PR #11929, opened 2019-07-28
*medic

Ty-Omaha:
**What does this PR do:**
Allows humans to call to get doctors' and medics' attention by emoting *medic

will be helpful in xeno outbreaks

make sure sound is on
https://streamable.com/ovoh1

credit to CM

**Changelog:**
🆑 
add: Humans may now use an emote to call for a *medic!
/ 🆑


Breenland:
I'd like it so much more if we could get an original sound clip. This doesn't seem to fit well to me, likely because it's the same one used on CM and it doesn't really fit this non-military setting in my mind.

AffectedArc07:
Its got the heavy in it
Its a +1 from me

Citinited:
I'm gonna bind this to ctrl-E and spam it every time I'm in medbay

KasparoVy:
The weirdest part will be hearing Vox or tree people using this emote. The voice just doesn't match hahaha

Ty-Omaha:
@KasparoVy it's humans only to keep everything immersed

KasparoVy:
My bad I'm dumb, good thinking

MRgaty:
I don't like it being human only, as other races can get as infected as humans, also sounds a bit... Weird... Like it gives the same voice to every human.

Ralta:
I'd rather remove the audio and just have it send a message to chat. This is just going to get spammed by people in medbay _constantly_ otherwise, and it will make medbay unbearable. I'm sure it'll happen in the brig too - inmates screaming `MEDIC!` to get attention, or piss off Security.

PidgeyThePirate:
Can we put up a little bubble like the typing bubble for a few seconds with a small red cross instead? That way every race can use it and it's relatively unintrusive.

Fox-McCloud:
This is really gamey and incredibly memey.

I don't support this.

This ain't TF2.

KasparoVy:
Where's the dispenser at?

Tayyyyyyy:
"medic" seems to be a military term. You wouldn't walk into a hospital and call for a medic.

taukausanake:
> You wouldn't walk into a hospital and call for a medic

Not unless you speak in memes. Speaking of which, why not get a Genji "I need healing" sound clip too?

Spacemanspark:
Characters have different voices, accents and tones. I don't think this fits well here, and don't see the need in the first place--people that feel they need to be 'faster' can custom make a macro.

variableundefined:
I'm against this PR for the same reasons as fox. This seems likely to be abused too.

Closing the PR.

Issue #7300, opened 2017-05-13
Anti drop implants result in undroppable items (shocking)


Anoonki:
**Problem Description**:
After being forced to drop an item that had nodrop, as part of my antidrop implant. (Hands got knocked off as an IPC), (After getting my hands fixed) I picked the item back up, and couldn't relax my hands or drop the item.

**What did you expect to happen**: 
For my anti-drop implant to be able to be turned off as normal.

**What happened instead**:
Super glued a baton and teleshield to my hands, turning me into Harmbaton law enforcement

**Why is this bad/What are the consequences:**
I lost use of my hands until an admin came and Varedited my baton and shield to be droppable.

**Steps to reproduce the problem**:
Spawn as an IPC, acquire an antidrop, pick up two items of your choice, activate your antidrop, damage your hands until they come off, reattach your hands, (Deactivate your Antidrop implant before picking up dropped items)pick up dropped items, have them replace your hands for eternity.

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was
anything else out of the ordinary happening?)**: Should be covered in the rest of the report. [But have this image too.](http://i.imgur.com/PfOL23v.png)

Anoonki:
I looked and couldn't find a duplicate issue, so if this' already reported (which I imagine it is), lemme know to close this.

Tayyyyyyy:
I ran into this issue too when I was decapitated, losing the skill in the upper left corner. When my head was reattached, activating the no-drop spell didn't actually affect me.

TullyBurnalot:
Cannot reproduce.

Spawned as IPC, picked up items, activated anti-drop, got my hands chopped off, reattached hands, picked up items again, could freely turn off anti-drop and drop all items.

Anoonki:
Odd, maybe I'm missing a step, but Flattest (and I think Marsmond) ended up having to fix it for me. 
Given what they were telling me, the Anti-drop implant was toggling on and off, but my items just had a permanent undroppable value of 2 until manually varedited otherwise.

Anoonki:
@TullyBurnalot Found the missing link after spending like half an hour testing. After getting repaired and my hand reattached, I deactivated the antidrop implant *before* picking the dropped item(s) back up [As seen in this image](http://i.imgur.com/T0k91UC.png) (At the end, after I relax my hands, then get the notification "not holding anything, hands relaxed" I picked the items up and started toggling the implant to keep relaxing my hands but not really.)

[...]

PR #11212, opened 2019-04-02
Dropped anti drop items now lose the antidrop status if not initialized with it

farie82:
**What does this PR do:**
If an item magically drops (or drops due to surgery) while it has nodrop it'll now get nodrop removed from it's flags if the item didn't have it when it was created.
There might be a few items that implement antidrop weirdly but those are usually admin spawned stuf.

Fixes: #7300

**Changelog:**
:cl:
fix: Items with NODROP that are dropped now lose the flag if they didn't get created with it
/:cl:

PR #11936, opened 2019-07-28
Dropped anti drop items now lose the antidrop status if not initialized with it

farie82:
What does this PR do:
If an item magically drops (or drops due to surgery) while it has nodrop it'll now get nodrop removed from it's flags if the item didn't have it when it was created.
There might be a few items that implement antidrop weirdly but those are usually admin spawned stuf.

Fixes: #7300

Remake of #11212 Git fucked up so yeah

Changelog:
🆑
fix: Items with NODROP that are dropped now lose the flag if they didn't get created with it
/🆑

### DIFF
--- a/code/datums/spells/horsemask.dm
+++ b/code/datums/spells/horsemask.dm
@@ -36,7 +36,7 @@
 		return
 
 	var/obj/item/clothing/mask/horsehead/magichead = new /obj/item/clothing/mask/horsehead
-	magichead.flags |= NODROP		//curses!
+	magichead.flags |= NODROP | DROPDEL	//curses!
 	magichead.flags_inv = null	//so you can still see their face
 	magichead.voicechange = 1	//NEEEEIIGHH
 	target.visible_message(	"<span class='danger'>[target]'s face  lights up in fire, and after the event a horse's head takes its place!</span>", \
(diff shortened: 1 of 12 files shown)


Bram961:
Do they also become useless?

farie82:
The items will lose the nodrop trait if they didn't have it upon spawning. I've made edge case items delete themselves or do something else. Crude info on them in the previous PR.

The items will still be usefull as in they behave as normal. Just without nodrop

Bram961:
> > Do they also become useless?
> 
> The items will lose the nodrop trait if they didn't have it upon spawning. I've made edge case items delete themselves or do something else. Crude info on them in the previous PR.
> 
> The items will still be usefull as in they behave as normal. Just without nodrop

So essentially what you've made is a feature whereby someone can just go to robotics to get the limb replaced with a robotic part so they can use the no-drop item freely without the downside?

farie82:
The only issues will come from custom admin made items. Items that do not spawn with nodrop. But are given nodrop by admins. However these are edge cases and admins can just add DROP_DEL to it as well then.

Couls:
wouldn't it be better to just override the /obj/item/dropped proc instead of doing it for each individual item?

farie82:
> wouldn't it be better to just override the /obj/item/dropped proc instead of doing it for each individual item? (check if it has the NODROP flag and if it does remove it when it's dropped)

I currently can't remember why I've overriden it for the ninja stuff. Since the item/dropped proc will handle those as well. I will remove the ninja code since it's not needed

Issue #11890, opened 2019-07-18
Wiring a crate breaks borg cable coils


enderman51505:
**Problem Description**:
apon rigging a crate with cable coils and having none left causes it not to regenerate when charging

**Why is this bad/What are the consequences:**
it takes a rather large part of a engiborgs kit away

**Steps to reproduce the problem**:
1: have less then or equal too the amount of wires to rig a crate
2: rig the crate
3: cable coil dissapears and doesnt come back

SteelSlayer:
I was able to recreate this bug

PR #11937, opened 2019-07-28
Rigging a crate now costs 15 coils instead of deleting your stack

farie82:
**What does this PR do:**
Rigging a crate now costs 15 coils instead of deleting your stack.
Makes borgs also able to use it again.

Fixes: #11890

**Changelog:**
:cl:
fix: Rigging a crate now doesn't delete your coil stack anymore but instead uses 15 of it. Borgs can use it safely again
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -5,9 +5,9 @@
 	icon_state = "crate"
 	icon_opened = "crateopen"
 	icon_closed = "crate"
-	climbable = 1
+	climbable = TRUE
 //	mouse_drag_pointer = MOUSE_ACTIVE_POINTER	//???
-	var/rigged = 0
+	var/rigged = FALSE
 	var/obj/item/paper/manifest/manifest
 	// A list of beacon names that the crate will announce the arrival of, when delivered.
 	var/list/announce_beacons = list()
@@ -23,16 +23,16 @@
 		overlays += "manifest"
 
 /obj/structure/closet/crate/can_open()
-	return 1
+	return TRUE
 
 /obj/structure/closet/crate/can_close()
-	return 1
+	return TRUE
 
 /obj/structure/closet/crate/open()
 	if(src.opened)
-		return 0
+		return FALSE
 	if(!src.can_open())
-		return 0
+		return FALSE
 
 	if(rigged && locate(/obj/item/radio/electropack) in src)
 		if(isliving(usr))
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -154,6 +154,8 @@
 	else if(istype(W, /obj/item/stack/packageWrap))
 		return
 	else if(istype(W, /obj/item/stack/cable_coil))
+		if(isrobot(user))
```
Citinited: This is a disgusting hack BTW; you should really find out what the underlying issue is and fix that, rather than applying bandaid solutions like this.

Plus, if other things exhibit similar behaviour like this in the future, you'll just have to apply more bandaids like this instead of fixing the actual code that breaks it

farie82: Fair point. Looking at it a bit better I know another fix. Bit to zealous here :P

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -154,14 +154,15 @@
...
-		to_chat(user, "<span class='notice'>You rig [src].</span>")
-		user.drop_item()
-		qdel(W)
-		rigged = 1
-		return
+		if(C.use(15))
+			to_chat(user, "<span class='notice'>You rig [src].</span>")
+			rigged = 1
+		else 
+			to_chat(user, "<span class='warning'>You need atleast 15 wires to rig [src]!</span>")
```
Couls: should add a return after this else statement

farie82: In this case it isn't needed due to the else if structure. But I'll include it to be more readable

Fox-McCloud:
Might want to change the CL and description and title.

farie82:
> Might want to change the CL and description and title.

oh whoops

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -154,14 +154,16 @@
...
 			to_chat(user, "<span class='notice'>[src] is already rigged!</span>")
 			return
-		to_chat(user, "<span class='notice'>You rig [src].</span>")
-		user.drop_item()
-		qdel(W)
-		rigged = 1
-		return
+		if(C.use(15))
+			to_chat(user, "<span class='notice'>You rig [src].</span>")
+			rigged = 1
```
variableundefined: TRUE

farie82: good catch

Issue #11883, opened 2019-07-18
Runtime error on advpinpointer caused server crash.


xProlithium:
**Problem Description**:
Last runtime log shown before server crash is
`[2019-07-18T04:26:59] Runtime in code/game/gamemodes/nuclear/pinpointer.dm,175: DEBUG: to_chat called with invalid message/target.`
`[2019-07-18T04:26:59]   Message: '(null)'`
`[2019-07-18T04:26:59]   Target: 'Fiddler', /mob/living/carbon/human`
`[2019-07-18T04:26:59]   src: The advanced pinpointer (/obj/item/pinpointer/advpinpointer)`
`[2019-07-18T04:26:59]   src.loc: Fiddler (58,156,4) (/mob/living/carbon/human)`
`[2019-07-18T04:26:59]   usr: Fiddler (muhdi345) (/mob/living/carbon/human)`
`[2019-07-18T04:26:59]   usr.loc: The floor (58,156,4) (/turf/simulated/floor/plasteel)`

**What did you expect to happen**: 
Advanced Pinpointer to locate Red Documents.

**What happened instead**:
Server hard crashed.

**Why is this bad/What are the consequences:**
Premature round ending. Possibly exploitable.

**Steps to reproduce the problem**:
Somehow targetting an invalid person or item with the pinpointer.

**When did the problem start happening**:
The issue was recorded on 7/18/2019 at 4:26 GMT.

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was anything else out of the ordinary happening?)**:
It was a traitor round, mid round.

Muhdi345:
I'll try to add to this, I turned my pinpointer on first and then set it to the red docs. And then I tried to stash it in my satchel. And then I killed the server... Somehow....

PR #11938, opened 2019-07-28
Fixes the advanced pinpointer crashing the server

farie82:
**What does this PR do:**
Stops the advanced pinpointer from crashing the server

Fixes: #11883

**Changelog:**
:cl:
fix: The syndicate can't use meta warfare no more. Advanced pinpointers no longer crash the server
/:cl:

### DIFF
--- a/code/game/gamemodes/nuclear/pinpointer.dm
+++ b/code/game/gamemodes/nuclear/pinpointer.dm
@@ -168,7 +168,7 @@
 					
 					var/list/target_candidates = get_all_of_type(item_paths[targetitem], subtypes = TRUE)
 					for(var/obj/item/candidate in target_candidates)
-						if(!is_admin_level(candidate.loc.z))
+						if(!is_admin_level((get_turf(candidate)).z))
 							target = candidate
 							break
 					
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/nuclear/pinpointer.dm
```diff
@@ -165,10 +165,13 @@
...
 					if(!targetitem)
 						return
-					var/list/obj/item/target_candidates = get_all_of_type(item_paths[targetitem], subtypes = TRUE)
-						for(var/obj/item/candidate in target_candidates)
-							if(!is_admin_level(candidate.loc.z))
-								target = candidate
+					
+					var/list/target_candidates = get_all_of_type(item_paths[targetitem], subtypes = TRUE)
+					for(var/obj/item/candidate in target_candidates)
+						if(!is_admin_level(candidate.loc.z))
```
Ty-Omaha: shouldn't this use get_turf anyways instead?

Couls: get_turf wouldn't give you the z level

farie82: Damn you just found another bug that caused documents and the like not to show up on the pinpointer

farie82: Update. Doesn't seem to have fixed that

PR #11942, opened 2019-07-29
Waddling component, clowns can now choose to waddle while they walk, penguins automatically waddle

Couls:
**What does this PR do:**
Port over TG's waddling component and allow clowns to pick whether or not they waddle while they walk

ezgif com-gif-maker

**Changelog:**
:cl:
add: TG waddle component, clowns can now optionally waddle, penguins always waddle. Ctrl Click clown shoes in hand to toggle waddling
/:cl:

### DIFF
--- a/code/modules/clothing/shoes/magboots.dm
+++ b/code/modules/clothing/shoes/magboots.dm
@@ -87,7 +87,7 @@ obj/item/clothing/shoes/magboots/syndie/advance //For the Syndicate Strike Team
 	if(!isliving(user))
 		return
 	if(user.get_active_hand() != src)
-		to_chat(user, "You must hold the [src] in your hand to do this.")
+		to_chat(user, "You must hold [src] in your hand to do this.")
 		return
 	if(!enabled_waddle)
 		to_chat(user, "<span class='notice'>You switch off the waddle dampeners!</span>")
--- a/code/modules/clothing/shoes/magboots.dm
+++ b/code/modules/clothing/shoes/magboots.dm
@@ -71,6 +71,30 @@ obj/item/clothing/shoes/magboots/syndie/advance //For the Syndicate Strike Team
 	silence_steps = 1
 	shoe_sound = "clownstep"
 	origin_tech = "magnets=4;syndicate=2"
+	var/enabled_waddle = TRUE
+	var/datum/component/waddle
+
+/obj/item/clothing/shoes/magboots/clown/equipped(mob/user, slot)
+	. = ..()
+	if(slot == slot_shoes && enabled_waddle)
+		waddle = user.AddComponent(/datum/component/waddling)
+
+/obj/item/clothing/shoes/magboots/clown/dropped(mob/user)
+	. = ..()
+	QDEL_NULL(waddle)
+
+/obj/item/clothing/shoes/magboots/clown/CtrlClick(mob/living/user)
+	if(!isliving(user))
+		return
+	if(user.get_active_hand() != src)
+		to_chat(user, "You must hold the [src] in your hand to do this.")
+		return
+	if(!enabled_waddle)
+		to_chat(user, "<span class='notice'>You switch off the waddle dampeners!</span>")
+		enabled_waddle = TRUE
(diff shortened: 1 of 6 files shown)


code/modules/clothing/shoes/miscellaneous.dm
```diff
@@ -63,6 +63,30 @@
...
+	QDEL_NULL(waddle)
+
+/obj/item/clothing/shoes/clown_shoes/CtrlClick(mob/living/user)
+	if(!isliving(user))
+		return
+	if(user.get_active_hand() != src)
+		to_chat(user, "You must hold the [src] in your hand to do this.")
+		return
+	if(!enabled_waddle)
+		to_chat(user, "<span class='notice'>You switch off the waddle dampeners!</span>")
```
Citinited: Shouldn't these messages be the other way around?

code/modules/clothing/shoes/miscellaneous.dm
```diff
@@ -63,6 +63,30 @@
 	item_color = "clown"
 	var/footstep = 1	//used for squeeks whilst walking
 	shoe_sound = "clownstep"
+	var/enabled_waddle = TRUE
+	var/datum/component/waddle
```
Citinited: I don't think this var is actually called anywhere in the code, so it can be removed right?

code/modules/clothing/shoes/miscellaneous.dm
Couls: Derp, good catch

code/modules/clothing/shoes/miscellaneous.dm
Couls: Need to change the code to use the holder, another good catch thanks!

MRgaty:
Only works if the clown walks?

davidchan13:
Make sure this applies to Cluwnes as well.

This would be hilarious if it was tied to the clown trait as well as induced by the clown virus. Entire station starts to waddle and honk hysterically.

Couls:
Applies to all clown shoe types, except chameleon and magboots and yeah they only waddle when they walk

Citinited:
Give this to people affected by jestosterone please 👍

Ralta:
> Give this to people affected by jestosterone please 👍

It looks like it's applied to the shoes themselves, so it should work for anyone who wears them.

Citinited:
> It looks like it's applied to the shoes themselves, so it should work for anyone who wears them.

That's true, but I mean if you take jestosterone, IMO it should also add the jestosterone component to you — regardless of if you're wearing clown shoes. :smile:

datlo:
This should be added to clown magboots, theyre a traitor item meant to look the same as regular ones.

code/modules/clothing/shoes/miscellaneous.dm
Couls: Actually no, dampening is the act of cancellation this was right the first time

Citinited:
Cheers 👍 

Can you check if removing clown shoes while under the effects of jestosterone removes the waddling? I'm not too familiar with components to check that this is the case from reading the code. If so, can you see about changing that?

Couls:
In theory, it shouldn't but it wouldn't hurt to test it

code/modules/clothing/shoes/magboots.dm
```diff
@@ -71,6 +71,30 @@ obj/item/clothing/shoes/magboots/syndie/advance //For the Syndicate Strike Team
...
+
+/obj/item/clothing/shoes/magboots/clown/dropped(mob/user)
+	. = ..()
+	QDEL_NULL(waddle)
+
+/obj/item/clothing/shoes/magboots/clown/CtrlClick(mob/living/user)
+	if(!isliving(user))
+		return
+	if(user.get_active_hand() != src)
+		to_chat(user, "You must hold the [src] in your hand to do this.")
```
Fox-McCloud: no `the`

Couls: fixed

PR #11944, opened 2019-07-29
Lavaland changes

Emanthealmighty:
**What does this PR do:**
Changes a few things in lavaland.
Ash walkers are now slightly slower but more durable in terms of brute and burn damage.
Ash walkers can now dig rocks without a pickaxe, however, it takes longer.
Melee and ranged swarmers are now slightly slower so they don't chase you around with nothing you can do about it.
Watchers reworked to be less tedious to fight with melee weapons.

These are all changes I have decided to make after playing for some time as an ash walker in lavaland. They'll not only make the whole experience a bit more fun, but also change the walkers to act more like tough lavaland mobs, which can take quite a beating and tunnel around without the need of any tools, rather than just being fast unathis with a different name.

Specifics:
Ash Walkers can now be defined with "isashwalker" for future ash walker only things.
Ash walkers' slowdown changed to -0.90 from -0.80, making them slower.
Ash walkers now have a brute and burn modifiers of 0.8, making them more resistant to those damage types.
Ranged and melee swarmers now have a speed of 1, from 0, making them slower.
Watchers now have a lower damage of 8, and a higher damage of 13, from both of them being 15.
Watcher projectile changes:
Speed was set to 1.6 from 1.0, making it a bit slower and easier to dodge.
Now deals 3 burn damage per hit.
Temperature changed to 225 from 50, making it have less of a freezing effect on its victim.
Magmawing projectile deals 10 burn damage from 5, icewing deals 7 burn damage from 5.

**Changelog:**
:cl:
add: Ash walkers can now dig into rocks without pickaxes at the cost of taking longer to do so.
tweak: Ash walkers have been slowed down, however they are still faster than normal humans.
tweak: Ash walkers can now sustain more brute and burn damage.
tweak: Ranged and melee swarmers are now slower.
balance: Rebalanced watchers to be slightly less tedious to fight without ranged weapons.
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -168,6 +168,7 @@
 #define isskeleton(A) (is_species(A, /datum/species/skeleton))
 #define ishumanbasic(A) (is_species(A, /datum/species/human))
 #define isunathi(A) (is_species(A, /datum/species/unathi))
+#define isashwalker(A) (is_species(A, /datum/species/unathi/ashwalker))
 #define istajaran(A) (is_species(A, /datum/species/tajaran))
 #define isvulpkanin(A) (is_species(A, /datum/species/vulpkanin))
 #define isskrell(A) (is_species(A, /datum/species/skrell))
--- a/code/game/turfs/simulated/minerals.dm
+++ b/code/game/turfs/simulated/minerals.dm
@@ -48,6 +48,14 @@
 	setDir(angle2dir(rotation + dir2angle(dir)))
 	queue_smooth(src)
 
+/turf/simulated/mineral/attack_hand(mob/living/carbon/human/M)
+	if(isashwalker(M)) // Ash walkers can dig without a pickaxe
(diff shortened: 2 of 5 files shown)


Quantum-M:
Probably too strong of a buff, but what about making ash walkers immune to ash storms? Considering that other lavaland mobs native to the place are immune to it.

datlo:
I wouldnt make ashwalkers completely immune to storms. Maybe some resistance but not immune. Ashstorms force ashwalkers to stay near their nest/a safe area which imo is better for gameplay than just letting them free roam lavaland

Couls:
dunno if you need to nerf the fauna, lavaland is meant to be difficult and honestly i've been seeing too many people come back from lavaland with insane loot as opposed to situations where miners wouldn't come back at all

variableundefined:
Ashwalkers are meant to have strength in number, instead of being strong individually, they don't need buffs.

PR #11948, opened 2019-07-30
Blobbernaut major changes (player control!)

Emanthealmighty:
**What does this PR do:**
Blobbernauts can now be controlled by players through a prompt which appears when they are spawned by blobs.
Blobbernauts have been nerfed, they now have less health, deal less damage and cannot break walls anymore.
Blobbernauts now regenerate health overtime when standing on blob structures.
Blobbernauts will start losing health overtime if they are not standing on blob structures and they are not at full health.
Blobbernauts and blobs can now communicate with each other.
Blob's "Produce Blobbernaut" ability has been made more expensive to use, now costing 60 resources instead of 20.

Blobbernauts are now smart "minions" the blobs can rely on. They are best used defensively as they can regenerate health when they're on blob structures, but they can still be used to do other strategic things, such as moving out in groups and attacking some specific department, or killing unsuspecting people and dragging them back to the blob to be zombified, etc. 

Specifics:
Blobbernauts now have 200 health down from 240.
They now deal 10 lower/15 higher damage down from 20/20.
Their environment smash has been set to "SMASH_STRUCTURES" from "SMASH_RWALLS" which is the same as blob spores now.
They now heal by 0.25 HP every tick if they are standing on blob structures.
They now lose health by 0.2 HP every tick if they are not standing on blob structures and they are not at max health. (Until you lose a bit of health, your integrity will remain the same no matter where you are)
They now have a random ID next to their name which can go from 0 to 1000, this is the same system used by monkeys, aliens, etc to help identify them.
They now have a verb which lets them type what they want to say in the blob chat. Said message will reach blobs and blobbernauts alike. Blobs' overmind chat has also been changed to send the messages to blobbernauts too.

**Changelog:**
:cl: EmanTheAlmighty
add: Blobbernauts can now be controlled by players through a prompt which appears when they are spawned by blobs.
add: Blobbernauts and blobs can now communicate with each other.
add: Blobbernauts now regenerate health overtime when standing on blob structures.
add: Blobbernauts will slowly lose health if they are not standing on blob structures while not at full health.
balance: Blobbernauts have been nerfed, they now have less health, deal less damage and cannot break walls anymore.
balance: Blob's "Produce Blobbernaut" ability has been made more expensive to use, now costing 60 resources instead of 20.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -184,8 +184,8 @@
 /mob/living/simple_animal/hostile/blob/blobbernaut/Life(seconds, times_fired)
 	if(stat != DEAD && (getBruteLoss() || getFireLoss())) // Heal on blob structures
 		if (locate(/obj/structure/blob) in get_turf(src))
-			adjustBruteLoss(-0.5)
-			adjustFireLoss(-0.5)
+			adjustBruteLoss(-0.25)
+			adjustFireLoss(-0.25)
 
 /mob/living/simple_animal/hostile/blob/blobbernaut/blob_act()
 	return
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -183,9 +183,12 @@
 
 /mob/living/simple_animal/hostile/blob/blobbernaut/Life(seconds, times_fired)
 	if(stat != DEAD && (getBruteLoss() || getFireLoss())) // Heal on blob structures
-		if (locate(/obj/structure/blob) in get_turf(src))
+		if(locate(/obj/structure/blob) in get_turf(src))
 			adjustBruteLoss(-0.25)
 			adjustFireLoss(-0.25)
+		else
+			adjustBruteLoss(0.2) // If you are at full health, you won't lose health. You'll need it. However the moment anybody sneezes on you, the decaying will begin.
+			adjustFireLoss(0.2)
 
 /mob/living/simple_animal/hostile/blob/blobbernaut/blob_act()
 	return
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -163,10 +163,10 @@
 	icon_state = "blobbernaut"
 	icon_living = "blobbernaut"
 	icon_dead = "blobbernaut_dead"
-	health = 240
-	maxHealth = 240
-	melee_damage_lower = 20
-	melee_damage_upper = 20
+	health = 200
+	maxHealth = 200
+	melee_damage_lower = 10
+	melee_damage_upper = 15
 	obj_damage = 60
 	attacktext = "hits"
 	attack_sound = 'sound/effects/blobattack.ogg'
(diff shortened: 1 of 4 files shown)


JKnutson101:
Not sure about the specifics on balance, but I really like the idea of giving ghosts more ways to get involved in blob rounds. Currently all I can think of is getting really lucky and hopping in when the blob splits, or signing up for an ERT, if one is even called.

Couls:
Smart blobs can camp blob tiles to both heal and take cover and this might change blob in a huge way, Ive always preferred blobs giving ghost roles but you have to keep in mind the impact this could have

AdamElTablawy:
@Couls 
in practice on /tg/ this didn't significantly change blob playstyle aside from encouraging more blobbernaut use 

that being said this PR is missing a few features

first and foremost blobbernauts as this pr makes them are able to go away from their boss blob independently. this is a bad idea as tg learned real fast. you'll want to force them to remain within some distance of a blob tile or they'll just run off and start beating the medbay to death.

Emanthealmighty:
> @Couls
> in practice on /tg/ this didn't significantly change blob playstyle aside from encouraging more blobbernaut use
> 
> that being said this PR is missing a few features
> 
> first and foremost blobbernauts as this pr makes them are able to go away from their boss blob independently. this is a bad idea as tg learned real fast. you'll want to force them to remain within some distance of a blob tile or they'll just run off and start beating the medbay to death.

They have less health and damage than they did before. If they run off from the blob, they won't have anywhere to run off to to heal. They'll be an easy target.

AdamElTablawy:
@Emanthealmighty that's not much of an obstacle, considering it's a measly 40 points and an escaped blobbernaut in the back lines is going to be causing some trouble that can distract people from fighting the blob

take it from experience-that argument got made on /tg/ too, but four-man gangs of blobbernauts running into science and beating the guy doing R&D to death shouting UNGA BUNGA is only funny the first seven times

Emanthealmighty:
> @Emanthealmighty that's not much of an obstacle, considering it's a measly 40 points and an escaped blobbernaut in the back lines is going to be causing some trouble that can distract people from fighting the blob
> 
> take it from experience-that argument got made on /tg/ too, but four-man gangs of blobbernauts running into science and beating the guy doing R&D to death shouting UNGA BUNGA is only funny the first seven times

I want to keep the aspect of being able to go behind the battlefield and hit the unprepared. I'll just raise the amount of points it costs so you can't waste them as easily only to then tell the blobbernaut to go out and get themselves killed while beating people up.

Lunaimon:
I really think they should have something like purple Terror Spiders - which die off if they're away for their queen for too long, but still have enough time to do things like assassinations if they want.

Taaketa:
Keep the backline aspect but have them steadily lose health if they do not remain on a Blob tile. Restore their health value to original levels and consider that extra hp their drain gap.

Emanthealmighty:
Blobbernauts will now start losing health if they are below their max health while outside of blob structures. 

This means that a full health blobbernaut can retain its health outside of the blob as long as they are not damaged in any way. This makes it so they can "travel" away and retain their health, to then use it all on killing and ravaging whatever they want. (So that they don't get to the place they want to be in with half health)

davidchan13:
Blob should be expanding to allow Blobnauts to push, not the other way around. They absolutely should lose health when not on blob tiles, otherwise we'll get blobnaughts scattering about the station smashing every window they can find to vent the station.

Emanthealmighty:
> Blob should be expanding to allow Blobnauts to push, not the other way around. They absolutely should lose health when not on blob tiles, otherwise we'll get blobnaughts scattering about the station smashing every window they can find to vent the station.

First off, they now cost as much as it takes to actually build a factory, which is a lot. It would take a long time for there to be "blobbernauts scattering about", at which point it'll just be the crew's fault.

Second, if they want to scatter about, they need to actually "leave" the vicinity of the blob, which most certainly isn't unguarded. All it takes is one hit for the blobbernauts to start decaying overtime outside of the blob. So they DO lose health when not on blob tiles, HOWEVER:

If they manage to leave the blob's vicinity and sneak past all the crew members gone rabid, rushing from every direction to kill this one blob, then they deserve to not be losing any health. When they do the thing they wanted to do, they will also have to GET  BACK to the blob while also losing health constantly and definitely getting chased by angry people. 

It's a pretty risky and an overall not-worth-it investment to buy a blobbernaut and tell it to run off.
So it won't happen in most cases.

code/game/gamemodes/blob/blobs/blob_mobs.dm
```diff
@@ -175,18 +175,48 @@
...
-	environment_smash = ENVIRONMENT_SMASH_RWALLS
+	environment_smash = ENVIRONMENT_SMASH_STRUCTURES
 	pressure_resistance = 100    //100 kPa difference required to push
 	throw_pressure_limit = 120  //120 kPa difference required to throw
 	see_in_dark = 8
 	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE
 
+/mob/living/simple_animal/hostile/blob/blobbernaut/Life(seconds, times_fired)
+	if(stat != DEAD && (getBruteLoss() || getFireLoss())) // Heal on blob structures
+		if (locate(/obj/structure/blob) in get_turf(src))
```
variableundefined: No space between if and (

Kyep:
I looked at the first round this was used in.
2/3 of the blobbernauts died while under AI control, before a player could even take control of them.

IMHO, it might be a good idea to figure out which player will control the blobbernaut BEFORE the blobbernaut is spawned. Or at the very least, don't make their AI charge at enemies before a player takes control.

Emanthealmighty:
I'm thinking of shortening the time before a player is pulled and making them stand while the poll is going on.

PR #11949, opened 2019-07-30
Hierophant layer fix

### LINKED ISSUE #11947: Hierophant abilities prevent clicking on yourself
**Problem Description**:
The particle effects of the hierophant abilities have a higher overlay priority than that of a player.

For the hierphant abilities to be a layer below the player so you can click on yourself with legion souls/healing patches.

**What happened instead**:
You cannot click on yourself if the hierphants abilities are currently on your same tile and you instead have to alt click your tile and then click on your player in the floor tab menu.

Its a speed intensive fight where you have to keep moving to avoid being hit.  Being unable to heal yourself on the fly means you have to stand still and are a sitting duck.

Fight the hierophant and try to click on yourself with a legion core/healing patch while its abilities are proccing on your tile.

Most likely was a problem that came along with the megafauna but was never noticed.

florkeger:
**What does this PR do:**
Adjusts the effect layer for the Hierophant blast to always be between a laying mob and a further above mob in order to make sure the effect doesn't ever apply over the mob. This is not really a fix, merely a bandaid for an issue that supposedly happens, but is hard to reproduce.

Fixes #11947

**Changelog:**
:cl: kazboo
fix: adjusts the hierophant blast layer to make for it to always be below the effect. this should be the case already due to mouse_opacity, though for some unknown reason, it just doesn't work sometimes, so this should fix it.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/hierophant.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/hierophant.dm
@@ -498,6 +498,7 @@ Difficulty: Hard
 	name = "vortex blast"
 	luminosity = 1
 	desc = "Get out of the way!"
+	layer = BELOW_OBJ_LAYER
 	duration = 9
 	var/damage = 10 //how much damage do we do?
 	var/list/hit_things = list() //we hit these already, ignore them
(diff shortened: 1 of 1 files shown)


Dave-TH:
Just adjust the mouse_opacity of the heirophant effect instead.

florkeger:
Considering the mouse_opacity doesn't work for an odd reason, I won't change anything about it and leave it as it is until someone figured out why mouse_opacity = 0 applied to /hierophant/blast doesn't allow for beneath items to be clickable.

florkeger:
After re-consideration and testing the reported bug countless times, we came to the conclusion that the values as they affect the hierophants effect are working as intended, and the bug couldn't be reproduced, at least not by me.

Should be observed more often and noted or something, but for now, I'll close this myself.

PR #11951, opened 2019-07-31
New damage type: Critical, for races that don't breathe

Couls:
**What does this PR do:**
Creates a new critical damage type for the special snowflakes so they don't take brute damage in crit instead, as brute is much harder to heal and I don't think these races should be punished for getting stabilized. The only way to currently accumulate critical damage is through admemes or being in critical, you take 1 critical damage while in crit per tick(this is on par with 1 oxyloss per tick in crit for oldcrit races that breathe), you also heal 5 critical damage every tick spent outside of crit which is again on par with breathing races. Epinephrine and Ephedrine lower critical damage by small amounts to simulate the same effect these chems have on oxyloss.

image

The new damage type appears in scanners, advanced scanners, sleepers and cryocells

**Changelog:**
:cl:
add: Unique critical damage type for non-breathing races
tweak: replaces paralysis % field in advanced scanner with critical damage field
/:cl:

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -475,7 +475,7 @@ GLOBAL_LIST_INIT(do_after_once_tracker, list())
 			if(DEAD)
 				status = "<font color='red'><b>Dead</b></font>"
 		health_description = "Status = [status]"
-		health_description += "<BR>Oxy: [L.getOxyLoss()] - Tox: [L.getToxLoss()] - Fire: [L.getFireLoss()] - Brute: [L.getBruteLoss()] - Clone: [L.getCloneLoss()] - Brain: [L.getBrainLoss()]"
+		health_description += "<BR>Oxy: [L.getOxyLoss()] - Tox: [L.getToxLoss()] - Fire: [L.getFireLoss()] - Brute: [L.getBruteLoss()] - Clone: [L.getCloneLoss()] - Brain: [L.getBrainLoss()] - Crit: [L.getCritLoss()]"
 	else
 		health_description = "This mob type has no health to speak of."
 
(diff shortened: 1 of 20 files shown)


JohnBigless:
I approve.

Ty-Omaha:
Crit is such a weird damage name

Shockpoint:
This appears to be a hard nerf to slimes in particular. They already take toxin damage in place of oxyloss, and seeing how bad medical is on a normal basis and the difficulty of fixing them with significant toxin damage as is, this will only make things worse. 

I suggest something that balances all nobreathe races for crit without hindering them in places they're already struggling.

MRgaty:
I would suggest taking a picture of the analysis of a slime that died during crit, so we can see how much damage do they get in total.

Couls:
They would take 100 extra brute damage at death if they died through crit, crit damage itself is not very difficult to heal. Using the normal methods of stabilizing(epi or ephe or atropine) all would work to help heal crit damage the issue with taking brute damage while in crit is that bones can be broken with a very small chance every tick you’re in crit and for things like IPCs you’ll see their limbs falling off. SR heals all crit damage and defibs heal the same amount of crit damage as they do oxyloss so current doctors can just treat slimes as they did before and the results would mostly be the same

Couls:
Ty do you have a better name suggestion? I didn’t know what else to call the damage no breathe races would take in crit

datlo:
'Shock'?

Couls:
That can be misleading as we already have the newcrit disease shock

davidchan13:
If it's only Slimes and IPC "integrity" might work. Otherwise Fortitude or Cohesion maybe?

JohnBigless:
Not really sure. For IPCs, it could perhaps be 'CCI' Damage, (Critical Chassis Integrity)

Ty-Omaha:
>Critical Chassis Integrity

that's worse

the word "critical" makes it weird

I don't have any other ideas

JohnBigless:
> > Critical Chassis Integrity
> 
> that's worse
> 
> the word "critical" makes it weird
> 
> I don't have any other ideas

What about (Terminal Chassis Integrity)?

Ty-Omaha:
if its just IPC then "Chassis Integrity" is better

Couls:
It’s not just IPC, diona and slimes can accumulate this as well

Kyep:
I'm against this. Our health system is already complicated enough without adding this.

Fox-McCloud:
Going to be closing this. I don't feel another damage type is necessary at this point---having a damage type behave exactly like breathing, only it's not breathing, just so non-breathing races can take it so behave and act like they have breathing----and have chems that heal this....really just kinda demonstrates and calls into question  *why* those races don't breathe to begin with.

PR #11859, opened 2019-07-15
Moves all stuns to status effects

Couls:
**What does this PR do:**
Rename Paralyze to Unconscious, Weaken to Knockdown. ports https://github.com/tgstation/tgstation/pull/28603 https://github.com/tgstation/tgstation/pull/28696 https://github.com/tgstation/tgstation/pull/28654
This is mainly framework code, on the surface everything remains the same but its necessary setup to make porting crawling slightly easier for me, it also improves code reability(Why the heck was paralysis knocking people unconscious??). All stuns are now in deciseconds instead of ticks

I've also added screen alerts for being stunned and knocked down but those can be very easily removed, images below.
image
image

**Changelog:**
:cl:
tweak: All stuns are now status effects
add: alerts for being knocked down and stunned
/:cl:

Definitely gonna need a testmerge for this one to make sure I didn't miss any stuns!

PR #11959, opened 2019-08-03
Stun system rework

Couls:
**What does this PR do:**
Port of https://github.com/tgstation/tgstation/pull/39967, uses code from #11859 overhauls how stuns work, description from TG's PR:
Moves canmove to a bitflag in a new variable called mobility_flags, that will allow finer grain control of what someone can do codewise, for example, letting them move but not stand up, or stand up but not move.

Adds Immobilize()d status effect that freezes movement but does not prevent anything else.
Adds Paralyze()d which is oldstun "You can't do anything at all and knock down).
Stun() will now prevent any item/UI usage and movement (which is similar to before).
Knockdown() will now only knockdown without preventing item usage/movement.
People knocked down will be able to crawl at softcrit-speeds(slower speeds in para's case)
Refactors some /mob variables and procs to /mob/living.
update_canmove() refactored to update_mobility() and will handle mobility_flags instead of the removed canmove

**I'm not a code nerd what does this mean?**: It means we now have a framework for crawling and stuns that cause you to fall over without dropping your items, this makes it possible to rework our disarm to become like TG's if thats wanted(shoving your attacker away from you into walls and tables instead of just an RNG drop) and change slips to make it so they make you fall but changing it so it doesn't drop your weapons and make you immobile, which still slows you down without being extremely effective in making you helpless as well.

Currently this code has small changes in terms of actual gameplay, the rest will come later after the design and balances ideas are fleshed out and discussed by the heads and maintainers.

**Okay so then what's changed?**: You can now crawl when you rest, getting back up takes a few seconds and has a progress bar, you no longer drop items when you rest so now you can lie down and do things like use your PDA. Disarms have also reworked, instead of stunning and making the person you knock down immobile, they now drop the item they were holding and are forced to crawl. this means if they're quick enough they can pick up the item they've dropped again, but if you're quicker you can steal it from right under them!

**Changelog:**
:cl:
tweak: overhaul stuns and move them to a new system
tweak: disarms now trip you for four seconds
/:cl:


taukausanake:
> you can lie down and do things like use your PDA

:heart: I love this. Now I can sleep with my space carp plushie.
Does this also mean you can tactical shoot at people?

Couls:
Yeah, it does but you move much slower which SHOULD make you an easier target. later on I can change the projectile code to hit people resting as well if tactical shooting is too powerful

taukausanake:
> change the projectile code to hit people resting

I wouldn't change that. If you're not clicking on your target then you shouldn't hit them. Likewise if you're trying to hit someone past a downed/resting body you will only hit the downed body and not your target. Imagine if you were trying to mow down a group of people. The first row of bodies would flop down and become bullet sponges. Sure you could probably make it discern between resting and knocked down but that's probably too much

Couls:
it's actually not too much, but I guess it depends on whether or not we keep using new crit, as putting someone in crit would allow you to mow down everyone behind them.

AffectedArc07:
image
image

You may be waiting a while for code review. Maints are busy as is, and this PR is a monster

Couls:
That being said, most of the changes are regex's and it was partially reviewed in the stun rework PR, so that should help make it a bit easier to go over. The biggest change is just the update_mobility proc replacing update_canmove and a lot of the checks for incapacitation and lying down

Ralta:
Nice. The cooldown on crawling is good, even if it's on the getting up; prevents people bunny-hopping around.

This looks like a good and sensible change to differentiate paralysis versus stun, too.

SteelSlayer:
I like some of the stuff on here but what I don't like is this: **"and change slips to make it so they make you fall but changing it so it doesn't drop your weapons and make you immobile."** While this nerfs lube which I like, it also nerfs soap and water slips, if I'm understanding correctly.

Soap and water slips are great tatics to be used by antags when they want to stun/kidnap someone. Now you're forced to get a stunprod or some other "real" stunning tool if you want to actually stun someone. And if we're going with this change, I think slips should at least make you drop your held items.

Couls:
That change actually isn’t in this PR, it was more what the framework would be capable of in the future. Though I don’t think slipping people should be a viable tactic for kidnapping I’m more under the impression that slipping should be a defensive tool used to slow down people chasing you, if you do want to kidnap you’re gonna need to get your hands dirty with something like a stunprod, also you’re probably right and slipping should probably cause you to still drop your items but not make you unable to pick them back up

Ralta:
>...lube making you drop your items and slip while normal water would just cause you to fall over. Definitely something to think about

This is what I would prefer, personally. Though I appreciate it's not in this PR, I feel that it's a more balanced change.

code/datums/status_effects/debuffs.dm
```diff
@@ -1,3 +1,110 @@
+//Largely negative status effects go here, even if they have small benificial effects
```
Couls: Most of this PR is regex outside of here,

code/modules/mob/living/status_procs.dm
```diff
@@ -38,7 +38,7 @@
 	*	Dizzy						*
 			The screen goes all warped
 	*	Drowsy
-			You begin to yawn, and have a chance of incrementing "Paralysis"
+			You begin to yawn, and have a chance of incrementing "unconscious"
```
Couls: here(and all changes in file below)

code/modules/mob/living/update_status.dm
```diff
@@ -75,45 +64,86 @@
 	return eye_blurry
 
 //Updates canmove, lying and icons. Could perhaps do with a rename but I can't think of anything to describe it.
-/mob/living/update_canmove(delay_action_updates = 0)
-	var/fall_over = !can_stand()
-	var/buckle_lying = !(buckled && !buckled.buckle_lying)
-	if(fall_over || resting || stunned)
-		drop_r_hand()
-		drop_l_hand()
+/mob/living/proc/update_mobility()
```
Couls: here is basically how the new mobility stuff is handled, replaced update_canmove with update_mobility.

code/modules/mob/mob_helpers.dm
```diff
@@ -388,24 +388,36 @@ var/list/intents = list(INTENT_HELP,INTENT_DISARM,INTENT_GRAB,INTENT_HARM)
...
 
 	if(!resting)
-		client.move_delay = world.time + 20
-		to_chat(src, "<span class='notice'>You are now resting.</span>")
-		StartResting()
+		set_resting(TRUE, FALSE)
 	else if(resting)
-		to_chat(src, "<span class='notice'>You are now getting up.</span>")
-		StopResting()
+		if(do_after(src, 10, 0, src, 1, null, MOBILITY_MOVE))
```
Couls: resting now has an do_after

code/modules/mob/status_procs.dm
```diff
@@ -111,14 +111,19 @@
 /mob/proc/AdjustLoseBreath()
 	return
 
+/mob/proc/IsUnconscious() //non-living mobs shouldn't be unconscious
```
Couls: Mainly just proc renames here

code/modules/mob/mob_helpers.dm
KasparoVy: RIP matrixing under lasers and bullets once again 🤣

Couls: We'll have the issue of people crawling around, as bullets won't hit them. Til the maints or staff tell me how they wanna go with that at least. It wouldn't be hard to make bullets hit people regardless of resting or not but I didn't include any of those changes in this PR so you can crawl around to dodge bullets at the cost of being slower.

KasparoVy: Does this PR add the ability to crawl voluntarily while rested, or only while stunned?
It makes sense that crawling to avoid projectiles lazily aimed for center of mass. Perhaps down the road it could have some rng to see whether the shot 'dipped' to hit a crawler as it passed the turf.

Couls: It adds in the ability to crawl voluntarily while rested. You can even use items while rested there's no restrictions currently. Though I think a nice compromise to the system between "no you can't duck under bullets" and "yes you can lie down and machine gun people" would just be to restrict gun/ranged weapon usage on the floor, and then just having it as a melee thing. of course you could still throw stuff but it'd be cool to see someone with a stun baton slide under some bullets and then whack someone

KasparoVy: Lmfao! Great. Thanks for the quick response, it would be pretty badass-- but that's assuming the guy doesn't actually click you dead-on to score actual hits 😄 not something I myself can pull off but I know there's sharpshooters among us

Couls: Would make the moment even better considering that would even be a possibility. No problem with the responses it's fun talking to someone about code and stuff. Feel free to comment on any of my other PRs or anything and i'll be happy to reply when I can get to it

PR #11961, opened 2019-08-04
Limb and stamina damage rework

Couls:
**What does this PR do:**
Port of a few PRs from TG: https://github.com/tgstation/tgstation/pull/38436 , https://github.com/tgstation/tgstation/pull/39304, https://github.com/tgstation/tgstation/pull/36274

**Things i've changed from these PRs**: hands and feet now only contribute to 50% of your health, if you deal 30 brute damage to someones hand they will only take 15 damage for the purposes of crit and death. My reasoning for this being that non-vital limbs have no downside to targetting at the moment and theres no reason not to aim for the hands and feet because not only can you break the limb but you also do full damage. this change also makes sense in terms of logic as dying from hurting your hand was kind of strange. Arms and legs contribute to 75% of your total health and chest head and groin do 100% damage in terms of your total health. This means its better to aim for the legs and arms if you want to weaken someone without killing them but aim for the head if you want to lethal them. Disabled arms means you drop items as if the arm was broken, disabled hands means you can't use your hands at all(it treats them as if they were dismembered but they can still be used for cuffing)

**Changes listed for those who don't want to read the PRs above**: Limbs can now take stamina damage, when a limb takes maximum stamina damage(this varies for each limb) the limb becomes disabled and unusable, this encourages security to use disablers more often as it will slow the criminal you're chasing down if you disable their legs or cause them to drop their items if you disable their arms or hands.  Also reworked health dolls to show disabled limbs and be a bit more consistent and less buggy. Another inclusion is that when a limb takes max damage it becomes disabled as well, the way to fix it after taking max damage is to simply heal it to at least 50% health(if your hand which has a max damage of 30 was disabled, simply getting it to 15 health would heal it.)

**Why?**: I want to kill off the meta of aiming for legs and hands, it also makes sense for damage to limbs to be less lethal than directly hitting the head or chest. This also encourages security to use disablers which I think is much better than always using your instant stun weapon, this should make security slightly more fun hopefully by widening their arsenal of viable weapons.

**Changelog:**
:cl:
add: Stamina and limb damage has been reworked.
balance: damaging limbs is slightly less lethal
tweak: stamina is now applied to limbs
tweak: limbs can now be disabled by disablers and other forms of stamina damage
tweak: new inventory overlay to show that you can't use your hands when they're disabled or missing
/:cl:


Ralta:
REEEEEEEEEEEEEE CHANGES!

I kid. I like it. As long as blood loss and the like still exist, this is more balanced and makes sense. As an aside, there is good reason to attack the chest - damaging internal organs. Not as obvious a place to attack someone, as you'd really want to stop them from fighting back easily (so legs/arms make more sense for that purpose) but still.

taukausanake:
This actually sounds fine. Doesn't seem like there would be any balance problems.

You said you reworked the health doll? Can we get some screenshots of that?

TDSSS:
I like it a lot.

KasparoVy:
Lovely!

Couls:
@taukausanake it was mainly just the code that was reworked you shouldn’t notice anything outside of the healthdoll being a bit less buggy but there are new sprites for disabled limbs on the health roll which I will post when I have the time!

KasparoVy:
I foresee a bunch more entertaining 'tis but a scratch!' scenarios with limbless folks hahaha.

taukausanake:
@Couls cool, it was the new sprites for disabled limbs I was asking about. I'll just check the dmi in the meantime

Couls:
Yep! That’s them, credit for TG for the sprites I just cut them up for all of our limb types

code/__DEFINES/mobs.dm
```diff
@@ -7,6 +7,9 @@
 #define ORGAN_DEAD       8
 #define ORGAN_MUTATED    16
 
+// /obj/item/bodypart on_mob_life() retval flag
+#define BODYPART_LIFE_UPDATE_HEALTH (1<<0)
```
farie82: just make it 1. 1<<0 = 1 1<<1 = 2.
Since we don't use bitshift logic to define other bit related things.

code/_onclick/hud/screen_objects.dm
```diff
@@ -388,9 +389,13 @@
 
 	if(hud && hud.mymob)
 		if(iscarbon(hud.mymob))
-			var/mob/living/carbon/C = hud.mymob
+			var/mob/living/carbon/human/C = hud.mymob
```
farie82: casting can go wrong. Due to you checking if it's carbon. Then casting it to human

code/modules/mob/living/carbon/human/species/_species.dm
```diff
@@ -571,15 +574,14 @@
...
+		H.healthdoll.cut_overlays()
 		if(H.stat == DEAD || (H.status_flags & FAKEDEATH))
 			H.healthdoll.icon_state = "healthdoll_DEAD"
-			if(H.healthdoll.overlays.len)
-				H.healthdoll.overlays.Cut()
 		else
-			var/list/new_overlays = list()
-			var/list/cached_overlays = H.healthdoll.cached_healthdoll_overlays
+/* 			var/list/new_overlays = list()
+			var/list/cached_overlays = H.healthdoll.cached_healthdoll_overlays */
```
farie82: can be removed if it's not required anymore

code/modules/mob/living/carbon/human/species/_species.dm
```diff
@@ -571,15 +574,14 @@
...
 			// Use the dead health doll as the base, since we have proper "healthy" overlays now
```
farie82: comment doesn't make sense anymore

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -41,6 +42,7 @@
 	limb_name = "groin"
 	icon_name = "groin"
 	max_damage = 100
+	max_stamina_damage = 100
 	min_broken_damage = 35
 	w_class = WEIGHT_CLASS_BULKY // if you know what I mean ;)
```
farie82: wowie

code/_onclick/hud/screen_objects.dm
Couls: Good catch, not sure how I missed this!

code/modules/surgery/organs/subtypes/standard.dm
Couls: ;)

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -381,7 +387,10 @@ This function restores all organs.
 			if(dna.species)
 				damage = damage * dna.species.burn_mod
 
-			if(organ.receive_damage(0, damage, sharp, used_weapon))
+			if(organ.receive_damage(0, damage, 0, sharp, used_weapon))
+				UpdateDamageIcon()
+		if(STAMINA)
```
ninjanomnom: At least be consistent with the amount of spaces between cases

code/modules/mob/living/carbon/human/species/_species.dm
```diff
@@ -461,6 +461,9 @@
 		var/obj/item/organ/external/temp = M.bodyparts_by_name["r_hand"]
 		if(M.hand)
 			temp = M.bodyparts_by_name["l_hand"]
+		if(temp && temp.disabled)
```
ninjanomnom: A neat bit of 512 that simplifies these kinds of checks. It also really minorly improves performance but not enough to matter

```suggestion
		if(temp?.disabled)
```

code/modules/mob/living/carbon/human/species/_species.dm
```diff
@@ -538,7 +541,7 @@
 	return H.health - H.getStaminaLoss()
 
 /datum/species/proc/handle_hud_icons(mob/living/carbon/human/H)
-	if(!H.client)
+	if(!H || !H.client)
```
ninjanomnom:
```suggestion
	if(!H?.client)
```

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -14,8 +14,10 @@
 	return 0
 
 
-/mob/living/carbon/human/proc/has_organ(name)
+/mob/living/carbon/human/proc/has_organ(name, check_disabled = FALSE)
 	var/obj/item/organ/external/O = bodyparts_by_name[name]
+	if(check_disabled && O && O.disabled)
```
ninjanomnom:
```suggestion
	if(check_disabled && O?.disabled)
```

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -55,12 +57,25 @@
...
 	body_part = ARM_LEFT
 	parent_organ = "chest"
 	amputation_point = "left shoulder"
 	can_grasp = 1
 	convertable_children = list(/obj/item/organ/external/hand)
+	stam_heal_tick = 2
+	max_stamina_damage = 50
+
+/obj/item/organ/external/arm/set_disabled(new_disabled = TRUE)
+	..()
```
ninjanomnom: You should default to this unless you have a good reason so that you can save some work later when you need a return value. The cost is minimal and you get a nice benefit in terms of readability as well if it's standard
```suggestion
	. = ..()
```

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -76,13 +91,22 @@
...
 	icon_position = LEFT
 	parent_organ = "groin"
 	amputation_point = "left hip"
 	can_stand = 1
 	convertable_children = list(/obj/item/organ/external/foot)
+	stam_heal_tick = 2
+	max_stamina_damage = 50
+
+/obj/item/organ/external/leg/set_disabled(new_disabled = TRUE)
+	..()
```
ninjanomnom:
```suggestion
	. = ..()
```

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -99,12 +123,21 @@
...
 	body_part = FOOT_LEFT
 	icon_position = LEFT
 	parent_organ = "l_leg"
 	amputation_point = "left ankle"
 	can_stand = 1
+	stam_heal_tick = 2
+	max_stamina_damage = 30
+
+/obj/item/organ/external/foot/set_disabled(new_disabled = TRUE)
+	..()
```
ninjanomnom:
```suggestion
	. = ..()
```

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -204,8 +257,8 @@
 	name = limb_name
 	..()
 
-/obj/item/organ/external/head/receive_damage(brute, burn, sharp, used_weapon = null, list/forbidden_limbs = list(), ignore_resists = FALSE)
-	..(brute, burn, sharp, used_weapon, forbidden_limbs, ignore_resists)
+/obj/item/organ/external/head/receive_damage(brute, burn, stamina, sharp, used_weapon = null, list/forbidden_limbs = list(), ignore_resists = FALSE)
+	..(brute, burn, stamina, sharp, used_weapon, forbidden_limbs, ignore_resists)
```
ninjanomnom: why

Couls: THAT'S A VERY GOOD QUESTION

code/modules/surgery/organs/organ_external.dm
```diff
@@ -234,34 +258,59 @@
...
 		perma_injury = 0
 
 	if(updating_health)
 		owner.updatehealth("limb heal damage")
-
+	consider_processing()
+	check_disabled()
 	return update_icon()
 
+/obj/item/organ/external/proc/total_damage(include_stamina = FALSE)
```
Couls: renamed proc to total_damage due to having our own get_damage proc that uses a perma_injury var. let me know if you would rather I remove the old get_damage proc

PR #11964, opened 2019-08-05
Add the NO_MACHINES tag to ashwalkers

datlo:
This PR adds a new tag, NO_MACHINES, to ashwalkers. This tag prevents them from using most machinery/computers (basically everything you can find on lavaland aside from airlocks and airlock controllers), and entering mechs/space pods.

This PR solves : 

1. ashwalkers invading/visiting the station. they need assistance, either from crewmembers or admins, to do so now.
2. ashwalkers blowing up the syndi base easy peasy by deconstructing the outside wall and triggering the base's self destruct. the base's self destruct really should also check for a syndicate ID before triggering but thats outside the scope of this PR.
3. ashwalkers stealing miner mechs who were too dumb dumb to DNA lock it, cavelizards shouldnt be able to pilot mechs

this PR also adds a new mob proc, can_use_machines(), which checks for the tag and if youre an advanced tool user.

:cl:
tweak: Ashwalkers can no longer operate most machinery, computers, mechs, space pods, and shuttles.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -2020,10 +2020,7 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 		visible_message("<span class='danger'>[src] is engulfed in shadows and fades into the darkness.</span>", "<span class='danger'>A sense of dread washes over you as you suddenly dim dark.</span>")
 
 /mob/living/carbon/human/can_use_machinery()
-	if(!IsAdvancedToolUser())
-		to_chat(src, "<span class='warning'>You don't have the dexterity to do this!</span>")
-		return FALSE
 	if(NO_MACHINES in dna.species.species_traits)
 		to_chat(src, "<span class='warning'>You can't figure out how to do this!</span>")
 		return FALSE
-	return TRUE
+	. = ..()
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1823,6 +1823,10 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 			to_chat(src, "<span class='warning'>Your fingers don't fit in the trigger guard!</span>")
 			return 0
 
+	if(TRIBAL in dna.species.species_traits)
+		to_chat(src, "<span class='warning'>You can't figure out how to use this weapon!</span>")
+		return 0
+
 	if(martial_art && martial_art.no_guns) //great dishonor to famiry
 		to_chat(src, "<span class='warning'>[martial_art.no_guns_message]</span>")
 		return 0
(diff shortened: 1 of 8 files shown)


Markolie:
You should still keep the NOGUNS trait separate from TRIBAL and give them both. That adds more granularity and use for the TRIBAL trait in the future.

I think it's also better to add a can_use_machinery() proc on a basic mob level which is overridden on the human level which makes this check. It makes it much more future proof and easier to work with.

datlo:
Ill look into those when i can

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -2016,4 +2017,13 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 	var/obj/item/organ/internal/lantern/O = get_int_organ(/obj/item/organ/internal/lantern)
 	if(O && O.glowing)
 		O.toggle_biolum(TRUE)
-		visible_message("<span class='danger'>[src] is engulfed in shadows and fades into the darkness.</span>", "<span class='danger'>A sense of dread washes over you as you suddenly dim dark.</span>")
\ No newline at end of file
+		visible_message("<span class='danger'>[src] is engulfed in shadows and fades into the darkness.</span>", "<span class='danger'>A sense of dread washes over you as you suddenly dim dark.</span>")
+
+/mob/living/carbon/human/can_use_machinery()
+	if(!IsAdvancedToolUser())
```
Markolie: This check should be added to the mob-level proc definition instead, as IsAdvancedToolUser() is actually defined on the mob level as well.

Looks good otherwise!

MrMagolor:
Will this prevent them using drying racks?

datlo:
good question i dont believe so but i will doublecheck

datlo:
they can use drying racks

Fox-McCloud:
I really question the necessity of this. Ashwalkers have some pretty sizeable downsides to them and are easily controlled with ranged weaponry.

If a miner leaves a mech/pod unattended and doesn't DNA lock it, that's kinda on them.

The syndicate base is kinda a separate issue, and it would be better to address that directly, than indirectly.

MrMagolor:
>The syndicate base is kinda a separate issue, and it would be better to address that directly, than indirectly.

There is a note in the syndicate base that mentions the walls are filled with explosives- could make it so that attempts to deconstruct the walls trigger the destruct?

Necaladun:
that seems it'd make it even easier to destruct the whole base?

MrMagolor:
>that seems it'd make it even easier to destruct the whole base?

I mean... aren't you supposed to detonate the self-destruct when you're compromised, anyways?

datlo:
Syndis are equipped with weapons more than strong enough to handle a single ambitious ashwalker.

SpessX:
The obvious next step should be giving them the option to tame goliaths and ride them like Goliamechas.

variableundefined:
Closing for reasons Fox stated above.

PR #11965, opened 2019-08-05
Give cult talisman different icons

TDSSS:
**What does this PR do:**
This PR makes it so cult talismans get different icons so you can tell them apart without examining them. No more reaching for the wrong talisman in the middle of combat.

This also means non-cult crew can tell them apart too but what big difference does it really make if the cultist you just caught had a stunning or a horror talisman? You're going to burn it anyway.

All credit for the sprites goes to "Ionward (Geminerd)" on Discord, except the supply icon which is the old one we had before.
Capture

**Changelog:**
:cl:
tweak: cult talisman got unique icons to tell them apart
/:cl:


code/modules/paperwork/paper.dm
```diff
@@ -54,7 +54,7 @@
 
 /obj/item/paper/update_icon()
 	..()
-	if(icon_state == "paper_talisman")
+	if(istype(src, /obj/item/paper/talisman))//let's not override talisman sprites, tyvm
```
Markolie: This should be overriden in the talisman’s update_icon instead.

TDSSS: You're right, I should just improve the code when I see such stuff instead of going with it, lmao. Will fix tomorrow.

MRgaty:
The problem I see with this is that people will recognize each one and it's functionality when in theory people should not be able to read them. For the rest the icons are neat and helpful.

TDSSS:
Sure, but...what advantage does that actually give them, to recognise the talismans? The only circumstances you see talismans as objects is when searching a cultist or raiding a cult base. It's really not a big gameplay change for them.

code/modules/paperwork/paper.dm
TDSSS: Should be fixed now

MRgaty:
> Sure, but...what advantage does that actually give them, to recognise the talismans? The only circumstances you see talismans as objects is when searching a cultist or raiding a cult base. It's really not a big gameplay change for them.

Is more like an IC thing rather than a mechanical thing, people will meta the different uses of the runes, that's all, is not gonna affect gameplay for sure.

Mitchs98:
Much better than seeing same old paprr. Unsure how anyone would really meta anything at all from this. Or what any real downsides are.

MrMagolor:
>Much better than seeing same old paper. Unsure how anyone would really meta anything at all from this. Or what any real downsides are.

I feel like it's more immersion than anything: Joe McRedshirt somehow possessing the occult knowledge to identify a cultist talisman at a glance: not to mention, the non-cultist description for talismans mentions that it just looks like gibberish.

TDSSS:
Considering Joe McRedshirt somehow knows everything about Terror Spiders, Shadowlings, Xenomorph lifecycles, syndicate top secret technology, etc. I'm okay with this.

Fox-McCloud:
I really think these are *cool*.

PR #11970, opened 2019-08-08
Fix conflicting traitor objectives

datlo:
This PR changes objective code to check that you are not targeting someone who's already a target for that traitor, avoiding situations where you have to protect and assassinate the same person.

~~There's still a bug where somebody with a steal objective (or other martyr incompatible objs) can apparently get the Die a glorious death obj when they are not meant to, the check is apparently buggy. I cant figure out why and its hard to test it for stuff like this on a private server, I've never seen it on live.~~

There might also be a bug where somebody can get both sets of documents at once, although looking at the code I am not sure how that happens and I havent seen it happen myself.

Once this PR is merged the config should be set back to two objectives per traitor

:cl:
fix: Fix some cases of traitors getting conflicting objectives, such as assassinating and protecting the same target.
/:cl:

### DIFF
--- a/code/__DEFINES/gamemode.dm
+++ b/code/__DEFINES/gamemode.dm
@@ -6,6 +6,7 @@
 #define TARGET_INVALID_UNREACHABLE	5
 #define TARGET_INVALID_GOLEM		6
 #define TARGET_INVALID_EVENT		7
+#define TARGET_INVALID_IS_TARGET	8
 
 //gamemode istype helpers
 #define GAMEMODE_IS_BLOB		(SSticker && istype(SSticker.mode, /datum/game_mode/blob))
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -42,6 +42,7 @@
 	var/list/kills = list()
 	var/list/datum/objective/objectives = list()
 	var/list/datum/objective/special_verbs = list()
+	var/list/targets = list()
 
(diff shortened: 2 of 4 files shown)


FreeStylaLT:
datlo literally taking the hilarious fun away smh

datlo:
I wouldnt have made this PR if admins didnt change the config to only have 1 obj per traitor due to this bug so go yell at them smh

Couls:
should also ensure someone who has to exchange folders can't get the die a glorious death objective

datlo:
As I said on description this is a bug with the martyr check that I cant figure out or test for

datlo:
I have tested setting the martyr chance to 100 percent prob. The check worked properly and gave me the escape objective instead of martyr : https://cdn.discordapp.com/attachments/145700319819464704/608952165880299531/unknown.png
Unless I get more details on this, a screenshot of it happening on live, or detailed steps to reproduce, I am considering the check to be working

MRgaty:
Thing is, if you give 2 objectives to 17 traitors, as we had in some highpop rounds, the round will be extremely chaotic. I would suggest also reducing the number of traitors per crew too.

datlo:
That's for admins to decide and isnt directly relevant to the PR, it's just a suggestion based on what I've been told (objectives were reduced because of this bug)

PR #11633, opened 2019-05-31
Lavaland update!

CornMyCob:
**What does this PR do:**
I have added numerous things with this PR, including:
- Kinetic crusher code refactor, it has its own file now
- Ported kinetic crusher loot from /tg/
- Ported the animated legion sprite from /tg/
- Added the survival medipen to the mining rewards vendor for 400 points
- Sorted the items in the mining rewards vendor to be in order of price
- Added the book of babel as a possible drop from the necropolis chests, which grants all languages
- Added the remaining 2 biodome maps (clown planet and snow biodome) and all their content
- Added the H.E.C.K suit, a drop from bubblegum
- Added the fire first aid kit to the mining vendors
- Fixed ripleys being slow on lavaland and as a result being shit
- Balancing for bubblegum and colossus

Regular fauna each have a different drop chance for their loot item when killed with the crusher, and you must deal over half a megafauna's health with the crusher to get their drop. The rewards are quite fancy, and hopefully this will mean that the crusher will get the love it deserves. You can view them all and their effects [here](https://tgstation13.org/wiki/Shaft_Miner#Crusher_trophies)

The book of babel is one of the few items which hasn't been ported from /tg/ yet for the necropolis chests, and was easy enough to get in so I did that.

For the balance of the bubblegum and colossus, I reduced the speed of the death bolts so that they are about as easy to avoid as they are on /tg/, and reduced their damage by 5. For bubblegum, I reduced its movement speed a bit. I'm open for discussion on megafauna balance, these are the two ones with the biggest balance issues in my opinion, the legion may also need a movement speed nerf.

Credit goes to numerous coders over on /tg/ for most of the work here.

**Changelog:**
:cl: CornMyCob
add: Trophies for killing fauna with the kinetic crusher which you can attach to your crusher
add: 2 of the biodome ruins for lavaland
add: H.E.C.K suit drop from bubblegum
add: You can now purchase the fire first-aid kit from the mining rewards vendor
fix: Ripleys are no longer slow on lavaland
tweak: The megafauna legion's sprite is now animated
balance: Colossus and bubblegum are now more suited for paradise movement speed
/:cl:

PR #11981, opened 2019-08-11
Major Lavaland Update

Fox-McCloud:
Slowly bringing Lavaland up to parity with TG.

Makes a number of changes to Lavaland to make it closer to parity with TG; has a lot of updates to starting equipment, how existing equipment works, and what you can purchase, as an upgrade.

Things I haven't done and won't be doing in this PR, so please don't ask:
- Sandbags (requires an overhaul of barricades)
- Fultons (Will need more testing and work that I'd prefer be more fixated on)
- Kinetic Crusher Trophies (quite involved)

Some highlights:
- All miners start off with kinetic accelerators
- Kinetic Accelerator and Advanced Mining scanner are no longer an option for a voucher kit
- Conscription kit includes survival gas mask, survival suit, survival knife, and kinetic accelerator
- Miners start out with a survival gas mask in their box
- Mining lockers now appropriately have mini pickaxes and a plant bag (for lavaland snacks)
- Removed marker beacon from lockers
- Re-ordered equipment vendor
- Added in a boxed implant kit and minebot KA passthrough kit (more on that later)

Updated the minebot
- Minebot no longer has a hacky KA gun; it has an actual kinetic accelerator. 
  - This means you can actually install KA upgrades and mods to the gun it has. Note that the cooldown mod from regular KA's cannot be applied to minebots; they require their own special minebot KA cooldown mod 
- There's a special upgrade for non-minebot KA's that allow your accelerator shots to pass through a minebot without hurting them; useful for avoid getting your shots blocked
- Sentient minebots nerfed; they can no longer accept minebot upgrades (like the armor and melee upgrade). They can still accept upgrades to their kinetic accelerator, though (overall, IMO, this is still a buff)

Updated Kinetic accelerators
- This is mostly just backend changes; instead of the projectile being directly modified by the gun itself; the projectile references the KA's internal mods---this means we can have more unique mods that have unique behaviors on hit
- buffed the cooldown mod a bit
- some new tendril only mods that can act as lifesteal or dramatically lower the cooldown of your gun, so long as you don't miss. There's also a KA blast that leaves behind a resonator field
- the indoor pressure mod has been removed from R&D; it's now a miner specific traitor item

Updates Resonators
- both resonators have +1 field limit, compared to before
- Resonators do 60 damage at full field strength, up from 50
- You can now burst resonator fields early by immediately placing another field on top of it; there is a damage penalty to for doing so on the regular resonator, but not the upgraded one
- Resonator melee damage increased from 8 to 15

Updated Plasmacutters
- No longer does bonus damage in low pressure
- increased the base damage of advance plasma cutter, but lowered how many tiles it can mine
- Cut the cost of both plasma cutters

Straight Up Removals
- Removed the highest tier kinetic accelerator from pods. Snowflaking yet another KA type for these guys is something I'm not going to do. Pods are a pain, buggy, and ~~total trash~~. Either case, this just means that you can't have a pod that can do AoE Mob+turf mining at the same time.....not a big loss, IMO, as literally no one cares about mining with a pod.

Fixes
- Ensuring consistent pressure checks that will work in lavaland

:cl: Fox McCloud
add: Majorly updates lavaland
add: Miner starting equipment altered. They now start with a kinetic accelerator
add: Miners start off with a mini pickaxe
add: Adds in the boxed implant kit, regular shelter capsule, and mining drone passthrough upgrade to the mining equipment vendor
tweak: re-ordered the equipment vendor, so things are grouped together more logically
add: Added a number of new kinetic accelerator upgrades acquirable through tendril chests
tweak: Mining bot now uses an actual kinetic accelerator as opposed to a "Look alike"; you can upgrade its internal gun with various kinetic accelerator modkits.
tweak: Sentient mining bot nerfed; you can no longer apply the armor or melee buff to them; offset by the fact you can upgrade their kinetic gun.
tweak: Indoor pressure mod removed from R&D and made into a miner traitor item
tweak: Resonators have an increased amount of fields and do more damage in low pressure. Additionally, if you place a field on a tile that already has a field, it'll cause it to detonate early
tweak: Plasmacutter no longer does increased damage in low pressure, Advanced plasma cutter does slightly more base damage but has a bit less range. Price has been cut at R&D
del: removed the AoE Mob+Turf upgrade for mining pods
fix: Fixes ripleys being slow on lavaland
/:cl:


Quantum-M:
I disagree with the miners having a headset with science radio.
My reasoning is that:

1. Miners have been able to manage in not having to have direct radio contact with science, both in the old asteroid version and the current lavaland version up to this point.

2. All other jobs, with exception of command, the NT Rep., and genetics do not have access to a radio that isn't directly part of their own department. While mining brings ores for science, they are not part of the science department.

3. It could end up miners/scientists go into a constant argument when one gets peeved at the other, especially as taking a radio from miners would be more difficult due to the distance it can take to reach lavaland/whatever point they are in lavaland.

lolzyking:
Don't worry about the crusher trophies, theres already a PR for it fox, 

I'm also the only person that mines with space pods, but yes we don't need another mining gun, the ones we have on space pods are good enough as is.

Kyep:
I'm really not sure that giving miners access to science radio is a good idea.

Denthamos:
Hmmn, if people are that sketchy about all miners having both supply and science comms, what if just the QM had the additional science comms on their headset?  So then the RD/science and the QM coordinate orders and miners requests back and forth while preventing miners spamming up the science comms.

Quantum-M:
@Denthamos they could still use their pda or the request consoles, or show up  at cargo since there's usually one or two scientists with nothing to do in a round (or the Director themselves who is supposed to direct science, not actually do it all).

If science isn't getting minerals, that would be largely due to:
A. Miners are newbies and aren't fully sure what they are doing.
B. They dead or are non-existent (which would be the QMs and HoP job to replace/add to).
C. The miners are antags and are busy doing antag stuff.
D. Miners are peeved at science and does not want to give them anything.

Leanfrog:
So, what exactly is contained within the boxed implant kit mentioned at the bottom the highlight list?

Fox-McCloud:
The same exact thing as a security implant kit, only 1 less implant case.

Fox-McCloud:
Alright, PR has been updated and tested. Problems with it have been resolved.

I really don't think the science portion of the headset will be  hugely problematic, personally---I think it's definitely worth trying out, especially since it's a very very minor technical change.

If it becomes a problem, it'll be incredibly easy to revert; it's not heavily integrated into the map or code.

Couls:
I personally think science comms is a good idea, they're kind of underused in their current state since there's not really much to talk about among scientists, being able to call miners when ORM is upgraded and when their items are researched will be a very useful tool and vice versa when miners bring in ore as its easy to miss the materials arrived message you get from those announcement things no one uses

MrMagolor:
Will this have any conflicts with #11633?

Kyep:
I'm okay with this once the bit about mining getting science headsets is removed.

Until then, that bit makes me sufficiently uncomfortable I feel I have no choice but to object.

Spacemanspark:
Would it be too out of scope to add the ability to place a pickaxe on the back slot in this PR @fox-mccloud

Fox-McCloud:
removed the science channel *for now*.

SpessX:
Maybe survival shelters could come with a holopad, then they'd have another thing to do while waiting out the storm.

PR #11971, opened 2019-08-08
Port and modify delta escape shuttle to cyberiad

datlo:
Port and updates the delta station escape shuttle to the cyberiad. It has much more open spaces, more chairs, and is just a better fit for a server with high pop.

The syndicate base was also slightly enlarged to fit the shuttle dock when hijacked, and some of the transit areas were shuffled to make space for the bigger transit necessary for this shuttle. Also I deleted the transit area for the science shuttle. *shrugs*

Screenshots with details where needed : 

Overview of the shuttle : 
shuttle7

Remap of the escape pod bay so that it doesnt get blocked by the escape shuttle : 
shuttle8

Cockpit, one chair for each command/vip minus the Blueshield : 
shuttle2

Security cell and cockpit foyer. Both command and security have access to the foyer. Comes with a chair for the Blueshield/security officer on guard to strap in.
shuttle3

The one room that is entirely new : The emergency external airlock. The external airlocks are public access and can be entered by any crewmember, however the internal white doors are security/command only and are see-through glass, meaning it will be difficult for antags to sneak in without being noticed that way. This also provides an alternative entrance to hijack antags who dont mind going loud to emag the shuttle away. The external airlocks (and only those) have atmos blocking tiny fans so the pressure doesnt get fucked whenever someone enters or exits the shuttle using them.
shuttle4

Cargo and medical. Oxygen crates contain 4 yellow emergency tank and breath mask, medical crate contains a few various medkits.
shuttle5

Pleb room
shuttle6

:cl:
add: Replaced the current escape shuttle with a larger, improved one modified from Delta station.
/:cl:

PR #11696, opened 2019-06-07
Add shuttle chairs (and fix newlines on all maps)

Markolie:
**What does this PR do:**
This pull request ports over the shuttle chairs from /tg/ and replaces all chairs on our shuttles with them. While I was at it, I also ran the TGM converter on all maps to fix their newlines.

Image: 

**Changelog:**
:cl: Markolie
add: All chairs on shuttles have been replaced with brand new shuttle chairs.
/:cl:

PR #11982, opened 2019-08-11
Port and modify delta escape shuttle to cyberiad

datlo:
Deconflicted version of #11971, with new shuttle chairs added to the shuttle from #11696

See linked PR for details on the shuttle and comments

newshuttle

:cl:
add: Replaced the current escape shuttle with a larger, improved one modified from Delta station.
tweak: Modified the escape pod bay layout so that the dock isnt blocked when the new larger shuttle is docked
/:cl:


craftxbox:
What about the mech recharger?

datlo:
I dont really see the need for a mech recharger on the escape shuttle imo

Also ree this is conflicted again

taukausanake:
I see the mech recharger as a docking port for mechs. A place meant for them to be placed. You can fit one right next to the windows in the cargo section.

Can I ask what those external airlocks are doing on the top-ish right? Where are they meant to go to and I can see them as a potential entrance into the shuttle besides the main doors. They also aren't shuttle doors so really wondering what they are doing there

datlo:
from the linked PR : 
'The one room that is entirely new : The emergency external airlock. The external airlocks are public access and can be entered by any crewmember, however the internal white doors are security/command only and are see-through glass, meaning it will be difficult for antags to sneak in without being noticed that way. This also provides an alternative entrance to hijack antags who dont mind going loud to emag the shuttle away. The external airlocks (and only those) have atmos blocking tiny fans so the pressure doesnt get fucked whenever someone enters or exits the shuttle using them'

Kyep:
What happens if an admin loads one of the alternate escape shuttle templates?

datlo:
Shuttle is loaded as you'd expect however you will not be able to switch back to this new escape shuttle because the code for the shuttle loader cant handle its size

taukausanake:
Maybe that should be fixed first? Loading of different sized ships?

datlo:
I was told its fine

However with the new transit system that I believe was merged it should be much easier to fix I will look into it when I get around to updating this.

datlo:
This PR needs more work that I dont really want to deal with right now. It WILL be reopened eventually, ideally after Free's engineering remap is merged.

If anyone wants to pick up the files and make their own deconflicted PR they are free to do so, ideally also with a fix for shuttle loading (it cant be loaded right now because its too big for the shuttle manipulator).

PR #11515, opened 2019-05-10
Sanitize all inputs

Tayyyyyyy:
**What does this PR do:**
Sanitizes all text input. Admins can no longer send arbitrary html via global narrate/pms/direct message, but they can use pencode.

**Changelog:**
:cl: Tayyyyyyy
tweak: Inputs sanitized
/:cl:

PR #11985, opened 2019-08-12
Add admin pencode: style, class, and img

Tayyyyyyy:
# What does this PR do
#11515 reduced the flexibility of admin message formatting. This introduces 3 tags that are part of admin pencode, [class], [style], and [img]. This is enabled for subtle, direct, global, admin pms, make sound, and admin faxes.

# Examples
## class
Let's say you want to get someone's attention. You can use any span that you could use before html was removed.
### input
image
### output
image

## style
Maybe you like the classic big red text.
### input
image
### output
image

### input
image
### output
image

You can also specify a width x height using semicolons:
### input
image

### output
image

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -546,7 +546,8 @@ proc/checkhtml(var/t)
 	return text
 
 /proc/convert_pencode_arg(text, tag, arg)
-	arg = replacetext(replacetext(html_encode(arg), "'", ""), "\"", "")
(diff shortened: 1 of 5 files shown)


Terilia:
Awesome thankies!

Spacemanspark:
The ability to use pencode doesn't function with the 'Make-Sound' verb, which is very useful for running events.

Tayyyyyyy:
Oops! I missed that one. Fixed.

PR #11990, opened 2019-08-12
Slime people rework

Couls:
**What does this PR do:**
Moves slime people over to new crit, gives them lungs and hearts(sprites for those will be coming soon)

Changes:

- Slime people now have hearts and lungs which means they breathe

- Slime people no longer have bones

- Slime people no longer take brute damage from crit or toxin damage from blood loss

- Slime people now use new crit

heart:
image
lungs:
image

Credit to drakeven for the sprites!
**Changelog:**
:cl:
add: Slime lungs and heart
del: Slimes no longer take toxins damage from blood loss
del: Slimes no longer take brute damage in crit
add: Slimes now breathe
tweak: slimes now use new crit
/:cl:


code/__DEFINES/genetics.dm
```diff
@@ -146,4 +146,5 @@
 #define RESISTHOT		16
 #define RESISTCOLD		17
 #define NO_EXAMINE		18
-#define CAN_WINGDINGS	19
\ No newline at end of file
+#define CAN_WINGDINGS	19
+#define NO_BONES        20
```
Fox-McCloud: We're trying to get away from having flags determine abilities, where possible; it'd be better if we did what we did with Diona, where they have a limb subtype, and that subtype has the `unbreakable` var set to `TRUE`.

code/modules/mob/living/carbon/human/species/slime.dm
```diff
@@ -176,6 +176,16 @@
 	else
 		to_chat(H, "<span class='warning'>You need to hold still in order to regrow a limb!</span>")
 
+
+/obj/item/organ/internal/heart/slime
+	name = "osmatic pressure regulator"
+	desc = "A wet squishy heart."
+
+/obj/item/organ/internal/lungs/slime
```
Fox-McCloud: These are defined twice.

MRgaty:
I like this a bit less, because as I stated in the discord discussion, not having bones will make doctors frown a bit for sure but is not as characteristic as not breathing and is not as easily identifiable.
So slimes now are a bit more similar to humans than before.

code/modules/mob/living/carbon/human/species/slime.dm
Couls: fixed, I forgot to remove them after I moved them to their own file

code/__DEFINES/genetics.dm
Couls: Would the var reset if the limb was removed? I can see people amputating their boneless limbs and giving them to others

Imotepchief:
Slime people can already be revived by defib, which if you think about it, is an odd function for a race without a heart...
The no breathing is also a little strange, as it prevents them from taking gaseous chems and the like, which considering they are made of slime and I would imagine possess an entire body of permeable membrane, is really odd.

I do like the idea of slime people no longer having bones, but I would like to see their unique nature be preserved somewhat.
Maybe instead of having a conventional "heart" and "lungs", they could possess additional "nodes" in their head, smaller than the slime core, which allow them to regulate their enlarged form.
Make these "nodes" a required function, that operate similar to heart and lungs, but don't outright kill the the slime person if they are damage and/or removed, but instead reduce their functionality somewhat.
For example:
Motor neuron core, which allows them to control their finer motor skills.
It's damage would cause them to drop things more often, or fall over.
It's destruction would reduces them to virtually no movement and the inability to pick things up.

Maybe the other node could be their thermal regulator, without which they couldn't maintain their thermal equilibrium at all, which could lead them to easily melt and/or freeze.

I would also really like to see slimes no longer suffer from IB, not only is the slipping because of "coughing up blood" frustrating, but it doesn't make sense.

Oh oh!

Also, what I believe to be a fantastic idea, is for slime heads to be able to cut it and run from their body if their body specifically, takes too much damage.
Imagine, slime people heads rolling down the corridor to escape a terror spider who is happily munching on their body!

Couls:
Great ideas and suggestions, sadly these are all out of the scope of the PR and probably incompatible with new crit as well.

code/modules/surgery/organs/subtypes/slime.dm
```diff
@@ -0,0 +1,7 @@
+/obj/item/organ/internal/heart/slime
+	name = "osmatic pressure regulator"
+	desc = "A wet squishy heart."
```
Fox-McCloud: Should probably have a better description, similar the membrane.

Couls: Not sure what i'd give it, I went off reference of the other races' descriptions and they didn't have much detail as well(looked at skrell and vox)

Fox-McCloud: how about "it appears to be some kind of bilogical pump that uses osmotic pressure to regulate water flow."

also, yeah `osmotic`, not `osmatic`

Fox-McCloud:
just add the sprites when you get them.

Couls:
added sprites but they might be a bit big, may need to resize it down a bit

SteelSlayer:
I really don't like this change. I mean having no bones sounds cool and all but not with all the other changes. I feel like slime people are now simply an upgrade to human if you think about it. 

Pros over humans: 
-Never have to worry about bones breaking
-Blood for water, easily replenishable
Cons:
-Can't heal toxins due to lack of liver

The one con they do have is negligible in my opinion. Slime people will function almost exactly the same, except they have a few major benefits. Also, I think that having unbreakable bones on a new crit race might actually be OP.

Couls:
Toxins damage for slime people is going to be rarer now that they take oxyloss from blood loss, taking that in mind a liver would not save you in a lot of situations where toxins would be bad, having no bones means their internal organs can’t really be damaged as easily and with them finally having new crit my only concern is that they’ll take too much brain damage in the heart attack stage of crit but that’s something we’ll have to see in action especially when it comes to treatment in medbay

pinatacolada:
Hate this, slimes have always been known to have no organs except for the core, which controls all their body functions. Now we're adding hearts and lungs for what? Neater code? This strips a lot of their uniqueness.

No bones is neat tho

Fox-McCloud:
>Can't heal toxins due to lack of liver

Uh, no one heals toxins by default. Not having a liver is mostly just asking to get wrecked by alcohol.

SteelSlayer:
What I meant was Slimes don't vomit like some of the other races when they have a fair amount of toxins, and vomiting removes a bit a toxins. I admit I'm not sure know why slimes do not vomit, I thought it was because they lacked a liver, perhaps that's not the reason.

BakKeener:
Hearing about and now looking at this PR, I'm finding a trend that is happening as of late with multiple races in a effort to move them to newcrit, which was a bad idea in itself and is still disliked to this day by a large number of players but moving from that and to what I do think about this PR.

No bones I can very much get behind because **SLIMES DON'T HAVE BONES**. Having a slime come into medbay for bone repair surgery is mind boggling because they shouldn't have bones as they are, as they are called _Slime_. 

I do find it very annoying to have to treat a slime who is suddenly gaining toxins just because they are in crit as it is one of the slowest things to heal which I'm fine with having buffed and I didn't notice this before that they took brute damage while in crit while others didn't.

I don't think they should have a heart as the core already functions as such and just makes the core function as only the brain, which it already is pretty much but that takes some of it functionality away and what makes a slime unique. I have also never heard of a slime person outside of SS13 having said organs either. 

Lungs aren't exactly needed and to respond to Imotepchief about the breathing part and exchange of gasses I would leave that to the membrane coating they have. IB is definitely a thing to have removed as it doesn't make sense for slimes since they have something closer to an if not open circulation system in comparison to the other races closed ones.

Sikroth:
Not a fan of these changes. The no bones is a nice addition, but making slimes have heart and lungs feels a bit shoehorned just to make them newcrit friendly.

Here's a few things that come to mind:
- Would it be possible to conserve no-breathing while moving them to new-crit? As in, having them still take oxyloss in crit, but not really needing any particular atmos gases to live.
- Is it possible to have an organ do several functions? For example, make their core also work as the heart. This would honestly be quite a nerf because their "heart" will rapidly take damage during crit but it's still thematically more fitting. Not sure how feasible it is though.

Mitchs98:
100% absolutely against these changes. I thought the Dionae changes were bad and being done solely to move them over to newcrit, but this takes the cake and has no excuses otherwise. These changes are quite frankly appalling and EXACTLY what I knew Fox would push for.

This entirely guts the major unique thing slimes had, and for what...no bones? Sorry, but that's really not a good 'compromise' as stated on the discord. By far. You don't entirely mess up a race that was, in my opinion, one of the most unique by their traits and what they could do solely to push it onto a system hated by many players.

You're ruining races left and right just to push them onto a system many didn't even want or ask for in the first place.

This also doesn't address their HUGE brain damage modifier which WILL have to be removed if they're put onto newcrit. They will die much, much, MUCH faster than any other race when in crit due to it. Its not a matter of needing a playtest its a matter of legitimately needing it removed.

Frankly I'm against it regardless of it being removed. It makes no sense for a slime person to breathe or have a heart, or any other organs. Whatsoever. At the end of the day you can say what you want but in absolutely no lore for anything does a slime creature have a heart or lungs. They have a singular core that handles all body functions.

Just as they should here, just as they do in lore here. Yes lore is built around gameplay but thats a pitiful excuse to just gut a race like this. Especially for the reasons its being done. It's honestly baffling its even being supported by anyone, much less PR'd by anyone other than Fox or another staff member that supported newcrit.

Spacemanspark:
I'd rather slimes kept their nobreathe ability.

Terilia:
To me it looks like breathing is a limitation of newcrit. This is just pure guessing on my part, but in this proposal both Slime and Diona would receive it.

While I don't mind tree's breathing I like my slime nonbreathing. Everything else I am fine with.

DerpehBoi:
Slime are fine how they are. All this does is basically turns them LITERALLY INTO A HUMAN. This is just removing what few things they get that are unique. Very much against this and the diona one.

Spacemanspark:
I suppose I should expand my statement.
I see no real reason to change slime people right now. They aren't unbalanced and they're quite interesting to play for the most part.

Diona, on the other hand, could use some new changes... but what those changes are is a different topic.

I'm giving a -1 to this based on that. Species diversity is a pretty big thing and while we aren't perfect with how we do it, I don't think making them more humanlike is a direction we should be taking.

SomeSortaSquid:
I would be wholly in support of this if these were unicellular-organism-people, not slime-people. Unicellular organisms do have multiple organelles responsible for various functions.
Slimes do not.

The idea of a slime, correct me if I'm mistaken, came to be in 80s, with Gary Gygax's original Dungeon and Dragons set. They were a type of ooze creature characterized by being an entity made entirely out of what's essentially a gastric acid.
Later incarnations of the creature, mostly in sources which clearly took inspiration from DnD, added a singular core, an organ(elle?) around which the "body" of an ooze would gather.
As such, we have essentially two forms of slimes: A living gastric acid and a core surrounded by gastric acid. They are seemingly unicellular organisms, but in practice make no sense. You could consider them a form of prokaryotes, but it's rather far-fetched comparison.

My point is simple: Calling what this PR suggest a "slimeperson" is misleading, as it clearly takes inspiration from unicellular organisms, like Gygax' oozes did, but is not a slime.

I know I am basically arguing semantics here, but it's just a lengthy glorified "don't rework the entire species fantasy".

farie82:
I dislike new crit myself I'm not gonna lie. So I'm biased to anything making that more common. 

That said I don't like that lore and existing gameplay have to change to a new feature that is ill received among many people. Best is to make that feature adapt to the current situation.

How about keeping slimes how they are and instead giving them resp damage.
Yes this will mean a rework on the NOBREATH trait. And no it's not undoable. It'll also make it "doable" for the other NOBREATH races.

Short of it. Just make NOBREATH mean that you actually don't have to breath. But still can get resp damage. Make resp a kind of damage that signals that the body can't get the supplies to places it needs to get it to. In humans this is oxygen. In slimes this is something else.

EvadableMoxie:
Xarxis 5 is a mostly aquatic planet, with an atmosphere made up of Argon.  It would make absolutely no sense within the lore for them to breathe, let alone breathe oxygen, a gas that's barely even present on their homeplanet.

Unless we're going to entirely change all of the lore surrounding Slime people overnight. I kinda feel like lore is sort of important to a roleplay server, myself.

I suppose if this happens it'll just be even more collateral damage caused by newcrit, a system we are apparently willing to pay any price for.

MRgaty:
I have to agree that newcrit is not as well done as it should, the necessity of lungs and heart to work (and the fact that people just don't die when they have) proves it. Slimes right now kinda suck because every race can resist infinite amount of attacks if they are lucky when slimes have 200 health. 
I don't think the problem is the non breathing traits but how newcrit is handle.
Making all species more human-like when we are in a futuristic space station filled with alien creatures with their own lore (which newcrit doesn't seem to respect) and behavior is, as evadable stated, not a good way to go.
I dislike this change, I would rather play with a highly unbalanced species rather than snowflake newcrit on them, removing their most characteristic traits.

EvadableMoxie:
Trying to be a bit more constructive: If the goal is to move slime people to newcrit as I assume it is, I think there's better ways.

If slimes must breathe, they should be able to breathe in water, like Skrell. Coming from an aquatic planet and having blood that's literally water in the first place, this would make sense.

They should breathe nitrogen and not oxygen.  (Yes, there is nitrogen in water to breathe) The lore can be changed from Xarxis 5 having an atmosphere of argon to one of nitrogen and not impact lore too heavily, and it's a bit more friendly mechanically since argon doesn't exist in game but nitrogen does.  Vox already breathe Nitrogen so this isn't that snowflakey, and it means we can add more nitrogen tanks that Vox and Slimes can share.

Instead of giving slime people two new organs which are somehow fully compatible with alien organs, just have the slimecore handle breathing. It's supposed to do everything anyway. I understand this would probably be annoying to code, but it's much less jarring to the lore.  It would also help balance out the advantages slime people are getting, since their core takes extra damage and now it'll cause oxy damage when damaged.  They'd really want to protect their core, which makes sense.

Instead of giving slimes no bones and thus full immunity to bone breaking in combat, give them the option to regenerate their bones in addition to their limbs. This way they still need to worry about a broken hand or leg in combat, but they have the advantage of easily tending to it, as long as they have calories to burn and some down time.

Mitchs98:
...Was the slime brain damage multiplier ever removed? I really feel it needs to be removed for this to be viable. Bit late since its been merged...but it should be addressed.

KasparoVy:
The madness continues

Bigtimetripod:
250% brain damage + brain damage from new crit?

Slimes gonna be dead of brain damage before heart attack does its thing if my math is right... Even if slimes are revived and were in crit for a long time, can you revive someone whos brain is destroyed? My answer is no.

Either way, new meta for slimes in crit is gonna be emergency injector + 20 units of mannitol to even survive near crit states. Even if crit knocks on the door, simple emergency injector is not gonna do for slimes since when cardiac arrest sets in, slimes gonna be too dumb already that they smash their head against the airlock trying to open it.

JKnutson101:
> can you revive someone whos brain is destroyed? My answer is no.

While technically correct, 'brain destroyed' and 'brain dead' are two different things, and you can bring someone back with mannitol if they're brain dead. New crit is not gonna be permadeath to slimes in this way.

Bigtimetripod:
> > can you revive someone whos brain is destroyed? My answer is no.
> 
> While technically correct, 'brain destroyed' and 'brain dead' are two different things, and you can bring someone back with mannitol if they're brain dead. New crit is not gonna be permadeath to slimes in this way.

Well... An dead body cannot process reagents and defib has an limited timeframe. If you do not get some mannitol within the slime while its still alive, there is no real way to fix brain damage without surgery. Also with surgery, once the body dies, decay sets in giving an even more limited timeframe. That coupled with no cloning is near permadeath and once the brain has decayed, thats it. While SR can be used to some extent, decay kills all and leaves slimes forever in afterlife. While you could argue that transplanting the brain into an another body would fix this issue a bit, you are no longer slime at this point.

Also i'm aware at some point, normal humans decay into skeletons leaving them unclonable, but that takes a long time compared to an handful of minutes slimes have. 

Ill run some tests on private server today and compile the results.

Bigtimetripod:
So... After running multiple tests, i can say new crit is permadeath to slimes unless you can get to the body before decay can.

Long story short, after defib window has passed only way to revive a slime is by an doctor that knows how to do medical quite well and good chemist. Chemist needs to provide the station with SR and doctor needs to inject the subject with epinephrine, mannitol, oxygen curing drug (Normal Sambutanol will do) and then SR (and luck). Without any of the previous drugs, the body will keep dying and rack up Genetic damage cause of SR revives until it bursts open. Luck cause you need to hope that cardiac arrest does not continue to rack up brain damage cause Mannitol cannot out heal slimes 250% brain damage multiplier. Slimes get 7.5 brain damage per cycle and mannitol only cures 3 per cycle (normal brain damage rate is 3 per cycle for other races). If SR fails 5 times, genetic damage is too great to heal and it cannot be healed without cyro and person needs to be alive for cryo to work.

If you are fast enough and have the dead slime pumped full of drugs, you can stand infront of the cryo, force down an SR pill and at the same cycle force the slime to cryo. It improves the chances of out running the continued cardiac arrest "illness" and curing it, causing other damage to be cured in its wake.

Mitchs98:
@bigtimetripod For the record. SR decay is a non-issue when it comes to reviving slimes(usually). And you can still clone skeletonized mobs. You can also SR skeletonized mobs. The issue here is the brain damage modifier itself and the limiter on genetic damage for SR, the former of which should of been removed before this PR was merged.

PR #11991, opened 2019-08-12
Modify antag creating lavaland loot

datlo:
- Wishgranter ruin and associated code were completely removed. They are no longer really usable to begin with with telescience being removed and I dont think the ruin layout is particularly interesting or fitting. The Wild West gateway's wish granter was not touched or removed as it is a different type of machinery entirely. (wish_granter_dark)

- The Die of Fate from the Greed ruin no longer turns you into a wizard antagonist on a nat20. Instead, it drops full wizard gear, including bound spellbook with 10 spell points and teleport scroll, and allow the user to use magical artifacts. They get a big notice that they are NOT a wizard antagonist, and to follow server rules.

For non-antags : 
image

For antags : 
image

- The Gluttony loot was changed from a morph transformation syringe to a Matter Eater DNA injector.

:cl:
del: Removed the wishgranter cube lavaland ruin.
tweak: The Die of Fate no longer turns you into a wizard antagonist on a nat20. You still get full wizard gear and powers, just not the antag status.
del: Removed Gluttony's Blessing from Gluttony's ruin
add: Added a Matter Eater DNA injector in Gluttony's ruin instead
/:cl:


Markolie:
Leave the wishgranter ruin and associated code be please. It will be disabled through a configuration setting. That way it and the wish granter can still be spawned in manually. 

In addition, I think it’s a lot cleaner to just spawn the wizard through the regular manner and give them the non-antagonist message, instead of manually spawning in stuff beneath them. If they get full wizard powers anyway, I don’t see much reason not to. 

I’d also prefer at least a heads up before you directly modify features that I added less than a week ago.

datlo:
Sure, I'll change it and leave the wishgranter code, was told to make a PR to remove all antag creation from lavaland loot so thats what I did.

I disagree. There's a lot of consequences from using the make_wizard proc, which is meant to be used for the admin panel antag creation. It also makes it so that the wizard gets a new name, can change his species, gets the same setup as a regular wizard, which I don't think is the best way to do this. This is lavaland loot, it makes much more sense to have the items given to the user without teleporting him out of lavaland.

Sure, I'll keep that in mind for next PRs.

Markolie:
I think adding a parameter to that proc to prevents those unwanted changes is cleaner than effectively copying the code into the dice code. The antag panel also has separate procs for the different wizard actions, so part of that is already available. 

Having access to the wizard lair changes nothing except having access to the magic mirror and some fluff items. Considering the insanely low odds of of getting the wizard roll (it’d surprise me if it’ll happen more than once per two weeks, if not per month) I think it’s fine to give them access to the mirror.

Fox-McCloud:
The morph one arguably doesn't even make you antag.

@Markolie the die of fate can only be utilized once, correct?

Rurik123:
> The Gluttony loot was changed from a morph transformation syringe to a Matter Eater DNA injector.

Why though? Have miners becoming a morph been that much of an issue? Having a unique item that could make for an interesting round compared to a DNA block that genetics can get you (iirc at least) takes it from a 100 to 0 real fast on the funometer.

datlo:
Closing this as this PR needs more thinking and I'd rather let somebody else handle it

For the record I still think linking together the die of fate with the wizard gamemode proc is absolutely awful coding practice. They may do something similar but the two use cases (setting up a gamemode before the round starts/giving a miner loot midround) are completely different and you'll make life hard for anyone trying to modify/refactor wiz gamemode in the future just to save a few lines of code.

PR #10484, opened 2018-12-20
Allows jetpacks to work in suit storage slot

Eniraz:
**What does this PR do:**
Makes jetpacks actually work when they are on the suit storage slot.

**Changelog:**
:cl: Eschess
add: Jetpacks work properly when placed on suit slot
/:cl:

PR #11993, opened 2019-08-12
Reverts #10484

FreeStylaLT:
**What does this PR do:**
Reverts #10484. Jetpacks no longer work from the suit storage slot.

**Changelog:**
:cl:FreeStylaLT
del:Jetpacks no longer work in the suit storage slot
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -13,8 +13,6 @@
 	var/obj/item/tank/jetpack/thrust
 	if(istype(back,/obj/item/tank/jetpack))
 		thrust = back
-	else if(istype(s_store,/obj/item/tank/jetpack))
-		thrust = s_store
 	else if(istype(back,/obj/item/rig))
 		var/obj/item/rig/rig = back
 		for(var/obj/item/rig_module/maneuvering_jets/module in rig.installed_modules)


datlo:
As convenient as this was, I feel like that PR was too much of a buff to jetpacks. :+1:

Dave-TH:
How did that even get merged?

AffectedArc07:
Yes

Denthamos:
Lets not start this shit show again.  Regardless of how/when it got merged based on my eight months of observing gameplay since it has been in I can count on a single hand the amount of times it had any major impact to the game.  

In three of the four situations it was during nuclear ops where there was crew members with jetpacks fighting operatives with their jetpacks in space which made for a fun battle for those involved.  For situations in which I caused frustration for those involved I cannot name a single instance.  Again, a single instance over the past eight months.

Generally it seems (for those that actually do play consistently on the server and not just argue things on github) that players like the change and it has a net positive on gameplay and experience.

Fox-McCloud:
As I said, originally, finally getting off our butts and unifying hardsuits to be a single object instead of needlessly attaching helmets all the time.....then adding in the jetpack upgrade? That's acceptable.

There's some instances where jetpacks being built in makes sense (nukies), but on station roles? There should be some *cost* of some sort associated with it--either a time+mineral sink (R&D), or they have to give up some storage for it.

Denthamos:
@Fox-McCloud What if we were to bring in the jetpack upgrade modules from TG?  I know the syndicate jetpacks have them by default where the jetpack is built into the suit and runs off the users equipped air canisters.  Miners can get the jetpack upgrade from their vendor, and if I remember correctly there was a method for the rest of the crew (IE security/engineering) to get such upgrades as well besides mining.

Ty-Omaha:
This PR does nothing except force people to deal with a clunky unintuitive UI.

People don't give up storage
They just hold their bag in their hand
Forcing people to be subject to clunky unintuitive UI is not balance

Kyep:
@Ty-Omaha is correct. Every single time I have ever used a jetpack, I have just held my bag in my hand while the jetpack was on my back slot. Not once have I ever, or seen anyone, actually give up their suit storage. The only cost imposed is forcing people to juggle inventory in a clumsy, clunky way... which shouldn't be a method we use to balance anything.

TDSSS:
The only way in which I could see people actually take a trade-off and go with less storage is if we added some kind of small satchel that goes in a hardsuit storage slot, so you can stick that in the suit and the jetpack on your back. Trying to get anything done without a backpack-sized storage is just a total PITA.

MRgaty:
I have never seen or suffered any issues with this. Jetpacks having to be equipped in the back slot was, indeed, very clunky and made very little sense.
You still need to use storage to either a jetpack or an oxygen tank, if you uae the jetpack in the storage slot you need to either use the back slot for the tank or uae small ones. 
I think is pretty balanced right now.

MrMagolor:
>If, nevertheless, you're absolutely determined that use of jetpacks should have a downside, then figure out some other downside that isn't just punishing the player by making them deal with a fiddly interface.

If you were to ask my suggestion, jetpacking in space should have momentum, similar to the new spacepod movement on goon.

Kyep:
This is going to be closed due to objections from heads.

FreeStylaLT:
😢

PR #11994, opened 2019-08-13
Improve water handling for slimes, Grey mask/head water protection tweak

Markolie:
**What does this PR do:**
This pull request changes two water related things:

1. Water handling for slimes of the Xenobiology variety now uses the water_act system instead of relying on the turf propagating water to slimes. As a result, slimes are now affected by most sources of water. Fire extinguisher damage is roughly equal. Showers and pools effectively one-shot them.
2. Previously, if water touched a Gray wearing a mask or hat it wouldn't affect them if the volume was over 25. If it was below that, it'd still damage them. As far as I can tell, this was caused by faulty logic in the head/mask checking. It also made very little sense. Wearing a mask or hat now fully protects Gray from water damaging them.
3. Water from sinks and water effects is now 10°C (cold sink water temperature) instead of body temperature.

**Changelog:**
:cl: Markolie
balance: Slimes (of the Xenobiology variety) are now affected by regular sources of water.
balance: If a Gray is wearing a hat/mask and water touches them, it will no longer affect them. Previously, it'd still affect them in volumes less than 25.
balance: Water coming from sinks and certain water effects is now 10°C instead of body temperature.
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -405,3 +405,6 @@
 
 // Cult summon possibilities
 #define SUMMON_POSSIBILITIES 3
+
+// Water temperature
+#define COLD_WATER_TEMPERATURE 283.15 // 10 degrees celsius
--- a/code/game/objects/effects/effect_system.dm
+++ b/code/game/objects/effects/effect_system.dm
@@ -79,7 +79,7 @@ would spawn and follow the beaker, even if it is carried or thrown.
 		reagents.reaction(A)
 	if(istype(A,/atom/movable))
 		var/atom/movable/AM = A
-		AM.water_act(life, 310.15, src)
+		AM.water_act(life, COLD_WATER_TEMPERATURE, src)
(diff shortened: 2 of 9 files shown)


Markolie:
Not sure if there are huge balance implications for pools/showers one shotting slimes. The goal was mainly just to make sure slimes use the proper water system.

Fox-McCloud:
Nice refactor. I like it.

I don't think showers one-shotting slimes is a big deal; Xenobiologists who know what they're doing can already one-shot slimes with an extinguisher---so even if it becomes something people do....not really a big deal, IMO.

MrMagolor:
As a Grey player, I feel like unless you were a complete body-covering suit, immersing yourself in water (i.e. pool or lavaland beach ruin) should also result in a swift death: but I was able to swim around with just an explorer mask and suit with internals on.

muh immersion

Fox-McCloud:
conflicts.

PR #12005, opened 2019-08-14
Nerf the throwing range of facehuggers

datlo:
Shameless I ded pls nerf PR

The most frustrating part of our current xenomorphs IMO is their ranged instant stun/instant kill, which, due to throwing mechanics, is even faster than freaking bullets.

This PR nerfs the range at which you can throw xenos from 5 tiles to 2 tiles, that way the xeno actually has to expose himself to return fire to attempt a facehug throw.

:cl:
balance: Xeno facehuggers throwing range reduced from 5 tiles to 2 tiles.
/:cl:


TheSardele:
I never really found this to be the real issue, a decent helmet makes you completely immune, these sort of antags are meant to be basically unbeatable if you're not prepared. Though, I do find myself getting frustrated at the xenomorph spit, seen entire geared up sec teams get fucked over by that in the blink of an eye, though one might attribute that to bad shield RNG.

Fox-McCloud:
Of all the thing Xenos could possibly even remotely need for balance, this is absolutely not it.

datlo:
Seems my 5 minutes idedplsnerf PR isnt popular (i wonder why) so guess im closing this

Xenos do need balance work done tho

Couls:
Oh trust me, no one disagrees with that

SpessX:
2 is too little, tho I agree a facehugger shouldn't be very aerodynamic.

PR #12011, opened 2019-08-16
Corrects rods used in mass driver construction

Shadow-Quill:
**What does this PR do:**
You use three rods to make a mass driver, but only get two back.

Also lowers the number of cable coil needed to construct to two.

**Changelog:**
:cl:
tweak: Mass drivers now need two rods/cable coil instead of three to construct.
/:cl:

### DIFF
--- a/code/game/machinery/mass_driver.dm
+++ b/code/game/machinery/mass_driver.dm
@@ -160,8 +160,8 @@
 				var/obj/item/stack/cable_coil/C = W
 				to_chat(user, "You start adding cables to \the [src]...")
 				playsound(get_turf(src), C.usesound, 50, 1)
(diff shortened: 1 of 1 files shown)


code/game/machinery/mass_driver.dm
```diff
@@ -180,7 +180,7 @@
 				to_chat(user, "You begin to complete \the [src]...")
 				playsound(get_turf(src), R.usesound, 50, 1)
 				if(do_after(user, 20 * R.toolspeed, target = src) && (R.amount >= 3) && (build == 3))
```
Markolie: This amount check also needs to be adjusted in that case, I think.

Shadow-Quill: Good catch

Issue #8932, opened 2018-04-26
Unable to create sorting disposal pipes


Muhdir:
**Problem Description**:
Ever had a bomb blow all your pipes away? Felt ambitious enough to try to fix all that? Well, if the sorting disposal pipes are ever broken you won't be able to create them by any means with the RPD or the not so mobile disposal pipe dispenser. You can try to fill them in by the regular junctions, but that will mean that depending on the type of junction, either deliveries will never make it to that department or that everything through that pipe will end up in that department.

This is the pipe in question
![disposals](https://user-images.githubusercontent.com/38748088/39326801-625a5ce0-4996-11e8-9937-87e501ab44d1.png)

**What did you expect to happen**: 
I would've been able to create those disposal pipe sections. And with an destination tagger repaired the delivery system.

**What happened instead**:
I neither was able to make sorting disposal pipes nor do I start with a destination tagger to set the destination of that pipe.

**Why is this bad/What are the consequences:**
If someone either removes, destroys or displaces the sorting pipe, disposals deliveries will cease to function properly depending where that section was removed. And if you are a clown, removing and placing the sorting pipe unsets the destination and will force everything out of that departments disposal bin. 

**Steps to reproduce the problem**:
Destroy a sorting disposal pipe then scream to everyone that disposals deliveries are fucked and shouldn't be used anymore

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was
anything else out of the ordinary happening?)**: Anything else you can tell us.

Repairing pipes is obviously not done on the short gamemodes, but traitors can obviously hijack crates with less suspicion if they have these tools available to them.

Citinited:
I think (I could be wrong) disposals needs an entire overhaul before you can begin to think of dispensing sort junctions, as they're sort of hardcoded into the game right now - so while it is an issue, it would take a refactor to fix.

Fox-McCloud:
```
I think (I could be wrong) disposals needs an entire overhaul before you can begin to think of dispensing sort junctions, as they're sort of hardcoded into the game right now - so while it is an issue, it would take a refactor to fix.
```

This. Disposals code is utterly *horrifying*.

Citinited:
Right, so I took a look at this and actually it does seem like you could just make the RPD dispense a sortjunction pipe and then use the tagger to change destination, like you said. You won't be able to do anything like add new destinations to the sortjunction though. I'll look into adding it to the RPD, once the refactor goes through.

[...]

PR #12015, opened 2019-08-16
Atmos tech QoL & Disposal pipe tweaks

SteelSlayer:
**What does this PR do:**

- Rapid pipe dispensers can now place disposal sorting pipes
- When you use a destination tagger to change the location of sorting pipes, it now updates the name of the pipe as well
- Added atmospherics access to the Engineering checkpoint airlocks and the Engi-Vend vendors
- I removed some random grey pixels that I thought were out of place on disposal sorting pipes. You can see the before and after below

sortpipes

- Also removed any grey pixels I found on normal disposal junctions as well. Some pipes and orientations had them while others did not so I tried to fix them all

**Changelog:**
:cl:
add: Rapid pipe dispensers can now place disposal sorting pipes
tweak: When changing the sort location of disposal sorting pipe, the name of the pipe now also changes to the desired destination
tweak: Added atmospherics access to Engi-Vend's and Engineering checkpoint airlocks
tweak: Removed some random grey pixels from disposal sorting junction pipe sprites, and certain normal disposal junction pipe sprites
/:cl:


Muhdi345:
N I C E. My old issue I made a year ago is finally fixed with this. #8932

Will Atmosians start out with a destination tagger or will we have to bite the bullet and hope the Autolathe board doesn't get taken from us?

SteelSlayer:
Oh wow, I didn't see that issue. I don't see why they shouldn't have a destination tagger if it relates to their job, so I'll add one to their locker.

Fox-McCloud:
`Added atmospherics access to Engi-Vend's and Engineering checkpoint airlocks`

The one thing I don't care for; we went in the direction away from this for the simple matter that, well; atmostech has a lot of neat things they can do and fiddle with that engineering can't---likewise for engineers---we wanted to avoid the situation of "atmostech is just an engineer but BETTER and doesn't have to worry about setting up the engine!"

SteelSlayer:
My reasoning for the engi-vend's change is that they contain some useful things like air alarm electronics, fire alarm electronics etc. which are useful for atmos techs. My reasoning for the engineering checkpoint is that its a nice shortcut to get to science in order to fight any fires over there. 

In my atmos tech shifts I usually just hack the engi-vend to be all access and give atmospheric access to the engineering checkpoint doors anyway, as I felt atmos techs should have access to these two things. However, I can easily remove these things if you don't want them.

PR #12016, opened 2019-08-16
Karma purchase infinite loop fix

### LINKED ISSUE #11398: Misleading text when unlocking karma role
**Problem Description**:
When unlocking a karma job, I get the following text, one after another:
```
You have unlocked Nanotrasen Representative.
You have been charged 30 karma.
You do not have enough karma!
```
At the start I had enough karma for the job, but not enough to buy it twice, so it seems it first charges me the karma and then checks again if I have enough karma?

The game not to tell me I don't have enough karma.

**What happened instead**:
To get the first two messages, but not 'You do not have enough karma!'

Confusing message. I also wonder if it tries to buy the karma job twice and if I could have gotten charged twice if I had enough karma to buy the job twice.

I just opened the karma shop and bought a role. I initially joined as an observer if that matters?

I don't remember it happening the last karma purchase I made, but that was a while ago, so I dunno.

SomeSortaSquid:
**What does this PR do:**
Fixes #11398
I can't exactly explain it, but switch() responsible for choosing karma_purchase() are parts of an infinite loop I couldn't locate.
Return statements break that loop.
Client appeared to remain fully functional with these changes on private server.

**Changelog:**
:cl:
fix: Karma purchases no longer loop infinitely, leading to repeated purchase prompts or "You do not have enough karma!" message after successful purchase.
/:cl:

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -127,6 +127,7 @@
 							karma_purchase(karma,45,"job","Magistrate")
 						if("9")
 							karma_purchase(karma,30,"job","Security Pod Pilot")
+					return
 				if(href_list["KarmaBuy2"])
 					var/karma=verify_karma()
 					if(isnull(karma)) //Doesn't display anything if karma database is down.
@@ -146,6 +147,7 @@
 							karma_purchase(karma,100,"species","Plasmaman")
 						if("7")
 							karma_purchase(karma,30,"species","Drask")
+					return
 				if(href_list["KarmaRefund"])
 					var/type = href_list["KarmaRefundType"]
 					var/job = href_list["KarmaRefund"]


AffectedArc07:
If youre messing with karma, make sure you absolutely tested **with the SQL backend**.

SomeSortaSquid:
Honestly, it's a good advice. While I did check the client, I admittedly forgot to check is database itself still working as intended.
I am perfectly aware that karma system is a delicate thing that may be better left alone with a minor bug instead of patched with something potentially breaking it in an unknown way. This is why, if maintainers simply decide to close this PR, I will understand.

The database is working properly, I dare say better than before, given how this removes an infinite loop of karma_purchase() procs.
I would, admittedly, still like to find exactly _where_ that loop begins. Perhaps only one return would be enough. And it would look much better than "yeah, this code works, but don't ask why or how".

variableundefined:
Tested locally with my own SQL database. The issue exists, the PR do fix the problem. But I am left wondering why it is looping.

PR #12017, opened 2019-08-17
Fixes Lavaland Sounding Like Metal

Fox-McCloud:
Lavaland might be metal, but not that kind of metal.

:cl: Fox McCloud
fix: Fixes walking on lavaland sounding like walking on metal plating
/:cl:


KasparoVy:
Footstep sound should be deathmetal riffs! Lmao

PR #12020, opened 2019-08-17
Readd stimpacks to mining vendors

datlo:
This PR readds stimpacks, teporone injectors and the medipen bundle to mining vendors. 

They were removed as an undocumented change in #11981. Since it was undocumented and not discussed, I assume the change was not intentional.

If stimpacks are a balance concern, they should be removed or rebalanced in their own PR.

:cl:
add: Readds stimpacks, teporone and medipen bundle to mining vendors.
/:cl:

### DIFF
--- a/code/modules/mining/equipment_locker.dm
+++ b/code/modules/mining/equipment_locker.dm
@@ -404,6 +404,7 @@
 		new /datum/data/mining_equipment("1 Marker Beacon",				/obj/item/stack/marker_beacon,										10),
 		new /datum/data/mining_equipment("10 Marker Beacons",			/obj/item/stack/marker_beacon/ten,									100),
 		new /datum/data/mining_equipment("30 Marker Beacons",			/obj/item/stack/marker_beacon/thirty,								300),
+		new /datum/data/mining_equipment("Stimpack",					/obj/item/reagent_containers/hypospray/autoinjector/stimpack, 		50),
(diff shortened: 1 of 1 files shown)


SpinningSalami:
The teporone injector and medipen bundle were also removed with that PR.

datlo:
good catch

Fox-McCloud:
Yeah, this was unintentional, though I have half a notion of buffing the survival pen and keeping only a single injector as opposed to adding these (the stimpacks are *especially* suspect as a legitimate mining item).

datlo:
Maybe its just me being unrobust but I find beating Legion/Bubblegum without stimpacks to be near impossible unless you got something like the teleport cubes or other ways to cheese em. I regularly used stimpacks on lavaland and not just for antag memechem memes (although I also did that :+1: )

Denthamos:
Removal of the stimpacks and buffing of the survival pens can come in a different PR later.  For now lets focus on fixing an unintentional/undocumented change.  @Fox-McCloud

Fox-McCloud:
This is going to be closed.

After discussing it, it boils down to: yeah, it was accidentally removed---the survival injector PR was going to come (and did), and had the stimulant+teporone pens not been removed, you would have seen them removed in that PR---rather than go through the hoops of merging this only to re-remove it, we're just opting ot handle it this way.

Sorry for the confusion!

PR #12033, opened 2019-08-18
Ports TG's Fultons

Fox-McCloud:
Ports TG's fultons.

Fultons are 3-use extraction device that can be linked to an extraction beacon; you can use them to move moderately sized non-anchored items or mobs from an outdoor location to a fulton beacon (the beacon doesn't have to be outside). Useful for moving things like the dance machine or rescuing dead miners without having to interrupt your mining adventures.

Fultons are available as a mining vendor purchase, or you can spend your voucher to get a kit that has 30 marker beacons, a fulton beacon, and a fulton extraction pack.

I thought these would be a royal pain to implement---nope. One of the easier ports I've done

:cl: Fox McCloud
add: Add fulton extract packs to the mining vendor; rescue precious valuable or dead miners!
/:cl:


TDSSS:
feel pretty silly and memey for Paradise, personally.

Fox-McCloud:
You do realize it's an actual, real thing, right? https://en.wikipedia.org/wiki/Fulton_surface-to-air_recovery_system

JKnutson101:
Cool! Now if only Miner's didn't disconnect when they died so we could actually recover them. 👀

Lunaimon:
Can these carry things across z-levels?

Leanfrog:
Will you be able to fulton yourself to get on out of there in a bad situation?

Fox-McCloud:
>Will you be able to fulton yourself to get on out of there in a bad situation?

Yes, but there's a long delay; jaunters are better for poofing out of danger.

Fultons are better for rescuing.

Citinited:
Finally, a better way than the RCS to send stuff back to the station!

TDSSS:
>You do realize it's an actual, real thing, right?

I thought metal gear made that one up, lol. Fair enough.

JohnBigless:
Would it be possible to fulton another player who is still alive?

Melloreaper:
This may be needed sense the only other method of extracting off z levels is gone.(tele science)

PR #12035, opened 2019-08-19
Adds Fans to Mining and Engineering Shuttles

Fox-McCloud:
Due to the nature of space wind on TG versus Paradise, this is probably necessary, even if I don't like it.

Needless to say, the amount of miners getting spaced by shuttles, these days, is ridiculous.

This adds fans to the docks on the Cyberiadd and also on the mining+engineering shuttle itself.

This should prevent you from rocketing out, instantly, if someone suddenly calls the shuttle.

Likewise, if someone calls it while you're standing right in the door, you should just face plant, and as long as you don't walk out, when you get up, you should be able to walk back in.

:cl: Fox McCloud
tweak: Adds mini-fans to the mining, engineering, and security shuttles
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -31407,6 +31407,7 @@
 	id_tag = "s_docking_airlock";
 	req_access_txt = "48"
 	},
+/obj/structure/fans/tiny,
 /turf/simulated/shuttle/floor,
 /area/shuttle/mining)
 "bgZ" = (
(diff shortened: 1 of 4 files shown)


Shadow-Quill:
Could you do this for delta&meta as well?

TDSSS:
I dislike this, tiny fans are a cop-out. I'd much prefer it if we added a delay to shuttle calling/recalling, with some time for doors to close and bolt or whatever.

Shadow-Quill:
@TDSSS This would be nice in the meantime, while we don't have that.

JohnBigless:
> I dislike this, tiny fans are a cop-out. I'd much prefer it if we added a delay to shuttle calling/recalling, with some time for doors to close and bolt or whatever.

Why not **BOTH?**

PR #12037, opened 2019-08-19
Mining shuttle launch delay and warning message

Couls:
**What does this PR do:**
The mining shuttle now has a three second delay before launching and sends a warning message prior to launching. Also cleaned up a bit of the timer code(credit to TG for a large portion of this code!)

image

**Changelog:**
:cl:
add: mining shuttle warning message and delay before launching
/:cl:

### DIFF
--- a/code/__DEFINES/stat.dm
+++ b/code/__DEFINES/stat.dm
@@ -41,6 +41,7 @@
 #define SHUTTLE_STRANDED 4
 #define SHUTTLE_ESCAPE 5
 #define SHUTTLE_ENDGAME 6
+#define SHUTTLE_IGNITING 7
 
 // Shuttle return values
 #define SHUTTLE_CAN_DOCK "can_dock"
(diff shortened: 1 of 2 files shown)


Breenland:
I do agree a 5 second delay seems too long. A 3 second delay sounds good, and if possible (though this is maybe another PR), have the door and exterior airlock clones and bolt on the final second. This would also be helpful to have on any other shuttle, like the Engineering, Labor Camp, and possibly ERT. The latter two may not need it so much, but Engineering still has people get lost ejected  and then. 

This makes me think now, it’d be useful to have a final countdown for when the emergency shuttle and escape pods are leaving.

Couls:
This kind of lays down a framework for something like that, which TG already has. I just wanted to get this small thing done and out first considering its a much more pressing and common issue due to lavalands popularity and the upcoming updates and current PRs out for it. I'll bump down the delay to three seconds though! Thanks for the feedback.

Emanthealmighty:
Could you add one for the labor shuttle at sec too? Just yesterday a sec officer died cause someone launched the shuttle while he was near the open airlock.

Couls:
This should work for labour shuttle as well, basically everything but the emergency shuttle

Issue #10286, opened 2018-11-19
APC menu doesn't auto-open on powerloss hack success.


Shadow-Quill:
**Problem Description**:
What is the problem?

After an unpowered AI hacks the area APC, the control panel doesn't open.

**What did you expect to happen**: 
Why do you think this is an issue?

The APC control panel to automatically open after the hack succeeds.

**What happened instead**:
How is what happened different from what you expected?

Okay, okay. You might think I'm crazy but... 
It did not.

**Why is this bad/What are the consequences:**
Why do you think this is an important issue?

If an AI loses power from an APC outside of their FOV, they cannot interface with it and re-enable power.

**Steps to reproduce the problem**:
The most important section. Review everything you did leading up to causing the issue.

1. Be an AI,
2. Turn off power in your core.
3. Wait for the hacking sequence to finish.
4. Observe how nothing happens, but you can now interact with the APC by clicking on it.

**When did the problem start happening**:
If your report is about something that used to work but no longer does, when was the last time you remember it working?

Just found it out today.

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was anything else out of the ordinary happening?)**:
Anything else you can tell us.

Private testing server, only I was on it. 

Probably has something to do with [this line](https://github.com/ParadiseSS13/Paradise/blob/bed2c8ae3464783254bb30538d3cd47193fcf226/code/modules/mob/living/silicon/ai/life.dm#L126).

PR #12038, opened 2019-08-19
Fixes APC menus not opening after the powerloss hack

Shadow-Quill:
**What does this PR do:**
Fixes: #10286

You're supposed to open the menu of whatever area you're in (after around 30 seconds) so you can save yourself.
This fixes not being able to do that.

Also moves the law reminder to a more sensible place.

**Changelog:**
*Remove this line, and appropriate fields from the changelog*
:cl:
fix: AIs not being able to restore their own power if in an area with an APC.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/life.dm
+++ b/code/modules/mob/living/silicon/ai/life.dm
@@ -107,6 +107,8 @@
 							aiRestorePowerRoutine = 0
 							update_blind_effects()
 							update_sight()
+							to_chat(src, "Here are your current laws:")
+							src.show_laws()
 							return
 
 						switch(PRP)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/ai/life.dm
```diff
@@ -107,6 +107,8 @@
 							aiRestorePowerRoutine = 0
 							update_blind_effects()
 							update_sight()
+							to_chat(src, "Here are your current laws:")
+							src.show_laws()
```
variableundefined: implicit src

Shadow-Quill:
I'm stoopid. First committed to master then committed without muh commit message.

Also: Mobile webedit.

variableundefined:
😱

PR #12040, opened 2019-08-20
Improve shuttle parallax stopping + fix shuttles flying sideways

Tayyyyyyy:
**What does this PR do:**
Just found out `/area/shuttle` exists. Updated shuttle parallax to be more similar to tg's using `/area/shuttle`. This is a better/faster way to stop shuttle parallax (if you fall out or leave the shuttle), and also reactivates parallax if you move back into the transiting shuttle.

Also updates a bunch of shuttles to move east/west/south instead of north

:cl: Tayyyyyyy
fix: Many shuttle flight directions fixed
/:cl:


Fox-McCloud:
Not ideal, but until we get dynamic shuttle areas and can implement this properly...

PR #11364, opened 2019-04-24
Kills off Object and Fast Object Processes

Fox-McCloud:
Kills off object processing and fast processing; ports them to subsystems.

This makes `/datum/proces/process` use `set waitfor = FALSE` so it doesn't block if it sleeps.

This has the effect of completely destroying the functionality of the Process Scheduler---I'm also not sure how it'll work with the Ticker either.

As such, the names of these `process` functions have been renamed to `process_decrepit` as a temporary name until Ticker is converted and/or the scheduler is completely destroyed.

I've done basic testing of this, and it seems to work, but I'd like to do a bit more testing, just to be sure.

:cl: Fox McCloud
tweak: object and fast processing are now proper subsystems
/:cl:

PR #9114, opened 2018-06-14
Fix emitters

MINIMAN10000:
The order in which power is consumed is dependent on the order in which they were modified on the map. Defer the SMES processing until after all normal processing has been done so they don't suck up all the power before it reaches the containment emitters. Fixes #9095
:cl:
fixes: Containment emitters
add: deferred processing of SMES
/:cl:

PR #12042, opened 2019-08-20
Fix emitters again

MINIMAN10000:
Back in #9114 I fixed emitters by making it so batteries would act like batteries by being the last in line to take power in.
This code got removed in #11364 I blame the lack of obvious naming. So this time around I went the route of adding a "is_battery" helper to try and make it a bit more obvious what's going on so hopefully it doesn't end up getting dropped again.

**What does this PR do:**
SMES once again act like batteries and are last in line to get power, this means no more having to rebuild emitters while setting up the engine ( or otherwise turning off SMES power input ) in order to force the powernet to do what it should have done in the first place. Which is use the SMES as a battery instead of it feeding itself.

**Changelog:**
:cl: MINIMAN10000
fix: SMES once again act like batteries and are last in line to get power
/:cl:

### DIFF
--- a/code/__DEFINES/MC.dm
+++ b/code/__DEFINES/MC.dm
@@ -19,7 +19,7 @@
 
 #define NEW_SS_GLOBAL(varname) if(varname != src){if(istype(varname)){Recover();qdel(varname);}varname = src;}
 
-#define START_PROCESSING(Processor, Datum) if (!Datum.isprocessing) {Datum.isprocessing = TRUE;Processor.processing += Datum}
+#define START_PROCESSING(Processor, Datum) if (!Datum.isprocessing) {Datum.isprocessing = TRUE;if(is_battery(Datum)){Processor.processing.Insert(1,Datum)}else{Processor.processing += Datum}}
 #define STOP_PROCESSING(Processor, Datum) Datum.isprocessing = FALSE;Processor.processing -= Datum
 
 //SubSystem flags (Please design any new flags so that the default is off, to make adding flags to subsystems easier)
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -15,6 +15,8 @@
 
 #define ismachinery(A) (istype(A, /obj/machinery))
 
+#define is_battery(A) (istype(A, /obj/machinery/power/smes)) // If something needs to be last in line to get power from the powernet
+
 #define ismecha(A) (istype(A, /obj/mecha))
 
 #define iseffect(A) (istype(A, /obj/effect))


datlo:
Speedmerge please :+1:

SomeSortaSquid:
Doesn't that mean SMES can no longer be turned into powersinks?
I think that possibly bug should be a feature.

Fox-McCloud:
This occurred in conjunction with the power refactor and was not accidental.

After testing both the TG and Paradise power systems and how they behave and interact, I came to the conclusion that it was probably me that was doing something wrong, with power, and that this was most likely not a "bug".

So, what's actually happening is that the SMESes are taking in enough power that there isn't any power left for anything else.

This is very easily remedied though, and I updated the engine setup guide for how to avoid and correct it without having an engine release or other problems; there was an announcement made about this, and info was handed off to mentors as well.

It seems some returning players haven't gotten this memo though, so they continue to believe this behavior is a mistake.

Short of it: following proper engine setup guidelines means the engine/emitters will start fine; this is less a problem/bug and more just initial misunderstanding of how powernets work.

MINIMAN10000:
To be clear my definition of a battery is an item that provides power.
This means that a battery should not be consuming power.
Excess power however can be fed back into a battery as it charges.
The way power works in beyond is the powernet will have a specific amount of power in the network. The order in which items consume power from the network is based off when they were placed. This means you can arbitrarily change the order in which items get power by constructing them in a specific order.
This applies both to mapping and constructing them in the game as well.
This power is consumed until there is nothing left. So you can in fact place an SMES to be the first machine in that list making it so while charging and providing power you can make it so no other machines are able to consume power. Thus making it not a battery.
SMES is as the name says energy storage IE battery.
This PR is to make the behavior of an SMES behave as a battery should.

> Short of it: following proper engine setup guidelines means the engine/emitters will start fine; this is less a problem/bug and more just initial misunderstanding of how powernets work.

Yes you can in fact work around this bug, we can also go into the map editor and replace the entities in a different order to fix the behavior that way as well.

But the reality is a battery should act as a battery and it should be enforced through code.

Crazylemon64:
Are you sure this will work? From what I understand, the whole problem comes about due to the SMES coming earlier in the machinery list, and changing SMES init code to put them at index 1 wouldn't modify that relation. However, an approach that uses `LateInitialize` would be capable of adding itself to the list after all other machinery is in place - though newly constructed machinery attached to the powernet (APCs, primarily) would still come after the SMES.

Looking closer at the SMES process code, it performs two passes:
[The first](https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/power/smes.dm#L226), takes in the available power at the terminal, and stores it
[The second](https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/power/smes.dm#L248), adds power to the powernet under the SMES machine itself.

[At the end of every power tick](https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/power/powernet.dm#L80), the powernet undergoes a reset where it stores all excess energy in  SMESs by calling the `restore` function.

If I'm understanding correctly, by moving the SMES to the first entry in the list, it'll draw power before any other machine, but since there hasn't been any energy injected into the network yet by the SMES on that power tick, it won't be able to siphon itself.

However, if you had multiple SMES hooked in a shorted-terminal configuration, the second SMES in the list should be capable of siphoning the entire output of the first, regardless of this change.

Now I'm wondering what prevents machinery placed earlier in power processing than the SMES in the network from failing to draw power before the SMES has had a chance to tick for that power cycle.

Crazylemon64:
On further reading, it appears `newavail` is what the powernet uses to carry over power across ticks.
So it goes
`smes storage -> newavail -> (powernet reset)avail -> (powernet reset)smes storage`?

Crazylemon64:
If that's the case, I'd imagine moving the SMES to the first entry would cause it to siphon off the lion's share of `newavail` before any other machine, correct?

MINIMAN10000:
Thanks for looking into that. I'll look closer into it and mark it WIP in the meantime.

Crazylemon64:
The strategy for getting around this, without any code changes, would be to make sure the input and output terminals are on different power networks - that way an SMES can't eat all of its own power.
Perhaps if some kind of indicator was added that showed that an SMES's input was connected to the output (by seeing if the powernet the SMES is connected to is the same powernet as the terminal), that could help make these sorts of problems obvious to the electricians.

Issue #12048, opened 2019-08-22
Cardboard boxes are extremely fast when trying to move diagonally


Lunaimon:
**Problem Description**:
Cardboard boxes become super fast when moving diagonally into a wall.

**What did you expect to happen**: 
The box to move at normal speed and not go really fast.

**What happened instead**:
My box moved very fast whenever I held two movement keys while next to a wall.

**Why is this bad/What are the consequences:**
Cardboard boxes are as fast as a speedbike! New antag strat!

**Steps to reproduce the problem**:
Get a cardboard box. Get inside it and close it. Move up to a wall, then press two movement keys together, one towards the wall and the other in the direction you want to go. Become very fast.
Wall not required, also works with doors, tables, and other things that normally impede movement.

**When did the problem start happening**:
Since the diagonal movement PR, probably.

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was anything else out of the ordinary happening?)**:
I had changed the box's look a few times with a pen before I started moving around with it. Probably not related.

PR #12053, opened 2019-08-22
Fix cardboard box walldash

Couls:
**What does this PR do:**
Not really that urgent but may as well fix it anyway. credit to TG for the code here

Fixes: #12048 

**Changelog:**
:cl:
fix: cardboard boxes are no longer sonic speed
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/cardboardbox.dm
+++ b/code/game/objects/structures/crates_lockers/closets/cardboardbox.dm
@@ -18,12 +18,14 @@
 /obj/structure/closet/cardboard/relaymove(mob/user, direction)
 	if(opened || move_delay || user.stat || user.stunned || user.weakened || user.paralysis || !isturf(loc) || !has_gravity(loc))
 		return
-	move_delay = 1
-	if(step(src, direction))
+	move_delay = TRUE
+	var/oldloc = loc
+	step(src, direction)
+	if(oldloc != loc)
 		spawn(config.walk_speed)
-			move_delay = 0
+			move_delay = FALSE
 	else
-		move_delay = 0
+		move_delay = FALSE
 
 /obj/structure/closet/cardboard/open()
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/closets/cardboardbox.dm
```diff
@@ -18,12 +18,14 @@
 /obj/structure/closet/cardboard/relaymove(mob/user, direction)
 	if(opened || move_delay || user.stat || user.stunned || user.weakened || user.paralysis || !isturf(loc) || !has_gravity(loc))
 		return
-	move_delay = 1
-	if(step(src, direction))
+	move_delay = TRUE
+	var/oldloc = loc
+	step(src, direction)
```
Fox-McCloud: did you try what TG did with `if(step(src, direction))` then using timers?

Couls: the issue isn't timers, the if check is the reason you have no move delay when you move diagonally against walls

Fox-McCloud: I'm...not suggesting the issue is with timers.

Fox-McCloud: Please make this:

```
/obj/structure/closet/cardboard/relaymove(mob/living/user, direction)
	if(!istype(user) || opened || move_delay || user.incapacitated() || !isturf(loc) || !has_gravity(loc))
		return
	move_delay = TRUE
	var/oldloc = loc
	step(src, direction)
	if(oldloc != loc)
		addtimer(CALLBACK(src, .proc/ResetMoveDelay), config.walk_speed)
	else
		move_delay = FALSE

/obj/structure/closet/cardboard/proc/ResetMoveDelay()
	move_delay = FALSE
```

Issue #12054, opened 2019-08-22
Airlocks don't close themselves anymore


farie82:
**Problem Description**:
Airlocks don't close themselves anymore and changing the timing won't do anything

**What did you expect to happen**: 
For them to close per usual

**What happened instead**:
Bridge and brig wide open because somebody didn't close it behind their back

**Why is this bad/What are the consequences:**
It's broken AF

**Steps to reproduce the problem**:
(might have to set the timers first, aka shortcut them)
Open a door

**When did the problem start happening**:
I've noticed it starting since of today (22-8-2019)

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was anything else out of the ordinary happening?)**:
Game was chugging pretty badly as well

TDSSS:
Probably caused by #12047

Shadow-Quill:
Can confirm, only happens with overridden timed doors

TDSSS:
Can reproduce it locally with overriden timing, yeah.

Fox-McCloud:
Not entirely correct; this only applies to overriden ones, as @Shadow-Quill technically pointed out.

PR #12056, opened 2019-08-22
Fixes Door Timer Override

Fox-McCloud:
Fixes door timer overrides not working.

When I did the refactor, I kept the old value to keep balance more or less the same; little did I know, that'd have an actual impact---the timer value was just too small for actually be registered on the timer SS, so the door would never close. Very strange---actual delay difference you probably won't notice

fixes: https://github.com/ParadiseSS13/Paradise/issues/12054

PS: Door timing overriding is powergaming, change my mind.

:cl: Fox McCloud
fix: Fixes overriden door timers not working
/:cl:

### DIFF
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -20,7 +20,7 @@
 	var/welded = FALSE
 	var/normalspeed = 1
 	var/auto_close_time = 150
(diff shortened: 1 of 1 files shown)


Shadow-Quill:
You have been redeemed.

Issue #12058, opened 2019-08-23
found a bug while codediving for a friend


ATH1909:
It's in the code for the sneezing symptom:
https://github.com/ParadiseSS13/Paradise/blob/6fe83a698757f023d26171355fff906d95ee764b/code/datums/diseases/advance/symptoms/shivering.dm

For some reason, it will only try to chill you if your body temperature is BELOW the body temperature cold damage limit.

PR #12059, opened 2019-08-23
Shivering should now cool you down properly

Shadow-Quill:
**What does this PR do:**
Fixes: #12058
Switched a < to a >.

Untested mobile webedit.

**Changelog:**
:cl:
fix: The shivering symptom in virology now properly chills people.
/:cl:

Sent from my iPhone.


ATH1909:
Say, while you're here, are you sure that that bodytemperature adjustment is supposed to use a min instead of a max? Right now, it looks like shivering CAN cool you down below a temperature limit, while fever CAN'T heat you up above a temperature limit.

PR #12075, opened 2019-08-25
IV bag tweaks, adds handheld defib to paramedic's locker

SteelSlayer:
**What does this PR do:**
The IV bag can now transfer its contents to other containers by clicking on the container with the bag. It has a few new transfer amounts being: 15, 25, 30, 50. 

However, you cannot use these new transfer amounts when using the IV on a person as it would be silly to be able to push 50 units of something through the small tube on an IV bag. If you try to inject someone while on these high settings you'll get this message:

Untitled-1

Likewise you cannot change the transfer amount from something acceptable to a value not allowed while you have an IV in someone. It will display a similar error message and remove the needle from them. Hopefully that retains the balance on IV bags, while making it not so annoying to remove chems from them using a syringe at 5u per click.

**Changelog**:
:cl:
tweak: IV bags are able to transfer their contents to other containers by clicking on them with the bag
add: Adds a handheld defibrillator to the Paramedic's EVA closet
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/medical.dm
@@ -273,6 +273,7 @@
 	new /obj/item/key/ambulance(src)
 	new /obj/item/pinpointer/crew(src)
 	new /obj/item/clothing/shoes/magboots(src)
+	new /obj/item/handheld_defibrillator(src)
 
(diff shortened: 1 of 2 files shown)


Fox-McCloud:
This PR needs to be split in two--as they're extremely unrelated.

SteelSlayer:
Alright, I'll just remove the defib part and put it into its own PR then

PR #12082, opened 2019-08-25
[WIP] Add vox/plasmaman anesthetics, remove sleep surgery

datlo:
This PR adds one Vox and one Plasmaman anesthetics tank to surgery ORs, and removes being able to sleep to undergo surgery safely. 

If any surgery step is performed without anesthetics, the patient will wake up with a message warning him of the pain, making it very clear that it is no longer an effective method of anesthesia.

Reagents that force you to sleep such as ketamine (sleepy pens) are effective painkillers too.

Aside from the patient sleeping being an effective form of painkiller making no sense, it is not intended in code either :
 
`
if(M.stat) //stat=0 if CONSCIOUS, 1=UNCONSCIOUS and 2=DEAD. Operating on dead people is easy, too. Just sleeping won't work, though.
`

TODO : Sprites for the new anesthetics tanks.

:cl:
add: Added vox and plasmaman anesthetics tanks to OR closets.
tweak: You can no longer use the Sleep verb as an effective method of anesthesia.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/pain.dm
+++ b/code/modules/surgery/organs/pain.dm
@@ -5,7 +5,9 @@
 // partname is the name of a body part
 // amount is a num from 1 to 100
 /mob/living/carbon/human/proc/pain(partname, amount)
-	if(stat >= UNCONSCIOUS)
+	if(stat == DEAD || status_flags & FAKEDEATH)
+		return
+	if(stat == UNCONSCIOUS && paralysis)
 		return
 	if(reagents.has_reagent("sal_acid"))
 		return
@@ -31,9 +33,10 @@
 
 // message is the custom message to be displayed
 mob/living/carbon/human/proc/custom_pain(message)
-	if(stat >= UNCONSCIOUS)
+	if(stat == DEAD || status_flags & FAKEDEATH)
 		return
-
+	if(stat == UNCONSCIOUS && paralysis)
+		return
 	if(NO_PAIN in dna.species.species_traits)
 		return
 	if(reagents.has_reagent("morphine"))
@@ -41,6 +44,9 @@ mob/living/carbon/human/proc/custom_pain(message)
 	if(reagents.has_reagent("hydrocodone"))
 		return
 
+	if(sleeping)
+		SetSleeping(0)
+		to_chat(src, "<span class='userdanger'>Pain wakes you up!</span>")
 	var/msg = "<span class='userdanger'>[message]</span>"
 
 	// Anti message spam checks
--- a/code/game/objects/items/weapons/tanks/tank_types.dm
+++ b/code/game/objects/items/weapons/tanks/tank_types.dm
@@ -59,6 +59,28 @@ obj/item/tank/oxygen/empty/New()
 
 	air_contents.trace_gases += trace_gas
 
+/obj/item/tank/anesthetic/vox
+	name = "vox anesthetic tank"
+	desc = "A tank with an N2O/N2 gas mix."
+	icon_state = "anesthetic"
+	item_state = "an_tank"
+
+/obj/item/tank/anesthetic/vox/New()
+	..()
+	air_contents.oxygen = 0
+	air_contents.nitrogen = (3*ONE_ATMOSPHERE)*70/(R_IDEAL_GAS_EQUATION*T20C) * O2STANDARD
+
+/obj/item/tank/anesthetic/plasmaman
+	name = "plasmaman anesthetic tank"
+	desc = "A tank with an N2O/Plasma gas mix."
+	icon_state = "anesthetic"
+	item_state = "an_tank"
+
+/obj/item/tank/anesthetic/plasmaman/New()
+	..()
+	air_contents.oxygen = 0
+	air_contents.toxins = (3*ONE_ATMOSPHERE)*70/(R_IDEAL_GAS_EQUATION*T20C) * O2STANDARD
+
 /*
  * Air
  */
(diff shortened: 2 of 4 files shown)


pinatacolada:
I think its more unique not having anesthetics for those 2, like its "oh this is a giant dumb bird, i need to use a different way to sedate it appart from the usual one used on normal human beans", vs "oh its a bird, i need a different gas". To sum I think its more engaging for medbay not having these.

But the sleep part yes fuck yes 💯💯👌👌👏🤾‍♀️🎣✌️ speed merge

Can you check salicylic acid too, i think the times i used it surgery steps were still failing, which it probably shouldnt since its a painkiller just like hydrocodone which works great

datlo:
Saly acid is a very mild painkiller and while it will reduce the chance to fuck up surgery you can still fuck up, its working as intended.

pinatacolada:
Guess i just have shit rng then

lolzyking:
This doesn't mess with ipc snowflake surgery right?

datlo:
IPCs have the NO_PAIN species trait so they arent affected by this

code/modules/surgery/organs/pain.dm
```diff
@@ -31,16 +33,20 @@
...
 	if(NO_PAIN in dna.species.species_traits)
 		return
 	if(reagents.has_reagent("morphine"))
 		return
 	if(reagents.has_reagent("hydrocodone"))
 		return
 
+	if(sleeping)
+		SetSleeping(0)
+		to_chat(src, "<span class='userdanger'>Pain wakes you up!</span>")
```
Fox-McCloud: Broken bones an damaged organs now counter some antag tactics and knockouts with this.

Yeah, not a good idea.

KasparoVy: >Gets sleepypenned

>VIGOROUSLY STABS SELF WITH SCREWDRIVER TO STAY AWAKE

Fox-McCloud:
As if Ghetto surgery wasn't unviable enough, as is, this pretty much completely destroys it.

One if the big aspects of the "need to be unconscious for surgery without effort" is to help cut down on the powergaming of "NO ANESTHETIC EVER"---and to re-introduce a bit of additional risk with surgery.

self-sleeping still keeps that risk in place and allows for cases where things are completely and totally screwed (and ghetto surgery) for surgery to still be somewhat viable.

Not keen on this change, personally.

code/modules/surgery/organs/pain.dm
datlo: Chemicals set you to Paralyse and not sleeping, which wouldnt get you to wake up.

The sleep verb is basically the only thing that directly sets Sleeping without going through paralyse.

datlo:
Self sleeping is super powergamey. 'I sleep so the surgery can be done faster and I can get back to antagging/validhunting/whatever faster.'. It also allows you to wake up halfway through the surgery so if the doctor is killing you you can stop it if you are lucky, otherwise just go back to sleep.
 
Forcing people to use anesthetics or have surgery last forever from all the organ damage caused is a reduction in powergaming, not an increase.

SomeSortaSquid:
Perhaps add pain-killing properties to some alcohols as a form of ghetto anesthetics?
I don't think ghetto surgery comes into the spotlight often enough to warrant balancing the system around it. Very often, you have to rely on partial ghetto, which still has decent success chances, even without anesthetics.
Full ghetto, with just a table and no surgical tools, happens pretty much never.
As much as I'd like to make it slightly more viable, I would like self-sleeping through the surgery gone even more. Medbay has a lot of flavourful anesthetic options and they're all wasted because "I'll just sleep". I have had cases where I couldn't even use hydrocodone, because they wanted to sleep instead.

datlo:
Being drunk is already a relatively effective painkiller.

TDSSS:
You want ghetto surgery, just chug some whiskey/vodka. Being totally drunk reduces failure chance from 20% to 5%

Mitchs98:
Entirely against removing sleep surgery. Overall makes things a huge PITA for slimes, vox, plasmamen, and slimes. Unintended or not, it's an accepted mechanic by most players. Only some people want it changed/removed for being 'powergamey' or 'metagamey'. There are many people choose sleep surgery over anesthetics, and most of them being races that don't take conventional anesethics/need to be drugged. 

Personally I prefer not using them either way because many doctors are extremely new and staring at a black screen for 30+ minutes while drugged/forced unconscious while they figure out how to patch up a few bones ain't fun.

Fox-McCloud:
>Self sleeping is super powergamey. 'I sleep so the surgery can be done faster and I can get back to antagging/validhunting/whatever faster.'. It also allows you to wake up halfway through the surgery so if the doctor/an intruder is killing you you can stop it if you are lucky, otherwise just go back to sleep.

Under that light, so is using painkillers, as they allow you to be FULLY awake with the full chance of retaliation at a split second---compared to sleeping, which you can't wake up from and are full out blacked out; the argument could easily be made that painkiller are the most powergamey of the 3 surgery supression methods.

>As for ghetto surgery, its perfectly doable for an enterprising maintenance surgeon to get a bottle of morphine, or hell steal/borrow an anesthetic tank. 

This isn't ghetto surgery; that's just setting up surgery in a shadey place. Ghetto surgery is using stuff like hatchets, cigarettes, kitchen knives, cable coils, and the likes to perform surgery. The success rate of these tools is (appropriately) low; the one thing that kept ghetto surgery somewhat viable, after the "you have to be asleep or have painkillers or surgery fails most of the time" PR, was the fact that you *could* put yourself to sleep.

Other considerations:

Vox are going to have to be poisoned just to get the anesthetic on---not great
Slime people don't even breathe and can't even have anesthetic to begin with; unless you specifically have access to a couple of chems, you're going to have a not great time.

datlo:
Seems medical in general will need larger work done for something like this to be implemented. I'll give it more thought and might give this another approach later. Closing for now.

PR #12083, opened 2019-08-25
Multiple proposed changes to antag theft/document objectives

AzuleUtama:
**What does this PR do:**
Based off the community meeting's discussion regarding certain objectives antags can be assigned, I've made a few suggested changes both to the items traitors can be asked to steal, as well as how document trading is handled for traitors.

**Removal**
The objectives to steal a slime extract or a tank of plasma have been removed, it was brought up in the meeting that these objectives were too easy to complete and in many cases you could just walk up and ask for either, which doesn't make for an interesting traitor round.

**Documents**
Having to trade documents (or backstab the other traitor afterwoods) no longer counts towards the total number of objectives assigned to an antagonist. This means that the two agents who get this assigned will have an extra, normal objective to complete alongside it. I implemented this suggestion as currently document traders have nothing to do following the trade itself, which usually occurs in the first few minutes of the round.

**Additions**
The following item has been added to the list of potential objects antags can be asked to steal:

- Captain's Rapier

If anyone has any additional or better suggestions for potential new items for antagonists to be tasked with stealing, please let me know.

Here is an example screenshot showing a document trader now having an extra objective, as well as one of the new objectives:

image

**Changelog:**
:cl:
add: The Captain's rapier can now be assigned as a theft objective.
del: Slime extracts and plasma tanks have been removed as theft objectives.
tweak: Document trading no longer counts towards a traitor's total number of assigned objectives.
/:cl:


TDSSS:
Since I don't have BS unlocked, I'll just ask...does the BS spawn with the revolver or does it spawn in his locker? In the first case, you might be totally unable to complete your objective if there is no BS. If it spawns in his locker, it seems like an objective that's too easy if there's no BS.

I like the other objectives, though.

datlo:
'The objectives to steal a slime extract or a tank of plasma have been removed' I definitely approve this change. It was refused the last time I tried to change that, and it wasnt the first time I believe.

The Blueshield revolver spawns on the BS, making the objective potentially impossible to complete. 

I am not sure I like the medals being antag objectives. This will turn a fun RP thing into another must prevent redtext item.

AzuleUtama:
> 'The objectives to steal a slime extract or a tank of plasma have been removed' I definitely approve this change. It was refused the last time I tried to change that, and it wasnt the first time I believe.
> 
> The Blueshield revolver spawns on the BS, making the objective potentially impossible to complete.
> 
> I am not sure I like the medals being antag objectives. This will turn a fun RP thing into another must prevent redtext item.

The cryopod code should prevent any theft objectives from being deleted if the BS goes to cryo. So it should appear in the console along with other items like the compact defib or reactive armor. Unless I'm missing something there.

I can see people potentially doing that with the medals, but you could also say the same about the CMO not giving the paramedic the compact defib to make their job a bit easier because "you could be an antag!", or other departmental items. I'll consider changing or removing them if the general opinion is that they're a bad idea though.

Quantum-M:
I don't think the Blueshield revolver should be a theft objective, given that it only spawns on the Blueshield. I don't think it should start in their locker, given to how isolated their office is and that it's a karma job.

Abydos2:
The Blueshield's revolver is spawned on him and that is if someone decides to be a Blueshield.

Also, document trading, the traitor has a 50% chance of gaining a secondary objective already that states they are to get the red document but not give up the blue document. Meaning they have to trick or even kill the other person to get it.

AzuleUtama:
> The Blueshield's revolver is spawned on him and that is if someone decides to be a Blueshield.
> 
> Also, document trading, the traitor has a 50% chance of gaining a secondary objective already that states they are to get the red document but not give up the blue document. Meaning they have to trick or even kill the other person to get it.

It's only a 20% chance for each document holder to get backstab, not 50%. The probability is set on line 413 of traitor.dm.

Even if it were 50% though, it's still effectively a gamble to get an extra objective to kill a fellow traitor you could be working alongside following the document trade. Hell, you could even work with one another on an objective, hang out and watch each others backs only to then backstab them at the worst moment. Right now with backstab, it's either play the old "replace the documents in the folder" trick on the first meetup and hope they don't notice or just kill them once you meet up since neither of you have anything else to do. Neither of which are very interesting.

Kyep:
Having the dept medals be antag theft objectives undermines their effectiveness as incentives for the crew to do their jobs, and makes heads less likely to award them. It also makes people less likely to accept them, knowing that accepting a medal may make them a target.

If you want to add new unique theft objectives, I'd start out with, say, the head of department stamps, the department pets (not sure how you'd capture them), unique and useful items (the Captain's rapier) and so forth.

KasparoVy:
Pokeballing department pets after emagging the mining vendor and stealing some points somehow? Hmmm.

Mitchs98:
Against the medals and revolver being steal objectives personally for reasons stated above. Having the detectives' gloves be an objective just feels tacked on as well. Sure there's only one pair on the station, but at the same time there's also only one champion belt in the vault for example. It doesn't feel like an item NT would go after. It's just specially made gloves so that no fibres or fingerprints contaminate a crimescene. It may only be fingerprints too, unsure. Which makes them even less special.

Good ideas in theory, probably won't be in practice.

AzuleUtama:
Following some of the initial feedback, I've removed the medals and BS's laser gun from the list of potential items to steal. 

It's clear now that the BS's gun will cause problems if no Blueshield ever joins, and leaving only one inside the locker will likely make things too easy when it comes to stealing it should there not be one at roundstart. I can also now see the medals causing 'valid' issues with heads being reluctant to ever hand them out because 'you might be an antag!'.

I've implemented Kyet's suggestion of the captain's rapier as a theft objective. Captains almost always keep it on their person and if they don't, there's a few more doors to get to it compared to the laser gun / jetpack (albeit this makes little difference with an emag). It also fits nicely alongside the other captain unique items they already have to steal.

I'm not personally a fan of stamps as targets, they seem a bit.... underwhelming of an item to steal. Most heads probably won't even need to use them or notice they've been stolen should someone break in and swipe it. Door remotes for each department might be an alternative since heads will usually secure them better and the item itself is useful to traitors (namely for getting out of the department after stealing it).

Department pets does sound pretty unique and funny as an idea, however I can see a couple of issues with it. The biggest is the greytide randomly killing and gibbing the pets at roundstart, making the objective impossible to complete. From memory there's no rule against doing this, just a fairly minor jail time and possible lynching. The other is exactly how they need to steal them in the first place. Capsules might be an idea, but do we need to give traitors one at roundstart to steal them with or expect them to get ahold of one themselves? It might be better to implement this in a separate PR given the immediate issues with it.

Fox-McCloud:
Having come from the old days, I can tell you that the medals being an objective will be bad.

Department heads will be more leery to hand them out, and if someone has one, even for shiggles, then they'll demand security arrest them as an EoC rather than just petty theft.

Why do I say this? We used to have objectives to steal head's jumpsuits. If you had a jumpsuit with you, security would usually try to pin you as an EoC, and even if a head gave you the suit because you wanted to wear it for its aesthetics (I used to wear the RD suit even as an assistant), security would get upset.

TDSSS:
It would be cool if you could request a new objective if your old one was done as a function of a traitor uplink to make slime extract less silly, but that's for another PR

As for new objectives, maybe something like stealing the nuke core that other servers have? Or how about the captain's painting?

PR #12085, opened 2019-08-25
Adds a disk compartmentalizer to hydroponics

SteelSlayer:
**What does this PR do:**
Adds a disk compartmentalizer in hydroponics, which gives botanists and convenient place to store all their plant data disks. The life bringer ghost role area already has one of these so I figured why not give one to the station as well? I removed one of the high capacity water tanks, but there's already two of these in hydroponics anyway, so it shouldn't be a problem.

Before:
before

After:
after

**Changelog:**
:cl:
del: Removed one of the high capacity tanks in hydroponics
add: Added a disk compartmentalizer to hydroponics
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -32862,8 +32862,8 @@
 /obj/machinery/light{
 	dir = 1
 	},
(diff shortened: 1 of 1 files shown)


SomeSortaSquid:
Given how often I move that tank out of the corner, build a table and store the disks there? Quite good change.

PR #11978, opened 2019-08-10
Adds Sprites for Some Armalis+Vox Gear & Species-fitting Inhand Icons, etc.

KasparoVy:
**What does this PR do:**
I wanted to do some work on Armalis some years ago but forgot. #11975 reminded me so I fixed that and added some basic gear sprites for Armalis, species-fit earwear for Vox Primalis and fixed a couple minor issues in humanoid icon generation.

Resolves #11975

**Images of sprite/map changes (IF APPLICABLE):**
https://imgur.com/a/6UdGpuq

**Changelog:**
:cl:
imageadd: Adds sprites for all Vox & Armalis earwear.
imageadd: Adds sprites for Armalis default backpack, nitrogen tank (old-style back & new-style belt) & all breath mask variations.
add: You can now species-fit in-hand icons: Use a single .dmi file and suffix the icon states with _l or _r.
fix: Fixes an issue preventing Vox Armalis from being rendered with the proper in-hand icons for their massive noise cannons.
fix: Fixes an issue preventing the species-fitting of earwear.
fix: Fixes an issue preventing Vox Armalis from wearing the 'Vox' breath mask.
/:cl:

I've always wondered wtf was wrong with the noise-cannons all this time. Thankfully since the massive icon generation refactor it was a more straightforward fix.

PR #12087, opened 2019-08-26
Fixes Bug with Species-fitted Inhand Icons

KasparoVy:
Caused by myself in #11978, the bug presented where a species that was in the sprite_sheets of an item but didn't have a custom inhand sprite it picked up an item like a uniform and suddenly a grey jumpsuit was rendered on them. Happens for Vox when picking up regular uniforms because they have species-specific uniform sprites but not species-specific inhands, so it triggered the conditional and caused silly issues like rendering a default grey jumpsuit on them when they pick up a uniform.

Thought it was a BYOND sprite glitch at first but after reproducing it a couple times I figured I probably screwed something up.

**What does this PR do:**
Fixes a bug causing species with custom onmob sprites to have issues when they don't have custom inhand icons as well.

**Changelog:**
:cl:
fix: Picking up uniforms no longer renders a default grey jumpsuit on Vox.
/:cl:

**Ingame Screenshot:**
_The bug in action. Absolutely perfect timing as my character was blind and the overlay gives you a couple frames of vision once a minute._
para ss13 uniform inhands bug

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -87,6 +87,7 @@ var/global/image/fire_overlay = image("icon" = 'icons/goonstation/effects/fire.d
 	If index term exists and icon_override is not set, this sprite sheet will be used.
 	*/
 	var/list/sprite_sheets = null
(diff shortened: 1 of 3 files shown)


KasparoVy:
Screenshot pending

PR #12091, opened 2019-08-26
Reverts "Fixes Grid Check"

Fox-McCloud:
Yeah uh, this wasn't the proper way to do this. There is a way, but this isn't it--this is basically generating all power distribution on the grid, which invariably means the Tesla/Singulo is going to get released.

:cl: Fox McCloud
fix: Fixes random singularity releases from the grid check event
/:cl:


KasparoVy:
Our lord will be freed.

PR #12092, opened 2019-08-26
Loadout Updates

Fox-McCloud:
Adds a few more things to loadouts and tweaks costs.

For the most part, loadout costs have been cut--most things are only 1 point now, instead of 2; there's still some exceptions for exceptional items, but I really don't see much point in the majority of these suits, uniforms, and such costing 2 points.

Additions:
- Fingerless gloves
- Eyepatch
- Hipster glasses
- Prescription glasses
- Black shoes
- Brown shoes
- White Shoes
- Ian Shirt

:cl: Fox McCloud
tweak: loadout costs for most things cut from 2 to 1
add: Adds fingerless gloves, eyepatch, hipster glasses, prescription glasses, black shoes, brown shoes, white shoes, and ian's shirt to the loadout
/:cl:

### DIFF
--- a/code/modules/client/preference/loadout/loadout_donor.dm
+++ b/code/modules/client/preference/loadout/loadout_donor.dm
@@ -1,5 +1,4 @@
 /datum/gear/donor
-	donator_tier = 2
 	sort_category = "Donor"
 	subtype_path = /datum/gear/donor
 
@@ -18,12 +17,10 @@
 /datum/gear/donor/furcape
 	display_name = "Fur Cape"
 	path = /obj/item/clothing/suit/furcape
-	cost = 2
 
 /datum/gear/donor/furcoat
 	display_name = "Fur Coat"
 	path = /obj/item/clothing/suit/furcoat
-	cost = 2
 
 /datum/gear/donor/kamina
 	display_name = "Spiky Orange-tinted Shades"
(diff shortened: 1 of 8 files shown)


Ionward:
Hipster glasses are already in the patron loadout options, but aside from that I like this a lot.

KasparoVy:
Eyepatch FTW

code/modules/client/preference/loadout/loadout_donor.dm
```diff
@@ -1,5 +1,4 @@
 /datum/gear/donor
-	donator_tier = 2
```
Kyep: Um... isn't this going to make all donation items available to donors at a lower level than our patreon says they should be available at?

Fox-McCloud: whoops, didn't mean to remove that.

Fox-McCloud:
Updated it; made note that the hipster glasses are moving.

I absolutely do NOT want to make a habit of making donator things non-donator--the reason I wanted to move the hipster glasses, explicitly, is because they're a bit better at showing someone's eyes than the regular prescription glasses, and someone is bound to prefer that on their character---and not something that should really be locked behind donator status, IMO.

KasparoVy:
If it's a choice between the regular eyeglasses and the hipster ones I always end up picking the hipsters, regardless of species.

PR #12093, opened 2019-08-26
Adds Sandbags Reworks Security Barricades

Fox-McCloud:
Finally adds in sandbags and sandbag barriers.

Additionally refactors barricades and ties them into the object damaging system.

From TG.

Barricades have been refactored to tie into the object damaging system so things like hulks, mechs, animals, and what have you can all damage barricades---a side effect of this is that the security barricade has been dramatically reworked. It is no longer something that security can toggle on/off with an ID, but rather a permanent structure that has to be damaged to be destroyed.

Sandbags

Empty sandbags are made from cloth and can be filled with any type of sand, which makes them a stack of standbags.

Full sandbags can be deployed to construct a barrier; these are climbable and you can throw things over them, but they block direct movement; they're good for stopping small legion skulls, but everything else will attack and break them quite easily.

Miners start off with a stack of 5 full sandbags and a box with 7 empty sandbags

Security Barriers

Security barriers as a discrete object, on the map, are gone; instead they're replaced with security deployable grenades. These grenades have 3 modes: single, vertical, and horizontal. After the grenade is primed, it'll go off, then 3 security barriers (though it can do only one if you want it to); after a short delay they'll fully deploy and lock into place, becoming unpassable.

You'll find 4 of these grenades in the armory; more can be ordered from cargo

Other Changes

-Can Rip up bedhseets and gauze to make cloth

:cl: Fox McCloud
add: Adds in sandbags. Use them to deploy destructable barriers; miners start off with a few
del: Removes access based security barriers
add: Adds in security barrier grenade. When it goes off, it will deploy up to 3 barriers that will lock into place; should find 4 of these in the armory
add: can tear up bedsheet and gauze to get cloth--requires a sharp object
/:cl:


TDSSS:
I feel like there should be some kind of visual effect when the barrier grenade explodes. Maybe a small cloud or some light or _something_ instead of the barricade just appearing from thin air. Minor nitpick tho.

PR #12099, opened 2019-08-27
Fixes Jackets Not Keeping You Warm

Fox-McCloud:
Jackets are supposed to have some cold protection, but for whatever reason, this wasn't properly implemented when these were ported from TG.

:cl: Fox McCloud
fix: Fixes jackets not keeping you warm
/:cl:


MRgaty:
New way to kill drasks

PR #12108, opened 2019-08-27
Medbay/Science Remap

SteelSlayer:
**What does this PR do:**
This will remap some parts of medbay and science and the surrounding maint areas. You can see an image of the map here:
https://affectedarc07.co.uk/ParaMapGen/Images/119.png

**Summary of changes:**
My initial reasoning for wanting to remap medbay was that it just felt too cramped to move around in some of the areas, especially on high pop rounds. Everyone would constantly be rushing around, pushing you and interrupting your defib attempts etc. With this remap I tried to keep most of medbay and science the same. Thus, most parts of the departments look familiar, but areas are shuffled around a lot.

**Medbay changes:**
First, we have the medbay treatment area with the sleepers and body scanner, and the cryotube area. I dramatically increased the size of these two rooms. These rooms are where the majority of traffic in medbay happens, so I felt like giving them a lot of room to move around in would be nice. 

Now doctors can stand off to the side near the sleepers, and treat their patients without getting pushed all over the place by random people running around. This also makes these areas very ambulance friendly for the paramedic since there is so much more room. Same goes with the cryotube area.

Next, the decision to add a third surgery room. This all follows the same idea: we have more patients we need to treat due to high pop, and we have more doctors, but not enough space to treat them. Well I added an additional surgery room to handle the increased population.

Obviously virology has been shifted from the south side of medbay to the east side. The only other thing that has changed with virology is that I have slightly reduced the size of it, I didn't feel it needed to be so large, plus cutting down on its size gave room for the new maintenance tunnel that runs behind it.

**The new Maintenance between medbay and science:**
Something that sort of evolved as the remap was progressing was the maintenance tunnel that runs from the robotics hallway to medbay maintenance. Originally I had virology butted up against toxins and science, but it felt odd. 

So, there is a now a new maint here which I quite like how it turned out. It was awkward getting deep into medbay maint from the rest of the station, you had to take a long way around. I'm not sure how this will affect balance with security response time and such, we'll have to see.

**Science changes:**
Not a whole lot has been changed in science really, it was just a lot of shuffling rooms around and making subtle tweaks here and there. I suppose the biggest change would be the new location and reorganization of the experimentor room. It's moved down to the southern half of science, has a more horizontal feel, and contains two more scientist lockers.

**========================================**
I know some of these changes are probably controversial. I tried to gather feedback on the forums back when I started a month ago, but not a whole lot of people responded. You can see the forum post here if you're interested: https://nanotrasen.se/forum/topic/16553-medbay-and-science-remap/ As always, let me know what you think.

**Changelog:**
:cl:
tweak: Remaps medbay and science
/:cl:


pinatacolada:
Way too much empty space next to cryo, and the long hallway connecting to genetics feels weird and forced into existance. So does the maintenance next to genetics. maybe having it all be maintenance without the weird kinda maintenance kinda not airlock, and having the sci side hallway shortened could help.

1 OR having its own dedicated storage room while the other 2 share one disconnected from them is very ehh to me, Id be inclined to each OR having its own room as it is on live, or none having one but instead one larger room servicing all 3. If you swap the atmos closet with a storage room, and place another one to the left of the topmost OR I am pretty sure it all fits.
Regardless there probably should be one cleaner, one IV thing and one blood crate per OR, as it is on live.

Hey with that big ass space you could make a nice looking patient treating area with curtains and beds and IV drips, sort of like the OG medbay sleepers were inside small room things.

FreeStylaLT:
Now that it's officially been PRed and this seems like a final to-be-presented version, I have some comments.

img
These spaces, while bigger, are completely empty and bare. While _yes_ you can now find a spot to not be disturbed while you defib more easily, it comes with some issues:

- It doesn't quite fix the underlying issue, but addresses its symptoms. The area is _still_ going to be the main gathering point for everyone. It _still_ has the same chokepoint of a 3-wide entrance / exit that will be used by nearly everyone.

- More intricately, the sleepers and the scanner are laid out around the edges of the room, meaning that to reach the scanner / two of the sleepers (so, most of the devices), you'd have to go _across_ the entire room, likely getting in the way of anyone and everyone in the space. 

- While the ^ could be fixed by centralizing the sleepers / scanner in the space, it would then get in the way of anyone going to / from surgery / cryo in an annoying way, as it'd be literally in the way of the main path in / out of the space.

- The cryo space was expanded in consequence with the treatment center, but I doubt that it ever needed to be _this_ large.

- And, finally, as a slightly pedantic point, yet most relevant to the ~~normie~~ regular player masses is that empty, bare spaces look unappealing / ugly, most of the time.

As far as I can tell, the underlying, true issue is that the tiny treatment center connects almost everything in Medbay. It is meant to act both as a hallway _and_ the main primary treatment place with sleepers / scanners. This was a mistake on behalf of whoever decided to retardedly map it this way. What a jerk. The consequence of this mistake, however, is that in order to fix the crowding, you'd have to remap the entire block, essentially, and split it up to not be directly connected to one vital place. 

You can see, hopefully, why I was saying it'd probably have to be a big(ger) remap on Discord.

**MOVING ON**

Slightly irrelevant, but important for general map consistency:

- Surgery storage. Either all ORs should have one attached or there should be a larger, combined one. Keep in mind that alternative entrances / exits to surgery were made intentionally, both for antaggery and other purposes. I am not entirely sure whether they've proven themselves useful, so they may be fine to remove.

img

This space, the hallway itself. There is nothing there for anyone and all it does is connect Viro windows to an empty hallway. I'd suggest simply making Maint extend inward more, make it less of a hallway and more of an indent.

As for maint. I am not opposed to the idea of bringing back robo maint, but if you're remapping science / medbay, then addressing the Genetics connection to science being one long thin desolate hallway should be some sort of priority. You're remapping Science anyway.

SteelSlayer:
The number 1 most common complaint I've seen is that there is too much space in the sleeper and cryo area. I suppose I understand why people think they're ugly, I did go for a functionality over form thing here. The idea was to have an open floor plan for these high traffic areas, but perhaps it can be solved in a different way.

One OR having a storage room while the others don't is bad, agree.
I'm really not sure what to do with that hecking west science hallway, it's just sort of there now.

I'm probably can't solve these problems by just shuffling things around. What is probably going to happen is I'll just completely redo medbay and science so I can deal with these problems better. Plus @FreeStylaLT's map is going to conflict to hell with mine. It's obvious his map is better than my attempt here, and will likely get merged, so there's no point in having to go through all that right now.

I'll likely come back once his PR is merged and then attempt some sort of remap, if I do at all. But thanks for the feedback, I'll leave this up just to see what happens to it, and to gather any more feedback etc.

KorpoPapiez:
That new maintence thing makes server room way too easy to access

variableundefined:
Honestly. I'm neutral and leaning toward no on this, as I agree with the empty space problems pointed out above.

SteelSlayer:
This is a poor attempt. Will attempt a version 2 at some point

PR #12118, opened 2019-08-28
Add the Banana Touch and Mime Malaise wizard spells

datlo:
This PR adds two new spells to the Wizard Spellbook. Banana touch is more permanent compared to my previous PR with it, and this PR does not touch or remove the Cluwne Curse spell. 

Banana Touch : 1 spellpoint, 30 seconds cooldown

 "A spell popular at wizard birthday parties, this spell will put on a clown costume on the target, stun them with a loud HONK, and mutate them to make them more entertaining! Warning : Effects are permanent on non-wizards."

Mime Malaise : 1 spellpoint, 30 seconds cooldown

 "A spell popular with theater nerd wizards and contrarian pranksters, this spell will put on a mime costume on the target, stun them so that they may contemplate Art, and silence them. Warning : Effects are permanent on non-wizards."

Gameplay wise, the first spell will force the target into a nodrop clown outfit with shoes and mask and set their CLUMSY and COMIC block to true, and mark them as default block so they cant be removed with mutadone. 

This is pretty crippling for sec/crew who will not be able to use their stun weaponry and guns reliably, and who will be slowed down by the clown shoes they cant remove. And look ridiculously funny trying to get you.

The second spell similarly forces a mime outfit and add the MUTE mutation as default to the target. This will prevent them from communicating your position. Also, mime.

Both spells stun on hit for a fairly long time. If the target is a wizard, the items will not be cursed, and side effects wont be permanent.

This PR also slightly refactor pierrot's throat disease to avoid having redundant item defines.

:cl:
add: Added the Banana Touch and Mime Malaise spells to the Wizard spellbook.
/:cl:


datlo:
to clarify ; The outfits you get are cursed (cant be removed), however they are NOT the green cluwne outfit. they look like regular clowns and wont be valid like cluwnes are.

datlo:
I mean consider that this is a melee spell and that the wiz could have just desintegrated the target instead (albeit at a higher cost/cooldown). You could just use a stunprod/baton and get similar results.

SomeSortaSquid:
Perhaps consider setting clothes_req to 0?
It could be too strong, but it would allow the wizard to dress up as a clown/mime and hide among the people they cursed.
This could lead to amusing cases of people being afraid of the wizard's victims. However, if played too hard with the mime spell, it could make wizard borderline untraceable. If you take an ID and PDA of someone with default muteness and cursed mask away, confirming their identity will require some time-consuming tests.

Overall, clothes_req change is just a thing to consider. If it doesn't break the balance too much, it's a buff that would definitely be fun to play around with.

datlo:
I'm not really a fan of these spells being used by a stealth wizard at no cost, No Clothes can be bought if thats the meme you want to go for.

Issue #12141, opened 2019-08-30
Blood Drunk Eye causes you to randomly drop dead


MrMagolor:
**Issue Description**:

Using the Eye of A Blood Drunk Miner in active combat will cause you to randomly die despite being high above the HP threshold.
**What did you expect to happen**: 

To not instantly die at *yellow* HP.
**What happened instead**:

YOU DIED
**Why is this bad/What are the consequences**:

It makes the blood drunk eye not act like its description would indicate.
**Steps to reproduce the problem**:

1. Get miner eye
2. Walk up to ash drake
3. Start whamming it
4. Randomly die 
**When did the problem start happening**:

Presumably when crusher trophies were merged.
**Extra information**:

It might have to do with some sort of brute loss stacking eventually going over the limit?

Furthermore, I've had this happen with both a drake, and Hierophant.

Fox-McCloud:
I was afraid something like that would happen.

MrMagolor:
This could also have something to do with how newcrit handles instant death with calculating the total damage.

PR #12142, opened 2019-08-30
Fixes Blooddrunk Buff

Fox-McCloud:
As I suspected, TG's logic of handling the blood drunk buff wasn't adequate.

It should now function, as intended; a short, 1 second buff that makes you stun immune+reduces the damage you take by 90% in addition to allowing you to ignore slowdown.

Also fixes the logic with attacking a crusher trophy with your Kinetic Crusher; it'd properly equip the trophy, but make you drop your Crusher; now it just equips the trophy to the crusher, properly

fixes: https://github.com/ParadiseSS13/Paradise/issues/12141

:cl: Fox McCloud
fix: Fixes blood drunk buff
fix: Fixes using a crusher on a trophy to equip it causing you to drop the crusher
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -45,13 +45,6 @@
 	duration = 10
 	tick_interval = 0
 	alert_type = /obj/screen/alert/status_effect/blooddrunk
-	var/last_health = 0
-	var/last_bruteloss = 0
-	var/last_fireloss = 0
-	var/last_toxloss = 0
-	var/last_oxyloss = 0
-	var/last_cloneloss = 0
-	var/last_staminaloss = 0
 
 /obj/screen/alert/status_effect/blooddrunk
 	name = "Blood-Drunk"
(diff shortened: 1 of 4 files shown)


code/datums/status_effects/buffs.dm
```diff
@@ -61,108 +54,32 @@
...
+			H.dna.species.tox_mod *= 0.1
+			H.dna.species.oxy_mod *= 0.1
+			H.dna.species.clone_mod *= 0.1
+			H.dna.species.stamina_mod *= 0.1
 		add_attack_logs(owner, owner, "gained blood-drunk stun immunity")
 		var/status = CANSTUN | CANWEAKEN | CANPARALYSE | IGNORESLOWDOWN
 		owner.status_flags &= ~status
 		owner.playsound_local(get_turf(owner), 'sound/effects/singlebeat.ogg', 40, 1)
 
-/datum/status_effect/blooddrunk/tick() //multiply the effect of healing by 10
```
Fox-McCloud: So, this can go, because of how TG handled it--they multiply EVERYTHING but 10, then, when the buff ends, it divides everything by 10---we're just going to apply a mod that makes whatever resistances you have divide by 10, then, when the buff ends, get multiplied by 10, instead. 

Should get rid of any weirdness.

code/datums/status_effects/buffs.dm
```diff
@@ -61,108 +54,32 @@
...
-		owner.updatehealth()
-		last_health = owner.health
-		last_bruteloss = owner.getBruteLoss()
-		last_fireloss = owner.getFireLoss()
-		last_toxloss = owner.getToxLoss()
-		last_oxyloss = owner.getOxyLoss()
-		last_cloneloss = owner.getCloneLoss()
-		last_staminaloss = owner.getStaminaLoss()
+				BP.brute_mod *= 0.1
+				BP.burn_mod *= 0.1
```
Fox-McCloud: So the reason I didn't use the species burn and brute mods is because it'll make augments bodyparts insanely strong (and because I"m ensuring consistency with TG's behavior).

If it was *species* `brute_mod` of `*= 0.1` and you took 100 damage, it'd reduce that damage to 10---and if you had an augmented limb, it'd then reduce that damage further to 5.

Contrast this with using a *limb's* `brute_mod` of `*= 0.1`. The 100 damage will get reduced to 95 if the limb is augmented, then the further reduction reduces it down to 9.5.

The first method means you can shrug off any damage, under 50, while blood drunk, which is not the intention, at all, so I went with the second one.

KasparoVy:
Damn that's hardcore.

PR #12114, opened 2019-08-28
Fixes Bugs with Shining Eyes, Disembodied Heads and Admin-Reviving Beheaded Humanoids

KasparoVy:
**What does this PR do:**
Been noticing a weird issue where character eyes are darker than they should be under circumstances. I narrowed it down to the issue occurring only for characters with shining eyes. Bit more troubleshooting & comparison to Polaris code revealed eyeshine's render plane/layer needed to be updated.

Took the opportunity to do a bit of cleanup and fix another long-time issue preventing the rendering of appearance features (hair, eyes, etc.) on disembodied heads.

These fixes were salvaged from a really old ass branch I have on my fork and figured it was better they actually see the light of day.

Admin-reviving a decapitated body will now regrow the head with the ears organ and all the appearance features (hair, eyes, etc.) it used to have according to the mob's DNA.

**Images of sprite/map changes (IF APPLICABLE):**
_Before and after fixing the plane/layer_
para ss13 shining eyes colour fix

_Gotta roll some heads - sometimes your own_
para ss13 reheader fix

**Changelog:**
:cl:
fix: Resolves an issue causing shining eyes to be rendered darker than intended.
fix: Resolves an issue preventing appearance features (hair, eyes, etc.) from rendering on disembodied heads.
fix: Resolves an issue preventing Admin-revived decapitated bodies from having appearance features or ears.
/:cl:

PR #12147, opened 2019-08-30
Fix Incorrect Eye-Colour for Races Without Shiny Eyes

### LINKED ISSUE #12143: Eyes are Now Sometimes Weird Colors
@KasparoVy 

Players are apparently spawning in with incorrect/strange eye colors that are not what they set them to, on character setup.

Seems to have occurred ever since: https://github.com/ParadiseSS13/Paradise/pull/12114

KasparoVy:
Overlays don't carry over from organ get_icon() and have to be blended if we want to use them in that way - currently eyes are the only thing update_icons() doesn't have its own way of rendering so eyes are the only thing here that get blended AND overlayed

**What does this PR do:**
Reverts a change I made in #12114 because `update_icons.dm` requires the head organ get eyes blended onto it from `get_icon()` as it's the only thing `update_icons.dm` doesn't have its own method of rendering.

See a more detailed explanation in the linked issue - resolves #12143 

_Spawned in with proper eye colour on left - used C.M.A. Admin to become satanic on the right_
para ss13 fix the eyes I broke

**Changelog:**
:cl:
fix: Non-shining eyes render the proper colour again.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/organ_icon.dm
+++ b/code/modules/surgery/organs/organ_icon.dm
@@ -103,6 +103,7 @@ var/global/list/limb_icon_cache = list()
 	if(dna.species.has_organ["eyes"])
 		var/icon/eyes_icon = owner.get_eyecon()
 		if(eyes_icon)
+			mob_icon.Blend(eyes_icon, ICON_OVERLAY)
 			add_overlay(eyes_icon)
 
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Thank you for hopping on this so quickly.

KasparoVy:
Thanks for the detailed issue you logged, sorry about that

PR #12150, opened 2019-08-30
Add the command crew monitor

datlo:
This PR adds the command crew monitor to the Blueshield's locker, the Emergency Response Team Leader's backpack (all levels), and the Protolathe's medical designs (same cost and requirements as the standard crew monitor).

The command crew monitor functions just like the standard crew monitor, however it filters out anyone who isnt a Command member, a NT VIP, or an ERT/CC official. 

Somebody with the custom title 'Acting Captain' will show up on the command monitor as they should be a member of Command proper, unlike other acting command staff.

Deathsquad and undercover NT agents are not shown for obvious reasons. 

In the hands of the Blueshield, this will help focus them on their mission of protecting their charges without distraction from people that should be helped by medical and/or sec.

In the hands of the ERT leader, it lets them monitor the status of their team, and locate the members of Command they should be coordinating with.

Replacements can be printed at the protolathe if lost.

Using this instead of a standard crew monitor does have a downside : The filter is ID based. If a command member is stripped of their ID or if they alter their title to a custom job that isnt 'Acting Captain', they wont show up.

:cl:
add: Added the command crew monitor to Blueshield's locker and ERT Leader's backpack, which only shows the sensors of Command, NT VIPs, ERTs, and CC.
/:cl:

### DIFF
--- a/code/_globalvars/lists/misc.dm
+++ b/code/_globalvars/lists/misc.dm
@@ -58,4 +58,10 @@ GLOBAL_LIST_INIT(cooking_reagents, list(RECIPE_MICROWAVE = list(), RECIPE_OVEN =
 
 GLOBAL_LIST(station_level_space_turfs)
 
+GLOBAL_LIST_INIT(command_and_vips, list("Captain", "Acting Captain", "Head of Personnel", "Head of Security", "Chief Medical Officer", "Chief Engineer",
+						"Research Director", "Magistrate", "Nanotrasen Representative", "Blueshield", "Nanotrasen Navy Officer", "Special Operations Officer",
+						"Nanotrasen Navy Captain", "Nanotrasen Diplomat", "Emergency Response Team Leader", "Emergency Response Team Officer",
+						"Emergency Response Team Engineer","Emergency Response Team Medic", "Emergency Response Team Inquisitor",
+						"Emergency Response Team Janitor", "HONKsquad"))
+
 #define EGG_LAYING_MESSAGES list("lays an egg.", "squats down and croons.", "begins making a huge racket.", "begins clucking raucously.")
--- a/code/datums/cache/crew.dm
+++ b/code/datums/cache/crew.dm
@@ -1,4 +1,5 @@
 var/global/datum/repository/crew/crew_repository = new()
+var/global/datum/repository/crew/command/command_repository = new()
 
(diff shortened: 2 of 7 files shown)


JKnutson101:
If I remember correctly, the whole 'Acting' title isn't actually supposed to be included when you make someone an acting captain or head of staff. It's just supposed to be the job title, with the understanding that the position is acting.

Could be wrong though. Been a long while since I've dealt with being an acting head of staff.

datlo:
It is very common from my experience that the HOP/HOS sets themselves as 'Acting Captain' in order to make clear that they are not a CC appointed cap, which is why I decided to add it to the list. If the HOP set the acting Captain as 'Captain', theyll show up on the monitor anyway so its not really an issue.

pinatacolada:
Not opposed to this but consider instead of this having and "advanced crew monitor 3000 deluxe edition" printable from rnd that lets you filter out which crew members you want to see. Could be useful for brig phys wanting to see prisioners and sec only, medbay to see everyone except assistants, antags only their target, you name it

datlo:
It'd be nice but it'd also be a massive pain in the ass to code to be honest.

Quantum-M:
I disagree with this pr. Blueshields already have a tendency of getting the spare ID and the captain's spare headset, and with that access to many areas and tools, which with this crew monitor would make antags who go after command, either for assassination or object theft, more harder to evade the Blueshield (or security since it's able to be made in the protolathe.

If it's hard to tell who is the command member in the normal crew member, then the icons for those his jobs on the monitor can just be made more unique for it to stand out.

Fox-McCloud:
Not a fan. 

Command members are already highly favored on paradise and killing them, stealthily, is impossible; making it *even harder*? I'm going to have to disagree.

datlo:
I personally think it's fine for command members to be difficult to impossible to be killed without being noticed, forcing antags to be loud to do so. If admins dont, no hard feelings if you close the PR.

variableundefined:
Agree with Fox on this one. With a well-organized security team it will make killing any head that isn't an obvious idiot impossible.

datlo:
Since two out of three maints agree on it, I'll go ahead and close the PR

~~Blueshield mains are crying now, look at what you've done.~~

PR #12151, opened 2019-08-30
Add the paranormal scanner to Inquisitor ERT

datlo:
This PR adds the paranormal scanner to Red Inquisitors and the advanced paranormal scanner to Gamma Inquisitors.

The paranormal scanner when used on a human will scan them. After 20 (10 for advanced) seconds, it will reveal if the target is a Vampire, Vampire Thrall, Cultist, Devil, or Wizard.

The main purpose and use case of the scanner is as an admin response to an effective mindswap wizard, which currently has no existing counters as there is no possible test for wizards (other than knowing who was swapped or seeing the wiz cast shit, which removes the need for a test).

As it is slow to use and requires both you and the target to be immobile, it is probably still better to use holy water to test for vamps and cultists, which is intentional. This can still be used to test non-mindshielded command and make sure they are trustworthy, for example.

Currently the scanner uses the spectrometer/advanced spectrometer sprites. I will look into getting unique sprites for it if feedback is positive.

:cl:
add: Added the paranormal scanner to Red and Gamma Inquisitor ERT.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -856,3 +856,38 @@ proc/healthscan(mob/user, mob/living/M, mode = 1, upgraded = FALSE)
 		dat += "<font color='red'>Retinal misalignment detected.</font><BR>"
 
 	return dat
+
+/obj/item/paranormal_scanner
+	name = "paranormal scanner"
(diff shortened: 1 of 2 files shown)


AdvnetureDrake:
Although only an admeme round type at the moment, should Devils be included in the list of detected antags? They seem like the should show up on such a device.

datlo:
Will add devils.

datlo:
Closing since it seems admins dont think this is a good idea and this is an admin tool.

Issue #12148, opened 2019-08-30
Playtime-locked jobs in Occupation Preferences are very hard to read


moxian:
**Issue Description**:
The jobs that don't have playtime requirements met are shown in white font on light background.
![2019-08-30_11-27-33](https://user-images.githubusercontent.com/7831163/64017542-5740db80-cb19-11e9-961d-49e8f04f6785.png)

**What did you expect to happen**: 
Text having decent contrast
**What happened instead**:
Text is unreadable
**Why is this bad/What are the consequences**:
Gives bad impression of the game to new players.
Hard on the eyes, especially for people with poor eyesight.
**Steps to reproduce the problem**:
1) (optional) Create a new account if your has all jobs unlocked already
2) go to setup character -> Occupation preferences
3) be unable to read text

**When did the problem start happening**:
Probably since forever.
**Extra information**:
Karma-locked jobs are displayed in nice bold readable blood-red. Adapting it for the playtime-locked jobs is one of the solutions i'd be happy with (devising a separate contrrasting colors is another one, but probably more difficult?)

Kyep:
I have no objection to you doing a PR that changes the white text to a more readable color. I only did it that way originally because the previous method did too. In retrospect I should have changed the text colors while I was at it.

moxian:
I'd gladly do a PR, but unfortunately I cannot [easily] do so for legal reasons.
So I'll choose to interpret "you" in the comment above as collective "you, the people who might be reading this message".

PR #12154, opened 2019-08-30
Changes the colour of playtime-required jobs

Shadow-Quill:
**What does this PR do:**
Fixes: #12148
Makes jobs which you do not have access to due to insufficient playtime red, instead of white.

image

(**Note**: The `none as ...` thing shouldn't happen normally. I just didn't bother to setup a DB or anything, and I just set the check to TRUE to test it.)

**Changelog:**
*Remove this line, and appropriate fields from the changelog*
:cl:
tweak: Jobs which you have insufficient playtime for are now red.
/:cl:

### DIFF
--- a/code/modules/client/preference/preferences.dm
+++ b/code/modules/client/preference/preferences.dm
@@ -642,7 +642,7 @@ var/global/list/special_role_times = list( //minimum age (in days) for accounts 
 			continue
 		var/available_in_playtime = job.available_in_playtime(user.client)
 		if(available_in_playtime)
-			HTML += "<del>[rank]</del></td><td> \[ " + get_exp_format(available_in_playtime) + " as " + job.get_exp_req_type()  + " \]</td></tr>"
+			HTML += "<del><font color=red>[rank]</font></del></td><td><font color=red> \[ " + get_exp_format(available_in_playtime) + " as " + job.get_exp_req_type()  + " \]</font></td></tr>"
 			continue
 		if(job.barred_by_disability(user.client))
 			HTML += "<del>[rank]</del></td><td> \[ DISABILITY \]</td></tr>"


Kyep:
Maybe they should all be in black, rather than red.
The red looks... I dunno, really aggressive.
Also, in this screenshot, they have a white strikethrough, which does not show up well.
How does black, with a black strikethrough, look?

code/modules/client/preference/preferences.dm
```diff
@@ -642,7 +642,7 @@ var/global/list/special_role_times = list( //minimum age (in days) for accounts
 			continue
 		var/available_in_playtime = job.available_in_playtime(user.client)
 		if(available_in_playtime)
-			HTML += "<del>[rank]</del></td><td> \[ " + get_exp_format(available_in_playtime) + " as " + job.get_exp_req_type()  + " \]</td></tr>"
+			HTML += "<del><font color=red>[rank]</font></del></td><td><font color=red> \[ " + get_exp_format(available_in_playtime) + " as " + job.get_exp_req_type()  + " \]</font></td></tr>"
```
Kyep: This should probably use a css class, if anything, not "font color=red". 
Also, it should probably get rid of the "<del>", because that creates a white strikethrough which is still hard to read.

Perhaps simple black text with no (or black) strikethrough would be best.

Shadow-Quill: And how would I do that?

Kyep: Well, start by removing the del tags.
Then, see how it looks with no font color tags. Is the default text black? That might be okay.
If not, find the class in that HTML which sets the default color to white, and undo that so the default color is black.

variableundefined:
Can I have an ETA on when this can be updated.
@Shadow-Quill

Shadow-Quill:
Whoops, forgot about this. I'll work on it later tonight.

Shadow-Quill:
Yeah, I don't have a clue.

Closing this.

code/modules/client/preference/preferences.dm
KasparoVy: `HTML += "<span class="bad">[rank]</span></td><td><span class="bad"> \[ " + get_exp_format(available_in_playtime) + " as " + job.get_exp_req_type()  + " \]</span></td></tr>"`

The "bad" class from [this file](https://github.com/ParadiseSS13/Paradise/blob/add5787179fe220b0eddc4bd26099eb5a6761031/goon/browserassets/css/browserOutput.css) is red (#FF0000) with no other formatting.

Not sure what `<del>` does though.

Shadow-Quill: `<del>`does ~~strikethroughs~~

KasparoVy: Ohhh, I see. I looked to see if there was a more suitable span class but it actually seems that in this case we may have to create a new one in [this file](https://github.com/ParadiseSS13/Paradise/blob/add5787179fe220b0eddc4bd26099eb5a6761031/goon/browserassets/css/browserOutput.css)

Could do it like this:

`.not_available {color: #FF0000; text-decoration: line-through;}`
or however the spacing needs to be. I referred to documentation [here](https://www.electrictoolbox.com/css-strikethrough/) for the CSS information.

PR #12155, opened 2019-08-30
Mice are no longer zombies

Couls:
**What does this PR do:**
Mice were coming back to life due to some interesting code in the mouse file.

**Changelog:**
:cl:
fix: mice will no longer resurrect from death randomly
/:cl:


Couls:
Fix for this is in #12157
(Fox is a nerd)

Issue #12163, opened 2019-08-31
Blood lags the game out hard


farie82:
**Issue Description**:
Blood stacks on top of each other. Creating lag piles as a cultist

**What did you expect to happen**: 
Blood not to make your PC crawl to a halt.

**What happened instead**:
In a fight one of the crewmembers reached a whopping 2 FPS (probably a bit more but it was horrid)

**Why is this bad/What are the consequences**:
SS13 should be playable for everybody not just the elite with their 5Ghz CPUs

**Steps to reproduce the problem**:
Blood. Blood blood!!!!!!!!!
No but really just stack multiple blood paths on top of each other. We had like 50+ blood effects on one tile. In the drug base and it was completely full with those tiles

**When did the problem start happening**:
I think since #11797

Couls:
Can't reproduce this on my test server, how did you manage this?

farie82:
might have to do with cult spawned blood. It was in a base which was tightly packed. Lots of fresh runes and teleporting (which damages you since you can bleed from it)

farie82:
Still an issue that breaks the game

PR #12167, opened 2019-09-01
Fix blood stacking

Couls:
**What does this PR do:**
after some tests and looking through comments in the code it seems that INITIALIZE_HINT_QDEL doesn't work as intended. I can't figure out how to fix it myself but for the time being this is the only way to reliably qdelete items during initialization until the initialize hint can be fixed

fixes: #12163 

**Changelog:**
:cl:
fix: blood no longer stacks on itself endlessly
/:cl:


Fox-McCloud:
I'd rather fix the base level problem: https://github.com/ParadiseSS13/Paradise/pull/12170

Once this is tested and confirmed to resolve the issue, then this is not necessary.

Fox-McCloud:
Merging this, because it's causing quite a problem and the initializations PR is a bigger nightmare than I anticipated.

PR #12169, opened 2019-09-01
Lavaland Syndiebase Fixes & Tweaks

Kyep:
:cl: Kyep
fix: Fixed a bug that made lavaland syndie scientist and comms agent headsets fail to spawn on their ears. Removed the temporary workaround for this bug (extra headsets spawning on the tables in their dorm rooms).
tweak: Lavaland syndie scientists can no longer contact the station (only comms agent headsets can speak on NT common, only comms agents get the monitor key for the PDA messages console, etc). Previously, syndie scientists would often either try to take the comms agent's job, or just out themselves to the Cyberiad by speaking on ";" radio wearing their syndie IDs.
/:cl:


Markolie:
The Lavaland Syndie comms agent spawning all the time is absolutely not a bug. If you have a look at the type path of the specific spawner that implements that, it’s the /space/ one. That one’s used by /tg/ on the space listening outpost. We currently do not use that spawner. 

As mentioned on the other pull request, I’m also fully against disallowing the comms agent to spawn. With the new ruins, the Syndicate base spawning is uncommon and the comms agent can have interesting interaction with the crew nonetheless. 

The scientist changes are fine.

VEN0M5:
They still can contact the station if they print the normal headset or radio from the autolathe

datlo:
I feel like 10 percent is too low of a spawn rate considering the ruin is not guaranteed to spawn in the first place. add to that that traitors being on station is not a guarantee either.

Kyep:
D'oh. Looks like the paths are indeed different. The lavabase uses /obj/effect/mob_spawn/human/lavaland_syndicate/comms and only /obj/effect/mob_spawn/human/lavaland_syndicate/comms/space has the 10% spawn rate.

Now I have to figure out if I want to:
- remove the comms agent
- apply the 10% spawn chance to the comms agent as a non-fix change
- leave the comms agent as-is
Not sure which I want to do yet.

I'm okay, I think, with scientists being able to print normal radios from an autolathe. If they do that, they're breaking role (they are obligated to keep the base secret, and using NT radio breaks that). The main thing with the scientist is that say ";hi" should not break their role and give away the base's existence, nor should they easily fall into the pattern of trolling the crew all round.

Kyep:
Update:
- Reverted change that made comms agent only spawn 10% of the time.
- Kept changes that prevented the scientists from accidentally outing themselves, or stealing the comms agent's job.

AffectedArc07:
Ignore that MapDiffBot error, my DB server shat itself again

PR #12002, opened 2019-08-14
False wall secure fix, add extra step to plating deconstruction

Markolie:
**What does this PR do:**
Fixes https://github.com/ParadiseSS13/Paradise/issues/12001.
Fixes https://github.com/ParadiseSS13/Paradise/issues/11999.

**Changelog:**
:cl: Markolie
fix: Tightening the bolts of a falsewall now works properly.
tweak: Deconstructing plating now requires it to be unfastened using a screwdriver first, in order to prevent it from being accidentally deconstructed with a welder.
/:cl:

PR #12171, opened 2019-09-01
Welding a tile away now gives a new message

farie82:
**What does this PR do:**
Welding a tile now gives a more clear message to the user

**Changelog:**
:cl:
tweak: Welding a tile away now gives a more useful message when doing so.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/floor/plating.dm
+++ b/code/game/turfs/simulated/floor/plating.dm
@@ -97,7 +97,7 @@
 				broken = FALSE
 				update_icon()
 			if(unfastened)
-				user.visible_message("<span class='danger'>[user] starts removing [src]!</span>", "<span class='warning'>You start removing [src] exposing space after you're done!</span>")
+				to_chat(user, "<span class='warning'>You start removing [src] exposing space after you're done!</span>")
 				playsound(src, welder.usesound, 100, 1)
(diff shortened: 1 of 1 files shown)


JohnBigless:
Gone would be the days of accidentally breaking plating with welders during blob rounds.
And gone would be the possibility of stealthily removing floor tiles under someone and watching them get annihilated by Fastmos.
+1

Lunaimon:
Don't you already need to use a screwdriver before the welder to remove plating? #12002

Citinited:
Agreed with Lunaimon -- this seems unnecessary given #12002 is in

farie82:
It still confused me after I tried to deconstruct something. Since it was a notice and not a warning that i was about to space myself

variableundefined:
I think the screwdriver should be enough. Just change the span class of the tile removal?

farie82:
Would be nice to get a direction in how we do dangerous deconstructions. As some use the harm which means you're about to do something harmful or potentially dangerous.

I myself think this is a good thing to have. You can also prime floorings now and have people accidently hit them

Fox-McCloud:
yeah, just add a warning/danger class

variableundefined:
Changelog no longer accurate. I think it's better to just stick with to_chat for now.

@farie82

farie82:
Updated to just be a changed to_chat

Issue #12176, opened 2019-09-01
Can't make fruit wine


variableundefined:
**Issue Description**:

You cannot make fruit wine

**What did you expect to happen**: 

Make fruit wine

**What happened instead**:

You can't

**Why is this bad/What are the consequences**:

**Steps to reproduce the problem**:

**When did the problem start happening**:

**Extra information**:

To be fixed. List out of bound thing.

Issue #12175, opened 2019-09-01
Borgi causes run time while tasting food


variableundefined:
**Issue Description**:

Borgi causes runtime!

**What did you expect to happen**: 

No runtime

**What happened instead**:

**Why is this bad/What are the consequences**:

**Steps to reproduce the problem**:

Eat. As. Borgi.

**When did the problem start happening**:

**Extra information**:

Self reminder to fix it later.

PR #12177, opened 2019-09-01
Fixes Some Null Tastes and Fruit Wine

Fox-McCloud:
Big thanks to @Crazylemon64 for helping me pin down the problem; I was starting to suspect reagents were involved, but I wasn't completely sure, and I doubted myself enough that I ruled it out. 

Either case, this fixes fruit wine causing runtimes and not actually fermenting.

This should also fix some instances of "you taste sweetness, ,and ". For a few null taste reagents. If we want something to not have a taste, it shouldn't be `null`, it should just have a `taste_multiplier` of `0`.

Other:
Implemented some of TG's code for taste (not really functionally different).

fixes: https://github.com/ParadiseSS13/Paradise/issues/12175
fixes: https://github.com/ParadiseSS13/Paradise/issues/12176

:cl: Fox McCloud
fix: Fixes some blank tastes
fix: Fixes not being able to brew fruit wine
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -20,6 +20,9 @@
 
 /obj/item/reagent_containers/food/snacks/grown/New(newloc, var/obj/item/seeds/new_seed = null)
 	..()
+	if(!tastes)
+		tastes = list("[name]" = 1)
+
 	if(new_seed)
 		seed = new_seed.Copy()
 	else if(ispath(seed))
@@ -68,7 +71,7 @@
 
 			if(!isturf(loc) || !(locate(/obj/structure/table) in loc) && !(locate(/obj/machinery/optable) in loc) && !(locate(/obj/item/storage/bag/tray) in loc))
 				to_chat(user, "<span class='warning'>You cannot slice [src] here! You need a table or at least a tray to do it.</span>")
-				return TRUE	
+				return TRUE
 
 			var/slices_lost = 0
 			if(!inaccurate)
--- a/code/modules/hydroponics/seeds.dm
+++ b/code/modules/hydroponics/seeds.dm
@@ -160,23 +160,23 @@
 	return result
 
 
-/obj/item/seeds/proc/prepare_result(var/obj/item/T)
+/obj/item/seeds/proc/prepare_result(obj/item/T)
 	if(!T.reagents)
 		CRASH("[T] has no reagents.")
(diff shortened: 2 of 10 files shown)


code/modules/hydroponics/grown.dm
```diff
@@ -20,6 +20,9 @@
 
 /obj/item/reagent_containers/food/snacks/grown/New(newloc, var/obj/item/seeds/new_seed = null)
 	..()
+	if(!tastes)
+		tastes = list("[name]" = 1)
```
Fox-McCloud: This is part of the fix....but...

code/modules/hydroponics/seeds.dm
```diff
@@ -160,23 +160,23 @@
...
 			data = list("blood_type" = "O-")
-			continue
-		var/list/nutritious_reagents = list("nutriment", "vitamin", "protein", "plantmatter")
-		if(reagent_id in nutritious_reagents)
+		if(rid == "nutriment" || rid == "vitamin" || rid == "protein" || rid == "plantmatter")
+			// apple tastes of apple.
 			if(istype(T, /obj/item/reagent_containers/food/snacks/grown))
 				var/obj/item/reagent_containers/food/snacks/grown/grown_edible = T
-				data = grown_edible.tastes
+				data = grown_edible.tastes.Copy()
```
Fox-McCloud: This is the real fix.

Basically, TG doesn't `Cut` the `data` list of reagents when reagents are destroyed, mostly because they don't have anything too crazy like Stable Mutagen; we do though, so we want to clear out those references.....but it had implications.

Really, TG should do this, too---only reason they're getting away with it is because their `data` isn't cut; that said, I suspect it's probably creating hard deletions for some reagents for them.

code/modules/reagents/chemistry/reagents.dm
```diff
@@ -27,7 +27,7 @@
 	var/drink_name = "Glass of ..what?"
 	var/drink_desc = "You can't really tell what this is."
 	var/taste_mult = 1 //how easy it is to taste - the more the easier
-	var/taste_description = "bitterness" //life's bitter by default. Cool points for using a span class for when you're tasting <span class='userdanger'>LIQUID FUCKING DEATH</span>
+	var/taste_description = "metaphorical salt"
```
Fox-McCloud: You should never see this.

PR #12184, opened 2019-09-02
Removes headpat gloves

Ty-Omaha:
Some admins are lagging from this and the logs are being spammed

We really don't need this much spam in game either way

🆑 
del: Headpat gloves
/ 🆑

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -152,25 +152,10 @@
 /obj/item/clothing/gloves/fingerless/rapid
 	name = "Gloves of the North Star"
 	desc = "Just looking at these fills you with an urge to beat the shit out of people."
-	var/accepted_intents = list(INTENT_HARM)
-	var/click_speed_modifier = CLICK_CD_RAPID
 
 /obj/item/clothing/gloves/fingerless/rapid/Touch(mob/living/target, proximity = TRUE)
 	var/mob/living/M = loc
 
-	if(M.a_intent in accepted_intents)
-		M.changeNext_move(click_speed_modifier)
+	if(M.a_intent == INTENT_HARM)
+		M.changeNext_move(CLICK_CD_RAPID)
 	.= FALSE
-
-/obj/item/clothing/gloves/fingerless/rapid/admin
-	name = "Advanced Interactive Gloves"
-	desc = "The gloves are covered in indecipherable buttons and dials, your mind warps my merely looking at them."
-	accepted_intents = list(INTENT_HELP, INTENT_DISARM, INTENT_GRAB, INTENT_HARM)
-	click_speed_modifier = 0
-	siemens_coefficient = 0
-
-/obj/item/clothing/gloves/fingerless/rapid/headpat
-	name = "Gloves of Headpats"
-	desc = "You feel the irresistable urge to give headpats by merely glimpsing these."
-	accepted_intents = list(INTENT_HELP)
-	click_speed_modifier = 0 // That's some serious headpatting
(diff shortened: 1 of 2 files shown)


Drakeven:
We hardly knew ye...

AffectedArc07:
Or just make them not admin log, or increase the cooldown


***imPrOve donT RemOVE***

Ty-Omaha:
@AffectedArc07 
lots of things get removed instead of improved

Lunaimon:
How did these get in lmao

Ty-Omaha:
@AffectedArc07 updated to keep the refactor

code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -161,16 +161,3 @@
 	if(M.a_intent in accepted_intents)
 		M.changeNext_move(click_speed_modifier)
 	.= FALSE
-
-/obj/item/clothing/gloves/fingerless/rapid/admin
```
AffectedArc07: You sure you dont want to keep these as well?

FreeStylaLT:
At first I thought we should just OOCly remind spammers not to spam excessively.
Then I remembered that spam is the entire purpose of these items. They probably shouldn't be publicly accessible like that, at least.

code/modules/clothing/gloves/miscellaneous.dm
Ty-Omaha: eh sure

Qwertytoforty:
Honestly, gloves are annoying, probably could be abused by shaking people up to remove stuns fast, so it is better to remove them or keep them admin only.

Spacemanspark:
I'd be fine with making it admin only.

Ty-Omaha:
@Spacemanspark the admin only version is kept

datlo:
just remove the gloves from the vendor instead

Fox-McCloud:
There's several different ways this could be improved; kneejerk removing it when it's very modular is silly.

Instead of out and out removing it, you could:

- Make the cooldown higher
- Have it print a message to people in a small radius instead of it being a visible message
- Move it from the vendor (where people are exploiting the heck out of decon and rebuilt vending machines to get more) to somewhere else harder to get (like an arcade thing)
- Fix logging on help intent so it doesn't show up for everything
- Any combination of the above.

This isn't some snowflake code that's hard to modify; I had this this way so it was *easy* to modify.

Ty-Omaha:
>Make the cooldown higher

What's the point of it then? Slightly faster hugs?

>Have it print a message to people in a small radius instead of it being a visible message

Those people in a small radius will still get spammed

>Move it from the vendor (where people are exploiting the heck out of decon and rebuilt vending machines to get more) to somewhere else harder to get (like an arcade thing)

I don't see the purpose of having an item just to circumvent spam protection, arcades can be exploited and even if someone wins it then we are at the same issue of having spam

>Fix logging on help intent so it doesn't show up for everything

As above



What value does this add to the game other than spamming hugs?

Fox-McCloud:
>What's the point of it then? Slightly faster hugs?

What's the point of gloves of the north star? Slightly faster punching? 

There's a world of difference between "reduced cooldown" and "literally none at all; as fast as you can click"

>Those people in a small radius will still get spammed

The vast bulk of the complaint as about logging and causing it to be shown to *everyone* on the screen.

>arcades can be exploited and even if someone wins it then we are at the same issue of having spam

The problem has been 3-4+ people having them and all using them, not to mention the *literally non-existent click cooldown*.

Ty-Omaha:
>What's the point of gloves of the north star? Slightly faster punching?

Yes, to give antags another way of killing people.

variableundefined:
Closing in favour of fixing the cooldown by using headpat's parent's cooldown.

PR #12187, opened 2019-09-02
Fixes Fore Port Solars APC doesn't work

### LINKED ISSUE #11752: Fore Port Solar APC doesn't work
Lt. Diggs was able to find the issue with the fore port solar APC.  It seems that the tile that the APC is on is not technically part of the room and is in space or something which is why it can't charge or anything.  It also seems that the rest of the room's reinforced walls aren't part of the room either.  Since the APC doesn't work neither does the solar tracker for that solar either.

### LINKED ISSUE #12185: Fore Port Solar Array APC
**Issue Description**:
The Fore Port Solar Array is not connected to the Area and will not take External Power. It stays stuck at 90% on the monitor computer and directly.

This issue really is a map oversight. The Area is not fully enclosed so therefore, the Area is basically broken. (See image below)

image

This is how is should be:
image

Even after removing the APC from the wall. You cannot put it back on the wall. We tried using the blueprint to make it work, but it doesn't work either.

The APC to work

**What happened instead**:
APC no work

SMES doesn't distribute power to the station. So therefore only 3/4 of the Solar Arrays throughout the station are running at any given time. The only solution would be to directly power the solar to the main line.

N/A This is a mapping problem.

I don't know. Recent Map merges in that area?

**Extra information**:
I have video clips, if you need additional proof.

Shadow-Quill:
**What does this PR do:**
changes the area to fore port solar _control_

im just as confused as you are

fixes #11752
fixes #12185

**Changelog:**
*Do not remove this line, but do remove appropriate fields from the changelog*
:cl:
fix: Fixed Fore Port Solars' APC not charging.
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -15799,7 +15799,7 @@
 "aAd" = (
 /obj/effect/spawner/window/reinforced,
 /turf/simulated/floor/plating,
-/area/solar/auxport)
+/area/maintenance/auxsolarport)
 "aAe" = (
 /turf/space,
 /turf/simulated/shuttle/wall{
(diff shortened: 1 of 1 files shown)


Shadow-Quill:
mfw mapdiffbot fails

AffectedArc07:
MDB is failing GitHub wide due to a DB table being full

PR #12189, opened 2019-09-02
Increases spam hug glove cooldown

Ty-Omaha:
**What does this PR do:**
Increases the cooldown of the headpat gloves because it was being spammed:

Alternative to: #12184

**Changelog:**
:cl:
tweak: Headpat gloves now have an increased cooldown due to spam
/ 🆑

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -173,4 +173,4 @@
 	name = "Gloves of Headpats"
 	desc = "You feel the irresistable urge to give headpats by merely glimpsing these."
 	accepted_intents = list(INTENT_HELP)
-	click_speed_modifier = 0 // That's some serious headpatting
+	click_speed_modifier = 6 // That's some serious headpatting


code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -173,4 +173,4 @@
 	name = "Gloves of Headpats"
 	desc = "You feel the irresistable urge to give headpats by merely glimpsing these."
 	accepted_intents = list(INTENT_HELP)
-	click_speed_modifier = 0 // That's some serious headpatting
\ No newline at end of file
+	click_speed_modifier = 6 // That's some serious headpatting
```
Fox-McCloud: just get rid of this line and let inheritance handle what the modifier is for the original gloves of the north star

Mitchs98:
Would definitely prefer this to their removal. I haven't even got to use them yet ;-;

code/modules/clothing/gloves/miscellaneous.dm
Ty-Omaha: thats still too fast

Fox-McCloud: There's virtually no point to it being 6.

You can make a *gun* click faster than this.

florkeger:
> Would definitely prefer this to their removal. I haven't even got to use them yet ;-;

Medium roleplay??? I didn't use them yet, but I don't think they have much of a place in my opinion. But this definitely looks too fast.

Mitchs98:
Medium roleplay allows for memeing to reasonable degrees. You can't just passively quip 'medium roleplay' at everything. They fit just as much as anything else.

florkeger:
> Medium roleplay allows for memeing to reasonable degrees. You can't just passively quip 'medium roleplay' at everything. They fit just as much as anything else.
I mean if you say so, I still don't like the inclusion. Too memey for my taste.

Quantum-M:
I agree with @florkeger, besides being a memey thing, there isn't much reason to have those gloves, you already can pat people with the reasonable one second delay without these gloves, without spamming the chat severely.

I think they are better removed or just left as a admin only item.

code/modules/clothing/gloves/miscellaneous.dm
variableundefined: Agreed with above

Fox-McCloud: Might I again, also suggest, this be an arcade reward? Not horrifyingly expensive, but enough that it's not something that shows up 5 minutes in, every shift, and peopel can make hundreds of them.

farie82:
Unsure if you can do this. But how about changing the message to 1 instead of a choice of 3-5 (dunno how many there are). That way the spam filter on your chat also reduces the effect

Fox-McCloud:
Tick dock, this is a 5 minute, tops, fix. 

Don't let this hang forever.

variableundefined:
Required changes implemented on #12218. Closing

PR #12190, opened 2019-09-02
Moves the admin "Shake" log to the actual waking up people part of the code

farie82:
**What does this PR do:**
Help intent now only logs when somebody is really helped up from a stun or sleep effect. Instead of the full help intent mode.

**Changelog:**
:cl:
tweak: Admins now don't get spammed by hugs.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -213,7 +213,6 @@
 		swap_hand()
 
 /mob/living/carbon/proc/help_shake_act(mob/living/carbon/M)
-	add_attack_logs(M, src, "Shaked", ATKLOG_ALL)
 	if(health >= HEALTH_THRESHOLD_CRIT)
 		if(src == M && ishuman(src))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -213,7 +213,6 @@
 		swap_hand()
 
 /mob/living/carbon/proc/help_shake_act(mob/living/carbon/M)
-	add_attack_logs(M, src, "Shaked", ATKLOG_ALL)
```
Fox-McCloud: I suspect we inherited this from Bay---they have a proclivity for logging just about every single interaction just to make sure "dat ARRR PEE WAS JOOSTEHFEYED!"

farie82: BWOINK!! WHY DID YOU HEADPAT THAT CIVILIAN!?

AffectedArc07: inb4 the server runs out of storage space

PR #12199, opened 2019-09-04
Changes behavior of the malf AI's overload/override machine abilities

SteelSlayer:
**What does this PR do:**
Edit: Alright, upon Fox's request, I scrapped all that spaghetti shitcode and just implemented TG's version of overload/override machine. Now, when you you click either of these ability's action buttons, your mouse pointer icon will transform to a different icon. The next click you perform on a machine will overload or override that machine. You can cancel it's use by clicking on the action button again 

Ability demo video:
https://streamable.com/jinhx

**Changelog:**
:cl:
add: Malf AI's overload machine & override machine action button behavior is changed to a point and click style ability.
imageadd: Added some new mouse pointer icons for the machine overload and machine override abilities
/:cl:


code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -36,6 +36,10 @@
 		if(initial(uses) > 1) //no need to tell 'em if it was one-use anyway!
 			to_chat(owner, "<span class='warning'>[name] has run out of uses!</span>")
 		qdel(src)
+		if(istype(src, /datum/action/innate/ai/overload_machine))
```
variableundefined: Never do this. OOP Principles means that object behavior should be attached to the object.

This means that instead of every action checking whether it is overload machine, it should be overload machine doing this adjustment instead.

Now I'm not too sure if this is necessary as I don't see this system being used to add actions in the old PR. Cannot comment on that.

code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -36,6 +36,10 @@
 		if(initial(uses) > 1) //no need to tell 'em if it was one-use anyway!
 			to_chat(owner, "<span class='warning'>[name] has run out of uses!</span>")
 		qdel(src)
+		if(istype(src, /datum/action/innate/ai/overload_machine))
+			usr.verbs -= /mob/living/silicon/ai/proc/overload_machine
+		else if(istype(src, /datum/action/innate/ai/override_machine))
```
variableundefined: Ditto

code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -105,6 +109,8 @@
 		for(var/datum/action/A in AI.actions)
 			if(istype(A, initial(AM.power_type)))
 				qdel(A)
+	AI.verbs -= /mob/living/silicon/ai/proc/overload_machine
```
variableundefined: Ditto

code/game/gamemodes/malfunction/Malf_Modules.dm
SteelSlayer: Yeah this is something I added actually. I felt like it didn't make sense for the action button to be removed when amount of charges reaches 0 but then have the verb not get removed.

I see what you mean about having the overload/override procs handle this.

SteelSlayer:
If I understand what you mean, that's how it currently works. Right now when you add either of these powers, you get an old-style verb in the malfunction tab. If you click on those verbs in the tab, you get the giant list of all machines. But you can also right click on machines, and get the verb in the right click menu

code/game/gamemodes/malfunction/Malf_Modules.dm
SteelSlayer: I already had deletion code in the proc itself for when it ran out of charges, what was I thinking..

Fox-McCloud:
I'm pretty sure TG handled this by having malf AI use `/obj/effect/proc_holder/ranged_ai/overload_machine`; you may want to look into their latest implementation of AI abilities.

SteelSlayer:
Sounds good, I'll check it out

SteelSlayer:
@Fox-McCloud @variableundefined I did what Fox suggested and just implemented the point and click style system TG uses for these abilities

code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -77,13 +77,13 @@
 	var/disable_text = "<span class='danger'>Goodbye Cruel World!</span>" //Context clues!
 	var/datum/action/innate/ai/ranged/attached_action
 
-/obj/effect/proc_holder/ranged_ai/Destroy()
-	QDEL_NULL(attached_action)
-	return ..()
+// /obj/effect/proc_holder/ranged_ai/Destroy()
```
SteelSlayer: I'm not sure if this is necessary. This proc kept giving me runtimes about multiple qdels on the same object. When I commented them out, the runtimes went away. All the action buttons function normally and still get removed when they run out of charges so I don't think this is affecting anything.

code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -349,7 +350,7 @@
 	for(var/obj/machinery/door/D in GLOB.airlocks)
 		if(!is_station_level(D.z))
 			continue
-		INVOKE_ASYNC(D, /obj/machinery/door.proc/hostile_lockdown, src)
+		INVOKE_ASYNC(D, /obj/machinery/door.proc/hostile_lockdown, owner)
```
SteelSlayer: Having src instead of owner here gave me runtimes on every single door that ability affected. I didn't touch this ability so these runtimes might be present on live right now

code/game/gamemodes/malfunction/Malf_Modules.dm
variableundefined: Confirmed runtime also exists on the original TG version

PR #11863, opened 2019-07-15
Middle Mouse button now toggles airlock safeties for AIs.

Emanthealmighty:
**What does this PR do:**
The AI can now quickly toggle safeties instead of toggling door bolt lights.

**Changelog:**
:cl: EmanTheAlmighty
tweak: The AI can now toggle airlock safeties instead of toggling door bolt lights by pressing middle mouse.
/:cl:

PR #12207, opened 2019-09-04
Revert "Middle Mouse button now toggles airlock safeties for AIs."

Spacemanspark:
Reverts ParadiseSS13/Paradise#11863

Having played around with this for a few rounds... not digging it even remotely. Had this been attached to a different combination, maybe, but even then I and a good few others might argue that door safeties probably shouldn't be a macro.

As is, door bolt lights see much more use than the safeties. It makes a lot more sense to be macro'd where it is.

### DIFF
--- a/code/_onclick/ai.dm
+++ b/code/_onclick/ai.dm
@@ -209,13 +209,11 @@
 /atom/proc/AIMiddleClick()
 	return
 
-/obj/machinery/door/airlock/AIMiddleClick() // Toggles door safety.
-	if(safe)
-		Topic(src, list("src" = UID(), "command"="safeties", "activate" = "1"), 1) // 1 meaning no window (consistency!)
-		to_chat(usr, "The door safeties have been overridden.")
+/obj/machinery/door/airlock/AIMiddleClick() // Toggles door bolt lights.
+	if(!src.lights)
+		Topic(src, list("src" = UID(), "command"="lights", "activate" = "1"), 1) // 1 meaning no window (consistency!)
 	else
-		Topic(src, list("src" = UID(), "command"="safeties", "activate" = "0"), 1)
-		to_chat(usr, "The door safeties have been reset.")
+		Topic(src, list("src" = UID(), "command"="lights", "activate" = "0"), 1)
 	return
 
 /obj/machinery/ai_slipper/AICtrlClick() //Turns liquid dispenser on or off


Emanthealmighty:
Why should door bolt lights be a macro, but not door safeties? What is there to bolt lights that requires you to set them quickly?

Opening doors, bolting them and electrifying them are all macroed to be doable on the run, yet for some reason setting door safeties, which makes you capable of crushing people with doors, wasn't.

That means you had to run ahead while nukies or shadowlings are going about and click on every single door to disable the safeties and then hope they're actually walking there. And then when whatever you wanted to crush has passed by? You'd have to go back to those doors to re enable the safeties by again, clicking on them and opening up the whole menu and fidgeting with the barely responsive buttons to make sure the normal crew doesn't get crushed accidentally. 

I'd rather keep macros to combat related things that need to be done quickly to at least offer some better actions for AIs allowed to harm certain people.

variableundefined:
Although it is named revert, it is more like a tweak

Spacemanspark:
Door opening, closing and bolting can be circumvented easily enough with tools and know how in most situations.
Door electrifying is a great opener to attacks, but you can quite literally just walk away from it if it's just the AI or just wear gloves.

Door crushing, however, is one of the few methods of killing someone that allows you to near infinitely chain-stun-kill someone at the same time. Without outside help, you're fucked.
It's also something that takes a bit of an art to master. Proper timing, largely. It's not something that should necessarily be bound to a macro.

That said, I probably wouldn't have made this PR had it been bound to a different button combination, as I requested in the PR you made. Door lights is actually very useful for trickery, and having it bound to a macro makes, say, bolting down the AI satellite much easier for those of us who do it, as most of us tend to hide the AI bolt lights.

There are still several combos you could have probably pulled off, but you chose to override something that already had a use... for the intention of giving something else that probably shouldn't have a macro in the first place a macro.

Emanthealmighty:
I still do not see why you would need to disable door bolts on the run. 

The only way for it to trick someone is if they walk up to a door which has been bolted and with disabled door bolt lights for a while... But even then, it still takes no time to notice the bolts are down by opening the maintenance panel.

If you bolt someone in a room, they will hear. Even if you disable the door bolt lights they will know some doors were bolted from the sound they make and they'll most likely be prepared for it.

But disabling the safeties? Nobody expects that.

I agree that the stunlock is way too powerful, so door crushing could be tweaked, but it certainly still seems more useful than door bolt lights. That is why I just replaced bolt lights with that.

Spacemanspark:
My point still stands that you could have used another combination for it.

As for door bolt lights, I've used it quite effectively at confusing people, especially those that are trying to run away. Sure, they'll hear the sound of door bolts going down (most of the time, anyways, sometimes I plan ahead and disable the door bolt lights and bolt the door before they can be within hearing distance of said door) but they won't know _what_ was bolted specifically. It forces a double take.

As someone that's played AI for several years at this point, I find door bolt lights being disabled is much more useful at times, especially bound to the middle mouse click button.

pinatacolada:
I use door bolt lights to confuse ppl way wayyy more often than door crushing, since its more common to have laws that prevent it.

PR #12208, opened 2019-09-05
stops bloody footprints from stacking

Couls:
**What does this PR do:**
Follow up to the first blood fix, this one should fix it for real this time. The issue was blood from multiple species would endlessly stack, now the code for the footprint checks all the blood on the tile rather than only the first one. There should now only be one footprint per blood colour, when testing there were three blood tracks one being human blood, another being IPC oil and the third being vox blood.

**Changelog:**
:cl:
fix: fixes footprints stacking
/:cl:


AdvnetureDrake:
Don't forget, Diona also have blood with its own color.

PR #12209, opened 2019-09-05
Further Updates Lavaland

Fox-McCloud:
Further updates lavaland aspects to parity with TG

DO NOT MERGE, WIP, AND NEEDS TESTING AND REVIEW

@Markolie 

Changes:
- Spawning in of megafauna is a bit different--you should see far less megafauna spawning super close to the base
- Explorers belts can now hold shovels and pill bottles
- Mining scanner overlays are now much brighter and stand out; they also clear instantly when you mine a turf
- Added two alternative version of the mining belt that are found on legion corpses
- Gutlunches now produce small version of themselves that eventually grow into adult version
- Basilisks can now spawn from gold cores
- Legions can damage objects (this is the full sized ones, not the tiny legion skulls)
- Legion skulls (tiny floating ones) should not longer block movement
- Altered mineral turfs fairly dramatically in terms of how they are handled
  - There were a lot of turfs that had `planetary_atmos` set for some reason, despite being a wall turf---we really don't want these to process or be factored into calculations; I suspect this was causing far more processing than we otherwise would have needed; it should really only be lava, chasms, and the surface that have air on them and process air; the mineral turfs should be "sleeping"
- Mining ripley starting equipment tweaked a bit; the cell is going to be drained and it can no longer have multiple mining pieces attached to it (it'll be either a drill, diamond drill, OR a plasma gun)
- Ripley speed buffed
- Ripley drill functionality changed; it will now have a delay before it will attempt to repeatedly drill a target
- Drills will butcher mobs that have over 200 brute and are dead; if they can't be butchered, it'll gib them
- Ripleys can no longer drill through the floor
- Speed of the mining scanner dramatically improved for the ripley
- plasma gun has lower CD for the ripley
-Removed secure crates as lot potential
- Fixes Lava Rivers Not Spawning

:cl: Fox McCloud
tweak: Should see less megafauna spawning near the lavaland base
tweak: Explorer belts can hold shovels and pill bottles
add: Legion corpses can now spawn with a few uniquely themed mining belts
add: Gutlunches no longer produce adult version of themselves, but babies that eventually grow into adults
tweak: basilisks are goldcore spawnable
tweak: Legions (not the tiny legion skulls) can damage objects
tweak: legion skulls shouldn't block movement
tweak: overlays for ores are more clear and now clear, instantly, when a turf is mined
tweak: Ripleys are a bit faster
tweak: Mining drills work differently when targeting something; they have a delay before drilling, then they start drilling rapidly, dealing continuous damage to something
Add: Mining drills will now gib or butcher mobs over 200 damage
del: Ripleys can no longer drill through to the floor
tweak: ripley mineral scanner works a lot more quickly now
tweak: Lowered the cooldown on Ripley plasma gun
del: removes numperlocked crates as potential loot
fix: Fixes Lava Rivers not spawning
/:cl:


code/game/turfs/simulated/floor/asteroid.dm
```diff
@@ -31,79 +47,42 @@
...
-		var/obj/item/stack/tile/Z = W
-		if(!Z.use(1))
-			return
-		if(istype(Z, /obj/item/stack/tile/plasteel)) // Turn asteroid floors into plating by default
-			ChangeTurf(/turf/simulated/floor/plating, keep_icon = FALSE)
-		else
-			ChangeTurf(Z.turf_type, keep_icon = FALSE)
-		playsound(src, 'sound/weapons/Genhit.ogg', 50, 1)
-
-/turf/simulated/floor/plating/asteroid/gets_drilled()
```
Fox-McCloud: May need to be re-added.

code/game/turfs/simulated/floor/asteroid.dm
```diff
@@ -31,79 +47,42 @@
 /turf/simulated/floor/plating/asteroid/remove_plating()
 	return
 
-/turf/simulated/floor/plating/asteroid/ex_act(severity, target)
```
Fox-McCloud: May need to be re-added.

code/modules/mob/living/simple_animal/hostile/mining/gutlunch.dm
```diff
@@ -133,3 +133,34 @@
...
 		regenerate_icons()
+
+/mob/living/simple_animal/hostile/asteroid/gutlunch/grublunch
+	name = "grublunch"
+	wanted_objects = list() //They don't eat.
+	gold_core_spawnable = CHEM_MOB_SPAWN_INVALID
+	var/growth = 0
+
+//Baby gutlunch
+/mob/living/simple_animal/hostile/asteroid/gutlunch/grublunch/Initialize()
```
Fox-McCloud: mapload

code/modules/mob/living/simple_animal/hostile/mining/gutlunch.dm
```diff
@@ -133,3 +133,34 @@
...
+	var/mob/living/L
+	if(prob(45))
+		L = new /mob/living/simple_animal/hostile/asteroid/gutlunch/gubbuck(loc)
+	else
+		L = new /mob/living/simple_animal/hostile/asteroid/gutlunch/guthen(loc)
+	mind?.transfer_to(L)
+	L.faction = faction.Copy()
+	L.setDir(dir)
+	visible_message("<span class='notice'>[src] grows up into [L].</span>")
+	qdel(src)
```
Fox-McCloud: Whoever wrote this on TG didn't know what they were doing.

They hard referenced the old mobs factions and called `Destroy` directly---OUCH. Never call `Destroy` directly.

code/game/turfs/simulated/floor/asteroid.dm
```diff
@@ -31,79 +47,42 @@
...
-	else if(istype(W,/obj/item/storage/bag/ore))
-		var/obj/item/storage/bag/ore/S = W
+	else if(istype(I, /obj/item/storage/bag/ore))
+		var/obj/item/storage/bag/ore/S = I
 		if(S.collection_mode == 1)
 			for(var/obj/item/stack/ore/O in src.contents)
-				O.attackby(W,user)
-				return
-
-	else if(istype(W, /obj/item/stack/tile))
```
Markolie: I think we'll want to keep this. As our turf system is vastly outdated, this is necessary to make it possible to build bases on Lavaland with normal turfs.

code/game/turfs/simulated/minerals.dm
```diff
@@ -6,7 +6,7 @@
 	icon_state = "rock"
 	var/smooth_icon = 'icons/turf/smoothrocks.dmi'
 	smooth = SMOOTH_MORE | SMOOTH_BORDER
-	canSmoothWith
+	canSmoothWith = null
```
Fox-McCloud: This file is what I'm most concerned about looking at.

code/game/turfs/simulated/floor/asteroid.dm
```diff
@@ -166,31 +144,29 @@ GLOBAL_LIST_INIT(megafauna_spawn_list, list(/mob/living/simple_animal/hostile/me
...
 /turf/simulated/floor/plating/asteroid/airless/cave/volcanic/has_data //subtype for producing a tunnel with given data
 	has_data = TRUE
 
-/turf/simulated/floor/plating/asteroid/airless/cave/New()
+/turf/simulated/floor/plating/asteroid/airless/cave/Initialize(mapload)
 	if (!mob_spawn_list)
 		mob_spawn_list = list(/mob/living/simple_animal/hostile/asteroid/goldgrub = 1, /mob/living/simple_animal/hostile/asteroid/goliath = 5, /mob/living/simple_animal/hostile/asteroid/basilisk = 4, /mob/living/simple_animal/hostile/asteroid/hivelord = 3)
 	if (!megafauna_spawn_list)
-		megafauna_spawn_list = GLOB.megafauna_spawn_list
+		megafauna_spawn_list = list(/mob/living/simple_animal/hostile/megafauna/dragon = 4, /mob/living/simple_animal/hostile/megafauna/colossus = 2, /mob/living/simple_animal/hostile/megafauna/bubblegum = SPAWN_BUBBLEGUM)
```
Markolie: Are you sure this works? I had to use the global list workaround as this method allowed more than one bubblegum to spawn for some reason.

code/game/turfs/simulated/floor/asteroid.dm
```diff
@@ -143,10 +121,10 @@
 	oxygen = 0
 	nitrogen = 0
 	turf_type = /turf/simulated/floor/plating/asteroid/airless
```
Fox-McCloud: this top one should be `baseturf`

code/game/turfs/simulated/minerals.dm
```diff
@@ -184,11 +177,10 @@
 /turf/simulated/mineral/random/high_chance/volcanic
 	environment_type = "basalt"
 	turf_type = /turf/simulated/floor/plating/asteroid/basalt/lava_land_surface
-	baseturf = /turf/simulated/floor/plating/lava/smooth/lava_land_surface
+	baseturf = /turf/simulated/floor/plating/asteroid/basalt/lava_land_surface
```
Fox-McCloud: right, so, I noticed that the baseturf of a lot of these is uh...lava? This is peculiar--not sure why it varies on TG versus here.

The only thing I can think of is that this is the ANCIENT version of lavaland where `ex_act(1)` or the singularity would remove it to the base turf, and the base turf WAS lava back (or was it chasms? I forget) then (that didn't last long and was too problematic)

code/game/turfs/simulated/floor/asteroid.dm
Fox-McCloud: I'll give it a go to see what happens---I pretty much just ensured consistency with TG; don't have time to test this tonight, but thanks for notifying me.

code/game/turfs/simulated/floor/asteroid.dm
Fox-McCloud: Ahh, thanks for letting me know.

code/game/turfs/simulated/minerals.dm
Fox-McCloud: Alright, so after looking at this...yeah, going to have to revert this portion. 

Basically, they don't have the same baseturf system and they more gracefully handle turfs changing from one to another.

This means that you can RCD a turf or deconstruct things and it'll end up being lavaland's surface, but the singularity will still strip it down to the lava.

We don't have such a system where it acts as a list; it's one turf type and that's it, so changing this to the basalt versions is going to have unintended consequences as our system expects it to be the smooth lava and for us to use mapping helpers (TG doesn't use those anymore either).

code/game/turfs/simulated/minerals.dm
```diff
@@ -253,7 +243,6 @@
 	oxygen = 14
 	nitrogen = 23
 	temperature = 300
-	planetary_atmos = TRUE
```
Fox-McCloud: All of these need to go though.

Fox-McCloud:
Alright, just need to test out the spawning, repeatedly to make sure it spawns the megafauna properly and doesn't wind up giving duplicate Bubblegums, plus no other general oddities.

Fox-McCloud:
Hmm...

@Markolie I"m getting "bad del" errors with this when some megafauana spawn---aside from that, all bugs on this have been addressed.

Any idea what's up? Ever encountered anything like that before?

PR #12219, opened 2019-09-06
Adds comments with the chance of winning at slots

Shadow-Quill:
**What does this PR do:**
`// .02%`
So you don't have to do it yourself.
(Correct me if I'm wrong, please.)

### DIFF
--- a/code/game/machinery/slotmachine.dm
+++ b/code/game/machinery/slotmachine.dm
@@ -92,7 +92,7 @@
 			result = "You win ten credits!"
 			resultlvl = "good"
 			win_money(10)
-		else // 74.05%
+		else // 75.31%
 			result = "<span class='warning'>No luck!</span>"
 			resultlvl = "average"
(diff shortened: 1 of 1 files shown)


code/game/machinery/slotmachine.dm
```diff
@@ -55,44 +55,44 @@
...
 			resultlvl = "good"
 			win_money(50)
-		if(301 to 1000)
+		if(301 to 1000) // 17.26%
 			atom_say("Winner! [userName] has won ten credits!")
 			result = "You win ten credits!"
 			resultlvl = "good"
 			win_money(10)
-		else
+		else // 74.05%
```
Couls: 75.3%, 1000 to 4050 is 3050, divide that by 4050 and multiply by 100

Shadow-Quill: how did I mess that up 🤔

PR #12220, opened 2019-09-06
Lavaland Update: Mechs and Scanners

Fox-McCloud:
There was some oddities about the previous Lavaland PR, and to be honest, it was too big for too many  disparate changes anyway. Consider this a basic re-upload, only it does far less than the other PR.

Re-upload, with lesser scope of: https://github.com/ParadiseSS13/Paradise/pull/12209

Updates Ripleys and the mining drill to make them more viable. Also updates mining scanners

Ripleys

- Move slightly faster in high pressure, but dramatically faster in low pressure
- Mining drill reworked
  - It no longer just drills once; there's a short delay before it will attempt to drill, repeatedly
  - Mobs take less damage from a drill and are no longer instantly knocked out, but there's now a chance to dismember them--if they're dead and have over 200 brute, then they'll either be harvested or gibbed
- Can no longer drill through floors
- Lowered the cooldown on the mech plasma gun
- Mech mining ripley will never have multiple mining tools on it and now starts off with a depleted cell

Mining Scanners

- Mining scanners now apply a much brighter and visible overlay
- Mining overlays clear instantly when a new overlay is applied to them OR you mine the turf

:cl: Fox McCloud
tweak: Ripleys are slightly faster on station, but dramatically faster on Lavaland
tweak: Mech mining drings can no longer drill through floors; they also don't just drill once. After a short delay, they'll attempt to drill the target, repeatedly.
tweak: Mech mining drills no longer instantly knock out mobs; if the mob is dead and has over 200 brute, the drill will attempt to harvest or gib the mob; can also disembowel
tweak: Mech plasma gun has a lower cooldown
tweak: Mining scanners apply a much brighter and visible overlay; mining a turf will now instantly clear the overlay
tweak: Mining mech (the one that's a lavaland find) has a partially depleted cell; it'll also no longer have multiple mining tools on it
/:cl:


MrMagolor:
Would this make Ripley's (somewhat) more viable for megafauna combat, as you no longer need to manually spam the drill?

PR #12227, opened 2019-09-07
Introduces local gps signals

TDSSS:
**What does this PR do:**
This PR modifies gps so that they can be marked as 'local'. A local gps is only visible to other gps on the same z level. For the moment the only local gps are the internal ones in lavaland megafauna, but I plan to add some to some space ruins to make them easier but not totally trivial to find. 

They aren't intended for being used by crew for the moment, as seeing other gps and your own coordiantes without being seen in turn has balance implications.

Just compare the before/after when going to/from lavaland:
Capture1
Capture2
I got really tired of the gps clutter

**Changelog:**
:cl:
add: local gps signals, visible only on the same z level
tweak: lavaland mob signals were made local, reducing gps clutter
/:cl:

### DIFF
--- a/code/modules/telesci/gps.dm
+++ b/code/modules/telesci/gps.dm
@@ -11,6 +11,7 @@ var/list/GPS_list = list()
 	var/emped = 0
 	var/turf/locked_location
 	var/tracking = TRUE
+	var/local = FALSE	//local gps show up only to gps on same z level
 
 /obj/item/gps/New()
 	..()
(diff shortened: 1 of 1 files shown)


lolzyking:
Thanks Tdss, gps spam is bad.

for the space ruins, I don't think you need to bother with the static ones on the fixed z levels ( we all know those by heart) but for the random ones, perhaps giving them bland names "a facility"  "a crash site" and have them shared between  multiple ruins would be a good move?

TDSSS:
Space ruins are all randomly placed, the things with fixed locations are just...places in space.

PR #12230, opened 2019-09-08
Refactors Spawners/Tendrils

Fox-McCloud:
Refactors spawners to be a structure type with a spawner component instead of being a snowflakey simple mob subtype. Better way of fixing the anchored/moveable tendril problem, not to mention this is better expandable in the future.

:cl: Fox McCloud
tweak: Refactored tendrils; they won't show up on health HUDs anymore
fix: Fixes Tendrils not emitting light
fix: Fixes tendrils not changing the terrain around them
/:cl:


code/_onclick/hud/screen_objects.dm
```diff
@@ -11,7 +11,8 @@
 	icon = 'icons/mob/screen_gen.dmi'
 	layer = HUD_LAYER
 	plane = HUD_PLANE
-	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF
+	resistance_flags = INDESTRUCTIBLE
```
Fox-McCloud: I went ahead and changed all things that had a resistance flag of `LAVA_PROOF` or `FIRE_PROOF` to be a `burn_state` instead.

Mixing the two results in incorrect flag logic; when I originally had tendrils with `resistance_flags = LAVA_PROOF | FIRE_PROOF`, they were invincible.

code/game/objects/structures/spawner.dm
```diff
@@ -0,0 +1,124 @@
...
+/obj/structure/spawner/clown
+	name = "Laughing Larry"
+	desc = "A laughing, jovial figure. Something seems stuck in his throat."
+	icon_state = "clownbeacon"
+	icon = 'icons/obj/device.dmi'
+	max_integrity = 200
+	max_mobs = 15
+	spawn_time = 150
+	mob_types = list(/mob/living/simple_animal/hostile/retaliate/clown)
+	spawn_text = "climbs out of"
```
KasparoVy: "stumbles out of" hahaha. They're clowns, after all

code/game/objects/structures/lavaland/necropolis_tendril.dm
```diff
@@ -0,0 +1,100 @@
...
+GLOBAL_LIST_INIT(tendrils, list())
+
+/obj/structure/spawner/lavaland/Initialize(mapload)
+	. = ..()
+	emitted_light = new(loc)
+	gps = new /obj/item/gps/internal(src)
+	GLOB.tendrils += src
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/structure/spawner/lavaland/LateInitialize()
```
Fox-McCloud: @Markolie 

I noticed this was commented out; I wasn't sure why at first, until I noticed it causes runtimes; changing it to a `LateInitialize` resolved the issue and properly flattens the turfs around it.

code/__DEFINES/flags.dm
```diff
@@ -137,8 +137,11 @@
 #define FLAMMABLE 0
 #define ON_FIRE 1
 
-//resistance_flags
-#define INDESTRUCTIBLE 64 //doesn't take damage
+//Fire and Acid stuff, for resistance_flags
+#define UNACIDABLE		(1<<4) //acid can't even appear on it, let alone melt it.
```
Fox-McCloud: Note to future coders; this format is preferable to the previously 1, 2, 4, 8, 16, 32.

KasparoVy: Sorry for the bother but would you mind elaborating on that a bit? Is it for readability purposes or does it present more usable flags than the 1 2 4 8 16 32 etc. format?

Just curious! I've seen it in Baycode but don't understand

Fox-McCloud: It's for readability, mostly.

It tells us just how many bits we've used up, and is a lot easier when you get into higher numbers.

KasparoVy: Thanks, I appreciate the explanation

PR #12242, opened 2019-09-09
Ports Frozen Status Effect

Fox-McCloud:
Very simple port.

Ports over TG's frozen status effect---normally freon applies this; we don't have Freon, but someone might want to use this for something, in the future.

Currently, only Watcher Icewwings use this; if you get hit by one of their projectiles, it'll freeze you temporarily.

Being frozen gives you an ice cube overlay and prevents movement; you can still attack or use items in this state, you just can't move.

:cl: Fox McCloud
add: Watcher Icewing blasts will briefly freeze you
/:cl:


JKnutson101:
Are Drask immune to the effects?

PR #12243, opened 2019-09-09
Adds the recharger belt, a belt that recharges energy guns stored inside it

TheSardele:
**What does this PR do:**
Adds the recharger belt, a belt that charges energy guns stored inside of it using a separate battery and can be ordered by cargo. I made a forum post a while ago to explain why I think something like this was needed https://nanotrasen.se/forum/topic/16391-why-i-think-energy-lethals-suck-and-some-ideas-on-how-one-could-improve-them/ .

Sprites by Drakeven
rechargerbelt

**Changelog:**
:cl:
add: Adds the recharger belt, a belt that charges energy guns stored inside of it using a separate battery and can be ordered by cargo.
/:cl:


Fox-McCloud:
Absolutely not.

AzuleUtama:
While I like the idea, I think having this as something that cargo can easily order in for a low 20 points isn't a great idea, since it allows security to turn all their tasers / laser guns into makeshift AEGs right when the round starts.

It looks like something that would either be better as an R&D item (though they would be outclassed by AEGs here) or something that Red/Gamma ERT members could spawn with so they have something to charge their little taser pistol.

Also, the thread you linked specifically talks about LETHAL energy weapons. I don't think anyone would argue that disablers and especially taser shots are currently weak, which this item would also buff.

SomeSortaSquid:
Think of the bigger picture, too.
Yes, it does make energy weapons more appealing, but at what price?
You now have pretty much every officer wanting two laser guns and this belt, so they can spray and pray more, without ever having to stop.
Not to mention cases where security will foam their mouths up when cargo says "no, not enough points for your belts". Don't even pretend it won't lead to security demanding them from cargo every single round, possibly even at the very beginning.

Unpopular opinion: Armoury is fine. Security can beat changeling, holoparasite users, energy sword wielders and other nasties using it. Ordering more weapons from cargo or RnD is also a valid option. The problem seems to be security being too lazy to arm themselves up, not energy weapons being what they're meant to be. Given all of that, why would you want to implement basically an entirely new meta?

TheSardele:
I was aware that effectively limitless gun recharging was very powerful, so I tried taking some steps to balance that. For example, you can't use potato batteries in it, so you can't use 400k batteries, you usually won't get anything better than a bluespace battery, which is effectively 4 reloads. The belt is also bulky, so unless you have a BoH you can only carry one on your belt slot, which is a rather useful slot to begin with. Using a yellow slime core with it is slower than an AEG, so they still have their place, especially since they are just normal sized and easy to carry multiple of. During blob rounds you're still better of just ordering a ton of laser guns and moving stationary rechargers around.

If being able to get this through cargo is too powerful, would this be worth considering as an ERT or a trader item?

SomeSortaSquid:
ERT is already heavily overloaded. You basically have to choose which gear you'll discard to keep your equipment manageable. Taking away their belts reduces the amount of space they have, so it's questionable idea.
You won't throw your energy gun away with this belt, sure. However, throwing that gun away is a perfectly valid option to free up some much-needed space, because its limited usage makes it relatively unappealing.
Overall, I'm neutral on giving it to ERT and traders, but definitely don't want to see it as a security meta.

Couls:
I understand that a lot of work went into this but I agree with fox on this, this can seriously upset balance and I think energy guns are already in a fine place, I’ve been playing security more and I don’t see this as something that’s necessary or even more convenient.. just feels overpowered. Maybe it can be an admin item

Fox-McCloud:
To put things into perspective.

TG has a power inducer; it's this awesome device that can transfer power from its internal cell (which you can change) to anything else that has a cell.

With one exception; energy guns.

A long time ago we allowed you to change out the cells on energy guns--while this PR is nothing like that, it lends itself a bit towards that. The end result of being able to do so? Sec tried to get ahold of higher capacity cells ASAP and the impact it had on game balance was more than a bit nuts; people will hyper capacity cells (bluespace cells) could stand and deliver for an incredible amount of shots--there was no retreating or considering how carefully placed your shots were; it was just spam.

Again, I can't stress enough that I'm not saying your PR goes that far--but it lends itself towards that mentality, which I do not think is healthy for the overall game's design or balance.

TheSardele:
Alright, you make fair points. I made this PR fully knowing that this was potentially meta-changing, and was likely not going to be accepted. At any rate, I feel like I learnt a lot just making this PR, and I had fun. Thanks for the feedback.

Issue #11902, opened 2019-07-20
Ash walker nest is invincible


Denthamos:
**Problem Description**:
The ash walker tendril cannot be destroyed by melee or guns that miners have.  Additionally KA's do not work in the room since it is pressurized for some reason unlike the rest of lavaland surface.

**What did you expect to happen**: 
To destroy the ash walker tendril by KA or knife.

**What happened instead**:
It cannot be killed.

**Why is this bad/What are the consequences:**
You cannot destroy the tendril nest as a miner which leads to constant ash walkers spawning as long as some are alive.

**Steps to reproduce the problem**:
Attack the ash walker tendril.  You cannot destroy it.

**When did the problem start happening**:
For a while, just the people I asked to report it before never reported it.

Markolie:
This will be fixed by the universal damage PR, which is why I haven't touched this yet. Hopefully that will still go through.

Issue #10930, opened 2019-02-20
Meteor strikes leave solars catwalk tiles in a broken state


SpessX:
**Problem Description**:
When meteors hit solars catwalks, the catwalk tile is left in a broken state (usually a black tile, but sometimes apprently intact) that you can't repair or lay wire over.

**What did you expect to happen**: 
Catwalks should be ok or not ok, not "sort of ok but actually not."

**What happened instead**:
The catwalk tile refused to cooperate with my repair work.

**Why is this bad/What are the consequences:**
In most cases meatspace crew aren't evolved enough to operate PACMANs during meteors, so it's up to lil repair buddies to go out and fix solars. Although you can usually add more catwalk to route things around the bad tiles, you may not manage to reconnect a good chunk of arrays.

It also looks like the kind of really broken stuff that can make people distrust the state of the server and go elsewhere.

**Steps to reproduce the problem**:
1. Admin becomes bored and triggers meteors
2. Wait for meteor to hit solars
3. Try to lay wire on catwalks within the damage area of a meteor

**When did the problem start happening**:
Been like that for years, specifically seen on fore starboard. 

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was anything else out of the ordinary happening?)**:
Reporting the issue for the record, I don't think anyone else really cares about those catwalks being broken.

Issue #3327, opened 2016-01-22
Can wrench holodeck stools for real metal.


tigercat2000:
The stools in the template are normal stools and the holodeck doesn't have a "holo stool"

Spacemanspark:
This reminds me of the holodeck issues with racks.

tigercat2000:
i didn't want to report this because it's the most useful exploit


but gotta get dem quads mang

taukausanake:
Someone should check every object that the holodeck uses just to make sure this issue doesn't come up again

Anticept:
Confirming this is still an issue.

Holodeck metal tables actually prevent you from wrenching them with the chat message "It's a holotable! There are no bolts!", so I doubt this would be difficult to implement for wood tables and stools.

taukausanake:
Yeah but with the VR PR going this might not be relevant anymore. Also I thought this WAS fixed?

tristan1333:
According to all known laws of physics, matter cannot be created nor destroyed. The holodeck, of course, creates matter anyway because holodecks don't care what humans think is possible.

Ty-Omaha:
Still active, just tested

Anticept:
@tristan1333 Energy cannot be created nor destroyed (except for maybe some quantum voodoo). With our current level of understanding of physics, we actually do create and destroy matter (fission releases some mass as energy, particle accelerators, etc).

taukausanake:
I think I'll just on this today if no one else is

Issue #2805, opened 2015-12-11
Ripley pretty much broken down will never move again and possibly related heat bug


Ar3nn:
http://prntscr.com/9cqurg

Ripley broke down, fully repaired it, but...

http://prntscr.com/9cqvdp

No issues other than a tank temp of 6000C, the cabin temp is 50C and I still take damage faster than Diona heal while riding, and it cannot move.

Ar3nn:
http://prntscr.com/9cqwau

After riding in it for a couple minutes, almost critical.

Crazylemon64:
Is this able to be reproduced?

Jovaniph:
I had a similar issue about this. My job was a Roboticist, and some Shaft Miner brought a Ripley that had been damaged. I manage to fix everything but the Ripley kept draining the new batteries I put in it.

I think its more of a missing feature than a bug however because I had admins and mentors trying to figure out why nothing worked.

KasparoVy:
Man that heat bug was such a throwback. Back when I first started playing it cooked me alive a few times and I thought I got fucked over by a 'solar flare' or some shit.

Haven't seen it in years.

Fox-McCloud:
You can still get cooked inside it, under the right conditions.

There *IS* a damage flag for mechs that's like `damage_internal_fire` or something; that can cause you to heat up, yourself.

That said, damage oddities should be far more consistent now.

PR #12245, opened 2019-09-09
Universal Object Damage

Fox-McCloud:
This needs to be done. Need I say more?

Obviously WIP and obviously DNM.

Universal damage turned on; acid melts things over time now; armor values now have acid and fire attached to them.

Should fix a ton of things including lava's behavior--also means a mostly destructable environment.

img
img

fixes: https://github.com/ParadiseSS13/Paradise/issues/12246
fixes: https://github.com/ParadiseSS13/Paradise/issues/11902
fixes: https://github.com/ParadiseSS13/Paradise/issues/10930
fixes: https://github.com/ParadiseSS13/Paradise/issues/2805
fixes: https://github.com/ParadiseSS13/Paradise/issues/3327

:cl: Fox McCloud
add: Most things are universally damageable
tweak: Acid melts things a bit differently now
fix: Lava actually sets most things on fire.
tweak: Captain no longer has access to his display case by default
add: Xenos can break open vents
/:cl:


Aurorablade:
I like the orignal title..the theme song of this PR is now Ring of FIre

Ty-Omaha:
can we like

shoot the tesla to kill it

or throw rubber gloves at it

Fox-McCloud:
Also, there is probably balance altering armor changes in here for a number of things. I'm not going to document them all; for the most part, it's minor buffs/nerfs, with the exception of maybe windows. 

This is mostly copy-pasting form TG to ensure consistency in behavior--this isn't me trying to sneak in changes; it's just that with something this massive, I really don't have time/willpower to document every single armor change.

Fox-McCloud:
There's one thing I'm not going to be porting over to this system, because they're atrocious, snowflakey, and they can get wrecked for all I care; spacepods.

The changes here shouldn't be a problem for spacepods; they *should* still behave as before, but I'm not going to invest a lot of time into getting them to work with this system.

To be blunt, spacepods should probably just be a type of mech subtype that can process space movement with little delay, but that's for another day.

code/ATMOSPHERICS/atmospherics.dm
```diff
@@ -31,8 +31,8 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 	var/image/pipe_image
 
 /obj/machinery/atmospherics/New()
-	if(!armor)
-		armor = list(melee = 25, bullet = 10, laser = 10, energy = 100, bomb = 0, bio = 100, rad = 100)
+	if (!armor)
```
Couls: space

Fox-McCloud:
My face when someone set `unacidable` all over the place on an away mission.

No wonder TG considers those things almost universally trash.

KasparoVy:
Polyacid your way to victory comrade

Fox-McCloud:
had to do a couple of refactors in the way; display cases and ash trays. 

Can't construct alarmed display cases, the cap's display case is no longer fingerprint locked, and it has access that not even cap has (he'll have to break it).

I may change the latter most part of it, but we'll see; it was mostly just trying to keep it up to date and avoiding the problems that the display case would cause with how object damage works.

Either case, most `var/health` is killed off. 

Shields and shield generators still use it, because they need an extensive refactor to work with this system--portable turrets also have `var/health` I'm not sure if I'll tweak them, revert them back to pre universal object damage....or what; the three problem things with this PR are space pods, portable turrets, and field emitters/forcefield generators/shields.

Either case, I'll think about what to do with portable turrets; once I've decided and handled that (or not), then it's onto the final stretch; `attackby` returning parent value.

Then, and finally then, I can start testing and how much this broke and how badly this works.

Fox-McCloud:
Ok, yeah, I'm so sick of this PR and the other refactors I have to do to get it to work....

I'm going to just save the portable turrets for later, along with the field generator and space pods.

Those are all inconsequential enough (and/or work fine with their snowflake BS), this is so long overdue, and I'm so tired of working on this, that it can wait. 

In any event, this is the PR "done" from what I believe needs to be changed.

Tomorrow I start testing.

Fox-McCloud:
Ok, another thing that will use the old system: disposals.

Disposal code is goodness *awful* and is something else that needs to be refactored.

Fox-McCloud:
This is the best I can do. Done as much testing as I feel is necessary and within reasonable limits.

KasparoVy:
Legendary, way to go

PR #12249, opened 2019-09-09
The gravitational catapult no longer pushes ghosts when used

SteelSlayer:
**What does this PR do:**
See title & CL

**Changelog:**
:cl:
fix: The gravitational catapult no longer pushes ghosts when used
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/other_tools.dm
+++ b/code/game/mecha/equipment/tools/other_tools.dm
@@ -119,7 +119,7 @@
 				atoms = orange(3, target)
 			for(var/atom/movable/A in atoms)
 				if(A.anchored) continue
-				if(isobserver(A)) continue
+				if(A.move_resist == INFINITY) continue
 				spawn(0)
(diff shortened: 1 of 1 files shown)


variableundefined:
Should be counted as a fix too

SteelSlayer:
Ok change it, `move_resist == INFINITY` works just fine

code/game/mecha/equipment/tools/other_tools.dm
```diff
@@ -119,6 +119,7 @@
 				atoms = orange(3, target)
 			for(var/atom/movable/A in atoms)
 				if(A.anchored) continue
```
variableundefined: You can just do if(A.anchored || A.move_resist == INFINITY)

instead of two separate line. They're mostly the same thing after all.

SteelSlayer: That is true, changing

PR #12255, opened 2019-09-10
Refactors Shards, Improved Light Replacer, Adds Plasmaspear

Fox-McCloud:
You May find this combination odd, but this stemmed from me looking at glass shards and wanting to improve their code, myself, which led me to looking at where else glass shards are used.

Most of the `Crossed` improvements come from TG, but I refactored welding glass shards to be more unified instead of it being snowflaked for each subtype of shard.

The light replacer improvement mostly just makes the light replacer work better by properly incrementing/using lights; the additional feature is that you can use the light replacer in your hands and it'll attempt to replace lights on the same tile you're on.

Plasma glass shards damage has been nerfed, but as a bonus, they do more damage when you step on them instead of the same as regular glass; additionally if you construct a spear out of a plasma glass shard it'll do 1 more damage than regular shards.

:cl: Fox McCloud
tweak: Plasma glass shards do less damage when attacking, but slightly more when stepping on them
tweak: Spears made out of plasma glass do slightly more damage
tweak: Light replacer can now be used in-hand to attempt to reload empty lighting fixtures around you in addition to showing how many lights it has
tweak: Light replacer needs less shards to increment bulbs
/:cl:


code/game/objects/items/weapons/shards.dm
```diff
@@ -48,80 +54,66 @@
...
+		I.attackby(src, user)
+		return
+	return ..()
 
-/obj/item/shard/Crossed(AM as mob|obj, oldloc)
-	if(isliving(AM))
-		var/mob/living/M = AM
-		if(M.incorporeal_move || M.flying || M.throwing)//you are incorporal or flying or being thrown ..no shard stepping!
+/obj/item/shard/Crossed(mob/living/L, oldloc)
+	if(istype(L) && has_gravity(loc))
```
Fox-McCloud: this gravity check should prevent the original problem of getting wrecked, in space, by glass while still allowing the *tactical shoe steal and throw* strat.

code/game/objects/items/weapons/shards.dm
```diff
@@ -48,80 +54,66 @@
...
-			var/mob/living/carbon/human/H = M
+		playsound(loc, 'sound/effects/glass_step.ogg', 50, TRUE)
+		if(ishuman(L))
+			var/mob/living/carbon/human/H = L
+			var/obj/item/organ/external/affecting = H.get_organ(pick("l_foot", "r_foot"))
+			if(!affecting)
+				return
+			if(affecting.is_robotic())
+				return
+			var/feetCover = (H.wear_suit && H.wear_suit.body_parts_covered & FEET) || (H.w_uniform && H.w_uniform.body_parts_covered & FEET)
```
Fox-McCloud: This also improves it so if you have no shoes, but have a suit that covers your feet, you're protected.

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -562,10 +562,6 @@
 /obj/item/robot_module/drone/respawn_consumable(mob/living/silicon/robot/R)
 	var/obj/item/reagent_containers/spray/cleaner/C = locate() in modules
 	C.reagents.add_reagent("cleaner", 3)
-
-	var/obj/item/lightreplacer/LR = locate() in modules
-	LR.Charge(R)
```
Fox-McCloud: This is removed because it's handled on the recharge,r as disgusting as that is. It would be *better* if it was handled here, in `respawn_consumable`, but that'd require a larger refactor of borg rechargers and appropriately using `respawn_consumable`, which I really didn't feel like doing in this PR.

code/modules/power/lighting.dm
```diff
@@ -303,10 +303,7 @@
 	//Light replacer code
 	if(istype(W, /obj/item/lightreplacer))
 		var/obj/item/lightreplacer/LR = W
-		if(isliving(user))
```
Fox-McCloud: This is the strangest safety check I've seen in a while--someone was apparently very afraid of ghosts somehow replacing lights.

code/game/objects/items/weapons/shards.dm
```diff
@@ -33,8 +36,11 @@
 		if("large")
 			pixel_x = rand(-5, 5)
 			pixel_y = rand(-5, 5)
+	if(icon_prefix)
+		icon_state = "[icon_prefix][icon_state]"
 
 /obj/item/shard/afterattack(atom/movable/AM, mob/user, proximity)
+	. = ..()
```
Fox-McCloud: Remove this

Ionward:
Would it be too much if Shark teeth also made slightly stronger spears? They function like glass shards as well.

Fox-McCloud:
@Ionward  

Probably not in the PR--they're rare enough that I'd rather not make an explicit exception for them.

If there ends up being more glass subtypes that are more commonly accessible, then potentially, yeah; then I'd probably make glass shards more OOP in boosting damage.

PR #12258, opened 2019-09-10
Ban evasion autoban 

Ty-Omaha:
adds an autoban for evasion

:cl:
tweak: Minor ban evasion update
/ 🆑


xProlithium:
There are certain situations, like with a person who has a banned ckey and has created a new one, where an auto-ban feature would actually have a negative effect.

Ty-Omaha:
That already exists, so this isn't introducing a new problem.

PR #11733, opened 2019-06-17
Adds Auto-Note for Manual Bans

Ty-Omaha:
**What does this PR do:**
Adds auto noting for manual bans because it has angered me long enough.

🆑 
fix: Manual bans now auto-note
/ 🆑

PR #12259, opened 2019-09-10
Revert "Adds Auto-Note for Manual Bans"

Ty-Omaha:
Removes auto note for manual bans

grouped manual bans such as "security" bans spam notes because of how bans are handled and this doesn't have the proper logic to not spam the notes with "banned banned banned banned banned banned" in 1 group such as "banned from security roles"

also they are getting flagged as all permabans for whatever reason

:cl:
del: Autonote for manual bans removed
/ 🆑


variableundefined:
While this is in fact, spammy, I think having note-less ban might be *very* bad from an administration / accountability standpoint.

Without a system to search logs conveniently having notes to keep track of ban is very important.

Ty-Omaha:
@variableundefined We have had note less manual bans for the entire time paradise existed except for a little over a month when #11733 was merged. Admins are capable of adding a note when they manual ban, having someone's entire notes be spammed is a much bigger issue that will take more of admin time than them simply adding a note when they manual ban.

Admins have a for certain way of seeing bans if we lookup their bans and not notes, if needed.

PR #9201, opened 2018-07-12
Add nutrient dispenser to Abductor ships

datlo:
Replaces one of the alien lockers in each ship by a Nutrient Dispenser
vending machine, which contains 20 chicken cup soups and 12 warm donk
pockets (15u omnizine when eaten).

Also contains two bottles of vodka if hacked.

Has sprite for the machine's various states.
nutrientdispenser

I believe abductors should influence their testing environment as little
as they can, and this includes station supplies. This vending machine
supplies both food and healing, cutting some of the Abductors' reliance
on stealing station supplies.

🆑
rscadd: Added nutrient dispenser to Abductor ships.
/🆑

PR #12260, opened 2019-09-10
Gives abductors an objective to stay hidden/limit crew disruption

TDSSS:
**What does this PR do:**
This gives all abductors a secondary objective, to keep disruption to the crew to a minimum. Abductors are supposed to mostly stay hidden and just kidnap people, but sometimes they steal the AI, space armory, get into shootouts with sec and dab on the captain. From talking with admins I know that's not really intended so I figured a second objective should make it clear what kind of playstyle is expected of abductor players.
The objective's exact text is:
"Disrupt the primitives as little as possible."
The objective defaults to true and can only be failed if an admin manually toggles it, perhaps because someone did something very disruptive. It is on a per abductor basis instead of per team so you can selectively have people fail, but I expect that won't be done much, if at all.

**Changelog:**
:cl:
add: Objective to abductors to not disrupt crew too much.
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -1524,6 +1524,10 @@
 
 	SSticker.mode.abductors |= src
 
+	var/datum/objective/stay_hidden/hidden_obj = new
+	hidden_obj.owner = src
+	objectives += hidden_obj
+
 	var/datum/objective/experiment/O = new
 	O.owner = src
 	objectives += O
--- a/code/game/gamemodes/miniantags/abduction/abduction.dm
+++ b/code/game/gamemodes/miniantags/abduction/abduction.dm
@@ -150,6 +150,8 @@
 	update_abductor_icons_added(scientist)
 
 /datum/game_mode/abduction/proc/greet_agent(datum/mind/abductor,team_number)
+	var/datum/objective/stay_hidden/O = new
+	abductor.objectives += O
 	abductor.objectives += team_objectives[team_number]
 	var/team_name = team_names[team_number]
 
@@ -160,6 +162,8 @@
 	abductor.announce_objectives()
 
 /datum/game_mode/abduction/proc/greet_scientist(datum/mind/abductor,team_number)
+	var/datum/objective/stay_hidden/O = new
+	abductor.objectives += O
 	abductor.objectives += team_objectives[team_number]
 	var/team_name = team_names[team_number]
 
@@ -227,6 +231,13 @@
 /datum/objective/experiment
 	target_amount = 6
 	var/team
+
+/datum/objective/stay_hidden
+
+/datum/objective/stay_hidden/New()
+	explanation_text = "Disrupt the primitives as little as possible."
+	completed = TRUE
+//No check completion, it defaults to being completed unless an admin sets it to failed.
 
 /datum/objective/experiment/New()
 	explanation_text = "Experiment on [target_amount] humans."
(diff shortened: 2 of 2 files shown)


MrMagolor:
On the flipside, if an abductor gets hurt, they NEED to use crew supplies. Recharging the pistol isn't that big a deal since you shouldn't NEED to use it anyways.

My idea is adding a special sort of stabilized hivelord core that you can buy for an equipment point or two.

TDSSS:
I mean, if you're hurt and need to heal, that's a necessary disruption and allowed. You just shouldn't troll crew unnecessarily, or steal every set of surgery tools on station "just in case"

taukausanake:
[Hmmmmmmmm...](https://github.com/ParadiseSS13/Paradise/pull/9201#issuecomment-404640852) Yeah I kind of think ayyliums get a free pass to do all sorts of weird shit. Objectives are good and all but they're not as fun as fucking with the crew

Fox-McCloud:
I feel like keeping them *mostly* non-lethal is a good thing, but having them not interfere with the crew at all beyond what they need to do? Ehhh,kinda iffy on this one.

Abductors are kinda designed around the concept of "messing with the crew"; over the years, the vast bulk of their lethal things have been stripped from them and they've been turned into more of a "cause chaos and insanity". 

Taken more literally, this would even actively forbid messing with your abductees after you put them back on station---which, again, kinda negates the whole point of the device that lets you speak with them and implant objectives into their mind.

TDSSS:
Hmm, maybe the objective's not worded ideally.

Maybe it should forbid 'overt' interference or something like that? Open for suggestions.

polartron:
Secondary Objective: _"Complete your primary objective without the station going to Code Red"_
Mid-round antags will most likely trigger it, but it will give you an optional time limit.

The fact that you **have** to give them glands and that you have a limited supply of them makes it challenging in the sense that you quickly run out of glands that makes it less obvious that they have been abducted. Choose wisely.

Quantum-M:
I strongly agree with this PR. Abductors are considerably advanced species, most likely way, way far ahead than any species on the station, all the acts mentioned (e.g. kidnapping the AI, spacing/stealing the armory, stealing IDs from crew, especially all access) makes them see more like a crappy form of greytide that can teleport. They are also supposed to be a "science team" meant to experiment on the crew, and as with a lot of experiments, the less impact the science team has, the better the "outcome".

It doesn't necesarily mean that the abductors shouldn't take the occasional item (taking a few medkits is fine when it's needed, but stealing all medbay items seems far too much of a thing), or "control" their test subjects (e.g. telling them to distract security to a minimal intent would be reasonable, telling them to kill people/themselves is not reasonable).

**_In short_**, the less abductors act out like greytide, the less dumb/unreasonable like antag they would seem (especially since they are the few non-highly dangers midround antags).

The objective could probably be worded better though, something like: "Avoid causing significant impact to the station crew as possible, with exception of your experiments."

variableundefined:
I'm not good at making suggestions but preventing code red don't walk given that stations tends to be in permanent code red.

Lack of perfect information means that even if the threat is completely dealt with (Usually, they are not), security / command wouldn't know it and won't code blue.

Of course, there's also the fact that Code Red grant more authority.

MrMagolor:
The only real reason currently for station to leave code red is the off chance that there are traders; the event will fire but not truly activate, with an announcement saying that they were denied docking due to heightened security level.

TDSSS:
Best suggestion I've seen so far was "Limit contact with your targets outside of conducting your experiments and abduction." I think limiting contact is still rather permissive, you can have _some_ contact, etc.

variableundefined:
That should be good. Let's try that out.

TDSSS:
I tried to do it with the web editor, fled in terror at how bad it is and then promptly forgot. oops. Wording has been changed now.

PR #12261, opened 2019-09-10
Small Headcrab fix. 

Terilia:
**What does this PR do:**
This PR fixes headcrab behaviour. They are supposed to kill and couple with the people they kill in order to create zombies. They are not designed to stop when the person drops to the ground. 

**Changelog:**
:cl:
fix: stat_attack from unconscious to dead
/:cl:


Ty-Omaha:
your cl is broken
do it like
🆑 
fix: stuff
/ 🆑

PR #12263, opened 2019-09-11
Remove normal subtype of wooden chair

### LINKED ISSUE #12252: Placing wooden chairs instead places down a normal metal chair
**Issue Description**:
Placing wooden chairs instead places down a normal metal chair

1. Pick up a wooden chair
2. Press Z with it in your active hand

Unknown, reported on 9/9/2019 by Ansari

variableundefined:
**What does this PR do:**
Fix #12252 

Remove the /normal subtype of wooden chair. Why does it exists anyway? 

The issue in #12252 was caused by normal wooden chair having a parent that don't have icon_state specified and inherit its parent's icon state, which is normal chair icon.

100+ lines but basically only 7 or so matter, rest is just typepath change.

Sorry mapper, but this should be an easy change to make.

**Changelog:**
:cl:
fix: Refactored wooden chair so placing them won't make you see normal metallic chair anymore.
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -22514,7 +22514,7 @@
 	},
 /area/crew_quarters/bar/atrium)
 "aRy" = (
-/obj/structure/chair/wood/normal,
+/obj/structure/chair/wood,
 /turf/simulated/floor/carpet,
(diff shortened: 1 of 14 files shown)


Fox-McCloud:
the heck.

Issue #12215, opened 2019-09-06
Megafauna


DoctorDrugs:
**Issue Description**:
Megafauna are unable to gib/destroy dead mobs, they will continuously attack a dead body and will focus on constantly attacking them. 

**What did you expect to happen**: 
for the dead body to be gibbed or turned into remains

**What happened instead**:
the megafauna hyper focused on the dead instead of gibbing it and continuing down the list of targets it has

**Why is this bad/What are the consequences**:
someone can exploit this, and make it trivial to kill megafauna

**Steps to reproduce the problem**:
toss a drake/colossus/hiero a dead body they will just constantly pummel it

**When did the problem start happening**:
about 2-3 days ago is when i noticed it

**Extra information**:
N/A

MrMagolor:
I think the title of this should be changed to be more specific.

Not to mention, the other bad part is that it removes the joy from people who actually want to kill megafauna in a fun fight.

DoctorDrugs:
late night issue reports leads to bad titles and yes the TG simplemob refractor is most likely the culprit

PR #12270, opened 2019-09-11
Reworks Ash Drake, Refactors Colossus, Fixes Endless Animal Attacks

Fox-McCloud:
From TG

Lavaland is nearing completion, so it's time to make it *even harder*.

This reworks the Ash Drake--instead of it just being a mob that spews out fire in cardinal directions and makes occasional swoops at you, it will now blow out fire in a big array, turn the floor to lava, and attempt to seal you into its arena and make you stand on a turf before swooping down on you.

This also refactors the Colossus to use better code; for the most part, it has the same impacts and behaviors as it did before.

Why? Megafauna fights should be epic and challenging, not "walk backwards and occasionally change the turf you're on and you'll be good.

Also fixes a bug where megafauna would endlessly attack you and never dust/gib you.

Fixes: https://github.com/ParadiseSS13/Paradise/issues/12215

:cl: Fox McCloud
add: Ash Drake reworked; it's much more powerful now and has a diverse set of attacks--its fire breath is more random, it moves faster, it can turn the ground into lava, and once it's lower on health, will seal you into an arena and make you play the "only one turf is safe; stand on it or suffer" mini-game
fix: Fixes megafauana endlessly attacking dead mobs
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/blood_drunk_miner.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/blood_drunk_miner.dm
@@ -30,7 +30,7 @@ Difficulty: Medium
 	icon = 'icons/mob/alienqueen.dmi'
 	light_color = "#E4C7C5"
 	speak_emote = list("roars")
-	speed = 1
+	speed = 3
 	move_to_delay = 3
 	projectiletype = /obj/item/projectile/kinetic/miner
 	projectilesound = 'sound/weapons/kenetic_accel.ogg'
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
@@ -1,3 +1,9 @@
+#define DRAKE_SWOOP_HEIGHT 270 //how high up drakes go, in pixels
+#define DRAKE_SWOOP_DIRECTION_CHANGE_RANGE 5 //the range our x has to be within to not change the direction we slam from
+
+#define SWOOP_DAMAGEABLE 1
+#define SWOOP_INVULNERABLE 2
+
 /*
 
 ASH DRAKE
@@ -6,10 +12,11 @@ Ash drakes spawn randomly wherever a lavaland creature is able to spawn. They ar
 
 It acts as a melee creature, chasing down and attacking its target while also using different attacks to augment its power that increase as it takes damage.
 
-Whenever possible, the drake will breathe fire in the four cardinal directions, igniting and heavily damaging anything caught in the blast.
-It also often causes fire to rain from the sky - many nearby turfs will flash red as a fireball crashes into them, dealing damage to anything on the turfs.
-The drake also utilizes its wings to fly into the sky and crash down onto a specified point. Anything on this point takes tremendous damage.
+Whenever possible, the drake will breathe fire directly at it's target, igniting and heavily damaging anything caught in the blast.
+It also often causes lava to pool from the ground around you - many nearby turfs will temporarily turn into lava, dealing damage to anything on the turfs.
+The drake also utilizes its wings to fly into the sky, flying after its target and attempting to slam down on them. Anything near when it slams down takes huge damage.
  - Sometimes it will chain these swooping attacks over and over, making swiftness a necessity.
+ - Sometimes, it will encase its target in an arena of lava
 
 When an ash drake dies, it leaves behind a chest that can contain four things:
  1. A spectral blade that allows its wielder to call ghosts to it, enhancing its power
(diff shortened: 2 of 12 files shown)


Couls:
Now that fastmos is gone do you think you can also change the airlocks at the outpost so we don’t need to cycle them?

MrMagolor:
Furthermore- how would you change the blood drunk fight to be more engaging, given that all he does is run at you and whack you, occasionally teleporting a few tiles?

taukausanake:
Ash Drake sounds like lava Ridley now.

Also, Fastmos is dead? What?

variableundefined:
Might want to note down the blood drunk miner speed change too

code/modules/mob/living/simple_animal/hostile/megafauna/blood_drunk_miner.dm
```diff
@@ -30,7 +30,7 @@ Difficulty: Medium
 	icon = 'icons/mob/alienqueen.dmi'
 	light_color = "#E4C7C5"
 	speak_emote = list("roars")
-	speed = 1
+	speed = 3
```
variableundefined: Should probably be included in changelog / PR desc

MrMagolor: That _decreases_ his speed, right? What an odd change, given that my impression was that you had to facetank him

Couls:
> Ash Drake sounds like lava Ridley now.
> 
> Also, Fastmos is dead? What?

not as strong as it used to be, but that's off-topic

code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
```diff
@@ -39,31 +46,358 @@ Difficulty: Medium
 	armour_penetration = 40
 	melee_damage_lower = 40
 	melee_damage_upper = 40
-	speed = 1
-	move_to_delay = 10
-	ranged = 1
+	speed = 5
+	move_to_delay = 5
```
variableundefined: Only significant divergance I can see is no GPS signal (Might or might not be a good thing?) for dragon and also we don't use slaughter demon attack sound for our drake.

MrMagolor: That seems... not right. The point of the GPS signals was to also help people who didn't want to go megafauna hunting and just mine in peace, right?

code/modules/mob/living/simple_animal/hostile/megafauna/blood_drunk_miner.dm
Fox-McCloud: the `speed` var only applies to when players are controlling the mob.

code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
Fox-McCloud: @variableundefined the GPS is defined below, as `internal_type`; we don't have component based GPS like TG does, as such, it's an actual item in the mob itself.

code/modules/mob/living/simple_animal/hostile/megafauna/blood_drunk_miner.dm
MrMagolor: So this is just to make adminbussed blood drunks a little less ridiculous? Sounds fair

code/modules/mob/living/simple_animal/hostile/megafauna/drake.dm
variableundefined: This is me being dumb and misreading

KasparoVy:
Tfw you get your first crusher drake kill and immediately lose all hope of a repeat performance

variableundefined:
Should probably add the space dragon to the changelog too

Fox-McCloud:
Went ahead and reworked bubblegum too.

variableundefined:
Will re-re-review this tommorow

code/modules/flufftext/Hallucination.dm
```diff
@@ -269,51 +269,59 @@ Gunshots/explosions/opening doors/less rare audio (done)
 	px = -32
 
 /obj/effect/hallucination/oh_yeah
-	var/turf/simulated/wall/wall
-	var/obj/effect/hallucination/simple/bubblegum/bubblegum = null
+	var/obj/effect/hallucination/simple/bubblegum/bubblegum
```
Fox-McCloud: all these changes are entirely code improvements, no actual changes to behavior.

Denthamos:
Bubblegum feels too powerful now after this merge due to slower paradise movement speeds and how fast bubblegum moves in comparison with its abilities now.  Miners have zero chance to even respond and die in under 5 seconds.

PR #12288, opened 2019-09-13
Allows people to float on shallow water

Ty-Omaha:
You can now float (lie down) on water that isn't deep without drowning

Resolves the issue of SSD people from being dragged to the pool to drown as a joke.

I've never seen a case where someone has legitimately drowned because of this, just SSDs getting griefed.

🆑 
tweak: Players will no longer drown and instead float when lying down on shallow water.
/ 🆑

### DIFF
--- a/code/game/machinery/poolcontroller.dm
+++ b/code/game/machinery/poolcontroller.dm
@@ -116,7 +116,7 @@
 	if(!drownee)
 		return
 
-	if(drownee && (drownee.lying || deep_water)) //Mob lying down or water is deep (determined by controller)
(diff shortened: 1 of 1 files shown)


Couls:
don't like this because I don't think its a necessary removal, I once drowned a holopara user at the beach gateway because I didn't have lethals and I couldn't get close to them. The only way for me to have won that is to keep drowning them by stunning them until the holopara died sure it's rarely used but I still like the feature itself and there's really no reason not to just add an SSD check and get the best of both worlds here

PR #12291, opened 2019-09-13
Hierophant Update

Fox-McCloud:
From TG.

Updates the Hierophant.

Instead of being some weird giant bird thing, the Hierophant is now an ancient machine that forces you to dance to *its* tune. The overall battle mechanics are mostly the same, but if you attempt to flee the fight arena, it'll trap you inside its own created arena and teleport to you; said arena will persist for 10 seconds---after it expires, you'll have 10 more seconds to get out of dodge before it'll try to trap you again.

Thematically, the arena is completely different too; far less esoteric+magical, more indecipherably advanced technology.

The Hieorophant club does a bit less damage, but it now does increasingly stronger effects based on how weak you are; the weaker you are, the stronger the club becomes, when wielded.

pic:

img

:cl: Fox McCloud
add: Hierophant has been updated--it'll be a bit tougher and will attempt to trap you if you run for it. 
tweak: Heirophant is no longer a weird bird thing, but a massive, indecipherable piece of machinery that will make you dance to ITS tune
tweak: Hierophant club is weaker up front, but becomes stronger the lower health you are
/:cl:


MrMagolor:
Can we get pictures of the sprite/map changes?

Also will giant birb still exist as an admin thing?

PR #12292, opened 2019-09-13
Assassin Guardians no longer have Armor

AzuleUtama:
## What Does This PR Do
It states in the description and welcome message for assassin guardians that they lack any armor in favour of their highly damaging stealth attack.
This is a lie, since they actually have a whopping **40%** reduction to all damage they take. This makes them even tankier than healer or chaos. This PR changes them to have no armor value, making them more accurate to their description.

## Why It's Good For The Game
Stops a class of guardian being stronger than it should be.

## Changelog
:cl:
balance: Assassin guardians no longer have any armor.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/guardian/types/assassin.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/assassin.dm
@@ -2,7 +2,6 @@
 	melee_damage_lower = 15
 	melee_damage_upper = 15
 	armour_penetration = 0
-	damage_transfer = 1
 	playstyle_string = "As an <b>Assassin</b> type you do medium damage and have no damage resistance, but can enter stealth, massively increasing the damage of your next attack and causing it to ignore armor. Stealth is broken when you attack or take damage."
 	magic_fluff_string = "..And draw the Space Ninja, a lethal, invisible assassin."
 	tech_fluff_string = "Boot sequence complete. Assassin modules loaded. Holoparasite swarm online."
--- a/code/game/gamemodes/miniantags/guardian/types/assassin.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/assassin.dm
@@ -2,7 +2,7 @@
 	melee_damage_lower = 15
 	melee_damage_upper = 15
 	armour_penetration = 0
-	damage_transfer = 0.6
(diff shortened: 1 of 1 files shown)


TDSSS:
I feel like they should get a buff to compensate for this extremely large nerf. Assassin guardians are not exactly picked a lot anyway, making them super fragile will not help at all.

AzuleUtama:
> I feel like they should get a buff to compensate for this extremely large nerf. Assassin guardians are not exactly picked a lot anyway, making them super fragile will not help at all.

I wouldn't be against buff suggestions if they're underused, it just makes no sense that something classed as being high damage and stealthy is THIS tanky.

TDSSS:
Hmm, I mean...
Standard and Protector get 0.4, charger, bomb and assassin are 0.6, lightning 0.7, chaos 0.8, ranged 0.9, with only healers totally unarmored. I think it could be reasoned to give them the 0.9 of ranged or something instead of totally no armor.

And consider, having no armor means the full damage of the attack is caused to the host. That's more than if the host just took the hit since he probably has _some_ armor on his clothes ( I don't think host armor is applied on top of that)

Couls:
maybe 10 to 20% would be better, though I agree its strange to advertise no armor and then have the exact same amount as standard

TDSSS:
You could always change the description too, mentioning 'weak armor' or something instead of none.

AzuleUtama:
So I've done some thinking and there's a few options I'm considering to make this less of a nerf while keeping the assassin guardian themed like an assassin:

- Go with above and make them have 10% armor rather than none. No other changes.
- Give them 10% armor and a slight increase to base (not the stealth attack's) damage. The increase would be enough for them to put someone into critical health in one less attack than they can currently, but will still make them deal less damage than normal ones.
- Keep them with no armor but give them thermal sight to help them locate targets for their master and warn them of incoming dangers. Not sure if the removal of armor alone is enough to justify this.

Any thoughts on which would be the best option?

TDSSS:
Personally I like the second option, but the third one sounds fine too. It fits the assassin theme to know where people are so you can jump them when they're alone like that.

Fox-McCloud:
This seems like a copy-pasta error to begin with.

PR #12296, opened 2019-09-15
limit the amount of monkeys spawned from monkey cubes

Couls:
## What Does This PR Do

Idea and some code taken from TG, adds a configurable limit to the amount of monkeys that can be spawned by monkey cubes
## Why It's Good For The Game

One person can no longer make 1000 monkeys roam the halls

## Changelog
:cl:
add: configurable limit to amount of monkeys that can be spawned by monkey cubes
tweak: handle_death is passed the gibbed variable
tweak: spec_death replaced with handle_death in romerol
/:cl:


code/modules/mob/living/carbon/human/species/monkey.dm
```diff
@@ -46,6 +46,10 @@
 /datum/species/monkey/get_random_name()
 	return "[lowertext(name)] ([rand(100,999)])"
 
+/datum/species/monkey/handle_death(gibbed, mob/living/carbon/human/H)
+	. = ..()
+	SSmobs.cubemonkeys -= H
```
Couls: Issue with handling it on the species is that if they're humanized before they're killed it won't delete them from the list properly. Don't know how to work around this unless we change the expand proc to use a monkey subtype and then overwrite it's destroy proc like TG did

Ty-Omaha: monkey faction?

variableundefined:
Sure you'd like to make the default 32 instead of 64 in TG?

Couls:
Since it's a config option I dunno how much that matters but I can change it if you'd like

code/modules/mob/living/carbon/human/species/monkey.dm
Couls: that'd be pretty snowflake since it's not really the intended use of factions, could have it removing from the carbon/human/death proc but that wouldn't be optimal too because the code would run for every non-monkey as well

variableundefined:
Not sure if code: exists in our changelog

code/modules/mob/living/carbon/human/species/monkey.dm
Fox-McCloud: You could safely put this in `human/death`

It's a needless check 99% of the time, but it will work.

Couls:
Will replace with tweak? Or what tag should I use

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -101,7 +101,7 @@
 		return FALSE
 
 	set_heartattack(FALSE)
-
+	SSmobs.cubemonkeys -= src
```
Couls: This should be fine, if src is not in the list then it won't remove anything

variableundefined:
Also, config conflicts with my previous PR

Couls:
Our expand proc doesn’t create a new monkey, it creates a human and then assigns the species afterwards, I didn’t really want to change how this was done and add additional parameters to the initialize proc on humans I also would have to check when destroy is called on humans but I assume it doesn’t run when monkeys are killed(maybe gibbed?) so it depends on how strict you want the cap( x living monkeys versus x monkeys )

code/modules/mob/living/carbon/human/species/zombies.dm
```diff
@@ -67,7 +67,7 @@
 		playsound(H, pick(spooks), 50, TRUE, 10)
 
 //Congrats you somehow died so hard you stopped being a zombie
-/datum/species/zombie/infectious/spec_death(gibbed, mob/living/carbon/C)
+/datum/species/zombie/infectious/handle_death(gibbed, mob/living/carbon/C)
```
Couls: This is actually a bad idea, the reason this works on TG is because zombies don’t die there due to their nodeath trait, but in our code zombies do die and then revive afterwards. I need to revert this back

code/modules/food_and_drinks/food/foods/meat.dm
```diff
@@ -271,6 +271,12 @@
 	return 1
 
 /obj/item/reagent_containers/food/snacks/monkeycube/proc/Expand()
+	if(LAZYLEN(SSmobs.cubemonkeys) >= config.cubemonkeycap)
+		if(fingerprintslast)
+			to_chat(get_mob_by_ckey(fingerprintslast), "<span class='warning'>Bluespace harmonics prevent the spawning of more than [config.cubemonkeycap] monkeys on the station at one time!</span>")
+		else
+			visible_message("<span class='notice'>[src] fails to expand!</span>")
+		qdel(src)
```
Fox-McCloud: you need an explicit return here.

Yes, you *can* rely on `QDELETED`, but I'd rather explicitly return.

Fox-McCloud: Now that I think about it, why does this have to delete if it fails? Seems like it would kinda suck for a cube to get deleted if you happen to bump up against the limit and aren't being a spammy jerk.

code/modules/research/xenobiology/xenobio_camera.dm
```diff
@@ -219,6 +219,7 @@
 	if(cameranet.checkTurfVis(remote_eye.loc))
 		if(X.monkeys >= 1)
 			var/mob/living/carbon/human/monkey/food = new /mob/living/carbon/human/monkey(remote_eye.loc)
```
Fox-McCloud: you need a check here to prevent the monkey from being created, at all, if the limit has been reached--and some bluspacey message about it too.

code/modules/food_and_drinks/food/foods/meat.dm
Couls: no longer deletes on fail

PR #12298, opened 2019-09-15
Adminhelp update

Ty-Omaha:
Adds the ability for admins to autorespond with common things such as "This is an IC issue" "Thanks" "Clear your cache, this is how you do it". When an autoresponse is sent, the ticket is automatically resolved.

Also changes the neon green color of adminticket stuff to a less obnoxious green.

Plays a new sound when someone opens a ticket. Old Bwoink sound for PMs is kept as is.

Adds "AUTO" on incoming adminhelps. Players should get an auto response instead of "your ticket is resolved" to make sure they don't think they are being ignored.

Video of it https://i.imgur.com/3vuONf9.gifv

Picture of it: (this is sending **both** the admin messages and the ticket opener messages FYI, so it won't be as confusing in this picture)
image 

Sounds _(First one is admin PM, Second is opening a ticket. The audio crackle is not there in the actual files)_: https://streamable.com/h4nx5

_Sounds are taken from a producer on youtube who gave permission in his video_

Credit to CM for the idea

🆑 
add: Admins can now autoreply with common answers to common questions / issues.
soundadd: Opening an adminhelp/mentorhelp ticket will play a sound
tweak: Adminhelp ticket information color changed from neon green to a more toned down green
/ 🆑


TDSSS:
So I can no longer just resolve an ahelp ticket from chat without opening the ticket?

Ty-Omaha:
@TDSSS When you autoreply it will automatically resolve the ticket

TDSSS:
But what if I don't want to autoreply? What if I click on their name, write "we are aware thanks", then want to close it? I need to do an additional click now?

SteelSlayer:
I like everything except the new bwoink sound. Please keep the legacy one.. it's just so iconic to spess station

Ty-Omaha:
@TDSSS In my experience it's more common for admins to just hit resolve and use that to convey that we are looking into it / already aware. This change encourages them to send an auto message which is essentially the same thing, except the player has more of an explanation and feeling that their adminhelp was received and listened to, instead of just telling a player "your ticket is marked as resolved".

variableundefined:
I personally prefer the old(new?) resolve button since it allows me to quickly dismiss trivial ahelp

Ty-Omaha:
Resolve re-added

Breenland:
I don’t like the fact that it autoresolves, honestly. If I could send the autoreply and have it end with a statement such as, “If issues continue, please let us know.” or something along those lines, that’d be nice. I noticed it was on one of the responses, wouldn’t 
I don’t it on most of them.

If I could add a suggestion: something like “There is a large volume of ahelps at this time. We will review this when we have a moment.” Something along those lines.

Mitchs98:
Really dislike the change in the bwoink sound. It sounds nothing like SS13. Current bwoink is iconic and incites fear as it should, not staff but as a player wouldn't want it changed. And frankly the new sound isn't as ear catching.

code/controllers/subsystem/tickets/tickets.dm
```diff
@@ -94,6 +94,8 @@ SUBSYSTEM_DEF(tickets)
 
 	//Inform the user that they have opened a ticket
 	to_chat(C, "<span class='[span_class]'>You have opened [ticket_name] number #[(getTicketCounter() - 1)]! Please be patient and we will help you soon!</span>")
+	var/ticket_open_sound = sound('sound/effects/adminticketopen.ogg')
```
AffectedArc07: Keep the old bwoink sound

code/controllers/subsystem/tickets/tickets.dm
```diff
@@ -111,7 +113,43 @@ SUBSYSTEM_DEF(tickets)
...
+	if((T.staffAssigned && T.staffAssigned != C) || (T.lastStaffResponse && T.lastStaffResponse != C) || T.ticketState != TICKET_OPEN) //if someone took this ticket, is it the same admin who is autoresponding? if so, then skip the warning
+		if(alert(usr, "[T.ticketState == TICKET_OPEN ? "This ticket is already marked as closed or resolved" : "Another admin appears to already be handling this."] Are you sure you want to continue?", "Confirmation", "Yes", "No") != "Yes")
+			return
+	T.assignStaff(C)
+	
+	//try to keep this list where more common things are at the top yeah? thx
+	var/response_phrases = list("Cancel" = "Cancel",
+		"Thanks" = "Thanks, have a Paradise day!", 
+		"Already Resolved" = "The problem has been resolved already.",
+		"Mentorhelp" = "Please redirect your question to Mentorhelp, as they are better experienced with these types of questions.",
```
variableundefined: Not an English expert, but wouldn't "more suitable to" instead of "better experienced" be better phrasing?

Admin does not necessarily have less experience but mentors are generally the better candidate for gameplay question.

code/controllers/subsystem/tickets/tickets.dm
```diff
@@ -111,7 +113,43 @@ SUBSYSTEM_DEF(tickets)
...
+		"Clear Cache" = "To fix a blank screen, please leave the game and clear your byond cache. To clear your Byond Cache, there is a Settings icon in the top right of the launcher. After you click that, go into the Games tab and hit the Clear Cache button. If the issue persists a few minutes after rejoining and doing this please adminhelp again and state you cleared your cache." ,
```
variableundefined: Probably should add a comma between please and adminhelp

code/controllers/subsystem/tickets/tickets.dm
```diff
@@ -111,7 +113,43 @@ SUBSYSTEM_DEF(tickets)
...
+		"IC Issue" = "This is an In Character (IC) issue and will not be handled by admins. You could speak to Security, Internal Affairs, a Departmental Head, Nanotrasen Representetive, or any other relevant authority currently on station.",
```
variableundefined: This is good

code/controllers/subsystem/tickets/tickets.dm
variableundefined: Can the old bwoink sound be kept. Thanks

code/controllers/subsystem/tickets/tickets.dm
Ty-Omaha: They are better experienced, because that's what they do all the time - gameplay mechanic questions - while admins deal with the rules all the time.

code/controllers/subsystem/tickets/tickets.dm
Ty-Omaha: Its dramatic to have a pause between please and something else I'll put it infront of the please

code/controllers/subsystem/tickets/tickets.dm
Ty-Omaha: I assume you commented on the wrong line because this is the ticket opening sound, not the PM reply sound (bwoink) which isn't the new bwoink sound. The old bwoink sound has been kept.

code/controllers/subsystem/tickets/tickets.dm
Ty-Omaha: I'm gonna mark this as resolved so it doesnt take up a bunch of space on the files page but if you want it changed lmk

Ty-Omaha:
Added an option to reject the adminhelp
Made the list sort alphabetically
Made the cancel a real cancel instead of a list entry

variableundefined:
Will merge once I re-review and make sure nothing's wrong

code/controllers/subsystem/tickets/tickets.dm
```diff
@@ -407,6 +449,11 @@ UI STUFF
 		takeTicket(indexNum)
 		showDetailUI(usr, indexNum)
 
+	if(href_list["autorespond"])
+		var/indexNum = text2num(href_list["autorespond"])
+		takeTicket(indexNum)
```
Ty-Omaha: pretty sure this takeTicket is redundant let me test

PR #12301, opened 2019-09-15
The most controversial PR yet

florkeger:
## What Does This PR Do
This PR changes the area name of **Escape** to **Departures** and that's about it. It also adjusts the map editor icon for it accordingly.

## Why It's Good For The Game
Actually, this doesn't particularly change anything special. The reason I think of the Escape Hallway being renamed to Departures is that I, quite frankly, hold the opinion of "Departures" sounding more professional and less threatening to the crew. "Escape Hallway" sounds like it's a permanent evaucation bay that will only be used in emergencies when - in reality - it's meant to be used as the standard exit point for the crew once their tenure is transpired. 

Yes, people are most likely used to say "Escape" but honestly, changing it to Departures just fits the schema of an appropriate area point better from a professional and architectural standpoint. You won't even notice it.

## Changelog
:cl: kazboo
tweak: renamed "Escape Hallway" to "Departures" and the adjacent podbay respectively
/:cl:


Fox-McCloud:
No.

florkeger:
> No.

Owh, why not?

Fox-McCloud:
It's been called "Escape" for literally years. There's no reason to change this just because like 5 people insist on calling it "departures".

florkeger:
> It's been called "Escape" for literally years. There's no reason to change this just because like 5 people insist on calling it "departures".

Change. Well, as far as I'm aware it's been Departures on a few other stations as well. As I mentioned, it's a minor subtle change that doesn't really affect anything outside of people who will just read Departures instead of Escape on Air Alarms and APCs.

Fox-McCloud:
We're not other station. ICly we've called this escape, since inception---I only see a tiny amount of players that insist on calling it "departures". If it *really* doesn't impact much then....why make a PR at all to begin with.

florkeger:
> We're not other station. ICly we've called this escape, since inception---I only see a tiny amount of players that insist on calling it "departures". If it _really_ doesn't impact much then....why make a PR at all to begin with.
> 
> Change, purely for the sake of change is something I will generally be opposed to.

But we do a lot of things like other stations and try to adapt their ways, so I won't really accept that as a counter-argument in this case. Change for the sake of change doesn't always have to include fundamental modifications to the greater part of gameplay of features when it can just be minor adjustments that flavour the detail. How would it be explained that Escape is called Escape? Does NT expect their stations to be pulled into a variety of differing disasters which warrants an entire room dedicated to evacuation? Personally, don't think so.

Sure it doesn't impact a whole lot in the sense of players going like "o wow we were wrong the entire time, it's departures!" or something, it's simply fitting the design better than having it designed as a flat out evac zone.

JohnBigless:
I personally agree with this.
About 50/50, Escape, or should i say Depatrues ends the round via the Crew transfer shuttle arrivin, and not the actual shuttle being called.

variableundefined:
I object.

Drakeven:
I mean, it certainly fits better.

florkeger:
> I object.

Reason being?

MRgaty:
"escape" sounds more like it is only for emergencies, taking in count that the station always gets evacuated after 2 hours, independently of what is going on, and nothing is supposed to go on, I see departures fitting more than escape. As you are, usually, not escaping from anything lore wise.

Fox-McCloud:
img

Upvoting your own PR.

Shame. *Shame*.

variableundefined:
I simply prefer the old name as it has been that way for a long time.

Besides, we kinda accept in canon that we're on a hazardous junkpile of a station that's prone to shit going wrong at any moment. Things explode, dangerous researches are being undertaken.

Surely calling it Escape wouldn't be completely out of character?

taukausanake:
I'm kind of in both teams even though I voted :+1: but this is how I'm seeing it: How often is the "escape" shuttle called verse the crew transfer/"departure" shuttle? I'd vote that would be a good judgement on what the wing should be called.

Also:
> We're not other station

True, Paradise is more of a bastardization of Bay, /TG/, and somethingsomethingfurriesreeee but that's never been a bad thing, _right?_

florkeger:
> img
> 
> Upvoting your own PR.
> 
> Shame. _Shame_.

_Free did it too._



> I simply prefer the old name as it has been that way for a long time.
> 
> Besides, we kinda accept in canon that we're on a hazardous junkpile of a station that's prone to shit going wrong at any moment. Things explode, dangerous researches are being undertaken.
> 
> Surely calling it Escape wouldn't be completely out of character?

Yes, change surely is a thing majorly disliked in Paradise. A lot of people are used to a lot of things that have been existant in a certain way over a long time, but one can always adapt. As I answered to Fox already, you won't even notice the difference of it, it's a mere formality change.

After checking into the lore, I couldn't find anywhere stating that the Cyberiad is in a completely hazardous environment surrounded by dangers that would ultimately suggest mass evacuation in most shifts. The only nearest reference I could find was how antag insurgencies like blob, xenos etc. are, at most, always only the most recent events taking place. I don't think any station engineer that constructs a place of this size would call it's exit transfer point for the crew "Escape" but rather Departures, as this seems more appropriately fitting the circumstances. "Escape", in my opinion, should be nothing but a slang term.

And what do you mean by out of character?

variableundefined:
Closing due to two maintainer objections.

PR #12308, opened 2019-09-15
Changes station goal crate to engi crates

datlo:
This PR changes the station goal crates to engineering crates.

I've seen a few techs confused at to where station goals should be delivered as they come in grey crates, with the engineering paint job it should be clearer.

:cl:
tweak: Station goal crates now come in engineering crates.
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -1252,7 +1252,9 @@ var/list/all_supply_groups = list(supply_emergency,supply_security,supply_engine
 					/obj/item/clothing/suit/beekeeper_suit,
 					/obj/item/melee/flyswatter)
 	cost = 15
+	containertype = /obj/structure/closet/crate/secure/hydrosec
 	containername = "beekeeping starter kit"
+	access = access_hydroponics
 
 /datum/supply_packs/organic/hydroponics/beekeeping_suits
 	name = "2 Beekeeper suits"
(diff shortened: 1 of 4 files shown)


Mitchs98:
I'd personally rather bee crates stay without a lock. Only because Squishy and I sometimes mess with bees in the public garden. And some others do too I believe.

Don't strongly disagree either way but ye.

datlo:
You could get the HOP or a botanist to unlock it, or get botany access. I dont think bees having some oversight is an issue considering holy shit its BEEES AAAAH

Mitchs98:
Yeah I get ya. It's definitely annoying when people that don't know how to handle them get them. Really my only complaint is an extra hurdle to go through. I'm fine with the change.

Fox-McCloud:
I'm ok with the station goals being an engineering subtype with that particular crate, but I'm generally not ok with them being locked. These require massive interdepartmental co-operation; if it's an assistant that spearheads it, then so be it.

Locking the bees crate because they might make a mistake is also kinda questionable; bees hardly wipe out the station and are incredibly easy to deal with; sure, some mistakes will be made; they'll just have to learn.

Birdtalon:
Can't you just unwrench the beehive if it's opened prematurely?

taukausanake:
> Can't you just unwrench the beehive if it's opened prematurely?

It's not that it anchors when the crate is opened. It's that now you have to move multiple things rather than one crate

PR #12311, opened 2019-09-15
Small global midi improvements

Ty-Omaha:
When playing a song when people are in the lobby, that music would be playing overtop of the lobby music. This will now silence the lobby music for those who have admin midis enabled when they are in the lobby listening to lobby music and a midi is played globally.

🆑 
tweak: Global admin midis will now silence lobby music for those who have midis enabled.
/ 🆑

### DIFF
--- a/code/modules/admin/verbs/playsound.dm
+++ b/code/modules/admin/verbs/playsound.dm
@@ -28,9 +28,12 @@ var/list/sounds_cache = list()
 
 	log_admin("[key_name(src)] played sound [S]")
 	message_admins("[key_name_admin(src)] played sound [S]", 1)
+
 	for(var/mob/M in GLOB.player_list)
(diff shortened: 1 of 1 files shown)


variableundefined:
It probably won't improve performance as the macro is just the same AFAIK.

But it is good from a readability and code quality standpoint 👍

Ty-Omaha:
I remember being told to use SEND_SOUND because its more efficient one time 🤷‍♂

Fox-McCloud:
You should listen to Ansari.

```
#define DIRECT_OUTPUT(A, B) A << B
#define SEND_SOUND(target, sound) DIRECT_OUTPUT(target, sound)
```

It doesn't increase performance; it's purely to keep things nice and clean.

Ty-Omaha:
So wouldnt this make it lag more since it has to use a define which has to access the actual `<<` ?

PR #12316, opened 2019-09-16
Wall repair fixes

### LINKED ISSUE #12314: You can't repair damage on walls
**Issue Description**:
You can't repair damage on walls. Video evidence: https://streamable.com/13wpv

To have the wall get repaired

**What happened instead**:
The wall never got repaired

This makes is so you have to dismantle and re-add the metal to girders every time you want to repair a wall, very annoying and time consuming

1. Get a wall into a semi-damaged state
2. Try to repair wall with a welder

Unknown

SteelSlayer:
## What Does This PR Do
It turns out that if the list `dent_decals` that stores bullet hole decals on walls, was `null`, it would cause a runtime and not allow a wall with only structural damage to be repaired. I just added a check to see if the list was null before attempting to call the `Cut()` function which caused the runtime.

I also made it so you can repair and remove the bullet holes from reinforced walls even if there is no structural damage on them.

Fixes  #12314

## Why It's Good For The Game
It allows engineers and others to repair walls properly withing having to dismantle and rebuild them just to get damage to go away. Also it prevents runtimes.

## Changelog
:cl:
fix: Fixes a runtime preventing walls from being repaired properly
add: Adds the ability to repair and remove bullet holes from reinforced walls that have no structural damage
/:cl:

### DIFF
--- a/code/game/turfs/simulated/walls_reinforced.dm
+++ b/code/game/turfs/simulated/walls_reinforced.dm
@@ -40,7 +40,7 @@
 	return ..()
 
 /turf/simulated/wall/r_wall/proc/try_repair(obj/item/I, mob/user, params)
-	if(damage && iswelder(I))
+	if((damage || LAZYLEN(dent_decals)) && iswelder(I))
 		var/obj/item/weldingtool/WT = I
 		if(!WT.remove_fuel(0, user))
 			to_chat(user, "<span class='warning'>You need more welding fuel to complete this task.</span>")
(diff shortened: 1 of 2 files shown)


code/game/turfs/simulated/walls_reinforced.dm
```diff
@@ -50,6 +50,8 @@
 		playsound(src, WT.usesound, 100, 1)
 		if(do_after(user, max(5, damage / 5) * WT.toolspeed, target = src) && WT && WT.isOn())
 			to_chat(user, "<span class='notice'>You finish repairing the damage to [src].</span>")
+			cut_overlay(dent_decals)
+			dent_decals?.Cut()
```
Fox-McCloud: Why is `dent_decals` ever null?

PR #12318, opened 2019-09-16
Makes Wizards Twice As Cool

### LINKED ISSUE #12309: Dual wielding magical staves dont work
**Issue Description**:

If you fire a staff on harm intent while holding another in your off-hand, dual wielding firing will proc as it normally would but the second staff 'whizzles quietly' as if it had no charge, no matter how much charge it actually has

Dual wieldin shootin tootin wizard action

**What happened instead**:

*sad horn*

Feature doesnt work properly, should either be fixed or wizard staff should not trigger dual wielding at all (as you do lose accuracy from it)

1. Hold a stafff in both hands
2. Shoot with one in harm intent
3. You will lose accuracy but the second staff will whizzle instead of firing

Dual wielding PR

**Extra information**:

Only tested locally

Dave-TH:
## What Does This PR Do
Fixes dual wielding with magical weapons. Fixes #12309. The error was that the chambering of the rounds was handled by the afterattack on magical weapons, but the offhand weapon went directly to process_fire which caused problems. Further problems may exist with the dual wielding system regarding things like the clumsy gene and heavy weapons falling out of hands.

## Why It's Good For The Game
Because we definitely need a wizard running around with two staves of chaos. D-Definitely.

## Changelog
:cl:
fix: Magical weapons can now be properly dual-wielded.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/magic.dm
+++ b/code/modules/projectiles/guns/magic.dm
@@ -23,7 +23,6 @@
 	righthand_file = 'icons/mob/inhands/items_righthand.dmi'
 
 /obj/item/gun/magic/afterattack(atom/target, mob/living/user, flag)
-	newshot()
 	if(no_den_usage)
 		var/area/A = get_area(user)
 		if(istype(A, /area/wizard_station))
@@ -37,9 +36,13 @@
 	return charges
 
 /obj/item/gun/magic/newshot(params)
-	if(charges && chambered)
+	if(charges && chambered && !chambered.BB)
 		chambered.newshot(params)
 	return
+
+/obj/item/gun/magic/process_fire()
+	newshot()
+	return ..()
 
 /obj/item/gun/magic/process_chamber()
 	if(chambered && !chambered.BB) //if BB is null, i.e the shot has been fired...


code/modules/projectiles/guns/magic.dm
```diff
@@ -37,10 +36,14 @@
...
 
 /obj/item/gun/magic/newshot(params)
-	if(charges && chambered)
+	if(charges && chambered && !chambered.BB)
 		chambered.newshot(params)
 	return
 
+/obj/item/gun/magic/process_fire()
+	newshot()
+	return ..()
```
Couls: Use . = ..() because it looks bettet

code/modules/projectiles/guns/magic.dm
```diff
@@ -37,10 +36,14 @@
 	return charges
 
 /obj/item/gun/magic/newshot(params)
-	if(charges && chambered)
+	if(charges && chambered && !chambered.BB)
```
Couls: Looks like you added the additional check in the end 😏

code/modules/projectiles/guns/magic.dm
Dave-TH: newshot needs to happen before the call to super due to the weirder points of how the magic weapons works.

code/modules/projectiles/guns/magic.dm
Dave-TH: Yes but it didn't solve the problem.

Couls: Well yes

code/modules/projectiles/guns/magic.dm
Couls: Ughhhhh

TDSSS:
two times zero is still zero, cause wiznerds were never cool *dabs

But yeah nice catch, didn't ever consider dual wielding staves.

MrMagolor:
Let's be glad spellblade still isn't in as a normal purchasable wiz item...

PR #12320, opened 2019-09-16
Updates Status Effects: Fixes Blood Drunk Status Effect, Buffs Regen Core

Fox-McCloud:
Updates Status effects to TG's current; most of this is the same; this just adds support for some new status effect types and adds support for refreshing status effects; neither are actually used in game.

This also ensures the blood drunk buffs uses the proper stun absorption as opposed to utilizing status flags being turned on and off; this also fixes the buff not properly giving you NOSLOWDOWN.

This also reworks the regenerative core a little into a status effect; when you apply it to youreslf, it'll instantly heal all damage, but also give you a 1 minute buff of NOSLOWDOWN.

The regenerative core rework is partially from TG; they nerfed the regenerative core so it's not a full heal anymore; eventually, I'd like to follow suit, but until players get used to and start beating the new Megafauna, it still giving a full revive is intended.

That said, don't expect the cores to stay that way; one way, or another, they're getting nerfed, in the future.

:cl: Fox McCloud
fix: Fixes blood drunk buff not making you slow immune
add: Regenerative core will give you 1 minute of slow immunity on application
/:cl:


code/datums/status_effects/buffs.dm
```diff
@@ -65,8 +66,7 @@
 			H.dna.species.clone_mod *= 0.1
 			H.dna.species.stamina_mod *= 0.1
 		add_attack_logs(owner, owner, "gained blood-drunk stun immunity", ATKLOG_ALL)
-		var/status = CANSTUN | CANWEAKEN | CANPARALYSE | IGNORESLOWDOWN
```
Fox-McCloud: @Fox-McCloud was a moron and was toggling OFF slow immunity

code/datums/status_effects/neutral.dm
```diff
@@ -43,5 +43,5 @@
 	duration = 40
 	alert_type = null
 
-/datum/status_effect/high_five/on_timeout()
-	owner.visible_message("[owner] was left hanging....")
+/datum/status_effect/high_five/on_remove()
```
Fox-McCloud: Having `on_timeout` is literally pointless, since it occurred just before `qdel(src)`, which leads to `on_remove` being called.

MrMagolor:
How DO tg's regenerative cores work? Do they heal over time like the survival medipen?

variableundefined:
I'll look at this tommorow, probably

Fox-McCloud:
TG regen cores give 1 minute of slow immunity----on application they heal 25 brute, 25 burn, and remove all CC+set stamina loss to 0.

PR #12322, opened 2019-09-16
Updates Space Carp

Fox-McCloud:
From TG.

Updates Space Carp; they now have a base sprite that gets color applied to it, instead of having discrete sprites for every single carp.

Megacarp are also buffed; they regen, over time, when they haven't taken damage; they also have randomized stats and names on creation.

img

:cl: Fox McCloud
tweak: carp sprites now support a wider range of colors
tweak: Megacarp names randomized in addition to their stats
/:cl:


taukausanake:
Not bad, but these look way more washed out than the existing sprites. Not sure how I feel about that. Can the contrast be bumped up a little?

PR #12327, opened 2019-09-16
Crayon Fixes

### LINKED ISSUE #12244: Can no longer draw multiple fake runes on a single tile.
**Issue Description**:
One used to be able to draw multiple fake rune on a tile in crayon, now can only draw one fake rune per tile. I believe this feature was removed unintentionally, because you can still attempt to draw multiple runes, and a timer will pop up while drawing, but nothing appears once the drawing is finished.

I expected to be able to draw multiple runes per tile. Alternatively, to just reject the attempt to draw a second rune and not start a needless timer.

**What happened instead**:
I got a timer AND no extra fake rune! It took a whole 3 SECONDS from my life!

The high levels of roleplay achievable by drawing multiple fake runes are no longer possible in the game's current state, and it's clear that everyone is suffering terribly from their loss. Alternatively, if the performance increase of removing the ability to add multiple fake runes has more value, then spawning timers that inevitably fail with no feedback to the player is minor, but still an issue.

Find crayon. Draw fake rune. Switch fake rune number you wish to draw. Try drawing on the same tile. See the deception of the progress bar and lament as your newly founded cult is stunted by artistic limitations.

Pretty sure it was #12167 . It's a side effect of a MUCH NEEDED fix to the blood stacking issue.

**Extra information**:
I understand this is a real non-issue to 99.99% of the playerbase, but I've had a lot of fun scribbling a bunch of fake runes on a single tile, and making silly summoning circles. Enough fun that I felt the need to write an entire issue report about it. Will close if maintainers deem far too trivial an issue.

Dave-TH:
## What Does This PR Do
Crayons and spraycans now allow multiple decals on the same tile again.
Adds an info span class to the messages crayons give to the user as previously they were unformatted.
Crayons now cannot draw more than one drawing at a time.
Fixes #12244

## Why It's Good For The Game
Because of the freedom of artistic expression!
Also because it probably ought to be a bit more difficult to completely fill hallways with graffiti.

## Changelog
:cl:
fix: Crayon drawings can stack again.
fix: Crayons can no longer draw on multiple tiles at the same time, sorry octo-artist!
/:cl:

### DIFF
--- a/code/game/objects/effects/decals/crayon.dm
+++ b/code/game/objects/effects/decals/crayon.dm
@@ -6,6 +6,7 @@
 	layer = MID_TURF_LAYER
 	plane = GAME_PLANE //makes the graffiti visible over a wall.
 	anchored = TRUE
+	mergeable_decal = FALSE // Allows crayon drawings to overlap one another.
 
 
 /obj/effect/decal/cleanable/crayon/Initialize(mapload, main = "#FFFFFF", var/type = "rune1", var/e_name = "rune")
--- a/code/game/objects/items/crayons.dm
+++ b/code/game/objects/items/crayons.dm
@@ -19,6 +19,7 @@
 	var/instant = 0
 	var/colourName = "red" //for updateIcon purposes
 	var/dat
+	var/busy = FALSE
 	var/list/validSurfaces = list(/turf/simulated/floor)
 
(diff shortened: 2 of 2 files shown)


JKnutson101:
You've saved my cult dreams.

PR #12333, opened 2019-09-17
Replace emergency toolbox in hermit ruin by regular

datlo:
This PR replaces the emergency red toolbox in the lavaland hermit ruin by a regular blue one that just contains tools.

The red toolbox contains an handheld station bounced radio, letting the hermit immediately contact the Cyberiad and ask for a rescue. An hermit that isnt actually stranded does not make much sense, and makes for IMO a less interesting/difficult character to play as.

:cl:
tweak: Replaced the red emergency toolbox in hermit ruin by a regular blue toolbox.
/:cl:

### DIFF
--- a/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_hermit.dmm
+++ b/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_hermit.dmm
@@ -50,7 +50,7 @@
 	},
 /area/ruin/powered)
 "l" = (
-/obj/item/storage/toolbox/emergency,
+/obj/item/storage/toolbox/mechanical,
 /turf/simulated/floor/plating/asteroid{
 	name = "dirt"
 	},


Fox-McCloud:
Why not make a subtype of the toolbox that just doesn't have the radio in it? The emergency one makes a lot more sense, tool wise, for what the hermit is doing.

lolman360:
This PR also allows hermit to rearrange their base/make hidden doors and such, which is nice. They lose an extinguisher but they have a bucket so it's balanced out.

datlo:
Make a subtype for this specific item feels unnecessary, the only item the hermit really loses from this trade is the radio (intentional) and the extinguisher, as they have a lantern for light.

Denthamos:
Mmn, while the emergency makes more sense theme wise, gameplay wise a regular toolbox is better so the player in the role can modify their domicile and improve upon it.

PR #12334, opened 2019-09-17
Second Most Important PR

Shadow-Quill:
fixes a typo

`then` > `than`


code/__DEFINES/math.dm
```diff
@@ -1,7 +1,7 @@
 #define PI 3.1415
 #define SPEED_OF_LIGHT 3e8 //not exact but hey!
 #define SPEED_OF_LIGHT_SQ 9e+16
-#define INFINITY 1e31 //closer then enough
```
Shadow-Quill: why

AffectedArc07: what

PR #12337, opened 2019-09-17
Can Move Dense Objects Around Corners

Fox-McCloud:
From TG.

If you push a dense object into a corner, you can wind up in a situation where you're kinda stuck.

With this change, you will be able to start dragging the object in question, click around the corner and move it.

Also uh, non human mobs can't use laser eyes or TK anymore---not that it really mattered anyhow.

Fixes using gloves over-rides that are ranged

:cl: Fox McCloud
add: can move dense objects around corners
/:cl:


TDSSS:
Just think of it as a learning experience when you trap yourself in a corner like that.

PR #12342, opened 2019-09-18
The Ending of an Era. The Turning of the Page

Fox-McCloud:
Lavaland has been around for a while now; it's stable, people enjoy it, and it's been *mostly* brought up to date with TG (still work to do), and even our space/parallax background demonstrates how much of an icon it is on our server.

Either case, this puts to final rest the asteroid mining level; all maps and code related to it are now gone.

It served us well; the amount of memories I built and formed there were great and amazing (especially the science station and Xenoarch), but all things eventually end. It really does makes me sad to see it go, but the map is just going to very slowly languish over time.

:cl: Fox McCloud
del: Completely transitions to Lavaland for all maps. The mining asteroid and it's related bits are removed
/:cl:


_maps/cyberiad.dm
```diff
@@ -21,7 +21,6 @@ z7 = empty
 	#include "map_files\cyberiad\z6.dmm"
 	#include "map_files\generic\z7.dmm"
 
-	#define MINETYPE "lavaland"
```
Shadow-Quill: Rest in peace. You shall be missed.
May you find peace in the mapterlife.

JKnutson101:
o7

Sleep well.

TDSSS:
I'm sad I never got to see the famous xenoarch outpost

KasparoVy:
Man old asteroid, xenoarch and all was such a trip. That was right up my alley, there was just something so cool about mining a damn space rock (bone dry as we added better tools) though for most it really didn't have all that much value.
Considering  the direction the server and game as a whole has progressed (as it needed to) all I got left to say is RIP to the place I spent most my playtime
For every aspect I enjoyed, thinking critically about what modern features deliver outside I just can't knock the new aside from 'muh feels'.

Just gonna have to get my spacerock fix in places still clinging to a legacy, but I'm mining hell does do a hell of a lot better for just about everything else.

Deep 666'd

taukausanake:
So how is this going to work for any downstreams that might want to keep the mining asteroid around? Maybe even use it in a circulation thing like the map rotation (surely rotating two maps with similar shuttles can kill the server). When Lavaland was first added the asteroid was left in as an option and I kind of would like to see that option left in.

Ty-Omaha:
>So how is this going to work for any downstreams that might want to keep the mining asteroid around?

just revert?

variableundefined:
I'm highly doubtful any downstream of ours can handle rotating asteroid in lieu of lavaland when we haven't even pulled that off ourselves yet

TDSSS:
I mean, was there any interest in having a rotating mining level in the first place? Its probably less a question of ability and more of will.

JohnBigless:
I'm glad i got the older versions for archiving purposes.

PR #12347, opened 2019-09-18
Syndie and Binary keys now use their own sprites

AzuleUtama:
## What Does This PR Do
The Syndicate and Binary radio keys have their own sprites labelled in the radio.dmi file, but the code instead makes them call the default sprite for encryption keys. This PR simply points both keys (as well as the syndteam variant) to their unique sprites.

## Why It's Good For The Game
Gives the items a unique look they were presumably meant to always have.

## Images of changes
**Before:**
image

**After:**
image

(Syndie key on the left, Binary on the right)

## Changelog
:cl:
fix: Syndie and Binary keys now use their own sprites rather than the default ones.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -17,7 +17,7 @@
 /obj/item/encryptionkey/attackby(obj/item/W as obj, mob/user as mob, params)
 
 /obj/item/encryptionkey/syndicate
-	icon_state = "cypherkey"
+	icon_state = "syn_cypherkey"
 	channels = list("Syndicate" = 1)
 	origin_tech = "syndicate=1;engineering=3;bluespace=2"
 	syndie = TRUE //Signifies that it de-crypts Syndicate transmissions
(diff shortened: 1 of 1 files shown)


datlo:
I believe the syndi key is intentionally disguised as a 'standard' key but honestly you could change it to just 'Syndicate Encryption Key' and add this sprite, nobody's getting fooled by it

AzuleUtama:
> I believe the syndi key is intentionally disguised as a 'standard' key but honestly you could change it to just 'Syndicate Encryption Key' and add this sprite, nobody's getting fooled by it

Agreed, not like any security officer is going to find it out by just looking at the key when it's never going to leave your headset.

Fox-McCloud:
Wouldn't it be better that they *do* look the same, so you can't just "AHH YES A SYNDICATE KEY" from a split second glance?

AzuleUtama:
> Wouldn't it be better that they _do_ look the same, so you can't just "AHH YES A SYNDICATE KEY" from a split second glance?

You're rarely, if ever, going to find out someone has a syndi key by looking at the key on its own. The only likely scenario in which they'll get caught from that is immediately after they buy it.

Additionally, there isn't currently a standard encryption key for it to mimic - the general channel is built into all radios by default. Meaning that seeing a "standard" key is very likely to mean it's one of the traitor keys as of right now.

PR #12348, opened 2019-09-18
Fixes message resolve in admin/mentorhelps

Ty-Omaha:
allows you to press the resolve button when an a/m help appears and it will actually resolve again

🆑 
fix: Resolve button on adminhelps and mentorhelps work again
/ 🆑


variableundefined:
For the future try to have slightly more descriptive commit name like "Fixes msg resolve".

With the new squash & merge practice your first commit name is likely to show up for future references.

PR #12350, opened 2019-09-19
Autoreply confirmation alert logic fix

Ty-Omaha:
Fixes reversed logic in autorespond confirmation where it would tell the user that the ticket is closed when open or another admin handling it when closed

🆑 
fix: Fixes reversed logic in autorespond confirmation alert
/ 🆑


Ty-Omaha:
someone please review the logic

PR #12353, opened 2019-09-19
Buffs the uplink syringe gun, it now comes with some syringes

TDSSS:
## What Does This PR Do
This PR changes the 'dart pistol' in the uplink into a dart pistol kit, which contains the dart pistol itself, 1 syringe of sarin for killing, 1 syringe of capulettium plus for silencing and 1 syringe of pancuronium as a knockout toxin.

## Why It's Good For The Game
The dart gun is pretty damn useless. On its own it does nothing, requiring you to get nasty chems and syringes for use with it. The issue is, all jobs that have access to nasty chems (medbay and science) also have access to both normal syringe guns and rapid syringe guns. They'd be effectively paying 4 TC for a minor stealth boost, which is in no way worth it.

With this PR, the syringe gun is an actual worthwhile purchase for any traitor looking for an alternative weapon. Other weapons come with some ammo to use with them. However, the kit shouldn't be too overpowered because it's still just 3 syringes, one of which just silences. Considering you only get 2 real shots, it seems still fair at its old cost of 4 TC, but I could be convinced to go to let's say 5 or 6 TC.

## Changelog
:cl:
tweak: The uplink dart gun now comes with some syringes to use with it.
/:cl:

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -925,10 +925,10 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	surplus = 50
 
 /datum/uplink_item/stealthy_weapons/dart_pistol
-	name = "Dart Pistol"
-	desc = "A miniaturized version of a normal syringe gun. It is very quiet when fired and can fit into any space a small item can."
+	name = "Dart Pistol Kit"
+	desc = "A miniaturized version of a normal syringe gun. It is very quiet when fired and can fit into any space a small item can. Comes with 3 syringes, a knockout poison, a silencing agent and a deadly neurotoxin."
 	reference = "DART"
-	item = /obj/item/gun/syringe/syndicate
+	item = /obj/item/storage/box/syndie_kit/dart_gun
 	cost = 4
 	surplus = 50
 	excludefrom = list(/datum/game_mode/nuclear)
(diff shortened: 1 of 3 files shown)


AzuleUtama:
Having done a little looking into each chem, it seems more worthwhile to mix all the chems in each syringe into a beaker and then use them as an equal mix instead of firing them individually. Sarin and pancuronium in particular have very low metabolization rates - you won't need more than about 5 units of each to get the job done.

It might need to be tested first, but having all 3 chemicals pre-mixed into a syringe might be better than just giving them 3 individual syringes. That or possibly giving them the chemicals in bottles and throwing in a box of syringes to emphasis mixing the chemicals as they want.

Either way, it's still an improvement over the original dart pistol when you consider traitors have nothing remotely lethal or debilitating to load into it available on the uplink right now.

TDSSS:
Hmm, part of the reason they're in separate syringes is to not have them be too similar to bioterror syringes, a nukie item. It also gives you a non-lethal option that a mix wouldn't. And if people do want to create their own custom mix that might be more optimal, that's okay with me too. Skill and game knowledge can be rewarded.

AzuleUtama:
Fair point with the bioterror syringes. They can always just take the chemicals out of the syringes if they want to mix them instead, I just figured having them come in bottles makes it a bit less clunky for doing so while still allowing them to fill a syringe with sarin if they desire.

TDSSS:
Bottles hold more than 15u each, so I'd either buff the item more by giving you like 50u each or give you almost empty bottles, which feels kind of inelegant to me.

PR #12355, opened 2019-09-19
Removes custom admin event info verbs

Ty-Omaha:
There are two of these. One for admins, one for everyone. They are too similar and admins always change the wrong one that is admin only. 

There's been no practical use to the intentions of this verb as everyone is in asay anyways, and they will get info via the regular event info verb.

🆑 
del: Removed custom admin event info that only admins could view due to confusion with the regular non-admin verb.
/ 🆑


code/modules/admin/admin_verbs.dm
```diff
@@ -110,8 +110,6 @@ var/list/admin_verbs_event = list(
 	/client/proc/toggle_ert_calling,
 	/client/proc/show_tip,
 	/client/proc/cmd_admin_change_custom_event,
-	/client/proc/cmd_admin_custom_event_info,
-	/client/proc/cmd_view_custom_event_info,
```
AffectedArc07: Why did you remove this

Ty-Omaha: Because those procs don't exist anymore?

AffectedArc07: Oh I can’t read

Issue #10048, opened 2018-10-29
Cortical Borer/Monkey Domination UI Issue


Breenland:
**Problem Description**:
When in a monkey as a cortical borer and in direct control, you can put on a backpack. However, when you release control and then take control again, the backpack is invisible, inaccessible, and unable to be removed as well. I don't think this happens with other smallmobs like farwas, stoks, wolpins, and neareas, but I'm not 100% sure.

**What did you expect to happen**: 
When I relinquish control, then take control again, the backpack should still be visible and able to be interacted with.

**What happened instead**:
When taking control again, the backpack is invisible and unable to be interacted with at all. Using hotkeys still allows you to put stuff into it.

**Why is this bad/What are the consequences:**
If you're a borer attempting to stay alive in a small creature and you somehow lose control, you lose access to any equipment you had unless someone removes the bag for you.

**Steps to reproduce the problem**:
1. Be a cortical borer, infest monkey.
2. Take control of monkey, put on backpack/satchel/similar item.
3. Relinquish control.
4. Take control again. The backpack is invisible.

**When did the problem start happening**:
It's been happening as long as I can remember, though I've always forgotten to make an issue for it.

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was
anything else out of the ordinary happening?)**: As stated before, I don't think it affects other smallmobs, but they also have a full UI compared to monkeys.

PR #12358, opened 2019-09-20
Rebalances Golems, Refactors Species Based Equipping

Fox-McCloud:
This PR does several things, mostly related to golems.

When golems were updated/ported, they had their damage numbers tweaked on them---that said, only the brute and burn mods were adjusted---the `armor` value on TG is supposed to apply to all damage types; as such, golems have been adjusted to have damage reduction for all damage types equal to what their armor would be, on TG. It also corrects some mathematical errors with regards to how much damage reduction a few golems should get.

Additionally, golems are 1 speed faster than they were previously; their limbs also cannot be dismembered or broken; they have lost their virus immunity, however.

Some other minors tweaks, like giving bananium golems the ability to deal at least 1 damage, half the time they punch, fixing herbicides not working on wood golems, and adjusting their light healing proc to be the same as Diona (since Diona were changed).

Also fixes zombies not taking reduced damage. @Couls literally added the `armor` var to species and didn't plug it in anywhere. What the actual heck.

Lastly, this refactors species equipping to allow us to be able to block off specific slots for specific species; this means we can avoid doing hacky things like with golems, whereby they have an invisible "jumpsuit" or "gloves" equipped to them to prevent them from wearing anything else in this slot. There's the ability, however, to allow overrides for this; Golems, for instance, can wear labcoat and can still wear boxing gloves.

If you can't equip something to a slot, as a species, that slot is now a distinctively different shade, to mark it as "inactive".

Additionally, this gives the ability for species to have a pseudo jumpsuit; if the `nojumpsuit` var is set, then that species will be able to have a belt, PDA, ID, and pockets equipped without actually wearing a jumpsuit (not surprisingly, golems have this enabled).

And lastly, but certainly not least, if you have a robotic body, you can wear a belt, PDA, and ID without wearing a jumpsuit. If your left leg is robotic then your left pocket can be utilized without a jumpsuit--likewise for the right leg.

fixes: https://github.com/ParadiseSS13/Paradise/issues/10048

Well, probably.

I also implemented the `PIERCEIMMUNE` trait, which doesn't allow syringes and such to be used on you, but protects again using glass shards, tripping on things like dice, cacti, glass shards, and object embedding; Golems and skeletons have this, by default.

In order to avoid copy-pasta snowflake, this implement's TG's caltrop component; currently glass shards, cacti, D4 dice, and light tubes use this.

:cl: Fox McCloud
tweak: Golems have been buffed. 
tweak: Golems are faster, better protected against damage, and have limbs that cannot break or be dismembered.
tweak: Golems are no longer virus immune
tweak: Golems can now wear labcoats and boxing gloves
fix: Fixes golems being able to be healed/injected on parts other than the chest
tweak: Having a robotic chest allow you to wear an ID, PDA, and belt without a jumpsuit. Having a robotic right leg allows you to use your right pocket without a jumpsuit; same for your left
add: stepping on glass tubes now shatters them; can trip you up if you aren't wearing shoes
add: Stepping on a D4 will deal between 1 and 4 damage to you if you aren't wearing shoes and trip you up
tweak: skeletons have built in pierce immunity
/:cl:


code/modules/mob/living/carbon/human/species/monkey.dm
```diff
@@ -58,44 +59,6 @@
 		H.dna.SetSEState(MONKEYBLOCK, TRUE)
 		genemutcheck(H, MONKEYBLOCK, null, MUTCHK_FORCED)
 
-/datum/species/monkey/handle_can_equip(obj/item/I, slot, disable_warning = 0, mob/living/carbon/human/user)
```
Fox-McCloud: Good riddance.

Fox-McCloud:
Refactored monkey HUDs, too,

There's really no reason to have them around anymore--the different thing about the monkey UI (besides less icons) is that they can't equip to certain slots; the setting of the alpha of slots you can't equip to already is a visual indicator.... the only difference on top of that is that monkeys can use crafting---I added the ability to blacklist certain species from crafting--with monkeys being the only ones who can't.

That should knock out bugs with changelings and other humans that changes HUDs from one to the other.

code/_onclick/hud/monkey.dm
```diff
@@ -1,134 +0,0 @@
-/mob/living/carbon/human/monkey/create_mob_hud()
-	if(client && !hud_used)
-		hud_used = new /datum/hud/monkey(src, ui_style2icon(client.prefs.UI_style), client.prefs.UI_style_color, client.prefs.UI_style_alpha)
-
-/datum/hud/monkey/New(mob/living/carbon/human/owner, var/ui_style = 'icons/mob/screen_white.dmi', var/ui_color = "#ffffff", var/ui_alpha = 255)
```
Fox-McCloud: You won't be missed.

PidgeyThePirate:
Hey Fox, now that we have the vox rework, can we make it so golems don't decay either? I remember having some pain with that earlier but it might have already been fixed.

JKnutson101:
> And lastly, but certainly not least, if you have a robotic body, you can wear a belt, PDA, and ID without wearing a jumpsuit. If your left leg is robotic then your left pocket can be utilized without a jumpsuit--likewise for the right leg.

Is this.... is this an IPC buff??????? What?! 👀 
Or does this not apply to IPCs?

Fox-McCloud:
>Hey Fox, now that we have the vox rework, can we make it so golems don't decay either? I remember having some pain with that earlier but it might have already been fixed.

Not In yet

>Or does this not apply to IPCs?

It applies to IPCs.

Kyep:
Why do golems need to be buffed?
"Because they are like that on TG" is not a valid answer - you need a justification for why its good for Paradise to have stronger golems.

Fox-McCloud:
Because they're a painfully slow race that offers relatively mediocre upsides that you have to ultimately persuade ghosts to play as. Diona were shied away from because they were painfully slow; golems were *just* about as bad--and are still quite slow, even after all these changes.

PR #12359, opened 2019-09-20
Increases timer for blob mouse spawn

Ty-Omaha:
Increases the timer from 60 seconds to 120 seconds regarding when the blob mouse will automatically burst.

Blob is rare as it is, and 60 seconds isn't enough to find a good spot. The rare event is frequently wasted due to bad spawns and just getting roflstomped within 3 minutes. Also increases the announcement timer to compensate.

🆑 
balance: Blob mice now have 120 seconds until bursting instead of 60 and the announcement delay is increased
/ 🆑

### DIFF
--- a/code/modules/events/blob.dm
+++ b/code/modules/events/blob.dm
@@ -1,6 +1,6 @@
 /datum/event/blob
-	announceWhen	= 120
-	endWhen			= 180
+	announceWhen	= 180
+	endWhen			= 240
 
 /datum/event/blob/announce()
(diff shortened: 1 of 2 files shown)


Kyep:
This might not actually be the buff to blob that you think it is.
Without changing the timer on the announcement, it means that the blob has 60 fewer seconds of growth time post bursting before they're announced, crew start looking for them, and they probably get swarmed.

Ty-Omaha:
updated

PR #12360, opened 2019-09-20
Fixes hierophant club and warp cube ignoring area/tele_proof

Kyep:
## What Does This PR Do
The Hierophant club and warp cube (from lavaland) currently allow you to teleport into/out of areas even when they're marked as having teleportation disallowed (via the area/tele_proof flag).
This PR fixes the oversight, so that these items no longer allow you to teleport into or out of areas flagged as tele_proof.

## Why It's Good For The Game
The tele_proof flag exists to prevent players using teleports to largely negate the difficulty of certain areas. Items ignoring this flag is bad, as opens up cheese strategies which allow players to defeat the content in those areas without the risk that was intended.

:cl: Kyep
fix: The Hierophant club and warp cubes (lavaland loot) no longer allow you to teleport into or out of areas that are set to forbid teleports.
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/hierophant_loot.dm
+++ b/code/modules/mining/lavaland/loot/hierophant_loot.dm
@@ -114,6 +114,10 @@
 	if(user.is_in_active_hand(src) && user.is_in_inactive_hand(src)) //you need to hold the staff to teleport
 		to_chat(user, "<span class='warning'>You need to hold the club in your hands to [beacon ? "teleport with it":"detach the beacon"]!</span>")
 		return
+	var/area/A = get_area(user)
+	if(A.tele_proof)
+		to_chat(user, "<span class='warning'>[src] sparks and fizzles.</span>")
+		return
 	if(!beacon || QDELETED(beacon))
 		if(isturf(user.loc))
 			user.visible_message("<span class='hierophant_warning'>[user] starts fiddling with [src]'s pommel...</span>", \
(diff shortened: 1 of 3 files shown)


Qwertytoforty:
For the best, these items could be used as an escape card from anywhere, with one an instant teleport, and made the depot quite easy for me to escape from, should be under the same restrictions as the other teleport items, as is done here.

code/modules/mining/lavaland/loot/hierophant_loot.dm
```diff
@@ -114,6 +114,10 @@
 	if(user.is_in_active_hand(src) && user.is_in_inactive_hand(src)) //you need to hold the staff to teleport
 		to_chat(user, "<span class='warning'>You need to hold the club in your hands to [beacon ? "teleport with it":"detach the beacon"]!</span>")
 		return
+	var/area/A = get_area(user)
+	if(A.tele_proof)
+		to_chat(user, "<span class='warning'>[src] sparks and fizzles.</span>")
+		return
```
Fox-McCloud: This should probably be made into a helper instead of using basically the same thing everywhere.

Kyep: How about the lavaland items just using do_teleport like most teleportation items?
do_teleport already respects tele_proof.
I could make a helper for the cases that don't use do_teleport... but if you want standardization, then just making the lavaland items use do_teleport would probably be better.

Fox-McCloud: These aren't really designed to work with those--not to mention it'll break things if you're using a bag of holding.

Kyep: Fair point. I had not considered miners using bluespace bags. That could get wonky.

Updated to use a helper instead.

PR #12362, opened 2019-09-20
Fixes incorrect names of syndie/antag coins

Kyep:
## What Does This PR Do
Currently, there are two syndicate-like coins in the codebase:
/obj/item/coin/antagtoken named "antag token"  https://github.com/ParadiseSS13/Paradise/blob/21155b69401af8f3e4fe6608dcc5b8612ce6d503/code/modules/mining/ores_coins.dm#L402
and
/obj/item/coin/antagtoken/syndicate named "syndicate coin" https://github.com/ParadiseSS13/Paradise/blob/21155b69401af8f3e4fe6608dcc5b8612ce6d503/code/modules/mining/ores_coins.dm#L410
When you actually spawn these in game though, they don't get their intended names. Their New() proc overrides their name, turning them both into "valid coin" and making them indistinguishable from each other.

This PR fixes this, by adding a name_by_cmineral var (default TRUE, but false for these coins) which stops them being renamed in New() and ensures they get their correct names.
Changing their cmineral wasn't possible because that would result in issues with their icon states (which are based on cmineral).

## Why It's Good For The Game
Fixes a bug that results in syndicate coins getting incorrect names.

:cl: Kyep
fix: Fixed a small bug that caused coins named "syndicate coin" or "antag token" to both incorrectly appear as "valid coin". 
/:cl:

### DIFF
--- a/code/modules/mining/ores_coins.dm
+++ b/code/modules/mining/ores_coins.dm
@@ -322,6 +322,7 @@ GLOBAL_LIST_INIT(sand_recipes, list(\
 	var/string_attached
 	var/list/sideslist = list("heads","tails")
 	var/cmineral = null
+	var/name_by_cmineral = TRUE
 	var/cooldown = 0
 	var/credits = 10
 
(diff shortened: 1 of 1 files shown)


JKnutson101:
There's a 'valid' mineral hmm? Can we make Valid Golems next? 😄

AffectedArc07:
Golems are already valid

**Disclaimer: They are not**

PR #12373, opened 2019-09-21
Ghost HUD QoL change

TDSSS:
## What Does This PR Do
The toggle huds button for ghosts now first goes to ALL huds, then cycles through the other options

## Why It's Good For The Game
Most people use all huds anway, less clicks.

## Changelog
:cl:
tweak: all huds are the first option when ghosts cycle through huds
/:cl:


Markolie:
Honestly all huds should just be on by default when you observe.  I dislike always having to toggle it.

TDSSS:
Sure, why not

Kyep:
Make 'all' the default for ghosts.
There's no reason not to. It is annoying to have to re-activate HUDs every time you ghost.

PR #12382, opened 2019-09-22
mechs now start with a high-capacity power cell+

Evankhell561:
## What Does This PR Do
mechs now start with a high-capacity power cell+ instead basic power cell. The initial charge is the same

## Changelog
:cl:
tweak: mechs now start with a high-capacity power cell+
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -146,14 +146,12 @@
 	internal_tank = new /obj/machinery/portable_atmospherics/canister/air(src)
 	return internal_tank
 
-/obj/mecha/proc/add_cell(var/obj/item/stock_parts/cell/high/plus/C=null)
+/obj/mecha/proc/add_cell(var/obj/item/stock_parts/cell/C=null)
 	if(C)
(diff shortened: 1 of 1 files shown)


code/game/mecha/mecha.dm
```diff
@@ -146,7 +146,7 @@
 	internal_tank = new /obj/machinery/portable_atmospherics/canister/air(src)
 	return internal_tank
 
-/obj/mecha/proc/add_cell(var/obj/item/stock_parts/cell/C=null)
+/obj/mecha/proc/add_cell(var/obj/item/stock_parts/cell/high/plus/C=null)
```
Fox-McCloud: please fix the argument. This is really gross. 

Also, this isn't the proper way too do it, since this proc is utilized for more than just creating the cell.

You'd need to do `cell = new /obj/item/stock_parts/cell/high/plus(src)` and axe the `cell.charge` and `cell.maxcharge` things.

taukausanake:
I don't remember batteries even being part of mech construction. Shouldn't that be done instead?

PR #12388, opened 2019-09-22
Wizard Battlemage Armor

Fox-McCloud:
Ports the Battlemage Armor, from TG.

The wizard's hardsuit is pretty crappy, to be honest; it gives you a bit of armor and the ability to go into space, protects you from fire, and slows you down....that's it. Why bother getting it when you can be a Lich, who's immune to space AND fire and isn't slowed down day to day.

Either case, this reworks the wizard's hardsuit into battlemage armor.

This armor is *partially* space resistant; you won't take pressure damage while wearing it, but you will get cold. It doesn't have quite as much armor as the old wizard hardsuit, but it does have one important caveat; it's a shielded hardsuit. Unlike other shielded hardsuits, this one doesn't recharge; you'll get 16 attacks blocked, and that's it. After the shields deplete, it's just a weaker wizard robes. Oh, did I mention? It doesn't have a slowdown either.

Additionally, for 1 spell point, you can buy additional charges for the suit--it'll give you 8 more shields.

Lastly, this actually ensures that shielded hardsuits have a shield icon, when worn--we should really probably implement TG's `worn_icon` system, but...that'll be a huge pain in the butt to implement since we have species specific clothing---so I opted for a simple work-around that allows players to see this important visual indicator.

I didn't use TG's battlemage armor  sprite, because I didn't think it was that great, especially compared to the gem encrusted armor.

img

:cl: Fox McCloud
tweak: Wizard's hardsuit is now battlemage armor
tweak: Battlemage armor protects you from pressure of space (but not the cold), has no slowdown, and a moderate amount of armor, however, it has 16 shields to block incoming attacks
add: Can purchase additional charges for your battlemage armor for 1 spell point
tweak: Shielded hardsuits now show the shield  on the mob
/:cl:


TDSSS:
Well, that should keep some wizards alive for at least a little bit. I'm not sure i like the suit never recharging

BakKeener:
I can see shotguns absolutely shredding those charges, In fact I can see those charges getting munched fast regardless since it would be better to just have evasion spells like rod form or jaunt. Maybe let them recharge with having a lower max charge count and a much higher recharge time compared to the shielded hardsuit that nukies

variableundefined:
Shotgun will probably still murderize them, yes. But that could be something adjusted later on

TDSSS:
Yeah, I'd keep inm ind there's 5 nukies but only 1 wizard to go up against all of the station.

Fox-McCloud:
Sure, it'll chew through it fast, but this is still better than the regular hardsuit, before---it's not just damage reduction; it's complete damage *mitigation* that can be combined with other evasive/defensive spells.

If it recharges, by itself, I have a bad feeling it's going to wind up with situations that are frustratingly difficult to deal with---think of the nightmare of recharging+blink spam; you'd really only need one really strong spell, like mutate to be pretty much unkillable---go ham when mutate is off cooldown, and if you get in a dicey situation, spam the heck out of blink until your charges come back, then rinse, wash repeat.

Even without recharges, this armor set puts the wizard in an *objectively* better spot than he was before.

PidgeyThePirate:
Why not instead of 16 charges you get 4-8 charges where each charge confers a couple of seconds of invulnerability on the wearer? Enough to block one or two entire shotgun blasts per charge. That differentiates it from the shielded hardsuit and makes hiding and taking potshots at the wizard more meaningful as it depletes their resources more efficiently than going toe-to-toe.

The number of charges would be based on exactly how long the invulnerability period actually is.

TDSSS:
What if you gave them a mechanic to recharge the suit, instead of having it be automatic. Something that means they can't just kite forever. Like feeding it with human sacrifice?

PidgeyThePirate:
If you did that it would have to be fairly stingy. like a couple of charges per sacrifice (or one per sacrifice if using the invuln.)

taukausanake:
This just sounds like a different kind of suck to be honest. You're giving up a little extra armor and full space resistance for limited invulnerability and no slowdown. Why not make the hardsuit a little better?
16 charges does seem fine but the wizard is by no means tanky with this armor. I'd just skip it completely, doesn't seem worth much to me

AzuleUtama:
The lack of slowdown is a big buff compared to the old suit - one of the main reasons you never saw wizards using it was because anyone could outrun them with zero effort. The shield not recharging is also a very good thing when you remember what wizards can do - blink, jaunt, forcewall, magic missile and many other defensive spells combined with recharging invincibility would be even more broken than the current d-esword + shielded suit nuke ops meta.

Even if it doesn't recharge, it's still giving wizards more chances against taser shots and stun batons hits which would otherwise end their round pretty quickly without blink. It also synergizes well with the wizard melee equipment since they can't just be insta-stunned before they reach someone with it.

variableundefined:
@Fox-McCloud That Fox guy's PR conflicted with you. How dare him.

Fox-McCloud:
Wow, what a butt. Maybe we should git ban him.

PR #12396, opened 2019-09-23
A red headcrab day

Terilia:
## What Does This PR Do
_**Rise and shine, Mister Spaceman. Rise and... shine. Not that I... wish to imply you have been graytiding on the job. No one is more deserving of a rest... and all the effort in the world would have gone to waste until... well, let's just say your shift has... come again.**_

_**The right man in the wrong place can make all the difference in the world. So, wake up, Mister Spaceman. Wake up and... smell the toolbox...**_

It is indeed once again time for another headcrab PR. We didn't have enough of those lately. 

I've tested the new headcrabs, poison and fast one included, in a lot of detail on the server and the admin team came to the conclusion that in the current state they were too weak and 'weird'. 

Due to this, I've discussed possible changes with Kyet and present you the meaner, buffer, headcrabs. 

What has changed? 
+Headcrabs have more HP throughout
+Headcrab Nests now exist, which will create a headcrab colony of 15 .
+Headcrabs can now eat dead simple_animals in order to regain a little bit of HP. This also means that they can eat each other's corpses. This should make fighting a swarm more difficult. 
+Headcrabs now have dodging enabled for combat.
+Headcrab will bite and attack people now correctly until they are dead. 
+Zombies received a whopping HP buff of 200. 
+Poison Headcrabs will now blur vision of anyone who they successfully attack. 
+Headcrabs now correctly try to pierce someone's armor and if unsuccessful cannot 'poison' their target. 
-Poison Headcrabs will not inject LSD anymore. 

New Stats: 

Normal Headcrab:
HP: 60
Damage: 5-10

Fast Headcrab:
HP: 40
Damage: 5-10

Poison Headcrab:
HP: 80
Damage: 8-20 (to counterbalance the nerfed poison)

Headcrab Nest:
Spawns 1 Headcrab every 10 seconds to a maximum of 15. 
600 HP
move_resist: MOVE_FORCE_EXTREMELY_STRONG

## Why It's Good For The Game
Saying it bluntly: Headcrabs were not strong enough. During our testing we saw that even in a full fledged assault over half an hour with 30 minutes, only 2 zombies occurred with 200 dead headcrabs. 

## Images of changes
A picture of the nest: 
image
## Changelog
:cl:
add: Headcrab Nests
tweak: Headcrabs can now eat dead simple_animal corpses to regain 10 hp. 
tweak: Poison Headcrabs will not inject LSD anymore.
tweak: Headcrabs have more HP and some do more damage.
tweak: Headcrabs will not stop attacking people after they are unconscious. 
imageadd: Headcrab Nest DMI
/:cl:


SteelSlayer:
Very cool and 10/10 title reference

Carthusia:
I like, perhaps one more headcrab type, as in HL1 its shown headcrabs are born or can be born from a Gonarch, basically a crab that's survived long enough to breed it seems.

In HL1 they spit some kind of acid, charge you and have enough force to basically bash down a wall if they hit it, are heavily armored and birth smaller newborn crabs that just bite, but are not old enough yet to actually take a host.

https://vignette.wikia.nocookie.net/half-life/images/2/2a/Gonarch_model.jpg/revision/latest?cb=20090607012302&path-prefix=en

Perhaps this could be converted into kind of like a nest guard concept, like the purple terror guards for their queen,

Kyep:
Assume:
- Headcrab nests that spawn in maint always go at least 3 minutes before being discovered (likely)
- This ensures that when you encounter them, they always have 15 headcrabs, thus a total of (60*15 = 900) HP and potentially up to 7.5*15 per attack wave damage (just counting normal headcrabs). 
- Further, if someone attacks them, kills four of them, then retreats for a minute, those 4 headcrabs will have respawned by the time they (or someone else) comes back a minute later. Which means nothing but overwhelming force will kill a nest.
- Going after the nest directly isn't viable, as it has 600 HP. You have to kill every crab first, including the ones respawned by the nest.

Balance Conclusion:
- Spawning one headcrab per 10 seconds is spawning way, way too often. Maybe 5 initial crabs plus 1-2 headcrabs per MINUTE would be more appropriate. Otherwise they're always going to kill a bunch of people every time they appear.

code/game/objects/structures/spawner.dm
```diff
@@ -121,4 +121,16 @@
 /obj/structure/spawner/mining/basilisk
 	name = "basilisk den"
 	desc = "A den housing a nest of basilisks, bring a coat."
-	mob_types = list(/mob/living/simple_animal/hostile/asteroid/basilisk)
\ No newline at end of file
+	mob_types = list(/mob/living/simple_animal/hostile/asteroid/basilisk)
+
+/obj/structure/spawner/headcrab
+	name = "headcrab nest"
+	desc = "A living nest for headcrabs. It is moving ominous."
```
Kyep: "ominously."

code/modules/events/event_container.dm
```diff
@@ -181,7 +181,8 @@ var/list/event_last_fired = list()
 		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Revenant", 				/datum/event/revenant, 					150),
 		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Swarmer Spawn", 			/datum/event/spawn_swarmer, 			150, is_one_shot = 1),
 		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Morph Spawn", 				/datum/event/spawn_morph, 				40, is_one_shot = 1),
-		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Disease Outbreak",			/datum/event/disease_outbreak, 			0,		list(ASSIGNMENT_MEDICAL = 150), 1)
+		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Disease Outbreak",			/datum/event/disease_outbreak, 			0,		list(ASSIGNMENT_MEDICAL = 150), 1),
+		new /datum/event_meta(EVENT_LEVEL_MODERATE, "Headcrabs",				/datum/event/headcrabs, 				80)
```
Kyep: Maybe use a zero base chance plus some ASSIGNMENT_SECURITY here, so it won't happen during lowpop where there is no sec to deal with it.

code/modules/events/headcrabs.dm
```diff
@@ -16,15 +16,14 @@
...
-#define LOC_GAMBLING_DEN 19
 
 #define HEADCRAB_NORMAL 0
 #define HEADCRAB_FASTMIX 1
 #define HEADCRAB_FAST 2
 #define HEADCRAB_POISONMIX 3
 #define HEADCRAB_POISON 4
-/*#define HEADCRAB_CLOWN 5 */ //Planned for the future. maybe
+#define HEADCRAB_SPAWNER 5
+/*#define HEADCRAB_CLOWN 6 */ //Planned for the future. maybe
```
Kyep: It is generally not good to add commented-out code.

Also, for the love of all that is holy, please no clown headcrabs.

code/modules/events/headcrabs.dm
```diff
@@ -35,7 +34,7 @@
 
 /datum/event/headcrabs/start()
 
-	location = rand(0,19)
+	location = rand(0,17)
```
Kyep: I'm really not sure how I feel about hardcoding a long list of pre-defined /areas on this event.
What if we use an alternate map which doesn't have one of these areas, and that area is rolled? What happens?

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -5,8 +5,9 @@
 	icon_state = "headcrab"
 	icon_living = "headcrab"
 	icon_dead = "headcrab_dead"
-	health = 40
-	maxHealth = 40
+	health = 60
+	maxHealth = 60
+	dodging = 1
```
Kyep: Has this been tested?

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -28,6 +30,13 @@
 			if(H.stat == DEAD || (!H.check_death_method() && H.health <= HEALTH_THRESHOLD_DEAD))
 				Zombify(H)
 				break
+	for(var/mob/living/simple_animal/K in oview(src, 1)) //Only for corpse right next to/on same tile
```
Kyep: This is going to do a scan of its vicinity every Life() tick.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -167,6 +178,10 @@
 
 /mob/living/simple_animal/hostile/headcrab/poison/AttackingTarget()
 	. = ..()
-	var/mob/living/carbon/C = target
-	if(C.can_inject(null, 0, "head", 0))
-		C.reagents.add_reagent("lsd", 5)
+	if(iscarbon(target) && target.reagents)
+		var/inject_target = pick("chest", "head")
+		var/mob/living/carbon/C = target
+		if(C.stunned || C.can_inject(null, 0, inject_target, 0))
```
Kyep: Unsure if you want to have an else clause for this condition, giving a different message indicating that it failed to pierce armor.
Also, unsure if this will result in 2 messages per attack given that you use . = ..() above.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -144,11 +153,13 @@
 	icon_state = "poison_headcrab"
 	icon_living = "poison_headcrab"
 	icon_dead = "poison_headcrab_dead"
-	health = 60
-	maxHealth = 60
+	health = 80
```
Kyep: Doesn't dodging give >50% chance of immunity to all ranged attacks?
If so, that already pretty much negates half of the weapons you can use against these.
Further buffing their health so they take 8 harmbatons per headcrab to kill means that clearing out a nest of these will take 8*num_headcrabs hits, which could easily be over 100 attacks. Not including any necessary to destroy the nest.

Maybe giving them dodging is a bad idea. Raising their health disproportionately makes them more difficult for assistants, but sec doesn't care as much as it will just use 20 damage/shot lasers.

Giving them dodging massively disadvantages sec trying to fight them, as sec is the only type of crew likely to have ranged weapons, and sec are MEANT to be able to take these out much easier than regular crew.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -167,6 +178,10 @@
...
 	. = ..()
-	var/mob/living/carbon/C = target
-	if(C.can_inject(null, 0, "head", 0))
-		C.reagents.add_reagent("lsd", 5)
+	if(iscarbon(target) && target.reagents)
+		var/inject_target = pick("chest", "head")
+		var/mob/living/carbon/C = target
+		if(C.stunned || C.can_inject(null, 0, inject_target, 0))
+			if(C.eye_blurry < 60)
+				C.AdjustEyeBlurry(10)
```
Kyep: Assuming someone gets 60 eyeblurry from this, how long does that take to wear off?

code/modules/events/headcrabs.dm
```diff
@@ -107,7 +100,7 @@
 
 	var/list/spawn_types = list()
 	var/max_number
-	headcrab = rand(0, 4) //rand(0,x) for the future
+	headcrab = rand(0, 5) //rand(0,x) for the future
```
Kyep: This var might be better named 'headcrab_type' or similar.

code/game/objects/structures/spawner.dm
```diff
@@ -121,4 +121,16 @@
...
+	icon_state = "headcrab_nest"
+	icon = 'icons/mob/headcrab.dmi'
+	max_integrity = 600
```
Kyep: This is probably too high.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
variableundefined: From what I check, they don't give immunity to ranged attack but rather make them engage in erratic movement.

Health wise seems to be fine

code/modules/mob/living/simple_animal/hostile/headcrab.dm
variableundefined: Believe this is 20 seconds. Same as green spider

Terilia: 60 is the same as green spider, yup

code/modules/mob/living/simple_animal/hostile/headcrab.dm
Terilia: I honestly have no better solution to this currently.

Terilia:
All changes have been implemented except: https://github.com/ParadiseSS13/Paradise/pull/12396#discussion_r327395063 because I honestly don't know how I should approach this.

Terilia:
Resolved Conflict, merged everything requested - Review pls <3 and lets go.

PR #12399, opened 2019-09-23
Fixes Sequencing on Mob Spawner Safeties

### LINKED ISSUE #12391: Accepting a Lavaland ghost role, and having another player accept the same role before you pick your race, sticks you into Null Space.
**Issue Description**:
If you try to pick a lavaland role, the prompt to accept 'no-respawns' and then wait on the 'choose your race' selection, and then a second player accepts the same lavaland role and finishes the racial choice before you, upon accepting you are placed into Null Space.

I expected to either lose out of the choice or reserve it once I pressed yes, not disappear.

**What happened instead**:
null_space

Null Space is very bad. Had to AHelp to become real.

Try to join lavaland ghost role. Accept the first prompt and don't accept the second prompt. Have another player join on the same role, and accept both prompts before you. Accept the prompt now. Appear in Null Space.

#12346 

**Extra information**:
I think it can fixed if the order of the prompts is reversed. No idea for sure though.

Dave-TH:
## What Does This PR Do
Fixes some improper ordering of safety checks in the mob spawners. Fixes #12391 

## Changelog
:cl:
fix: Mob spawners will no longer sometimes send players to the void.
/:cl:

### DIFF
--- a/code/modules/awaymissions/corpse.dm
+++ b/code/modules/awaymissions/corpse.dm
@@ -48,12 +48,13 @@
 	if(!O.can_reenter_corpse)
 		to_chat(user, "<span class='warning'>You have forfeited the right to respawn.</span>")
 		return
-	if(QDELETED(src) || QDELETED(user))
-		return
 	var/ghost_role = alert("Become [mob_name]? (Warning, You can no longer be cloned!)",,"Yes","No")
-	if(ghost_role == "No" || !loc)
+	if(ghost_role == "No")
 		return
 	if(!species_prompt())
+		return
+	if(!loc || !uses || QDELETED(src) || QDELETED(user))
+		to_chat(user, "<span class='warning'>The [name] is no longer usable!</span>")
 		return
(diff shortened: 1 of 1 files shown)


code/modules/awaymissions/corpse.dm
```diff
@@ -51,10 +51,12 @@
 	if(QDELETED(src) || QDELETED(user))
 		return
 	var/ghost_role = alert("Become [mob_name]? (Warning, You can no longer be cloned!)",,"Yes","No")
-	if(ghost_role == "No" || !loc)
+	if(ghost_role == "No")
 		return
 	if(!species_prompt())
 		return
+	if(!loc || !uses)
+		to_chat(user, "<span class='warning'>The [name] is no longer usable!</span>")
```
Fox-McCloud: this doesn't even return.

Dave-TH:
Review concerns addressed, moved the qdel control structures down the chain to avoid replicating code.

PR #12401, opened 2019-09-23
Fixes durathread mutating into durathread, reorganises seed vendor

AzuleUtama:
## What Does This PR Do
Simple fix for durathread after I noticed a bug with it during a round. It will no longer mutate into more durathread if you use mutagen on it.

This PR also re-alphabetises the seed vendor by the product names it displays, intended as minor QOL change to make it easier to find specific seeds. Contraband will still slot into the bottom of the list, but that group is also re-alphabetised.

## Why It's Good For The Game
Fixes people potentially resetting their durathread's growth status when mutating it for better stats, makes it a bit easier to find seeds in the vendor.

## Changelog
:cl:
tweak: Reorganises products in seed vendor.
fix: Fixes Durathread mutating into more Durathread
/:cl:


Fox-McCloud:
The durathread mutates into the ----Durathread?

PR #12402, opened 2019-09-23
Sword Shade Fix

### LINKED ISSUE #12395: Possessed Sword takes damage from ash storms
**Issue Description**:

The possessed sword takes damage from ash storms and from what I am guessing then other damage sources.

For the sword to not get damaged and die from an ash storm.

**What happened instead**:

The sword died and the ghost inside the sword was ejected.

This makes the weapon as lavaland loot a bit useless as such since a single ash storm will kill your sword.  Especially if you yourself are immune to ash storms from certain armors but your sword on your person will still take damage.

Get an possessed sword and have a ghost in control of it, have out in a ash storm, the ghost inside the door will die in the storm and get ejected from the sword.

Probably has been in for a while but really only noticed now with lavaland

**Extra information**:

From what I hear its people saying the sword works off of PAI mob code.

Dave-TH:
## What Does This PR Do
The ghosts inside of possessed swords now have godmode enabled to prevent scenarios where they are getting damaged when they should not be. The shade will still be destroyed if the blade is destroyed, but this will prevent scenarios where the blade is unharmed but the ghost inside mysteriously dies. Fixes #12395

## Changelog
:cl:
fix: The ghosts inside possessed blades are now invincible and will not die to weird edge cases.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/shade.dm
+++ b/code/modules/mob/living/simple_animal/shade.dm
@@ -46,3 +46,7 @@
 /mob/living/simple_animal/shade/sword
 	universal_speak = 1
 	faction = list("neutral")
+
+/mob/living/simple_animal/shade/sword/Initialize(mapload)
+	.=..()
+	status_flags |= GODMODE


code/modules/mob/living/simple_animal/shade.dm
```diff
@@ -46,3 +46,7 @@
 /mob/living/simple_animal/shade/sword
 	universal_speak = 1
 	faction = list("neutral")
+
+/mob/living/simple_animal/shade/sword/Initialize()
```
Fox-McCloud: needs the mapload argument.

code/modules/mob/living/simple_animal/shade.dm
```diff
@@ -46,3 +46,7 @@
 /mob/living/simple_animal/shade/sword
 	universal_speak = 1
 	faction = list("neutral")
+
+/mob/living/simple_animal/shade/sword/Initialize()
+	..()
```
Fox-McCloud: `. = ..()`

Dave-TH:
Review concerns addressed

PR #12405, opened 2019-09-24
Fixes space syndis being unable to navigate space

### LINKED ISSUE #12403: Syndicate Depot Backup is Non-Hostile.
**Issue Description**:
The Syndicate Back-up agents that surround the suspicious depot once it goes red alert, are non-hostile.

The Backups to chase and attack, and punish me for going red alert.

**What happened instead**:
They stood there and didn't even fight back. Could even push off of them like they're made of cardboard or something.

Pretty big consequence of sending depot to red is completely nullified. 

Get the Depot to call red alert. Go outside and pester a floating operative and watching him not attack or even defend himself.

Maybe sometime around the simple mob refactor. #12325 
Or perhaps when the depot was last updated. #11677

**Extra information**:
Encountered on live and then double checked on private server.

Kyep:
## What Does This PR Do
Fixes #12403 
Fixes NPC syndies in red hardsuits (nuke op suits) being unable to move in space.

Fixes NPC bots thinking that they can path through airlocks that are completely welded shut.

Refactors some safety checks to make them more reliable, which prevents runtimes in syndicate.dm in rare situations (like admins manually spawning depot mobs outside of the correct /area).

## Changelog
:cl: Kyep
fix: Fixed NPC redsuited syndicates from being unable to move in space.
fix: Fixed NPC bots thinking that they can path through airlocks which are welded shut.
refactor: Refactors some safety checks so that its impossible for depot mobs to generate runtimes in rare cases (e.g: admins spawning them outside the depot for testing)
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/syndicate.dm
+++ b/code/modules/mob/living/simple_animal/hostile/syndicate.dm
@@ -288,7 +288,7 @@
 	alert_on_spacing = FALSE
 
 /mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/space/Process_Spacemove(var/movement_dir = 0)
-	return
+	return TRUE
 
 
 
@@ -302,7 +302,7 @@
 	loot = list(/obj/effect/mob_spawn/human/corpse/syndicatecommando, /obj/item/melee/energy/sword/saber/red, /obj/item/shield/energy)
 
 /mob/living/simple_animal/hostile/syndicate/melee/space/Process_Spacemove(var/movement_dir = 0)
-	return
+	return TRUE
 
 
 /mob/living/simple_animal/hostile/syndicate/ranged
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/hostile/syndicate.dm
```diff
@@ -116,7 +116,9 @@
 
 /mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/Aggro()
 	. = ..()
-	if(target && istype(depotarea))
+	if(!istype(depotarea))
+		return .
```
variableundefined: Explicit return not necessary as 
``` . = ..() ``` Is already done which already return the super call

File: code/modules/mob/living/simple_animal/hostile/syndicate.dm [hunk 2]
```diff
@@ -148,11 +150,13 @@
 
 /mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/handle_automated_movement()
 	. = ..()
+	if(!istype(depotarea))
+		return .
```
variableundefined: Same

code/modules/mob/living/simple_animal/hostile/syndicate.dm
Kyep: Changed from "return ." to "return".

code/modules/mob/living/simple_animal/hostile/syndicate.dm
Kyep: Changed from "return ." to "return".

PR #12413, opened 2019-09-25
Add bitshift and squash merge instructions

variableundefined:
## What Does This PR Do

As per title. Add instructions about bitflags and that we prefer bitshifts be used to represent them, like TG, for readability reason.

Also add that we now prefer squash merge to keep commit history clean for git blame and other purpose.

## Why It's Good For The Game

Better code maintainability

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -327,6 +327,21 @@ This prevents nesting levels from getting deeper then they need to be.
   * WRONG: list(a = "b")
   * RIGHT: list("a" = "b")
 
+#### Bitflags
+* We prefer using bitshift operators instead of directly typing out the value. I.E.
+    ``` 
+    #define MACRO_ONE (1<<0)
+    #define MACRO_TWO (1<<1)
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -327,6 +327,21 @@ This prevents nesting levels from getting deeper then they need to be.
...
+    ``` 
+    #define MACRO_ONE (1<<0)
+    #define MACRO_TWO (1<<1)
+    #define MACRO_THREE (1<<2)
+    ```
+    Is preferable to
+    ``` 
+    #define MACRO_ONE 0
+    #define MACRO_TWO 1
+    #define MACRO_THREE 2
```
moxian: These are no equivalent. The first group of defines define 1 2 4, and the second one (as you can see) 0 1 2 .
I guess you meant 
```
    #define MACRO_ONE 1
    #define MACRO_TWO 2
    #define MACRO_THREE 4
```
?

variableundefined: Woops. Let me fix that.

PR #12415, opened 2019-09-25
Nerfs Book of Babel

Kyep:
## What Does This PR Do
Alters the mining loot item "Book of Babel" so that instead of granting you EVERY language, it only grants you non-hivemind languages.

## Why It's Good For The Game
Currently, the Book of Babel lets you spy on Xeno/Borer/Abductor/TerrorSpider/etc hivemind chat.
This completely breaks all of those antags' ability to be stealthy, or to co-ordinate as a team.
With this PR in place, the book still grants you standard crew racial languages for crew races, even unusual languages like chimp, but won't grant you the ability to single-handedly wreck most of the late-round antags by spying on their hivemind and relaying their thoughts to the crew.

:cl: Kyep
balance: Nerfed Book of Babel so it is no longer wrecking the balance of late-round antags.
/:cl:

### DIFF
--- a/code/modules/mob/language.dm
+++ b/code/modules/mob/language.dm
@@ -773,8 +773,13 @@
 	flags = RESTRICTED
 	syllables = list("BRAAAAAAAAAAAAAAAAINS", "BRAAINS", "BRAINS")
 
-/mob/proc/grant_all_languages()
+/mob/proc/grant_all_nonhivemind_languages()
 	for(var/la in GLOB.all_languages)
-		add_language(la)
+		var/datum/language/new_language = GLOB.all_languages[la]
+		if(!istype(new_language) || (new_language in languages))
+			continue
+		if(new_language.flags & HIVEMIND)
+			continue
+		languages |= new_language
 
 #undef SCRAMBLE_CACHE_LEN
--- a/code/modules/mob/language.dm
+++ b/code/modules/mob/language.dm
@@ -454,7 +454,7 @@
 	exclaim_verb = "chitters"
 	colour = "alien"
 	key = "y"
-	flags = RESTRICTED | HIVEMIND
+	flags = RESTRICTED | HIVEMIND | NOBABEL
 	follow = TRUE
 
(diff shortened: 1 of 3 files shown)


MrMagolor:
Does Grey psionic communications count as a hivemind chat? Would this PR remove the ability to speak in it?

Kyep:
@MrMagolor Yes, and yes.

Fox-McCloud:
This won't ruin abductors, since they speak to only members of their own team--and you get team assigned at round-start; the default team, I believe is an empty team that never gets filled, so removing that is pointless----and grey should absolutely not be immune to this, as they're not an antag race.

MrMagolor:
>This won't ruin abductors, since they speak to only members of their own team--and you get team assigned at round-start; the default team, I believe is an empty team that never gets filled, so removing that is pointless

Actually, would it have a use in communicating with alien-alloy golems?

Leanfrog:
Alloy golems have a second mind link separate  from the normal language oddly enough, not sure why they have their own and not just a general link, I noticed this when I found the book for the first time

code/modules/mob/language.dm
```diff
@@ -773,8 +773,13 @@
 	flags = RESTRICTED
 	syllables = list("BRAAAAAAAAAAAAAAAAINS", "BRAAINS", "BRAINS")
 
-/mob/proc/grant_all_languages()
+/mob/proc/grant_all_nonhivemind_languages()
 	for(var/la in GLOB.all_languages)
-		add_language(la)
+		var/datum/language/new_language = GLOB.all_languages[la]
+		if(!istype(new_language) || (new_language in languages))
```
Markolie: I don't think these sanity checks are necessary. The all_languages list only contains languages (if it doesn't, we should fix the cause instead of adding this sanity check) and the |= list addition already makes sure the mob's languages won't contain duplicates.

Kyep:
Changed:
- No longer blocks gray telepathy
- Removed check pointed out by Mark

PR #12418, opened 2019-09-25
Cult Pylon Changes

MrMagolor:
## What Does This PR Do

This PR edits cult pylons. First of all, it now ignores unsimulated walls when corrupting tiles (looking at you, white ship cult).

Secondly, pylons now turn simulated walls into cult walls. This is not as big a deal as you may think, as if you're using pylons on station, people will notice the floors anyways, regardless of walls. It also makes more sense in general - if pylons are capable of corrupting large chunks of metal on the floor, why wouldn't the same apply for walls?
## Why It's Good For The Game

It fixes an annoying bug, and makes pylons work more in line with their corruption functionality.

## Changelog
:cl:
fix: Cult pylons no longer delete unsimulated walls and turn them into floors.
tweak: Cult pylons now turn simulated walls into cult walls.
/:cl:


code/game/gamemodes/cult/cult_structures.dm
```diff
@@ -167,11 +167,13 @@
...
-	/turf/simulated/floor/plating/lava,
-	/turf/simulated/floor/chasm,
-	/turf/simulated/wall,
+    /turf/simulated/floor/engine/cult,
+    /turf/space,
+    /turf/simulated/floor/plating/lava,
+    /turf/simulated/floor/chasm,
+    /turf/simulated/wall/cult,
+    /turf/simulated/wall/cult/artificer,
+    /turf/unsimulated/wall
```
MrMagolor: Dunno what happened here - I doubt it causes any issues, though, seeing as I was able to test it successfully.

PR #12421, opened 2019-09-26
Mining borg QoL update and fix

### LINKED ISSUE #12384: GPS location status not applied to mining borg
**Issue Description**:
Mining borg does not have the GPS coordinates shown in the status tab.

Mining borg to have the GPS coordinates shown in the status tab like anyone else with a GPS.

**What happened instead**:
GPS coordinates not shown in status tab

My 'Z' key is breaking from all the spamming

1. Be mining borg
2. Break your 'Z' key when trying to take shelter from an ash storm.

Just noticed it today.

### LINKED ISSUE #12407: Borg proto-kinetic accelerator fires at an ungodly speed.
**Issue Description**:
 The default mining borg KA fires at an ungodly speed.

The KA to be normal like every other one.

**What happened instead**:
You fire at the speed or faster than a fully upgraded KA with cooldown modules.

You can fire the same speed no matter the pressure, your power is drained absurdly fast as well. Can be used to cheese mobs such as Terrors. 

Be borg, choose mining module, fire KA.

Probably sometime during the lavaland updates, I don't know.

**Extra information**:
Not sure if it's intended or not. But i'm leaning towards unintended.

SteelSlayer:
## What Does This PR Do

Adds GPS coordinates under the status tab for cyborgs which have a GPS. Also fixes the mining cyborg's ability to fire their KA with no cooldown. 

Unfortunately I had to snowflake an `istype` check for KAs into the `robocharge()` proc . This proc is used when borgs fire energy weapons, and is used to subtract the appropriate power cost of the shot from their battery. This instantly recharges the borg's KA which can then be fired again immediately.

Since the KA is a subtype of energy weapon, and that proc needs to be where it is, I'm not sure if there is another fix for it. Open to any help/suggestions on this.

Fixes #12407
Closes #12384

## Why It's Good For The Game

For the first change: human miners carrying GPSs already have their GPS coordinates displayed under the status tab so I think its reasonable to give cyborg miners one as well. Second change is a bug fix.

## Images of changes

Where the cyborg GPS text will be placed:

GPS

## Changelog
:cl:
add: Added GPS coordinates under that status tab for cyborgs which have a GPS
fix: You can no longer fire mining cyborg's KA with no cooldown
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -111,6 +111,13 @@
 	chambered = null //either way, released the prepared shot
 	newshot()
 
+/obj/item/gun/energy/kinetic_accelerator/process_chamber()
+	if(chambered && !chambered.BB)
+		var/obj/item/ammo_casing/energy/shot = chambered
+		power_supply.use(shot.e_cost)
+	chambered = null
+	newshot()
+
 /obj/item/gun/energy/proc/select_fire(mob/living/user)
 	select++
 	if(select > ammo_type.len)
@@ -198,8 +205,7 @@
 		if(R && R.cell)
 			var/obj/item/ammo_casing/energy/shot = ammo_type[select] //Necessary to find cost of shot
 			if(R.cell.use(shot.e_cost)) 		//Take power from the borg...
-				if(!istype(src, /obj/item/gun/energy/kinetic_accelerator/cyborg)) // Mining borg KA's can be spammed with no cooldown without this check
-					power_supply.give(shot.e_cost)	//... to recharge the shot
+				power_supply.give(shot.e_cost)	//... to recharge the shot
 
 /obj/item/gun/energy/proc/get_external_power_supply()
 	if(istype(loc, /obj/item/rig_module))
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -198,7 +198,8 @@
 		if(R && R.cell)
 			var/obj/item/ammo_casing/energy/shot = ammo_type[select] //Necessary to find cost of shot
 			if(R.cell.use(shot.e_cost)) 		//Take power from the borg...
-				power_supply.give(shot.e_cost)	//... to recharge the shot
+				if(!istype(src, /obj/item/gun/energy/kinetic_accelerator/cyborg)) // Mining borg KA's can be spammed with no cooldown without this check
+					power_supply.give(shot.e_cost)	//... to recharge the shot
 
 /obj/item/gun/energy/proc/get_external_power_supply()
 	if(istype(loc, /obj/item/rig_module))
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -111,13 +111,6 @@
 	chambered = null //either way, released the prepared shot
 	newshot()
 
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/guns/energy.dm
```diff
@@ -198,7 +198,8 @@
 		if(R && R.cell)
 			var/obj/item/ammo_casing/energy/shot = ammo_type[select] //Necessary to find cost of shot
 			if(R.cell.use(shot.e_cost)) 		//Take power from the borg...
-				power_supply.give(shot.e_cost)	//... to recharge the shot
+				if(!istype(src, /obj/item/gun/energy/kinetic_accelerator/cyborg)) // Mining borg KA's can be spammed with no cooldown without this check
```
Fox-McCloud: This is such a ridiculous and egregious violation of OOP.

this needs to be addressed in a more fundamental way. This is absolutely not the way to do it.

SteelSlayer: I know this is a bad fix and I don't know *how* to fix it other than doing some drastic change. Since KAs are subtypes of energy weapons, it passes through the `process_chamber()` proc on line 106 of the same energy.dm file and the `robocharge()` proc is called inside here.

As far as I can tell, you either make a check before calling `robocharge()` or you make a check inside of it, like I did. I don't see another way of doing this unless you made the borg KA not a subtype of energy weapon or something like that.

SteelSlayer:
Ok, I removed the `istype` check. I created a new overridden function for `process_chamber()` specifically for KAs which do not call the `robocharge()` function. I *hope* this is a better fix, and is less hacky.

code/modules/projectiles/guns/energy.dm
```diff
@@ -111,6 +111,13 @@
 	chambered = null //either way, released the prepared shot
 	newshot()
 
+/obj/item/gun/energy/kinetic_accelerator/process_chamber()
```
Fox-McCloud: it'd be better to overload the `robocharge` proc for kinetic accelerators. They're not supposed to use a borg's internal cell, anyway.

Also, why are you doing this here as opposed to under kinetic accelerators?

SteelSlayer: Ok I'll look at overloading robocharge and making it so it doesn't pull power from their cell. 

My reasoning for putting it there was to just to have both `process_chamber` procs in the same place. I suppose I should put the overloaded robocharge proc in the KA file then.

SteelSlayer: Wait, hold up. The `robocharge` proc is actually the problem here for the bug I'm trying to fix. If we let the proc run on the miner cyborg, it will recharge their KA prematurely and they will be able to spam it at the speed of other weapons like autorifles, revolvers, etc.

And if we're not using the proc to recharge their KA or pull from their battery, then the `robocharge` function would be useless to call, as it does nothing else.

Fox-McCloud: And that's precisely what will accomplish that.

```
/obj/item/gun/energy/kinetic_accelerator/robocharge()
	return
```

Now it doesn't draw from the internal cell and uses its own cell, like it should. It's no longer insanely faster and it no longer drains cyborgs after a couple of shots, which is what it's supposed to do.

SteelSlayer: Ack, such a simple yet effective solution.. will change that

PR #12423, opened 2019-09-26
Diona nymphs can eat veggies

PidgeyThePirate:
## What Does This PR Do
Pretty much what it says in the title. Gives Diona nymphs the option of eating vegetables instead of uprooting weeds in order to grow. Gained nutrition is based on the plant matter and/or nutriment content. Nymphs can't eat more than they need to evolve to prevent griefing by overeating botany products.

## Why It's Good For The Game
Adds an option for Diona nymphs in case weeds are not readily available (usually due to overzealous Earthsblood usage).

## Changelog
:cl:
add: Diona nymphs can now consume veggies as well as weeds in order to gain nutrition and grow.
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -182,3 +182,11 @@
 		user.put_in_hands(T)
 		to_chat(user, "<span class='notice'>You open [src]\'s shell, revealing \a [T].</span>")
 	qdel(src)
+
+// Diona Nymphs can eat these as well as weeds to gain nutrition.
+/obj/item/reagent_containers/food/snacks/grown/attack_animal(mob/living/simple_animal/M)
+	if(istype(M, /mob/living/simple_animal/diona))
+		var/mob/living/simple_animal/diona/D = M
+		D.consume(src)
+	else
+		..()
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -189,4 +189,4 @@
 		var/mob/living/simple_animal/diona/D = M
 		D.consume(src)
 	else
-		..()
+		return ..()
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -185,7 +185,7 @@
 
 // Diona Nymphs can eat these as well as weeds to gain nutrition.
 /obj/item/reagent_containers/food/snacks/grown/attack_animal(mob/living/simple_animal/M)
-	if(istype(M, /mob/living/simple_animal/diona))
+	if(isnymph(M))
 		var/mob/living/simple_animal/diona/D = M
 		D.consume(src)
 	else
--- a/code/modules/mob/living/simple_animal/friendly/diona.dm
+++ b/code/modules/mob/living/simple_animal/friendly/diona.dm
@@ -97,14 +97,14 @@
 		if(nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
 			to_chat(src, "<span class='warning'>You're too full to consume this! Perhaps it's time to grow bigger...</span>")
 		else
-			if(do_after(src, 20, target = A))
+			if(do_after_once(src, 20, target = A))
 				visible_message("[src] ravenously consumes [A].")
-				playsound(src.loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 1.5)
+				playsound(loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 1.5)
 				var/obj/item/reagent_containers/I = A
 				if(I.reagents.get_reagent_amount("nutriment")+I.reagents.get_reagent_amount("plantmatter") < 1)
-					nutrition += 2
+					adjust_nutrition(2)
 				else
-					nutrition += (I.reagents.get_reagent_amount("nutriment")+I.reagents.get_reagent_amount("plantmatter"))*2
+					adjust_nutrition((I.reagents.get_reagent_amount("nutriment")+I.reagents.get_reagent_amount("plantmatter"))*2)
 				qdel(A)
 	else
 		..()
--- a/code/modules/mob/living/simple_animal/friendly/diona.dm
+++ b/code/modules/mob/living/simple_animal/friendly/diona.dm
@@ -93,15 +93,15 @@
 	if(isdiona(A) && (src in A.contents)) //can't attack your gestalt
 		visible_message("[src] wiggles around a bit.")
 	// Nymphs can consume grown food as well as weeds.
-	else if (istype(A, /obj/item/reagent_containers/food/snacks/grown))
-		if (nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
+	else if(istype(A, /obj/item/reagent_containers/food/snacks/grown))
+		if(nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
 			to_chat(src, "<span class='warning'>You're too full to consume this! Perhaps it's time to grow bigger...</span>")
 		else
 			if(do_after(src, 20, target = A))
 				visible_message("[src] ravenously consumes [A].")
 				playsound(src.loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 1.5)
 				var/obj/item/reagent_containers/I = A
-				if (I.reagents.get_reagent_amount("nutriment")+I.reagents.get_reagent_amount("plantmatter") < 1)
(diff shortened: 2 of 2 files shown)


PidgeyThePirate:
Hope I did this right.

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -92,6 +92,20 @@
 /mob/living/simple_animal/diona/UnarmedAttack(var/atom/A)
 	if(isdiona(A) && (src in A.contents)) //can't attack your gestalt
 		visible_message("[src] wiggles around a bit.")
+	// Nymphs can consume grown food as well as weeds.
+	else if(istype(A, /obj/item/reagent_containers/food/snacks/grown))
+		if(nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
+			to_chat(src, "<span class='warning'>You're too full to consume this! Perhaps it's time to grow bigger...</span>")
+		else
+			if(do_after(src, 20, target = A))
```
farie82: do_after_once might fit better here

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -92,6 +92,20 @@
...
+				visible_message("[src] ravenously consumes [A].")
+				playsound(src.loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 1.5)
```
farie82: Just loc. No src.

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -92,6 +92,20 @@
...
+				var/obj/item/reagent_containers/I = A
+				if(I.reagents.get_reagent_amount("nutriment")+I.reagents.get_reagent_amount("plantmatter") < 1)
+					nutrition += 2
```
farie82: I think there was a proc to add nutrition. Best to use that

Ty-Omaha:
so they're cannibals

PidgeyThePirate:
Only if you consider humans eating cows cannibalism.

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -92,6 +92,20 @@
...
 	if(isdiona(A) && (src in A.contents)) //can't attack your gestalt
 		visible_message("[src] wiggles around a bit.")
+	// Nymphs can consume grown food as well as weeds.
+	else if(istype(A, /obj/item/reagent_containers/food/snacks/grown))
+		if(nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
+			to_chat(src, "<span class='warning'>You're too full to consume this! Perhaps it's time to grow bigger...</span>")
+		else
+			if(do_after_once(src, 20, target = A))
+				visible_message("[src] ravenously consumes [A].")
+				playsound(loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 1.5)
```
Kyep: I always knew nymphs were demonic.

PidgeyThePirate: It's true! Also no, they still have to take blood, it just lets them consume plant matter other than weeds to gain nutrients.

PidgeyThePirate:
That feels a little arbitrary given that stealing blood doesn't hurt people. If I made it so you had to eat 5 kinds of veggies to have the DNA to grow up then you'd still need roughly the same number of veggies as when taking nutrition from a decent-potency plant.

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -92,6 +92,20 @@
 /mob/living/simple_animal/diona/UnarmedAttack(var/atom/A)
 	if(isdiona(A) && (src in A.contents)) //can't attack your gestalt
 		visible_message("[src] wiggles around a bit.")
+	// Nymphs can consume grown food as well as weeds.
+	else if(istype(A, /obj/item/reagent_containers/food/snacks/grown))
```
Fox-McCloud: This is a violation of OOP and is snowflake code. You want the behavior defined on the object in question. We already have `attack_animal` for stuff like this--you should be doing something like that, then only allowing for nymphs to benefit from this.

The idea is ok, but the code implementation is not.

code/modules/mob/living/simple_animal/friendly/diona.dm
Fox-McCloud: Though I do take issue with using this sound; it should be the same generic eat sound everyone else has.

code/modules/mob/living/simple_animal/friendly/diona.dm
PidgeyThePirate: Unique interactions and type checks should be on the interacted, not the interactor (or a bunch of vaguely related objects) and that's what "snowflaked" means in most of the previous work I've done. That makes sense, thank you for your feedback.

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -202,6 +202,20 @@
...
+	if(nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
+		to_chat(src, "<span class='warning'>You're too full to consume this! Perhaps it's time to grow bigger...</span>")
+	else
+		if(do_after_once(src, 20, target = G))
+			visible_message("[src] ravenously consumes [G].", "You ravenously devour [G].")
+			playsound(loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 2.0)
+			if(G.reagents.get_reagent_amount("nutriment")+G.reagents.get_reagent_amount("plantmatter") < 1)
+				adjust_nutrition(2)
+			else
+				adjust_nutrition((G.reagents.get_reagent_amount("nutriment")+G.reagents.get_reagent_amount("plantmatter"))*2)
```
Fox-McCloud: spacing.

code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -202,6 +202,20 @@
...
 	return TRUE
 
+// Consumes plant matter other than weeds to evolve
+/mob/living/simple_animal/diona/proc/consume(obj/item/reagent_containers/food/snacks/grown/G)
+	if(nutrition >= nutrition_need) // Prevents griefing by overeating plant items without evolving.
+		to_chat(src, "<span class='warning'>You're too full to consume this! Perhaps it's time to grow bigger...</span>")
+	else
+		if(do_after_once(src, 20, target = G))
+			visible_message("[src] ravenously consumes [G].", "You ravenously devour [G].")
+			playsound(loc, 'sound/misc/demon_consume.ogg', 30, 0, frequency = 2.0)
```
Fox-McCloud: this should be the eat sound, not the demon one.

code/modules/hydroponics/grown.dm
```diff
@@ -182,3 +182,11 @@
...
 		to_chat(user, "<span class='notice'>You open [src]\'s shell, revealing \a [T].</span>")
 	qdel(src)
+
+// Diona Nymphs can eat these as well as weeds to gain nutrition.
+/obj/item/reagent_containers/food/snacks/grown/attack_animal(mob/living/simple_animal/M)
+	if(istype(M, /mob/living/simple_animal/diona))
+		var/mob/living/simple_animal/diona/D = M
+		D.consume(src)
+	else
+		..()
```
Fox-McCloud: `return ..()`

code/modules/hydroponics/grown.dm
```diff
@@ -182,3 +182,11 @@
 		user.put_in_hands(T)
 		to_chat(user, "<span class='notice'>You open [src]\'s shell, revealing \a [T].</span>")
 	qdel(src)
+
+// Diona Nymphs can eat these as well as weeds to gain nutrition.
+/obj/item/reagent_containers/food/snacks/grown/attack_animal(mob/living/simple_animal/M)
+	if(istype(M, /mob/living/simple_animal/diona))
```
Ty-Omaha: there's an `isnymph `proc

https://github.com/ParadiseSS13/Paradise/blob/bae5faae27c8589e158cc9279947b4aea1c953e2/code/__DEFINES/mobs.dm#L207

Drakeven:
Aw, I like this!

taukausanake:
This is nice, but I was just thinking of how interesting it would be if there was a locust variant that had no limitations on how much it could eat. This is a good alternative to weeds but then it probably means they won't eat weeds at all because _it's a good alternative_ so I'm also kind of mixed

PR #12425, opened 2019-09-26
Diona Reproduction

PidgeyThePirate:
## What Does This PR Do
Grants every Roundstart or Arriving diona the ability to plant a diona seed pod that produces a single diona nymph. This nymph is not in any way subservient to the planter, and may only be planted once. Terrarium and Pod Dionaea do not receive this ability. 

## Why It's Good For The Game
1. Lore: Since the rework Dionaea have lost a lot of their identity. My hope is to lean into the "life giver" origin that the terrarium dionaea suggest. 
2. Mechanics: The ability to "grow your own friend" is interesting and situationally useful, but is highly limited in situations where the subject is expected to be alone anyway, basically providing little more than moral support. It also relies heavily on there being available ghosts whose attention you can grab.

## Issues: 
- This comes with some amount of implied subservience which may be an issue if the parent Diona is a traitor; the station may wind up facing two traitors instead of one despite having no mechanical subservience. This may or may not be an acceptable outcome, but it is important to remember that the same can happen with both golems and the standard diona nymph seeds in botany.

## Changelog
:cl:
add: Roundstart and Arriving Dionaea can asexually reproduce by planting a single diona nymph seed. This seed is the same as standard diona nymph seeds in every respect except the resulting plants may only ever produce one nymph.
tweak: Changed the Diona blurb slightly to remove some outdated references to "slowness", rad resistance and so on.
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/nymph.dm
+++ b/code/modules/hydroponics/grown/nymph.dm
@@ -12,33 +12,6 @@
 	yield = 1
 	reagents_add = list("plantmatter" = 0.1)
 
-// If the nymph pods are fully grown, allow ghosts that click on them to spawn as a nymph (decreases yield by 1 and kills the plant when yield reaches 0)
-/obj/item/seeds/nymph/GhostAttackWhenPlanted(mob/dead/observer/O, obj/machinery/hydroponics/T)
-	if (T.harvest && !T.dead)
-		if (!(O in GLOB.respawnable_list))
-			to_chat(O, "You are not permitted to rejoin the round.")
-			return
-		else if (cannotPossess(O))
-			to_chat(O, "You have enabled antag HUD and are unable to re-enter the round.")
-			return
-		var/nymph_ask = alert("Become a Diona Nymph? You will not be able to be cloned!", "Diona Nymph Pod", "Yes", "No")
-		if (nymph_ask == "No" || !src || QDELETED(src))
-			return
-		if (!(T.myseed == null) && yield > 0 && !T.dead)
-			yield -= 1
-			var/mob/living/simple_animal/diona/D = new /mob/living/simple_animal/diona(get_turf(T))
-			D.key = O.key
-			GLOB.respawnable_list -= O
-			visible_message(D, "A new diona nymph emerges from the pod, its antennae waving excitedly.")
-			if (yield <= 0)
-				visible_message("The seed pod withers away, now merely an empty husk.")
-				T.plantdies()
-			spawn(5)
-				GLOB.respawnable_list += usr
-		else
-			to_chat(O, "Seed: [T.myseed.name], Yield: [yield], Dead: [T.dead]")
-			to_chat(O, "The seed pod is no longer functional. It has probably been used up or destroyed in some way.")
-
 /obj/item/reagent_containers/food/snacks/grown/nymph_pod
 	seed = /obj/item/seeds/nymph
 	name = "nymph pod"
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/simple_animal/friendly/diona.dm
```diff
@@ -90,7 +90,7 @@
 	steal_blood_action.Grant(src)
 
 /mob/living/simple_animal/diona/UnarmedAttack(var/atom/A)
-	if(isdiona(A) && (src in A.contents)) //can't attack your gestalt
+	if (isdiona(A) && (src in A.contents)) //can't attack your gestalt
```
farie82: Spacing. Prob a mistake

Fox-McCloud:
This is *incredibly* snowflakey.

Generally speaking, if something can't be contained to a species trait for the species datum, then it's probably not a good idea.

Also, being able to spawn more of yourself is really questionable, especially since said spawn can be player controlled---even more so on a round start species.

code/modules/mob/living/carbon/human/species/diona.dm
```diff
@@ -70,16 +72,43 @@
 		return 1
 	return 0
 
+// Only roundstart and arriving dionaea should have had time to prepare a seed pod. 
+// Terrarium and Pod dionaea will likely have access to botany and won't need this ability. 
+/datum/species/diona/after_equip_job(datum/job/J, mob/living/carbon/human/H)
```
Fox-McCloud: This is hacky at heck to do this.

PidgeyThePirate:
Fair enough. I don't really understand why that particular sample is "hacky" and vox equipment (which I based it off) is not, but I trust you know what you're doing. For future reference, would there be a better way to implement similar kinds of roundstart ability or is that just a design no-no?

I based most of this on what diona nymphs already had in an attempt to make the code better. I'm sorry it didn't work out.

TDSSS:
I like the idea, it would give Diona something genuinely unique and flavorful without being OP or anything like that. After all, the new Diona is not your servant in any way, it's basically just a way for ghosts to rejoin the round.

PidgeyThePirate:
Would it be better theoretical design to give the seed directly to the diona as an item instead of hiding it in an ability? I housed it in an ability so it couldn't be stolen and the seed was less available to a DNA machine, but there were probably better ways to do it.

code/modules/mob/living/carbon/human/species/diona.dm
PidgeyThePirate: Is this a bit closer to alright?

Couls: It would be more helpful if you told them why it’s hacky and what they should do or use instead

code/modules/hydroponics/grown/nymph.dm
```diff
@@ -23,4 +23,15 @@
...
+	qdel(src)
+
+// The seed granted to a diona at round start.
+/obj/item/seeds/nymph/diona_innate
+	desc = "A single seed produced by a small diona gestalt. It can typically only grow one nymph."
+	name = "diona nymph seed"
+
+// The Diona's innate pod will not yield more than one nymph at harvest and perennial growth cannot be added (see plant_genes.dm).
+// A seed extractor can possibly make more, but if the player is in a position to do that they can just get the nymph seeds from the vendor anyway.
+/obj/item/seeds/nymph/diona_innate/getYield()
```
SteelSlayer: Seeing as how the parent `getYield()` is a proc, I believe you need to add in "proc" before the `getYield()`.
Like this: `/obj/item/seeds/nymph/diona_innate/proc/getYield()`

code/modules/mob/living/carbon/human/species/diona.dm
```diff
@@ -70,6 +70,12 @@
 		return 1
 	return 0
 
+/datum/species/diona/after_equip_job(datum/job/J, mob/living/carbon/human/H)
+	var/obj/item/seeds/nymph/diona_innate/seedpod = new /obj/item/seeds/nymph/diona_innate
```
SteelSlayer: While you *can* include the path name of the thing you're creating after the `new`, it's not needed. You can just write `= new` here if you want.

code/modules/hydroponics/grown/nymph.dm
PidgeyThePirate: getYield() is a proc in /obj/item/seeds which is altered in this type. Adding /proc/ would throw a duplicate definition error; I know. I've tried.

PidgeyThePirate:
I figured as much. So be it, then.

Issue #12408, opened 2019-09-24
Maintenance Drones need both robotics and engineering access to reboot


Shadow-Quill:
**Issue Description**:

Title

**What did you expect to happen**: 

To be able to reboot/shutdown a maintenance drone as an engineer or the roboticist.

**What happened instead**:

You have to get the captain.

**Why is this bad/What are the consequences**:

My poor baby maintenance drones won't get rebooted :(

**Steps to reproduce the problem**:

1) Be engineer/roboticist and swipe your ID on a drone.
2) Have a breakdown because you can't restart the purest thing on this station.
3) ~~Throw yourself out the airlock, embracing the empty shell of your last friend.~~

**When did the problem start happening**:

?

**Extra information**:

Robots use `req_access`, not `req_one_access`

PR #12431, opened 2019-09-27
Issue #12408 - Fixes drones requiring all-access to reboot.

PidgeyThePirate:
## What Does This PR Do
Fix Issue #12408 by adding req_one_access to robot access checks.

## Why It's Good For The Game
You can reboot drones without bothering the captain now.

## Changelog
:cl:
fix: Rebooting a drone no longer requires both engineering and robotics access.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone.dm
@@ -13,7 +13,8 @@
 	braintype = "Robot"
 	lawupdate = 0
 	density = 0
-	req_access = list(access_engine, access_robotics)
+	req_access = null
+	req_one_access = list(access_engine, access_robotics)
 	ventcrawler = 2
 	magpulse = 1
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -13,7 +13,8 @@
 	braintype = "Robot"
 	lawupdate = 0
 	density = 0
-	req_access = list(access_engine, access_robotics)
+	req_access = null
```
variableundefined: This statement should be deleted if it is being replaced.

PidgeyThePirate: req_access is by default list(access_robotics), so it needs to be nulled or engineers won't be able to reboot the drone. There might be a better way but I'm leery of substantially altering the structure of the robot type for the sake of a single edge case.

Issue #12331, opened 2019-09-17
Bluespace translocation event teleports observers to the void


dovydas12345:
**Issue Description**:

If you happen to be a observer and inside of the area, where the bluespace anomaly triggers the bluespace translocation event, it just puts you into the void where your screen is black and you can't do anything without admin's intervention.

**What did you expect to happen**: 

To not be put into the void by the bluespace translocation event as a observer.

**What happened instead**:

A observer was put into the void where they can't do anything without admin intervention.

**Why is this bad/What are the consequences**:

You cannot fly around as a ghost while in the void.

**Steps to reproduce the problem**:

1. Spawn in a bluespace anomaly
2. Be in the area of the bluespace anomaly as a observer
3. Wait for it to go off

**When did the problem start happening**:

Not certain, probably for a long time, but today (2019-17-09) was the first time I seen it happen.

**Extra information**:

Teleporting the player fixed the issue.

PidgeyThePirate:
I attempted to replicate this on my personal test server and couldn't do it. The anomaly does however teleport ghosts, which shouldn't happen, and fixing that should fix the issue.

dovydas12345:
I suppose something broke with teleporting then. So yeah fixing it teleporting ghosts should fix the issue.

JKnutson101:
@variableundefined sorry for the ping, just going through the issues and found one to close cause #12433 fixed it.

PR #12433, opened 2019-09-27
Issue #12331 - Stops bluespace anomaly from teleporting ghosts.

PidgeyThePirate:
## What Does This PR Do
Couldn't reproduce #12331. Stopped the bluespace anomaly from teleporting ghosts anyway just in case.

## Why It's Good For The Game
Ghosts shouldn't be affected by most ingame events, and I'm pretty sure this was unintended.

## Changelog
:cl:
tweak: bluespace anomalies no longer teleport ghosts when their event fires.
/:cl:

### DIFF
--- a/code/modules/events/anomaly_bluespace.dm
+++ b/code/modules/events/anomaly_bluespace.dm
@@ -45,9 +45,7 @@
 				for(var/atom/movable/A in range(12, FROM )) // iterate thru list of mobs in the area
 					if(istype(A, /obj/item/radio/beacon)) // don't teleport beacons because that's just insanely stupid
 						continue 
-					if(isobserver(A)) // don't teleport ghosts.
-						continue
-					if(A.anchored) 
+					if(A.anchored || A.move_resist == INFINITY)
 						continue
 					var/turf/newloc = locate(A.x + x_distance, A.y + y_distance, TO.z) // calculate the new place
 					if(!A.Move(newloc)) // if the atom, for some reason, can't move, FORCE them to move! :) We try Move() first to invoke any movement-related checks the atom needs to perform after moving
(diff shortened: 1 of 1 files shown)


code/modules/events/anomaly_bluespace.dm
```diff
@@ -43,9 +43,12 @@
 				var/y_distance = TO.y - FROM.y
 				var/x_distance = TO.x - FROM.x
 				for(var/atom/movable/A in range(12, FROM )) // iterate thru list of mobs in the area
-					if(istype(A, /obj/item/radio/beacon)) continue // don't teleport beacons because that's just insanely stupid
-					if(A.anchored) continue
-
+					if(istype(A, /obj/item/radio/beacon)) // don't teleport beacons because that's just insanely stupid
+						continue 
+					if(istype(A, /mob/dead/observer)) // don't teleport ghosts.
```
Kyep: Might be better to use "if(isobserver(A))"

https://github.com/ParadiseSS13/Paradise/blob/bae5faae27c8589e158cc9279947b4aea1c953e2/code/__DEFINES/mobs.dm#L231

code/modules/events/anomaly_bluespace.dm
```diff
@@ -43,9 +43,12 @@
 				var/y_distance = TO.y - FROM.y
 				var/x_distance = TO.x - FROM.x
 				for(var/atom/movable/A in range(12, FROM )) // iterate thru list of mobs in the area
-					if(istype(A, /obj/item/radio/beacon)) continue // don't teleport beacons because that's just insanely stupid
-					if(A.anchored) continue
-
+					if(istype(A, /obj/item/radio/beacon)) // don't teleport beacons because that's just insanely stupid
+						continue 
+					if(isobserver(A)) // don't teleport ghosts.
```
variableundefined: Ghost has their move resist set to infinite. Instead of checking for observer, check for A.move_resist = INFINITY

Like this:
``` if(A.anchored || A.move_resist == INFINITY) ```

(Anchored and move_resist are bundled together since they're the same concept, except move_resist applies to moving creatures)

PR #12439, opened 2019-09-27
replace high power cells with high power cell+

Evankhell561:
## What Does This PR Do
change all high power cells  with a 15,000 max charge with high power cell+ on the map

## Changelog
:cl:Evankhell
tweak: replace high power cells with high power cell+
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -5205,9 +5205,8 @@
 	pixel_x = 0;
 	pixel_y = 10
 	},
-/obj/item/stock_parts/cell/high{
-	charge = 100;
-	maxcharge = 15000
+/obj/item/stock_parts/cell/high/plus{
+	charge = 100
 	},
 /obj/effect/decal/warning_stripes/east,
 /obj/item/clothing/glasses/welding,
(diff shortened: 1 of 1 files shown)


datlo:
Please explain what are you exactly doing and why you are doing it, please

variableundefined:
If I have to hazard a guess, it's because high power cell is defined as a cell with 10000 max_charge, while the one with 15000 charge is high/plus.

This reduces the amount of snowflake varedit on the map.

That being said you should probably explain *why*

_maps/map_files/cyberiad/cyberiad.dmm
```diff
@@ -84452,13 +84443,11 @@
 /obj/item/rcd_ammo,
 /obj/item/rcd_ammo,
 /obj/item/rcd_ammo,
-/obj/item/stock_parts/cell/high{
-	charge = 100;
-	maxcharge = 15000
+/obj/item/stock_parts/cell/high/plus{
+	charge = 100
```
Fox-McCloud: this shouldn't exist either.

Evankhell561:
> If I have to hazard a guess, it's because high power cell is defined as a cell with 10000 max_charge, while the one with 15000 charge is high/plus.
> This reduces the amount of snowflake varedit on the map.
Yes, it is the reason. Also, players should know what the maximum charge is only by the name of the item.

Issue #12104, opened 2019-08-27
Deja Vu! I've just been in this place before! Drifting on vehicles is too fast.


JKnutson101:
**Issue Description**:
Diagonal movement on vehicles like the janicart are not slowed down like other diagonal movement.
**What did you expect to happen**: 
For the diagonal movement of vehicles to match up with the rest of the game.
**What happened instead**:
K-Kono Speedo?! NANI?! I went so fast I felt like I was in a famous Japanese anime known for it's Eurobeat soundtrack.
**Why is this bad/What are the consequences**:
This could actually lead to some really silly powergaming with the janicart or other vehicles.
**Steps to reproduce the problem**:
Get on a vehicle and move diagonally. Experience high speed drifting!
**When did the problem start happening**:
Probably since Diagonal movement was merged.
**Extra information**:
Drifting at high speeds is pretty fun, enjoy it while it lasts!

JKnutson101:
This was fixed in #12441

PR #12441, opened 2019-09-28
Minor Vehicle Refactor and Balance

Fox-McCloud:
I wasn't going do any more refactors until universal object damage was done, but vehicles had a number of bugs with them, so might as well refactor them a bit in addition to bugfixing them; this is largely from TG, with a single noteworthy change

Here's what has changed
- You can insert keys into some vehicles; when the key is inserted, you can freely drive the vehicle without holding the key in your hand; lavaboats still require you to actually hold the key
  - Click the key on a vehicle to insert it; alt-click while riding a vehicle to remove the key
- Fixed a bug where moving diagonally with a vehicle results in faster movement speed than otherwise intended
- Fixed a bug where being on a vehicle didn't allow for proper z-level registered transitions
- Increased the speed on the secway from 2 to 1
- Vehicles now factor in human movement delay. This is changed, from TG. While TG has had the ability to change default mob movement speed, they've never really utilized it. As such, a lot of their systems are balanced around the assumption that the configured human movement delay is `0`. We use `1.5` for human movement delay, so it makes several things stand out that TG hasn't plugged into their system yet
  - This means vehicles are not the be all and end all they used to be; unless they have a movement delay of `1` or less, you will be slower riding a vehicle than running about. It's the price you pay for being immune to damage slow, being a lot harder to click, and immune to most slips
  - It's also worth mentioning this means speedbike and ATVs are no longer "cross the entire station in literally 3-4 seconds"--speedbikes, in particular, are still faster than human movement (intended), but it's not so fast as to be unclickable
- Updates the ATV sprites and fixes the overlay

img

:cl: Fox McCloud
add: can insert keys into vehicles instead of having to hold them in your hand (alt-click while riding a vehicle to remove the key); lava boat still requires holding the key in your hand
fix: Fixes moving diagonally with vehicles making you super fast
tweak: Vehicles now factor in the configured human movement delay; this means all vehicles are slower than they previously were
fix: fixes a bug with mobs on vehicles causing incorrect space transitions
tweak: ATV sprites updated
add: Slime speed potions can now apply to vehicles, which makes them faster
/:cl:


code/modules/vehicle/vehicle.dm
```diff
@@ -93,25 +134,30 @@
...
 		if(!Process_Spacemove(direction) || !isturf(loc))
 			return
 		step(src, direction)
 
-		if(has_buckled_mobs())
-			for(var/m in buckled_mobs)
-				var/mob/living/buckled_mob = m
-				if(buckled_mob.loc != loc)
-					buckled_mob.buckled = null //Temporary, so Move() succeeds.
-					buckled_mob.buckled = src //Restoring
```
Fox-McCloud: This ugly hack, above, isn't needed anymore, since we have a system in place that relays buckled mobs movement when the thing they're buckled to moves---this hack of unbuckling then rebuckling was only necessary because of that.

code/modules/vehicle/vehicle.dm
```diff
@@ -93,25 +134,30 @@
...
+		if((direction & (direction - 1)) && (loc == next))		//moved diagonally
+			last_move_diagonal = TRUE
+		else
+			last_move_diagonal = FALSE
 
+		if(has_buckled_mobs())
```
Fox-McCloud: Sadly, the lines below this (related to lube slipping) are still needed.

TDSSS:
Repeating from discord: really sad to see ambulances needed like that, no real reason to use them anymore when.you can just pull people and be faster.

JKnutson101:
Fixes #12104

Glad to see this vehicle refactor! (Might make ridable cyborgs PR possible now =P)

>Repeating from discord: really sad to see ambulances needed like that, no real reason to use them anymore when.you can just pull people and be faster.

Aside from the mentioned benefits from Fox...
>immune to damage slow, being a lot harder to click, and immune to most slips
You can also grab and drag multiple bodies. If I'm correct, a Paramedic can now drag up to four bodies on the ambulance. One on the trolley, which is hooked to the ambulance, one being pulled, and one being passive grabbed in each hand now that your can have both hands free. There's still reason to use it, just not all the time.

Fox-McCloud:
Made some speed adjustments (TG has a different `run_speed` config and made adjustments to their vehicles because of it); also ported over the slime speed potion being able to impact vehicle speed.

taukausanake:
Maybe the ambulance can benefit from multiple beds being connected to it? Kind of like how the cargo train works. I always though that was the best benefit to vehicles.

_I miss the cargo train_ :(

TDSSS:
>Aside from the mentioned benefits from Fox...

I mean...the paramedic shouldn't be in combat, he should not chase after antags to rip their victims from their hands. Being more tanky and more able to withstand antags but less good at actually saving lives doesn't seem that good of a trade-of.

Being able to pull a total of 4 instead of 3 is a benefit, sure, but I rarely see such mass causality events, tbh.

lolzyking:
Gotta agree with Tdss on the casualties if we're looking at 3+ we're just gonna locker drag them anyway....

Fox-McCloud:
The ambulance being a go-to combat vehicle is also not a desirable thing, either. Yes, the paramedic "shouldn't" be doing this, but he can also be a traitor---it can also be used by the CMO, or it could be a situation where if he's "validing" it doesn't matter (Xenos, Warops, Wizard, the likes). The primary point of vehicles is to generally assist with situations and help improve the slowdown from certain equipment (for example, carrying a duffel+wearing a hardsuit)--not a be all and end all of movement speed.

shawnhind:
This change honestly baffles me.  If the ambulance is slower than walking there's no reason to use it. 

Driving the ambulance with the sirens on with people just walking past you has to be one of the silliest things I've seen.

shawnhind:
@Fox-McCloud I've never even seen it getting abused as much as you are making it out to be.  Why nerf it in 100% of cases if the cases of abuse are seemingly less than 10%.  If the paramedic is a traitor and he uses the ambulance to his advantage then so what?  Who decided what the primary point of vehicles is?  I would have never thought their primary purpose would be anything other than speed.

PR #12442, opened 2019-09-28
change the name of slime organs to be less confusing

Couls:
## What Does This PR Do

Renames the slime lungs to slime lungs and slime's heart to slime heart

## Why It's Good For The Game

Newer players will be confused as to what these organs are and I think gameplay comes over lore in this case, the details of the organ are kept in the description but the names are simplified so doctors don't have to learn new terminology to deal with slime people.

## Changelog
:cl:
tweak: Renames slime organs to be more player friendly
/:cl:


Lunaimon:
Diona have had confusing names for all their parts since forever. Why should slimes be different?

Couls:
Dunno, I just wanted this to be an option if slime organ names proved to be confusing people or if it becomes an issue. I personally found myself doubting which organ did which and I feel like this is less of an issue for dionae because they hardly came into medbay in the first place

MRgaty:
You could use terms like pump for the heart and filter for the lungs.

Carthusia:
I feel it's only confusing at first, then you learn it.

Like how the OR machine doesn't say "Use hemostat" it just says Clamp Bleeders

At first yea, i had no idea what that meant, but you learn quick as a doctor.

It's not called Malpractice Bay for nothin, you make mistakes and learn from them, its part of the fun

variableundefined:
@Couls Conflicts

Couls:
@variableundefined resolved!

Issue #10918, opened 2019-02-17
Abductor Fleshy Mass Sprite Dissapears When Removed


Pckables:
**Problem:**
Fleshy Masses removed via surgery have invisible/no sprites. The object still exists, there is just no visible sprite for it.

**Why is This An Issue**
Invisible objects are generally bad for gameplay. Less experienced surgeons might get confused when their seemingly empty hand can't pick anything up.

**Steps to reproduce the problem**:
Have an abductor insert a fleshy mass into a victim. Afterwards, have a normal surgeon remove this fleshy mass. If it has a sprite when initially removed (it may be visible initially), it will lose its sprite when dropped or put into a bag.

**When did the problem start happening**:
This has always been an issue for as long as I've been around.

**Other Notes**:
Sprite may still appear on examine, despite being invisible in-hand.
The Runic Mind (brain) of a golem has a similar issue at the moment.

PR #12444, opened 2019-09-28
fix slime heart going invisible

Couls:
## What Does This PR Do

update_icon proc on the heart was making its icon go invisible, overriding the proc for slime heart fixes this. This was also an issue for abductor fleshy masses, they also went invisible after being bitten or necrotizing due to having the dead_icon being defined as [icon]-off which fleshy masses and slime hearts both don't have

fixes: #10918
## Why It's Good For The Game

being able to see organs is always nice.

## Changelog
:cl:
fix: slime heart and fleshy masses no longer go invisible after being removed through surgery or being eaten or necrotizing
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/abduction/gland.dm
+++ b/code/game/gamemodes/miniantags/abduction/gland.dm
@@ -3,6 +3,7 @@
 	desc = "A nausea-inducing hunk of twisting flesh and metal."
 	icon = 'icons/obj/abductor.dmi'
 	icon_state = "gland"
+	dead_icon = null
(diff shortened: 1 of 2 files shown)


Couls:
It's two seperate issues, you need to set dead_icon to null so when the heart dies it's icon isn't changed but you also need to override update_icon because it attempts to give the heart beating animations
https://github.com/ParadiseSS13/Paradise/blob/10ef2311ab0b5283422da0ba63783ab97430e1e2/code/modules/surgery/organs/heart.dm#L12

PR #12446, opened 2019-09-28
Reduces slime damage mod to 150% from 250%.

PidgeyThePirate:
## What Does This PR Do
Reduces Slime Person brain damage mod to 150% from 250%.

## Why It's Good For The Game
Now that slimes use newcrit, they die ridiculously fast due to the oversized brain damage mod, which in turn makes it harder to defibrillate them in time. Since they have fewer organs, the important ones turn septic much faster making them harder to revive after death. This should hopefully give them a fighting chance while maintaining some of their previous flavour and vulnerability.

## Changelog
:cl:
balance: Slime brain damage multiplier reduced to 150% from 250%.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/slime.dm
+++ b/code/modules/mob/living/carbon/human/species/slime.dm
@@ -20,7 +20,7 @@
 	cold_level_3 = 200
 	coldmod = 3
 
-	brain_mod = 2.5
+	brain_mod = 1.5
 
 	male_cough_sounds = list('sound/effects/slime_squish.ogg')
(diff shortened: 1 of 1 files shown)


Mitchs98:
I'd rather it be removed outright. As said on Discord it was only there due to the fact they had only one organ, and thus to compensate for that they had to keep core safe. With brain damage being an integral part of newcrit no race should have a modifier that raises brain damage, much less a race that's already fairly hard to revive due to the huge health thresh-holds newcrit provides. 

That said. Sizeable reduction is better than nothing even if I see no reason for it to continue to exist at all.

PR #9223, opened 2018-07-17
Makes Species datum Non-Global

Fox-McCloud:
What a massive pain in the ass.

@Aurorablade sorry to steal your thunder, not trying to. It's just a project like this I can't really work through and review by myself and pinch and poke at; it's something I have to work through myself.

Here it is, the fully functional and working non-global species datums. This not only makes the species datums non-global, but makes it held on the mob's DNA (or the organ's DNA).

Code Changes
- Species datums are now held on a human mob's DNA
- Each human mob's DNA has a unique species datum instance, so you can alter it without it impacting every other of that species type on the map
- Organs no longer explicitly have a species datum, but instead rely on their own DNA datum species holder var (This worked out very nicely
- Refactored a lot of explicit species checks to be datum based
  - There's some species checks that aren't datum based yet; these are mostly for game modes where the user's prefs (which stores species as a name string) are checked against a list. Don't want to touch these in this PR

Fixes:
- Fixes monkey HUDs not showing up
- Fixes going from monkey to human and not having a HUD
- Fixes some items unequipping when you change species
- Fixes mitocholide not rejuvinating some organs
- Fixes shadowlings not being forced to hatch after a time
- Fixes abductor console not properly clearing some action buttons
- Fixes Vox name generation being based on incorrect syllables
- Consolidates Diona name generation instead of it randomly generating a name in player select and then asking you again, on spawn with a different set of names (the lists have been combined together
- Fixes DNA scrambler so it generates an appropriate species name
- Fixes runtimes with spawning in `/mob/living/carbon/human/interactive`
- Fixes a runtime with apply DNA to a head already deleted
  - This was occurring because of Vox player preview during start up; it was caused by spawns in `set_species` and some organ icon updates
- Fixes Vox lungs going nectrotic/being removed and causing permanent "low on nitrogen" warnings.
- Fixes operatives being bald most of the time

Additions
- Having a robotic head will now allow you to beep, buzz, ping, and the likes
- Claw sharpening actually works
- Abductors have their own headsets
- Can put monkeys in the gibber
- Different species now produce different hide types when put into the gibber (monkies will produce monkey hide, Unathi will produce lizard hide, Diona will produce sheets of wood, and other similar things; not every species has a hide).

Removal
- Removes ability to pick name as a diona nymph when evolving (consistent with ghost roles)

Thus far, no runtimes, icons works, and everything seems a-ok.

:cl: Fox McCloud
fix: Fixes monkeys not having their own HUD
[...]

PR #12455, opened 2019-09-30
DNA Vault Rework

Fox-McCloud:
From TG.

DNA Vault is currently a monumental task to finish; it's probably one of the single most engaging station goals, but it's almost incredibly difficult to complete, in part due to the samples---and the rewards it gives are....well, pretty much just what you can get from genetics. This ensures the things you get the from the DNA vault are unique while still having somewhat of an impact.

This does two things:
- Makes the DNA vault easier to construct (5 super capacitors+5 pico manipulators as opposed to 5 quadratic capacitors)
- Adjusts what the DNA vault gives you
  - All the previous powers have been removed, with the exception of the no breathing one
  - None of these are genetic powers, but are, instead, modifications to your species (yay for non-global species datums)
  - It no longer allows you to have unlimited amounts of genetics powers

New things the DNA vault gives
- Makes your lungs take no toxin damage in addition to granting you virus immunity
- Makes you heat resistant and lowers fire damage you take by 50%
- Halves your stun times
- Armors your species by taking 30% less damage and making you `PIERCEIMMUNE`
- Increases your movement speed by 1 level
- Decreases the cooldown between actions (punch/disarm/shoot/stab/whatever faster)

Other stuff
- Refactors species traits to actually have a text string so they're easier to debug
- `speedmod` was added by @Couls without plugging it into anything, despite the fact `slowdown` existed. Redundancy removed and all instances of species related `slowdown` or `speedmod` modified to `speed_mod`

:cl: Fox McCloud
tweak: DNA vault requires 5 pico manipulators and 5 super capacitors instead of quadratic capacitors
tweak: Powers of the DNA vault adjusted
del: DNA vault no longer allows you to have unlimited genetics powers.
del: All previous powers removed with the exception of no breathing
add: New powers: breathable plasma immunity+virus immunity, heat immunity+burn reduction, speed increase, species armoring+pierce immunity, stun reduction, and action speed increase
/:cl:


PidgeyThePirate:
I like this. We already have a tough science/engineering challenge with the BSA, this puts the focus on the other parts of the station to finish the vault as the vault itself can be completed as a matter of course. The rewards are nicer too.

MrMagolor:
I think it's good that the DNA vault, being a station goal, has generally better and more unique rewards than what is basically just genetics.

code/modules/station_goals/dna_vault.dm
```diff
@@ -267,34 +275,49 @@ var/list/non_simple_animals = typecacheof(list(/mob/living/carbon/human/monkey,/
 		return
 	if(!completed)
 		return
-	if(!H.ignore_gene_stability)
-		to_chat(H, "<span class='notice'>[src] stabilizes your genes, granting you the ability to have multiple powers.</span>")
-		H.ignore_gene_stability = 1
+	var/datum/species/S = H.dna.species
+	if(NO_DNA in S.species_traits)
```
Fox-McCloud: This check is now necessary; as the DNA vault was previously genetics based, IPCs (and other potentially future species without DNA) weren't able to utilize it as setting blocks for them will fail; as this is all species datum based and altering vars more directly, we now have to explicitly check for this.

code/modules/station_goals/dna_vault.dm
```diff
@@ -267,34 +275,49 @@ var/list/non_simple_animals = typecacheof(list(/mob/living/carbon/human/monkey,/
...
+		to_chat(H, "<span class='warning'>Erorr, no DNA detected.</span>")
```
Fox-McCloud: spelling

Fox-McCloud:
Nothing else about the DNA vault has changed.

It still only rolls for 2 random powers at shift start and grants only one of those.  You can still only select one of those two powers and are blacklisted from selecting them in the future.

Saying that we don't expect species datums to change is utter nonsense. The whole point of: https://github.com/ParadiseSS13/Paradise/pull/9223 was precisely so we *could* modify species datums  at runtimes and have a uniquely held instance on each mob; if we ever intended to keep species static, globally, then making species datums non-global would be a waste of memory resources as you now have several thousand new instances of species datums floating around for literally zero reasons or benefits.

code/modules/station_goals/dna_vault.dm
streather: From my own experience the DNA vault previously had no problems working with IPCs and other synthetics, the rework currently doesn't seem to work, when I tried to use it in a round earlier I got an error saying that I had no DNA.

JKnutson101: Actually speaking for a *lot* of previous experience, IPCs were very much able to use the DNA Vault before and now they cannot. I thought this change was intended.

Issue #12338, opened 2019-09-17
Apostrophes dont work


Ty-Omaha:
**Issue Description**:

![](https://i.imgur.com/E0PjYn2.png)

![](https://i.imgur.com/NCuOxmG.png)

**What did you expect to happen**: 
For apostrophes to not be escaped

**What happened instead**:
They are

**Why is this bad/What are the consequences**:
Because it looks stupid

**Steps to reproduce the problem**:
Have a name with apostrophes in it, mouse over the mob and view name in bottom left. Have medibot or automated announcement announce your name.

**When did the problem start happening**:
Probably #11515

https://github.com/ParadiseSS13/Paradise/blob/bb4a6536953d34054e74bbdc405ba032b3653764/code/modules/client/preference/preferences.dm#L1281

https://github.com/ParadiseSS13/Paradise/blob/5d9fc8e9ea543e313a0225df513cb7244ca68984/code/__HELPERS/text.dm#L132

https://github.com/ParadiseSS13/Paradise/blob/5d9fc8e9ea543e313a0225df513cb7244ca68984/code/__HELPERS/text.dm#L32

PR #12456, opened 2019-09-30
Simple Animal Fixes

### LINKED ISSUE #3596: can't give karma to pokeball captured pets
I couldn't give karma to the player controlled nurse ian that dragged patients to me and helped me in medbay, and that broke my heart. It says you can't give karma to yourself.

### LINKED ISSUE #11561: "Respawn as NPC" Mobs with broken pulling after respawning again/resurrection
**Problem Description**:
"Respawn as NPC" Mobs (Ian, Runtime, Poly, etc) have their ability to pull break when you choose "Respawn as NPC" more than one time in a round, or get brought back from death.

For pulling to work consistently as an NPC, irrespective of how many times you're resurrected with Strange Reagent/Respawn as a different NPC

**What happened instead**:
When you're brought back to life, or you choose a different NPC to respawn as, you can no longer pull anything. Tested both Ctrl and Right Click, {Item}, Pull. Quit and reconnect did not resolve either. Works as expected the next round (On the first life).

As an NPC, your options are so limited that losing pull hurts your play options significantly.

Do not choose Join Game/Ready, go straight into Observe (you can wait until round starts or not), at game start choose Respawn as NPC, pick your poison, pull an object (or player) successfully, die, choose Respawn as NPC, pick somebody else, spawn, try and pull an object.

Ever since I've started to play NPC heavily... I pretty much never choose Respawn as NPC when I die during a round so I'm not sure if joining the game or not affects this.

**Possibly related stuff : 
So I've never tested this when I've played as a PC, died, and chosen Respawn as NPC, died as NPC, and Respawned as NPC again/Been Strange Reagent'ed so if pulling works fine after you join and follow those steps, then it's exclusively when you go into NPC after immediately Observing the round on Join.
Also, the only two NPCs I got this with was with a mouse (after dying as Clucky) and playing as Poly and getting SR'ed. I have a feeling it's not going to be specific to NPCs but if you're having trouble reproducing, try the scenarios above.

### LINKED ISSUE #12032: Cant pull objects / Climb onto surfaces after SR as Ian
**Problem Description**:
After SR'd as Ian I realized I could no longer pull objects or climb onto surfaces anymore as Ian.  John Clarke asked me to post this. :)

Being able to pull things and climbing onto things, Usual Ian good-boy stuffs.

**What happened instead**:
Nothing happened at all.  No pull icon appeared.  Tried Control + Clicking and Right Click + Pull.

Ian cant be the bestest of boys if not able to play fetch! Or Scamp around on tables

Die horrificly to the dismay of others as Ian.  Wait for someone to SR you.

After I was SR'd.

Ian was being a good boy and was SR'd, went immediately running to HoP!  Tried to grab HoP a snack from kitchen counter on way back and realized I couldnt pull.  Then tried to jump onto the HoP counter, cutting in line, to jump into owners arms,  Realized could no longer climb surfaces.

PidgeyThePirate:
## What Does This PR Do
- Fixes #11561, Fixes #12032: Revived simple animals keep their ability to pull things, climb on things and their density if they originally had it.

## Why It's Good For The Game
Squishes a few more bugs, closes a few more issues.

## Changelog
:cl:
fix: Revived simple animals keep their ability to pull things and their density if they originally had it.
/:cl:

### DIFF
--- a/code/modules/mining/equipment/lazarus_injector.dm
+++ b/code/modules/mining/equipment/lazarus_injector.dm
@@ -96,7 +96,7 @@
 	else
 		if("neutral" in S.faction)
 			S.forceMove(src)
-			S.name = "[M.name]'s [initial(S.name)]"
+			S.name = "[M.name]&#39s [initial(S.name)]"
 			S.cancel_camera()
 			name = "Lazarus Capsule: [initial(S.name)]"
 			to_chat(M, "<span class='notice'>You placed a [S.name] inside the Lazarus Capsule!</span>")
--- a/code/modules/mining/equipment/lazarus_injector.dm
+++ b/code/modules/mining/equipment/lazarus_injector.dm
@@ -96,7 +96,7 @@
 	else
 		if("neutral" in S.faction)
 			S.forceMove(src)
-			S.name = "[M.name]&#39s [initial(S.name)]"
+			S.name = "[M.name]'s [initial(S.name)]"
 			S.cancel_camera()
 			name = "Lazarus Capsule: [initial(S.name)]"
 			to_chat(M, "<span class='notice'>You placed a [S.name] inside the Lazarus Capsule!</span>")
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -84,7 +84,6 @@
 	var/del_on_death = 0 //causes mob to be deleted on death, useful for mobs that spawn lootable corpses
 	var/deathmessage = ""
 	var/death_sound = null //The sound played on death
-	var/initial_density //Whether the density of a simple animal starts as 0 or 1. Used for revive().
 
 	var/allow_movement_on_non_turfs = FALSE
 
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -338,8 +338,7 @@
 
 /mob/living/simple_animal/revive()
 	..()
-	if(initial_density)
-		density = 1
+	density = initial(density)
 
 /mob/living/simple_animal/death(gibbed)
 	// Only execute the below if we successfully died
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -334,6 +334,10 @@
 		for(var/i in loot)
 			new i(loc)
 
+/mob/living/simple_animal/revive()
+	..()
+	density = 1
+
 /mob/living/simple_animal/death(gibbed)
 	// Only execute the below if we successfully died
 	. = ..()
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -102,7 +102,6 @@
 
 /mob/living/simple_animal/Initialize(mapload)
 	. = ..()
-	initial_density = density
 	GLOB.simple_animals[AIStatus] += src
 	if(gender == PLURAL)
 		gender = pick(MALE, FEMALE)
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -84,6 +84,7 @@
 	var/del_on_death = 0 //causes mob to be deleted on death, useful for mobs that spawn lootable corpses
 	var/deathmessage = ""
 	var/death_sound = null //The sound played on death
+	var/initial_density //Whether the density of a simple animal starts as 0 or 1. Used for revive().
 
 	var/allow_movement_on_non_turfs = FALSE
 
@@ -101,6 +102,7 @@
 
 /mob/living/simple_animal/Initialize(mapload)
 	. = ..()
+	initial_density = density
 	GLOB.simple_animals[AIStatus] += src
 	if(gender == PLURAL)
 		gender = pick(MALE, FEMALE)
@@ -336,7 +338,8 @@
 
 /mob/living/simple_animal/revive()
 	..()
-	density = 1
+	if(initial_density)
+		density = 1
 
 /mob/living/simple_animal/death(gibbed)
 	// Only execute the below if we successfully died
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1098,25 +1098,23 @@ var/list/slot_equipment_priority = list( \
 				if(!L.key)
 					creatures += L
 		var/picked = input("Please select an NPC to respawn as", "Respawn as NPC")  as null|anything in creatures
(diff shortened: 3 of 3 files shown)


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -84,6 +84,7 @@
 	var/del_on_death = 0 //causes mob to be deleted on death, useful for mobs that spawn lootable corpses
 	var/deathmessage = ""
 	var/death_sound = null //The sound played on death
+	var/initial_density //Whether the density of a simple animal starts as 0 or 1. Used for revive().
```
variableundefined: Two thing.

You can use initial(variable) to find its initial value. It is a built in proc in BYOND that find its compile time value. So this is unnecessary

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -334,6 +336,11 @@
 		for(var/i in loot)
 			new i(loc)
 
+/mob/living/simple_animal/revive()
+	..()
+	if(initial_density)
```
variableundefined: This is band-aid over the problem. Somewhere in the code, someone messed up setting the density. This means that if the density change for whatever legit reason (For example, a future potion), it will always get set back to its original value on revive.

This just introduces another bug for the future

PidgeyThePirate: Would changing initial_density to living_density and setting it in death() be better? Death() is where the density is set to 0 and it isn't reset on revive.

variableundefined: Looked at what tg did.

They just set density = initial(density) in revive.

The approach is OK but not the method.

initial is a proc (function) that returns the initial, compile time value of something. There's no need to have initial_density set in initialize

code/modules/mining/equipment/lazarus_injector.dm
```diff
@@ -96,7 +96,7 @@
 	else
 		if("neutral" in S.faction)
 			S.forceMove(src)
-			S.name = "[M.name]'s [initial(S.name)]"
+			S.name = "[M.name]&#39s [initial(S.name)]"
```
variableundefined: Just tested on live.

I can interact with a lazarus capsuled runtime without any issue. This make the name not display properly visually but fine in text panel.

On live, both were displaying just fine

code/modules/mob/mob.dm
```diff
@@ -1098,25 +1098,23 @@ var/list/slot_equipment_priority = list( \
...
-					GLOB.respawnable_list += usr
+				become_mouse(O)
 			else
 				var/mob/living/NPC = picked
 				if(istype(NPC) && !NPC.key)
-					GLOB.respawnable_list -= usr
-					NPC.key = key
-					spawn(5)
-						GLOB.respawnable_list += usr
+					if(mind)
```
variableundefined: There's multiple thing wrong with this.

Antagonist status and traitor memory, objectives etc. is stored in the mind.

This means:

image


I think it's self-evident what goes wrong

Not to mention you never removed them from the respawnable_list which is going to be problematic when some proc look for a respawnable and see someone who is already a mouse.

Remove this fix, this creates more problems than it solve.

code/modules/mining/equipment/lazarus_injector.dm
PidgeyThePirate: By interact I mean give karma or examine. If you try to give karma to your capsuled pet it will try to give the karma to you. Similarly, if you examine your capsuled pet it will examine you instead. Are you sure that's fine, because I replicated the issue myself when I was trying to fix it.

code/modules/mob/mob.dm
PidgeyThePirate: Fox told me that you should never destroy the mind of a ghost of you can help it, when I used code based on this very function in the diona nymph thing. So why is it acceptable here and not there? Obviously mice keeping antag objectives is a problem, but given the circumstance I'm pretty sure my diona pods would do the same thing, and that would be even worse.

variableundefined: That's something that I cannot comment on as I do not know the specific of this PR. All I know is that this fix create more problems than it solve.

In fact, I don't think this even solve any issue at all. This one should be removed. You should avoid refactoring things you do not understand. Even I am not confident in refactoring datum, and you, for one, did not know that mind was responsible for storing antag status and abilities.

code/modules/mining/equipment/lazarus_injector.dm
Ty-Omaha: This is so atypical, there is literally nothing else in the game that uses `&#39s` written out like this. I don't think this is a proper way to fix anything, I'm pretty sure it's related to #12338 which should be fixed in its own.

variableundefined:
Above. Something somewhere probably stripped the input improperly leading to karma problem.

code/modules/mob/mob.dm
```diff
@@ -1098,23 +1098,25 @@ var/list/slot_equipment_priority = list( \
 				if(!L.key)
 					creatures += L
 		var/picked = input("Please select an NPC to respawn as", "Respawn as NPC")  as null|anything in creatures
-		var/mob/dead/observer/O = src
 		switch(picked)
 			if("Mouse")
-				become_mouse(O)
+				GLOB.respawnable_list -= usr
+				become_mouse()
+				spawn(5)
```
KasparoVy: Sounds like a job for addtimer callback!

code/modules/mob/mob.dm
```diff
@@ -1098,23 +1098,25 @@ var/list/slot_equipment_priority = list( \
...
 			else
 				var/mob/living/NPC = picked
 				if(istype(NPC) && !NPC.key)
-					if(mind)
-						O.mind.transfer_to(picked)
-						GLOB.non_respawnable_keys -= ckey
-						O.reenter_corpse()
+					GLOB.respawnable_list -= usr
+					NPC.key = key
+					spawn(5)
```
KasparoVy: Ditto

PidgeyThePirate: I think you'll find that that is the original file. I'm not keen on touching it again, for the time being at least.

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -334,6 +334,10 @@
 		for(var/i in loot)
 			new i(loc)
 
+/mob/living/simple_animal/revive()
+	..()
+	density = initial(density)
```
Fox-McCloud: I'd still rather figure out why density is breaking.

PidgeyThePirate: It's because density is set to 0 on death(), so people can walk over the corpse, I assume. I just reset the thing that previously wasn't reset after revival.

variableundefined: I don't know of a better solution than this at the moment, it is what TG did to fix the issue.

PR #12458, opened 2019-09-30
Issue #5066 - Abductor silencer now properly silences radios temporarily.

### LINKED ISSUE #5066: The abductor silencer does not, in fact, silence radio headsets.
**Problem Description**:
During an abductor round I hit several people with the abductor silencer and yet they were still able to call for help on the radio despite me hitting them BEFORE they called for help.

For their calls to help not to be broadcast over the radio once they were silenced.

**What happened instead**:
They called for help, crew busted in and nearly dunked me several times.

The abductors have no way to prevent calls for help.  It's a severe nerf to their abilities given that the item is specifically meant to do one thing and it doesn't actually do that thing.

Be abductor.  Run in, stun someone, use abductor silencer, let them try calling for help, get dunked on.

anything else out of the ordinary happening?)**: Gamemode was changelings, but the abductor team got going about 10 minutes into the round.  Nothing else special that I noted about the round, but I was de-admin'd the whole time so I could focus on my abducting.

IMAGES:
https://i.gyazo.com/d9f3d693873d776d194dceba096d908a.png
https://i.gyazo.com/d65ab472b090096881f7d49da8c11a50.png

PidgeyThePirate:
## What Does This PR Do
Fixes #5066 - The silencer now does an emp_act on all the target's radios as well as turning broadcasting off, knocking them out for 10-20 seconds.

## Why It's Good For The Game
The silencer doesn't seem to have ever worked. Now it does.

## Changelog
:cl:
fix: The abductor's silencer now actually silences radios for 10-20 seconds.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/abduction/abduction_gear.dm
+++ b/code/game/gamemodes/miniantags/abduction/abduction_gear.dm
@@ -268,8 +268,9 @@
 	for(var/obj/I in all_items)
 		if(istype(I, /obj/item/radio))
 			var/obj/item/radio/r = I
+			I.emp_act(1)
 			r.listening = 0
-			if(!istype(I, /obj/item/radio/headset))
+			if(istype(I, /obj/item/radio/headset))
 				r.broadcasting = 0 //goddamned headset hacks
 
 /obj/item/abductor/mind_device
--- a/code/game/gamemodes/miniantags/abduction/abduction_gear.dm
+++ b/code/game/gamemodes/miniantags/abduction/abduction_gear.dm
@@ -267,11 +267,9 @@
 
 	for(var/obj/I in all_items)
 		if(istype(I, /obj/item/radio))
-			var/obj/item/radio/r = I
-			I.emp_act(1)
-			r.listening = 0
-			if(istype(I, /obj/item/radio/headset))
-				r.broadcasting = 0 //goddamned headset hacks
+			var/obj/item/radio/R = I
+			R.listening = 0 // Prevents the radio from buzzing due to the EMP, preserving possible stealthiness.
+			R.emp_act(1)
 
 /obj/item/abductor/mind_device
 	name = "mental interface device"


Fox-McCloud:
This is papering over a bug and is absolutely the worst kind of bugfixing.

It's like when species is `null` and throws a runtime because of it, so you throw in an `if(species)` check to prevent it from runtiming.

Congrats, you fixed the runtime from showing up, but you didn't fix the larger problem of why species is null to begin with, which is the *real* problem.

So unless there's a specific reason why setting the vars on these radios isn't acceptable/doesn't work at all, this is papering it over, especially considering that the radio silencers *do* work a good portion of the time.

PidgeyThePirate:
As far as I can tell all the silencer does normally is turn the microphone and speakers off, which the microphone usually is anyway. Whether the microphone or speaker is on appears to have no bearing on whether or not someone can talk through it using Say. After testing I found that I could still talk through the headset perfectly fine once I got hit with the silencer, even though the microphone and speaker were off.

This is I believe the most elegant way to fix the issue as it uses a function that makes sense in the context of the silencer device itself, and as a bonus it keeps the speaker and microphone off after the EMP runs out, all without adding more code that's only used for that one interaction. You're against that, right?

Markolie:
Yup, I think this emp_act fix is the best solution as well as disabling the microphone doesn't actually prevent _usage_ of the microphone, it just stops it from broadcasting _everything_ in an X radius of being broadcast through it, similar to intercoms.

code/game/gamemodes/miniantags/abduction/abduction_gear.dm
```diff
@@ -268,8 +268,9 @@
 	for(var/obj/I in all_items)
 		if(istype(I, /obj/item/radio))
 			var/obj/item/radio/r = I
+			I.emp_act(1)
 			r.listening = 0
-			if(!istype(I, /obj/item/radio/headset))
```
variableundefined: EMP_act already set broadcasting to false

PidgeyThePirate:
Removed broadcasting=0 and changed the order of the remaining 2 lines so that listening = 0 comes first. That stops the radio from buzzing visibly to everyone in sight due to the EMP act.

PR #12461, opened 2019-09-30
Issue #12223 Ripley Ore Pickup Fix

### LINKED ISSUE #12223: Ripley's no longer auto-pickup all ore's they mine.
**Issue Description**:
When a Ripley with a drill mines a 3x1 section of ore with an ore crate in its cargo, it will pick up 
the ore directly in front of it, but it will not the ores adjacent to it. You will have to turn and pick up each ore individually.

I expected for all the ores from the walls mined to instantly go into the ore box. 

**What happened instead**:
The ores from the walls mined adjacent to the Ripley fell to the floor.

Slows down the mining process for Ripley's greatly, as you have to constantly turn about and pickup the loose ore.

Get a Ripley with a mining scanner, drill, and clamp. Pick up an ore box with the clamp. Find a wall with 3 ores adjacent. Moving into the wall in the Ripley with the drill equipped. Witness as the front facing ore is mined, but the adjacent ores are not.

Almost 100% the new mech changes, intended to make mech mining faster. #12220

PidgeyThePirate:
## What Does This PR Do
Fixes #12223: the collect_ore() event now correctly fires when drilling, causing the Ripley to pick up all its trash.

## Changelog
:cl:
fix: The Ripley now collects all the ore it unearths after drilling. 
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/mining_tools.dm
+++ b/code/game/mecha/equipment/tools/mining_tools.dm
@@ -82,7 +82,7 @@
 	drill.move_ores()
 
 /obj/item/mecha_parts/mecha_equipment/drill/proc/move_ores()
-	if(locate(/obj/item/mecha_parts/mecha_equipment/hydraulic_clamp) in chassis.equipment && istype(chassis, /obj/mecha/working/ripley))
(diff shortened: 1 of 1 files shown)


JKnutson101:
Yay! I for one love mining in a Ripley and this bug kinda killed the experience!

KasparoVy:
A set of brackets, what a classic

Issue #12454, opened 2019-09-29
The proc cmd_admin_pm needs a Refactor


Ty-Omaha:
**Issue Description**:
The proc `cmd_admin_pm` is all over the place

**What did you expect to happen**: 
For legible code, allowing the ability to easily modify it.

**What happened instead**:
It's too confusing

**Why is this bad/What are the consequences**:
Because it hinders progress

**Extra information**:
https://github.com/ParadiseSS13/Paradise/blob/d2e2c91eca9261091984cd372e849fed7710c1b7/code/modules/admin/verbs/adminpm.dm#L61

AffectedArc07:
Requesting a refactor is not a valid issue.

PR #12462, opened 2019-09-30
[needs review]Refactors cmd_admin_pm and adds autotake to adminhelps

Ty-Omaha:
What Does This PR Do
Refactors cmd_admin_pm proc and adds an autotake and confirmation for admins to avoid duplicate PMing people

Fixes: #12454

## Images of changes

after removing permissions from both accounts:

:cl:
tweak: Refactor'd cmd_admin_pm proc, aka admin PM code.
add: Autotake and double PM check for admins in case another admin has already responded to the player.
/ 🆑

### DIFF
--- a/code/modules/admin/verbs/adminpm.dm
+++ b/code/modules/admin/verbs/adminpm.dm
@@ -157,7 +157,7 @@
 
 
 	var/emoji_msg = "<span class='emoji_enabled'>[msg]</span>"
-	recieve_message = "<span class='[recieve_span]'>[type] from-<b>[recieve_pm_type][ranked ? key_name(src, TRUE, type) : key_name_hidden(src, TRUE, type)]</b>: [emoji_msg]</span>"
+	recieve_message = "<span class='[recieve_span]'>[type] from-<b>[recieve_pm_type][C.holder ? key_name(src, TRUE, type) : key_name_hidden(src, TRUE, type)]</b>: [emoji_msg]</span>"
 	to_chat(C, recieve_message)
 	var/ping_link = check_rights(R_ADMIN, 0, mob) ? "(<a href='?src=[pm_tracker.UID()];ping=[C.key]'>PING</a>)" : ""
 	var/window_link = "(<a href='?src=[pm_tracker.UID()];newtitle=[C.key]'>WINDOW</a>)"
--- a/code/modules/admin/verbs/adminpm.dm
+++ b/code/modules/admin/verbs/adminpm.dm
@@ -71,44 +71,48 @@
 	else if(istype(whom,/client))
 		C = whom
 
-	if(!C)
-		if(holder)
-			to_chat(src, "<span class='danger'>Error: Private-Message: Client not found.</span>")
-		else
-			adminhelp(msg)	//admin we are replying to left. adminhelp instead
-		return
-
-	/*if(C && C.last_pm_recieved + config.simultaneous_pm_warning_timeout > world.time && holder)
-		//send a warning to admins, but have a delay popup for mods
-		if(holder.rights & R_ADMIN)
-			to_chat(src, "<span class='danger'>Simultaneous PMs warning:</span> that player has been PM'd in the last [config.simultaneous_pm_warning_timeout / 10] seconds by: [C.ckey_last_pm]")
-		else
-			if(alert("That player has been PM'd in the last [config.simultaneous_pm_warning_timeout / 10] seconds by: [C.ckey_last_pm]","Simultaneous PMs warning","Continue","Cancel") == "Cancel")
-				return*/
-
+	//is this person an admin/mentor? is this person either an admin or a mentor or hold some sort of rank, if so, set them as the value. if not, set as false.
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/adminpm.dm
```diff
@@ -164,14 +162,9 @@
...
 	to_chat(src, "<span class='pmsend'>[send_pm_type][type] to-<b>[holder ? key_name(C, TRUE, type) : key_name_hidden(C, TRUE, type)]</b>: [emoji_msg]</span> [ping_link] [window_link]")
-
-	/*if(holder && !C.holder)
-		C.last_pm_recieved = world.time
-		C.ckey_last_pm = ckey*/
-
 	//play the recieving admin the adminhelp sound (if they have them enabled)
 	//non-admins always hear the sound, as they cannot toggle it
-	if((!C.holder) || (C.prefs.sound & SOUND_ADMINHELP))
+	if((!admin) || (C.prefs.sound & SOUND_ADMINHELP))
```
Ty-Omaha: I'm really logic'd out. C.holder is not src, right? So the verb admin, which is based off of src, would not be applicable right?

code/modules/admin/verbs/adminpm.dm
```diff
@@ -71,79 +71,77 @@
...
+					if(alert("[T.staffAssigned] is already responding, are you sure you want to continue?", ,"Yes","No") != "Yes") //ask them if they wanna continue
+						return //if they don't, then return
+				if(!T.staffAssigned) //verify that the person we are dealing with is an admin to avoid setting the assignedStaff as nobody (FALSE)
+					T.assignStaff(admin) //If nobody is assigned, we want to assign the admin to avoid duplicate PMing
 
 	//clean the message if it's not sent by a high-rank admin
 	if(!check_rights(R_SERVER|R_DEBUG,0))
 		msg = sanitize_simple(copytext(msg,1,MAX_MESSAGE_LEN))
-		if(!msg)
-			return
```
Ty-Omaha: This should probably be re-added now that I am looking at it, since it's after the sanitize

code/modules/admin/verbs/adminpm.dm
```diff
@@ -71,79 +71,77 @@
...
-		//mod PMs are maroon
-		//PMs sent from admins and mods display their rank
-		if(holder)
-			if(check_rights(R_MOD|R_MENTOR,0) && !check_rights(R_ADMIN,0))
-				recieve_span = "mentorhelp"
-			else
-				recieve_span = "adminhelp"
-			send_pm_type = holder.rank + " "
-			recieve_pm_type = holder.rank
+	if(ranked)
```
Ty-Omaha: this can probably be removed and added in an and operator to the line below

Ty-Omaha: no but then what happens to line 128 and 129

Ty-Omaha: and 131. nevermind.

code/modules/admin/verbs/adminpm.dm
Ty-Omaha: if not then this should be changed and the commit fix should be reverted

code/modules/admin/verbs/adminpm.dm
Ty-Omaha: eh I guess it could stay the way it is

code/modules/admin/verbs/adminpm.dm
```diff
@@ -71,79 +71,77 @@
...
+
+	if(tickets)
+		for(var/datum/ticket/T in tickets)
+			T.addResponse(src, msg) // Add this response to their open tickets.
+			if(admin && (type != "Mentorhelp")) //See if we are dealing with an admin and avoid autotaking mentorhelps. Handle auto take when admin sends PM  to avoid duplicate PMing
+				if(T.staffAssigned && (T.staffAssigned != admin)) //if there is currently an admin assigned and if so, is this admin not the admin assigned to the ticket?
+					if(alert("[T.staffAssigned] is already responding, are you sure you want to continue?", ,"Yes","No") != "Yes") //ask them if they wanna continue
+						return //if they don't, then return
+				if(!T.staffAssigned) //verify that the person we are dealing with is an admin to avoid setting the assignedStaff as nobody (FALSE)
+					T.assignStaff(admin) //If nobody is assigned, we want to assign the admin to avoid duplicate PMing
```
variableundefined: Would it be better to use takeTicket instead of assignStaff? 

AssignStaff works well and actually assign them, but offer no indicator to the player that the admin in question is handling them. It also don't inform admins / mentors that it is being handled.

Ty-Omaha: oh yup

code/modules/admin/verbs/adminpm.dm
Ty-Omaha: yeah i think i need to experiment with having "C.check_rights(stuff)" for this instead of src

Ty-Omaha:
@variableundefined will do it later today

Ty-Omaha:
this needs more testing and im too busy at the moment

Issue #12306, opened 2019-09-15
Lethal Injections Syringes aren't lethal


Ty-Omaha:
**Issue Description**:
Lethal injection syringes don't kill

**What did you expect to happen**: 
For someone to die when injected with a lethal injection syringe
They didnt

**When did the problem start happening**:
Probably newcrit

TheSardele:
Having tested this locally, a single syringe is just about enough to kill someone, though not very humanely. Judging from the discord, the reason that someone survived it was that they had a healing virus. Either way, they should probably be made more lethal.

BakKeener:
Yeah the reason they survived was purely due to healing virus which was overly stated that round. Also Cyanide is  still a poor chem for murder.

PR #12460, opened 2019-09-30
Improved Lethal Injections

PidgeyThePirate:
## What Does This PR Do
Fixes #12306
- The lethal injection syringe used for nothing else is no more.
- Added a set of 3 numbered syringes used for lethal injection for a faster and more humane(-looking) kill.
- Put them all in a kit in a custom box with an accompanying leaflet and replaced the old lethal syringes with them.
- There was one empty ld50 syringe on metastation z3. I replaced it with a normal one. The ld50 syringe sucked anyway.

## Why It's Good For The Game
The lethal syringe was made just for the purpose of giving lethal injections and it was bad for both executions and as an item for antags to steal and use. These new kits have some fairly nasty stuff in them, especially Thiopental, which may make them worth stealing (but maaaybe too worth stealing? Might want to make it so there's only one kit in the locker and another in the armory or something).

Also 3 syringes means a higher chance of botched executions, which is always... fun?

## Images of changes
[The contents of each syringe and their position.](https://imgur.com/a/m704s7O)

## Changelog
:cl:
add: A new box, a new paper, 3 new syringes as part of a Lethal Injection Kit, two of which have been placed in security instead of the two lethal injection syringes.
del: The snowflakey lethal injection syringe.
tweak: replaced the empty lethal injection syringe on metastation z3 with a regular one.
/:cl:

PR #12467, opened 2019-09-30
Syringe Refactor. Overhauls Lethal Injection Syringe

Fox-McCloud:
Refactors syringes, mostly from TG, not that it changes much.

This mostly ensures that LD50 syringes aren't their own snowflaked reagent container, and, instead, lethal syringes are just regular syringes with 50 units that can inject them all at once.

other things
- Fixes in-hand of syringes not updating on inject/drawing
- plugs in piercing darts (not used)
- Refactors the `is_injectable` and `is_drawable` to pass the user argument
- Fixes syringse not having metal/glass
- Lethal injection syringes now have toxin, cyanide, pancuronium, fluorine, and fluorosulfuric acid

Guess this is an alternative to: https://github.com/ParadiseSS13/Paradise/pull/12460

fixes: https://github.com/ParadiseSS13/Paradise/issues/12306

:cl: Fox McCloud
fix: Fix syringe in-hand icon not updating
fix: Fixes syringes not having materials
tweak: Lethal injection syringes should be a bit more...lethal
/:cl:


PidgeyThePirate:
What the hell, fox? you're going to inject a lethal injection recipient with ACID? I can understand you not wanting to use traitor chems, but I chose that formulation and syringe setup in order to quickly knock out the subject before getting down to the business of killing them as quickly as possible without leaving any marks. You know, like I'm pretending nanotrasen aren't a bunch of psychopaths.

Your syringe refactor is better, but you should go back to the drawing board on those chems. In my opinion you need a knockout drug at least (ether will serve, I guess) and chems that kill with _toxin damage_ so they don't visibly harm the subject.

Ty-Omaha:
@Fox-McCloud  Fixes #12306

Fox-McCloud:
Nanotrasen allows execution by electric chair, asphyxiation, firing squad, and gas inhalation as well.

Most of these are pretty brutal, if not downright cruel, not to mention physically harms the person in question; none of the execution methods are really "humane".

Ty-Omaha:
I don't think the chem mix is that important really, as long as it will kill people not just _barely_ kill people.

NT has a chamber where a guy sits and it opens to space to choke them out, they're not really a pillar of humanity.

PidgeyThePirate:
Nah, you're right. I'm probably just salty because I spent more time on my version than I should have. I researched lethal injection methods and experimented with several mixes to find something that could (usually, because new crit) kill painlessly within 2 minutes and 3 standard syringes. That's not a lot of room to work with, I could probably have done better if I used bigger syringes.

TDSSS:
NT is not nice, yes. But if they go through the trouble of making lethal injections instead of just using their cheap, rechargable lasers for an execution I think 'criminal isn't in horrible agony and/or literally melting' is still something they'd pay attention to. It's not like they lose much by putting their victim to sleep first. If nothing else, do it so it doesn't horribly traumatize the people carrying out or witnessing the execution.

Carthusia:
I feel the purpose of the Lethal injection is to be the more "Humane" approach to all NTs methods of exicutions. 

Dont wana get shot, spaced, suffocated, but a nice sleepy time injection that you wont feel a thing for? Seems appealing.

Unless that injection begins to MELT YOU, then why even choose it in the first place, at that point a forced injection would be the absolute cruelest thing anyone could do to a player IC death wise.

JKnutson101:
I do feel like since the method of execution doesn't really matter except for RP and flavor, the lethal injection would be the choice of someone that doesn't want a painful death. While effective, acid isn't exactly painless. It's not a big deal though.

PR #12469, opened 2019-09-30
Give Heads of Staff Single-Use Autoimplanters instead of raw implants.

craftxbox:
## What Does This PR Do
Give Heads of Staff Single-Use Autoimplanters instead of raw implants.

## Why It's Good For The Game
Makes it easier for heads to use their implants, encouraging them to use them more rather than them sitting unused in most shifts.

## Changelog
:cl:
add: added pre-loaded autoimplanters with huds
add: functionality to have autoimplanters select between implants
add: var to restrict the amount of times an autoimplanter may be used
add: var to set the amount of time it takes for an autoimplanter to implant
tweak: replaced head's implants with autoimplanters
/:cl:

Do let me know if the captain having a hud on spawn isnt something that's wanted and i'll take it out.

### DIFF
--- a/code/modules/surgery/organs/autoimplanter.dm
+++ b/code/modules/surgery/organs/autoimplanter.dm
@@ -6,14 +6,19 @@
 	item_state = "walkietalkie"//left as this so as to intentionally not have inhands
 	w_class = WEIGHT_CLASS_SMALL
 	usesound = 'sound/weapons/circsawhit.ogg'
+	var/uses = -1  // -1 is infinite as far as we care
 	var/obj/item/organ/internal/cyberimp/storedorgan
 
 /obj/item/autoimplanter/attack_self(mob/user)//when the object it used...
+	if(uses == 0) // uses check before stored check here because a single use implanter probably wont have an implant in it after its been used up.
+		to_chat(user, "<span class='notice'>[src] appears to be permanently jammed.</span>")
+		return
 	if(!storedorgan)
 		to_chat(user, "<span class='notice'>[src] currently has no implant stored.</span>")
 		return
 	storedorgan.insert(user)//insert stored organ into the user
 	user.visible_message("<span class='notice'>[user] presses a button on [src], and you hear a short mechanical noise.</span>", "<span class='notice'>You feel a sharp sting as [src] plunges into your body.</span>")
+	uses = uses > 0 ? uses-1 : uses
 	playsound(get_turf(user), usesound, 50, 1)
 	storedorgan = null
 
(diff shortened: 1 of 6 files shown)


Fox-McCloud:
TG implemented this a bit more gracefully, IMO. I'd do that.

I'm still opposed to giving all these individuals easy-mode augments though.

TDSSS:
Eh, not a fan of this. If people are too lazy to get a simple 5 minute implantation surgery then they shouldn't have implants.

Shockpoint:
I personally don't mind this at all, but it does encourage laziness. However, for some jobs like HoS or CMO, this would be prime as it is as even a good CMO will be too swamped carrying their department of baby surgeons who can't even figure out how a scalpel works.

However, would this give the HoS a pair of regular sunglasses as they would have their HUD implanted? It would be very redundant to have the HUD implanted but also wear HUDglasses. Same with the Blueshield if they are getting one. (Would suggest adding a special one that lets them choose between Med/Sec HUD as they currently get both in their locker)

craftxbox:
I'll set those up tomorrow, never thought about that

JKnutson101:
I think giving free implants with zero outside interaction, while convenient, kills off some potential interesting situations, like the surgeon being a traitor using a mindslave implant during the surgery.

I do also understand these are hardly ever used (outside of Kidan that need them) so the amount of killed off potential is ultimately really low. 

Only problem I can think of is just giving out free upgrades to heads of staff with zero investment of time or effort. The implants themselves I don't think would be too strong, but something about giving people free upgrades feels 'wrong' for lack of a better word. Ultimately I don't mind if this gets in, best of luck.

craftxbox:
I can make them have a timer to complete instead of being instantaneous

code/modules/surgery/organs/autoimplanter.dm
```diff
@@ -7,18 +7,27 @@
...
 	var/obj/item/organ/internal/cyberimp/storedorgan
+	var/possible_implants = list() //can put implant type references in here to let the user pick which one they want
 
 /obj/item/autoimplanter/attack_self(mob/user)//when the object it used...
 	if(uses == 0) // uses check before stored check here because a single use implanter probably wont have an implant in it after its been used up.
 		to_chat(user, "<span class='notice'>[src] appears to be permanently jammed.</span>")
 		return
-	if(!storedorgan)
+	if(!storedorgan && length(possible_implants) > 0)
+		var/selectedimplant = input("Select what to program the implant as.","Program Implant",/obj/item/organ/internal/cyberimp/eyes/hud/medical) in possible_implants
```
craftxbox: This results in showing the user type paths, and I'm not quite sure how I fix it.

Fox-McCloud:
img

Upvoting your own PR.

Shame. Shame.

Carthusia:
I mean, if your wishing to add auto implanters so they can just instantly implant their provided implant, why not just make them start with the implant already in their head?

Not for or against the change, just feel like if your going to go that route, may as well cut out the extra step.

variableundefined:
It's a good thing that you started making PRs by doing something simple like this, namely adding a new feature.

Unfortunately we cannot accept this as it gives the head of staffs easy augments.

PR #11166, opened 2019-03-29
Highlight string now uses regex

craftxbox:
**What does this PR do:**
Highlight string in the chat panel now uses regex

**Changelog:**
:cl:
tweak: Highlight string uses regex
del: Removed jquery mark
/:cl:

PR #12474, opened 2019-10-01
Highlight regex 2 Electric boogaloo

craftxbox:
Remake of #11166 

## What Does This PR Do
Adds an option to use regex for word highlighting, disabled by default for compatibility

## Why It's Good For The Game
Finer control over what gets highlighted, no more false highlights.

## Changelog
:cl:
add: Vanilla JS Regex for chat highlighting
del: Removed jquery mark
/:cl:

Requesting that this be testmerged instead of fullmerged immediately as its had issues in the past.


goon/browserassets/js/browserOutput.js
```diff
@@ -965,18 +971,16 @@ $(function() {
...
 		var termInputs = '';
 		for (var i = 0; i < opts.highlightLimit; i++) {
 			termInputs += '<div><input type="text" name="highlightTermInput'+i+'" id="highlightTermInput'+i+'" class="highlightTermInput'+i+'" maxlength="255" value="'+(opts.highlightTerms[i] ? opts.highlightTerms[i] : '')+'" /></div>';
 		}
 		var popupContent = '<div class="head">String Highlighting</div>' +
 			'<div class="highlightPopup" id="highlightPopup">' +
-				'<div>Choose up to '+opts.highlightLimit+' strings that will highlight the line when they appear in chat.</div>' +
+				'<div>Choose up to '+opts.highlightLimit+' strings that will highlight the line when they appear in chat.<br>'+
+		    			'<input name="highlightRegex" id="highlightRegexEnable" type="checkbox">Enable Regex</input>'+
+		    		'<br><a href="" onclick="window.open(\'https://nanotrasen.se/wiki/index.php/Guide_to_Regex\')">See here for details</a></div>' +
```
craftxbox: As like the last pr, I wasnt able to figure out how to open in anything but IE.
Additionally, this wiki page doesnt exist, and needs to be created, or at the very least redirect to a regex reference, probably like http://www.byond.com/docs/ref/#/{notes}/regex

goon/browserassets/js/browserOutput.js
```diff
@@ -157,13 +158,13 @@ function setHighlightColor(match) {
...
-		return
-	}
 	for (var i = 0; i < opts.highlightTerms.length; i++) { //Each highlight term
 		if(opts.highlightTerms[i]) {
-			element.mark(opts.highlightTerms[i], {"element" : "span", "each" : setHighlightColor});
+			var rexp = new RegExp("("+opts.highlightTerms[i]+")","gmi")
+			if(!opts.highlightRegexEnable){
+				rexp = new RegExp("("+opts.highlightTerms[i].replace(new RegExp("([^a-zA-Z0-9])","gmi"),"\\$1")+")","gmi")
+			}
+			el.innerHTML = el.innerHTML.replace(rexp,"<span style=\"background-color:"+opts.highlightColor+"\">$1</span>")
```
Tayyyyyyy: Should this be `innerText` and not `innerHTML`? Does this break html tags if you set the regex to something like "span" or "warning"?

craftxbox: Thats true, however if you're using regex you should understand that its to be expected if you dont write your regex correctly, ie: if you want to highlight span or warning without fucking the html, youd \b(span|warning)\b , where \b is the token for word boundary.

Tayyyyyyy: That doesn't sound like the typical user case. Most users don't have an understanding of the HTML that goes into the chat window, and seeing as there's no way to view the HTML content of the chat window without code diving, even the users that would want to change it probably wouldn't bother. Most users just want to highlight "AI" when "AI" is spoken, and will be confused when something like that breaks their window.

Also, this basically gives the user the ability to insert any html they want into their chat window. While that's not really a genuine security risk, Byond right now as far as I know doesn't let you do this. My feeling is that there's no need to edit the html so why complicate things? Not to mention that trying to parse html with regex is not a great idea: https://stackoverflow.com/questions/6751105/why-its-not-possible-to-use-regex-to-parse-html-xml-a-formal-explanation-in-la

Have you tried this with `innerText` instead? I'm almost certain that would be a better user experience for the vast majority of our users.

craftxbox: This wont break anyone's previously defined highlights, Regex is an option here that is effectively disabled by default, And no, you cannot insert any html anywhere, as this is a replace with a predefined replacement, therefore not making it an issue at all. The point is less of parsing the html, and more of parsing around while preserving the html.

craftxbox:
The main question i have at the moment is wether its expected to capture words within other words, ie, **AI** in M**AI**ntenance, or if that would be an unexpected behavior and we should assume the strings people are using as highlights are on word boundaries when the regex option is unticked

goon/browserassets/js/browserOutput.js
Tayyyyyyy: > And no, you cannot insert any html anywhere, as this is a replace with a predefined replacement

You're right, I forgot about that.

I still don't see any use case for trying to read the HTML itself.

craftxbox: You can use noncapturing groups to anchor to a specific location before reading ahead when you want to highlight something, but its dynamic and unknown, which is where reading the html would come in handy.

moxian: Right now highlighting "warning" would break all the warning classes since `<span class="warning">the mouse nuzzles you</span>`` becomes `<span class="<span style="background-color:yellow">warning</span>">the mouse nuzzles you</span>`.
In other words, would break spectacularly.

So instead of matching HTML you should really walk over DOM tree anchored on the `el` and and only mess with text(!) nodes by wrapping them in the colored span.

Separately, if the eventual goal of this PR is to highlight all warnings - you be to add custom CSS that would augment the `warning` class. (That custom CSS supportmight already exist? I'm unsure.. How does dark chat theme work? )

craftxbox: Totally forgot i could just match on innerText when the whole regex bit is off and innerHtml when its on. I brain.

moxian: You would still have the issue of people searching for `\bwarning\b` (or, hell, `war`)in regex mode would have their HTML messed up.

I strongly suggest you don't match on HTML, in either mode.

craftxbox: Limiting to the text only severely limits the flexibility goal of the system in the first place.

craftxbox:
thanks github

goon/browserassets/js/browserOutput.js
```diff
@@ -157,13 +158,15 @@ function setHighlightColor(match) {
...
 	for (var i = 0; i < opts.highlightTerms.length; i++) { //Each highlight term
 		if(opts.highlightTerms[i]) {
-			element.mark(opts.highlightTerms[i], {"element" : "span", "each" : setHighlightColor});
+			var rexp = new RegExp(opts.highlightTerms[i],"gmi")
+			if(!opts.highlightRegexEnable){
+				rexp = new RegExp("("+opts.highlightTerms[i].replace(new RegExp("([^a-zA-Z0-9])","gmi"),"\\$1")+")","gmi")
+				el.innerText = el.innerText.replace(rexp,"<span style=\"background-color:"+opts.highlightColor+"\">$1</span>") //disabling regex disables html matching too
+				continue;
+			}
+			el.innerHTML = el.innerHTML.replace(rexp,"<span style=\"background-color:"+opts.highlightColor+"\">$0</span>") //i cant figure out a proper, non snowflakey way to let people select the group that gets highlighted
```
craftxbox: If anyone has a good solution to the problem of the user being able to select the capture group they want to highlight without adding another textbox to specify, please let me know.  I've talked with a few people and had came up with only the following:
1: Use $1 (Not preferred because backreferences can cause issue)
2: Use the last group (Mildly arbitrary and can be confusing)
3: Use the group with the longest match (Could be odd if a group not meant to be highlighted is longer than the one that is highlighted, but i cant think of a situation that would cause this, also mildly arbitrary and confusing)
4: Use the first group with the most identical matches (seemingly the best option here but a bit snowlakey and confusing, ex: `(a)((ab))` would match `aab` as `a,ab,ab` and we would select `ab` since its repeated)

PR #12475, opened 2019-10-02
Issue #12473: Lavaland Winter Dome Vehicle Fix

### LINKED ISSUE #12473: Lavaland Winter Dome Key/ATV Bug
**Issue Description**:

Something replaced the keys for the ATV in the lavaland winter dome with a snowmobile key so both are useless now as the snowmobile key does not work for the ATV.

For the key to be an ATV key or for the vehicle to be a snowmobile for the snowmobile key.

**What happened instead**:

Its now a snowmobile key.

Makes one of the loot items from the dome pointless and to be ignored.

Spawn the winter dome, get the key and ATV from the dome, cannot use either.

Only recently as the first ahelp I received about it was this past Sunday.

PidgeyThePirate:
## What Does This PR Do
Fixes #12473: Replaced the ATV in the Lavaland winter dome with a snowmobile since the ATV key changed to a snowmobile key and I assume that's what the intended effect was.

## Changelog
:cl:
tweak: replaced the Lavaland winter dome ATV with a snowmobile.
fix: The snowmobile key in the winterdome now accompanies an actual snowmobile instead of an incompatible vehicle.
/:cl:

### DIFF
--- a/_maps/map_files/RandomRuins/LavaRuins/lavaland_biodome_winter.dmm
+++ b/_maps/map_files/RandomRuins/LavaRuins/lavaland_biodome_winter.dmm
@@ -158,7 +158,7 @@
 /turf/simulated/floor/plating/asteroid/snow/atmosphere,
 /area/ruin/powered/snow_biodome)
 "aJ" = (
-/obj/vehicle/atv,
+/obj/vehicle/snowmobile,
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
This is meant to be an ATV, not a snowmobile. Change the key type, not the vehicle

PidgeyThePirate:
Okay.

PR #12476, opened 2019-10-02
Prevent facehuggers from pouncing on masks that cover the mouth

Couls:
## What Does This PR Do

masks are now immune to facehugger cheese. also am ded please nerf
## Why It's Good For The Game

It makes xenos have to actually do more than machine gun toss facehuggers at you to win and heavily slows down their snowball rate

## Changelog
:cl:
tweak: facehuggers can no longer tear off masks that cover mouths
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/alien/special/facehugger.dm
+++ b/code/modules/mob/living/carbon/alien/special/facehugger.dm
@@ -129,6 +129,11 @@ var/const/MAX_ACTIVE_TIME = 400
 								"<span class='userdanger'>[src] smashes against [H]'s [H.head]!</span>")
 			Die()
 			return 0
+		else if(H.wear_mask && H.wear_mask.flags_cover & MASKCOVERSMOUTH)
+			H.visible_message("<span class='danger'>[src] smashes against [H]'s [H.wear_mask]!</span>", \
+								"<span class='userdanger'>[src] smashes against [H]'s [H.wear_mask]!</span>")
+			Die()
+			return 0
 	if(iscarbon(M))
 		var/mob/living/carbon/target = M
 		if(target.wear_mask)


Fox-McCloud:
Please link to the PR in question.

Couls:
@Fox-McCloud https://github.com/tgstation/tgstation/pull/25488

Fox-McCloud:
That's...not what that PR does at all.

The logic was borked and `is_mouth_covered` was never checked for, therefore, no matter WHAT you wore, your mask would get ripped off.

`is_mouth_covered` in that case has a `heads_only = TRUE` argument, which leads to: https://github.com/tgstation/tgstation/blob/8cfe0f6467a2b7ca815a61e3f37ab10879d168ee/code/modules/mob/living/carbon/carbon_defense.dm#L23

Only helmets that cover the mouth protect against facehuggers.

Couls:
Hmm guess you're right, regardless I still think gas masks should protect from facehuggers

pinatacolada:
Scarfs will prevent facehuggers
https://github.com/ParadiseSS13/Paradise/blob/7918692dd048890ddaebf932987aeb554a30d9e3/code/modules/clothing/masks/miscellaneous.dm#L254
Probably best to nerf the scarfs, they had it comming anyway

Couls:
I’ve never actually seen someone wear a mask scarf only the accessory ones, where can you get the mask version?

KasparoVy:
Never seen mask scarves either 🤔

pinatacolada:
Huh i thought these were the ones in the lockers in the dressing room, turns out theyre not, theyre not used at all as far as i can tell... Nevermind then

Denthamos:
If ALL mouth covering masks worked then the face huggers would be near pointless unless the player is already stunned and nested.  It should check for certain helmet types and not masks.

Fox-McCloud:
>If ALL mouth covering masks worked then the face huggers would be near pointless unless the player is already stunned and nested. It should check for certain helmet types and not masks.

That's currently what it does.

Couls:
xenos already have an acid spit why do they need a ranged facehugger toss attack ontop of that? They should keep the huggers in the nest and prepare victims by removing their masks prior

Pixelzilla:
Have you actually watched the movies? Facehuggers are pretty strong and generally /can/ get through some harder helmets. I dont understand why they'd have trouble getting through a gasmask, scarf, medical mask or anything like that.

Not all xenos have an acid spit. Only queen/sentinel do. Drones/Hunters are primarily melee with the latter having their pounce.

How would you propose a xeno transport a unstunned/fully awake person to the nest while not directly leading other crew there or letting that resisting person get away/scream for help? The stun doesn't last long enough to devour someone and devouring an awake individual means they can attack and kill you from within depending on weapons etc. If you injure the host too much they'll die since they take damage while being devoured and this also means they have less time left to be infected. Crew have to be alive for a period of time before becoming infected, if they die before that then there is no infection.

Facehuggers provide that ability to both stun/silence and that allows xenos to safely transport that host to the nest. Throwing huggers still depends on the crew being 1.) stupid/unprepared with proper facial protection 2.) fighting xenos alone or being overextended so other crew cant rescue you. A lot of the time if i throw a facehugger its almost always at someone who is not wearing proper facial protection, who is running /into/ the fight because they think its a GREAT idea and they end up being overextended and away from the support from other more prepared people. No one throws huggers at people who are wearing proper facial protection - it just wastes the hugger.

Couls:
The stun can be spammed and usually the xeno nest is known and approaching them is actually harder than locating them, every xeno can stun lock someone by spamming disarm intent and drones shouldn't be bringing people into the nest anyway

PR #12477, opened 2019-10-02
modify xenos tackle

Couls:
## What Does This PR Do

Part 2 of my proposed xenomorph nerfs, tweak the xeno disarm to only tackle down if the target is not holding an item
## Why It's Good For The Game

This allows crew to fight xenos with something other than a mech, it's still unsafe but it allows you to charge into battle with a shield and not be immediately knocked down and dragged away. adds to the struggle of fighting xenos and makes it a bit more interesting for both sides.

## Changelog
:cl:
balance: xenomorphs will no longer knock over people holding items, they are guaranteed to knock anyone not holding anything
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -421,20 +421,16 @@ emp_act
 			updatehealth("alien attack")
 
 		if(M.a_intent == INTENT_DISARM)
-			if(prob(80))
+			var/obj/item/I = get_active_hand()
+			if(I && drop_item())
+				playsound(loc, 'sound/weapons/slash.ogg', 25, 1, -1)
+				visible_message("<span class='danger'>[M] disarmed [src]!</span>")
+			else
 				var/obj/item/organ/external/affecting = get_organ(ran_zone(M.zone_sel.selecting))
 				playsound(loc, 'sound/weapons/pierce.ogg', 25, 1, -1)
 				apply_effect(5, WEAKEN, run_armor_check(affecting, "melee"))
 				add_attack_logs(M, src, "Alien tackled")
 				visible_message("<span class='danger'>[M] has tackled down [src]!</span>")
-			else
-				if(prob(99)) //this looks fucking stupid but it was previously 'var/randn = rand(1, 100); if(randn <= 99)'
-					playsound(loc, 'sound/weapons/slash.ogg', 25, 1, -1)
-					drop_item()
-					visible_message("<span class='danger'>[M] disarmed [src]!</span>")
-				else
-					playsound(loc, 'sound/weapons/slashmiss.ogg', 50, 1, -1)
-					visible_message("<span class='danger'>[M] has tried to disarm [src]!</span>")
 
 /mob/living/carbon/human/attack_animal(mob/living/simple_animal/M)
 	. = ..()


Fox-McCloud:
Just port TG's implementation of this entire proc, don't half-butt it.

Couls:
Sure thing, wanted to keep our code mostly the same though just change the logic

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -421,20 +421,16 @@ emp_act
 			updatehealth("alien attack")
 
 		if(M.a_intent == INTENT_DISARM)
-			if(prob(80))
+			var/obj/item/I = get_active_hand()
+			if(I && drop_item())
```
Fox-McCloud: unEquip(I) would be preferred

Fox-McCloud:
But...they're not the same. ANd I'm not even talking about the dismembering strike stuff.

Couls:
do you want me to port the visible_messages? or the to_chats? the paralyze proc is the same as our weaken proc here because we didn't convert stuns into status effects or do you want me to port the 10% chance to do no damage part? the disarm portion already adds logs and plays sounds and stuns for 10 seconds

farie82:
Issue I see with this is that nodrop will prevent hunters from dropping those targets. Bit odd logic to me as well tbh

Shockpoint:
That's it, I'm always holding onto a cup.

Shockpoint:
But real talk, it does seem a bit weird to have items (any item at that) prevent knockdowns. While I am def in favour of having _something_ countering xeno op tackles, this kinda just seems weird, 

Also what Farie said. Nodrop would become a meta against xenos and you'd never get to grab anyone at all. There must be another way to counter it other than by holding an item. Maybe restrict it only to shields?

Ty-Omaha:
So I could hold a pen and be immune to xeno tackles.

You should find a better way to balance it, this is just silly. Shields have actual block values, use that instead of classifying everything as a shield.

Couls:
xenos have a harm attack that can do anywhere between 15 and 30 damage and would stun you regardless of what you were holding, anti-drop wouldn't stop xenos from being able to take you down they'd just have to lethal you similarly to how security would in the exact same situation, a pen would buy you a few seconds but it really wouldn't make a difference because a hunter would still outrun you and a pen doesn't do any damage to xenos

Ty-Omaha:
Yes they can still harm slash. It doesn't change how having a pen act as a shield is stupid.

Spacemanspark:
The pen is mightier than the sword. 
Write up that dumb alien for harassment and get them fired, that'll show 'em

Tayyyyyyy:
In addition to breaking suspension of disbelief, it's also not very intuitive to new players at all on either side. Can you even easily tell that someone is holding a pen in their hand? That's something that could be missed in a combat situation. Mechanics should make sense, not only for RP reasons but also to make an already complex game less confusing to play. This change just screams gaminess and balance for the sake of "balance".

Sidenote: I don't see the need to nerf xenos. It's already hard enough to get started for players without experience, and if they do get out of control it's still fun for the apocalyptic atmosphere. Not everything needs to be "balanced".

Couls:
closing due to unpopularity

PR #12478, opened 2019-10-02
slightly slow down hunters

Couls:
## What Does This PR Do

This is the final proposed xeno nerf and likely to be the most controversial, so i'll spend a bit explaining my reasoning for this.
## Why It's Good For The Game

Hunters are too fast and they don't give the victim enough time to react to them attacking, the issue is that this combined with their thermal vision makes it really hard to counter a xeno outside of having a mech and the fact that they can get in and out of places very easily helps contribute to how powerful xenos are especially considering that dragging someone away is no slower than grabbing them in the first place. slightly slowing them down will give crew more time to react to their approach and also allow them to give chase if their xeno hunting buddy happens to be taken down early in the fight
## Changelog
:cl:
balance: xeno speed is more inline with paradise movespeed
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/alien/humanoid/caste/hunter.dm
+++ b/code/modules/mob/living/carbon/alien/humanoid/caste/hunter.dm
@@ -14,7 +14,7 @@
 	..()
 
 /mob/living/carbon/alien/humanoid/hunter/movement_delay()
-	. = -1		//hunters are sanic
+	. = -0.5		//hunters are sanic but our movespeed is different!
 	. += ..()	//but they still need to slow down on stun
 
 /mob/living/carbon/alien/humanoid/hunter/handle_hud_icons_health()


Fox-McCloud:
The movement delay on TG (if running) is 1.5; for here, it's 2.5

What they have on their config is NOT what htey use on the live server

Couls:
oh god in that case we should slow down hunters even MORE

Fox-McCloud:
not...really?

It takes ~5.3 second to cross from one side of the hall to the other (in front of the bridge) using TG's configured movement delay for human mobs. 

How long does it take hunters using their configuration? ~1.8 seconds

If we use OUR configuration, it takes ~8.8 seconds to cross the same distance as a human. 

It takes hunters ~5.2 seconds

If you're using TG as a metric, then hunters are *too slow*.

The proportional speed between hunters and humans on paradise is the same as it is on TG, as all other delays are held equal--which is to say, they're 1 level faster.

Couls:
right after testing for a bit I see what you mean now, the difference in delay is actually always 1 which makes a lot more sense now. I still feel like hunters are too fast here but for different reasons, I'll rewrite the "why it's good for the game" section later

MRgaty:
In my opinion, even playing as a hunter is annoying, as you go so fast is difficult to click on other spess men

Shockpoint:
Controversial? More like one of the best PRs in the world. Hunters are disgustingly fast.

Pixelzilla:
Are we just going to overlook the fact that hunters are supposed to be fast and they only have melee as a combat option? If you're stupidly fighting a xeno in melee range you absolutely deserve to get wrecked by one. They should be fast, their role is to hunt effectively and efficiently. Xenos already have numerous things that make it difficult to get them going (the event being so late in the round when R&D is generally done, cargo has a lot of points) and their only advantage before the crew noticing is the fact that they have the element of surprise and can do quick hit and runs to attempt to down someone before they can scream "xenos!". Making hunters slower is going to chip away at that advantage and try too hard to "even" the playing field against something that is quite honestly supposed to be an extremely difficult challenge.

In a 1v1 situation a xeno should almost always win against an unprepared crewmember. We dont need to nerf xenos into the ground. People just need to man up and deal with it. Sometimes you're victim 1 and its true there isnt much you can do when you're surprised but your death is part of providing the "action" for everyone else, thats just how the cards are delt and people can be a little irritated at that sometimes but thats how this game is supposed to be. Everyone cant live. Everything cant be 100% fair.

Couls:
yeah and in a 1vmany scenario a hunter can run in, stun and drag out before the rest can react and if xenos won everytime in a game that would be a very boring game in my opinion, I understand them being strong in the movies but we all just want to have fun here. if xenos are SUPPOSED to win every round then what's the point of even fighting back? might as well just cryo or walk directly into the nest then

Tayyyyyyy:
> if xenos are SUPPOSED to win every round then what's the point of even fighting back? might as well just cryo or walk directly into the nest then

This is not a pvp game. You can still have fun even while "losing". In fact, there is no win or lose condition in this game, so I don't even know. I just don't understand this mentality of everything needing to be a pitched 50/50 battle. That's what nukies is for.

> if xenos won everytime in a game that would be a very boring game in my opinion

Xenos do not "win" every time. They don't make it out of the early stealth stage much of the time

PR #12479, opened 2019-10-02
Fixes Soul Stone Shards, they can now poll deadchat if the original soul is gone.

JKnutson101:
## What Does This PR Do
This PR fixes a line that prevented Soul Stone Shards from polling deadchat to be filled, if the original soul is unreachable. This also prevents using soul shards on mobs that have never had a ckyey, to prevent soul sharding monkeys for free constructs. 

## Why It's Good For The Game
Cultists should not be denied their filled soul stone just because their target has left the game. Ghost chat has another way to rejoin the game. Prevents a cheese strategy before it could be used.

## Changelog
:cl:
tweak: Soul Shards fail when used on mobs that have never had a ckey.
fix: Soul Stone Shards properly poll dead chat to be filled.
/:cl:

EDIT: I forgot to say: Big thanks to SteelSlayer and Faire82 for help with the ckey problem!

### DIFF
--- a/code/game/gamemodes/wizard/soulstone.dm
+++ b/code/game/gamemodes/wizard/soulstone.dm
@@ -377,7 +377,7 @@
 			break
 
 	if(!chosen_ghost)	//Failing that, we grab a ghost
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Good heavens almighty soulstone code is bad.

PR #12480, opened 2019-10-02
Issue #12466: IPC EMP Identity Loss Fix

### LINKED ISSUE #12466: cybernetics repair doesn't fix unknown face for IPC
**Issue Description**:Cybernetics repair doesn't fix unknown face anymore

**What did you expect to happen**: Unknown face repaired with cybernetic repair like it used to be

**What happened instead**:cybernetic face surgery doesn't work. Welder can't be used at full health. Even when the head is injured the face is not repaired.

**Why is this bad/What are the consequences**: IPC face can't be repaired and always shows as Unknown (as Name of IPC).

1. Get a EMP implant and use the EMP implant.
2. Try to fix face from surgery cybernetics repair.

**When did the problem start happening**: I don't know. I know that cybernetic used to work few month ago. 

**Extra information**:

PidgeyThePirate:
## What Does This PR Do
Fixes #12466: NO_DNA humans (IPCs) can no longer husk.

I'm pretty sure IPCs aren't meant to husk under any circumstances, but after taking massive burn damage from an EMP they husk and remain so after reviving.

## Changelog
:cl:
fix: NO_DNA humans (IPCs) no longer husk under any circumstances.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -158,10 +158,14 @@
 	return
 
 /mob/living/carbon/human/proc/ChangeToHusk()
-	var/obj/item/organ/external/head/H = bodyparts_by_name["head"]
+
+	// If the target has no DNA to begin with, its DNA can't be damaged beyond repair.
+	if(NO_DNA in dna.species.species_traits)
+		return
 	if(HUSK in mutations)
 		return
 
+	var/obj/item/organ/external/head/H = bodyparts_by_name["head"]
 	if(istype(H))
 		H.disfigured = TRUE //makes them unknown without fucking up other stuff like admintools
 		if(H.f_style)


Fox-McCloud:
1) You can't fix husking on anyone, not even normal species
2) Is there a surgery to fix the disfigured status for heads that works for IPCs? If so, this is a non-issue and a case of "working as intended".

PidgeyThePirate:
I'm pretty sure husks are supposed to stay dead. If you can repair and revive an IPC, you should be able to restore its identity as you're reconstructing it, I do not agree that not being able to do so is in any way "working as intended". Unless you want a snowflake surgical procedure to fix husking in an IPC or a change to the way names are determined to check for the mob being an IPC, I think this is the best way to solve the issue. 

Pretty sure the only thing husking does to an IPC is blocking the real name from showing anyway.

Cazdon:
Yeah Husking is entirely weird on IPCs, like Pidgey said it only blocks your identity. There's no husk sprite, and you can even fix all the damage, wash off the scorch marks, and be as you were before any damage and you'll still be missing your identity. You can even replace your head entirely, and your identity is still unknown, as it's weirdly enough tied to your body. The only way to get rid of the lack of 'facial recognition' is to completely replace the middle body, which is extremely weird.

NoDNA preventing husking makes plenty of sense.

PR #12481, opened 2019-10-02
Welded vents burst open under strong overpressure

TDSSS:
## What Does This PR Do
This PR makes it so that vents can burst open if they are welded, turned on and have a strong internal pressure. If there are 50 atmospheres worth of pressure or more in the pipe net, then welded vents have a chance of bursting open while on (I assume the vents actual...vents are able to withstand the pressure, hence why it only happens if it's toggled on).

A bursting vent sends people on or adjacent to it flying like unwrenching pipes does and dumps some of its content on its tile.

It is just a 5% chance to not every vent on station bursts at the same time the moment pressure ticks over the threshold.

For reference, 50 atmospheres of internal pressure, assuming the vent has a 1 square meter area, would exert 5 million Newton worth of force on the vent ( if my math and wolfram alpha are correct)

## Why It's Good For The Game
This introduces a (small) downside to welding vents which right now is consequence free. You can still weld vents but you have to either:
1. Have atmos lower the distro pressure, giving them something to do. This also comes with downsides like allowing people with the ventcrawling jumpsuit to get places faster, slower pressurisation of rooms, easier sabotage by pumping poison into distro.
2. Have someone with atmos access turn off vents with an air alarm. This takes more time and effort than just having twenty assistants with welders weld every vent they see.

## Changelog
:cl:
add: Welded vents that are on now burst open at more than 50 atmospheres of internal pressure.
:cl:

### DIFF
--- a/code/ATMOSPHERICS/atmospherics.dm
+++ b/code/ATMOSPHERICS/atmospherics.dm
@@ -197,20 +197,20 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 
 //Called when an atmospherics object is unwrenched while having a large pressure difference
 //with it's locs air contents.
-/obj/machinery/atmospherics/proc/unsafe_pressure_release(var/mob/user,var/pressures)
+/obj/machinery/atmospherics/proc/unsafe_pressure_release(mob/user, pressures)
 	if(!user)
 		return
 
 	if(!pressures)
 		var/datum/gas_mixture/int_air = return_air()
 		var/datum/gas_mixture/env_air = loc.return_air()
-		pressures = int_air.return_pressure()-env_air.return_pressure()
+		pressures = int_air.return_pressure() - env_air.return_pressure()
 
-	var/fuck_you_dir = get_dir(src,user)
-	var/turf/general_direction = get_edge_target_turf(user,fuck_you_dir)
+	var/fuck_you_dir = get_dir(src, user)
+	var/turf/general_direction = get_edge_target_turf(user, fuck_you_dir)
 	user.visible_message("<span class='danger'>[user] is sent flying by pressure!</span>","<span class='userdanger'>The pressure sends you flying!</span>")
 	//Values based on 2*ONE_ATMOS (the unsafe pressure), resulting in 20 range and 4 speed
-	user.throw_at(general_direction,pressures/10,pressures/50)
+	user.throw_at(general_direction, pressures/10, pressures/50)
 
 /obj/machinery/atmospherics/deconstruct(disassembled = TRUE)
 	if(can_deconstruct)
--- a/code/ATMOSPHERICS/components/unary_devices/vent_pump.dm
+++ b/code/ATMOSPHERICS/components/unary_devices/vent_pump.dm
@@ -38,6 +38,7 @@
 	var/pressure_checks_default = PRESSURE_CHECKS
 
 	var/welded = 0 // Added for aliens -- TLE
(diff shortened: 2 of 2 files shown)


JKnutson101:
Giving atmos techs something else to do in their current state is a good thing in my eyes. (now if only there was a way to counter wrenching pipe cheese)

Couls:
> Giving atmos techs something else to do in their current state is a good thing in my eyes. (now if only there was a way to counter wrenching pipe cheese)

make scrubber pipes release gas contents like it does on TG, that way people will release hot plasma from the scrubbers or overpressure air from the vents and get screwed either way

Fox-McCloud:
While I dislike how people keep vents welded shut all the time....

The vents are *always* overpressured on Paradise because we have damn volumetric pumps by default.

>make pipes release gas contents like it does on TG, that way people will release hot plasma from the scrubbers or overpressure air from the vents and get screwed either way

We already have that.

TDSSS:
> The vents are _always_ overpressured on Paradise because we have damn volumetric pumps by default. Once they're overpressurized, there's not much you can do to fix, bar voiding the entire pipenet of its gas.

Well yes, that's the idea. Just pump some of distro into waste, it will get filtered back into the air tanks. It's not like you'll be dumping them into space or anything. And I'm sure some atmos techs will start turning the roundstart pump off or something at roundstart once there's a reason for controlling distro pressure.

Fox-McCloud:
quite frankly, with how quickly LINDA moves, those pipes should *probably* be changed back to regular pumps.

They're relics of ZAS where the pipes were prone to plugging if you so much as *looked* at them the wrong way.

Kyep:
I'm really unsure if this will make that big a difference.
Sooner or later, someone will want to weld a vent every shift. At which point they'll just ask the AI to turn down the distro pressure. Since the pumps are electronic, they can do so quickly, easily, and without fail.
Heck, atmos techs might turn down the pressure at shift start.
Once the pressure is turned down, you can then safely weld event vent on the station.

I don't think this is going to have that much effect.
Convince me otherwise?

TDSSS:
I don't expect it to single-handedly solve the excessive welding issue, if that's what you're asking. It is intended as one step in the right direction.

>At which point they'll just ask the AI to turn down the distro pressure. Since the pumps are electronic, they can do so quickly, easily, and without fail.

If distro pressure is already above the threshold it will take a bit to get it back down, and lowering it has other minor downsides. Also, many AIs, I've found, really do not know how atmos or atmos machines work. I bet at least some AIs wouldn't even know how to do that.

>Heck, atmos techs might turn down the pressure at shift start.
Once the pressure is turned down, you can then safely weld event vent on the station.

And that's okay, IMO. That is a valid trade-off that an atmos tech can make and a thing to make them marginally less useless. But if the pump is off, that means one major breach and suddenly rooms aren't pressurising anymore because distro is empty, requiring them to undo the work they did. Already, we've given the atmos tech two things to do. Especially if those 'atmosphere anomalies' or whatever get in that I've heard talk about, that mess with atmos and require active scrubbers+vents to fix.

SpessX:
> > make pipes release gas contents like it does on TG, that way people will release hot plasma from the scrubbers or overpressure air from the vents and get screwed either way
> 
> We already have that.

This only happens currently if you unwrench a pipe and the adjacent pipe section is not connected to anything and comes loose. Other than that, you only get launched around when unwrenching, and only if you aren't standing on top of the pipe. If you stand on the pipe and you aren't cutting loose an adjacent single pipe, unwrenching pressurized pipes is free.

PR #12486, opened 2019-10-03
Adds minor armor to combat gloves and jackboots

TDSSS:
## What Does This PR Do
Adds very minor armor to combat gloves and security jackboots.

## Why It's Good For The Game
Combat gloves and jackboots are both tough garments designed to provide some protection. If even a jumpsuit can provide minor melee armor, then tough boots should do too. It also is a reason to wear them over just any normal shoes from a vendor. In other words, it's for consistency, it is counter-intuitive for them to not provide at least some protection, even if it's not a game-changer or anything.

## Changelog
:cl:
add: minor armor to jackboots and combat gloves
/:cl:

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -29,6 +29,7 @@
 	name = "combat gloves"
 	icon_state = "combat"
 	item_state = "swat_gl"
+	armor = list(melee = 5, bullet = 5, laser = 5, energy = 5, bomb = 0, bio = 10, rad = 0)
 	siemens_coefficient = 0
 	permeability_coefficient = 0.05
 	strip_delay = 80
--- a/code/modules/clothing/shoes/miscellaneous.dm
+++ b/code/modules/clothing/shoes/miscellaneous.dm
@@ -102,6 +102,7 @@
 	can_cut_open = 1
 	icon_state = "jackboots"
 	item_state = "jackboots"
+	armor = list(melee = 10, bullet = 10, laser = 5, energy = 5, bomb = 0, bio = 10, rad = 0)
 	item_color = "hosred"
 	strip_delay = 50
 	put_on_delay = 50


AdamElTablawy:
bone talisman is 5 melee

sec jumpsuit is 10 melee

sec armor is 25 melee, riot armor is 50 melee

so if you pair sec armor with a bone talisman, jackboots, combat gloves, and their jumpsuit, you'll end up with an uncomfortably high armor value

one that'll only get worse if they get augumented

TDSSS:
Armor doesn't stack like that. Gloves only cover the hands, sec armor won't cover the hands, etc. You don't just add all the armor values together.

Fox-McCloud:
Not keen on giving gloves or boots armor; currently only the bone bracers have it (and that covers the arms), and the combat boots have it, which are pretty much limited to ops and adminbus only.

TDSSS:
I don't see why we should have some parts of the body that you just can't get armor for. Like, it doesn't make much sense from a fluff perspective and gameplay wise do we really want the optimal play to be targeting hands/feet even more? It feels very gamey to me.

TBH it feels like an oversight to me that was then post-hoc justified in some way, but maybe you have good reasons for it.

craftxbox:
s t e e l   t o e   b o o t s

Couls:
Why not just remove the jackboot sound? People put tape on their boots anyway and I think most players find the noise the jackboots make annoying more than anything

TDSSS:
The noises are kind of cool, though. As long as there was a proper upside to jackboots. ~~Could always remove jackboot taping then~~

AdamElTablawy:
@TDSSS _some_ armor doesn't cover the hands

other armor, like spacesuits, hos clothes, riot suit, etc, offer full body coverage, so it will stack

Fox-McCloud:
Yeah, and that's generally the problem, especially for stuff like the boots--again, with antag things, it's less concerning, but I don't think things generally available to the crew should exacerbate the problem of "armor stacking that hits `X >= 100` for melee damage"

TDSSS:
I mean in general, it probably makes sense to sanity check armor values and cap them somewhere high, IE don't allow 100% armor to most things? ~~Also can we just get rid of hands and feet as a separate zone in general?~~

Ty-Omaha:
I think it would be cool if it was just melee and bullet protection

Spacemanspark:
The sound is cool, don't remove it.

I have no real issues with this PR either way it goes, but it does make a bit of sense for those items, boots in particular, to give a very tiny bit of protection.

variableundefined:
I agree with Fox on the armor stacking issues.

Closing for the reasons he's stated above.

PR #12496, opened 2019-10-04
Shade related Soul Stone fixes and tweak.

### LINKED ISSUE #12495: Shades cannot return to soul stones.
**Issue Description**:
Shades cannot return to their soul stones.

Bapping a shade with its soul stone should put it back inside.

**What happened instead**:
Capture failed!: The soul stone has already been imprinted with "empty"'s mind!

Shades can't go back in their spooky rocks.

Summon shade. Bap shade with the same stone. Be sad.

Who knows.

**Extra information**:
var/imprinted is initialized to "empty" but never set. I honestly don't think we need imprinted at all.

JKnutson101:
## What Does This PR Do
Fixes #12495

Fixes unlisted issue where non-cultists and non-wizards can use cult-spawned and wizard-spawned Soul Stones.

Removes all Soul Stone references to the 'imprinted' variable that was defined, never set anywhere, but still checked for. Functionally, this means any empty Soul Stone can fit any shade in them now. 

## Why It's Good For The Game
Besides fixing two bugs, one restoring an intended feature, one fixing an unintended feature, the removal of 'imprinting' makes for a small quality of life change for Shade users, as they won't have to remember which Shade goes in which stone.

## Changelog
:cl:
tweak: Shades can be stored in any Soul Stone.
fix: Shades can once again be stored in Soul Stones.
fix: Non-cultists and Non-wizards can no longer use most Soul Stones.
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/soulstone.dm
+++ b/code/game/gamemodes/wizard/soulstone.dm
@@ -223,10 +223,6 @@
 		if("VICTIM")
 			var/mob/living/carbon/human/T = target
 			var/obj/item/soulstone/C = src
-			if(!C.can_use(U))
-				U.Paralyse(5)
-				to_chat(U, "<span class='userdanger'>Your body is wracked with debilitating pain!</span>")
-				return
 			if(T.stat == 0)
 				to_chat(U, "<span class='danger'>Capture failed!</span>: Kill or maim the victim first!")
 			else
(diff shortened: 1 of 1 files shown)


JKnutson101:
Was just double checking some code here, I've a few redundant lines, not vital but Do Not Merge yet please, I should be able to fix these in a few hours.

JKnutson101:
Alright, a duplicate line was removed. Tested, everything still worked.

PR #12506, opened 2019-10-06
Removes roundstart bulldogs for nuke ops, adds them to uplink

AzuleUtama:
## What Does This PR Do
This removes the free bulldog shotguns from inside the syndicate nuke ops shuttle and instead lets nukies buy them from the uplink for 6TC.

I don't really understand why they get these for free right now, they get more than enough TC for equipment both in general and if they declare war. The shotgun is a pretty good weapon in its own right and can be loaded with a variety of different ammo types on top of the starting 60 damage slugs.

Taking into account some of the other strong equipment nukies can get such as double-bladed eswords, elite suits, xrays and adrenals - getting a free 60 damage gun in addition doesn't seem necessary.

This also adjusts the price of the bulldog bundle to match up with the new price of the bulldog.

## Why It's Good For The Game
Cuts down on the amount of free gear nuke ops get in addition to all the TC they start with.

## Changelog
:cl:
balance: Bulldog shotguns removed from roundstart nuke ops locker, can now be bought from the uplink for 6TC.
/:cl:

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -472,6 +472,14 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	cost = 13
 	surplus = 50
 
+/datum/uplink_item/dangerous/bulldog
+	name = "Bulldog Shotgun"
+	reference = "BDS"
+	desc = "A compact automatic shotgun that fires various different kinds of 12 gauge shells with an 8-round magazine."
+	item = /obj/item/gun/projectile/automatic/shotgun/bulldog
+	cost = 6
+	gamemodes = list(/datum/game_mode/nuclear)
+
 /datum/uplink_item/dangerous/smg
 	name = "C-20r Submachine Gun"
 	reference = "SMG"
@@ -1637,7 +1645,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			Bulldog shotgun, two 12g buckshot drums, and a pair of Thermal imaging goggles."
 	reference = "BULB"
 	item = /obj/item/storage/backpack/duffel/syndie/bulldogbundle
-	cost = 9 // normally 12
+	cost = 12 // normally 16
 	gamemodes = list(/datum/game_mode/nuclear)
 
 /datum/uplink_item/bundles_TC/c20r
--- a/code/game/objects/structures/crates_lockers/closets/syndicate.dm
+++ b/code/game/objects/structures/crates_lockers/closets/syndicate.dm
@@ -32,19 +32,11 @@
 
 /obj/structure/closet/syndicate/nuclear/New()
 	..()
-	new /obj/item/ammo_box/magazine/m10mm(src)
-	new /obj/item/ammo_box/magazine/m10mm(src)
-	new /obj/item/ammo_box/magazine/m10mm(src)
-	new /obj/item/ammo_box/magazine/m10mm(src)
-	new /obj/item/ammo_box/magazine/m10mm(src)
+	for(var/i in 1 to 5)
+		new /obj/item/ammo_box/magazine/m10mm(src)
 	new /obj/item/storage/box/teargas(src)
 	new /obj/item/storage/box/flashbangs(src)
 	new /obj/item/storage/backpack/duffel/syndie/med(src)
-	new /obj/item/gun/projectile/automatic/shotgun/bulldog(src)
-	new /obj/item/gun/projectile/automatic/shotgun/bulldog(src)
-	new /obj/item/gun/projectile/automatic/shotgun/bulldog(src)
-	new /obj/item/gun/projectile/automatic/shotgun/bulldog(src)
-	new /obj/item/gun/projectile/automatic/shotgun/bulldog(src)
 	new /obj/item/pda/syndicate(src)
 
 /obj/structure/closet/syndicate/sst


JKnutson101:
I honestly like this, I always thought the 'spare weapon' was supposed to be the stechkin, but I often see nukies taking the bulldog as their spare because it's much stronger and still free. Bulldog should definitely a be a cheap primary, but still something you have to purchase.

My two cents anyway.

Mitchs98:
100% agree. Always found it weird they got a 'full auto' shotgun for free when they have to buy damn near everything else.

datlo:
This is mostly a nerf to stealth/blitz ops as frankly ballistics arent worth bothering with on war ops right now, melee/explosives is where its at.

Considering stealth/blitz is way way stronger than war right now, I think this is a fine change.

TDSSS:
This is a pretty sizeable nerf to nukies, do they actually need it? Most of the time I see nukies they fluke out etc.

AzuleUtama:
The fails i've noticed from nukies tend to be from their own mistakes or the crew outplaying them, not due to poor equipment. True stealth ops don't usually make use of bulldogs and war ops have more than enough TC for whatever they're going to need. Blitz ops might be nerfed somewhat since they get less TC for combat gear, but given I've witnessed them having a much higher success rate than other strategies I don't see this as a huge issue for them.

Mitchs98:
I don't see how this is a pretty sizeable nerf at all. They really shouldn't be given the shotgun in the first place. As Azule said, stealth ops and even really blitz ops don't need it in the first place. The action is usually over before any major actual combat with the crew.

TDSSS:
>I don't see how this is a pretty sizeable nerf at all

Taking free and useful stuff away from an antag nerfs them. It's a very good gun that just needs some ammo to be useful. It's great if you're going for a support/quirky build and for example take a medibeam gun but still want to defend yourself, or go for a borg/mech/etc.

Mitchs98:
If you're going for a support build you should in no way shape or form be entitled to a very strong gun as a side arm for free to begin with, nor should you if you're calling in strong support such as a mech. An antag shouldn't be given a very strong gun for free to begin with. I don't see it as a sizeable nerf. I see it as removing something they should of really never had for free.

There's limits to things antags should be given for free, the Bulldog is definitely one of them.

TDSSS:
>remove a powerful thing
>somehow not a nerf 🤔

I think we have very different definitions of the word nerf here. Just because you think it's justified doesn't mean it's not a nerf.


And it is a very strong gun in general, but compared to the other options available to nuke ops it's really subpar, tbh. And it's not really free since you still need to buy ammo for it.

florkeger:
I'd propose instead of removing them and making them a buyable item, to instead just modify the Syndicate Outpost a little bit to where those extra guns are locked in a room only the team leader can access. That way, he can decide whether or not they will be used, if at all.

Mitchs98:
Disagree on the team leader room. No one in their right mind would pass up free gear. It wouldn't change a single thing unless you had to pay for it.

And no, I don't consider changing something I consider should of ever been a thing a nerf.

florkeger:
> Disagree on the team leader room. No one in their right mind would pass up free gear. It wouldn't change a single thing unless you had to pay for it.
> 
> And no, I don't consider changing something I consider should of ever been a thing a nerf.

What? Why wouldn't anyone pass up free gear? It can restrict the decision on whether or not they get free guns to the team leader rather than just being flat out able to take them with you.

I don't see your point in that argument.

Mitchs98:
That was my point. No one would pass up the free gear. The leader would always take the free gear.

florkeger:
> That was my point. No one would pass up the free gear. The leader would always take the free gear.

Then it's just a bad leader, and no, still not really. If they advocate for a stealth mission, the leader who makes the call on the equipment will much rather NOT give his team weapons that could compromise his idea.

AzuleUtama:
If we kept them in a locker id'd to the leader then as Mitch said they'll either open it at roundstart anyway or get begged to open it from one of the other ops. Free gear is free gear, they're not gonna just leave it there - especially when it fits into a backpack allowing them to conceal it if they are planning on being stealthy.

While I understand you also need to invest in some ammo for the gun currently, you also need to do this for all other guns available to nuke ops - the difference is that you still need to buy those guns as well.

One of the best combos you can currently get with it is the grab bag of ammo (duffelbag with 6 slug, 2 buck and 1 fire magazine), which costs just 12TC for the bag since the gun is free anyway. Compare this with the cheapest nukie longarm you can buy, the C20, which costs 14TC for the gun on its own. That's 80 shots (counting the bulldog's initial magazine) against 20.... for 2TC less.

Ty-Omaha:
Really weird having a "bulldog bundle" and just a bulldog in the uplink.

Leanfrog:
Just to note, there's also the C20 bundle which costs like 4 more TC and comes with plenty of ammo for what you'll need as well as the C20

AzuleUtama:
> Objecting to this, on the grounds that nuke ops don't need nerfs.
> If you want to remove them and add an equivalent amount of TC to the nuke ops uplinks, sure. But as is, its just a nerf, and nuke ops win infrequently enough they don't need nerfs.

I wouldn't be against adding an equivalent amount of TC to the warops bonus, since I agree they don't need to be nerfed by this PR as they don't generally have a high level of success against a station armed with rifles and shotguns. Giving the TC to nuke ops in general I'm not as keen on however, since I've witnessed them having a much greater success rate over war ops and giving them more TC in exchange for losing a free gun they might not have taken anyway doesn't seem ideal to me.

variableundefined:
Closing due to objections

PR #12509, opened 2019-10-06
BSA Signal beacon

Ty-Omaha:
## What Does This PR Do
Creates a signal beacon to call in a BSA strike. 
You must stand still for 6 seconds to arm it, if you are interrupted it will cancel.
It plays an audible local warning as well as a visual red light, in addition to the station wide announcement.
It is a one time use item.

## Updates:

- Bigger explosion
- Blob can't eat it for balance, logic, and code reasons _(its a 1 time use item for a rare gamma team, also the NDV Brutus is already locked onto the target and fired and would have to snowflake around that)_
- Arm time reduced to 4 seconds from 6
- Code Improvements
- Sprites (by ionward)

## Why It's Good For The Game
It allows Gamma ERT commanders, DS, or CC officials to call in 1 BSA strike on a troublesome target.

Video (up to date): https://gofile.io/?c=9dMebt

## Changelog
🆑 Ty-Omaha and Ionward
add: BSA Signal Beacon. Gamma ERT Commanders now spawn with a 1 time use signal beacon to call in a BSA strike.
/ 🆑


code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,79 @@
...
+		to_chat(user, "<span class = 'warning'>Unauthorized user detected.</span>")
+	else
+		for(var/obj/item/grenade/bsa/B in world) //check if anyone else is calling a strike currently, we don't want to spam the announcements
+			if(B.active && B != src)
+				to_chat(user, "<span class = 'warning'>Another strike is already in progress!</span>")
+				return
+		active = TRUE
+
+		var/area/A = get_area(bombturf)
+		to_chat(user, "<span class = 'warning'>Calibrating GPS, remain still to continue with strike.</span>")
```
Ty-Omaha: Need to fix the grammar: "with the strike"

craftxbox: Personally `with strike` seems correct, while the the would also work.

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,83 @@
...
+			to_chat(user, "<span class = 'warning'>Calibration cancelled. Strike sequence aborted.</span>")
+			message_admins("[key_name_admin(user)] has cancelled a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>")
+		
+		
+/obj/item/grenade/bsa/prime()
+	update_mob()
+	var/ex_power = 3
+	var/turf/bombturf = get_turf(src)
+	var/area/A = get_area(bombturf)
+	explosion(get_turf(src), ex_power * 2, ex_power * 5, ex_power * 6)
```
Kyep: Instead of defining ex_power in prime(), it should be defined at the object level, so it can be varedited.
Or, better yet, scrap the idea of using a single ex_power var, and have devastation/high/light/flame/etc power as independent vars defined at the object level, perhaps as a var/list/bsa_explosive_effects(...). Put the announcement text in a var, too.
And have the actual detonation handled by a proc/call_in_bsa defined at the object level, which can be overriden.

That would allow several things:
- easy varediting of the bsa grenade to alter its yield
- easier creation of variable power subtypes, such as a low power variant for red team commanders, or a high power variant for specops/DS
- creation of special effect subtypes, such as a type that functions as a huge-radius EMP nade ("Orbital Ion Cannon shot"), purely fire nade ("thermal beam strike"), or even other effects ("purifier ray" v cult/vampires/etc). Heck, even a syndicate version would be possible with this system (probably reserved for nukeops).

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,83 @@
...
+	if(active)
+		return
+
+	if(is_admin_level(bombturf.z) && !centcom_cancast)
+		to_chat(user, "<span class = 'warning'>The strike craft isn't authorized to fire near Central Command!</span>")
+		return
+
+	if((!ismindshielded(user)) || ((user.mind.special_role != SPECIAL_ROLE_ERT) && (user.mind.special_role != SPECIAL_ROLE_DEATHSQUAD) && (!allowed(user)))) //if is not mindshielded, or is not an ERT/DS AND does not have high level CC access. naval officers have access_cent_commander, but not special role so we don't want to lock them out.
+		to_chat(user, "<span class = 'warning'>Unauthorized user detected.</span>")
+	else
```
Kyep: Instead of putting the entire rest of the proc under 'else', just have it return early if it is unauthorized. That has the same effect, without having to indent the entire rest of the proc an extra level.

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,83 @@
...
+				C.throw_mode_on()
+			addtimer(CALLBACK(src, .proc/prime), strike_init_time * 10)
+		else
+			active = FALSE
+			to_chat(user, "<span class = 'warning'>Calibration cancelled. Strike sequence aborted.</span>")
+			message_admins("[key_name_admin(user)] has cancelled a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>")
+		
+		
+/obj/item/grenade/bsa/prime()
+	update_mob()
```
Kyep: Could you explain what this is for?

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,83 @@
...
+	var/mob/user = usr
+	var/turf/bombturf = get_turf(src)
+	if(active)
+		return
+
+	if(is_admin_level(bombturf.z) && !centcom_cancast)
+		to_chat(user, "<span class = 'warning'>The strike craft isn't authorized to fire near Central Command!</span>")
+		return
+
+	if((!ismindshielded(user)) || ((user.mind.special_role != SPECIAL_ROLE_ERT) && (user.mind.special_role != SPECIAL_ROLE_DEATHSQUAD) && (!allowed(user)))) //if is not mindshielded, or is not an ERT/DS AND does not have high level CC access. naval officers have access_cent_commander, but not special role so we don't want to lock them out.
```
Kyep: This check should probably be a sub-proc too, so that subtypes of this item can easily override that check.
I could see us adding a version that's authorized for DS/Specops but not ERT, or for syndicate but not anyone else, etc.

code/game/jobs/job/central.dm
```diff
@@ -34,6 +34,9 @@
 		/obj/item/implant/dust
 	)
 	backpack = /obj/item/storage/backpack/satchel
+	backpack_contents = list(
+		/obj/item/grenade/bsa = 1
```
Kyep: Why do I feel like sooner or later, this is going to get thrown at an incompetent Captain.

code/game/jobs/job/central.dm
```diff
@@ -83,7 +86,8 @@
 	backpack = /obj/item/storage/backpack/satchel
 	backpack_contents = list(
 		/obj/item/clothing/shoes/magboots/advance = 1,
-		/obj/item/storage/box/zipties = 1
+		/obj/item/storage/box/zipties = 1,
+		/obj/item/grenade/bsa = 2
```
Kyep: This might be a prime example of where a subtype might be useful. Specifically a variant that has a configurable blast yield, with configuration working using the same series of prompts that the event 'drop bomb' verb uses.

code/game/objects/items/weapons/grenades/custom_grenades.dm
Ty-Omaha: I have no idea. All the other grenades have them

Kyep: You should probably understand what every line of the code you are using actually does, rather than just copy-pasting it.

code/game/objects/items/weapons/grenades/custom_grenades.dm
Ty-Omaha: then I need to make a proc on the mind/mob level to check their roles

Fox-McCloud: Hardcoded checks like this are *strongly* frowned upon in nearly all circumstances, with few exceptions, especially when it comes to using devices. Yes, they do exist else where, but they're typically *edge case*

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,83 @@
...
+		to_chat(user, "<span class = 'warning'>Unauthorized user detected.</span>")
+	else
+		for(var/obj/item/grenade/bsa/B in world) //check if anyone else is calling a strike currently, we don't want to spam the announcements
```
Fox-McCloud: `in world` checks.

🦊🗡

Fox-McCloud:
This is a lot of special code and arbitrary checks to effectively make a grenade that just causes an explosion.

Not a fan to be honest.

code/game/objects/items/weapons/grenades/custom_grenades.dm
Kyep: This check is probably not a good idea at all.
Right now, only ERT/admins can get this. The chances of multiple ones being thrown at once is small, and if it did happen, its probably temporary, and has a good reason.
Later on, with more subtypes, it would not make sense, for example, for a centcom BSA nade to block a syndie BSA nade, or vice versa.
I'd just scrap this check entirely.
If certain roles try to spam this... instead of giving them many smaller ones, just give them one bigger one.

code/game/objects/items/weapons/grenades/custom_grenades.dm
Ty-Omaha: Did some, but most of it can be saved for another PR so that it's not a massive PR with the feature chill

Ty-Omaha:
Round atmosphere

code/game/objects/items/weapons/grenades/custom_grenades.dm
Ty-Omaha: oh i thought i fixed that

Ty-Omaha: fixed

Ty-Omaha:
conflicts

taukausanake:
Instead of a grenade, why not make it a gun/tracer that you point at something? That's what most airstrike mechanics use. Would that be possible to implement?

Ty-Omaha:
>Instead of a grenade, why not make it a gun/tracer that you point at something? That's what most airstrike mechanics use. Would that be possible to implement?

I originally thought about that but then decided against it for balance reasons

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	det_time= 8 //in seconds, we multiply by 10 below so we can use this var in the messages we send. so a value of 8 here would be a value of 80 for the timer.
+	var/arm_time = 40 //in deciseconds, so 40 would equal 4 seconds.
+	var/caller
+	var/centcom_cancast = FALSE //incase admin wants to varedit to allow on Z 2 for whatever reason.
+	var/ex_power = 3
+	var/authorized_special_roles = list(SPECIAL_ROLE_ERT, SPECIAL_ROLE_DEATHSQUAD)
+	var/ID_access_override = TRUE
+	var/require_mindshield = TRUE
+
+/obj/item/grenade/bsa/proc/authenticate(var/mob/user, var/mindshield_check, var/role_check, var/access_check)
```
farie82: vars implied

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+		return FALSE
+	//access check
+	if(access_check && !allowed(user))
+		return FALSE
+	//special role check
+	if(role_check)
+		var/role = FALSE
+		for(var/R in authorized_special_roles)
+			if(user.mind.special_role == R)
+				role = TRUE
```
farie82: Can just return TRUE upon finding it and returning false otherwise

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+				break
+		if(!role)
+			return FALSE
+	return TRUE
+
+/obj/item/grenade/bsa/attack_self()
+	var/mob/user = usr
```
farie82: attack_self already takes user as param. Just add mob/user to the params

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	var/turf/bombturf = get_turf(src)
```
farie82: Small thing: Can be put after the active check since it's not required before that

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	if(active)
+		return
+	if(is_admin_level(bombturf.z) && !centcom_cancast)
```
farie82: Minor thing: Can switch these around. A bool check is cheaper to do than the proc check

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	if(!authenticate(user, require_mindshield, authorized_special_roles, FALSE) && (ID_access_override && !authenticate(user, FALSE, FALSE, TRUE))) //if is not mindshielded, not an ERT/DS AND does not have high level CC access. naval officers have access_cent_commander, but not special role so we don't want to lock them out.
+		to_chat(user, "<span class = 'warning'>Unauthorized user detected.</span>")
+		return
+	active = TRUE
+	icon_state = initial(icon_state) + "_active"
+	item_state = initial(item_state) + "_active"
+	var/area/A = get_area(bombturf)
+	to_chat(user, "<span class = 'warning'>Calibrating GPS, remain still to continue with the strike.</span>")
+	//tell admins and ghosts we are initiating a strike
+	message_admins("[key_name_admin(user)] is initiating a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>")
```
farie82: Design:
The message_admins might be a bit much. Maybe only on it fully activating
Technically:
`ADMIN_VERBOSEJMP` or `ADMIN_COORDJMP` or such can be used for the jump.

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+		return
+	active = TRUE
+	icon_state = initial(icon_state) + "_active"
+	item_state = initial(item_state) + "_active"
+	var/area/A = get_area(bombturf)
+	to_chat(user, "<span class = 'warning'>Calibrating GPS, remain still to continue with the strike.</span>")
+	//tell admins and ghosts we are initiating a strike
+	message_admins("[key_name_admin(user)] is initiating a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>") 
+	notify_ghosts("[user] is initiating a Bluespace Artillery Strike in [get_area(src)]!", source = src)
+	log_game("[key_name_admin(user)] is initiating a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>")
```
farie82: Writing html code in a log is probably not quite what is wanted here.

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	
+	playsound(src, 'sound/effects/bsainit.ogg', 50, 0)
+	
+	//make sure they stay still, if they move it's cancelled
+	if(!do_after(user, arm_time, target = user))
+		active = FALSE
+		to_chat(user, "<span class = 'warning'>Calibration cancelled. Strike sequence aborted.</span>")
+		message_admins("[key_name_admin(user)] has cancelled a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>")
```
farie82: Same as above

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	light_color = "blue"
+	light_range = 10
+	update_light()
+	priority_announcement.Announce("This is NDV Brutus, we've received your strike signal loud and clear, [user]! Bluespace Artillery Strike incoming near [A.name]!", "Bluespace Artillery Strike", 'sound/effects/engine_alert2.ogg')
+	to_chat(user, "<span class = 'warning'>Strike sequence initiated! Deploy beacon to target! [det_time] seconds!</span>")
+	caller = user
+	if(iscarbon(user))
+		var/mob/living/carbon/C = user
+		C.throw_mode_on()
+	addtimer(CALLBACK(src, .proc/prime), det_time * 10)
```
farie82: Minor thing. Can use `10 SECONDS` here. The * 10 is fine as well though

code/game/objects/items/weapons/grenades/custom_grenades.dm
```diff
@@ -303,3 +303,97 @@
...
+	
+/obj/item/grenade/bsa/prime()
+	var/turf/bombturf = get_turf(src)
+	var/area/A = get_area(bombturf)
+	explosion(get_turf(src), ex_power * 2, ex_power * 5, ex_power * 6)
+	message_admins("[key_name_admin(caller)] has sucessfully called a Bluespace Artillery Strike at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>")
```
farie82: Same as the above comments about the logging/messaging

Fox-McCloud:
Closing. This hasn't been touched in forever, even after a review was left---and even after github added a label to it.

PR #12512, opened 2019-10-06
Removes Console Screens

Fox-McCloud:
Removes console screens and replaces them with just *plane* old glass.

From TG. 

There's not much point to console screens; you can *only* get them from an autolathe, and *actual* consoles don't even use console screens. They're really strange, from a stock part perspective. They're annoying, making construction more difficult, and they're not a particularly great construction item.

This pretty much means things that have "screens" are consistent; any place you'd previously use a "console screen" for, you'll now use a sheet of glass (same ratio).

:cl: Fox McCloud
tweak: Removes consoles screens; use plane old glass in place of them. Holy sheet, they were a pane.
/:cl:


Couls:
typo in CL, should be plain or pane(if the pun is intended) instead of plane

variableundefined:
yes

JKnutson101:
> 
> 
> I don't have a pun so I approve

Yeah, you're just here to **_screen_** the code.

KasparoVy:
Console screens were nothing but a pane so instead we switch to glass to improve the game

PidgeyThePirate:
I should note you can put console screens in RPDs but not glass sheets, as far as I know. That may prove to be a minor annoyance as you also have to carry cable separately. I'd like to be able to add small amounts of metal and glass into an rpd to provide cable and glass, if we're doing stock part changes. Maybe I could look into that?

Fox-McCloud:
Storage doesn't work like that, so the alternative option is holding almost infinite amounts of glass in one container. 

Carrying around glass really *isn't* that hard.

PR #12515, opened 2019-10-07
Makes Pants All Powerful

Fox-McCloud:
Simple change; if you equip pants and are wearing an undershirt, then they cover your arms and chest.

I realize that technically some undershirts don't cover the arms (like tank tops), but making a list of which shirts do which just for that would be absurd.

Doesn't really change much, except if you fireproof your pants while having an undershirt, it'll provide more fire protection.

:cl: Fox McCloud
tweak: wearing pants while wearing an undershirt means your arms and chest are covered
/:cl:

### DIFF
--- a/code/modules/clothing/under/pants.dm
+++ b/code/modules/clothing/under/pants.dm
@@ -1,7 +1,17 @@
 /obj/item/clothing/under/pants
 	gender = PLURAL
 	body_parts_covered = LOWER_TORSO|LEGS
-	displays_id = 0
+	displays_id = FALSE
+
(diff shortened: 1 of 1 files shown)


KasparoVy:
Mr. Fancy himself would be proud.
fancy pants logo

taukausanake:
I'd personally just get rid of underclothes and make a modular clothing system. One with tops and bottoms, shirts and pants. Something like the accessory system but they still take up a "slot" so to say. That would be a major refactor of clothing and inventory slots though

Fox-McCloud:
Have fun coding that one. No thanks.

PR #12519, opened 2019-10-07
Removes Dirty Floors

Fox-McCloud:
Dirt doesn't add anything particularly fun to the game, and ever since we got blood trails and blood drips and what have you, the station is, more often than not, *covered* in blood.

Dirt is impossible to keep up with, as a janitor, particularly on high pop...not only that, but it's lagging the server and gumming up the garbage collector; taking thousands of deciseconds of deletion time.

With how much vomit, blood, clothing shreds, and other messes that the crew can make these days, dirt is just adding insult to injury and doesn't really let the janitor fixate on tasks he can actually handle.

Plus, did i mention it's literally lagging the server right now?

img

It only failed 50 times and it's already used up 3 seconds of calculation time. That's 60 deciseconds per delete---yikes.

:cl: Fox McCloud
del: Floors no longer get dirty by walking on them
/:cl:

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -194,8 +194,7 @@
 
 	for(var/obj/effect/decal/cleanable/M in view(2, E.loc))//germs from messes
 		if(AStar(E.loc, M.loc, /turf/proc/Distance, 2, simulated_only = 0))
-			if(!istype(M,/obj/effect/decal/cleanable/dirt))//dirt is too common
-				germs++
+			germs++
 
 	if(tool.blood_DNA && tool.blood_DNA.len) //germs from blood-stained tools
 		germs += 30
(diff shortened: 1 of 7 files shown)


code/modules/surgery/surgery.dm
```diff
@@ -194,8 +194,7 @@
 
 	for(var/obj/effect/decal/cleanable/M in view(2, E.loc))//germs from messes
 		if(AStar(E.loc, M.loc, /turf/proc/Distance, 2, simulated_only = 0))
-			if(!istype(M,/obj/effect/decal/cleanable/dirt))//dirt is too common
```
Fox-McCloud: Since dirt isn't as common now, it should now count.

AzuleUtama:
I don't think anyone really bothered with dirt anyway, so if it improves server performance by removing it I see it as a win-win.

PR #12523, opened 2019-10-08
Refactors Zone Selection

Fox-McCloud:
From TG.

This is long overdue. 

Instead of the zone selector being held on the mob, it's now held on the HUD datum...as such, it allows us to get rid of `user.zone_zel.selecting` and now just have it be `user.zone_selected`, which is a lot cleaner; it also means we don't have to do ugly hacks like creating a screen objects for simple mobs that happen to utilize actual objects in their behavior (like Blood drunk, mining bots, and possessed objects).

Lastly, this means no more items cross referencing each other, which typically means a better time garbage collecting.

:cl: Fox McCloud
tweak: Zone selection tweaked; shouldn't notice anything different
/:cl:

### DIFF
--- a/code/_onclick/hud/hud.dm
+++ b/code/_onclick/hud/hud.dm
@@ -24,6 +24,7 @@
 	var/obj/screen/alien_plasma_display
 	var/obj/screen/nightvisionicon
 	var/obj/screen/action_intent
+	var/obj/screen/zone_select
 	var/obj/screen/move_intent
 	var/obj/screen/module_store_icon
 
@@ -67,6 +68,7 @@
 
 	inv_slots.Cut()
 	action_intent = null
+	zone_select = null
 	move_intent = null
 
 	QDEL_LIST(toggleable_inventory)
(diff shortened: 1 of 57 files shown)


code/_onclick/hud/hud.dm
```diff
@@ -80,7 +82,6 @@
 	mymob.healths = null
 	mymob.healthdoll = null
 	mymob.pullin = null
```
Fox-McCloud: these four above also need the same treatment, but I'm not going to bother with that in this PR.

code/modules/mob/living/living.dm
```diff
@@ -376,15 +376,6 @@
 /mob/living/is_drawable(mob/user, allowmobs = TRUE)
 	return (allowmobs && reagents && can_inject(user))
 
-/mob/living/proc/get_organ_target()
-	var/mob/shooter = src
-	var/t = shooter:zone_sel.selecting
```
Fox-McCloud: What the actual hell.

Issue #12530, opened 2019-10-08
Reflective Blobs don't work


Ty-Omaha:
**Issue Description:**
Reflective blobs are supposed to reflect energy projectiles (at a high percentage), now they take damage.

Universal object damage broke it #12245

https://i.imgur.com/DQePSUG.png

Ty-Omaha:
lines 57-63

https://i.imgur.com/ABY8rKS.png

Issue #12528, opened 2019-10-08
Can open lockboxes using facid


farie82:
**Issue Description**:
As the title says. Lockboxes are easily opened by facid

**What did you expect to happen**: 
Science not to have 10 AEG's in their bluespace backpack

**What happened instead**:
Every scientist and his RD now have guns and are even more armed than cargonia in an identity crisis

**Why is this bad/What are the consequences**:
Don't have to explain this I hope

**Steps to reproduce the problem**:
1. Get lockbox
2. Throw facid over it. Or use grenade

**When did the problem start happening**:
Since the universal damage PR

Spacemanspark:
Oh God, the floodgates have been opened once again

PR #12532, opened 2019-10-08
Fixes Universal Damage Issues. Implements Proper Ricochets

Fox-McCloud:
So, when ricochets were ported, it wasn't done very well at all. Blobs, themselves, checked for ricochets, but there was a `ricochet_chance` var that could have handled this just fine...but wasn't used. *What*.

Either event, this fixes ricochetting with blobs and enables it for walls as well---currently only Titatnium walls will ricochet.

On that note, everything can ricochet now; bullets only ricochet 30% of the time; beams do so 80%.

Emitters have also been converted over to NOT use legacy projectiles...I don't see why they should use them; they still work just fine and are reflected properly now (still work ok with reflector boxes too).

Removes the unused `unacidable` var that was still on mobs and turfs

Also melting or calling `deconstruct` on research lockboxes just deletes them; it no longer dumps the contents.

This is temporary; eventually we'll probably move away from *most* lockboxes, but as a stopgap, this is necessary.

fixes: https://github.com/ParadiseSS13/Paradise/issues/12530
fixes: https://github.com/ParadiseSS13/Paradise/issues/12528

On a personal note, you science mains who intentionally didn't report the issue with lockboxes and immediately began abusing it, shame on you; only reporting bugs when they disadvantage you, but conveniently "forgetting" to report positive bugs on the Github is a pet peeve of mine. Good on @farie82 for reporting it.

:cl: Fox McCloud
fix: melting research lockboxes no longer dumps the contents
fix: Fixes reflective blobs using horrifying code that didn't reflect things *properly*
/:cl:


code/game/gamemodes/blob/blobs/shield.dm
```diff
@@ -38,26 +38,16 @@
...
-	if(P.is_reflectable && prob(reflect_chance) && !P.legacy)
-		var/P_turf = get_turf(P)
-		var/face_direction = get_dir(src, P_turf)
-		var/face_angle = dir2angle(face_direction)
-		var/incidence_s = GET_ANGLE_OF_INCIDENCE(face_angle, (P.Angle + 180))
-		if(abs(incidence_s) > 90 && abs(incidence_s) < 270)
-			return FALSE
-		var/new_angle_s = SIMPLIFY_DEGREES(face_angle + incidence_s)
-		P.setAngle(new_angle_s)
-		P.firer = src //so people who fired the lasers are not immune to them when it reflects
```
Fox-McCloud: Really. Let's set the firer, which is always a mob to be an object, what could go wrong? Never runtimes like `undefined var: /obj/structure/blob/shield/reflective/var/ckey` No, that would never happen, oh no no.

code/game/gamemodes/blob/blobs/shield.dm
```diff
@@ -38,26 +38,16 @@
...
-		visible_message("<span class='warning'>[P] reflects off [src]!</span>")
-		return -1// complete projectile permutation
-	else if(P.is_reflectable && P.legacy) //to stop legacy projectile exploits
-		visible_message("<span class='warning'>[P] disperses into energy from [src]!</span>")
-		qdel(P)
-	else
-		playsound(src, P.hitsound, 50, 1)
-		visible_message("<span class='danger'>[src] is hit by \a [P]!</span>")
-		take_damage(P.damage, P.damage_type)
```
Fox-McCloud: *but why though* This is obviously how ricochets are NOT supposed to work.

code/modules/projectiles/projectile.dm
```diff
@@ -185,13 +185,13 @@
 
 	if(check_ricochet(A) && check_ricochet_flag(A) && ricochets < ricochets_max)
 		ricochets++
-	if(A.handle_ricochet(src))
```
Fox-McCloud: *sigh*

code/modules/projectiles/projectile.dm
```diff
@@ -185,13 +185,13 @@
...
-		on_ricochet(A)
-		ignore_source_check = TRUE
-		range = initial(range)
-		return TRUE
-	if(firer)
-		if(A == firer || (A == firer.loc && istype(A, /obj/mecha))) //cannot shoot yourself or your mech
+		if(A.handle_ricochet(src))
+			on_ricochet(A)
+			ignore_source_check = TRUE
```
Fox-McCloud: This var was added, then promptly not utilized or plugged in properly ANYWHERE. Seriously. IF YOU PORT SOMETHING MAKE SURE YOU UNDERSTAND THE SYSTEM. THIS IS ALSO WHY YOU DON'T SET `firer` TO A NON MOB TO AVOID THE BUG OF HITTING THE MOB WHO ACTUALLY FIRED. FOR PETES SAKE.

Issue #12535, opened 2019-10-09
The Syndicate Depot doesn't get fully destroyed.


JKnutson101:
**Issue Description**:
When Code Delta is declared at the Syndicate Depot, it doesn't fully wipe out the area like it used to, and leaves behind a lot of loot and debris that shouldn't have remained.

**What did you expect to happen**: 
For the suspicious asteroid to be almost completely destroyed

**What happened instead**:
A lot of chunks of depot remained, and several bits of unclaimed loot as well.

**Why is this bad/What are the consequences**:
Free loot for 'failing' the depot is bad.

**Steps to reproduce the problem**:
Trigger enough alarms to set the depot to delta. Scavenge free loot post explosion.

**When did the problem start happening**:
Probably related to the changes with Universal Object Damage.

Kyep:
Having just tested this, the main things left behind appear to be:

- airlocks (which break, and drop circuits)
- metal (stacks of it everywhere)
- atmos pipes and vents (which appear displaced, not destroyed)
- stuff inside boxes inside lockers (e.g pills from pill bottles)
- gibs and blood
- broken lights

Most of this stuff is semi-useless debris.
Of the actual loot, it appears to be mostly stuff inside boxes inside lockers.

Really though I should probably just recode the explosion to:
- change all turfs in the area to /turf/space
- qdel everything else

Explosions don't actually destroy stuff reliably - even dropping a bomb directly on top of a box only destroys the box - everything inside the box is unaffected.

If all containers work like this, it means you can protect an object from a reactor meltdown explosion by wrapping it in paper.

Couls:
I thought TG explicitly made a proc to manually delete things in containers when they explode, its called something like contents_explosion, perhaps we can port that?

KasparoVy:
I remember when you could survive massive fucking bombs by stowing yourself in a locker, bodybag etc. back in the day
Did that come back with a vengeance?

Fox-McCloud:
@KasparoVy I mean, you can try: https://github.com/ParadiseSS13/Paradise/blob/5936bb98fea2bc46769df7ee6bdf57a5f3f3db37/code/game/objects/structures/crates_lockers/closets.dm#L402

Kyep:
"contents_explosion" doesn't exist in our codebase.
I'm not sure trying to port that whole system from TG is a good idea.
Its certainly way beyond the scope of just fixing this bug.

PR #12539, opened 2019-10-09
Fixes Explosions

Fox-McCloud:
I forgot  to include the `ex_act` for storage which calls `ex_act` on the contents of said storage, which boxes dumping their contents and leaving it mostly unharmed.

Also fixes sleepers making you immune to explosions.

Fixes: https://github.com/ParadiseSS13/Paradise/issues/12535

The yellow documents will ALWAYS survive, because they have the `INDESTRUCTIBLE` flag, as the vast bulk of objective items do.

:cl: Fox McCloud
fix: Fixes explosions not properly interacting with storage items
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -346,7 +346,19 @@
 /obj/machinery/sleeper/ex_act(severity)
 	if(filtering)
 		toggle_filter()
+	if(occupant)
+		occupant.ex_act(severity)
 	..()
+
+/obj/machinery/sleeper/handle_atom_del(atom/A)
+	..()
+	if(A == occupant)
+		occupant = null
+		updateUsrDialog()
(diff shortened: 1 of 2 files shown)


TDSSS:
Since nobody ever has to specifically get the yellow documents maybe they should lose that flag? They certainly were destroyed in depot explosions before and nobody minded.

Kyep:
Yes, the yellow docs should lose that flag. They're not intended to be recoverable if the depot blows up.

Fox-McCloud:
No one minded deleting the HoS's gun, Cap's gun, and all that jazz either, but they can't be deleted so easily now---it was largely that way because there wasn't a good system in place to handle it; now there is.

I'm not going to bother changing that in this PR--feel free, but the argument you're advancing isn't a good one in relation to "people not minding before".

PR #12544, opened 2019-10-09
Nerfs Glass Armor

Fox-McCloud:
Dunno how I feel about the buff if it's good or bad--so I'm going to revert it for now

:cl: Fox McCloud
tweak: windows are easier to break
/:cl:


taukausanake:
And here I thought there was actual _glass armor_ and not the glass _armor_ var

Carthusia:
does this apply to reinforced glass and its other variants or just normal plain glass?

PR #12548, opened 2019-10-09
Traits framework

Fox-McCloud:
This ports over the basic framework for adding/removing traits, from TG.

This doesn't make an attempt to replace all existing mutations/behaviors/species traits with traits, as I reallyyy don't want to do another huge refactor, but the basics of the system will be in place if anyone is crazy enough to do it.

This does, however, implement ONE trait; pacifism.

Pacifism is a very debilitating trait that prevents you from engaging in direct harm against another being. There's definitely convoluted ways to circumvent this, like pushing someone out in space, pressing a button that will launch them into an emagged gibber, and what have you, but what it does prevent is a lot of *direct* harm--stuff like hitting with a weapon, grabbing aggressively, shooting someone with lethal rounds, or throwing lethal things at people. 

This also implements Rod of Asclepius, a Lavaland loot item; when you pick it up, it's an inactive artifact, but if you pledge your very mind and will to "First, Do No Harm", then the rod will permanently attach to your arm while also providing healing in an AoE (and a little bonus to the pledger) around the pledger. The downside is, well...you get pacifism permanently added to you and the loss of that arm.

Even if the item in your hand is somehow replaced, the rod will grow back; if your arm or hand gets hacked off, it'll grow back. 

That said, if you happen to die while pledged, then the rod will absorb the soul of the person in question....so while the benefits are great, the downsides are extreme.

Also implement's TG's snakes; they have a better sprite than our current ones (which are now removed). They're not immediately hostile; they're a retaliate mob--just don't punch them and you'll be fine.

:cl: Fox McCloud
add: Adds in the Rod of Asclepius, a healing item that carries an immense burden
/:cl:


code/datums/status_effects/buffs.dm
```diff
@@ -99,6 +99,108 @@
...
+				else
+					if(itemUser.put_in_r_hand(newRod))
+						to_chat(itemUser, "<span class='notice'>The Rod of Asclepius suddenly grows back out of your arm!</span>")
+					else
+						if(!itemUser.has_organ("r_arm"))
+							new /obj/item/organ/external/arm/right(itemUser)
+						new /obj/item/organ/external/hand/right(itemUser)
+						itemUser.update_body()
+						itemUser.put_in_r_hand(newRod)
+						to_chat(itemUser, "<span class='notice'>Your arm suddenly grows back with the Rod of Asclepius still attached!</span>")
```
Fox-McCloud: This code...I'm not fond of. Sadly, we don't have a way to get an item index to a particular hand and what bodypart that hand is index to, as well, so this is about the best we can do. If anyone has any code recommendations, let me know.

code/modules/projectiles/gun.dm
```diff
@@ -207,14 +207,17 @@
...
+				to_chat(user, "<span class='warning'>[src] is lethally chambered! You don't want to risk harming anyone...</span>")
+				return
 		firing_burst = 1
 		for(var/i = 1 to burst_size)
 			if(!user)
 				break
 			if(!issilicon(user))
 				if( i>1 && !(src in get_both_hands(user))) //for burst firing
 					break
-			if(chambered)
```
Fox-McCloud: oops.

Shockpoint:
I like that Rod. Whew.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -582,6 +582,8 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 		qdel(G)	//We delete the grab.
 		if(throwable_mob)
 			thrown_thing = throwable_mob
+			if(HAS_TRAIT(src, TRAIT_PACIFISM))
+				to_chat(src, "<span class='notice'>You gently let go of [throwable_mob].</span>")
 			var/turf/start_T = get_turf(loc) //Get the start and target tile for the descriptors
```
Couls: shouldn't we return here?

reschreiter1311:
You heard it here first boys, shock likes fox`s rod!

KasparoVy:
Taking pacifism straight up. Lovely
There's a lot of neat traits that can be referred to in other implementations so the opportunity is massive!
I'll look back to Terilia's PR to jog my memory on just how far we were on some of that stuff.

Fox-McCloud:
@KasparoVy 

I haven't implemented *quirks* yet, which is what oyu may be thinking of. On TG, one of the quirks gives you pacifism, but I haven't implemented that here---so you won't be able to get this except via the rod, at the moment.

That said, if/when we have converted all we can over to traits (oh boy), then yeah, I think quirks is a natural shoe-in, and pacifism would most definitely be included as negative quirk.

KasparoVy:
Thanks for clearing that up, my bad on the misunderstanding

code/modules/mob/living/carbon/carbon.dm
Fox-McCloud: Yeah; this is broke, even on TG.

procdrone:
Oh my. I would totally love if the Pacifism trait was applied to people who are in crit and crawling. This would resolve so many issues people have with newcrit

PR #12549, opened 2019-10-10
Fixes Tribute Rune not producing full soul stone + light refactor.

### LINKED ISSUE #12494: Rite of Tribute is completely broken.
**Issue Description**:
The Rite of Tribute, the sacrifice rune, is completely wack.

I expected sacrificing human mobs, or cyborgs with ckeys, living or dead to yield a filled soul stone. Sacrificed simple mobs to simply die. Sacrificed humans or cyborgs without ckeys to not yield soul stones at all. I also expected all of these mobs to gib, or in the case of cyborgs, dust.

**What happened instead**:
When cultists sacrifice their target, while the target is alive, the target gibs as they're supposed to, but is not put into a soul stone. Objectives update properly.

When sacrificing a live cyborg, the cyborg dusts properly, but they're not placed into a soul stone.

When sacrificing a dead cyborg, reality shatters. 
sacrifice0
A bunch of items that should never appear, appear, and the ghost of the dead cyborg is placed into a soul stone, that is instantly deleted, causing them to be ghosted instantly as a shade.

Sacrificing a live PAI works as intended. The PAI gibs and is NOT placed into a soul stone.

Sacrificing a dead PAI causes God to weep.
sacrifice1
Even more items that should never appear, appear, and the same thing with the dead cyborg happens to the PAI's ghost. Additionally, I have a ckeyless PAI with the name PAI debris.

Sacrificing ckey-less simple mobs works as intended both dead and alive.
Sacrificing dead simple mobs with a ckey, attempts to put them in a soul stone but fails. Ideally there would be no attempt.

Sacrificing a ckey-less human mob, alive and dead, works as intended and doesn't give a soul stone.

Sacrificing a human mob with a ckey does not give a soul stone, alive or dead. Dead however, attempts to put the ghost into the soulstone and fails.

Finally, attempting to sacrifice dogs both fails and succeeds. Normally you're just supposed to catch fire, but instead you both catch fire and the sacrifice goes through.

Aside from spawning things that should never be spawned, it also dusts people without putting them back in the round.

Try out the rite of tribute with some fellow cultists. 

No idea. Probably since forever.

**Extra information**:
Cult code is broke.

JKnutson101:
## What Does This PR Do
Fixes #12494 

While a straight fix could have been much easier if I just gutted a lot of the rune, I also made it a point to maintain previously intended functionality. What I mean by this is, you can still sacrifice just about any mob, with a exception of dogs. However, I adjusted it so only humans and cyborgs produce and attempt to fill soul stones when sacrificed, and only when the said human or cyborg has had a ckey at one point in its existence. So no cheesy free cult ghosts by sacrificing PAI's or something like that.

I changed a few lines here and there in soul stone code, like lines deleting mobs before we're done with them. (causing cyborgs to spill out their internal bits you're never supposed to see)

## Why It's Good For The Game
Makes an intended feature *work* while accounting for potential exploits. I did extensive testing. Sacrificing objective still works, as that part of the code wasn't touched. Adjusted for strange edge cases like a client leaving and then rejoining mid ghost-chat poll for filling the soul stone.

In short, it makes the Sacrifice Rune work as intended, giving the one sacrificed a bit of a consolation prize by putting them in a soul stone and letting them do some cult stuff, or lets ghosts take their place if the player has already ghosted.

My coding is likely poor quality, and I would appreciate feedback. If owners think we should do without some of the features like Nar'Sie being a dog person, or sacrificing any and all mobs for a meager sacrifice flavor text, I am 100% willing to ditch those.

Edge cases I did not test: Cortical Borers in the sacrificee, and the sacrificee being antag banned.

## Changelog
:cl:
fix: Fixed sacrificing rune not producing filled soul stones.
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/soulstone.dm
+++ b/code/game/gamemodes/wizard/soulstone.dm
@@ -206,19 +206,25 @@
 	switch(choice)
 		if("FORCE")
 			var/obj/item/soulstone/C = src
-			if(!iscarbon(target))		//TODO: Add sacrifice stoning for non-organics, just because you have no body doesnt mean you dont have a soul
-				return 0
-			if(contents.len)
-				return 0
-			var/mob/living/carbon/T = target
-			if(T.client != null)
+			var/mob/living/T = target
+			if(issilicon(T))
+				T.Paralyse(10) //In the off chance there's no client, but it suddenly gets one while polling deadchat.
+				if(T.client == null)
+					C.getCultGhost(T,U)
+				else
+					C.init_shade(T,U)
+				playsound(T, 'sound/effects/EMPulse.ogg', 100, 1)
+				T.dust() //To prevent the MMI from remaining
+			else
+				T.Paralyse(10) //In the off chance there's no client, but it suddenly gets one while polling deadchat.
+				if(T.client == null)
+					C.getCultGhost(T,U)
+				else
+					C.init_shade(T, U)
 				for(var/obj/item/W in T)
 					T.unEquip(W)
-				C.init_shade(T, U)
-				return 1
-			else
-				to_chat(U, "<span class='userdanger'>Capture failed!</span>: The soul has already fled its mortal frame. You attempt to bring it back...")
-				return C.getCultGhost(T,U)
+				playsound(T, 'sound/misc/demon_consume.ogg', 100, 1)
+				qdel(T)
 
 		if("VICTIM")
 			var/mob/living/carbon/human/T = target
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/wizard/soulstone.dm
```diff
@@ -395,8 +404,8 @@
 	if(contents.len) //If they used the soulstone on someone else in the meantime
 		return 0
 	T.ckey = chosen_ghost.ckey
-	for(var/obj/item/W in T)
-		T.unEquip(W)
+	if(ishuman(T))
```
Couls: why is this check here?

JKnutson101: If I just do the T.unEquip(W) loop on something without making sure it's human, like on say a cyborg, it will dump out all of it's internal pieces it shouldn't. 

At least that's what I assume was causing this issue.

sacrifice0

JKnutson101:
I'm thinking of closing this and trying a straight up refactor into TG's combination of Sacrifice and Tribute runes + the fix.

PR #12552, opened 2019-10-10
Implements signals and some components. Also refactors handle_changeling to use a signal

farie82:
## What Does This PR Do
Adds a few components I've been using for my project but might be handy for others. If not just as an example

Also refactors handle_changeling to use the new life signal

## Why It's Good For The Game
Can be handy for things like replacing the ugly handle_changeling call in every carbon
Less cost for Life due to handle_changeling not being called each time.

## Changelog
No CL since no game feature changes


Couls:
Why does the on life component have to be on a carbon? Should work on living, same with damage callback

farie82:
Honest answer? Because I ported them from my code which had that as a must. I'll change it in a bit

farie82:
Better answer. Because I wanted to have the callbacks be done after the life proc. And the ones before carbon (living etc) are called prior to the carbon life proc.

For apply damage it's the case that humans don't use the base procs.

You can add this in if you need it. Make a check on the type of the parent and register based on that.

code/datums/components/living_no_move.dm
```diff
@@ -0,0 +1,19 @@
...
+	RegisterSignal(parent, COMSIG_LIVING_UPDATE_CAN_MOVE, .proc/update_canmove)
+
+/datum/component/living_no_move/UnregisterFromParent()
+	var/mob/living/M = parent
+	M.canmove = TRUE
+	UnregisterSignal(parent, COMSIG_LIVING_UPDATE_CAN_MOVE)
+
+/datum/component/living_no_move/proc/update_canmove()
+	var/mob/living/M = parent
+	M.canmove = FALSE
```
Fox-McCloud: I really don't get the point of this, to be honest, when there's multiple effects that can be applied that prevent someone from moving---like `Stun`. 

I also don't like procs that have the same exact proc name as a different proc, but their functionality is different (but *semi* related). It's bad practice and just generates confusion.

code/datums/components/human_damage_callback.dm
```diff
@@ -0,0 +1,26 @@
...
+	var/total_damage = 0
+
+// Takes a limit and a callback as manditory arguments. Will not unregister itself after the callback is called
+// Optionally takes the allowed damage types as argument in the form of (BRUTE, BURN) for example
+/datum/component/human_damage_callback/Initialize(damage_limit, callback, damage_types = null)
+	var/mob/living/carbon/human/H = parent
+	if(!istype(H) || damage_limit <= 0 || !callback) //Something went wrong
+		return COMPONENT_INCOMPATIBLE
+	src.damage_limit = damage_limit
+	src.callback = callback
```
Fox-McCloud: To quote a former maintainer. If you have to use `src.var_name = var_name`, then you should probably rename your vars.

code/datums/components/carbon_on_life.dm
```diff
@@ -0,0 +1,18 @@
...
+		return COMPONENT_INCOMPATIBLE
+	src.callback = callback
+	RegisterSignal(C, COMSIG_CARBON_LIFE, .proc/Life)
+
+/datum/component/carbon_on_life/UnregisterFromParent()
+	UnregisterSignal(parent, COMSIG_CARBON_LIFE)
+
+/datum/component/carbon_on_life/proc/Life(mob/living/carbon/C)
+	if(C.stat != DEAD)
+		callback.Invoke(C)
```
Fox-McCloud: This is another one I'm a *bit* sketchy on too...what needs to have a component that gets called every `Life` that isn't an organ, genetic mutation, or reagent?

code/datums/components/human_damage_callback.dm
```diff
@@ -0,0 +1,26 @@
...
+	RegisterSignal(H, COMSIG_HUMAN_APPLY_DAMAGE, .proc/apply_damage)
+
+/datum/component/human_damage_callback/UnregisterFromParent()
+	UnregisterSignal(parent, COMSIG_HUMAN_APPLY_DAMAGE)
+
+/datum/component/human_damage_callback/proc/apply_damage(mob/living/carbon/human/H, damage, damagetype, def_zone, blocked, sharp, obj/used_weapon)
+	if(damage > 0 && (!allowed_damage_types || damagetype in allowed_damage_types)) // Only count damage. Not healing
+		total_damage += damage
+		if(total_damage >= damage_limit)
+			callback.Invoke()
```
Fox-McCloud: If I'm understanding this correctly, this could be used like---if a debuff is placed and you deal 10 damage to them them, then they fart or something like that.

Interesting, though I do wonder if this would be better as a status effect than a component, for the simple fact you might want it to be a debuff that is applied for a short time period that expires, rather than something you have to turn on/turn off as a component (which is a bit less ideal).

farie82: Ill give some context to why I made these. 
I wanted to make a cast of a channel ability (ability with more than 1 stage usually) stop once the caster received enough damage.
Stepping away from the stun wins against everything and giving lethals some use as well.

The debuff using a status effect does sound nice

code/datums/components/carbon_on_life.dm
farie82: Again some context.
Ive used this one to make a new antag that does things on life. Like our current changeling which just has "handle_changeling" as proc which is called for every carbon in the game.

code/datums/components/human_damage_callback.dm
farie82: Will change once able

code/datums/components/living_no_move.dm
farie82: Will change the naming.
Some context why I made it.
I wanted to have abilities take a long time and once they were channeling I wanted the user to be unable to move. Using stun here would break the channel since I use a do_after

code/datums/components/carbon_on_life.dm
farie82: This component is not required. I didn't fully understand signals. Will update the PR soon

farie82:
Thanks to insights from ninja I changed the PR a bit. The carbon life signal now is used by changelings and the component I made for it is not required.

PR #12556, opened 2019-10-10
Gives Captains access to unlock their display case again

Ty-Omaha:
## What Does This PR Do
Changes the access requirements of the Captain's Display Case back to Captain level access from CC access 

:cl:
tweak: Captains can access their display case again
/ :cl:

### DIFF
--- a/code/game/objects/structures/displaycase.dm
+++ b/code/game/objects/structures/displaycase.dm
@@ -217,12 +217,11 @@
 	else
 		return ..()
 
-//The captains display case requiring specops ID access is intentional.
-//The lab cage and captains display case do not spawn with electronics, which is why req_access is needed.
+
 /obj/structure/displaycase/captain
 	alert = TRUE
 	start_showpiece_type = /obj/item/gun/energy/laser/captain
-	req_access = list(access_cent_specops)
+	req_access = list(access_captain)
 
 /obj/structure/displaycase/labcage
 	name = "lab cage"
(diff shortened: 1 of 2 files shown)


variableundefined:
Snuck?
It is right there in the changelog
image

Fox-McCloud:
"snuck in"

I love the hostility this implies like I was trying to be sneaky, when it was one of the things *deliberately mentioned* in the PR itself, multiple times: https://github.com/ParadiseSS13/Paradise/pull/12245 https://github.com/ParadiseSS13/Paradise/pull/12245#issuecomment-538592459

I *did* forget to ask a few people about this, admittedly, but to suggest that it wasn't documented or I was trying to intentionally "sneak" it in, is uncalled for.

florkeger:
I don't get it. What's the idea behind making the Captains personal display case be only accessable by CC staff? Sort of a unneccessary change without a lot reason behind it, in my opinion.

Gotta' slap an agree on the PR for this one.

Mitchs98:
Yeahh. Seems really pointless to disallow the Captain from accessing the case. If they really want the gun they'll just break the case and...steal their own gun? :thinking:

Carthusia:
I've seen more than a few times while playing security "Ignore the theft alarm in my office" as the captain wants his pistol

Spacemanspark:
Looking back and seeing the PR I don't think it was really snuck in, although I think this could have been spoken out a bit better as it is a rather interesting change.

I don't think it's a change that makes much sense, however. I don't see why the captain shouldn't be entitled to his own weapon, and nobody is really going to stop them from taking it.

Mitchs98:
Time to brig the Captain for Grand Theft of his own gun!

PR #12560, opened 2019-10-11
Make cult whispering use their real name

farie82:
## What Does This PR Do
Makes cult whispering use their real name instead of their visible name

## Why It's Good For The Game
It's handy to know which cultist is talking. And not what ID they are wearing

## Changelog
:cl:
tweak: Cult whispering now uses your real name
/:cl:


JKnutson101:
I always forgot this would happen, and if you're doing well as a cult and getting a few spare IDs from officers and the like, it can get really confusing. spd mrg plz

craftxbox:
oh my god i was literally thinking about this yesterday.

KasparoVy:
Will the real Urist McCultist please speak up?

SteelSlayer:
Yay no more unknowns speaking in cult chat

PR #12561, opened 2019-10-11
Fixes IsBanned Bug

Kyep:
## What Does This PR Do
Fixes a bug with an IsBanned check where the arguments are passed in the wrong order, which stops banned people from being recognized as banned.

## TG Information
TG noticed as well (code is shared) and implemented the same fix here: https://github.com/tgstation/tgstation/pull/42899/files#diff-6895b8588eaf4eb88f939ed4ceb7d67b

## CL
No CL as this is backend only.


Fox-McCloud:
RIP

PR #12566, opened 2019-10-12
Makes LINDA Aggressive

Fox-McCloud:
Requesting that @MrStonedOne take a look at this.

I realize you guys have datumized gases and we don't, but I figure we can get the code *relatively* the same, function wise.

Either case, this ports @MrStonedOne's LINDA update from eons ago that makes LINDA more aggressive about breaking down excited groups and evening out the gas among them. 

It means LINDA is, overall, less active and moves gas faster among a larger amount of turfs.

This also fixes planetary atmos--our code expects that if there *IS* a difference between two turfs with gas, it returns `FALSE`. When @Markolie ported planetary atmos, he didn't realize that TG has inverted that logic (and made is more readable IMO...who expects a `FALSE` return to mean 'Yep, there's a difference!"), ergo, the only time atmos was sharing its gas with turfs was when there *wasn't* a difference between turfs.

:cl: Fox McCloud
tweak: LINDA is much more aggressive at distributing gas between turfs in an excited group
/:cl:

### DIFF
--- a/code/datums/gas_mixture.dm
+++ b/code/datums/gas_mixture.dm
@@ -368,20 +368,20 @@ What are the archived variables for?
 
 	var/delta_temperature = (temperature_archived - model.temperature)
 
-	if(((abs(delta_oxygen) > MINIMUM_AIR_TO_SUSPEND) && (abs(delta_oxygen) >= oxygen_archived*MINIMUM_AIR_RATIO_TO_SUSPEND)) \
-		|| ((abs(delta_carbon_dioxide) > MINIMUM_AIR_TO_SUSPEND) && (abs(delta_carbon_dioxide) >= carbon_dioxide_archived*MINIMUM_AIR_RATIO_TO_SUSPEND)) \
-		|| ((abs(delta_nitrogen) > MINIMUM_AIR_TO_SUSPEND) && (abs(delta_nitrogen) >= nitrogen_archived*MINIMUM_AIR_RATIO_TO_SUSPEND)) \
-		|| ((abs(delta_toxins) > MINIMUM_AIR_TO_SUSPEND) && (abs(delta_toxins) >= toxins_archived*MINIMUM_AIR_RATIO_TO_SUSPEND)))
-		return 0
+	if(((abs(delta_oxygen) > MINIMUM_MOLES_DELTA_TO_MOVE) && (abs(delta_oxygen) >= oxygen_archived*MINIMUM_AIR_RATIO_TO_MOVE)) \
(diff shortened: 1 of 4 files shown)


code/datums/gas_mixture.dm
```diff
@@ -638,46 +638,53 @@ What are the archived variables for?
...
-		var/datum/gas/trace_gas = gas
-		if(trace_gas.moles_archived > MINIMUM_AIR_TO_SUSPEND)
-			var/datum/gas/corresponding = locate(trace_gas.type) in trace_gases
-			if(corresponding)
-				if((abs(trace_gas.moles - corresponding.moles) > MINIMUM_AIR_TO_SUSPEND) && \
-					((corresponding.moles < (1-MINIMUM_AIR_RATIO_TO_SUSPEND)*trace_gas.moles) || (corresponding.moles > (1+MINIMUM_AIR_RATIO_TO_SUSPEND)*trace_gas.moles)))
-					return 0
-			else
-				return 0
+	delta = abs(carbon_dioxide - sample.carbon_dioxide)
```
Fox-McCloud: This is the portion that I'd like to have the most scrutiny, since we don't have list-based gases.

code/datums/gas_mixture.dm
```diff
@@ -368,20 +368,20 @@ What are the archived variables for?
 
 	var/delta_temperature = (temperature_archived - model.temperature)
 
-	if(((abs(delta_oxygen) > MINIMUM_AIR_TO_SUSPEND) && (abs(delta_oxygen) >= oxygen_archived*MINIMUM_AIR_RATIO_TO_SUSPEND)) \
```
Fox-McCloud: This and the proc, below, as well...they're both disgusting procs that mimic a comparison between a non-simulated turf and a simualted one.

taukausanake:
So it's NOT actively trying to kill me more than normal? And here I thought "aggressive" was something to get excited over :(

farie82:
So what exactly does this do? That turfs now share atmos better making gas flow better or?

Fox-McCloud:
Basically, when you have a group of turfs, adjacent to each other, all excited, they're part of a group called an "excited group". This is kinda like a ZAS zone,  but it still behaves on a per turf basis.

Eventually, when the group has been in an excited state for a long enough time, it eventually dismantles itself, which averages the gas of the group out over all of the turfs.

This makes LINDA much more aggressive about averaging out the gas of the turfs in an excited group when only a small amount of gas is moving around in them---and excited groups themselves with be less aggressive about putting htemselves in a sleeping state.

This means LINDA runs less and averages out gas among excited groups far more often, which helps cut down on "someone farted in maintenance, now most of it is in an activated state for the rest of the shift just to spread that tiny atmospheric change around".

the fix for planetary atmos means far far far less active turfs on lavaland.

PR #12571, opened 2019-10-13
Rite of Offering - Refactors Convert and Sacrifice into one rune, fixes Sacrifice bugs.

### LINKED ISSUE #12494: Rite of Tribute is completely broken.

JKnutson101:
## What Does This PR Do
Fixes #12494

This PR combines the Rite of Tribute and Rite of Enlightenment into one rune that converts the victim if it can, and if it's unable to convert due to mindshield, chaplainship, etc. it will smartly sacrifice the victim instead.  With the above listed fix, it will now properly put those sacrifice victims into soul stones. Humans and Cyborgs that have had C-Keys at some point, but currently don't have one, can still be sacrificed and poll dead chat to fill the soul. Other mobs and Humans and Cyborgs that have never had C-Keys, can still be sacrificed but no reward comes of it. 

Also included are very light refactors of soul stones to stop cyborgs from spilling their bits that should never exist detailed in #12494. Another light refactor of is_convertable_to_cult to include a check for golems.

Credit to TG's code for the name, idea, and the way the code was formatted.

## Why It's Good For The Game
Combining two runes where you essentially do the same thing, and only switch out which rune to use if the person is mind-shielded or not, saves on the bloat of choice for the cult and makes things just a tad less overwhelming for new cult members. Fixing the sacrifice to actually function and stick the victim into a soul stone gets rid of a nasty bug that takes people out of the round permanently.

Finally, a few cleanups of cult/soulstone code to make it easier to work with, and handles a few unfair exploits with sacrificing before they have a chance to become a problem.

One step closer to TG Cult. 👀 

## Changelog
:cl:
tweak: Combined the Rite of Tribute and Rite of Enlightenment into one Rite of Offering.
fix: Sacrificing now properly puts victims that have at one point had CKeys into soul stones, or polls dead-chat to do so.
/:cl:


datlo:
There's an undocumented change in your PR : Golems can no longer be converted.

JKnutson101:
> 
> 
> There's an undocumented change in your PR : Golems can no longer be converted.
> 
> EDIT : Turns out that golems can already not be converted from another PR that placed the check in the conversion rune itself. _shrugs_

Yeah I filed that under cleanups, considering we have a is_convertable_to_cult helper, but we included golems explicitly on the conversion I just moved the check to there.

code/game/gamemodes/cult/runes.dm
```diff
@@ -340,144 +340,101 @@ var/list/teleport_runes = list()
...
-				return
-		if(isrobot(T))
-			playsound(T, 'sound/effects/EMPulse.ogg', 100, 1)
-			T.dust() //To prevent the MMI from remaining
+	if(offering.mind)
+		sacrificed.Add(offering.mind)
+		if(is_sacrifice_target(offering.mind))
+			sacrifice_fulfilled = 1
+	new /obj/effect/temp_visual/cult/sac(loc)
+	if(SSticker && SSticker.mode && SSticker.mode.name == "cult")
```
farie82: SSTicker should be checked earlier. Since you already take the mode.
Or better yet move `var/datum/game_mode/cult/cult_mode = SSticker.mode` to after the if or refactor it a bit.

code/game/gamemodes/wizard/soulstone.dm
```diff
@@ -206,19 +206,14 @@
 	switch(choice)
 		if("FORCE")
 			var/obj/item/soulstone/C = src
-			if(!iscarbon(target))		//TODO: Add sacrifice stoning for non-organics, just because you have no body doesnt mean you dont have a soul
-				return 0
-			if(contents.len)
-				return 0
-			var/mob/living/carbon/T = target
+			var/mob/living/T = target
```
farie82: You forgot an istype check here for target prior to parsing

code/game/gamemodes/cult/runes.dm
```diff
@@ -340,144 +340,101 @@ var/list/teleport_runes = list()
...
-			if(!T)
-				rune_in_use = 0
-				return
-		if(isrobot(T))
-			playsound(T, 'sound/effects/EMPulse.ogg', 100, 1)
-			T.dust() //To prevent the MMI from remaining
+	if(offering.mind)
+		sacrificed.Add(offering.mind)
+		if(is_sacrifice_target(offering.mind))
+			sacrifice_fulfilled = 1
```
farie82: TRUE

code/game/gamemodes/cult/runes.dm
```diff
@@ -340,144 +340,101 @@ var/list/teleport_runes = list()
...
-	if(!is_convertable_to_cult(new_cultist.mind) || new_cultist.null_rod_check())
-		for(var/M in invokers)
-			to_chat(M, "<span class='warning'>Something is shielding [new_cultist]'s mind!</span>")
-			if(is_sacrifice_target(new_cultist.mind))
-				to_chat(M, "<span class='cultlarge'>\"I desire this one for myself. <i>SACRIFICE THEM!</i>\"</span>")
+	var/list/myriad_targets = list()
+	var/turf/T = get_turf(src)
+	for(var/mob/living/M in T)
+		if(!iscultist(M))
+			myriad_targets |= M
```
farie82: should not need to be |=. Unless the same target is listen twice or more on the turf for some reason

code/game/gamemodes/cult/cult.dm
```diff
@@ -26,6 +26,8 @@ var/global/list/all_cults = list()
 		var/mob/living/simple_animal/hostile/guardian/G = mind.current
 		if(!iscultist(G.summoner))
 			return 0 //can't convert it unless the owner is converted
+	if(isgolem(mind.current))
+		return 0
```
farie82: return FALSE. But then it would be better to refactor the whole proc if you change this

code/game/gamemodes/cult/runes.dm
```diff
@@ -340,144 +340,101 @@ var/list/teleport_runes = list()
...
-	cultist_desc = "converts a normal crewmember on top of it to the cult. Does not work on mindshielded crew."
+	cultist_name = "Rite of Offering"
+	cultist_desc = "Offers a non-cultists on top of it to your deity, either converting or sacrificing them."
 	invocation = "Mah'weyh pleggh at e'ntrath!"
 	icon_state = "3"
-	req_cultists = 2
-
+	req_cultists = 1
+	allow_excess_invokers = 1
+	rune_in_use = 0
```
farie82: FALSE

code/game/gamemodes/cult/runes.dm
```diff
@@ -340,144 +340,101 @@ var/list/teleport_runes = list()
...
-				to_chat(M, "<span class='cultlarge'>\"I desire this one for myself. <i>SACRIFICE THEM!</i>\"</span>")
+	var/list/myriad_targets = list()
+	var/turf/T = get_turf(src)
+	for(var/mob/living/M in T)
+		if(!iscultist(M))
+			myriad_targets |= M
+	if(!myriad_targets.len)
 		fail_invoke()
-		log_game("Convert rune failed - convertee could not be converted")
+		log_game("Offer rune failed - no eligible targets")
```
farie82: this happens everytime a cultist spams the rune right? Just be aware of that

code/game/gamemodes/wizard/soulstone.dm
```diff
@@ -206,19 +206,14 @@
...
 			if(T.client != null)
-				for(var/obj/item/W in T)
-					T.unEquip(W)
 				C.init_shade(T, U)
-				return 1
 			else
 				to_chat(U, "<span class='userdanger'>Capture failed!</span>: The soul has already fled its mortal frame. You attempt to bring it back...")
-				return C.getCultGhost(T,U)
+				T.Paralyse(10)
```
farie82: Some items and chems will reduce this time. Shouldn't be that much of an issue though since the player is out of the mob.

code/game/gamemodes/cult/runes.dm
JKnutson101: Well that's where SSticker is checked in the current sacrifice rune, and the current sacrifice rune properly works, and my testing showed it still did with my refactor. I'm completely inexperienced with SSticker and modes, if you know exactly where it should go please share, otherwise I'm not gonna toy with it.

code/game/gamemodes/wizard/soulstone.dm
JKnutson101: None of the other Soul Stone choices use an istype check before parsing target to a type. Also the only two ways of calling the "FORCED" choice is the sacrifice rune here, which won't be called at all unless it's something living, and the unused Cursed Russian Revolver. 

Right now I see adding a check there as redundant, and I would like to know the reasons it's needed if I'm wrong.

code/game/gamemodes/cult/runes.dm
JKnutson101: Thanks.

code/game/gamemodes/cult/runes.dm
JKnutson101: Ah, no idea what |= did, it was just part of how TG Code listed the targets. Thanks.

code/game/gamemodes/cult/cult.dm
JKnutson101: Thanks.

code/game/gamemodes/cult/runes.dm
JKnutson101: Thank you.

code/game/gamemodes/cult/runes.dm
JKnutson101: Yeah this looks unnecessary, I'll remove it.

code/game/gamemodes/wizard/soulstone.dm
JKnutson101: I'll just double the time since the duration doesn't really matter.

code/game/gamemodes/cult/runes.dm
farie82: Currently at the top it makes a variable cult_mode which takes the tickers game mode and casts it to an type.
Here in this if you check if SSTicker is not null and SSTicker.mode is not null and if SSTicker.mode is a cult objective. The first check, if Ticker is null is caught as an exception if it is null due to you using SSTicker above. The other 2 are tested because you parse the mode to the cult mode. If the mode is not cult then the parsed var (cult_mode) will fail if you use it. So you can do istype(cult_mode).

code/game/gamemodes/cult/runes.dm
```diff
@@ -360,7 +360,7 @@ var/list/teleport_runes = list()
 	var/turf/T = get_turf(src)
 	for(var/mob/living/M in T)
 		if(!iscultist(M))
-			myriad_targets = M
+			myriad_targets |= M
```
JKnutson101: Turns out removing that made it so the rune didn't work at all. No idea at all why. Added it back.

code/game/gamemodes/cult/runes.dm
JKnutson101: After chatting about it on the Discord, I get a bit more of what you're saying, but I still don't think I've got the chops for handling this. As it is now, the code works, and it's the code we already have, unchanged. I'll leave this refactor for someone that really knows what they're doing. This PR was about combing two existing runes into one, while fixing a couple functionality bugs.

code/game/gamemodes/cult/runes.dm
farie82: as reference for others. It adds an item to a list if it's not yet in the list. This is the case for lists. For other variables this is a bitwise operation.
+= will do here since the turf SHOULD not have duplicates of mobs.

variableundefined:
@JKnutson101 Conflicts, is this tested?

JKnutson101:
Heck! Let me check these really quick.

JKnutson101:
Okay okay dear fucking ME I fucked up hard.

JKnutson101:
Okay successfully reverted that terrible 3.A.M. rebase! Wowee!

JKnutson101:
Okay 3:30am merges work out just perfectly though.

JKnutson101:
And yes @variableundefined I did test it quite a bit back when I made this. However, if you would like me to test it again, with pictures and accounting for as many edge cases as I can, I'd be happy to.

PR #10882, opened 2019-02-11
Redundent cult convert checks

Aurorablade:
So uhh way back i had made it so you couldn't convert golems...or animals. (or i thought i did) I am hearing..that broke. I am fixing it now.

I know we got some new golem types but i am not getting runic golems in in this PR (and if anyone else wants to feel free), Mainly as i want a fix and runic golems need me to eval ease to make versus repercussion on round.

:cl:
fix:unbreaks letting golems be cult converts.
/:cl:

PR #12577, opened 2019-10-13
Prevent offstation roles from being converted

datlo:
This PR changes the cultist conversion check to include an offstation role condition. 

This will prevent ghost spawns such as beach bums, ashwalkers, etc from being converted. 

This is due to people spawning as them when the cult settles there just to be converted, and in return encouraging cultists to settle there for free conversions. For some spawns like ashwalkers they can even reproduce for even more free cultists.

This will also prevent rare edge cases conversions, like removing the mindshield of an ERT and converting them which isnt possible anymore, or converting other offstation roles like honk squad members/syndi infiltrators/abductors/traders/etc. Shouldnt really ever be an issue.

:cl:
balance: Offstation roles such as beachbums, ashwalkers, etc can no longer be converted to the cult.
balance: Roles such as ERT, abductors, traders, etc can no longer be converted aswell.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult.dm
+++ b/code/game/gamemodes/cult/cult.dm
@@ -9,24 +9,28 @@ var/global/list/all_cults = list()
 
 /proc/is_convertable_to_cult(datum/mind/mind)
 	if(!mind)
-		return 0
+		return FALSE
 	if(!mind.current)
-		return 0
+		return FALSE
 	if(iscultist(mind.current))
-		return 1 //If they're already in the cult, assume they are convertable
+		return TRUE //If they're already in the cult, assume they are convertable
 	if(ishuman(mind.current) && (mind.assigned_role in list("Captain", "Chaplain")))
-		return 0
+		return FALSE
 	if(ishuman(mind.current))
 		var/mob/living/carbon/human/H = mind.current
-		if(ismindshielded(H))
-			return 0
+		if(ismindshielded(H)) //mindshield protects against conversions unless removed
+			return FALSE
+		if(mind.offstation_role) //cant convert offstation roles such as ghost spawns...
+			if(isgolem(mind.current) && owner && iscultist(owner)) //... unless it's a servant golem of a cultist
+				return TRUE
+			return FALSE
 	if(issilicon(mind.current))
-		return 0 //can't convert machines, that's ratvar's thing
+		return FALSE //can't convert machines, that's ratvar's thing
 	if(isguardian(mind.current))
 		var/mob/living/simple_animal/hostile/guardian/G = mind.current
 		if(!iscultist(G.summoner))
-			return 0 //can't convert it unless the owner is converted
-	return 1
+			return FALSE //can't convert it unless the owner is converted
+	return TRUE
 
 /proc/is_sacrifice_target(datum/mind/mind)
 	if(SSticker.mode.name == "cult")
--- a/code/game/gamemodes/cult/cult.dm
+++ b/code/game/gamemodes/cult/cult.dm
@@ -20,10 +20,8 @@ var/global/list/all_cults = list()
 		var/mob/living/carbon/human/H = mind.current
 		if(ismindshielded(H)) //mindshield protects against conversions unless removed
 			return FALSE
-		if(mind.offstation_role) //cant convert offstation roles such as ghost spawns...
-			if(isgolem(mind.current) && owner && iscultist(owner)) //... unless it's a servant golem of a cultist
-				return TRUE
-			return FALSE
+	if(mind.offstation_role) //cant convert offstation roles such as ghost spawns
+		return FALSE
 	if(issilicon(mind.current))
 		return FALSE //can't convert machines, that's ratvar's thing
 	if(isguardian(mind.current))


datlo:
this is what I get for doing changes to the PR and not testing them
fixing compile issue

TheSardele:
I thought golems have been unconvertable for a while now, but can be sharded?

datlo:
Cant see any checks for golems in cult code as is on live

TheSardele:
Could have sworn that was a thing, have seen cults fail to convert free golems recently. Think it was this PR https://github.com/ParadiseSS13/Paradise/pull/10882

farie82:
More a balance thing than a del. But good shit

datlo:
oh. they placed it in the rune itself instead of the dedicated conversion check... i'll just remove the golem check stuff then

AdamElTablawy:
okay, but, two things

what’s the issue with converting, say, podpeople, or abductors? i can understand the logic behind golems and ashwalkers because infinite converts, but if a cult manages to successfully stun and convert an abductor before its scientist pulls it up, props to the cult?

datlo:
it's not really an issue but its better for consistency to have all of those special roles excluded instead of picking and choosing which one can get converted and which one can't for arbitrary reasons

AdamElTablawy:
i don't really see the inherent issue in encouraging the cult to base there 

as for ghosts being encouraged to spawn in for free antag, well that's basically just 4 converts, and cultists could just manifest ghost those dudes instead. If they REALLY want free antag they'd spawn in anyway to get soulsharded after this PR

i can again understand the logic behind ashwalkers and so forth but for limited-role entities what's the issue there

also as for the abductors and stuff, it really isn't a 'consistency' thing at all. ERTs would get excluded because like the captain they'd be unconvertable period due to being loyal. there's no real reason to exclude abductors and so forth from the pool of viable candidates just because ERTs are being blacklisted.

datlo:
spawning just so you can get converted to cult is breaking server rules (although not explicitely, it's basically the same as running into maint/cult bases to get culted/thralled which isnt allowed)

when you can discourage breaking server rules by making it harder/impossible through code, it is better to do so

Pixelzilla:
I don't agree that you should always have a mechanical restriction to prevent unwanted behavior. That's what admin intervention, warnings and antag bans are for. You don't go nuclear and alter the code when the problem is individuals. You go after the individuals causing the problem.

There are very real negative repercussions for doing something like this that quite a few people seem to be overlooking. Essentially doing this sort of PR means that there is now going to be numerous roles that are absolutely 100% immune to being converted and that means they are now 100% trustworthy and ideal candidates for validhunters.

I can understand Ashwalkers being immune purely based on the fact that they basically are their own little cult with strong beliefs and would likely not be swayed from that. They also are not supposed to really leave their area or interact with people much outside of their community. Their lore/restrictions would support immunity.

Golems can just "go do whatever" so while I understand the frustration at having a seemingly endless army of them just because they can be converted doesn't really mean they will be and I at least haven't seen it happen as much as the PR seems to imply it does. If its really an issue with cult specifically going for golems I think the better option is to deal with the players who are actively going for golems every single time to curb the behavior before outright making yet another thing into an immune validhunter's dream.

Mitchs98:
Cultists shouldn't be given basically free recruits and should be discouraged from camping mid-round ghost spawn areas. That's the main issue here. That's the main issue this is trying to solve. Sure, there is the issue of having 100% trustworthy candidates, but in my opinion the bigger concern here is how badly Cultists are more or less encouraged to set up shop in said areas and/or camp them for members.

Frankly if it was up to me I'd have the spawn areas disallow runes and cultists structures in general, though that would have even bigger balance concerns and potentional outlash. This in my opinion is a happy medium.

Pixelzilla:
> 
> 
> Cultists shouldn't be given basically free recruits and should be discouraged from camping mid-round ghost spawn areas. That's the main issue here. That's the main issue this is trying to solve. Sure, there is the issue of having 100% trustworthy candidates, but in my opinion the bigger concern here is how badly Cultists are more or less encouraged to set up shop in said areas and/or camp them for members.
> 
> Frankly if it was up to me I'd have the spawn areas disallow runes and cultists structures in general, though that would have even bigger balance concerns and potentional outlash. This in my opinion is a happy medium.

The entire point of conversion antags is to convert people to their cause. Thats the point of those types of modes. The more people who cannot be converted because of some code restriction the more you complicate and water down the antagonist role while giving a significant number of roles an extreme advantage over the threat that is supposed to spice up the round.

SS13 is a game about paranoia, about chaos, about terrible things happening that make the round FUN!! It isn't about safezones, bubblewrap and balance/fairness for everyone.

Just consider for a moment the reason _why_ so many cults/conversion antags **do** go off station. Its because people who are not antags generally metacheck and screech the instant they notice something shady because there is zero antag knowledge restrictions and numerous people who play to "win" or to prevent antags from "winning". 

The implication that the cult is "given free converts" purely because some ghost roles can be converted is really weird, imo. The assumption seems to be that anyone playing a ghost role is self antagging and allowing conversion to happen and that absolutely isn't true. What makes ghost roles that are not infinite spawning roles (like golems/ashwalkers) any different than people arriving to the station late? Why does the ERT/Traders/Vet/BeachBum/Etc deserve special snowflake protections with no counterplay that late arrivals are not afforded?

Do you really think it adds anything positive to the the round as a whole to essentially have numerous roles _unable_ to participate in the round type in anything other than validhunting fashion? Well, either that or they're going to get murdered because they cant be converted, doesn't sound very fun.

MrMagolor:
I don't get why ERT shouldn't be able to be converted. That's like saying that sec shouldn't be able to be converted. The only thing ICly preventing an ERT's conversion is their mindshield. And conversion should always be preferable to sharding.

Fox-McCloud:
Highly skeptical of this. If people are abusing offstation roles to the degree of "intentionally set up near obvious off station spawns" or "spawn in just to be converted", then that's really an administrative issue where-in the person should probably be job-banned from cult or all ghost roles.

Patching over every issue, mechanically, takes away a lot of varied and legitimate options when administration exists for this very reason.

variableundefined:
Let's try this out

PR #12580, opened 2019-10-13
Captain Nanotrasen: The First Vindicator

Ty-Omaha:
## What Does This PR Do
Adds CAPTAIN NANOTRASEN as an admin spawn outfit. Has CQC training.
Sprites by Ionward
Suit compatibility with lots of species too.
Adds a clangsingle sound where its a single clang instead of double

## Why It's Good For The Game
So Admins can send in Captain Nanotrasen to save the day. _(or record a video for Nanotrasen sponsored school gyms)_

## Changelog
:cl: Ty-Omaha and Ionward
add: Captain Nanotrasen, can be spawned as an admin outfit equip
soundadd: Added clangsingle.ogg, version of clang but with a single clang
/ 🆑


Terilia:
https://www.youtube.com/watch?v=cVU4HURKEXs

Yup.

taukausanake:
I like this, but I'm disappointed the shield won't fly back to you if you throw it at someone. I'd really like that. Maybe even a special photocopy ass that represents Nanotrasen

lolzyking:
Perhaps give them recall?

McRamon:
Custom recall which is used on shield by default but cant be changed maybe

Fox-McCloud:
Not so sure this is a good idea, to be honest.

AffectedArc07:
Reminder that marvel is owned by disney, who have probably the most aggro lawyer team, possibly in the world

taukausanake:
Reminder that we also have Spiderman sprites which, while not Disney right was was for a while.

While we're at that, Star Wars is owned by Disney too and don't we have literal droidekas for combat cyborgs? They may be all black but their structure and design is the same. I'm sure there's a few more examples laying about. Not a lot but at least a few others

JKnutson101:
> Reminder that we also have Spiderman sprites which, while not Disney right was was for a while.
> 
> While we're at that, Star Wars is owned by Disney too and don't we have literal droidekas for combat cyborgs? They may be all black but their structure and design is the same. I'm sure there's a few more examples laying about. Not a lot but at least a few others

Let's also not forget that the E-Sword is literally a lightsaber with it's sound effects, possibly one of the most iconic sound effects in media.

Ty-Omaha:
>Just because we have some copyrighted stuff in the code doesnt mean we want more things.

Except it makes your entire argument with false concern pointless. You're saying how big and bad disney lawyers are. If they are really that bad then the addition of this won't make a difference at all.

There's a reason why we break people's ribs in CPR - they're already dead and can't get deader.

We already have copyright stuff. The amount doesn't matter.

Necaladun:
Could we swap the colours to NT colours?

Mitchs98:
Love the outfit myself. Swapping colors would probably be a nice thing just for the NT aesthetic. Siding with Ty though. If you're going to get sued for copyright infringement you're going to get sued for it for one item. Adding more things won't really make things any more illegal. Especially when as stated we quite literally have lightsabers with sound included sans the name.

Megarop:
> I like this, but I'm disappointed the shield won't fly back to you if you throw it at someone. 

tg already has the bananium shield do exactly that so surely we can just yoink whatever code it uses for that

variableundefined:
Not a fan of this either. Closing due to objections.

PR #12582, opened 2019-10-13
Stops spamming admins with heirophant club attack logs

Ty-Omaha:
Changes the heirophant club attack log to ATKLOG_ALL unless they are on the station so that admins can enable attack logs again without having this take up the whole chat

🆑 
tweak: Admin attack log level of heirophant club updated
/ 🆑


farie82:
Will the hits also log? If so why not just log those instead of the usages of the staff?

Kyep:
This may help.
Personally, I'm increasingly of the opinion that the heiro/ash/lava staves should simply not work at all outside of lavaland. They have many issues... attack log spam is arguably the least serious. This PR will help a bit, perhaps, but really all of these items need a very hard look, and probably a hard nerf or restriction applied to them.

Ty-Omaha:
I agree, this is just a small admin log tweak to hopefully be merged soon compared to a balance tweak like that

PR #12584, opened 2019-10-13
Fixes Planetary Atmos

Fox-McCloud:
Inb4 this is the source of the problem in the previous PR.

This should fix planetary atmos not functioning properly

:cl: Fox McCloud
fix: Fixes planetary atmos not correctly happening
/:cl:


KasparoVy:
I hope so, that'd be a relief.

PR #12587, opened 2019-10-13
Training Wheels Come Off

Fox-McCloud:
Lavaland has been around a while and people are starting to figure out how to play effectively--megafauna have been updated (with the exception of Legion) and rebalanced; it's time for the training wheels to come off, however. 

From TG

Changes made:
- Survival medical wall vendor doesn't have any medicine in it anymore; it now just has two sets of splints
- Using a regenerative core is no longer a full heal, instead it heals 25 burn, 25  brute, removes all CC, sets your body temperature back to normal, mends all internal bleeding, and fixes all broken bones
  - Implanting a regenerative core will still provide a full heal, but you can only ever have one implanted in you, and you'll actually have to get surgery to get this

Other:
- Refactors fracturing mending to be more consistent; it also doesn't require the limb to be under minimum broken damage to mend the fracture, as fractures aren't guaranteed anymore

:cl: Fox McCloud
tweak: Survival pod medical vendors only have 2 pairs of splints instead of a wide array of medicines
tweak: Hivelord core, when used on someone, no longer fully revives them; instead it heals 25 burn, 25 brute, resets body temperature, removes all CC, cures all internal bleeding, and mends all fractures (slowdown bonus still stays
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -113,7 +113,16 @@
 
 /datum/status_effect/regenerative_core/on_apply()
 	owner.status_flags |= IGNORESLOWDOWN
-	owner.revive()
+	owner.adjustBruteLoss(-25)
+	owner.adjustFireLoss(-25)
+	owner.remove_CC()
+	owner.bodytemperature = BODYTEMP_NORMAL
+	if(ishuman(owner))
+		var/mob/living/carbon/human/H = owner
+		for(var/thing in H.bodyparts)
+			var/obj/item/organ/external/E = thing
+			E.internal_bleeding = FALSE
+			E.mend_fracture()
 	return TRUE
 
 /datum/status_effect/regenerative_core/on_remove()
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -579,8 +579,8 @@
 		adjustFireLoss(-60)
 		for(var/obj/item/organ/external/E in bodyparts)
 			if(prob(25))
-				if(E.mend_fracture())
-					E.perma_injury = 0
+				E.mend_fracture()
+
 		return
 	if(stat != DEAD)
 		if(weakened)
--- a/code/modules/mining/equipment/survival_pod.dm
+++ b/code/modules/mining/equipment/survival_pod.dm
@@ -165,15 +165,8 @@
 	desc = "Wall-mounted Medical Equipment dispenser. This one seems just a tiny bit smaller."
 	req_access = list()
 
-	products = list(/obj/item/reagent_containers/food/pill/patch/styptic = 5,
-					/obj/item/reagent_containers/food/pill/patch/silver_sulf = 5,
-					/obj/item/reagent_containers/food/pill/charcoal = 2,
-					/obj/item/stack/medical/bruise_pack/advanced = 1,
-					/obj/item/stack/medical/ointment/advanced = 1,
-					/obj/item/reagent_containers/hypospray/autoinjector = 2,
-					/obj/item/stack/medical/splint = 1)
-	contraband = list(/obj/item/reagent_containers/food/pill/tox = 2,
-	                  /obj/item/reagent_containers/food/pill/morphine = 2)
+	products = list(/obj/item/stack/medical/splint = 2)
+	contraband = list()
 
 //Computer
 /obj/item/gps/computer
--- a/code/modules/mob/living/carbon/human/species/skeleton.dm
+++ b/code/modules/mob/living/carbon/human/species/skeleton.dm
@@ -48,12 +48,7 @@
 			var/list/our_organs = H.bodyparts.Copy()
 			shuffle(our_organs)
 			for(var/obj/item/organ/external/L in our_organs)
-				if(istype(L))
-					if(L.brute_dam < L.min_broken_damage)
-						L.status &= ~ORGAN_BROKEN
-						L.status &= ~ORGAN_SPLINTED
-						H.handle_splints()
-						L.perma_injury = 0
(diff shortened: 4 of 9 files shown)


TDSSS:
I like nerfing the survival vendors but IMO there should be some gauzes and oinment in there too. Having just 1 item feels a bit empty and gauze/oinment aren't OP.

Ty-Omaha:
It doesn't make much sense for a "medical vendor" to only have splints, as tds said maybe throw gauze and ointment in there.

Fox-McCloud:
I would have removed it altogether, but broken bones are highly problematic in a combat context, so I felt they should stay---that said, leaving them on the table, or putting them in the fridge is a *bit* silly.

I'd like to keep medical supplies down to practically nothing, even for something as basic as bandages.

Qwertytoforty:
_**TDLR: A single medkit should remain in the survival pods, or be available for 400 points the loss of 3 off each patch type and the advanced brute / burn kit isn't that bad from survival pods.**_

I like the core changes. The pod med bay changes however...
Lets do some testing.
We will be generous, and give the miner 3 goliath plates, and an un-upgraded KA.

**_Healing:_**
A full load of saline glucose is 20 units, and drains at 0.15 per cycle. This means it will last 133 cycles. It has a 1/3 chance of healing 2 brute / burn. This means on average, it will heal 44.5 times during those cycles, for 88 / 88 brute and burn. Cool. But, 133 cycles is 260 seconds, or 4 minutes and 20 seconds. That's about 20 of each damage type per minute, if saline follows the rules.

**_Watcher_**. I kill a watcher, while not getting struck by it, hit by a few beams. I take 22 burn damage total, this would take a minute to heal off following our rules.
Fighting the watcher with KA point blank rewards me with 30 melee damage, with our 60% melee resist. This would take 1.5 minutes to heal

**_Legions_**. Lets fight 2 at once, like if we were taking a legion tendril for the first time, and one of them was far enough away to not aggro, and the one instantly spawned from the tendril also didn't aggro. 21 damage, went fine, but was not hit in hands, and again had 60% armor. Minute to heal if you don't use a core.

**_Goliaths._** We fight a few at a tendril, get hit twice, 35 total damage, as I was hit in the foot. about a minute 40 to heal. 

This might seem reasonable, but this is if you visit a pod every 4 minutes and 20 seconds, otherwise you loose healing. If you attack a watcher faster then once a minute, you will take more damage then you are healing. If you are fighting legions, unless you are fast on picking them up, or have a few cores ready to start the assault on the legion tendril, you will not out-heal. A few bad hits early on will slow you, and since you might not have a legion core, you have no way to heal yourself a lot of damage to stop pain slowdown. You also can eat donks and cactus fruit, but you will eventually get full.

Now, lets look at megafauna

**_Hierophant_** doing my best dodging with KA upgrades on fauna takes 4-5 burn patches, we will say 10 damage is wasted per patch, that's 160-200 burn, saline wont help you in that fight, you will either need 5 cores to just be above crit during the fight, slowdown wont apply because of the cores. Survival pens are nice, and heal quite fast, but those are a decent amount of points, and will heal 25 brute / burn over 10 seconds, while the diluted omnizine will heal 2.5 brute / burn in that time, and 22.5 more over the course of a minute. Optimal, this comes down to 2-3 cores and 1-2 pens on Hierophant, the easiest mega fauna. 

And now the real test. An **_ash drake._** For this fight, upgraded KA with 2 cooldown and a range, maxed miner armor, not tendril, and hireophant staff, cause why not. During that fight, I use 4 pens and 2 cores. I was never bitten, as a large arena was prepared.

As I was writing this, and testing this, I realized that megaufana are fine without patches and nerfed cores, but the real issue is normal fauna, they don't hit too hard, but the damage adds up, and takes  a lot of time to heal. A tendril can do a lot of damage real fast, and saline or a single core or survival pen wont save you. A quick heal of a lot of damage is required, where a heal patch comes in.

In conclusion, the Legion core changes are needed and great, but the survival pods should have a med kit for some basic quick healing, so that tendrils and normal fauna doesn't screw you over early on, spending 3000 mining points on stabilizers and pens is reasonable. This also presumes if every miner is knowing what they are doing... which a lot dont, based on the amount of dead I see.

AdamElTablawy:
alternative to the above,

no

they don't need a medkit

AzuleUtama:
Agreed with all the changes - I'm starting to see miners getting the hiero's club 10 minutes into a round and I would suspect the full heal cores and many healing items inside the pods to contribute towards them speedrunning it.

It's possible to buy brute medkits inside the vendors alongside the survival pens, so there is a means to instantly heal off damage from goliaths and the like without solely using cores - it's just that you don't get more than 2 medkit's worth of healing in a 400 credit capsule anymore.

Qwertytoforty:
Fair, but the issue with the brute medkits is they have the brute trauma kits in them, which you need to stand still for, and target the area that is injured, and is not instant.

TDSSS:
Two questions: How does this affect the luxury pod, if any? IMO it should have a bit more supplies considering the cost
What about the lukewarm donks in the fridge, those heal quite a bit too.

AzuleUtama:
> Fair, but the issue with the brute medkits is they have the brute trauma kits in them, which you need to stand still for, and target the area that is injured, and is not instant.

I'm pretty sure the vendor brute medkits are the same pink ones you see in medical storage - 4 styptic patches, a bandage, a sac-acid pill and the epi-pen.

Fox-McCloud:
The reason  I opted not to do include the vendors, at all, is to give medical something more to do, make miners git even more guud, and make lavaland even harder.

TG has has this for a while, and they get along just fine---and you die sooner there, than here; I'd rather only allow miners an easier way to deal with unique problems to our codebase related to medical+combat (and that's largely broken bones+internal bleeding).

PR #12591, opened 2019-10-15
Stamina Damage Rework

Fox-McCloud:
Throwing this up, mostly for test merge purposes.

TG's stamina damage rework.

The short of it is this:

Stamina damage is kinda problematic from a number of standpoints, and is incredibly frustrating to deal with, from an anti-combat point of view, in general---though the overall idea "damage type that stacks up and stuns upon a certain threshold" is fairly solid.

Changes:
- Stamina stuns no longer take into account total health; it's completely independent of how much damage you have
- Stamina slow no longer stacks with health damage slows; whichever is greater will apply, but never both
- Stamina damage will instantly and fully clear after having not taken any other stamina damage in the last 10 seconds
- You will instantly become unweakened if your stamina damage falls below 100, but before the stamina timer resets
- Stamina damage is capped at 120

Overall, this makes stamina damage significantly less toxic and irritating to play against, while still letting it have the pseudo-damage that eventually weakens element to it.

Other related changes:
- Disabler damage reduced from 36 to 30
- Blob reagent stamina damage is now full, instead of being reduced by 60%

For those of you more familiar with TG: No, I didn't move over stamina damage to be limb based.


JKnutson101:
So the weakened status effect is the equivalent of TG's 'Stamina Crit Paralysis' if I understand this correctly?

Fox-McCloud:
Weakened just means you flop on the floor, drop anything in your hands, and can't move.

Stamina paralysis lasts 10 seconds (weaken duration of 5) provided you aren't taking any other source of stamina damage; if you do, then the weaken duration gets reset.

It's very similar, duration wise, to how it functions now; the big difference is that if you go below 100 stamina damage, you pop up instantly and when the weaken wears off, you'll have 0 stamina damage instead of having to recover, very slowly, forever, over the next ~64 seconds

Denthamos:
Hmn, how will these changes interact with stun/stamina diminishes and removers like adrenals, meth/ect chem wise, and anti-stun implants?  Or is there no change with that interaction from such?

Fox-McCloud:
I've already explained that, twice:

>>You will instantly become unweakened if your stamina damage falls below 100, but before the stamina timer resets

>>It's very similar, duration wise, to how it functions now; the big difference is that if you go below 100 stamina damage, you pop up instantly and when the weaken wears off, you'll have 0 stamina damage instead of having to recover, very slowly, forever, over the next ~64 seconds

PR #12592, opened 2019-10-15
Magnetic gripper code cleanup/re-organization, and bug fix

### LINKED ISSUE #12026: solar tracker electronics can't be applied to a solar assembly as a cyborg.
**Problem Description**:
What is the problem?
by trying to add a solar tracker electronics to a solar assembly through the magnetic gripper nothing happens.

the solar tracker electronics is added successfully when clicking on the assembly with it as an human race with hands, The engineering cyborgs with their magnetic gripper are unable to add the part to the assembly.

**What happened instead**:
the solar tracker electronics should have been applied to the assembly instead of "disappear"

you can't complete a solar tracker as a cyborg, and this is pretty important when you get created in delta station and the human crew orders the cyborg to set up solars.

>acquire solar assembly;
>wrench assembly;
>pick up the solar tracker electronics with the magnetic gripper as a cyborg engineer;
>click on the assembly with the gripper containing the solar tracker electronics;
>the solar tracker electronics suddenly disappears.

You can drop the electronics without any problem after picking it up with the gripper, but this possibility disappears when you click with the gripper on the assembly. starting from now the solar tracking electronics seems lost.

### LINKED ISSUE #12541: Engi borg gripper cannot install power cell into APC
**Issue Description**:
When an engi borg tries to install a power cell into a APC, the power cell will magically disappear from the gripper. Picking up and dropping power cells is fine.

To be able to install power cell and finish the APC

**What happened instead**:
Power cell stolen by redspace

Engiborg is unable to do one of their jobs.

1. Build up an APC until you wire it.
2. Grab power cell with gripper.
3. Use gripper on APC.
4. Summon Power-sie with the 10 bluespace power cell sacrificed.

Should be still working a week ago

**Extra information**:
A mining jetpack I tried to put into the destructive analyser is disappeared too. Installing a shield satellite board into a computer frame works fine.

### LINKED ISSUE #12627: Cyborg magnetic grippers not functioning properly with APCs.
**Issue Description**:

When using a magnetic gripper on a station maintenance drone, I was unable to insert a battery properly. The battery disappeared from the gripper, but the APC I was inserting it into did not update.

Normally, the battery should disappear from the gripper's inventory, and the APC should update both its sprite and in-game menu showing that a battery is now inserted.

**What happened instead**:

The battery disappeared from the gripper, and the APC sprite was unchanged. I assessed and confirmed that I was able to pick up and drop the battery appropriately before attempting to insert it, and that another player (Fatbackwards) encountered the same bug when attempting to do so as a maintenance drone.

This effectively prevents cyborgs and maintenance drones from constructing or repairing APCs without admin intervention.

This original bug was tested on the Assembly Line of Boxstation, west of Engineering. I recommend spawning in as a maintenance drone, grabbing a spare battery from the EngiVend south of the drone fabricator, then heading west to the un-constructed APC frame at the Assembly line. Finish its construction (place on wall, wire, electronics, screwdriver), then try inserting a battery. The APC sprite should be unchanged while the battery disappears from the gripper.

I just noticed it as of submitting this report.

**Extra information**:

### LINKED ISSUE #12871: Power cell gone on insert into APC by borg
**Issue Description**:

The power cells getting deleted when an engi borg tries to place it into an APC.

The power cell end up in the APC and not in bluespace.

**What happened instead**:

The engi borg been recruited by powerfull wizards from the wizard federation and been thought how to destroy matter of a power cell by the ritual of placing it into the APC.

Engi borgs can't finish, repair APCs

Be engi borg. Open APC, take out Power cell with claw thingy, try to put it back... abraka-dabra, no more Power cell. (I tested it two different round before I started this bug report as Kyet requested.)

2020.01.01 07:22 - War round

**Extra information**:

The two places I tested it was the courtroom and Tcomms main chamber.

SteelSlayer:
## What Does This PR Do
This cleans up some of the really old, outdated, or bad code related to the magnetic gripper that maintenance drones and engineering borgs have. Most of the code changes are purely back end stuff that you won't notice in game.

There is one thing I changed that players will notice. I removed the "Drop gripped item" verb from the drone tab in favor of a less clunky and annoying way to drop items. Pressing Q will drop the item you're currently holding in the gripper, if there is any. Otherwise it will unequip the gripper like it did before.

Fixes #12627
Fixes #12541
Fixes #12026
Fixes #12871

## Changelog
:cl:
tweak: The "drop gripped item" verb has been removed in favor of simply pressing Q to drop the gripper's held item.
fix: Fixes a bug in which engineering borgs and maintenance drones could not insert a cell into an APC properly using their magnetic gripper
fix: Magnetic grippers can be used to place solar tracker electronics into a solar assembly again
/:cl:


code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +79,67 @@
...
+				if("Drop")
+					drop_gripped_item()
+				if("Cancel")
+					return FALSE
+		else
+			drop_gripped_item()
+	else
+		to_chat(user, "<span class='warning'>Your gripper is empty.</span>")
 
+/obj/item/gripper/proc/drop_gripped_item(silent = 0)
```
Citinited: There needs to be some check in either here or the `attack_self` to prevent people being able to teleport gripped items. (they can activate the `attack_self()` multiple times)

SteelSlayer: I'm not sure I know what you mean. How can items be teleported?

Citinited: Actually wait never mind, I saw you set `gripped_item` to null

Citinited:
Does this PR fix #12026?

SteelSlayer:
@Citinited It does indeed fix that! I didn't notice that issue, thanks

Citinited:
Yeah, I only saw it because I was looking through the issues section for things that may have been fixed by my own refactor

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +79,67 @@
...
 	if(!silent)
-		to_chat(src.loc, "<span class='warning'>You drop \the [wrapped].</span>")
-	wrapped.forceMove(get_turf(src))
-	wrapped = null
+		to_chat(src.loc, "<span class='warning'>You drop the [gripped_item].</span>")
```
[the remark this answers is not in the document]
farie82: "the" should also not be needed I think

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +79,67 @@
...
-			to_chat(user, "You collect \the [I].")
+		if(is_type_in_typecache(I, can_hold)) // Make sure the item is something the gripper can hold
+			to_chat(user, "You collect the [I].")
 			I.forceMove(src)
-			wrapped = I
-			return
+			gripped_item = I
 		else
-			to_chat(user, "<span class='warning'>Your gripper cannot hold \the [target].</span>")
+			to_chat(user, "<span class='warning'>Your gripper cannot hold the [target].</span>")
```
farie82: The not required

code/modules/mob/living/silicon/robot/drone/drone_items.dm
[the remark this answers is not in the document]
SteelSlayer: Whats annoying is that for some items like airlock electronics, it will say "You drop the airlock electronics" which is fine. But for other items like cells it will say "You drop the the high capacity cell". So in one instance its fine to leave it, but if the item already has a "the" attached to the front, it makes it look ugly.

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +79,67 @@
...
+				if("Configure/place")
+					gripped_item.attack_self(user)
+				if("Drop")
+					drop_gripped_item()
+				if("Cancel")
+					return FALSE
+		else
+			drop_gripped_item()
+	else
+		to_chat(user, "<span class='warning'>Your gripper is empty.</span>")
```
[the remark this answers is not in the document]
SteelSlayer: Makes sense

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +79,67 @@
...
 	if(!target || !proximity) //Target is invalid or we are not adjacent.
-		return
+		return FALSE
 
-	//There's some weirdness with items being lost inside the arm. Trying to fix all cases. ~Z
-	if(!wrapped)
-		for(var/obj/item/thing in src.contents)
-			wrapped = thing
-			break
+	else if(gripped_item) //Already have an item.
```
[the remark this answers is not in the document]
SteelSlayer: Yep

code/modules/mob/living/silicon/robot/drone/drone_items.dm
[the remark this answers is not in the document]
farie82: yeah I got no clue how the game figures that out tbh

code/modules/mob/living/silicon/robot/drone/drone_items.dm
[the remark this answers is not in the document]
SteelSlayer: Ack, I put [src] here but it reads "Your the magnetic gripper is empty". Damn the's ruining everything.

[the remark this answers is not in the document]
Citinited: either remove the Your before [src], or there might be a macro for it (try putting either \improper or \proper before [src])

farie82:
Tested in these scenarios
1. Happy flow (build an APC) working as suspected
2. Putting the gripper in the inventory and back. Item held does not disappear and keeps the gripped object in memory
3. Kill the borg while having an item gripped. Held item is dropped out of the gripper.

I did find Runtimes however when placing the battery in the APC:
image
The item is removed after the attack and is not checked after if it's null or not.

Also a runtime when the borg dies when nobody is in the borg (SSD or ghosted)
image

Cazdon:
Neat to see someone testing someone elses PR, and it turns out to be useful!

SteelSlayer:
Ah I see the issue with the first runtime and know how to fix it.

 The second runtime, I can't seem to replicate. I ghosted out of a borg with an item gripped, then shot the borg to death in another body but did not get the runtime, even though it should be runtiming. I suppose I should add a check to see if the gripper user has a client before the to_chat message.

Thanks for testing.

Cazdon:
> The second runtime, I can't seem to replicate. I ghosted out of a borg with an item gripped, then shot the borg to death


Well the error show he was a maintenance drone, would that make a difference?

SteelSlayer:
It shouldn't matter, because they have the same gripper item. But I tested it just in case, shot a no client maint drone to death, no runtime on the to_chat. However I did discover a new runtime which I've fixed so thats nice.

farie82:
> > The second runtime, I can't seem to replicate. I ghosted out of a borg with an item gripped, then shot the borg to death
> 
> Well the error show he was a maintenance drone, would that make a difference?

The exact scenario was that I hit the borg using a welder till it died while he had a high cap cell gripped

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +82,68 @@
...
+		to_chat(user, "<span class='warning'>[src] is empty.</span>")
 
-	if(!silent)
-		to_chat(src.loc, "<span class='warning'>You drop \the [wrapped].</span>")
-	wrapped.forceMove(get_turf(src))
-	wrapped = null
+/obj/item/gripper/proc/drop_gripped_item(silent = FALSE)
+	if(gripped_item)
+		if(!silent)
+			to_chat(usr, "<span class='warning'>You drop [gripped_item].</span>")
```
SteelSlayer: Changed this from `to_chat(loc` to `to_chat(usr`, because I think using `loc` means it would display the message to all mobs at the location. I'm not sure if using `usr` is the best here. It's simply the easiest solution though, and it works.

farie82: This will mean that attacking a borg to make him drop his gripped item will give the attacker the message since usr will be the attacker. Best to add user as param

SteelSlayer: I tried adding `mob/user` to params and then doing `to_chat(user` but the message would not show up. Which means every call of this this proc needs to pass a variable to use for user. Am I missing something?

farie82: Correct. Other procs using user as param also have the same mechanism. It's because (in this case) you will be the usr if you attack the borg and cause it to drop it's gripped item.
You can probably hack the user together in the proc but that won't be an elegant solution. Best will be to pass the real user as a param

Citinited: I think `loc` would always return the borg, no? Unless it's on the floor, which should never happen

SteelSlayer: Actually yeah, for some reason I though loc was giving turf location but that's not the case.

Citinited: That's probably why it used `loc` in the first place, heh

farie82: Loc in this case was the module seeing the gripper puts it in its contents or something odd

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -71,98 +82,68 @@
...
-		var/grab = 0
-		if(can_hold.len)
-			if(is_type_in_typecache(I, can_hold))
-				grab = 1
-
-		//We can grab the item, finally.
-		if(grab)
-			to_chat(user, "You collect \the [I].")
+		if(is_type_in_typecache(I, can_hold)) // Make sure the item is something the gripper can hold
+			to_chat(user, "You collect [I].")
```
farie82: Doesn't use spans

code/modules/mob/living/silicon/robot/inventory.dm
```diff
@@ -118,8 +118,8 @@
 	var/obj/item/I = get_active_hand()
 	if(istype(I, /obj/item/gripper))
 		var/obj/item/gripper/G = I
-		G.drop_item_p(silent = 1)
-	return
+		G.drop_gripped_item(silent = TRUE)
```
Crazylemon64: What if you were to modify bindings_robot.dm so that the drop key would make the gripper drop its item if it was holding something? I don't like overriding the `attack_self` capabilities, since I feel like that would easily bite us in the butt later if we want to add other attack-selfable things to grippers.

SteelSlayer: Alright, I'll see about doing that.

SteelSlayer:
Well that easy to fix up actually, should be ready to go, provided there are no other issues.

Funce:
There's already native support for dropping with Q in `bindings_mob.dm`. It also includes HOME (Northwest) for non-hotkey mode. (A minority, but it still exists)

Speaking of hotkeys, have you got a non-keyboard way of dropping the item? I don't believe you need the drop-gripped-item verb in the drone tab, but having it as a right click option is better than 'know the hotkeys'.

Funce:
As an amusing aside, the soap (of a drone) can be placed on a conveyor belt, stolen by a human, and now the human has a soap stuck in their hand. Test-merged locally.

code/modules/keybindings/bindings_robot.dm
```diff
@@ -11,8 +11,12 @@
 			return
 		if("Q")
```
Crazylemon64, line 12: Make sure to cover the case for "A" in "AZERTY" as well

PR #12603, opened 2019-10-17
fixes an issue with romerol

Couls:
## What Does This PR Do
In preparation for kasparovy's set_species rework which is now ready for review, this fixes one of the issues with romerol so it can soon(hopefully) be enabled and people and admins can start messing with it(Just in time for halloween too!)

The handle_death proc was removing the infection organ, which is fine on TG where zombies can't actually die, but here our zombies die and revive on a timer.. so we need to not be dezombifying them when they die(since removing the organ removes the infection). Now that we have the trait framework though it might be possible to code in the NO_DEATH traits and stuff but that's out of scope for this one!

## Why It's Good For The Game
Helps us get closer to romerol

## Changelog
:cl:
fix: zombies not properly reviving
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/zombies.dm
+++ b/code/modules/mob/living/carbon/human/species/zombies.dm
@@ -70,14 +70,6 @@
 	if(!H.InCritical() && prob(4))
 		playsound(H, pick(spooks), 50, TRUE, 10)
 
-//Congrats you somehow died so hard you stopped being a zombie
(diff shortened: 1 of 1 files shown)


Fikou:
on tg zombies do die though

Couls:
get out of here @Fikou

taukausanake:
All because you die doesn't mean you stay dead. Look at DC/Marvel

PR #12604, opened 2019-10-17
Fixes endless material exploit

TDSSS:
## What Does This PR Do
Fixes destructive analyzer giving more materials than a thing actually contains when fully upgraded, by hard-capping it so you can get at most 90% of the materials back

## Why It's Good For The Game
no infinite diamond and plasma

## Changelog
:cl:
fix: destructive analyzer endless materials exploit
/:cl:

PR #12608, opened 2019-10-18
Fixes R&D Materials Exploit

Fox-McCloud:
Alternative to: https://github.com/ParadiseSS13/Paradise/pull/12604

Fixes R&D being able to exploit infinite materials.

Also shame on you guys for actively exploiting this and not reporting it.

It only further gives credence to my thesis that a decent amount of players only report bugs when it *negatively* impacts them, but are blithely happy to conveniently "forget" to report exploits/bugs in their favor.

:cl: Fox McCloud
fix: Fixes being able to exploit the destructive analyzer for unlimited materials
/:cl:

### DIFF
--- a/code/modules/research/rdconsole.dm
+++ b/code/modules/research/rdconsole.dm
@@ -357,7 +357,8 @@ won't update every console in existence) but it's more of a hassle to do. Also, 
 							submenu = 0
 						if(linked_lathe) //Also sends salvaged materials to a linked protolathe, if any.
 							for(var/material in linked_destroy.loaded_item.materials)
-								linked_lathe.materials.insert_amount(min((linked_lathe.materials.max_amount - linked_lathe.materials.total_amount), (linked_destroy.loaded_item.materials[material]*(linked_destroy.decon_mod/10))), material)
+								var/can_insert = min((linked_lathe.materials.max_amount - linked_lathe.materials.total_amount), (min(linked_destroy.loaded_item.materials[material] * (linked_destroy.decon_mod / 10), linked_destroy.loaded_item.materials[material])))
+								linked_lathe.materials.insert_amount(can_insert, material)
 						linked_destroy.loaded_item = null
 					else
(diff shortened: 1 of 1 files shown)


Shockpoint:
What's the difference here to #12604?

Fox-McCloud:
They hardcapped the material amount to 9 (I dunno why 10) to fix the issue. This technically has balance implications, and while it solves the issue, the problem is really due to there not being a `min` function call that prevents the materials from going over what they original were.

There's not really an issue with 100% efficient reclamation, so long as you can't go over it.

Items produced in R&D that are cheaper have reduced materials in them, so that's already safeguarded for.

code/modules/research/rdconsole.dm
```diff
@@ -357,7 +357,8 @@ won't update every console in existence) but it's more of a hassle to do. Also,
 							submenu = 0
 						if(linked_lathe) //Also sends salvaged materials to a linked protolathe, if any.
 							for(var/material in linked_destroy.loaded_item.materials)
-								linked_lathe.materials.insert_amount(min((linked_lathe.materials.max_amount - linked_lathe.materials.total_amount), (linked_destroy.loaded_item.materials[material]*(linked_destroy.decon_mod/10))), material)
+								var/can_insert = min((linked_lathe.materials.max_amount - linked_lathe.materials.total_amount), (min(linked_destroy.loaded_item.materials[material] * (linked_destroy.decon_mod / 10), linked_destroy.loaded_item.materials[material])))
```
TDSSS: Couldn't you do this with just one min, instead of two? IE min(A, B, C) instead of min(A, min(B, C))? I looked it up and min does take more than 2 arguments.

Crazylemon64: That checks out to me, yeah

MrMagolor:
When people don't report bugs outside of discord/adminhelp and say that "Oh, I reported it"... 😠

PR #12616, opened 2019-10-20
Adds a "Hear/Silence AI Voice Announcements" verb to the preferences tab

SteelSlayer:
## What Does This PR Do
This adds a verb called "Hear/Silence AI Voice Announcements" to the preferences tab. If you have this on, it converts whatever you would normally hear from the AI over a voice announcement and instead displays it to your chat window. It's basically identical to the text announcement AI's can make, complete with the little ding notification noise.

## Why It's Good For The Game
Sometimes AI's like to spam the voice announcement. It can be funny sometimes, but other times, it can be annoying. This option allows you to silence it but still receive whatever information the AI is trying to say.

Lets see what our AI thinks about this...

AI

Well there you have it, need I say more?

## Changelog
:cl:
add: Adds the Hear/Silence AI announcements verb to the preferences tab. If toggled on, it will convert the AI's vocal announcements to text instead
/:cl:


AffectedArc07:
May be worth running
```sql
UPDATE players SET sound = sound + 1
```
This means that everyone will still hear AI things by default 

*This query was written from memory and not compared with the existing DB table*

code/modules/client/preference/preferences_toggles.dm
```diff
@@ -36,6 +36,16 @@
 	to_chat(usr, "You will [(prefs.toggles & CHAT_RADIO) ? "now" : "no longer"] see radio chatter from radios or speakers")
 	feedback_add_details("admin_verb","THR") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
+/client/verb/toggle_ai_voice_annoucements()
+	set name = "Hear/Silence AI Voice Announcements"
+	set category = "Preferences"
+	set desc = "Toggle hearing AI annoucements in voice form or in text form"
+	if(!holder)	return
```
AffectedArc07: `holder` is a check for admin status (iirc). I would check if this works when you’re not an admin.

Ty-Omaha: holder is anyone with a rank including mentors

AffectedArc07:
I would say dont forget to update
https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/client/preference/preferences_mysql.dm#L65

But we are nowhere near 65535 yet

code/modules/client/preference/preferences_toggles.dm
```diff
@@ -36,6 +36,16 @@
...
+	prefs.sound ^= SOUND_AI_VOICE
+	prefs.save_preferences(src)
+	to_chat(usr, "[(prefs.sound & SOUND_AI_VOICE) ? "You will now hear AI announcements." : "AI annoucements will now be converted to text."] ")
+	feedback_add_details("admin_verb","TAIvoice") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
```
Ty-Omaha: feedback_add_details isn't needed because it's for the old blackbox system

AffectedArc07: @Ty-Omaha The new blackbox system isnt in yet, so it is still required

code/modules/mob/living/silicon/ai/say.dm
```diff
@@ -134,6 +134,22 @@ var/const/VOX_PATH = "sound/vox_fem/"
...
+	var/formatted_message = "<h1 class='alert'>A.I. Announcement</h1>"
+	formatted_message += "<br><span class='alert'>[words_string]</span>"
+	formatted_message += "<br><span class='alert'> -[AI]</span>"
+
+	for(var/mob/M in GLOB.player_list)
+		if(M.client && !(M.client.prefs.sound & SOUND_AI_VOICE))
+			var/turf/T = get_turf(M)
+			if(T && T.z == AI.z && M.can_hear())
+				M << 'sound/misc/notice2.ogg'
+				to_chat(M, formatted_message)
```
Ty-Omaha: Instead of creating a new announcement proc you should just use the existing proc and send the words to that

code/modules/client/preference/preferences_toggles.dm
Ty-Omaha: @AffectedArc07 The blackbox isnt even used, we don't need a new one. Look at all the new preference toggles without it.

code/modules/mob/living/silicon/ai/say.dm
Ty-Omaha: oh but then you have the issue of people who CAN hear. I see. Maybe a var?

Ty-Omaha:
The conversion to an announcement if they dont have voice enabled is really smart

code/modules/mob/living/silicon/ai/say.dm
Ty-Omaha: oh you could just throw it in the for loop

SteelSlayer: Throw what in the for loop?

code/modules/client/preference/preferences_toggles.dm
SteelSlayer: Yeah, I'll remove that bit

code/modules/mob/living/silicon/ai/say.dm
Ty-Omaha: the announcement proc but it probably doesnt work that way so nevermind

SteelSlayer:
@AffectedArc07 The preference should be set to hear AI voice by default. But where should I put the
`UPDATE players SET sound = sound + 1` ?

AffectedArc07:
@SteelSlayer New players (who havnt joined the server before) will hear the AI things through their speakers, however existing players wont due to their defaults not being updated.

In terms of where that query goes, it goes into the SQL console if @Kyep feels its required

code/modules/client/preference/preferences_toggles.dm
SteelSlayer: It does look like *some* of the preference verbs aren't using the `feedback_add_details` line, so I suppose it could work without it. But I'm not familiar with this system, and if it matters that I remove it.

Couls:
Might be out of scope but can you add a third “both” option for people who want to see the announcement as text and hear it? Sometimes they can be hard to understand but I don’t mind hearing them

SteelSlayer:
@Couls I'm not sure I can. All of the other preference verbs are either on or off, but it might be possible.

Couls:
might have to split it up into two seperate preferences

code/modules/mob/living/silicon/ai/say.dm
```diff
@@ -135,6 +135,22 @@ var/const/VOX_PATH = "sound/vox_fem/"
...
+
+
+/proc/ai_voice_announcement_to_text(words, ai)
+	var/mob/living/silicon/ai/AI = ai
+	var/words_string = jointext(words, " ")
+	var/formatted_message = "<h1 class='alert'>A.I. Announcement</h1>"
+	formatted_message += "<br><span class='alert'>[words_string]</span>"
+	formatted_message += "<br><span class='alert'> -[AI]</span>"
+
+	for(var/mob/M in GLOB.player_list)
```
farie82: Can use `var/thingy` and then cast it to a mob in the loop seeing the list should only ever contain mobs.

code/modules/mob/living/silicon/ai/say.dm
```diff
@@ -135,6 +135,22 @@ var/const/VOX_PATH = "sound/vox_fem/"
 	for(var/word in words)
 		play_vox_word(word, src.z, null)
 
+	ai_voice_announcement_to_text(words, src)
+
+
+/proc/ai_voice_announcement_to_text(words, ai)
```
farie82: Make this a proc on the AI level. It'll only be used by the AI anyway since it requires an AI as param

farie82: Optional (might be out of the scope of the PR):
You can probably use the `datum/announcement` object for this. It'll handle all things in here except for the z level thing. 
However it would require (quite) some refactoring of that system.

code/modules/mob/living/silicon/ai/say.dm
```diff
@@ -135,6 +135,22 @@ var/const/VOX_PATH = "sound/vox_fem/"
...
+	var/words_string = jointext(words, " ")
+	var/formatted_message = "<h1 class='alert'>A.I. Announcement</h1>"
+	formatted_message += "<br><span class='alert'>[words_string]</span>"
+	formatted_message += "<br><span class='alert'> -[AI]</span>"
+
+	for(var/mob/M in GLOB.player_list)
+		if(M.client && !(M.client.prefs.sound & SOUND_AI_VOICE))
+			var/turf/T = get_turf(M)
+			if(T && T.z == AI.z && M.can_hear())
+				M << 'sound/misc/notice2.ogg'
```
farie82: SEND_SOUND is a bit more readable here

code/modules/mob/living/silicon/ai/say.dm
```diff
@@ -135,6 +135,22 @@ var/const/VOX_PATH = "sound/vox_fem/"
 	for(var/word in words)
 		play_vox_word(word, src.z, null)
 
+	ai_voice_announcement_to_text(words, src)
+
+
+/proc/ai_voice_announcement_to_text(words, ai)
+	var/mob/living/silicon/ai/AI = ai
+	var/words_string = jointext(words, " ")
```
farie82: The words here can be any case possible. Just to keep that in mind.

code/modules/mob/living/silicon/ai/say.dm
SteelSlayer: good idea

SteelSlayer: Yeeeep. I remember trying to use that I think. I don't want to touch that right now :smile:

code/modules/mob/living/silicon/ai/say.dm
SteelSlayer: It should be fine since AI voice announcement words are restricted to either Title Case or UPPERCASE

code/modules/mob/living/silicon/ai/say.dm
SteelSlayer: If you know they should all be mobs, wouldn't this be more efficient, with not having to cast it? Or can it lead to some fucky behavior?

farie82: I believe if you specify a type in the for loop that it will filter out the other types. A filter costs time linear to the list length. So if you already know it only has only one type of variable (or it's subtypes) then you can better use `var/thingy` and cast it to it's actual type in the loop itself

SteelSlayer: Fair enough

PR #12628, opened 2019-10-22
IPC balance changes

SteelSlayer:
## What Does This PR Do
Alternative to: #13440

Disclaimer: before you continue reading this PR, I need to say a few things. First, the changes and numbers in this PR might not be final and are subject to change by the wishes of headmins and maints.

Secondly while reading this, do keep in mind that IPCs are still using old crit. This means they have only **100 usable HP** before they fall over "unconscious" in crit and are helpless. Not to mention the 150% damage modifier for brute and burn which makes them die even quicker.

Compared to all other species, which are on new crit now, IPCs are extremely weak, and I feel it's time to give them *something* instead of just ignoring their situation. Please, before you respond with a kneejerk "No" on this, hear me out.

With that out of the way, let's get onto the changes:

- IPCs no longer have oil blood anymore:

Originally, I was going to make it so the lower oil they had, the slower they would move. However I could not find a good way to do this without using the stamina blood damage they currently use. But, if anyone has any suggestions on how to simply make low blood slow them without being too snowflake, I'd probably change it to that system.

- IPCs now take a flat amount of damage when they are affected by an EMP:

There are two types of EMP severity, level 1 and level 2. Severity 1s are always closer to the middle of EMPs. For example if you use an EMP implant, the first 4 tiles around the EMP are severity 1 and any more than that, are severity 2.

**Severity 1 EMPs will do ~5.28 damage per limb for a total of ~65 damage if the IPC has 11 limbs. Severity 2 EMPs will do ~3.63 damage per limb for a total of ~40 damage if the IPC has 11 limbs**.

This gives IPC's at least a chance to maybe fight back instead of taking 300 damage, losing all their limbs at once and turning into a nugget.

~~- Lowered the brute and burn damage modifier on IPCs to 120%, down from 150%~~

~~Remember how I said IPC only have 100 usable HP? Well, that got even worse because they have a 150% damage taken modifier. This attempts to lessen that without completely removing that weakness. It should give them a bit more survivability.~~

Keeping the damage modifier the same at 150%.

Well, these are my purposed balance changes, and they may not be perfect. However I really just wanted to make this PR to get the discussion going, because I think it's time we did something about IPC balance.

## Changelog
:cl:
tweak: Robotics lockers no longer have oil cans in them
balance: IPCs no longer bleed oil
balance: Changes the EMP damage IPCs receieve. Severity 1 EMPs will do ~5.28 damage per limb for a total of ~65 damage if the IPC has 11 limbs. Severity 2 EMPs will do ~3.63 damage per limb for a total of ~40 damage.
fix: Fixed humans and carbon mobs getting all of their body parts EMP'd twice from a single EMP.
/:cl:


TDSSS:
I admit, I never got why IPCs had to have oil, so 👍 on that one. Three buffs at once might be a bid much tho.

JohnBigless:
Can't wait for this PR to inevitably get bashed into the ground because no one likes IPC.

farie82:
The damage debuff can be a bit tweaked I think. But it should be lower than 150% atleast. Good shit

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -162,10 +162,21 @@ emp_act
 		return TRUE
 
 /mob/living/carbon/human/emp_act(severity)
-	for(var/obj/O in src)
-		if(!O)	continue
+	if(isSynthetic())
```
TDSSS: Overriding the emp_act on the IPC level is a much better way do to this than using an istype check, IMO

Couls: IPCs aren't a subtype they're a species

SteelSlayer:
Travis is being weird again. Says its building but if you check the details, the build has passed.

code/modules/mob/living/carbon/human/human_defense.dm
farie82: Issue with that is that it'll still do the rest below it. And you don't want that to happen this time.
Issue I see here with the PR code is that it does not go through the object list anymore. Does it still EMP the objects the IPC holds? Like the backpack stuff etc?

TDSSS: Right, I forgot how exactly species stuff is set up. Still, the current approach seems not very OOP.

SteelSlayer: Ack, you're right, I screwed up my logic, and forgot to include EMPing the IPC's contents

JKnutson101:
After over 600 hours as IPC, I'd like to think my opinion has some merit, so here it goes.

I'll start with the nerf to EMPs, which I think most people will agree on is welcomed. Much like the reason removal of EMP weakness for Vox, EMPs on IPCs is perhaps one of the _least_ challenging, interesting, and fun ways to kill anyone. There's no chance to resist or flee, and doesn't just discourage roleplay situations, but doesn't even allow a chance for them. EMPs are also incredibly cheap in terms of cost to antags, with the EMP Flashlight even being its own meme. For just two TC you can kill your target instantly with no chance to escape or scream, and it doesn't even look like a real weapon! However, even with the changes in this PR, the above is not going to change.

What is going to change about EMPs is the collateral damage. While getting EMP'd as a target is lame and boring, nothing is as frustrating as getting EMP'd when you were not the target. Many EMPs have large AoEs to hit many targets. Revenant's EMP, Cultist's EMP, and EMP Grenades all cover a huge range and often spread to nearby rooms, beyond walls. Dying instantly to something not even intended for you, an entire room over is perhaps one of the most frustrating experiences I've ever had, and I've had it multiple times. The worst example of this I can remember was getting EMP'd by fellow cultists that weren't thinking and wanted to 'kill the lights' just before the final summoning ritual, and then failing to have enough members to finish the ritual. The above changes WILL fix this, because if the IPC isn't the target of the EMP, then they'll take a bunch of damage, but be in a position where they can at least heal themselves or communicate that they need help. It will also let cultists use EMP Talismans a bit more freely around IPC cultists because now even if they take a ton of damage at least they're not completely dead.


I would rant a lot more about the other changes but this went on long enough, I'll just say I like 'em and leave it for now.


Did you test if the EMP damage scaled with the 120% burn damage? I don't want that 90 damage to be 108 or else the problems remain.

SteelSlayer:
@JKnutson101 Yes I did, the `take_overall_damage()` proc I used for this ignores species damage modifiers. I added a comment in the code saying so as well

code/modules/mob/living/carbon/human/human_defense.dm
Fox-McCloud: This is exactly the kind of snowflake code that makes a lot of coders despise IPCs to begin with; this is a no go how this is being handled.

Fox-McCloud:
This PR is a prime example why a lot of coders end up hating how IPCs are handled. To alter the behavior of EMPs, no less than 4 files has to be changed for a minor change.

I'm not keen on touching EMP.

Altering how much damage they take would be acceptable, but the other changes are questionable.

code/modules/mob/living/carbon/human/human_defense.dm
Couls: So uhh.. how should it be handled?

SteelSlayer:
@Fox-McCloud We'll see.. I *assumed* humans getting their entire contents EMP'd multiple times was not an intended feature, because, why would that happen? So I simply made the `carbon/emp_act()` proc and the `human/emp_act()` proc EMP a mobs contents and then internal organs and then removed the bit that calls the parent, since that is unnecessary.

SteelSlayer:
@Kyep I can live with removing the damage mod changes. The real important change is getting rid of the EMP instakill like you said. 

I don't really know why the number of files should have any affect on whether the PR should go through or not. The file changes are necessary to revert the oil blood mechanic IPCs have; removing oil cans, removing oil tanks etc. Basically I'm undoing all the changes in the PR from a year ago in which oil was added to IPCs.

Spacemanspark:
I would rather IPCs kept their current damage modifiers. It makes for a nice challenge.

The rest of the changes I also agree with. Despite what I said above, I've always felt that instant death is a bit much.

Oil I don't care about either way. I've grown used to it, but it doesn't matter to me one way or the other.

SteelSlayer:
Alright, I'm removing the changes to the damage modifiers, IPC will now still take 150% brute and burn damage. This is a necessary change if I want this PR to have a chance to pass.

thanatossage:
kinda late to this but god i would love for this also i kinda like the oil being gone since it is a pain to deal with some times as a robotist when you have a pile of ipcs that need to be fixed and the sole one and you have to sit there and give the ipc thats been dead for 20 minute enough oil so they can at least stand to drink it themselfs. also insta kills being gone is great like what most people said insta kill just insta fun. like having the rev just appear behind you and wipe you and like 2 other ipcs out is just an oh ok moment. its part of the reason why i stopped playing my ipc to begin with which is sad cause i love playing them. please let this thing pass.

code/modules/mob/living/carbon/human/human_defense.dm
farie82: Maybe have the species listen to the emp_act signal and act accordingly (signal prob needs to be made)

AffectedArc07:
As a 1600 hour IPC player, I think these changes are fine, since making IPCs newcrit compatible would be aids

Cazdon:
> As a 1600 hour IPC player, I think these changes are fine, since making IPCs newcrit compatible would be aids

Mmmm delicious robot air. I can't live without it. Otherwise my robotic internal wind-power generators cannot produce enough power to sustain my wind-powered brain.

Spacemanspark:
Turn Torque into a wind turbine

Asmerath:
I love these changes, might actually get to play the game and stop being a human but worse.

SteelSlayer:
Alright, I've updated this a bit to include less snowflake code. External limbs now have a `emp_shielded` var and IPC limbs have this set to true. This only gives them reduced EMP damage while being separate from the `tough` var that augmented limbs have.

As for the new numbers: 
1. Severity 1 EMPs will do ~5.28 damage per limb for a total of ~65 damage if the IPC has 11 limbs.
2. Severity 2 EMPs will do ~3.63 damage per limb for a total of ~40 damage if the IPC has 11 limbs.

Yes, these numbers are lower which is an even further buff, but keep in mind, IPCs fall unconcious at -1 HP and will be unable to fight back. you can still two-shot IPCs with severity 1 EMPs like the EMP flashlight or the implant (if you're close enough), or ion rifle shots etc. And severity 2's are three-shots for the IPC.

Also cleaned up some organization by moving `../organ/external/emp_act()` to the external limb file, and the `../organ/internal/emp_act()` to the internal organ file, instead of having them both in `organs.dm`

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -298,14 +298,16 @@
...
-	if(!isSynthetic() && blood_volume < BLOOD_VOLUME_SAFE)
+	if(blood_volume < BLOOD_VOLUME_SAFE)
 		msg += "[p_they(TRUE)] [p_have()] pale skin.\n"
 
 	if(bleedsuppress)
 		msg += "[p_they(TRUE)] [p_are()] bandaged with something.\n"
 	else if(bleed_rate)
-		var/bleed_message = !isSynthetic() ? "bleeding" : "leaking"
-		msg += "<B>[p_they(TRUE)] [p_are()] [bleed_message]!</B>\n"
+		if(reagents.has_reagent("heparin"))
```
farie82: This was removed in the recent heparin PR

SteelSlayer: Ah ok, I'll remove it

Spacemanspark:
Glorious

Merge naow

MRgaty:
Oh god yes

Spacemanspark:
I'll be writing up something very soon (tm) regarding the test merge of this PR and how it played out.
Feel free to PM me on discord to remind me if I don't get around to this by Friday. Long work hours at bad times right now, but I'll do my best.
@SteelSlayer

Spacemanspark:
I have been pinged, and will now write up my report.

Overall, there's a lot that remains the same. IPCs are still fairly tin foil and easily ripped apart in brute combat.

The EMP buff does, however, offer synthetics the chance to potentially react to threats, but you're most certainly still going to get destroyed if you're taken off guard and don't have support. IE, it's not complete one shot bullshit, but it's enough that it's still a vastly significant blow to any synthetic. I believe about three hits still kills or almost kills an IPC. This doesn't include limbloss and how that plays into it. 
The removal of oil isn't so much of an immediate thing during combat as it is a long term thing that will effect you after combat.  I think it would be best to have another test merge with some more ordinary rounds to see the full impacts on it, but as is I think it's removal is a good change.

Let me know if any other information is needed and I'll update this, somewhat still waking up.

Dumbdumn5:
Asked personally and on behalf of the heads for this to be re-opened. The existence of an alternative PR from a maintainer shouldn't mean the loss of something like this. With its extensive history, its continual test-merges, and its constant improvement towards something the community can agree on.

Twelve different files changed across 7 months with significant line changes and additions vs. a two-line change to pre-existing code with barely the same amount of attention to community opinion and detail isn't even a contest in my book.

This one's got my pick over the latter for sure.

:+1:

Spacemanspark:
Any update on this as far as heads/ maints deciding on it goes?

SteelSlayer:
So far that I know of:

Objections:
- Fox

Approvals:
- Kyet
- Dumbdumn

Undecided/Haven't commented:
- Ansari
- Crazylemon

lolzyking:
I kind of feel pr's like this should be voted on by heads only, maintainers should only be able to object a Pr based on the merit of its code, not its content.

Tokorizo:
IPCs are still terrible and ive never played an IPC. This got massive community praise but it seems to be ignored for almost a year now? Really sad that IPC players have to deal with this.

_maps/map_files/cyberiad/cyberiad.dmm
```diff
@@ -46238,7 +46238,6 @@
 	dir = 2;
 	network = list("Research","SS13")
 	},
-/obj/structure/reagent_dispensers/oil,
```
AffectedArc07, line 46241: *hopefully* this can be merged after the mapping queue is dealt with. I hope so atleast.

Spacemanspark:
Hi yes this is the galactic robotic foundation and we want our PR merged

If our demands aren't met we will burn an NT flag in front of your house

It will smell really bad

taukausanake:
Threaten to bleed oil everywhere and make a mess too if not merged. Use that feature for the cause

_maps/map_files/cyberiad/cyberiad.dmm
SteelSlayer: Oh right, map changes. RIP this PR for a while then.

TDSSS: If only there was another use for oil so the oil tanks could stay in...

taukausanake: Well cyborgs could probably use it seeing that they actually have oil spatter out of them when they are brute damaged. That could be another form of repair and/or maintenance for them, refilling oil. Less oil means slower operation and movement

Darkmight9: have robotics hand the tank to chemistry *shrugs

Asmerath:
I mean IPCs leave more blood around for slaughters than regular humans imo. Ever pulled a dying IPC from arrivals to robotics? talk about coating the station...

SteelSlayer:
I have now removed the map edits from this PR which means it can be merged, should the maintainers choose to. The consequence of this is that there will be a useless oil tank sitting in robotics for the time being.

Spacemanspark:
Merge when

AffectedArc07 (approved):

AffectedArc07:
This has 2 heads (and me) in favour, and an overwhelming response on the PR
image

I am feeling more and more inclined to merge this, especially since it has been around since october 2019.

Darkmight9:
@AffectedArc07 Please please merge it AA! The map edits are gone so it won't conflict with mapping standardization

Spacemanspark:
Please do

TDSSS:
If you removed the map changes, then your changelog needs updating

SteelSlayer:
You're right. Done

variableundefined:
Second approval from a maint for this PR

Spacemanspark:
HERE WE GOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO

Tokorizo:
It only took almost a year for a merge, incredible.

JohnBigless:
A buff for IPC? On Paradise? I must be dreaming.

PR #12630, opened 2019-10-22
Added the PASS alarm, a device that plays a very loud noise whenever it hasn't been moved for a set amount of time

SkullyRoberts:
## What Does This PR Do
It adds the PASS alarm. A special device that, when activated, continuously plays a very loud sound whenever it hasn't been moved for a set time period.

## Why It's Good For The Game
I'm not honestly sure what anyone would use this for. I'm sure an admin or someone will come up with a use eventually

## Changelog
:cl:RobinFox
add: The PASS alarm
/:cl:

### DIFF
--- a/code/modules/assembly/pass.dm
+++ b/code/modules/assembly/pass.dm
@@ -1,0 +1,48 @@
+var/list/PASS_list = list()
+
+/obj/item/assembly/timer/pass
+	name = "Personal Alert Safety System"
+	desc = "A device that plays a very loud alarm if it isn't moved for 30 seconds."
+	icon_state = "timer"
+	var/lastloc = null
+	var/loca = null
+	var/alarm = 0
+	var/list/pass_sounds = list('sound/effects/alert.ogg' = 1)
+
+
+
+/obj/item/assembly/timer/pass/Initialize
+	. = ..()
+	AddComponent(/datum/component/squeak, pass_sounds, 100)
+
+/obj/item/assembly/timer/pass/New()
+	PASS_list.Add(src)
+
+/obj/item/assembly/timer/pass/Destroy()
+	PASS_list.Remove(src)
+	return ..()
+
+/obj/item/assembly/timer/pass/describe()
+	if(timing)
+		return "The PASS will go off in [time] seconds if not moved!"
+	return "The timer is set for [time] seconds."
+
+
+/obj/item/assembly/timer/pass/process()
+	if(alarm == 1)
+		playsound(loca,'sound/effects/alert.ogg', 150, 1)
+	if(timing && (time > 0))
+		loca = get_turf(src)
+		if(loca == lastloc)
+			time -= 2 // 2 seconds per process()
+		else
+			time = set_time
+			alarm = 0
+		lastloc = loca
+
+	if(timing && time <= 0)
+		timing = repeat
+		timer_end()
+		alarm = 1
+		time = set_time
+
--- a/code/modules/assembly/pass.dm
+++ b/code/modules/assembly/pass.dm
@@ -15,12 +15,6 @@ var/list/PASS_list = list()
 	. = ..()
 	AddComponent(/datum/component/squeak, pass_sounds, 100)
 
-/obj/item/assembly/timer/pass/New()
-	PASS_list.Add(src)
-
-/obj/item/assembly/timer/pass/Destroy()
-	PASS_list.Remove(src)
-	return ..()
 
 /obj/item/assembly/timer/pass/describe()
 	if(timing)
--- a/code/modules/assembly/pass.dm
+++ b/code/modules/assembly/pass.dm
@@ -1,5 +1,3 @@
-var/list/PASS_list = list()
-
 /obj/item/assembly/timer/pass
 	name = "Personal Alert Safety System"
 	desc = "A device that plays a very loud alarm if it isn't moved for 30 seconds."
--- a/code/modules/assembly/pass.dm
+++ b/code/modules/assembly/pass.dm
@@ -11,7 +11,7 @@ var/list/PASS_list = list()
 
 
 
-/obj/item/assembly/timer/pass/Initialize
+/obj/item/assembly/timer/pass/Initialize()
 	. = ..()
 	AddComponent(/datum/component/squeak, pass_sounds, 100)
 
@@ -33,8 +33,6 @@ var/list/PASS_list = list()
 			time = set_time
 			alarm = 0
 		lastloc = loca
-
 	if(timing && time <= 0)
 		alarm = 1
 		time = set_time
-
--- a/code/modules/assembly/pass.dm
+++ b/code/modules/assembly/pass.dm
@@ -16,10 +16,10 @@ var/list/PASS_list = list()
 	AddComponent(/datum/component/squeak, pass_sounds, 100)
 
 
-/obj/item/assembly/timer/pass/describe()
-	if(timing)
-		return "The PASS will go off in [time] seconds if not moved!"
-	return "The timer is set for [time] seconds."
+///obj/item/assembly/timer/pass/describe()
+//	if(timing)
+//		return "The PASS will go off in [time] seconds if not moved!"
+//	return "The timer is set for [time] seconds."
 
 
 /obj/item/assembly/timer/pass/process()
--- a/code/modules/assembly/pass.dm
+++ b/code/modules/assembly/pass.dm
@@ -35,8 +35,6 @@ var/list/PASS_list = list()
 		lastloc = loca
 
 	if(timing && time <= 0)
-		timing = repeat
-		timer_end()
 		alarm = 1
 		time = set_time
 
--- a/paradise.dme
+++ b/paradise.dme
@@ -1261,6 +1261,7 @@
 #include "code\modules\assembly\igniter.dm"
 #include "code\modules\assembly\infrared.dm"
 #include "code\modules\assembly\mousetrap.dm"
+#include "code\modules\assembly\pass.dm"
 #include "code\modules\assembly\proximity.dm"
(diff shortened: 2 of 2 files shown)


datlo:
Can this be attached to an igniter and used to make a LOUD 30 seconds bomb? :thinking:

TDSSS:
But...why?

Couls:
admin annoys 160 people by hiding 30 pass alarms in a locker near bridge to all go off at the same time

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+	var/list/pass_sounds = list('sound/effects/alert.ogg' = 1)
+
+
+
+/obj/item/assembly/timer/pass/Initialize
+	. = ..()
+	AddComponent(/datum/component/squeak, pass_sounds, 100)
+
+
+/obj/item/assembly/timer/pass/describe()
```
Couls: This is not a proc here, you want examine

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+	desc = "A device that plays a very loud alarm if it isn't moved for 30 seconds."
+	icon_state = "timer"
+	var/lastloc = null
+	var/loca = null
+	var/alarm = 0
+	var/list/pass_sounds = list('sound/effects/alert.ogg' = 1)
+
+
+
+/obj/item/assembly/timer/pass/Initialize
```
Couls: Initialize(mapload) you can't define a proc without ()

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
+var/list/PASS_list = list()
```
Couls: does this list actually ever get used? I don't see it being added to or removed from

SkullyRoberts: It was used originally when I had them have built in GPS. Lemme remove it.

SkullyRoberts:
In case you can't tell, I haven't coded anything in byond for a while. XP

Citinited:
There's no way to get this in-game.

I'm also dubious about an item that screams "MY OWNER IS DEAD REE" 

Also grief applications

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+
+
+/obj/item/assembly/timer/pass/describe()
+	if(timing)
+		return "The PASS will go off in [time] seconds if not moved!"
+	return "The timer is set for [time] seconds."
+
+
+/obj/item/assembly/timer/pass/process()
+	if(alarm == 1)
```
Ty-Omaha: `if(alarm)`

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
+var/list/PASS_list = list()
+
+/obj/item/assembly/timer/pass
+	name = "Personal Alert Safety System"
+	desc = "A device that plays a very loud alarm if it isn't moved for 30 seconds."
+	icon_state = "timer"
+	var/lastloc = null
+	var/loca = null
+	var/alarm = 0
```
Ty-Omaha: Use FALSE and TRUE for booleans

SkullyRoberts:
@Citinited there's no way to actually make this as a crew member. it's an admeme only item

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+
+/obj/item/assembly/timer/pass/describe()
+	if(timing)
+		return "The PASS will go off in [time] seconds if not moved!"
+	return "The timer is set for [time] seconds."
+
+
+/obj/item/assembly/timer/pass/process()
+	if(alarm == 1)
+		playsound(loca,'sound/effects/alert.ogg', 150, 1)
```
Ty-Omaha: This will play a sound every second but the sound is 2 seconds long. I also don't know how this will affect performance if at all

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+		loca = get_turf(src)
+		if(loca == lastloc)
+			time -= 2 // 2 seconds per process()
+		else
+			time = set_time
+			alarm = 0
+		lastloc = loca
+
+	if(timing && time <= 0)
+		alarm = 1
```
Ty-Omaha: TRUE

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+/obj/item/assembly/timer/pass/process()
+	if(alarm == 1)
+		playsound(loca,'sound/effects/alert.ogg', 150, 1)
+	if(timing && (time > 0))
+		loca = get_turf(src)
+		if(loca == lastloc)
+			time -= 2 // 2 seconds per process()
+		else
+			time = set_time
+			alarm = 0
```
Ty-Omaha: FALSE

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,36 @@
...
+
+/obj/item/assembly/timer/pass/Initialize()
+	. = ..()
+	AddComponent(/datum/component/squeak, pass_sounds, 100)
+
+
+///obj/item/assembly/timer/pass/describe()
+//	if(timing)
+//		return "The PASS will go off in [time] seconds if not moved!"
+//	return "The timer is set for [time] seconds."
```
Ty-Omaha: Don't add commented out code

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,36 @@
+/obj/item/assembly/timer/pass
+	name = "Personal Alert Safety System"
+	desc = "A device that plays a very loud alarm if it isn't moved for 30 seconds."
+	icon_state = "timer"
+	var/lastloc = null
+	var/loca = null
```
Ty-Omaha: Renaming this to current_turf or something would be easier for other people to comprehend

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
+var/list/PASS_list = list()
+
+/obj/item/assembly/timer/pass
+	name = "Personal Alert Safety System"
+	desc = "A device that plays a very loud alarm if it isn't moved for 30 seconds."
+	icon_state = "timer"
+	var/lastloc = null
```
Ty-Omaha: last_turf would be better because it's using the turf not the loc, and loc frequently equals 0 such as when in lockers

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,40 @@
...
+
+
+
+/obj/item/assembly/timer/pass/Initialize
+	. = ..()
+	AddComponent(/datum/component/squeak, pass_sounds, 100)
+
+
+/obj/item/assembly/timer/pass/describe()
+	if(timing)
```
Ty-Omaha: Where does this even get set?

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,36 @@
...
+///obj/item/assembly/timer/pass/describe()
+//	if(timing)
+//		return "The PASS will go off in [time] seconds if not moved!"
+//	return "The timer is set for [time] seconds."
+
+
+/obj/item/assembly/timer/pass/process()
+	if(alarm == 1)
+		playsound(loca,'sound/effects/alert.ogg', 150, 1)
+	if(timing && (time > 0))
```
Ty-Omaha: Instead of checking for timing in both of these if statements you can do an` if(!timing) 
return` and then have an else on line 34

code/modules/assembly/pass.dm
```diff
@@ -0,0 +1,36 @@
...
+		loca = get_turf(src)
+		if(loca == lastloc)
+			time -= 2 // 2 seconds per process()
```
Ty-Omaha: Shouldn't it go down by 1 second so that way if you have the examine text it will be accurate?

code/modules/assembly/pass.dm
Ty-Omaha: Use /* commented out whatever blah blah blah */ for block comments

code/modules/assembly/pass.dm
SkullyRoberts: Ty. that is copy/pasted from the timer code. Soooo... for some reason, this is what works.

Ty-Omaha: Does it work though? The processing subsystem says it process every 1 second https://github.com/ParadiseSS13/Paradise/blob/0fb164d041aa7db36ea6fa432284559bac6aa250/code/controllers/subsystem/processing/processing.dm

Ty-Omaha: It is possible that the timer code is outdated

SkullyRoberts: I'm not sure. When I tested this locally I only really checked to see if 1) the alarm went off continuously when not moving for x seconds and 2) the alarm shut up when you moved

Mitchs98:
https://m.youtube.com/watch?v=7vIjBtdEQRE

SkullyRoberts:
For the record, this is an actual thing that pretty much every firefighter in the US carries whenever they are in an IDLH area.

Fox-McCloud:
I don't think so.

In some situations this will be *better* than a death alarm.

For those who are new to the game, this will just get them caught a lot.

For those who are experienced, it'll only encourage them to go even more lethal than before and/or take even more of their things.

code/modules/assembly/pass.dm
Citinited: You don't even need these two vars, when you have a perfectly good `loc` var to use.

Just run down the timer in `process()` and have the timer reset to its initial value upon `Move()`.

variableundefined:
I think we'll pass this one

PR #12632, opened 2019-10-22
Made megafauna have randomized stats between half and 2 times the default stats

SkullyRoberts:
…rent values, decided per mob

## What Does This PR Do
On map initialization, it randomizes the damage, health, speed, and armor penetration of all megafauna to be somewhere between 1/2 and 2 times the stats stated in the mob's file.

## Why It's Good For The Game
There's too much of an established meta for experienced lavalanders. Some bosses are hillariously easy if you know the strategy (hierophants) and some are very lethal even if you've done it hundreds of times (colossus). Let's throw a wrench in the meta strats!

## Changelog
:cl:RobinFox
balance: Gave all megafauna randomized stats
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
@@ -47,6 +47,11 @@
 	for(var/action_type in attack_action_types)
 		var/datum/action/innate/megafauna_attack/attack_action = new action_type()
 		attack_action.Grant(src)
+	maxHealth = maxHealth*(rand(25,100)/50)
+	armour_penetration = armour_penetration*(rand(25,100)/50)
+	melee_damage_lower = melee_damage_lower*(rand(25,100)/50)
+	melee_damage_upper = melee_damage_upper*(rand(25,100)/50)
+	speed = speed*(rand(25,100)/50)
 
 /mob/living/simple_animal/hostile/megafauna/Destroy()
 	QDEL_NULL(internal_gps)


TDSSS:
I like the idea of throwing off the meta, but pure stat changes won't really do that. If it also changed their behaviour or gave/took away special attacks or made them more common, that might break the meta.

SkullyRoberts:
@TDSSS i could have it take away random abilities, but that would mean either 1) making them weaker or 2) adding a bunch more abilities to each megafauna

SteelSlayer:
I think a better idea would be to have a chance of spawning "elite" versions of megafauna that have increased stats. Similar to how we have the more powerful unique verions of watchers and goliaths. You could tweak their colors and name a little bit to let players know that it is a stronger monster.

Mitchs98:
Against this. Unless a much larger rework is put into place adding modified versions of the megafauna, modifying their loot tables and as stated attack patterns or such, I don't think they need a change. If people have gotten good enough to be able to reasonably tackle content that's good on them. They earned that much. They shouldn't be slapped in the face, especially if there's no discernable way to notice the changes, and told their work was all for nothing.

On the flipside, there definitely shouldn't be easier to kill megafauna. At all. Harder megafauna are harder for a reason. I get they're like that to balance to opposite spectrum of 2x strength. But I don't think stat buffs are needed in general.

Fox-McCloud:
👎

AdamElTablawy:
ash drakes could have thirty million HP and kill you instantly and they'd still be just as easy if you know the pattern

this isn't a solution

variableundefined:
This is a terrible idea

PR #12639, opened 2019-10-23
Rework changeling swap form

datlo:
This PR reworks (and for the most part, heavily nerfs) the changeling's 'Swap form' ability. 

This is what the ability does on live : 

The ability cost 1 dna point and 40 chemicals.
You need a reinforced (not neck, just reinforced.) grab to sting. On sting, after a near instant delay (2 seconds if I'm correct), you will swap with the target. You will be paralised for a short duration, which you can instantly cancel with changeling's antistun ability.

Compare this with the WIZARD's mindswap : 

Cost 2 spellpoints and has a 60 seconds cooldown. After using this on a target, you  instantly swap, and both of you are asleep for about 20 to 30 seconds?

This is bonkers and I have no idea how changelings have had this ability be that overpowered for so long, I personally had no idea it was that fast before today.

This is what the ability will do if this PR is merged : 

Cost 1 dna, and 5 chemicals. You need a neck grab to perform the swap. 

Once you start swapping, you will start a slow process of pushing yourself into the new body. This will take the same time as an absorption (45 seconds). Once you've swapped with the target, your original body is husked.

This reworks the ability from being a wizard spell with different flavor into an alternative to absorbing, where instead of absorbing and husking the target, you become the target and husk your old body, allowing you to perfectly impersonate them.

TODO:

- Give the changeling the target's recent speech and memory similar to absorbing
- Test everything thoroughly

:cl:
balance: Changeling's swap form ability now requires a neck grab instead of aggressive grab.
balance: Changeling's swap form ability now takes 45 seconds to perform instead of 2.
balance: Changeling's swap form ability now kills and husk's the changeling's original body after swap.
balance: Changeling's swap form ability now costs 5 chemicals instead of 40.
/:cl:


TDSSS:
Any power that encourages CHangelings to do actual impersonating of crew instead of just running around going loud is good in my book, so a 👎 from me for making that harder. You can't compare it directly with wizard, and it being stronger is no reason it should be nerfed.

datlo:
The current bodyswap on live doesnt encourage you to do impersonation. It encourages you to instantly swap an officer who's unable to do anything about it and use his own weaponry back on him. This is speaking from experience as HOS today : Changeling stuns me, instantly swap form with me, cancel the stun with antistun, unload my shotgun into my new face. Changeling proceeds to go even louder than before with a ton of guns, sec gear and access.

datlo:
I've considered what you said and opened an alternative PR that only blocks the ability from working on mindshielded crew, changing nothing else. This will still lets you use the ability to impersonate crew and hide from security, however it'll prevent using it for free gear (in most situations).

AzuleUtama:
This PR nerfs the swap form ability to point of near-uselessness. I'd argue it's a worse version of absorbing someone to a degree - since you don't get a remap out of this.

I've had people successfully break out of a grab when using swap form and the process looks really suspect to anyone nearby (the person grabbing them suddenly falls asleep and both the cling and victim jitter around a lot before it kicks in if i recall right). You will also lose all the items you have on you when you swap into the other person, so it can backfire if you swap forms and they run off with a taser or objective item you had on you.

I agree the time it takes to swap could be increased a bit to give victims more chance to break free from the grab alone, but reworking it to be a different means of absorbing someone seems excessive.

datlo:
On restrospect I went a bit too heavy handed with this PR. i'll leave the alternative PR up as I do believe this needs to be changed in some way, and I might rethink another rebalance

PR #12644, opened 2019-10-24
Make His Grace simpler to link with

datlo:
His Grace will now link with the user when they press Its action button if It is currently unlinked, aswell as informing the user that using the action again will let them bind It to their hand.

You have to click It with an empty hand while holding It currently, which is a little obscure. It's still possible to link It this way.

:cl:
add: His Grace can now be linked with Its action button, as an alternative to clicking It with an empty hand.
/:cl:


TheSardele:
Can you make it so clicking on a corpse/unconcious person in crit starts consuming them? Having to grab the person, then click on the toolbox with the grab is not very intuitive.

datlo:
I likely will add it in a future PR where I rebalance/refactor the toolbox on a larger scale, which is on my todo list, this is just a simple change that makes it easier to use without affecting balance.

PR #12646, opened 2019-10-25
nerf friendship

Couls:
## What Does This PR Do
Makes it so you can no longer push the person pulling you in disarm intent to go SUPER FAST

## Why It's Good For The Game
This is exploiting the system to go faster than you're supposed to

## Changelog
:cl:
fix: going really fast by pushing someone who's pulling you
tweak: you can no longer push people pulling you, you need to move away from them instead
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -70,6 +70,9 @@
 				to_chat(src, "<span class='warning'>[L] is restrained, you cannot push past.</span>")
 			return 1
 
+		if(pulledby == L) //prevents boosting the person pulling you
+			return 1
+
 		if(L.pulling)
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Friendship isn't magic.

It's just an in-game exploit.

KasparoVy:
Actual troll physics in action

Fox-McCloud:
More seriously, I'm not sure about this from a "fix the problem" stance. We "fixed" the issue of preventing people from pushing stuff away from you, only to make people pulling things literally immovable forces that will never move from their spot when bumped into--I could see this generating equally as strange effects.

Couls:
I can make it so using help intent will always make you go past them instead

PR #12647, opened 2019-10-25
Makes some space ruins easier to find.

TDSSS:
## What Does This PR Do
This PR takes advantage of the introduction of local gps signals by adding a local gps signal to some space ruins, making them easier to find. It does this via a stationary 'navigation console' in those ruins that work as a gps but can't be moved, same as the gps in survival pods.

As a reminder, a local gps signal is only visible to other gps on the same z level.

The three ruins that got such consoles are the crashed mech transport, the listening outpost asteroid and the abandoned zoo. I decided against giving the survivalist bunker one as it does contain pretty good weapons and should be a challenge to find.

## Why It's Good For The Game
Exploring space is a crapshot. You yeet yourself out of the station hoping to hit something interesting before you loop around, or if you're very patient you systematically search but that takes forever and it isn't very exciting just looking at empty space pass by for 20 minutes. Now a space explorer keeping an eye on his gps occasionally can find those ruins once he is on the same z level as them.

## Images of changes
Capture
You can see the navigation console on the wall next to the console, here in the listening outpost. Many thanks to Darkmight9 for the sprite.

## Changelog
:cl:
add: Some space ruins now emit a weak signal that a gps on the same z level can detect.
/:cl:


JKnutson101:
Well this is a pleasant surprise. I always liked finding ruins in space but heck, it got so tiring searching for them I would always give up, especially if all I found was generic asteroids with nothing on them.

S-CR:
Icon is.. definitely a codersprite.
But I like the idea, having to just hope you hit what you hit is not the greatest thing.

TDSSS:
>Icon is.. definitely a codersprite.

I thought it looked pretty good, but I am sure he'd be open to constructive criticism if you have it.

AdvnetureDrake:
If there's any worry about balancing the ease of finding space ruins, perhaps instead of giving an exact location, the GPS signal can instead list it as an unknown location instead of exact coordinates. This way, you know something is in the sector, but don't know the exact location, requiring you to do some searching to find it. I believe the Syndicate Lavaland base does something like this already.

lolzyking:
I agree anything 'too good' for a gps location should get an uknown signal hint, just to encourage people to search the z level

Also minor suggestion A binocular with the same zoom as a sniper rifle for the mechanic to use in exploration.

Drakeven:
If you like I can finangle with the Icon a bit, animate it and such!

Darkmight9:
@Drakeven I give you the go ahead to make the sprite fancier I'm not too happy with it myself looking at it in game. It also doesn't look like it belongs on the wall there. @TDSSS Alternatively I can make a new sprite for it. Just credit us both if you go with Drake's improved sprite.

code/modules/awaymissions/mission_code/ruins/gps.dm
```diff
@@ -0,0 +1,17 @@
...
+/obj/item/gps/ruin
+	name = "navigation console"
+	desc = "A console for navigation in local space, gives off a weak signal that can be picked up if sufficiently close."
+	icon = 'icons/obj/terminals.dmi'
+	icon_state = "gps_console"
+	anchored = TRUE
+	local = TRUE
+	gpstag = "Unknown Signal"
+
+/obj/item/gps/ruin/New()
```
farie82, line 12: This proc is not required. The name is already set in the object define

TDSSS: Check gps/New(), it sets the name to "global positioning system ([gpstag])" on top of doing some other things.

farie82: Good point. Overlooked that one. Mappers can't change the name now though due to this. Not sure how `initial(name)` interacts with mapping variables overrides.

TDSSS: Hmm, what if I change generic gps in general so that their initial name is NULL and then the renaming in the New only happens if it is null?

farie82: that is one way to solve it yes. Another option would be to add a variable or such. But the one you suggested will work

PR #12652, opened 2019-10-27
Fix paperwork acquiring extra line breaks when you change register

### LINKED ISSUE #12158: Itallics on papers breaks enters
**Issue Description**:

When using itallics [i]&[/i] (or [sign], it puts a br or something on either side.

When typing `1[i]2[/i]3` to show `1_2_3`

**What happened instead**:

When typing `1[i]2[/i]3` it showed
```
1

2

3
```

muh paperwork

1) Get paper & pen.
2) Write `1[i]2[/i]3` or `1[sign]1`

Broke at this commit: https://github.com/ParadiseSS13/Paradise/pull/11806/commits/9d9bf046e3d54f6e18cefd9c6c2f7097bf21ed38

**Extra information**:

no

moxian:
## What Does This PR Do
Fixes extra linebreaks being added when they shouldn't be.
Specifically:
Fixes #12158 
Fixes unreported bug, where adding a new field wraps it in extra line breaks

Separately, split the paradise-specific customization of marked.js away from the core marked.js. This makes it easily updatable - simply download the new version from https://github.com/markedjs/marked and drop in here - and also, (i believe) makes the licensing situation less ambiguous.

Here's what was happening with the bug:
When you type `1[i]2[/i]3`, the pencode gets parsed into html as `1<I>2</I>3`.
After that marked.js comes in and wraps *every* single text node in `<p>` tags, so you get `<p>1</p><I><p>2</p></I><p>3</p>`, creating all those extra line breaks.

Similarly filling field with `some text here` actually fills them with `<font face="Verdana" color=black>some text here</font>` so you, again, get an extra tag wrapping your text, and get an extra `<p>` inserted.

#11845 attempted to fix a part of the problem by blacklisting certain specific tags, but the problem persists for the rest of them.

Instead what I'm doing here is preventing the *extra* wrapping of the entire DOM node content in `<p>` tags. The internal `<p>` that are actually needed to separate paragraphs stay as they are.

## Why It's Good For The Game
Paperwork looking as one expects it to.
Being able to write *italic* words amidst non-italic text without breaking for a new paragraph.

## Images of changes
dreamseeker_2019-10-27_06-51-46
dreamseeker_2019-10-27_06-52-06

## Changelog
:cl:
fix: fixed paperwork acquiring extra line breaks when you change register
/:cl:


craftxbox:
Does this change pencode? If so, im not sure if people would be happy.

moxian:
No, it does not. As you can see in the screenshot you can still mix and match pencode with markdown.

craftxbox:
oooh, fancy! 👍

Tayyyyyyy:
I'm so happy you figured this out! 😊

One thing I would note for future updates of marked.js is that the link feature needs to be disabled before being dropped in.

PR #12658, opened 2019-10-28
Dead mice squeak no tails

### LINKED ISSUE #11596: Mouses squik post mortem when you walk on them.
**Problem Description**:
What is the problem?
You walk on mouse, it squiks. You kill it, then walk on it, and it keeps making noises, even though its dead. Security hears, and comes to you.

**What happened instead**:
I expected for dead mouse to not make noises, when you walk over it.
Makes you less stealthy.

1. kill mouse
2. walk over it, it will squik
since few weeks ago I think

Kyep:
Fixes #11596. RIP undead mice everywhere.

:cl: Kyep
fix: Fixed dead mice still squeaking when stepped on.
fix: Fixed mice held in your hand, or a backpack, still chewing through cables.
/:cl:

### DIFF
--- a/code/datums/components/squeak.dm
+++ b/code/datums/components/squeak.dm
@@ -41,6 +41,10 @@
 		use_delay = use_delay_override
 
 /datum/component/squeak/proc/play_squeak()
+	if(ismob(parent))
+		var/mob/M = parent
+		if(M.stat == DEAD)
+			return
 	if(prob(squeak_chance))
 		if(!override_squeak_sounds)
 			playsound(parent, pickweight(default_squeak_sounds), volume, 1, -1)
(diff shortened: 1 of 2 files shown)


code/datums/components/squeak.dm
```diff
@@ -41,6 +41,10 @@
 		use_delay = use_delay_override
 
 /datum/component/squeak/proc/play_squeak()
+	if(ismob(parent))
```
farie82: Might want to just unregister on death. And register back when the mouse comes back alive. Unsure how viable that is though

Couls:
fixing spooky bugs during halloween, wait until thursday at least sheesh.

code/datums/components/squeak.dm
Kyep: I did explore during that, but then I realized there are a few different methods it could come alive again, from staves of healing, admin rejuv, lazarus, etc.
I felt that trying to remove the component and then re-add it in each possible case would just lead to more bugs and edge cases. For example, worries about incorrectly adding the component multiple times.
For all of these reasons I felt it was cleaner, clearer and less bug-prone to just disable the squeaking while the mob is dead.

farie82: we kinda need one way to revive things. Or atleast have the ways to revive them call a signal or such.

Crazylemon64: `update_revive` is the standard way to revive a mob

PR #12659, opened 2019-10-28
T-ray scanner now shows plumbing only to the users, not nearby mobs.

moxian:
## What Does This PR Do
From tg, more or less.

- Makes T-ray overlay only visible for the user, not for other people nearby.

- Minor refactor: the plumbing invisibility values are switched from numerical 101 (which is INVISIBILITY_ABSTRACT) to INVISIBILITY_MAXIMUM (which is 100). The INVISIBLITY_MAXIMUM seems to be the semantically correct value, but I don't think this would have any noticeable effect in game.

- The old code tries to reveal "invisible" monsters on scan, but I've never heard of scanner being actually used for this purpose. Probably because only such monsters are ninjas, which are currently admin-only...

Changes from tg:
- `for(var/turf/T in range(scan_range, viewer)) for(var/obj/O in T.contents)` instead of `for(var/obj/O in orange(scan_range, viewer)` . I don't know why but `range` iterates over turfs contents on /tg/, yet not on para...
- `range` instead of `orange`: because you are likely interested in seeing piping even on the tile you're currently standing

## Why It's Good For The Game
T-ray view is only really useful for the person holding the scanner, and people nearby are usually mildly annoyed by the flashing. This is especially bothersome when an unconscious engineer comes/is brought to medbay/cloning.

## Images of changes
N/A

## Changelog
:cl:
tweak: T-Ray imagery is now only visible to T-Ray scanner user, not to outside observers.
/:cl:

P.s.: as a future work items, one could introduce:
- Increase the T-Ray scanner range to make it easier to work with (now that it's unobtrusive to bystanders) (the range is 3 on tg, compared to 1 here)
- Add ability icon for toggling T-ray scanner
- Add T-ray goggles (combined with optical meson scanners probably?)
- Add T-ray ghost view

but these are all items for a different PR

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -20,7 +20,7 @@ REAGENT SCANNER
 	materials = list(MAT_METAL=150)
 	origin_tech = "magnets=1;engineering=1"
 	var/scan_range = 1
-	var/pulse_duration = 10
+	var/pulse_duration = 10  // in deciseconds
 
 /obj/item/t_scanner/longer_pulse
 	pulse_duration = 50
@@ -37,7 +37,7 @@ REAGENT SCANNER
 		STOP_PROCESSING(SSobj, src)
 	return ..()
 
-/obj/item/t_scanner/proc/toggle_on()
+/obj/item/t_scanner/attack_self(mob/user)
 	on = !on
 	icon_state = copytext(icon_state, 1, length(icon_state))+"[on]"
 	if(on)
@@ -45,21 +45,15 @@ REAGENT SCANNER
 	else
 		STOP_PROCESSING(SSobj, src)
 
-/obj/item/t_scanner/attack_self(mob/user)
-	toggle_on()
-
 /obj/item/t_scanner/process()
 	if(!on)
 		STOP_PROCESSING(SSobj, src)
 		return null
 	scan()
 
+// Show the holder the T-ray layout within `scan_range` for the `pulse_duration` deciseconds.
 /obj/item/t_scanner/proc/scan()
-	t_ray_scan(src.loc, pulse_duration=pulse_duration, scan_range=scan_range)
-
-// Show the viewer the T-ray layout within `scan_range` for the `pulse_duration` deciseconds.
-/proc/t_ray_scan(mob/viewer, pulse_duration, scan_range)
-	var/flick_time = pulse_duration
+	var/mob/viewer = src.loc
 	if(!ismob(viewer) || !viewer.client)
 		return
 
@@ -78,8 +72,8 @@ REAGENT SCANNER
 				I.appearance = MA
 				t_ray_images += I
 	if(t_ray_images.len)
-		flick_overlay(t_ray_images, list(viewer.client), flick_time)
-	
+		flick_overlay(t_ray_images, list(viewer.client), pulse_duration)
+
 
 /proc/chemscan(mob/living/user, mob/living/M)
 	if(ishuman(M))
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -37,14 +37,16 @@ REAGENT SCANNER
 		STOP_PROCESSING(SSobj, src)
 	return ..()
 
-/obj/item/t_scanner/attack_self(mob/user)
-
+/obj/item/t_scanner/proc/toggle_on()
 	on = !on
 	icon_state = copytext(icon_state, 1, length(icon_state))+"[on]"
-
 	if(on)
 		START_PROCESSING(SSobj, src)
+	else
+		STOP_PROCESSING(SSobj, src)
 
+/obj/item/t_scanner/attack_self(mob/user)
+	toggle_on()
 
 /obj/item/t_scanner/process()
 	if(!on)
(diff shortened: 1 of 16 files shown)


TDSSS:
>The old code tries to reveal "invisible" monsters on scan, but I've never heard of scanner being actually used for this purpose. Probably because only such monsters are ninjas, which are currently admin-only...

Does it no longer do that with your PR? It might be best to keep that functionality intact for future use.

moxian:
It no longer does that with my PR.

There are two flavours of invisibility that were being revealed here, so let's discuss them separately:

One is monsters with their opacity set to <100% - the icon is there, the item shows up when alt-clicking the tile, just hard to see with the naked eye. The reveal of these was introduced in 9e0e080376d63d324c4a1f75410cec9823efab92 while touching stealth genes. I didn't touch Para's genetics almost at all, but I think those mutations are longer obtainable?

The other one is ninjas, which were there from the beginning of time, but never actually used. I think the situation with their invisibility is neatly described by the following comments:
https://github.com/ParadiseSS13/Paradise/blob/54838ccfcf1579128d9dfdb2d13576b982f57e12/code/__DEFINES/sight.dm#L9-L12

In either case, I don't think leaving the code in "just in case" is a good idea. At best it would sit unused and confused the people who look at the file in the future; at worst they would introduce extra maintenance work when trying to preserve the behavior while doing yet another refactor.


Ideological differences aside, it's quite challenging to implement the reveal in the new system on the technical level. What T-scan does for pipes is creates the pipe image overlay, and presents it to the user for a set duration.
If we apply the same treatment to invisible mobs, you would get a still image of the mob at the place they were when the T-pulse happened - it would not be updated when the mob moves. I couldn't figure out a way to make the imagery update (neither did tg folks, apparently), or I'd use that here.
If we treat them the way they were treated before the PR - then they would suddenly become visible for _everyone_ - not just for T-scanner user - despite having no indication as to _why_, which sounds like bad gameplay experience to me.


All that said, I could still be convinced to reintroduce either/both of the invisibility flavours being revealed via either old/new mode (please specify which/how).

TDSSS:
>One is monsters with their opacity set to <100% - the icon is there, the item shows up when alt-clicking the tile, just hard to see with the naked eye. The reveal of these was introduced in 9e0e080 while touching stealth genes. I didn't touch Para's genetics almost at all, but I think those mutations are longer obtainable?

Uh, there's a few mobs that make use of this, I think. The ambush terror spiders have stealth while on webs, xenos can stealth and become harder to see and I think there's a vampire power for being less visible in darkness, no? Or do they use a different system for that?

code/game/objects/items/devices/scanners.dm
```diff
@@ -53,42 +55,31 @@ REAGENT SCANNER
...
 
 /obj/item/t_scanner/proc/scan()
+	t_ray_scan(src.loc, pulse_duration=pulse_duration, scan_range=scan_range)
 
-	for(var/turf/T in range(scan_range, src.loc) )
-
-		if(!T.intact)
-			continue
+// Show the viewer the T-ray layout within `scan_range` for the `pulse_duration` deciseconds.
+/proc/t_ray_scan(mob/viewer, pulse_duration, scan_range)
```
Citinited: Why is this its own proc instead of just doing this code inside `proc/scan()`?

moxian: Because that's the way it was done on tg.
They have both the hand held scanner and T-ray goggles, which both call this proc.

Honestly, I'm still very unsure of how close I should keep to the tg original when doing ports, especially when it comes to small things like this. I might be reading too much into this section of the contributing guide here:

> Document and explain your pull requests thoroughly. Failure to do so will delay a PR as we question why changes were made. This is especially important if you're porting a PR from another codebase (i.e. TG) and *divert from the original*.

(emphasis mine) but that was the motivation for keeping stuff as close to the port source as possible.

Should i remove the standalone proc?

Citinited: Oh, this will explain why the proc isn't typecast.

What I would do is make it a t-ray scanner level proc for now and then repath it once you port the goggles. I think the maintainers would be ok with it being a different path to tg code (obviously I'm not speaking for them here though)

moxian: Removed standalone proc. Thanks!

moxian:
> The ambush terror spiders have stealth while on webs
It is done differently and unaffected by today's T-ray scanner. Namely, when in webs they use a different semitransparent icon (with semitransparent pixels) - the icon *itself* is rendered at 100% opacity though.

Grey spider webs are implemented via alpha tweaking. However the webs are not /mob/living so they are skipped in the scanner check and not revealed.

> there's a vampire power for being less visible in darkness
This also via alpha, so this one _should_ be revealable. However in my testing I couldn't get it to work. I'm not sure why.
One explanation can be that the alpha *is* set to 100% opacity in the t-ray proc, but then immediately overriden in the vampire's Life proc, so it's never really visible.

TDSSS:
Hmm, in light of that you might be best off not having it work on vampires and co for now. Might be better to add a separate item later than can reveal invisible mobs, especially if any antags with full stealth are added (predators to go with our aliens, maybe?)

code/game/objects/items/devices/scanners.dm
```diff
@@ -38,56 +38,41 @@ REAGENT SCANNER
...
-						M.alpha = oldalpha
-
-		var/mob/living/M = locate() in T
-
-		if(M && M.invisibility == 2)
-			M.invisibility = 0
-			spawn(2)
-				if(M)
-					M.invisibility = INVISIBILITY_LEVEL_TWO
+			if(O.invisibility == INVISIBILITY_MAXIMUM || HAS_TRAIT(O, TRAIT_T_RAY_VISIBLE))
```
farie82, line 67: This will make things appear that were not shown prior. Might want to take another look at what items will be visible due to this. As an example the poolcontroller will become visible now.

You might have to change those to have a more fitting invisibility value though

Fox-McCloud:
No updates, still conflicted, sadly.

Nothing wrong with this idea, at all---but closing due to stale-ness. Feel free to re-submit though.

PR #12663, opened 2019-10-29
Allow observers to inspect storage contents by clicking it.

moxian:
## What Does This PR Do
Allows ghosts to see storage contents by clicking on the storage (boxes, toolboxes, belts etc).

From tg but not really.
tg's storage handling is much.. refactored from what it is on Paradise, and I'm not going to port all of it here in one go.

Just taking a few small parts in hopes we can maybe migrate there  gradually... (not really, but one can dream)

## Why It's Good For The Game
Ghosts get a bit more involved in the round, and have a bit more fun (without screwing with the living)

## Images of changes
dreamseeker_2019-10-29_10-22-48

## Changelog
:cl:
add: Ghosts can inspect containers by clicking on them
/:cl:

### DIFF
--- a/code/datums/components/storage/storage.dm
+++ b/code/datums/components/storage/storage.dm
@@ -1,0 +1,17 @@
+
+/datum/component/storage
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+
+/datum/component/storage/concrete
+
+/datum/component/storage/Initialize(datum/component/storage/concrete/master)
+	if(!isatom(parent))
+		return COMPONENT_INCOMPATIBLE
+
+	RegisterSignal(parent, COMSIG_ATOM_ATTACK_GHOST, .proc/show_to_ghost)
+
+/datum/component/storage/proc/show_to_ghost(datum/source, mob/dead/observer/M)
+	var/obj/item/storage/s = parent
+	if(istype(s))
+		return s.show_to(M)
+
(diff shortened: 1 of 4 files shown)


code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -32,7 +36,6 @@
 			return
 
 		if(over_object == M && Adjacent(M)) // this must come before the screen objects only block
-			orient2hud(M)          // dunno why it wasn't before
```
moxian: This orient2hud (and several others) are moved into show_to, since that is what ultimately needs it.

code/datums/components/storage/storage.dm
```diff
@@ -0,0 +1,17 @@
+
+/datum/component/storage
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+
+/datum/component/storage/concrete
```
farie82: why this subtype?

moxian: Only to mirror what tg is doing, no other reason.
Should I remove it if it's not serving any real purpose?

Fox-McCloud:
Rreally don't think we should be half-assed implementing the storage component, especially like that.

We have far far too many half finished half implemented things in the codebase, as is; the buck needs to stop somewhere.

I'm also pretty sure this doesn't really need components to work, to begin with. TG had this LONG before components were even a thought.

moxian:
Acknowledged, will rework.

PR #12664, opened 2019-10-29
Allow admins to rename people with numbers

Ty-Omaha:
Numbers were removed when admins try to rename people in VV because the argument wasn't set to TRUE.
:cl:
fix: Admins can now rename people to names that include numbers, this is useful for silicons and IPCs.
/ 🆑

### DIFF
--- a/code/datums/datumvars.dm
+++ b/code/datums/datumvars.dm
@@ -548,7 +548,7 @@
 			to_chat(usr, "This can only be used on instances of type /mob")
 			return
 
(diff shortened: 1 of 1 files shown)


Spacemanspark:
While we're on the topic, why can't the HoP computer edit names to add numbers?

Birdtalon:
@Spacemanspark same reason as this PR.

https://github.com/ParadiseSS13/Paradise/blob/1029bf7e7af28ad4b393a954641f22223b2be7db/code/game/machinery/computer/card.dm#L438

PR #12678, opened 2019-10-31
Fixes 12677 - depot self-destruct leaves loot behind

### LINKED ISSUE #12677: Syndicate Depot Delta still leaves treasure behind.
**Issue Description**:
Some loot still remains after the depot explodes, and the explosion leaves a lot of debris.

The depot to be completely gone.

**What happened instead**:
A lot of the depot remained with some very strong loot inside as well.
depot_explosion_live

Free loot bad.

Trigger Delta alert on Syndie Depot. Check remains.

Well #12245 most likely again.

**Extra information**:
#12535 is basically the same issue, which was supposed to be fixed by  #12539. No idea what the issue could be caused from.

Picture was taken on live server.

### LINKED ISSUE #12856: Syndicate Depot Self Destruct Disabler
**Issue Description**:
You can remove the entity that is causing the self destruct of the Syndicate Depot upon reaching the Delta Alert level.

wasn't expecting anything, was just dicking around

**What happened instead**:
The Tesla-thingy that causes the self destruction was bombed away.

Delta alert doesn't matter anymore.

https://www.youtube.com/watch?v=5I-9_CUHqo0&feature=youtu.be

28.12.2019
17:32

**Extra information**:
:^)

Kyep:
## What Does This PR Do
When depot self destructs, it first deletes all loot in its loot lockers, to prevent anyone recovering restricted loot after the blast.
~~Also fixes an obscure bug where the depot comms computer fails to recognize some syndie agents as such during traitorchan gamemode.~~
Makes the 'overload' effect (the object that handles depot self destruct) immune to explosions, so it cannot be exploited by destroying it with a grenade.

## Why It's Good For The Game
Fixes #12677 
Fixes #12856 

:cl: Kyep
fix: Fixed depot sometimes leaving valuable loot after self-destruct.
fix: Fixed depot self-destruct being possible to shut down with an explosive grenade.
/:cl:


Kyep:
@Cazdon Removed that part, on the basis of 12740.

code/game/objects/structures/depot.dm
```diff
@@ -113,6 +113,8 @@
 		depotarea.updateicon()
 
 	for(var/obj/structure/closet/L in range(30, T))
+		for(var/obj/O in L)
+			qdel(O)
 		L.open()
 	for(var/mob/living/M in range(30, T))
 		M.gib()
```
Fox-McCloud: calling `Destroy` directly is something you should *never* do under any circumstances ****ever***. 

This hsould probably just deal a crap ton of damage to them to break them.

code/game/objects/structures/depot.dm
```diff
@@ -113,6 +113,8 @@
 		depotarea.updateicon()
 
 	for(var/obj/structure/closet/L in range(30, T))
+		for(var/obj/O in L)
+			qdel(O)
 		L.open()
```
Fox-McCloud: I suspect the 4 lines above may not be necessary; I suspect that what's happening is that the closet opening, before the explosion, together with how explosion process, it leaving "gaps". 

If this is all gone, when explosions happen, it *should* result in all the contents getting `ex_act`'d. The fact the lockers don't open should prevent the atmos from pushing them around and causing weird issues.

code/game/objects/structures/depot.dm
Kyep: Changed to take_damage, verified works, resolving.

code/game/objects/structures/depot.dm
Kyep: Having tested this by spawning lockers, I found that leaving out this code caused some locker items to survive, e.g. grenades that armed, etc. So it does seem necessary.

PR #12679, opened 2019-10-31
[DNM] Some more mindslave code fixes

### LINKED ISSUE #12675: Wololo antag log/hud bug.
**Issue Description**: Antags that get wololo'd no longer show up as antags on attack logs and antag hud after wololo wears off.

**What did you expect to happen**: It to keep their original antag status on attack logs.

**What happened instead**: Attack logs and antag hud shows a normal player.

**Why is this bad/What are the consequences**: Creates inaccuracies in logs, and misinformation for people who turn antag hud on.

**Steps to reproduce the problem**: Wololo an antag, wait for it to wear off, check antag hud and attack logs.

**When did the problem start happening**: High pop traitor vamp shift.

SteelSlayer:
## What Does This PR Do
Fixes mindslaves and zealot's not getting their objectives removed when they lose the mindslave status. 

Antags who are mindslaved now revert to their previous special role when they lose the mindslave status. For example, if a traitor gets mindslaved and then that status gets removed, their `special_role` won't be null, it will get reverted back to traitor.

Fixes  #12675

## Changelog
:cl:
fix: Fixes mindslaves and zealots not getting their objective removed when they lose the mindslave status
fix: Antagonists who are mindslaved now revert to their previous special role when they lose the mindslave status
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -1662,7 +1662,7 @@
 	SSticker.mode.traitors += src
 
 	
-	var/datum/objective/protect/zealot_objective = new
+	var/datum/objective/protect/mindslave/zealot_objective = new
 	zealot_objective.target = missionary.mind
 	zealot_objective.owner = src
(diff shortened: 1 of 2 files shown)


SteelSlayer:
This isn't going to fix the issue currently

PR #12680, opened 2019-10-31
Fixes depot mobs reacting very slowly

Kyep:
## What Does This PR Do
Improves the reaction speed of depot AI mobs.

## Why It's Good For The Game
Currently, their reaction speed is so slow that someone can walk into their room, open their locker, take their item, and run out, before they notice that a player is within their vision range.
This improves their reaction time, so that they will now notice much faster when a player enters their vision range.

:cl: Kyep
fix: Fixed depot mobs taking a long time to react to the presence of nearby hostiles.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/syndicate.dm
+++ b/code/modules/mob/living/simple_animal/hostile/syndicate.dm
@@ -148,8 +148,10 @@
 			// This prevents someone from aggroing a depot mob, then hiding in a locker, perfectly safe, while the mob stands there getting killed by their friends.
 			LoseTarget()
 
-/mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/handle_automated_movement()
+/mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/handle_automated_action()
 	. = ..()
+	if(!.)
+		return
 	if(!istype(depotarea))
 		return
 	if(seen_enemy)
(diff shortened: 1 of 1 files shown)


JKnutson101:
Have you consider applying this to most simple mobs or do most simple mobs already react at a faster speed?

Kyep:
There are many issues with that, @JKnutson101.
The depot only has ~7 mobs that this change would affect. Even then, >90% of the time the mobs don't run AI at all, because their AI is disabled since there is nobody living on the zlevel. When they are active, they need to react at a decent speed for the depot to be challenging. So, for all these reasons, it makes sense for them to have their AI process faster. It makes a big difference to the experience without substantially increasing CPU load.
Applying the same change to ALL NPC mobs isn't the same. There are tons of NPC AI mobs on station (all pets, all xenobio monsters, etc). They're always active. And they don't need fast reaction time. So making all of them use the same system might consume more CPU than is worth it.
It might make sense to apply this update to some monsters, such as enemies in gateway missions (which tend to be reasonable and fixed in number, which need decent reaction speed to make the missions challenging, and crucially which don't run their AI 90% of the time as there's nobody in the gateway).
That's really a case-by-case determination though.
I don't want to apply a change like this in a blanket fashion to every simple animal or NPC. At most, I might apply it to other NPCs where I notice their sluggish reaction speed impairs their ability to perform their role.

JKnutson101:
Wow, thanks for the detailed response. Stuff like CPU usage hadn't even crossed my mind, so I learned something new!

PR #12681, opened 2019-10-31
Ghosts can browse though paper bundles.

### LINKED ISSUE #3090: Ghosts [obviously] can't turn pages in bundled papers
They need to have the paper in their active hand. Ghosts do not have hands.
Result: ghosts can only read the first page in bundled papers. It'd be swell if we could have a feature that would let ghosts access all the other pages. It should be visible only to them, of course, as we do not want ghosts to interact with the real world.
And yes, let's not forget that humans should still have to turn the pages physically, so no magical page-turning with hands cuffed.

moxian:
## What Does This PR Do
Allows ghosts to browse through paper bundles.
Fixes #3090

Living people still need to hold the bundle in hand to physically turn over the pages, but ghosts can now flip them over "virtually".

Also minor paper bundle refactor in general - we no longer need a separate `amount` variable which is basically `src.contents.len - 1` at all times, and we also don't need a separate "is this the first page, the last page or the middle page" variable, since we can calculate it when needed instead.

## Why It's Good For The Game
Ghosts have more fun.
Aghosts have more info to make informed decisions.

## Images of changes
N/A

## Changelog
:cl:
add: Ghosts can browse through paper bundles now
/:cl:

### DIFF
--- a/code/modules/paperwork/paper_bundle.dm
+++ b/code/modules/paperwork/paper_bundle.dm
@@ -11,7 +11,7 @@
 	layer = 4
 	pressure_resistance = 2
 	attack_verb = list("bapped")
-	var/opened_page = 1
+	var/opened_page = 1  // page number the bundle is opened on
 
 /obj/item/paper_bundle/New(default_papers = TRUE)
 	if(default_papers) // This is to avoid runtime occuring from a paper bundle being created without a paper in it.
@@ -91,7 +91,7 @@
 /obj/item/paper_bundle/examine(mob/user)
 	. = ..()
 	if(in_range(user, src))
-		show_content(user, src.opened_page)
+		show_content(user, opened_page)
 	else
 		. += "<span class='notice'>It is too far away.</span>"
 
@@ -99,7 +99,7 @@
 // `page_no` controls which page to show, 0 is a special value that means "show the currently opened one".
 /obj/item/paper_bundle/proc/show_content(mob/user as mob, var/page_no = 0)
 	if(page_no == 0)
-		page_no = src.opened_page
+		page_no = opened_page
 	var/dat
 	var/obj/item/W = src[page_no]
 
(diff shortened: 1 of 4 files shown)


moxian:
No, it won't.
Changing pages for the ghosts is purely virtual, and does not leave any trace in the world.

code/modules/paperwork/paper_bundle.dm
```diff
@@ -136,53 +127,45 @@
...
+				playsound(loc, "pageturn", 50, 1)
+			show_content(usr, new_page)
+
+		if(href_list["remove"] && can_flip)
+			var/obj/item/W = src[opened_page]
 			usr.put_in_hands(W)
 			to_chat(usr, "<span class='notice'>You remove the [W.name] from the bundle.</span>")
-			if(amount == 1)
-				var/obj/item/paper/P = src[1]
+			if(opened_page > contents.len)
```
farie82, line 156: Can use `max(opened_page, contents.len)` here

code/modules/paperwork/paper_bundle.dm
```diff
@@ -136,53 +127,45 @@
...
+		if(href_list["flip"])
+			var/flip_count = text2num(href_list["flip"])
+			if(!isnum(flip_count))  // should never happen
+				stack_trace("Paper bundle was flipped without specifying numeric flip amount. href_list\[\"flip\"] == [href_list["flip"]]")
+				flip_count = 0
+			var/new_page = Clamp(1, viewed_page + flip_count, contents.len)
+			if(can_flip && new_page != opened_page)
+				opened_page = new_page
+				playsound(loc, "pageturn", 50, 1)
+			show_content(usr, new_page)
```
farie82, line 150: This is already done in the attack_self call on line 168 which will override the page it's on with the `opened_page` value.
It works due to the attack_self call only going of for mobs holding it in their contents but this causes more networking and what not due to the double showing of the same window. You do want to keep the finger print stuff though I imagine that attack_hand gives.

Fox-McCloud:
Please deconflict. Otherwise, this is quite stale.

Spacemanspark:
Just like my life

moxian:
I guess i don't have enough interest to work on this anymore.

PR #12687, opened 2019-11-03
Adds the airlock painter

SteelSlayer:
## What Does This PR Do
Ports over a modified version of TG's airlock painter. This version doesn't include the toner mechanic, which has you refilling the painter to be able to use it. None of our other painters do this so I just didn't include that mechanic.

The airlock painter allows users to re-paint certain fully constructed airlocks. I've restricted the use of this to only include "normal" airlocks.

Excluded airlocks types: Vault, high security, all exotic material airlocks (uranium, plasma etc.), abductor, cult, hatches (the ones that have 4 bits that open), and shuttle doors.

These exclusions are mostly to prevent dumb shenanigans like making plasma airlocks and then painting them to look normal. Which can lead to people welding them and causing plasma fires randomly. Or for example cultists making cult airlocks, but disguising them for free with the painter.

Engineering spawns with two painters round-start and you can print more from an autolathe.

## Why It's Good For The Game
Changing the look of an airlock is extremely annoying at the moment. You have to deconstruct them either by hand, or by RCD, and then rebuilt them with the skin you want. This makes it far less annoying to change.

## Images of changes
Engineering spawns with two painters round-start:
painters

The airlock painter in action:
GIF

EDIT: I've changed the behavior of the painter slightly, so that you will not longer get this pop you see in the gif every time you click an airlock. It works like other painters now. You use it in your hand, select the paint job you want, and then apply it to airlocks.

## Changelog
:cl:
add: Adds the airlock painter. Two spawn in engineering on the large table, and more can be printed from an autolathe.
del: Deletes a multitool, t-ray scanner and engineering headset from the large table in engineering
imageadd: Adds the airlock painter item sprite, and inhand sprites for it
/:cl:


code/game/objects/items/devices/airlock_painter.dm
```diff
@@ -0,0 +1,61 @@
...
+
+	usesound = 'sound/effects/spray2.ogg'
+
+	w_class = WEIGHT_CLASS_SMALL
+	slot_flags = SLOT_BELT
+
+	materials = list(MAT_METAL = 3000, MAT_GLASS = 1000)
+
+
+/obj/item/airlock_painter/Initialize()
```
Couls: unnecessary if you're just gonna call . = ..()

code/game/machinery/doors/airlock.dm
```diff
@@ -500,6 +501,81 @@ About the new airlock wires panel:
...
+		return FALSE
+
+	var/list/optionlist
+	if(airlock_material == "glass")
+		optionlist = list("Standard", "Public", "Engineering", "Atmospherics", "Security", "Command", "Medical", "Research", "Science", "Mining", "Maintenance", "External", "External Maintenance")
+	else
+		optionlist = list("Standard", "Public", "Engineering", "Atmospherics", "Security", "Command", "Medical", "Research", "Freezer", "Science", "Mining", "Maintenance", "External", "External Maintenance")
+
+	var/paintjob = input(user, "Please select a paintjob for this airlock.") in sortList(optionlist)
+	if((!in_range(src, usr) && loc != usr) || !W.paint(user))
```
Couls: might be better to use user instead of usr, since user is already defined and you can stay consistent

code/game/objects/items/devices/airlock_painter.dm
```diff
@@ -0,0 +1,61 @@
...
+
+//Only call this if you are certain that the painter will be used right after this check!
+/obj/item/airlock_painter/proc/paint(mob/user)
+	playsound(loc, usesound, 50, TRUE)
+	return TRUE
+
+/obj/item/airlock_painter/suicide_act(mob/user)
+
+	var/obj/item/organ/internal/lungs/L = user.get_organ_slot("lungs")
+	var/lungs_name = "\improper[L.name]"
```
farie82: just [L] should do here

Couls: Don't even need the var then

SteelSlayer: I had some problems with it saying "X throws up his the lungs", but doing this stopped that issue.

code/game/machinery/doors/airlock.dm
SteelSlayer: makes sense

code/game/objects/items/devices/airlock_painter.dm
SteelSlayer: true that

_maps/map_files/cyberiad/cyberiad.dmm
```diff
@@ -84015,14 +84015,14 @@
 	tag = ""
 	},
 /obj/structure/table,
-/obj/item/t_scanner,
-/obj/item/radio/headset/headset_eng,
-/obj/item/multitool{
-	pixel_x = 5
-	},
```
Couls: Did you purposely remove these things to make room for airlock painter?

SteelSlayer: Yes I did.

Couls: should include in CL

SagaStars:
Could this be made in a protolathe/autolathe?

SteelSlayer:
@SagaStars Yes

taukausanake:
Can you give the actual painting a timer and not make it instant color change?

SteelSlayer:
@taukausanake I could yeah, but what for? Floor painters and pipe painters are instant

moxian:
> I could yeah, but what for? Floor painters and pipe painters are instant

There are much less doors than there are floors/pipes. Adding progress bar on floor/pipe painter would be annoying, but it won't be on doors.

I'm not a psychologist, but I'd say having an action take a bit longer than instant makes it feel more impactful and thus more rewarding (especially if you don't have to repeat that action 100x times).
It doesn't need to be a long delay. Something like half a second would probably be fine.

code/game/machinery/doors/airlock.dm
```diff
@@ -500,6 +501,81 @@ About the new airlock wires panel:
...
+/obj/machinery/door/airlock/proc/change_paintjob(obj/item/airlock_painter/W, mob/user)
+	if(!paintable)
+		to_chat(user, "<span class='warning'>You cannot paint this type of airlock.</span>")
+		return FALSE
+
+	var/list/optionlist
+	if(airlock_material == "glass")
+		optionlist = list("Standard", "Public", "Engineering", "Atmospherics", "Security", "Command", "Medical", "Research", "Science", "Mining", "Maintenance", "External", "External Maintenance")
+	else
+		optionlist = list("Standard", "Public", "Engineering", "Atmospherics", "Security", "Command", "Medical", "Research", "Freezer", "Science", "Mining", "Maintenance", "External", "External Maintenance")
```
moxian: The only extra option in non-glass is the freezer. What do you think about
```js
var/list/optionlist = list( < everything except freezer > )
if(airlock_material != "glass")
  optionlist |= "Freezer"
```
?

If freezer is not the only one missing - that's the more reason to make the distinction between the two lists clearer.

code/game/machinery/doors/airlock.dm
```diff
@@ -500,6 +501,81 @@ About the new airlock wires panel:
...
+		return
+	switch(paintjob)
+		if("Standard")
+			icon = 'icons/obj/doors/airlocks/station/public.dmi'
+			overlays_file = 'icons/obj/doors/airlocks/station/overlays.dmi'
+			assemblytype = /obj/structure/door_assembly
+		if("Public")
+			icon = 'icons/obj/doors/airlocks/station2/glass.dmi'
+			overlays_file = 'icons/obj/doors/airlocks/station2/overlays.dmi'
+			assemblytype = /obj/structure/door_assembly/door_assembly_public
```
moxian: Instead of copy-pasting these three values from `/obj/machinery/door/airlock/public`, maybe it's better to create a `/obj/machinery/door/airlock/public` value here and copy the values from it?
Something like (pseudocode, probably doesn't compile):
```js
var/obj/machinery/door/airlock/painted
switch(paintjob)
  if("Standard")
    painted = new/obj/machinery/door/airlock()
  if("Public")
    painted = new/obj/machinery/door/airlock/public()
  ...
icon = painted.icon
overlays_file = painted.overlays_file
assemblytype = painted.assemblytype
```

You can then even replace the switch with a assoc list of `name -> desired airlock type path` to make it even cleaner.

code/modules/research/designs/autolathe_designs.dm
```diff
@@ -195,6 +195,14 @@
 	build_path = /obj/item/floor_painter
 	category = list("initial", "Miscellaneous")
 
+/datum/design/airlock_painter
+	name = "Airlock painter"
+	id = "airlock_painter"
+	build_type = AUTOLATHE
+	materials = list(MAT_METAL = 3000, MAT_GLASS = 1000)
```
moxian: I'm not familiar with this code, but is both this and `materials` var in `/obj/item/airlock_painter` necessary? How are the two different, and what happens if you forget to specify one of them?

code/game/machinery/doors/airlock.dm
SteelSlayer: yeah that is a better option, I'll change it

code/modules/research/designs/autolathe_designs.dm
SteelSlayer: The materials list here is how much it costs to print said item. The materials list in the airlock painter file determines the amount of material that gets added whenever you recycle it into an autolathe, or anything else that will accept it. I think it also determines what's dropped on the floor if the item is destroyed, but not sure on that.

If you forgot to specify them, making these from an autolathe would be free and recycling them would give no materials.

code/game/machinery/doors/airlock.dm
SteelSlayer: I'm honestly not sure how to do what you're suggesting with the assoc lists, but you're first suggestion is definitely an improvement. I'll go with that for now.

moxian: > I'm honestly not sure how to do what you're suggesting with the assoc lists,

Once you have 
```js
var/obj/machinery/door/airlock/painted
switch(paintjob)
  if("Standard")
    painted = new/obj/machinery/door/airlock()
  if("Public")
    painted = new/obj/machinery/door/airlock/public()
  if("Engineering")
    painted = new/obj/machinery/door/airlock/door_assembly_eng()
  ...
icon = painted.icon
overlays_file = painted.overlays_file
assemblytype = painted.assemblytype
```
you can easily transform that to 
```js
var/list/airlocks = new list(
  "Standard" = /obj/machinery/door/airlock,
  "Public" = /obj/machinery/door/airlock/public,
  "Engineering" = /obj/machinery/door/airlock/door_assembly_eng,
  ...
)
var/chosen_path = airlocks[paintjob]
var/obj/machinery/door/airlock/painted = new chosen_path ()

icon = painted.icon
overlays_file = painted.overlays_file
assemblytype = painted.assemblytype
```

And once you have that, it's easy to generate the `var/list/airlocks` by iterating over existing types with `typesof` instead of hardcoding the `name -> type` correspondence

SteelSlayer: The issue is that all the names like "Standard", "Engineering", "Public" etc. are custom names. Those aren't actually the names of those types of airlocks, so I'm going to have to hardcode the list of names/paths for all the paintable airlocks. But even hardcoding them in, assoc lists are way better than what I have currently.

moxian: You can get the non-hardcoded names roughly this way:
```js
var/airlocks = new list()
for(var/airlock_path in typesof(/obj/machinery/door/airlock))
  var/obj/machinery/door/airlock/A = new airlock_path()
  var/obj/structure/door_assembly/AT = A.assemblytype
  var/door_name = AT.base_name
  airlocks[door_name] = airlock_path
```
these `door_name`s would be slightly different from the ones you have currently, but that's probably not a big deal.

You can be fancy and instead of generating this list each time you want to paint a door, you can store it in a new var, or in a datum of its own.
But I'm not sure it's worth the effort.
(I'm also not sure of the performance implication of the `typesof()` search above. I *think* it shouldn't be a problem at all, but I'm not sure. If it is - then hardcoding the options might be better)

SteelSlayer:
Alright, I added a do_after to the painter, takes about 3 seconds to complete a paint job. Also uses an associative list, which saves tons of copy and paste, thanks to @moxian. 

And from a discussion about this on discord, I've made the behavior of the airlock painter similar to the other painters in that you can use it in your hand to select a paint job and then apply it to airlocks without having a pop up menu appear each time.

code/game/machinery/doors/airlock.dm
```diff
@@ -500,6 +520,31 @@ About the new airlock wires panel:
...
+	if(airlock_material == "glass" && W.paint_setting == "Freezer") // Freezer doesn't have a glass version
+		to_chat(user, "<span class='warning'>The freezer paint job can only be applied to non-glass airlocks.</span>")
+		return FALSE
+
+	if((!in_range(src, user) && loc != user))
+		return FALSE
+
+	if(do_after(user, 20, target = src))
+		W.paint()
+		var/airlock_type = paintable_airlocks["[W.paint_setting]"]
```
moxian: If you move paintable_airlocks to be paint gun level var, you won't need to worry (as much) about it being desynced from paint gun's own option_list (since you'd be able to construct the latter from the former), and getting a null airlock_type here (which would at best runtime, and at worst make the door invisible).

Oh and (totally a nitpick!) probably a better name would be something like (airlock?)_(available?)_paintjobs, since this, confusingly, has nothing to do with the .paintable airlock var (that one only controls which doors can be changed away *from*, not which the paint gun can *produce*!)

code/game/machinery/doors/airlock.dm
```diff
@@ -500,6 +520,31 @@ About the new airlock wires panel:
 				sleep(6)
 				update_icon(AIRLOCK_CLOSED)
 
+
+/obj/machinery/door/airlock/proc/change_paintjob(obj/item/airlock_painter/W, mob/user)
+	if(!paintable)
+		to_chat(user, "<span class='warning'>You cannot paint this type of airlock.</span>")
+		return FALSE
+
+	if(airlock_material == "glass" && W.paint_setting == "Freezer") // Freezer doesn't have a glass version
```
moxian: I think it's better to inspect airlock's assembly .glass variable to see whether it has glass now, and indirect assembly .noglass var to see if it can be installed at all (instead of hardcoding freezer).

code/game/machinery/doors/airlock.dm
```diff
@@ -79,6 +98,7 @@ var/list/airlock_overlays = list()
 	var/note_overlay_file = 'icons/obj/doors/airlocks/station/overlays.dmi' //Used for papers and photos pinned to the airlock
 	var/normal_integrity = AIRLOCK_INTEGRITY_N
 	var/prying_so_hard = FALSE
+	var/paintable = TRUE
```
moxian: It's obvious what it does when you inspect the code deeper but probably not when you're skimming through the airlock definition here.
I suggest adding a short comment here. Something like "whether appearance can be charged with airlock painter"

code/game/machinery/doors/airlock.dm
SteelSlayer: You can't use the painter on assemblies, only fully constructed airlocks. Originally I wanted to do that but it would be more annoying. Airlocks that have glass will have the var `airlock_material == "glass"`

code/game/machinery/doors/airlock.dm
SteelSlayer: Both good points

code/game/machinery/doors/airlock.dm
moxian: Oh, you're right, the original assembly is hard to get from here, my bad!

You can still get the noglass as `(new src.assemblytype() ).noglass` (again, instead of hardcoding the freezer)

code/game/machinery/doors/airlock.dm
SteelSlayer: Yeah, will add one

code/game/machinery/doors/airlock.dm
SteelSlayer: Hm, yeah that actually works, neat little trick

code/game/objects/items/devices/airlock_painter.dm
```diff
@@ -0,0 +1,83 @@
...
+
+/obj/item/airlock_painter
+	name = "airlock painter"
+	desc = "An advanced autopainter preprogrammed with several paintjobs for airlocks. Use it on a completed airlock to change its paintjob."
+	icon = 'icons/obj/device.dmi'
+	icon_state = "airlock_painter"
+	item_state = "airlock_painter"
+
+	usesound = 'sound/effects/spray2.ogg'
+
```
Fox-McCloud, line 11: needs conduct and no bludgeon

code/game/machinery/doors/airlock.dm
```diff
@@ -502,6 +503,36 @@ About the new airlock wires panel:
...
+
+	if(!painter.paint_setting)
+		to_chat(user, "<span class='warning'>You need to select a paintjob first.</span>")
+		return
+
+	var/airlock_type = painter.available_paint_jobs["[painter.paint_setting]"] // get the airlock type path associated with the airlock name the user just chose
+	var/obj/machinery/door/airlock/airlock = new airlock_type // we need to create an new instance of the airlock and assembly to read vars from them
+	var/obj/structure/door_assembly/assembly = new airlock.assemblytype
+
+	if(airlock_material == "glass" && assembly.noglass) // prevents painting glass airlocks with a paint job that doesn't have a glass version, such as the freezer
```
Fox-McCloud, line 520: can't you abuse `initial` here?

code/game/machinery/doors/airlock.dm
```diff
@@ -502,6 +503,36 @@ About the new airlock wires panel:
...
+		to_chat(user, "<span class='warning'>This paint job can only be applied to non-glass airlocks.</span>")
+
+	else if(do_after(user, 20, target = src))
+		// applies the user-chosen airlock's icon, overlays and assemblytype to the src airlock
+		painter.paint(user)
+		icon = airlock.icon
+		overlays_file = airlock.overlays_file
+		assemblytype = airlock.assemblytype
```
Fox-McCloud, line 528: and the above two.

code/game/machinery/doors/airlock.dm
```diff
@@ -502,6 +503,36 @@ About the new airlock wires panel:
...
+		update_icon()
+
+	// these are just hanging around but are never placed, we need to delete them.
+	qdel(airlock)
+	qdel(assembly)
```
Fox-McCloud, line 533: that way we don't have to do this nonsense.

code/game/machinery/doors/airlock.dm
SteelSlayer: Yeah it works... somehow. Don't have to `new` them anymore.

code/game/objects/items/devices/airlock_painter.dm
SteelSlayer: Sure

PR #12684, opened 2019-11-02
Reverts regex. Keeps it in code for later to fix

farie82:
**do not merge unless #12690 is not working for some reason**

## What Does This PR Do
Reverts #12474 and #12655
But keeps them in code to fix for later

Fixes: #12676
Fixes: #12657
Fixes: #12656

## Why It's Good For The Game
Currently the chat breaks too often due to it.

:cl:
add: Readded the old highlighting system
del: Removed the regex highlighting. Code is still there but commented out
/:cl:

PR #12690, opened 2019-11-04
The (hopefully) final highlight regex fix

### LINKED ISSUE #12656: Regex breaks chat for others than just you
**Issue Description**:
You can't use multiple words on one line anymore. Not easily atleast.
\bfries|\bhenk|\bmentor|\blipih|\brise
for example will highlight
fries, henk, mentor, lipih and rise
BUT
it'll also do this:
image

If this was only styling then it was okay. But this will cause the whole message to be fucked up even for others who do not use the regex. I also noticed that the mentor ticket system didn't register any of the messages I send.

The ticket itself after having send the messages back:
image

how it looked on their end:
image
Notice that the left name is not clickable. Which normally is the case

To be able to use more than 1 word per line. I have up to 20 words highlighted depending on the role I play.

**What happened instead**:
Had to find a work around which broke the chat

It's a reduction of usability and the breaking of the chat for others can be used in exploits no doubt.

1. Use the text \bfries|\bhenk|\bmentor|\blipih|\brise as highlight
2. Put regex mode on.
3. Have somebody mhelp and have a mentor receive the mhelp.
4. Let the mentor with the regex take the ticket and answer it. Getting a response back as well
5. Having a second mentor/admin observe the chat

Since #12474

**Extra information**:
Dunno but the client regex code being able to break messages is quite bad

### LINKED ISSUE #12657: The highlighting highlights the entire message when a part matches
**Issue Description**:
If you for example put AI as a highlight word you will get a lot of lines highlighted fully.

To only have the string highlighted that you specified

**What happened instead**:
The whole message gets highlighted

This reduces the usability of the highlight system since you can't mark words like AI without getting a lot of false positives on your screen. It was the case like this but then it would just be a small highlight. Not one that screams for your attention and you can't find what string is highlighted.

1. Put AI as a string in a highlight line
2. Say a long ass sentence using ai as part of the string
3. See your chatbox turn yellow

since #12474

### LINKED ISSUE #12676: Highlight strings broken
**Issue Description**: After looking through past PRs, I see a couple were pushed pertaining to highlight strings. Seems nobody tested it because _**THEY DON'T WORK ANYMORE**_.

**What did you expect to happen**: The Highlight strings to either work better, or work period.

**What happened instead**: They don't work _at all_.

**Why is this bad/What are the consequences**: As a player, highlight strings can be useful, as an admin, they're invaluable.

**Steps to reproduce the problem**:N/A

**When did the problem start happening**: #https://github.com/ParadiseSS13/Paradise/pull/12474

craftxbox:
## What Does This PR Do
Fix highlight regex, everything about it, hopefully
Closes #12684 
Closes #12691
Fixes #12676 
Fixes #12657 
Fixes #12656 i think? i havent been able to reproduce it

## Why It's Good For The Game
Fixes this godfuckingdamned cursed motherfucking piece of shit

## Changelog
:cl:
tweak: highlights are split by space again
tweak: remove html matching from regex
fix: properly recurse through the dom to highlight without destroying styling
tweak: make it more obvious and eye-catching when a regex error occurs, detailing what happened and why it stopped working
/:cl:


goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
+			if(next_tag == -1) {
+				s+=replace(work.substring(ind));
+				break;
+			}
+			else if(next_term==-1) {
+				s+=work.substring(ind)
+				break;
+			}
+			else if(next_tag < next_term) {
+				var temp=work_lc.indexOf('>',next_tag)
```
moxian: The html of `one <span foo="bar>baz"> two </span> tree` is technically correct and well-formed, yet your algorithm would mangle it when trying to highlight `baz`.
I *think* we're not using this kind of thing anywhere, but I can't be certain...

As I've mentioned several times already the correct way to handle this to navigate the DOM tree using js-provided capabilities, and not via regex...

But this would probably work in the short term, yes...

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
-				rexp = new RegExp(opts.highlightTerms[i],"gmi")
-			} catch(e){
-				el.innerHTML+='<br/><span style="boldwarning"> Your highlight regex - '+opts.highlightTerms[i]+' - is malformed. Thrown exception: '+e+'</span>'
-				regexHasError = true;
-				return;
+			else {
+				try{
+					new RegExp(opts.highlightTerms[i],"gmi") // check to make sure the pattern wont cause issues
+				} catch(e){
+					el.innerHTML+='<br/><span style="color:#000;background-color:#FFFF00;" class="bold"> Your highlight regex pattern - '+opts.highlightTerms[i]+' - is malformed.<br/>Your highlights have been disabled until they are next edited<br/>Thrown exception: '+e+'</span>'
```
moxian: The user should be getting the error when they are editing the highlights, not when they are receiving a message.

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
 }
 
+function escapeRegex(input){ // put this in a function incase it ever needs to be used elsewhere, makes code cleaner
+	return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
 //Highlights words based on user settings
 function highlightTerms(el) {
 	if(regexHasError) return; //just stop right there ig the regex is gonna except
+	function highlightRecursor(element,term){ //recursor function to do the highlighting proper
```
moxian: Style: rest of the file has a space after coma in argument list.

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
+		var regex = new RegExp("("+term+")","gi")
+		function replace(str) {
+			return str.replace(regex,'<span style="background-color:'+opts.highlightColor+'">$&</span>')
+		}
+		var s=''
```
moxian: Style: rest of file has a space on both sides of assignment (same below)

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
+	return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
 //Highlights words based on user settings
 function highlightTerms(el) {
 	if(regexHasError) return; //just stop right there ig the regex is gonna except
+	function highlightRecursor(element,term){ //recursor function to do the highlighting proper
+		var regex = new RegExp("("+term+")","gi")
+		function replace(str) {
+			return str.replace(regex,'<span style="background-color:'+opts.highlightColor+'">$&</span>')
```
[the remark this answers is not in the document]
moxian: Nevermind, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace mentions `$&`, and does not mention `$0`. I'm fine with either of the two then - up to you.
(I tested both and they behaved exactly the same in practice).

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
 	match.style.background = opts.highlightColor
 }
 
+function escapeRegex(input){ // put this in a function incase it ever needs to be used elsewhere, makes code cleaner
+	return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
 //Highlights words based on user settings
 function highlightTerms(el) {
 	if(regexHasError) return; //just stop right there ig the regex is gonna except
```
moxian: typo. ig -> if

farie82:
I'll keep the revert open just in case. I'll close it once this one is merged

goon/browserassets/js/browserOutput.js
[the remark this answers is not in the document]
craftxbox: this also confused me, $0 has always worked but apparently isn't valid syntax

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
+	function highlightRecursor(element,term){ //recursor function to do the highlighting proper
+		var regex = new RegExp("("+term+")","gi")
```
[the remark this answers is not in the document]
craftxbox: Remnant of testing, literally doesn't do anything because the user can't pick their capture groups nanyways

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
 	match.style.background = opts.highlightColor
 }
 
+function escapeRegex(input){ // put this in a function incase it ever needs to be used elsewhere, makes code cleaner
```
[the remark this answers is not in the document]
craftxbox: I figured ' escape regex' was self explanatory???

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
 //Highlights words based on user settings
 function highlightTerms(el) {
 	if(regexHasError) return; //just stop right there ig the regex is gonna except
+	function highlightRecursor(element,term){ //recursor function to do the highlighting proper
+		var regex = new RegExp("("+term+")","gi")
+		function replace(str) {
+			return str.replace(regex,'<span style="background-color:'+opts.highlightColor+'">$&</span>')
+		}
+		var s=''
+		var work=element.innerHTML;
```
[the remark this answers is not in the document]
craftxbox: bah, nitpicking

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
+		var work_lc=work.toLowerCase()
+		var ind=0;
+		while(ind < work_lc.length) {
+			console.log(s);
+			var next_term=work_lc.substring(ind).search(regex);
+            if(next_term!=-1)next_term+=ind;
```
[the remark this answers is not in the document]
craftxbox: indentation happens, but how the fuck and why are braceless ifs discouraged

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
+		var regex = new RegExp("("+term+")","gi")
+		function replace(str) {
+			return str.replace(regex,'<span style="background-color:'+opts.highlightColor+'">$&</span>')
+		}
+		var s=''
+		var work=element.innerHTML;
+		var work_lc=work.toLowerCase()
+		var ind=0;
+		while(ind < work_lc.length) {
+			console.log(s);
```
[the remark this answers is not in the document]
craftxbox: whoops, .log doesn't even work here

goon/browserassets/js/browserOutput.js
```diff
@@ -171,25 +171,67 @@ function setHighlightColor(match) {
...
 function highlightTerms(el) {
 	if(regexHasError) return; //just stop right there ig the regex is gonna except
+	function highlightRecursor(element,term){ //recursor function to do the highlighting proper
+		var regex = new RegExp("("+term+")","gi")
+		function replace(str) {
+			return str.replace(regex,'<span style="background-color:'+opts.highlightColor+'">$&</span>')
+		}
+		var s=''
+		var work=element.innerHTML;
+		var work_lc=work.toLowerCase()
```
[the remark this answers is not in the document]
craftxbox: I believe this is just a leftover from when the recursor only did plaintext matches

goon/browserassets/js/browserOutput.js
craftxbox: Your highlight is invalid x10000

goon/browserassets/js/browserOutput.js
craftxbox: stray braces will NEVER appear, ever unless there's a security vulnerability, or an admin is being an idiot, theyre all filtered out already at the server side

goon/browserassets/js/browserOutput.js
craftxbox: reeeeeeeee

goon/browserassets/js/browserOutput.js
craftxbox: reeeeeeeee

goon/browserassets/js/browserOutput.js
craftxbox: it's... a comment

goon/browserassets/js/browserOutput.js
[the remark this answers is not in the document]
moxian: Here's how: https://www.imperialviolet.org/2014/02/22/applebug.html

Also see this stackoverflow question (though it's mostly the same thing): https://stackoverflow.com/questions/2125066/is-it-a-bad-practice-to-use-an-if-statement-without-curly-braces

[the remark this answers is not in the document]
craftxbox: Dont bring apple into this, we all know they're idiots already.
the SO question was closed for opinionated, the only reason it would be bad would be people being absolute idiots, which isnt a reason to not use it.

goon/browserassets/js/browserOutput.js
craftxbox: I actually tested this and I haven't been able to cause this cause a problem.

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +166,79 @@ function emojiparse(el) {
...
 //Highlights words based on user settings
 function highlightTerms(el) {
+
 	if(regexHasError) return; //just stop right there ig the regex is gonna except
+
+	function highlightRecursor(element,term){ //recursor function to do the highlighting proper
+		var regex = new RegExp("("+term+")","gi")
+
+		function replace(str) {
+			return str.replace(regex,'<span class="highlight" style="background-color:'+opts.highlightColor+'">$&</span>')
```
Crazylemon64, line 190: Would the "highlight" span class's rules get overridden by the explicit color set here

goon/browserassets/js/browserOutput.js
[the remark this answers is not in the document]
Crazylemon64: I find it a bit ambiguous - some more detail could be helpful for rapidly understanding what it's supposed to do.

goon/browserassets/js/browserOutput.js
Crazylemon64: don't ignore style advice and mark it as "resolved"

goon/browserassets/js/browserOutput.js
Crazylemon64: Isn't this missing a semicolon? I'd expect this to fail to run

goon/browserassets/js/browserOutput.js
Crazylemon64: are you going to actually do anything about it
it looks like around 1107 would be a good place to do that actual check

goon/browserassets/js/browserOutput.js
craftxbox: No, the highlight class will still follow the css rules, however inline styling will override  a stylesheet for that element, but nnot the rest of the class

goon/browserassets/js/browserOutput.js
craftxbox: Javascript doesnt particularly care about colons, unless theres a case in where the separation between two statements can be ambiguous, ex:
```
var a = "a"
var b = "b"
```
is okay, but it will throw an error if you do:
```
var a = "a" var b = "b"
```
unless theres a semicolon between the two statements.

goon/browserassets/js/browserOutput.js
craftxbox: The `internalOutput` at 1108 causes this check to be done already, however if a saved invalid regex is in options at the time the chat loads checking only in edit wouldnt catch that and cause the chat panel to fail to load.

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +166,79 @@ function emojiparse(el) {
...
 }
 
-// Colorizes the highlight spans
-function setHighlightColor(match) {
-	match.style.background = opts.highlightColor
+// Recolorizes the highlight spans
+function setHighlightColor() {
+	var highlightspans = document.getElementsByClassName("highlight")
+	for(var i in highlightspans){
+		highlightspans[i].setAttribute("style","background-color:"+opts.highlightColor)
```
Crazylemon64, line 173: Let's see, you do it this way because the js interface for modifying existing css rules sucks?

craftxbox: Yep, from experience `.style` will occasionally either do nothing, or be undefined even when it IS defined on the element.

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +166,79 @@ function emojiparse(el) {
...
-				rexp = new RegExp(opts.highlightTerms[i],"gmi")
-			} catch(e){
-				el.innerHTML+='<br/><span style="boldwarning"> Your highlight regex - '+opts.highlightTerms[i]+' - is malformed. Thrown exception: '+e+'</span>'
-				regexHasError = true;
-				return;
+			else {
+				try{
+					new RegExp(opts.highlightTerms[i], "gmi"); // check to make sure the pattern wont cause issues
+				} catch(e){
+					el.innerHTML += '<br/><span style="color:#000;background-color:#FFFF00;" class="bold"> Your highlight regex pattern -- ' + opts.highlightTerms[i] + ' -- is malformed.<br/>Your highlights have been disabled until they are next edited<br/>Thrown exception: '+e+'</span>';
```
AffectedArc07, line 236: Cant you use `internalOutput()` here instead of appending straight onto the page?

PR #12691, opened 2019-11-04
Fix regex highlightning.

### LINKED ISSUE #12656: Regex breaks chat for others than just you

### LINKED ISSUE #12657: The highlighting highlights the entire message when a part matches

### LINKED ISSUE #12676: Highlight strings broken

moxian:
## What Does This PR Do
Fixes highlighting that was broken with the recent highlighting regex changes.
Conflicts with/reimplements #12690 (and, transitively, #12684)

Fixes #12676
Fixes #12657
Fixes #12656

You see, Cunningham's Law is true, as always - "the best way to ~~get the right answer~~ get a feature implemented on the internet is not to ~~ask a question~~ create a bug; it's to post the ~~wrong answer~~ suboptimal PR"
But that's, of course, a matter of personal taste - feel free to close this one if you like the other PR handles it better.

The prime difference is that this PR operates on DOM entities rather than on raw HTML, which makes it more robust to the weird layouts that you cannot really parse with a regex. Other changes (including the regex check being moved to when it's saved over when it's used) are more or less superficial.

Credits to @craftxbox for the original implementation.

## Why It's Good For The Game
I'm not sure why regex highlights are needed, but you currently get broken HTML if you use the feature at all, and that's bad user experience.

## Images of changes
dreamseeker_2019-11-04_08-00-48

## Changelog
:cl:
fix: fixed highlighting messing up the chat
/:cl:

### DIFF
--- a/goon/browserassets/js/browserOutput.js
+++ b/goon/browserassets/js/browserOutput.js
@@ -159,7 +159,7 @@ function escapeRegex(input){
 // Recursively highlights all text matching `term` in text nodes reachable from the given root `element`.
 // Returns the last (right-most, DOM-wise) element that was analyzed and potentially highglighted.
 function highlightRecursively(element, term) {
-	var regex = new RegExp(term, "gi")
+	var regex = new RegExp(term, "gi");
 	// if that's a text node - wrap found words in a highlighter span.
 	if (element.nodeType == Node.TEXT_NODE) {
 		// The underlying text element would (likely) be destroyed, and we'll be left with
@@ -179,20 +179,20 @@ function highlightRecursively(element, term) {
 			lastStop = newStop;
 		}
 		// Push the rest of the string.
-		newElements.push(document.createTextNode(element.textContent.substring(lastStop)))
+		newElements.push(document.createTextNode(element.textContent.substring(lastStop)));
 
 		// If we found any matches - replace the current text element with the bunch of new nodes we've just created
 		var lastInserted = element;
 		if (newElements.length > 0) {
 			for (var i = 0; i < newElements.length; i++){
-				lastInserted = element.parentNode.insertBefore(newElements[i], element)
+				lastInserted = element.parentNode.insertBefore(newElements[i], element);
 			}
 			element.parentNode.removeChild(element);
 		}
 		return lastInserted;
 	} else {
 		// If it's not a plain text node - recurse further.
-		var currentChild = element.childNodes[0]
+		var currentChild = element.childNodes[0];
 		while (currentChild){
 			currentChild = highlightRecursively(currentChild, term);
 			currentChild = currentChild.nextSibling;
@@ -1065,7 +1065,7 @@ $(function() {
 			count++;
 		}
 
-		opts.highlightRegexEnable = document.querySelector("#highlightRegexEnable").checked
+		opts.highlightRegexEnable = document.querySelector("#highlightRegexEnable").checked;
 		if (opts.highlightRegexEnable) {
 			// Check that all the patterns are valid regexes.
 			for(var i = 0; i < opts.highlightTerms.length; i++){
--- a/goon/browserassets/js/browserOutput.js
+++ b/goon/browserassets/js/browserOutput.js
@@ -72,8 +72,6 @@ var opts = {
 	'enableEmoji': true
 };
 
-var regexHasError = false; //variable to check if regex has excepted 
-
 function outerHTML(el) {
     var wrap = document.createElement('div');
     wrap.appendChild(el.cloneNode(true));
@@ -90,19 +88,6 @@ if (!Date.now) {
 if (typeof String.prototype.trim !== 'function') {
 	String.prototype.trim = function () {
 		return this.replace(/^\s+|\s+$/g, '');
-	};
-}
-
-//Polyfill for string.prototype.includes. Why the fuck. Just why the fuck.
-if (!String.prototype.includes) {
-	String.prototype.includes = function(search, start) {
-	  'use strict';
-  
-	  if (search instanceof RegExp) {
-		throw TypeError('first argument must not be a RegExp');
-	  } 
-	  if (start === undefined) { start = 0; }
-	  return this.indexOf(search, start) !== -1;
 	};
 }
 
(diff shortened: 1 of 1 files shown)


goon/browserassets/js/browserOutput.js
```diff
@@ -1042,15 +1073,30 @@ $(function() {
...
+		opts.highlightRegexEnable = document.querySelector("#highlightRegexEnable").checked;
+		if (opts.highlightRegexEnable) {
+			// Check that all the patterns are valid regexes.
+			for(var i = 0; i < opts.highlightTerms.length; i++){
+				if (opts.highlightTerms[i]){
+					try {
+						new RegExp(opts.highlightTerms[i], "gmi");
+					} catch(e) {
+						var message = "Your highlight pattern '" + opts.highlightTerms[i] + "' is not a valid regex. Regex mode was automatically disabled.";
+						opts.highlightRegexEnable = false;
```
craftxbox: The feature shouldnt literally disable itself when it encounters an error, this can also cause issue with the option enable being desyncronized from the actual value in cookies.

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +151,76 @@ function emojiparse(el) {
...
-			var rexp;
-			try{
-				rexp = new RegExp(opts.highlightTerms[i],"gmi")
-			} catch(e){
-				el.innerHTML+='<br/><span style="boldwarning"> Your highlight regex - '+opts.highlightTerms[i]+' - is malformed. Thrown exception: '+e+'</span>'
-				regexHasError = true;
-				return;
+		if (opts.highlightTerms[i]) {
+			if (!opts.highlightRegexEnable) {
+				var innerTerms = opts.highlightTerms[i].split(" ")
```
craftxbox: stolen from my pr, again

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +151,76 @@ function emojiparse(el) {
 	}
 }
 
-// Colorizes the highlight spans
-function setHighlightColor(match) {
```
craftxbox: You broke this by smashing its function, this will except when the user tries to set their highlight color

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +151,76 @@ function emojiparse(el) {
...
-	match.style.background = opts.highlightColor
+// Escape any special regex characters in a given string, so that one can match it literally.
+function escapeRegex(input){
```
craftxbox: carbon copy from my pr.

goon/browserassets/js/browserOutput.js
```diff
@@ -93,19 +91,6 @@ if (typeof String.prototype.trim !== 'function') {
 	};
 }
 
-//Polyfill for string.prototype.includes. Why the fuck. Just why the fuck.
```
craftxbox: Why remove the includes polyfill?

Couls:
Highlightning

goon/browserassets/js/browserOutput.js
moxian: Disabling the feature is the least bad behavior imo.
Feature being desynchronized from the cookies is possible, yes, but only realistic during the transition period. Reopening the highlights dialog would fix it (by disabling the regex), and you'd never be bothered with it again.

goon/browserassets/js/browserOutput.js
moxian: I was basing this on your PR, correct. I didn't need to, but this is how it happened.

goon/browserassets/js/browserOutput.js
moxian: This function was never used, which is why I deleted it.
Changing color still works
dreamseeker_2019-11-04_19-42-24

goon/browserassets/js/browserOutput.js
moxian: Correct.

goon/browserassets/js/browserOutput.js
craftxbox: opts is meant for the user's preferences, and generally shouldnt be touched to preserve the user's actual preferences

goon/browserassets/js/browserOutput.js
moxian: It is not needed.
It was added in #12655 
String.includes call is removed in both this PR and in your #12690

goon/browserassets/js/browserOutput.js
craftxbox: that's probably a bug, i presume it was meant to retroactively recolor previous highlights.

goon/browserassets/js/browserOutput.js
craftxbox: it still was a nice thing to have regardless
~~babel on text panel when~~

moxian:
> you should reconsider waiting for people to post their pr to post your counterpr, it's really fucking infuriating, especially when you nitpick their pr first.

Sorry about nitpicks - I did mention that they are small and mostly insignificant.
As for the counterpr, well, I did try to convince you to implement things this way since at least https://github.com/ParadiseSS13/Paradise/pull/12474#discussion_r332174242 (so, for a month now). I hoped your fix in #12690 would finally change this approach to work with DOM, but since it didn't, I filed this my PR instead.

> Highlightning
I'm pathologically unable to spell that word. I'm very sorry.

goon/browserassets/js/browserOutput.js
moxian: Let's not bring the entire js ecosystem into paradise just because it's "nice to have"

craftxbox:
> > Highlightning

> I'm pathologically unable to spell that word. I'm very sorry.

highlightininighitnigiting

> I hoped your fix in #12690 would finally change this approach to work with DOM, but since it didn't, I filed this my PR instead.

Actually going through the entire dom is alot slower than just doing it by characters in the innerhtml, and, as far as i can tell, doesnt cause issue with any valid html. ( `<span class="a>b">` is actually not valid (should be `&gt;` NOT `>`), and would resolve to `<span class=&quot;a>b&quot;&gt;` as far as i can tell )

goon/browserassets/js/browserOutput.js
moxian: No, it's not.
Highlights are only applied on new messages. Have always been.
I think it's intentional, as going through all the logs would be fairly slow (and since you probably don't really care about highlights on the old messages).

goon/browserassets/js/browserOutput.js
craftxbox: I know, its just a joke. Anyways, i could see includes being used in other places (the whole admin filter bit could probably be rewritten using it)

moxian:
> ( <span class="a>b"> is actually not valid (should be &gt; NOT >), and would resolve to <span class=&quot;a>b&quot;&gt; as far as i can tell )

Both are valid. Both are treated as `<span class="a>b">` for the purpose of innerHTML matching. i.e. you'll find `>` inside the tag.
See this demo: https://jsfiddle.net/8jxqtcyn/

craftxbox:
> Both are valid. [...]

Either way, that sort of situation will not, or, should not ever occur, ever, at all

goon/browserassets/js/browserOutput.js
craftxbox: it isnt a thing that happens often enough for speed to matter much anyways.

goon/browserassets/js/browserOutput.js
Crazylemon64: using the `regexHasError` system here to do this without desyncing the opts object from the cookie might be a good thing to do
as well as intercepting the saving of the patterns to the cookie if they're invalid

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +151,76 @@ function emojiparse(el) {
...
+			if (match[0].length == 0) {
+				// Matching on .* or other non-consuming pattern would get us here
+				break;
+			}
+			var start = match.index;
+			var newStop = match.index + match[0].length;
+			// push text between the end of the last match and start of this one
+			newElements.push(document.createTextNode(element.textContent.substring(lastStop, start)));
+			// wrap the match in a span, and push the span
+			var span = document.createElement("span");
```
Crazylemon64, line 183: I would advocate using a proper class here so that it would be easy to restyle it with css later, to allow restyling old highlights

goon/browserassets/js/browserOutput.js
```diff
@@ -166,30 +151,76 @@ function emojiparse(el) {
...
+function highlightRecursively(element, term) {
+	var regex = new RegExp(term, "i");
+	// if that's a text node - wrap found words in a highlighter span.
+	if (element.nodeType == Node.TEXT_NODE) {
+		// The underlying text element would (likely) be destroyed, and we'll be left with
+		// a collection of smaller text nodes and highlighter spans. They'd be stored in this array.
+		var newElements = [];
+		var lastStop = 0;  // The end of the last found match.
+		// The loop is ugly.
+		// There are reasons for it, namely: BYOND does not support matchAll regex method, and the `regexp.exec()` which is
```
Crazylemon64, line 170: What exactly do you mean by "BYOND" here? We're dealing with an IE browser instance, right?

moxian:
I've just checked this, and this PR does not quite work on latest master for some reason.
Namely, it highlights the word normally, but fails to do so if the word occurs more than once (none of the occurrences get highlighted). I've also had issues with `say` not actually saying anything, but that might be completely unrelated (i didn't check if that's the problem on master with similar setup).

I'm afraid I wouldn't have time (or motivation) to fix this in the near future (think, this week); I'd advise against merging this in the meantime.
And feel free to close this outright.

PR #12398, opened 2019-09-23
Makes Xenobiology Simpler. Removes Slimes

Fox-McCloud:
and adds simple mob slimes

😛 

From TG.

This is mostly just a refactor; there's not really all that many changes here; this just kills off stupid carbon based slimes and converts them over to simple mobs, as they should be.

Slimes now have a HUD and actions buttons; the HUD is pretty simple; it's really just a health icon

fixes: https://github.com/ParadiseSS13/Paradise/issues/11977

Actual changes:
- Slimes use action buttons instead of verbs
- Using a Docility potion on a slime that's rabid will make it not rabid (but not docile)
- Using bloodlust potion on a slime that's docile will make it not docile (but not rabid)

:cl: Fox McCloud
tweak: Slimes are a simple mob subtype
add: using a bloodlust slime potion on a docile slime removes the docility
add: using a docility potion on a rabid slime calms them down
tweak: slime core removal surgery steps are now just scalpel->hemostat
add: Slimes have action buttons to feed on, evolve, or reproduce
fix: Fixes beepsky arresting slimes (maybe it should be a feature....)
/:cl:

PR #12696, opened 2019-11-04
Makes simple slimes ignore slime people when searching for targets again

### LINKED ISSUE #12518: Slimes eat slime people
Xenobiology slimes would previously not attack slime people.

This was changed with an undocumented tweak introduced in #12398

https://github.com/ParadiseSS13/Paradise/pull/12398/files#diff-3b7ff7acf97c44b0c42afd443293685cL328

S Q U I S H

SteelSlayer:
## What Does This PR Do
Simple slimes will ignore slime people when searching for targets, as they used to do.

EDIT: Updated to include species `inherent_factions` and gives slime people the "slime" faction. Since simple slimes have this faction, they will never attack other simple slimes, or slime people.

## Why It's Good For The Game
Restores a gameplay mechanic that was removed in the slime refactor (#12398) as an undocumented change.

Fixes #12518

## Changelog
:cl:
add: Adds in an inherent_factions var to all species. Gives slime people the "slime" faction
fix: Slimes will ignore slime people when searching for targets again
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/slime/life.dm
+++ b/code/modules/mob/living/simple_animal/slime/life.dm
@@ -314,7 +314,7 @@
 
 				for(var/mob/living/L in view(7,src))
 
-					if(isslime(L) || isslimeperson(L) || L.stat == DEAD) // Ignore other slimes, slime people, and dead mobs
+					if(L.stat == DEAD) // Ignore dead mobs
 						continue
 
 					if(L in Friends) // No eating friends!
--- a/code/modules/mob/living/simple_animal/slime/life.dm
+++ b/code/modules/mob/living/simple_animal/slime/life.dm
@@ -314,7 +314,7 @@
 
 				for(var/mob/living/L in view(7,src))
 
-					if(isslime(L) || L.stat == DEAD) // Ignore other slimes and dead mobs
+					if(isslime(L) || isslimeperson(L) || L.stat == DEAD) // Ignore other slimes, slime people, and dead mobs
 						continue
 
 					if(L in Friends) // No eating friends!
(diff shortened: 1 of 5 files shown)


Mitchs98:
Yay. *squish

Fox-McCloud:
Instead of being a ❄️ why not make it so that you use something like TG's `inherent_factions` list.

SteelSlayer:
I'll look into it

Mitchs98:
With any luck this'll get merged soon 👀

Fox-McCloud:
You'll have to alter pod diona and shadow people, as well.

Then this should be ready to go.

SteelSlayer:
Alright, just confirming this is what you want.. I looked up what factions they have on tg. For pod people, they'll get the "plants" and "vines" faction, and for shadow's, they'll get the "faithless" faction, is that right?

code/modules/mob/living/simple_animal/slime/life.dm
```diff
@@ -314,7 +314,7 @@
 
 				for(var/mob/living/L in view(7,src))
 
-					if(isslime(L) || L.stat == DEAD) // Ignore other slimes and dead mobs
+					if(L.stat == DEAD) // Ignore dead mobs
```
Fox-McCloud, line 317: out of curiosity, why did you remove this?

SteelSlayer: Further down on line 327, the slime checks if the target mob has one of its own factions. Simple animal slimes get the "slime" faction which would mean they ignore other slimes. The check is redundant in this case.

PR #12697, opened 2019-11-05
Fix odysseus syringes breaking when flying beyond max range.

### LINKED ISSUE #7132: Syringes that don't hit from a Odysseus turn invisible and break
If you shoot a syringe and it doesn't properly hit (or perhaps hits a hardsuit or armor) it turns invisible and can no longer be properly used. You can still see the chem, however, and pick it up, but it doesn't seem to be possible to do anything with it.

### LINKED ISSUE #12685: Oddyseus Syringe Gun Syringes Usually Don't Work Properly
**Issue Description**:

The syringes fired from an Oddyseus' exosuit syringe gun usually do not work, resulting in a vertical syringe laying on the ground still filled with reagents.

Syringes fired from an oddyseus should have the same effect as the ones from any other syringe gun.

**What happened instead**:

That for whatever reason the projectiles don't inject and use a weird icon.

It really makes a lot of interesting things that could be possible with the Oddyseus much less feasible.

Step 1. Get oddyseus
Step 2. Get some reagents analyzed, and syringes loaded
Step 3. Process reagents, and fire a syringe at something
Step 4. Notice how the syringe is vertical and filled with reagents still, and the person you hit with it has none of the reagents from it in them.

I'm not sure - whenever I've tried using an oddy, this has happened.

**Extra information**:

Sometimes the syringes work properly - they leave an empty normal syringe on the ground. The broken syringes still have a small line indicating that they have chems in them.

moxian:
## What Does This PR Do
Related: #12685 
Related: #7132 
Late edit: According to https://github.com/ParadiseSS13/Paradise/issues/12685#issuecomment-552586988 there is another case of syringes breaking that I did not explicitly test against. I think this PR should handle that as well, but not 100% confident.

Prior to this PR, the syringe gun icon would be reverted to horizontal only if:
 - it hits someone
 - it hits a wall and can't make any progress towards its destination
 - it reaches its destination

but it would stay in its projectile (vertical) icon if the target you are aiming for is out of range (i.e. 7 tiles away)

## Why It's Good For The Game
Bugfix. Syringes are now reusable.

## Images of changes
N/A

## Changelog
:cl:
fix: fixed syringe fired from odysseus syringe gun sprite being broken when flying beyond max range.
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/medical_tools.dm
+++ b/code/game/mecha/equipment/tools/medical_tools.dm
@@ -316,40 +316,38 @@
 	var/mob/originaloccupant = chassis.occupant
 	spawn(0)
 		src = null //if src is deleted, still process the syringe
-		for(var/i=0, i<6, i++)
+		var/max_range = 6
+		for(var/i=0, i<max_range, i++)
 			if(!mechsyringe)
 				break
-			if(step_towards(mechsyringe,trg))
-				var/list/mobs = new
-				for(var/mob/living/carbon/M in mechsyringe.loc)
-					mobs += M
-				var/mob/living/carbon/M = safepick(mobs)
-				if(M)
-					var/R
-					mechsyringe.visible_message("<span class=\"attack\"> [M] was hit by the syringe!</span>")
-					if(M.can_inject(null, TRUE))
-						if(mechsyringe.reagents)
-							for(var/datum/reagent/A in mechsyringe.reagents.reagent_list)
-								R += A.id + " ("
-								R += num2text(A.volume) + "),"
-						add_attack_logs(originaloccupant, M, "Shot with [src] containing [R], transferred [mechsyringe.reagents.total_volume] units")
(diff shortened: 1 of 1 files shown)


code/game/mecha/equipment/tools/medical_tools.dm
```diff
@@ -316,40 +316,38 @@
 	var/mob/originaloccupant = chassis.occupant
 	spawn(0)
 		src = null //if src is deleted, still process the syringe
-		for(var/i=0, i<6, i++)
+		var/max_range = 6
+		for(var/i=0, i<max_range, i++)
 			if(!mechsyringe)
 				break
-			if(step_towards(mechsyringe,trg))
```
moxian: github shows lots of changes here, but they are all whitespace. I.e.
```js
if(step_towards(mechsyringe,trg))
  // do stuff
```
changed to 
```js
if(!step_towards(mechsyringe,trg))
  break
// do stuff
```

Toggling "hide whitespace changes" should help make it more readable.

PR #12708, opened 2019-11-07
Corpses on the shuttle no longer prevent hijack greentext.

### LINKED ISSUE #12641: Hijack bugged
**Issue Description**: Hijack seems to be bugged with its parameters. 

So it doesn't seem to correctly make checks on the shuttle and whether it was hijacked or not. I had seen one round where only one person on the shuttle wasnt a syndicate and they were ziptied (note: they were standing in case that needs to be apart of it). That incident was 2 days ago. And just yesterday I had hijacked the shuttle and i made sure that no one else was on the shuttle and every ghost agreed I was the only one present. I will say though both times I am reffering to the hijacker had a holo parasite. Im wondering if the holo parasite is bugged so that he isnt counted as syndicate and instead is counted as crew. 

Greentext.

**What happened instead**:

red text

Because it makes it so that your objective was uncompletable. and it makes people think that you didnt hijack. 

try hijacking the shuttle with a holo parasite present and out and see if it counts as a proper hijack. 

I dont know. Maybe this is a bug thats existed for a long time. maybe its only been since the last merge. 

**Extra information**:

### LINKED ISSUE #12672: Demons and revenants count against hijack
**Issue Description**:

If you have the hijack objective and are the only person on the shuttle but a demon or revenant are also on the shuttle while not revealing themselves, they count against the hijack objective and you cannot do anything about it.

For the hijacker to get greentext when no one else was on the shuttle besides them and the revenant/demon.

**What happened instead**:

The hijack failed.

The hijacker has nothing they can do to prevent it as the revenant/demon is invulnerable and invisible when not revealed.

Have hijack, emag the console, have a demon/revenant on the shuttle too.  Hijack fails.

Unknown as its a very rare circumstance.

**Extra information**:

moxian:
## What Does This PR Do
Corpses on the shuttle no longer prevent hijack greentext.
Slaughter demons and other special roles don't count against it either.
Instead of having a long list of roles that are allowed on the shuttle, it's easier to list *dis*allowed ones instead.
Probably fixes #12641
Fixes #12672

The live (ha!) code checks for aliveness as  `if(player.stat != DEAD || !iszombie(player))`, which.. happens to be true for corpses. It probably meant to be `if(player.stat != DEAD && !iszombie(player))`. So I fixed that.

The special role check is long and tries to list every possible traitor-relevant role as "acceptable". But really all the special roles are traitor-relevant anyway. What matters is that no crew escapes!
The only ones [from the list](https://github.com/ParadiseSS13/Paradise/blob/8b9f6632038bbc98f93ce1c0740c3d7ea5f0f9cc/code/__DEFINES/gamemode.dm#L19) that I can see as arguable are revolutionaries (but I'm happy to have them count against hijack, if that is the decision).  

Note: **I did not test the changes**, since that requires a fairly involved setup. However, I hope the logic is straightforward enough that simply reading the code would suffice.

## Why It's Good For The Game
Greentext good.

## Images of changes
N/A

## Changelog
:cl:
fix: corpses on the shuttle no longer count against hijack objective.
tweak: slaughter demons, borers, wizards, xenomorphs, etc on the shuttle no longer count against hijack objective
/:cl:

### DIFF
--- a/code/modules/shuttle/emergency.dm
+++ b/code/modules/shuttle/emergency.dm
@@ -174,36 +174,37 @@
 
 /obj/docking_port/mobile/emergency/proc/is_hijacked()
 	for(var/mob/living/player in GLOB.player_list)
-		if(player.mind)
-			if(player.stat == DEAD)  // Corpses
+		if(!player.mind)
+			continue
+		if(player.stat == DEAD)  // Corpses
+			continue
+		if(iszombie(player))  // Walking corpses
+			continue
+		if(issilicon(player)) //Borgs are technically dead anyways
+			continue
+		if(isanimal(player)) //Poly does not own the shuttle
+			continue
+		if(ishuman(player)) //hostages allowed on the shuttle, check for restraints
+			var/mob/living/carbon/human/H = player
+			if(!H.check_death_method() && H.health <= HEALTH_THRESHOLD_DEAD) //new crit users who are in hard crit are considered dead
 				continue
-			if(iszombie(player))  // Walking corpses
+			if(H.handcuffed) //cuffs
 				continue
-			if(issilicon(player)) //Borgs are technically dead anyways
+			if(H.wear_suit && H.wear_suit.breakouttime) //straight jacket
 				continue
-			if(isanimal(player)) //Poly does not own the shuttle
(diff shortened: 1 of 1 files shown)


Couls:
oops

PR #12711, opened 2019-11-07
Chemistry bag QoL improvement

SteelSlayer:
## What Does This PR Do
Makes the chemistry bag able to transfer pills and patches to the all-in-one grinder.

EDIT: Staring at all the double indentation in this file gave me a headache so I changed all the double indentation to singles in this proc. The actual new/modified code is on line 177-202. Also this has the added benefit of allowing the plant bag transfer certain items that weren't transferable before, such as nettles and novaflowers.

## Why It's Good For The Game
Useful when you accidentally make the wrong dosage on pills/patches. If you want to grind them back down to liquid form and try again, you have to place them 1 by 1 into the grinder. Now you can easily transfer a bulk amount from the chemistry bag just like you can when using the plant bag.

## Changelog
:cl:
tweak: Chemistry bags can now transfer their pills and patches to all-in-one grinders
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/machinery/reagentgrinder.dm
+++ b/code/modules/reagents/chemistry/machinery/reagentgrinder.dm
@@ -134,76 +134,88 @@
 		icon_state = "juicer0"
 
 /obj/machinery/reagentgrinder/attackby(obj/item/I, mob/user, params)
-		if(default_unfasten_wrench(user, I))
-				return
+	if(default_unfasten_wrench(user, I))
+		return
 
-		if(exchange_parts(user, I))
-				return
+	if(exchange_parts(user, I))
+		return
 
-		if(anchored && !beaker)
-				if(default_deconstruction_screwdriver(user, "juicer_open", "juicer0", I))
-						return
+	if(anchored && !beaker)
+		if(default_deconstruction_screwdriver(user, "juicer_open", "juicer0", I))
+			return
 
-				if(panel_open && istype(I, /obj/item/crowbar))
-						default_deconstruction_crowbar(I)
-						return
+		if(panel_open && istype(I, /obj/item/crowbar))
+			default_deconstruction_crowbar(I)
+			return
 
-		if (istype(I, /obj/item/reagent_containers) && (I.container_type & OPENCONTAINER) )
-				if(beaker)
-						to_chat(user, "<span class='warning'>There's already a container inside.</span>")
-				else if(panel_open)
-						to_chat(user, "<span class='warning'>Close the maintenance panel first.</span>")
-				else
-						if(!user.drop_item())
-								return 1
-						beaker =  I
-						beaker.loc = src
-						update_icon()
-						src.updateUsrDialog()
-				return 1 //no afterattack
+	if(istype(I, /obj/item/reagent_containers) && (I.container_type & OPENCONTAINER) )
+		if(beaker)
+			to_chat(user, "<span class='warning'>There's already a container inside.</span>")
+		else if(panel_open)
+			to_chat(user, "<span class='warning'>Close the maintenance panel first.</span>")
+		else
+			if(!user.drop_item())
+				return TRUE
+			beaker =  I
+			beaker.loc = src
+			update_icon()
+			updateUsrDialog()
+		return TRUE //no afterattack
 
-		if(is_type_in_list(I, dried_items))
-				if(istype(I, /obj/item/reagent_containers/food/snacks/grown))
-						var/obj/item/reagent_containers/food/snacks/grown/G = I
-						if(!G.dry)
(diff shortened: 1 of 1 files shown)


Mitchs98:
You're literally space jesus.

code/modules/reagents/chemistry/machinery/reagentgrinder.dm
```diff
@@ -176,15 +176,15 @@
 		//Fill machine with a bag!
 		if(istype(I, /obj/item/storage/bag))
 				var/obj/item/storage/bag/B = I
-				for (var/obj/item/reagent_containers/food/snacks/grown/G in B.contents)
+				for (var/obj/item/reagent_containers/food/G in B.contents)
```
moxian: You fix it for a single use case of pills, but there's still a lot of stuff that can fit into grinder, but can't be moved from bags (honeycomb, minerals, weird plants such as nettles).
I suggest putting a more generic check instead. Something like:
```js
for(var/item in B.contents)
  if(!is_type_in_list(item , blend_items) && !is_type_in_list(item, juice_items))
    continue
  // do the item move logic
```
This doesn't handle dried items well, though, since they need the dry check.. Except it does, since all the dried items happen to be a subtype of /food/snacks/grown, so they happen to be handled already... Except if you transfer them via a bag, you don't need to dry them... But at least you are not introducing a new bug here... But I digress... (i couldn't figure out a way to substantially improve this code, so this is probably fine?)

code/modules/reagents/chemistry/machinery/reagentgrinder.dm
```diff
@@ -176,15 +176,15 @@
...
 						if(holdingitems && holdingitems.len >= limit) //Sanity checking so the blender doesn't overfill
 								to_chat(user, "<span class='notice'>You fill the All-In-One grinder to the brim.</span>")
 								break
 
 				if(!I.contents.len)
-						to_chat(user, "<span class='notice'>You empty the plant bag into the All-In-One grinder.</span>")
+						to_chat(user, "<span class='notice'>You empty [I] into the All-In-One grinder.</span>")
 
 				src.updateUsrDialog()
 				return 1
```
moxian: Should this be `0`? Should it be `0` iff at least one item got transferred?

code/modules/reagents/chemistry/machinery/reagentgrinder.dm
SteelSlayer: Probably a better thing to do yeah. Honestly looking at the grinder code makes me want to just refactor/clean up the whole thing, lots of ugly crap in here but that's out of the scope of the PR.

moxian: > but that's out of the scope of the PR
I agree. Feel free to simply resolve this comment if that's not something you want to tackle at the moment.

code/modules/reagents/chemistry/machinery/reagentgrinder.dm
SteelSlayer: I'm honestly not even sure how that return statement is there and the for loop doesn't terminate after the first item. I think it should be one tab to the left and it should be return TRUE.

code/modules/reagents/chemistry/machinery/reagentgrinder.dm
SteelSlayer: I was referring to a complete grinder refactor/clean up being out of the scope. Your changes are a good idea, and I will be implementing them. :+1:

code/modules/reagents/chemistry/machinery/reagentgrinder.dm
```diff
@@ -134,76 +134,88 @@
...
+	if(holdingitems && holdingitems.len >= limit)
+		to_chat(usr, "The machine cannot hold anymore items.")
+		return TRUE
+
+	//Fill machine with a bag!
+	if(istype(I, /obj/item/storage/bag))
+		var/obj/item/storage/bag/B = I
+		if(!B.contents.len)
+			to_chat(user, "<span class='warning'>[B] is empty.</span>")
+			return FALSE
```
moxian: I think the logic is still inverted here (and below)?
In the rest of the proc, `TRUE` seems to be returned on failure of any kind, and `FALSE` on success; but it's the opposite in bag handling.
(Honestly I'm not sure `TRUE`/`FALSE` make it any better than `0`/`1` at all... but that's tangential)

SteelSlayer: You're right, the return logic is reverse for a lot of the other stuff. It should be `TRUE` on success and `FALSE` on failure. I'll have to change it. I've been told to return T/F instead of 1/0, so that's why I'm doing that

PR #12718, opened 2019-11-09
Waterbreath Trait, Snowflake Removal

Leanfrog:
## What Does This PR Do
Creates a new trait that allows for breathing underwater which was given to Slimes, Skrell, and Neara, it also removes the snowflake from the poolcontroller code which was meant to allow Skrell and Neara to breath underwater.

## Why It's Good For The Game
The removal of snowflaking is for the better, it also returns the ability to be in water without internals to Slime People which was removed as part of their rework as well as creating a new trait for possible future use.

## Images of changes
N/A

## Changelog
:cl:
add: new water breathing trait, Skrell, Slime People, and Neara now use the new waterbreathing trait 
tweak: cleaned out the snowflake code in poolcontroller.dm
/:cl:

### DIFF
--- a/code/__HELPERS/traits.dm
+++ b/code/__HELPERS/traits.dm
@@ -62,6 +62,7 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 
 //mob traits
 #define TRAIT_PACIFISM			"pacifism"
+#define TRAIT_WATERBREATH       "waterbreathing"
 
 // common trait sources
 #define ROUNDSTART_TRAIT "roundstart" //cannot be removed without admin intervention
--- a/code/modules/mob/living/carbon/human/species/skrell.dm
+++ b/code/modules/mob/living/carbon/human/species/skrell.dm
@@ -12,6 +12,7 @@
 	herbivores on the whole and tend to be co-operative with the other species of the galaxy, although they rarely reveal \
 	the secrets of their empire to their allies."
 
+
 	species_traits = list(LIPS)
 	clothing_flags = HAS_UNDERWEAR | HAS_UNDERSHIRT | HAS_SOCKS
 	bodyflags = HAS_SKIN_COLOR | HAS_BODY_MARKINGS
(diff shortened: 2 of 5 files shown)


KasparoVy:
Smoke dihydrogen monoxide every day

code/__HELPERS/traits.dm
```diff
@@ -62,6 +62,7 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 
 //mob traits
 #define TRAIT_PACIFISM			"pacifism"
+#define TRAIT_WATERBREATH       "waterbreathing"
```
Ty-Omaha: It would make more sense to add it here: https://github.com/ParadiseSS13/Paradise/blob/99a53306847df59f054931bbd07c045e32f2f58a/code/__DEFINES/genetics.dm#L144

code/modules/mob/living/carbon/human/species/skrell.dm
```diff
@@ -43,4 +44,12 @@
...
\ No newline at end of file
+		"is strangling themselves with their own tendrils!")
+
+/datum/species/skrell/on_species_gain(mob/living/carbon/human/H)
+	..()
+	ADD_TRAIT(H, TRAIT_WATERBREATH, "species")
+
+/datum/species/skrell/on_species_loss(mob/living/carbon/human/H)
+	..()
+	REMOVE_TRAIT(H, TRAIT_WATERBREATH, "species")
```
Ty-Omaha: Can't you just add the trait to the `species_traits ` list instead of `ADD_TRAIT` and `REMOVE_TRAIT`?

https://github.com/ParadiseSS13/Paradise/pull/12718/files#diff-6befeb8738a945b5590591fce266433cR16

code/__HELPERS/traits.dm
Leanfrog: It was suggested to me to use this version as the framework for traits was recently added and it also lays the ground work for future use in items and other things.

code/modules/mob/living/carbon/human/species/skrell.dm
Couls: I assume the plan is to eventually get rid of species_trait entirely and just use the framework we have here, better to future proof this PR it’s also more readable to have HAS_TRAIT instead of TRAIT_X in dna.species.species_traits

Ty-Omaha: right but then with multiple traits you'd have to spam ADD/REMOVE

Couls: When all species traits are turned into real ones you can just go through the list and add them in a loop like TG does, I didn’t want to ask lumpy to do this because this is their first PR

PR #12721, opened 2019-11-10
Fix Re-admining self on test server (with legacy admin config)

moxian:
## What Does This PR Do
You can now readmin self after de-admining self on test server.

Moved the functionality to parse admins.txt to a dedicated function, and call that, instead of reimplementing the config parser each time it's needed (it's only two, but 50% of those happened to runtime on default example config).

## Why It's Good For The Game
bugfix; makes easier to test stuffs on local server

### DIFF
--- a/code/modules/admin/admin_ranks.dm
+++ b/code/modules/admin/admin_ranks.dm
@@ -58,6 +58,28 @@ var/list/admin_ranks = list()								//list of all ranks with associated rights
 	load_admins()
 	return 1
 
+/proc/parse_admins_config_file()
+	var/list/admins = new/list()
+	for(var/line in file2list("config/admins.txt"))
+		// strip out comments
+		var/comment_start = findtext(line, "#")
+		if(comment_start > 0)
+			line = copytext(line, comment_start)
+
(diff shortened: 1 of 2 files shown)


code/modules/admin/admin_ranks.dm
```diff
@@ -58,6 +58,28 @@ var/list/admin_ranks = list()								//list of all ranks with associated rights
 	load_admins()
 	return 1
 
+/proc/parse_admins_config_file()
+	var/list/admins = new/list()
+	for(var/line in file2list("config/admins.txt"))
+		// strip out comments
+		var/comment_start = findtext(line, "#")
+		if(comment_start > 0)
+			line = copytext(line, comment_start)
```
farie82: idea of the comments is to ignore that line. Not to remove the # of the line then continue

moxian: The intention was to remove everything after `#`, not everything before... Ooops...
Fixed now, thanks!

Fox-McCloud:
Please fix.

Fox-McCloud:
No response. Closing.

PR #12728, opened 2019-11-12
Changes the spray bottle admin attack log messages

SteelSlayer:
## What Does This PR Do
I'm making this based on @TDSSS's requests in the discord coder chat.

Changes spray bottle attack logging to now include all the contents of the spray and the temperature of the spray. By default, this will appear to admins that have their attack log prefs set to MOST, ALMOST ALL, or ALL.

If the bottle contains Sacid, Facid or space lube, the logs will also show up for admins on the FEW preference in addition to the prefs above. Basically everything except the NO preference.

If the bottle contains nothing but space cleaner, no logs will be shown of the spray, *unless* the temperature is above 300 or below 280. This covers all species temperature thresholds.

~~Sidenote: I don't think the coordinates for this log are particularly useful and could be removed, but correct me if I'm wrong and they're actually useful for something.~~

## Images of changes
spray

## Changelog
:cl:
tweak: Changes how spray bottle logging works. Show's its contents and temperature in attack logs
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/spray.dm
+++ b/code/modules/reagents/reagent_containers/spray.dm
@@ -51,7 +51,7 @@
 	user.newtonian_move(get_dir(A, user))
 
 	if(reagents.reagent_list.len == 1 && reagents.has_reagent("cleaner")) // Only show space cleaner logs if it's burning people from being too hot or cold
-		if((reagents.chem_temp < 300) && (reagents.chem_temp > 280)) // 280 is the cold threshold for slimes, 300 the hot threshold for drask
+		if(reagents.chem_temp < 300 && reagents.chem_temp > 280) // 280 is the cold threshold for slimes, 300 the hot threshold for drask
 			return
 
 	var/attack_log_type = ATKLOG_MOST
--- a/code/modules/reagents/reagent_containers/spray.dm
+++ b/code/modules/reagents/reagent_containers/spray.dm
@@ -43,20 +43,22 @@
 		to_chat(user, "<span class='notice'>[src] is empty!</span>")
 		return
 
-	var/contents_log = get_formatted_contents_string()
+	var/contents_log = reagents.reagent_list.Join(", ")
 	spray(A)
 
 	playsound(loc, 'sound/effects/spray2.ogg', 50, 1, -6)
 	user.changeNext_move(CLICK_CD_RANGE*2)
 	user.newtonian_move(get_dir(A, user))
 
-	if(reagents.reagent_list.len == 1 && reagents.has_reagent("cleaner")) // Don't show logs for bottles spraying only space cleaner
-		return
+	if(reagents.reagent_list.len == 1 && reagents.has_reagent("cleaner")) // Only show space cleaner logs if it's burning people from being too hot or cold
+		if((reagents.chem_temp < 300) && (reagents.chem_temp > 280)) // 280 is the cold threshold for slimes, 300 the hot threshold for drask
+			return
 
 	var/attack_log_type = ATKLOG_MOST
 	if(reagents.has_reagent("sacid") || reagents.has_reagent("facid") || reagents.has_reagent("lube"))
 		attack_log_type = ATKLOG_FEW
 	msg_admin_attack("[key_name_admin(user)] used a spray bottle at [COORD(user)] - Contents: [contents_log] - Temperature: [reagents.chem_temp]K", attack_log_type)
+	log_game("[key_name(user)] used a spray bottle at [COORD(user)] - Contents: [contents_log] - Temperature: [reagents.chem_temp]K")
 	return
 
 
--- a/code/modules/reagents/reagent_containers/spray.dm
+++ b/code/modules/reagents/reagent_containers/spray.dm
@@ -43,21 +43,20 @@
 		to_chat(user, "<span class='notice'>[src] is empty!</span>")
 		return
 
+	var/contents_log = get_formatted_contents_string()
 	spray(A)
 
 	playsound(loc, 'sound/effects/spray2.ogg', 50, 1, -6)
 	user.changeNext_move(CLICK_CD_RANGE*2)
 	user.newtonian_move(get_dir(A, user))
 
-	if(reagents.has_reagent("sacid"))
-		msg_admin_attack("[key_name_admin(user)] fired sulphuric acid from \a [src] at [COORD(user)].", ATKLOG_FEW)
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -43,21 +43,20 @@
...
-		log_game("[key_name(user)] fired fluorosulfuric Acid from \a [src] at [COORD(user)].")
-	if(reagents.has_reagent("lube"))
-		msg_admin_attack("[key_name_admin(user)] fired space lube from \a [src] at [COORD(user)].")
-		log_game("[key_name(user)] fired space lube from \a [src] at [COORD(user)].")
+	if(reagents.reagent_list.len == 1 && reagents.has_reagent("cleaner")) // Don't show logs for bottles spraying only space cleaner
+		return
+
+	var/attack_log_type = ATKLOG_MOST
+	if(reagents.has_reagent("sacid") || reagents.has_reagent("facid") || reagents.has_reagent("lube"))
+		attack_log_type = ATKLOG_FEW
```
Ty-Omaha: Reversed logic with ATKLOG_MOST and ATKLOG_FEW

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -43,21 +43,20 @@
...
+	msg_admin_attack("[key_name_admin(user)] used a spray bottle at [COORD(user)] - Contents: [contents_log] - Temperature: [reagents.chem_temp]K", attack_log_type)
```
Ty-Omaha: You also need to use `log_game`

Ty-Omaha:
>Sidenote: I don't think the coordinates for this log are particularly useful and could be removed, but correct me if I'm wrong and they're actually useful for something.

They are very useful in verifying people's stories

TDSSS:
Couldn't you still harm people with superheated space cleaner and it wouldn't be logged?

code/modules/reagents/reagent_containers/spray.dm
Ty-Omaha: Maybe I shouldnt review 30 min after I wake up. I messed up the logic, I think.

SteelSlayer: Well you see the attack log defines look like this:
```js
#define ATKLOG_ALL	0
#define ATKLOG_ALMOSTALL	1
#define ATKLOG_MOST	2
#define ATKLOG_FEW	3
#define ATKLOG_NONE	4
```
And inside msg_admin_attack we have:
```js
if(C.prefs.atklog <= loglevel)
    to_chat(C, msg)
```
My code is intended to be written this way.

code/modules/reagents/reagent_containers/spray.dm
SteelSlayer: Ah yep, left that out..

SteelSlayer:
@TDSSS Yeah that's true, you could. I just didn't want to spam the logs for the no doubt, many MANY times space cleaner is used during a round.

Ty-Omaha:
What about a temp check in where the acids and lube are?

TDSSS:
Only log it if the space cleaner is superheated/cooled then? I don't expect people to go cleaning with scalding hot space cleaner after all

SteelSlayer:
Suppose I gotta find out at which temperature, hot or cold, that a chem can burn people

TDSSS:
It's somewhere in the species I think, since if depends on species. For simplicity I would use the temps that burn humans

SteelSlayer:
Alright I updated this. It now only logs pure space cleaner sprays if the temp is above 300 or below 280. 300 is the heat threshold for drasks and 280 is the cold threshold for slime people. This should cover all  species.

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -43,21 +43,22 @@
...
-		msg_admin_attack("[key_name_admin(user)] fired sulphuric acid from \a [src] at [COORD(user)].", ATKLOG_FEW)
-		log_game("[key_name(user)] fired sulphuric acid from \a [src] at [COORD(user)].")
-	if(reagents.has_reagent("facid"))
-		msg_admin_attack("[key_name_admin(user)] fired fluorosulfuric acid from \a [src] at [COORD(user)].", ATKLOG_FEW)
-		log_game("[key_name(user)] fired fluorosulfuric Acid from \a [src] at [COORD(user)].")
-	if(reagents.has_reagent("lube"))
-		msg_admin_attack("[key_name_admin(user)] fired space lube from \a [src] at [COORD(user)].")
-		log_game("[key_name(user)] fired space lube from \a [src] at [COORD(user)].")
+	if(reagents.reagent_list.len == 1 && reagents.has_reagent("cleaner")) // Only show space cleaner logs if it's burning people from being too hot or cold
+		if((reagents.chem_temp < 300) && (reagents.chem_temp > 280)) // 280 is the cold threshold for slimes, 300 the hot threshold for drask
```
Citinited: Just so you know, these extra brackets around the `<` `>` aren't necessary as BYOND computes `<>` before `&&`

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -43,21 +43,22 @@
...
-	if(reagents.has_reagent("sacid"))
-		msg_admin_attack("[key_name_admin(user)] fired sulphuric acid from \a [src] at [COORD(user)].", ATKLOG_FEW)
-		log_game("[key_name(user)] fired sulphuric acid from \a [src] at [COORD(user)].")
-	if(reagents.has_reagent("facid"))
-		msg_admin_attack("[key_name_admin(user)] fired fluorosulfuric acid from \a [src] at [COORD(user)].", ATKLOG_FEW)
-		log_game("[key_name(user)] fired fluorosulfuric Acid from \a [src] at [COORD(user)].")
-	if(reagents.has_reagent("lube"))
-		msg_admin_attack("[key_name_admin(user)] fired space lube from \a [src] at [COORD(user)].")
-		log_game("[key_name(user)] fired space lube from \a [src] at [COORD(user)].")
+	if(reagents.reagent_list.len == 1 && reagents.has_reagent("cleaner")) // Only show space cleaner logs if it's burning people from being too hot or cold
```
Crazylemon64: It could possibly be worth adding a variable to reagents which marks them as "benign" and not reporting if all compounds have that variable set and the mixture is within a sane temperature range 

but this should work fine

TDSSS: There is such a var, harmless. It was added in https://github.com/ParadiseSS13/Paradise/pull/12302

PR #12740, opened 2019-11-20
Traitor Special Role Fix

### LINKED ISSUE #12739: Syndicate Deport will not register visiting agents
**Issue Description**:

When an Agent ID is used to try and log into the Syndicate Deport, it will not register it, and rather treat it like a regular crew ID despite it having access to doors and turrets. 

Being able to sign into the Syndicate Deport, making mobs non-hostile. 

**What happened instead**:

I was not able to sign in at all.

This makes the ability to sign in as a syndicate agent useless other than the ability to send a one time message to the Syndicate. 

Use an Agent ID, get into the Deport, and try to sign in. 

It first happened to me at 17/11/2019 (Australian Dating) 

**Extra information**:

Here are pictures of two different examples of it happening to me, including my objectives, my agent ID, and the red text forbidding me from signing in (Slightly obscured in one of the pictures) 
https://gyazo.com/1df219807ebeafdb317e3629c8839d14
 https://gyazo.com/59bb24d0ae9522fb0518168b61c700c3

Drakeven:
## What Does This PR Do
Fixes #12739 caused by the mislabeling of the special role for traitor and mindslaves, it also fixes the unlisted issue of traitors being unable to hijack the shuttle, as it also checks for the special roll. 

## Why It's Good For The Game
It fixes some awfully inconvenient bugs regarding traitors 

## Changelog
:cl:
fix: Fixed mislabeled traitor and mindslave special role.
/:cl:

### DIFF
--- a/code/__DEFINES/role_preferences.dm
+++ b/code/__DEFINES/role_preferences.dm
@@ -10,7 +10,6 @@
 // justice if someone's abusing your role
 #define ROLE_SYNDICATE			"Syndicate"
 #define ROLE_TRAITOR			"traitor"
-#define ROLE_MINDSLAVE			"mindslave"
 #define ROLE_OPERATIVE			"operative"
 #define ROLE_CHANGELING			"changeling"
 #define ROLE_WIZARD				"wizard"
--- a/code/game/gamemodes/autotraitor/autotraitor.dm
+++ b/code/game/gamemodes/autotraitor/autotraitor.dm
@@ -52,8 +52,7 @@
 		if(!traitor || !istype(traitor))
 			pre_traitors.Remove(traitor)
 			continue
-		if(istype(traitor))
-			traitor.special_role = SPECIAL_ROLE_TRAITOR
+		if(istype(traitor))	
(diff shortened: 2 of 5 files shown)


Drakeven:
Thanks to @JKnutson101 for pointing out where exactly the issue was coming from, and giving me a hand fixing it!

SteelSlayer:
Nice catch. Since you're already doing this PR, maybe you can sneak in some code touch ups related to this?

1) I made ROLE_MINDSLAVE just for mindslaves but since its really not needed anymore, you can delete the define I made for it.

2) This line can be deleted from the traitor gamemode since when we add the traitor datum it will assign the special role itself: https://github.com/ParadiseSS13/Paradise/blob/master/code/game/gamemodes/traitor/traitor.dm#L52

3) Likewise with autotraitor, you can delete this line as well: https://github.com/ParadiseSS13/Paradise/blob/master/code/game/gamemodes/autotraitor/autotraitor.dm#L56

Drakeven:
Hopefully that should do it!

PR #12743, opened 2019-11-21
Added Pylon healing to Wizards

Drakeven:
## What Does This PR Do
This PR allows wizards to heal from cult pylons, which they can construct via the use of soulstones and the 'Create Shell' spell. 

## Why It's Good For The Game
It is strange that the wizard's constructs are healed, but not the wizard themselves, by pylons. This small buff gives consistency to the mechanics of the pylons. 

## Changelog
:cl:
balance: Allows wizards to be healed by cult pylons. 
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult_structures.dm
+++ b/code/game/gamemodes/cult/cult_structures.dm
@@ -198,7 +198,7 @@ var/list/blacklisted_pylon_turfs = typecacheof(list(
 	if(last_heal <= world.time)
 		last_heal = world.time + heal_delay
 		for(var/mob/living/L in range(5, src))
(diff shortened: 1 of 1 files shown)


TDSSS:
This only heals wizards, not apprentices, right? Might be worth considering those too.

taukausanake:
I know it probably doesn't matter, but I would consider a subtype of pylon for wizards only, if just so they can have a Protoss themed one but also just in case for some reason you get cult and wizards going at the same time. I know it won't ever happen but that's something I'd do

Drakeven:
I'm currently working on some unique Wizard sprites! So I'll fiddle with the details then. Good idea about apprentices, though.

PR #12744, opened 2019-11-21
Getting cluwned ignores species brain damage modifiers

SteelSlayer:
## What Does This PR Do
Makes the brain damage caused from honk tumors and direct cluwning methods a flat 80 damage. It is no longer affected by species brain damage modifiers.

## Why It's Good For The Game
First of all, insta kills are bad mkaaay. Second, this allows slime people to experience the gloriousness of becoming a cluwne without dying immediately. You can now instead get beaten to death by your fellow crewmembers, HONK!!

 This also means that certain golems which take reduced brain damage will properly be turned into a stumbling idiot, as is intended with cluwning.

## Changelog
:cl:
tweak: The brain damage caused from honk tumors and direct cluwning methods is now a flat 80 damage. It is no longer affected by species brain damage modifiers
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -39,7 +39,7 @@
 		update_stat("adjustBrainLoss")
 	return STATUS_UPDATE_STAT
 
-/mob/living/carbon/human/setBrainLoss(amount, updating = TRUE)
+/mob/living/carbon/human/setBrainLoss(amount, updating = TRUE, ignore_brain_mod = FALSE)
 	if(status_flags & GODMODE)
 		return STATUS_UPDATE_NONE	//godmode
 
@@ -47,7 +47,8 @@
 		var/obj/item/organ/internal/brain/sponge = get_int_organ(/obj/item/organ/internal/brain)
 		if(sponge)
 			if(dna.species && amount > 0)
-				amount = amount * dna.species.brain_mod
+				if(!ignore_brain_mod)
+					amount = amount * dna.species.brain_mod
 			sponge.damage = Clamp(amount, 0, 120)
 			if(sponge.damage >= 120)
 				visible_message("<span class='alert'><B>[src]</B> goes limp, [p_their()] facial expression utterly blank.</span>")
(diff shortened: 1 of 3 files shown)


JKnutson101:
Now we can beat up Slimefolk! Hurray?

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -39,15 +39,16 @@
 		update_stat("adjustBrainLoss")
 	return STATUS_UPDATE_STAT
 
-/mob/living/carbon/human/setBrainLoss(amount, updating = TRUE)
+/mob/living/carbon/human/setBrainLoss(amount, updating = TRUE, ignore_brain_mod = FALSE)
```
TDSSS: I know it's not really on you since it was like that before but man, the duplicate code between setbrainloss and adjustbrainloss really bugs me. As does the fact now one has an additional var/check while the other one doesn't.

SteelSlayer: Yeah that's true. I suppose I can add this arg to adjustBrainLoss as well. I'm sure it will also be useful at some point

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -22,15 +22,16 @@
 	med_hud_set_status()
 	handle_hud_icons_health()
 
-/mob/living/carbon/human/adjustBrainLoss(amount, updating = TRUE)
+/mob/living/carbon/human/adjustBrainLoss(amount, updating = TRUE, ignore_brain_mod = FALSE)
```
moxian: minor style thing: Try to have variables named positively whenever you can, and tweak their values as necessary instead.
To illustrate, `ignore_brain_mod = FALSE` is the same as `avoid_not_ignoring_brain_mod = FALSE`, but figuring out what does `if(!avoid_not_ignoring_brain_mod)` means takes an unreasonable amount of time.
Similarly, it's much more straightforward to reason about `if(use_brain_mod)` than about `if(!ignore_brain_mod)`.

farie82: Using positive variables (do_x) is usually more readable than negative variables(do_not_do_x).
Negative does not always exclude what it does do example. do_not_harm, what will it do then? Heal?

SteelSlayer: I like use_brain_mod instead

PR #12748, opened 2019-11-22
Adds examine text for people who have open surgical incisions

SteelSlayer:
## What Does This PR Do
Idea taken from this forum post: 
https://nanotrasen.se/forum/topic/17316-add-examine-text-for-surgically-opened-limbs/
@Spacemanspark 

Organic limbs and robotic limbs have distinct messages which you can see in the picture below. IPC limbs will have the robotic examine text.

## Why It's Good For The Game
As was said in the forum post, it would be nice for when roboticists are working on IPCs. They don't have a body scanner and it can be annoying if the patient has a bunch of open body parts. And in general, it makes sense that you should be able to see if someone has an open surgical cut or an open maintenance hatch, in the case of IPC/robolimbs.

## Images of changes
E

## Changelog
:cl:
add: Adds examine text for people with open incisions from surgery
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -215,6 +215,12 @@
 				else if(E.status & ORGAN_SPLINTED)
 					wound_flavor_text["[E.limb_name]"] = "[p_they(TRUE)] [p_have()] a splint on [p_their()] [E.name]!\n"
 
+			if(E.open)
+				if(E.is_robotic())
+					msg += "<b>The maintenance hatch on [p_their()] [ignore_limb_branding(E.limb_name)] is open!</b>\n"
+				else
+					msg += "<b>[p_their(TRUE)] [ignore_limb_branding(E.limb_name)] has an open incision!</b>\n"
+
 			for(var/obj/item/I in E.embedded_objects)
 				msg += "<B>[p_they(TRUE)] [p_have()] \a [bicon(I)] [I] embedded in [p_their()] [E.name]!</B>\n"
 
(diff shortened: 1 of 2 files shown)


Spacemanspark:
Well, guess Dumbdumn's computer is spared
:+1:

code/modules/surgery/organs/organ_external.dm
```diff
@@ -47,7 +47,7 @@
 	var/damage_msg = "<span class='warning'>You feel an intense pain</span>"
 	var/broken_description
 
-	var/open = 0
+	var/open = 0  // If the body part has an open incision from surgery
```
TDSSS: Shouldn't these be FALSE?

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -215,6 +215,12 @@
 				else if(E.status & ORGAN_SPLINTED)
 					wound_flavor_text["[E.limb_name]"] = "[p_they(TRUE)] [p_have()] a splint on [p_their()] [E.name]!\n"
 
+			if(E.open)
+				if(E.is_robotic())
+					msg += "<b>The maintenance hatch on [p_their()] [ignore_limb_branding(E.limb_name)] is open!</b>\n"
```
TDSSS: Wouldn't it be better to use span classes for these?

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -215,6 +215,12 @@
 				else if(E.status & ORGAN_SPLINTED)
 					wound_flavor_text["[E.limb_name]"] = "[p_they(TRUE)] [p_have()] a splint on [p_their()] [E.name]!\n"
 
+			if(E.open)
+				if(E.is_robotic())
+					msg += "<b>The maintenance hatch on [p_their()] [ignore_limb_branding(E.limb_name)] is open!</b>\n"
+				else
+					msg += "<b>[p_their(TRUE)] [ignore_limb_branding(E.limb_name)] has an open incision!</b>\n"
```
TDSSS: same for here? As far as I can tell, spans should work here, right?

code/modules/surgery/organs/organ_external.dm
SteelSlayer: Yeah I can change it

code/modules/mob/living/carbon/human/examine.dm
SteelSlayer: I'm not sure if it it's better to do `<span class='bold'>` or just use `<b>` to be honest. I see both used in this file.

TDSSS: I thought generally we prefer using span classes, but would have to ask a maint what the actual policy is.

Cazdon:
Downloaded and did some testing. Everything seemed to work just fine.

surgery_tested

Also tried to see if anything would be weird if I cut off limbs mid surgery. Happy to say it worked as expected.
surgery_tested1


Functionally at least it looks good! No runtimes or weird issues.

PR #12751, opened 2019-11-22
fixes ticket system bug

### LINKED ISSUE #12750: Ticket system can bug out, making it not work.
**Issue Description**:
Sometimes, you can't take or resolve tickets and the ticket window doesn't work.

for it to work

**What happened instead**:
take/resolve and ticket button do nothing, ticket window lists no tickets.

much easier to  to lose an ahelp or mhelp

I _think_ this is caused by someone ahelping/mhelping before the ticket system is initialized, which makes the alltickets var hold a single ticket instead of a list of tickets, since it's only initialized after that (I suspect), so try this:
start a server, immediately ahelp/mhelp before the server has initialized those systems and check.

**Extra information**:
See this runtime it throws when someone ahelps:
[2019-11-22T02:25:04] Runtime in tickets.dm,90: type mismatch: /datum/ticket (/datum/ticket) += /datum/ticket (/datum/ticket)
   proc name: newTicket (/datum/controller/subsystem/tickets/proc/newTicket)
   usr: CKEY CENSORED (/mob/new_player)
   usr.loc: The floor (8,174,2) (/turf/unsimulated/floor)
   src: Admin Tickets (/datum/controller/subsystem/tickets)

TDSSS:
## What Does This PR Do
fixes #12750
The simplest way to fix this seems to be to just create the actual list when the ticket system is created. The alternative was checking for the existence of the list each time we add a new ticket.

## Why It's Good For The Game
bugs are bad

## Changelog
:cl:
fix: fixed a bug preventing ticket system from working under some circumstances.
/:cl:

### DIFF
--- a/code/controllers/subsystem/tickets/tickets.dm
+++ b/code/controllers/subsystem/tickets/tickets.dm
@@ -87,6 +87,7 @@ SUBSYSTEM_DEF(tickets)
 		title = passedContent
 
 	var/datum/ticket/T =  new(title, passedContent, getTicketCounterAndInc())
+	LAZYINITLIST(allTickets)	//this checks if the list exists yet and if not creates it. Feels weird to have this each time we make a ticket but I think it's correct.
 	allTickets += T
 	T.clientName = C
 	T.locationSent = C.mob.loc.name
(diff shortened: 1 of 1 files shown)


TDSSS:
Found a much simpler way to fix that bug, changes pushed.

PR #12763, opened 2019-11-30
Massive Clown Nerfs - Unbuckle time while restrained is now equal to breakout time.

Cazdon:
## What Does This PR Do
Title.

## Why It's Good For The Game
Mostly to correct Toy Handcuffs. They have a breakout time of 0, meaning they're meant for jokes and if anyone wants to break free they can instantly. However if you bucklecuff someone with toy cuffs, they would have to break out in 60 seconds. This also very slightly nerfs things like zipties which have a 45 second breakout time compared to the handcuff's 60. 

EDIT: Meant to include this also slightly nerfs cable-restraints similar to zipties. 30 Second unbuckle time as opposed to 60.

(Please inform me of better ways to do this if there are any.)

## Changelog
:cl:
balance: Unbuckle time is now equal to breakout time.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -771,13 +771,11 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 	return(istype(src.wear_mask, /obj/item/clothing/mask/muzzle))
 
 /mob/living/carbon/resist_buckle()
-	var/breakouttime = 600
-
 	spawn(0)
 		resist_muzzle()
 	if(restrained())
 		var/obj/item/I = handcuffed
-		breakouttime = I.breakouttime
+		var/breakouttime = I.breakouttime
 		var/displaytime = breakouttime / 600
 		changeNext_move(CLICK_CD_BREAKOUT)
 		last_special = world.time + CLICK_CD_BREAKOUT
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -771,14 +771,19 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 	return(istype(src.wear_mask, /obj/item/clothing/mask/muzzle))
 
 /mob/living/carbon/resist_buckle()
+	var/breakouttime = 600
+
 	spawn(0)
 		resist_muzzle()
 	if(restrained())
+		var/obj/item/I = handcuffed
+		breakouttime = I.breakouttime
+		var/displaytime = breakouttime / 600
 		changeNext_move(CLICK_CD_BREAKOUT)
 		last_special = world.time + CLICK_CD_BREAKOUT
 		visible_message("<span class='warning'>[src] attempts to unbuckle [p_them()]self!</span>", \
-					"<span class='notice'>You attempt to unbuckle yourself... (This will take around one minute and you need to stay still.)</span>")
-		if(do_after(src, 600, 0, target = src))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -771,14 +771,19 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
 	return(istype(src.wear_mask, /obj/item/clothing/mask/muzzle))
 
 /mob/living/carbon/resist_buckle()
+	var/breakouttime = 600
+
 	spawn(0)
 		resist_muzzle()
 	if(restrained())
+		var/obj/item/I = handcuffed
+		breakouttime = I.breakouttime
```
FlattestGuitar: Don't cable ties have a different value than 600?

Cazdon: Yeah I just set it to 600 by default to avoid possible complications. The breakout time is adjusted to the item they're 'handcuffed' with, with `breaktouttime = I.breakouttime`.

FlattestGuitar: I mean this: https://github.com/ParadiseSS13/Paradise/blob/5936bb98fea2bc46769df7ee6bdf57a5f3f3db37/code/game/objects/items/weapons/handcuffs.dm#L95 will override the 60 seconds that are in the code right now.

Cazdon: Yeah that's what the beakouttime variable gets set to. I declare them at 600 just as a default, but then the `var/obj/item/I = handcuffed` gets the item the person is handcuffed with, and then we set the breakoutime to whatever that item's breakout time is, meaning the 600 gets changed to 300 for cable cuffs, 0 for toy cuffs, 450 for zipties, etc. 

if this is about the display time, then that's the same display time the regular handcuff code uses. (This will take 0.5 minutes to break out of, as an example.)

TDSSS: In what scenario are you handcuffed with cuffs without a breakout time? ANd even if it did, then breakouttime would just be treated as 0 which isn't the worst issue, imo.

Cazdon: That is true. I'll look into cleaning this up a little bit.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -774,11 +774,14 @@ var/list/ventcrawl_machinery = list(/obj/machinery/atmospherics/unary/vent_pump,
...
 	if(restrained())
+		var/obj/item/I = handcuffed
+		var/breakouttime = I.breakouttime
+		var/displaytime = breakouttime / 600
 		changeNext_move(CLICK_CD_BREAKOUT)
 		last_special = world.time + CLICK_CD_BREAKOUT
 		visible_message("<span class='warning'>[src] attempts to unbuckle [p_them()]self!</span>", \
-					"<span class='notice'>You attempt to unbuckle yourself... (This will take around one minute and you need to stay still.)</span>")
-		if(do_after(src, 600, 0, target = src))
+					"<span class='notice'>You attempt to unbuckle yourself... (This will take around [displaytime] minute and you need to stay still.)</span>")
```
Crazylemon64: For values of breakouttime other than 600 (1 minute) this will display weird things like `0.75 minute`, `0 minute`, or similar

Cazdon: Right let me see if I can convert that to seconds.

Cazdon: Got this wrapped up. Display time is in seconds, and changed uncuff time to follow suit.

Issue #12764, opened 2019-12-01
On station rolls are getting flagged as off station roles.


Cazdon:
Crewmembers are randomly having their Offstation_Role flag set to TRUE, when they've been crew from roundstart.

**What did you expect to happen**: 
To never have the Offstation_Role flag set to TRUE unless they are truly offstation.

**What happened instead**:
Some Civillian had their Offstation_Role set to TRUE. Some Borg had his Offstation role set to TRUE. I can't re-create this but these were confirmed by @Drakeven and @TDSSS that this happened on live.

**Why is this bad/What are the consequences**:
As of #12577 Off Station roles could no longer be converted, and with #12571 non-convertibles would get sacrificed, meaning perfectly convertible people were wound up sacrificed, to the fault of neither PR mind you.

**Steps to reproduce the problem**:
Not sure. We can rule out this. The civilian in question had Scientist roll set to high, but didn't get their job, and had 'become civilian instead' enabled. Recreating this scenario showed no issues, and nothing else was out of the ordinary about the civilian. 

**When did the problem start happening**:
No idea, this might have been a bug for the games whole lifespan, but only became a problem with the two above mentioned PRs.

**Extra information**:
Big thanks to @TDSSS , @Drakeven , and @TheSardele for answering my questions and helping me find this issue.

datlo:
Offstation role is a fairly recent variable I added myself. I have no idea what could cause it to get flipped on outside of actually spawning in as a spawner or other special role.

Kyep:
Needs to be investigated/fixed, so the off-station role check can be re-added.

PR #12765, opened 2019-12-01
Cult Hotfix - Comments out Off_Station role check 

Cazdon:
## What Does This PR Do
Comments out the Off_Station role check introduced by #12577 because as described in #12764 crew are randomly getting their offstation_role flag set to TRUE, causing them to be non-convertible to the cult.

## Why It's Good For The Game
Need functionality while this strange bug is looked into, nowhere in code is offstation_role set to TRUE except when initializing actual off station_roles. 

## Changelog
:cl:
fix: Potential Convertee's being unable to be converted.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult.dm
+++ b/code/game/gamemodes/cult/cult.dm
@@ -20,8 +20,8 @@ var/global/list/all_cults = list()
 		var/mob/living/carbon/human/H = mind.current
 		if(ismindshielded(H)) //mindshield protects against conversions unless removed
 			return FALSE
-/*	if(mind.offstation_role) //cant convert offstation roles such as ghost spawns
-		return FALSE */ //Commented out until we can figure out why offstation_role is getting set to TRUE on normal crew.
+/*	if(mind.offstation_role) cant convert offstation roles such as ghost spawns
+		return FALSE Commented out until we can figure out why offstation_role is getting set to TRUE on normal crew. */
 	if(issilicon(mind.current))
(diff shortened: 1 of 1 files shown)


Aurorablade:
This is seriously not the best way or the wisest way to go about this. https://github.com/ParadiseSS13/Paradise/pull/12756/ may have some part of the answer here but it needs merged.

Cazdon:
> 
> 
> This is seriously not the best way or the wisest way to go about this. #12756 may have some part of the answer here but it needs merged.

Not trying to be rude, but what would autotraitor/traitor changes have to do with a problem that's effecting multiple round types?

variableundefined:
@Cazdon Does not compile

Cazdon:
@variableundefined fixed. Sorry about that, I should know better.

Aurorablade:
Yes it could and also simply commenting out the check is going to open up so much cheese for cult it is not funny

Cazdon:
> 
> 
> Yes it could and also simply commenting out the check is going to open up so much cheese for cult it is not funny

#12577 has plenty of discussion on the topic. As Fox stated in that PR people spawning in as ghost roles just to be converted are an administration issue. We've had this check for less than a month, and it causes normal crew to wind up perma-killed when they should be converted, and not even admins can convert because culting in the traitor panel uses the same check. So currently having this offstation check is breaking cult, and the cheese described hasn't been out of the game for long, and can/should be handled admin side.

PR #12772, opened 2019-12-03
Fixes the null pointer in get_malf_status, refactors the proc to use enums

farie82:
## What Does This PR Do
Title says it
Fixes a missed logical path in #12760 (https://github.com/ParadiseSS13/Paradise/pull/12772/files/dcfcd31c866406b90548191ae3612ffcdea43e55#r357193287 is the logical change) 
Refactors the proc to use enums instead of plain numerical returns

## Changelog
:cl:
fix: Using APCs won't cause runtimes if used by non AIs
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -749,7 +749,7 @@
 
 
 /obj/machinery/power/apc/proc/get_malf_status(mob/living/silicon/ai/malf)
-	if(istype(malf) && malf.mind.has_antag_datum(/datum/antagonist/traitor) || malf.malf_picker) // If they're a traitor OR they have the malf picker from the combat module
+	if(istype(malf) && (malf.mind.has_antag_datum(/datum/antagonist/traitor) || malf.malf_picker)) // If they're a traitor OR they have the malf picker from the combat module
 		if(malfai == (malf.parent || malf))
 			if(occupier == malf)
 				return 3 // 3 = User is shunted in this APC
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -30,6 +30,11 @@
 
 #define APC_UPDATE_ICON_COOLDOWN 200 // 20 seconds
 
+// APC malf status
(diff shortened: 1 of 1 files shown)


code/modules/power/apc.dm
```diff
@@ -749,7 +749,7 @@
 
 
 /obj/machinery/power/apc/proc/get_malf_status(mob/living/silicon/ai/malf)
-	if(istype(malf) && malf.mind.has_antag_datum(/datum/antagonist/traitor) || malf.malf_picker) // If they're a traitor OR they have the malf picker from the combat module
+	if(istype(malf) && (malf.mind.has_antag_datum(/datum/antagonist/traitor) || malf.malf_picker)) // If they're a traitor OR they have the malf picker from the combat module
```
moxian: Early return is recommended both to eliminate excessive nesting, and to make stuff easier to read and prevent this kind of mistake.
I.e.
```js
if(!istype(malf))
  return 0
if(!malf.mind.has_antag_datum(/datum/antagonist/traitor))
  reuturn 0
if(!malf.malf_picker)
  return 0
```

farie82: The given solution is incorrect. But I get what you mean. Will change once able

code/modules/power/apc.dm
```diff
@@ -749,18 +754,22 @@
...
-				return 3 // 3 = User is shunted in this APC
-			else if(istype(malf.loc, /obj/machinery/power/apc))
-				return 4 // 4 = User is shunted in another APC
-			else
-				return 2 // 2 = APC hacked by user, and user is in its core.
+	if(!istype(malf))
+		return FALSE
+	
+	// Only if they're a traitor OR they have the malf picker from the combat module
+	if(!malf.mind.has_antag_datum(/datum/antagonist/traitor) && !malf.malf_picker)
```
farie82: only logic change is here. Prior it was:
if((istype(malf) && isantag) || combat_modules)
    do_stuff

now it is:
if(istype(malf) && (isantag || combat_modules))

PR #12783, opened 2019-12-08
Tweaks to admin freeze and sleeping

SteelSlayer:
## What Does This PR Do
While a mob is admin frozen, their sleeping variable is unable to be changed

## Why It's Good For The Game
Admin freeze should freeze indefinitely and should not be able to be cleared by effects that set sleeping to 0.

## Changelog
:cl:
tweak: While a mob is admin frozen, their sleeping variable is unable to be changed
/:cl:

### DIFF
--- a/code/modules/admin/verbs/freeze.dm
+++ b/code/modules/admin/verbs/freeze.dm
@@ -26,7 +26,7 @@ var/global/list/frozen_mob_list = list()
 /mob/living/var/frozen = null //used for preventing attacks on admin-frozen mobs
 /mob/living/var/admin_prev_sleeping = 0 //used for keeping track of previous sleeping value with admin freeze
 
-/mob/living/proc/admin_Freeze(var/client/admin, skip_overlays = FALSE)
+/mob/living/proc/admin_Freeze(client/admin, skip_overlays = FALSE)
 	if(istype(admin))
 		to_chat(src, "<b><font color= red>You have been frozen by [admin]</b></font>")
 		message_admins("<span class='notice'>[key_name_admin(admin)]</span> froze [key_name_admin(src)]")
@@ -34,24 +34,25 @@ var/global/list/frozen_mob_list = list()
 
 	var/obj/effect/overlay/adminoverlay/AO = new
 	if(skip_overlays)
-		src.overlays += AO
+		overlays += AO
 
-	anchored = 1
-	frozen = AO
+	canmove = FALSE
 	admin_prev_sleeping = sleeping
 	AdjustSleeping(20000)
+	frozen = AO
 	if(!(src in frozen_mob_list))
 		frozen_mob_list += src
 
-/mob/living/proc/admin_unFreeze(var/client/admin, skip_overlays = FALSE)
+/mob/living/proc/admin_unFreeze(client/admin, skip_overlays = FALSE)
 	if(istype(admin))
 		to_chat(src, "<b><font color= red>You have been unfrozen by [admin]</b></font>")
 		message_admins("<span class='notice'>[key_name_admin(admin)] unfroze [key_name_admin(src)]</span>")
 		log_admin("[key_name(admin)] unfroze [key_name(src)]")
 
-	anchored = 0
 	if(skip_overlays)
 		overlays -= frozen
+
+	canmove = TRUE
 	frozen = null
 	SetSleeping(admin_prev_sleeping)
 	admin_prev_sleeping = null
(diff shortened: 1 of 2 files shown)


code/modules/admin/verbs/freeze.dm
```diff
@@ -26,32 +26,33 @@ var/global/list/frozen_mob_list = list()
...
 		to_chat(src, "<b><font color= red>You have been frozen by [admin]</b></font>")
 		message_admins("<span class='notice'>[key_name_admin(admin)]</span> froze [key_name_admin(src)]")
 		log_admin("[key_name(admin)] froze [key_name(src)]")
 
 	var/obj/effect/overlay/adminoverlay/AO = new
 	if(skip_overlays)
-		src.overlays += AO
+		overlays += AO
 
-	anchored = 1
```
TDSSS: Can you explain why you're removing this? (though an anchored mob does seem really weird)

SteelSlayer: Yeah, good point I should explain. Unless I'm mistaken, I found that turning a mob's anchored to 1 doesn't actually stop them from moving. I replaced this with `canmove = 0/1` which actually stops them from moving.

TDSSS: Maybe it stops them from being moved by other things, ie space wind? But it's possible that it's really just useless, true.

SteelSlayer: Yeah so you're right, anchored prevents the anchored mob from getting pulled/moved by other things, so adding that back in.

code/modules/mob/living/status_procs.dm
```diff
@@ -344,6 +344,8 @@
 	return SetSleeping(max(sleeping, amount), updating, no_alert)
 
 /mob/living/SetSleeping(amount, updating = 1, no_alert = FALSE)
+	if(frozen) // If the mob has been admin frozen, sleeping should not be changeable
+		return FALSE
```
moxian: Do not return value [without a clearly-defined semantics for the return value].
Nothing checks AdjustSleeping for whether it succeeded. However *some* things (alcohol and friends) have stuff like `update_flags |= M.AdjustSleeping(-2, FALSE)`. Your particular change happens to be a no-op, but if you returned `TRUE` on success, that would have gotten reinterpreted as "please update health since it likely changed due to metabolism"

code/modules/admin/verbs/freeze.dm
```diff
@@ -26,32 +26,35 @@ var/global/list/frozen_mob_list = list()
...
 		to_chat(src, "<b><font color= red>You have been unfrozen by [admin]</b></font>")
 		message_admins("<span class='notice'>[key_name_admin(admin)] unfroze [key_name_admin(src)]</span>")
 		log_admin("[key_name(admin)] unfroze [key_name(src)]")
 
-	anchored = 0
 	if(skip_overlays)
 		overlays -= frozen
+
+	anchored = FALSE
+	canmove = TRUE
```
moxian: Freezing and then unfreezing a thing that was anchored (or couldn't move) for whatever reason would make it lose its anchoredness.
I can't come up with realistic examples where that would be a problem off top of my head, but it feels the proper way would be to store all those along with `admin_prev_sleeping` somewhere on freeze and restore on unfreeze.

(I don't feel particularly strong about this, feel free to ignore)

code/modules/admin/verbs/freeze.dm
```diff
@@ -26,32 +26,35 @@ var/global/list/frozen_mob_list = list()
...
 
 	var/obj/effect/overlay/adminoverlay/AO = new
 	if(skip_overlays)
-		src.overlays += AO
+		overlays += AO
 
-	anchored = 1
-	frozen = AO
+	anchored = TRUE
+	canmove = FALSE
```
Crazylemon64: Does `update_canmove` check for the value of `frozen`? If it doesn't, any other effect that would call `update_canmove` could possibly set this back to `TRUE`.

SteelSlayer: It doesn't check for frozen specifically. However `canmove` will ultimately be set to `FALSE` if the person is unconscious. And because they're asleep as long as the admin freeze is in effect, `canmove` will never be `TRUE`. 

Setting the frozen person's canmove is probably uneccessary now that I think about it. At the time I thought it would be extra security but I think having their sleeping var unable to be changed is enough.

code/modules/mob/living/status_procs.dm
```diff
@@ -344,6 +344,8 @@
 	return SetSleeping(max(sleeping, amount), updating, no_alert)
 
 /mob/living/SetSleeping(amount, updating = 1, no_alert = FALSE)
+	if(frozen) // If the mob has been admin frozen, sleeping should not be changeable
```
Crazylemon64: i feel a little iffy about this but,
eh

PR #12788, opened 2019-12-10
Fixes headcrabs getting stuck attacking corpses

Kyep:
## What Does This PR Do
Fixes both zombie and non-zombie headcrabs getting stuck in an AI loop whereby they endlessly attack a corpse.
Zombie headcrabs now ignore all unconscious/dead bodies, as they have no reason to care about them (they're already a zombie, and only non-zombie headcrabs can infect corpses, thus zombie headcrabs have no reason to care about corpses).
Non-zombie headcrabs no longer target corpses which are not human. This means they will no longer get stuck in an AI loop attacking say, silicons or other non-human living mobs. They will still consume nearby corpses for health.
Requesting a review from @Terilia 

## Why It's Good For The Game
Fixes a pair of bugs that caused headcrab AI to lock up, making headcrabs trivial to defeat, and spamming the admins with attack logs.

## Changelog
:cl: Kyep
fix: Fixed headcrabs getting stuck attacking corpses
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -19,7 +19,7 @@
 	attack_sound = 'sound/creatures/headcrab_attack.ogg'
 	speak_emote = list("hisses")
 	var/is_zombie = 0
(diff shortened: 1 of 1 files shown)


Terilia:
That is a very neat solution :)

PR #12789, opened 2019-12-11
Exosuit fabricator nano-UI conversion

SteelSlayer:
## What Does This PR Do
This converts the exosuit fabricator's UI to use nano UI, which functions similarly to the autolathe and R&D console UIs. Pardon the gif to show off the features. Apparently OBS can't pick up nanoUI, so I had to settle for a gif instead.

**The swapping of what the exofab's parts determine:**

For some reason the exofab's build time was determined by its manipulator and its resource cost was determined by its micro-laser. I'm not sure why this is the case, however this is the opposite for other fabricating machines (autolathe, protolathe, circuit imprinter, etc.). 

All of these use the *manipulator* to determine cost. So to keep it consistent, I've swapped the part functions on the exofab to be in line with these other machines. The exosuit fabricator's resource cost adjustment is now determined by its manipulator, and its time to build is determined by its micro-laser.

**The removal of the part description and build time features from the exofab:**

To be honest I've maybe used the part description feature about one time. I've never heard anyone talk about using this feature or complain that some of the descriptions are empty. So for the sake of keeping the UI clean I've removed this button.

Similar to the part description feature, I've removed the bit where the UI tells you how long it will take to build the part. It's nice to see the time to build I suppose, but its not that particularly useful and once again I think it would clutter the UI with not that useful information.

## Why It's Good For The Game
The exofab's UI was old, clunky, and lacked search bar. Now it's less annoying to use and to find what you need via the serach bar. If you're lacking a material for a given part, the missing material will show up red, making it easier to see what you need more of.

## Images of changes

Demo gif of the UI:

GIF

Main menu: https://i.imgur.com/buuVJVf.png
Category menu (misc): https://i.imgur.com/ZsauhOK.png
Material storage manu: https://i.imgur.com/dw20lNC.png

## Changelog
:cl:
tweak: The exosuit fabricator now uses nano-UI
tweak: The exosuit fabricator's resource cost adjustment is now determined by its manipulator and its time to build is determined by is micro-laser
del: Removed the time to build information from the exosuit fabricator UI
del: Removed the part description button from the exosuit fabricator UI
/:cl:

### DIFF
--- a/code/game/mecha/mech_fabricator.dm
+++ b/code/game/mecha/mech_fabricator.dm
@@ -1,3 +1,9 @@
+
+#define MECHFAB_MAIN_MENU		1
+#define MECHFAB_CATEGORY_MENU	2
+#define MECHFAB_SEARCH_MENU		3
+#define MECHFAB_MATERIALS_MENU	4
+
 /obj/machinery/mecha_part_fabricator
 	icon = 'icons/obj/robotics.dmi'
 	icon_state = "fab-idle"
@@ -8,39 +14,30 @@
 	use_power = IDLE_POWER_USE
 	idle_power_usage = 20
 	active_power_usage = 5000
-	var/time_coeff = 1
-	var/component_coeff = 1
+	var/time_coeff = 1.15  // Determines build time for parts
+	var/component_coeff = 1.20  // Determines cost of parts
 	var/datum/research/files
 	var/fabricator_type = MECHFAB
-	var/id
-	var/sync = 0
-	var/part_set
-	var/datum/design/being_built
-	var/list/queue = list()
-	var/processing_queue = 0
-	var/screen = "main"
-	var/temp
-	var/list/part_sets = list(
-								"Cyborg",
-								"Cyborg Repair",
-								"Ripley",
-								"Firefighter",
-								"Odysseus",
-								"Gygax",
-								"Durand",
-								"H.O.N.K",
-								"Reticence",
-								"Phazon",
-								"Exosuit Equipment",
-								"Cyborg Upgrade Modules",
-								"Medical",
-								"Misc"
-								)
+	var/temp_search
+	var/wait_message // Used for popping up a wait message when the user presses sync
+	var/wait_message_timer
+	var/datum/design/being_built // The design currently being built
+	var/list/queue = list()  // list of designs currently in the queue
+	var/screen = MECHFAB_MAIN_MENU  // Which screen the mechfab is at. 1 is main menu, 2 is category etc. From the #defines
+	var/selected_category  // Which category the mechfab is currently displaying. See list below
+	var/list/categories = list("Cyborg", "Cyborg Repair", "Cyborg Upgrade Modules", "Ripley", "Firefighter", "Odysseus", "Gygax", "Durand", \
+							   "H.O.N.K", "Reticence", "Phazon", "Exosuit Equipment", "Medical", "Misc")
 
 /obj/machinery/mecha_part_fabricator/New()
-	var/datum/component/material_container/materials = AddComponent(/datum/component/material_container,
-		list(MAT_METAL, MAT_GLASS, MAT_SILVER, MAT_GOLD, MAT_DIAMOND, MAT_PLASMA, MAT_URANIUM, MAT_BANANIUM, MAT_TRANQUILLITE, MAT_TITANIUM, MAT_BLUESPACE), 0,
-		FALSE, /obj/item/stack, CALLBACK(src, .proc/is_insertion_ready), CALLBACK(src, .proc/AfterMaterialInsert))
+	var/datum/component/material_container/materials = AddComponent(
+		/datum/component/material_container,
+		list(MAT_METAL, MAT_GLASS, MAT_SILVER, MAT_GOLD, MAT_DIAMOND, MAT_PLASMA, MAT_URANIUM, MAT_BANANIUM, MAT_TRANQUILLITE, MAT_TITANIUM, MAT_BLUESPACE),
+		0,
+		FALSE,
+		/obj/item/stack,
+		CALLBACK(src, .proc/is_insertion_ready),
+		CALLBACK(src, .proc/AfterMaterialInsert)
+	)
 	materials.precise_insertion = TRUE
 	..()
 	component_parts = list()
@@ -78,82 +75,131 @@
 	GET_COMPONENT(materials, /datum/component/material_container)
 	materials.max_amount = (200000 + (T*50000))
 
-	//resources adjustment coefficient (1 -> 0.85 -> 0.7 -> 0.55)
-	T = 1.15
-	for(var/obj/item/stock_parts/micro_laser/Ma in component_parts)
-		T -= Ma.rating*0.15
-	component_coeff = T
+	// build time adjustment coefficient (1 -> 0.85 -> 0.7 -> 0.55)
+	for(var/obj/item/stock_parts/micro_laser/Ml in component_parts)
+		time_coeff = initial(time_coeff) - Ml.rating*0.15
 
-	//building time adjustment coefficient (1 -> 0.8 -> 0.6)
-	T = -1
-	for(var/obj/item/stock_parts/manipulator/Ml in component_parts)
-		T += Ml.rating
-	time_coeff = round(initial(time_coeff) - (initial(time_coeff)*(T))/5,0.01)
+	// resource cost adjustment coefficient (1 -> 0.8 -> 0.6 -> 0.4)
+	for(var/obj/item/stock_parts/manipulator/Ma in component_parts)
+		component_coeff = initial(component_coeff) - Ma.rating*0.20
 
+/obj/machinery/mecha_part_fabricator/interact(mob/user)
+	ui_interact(user)
 
-/obj/machinery/mecha_part_fabricator/proc/output_parts_list(set_name)
-	var/output = ""
-	for(var/v in files.known_designs)
-		var/datum/design/D = files.known_designs[v]
-		if(D.build_type & fabricator_type)
-			if(!(set_name in D.category))
-				continue
-			output += "<div class='part'>[output_part_info(D)]<br>\["
-			if(check_resources(D))
-				output += "<a href='?src=[UID()];part=[D.id]'>Build</a> | "
-			output += "<a href='?src=[UID()];add_to_queue=[D.id]'>Add to queue</a>\]\[<a href='?src=[UID()];part_desc=[D.id]'>?</a>\]</div>"
-	return output
+/obj/machinery/mecha_part_fabricator/ui_interact(mob/user, ui_key = "main", datum/nanoui/ui = null, force_open = 1)
+	user.set_machine(src)
+	ui = SSnanoui.try_update_ui(user, src, ui_key, ui, force_open)
+	if(!ui)
+		ui = new(user, src, ui_key, "mech_fabricator.tmpl", name, 900, 600)
+		ui.open()
 
-/obj/machinery/mecha_part_fabricator/proc/output_part_info(datum/design/D)
-	var/output = "[initial(D.name)] (Cost: [output_part_cost(D)]) [get_construction_time_w_coeff(D)/10] seconds"
-	return output
+/obj/machinery/mecha_part_fabricator/ui_data(mob/user, ui_key = "main", datum/topic_state/state = GLOB.default_state)
+	var/data[0]
 
-/obj/machinery/mecha_part_fabricator/proc/output_part_cost(datum/design/D)
-	var/i = 0
-	var/output
-	for(var/c in D.materials)
-		output += "[i?" | ":null][get_resource_cost_w_coeff(D, c)] [material2name(c)]"
-		i++
-	return output
+	files.RefreshResearch()
 
-/obj/machinery/mecha_part_fabricator/proc/output_available_resources()
-	var/output
 	GET_COMPONENT(materials, /datum/component/material_container)
-	for(var/mat_id in materials.materials)
-		var/datum/material/M = materials.materials[mat_id]
-		output += "<span class=\"res_name\">[M.name]: </span>[M.amount] cm&sup3;"
-		if(M.amount >= MINERAL_MATERIAL_AMOUNT)
-			output += "<span style='font-size:80%;'>- Remove \[<a href='?src=[UID()];remove_mat=1;material=[mat_id]'>1</a>\]"
-			if(M.amount >= (MINERAL_MATERIAL_AMOUNT * 10))
-				output += " | \[<a href='?src=[UID()];remove_mat=10;material=[mat_id]'>10</a>\]"
-			output += " | \[<a href='?src=[UID()];remove_mat=50;material=[mat_id]'>All</a>\]</span>"
-		output += "<br/>"
-	return output
+	data["screen"] = screen
+	data["total_amount"] = materials.total_amount
+	data["max_amount"] = materials.max_amount
+	data["wait_message"] = wait_message
 
-/obj/machinery/mecha_part_fabricator/proc/get_resources_w_coeff(datum/design/D)
+	switch(screen)
+		if(MECHFAB_MAIN_MENU)
+			data["uid"] = UID()
+			data["categories"] = categories
+
+		if(MECHFAB_CATEGORY_MENU)
+			var/list/designs = list()
+			for(var/I in files.known_designs)
+				var/datum/design/D = files.known_designs[I]
+				if(!(D.build_type & fabricator_type) || !(selected_category in D.category))
+					continue
+				var/list/current_design = list()
+				designs[++designs.len] = current_design
+				current_design["name"] = D.name
+				current_design["id"] = D.id
+				current_design["materials"] = design_cost_data(D)
+			data["designs"] = designs
+			data["selected_category"] = selected_category
+			data["designs_len"] = designs.len
+
+		if(MECHFAB_SEARCH_MENU)
+			var/list/designs = list()
+			for(var/I in files.known_designs)
+				var/datum/design/D = files.known_designs[I]
+				if(!(D.build_type & fabricator_type))
+					continue
+				if(findtext(D.name, temp_search))
+					var/list/current_design = list()
+					designs[++designs.len] = current_design
+					current_design["name"] = D.name
+					current_design["id"] = D.id
+					current_design["materials"] = design_cost_data(D)
+			data["designs"] = designs
+			data["search"] = temp_search
+			data["designs_len"] = designs.len
+
+		if(MECHFAB_MATERIALS_MENU)
+			var/list/materials_list = list()
+			var/list/mats = materials.materials // Materials list in the material_container component. Shortening the name to make it less abnoxious.
+			for(var/M in mats)
+				materials_list[++materials_list.len] = list("name" = mats[M].name, "id" = mats[M].id, "amount" = mats[M].amount)
+			data["loaded_materials"] = materials_list
+
+	data["processing"] = being_built ? "PROCESSING: [initial(being_built.name)]" : null
+	var/list/data_queue = list()
+	if(queue?.len)
+		for(var/datum/design/D in queue)
+			data_queue[++data_queue.len] = list("name" = initial(D.name), "can_build" = can_build(D))
+		data["queue"] = data_queue
+	else
+		data["queue"] = null
+
+	return data
+
+// Gets the cost of each material type from the design as a list
+/obj/machinery/mecha_part_fabricator/proc/design_cost_data(datum/design/D)
+	GET_COMPONENT(materials, /datum/component/material_container)
+	var/list/data = list()
+	var/material_name
+	var/material_amount
+
+	for(var/R in D.materials)
+		material_amount = get_resource_cost_w_coeff(D, R)
+		material_name = CallMaterialName(R) // Formatted name so it looks nice. Ex: Metal, Solid Plasma
+		data[++data.len] = list("name" = material_name, "amount" = material_amount, "is_red" = materials.amount(R) < material_amount)
+	return data
+
+// Returns true if the mech fab can build the specified design
+/obj/machinery/mecha_part_fabricator/proc/can_build(datum/design/D)
+	GET_COMPONENT(materials, /datum/component/material_container)
+	if(materials.has_materials(get_all_resources_w_coeff(D)))
+		return TRUE
+	return FALSE
+
+/obj/machinery/mecha_part_fabricator/proc/get_all_resources_w_coeff(datum/design/D)
 	var/list/resources = list()
 	for(var/R in D.materials)
 		resources[R] = get_resource_cost_w_coeff(D, R)
 	return resources
 
-/obj/machinery/mecha_part_fabricator/proc/check_resources(datum/design/D)
-	if(D.reagents_list.len) // No reagents storage - no reagent designs.
-		return FALSE
-	GET_COMPONENT(materials, /datum/component/material_container)
-	if(materials.has_materials(get_resources_w_coeff(D)))
-		return TRUE
-	return FALSE
+/obj/machinery/mecha_part_fabricator/proc/get_resource_cost_w_coeff(datum/design/D, resource, roundto = 1)
+	return round(D.materials[resource] * component_coeff, roundto)
+
+/obj/machinery/mecha_part_fabricator/proc/get_construction_time_w_coeff(datum/design/D, roundto = 1) //aran
+	return round(initial(D.construction_time) * time_coeff, roundto)
 
 /obj/machinery/mecha_part_fabricator/proc/build_part(datum/design/D)
 	being_built = D
 	desc = "It's building \a [initial(D.name)]."
-	var/list/res_coef = get_resources_w_coeff(D)
+	var/list/res_coef = get_all_resources_w_coeff(D)
 
 	GET_COMPONENT(materials, /datum/component/material_container)
 	materials.use_amount(res_coef)
 	overlays += "fab-active"
 	use_power = ACTIVE_POWER_USE
-	updateUsrDialog()
+	SSnanoui.update_uis(src)
 	sleep(get_construction_time_w_coeff(D))
 	use_power = IDLE_POWER_USE
 	overlays -= "fab-active"
@@ -172,34 +218,26 @@
 	atom_say("[I] is complete.")
 	being_built = null
 
-	updateUsrDialog()
+	SSnanoui.update_uis(src)
 	return TRUE
 
-/obj/machinery/mecha_part_fabricator/proc/update_queue_on_page()
-	send_byjax(usr,"mecha_fabricator.browser","queue",list_queue())
-	return
-
-/obj/machinery/mecha_part_fabricator/proc/add_part_set_to_queue(set_name)
-	if(set_name in part_sets)
-		for(var/v in files.known_designs)
-			var/datum/design/D = files.known_designs[v]
-			if(D.build_type & fabricator_type)
-				if(set_name in D.category)
-					add_to_queue(D)
-
-/obj/machinery/mecha_part_fabricator/proc/add_to_queue(D)
+// Add a single part to the queue
+/obj/machinery/mecha_part_fabricator/proc/add_to_queue(datum/design/D)
 	if(!istype(queue))
 		queue = list()
-	if(D)
+	if(D && istype(D))
 		queue[++queue.len] = D
-	return queue.len
+		return TRUE
+	return FALSE
 
+// Removes an item from the queue at the specified index
 /obj/machinery/mecha_part_fabricator/proc/remove_from_queue(index)
-	if(!isnum(index) || !istype(queue) || (index<1 || index>queue.len))
+	if(!isnum(index) || !istype(queue) || (index < 1 || index > queue.len))
 		return FALSE
-	queue.Cut(index,++index)
+	queue.Cut(index, index + 1)
 	return TRUE
 
+// Starts building whatever is in the queue
 /obj/machinery/mecha_part_fabricator/proc/process_queue()
 	var/datum/design/D = queue[1]
 	if(!D)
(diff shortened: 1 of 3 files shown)


Markolie:
Very nice work!

code/game/mecha/mech_fabricator.dm
```diff
@@ -172,34 +239,26 @@
...
-				if(set_name in D.category)
-					add_to_queue(D)
-
-/obj/machinery/mecha_part_fabricator/proc/add_to_queue(D)
+// Add a single part to the queue
+/obj/machinery/mecha_part_fabricator/proc/add_to_queue(datum/design/D)
 	if(!istype(queue))
 		queue = list()
-	if(D)
+	if(D && istype(D))
```
moxian: This is the same as `if(istype(D))`, right?

code/game/mecha/mech_fabricator.dm
```diff
@@ -172,34 +239,26 @@
...
-		for(var/v in files.known_designs)
-			var/datum/design/D = files.known_designs[v]
-			if(D.build_type & fabricator_type)
-				if(set_name in D.category)
-					add_to_queue(D)
-
-/obj/machinery/mecha_part_fabricator/proc/add_to_queue(D)
+// Add a single part to the queue
+/obj/machinery/mecha_part_fabricator/proc/add_to_queue(datum/design/D)
 	if(!istype(queue))
```
moxian: Unrelated: I'm not a fan of this many safety checks. There's no way D would be null or a non-design type.
There is absolutely no way `istype(queue)` returns false (like what??).
What are you even going to do with the return values? If it's not TRUE, something went terribly wrong already, and you have bigger problems than to handle  the "sorry couldn't add to queue" case.

I mean, I understand where this is coming from, but it all feels very weird.

code/game/mecha/mech_fabricator.dm
```diff
@@ -172,34 +239,26 @@
...
+		return TRUE
+	return FALSE
 
+// Removes an item from the queue at the specified index
 /obj/machinery/mecha_part_fabricator/proc/remove_from_queue(index)
-	if(!isnum(index) || !istype(queue) || (index<1 || index>queue.len))
+	if(!isnum(index) || !istype(queue) || (index < 1 || index > queue.len))
 		return FALSE
-	queue.Cut(index,++index)
+	queue.Remove(queue[index])
```
moxian: `.Remove` is 1) more expensive than `.Cut`, 2) does not always do what you expect.

Scenario: the queue has `["left leg", "honkbot", "right leg", "left arm", "right arm"]`.
You then realize that honkbot is probably not as important as the rest, and try to move it to the end of the queue. To do that you add a new honkbot, giving you `["left leg", "honkbot", "right leg", "left arm", "right arm", "honkbot"]`. Then you click "remove" on the earlier one and get up with `["left leg", "honkbot", "right leg", "left arm", "right arm"]`, exactly what you started with.

Bonus points: this also breaks for "normal" queue processing of `["left leg", "honkbot", "right leg", "left arm", "right arm", "honkbot"]` - both honkbots would get finished before right leg is started.

code/game/mecha/mech_fabricator.dm
```diff
@@ -208,67 +267,47 @@
...
 		RDC.files.push_data(files)
-		temp = "Processed equipment designs.<br>"
-		//check if the tech coefficients have changed
-		temp += "<a href='?src=[UID()];clear_temp=1'>Return</a>"
-
-		updateUsrDialog()
+		SSnanoui.update_uis(src)
 		atom_say("Successfully synchronized with R&D server.")
-		return
+		return TRUE
```
moxian: I don't think there's value in adding return values (ha!) when nobody is checking them.
It's better to leave them bare, and only populate once you know what semantics you *want* to act on (and document that semantics).

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
-				if(D.id == T)
+				if(D.id == design)
 					add_to_queue(D)
 					break
-		return update_queue_on_page()
+
+	if(href_list["category_add_all"])
+		for(var/I in files.known_designs)
+			var/datum/design/D = files.known_designs[I]
+			if(D.build_type & fabricator_type && selected_category in D.category)
```
moxian: Can you pass `selected_category` from the UI here?
Situation: two people operating the same fabricator, one select category A, second switches to category B, the UI does not update for person 1 for whatever reason, person 1 clicks "add all", and adds everything from cat B, which is surprising.

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
-		remove_from_queue(afilter.getNum("remove_from_queue"))
-		return update_queue_on_page()
-	if(href_list["partset_to_queue"])
-		add_part_set_to_queue(afilter.get("partset_to_queue"))
-		return update_queue_on_page()
+		var/index = text2num(href_list["remove_from_queue"])
+		remove_from_queue(index)
+
 	if(href_list["process_queue"])
-		spawn(0)
```
moxian: Removing `spawn(0)` does not look right...
Please revert, or explain why this is safe.

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
-				queue.Swap(index,new_index)
-		return update_queue_on_page()
+		if(processing_queue || being_built)
+			return FALSE
+		processing_queue = TRUE
+		process_queue()
+		processing_queue = FALSE
+
 	if(href_list["clear_queue"])
 		queue = list()
```
moxian: Not your code, but while you're touching it:
use `queue.Cut()` instead.
Subtle difference, and unlikely to be relevant in practice, but `.Cut()` ensures that if we had `process_queue(src.queue)`, then that, passed `queue` gets cleared as well.

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
-		return update_queue_on_page()
+
+	if(href_list["queue_move"] && href_list["index"])
+		var/new_index = text2num(href_list["queue_move"])
+		var/index = text2num(href_list["index"])
+		if(isnum(index) && isnum(new_index))
+			if(IsInRange(new_index, 1, queue.len))
+				queue.Swap(index, new_index)
```
moxian: That's swap, not move.

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
-		materials.retrieve_sheets(text2num(href_list["remove_mat"]), href_list["material"])
+		var/desired_num_sheets
+		if(href_list["remove_mat_amount"] == "custom")
+			desired_num_sheets = input("How many sheets would you like to eject from the machine?", "How much?", 1) as null|num
+			desired_num_sheets = max(0, desired_num_sheets) // If you input too high of a number, the mineral datum will take care of it either way
+			if(!desired_num_sheets)
+				return
+			desired_num_sheets = round(desired_num_sheets) // No partial-sheet goofery
+		else
+			desired_num_sheets = text2num(href_list["remove_mat_amount"])
```
moxian: href_list is user-controlled ("attacker-controlled" if you wish), you still need to sanitize it the same way you do for custom amount (just without the `input()`)

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
+
 	if(href_list["clear_queue"])
 		queue = list()
-		return update_queue_on_page()
+
+	if(href_list["queue_move"] && href_list["index"])
+		var/new_index = text2num(href_list["queue_move"])
+		var/index = text2num(href_list["index"])
+		if(isnum(index) && isnum(new_index))
+			if(IsInRange(new_index, 1, queue.len))
```
moxian: you're checking only new_index, but not the old_index

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +320,103 @@
...
+				return
+			desired_num_sheets = round(desired_num_sheets) // No partial-sheet goofery
+		else
+			desired_num_sheets = text2num(href_list["remove_mat_amount"])
+		materials.retrieve_sheets(desired_num_sheets, href_list["remove_mat"])
 
-	updateUsrDialog()
-	return
+	SSnanoui.update_uis(src)
+	return TRUE
```
moxian: useless return value?

code/game/mecha/mech_fabricator.dm
```diff
@@ -475,7 +461,15 @@
 	component_parts += new /obj/item/stock_parts/micro_laser(null)
 	component_parts += new /obj/item/stack/sheet/glass(null)
 	RefreshParts()
+	files = new /datum/research(src) //Setup the research data holder.
+	matching_designs = list()
+
+/obj/machinery/mecha_part_fabricator/spacepod/interact(mob/user)
+	. = ..()
```
moxian: Do you need to write this? If you remove this proc does this not automatically happen? (I'm don't know)

code/game/mecha/mech_fabricator.dm
```diff
@@ -1,3 +1,10 @@
+
+#define MECHFAB_MAIN_MENU		1
+#define MECHFAB_CATEGORY_MENU	2
```
moxian: weird indentation

code/game/mecha/mech_fabricator.dm
```diff
@@ -1,3 +1,10 @@
+
+#define MECHFAB_MAIN_MENU		1
+#define MECHFAB_CATEGORY_MENU	2
+#define MECHFAB_SEARCH_MENU		3
+#define MECHFAB_MATERIALS_MENU	4
```
moxian: I suggest using strings instead of numbers here. AFAIK comparing those is just as cheap, but you get a tiny bit of debug info that way.

moxian:
Oh, and to elaborate on/clarify my above comment a bit: a lot of the changes you've made here have nothing to do with nanoUI. *Which is* why a lot of my comments have nothing to do with the code *you* wrote.
If you rework this PR to only touch nanoUI (and opt out of stylistic to unrelated code), I'll only comment on that, but in its current state I treat it more as a complete mechfab refactor (and this is probably a bad thing).

code/game/mecha/mech_fabricator.dm
SteelSlayer: >.Remove is 1) more expensive than .Cut, 2) does not always do what you expect.

Perhaps. How do you know this/where did you see this information?

Also I could not get Cut to simply cut a single value. I don't remember exactly why I couldn't get it working, as I changed this early in life of this project. So I switched to remove instead.

But I see the problem you're talking about. I tried adding a duplicate item at the end of the queue and removing the old item further up but it instead removes the one on the end. I have NO idea why this is happening because I'm passing the specific index of the item from the template.

code/game/mecha/mech_fabricator.dm
SteelSlayer: fair enough

code/game/mecha/mech_fabricator.dm
SteelSlayer: Not sure what you mean. Queue moving works fine though. In the gif, you can see I move a single item from the end of the queue, to the beginning and then back to the end.

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,72 +75,142 @@
...
+			materials_list[++materials_list.len] = list("name" = "Solid Plasma", "id" = MAT_PLASMA, "amount" = materials.amount(MAT_PLASMA))
+			materials_list[++materials_list.len] = list("name" = "Uranium", "id" = MAT_URANIUM, "amount" = materials.amount(MAT_URANIUM))
+			materials_list[++materials_list.len] = list("name" = "Diamond", "id" = MAT_DIAMOND, "amount" = materials.amount(MAT_DIAMOND))
+			materials_list[++materials_list.len] = list("name" = "Bananium", "id" = MAT_BANANIUM, "amount" = materials.amount(MAT_BANANIUM))
+			materials_list[++materials_list.len] = list("name" = "Tranquillite", "id" = MAT_TRANQUILLITE, "amount" = materials.amount(MAT_TRANQUILLITE))
+			materials_list[++materials_list.len] = list("name" = "Titanium", "id" = MAT_TITANIUM, "amount" = materials.amount(MAT_TITANIUM))
+			materials_list[++materials_list.len] = list("name" = "Plastic", "id" = MAT_PLASTIC, "amount" = materials.amount(MAT_PLASTIC))
+			materials_list[++materials_list.len] = list("name" = "Bluespace Mesh", "id" = MAT_BLUESPACE, "amount" = materials.amount(MAT_BLUESPACE))
```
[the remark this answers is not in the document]
SteelSlayer: Yeah I wonder if something like this would be better: 
Make a `var/list/materials_list` innate to the mech fabricator instead of locally on line 143. Loop through every material in the material datum and add each material list to `materials_list`. That way its not hardcoded.

That way when the `data["loaded_materials"] = materials_list` line executes it will just assign all the information we already assigned to `materials_list`.

code/game/mecha/mech_fabricator.dm
TDSSS: If I understand it right, spawn spawns a separate process at that point after the wait time. Spawn(0) makes it so the rest doesn't wait for this stuff to finish I think.

code/game/mecha/mech_fabricator.dm
moxian: You check that `new_index` is between `1` and `queue.len`
You do NOT check that `index` is between `1` and `queue.len`.
That's inconsistent (and prone to (harmless) runtimes).

Separately, does `.Swap` runtime if both its arguments are the same?

code/game/mecha/mech_fabricator.dm
moxian: > Queue moving works fine though. In the gif, you can see I move a single item from the end of the queue, to the beginning and then back to the end.

That's true! It does indeed work fine since you only move the item by 1.
For some reason I thought the "move" functionality involves fancy drag-and-drop, and that wouldn't work with your current code (having a list of `["a", "b", "c", "d"]` and issuing `queue_move` with `index=2` and `old_index=4` would give `["a", "d", "c", "b"]` not `["a", "d", "b", "c"]`).

So, yeah, feel free to ignore this.

On a mostly unrelated note, consider having `index` and `offset` (the latter beging `1` or `-1`) instead of `old_index` and `new_index`. (I'm not(!) saying it *would* be better that way, just that it might, and that it's a thing that can be considered).

code/game/mecha/mech_fabricator.dm
```diff
@@ -8,34 +15,21 @@
...
-								"Cyborg Upgrade Modules",
-								"Medical",
-								"Misc"
-								)
+	var/temp_search
+	var/wait_message // Used for popping up a wait message when the user presses sync
+	var/wait_message_timer
+	var/list/datum/design/matching_designs  // Used for designs that will appear after a search
+	var/datum/design/being_built // The design currently being built
+	var/list/queue = list()  // list of designs currently in the queue
```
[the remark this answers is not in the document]
moxian: That's not what I mean.
(If my understanding of the DM object model is right) creating a list here in the object definition means that every single fabricator has this particular `list()` object as their `queue` field.
I.e. if you create 5 fabricators, and add something to a queue in one of them, then ALL of them will have that item added. Since the queue is literally shared.
Moving the initialization to something like `New()` would force DM to create a new `list()` for the queue each time a new fabricator is spawned in the world.

Does that make sense?

code/game/mecha/mech_fabricator.dm
moxian: Right, but isn't `if(D != null && istype(D))` equivalent to `if(istype(D))` ?

code/game/mecha/mech_fabricator.dm
moxian: > How do you know this/where did you see this information?

The "more expensive" part? Generic algorithmic knowledge.
To remove an item by value from the list you need to inspect each element if it equals that value, and if it does - remove that index.
To remove an item by index - you just remove that index, skipping the "check for equality" part.

code/game/mecha/mech_fabricator.dm
SteelSlayer: Yeah I don't think queue would ever not be a list, so this is probably unnecessary. But it was here before so I just left it.

code/game/mecha/mech_fabricator.dm
SteelSlayer: Now that I take a closer look, I know what you mean, I did `Remove(queue[index])` which is a value instead of just doing `Remove(index)` which, I don't think Remove can even do. So I see why Cut is preferred, you can feed it indexes.

code/game/mecha/mech_fabricator.dm
SteelSlayer: Hm, I don't quite understand this yet so I'm going to add it back in.

code/game/mecha/mech_fabricator.dm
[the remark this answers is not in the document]
SteelSlayer: Oh I get what you mean, that is not the case however. Each fabricator's `queue` will be unique to that specific machine. If you add something to one, it won't be added to all of them.

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,72 +75,142 @@
...
+					design["name"] = D.name
+					design["id"] = D.id
+					design["materials"] = design_cost_data(D)
+			data["designs_len"] = designs.len
+
+		if(MECHFAB_SEARCH_MENU)
+			data["search"] = temp_search
+			var/list/designs = list()
+			data["designs"] = designs
+			for(var/datum/design/D in matching_designs)
```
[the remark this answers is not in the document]
SteelSlayer: I'm not sure what you mean here, can you explain more?

code/game/mecha/mech_fabricator.dm
SteelSlayer: I'm honestly not sure what istype does if you do `istype(null)`. I notice in most cases these types of safety checks include both of these, similar to how I have it written.

code/game/mecha/mech_fabricator.dm
SteelSlayer: You're right, adding a check for index. Also, can't say for certain if Swap runtimes on indentical arguments, I'm assuming it doesn't but if I find out it does, I can't add a check making sure they're unique.

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,72 +75,142 @@
...
+			materials_list[++materials_list.len] = list("name" = "Tranquillite", "id" = MAT_TRANQUILLITE, "amount" = materials.amount(MAT_TRANQUILLITE))
+			materials_list[++materials_list.len] = list("name" = "Titanium", "id" = MAT_TITANIUM, "amount" = materials.amount(MAT_TITANIUM))
+			materials_list[++materials_list.len] = list("name" = "Plastic", "id" = MAT_PLASTIC, "amount" = materials.amount(MAT_PLASTIC))
+			materials_list[++materials_list.len] = list("name" = "Bluespace Mesh", "id" = MAT_BLUESPACE, "amount" = materials.amount(MAT_BLUESPACE))
+
+	data = queue_data(data)
+	return data
+
+// Appends information about the queue to the "data" list in ui_data
+/obj/machinery/mecha_part_fabricator/proc/queue_data(list/data)
```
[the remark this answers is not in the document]
SteelSlayer: Similar to one of your other suggestions, I didn't even need this proc anymore. I just stuck it inside `ui_data` in place of `data = queue_data(data)` and it works fine.

code/game/mecha/mech_fabricator.dm
```diff
@@ -406,10 +381,12 @@
 			desired_num_sheets = round(desired_num_sheets) // No partial-sheet goofery
 		else
 			desired_num_sheets = text2num(href_list["remove_mat_amount"])
+			if(!isnum(desired_num_sheets))
+				CRASH("A non-number got passed to a mach fabricator's remove_mat section")
```
moxian: You can still "remove" negative amount of sheets this way.

code/game/mecha/mech_fabricator.dm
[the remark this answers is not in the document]
moxian: Currently, when search happens (in `Topic`), you populate `matching_designs` with list of all matching designs, then pass the compiled list to the display function (here, in `ui_data`).
What I suggest is not storing the `matching_designs` list as the object variable at all, but instead store only the search terms (which you store anyway) and create it right here in `ui_data` function.

code/game/mecha/mech_fabricator.dm
```diff
@@ -76,22 +80,17 @@
...
-		T += Ml.rating
-	component_coeff = round(initial(time_coeff) - (T / 5), 0.01)
+	for(var/obj/item/stock_parts/manipulator/Ma in component_parts)
+		component_coeff -= Ma.rating*0.20
 
 /obj/machinery/mecha_part_fabricator/interact(mob/user)
 	ui_interact(user)
 
 /obj/machinery/mecha_part_fabricator/ui_interact(mob/user, ui_key = "main", datum/nanoui/ui = null, force_open = 1)
-	user.set_machine(src)
```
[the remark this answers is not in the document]
SteelSlayer: Oh, I removed this because I though it did something else, but after some research, I actually need this. It ultimately tells the machine if someone is using it, and if it has to start processing/updating things.

code/game/mecha/mech_fabricator.dm
SteelSlayer: After doing some testing, I don't think this will ever happen. Even if the UI doesn't update for the second person, as soon as he clicks to go to a new category, `selected_category` gets updated, which means the adding would fail for the other person (if they are still somehow are viewing the old category).

However, I can pass `data.selected_category` from nano for extra safety, but I'm not sure its necessary.

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,82 +75,131 @@
...
-	for(var/c in D.materials)
-		output += "[i?" | ":null][get_resource_cost_w_coeff(D, c)] [material2name(c)]"
-		i++
-	return output
-
-/obj/machinery/mecha_part_fabricator/proc/output_available_resources()
-	var/output
+	// build time adjustment coefficient (1 -> 0.85 -> 0.7 -> 0.55)
+	for(var/obj/item/stock_parts/micro_laser/Ml in component_parts)
+		time_coeff = initial(time_coeff) - Ml.rating*0.15
```
farie82, line 80: spacing

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,82 +75,131 @@
...
+
+	// resource cost adjustment coefficient (1 -> 0.8 -> 0.6 -> 0.4)
+	for(var/obj/item/stock_parts/manipulator/Ma in component_parts)
+		component_coeff = initial(component_coeff) - Ma.rating*0.20
```
farie82, line 84: spacing

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,82 +75,131 @@
...
+				if(!(D.build_type & fabricator_type) || !(selected_category in D.category))
+					continue
+				var/list/current_design = list()
+				designs[++designs.len] = current_design
+				current_design["name"] = D.name
+				current_design["id"] = D.id
+				current_design["materials"] = design_cost_data(D)
+			data["designs"] = designs
+			data["selected_category"] = selected_category
+			data["designs_len"] = designs.len
```
farie82, line 125: `LAZYLEN` is prefered over `.len`

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,82 +75,131 @@
...
+					continue
+				if(findtext(D.name, temp_search))
+					var/list/current_design = list()
+					designs[++designs.len] = current_design
+					current_design["name"] = D.name
+					current_design["id"] = D.id
+					current_design["materials"] = design_cost_data(D)
+			data["designs"] = designs
+			data["search"] = temp_search
+			data["designs_len"] = designs.len
```
farie82, line 141: Same as above

code/game/mecha/mech_fabricator.dm
```diff
@@ -78,82 +75,131 @@
...
+		if(MECHFAB_MATERIALS_MENU)
+			var/list/materials_list = list()
+			var/list/mats = materials.materials // Materials list in the material_container component. Shortening the name to make it less abnoxious.
+			for(var/M in mats)
+				materials_list[++materials_list.len] = list("name" = mats[M].name, "id" = mats[M].id, "amount" = mats[M].amount)
+			data["loaded_materials"] = materials_list
+
+	data["processing"] = being_built ? "PROCESSING: [initial(being_built.name)]" : null
+	var/list/data_queue = list()
+	if(queue?.len)
```
farie82, line 152: using `LAZYLEN(queue)` will result in the same behaviour here

code/game/mecha/mech_fabricator.dm
```diff
@@ -172,34 +218,26 @@
...
+	if(D && istype(D))
 		queue[++queue.len] = D
-	return queue.len
+		return TRUE
+	return FALSE
 
+// Removes an item from the queue at the specified index
 /obj/machinery/mecha_part_fabricator/proc/remove_from_queue(index)
-	if(!isnum(index) || !istype(queue) || (index<1 || index>queue.len))
+	if(!isnum(index) || !istype(queue) || (index < 1 || index > queue.len))
```
farie82, line 235: `LAZYLEN` here as well

code/game/mecha/mech_fabricator.dm
```diff
@@ -208,67 +246,46 @@
...
+	wait_message = message
+	wait_message_timer = addtimer(CALLBACK(src, .proc/clear_wait_message), delay, TIMER_UNIQUE | TIMER_STOPPABLE)
 
-/obj/machinery/mecha_part_fabricator/proc/get_construction_time_w_coeff(datum/design/D, roundto = 1) //aran
-	return round(initial(D.construction_time)*time_coeff, roundto)
+/obj/machinery/mecha_part_fabricator/proc/clear_wait_message()
+	wait_message = ""
+	if(wait_message_timer)
+		deltimer(wait_message_timer)
+		wait_message_timer = 0
```
farie82, line 287: `null` would make more sense here

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +298,99 @@
...
-				</table>"}
-	var/datum/browser/popup = new(user, "mecha_fabricator", name, 1000, 600)
-	popup.set_content(dat)
-	popup.open(0)
-	onclose(user, "mecha_fabricator")
-	return
-
 /obj/machinery/mecha_part_fabricator/Topic(href, href_list)
 	if(..())
 		return 1
```
farie82, line 303: `TRUE`

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +298,99 @@
...
-		remove_from_queue(afilter.getNum("remove_from_queue"))
-		return update_queue_on_page()
-	if(href_list["partset_to_queue"])
-		add_part_set_to_queue(afilter.get("partset_to_queue"))
-		return update_queue_on_page()
+		var/index = text2num(href_list["remove_from_queue"])
+		remove_from_queue(index)
+
 	if(href_list["process_queue"])
 		spawn(0)
```
farie82, line 348: Gross. `INVOKE_ASYNC` would be nice here as an added refactor

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +298,99 @@
...
 	if(href_list["clear_queue"])
-		queue = list()
-		return update_queue_on_page()
+		queue.Cut()
+
+	if(href_list["curr_index"] && href_list["new_index"])
+		var/curr_index = text2num(href_list["curr_index"])
+		var/new_index = text2num(href_list["new_index"])
+		if(isnum(curr_index) && isnum(new_index))
+			if(IsInRange(curr_index, 1, queue.len) && IsInRange(new_index, 1, queue.len))
```
farie82, line 360: `LAZYLEN` here as well. Can cache the result

code/game/mecha/mech_fabricator.dm
```diff
@@ -281,151 +298,99 @@
...
+		if(href_list["remove_mat_amount"] == "custom")
+			desired_num_sheets = input("How many sheets would you like to eject from the machine?", "How much?", 1) as null|num
+			if(!desired_num_sheets)
+				return
+		else
+			desired_num_sheets = text2num(href_list["remove_mat_amount"])
 
-	updateUsrDialog()
+		if(!isnum(desired_num_sheets))
+			CRASH("A non-number got passed to a mach fabricator's remove_mat section")
```
farie82, line 380: type here. mach -> mech. Maybe also adding the value in the crash would be nice

code/game/mecha/mech_fabricator.dm
```diff
@@ -434,7 +399,7 @@
 	if(exchange_parts(user, W))
 		return
 
-	if(default_deconstruction_crowbar(user, W))
+	if(default_deconstruction_crowbar(W))
```
farie82, line 402: `user` is needed here as first param.

code/game/mecha/mech_fabricator.dm
```diff
@@ -1,3 +1,9 @@
+
+#define MECHFAB_MAIN_MENU		1
+#define MECHFAB_CATEGORY_MENU	2
+#define MECHFAB_SEARCH_MENU		3
+#define MECHFAB_MATERIALS_MENU	4
```
farie82, line 5: These are not undefined at the end of the file

SteelSlayer:
I am closing this because we're *probably* getting TGUI soon and its kinda pointless to work on this anymore with that coming. The DM code for the TGUI mech fab will be 95% the same so its not a huge deal. Thanks to everyone who reviewed this, your reviews will still be put to use whenever I remake this PR for TGUI in the future.

PR #12795, opened 2019-12-12
Emagproofs CC, adds admin warning if ert shuttle is moved by non-ert

TDSSS:
## What Does This PR Do
This replaces the blast doors in CC with an emag-proof variant used in the syndicate base as well.

Edit: It also includes a message informing admins whenever a player that is not an ert, a deathsquad or himself an admin moves the ert shuttle, so they do not miss people sneaking into CC.

## Why It's Good For The Game
It stops a traitor that manages to steal an ert shuttle from raiding the gamma armory or firing the CC BSA.

Admins can now appropriately react to anyone attempting to attack CC via the shuttle if they miss it.

## Images of changes
the blast doors look exactly the same as the normal version, so no visual changes.

## Changelog
:cl:
tweak: CC is now emag-proof
fix: Minor copy-paste error
tweak: admins are now messaged if a non-ert/DS player moves the ert shuttle.
/:cl:

### DIFF
--- a/code/modules/shuttle/ert.dm
+++ b/code/modules/shuttle/ert.dm
@@ -9,7 +9,7 @@
 	if(href_list["move"])
 		var/authorized_roles = list(SPECIAL_ROLE_ERT, SPECIAL_ROLE_DEATHSQUAD)
 		if(!((usr.mind.assigned_role in authorized_roles) || is_admin(usr)))
-			message_admins("Potential ERT shuttle hijack, ert shuttle moved by unauthorized user: [key_name_admin(usr)]")
+			message_admins("Potential ERT shuttle hijack, ERT shuttle moved by unauthorized user: [key_name_admin(usr)]")
 	..()
 	
 
--- a/code/modules/shuttle/ert.dm
+++ b/code/modules/shuttle/ert.dm
@@ -5,6 +5,14 @@
 	possible_destinations = "specops_home;specops_away;specops_custom"
 	resistance_flags = INDESTRUCTIBLE
 
+/obj/machinery/computer/shuttle/ert/Topic(href, href_list)
+	if(href_list["move"])
+		var/authorized_roles = list(SPECIAL_ROLE_ERT, SPECIAL_ROLE_DEATHSQUAD)
+		if(!((usr.mind.assigned_role in authorized_roles) || is_admin(usr)))
+			message_admins("Potential ERT shuttle hijack, ert shuttle moved by unauthorized user: [key_name_admin(usr)]")
+	..()
+	
+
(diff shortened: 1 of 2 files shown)


Kyep:
This is probably not a good idea. Once someone gets to CC they have other potential methods of getting past the blast doors, like ripley drills.

If you want to stop people breaking into CC, the best way to do it would be to secure the ERT shuttle. Make it a subtype of the ferry, so it requires admin permission to return to CentCom.
Alternatively, don't secure it, just make it create an alert for admins when it is moved back to CentCom and anyone other than ERT is on it.

TDSSS:
The impassable blast doors have
resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | UNACIDABLE | ACID_PROOF
I am fairly sure that prevents ripley drills too, will test in a bit.

Kyep:
This PR is probably not going to pass unless you change it to focus on securing the shuttle, instead of altering things on z2.

ppi13:
Originally the shuttle console was emag proof, change the shuttle console instead, you still need to find a dead ert or kill one to properly get there, and if you can get there, well gamma armory is protected by a pulse laser, however you can pretend to be a centcom official is you also use a dna scrambler and emag yourself to the super-all-access id, this will also allow you to access the mechs.  
Generally outside of hijack you can't do much other then probably create some amazing RP, and more RP is always healthy.

TDSSS:
After having a talk with Kyet, included a warning to online admins when the shuttle is moved by a player that's not an ert, deathsquad or an admin.

code/modules/shuttle/ert.dm
```diff
@@ -5,6 +5,14 @@
 	possible_destinations = "specops_home;specops_away;specops_custom"
 	resistance_flags = INDESTRUCTIBLE
 
+/obj/machinery/computer/shuttle/ert/Topic(href, href_list)
+	if(href_list["move"])
+		var/authorized_roles = list(SPECIAL_ROLE_ERT, SPECIAL_ROLE_DEATHSQUAD)
+		if(!((usr.mind.assigned_role in authorized_roles) || is_admin(usr)))
```
farie82: More a design point:
This would mean that admins themselves as characters won't trigger the 'alarm'. If this is undesired then a simple check on the mobtype of the users mob will fix it. (aghosting etc)

TDSSS: Yeah that is intended behaviour. If an admin spawns in and takes the shuttle there or back, he doesn't want an alert. And if an admin as normal station crew takes the shuttle back to CC, then he's already aware of what's going on.

PR #12798, opened 2019-12-12
Move the handle_changeling proc call to the (new) life signal

farie82:
## What Does This PR Do
Makes it so that handle_changeling is not called every Life tick for every carbon and instead uses a signal.

Code is from #12552

## Why It's Good For The Game
More modular and probably a bit faster on performance

## Changelog
No CL since nothing feature wise changed


code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -421,6 +424,8 @@ var/list/sting_paths
 		if(hud_used)
 			hud_used.lingstingdisplay.icon_state = null
 			hud_used.lingstingdisplay.invisibility = 101
+			hud_used.lingchemdisplay.invisibility = 101
```
farie82: Required since handle_changeling used to set this to 101 if you were not a cling. Refactor of handle_changeling can be done as well but I'll keep it out of the PR

PR #12119, opened 2019-08-28
The Great Engineering / South Remap

FreeStylaLT:
**What does this PR do:**
Heeeere we go.
Core aspects of this PR:

- Adds a second Engine (Supermatter), straight from #11311 
- Remove Mechanic / Assembly Line hallway, which served no purpose (Similarly to North remap's removal of clown / mime / IAA / magi hallway)
- Remaps Engineering to feel more like a proper department rather than an inter-connected room salad. (Basically adds a hallway to Engineering)
- Moves Turbine / Engineering shuttle into Engi-maint "space pockets" to make them much more practically accessible.
- Connects Arrival maint with Engineering maint. More info on that at pics!
- Adds a couple of new maint areas

### https://affectedarc07.co.uk/ParaMapGen/Images/150.png

As you can see, it's quite a big one.

Let's take it from the top, or, rather, the bottom:

### SOUTH HALLWAY / ENGI LOBBY:
**Hallway:**
img
First thing you might notice is that the giant reception schlong from Engineering is now gone. Instead, replaced by Blueshield / NT Rep, the janitor and a public construction area. The goal here was to turn the hallway into less of an Engi-only place and into more of a proper hallway used by jobs.

The construction area has also been moved to the public, to fill the niche of a lack of a public building spot for the crew, especially relevant with the various build-oriented station objectives. Engineering itself has no trouble building anything anywhere, so the area was fairly useless for them.

**Auxiliary storage** has also been moved to the construction area. It made little sense to have two tool / engineering related storages so close to each other in Arrivals. This is the replacement:

img

**Engineering Lobby:**
img
The entire schlong has been shrunk into two rooms, with a proper, far more practical reception. A public EVA-like area has also been added, to fill in the niche of Engineering supplying the crew in case of emergencies more practically than three firefighting lockers.

Tech storage has been moved to be adjacent to Engineering's lobby, and the secure storage area is now visible from Engineering in general, thus increasing the security of that easily-ransackable area.

Assembly line has also been moved to be directly part of Engineering, but also having direct adjacency to public maint. This felt like the best of two worlds that the current iteration tries to accomplish (Easily accessible by Engineering and public). Since Construction area is now public, having Assembly line be more private felt like a fair exchange.

### ENGINEERING:
img
[...]

PR #11311, opened 2019-04-17
Supermatter port

Fox-McCloud:
Ports over TG's supermatter. 

So begins the ending of the age of the Singulo and Tesla

TG's supermatter fully replaces the singulo and Tesla. You will still be able to construct the singulo and Tesla, but it being a round-start engine available for players is gone (order it from cargo). 

The supermatter is an engine that interacts with gas, pressure, and temperature, while giving off radiation. For those of you interested in the finer details of the engine, I recommend TG's guide: https://tgstation13.org/wiki/Guide_to_the_Supermatter

The engine can be ran "hot" to intentionally produce anomalies, which can then be deactivated and given to science for bonuses in research. Of course, the anomalies themselves are dangerous, and if your'e not careful, things could spiral out of control and the engine could ultimately get away from you in your attempt to engage in this.

If the supermatter reaches critical point, it'll delaminate, causing a hug explosions while bathing the station in radiation and causing everyone to hallucinate.

If you managed to have a lot of gas in the chamber of the supermatter when it delaminates, it can compress down into a super singularity, which will begin consuming the station.

Alternatively, if the supermatter is particularly engergetic when it delamintes, it'll explode into an energy ball that'll begin shocking the station into oblivion.

pic:
img

:cl: Fox McCloud
add: Updates the supermatter
/:cl:

PR #12799, opened 2019-12-12
Fixes the construct reflect and refactors the reflect code doing it

farie82:
## What Does This PR Do
Fixes #12782
Refactors other parts of the code who all do the same but just copy pasta'd the code. The construct code was one that didn't fully copy pasta all the code.

## Why It's Good For The Game
Less bugs more gut

## Images of changes
image

## Changelog
:cl:
fix: Constructs now correctly reflect energy and beam projectiles
/:cl:

PR #12808, opened 2019-12-15
Fixes reflectors and a Projectile angle bug

FreeStylaLT:
## What Does This PR Do
- Fixes Reflectors not reflecting at all 
- - Reflected projectiles needed an angle to be set to them. This happens now.
- Fixes Reflectors (and anything else) reflecting incorrectly NORTH (angle 0) in projectile.dm
- - null angle and 0 angle were treated the same (as a non-existing angle), meaning it'd get reset. 0th angle, however, is the reflection to North.

## Why It's Good For The Game
Makes reflections, you know, work.
This bug is also integral to both #11311 and #12119 due to the SM relying on reflectors almost completely. Merge this before you merge any of those.

## Images of changes
https://streamable.com/tty26
Reflectors work properly now at all angles.

## Changelog
:cl: FreeStylaLT
fix: Reflectors not reflecting
fix: North reflections in general
/:cl:

### DIFF
--- a/code/game/objects/structures/reflector.dm
+++ b/code/game/objects/structures/reflector.dm
@@ -12,9 +12,9 @@
 /obj/structure/reflector/bullet_act(obj/item/projectile/P)
 	var/turf/reflector_turf = get_turf(src)
 	var/turf/reflect_turf
-	var/new_dir = get_reflection(dir, P.dir)
 	if(!istype(P, /obj/item/projectile/beam))
 		return ..()
+	var/new_dir = get_reflection(dir, P.dir)
 	if(new_dir)
 		reflect_turf = get_step(reflect_turf, new_dir)
 	else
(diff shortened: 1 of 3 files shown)


FreeStylaLT:
This also shouldn't conflict with #12799 as we're handling different parts of reflection

code/game/objects/structures/reflector.dm
```diff
@@ -30,6 +30,10 @@
 	P.current = reflector_turf
 	P.yo = reflect_turf.y - reflector_turf.y
 	P.xo = reflect_turf.x - reflector_turf.x
+	P.ignore_source_check = TRUE		//If shot by a laser, will now hit the mob that fired it
+	var/reflect_angle = dir2angle(new_dir)
+	P.setAngle(reflect_angle)
```
FreeStylaLT: This is all there was to make the Reflectors actually reflect

code/modules/power/singularity/emitter.dm
```diff
@@ -254,6 +254,7 @@
 		shot_number = 0
 	P.setDir(dir)
 	P.starting = loc
+	P.Angle = null
```
FreeStylaLT: Without this, Emitters would fire at angle 0, IE "north" regardless of dir.

Kyep:
Thank you for doing this.

code/game/objects/structures/reflector.dm
farie82: simple solutions are the best usually. Good find

PR #12810, opened 2019-12-16
Makes the Handheld Defib not sleep people who are not in crit.

Qwertytoforty:
## What Does This PR Do

This changes the handheld defibrillator, from sleeping people who are not in crit when safeties are disabled..

## Why It's Good For The Game

Handheld defibrillators are items easily available to anyone, being found all over medical, or mass produced in science. If the defibrillator is emaged or emped, it used to apply Paralysis(3) to the target upon using it on them. If you have 3 defibs, you would be able to instantly and permanently sleep someone, preventing them from calling for help and stunning them in one hit. Worse, if you combine the defibrillator with a soul stone, you could instantly sleep and shard an officer by standing beside them, zapping, and switching hands. This PR makes it so the person must be in crit, or undergoing cardiac arrest, for the paralysis affect to be applied.

## Changelog
:cl:
balance: Handheld defibrillator only paralyses the target if they are in crit or undergoing cardiac arrest.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/handheld_defib.dm
+++ b/code/game/objects/items/devices/handheld_defib.dm
@@ -63,7 +63,9 @@
 				else
 					to_chat(user, "<span class='danger'>[H] doesn't respond!</span>")
 
-			H.AdjustParalysis(3)
+			if((H.health <= HEALTH_THRESHOLD_CRIT) || (H.undergoing_cardiac_arrest())) //Don't sleep if they aren't in crit
+				H.AdjustParalysis(3)
+
 			H.AdjustStunned(5)
 			H.AdjustWeakened(5)
 			H.AdjustStuttering(10)
--- a/code/game/objects/items/devices/handheld_defib.dm
+++ b/code/game/objects/items/devices/handheld_defib.dm
@@ -52,6 +52,14 @@
 		if(H.stat == DEAD)
 			to_chat(user, "<span class='danger'>[H] doesn't respond at all!</span>")
 		else
+
+			if((H.health <= HEALTH_THRESHOLD_CRIT) || (H.undergoing_cardiac_arrest())) //Don't sleep if they aren't in crit
+				H.AdjustParalysis(3)
+
+			H.AdjustStunned(5)
+			H.AdjustWeakened(5)
+			H.AdjustStuttering(10)
+
 			H.set_heartattack(FALSE)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/handheld_defib.dm
```diff
@@ -63,7 +63,9 @@
 				else
 					to_chat(user, "<span class='danger'>[H] doesn't respond!</span>")
 
-			H.AdjustParalysis(3)
+			if((H.health <= HEALTH_THRESHOLD_CRIT) || (H.undergoing_cardiac_arrest())) //Don't sleep if they aren't in crit
```
farie82: undergoing_cardiac_arrest won't work seeing it's set to false prior on line 55

Qwertytoforty: Moved it and the surrounding stun to above the line that sets heart attack to false

TDSSS:
>If you have 3 defibs, you would be able to instantly and permanently sleep someone, preventing them from calling for help and stunning them in one hit.

that seems like a rather big edge case. If people abuse it as a stun weapon too much, maybe just put a short delay on it during which you have to stay still. Emagged equipment is supposed to be dangerous.

Qwertytoforty:
> 
> 
> > If you have 3 defibs, you would be able to instantly and permanently sleep someone, preventing them from calling for help and stunning them in one hit.
> 
> that seems like a rather big edge case. If people abuse it as a stun weapon too much, maybe just put a short delay on it during which you have to stay still. Emagged equipment is supposed to be dangerous.

The emagged flag is set to true from an emp act, as well as an emag, making it quite easy to obtain even to people who don't have emags, like cult. Perhaps adding a delay to the stun would work, however this normally does not need a delay to function, and would still sleep people incredibly fast, preventing calls for help or still quick sharding. 

The only thing this does is to stop quick instant sleep, which was the goal of this pr.

Necaladun:
I think this goes a bit too far to nerfing them, some kinda delay to make it more akin to a normal stun weapon is best, not making it near useless - stunning someone already in crit is basically useless.

Qwertytoforty:
Sleeping was the only change I new how to change, and the main problem, so only other change would be to allow shields / hard suits / shielded hard suit to block it, someone else can do that if they want.

PR #12775, opened 2019-12-05
Updates links for new domain

Kyep:
## What Does This PR Do
Updates some links on our github to point to our new domain.

## Why It's Good For The Game
Nanotrasen.se is depreciated - we now use ParadiseStation.org for everything.

## Changelog
:cl: Kyep
tweak: Updated github links to point to our new domain.
/:cl:

PR #12811, opened 2019-12-17
Domain link updates 2

Kyep:
## What Does This PR Do
Updates various links in code/comments that still point at our old domain name.
Corrects various links that still point at phpBB URLs that no longer work.

In #12775 some cases were missed, because they did not show up in github search.
This version uses a recursive grep to catch every remaining case.
It also goes after (and fixes) phpbb links in general (except for ones to tg's forums, those are still valid).

## Why It's Good For The Game
Fixes out of date links.

## Changelog
:cl: Kyep
fix: Fixed more references to our old domain name.
/:cl:


AffectedArc07:
Question


Shouldn’t these links really be using the domain config option to avoid issues like this?

PR #12815, opened 2019-12-17
Clowns learns clownish on spawn

datlo:
With a grant from clown planet, Nanotrasen is now teaching Clownish classes to all hired clowns.

Clowns now get clownish on spawning. 

The reason for this change is that clowns should be the one job on the station that oughta speak clownish, and that due to it I suspect a lot of people have clownish as their second language just because they play clown now and then. 

This change should hopefully see people using other languages, as they are guaranteed to know clownish when they do play clown.

image

:cl:
add: Clowns can now always speak clownish.
/:cl:


MrMagolor:
10/10, now I won't need to pray whenever I'm clown as a character without clownish

PR #12735, opened 2019-11-16
Improves Construct/Shade faction handling.

Cazdon:
## What Does This PR Do
In short, fixes a lot of unlisted issues regarding Constructs/Shades and how they're given factions, and gives Wizard Constructs a Wizard HUD.

The long version:

- Fixed an issue where creating constructs made an entirely second mind out of the shade, causing the shade to not have a body, but the mind would still count for cult objectives. Now the mind of the shade simply transfer into the construct.

- Fixed an issue when trying to place a soul shard into a construct shell, but before picking a type, you summoned the shade from the soul stone, the shade would still exist alongside the new construct. This issue could have been exploited to have a free 'dummy invoker.'

- Tweaked Constructs and Shades to be removed from the Cult on death, as to not count for the 'total number of cultists' objective and as an added effect, tidies up the End of Round Cult Reveal to not include dead Shades/Constructs.

- Fixed Non-Cultist Constructs and Shades not having a Note in their mind about who their Master is, despite having the objective to serve them. Cult Constructs still have a generic 'Serve the Cult' objective.

- Tweaked Cult and Wizard Constructs/Shades to log as antags with the big red (A) in the attack logs.

- Tweaked Wizard-Made Constructs to have a Wizard HUD, to identify Wizards and Apprentices, and subsequently have an Apprentice Icon over their heads as well. 

- Tweaked Soulstone code slightly to make creating a new construct just a bit cleaner.

## Why It's Good For The Game
Aside from the fixes, which I'll just assume most people can agree are good for the game, the one thing I would like to justify is the WizardHUD. I personally have had my one Wizard round ended early by a construct who couldn't tell who I was, and subsequently murdered me for it. Very unfortunate. Turns out this has happened before and since, and I feel this change is sort of necessary for Wizards to use constructs effectively as the amount of chaos in a Wizard round leaves a LOT of room for Constructs to get confused on who's who.

## Changelog
:cl:
fix: Creating a Construct from Shade no longer makes a duplicate mind, and transfers minds instead.
fix: Summoning a Shade in the middle of selecting a construct no longer leaves a brain dead shade.
tweak: Constructs/Shades are removed from the cult on death, and no longer count for objectives.
fix: Constructs/Shades made by non-cultists properly have a Master listed in their notes.
tweak: Constructs/Shades made by Cult/Wizards log as Antags in attack logs.
tweak: Wizard-Made Constructs now have a WizardHud and can identify Wizards and Apprentices.
tweak: Construct code maintenance.
/:cl:

PR #12816, opened 2019-12-18
Cult Fixes and Tweaks: Admins can cult anyone, Constructs get their comms back.

Cazdon:
## What Does This PR Do
Title.

The first part is achieved by removing a redundant is_convertible_to_cult check in the add_cultist proc, and by editing the traitor panel to not gray out 'make cultist' when the character has a mind shield.

In #12735 I changed constructs to just be mind transfers from shades. This resulted in everything transferring except the spell to speak over cult comms, which is tied to the body itself. We just check if the new construct is a cultist, and give it the comms if so.

## Why It's Good For The Game
Fixes good. Also removing that redundant check fixes really weird situations where someone would fail the is_convertible_to_cult check without a mind shield, like say a Chaplain. That would cause the player to hear the conversion sound, get text, but not actually be in the cult at all, leaving a very confused admin and player.

Also admins should be able to cult literally anyone, even through mind shields. It's adminbus.

## Changelog
:cl:
tweak: Admins can force anyone into the cult.
fix: Cult Constructs get their comms again.
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -225,9 +225,7 @@
 
 /datum/mind/proc/memory_edit_cult(mob/living/carbon/human/H)
 	. = _memory_edit_header("cult")
-	if(ismindshielded(H))
-		. += "<B>NO</B>|cultist"
-	else if(src in SSticker.mode.cult)
+	if(src in SSticker.mode.cult)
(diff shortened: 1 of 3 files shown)


Cazdon:
Oh and just in case, I tested a few round starts and they remain working just fine. No Chaplains, Captains, or HoSes becoming round start cultists. :^]

Cazdon:
Whoops

PR #12819, opened 2019-12-18
ID computer can no longer encourage Magi/BS/NTR as job transfer options

Kyep:
## What Does This PR Do
Prior to this PR, Magi/Blueshield/NTR were treated the same as other karma jobs - they could be shown as 'encouraged' (ie: green highlight) job transfer options under certain conditions (job slot is free, player has job unlocked, etc).
With this PR, these three jobs are now always shown as 'discouraged' job transfer options. IE, they can never be shown in green, they can only be shown in gray, with text that turns red if you mouse over it.

This is a purely cosmetic change. It does not stop you from assigning these jobs to an ID.
All it does is change the way these options are displayed, such that the ID computer never encourages you to pick these options anymore.

## Why It's Good For The Game
Last week, the heads made a decision to stop allowing job transfers to Magi/BS/NTR. Primarily because we did not want Captains giving these jobs to their allies, which would undermine the vital oversight these jobs are meant to provide.
Currently, the ID computer still treats them like other karma jobs, and encourages them in some cases.
With this PR, the ID computer stops encouraging things that are against SOP and which the heads have decided shouldn't happen.

:cl: Kyep
tweak: ID computer will no longer encourage job transfers to Magi/BS/NTR.
/:cl:

### DIFF
--- a/code/game/jobs/job/job.dm
+++ b/code/game/jobs/job/job.dm
@@ -53,6 +53,7 @@
 	var/exp_type = ""
 
 	var/disabilities_allowed = 1
+	var/transfer_allowed = TRUE // If false, ID computer will always discourage transfers to this job, even if player is eligible
 
 	var/admin_only = 0
 	var/spawn_ert = 0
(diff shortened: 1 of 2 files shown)


Cazdon:
So the functionality is left in for traitor HoPs and such?

Kyep:
@Cazdon 
Yes.
This PR doesn't stop you from assigning these jobs to a card. (Necessary for replacement IDs and such).
It just changes how the jobs are displayed - displaying them in a way that stops encouraging HoPs to pick them.

PR #12820, opened 2019-12-18
Gives sleepers and cryotubes the ability to auto-eject dead people

SteelSlayer:
## What Does This PR Do
Adds a button to medical sleepers and cryotubes, that when toggled on, auto-ejects dead occupants and plays a buzz noise when it's ejects them.

## Why It's Good For The Game
I think it's bad that cryotubes and sleepers have no visual indicator that the person inside is currently alive or dead. This at least gives people the option to auto-eject those dead patients so they don't spend lots of time sitting there dead while no one notices them.

You could argue this is a "git gud" issue for doctors, and that they need to pay attention more. However, I would say that when things get really hectic and you have doctors running all over, frantically trying to heal and perform surgery during a crisis, they can easily forget to check these two machines, *especially* the sleeper. I've seen players sit in these for upwards of 10-15 minutes, myself included, and that's just frustrating and annoying for the player stuck in the sleeper/cryotube.

## Images of changes
c
s

## Changelog
:cl:
add: Gives sleepers and cryotubes the ability to auto-eject dead people via a toggleable button on their interface.
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -23,6 +23,7 @@
 	var/initial_bin_rating = 1
 	var/min_health = -25
 	var/controls_inside = FALSE
+	var/autoeject_dead = FALSE
 	idle_power_usage = 1250
 	active_power_usage = 2500
 
@@ -81,8 +82,19 @@
 	go_out()
 
 /obj/machinery/sleeper/process()
-	if(filtering > 0)
-		if(beaker)
+	for(var/mob/M as mob in src) // makes sure that simple mobs don't get stuck inside a sleeper when they resist out of occupant's grasp
+		if(M == occupant)
+			continue
+		else
+			M.forceMove(loc)
+
+	if(occupant)
+		if(autoeject_dead && occupant.is_dead())
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 40)
+			go_out()
+			return
+
+		else if((filtering > 0) && beaker)
 			// To prevent runtimes from drawing blood from runtime, and to prevent getting IPC blood.
 			if(!istype(occupant) || !occupant.dna || (NO_BLOOD in occupant.dna.species.species_traits))
 				filtering = 0
@@ -94,7 +106,6 @@
 					occupant.reagents.trans_to(beaker, 3)
 					occupant.transfer_blood_to(beaker, 1)
 
-	if(occupant)
 		for(var/A in occupant.reagents.addiction_list)
 			var/datum/reagent/R = A
 
@@ -105,12 +116,6 @@
 				to_chat(occupant, "<span class='notice'>You no longer feel reliant on [R.name]!</span>")
 				occupant.reagents.addiction_list.Remove(R)
 				qdel(R)
-
-	for(var/mob/M as mob in src) // makes sure that simple mobs don't get stuck inside a sleeper when they resist out of occupant's grasp
-		if(M == occupant)
-			continue
-		else
-			M.forceMove(loc)
 
 	updateDialog()
 	return
@@ -201,6 +206,7 @@
 	data["maxchem"] = max_chem
 	data["minhealth"] = min_health
 	data["dialysis"] = filtering
+	data["autoeject_dead"] = autoeject_dead
 	if(beaker)
 		data["isBeakerLoaded"] = 1
 		if(beaker.reagents)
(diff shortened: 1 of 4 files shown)


Cazdon:
It seems like it would be neat. Might help a little bit. inb4 new doctors think the dead guy falling out of the cryotube is alive and healthy because it auto-ejected. 😛

Ty-Omaha:
I like how it's toggle able in case a traitor wants to do something with a body etc

VeteranKyl:
I was a little concerned how this might play out with antags hiding dying/dead people in the tubes or sleepers while they make their escape, but with it being a toggle option, I guess you just need to remember to turn off the auto eject.

code/game/machinery/Sleeper.dm
```diff
@@ -81,8 +82,19 @@
...
-		if(beaker)
+	for(var/mob/M as mob in src) // makes sure that simple mobs don't get stuck inside a sleeper when they resist out of occupant's grasp
+		if(M == occupant)
+			continue
+		else
+			M.forceMove(loc)
+
+	if(occupant)
+		if(autoeject_dead && occupant.is_dead())
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 40)
```
farie82, line 93: src implied

code/game/machinery/Sleeper.dm
```diff
@@ -23,6 +23,7 @@
 	var/initial_bin_rating = 1
 	var/min_health = -25
 	var/controls_inside = FALSE
+	var/autoeject_dead = FALSE
```
farie82, line 26: Minor styling. it reads a bit odd with auto and eject being put right after another. Putting a _ in between will improve readability

code/game/machinery/Sleeper.dm
```diff
@@ -81,8 +82,19 @@
...
+		else
+			M.forceMove(loc)
+
+	if(occupant)
+		if(autoeject_dead && occupant.is_dead())
+			playsound(src.loc, 'sound/machines/buzz-sigh.ogg', 40)
+			go_out()
+			return
+
+		else if((filtering > 0) && beaker)
```
farie82, line 97: no need for the brackets around `filtering > 0`

code/game/machinery/Sleeper.dm
```diff
@@ -81,8 +82,19 @@
...
 			// To prevent runtimes from drawing blood from runtime, and to prevent getting IPC blood.
 			if(!istype(occupant) || !occupant.dna || (NO_BLOOD in occupant.dna.species.species_traits))
```
farie82, line 99: You've lost this feature now. If an occupant leaves the sleeper that filtering will stop on it's own

code/game/machinery/cryo.dm
```diff
@@ -430,6 +438,14 @@
 	for(var/atom/movable/A in contents - component_parts - list(beaker))
 		A.forceMove(get_step(loc, SOUTH))
 
+/obj/machinery/atmospherics/unary/cryo_cell/proc/auto_eject(eject_healthy, eject_dead)
+	on = FALSE
+	go_out()
+	if(eject_healthy)
```
farie82, line 444: Maybe use a flag here instead of two exclusive bools. If both are true only eject_healthy will get through

code/game/machinery/Sleeper.dm
SteelSlayer: I'm not sure what you mean by this.

farie82: In current code the filtering option is turned off when an occupant leaves the sleeper. In your PR this wont happen since this code is only executed when there is an occupant

SteelSlayer: Ah I understand now, however we have this in the go_out proc:
```js
/obj/machinery/sleeper/proc/go_out()
	if(filtering)
		toggle_filter()
```
This ensures filtering is toggled when they exit.

farie82: yeah okay that should cover it all then. Thanks for the clarification!

code/game/machinery/cryo.dm
```diff
@@ -10,11 +13,11 @@
 	interact_offline = 1
 	max_integrity = 350
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 100, "bomb" = 0, "bio" = 100, "rad" = 100, "fire" = 30, "acid" = 30)
-	var/on = 0
+	var/on = FALSE
 	var/temperature_archived
 	var/mob/living/carbon/occupant = null
 	var/obj/item/reagent_containers/glass/beaker = null
-	var/autoeject = 0
+	var/auto_eject_prefs = 0
```
farie82, line 20: small thing. You can use NONE here to make it denote it holds something and is not a number value.

SteelSlayer:
Tested out the cling stuff a bit. Since their stasis makes them unconscious instead of dead, they won't get auto-ejected with the auto eject dead setting on while they are "dead" from stasis, which is technically a cling test. 

However, currently even without my PR changes, if the cling just goes into stasis at full health, and is inside a cryo pod or sleeper, it will show them at full HP and they will be "unconscious". So *technically* these machines are already cling tests I guess.

PR #12827, opened 2019-12-21
Removes the Freeze Mech verb, and bundles its functionality into the Freeze verb

SteelSlayer:
## What Does This PR Do
Edit: This removes the "Freeze Mech" verb and allows you to call the normal "Freeze" verb on mechs. I also removed all `admin_unFreeze` procs in favor of just having one proc handle both freezing and unfreezing for mob/mechs, that proc being `admin_Freeze`.

 I did this to cut down on a bit of copy and paste in each of the two procs. And so you never have to do things like this outside of the freeze proc:
```js
if(M in frozen_atom_list)
        M.admin_unFreeze(src)
else
        M.admin_Freeze(src)
```
This check will happen inside `admin_Freeze` itself, and you can instead just write:
```js
M.admin_Freeze(src)
```
___
Freezing a mech currently doesn't do anything. This PR makes it so the mech cannot move at all, and adds an admin freeze overlay to it. In addiation, it calls freeze on the pilot himself, so if somehow the pilot falls out of the mech (or it gets destroyed), the pilot himself will still be frozen. This is also done to prevent the pilot from firing the weapons of the mech.

If the pilot and mech get seperated somehow, but are still each frozen, you can can simply call the Freeze verb on them individually to unfreeze them.

When someone tries to enter a frozen mech, they get the message "Do not enter Admin-Frozen mechs." It's similar to the message that gets displayed when you try to touch frozen humans.

## Changelog
:cl:
tweak: Removes the Freeze Mech verb, and bundles its functionality into the Freeze verb. Freeze is now callable on both mobs and mechs.
tweak: When freezing a mech, the mech's pilot becomes frozen in addition to freezing the mech
fix: Freezing a mech will now actually freeze the mech instead of doing nothing
/:cl:


code/modules/admin/verbs/freeze.dm
```diff
@@ -6,7 +6,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
 var/global/list/frozen_mob_list = list()
-/client/proc/freeze(var/mob/living/M as mob in GLOB.mob_list)
+/client/proc/freeze(mob/living/M in GLOB.mob_list)
```
farie82: Perhaps possible to rewrite this proc to be valid for all atoms?
Seeing the freezemecha proc is on every obj ingame it won't change much I assume.

In the proc here you'd then check the type. mob? Do the usual. Mech? Freeze it's occupant

code/game/mecha/mecha.dm
```diff
@@ -263,7 +264,7 @@
 		return 1
 
 /obj/mecha/relaymove(mob/user, direction)
-	if(!direction)
+	if(!direction || frozen)
```
farie82: perhaps better to check if the occupant is frozen? 
if(occupant in frozen_mob_list)
    he_froze_yo

code/modules/admin/verbs/freeze.dm
```diff
@@ -87,7 +87,7 @@ var/global/list/frozen_mob_list = list()
 
 //////////////////////////Freeze Mech
 
-/client/proc/freezemecha(var/obj/mecha/O as obj in GLOB.mechas_list)
+/client/proc/freezemecha(obj/mecha/O in GLOB.mechas_list)
```
farie82: maybe make this /obj/mecha/M/admin_Freeze and have a counter part of it. Then declare a proc for these on atom level  which will be called by the freeze verb admins have

code/modules/admin/verbs/freeze.dm
SteelSlayer: Not a bad idea. Would it be like `atom/movable/M` or something?

farie82: Yeah something the like. Most detailed path a player can use/be

code/game/mecha/mecha.dm
SteelSlayer: Well see I also want mechs themselves able to be frozen, as I think that's the intent of the freeze mech verb

farie82: Boom you got it :P 
But I can see the need for freezing mechs. For instance if the AI is using it

code/modules/admin/verbs/freeze.dm
```diff
@@ -5,121 +5,95 @@
 //////Allows admin's to right click on any mob/mech and freeze them in place.///
 ////////////////////////////////////////////////////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
-var/global/list/frozen_mob_list = list()
-/client/proc/freeze(var/mob/living/M as mob in GLOB.mob_list)
+var/global/list/frozen_atom_list = list()
```
AffectedArc07: Can you make this into a managed global list? (Use the GLOBAL_LIST_INIT helpers)

SteelSlayer: Good idea

code/modules/admin/verbs/freeze.dm
```diff
@@ -5,121 +5,95 @@
...
 
 	if(!check_rights(R_ADMIN))
 		return
 
-	if(!istype(M))
-		return
-
-	if(M in frozen_mob_list)
-		M.admin_unFreeze(src)
+	if(istype(M, /mob/living))
```
farie82: perhaps instead of making it like this (harder to maintain) you can define admin_Freeze on movable level and make it return FALSE by default.
Then have living and mecha override it to actually do something and make it return TRUE (pods might also be a logical next step)
If it returns false you can display the message saying that it's not suited for this thing.

SteelSlayer: Holy shit, I understand what you're saying now, and it works.

image

PR #12844, opened 2019-12-24
Revert "Move the handle_changeling proc call to the (new) life signal"

farie82:
Reverts ParadiseSS13/Paradise#12798

Clings got broken when they change into another body. Don't have the time myself to fix this so I'll put the revert up instead.

Issue is caused because the signal listens on the body where the cling spawns with. This is not changed on a mind tranfer or such. But should be done (signals maybe for that)


farie82:
Workaround till this is merged or the issue is fixed is calling make_changeling without parameters on the affected mob.

PR #12851, opened 2019-12-26
Activating internals automatically pushes breath masks into place

SteelSlayer:
## What Does This PR Do
When you press the "set internals" action button and have a breathe mask that is pushed down, it will automatically push the mask up and activate the internals.

## Why It's Good For The Game
Small QoL thing, and the reverse of this already exists. If you have internals activated and press the "push down breath mask" button, It will auto-deactivate your internals.

## Images of changes
GIF

## Changelog
:cl:
tweak: Activating internals via the action button automatically pushes breath masks into place, if the mask is pushed down.
/:cl:


Fox-McCloud:
Does this really need to be more brainless than it already is? What's the point of mushing a mask up and down at all then, other than aesthetics?

I know this is a small thing, and it's certainly not a hill I'm willing to die on, but this seems like it goes over the line of "QoL" into "people do dumb"-handholding.

SteelSlayer:
I might not be making this PR if the reverse of this didn't exist, where you can push down your mask and your internals deactivate automatically. So given that, I think it makes sense to have, but, up to you of course.

variableundefined (approved):
Minor QOL, I don't see any issue with this.

PR #12858, opened 2019-12-28
Make can_inject deterministic; fix odysseus syringes sometimes not working.

### LINKED ISSUE #12857: Odysseus syringe gun does not work properly
**Issue Description**:
the odysseus syringe gun does not work properly 

for the syringe to hit the target and transfer its reagents to it, saying in admin log that it transferred the reagent

**What happened instead**:
the syringe hits the target but no reagents get transferred, also no admin log

it makes the syringe's use impractical.

the situation where the launcher did not inject the syringe was the one in the image(notice for the syringe beneath the vulpin 
image ).

no idea.

**Extra information**:
it seems that the failure is random.

moxian:
## What Does This PR Do
Fixes #12857 by passing in targeting_zone that the mecha occupant aims at.
The problem here is that when neither targeting zone nor the user are passed to `can_inject()` , it picks a random zone (which is weird, since one would expect `can_inject()` to be deterministic), and many of those random choices are not really zones (`"left leg"` is not a zone, but `"l_leg"` is).
Anyhow, now all callers to `can_inject` pass in either non-null user or properly pass in targeting zone (or both).

Also added test for #12857  . It probably doesn't add much value at this point, but I think this time is as good as any to start building up a test suite.

Separately, my previous pr of #12697 was overly cautions with not closing the bugs it fixed, so I'll do it here:
Fixes: #12685
Fixes: #7132

## Why It's Good For The Game
Bugfix good
Automated testing good

## Images of the changes
B8g9ZeWIaH

## Changelog
:cl:
fix: Odysseus syringes now inject the victim consistently
/:cl:

### DIFF
--- a/code/game/mecha/equipment/tools/medical_tools.dm
+++ b/code/game/mecha/equipment/tools/medical_tools.dm
@@ -314,6 +314,7 @@
 	playsound(chassis, 'sound/items/syringeproj.ogg', 50, 1)
 	log_message("Launched [mechsyringe] from [src], targeting [target].")
 	var/mob/originaloccupant = chassis.occupant
+	var/original_target_zone = originaloccupant.zone_selected
 	spawn(0)
 		src = null //if src is deleted, still process the syringe
 		var/max_range = 6
@@ -330,7 +331,7 @@
 			if(M)
 				var/R
 				mechsyringe.visible_message("<span class=\"attack\"> [M] was hit by the syringe!</span>")
-				if(M.can_inject(null, TRUE))
+				if(M.can_inject(originaloccupant, TRUE, original_target_zone))
 					if(mechsyringe.reagents)
 						for(var/datum/reagent/A in mechsyringe.reagents.reagent_list)
 							R += A.id + " ("
--- a/code/game/mecha/equipment/tools/medical_tools.dm
+++ b/code/game/mecha/equipment/tools/medical_tools.dm
@@ -565,3 +565,46 @@
 		if(M.equipment.len < M.max_equip)
 			return TRUE
 	return FALSE
+
+
+/datum/tests/mecha_syringe_gun/gh12857/proc/run_test(mob/admin, cleanup="on_pass")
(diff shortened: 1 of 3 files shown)


farie82:
Interesting idea with the testing proc. Unsure how much use it will see.

code/game/mecha/equipment/tools/medical_tools.dm
```diff
@@ -330,7 +331,7 @@
 			if(M)
 				var/R
 				mechsyringe.visible_message("<span class=\"attack\"> [M] was hit by the syringe!</span>")
-				if(M.can_inject(null, TRUE))
+				if(M.can_inject(originaloccupant, TRUE, original_target_zone))
```
moxian: Btw, this also makes the occupant see the error message if the body part they are targeting is mechanic/missing/otherwise uninjectable. I'm assuming this was the original intent given the errors=TRUE argument.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1007,34 +1007,34 @@
...
 
 /mob/living/carbon/human/can_inject(mob/user, error_msg, target_zone, penetrate_thick = FALSE)
-	. = 1
+	. = TRUE
 
 	if(!target_zone)
 		if(!user)
-			target_zone = pick("chest","chest","chest","left leg","right leg","left arm", "right arm", "head")
+			. = FALSE
+			CRASH("can_inject() called on a human mob with neither a user nor a targeting zone selected.")
```
farie82, line 1015: Looks like you got all the obvious usages of this proc that have both user and target_zone set to null. Be sure that other less obvious paths are also covered. Ones that get their target zone or user as a parameter

Kyep:
@moxian I am willing to vote for this if you:
- confirm it won't break things like terror spider bite attack injections
and
- resolve the merge conflicts

Kyep:
@moxian

Fox-McCloud:
No response after a week. Closing.

PR #12860, opened 2019-12-29
Kudzu Mutativeness Fix

Arvenius169:
## What Does This PR Do
Changes the lower bound of kudzu mutativeness from infinity at 0 potency(math-breaking) and having 0 mutativeness at 1 potency to having 0 mutativeness at 0 potency, as mutativeness is a squared log of potency.

## Why It's Good For The Game
Kudzu with potency below 1 will now not have busted mutativeness. Kudzu at 0 potency will not mutate.

## Changelog
:cl:
fix: Changed some bad math to make kudzu at 0 potency to have infinite mutativeness.
/:cl:

### DIFF
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -544,7 +544,7 @@
 	if(potency != null && potency > 0)
 		// ~1 mutativeness at 6 potency
 		// 4 mutativeness at 100 potency
-		mutativeness = sqrt(potency) * .4
+		mutativeness = sqrt(potency) * 0.4
 	if(production != null)
 		// 1 production is crazy powerful
 		var/spread_value = max(10 - production, 1)
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -544,7 +544,7 @@
 	if(potency != null && potency > 0)
 		// 1 mutativeness at 10 potency
 		// 4 mutativeness at 100 potency
-		mutativeness = log(10, potency) ** 2
+		mutativeness = log(10, (potency - 1)) ** 2
 	if(production != null)
 		// 1 production is crazy powerful
 		var/spread_value = max(10 - production, 1)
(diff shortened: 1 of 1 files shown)


code/modules/events/spacevine.dm
```diff
@@ -544,7 +544,7 @@
 	if(potency != null && potency > 0)
 		// 1 mutativeness at 10 potency
 		// 4 mutativeness at 100 potency
```
moxian: This comment is now technically wrong.

moxian:
Honestly, squared logarithm is such a weird function, I'm surprised it's used anywhere outside of hardcore math. I'd suggest changing the `log_10(x) ** 2` to something simpler like `sqrt(x)/2.5` which looks almost exactly the same, but behaves in a much more straightforward way.
firefox-20191229-031450

Arvenius169:
> 
> 
> Honestly, squared logarithm is such a weird function, I'm surprised it's used anywhere outside of hardcore math. I'd suggest changing the `log_10(x) ** 2` to something simpler like `sqrt(x)/2.5` which looks almost exactly the same, but behaves in a much more straightforward way.
> firefox-20191229-031450

Good point, reminding me to fix the comments. But also, I hadn't even noticed that the full equation was log10(p)^2. That's genuinely bizarre. The equation you suggested is much better, and I'll add documentation that reflects it.

Kyep:
Does this fix the fact that space vines always seem to spawn a ton of mutated variants, including explosive, when they get out of control?

Arvenius169:
> Does this fix the fact that space vines always seem to spawn a ton of mutated variants, including explosive, when they get out of control?

I think that that might just be a function of vines having more opportunities to mutate the more sizeable they are. Whether or not that's the case, this new equation makes kudzu slightly more mutative than before  at potencies below ~25, with the increase being larger the closer you get to 0, and has very slightly lower mutativeness than before while beyond that potency threshold. The curves are really pretty close, but the main difference is that the new one doesn't suddenly rocket to infinity as you pass under 1 potency.

Fox-McCloud:
Deconflict. OTherwise, this is quite stale.

Fox-McCloud:
Failure to respond. Still has conflicts. Stale.

PR #12863, opened 2019-12-30
Adds an alt-click shortcut for opening storage items

SteelSlayer:
## What Does This PR Do
We currently have two ways of opening storage items such as backpacks. Either you click it with an open hand while holding it in your other hand, or you click and drag the item onto your character. 

This adds a third way to open bag inventories, by performing an alt-click on the storage object.

This works for any  `/obj/item/storage` type object or child of that type. Which includes, but is not limited to, all types of backpacks, medkits, boxes, patch packs, pill bottles, chemistry bags and more.

## Why It's Good For The Game
The main reason for making this was to cut down the tediousness of having to drag and drop things onto yourself when searching through bags. For example, if there are a lot of boxes or medkits you need to search through in a backpack. And even when you're not searching through a bag, I find alt-clicking is much easier than dragging and dropping.

## Changelog
:cl:
add: Adds an alt-click shortcut for opening storage items
/:cl:


moxian:
1) why do we need three different ways to do the same thing
2) this is non-discoverable
3) this is very ad-hoc
Can we sit down and rethink when should each type of click (click/drag-n-drop/alt-click/ctrl-click/middle-click) do in each situation implement that, and cut the rest? I mean, I know this is hard and ambitious, and people would be upset once the way they are used to stops working, but the UI inconsistency is already a problem in ss13, which imho has to be addressed, and this PR moves the opposite direction. (Btw did you know that drag-n-dropping pill bottle onto yourself opens its contents, but drag-n-dropping patch pack starts applying all the patches to you? :stars: )

SteelSlayer:
@moxian 

1. *Technically* we don't need the drag-and-drop method for opening bags if this becomes a thing. As for the first method (clicking on the bag with an open hand), that's just the default way to open things and this will probably never be removed. But also, I did not want to remove the drag-and-drop method because many people are use to this and may prefer it.
2. Well yes, but a lot of things in SS13 are non-discoverable including dragging and dropping bags onto yourself to open them, or similarly opening other people's equipment screen by drag-n-drop. It is simply the nature of SS13, unless we have some sort of tutorial/cheat sheet for these types of things.
3. You may think its unnecessary and that's fine, but as I wrote in the PR description, it cuts down on some tediousness of having to drag-and-drop multiple times in a row. This is akin to adding the control-click shortcut to atmos devices like pumps to switch them on and off.

As for the rest of what you said, that is topic for some kind of "click behavior refactor" where you make clear what type of clicks do what, and change them according. I am not aiming to address that at this time. 

I did not know about that patch pack thing and I think it's sort of silly.

moxian:
I know my opinion has pretty much no weight in this, but I'd be fine with adding alt-click-to-open if - and only if - drag-and-drop-to-open gets removed. I'm not fine if both are to stay.

Mitchs98:
I see no reason both can't remain for people that prefer to do one or the other. It's not like it's adding some big thing that'll need to be carefully maintained.

farie82:
All for this but it will make alt clicking on tiles harder. Not much you can do about that I guess though

Fox-McCloud:
Something tells me this is going to break things, especially on down the line--mostly being able to access the "storage" of things you're really not supposed to be able to open.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -82,6 +82,12 @@
 			return
 	return
 
+/obj/item/storage/AltClick(mob/user)
+	if(user.Adjacent(get_turf(src)))
```
Fox-McCloud: Why not `if(Adjacent(user))`?

SteelSlayer: Eh yeah I'm not sure why I did this

RV666:
This implementation lacks incapacitance check and lock status check for secure briefcase and wall safe.
Why not use just 
`/obj/item/storage/AltClick(mob/user) 
	MouseDrop(user) `

SteelSlayer:
Yeah you're right, I'm going to add lock checks and incapacitated checks in another PR. Thanks for pointing that out.

RaidarV:
Thank you, you are a good man.

PR #12865, opened 2019-12-30
Racial Loadout Items Pathway Fix

Arvenius169:
## What Does This PR Do
Fixes a small, inconsequential error from a previous PR I made, which showed up in dream daemon but otherwise had no effect.

## Why It's Good For The Game
Doesn't make an unused parent object anymore.

## Changelog
:cl:
fix: Removes an ugly error that gets thrown up on server initialization.
/:cl:


Kyep:
I'll be so glad when I don't have to go "what is this?!?" every time I start a server.

PR #12870, opened 2020-01-01
Makes radio connections update properly when you change their host machine's frequency

SteelSlayer:
## What Does This PR Do
What happens currently is that this line in ` code/game/machinery/machinery.dm ` was executing and setting only that machine's frequency var:
```js
src:frequency = sanitize_frequency(newfreq, RADIO_LOW_FREQ, RADIO_HIGH_FREQ)
```
Instead I made it call the `set_frequency` proc:
```js
set_frequency(sanitize_frequency(newfreq, RADIO_LOW_FREQ, RADIO_HIGH_FREQ))
```
This proc is used on most machines (mainly atmos devices), which communicate with other machines using radio signals. So while this was setting that machine's `frequency` var, it does not update the machine's `radio_connection` and ITS frequency var. Calling `set_frequency` will update this properly for each respective device.

Also make's air injectors able to be properly linked with atmospherics tank control consoles.

Little demo video showing the linking of an air injector to a tank control console: 
https://streamable.com/3a154u

## Why It's Good For The Game
Fixes things not working as they should.

## Changelog
:cl:
fix: Radio connections update properly when you change their host machine's frequency
fix: Air injectors can connect to atmospherics tank control consoles properly
/:cl:


code/game/machinery/atmo_control.dm
```diff
@@ -407,6 +407,7 @@
 
 /obj/machinery/computer/general_air_control/large_tank_control/linkWith(mob/user, obj/O, list/context)
 	if(context["slot"]=="input" && is_type_in_list(O,input_linkable))
+		input_tag = O:id_tag
```
Fox-McCloud: Do you want to get stabbed?

SteelSlayer: I know... this crap is all over these files, and I really didn't want to mess with it. But if you want I can refactor this stuff to not use `:`

SteelSlayer:
There is so much shitcode with this radio signal crap. I plan on doing a more in-depth refactor of this.

code/ATMOSPHERICS/components/binary_devices/pump.dm
```diff
@@ -121,7 +121,7 @@ Thus, the two variables affect pump operation are set in New():
 	return 1
 
 //Radio remote control
-/obj/machinery/atmospherics/binary/pump/proc/set_frequency(new_frequency)
+/obj/machinery/atmospherics/binary/pump/set_frequency(new_frequency)
 	SSradio.remove_object(src, frequency)
 	frequency = new_frequency
 	if(frequency)
```
AffectedArc07, line 127: If the child proc is the same on all of these, why not just define it at parent level?

SteelSlayer: That is a very good point, one that 9 month ago me didn't realize. I will fix it.

SteelSlayer: Addressed in the latest commit.

SteelSlayer:
Updated to remove more copy and paste / re-declaration of the same variables across a bunch of machines.
- The `datum/radio_frequency/radio_connection` var that a bunch of machines had, but kept re-declaring all over the place is now defined on the base `/obj/machinery`.
- Similarly, `var/frequency` was being declared everywhere, but it is now only declared on the base `/obj/machinery`.
- Moved this proc: 
```dm
/obj/machinery/atmospherics/set_frequency(new_frequency)
	SSradio.remove_object(src, frequency)
	frequency = new_frequency
	if(frequency)
		radio_connection = SSradio.add_object(src, frequency, RADIO_ATMOSIA)
```
To the `/obj/machinery/atmospherics` level, and deleted the copy and paste of this on several subtype atmospherics machines.

AffectedArc07:
Upon testing, this happens
image

SteelSlayer:
Ok I fucked up by removing air sensor's and the general air computer's `set_frequency()`, not realizing at the time it wouldn't get covered by my `/obj/machinery/atmospherics/set_frequency()` proc. Because duh, these two are not `/atmospherics` subtypes. It's fixed now though.

AffectedArc07 (approved):
Tested locally, seems to work

PR #12874, opened 2020-01-02
Add missing grey translator implant check

datlo:
Add a missing translator implant check in say code. 

This missing check caused radios to stop working for greys when suffering oxygen damage, as the radio checked IsVocal without going through the usual say code. 

Due to this, while the grey is able to speak, they were unable to speak into the radio.

With this check in, the radio works properly.

:cl:
fix: Fixed greys with translator implant being unable to use their headset when suffering oxygen damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/say.dm
+++ b/code/modules/mob/living/carbon/human/say.dm
@@ -97,6 +97,9 @@
 	return real_name
 
 /mob/living/carbon/human/IsVocal()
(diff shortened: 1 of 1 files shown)


LegendWookie:
i love you thanks

Crazylemon64:
Wait, why would the gray be able to speak normally if `IsVocal` would return false due to low oxy?

datlo:
The grey translator implant bypasses all speech inpendiments without exception by design, as it translates thoughts into speech directly. I just missed implementing it into this proc when I coded the translator PR, as it wasnt relevant back then (you could speak with low oxy)

PR #4777, opened 2016-06-25
Renames Loyalty implants to Mindshield implants

Kyep:
Exactly what it says on the tin.

Note: this change only affects player-visible references to loyalty implants. Not changing their names in the code to avoid map conflicts.

Reason: lots of ahelps like "does a loyalty implant make me loyal?"
No, it doesn't. It just prevents most forms of brainwashing.

:cl: Kyep
tweak: "Loyalty" implants are now known as "Mindshield" implants
/:cl:

PR #12876, opened 2020-01-03
Removes Loyalty Implant aspect of Mindshield Implants

ppi13:
## What Does This PR Do
Makes it so that Mindshields provides protection against brainwashing, and does not work like a loyalty implant or mind slave implant
Adds Loyalty Implants back into the game

## Why It's Good For The Game
Works as the name says and not as remove brainwashing implant, with loyalty implants taking over as the brainwashing part.

## Changelog
:cl:
del: Remove Mindshield implants reverting cultists, revs or shadowling thralls
tweak: Changed the Implant data for mindshields to only protect against brainwashing.
add: Adds back Loyalty Implants, Nanotrasens answer to the Syndicates Mindslave implants.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/implants/implant_mindshield.dm
+++ b/code/game/objects/items/weapons/implants/implant_mindshield.dm
@@ -12,23 +12,14 @@
 				<HR>
 				<b>Implant Details:</b><BR>
 				<b>Function:</b> Contains a small pod of nanobots that manipulate the host's mental functions.<BR>
-				<b>Special Features:</b> Will prevent and cure most forms of brainwashing.<BR>
+				<b>Special Features:</b> Will prevent most forms of brainwashing.<BR>
 				<b>Integrity:</b> Implant will last so long as the nanobots are inside the bloodstream."}
 	return dat
 
 
 /obj/item/implant/mindshield/implant(mob/target)
 	if(..())
-		if(target.mind in SSticker.mode.head_revolutionaries || is_shadow_or_thrall(target))
-			target.visible_message("<span class='warning'>[target] seems to resist the implant!</span>", "<span class='warning'>You feel the corporate tendrils of Nanotrasen try to invade your mind!</span>")
-			removed(target, 1)
-			qdel(src)
-			return -1
-		if(target.mind in SSticker.mode.revolutionaries)
-			SSticker.mode.remove_revolutionary(target.mind)
-		if(target.mind in SSticker.mode.cult)
-			to_chat(target, "<span class='warning'>You feel the corporate tendrils of Nanotrasen try to invade your mind!</span>")
-		else
+		if(target.stat != DEAD)
 			to_chat(target, "<span class='notice'>Your mind feels hardened - more resistant to brainwashing.</span>")
 		return 1
 	return 0
@@ -36,7 +27,7 @@
 /obj/item/implant/mindshield/removed(mob/target, var/silent = 0)
 	if(..())
 		if(target.stat != DEAD && !silent)
-			to_chat(target, "<span class='boldnotice'>You feel a sense of liberation as Nanotrasen's grip on your mind fades away.</span>")
+			to_chat(target, "<span class='boldnotice'>Your mind feels vulnerable - more open to outside influences.</span>")
 		return 1
 	return 0
 
--- a/code/game/objects/items/weapons/implants/implant_loyalty.dm
+++ b/code/game/objects/items/weapons/implants/implant_loyalty.dm
@@ -1,0 +1,59 @@
+/obj/item/implant/loyalty
+	name = "loyalty implant"
+	desc = "Makes you loyal to Nanotrasen."
+	origin_tech = "materials=5;biotech=5;programming=7"
+	activated = 0
+
+/obj/item/implant/loyalty/get_data()
+	var/dat = {"<b>Implant Specifications:</b><BR>
+				<b>Name:</b> Nanotrasen Employee Management Implant<BR>
+				<b>Life:</b> Ten years.<BR>
+				<b>Important Notes:</b> Personnel injected with this device tend to be much more loyal to the company.<BR>
+				<HR>
+				<b>Implant Details:</b><BR>
+				<b>Function:</b> Contains a small pod of nanobots that manipulate the host's mental functions.<BR>
+				<b>Special Features:</b> Will prevent and cure most forms of brainwashing.<BR>
+				<b>Integrity:</b> Implant will last so long as the nanobots are inside the bloodstream."}
+	return dat
+
+
+/obj/item/implant/loyalty/implant(mob/target)
+	if(..())
+		if(target.mind in SSticker.mode.head_revolutionaries || is_shadow_or_thrall(target) || ismindshielded(target) || ismindslave(target))
+			target.visible_message("<span class='warning'>[target] seems to resist the implant!</span>", "<span class='warning'>You feel the corporate tendrils of Nanotrasen try to invade your mind!</span>")
+			removed(target, 1)
+			qdel(src)
+			return -1
+		if(target.mind in SSticker.mode.revolutionaries)
+			SSticker.mode.remove_revolutionary(target.mind)
+		if(target.mind in SSticker.mode.cult)
+			to_chat(target, "<span class='warning'>You feel the corporate tendrils of Nanotrasen try to invade your mind!</span>")
+		else
+			to_chat(target, "<span class='notice'>You feel a surge of loyalty towards Nanotrasen.</span>")
+		return 1
+	return 0
+
+/obj/item/implant/loyalty/removed(mob/target, var/silent = 0)
+	if(..())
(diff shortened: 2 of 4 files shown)


TDSSS:
So this makes revs completely unrevertable? Security's only option for them is executing them? Because to my knowledge this is the only way to revert those. 

And if I am reading this code right, then the implant doesn't even...remove cultists or shadowling thralls. Like, two thirds of what you say this does is already done, you just removed a flavour message.

taukausanake:
> only option for them is executing them

To be honest, that's pretty hardcore and I like the sound of it. Not so much fun for the revs but then they either play hard or die soft.

You think how you have holy water and surgery for curing cults and thralls that revs might have a reprogramming way of being cured? How about electroshock therapy? That might be fun

FreeStylaLT:
Revolution isn't in the rotation, I'm not sure it should really factor into whether or not this should get merged.

Carthusia:
I always thought a MS was just that, to prevent brainwashing, it was never a "loyalty implant" where your 100% loyal to NT, if that was the case, security being corrupt, taking bribes and what not wouldn't be possible

taukausanake:
@Carthusia #4777 is where things changed. Man that was a long time ago

Quantum-M:
I thought mindshields only prevented brainwashing and not reversed it with exception of non-head revolutionaries?

Spacemanspark:
While Rev isn't in the code, I see no reason why we need to make the game mode even more ridiculous than it already is.
To add to that, as far as I'm aware, the implants already don't deconvert cultists/ shadowlings. What's the point in this PR?

datlo:
This PR pretty much only affects rev gameplay wise and while rev definitely needs something changed to be a more interesting gamemode I dont think this is it.

VeteranKyl:
There's actually a way to cure revs outside of mindshields. I believe a hit to the head is able to fix it. So replace security mindshielding folks with sec harm batoning people.

Cazdon:
Mindshields already don't deconvert cultists or sling thralls, only prevent conversion.

Mitchs98:
@TDSSS Actually beating the shit out of a rev over the head with a baton can/will deconvert them.

But uh, yeah, I don't see any point in this PR. At all.

ppi13:
This PR does multiple things, first up, you can't detect a thrall or pre-transformation shadowling by implanting them, it removes the mind slave like message you get if you are a cultist.  
As for revs, beating them over the head may de-convert them, but here is the deal, when dealing with revolution people Are gonna die, innocent people, revs, security, heads.  

And if we look at Space Law, it says Mutiny is a Capital Crime, when security is being out numbered 8 to 1 then lethal force is the only option.

Spacemanspark:
_willful_ mutiny is a crime. Headrevs? Yes, they fall under this catagory. 
Revs are brainwashed, and not acting of their own willpower. That's why they're meant to be deconverted first.
When things go to shit, yes, martial law comes into play, but until then they're meant to be saved, not killed.

Mitchs98:
What Spark said. I'm entirely unconvinced that there's a point or purpose to this PR. And you absolutely do not need a mindshield to detect a thrall, and I'm 80% sure you can't detect a pre-hatched shadowling with one(why would you even do that, when would that even be an issue even if you could?).

Now, the loyalty implant aspects I agree with removing. As they aren't loyalty implants in any way. Everything else? Absolutely not.

code/game/objects/items/weapons/implants/implant_mindshield.dm
```diff
@@ -12,31 +12,22 @@
 				<HR>
 				<b>Implant Details:</b><BR>
 				<b>Function:</b> Contains a small pod of nanobots that manipulate the host's mental functions.<BR>
-				<b>Special Features:</b> Will prevent and cure most forms of brainwashing.<BR>
+				<b>Special Features:</b> Will prevent most forms of brainwashing.<BR>
```
Kyep: Reasonable change.

code/game/objects/items/weapons/implants/implant_mindshield.dm
```diff
@@ -12,31 +12,22 @@
...
 				<b>Integrity:</b> Implant will last so long as the nanobots are inside the bloodstream."}
 	return dat
 
 
 /obj/item/implant/mindshield/implant(mob/target)
 	if(..())
-		if(target.mind in SSticker.mode.head_revolutionaries || is_shadow_or_thrall(target))
```
Kyep: Meh on this. Shadowlings are easy to identify (they're dying because of the flashlight in your pocket). Thralls are also easy to identify (remove their mask, examine them). So... I'm not sure this matters much either way.

code/game/objects/items/weapons/implants/implant_mindshield.dm
```diff
@@ -12,31 +12,22 @@
...
-			target.visible_message("<span class='warning'>[target] seems to resist the implant!</span>", "<span class='warning'>You feel the corporate tendrils of Nanotrasen try to invade your mind!</span>")
-			removed(target, 1)
-			qdel(src)
-			return -1
-		if(target.mind in SSticker.mode.revolutionaries)
```
Kyep: This is a big deal. Rev is rarely played, but mindshields are the only reliable way of countering revs. Removing this essentially makes rev unwinnable for the crew. Well.... even more unwinnable than it already is.

code/game/objects/items/weapons/implants/implant_mindshield.dm
```diff
@@ -12,31 +12,22 @@
...
-			SSticker.mode.remove_revolutionary(target.mind)
-		if(target.mind in SSticker.mode.cult)
```
Kyep: This is reasonable.

code/game/objects/items/weapons/implants/implant_mindshield.dm
```diff
@@ -12,31 +12,22 @@
...
+		if(target.stat != DEAD)
 			to_chat(target, "<span class='notice'>Your mind feels hardened - more resistant to brainwashing.</span>")
 		return 1
 	return 0
 
 /obj/item/implant/mindshield/removed(mob/target, var/silent = 0)
 	if(..())
 		if(target.stat != DEAD && !silent)
-			to_chat(target, "<span class='boldnotice'>You feel a sense of liberation as Nanotrasen's grip on your mind fades away.</span>")
+			to_chat(target, "<span class='boldnotice'>Your mind feels vulnerable - more open to outside influences.</span>")
```
Kyep: Yeah, this is reasonable.

code/game/objects/items/weapons/implants/implant_mindshield.dm
ppi13: I added back Loyalty Implants that can be part of ERT kit, or spawned for the ert since rev is admin only , can also replace the mindshield implants of centcom personnel for rp reasons.

Mitchs98:
I'd rather not add loyalty implants back into the game in any kind of form, especially from ERT admin spawned or not. Again, see no reason to remove the function of removing rev status from them.

Spacemanspark:
We specifically removed loyalty implants from the game years ago due to their negative impact on the server atmosphere.

I see no need to add them back.

variableundefined:
Objection for reasons above.

Kyep:
When I made my comments about revs, I was suggesting that you remove those changes.
Not that you go further and proceed to re-implement loyalty implants overall. They were removed for a reason.
Objection.

Vivalas:
I'm curious, since this actually popped up in a round last night, what's the whole thing about NT "owning everyone" about? That seems to conflict with removal of loyalty implants and afaik only exists lorewise because of the addition of devil. It would seem to have the same negative impact on the atmosphere as loyalty implants would.

Carthusia:
There's another conflict that feels like its leftover from old loyalty implant lore, the fact a mind shielded person's testimony is admissible as evidence, without actual evidence to back it.

This makes sense if they have Loyalty implants which would prevent them from lieing, but as they dont do this, the fact they have a shield implant in no way prevents them from lieing, being bribed, or being corrupt, thus the logic behind allowing MS testimony to be accepted in space law without any evidence backing the claim seems...odd considering how much evidence matters.

Kyep:
Practically, sec officers have to be able to arrest someone for something and have "I saw them do it" treated as sufficient evidence. Otherwise sec would be unplayable.

Historically that was tied to mindshield implants - but perhaps instead it should be tied to someone being part of Security. IE: a civilian getting a mindshield implant doesn't make them trustworthy, nor does a sec officer losing their implant due to cloning make them untrustworthy. Maybe start a forum thread for that discussion - it doesn't really belong here, as its a policy/spacelaw change, not a code change.

Carthusia:
Right-o

Fox-McCloud:
NT doesn't need mindslave implants of their own.

We've seen what happens when these are utilized; it's terrible, because players are terrible.

Also making this more annoyingly difficult to understand what is what isn't worth the additional "nuance".

👎

variableundefined:
Closing due to majority head and maintainer objections.

PR #12882, opened 2020-01-05
Adds Calcium Zeolite and Tranexamic Acid

Aquilar:
## What Does This PR Do

This adds a pair of bleed rate affecting chemicals and the code additions required to make it work, specifically handle_blood in blood.dm was added too as these two chemicals have no effects of their own other than overdose.

update: originally this relied on blood.dm using a proc to figure out what level of suppression it should use, now I moved it out of there because i didn't think it was it was a good idea now.
Added vars to human_defines.dm for bleed rate modifiers, which are used by the chems to modify the bleed rate. 
These chems respect the lowest modifier and reset on filtering out of the system. 
Done some specific fixes because of a weird floating point error involving internal_bleed_rate going to negatives.

-Tranexamic acid slows bleeding and internal bleeding significantly, however significant damage and instances of IB overcome it through sheer brute force, at least giving people time to get medical attention if nothing else.  Good to inject during surgery to reduce blood loss.

-Calcium Zeolite halts all bleeding entirely while it remains in system and has a low overdose threshold with nasty side effects and a small chance of the user getting a heart attack.

## Why It's Good For The Game

These chemicals add some quality of life to doctors and crew, there's very little in the way to levitate internal bleeding outside of surgery, pumping fresh blood, saline glucose solution or iron.
This adds two chemicals, one cheap and accessible and one harder to get, that slow bleeding and internal bleeding so multiple crew do not end up in medical and die from bleeding out because a doctor could not or can not make it to them in time.

## Changelog
:cl:
add: Tranexamic acid, reagent and recipe.
add: Calcium Zeolite ,reagent and recipe. 
~~add: new proc called get_bleed_modifier which is affected by whether the two chems are in the system with the stronger one overriding the other.~~
add: new parent specifically for bleed suppressing chems 
change: changes how bleed_rate was calculated
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine.dm
@@ -106,6 +106,59 @@
 		if(!Org.is_robotic())
 			Org.rejuvenate()
 
+/datum/reagent/medicine/tran_acid
+	name = "Tranexamic acid"
+	id = "tran_acid"
+	description = "A specialised chemical used for the treatment of truamatic bleeding and to control bleeding during surgery."
+	reagent_state = LIQUID
+	color = "#fcec96"
+	overdose_threshold = 40
+	harmless = FALSE
+	taste_description = "time running out"
+
+// This reagent's effects are handled in handle blood  code
+
+/datum/reagent/medicine/quikclot
+	name = "Calicum Zeolite"
+	id = "quikclot"
+	description = "A naturally occuring mineral used for treating mass hemorrhage. Use with caution."
+	reagent_state = LIQUID
+	color = "#f3d7a4"
+	overdose_threshold = 15
+	harmless = FALSE
+	taste_description = "Bonemeal"
+
+// This reagent's effects are handled in handle blood code
+
+/datum/reagent/medicine/quikclot/overdose_process(mob/living/M, severity)
+	var/list/overdose_info = ..()
+	var/effect = overdose_info[REAGENT_OVERDOSE_EFFECT]
+	var/update_flags = overdose_info[REAGENT_OVERDOSE_FLAGS]
+	if(severity == 1)
+		if(effect <= 1)
+			M.visible_message("<span class='warning'>[M] suddenly and violently vomits!</span>")
+			M.fakevomit(no_text = 1)
+		else if(effect <= 3)
+			M.emote(pick("groan","moan"))
+		if(effect <= 8)
+			M.emote("collapse")
+	else if(severity == 2)
+		if(effect <= 2)
+			M.visible_message("<span class='warning'>[M] suddenly and violently vomits!</span>")
+			M.fakevomit(no_text = 1)
+		else if(effect <= 5)
+			M.visible_message("<span class='warning'>[M.name] staggers and drools!</span>")
+			M.Dizzy(2)
+			update_flags |= M.Weaken(3, FALSE)
+		if(effect <= 15)
+			M.emote("collapse")
+	if(prob(5))	
+		if(istype(M,/mob/living/carbon/human))
+			var/mob/living/carbon/human/H = M
+			if(!H.undergoing_cardiac_arrest())
+				H.set_heartattack(TRUE) //this is a powerful clotting factor, overdose and experience SCD
+	return list(effect, update_flags)
+
 /datum/reagent/medicine/cryoxadone
 	name = "Cryoxadone"
 	id = "cryoxadone"
--- a/code/modules/reagents/chemistry/reagents/medicine.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine.dm
@@ -106,55 +106,6 @@
 		if(!Org.is_robotic())
 			Org.rejuvenate()
 
-/datum/reagent/medicine/tran_acid
-	name = "Tranexamic acid"
-	id = "tran_acid"
-	description = "A specialised chemical used for the treatment of truamatic bleeding and to control bleeding during surgery."
-	reagent_state = LIQUID
-	color = "#fcec96"
-	overdose_threshold = 40
-	taste_description = "time running out"
-// This reagent's effects are handled in handle_blood() code
-
-/datum/reagent/medicine/calzeo
-	name = "Calicum Zeolite"
-	id = "calzeo"
-	description = "A naturally occuring mineral used for treating mass hemorrhage. Use with caution."
-	reagent_state = LIQUID
-	color = "#f3d7a4"
-	overdose_threshold = 15
-	harmless = FALSE
-	taste_description = "Bonemeal"
-// This reagent's effects are handled in handle_blood() code
-
-/datum/reagent/medicine/calzeo/overdose_process(mob/living/M, severity)
-	var/list/overdose_info = ..()
-	var/effect = overdose_info[REAGENT_OVERDOSE_EFFECT]
-	var/update_flags = overdose_info[REAGENT_OVERDOSE_FLAGS]
-	if(severity == 1)
-		if(effect <= 1)
-			M.visible_message("<span class='warning'>[M] suddenly and violently vomits!</span>")
-			M.fakevomit(no_text = 1)
-		else if(effect <= 3)
-			M.emote(pick("groan", "moan"))
-		if(effect <= 8)
-			M.emote("collapse")
-	else if(severity == 2)
-		if(effect <= 2)
-			M.visible_message("<span class='warning'>[M] suddenly and violently vomits!</span>")
-			M.fakevomit(no_text = 1)
-		else if(effect <= 5)
-			M.visible_message("<span class='warning'>[M.name] staggers and drools!</span>")
-			M.Dizzy(2)
-			update_flags |= M.Weaken(3, FALSE)
-		if(effect <= 15)
-			M.emote("collapse")
-	if(prob(5))	
-		if(istype(M, /mob/living/carbon/human))
-			var/mob/living/carbon/human/H = M
-			H.set_heartattack(TRUE) //this is a powerful clotting factor, overdose and experience SCD
-	return list(effect, update_flags)
-
 /datum/reagent/medicine/cryoxadone
 	name = "Cryoxadone"
 	id = "cryoxadone"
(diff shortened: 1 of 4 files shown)


Carthusia:
ooohhhh, another chem to add to my paramedic kit, always keep an injector with 30u salb and a saline/epi mix, soon as you arrive at a near dead person, inject 5-10u of each, with this, this would help keep them from bleeding out during transit after the bandages wear off.

code/modules/surgery/organs/blood.dm
```diff
@@ -72,11 +72,17 @@
 
 		bleed_rate = max(bleed_rate - 0.5, temp_bleed)//if no wounds, other bleed effects (heparin) naturally decreases
 
-		if(internal_bleeding_rate && !(status_flags & FAKEDEATH))
-			bleed_internal(internal_bleeding_rate)
+		if(internal_bleeding_rate && !(status_flags & FAKEDEATH) && !reagents.has_reagent("quikclot"))
+			if(reagents.has_reagent("tran_acid"))
+				bleed_internal(internal_bleeding_rate * 0.3) 	//bleed_internal(internal_bleeding_rate - 0.7 * reagents.has_reagent("tran_acid") * internal_bleeding_rate) efficient alternative 
+			else
+				bleed_internal(internal_bleeding_rate)
```
moxian: Move this whole "bleed rate is modified by this percentage when you have reagents in you" logic into a standalone proc (`get_bleed_modifier()` or something). You'll then be able to plug it in both for internal and external bleeding without the need to duplicate code.

code/modules/reagents/chemistry/recipes/medicine.dm
```diff
@@ -12,6 +12,23 @@
...
+	result = "tran_acid"
+	required_reagents = list("oil" = 1, "ethanol" = 1, "ammonia" = 1)
+	result_amount = 3
+
+/datum/chemical_reaction/quikclot
+	name = "Calcium Zeolite"
+	id = "quikclot"
+	result = "quikclot"
+	required_reagents = list("milk" = 1, "aluminum" = 1, "silicon" = 1, "plasma_dust" = 1, "water" = 1)
+	result_amount = 3
```
moxian: 5u reagents give 3u zeolite. Is this intentional?

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -106,6 +106,59 @@
...
+	description = "A specialised chemical used for the treatment of truamatic bleeding and to control bleeding during surgery."
+	reagent_state = LIQUID
+	color = "#fcec96"
+	overdose_threshold = 40
+	harmless = FALSE
+	taste_description = "time running out"
+
+// This reagent's effects are handled in handle blood  code
+
+/datum/reagent/medicine/quikclot
```
moxian: Having name and id be so different is way too confusing, and not worth it IMO.
I agree that `calcium_zeolite` is awkward to write, but you write code once, and read... a lot. Even you, the PR author, would likely get a "wait, what the heck is quickclot?" moment several months down the road.

Aquilar: I was using the brand name of the mineral because it was fast and easy. I can change that to something else.

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -106,6 +106,59 @@
...
+		else if(effect <= 5)
+			M.visible_message("<span class='warning'>[M.name] staggers and drools!</span>")
+			M.Dizzy(2)
+			update_flags |= M.Weaken(3, FALSE)
+		if(effect <= 15)
+			M.emote("collapse")
+	if(prob(5))	
+		if(istype(M,/mob/living/carbon/human))
+			var/mob/living/carbon/human/H = M
+			if(!H.undergoing_cardiac_arrest())
```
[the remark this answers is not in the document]
Aquilar: yeah I'll get to fixing that

Aquilar:
> Design-wise, Calcium Zeolite seem way too difficult to make for not enough payoff. I don't think I would use it, but that might just be me.

the purpose of calcium zeolite is to cancel out all bleeding for a short duration so that the subject can be tended to, or tend to themselves. it's your, "Oh no, they going to bleed out before they can fix anything" chem. 
From a practical perspective, 15u will last around a minute, which is more than enough time to get to med bay and/or patch up the more serious bleeding, splint up, take pills.  

> Don't forget to update the wiki.

I will, that's if it ever gets merged

SteelSlayer:
I'd say that the zeolite creation difficulty is fine. The "hardest" things to get for it is the milk and plasma obviously, but even then its just a stroll to the bar for milk or a stroll to the ORM for plasma (provided there's actually plasma). And the other 3 ingredients are single clicks from the chem dispenser.

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -106,6 +106,55 @@
...
+// This reagent's effects are handled in handle_blood() code
+
+/datum/reagent/medicine/calzeo/overdose_process(mob/living/M, severity)
+	var/list/overdose_info = ..()
+	var/effect = overdose_info[REAGENT_OVERDOSE_EFFECT]
+	var/update_flags = overdose_info[REAGENT_OVERDOSE_FLAGS]
+	if(severity == 1)
+		if(effect <= 1)
+			M.visible_message("<span class='warning'>[M] suddenly and violently vomits!</span>")
+			M.fakevomit(no_text = 1)
```
moxian: `no_text = TRUE`
Same below. (You did apply the change via github, but that was subsequently lost due to a force-push)

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -106,6 +106,55 @@
...
+		else if(effect <= 5)
+			M.visible_message("<span class='warning'>[M.name] staggers and drools!</span>")
+			M.Dizzy(2)
+			update_flags |= M.Weaken(3, FALSE)
+		if(effect <= 15)
+			M.emote("collapse")
+	if(prob(5))	
+		if(istype(M,/mob/living/carbon/human))
```
moxian: space after coma

code/modules/surgery/organs/blood.dm
```diff
@@ -70,13 +70,21 @@
 			if(BP.internal_bleeding)
 				internal_bleeding_rate += 0.5
 
-		bleed_rate = max(bleed_rate - 0.5, temp_bleed)//if no wounds, other bleed effects (heparin) naturally decreases
+		bleed_rate = max(bleed_rate - 0.5, temp_bleed * get_bleed_modifier()) //if no wounds, other bleed effects (heparin) naturally decreases
```
moxian: `bleed_rate` is a `/mob/living/carbon/human` field, not a local variable.
So if you get calzeo in your blood for a single tick, `bleed_rate` would be immediately set to zero (`max(1234 - 0.5, 1234 * 0) == 0`), and stay there even after the chemical leaves your body.
I would suggest something like
```dm
bleed_rate = max(bleed_rate - 0.5, temp_bleed)  // bleeding heals naturally
temp_bleed = bleed_rate * get_bleed_modifier()  // bleed modifier is temporary and does accumulate across ticks
internal_bleeding_rate *= get_bleed_modifier()
//..
bleed(temp_bleed)
```

Aquilar: I made sure to test it before pushing, it resets after the chemical leaves. I don't see why this change would be necessary?

code/modules/reagents/chemistry/reagents/medicine.dm
Aquilar: this would make it inconsistent with the rest of the reagents, not just medicine.dm and rest of the reagents would have to be updated to reflect that and that would be beyond the scope of the RP

code/modules/surgery/organs/blood.dm
```diff
@@ -70,13 +70,26 @@
 			if(BP.internal_bleeding)
 				internal_bleeding_rate += 0.5
 
-		bleed_rate = max(bleed_rate - 0.5, temp_bleed)//if no wounds, other bleed effects (heparin) naturally decreases
+		if(temp_bleed > 0 || bleed_rate > 0 )
```
moxian:
```dm
if( a || b )
  if(a)
    stuff1()
  else
    stuff2()
```
is equivalent to
```dm
if(a)
  stuff1()
else if(b)
  stuff2()
```

code/modules/surgery/organs/blood.dm
```diff
@@ -70,13 +70,26 @@
...
+			if(temp_bleed)
+				bleed_rate = temp_bleed * get_bleed_modifier() //temp_bleed to bleed_rate
+			else
+				bleed_rate = max(0, bleed_rate - 0.5) //if no wounds, other bleed effects (heparin) naturally decreases and can not go below 0
```
moxian: I'm very sorry, I spent way too much time being confused over this.

So, your previous version of `bleed_rate = max(bleed_rate - 0.5, temp_bleed * get_bleed_modifier()) ` was totally correct. Apologies for misleading your here.
The only "problem" with that one is if you swallow heparin, get to, say 4 bleed_rate, and get a tiny bit of calzeo after, then after calzeo wears off, heparin would start working its way back from zero. But that's arguably *intentional*, and not a problem whatsoever.

Current code however has an interesting bug in that heparin is completely ineffective if you have a minor bruise and are bleeding at a rate of 0.01 from it - the `temp_bleed` would get set to a small, yet nonzero value, and `bleed_rate` would be set to `temp_bleed` in turn, forgetting whatever changes heparin has done to it.

TL;DR: `bleed_rate = max(bleed_rate - 0.5, temp_bleed * get_bleed_modifier())` is the right way to handle things; I'm an idiot and I'm sorry.

Fox-McCloud:
Now that Goon has publicly released a more up to date version of their code, I plan on implementing the blood pressure, clotting, and various blood chems, from there (filgrastim, proconvertin, heparin rework), so I'd prefer that, personally.

We're extremely accurate to goonchem (minus a few things), and we have gooncrit+goon med, more or less, so I'd prefer us stay in that direction.

Aquilar:
> Now that Goon has publicly released a more up to date version of their code, I plan on implementing the blood pressure, clotting, and various blood chems, from there (filgrastim, proconvertin, heparin rework), so I'd prefer that, personally.
> 
> We're extremely accurate to goonchem (minus a few things), and we have gooncrit+goon med, more or less, so I'd prefer us stay in that direction.

Will there at least be away that temporarily halts internal bleeding or is possible to cause?

Aquilar:
> How do these compare to the anti-bleed options available in goonchem?

for a simplicity sake: goon blood affecting reagent repair or increase bleeding, they also have a very different blood system. These bleeding repair mechanics do not affect internal bleeding or the rate of bleeding in anyway. 
these two as is, they slow or halt internal and external bleeding temporarily with the full force of the bleeding returning once they leave system.
through I'm willing to close and restart this should fox port goons blood handle to paradise, it looks great.

PR #12890, opened 2020-01-08
Fixes a runtime with revenants when they click on mobs

SteelSlayer:
## What Does This PR Do
Revenants use `/mob/living/attack_ghost(mob/dead/observer/user)` when they click on mobs, which passes itself as user. However there is no safety check to make sure that the user is actually an observer, which lead to runtimes about the revenant not having a `health_scan` var, which only observers should have.

## Why It's Good For The Game
Fixes runtimes

## Changelog
:cl:
fix: Fixes a runtime with revenants when they click on mobs
/:cl:

### DIFF
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -61,6 +61,8 @@
 
 // health + cyborg analyzer for ghosts
 /mob/living/attack_ghost(mob/dead/observer/user)
+	if(!istype(user))
+		return
 	if(user.client && user.health_scan)
 		if(issilicon(src) || ismachine(src))
 			robot_healthscan(user, src)
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -61,7 +61,7 @@
 
 // health + cyborg analyzer for ghosts
 /mob/living/attack_ghost(mob/dead/observer/user)
-	if(!istype(user))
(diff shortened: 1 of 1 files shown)


Cazdon:
It's nice to see the new runtime transparency policy already paying off!

moxian:
If `attack_ghost` is used by revenant clicking, but it is supposed to do nothing, then it should be removed from revenant clicking instead of papering over it here. 
Alternatively, we might decide to allow revenants to click things as if they were ghosts, but that's a gameplay change.

Revenant clicking bit:
https://github.com/ParadiseSS13/Paradise/blob/076ef673cda2630eed893db11eb27154a78d56ee/code/game/gamemodes/miniantags/revenant/revenant_abilities.dm#L24

code/_onclick/observer.dm
```diff
@@ -61,6 +61,8 @@
 
 // health + cyborg analyzer for ghosts
 /mob/living/attack_ghost(mob/dead/observer/user)
+	if(!istype(user))
```
moxian: Let's make it a bit more obvious to people reading this in the future why the type check is needed.
```suggestion
	// Revenants use attack_ghost when they click on things, but they are not observers and don't have .health_scan .
	if(!istype(user))
```

SteelSlayer: Sure why not.

PR #12893, opened 2020-01-09
Fixes a few runtimes 

### LINKED ISSUE #12892: Changeling Transform Sting Runtime
**Issue Description**:
Changeling transform sting can generate runtimes due to changeling eyes.

No runtimes.

**What happened instead**:
Runtimes. Transform sting may fail.

Unknown - but it is a fairly common runtime in the log, and hard to identify from runtime_summary since it doesn't list the file (the runtime happens in the /icon/proc/Blend builtin)

Use changeling transform sting while you and/or the target have changeling eyes active.

unknown

**Extra information**:
[2020-01-07T20:58:13] Runtime in ,: bad icon operation
   proc name: Blend (/icon/proc/Blend)
   usr: yyyyyy (as zzzzzzzzz) (xxxxxxxxx) (/mob/living/carbon/human)
   usr.loc: The plating (60,124,1) (/turf/simulated/floor/plating)
   src: /icon (/icon)
   call stack:
   /icon (/icon): Blend(null, 0, 1, 1)
   the protective membranes (/obj/item/organ/internal/cyberimp/eyes/shield/ling): generate icon(null)
   yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human): get eyecon()
   the head (/obj/item/organ/external/head): get icon(0)
   yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human): update body(0, 1)
   yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human): regenerate icons()
   yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human): UpdateAppearance(null)
   Transform (/datum/action/changeling/transform): transform dna(yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human), /datum/dna (/datum/dna))
   Transform (/datum/action/changeling/transform): sting action(yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human), null)
   Transform (/datum/action/changeling/transform): try to sting(yyyyyy (as zzzzzzzzz) (/mob/living/carbon/human), null)
   Transform (/datum/action/changeling/transform): Trigger()
   Transform (/obj/screen/movable/action_button): Click(null, "mapwindow.map", "icon-x=25;icon-y=23;left=1;scr...")

farie82:
## What Does This PR Do
Fixes
image
image
image
image

Fixes simple_mobs being able to nuzzle grilles to death at hyper speed

fixes #12892

## Why It's Good For The Game
Less runtimes more good

## Changelog
:cl:
fix: Fixes a runtime in npcpool.dm
fix: Fixes a runtime in atmospherics.dm
fix: Fixes a runtime in overmind.dm
fix: Fixes a runtime in Hallucinations.dm
fix: Fixes a runtime in augments_eyes.dm due to cling eyes not having color
fix: Losing an arm due to overal damage now won't cause a runtime in the hand attached to the arm
fix: Nuzzling a grille now won't kill it as simple animal
fix: Hitting a grille now won't cause a runtime as simple animal
/:cl:

### DIFF
--- a/code/ATMOSPHERICS/atmospherics.dm
+++ b/code/ATMOSPHERICS/atmospherics.dm
@@ -64,10 +64,8 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 
 // Icons/overlays/underlays
 /obj/machinery/atmospherics/update_icon()
-	var/turf/T = loc
-	if(!istype(T)) // Somehow loc became null
-		qdel(src)
-	if(level == 2 || !T.intact)
+	var/turf/T = get_turf(loc)
+	if(!T || level == 2 || !T.intact)
 		plane = GAME_PLANE
 	else
 		plane = FLOOR_PLANE
--- a/code/ATMOSPHERICS/atmospherics.dm
+++ b/code/ATMOSPHERICS/atmospherics.dm
@@ -66,7 +66,7 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 /obj/machinery/atmospherics/update_icon()
 	var/turf/T = loc
 	if(!istype(T)) // Somehow loc became null
-		return
+		qdel(src)
 	if(level == 2 || !T.intact)
 		plane = GAME_PLANE
 	else
--- a/code/ATMOSPHERICS/atmospherics.dm
+++ b/code/ATMOSPHERICS/atmospherics.dm
@@ -65,6 +65,8 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 // Icons/overlays/underlays
 /obj/machinery/atmospherics/update_icon()
 	var/turf/T = loc
+	if(!istype(T)) // Somehow loc became null
+		return
 	if(level == 2 || !T.intact)
 		plane = GAME_PLANE
 	else
--- a/code/controllers/subsystem/idlenpcpool.dm
+++ b/code/controllers/subsystem/idlenpcpool.dm
@@ -29,6 +29,7 @@ SUBSYSTEM_DEF(idlenpcpool)
 		var/mob/living/simple_animal/SA = currentrun[currentrun.len]
 		--currentrun.len
 		if(!SA)
+			log_debug("idlenpcpool encountered an invalid entry, resumed: [resumed], SA [SA], type of SA [SA?.type], null [SA == null], qdelled [QDELETED(SA)], SA in AI_IDLE list: [SA in GLOB.simple_animals[AI_IDLE]]")
 			GLOB.simple_animals[AI_IDLE] -= SA
 			continue
 
--- a/code/controllers/subsystem/npcpool.dm
+++ b/code/controllers/subsystem/npcpool.dm
@@ -22,6 +22,7 @@ SUBSYSTEM_DEF(npcpool)
         var/mob/living/simple_animal/SA = currentrun[currentrun.len]
         --currentrun.len
         if(!SA)
+            log_debug("npcpool encountered an invalid entry, resumed: [resumed], SA [SA], type of SA [SA?.type], null [SA == null], qdelled [QDELETED(SA)], SA in AI_ON list: [SA in GLOB.simple_animals[AI_ON]]")
             GLOB.simple_animals[AI_ON] -= SA
             continue
 
--- a/code/controllers/subsystem/npcpool.dm
+++ b/code/controllers/subsystem/npcpool.dm
@@ -21,6 +21,9 @@ SUBSYSTEM_DEF(npcpool)
     while(currentrun.len)
         var/mob/living/simple_animal/SA = currentrun[currentrun.len]
         --currentrun.len
+        if(!SA)
+            GLOB.simple_animals[AI_ON] -= SA
+            continue
 
         if(!SA.ckey && !SA.notransform)
             if(SA.stat != DEAD)
--- a/code/game/gamemodes/blob/overmind.dm
+++ b/code/game/gamemodes/blob/overmind.dm
@@ -48,13 +48,14 @@
 	update_health_hud()
 
 /mob/camera/blob/update_health_hud()
-	if(blob_core)
+	if(blob_core && hud_used)
 		hud_used.blobhealthdisplay.maptext = "<div align='center' valign='middle' style='position:relative; top:0px; left:6px'><font color='#e36600'>[round(blob_core.obj_integrity)]</font></div>"
 
 /mob/camera/blob/proc/add_points(var/points)
 	if(points != 0)
 		blob_points = Clamp(blob_points + points, 0, max_blob_points)
-		hud_used.blobpwrdisplay.maptext = "<div align='center' valign='middle' style='position:relative; top:0px; left:6px'><font color='#82ed00'>[round(src.blob_points)]</font></div>"
+		if(hud_used)
+			hud_used.blobpwrdisplay.maptext = "<div align='center' valign='middle' style='position:relative; top:0px; left:6px'><font color='#82ed00'>[round(src.blob_points)]</font></div>"
 
 /mob/camera/blob/say(var/message)
 	if(!message)
--- a/code/modules/surgery/organs/eyes.dm
+++ b/code/modules/surgery/organs/eyes.dm
@@ -5,7 +5,7 @@
 	organ_tag = "eyes"
 	parent_organ = "head"
 	slot = "eyes"
-	var/eye_colour = "#000000" // Should not be null
+	var/eye_colour = "#000000" // Should never be null
 	var/list/colourmatrix = null
 	var/list/colourblind_matrix = MATRIX_GREYSCALE //Special colourblindness parameters. By default, it's black-and-white.
 	var/list/replace_colours = LIST_GREYSCALE_REPLACE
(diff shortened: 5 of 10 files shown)


farie82:
Travis being blocked by Byond. Sad
image

Couls:
😭

code/modules/surgery/organs/eyes.dm
```diff
@@ -5,7 +5,7 @@
 	organ_tag = "eyes"
 	parent_organ = "head"
 	slot = "eyes"
-	var/eye_colour = "#000000"
+	var/eye_colour = "#000000" // Should not be null
```
Couls: Might be a better idea to check if the eye_colour is null on init/new and then set it to this instead, future-proof!

code/ATMOSPHERICS/atmospherics.dm
```diff
@@ -65,6 +65,8 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 // Icons/overlays/underlays
 /obj/machinery/atmospherics/update_icon()
 	var/turf/T = loc
+	if(!istype(T)) // Somehow loc became null
```
Couls: might be a good idea to qdel it if it's loc is null

code/modules/surgery/organs/eyes.dm
farie82: Added the lazy programmer protection. Not sure how I feel about it but I know it will happen again so why not add it

code/ATMOSPHERICS/atmospherics.dm
farie82: Can't think of a situation where it should still exist so yeh good one

Kyep: Maybe check if its already qdeleted before qdeling it.
Also... can there be situations where loc is not a turf, but also not null?
For example, if someone is transporting a stack of pipes or other atmos machinery unwrenched, inside a locker?

code/ATMOSPHERICS/atmospherics.dm
```diff
@@ -65,6 +65,8 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 // Icons/overlays/underlays
 /obj/machinery/atmospherics/update_icon()
 	var/turf/T = loc
+	if(!istype(T)) // Somehow loc became null
+		qdel(src)
```
Kyep: What if loc is, say, a locker or other container?
It can be things other than null, even if it isn't a turf.

code/modules/surgery/organs/eyes.dm
```diff
@@ -17,6 +17,11 @@
 	var/see_invisible = SEE_INVISIBLE_LIVING
 	var/lighting_alpha = LIGHTING_PLANE_ALPHA_VISIBLE
 
+/obj/item/organ/internal/eyes/Initialize(mapload, ...)
+	. = ..()
+	if(!eye_colour)
```
Kyep: If all the eyes have a var/eye_colour set to #000000 by default, is this needed?

code/ATMOSPHERICS/atmospherics.dm
moxian: What you want is `var/turf/T = get_turf(src)`, and skip istype check.
`istype` is the wrong fix here anyway. If you don't like `loc` getting null, you should investigate why didn't it qdel its contents upon becoming `null`, and not just paper over with an istype check. Deleting `src` in `update_icon` of all places is.. weird

code/game/gamemodes/blob/overmind.dm
```diff
@@ -48,13 +48,14 @@
...
 /mob/camera/blob/update_health_hud()
-	if(blob_core)
+	if(blob_core && hud_used)
 		hud_used.blobhealthdisplay.maptext = "<div align='center' valign='middle' style='position:relative; top:0px; left:6px'><font color='#e36600'>[round(blob_core.obj_integrity)]</font></div>"
 
 /mob/camera/blob/proc/add_points(var/points)
 	if(points != 0)
 		blob_points = Clamp(blob_points + points, 0, max_blob_points)
-		hud_used.blobpwrdisplay.maptext = "<div align='center' valign='middle' style='position:relative; top:0px; left:6px'><font color='#82ed00'>[round(src.blob_points)]</font></div>"
+		if(hud_used)
```
moxian: Do you maybe want to to fold it with `if` above? (i don't know)

code/ATMOSPHERICS/atmospherics.dm
farie82: It might be due to what kyet posted. That the thing is in a container. I'll change it

code/modules/surgery/organs/eyes.dm
farie82: It's not but I thought I'd add it seeing how the issue already was caused by somebody forgetting that it couldn't be null.
I'm myself still on the fence for this one though since it's only here to prevent coding mistakes (which prob will happen)

code/game/gamemodes/blob/overmind.dm
farie82: Fold it? As in add it to the points if? If so then no seeing the blobs points do need to be added on line 56

code/modules/surgery/organs/eyes.dm
moxian: Future coding mistakes are as likely to happen with eye_colour being null as, say, colourblind_matrix being null. If you argue that this check is useful, then you have to admit that this is only a half-measure, and you really need to set *all* the fields to their default values (and do that pretty much for every object). I hope you would agree that is overkill, since you *already* have default values set in the object definition.

code/game/gamemodes/blob/overmind.dm
moxian: You're right, ignore my comment.

code/ATMOSPHERICS/atmospherics.dm
farie82: Edit. I got no clue why that wouldn't be a turf or null. I'll change it to be more robust to runtimes but won't find the source of the issue

code/modules/surgery/organs/eyes.dm
farie82: Changed. eye_colour I can see getting set to null. But having this check promotes lazy/non_tested coding which I don't like. Also sets the wrong expectation seeing you can set it to null on init

code/controllers/subsystem/npcpool.dm
```diff
@@ -21,6 +21,9 @@ SUBSYSTEM_DEF(npcpool)
     while(currentrun.len)
         var/mob/living/simple_animal/SA = currentrun[currentrun.len]
         --currentrun.len
+        if(!SA)
+            GLOB.simple_animals[AI_ON] -= SA
+            continue
```
Fox-McCloud: This is a cheap way to fix it and not proper at all. We need to find the *source* of why it is null, and not paper over the problem.

code/ATMOSPHERICS/atmospherics.dm
```diff
@@ -64,8 +64,8 @@ GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 
 // Icons/overlays/underlays
 /obj/machinery/atmospherics/update_icon()
-	var/turf/T = loc
-	if(level == 2 || !T.intact)
+	var/turf/T = get_turf(loc)
+	if(!T || level == 2 || !T.intact)
```
Fox-McCloud: Similar to the other statement, why is `update_icon` being called when it's `loc` is something other than a turf?

code/controllers/subsystem/npcpool.dm
farie82: Copied from idlenpcpool.dm
But I'll see about finding the source of the issue

farie82: Tried to find some causes. Inbetween update (else I forget):
All destroys will remove the animal from the AIStatus they are assigned to.
There are some ways that AIStatus is changed without it using toggle_ai() and thus removing it from the list that NPCPool uses.
I'll check those ways to see if they fuck shit up

farie82: Probably won't have time myself to find the real cause. This will however fix the issue for now since it is affecting the game. Bots not working etc.

code/ATMOSPHERICS/atmospherics.dm
farie82: Happens due to blob_act and other reasons destroying the object. The destroy calls disconnect which calls update_icon.

code/controllers/subsystem/npcpool.dm
Crazylemon64: could you add some kind of logging to these null checks so that we have a lead on what's causing the nulls, since these kinds of errors will have impacts in other parts of the code, too

farie82: I'll add some debugging with tons of info.
Another reason that comes to mind for it to happen might be due to the resuming of the SS.

code/controllers/subsystem/idlenpcpool.dm
```diff
@@ -29,6 +29,7 @@ SUBSYSTEM_DEF(idlenpcpool)
 		var/mob/living/simple_animal/SA = currentrun[currentrun.len]
 		--currentrun.len
 		if(!SA)
+			log_debug("idlenpcpool encountered an invalid entry, resumed: [resumed], SA [SA], type of SA [SA?.type], null [SA == null], qdelled [QDELETED(SA)], SA in AI_IDLE list: [SA in GLOB.simple_animals[AI_IDLE]]")
```
farie82: @Crazylemon64 extra debugging added. Hopefully this will help

Crazylemon64:
sounds good

PR #12895, opened 2020-01-10
Adds the job and antag status to the offer to ghost question

farie82:
## What Does This PR Do
Adds the job and, if chosen by the offering admin also the antag status to the offer to ghosts verb.

## Why It's Good For The Game
Makes it easier for ghosts to decide if they want to play a random character.

## Images of changes
image

## Changelog
:cl:
tweak: The offer control to ghosts question now includes the job title and optionally the antag status if chosen so.
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -118,7 +118,10 @@
 	log_admin("[key_name(usr)] has offered control of ([key_name(M)]) to ghosts.")
 	var/minhours = input(usr, "Minimum hours required to play [M]?", "Set Min Hrs", 10) as num
 	message_admins("[key_name_admin(usr)] has offered control of ([key_name_admin(M)]) to ghosts with [minhours] hrs playtime")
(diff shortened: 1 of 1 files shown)


Kyep:
How does this handle cases where the name is really long?

Also, in cases where the 'Never' button does not work, could you either make it work, or remove it? (Not necessary for the PR, but a nice thing to have, and may touch the same code)

farie82:
> How does this handle cases where the name is really long?

image
Still looks decent. That name is the biggest players can make it.

> Also, in cases where the 'Never' button does not work, could you either make it work, or remove it? (Not necessary for the PR, but a nice thing to have, and may touch the same code)

I can take a look at the never button

farie82:
@Kyep I can fix it in this PR but then I'll remove the question for the admins to include the antag type to the question. Seeing otherwise players won't know to what they say never to.

In that question it will be hard to include the special_role value otherwise. It can be set to Event Character but this will be to generic. Admins making a special clown team? They will be the same as an admin polling for ghosts trying to replace a SSD sling

PR #12897, opened 2020-01-10
Antags that have their job unavailable will now spawn according to their preferences

### LINKED ISSUE #12771: Rolling antag causes you to ignore default role prefs
**Issue Description**:

If you roll antag and have 'be a civilian if preferences unavailable', you will be given a random job if none are available instead of being a civilian.

Be a civilian since I didnt roll the jobs I wanted.

**What happened instead**:

I became a random job.

Fairly edge case but this doesnt respect the user's preferences.

1. roll antag 
2. dont roll any of your preferred roles, have civilian as default
3. dont get civilian but a random job

Probably an edge case thats always been there.

**Extra information**:

Happened today 3/12/2019 on live server.

farie82:
## What Does This PR Do
Being an antag now makes you a civ if no jobs are available and you have that preference turned on.

Fixes #12771

## Why It's Good For The Game
Best to listen to peoples preferences init?

## Changelog
:cl:
fix: Antags now will spawn as civilians if their selected jobs are unavailable and they have the spawn as civ when job unavailable preference turned on
/:cl:

### DIFF
--- a/code/controllers/subsystem/jobs.dm
+++ b/code/controllers/subsystem/jobs.dm
@@ -389,8 +389,11 @@ SUBSYSTEM_DEF(jobs)
 	// Antags, who have to get in, come first
 	for(var/mob/new_player/player in unassigned)
 		if(player.mind.special_role)
-			GiveRandomJob(player)
-			if(player in unassigned)
+			if(player.client.prefs.alternate_option != BE_ASSISTANT)
+				GiveRandomJob(player)
+				if(player in unassigned)
+					AssignRole(player, "Civilian")
+			else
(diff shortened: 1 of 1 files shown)


Cazdon:
Just from looking at the code, that also means if they have 'return to lobby' they'll still be given a random job instead? So that'll avoid potential weird issues with antags being sent back to lobby hopefully.

farie82:
> Just from looking at the code, that also means if they have 'return to lobby' they'll still be given a random job instead? So that'll avoid potential weird issues with antags being sent back to lobby hopefully.

Those issues are solved by the other PR I posted. They were caused by another issue

PR #12902, opened 2020-01-12
Wizard Apprentices pass the IsWizard check. Wizard rounds end when all Wizards and Apprentices are dead.

Cazdon:
## What Does This PR Do
Title.

## Why It's Good For The Game
Apprentices not passing the IsWizard check causes a few checks to fail when you'd think they shouldn't. In example, a Soul Stone Shard cannot be used by an apprentice cause they don't pass the check. 

Previously, apprentices were listed as traitors, so instead they've been listed as apprentices. As a consequence, wizard no longer checks if there are any traitors left alive, which means wizard rounds will end even if there are survivalists alive due to summon guns or magic.

Personally I think this change would be appropriate, as summon guns and magic extends wizard rounds by a ton of time, when they're normally short and chaotic.

## Changelog 
:cl:
tweak: Wizard Apprentices pass the IsWizard check.
tweak: Wizard rounds end if all wizards and apprentices are dead, even if there's survivalists.
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/artefact.dm
+++ b/code/game/gamemodes/wizard/artefact.dm
@@ -99,7 +99,7 @@
 				new_objective:target = H:mind
 				new_objective.explanation_text = "Protect [H.real_name], the wizard."
 				M.mind.objectives += new_objective
-				SSticker.mode.traitors += M.mind
+				SSticker.mode.apprentices += M.mind
 				M.mind.special_role = SPECIAL_ROLE_WIZARD_APPRENTICE
 				SSticker.mode.update_wiz_icons_added(M.mind)
 				M.faction = list("wizard")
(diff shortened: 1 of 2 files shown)


Leanfrog:
So, take what I'm saying here with a grain of salt but what if the apprentice adopts the antag status of the one who summoned them, this is an edge case but earlier today I saw a member of command somehow get an apprentice scroll and they summoned one, if this were to potentially happen again wouldn't this mean that despite the apprentice not being bound to the wizard, the round wouldn't end because they're still alive

Cazdon:
@Leanfrog I've seen that happen too before the 'right and wrong' change. The crew proceeded to kill the apprentice and the round ended. While I agree it would be neat to have the apprentice scroll be not completely useless in the hands of a crew member, it's a bit outside the scope of this PR.

Issue #12807, opened 2019-12-15
Security Beret Has Better Melee Resist Than Helmet


TinnCatt:
Originally brought up by the player Gaty on the discord. 

Issue Description:
Currently our security berets have 5 more points of melee resist than the armored security helmets do, with the downside of marginally less fire resist on the beret (B-20 fire resist/H-50 fire resist). Those who brought this up wonder why this strange oversight or obvious beret bias has to be. 

What did you expect to happen:
For the values to be equal, or well, for the armored helmet's to be more.

What happened instead:
As stated above, the security berets have a better melee resist value than the security helmet. 

Why is this bad/What are the consequences:
Well, it means during normal rounds, it's slightly better for officers to ditch their stylish and iconic helmets in favor of the berets for a combat advantage. 

Steps to reproduce the problem:
None, just check the armor values for both.

When did the problem start happening:
Unsure. Not exactly a coder, so, I assume it's been like this for awhile and has been unnoticed. 

Extra information:
Here's two screenshots of the armor values after examining each with View Variables in-game, and a screenshot Gaty showed me. 

![Sec Helmet](https://user-images.githubusercontent.com/35948390/70857708-d50ad600-1ec1-11ea-9c2e-0335827799cf.PNG)

![Sec beret](https://user-images.githubusercontent.com/35948390/70857715-e94ed300-1ec1-11ea-96f5-43617c63fd5b.PNG)

![Values](https://user-images.githubusercontent.com/35948390/70857730-261aca00-1ec2-11ea-92f7-a281e8aa611d.png)

My suggestion, either make the stats equal, with the beret just being a fancier-looking sec helmet, or decrease the beret's melee resist by 5 so it's not slightly better than the helmet itself. 

Sorry this isn't mega helpful in finding the code for these, my knowledge of coding is extremely limited.

Shockpoint:
I only even took the beret because fashion...

I had no idea the beret was stronger LMAO

farie82:
Helmets have the advantage of blocking your ears when being stripped.

Ty-Omaha:
@TinnCatt When posting code don't post a screenshot of it. Click the line number and then copy the URL at the top.

PR #12907, opened 2020-01-16
Fixes some armour value oversights

Aquilar:
## What Does This PR Do
Fixes some oversights with various values of security armour
I consider the value changes for the berets and wintercoat fixes since the original values were probably never intended to be as high as they were so the security officer berets were brought to be inline with security soft cap because the accidentally inherited the same values as the hos and warden berets, as the general intention is your sacrificing protection for looks. 
while the security winter coat is an alternative the security jacket but ended up inheriting the armour values of the captains winter jacket, this in turn make it unintentionally better than the armour vest when accounting for the jumpsuit due the increased coverage it provided, it's also a load out item.

so the numbers:
- security and officers beret had their melee resistance lower by 5 and bomb resistance removed
- the HoS' beret get it's laser resistance increased by 5 to match the warden's
- security soft cap has it's armour values brought up to the berets along with removing the explosion resistance
- the armour values for the security winter coat were lowered to match the security jacket.

## Why It's Good For The Game

This solves some unintentional buffs to security head and suit gear.

## Changelog
:cl:
balance: fixes armour values across pieces of security gear, specifically berets and wintercoat
/:cl:

edit: removed all the changes related to the plasmamen helmets for another PR, cleaned up text.


Ty-Omaha:
What exactly is "Fashionsec"?

Ty-Omaha:
These look more like balance changes not fixes / oversights

Aquilar:
> What is "Fashionsec"?

security officers that are wearing the sec jacket/winter coat, the wintercoat is pretty strong and is a load out only item, and wearing the beret rather than the helmet because at the moment, there is zero reason to ever wear the helmet outside of being set on fire. 

and in response to this issue https://github.com/ParadiseSS13/Paradise/issues/12807

Ty-Omaha:
All of these values are not the same as what the security helmet currently has. 
https://github.com/ParadiseSS13/Paradise/blob/5936bb98fea2bc46769df7ee6bdf57a5f3f3db37/code/modules/clothing/head/helmet.dm#L8

Aquilar:
because they been pulled down to the values of the soft cap.

Ty-Omaha:
So it is not fixing any oversight at all, but changing the balance.

Aquilar:
The armour values for beret and wintercoat were added 4 months ago, the oversight here is the values were implemented without consideration or notice of 3 existing pieces of sec gear, the security cap, security jacket and security helmet. this ended up making the security wintercoat better than the standard armoured vest because similar armour values and better coverage, this is clearly unintentional, and the same goes for the beret. 

the additions to the plasmamen helmets were purely to bring them line with the regular head gear on the jobs those belong to.

Cazdon:
Why not make all the hat values on par with the helmet? I can at least see the reasoning behind the winter coat being put on par with the security jacket because it covers the arms, and the regular armor does not.

Why not just put all the hat values on par with the helmet and leave the players with more freedom of expression? Why make them strictly worse? I understand that they're not hard helmets, so maybe they shouldn't be as strong, but we already make big stretches in logic, and making another one as small as armor values for better freedom of expression isn't a bad trade at all. There's no reason to kill fashionsec.

Also you might want to explain why you view this as a fix in the original post, most people don't know why you would consider this a fix and it comes off as passive aggressively calling your balances changes fixes, despite the fact that I agree the Winter Coat thing is a fix.

Aquilar:
Duly noted, though many people would consider trading protection for looks a while worth trade. 
To be honest, it's 5 points of difference in armour has little impact in practice until you start hitting the bigger numbers. yeah lets make all the values same but keep the explosion resistance removed and lessened fire resistance, helmets needs to have a reason to be worn.

Krysonism:
> Why not make all the hat values on par with the helmet? I can at least see the reasoning behind the winter coat being put on par with the security jacket because it covers the arms, and the regular armor does not.
> 
> Why not just put all the hat values on par with the helmet and leave the players with more freedom of expression? Why make them strictly worse? I understand that they're not hard helmets, so maybe they shouldn't be as strong, but we already make big stretches in logic, and making another one as small as armor values for better freedom of expression isn't a bad trade at all. There's no reason to kill fashionsec.

If you want to wear a nice hat but you choose not to because it has less stats, you have a powergaming mindset that should not catered to on a RP server.

Even a lower RP level server like tg, has substantially lowered stats for the standard security berets and soft caps because it just makes sense.

This is a role playing game not an arena shooter, lower armour values have never stopped me from wearing a softcap on tg.

Denthamos:
Mmn, Aquilar has a point however.  The armor values being changed was an unintentional change and buff which was not listed nor detailed in the previous PR.  If a buff should be given to security gear then that should be handled in a separate PR for deliberation because I believe a security buff would be highly contested by people.

In the meantime a fix should be implemented so things are back to how they should be.

Aquilar:
> Mmn, Aquilar has a point however. The armor values being changed was an unintentional change and buff which was not listed nor detailed in the previous PR. If a buff should be given to security gear then that should be handled in a separate PR for deliberation because I believe a security buff would be highly contested by people.
> 
> In the meantime a fix should be implemented so things are back to how they should be.

so do I send the berets down to the softcap's original level or bring them in line with the helmet?

MRgaty:
I think there is no reason to not give all the hats the same armor values, even if it makes no sense ICly, I dont see a reason to punish the players who simply want to wear a hat with less armor.

Necaladun:
Armor values should be fairly consistent, this looks like oversight and mistakes, so it's good they're corrected.

Fox-McCloud:
please fix conflicts.

PR #12909, opened 2020-01-17
IPC survival kits

nl-208:
## What Does This PR Do

Adds in an IPC survival kit containing:
-  an emergency welder
-  a cable coil
-  a glowstick
-  a half-full oilcan (50u)
A new sprite for the kit is also included.

## Why It's Good For The Game

Currently, IPCs get the plain human survival kit at roundstart. However, most of the gear the kit contains is useless for the IPCs - they can't use an oxygen mask or inject epinephrine. The new kit gives them some basic gear to make repairs in a pinch while also not leaving trips to robotics or improved tools unneeded.
## Images of changes

closeup ingame
## Changelog
:cl:
add: add IPC survival kit containing basic repair tools
imageadd: add sprite for the new survival kit
/:cl:


Citinited:
I would support this PR only if you also add brute and burn patches to the standard survival box

Citinited:
This is less of a survival kit, and more of a medkit.

If there are any chems that only work on critical IPCs to prevent them dying then I would be in favour of those being added to their roundstart box; otherwise I'm not sure that giving free heals to machine people is a good idea.

Spacemanspark:
If an IPC doesn't pursue a welding tool and cable coil upon arrival, they are not worthy of survival

Silliness aside, I really don't like this. Maybe make an item with the fuel can's code that only fits around 50u of oil (if not less), but even then... eh? I admit I don't like the oil mechanic, but honestly this won't really be terribly handy considering how easy it is to damage an IPC and how easy it is to lose oil.

As for the other items, no. It would be almost on par with giving organics synth flesh at round start.

Yes, granted, IPCs are tin foil... but you can still get a welding tool and cable coil from cargo, worse comes to worse.

IPCs are and will always be hard mode (something I quite enjoy, even if a few of the mechanics are a bit lopsided--something Steelslayer is addressing in his current IPC mechanics PR). I don't see why they need to have a health kit at round start.

MRgaty:
This has been discussed many times and the answer has always been the same.
While rhe oil can is not too bad giving IPCs a way of healing at roundstart is not ok, the emergency boxes have a way to delay your death, they are not for healing purposes.
Giving IPCs literal brute kits and burn kits that can even be used as a weapon and a tool would not help the balance of the race.

Denthamos:
A better solution would be another chem created which could be used on IPC's to delay death and not heal for the same purpose that Epi pens have.

nl-208:
I've got the perfect solution to balance this out - just fill the damn box with nanopaste and ultralube.

jokes aside, I've read through the responses and according to consensus so far, it seems like making what's basically a version of the epipen for IPCs is the way to go, rather than regular repair tools in a box. I may have a look at chems this weekend and see what it'd take to make something like it.

also, @MRgaty - I'd like to hear what exactly you'd like changed with the box sprite, as I may come back to this at some point and might also change up the box's design a bit.

MRgaty:
Maybe you could use the robotics IPC repair kit to make it look like the rest of the emergency boxes, purple and white instead of black and maybe with their "epipen" as an icon, similar to how organics have their internals shown in their emergency box.

Carthusia:
Nanite Pen: Does the exact same thing epipens do, but for IPCs to prevent hardcrit for just a bit longer, one single injection, 10u?

Does nothing if injected into an organic

MRgaty:
but nanites are already in the game and they are for mending damage, having another version of the nanites but in a pen would be too confusing

Carthusia:
mmmm, I thought it was Nanopaste that mended the damage, not an item named Nanites, or does that exist to?

Perhaps then, i dono, just some kind of 1 use burner shock thing that, ICly just keeps them going for a bit longer like the Emergency Injector does?

nl-208:
I believe there are "nanite injectors", but they're fairly uncommon (only show up on DS and possibly gamma-level medical ERTs) and show up as "combat stimulants" or somesuch in the game. Still, though, we can definitely discuss different names for the chem.

from there, I'll just have whatever it ends up being called it work the same as epinephrine in the context of a synthetic - once they drop below a certain health level, try to keep them out of crit and standing for as long as the chem's in their system.

about the color scheme, that doesn't sound like a bad idea, actually - it'd definitely make the kit more recognizable as a synthetic-oriented one. I'll probably take another look at the design once the synthi-pens are coded and working.

Fox-McCloud:
This is free tools and effectively, medication, at round start, for a particular species.

Not a good precedent, at all, to set.

Humans get a single use pen that is only effective when they're in critical.

What exists here is more akin to a free burn patch, brute patch, and half a blood bag.

👎

nl-208:
right. well, now that the maintainers have weighed in, that's that - thanks for the feedback, all of you. it'll be taken into account for any future work on this. for now, I'll be closing this one.

In the coming week or two, I'll *actually* start work on the synthinephrine as a chem and the pens to administer it. from there, in the interest of keeping things atomized and if that one gets passed, I'll reintroduce a version of this PR.

Citinited:
Good luck!

PR #12910, opened 2020-01-17
Fixes the array index out of range in keyDown() due to players joining prior to SSInput is initialised

farie82:
## What Does This PR Do
Should fix the runtime where keys_held is accessed using a wrong index due to the array not being made properly.
Runtime in bindings_client.dm,39: list index out of bounds: proc name: keyDown (/client/verb/keyDown)

## Why It's Good For The Game
Very eager players who join before SSInput is started now won't cause runtimes

## Changelog
:cl:
fix: Players who join before SSInput has started now won't cause runtimes anymore
/:cl:

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -367,6 +367,8 @@
 
 	if(SSinput.initialized)
 		set_macros()
+	else
+		reset_keys_held()
 
 	donator_check()
 	check_ip_intel()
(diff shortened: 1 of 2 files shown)


Couls:
Don’t know if this will fix it, when the subsystem is done initializing it calls set macro on every client already https://github.com/ParadiseSS13/Paradise/blob/5b1941e784220cb7752909e44002754342dcf91d/code/controllers/subsystem/input.dm#L20

farie82:
> Don’t know if this will fix it, when the subsystem is done initializing it calls set macro on every client already

Might have happened due to a crash of the MC and thus the input SS perhaps? Although then this won't fix it either due to other things breaking
Runtime hasn't occured in the last few days either. 

Will close it since it won't fix the issue

PR #12911, opened 2020-01-17
Makes traitor AIs always malf

Ty-Omaha:
This gives AI traitors the objective to hijack always.

There is only 1 AI on station. AI traitors are rare as it is, and on top of that there was a 70% chance to **not** even try to make a hijack objective, but if it did pass, there was another 50% chance ontop of that. These odds are so low and make things very uninteresting and stale.

AI traitor objectives are "Turn off telecomms, announce ion, have secborg murder guy in maint and then  space/cremate him".  This allows for things to be more interesting and allows the crew team up and work together to stop the AI.

🆑 
balance: AI traitors will now always have hijack
/ 🆑

### DIFF
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -152,8 +152,7 @@
 	var/objective_count = 0
 	var/try_again = TRUE
 
-	if(prob(30))
-		objective_count += forge_single_objective()
+	objective_count += forge_single_objective()
 
 	for(var/i = objective_count, i < config.traitor_objectives_amount)
 		var/datum/objective/assassinate/kill_objective = new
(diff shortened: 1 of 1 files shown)


farie82:
Combining this with the protect objective will be interesting as well I'd say. Maybe with a chance for it to happen though

Birdtalon:
AI rouge.

AzuleUtama:
Makes sense to give them more freedom. Nothing would suck more than landing Malf. and gaining access to all those cool toys only to be told you can kill one civilian and that's it.

SteelSlayer:
I've been told by admins before even if you don't get hijack as AI, you're allowed to do things like use the borg machine or go delta. This would just cement that and make it clear though.

Spacemanspark:
@SteelSlayer that's actually false and at least one head admin has confirmed such.

However, it is agreed that tator AI seems kind of pointless without the hijack/ doomsday objective.

Cazdon:
Yeah, a traitor AI is a grey area in the rules. This makes things a lot simpler. It's not like AI assassinations won't take place either, subversions still are quite common.

The Malf Abilities are just way too overkill for taking out one or two dudes. The abilities are *station wide* and with 1 or 2 targets they're really unnecessary. Give 'em a reason to use 'em.

Also watching traitor AIs going back to doing normal AI things after successful kills is really lame.

SynthNumerics:
200% agreed. as malf ai you have so much potential an often uh, it attracts a lot of borgs who want to be part of the fun- which is great- but when you're just playin normally while waitin for later in the round so you dont mess your target up too soon or you get things done an you now have 45 minutes to kinda... do not a lot it sucks pretty hard. bein able to actually keep somethin goin w a big plan makes it way more interestin, esp like as mentioned its super rare.

subversions are for sure in a good place to do that kinda thing as is.

Mitchs98:
I'm on the fence about this. On one hand, it could make things more interesting. On the other hand, it could just lead to the AI mass murdering/borg machining the crew every single time they roll malf. It's, overall, much easier for an AI to cause such chaos than it is a traitor.

But on the other hand, traitor AI's are rare-ish as it already is so it shouldn't be too big of a problem.

Probably.

Quantum-M:
I agree with @Mitchs98 . It does allow the AI to deal with crew that, as usual every time the AI is discovered rogue, to rush the AI.

But as stated, it could just lead the AI to just murdering the crew the moment it has enough points/plasma flood the station (and it would most likely end up prolonging the round with a lot of dead people, as the command staff are most likely to be dealt with first by the AI, if the crew manages to kill the AI).

It also has to be considered that traitors, or any rounds that includes traitors (e.g. changeling/vampires), will most likely end up having no obligation to help the AI (since the AI has traitor comms) and in fact may be encouraged to go against it.

Maybe a consideration on giving the AI some different objectives that isn't just plainly taking out/protecting or destroying the station/surviving?

E.g. 
Accumulate X amounts of cpu points.
Have X amount of cameras at the end of the round
Have X amount of surviving cyborgs at the end of the round.

PR #12912, opened 2020-01-17
Fixes barsigns turning into regular signs when attacked by a screwdriver; attacking it with an ID card will let you pick the sign

Citinited:
Title

Stops you ruining the bartender's day with a screwdriver. Instead it will just open and close the maintenance panel

:cl:
fix: Unscrewing a barsign won't drop it as an invisible regular sign any more
tweak: Attacking a barsign with an ID card will open the sign pick menu
/:cl:


Citinited:
This will be picked up in another PR (where I edit a tonne of attackbys)

PR #12918, opened 2020-01-22
Camera action buttons & better handling of camera movement/double-clicking

Citinited:
**Please note: this PR depends on global signals, which I will port along with the DCS subsystem at a later date. So don't merge until that's done**

**What does this PR do:**
- Tweaks the movement between cameras when watching the camera console. There are now action buttons for that. Moving with the arrow keys will move your mob as normal.
- Changes how double-click-to-move-to-new-camera works: before, it put that in base of `atom/DblClick()`. Now, it uses a signal.
- Performance-related: cameras now cache the result of their movement to help cut down on CPU usage

This should eliminate a bunch of overhead by not checking for camera-related stuff on 99.999% of mob moves and double clicks

**Obligatory picture**
image

:cl:
add: Moving between cameras using the camera console now uses action buttons.
tweak: Stops camera console-specific checks on 99.999% of mob movement and double-click calls
/:cl:


farie82:
Does this also make camera movement more smooth? Since it was a bit ... off from time to time. Going left made you not able to go right again etc

Citinited:
Yeah, the old camera movement definitely had some weird behaviour. Previously, the camera move selected a tile 10 units away from itself, then found the closest camera it could inside that turf’s area, then jumped to it if possible. This is why it was so buggy – for example it might completely skip over a camera because it was in a different area to where the camera was searching. One amusing bug I found is that sometimes you’ll end up jumping to the telecoms satellite (because your console was searching in space, and the closest camera in space was at telecoms) and be unable to return via camera movement :smile:

Instead, now the camera console searches in a narrowish cone in one particular direction, and jumps to the first valid camera that it can find. Therefore, any camera it can jump to is at least in the desired general direction of where you wanted to jump to, and closer than any other camera in that direction. It can still be a bit weird but the camera move should make a lot more sense when this is merged :smile:

In the below picture, the green squares are the ones it will search for, while ignoring anything in white.

image

farie82:
Love the detailed answer. Thanks and good shit man

Citinited:
You're welcome!

Citinited:
@Kyep Does that apply to all of this PR or just a part of it? Just curious why this one would be a maintainer-only thing 😄

Fox-McCloud:
we have DCS now.

If I don't hear from you in a week though, this will be closed due to staleness.

Fox-McCloud:
CLosing due to staleness.

PR #12924, opened 2020-01-25
Terror queen spit is now blockable

farie82:
## What Does This PR Do
Terror queens spit is now blockable like the message you get says it does.

## Why It's Good For The Game
It doesn't make sense to block something and still get hit by it

## Changelog
:cl:
fix: Queen terror spit is now blockable like the message says it is
/:cl:


Kyep:
Blockable by what?

Drakeven:
I presume by shields, riot shields, energy shields, ect.

farie82:
> Blockable by what?

Indeed blockable by shields, weapons, suits etc. Anything that can block and will say that it blocked the shot but it currently still does damage

PR #12934, opened 2020-01-29
Make every obj/.../New() call it's parents, Fixes things not burning on lava and causing runtimes

farie82:
## What Does This PR Do
Makes all objects call their parents New
Fixes:
image

I would like a test merge for this just to be sure that nothing broke. 

"\/obj\/item(\/[A-Z_0-9]*)*\/New\(\)" was used to find all the News

## Why It's Good For The Game
Less runtimes when things burn on lava and just more expected behaviour and better code

## Changelog
:cl:
fix: All things now burn properly on lava
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/melee/energy.dm
+++ b/code/game/objects/items/weapons/melee/energy.dm
@@ -119,6 +119,7 @@
 	var/hacked = 0
 
 /obj/item/melee/energy/sword/New()
+	..()
 	if(item_color == null)
 		item_color = pick("red", "blue", "green", "purple")
 
@@ -220,7 +221,7 @@
 	light_color = LIGHT_COLOR_RED
 
 /obj/item/melee/energy/sword/pirate/New()
-	return
+	..()
 
 /obj/item/melee/energy/blade
 	name = "energy blade"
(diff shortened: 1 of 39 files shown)


Citinited:
Would it be better to convert all these `New()`s to `Initialize()`? Given we probably want to do that for everything eventually 😄

farie82:
> Would it be better to convert all these `New()`s to `Initialize()`? Given we probably want to do that for everything eventually 😄

I did think about that. But that would also include changing the code below these and such... bit much for myself currently to do. But feel free to do it :D

Citinited:
Ahh right, thanks for checking! If I were to do that it would be in a mass refactor :P

code/game/objects/items/weapons/melee/energy.dm
```diff
@@ -220,7 +221,7 @@
 	light_color = LIGHT_COLOR_RED
 
 /obj/item/melee/energy/sword/pirate/New()
-	return
+	..()
```
Fox-McCloud: this shouldn't exist at all.

code/game/objects/items/weapons/storage/backpack.dm
```diff
@@ -63,7 +63,6 @@
 
 /obj/item/storage/backpack/holding/New()
 	..()
-	return
```
Fox-McCloud: same

code/game/objects/items/weapons/melee/energy.dm
farie82: Darn forgot about those. Removed in the newest commit

PR #12938, opened 2020-01-29
Fixes the issue with files being locked by BYOND

AffectedArc07:
## What Does This PR Do
This fixes a long standing bug where BYOND & RUST-G would lock the files and not write the contents until DreamDaemon was either closed or crashed. This is bad because it doesnt allow background file sync tasks, and will may RAM usage higher if the files are not being written. This PR fixes that issue by simply recompiling RUST-G. 

Dont question it, I am just as confused as you are

## Why It's Good For The Game
DD shouldnt be locking files just because it touched them **once**, and apparently this is a pain for backend stuffs

## Changelog
:cl: AffectedArc07
tweak: RUST-G (The logging library for the server) no longer locks files when the server is running
/:cl:


Kyep:
Context:

- The idea came up of having a "status" command in Discord which would show current crew manifest, round duration, etc.
- Instead of making Topic() calls for this data, I proposed instead that, every 5 minutes, during the EXP update proc, the server simply write out a HTML file containing a manifest and any other common status information to its data directory, which could then be SFTPed up to the webserver and looked at by anyone. Effectively, rather than a bot checking the server whenever anyone ran a command, a "within the last 5 minutes" status page would always be available.
- However, this relies on byond being able to write a status file without keeping a permanent lock on it (as sftp won't upload a file that another process has a write lock on, it seems).
- So, this PR is updating rust so that byond no longer keeps unnecessary file locks and thus exporting stuff from the server in other PRs for status pages/etc is viable.

This is not a super-critical PR, and it would be worth checking the md5s of the compiled rust DLL versus one you compile yourself, but it could be helpful.

AffectedArc07:
I plan to add RUST-G building CI and CD to the rust-G repo, but that can wait.

tigercat2000:
Not that it really matters without a logging back-end anymore, but Paradise technically uses it's own version of rust_g that tweaks the format codes. https://github.com/ParadiseSS13/rust-g

AffectedArc07:
@tigercat2000 I literally just rebuilt para's RUST-G. No actual changes made.

tigercat2000:
Oh, so that's what you were saying was confusing. That sure is confusing. What.

AffectedArc07:
My only guess is that dependencies got updated and it no longer locks a file.

PR #12946, opened 2020-01-30
Fixed a few funtimes and bugs

farie82:
## What Does This PR Do
Fixes constructs not being able to teleport using teleport runes.
Fixes constructs not being able to be teleported by the Rite of Joined Souls rune.

Fixes these runtimes:
image

New fix since 2-3-2020 (occurences on 3-2-2020: 24 times):
image

New fix since 2-3-2020 (occurences on 3-2-2020: 2 times):
image

## Why It's Good For The Game
Less runtimes more funtimes

## Changelog
:cl:
fix: Cult constructs can now use the teleport rune without causing a runtime
fix: Cult constructs can be teleported using the joined souls rune without causing a runtime
fix: Nuking the station won't cause runtimes anymore. People can now also walk after being nuked and living
/:cl:


code/controllers/subsystem/ticker.dm
```diff
@@ -309,15 +309,12 @@ SUBSYSTEM_DEF(ticker)
 	cinematic.mouse_opacity = MOUSE_OPACITY_TRANSPARENT
 	cinematic.screen_loc = "1,0"
 
-	var/obj/structure/bed/temp_buckle = new(src)
```
farie82: This is the source of quite some runtimes and I myself don't see the need for buckling mobs down. Walking around doesn't break shit

Crazylemon64: sure why not

code/modules/customitems/item_spawning.dm
```diff
@@ -11,6 +11,9 @@
 		var/propadjust = query.item[2]
 		var/jobmask = query.item[3]
 		var/ok = 0
+		if(!path)
+			log_debug("Incorrect database entry found in table 'customuseritems', cuiPath is null. cuiCKey='[M.ckey]' AND (cuiRealName='[sanitizeSQL(M.real_name)]' OR cuiRealName='*'")
```
Crazylemon64: couldn't this also be caused by the path being typed incorrectly? Or does that just do a runtime

farie82: I'll test this case. But I think you might be right

farie82: It also caused the same exception. Added a ispath check and the path value to the logging

code/game/gamemodes/cult/runes.dm
```diff
@@ -332,10 +332,12 @@ var/list/teleport_runes = list()
 			user.forceMove(get_turf(actual_selected_rune))
 		var/mob/living/carbon/human/H = user
 		if(user.z != T.z)
-			H.bleed(5)
+			if(istype(H))
```
Fox-McCloud: Why not just return, above, if it's not human? This doesn't seem to be something anything other than a human can do anyway.

farie82: Constructs can also do this. Or they should be able to do this. Currently they cannot because it runtimes on them

PR #12948, opened 2020-02-01
Crossbow Update + Skin rework  :)

Diegoflores31:
## What Does This PR Do
*Creates new Skin for The crossbow 
*Gives it an extra damage 
*Deletes the Tension Adjustments 
*removes the throw range of the makeshift bolt to make it only usable in crossbow 
*adds embeed effect to the makeshift Bolts 

## Why It's Good For The Game
Crossbow has ben forgotten in a very very dark place long time ago , its time to rise him up to life , new dmg , new skin less Wait times compared to the old one to be more attractive and introduces the embeed , something that players dont usualy has to deal with , and has a penalisation for removing by hand so medbay will have some nice work performing Forgotten surgery

## Images of changes

pixil-frame-0 (5)
pixil-frame-0 (4)
pixil-frame-0 (3)
pixil-frame-0 (8)
pixil-frame-0 (7)
crosbow-pixilart

## Changelog
:cl:
del: Most of Old crossbow code 
balance: Damage increased for makeshift bolts
wip:  Still needs a little tweak in the Firing process since you can still fire it winout need to Tension it 
imageadd:  adds new icons for the crossbow
/:cl:


MRgaty:
Mmmh am a little concerned about giving crew a reliable craft-able ranged weapon.
The stunprod are already ridiculous but at least they are not ranged and has low damage...

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -29,41 +21,48 @@
...
 	. = ..()
 	if(cell)
 		. += "<span class='notice'>\A [cell] is mounted onto [src]. Battery cell charge: [cell.charge]/[cell.maxcharge]"
+		update_icon()
 	else
 		. += "<span class='notice'>It has an empty mount for a battery cell.</span>"
 
-/obj/item/gun/throw/crossbow/modify_projectile(obj/item/I, on_chamber = 0)
-	if(cell && on_chamber && istype(I, /obj/item/arrow/rod))
+/obj/item/gun/throw/crossbow/modify_projectile(obj/I,)
```
Citinited: This comma should be removed

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -29,41 +21,48 @@
...
+	if( tension && (cell.charge >= 1999))
```
Citinited: Remove this space between the `(` and `cell.charge`

Also, you should change `1999` to `2000` because in the event of the cell having exactly 1999 charge, it won't use any (given you are doing `cell.use(2000)` later on)

I would put all your code below this in something like the following:

```
if(tension && cell.use(2000))
    code_goes_here
```
This is because the `use()` proc returns `TRUE` if the cell actually used the charge (and didn't fail), so we know we can proceed with the rest of the code

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -29,41 +21,48 @@
...
 		var/obj/item/arrow/rod/R = I
 		visible_message("<span class='danger'>[R] plinks and crackles as it begins to glow red-hot.</span>")
-		R.throwforce = 15
+		R.throwforce = 40
```
Citinited: Seems a bit high... 40 damage is quite a bit of damage!

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -29,41 +21,48 @@
...
 /obj/item/gun/throw/crossbow/get_throwrange()
-	return tension * range_multiplier
+	return 5 * range_multiplier
 
 /obj/item/gun/throw/crossbow/process_chamber()
 	..()
 	if(to_launch)
 		modify_projectile(to_launch, 1)
-	update_icon()
+	  update_icon()
```
Citinited: There's some extra whitespace here which is why this isn't compiling. You only want one tab here, no spaces

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -87,9 +85,12 @@
 	if(!to_launch)
 		to_chat(user, "<span class='warning'>You can't draw [src] without a bolt nocked.</span>")
 		return
+	if(!cell)
+		to_chat(user, "<Span Class='warning'> Baterry Cell is missing.</span>")
```
Citinited: `span class` shouldn't be capitalised.

Also battery is spelt incorrectly

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -29,41 +21,48 @@
...
 
-/obj/item/gun/throw/crossbow/update_icon()
-	if(!tension)
-		if(!to_launch)
-			icon_state = "[initial(icon_state)]"
-		else
-			icon_state = "[initial(icon_state)]-nocked"
-	else
-		icon_state = "[initial(icon_state)]-drawn"
+obj/item/gun/throw/crossbow/update_icon()
```
Citinited: I think you need to call `..()` somewhere in this

Also, you need a `/` in front of `obj`

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -1,26 +1,18 @@
...
-#define XBOW_TENSION_80 "80%"
-#define XBOW_TENSION_FULL "100%"
-
 /obj/item/gun/throw/crossbow
-	name = "powered crossbow"
-	desc = "A modern twist on an old classic. Pick up that can."
-	icon_state = "crossbow"
-	item_state = "crossbow-solid"
+	name = "Makeshift crossbow"
+	desc = "A Wood Cell Powered kind of Crossbow."
```
Citinited: Is it powered by a wood cell, or is it wooden and cell-powered? May I suggest `A wooden, cell-powered crossbow.` as the description?

code/modules/projectiles/guns/throw/crossbow.dm
```diff
@@ -29,41 +21,48 @@
...
+	if(tension)
+		icon_state += "-Charged"
+	if(cell.charge >= 500 && tension)
+		icon_state +="-Hot"
 
 /obj/item/gun/throw/crossbow/examine(mob/user)
 	. = ..()
 	if(cell)
 		. += "<span class='notice'>\A [cell] is mounted onto [src]. Battery cell charge: [cell.charge]/[cell.maxcharge]"
+		update_icon()
```
Citinited: Why does it call `update_icon()` in the examine proc?

code/modules/projectiles/guns/throw/crossbow.dm
taukausanake: Double the original, maybe 25, would be a more reasonable buff, but maybe 15 isn't bad either

code/modules/projectiles/guns/throw/crossbow.dm
taukausanake: _I see Botany has been busy again_

Citinited: Wooden you know it 🤔

Spacemanspark:
Needs pictures of the new sprites for all to see.

Also not sure how I feel about this overall.

code/modules/projectiles/guns/throw/crossbow.dm
Diegoflores31: its not exactly 1999 as you can see the cell charge thing says >= that means more than 1999
starting from 2000

Diegoflores31:
> Needs pictures of the new sprites for all to see.
> 
> Also not sure how I feel about this overall.

Done :)

code/modules/projectiles/guns/throw/crossbow.dm
Diegoflores31: Fixed , and yeah actually would be good if could be only loaded with potato batteries nice idea

code/modules/projectiles/guns/throw/crossbow.dm
blinkdog: `x >= 1999` means: x is greater than or equal to 1999
If you want starting from 2000, it should probably be as Citnited suggested.

`x > 1999` also works for integers, but if `cell.charge` was a floating value, then `1999.5` would pass the check. It seems you want strict 2000+ ?

Diegoflores31:
> Needs pictures of the new sprites for all to see.
> 
> Also not sure how I feel about this overall.
done

code/modules/projectiles/guns/throw/crossbow.dm
Diegoflores31: done , thanks for helping me

Ty-Omaha:
The neon red looks extremely out of place. I thought it was you circling something at first

Diegoflores31:
the red is the cable coil used for the crossbow its made like that because making a fine piece out of a common wire its not something you do with a screwdriver and a blowtorch

Ty-Omaha:
@Diegoflores31 if I used yellow cable coil would it be yellow?

taukausanake:
@Ty-Omaha I think the improvised shotgun can use cables for a sling so you can put it on  your back (or at least the sprites exist for it) and those are just red. You could probably use mutable icons for this if you really want to. That would be neat

Diegoflores31:
is mean to be like a shoot and chase weapon or a shoot and run , it doesnt really do too much damage but will stop your target from running away faster than you or it can give you a chance to escape in a complicate situation , the rod embeeds into the target dealing some damage and starts a little damage every cicle until its removed from the body , if removed wildly will deal more damage but if a medic removes it you will recieve no damage

MRgaty:
I think is way too powerful for an easily crafted ranged weapon that can fit in your bag along with its easily crafted ammo.

Spacemanspark:
The crossbow could certainly use some love, but this isn't the way to go about it.

Diegoflores31:
ok , i will work on a better sprite and making a new Pr later

Issue #12723, opened 2019-11-10
Malf AI neutron bomb not counting down nor working


farie82:
**Issue Description**:
The countdown of the neutron bomb doens't go down nor does the bomb go of on it's own

**What did you expect to happen**: 
All humanoids to die and sillicons to live

**What happened instead**:
A lone rogue AI having to pray to the sillicon gods to make the humanoids go away. Aka no boom without admin intervention

**Why is this bad/What are the consequences**:
It's broke yo

**Steps to reproduce the problem**:
Use it

**When did the problem start happening**:
Unknown, for atleast 1 or more months now

**Extra information**:

farie82:
Happens due to the processing of the doomsday device starting before it is properly set up. Then it will try to start it under another processing system, fail and think it is started.

PR #11700, opened 2019-06-07
[Testmerge ready] Ports tool behaviours; refactors all tools; adds functionality for self-filling reagent containers

Citinited:
**What does this PR do:**
- Ports tool behaviours from tgstation, a very nice method of handling tools
- Reworks welder code to not be so large 
- Refactors all tools to use the new system and some minor refactoring of tool logic
- Refactors personal crafting to use the new procs
- Refactors construction steps to use the new procs
- ~~Refactors disposals to use the new system~~ haaahahahaaha
- Provides defines for some commonly-used tool messages, to save on typing, and gives neatness and consistency
- Numerous tweaks and bugfixes (see the changelog)
- Splits up tools into their own files as opposed to one big one
- Fixes #6838
- Fixes #6844
- Fixes #12205
- Fixes #10544
- Fixes #12623

Non-tool-related tweaks:
- Allows you to make reagent containers refill with specified reagents
- Slightly modifies welder repair times on objects - ranges from 0.5 to 5 seconds, depending on object damage

**Important note**
After this PR, you should handle all tool interactions via the relevant `tool_act()` proc.

**To-do:**

- [x] Initial tool framework and set tool behaviours on the tools
- [x] Convert all existing code to be compliant
- [x] Hunt for bugs / test
- [x] Fix #6838

**Changelog:**
:cl:
tweak: Refactors tool code by a lot; ported from tgstation.
tweak: You can now set plasma floors on fire using a welder.
tweak: You can no longer construct flamethrowers using the old method - use the crafting menu, instead.
tweak: Using a welder to repair things has a variable timer now.
tweak: You can now carve out books using more than just knives and wirecutters (anything sharp also works).
tweak: Use a crowbar to deconstruct the cargo telepad. (Instead of a welder, for some reason.)
fix: Fixes a bug allowing you to instantly trap people inside lockers.
fix: Fixes a bug allowing you to unanchor powered emitters.
fix: Fixes a few bugs related to using multiple tools during construction of stuff.
fix: Welders that turn off automatically will now update their mob's sprite.
fix: Welding a plating that you aren't either fixing or removing no longer burns your eyes.
fix: You can no longer turn any type of vendor into another vendor.
fix: Cutting the AI control wire on air alarms now properly disables that control.
/:cl:

PR #12952, opened 2020-02-03
[DNM] Machinery processing refactor; Converts some machines to event-based processing

SteelSlayer:
Prerequisite PR: #11700

Putting a DNM on this because it's going to conflict to hell with the tool refactor and probably several other PRs since it modifies so many files.

## What Does This PR Do
This PR converts about 30 machines over "event-based processing" for lack of a better term. The idea is that these machines will now only begin processing when a certain criteria is met. For example, sleepers and cryotubes will now only process when there is an occupant, since there is no reason for them to be processing otherwise.

Replaces the `speed_process` var with a `processing_flags` var. Four different bitflags are meant to go here, which are:
```js
#define START_PROCESSING_ON_INIT     1
#define START_PROCESSING_MANUALLY    2
#define NORMAL_PROCESS_SPEED         4
#define FAST_PROCESS_SPEED           8
```
By default, all `obj/machinery` types start with these flags:
```js
processing_flags = START_PROCESSING_ON_INIT | NORMAL_PROCESS_SPEED
```
if you want to tell a machine to *not* start processing when the round starts, just replace the first flag with `START_PROCESSING_MANUALLY` instead. This allows it bypass the code in `/obj/machinery/Initialize()` that would normally tell it to immediately start processing. Then, simply find the places in the code in which you want to tell the machine to either being or end processing and use the `begin_processing()` and `end_processing()` procs respectively.

I won't explain absolutely everything I added here, but that was just a little introduction to the basics. Check out `machinery.dm` or `objs.dm` for more information on the procs available. If you have questions about specifics, just ask.

## Why It's Good For The Game
This will save a small amount of server resources, because the game won't have to pointlessly execute code when it doesn't need to.

## Changelog
:cl:
tweak: Modifies a number of machines so that they will not constantly be processing even though there is no need for them to be. This will save some server resources. Back-end change only, there are no gameplay changes.
/:cl:

### DIFF
--- a/code/__DEFINES/machines.dm
+++ b/code/__DEFINES/machines.dm
@@ -12,6 +12,12 @@
 #define IDLE_POWER_USE 1
 #define ACTIVE_POWER_USE 2
 
+// bitflags for a machine's processing preferences
+#define START_PROCESSING_ON_INIT 	1	// if the machine should start processing on Initialize()
+#define START_PROCESSING_MANUALLY 	2	// if the machine will start processing in the future, after Initialize(), such as when a certain condition is met
+#define NORMAL_PROCESS_SPEED		4	// normal processing speed
+#define FAST_PROCESS_SPEED			8	// fast processing speed
+
 //computer3 error codes, move lower in the file when it passes dev -Sayu
  #define PROG_CRASH      1  // Generic crash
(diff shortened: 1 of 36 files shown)


farie82:
Fixes: #12723

SteelSlayer:
This is never going be to be merged.

PR #12963, opened 2020-02-09
Scan mind and Project mind now use vision to determine the naming in the target list

farie82:
## What Does This PR Do
Both Scan Mind and Project Mind now use vision of the mob using it to determine if they know who they sense. The target will be listed as "Unknown entity" If they lack vision to the target.

## Why It's Good For The Game
Reduces the meta radar that greys get at round start. And seeing it's not a mindreading ability they shouldn't know.

## Images of changes
Test situation:
image

Without xray:
image
image

With xray:
image
image

## Changelog
:cl:
balance: Project mind now uses vision to determine if you see the name in the target list. If you can't see it it will be listed as unknown entity
balance: Scan mind now uses vision to determine if you see the name in the target list. If you can't see it it will be listed as unknown entity
/:cl:


datlo:
*SCREAM*

good change

Tayyyyyyy:
Since it's based on vision now, couldn't you get rid of the awkward popup menu entirely? Maybe make it like the changeling sting where once you activate the ability you can click on the target to use it on them?

code/game/dna/genes/vg_powers.dm
```diff
@@ -215,23 +215,37 @@
...
+			names += mob_name
+			validtargets[mob_name] = M
 
 	if(!validtargets.len)
 		to_chat(user, "<span class='warning'>There are no valid targets!</span>")
 		start_recharge()
 		return
 
-	targets += input("Choose the target to talk to.", "Targeting") as null|mob in validtargets
+	var/target_name = input("Choose the target to listen to.", "Targeting") as null|anything in names
```
datlo: I believe that's supposed to be 'Choose the target to talk to'

farie82: Good catch. Yep copy pasta fail

code/game/dna/genes/vg_powers.dm
```diff
@@ -268,23 +286,41 @@
 	var/list/targets = new /list()
 	var/list/validtargets = new /list()
 	var/turf/T = get_turf(user)
+	var/list/mobs_in_view = user.get_visible_mobs()
```
Crazylemon64, line 289: since you only use this list to check for the presence of items in, would it make more sense to just replace this with the applicable `view` call

farie82: I'm expecting the `in` operator to do a check that costs `O(m + n)` time. Where m is the amount of mobs in sight and n the rest objects. In the current situation it will only cost `O(m + n) + m*O(M)` time. Instead of `m * O(m+n)` time

SteelSlayer (commented):
Something I noticed is that the two procs `/obj/effect/proc_holder/spell/targeted/remotetalk/choose_targets` and `/obj/effect/proc_holder/spell/targeted/mindscan/choose_targets` have basically identical logic for finding targets. For example, line 215 to 238, and 286 to 309, could be made into a proc.

And then for example, you can do something like
```js
/obj/effect/proc_holder/spell/targeted/remotetalk/choose_targets(mob/user = usr)
	var/list/mytargets = get_telepathic_targets() // Made up a proc name here.

	if(!mytargets.len)
		to_chat(user, "<span class='warning'>There are no valid targets!</span>")
		... // and so on
```
Where `get_telepathic_targets()` is the proc that has the duped logic. Bit of pseudo code and such but hopefully that makes sense.

farie82:
> Something I noticed is that the two procs `/obj/effect/proc_holder/spell/targeted/remotetalk/choose_targets` and `/obj/effect/proc_holder/spell/targeted/mindscan/choose_targets` have basically identical logic for finding targets. For example, line 215 to 238, and 286 to 309, could be made into a proc.
> 
> And then for example, you can do something like
> 
> ```js
> /obj/effect/proc_holder/spell/targeted/remotetalk/choose_targets(mob/user = usr)
> 	var/list/mytargets = get_telepathic_targets() // Made up a proc name here.
> 
> 	if(!mytargets.len)
> 		to_chat(user, "<span class='warning'>There are no valid targets!</span>")
> 		... // and so on
> ```
> 
> Where `get_telepathic_targets()` is the proc that has the duped logic. Bit of pseudo code and such but hopefully that makes sense.

Yep good one. Fixed it in the latest commit

code/modules/mob/mob.dm
```diff
@@ -500,6 +500,41 @@ var/list/slot_equipment_priority = list( \
...
+/**
+ * Returns an assoc list which contains the mobs in range and their "visible" name.
+ * Mobs out of view but in range will be listed as unknown. Else they will have their visible name
+*/
+/mob/proc/get_telepathic_targets()
+	var/list/validtargets = new /list()
+	var/turf/T = get_turf(src)
+	var/list/mobs_in_view = get_visible_mobs()
+
+	for(var/mob/living/M in range(14, T))
```
SteelSlayer, line 519: Since `range()` returns a list.. is it possible to do something like:
```dm
for(var/mob/living/M in range(14, T) - src)
```
and then skip the:
```dm
    if(M == src)
        continue
```
?

farie82: Interesting idea. I've benchmarked both and the check version is slightly faster. about 69 MS vs 56 MS.
I'll keep it how it is now to be more conform to other checks as well

SteelSlayer: Oh, sad..

variableundefined (approved):

PR #12964, opened 2020-02-09
Fixes bots not being trackable using the chat link as AI

farie82:
## What Does This PR Do
Makes bots like beepsky trackable as ai when clicked on the name in chat when they say something

## Why It's Good For The Game
It's currently broken

## Changelog
:cl:
fix: Makes bots like beepsky trackable as when clicked on their name in chat
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/bot.dm
+++ b/code/modules/mob/living/simple_animal/bot/bot.dm
@@ -146,7 +146,7 @@
 	access_card.access += access_robotics
 	set_custom_texts()
 	Radio = new/obj/item/radio/headset/bot(src)
-
+	Radio.follow_target = src
 	add_language("Galactic Common", 1)
 	add_language("Sol Common", 1)
 	add_language("Tradeband", 1)


Citinited:
This has been broken since forever, good work 👍

Shockpoint:
Will this work for ghosts too, or just AI?

farie82:
> Will this work for ghosts too, or just AI?

It should fix it for ghosts as well seeing that the track target was null and I assume that the ghosts also use that

Spacemanspark:
I think at the time of the last bot changes, it was said this was some unfortunate side effect.

Thank fuck if you've managed to fix it.
Now all we need is for the AI to be able to click on floor tiles when moving bots around once again...

farie82:
> Now all we need is for the AI to be able to click on floor tiles when moving bots around once again...

Thanks for bringing that up. Kinda forgot about it myself. Fixed in #12968

Spacemanspark:
Yay

PR #12978, opened 2020-02-15
Antag objectives are now announced when a target cryos

AzuleUtama:
## What Does This PR Do
When an antagonist's target is cryo'd, it will now announce the updated objectives to them rather than having them check their notes.

## Why It's Good For The Game
Less M/AHelps about where a traitor needs to find their new target when they get the alert.

## Changelog
:cl:
tweak: Antagonists will now have their new assassination targets announced should their original target enter cryo.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -394,11 +394,11 @@
 				spawn(1) //This should ideally fire after the occupant is deleted.
 					if(!O) return
 					O.find_target()
-					O.owner.announce_objectives()
 					if(!(O.target))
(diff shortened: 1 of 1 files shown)


code/game/machinery/cryopod.dm
```diff
@@ -394,6 +394,7 @@
 				spawn(1) //This should ideally fire after the occupant is deleted.
 					if(!O) return
 					O.find_target()
+					O.owner.announce_objectives()
```
Crazylemon64: This might work better after the if block below, so that it doesn't announce the objective if it's got a null target

Issue #12973, opened 2020-02-13
Picket signs have no character limit


Sikroth:
**Issue Description**:

What the title says.

**What did you expect to happen**: 

To not be able to copypaste the entire Bee Movie script into a sign.

**What happened instead**:

I was in fact able to copypaste the entire* Bee Movie script into a sign.

*disclaimer: I didn't actually paste the whole thing, just the first 10000 characters. Seemed like an arbitrarily long string but luckily didn't result in a broken chat.

**Why is this bad/What are the consequences**:

If 10000 chars fit in I imagine there's no limit to it. So you could in theory paste millions of chars, breaking the chat or at the very least being extremely annoying. Or potentially worse. Also you might get bwoinked for copypasting the entire Bee Movie script into a sign.

**Steps to reproduce the problem**:

1. Get a picket sign (I actually don't know how to craft one ingame. Maybe with wood and cardboard?)
2. Paste / write a really long string in it using a pen.
3. Annoy the hell out of everyone by filling the entire chat window with a button press, or even when someone simply examines you.

**When did the problem start happening**:

I imagine since they were first introduced. This is why you stress test things.

**Extra information**:

According to all known laws of aviation, there is no way a bee should be able to fly. Its wings are too small to get its fat little body off the ground. The bee, of course, flies anyway because bees don't care what humans think is impossible.

taukausanake:
>Not posting WGW or the Navy Seal copypasta instead

but I guess those are quite a bit shorter than the Bee Movie

Sikroth:
My first idea was WGW (just hinted at doing it IC, didn't really plan on doing it) but then a helpful Poly suggested the bee movie script which was both funny and a great way to test huge text lenghts. So that's what I did :D

PR #12983, opened 2020-02-15
stripped_input now actually uses the max length at all times

farie82:
## What Does This PR Do
Makes stripped_input use the max_length parameter.

Fixes: #12973

## Why It's Good For The Game
Fixes an exploit where people could fill in unlimited characters to certain objects like picket signs.

## Changelog
:cl:
fix: stripped_input now uses the max_length parameter. Picket signs now have a limit
/:cl:

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -108,11 +108,10 @@
 
 // Used to get a sanitized input.
 /proc/stripped_input(mob/user, message = "", title = "", default = "", max_length=MAX_MESSAGE_LEN, no_trim=FALSE)
-	var/name = input(user, message, title, default) as text|null
-	if(no_trim)
-		return copytext(html_encode(name), 1, max_length)
-	else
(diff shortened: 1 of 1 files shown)


code/__HELPERS/text.dm
```diff
@@ -108,11 +108,10 @@
 
 // Used to get a sanitized input.
 /proc/stripped_input(mob/user, message = "", title = "", default = "", max_length=MAX_MESSAGE_LEN, no_trim=FALSE)
-	var/name = input(user, message, title, default) as text|null
-	if(no_trim)
-		return copytext(html_encode(name), 1, max_length)
-	else
-		return trim(html_encode(name), max_length) //trim is "outside" because html_encode can expand single symbols into multiple symbols (such as turning < into &lt;)
```
Crazylemon64: weird that this even compiles, trim doesn't take more than 1 argument

PR #12988, opened 2020-02-16
Fixes welders blinding you when you repair robo-limbs and cyborgs

### LINKED ISSUE #12987: IPCs are blinding themselves during self-repair
**Issue Description**:

IPCs are blinding themselves during self-repair with the welders.

It is annoying, and it wasn't like this about 2 days ago.

**What happened instead**:

Got flashed each and every time I fixed a dent.

Making it annoying to play.

Get denting damage, use welder on the targeted area, get flashed by it.

2 days ago it was fine, or even just a few hours ago.

**Extra information**:

Not much to add.

SteelSlayer:
## What Does This PR Do
What the title says basically. This was an unintentional change/oversight from the tool refactor, judging by the CL from that PR. 

This may not be the right way to handle this but it seemed fine to me and I wanted to take a crack at messing with the new tool behavior stuff. Maybe @Citinited can correct me if I'm wrong.

Fixes #12987

## Why It's Good For The Game
Adds back in a feature that got unintentionally removed.

## Changelog
:cl:
fix: Welders no longer blind you while repairing your own robo-limbs
/:cl:


code/game/objects/items/tools/tool_behaviour.dm
```diff
@@ -39,8 +39,8 @@
 
 // Called before use_tool if there is a delay, or by use_tool if there isn't.
 // Only ever used by welding tools and stacks, so it's not added on any other use_tool checks.
-/obj/item/proc/tool_start_check(mob/living/user, amount=0)
-	return tool_use_check(user, amount)
+/obj/item/proc/tool_start_check(atom/target, mob/living/user, amount=0)
```
Citinited: You don't need to pass the target to this proc, this is used only to check if the tool can be used (which is independent of what the target is)

Citinited:
For reference, this is the line that actually blinds you when fixing your limbs: https://github.com/ParadiseSS13/Paradise/blob/6455f645096fcf7c31b4d834e0d8c5865be5e073/code/modules/mob/living/carbon/human/human_defense.dm#L60

`use_tool()` on a welding tool always blinds you if it succeeds, which makes sense (even though previously it would not blind you while repairing your own limbs). 

off the top of my head I can't think of a nice way to make this not blind you without passing a var to `use_tool()` which is not a good idea IMO. You could make use of the `use()` proc though, which doesn't blind you if it succeeds

SteelSlayer:
@Citinited But the tool_start_check for welders contains the code that will flash the user with `user.flash_eyes(light_intensity)`. The target gets passed up the chain for a quite a while but then stops here. I don't see what else you could do to negate this unless I'm missing something.

Citinited:
`tool_use_check` doesn't blind you

SteelSlayer:
Well it turns out that if you don't pass target to `tool_start_check`, you can't pass the target to `tool_use_check`. The proc as is without target being passed:
```js
/obj/item/weldingtool/tool_start_check(mob/living/user, amount=0)
	. = tool_use_check(user, amount)
```

code/game/objects/items/tools/welder.dm
```diff
@@ -100,9 +100,9 @@
 		return FALSE
 
 // When welding is about to start, run a normal tool_use_check, then flash a mob if it succeeds.
-/obj/item/weldingtool/tool_start_check(mob/living/user, amount=0)
+/obj/item/weldingtool/tool_start_check(atom/target, mob/living/user, amount=0)
 	. = tool_use_check(user, amount)
-	if(. && user)
+	if(. && user && target != user)
```
SteelSlayer: Leaving this as reminder. Apparently people are getting flashed from other people's robo-limbs and also from repairing borgs. Instead of what I wrote here, this should probably have a check for `!issythetic(target)` or whatever. It's a snowflakey method, but maybe there's a better way

KasparoVy:
This is intentional balancing factor elsewhere but here it be boog/oversight

PR #12989, opened 2020-02-16
Consolidates Jaws of life and hand drill into a single file and de-spaghettis them [DNM]

Aquilar:
## What Does This PR Do

Does it modify too many files? yes
Could be done better? also yes
Is it necessary? Absolutely 

This PR consolidates and changes the entries of the jaws of life and hand drill into a single new file, removing the previous entries as well.
~~Adding some coding that allows for tools with two states to have different tool behaviours without them interfering with each other. Allowing them to swap between sprites and the two behaviours without issue, they could even have no tool behaviours if wished.~~
Completely reworked this so it'll now accept as many tool behaviours and icon states as one wants.
rework required adding a proc to do it though. 
also added to so it tells you which mode the item is in currently on examine. 

Preserves the suicide_acts and the changing function sounds.
No longer requires an item to destroy itself in game to just swap to another tool behaviour.

## Why It's Good For The Game

Having an item destroy itself to spawn another item is terrible now that we finally have tool behaviours, the previous system it was the only way of doing it this is slicker and allows coders to add more similar tools without having to edit 6 different files, now they can edit just 2 or 3.

## Changelog
:cl:
add: New ~~powertool~~ multifunctiontool path
add: Reworked jaws of life and hand drill
change: Various file paths that needed adjusting
del: removed old power tool entries and paths
/:cl:


code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
+//Power Tools
+/obj/item/powertool
+	var/stateswap = 0 //what state is it in
+	var/icon_state_swap = 0 //which icon should it be
+	var/tool_behaviour_swap = 0 //what tool behaviour(s) should be here
+	var/swapsound = 0 //the sound that plays for swapping
+
+/obj/item/powertool/attack_self(mob/living/carbon/user) //activating the item causes it to swap icons and behaviours
+	stateswap = !stateswap
```
moxian: It would be much neater if this could support more than two states. Something like
```dm
/obj/item/powertool
  var/list/tool_behaviors= list()
  var/list/tool_icon_states = list()
  var/active_behavior_index = 1

/obj/item/powertool/attack_self(mob/living/carbon/user)
  active_behavior_index = (active_behavior_index + 1) % possible_tool_behaviors.len + 1
  tool_behaviour  = possible_tool_behaviors[active_behavior_index]
  icon_state = tool_icon_states[active_behavior_index]
...
/obj/item/powertool/drill
  ...
  tool_behaviors = list(TOOL_SCREWDRIVER, TOOL_WRENCH)
  tool_icon_states = list("drill_screw", "drill_bolt")
```

(But given that none of the tools currently *need* this, this might not be necessary)

code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
...
+	icon_state_swap = "drill_bolt"
+	hitsound = 'sound/items/drill_hit.ogg'
+	usesound = 'sound/items/drill_use.ogg'
+	swapsound = 'sound/items/change_drill.ogg'
+	flags = CONDUCT
+	slot_flags = SLOT_BELT
+	force = 5
+	throwforce = 7
+	w_class = WEIGHT_CLASS_SMALL
+	materials = list(MAT_METAL=150,MAT_SILVER=50,MAT_TITANIUM=25)
```
moxian: Nit: space after commas (same below).
I know you've simply moved this from another place, but this is inconsistent with the other lists in here if nothing else.
(feel free to ignore)

code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
...
+	origin_tech = "engineering=1;combat=1"
+	attack_verb = list("attacked", "bashed", "battered", "bludgeoned", "whacked")
+	toolspeed = 0.25
+	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 50, "acid" = 30)
```
moxian: Armor? On a drill? (same below)

code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
...
+	throwforce = 7
+	w_class = WEIGHT_CLASS_SMALL
+	materials = list(MAT_METAL=150,MAT_SILVER=50,MAT_TITANIUM=25)
+	origin_tech = "materials=2;engineering=2"
+	attack_verb = list("attacked", "bashed", "battered", "bludgeoned", "whacked")
+	toolspeed = 0.25
+	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 50, "acid" = 30)
+	tool_behaviour = TOOL_CROWBAR
+	tool_behaviour_swap = TOOL_WIRECUTTER
+	var/airlock_open_time = 100 // Time required to open powered airlocks
```
moxian: `airlock_open_time` is never used anywhere.

code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
...
+	icon_state = "drill_screw"
+	item_state = "drill"
+	icon_state_swap = "drill_bolt"
+	hitsound = 'sound/items/drill_hit.ogg'
+	usesound = 'sound/items/drill_use.ogg'
+	swapsound = 'sound/items/change_drill.ogg'
+	flags = CONDUCT
+	slot_flags = SLOT_BELT
+	force = 5
+	throwforce = 7
```
moxian: This used to be
```
  force = 8 //might or might not be too high, subject to change
  throwforce = 8
  throw_speed = 2
  throw_range = 3
```
in the screwdriver form and
```
  force = 8 //might or might not be too high, subject to change
  throwforce = 8
  // /obj/item defaults:
  throw_speed = 2
  throw_range  = 7
```
in the wrench form.

Why the change to 5/7 ? Where do those numbers come from?

code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
...
+	name = "jaws of life"
+	desc = "A set of jaws of life, the magic of science has managed to fit it down into a device small enough to fit in a tool belt. has prying and cutting attachments."
+	icon = 'icons/obj/tools.dmi'
+	icon_state = "jaws_pry"
+	item_state = "jawsoflife"
+	icon_state_swap = "jaws_cutter"
+	usesound = 'sound/items/jaws_pry.ogg'
+	swapsound = 'sound/items/change_jaws.ogg'
+	flags = CONDUCT
+	slot_flags = SLOT_BELT
```
moxian: I think it might be a good idea to move `SLOT_BELT` and `WEIGHT_CLASS_SMALL` to the base `/obj/item/powertool` class (but I don't feel too strong about this)

code/game/objects/items/tools/crowbar.dm
```diff
@@ -58,28 +58,4 @@
...
-	playsound(loc, 'sound/items/jaws_pry.ogg', 50, 1, -1)
-	return BRUTELOSS
-
-/obj/item/crowbar/power/attack_self(mob/user)
-	playsound(get_turf(user), 'sound/items/change_jaws.ogg', 50, 1)
-	var/obj/item/wirecutters/power/cutjaws = new /obj/item/wirecutters/power
-	to_chat(user, "<span class='notice'>You attach the cutting jaws to [src].</span>")
-	qdel(src)
-	user.put_in_active_hand(cutjaws)
+	toolspeed = 0.5
```
moxian: Add newline at the end of file. (Same in powertool.dm)

code/__HELPERS/tool_helpers.dm
```diff
@@ -36,7 +36,7 @@
 	return FALSE
 
 /proc/ispowertool(O)//used to check if a tool can force powered doors
-	if(istype(O, /obj/item/crowbar/power) || istype(O, /obj/item/mecha_parts/mecha_equipment/medical/rescue_jaw))
+	if(istype(O, /obj/item/powertool/powerjaws) || istype(O, /obj/item/mecha_parts/mecha_equipment/medical/rescue_jaw))
```
moxian: Do we need to check that powerjaws are on the "crowbar" setting?

code/game/objects/items/tools/powertool.dm
```diff
@@ -0,0 +1,86 @@
...
+	force = 15
+	throwforce = 7
```
moxian: Just for the record, the current values are:

For the crowbar form:
```
  force = 15
  throwforce = 7
  throw_speed = 2
  throw_range = 7
```
For the wirecutters form:
```
  force = 6
  throwforce = 0
  throw_speed = 3
  throw_range = 7
```

code/__HELPERS/tool_helpers.dm
```diff
@@ -36,7 +36,7 @@
 	return FALSE
 
 /proc/ispowertool(O)//used to check if a tool can force powered doors
```
moxian: Interesting that not everything under `/obj/item/powertool` counts as "power tool"...
(Not something you can really fix, just an observation).

code/game/objects/items/tools/powertool.dm
Aquilar: still good for posterity

Aquilar:
I'm gonna need a do not merge tag until all these istype checks are dealt with.

code/__HELPERS/tool_helpers.dm
Citinited: Might be better to have this as an item level proc and just return TRUE if the item is a powertool

code/game/objects/items/tools/powertool.dm
Aquilar: lets go a bit further and add on examine as well.

code/game/objects/items/tools/multifunctiontool.dm
```diff
@@ -1,76 +1,89 @@
...
+/obj/item/multifunctiontool/attack_self(mob/living/carbon/user) //activating the item causes it to swap icons and behaviours
+	if(tool_behaviours.len > 1)
+		select_mode(user)
+	tool_behaviour = tool_behaviours[active_behaviour_index]
+	icon_state = tool_icon_states[active_behaviour_index]
 	playsound(get_turf(user), swapsound, 50, 1)
-	to_chat(user, "<span class='notice'>You attach a different bit to [src].</span>")
+	to_chat(user, "<span class='notice'>You use a different part on [src].</span>")
+
+/obj/item/multifunctiontool/proc/select_mode(mob/living/user)
```
moxian: Several issues here (read: probably just one issue, but can be addressed in multiple different ways):
1) I think a better name would be `cycle_mode`. I think `select` implies that it gets passed an index of the mode you want to switch towards.
2) this function doesn't actually change the mode, it only adjusts the index. It would be cleaner if it *returned* the  desired active_behaviour_index value instead of adjusting it in-place
3) As is, i would much prefer this to be inlined into attack_self, and not be a standalone function
4) It can be condensed from 3 lines down to one `active_behaviour_index = (active_behaviour_index + 1) % tool_behaviours.len + 1`

I think at least either 2 or 3 need to be addressed here (the rest are whatever-ish)

code/game/objects/items/tools/multifunctiontool.dm
```diff
@@ -1,76 +1,89 @@
...
+	active_behaviour_index++
+	if(active_behaviour_index > tool_behaviours.len)
+		active_behaviour_index = 1
 	return
+
+/obj/item/multifunctiontool/examine(mob/user)
+	. = ..()
+	. += "[tool_mode_names[active_behaviour_index]]"
```
moxian: Does this properly add spaces/punctuation? I.e. does this not give you `Power DrillScrewdriver bit attached`?

Fox-McCloud:
This is arguably worse.

Better to just do the refactor here: https://github.com/tgstation/tgstation/pull/46974

Inheritance+the tool behavior refactor makes it much cleaner.

Aquilar:
@Fox-McCloud 
noted, still have to wait until all the tool related istype checks are changed to tool behaviours otherwise doing either is point at the moment.

PR #12999, opened 2020-02-17
Replaces a bunch of food item sprites with new and improved versions

SteelSlayer:
## What Does This PR Do
**Sprites by: @Pewtershmitz**

Replaces a bunch of food item sprites with new and improved versions. There's some on here that I didn't bother with right now, namely the stew bowls and the chocolate bar with wrapper ones. But maybe someone can use those in the future, or I may come back and do something with them myself.

## Why It's Good For The Game
Improved sprites for things that looked bad.

## Images of changes
**Note: not all of these icon names are representative of the actual item's name**
"cocoa" is actually the sprite for the pile of chocolate for example

R

E

## Changelog
:cl: SteelSlayer, Pewtershmitz
add: Adds the ability to change the look of a cereal box using a pen
imageadd: Replaces a bunch of food item sprites with new and improved versions, also adds completely new ones such as more types of cereal boxes
/:cl:


Cazdon:
That looks *really* good! 


I will say the Cocoa powder looks a little.... lumpy... and not so powdery. It gives me the wrong idea at a glance, if you catch my drift.

SteelSlayer:
Yeah actually, I should rename that because that's not the sprite for cocoa powder, its suppose to be the sprite for a pile of chocolate. Actually the picture of all the sprites and names is from the original sprite sheet I was handed and I changed most of the names. I'll put this into the description

Carthusia:
I like the sprites, though they seem a tad...big? Today I was thinking our current muffin sprite was fairly huge, but the one shown above looks like its at least 25% larger.

Are they just large in the picture to show how they look? A single box of cereal that big could crush a man to death

SteelSlayer:
I was thinking some of them may be too big too, but 32x32 is hard to sprite with, and you lose detail when go smaller, which is why a lot of ss13 sprites are huge like that.

Also yeah, the in-game screenshot is the actual size of the sprites, and the new cereal boxes use the existing plain cereal box as a base, so these are the same size.

Carthusia:
I like the sprites quality and look alot, i just can't really get over how huge they are in comparison to other objects, its really off putting and in all honesty, kind of goofy

https://i.gyazo.com/8f4ec33d6ab1358300ae63aa5b0c4659.png

If each food item was that big, I would have an extremely difficult time organizing my kitchens, it's already difficult with muffins and some food items being as huge as they are to place them on tables with multiple objects and make them look presentable.

Is it possible to shrink them to be more on part with other food objects?

Pewtershmitz:
Heyo, Pewts here!

I figured I should refer to the feedback on the sprite size. Yeah, some of them are ridiculously big, but you need to understand that you It would take an enormous amount of work, not to mention "elbow grease" to do everything to scale. Like Steel mentioned, we spriters are stuck to 32x. 

The issue is the smaller you go, the harder It is to make shapes, and add shading. Another issue is the smaller the sprite, the harder it is to click it. This is one of the reasons that items are generally bigger than they have to be. I should also mention these sprites used the ones that existed as the base, meaning they pretty much retain their dimension. Either that, or they have been changed.

As for making sprites smaller, I'd like to mention that this whole game is a mix of styles, parts and perspectives, and as such, items are no homogeneous. It comes down to it being 32x for -everything-, unless you're making special sprites that can reach 64x or higher. I personally don't think how I could make some of these items smaller.

SteelSlayer:
Closing because Pewter is currently dealing with real life stuff and is unable to update his sprites for the time being. We'll probably re-open this in a few months, if the sprites don't get updated by someone else in that time.

PR #13003, opened 2020-02-18
Grants the Blueshield additional access

datlo:
The Blueshield currently has a frustratingly limited amount of station access to start with.

In theory, this can be remedied in rounds by the HOP or Captain granting the Blueshield the extra access they need. 

In practice, to save time and because 1. The Blueshield cant be an antag and 2. The Blueshield is rarely completely incompetent due to server rules and karma requirements, the captain just chucks the spare ID at the blueshield to do with as they please.

This is, IC, a violation of SOP : only the acting captain is supposed to have captain level access. Not even the HOP is allowed to!

I, personally, dislike this current meta, which I've seen happen almost every round I am playing command or Blueshield (being handed the spare ID without even asking for it).

This would be less prevalent if the Blueshield had more access, as it would be less of a pain to give the BS any additional access they require instead of just AA, if they do require any.

This PR gives the Blueshield access to surgery and all command areas : Teleporter, AI upload, EVA, telecomms, gateway, vault and the AI satellite, with the exception of the actual head of staff offices.

While this is a technically a buff to Blueshield, please consider that nearly all Blueshields currently walk around with all access, so it isnt that large of one.

:cl:
balance: The Blueshield now has roundstart access to : Surgery, AI upload, EVA, Telecomms, Gateway, Teleporter, Vault, and the AI satellite.
/:cl:

### DIFF
--- a/code/game/jobs/job/supervisor.dm
+++ b/code/game/jobs/job/supervisor.dm
@@ -162,8 +162,9 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 0)
 			            access_crematorium, access_kitchen, access_cargo, access_cargo_bot, access_mailsorting, access_qm, access_hydroponics, access_lawyer,
 			            access_theatre, access_chapel_office, access_library, access_research, access_mining, access_heads_vault, access_mining_station,
 			            access_clown, access_mime, access_hop, access_RC_announce, access_keycard_auth, access_gateway, access_weapons, access_blueshield)
-	minimal_access = list(access_forensics_lockers, access_sec_doors, access_medical, access_construction, access_engine, access_maint_tunnels, access_research,
-			            access_RC_announce, access_keycard_auth, access_heads, access_blueshield, access_weapons)
+	minimal_access = list(access_forensics_lockers, access_sec_doors, access_medical, access_surgery, access_construction, access_engine, access_minisat, access_maint_tunnels, access_research,
+			            access_RC_announce, access_keycard_auth, access_heads, access_ai_upload, access_teleporter, access_eva, access_tcomsat, access_gateway, access_heads_vault,
+			            access_blueshield, access_weapons)
 	outfit = /datum/outfit/job/blueshield
 
 /datum/outfit/job/blueshield


TDSSS:
I see the reason for giving the BS access to surgery, but why do they need access to EVA, Gateway or Telecomms?

TheSardele:
Split on this, I'm fine with EVA (since every command member gets that anyway , and it can come in handy rescuing heads), but the others don't seem like they relate to the blueshield's duties. Arguably surgery, but while they are allowed to perform it as we don't restrict knowledge, it is discouraged to perform other department's duties.

datlo:
Surgery : Heads might need surgery work done. Performing surgery on heads fall within a Blueshield's duties.
EVA : Heads will sooner or later be spaced, or be kidnapped in space. I've often used EVA to go rescue them.
Telecomms : It's a command area that heads can find themselves in trouble in. Same reasoning for vault and AI upload.
Gateway : This one admittedly I'm a bit iffy on, but it's the only command area remaining and it felt out of place not to include it at this point. If a head decides to be a dumbass and go through the gateway, you should be able to follow them.

Cazdon:
EVA seems reasonable, I don't like the others, in particular surgery.

It's really getting into the whole 'doing other players jobs' territory. Blueshield should carefully watch over a head of staff while they're getting surgery in the operating theater, not be a bodyguard + personal surgeon combo.

It highlights the 'gamey' nature of us not restricting job knowledge, while simultaneously squashing the chances for interesting plays by potential antagonist surgeons. (which is also a problem I have with brig physician's existence) This happens right now anyway, but I don't think we should be encouraging it mechanically.

Quantum-M:
I disagree with most of those additional access. 

In terms of telecomms, the vault, and AI satellite, it's really on fault of the command member who goes to those areas without being accompanied by the Blueshield, as those areas are usually restricted to a small amount of crew (e.g. Captain, CE, HoP, RD).

The AI upload access will probably cause issues with Blueshields messing with the upload without authorization (even if there is a good reason to) and that usually command (specifically the Captain and RD) is supposed to keep an eye as a whole on whoever goes into it.

The teleporter usually has very few teleport beacons at round start (arrivals, teleport, bridge, science test chamber, AI satellite, telecomms, abandoned satellite) to even warrant a fast travel through the station.

If there is a concern about the whole meta of giving the Blueshield the spare ID/all access (which I think they shouldn't get it at all), then it's more of a player issue and should probably be brought up to the admins.

datlo:
Closing due to reasons above. I do think all access blueshields are an issue that needs eventual addressing.

PR #13006, opened 2020-02-19
Increases the defib timer back to 5 minutes

SteelSlayer:
## What Does This PR Do
Increases the defib timer back to 5 minutes, up from 2 minutes. This means you'll have 5 minutes after a patient dies before they are no longer able to be defib'd.

## Why It's Good For The Game
In my opinion, having a shorter defib timer does not make death more meaningful/impactful, it may in theory but it does not in practice. Currently, the timer is so short, that bodies often get dumped at medbay already unable to be defib'd. The bodies are then promptly whisked away to cloning, where the doctor presses 3 buttons and then sits around and waits for the body to be dumped out.

So what this actually does is place more importance on cloning, since often that is the only option.
I think it not only encourages doctors to be more lazy with the "just wait till they die and then clone" attitude. But also, for doctors who would like to be able to defib and put in the extra work on bringing patients back from that state often aren't able to get the chance, and are forced to clone instead.

I honestly don't think this will mess with the balance too much. The cloner is still there and it is a viable option. This just gives more room for the doctors to do actual doctor work of healing and fixing people, if they choose to.

One last note: I know that SR exists and can be seen as a sort of alternative to defibbing, but it's actually quite easy to make 90 pills of in the first 5-10 minutes of the round if you actually care to make it. This PR addresses the defib timer only.

## Changelog
:cl:
balance: Increases the the defib timer to 5 minutes, up from 2 minutes
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -345,7 +345,7 @@
 #define MOUSE_OPACITY_OPAQUE 2
 
 // Defib stats
-#define DEFIB_TIME_LIMIT 120
+#define DEFIB_TIME_LIMIT 300
 #define DEFIB_TIME_LOSS 60
 
 //different types of atom colorations


SagaStars:
I'm totally into this, very good argument.

Shockpoint:
I agree, just by watching medbay as an admin. Some people can die in, say, arrivals or something, and be out of defib time by the time they actually make it to medbay upon immediate retrieval. or someone's been killed in security and they're taking them to medbay a little late.

It certainly does not seem more impactful, just more of an obstacle and irritant for a medbay that is probably already drowning in bodies.

Pewtershmitz:
Please, yes! This is great! Just as It should be!

Quantum-M:
I strongly agree, as I think it would slightly push medical staff a bit to not default to cloning, especially if the person in question who died was too far/took too long to be transported to medbay.

One has to keep in mind that once someone is defibbed, Medical Doctors still have to deal with whatever other injuries they may had sustained that led to their death/while they were dead (e.g. broken bones, internal bleeding, toxins, burns, brute damage, septic/damaged organs).

Compared to cloning, which the only concerns are largely if there is a biomass, time spent cloning, any mutations/genetic damage you get, and whether or not cloning suddenly runs out of power and you end up being spit out without vital organs.

Carthusia:
I'm all for this, a clone should be a last resort, as for all intents and purposes, that person is dead, dead forever, all you've done is make a new "version" of them, a doctor should be doing all they can to SAVE that current "version" of the person, and when it comes to say the Vox or slime people, god help them.

SR is basically the only option unless they want a full body stack/brain transplant, which never sits well with vox or slime people (They never look the same) AND the SR kills most all of their internals almost immediately, nullifying the "They don't decay" aspect of vox and forcing a massive surgical operation to save them, putting them in a worse situation compared to where they started in the first place had they just been able to be defibbed at 2 minutes 5 seconds.

So an extended 5m for them is a god send in terms of reducing frustration.

Necaladun:
I think 5 minutes is still a reasonable yet short time. Cloning is still preferred if there's any extra damage, causing more work for med overall.

Birdtalon:
Anything which encourages revival and treatment of patients over just throwing them into the cloning room is a good change.

DanaDririon:
So, is this still in discussion or what? I would really like more time to defib the current clown that was beaten to death

Carthusia:
Would be nice to see this implemented ASAP if it approved

SteelSlayer:
Yeah still in discussion basically, waiting for a verdict from the the headmins and maints.

PR #13009, opened 2020-02-20
Fixes Blood Eat

Leanfrog:
## What Does This PR Do
This fixes a bug that prevented someone from obtaining the BLOODEAT power after consuming two demon hearts.

## Why It's Good For The Game
It fixes a bug that shouldn't exist in the first place

## Images of changes
N/A

## Changelog
:cl:
fix: fixed a bug with eating demon hearts where it would remove your currently existing demon heart setting your bloodcrawl value to 0
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/slaughter/slaughter.dm
+++ b/code/game/gamemodes/miniantags/slaughter/slaughter.dm
@@ -247,10 +247,7 @@
 	user.visible_message("<span class='warning'>[user] raises [src] to [user.p_their()] mouth and tears into it with [user.p_their()] teeth!</span>", \
 						 "<span class='danger'>An unnatural hunger consumes you. You raise [src] to your mouth and devour it!</span>")
 	playsound(user, 'sound/misc/demon_consume.ogg', 50, 1)
-	for(var/obj/effect/proc_holder/spell/knownspell in user.mind.spell_list)
-		if(knownspell.type == /obj/effect/proc_holder/spell/bloodcrawl)
-			qdel(src)
-			return
+
 
 	if(user.bloodcrawl == 0)
 		user.visible_message("<span class='warning'>[user]'s eyes flare a deep crimson!</span>", \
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/miniantags/slaughter/slaughter.dm
```diff
@@ -247,10 +247,7 @@
 	user.visible_message("<span class='warning'>[user] raises [src] to [user.p_their()] mouth and tears into it with [user.p_their()] teeth!</span>", \
 						 "<span class='danger'>An unnatural hunger consumes you. You raise [src] to your mouth and devour it!</span>")
 	playsound(user, 'sound/misc/demon_consume.ogg', 50, 1)
-	for(var/obj/effect/proc_holder/spell/knownspell in user.mind.spell_list)
-		if(knownspell.type == /obj/effect/proc_holder/spell/bloodcrawl)
-			qdel(src)
-			return
```
SteelSlayer: These lines shouldn't have been removed completely, only the return should have been. I tested this out, and without these lines, when someone eats a second heart to get `BLOODCRAWL_EAT`, the heart doesn't get deleted, and it instead gets dropped on the floor. Which means another person can pick up the heart and use it to get normal `BLOODCRAWL`.

code/game/gamemodes/miniantags/slaughter/slaughter.dm
```diff
@@ -259,10 +256,13 @@
 	else if(user.bloodcrawl == BLOODCRAWL)
 		to_chat(user, "You feel diffr-<span class = 'danger'> CONSUME THEM! </span>")
 		user.bloodcrawl = BLOODCRAWL_EAT
-	else
+	else if(user.bloodcrawl == BLOODCRAWL_EAT)
```
SteelSlayer: The if check is not needed here. There are only three options.
1. bloodcrawl == 0
2. bloodcrawl == 1 (`BLOODCRAWL`)
3. bloodcrawl == 2 (`BLOODCRAWL_EAT`)

If bloodcrawl does not equal 0 or 1, the only other option is 2, so a simple `else` works fine here.

Ty-Omaha: It makes it easier to understand imo, also else should be for 0 since else is usually called if things are false and 0 is attributed to things being false

code/game/gamemodes/miniantags/slaughter/slaughter.dm
Leanfrog: Alright, weirdly enough during my initial testing removal of these lines actually got some of the code actually working, I'll fix this then

code/game/gamemodes/miniantags/slaughter/slaughter.dm
SteelSlayer: Actually yeah, I like the clarity of having this here, so ignore what I said.

Fox-McCloud:
Please update.

Fox-McCloud:
23 days passed--no update. Very stale. Closing.

Issue #11635, opened 2019-06-01
Limb augmentation not augmenting with armour


MarsM0nd:
**Problem Description**:
Attempting to augment a limb with robot armour plating (/obj/item/robot_parts/robot_component/armour) doesn't augment.

**What did you expect to happen**: 
The limb to be augmented, or to told it's not the correct tool for this job.

**What happened instead**:
Surgery completes successfully, but besides a disappeared armour plating, nothing happens.

**Why is this bad/What are the consequences:**
You would figure you have an augmented limb/subdermal armor with no visual change, but are unchanged.

**Steps to reproduce the problem**:
Do a limb augmentation surgery with robot armour plating

**When did the problem start happening**:
Seems to be since the addition of it.

PR #13015, opened 2020-02-22
Surgery rework (continues surgery and tools refactor)

farie82:
Two test merges are done and the found issues were fixed. All ready to be merged if it's approved!

## What Does This PR Do
Makes surgery one continues workflow. Want to fix both IB and his lungs? You can do that. Be sure to fix his bones after though since cutting bones now breaks the bones.

Changes:
- The allowed tools per surgery step are now defines. This allows for a clear overview and for easier adding/changing steps
- Surgery now has less menu's and you can do multiple surgery types in one go.
- Removed the slime core surgery... why was this still in? (Was more effort to make it work than it is worth)
- Facial reconstruction now follows the same basic steps like all other surgeries. Incision, Clamp, Retractor, then the cutting. (Did this so I didn't have to add another surgery stage)
- Makes the healing surgery for IPCs (brute and burn) continue till all the damage is healed or once the surgery conditions are no longer valid
- The early stop (cautery + scalpel etc) has been removed. Now you can just use a cautery on organics. And the logical thing on IPCs. Just screwed him open? Screw again to stop etc
- Surgically cutting bones now breaks said bones.
- The clean step in organ manipulation now allows you to inject mitocholide into the organs. Every organ should get 3 units of mito to be revived.
- The clean surgery step now doesn't waste chems if they won't do anything.
- The debridement mito step now in addition to reviving the limb also set the germ level to 0. This should keep the limb alive for a bit longer
- Updated the Surgery computer to also show the active surgeries and their information
- Makes the abductor surgery's last step heal your bones, crit symtoms and resp damage. This is to counter the bones breaking in this PR and the "recent" new crit changes
- One less step for IPC abduction surgery. No more useless step in between to "prepare" the chest
- robot_parts that don't have an organ attached now won't start a surgery and instead give a fitting message.
- You can't heal the organs in the head through the eyes/mouth using organ manip anymore. Need to crack the head open for this.
- Using a surgery tool (or hand) again when doing a step will now give a warning and you won't attack the patient on accident.
- If the targets surgery stage is after the retractor stage then you can't heal them using brute kits etc. Used to be only on the retractor step and equal steps. Having cut open ribs didn't prevent you from healing them
- The check if you're on a bed/table/surgery table is simplified. It now just checks if you're lying instead of checking every stat that leads to you lying. Fixes some weird cases where a mob being buckled on a chair above a bed is surgery viable

The user will be asked which step to do in case of multiple possible. This does bring the issue of the IMS having tons of menu's... still thinking about a way to fix this. IPCs also always have the cavity implant surgery available when you have their hatch open.

I've given each step a priority. This is used to determine which step should be placed higher in the list of options. The standard steps have a pretty high priority while for instance cauterizing the wound has a low priority.

Tested in most situations using the happy flow and a few unhappy flows.

## Why It's Good For The Game
Makes surgery a more dynamic flow and the refactor is just nice.

## Images of changes
ESXA5klo2Y
Organic surgery. (Fix-Ovein didn't work due to the patient not having IB in his chest)

image

Q18tk5Ds9O
IPC surgery. Fixing the brute damage

image
Surgery computer

Fixes:
```
[2020-01-30T04:19:59] Runtime in action.dm,116: Cannot read null.actions
[2020-01-30T04:19:59]   proc name: New (/datum/action/item_action/New)
[2020-01-30T04:19:59]   usr: Hazel Ray (CKEY) (/mob/living/carbon/human)
[2020-01-30T04:19:59]   usr.loc: The floor (146,107,1) (/turf/simulated/floor/plasteel)
[2020-01-30T04:19:59]   src: Activate Pill (/datum/action/item_action/hands_free/activate_pill)
[2020-01-30T04:19:59]   call stack:
[2020-01-30T04:19:59]   Activate Pill (/datum/action/item_action/hands_free/activate_pill): New(null, null, null)
[2020-01-30T04:19:59]   insert pill (/datum/surgery_step/insert_pill): end step(Hazel Ray (/mob/living/carbon/human), Vixx Vaxx (/mob/living/carbon/human), "mouth", Water (50u) pill (/obj/item/reagent_containers/food/pill), dental implant (/datum/surgery/dental_implant))
[2020-01-30T04:19:59]   insert pill (/datum/surgery_step/insert_pill): initiate(Hazel Ray (/mob/living/carbon/human), Vixx Vaxx (/mob/living/carbon/human), "mouth", Water (50u) pill (/obj/item/reagent_containers/food/pill), dental implant (/datum/surgery/dental_implant))
[2020-01-30T04:19:59]   insert pill (/datum/surgery_step/insert_pill): try op(Hazel Ray (/mob/living/carbon/human), Vixx Vaxx (/mob/living/carbon/human), "mouth", Water (50u) pill (/obj/item/reagent_containers/food/pill), dental implant (/datum/surgery/dental_implant))
[2020-01-30T04:19:59]   dental implant (/datum/surgery/dental_implant): next step(Hazel Ray (/mob/living/carbon/human), Vixx Vaxx (/mob/living/carbon/human))
[2020-01-30T04:19:59]   Vixx Vaxx (/mob/living/carbon/human): attackby(Water (50u) pill (/obj/item/reagent_containers/food/pill), Hazel Ray (/mob/living/carbon/human), "icon-x=15;icon-y=15;vis-x=16;v...")
[2020-01-30T04:19:59]   Water (50u) pill (/obj/item/reagent_containers/food/pill): melee attack chain(Hazel Ray (/mob/living/carbon/human), Vixx Vaxx (/mob/living/carbon/human), "icon-x=15;icon-y=15;vis-x=16;v...")
[2020-01-30T04:19:59]   Hazel Ray (/mob/living/carbon/human): ClickOn(Vixx Vaxx (/mob/living/carbon/human), "icon-x=15;icon-y=15;vis-x=16;v...")
[2020-01-30T04:19:59]   Vixx Vaxx (/mob/living/carbon/human): Click(the floor (146,106,1) (/turf/simulated/floor/plasteel), "mapwindow.map", "icon-x=15;icon-y=15;vis-x=16;v...")
```

Fixes: #11635
Fixes: #16984

## Changelog
:cl:
del: Slime core surgery. Good bye old friend
tweak: Surgery is a continues workflow now. Aka you can do multiple of the old surgery types at once now. IB and bones? Sure go ahead
tweak: Facial reconstruction now starts with the standard 3 steps. Incision, clamp, retractor then the face cutting
tweak: The clean surgery step in organ manipulation now accepts mito as a chemical. You need to inject atleast 3u of pure mito into each dead organ to revive it.
tweak: The clean surgery step now doesn't waste chems if they won't do anything.
tweak: The debridement mito step now also cleans the limb. So no instant dead limbs again
balance: Sawing a bone in the bone saw surgery step now breaks the bone
fix: Fixes a runtime in dental pill implants
tweak: The surgery computer now shows the active surgeries and their information
tweak: The IPC abduction surgery now has one less step. The one before you put the gland in is removed since it was useless
fix: using robot_parts that have no organ attached now won't start the surgery
tweak: You can't heal the organs in the head anymore through the eyes/mouth organ manip surgery
tweak: Using a surgery tool (or hand) again when doing a step will now give a warning and you won't attack the patient on accident.
tweak: You can't heal limbs using healing packs etc if those limbs are at or after the retractor step. Used to be only on the retractor step
fix: Being buckled on a chair above a bed now won't make you viable for surgery
/:cl:


code/game/machinery/computer/Operating.dm
```diff
@@ -120,9 +120,8 @@
 		if(occupant.surgeries.len)
 			occupantData["inSurgery"] = 1
 			for(var/datum/surgery/procedure in occupant.surgeries)
-				occupantData["surgeryName"] = "[capitalize(procedure.name)]"
-				var/datum/surgery_step/surgery_step = procedure.get_surgery_step()
-				occupantData["stepName"] = "[capitalize(surgery_step.name)]"
+				occupantData["surgeryLocation"] = "[capitalize(procedure.location)]"
```
farie82: Needs a method to make the location into a readable string. Instead of L_arm

code/modules/mob/living/carbon/alien/special/alien_embryo.dm
```diff
@@ -57,8 +57,8 @@
 			RefreshInfectionImage()
 
 	if(stage == 5 && prob(50))
-		for(var/datum/surgery/S in owner.surgeries)
-			if(S.location == "chest" && istype(S.get_surgery_step(), /datum/surgery_step/internal/manipulate_organs))
+		for(var/datum/surgery/S in owner.surgeries)// TODO
+			if(S.location == "chest" && S.current_stage == SURGERY_STAGE_OPEN_INCISION_BONES)
```
farie82: Will extend upon this later

code/modules/surgery/generic.dm
```diff
@@ -140,104 +125,40 @@
 		self_msg = "<span class='warning'> Your hand slips, damaging several organs [target]'s lower abdomen with \the [tool]!</span>"
 	user.visible_message(msg, self_msg)
 	target.apply_damage(12, BRUTE, affected, sharp = 1)
-	return 0
+	return FALSE
 
 /datum/surgery_step/generic/cauterize
-
+	priority = -1 // Always on the bottom of possibilities
```
farie82: Todo actually use the priority value. Maybe to fix the available steps issue with say IMS

code/modules/surgery/limb_augmentation.dm
```diff
@@ -1,38 +1,36 @@
...
+	if(affected.is_robotic())
+		return FALSE
+	
 	var/obj/item/robot_parts/p = tool
 	if(p.part)
 		if(!(target_zone in p.part))
-			to_chat(user, "<span class='warning'>[tool] cannot be used to augment this limb!</span>")
-			return 0
-	return 1
+			//to_chat(user, "<span class='warning'>[tool] cannot be used to augment this limb!</span>")
```
farie82: can_use procs are now used to determine which surgery step is available so no to_chats here

code/_onclick/item_attack.dm
```diff
@@ -1,5 +1,5 @@
 /obj/item/proc/melee_attack_chain(mob/user, atom/target, params)
-	if(!tool_attack_chain(user, target) && pre_attackby(target, user, params))
+	if(!tool_attack_chain(user, target) && !surgery_attack_chain(user, target) && pre_attackby(target, user, params))
```
Citinited: I've not looked at any of the rest of the code, but won't the order of this line interfere with things like welding tools, which can be used both as a tool and as a surgery tool?

You might want to stick the surgery attack chain first if it does

code/modules/mob/living/living_defense.dm
```diff
@@ -281,6 +281,12 @@
 
 	return G
 
+/mob/living/proc/surgery_act(mob/living/user, obj/item/I, datum/surgery/S)
```
Citinited: Does this mean we can do surgery on borgs? 😄

code/_onclick/item_attack.dm
```diff
@@ -12,6 +12,18 @@
 		return FALSE
 	return target.tool_act(user, src, tool_behaviour)
 
+
+/obj/item/proc/surgery_attack_chain(mob/user, atom/target)
```
Citinited: Given this is `mob/living` level stuff you might want to reflect that in these args

code/modules/surgery/surgery.dm
```diff
@@ -27,19 +18,47 @@
...
-			return 1
-	return 0
-
-/datum/surgery/proc/get_surgery_step()
-	var/step_type = steps[status]
-	return new step_type
-
+		var/result
+		. = TRUE
+		do
```
Citinited: Don't think I've ever seen a `do...while` in this codebase before 🤔

AffectedArc07: I avoid them if I can, seems like a weird operation to me

farie82: It is used in some places. I can hack it out and set result to SURGERY_CONTINUE
But Id rather have it like this. It will always do the loop once. Then it'll act like a normal while loop

code/_onclick/item_attack.dm
farie82: It will interfere if the target (the mob in case of surgery) also has a tool_act. I hope this isn't ever the case though :P

code/modules/mob/living/living_defense.dm
farie82: image

All surgery steps expect atleast a carbon now. It's a mess and I'll have to clean that up though

code/_onclick/item_attack.dm
farie82: Good point. Will change

code/_onclick/item_attack.dm
farie82: Changed it just in case. Thanks for the tip!

SteelSlayer:
Cool stuff, I like the ability to multisurgeries. I tested out nearly every organic surgery, no bugs or runtimes so far. Here are my findings/feedback:

Thoughts:
1. After you use retractors on someone, I feel like the surgery computers should give better feedback on this step. It currently says `Current stage: Open incision`, but I was confusing this for having just used a scalpel to make an incision. Could have this be `Skin retracted` or something.
2. Similar to the above, `Current stage: Open incision bones` could be made more clear. Could have it be `Bones retracted` instead.
3. I sometimes got confused while doing surgery, mostly trying to keep track of what step I was on and was I was trying accomplish next. Idea: give the surgery computer a list of "Possible next steps", for people who like to follow along in the computer. I realize this could be a pain to code in though. Maybe its just a git gud situation for us.

Bugs:
1. The debridement surgery seemed to not work sometimes. I tried to use a mitocholide beaker on someone's dead limb, the limb appeared to get healed but didn't. The limb's color went from green and dead looking, to healed, but then 1 second later it went back to green and dead. Not sure if I'm doing something wrong there but thought I should mention it.

farie82:
> Thoughts:
> 
> 1. After you use retractors on someone, I feel like the surgery computers should give better feedback on this step. It currently says `Current stage: Open incision`, but I was confusing this for having just used a scalpel to make an incision. Could have this be `Skin retracted` or something.
> 2. Similar to the above, `Current stage: Open incision bones` could be made more clear. Could have it be `Bones retracted` instead.

Good points. I've not spend to much time on those names in all honesty. I'll change them to something more sensible and easier to understand

> 3. I sometimes got confused while doing surgery, mostly trying to keep track of what step I was on and was I was trying accomplish next. Idea: give the surgery computer a list of "Possible next steps", for people who like to follow along in the computer. I realize this could be a pain to code in though. Maybe its just a git gud situation for us.

It is certainly doable and will most likely help people switch between the old and new system. Good idea

> Bugs:
> 
> 1. The debridement surgery seemed to not work sometimes. I tried to use a mitocholide beaker on someone's dead limb, the limb appeared to get healed but didn't. The limb's color went from green and dead looking, to healed, but then 1 second later it went back to green and dead. Not sure if I'm doing something wrong there but thought I should mention it.

I think this is an old bug from what I heard from others. I will however fix it.

Thanks for the testing and feedback!

code/__DEFINES/surgery_stages.dm
```diff
@@ -0,0 +1,29 @@
...
+
+#define SURGERY_STAGE_BONES_GELLED			"bones gelled"
+#define SURGERY_STAGE_BONES_SET				"bones set"
+
+#define SURGERY_STAGE_ROBOTIC_HATCH_UNLOCKED	"hatch unlocked"
+#define SURGERY_STAGE_ROBOTIC_HATCH_OPEN		"hatch open"
+
+#define SURGERY_STAGE_CARAPACE_SAWN				"carapace sawn open"
+#define SURGERY_STAGE_CARAPACE_CUT				"carapace cut"
+#define SURGERY_STAGE_CARAPACE_OPEN				"carapace open"
```
farie82: If people have better names for the stages then please let me know! They will be used by the surgery computer so a short descriptive name would be welcome

SteelSlayer:
Feedback from more testing:

Nitpick but I'm was doing surgery on the arm here and it says "heal *organs*", "clean *organs*" etc. but the arms and legs do not have organs. Basically the code is referring to implants as "organs". Not sure if you can fix that without spaghetti.
A

An issue also related to arm surgery: in the picture above you can see the step labeled "cut tissue" which uses a scalpel. I used a scalpel at this step but the surgery computer did not update the current stage or possible steps, it stayed on "skin retracted" and kept the "cut tissue" in the list of possible steps.

<-------------------------------------------------------------------------->

I also found another bizzare bug. While messing around with organic and IPC surgery, I was operating on this skrell. I closed the computer window and dragged the skrell off the operating table. The skrell was on the step "bones sawn". I then dragged the IPC onto the table which also had his chest hatch open. I then opened the computer window again and this is what I saw:

B

The same happens in reverse. Somehow the surgery computer was storing the previous surgery current step and possible steps. You can avoid this by **not** closing the surgery window when you transfer the patients and instead click the computer to get the "no patient detected" text on the screen, and then transfer them. This will make the surgeries properly update.

This bug also occurs between 2 organic patients, provided they both have ongoing surgeries (the computer was telling me to saw eyes open for example). This is a pretty niche/edge case type of thing here, sorry to put you through this :smile:. Let me know if the explanation doesn't make sense.

farie82:
> Nitpick but I was doing surgery on the arm here and it says "heal _organs_", "clean _organs_" etc. but the arms and legs do not have organs. Basically the code is referring to implants as "organs". Not sure if you can fix that without spaghetti.

It's due to how the surgery step is currently named. I can't think of a better name in all honesty so if others have suggestions please do let me know!

> A
> An issue also related to arm surgery: in the picture above you can see the step labeled "cut tissue" which uses a scalpel. I used a scalpel at this step but the surgery computer did not update the current stage or possible steps, it stayed on "skin retracted" and kept the "cut tissue" in the list of possible steps.

Nice find. I'll look into it once I can. Did the step actually start btw? Like did you have to wait and get messages?

> I also found another bizzare bug. While messing around with organic and IPC surgery, I was operating on this skrell. I closed the computer window and dragged the skrell off the operating table. The skrell was on the step "bones sawn". I then dragged the IPC onto the table which also had his chest hatch open. I then opened the computer window again and this is what I saw:
> 
> B
> 
> The same happens in reverse. Somehow the surgery computer was storing the previous surgery current step and possible steps. You can avoid this by **not** closing the surgery window when you transfer the patients and instead click the computer to get the "no patient detected" text on the screen, and then transfer them. This will make the surgeries properly update.
> 
> This bug also occurs between 2 organic patients, provided they both have ongoing surgeries (the computer was telling me to saw eyes open for example). This is a pretty niche/edge case type of thing here, sorry to put you through this 😄. Let me know if the explanation doesn't make sense.

I think I know why this is. Will fix it once able.

Thanks again for the testing! This stuff really helps find and fix the edge cases!

SteelSlayer:
@farie82 Yes I was able to actually use the scalpel and have a progress bar for that cut tissue step, I believe it's just the computer that didn't update.

farie82:
> @farie82 Yes I was able to actually use the scalpel and have a progress bar for that cut tissue step, I believe it's just the computer that didn't update.

I could and I couldn't reproduce the bug. I think this is due to how the UI stuff works. Sometimes a window loses focus and then it stops auto updating.
I know I had issues with this while doing virology as well. Clicking on the machine again will fix the issue

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
```diff
@@ -1,40 +1,33 @@
...
-		return TRUE
-	return FALSE
+/datum/surgery_step/internal/manipulate_organs/abduct/is_valid_target(mob/living/carbon/human/target)
+	return ishuman(target)
+
+/datum/surgery_step/internal/manipulate_organs/abduct/is_valid_user(mob/living/carbon/human/user)
+	return ishuman(user) && (isabductor(user) || (locate(/obj/item/implant/abductor) in user))
 
-/datum/surgery_step/internal/extract_organ
+/datum/surgery_step/internal/manipulate_organs/abduct/extract_organ
```
farie82: Todo make work with augmented chest etc. Damn you fake IPCs

code/modules/paperwork/pen.dm
```diff
@@ -23,6 +23,7 @@
 	materials = list(MAT_METAL=10)
 	var/colour = "black"	//what colour the ink is!
 	pressure_resistance = 2
+	surgery_behaviours = list(SURGERY_MAKE_CAVITY = 90)
```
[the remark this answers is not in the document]
farie82: Changed in the newest commit.
I've not yet included the other changes we talked about like keeping track of the amount of times you were on the same stage (for surgeries like TG has which use the same step multiple times in a row). I still need to think about this how I want to implement this. Currently this is possible by making a new surgery step which leads to a newly defined surgery state.

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
```diff
@@ -1,40 +1,36 @@
-/datum/surgery/organ_extraction
-	name = "Experimental Dissection"
-	steps = list(/datum/surgery_step/generic/cut_open, /datum/surgery_step/generic/clamp_bleeders, /datum/surgery_step/generic/retract_skin, /datum/surgery_step/open_encased/saw, /datum/surgery_step/open_encased/retract, /datum/surgery_step/internal/extract_organ, /datum/surgery_step/internal/gland_insert, /datum/surgery_step/generic/cauterize)
-	possible_locs = list("chest")
+/datum/surgery_step/internal/manipulate_organs/abduct
```
farie82: Test

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
```diff
@@ -1,40 +1,36 @@
...
-/datum/surgery_step/internal/extract_organ
+/datum/surgery_step/internal/manipulate_organs/abduct/extract_organ
 	name = "remove heart"
-	accept_hand = 1
-	time = 32
+	accept_hand = TRUE
+	surgery_start_stage = list(SURGERY_STAGE_OPEN_INCISION, SURGERY_STAGE_OPEN_INCISION_BONES, SURGERY_STAGE_ROBOTIC_HATCH_OPEN)
+	next_surgery_stage = SURGERY_STAGE_SAME
+	allowed_surgery_tools = SURGERY_TOOLS_EXTRACT_ORGAN
 	var/obj/item/organ/internal/IC = null
```
Citinited: I don't know how much work it would be but IC is such a vague var name it might be worth renaming

code/modules/surgery/surgery.dm
```diff
@@ -206,19 +246,3 @@
...
 
 	E.germ_level += germs
-
-/proc/sort_surgeries()
-	var/gap = GLOB.surgery_steps.len
-	var/swapped = 1
-	while(gap > 1 || swapped)
-		swapped = 0
-		if(gap > 1)
-			gap = round(gap / 1.247330950103979)
```
Citinited: How oddly specific 😂

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
```diff
@@ -1,100 +1,83 @@
...
+	for(var/obj/item/I in target.get_organs_zone(target_zone))
 		// Allows for multiple subtypes of heart.
 		if(istype(I, /obj/item/organ/internal/heart))
-			IC = I
+			foundHeart = I
 			break
-	user.visible_message("[user] starts to remove [target]'s organs.", "<span class='notice'>You start to remove [target]'s organs...</span>")
+	if(!foundHeart)
+		to_chat(user, "<span class='warning'>You can't seem to find an organ that functions like a heart in [target].</span>")
+		return -1
```
Citinited: I think it would be better to use a define in these return values instead of a magic number

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
farie82: Just a few clicks and a bit of creative thinking *coughs* and it's done! Thanks. Kinda of looked over the abductor surgery for the most part

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
farie82: Will change it. There are some inconsistencies with this as well so it's needed.

code/modules/surgery/cavity_implant.dm
```diff
@@ -154,13 +152,12 @@
 			user.drop_item()
 			affected.hidden = tool
 			tool.forceMove(affected)
-			return TRUE
 	else
 		if(IC)
 			user.visible_message("[user] pulls [IC] out of [target]'s [target_zone]!", "<span class='notice'>You pull [IC] out of [target]'s [target_zone].</span>")
 			user.put_in_hands(IC)
 			affected.hidden = null
-			return FALSE
```
farie82: In the old system it returned FALSE to stop the stage from advancing. Not required now. This is a total success. Same for the others that used to return FALSE below

code/modules/surgery/surgery.dm
farie82: Somebody put a lot of effort into that number. And it was not even used in the previous code

Crazylemon64:
It seems that splinted limbs can't be bone-repaired...?

Mitchs98:
Few bugs I encountered, overall it seems very well done.
- Scanner will sometimes show an open incision that doesn't exist, leading to you stabbing/cauterying people repeatedly
- Organ healing surgery wastes the mitocholide even if it doesn't need treatment, probably shouldn't do that.
- Using a bone setter on the skull gives an option to mend the skull or set the skull, seems redundant. After using set the skull the bone gel tells you it both mends and heals the bones?
- Mending with bone gel on the final step sends two messages.
- Sometimes when using the same tool on multiple limbs it can fail to properly do the step on one or part of the limbs you attempt it on.

The mito step thing is more something that should be in rather than a bug, I'd assume. Makes no sense to waste the mito on organs that aren't damaged, should use the same checks that the brute trauma kit does and notice there's no damage BEFORE you inject the mito imo.

Necaladun:
Need to test some of these, but it all LGTM in theory

farie82:
> * Scanner will sometimes show an open incision that doesn't exist, leading to you stabbing/cauterying people repeatedly

Check

> * Organ healing surgery wastes the mitocholide even if it doesn't need treatment, probably shouldn't do that.

Can't find the source in the code so will have to test.

> * Using a bone setter on the skull gives an option to mend the skull or set the skull, seems redundant. After using set the skull the bone gel tells you it both mends and heals the bones?

Oh whoops. Fixed

> * Mending with bone gel on the final step sends two messages.

Check

> * Sometimes when using the same tool on multiple limbs it can fail to properly do the step on one or part of the limbs you attempt it on.

In what sense do you mean this? Like it fails the step and gives you the message?

farie82:
> It seems that splinted limbs can't be bone-repaired...?

Unable to reproduce this behaviour. I was able to break somebodies bone. Splint it and repair it while the splint was still on

code/modules/surgery/other.dm
```diff
@@ -2,29 +2,24 @@
 //////////////////////////////////////////////////////////////////
 //					INTERNAL WOUND PATCHING						//
 //////////////////////////////////////////////////////////////////
-/datum/surgery_step/debridement
```
farie82: Was the exact same as cauterize but didn't work...

code/__DEFINES/surgery_tools.dm
```diff
@@ -0,0 +1,103 @@
...
+
+#define SURGERY_TOOLS_ROBOTIC_REPAIR	list(/obj/item/stack/cable_coil = 100, \
+											/obj/item/weldingtool = 100, \
+											/obj/item/gun/energy/plasmacutter = 50)
+
+#define SURGERY_TOOLS_ROBOTIC_REMOVE_ORGAN	list(/obj/item/multitool = 100)
+
+#define SURGERY_TOOLS_ROBOTIC_MEND_ORGAN	list(/obj/item/stack/nanopaste = 100,/obj/item/bonegel = 30, /obj/item/screwdriver = 70)
+
+#define SURGERY_TOOLS_ROBOTIC_REPORGRAM		list(/obj/item/multitool = 100)
```
Citinited: Typo `REPORGRAM`

code/__DEFINES/surgery_tools.dm
```diff
@@ -0,0 +1,103 @@
...
+#define SURGERY_TOOLS_RETRACT_SKIN 	list(/obj/item/scalpel/laser/manager = 100, \
+										/obj/item/retractor = 100, 	\
+										/obj/item/crowbar = 90,	\
+										/obj/item/kitchen/utensil/fork = 60)
+
+#define SURGERY_TOOLS_CAUTERIZE		list(/obj/item/scalpel/laser = 100, \
+										/obj/item/cautery = 100,			\
+										/obj/item/clothing/mask/cigarette = 90,	\
+										/obj/item/lighter = 60,			\
+										/obj/item/weldingtool = 30)
```
Citinited: There doesn't seem to be any checks done here to ensure the tool is actually working (e.g. a welder will always act as a cautery, regardless of if it's lit). It might be an idea in `/datum/surgery_step/proc/initiate` to include a `if(!item.tool_enabled)` to fail if the tool isn't enabled

It'll probably need a bit of work for cigarettes and the like to not work as surgical tools if unlit, but that can come at a later date and this tweak shouldn't impact on any surgical behaviours at present

farie82: Placed it in the initiate.
A later PR can use tool_enabled for cigs etc as well. Instead of the lit var they use there. Out of the scope of this one though.
Thanks for the remark! Fixes another case as well so nice

Mitchs98:
> This should not occur if the organ is not damaged/dead. It will still do it if say the organ has 1 point of damage and is not dead.
> image

Hmm. Definitely odd. Because when I injected it in someones chest(Because I thought the liver was there briefly I derped lol) it injected it into EVERY organ in the chest and used the mito. It had a blue message saying something...I forget what though >_<. At-least 80% sure their chest organs weren't damaged. Did the same thing when I did finally fix their liver.

farie82:
Second TM done.
One bug found: Using a scalpel on the open incision step, then a drill and then the cautery to close the cavity would bypass the ribs/skull and lead you to the open bones stage.
Fix was to add a missing check to the cavity surgery (drill).

No other issues found luckily!

Aquilar:
An idea for the IMS, 
- Act as scalpel, hemostat and retractor on the opening (ribcage movement/skin replacement too)
- Act as scalpel for the rest of the operation
- Act as cautery only on the cauterise step
This will in turn,
- Require a hemostat for organ manipulation since the IMS can not do it
- Require a cautery if the operation needs to be cancelled since the IMS can not do it
However this will result in a lot less menus popping up constantly asking you what to do and would keep it more in line with it's current limitations.

farie82:
Quick close and reopen to give travis a reason to build again

farie82:
> An idea for the IMS,
> 
> * Act as scalpel, hemostat and retractor on the opening (ribcage movement/skin replacement too)
> * Act as scalpel for the rest of the operation
> * Act as cautery only on the cauterise step
> 
> This will in turn,
> 
> * Require a hemostat for organ manipulation since the IMS can not do it
> * Require a cautery if the operation needs to be cancelled since the IMS can not do it
> 
> However this will result in a lot less menus popping up constantly asking you what to do and would keep it more in line with it's current limitations.

The organ manipulation part the IMS can already not do. I'll think about the cautery part though. Will solve most of the popups

Aquilar:
this can also go for the laser scalpel to, since it also can cauterise and it should do that only at the cauterise step as well

farie82:
I think I'm gonna leave it as is. I've attempted a few ways of coding this. And they all resulted in more or less mostly snowflake code. I've put the cautery step as priority -1 so it'll always be at the bottom of the list. Experienced surgeons will find out that they can just spam enter if they use the IMS.
Another PR can be made if it really is an issue. Or even another PR can be made to make the choice pop up a radial menu or something like that. So the user is not pulled out of the main game window

code/modules/surgery/rig_removal.dm
```diff
@@ -1,59 +1,45 @@
 //Procedures in this file: Unsealing a Rig.
```
SteelSlayer, line 1: I think this entire file will be deleted when you merge upstream, because of the rig removal.

code/modules/surgery/surgery.dm
```diff
@@ -1,173 +1,217 @@
...
-	var/datum/surgery_step/S = get_surgery_step()
+	var/current_stage = SURGERY_STAGE_START
+	var/step_in_progress = FALSE
+	var/location = "chest"						//Surgery location
+	var/list/surgery_steps_history = list()		// For debugging and maybe later expansions that require previous surgeries
+
+/datum/surgery/Destroy(force, ...)
+	QDEL_LIST(surgery_steps_history)
+	surgery_steps_history = null
+	. = ..()
```
SteelSlayer, line 11: Probably better to do
```suggestion
	return ..()
```

code/modules/surgery/surgery.dm
```diff
@@ -1,173 +1,217 @@
...
+
+/datum/surgery/proc/next_step(mob/living/user, mob/living/carbon/target, obj/item/tool)
+	if(step_in_progress)	return FALSE
+	. = TRUE // Person
+	var/list/steps = get_surgery_steps(user, target, tool)
+	var/datum/surgery_step/S
+	if(steps.len == 0)
```
SteelSlayer, line 18: Length over len, won't mention it again for this file.
```suggestion
	if(!length(steps))
```

code/modules/surgery/surgery.dm
```diff
@@ -1,173 +1,217 @@
...
+		if(S.possible_locs?.len && !(location in S.possible_locs))
+			continue
+		if(S.is_valid_target(target) && (isobserver(user) || S.is_valid_user(user)) && S.is_zone_valid(target, location, current_stage))
+			possible_steps.Add(S)
+
+	return sortInsert(possible_steps, /proc/compare_surgery_steps)
 
+/* SURGERY STEPS */
+/datum/surgery_step
+	var/name			// Don't forget to name actual steps and make sure the name is unique. This will be used in the selecting logic
```
SteelSlayer, line 71: `///` for autodoc :smirk:

SteelSlayer:
Did some more testing of the core surgeries. Some more findings/questions: 

1) On the "Tissue cut" step. You're suppose to be allowed to use any tool from the `SURGERY_TOOLS_CLEAN_ORGAN` list. So for example, I'm using a bottle of water as a test, but it doesn't pour the bottle over the targeted area. I instead try to feed the bottle's contents to the mob.
2) While trying to use a brute kit on someone's eyes, I got: `Quumqooqixqix Quum*'s eyes does not seem to have any organs.` I'm fairly sure you can heal eyes with a brute kit on live, although I could be wrong.
3) While doing eye surgery or mouth surgery, after the "Retract skin" step, you have the option to saw their eye/mouth... bones :smile:. And this is what you end up with:
m5jaO5alAZ
This actually causes the **head** to get a fracture, which is quite interesting.

farie82:
> 1. On the "Tissue cut" step. You're suppose to be allowed to use any tool from the `SURGERY_TOOLS_CLEAN_ORGAN` list. So for example, I'm using a bottle of water as a test, but it doesn't pour the bottle over the targeted area. I instead try to feed the bottle's contents to the mob.

Yep that sounds like a bug. Will test and fix!

> 2. While trying to use a brute kit on someone's eyes, I got: `Quumqooqixqix Quum*'s eyes does not seem to have any organs.` I'm fairly sure you can heal eyes with a brute kit on live, although I could be wrong.

On live you can indeed do it... and more
image
I'll see about fixing/changing that

> 3. While doing eye surgery or mouth surgery, after the "Retract skin" step, you have the option to saw their eye/mouth... bones 😄. And this is what you end up with:
>    m5jaO5alAZ
>    This actually causes the **head** to get a fracture, which is quite interesting.

Also sounds like a bug yeah. Unless they got encased eyes which I doubt 😁 

Thanks for testing! Helps a bunch

Aquilar:
do you think you could add Toxin/Burn/Trauma would care as a set of surgeries? or should that be in a separate PR?

farie82:
> do you think you could add Toxin/Burn/Trauma would care as a set of surgeries? or should that be in a separate PR?

It is something I plan on doing once this is merged yeah. I also want to add more ways to revive people instead of just the SR, clone way. But I'll all do it in separate PRs so that this one will eventually get merged :D

moxian (approved):
Didn't read the individual surgeries too close, given that they have retained most of their behavior from the existing code.
The framework looks good.
My only real concern is that clicking people with the surgery tool twice in a row stabs them instead of saying "hey, you're already doing that, wait a moment" - the rest of the comments are nitpicks.

Good work!

code/modules/surgery/surgery.dm
```diff
@@ -1,173 +1,232 @@
...
 
 // does stuff to begin the step, usually just printing messages. Moved germs transfering and bloodying here too
-/datum/surgery_step/proc/begin_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool,datum/surgery/surgery)
+/datum/surgery_step/proc/begin_step(mob/living/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	if(ishuman(target))
 		var/obj/item/organ/external/affected = target.get_organ(target_zone)
 		if(can_infect && affected)
 			spread_germs_to_organ(affected, user, tool)
-	if(ishuman(user) && !(istype(target,/mob/living/carbon/alien)) && prob(60))
+	if(blood_level && ishuman(user) && !(istype(target,/mob/living/carbon/alien)) && prob(60))
```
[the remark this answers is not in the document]
farie82: I'm gonna keep it 60 here. It's a random chance that is not used more often

tgui/packages/tgui/interfaces/OperatingComputer.js
```diff
@@ -144,11 +144,24 @@ const OperatingComputerPatient = (props, context) => {
...
-            <LabeledList.Item label="Procedure">
-              {occupant.surgeryName}
+            <LabeledList.Item label="Ongoing Surgeries">
+              {occupant.surgeries.map((a, i) => (
+                <Button
+                  key={i}
+                  selected={a.location === occupant.selected_surgery.location}
+                  icon="arrow-down"
+                  content={a.location}
+                  onClick={() => act('change_surgery', {
```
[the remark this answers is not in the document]
farie82: I'm gonna keep that for a later PR. I ain't to great with TGUI yet so I'll keep it open for somebody else to improve

code/modules/surgery/surgery.dm
```diff
@@ -1,173 +1,232 @@
...
-	// put special restrictions here
-	return 1
-
-
-/datum/surgery/proc/next_step(mob/user, mob/living/carbon/target)
-	if(step_in_progress)	return
-
-	var/datum/surgery_step/S = get_surgery_step()
+	var/current_stage = SURGERY_STAGE_START
+	var/step_in_progress = FALSE
```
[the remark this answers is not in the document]
farie82: Gonna refrain from adding that now. Adding a reference to a mob is something you have to think about since having a reference means that you also need to dereference it. Else the referenced mob won't properly GC.
Would be something for the operation computer improvement though

code/modules/mob/living/living_defense.dm
```diff
@@ -280,6 +280,30 @@
...
+		return FALSE
+	if(can_operate(src))
+		var/datum/surgery/S = get_or_initiate_surgery(src, user)
+		if(S)
+			return surgery_act(user, I, S)
+	return FALSE
+
+/mob/living/proc/surgery_act(mob/living/user, obj/item/I, datum/surgery/S)
+	if(S.step_in_progress)
+		return FALSE
```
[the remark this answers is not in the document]
farie82: Good suggestion. Changed in the latest commit

farie82:
> Didn't read the individual surgeries too close, given that they have retained most of their behavior from the existing code.
> The framework looks good.
> My only real concern is that clicking people with the surgery tool twice in a row stabs them instead of saying "hey, you're already doing that, wait a moment" - the rest of the comments are nitpicks.
> 
> Good work!

Thanks for the review!
Latest commit has all the proposed changes!

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+/datum/surgery_step/robotics/manipulate_robotic_organs/extract/begin_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
 
-	else if(current_type == "extract")
-		if(I && I.owner == target)
-			user.visible_message("<span class='notice'> [user] has decoupled [target]'s [I] with \the [tool].</span>" , \
-		"<span class='notice'> You have decoupled [target]'s [I] with \the [tool].</span>")
+	organ_being_removed = null
+	var/list/organs = target.get_organs_zone(target_zone)
+	if(!length(organs))
+		to_chat(user, "<span class='notice'>There is no removeable organs in [target]'s [parse_zone(target_zone)]!</span>")
```
dearmochi, line 329:
```suggestion
		to_chat(user, "<span class='notice'>There are no removable organs in [target]'s [parse_zone(target_zone)]!</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+			organs[O.name] = O
+
+		organ_being_removed = input("Remove which organ?", "Surgery", null, null) as null|anything in organs
+		if(organ_being_removed && user && target && user.Adjacent(target) && user.get_active_hand() == tool)
+			organ_being_removed = organs[organ_being_removed]
+			if(!organ_being_removed)
+				return SURGERY_FAILED
+			var/obj/item/organ/external/affected = target.get_organ(target_zone)
+			user.visible_message("<span class='notice'>[user] starts to decouple [target]'s [organ_being_removed] with \the [tool].</span>", \
+			"<span class='notice'>You start to decouple [target]'s [organ_being_removed] with \the [tool].</span>" )
```
dearmochi, line 344:
```suggestion
			user.visible_message("<span class='notice'>[user] starts decoupling [target]'s [organ_being_removed] with [tool].</span>", \
			"<span class='notice'>You start decoupling [target]'s [organ_being_removed] with [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
-			if(!istype(thing))
-				thing.forceMove(get_turf(target))
-			else
-				user.put_in_hands(thing)
+	return ..()
+
+/datum/surgery_step/robotics/manipulate_robotic_organs/extract/end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	if(organ_being_removed && organ_being_removed.owner == target)
+		user.visible_message("<span class='notice'> [user] has decoupled [target]'s [organ_being_removed] with \the [tool].</span>" , \
+			"<span class='notice'> You have decoupled [target]'s [organ_being_removed] with \the [tool].</span>")
```
dearmochi, line 355:
```suggestion
	if(organ_being_removed?.owner == target)
		user.visible_message("<span class='notice'>[user] decouples [target]'s [organ_being_removed] with [tool].</span>", \
			"<span class='notice'>You decouple [target]'s [organ_being_removed] with [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+	else
+		user.visible_message("<span class='notice'>[user] can't seem to extract anything from [target]'s [parse_zone(target_zone)]!</span>",
+			"<span class='notice'>You can't extract anything from [target]'s [parse_zone(target_zone)]!</span>")
+	return SURGERY_SUCCESS
 
-	else if(current_type == "extract")
-		user.visible_message("<span class='warning'> [user]'s hand slips, disconnecting \the [tool].</span>", \
+/datum/surgery_step/robotics/manipulate_robotic_organs/extract/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	user.visible_message("<span class='warning'> [user]'s hand slips, disconnecting \the [tool].</span>", \
 		"<span class='warning'> Your hand slips, disconnecting \the [tool].</span>")
```
dearmochi, line 371:
```suggestion
	user.visible_message("<span class='warning'>[user]'s hand slips, disconnecting [tool].</span>", \
		"<span class='warning'>Your hand slips, disconnecting [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+	if(!target.dna.species.has_organ["brain"])
+		to_chat(user, "<span class='danger'>You're pretty sure [target.dna.species.name_plural] don't normally have a brain.</span>")
+		return SURGERY_FAILED
 
+	if(target.get_int_organ(/obj/item/organ/internal/brain/))
+		to_chat(user, "<span class='danger'>Your subject already has a brain.</span>")
+		return SURGERY_FAILED
+
+	user.visible_message("<span class='notice'>[user] starts installing \the [tool] into [target]'s [affected.name].</span>", \
+	"<span class='notice'>You start installing \the [tool] into [target]'s [affected.name].</span>")
```
dearmochi, line 411:
```suggestion
	user.visible_message("<span class='notice'>[user] starts installing [tool] into [target]'s [affected.name].</span>", \
	"<span class='notice'>You start installing [tool] into [target]'s [affected.name].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+
+	return ..()
+
+/datum/surgery_step/robotics/manipulate_robotic_organs/mmi/end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/mmi/tool, datum/surgery/surgery)
+	var/obj/item/organ/external/affected = target.get_organ(target_zone)
+	user.visible_message("<span class='notice'> [user] has installed \the [tool] into [target]'s [affected.name].</span>", \
+		"<span class='notice'> You have installed \the [tool] into [target]'s [affected.name].</span>")
```
dearmochi, line 418:
```suggestion
	user.visible_message("<span class='notice'>[user] installs [tool] into [target]'s [affected.name].</span>", \
		"<span class='notice'>You install [tool] into [target]'s [affected.name].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+
+	user.unEquip(tool)
+	tool.attempt_become_organ(affected, target)
+
+	return SURGERY_SUCCESS
+
+/datum/surgery_step/robotics/manipulate_robotic_organs/mmi/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	user.visible_message("<span class='warning'> [user]'s hand slips!</span>.", \
+		"<span class='warning'> Your hand slips!</span>")
```
dearmochi, line 427:
```suggestion
	user.visible_message("<span class='warning'>[user]'s hand slips!</span>.", \
		"<span class='warning'>Your hand slips!</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+	return SURGERY_FAILED
 
 /datum/surgery_step/robotics/external/amputate
 	name = "remove robotic limb"
-
-	allowed_tools = list(
-	/obj/item/multitool = 100)
+	allowed_surgery_tools = SURGERY_TOOLS_ROBOTIC_REMOVE_ORGAN
+	possible_locs = list("head","l_arm", "l_hand","r_arm","r_hand","r_leg","r_foot","l_leg","l_foot")
```
dearmochi, line 433:
```suggestion
	possible_locs = list("head", "l_arm", "l_hand", "r_arm", "r_hand", "r_leg", "r_foot", "l_leg", "l_foot")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+	surgery_start_stage = SURGERY_STAGE_START
+	next_surgery_stage =  SURGERY_STAGE_START
 
 	time = 100
```
dearmochi, line 437:
```suggestion
	time = 10 SECONDS
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+
 	var/obj/item/organ/external/affected = target.get_organ(target_zone)
-	user.visible_message("[user] starts to decouple [target]'s [affected.name] with \the [tool].", \
-	"You start to decouple [target]'s [affected.name] with \the [tool]." )
+	return !affected.cannot_amputate
+
+/datum/surgery_step/robotics/external/amputate/begin_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	var/obj/item/organ/external/affected = target.get_organ(target_zone)
+	user.visible_message("<span class='warning'>[user] starts to decouple [target]'s [affected.name] with \the [tool]!</span>", \
+	"<span class='notice'>You start to decouple [target]'s [affected.name] with \the [tool].</span>" )
```
dearmochi, line 449:
```suggestion
	user.visible_message("<span class='warning'>[user] starts decoupling [target]'s [affected.name] with [tool]!</span>", \
	"<span class='notice'>You start decoupling [target]'s [affected.name] with [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
 	target.custom_pain("Your [affected.amputation_point] is being ripped apart!")
-	..()
+	return ..()
 
-/datum/surgery_step/robotics/external/amputate/end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool,datum/surgery/surgery)
+/datum/surgery_step/robotics/external/amputate/end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	var/obj/item/organ/external/affected = target.get_organ(target_zone)
-	user.visible_message("<span class='notice'> [user] has decoupled [target]'s [affected.name] with \the [tool].</span>" , \
+	user.visible_message("<span class='warning'> [user] has decoupled [target]'s [affected.name] with \the [tool].</span>" , \
 	"<span class='notice'> You have decoupled [target]'s [affected.name] with \the [tool].</span>")
```
dearmochi, line 457:
```suggestion
	user.visible_message("<span class='warning'>[user] decouples [target]'s [affected.name] with [tool].</span>", \
	"<span class='notice'>You decouple [target]'s [affected.name] with [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
 
-
 	add_attack_logs(user, target, "Surgically removed [affected.name] from. INTENT: [uppertext(user.a_intent)]")//log it
 
 	var/atom/movable/thing = affected.droplimb(1,DROPLIMB_SHARP)
 	if(istype(thing,/obj/item))
 		user.put_in_hands(thing)
```
dearmochi, line 463:
```suggestion
	var/obj/item/limb = affected.droplimb(TRUE, DROPLIMB_SHARP)
	if(limb)
		user.put_in_hands(limb)
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
 
-	return 1
-
-/datum/surgery_step/robotics/external/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool,datum/surgery/surgery)
+	return SURGERY_SUCCESS
 
+/datum/surgery_step/robotics/external/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	user.visible_message("<span class='warning'> [user]'s hand slips!</span>", \
 	"<span class='warning'> Your hand slips!</span>")
```
dearmochi, line 469:
```suggestion
	user.visible_message("<span class='warning'>[user]'s hand slips!</span>", \
	"<span class='warning'>Your hand slips!</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
+	return SURGERY_FAILED
 
 /datum/surgery_step/robotics/external/customize_appearance
 	name = "reprogram limb"
-	allowed_tools = list(/obj/item/multitool = 100)
+	allowed_surgery_tools = SURGERY_TOOLS_ROBOTIC_REPROGRAM
+	possible_locs = list("head", "chest", "l_arm", "l_hand", "r_arm", "r_hand", "r_leg", "r_foot", "l_leg", "l_foot", "groin")
+	surgery_start_stage = SURGERY_STAGE_ROBOTIC_HATCH_UNLOCKED
+	next_surgery_stage =  SURGERY_STAGE_START
 	time = 48
```
dearmochi, line 478:
```suggestion
	time = 4.8 SECONDS
```

code/modules/surgery/robotics.dm
```diff
@@ -2,582 +2,491 @@
...
-		return TRUE
-
-/datum/surgery_step/robotics/external/customize_appearance/begin_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool,datum/surgery/surgery)
+/datum/surgery_step/robotics/external/customize_appearance/begin_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	var/obj/item/organ/external/affected = target.get_organ(target_zone)
-	user.visible_message("[user] begins to reprogram the appearance of [target]'s [affected.name] with [tool]." , \
-	"You begin to reprogram the appearance of [target]'s [affected.name] with [tool].")
-	..()
+	user.visible_message("<span class='notice'>[user] begins to reprogram the appearance of [target]'s [affected.name] with [tool].</span>" , \
+	"<span class='notice'>You begin to reprogram the appearance of [target]'s [affected.name] with [tool].</span>")
```
dearmochi, line 483:
```suggestion
	user.visible_message("<span class='notice'>[user] begins reprogramming the appearance of [target]'s [affected.name] with [tool].</span>", \
	"<span class='notice'>You begin reprogramming the appearance of [target]'s [affected.name] with [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -589,11 +498,10 @@
 	target.UpdateDamageIcon()
 	user.visible_message("<span class='notice'> [user] reprograms the appearance of [target]'s [affected.name] with [tool].</span>", \
 	"<span class='notice'> You reprogram the appearance of [target]'s [affected.name] with [tool].</span>")
```
dearmochi, line 500:
```suggestion
	user.visible_message("<span class='notice'>[user] reprograms the appearance of [target]'s [affected.name] with [tool].</span>", \
	"<span class='notice'>You reprogram the appearance of [target]'s [affected.name] with [tool].</span>")
```

code/modules/surgery/robotics.dm
```diff
@@ -589,11 +498,10 @@
...
-	affected.open = 0
-	return TRUE
+	return SURGERY_SUCCESS
 
-/datum/surgery_step/robotics/external/customize_appearance/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool,datum/surgery/surgery)
+/datum/surgery_step/robotics/external/customize_appearance/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	var/obj/item/organ/external/affected = target.get_organ(target_zone)
 	user.visible_message("<span class='warning'> [user]'s [tool.name] slips, failing to reprogram [target]'s [affected.name].</span>",
 	"<span class='warning'> Your [tool.name] slips, failing to reprogram [target]'s [affected.name].</span>")
```
dearmochi, line 506:
```suggestion
	user.visible_message("<span class='warning'>[user]'s [tool.name] slips, failing to reprogram [target]'s [affected.name].</span>",
	"<span class='warning'>Your [tool.name] slips, failing to reprogram [target]'s [affected.name].</span>")
```

dearmochi (changes requested):
80% of the requested changes are rewriting `visible_message`s to all include `span`s while eliminating spaces and `\the`s. Sorry there are so many.

code/game/gamemodes/miniantags/abduction/abduction_surgery.dm
```diff
@@ -1,100 +1,83 @@
...
-			return FALSE
-		if(affected.is_robotic())
-			return FALSE
-	var/mob/living/carbon/human/H = user
-	// You must either: Be of the abductor species, or contain an abductor implant
-	if((isabductor(H) || (locate(/obj/item/implant/abductor) in H)))
-		return TRUE
-	return FALSE
+/datum/surgery_step/internal/manipulate_organs/abduct/is_valid_user(mob/living/carbon/human/user)
+	return ishuman(user) && (isabductor(user) || (locate(/obj/item/implant/abductor) in user))
```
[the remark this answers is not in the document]
farie82: Will slightly change it since the right part of the `||` will require the type check

code/modules/mob/living/living_defense.dm
```diff
@@ -281,6 +281,31 @@
...
+
+/mob/living/proc/surgery_act(mob/living/user, obj/item/I, datum/surgery/S)
+	if(S.step_in_progress)
+		to_chat(user, "<span class='warning'>This body part is already under surgery!</span>")
+		return TRUE
+	var/obj/item/organ/external/E = get_organ(S.location)
+	. = S.next_step(user, src, I)
+	if(!QDELETED(S) && S.current_stage == SURGERY_STAGE_START) // Remove surgeries that haven't started yet. Amputate will qdel the surgery itself
+		if(E)
+			E.open = 0
```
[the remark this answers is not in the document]
farie82: It is an attempt of making the surgery stage modular. Any value means there is a cut (or equivalent). With higher values meaning the surgery has advanced further. More open so to say. I'll make defines for this

code/modules/surgery/dental_implant.dm
```diff
@@ -1,46 +1,57 @@
...
-	allowed_tools = list(/obj/item/reagent_containers/food/pill = 100)
+	surgery_start_stage = SURGERY_STAGE_DENTAL
+	next_surgery_stage = SURGERY_STAGE_START
+	accept_any_item = TRUE // can_use will check if it's a pill or not
 	time = 16
 
-/datum/surgery_step/insert_pill/begin_step(mob/living/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
-	user.visible_message("[user] begins to wedge \the [tool] in [target]'s [parse_zone(target_zone)].", "<span class='notice'>You begin to wedge [tool] in [target]'s [parse_zone(target_zone)]...</span>")
-	..()
+/datum/surgery_step/dental/insert_pill/can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/reagent_containers/food/pill/tool, datum/surgery/surgery)
```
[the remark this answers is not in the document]
farie82: Bit odd indeed... but I think I'll keep it as is for now. Maybe for a refactor later

code/modules/surgery/limb_reattach.dm
```diff
@@ -2,138 +2,113 @@
...
+/datum/surgery_step/limb/amputate/end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	var/obj/item/organ/external/affected = target.get_organ(target_zone)
+	user.visible_message("<span class='notice'> [user] amputates [target]'s [affected.name] at the [affected.amputation_point] with \the [tool].</span>", \
+	"<span class='notice'> You amputate [target]'s [affected.name] with \the [tool].</span>")
+
+	add_attack_logs(user, target, "Surgically removed [affected.name]. INTENT: [uppertext(user.a_intent)]")//log it
+
+	var/atom/movable/thing = affected.droplimb(1,DROPLIMB_SHARP)
+	if(istype(thing,/obj/item))
+		user.put_in_hands(thing)
```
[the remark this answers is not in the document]
farie82: I'm not 100% sure the returned object always is an `external`. The code in droplimb expects a `movable` there. It's probably always of the type external currently but it's not guaranteed by the code.

code/modules/surgery/robotics.dm
farie82: Same as before

code/__HELPERS/global_lists.dm
```diff
@@ -21,7 +21,16 @@
...
+	for(var/path in (subtypesof(/datum/surgery_step)))
+		var/datum/surgery_step/S = new path()
+		for(var/start_step in S.surgery_start_stage)
+			if(!S.name) // abstract subtype
+				continue
+			if(!GLOB.surgery_steps[start_step])
+				GLOB.surgery_steps[start_step] = list(S)
+			else
+				var/list/SS = GLOB.surgery_steps[start_step]
+				SS.Add(S)
```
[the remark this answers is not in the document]
farie82: image
Won't work since it'll just assign it instead of make a list

farie82:
> 80% of the requested changes are rewriting `visible_message`s to all include `span`s while eliminating spaces and `\the`s. Sorry there are so many.

Thanks for the *small* review. I think I got them all now plus some added refactors. Thanks!

farie82:
Damn linter not running. Close and reopen time

dearmochi (approved):

code/modules/mob/living/living_defense.dm
```diff
@@ -284,6 +284,55 @@
...
+	return FALSE
+
+
+//check if mob is lying down on something we can operate him on.
+/mob/living/proc/can_be_operated_on()
+	if(locate(/obj/machinery/optable, loc) && (lying || resting))
+		return TRUE
+	if(locate(/obj/structure/bed, loc) && (buckled || lying || IsWeakened() || stunned || paralysis || sleeping || stat))
+		return TRUE
+	if(locate(/obj/structure/table, loc) && (lying || IsWeakened() || stunned || paralysis || sleeping || stat))
```
SteelSlayer, line 303: I think you could use `incapacitated()` here to cut down on copy paste, since it covers *most* of the checks you want.

farie82: will see about changing that mess there. Looks like it can be simplified by a lot

farie82: Changed it to just check if you're lying. I don't see any reason to have the other checks since all cases I can think of will make the `lying` variable have a value.

Oyu-De:
holy crap this is still open?

Aurorablade:
a part of me kinda dies on seeing alot of the work i did way back gutted out but i guess its okay. The amount of defines seems a bit lengthy though but i dunno how else you would go about it.

farie82:
Smoke test held on the test server. Special thanks to @Qwertyotoforty @Sirryan2002 Kerbal Citizen and Praise Dun Bun for helping with testing the PR!

The following findings were fixed in the latest commit:
- Xenomorph surgery not working. They have no external organs which the code was expecting that they had.
- The cavity surgery fail message did not take into account that you not always have a tool

In addition cavity surgery now only has one stage. This means that once you opened a cavity you can put something in and then take it out again without having to close it off. This made 0 sense to me in hindsight so that got changed. Also removes one stage define

Fox-McCloud:
I'd like to first apologize for this being let go for far too long. It really shouldn't have and I take responsibility for that.

That said, there's some strong concerns among the team with respect to this PR. It is largely a shoe-in that we don't feel is apt to work for us in the long term. This brings us out of line with TG, which will make implementing their advanced surgeries quite difficult and leave us dependent on a solo coder for this system to be maintained and work with it. It also does not have a way to handle things like the Incision Management System, which is a sad loss.

Additionally, I don't feel the way this was done, refactor wise is ideal; we now have the standard tools use of `screwdriver_act`, `welder_act` and what have you---in conjunction to this we also have `tool_behavior`; this makes use of neither of those.

On the whole, due to overall design and coding issues, we formally object to this PR, and it will be closed.

PR #13016, opened 2020-02-22
Fix cult related runtimes

datlo:
Fix some cult related runtimes, most notably cloning a cultist will now properly give them their abilities/hud back.

Runtimes bad. 

:cl:
fix: Fix some cult related runtimes.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/runes.dm
+++ b/code/game/gamemodes/cult/runes.dm
@@ -167,11 +167,12 @@ structure_check() searches for nearby cultist structures required for the invoca
 /obj/effect/rune/proc/burn_invokers(var/list/mobstoburn)
 	for(var/M in mobstoburn)
 		var/mob/living/L = M
-		to_chat(L, "<span class='cultlarge'><i>\"YOUR SOUL BURNS WITH YOUR ARROGANCE!!!\"</i></span>")
-		if(L.reagents)
-			L.reagents.add_reagent("hell_water", 10)
-		L.Weaken(7)
-		L.Stun(7)
+		if(L)
+			to_chat(L, "<span class='cultlarge'><i>\"YOUR SOUL BURNS WITH YOUR ARROGANCE!!!\"</i></span>")
+			if(L.reagents)
+				L.reagents.add_reagent("hell_water", 10)
+			L.Weaken(7)
+			L.Stun(7)
 	fail_invoke()
 
 /obj/effect/rune/proc/do_invoke_glow()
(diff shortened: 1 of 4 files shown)


code/game/gamemodes/cult/runes.dm
```diff
@@ -167,11 +167,12 @@ structure_check() searches for nearby cultist structures required for the invoca
 /obj/effect/rune/proc/burn_invokers(var/list/mobstoburn)
 	for(var/M in mobstoburn)
 		var/mob/living/L = M
-		to_chat(L, "<span class='cultlarge'><i>\"YOUR SOUL BURNS WITH YOUR ARROGANCE!!!\"</i></span>")
-		if(L.reagents)
-			L.reagents.add_reagent("hell_water", 10)
-		L.Weaken(7)
-		L.Stun(7)
+		if(L)
```
Crazylemon64: Not sure how a null entry would get in this list

code/game/machinery/cloning.dm
```diff
@@ -433,8 +433,7 @@
 	if(H.mind in SSticker.mode.syndicates)
 		SSticker.mode.update_synd_icons_added()
 	if(H.mind in SSticker.mode.cult)
-		SSticker.mode.add_cultist(occupant.mind)
-		SSticker.mode.update_cult_icons_added() //So the icon actually appears
+		SSticker.mode.update_cult_icons_added(H.mind) //So the icon actually appears
 		SSticker.mode.update_cult_comms_added(H.mind) //So the comms actually appears
```
Crazylemon64: I assume `add_cultist` was removed because it was redundant?

code/game/gamemodes/cult/runes.dm
datlo: No clue either. It did happen, though, got the runtime from the public runtime logs of the live server

code/game/machinery/cloning.dm
datlo: Yeah if the mind is in SSticker.mode.cult then it's already a cultist.

code/game/gamemodes/cult/runes.dm
farie82: It's because of this line https://github.com/ParadiseSS13/Paradise/blob/49bdfbf3f02b91b56be4a1bdface2d51f6544dcf/code/game/gamemodes/cult/runes.dm#L147
if List len == req_cultists
then it'll pick req_cultists + 1 cultists from that list. And if the list is empty it'll put a null in there. Change it from 0 to 1 and it'll work. No need for this if here then

datlo:
Closing this as the blood cult update includes these changes and is ready for merging

PR #13020, opened 2020-02-23
Caps the health and damage of sentient animals

farie82:
## What Does This PR Do
Caps the health of a sentient animal to 300, min damage to 20 and max damage to 25

## Why It's Good For The Game
Animals like Araneus are stronger than nukies when they get chosen for said event. It is rather lame to fight against due to them being that strong

## Changelog
:cl:
balance: Sentient animal event now has a max cap for the health and damage values
/:cl:

### DIFF
--- a/code/modules/events/sentience.dm
+++ b/code/modules/events/sentience.dm
@@ -36,9 +36,7 @@
 	SA.universal_speak = 1
 	SA.sentience_act()
 	SA.can_collar = 1
-	SA.maxHealth = min(SA.maxHealth + 200, max(SA.maxHealth, 300))
-	SA.melee_damage_lower = min(SA.melee_damage_lower + 15, max(SA.melee_damage_lower, 20))
-	SA.melee_damage_upper = min(SA.melee_damage_upper + 15, max(SA.melee_damage_upper, 25))
+	SA.maxHealth = max(SA.maxHealth, 200)
 	SA.health = SA.maxHealth
 	SA.del_on_death = FALSE
 	greet_sentient(SA)
--- a/code/modules/events/sentience.dm
+++ b/code/modules/events/sentience.dm
@@ -36,9 +36,9 @@
 	SA.universal_speak = 1
 	SA.sentience_act()
 	SA.can_collar = 1
-	SA.maxHealth += 200
-	SA.melee_damage_lower += 15
-	SA.melee_damage_upper += 15
+	SA.maxHealth = min(SA.maxHealth + 200, 300)
+	SA.melee_damage_lower = min(SA.melee_damage_lower + 15, 20)
+	SA.melee_damage_upper = min(SA.melee_damage_upper + 15, 25)
 	SA.health = SA.maxHealth
 	SA.del_on_death = FALSE
 	greet_sentient(SA)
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
I really question if Araneus is the best inclusion these days.

It's really only worth it for the meme of the HoS attacking him then getting wrecked.

lolzyking:
I'm against this because barring nuke rounds, sentient  animals often get valided, they need the huge health to survive the grey tide lynching.

Maybe if we buffed sentient animal moves peed as compensation this would be fine.

farie82:
> I'm against this because barring nuke rounds, sentient animals often get valided, they need the huge health to survive the grey tide lynching.
> 
> Maybe if we buffed sentient animal moves peed as compensation this would be fine.

Most if not all except araneus their stats won't be affected by this. A cat etc have 20 HP and damage of 0 even....

Leanfrog:
Something else to take into account for simple mobs is how easily they're healed, all it takes is a trauma kit to heal most if not all of their health, and trauma kits are a very abundant resource on station, that 450 HP is restored incredibly easily for the power that it packs behind it

lolzyking:
Fair I thought the nerf would negatively impact some mobs, but with the buff nerf only really noticeable on adult slimes and araneus this is probably a good solution.

Fox-McCloud:
In all honest, I'd just do what TG does. 

Their nerf was even harsher; it just boosts health, but doesn't boost attack damage, at all.

Quite frankly, that's probably a good thing.

code/modules/events/sentience.dm
```diff
@@ -36,9 +36,9 @@
 	SA.universal_speak = 1
 	SA.sentience_act()
 	SA.can_collar = 1
-	SA.maxHealth += 200
-	SA.melee_damage_lower += 15
-	SA.melee_damage_upper += 15
+	SA.maxHealth = min(SA.maxHealth + 200, 300)
+	SA.melee_damage_lower = min(SA.melee_damage_lower + 15, 20)
+	SA.melee_damage_upper = min(SA.melee_damage_upper + 15, 25)
```
Fox-McCloud: `SA.maxHealth = max(SA.maxHealth, 200)`

farie82: Will change once the surgery stuff is fixed. Thanks!

code/modules/events/sentience.dm
```diff
@@ -36,9 +36,9 @@
 	SA.universal_speak = 1
 	SA.sentience_act()
 	SA.can_collar = 1
-	SA.maxHealth += 200
-	SA.melee_damage_lower += 15
-	SA.melee_damage_upper += 15
+	SA.maxHealth = min(SA.maxHealth + 200, max(SA.maxHealth, 300))
+	SA.melee_damage_lower = min(SA.melee_damage_lower + 15, max(SA.melee_damage_lower, 20))
+	SA.melee_damage_upper = min(SA.melee_damage_upper + 15, max(SA.melee_damage_upper, 25))
```
Fox-McCloud: Again, I'd remove the melee damage upper/lower boost altogether.

Then just make the health boost this: `SA.maxHealth = max(SA.maxHealth, 200)`

farie82: Oh like that. Will do

PR #13021, opened 2020-02-24
Adds CBD

Fox-McCloud:
Credit to Goon for this.

Adds cbd to cannabis--it's a non-psychoactive drug that helps you relax and also promotes healing; it also has mild anti-stun properties to it.

Also adds a contented sigh emote

:cl: Fox McCloud
add: Marijuana plants now have cbd in them; it's a non-psychoactive/legal drug that promotes relaxation and healing
add: Adds a contented sigh emote *hsigh
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -590,7 +590,7 @@
 				message = "<B>[src]</B> sighs contentedly."
 				m_type = 2
 			else
-				message = "<B>[src]</B> makes a [pick("chill", relaxed")] noise"
+				message = "<B>[src]</B> makes a [pick("chill", "relaxed")] noise"
 				m_type = 2
 
 		if("laugh", "laughs")
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -590,7 +590,7 @@
 				message = "<B>[src]</B> sighs contentedly."
 				m_type = 2
 			else
-				message = "<B>[src]</B> makes a weak noise"
+				message = "<B>[src]</B> makes a [pick("chill", relaxed")] noise"
 				m_type = 2
 
 		if("laugh", "laughs")
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -585,6 +585,14 @@
 					message = "<B>[src]</B> makes a weak noise"
 					m_type = 2
 
+		if("hsigh", "hsighs")
+			if(!muzzled)
+				message = "<B>[src]</B> sighs contentedly."
+				m_type = 2
+			else
+				message = "<B>[src]</B> makes a weak noise"
+				m_type = 2
+
 		if("laugh", "laughs")
 			var/M = handle_emote_param(param)
 			if(miming)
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -585,6 +585,14 @@
 					message = "<B>[src]</B> makes a weak noise"
 					m_type = 2
 
+		if("hsigh", "hsighs")
+			if(!muzzled)
+				message = "<B>[src]</B> sighs contentedly."
+				m_type = 2
+			else
+				message = "<B>[src]</B> makes a weak noise"
```
KasparoVy: `makes a chill/relaxed noise`

KasparoVy:
Blaze it

Mitchs98:
Dave's not here man.

code/modules/mob/living/carbon/human/emote.dm
Fox-McCloud: good suggestion!

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -590,7 +590,7 @@
 				message = "<B>[src]</B> sighs contentedly."
 				m_type = 2
 			else
-				message = "<B>[src]</B> makes a weak noise"
+				message = "<B>[src]</B> makes a [pick("chill", relaxed")] noise"
```
KasparoVy: Hell yeah!

pinatacolada:
Should also make you hungry

Carthusia:
heh, you get a little moodle in the top right under your hunger bar that notes you got the "Munchies"

To confirm though: This will remove the spacedrug like effect from it and replace it with the proper CBD effect?

Fox-McCloud:
THC is still around; its still does stereotypical psycho-active THC things.

This is a completley and utterly separate chem.

code/modules/reagents/chemistry/reagents/drugs.dm
```diff
@@ -517,6 +517,32 @@
...
+	if(prob(5))
+		to_chat(M, "<span class='notice'>[pick("You feel peaceful.", "You breathe softly.", "You feel chill.", "You vibe.")]</span>")
+	if(prob(10))
+		M.AdjustConfused(-5)
+		update_flags |= M.SetWeakened(0, FALSE)
+	if(volume >= 70 && prob(25))
+		if(M.reagents.has_reagent("thc") <= 20)
+			M.Drowsy(10)
+	if(prob(25))
+		update_flags |= M.adjustBruteLoss(-2, FALSE)
```
Kyep: This healing seems pretty substantial. Even Saline Glucose, which is the basic medical chem, heals only 0.66 brute/burn per tick. This heals 0.5 brute+burn per tick, as well as reducing confusion. Not sure about its effects on weakening, too. 

Is this meant to effectively replace most brute/burn treatment for botanists, and whoever they give it to?

SynthNumerics:
for clarification on cbd itself, its actually the chemical that prevents/limits the high that you gain from thc- the chemical you typically think of when you consider cannabis that makes you super out of it, get peckish etc; a certain level of it is honestly needed to prevent "bad trips". cbd gives a kind of "body high" where your mind is clear but your body as mentioned gains a level of relaxation with its pretty potent painkiller an mild healin properties (i use it daily to help manage my pain) among other things that aid the health of the body (i can talk more abt it on disc if folks want). it's a bit of a miracle chemical in a way.

im feelin a lil late but regardin how it affects the game however, i have questions.
1. would the cbd be present in premade (ie medicinal from the psych office) cigarettes an would they be altered for the straight cbd version due to its connection or would they retain both chemicals in certain percentages for both effects?
2. does it have the blockin affect on thc or does it remain independent in this case?

PR #13022, opened 2020-02-24
Makes Most Emotes Age Pitched

Fox-McCloud:
Makes most emotes age pitched.

Things like buzz, ping, yes, no, quill rustling, hissing, howling, growling, and the likes are now all age pitched.

Clapping and snapping are the exception, at this is less part of the SS13 stereotype.

Additionally, the chance to make noise when speaking (common to Diona and Vox) are now also age pitched.

Should add a lot more diversity to the crew and how they speak/interact beyond just screams  and deathgasps.

:cl: Fox McCloud
add: Most emotes (minus clapping and snapping) are now age pitched
add: The speak sounds are now also age pitched
/:cl:

### DIFF
--- a/code/modules/mob/hear_say.dm
+++ b/code/modules/mob/hear_say.dm
@@ -47,7 +47,7 @@
 		. = trim(. + trim(msg))
 		. += "\""
 
-/mob/proc/hear_say(var/list/message_pieces, var/verb = "says", var/italics = 0, var/mob/speaker = null, var/sound/speech_sound, var/sound_vol)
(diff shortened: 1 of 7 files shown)


KasparoVy:
NICE! Howling and growling will sound pretty awesome age-pitched. Old ass Vulp growling sounds utterly wicked

Carthusia:
Very nice.

Question: This may be out of the scope of the PR, but can an option be added to disable fingers snapping OFF? The sound it makes is so out of place when your expecting the normal one, and it kind of wrecks a moments immersion as people get worried when a finger snaps off, even more so when you get 5 fingers snapping off within 5 minutes as a chef who snaps his fingers after each meal served, and your a vox with only 3 fingers.

PR #13025, opened 2020-02-24
Alphabetically sorts hairstyles & appearance things

Ty-Omaha:
## What Does This PR Do
Sorts all the hairstyles and accessories in alphabetical order.

this is a direct result of me getting angry because I couldn't find if there were dreadlocks or not.

goes with #13038

🆑 
tweak: Alphabetically sorts Appearance list items such as hairstyles
/ 🆑

### DIFF
--- a/code/modules/client/preference/preferences.dm
+++ b/code/modules/client/preference/preferences.dm
@@ -1308,7 +1308,7 @@ var/global/list/special_role_times = list( //minimum age (in days) for accounts 
 					else //Not using the whitelist? Aliens for everyone!
 						new_species += GLOB.whitelisted_species
 
-					species = input("Please select a species", "Character Generation", null) in new_species
+					species = input("Please select a species", "Character Generation", null) in sortTim(new_species, /proc/cmp_text_asc)
 					var/datum/species/NS = GLOB.all_species[species]
 					if(!istype(NS)) //The species was invalid. Notify the user and fail out.
 						species = prev_species
@@ -1406,7 +1406,7 @@ var/global/list/special_role_times = list( //minimum age (in days) for accounts 
 						if(!(lang.flags & RESTRICTED))
 							new_languages += lang.name
 
-					language = input("Please select a secondary language", "Character Generation", null) in new_languages
+					language = input("Please select a secondary language", "Character Generation", null) in sortTim(new_languages, /proc/cmp_text_asc)
 
 				if("autohiss_mode")
 					if(S.autohiss_basic_map)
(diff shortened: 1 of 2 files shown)


AffectedArc07:
Any chance we can get the scissors menu alphabetised? IIRC it isnt at the moment/

Ty-Omaha:
I think would need a whole refactor since from a quick glance it appears to just hardcode the code of the procs that are changed in this PR instead of just using them, will look more into it later

https://github.com/ParadiseSS13/Paradise/blob/3fee0f2019bb83f27aca6f44c0f0770187a7ed59/code/game/objects/items/weapons/scissors.dm

Mitchs98:
I love you.

code/modules/client/preference/preferences.dm
```diff
@@ -1308,7 +1308,7 @@ var/global/list/special_role_times = list( //minimum age (in days) for accounts
 					else //Not using the whitelist? Aliens for everyone!
 						new_species += GLOB.whitelisted_species
 
-					species = input("Please select a species", "Character Generation", null) in new_species
+					species = input("Please select a species", "Character Generation", null) in sortTim(new_species, /proc/cmp_text_asc)
```
farie82: I think you can just use sortList() instead of these. Same effect with less overhead

taukausanake:
Can we at least get the bald/shaven options first in the lists? It's annoying having to look through the list, even if in order, to find the blank state

Ty-Omaha:
@taukausanake the way it works is sorting by alphabetical order, when it encounters numbers it puts them first. doing that would require snowflake code

code/modules/client/preference/preferences.dm
Ty-Omaha: I can, but sortTim(uses Timsort) is more optimized compared to sortList(Mergesort)

Ty-Omaha:
adjusting scissors is out of scope for this PR they can get their own refactor PR

Ty-Omaha:
@taukausanake why are you downvoting that? you want me to add a specific thing that says "if list item == "bald" then put at front" that will run everytime anything in the code uses sortTim? 😐

taukausanake:
Downvote means I disagree. It's annoying having to try to find bald or shaven in these lists. You'd expect the blank states to be at the top. That's all I'm saying.

_People actually pay attention to the votes?_ I didn't know we were a democracy now /s

Crazylemon64:
@Ty-Omaha the way you'd implement that would be to exclude "bald" from the list beforehand, and then prepend it to the list at the point of return

Ty-Omaha:
but you'd have to use an if statement to exclude it from the list

Ty-Omaha:
unless theres some other way

PR #13027, opened 2020-02-24
Goonchem: Heparin Rework

Fox-McCloud:
Credit largely to Goon.

Reworks heparin from being an explicit poison to a medical chem. Likely to be expanded upon later by impacting blood clots and maybe tying into the blood system a little more.

Heparin is less directly a reagent meant exclusively for poisoning someone and is now a medical chem.

Here's what it'll do:
- Worsen existing internal and regular bleeding (but not cause bleeding on its own)
- If you overdose on it, the individual will vomit up blood or bleed excessively from their pores
- May stop heart failure

The recipe has also changed:
- meat slurry, phenol, sulfuric acid, and sugar

Medical will start with 4 syringes of it per vendor. 

:cl: Fox McCloud
tweak: reworked heparin. It's no longer explicitly a poison, but a medical chem. It can help with heart failure, but has the downside of making existing bleeding worse, and if overdosed, will cause the patient to lose quite a bit of blood
tweak: heparin has a new recipe: sulfuric+meat slurry+phenol+sugar
/:cl:

### DIFF
--- a/code/datums/diseases/critical.dm
+++ b/code/datums/diseases/critical.dm
@@ -80,8 +80,8 @@
 	spread_text = "The patient is having a cardiac emergency"
 	max_stages = 3
 	spread_flags = SPECIAL
-	cure_text = "Atropine or Epinephrine"
-	cures = list("atropine", "epinephrine")
+	cure_text = "Atropine, Epinephrine, or Heparin"
+	cures = list("atropine", "epinephrine", "heparin")
 	cure_chance = 10
 	needs_all_cures = FALSE
 	viable_mobtypes = list(/mob/living/carbon/human)
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1178,7 +1178,7 @@
 					/obj/item/reagent_containers/glass/bottle/salicylic = 4, /obj/item/reagent_containers/glass/bottle/potassium_iodide =3, /obj/item/reagent_containers/glass/bottle/saline = 5,
 					/obj/item/reagent_containers/glass/bottle/morphine = 4, /obj/item/reagent_containers/glass/bottle/ether = 4, /obj/item/reagent_containers/glass/bottle/atropine = 3,
 					/obj/item/reagent_containers/glass/bottle/oculine = 2, /obj/item/reagent_containers/glass/bottle/toxin = 4, /obj/item/reagent_containers/syringe/antiviral = 6,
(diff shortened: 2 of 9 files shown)


TDSSS:
I dislike putting every chem medical could need in the vendors, give chemistry something to do. Since defibs are a thing you don't strictly speaking need it, it's just a luxury/convenience thing.

Kyep:
Why is this better?

AdamElTablawy:
if you wanted to make heparin not a poison you'd port goon's blood hypertension, this just makes it so niche it's useless

Mitchs98:
While I do like the idea that it won't be an ezpz method of shooting and killing people anymore. It's gonna be really awkward for people to realize it's not a lethal chem anymore, especially with our changelog not working. That said just something for people to get used too I guess, especially as I feel heparin is overpowered as hell as it is currently.

Offtopic; Though...I'd also rather not have hypertension buut...if it's added I'd guess this'd help with it? If it must be added, anyways.

Fox-McCloud:
>if you wanted to make heparin not a poison you'd port goon's blood hypertension

`Likely to be expanded upon later by impacting blood clots and maybe tying into the blood system a little more.`

MRgaty:
I honestly dont see the usefulness in medical

Quantum-M:
It seems to be more to be too specific in a situation that can be remedied in different ways that are less harmful (given that a lot of injuries tend to be brute, it is most likely a patient will come in bleeding in eithercase)

I also agree with @TDSSS , at the very least it should either be in even lower quantities or none at all in the vendors, otherwise it will make chemistry increasingly redundant.

Aquilar:
I don't see this as useful for medical outside of giving people possibly incurable internal bleeding and killing people via blood loss when it's meant to save them.
Especially right now as a round start chem when atropine less lethal alternative.

You mention it doing stuff down the line, wouldn't it be better served being bundled along with those features it'll be affecting instead of now?

farie82:
In the current form it will serve as chemical bloat in the vendor. Nerfing heparin is fine but the current use case for the medicine is not clear to me. I'd say adding it later with more related changes would make more sense like Aquilar said

tiggypuff:
> I have concerns that this PR makes heparin fundamentally useless. Why would you bother using it to treat heart issues? Isn't it just easier to defib or clone people? Then again, I'm not a medical player. Perhaps it won't be useless.

Since this PR makes heparin purge cholesterol and cure cardiac failures, it can be used to treat cholesterol induced cardiac failures. This kind of cardiac failure is rather uncommon though. in most cases the cardiac failures treated in medbay are caused by the patients being in critical condition.
Also, due to the way cures work for the critical subtype of diseases, heparin could be used to boost the probability of curing a cardiac arrest. 
A major disadvantage is that heparin worsens bleedings, which are pretty common with patients in critical condition. This limits the usefulness of heparin.

PR #13029, opened 2020-02-25
Changes power tools to longer destroy themselves. [DNM]

Aquilar:
## What Does This PR Do

as a title. 
doing as suggested from https://github.com/ParadiseSS13/Paradise/pull/12989#issuecomment-590116650

## Why It's Good For The Game

Jaws of life and power drill no longer destroy themselves to change function.

This can not be merged due to one glaring problem exhibited by this image here
weh

there is too many istype checks for specific items missed by the tool refactor and these would need to be dealt with before this could be implemented.

## Changelog
:cl:
del: duplicate entires
tweak: changes the 'power' tools to be single entries 
:cl:


AffectedArc07:
This will probably break the hell out of @Citinited 's tool rework.

Proceed with caution

code/game/objects/items/tools/crowbar.dm
```diff
@@ -70,16 +70,44 @@
 	usesound = 'sound/items/jaws_pry.ogg'
 	force = 15
 	toolspeed = 0.25
-	var/airlock_open_time = 100 // Time required to open powered airlocks
+
+obj/item/crowbar/power/examine()
```
Citinited: Missed a `/` on the start of this

Also I think you should put `mob/user` in the args of this, to remain consistent with the rest of the codebase

code/game/objects/items/tools/screwdriver.dm
```diff
@@ -83,16 +83,28 @@
 	toolspeed = 0.25
 	random_color = FALSE
 
+/obj/item/screwdriver/power/examine()
```
Citinited: Same as above

Fox-McCloud:
Going to close this---until someone finishes up @Citinited 's attackby stuffs, this will just continue to become more and more dated, but, like the other PR, this is *much* needed.

Issue #13031, opened 2020-02-26
Personal crafting is broke for many recipes


Aquilar:
**Issue Description**:

Personal crafting no longer works for any recipe that has tools like sushi mats, pens, damp rags, crayons and so on.

**What did you expect to happen**: 

trying to craft various paper decorations

**What happened instead**:

nothing, got construction failed, tool missing.

**Why is this bad/What are the consequences**:

means most crafting recipes in personal crafting no longer work.

**Steps to reproduce the problem**:

try making any craft-able item that uses a pen, crayon or damp rag.

**When did the problem start happening**:

probably since tool_refactor got merged. 

**Extra information**:

none

PR #13032, opened 2020-02-26
Makes personal crafting for all recipes work again.

Aquilar:
## What Does This PR Do

fixes: https://github.com/ParadiseSS13/Paradise/issues/13031#issue-571330753

also converts remaining /obj/item/"tool" paths in recipes to tool behaviours.
splits the two types of tools into paths and tool_behaviours for cleaner and more readable lists with a different proc reading the new list.

## Why It's Good For The Game

people can make sushi, decorations and fake toolboxes again!

## Changelog
:cl:
tweak: changes old tool paths to tool behaviours
add: new proc and list for recipes.
fix: fixes personal crafting using tools that are not tool_behaviours.
/:cl:

### DIFF
--- a/code/modules/crafting/craft.dm
+++ b/code/modules/crafting/craft.dm
@@ -109,19 +109,28 @@
 		return TRUE
 	var/list/possible_tools = list()
 	var/list/tools_used = list()
+	var/list/other_tools = list()
 	for(var/obj/item/I in user.contents)
 		if(istype(I, /obj/item/storage))
 			for(var/obj/item/SI in I.contents)
 				if(SI.tool_behaviour) //Only add things that we could actually use as a tool
 					possible_tools += SI
+				else
+					other_tools += SI.type				
 		if(I.tool_behaviour)
 			possible_tools += I
+		else
+			other_tools += I.type
 	possible_tools |= contents["other"]
+	other_tools |= contents["other"]
 	main_loop:
 		for(var/A in R.tools)
 			for(var/obj/item/I in possible_tools)
 				if(A == I.tool_behaviour)
 					tools_used += I
+					continue main_loop
+			for(var/I in other_tools)
+				if(ispath(I,A))
 					continue main_loop
 			return FALSE
 	for(var/obj/item/T in tools_used)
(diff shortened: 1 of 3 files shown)


Carthusia:
I was wondering why I no longer could make complex sushi even with all I needed, huzzah!

code/modules/crafting/recipes.dm
```diff
@@ -578,15 +578,15 @@
 	result = /obj/item/decorations/sticky_decorations/flammable/heart
 	reqs = list(/obj/item/paper = 1,
 				/obj/item/stack/tape_roll = 1)
-	tools = list(/obj/item/wirecutters,
+	tools = list(TOOL_WIRECUTTER,
```
farie82: I'd make a new list for the "other tools".
Call it extra_required_items or such.
This will keep the list of one type and will be easier to understand later on for people

Aquilar: it's fixable enough.

code/modules/crafting/craft.dm
```diff
@@ -129,26 +130,49 @@
...
+		if(istype(I, /obj/item/storage))
+			for(var/obj/item/SI in I.contents)
+				other_possible_tools += SI.type	// filters type paths
+		other_possible_tools += I.type
+	
+	other_possible_tools |= contents["other"] // This check for type paths
+	main_loop: // checks if all tools found are usable with the recipe
+		for(var/A in R.pathtools)
+			for(var/I in other_possible_tools)
+				if(ispath(I,A))
```
farie82: Need to include an else to this due to contents["other"] being a list of items.
So something like:
```suggestion
				if(ispath(I,A))
					continue main_loop
				else if(isitem(I) && ispath(I.path, A)
					continue main_loop
```
Or put both checks in one line

code/modules/crafting/craft.dm
```diff
@@ -103,21 +104,24 @@
 					for(var/datum/reagent/A in RC.reagents.reagent_list)
 						.["other"][A.type] += A.volume
 			.["other"][I.type] += 1
+	for(var/obj/item/B in get_environment(user))
```
farie82: Can do this in one for loop. Also would want the hologram check in there.

PR #13034, opened 2020-02-27
Goonchem: Addiction Rework

Fox-McCloud:
Credit to goon for the idea and *most* of the implementation.

Reworks addiction to make it less of an avoidable mallus.

Currently, addiction only rolls for a chance to make you addicted when you initially swallow the reagent; if you can manage to get the reagent into your bloodstream some other way, you can bypass the issues with addiction completely.

Likewise, regardless of how much you take, even if it's 0.00001 units, you can still get addicted to it; likewise, if you're eating something with a low addiction chance in it (say a salad with omnizine), you increase your chances of getting addicted based on how many times you take a bite. This is...silly.

The last issue is that sometimes addiction withdrawl is awfully severe if you don't feed the addiction; this is warranted for the *vast* bulk of drugs, but for a few of the more silly/gimmick ones, it can be a bit annoying. There's now a lighter version of addiction--you'll still twitch and have a headache, but you'll no longer fall on the floor stunned--instead you'll just be slowed for a short duration.

As for how addiction works now:
Addiction is still probability based and rolled for each time you process reagents. You must cross a threshold for the amount of reagents you process before you can get addicted; while this amount is usually fairly low (5-10 typically), this avoids situations of suddenly getting addicted because you were exposed to 1 unit of meth. As you process these reagents, it'll keep track of how much you've processed---once you stop processing that reagent, the tally will ***very slowly*** deplete; resuming taking an addictive reagent during this time will start increasing the tally again.

All addiction probabilities have been adjusted to be much lower to compensate for this; typically it's about the duration of your drug use that is going to govern if you're addicted now as opposed to just being exposed, once, to the drug.

And yes, this means that cigarettes and other methods of smoking are now definitively addictive.

Some new things you can now get addicted to:
- alcohol (very very low addiction chance)
- tea (insanely low addiction chance; minor withdrawl)
- capsaicin
- fake cheese

Things that have minor addiction:
- coffee
- tea
- capsaicin
- fake cheese
- weird cheese
- nicotine

This should make addiction a bit more impactful for those who are cheesing it with alternative injection methods while also discouraging doping a ton of chemicals at once while still making minor "recreational"  (or emergency edge case) drug use more than viable.

:cl: Fox McCloud
tweak: addiction has been overhauled. It's more metabolism based than just ingesting it; having a reagent in your system longer is going to increase your chances of becoming addicted
tweak: some reagents have a new addictions, like ethanol, capsaicin, tea, and fake cheese
tweak: new more minor addiction that is less debilitating: coffee, tea, capsaicin, fake cheese, weird cheese, and nicotine all have minor addiction effects
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -237,7 +237,7 @@ var/const/INGEST = 2
 
 	if(LAZYLEN(addiction_threshold_accumulated))
 		for(var/thing in addiction_threshold_accumulated)
-			if(reagent_list.Find(thing))
+			if(has_reagent(thing))
 				continue // if we have the reagent in our system, then don't deplete the addiction threshold
 			addiction_threshold_accumulated[thing] -= 0.01 // Otherwise very slowly deplete the buildup
 			if(addiction_threshold_accumulated[thing] <= 0)
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -11,6 +11,7 @@ var/const/INGEST = 2
 	var/atom/my_atom = null
 	var/chem_temp = T20C
 	var/list/datum/reagent/addiction_list = new/list()
+	var/list/addiction_threshold_accumulated = new/list()
 	var/flags
 	var/list/reagents_generated_per_cycle = new/list()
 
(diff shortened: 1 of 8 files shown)


code/modules/reagents/chemistry/holder.dm
```diff
@@ -233,6 +234,15 @@ var/const/INGEST = 2
 	if(M)
 		temperature_reagents(M.bodytemperature - 30)
 
+
+	if(LAZYLEN(addiction_threshold_accumulated))
```
farie82: Is this check needed? The for won't do anything if the list is empty. 
Unless this is for performance reasons ofc

Fox-McCloud: No point in really running the below with an empty list.

Issue #12971, opened 2020-02-11
[EASY] Old rule of adminhelp when disconnect still in game text


Ty-Omaha:
**Issue Description**:
Tells people an outdated rule to adminhelp if they disconnect in an important role.

**What did you expect to happen**: 
They should be told to inform command and go into cryo or wipe core, if they cant do that then they should adminhelp. The current rule says: 

> If you need to leave the round early as a VIP, Head of Staff, or the station AI you must inform Command and place yourself in cryo, or in the case of the AI, use the Wipe Core verb in the OOC tab. If you cannot do this for any reason, then you must instead Adminhelp prior to leaving so that an Administrator can remove you after logging out. Failing to inform Command and cryoing yourself, wiping your core, or informing the admins will lead to a VIP and Command, or Synthetic Jobban where appropriate;

**Why is this bad/What are the consequences**:
Inconsistency

https://github.com/ParadiseSS13/Paradise/blob/cc72b2fa5e452b8bc7cd555277c15e3cb332da8e/code/controllers/subsystem/jobs.dm#L442

PR #13039, opened 2020-02-28
Updates notify admin text for command roles

Ty-Omaha:
Updates the message people get to inform admins if they have to disconnect to match our current rules.

Fixes: #12971

:cl:
tweak: Updated wording of the round start text command roles receive if they have to disconnect to reflect our current rules.
/ :cl:


shawnhind:
> notify your department/command

That makes it sound like they could either notify just their own department or notify command and doing either one of those would be okay but the rule specifically says that command should be notified.

Maybe change to notify your "department and command"?

Ty-Omaha:
@Kyep k done

Issue #12991, opened 2020-02-16
Nukie shuttle consoles are destructible despite INDESTRUCTIBLE tag


datlo:
**Issue Description**:

You can destroy nukie shuttle consoles when they have the INDESTRUCTIBLE tag.

**What did you expect to happen**: 

You shouldnt be able to.

**What happened instead**:

Sad nukies.

**Why is this bad/What are the consequences**:

Bug.

**Steps to reproduce the problem**:

I am not certain but likely not difficult to reproduce since it's happened on live more than once

**When did the problem start happening**:

Likely universal damage

**Extra information**:

Happened today on live 16/02/2020

Ty-Omaha:
the INDESTRUCTIBLE flag is for damage only and not deconstruction which is why it still could be deconstructed

NODECONSTRUCT is for deconstruction with tools which is what I assume this is about

PR #13042, opened 2020-02-28
Fixes nukie shuttle computer being destructible & others

Ty-Omaha:
Fixes the nukie shuttle computer and a whole other special bunch like ERT and Vox from being able to be deconstructed.
If they have the indestructible flag chances are they shouldn't be deconstructed with a screwdriver either.

fixes: #12991

the INDESTRUCTIBLE flag is for damage only and not deconstruction which is why it still could be deconstructed

:cl:
fix: Special shuttles such as those for nuclear operatives can no longer be deconstructed with tools.
/ :cl:


Ty-Omaha:
#13042

Issue #12998, opened 2020-02-17
Cult Sac Objectives Can Target Offstation Roles


Kyep:
**Issue Description**:
CC officer on z2 was selected as cult sac target when one cryoed, despite having mind.offstation_role = 1.

**What did you expect to happen**: 
A crew member to be picked as the sac objective target.

**What happened instead**:
A mob with mind.offstation_role==1 was picked instead.

**Why is this bad/What are the consequences**:
Cult can get sac objectives which target mobs they cannot possibly get to.
Such mobs may have vox/etc names which don't make it obvious they're offstation roles.
So cult can spend entire round searching for mysterious voxxies who cannot be found anywhere.

**Steps to reproduce the problem**:
Cryo the cult sac target.
Have a offstation role character exist and be the only human.
Cult obj should target the offstation role.

Cazdon:
`/datum/game_mode/cult/proc/get_possible_sac_targets()
	var/list/possible_sac_targets = list()
	for(var/mob/living/carbon/human/player in GLOB.player_list)
		if(player.mind && !is_convertable_to_cult(player.mind) && (player.stat != DEAD))
			possible_sac_targets += player.mind
	if(!possible_sac_targets.len)
	//There are no living Unconvertables on the station. Looking for a Sacrifice Target among the ordinary crewmembers
		for(var/mob/living/carbon/human/player in GLOB.player_list)
			if(is_secure_level(player.z)) //We can't sacrifice people that are on the centcom z-level
				continue
			if(player.mind && !(player.mind in cult) && (player.stat != DEAD))//make DAMN sure they are not dead
				possible_sac_targets += player.mind
	return possible_sac_targets`


looks like it only checks if 'is_secure_level' if they're already going through the list of convertables, and considering the mindshield puts them in the list of non-convertables, it searches for them too. Should be a simple fix.

PR #13046, opened 2020-02-28
Fixes cult sacrifice targeting offstation roles

Ty-Omaha:
Stops cult sacrifice from targeting offstation roles

Fixes: #12998 

:cl:
fix: Cult sacrifice objective will no longer target offstation roles
/ :cl:

### DIFF
--- a/code/game/gamemodes/cult/cult_objectives.dm
+++ b/code/game/gamemodes/cult/cult_objectives.dm
@@ -155,12 +155,12 @@
 /datum/game_mode/cult/proc/get_possible_sac_targets()
 	var/list/possible_sac_targets = list()
 	for(var/mob/living/carbon/human/player in GLOB.player_list)
-		if(player.mind && !is_convertable_to_cult(player.mind) && (player.stat != DEAD))
+		if(player.mind && !is_convertable_to_cult(player.mind) && (player.stat != DEAD) && (player.mind.offstation_role != TRUE) )
 			possible_sac_targets += player.mind
 	if(!possible_sac_targets.len)
 	//There are no living Unconvertables on the station. Looking for a Sacrifice Target among the ordinary crewmembers
 		for(var/mob/living/carbon/human/player in GLOB.player_list)
-			if(is_secure_level(player.z)) //We can't sacrifice people that are on the centcom z-level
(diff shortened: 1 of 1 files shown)


datlo:
Unfortunately we are likely going to have the same issue with this PR as we did when I added that check, some stationside crewmembers are apparently getting the offstation_role var set to true despite not being ghost roles. I still dont know how and when this happens.

datlo:
I misread the pr and the bug i mentionned will only have a minor effect in game, its fine

code/game/gamemodes/cult/cult_objectives.dm
```diff
@@ -155,12 +155,12 @@
 /datum/game_mode/cult/proc/get_possible_sac_targets()
 	var/list/possible_sac_targets = list()
 	for(var/mob/living/carbon/human/player in GLOB.player_list)
-		if(player.mind && !is_convertable_to_cult(player.mind) && (player.stat != DEAD))
+		if(player.mind && !is_convertable_to_cult(player.mind) && (player.stat != DEAD) && (player.mind.offstation_role != TRUE) )
```
Fox-McCloud, line 158: Not the proper check. This is a boolean, so explicitly checking for a specific value is unnecessary.

code/game/gamemodes/cult/cult_objectives.dm
```diff
@@ -155,12 +155,12 @@
...
 			possible_sac_targets += player.mind
 	if(!possible_sac_targets.len)
 	//There are no living Unconvertables on the station. Looking for a Sacrifice Target among the ordinary crewmembers
 		for(var/mob/living/carbon/human/player in GLOB.player_list)
-			if(is_secure_level(player.z)) //We can't sacrifice people that are on the centcom z-level
+			if(is_secure_level(player.z) || player.mind.offstation_role == TRUE) //We can't sacrifice people that are on the centcom z-level or offstation roles
```
Fox-McCloud, line 163: see above.

code/game/gamemodes/cult/cult_objectives.dm
Ty-Omaha: I put it this way because I recall one time I was trying something without the specific value and it didn't work, so did aa but I'll change it. I'll change it anyways I guess

PR #13051, opened 2020-02-28
Changelog Overhaul

AffectedArc07:
## What Does This PR Do
This PR completley overhauls our changelog system to use a database backend instead of YML files. This will allow the changelog to be auto updated each round by the game without having to run a changelog bot. Users still make their changelogs in the existing format and the webhook processor just throws the data into a database instead of chucking it into a YML file. When you load the changelog ingame, it shows you items from the last 30 days. If needed I can write a separate PHP script to be ran on the webserver to view a changelog that is older than 30 days.

The new changelog is now grouped by PR number not date + author, and the date is the day the PR was merged, not the date the generator was ran. The game now also stores your last viewed changelog as the timestamp of the last entry you looked at, not an md5 of the changelog file.

## Why It's Good For The Game
Since alffd stepped down we havnt had a fully automatic changelog, and it has become an issue, so this will now auto create from the github processor automatically.

## Images of changes
image

## Note
All current changelog entries have been archived, and the archived file will have to be linked to seperately due to only showing dates of when the script was ran, which doesnt allow for it to be grouped by PR, and is thus not compatible with the new system.

## Changelog
Preferably install the new generator script before merging this PR so that there is a new CL entry in the new changelog system.
:cl: AffectedArc07
add: Completely overhauled the game changelog system
/:cl:


SQL/updates/10-11.sql
```diff
@@ -38,4 +38,4 @@ ALTER TABLE feedback.population CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4
 ALTER TABLE feedback.privacy CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 ALTER TABLE feedback.vpn_whitelist CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 ALTER TABLE feedback.watch CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
-ALTER TABLE feedback.whitelist CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
AffectedArc07: Ignore this, its just an artefact of conflict resolving

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+			return "<i class='fas fa-font'></i>"
+		if("EXPERIMENT")
+			return "<i class='fas fa-exclamation-triangle'></i>"
+		// No default here because its not possible to be anything else unless the DB schema changes
+
+// This proc is the star of the show
+/datum/controller/subsystem/changelog/proc/GenerateChangelogHTML()
+	// Modify the code below to modify the header of the changelog
+	var/changelog_header = {"
+		<link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.1/css/all.min.css" rel="stylesheet">
```
AffectedArc07: This probably needs to be local CSS, but for that I need to figure out how the asset caching system works

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+		return ..()
+
+	while(latest_cl_date.NextRow())
+		current_cl_timestamp = latest_cl_date.item[1]
+
+	if(!GenerateChangelogHTML()) // if this failed to generate
+		to_chat(world, "<span class='alert'>WARNING: Changelog failed to generate. Please inform a coder/server dev</span>")
+		return ..()
+
+	ss_ready = 1
```
farie82: TRUE

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+
+/datum/controller/subsystem/changelog/Initialize()
+	// This entire subsystem relies on SQL being here.
+	if(!config.sql_enabled)
+		return ..() 
+	
+	var/DBQuery/latest_cl_date = dbcon.NewQuery("SELECT UNIX_TIMESTAMP(date_merged) AS ut FROM [format_table_name("changelog")] ORDER BY date_merged DESC LIMIT 1")
+	if(!latest_cl_date.Execute())
+		var/err = latest_cl_date.ErrorMsg()
+		log_game("SQL ERROR during SSchangelog initialization L24. Error: \[[err]\]\n")
```
farie82: `log_and_message_admins` can be used for this

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+	while(latest_cl_date.NextRow())
+		current_cl_timestamp = latest_cl_date.item[1]
+
+	if(!GenerateChangelogHTML()) // if this failed to generate
+		to_chat(world, "<span class='alert'>WARNING: Changelog failed to generate. Please inform a coder/server dev</span>")
+		return ..()
+
+	ss_ready = 1
+	// Now we can alert anyone who wanted to check the changelog
+	for(var/client/C in startup_clients_button)
```
farie82: if you're sure the list only contains clients then you can loop through it using `var/thingy` and then cast it in the loop scope. Bit faster

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+		var/pr_block = "" // HTML for the changelog section
+		var/author = "" // Author of the PR
+		var/merge_date = "" // Timestamp of when the PR was merged
+		
+		// Now we gather the data from the DB
+		// Also we probably dont need to sanitize the PR number but you never know
+		var/DBQuery/pr_meta = dbcon.NewQuery("SELECT author,DATE(date_merged) AS date FROM changelog WHERE pr_number = [sanitizeSQL(pr_number)] LIMIT 1")
+		if(!pr_meta.Execute())
+			var/err = pr_meta.ErrorMsg()
+			log_game("SQL ERROR during CL generation L190. Error: \[[err]\]\n")
```
farie82: Same

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+
+		while(pr_meta.NextRow())
+			author = pr_meta.item[1]
+			merge_date = pr_meta.item[2]
+
+		// Now for each actual entry
+		var/DBQuery/db_entries = dbcon.NewQuery("SELECT cl_type, cl_entry FROM changelog WHERE pr_number = [sanitizeSQL(pr_number)]")
+		if(!db_entries.Execute())
+			var/err = db_entries.ErrorMsg()
+			log_game("SQL ERROR during CL generation L204. Error: \[[err]\]\n")
```
farie82: Same

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+
+	This SS also handles the checking of player CL dates and informing them if it has changed
+
+*/
+
+SUBSYSTEM_DEF(changelog)
+	name = "Changelog"
+	flags = SS_NO_FIRE
+	var/current_cl_timestamp = "0" // Timestamp is seconds since UNIX epoch (1st January 1970)
+	var/ss_ready = 0 // Is the SS ready? We dont want to run procs if we have not generated yet
```
farie82: FALSE

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+		if(C.prefs.toggles & UI_DARKMODE)
+			winset(C, "rpane.changelog", "background-color=#40628a;text-color=#FFFFFF")
+		else
+			winset(C, "rpane.changelog", "background-color=none;text-color=#000000")
+
+
+/datum/controller/subsystem/changelog/proc/OpenChangelog(client/C)
+	// If SQL is enabled but we aint ready, queue them up
+	if(!ss_ready)
+		startup_clients_open += C
```
farie82: If you press multiple times you'll be added multiple times to the list. Might want to use |= here

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+			return "<i class='fas fa-minus'></i>"
+		if("IMAGEADD")
+			return "<i class='fas fa-folder-plus'></i>"
+		if("IMAGEDEL")
+			return "<i class='fas fa-folder-minus'></i>"
+		if("SPELLCHECK")
+			return "<i class='fas fa-font'></i>"
+		if("EXPERIMENT")
+			return "<i class='fas fa-exclamation-triangle'></i>"
+		// No default here because its not possible to be anything else unless the DB schema changes
```
farie82: Maybe add a default case with a log. Just to catch some database corruptions etc

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+	// Technically this is only for the date but we can also do the UI button at the same time
+	if(preferences_datums[C.ckey].toggles & UI_DARKMODE)
+		winset(C, "rpane.changelog", "background-color=#40628a;font-color=#ffffff;font-style=none")
+	else
+		winset(C, "rpane.changelog", "background-color=none;font-style=none")
+	C.prefs.lastchangelog = current_cl_timestamp
+	var/DBQuery/updatePlayerCLTime = dbcon.NewQuery("UPDATE [format_table_name("player")] SET lastchangelog='[sanitizeSQL(current_cl_timestamp)]' WHERE ckey='[C.ckey]'")
+	if(!updatePlayerCLTime.Execute())
+		var/err = updatePlayerCLTime.ErrorMsg()
+		log_game("SQL ERROR during lastchangelog updating. Error: \[[err]\]\n")
```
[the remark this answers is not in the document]
AffectedArc07: This will only fire off the error message there are problems with the DB, and every other DB query is handled like this

code/controllers/subsystem/changelog.dm
```diff
@@ -0,0 +1,220 @@
...
+			<p><a href='[config.forum_link_url]'>Forum</a> - <a href='[config.wikiurl]'>Wiki</a> - <a href='[config.githuburl]'>GitHub</a></p>
+		</center>
+	"}
+
+	var/list/prs_to_process = list()
+	// Grab all from last 30 days
+	var/DBQuery/pr_list_query = dbcon.NewQuery("SELECT DISTINCT pr_number FROM changelog WHERE date_merged BETWEEN NOW() - INTERVAL 30 DAY AND NOW()")
+	if(!pr_list_query.Execute())
+		var/err = pr_list_query.ErrorMsg()
+		log_game("SQL ERROR during CL generation L143. Error: \[[err]\]\n")
```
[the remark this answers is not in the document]
AffectedArc07: Again here, it only errors and this if the DB server shat itself, and its the same as it is everywhere else in the code

code/controllers/subsystem/changelog.dm
AffectedArc07: Same as I have said above, though remind me to add a default for this incase no changelogs exist

code/controllers/subsystem/changelog.dm
AffectedArc07: Same as I have said

code/controllers/subsystem/changelog.dm
AffectedArc07: The list *should* only ever contain clients so ill do that

code/controllers/subsystem/changelog.dm
AffectedArc07: Same as I have said

AffectedArc07:
Closing & Reopening to make sure I dont have any CRLF

Ty-Omaha:
I like it, maybe some color to the different icons to make it easy to see what is what

AffectedArc07:
image
Added some colour and hoverable text to the items

AffectedArc07:
Pushed a new commit to kick travis into working

code/controllers/subsystem/changelog.dm
AffectedArc07: I have discovered why this is needed.

TODO: Add fontawesome webfont files locally

tools/githubChangelogProcessor.php
```diff
@@ -0,0 +1,347 @@
...
+
+//Github lets you have it sign the message with a secret that you can validate. This prevents people from faking events.
+//This var should match the secret you configured for this webhook on github.
+//This is required as otherwise somebody could trick the script into leaking variables.
+$hookSecret = '08ajh0qj93209qj90jfq932j32r';
+
+$dbServer = "localhost"; // Hostname of the database server (default localhost)
+$dbUser = "root"; // Database username (default root)
+$dbPassword = ""; // Database password (default blank)
+$dbDatabase = "feedback"; // Database name (default feedback)
```
Kyep, line 31: This requires an additional setting: port.
We use a non-standard port for our mysql connection, so the script needs a config setting for this.

tools/githubChangelogProcessor.php
```diff
@@ -0,0 +1,347 @@
...
+//CONFIG START (all defaults are random examples, do change them)
+//Use single quotes for config options that are strings.
+
+//Github lets you have it sign the message with a secret that you can validate. This prevents people from faking events.
+//This var should match the secret you configured for this webhook on github.
+//This is required as otherwise somebody could trick the script into leaking variables.
+$hookSecret = '08ajh0qj93209qj90jfq932j32r';
+
+$dbServer = "localhost"; // Hostname of the database server (default localhost)
+$dbUser = "root"; // Database username (default root)
```
Kyep, line 29: This requires a comment describing the permissions that the user needs. I assume it only needs all permissions to one table, 'changelog', and no permissions to any other table.

html/archived_changelog/changelog.html
```diff
@@ -25,7 +25,8 @@
 		<td valign='top'>
 			<div align='center'><font size='3'><b>Paradise Station</b></font></div>
 			
-			<p><div align='center'><font size='3'><a href="https://www.paradisestation.org/forum/">Forum</a> | <a href="https://www.paradisestation.org/wiki/index.php/Main_Page">Wiki</a> | <a href="https://github.com/ParadiseSS13/Paradise">Source</a></font></div></p>
+			<p><div align='center'><font size='3'><a href="https://www.paradisestation.org/forum">Forum</a> | <a href="https://www.paradisestation.org/wiki/index.php/Main_Page">Wiki</a> | <a href="https://github.com/ParadiseSS13/Paradise">Source</a></font></div></p>
+			<font size='2'><b>Visit our Discord channel:</b><a href='https://discordapp.com/invite/nuqD478<'>-Click Here-</a></font>
```
Kyep, line 29: Remove the Discord invite link. It is outdated and the actual link changes every once in awhile, we don't want to have static links in the HTML.

PR #13054, opened 2020-02-29
adds a cyborg APC adaptor for cyborgs.

Aquilar:
## What Does This PR Do

as title
~~originally it was intended just for ERT and syndicate borgs, I couldn't find a way to add to either without the solution being a hacked mess especially since ERT borgs use standard modules and not their own. So now it's just an expensive upgrade from the mechfab.~~
ports over TG borg charger with the changes needed for it to work.

## Why It's Good For The Game

it's an upgrade module allowing cyborgs to draw power from apcs and power cells should a charging station be unavailable. it can also recharge powercells

## Images of changes

charger

## Changelog
:cl:
add: cyborg apc power adaptor and a sprite for it.
/:cl:


code/modules/mob/living/silicon/robot/robot_items.dm
```diff
@@ -95,4 +95,59 @@
...
+	while(do_after(H, 10, target = A))
+		if(loc != H)
+			to_chat(H, "<span class='warning'>You must keep your connector out while charging!</span>")
+			break
+		if(A.cell.charge == 0)
+			to_chat(H, "<span class='warning'>\The [A] has no more charge.</span>")
+			break
+		A.charging = 1
+		if(A.cell.charge >= 200)
+			H.cell.charge += 200 * H.cell.rating // replenishment scales with higher capacity cells
```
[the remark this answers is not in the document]
Aquilar: the borg getting more power based on cell rating is to keep the charging times consistent, otherwise it'll be standing there a very long time.

code/modules/mob/living/silicon/robot/robot_items.dm
```diff
@@ -95,4 +95,59 @@
...
+		if(loc != H)
+			to_chat(H, "<span class='warning'>You must keep your connector out while charging!</span>")
+			break
+		if(A.cell.charge == 0)
+			to_chat(H, "<span class='warning'>\The [A] has no more charge.</span>")
+			break
+		A.charging = TRUE
+		if(A.cell.charge >= 200)
+			H.cell.give(200 * H.cell.rating) // replenishment scales with higher capacity cells
+			A.cell.use(200) // as to not rapidly drain and depower rooms
```
Crazylemon64: This would be better implemented as a subtype of `apc_powercord`, where you replace the actions in this part of the loop with a subtype-specific action - for the human version give "nutrition", but for the borg version provide cell charge

Aquilar: I would have to refactor apc_powercord just to accommodate the new item, which beyond the scope of this pr.

Crazylemon64: "Refactor" isn't a scary word. All you need to do is add a couple of more procs to make the behavior match between what you copy pasted it from.
https://gist.github.com/Crazylemon64/2f16dd661944326562e5f4d5059d67cc

Granted, now that you've changed your implementation from a copypasta to a proper port, this isn't directly applicable, but keep in mind: When you do a copypasta, this kind of "refactor" is expected. It is definitely NOT out of scope.

Crazylemon64: The reason why we want these refactors on copypasta'd code is to avoid having multiple implementations for the same behavior. Say that we want, as a hypothetical example, the charger to hack the user if used on an emagged APC. We would need to implement it on each copy of the cable, meaning more places to check, and more chances to make mistakes. A similar argument exists for bugfixing - if you have many copies, you might only fix a bug on a single instance when it exists on them all.

Aquilar: sorta shows my relative inexperience and I'll keep that in mind for the future.

Fox-McCloud:
I'm not really sure I like this idea, to be honest.

I'd much rather see something like the power inducer implemented, than this--as that as a lot broader use and requires more input+help to get charged quickly. This is more just "self sustain in the field anywhere".

code/modules/mob/living/silicon/robot/robot_items.dm
```diff
@@ -95,4 +95,132 @@
...
+					break
+
+				if(!user.cell.give(150))
+					break
+
+				M.use_power(200)
+
+			to_chat(user, "<span class='notice'>You stop charging yourself.</span>")
+
+		else if(is_type_in_list(target, charge_items)) // drawing power from powercells in side objects
```
Crazylemon64: `is_type_in_typecache`, since you declare those lists with `typecacheof`

TDSSS:
It would be very useful for ERT borgs, atl east. Lack of chargers can be a big issue for them.

Aquilar:
> I'd much rather see something like the power inducer implemented, than this--as that as a lot broader use and requires more input+help to get charged quickly. This is more just "self sustain in the field anywhere".

it's a pretty low power self sustain if you can't get a hold of any loose power cells. 
The inducer sounds fun to port.

> It would be very useful for ERT borgs, atl east. Lack of chargers can be a big issue for them.

There's no easy way to add it to ERT borgs without either duplicating the modules to be separate from the stations borgs, which is to be avoided, or having the upgrade action applied on picking a module as ERT, which is probably not advised.

I'll just comment out the design for now and just leave it in case anyone wants it.

Denthamos:
Gonna agree with @Fox-McCloud the inducer would work better because borgs should not be able to handle everything on their own without proper humanoid interaction.

However, I can see this working very well with syndie borgs for nuclear operatives since they have to currently rely on their ship recharger or station rechargers.  I am not sure about ERT borgs getting this due to Kyet's current changes he is implementing to the ERT ones as it could create a power bloat with them.

PR #13056, opened 2020-02-29
Cyborg Improvements

### LINKED ISSUE #11657: Death Squad borgs are vastly underpowered
**Problem Description**:
What is the problem?
Death squad borgs lack several features that you would expect, making them almost inferior to gamma borgs.
Death Squad borgs should be at least as competent as their organic counterparts.
**What happened instead**:
No energy shield to prevent friendly fire;
No self-repair;
No self-recharge to counteract incredible power loss;
As vulnerable to damage as a regular borg
Also still incredibly vulnerable to ion (not as important)
As stated above, picking borg should not feel like it's actively harming the DS.
Step 1. Pick DS borg
Step 2. Get on station
Step 3. Watch as you die almost immediately
Forever afaik
Not really

### LINKED ISSUE #13055: Some ERT borg module pick runtimes
**Issue Description**:

An ERT borg choosing either the engineering or medical module gets the following runtimes:
image

Module picking runtimes halfway through the module picking code which means the ERT borgs never get their deparment sprite and will instead will be the standard grey borg. Also some important code never executes like the engineering borg getting its "magboot" setting, for example.

1. Put yourself into a `mob/living/silicon/robot/ert` borg
2. Try to choose either the medical or engineering module

Kyep:
Mining Borgs:
- Mining (and other) borgs are now immune to ash storms on lavaland. This was necessary for them to be viable, as they have no way to generate shelters like organic miners can, which results in them dying very quickly to ash storms.
- Mining borgs can now get a "lavaproof chassis" upgrade from Robotics which lets them walk on lava (ported from TG, requires tech and mats).
- Mining borgs now have access to a dark lavaland-themed sprite (ported from TG)

Crew Borgs:
- The "Standard" robot module has been reworked to be viable. It is now called "Generalist". It gets basic gear for every department, but doesn't specialize in any. For example, it can clean with soap and replace broken lights like a jani-borg, but doesn't clean as it moves. It can extinguish fires and do basic repairs like an engi-borg, but lacks an RCD. It can diagnose and treat basic injuries like a medi-borg, but cannot do surgery. Etc. Overall, it should now be useful enough in a variety of situations to be worth picking at least occasionally.
- Janitor borgs now have a mini-extinguisher, so they don't get permanently stuck in space or zero gravity.
- The number of security borgs is now limited. At roundstart (green alert) you're limited to a single security borg. At higher alert levels, you can only have two at most. ERT borgs, emagged borgs, and borgs with the "safety override" upgrade can bypass this limit.

Special Borgs:
- Fixed bug where ERT medical/engi borgs were not getting sprites, or cameras, and were generating a runtime error. Fixes #13055 
- Power cells in ERT/DS borgs now scale based on how likely it is that the station will have working rechargers when the borgs get there. Amber ERT get super cells. Red ERT get hyper cells. Gamma/DS get bluespace cells. This also has the side effect of solving balance issues whereby ERT borg weapons/items were balanced with one cell in mind, but then become overpowered when science upgrades their cell.
- ERT borgs now always have their ERT level in their name. E.g: "Amber ERT 934", "Red ERT 123", "Gamma ERT 834", etc.
- DS Borgs now have permanent eshields that reflect all incoming laser-based fire, including pulse lasers. Without this, they always get killed by stray pulse shots from the organic DS officers (who already have eshields that reflect energy projectiles, and are not in any way careful to avoid hitting their own borgs with their shots). Fixes #11657
- Reworked Gamma borg completely. It is now a specialist borg, intended for assaulting entrenched blobs/xenos/terrors. Its primary weapon is similar to the gamma sec ERT, with the benefit of way more shots, but also the cost of taking way longer to recharge after depletion. Unlike organic ERT, borgs cannot pick up extra weapons from science either. This borg has a new sprite, no roll ability, and its energy shields have been replaced with damage reduction armor. Overall it is much stronger, but due to the fact **crew borgs can no longer become gamma borgs, even on gamma alert**, it is also much rarer. It is intended to be an absolute last resort if admins are trying to send something to save the station, as opposed to just sending DS to nuke it.
- Moved the "rolling borg" sprite/concept/module to a new borg type, /destroyer borgs. These are the borg equivalent of the CC Special Operations Officer, and as such are admin-only. They're immune to almost everything that can stun a borg, including EMP. They've also been refactored to eliminate snowflake code, weird mechanics, sprite glitches, etc. 
- Added the "damage reduction" mechanic for borgs, which reduces all incoming damage by a set number. Only DS/Gamma/Destroyer borgs get any of this. It replaces the old combat shield mechanics of gamma borgs as well, for simplicity.

Misc:
- The "illegal equipment" upgrade has been renamed to "safety override", to be clear about what it does, and stop implying the upgrade is illegal when it isn't.
- The number of roundstart cyborg jobs has been increased from 1 to 2, to account for the greater variety of viable borg types now, such as mining and standard borgs. In addition, the number of spawn points for cyborgs on box/meta/delta has been increased from 1 to 4, so that if we ever decide to increase their job slots in future, or run an event round with a higher-than-normal number of cyborg slots, these things can be done without the need for further map edits.
- Refactored some vars that were 1/0 to be TRUE/FALSE instead.
- Fixes a bug where borgs using non-disabler energy weapons (like the sec borg's laser gun) who fully deplete both the borg's cell and the gun's cell can end up with a permanently broken gun that will never recharge or be usable again.

## Images of changes
new_mining_borg_sprite
cyborgstart_box
cyborgstart_meta
cyborgstart_delta
ertgamma_small

## Changelog
:cl:
add: Mining cyborg module can now mine in ash storms, be upgraded with lava walking by science, and use TG mining borg sprites.
add: Standard cyborg module replaced by new (viable) generalist module.
add: DS borgs now reflect lasers like organic DS officers with eshields.
add: Crew borgs can no longer respec to gamma borgs during gamma alert.
add: Sec borgs are normally capped to 1 at roundstart/green. 2 at higher alert levels.
add: Gamma ERT borgs extensively reworked. 
tweak: Added additional roundstart cyborg job slot.
tweak: Illegals upgrade for borgs renamed to safety override upgrade.
refactor: Improved code quality for cyborgs.
fix: Fixed runtimes related to cyborgs.
tweak: Many other minor tweaks to cyborgs.
fix: Fixed a bug where cyborg laser weapons (e.g. energy gun) could permanently break when borg+cell energy was completely depleted.
/:cl:


Cazdon:
Always nice to see cyborgs getting some love.

MRgaty:
This is awesome

TDSSS:
What exactly is a "weapons-unlocked borg"? Is it one with the safety override module?

I am also a bit concerned about the number of mining borg upgrades in the fabricator compared to the other borgs. Feels like most of the list is just miner stuff.

datlo:
Interesting that standard borgs get atmos holofans while IIRC engiborgs dont. I get the logic, tho. 
:+1:

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1307,56 +1316,69 @@ var/list/robot_verbs_default = list(
...
 	icon_state = "droidcombat"
 	modtype = "Combat"
 	designation = "Combat"
+	default_cell_type = /obj/item/stock_parts/cell/bluespace
 
 /mob/living/silicon/robot/combat/init()
 	..()
 	module = new /obj/item/robot_module/combat(src)
 	module.channels = list("Security" = 1)
-	//languages
```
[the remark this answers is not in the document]
Kyep: Because things like module.add_languages, module.add_subsystems_and_actions, etc are obvious and don't need comments to explain what they do.

Kyep:
@TDSSS weapons_unlock means ERT, or station borg with the "safety override" upgrade. If you have suggestions for other borgs, message me on Discord. I'm willing to consider them, but (A) they'll probably be another PR at this point, and (B) I don't want to buff things like the sec borg or medical borg.
@datlo I have since removed the atmos holobarrier projector from the standard borg, and instead given them subsystems and metal/rod/cable/tile stacks. This enables them to do something about breaches, as before, as well as gives them something to do on green when nothing is happening (construction) as well as avoiding giving them an engi tool that engi borgs do not get.
@AffectedArc07 Addressed all your suggestions.

lolzyking:
Mining borgs lack a method of removing Kinetic accelerator upgrades.

TDSSS:
You wrote:
>Exceptions are made for the ERT, emagged borgs, and weapons-unlocked borgs.
and now you say
>weapons_unlock means ERT borg, emagged borg, or borg with the "safety override" upgrade.

So shouldn't you just say 'Exceptions are made for ert, emagged or safety override borgs'? Or alternatively, exceptions are made for weapons_unlock borgs. As is you list two things, then a third category that includes the first two.

Kyep:
I could, but not everyone knows that ERT borgs get the safety override upgrade by default.

Cazdon:
@lolzyking Roboticists have to take them out for you.

code/game/objects/items/robot/robot_upgrades.dm
```diff
@@ -193,16 +193,27 @@
...
-
 	R.emagged = 1
+	return 1
 
+/obj/item/borg/upgrade/lavaproof
+	name = "mining cyborg lavaproof chassis"
+	desc = "An upgrade kit to apply specialized coolant systems and insulation layers to a mining cyborg's chassis, enabling them to withstand exposure to molten rock."
+	icon_state = "ash_plating"
+	resistance_flags = LAVA_PROOF | FIRE_PROOF
+	require_module = 1
```
farie82: TRUE

code/game/objects/items/robot/robot_upgrades.dm
```diff
@@ -193,16 +193,27 @@
...
+	module_type = /obj/item/robot_module/miner
+
+/obj/item/borg/upgrade/lavaproof/action(mob/living/silicon/robot/R)
+	if(..())
+		return
+	if(istype(R))
+		R.weather_immunities += "lava"
 	return 1
```
farie82: TRUE

AffectedArc07:
IconDiffBot got stuck in a loop processing, pushing another commit should kick it into working.

Quantum-M:
I don't think the generalist module should get all of those different radios, as they only do a small portion/minimal portion of each of the different departments, especially with the security radio, as they only get a baton and, if emagged, a energy sword.

Kyep:
The reason that radio was added is that, during testing, generalist borgs spent a lot of time wandering the station, trying to figure out where they could be helpful. Letting them hear those comms enables them to identify opportunities to help, and go there. E.g: engineering building a BSA room, medical having an influx of patients, security getting swarmed by hostiles in brig, etc.

If generalists turn out to be OP, I will nerf them, but I don't want to nerf them by making it harder for them to find anything to do - that just makes them un-fun to play. If I need to nerf them, I will nerf them in some other way that doesn't go against the flexible generalist nature they're supposed to have.

Aquilar:
I would suggest giving the new combat module a disabler for crowd control.
and I would suggest giving the new destroyer module a hybrid taser for similar reasons.

variableundefined:
@Kyep 
Commenting on design (Haven't reviewed actual code / sprites) only: 
- Mining borgs looks good. 
Crew Borgs: 
- Standard changes looks good 
- Janitor Borgs: looks good 
- Security Borgs: I am highly biased toward Security borgs and addicted to playing as them, but I can 100% understand why more than 2 secborg is cancer. So looks good to me.

Power cells upgrade: Seems reasonable to me
Bugfixes: No objection

DS Borgs: It's DS, so buffing it should be fine

Gamma Borg: Moving them away from Crew borg make sense as I remember time where I'd stay as a standard borg, waiting for Gamma to be autotriggered by nuclear disk being stolen, and then become a Gamma borg to rip the nuke op team apart.

Destroyer / Rolling Borg: I am somewhat biased toward droideka gamma borg and would like it retained for Gamma borg. 

Damage Reduction: Probably less obtuse than turning on the shield and occupying one slot for damage reduction. I'll need to see how they actually work.

Illegal Equipment rename: Make senses

Roundstart Cyborg Job: This improve my odd of giving the playerbase cancer by playing roundstart security borg by +100%, naturally, I approve.

Refactoring Vars: duh, no objection.

Aquilar:
thinking about it more, the combat and destroyer borgs could really do with wrench so they can move stuff out of the way after busting down walls with the jackhammer, for the same token, the "generalist" would also need a wrench in case it screws up building.

lolzyking:
Another borg quality of life  change I've seen on other servers, the borg shell unlocking on borg death, so the MMI, Robobrain or posibrain can be removed, stolen, or interrogated OR the body can be resynced back or removed from the AI then repaired.

Kyep:
@Aquilar @lolzyking I don't want to add any more significant features to the PR at this point. I might add a wrench, maybe, but I'm skeptical about enabling anyone to steal the brain of any dead borg.

Kyep:
Update: added a fix which prevents borgs that use non-disabler energy weapons from permanently breaking their gun in some situations. This bug existed before this PR, but this PR now fixes it.

code/datums/weather/weather_types/ash_storm.dm
```diff
@@ -78,6 +78,8 @@
 	while(L && !isturf(L))
 		if(ismecha(L)) //Mechs are immune
 			return TRUE
+		if(issilicon(L))
+			return TRUE
```
Fox-McCloud, line 82: icky way to do it.

Have `weather_immunities = list("ash")` on all silicons.

code/game/objects/items/robot/robot_upgrades.dm
```diff
@@ -181,35 +181,46 @@
...
+	resistance_flags = LAVA_PROOF | FIRE_PROOF
+	require_module = TRUE
+	module_type = /obj/item/robot_module/miner
 
-	return 1
+/obj/item/borg/upgrade/lavaproof/action(mob/living/silicon/robot/R)
+	if(..())
+		return
+	if(istype(R))
+		R.weather_immunities += "lava"
```
Fox-McCloud, line 216: need to make sure this is properly reset when the module is removed.

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -118,20 +118,42 @@
...
+		/obj/item/stack/sheet/metal/cyborg = 50,
+		/obj/item/stack/cable_coil/cyborg = 50,
+		/obj/item/stack/rods/cyborg = 60,
+		/obj/item/stack/tile/plasteel = 20
+		)
 
 /obj/item/robot_module/standard/New()
 	..()
+	// sec
 	modules += new /obj/item/melee/baton/loaded(src)
```
Fox-McCloud, line 138: I'd rather we just do what TG has done here, with this set of items;

```
		/obj/item/assembly/flash/cyborg,
		/obj/item/reagent_containers/borghypo/epi,
		/obj/item/healthanalyzer,
		/obj/item/weldingtool/largetank/cyborg,
		/obj/item/wrench/cyborg,
		/obj/item/crowbar/cyborg,
		/obj/item/stack/sheet/metal/cyborg,
		/obj/item/stack/rods/cyborg,
		/obj/item/stack/tile/plasteel/cyborg,
		/obj/item/extinguisher,
		/obj/item/pickaxe,
		/obj/item/t_scanner/adv_mining_scanner,
		/obj/item/restraints/handcuffs/cable/zipties,
		/obj/item/soap/nanotrasen
```

And do away with the stun baton. I'd rather not de-factor have the standard module become secborg 2.0

code/modules/projectiles/ammunition/energy.dm
```diff
@@ -68,20 +68,28 @@
 /obj/item/ammo_casing/energy/immolator
 	projectile_type = /obj/item/projectile/beam/immolator
 	fire_sound = 'sound/weapons/laser3.ogg'
-	e_cost = 125
+	e_cost = 100 // 10 shots for the gamma sec ERT who get this, used to be 8
```
Fox-McCloud, line 71: If you're making something for the borgs, this needs to be a subtype; this is impacting the balance of existing immolators.

Fox-McCloud, line 71: This should probably be a separate PR.

code/modules/projectiles/ammunition/energy.dm
```diff
@@ -68,20 +68,28 @@
...
 
 /obj/item/ammo_casing/energy/immolator/strong
 	projectile_type = /obj/item/projectile/beam/immolator/strong
-	e_cost = 125
+	e_cost = 100 // 10 shots for the gamma sec ERT who get this, used to be 8
```
Fox-McCloud, line 75: same

code/game/jobs/job/silicon.dm
```diff
@@ -24,8 +24,8 @@
 	title = "Cyborg"
 	flag = JOB_CYBORG
 	department_flag = JOBCAT_ENGSEC
-	total_positions = 1
-	spawn_positions = 1
+	total_positions = 2
+	spawn_positions = 2
```
Fox-McCloud, line 28: I'm inclined to say if you're going to do this, there should only ever be *ONE* security borg active at a time.

procdrone: for the population on the server we have, i think 1 is not enough.

Kyep: One is definitely not enough.

code/modules/mob/living/silicon/robot/robot_modules.dm
Kyep: Comparing mine to the TG borg:
1) mine has a stunbaton but no cuffs, theirs has cuffs but no stunbaton. I'm inclined to think mine is better, as the stunbaton has non-sec uses (like letting the borg defend itself from carp and other NPCs). Whereas the sole purpose of zipties is detaining people, for sec or antag purposes.
2) theirs has mining items (mining scanner, pickaxe).... which mine doesn't. It isn't a big deal, I just feel like there's no point in a generalist borg going to mining. 
3) theirs has more engi gear (they get a largetank welder, and a wrench)
4) mine's injector has saline, in addition to the epi theirs has. So a little better at medical. 

Overall, I don't think TG's version is better. 
I'm okay with changes 2 and 3. Though I worry they'll make the borg a bit too strong.
My worry is with changes 1 and 4. 1 I feel strips the borg of its primary ability to defend itself from stuff like carp. 4 I feel makes it nigh-useless at doing anything medical.

Fox-McCloud: Fair on the injector.

I'd go ahead and add the engineering and mining gear.

The reason the baton is more egregious than the cablecuffs is it's pretty easy to baton and drag someone all the way to security by repeatedly applying the baton before they get unstunned. Chances are you aren't going to completely drain your own cell with 4-5 stun baton whack---but you can effectively keep someone stunned for over a full minute under those circumstances.

code/game/jobs/job/silicon.dm
Fox-McCloud: You'd be surprised---a lot of AI's are pretty validy and specifically *order* their borgs to go security.

Objectively speaking, on average, there's going to be *more* sec borgs running around with this change than before.

Even if only 1 in 4 rounds the second borg goes security borg, that's still, on average, more than current.

AffectedArc07:
Looking through the icon diff checks
image
IDB has an issue with rogue pixels sometimes but none of the other images have them. Is this an icon-issue with a rogue pixel on a second frame, or just IDB being IDB

Fox-McCloud:
Looking at things more:

img

The borg sprite on the right really doesn't fit, at all.

The one on the left should just replace the other borg sprite with the tread---it has the old color scheme that matches the old shaft miner's purple+brown theme.

code/modules/mob/living/silicon/robot/robot_modules.dm
Kyep: Generalist borg loadout has been tweaked - it now gets cuffs instead of a baton, and gets engi/mining gear. I also gave it an ore satchel because without one it can't actually use its mining gear effectively.

code/game/jobs/job/silicon.dm
Kyep: I've adjusted the limit on secborgs down to 1 at roundstart, and on green alert. 2 on higher security alert levels. This will stop multiple roundstart cyborgs going secborg. I'm not willing to set the cap on secborgs to 1 at all times, even on red alert. I feel that goes too far. I hope this compromise will be enough.

Kyep:
Affected: that's an IDB problem - if you look at the sprite file, that sprite is unchanged.
Fox: the brown spider mining borg sprite has been removed.

Spacemanspark:
I personally don't see an issue with having both tread sprites @fox-mccloud
Nothing wrong with mixing a bit of old and new.

taukausanake:
Is there a reason the nano bloodhound sprite was edited? Looks like the only change is a white pixel on the top-left of the South sprite

PR #11860, opened 2019-07-15
Slime People can regrow limbs while both legs are missing. Reworks how the game treats legless people

SteelSlayer:
**What does this PR do:**
Adds a new status labeled "legless". When people missing both their legs fall over, legless is set to True. The person is no longer spammed with weakness forever. Having weakness or not didn't seem to matter to me, as the person couldn't move, get up, or interact with anything anyway. In that regard nothing has changed, legless people will still fall over and cannot move or do anything.

This allows Slime People who are missing both legs to regrow their limbs, regardless if they're buckled into something or not.

I tested this on my local server by amputating legs through surgery, hacking legs off with a weapon or using admin proc calls to delete the legs. Everything was working as intended in all those instances.

Fixes #10066
Fixes #11200

**Changelog:**
:cl:
add: Added the "legless" status effect to mobs who have both legs and both feet missing
add: Admin Rejuvenate sets legless to False, allowing people to move properly again
fix: Slime People can now regrow their limbs while missing both legs and/or resting
del: People missing both legs no longer receive weakness constantly
/:cl:

PR #13057, opened 2020-02-29
Adds support for multiple callback checks for the do_after and incapacitated procs

### LINKED ISSUE #10066: Slimes can't regrow limbs if they're missing both legs
**Problem Description**:
If a slime is missing both legs and are not buckled to a chair or roller bed or something, then they can't regrow any missing limbs. They get this error:

You begin to focus on regrowing your missing left leg... (This will take 45 seconds, and you must hold still.)

You need to hold still in order to regrow a limb!

To be able to regrow missing limbs while immobilized.

**What happened instead**:
Unable to regrow limbs while missing both legs.

It's a bit of a downer on the racial ability if it can't fix the exact sort of thing it's meant to be able to fix.

Be a slime, have both legs amputated or otherwise removed, and try to grow them back.

### LINKED ISSUE #11200: Limbless slime person can't regenerate without wheelchair
**Problem Description**:
What is the problem?
As slime person with no arms or legs, when on ground, or buckled to chair/bed, trying to regenerate limb results in message saying that You must stay still to do that. In result, ability loses part of its usefulness.
I expected to be able to regenerate my legs and arms, without having to be specifically on wheelchair. Buckling to bed, or chair did not fix problem
**What happened instead**:

Limb regeneration loses part of its usefulness.

SteelSlayer:
## What Does This PR Do
This is technically a remake of this PR: #11860. As such, it does what I tried to accomplish in that PR. Slime people can now regrow any missing limbs while they are currently missing both of their legs. Previously it would have given a message about how you needed to "hold still" to regrow, even though you couldn't move. This was because legless people get spammed with weakness constantly.

#### Code improvements needed for this:
Swaps `do_after()`'s `extra_checks` from a singular callback argument, to a list, which is designed to hold multiple callbacks. Also adds a new variable `use_default_checks` which is set to `TRUE` by default. I've removed the `user.stunned` and `user.IsWeakened()`in favor of using callbacks:
```js
if(use_default_checks)
    extra_checks += CALLBACK(user, /mob.proc/IsWeakened)
    extra_checks += CALLBACK(user, /mob.proc/IsStunned)
```
This means that all do_after's will still check for weakness and stunned by default, provided you don't set `use_default_checks` to `FALSE`.

Similarly `incapacitated()` now gets the same two arguments, and gains callback checks for stunned and weakened, just like do_after.

## Why It's Good For The Game
Allows for better customization of the `incapacitated` and `do_after` procs.

Fixes #10066
Fixes #11200

## Changelog
:cl:
fix: Slime people can now regrow their limbs while missing both legs. They will no longer get the message about needing to hold still.
/:cl:


code/__HELPERS/mobs.dm
```diff
@@ -346,62 +346,92 @@ This is always put in the attack log.
...
 	var/endtime = world.time + delay
 	var/starttime = world.time
-	. = 1
+	. = TRUE
+
+	// By default, checks for weakness and stunned get added to the extra_checks list.
+	// Setting `use_default_checks` to FALSE means that you don't want the do_after to check for these statuses, or that you will be supplying your own checks.
+	if(use_default_checks)
+		extra_checks += CALLBACK(user, /mob.proc/IsWeakened)
+		extra_checks += CALLBACK(user, /mob.proc/IsStunned)
```
Crazylemon64: Careful - `extra_checks` will have these callbacks added on every call of the function, and since `extra_checks` is a singleton (modifications to it it will be retained across function calls) it'll have more and more copies of these callbacks added for every call that uses `use_default_checks`.

Instead, you may want to have `extra_checks` be `null`, and then if not supplied, replace it with a list that contains these two callbacks.

code/modules/mob/living/update_status.dm
```diff
@@ -65,8 +65,14 @@
 	return !(IsWeakened() || paralysis || stat || (status_flags & FAKEDEATH))
 
 // Whether the mob is capable of actions or not
-/mob/living/incapacitated(ignore_restraints = FALSE, ignore_grab = FALSE, ignore_lying = FALSE)
-	if(stat || paralysis || stunned || IsWeakened() || (!ignore_restraints && restrained()) || (!ignore_lying && lying))
+/mob/living/incapacitated(ignore_restraints = FALSE, ignore_grab = FALSE, ignore_lying = FALSE, list/extra_checks = list(), use_default_checks = TRUE)
+	// By default, checks for weakness and stunned get added to the extra_checks list.
+	// Setting `use_default_checks` to TRUE means that you don't want it checking for these statuses or you are supplying your own checks.
```
Crazylemon64: I think you mean FALSE?

code/__HELPERS/mobs.dm
SteelSlayer: I'm not sure what you mean here, every time something calls do_after, `extra_checks` starts as a blank list, provided the coder didn't supply anything for the argument. But even then, it should be unique per proc call.

I just tested this out in game. I added some print statements for the amount of callbacks in `extra_checks`. Then I spammed clicked a thing which created uses a do_after a bunch of times. (tested with limb regrowing, but also cutting wires out of a SMES). The amount of callbacks in `extra_checks` was always 2.

```js
for(var/datum/callback/C in extra_checks)
    message_admins("[C]")
```

image

Crazylemon64: Oh - it looks like this isn't the case in BYOND, good. I think I might have had a wire crossed from python, where that pattern would cause the list to balloon in size

code/__HELPERS/mobs.dm
```diff
@@ -346,62 +346,87 @@ This is always put in the attack log.
...
 					break
 			if(user.get_active_hand() != holding)
-				. = 0
+				. = FALSE
 				break
 	if(progress)
 		qdel(progbar)
 
+// Upon any of the callbacks in the list returning TRUE, the proc will return TRUE.
+/proc/invoke_callback_list(list/extra_checks)
```
moxian: I suggest renaming this to `any_callback_true` or similar, because right now the semantics are unobvious, if you see this called in the code somewhere and don't bother checking the comment in the definition. I originally thought it would return true if *all* the callbacks return true, but this is not the case here.

SteelSlayer: That's a funky sounding proc name, but I guess its name is more telling of its purpose than my current name.

PR #13061, opened 2020-03-01
AI Cores are no longer able to fit in a locker

Quantum-M:
## What Does This PR Do

This PR prevents AI Cores (the mob kind with the AI in it) from being closed in a locker.

## Why It's Good For The Game

This is to prevent the cheese that can be done with the lockers, in which a AI would be closed in a locker, and they would be easily teleported via the teleporter (which the teleporter does not allow at all for AI cores to go through it).

Also a consideration that AI cores are larger than the lockers themselves.

## Images of changes

N/A

## Changelog
:cl: Quantum-M
tweak: AI Cores (the mob kind with the AI) are no longer able to fit inside lockers.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -46,6 +46,10 @@
 	for(var/obj/structure/closet/closet in get_turf(src))
 		if(closet != src && closet.anchored != 1)
 			return FALSE
+	for(var/mob/living/silicon/ai/ai in get_turf(src))
+		if(ai)
+			return FALSE
+
 	return TRUE
 
 /obj/structure/closet/proc/dump_contents()
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -46,9 +46,6 @@
 	for(var/obj/structure/closet/closet in get_turf(src))
 		if(closet != src && closet.anchored != 1)
 			return FALSE
-	for(var/mob/living/silicon/ai/ai in get_turf(src))
-		if(ai)
-			return FALSE
 
 	return TRUE
 
@@ -109,6 +106,8 @@
 		if(istype(M, /mob/living/simple_animal/bot/mulebot))
 			continue
 		if(M.buckled || M.anchored || M.has_buckled_mobs())
+			continue
+		if(isAI(M))
 			continue
 
 		M.forceMove(src)


Qwertytoforty:
Honestly, as much as everyone does this, the AI has a built in mechanism that is supposed to keep it from teleporting for a reason, and the fact you could bypass this by putting it in a locker was silly. Probably a change for the best.

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -46,6 +46,10 @@
 	for(var/obj/structure/closet/closet in get_turf(src))
 		if(closet != src && closet.anchored != 1)
 			return FALSE
+	for(var/mob/living/silicon/ai/ai in get_turf(src))
```
Citinited: I wonder if it's cheaper to do two `for` loops vs. doing something like
```
for(var/atom/A in get_turf(src))
    if(isAI(A))
        return FALSE
    if(istype(A, /obj/structure/closet) && !A.anchored)
        return FALSE
return TRUE
```
Anyone know?

Ty-Omaha: @farie82 might?

Ty-Omaha: i'd think it would be cheaper to do 1

AffectedArc07:
Note to self: What the fuck has happened to mapdiffbot

code/game/objects/structures/crates_lockers/closets.dm
farie82: Can also use `locate(mob/living/silicon/ai) in get_turf(src)` to find one thing.

I'd say one loop will be faster since it will still get all the contents of the turf and filter them. Twice in the current case. 
The given suggestion will get them once then loop through them. Which should be better performing in theory. But I'm not to sure how byond does this all
Can prob even do `var/A` instead of `var/atom/A`.

AffectedArc07:
Fox fixed it

code/game/objects/structures/crates_lockers/closets.dm
Quantum-M: Which one would be best recommended?
What I currently have, the suggestion made by @Citinited , or the usage of the `locate` proc?

Crazylemon64: Why don't you put the change at line 199, inside the `close` loop instead

Quantum-M: So it would be something like this, @Crazylemon64 ?

```
for(var/obj/structure/closet/closet in get_turf(src))
	if(closet != src && closet.anchored != 1)
		return FALSE
	for(var/mob/living/silicon/ai/ai in get_turf(src))
                if(ai)
	               return FALSE
```
Because if so, what @Citinited suggested would look more cleaner.

Citinited: The second for loop shouldn't be inside the first, and you don't need the `if(ai)` bit because that will always be TRUE (just `return FALSE`)

Crazylemon64: It'd be in that `for(mob...)` loop - alongside that mulebot check

Quantum-M:
Requested changes done, @Crazylemon64 , changes moved to the close() proc, using the isAI() proc, and has been tested.

PR #13062, opened 2020-03-02
Patch rework

Fox-McCloud:
Credit to Goon for the ideas, a decent portion of the implementation, sprites, and sound. Sounds and sprites have been placed in the appropriate folder.

Reworks patches and adds a new medical applicator device.

Patches:

Patches no longer apply instantly; instead they stay "stuck" on a person and periodically transfer reagents and apply a minor `TOUCH` reaction to the person they're stuck to. The more patches you have stuck to the person, the more reagents they transfer in total. 

If you want a slow, steady heal over time, then just stick with one patch, where as slapping on a whole bunch is going to heal more up front; it also relies on the chem's passive healing to heal you up.

Why? 

Patches are great and nice, and having a different application method is a good thing that adds more gameplay and nuance. That said, a huge burst heal in the middle of combat can get a bit ridiculous at times, especially when someone is carrying a patch pack with 50 styptic patches---using them to go from -50 health to 100 health in the span of literally 2 seconds. It also makes cheesing hostile mobs, especially on Lavaland, pretty easy. Just face tank the mob and slap one a patch or two when you're low.

This still keeps patches more or less effective for day to day use, but largely relegates them to a low tier treatment method meant for patching up someone slowly who's not at risk of dying.

That said, patches are not a viable solution for treating patients in critical states, which is why there's:

Auto-mender:

Auto-mender is a device that can be loaded with styptic, silver sulfa, or synthflesh (or anything if emagged). They apply a tiny little splash of a chemical instantaneously, but then start a loop of applying a lot more of their reservoir, over time, at intervals of 1 second.

This is the doctor's new best friend for treatment; they rapidly can apply chemicals and heal up a patient. They have a total volume of 200 units.

Why?

These allow doctors to more effectively and rapidly treat patients, while keeping them out of other's hands a bit more---they can rapidly treat a patient, similar to old patches, but their "stand still to use" nature makes them very very difficult to use in combat unless you get an extended breather. 

img

Other changes related:
- Brute kits now have a styptic auto-mender and single mini styptic patch (15 units) instead of 2 full sized styptic patches
- Burn kits received the same treatment
- Doctors now spawn in with a special medkit that has 1 brute auto-mender, 1 burn auto-mender, 1 full burn patch, 1 full brute patch, 1 upgraded health analyzer, salicyllic pill, and autoinjector
- Patch total volume reduced from 40 to 30
- Reduced the number of patches in the medical vendor from 10 of each type to 4
- Included 3 auto-menders of each type in the medical vendor
- Auto-menders can be produced in the autolathe
- Tactical medkit now has a single synthflesh automender instead of 4 syntflesh patches

Combat healing that is done in huge bursts is overall problematic, especially with Gooncrit; it turns engagements against powergamers, well prepared, and lucky individuals into lengthy process. Altering patches to be more heal over time helps cut down on this abuse. Cutting down on this abuse obviously introduces a new problem of not being able to effectively treat patients in medbay in a timely manner, especially in critical condition; the introduction of the auto-mender fills this gap; it can rapidly apply healing chemicals while being rare enough to make it less viable for combat, especially in combination with its "stand still to use" nature.

Other
- Implements Goon's idea for limiting reagent temperatures; we handle this quite a bit differently than they do (more OOP and easier to use for future reagent containers)---basically prevents something from hitting a certain temperature, either too hot or too cold. Patches and auto-menders both use this

:cl: Fox McCloud
add: Patches no longer instantly heal, but apply slowly over time. Having more patches increases the rate at which they transfer reagents
add: adds new auto-mender device that can apply styptic, silver sulfadiazine, or synthflesh to patients in a rapid manner
tweak: patches now hold 30 units instead of 40
tweak: Removed 6 silver sulfadiazine patches and 6 brute patches from the medical vendor
add: Added 3 brute auto-menders and 3 burn auto-menders to the medical vendor
add: Can produce automendors at the autolathe
tweak: doctors now start with a medkit that has 1 brute auto-mender, 1 burn auto-mender, 1 styptic patch, 1 burn patch, 1 upgraded analyzer, 1 painkiller, and 1 autoinjector
tweak: brute and burn kits now have 1 less of their name-sake healing patches and instead have a brute/burn auto-mender instead; the second patch has now been replaced by a mini healing patch that has 15 units of healing in it instead of 30
tweak: tactical medkit has a single synthflesh auto-mender instead 4 synthflesh patches
/:cl:


code/modules/reagents/reagent_containers/applicator.dm
```diff
@@ -0,0 +1,104 @@
...
+	if(!reagents.total_volume)
+		to_chat(user, "<span class='warning'>[src] is empty!</span>")
+		return
+	if(applying)
+		to_chat(user, "<span class='warning'>You're already applying [src].</span>")
+		return
+	if(!iscarbon(M))
+		return
+
+	if(reagents.total_volume && (ignore_flags || M.can_inject(user, TRUE)))
```
farie82: reagents.total_volume is already checked above

code/modules/mob/living/carbon/life.dm
```diff
@@ -437,4 +438,22 @@
...
+	if(LAZYLEN(processing_patches))
+		var/multiple_patch_multiplier = processing_patches.len > 1 ? (processing_patches.len * 1.5) : 1
+		var/applied_amount = 0.35 * multiple_patch_multiplier
+
+		for(var/patch in processing_patches)
+			var/obj/item/reagent_containers/food/pill/patch/P = patch
+
+			if(P.reagents && P.reagents.total_volume)
+				var/fractional_applied_amount = applied_amount  / P.reagents.total_volume
+				P.reagents.reaction(src, TOUCH, fractional_applied_amount, P.reagents.total_volume == P.reagents.maximum_volume ? TRUE : FALSE)
```
farie82: Can remove the ternary operator here. Maybe put ()'s around the logical statement to make clear it is a param instead.

code/modules/mob/living/carbon/life.dm
```diff
@@ -437,4 +438,22 @@
...
+				P.reagents.trans_to(src, applied_amount * 0.5)
+				P.reagents.remove_any(applied_amount * 0.5)
+			else
+				if(P.reagents.total_volume <= 0)
```
farie82: can runtime if P.reagents is null. You checked it above but not here

Cazdon:
The auto-menders look like wee little bibles.

Crazylemon64:
Hmm, beakers/buckets would still work as "patches", right? Although they're a good deal less portable (due to them not fitting in pill bottles)

code/modules/mob/living/carbon/life.dm
```diff
@@ -437,4 +438,22 @@
...
+	if(LAZYLEN(processing_patches))
+		var/multiple_patch_multiplier = processing_patches.len > 1 ? (processing_patches.len * 1.5) : 1
+		var/applied_amount = 0.35 * multiple_patch_multiplier
+
+		for(var/patch in processing_patches)
+			var/obj/item/reagent_containers/food/pill/patch/P = patch
+
+			if(P.reagents && P.reagents.total_volume)
+				var/fractional_applied_amount = applied_amount  / P.reagents.total_volume
+				P.reagents.reaction(src, TOUCH, fractional_applied_amount, P.reagents.total_volume == P.reagents.maximum_volume)
```
Crazylemon64: Wouldn't this mean that patches smaller than the maximum amount (less than 40) would not display a message?

Fox-McCloud:
`Hmm, beakers/buckets would still work as "patches", right? Although they're a good deal less portable (due to them not fitting in pill bottles)`

Yeah, they would, but as you noted---carrying a lot of buckets/beakers is very difficult and it tends to be a once off thing--as opposed to carrying, literally, 20+ patches in an item that fits in your pocket.

code/modules/mob/living/carbon/life.dm
Fox-McCloud: Yeah, it does. 

Any other ideas? `initial(P.reagents.total_volume) == P.reagents.total_volume`  doesn't work.

Crazylemon64: Ideas:
- Set `maximum_volume` in initialize of a patch, although this could lead to problems if we somehow add more chems to a patch after creation
- Give a free "reaction" cycle directly on patch application, and don't show a message in the loop
- Add a flag to patches which says whether they've given their reaction message yet or not

lolzyking:
Does this affect synth flesh for cyborg/ ipc mending

Fox-McCloud:
Went ahead and implemented the temperature limiting on certain reagent containers; this is applied to patches and auto-menders so people aren't cheeky with them.

code/modules/mob/living/carbon/life.dm
Fox-McCloud: Been addressed!

Necaladun:
Quite like it

Mitchs98:
My only question as overall I'm pretty okay with this and the menders sound cool, what are the specifics on the automenders apply rate? Configurable or set? 

I read the code but I can't tell if it applies 10u or 1u at a time 🤔

pinatacolada:
does this play well with dead people? from what i gather the applicator is essentially a loop of "contact" or whatever it was called reagent transfers, but how much how fast? 10u per tick? I'm only concerned about someone that has for example 300 brute not being healed fast enough for a defib

PR #6590, opened 2017-02-25
Cloning Instability

Crazylemon64:
Cloning now causes permanent genetic instability - this can be mitigated by either upgrading the cloner (removed entirely at tier 4), or by keeping gene disk backups of the subject's DNA to clone them when the cloner is more sophisticated.
Do keep in mind that the cloner genetic instability change is mitigated by the negative mutations clones from low-tier cloners suffer from - meaning that in some cases, it may be wiser to skip the mutadone step, so that the patient may live.

At base level, patients are able to be cloned once without permanent detriment to everyday behavior. The permanent instability caused then drops as the cloner is upgraded, scaling linearly until tier 4 parts(scanners), where this is eliminated.

Hopefully, this should push medical to either not let patients die, or stick to more constrained/difficult, yet less punishing means of revival, such as the defib, or botany.

:cl:Crazylemon
experiment: Cloning now reduces genetic stability each time someone is cloned. This can be mitigated by either keeping gene disks handy, or by upgrading the cloner's scanning unit.
/:cl:

PR #13064, opened 2020-03-02
Cloning Nerf: Goon Edition

Fox-McCloud:
Credit to Goon

Cloning now degrades your genome.

Each time you get cloned, it will degrade your genetic material which reduces your max health each time you do it. As it currently stand, you get 1 "free" clone; being cloned a second time after this will reduce your max health by 15  per each iteration after your original cloning.

:cl: Fox McCloud
tweak: excessive amounts of cloning will reduce your maximum health
/:cl:

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -81,6 +81,7 @@ var/global/list/bad_blocks[0]
 	var/uni_identity="" // Encoded UI
 	var/struc_enzymes="" // Encoded SE
 	var/unique_enzymes="" // MD5 of player name
+	var/clone_generation = 0 //This increments every time you clone, to detrimental effects.
 
 	// Original Encoded SE, for use with Ryetalin
 	var/struc_enzymes_original="" // Encoded SE
@@ -111,6 +112,7 @@ var/global/list/bad_blocks[0]
 	new_dna.blood_type = blood_type
 	new_dna.real_name=real_name
 	new_dna.species = new species.type
+	new_dna.clone_generation = clone_generation
 	for(var/b=1;b<=DNA_SE_LENGTH;b++)
 		new_dna.SE[b]=SE[b]
 		if(b<=DNA_UI_LENGTH)
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -258,6 +258,7 @@
 		H.real_name = R.dna.real_name
 
 	H.dna = R.dna.Clone()
+	H.dna.clone_generation++
 
 	for(var/datum/language/L in R.languages)
 		H.add_language(L.name)
@@ -573,6 +574,7 @@
 
 	H.setCloneLoss(CLONE_INITIAL_DAMAGE, FALSE)
 	H.setBrainLoss(BRAIN_INITIAL_DAMAGE)
+	H.maxHealth -= (H.dna.clone_generation - 1) * 15
 
 	for(var/o in H.internal_organs)
 		var/obj/item/organ/O = o


Cazdon:
I like the idea, and I even think the 'freebie' shouldn't be given. It's rare to see a person cloned more once in my experience, so I think this penalty won't come into effect as often as hoped.

Also are there plans for a change to SR in a similar light?

TDSSS:
So if you get cloned enough, you'll come out of the pod already in crit/dead? Should it even let you clone at that point? IMO the cloning should just fail at that point.

I assume this doesn't affect long-ranged cloner or other ways of reviving the dead.

Can you bypass the health loss by getting cloned, then having your brain transplanted into a monkey or similar, then changing the monkey to look like you?

Fox-McCloud:
>So if you get cloned enough, you'll come out of the pod already in crit/dead? Should it even let you clone at that point?

Yes, it will. I do think the nature of popping out and being in perpetual crit has some interesting/hilarious merit---being a shambling, gurgling abomination who's only alive because of continuous chems and treatment--but it's worth it because they're the best darn captain there ever was, or something ridiculous like that.

Tayyyyyyy:
I wonder if the cloning machine should read the cloning generation number and warn about potential genetic damage if they've been cloned already. That way it'll deter people from abusing cloning too much.

> Can you bypass the health loss by getting cloned, then having your brain transplanted into a monkey or similar, then changing the monkey to look like you?

At this point, why even clone? You would just do the brain transfer, I would think. Which means the PR is working.

Kyep:
Not sure how I feel about this.
At minimum, it needs an explanation for why allowing people to be cloned multiple times is bad for game balance / results in a worse game. Simply so we can examine the reasoning, not just the change list.
I do think we probably don't *need* the ability to clone people a bunch of times... it does make death a bit cheap. At the same time though, a summary of how the mechanic works is NOT a substitute for the argument that needs to be presented about why this approach is better.

farie82:
Good idea which should however be paired with a more forgiving defib timer or a way to extend said timer.
I hate cloning to revive people. It's cheap and skill less. However defibbing takes time and skill so it should be rewarded and possible

Mitchs98:
While I do like the idea of making cloning much less appealing, I agree with Farie. If this is going to go through alternative methods that award a doctor's skill and time should be more accessible to use/not as terrible an option to use. Defib timers being extended, SR downsides being lessened(Reasonably, just so everything won't go septic past the ten minute mark if RNG decides fuck you), etc. 

I also certainly don't agree in changing SR to also do something similar to this PR as Cazdon mentioned. Overall, neutral without any further changes to revival methods.

Crazylemon64:
I feel like I made this PR before
I think it was with genetic stability itself, instead of maxhealth

and you can get around it just fine by using cloning disks instead

Crazylemon64:
Yeah, this one
https://github.com/ParadiseSS13/Paradise/pull/6590

taukausanake:
I really liked the genetic stability thing. I rather have that than this to be honest because the stability is at least more of an impact on a person than less health in my opinion. You come out of the cloner and your body doesn't work as well as it did before rather than just being overall weaker.

Out of curiosity, if we are getting a freebie clone at start, how many times does a single spaceman get cloned per round? If I was running the cloner I'd just toss the body to the morgue if I had a cloned body come in. You had your second chance

Crazylemon64:
idk i'm partial on gene stability bc it leads to neat emergent behavior like "has to have mutations or you're going to die" on multi-clones, among other things

Carthusia:
Im mixed on this idea, while I see it making a sort of sense, it also contradicts the upgraded cloner effects where with its baseline parts you see cellular damage and defects, but once its been fully upgraded it does not have this downside, this PR would be a downside no matter what the crew do to prevent it.

Would there be a method for genetics to reverse the effects given some hands on tweaking of your code? That is their whole schtick and gives them something other than RnD minesweeper to do.

What about SR? The person is not being cloned, just more or less rapidly restored and jolted back to life with cellular defects that need to be fixed via cryo and more than likely surgery. 

Feels like it would be even more of a slap to the face to the non-cloner races should they die if get almost guaranteed septic/dead organs/limbs upon being SRd, require possibly lengthy surgeries to correct the problem or simply die again should the heart go necrotic due to the SR, requiring a full heart replacement, and then to have their health permanently reduced by 15 for the rest of the round, just seems like salt on the wound.

Again im mixed on this, leaning towards disliking the idea entirely

Crazylemon64:
You get a free clone in either system

Aquilar:
Couldn't it be cheesed by copying the DNA from the original body over and over? and I know there will be people that keep their dead bodies just to do that. 
Or even just using SM to negate the clone generation?

Crazylemon64:
@Aquilar scanning doesn't work once a mob is dissociated from its ghost

Aquilar:
> scanning doesn't work once a mob is dissociated from its ghost
> However, stable mutagen could get around it
> As well as cloning disks

 I sort of meant a person could just be brain swapped into their original body over and over, so long as it's preserved that is.

Fox-McCloud:
maybe I'll revisit one day.

PR #13066, opened 2020-03-03
SOP Books now read the SOP on the wiki and has new sprites

Quantum-M:
## What Does This PR Do
This makes it so the code for the SOP books reads the data from the SOP entries in the wiki, and gives them more unique sprites.

## Why It's Good For The Game
I think this would make it more easier to both search the SOP information needed, and cite them, without the the SOP books having outdated information.

The sprites is to make them look nicer and have them stand out.

## Images of changes

image

## Changelog
:cl: Quantum-M
add: Adds new Sprites for the SOP books.
tweak: The SOP books now pull information from the SOP on the wiki.
/:cl:


Quantum-M:
A concern I had when testing this was that there were some links when I checked the books, I am not sure if it's due to it running locally at my computer, as I also checked the Space Law book (which is the code I used to base off the SOP book ones) and it had links.

farie82:
Best to use `config.wikiurl` if you can somehow. This will make it easier if we migrate the wiki URL again or such. And for downstreams

AffectedArc07:
Linking to the `config.wikiurl` is highly dependant on what it is. If it is `https://www.paradisestation.org/wiki/index.php/Main_Page`, we wont be able to use this system.

PR #13067, opened 2020-03-03
Fix global narrate not using admin pencode

Tayyyyyyy:
## What Does This PR Do
admin pencode was left out of global narrate

## Why It's Good For The Game
Need to be able to global narrate "Narsie has risen!"


AffectedArc07:
`[class notice][b]The gamemode is now: raginmages[/b][/class]`

PR #13076, opened 2020-03-04
Mountable frame bugfixes and minor refactor

### LINKED ISSUE #12598: Certain mountable frames aren't getting applied to walls with the proper orientation and location
**Issue Description**:

Currently the fire alarm frame, intercom frame, and fire extinguisher cabinet frame are broken in terms of placing them on walls. Trying to place any one of these doesn't work properly no matter which side of a wall you apply it to.

Video of the problem:
https://streamable.com/oqmpd

I *think* the problem code lies somewhere in one of these files here:
https://github.com/ParadiseSS13/Paradise/tree/master/code/game/objects/items/mountable_frames

And specifically this proc, which is also present on the other frames.
https://github.com/ParadiseSS13/Paradise/blob/master/code/game/objects/items/mountable_frames/fire_alarm.dm#L8-L10

However, what I don't understand is that all of this code hasn't been touched in months or years, save for the APC code touched in universal object damage. But that's irrelevant, because placing APC is working properly right now.

I've already tried fiddling around with it to no avail, hoping someone else has any ideas as to what the problem could be.

Take either a fire alarm frame, intercom frame, and fire extinguisher cabinet frame and attempt to place it like I've shown in the video.

Can't have been more than a week or so ago.

### LINKED ISSUE #12610: Attacking a wall with an air alarm frame triggers a runtime
Runtime when attacking a wall with an empty air alarm frame. Occurs regardless of whether or not another alarm is in the same place.

```
[2019-10-18T20:12:17] Runtime in code/game/machinery/alarm.dm,252: Air alarm /obj/machinery/alarm lacks alarm_area and areaMaster vars during proc/master_is_operating()
   src: The alarm (/obj/machinery/alarm)
   src.loc: The floor (123,92,1) (/turf/simulated/floor/plasteel)
   usr: Mackenzie Shafer (citinited) (/mob/living/carbon/human)
   usr.loc: The floor (123,92,1) (/turf/simulated/floor/plasteel)
```
Perhaps related to #12598?

SteelSlayer:
## What Does This PR Do
Fixes this behavior, with these mountable frames: https://streamable.com/oqmpd

I also went ahead and cleaned up / refactored some duplicate code that was used with mountable frames to pick their pixel locations. 

## Why It's Good For The Game
Fixes #12598
Fixes #12610

## Changelog
:cl:
fix: Fixes extinguisher cabinets, intercom frames and fire alarm frames getting applied to walls with the incorrect orientation and location
fix: Fixes mounted intercom frames being unable to be welded from walls
fix: Fixes a runtime when placing an air alarm frame onto a wall
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -164,6 +164,34 @@
 	SEND_SIGNAL(src, COMSIG_ATOM_DIR_CHANGE, dir, newdir)
 	dir = newdir
 
+/*
+	Sets the atom's pixel locations based on the atom's `dir` variable, and what pixel offset arguments are passed into it
+	If no arguments are supplied, `pixel_x` or `pixel_y` will be set to 0
+	Used primarily for when players attach mountable frames to walls (APC frame, fire alarm frame, etc.)
+*/
+/atom/proc/set_pixel_offsets_from_dir(pixel_north = 0, pixel_south = 0, pixel_east = 0, pixel_west = 0)
+	switch(dir)
+		if(NORTH)
+			pixel_y = pixel_north
+		if(SOUTH)
+			pixel_y = pixel_south
+		if(EAST)
+			pixel_x = pixel_east
+		if(WEST)
+			pixel_x = pixel_west
+		if(NORTHEAST)
+			pixel_y = pixel_north
+			pixel_x = pixel_east
+		if(NORTHWEST)
+			pixel_y = pixel_north
+			pixel_x = pixel_west
+		if(SOUTHEAST)
+			pixel_y = pixel_south
+			pixel_x = pixel_east
+		if(SOUTHWEST)
+			pixel_y = pixel_south
+			pixel_x = pixel_west
+
 ///Handle melee attack by a mech
 /atom/proc/mech_melee_attack(obj/mecha/M)
 	return
--- a/code/game/objects/items/devices/radio/intercom.dm
+++ b/code/game/objects/items/devices/radio/intercom.dm
@@ -62,7 +62,7 @@
 
 /obj/item/radio/intercom/department/medbay/New()
 	..()
-	internal_channels = default_medbay_channels.Copy()
+	internal_channels = GLOB.default_medbay_channels.Copy()
 
 /obj/item/radio/intercom/department/security/New()
 	..()
@@ -270,4 +270,4 @@
 
 /obj/item/radio/intercom/locked/prison/New()
 	..()
-	wires.CutWireIndex(WIRE_TRANSMIT)
+	wires.CutWireIndex(RADIO_WIRE_TRANSMIT)
--- a/code/game/objects/items/devices/radio/intercom.dm
+++ b/code/game/objects/items/devices/radio/intercom.dm
@@ -68,7 +68,7 @@
 	..()
 	internal_channels = list(
 		num2text(PUB_FREQ) = list(),
-		num2text(SEC_I_FREQ) = list(access_security)
+		num2text(SEC_I_FREQ) = list(ACCESS_SECURITY)
 	)
 
 /obj/item/radio/intercom/syndicate
@@ -80,7 +80,7 @@
 
 /obj/item/radio/intercom/syndicate/New()
 	..()
-	internal_channels[num2text(SYND_FREQ)] = list(access_syndicate)
+	internal_channels[num2text(SYND_FREQ)] = list(ACCESS_SYNDICATE)
 
 /obj/item/radio/intercom/pirate
 	name = "pirate radio intercom"
--- a/code/game/objects/items/devices/radio/intercom.dm
+++ b/code/game/objects/items/devices/radio/intercom.dm
@@ -54,7 +54,7 @@
 	else
 		if(direction)
 			setDir(direction)
-			set_pixel_offsets(28, -28, 28, -28)
+			set_pixel_offsets_from_dir(28, -28, 28, -28)
 		b_stat=1
 		on = 0
 	GLOB.global_intercoms.Add(src)
--- a/code/game/objects/items/devices/radio/intercom.dm
+++ b/code/game/objects/items/devices/radio/intercom.dm
@@ -46,16 +46,15 @@
 	name = "station intercom (Security)"
 	frequency = SEC_I_FREQ
 
-/obj/item/radio/intercom/New(turf/loc, ndir, building = 3)
-	..()
+/obj/item/radio/intercom/New(turf/loc, direction, building = 3)
+	. = ..()
 	buildstage = building
 	if(buildstage)
 		START_PROCESSING(SSobj, src)
 	else
-		if(ndir)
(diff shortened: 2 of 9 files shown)


code/game/objects/items/devices/radio/intercom.dm
```diff
@@ -204,7 +204,7 @@
 	STOP_PROCESSING(SSobj, src)
 
 /obj/item/radio/intercom/welder_act(mob/user, obj/item/I)
-	if(!buildstage)
+	if(buildstage != 0)
```
Ty-Omaha: can just do 
```suggestion
	if(buildstage)
```

Ty-Omaha: because if its not 0 it will be true

SteelSlayer: You are correct, I just wrote it like that for clarity

code/game/objects/items/mountable_frames/fire_alarm.dm
```diff
@@ -6,5 +6,5 @@
 	mount_reqs = list("simfloor", "nospace")
 
 /obj/item/mounted/frame/firealarm/do_build(turf/on_wall, mob/user)
-	new /obj/machinery/firealarm(get_turf(src), get_dir(on_wall, user), 1)
```
farie82: nice

code/game/objects/structures/extinguisher.dm
```diff
@@ -165,9 +165,9 @@
 	else
 		icon_state = "extinguisher_empty"
 
-/obj/structure/extinguisher_cabinet/empty/New(turf/loc, ndir = null)
+/obj/structure/extinguisher_cabinet/empty/New(turf/loc, direction = null)
 	extinguishertype = NO_EXTINGUISHER
```
farie82: Know per chance why this is put in here and not overriden in the object define?

code/game/objects/structures/extinguisher.dm
```diff
@@ -17,11 +17,11 @@
 	var/opened = 0
 	var/material_drop = /obj/item/stack/sheet/metal
 
-/obj/structure/extinguisher_cabinet/New(turf/loc, ndir = null)
+/obj/structure/extinguisher_cabinet/New(turf/loc, direction = null)
 	..()
```
farie82: You don't return the ..() here while you do in the subtype. Probably not required but it's inconsistent

code/game/objects/structures/extinguisher.dm
SteelSlayer: Actually its because this is some weird hybrid subtype define and `New()` proc merged into one

code/game/atoms.dm
```diff
@@ -164,6 +164,22 @@
 	SEND_SIGNAL(src, COMSIG_ATOM_DIR_CHANGE, dir, newdir)
 	dir = newdir
 
+/*
+	Sets the atom's pixel locations based on what pixel offset arguments are passed into it
+	If no arguments are supplied, `pixel_x` or `pixel_y` will be set to 0
+	Used primarily for when players attach mountable frames to walls (APC frame, fire alarm frame, etc.)
+*/
+/atom/proc/set_pixel_offsets(pixel_north = 0, pixel_south = 0, pixel_east = 0, pixel_west = 0)
```
moxian: This is named much more generically than it actually is. Many things want to set pixel offsets. Only mountable frames want to use this function.
It would be a fine name if it was a method on /obj/item/mounted/frame/ but it isn't (understandably so) - it's a global proc.
Please, please include some kind of "mountable frame" mention in the name
`set_pixel_offsets_for_mountable_frames` ,  `pixel_offsets_apc` or even `apc_pixel_fix` would all work.

code/game/objects/structures/extinguisher.dm
farie82: Oh like no subtype definition? You can just make one then and remove this New()

code/game/atoms.dm
```diff
@@ -164,6 +164,22 @@
 	SEND_SIGNAL(src, COMSIG_ATOM_DIR_CHANGE, dir, newdir)
 	dir = newdir
 
+/*
+	Sets the atom's pixel locations based on what pixel offset arguments are passed into it
+	If no arguments are supplied, `pixel_x` or `pixel_y` will be set to 0
+	Used primarily for when players attach mountable frames to walls (APC frame, fire alarm frame, etc.)
+*/
+/atom/proc/set_pixel_offsets(pixel_north = 0, pixel_south = 0, pixel_east = 0, pixel_west = 0)
+	switch(dir)
```
Citinited: What happens if `dir` is intercardinal?

code/game/atoms.dm
SteelSlayer: I honestly can't decide on a name that I like, maybe I *should* just make it a mountable frame only proc.

code/game/atoms.dm
SteelSlayer: Yeah, why not, I'll include them

code/game/atoms.dm
moxian: Actually, yeah, you can technically make a new `/obj/item/mounted/frame/proc/set_pixel_offsets` proc, and call that from other places via its full path (i.e. `/obj/item/mounted/frame/proc/set_pixel_offsets(-24, 24, -24, 24)`) (at least i'm 83% sure this is the case) (you might need to pass the atom as an extra argument, since I'm not 100% it would work with `src`)

SteelSlayer: Agghh, actually I can't because all these objects I'm touching have different paths:
1. `/obj/structure/extinguisher_cabinet`
2. `/obj/machinery/defibrillator_mount`
3. `/obj/item/radio/intercom`

For now I think I will go with `set_pixel_offsets_from_dir`. Generic yes, but hopefully you can tell what its suppose to do by its name, and by then reading the documentation for it.

moxian: no, what I'm saying is you can have

```dm
/obj/item/mounted/frame/proc/set_pixel_offsets(/obj/item/frame, pixel_north = 0, pixel_south = 0, pixel_east = 0, pixel_west = 0)
  switch(frame.dir)
    if (NORTH):
      frame.pixel_y = pixel_north
  // similarly other cases

// and then, later, in a different dm:

/obj/machinery/alarm/New(loc, direction, building = 0)
  // ..
  if(building):
    /obj/item/mounted/frame/proc/set_pixel_offsets(src, -24, 24, -24, 24)
```

Again, I'm only ~80% sure that this is the correct syntax (can't test rn :c ), but I think it should work.

SteelSlayer: I don't think that's even posssible to call procs like that, and even if it was, I would get stabbed for trying :smile:. 

But see the issue is that I can't create a single proc for all the mountable frames becuase, as I wrote in my previous comment, a lot of them have quite different paths, (one is `/obj/item`, one is `/obj/machinery`, etc.). Which is why I just went down the `/atom` level. But now that I think about it, I could also just do it on the `/obj` level.

PR #13078, opened 2020-03-05
Blocked Traders event triggers another major event

blinkdog:
## What Does This PR Do

If the 'Traders' major event fires, but the alert level is red or higher, so the traders don't come, then set the timer until the next major event to 60 seconds.

## Why It's Good For The Game

This will mean that there's actually a disincentive to stay on red alert pointlessly, because trader event on under red alert is beneficial, but trader event on red alert or higher causes another (almost certainly harmful) event to fire that would not otherwise have fired. 

The text describing this change is copied verbatim from [Kyet's Changes Wanted](https://www.paradisestation.org/forum/topic/18118-kyets-changes-wanted/)

## Changelog
:cl:
tweak: Blocked Traders event triggers another major event
/:cl:

### DIFF
--- a/code/modules/events/traders.dm
+++ b/code/modules/events/traders.dm
@@ -18,6 +18,9 @@ var/global/list/unused_trade_stations = list("sol")
 		return
 	if(seclevel2num(get_security_level()) >= SEC_LEVEL_RED)
 		event_announcement.Announce("A trading shuttle from Jupiter Station has been denied docking permission due to the heightened security alert aboard [station_name()].", "Trader Shuttle Docking Request Refused")
+		// if the docking request was refused, fire another major event in 60 seconds
+		var/list/datum/event_container/EC = SSevents.event_containers[EVENT_LEVEL_MAJOR]
+		EC.next_event_time = world.time + (60 * 10)
 		return
 
 	var/list/spawnlocs = list()


farie82:
Now prompt 10 denied traders trying to dock due to RNG

Cazdon:
Looked in a different light, it could give bored security another reason to be on Red Alert 24/7 cause there's nearly guaranteed action. That's me mostly joking however. Good change, I like it!

blinkdog:
> Now prompt 10 denied traders trying to dock due to RNG

When one-shot events are selected, they are disabled and can no longer be randomly selected. (Admins can manually re-enable or schedule the event though.)
https://github.com/ParadiseSS13/Paradise/blob/fe396a40c9783c033ed56c45397c7ae786de2540/code/modules/events/event_container.dm#L45

Traders is a one-shot event, so it will not fire more than once without admin intervention.
https://github.com/ParadiseSS13/Paradise/blob/fe396a40c9783c033ed56c45397c7ae786de2540/code/modules/events/event_container.dm#L198

Ty-Omaha:
I don't think traders shouldn't even fire if it's red. Whats the point in saying "trade request denied?" when it's just going to choose something else? Why not just choose something else in the first place?

farie82:
> I don't think traders shouldn't even fire if it's red. Whats the point in saying "trade request denied?" when it's just going to choose something else?
> 
> Why not just choose something else in the first place?

To make security/command feel bad for leaving it on red for the whole shift even though nothing is going on

Ty-Omaha:
>To make security/command feel bad for leaving it on red for the whole shift even though nothing is going on

Good point

PR #13079, opened 2020-03-05
Underage Drinking Is Now Illegal

Fox-McCloud:
We're Space America now.

Credit to Goon for the idea; I included their comment  because it's just too good to pass up.

Drinking alcohol will now incur an arrest with a minor crime of Underage Drinking if you're under the age of 21.

Being hidden from camera while or having no records makes you immune to this, of course.

For those who are asking why I didn't make this a `reaction_mob` on `INGEST`; it's in part so you can't cheese it, but also to avoid looping through the entire datacore every time someone takes a sip of any of the myriad of alcohol types.

:cl: Fox McCloud
add: Underage Drinking is now illegal. Welcome to Space America. This is canon now
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol.dm
@@ -15,6 +15,25 @@
 	M.AdjustDizzy(dizzy_adj)
 	return ..()
 
+/datum/reagent/consumable/ethanol/on_mob_add(mob/living/carbon/M)
+	..()
+	if(!ishuman(M))
+		return
+	var/mob/living/carbon/human/H = M
+	if(H.age >= 21) // Yes. Its 21. This is Space America. That is canon now.
+		return
+	if(!cameranet.checkTurfVis(get_turf(M)))
+		return
+	for(var/datum/data/record/E in data_core.general)
+		if(E.fields["name"] == M.real_name)
+			for(var/datum/data/record/R in data_core.security)
+				if(R.fields["id"] == E.fields["id"])
+					R.fields["criminal"] = "*Arrest*"
+					R.fields["mi_crim"] = "Underage Drinking"
+					R.fields["comments"] += "Set to *Arrest* by Nanotrasen Alcohol Oversight Committee on [current_date_string] [station_time_timestamp()]."
+					update_all_mob_security_hud()
+				break
+
 /datum/reagent/consumable/ethanol/reaction_obj(obj/O, volume)
 	if(istype(O,/obj/item/paper))
 		if(istype(O,/obj/item/paper/contract/infernal))


Quantum-M:
But why? This seems it would be more of a hassle to Security, especially since age doesn't have much of a factor in game besides mob sounds.

Necaladun:
Currently not a valid charge in space law, as hilarious as it is. I think it'd be interesting to add to give some consequence to playing a stupidly young character. Secret drinking bars for teenagers would be hilarious.

Fox-McCloud:
Because the vast majority of characters are over the age of 21, it's comedic, and gives security something silly to do, if they so desire when they have more time for it (much like the indecent exposure law).

Also, NT takes underage alcohol consumption very seriously. Bad habits start young. Alcoholism is a *very* serious problem---we only want the most productive of employees!

Mitchs98:
My only concern is security going _really hard_ on people and arresting them for drinking just because they can for no real reason other than 'the law'. Even if it does make sense and I could see a lot of interesting things come from it, security will often already ignore actual threats in favor of pursuing that one naked man or that one assistant that pushed them over once.

I'd actually be in favor of it due to funny moments that could come from it if it weren't for the vast amount of shitcurity that is likely to occur from it.

Otherwise neutral, I guess.

farie82:
Please no. People will make young chars now and troll security. Everybody set to arrest

Mitchs98:
> Please no. People will make young chars now and troll security. Everybody set to arrest

This is...also a concern I didn't think of. Oh lord.

Shockpoint:
Imagine having your legal drinking age be above 18.

Real talk though, I don't like this because the above issues will happen for sure. We're more often than not going to have security focus on stupid shit like this than actually care about real threats. We already have problems with security caring about lesser things than station-wide threats; we shouldn't be giving them (and ourselves) more things to deal with and worry about.

dovydas12345:
Well it may be funny and hilarious at first, but the possible issues mentioned above worry me. Don't think that this is a good idea.

MarsM0nd:
Shouldn't it depend on the species? For one, organic species are affected differently by alcohol, if only due to their liver. Another thing will be IPCs. They sure can be young, and still be a full adult, and aren't even affected by ethanol, but would get set to arrest for drinking that. The other way around is probably fine, synthanol, even if it gets removed quickly, is alcoholic for organics.

Citinited:
Perhaps drinking while operating heavy machinery should also be a punishable offence :smile:

Spacemanspark:
Not really a fan of this myself. Between IPCs, people trolling security and vice versa, along with the fact that it's not actually against space law, I don't think we should add this.

Birdtalon:
>Imagine having your legal drinking age be above 18.

Literally this.

Carthusia:
I mean, if it was a normally applied law that required security to check your ID to discover your drinking under the age, then it would be a nice little fluff law that gets some use on green alerts during slow times from time to time, but I dislike the fact that it sounds like its an automatically applied warrant, no warrant should be auto applied.

On the other hand, considering how NT is, it seems unlikely they would care if a 19 year old was drinking on their station, your old enough to be on the station working to death, your old enough to drink, same logic they use in the US army, if your old enough to join, your old enough to get smokes and drink while off duty on deployment.

datlo:
I dont believe there is currently any automated system that set people to arrest in the game and I think it should be kept that way. :-1:

TDSSS:
NT engineer: Our new crime detection system is complete. Our cameras can now automatically set people to arrest if they detect certain crimes. What crimes shall we set them to, boss? Murder? Breaking and entering?
NT manager: How about...underaged drinking instead! Forget about all that other stuff!

Cazdon:
While I think the gimmick of arresting people for under-age drinking is funny, I don't think an automated system would fit well. This could honestly just be a small change to Space Law, but on top of what everyone else said, even then its worth considering if it's worth adding another petty reason for security and crew to come into conflict.

I like the idea of cameras automatically catching certain things though, it can probably be used for some very interesting game play elements and this is kinda just the start.

Carthusia:
> I like the idea of cameras automatically catching certain things though, it can probably be used for some very interesting game play elements and this is kinda just the start.

I'm kind of on the opposite side of that, it should always be up to players paying attention to the station to spot crimes and warrant them accordingly, that moment you know the guards are taking a coffee break is the moment to slip past, or do your crime.

If its automated, even for small crimes, crimes that otherwise would just be ignored by security for more important issues at hand, start to flood their huds, and now take their attention away from those who did have them manually set for more serious offences.

thatdanguy23:
listen we get it, goon has great features and some of them are worth importing

but this one really isn't. it isn't suitable for our playbase, it isn't suitable for our space law or SOP, and of all things in the world that need adding and changing to the code, you chose /this/?
seriously?

all of the comments above that reveal issues with this PR are valid and things to be genuinely concerned about. nobody is going to enjoy this if it's passed.

Landerlow:
Dear god no.This would be abused by trolls if passed and it will most likely end up annoying a LOT of people. I'm also not a fan of people automatically being set to arrest. 

Other than that, underage drinking is not defined in space law which is a must before even implementing something like this. 

Also, as someone who plays security regularly (when I'm not adminning a round), I don't think I could be bothered arresting people for something like underage drinking, in this game. There are way more important things to focus on. 

In my opinion .. this wouldn't be fun at all and I'd rather this did not get merged.

Fox-McCloud:
I've never seen people take a *not completely serious* PR so seriously before.

Ya'all are alcohol addicts.

More seriously, even if this was merged, I really doubt it would have that much of an impact (especially the trolling), in the long term.

Running around naked is even lower effort, and that's really not done that much, nor does sec really care about it.

PR #13080, opened 2020-03-05
Panic bunker threshold is configurable

blinkdog:
## What Does This PR Do

Updates auto panic bunker code.
* Make the threshold at which it turns on (currently 150) into a config setting (`panic_bunker_threshold`).
* Make it automatically turn off again if player count goes below the threshold while it is on.
* When a player is blocked by it, give them the message "Server is not accepting connections from never-before-seen players until player count is less than (number)".
* When a player is blocked by it, don't display a message to admins.

The text above is paraphrased from [Kyet's Changes Wanted](https://www.paradisestation.org/forum/topic/18118-kyets-changes-wanted/)

* Admin verb to control panic bunker state is removed; adjust the configuration value and rely on automation instead.
(Edited: Kyet's request did not specify removing the verb. I have moved it below the attribution to make it clear that this was an informative point about the changes in PR, not a part of the original request specification.)

## Why It's Good For The Game

Panic bunker is automated and configurable. That's a good thing, right?

## Changelog
:cl:
add: panic_bunker_threshold configuration value added
add: Panic bunker will activate/deactivate automatically based on player count
del: Removed Admin verb to manually control panic bunker
del: Removed message to admins for users blocked by panic bunker.
tweak: Changed message sent to never-before-seen users blocked by panic bunker.
/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -78,8 +78,7 @@ var/list/admin_verbs_admin = list(
 	/client/proc/toggle_advanced_interaction, /*toggle admin ability to interact with not only machines, but also atoms such as buttons and doors*/
 	/client/proc/list_ssds_afks,
 	/client/proc/cmd_admin_headset_message,
-	/client/proc/spawn_floor_cluwne,
-	/client/proc/toggle_panic_bunker
+	/client/proc/spawn_floor_cluwne
 )
 var/list/admin_verbs_ban = list(
 	/client/proc/unban_panel,
@@ -1017,16 +1016,3 @@ var/list/admin_verbs_ticket = list(
 
 	log_admin("[key_name(usr)] has [advanced_admin_interaction ? "activated" : "deactivated"] their advanced admin interaction.")
 	message_admins("[key_name_admin(usr)] has [advanced_admin_interaction ? "activated" : "deactivated"] their advanced admin interaction.")
-
-/client/proc/toggle_panic_bunker()
-	set name = "Toggle Panic Bunker"
-	set category = "Admin"
-	set desc = "Disables new players connecting."
-
-	if(!check_rights(R_ADMIN))
-		return
-
-	GLOB.panic_bunker_enabled = !GLOB.panic_bunker_enabled
-
-	log_admin("[key_name(usr)] has [GLOB.panic_bunker_enabled  ? "activated" : "deactivated"] the panic bunker.")
-	message_admins("[key_name_admin(usr)] has [GLOB.panic_bunker_enabled  ? "activated" : "deactivated"] the panic bunker.")
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -78,8 +78,7 @@ var/list/admin_verbs_admin = list(
 	/client/proc/toggle_advanced_interaction, /*toggle admin ability to interact with not only machines, but also atoms such as buttons and doors*/
 	/client/proc/list_ssds_afks,
 	/client/proc/cmd_admin_headset_message,
-	/client/proc/spawn_floor_cluwne,
-	/client/proc/set_server_capacity
+	/client/proc/spawn_floor_cluwne
 )
 var/list/admin_verbs_ban = list(
 	/client/proc/unban_panel,
@@ -144,8 +143,7 @@ var/list/admin_verbs_server = list(
 	/client/proc/toggle_antagHUD_use,
 	/client/proc/toggle_antagHUD_restrictions,
 	/client/proc/set_ooc,
-	/client/proc/reset_ooc,
-	/client/proc/set_server_capacity
+	/client/proc/reset_ooc
 	)
 var/list/admin_verbs_debug = list(
 	/client/proc/cmd_admin_list_open_jobs,
@@ -1018,17 +1016,3 @@ var/list/admin_verbs_ticket = list(
 
 	log_admin("[key_name(usr)] has [advanced_admin_interaction ? "activated" : "deactivated"] their advanced admin interaction.")
 	message_admins("[key_name_admin(usr)] has [advanced_admin_interaction ? "activated" : "deactivated"] their advanced admin interaction.")
-
-/client/proc/set_server_capacity()
-	set name = "Set Server Capacity"
-	set category = "Admin"
-	set desc = "Prevent new players connecting when over capacity."
-
-	if(!check_rights(R_ADMIN))
-		return
-
-	var/capacity = input("Please input the new Server Capacity (player count)", "Set Server Capacity", 150) as num
-	config.panic_bunker_threshold = Clamp(capacity, 0, 500)
-
-	log_admin("[key_name(usr)] has set the Server Capacity to [capacity].")
-	message_admins("[key_name_admin(usr)] has set the Server Capacity to [capacity].")
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -78,7 +78,8 @@ var/list/admin_verbs_admin = list(
 	/client/proc/toggle_advanced_interaction, /*toggle admin ability to interact with not only machines, but also atoms such as buttons and doors*/
 	/client/proc/list_ssds_afks,
 	/client/proc/cmd_admin_headset_message,
-	/client/proc/spawn_floor_cluwne
+	/client/proc/spawn_floor_cluwne,
+	/client/proc/set_server_capacity
 )
 var/list/admin_verbs_ban = list(
 	/client/proc/unban_panel,
@@ -143,7 +144,8 @@ var/list/admin_verbs_server = list(
 	/client/proc/toggle_antagHUD_use,
 	/client/proc/toggle_antagHUD_restrictions,
 	/client/proc/set_ooc,
-	/client/proc/reset_ooc
+	/client/proc/reset_ooc,
+	/client/proc/set_server_capacity
 	)
 var/list/admin_verbs_debug = list(
 	/client/proc/cmd_admin_list_open_jobs,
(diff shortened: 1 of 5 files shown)


code/modules/client/client procs.dm
```diff
@@ -414,9 +414,16 @@
 		if(M.client)
 			playercount += 1
 
-	if(playercount >= 150 && GLOB.panic_bunker_enabled == 0)
-		GLOB.panic_bunker_enabled = 1
-		message_admins("Panic bunker has been automatically enabled due to playercount surpassing 150")
+	// Update the state of the panic bunker based on current playercount
+	var/threshold = config.panic_bunker_threshold
+
+	if((playercount > threshold) && (GLOB.panic_bunker_enabled == FALSE))
```
farie82: Can write the == FALSE as `!GLOB.panic_bunker_enabled`

code/modules/client/client procs.dm
```diff
@@ -414,9 +414,16 @@
...
+		GLOB.panic_bunker_enabled = TRUE
+		message_admins("Panic bunker has been automatically enabled due to playercount rising above [threshold]")
+
+	if((playercount < threshold) && (GLOB.panic_bunker_enabled == TRUE))
```
farie82: Same and you can put this in an else if

AffectedArc07:
I’m not 100% sure if the removal of the toggle verb is a great idea.

TDSSS:
Admins should still be able to toggle it I'd say, yeah.

Ty-Omaha:
yeah we manual toggle it more than it automatically toggles

TheSardele:
Does manually toggling actually do anything? IIRC it just checks every time someone connects anyway, and turns the bunker on or off.

AffectedArc07:
Manually toggling the PB turns the check on and off

code/modules/client/client procs.dm
blinkdog: I prefer the explicit use of `== FALSE` in the comparison expression.
I won't change it, but I wouldn't be upset if a maintainer chose to do so before merging.

code/modules/client/client procs.dm
blinkdog: I prefer the explicit use of `== TRUE` in the comparison expression.
I won't change it, but I wouldn't be upset if a maintainer chose to do so before merging.

I prefer the logic as-is, as I feel there is less cognitive load in understanding the conditions that trigger the action specified in the code.
```
if ((cond-a) && (cond-b))
   // do something here if (cond-a) and (cond-b)

if ((not-cond-a) && (not-cond-b))
   // do something here if (not-cond-a) and (not-cond-b)
```
vs.
```
if ((cond-a) && (cond-b))
   // do something here if (cond-a) and (cond-b)
else if ((not-cond-a) && (not-cond-b))
   // do something here if (((not-cond-a) or (not-cond-b)) and (not-cond-a) and (not-cond-b))
```

There is a small caveat in that the code may flap if the conditions are adjusted the wrong way. If someone changed the comparisons between `playercount` and `threshold` the wrong way, the first block could enable the bunker while the second would immediately disable the bunker. The logic as it written now should be okay though.

I'm not opposed to a change to `else if`, if others are strongly in favor of it.

blinkdog:
> Does manually toggling actually do anything? IIRC it just checks every time someone connects anyway, and turns the bunker on or off.

This is correct. The admin verb is a simple toggle of the current state of the panic bunker.
https://github.com/ParadiseSS13/Paradise/blob/bd85adb11c7ab2a78060334b605764c9617c0f32/code/modules/admin/admin_verbs.dm#L1029

So in this case, it was less of a toggle than an early activation button. If an admin attempted to toggle it off, when `playercount > 150`, the very next connection would have re-enabled the panic bunker due to meeting the activation conditions.

Adding the opposite action, disabling the panic bunker below the threshold, meant that the admin toggle button was nearly useless, as the automation would forcibly override the admin's toggle. If I were an admin, this would be a surprising result, and immediately lead to the question: Why wasn't this toggle removed if no longer works?

I considered keeping the toggle, but it would require two toggles:
* Toggle Panic Bunker Operating Mode: Manual or Automatic
* Toggle Panic Bunker State: Enabled or Disabled

This would also complicate the logic a little further, as each condition would have to check if the bunker was operating in automatic mode.

I decided that adding another toggle was probably confusing; does the busy admin handling 150+ players want to look up the subtle difference between two toggle switches? And decided instead to remove the toggle that didn't actually do what it said on the tin.

A possible improvement would be to change the toggle verb to a verb that allowed the admin to update the configuration value for `panic_bunker_threshold` with a little dialog box; like the way one sets the target temperature for a chem heater.

config/example/config.txt
```diff
@@ -190,6 +190,9 @@ GUEST_JOBBAN
 ## Uncomment this to stop people connecting to your server without a registered ckey. (i.e. guest-* are all blocked from connecting)
 GUEST_BAN
 
+## above this player count threshold, never-before-seen players are blocked from connecting
+PANIC_BUNKER_THRESHOLD 150
```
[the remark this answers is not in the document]
blinkdog: The configuration variable `panic_bunker_threshold` is set to a default of 150:
https://github.com/ParadiseSS13/Paradise/blob/4fffcd676d43caaaaaf4ac9bc20dce0f22728101/code/controllers/configuration.dm#L60

And it is overridden by the configuration only if it is specified:
https://github.com/ParadiseSS13/Paradise/blob/4fffcd676d43caaaaaf4ac9bc20dce0f22728101/code/controllers/configuration.dm#L507

I think if this line is missing from `config.txt` it will default to 150, which is probably worth noting in the comment. I don't think the game will break?

code/modules/client/client procs.dm
farie82: It is also a case of making viable paths appear. In the current case it suggests that both options are possible. While it's exclusive
Performance is also one to consider. But in this case it's not much of an issue but it's good to always keep that on in the back of your head

AffectedArc07:
What I would do here is change it to `GLOB.panic_bunker_status` and have it able to be set to `AUTO_ENABLED`, `AUTO_IDLE`, `MANUAL_ENABLED`, `DISABLED`
Auto enabled: panic bunker auto mode is on and the bunker is enabled 
Auto idle: auto mode is on but the bunker is disabled 
Manual enabled: an admin has manually enabled the bunker and it should not be auto disabled 
Disabled: the bunker should not auto enable, and it is not currently enabled

code/modules/admin/admin_verbs.dm
```diff
@@ -1017,16 +1016,3 @@ var/list/admin_verbs_ticket = list(
 
 	log_admin("[key_name(usr)] has [advanced_admin_interaction ? "activated" : "deactivated"] their advanced admin interaction.")
 	message_admins("[key_name_admin(usr)] has [advanced_admin_interaction ? "activated" : "deactivated"] their advanced admin interaction.")
-
-/client/proc/toggle_panic_bunker()
```
Crazylemon64: Don't remove this, add a flag for "manual override" or something

code/modules/client/client procs.dm
Crazylemon64:
```
if(GLOB.panic_bunker_enabled)
    if(playercount < threshold)
        GLOB.panic_bunker_enabled = FALSE
else
    if(playercount > threshold)
        GLOB.panic_bunker_enabled = TRUE
```
Hmm, I prefer how it looks the most this way due to the parallel structure, but the else and if on separate lines violates convention
```
if(GLOB.panic_bunker_enabled)
    if(playercount < threshold)
        GLOB.panic_bunker_enabled = FALSE
else if(playercount > threshold)
        GLOB.panic_bunker_enabled = TRUE
```
yeah this doesn't look nearly as nice

Crazylemon64: (idk if the other "allowed" solutions don't look/feel intuitive, I'd say it's fine as it is)

code/modules/admin/admin_verbs.dm
Kyep: I'm actually fine with removing this, and having the system be entirely automated. It isn't a requirement, but this really isn't a system that needs manual control. Server capacity doesn't change round to round.

code/modules/client/client procs.dm
blinkdog: I have rewritten and tested the bunker automatic logic in your first style:
```dm
if(GLOB.panic_bunker_enabled)
    if(playercount < threshold)
        GLOB.panic_bunker_enabled = FALSE
else
    if(playercount > threshold)
        GLOB.panic_bunker_enabled = TRUE
```
It may not go with convention, but I also like the way it looks.

code/modules/admin/admin_verbs.dm
blinkdog: I have added a new admin verb `Set Server Capacity` that prompts the admin for a number.
The configuration variable `config.panic_bunker_threshold` will be set to the value provided.
It's not the same as a toggle on/off, but it allows admin control just the same.

Kyep: I don't want a Set Server Capacity verb for admins.
I'm okay with a literal on/off switch for admins, so long as it affects only the remainder of the current round.
I'm not okay with any admin being able to change this in a way that persists beyond the current round.
I'm not okay with any admin being able to set this to say, 50, or 200. That just seems like it would create a bunch of problems for me to deal with.

AffectedArc07:
@Kyep Modifying `config.panic_bunker_threshold` will only stick for that round, because `config` is just a datum on the server, it doesnt re-write to the text files on disk, only affects that round.

I still agree that having the set capacity verb is a bad idea, and think that it should have a manual override for on and off, instead of a capacity limiter.

Kyep:
@blinkdog 
In your latest commit, your code supports 3 states:
1) Off
2) Automatic (based on config)
3) On (toggled on by an admin, then automatic turned off)

I am really uneasy with the fact your code still supports (3).
I do not want admins manually keeping the server in panic bunker mode when we are at like 50 pop.
To me, that's just as bad, if not worse, than keeping it off when we are at 200 pop.

Ideally there would be no verbs at all for admins.
But CL did request one, so I'll talk to @Crazylemon64 and see what we can agree on.
Then get back to you.

In the meantime, my objection stands.

blinkdog:
> Ideally there would be no verbs at all for admins.
> But CL did request one, so I'll talk to @Crazylemon64 and see what we can agree on.
> Then get back to you.

Sure. In this PR branch, there are three options:

1. Static Config, Automatic Only, No Verbs https://github.com/ParadiseSS13/Paradise/pull/13080/commits/4fffcd676d43caaaaaf4ac9bc20dce0f22728101

2. Dynamic Config, Automatic Only, One Verb https://github.com/ParadiseSS13/Paradise/pull/13080/commits/b6a6d4971462511c16e9f7524cdc2bc30db97765

3. Static Config, Manual or Automatic, Two Verbs https://github.com/ParadiseSS13/Paradise/pull/13080/commits/398a8615a584c97fceb732605f381d638a4ecddd

If you like any of these options as-is, feel free to merge it.
If you still need changes, let me know.

blinkdog:
> Kyet doesn't want the manual panic bunker verb

Sure. I'll push another commit to get us back to https://github.com/ParadiseSS13/Paradise/commit/4fffcd676d43caaaaaf4ac9bc20dce0f22728101 when I get home from work.

blinkdog:
@Crazylemon64 Now it should put us back to the original commit; a new configuration variable to define the panic bunker threshold, logic to automate the activation and deactivation of the panic bunker as player count rises and falls, and no admin verbs to muck anything up.

PR #11921, opened 2019-07-25
Port TG's chat subsystem

Couls:
**What does this PR do:**
Port TG's new chat subsystem https://github.com/tgstation/tgstation/pull/45354 which they ported from TGMC https://github.com/tgstation/TerraGov-Marine-Corps/pull/1043

The chat subsystem queues new chat messages and prevents lag spikes that usually happened when you'd be in a situation where you'd get a lot of messages at once, this means fastmos no longer KILLS your game, at least in my tests with 5 monkeys and 15 glass shards and deleting the floor, the game didn't even hang and this was on my laptop. Big performance improvement in that regard

Shotguns were also a culprit of this lag, shotgun buff or maybe nerf? depends I guess

**Changelog:**
:cl:
add: new chatsubsystem from TG!
/:cl:

PR #13084, opened 2020-03-05
Turns on the chat subsystem

SteelSlayer:
## What Does This PR Do
Turns on the chat subsystem. 

Originally ported here: #11921. However it was left with the flag `SS_NO_INIT` which mean's it never runs. TG corrected this same issue in: https://github.com/tgstation/tgstation/pull/47653.

## Changelog
:cl:
fix: Makes the chat subsystem actually initialize and run properly
/:cl:

### DIFF
--- a/code/controllers/subsystem/chat.dm
+++ b/code/controllers/subsystem/chat.dm
@@ -1,6 +1,6 @@
 SUBSYSTEM_DEF(chat)
 	name = "Chat"
-	flags = SS_TICKER|SS_NO_INIT
+	flags = SS_TICKER
 	wait = 1
 	priority = FIRE_PRIORITY_CHAT
 	init_order = INIT_ORDER_CHAT
--- a/code/controllers/subsystem/chat.dm
+++ b/code/controllers/subsystem/chat.dm
@@ -15,7 +15,7 @@ SUBSYSTEM_DEF(chat)
 			C << output(payload[C], "browseroutput:output")
 			payload -= C
 		else
-			qdel(i)
+			payload -= i
 
 		if(MC_TICK_CHECK)
 			return
--- a/code/controllers/subsystem/chat.dm
+++ b/code/controllers/subsystem/chat.dm
@@ -13,9 +13,7 @@ SUBSYSTEM_DEF(chat)
 		var/client/C = i
 		if(C)
 			C << output(payload[C], "browseroutput:output")
-			payload -= C
-		else
-			payload -= i
+		payload -= C
 
 		if(MC_TICK_CHECK)
 			return
--- a/code/controllers/subsystem/chat.dm
+++ b/code/controllers/subsystem/chat.dm
@@ -11,8 +11,11 @@ SUBSYSTEM_DEF(chat)
 /datum/controller/subsystem/chat/fire()
 	for(var/i in payload)
 		var/client/C = i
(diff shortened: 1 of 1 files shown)


SteelSlayer:
It means the subsystem will initialize on round start, and actually run. Currently it just sits there and does nothing.

Birdtalon:
>I'm gonna have to refer you to the original PR for the details: #11921. I don't know how it works exactly, beyond what I'm fixing here.

Turning on something without knowing what it does seems a bit illogical to me.

farie82:
Turning on the subsystem will allow for queueing of chat messages. Which should reduce the load of other to_chat heavy procs like explosions and such.

However at https://github.com/ParadiseSS13/Paradise/blob/49bdfbf3f02b91b56be4a1bdface2d51f6544dcf/code/controllers/subsystem/chat.dm#L12 I'm a bit concerned that the Client is not checked again if it's still valid

AffectedArc07:
@Birdtalon the person who made this subsystem did a copy paste without actually checking if their code was valid. This SS is designed to make chat messages less intensive by queuing them to send them out in blocks, which can help with events that make 600+ chat lines (such as the Tesla in xenobio).

SteelSlayer:
@farie82 So basically you want something like this?
```js
var/client/C = i
if(C)
    ...
```

farie82:
> @farie82 So basically you want something like this?
> 
> ```js
> var/client/C = i
> if(C)
>     ...
> ```

Aye aye would probably be safest.
However the list still needs to be cleaned from disconnected clients and their messages.

SteelSlayer:
Added in farie's suggested improvements. Also tested this with a tesla shocking a ton of slimes in xenobio like affected said. Seems to work fine with no issues. I wasn't getting those huge ~5 second game hangs while all the messages were being written to the chat.

code/controllers/subsystem/chat.dm
```diff
@@ -11,8 +11,11 @@ SUBSYSTEM_DEF(chat)
 /datum/controller/subsystem/chat/fire()
 	for(var/i in payload)
 		var/client/C = i
-		C << output(payload[C], "browseroutput:output")
-		payload -= C
+		if(C)
+			C << output(payload[C], "browseroutput:output")
+			payload -= C
+		else
+			qdel(i)
```
Crazylemon64: Wait a minute, this would only be reached if `C` is `null`, and so `i` would be `null`, and so this would always be `qdel(null)`. Instead, I would advocate that you do `payload -= i`
(Granted, I'm not sure how matching keys in an associated list would work - I think the game does a hard-del when a client disconnects either way)

SteelSlayer: Yeah you're right

code/controllers/subsystem/chat.dm
```diff
@@ -11,8 +11,11 @@ SUBSYSTEM_DEF(chat)
 /datum/controller/subsystem/chat/fire()
 	for(var/i in payload)
 		var/client/C = i
-		C << output(payload[C], "browseroutput:output")
-		payload -= C
+		if(C)
+			C << output(payload[C], "browseroutput:output")
+			payload -= C
```
Crazylemon64: tbh just shift `payload -= C` to the left both lines do the same thing

PR #13085, opened 2020-03-06
Maintenance drones no longer get a flash

### LINKED ISSUE #13053: Maintenance Drones have Flashes
**Issue Description**:

Maintenance drones had flashes in their inventory, despite not having them before and that it can only be used to break their laws.

I think its unintended, as an item that would only break the users laws seems wrong. Drones also didn't have a flash before.

**What happened instead**:

Crack open inventory to find t-ray scanner, find flash

See above.

Spawn as drone, check your modules.

Drones didn't have a flash a few days ago, last I played one was maybe two or three days ago.

**Extra information**:

Its probably a drone uprising.

SteelSlayer:
## What Does This PR Do
#12934 Made all `New()` procs call their parent, but in the case of drones, it gave them a flash. This removes their flash, because drones should not have them.

## Why It's Good For The Game
Fixes #13053

## Changelog
:cl:
fix: Maintenance drones no longer get a flash
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -27,9 +27,10 @@
 	..()
 
 
-/obj/item/robot_module/New()
+/obj/item/robot_module/New(loc, should_have_flash = TRUE)
 	..()
-	modules += new /obj/item/flash/cyborg(src)
+	if(should_have_flash)
+		modules += new /obj/item/flash/cyborg(src)
 	emag = new /obj/item/toy/sword(src)
 	emag.name = "Placeholder Emag Item"
 
@@ -539,7 +540,7 @@
 		/obj/item/stack/cable_coil/cyborg = 30
 		)
 
-/obj/item/robot_module/drone/New()
+/obj/item/robot_module/drone/New(loc, should_have_flash = FALSE)
 	..()
 	modules += new /obj/item/weldingtool/largetank/cyborg(src)
 	modules += new /obj/item/screwdriver/cyborg(src)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -27,9 +27,10 @@
 	..()
 
 
-/obj/item/robot_module/New()
+/obj/item/robot_module/New(loc, should_have_flash = TRUE)
```
AffectedArc07: Sidenote: A lot of these procs missing args will be fixed when I add in dreamchecker, which should kill off a fair few runtimes.

SteelSlayer: Dreamchecker will be nice

Crazylemon64:
Nah it should be fine that's Initialize that I'm thinking about

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -27,9 +27,10 @@
 	..()
 
 
-/obj/item/robot_module/New()
+/obj/item/robot_module/New(loc, should_have_flash = TRUE)
 	..()
-	modules += new /obj/item/flash/cyborg(src)
+	if(should_have_flash)
+		modules += new /obj/item/flash/cyborg(src)
```
Crazylemon64: Wait no what was I thinking
Adding extra items to a module should be put in a separate proc, and then override that proc in the drone not to give the item you don't want them to have

SteelSlayer: I feel like I should just go ahead and do that for every module, medical, mining, sec etc. Just throw every module's item giving code into some proc, name it `add_module_specific_items` or something, then call that in `New()`

Spacemanspark:
You are droneist sir

Maint drones rise up

PR #13086, opened 2020-03-06
Ports Floor Closet from Goonstation

MrMagolor:
## What Does This PR Do
This PR ports the "Floor Closet" traitor item from Goonstation; the Floor Closet itself is a closet which takes the appearance of any floor that it is closed on, providing an excellent stealthy storage, or a means to hide yourself. The disguise isn't perfect though, as a careful right-click or even examine will show the suspiciousness. And, of course, like smugglers' satchels, you have to remember where you put it! Currently, it costs 4 TC, and is allowed for both traitors and nuke ops, and can appear in surplus crates rarely.

## Why It's Good For The Game
It provides a good, fun, halfway point between smugglers' satchels and storage implants, while not being too strong to be ridiculous.

## Changelog
:cl:
add: Ported Floor Closet from Goonstation. Stealthy ambushes or a secret stash, your choice!
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/floor_closet.dm
+++ b/code/game/objects/structures/crates_lockers/closets/floor_closet.dm
@@ -1,0 +1,32 @@
+/obj/structure/closet/floor_closet
+	name = "floor"
+	desc = "Where did this come from?"
+	density = 0
+	icon_state = "bluespace"
+	icon_closed = "bluespaceclosed"
+	icon_opened = "bluespaceopen"
+	storage_capacity = 30
+	var/materials = list(MAT_METAL = 5000, MAT_PLASMA = 2500, MAT_TITANIUM = 500, MAT_BLUESPACE = 500)
+
+
+/obj/structure/closet/floor_closet/close()
+	var/turf/T = get_turf(src)
+		if (T)
+			src.icon = T.icon
+			src.icon_closed = T.icon_state
+			src.desc = T.desc + " It looks odd."
+			src.plane = T.plane
+		else
+			src.icon = 'icons/obj/closet.dmi'
+			src.icon_closed = "closet"
+
+    . = ..()
+	return
+
+/obj/structure/closet/floor_closet/open()
+		if (src.welded)
+			return
+		src.icon = 'icons/obj/closet.dmi'
+		src.plane = initial(src.plane)
+		..()
+		return
--- a/code/game/objects/structures/crates_lockers/closets/floor_closet.dm
+++ b/code/game/objects/structures/crates_lockers/closets/floor_closet.dm
@@ -1,27 +1,38 @@
 /obj/structure/closet/floor_closet
-	name = "floor"
+	name = "floor(?)"
 	desc = "Where did this come from?"
 	density = 0
 	icon_state = "bluespace"
 	icon_closed = "bluespaceclosed"
 	icon_opened = "bluespaceopen"
-	storage_capacity = 30
+	plane = OBJ_LAYER
+	layer = BELOW_OBJ_LAYER
+	storage_capacity = 30 //With the chameleon technology it doesn't hold as much as a normal one
 
 
 /obj/structure/closet/floor_closet/close()
-	. = ..()
 	var/turf/T = get_turf(src)
 	if (T)
 		src.icon = T.icon
 		src.icon_closed = T.icon_state
 		src.desc = T.desc + " It looks odd."
 		src.plane = T.plane
+		src.layer = T.layer
+		src.name = T.name
 	else
 		src.icon = 'icons/obj/closet.dmi'
 		src.icon_closed = "bluespaceclosed"
+	. = ..()
+	src.density = 0
 	return
+	
 
 /obj/structure/closet/floor_closet/open()
 	. = ..()
 	src.plane = initial(src.plane)
+	src.layer = initial(src.layer)
+	src.icon = initial(src.icon)
+	src.icon_opened = initial(src.icon_opened)
+	src.name = initial(src.name)
+	src.desc = initial(src.desc)
 	return TRUE
--- a/code/game/objects/structures/crates_lockers/closets/floor_closet.dm
+++ b/code/game/objects/structures/crates_lockers/closets/floor_closet.dm
@@ -6,27 +6,22 @@
 	icon_closed = "bluespaceclosed"
 	icon_opened = "bluespaceopen"
 	storage_capacity = 30
-	var/materials = list(MAT_METAL = 5000, MAT_PLASMA = 2500, MAT_TITANIUM = 500, MAT_BLUESPACE = 500)
 
 
 /obj/structure/closet/floor_closet/close()
+	. = ..()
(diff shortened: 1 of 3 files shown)


code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+	icon_opened = "bluespaceopen"
+	plane = OBJ_LAYER
+	layer = BELOW_OBJ_LAYER
+	storage_capacity = 30 //With the chameleon technology it doesn't hold as much as a normal one
+
+
+/obj/structure/closet/floor_closet/close()
+	var/turf/T = get_turf(src)
+	if (T)
+		src.icon = T.icon
```
farie82: src implied on all the vars you use here

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+		src.icon_closed = T.icon_state
+		src.desc = T.desc + " It looks odd."
+		src.plane = T.plane
+		src.layer = T.layer
+		src.name = T.name
+	else
+		src.icon = 'icons/obj/closet.dmi'
+		src.icon_closed = "bluespaceclosed"
+	. = ..()
+	src.density = 0
```
farie82: src implied and best to use FALSE/TRUE defines for boolean values.

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+	return
+	
+
+/obj/structure/closet/floor_closet/open()
+	. = ..()
```
farie82: If say the closet is welded shut and ..() returns FALSE then it will still reset the variables below.

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+	src.plane = initial(src.plane)
+	src.layer = initial(src.layer)
+	src.icon = initial(src.icon)
+	src.icon_opened = initial(src.icon_opened)
+	src.name = initial(src.name)
+	src.desc = initial(src.desc)
+	return TRUE
```
farie82: Always returning TRUE here will break logic. If say the closet is welded it will return FALSE on ..() yet it'll return TRUE here. Some code relies on the return of the open proc so it's best to remove this line.

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+		src.icon = T.icon
+		src.icon_closed = T.icon_state
+		src.desc = T.desc + " It looks odd."
+		src.plane = T.plane
+		src.layer = T.layer
+		src.name = T.name
+	else
+		src.icon = 'icons/obj/closet.dmi'
+		src.icon_closed = "bluespaceclosed"
+	. = ..()
```
farie82: ..() might return FALSE if for any reason the closet can't close and won't actually close (like having other closets on the same tile).  Yet all the variables will be set above and thus making it camouflaged.
This will also make the description add " It looks odd" each time you try it

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
+/obj/structure/closet/floor_closet
+	name = "floor(?)"
+	desc = "Where did this come from?"
+	density = 0
```
farie82: Best to use the logical defines TRUE/FALSE for boolean values

MrMagolor:
@farie82 will make changes tomorrow

Mitchs98:
Personally while it's a cool item, people absolutely do not need a new way to hide bodies. It's easy enough as it is, they don't need something the average(or really anyone) person would ever find at all. The only way to come across it would be sheer dumb luck.

Yes, you can space bodies, debrain them, and hide them in maint/wall lockers. But all of these still have a higher chance of being found if the brain isn't destroyed than this ever would.

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+		src.desc = T.desc + " It looks odd."
+		src.plane = T.plane
+		src.layer = T.layer
+		src.name = T.name
+	else
+		src.icon = 'icons/obj/closet.dmi'
+		src.icon_closed = "bluespaceclosed"
+	. = ..()
+	src.density = 0
+	return
```
Crazylemon64: No trailing returns if you aren't returning anything

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
```diff
@@ -0,0 +1,38 @@
...
+	icon_closed = "bluespaceclosed"
+	icon_opened = "bluespaceopen"
+	plane = OBJ_LAYER
+	layer = BELOW_OBJ_LAYER
+	storage_capacity = 30 //With the chameleon technology it doesn't hold as much as a normal one
+
+
+/obj/structure/closet/floor_closet/close()
+	var/turf/T = get_turf(src)
+	if (T)
```
Fox-McCloud: spacing.

Kyep:
Smuggle satchels already exist.
The only advantage of this over them is that it lets you hide corpses.
That is not a good thing.
Traitors already have a wide variety of options to hide corpses. Adding another isn't really unique or interesting. Why would a traitor buy this over, say, just spacing the body? Spacing makes the body harder to find (T-ray is a thing) and is free.
Personally I'd be more inclined to support, say, an implant, that lets a traitor implant a corpse to make them unscannable. That would let traitors prevent the revival of their targets, while at the same time leaving a mystery for the crew to solve, and also be per-target, rather than a closet that can hide many corpses.

Fox-McCloud:
This was always a great item on Goon that as a myriad of really interesting uses. Easier to find than a floor satchel, but more potential.

👍

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
MrMagolor: would I make an an if(. = . . ()) then?

MrMagolor:
Given the point being made for the vastly increased usability in hiding bodies, 4 TC does seem too low. Would somewhere from 6-8 work, instead? Or more?

code/game/objects/structures/crates_lockers/closets/floor_closet.dm
MrMagolor: or rather, an if(opened) ?

Fox-McCloud:
Yeah, likely. Maybe a bit more; I'll think about it, but 4 is too low.

Fox-McCloud:
no updates on this in a month--I still support this, but not even formatting was fixed. Closing.

Feel free to re-open one day.

PR #13090, opened 2020-03-08
Makes the singularity more realistic

Citinited:
**What does this PR do**
Makes our singularity behave much more realistically, by ensuring the latest scientific theories are upheld. Approved by physicists everywhere!

Also gives engineering the ability to act as a makeshift kitchen

:cl:
tweak: Humans now get spaghettified when the singularity eats them
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/food/foods/pasta.dm
+++ b/code/modules/food_and_drinks/food/foods/pasta.dm
@@ -47,6 +47,11 @@
 	list_reagents = list("nutriment" = 6, "tomatojuice" = 10, "vitamin" = 4)
 	tastes = list("pasta" = 1, "tomato" = 1)
 
+/obj/item/reagent_containers/food/snacks/pastatomato/singulo
+	name = "singularity spaghetti"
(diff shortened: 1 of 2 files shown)


Citinited:
I don't joke when it comes to the singularity Kyep, it's a feature™️ 🔫 😄

Spacemanspark:
I think Citinited broke

taukausanake:
Is `death(gibbed)` called only when you die by singulo? It sounds like if you die via gibbed and I for one know you can gib people by other means (mainly badmin but still). I would think you'd need to put this is a `singularity_act` proc instead but I don't know everything

Citinited:
@taukausanake Ahh right you are, should be under `singularity_act`. This will make anyone who gibbed turn into spaghetti. I'll change that as soon as I get home tonight. Good spot!

Stupid spaghetticode 😄

Citinited:
@taukausanake fixed 🍝

Fox-McCloud:
Closing, as this is likely a joke. If not, then uh, DM me about this.

Still absolutely hilarious though.

PR #13094, opened 2020-03-08
Update cloning scan messages

shawnhind:
## What Does This PR Do
Clarifies cloning scan messages to make them more clear for new players.

## Why It's Good For The Game
Easier to understand if you should keep trying to clone someone or not.  Makes it clear why someone can't be cloned if they can't be.

## Changelog
:cl:
tweak: adjusted scan messages for cloning
/:cl:

### DIFF
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -384,7 +384,7 @@
 		SSnanoui.update_uis(src)
 		return
 	if((!subject.ckey) || (!subject.client))
-		scantemp = "<span class=\"bad\">Error: Subject's brain is not responding. Further attempts may succeed.</span>"
+		scantemp = "<span class=\"bad\">Error: Subject's brain is not responding. Further attempts after a short delay may succeed.</span>"
 		SSnanoui.update_uis(src)
 		return
 	if((NOCLONE in subject.mutations) && src.scanner.scan_level < 2)
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -359,9 +359,15 @@
 	if(scan_brain && !can_brainscan())
 		return
 	if(isnull(subject) || (!(ishuman(subject))) || (!subject.dna))
-		scantemp = "<span class=\"bad\">Error: Subject is not scannable.</span>"
-		SSnanoui.update_uis(src)
-		return
+		if(isalien(subject)) 
+			scantemp = "<span class=\"bad\">Error: Xenomorphs are not scannable.</span>"
+			SSnanoui.update_uis(src)
+			return
+		// can add more conditions for specific non-human messages here
+		else 
+			scantemp = "<span class=\"bad\">Error: Subject species is not scannable.</span>"
+			SSnanoui.update_uis(src)
+			return
 	if(subject.get_int_organ(/obj/item/organ/internal/brain))
 		var/obj/item/organ/internal/brain/Brn = subject.get_int_organ(/obj/item/organ/internal/brain)
 		if(istype(Brn))
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -358,35 +358,31 @@
 		return
 	if(scan_brain && !can_brainscan())
 		return
-	if((isnull(subject)) || (!(ishuman(subject))) || (!subject.dna) || (NO_SCAN in subject.dna.species.species_traits))
-		scantemp = "<span class=\"bad\">Error: Unable to locate valid genetic data.</span>"
+	if(isnull(subject) || (!(ishuman(subject))) || (!subject.dna))
+		scantemp = "<span class=\"bad\">Error: Subject is not scannable.</span>"
 		SSnanoui.update_uis(src)
 		return
 	if(subject.get_int_organ(/obj/item/organ/internal/brain))
 		var/obj/item/organ/internal/brain/Brn = subject.get_int_organ(/obj/item/organ/internal/brain)
 		if(istype(Brn))
 			if(NO_SCAN in Brn.dna.species.species_traits)
-				scantemp = "<span class=\"bad\">Error: Subject's brain is incompatible.</span>"
+				scantemp = "<span class=\"bad\">Error: [subject.dna.species.name_plural] are not scannable.</span>"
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/cloning.dm
```diff
@@ -358,35 +358,31 @@
 		return
 	if(scan_brain && !can_brainscan())
 		return
-	if((isnull(subject)) || (!(ishuman(subject))) || (!subject.dna) || (NO_SCAN in subject.dna.species.species_traits))
-		scantemp = "<span class=\"bad\">Error: Unable to locate valid genetic data.</span>"
+	if(isnull(subject) || (!(ishuman(subject))) || (!subject.dna))
+		scantemp = "<span class=\"bad\">Error: Subject is not scannable.</span>"
```
shawnhind:
```suggestion
		scantemp = "<span class=\"bad\">Error: Unable to locate valid genetic data.</span>"
```

code/game/machinery/computer/cloning.dm
```diff
@@ -358,35 +358,31 @@
...
 		SSnanoui.update_uis(src)
 		return
 	if(subject.suiciding)
-		scantemp = "<span class=\"bad\">Error: Subject's brain is not responding to scanning stimuli.</span>"
+		scantemp = "<span class=\"bad\">Error: Subject has committed suicide and is not scannable.</span>"
 		SSnanoui.update_uis(src)
 		return
 	if((!subject.ckey) || (!subject.client))
-		scantemp = "<span class=\"bad\">Error: Mental interface failure.</span>"
+		scantemp = "<span class=\"bad\">Error: Subject's brain is not responding. Further attempts may succeed.</span>"
```
Kyep: Perhaps 'further attempts after a short delay may succeed', so they don't just hit it 3 times in 3 seconds, then give up, before the ghost has a chance to re-enter.

code/game/machinery/computer/cloning.dm
```diff
@@ -358,35 +358,31 @@
 		return
 	if(scan_brain && !can_brainscan())
 		return
-	if((isnull(subject)) || (!(ishuman(subject))) || (!subject.dna) || (NO_SCAN in subject.dna.species.species_traits))
+	if(isnull(subject) || (!(ishuman(subject))) || (!subject.dna))
```
Kyep: This should really be split into 3 separate checks, with 3 separate errors:
Error: no subject in scanner
Error: subject is not humanoid
Error: subject has no DNA

code/game/machinery/computer/cloning.dm
shawnhind: Good point.  Only reason I used this message is because it matches what the defib says when the ghost isn't in the body.

Qwertytoforty:
Looks like the no clone message is gone compleatly, might want to be re added so people know why husks can not be cloned

code/game/machinery/computer/cloning.dm
shawnhind: To be honest, I wasn't even sure how most of these conditions would ever happen.  The scan button is disabled if there's no subject inside the scanner.  You can't put in animals (non-humans?) into the scanner, and I'm not sure how someone could have no DNA.  

Any suggestions on how these things could happen so I can test them individually and add messages?

shawnhind:
> Looks like the no clone message is gone compleatly, might want to be re added so people know why husks can not be cloned

What do you mean by the no clone message?

code/game/machinery/computer/cloning.dm
Kyep: There might be certain species without DNA, like for example IPC, possibly, or slimes.

I also wonder if you can put a simple_animal into the scanner.

Fair enough if you can't, but if the reason is "you put an IPC into the scanner, they have no DNA, they're not even organic" then that should be obvious from the error message.

code/game/machinery/computer/cloning.dm
```diff
@@ -358,35 +358,31 @@
 		return
 	if(scan_brain && !can_brainscan())
 		return
-	if((isnull(subject)) || (!(ishuman(subject))) || (!subject.dna) || (NO_SCAN in subject.dna.species.species_traits))
```
Qwertytoforty: This message at the end, NO_SCAN in subject dna, it gets applied to people when they are absorbed by a changeling, and prevents cloning from happening if the cloner is not upgraded enough, could use a message why the person failed to be scanned correctly

shawnhind: Isn't that what this condition is?

```
if((NOCLONE in subject.mutations) && src.scanner.scan_level < 2)
```

That's still there.  The condition I removed could actually never be hit as far as I could tell.
```
if(scan_brain && !subject.get_int_organ(/obj/item/organ/internal/brain))
```

That's the condition I removed.  There's a condition which checks to see if they have no brain removed a few checks above so that one would always be hit before this one was ever checked.

Qwertytoforty: ah my mistake, I am blind, saw the red text under it and presumed it was removed as well

code/game/machinery/computer/cloning.dm
shawnhind: Simple animals I tried and it didn't let me put them in.  You make a good point about IPCs and slimes and I'll check to see if I need to add another message for them or if they already fall under the "[race] is not scannable" message.

Qwertytoforty: Think xenomorphs can be shoved in

shawnhind: Will check that too

Mitchs98:
Definitely a welcome change for the cloner, and will hopefully stop people from ramming an IPC in over and over before morguing them after wandering why it didn't work. Same for the other messages too, overall quite useful.

Ty-Omaha:
Even after all these years I didn't know what all of them meant.

code/game/machinery/computer/cloning.dm
shawnhind: Checked Xenomorphs and they are hitting the !ishuman condition.  Going to adjust messages to give more specific messages for IPCs and Xenos. Will do today

shawnhind:
Added a message for xenomorphs.  IPCs already got caught by the racial one.  Checked slimes and other simple animals and they can't be inserted.

PR #13098, opened 2020-03-09
Hypoglycemia

Fox-McCloud:
Adds hypoglycemia to human mobs. From Goon. Credit to them.

This obviously doesn't impact mobs that don't eat (like Abductors), and IPCs are also immune (though I'm not opposed to adding a specific version for them being low on energy).

Either case, hypoglycemia adds an additional penalty for if you're especially low on food. Ever noticed the flavor text of when someone is severely malnourished? That's when you get hypoglycemia--meaning you'll have to have the "starving" indicator for a little.

Hypoglycemia slows you down, makes you stutter, occasionally makes you fall down, and, if left untreated, can lead to you going into shock. 

Gives a little bit more "umph" to having to eat--for the majority of the time, most players won't notice this, but those that repeatedly ignore hunger and continue to stand at their R&D desk/chem station/whatever for the entire shift? Those individuals will run into issues.

Curing it as simple as getting something to eat. 

Also adjusts shock to require you to have nutrition levels above the hypoglycemia threshold to cure (otherwise it eventually becoming shock is pointless).

related:
- Fixes being "exercised" not protecting against heart failure.

:cl: Fox McCloud
add: Mobs will develop hypoglycemia if they don't eat for an exceptionally long time and are exceptionally hungry; can develop into a medical emergency if you don't eat
fix: Fixes being exercised not protecting against heart failure
/:cl:


TDSSS:
This is going to be brutal for vampires. They already have issues if they don't get their first victim relatively soon, getting stunned and going into shock will make getting some food basically impossible once the initial window has passed.

I'm okay with this affecting general crew, though.

Darkmight9:
^^^A valid point may want it to not affect vampires

Fox-McCloud:
The threshold to trigger it is *exceptionally* low.

A vamp that's just "hungry" is highly unlikely to trigger it; I'd rather see how it plays out, first, before jumping to conclusions that it's instantly a problem.

I get the concern, but < 100 nutrition is pretty doggone low.

Ty-Omaha:
This would be cool if people had diabetes as a disability

Fox-McCloud:
>This would be cool if people had diabetes as a disability

That's ***HYPER***glycemia.

Ty-Omaha:
Diabetics can have hyper or hypo glycemia, which is why in emergency situations they're just given glucose or sugary foods. It won't harm them, however having too much insulin will.

https://www.mayoclinic.org/diseases-conditions/diabetic-hypoglycemia

Aquilar:
Couldn't you call this starving or malnutrition, not hypoglycemia since being either of those does not usually lead the body losing control of blood sugar, unless there are additional conditions such as being a pre diabetic or raging alcoholic for example?

Fox-McCloud:
>Couldn't you call this starving or malnutrition, not hypoglycemia since. . .

Nope. That doesn't convey the same sense of medical urgency or have the slightly more immersive/medical-ish sounding nature.

"I had to feed him food; he was malnourished" doesn't have quite the immersive punch to it as "he had to be treated for his hypoglycemia".

Sure, both are just cramming a twinky in their mouth, but the latter has more of an immersive *umph*.

code/datums/diseases/critical.dm
```diff
@@ -133,4 +138,65 @@
...
+					affected_mob.Slowed(rand(4, 16))
+					to_chat(affected_mob, "<span class='warning'>You feel [pick("tired", "exhausted", "sluggish")].</span>")
+				if(prob(5))
+					affected_mob.Weaken(6)
+					affected_mob.Stuttering(10)
+					to_chat(affected_mob, "<span class='warning'>You feel [pick("numb", "confused", "dizzy", "lightheaded")].</span>")
+					affected_mob.emote("collapse")
+			if(3)
+				if(prob(8))
+					var/datum/disease/D = new /datum/disease/critical/shock
```
Kyep: Shock? For not eating?
Can't shock advance to other, lethal conditions?
It seems really weird that someone could fall into shock, then heart failure, then die, just because they did not eat for two hours.
Even if you assume a typical work shift is more like 8 hours, compressed.... people skip lunch all the time. It doesn't kill them.
This seems over the top.

TDSSS: Doesn't shock heal itself if your health is high enough anyway? Am I misremembering newcrit?

Darkmight9: Yeah it has a chance to heal itself if you have high health

MRgaty:
@Kyep  I mean people usually dont become severely malnourished for not eating for two hours, I guess this is a way to make not eating actually dangerous, a feature that focus in mechanics rather than being realistic, maybe this is a right step to make chefs role more important.

Kyep:
I'm really skeptical that I want the chef's role to be more important.
In my experience, quite often:
- there is no chef
- chef doesn't make anything
- chef makes food with severe downsides, like spider legs
- chef's food is stolen, poisoned, or otherwise messed with

I'm really not keen on the idea of making the chef more critical, given how often you just can't rely on them.

With, say, engineering, its different. There, you have the CE plus a whole bunch of engineers. Only one of them needs to be present and competent. 
With the chef, there's only one chef slot? So your chances of having a competent chef are much lower. Plus, its self-antagging to mess with the whole station's power grid as a non-antag, but nobody bats an eye when someone swipes or mildly poisons all the chef food.

I am really skeptical that it is a good idea to make the chef role more important.

IMHO, the chef is much like the librarian - an optional flavor role, not a vital role.

Carthusia:
I do like the idea

Kyet, while the chef is kind of just flavor, it is this due to the abundance of chicken noodle around the station. why go to the kitchen when you can just pay 120c to get a 2 cans of chicken noodle of which feed you for an entire shift.

Seriously, chicken noodle soup is ungodly potent to the point it makes you super fat after 1 can if you ate it at shift start with 75% hunger.

I like that kitchens would be more important because of this in extreme situations, but I doubt people will be starving to death more than they do now.

code/datums/diseases/critical.dm
farie82: It's part of the PR as well. It will heal if you have over 25 HP currently. The PR adds the condition that you need over the threshold amount of hunger. So this just means don't be starving and you'll heal from it. Unless you're dying

code/datums/diseases/critical.dm
```diff
@@ -133,4 +138,65 @@
...
\ No newline at end of file
+					H.set_heartattack(TRUE)
+
+/datum/disease/critical/hypoglycemia
+	name = "Hypoglycemia"
+	form = "Medical Emergency"
+	max_stages = 3
+	spread_flags = SPECIAL
+	spread_text = "The patient has low blood sugar."
+	cure_text = "Eating or glucose treatment"
```
farie82: Glucose treatment will might make people think they need to inject saline glucose.
Having the cure text be as clear as can be is best.

Spacemanspark:
As iffy as I am on this idea, it made me chuckle a bit as I remembered I have a sister with hyperglycemia, I should show her this

code/datums/diseases/critical.dm
Ty-Omaha: Saline glucose should cure it too otherwise it's a massive oversight that makes no sense

code/datums/diseases/critical.dm
```diff
@@ -133,4 +138,65 @@
...
+		var/mob/living/carbon/human/H = affected_mob
+		if(NO_HUNGER in H.dna.species.species_traits)
+			return TRUE
+		if(ismachine(H))
+			return TRUE
+	return ..()
+
+/datum/disease/critical/hypoglycemia/stage_act()
+	if(..())
+		if(affected_mob.nutrition > NUTRITION_LEVEL_HYPOGLYCEMIA)
```
Ty-Omaha: Saline glucose should cure it as well

Fox-McCloud:
This only kicks in an extremely low hunger intervals. It can develop into something worse to compel you to actually do something about it---it's very trivial to address and requires repeatedly ignoring getting there to begin with.

It's like if someone has the minim amount of brute to cause bleeding. That person is going to have to deliberately ignore it for a very long time before it to develop into an issue---and it's extremely trivial to fix.

It's not like food is rare outside the chef anyway. Even if it was harder to acquire, there's absolutely nothing wrong with making a game element in a **_roleplaying_** have more impact than being *pure* fluff.

Carthusia:
I like the idea to that say if the gateway missions are expanded on or even the current ones, bring food and other supplies with you, or you risk starving to death, cold and alone hiding in a closet after becoming stranded

Gatchapod:
While the mechanic sounds fun and amusing, I am not sure you can justify it with RP.
I get that running for two hours straight is unrealistic, but so is dying of hypoglycemia if you don't eat for 60 minutes.

variableundefined:
LF conflicts

PR #13108, opened 2020-03-11
Makes the FLW command ignore permissions if you're already an observer

farie82:
## What Does This PR Do
Will make the FLW command on tickets and such ignore permissions if you're already a ghost

## Why It's Good For The Game
This way ghosted mentors can use the button as well. Since if the user is already a ghost it will behave like orbiting the mob. Which they can already do

## Changelog
:cl:
tweak: The FLW button on tickets etc now ignores permissions if you're already an observer
/:cl:

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -1551,13 +1551,12 @@
 		show_player_panel(M)
 
 	else if(href_list["adminplayerobservefollow"])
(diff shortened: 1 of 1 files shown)


Shockpoint:
I would have loved this as a mentor...

Spacemanspark:
Same

TDSSS:
Good idea, I like it

PR #13112, opened 2020-03-12
Filters say of illegal characters

farie82:
## What Does This PR Do
Makes the say verb filter out illegal characters that can break the chat or just shouldn't be used in character.

Current list: {}[]<>^

This is done on the say verb. Meaning that other ways of saying stuff (being drunk etc) are not affected. This only clears direct player input.

## Why It's Good For The Game
< and > will just stop your message from going through. The others are unwanted characters

## Changelog
:cl:
tweak: Certain characters are removed from your say input
fix: Accidently typing < or > now doesn't kill your message
/:cl:

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -1,5 +1,5 @@
 
-#define ILLEGAL_CHARACTERS_LIST list("~" = "", "<" = "", ">" = "", "^" = "", \
+#define ILLEGAL_CHARACTERS_LIST list("<" = "", ">" = "", "^" = "", \
 	"\[" = "", "]" = "", "{" = "", "}" = "")
 
 /mob/proc/say()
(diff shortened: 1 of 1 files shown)


Darkmight9:
anyone else feel this is getting out of hand regarding tildes (~)?

Darkmight9:
I approve of the rest of this though

farie82:
Honestly speaking I don't care about it being used ingame. But I'd hate to see a mechanical impact being attached to it. So I'd rather have it removed compared to that

Carthusia:
Huh, so <> stop chat from going through, that explains why my chat just vanishes from time to time

farie82:
I've removed the ~from the list. Thinking about it a bit more made me realize this is an administration thing. And not allowing it to be used can restrict RP as well. Think of singing and such

PR #13113, opened 2020-03-12
changes the way of handling addictive drinks, also adds caffeine.[DNM]

Aquilar:
## What Does This PR Do

makes it so you get addicted to ethanol or the new drug caffeine instead of addiction to beer, whisky, coffee, iced tea, cafe latte and so on

## Why It's Good For The Game

Ending up addicted to multiple different drinks instead of the underlying chem that causes said addiction is bad for the game, technically this should have been done in some form or another when the addiction rework was added. ~~so the goon method is being ported to solve the current problem.~~
using an original method now, special thanks to farie82 for helping me get it there.

## Changelog
:cl:
add: caffeine, it does most of what coffee did but nothing extra
add: caffeine supplement pill.
add: new parent chemical for alcohols
tweak: removed the addictive aspects of alcohols, coffee and tea
tweak: instead of being addicted to the drink you'll be addicted to caffeine or ethanol instead. ~~technically makes drinking multiple types of alcohol highly lethal, like in the real world.~~
tweak: changes needed to make this all work.
/:cl:


TDSSS:
If we have an ethanol reagent, why doesn't every alcoholic drink just decay to ethanol and let's that deal all the drunkenness, addiction etc?

Aquilar:
> If we have an ethanol reagent, why doesn't every alcoholic drink just decay to ethanol and let's that deal all the drunkenness, addiction etc?

because all the alcoholic drinks were based off of ethanol as a parent.

Fox-McCloud:
I thought about implementing this, but I have reservations, for sure. 

On one hand, it makes handling stuff like ethanol addiction a bit easier, but I do wonder if we could handle that in addiction processing a bit better.

I just have concerns this is going to severely impact the overall existing balance and handling of alcohol and its behaviors.

I'm all for fixing the issue of alcohol overlap (though I don't see the same issue with tea and coffee; there's not an insane amount of subtypes, nor are people as prone to drinking multiple coffee/tea types), I'm just not really sure this is the best way to go about it.

Aquilar:
Farie suggested I set it up as a flag or type system specifically for addictive chems, through it mean you'll need to see up a new definition for each chem you wanted to be addictive.
I could not do it as it's a bit beyond me.

code/modules/reagents/chemistry/reagents/drugs.dm
```diff
@@ -150,6 +150,65 @@
...
+			update_flags |= M.adjustBrainLoss(1, FALSE)
+			update_flags |= M.adjustToxLoss(1, FALSE)
+			update_flags |= M.Stun(2, FALSE)
+		else if(effect <= 7)
+			M.visible_message("<span class = 'warning'>[M] looks irritated!</span>")
+			M.Jitter(30)
+			M.emote("grumble")
+	if(severity == 2)
+		if(effect <= 2)
+			M.visible_message("<span class='warning'>[M] suddenly cluches [M.p_their()] chest!</span>")
```
Crazylemon64: clutches

Aquilar:
waiting on https://github.com/ParadiseSS13/Paradise/pull/13181 first

Fox-McCloud:
Given the numerous tweaks to addiction and the removal of alcohol addiction, would it be ok if I closed this?

Aquilar:
Through we may need the general concept again later. I haven't adapted it so yeah, close it.

PR #13117, opened 2020-03-14
Rework power outage event

blinkdog:
## What Does This PR Do
This PR is a rework of the power outage event called `Grid Check`.

As background, there are three levels of events: Mundane, Moderate, and Major. These are listed in increasing order of severity and decreasing order of frequency. The `Grid Check` event was a Moderate level event. It would temporarily drain the SMES of power for a random period of time, then restore the power to the SMES after the event was over. There were some shortcomings to this event, as noted in the Why It's Good For The Game section below.

This rework replaces `Grid Check` with two new events:
* Moderate Event: `APC Short`
* Major Event: `APC Overload`

**APC Short**
`APC Short` is a Moderate level event that may be selected at random. It has the same probability of being selected as the original `Grid Check`. The differences between the events are as follows:

- The event no longer affects the SMES.
- Each APC has a 25% chance of being shorted out (power wires inside getting cut).
- In addition to the engine, the AI is added to the list of "safe" areas that will not have their APCs shorted out by the event.
- The event does not have a definite end time or fix itself. Instead, the APCs are repaired if/when engineering decides to fix them.

_Fixing APC Short_
The following steps can be taken to repair an APC that has been shorted out by this event.

0. Wear Insulated Gloves
1. Apply Screwdriver to the APC
2. Apply Wirecutter to the APC
3. [Mend] both broken wires (color is random)
4. Apply Screwdriver to the APC
5. Apply Engineering ID to the APC
6. [On] the Main Breaker of the APC
7. Apply Engineering ID to the APC

**APC Overload**
`APC Overload` is a Major level event that is NOT selected at random. It may be fired manually by an Admin if they choose to inflict it upon the station. It is similar to `APC Short` but is harder to fix.

- The event does not affect the SMES.
- Each APC has a 25% chance of having it's underfloor terminal destroyed.
- In addition to the engine, the AI is added to the list of "safe" areas that will not have their APCs damaged by the event.
- The event does not have a definite end time or fix itself. Instead, the APCs are repaired if/when engineering decides to fix them.

_Fixing APC Overload_
The following steps can be taken to repair an APC that has been damaged by this event.

0. Wear Insulated Gloves
1. Apply Crowbar to Floor Tile in front of the APC
2. Apply Engineering ID to the APC
3. [Disengaged] the Cover Lock of the APC
4. Apply Crowbar to the APC
5. Apply Cable Coils (10 stack) to the APC
6. Apply Crowbar to the APC
7. [On] the Main Breaker of the APC
8. Apply Engineering ID to APC
9. Apply Floor Tile to Plating in front of the APC

Other minor changes:
- A tiny bit of incorrect grammar in the Examine message for the APC caught my eye; fixed that message.
- Removed "holder" variables in SMES as they are no longer used/needed by the event.
- Updated the announcement text displayed when the event begins.
- Updated the voice samples played when the event begins.

Some things to note:
- As of this PR, there are 154 APCs that can be affected by the event. At 25% probability, there will be roughly ~40 APCs that are affected by either event.
- APC Overload is more difficult to diagnose because of the underfloor wiring. It also requires the engineer to bring a stack of 10 Cable Coils in order to repair the APC. At ~40 APCs getting damaged, this means it will require about ~400 Cable Coils to repair all the damage caused by the major event.

Shortcomings:
- I have replaced the voice sounds, but I am ambivalent about how well they fit.
- I would love to be able to do the new announcement text in the original voice; if anybody can help with that, please reach out.

## Why It's Good For The Game
The current 'power outage' event gets ignored, because it is usually very short, nothing can be done about it, and if you've hotwired the grid it has no impact. Rework it so that it is longer, can be actively dealt with, and cannot be completely prevented in advance. For example, rework it so that instead of draining the grid, it instead flips the main breaker to OFF for say 25% of APCs on the station. This would be impossible to ignore, give eng/AI something to do in fixing it, and not be something the crew could prevent in advance by metagame hotwiring tactics.

The above text is copied verbatim from [Kyet's Changes Wanted](https://www.paradisestation.org/forum/topic/18118-kyets-changes-wanted/) 

## Changelog
:cl:
add: Power outage event has been completely reworked
/:cl:


datlo:
I feel like simply flipping the breaker is a bit too easy to fix, especially for an AI. This event is also an opportunity for antags to break into places by crowbarring : Not knowing that the place you wanna go in is open makes this a less interesting event for that.

Kyep:
Could you make it so that the deactivated APCs just remain off (main breaker is off, or wire that makes them shorted is cut) until someone fixes them? IE: the outage doesn't fix itself?

Eng needs more to do, and I feel like the current outage events just get ignored for the ~2 minutes it takes to fix them. We need something that engineers have to fix, or at least get the AI to fix, to be a meaningful event engineering-wise.

TDSSS:
I think it would be better if  every or most APCs were turned off, and say 25% of them didn't turn back on when the event ends. That way, crew and AI have something to do either way, and antags hav emore chances to break into places.

blinkdog:
> Could you make it so that the deactivated APCs just remain off (main breaker is off, or wire that makes them shorted is cut) until someone fixes them? IE: the outage doesn't fix itself?

I added a compile time flag: `#define EVENT_FIXES_ITSELF_AT_END FALSE`

If somebody decides in the future they want the event to fix itself, just change that FALSE to TRUE, and the event will call the fix itself part. By adding this, 25%(ish) of the APCs get switched off and they don't turn back on until somebody manually turns the main breaker back on at the APC panel.

blinkdog:
> I think it would be better if every or most APCs were turned off, and say 25% of them didn't turn back on when the event ends. That way, crew and AI have something to do either way, and antags hav emore chances to break into places.

The probability is defined at the top of the event file: `#define APC_TOGGLE_OFF_PROBABILITY 25`

Pretty easy to change, but I'd like Kyet to weigh in with the final number before I add a commit to change it.

code/modules/events/grid_check.dm
```diff
@@ -1,11 +1,20 @@
-/datum/event/grid_check	//NOTE: Times are measured in master controller ticks!
-	announceWhen		= 5
+#define ANNOUNCE_DELAYED FALSE
+#define ANNOUNCE_NOW TRUE
+
+#define ANNOUNCE_AFTER_MC_TICKS 5
+#define APC_TOGGLE_OFF_PROBABILITY 25
+#define EVENT_FIXES_ITSELF_AT_END FALSE
+#define EVENT_MAX_LENGTH_MC_TICKS 150
+#define EVENT_MIN_LENGTH_MC_TICKS 60
```
AffectedArc07: These should be variables on the event object, not defines. They should also have less ambiguous naming, even though they are `#undef`'d at the EOF.

code/modules/events/grid_check.dm
```diff
@@ -1,11 +1,20 @@
...
+/datum/event/grid_check
+	announceWhen = ANNOUNCE_AFTER_MC_TICKS
 
 /datum/event/grid_check/setup()
-	endWhen = rand(30,120)
+	endWhen = rand(EVENT_MIN_LENGTH_MC_TICKS, EVENT_MAX_LENGTH_MC_TICKS)
 
 /datum/event/grid_check/start()
-	power_failure(0)
+	power_failure(ANNOUNCE_DELAYED)
```
AffectedArc07: Possibly change this to `power_failure(announce=ANNOUNCE_DELAYED)` for ease of readability.

code/modules/events/grid_check.dm
```diff
@@ -17,60 +26,51 @@
 	event_announcement.Announce("Abnormal activity detected in [station_name()]'s powernet. As a precautionary measure, the station's power will be shut off for an indeterminate duration.", "Automated Grid Check", new_sound = 'sound/AI/poweroff.ogg')
 
 /datum/event/grid_check/end()
-	power_restore()
+	if(EVENT_FIXES_ITSELF_AT_END)
+		power_restore()
+
+/proc/power_failure(var/announce = ANNOUNCE_NOW)
```
AffectedArc07: No `var/` here

code/modules/events/grid_check.dm
```diff
@@ -17,60 +26,51 @@
...
-			C.cell.charge = 0
+		// otherwise, make a note to remember if we need to turn it back on
+		C.last_operating = C.operating
+		if(prob(APC_TOGGLE_OFF_PROBABILITY) && (C.operating))
+			C.toggle_breaker()
 
-/proc/power_restore(var/announce = 1)
-	var/list/skipped_areas = list(/area/turret_protected/ai)
-	var/list/skipped_areas_apc = list(/area/engine/engineering)
+/proc/power_restore(var/announce = ANNOUNCE_NOW)
```
AffectedArc07: Ditto

code/modules/events/grid_check.dm
```diff
@@ -17,60 +26,51 @@
...
-		S.output_attempt = S.last_output_attempt
-		S.input_attempt = S.last_input_attempt
-		S.update_icon()
-		S.power_change()
+		// if it was operating before the event, turn it back on
+		if(C.last_operating && (C.operating == FALSE))
+			C.toggle_breaker()
 
-/proc/power_restore_quick(var/announce = 1)
+/proc/power_restore_quick(var/announce = ANNOUNCE_NOW)
```
AffectedArc07: Ditto

code/modules/events/grid_check.dm
blinkdog: I have made some changes here. Please review and resolve if the changes are OK?

blinkdog:
I made some changes to improve the code as suggested by @AffectedArc07 

>@TDSSS
>I think it would be better if every or most APCs were turned off, and say 25% of them didn't turn back on when the event ends. That way, crew and AI have something to do either way, and antags hav emore chances to break into places.

I added an `APC_TOGGLE_ON_PROBABILITY` to help support this as an option. For example, if the toggle off probability was set to 100, the toggle on probability set to 75, and `event_fixes_itself_at_end` was set to TRUE, then it would operate as @TDSSS has described above.

**As It Stands Now**: The event operates as @Kyep has specified; 25% of the APCs turn off when the event starts, and none of them are automatically fixed when the event ends.

Denthamos:
Hmmn, taking a page from Rimworld what if the grid check event also would fry/short out power cells in some APC's and SMES's, resulting in a need to replace them by engineers.

blinkdog:
> Hmmn, taking a page from Rimworld what if the grid check event also would fry/short out power cells in some APC's and SMES's, resulting in a need to replace them by engineers.

I like this idea. It makes me wonder if the event might be split into three difficulty levels to match the three event categories (mundane, moderate, major). A major power outage could involve cells being fried and needing replacement, while a mundane power outage could be just a few APCs tripping a circuit breaker and fixing themselves in relatively short order.

I'll let @Kyep weigh in if that's for this PR or better saved for another PR.

Kyep:
@blinkdog 
Lately I've been thinking that we should make the event break one of the shorting wire in the APC. That way it would require an engineer to fix - wouldn't be fixable by AI.
However, lately I've also noticed a lot of shifts where engineering is useless, and having your workplace powered down the whole shift because eng is useless is not fun gameplay.

So, for the moment, I support the event as is: flips main breaker in 25% of station APCs, requiring either hacking, an engineer, OR the AI to fix. That way it does give eng something to do at least sometimes, while not being an utter disaster if engineering is incompetent.

Yes, we could split it into several events, but I'd rather just have this one event in rotation to start with - see how the playerbase reacts to that, then consider expanding it into multiple events of different levels based on that.

VEN0M5:
I can see this outrage being fixed in 2 minutes by the AI as it stands.
1) check power grid monitor and note the apcs that are off.
2) go around and turn everything back on.

Kyep:
Yes, I worry about that too. But I'd rather have it be under-disruptive than over-disruptive.
If we jump straight to making it cut shorting wires / burned out power cells that engineering must fix... then yes we force engi to do stuff, and I do want that... but I worry for the consequences during shifts that engineering is incompetent. Spending half your shift with a depowered workplace because engineers never turn up to fix it is no fun.

Kyep:
@blinkdog 
Okay, after discussion with @Necaladun and other admins, we'll go whole hog on this.
Please modify the event so that it isn't possible for AIs to fix the broken APCs. Make it so that the event cuts one of the 'short' wires in the APC, breaking the APC until an engineer with gloves can open the APC and 'mend' the wire.

blinkdog:
> Please modify the event so that it isn't possible for AIs to fix the broken APCs. Make it so that the event cuts one of the 'short' wires in the APC, breaking the APC until an engineer with gloves can open the APC and 'mend' the wire.

Okay, I'll work on modifying the event this weekend.

blinkdog:
@Kyep I've modified the event such that the APCs have their internal power wires cut, and the main breaker toggled off. Fixing the APC requires an engineer with an ID that can unlock APCs, insulated gloves, a screwdriver, and a wirecutter.

I've also coded another version of this event that I've called `apc_overload`. I've added it as a MAJOR level event, but commented-out (non-functional) in the code. Like the `grid_check` event, it affects 25% of APCs, but the APC Overload version:

* Discharges the APC cell completely
* Disconnects and deletes the terminal associated with the APC
* Toggles off the main power breaker

This APC Overload version of the event is pretty nasty. You need a crowbar to pry up the floor, 10 cable coils per APC to reconnect it to the grid, and the cell has to recharge from zero. I have tested the event and it does work, but **I have not enabled it in the code as it stands right now**. If you also want this version, the whole hog with a side of bacon, sausage, and Canadian bacon, it's pretty easy to uncomment the one line to make it a working major event. Just let me know.

Kyep:
@blinkdog can you leave that major event enabled in the code, but with 0 weight? That way it can be at least be forced for admin/testing purposes.

code/modules/events/apc_overload.dm
```diff
@@ -0,0 +1,63 @@
...
+/datum/event/apc_overload/end()
+	return TRUE
+
+/proc/apc_overload_failure(announce=TRUE)
+	var/list/skipped_areas_apc = list(
+		/area/engine/engineering,
+		/area/turret_protected/ai)
+
+	if(announce)
+		GLOB.event_announcement.Announce("Abnormal activity detected in [station_name()]'s powernet. As a precautionary measure, the station's power will be shut off for an indeterminate duration.", "Critical Power Failure", new_sound = 'sound/AI/poweroff.ogg')
```
Kyep, line 35: Use:
Overload detected in [station_name()]'s powernet. Engineering, please replace destroyed APCs.

code/modules/events/grid_check.dm
```diff
@@ -14,63 +21,50 @@
...
-		S.last_charge			= S.charge
-		S.last_output_attempt	= S.output_attempt
-		S.last_input_attempt 	= S.input_attempt
-		S.charge = 0
-		S.inputting(0)
-		S.outputting(0)
-		S.update_icon()
-		S.power_change()
+	if(announce)
+		GLOB.event_announcement.Announce("Abnormal activity detected in [station_name()]'s powernet. As a precautionary measure, the station's power will be shut off for an indeterminate duration.", "Systems Power Failure", new_sound = 'sound/AI/poweroff.ogg')
```
Kyep, line 35: Use:
Overload detected in [station_name()]'s powernet. Engineering, please repair damaged APCs.

blinkdog:
> A few updates to messaging are needed.
> 
> Also, you need to pull the latest updates from master - you're based on an out-of-date master which creates issues for testing during testmerges.

@Kyep these latest commits should:
- Update this branch to the latest master, so to work around that do_after bug
- Fix up the messaging to what you've specified
- Add APC Overload as a MAJOR level event with zero weight; it should never be randomly selected, but admins can fire it through the event panel.

Please let me know if any further changes are needed before the next test merge.

Kyep:
- Events still use voice sound for "the station's power will be shut off for an indeterminate duration". Either use a new updated voice sound for the new text, or the generic ping sound.
- Major event does indeed break the terminal. Can you list for the record the steps needed to fix an APC with a destroyed terminal?
- Also, I note for the record that each event seems to affect ~40 APCs.

Ionward:
Bug during test merge on April 10th, there were two APC outages, and more during the second than the first tons of APCs had their power terminals deleted, under the tiles. I sat there like a goober for a good few minutes messing with something under a table until realizing that there was no terminal to get the grid power into the APC 😄

nl-208:
@Ionward - the second event was an APC overload, which is meant to delete the terminals under the APCs - it's probably working as intended.

TDSSS:
Maybe we should have a "destroyed terminal" sprite or something, showing a burnt out terminal?

blinkdog:
>     * Events still use voice sound for "the station's power will be shut off for an indeterminate duration". Either use a new updated voice sound for the new text, or the generic ping sound.

I have made some new sounds, but I am ambivalent about them.
Let me know if we should keep them or go with the generic sound?
 
>     * Major event does indeed break the terminal. Can you list for the record the steps needed to fix an APC with a destroyed terminal?

I have updated the PR description at the top to reflect the most recent incarnation of the events. Included is a list of steps to repair an APC under both scenarios.

>     * Also, I note for the record that each event seems to affect ~40 APCs.

Glad to hear this. 154 * 0.25 = 38.5 ... so ~40 is about how many APCs we should see at 25%.

blinkdog:
> (as a side note, if anyone wanted to make the messaging for overloads even clearer, could always put something like "**engineering, please check all underfloor APC terminals**" - may help avoid confusion with people thinking the APCs were destroyed outright.)

I love this wording and adopted it for the APC Overload event. Thank you for this.

Kyep:
This was TM'ed the other day. It went OK, I think this really needs a full merge and a period of evaluation to see where we go from there.

PR #13118, opened 2020-03-14
Cyborg recharger now throws occupants out when destroyed and heals IPCs even when they are not hungry

farie82:
## What Does This PR Do
As the title says:
Cyborg recharger now throws occupants out when destroyed and heals IPCs even when they are not hungry

## Why It's Good For The Game
One is a bug that puts people in null space and the other is just odd.

## Changelog
:cl:
tweak: Cyborg recharger stations now heal IPCs even if they are not hungry (and the station is upgraded)
fix: Cyborg recharger stations now throw the occupant out when destroyed
/:cl:

### DIFF
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -15,7 +15,7 @@
 
 /obj/machinery/recharge_station/Destroy()
 	go_out()
-	. = ..()
+	return ..()
 
 /obj/machinery/recharge_station/New()
 	..()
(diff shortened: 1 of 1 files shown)


code/game/machinery/rechargestation.dm
```diff
@@ -13,6 +13,10 @@
 	var/recharge_speed_nutrition
 	var/repairs
 
+/obj/machinery/recharge_station/Destroy()
+	go_out()
+	. = ..()
```
Fox-McCloud: please stick with `return ..()`

Unless you need to return the value prior to further procs, stick with explicit returning.

farie82: Done

PR #13120, opened 2020-03-14
Alphabetizes Access on Electronics.

Mitchs98:
## What Does This PR Do
Right what it says on the tin. The access list on airlock electronics are, finally, alphabetized. No more wading through a large list to find the access you need. If you know the alphabet you'll be able to find it pretty quick.

## Why It's Good For The Game
It's annoying as hell to have to sort through the electronic list for an access you need with them jumbled all over the place. This makes it so that all access' starting with S and etc. are alphabetized fully. No more finding AI upload at the bottom of the list for example(even if that didn't happen exactly).

## Changelog
:cl: Mitchs98
tweak: Access lists for electronics are now alphabetized properly.
/:cl:


Mitchs98:
> You are mixing tabs and spaces. Don't.

Eh? I didn't use any spaces for lines, at-least it shouldn't of. I never use spaces for those 🤔

shawnhind:
I was just getting really frustrated about this the other day. I needed to quickly repair a door and it took me way too long (like a minute or 2) to find the access I was looking for.  It's even harder to find when you're under pressure and trying to read comms at the same time.

Yes 100x to this PR :)

code/game/jobs/access.dm
```diff
@@ -129,18 +129,15 @@
...
-	            ACCESS_EXTERNAL_AIRLOCKS, ACCESS_CHANGE_IDS, ACCESS_AI_UPLOAD,
-	            ACCESS_TELEPORTER, ACCESS_EVA, ACCESS_HEADS, ACCESS_CAPTAIN, ACCESS_ALL_PERSONAL_LOCKERS,
-	            ACCESS_TECH_STORAGE, ACCESS_CHAPEL_OFFICE, ACCESS_ATMOSPHERICS, ACCESS_KITCHEN,
-	            ACCESS_BAR, ACCESS_JANITOR, ACCESS_CREMATORIUM, ACCESS_ROBOTICS, ACCESS_CARGO, ACCESS_CONSTRUCTION,
-	            ACCESS_HYDROPONICS, ACCESS_LIBRARY, ACCESS_LAWYER, ACCESS_VIROLOGY, ACCESS_PSYCHIATRIST, ACCESS_CMO, ACCESS_QM, ACCESS_CLOWN, ACCESS_MIME, ACCESS_SURGERY,
-	            ACCESS_THEATRE, ACCESS_RESEARCH, ACCESS_MINING, ACCESS_MAILSORTING,
-	            ACCESS_HEADS_VAULT, ACCESS_MINING_STATION, ACCESS_XENOBIOLOGY, ACCESS_CE, ACCESS_HOP, ACCESS_HOS, ACCESS_RC_ANNOUNCE,
-	            ACCESS_KEYCARD_AUTH, ACCESS_TCOMSAT, ACCESS_GATEWAY, ACCESS_XENOARCH, ACCESS_PARAMEDIC, ACCESS_BLUESHIELD, ACCESS_MECHANIC,ACCESS_WEAPONS,
-	            ACCESS_PILOT, ACCESS_NTREP, ACCESS_MAGISTRATE, ACCESS_MINERAL_STOREROOM, ACCESS_MINISAT, ACCESS_NETWORK)
+	return list(ACCESS_MINISAT, ACCESS_AI_UPLOAD,  ACCESS_ARMORY, ACCESS_ATMOSPHERICS, ACCESS_BAR, ACCESS_SEC_DOORS, ACCESS_BLUESHIELD,
```
farie82: weird that this is a proc to begin with and not a define.
You can perhaps make a global list and have it initialised lazily. Then upon actual usage you sort the accesses in there and then set the global list to this. Further uses will just return this list then.
Might be out of the scope of this PR but it's another option that is future proof if you want to have a small challenge

PR #13121, opened 2020-03-14
Adds an extra handheld defib to medical lockers.

Mitchs98:
## What Does This PR Do
Adds another handheld defib to all medical lockers except CMO/Paramedics. 

## Why It's Good For The Game
The current amount of handheld defibs does not support a full medbay. You get four lockers with one handheld defib in each of them with 6-8 doctor slots(unsure on the exact number). While it's true you CAN get RND to print them, handheld defibs were designed to be an extremely important part of newcrit and RND is far from reliable enough to expect them to be able to print it(tech wise) or even be able to print it(know how/are around to do so).

You can still be SOL if someone decides to steal them/loses them for whatever reason, but at the very least now you can spawn into a hectic round and be able to gear up properly.

## Changelog
:cl:
add: Adds one extra handheld defibrillator to all standard MD lockers.
/:cl:


datlo:
An alternative would be to add one to medical vendors. So theres two ways to get one. Would be nice since med vendors' access can be hacked while lockers cant. This is still good though. :+1:

Mitchs98:
Didn't even think of that as a possibility. I'll do whichever maints/heads prefer, really. Either one is definitely acceptable and would virtually be the same thing.

variableundefined:
Conflicts

PR #13134, opened 2020-03-17
Ratvar spear does burn damage

### LINKED ISSUE #12570: Ratvarian spear doesn't do extra burn damage
**Issue Description**:

The ratvarian spear doesn't do 5 extra burn damage like it says it does

for it to do the extra burn damage it should

**What happened instead**:
It doesn't do extra burn damage

It makes the loot less valuable

attack with spear, check target that's been hit. notice there is no burn damage

Probably since lavaland was ported?

**Extra information**:
the comment in the code says
``	force = 15 //Extra damage is dealt to targets in attack()``
but there is no code for the spears attack proc here

TDSSS:
## What Does This PR Do
Fixes #12570
However, I must admit my current naive solution has a bit of an issue in that the bonus burn damage isn't inflicted to the same zone as the brute. 

As far as I can tell there's really no way to fix that without snowflaking something into human defense code just for this specific item, or creating a dedicated 'deal multiple damage types at once' system which would amount to the same thing until more such items are created. 

If there is an obvious way to do it that I missed, that would of course be welcome, too.

Considering we're talking about a single, rarely seen lavaland loot item I don't think it's worth the effort and added complexity, though maints are of course free to disagree.

## Why It's Good For The Game
Less bugs good

## Changelog
:cl:
fix: ratvar spear now does burn damage on top of brute as intended
/:cl:

### DIFF
--- a/code/modules/ruins/lavalandruin_code/dead_ratvar.dm
+++ b/code/modules/ruins/lavalandruin_code/dead_ratvar.dm
@@ -258,3 +258,8 @@
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	w_class = WEIGHT_CLASS_BULKY
 	var/bonus_burn = 5
+
+/obj/item/clockwork/weapon/ratvarian_spear/attack(mob/living/M, mob/living/user, def_zone)
+	. = ..()
+	if(isliving(M))
+		M.apply_damage(bonus_burn, BURN, def_zone, used_weapon = src)


TDSSS:
Had to update, this should pass travis now.

TDSSS:
Or...not

Spacemanspark:
Would it be too out of scope to put the in hand sprites for the spear? Currently it's just invisible in hand

TDSSS:
Is there an existing inhand sprite I could use?

Spacemanspark:
Probably on the tg code base, I'd assume at the very least

code/modules/ruins/lavalandruin_code/dead_ratvar.dm
```diff
@@ -258,3 +258,8 @@
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	w_class = WEIGHT_CLASS_BULKY
 	var/bonus_burn = 5
+
+/obj/item/clockwork/weapon/ratvarian_spear/attack(mob/living/M, mob/living/user, def_zone)
+	. = ..()
+	if(isliving(M))
```
Crazylemon64: Might want to check the return result here so that stuff like pacifism works properly

TDSSS: right, I hadn't considered pacifism, will look into this.

Fox-McCloud:
Please update or close---since Clockcult is fully dead on TG, they just altered this spear to be a null rod option and pre-spawned it on the map; it's nothing special, to be honest; I'm fine with this just being closed, as well, on the basis of "the extra damage is explicitly a clockcult feature that is never going to happen anyway"

TDSSS:
Yeah I ran into some issues doing the return CL suggested (maybe I'm just dumb) and I honestly have never used this spear before or seen it used so meh.

Going to just close it, if someone else wants to look into getting it to work correctly with pacifism, be my guest.

PR #13138, opened 2020-03-18
Advanced Scanner print-outs now show if an organ is dead.

### LINKED ISSUE #6792: Terror Spider Hallucigen Debuff persists after cloning
I'm not 100% on how terror spiders work, but I do know they do something that makes you hallucinate. I got hit by it, died, got cloned, and still hallucinated.

### LINKED ISSUE #6972: Adv Scanner Printouts do not include "dead" status on limbs and organs
**Problem Description**:
Adv Scanner Printouts do not include "dead" status on limbs and organs

That it would state if something is dead. It shows all the other ailments as far as I know

**What happened instead**:
It doesn't print that a part is dead

If there are several dead organs, it requires the person to memorize it.

Have dead organ. Scan. Print. Not dead anymore! Oh wait...

Mitchs98:
## What Does This PR Do
Fixes #6972 https://github.com/ParadiseSS13/Paradise/issues/6972. Advanced Scanner print outs will now properly report if an organ or limb is dead. It will also now properly report Septic organs, previously it only(for some reason) reported up to Acute Infection++. 

## Why It's Good For The Game
Print outs can be properly relied on for infection levels and limb/organ status so doctors don't have to 1. Constantly scan a patient/memorize what's dead or 2. Get a print out, be in a rush due to swamped medbay, and see the print out reports nothing is dead and thus they don't debride/revive the organs causing a lot of issues.

## Changelog
:cl: Mitchs98
fix: Advanced Scanner print outs now properly display dead organs and limbs.
fix: Advanced Scanners now properly show septic organs.
/:cl:

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -479,7 +479,7 @@
 			if(i.status & ORGAN_DEAD)
 				dead = "DEAD:"
 			switch(i.germ_level)
-				if(INFECTION_LEVEL_ONE to INFECTION_LEVEL_ONE + 200)
+				if(1 to INFECTION_LEVEL_ONE + 200)
 					infection = "Mild Infection:"
 				if(INFECTION_LEVEL_ONE + 200 to INFECTION_LEVEL_ONE + 300)
 					infection = "Mild Infection+:"
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -457,9 +457,9 @@
 					infected = "Acute Infection:"
 				if(INFECTION_LEVEL_TWO + 200 to INFECTION_LEVEL_TWO + 300)
 					infected = "Acute Infection+:"
-				if(INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 400)
+				if(INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 399)
 					infected = "Acute Infection++:"
-				if(INFECTION_LEVEL_THREE to INFINITY)
+				if(INFECTION_LEVEL_TWO + 400 to INFINITY)
 					infected = "Septic:"
 
 			var/unknown_body = 0
@@ -489,9 +489,9 @@
 					infection = "Acute Infection:"
 				if(INFECTION_LEVEL_TWO + 200 to INFECTION_LEVEL_TWO + 300)
 					infection = "Acute Infection+:"
-				if(INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 400)
+				if(INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 399)
 					infection = "Acute Infection++:"
-				if(INFECTION_LEVEL_THREE to INFINITY)
+				if(INFECTION_LEVEL_TWO + 400 to INFINITY)
 					infection = "Septic:"
 
 			dat += "<tr>"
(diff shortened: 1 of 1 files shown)


code/game/machinery/adv_med.dm
```diff
@@ -483,11 +489,13 @@
 					infection = "Acute Infection:"
 				if(INFECTION_LEVEL_TWO + 200 to INFECTION_LEVEL_TWO + 300)
 					infection = "Acute Infection+:"
-				if(INFECTION_LEVEL_TWO + 300 to INFINITY)
+				if(INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 400)
 					infection = "Acute Infection++:"
+				if(INFECTION_LEVEL_THREE to INFINITY)
+					infection = "Septic:"
```
SteelSlayer: It turns out this is actually a problem. You have code to handle 800-900 and 1000-INFINITY but nothing to handle 900-1000. So while germ level is between 900 and 1000, the print out shows nothing about infection.

image

Mitchs98: Ironically this is how it's always been for limbs, so this means this entire time you could get really unlucky and have 920 infection on your chest and not even be aware of it VIA scanner. Beautiful. I'll figure out how to fix this tomorrow, shouldn't be that hard.

Mitchs98:
Sorry for the delay on that. Though now scanners are fully in line with actual infection code. Turns out, 900 germ level means it's septic. Not 1000. This didn't affect the scanner itself, only its printout, but now it's entirely accurate at every stage of infection ^^

code/game/machinery/adv_med.dm
```diff
@@ -483,11 +489,13 @@
 					infection = "Acute Infection:"
 				if(INFECTION_LEVEL_TWO + 200 to INFECTION_LEVEL_TWO + 300)
 					infection = "Acute Infection+:"
-				if(INFECTION_LEVEL_TWO + 300 to INFINITY)
+				if(INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 399)
 					infection = "Acute Infection++:"
+				if(INFECTION_LEVEL_TWO + 400 to INFINITY)
+					infection = "Septic:"
```
Crazylemon64: It might be helpful to add a case between "Acute Infection++" and "Septic", since iirc mitocholide is needed once sepsis hits

Mitchs98: 900 is when sepsis hits naturally in code, so for that I'd need to refactor code entirely or the print out itself would be inaccurate. Though mitocholide isn't needed for anything other than dead organs as far as I'm aware and remember with recent incidents.

Crazylemon64: Are you sure? `necrotize` is called past INFECTION_LEVEL_THREE, which is at 1000 germs - and INFECTION_LEVEL_TWO is at 500 germs, so INFECTION_LEVEL_TWO+400 is 900 germs - which leaves a window of 100 germs where it would be displayed as "septic" despite not actually having necrotized yet.

Mitchs98: I'm quite sure, yes. I had to test germ levels fairly extensively, when I put germ level 900 on a limb it was septic, 900 being the first germ level it occurs at. 890-899 was acute affection++. It says this in the advanced scanner and did before I changed the code, the code that you're reading now only affects the print out not what the scanner tells you.

GeVCuepXy7

This image is taken from my master without the scanner changes.

Crazylemon64: Hmm, I guess it'd make sense in parity with the body scanner's nano interface. Sounds good!

PR #13139, opened 2020-03-18
isSynthetic() proc now returns TRUE for silicon mobs

Gatchapod:
## What Does This PR Do
Makes all silicon mobs return TRUE when running isSynthetic() proc. The proc works the same way as before for human mobs, except returns FALSE instead of 0, which I believe is the Paradise standard.

With an actual difference between ismachine() and isSynthetic() procs, adjusts the code in a few places where it may matter.
Majority of isSynthetic() calls are made by /living/carbon/human mobs. Since this PR does not affect how the proc works for carbons, I did not touch them, even if nearly all the cases could be replaced with plain ismachine() procs.

## Why It's Good For The Game
Without this change, we might as well replace all instances of isSynthetic() with ismachine() and remove the former, because they do exactly the same thing. In fact, isSynthetic() does basically nothing outside of calling for ismachine().

I'd rather have this change, as I'm currently working on some minor changes and improvements to the cyborg rechargers. Having a proc that returns true for silicons and IPC would make the code significantly more readable and, if I'm not mistaken, efficient.

The only problem that may arise, as far as I can tell, is that changes to the swarmer trap make unfolded pAIs get affected by it (and AIs, but if AI somehow runs into a swarmer trap, this bit of code is the least broken part). They get properly stunned and can be destroyed by swarmers, but a pAI can technically be used to destroy swarmer minefields. I think using pAIs this way is so unlikely and inefficient it does not warrant addressing, but it can still be seen as a bug and if deemed as such, I will have to either make isSynthetic() call for isrobot() or exclude pAIs from swarmer traps manually. Oh, and Lavaland swarmers obviously hate pAIs, so it doesn't really affect miners.

## Changelog
:cl:
tweak: isSynthetic() and ismachine() are now two different procs
tweak: Unfolded pAIs are now affected by swamer traps
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -61,7 +61,7 @@ proc/get_radio_key_from_channel(var/channel)
 	return default_language
 
 /mob/living/proc/handle_speech_problems(list/message_pieces, var/verb)
-	var/robot = isSynthetic()
+	var/robot = ismachine()
 	for(var/datum/multilingual_say_piece/S in message_pieces)
 		if(S.speaking && S.speaking.flags & NO_STUTTER)
 			continue
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -61,7 +61,7 @@ proc/get_radio_key_from_channel(var/channel)
 	return default_language
 
 /mob/living/proc/handle_speech_problems(list/message_pieces, var/verb)
-	var/robot = ismachine()
+	var/robot = ismachine(src)
 	for(var/datum/multilingual_say_piece/S in message_pieces)
 		if(S.speaking && S.speaking.flags & NO_STUTTER)
 			continue
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
@@ -604,7 +604,7 @@
 		if(!istype(L, /mob/living/simple_animal/hostile/swarmer))
 			playsound(loc,'sound/effects/snap.ogg',50, 1, -1)
 			L.electrocute_act(0, src, 1, TRUE, TRUE)
-			if(isrobot(L) || L.isSynthetic())
+			if(L.isSynthetic())
 				L.Weaken(5)
 			qdel(src)
 	..()
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -13,12 +13,15 @@
 	return 0
 
 /mob/proc/isSynthetic()
-	return 0
+	return FALSE
 
 /mob/living/carbon/human/isSynthetic()
 	if(ismachine(src))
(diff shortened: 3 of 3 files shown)


code/modules/mob/living/say.dm
```diff
@@ -61,7 +61,7 @@ proc/get_radio_key_from_channel(var/channel)
 	return default_language
 
 /mob/living/proc/handle_speech_problems(list/message_pieces, var/verb)
-	var/robot = isSynthetic()
+	var/robot = ismachine()
```
SteelSlayer: `ismachine()` needs an argument, `src` in this case.

AffectedArc07: What steel said, this needs an argument.

Gatchapod:
Yup, I am stupid. I don't even know how I did it.
Possibly forgot to compile after I changed the slurring bit.

Gatchapod:
Regarding the technically correct assumption the code wasn't tested before, it seems like I have made the slurring change as an afterthought and forgot to compile it.

After correcting the mistake and testing the actual PR, not an unfinished version, it behaves as described above. I think keeping isSynthetic() proc in the slurring bit might even be desirable.
Currently, cyborgs can't slur, stutter or anything like that outside of var editing, but if an admin does that, they will use regular versions, not robotic.

SteelSlayer:
Just a heads up, I'm probably going to make a PR to kill off `isSynthetic()` in exchange for just using `ismachine()`, I wouldn't worry about refactoring it too much.

farie82:
> Just a heads up, I'm probably going to make a PR to kill off `isSynthetic()` in exchange for just using `ismachine()`, I wouldn't worry about refactoring it too much.

This PR changed the use of isSynthetic to be both ismachine and isrobot on a mob level. It would make checking mobs if they are robits easier. Since now you gotta do: ismachine || issilicon

SteelSlayer:
Yeah alright that will make the proc actually useful then.

Gatchapod:
After going through the instances of isSynthetic() and ismachine() in the code carefully, I have found only one case where it actually improves anything.

> 
> 
> There are some remaining uses of isSynthetic that can use a refactor.
> tiny_prick.dm, swarmer.dm and scanners.dm
> All of those above use checks that can be simplified

`tiny_pric.dm` benefits more from replacing `isSynthetic()` with `ismachine()`. It already checks if the target is a carbon mob, which excludes silicons and simple animals. Replacing both checks with `isSynthetic()` would require an additional check for simple animals, leading to no actual code improvement.

`swarmer.dm` suffers from `isSynthetic()` by making pAIs valid targets for swarmer traps. Not exactly a desirable feature.

`scanners.dm` is exactly the same case as `tiny_prick.dm`. Replacing the check you're thinking about would require a special check for simple animals. Without it, you can use scanner on simple animals as if they were carbon mobs.

The PR was, after all, a poorly thought through stepping stone. I am closing it and most likely the next thing I'm doing is removing the proc.

PR #13143, opened 2020-03-18
Makes classic secHUD available to Magistrate

NotDhu:
## What Does This PR Do
Adds magistrate to the classic security HUD whitelist in the character loadout.

## Why It's Good For The Game
IAA are already able to spawn with a classic security HUD, magistrates should be able to as well.

## Changelog
:cl:
tweak: Magistrates can now use the classic sechud in the loadout options.
/:cl:

### DIFF
--- a/code/modules/client/preference/loadout/loadout_general.dm
+++ b/code/modules/client/preference/loadout/loadout_general.dm
@@ -29,7 +29,7 @@
 /datum/gear/sechud
 	display_name = "a classic security HUD"
 	path = /obj/item/clothing/glasses/hud/security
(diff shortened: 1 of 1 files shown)


TheSardele:
Wait a minute, IAAs can spawn with classic sec huds? Can they use this to spawn with those, and get around the fact their normal sec huds are read only?

AffectedArc07:
@TheSardele I __think__ its based on access, not the actual HUD item.

TheSardele:
https://github.com/ParadiseSS13/Paradise/blob/04ba5c1cc902bf2eada751de7b1b72ebc64ecd86/code/modules/clothing/glasses/hud.dm#L123

Looks like it's an actual item

AffectedArc07:
Huh so maybe that is the case

PR #13150, opened 2020-03-19
Adds confirmation prompt to kicking

TheSardele:
## What Does This PR Do
Adds a confirmation alert for admins trying to kick people.

## Why It's Good For The Game
I don't have to feel like shit when my hand slips and I accidentally kick someone.

## Changelog
:cl:
add: Adds a confirmation prompt for admins when kicking people
/:cl:


FreeStylaLT:
Do we... Do we kick people?

MRgaty:
"My hands slips" yeah sure... OwO

code/modules/admin/topic.dm
```diff
@@ -902,14 +902,18 @@
...
 		if(ismob(M))
-			if(M.client && M.client.holder && (M.client.holder.rights & R_BAN))
-				to_chat(usr, "<span class='warning'>[key_name_admin(M)] cannot be kicked from the server.</span>")
-				return
-			to_chat(M, "<span class='warning'>You have been kicked from the server</span>")
-			log_admin("[key_name(usr)] booted [key_name(M)].")
-			message_admins("<span class='notice'>[key_name_admin(usr)] booted [key_name_admin(M)].</span>", 1)
-			//M.client = null
-			del(M.client)
+			switch(alert("Kick [M.ckey]?",,"Yes","No"))
```
farie82: No need for a switch here if only the yes has behaviour.

can replace the switch here with a simple if
```
if(alert(....) != "Yes")
    return
```

TheSardele: There, it uses if instead of switch now.

taukausanake:
> "My hands slips" yeah sure... OwO

_Oh plweeze daddy-admin kick me moar_

code/modules/admin/topic.dm
```diff
@@ -902,14 +902,15 @@
...
 		if(ismob(M))
-			if(M.client && M.client.holder && (M.client.holder.rights & R_BAN))
-				to_chat(usr, "<span class='warning'>[key_name_admin(M)] cannot be kicked from the server.</span>")
-				return
-			to_chat(M, "<span class='warning'>You have been kicked from the server</span>")
-			log_admin("[key_name(usr)] booted [key_name(M)].")
-			message_admins("<span class='notice'>[key_name_admin(usr)] booted [key_name_admin(M)].</span>", 1)
-			//M.client = null
-			del(M.client)
+			if(alert("Kick [M.ckey]?",,"Yes","No") == "Yes")
```
Crazylemon64: Note that this introduces a delay here - meaning that the mob could null out (gets deleted in some way) or mob's client could either null out (ghosting) or be swapped (mindswap) before the admin can click yes. It might be better to grab a reference to the client/ckey before starting the check.

TheSardele: Updated it, it now gets a reference to the client first, and makes sure it's not null (turns out things like already kicked players and aghosted admins can have ckeys, but no clients).

Fox-McCloud:
@Crazylemon64

Fox-McCloud:
please fix conflicts.

PR #13155, opened 2020-03-20
Paper Letterhead

Ty-Omaha:
Adds different standard letterheads for NT and Syndicate and an optional standard template for CC faxes.

CC:
image

Syndicate Letterhead: (NT one is the same except the NT logo)
image

- [x] Size images down and remove sizing html

- [x] Add special paper bins to map

- [x] See if I can do something about the 3 copied and pasted attack_hands

:cl:
add: Paper letterhead
/ :cl:


code/modules/admin/topic.dm
```diff
@@ -2194,9 +2194,15 @@
...
-
+		var/obj/item/paper/P
+		var/use_letterheard = alert("Use letterhead? If so, do not add your own header or a footer. Type and format only your actual message.",,"Nanotrasen","Syndicate", "No")
+		switch(use_letterheard)
+			if("Nanotrasen")
+				P = new /obj/item/paper/central_command(null)
+			if("Syndicate")
+				P = new /obj/item/paper/syndicate(null)
+			if("No")
+				P = new/obj/item/paper(null)
```
farie82: spacing.

code/modules/paperwork/paperbin.dm
```diff
@@ -10,7 +10,8 @@
 	pressure_resistance = 8
 	var/amount = 30					//How much paper is in the bin.
 	var/list/papers = list()	//List of papers put in the bin for reference.
-
+	var/obj/item/paper/letterhead
```
farie82: just `var/letterhead` here. Since it is a type and not the actual paper object.

letterhead_type or letterhead_path will for instance be better names for this so readers will instantly see that it is a path.

code/modules/paperwork/paperbin.dm
```diff
@@ -139,3 +143,15 @@
 
 	add_fingerprint(user)
 	return
+	
+
+/obj/item/paper_bin/nanotrasen
+	name = "nanotrasen paper bin"
+	icon_state = "paper_bin1"
```
farie82: icon_state is overridden here with the same value as it was before.

code/modules/paperwork/paperbin.dm
```diff
@@ -139,3 +143,15 @@
...
+	letterhead = /obj/item/paper/nanotrasen
+
+/obj/item/paper_bin/syndicate
+	name = "syndicate paper bin"
+	icon_state = "paper_bin1"
```
farie82: same as above

PR #13157, opened 2020-03-21
Lighter refactor + Reduces zippo lighter and gavel text and sound spam

SteelSlayer:
## What Does This PR Do
Basically what the changelog says. Adds a 5 second cooldown for the gavel and zippo lighter visible messages and sounds. Also refactors lighter code because it was gross.

## Why It's Good For The Game
Prevents massive walls of text spam and sound spam.

## Changelog
:cl:
tweak: Zippo lighter's visible message and on/off sound can now only be triggered once every 5 seconds.
tweak: Hitting a gavel block with a gavel now has a cooldown of 5 seconds.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/lighters.dm
+++ b/code/game/objects/items/weapons/lighters.dm
@@ -14,6 +14,7 @@
 	var/lit = FALSE
 	var/icon_on = "lighter-g-on"
 	var/icon_off = "lighter-g"
+	var/light_chance = 75
 
 /obj/item/lighter/random/New()
 	..()
@@ -39,7 +40,7 @@
 	hitsound = 'sound/items/welder.ogg'
 	attack_verb = list("burnt", "singed")
 
-	if(prob(75) || issilicon(user)) // Robots can never fail to light it.
+	if(prob(light_chance) || issilicon(user)) // Robots can never fail to light it.
 		to_chat(user, "<span class='notice'>You light [src].</span>")
 	else
 		var/mob/living/carbon/human/H = user
@@ -98,6 +99,7 @@
 	item_state = "zippo"
 	icon_on = "zippoon"
 	icon_off = "zippo"
+	light_chance = 100
 	var/next_on_message
 	var/next_off_message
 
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/lighters.dm
```diff
@@ -33,54 +23,46 @@
...
+		to_chat(user, "<span class='notice'>You light [src].</span>")
+	else
+		var/mob/living/carbon/human/H = user
+		var/obj/item/organ/external/affecting = H.get_organ("[user.hand ? "l" : "r" ]_hand")
+		if(affecting.receive_damage( 0, 5 ))		//INFERNO
+			H.UpdateDamageIcon()
+		to_chat(user,"<span class='notice'>You light [src], but you burn your hand in the process.</span>")
+
+	set_light(2)
+	START_PROCESSING(SSobj, src)
```
Crazylemon64: Note: The way this is written, zippo lighters will now have a 25% burn chance like regular lighters, so you might want to refactor this bit, perhaps splitting it into `attempt_light` or something

SteelSlayer: Oops yeah zippo's are not meant to have that chance.

SteelSlayer: Would giving lighters a `light_chance` var and then doning `if(prob(light_chance)..` be acceptable?

Crazylemon64: That ought to work, although you might also want a variable for the message to display to others, since the zippo has a unique one

SteelSlayer: Yeah I still display that in the `turn_on_lighter` override for the zippos.

Crazylemon64: I'm mostly concerned about it displaying multiple messages instead of just the one - that seems a bit more janky

SteelSlayer: Fair, it is janky. I'll do what you said earlier, throw it into a seperate proc, and make it so if the fancy message gets shown it doesn't display the other one

PR #13163, opened 2020-03-21
Add "Medical Intern" alternate job title

nok-ko:
## What Does This PR Do

A one-line tweak adding the "Medical Intern" alternate job title to the Medical Doctor job. This change will help players tell people who are new to the medical roles apart from experienced doctors and surgeons.

## Why It's Good For The Game

Newbie doctors can pick an alternate title that will quickly make it clear that they are new to the Medbay.

## Changelog
:cl: Nokko
add: Added "Medical Intern" alternate job title for newbie doctors.
/:cl:

### DIFF
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -52,7 +52,7 @@
 	selection_color = "#ffeef0"
 	access = list(ACCESS_MEDICAL, ACCESS_MORGUE, ACCESS_SURGERY, ACCESS_CHEMISTRY, ACCESS_VIROLOGY, ACCESS_GENETICS, ACCESS_MINERAL_STOREROOM)
 	minimal_access = list(ACCESS_MEDICAL, ACCESS_MORGUE, ACCESS_SURGERY, ACCESS_MAINT_TUNNELS)
-	alt_titles = list("Surgeon","Nurse")
+	alt_titles = list("Surgeon","Nurse","Medical Intern")
 	minimal_player_age = 3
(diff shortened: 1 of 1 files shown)


AffectedArc07:
I would make sure this is tested with the SQL backend locally. Job titles are a little finnicky in the database and may need more changes than just this.

nok-ko:
Fair enough, I'll give it an actual test.

Quantum-M:
Why not add it as it's own role with some less access?

That way it's more clear on who the newbies are for medbay to help/teach, keep them from messing too much with stuff they don't know much about, and to add more roles to the round that could help in cutting down the +20 civilians.

nok-ko:
> 
> 
> Why not add it as it's own role with some less access?
> 
> That way it's more clear on who the newbies are for medbay to help/teach, keep them from messing too much with stuff they don't know much about, and to add more roles to the round that could help in cutting down the +20 civilians.

Great idea, I'll look at doing that!

nok-ko:
> 
> 
> I'm skeptical of the benefits of this.
> New medical players won't realize that they can even change the job title for MD.
> Even if they do, and do it, it will likely just lead to other players avoiding letting the new player treat them, which is not a great experience.
That is something I didn't consider, actually. Hm.

nok-ko:
Leaving as closed for now, I will have to make a more sensical proposal. Thanks for the critique!

hunakos57:
And if we were to put the medical internet at the first place of the list, then the new players would still get that title and others can see that they are new.

PR #13164, opened 2020-03-22
Gives the Nuke Codes to the NTR

Ty-Omaha:
## What Does This PR Do
Gives the NTR a piece of paper with the nuke codes on it when they spawn

## Why It's Good For The Game
Allows for more player autonomy and more variable scenarios to create different stories.

- [ ] Make the line breaks on the paper actually work

:cl:
add: Nanotrasen Representatives now get the nuke codes when they spawn
/ :cl:

### DIFF
--- a/code/modules/paperwork/paper.dm
+++ b/code/modules/paperwork/paper.dm
@@ -439,6 +439,7 @@
 /*
  * Premade paper
  */
+ //Use New() if you want to add vars or procs to the info/header/footer etc
 /obj/item/paper/Court
 	name = "Judgement"
 	info = "For crimes against the station, the offender is sentenced to:<BR>\n<BR>\n"
@@ -501,9 +502,7 @@
 /obj/item/paper/fortune/update_icon()
 	..()
 	icon_state = initial(icon_state)
-/*
- * Premade paper
- */
+
 /obj/item/paper/Court
 	name = "Judgement"
 	info = "For crimes against the station, the offender is sentenced to:<BR>\n<BR>\n"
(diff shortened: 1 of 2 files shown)


code/modules/paperwork/paper.dm
```diff
@@ -600,6 +599,14 @@
 	header = "<p><img style='display: block; margin-left: auto; margin-right: auto;' src='ntlogo.png' width='220' height='135' /></p><hr />"
 	info =  ""
 	
+/obj/item/paper/nanotrasen/nuke_code
+	name = "CLASSIFIED: Directive 7-12"
+	
+//New because we cant use vars or procs on compile
+/obj/item/paper/nanotrasen/nuke_code/New()
+	info = "<b>FOR YOUR EYES ONLY.</b> <br><br> To be used only as a last resort where any other option is inviable/failed and failure to contain shall cause more harm to Nanotrasen assets across the galaxy. <br> The nuclear code is: [get_nuke_code()] <br> Remember it, and destroy this paper."
```
Ty-Omaha: Why are these line breaks not working?

Quantum-M:
On one hand, it makes it more easier to deal with the station once things go terribly wrong.

On the other, it makes command easily give up when a biohazard is taking over and just nuke stuff, considering that Central Commands (admins) need to know if it's warranted, and is probably more liable to abuse, especially by antags, or bad players.

Why not also just make it a memory note?

MRgaty:
I dont think this is a good idea

Mitchs98:
Can't really see anything positive from this. Sure, it'll allow players on some god awful rounds to nuke when it's really needed and admins are too busy/afk/don't want to nuke the station. But on the other hand as Quantum said this will just err Command more into the realm of setting off the nuke as soon as things get a bit too bad, even if it's not past a tipping point yet. They already do it often and already often enough try to request the nuke codes when they aren't warranted.

Sure, you'd get banned/job banned if you nuked for no reason. But the fact of the matter is admins are far better a judge of when the nuke is actually warranted, especially given I've seen Command try to get a nuke/Gamma NUMEROUS times over one final red terror spider running around or something of the like.

Ty-Omaha:
>If anyone should get the codes, it is the Captain, not the NT Rep. The Captain can defend themselves, and has a lot more defense in general, than the NTR.

The reason I gave it to the NTR was because the role is karma locked, and it would require for more than just 1 person to decide that the station needs to be nuked

Kyep:
No, it wouldn't. 
What would happen in practice is:

- Captains would start routinely asking NTRs what the code is at shift start "just in case" the NTR goes to cryo, gets killed, etc.
- Traitors would simply kill both Captain and NTR, acquiring both the paper and the disk. It is way easier to kill those two people for both parts, than it is to kill everyone trying to board the shuttle.
- Most shifts don't have a NT Rep, which would make nuking protocol much more complicated, because half the time crew would have to ask CC for the codes, and the other half they've had someone on station with the codes.

datlo:
If NT Rep had the nuke codes, I would rather have them have it in memory notes than as a physical item, for the reasons stated above.

AdamElTablawy:
This just leads to constant nuking of the station any time there's a major threat.

Spacemanspark:
Nuking of the station should only be authorized by CC.
On a mechanics end, this is going to lead to griefing, trigger happy nukes, and rounds cut short. 

Also, It's Nanotrasen property first and foremost. They and their crazy protocols aren't going to allow the station and its asylum-needing residents to detonate it at will.

PR #13170, opened 2020-03-22
Adds Enforcer crate to Cargo supply

Ty-Omaha:
## What Does This PR Do
Allows for Enforcers and their magazines to be ordered at Cargo

## Why It's Good For The Game
It would allow security (and antags) a wider variety of weaponry to deal with threats.

:cl:
add: Enforcers can be ordered from cargo
/ :cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -459,7 +459,24 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 					/obj/item/ammo_box/magazine/wt550m9,)
 	cost = 30
 	containername = "auto rifle ammo crate"
+	
+/datum/supply_packs/security/armory/enforcer
+	name = "Enforcer Pistol Crate"
+	contains = list(/obj/item/gun/projectile/automatic/pistol/enforcer/lethal,
+					/obj/item/gun/projectile/automatic/pistol/enforcer/lethal,
+					/obj/item/gun/projectile/automatic/pistol/enforcer/lethal)
+	cost = 25
 
+/datum/supply_packs/security/armory/enforcerammo
+	name = "Enforcer Ammo Crate"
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Crew really don't need more ballistic weapons--the designation between syndicate having them but not NT is both a thematic and balance reason.

Energy has the advantage of being easy to recharge (typically) and unlimited amount of times, but cumbersome and time consuming to do so--in order to "stand and deliver" you have to sacrifice a lot of space and resources.

Ballistics are the opposite. They're limited, but extremely easy to "stand and deliver", not to mention incredibly space efficient. Your magazine runs out and you slap another one in and you're right back at it.

Generally keeping the existing balance of security having to have retreats and advances is better for the overall game balance than them being able to "stand and deliver" constantly.

This derives from the nature of antags vs the station. Typically antags are always vastly outnumbered, so they need some mechanic that allows them to have a break and not be relentless pursued.

Sec rifles push the limits of this as is--but they're intentionally hamstrung by low damage and low firing rate as a downside---something these don't have.

Shotguns have the limitation of effectively being short range melee weapons with how they work---and even so, shotguns have *repeatedly* wound up as problematic in the past.

Crew having ballistics is just not a great thing, at the end of the day.

Landerlow:
The thing is though, crew already has access to ballistic weapons through cargo.

Spacemanspark:
Honestly we should just replace the 'auto rifles' with enforcers.

Let's be real, the 'auto rifles' are just pistols with a front grip, stock, and top loaded cartridge. There's very little 'automatic' anything to them.

Kyep:
I'm actually sympathetic to @Spacemanspark 's idea. What, exactly, is "auto" about the "auto" rifles? If enforcers are seen as more interesting, and balance-wise, are not stronger (is this the case?) then why not have enforcers instead of the auto rifles?

MRgaty:
Revolvers are badass

variableundefined:
I object to giving security more ballistics. 

Closing due to two maintainer and one head objection.

PR #13162, opened 2020-03-21
Automatic merge conflict labeling

AffectedArc07:
## What Does This PR Do
This PR adds in a GitHub action which will automatically run when a PR is merged. This action checks the mergability of a PR, and if it is conflicted, will add the appropriate label.

## Why It's Good For The ~~Game~~ **CODEBASE**
Allows maints to easily see at a glance whice PRs are conflicted and which ones arent

## Changelog
No CL because this is github only

PR #13173, opened 2020-03-22
Revert "Automatic merge conflict labeling"

AffectedArc07:
Reverts ParadiseSS13/Paradise#13162

It false-negatives all commits to master, and also doesnt auto remove the label, since its on commit not on PR update. I will look into better solutions (such as TGs webhook) in the future.

image


AffectedArc07:
Its not as bad as I made it out to be

Issue #12607, opened 2019-10-18
Floor cluwne water not drying


Couls:
**Issue Description**:

Floor cluwne water and lube did not dry at all the whole shift

**What did you expect to happen**: 

for tiles to not be permanently wet or lubed

**What happened instead**:
they were permanently wet and lubed

**Why is this bad/What are the consequences**:
Really inconvenient and makes navigation nearly impossible

**Steps to reproduce the problem**:
Summon floor cluwne, have cluwne kill a few people and spread water

**When did the problem start happening**:
Started on main server the other shift

**Extra information**:
This only happened on the live server and I haven't attempted to reproduce it locally yet

PR #13178, opened 2020-03-22
Makes floor cluwnes slippery tiles not stay there till the end of times

farie82:
## What Does This PR Do
Fixes: #12607

## Why It's Good For The Game
It's a bug

## Changelog
:cl:
fix: Makes floor cluwnes slippery tiles dissapear after a certain time now. As intended
/:cl:

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -64,11 +64,9 @@
 		return
 	if(time == null)
 		time =	rand(790, 820)
-	addtimer(CALLBACK(src, /turf/simulated./proc/MakeDry, wet_setting), time)
+	addtimer(CALLBACK(src, .proc/MakeDry, wet_setting), time)
 
-/turf/simulated/proc/MakeDry(wet_setting = TURF_WET_WATER)
-	if(!istype(src, /turf/simulated)) //Because turfs don't get deleted, they change, adapt, transform, evolve and deform. they are one and they are all.
-		return
+/turf/simulated/MakeDry(wet_setting = TURF_WET_WATER)
 	if(wet > wet_setting)
 		return
 	wet = TURF_DRY
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -62,7 +62,7 @@
 		overlays += wet_overlay
 	if(time == INFINITY)
 		return
-	if(time == null)
+	if(!time)
 		time =	rand(790, 820)
 	addtimer(CALLBACK(src, .proc/MakeDry, wet_setting), time)
 
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -35,7 +35,12 @@
 		assume_air(lowertemp)
 		qdel(hotspot)
 
-/turf/simulated/proc/MakeSlippery(wet_setting = TURF_WET_WATER, infinite = FALSE) // 1 = Water, 2 = Lube, 3 = Ice, 4 = Permafrost
+/*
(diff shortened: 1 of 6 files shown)


code/game/turfs/simulated.dm
```diff
@@ -55,13 +60,15 @@
...
-			MakeDry(wet_setting)
+	if(time == INFINITY)
+		return
+	if(time == null)
+		time =	rand(790, 820)
+	addtimer(CALLBACK(src, /turf/simulated./proc/MakeDry, wet_setting), time)
 
 /turf/simulated/proc/MakeDry(wet_setting = TURF_WET_WATER)
+	if(!istype(src, /turf/simulated)) //Because turfs don't get deleted, they change, adapt, transform, evolve and deform. they are one and they are all.
+		return
```
SteelSlayer: I know these lines were already here, but do you know why this is needed? If the callback for this proc was called on a `turf/simulated`, why do we need to check that `src` is not that type anymore? And the comment doesn't really help :smile:

code/game/turfs/simulated.dm
```diff
@@ -55,13 +60,15 @@
...
-	if(!infinite)
-		spawn(rand(790, 820)) // Purely so for visual effect
-			if(!istype(src, /turf/simulated)) //Because turfs don't get deleted, they change, adapt, transform, evolve and deform. they are one and they are all.
-				return
-			MakeDry(wet_setting)
+	if(time == INFINITY)
+		return
+	if(time == null)
+		time =	rand(790, 820)
+	addtimer(CALLBACK(src, /turf/simulated./proc/MakeDry, wet_setting), time)
```
SteelSlayer: I am like 99% sure you can just do `.proc/MakeDry` here. I did a quick test of it just to be sure, poured some water on a tile, and it dried just fine.

code/game/turfs/simulated.dm
```diff
@@ -55,13 +60,15 @@
...
 				wet_overlay = image('icons/effects/water.dmi', src, "wet_static")
 		overlays += wet_overlay
-	if(!infinite)
-		spawn(rand(790, 820)) // Purely so for visual effect
-			if(!istype(src, /turf/simulated)) //Because turfs don't get deleted, they change, adapt, transform, evolve and deform. they are one and they are all.
-				return
-			MakeDry(wet_setting)
+	if(time == INFINITY)
+		return
+	if(time == null)
```
SteelSlayer: Couldn't you just do `if(!timing)` here?

farie82: I did this one to be compatable with floors being wet for 1 frame. Not sure why it would be used but it can now! :P

code/game/turfs/simulated.dm
farie82: Will change once able!

code/game/turfs/simulated.dm
farie82: Yeah no clue why it's there. It probably does nothing currently. Will test it without and see the results if I go change a turf

farie82: It did nothing yet caused a runtime if it happened. The turf was changed and it tries to call the proc.
But if it's say space it will still runtime. Fixed in the last commit

code/game/turfs/simulated.dm
Fox-McCloud: I'd prefer `!time` personally--`== null` checks are kinda gross  unless it's absolutely required.

farie82: Changed it in the last commit

PR #13181, opened 2020-03-23
fixes smoking and similar incremental dosage systems for addiction.

Aquilar:
## What Does This PR Do

moves the sating of addiction to reagent instead of having it done in reagents/metabolise, the problem was addiction was not being sated by smoking and similar methods of incremental dosage as the reagent was being removed before metabolise had a chance to be called.

## Why It's Good For The Game

smoking sates addiction now.

## Changelog
:cl:
fix: incremental dosages like smoking now properly sate addictions.
/:cl:


code/modules/reagents/chemistry/holder.dm
```diff
@@ -299,21 +299,17 @@
 			if(R.addiction_stage < 5)
 				if(prob(5))
 					R.addiction_stage++
-			if(M.reagents.has_reagent(R.id))
```
farie82: Still need to do this check. But the the negative version. Currently you will display stage 1 addiction stuff even if you have the chem inside of you

Aquilar: stage one is the default stage of addiction.

farie82: Yet is the thing. Somebody can add a new withdrawal type that does something at stage 1.
It's just incorrect in my perspective that you can get withdrawal signs while you're still under influence of the reagent.

Aquilar: this is for another PR, this just a straight fix to get smoking and similar methods to sate addiction.

farie82: You do remove it in this PR. So it is part of the PR. Before it was like this:
Reagent is in body. You don't get symptoms.
Now you can have a reagent in your body and get level 1 symptoms

Fox-McCloud: Yeah, this functionality needs to remain.

nok-ko:
Please merge this. Please.

Fox-McCloud:
Please address the comments that @farie82 and myself stated.

Aquilar:
they've been addressed already.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -299,10 +300,7 @@
...
-				R.last_addiction_dose = world.timeofday
-				R.addiction_stage = 1
-			else
+			if(world.timeofday > (R.last_addiction_dose + ADDICTION_WITHDRAWAL_DELAY)) //time check so addiction act doesn't play over and over. Allows incremental dosages to work.
```
Fox-McCloud, line 303: I'm pretty sure that `if(world.timeofday > R.last_addiction_dose)` Would be enough.

You can throw some debug `to_chat(world, "blehhh")` in to test this with smoking cigs or something similar just to be sure that what's below this `if` statement isn't running---you can even see how much time has past by throwing in a `to_chat(world, "[world.timeofday]")` on `reagent/on_mob_life` and right about here to see how much time has passed; I'd venture to guess the game won't even register a time change, as this only has granularity on the order of 1/10th of a second, and reagent processing is not really knowing to induce lag (unless something horrendous went wrong).

Of the edge cases of there being lag, this is so inconsequential that `addiction_act_stage1` running once in a blue moon, only to return to proper behavior the next time this runs....I'm not worried.

Aquilar: yeah that works

PR #13182, opened 2020-03-23
Colours MC subsystems based on state

AffectedArc07:
## What Does This PR Do
This PR colours the statpanel for MC subsystems based on their current running state
image

*Making a 200x200 explosion to simulate mass lag*
wokC1xWf0C

Subsystems which are actively processing fine, will colour green
Subsystems which are in the queue to be processed, but are delayed, will colour orange
Subsystems which have their processing paused due to lag, will colour red
Subsystems which are currently sleeping will be coloured blue

## Why It's Good For The Game
This allows admins who arent as experienced with the MC panel layout to get a rough idea of what is killing the server, so it can be relayed over. It is a lot easier to say "Oh yeah the lighting SS was red" than "The lighting SS had a letter next to it, I couldnt quite remember which one though"

## Changelog
:cl: AffectedArc07
add: Subsystems will now colour themselves based on their status
/:cl:

### DIFF
--- a/code/controllers/subsystem.dm
+++ b/code/controllers/subsystem.dm
@@ -181,7 +181,7 @@
 	if(can_fire)
 		title = "\[[state_letter()]][title]"
 
-	stat(title, statclick.update(msg))
+	stat(title, "[state_colour()][statclick.update(msg)]</font>")
 
 /datum/controller/subsystem/proc/state_letter()
 	switch(state)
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem.dm
```diff
@@ -196,6 +196,18 @@
 		if(SS_IDLE)
 			. = "  "
 
+/datum/controller/subsystem/proc/state_colour()
```
farie82: minor thing. But this proc does not return the color. It returns the start of the font that has the color in it.

PR #13184, opened 2020-03-23
Makes it so that IRC can't be cloned

farie82:
## What Does This PR Do
Makes it so that IRC and such can't be cloned.
The made IRC had a holder brain that had a default species set. And the species was human....

## Why It's Good For The Game
Fixes a bug

## Changelog
:cl:
fix: IRCs can't be cloned
/:cl:

### DIFF
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -359,12 +359,12 @@
 	if(scan_brain && !can_brainscan())
 		return
 	if(isnull(subject) || (!(ishuman(subject))) || (!subject.dna))
-		if(isalien(subject)) 
+		if(isalien(subject))
 			scantemp = "<span class=\"bad\">Error: Xenomorphs are not scannable.</span>"
 			SSnanoui.update_uis(src)
 			return
 		// can add more conditions for specific non-human messages here
-		else 
+		else
 			scantemp = "<span class=\"bad\">Error: Subject species is not scannable.</span>"
 			SSnanoui.update_uis(src)
 			return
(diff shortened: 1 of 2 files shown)


code/game/machinery/computer/cloning.dm
```diff
@@ -359,20 +359,20 @@
...
+		else
 			scantemp = "<span class=\"bad\">Error: Subject species is not scannable.</span>"
 			SSnanoui.update_uis(src)
 			return
 	if(subject.get_int_organ(/obj/item/organ/internal/brain))
 		var/obj/item/organ/internal/brain/Brn = subject.get_int_organ(/obj/item/organ/internal/brain)
 		if(istype(Brn))
 			if(NO_SCAN in Brn.dna.species.species_traits)
-				scantemp = "<span class=\"bad\">Error: [subject.dna.species.name_plural] are not scannable.</span>"
+				scantemp = "<span class=\"bad\">Error: [Brn.dna.species.name_plural] are not scannable.</span>"
```
farie82: It would show wrong information. Put a robo brain in a human and it would say "Humans are not scannable"

Mitchs98:
This is beautiful.

Ty-Omaha:
nice

Spacemanspark:
The method we found this out with was hilarious.

"Oh great they're popping me into cloning, wonderful."

_Two minutes of grumbling later_

"Man I'm never going to be-"

_Spawns in again, cloned_

"WHAT"

code/game/machinery/computer/cloning.dm
Crazylemon64: It sounds like the proper fix to that would be to set the species on the brain's DNA appropriately

Crazylemon64: though it looks like you've done that, too

SynthNumerics:
i never even realised this was a thing, but fuck dude the code is beautiful

Fox-McCloud:
Make sure this doesn't break MMI cloning.

You can robotize someone's chest, then yank out their brain, put it in an MMI, then put the MMI in the chest and revive them--they'll walk around and  be alive.

It's a bit convoluted, but people in such a state should still be able to be cloned and the DNA of the MMI should match that of the brain.

farie82:
> Make sure this doesn't break MMI cloning.
> 
> You can robotize someone's chest, then yank out their brain, put it in an MMI, then put the MMI in the chest and revive them--they'll walk around and be alive.
> 
> It's a bit convoluted, but people in such a state should still be able to be cloned and the DNA of the MMI should match that of the brain.

I've not touched MMI code itself. Only the New of robo brains. I will still test it though just in case

Fox-McCloud:
I just wanted to be sure, because MMI's as brain are really hacky and robotic brains are intertwined with them, so.

Thanks for confirming.

PR #13185, opened 2020-03-23
Paintable Lights

Ty-Omaha:
## What Does This PR Do
Allows you to paint light bulbs with a spraycan. To clean it, use soap on the light bulb. 

The bulb must be removed from the fixture first.

## Why It's Good For The Game
You can add cool mood lighting to areas. If you paint it black it will be as if it is off. 

## Images of changes
image

image

:cl:
add: Paintable lights. You can now paint lights by removing the bulb and using a spraycan on it. To clean, remove and use soap.
/ :cl:

### DIFF
--- a/code/game/objects/items/weapons/soap.dm
+++ b/code/game/objects/items/weapons/soap.dm
@@ -45,6 +45,14 @@
 		if(do_after(user, cleanspeed, target = target))
 			to_chat(user, "<span class='notice'>You clean \the [target.name].</span>")
 			clean_turf(target)
+	else if(istype(target, /obj/item/light))
+		user.visible_message("<span class='warning'>[user] begins to clean \the [target.name] with [src].</span>")
+		if(do_after(user, cleanspeed, target = target) && target)	
+			to_chat(user, "<span class='notice'>You clean \the [target.name].</span>")		
+			var/obj/item/light/L = target
+			L.brightness_color = null
+			L.color = null
+			L.update()
 	else
 		user.visible_message("<span class='warning'>[user] begins to clean \the [target.name] with [src].</span>")
 		if(do_after(user, cleanspeed, target = target))
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -618,7 +618,6 @@
 	var/brightness_power = 1
 	var/brightness_color
 	var/time_to_paint = 30  //amount of time it takes to paint a light
-	var/time_to_clean = 40 //amount of time it takes to clean a light
 
 /obj/item/light/ComponentInitialize()
 	. = ..()
@@ -631,16 +630,10 @@
 		var/obj/item/toy/crayon/spraycan/S = O
 		brightness_color =  S.colour
 		color = S.colour
-		visible_message("<span class='warning'>[M] paints the [src]!</span>")
+		visible_message("<span class='notice'>[M] paints the [src]!</span>")
 		update()
 		M.do_attack_animation(src)
-	//clean the light
-	else if(istype(O, /obj/item/soap) && do_after(M, time_to_clean, target = src))
-		brightness_color = null
-		color = null
-		visible_message("<span class='warning'>[M] cleans the [src]!</span>")
-		update()
-		M.do_attack_animation(src)
+	//cleaning is handled in soap.dm
 
 /obj/item/light/Crossed(mob/living/L)
 	if(istype(L) && has_gravity(loc))
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -155,7 +155,7 @@
 	var/static_power_used = 0
 	var/brightness_range = 8	// luminosity when on, also used in power calculation
 	var/brightness_power = 1
-	var/brightness_color = "#FFFFFF"
+	var/brightness_color
 	var/status = LIGHT_OK		// LIGHT_OK, _EMPTY, _BURNED or _BROKEN
 	var/flickering = 0
 	var/light_type = /obj/item/light/tube		// the type of light item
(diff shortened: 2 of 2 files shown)


Carthusia:
This would be super nice to have, i've never been a huge fan of the floor lights, just to big, clunky and people can just walk up and turn them off or recolor them.

I can foresee many cool moodlights from this

Quantum-M:
Maybe make the black one give a more dark grey color instead of making it not give off color? Otherwise, you'll end up having that one person spraying all the lights "off", and given that soap is a bit rare around the station, it would be hard to clean/deal with. 
Especially since you can create a lot of spray cans from the autolathe, _I think_.

Ty-Omaha:
People have the ability go around and smash all the lights if they want to. Spraying all the light tubes is equivalent to that; with the exception that new lights don't _have_ to be put in. It would take longer to spray all the lights black because you have to remove the light tube, paint, and then place it back in.

code/modules/power/lighting.dm
```diff
@@ -614,12 +616,32 @@
...
+	//color the light with a spraycan
+	if(istype(O, /obj/item/toy/crayon/spraycan) && do_after(M, time_to_paint, target = src))
+		var/obj/item/toy/crayon/spraycan/S = O
+		brightness_color =  S.colour
+		color = S.colour
+		visible_message("<span class='warning'>[M] paints the [src]!</span>")
+		update()
+		M.do_attack_animation(src)
+	//clean the light
+	else if(istype(O, /obj/item/soap) && do_after(M, time_to_clean, target = src))
```
farie82: Forget to include toolspeed. Different soaps have different speeds at which they clean

code/modules/power/lighting.dm
```diff
@@ -614,12 +616,32 @@
...
+		brightness_color = null
```
farie82: This won't break? Since it used to always have a value before

code/modules/power/lighting.dm
```diff
@@ -257,7 +257,7 @@
 	if(on)
 		var/BR = nightshift_enabled ? nightshift_light_range : brightness_range
 		var/PO = nightshift_enabled ? nightshift_light_power : brightness_power
-		var/CO = nightshift_enabled ? nightshift_light_color : brightness_color
+		var/CO = nightshift_enabled && (brightness_color == "#a0a080" || !brightness_color) ? nightshift_light_color : brightness_color
```
farie82: Design point:
Can't you mix/blend the brightness_color with the painted color? This way nightshifts will still be darker

S-CR:
Having black lights that have luminosity without visually representing that will majorly impact shadowlings.

Denthamos:
The shadowling part worries me as well if its black but still giving 'light' to tiles in its radius that could hurt them.  Just making it not give off light however when painted black would be a large buff to them as well and not something which is a good solution.

Ty-Omaha:
>Having black lights that have luminosity without visually representing that will majorly impact shadowlings. Consider allowing autolathes to print colored bulbs or giving cargo a colored lightbulb pack as an alternative.

That is a very good point I will see what I can do.

>Just making it not give off light however when painted black would be a large buff to them as well and not something which is a good solution.

To my knowledge you can't repair smashed lights. This is directly equivalent to that. There would be no buff. It takes longer to paint a light than it does to smash it.

Carthusia:
Heh, just a funny thought, don't know if its possible or even within the scope of this PR

What if when the station goes epsilon, all the lights turn red? One final "Your fucked" thing to set the mood for murder and panic

code/modules/power/lighting.dm
Ty-Omaha: Good point, how some have a set color like the small bulbs. I will see what I can do

Ty-Omaha:
>Having black lights that have luminosity without visually representing that will majorly impact shadowlings. Consider allowing autolathes to print colored bulbs or giving cargo a colored lightbulb pack as an alternative.

So after making a fix for this after and hour, it turns out it's not even an issue because there is no luminosity if the color is dark enough.

code/modules/power/lighting.dm
Ty-Omaha: The paint takes the place in brightness_color. The painted colors are still darker in nightshift mode because brightness_power. All it would really do is change the hue

image

image

code/modules/power/lighting.dm
Ty-Omaha: I think it makes more sense to just remove any color something has after cleaning it since I'd have to check for a certain color and it's pretty much the same anyways - light peach and white.

Nothing breaks except it makes it white

code/modules/power/lighting.dm
```diff
@@ -614,12 +616,25 @@
...
 	AddComponent(/datum/component/caltrop, force)
 
+/obj/item/light/attackby(obj/item/O, mob/M)
+	..()
+	//color the light with a spraycan
+	if(istype(O, /obj/item/toy/crayon/spraycan) && do_after(M, time_to_paint, target = src))
+		var/obj/item/toy/crayon/spraycan/S = O
+		brightness_color =  S.colour
+		color = S.colour
+		visible_message("<span class='notice'>[M] paints the [src]!</span>")
```
farie82, line 633: `the [src]` will turn out as `the the src` probably. You can leave the `the` out of it I think

farie82: This is also missing a self message. Can be intended though

code/game/objects/items/weapons/soap.dm
```diff
@@ -45,6 +45,14 @@
 		if(do_after(user, cleanspeed, target = target))
 			to_chat(user, "<span class='notice'>You clean \the [target.name].</span>")
 			clean_turf(target)
+	else if(istype(target, /obj/item/light))
+		user.visible_message("<span class='warning'>[user] begins to clean \the [target.name] with [src].</span>")
```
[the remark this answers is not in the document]
farie82: Missing a self message as well. Might be intended though

AffectedArc07:
image
Travis isnt even present. I recommend you atleast fix those things to make travis actually appear.

variableundefined:
@Ty-Omaha I'd like to review/merge this PR - Can you resolve the merge conflicts within a week? 

Otherwise, I'll close it and you can reopen it at a latter date

PR #13187, opened 2020-03-24
Increases distance of accelerated particles back to how they were

### LINKED ISSUE #13183: Issue related to singularity upkeep
**Issue Description**:
Singularity engine reverts back to stage 1 with it setup perfectly. The engine setup followed the wiki guide https://www.paradisestation.org/wiki/index.php/Singularity_Engine.

I've been playing engineer for about a year and the singularity used to run like this. You would setup the engine like the guide describes. You allowed the singularity to grow to stage 3 after which you would set the PA to 0. The singularity would then fluctuate between stage 2 and 3 the entire shift and no further action was required.

**What happened instead**:
With the PA set to 0 the singularity no longer fluctuated between stage 2 and 3. Instead it shrank to stage 1 which cause power issues and is a hassle to deal with. Either you ask CC to remove the baby singulo or you move the PA to hit the singularity.

While I think I have some creative in-game solutions to this problem that I would love to try. The way described in the wiki guide for singularity setup would either cause power failures or have one engineer play to fun game of watching a big ball shrink and turning a dial for about 2 hours. The moment it shrinks is quite unpredictable to so someone would have to watch it constantly to prevent issues. In my experience engineers that really involve themselves in the game are very creative people who can make neat things or are fun to RP with. Having the singularity work this way would essentially keep them out of them game thanks to boring busy work.

1. setup the engine following this guide https://www.paradisestation.org/wiki/index.php/Singularity_Engine
2. Wait
3. Watch the big ball become a small ball and experience the joy of the singularity engine.

I first noticed it about 2 weeks ago but the singularity gets used so rarely that it might have been in game for longer.

**Extra information**:
Personally I believe that the singularity should have it's own quirks but the result this of this quirk is just a chore. The solution must not necessarily be the reversal of whatever changes that caused this. A better way to monitor the singularity would be a great solution. Like the way engineering gets informed if a supermatter shard is about to go critical, an announcement on engineering coms about the stage changes of the singularity would go a long way.

I do not know if it is related but I am damn certain that the engine starters all take longer to pop out an engine then before. I've never actually timed how long it takes but with the amount of times I've set up those engines I am sure it takes longer now. Not sure when this change happened.

Citinited:
Fixes #13183 - this was due to an oversight in the accelerated particle code making them delete themselves one turf earlier than how they used to.

Also removes unused speed var which was not used anywhere (I removed the code that cared about it in my earlier PR but nothing actually made use of that)

:cl:
fix: Fixes an oversight that made accelerated particles travel 1 less turf than they used to. The singularity will now act like it used to.
/:cl:

### DIFF
--- a/code/modules/power/singularity/particle_accelerator/particle.dm
+++ b/code/modules/power/singularity/particle_accelerator/particle.dm
@@ -5,20 +5,19 @@
 	icon_state = "particle"
 	anchored = TRUE
 	density = FALSE
-	var/movement_range = 10
+	var/movement_range = 11
 	var/energy = 10
-	var/speed = 1
 
 /obj/effect/accelerated_particle/weak
-	movement_range = 8
+	movement_range = 9
 	energy = 5
 
 /obj/effect/accelerated_particle/strong
-	movement_range = 15
+	movement_range = 16
 	energy = 15
 
 /obj/effect/accelerated_particle/powerful
-	movement_range = 20
+	movement_range = 21
 	energy = 50
 
 


AllStickyGithub:
Aw thanks man.

BakKeener:
Finally we can have a few more singulooses because of that 1 tile difference, no joke I seen many singularities revert to stage 1 and can't progress because the PA is just 1 tile away from reaching them.

PR #13190, opened 2020-03-24
Adds a helper proc to calculate luminance

Ty-Omaha:
Adds a proc to calculate the perceived brightness of a color
getLuminance

:cl:
add:Adds getLuminance() helper proc
/ :cl:


FlattestGuitar:
Changelogs are for players, not coders.

FlattestGuitar:
Also why add a proc you don't need when you could add it when you need it?

Ty-Omaha:
>Changelogs are for players, not coders.

afaik that's outdated

>Also why add a proc you don't need when you could add it when you need it?

so it's there for other people to use if they need it for something

Crazylemon64:
it's kinda weird to just add a helper like that on its own with nothing that uses it

but I figure it could come in handy for sdql hijinks

PR #11137, opened 2019-03-26
Can't break out of grabs while restrained now

farie82:
**What does this PR do:**
Does what the title says.
I found it odd that you can't move out of being dragged but you can break out of grabs. Especially annoying when you as a cling want to absorb somebody alive. Or say want to table somebody who is restrained so you can remove his head.

**Changelog:**
:cl:
balance: Can't break out of grabs while being restrained
/:cl:

PR #13193, opened 2020-03-25
Can't break out of grabs while incapacitated

farie82:
## What Does This PR Do
Makes you unable to break out of grabs while you're restrained or stunned or otherwise incapacitated (ignoring resting). 

Redo of #11137 
That PR was closed for unforeseen side effects. Yet I can't think of any that are not intended. This code only touches the Process_Grab code.
It is a balance change because people can now restrain people with grabs after they stun prod them. But making restraints is more efficient and won't slow them down once cuffed.
And you can still break out of grabs (except the kill grab) once you're no longer stunned. Just need to hope luck is on your side.

## Why It's Good For The Game
Currently cuffed people can be dragged but not grabbed. Same goes to stunned people.
This makes grabbing near useless and simply unintuitive to use.

## Changelog
:cl:
tweak: Being incapacitated now makes you unable to break out of grabs
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -215,6 +215,9 @@
 ///Called by client/Move()
 ///Checks to see if you are being grabbed and if so attemps to break it
 /client/proc/Process_Grab()
+	if(mob.incapacitated()) // Can't break out of grabs if you're incapacitated
+		return TRUE
+
 	if(mob.grabbed_by.len)
 		var/list/grabbing = list()
 
@@ -237,17 +240,17 @@
 				if(GRAB_AGGRESSIVE)
 					move_delay = world.time + 10
 					if(!prob(25))
-						return 1
+						return TRUE
 					mob.visible_message("<span class='danger'>[mob] has broken free of [G.assailant]'s grip!</span>")
 					qdel(G)
 
 				if(GRAB_NECK)
 					move_delay = world.time + 10
 					if(!prob(5))
-						return 1
+						return TRUE
 					mob.visible_message("<span class='danger'>[mob] has broken free of [G.assailant]'s headlock!</span>")
 					qdel(G)
-	return 0
+	return FALSE
 
 
 ///Process_Incorpmove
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob_movement.dm
```diff
@@ -215,6 +215,9 @@
 ///Called by client/Move()
 ///Checks to see if you are being grabbed and if so attemps to break it
 /client/proc/Process_Grab()
+	if(mob.incapacitated()) // Can't break out of grabs if you're incapacitated
```
farie82: The incapacitated check also takes a parameter to ignore grabs. But this does nothing

MRgaty:
I can understand how someone cuffed could resist a grab... Somehow... But being able to resist a grab while stunned is ridiculous

farie82:
> I can understand how someone cuffed could resist a grab... Somehow... But being able to resist a grab while stunned is ridiculous

Dragging currently also works like that. If you're incapacitated you can't break free. So I made them behave the same in that regard.
Those butt cheeks of clamping might be of pure steel but spessmans hands can hit a monkey with a toolbox for over an hour without pauses. Now if that can't keep a stunned or cuffed person in check then I don't know what will

BakKeener:
Does this happen to fix the annoying issue with trying to grab someone who is stunned but it won't let you properly grab them and keeps dropping the grab?

farie82:
> Does this happen to fix the annoying issue with trying to grab someone who is stunned but it won't let you properly grab them and keeps dropping the grab?

Yep it fixes exactly that. Same for weakening and cuffs.

code/modules/mob/mob_movement.dm
```diff
@@ -215,6 +215,9 @@
 ///Called by client/Move()
 ///Checks to see if you are being grabbed and if so attemps to break it
 /client/proc/Process_Grab()
+	if(mob.incapacitated(FALSE, TRUE, TRUE)) // Can't break out of grabs if you're incapacitated
+		return TRUE
```
Fox-McCloud: This is going to break the ability of the code on lines 247 from running.

farie82: The getting out of grabs? Aye that's the idea. If you mean the `move_delay` stuff. I don't think it's needed to set the `move_delay` if you don't actually move since you're stunned on the floor or cuffed etc.

variableundefined:
Travis Borked. Reopening.

variableundefined:
Has been in the queue for way too long. Approving and merging.

PR #13198, opened 2020-03-25
[DNM]Makes the sling glare ability targeted by mouse click using a circle around the click

farie82:
## What Does This PR Do
Removes the god awful selection menu out of the combat focused shadowling glare ability.
Replaces it with a mouse click overload which looks for a target in a 1 radius circle around the click position.
So (O is center):
XXX
XOX
XXX

Also fixes a bug where the button won't show it is recharging if the target is in range due to the sleep used.

DNM due to me having found a bug in this code.
I'm currently making another PR making this mechanic be used for more spells so I'll include this one in the other.

## Why It's Good For The Game
Having a selection menu that pulls focus on a combat focused ability is a horrible design.
This will remove frustration from using the ability.

## Images of changes
GaeOsIRacM

## Changelog
:cl:
tweak: Replaces the shadowlings glare target list with a mouse click override looking for a target on the location clicked in a 1 radius circle
/:cl:

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -18,80 +18,91 @@
 	return 0
 
 
-/obj/effect/proc_holder/spell/targeted/glare //Stuns and mutes a human target, depending on the distance relative to the shadowling
+/obj/effect/proc_holder/spell/glare //Stuns and mutes a human target, depending on the distance relative to the shadowling
 	name = "Glare"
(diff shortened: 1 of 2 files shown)


BakKeener:
I'm honestly in full support of this simply because it removes that dreadful popup menu which no-one likes to be honest, the menu is annoying to use when there's a group of people and you need to know which one is in front and actually get to their name in time before they pull something on you. With this new system it's as easy as click the ability, then c lick your target. Although I'm not sure if we should have it work just by clicking near someone instead of just on them but it doesn't seem to be a big deal considering the size of the targeting zone.

farie82:
#13220 does the same but better

PR #8616, opened 2018-01-28
Reduce shuttle cheese

Tayyyyyyy:
Once you up to red, a timer starts to tick down from 5 minutes. After 5 minutes passes, the shuttle time is reduced to 5 minutes.

For reasons why this is a good idea, see @Spartan6's [forum suggestion](https://nanotrasen.se/forum/topic/11900-nov-25th-meeting-questions/?tab=comments#comment-99408).

Shouldn't affect automatic crew transfer.

:cl: Tayyyyyyy, LP Spartan
tweak: Once you up to red, it takes 5 minutes before the shuttle transit time is reduced to 5 minutes.
/:cl:

PR #13199, opened 2020-03-25
5 Minute Shuttle

Ty-Omaha:
## What Does This PR Do
Makes the shuttle take 5 minutes to arrive on all code levels

Reverts: #8616
## Why It's Good For The Game
Courtesy of Kyet:
- removes incentive to keep the station on unnecessary red alert just for faster shuttle
- better aligns with what players want - when players want the round to end they generally want it to end now, or at least soon, instead of taking an extra 5m because its on green
- reduces chance that antags who wait until the shuttle is called to do their objectives will be able to greentext, which acts as a bit of a discouragement for antags doing this, which is good, because it makes rounds more interesting since antags will be trying to do their objectives throughout the shift instead of waiting for the end

:cl:
tweak: The shuttle will now always take 5 minutes
/ :cl:

### DIFF
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -13,10 +13,9 @@ SUBSYSTEM_DEF(shuttle)
 		//emergency shuttle stuff
 	var/obj/docking_port/mobile/emergency/emergency
 	var/obj/docking_port/mobile/emergency/backup/backup_shuttle
-	var/emergencyCallTime = 6000	//time taken for emergency shuttle to reach the station when called (in deciseconds)
+	var/emergencyCallTime = 3000	//time taken for emergency shuttle to reach the station when called (in deciseconds)
 	var/emergencyDockTime = 1800	//time taken for emergency shuttle to leave again once it has docked (in deciseconds)
 	var/emergencyEscapeTime = 1200	//time taken for emergency shuttle to reach a safe distance after leaving station (in deciseconds)
-	var/emergency_sec_level_time = 0 // time sec level was last raised to red or higher
 	var/area/emergencyLastCallLoc
 	var/emergencyNoEscape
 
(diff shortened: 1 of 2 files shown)


farie82:
I don't like this myself to be honest. The server seems to be pushing more towards the action oriented stuff. And this is just another step to that. Call the shuttle once the antags are gone or dead. Then better have it come quick so we can do it all over again.

Having a 10 minute shuttle timer isn't a bad thing if things are quiet. Means you can finish what you're doing and call it a day. If rounds are boring and players want it to end as fast as possible then maybe there is something wrong with the design of the game or there is a lack of events (either custom or the games own events)

Ty-Omaha:
I agree that most of the events are boring and we need to add new ones and tweak/change them

I think there also needs to be more meaningful things for people to do in general. Engineering basically sits around waiting for things to blow up except doing the station projects

Mitchs98:
Honestly against this for the reasons Farie said, and due to the fact one of the reasons would encourage people to rush their objectives in the round in order to 'win'; which is something I'm of the opinion we should discourage. I really don't think it'll make rounds more interesting, it'll just shift people to rushing their objectives to 'greentext' then stealthing the rest of the round.

Granted we're often on red anyway, it wouldn't occur that often, but my concern is rounds that it will.

Carthusia:
I'm 50/50 on this as well

If its red alert, everyone wants to get out alive as soon as possible, if its blue/green, then it's probably a calm shift and people are doing their thing enjoying themselves.

Maybe if it was also blue that made it a 5 minute call, but an expedited call on green seems odd.

Reducing it to 5 all around also has the implication that CC will not expedite a shuttle during an emergency, unless you want to 2 and a half minute call for red alerts, which im against.

Ty-Omaha:
>has the implication that CC will not expedite a shuttle during an emergency

Just think of it at 100% speed all the time

Carthusia:
You might say the shuttle only has one setting....ludicrous speed.

Just need to make the space that you see outside all plaid and we're good to go.

MRgaty:
Not a bad change imo, rounds are already pretty long for my taste anyways

Kyep:
Approving this, for all the reasons stated in the OP.

Landerlow:
I'm with Farie on this one.

Fox-McCloud:
Opposed to this. It's going to entrench, evne more, the idea of "antags dead, call shuttle".

Given that this is a roleplaying game and not a freaking *job simulator*, I don't think this change fits well with that vision. I don't think it's really going to have an impact on "last minute antag shenanigans" either.

This *might* slightly reduce the incentive to stay on red, but I have my doubts.....even then, I'd be more inclined to remove the "red alert = faster shuttle" than to make the shuttle take 5 minutes.

This doesn't have a good feel to it at all.

PR #13200, opened 2020-03-25
Make spell buttons light up slowly when recharging

farie82:
## What Does This PR Do
Makes spell action buttons slowly light up while they are recharging. A percentage is also shown below

## Why It's Good For The Game
Adds visual feedback to the user to see when the ability will be available again.

## Images of changes
lFdy4YptRJ

## Changelog
:cl:
add: Spell buttons now visually recharge by slowly regaining their color and a percentage completed is shown below
/:cl:

### DIFF
--- a/code/datums/action.dm
+++ b/code/datums/action.dm
@@ -463,6 +463,7 @@
 /datum/action/spell_action
 	check_flags = 0
 	background_icon_state = "bg_spell"
+	var/recharge_text_color = "#FFFFFF"
 
 /datum/action/spell_action/New(Target)
 	..()
@@ -501,15 +502,17 @@
 	return 0
 
 /datum/action/spell_action/UpdateButtonIcon()
-	if(button && !..())
+	if(button && !(. = ..()))
 		var/obj/effect/proc_holder/spell/S = target
 		if(!istype(S))
 			return
 		var/progress = S.get_availability_percentage()
 		var/col_val_high = 72 * progress + 128
 		var/col_val_low = 200 * progress
+		button.maptext = "<div style=\"font-size:6pt;color:[recharge_text_color];font:'Small Fonts';text-align:center;\" valign=\"bottom\">[round_down((1-progress) * 100)]%</div>"
 		button.color = rgb(col_val_high, col_val_low, col_val_low, col_val_high)
-
+	else
+		button.maptext = null
 /*
 /datum/action/spell_action/alien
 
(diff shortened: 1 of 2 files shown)


code/datums/action.dm
```diff
@@ -500,6 +500,16 @@
...
 
+/datum/action/spell_action/UpdateButtonIcon()
+	if(button && !..())
+		var/obj/effect/proc_holder/spell/S = target
+		if(!istype(S))
+			return
+		var/progress = S.get_availability_percentage()
+		var/col_val_high = 72 * progress + 128
+		var/col_val_low = 200 * progress
+		button.color = rgb(col_val_high, col_val_low, col_val_low, col_val_high)
```
Crazylemon64, line 511: I'm a bit nervous that there's not a clear delineation between "99% recharged" and "ready"

Could there be some kind of skip of the color level at 100%?

Crazylemon64: Or, wait - the gif you provide seems to imply that it does
let's see, `progress` is a value between `0` and `1`
at 0, `col_val_high = 128` and `col_val_low = 0`
at 1, `col_val_high = 200` and `col_val_low = 200`
What about this would cause the distinct color change between 0.99 completion and 1.0 completion...?

farie82:
Will change it to be more clear once I got time again

code/datums/action.dm
farie82: The way it's currently implemented will make the color shift from about 78% of the full color to 100% of the sprite. I'll see about making this more clear though

farie82:
Changed it to have a percentage charged on the button icon.
Might experiment a bit more with other visualizations later.
I've not yet tested this with stuff like wizard magic that can't be used if you don't wear clothes. I'll test/fix those later

code/datums/action.dm
```diff
@@ -500,6 +501,18 @@
 		return spell.can_cast(owner)
 	return 0
 
+/datum/action/spell_action/UpdateButtonIcon()
+	if(button && !(. = ..()))
```
SteelSlayer, line 505: The `!(. = ..())` here with the assignment of `.` inside the if statement is weird, does this actually work properly? I would think it needs to be `(!..())` or something else with no assignment.

farie82: In languages like C and C++ this is valid syntax. Same as in byond.
Written out it will act the same as:
```
if(!button)
  return
. = ..()
if(!.)
  stuf_here()
```
I prefer to write it like this due to the 3 lines it saves

SteelSlayer: Well then, didn't know you could do that, carry on :smile:

Fox-McCloud:
@Crazylemon64

variableundefined (approved):
Doesn't work with lightning. That's lightning's fault and should be fixed in a separate PR.

PR #13201, opened 2020-03-26
fix slime grenades.

Evankhell561:
## What Does This PR Do
fix the interaction of slime cores with large grenades

## Why It's Good For The Game
bugs are bad

## Changelog
:cl:
fix: slime grenades
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/chem_grenade.dm
+++ b/code/game/objects/items/weapons/grenades/chem_grenade.dm
@@ -349,8 +349,11 @@
 					for(var/obj/item/reagent_containers/glass/G in beakers)
 						S.reagents.trans_to(G, S.reagents.total_volume)
 				else
-					S.forceMove(get_turf(src))
 					no_splash = TRUE
+					if(beakers.len)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/grenades/chem_grenade.dm
```diff
@@ -349,8 +349,11 @@
 					for(var/obj/item/reagent_containers/glass/G in beakers)
 						S.reagents.trans_to(G, S.reagents.total_volume)
 				else
-					S.forceMove(get_turf(src))
 					no_splash = TRUE
+					if(beakers.len)
+						for(var/obj/item/slime_extract/O in beakers)
+							O.forceMove(get_turf(src))
+							beakers = list()
```
Crazylemon64, line 356: why are you setting the list inside of the loop
also why isn't the one at line 352 good enough - is the trouble that slime cores without uses are getting skipped? Is the trouble that the cores aren't being removed from the `beakers` list?

Fox-McCloud:
Hasn't been addressed since April 11th. Closing.

PR #13206, opened 2020-03-26
Orange Watchlist Names

Ty-Omaha:
People placed on a watchlist will now have Orange names to alert admins. Mentors are unable to see the orange names

If the name is orange in the player panel you can click it to bring up the watchlist reason

I changed the color after taking the screenshots, they are now orange instead of red.
image
image
image

This is the color:
image
:cl:
add:People on a watchlist will now have orange names to admins. Admins can click the name in the player panel to bring up the watchlist reason
/ :cl:

### DIFF
--- a/code/defines/procs/admin.dm
+++ b/code/defines/procs/admin.dm
@@ -12,6 +12,7 @@
 	var/mob/M
 	var/client/C
 	var/key
+	var/watchlisted = FALSE
 
 	if(!whom)
 		return "INVALID/(INVALID)"
@@ -39,6 +40,9 @@
 
 	. = ""
 
+	if(check_watchlist(key) && is_admin(src))
+		watchlisted = TRUE
+
 	if(key)
 		if(C && C.holder && C.holder.fakekey && !include_name)
 			if(include_link)
@@ -46,7 +50,10 @@
 			. += "Administrator"
 		else
 			if(include_link && C)
-				. += "<a href='?priv_msg=[C.UID()];type=[type]'>"
+				if(watchlisted)
+					. += "<a style='color: #FF0000;' href='?priv_msg=[C.UID()];type=[type]'>"
+				else
+					. += "<a href='?priv_msg=[C.UID()];type=[type]'>"
 			. += key
 
 		if(include_link)
--- a/code/defines/procs/admin.dm
+++ b/code/defines/procs/admin.dm
@@ -51,7 +51,7 @@
 		else
 			if(include_link && C)
 				if(watchlisted)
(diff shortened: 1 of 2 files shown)


code/defines/procs/admin.dm
```diff
@@ -39,14 +40,20 @@
...
 	if(key)
 		if(C && C.holder && C.holder.fakekey && !include_name)
 			if(include_link)
 				. += "<a href='?priv_msg=[C.findStealthKey()];type=[type]'>"
 			. += "Administrator"
 		else
 			if(include_link && C)
-				. += "<a href='?priv_msg=[C.UID()];type=[type]'>"
+				if(watchlisted)
+					. += "<a style='color: #db9737;' href='?priv_msg=[C.UID()];type=[type]'>"
```
Kyep, line 54: Not sure if its good practice to use color codes explicitly instead of a class.

farie82: I agree with Kyet. Best to make a special class or reuse one. This will make it more robust when people want to reuse this later or when say darkmode support will be added

code/modules/admin/admin.dm
```diff
@@ -68,7 +68,12 @@ GLOBAL_VAR_INIT(nologevent, 0)
 	var/body = "<html><head><title>Options for [M.key]</title></head>"
 	body += "<body>Options panel for <b>[M]</b>"
 	if(M.client)
-		body += " played by <b>[M.client]</b> "
+
+		if(check_watchlist(M.ckey)
+			body += " played by <a style = 'color: #db9737;' href = '?_src_=holder;watchsearch=[M.ckey];'><b>[M.client]</b></a> "
```
farie82, line 73: spacing in html spans and what not is usually limited:
So instead of `<span style = 'color: #val' href = '?link'>` you can use `<span style='color: #value' href='?link'>`

code/defines/procs/admin.dm
Ty-Omaha: It already supports darkmode, orange is fine in light and dark. I don't want to create a class just to make this 1 thing orange.

code/defines/procs/admin.dm
```diff
@@ -39,14 +40,20 @@
 
 	. = ""
 
+	if(check_watchlist(key) && is_admin(src))
```
AffectedArc07, line 43:
```
code/defines/procs/admin.dm, line 43, column 29:
error: undefined var: "src"
```
`src` is null here because this proc isnt attached to a mob or client, its just a bare proc with no parent.

Fox-McCloud:
1 week to fix this.

Fox-McCloud:
Stale. Closing

PR #13209, opened 2020-03-26
Check antag improvements

Ty-Omaha:
For the Check Antagonist panel admins have
Adds FLW button
Displays if antagonist is in perma

(the spacing of (PERMA) is fixed)
image

:cl:
add:Admins can now follow antagonists and see if they are in the permabrig when opening the check antagonist panel
/ :cl:

### DIFF
--- a/code/modules/admin/player_panel.dm
+++ b/code/modules/admin/player_panel.dm
@@ -392,11 +392,12 @@
 	var/logout_status
 	logout_status = M.client ? "" : " <i>(logged out)</i>"
 	var/dname = M.real_name
+	var/area/A = get_area(M)
 	if(!dname)
 		dname = M
 
-	return {"<tr><td><a href='?src=[UID()];adminplayeropts=[M.UID()]'>[dname]</a><b>[caption]</b>[logout_status][M.stat == 2 ? " <b><font color=red>(DEAD)</font></b>" : ""]</td>
-		<td><A href='?src=[usr.UID()];priv_msg=[M.client ? M.client.UID() : null]'>PM</A></td>[close ? "</tr>" : ""]"}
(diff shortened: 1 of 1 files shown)


code/modules/admin/player_panel.dm
```diff
@@ -492,7 +493,7 @@
 			dat += check_role_table("Ninjas", ticker.mode.ninjas)*/
 
 		if(SSticker.mode.cult.len)
-			dat += check_role_table("Cultists", SSticker.mode.cult, 0)
+			dat += check_role_table("Cultists", SSticker.mode.cult)
```
Crazylemon64: This means the objectives of the cultists would individually be displayed, right? So if we had 10 cultists, the same objectives would be printed 10 times, right?

Ty-Omaha: It just shows the "Show Objectives" button

Kyep:
A few extra requests:

- Sometimes, antags will show up in CA as "not found". Instead of showing these "not found" entries, just omit that line. We cannot do anything with a 'not found' entry so its pointless to display it. I assume these are antag minds whose bodies have been destroyed.
- For antags that have hijack, display a (HIJACK) next to their names.
- Instead of showing (PERMA) just show the name of the area they're in, with the name displayed in red if the area has the 'fast_despawn' flag on the /area. So for example if they're in brig or perma it shows in red. Otherwise its like "Joe Scmoe (Bar Maintenance)".
- Show the count of sec members who are alive, nonafk, nonssd and not antags at the top of the CA panel.  E.g: "Active Security: 6" This will help us quickly determine balance of sec v antags for purposes like deciding whether to grant an ERT request.

Also a bonus idea for another PR (not this PR):
- If possible, in traitor panel, compute the actual completion status of each objective and display that. For example looking at a traitor with kill X and steal Y objectives they'd show as complete if X was dead and Y was in their inventory, but incomplete otherwise. Helps us gauge how far along someone is with their objectives. I admit this may be out of scope for this PR.

PR #13214, opened 2020-03-27
Increases ATM Withdrawl Limit

Mitchs98:
## What Does This PR Do
Increases the max limit for ATM withdrawls from 10k to 100k. It's not often people get over 10k, but with some admin run events and slot winnings it's relevant. Mostly for the admin event side of things. Now, the comment on the line I edited said that it was done that way to prevent crashes. I personally noticed no CPU spikes when testing locally, and don't really think it would affect anything personally. I could be wrong though.

## Why It's Good For The Game
Allows events that involve a lot of credits being spawned/deposited/stolen and what not to run a lot more smoothly, withdrawing 10k at once is a huge pain in the rear and becomes tedious fast.

## Changelog
:cl:
tweak: You can now withdraw up to 100k at a time from an ATM.
/:cl:

### DIFF
--- a/code/modules/economy/ATM.dm
+++ b/code/modules/economy/ATM.dm
@@ -237,9 +237,9 @@ log transactions
 						playsound(src, 'sound/machines/chime.ogg', 50, 1)
 
 						//remove the money
-						if(amount > 10000) // prevent crashes
+						if(amount > 100000) // prevent crashes
 							to_chat(usr, "<span class='notice'>The ATM's screen flashes, 'Maximum single withdrawl limit reached, defaulting to 10,000.'</span>")
-							amount = 10000
+							amount = 100000
 						withdraw_arbitrary_sum(amount)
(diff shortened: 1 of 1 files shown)


code/modules/economy/ATM.dm
```diff
@@ -237,9 +237,9 @@ log transactions
 						playsound(src, 'sound/machines/chime.ogg', 50, 1)
 
 						//remove the money
-						if(amount > 10000) // prevent crashes
+						if(amount > 100000) // prevent crashes
 							to_chat(usr, "<span class='notice'>The ATM's screen flashes, 'Maximum single withdrawl limit reached, defaulting to 10,000.'</span>")
```
farie82: Did not update the text here

Mitchs98: Aw hell, you're right >_<. Thanks for that. Knew I forgot something because I'm a derp xD

PR #13216, opened 2020-03-28
Allow the Cloning Pod to accept monster meat as a source of biomass

nok-ko:
## What Does This PR Do

It makes goliath meat and other exotic xeno meats (spider meat, bear meat, etc.) into instances of `/obj/reagent_containers/food/snacks/meat`, and therefore accepted by the cloning pod as biomass. Special case: they however cannot be made into cutlets, as other `meat` can.

## Why It's Good For The Game

It is currently very possible to construct a cloning setup on the mining outpost, but gathering biomass for a clone requires a stationside trip, since goliath meat is arbitratily not accepted as a biomass source by the cloning pods. This PR removes this issue! It also makes butchering xenos and spiders more useful.

## Changelog
:cl: Nokko
tweak: made the Cloning Pod accept goliath, xeno, spider, bear, and lizard meat
/:cl:

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -4,7 +4,19 @@
 //Potential replacement for genetics revives or something I dunno (?)
 
 #define CLONE_BIOMASS 150
+
 #define BIOMASS_BASE_AMOUNT 50 // How much biomass a BIOMASSABLE item gives the cloning pod
+
+// Not a comprehensive list: Further PRs should add appropriate items here.
+// Meat as usual, monstermeat covers goliath, xeno, spider, bear meat
+GLOBAL_LIST_INIT(cloner_biomass_items, list(\
+/obj/item/reagent_containers/food/snacks/meat,\
+/obj/item/reagent_containers/food/snacks/monstermeat,
+/obj/item/reagent_containers/food/snacks/carpmeat,
+/obj/item/reagent_containers/food/snacks/salmonmeat,
+/obj/item/reagent_containers/food/snacks/catfishmeat,
+/obj/item/reagent_containers/food/snacks/tofurkey))
+
 #define MINIMUM_HEAL_LEVEL 40
 #define CLONE_INITIAL_DAMAGE 190
 #define BRAIN_INITIAL_DAMAGE 90 // our minds are too feeble for 190
@@ -307,7 +319,7 @@
 /obj/machinery/clonepod/process()
 	var/show_message = 0
 	for(var/obj/item/item in range(1, src))
-		if (item.BIOMASSABLE)
+		if(is_type_in_list(item, GLOB.cloner_biomass_items))
 			qdel(item)
 			biomass += BIOMASS_BASE_AMOUNT
 			show_message = 1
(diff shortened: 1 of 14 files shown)


nok-ko:
A question: why don't terror spiders have butcher results?

nok-ko:
@farie82 I briefly checked over and made sure that you can't make cutlets and bacon from monster meat, and I changed the references in the cooking recipe files (for monstermeat-specific recipes) to the new names. I don't believe anything ever calls for raw meat and not cutlets, but I will double-check and report back.

nok-ko:
Honestly I've done this a little ass-backwards. @farie82 pointed out in Discord that changing the subtype of the monster meats to `meat` is a far more wide-reaching change than what my commit really is about: makng the Cloning Pod accept more logical sources of biomass/protein. I'll adjust this PR significantly today.

nok-ko:
Everything works properly now, as far as I can tell. Should I edit my original description and update the changelog and stuff? My implementation and the functionality has changed since the first commit.

code/game/machinery/cloning.dm
```diff
@@ -364,12 +378,12 @@
 		use_power(200)
 
 //Let's unlock this early I guess.  Might be too early, needs tweaking.
-/obj/machinery/clonepod/attackby(obj/item/I, mob/user, params)
-	if(exchange_parts(user, I))
+/obj/machinery/clonepod/attackby(obj/item/item, mob/user, params)
```
farie82: No need to rename `I` to `item`. `I` is more commonly used. Usually people call scoped variables in this fashion.

code/modules/food_and_drinks/food/snacks.dm
```diff
@@ -273,6 +273,8 @@
 //		reagents.add_reagent("nutriment", 2)							//	this line of code for all the contents.
 //		bitesize = 3													//This is the amount each bite consumes.
 
+// Remember to add anything that should be accepted as biomass by the cloning pods into the GLOB.cloner_biomass_items list.
```
farie82: No need to place comments here in my opinion. The file has nothing to do with cloning itself

code/game/machinery/cloning.dm
```diff
@@ -302,13 +315,14 @@
...
 	var/show_message = 0
-	for(var/obj/item/reagent_containers/food/snacks/meat/meat in range(1, src))
-		qdel(meat)
-		biomass += BIOMASS_MEAT_AMOUNT
-		show_message = 1
+	for(var/obj/item/item in range(1, src))
+		if(is_type_in_list(item, GLOB.cloner_biomass_items))
+			qdel(item)
+			biomass += BIOMASS_BASE_AMOUNT
+			show_message = 1
```
farie82: show_message is a boolean value. Standard is to use the defines `TRUE` or `FALSE` for those.
It's also set to 0 above on line 320. That one can also be replaced by a bool define

code/game/machinery/cloning.dm
nok-ko: Okay, thanks. I don't know a lot of convention, thanks for showing me.

code/game/machinery/cloning.dm
nok-ko: I just worked my code around what was there previously, didn't look closely enough to realize that it could be replaced with the defines. Will do!

code/modules/food_and_drinks/food/snacks.dm
nok-ko: Fair enough.

nok-ko:
Implemented all of @farie82's style suggestions in the latest commit. Woo-hoo, my first actual PR!

VeteranKyl:
Surely nothing can go wrong with using xeno based biomass to regrow a person.

nok-ko:
@VeteranKyl MEAT IS MEAT.

PR #13217, opened 2020-03-28
Revert "Goonchem: Addiction Rework"

Spacemanspark:
Reverts ParadiseSS13/Paradise#13034

But Spaaaark, wh-
This is largely here because I've seen nothing good come from the system mechanics or roleplay-wise. While this might fit on Goonstation, it doesn't mesh well here. I haven't seen a single good comment on the new system. 

This all said, I understand there are PRs out there to change and rebalance the system. Should those be accepted and make all sides happy, I'll happily close this after a period of time if I see no further complaints on the system. Improve over removal, yes. 
However, If the majority of players still hold issues, I'm leaving this up. 

Hi yes CL stuff ahoy
:cl:
del: Reverts the addiction system back to the way it was prior to the goon port.
/ :cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -405,7 +405,6 @@
 		if(C.reagents)
 			C.reagents.clear_reagents()
 			QDEL_LIST(C.reagents.addiction_list)
-			C.reagents.addiction_threshold_accumulated.Cut()
 
 // rejuvenate: Called by `revive` to get the mob into a revivable state
 // the admin "rejuvenate" command calls `revive`, not this proc.
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -9,7 +9,6 @@
 	var/atom/my_atom = null
 	var/chem_temp = T20C
 	var/list/datum/reagent/addiction_list = new/list()
-	var/list/addiction_threshold_accumulated = new/list()
 	var/flags
 	var/list/reagents_generated_per_cycle = new/list()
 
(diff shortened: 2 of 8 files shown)


Carthusia:
Posted before Spark added the PR Description:

I assume this will revert the whole "You are addicted to (Black Russian, Vodka, Absinthe) kind of addictions that keep happening when people just have a few drinks at the bar or eat some icecream with liquor in the mix?

Spacemanspark:
Yes

Mitchs98:
Please do. I haven't been able to actually drink or smoke as any of my characters due to this. Either this, or a decent bump in the thresholds need to happen.

Denthamos:
While I do like the idea of addictions taking place with the addiction system on paper, in practice it makes people just avoid anything that may get them addicted.

MRgaty:
Till it can get tweaked I think we are better without it.

Spacemanspark:
I might close this since alcohol addiction was removed. Cigarette addiction is still an issue in my eyes, however.

Carthusia:
The cig addiction is a bit of an issue, i smoked a pipe for 30 seconds and became addicted, and not long after began to have withdrawls, even though the pipe was still burning.

I think the pipe got overlooked when it came to the rebalance.

Mitchs98:
Cig addiction is still very much an issue. Smoked last night for roughly 20-30 seconds and was addicted. Put it out and not long after symptoms kicked in. I really don't feel nicotine is that abusable to have such heavy handed addiction symptoms or chances.

Fox-McCloud:
Going to close this. Ethanol is no longer addictive and the addiction fix means if you chainsmoke you won't get a nicotine addiction.

PR #13220, opened 2020-03-29
Makes most targeted spells select a target on click with a radius

farie82:
## What Does This PR Do
Makes most targeted spells use a mouse click to select a target. The combat focused spells will look in a radius around the clicked atom if no suitable target is found.
The current code uses lists. Which is shit.

For instance glare has a radius of 1 and thus will try to target anything around the center.
So (O is center):
XXX
XOX
XXX

By default it will select the closest targets from the click position. Giving priority to the clicked atom if it's valid

## Why It's Good For The Game
Having a selection menu that pulls focus on a combat focused ability is a horrible design.
This will remove frustration from using the ability. Other spells have also been included if they made sense to include.

## Images of changes
GaeOsIRacM

Also fixes a few runtimes that occurs when a player with a fireball active DCs or otherwise goes out of the body.

## Changelog
:cl:
tweak: Most targeted spells that used lists to target now use clicking. A target will be searched for in a defined radius around the click. Combat spells will have a radius of 1 generally while others might require an accurate click
/:cl:

### DIFF
--- a/code/datums/spell.dm
+++ b/code/datums/spell.dm
@@ -444,8 +444,23 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 		if(!max_targets)
 			targets.Add(found_others)
 		else
-			while(targets.len < max_targets && found_others.len) // Add the others
-				targets.Add(pick_n_take(found_others))
+			if(max_targets <= found_others.len + targets.len)
+				targets.Add(found_others)
+			else
+				switch(random_target_priority) //Add in the rest
+					if(TARGET_RANDOM)
+						while(targets.len < max_targets && found_others.len) // Add the others
+							targets.Add(pick_n_take(found_others))
+					if(TARGET_CLOSEST)
+						var/list/distances = list()
+						for(var/target in found_others)
+							distances[target] = get_dist(user, target)
+						sortTim(distances, /proc/cmp_numeric_asc, TRUE) // Sort on distance
+						for(var/target in distances)
+							targets.Add(target)
+							if(targets.len >= max_targets)
+								break
+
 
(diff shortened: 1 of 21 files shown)


Fox-McCloud:
It might be best to make this auto-target if there's just 1 person in range, but rely on the click method, otherwise.

farie82:
> It might be best to make this auto-target if there's just 1 person in range, but rely on the click method, otherwise.

It will bring some design issues with them. Since for instance cryokenesis can be used on yourself. So if you accidently use it you'll hurt yourself this way. Unless you make an exception for the user itself.

I'll think about a way to do this elegantly

farie82:
Updated the PR to include auto targeting if only one target was found. Currently only works for singular targets. No other spell uses more targets anyway yet.

code/datums/spell.dm
```diff
@@ -114,10 +130,14 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
...
 
-/obj/effect/proc_holder/spell/proc/cast_check(skipcharge = 0, mob/living/user = usr) //checks if the spell can be cast based on its settings; skipcharge is used when an additional cast_check is called inside the spell
-	if(((!user.mind) || !(src in user.mind.spell_list)) && !(src in user.mob_spell_list))
-		to_chat(user, "<span class='warning'>You shouldn't have this spell! Something's wrong.</span>")
-		return 0
+/* Checks if the user can cast the spell
+ * @param charge_check If the proc should do the cooldown check
+ * @param start_recharge If the proc should set the cooldown
+ * @param user The caster of the spell
+*/
```
AffectedArc07: mm yes documentation.

PR #13226, opened 2020-03-30
primed nukes explode if blobbed

TDSSS:
## What Does This PR Do
This PR makes it so that when a blob eats an active, primed nuke that's counting down, it causes the nuke to explode.

## Why It's Good For The Game
Nukes are set to end the round, a blob eating the nuke means the round drags on and since midround blobs don't have a win condition it just forces admins to send in a deathsquad with another nuke while all the ghosts watch.

It also removes a source of confusion regarding rules on is it allowed for blobs to rush nukes or not, since it makes it into a pretty bad idea now.

## Changelog
:cl:
tweak: blobs eating active nukes makes them blow up early.
/:cl:

### DIFF
--- a/code/game/gamemodes/nuclear/nuclearbomb.dm
+++ b/code/game/gamemodes/nuclear/nuclearbomb.dm
@@ -51,8 +51,7 @@ GLOBAL_VAR(bomb_set)
 		GLOB.bomb_set = 1 //So long as there is one nuke timing, it means one nuke is armed.
 		timeleft = max(timeleft - 2, 0) // 2 seconds per process()
 		if(timeleft <= 0)
-			spawn
-				explode()
+			INVOKE_ASYNC(src, .proc/explode)
 		SSnanoui.update_uis(src)
 	return
 
@@ -348,8 +347,7 @@ GLOBAL_VAR(bomb_set)
 	if(timing == -1.0)
 		return
 	if(timing)	//boom
-		spawn
-			explode()
+		INVOKE_ASYNC(src, .proc/explode)
 		return
 	qdel(src)
 
--- a/code/game/gamemodes/nuclear/nuclearbomb.dm
+++ b/code/game/gamemodes/nuclear/nuclearbomb.dm
@@ -347,6 +347,10 @@ GLOBAL_VAR(bomb_set)
 /obj/machinery/nuclearbomb/blob_act(obj/structure/blob/B)
 	if(timing == -1.0)
 		return
+	if(timing)	//boom
+		spawn
+			explode()
+		return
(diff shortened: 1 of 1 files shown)


Darkmight9:
oh yes thank you please

code/game/gamemodes/nuclear/nuclearbomb.dm
```diff
@@ -347,6 +347,10 @@ GLOBAL_VAR(bomb_set)
 /obj/machinery/nuclearbomb/blob_act(obj/structure/blob/B)
 	if(timing == -1.0)
 		return
+	if(timing)	//boom
+		spawn
```
farie82: spawn(0) is usually used.

TDSSS: I copied it directly from the normal nuke explosion code. What's the practical difference between spawn and spawn(0)?

farie82: Should be the same I believe. Both are done after the proc is done "async". More a styling thing

Fox-McCloud: use invoke async. Spawns are extremely discourage these days.

TDSSS: Alright. Want me to change it in the process() proc too while I'm there?

TDSSS:
Should use invoke async now

daylight2:
Seems like this would make rushing the nuke even more of a priority, because if you wait too long, you're fucked

Cazdon:
Well, all it takes now is for a blob mouse to pop inside the vault for the round to instantly end, I'm not sure this is the best option and will just lead to another set of problems regarding that.

code/game/gamemodes/nuclear/nuclearbomb.dm
```diff
@@ -347,6 +346,9 @@ GLOBAL_VAR(bomb_set)
 /obj/machinery/nuclearbomb/blob_act(obj/structure/blob/B)
 	if(timing == -1.0)
 		return
+	if(timing)	//boom
+		INVOKE_ASYNC(src, .proc/explode)
```
Kyep, line 350: Will this still run if the object is qdel()ed?

TDSSS: It won't work correctly if the object gets qdeled. That's why I have it return early so the nuke isn't deleted.

Kyep: Yeah, fair.

Issue #12344, opened 2019-09-18
Kinetic Accelerators Runtime on Roundstart


Fox-McCloud:
They try to use a cell that's deleted. RIP.

Started being a problem when KA's used `addtimer` instead of a `spawn`.

PR #13227, opened 2020-03-30
Makes the KA and energy guns use Initialize instead of New fixing the KA runtime

farie82:
## What Does This PR Do
Makes energy guns and such use Initialize instead of New. Which caused some issues

Fixes: #12344

## Why It's Good For The Game
Less runtimes on roundstart

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -341,8 +341,8 @@
 	var/emagged = 0			//ups the temperature cap from 500 to 1000, targets hit by beams over 500 Kelvin will burst into flames
 	var/dat = ""
 
-/obj/item/gun/energy/temperature/New()
-	..()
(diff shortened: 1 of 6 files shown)


code/modules/projectiles/guns/energy/special.dm
```diff
@@ -341,8 +341,8 @@
 	var/emagged = 0			//ups the temperature cap from 500 to 1000, targets hit by beams over 500 Kelvin will burst into flames
 	var/dat = ""
 
-/obj/item/gun/energy/temperature/New()
-	..()
+/obj/item/gun/energy/temperature/Initialize(mapload, ...)
```
AffectedArc07: What is the use of `...` in this proc?

farie82: I believe that is a holder for arguments or such. Using `args` you can get them. I think this is what my intellisense generated when I selected Initialize

PR #13231, opened 2020-03-30
Removes Alcohol Addiction

Fox-McCloud:
As much as I like the idea of having alcohol addiction as a form of alcoholism, I don't think it's really worth it, in this case--especially since alcohol's combat advantage is virtually non-existent.

Even if  addiction is tweaked to ignore subtypes, I'm still not sure having alcohol cause addiction is worth it---so I'm opting to just can it, completely.

Way back in the day (2015), alcohol was addictive then, as well---it was also removed back then because it was just too annoying, most of the time. Seems it's still the case.

:cl: Fox McCloud
del: Alcohol no longer has an addiction
/:cl:


Carthusia:
Huzzah! I can drink once again without having to worry about gettin the Absinthe shakes

PR #13233, opened 2020-03-31
Processing refactor; Code and performance improvements

### LINKED ISSUE #12268: Cloning with cryo sleepers
**Issue Description**:
Cryoing someone while neck grabbing them occupies the sleeper but the person you are grabbing does not gets inside.
 

To have them actually stuffed inside the pod, stopping the neck grab.

**What happened instead**:
The sleeper got occupied but I still had the person grabbed out of the sleeper

You can prevent people from cryoing as you can have every sleeper full with just a single player.

Neck grab the player, proceed to put them inside the sleeper, then take a few steps away from it while still having them grabbed.

3 days ago or so.

**Extra information**: 
You can free the pods if the person you cryo'd wakes up and enters the same pods you stuffed them in.

### LINKED ISSUE #13102: Charging Pods are Secretly Assisted-Teleportation Machines
**Issue Description**:
You, when grabbing somebody as an IPC, can store yourself into a charging station and, from any distance except for at Central Command, can translocate yourself instantly to its location from inside mechs, restraints, or otherwise, with the assistance of one other person pressing the 'eject' button.

Normally charging pods aren't telepads.

**What happened instead**:
The charging pod was, instead, a telepad.

You can cross z-level translocate yourself across the infinitesimal warp with speeds high enough to instantaneously eviscerate your fragile aluminum body, but appear fine. This can be done to escape a myriad of situations, but it's mostly good fun if you're bored and want to show off some magic tricks.

Basically, it's teleportation.

Be IPC
Grab mob
Shimmy into charging pod
Your Mob gets stored into the pod but never actually enters it
Other person ejects charging pod
Teleport on top of now ejected pod
Profit

Not Applicable

**Extra information**:
It's one hell of a fun gag.

### LINKED ISSUE #13103: Schrödinger's HardHat - IPC teleporting
**Issue Description**:

It is not really a problem just an abusable bug. Any IPC can teleport to any cyborg rechargers by doing a little setup. It works over every Z levels. The teleport happens when somebody uses the eject on the cyborg recharger that been set up with this bug.

Found it by accident was not expecting anything.

**What happened instead**:

HardHat Magic, that what happened. The IPC moved to the same tile where the cyborg recharger is, the cyborg recharger showed that somebody was inside but the IPC was not.

You can teleport from anywhere to set locations and you can't even tell if somebody teleported since it looks like they just move out of the recharger.

Be an IPC. Stand next to a cyborg recharger that is EMPTY. Grab intent -> grab somebody. (monkey, other player), Move to the tile where the cyborg recharger is. Done.

This bug might as be as ancient as the game is also I named it Schrödinger's HardHat. :P

**Extra information**:

Sorry for not being serious with this issue report. I just already miss my HardHat magic. :( (side note, I never abused it or used it aside showing to a random vox who was needed for the first test so I could confirm the bug existence, Torque - Cazdon(admin), Reagan, Frost, and Slade - Dumbdumn5 admin)

SteelSlayer:
## What Does This PR Do
Republish of #12952, but this one is a bit different and a bit more expanded. I've added a lot of documentation and comments for new code that hopefully explains things.

This refactors a bunch of machinery so it only starts processing when it needs to, and stops when there's no longer a need for it to run. Also adds in various useful signals and just general cleanup to certain bits of the code.

For now I've tried to tackle the machines which appear the most in game, such as solar panels, holopads, intercoms, lightswitches and so on. In the future, if and when this goes through, I'll probably look into fixing up some of the other ones.

Almost all of the changes in here are backend changes, but there is one tweak I made. You can no longer grab a mob and bring them into cyborg rechargers or sleepers with you. Only 1 mob will be allowed in these machines at a time. There was a bunch of pointless code running for these machines involving ejecting stuck mobs, and I figured, why are people allowed to do this in the first place?

## Why It's Good For The Game
Should increase the performance of the server a bit since we won't have a bunch of code pointlessly executing when it doesn't need to.

This PR fixes the weird teleporting bug with cyborg rechargers. I know this PR wont be merged for a while, so if someone wants to fix these sooner though, go ahead.
Fixes #13103
Fixes #13102
Fixes #12268

## Changelog
:cl:
tweak: You can no longer enter a sleeper, cyborg recharger, or cryopod while you have a mob in a grab
/:cl:

### DIFF
--- a/code/datums/holocall.dm
+++ b/code/datums/holocall.dm
@@ -159,13 +159,13 @@
 /datum/holocall/proc/Check()
 	for(var/I in dialed_holopads)
 		var/obj/machinery/hologram/holopad/H = I
-		if((H.stat & NOPOWER) || !H.anchored)
+		if(!H.anchored)
 			ConnectionFailure(H)
 
 	if(QDELETED(src))
 		return FALSE
 
-	. = !QDELETED(user) && !user.incapacitated() && !QDELETED(calling_holopad) && !(calling_holopad.stat & NOPOWER) && user.loc == calling_holopad.loc
+	. = (user.loc == calling_holopad.loc) && !user.incapacitated() && !QDELETED(user) && !QDELETED(calling_holopad)
 
 	if(.)
 		if(!connected_holopad)
@@ -174,7 +174,7 @@
 				calling_holopad.atom_say("No answer received.")
 				calling_holopad.temp = ""
 
-	else if(!.)
+	if(!.)
 		qdel(src)
 
 /datum/action/innate/end_holocall
--- a/code/datums/holocall.dm
+++ b/code/datums/holocall.dm
@@ -165,7 +165,7 @@
 	if(QDELETED(src))
 		return FALSE
 
-	. = (user.loc == calling_holopad.loc) && !user.incapacitated() && !QDELETED(user) && !QDELETED(calling_holopad)
+	. = !QDELETED(user) && !QDELETED(calling_holopad) && (user.loc == calling_holopad.loc) && !user.incapacitated()
 
 	if(.)
 		if(!connected_holopad)
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -85,9 +85,6 @@
 		stack_trace("Warning: [src]([type]) initialized multiple times!")
 	initialized = TRUE
 
-	if(loc)
-		SEND_SIGNAL(loc, COMSIG_ATOM_CREATED, src) // Sends a signal that the new atom `src`, has been created at `loc`.
-
 	if(color)
 		add_atom_colour(color, FIXED_COLOUR_PRIORITY)
 
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -66,7 +66,8 @@
 		if(SSatoms.InitAtom(src, args))
 			// we were deleted
 			return
-
+		if(loc)
+			SEND_SIGNAL(loc, COMSIG_ATOM_CREATED, src) // Send a signal that a new atom was created at `loc`.
 
 //Called after New if the map is being loaded. mapload = TRUE
 //Called from base of New if the map is not being loaded. mapload = FALSE
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -84,6 +84,9 @@
 		stack_trace("Warning: [src]([type]) initialized multiple times!")
 	initialized = TRUE
 
+	if(loc)
+		SEND_SIGNAL(loc, COMSIG_ATOM_CREATED, src) // Sends a signal that the new atom `src`, has been created at `loc`.
+
 	if(color)
 		add_atom_colour(color, FIXED_COLOUR_PRIORITY)
 
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -90,15 +90,16 @@
 		end_processing()
 		return
 
-	if(filtering <= 0 || !beaker)
-		return
+	if((filtering > 0) && beaker)
+		// To prevent runtimes from drawing blood from runtime, and to prevent getting IPC blood.
+		if(!istype(occupant) || !occupant.dna || (NO_BLOOD in occupant.dna.species.species_traits))
+			filtering = 0
+			return
 
-	// To prevent runtimes from drawing blood from runtime, and to prevent getting IPC blood.
-	if(!istype(occupant) || !occupant.dna || (NO_BLOOD in occupant.dna.species.species_traits))
-		filtering = 0
-		return
+		// Beaker is full, we can't perform dialysis any more.
+		if(beaker.reagents.total_volume >= beaker.reagents.maximum_volume)
+			return
 
-	if(beaker.reagents.total_volume < beaker.reagents.maximum_volume)
 		occupant.transfer_blood_to(beaker, 1)
 		for(var/datum/reagent/x in occupant.reagents.reagent_list)
 			occupant.reagents.trans_to(beaker, 3)
@@ -409,6 +410,23 @@
 	for(var/atom/movable/A in contents - component_parts - list(beaker))
 		A.forceMove(loc)
 
+/// Try to get `target` to enter the sleeper. It's possible that `target` and `user` are the same mob, in the case of the "Enter Sleeper" verb.
+/obj/machinery/sleeper/proc/go_in(mob/living/carbon/human/target, mob/living/user)
+	if(!user || !target)
+		return
+	if(do_after(user, 20, target = target))
+		if(occupant)
+			to_chat(user, "<span class='boldnotice'>The sleeper is already occupied!</span>")
+			return
+		target.forceMove(src)
+		occupant = target
+		begin_processing()
+		icon_state = "[base_icon]"
+		to_chat(target, "<span class='boldnotice'>You feel cool air surround you. You go numb as your senses turn inward.</span>")
+		add_fingerprint(user)
+		if(user.pulling == target)
+			user.stop_pulling()
+
 /obj/machinery/sleeper/proc/inject_chemical(mob/living/user as mob, chemical, amount)
 	if(!(chemical in possible_chems))
 		to_chat(user, "<span class='notice'>The sleeper does not offer that chemical!</span>")
@@ -496,22 +514,7 @@
 		visible_message("[user] starts climbing into the sleeper.")
 	else
 		visible_message("[user] starts putting [L.name] into the sleeper.")
-
-	if(do_after(user, 20, target = L))
-		if(occupant)
-			to_chat(user, "<span class='boldnotice'>The sleeper is already occupied!</span>")
-			return
-		if(!L) return
-		L.forceMove(src)
-		occupant = L
-		begin_processing()
-		icon_state = "[base_icon]"
-		to_chat(L, "<span class='boldnotice'>You feel cool air surround you. You go numb as your senses turn inward.</span>")
-		add_fingerprint(user)
-		if(user.pulling == L)
-			user.stop_pulling()
-		return
-	return
+	go_in(L, user)
 
 /obj/machinery/sleeper/AllowDrop()
 	return FALSE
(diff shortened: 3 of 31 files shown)


Aquilar:
does this prevent you putting mobs into sleeper and rechargers?

SteelSlayer:
Yes.

Aquilar:
your gonna need to reserve that change, since unconscious mobs that need dialysis will not be able to get dialysis

Mitchs98:
Yeah that's definitely a huge problem for medbay in general. A lot of people don't know how to get into sleepers/will be in crit and unable to get into sleepers. A very much un-needed nerf for medbay all around.

SteelSlayer:
Sorry I misread, this prevents a human from grabbing another mob with grab (yellow) intent and bringing them into the sleeper with them.

Mitchs98:
Oh! I see....that's...I didn't know that was a thing at all. Yeah that's dumb as heck 🤔

Fox-McCloud:
Please actually include performance comparisons with performance PRs, unless its' somethings well established as a process that saves CPU cycles (istypeless loops, for instance).

SteelSlayer:
The performance improvement comes from only having a thing's `process()` run when its turned on or it has an occupant, etc. Or for example some of the `/mineral` type machines like the ORM will now just use signals and don't even process at all.

code/datums/holocall.dm
```diff
@@ -159,13 +159,13 @@
...
 		var/obj/machinery/hologram/holopad/H = I
-		if((H.stat & NOPOWER) || !H.anchored)
+		if(!H.anchored)
 			ConnectionFailure(H)
 
 	if(QDELETED(src))
 		return FALSE
 
-	. = !QDELETED(user) && !user.incapacitated() && !QDELETED(calling_holopad) && !(calling_holopad.stat & NOPOWER) && user.loc == calling_holopad.loc
+	. = (user.loc == calling_holopad.loc) && !user.incapacitated() && !QDELETED(user) && !QDELETED(calling_holopad)
```
farie82: The order here matters in the way that QDELETED will also check if it's null. Having it in this order might break if say the user is null or the calling_holopad is null

code/game/atoms.dm
```diff
@@ -84,6 +84,9 @@
 		stack_trace("Warning: [src]([type]) initialized multiple times!")
 	initialized = TRUE
 
+	if(loc)
+		SEND_SIGNAL(loc, COMSIG_ATOM_CREATED, src) // Sends a signal that the new atom `src`, has been created at `loc`.
```
farie82: The atom will most likely not be fully initialised here. Usually Initialize is overriden and the parent is called prior to the other code. Best would be to send this signal after Initialize is fully done and thus in the proc above calling this one.

code/datums/holocall.dm
SteelSlayer: Oop yeah, I tried to be clever here but this will runtime if the pad or user are deleted.

code/game/atoms.dm
SteelSlayer: As far as I can tell the atom is spawned in the game world at this point, and my tests worked with having it here, but, I'm not sure if it will cause weird behavior. Guess I should put it at the end of the init proc, yeah.

Fox-McCloud:
I'm aware of what this does--but I want to see numbers, please.

If code increase overall performance by 0.001 deciseconds, but makes the code significantly more complex/unreadable/difficult to work with, that's just pointless optimization--which is why I'd like to see a performance comparison, please.

code/game/atoms.dm
farie82: An atom can override it like this:
image

Currently the signal will go through before Link_Abduction_Equipment is called. So the atom is not yet fully initialized.

On line 87 of subsystem/atoms.dm in `InitAtom` the `Initialize` is called. If you send the signal after that call then you'll be sure it's Initialized. Can even use some logic there (it has some checks to see if the init actually worked)

SteelSlayer:
I ran two one-hour rounds, one on master, one on this branch.

On master, the machines SS tended to hover around 38-39ms at idle.

master1

---

On this branch, the machines SS hovered around 29-31ms at idle.

refactor1

---

Here is the profile of `process_machines()`. Added some lables in there, the master branch is the first row, and the refactor is the second row.

E


If you want to see anything else, I have the full profile txt files from both rounds.

code/game/atoms.dm
SteelSlayer: Yeah that's a good idea, I'm thinking this is a good spot:

image

I believe if the code reaches this point, we know for sure the atom was initialized.

Fox-McCloud:
2.5% CPU savings isn't much at all--it's about what I expected--machine processing isn't that expensive to begin with (especially since lights were moved to the power channel years ago).

~0.00068065099 seconds of processing time gained roughly every 2 seconds.

I'll have to think about this and discuss this with some others--I'm not sure if this will be worth it or not.

For clarity, if this didn't really come with any increase in code complexity or readability, this would be a non-issue--stuff like istypeless loops or going with defines over procs to save CPU are always appreciated, even when they save 1% or less of CPU---it's functionality loss/alteration, code readability, and increases in code complexity are a factor that we have to start thinking "is this really worth it?"

SteelSlayer:
How did you get 2.5%? I'm curious because I was looking at the difference between self CPU and the ms time of the subsystems, which seems to show a ~20% improvement there. But this is basically my first forte into the performance side of things so I'm not entirely sure what to look for yet.

Crazylemon64:
Let's see, there's a few levels to parse this at
In the figure you've provided, without this refactor it appears that the machinery controller consumes ~14u of cpu time in the top-level loop (self cpu), out of 74u total spent actually doing processing in each machine (total cpu)

However, it also appears that with the refactor installed, the machinery controller consumes ~11u of cpu time at the top-level loop (self cpu), or 3 less units, for around 20% (1 - 11/14 = .21) less time in that loop. Unfortunately, the extra processing/logic used to tell whether to start processing or not seems to actually increase the total CPU time from 73.5u to 75u (2% slower)  - so something you're doing in these checks appears to be wasting the cycles saved, and actually making matters worse. (When you're looking at top-level loops like this, what matters more is the "total time", since that's the amount of time spent running code inside all functions called within)

SteelSlayer:
Ok, knowing that the total CPU time figure is the most important is useful info, I didn't know that before. So this basically means that as is, this PR is not worth it and actually makes the server take up more resources overall?

Also this is confusing to me because, in that round, I just sat there AFK for an hour and didn't turn on any of the machines I touched. So I'm confused as to why total CPU is higher here. It should have been lower because the number of machines processing was almost cut in half. They never even need to do those checks to start or stop.

Crazylemon64:
Was there any event that might have caused more complex processing to certain machines?

Crazylemon64:
Seeing which machines were most costly in `process` between the trials might also shed some light

SteelSlayer:
So I've isolated the machine `process()` procs from both of my profiles. From what I can see, it *looks* like it was actually other machines that caused the total CPU number to be higher. 

Even though `process_machines` ran an equal number of times in both of the rounds, certain process procs that I didn't even touch, like turrets and APCs for example, were using more total CPU. I can't explain this because all this PR does is make certain machines only process when turned on, or when they have an occupant etc. I didn't touch the logic of these other machine's `process()` loops.

Here is the txt file with the comparison if you want to check it out.
[Master-vs-ProcessRefactor.txt](https://github.com/ParadiseSS13/Paradise/files/4465709/Master-vs-ProcessRefactor.txt)

Crazylemon64:
It could be possible that a space ruin or away mission loaded in extra turrets, resulting in a higher load

If you could  do another go where you control for ruins, away missions, and events, that would probably result in a more accurate image

SteelSlayer:
Yeah that was a good idea. I ran two more test rounds with these configs and got results similar to what I expected:

```
## Disable the loading of away missions
DISABLE_AWAY_MISSIONS

## Disable the loading of space ruins
DISABLE_SPACE_RUINS

## Minimum number of space ruins levels to generate
EXTRA_SPACE_RUIN_LEVELS_MIN 0

## Maximum number of space ruins levels to generate
EXTRA_SPACE_RUIN_LEVELS_MAX 0
```


PVM

code/game/atoms.dm
farie82: Won't this code get called twice now?

SteelSlayer: Yeah that was me being dumb and forgetting to remove the previous one.

code/game/machinery/Sleeper.dm
```diff
@@ -81,39 +83,36 @@
...
+		filtering = 0
+		return
+
+	if(beaker.reagents.total_volume < beaker.reagents.maximum_volume)
+		occupant.transfer_blood_to(beaker, 1)
+		for(var/datum/reagent/x in occupant.reagents.reagent_list)
+			occupant.reagents.trans_to(beaker, 3)
+			occupant.transfer_blood_to(beaker, 1)
+
+	for(var/A in occupant.reagents.addiction_list)
```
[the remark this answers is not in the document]
SteelSlayer: You're right, will fix

code/game/machinery/machinery.dm
```diff
@@ -176,9 +211,9 @@ Class Procs:
 		use_power(7500/severity)
 		new /obj/effect/temp_visual/emp(loc)
 	..()
+
 /obj/machinery/default_welder_repair(mob/user, obj/item/I)
-	. = ..()
```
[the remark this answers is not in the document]
SteelSlayer: Trying to over optimize without realizing the return value is actually needed, gonna re-add it.

code/game/objects/structures/watercloset.dm
```diff
@@ -482,6 +487,7 @@
 
 /obj/machinery/shower/process()
 	if(!on || !mobpresent)
+		end_processing()
```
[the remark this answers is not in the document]
SteelSlayer: Yeah, another case of over optimization too, I'll remove it and just have the return there again.

code/game/machinery/Sleeper.dm
```diff
@@ -523,6 +527,7 @@
 		usr.stop_pulling()
 		usr.forceMove(src)
 		occupant = usr
+		begin_processing()
```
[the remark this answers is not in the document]
SteelSlayer: Good idea

code/game/machinery/cryopod.dm
```diff
@@ -689,6 +696,7 @@
 		to_chat(usr, "<span class='notice'>[on_enter_occupant_message]</span>")
 		to_chat(usr, "<span class='boldnotice'>If you ghost, log out or close your client now, your character will shortly be permanently removed from the round.</span>")
 		occupant = usr
+		begin_processing()
```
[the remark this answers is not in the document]
SteelSlayer: Side rant: I really hate these move in and move out verbs on these cryopods and sleepers and other things, I wish I could just remove them because I've never once used them.

SteelSlayer:
As Fox said, the increase in complexity probably isn't worth it for the smallish performance increase. Closing this because there's a bunch of conflicts, and I'm not sure if this is actually wanted. If it is, just tell me and I can re-open though.

PR #13237, opened 2020-04-01
Disables the syndicate outpost self destruct device

farie82:
## What Does This PR Do
Server is currently becoming unstable if the syndicate outpost blows up. This is due to the way it is designed with it having a ton of vendors with explosives and such.
This PR disables the bomb from being used for the time being

## Why It's Good For The Game
It stops the server from dying or running horribly. And this is a better alternative in my eyes than blacklisting the base

## Changelog
:cl:
del: The syndicate outpost self destruct device now doesn't work. Done due to the server becoming unstable after detonation
/:cl:

### DIFF
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -300,15 +300,23 @@
 	..()
 	wires.CutAll()
 
+// NOTICE: The bomb has been deactivated codewise due to it causing instabilities to the server in combination with the current syndicate base.
 /obj/machinery/syndicatebomb/self_destruct
 	name = "self destruct device"
-	desc = "Do not taunt. Warranty invalid if exposed to high temperature. Not suitable for agents under 3 years of age."
+	desc = "Do not taunt. Warranty invalid if exposed to high temperature. Not suitable for agents under 3 years of age. This bomb does not seem to be working!"
 	req_access = list(ACCESS_SYNDICATE)
 	payload = /obj/item/bombcore/large
 	can_unanchor = FALSE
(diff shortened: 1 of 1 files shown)


Carthusia:
Awww, no more "Forbidden" pizza's that lead to its destruction.

Perhaps instead of a Self destruct, it instead moves to "Self destruct" the loot containers? Still keeps the worry about activating the device, but doesnt blow the entire thing up, just the loot they would have obtained had they not tried to rambo it.

farie82:
> Awww, no more "Forbidden" pizza's that lead to its destruction.
> 
> Perhaps instead of a Self destruct, it instead moves to "Self destruct" the loot containers? Still keeps the worry about activating the device, but doesnt blow the entire thing up, just the loot they would have obtained had they not tried to rambo it.

This is more of a temporary fix to stop the instability. There is work being done to find the real cause of the issue and fixing it

lolzyking:
> Awww, no more "Forbidden" pizza's that lead to its destruction.
> 
> Perhaps instead of a Self destruct, it instead moves to "Self destruct" the loot containers? Still keeps the worry about activating the device, but doesnt blow the entire thing up, just the loot they would have obtained had they not tried to rambo it.

This is in regards to the Outpost not the supply depot, the supply depot is static in every round space and as far as I know not the problem, the problem is the lavaland syndicate outpost and its bomb.

Carthusia:
Ohhhhhhhhhhhh, ok i got them mixed up, my bad

Crazylemon64:
The ruin is already disabled - and what actually needs to be fixed is the syndicate bomb itself, rather than something that only targets the symptoms

farie82:
Gonna close it then since the ruin itself got disabled

PR #13244, opened 2020-04-02
Welders no longer spam eye flashes

Citinited:
**What does this PR do:**

Tweaks the change added in #11700 for welders. Instead of spamming `flash_eyes()` every second, instead now it just adds the flash overlay every 2.5 seconds (which essentially makes your screen white all the time while welding). It also won't give you more and more damage over time.

**Obligatory picture:**
image

:cl:
tweak: Welders now apply the flash overlay without a copious amount of eye damage.
/:cl:

### DIFF
--- a/code/game/objects/items/tools/welder.dm
+++ b/code/game/objects/items/tools/welder.dm
@@ -29,7 +29,7 @@
 	var/deactivation_sound = 'sound/items/welderdeactivate.ogg'
 	var/light_intensity = 2
 	var/low_fuel_changes_icon = TRUE//More than one icon_state due to low fuel?
-	var/progress_flash_divisor = 10 //Length of time between each "eye flash"
+	var/progress_flash_divisor = 25 //Length of time between each "eye flash".
 
 /obj/item/weldingtool/Initialize(mapload)
 	..()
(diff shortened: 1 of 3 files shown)


Darkmight9:
oh god yes please!

code/game/objects/items/tools/welder.dm
```diff
@@ -122,8 +122,8 @@
 /obj/item/weldingtool/tool_check_callback(mob/living/user, amount, datum/callback/extra_checks)
 	. = ..()
 	if(. && user)
-		if(progress_flash_divisor == 0)
-			user.flash_eyes(min(light_intensity, 1))
+		if(progress_flash_divisor <= 0)
+			user.add_flash_overlay(min(light_intensity, 1))
```
Crazylemon64, line 126: flash_eyes checked for eye protection before doing the flash - this doesn't

Crazylemon64:
The PR is OK design-wise, but it should flash only when eye protection is insufficient

code/game/objects/items/tools/welder.dm
farie82: He moved the protection check to add_flash_overlay and made flash_eyes call add_flash_overlay as a base.
So the protection check will be done but only the eye damage and other stuff flash_eyes might do will not be applied

code/modules/mob/living/living.dm
```diff
@@ -756,11 +756,15 @@
 	return "You can't fit into that vent."
 
 //called when the mob receives a bright flash
-/mob/living/proc/flash_eyes(intensity = 1, override_blindness_check = 0, affect_silicon = 0, visual = 0, type = /obj/screen/fullscreen/flash)
+/mob/living/proc/flash_eyes(intensity, override_blindness_check, affect_silicon, visual, type)
```
farie82, line 759: The default values here are gone. Meaning all calls depending on those will break

Fox-McCloud: Yeah, this needs to be addressed.

Fox-McCloud:
Please deconflict and resolve the issues, please. This is quite stale.

Fox-McCloud:
Nothing wrong with this PR, but closing due to staleness.

PR #13245, opened 2020-04-03
New disease: Carpcough

TDSSS:
## What Does This PR Do
Inspired by a prayer of a player afflicted with anxiety, this adds a new disease: Carpcough. Carpcough, like anxiety or beesease makes you cough up a mob, namely space carp. The cure is also significantly harder, making this a very deadly disease. There is a non-transmissable version to use for admins, as well.

## Why It's Good For The Game
Smiting someone with spawning space carps around themselves sounds hilarious. Also another deadly disease is fun. Might also encourage people to farm some fish for carp toxin.

## Changelog
:cl:
add: A new, deadly disease: Carpcough
/:cl:

### DIFF
--- a/code/datums/diseases/carpcough.dm
+++ b/code/datums/diseases/carpcough.dm
@@ -26,7 +26,7 @@
 					affected_mob.adjustBruteLoss(2)
 		if(4)
 			if(prob(10))
-				affected_mob.visible_message("<span class='danger'>[affected_mob] clutches his stomach.</span>", \
+				affected_mob.visible_message("<span class='danger'>[affected_mob] clutches [affected_mob.p_their()] stomach.</span>", \
 												"<span class='userdanger'>Your stomach hurts!</span>")
 			if(prob(5))
 				to_chat(affected_mob, "<span class='danger'>You feel something moving in your throat.</span>")
@@ -34,8 +34,7 @@
 				affected_mob.emote("cough")
 				affected_mob.visible_message("<span class='danger'>[affected_mob] coughs up an angry space carp!</span>", \
 													"<span class='userdanger'>You cough up an angry space carp!</span>")
-				new /mob/living/simple_animal/hostile/carp(affected_mob.loc)
-	return
+				new /mob/living/simple_animal/hostile/carp(get_turf(affected_mob))
 
 /datum/disease/carpcough/fake
 	name = "Localized Carpcough"
--- a/code/datums/diseases/carpcough.dm
+++ b/code/datums/diseases/carpcough.dm
@@ -31,6 +31,7 @@
 			if(prob(5))
 				to_chat(affected_mob, "<span class='danger'>You feel something moving in your throat.</span>")
 			if(prob(1))
+				affected_mob.emote("cough")
 				affected_mob.visible_message("<span class='danger'>[affected_mob] coughs up an angry space carp!</span>", \
 													"<span class='userdanger'>You cough up an angry space carp!</span>")
 				new /mob/living/simple_animal/hostile/carp(affected_mob.loc)
--- a/code/datums/diseases/carpcough.dm
+++ b/code/datums/diseases/carpcough.dm
@@ -1,0 +1,44 @@
+/datum/disease/carpcough
+	name = "Carpcough"
+	form = "Infection"
+	max_stages = 4
+	spread_text = "On contact"
+	spread_flags = CONTACT_GENERAL
+	cure_text = "Carpotoxin"
+	cures = list("carpotoxin")
+	agent = "Fishy Infection"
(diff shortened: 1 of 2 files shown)


code/datums/diseases/carpcough.dm
```diff
@@ -0,0 +1,44 @@
+/datum/disease/carpcough
+	name = "Carpcough"
+	form = "Infection"
+	max_stages = 4
+	spread_text = "On contact"
+	spread_flags = CONTACT_GENERAL
+	cure_text = "Carpotoxin"
```
Kyep, line 7: How are crew supposed to get this?

code/datums/diseases/carpcough.dm
```diff
@@ -0,0 +1,44 @@
...
+				to_chat(affected_mob, "<span class='danger'>Your stomach suddenly hurts a lot. It feels like there's something sharp in there.</span>")
+				if(prob(20))
+					affected_mob.adjustBruteLoss(2)
+		if(4)
+			if(prob(10))
+				affected_mob.visible_message("<span class='danger'>[affected_mob] clutches his stomach.</span>", \
+												"<span class='userdanger'>Your stomach hurts!</span>")
+			if(prob(5))
+				to_chat(affected_mob, "<span class='danger'>You feel something moving in your throat.</span>")
+			if(prob(1))
```
Kyep, line 33: This has no limit on the number of carp that can be spawned.

code/datums/diseases/carpcough.dm
TDSSS: At the moment, it is admin only as I thought it was too lethal for a level 7 biohazard. We could think about a more severe virus event that only includes actually deadly diseases. Would you be interested in something like that?

Kyep: I mean, how are crew supposed to get enough carpotoxin to treat this?

And yes, I'd like more dangerous disease events.

TDSSS: Ah. Carp, when butchered, produce meat that contains some carpotoxin. So worst case, you wait for it to reach stage 4, then kill the carp and butcher it. You might have to try several times if the cure roll fails.

Mitchs98: You can also produce a nigh infinite amount of baby carps from the critter vendor in a fish tank.

code/datums/diseases/carpcough.dm
TDSSS: That is correct, but keep in mind:
1) A 1% chance is not a lot and
2) The carp it spawns will probably either be killed or kill the host, thus stopping the spawning of further carp. Because you can't cough up carp when dead.

code/datums/diseases/carpcough.dm
TDSSS: Also what mitch said, see
>Might also encourage people to farm some fish for carp toxin.

farie82:
The virus will be 1/3 as effective as lycancoughy in spawning dogs. But lycancoughy has a limit. 
Adding one of these viruses into rotation is not a good idea I think. Not until the way viruses work is reworked to be more manageable and less RNG based.

BryanPara:
Wanted to point out carpotoxin can be produced from koi beans mutated off soy beans in botany which is a more stable and substantial source.

code/datums/diseases/carpcough.dm
```diff
@@ -0,0 +1,44 @@
...
+				affected_mob.visible_message("<span class='danger'>[affected_mob] coughs up an angry space carp!</span>", \
+													"<span class='userdanger'>You cough up an angry space carp!</span>")
+				new /mob/living/simple_animal/hostile/carp(affected_mob.loc)
+	return
```
farie82, line 37: non required return here

code/datums/diseases/carpcough.dm
```diff
@@ -0,0 +1,44 @@
...
+		if(4)
+			if(prob(10))
+				affected_mob.visible_message("<span class='danger'>[affected_mob] clutches his stomach.</span>", \
+												"<span class='userdanger'>Your stomach hurts!</span>")
+			if(prob(5))
+				to_chat(affected_mob, "<span class='danger'>You feel something moving in your throat.</span>")
+			if(prob(1))
+				affected_mob.visible_message("<span class='danger'>[affected_mob] coughs up an angry space carp!</span>", \
+													"<span class='userdanger'>You cough up an angry space carp!</span>")
+				new /mob/living/simple_animal/hostile/carp(affected_mob.loc)
```
farie82, line 36: Will spawn carps in say lockers but also in mechs if the victim coughs one up while he's in there. get_turf will prevent this

TDSSS:
Fair enough.

PR #13246, opened 2020-04-03
Major disease event Draft

TDSSS:
## What Does This PR Do
Since kyet mentioned he'd like more dangerous disease events, I threw together a rough idea of what it would look like, taking advantage of my carpcough PR as well (since this pr branches off of that one it can be merged instead of or in addition to it)
This probably needs a lot of discussion on what diseases should qualify as this major and the paramteres for advanced diseases (maybe make some premade ones to use too?), which is why this is a draft.
Also not really tested yet so might not even work.

Note that I want to keep the scope of this small so it can be quickly merged and doesn't spiral out into porting some other viro system.

## Why It's Good For The Game
Make diseases actually scary in day to day life, instead of just a flu or a cold.

## Changelog
:cl:
add: more dangerous disease event
/:cl:

### DIFF
--- a/code/datums/diseases/carpcough.dm
+++ b/code/datums/diseases/carpcough.dm
@@ -31,6 +31,7 @@
 			if(prob(5))
 				to_chat(affected_mob, "<span class='danger'>You feel something moving in your throat.</span>")
 			if(prob(1))
+				affected_mob.emote("cough")
 				affected_mob.visible_message("<span class='danger'>[affected_mob] coughs up an angry space carp!</span>", \
 													"<span class='userdanger'>You cough up an angry space carp!</span>")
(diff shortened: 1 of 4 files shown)


Ty-Omaha:
Should also expand on the superficial cures we have: https://github.com/ParadiseSS13/Paradise/blob/210f8badf474adbd9f710a5f22fc19bccf4516e7/code/datums/diseases/advance/advance.dm#L12

https://github.com/tgstation/tgstation/blob/6401ba57081efb3db2f39320eeefeb13c1be4d8c/code/datums/diseases/advance/advance.dm#L37

taukausanake:
I really do miss the usage of anti-bodies to find cures. You think it might be possible to combine the two types of cures? Instead of treating the person you test a sample of their virus and the result is the cure you need to distribute?

PR #13247, opened 2020-04-03
Makes logging based on ckeys. Adds dchat/LOOC/OOC logging

### LINKED ISSUE #13610: Lesser Ash Drake Reversion Seems To Cause Increasing Lag
**Issue Description**:
Every time I reverted to my humanoid form as a lesser ash drake I appeared to get increasing amounts of lag each time I did it, it may have also affected people in the near vicinity

To revert back to normal without any lag

**What happened instead**:
Each time I reverted it appeared to get more and more laggy as well as potentially impacting others

Lag is bad and being able to intentionally lag others is bad

1, Get lesser drake transformation
2. Transform back and forth a number of times
3. Experience progressively worse lag

No clue, but it happened to me during the round starting at 8:46 PM EST on the 13th of June

**Extra information**:

farie82:
## What Does This PR Do
1. Makes logging ckey based and there now is a global datum that keeps track of all logs.
2. Logs dchat
3. Few view changes to accompany the changes
4. Logs OOC chat
5. Logs LOOC chat

Mobs added will use their last_known_ckey value so admins can use the bodies of dead players to check the logs of the player as well. The ckey will always be displayed next to the mob

Once merged #14037 can be reverted again

Fixes #13610

## Why It's Good For The Game
1. Makes keeping track of logs in code way easier and makes logging dchat and what not easier.
2. Just handy for admins
3. Required for the new way of logging.
4. Same as 2
5. Same as 2

## Images of changes
image
New layout and the new log types

## Changelog
:cl:
add: Logs are now tracked in a global logging variable. Allowing easier logging
add: The logging view is updated to accompany the change above
add: Dchat is logged by the new system
add: OOC is logged by the new system
add: LOOC is logged by the new system
del: The add mobs via ckey button is removed from the logging view
/:cl:


code/__DEFINES/subsystems.dm
```diff
@@ -47,7 +47,7 @@
 
 #define INIT_ORDER_GARBAGE 19
 #define INIT_ORDER_DBCORE 18
-#define INIT_ORDER_BLACKBOX 17
+#define INIT_ORDER_LOGGING 17
```
farie82, line 50: Reused this one since it wasn't used anywhere

AffectedArc07: This will be used by the blackbox SS when I make it. Also if this is logging OOC stuff I recommend making it one of the first things to init if it needs to log OOC pre-round.

code/controllers/subsystem/logging.dm
```diff
@@ -0,0 +1,57 @@
+SUBSYSTEM_DEF(logging)
+	name = "Logging"
+	priority = INIT_ORDER_LOGGING
+	flags = SS_NO_FIRE
+	var/list/datum/log_record/logs	// Assoc list of assoc lists (ckey, (log_type, list/logs))
+
+/datum/controller/subsystem/logging/Initialize()
+	LAZYINITLIST(logs)
```
AffectedArc07, line 8: If LAZYINITLIST is just `if (!L) L = list()`, why not just initialise the list on line 5, then you don’t even need the SS initialiser.

code/__DEFINES/subsystems.dm
farie82: I've changed it back and made it go before the DBCORE

code/controllers/subsystem/logging.dm
farie82: Made this incase a log is made before the system is initliased. I'll change this though since it now inits as one of the earliest systems.

MRgaty:
I am sorry for your dog man

farie82:
> I am sorry for your dog man

Life is difficult in a local server

code/modules/mob/dead/observer/observer.dm
```diff
@@ -54,7 +54,6 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 	if(ismob(body))
 		T = get_turf(body)				//Where is the body located?
 		attack_log_old = body.attack_log_old	//preserve our attack logs by copying them to our ghost
```
Fox-McCloud, line 56: shouldn't this be a `Copy` function as well?

farie82: This is the old way of doing it. Ghosts get the reference of the attack logs of the mob they belong to. I've left the old way in for legacy sake but I'm willing to fully remove it if admins like the new system enough

PR #13255, opened 2020-04-04
Slaughter Demon Exploit Fix

Fox-McCloud:
Knew about this a long time ago---but hardly anyone abused it.

Seeing it abused a lot now though--shame on you guys.

:cl: Fox McCloud
fix: Fixes being able to determine slaughter demon location with medical devices
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/slaughter/bloodcrawl.dm
+++ b/code/game/gamemodes/miniantags/slaughter/bloodcrawl.dm
@@ -76,13 +76,17 @@
 				adjustToxLoss(-1000)
 
 				if(istype(src, /mob/living/simple_animal/slaughter)) //rason, do not want humans to get this
-
 					var/mob/living/simple_animal/slaughter/demon = src
 					demon.devoured++
 					to_chat(kidnapped, "<span class='userdanger'>You feel teeth sink into your flesh, and the--</span>")
 					kidnapped.adjustBruteLoss(1000)
 					kidnapped.forceMove(src)
 					demon.consumed_mobs.Add(kidnapped)
+					if(ishuman(kidnapped))
+						var/mob/living/carbon/human/H = kidnapped
+						if(H.w_uniform && istype(H.w_uniform, /obj/item/clothing/under))
+							var/obj/item/clothing/under/U = H.w_uniform
+							U.sensor_mode = SENSOR_OFF
 				else
 					kidnapped.ghostize()
 					qdel(kidnapped)


Carthusia:
Is that an exploit?

I always thought it was intentional because they ate someone with suit sensors on, and the body was still inside the monster, like a morph.

Pixelzilla:
Thank god. Its just really annoying to have an entire antag shut down by people running around using crew pinpointers, handheld monitors, gps and all that nonsense. They're spooky entities with spooky powers and shouldn't be so easily tracked. How can sensors track you when you're on another dimensional plane anyways? It cant even track you when you're off zlevels.

TDSSS:
What about morphs, they can be located like that as well.

Qwertytoforty:
I mean, demons do seem to be in another plane, but not when popping out of blood. Doesnt make sense for sensors to magically turn off.

Not to mention the antag doesn't instantly get shut down, considering how fast it can move in blood, that they can instantly sink into blood, and unless you have a gun you will die fighting one. Demons are still very strong, with or without people tracking them

datlo:
This is an exploit? I thought this was working as intended. :+1: I guess

Ty-Omaha:
>or a person geta stuck in a locker.

Why would suit sensors turn off if someone is in a locker? That makes no sense at all.

Carthusia:
> > or a person geta stuck in a locker.
> 
> Why would suit sensors turn off if someone is in a locker? That makes no sense at all.

I think that was a thing a while back, if you went into a locker, your suit sensors/tracking implants could not be picked up for some odd reason, i remember security rushing perma because we couldnt find the guy, whole place was fine, no hidden doors, then we checked the locker and BOOM, there the vampire was.

It got fixed

code/game/gamemodes/miniantags/slaughter/bloodcrawl.dm
```diff
@@ -76,13 +76,17 @@
...
 					demon.devoured++
 					to_chat(kidnapped, "<span class='userdanger'>You feel teeth sink into your flesh, and the--</span>")
 					kidnapped.adjustBruteLoss(1000)
 					kidnapped.forceMove(src)
 					demon.consumed_mobs.Add(kidnapped)
+					if(ishuman(kidnapped))
+						var/mob/living/carbon/human/H = kidnapped
+						if(H.w_uniform && istype(H.w_uniform, /obj/item/clothing/under))
+							var/obj/item/clothing/under/U = H.w_uniform
+							U.sensor_mode = SENSOR_OFF
```
farie82, line 89: Won't this break prisoner jumpsuits?

Qwertytoforty:
Delayed reply, but, you are right, it doesn't make sense that sensors fail when someone goes in a locker made of metal, so why do sensors go off when in a body made of flesh. And why do sensors fail in the demon, while GPS, borg tracking, and the nuclear disk pointer still works?

Ty-Omaha:
> so why do sensors go off when in a body made of flesh

Because it's a redspace being travelling through a blood dimension

>And why do sensors fail in the demon, while GPS, borg tracking, and the nuclear disk pointer still works?

Oversight and should befixed

farie82:
This sounds more like a tweak/balance change than a fix in all honesty. There is no code or such trying to hide them from the monitor (unless I'm missing something).

Mitchs98:
Kinda agree with Farie. The PR seems to just add a block of code that adds in the functionality. Definitely don't see how GPS and nuke disk tracking is an oversight and in need of a 'fix' either. Definitely going to be a rather large buff towards slaughter demons that I don't feel is entirely warranted.

PR #13257, opened 2020-04-05
Fixes MapMerge Requirements

Kyep:
:cl: Kyep
tweak: Fixed an issue where github contributors cannot install the dependencies for the 'mapmerge' tool, by updating the desired version some dependencies.
/:cl:

### DIFF
--- a/tools/mapmerge2/requirements.txt
+++ b/tools/mapmerge2/requirements.txt
@@ -1,3 +1,3 @@
-pygit2==0.27.2
+pygit2==1.0.1
 bidict==0.13.1
 Pillow==6.2.0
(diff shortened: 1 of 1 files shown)


AffectedArc07:
I requested this PR because the current version of pygit (0.27.2) no longer has binaries, which causes issues when trying to install mapmerge. This doesnt help when youre new to this and dont get how python or pip works, so seeing your entire command line fill with red error text isn't good.
image
The newer version of pygit doesnt have this issue

tools/mapmerge2/requirements.txt
```diff
@@ -1,3 +1,3 @@
-pygit2==0.27.2
+pygit2==1.0.1
 bidict==0.13.1
 Pillow==6.2.0
```
AffectedArc07, line 3: `pygit2` works, but could you also change `Pillow==6.2.0` to `Pillow==7.1.1`, as that works on the latest Python (3.8.2)

PR #13266, opened 2020-04-06
Ion storms are no longer announced by default

datlo:
Currently ion storms have a 33 percent chance of being announced to crew. 

Ion storms too often results in the captain resetting the law with zero conflicts with the AI, and the announcement doesnt help here, as this prevents the AI from making prepared plans on how best to apply the law (if their lawset allows hiding the ion law)

This PR sets ion storms from random events to no announcement by default. Admin triggered ion storms still lets admins choose to show an announcement when triggered.

:cl:
tweak: Ion storms are no longer announced to the station.
/:cl:

### DIFF
--- a/code/modules/events/ion_storm.dm
+++ b/code/modules/events/ion_storm.dm
@@ -31,15 +31,6 @@
 				to_chat(M, "<span class='danger'>[message] ...LAWS UPDATED</span>")
 				to_chat(M, "<br>")
 
-	for(var/mob/living/silicon/robot/R in GLOB.living_mob_list)
-		if(R.stat != DEAD && !R.connected_ai && !R.emagged)
-			var/message = generate_ion_law(ionMessage)
-			if(message)
-				R.add_ion_law(message)
-				to_chat(R, "<br>")
-				to_chat(R, "<span class='danger'>[message] ...LAWS UPDATED</span>")
-				to_chat(R, "<br>")
-
 	if(botEmagChance)
 		for(var/mob/living/simple_animal/bot/bot in GLOB.machines)
 			if(prob(botEmagChance))
--- a/code/modules/events/ion_storm.dm
+++ b/code/modules/events/ion_storm.dm
@@ -1,14 +1,15 @@
+#define ION_NOANNOUNCEMENT -1
 #define ION_RANDOM 0
 #define ION_ANNOUNCE 1
 
 /datum/event/ion_storm
 	var/botEmagChance = 10
-	var/announceEvent = ION_RANDOM // -1 means don't announce, 0 means have it randomly announce, 1 means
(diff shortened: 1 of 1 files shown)


code/modules/events/ion_storm.dm
```diff
@@ -19,18 +20,26 @@
 	if(announceEvent == ION_ANNOUNCE || (announceEvent == ION_RANDOM && prob(ionAnnounceChance)))
 		GLOB.event_announcement.Announce("Ion storm detected near the station. Please check all AI-controlled equipment for errors.", "Anomaly Alert", 'sound/AI/ionstorm.ogg')
 
-
 /datum/event/ion_storm/start()
 	//AI laws
 	for(var/mob/living/silicon/ai/M in GLOB.living_mob_list)
-		if(M.stat != 2 && M.see_in_dark != 0)
+		if(M.stat != DEAD && M.see_in_dark != FALSE)
```
farie82, line 26: What kind of check is this. see_in_dark? wot
Is this to check if they are carded or....
Nothing to do with your PR but just something I noticed

datlo: :man_shrugging:

Carthusia:
I like the idea of people not automatically being told this is happening, you have to pay attention.

Why are the borgs carrying bob around and praising him as Ali Din?

Why are the borgs stealing all the salt?

Why are the borgs pulling each other in a train, all saying choo choo?

It rewards paying attention, most of the time the laws are not super harmful and gives the AI and borg players a fun little thing to do until it's eventually discovered and fixed

code/modules/events/ion_storm.dm
```diff
@@ -19,18 +20,26 @@
...
 			var/message = generate_ion_law(ionMessage)
 			if(message)
 				M.add_ion_law(message)
 				to_chat(M, "<br>")
 				to_chat(M, "<span class='danger'>[message] ...LAWS UPDATED</span>")
 				to_chat(M, "<br>")
 
+	for(var/mob/living/silicon/robot/R in GLOB.living_mob_list)
+		if(R.stat != DEAD && !R.connected_ai && !R.emagged)
+			var/message = generate_ion_law(ionMessage)
```
Kyep, line 36: Objection.
This seems like it would give ERT borgs permanent, unremovable ion laws.
It also seems like it is buffing the ion storm event too much. Not just making it silent, but also making it affect unlinked borgs.

datlo: True I didnt think of that. I'll just remove this part of the PR as in most situations a cyborg will have a linked AI and this could cause too many issues/edge cases with stuff like ERT borgs... maybe even syndicate ones now that I think about it.

Updating now.

TDSSS:
Aww, I liked unlinked borgs getting a law, too. The borg law upload console never gets used, couldn't we just exclude ERT and other special borgs from it?

PR #13267, opened 2020-04-06
Fix reset/purge boards not clearing zeroth laws

### LINKED ISSUE #13264: Purge board doesn't remove the "One crewmember" law
**Issue Description**:
The "One crewmember" law cannot be removed by normal means, not even the purge board.

For the purge board to remove the "One crewmember" law 0.

**What happened instead**:
It didn't.

This law cannot be removed by normal means, making subversions using it way harder to deal with.

1. Upload the "One crewmember" law to the AI.
2. Use a purge module on the AI.
3. Check the AIs laws.

Uncertain, a long time ago likely.

**Extra information**:
I think this is due to the purge board code just setting the Law 0 to an empty string, which as far as I understand it, counts as null.
https://github.com/ParadiseSS13/Paradise/blob/210f8badf474adbd9f710a5f22fc19bccf4516e7/code/game/objects/items/weapons/AI_modules.dm#L241
Setting the zeroth law immediately returns if the law is null.
https://github.com/ParadiseSS13/Paradise/blob/04ba5c1cc902bf2eada751de7b1b72ebc64ecd86/code/datums/ai_laws.dm#L118
I wonder why the following proc isn't used instead.
https://github.com/ParadiseSS13/Paradise/blob/04ba5c1cc902bf2eada751de7b1b72ebc64ecd86/code/datums/ai_laws.dm#L236

datlo:
Reset and purge boards now clear zeroth laws of non-antag AIs as they are meant to do.

Also boldens some law notifications to AI that dont have a span and are easy to miss

Fix #13264

:cl:
fix: Fixed reset and purge boards not clearing zeroth laws of non-antag AIs, as they are meant to do.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/AI_modules.dm
+++ b/code/game/objects/items/weapons/AI_modules.dm
@@ -143,9 +143,8 @@ AI MODULES
 		target.set_zeroth_law(law)
 		GLOB.lawchanges.Add("The law specified [targetName]")
 	else
-		to_chat(target, "<b>[sender.real_name] attempted to modify your zeroth law.</b>")// And lets them know that someone tried. --NeoFite
-
-		to_chat(target, "<b>It would be in your best interest to play along with [sender.real_name] that [law]</b>")
+		to_chat(target, "<span class='bnotice'>[sender.real_name] attempted to modify your zeroth law.</span>")// And lets them know that someone tried. --NeoFite
+		to_chat(target, "<span class='bnotice'>It would be in your best interest to play along with [sender.real_name] that [law]</span>")
 		GLOB.lawchanges.Add("The law specified [targetName], but the AI's existing law 0 cannot be overridden.")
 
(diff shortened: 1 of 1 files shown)


datlo:
Updated to spans

BryanPara:
I have no problem with the Purge board getting rid of a zero law however I don't think the reset board should do so as well.

If the reset board is able to get rid of it, then really its no incentive to use the purge board except if one is an antagonist, which even then would be quite rare. I also feel as though removing a law that would be listed as inherent should be more risky and temporarily removing the AI's shackles to do so would at least provide more than simply popping in a reset board.

datlo:
The reset board is intended to remove zeroth law, it's just bugged and not doing so. The only case where removing zeroth law even matters is when OneCrew is used, as malf AI zeroth law cannot be removed by the AI upload. There is no concern for balance here.

PR #13268, opened 2020-04-07
Bonfire Logs

AffectedArc07:
## What Does This PR Do
Requested by Necaladun. This PR adds in logging for when someone lights a bonfire, as well as when people are burned by it

image
image
image

It also admin logs immediately when its crafted
image

## Why It's Good For The Game
Admins should know when people light people up

## Changelog
:cl: AffectedArc07
add: Bonfire lighting and burn events are now logged
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/towercap.dm
+++ b/code/modules/hydroponics/grown/towercap.dm
@@ -165,7 +165,7 @@
 
 /obj/structure/bonfire/Crossed(atom/movable/AM, oldloc)
 	if(burning)
-		add_attack_logs(src, AM, "burned by a bonfire (Lit by [lighter])", ATKLOG_FEW)
+		add_attack_logs(src, AM, "burned by a bonfire (Lit by [lighter])", ATKLOG_ALMOSTALL)
 		Burn()
 
 /obj/structure/bonfire/proc/Burn()
--- a/code/modules/hydroponics/grown/towercap.dm
+++ b/code/modules/hydroponics/grown/towercap.dm
@@ -110,6 +110,7 @@
 	anchored = TRUE
 	buckle_lying = FALSE
 	var/burning = 0
+	var/lighter // Who lit the fucking thing
 	var/fire_stack_strength = 5
 
 /obj/structure/bonfire/dense
@@ -125,6 +126,8 @@
 		var/image/U = image(icon='icons/obj/hydroponics/equipment.dmi',icon_state="bonfire_rod",pixel_y=16)
 		underlays += U
 	if(is_hot(W))
+		lighter = user.ckey
+		user.create_log(MISC_LOG, "lit a bonfire", src)
 		StartBurning()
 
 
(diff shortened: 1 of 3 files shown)


code/modules/hydroponics/grown/towercap.dm
```diff
@@ -162,6 +165,7 @@
 
 /obj/structure/bonfire/Crossed(atom/movable/AM, oldloc)
 	if(burning)
+		add_attack_logs(src, AM, "burned by a bonfire (Lit by [lighter])", ATKLOG_FEW)
```
farie82, line 168: `ATKLOG_FEW` means admins with Few logs turned on (aka they see few logs) get them. You prob want `ATKLOG_ALMOSTALL` for this one.

Kyep: Definitely not ATKLOG_FEW. That's reserved for things like attacks on SSD players, bombs, etc which are very possibly rule violations.
I'd probably go with ATKLOG_MOST. The major difference between ATKLOG_ALMOSTALL and ATKLOG_MOST is that ATKLOG_MOST hides NPC v player combat. You could argue that a bonfire should be treated like an NPC (its player-created, but not player-controlled).
(semi-joking) perhaps creating bonfires should be logged at a higher level than the burns they inflict, though. Perhaps creating bonfires in key areas, like the escape shuttle, or the Escape room, should even be logged as ATKLOG_FEW, since its likely griefing and will affect many players.

code/modules/hydroponics/grown/towercap.dm
```diff
@@ -162,6 +165,7 @@
 
 /obj/structure/bonfire/Crossed(atom/movable/AM, oldloc)
 	if(burning)
+		add_attack_logs(src, AM, "burned by a bonfire (Lit by [lighter])", ATKLOG_ALMOSTALL)
```
farie82, line 168: Checking `add_attack_logs` the proc: This will send a message to admins (who have the almostALL logging setting) even when say a ckeyless mob gets burned. Or just a locker.
Might want to do an istype check here.... that or rework `create_attack_logs` since that one also needs a rework

Denthamos:
Mmn, like Kyet and and was discussed before.  Creation of the bonfire is more important than the lighting of such.  If we know where it is being constructed and by whom, we can handle it then before it is even lighted.

code/modules/hydroponics/grown/towercap.dm
Crazylemon64: Seconded, the logging level should depend on whether what is being burned is a mob/player or not

AffectedArc07:
image
image

```dm
/obj/structure/bonfire/Crossed(atom/movable/AM, oldloc)
	if(burning)
		Burn()
		if(ishuman(AM))
			var/mob/living/carbon/human/H = AM
			add_attack_logs(src, H, "Burned by a bonfire (Lit by [lighter])", ATKLOG_ALMOSTALL)

```

Logs are now only created if the person being burned is a human

PR #13278, opened 2020-04-10
Makes solars auto-track again

AffectedArc07:
## What Does This PR Do
This PR fixes the issue of the tcomms solar tracker not auto-tracking the solars, and the AI having to either set them, or forget about it. The issue was the order of initializations, as visualised below
image

Now the solars initialize as part of the Sun subsystem (handles solar rotation and tracking, so I guess it makes sense)

## Why It's Good For The Game
The solars which have the auto-track flag in the code, should actually auto track instead of just being broken, redundant code.

## Changelog
:cl:
fix: Telecoms solars now auto-track again
/:cl:


AffectedArc07:
Well then, turns out this has no open issues for it.

Spacemanspark:
Thank fuck

code/modules/power/solar.dm
```diff
@@ -285,13 +285,14 @@
 	autostart = 1 // Automatically start
 
 /obj/machinery/power/solar_control/Initialize()
-	..()
-	if(!powernet)
-		return
+	SSsun.solars |= src
+	. = ..()
+
+/obj/machinery/power/solar_control/proc/setup()
```
Crazylemon64, line 291: How would this work for maploaded solars? I don't see a mechanism for calling setup outside of the sun subsystem startup

AffectedArc07: My bad

PR #13282, opened 2020-04-11
Remove crowbarring sound on blast doors and update the proc

datlo:
Remove crowbarring sound on blast doors, whether succesful or not.  Succesfully forcing a blast door open already plays the 'shutters opening' sound.

Trying to force open a powered blast doors (or shutters) will now display a message saying that the door resists your effort to force it.

Forcing open an unpowered blast doors or shutters will now take 5 seconds with a crowbar, or 1.25 seconds with a fire axe or jaws of life.

Fire axe toolspeed is now 0.25 to make it open the blast doors/shutters faster, was not used before this PR.

:cl:
tweak: Blast doors no longer make a sound when crowbarred.
tweak: Forcing open an unpowered blast doors now takes 5 seconds, faster with a fireaxe/power tool.
/:cl:

### DIFF
--- a/code/game/machinery/doors/poddoor.dm
+++ b/code/game/machinery/doors/poddoor.dm
@@ -59,16 +59,6 @@
 /obj/machinery/door/poddoor/try_to_activate_door(mob/user)
  	return
 
-/obj/machinery/door/poddoor/crowbar_act(mob/user, obj/item/I)
-	if(user.a_intent == INTENT_HARM)
-		return
-	. = TRUE
-	if(operating)
-		return
-	if(!I.use_tool(src, user, 0, volume = 0))
-		return
-	try_to_crowbar(user, I)
-
 /obj/machinery/door/poddoor/try_to_crowbar(mob/user, obj/item/I)
 	if(!density)
 		return
--- a/code/game/machinery/doors/poddoor.dm
+++ b/code/game/machinery/doors/poddoor.dm
@@ -59,9 +59,28 @@
 /obj/machinery/door/poddoor/try_to_activate_door(mob/user)
  	return
 
-/obj/machinery/door/poddoor/try_to_crowbar(obj/item/I, mob/user)
+/obj/machinery/door/poddoor/crowbar_act(mob/user, obj/item/I)
+	if(user.a_intent == INTENT_HARM)
+		return
+	. = TRUE
+	if(operating)
+		return
+	if(!I.use_tool(src, user, 0, volume = 0))
+		return
+	try_to_crowbar(user, I)
+
+/obj/machinery/door/poddoor/try_to_crowbar(mob/user, obj/item/I)
+	if(!density)
+		return
 	if(!hasPower())
-		open()
+		to_chat(user, "<span class='notice'>You start forcing [src] open...</span>")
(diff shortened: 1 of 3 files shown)


datlo:
Updated 'pod door' to blast doors because I forgot the type is pod door for blast doors for some reason

code/game/machinery/doors/poddoor.dm
```diff
@@ -59,9 +59,28 @@
...
  	return
 
-/obj/machinery/door/poddoor/try_to_crowbar(obj/item/I, mob/user)
+/obj/machinery/door/poddoor/crowbar_act(mob/user, obj/item/I)
+	if(user.a_intent == INTENT_HARM)
+		return
+	. = TRUE
+	if(operating)
+		return
+	if(!I.use_tool(src, user, 0, volume = 0))
```
Crazylemon64, line 68: The entire purpose of this redefinition of this proc is to make sure the tool doesn't play a sound here, right? Could you instead define a variable like `play_tool_sound_on_force` on `/obj/machinery/door/poddoor/crowbar_act` that would control whether to use the tool volume or 0 for the `use_tool` volume?

datlo: Now that I think bout it, I dont think any doors played the crowbar sound on being forced before the tool refactor

Ill just change the original door proc to not make a sound

Crazylemon64: Fine by me

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -179,6 +179,7 @@
 	slot_flags = SLOT_BACK
 	force_unwielded = 5
 	force_wielded = 24
+	toolspeed = 0.25
```
Kyep, line 182: Isn't this a general buff to fireaxes?

datlo: No, toolspeed isnt used by fireaxes anywhere before this pr.

code/game/machinery/doors/poddoor.dm
```diff
@@ -59,9 +59,18 @@
...
  	return
 
-/obj/machinery/door/poddoor/try_to_crowbar(obj/item/I, mob/user)
+/obj/machinery/door/poddoor/try_to_crowbar(mob/user, obj/item/I)
+	if(!density)
+		return
 	if(!hasPower())
-		open()
+		to_chat(user, "<span class='notice'>You start forcing [src] open...</span>")
+		if(do_after(user, 50 * I.toolspeed, target = src))
```
farie82, line 67: can put hasPower in as an extra check for the do_after. This will stop the do_after from working if the door regains power. Can also remove the line below then

datlo: If I do this it will remove user feedback on the fact the pod door is powered again, which is why the forcing open failed. Or I will need to check if door is powered to show a message or not which defeats the point of the change... Keeping it as is.

farie82: You can still have the feedback. Just move it down into the else of the `do_after`. That way the user won't do work while it won't result in anything and the user is notified on why it failed

PR #13286, opened 2020-04-12
Nanopaste resizeing (w_class)

hunakos57:
## What Does This PR Do

Resizes the Nanopaste to tiny size to match the organic equivalent of it, like the trauma and burn kit. 
## Why It's Good For The Game

Before this resizing, the Nanopaste couldn't be stored in one's pocket or in any medkit box, making it quite inconvenient to carry or to store. This small change is to fix this, may it make this item more commonly used among the crew due to not taking up space as 50 metal sheet would.
## Changelog
:cl:
tweak: Modified the 'size' of the nano paste to tiny
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/nanopaste.dm
+++ b/code/game/objects/items/stacks/nanopaste.dm
@@ -4,6 +4,7 @@
 	desc = "A tube of paste containing swarms of repair nanites. Very effective in repairing robotic machinery."
 	icon = 'icons/obj/nanopaste.dmi'
 	icon_state = "tube"
+	w_class = WEIGHT_CLASS_TINY
 	origin_tech = "materials=2;engineering=3"
 	amount = 6
 	max_amount = 6


Carthusia:
Huzzah, ive always found it odd how this tiny tube of nanites is the size of a full sized medkit holding 6-7 medical items

lolzyking:
Probably beyond the scope of this PR, but I think nanopaste should work like a brute/burn patch on ipc and not spend a whole minute playing target the damage.

farie82:
Not sure if tiny would be fitting here due to the potency of the item in question. It is an instant heal of 15 brute and burn for robotic parts that can be used multiple times.

Carthusia:
It is RnD produced and only viable for synthetics. So it kind of balances out, compared to organics who make up the majority of the crew at any given time, get non-RnD healing patches of 2 types, that can heal more than 20-30 damage per usage instantly, are of tiny size, and stack at round start, and can make synthflesh patches from chemistry within the first 5 minutes of the round that heal both burn and brute of the same amount.

That's not even counting other near instant heal methods organics get that are not RnD locked.

This being tiny seems on par with the organics items.

PR #13287, opened 2020-04-12
Travis cleanup

AffectedArc07:
## What Does This PR Do
This PR cleans up our current Travis pipelines, adds in a check I forgot about, and makes the file less of a mess. All dependencies are now in their own singular file instead of all over the place, and some job combination has been done.

NanoUI building has been merged into the `File Checks` stage, as this drastically reduces our build time. All map building has also been merged into the same check because I see no point of keeping the checks separate when it accomplishes the same thing.

The file checks stage also now verifies PHP and JSON files, to ensure they arent malformed with invalid syntax.

Old Travis completion time:
image

New Travis completion time:
image

This PR will build on the new Travis pipelines, so you can look at it and see the changes and improvements.

## Why It's Good For The Game
Our current Travis system takes a full 10 minutes to run, which is not good for crucial security PRs which need to be speedmerged. This PR increases the build speed by a full three times, so that in itself is justification of why this is good.

## Changelog
No changelog because this doesnt touch the game


AffectedArc07:
Oh and for those wondering why I have so many commits, you cant test travis locally properly

AffectedArc07:
Most recent commit is to fix these things which Travis complained at 
image

_build_dependencies.sh
```diff
@@ -1,2 +1,11 @@
...
+# For dreamchecker
 export SPACEMANDMM_TAG=suite-1.2
+# For NanoUI
+export NODE_VERSION=9
+# For the scripts in tools
+export PHP_VERSION=5.6
+# Byond Major
+export BYOND_MAJOR=513
+# Byond Minor
+export BYOND_MINOR=1505
```
farie82, line 11: Hmm annoying to have to update these using PRs everytime. Out of the scope of the PR though

AffectedArc07: Still easier than the old way which is finding the vars in the `.travis.yml` file. This is now actually properly laid out.

AffectedArc07:
The random EOF newlines are because some files got changed since which removed the trailing line. When I standardised them all, I forgot to setup the grep script in travis. Oops.

PR #13291, opened 2020-04-12
Replace traitor martyr objective by survive objective

datlo:
This PR replaces the traitor 'Die a glorious death' objective by 'Survive until the end' objective.

Die is a glorious death is a fairly unfullfilling objective on Paradise. You cant 'Kill until killed' like you can on LRP servers, and just killing yourself for greentext is meh. I can only remember a single round where I had fun thanks to it in my many hours of playing.

This PR replaces the objective by 'Survive until the end', which serves the same purpose of being an easier objective than escape that can help a newbie pull off a greentext, and can also lead to more interesting moments in security. (Traitor caught by security after the targets are killed. You may have perma'd me, but the deed is done.)

Having this objective also encourages traitors to take it lighter on security, as they risk execution/field execution if they take it too far. Lord knows the redshirts can use the help.

:cl:
tweak: The 'Die a glorious death' objective for traitors has been replaced by 'Survive until the end.'
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -105,7 +105,7 @@
 
 /datum/antagonist/traitor/proc/forge_human_objectives()
 	var/is_hijacker = prob(10)
-	var/martyr_chance = prob(20)
+	var/survive_chance = prob(20)
 	var/objective_count = is_hijacker 			//Hijacking counts towards number of objectives
 	if(!SSticker.mode.exchange_blue && SSticker.mode.traitors.len >= 8) 	//Set up an exchange if there are enough traitors
 		if(!SSticker.mode.exchange_red)
@@ -129,16 +129,16 @@
 	for(var/i = objective_count, i < objective_amount)
 		i += forge_single_objective()
 
-	var/martyr_compatibility = 1 //You can't succeed in stealing if you're dead.
+	var/survive_compatibility = TRUE //There isnt much point in stealing items if you dont deliver them to your employers.
 	for(var/datum/objective/O in owner.objectives)
 		if(!O.martyr_compatible)
-			martyr_compatibility = 0
+			survive_compatibility = FALSE
 			break
 
-	if(martyr_compatibility && martyr_chance)
-		var/datum/objective/die/martyr_objective = new
-		martyr_objective.owner = owner
-		add_objective(martyr_objective)
+	if(survive_compatibility && survive_chance)
+		var/datum/objective/survive/survive_objective = new
+		survive_objective.owner = owner
+		add_objective(survive_objective)
 		return
 
 	if(!(locate(/datum/objective/escape) in objectives))


Carthusia:
Ive never really understood the whole "Kill yourself" objectives to the point I ignore them if I do get one.

There are suicide missions where your paid VERY well if you CAN survive, but a job where your entire objective IS to die?

Fox-McCloud:
This is arguably even more boring.

Carthusia:
> This is arguably even more boring.
> 
> It also encourages security to execute.

Hmmm, the real question is then: 

What do we replace it with? 

We have steal objectives, kill objectives.

Proper sabotage objectives? Destroy/Disable (blank) station component, like say, destroy the gravity generator, destroy the stations labor shuttle, things that cant be replaced.
Just an idea

VeteranKyl:
What if it's changed to something like "Escape to a location not owned by Nanotrasen for extraction before the end of the shift." This could include places like Lavaland, the various space derelicts, or the gateway. It goes with the idea that the traitor will be picked up by the syndicate once the shift is over, but must be in a location where Nanotrasen isn't present.

Carthusia:
> What if it's changed to something like "Escape to a location not owned by Nanotrasen for extraction before the end of the shift." This could include places like Lavaland, the various space derelicts, or the gateway. It goes with the idea that the traitor will be picked up by the syndicate once the shift is over, but must be in a location where Nanotrasen isn't present.

I like, the problem with space derelicts though is how mostly impossible they are to locate without a LONG time spent searching.

I feel players would spend more time searching for their extraction point than their objectives.

Unless the derelicts and locations to find in space had GPS markers like they do in lavaland, then this would be a non-issue.

VeteranKyl:
> I like, the problem with space derelicts though is how mostly impossible they are to locate without a LONG time spent searching.
>
>I feel players would spend more time searching for their extraction point than their objectives.
>
> Unless the derelicts and locations to find in space had GPS markers like they do in lavaland, then this would be a non-issue.

Oh, I had it in mind that the objective would not have a specific location the player had to be to green text. They just had to be either at a completely different location like the lavaland or gateway, or in space on a Z level where there are no NT structures. So even if you're in an empty Z level, that'd still count as escaping the station. I see it like the traitor has a transponder with them to let their syndicate buddies know where they are once the shift is over, but using it near a Nanotrasen controlled location could alert NT.

DarthDangermouse:
I'm not a fan of this one, that'll take the fun out of the more crazy suicide plays, this very much makes it much more boring, and as Fox would put it, Sec would have no reason just to execute rather than Perma, other than Space Law, which hardly any of them follow half the time anyways.

Aquilar:
why not both?

daylight2:
> 
> 
> This is arguably even more boring.
> 
> It also encourages security to execute.

If one of the reasons security puts people into perma instead of executing them is to avoid the chance of giving them greentext, then something is seriously wrong in the security mindset.



> 
> 
> I'm not a fan of this one, that'll take the fun out of the more crazy suicide plays, this very much makes it much more boring, and as Fox would put it, Sec would have no reason just to execute rather than Perma, other than Space Law, which hardly any of them follow half the time anyways.

Again, if the only reason for security to exists is to deny greentext at the cost of everything else, there's a problem. And, currently, there are very few things you could get away with using the 'die a glorious death' objective that you couldn't do as a non-antag. Both would get bwoinked for suicide bombing the bridge, neither would get bwoinked for using the guillotine whilst surrounded by the bridgewatchers.

MRgaty:
Like someone pointed out in discord, it incentives the "accomplishing objective and hiding for the rest of the shift". I dont like this

daylight2:
> 
> 
> Like someone pointed out in discord, it incentives the "accomplishing objective and hiding for the rest of the shift". I dont like this

As opposed to "Accomplishing objective and then typing in suicide"? or perhaps the "Accomplishing objective and hiding for the rest of the shift, and then suiciding"? At least with survive, the antag has the opportunity to get caught.
There's also the option of asking admins/syndicate for more objectives - I often do this when I accomplish my objectives early, but glorious death only matches with kill and protect objectives, which get a bit old after a few repetitions.

taukausanake:
There was talk about a drop off objective one time I believe. How about instead of survive until the end of the round you get an early escape that the station learns of and must prevent? A small syndicate runner ship could come to the station and pick up the traitor but CentCom learns of it and alerts the station. You're not commiting suicide or blowing yourself and a big hole in the station, but you still get to end early if you want. Maybe even let the traitor decide how they want to escape. Either stay quiet until the end (like normal) or go out semi-loud in an escape sequence

procdrone:
Antag endgame is always boring. After objectives are acomplished you either hide, are in perma, or are dead. There is nothing you can really do, and in most cases you just hide around in space untill shift end to board one of the pods. 

It is very boring. It removes you from the round and just forces you to linger around.

I like the extraction idea. but it would need testmerge to see how it works.
(one of the ideas is antag highjacking an escape pod, station docking computer would announce the pod is being launched without authorisation and it would take some small time. not enough for full on siege but enough for someone to react and maybe stop it)

I also think that such antags that complete their job should have a way to reenter the round. New character perhaps? Or maybe some other thing? Being a ghost already allows you to drone, npc, or mob spawn. Not to mention midrounds or ERT/event characters.

Maybe more ghost roles?

Rurik123:
I'm very late to this discussion but this should have been done a while ago. 

This lol xd kill yourself objective literally only exists to confuse newbies, who assume it means going out in a blaze of glory. Cause ya know, "Die a glorious death" is ambiguous as fuck and I'd wager most of mankind assumes it would mean dying in mass combat, making a spectacle. I know I did. 

If we can't remove it, could we at least have it renamed to "Type *suicide" so new players don't fall for it thinking they could do something cool, at least?

Spacemanspark:
Slightly off topic, but if we want a more interesting antagonist end game, why not a more challenging 'protect' objective?

"Protect x and ensure they escape the station alive at the end of the shift, either via a pod or the shuttle."
Or perhaps a harder variant that only allows them to escape on the shuttle.

Maybe also have it so at least one person has the objective to kill X, although that might get complicated.

datlo:
Git is a fucking massive PITA and I can't get it to work. I'm done for now.

PR #13293, opened 2020-04-13
Zeroth law is now above ion laws

datlo:
In the last community meeting, admins explained that a Malf AI antag status take precedence over any hacked/ion laws, and is thus free to delete/ignore those laws.

However, this causes confusion, as server rules are written bind you to follow those hacked/ion laws since they are above your law 0 : Accomplish your objectives at all costs.

This solves the issue by putting the law 0 above everything else, obviously for both cyborgs and AIs.

Law notice/Stating laws :

image

Admin law manager : 

image

Malf AI law manager : 

image

:cl:
tweak: Zeroth laws are now above ion and hacked laws in law order.
/:cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -62,14 +62,14 @@
 	if(sorted_laws.len)
 		return
 
-	for(var/ion_law in ion_laws)
-		sorted_laws += ion_law
-		
-	for(var/evil_law in devil_laws)
-		sorted_laws += evil_law
-
 	if(zeroth_law)
 		sorted_laws += zeroth_law
+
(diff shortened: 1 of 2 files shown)


Aurorablade:
I disagree with this as a ION law is more a software glitch in the system. It seems like the wording on the 0th law needs to be reworded if for some reason its causeing confusing. I don't see why its causing confusion. EVEN with the change and how you assume someone can ignore it, people may still just consider it a objective anyway if that is the issue. This seems like a player issue in the long run.

datlo:
Antag status overrides any laws, ion or uploaded, that the AI gets. This PR reflects currently existing admin policy.

PR #13295, opened 2020-04-13
Unathi 5% brute damage taken (Removed)

hunakos57:
## What Does This PR Do

Removes hidden damage taken multiplier from the Unathi races which were not documented on the wiki.
## Why It's Good For The Game

A near 2m tall dino is not wimps.

## Changelog
:cl:
tweak: 1.05 damage taken multiplier removed
/:cl:

(Why was this added in the first place?)

### DIFF
--- a/code/modules/mob/living/carbon/human/species/unathi.dm
+++ b/code/modules/mob/living/carbon/human/species/unathi.dm
@@ -36,7 +36,6 @@
 	default_headacc = "Simple"
 	default_headacc_colour = "#404040"
 	butt_sprite = "unathi"
-	brute_mod = 1.05
 
 	has_organ = list(
 		"heart" =    /obj/item/organ/internal/heart/unathi,


Spacemanspark:
When was this added...?

Qwertytoforty:
2 years ago, according to git blame

TDSSS:
Apparently it was added as compensation for adding the tail whip attack. Might need to give them another slight downside.

Carthusia:
Hungrier faster due to their large size and requiring more food to keep the body functioning?

hunakos57:
Hunger is a common one to give as a con, but speed debuff after the use of tailspin could be nice, like stamina damage maybe caused by it to the user.

Mitchs98:
I see no reason they need any sort of debuff at all for tail whip. It barely does anything, it's literally just a meme. I don't understand why the initial debuff was added, let alone why they need any others, personally.

Carthusia:
For me, its more of a fluff kind of thing, their a pretty large species, thus they would require more food to function.

Plazma Men are about the same height, though they are plasma ridden skeletons, not much "meat" needing nutrients

Ty-Omaha:
I've updated the wiki to reflect this modifier. Just a reminder that the wiki is a community project and anyone can contribute to it if they see anything missing

MRgaty:
I domt see a reasom to remove this, modifying the wiki would be easier

hunakos57:
Looks like no need to remove it.

PR #13303, opened 2020-04-14
Fixes certain worn items being impossible to strip off of IPCs if they weren't wearing a jumpsuit

### LINKED ISSUE #13262: IPCs can hide and prevent removal of items in their pockets
**Issue Description**:
IPCs who have items in their pockets but are not wearing a jumpsuit cannot have these items removed or seen by someone searching them.

I expected to be able to remove items from an IPC on the ground after examining them and seeing that they had their ID with them and full pockets.

**What happened instead**:
The search UI wouldn't list any items in the IPCs pockets or their ID or PDA as it gets hidden when no jumpsuit is being worn, despite having their ID it was impossible to retrieve.

An IPC playing as an antag such as a syndicate would be able to store bombs for example in their pockets which security would have no way of knowing were there and would be unable to retrieve said bombs if they did. The syndicate if then perma brigged could bomb the brig without security ever knowing they had bombs, other controband such as emags and Eswords would also be hidden from them. It massively unbalances IPC syndicates in relation to security as they have no way of dealing with them. 

Get a naked IPC to equip their ID and items into their pockets, then have another character try to search the IPC, items will be hidden in their pockets and ID slot.

Noticed this when I tried to get the ID of a robotics partner to unlock their mechs. Also explained how sometimes when I played as a naked IPC, security wouldn't empty my pockets which had tools in when they put me in the brig.

**Extra information**:
I think it's very lucky that this hasn't been discovered by a wide range of people who do play IPC because it would be abused alot.

SteelSlayer:
## What Does This PR Do
IPC could equip items into their belt, ID, PDA, and pocket slots without having to wear a jumpsuit. But, when people tried to stripped them, the items were hidden due to them not wearing a jumpsuit. This corrects that, and makes the items visible.

## Why It's Good For The Game
Fixes #13262

## Images of changes
image

## Changelog
:cl:
fix: Fixes certain worn items being impossible to strip off of IPCs if they weren't wearing a jumpsuit
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/_species.dm
+++ b/code/modules/mob/living/carbon/human/species/_species.dm
@@ -634,11 +634,9 @@
 		if(slot_belt)
 			if(H.belt)
 				return FALSE
-			var/obj/item/organ/external/O = H.get_organ(BODY_ZONE_CHEST)
-
-			if(!H.w_uniform && !nojumpsuit && (!O || !(O.status & ORGAN_ROBOT)))
+			if(!H.w_uniform && !nojumpsuit)
 				if(!disable_warning)
-					to_chat(H, "<span class='alert'>You need a jumpsuit before you can attach this [name].</span>")
+					to_chat(H, "<span class='alert'>You need a jumpsuit before you can attach [I].</span>")
 				return FALSE
 			if(!(I.slot_flags & SLOT_BELT))
 				return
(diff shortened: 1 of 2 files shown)


SteelSlayer:
Feels like we should just remove the `O.status & ORGAN_ROBOT` part to these checks because, if you want your species to be able to equip stuff just set `nojumpsuit` to true

farie82:
> Feels like we should just remove the `O.status & ORGAN_ROBOT` part to these checks because, if you want your species to be able to equip stuff without a jumpsuit, just set `nojumpsuit` to true

Aye that or changing the checks everywhere. Your choice :D

Fox-McCloud:
>Feels like we should just remove the O.status & ORGAN_ROBOT part to these checks because, if you want your species to be able to equip stuff without a jumpsuit, just set nojumpsuit to true

The entire point is to allow people, specifically, with aug'd limbs to have built in "pockets" so to speak. Setting `nojumpsuit` is reserved for species like Golems, who can't wear clothes, but we want to be able to have pockets, by default, but aren't explicitly robotic.

SteelSlayer:
Oh heck and I just pushed that change as well. Well thhe solution I guess should be to add a check to see if the person has a robo chest when you go to strip them.

SteelSlayer:
I don't have time to find an elegant solution to this. This is also extra snowflakey since having a robotic chest deals with PDA, belt, and ID slots. Then having a robotic left leg controls the left pocket and a robotic right leg controls the right pocket.

PR #13322, opened 2020-04-20
Adds muzzle flashes for guns depending on the ammo

farie82:
## What Does This PR Do
Adds muzzle flashes to most guns using their ammo type.

- Bigger ammo types have bigger flashes while smaller ammo types will have smaller flashes.
- Energy based weapons have a small muzzle flash.
- Suppressors will lessen the flash and flash duration
- Ammo types also have flashes depending on their projectile type. Lasers are red. Disablers are blue. Tasers are yellow and say fuck you.

Guns like the KA or crossbows do not have lighting when fired. The KA does have a muzzle flash effect.

Part of Ansari's wanted list
```
-  Muzzle flash - create a brief, coloured (if appropriate for laser), lighting effect in front of the gun when fired - We might accept sprite if it is well-drawn (e.g. not the current Bay one)
```

Also fixes a bug with the `/obj/item/gun/magic/staff/focus`. Which just didn't work.

## Why It's Good For The Game
Makes guns feel more like guns.

## Images of changes
SRaJds3dMd
The muzzle flash effect itself.

6XwsQnmmKS
The effect in dark

RNCFsyg9CN
Shotgun with buckshot and fire shots

4bLI1oUj4U
Magic "guns" also have an effect

GrYQZYgbQZ
New. Muzzle flashes are less visible when a gun is silenced. Bigger rounds still produce a slightly bigger alpha so they are more visible when silenced

## Changelog
:cl:
add: Adds muzzle flashes to guns depending on the ammo they use.
fix: The mental focus staff can be used now. The ammo is no longer a list
/:cl:


farie82:
Found a bug while testing a bit more. Using flashlights and muzzle flashes does not work ATM. Gonna try to fix that. Till then it's  DNM

PR #13326, opened 2020-04-22
Lowers max charges and recharge time for Staff of Slipping

AzuleUtama:
## What Does This PR Do
Reduces both the maximum number of charges and the recharge time for the staff of slipping - an item that can be found on the clown biodome in lavaland and bought by Wizards for 1SP.

## Why It's Good For The Game
Having witnessed a round where a wizard used this combined with instant summons, as well as having used it myself as a cling, I would argue that it is far too strong in its current state regardless of whether you get this as a wizard or a lucky miner antag. The bananas can be fired very quickly, have a range of 50 unlike tasers and also work on borgs and anyone with slip protection.

The staff itself has the same charging stats as the chaos and door creation staff, where chaos is very unpredictable and thus needs the better charge stats to work around this disadvantage and door creation is purely support based and has no combat use whatsoever. This PR brings it to same level as the other more reliable and powerful staffs such as change and animation (**6 max charges, 8 second recharge time**).

I've left the price at 1SP for wizards, since the staff is powerful regardless of whether you're a wizard or miner and this would be an additional nerf for them.

## Changelog
:cl:
tweak: Reduces max charges and recharge time for staff of slipping.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/magic/staff.dm
+++ b/code/modules/projectiles/guns/magic/staff.dm
@@ -58,8 +58,6 @@
 	ammo_type = /obj/item/ammo_casing/magic/slipping
 	icon_state = "staffofslipping"
 	item_state = "staffofslipping"
-	max_charges = 10
-	recharge_rate = 2
 	fire_sound = 'sound/items/bikehorn.ogg'
 
 /obj/item/gun/magic/staff/slipping/honkmother


SteelSlayer:
It's important to note that the `recharge_rate` is in ticks, and not seconds. Each `process()` call for the staff is 2 seconds (since it uses `SSobj`), and with a recharge rate of 4, that means the staff will gain a charge every **8 seconds**.

I still think this is fine, because the staff is very power and only costs 1 spell point, just pointing that out.

AzuleUtama:
Thanks, I've updated the PR to reflect that.

datlo:
Looks good to me. :+1:

Balthusdire:
Well considering the round happening right now has a miner vamp with this staff who is essentially invulnerable because of it I'd say this is a good nerf.

BakKeener:
> 
> 
> Well considering the round happening right now has a miner vamp with this staff who is essentially invulnerable because of it I'd say this is a good nerf.

Bro, watch the metagame.

PR #13329, opened 2020-04-24
pipeline update, redundant case removal

Procdox:
- This adds some minor optimizations to the pipeline updating system, and fixes a order dependent bug that created invalid pipe 
- Portable atmo devices no longer force their connected pipeline to continually recalculate if nothing is changing.
- Portable atmo devices alert the pipeline they are connected to if they ARE changing.
- Valve connected pipelines no longer redundantly recalculate for each changing pipeline.
- If a portable atmo connector initialized BEFORE the setup_atmos_machinery processing step, it would create a network, with only itself in it. Then, when the network is built, it would be attached to another network, without deleting the first. The connector no longer creates a pipeline, when one is it will automatically update. If in the reverse case, the network will be told to update again.

Hopefully this will provide (albeit minor) runtime improvements, and reduce the number of invalid pipelines.

:cl:
tweak: reduce the number of reconcile_air calls
fix: initialization order dependent pipeline creation bug
/:cl:


Fox-McCloud:
Merge conflicts, please update. SOrry for the lateness of the reply.

Also, could you go over this in a bit more depth--atmos is...very easy to break.

PR #13334, opened 2020-04-25
Direction Locking Help Text

Kyep:
:cl: Kyep
tweak: Tweaks the help text for direction locking so that when you do it, it tells you how to cancel it. This solves all the mhelps/ahelps we get from confused new players who activate it by accident, and want to know how to turn it off.
fix: Fixes a bug where it would say you are "now facing ." (missing word) when shift-middleclicking yourself or your own tile.
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -261,12 +261,12 @@
 	if(incapacitated())
 		return
 	var/face_dir = get_cardinal_dir(src, A)
-	if(forced_look == face_dir)
+	if(!face_dir || forced_look == face_dir || A == src)
 		forced_look = null
-		to_chat(src, "<span class='notice'>You are no longer facing any direction.</span>")
+		to_chat(src, "<span class='notice'>Cancelled direction lock.</span>")
 		return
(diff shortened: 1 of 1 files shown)


Spacemanspark:
Why not just remove the feature entirely, honestly?

BakKeener:
I remember myself accidentally triggering this and it took me a bit to figure out what I did and I don't see any issue with adding a bit of text to help new users from locking themselves in facing a direction. Perhaps add a new function such as Tool Tips or something like that in Preferences so that new players will get that message but older players who are aware of this could turn it off and keep the old text.

MRgaty:
I can't think of an actual use for this feature except for memeing.
I would personally not mind this gone.

SteelSlayer:
There's only 1 gameplay aspect I use this for besides memeing and that's to always face my target as a xeno sentinel or queen. Since you have to use a stupid verb to spit, it helps to be pre-facing the spitting direction. But that could always be changed to be a click targeted fire like the simple xenomorphs or terror queen.

PR #13335, opened 2020-04-26
Makes 'nodecon' walls on derelict russian station immune to welders

### LINKED ISSUE #13304: /nodecon variant of plasteel/titanium walls are deconstructable
**Issue Description**:
You are able to deconstruct variants of walls that are not supposed to be able to be deconstructable.  Example of this being the derelict walls.

It allows player to bypass hard restrictions put into place in such space ruins and hampers the ability to modify shuttles without them being able to be nodecon.

**What happened instead**:
Players deconned the derelict walls with a welder.

It is not the intentional purpose of the walls allows players to skip challenges required to get loot.

Place down the /nodecon walls and you can deconstruct them/go to the derelict.

I noticed recently, may has been a thing for a few months now because the players that were doing this without reporting it knew exactly what to do and where to go in order to bypass everything.

Kyep:
## What Does This PR Do
Fixes #13304 - an oversight whereby the supposedly 'nodecon' titanium walls of the Russian derelict station could be easily deconstructed with a simple welder and wrench.
Does NOT make them absolutely indestructible walls such as https://github.com/ParadiseSS13/Paradise/blob/master/code/game/turfs/simulated/walls_indestructible.dm - you can still break the derelict walls if you have high explosives or similar, but it is no longer as easy as it was.

## Why It's Good For The Game
Stops people cutting through the derelict walls straight into the armory.

## Changelog
:cl: Kyep
fix: The walls of the derelict Russian station now properly resist welding tools.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/walls_mineral.dm
+++ b/code/game/turfs/simulated/walls_mineral.dm
@@ -277,6 +277,9 @@
 /turf/simulated/wall/mineral/titanium/nodecon/burn_down()
 	return
 
+/turf/simulated/wall/mineral/titanium/nodecon/welder_act()
+	return
+
 /////////////////////Plastitanium walls/////////////////////
 
 /turf/simulated/wall/mineral/plastitanium


MRgaty:
Wonder if this will affect RCDs, they are pretty easy to get and you can skip the entire delerict with them.

Kyep:
RCDs can destroy any wall so long as:

1. They have ammo
2. The wall is /turf/simulated/wall or a subtype (won't work for /turf/simulated/shuttle/wall walls)
3. The wall isn't /turf/simulated/wall/r_wall or a subtype

So no, this won't stop you breaking in with an RCD. I'm unsure if I should add that. @Denthamos

BakKeener:
I mean RCDs can only get made autolathes I think? Cargo normally wouldn't hand those out to anyone without good reason so it'll take at least mid-round to hand out due to resource constraints, plus the only pre-made one on the station is in the CE possession with the exception of engi borgs.

Kyep:
RCDs can only be created by hacked autolathes, and without upgrades from R&D, they're still very expensive in terms of metal and glass to make. They're obtainable, but most space explorers won't have one.

PR #13336, opened 2020-04-26
Makes Battlemage Armor Coldproof

Daylight2:
## What Does This PR Do
Makes the already pressureproof wizard hardsuit coldproof.

## Why It's Good For The Game
I've heard that this was originally left out so that wizards wouldn't hide in space, but we don't really have a problem of wizard rounds being too long, and the hardsuit-like appearance and action buttons are currently very misleading (Even with the description saying it's not spaceproof - Very few people examine their hardsuits before going out to spess). 

## Changelog
:cl:
tweak: Makes battlemage armor coldproof
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/spellbook.dm
+++ b/code/game/gamemodes/wizard/spellbook.dm
@@ -448,7 +448,7 @@
 //Weapons and Armors
 /datum/spellbook_entry/item/battlemage
 	name = "Battlemage Armour"
-	desc = "An ensorceled suit of armour, protected by a powerful shield. The shield can completely negate sixteen attacks before being permanently depleted. Despite appearance it is NOT spaceproof."
+	desc = "An ensorceled suit of armour, protected by a powerful shield. The shield can completely negate sixteen attacks before being permanently depleted. Completly spaceproof."
 	item_path = /obj/item/clothing/suit/space/hardsuit/shielded/wizard
 	limit = 1
 	category = "Weapons and Armors"
--- a/code/modules/clothing/suits/wiz_robe.dm
+++ b/code/modules/clothing/suits/wiz_robe.dm
@@ -156,7 +156,7 @@
 
 /obj/item/clothing/suit/space/hardsuit/shielded/wizard
 	name = "battlemage armour"
-	desc = "Not all wizards are afraid of getting up close and personal. Not spaceproof despite its appearance."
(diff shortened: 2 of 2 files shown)


AffectedArc07:
image
image

daylight2:
Sorry, babies first PR - still figuring out the rules

datlo:
Shouldnt wizards hiding in space be prevented some other way than removing a useful and expected element of a protective magical hardsuit? (call it what you want, it looks like and acts like a hardsuit, and players will keep expecting it to be spaceproof for ages because wizard is the rarest roll in the game)

Liches are still spaceproof...

Dumbdumn5:
Wizards hiding in space for a round would be against their role, making it more of a rule issue if they choose to bum out the entire time in there. If it's for a tactical bit where they shell the station with fireballs or legitimately use it to cause chaos, that should reasonably be an option. Leaving out the utility of a single hardsuit with the intent to corral playstyles is poor design, especially considering the alternatives given to wizards in terms of, as the above mentioned, on-station hardsuits, which are more plentiful than the wizard's suit and which are often unguarded in the case of engineering post round-start engine setup and medical's emergency lockup, both of which are easy to reach with somebody who has jaunt.

If we want to combat people sitting in space doing nothing but playing hide and seek with a pod pilot, keeping a hardsuit away from the point of a hardsuit in the first place isn't going to help much, the way we as admins enforce the 'play your role' rule would. Mechanically leaving the avenue open for a non-lich wizard to be exposed to space at all in the first place while enforcing the rules against people who are actually abusing that spaceproof nature would make more sense to me personally.

Daylight2:
Also, if a plasmaman rolls wizard, they get a free spacesuit (can't get it with the magic mirror though). No reason other wizards shouldn't be able to buy that for 2TC. 
The more robust wizards already hide their phylactery in space, using their spaceproofness to 'extend the round', so not having the hardsuit be spaceproof really doesn't stop wizards from goofing off in space.

variableundefined:
Hiding in Space does not make for fun gameplay. Wizards are supposed to be active antagonist causing as much as chaos as they can until they die, rather than playing hide and seek.

Playing hide and seek is more suitable for a normal traitor mode where there are multiple antagonists to create fun / chaos, so some of them can hide after completing their objectives. 

Closing due to two maintainer objection + 1 head objection.

PR #13338, opened 2020-04-27
ID Computer Demotion/Termination Upgrades

### LINKED ISSUE #13494: Input Validation: ID Computer Custom Titles & Non-English Characters
**Issue Description**:
It is possible to use non-English characters in custom job titles, by setting them on the ID computer.

Non-English characters should be rejected.

**What happened instead**:
Characters were accepted.

Possible display bugs.
We're an English-only server - this should not be allowed.

1) Set the custom title of an ID card using the ID computer to "の は でした" or similar Japaneseor other non-English characters.
2) Observe that the computer accepts this.

Unknown.

**Extra information**:
Unknown exactly what characters are accepted - likely a lot more than just Japanese.

Kyep:
## What Does This PR Do
1) Demoting/Terminating an ID now requires you to enter a legal justification (reason text). The reason is logged to game.log and shown to online admins.  If the ID is a crew ID, the ckey of the player associated with the demoted ID is also shown.
2) Demoting/Terminating an ID now notifies the head of that department via PDA. This re-uses the code implemented in #12295 but moves it to SSJobs and makes it a proc any machine can use to notify department heads about key events related to their employees.
3) Wiping all logs from the ID computer (or ID card program) now notifies admins. Admins can still see wiped logs using a centcom ID computer.
4) ID computer now allows you to set the registered/assigned name of an ID card to a name that contains numbers.
5) ID computer no longer changes job slots when manipulating a duplicate or replacement ID.
6) Fixes #13494 

## Why It's Good For The Game
1) helps admins spot demotions which reveal problematic player conduct, such as someone repeatedly demoted for sabotage as an engineer, severe incompetence in a key role, etc
2) helps heads of department stay informed about important changes to their department (like, for example, if the CMO's only chemist is terminated for being a syndicate, that's something the CMO should be told)
3) helps admins spot cases where a HOP/Captain/etc is doing ID changes that are a problem, and then wiping the records to hide the evidence.
4) allows IPC/machine crew members to get replacement ID cards that show their name correctly, including the numbers in their name. E.g.: "HIU-393".
5) stops HOPs from adding, say, another blueshield slot by making a new ID, setting it to blueshield, then transferring it to civilian.
6) stops HOPs assigning custom jobs using non-English languages (and weird symbols).

## Changelog
:cl: Kyep
add: Demoting or terminating an ID now requires you to enter a legal justification. Admins are notified, including the reason, demoted player's name/job/ckey, etc. The head of the demoted crew member's department is also notified via PDA message.
add: Wiping all logs from an ID computer now notifies admins.
fix: Fixed a restriction that was preventing ID computers from including numbers in the name printed on an ID card. This enables IPCs to get replacement IDs that show their name correctly.
fix: Fixed an issue where manipulating spare/replacement/blank IDs in a certain way could free job slots in an unintended way.
/:cl:


Quantum-M:
When you say Crew ID, you are referring to the roundstart ones and not replacement IDs later on, right?

Kyep:
@Quantum-M Correct.

procdrone:
How about an option to delete select records only? Wiping whole history just for a single change places traitor HoPs in much suspicion. Traitoring as HoP is kind of complicated already due to meta knowledge, you already know if HoP isn't doing their job and is not in his office most of the time, he is scheming and likely antag.

Kyep:
@procdrone 
I'm unlikely to add that. Traitor HOPs don't need it. They can already avoid creating records in the first place by toggling access on a card (which generates no record) as opposed to reassigning its job (which generates a record). That's even assuming they care about leaving records - most shift, records aren't checked.

ZomgPonies:
@Kyep This probably needs to be updated to also use the new sechud 'Demotion' status.

Kyep:
@ZomgPonies No, it doesn't. The "demote" sechud status means "take this person to the HOP for demotion". Setting that status on someone already demoted is pointless. People who have already been demoted using the demote button have a "demoted" title that marks them out and causes the ID computer to recognize them regardless of their sechud status.

code/controllers/subsystem/jobs.dm
```diff
@@ -643,6 +643,26 @@ SUBSYSTEM_DEF(jobs)
...
+	var/datum/job/tgt_job = GetJob(jobtitle)
+	if(!tgt_job)
+		return
+	if(!tgt_job.department_head[1])
+		return
+	var/boss_title = tgt_job.department_head[1]
+	var/obj/item/pda/target_pda
+	for(var/obj/item/pda/check_pda in GLOB.PDAs)
+		if(check_pda.ownrank == boss_title)
+			target_pda = check_pda
```
farie82, line 659: best to break after this. Behavior might change a bit but it's arbitrary anyway. Will save a bit of performance

code/game/machinery/computer/card.dm
```diff
@@ -389,7 +389,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 				if(!job_in_department(SSjobs.GetJob(t1)))
 					return 0
 				if(t1 == "Custom")
-					var/temp_t = sanitize(copytext(input("Enter a custom job assignment.","Assignment"),1,MAX_MESSAGE_LEN))
+					var/temp_t = sanitize(reject_bad_name(copytext(input("Enter a custom job assignment.","Assignment"),1,MAX_MESSAGE_LEN), 1))
```
farie82, line 392: spacing here and `TRUE` instead of the last 1

code/game/machinery/computer/card.dm
```diff
@@ -464,6 +465,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 			if(is_authenticated(usr) && !target_dept)
 				var/delcount = SSjobs.delete_log_records(scan.registered_name, TRUE)
 				if(delcount)
+					message_admins("[key_name_admin(usr)] has wiped all ID computer logs.")
```
farie82, line 468: Would be best to add a log here. `usr.create_log(MISC_LOG...` or something

code/game/machinery/computer/card.dm
```diff
@@ -435,7 +436,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 			if(is_authenticated(usr) && !target_dept)
 				var/t2 = modify
 				if((modify == t2 && (in_range(src, usr) || (istype(usr, /mob/living/silicon))) && istype(loc, /turf)))
-					var/temp_name = reject_bad_name(href_list["reg"])
+					var/temp_name = reject_bad_name(href_list["reg"], 1)
```
farie82, line 439: `TRUE` here.

code/game/machinery/computer/card.dm
```diff
@@ -503,9 +505,15 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 		if("terminate")
 			if(is_authenticated(usr) && !target_dept)
 				var/jobnamedata = modify.getRankAndAssignment()
-				log_game("[key_name(usr)] has terminated the employment of \"[modify.registered_name]\" the \"[jobnamedata]\".")
-				message_admins("[key_name_admin(usr)] has terminated the employment of \"[modify.registered_name]\" the \"[jobnamedata]\".")
+				var/reason = sanitize(copytext(input("Enter legal reason for termination. Enter nothing to cancel.","Employment Termination"),1,MAX_MESSAGE_LEN))
```
farie82, line 508: Spacing here

code/game/machinery/computer/card.dm
```diff
@@ -503,9 +505,15 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
+				if(!reason || !is_authenticated(usr) || !modify)
+					return 0
```
farie82, line 510: `FALSE`

code/game/machinery/computer/card.dm
```diff
@@ -503,9 +505,15 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
+				var/m_ckey = modify.getPlayerCkey()
+				var/m_ckey_text = m_ckey ? "([m_ckey])" : "(no ckey)"
+				log_game("[key_name(usr)] has terminated the employment of \"[modify.registered_name]\" [m_ckey_text] the \"[jobnamedata]\" for: \"[reason]\".")
+				message_admins("[key_name_admin(usr)] has terminated the employment of \"[modify.registered_name]\" [m_ckey_text] the \"[jobnamedata]\" for: \"[reason]\".")
```
farie82, line 514: Same here. A log would be nice here

code/game/machinery/computer/card.dm
```diff
@@ -517,16 +525,19 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
 				var/list/access = list()
 				var/datum/job/jobdatum = new /datum/job/civilian
 				access = jobdatum.get_access()
-
 				var/jobnamedata = modify.getRankAndAssignment()
-				log_game("[key_name(usr)] has demoted \"[modify.registered_name]\" the \"[jobnamedata]\" to \"Civilian (Demoted)\".")
-				message_admins("[key_name_admin(usr)] has demoted \"[modify.registered_name]\" the \"[jobnamedata]\" to \"Civilian (Demoted)\".")
+				var/m_ckey = modify.getPlayerCkey()
+				var/m_ckey_text = m_ckey ? "([m_ckey])" : "(no ckey)"
+				log_game("[key_name(usr)] has demoted \"[modify.registered_name]\" the \"[jobnamedata]\" [m_ckey_text] to \"Civilian (Demoted)\" for: \"[reason]\".")
```
farie82, line 537: Once more I ask for a player log
image

code/modules/modular_computers/file_system/programs/command/card.dm
```diff
@@ -221,7 +221,7 @@
 			if(is_authenticated(usr) && modify)
 				var/t1 = href_list["assign_target"]
 				if(t1 == "Custom")
-					var/temp_t = sanitize(copytext(input("Enter a custom job assignment.","Assignment"),1,MAX_MESSAGE_LEN))
+					var/temp_t = sanitize(reject_bad_name(copytext(input("Enter a custom job assignment.","Assignment"), 1, MAX_MESSAGE_LEN), 1))
```
farie82, line 224: `TRUE` instead of the last 1

code/modules/modular_computers/file_system/programs/command/card.dm
```diff
@@ -265,7 +266,7 @@
 
 		if("PRG_reg")
 			if(is_authenticated(usr))
-				var/temp_name = reject_bad_name(href_list["reg"])
+				var/temp_name = reject_bad_name(href_list["reg"], 1)
```
farie82, line 269: `TRUE`

code/modules/modular_computers/file_system/programs/command/card.dm
```diff
@@ -289,6 +290,7 @@
 			if(is_authenticated(usr))
 				var/delcount = SSjobs.delete_log_records(scan.registered_name, TRUE)
 				if(delcount)
+					message_admins("[key_name_admin(usr)] has wiped all ID computer logs.")
```
farie82, line 293: A log here would be great again

code/modules/modular_computers/file_system/programs/command/card.dm
```diff
@@ -330,9 +332,13 @@
 		if("PRG_terminate")
 			if(is_authenticated(usr))
 				var/jobnamedata = modify.getRankAndAssignment()
-				log_game("[key_name(usr)] has terminated the employment of \"[modify.registered_name]\" the \"[jobnamedata]\".")
-				message_admins("[key_name_admin(usr)] has terminated the employment of \"[modify.registered_name]\" the \"[jobnamedata]\".")
+				var/reason = sanitize(copytext(input("Enter legal reason for termination. Enter nothing to cancel.","Employment Termination"),1,MAX_MESSAGE_LEN))
+				if(!reason || !is_authenticated(usr) || !modify)
+					return
+				log_game("[key_name(usr)] has terminated the employment of \"[modify.registered_name]\" the \"[jobnamedata]\" for: \"[reason]\".")
+				message_admins("[key_name_admin(usr)] has terminated the employment of \"[modify.registered_name]\" the \"[jobnamedata]\" for: \"[reason]\".")
```
farie82, line 339: And the last log request

code/controllers/subsystem/jobs.dm
Kyep: The reason I did not was that there are some cases where a job might have multiple bosses, and both deserve to be notified. E.g: genetics. If I did insert a break here, it would lead to bugs where notifications wouldn't be reliable for jobs like that.

farie82: Only one of them will be notified here. Either the RD or the CMO. Depending who is first in the `department_head` list

PR #13339, opened 2020-04-27
Fixes #9971 - manually spawned paranormal ERT cannot reskin nullrods

### LINKED ISSUE #9971: Paranormal ERT null rod
**Problem Description**:
When manually turning a mob into a paranormal ERT, they cannot change their null rod from base form.

Like all chaplains/paranormal ERT, manually created paranormals should have the ability to change the null rod.

**What happened instead**:
Null rod cannot be changed.

Players manually made paranormal ERT cannot select their null rod form.

Get regular human mob, select equipment and gear them up as paranormal ERT member.  Cannot change null rod.

Kyep:
## What Does This PR Do
Fixes #9971
Manually spawning a human mob and dressing them as a paranormal ERT will no longer leave them unable to reskin/usr their nullrod.

## Changelog
:cl: Kyep
fix: fixed humans manually spawned in as paranormal ERT being unable to reskin their nullrods.
/:cl:

### DIFF
--- a/code/modules/response_team/ert_outfits.dm
+++ b/code/modules/response_team/ert_outfits.dm
@@ -401,8 +401,9 @@
 
 /datum/outfit/job/centcom/response_team/paranormal/pre_equip(mob/living/carbon/human/H, visualsOnly = FALSE)
 	. = ..()
-	if(H.mind)
-		H.mind.isholy = TRUE
+	if(!H.mind)
+		H.mind = new()
+	H.mind.isholy = TRUE
 
 /datum/outfit/job/centcom/response_team/paranormal/amber
 	name = "RT Paranormal (Amber)"


farie82:
I'm not sure that is the best place to put it. Other cases will probably also fail if they try to use the mind and the mob has no mind yet. I'd say a better place will be closer to the spawning or when you start equiping the mob

Kyep:
I went with this approach as it changes as little as possible while still fixing the bug.
AFAIK no other cases fail due to lacking minds.

Fox-McCloud:
this is probably going to result in mobs that have a mind, but no player---needlessly. I'm concurring with @farie82 that this probably isn't the best way to handle this.

Kyep:
CL plans to create a more general system which allows a playerless mob to be labeled, such that an effect is applied to it when that mob becomes player-controlled. As this would allow a more specific fix (label the mind as holy upon creation if a player inhabits it and a mind is thus created for it), I'm closing this in favor of CL's system.

PR #13342, opened 2020-04-27
You can now use movement keys to move out of a cryosleep pod

SteelSlayer:
## What Does This PR Do
Does as the title and CL suggest.

## Why It's Good For The Game
Currently the only way to get out of a pod is to right click the pod, and choose "eject". But I think this can confuse people, and newer players often don't know about it, so they have to make an mhelp asking how they can get out of the pod. I think this will make it much more intuitive.

## Changelog
:cl:
tweak: You can now use movement keys to move out of a cryopod
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -275,6 +275,12 @@
 
 	return control_computer != null
 
+// So the user can use movement keys to get out of the cryopod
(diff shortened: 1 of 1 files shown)


BakKeener:
Honestly surprised we haven't done this sooner, can't wait.

PR #13331, opened 2020-04-24
Travis Update Finale: Unit Tests + SQL Validation

AffectedArc07:
## What Does This PR Do
This PR is the last piece in my Travis Update set, and oh boy is it a benefit. This will now incorporate unit tests and SQL validation, both explained in more detail below.

# Unit Tests
Travis now has a third run in which is will not only compile the server, but also start an instance of it up. This will ensure that all subsystems can initialize properly, that we have no start-game runtimes, and that unit tests can run. The current unit tests will check the following things:
- Components with invalid duplication modes and duplication types
- Chemistry reagent ID typos
- Humans can successfully spawn on the map
- All subsystems initialize properly and return their parent calls
- The timer system never reaches a negative bucket count

Assuming all of these tests are successful, and no runtimes are encountered, the server will cleanly exit and the run will pass. If any of the tests fail however, or runtimes are encountered, travis will fail. 

This also ensures that https://github.com/ParadiseSS13/rust-g successfully builds and integrates with the codebase. 

# SQL Validation
Travis will now validate all files placed into the `SQL/updates` directory, ensuring they can all be successively applied on existing data. A sample SQL file of version 0 has been included, complete with sample data. It will also verify the prefixed schema and the regular schema, and run the regular schema on the game server itself. This will also allow checking that the SQL version has successfully been incremented on a PR. 

## Why It's Good For The Game
Adding better test routines will allow us to detect problems before a PR is merged in, instead of finding out on the server, and having to hastily make revert PRs. There have also been issues in the past with invalid SQL files being submitted, so this should now pick up on that

## Changelog
:cl: AffectedArc07
tweak: Travis has been fully overhauled
/:cl:

PR #13343, opened 2020-04-27
Makes ai_monitored load the cams on LateInitialize

farie82:
## What Does This PR Do
Fixes a runtime when creating ai_monitored areas:
image
Was caused because the cameras were used before they were initialised. On local servers the runtime nearly never happened due to the spawn. But on live it would happen due to the heavy server load.

## Why It's Good For The Game
Less runtimes more funtimes

## Changelog
:cl:
fix: ai_monitored areas now won't runtime on round start
/:cl:

### DIFF
--- a/code/game/area/ai_monitored.dm
+++ b/code/game/area/ai_monitored.dm
@@ -3,16 +3,14 @@
 	var/obj/machinery/camera/motioncamera = null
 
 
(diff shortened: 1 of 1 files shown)


AffectedArc07:
Remind me on merge to remove this from #13331

PR #13344, opened 2020-04-27
Fixes #13243 - Makes wands no longer work in den

### LINKED ISSUE #13243: Wand of door creation works in the wizard base
**Issue Description**:

the wand of door creation in the wizard wand assortment belt works in the wizard base, allowing you to raid the syndi base

The wand of door creation in the wands belt to work like the staff, which is blocked

**What happened instead**:
Works just fine, allowing you to do a quick spacewalk south over to the syndi base, which you also apply wand of door creation to, before making your way over to the east, where all the fancy the SST items are.

I mean, the wizard having the entire syndicate armory at his disposal gives him a slight edge

- Be wizard

- Buy wand assortment

- Apply wand of door creation to south wall

- Raid the syndicate base

- ???

- Profit

**Extra information**:
Admins where lame and didn't actually give me permission to do this in a live game, which is a shame.

Daylight2:
## What Does This PR Do
Fixes #13243
~~Makes wand of door creation not work on wizard den~~
Makes wands not work on the wizard den

## Why It's Good For The Game
Right now, wizards can spacewalk into the SST prep room, Central, Admin room etc.
Also prevents people from wasting their charges on the den

## Changelog
:cl:
fix: Wizards can no longer raid admin shuttle
tweak: Wands no longer work on the den
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/magic/wand.dm
+++ b/code/modules/projectiles/guns/magic/wand.dm
@@ -8,6 +8,7 @@
 	can_charge = 0
 	max_charges = 100 //100, 50, 50, 34 (max charge distribution by 25%ths)
 	var/variable_charges = 1
+	no_den_usage = 1
 
 /obj/item/gun/magic/wand/New()
 	if(prob(75) && variable_charges) //25% chance of listed max charges, 50% chance of 1/2 max charges, 25% chance of 1/3 max charges
@@ -118,7 +119,6 @@
 	ammo_type = /obj/item/ammo_casing/magic/teleport
 	icon_state = "telewand"
 	max_charges = 10 //10, 5, 5, 4
-	no_den_usage = 1
 	fire_sound = 'sound/magic/wand_teleport.ogg'
 
 /obj/item/gun/magic/wand/teleport/zap_self(mob/living/user)
@@ -140,7 +140,6 @@
 	fire_sound = 'sound/magic/staff_door.ogg'
 	icon_state = "doorwand"
 	max_charges = 20 //20, 10, 10, 7
-	no_den_usage = 1
 
 /obj/item/gun/magic/wand/door/zap_self(mob/living/user)
 	to_chat(user, "<span class='notice'>You feel vaguely more open with your feelings.</span>")
--- a/code/modules/projectiles/guns/magic/wand.dm
+++ b/code/modules/projectiles/guns/magic/wand.dm
@@ -140,6 +140,7 @@
 	fire_sound = 'sound/magic/staff_door.ogg'
 	icon_state = "doorwand"
 	max_charges = 20 //20, 10, 10, 7
+	no_den_usage = 1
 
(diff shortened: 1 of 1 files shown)


Shockpoint:
Shouldn't the changelog say the wand of door creation no longer works in the den?

daylight2:
I mean, "fix: Wizards are no longer allowed to dual wield instakill rifles" is technically accurate, but I guess I'll make it more precise - hopefully the changelog gets fixed soon

code/modules/projectiles/guns/magic/wand.dm
```diff
@@ -140,6 +140,7 @@
 	fire_sound = 'sound/magic/staff_door.ogg'
 	icon_state = "doorwand"
 	max_charges = 20 //20, 10, 10, 7
+	no_den_usage = 1
```
farie82, line 143: Best to use the TRUE/FALSE defines for boolean values like this.

daylight2: I just went off what the other no_den_usage checks were using, but I can change it, sure.

datlo:
I feel like you could set this on the wand parent type instead and kill two birds with one var, would prevent new wizards from wasting wand charges in the den aswell.

daylight2:
> 
> 
> I feel like you could set this on the wand parent type instead and kill two birds with one var, would prevent new wizards from wasting wand charges in the den aswell.

There you go

Fox-McCloud:
This should only apply to teleport and door creation.

The other wands aren't problematic enough to warrant locking all of them.

Daylight2:
It's more if a QOL thing so people don't waste their charges - the only wand I could see being useful whilst inside the den is resurrection, and that's only if you manage to stab yourself

taukausanake:
I think the real issue is that you don't have a place to test your wands. Would it be possible for the wizard to get a little test dimension perhaps? Kind of a practice room but maybe a separate area from the den and with walls that the door wand/staff won't work on? Out of scope for this but I think it's a good idea. Maybe even make a wand charger, a battery charger that takes wands, in that room so you don't lose charges

Fox-McCloud:
Generally, taking away players agency is something that should be done as a last resort---which is why I'd rather not this change go through.

Just altering the door make this purely just a fix.

code/modules/projectiles/guns/magic/wand.dm
```diff
@@ -8,6 +8,7 @@
 	can_charge = 0
 	max_charges = 100 //100, 50, 50, 34 (max charge distribution by 25%ths)
 	var/variable_charges = 1
+	no_den_usage = TRUE
```
farie82, line 11: This will still make all wands not allowed to be used in the den

Daylight2: Yeah, was working on fixing that - didn't show up in the DM for some reason

PR #13346, opened 2020-04-28
Removes AI notification on pAI hacking doors

Daylight2:
## What Does This PR Do
Disables the notification the AI gets when a pAI is trying to hack a door

## Why It's Good For The Game
The average assistant can hack doors faster (As well as having the ability to bolt, shock etc,) than the pAI, which takes between 50 to 1000 seconds to hack a door. There's no need for there to be the extra downside of alerting the AI that you're doing some B&E and are unlikely to move for a while.

## Changelog
:cl:
tweak: pAIs hacking doors no longer alerts the AI
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/pai/software_modules.dm
+++ b/code/modules/mob/living/silicon/pai/software_modules.dm
@@ -382,14 +382,6 @@
 		return 1
 
 /mob/living/silicon/pai/proc/hackloop()
-	var/turf/T = get_turf_or_move(src.loc)
-	for(var/mob/living/silicon/ai/AI in GLOB.player_list)
-		if(!T || !is_station_contact(T.z))
-			break
-		if(T.loc)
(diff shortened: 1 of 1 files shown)


taukausanake:
That's sound logic, but what if we did make pAI hacking faster and kept the AI notification? Less work for the user and faster but risk getting caught vs more prep work, slightly longer hacking but the AI won't find out

Daylight2:
> 
> 
> That's sound logic, but what if we did make pAI hacking faster and kept the AI notification? Less work for the user and faster but risk getting caught vs more prep work, slightly longer hacking but the AI won't find out

Unless you want to make pAI hacking almost instantaneous, there would still be very little reason to use the pAI for hacking, as having the AIs atention on you is a really good way to get caught fast. They can just bolt the door closed behind you, and laugh as the entirety of security comes rumbling in over the brute-force encryption crack.

PR #13347, opened 2020-04-28
Makes emotes that require air gasps when you have 35+ oxygen damage

Daylight2:
## What Does This PR Do
Turns howls, growls, hisses, screams and sneezes into gasps if you have more than 35 oxygen damage

## Why It's Good For The Game
It seems a bit ridiculous that people who can't gather enough breath to say a single word (happens at 10+ oxy damage) can scream loud enough to be heard 2 screens away. Also prevents people from endlessly screamspamming when being beaten to death but staying alive due to newcrit.
Note: Getting sucked by a vampire would put you at 35+ oxy damage at 140u drained, and kill you at 180, so it would only kick in in the last few seconds.

## Changelog
:cl:
tweak: You can no longer use audible emotes when you have 35+ oxygen damage
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -52,12 +52,20 @@
 				return
 		if("howl", "howls")
 			if(isvulpkanin(src))		//Only Vulpkanin can howl
-				on_CD = handle_emote_CD(100)
+				if(getOxyLoss() > 40)
+					emote("gasp")
+					return
+				else
+					on_CD = handle_emote_CD(100)
 			else
 				return
 		if("growl", "growls")
 			if(isvulpkanin(src))		//Only Vulpkanin can growl
-				on_CD = handle_emote_CD()
+				if(getOxyLoss() > 40)
+					emote("gasp")
+					return
+				else
+					on_CD = handle_emote_CD()
 			else
 				return
 		if("squish", "squishes")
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Not exactly pleased with how this is handled; this is snowflakey and copy-pasta.

Daylight2:
> 
> 
> Not exactly pleased with how this is handled; this is snowflakey and copy-pasta.

This better?

Daylight2:
Also, if anyone feels like 35 is too low/too high, feel free to let me know

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -45,6 +45,10 @@
 
 			if(!found_machine_head)								//Everyone else fails, skip the emote attempt
 				return								//Everyone else fails, skip the emote attempt
+		if("growl", "growls", "howl", "howls", "hiss", "hisses", "scream", "screams", "sneeze", "sneezes")
```
farie82, line 48: This will break the lines below because the switch already found it's matching value in these.
Best to do this check before this switch

Daylight2: Could you elaborate on that? Not too familiar with the term 'switch' in this case.

farie82: A switch can be written out like this for instance:
```
if(case1)
  dostuff
else if(case2)
  dostuff2
else if(...)
  ....
```
Actually writing it out like that will perform worse btw due to how switches work internally. 
But due to this it will stop after the first case it finds. In your case it will find growl at line 48 and then it won't go to the lines below where it checks the cooldown

AffectedArc07:
@Fox-McCloud requesting your review on this

PR #13357, opened 2020-05-01
Update mob glide_size prior to movement

Arokha:
## What Does This PR Do
Adjusts mob glide_size prior to movement to smooth movement depending on your mob's next available movement time.

## What Does This PR Do - The Long Version
glide_size is the number of pixels per tick to move the sprite during movement glides to adjacent turfs. /tg/ and Bay seem to have opted to just make this number 8, for whatever reason. Instead, you can do math to figure out how long you have in ticks until the next movement is possible, and allow that much time for gliding by adjusting glide_size appropriately.

This is a port of my PR, https://github.com/VOREStation/VOREStation/pull/7534

You may want some of the other changes there, too, since I see you still have ye olde client/Move() for the most part. Note that this only affects client/Move(), so only player-controlled mobs.

## Why It's Good For The Game
It looks nice, removes the 'jumpies', and a cat that used to work here told me a fox that works here might want it.

## Images of changes
**Running:**
para_run

**Walking:**
para_walk

## Changelog
:cl:
tweak: Adjust mob.glide_size in client/Move()
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -187,8 +187,16 @@
 			direct = newdir
 			n = get_step(mob, direct)
 
+	delay = TICKS2DS(-round(-(DS2TICKS(delay)))) //Rounded to the next tick in equivalent ds
+	mob.glide_size = world.icon_size/max(DS2TICKS(delay),1) //Down to whatever decimal
+	
 	. = ..()
 	mob.setDir(direct)
+
+	if((direct & (direct - 1)) && mob.loc == n) //moved diagonally successfully
+		delay = mob.movement_delay() * 2 //Will prevent mob diagonal moves from smoothing accurately, sadly
+
+	move_delay += delay
 
 	for(var/obj/item/grab/G in mob)
 		if(G.state == GRAB_NECK)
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Thank you very much, for this. 

I've always found the previous movement quite jarring; one of the reasons I liked TG's movement---but the downside to theirs is it was more "checkerboard" movestyle. It didn't have the jarring effect, but it had a weird "move quickly from one tile to the next" I didn't like---so smoother moving with the jump was ..."acceptable".

Looks like you solved that problem though. Again, thanks a ton.

Arokha:
Noooo problem. Hopefully it doesn't break like all your movement.

KasparoVy:
@Fox-McCloud Gotta test with dragging things & mech movement, those may be interesting

Kyep:
@Arokha 
Did you test how it looks when someone is being pulled and/or grabdragged?
It may look a little odd now when grabdragging someone.

Sheppysneik:
from what i saw in a shift, dragging people and object like closets, will cause them "follow" the character slowly to the new location instead of remaining attached to them. also me and more players reported some kind of "input lag".

Issue #13355, opened 2020-04-30
Mindslave Staff (Missionary Kit) Breaks after a single use


AzuleUtama:
**Issue Description**:

After using the staff on one person successfully, the staff never recharges and spams a charging message while held in my hand.
![image](https://user-images.githubusercontent.com/44248086/80750876-c5aec100-8b20-11ea-8f7c-cff9e620dccf.png)

**What did you expect to happen**:
Normally, the staff slowly recharges for more than 1 use.

**What happened instead**:
Never recharged, instead kept giving out the message that it was unable to charge up repeatedly. Admin spawned me a second one, but that broke as well so I gave up and asked for it to be removed.

**Why is this bad/What are the consequences**:
It's a 15TC item and now it's even worse than a 10TC mindslave implant.

**Steps to reproduce the problem**:
Be traitor, buy or spawn the missionary kit, use it once on someone - should then get charging messages and be unable to re-use it.

**When did the problem start happening**:
I only noticed it today, one admin mentioned a similar bug was reported with it before.

PR #13360, opened 2020-05-02
Fixes the missionary staff being unable to charge properly

SteelSlayer:
## What Does This PR Do
Fixes the missionary staff being unable to charge (gain faith) properly. It properly works while its in the robe wearer's hands, or anyone else's hands, as long as the staff is within 3 tiles.

## Why It's Good For The Game
Fixes: #13355

## Changelog
:cl:
fix: Fixes the missionary staff being unable to charge properly
/:cl:

### DIFF
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -959,8 +959,7 @@
 		return
 
 	// Do not allow the staff to recharge if it's more than 3 tiles away from the robe. If get_dist returns 0, the robe and the staff in the same tile.
-	var/staff_dist = get_dist(H, linked_staff)
-	if(staff_dist > 3 || staff_dist < 0)
+	if(!(get_dist(H, linked_staff) <= 3))
 		if(prob(10))	//10% chance per process should avoid being too spammy, can tweak if it ends up still being too frequent.
 			to_chat(H, "<span class='warning'>Your staff is unable to charge at this range. Get closer!</span>")
 		return
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -959,8 +959,8 @@
 		return
 
 	// Do not allow the staff to recharge if it's more than 3 tiles away from the robe. If get_dist returns 0, the robe and the staff in the same tile.
-	// Using .loc instead of get_turf() here because if the turf of H is the same as the turf of the staff, get_dist() returns -1.
-	if(!(get_dist(H.loc, linked_staff.loc) in 0 to 3))
+	var/staff_dist = get_dist(H, linked_staff)
+	if(staff_dist > 3 || staff_dist < 0)
 		if(prob(10))	//10% chance per process should avoid being too spammy, can tweak if it ends up still being too frequent.
 			to_chat(H, "<span class='warning'>Your staff is unable to charge at this range. Get closer!</span>")
 		return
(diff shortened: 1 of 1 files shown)


code/modules/clothing/suits/miscellaneous.dm
```diff
@@ -958,7 +958,9 @@
 	if(linked_staff.faith >= 100)	//if the linked staff is fully recharged, do nothing
 		return
 
-	if(!(linked_staff in range(3, get_turf(src))))		//staff won't charge at range (to prevent it from being handed off / stolen and used)
+	// Do not allow the staff to recharge if it's more than 3 tiles away from the robe. If get_dist returns 0, the robe and the staff in the same tile.
+	// Using .loc instead of get_turf() here because if the turf of H is the same as the turf of the staff, get_dist() returns -1.
+	if(!(get_dist(H.loc, linked_staff.loc) in 0 to 3))
```
farie82, line 963: Can write this easier `if(get_dist(get_turf(H), get_turf(linked_staf)) > 3)`
Can most likely also be done without the `get_turf` and instead with the `loc` or just by using `H` and `linked_staff` but you'll have to check that.

SteelSlayer: Yeah I had something like that before I settled on this. There needs to be a 0 case for when the two items are on the same tile. But, as said in my comment there, if you use `get_turf` and both `get_turf()`s return the same turf, `get_dist` gives you a -1, which makes the check fail. 

However, in actuality, that's valid for the staff to be recharging. Also, the refs for get_dist said it can return -1 for other error conditions, so that's why I didn't want to mess with using get_turf.

farie82: Hmm looking at the reference. It doesn't specify other cases than the object being the same. Maybe it also returns `-1` when they are on other Z levels? If so that might be problematic

Having the `in 0 to 3` here is the thing that I find odd so to say. Most checks use a simple `get_dist > value`.

SteelSlayer: Yeah I want to see if it would work with just 1 get_dist call, I know its probably weird :smile:
I could do something like `if(get_dist(H, staff) > 3 || get_dist(H, staff) < 0)` instead I suppose.

code/modules/clothing/suits/miscellaneous.dm
```diff
@@ -958,7 +958,9 @@
 	if(linked_staff.faith >= 100)	//if the linked staff is fully recharged, do nothing
 		return
 
-	if(!(linked_staff in range(3, get_turf(src))))		//staff won't charge at range (to prevent it from being handed off / stolen and used)
+	// Do not allow the staff to recharge if it's more than 3 tiles away from the robe. If get_dist returns 0, the robe and the staff in the same tile.
+	var/staff_dist = get_dist(H, linked_staff)
+	if(staff_dist > 3 || staff_dist < 0)
```
Fox-McCloud, line 963: Why not just do `if(staff_dist <= 3)`?

SteelSlayer: So in the DM refs for `get_dist` it says it will return -1 for certain error conditions, but I'm not sure what those error conditions are exactly. So I only want to cover 0 to 3 and not the -1.

Fox-McCloud: `get_dist() will return -1 in error conditions, such as when Loc1 and Loc2 are the same object. `

I don't think this is something to worry about. I'd go with what I suggested.

PR #13364, opened 2020-05-02
Revert "Update mob glide_size prior to movement"

### LINKED ISSUE #13363: Wonky movement update
**Issue Description**:

After the merge of https://github.com/ParadiseSS13/Paradise/pull/13357 looking at other people moving left and right makes it look like they are teleporting every few tiles. Not that up and down movement doesn't look off as well, just not as obvious. I don't quite see it on my own sprite, but it's jarring to say the least. I don't think this is a case of getting used to it IMO.

Also, I only saw it twice in the nukie round I played but using meth/+adrenals is really damn weird now. I cannot describe it accurately. I *think* the effect was delayed because of damage/lag. But after a few seconds I was just blinking around, I hardly even saw any frames before I hit walls. I don't remember that combo being that fast in such a jarring manner.

Edit:

As someone has said on the PR itself, dragging objects/people immediately shows a problem, it'll slowly try to go to the tile like a slug. And now that I think about it, I WAS dragging a corpse with the adrenals... So it's probably that too.

And wheelchairs are crippled in the same way, go figure. Probably the other mountable vehicles too.

Having somewhat cleaner movement after the merge

**What happened instead**:

Not looking at passerby's and think they blink from one tile to the next every so often

My brain cannot handle it.

Look at people moving around. 

after the merge of https://github.com/ParadiseSS13/Paradise/pull/13357

**Extra information**:

As I said my stint with meth+adrenals was just twice. So that might need further looking into at some point.

AffectedArc07:
Reverts ParadiseSS13/Paradise#13357

After speaking with people ingame (players, admins, mentors), no one has really said that this is a positive improvement. Movement while moving really slow may look smooth, however moving at normal speed just looks jarring. Not to mention that when you pull objects they lag behind you and sort of jitter about, and it just looks really jarring.

Closes #13363

9JI5ctMfag

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -187,16 +187,8 @@
 			direct = newdir
 			n = get_step(mob, direct)
 
-	delay = TICKS2DS(-round(-(DS2TICKS(delay)))) //Rounded to the next tick in equivalent ds
-	mob.glide_size = world.icon_size/max(DS2TICKS(delay),1) //Down to whatever decimal
-	
 	. = ..()
 	mob.setDir(direct)
-
-	if((direct & (direct - 1)) && mob.loc == n) //moved diagonally successfully
-		delay = mob.movement_delay() * 2 //Will prevent mob diagonal moves from smoothing accurately, sadly
-
-	move_delay += delay
 
 	for(var/obj/item/grab/G in mob)
 		if(G.state == GRAB_NECK)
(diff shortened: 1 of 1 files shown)


Arokha:
I have a fix for the pulling that I'll PR momentarily if that's annoying.

Arokha:
https://github.com/ParadiseSS13/Paradise/pull/13367

Fox-McCloud:
fix is preferable to revert. @Arokha is handling this well and it'll be fixed by the end of today/tomorrow at worst.

Spacemanspark:
Even with the fixes merged, I'm still seeing a lot of players/ admins/ mentors say the system is really... eh?
Maybe this should be re-opened and discussed more broadly, because it still feels really off. I don't want to have a kneejerk reaction towards it, but despite trying to give it a chance it really just feels janky and disorienting to work with. Not to mention I feel a billion times slower than I did before.

AffectedArc07:
Pasting my comment that I made on the fix PR

What framerate are you playing at to obtain this movement, because 90% of the people on the server are still complaining at the broken-ness and jitter of this movement

PR #13366, opened 2020-05-02
Standardise computer attack_hands procs and fix access issues

datlo:
Current computer attack_hand procs are a huge mess. They dont have any inheritance and each console just sets its own attack_hand. Some have fingerprints, some dont, its a huge mess and a lot of duplicated code.

This PR cleans all that up by having all (I think?) consoles go through their parent's attack_hand proc, and have that one handle it.

I've also cleaned up stuff like implied src and other issues in old computer code. The HONKputer has been refactored and now actually works (Lets you send a message to honk planet once per 10 minutes) Still admin-spawn only. Some consoles that were replaced (old ds shuttle console) but werent deleted from the code have been removed.

This does have some gameplay changes, however, as some consoles had req_access set but this was never actually checked anywhere, causing them to be public access.

These consoles now require access (And can be emagged to bypass it)

Mecha tracker (RD)
Atmos central control computer (Atmos)
Brig cell monitor console (Brig)
RND server monitor console (RD)

Robotic control console now requires RD access, cant be emagged. This is because it has its own security.

Additionally, emagging the supply shuttle console will now remove the cargo access requirement, on top of allowing null crates.

Some consoles who did not register fingerprints from the people using them now should.

I've quickly tested every console and see that they function as expected as both human and AI, I dont believe I missed anything

:cl:
tweak: The following computers now require access to use. Can be bypassed with an emag. RD office computers, brig cell manager, Atmos central computer, all telecomms computers.
tweak: Emagging the supply shuttle console now removes the cargo access requirement.
/:cl:


datlo:
Fixed merge conflict caused by telecomms rework which dont use consoles anymore

Also I suggest a testmerge for this before a full merge, I did test everything but this changes a LOT of consoles' behaviour

SteelSlayer:
Is the `.all_changelogs.yml` file suppose to be registering all those line changes?

datlo:
Nope! No idea what happened there. Let me look into it.

Fox-McCloud:
Will close this one; it'll have to be re-done--eventually attackhand probably does need refactoring, but @datlo wants this closed, and TGUi is likely to conflict with this anyway.

PR #13367, opened 2020-05-02
Fix up smooth movement issues

### LINKED ISSUE #13363: Wonky movement update

Arokha:
## What Does This PR Do
Fixes up some oversights in the mob smooth movement PR, namely when things that aren't you want to move with you (pulling, grabs, vehicles, flips over people, beepsky, etc)

## Why It's Good For The Game
Looks nicer
Fixes #13363 (?) Though the issues in that issue sound somewhat like a byond version issue. Nobody downstream running 513 has noticed any sort of instajumps, unless it was related to the issues fixed in this one.

## Images of changes
ezgif-7-3c918878e8f0

## Changelog
:cl:
fix: fix a few mob smooth movement issues
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -1,6 +1,7 @@
 /atom/movable
 	layer = 3
 	appearance_flags = TILE_BOUND
+	glide_size = 8 // Default, adjusted when mobs move based on their movement delays
 	var/last_move = null
 	var/anchored = 0
 	var/move_resist = MOVE_RESIST_DEFAULT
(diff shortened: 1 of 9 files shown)


AffectedArc07:
What framerate are you playing at to obtain this movement, because 90% of the people on the server are still complaining at the broken-ness and jitter of this movement

PR #13369, opened 2020-05-03
Vendor refactor

MarsM0nd:
## What Does This PR Do

Refactors the vendors to directly use lists for their slogans and unimplemented ads, instead of a string, they use to create that list.

## Why It's Good For The Game

Should reduce the time needed in the initialization a bit, and needs less memory. Will also avoid duplicates in the list in case the vendor gets initialized again for some reason.

## Changelog
:cl:
tweak: Vendor slogans are handled nicer, no gameplay changes.
/:cl:

Not sure if the ads_list should be kept, as it's unimplemented for quite some time, as far as i can tell, can remove them if requested, otherwise, it's marked that it would be a todo from some time.

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -92,7 +92,7 @@
 		build_inventory(products, product_records)
 		build_inventory(contraband, hidden_records)
 		build_inventory(premium, coin_records)
-	if(slogan_list.len > 0)
+	if(LAZYLEN(slogan_list))
 		// So not all machines speak at the exact same time.
 		// The first time this machine says something will be at slogantime + this random value,
 		// so if slogantime is 10 minutes, it will say it at somewhere between 10 and 20 minutes after the machine is created.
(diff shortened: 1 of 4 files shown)


SteelSlayer:
Personally, I would just remove the `ads_list` thing, as its not even clear on what it was ever meant to do.

And my god, these pseudo lists with strings are awful. I had no idea they were written like that, good that they're now proper lists :+1:

code/game/machinery/vending.dm
```diff
@@ -95,17 +92,12 @@
 		build_inventory(products, product_records)
 		build_inventory(contraband, hidden_records)
 		build_inventory(premium, coin_records)
-	if(product_slogans)
-		slogan_list += splittext(product_slogans, ";")
-
+	if(slogan_list.len > 0)
```
Fox-McCloud, line 95: `if(LAZYLEN(slogan_list)`

There's another `slogan_list.len > 0` that needs to be corrected, somewhere.

MarsM0nd: Changed.

PR #13373, opened 2020-05-03
Blobbernauts can now use the Say verb to automatically speak into the blob hivemind

SteelSlayer:
## What Does This PR Do
Replaces the blobbernaut telepathy verb in favor of simply using "Say" to communicate over the blob hivemind.

## Why It's Good For The Game
I believe this will make it more user-friendly and less annoying because it eliminates the need to click on a verb every time you want to speak in the hivemind. Also this verb was especially bad since it created an entirely new verb tab just for itself.

## Changelog
:cl:
tweak: Blobbernauts can now speak in the blob hivemind by using the Say verb.
del: Blobbernauts can no longer speak out loud.
del: Removed Blobbernaut's "Blob telepathy" verb and the removed the Blobbernaut verb tab.
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -226,6 +226,7 @@
 
 #define isAIEye(A)		(istype((A), /mob/camera/aiEye))
 #define isovermind(A)	(istype((A), /mob/camera/blob))
+#define isblobbernaut(A)	(istype((A), /mob/living/simple_animal/hostile/blob/blobbernaut))
 
 #define isSpirit(A)		(istype((A), /mob/spirit))
 #define ismask(A)		(istype((A), /mob/spirit/mask))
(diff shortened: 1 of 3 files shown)


BakKeener:
Honestly just replace it with auto talk in blob telepathy, there is no reason for blobbernauts to need to talk to crew and I'm not even sure if they can understand them regardless.

SteelSlayer:
Probably a better idea yeah, I'll do that once I figure it out.

SteelSlayer:
Yeah it probably would be. The only issue I see is passing in the `overmind` var into the language, so when the blob mobs talk it says (ripping tendrils) and such. I've not worked with languages before, but if its possible then I could try.

SteelSlayer:
Going to remake this eventually, but use a proper hive mind language instead like Farie suggested.

PR #13374, opened 2020-05-03
Newcrit Death Threshold Tweaks

Fox-McCloud:
Credit to Goon for this one.

Tweaks newcrit just a bit to make brute and burn damage more effective towards killing someone.

This does not increase the probability of them dying from brute/burn---that remains unchanged---changing that is likely to have negative consequences that'll make treating deeply critical patients far more frustrating than it should be.

This alters the absolute threshold for dying from these damage types, by making brute and burn both count for 67% more towards that threshold.

What this means: you'll die, absolutely, if you've taken 360 brute (or burn, or combination there-of) damage, instead of 500. Toxin's weighting is the same (500 total), as is Oxy total (1000).

This should make engagements that are clearly one-sided less time consuming to finish while ultimately not impacting, much, the overall medical treatment system.

:cl: Fox McCloud
tweak: Brute and burn damage thresholds for death lowered from 500 to 360
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -810,7 +810,9 @@
 
 		handle_organs()
 
-		if(getBrainLoss() >= 120 || (health + (getOxyLoss() / 2)) <= -500)
+		var/guaranteed_death_threshold = health + (getOxyLoss() * 0.5) - (getFireLoss() * 0.67) - (getBruteLoss() * 0.67)
+
+		if(getBrainLoss() >= 120 || (guaranteed_death_threshold) <= -500)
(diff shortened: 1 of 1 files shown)


MRgaty:
It took so long... But its totally worth it.
Thank you, +1000

BakKeener:
So this merely makes it a tad bit easier to die then before, It takes around 250 damage total to fully down someone before this PR, does this happen to affect that at all? Because if you do not reach the 250 damage threshold you can and most likely will be able to alert and possibly call for help.

Fox-McCloud:
It's never been 250.

It's always been 200 to hard incap someone; that's never not been the case.

PR #6431, opened 2017-02-13
Makes Gateway Explorer an official job

Kyep:
Reasons why we should have Gateway Explorer as a job:
1) We need more job slots. With >100 people regularly on the server, adding new jobs is important in order to avoid having 50+ Civilians on shift. Gateway Explorer, as a potential job, is a popular job, with existing mechanics, that can support several slots, and requires almost no map edits to make it viable (just placing the starting positions). In terms of jobs we could add to boost slots, it does well on every measure. If we want to add new jobs to increase slots (which we have to, really), then Gateway Explorer should be at the top of the list.
2) It is already an unofficial job. We get several people asking for it every shift. This isn't good, because, as an unofficial job, its not managed or regulated very well. It doesn't show up on sec HUDs, it requires a lot of time from the HoP to just process 'I want to become a GE' requests, etc. We don't accept "oh, lets leave this job out of the code and make people ask for it manually every shift" for ANY other job on the station that people regularly do. Why should we accept it for Gateway Explorer? We shouldn't. It should be an official job like any other.
3) GE provides an outlet for people who would otherwise Greytide. Now, they can go off into the gateway and kill things / destroy stuff there. Fun for them, less hassle for everyone else. We have the pop to support it.

OK, so, you've convinced me. What do Gateway Explorers get?
- Starting position in Gateway room, with Gateway access.
- That's it. They have to scrounge all their supplies by ordering them from cargo, interacting with / persuading the crew, etc.

Example screenshot of a GE start:

The green jumpsuits are standard-issue. The white/gray 'G' that resembles a swirling whirlpool is their sec hud icon.

This was brought up on the forums recently here: https://nanotrasen.se/phpBB3/viewtopic.php?f=12&t=9907
While the forum thread calls for making them a Karma job, personally I think they'd do MUCH better as a regular job. So, I did not make them a karma job - they're available to everyone.

🆑 Kyep
add: After a long history of station crews sending out Explorer teams, NT has recognized this as a job.
/🆑

PR #13384, opened 2020-05-06
Explorer Transfers

Kyep:
## What Does This PR Do
The ID computer now has a predefined 'Explorer' option for job transfers under the 'support' section. This title always displays in gray.
This option grants maintenance, eva, gateway, and external airlock access, but revokes all other access. Using this means explorers do NOT get to keep their old job access anymore.
People transferred to this show up as "Explorer" in the "misc" section of crew manifest.
They also have a sec hud icon which makes them show up distinctly, instead of the old behavior from custom title where they would still show up as their old job.
Finally, transferring someone to explorer frees their original job slot. This is important because it means that if someone in a key role, such as Chaplain, Virologist, Warden, etc chooses to transfer to Explorer, their original slot is freed so that a replacement can arrive. Conversely, if an Explorer comes back to the HOP they can transfer back to their old job, but by doing so they give up their explorer access.

## Why It's Good For The Game
More convenient for HOPs to do transfers.
More convenient for Sec, who can recognize explorers on HUDs more easily.
More convenient for the crew in general, who are less likely to find themselves unexpectedly shorthanded in crucial areas because the crew in those areas chose to go into the gateway, or off exploring space.

## Images of changes
Sec HUD icon:
explorer

## Please note
Explorer will NOT show up in job prefs, or on the latejoin screen. You still have to go to the HOP/etc to get it.

## Changelog
:cl:
add: 'Explorer' is now a job transfer option in the ID computer. The only way to get this is to go to the HOP and ask permission to transfer to it.
/:cl:


Fox-McCloud:
I'd rather not entrench the notion that just because someone has a custom job, but they don't have an icon, that it's not valid or what have you. 

I'd also rather not encourage people to go off station more than they already do. IT was a bad idea back then: https://github.com/ParadiseSS13/Paradise/pull/6431

And it's still a bad idea now.

Gateway is a fundamentally at odds with the better overall design of SS13---I'd rather not entrench and encourage it.

Kyep:
@Fox-McCloud 
This PR reduces the impact of explorers on the main station:
- Explorers having their own defined access means they no longer get to keep access to their old department (and thus, will find it much harder to steal supplies from their old job)
- Explorers having their own sechud icon means they're easier to identify (and search for any contraband they bring back, as well as generating less confusion for sec)
- Explorer transfers updating job slots means that the station is far less likely to end up being short on crew in a critical role (like chaplain during vamp/cult/etc, viro during virus outbreaks, etc) because that person went exploring

These changes don't make it any easier to become an explorer.
If anything, they make it more difficult. Yeah, sure, its now an option in the ID computer... but its a discouraged option, one that always shows in gray with red highlight.

Upon someone taking the transfer, they lose access to their old department. If they later want to transfer back, they may find they can't because their slot was taken by someone else while they were transferred to explorer.

In short: I don't think this will encourage people to become explorers. I just think it will better account for the explorers we already get.

Custom jobs in general are fine.
Most custom jobs are on-station, so there's less need for those to be distinguished.
If someone's going to go off-station, though, and likely die there, never to be recovered... I think it makes sense for the ID computer to free up their old job so someone else can take it. Otherwise you end up with cases such as there being no chaplain during a cult round, since they went into the gateway.

taukausanake:
I'd like to see Explorers more integrated into research and discovery to be honest. The away missions are treated like free loot places so why not make that loot something the station can use? Strange objects that R&D can break down or study. Things that you can only bring back from there. I honestly wish most departments worked that way, make R&D depend on everyone else to get their work done rather than **make, destroy, repeat** but that's beyond this.

If Gateway is so bad to SS13 design then why hasn't it been removed? Sounds like to me it hurts the idea of the research station more than helps. If the Gateway doesn't contribute anything to gameplay surely it wouldn't hurt to remove it and clear up load time for server restarts, removing a whole map that needs to be loaded.
Can someone honestly tell me a downside to removing Gateway besides "improve don't remove" or "reeee my gameplay/options?" Someone please humor my devil's advocate here

procdrone:
I dont see why having the job is wrong.

You dont get that many people exploring space or gateway each round and changes suggested by Kyet are only for the benefit. 

**Its still command/hop decision how many people go off exploring**

And it always was. I don't see how getting the HUD icon and giving the HoP a button to quick change the job, changes anything in mindset of people.

People that wanted exploring went, and still go to HoP, HoP still did and does gives access and title. 

Its just convenience features.

Im all in support of this one.

code/__DEFINES/job.dm
```diff
@@ -51,7 +51,7 @@
 #define JOB_CLOWN				(1<<11)
 #define JOB_MIME				(1<<12)
 #define JOB_CIVILIAN			(1<<13)
-
+#define JOB_EXPLORER			(1<<14)
```
AffectedArc07, line 54: Tested locally, this flag can fit into the DB table

code/game/jobs/job/support.dm
```diff
@@ -450,3 +450,24 @@
...
+/datum/job/explorer
+	title = "Explorer"
+	flag = JOB_EXPLORER
+	department_flag = JOBCAT_SUPPORT
+	total_positions = 0
+	spawn_positions = 0
+	supervisors = "the head of personnel"
+	selection_color = "#dddddd"
+	minimal_access = list(ACCESS_MAINT_TUNNELS, ACCESS_GATEWAY, ACCESS_EVA, ACCESS_EXTERNAL_AIRLOCKS)
+	minimal_access = list(ACCESS_MAINT_TUNNELS, ACCESS_GATEWAY, ACCESS_EVA, ACCESS_EXTERNAL_AIRLOCKS)
```
AffectedArc07, line 463: This line is duplicated

farie82: Most likely meant to be `access` instead of `minimal_access`.

Kyep: Fixed.

Kyep:
@AffectedArc07 
- minimal_access issue fixed
- the icon is intended to look like a whirlpool/vortex. I picked it because its shorthand for space anomalies / gateway, and because its a 'G' that's also short for gateway. That said perhaps the icon should be changed to something else that's more shorthand for exploration in general. I'm no spriter though so I'll have to ask around for a possibly better icon.

code/__DEFINES/job.dm
Kyep: This should never end up in DB table, since players can't set a preference for it - it won't even show in the prefs screen.

taukausanake:
@Kyep maybe make the icon more match the colors that the Gateway uses? I figured it was meant to be the swirly bit (but stupid me wondered why it was a 'G') so maybe give it a little blue in there? Black borders too to match the Gateway frame? I know it's gray to match Assistants permissions but it's a nice design I feel. Just maybe give it some color

procdrone:
@taukausanake 

The job should emphasise that explorer is the civilian role and has absolutely no authority. it should hold its job tier color.

Ty-Omaha:
Why not just make it an E for Explorer

Dinarzad:
> 
> 
> Can someone honestly tell me a downside to removing Gateway besides "improve don't remove" or "reeee my gameplay/options?" Someone please humor my devil's advocate here


Devil's Advocate Engage:   'Options' is a pretty valid reason considering the huge player numbers we get.  As an example, I am in the Pacific Timezone and I play in the dead of night.  Like the -dead- of night, we're talkin' 2 to 4 AM kinda stuff, and I regularly see Paradise will 90 players at that time, at it's lowest I've seen 70 in the past month or two.

So we easily have enough players to populate almost every job on the station an still end up with a handful of civilians, some by choice cuz they wanna chill or some because there's nothing to do.   Having a Gateway missions gives an option for those restless civvies, and gets people off station, and there fore out of the hair of security, it keeps it from turning into restless greytide. And, odds are, they're gonna die.

Of course, the negative side to this is when Explorers actually -survive- the Gateway an come back bedecked in loot they are unwilling to part from and that makes them a target of Security and also a genuine threat.  I would prefer the gateway loot to be something more advantageous for the station/crew as a whole and maybe some personal upgrades, like the Scarab Eggs Guardian from the old mining (Maybe limit it to the support type only?).  But that is a hefty undertaking as it requires making either new maps or replacing a lot of the loot from the existing ones.
/Devil's Advocate Disengage

farie82:
I'm usually not in favor of the gateway. Yet we have it and it's not going to change anytime soon as it is a valid way to offload some of the overpopulation we are experiencing.
 
Currently the job is a confusing thing for security. "Hey this guy has a EVA suit but he is a civilian. Oh wait it's a gateway explorer". This PR will at least put a stop to that I hope.

See it also as an opportunity to integrate the gateways more with the station activity. Have it be something that can grant research or such.

taukausanake:
So we want to get people off the station because of over-population, but at the same time we, **Fox:** "rather not encourage people to go off station more than they already do." I wonder which is a bigger problem than the other.

I also like Gateway but I wouldn't mind if it was nuked from existence. Away missions always get the shaft so putting them out of their misery wouldn't be the worst thing to do. Really should give it more purpose than just over-population control because apparently that's a _bad thing_

MRgaty:
> Why not just make it an E for Explorer

E for explorer, E for EVA, sounds cool

DogsParadise:
See below example of icons

image

Kyep:
Updated sechud icon to one that's more popular:
explorer

PR #13386, opened 2020-05-06
Fixes Chem Grenades

Fox-McCloud:
Fixes chem grenades never reacting and them being *scalding* hot.

I made a mistake in the patch rework PR that switched around some values of `Clamp`. End result was that chem grenades would always be the maximum temperature possible (10,000 K) and this prevented them from ever reacting and causing them to spew out their contents at this insane temp.

EIther case, that fixes that. I also renamed some arguments so there's less overlap with the `min_temp` and `max_temp` of chemistry's reaction datums.

:cl: Fox McCloud
fix: Fixes chem grenades not working and being scalding hot
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -15,12 +15,12 @@
 	var/flags
 	var/list/reagents_generated_per_cycle = new/list()
 
-/datum/reagents/New(maximum = 100, min_temp, max_temp)
+/datum/reagents/New(maximum = 100, temperature_minimum, temperature_maxixmum)
 	maximum_volume = maximum
-	if(min_temp)
-		temperature_min = min_temp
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/holder.dm
```diff
@@ -15,12 +15,12 @@
...
+/datum/reagents/New(maximum = 100, temperature_minimum, temperature_maxixmum)
 	maximum_volume = maximum
-	if(min_temp)
-		temperature_min = min_temp
-	if(max_temp)
-		temperature_max = max_temp
+	if(temperature_minimum)
+		temperature_min = temperature_minimum
+	if(temperature_maxixmum)
+		temperature_max = temperature_maxixmum
```
AffectedArc07, line 23: What was the point in renaming these vars?

PR #13387, opened 2020-05-06
GC Fixes

Fox-McCloud:
Just some de-referencing fixes.

A few items weren't GC'ing because of circular references. 

Also fixes testing mode so it compiles properly.

:cl: Fox McCloud
fix: Fixed some GC issues
fix: fixes compile errors with testing mode
/:cl:

### DIFF
--- a/code/game/objects/items/devices/enginepicker.dm
+++ b/code/game/objects/items/devices/enginepicker.dm
@@ -14,8 +14,12 @@
 	var/list/list_enginebeacons = list()
 	var/isactive = FALSE
 
+/obj/item/enginepicker/Destroy()
+	list_enginebeacons.Cut()
+	return ..()
+
 /obj/item/enginepicker/attack_self(mob/living/carbon/user)
-	if(usr.stat || !usr.canmove || usr.restrained())
+	if(usr.incapacitated())
 		return
 
 	if(!isactive)
--- a/code/game/objects/items/devices/enginepicker.dm
+++ b/code/game/objects/items/devices/enginepicker.dm
@@ -19,7 +19,7 @@
 	return ..()
 
 /obj/item/enginepicker/attack_self(mob/living/carbon/user)
-	if(usr.incapacitated())
+	if(user.incapacitated())
 		return
 
 	if(!isactive)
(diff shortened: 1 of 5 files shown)


code/game/objects/items/devices/enginepicker.dm
```diff
@@ -14,8 +14,12 @@
 	var/list/list_enginebeacons = list()
 	var/isactive = FALSE
 
+/obj/item/enginepicker/Destroy()
+	list_enginebeacons.Cut()
+	return ..()
+
 /obj/item/enginepicker/attack_self(mob/living/carbon/user)
-	if(usr.stat || !usr.canmove || usr.restrained())
```
Fox-McCloud, line 18: @FreeStylaLT why. WHY?!

SteelSlayer: lol, also shouldn't we use `user` instead of `usr` here since its passed in?

Fox-McCloud: yeah

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -69,8 +69,8 @@
 		to_chat(user, "<span class='notice'>Locked In</span>")
 		new /obj/machinery/power/singularity_beacon/syndicate( user.loc )
 		playsound(src, 'sound/effects/pop.ogg', 100, 1, 1)
+		user.drop_item()
```
farie82, line 72: might want to force this. Antidrop can probably fuck it up

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -82,16 +82,21 @@
 		to_chat(user, "<span class='notice'>Locked In</span>")
 		new /obj/machinery/syndicatebomb( user.loc )
 		playsound(src, 'sound/effects/pop.ogg', 100, 1, 1)
+		user.drop_item()
```
farie82, line 85: Same here. Edge cases but they happen

code/game/objects/items/devices/enginepicker.dm
FreeStylaLT: 😅

code/game/objects/items/devices/radio/beacon.dm
Fox-McCloud: It'll still properly delete--it just might hard delete in this circumstance---sometimes it will, sometimes it work; for general circumstances like this, it'll be good enough.

PR #13392, opened 2020-05-07
Removes fires from depot

Kyep:
## What Does This PR Do
1) The depot reactor self destruct no longer spreads fire.
2) Sources of plasma have been removed from the depot. This includes tanks of plasma, a plasma airlock, and plasma floor tiles on the east side of the depot.
3) N2O gas grenade has been removed from depot armory loot table.

## Why It's Good For The Game
Recently, we have been having some issues where by the MC (master controller, a crucial piece of game code) would crash while atmos stuff was going on in the depot.
This series of changes is intended to prevent atmos stuff in the depot from affecting the stability of the server overall.

## Images of changes
Before:
depot_tank_changes

After:
depot_new
The two smaller red-boxed areas in this image have had plasma tanks removed from them.
The larger red-boxed area on the right has had a plasma tank, plasma airlock, plasma floor tiles, and falsewall spawners removed.

## Changelog
:cl: Kyep
fix: Removed fire-based hazards from depot, for the sake of server stability.
/:cl:

### DIFF
--- a/code/game/objects/effects/spawners/random_spawners.dm
+++ b/code/game/objects/effects/spawners/random_spawners.dm
@@ -289,7 +289,6 @@
 		/obj/item/clothing/glasses/thermal = 1,
 		/obj/item/chameleon = 1,
 		/obj/item/reagent_containers/hypospray/autoinjector/stimulants = 1,
-		/obj/item/storage/box/syndie_kit/atmosn2ogrenades = 1,
 		/obj/item/grenade/plastic/x4 = 1)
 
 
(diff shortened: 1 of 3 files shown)


code/game/objects/effects/spawners/random_spawners.dm
```diff
@@ -304,11 +303,6 @@
 		/turf/simulated/wall/mineral/plastitanium/nodiagonal = 2,
 		/obj/structure/falsewall/plastitanium = 2)
 
-/obj/effect/spawner/random_spawners/syndicate/layout/door/secret
-	name = "50pc falsewall 50pc wall"
-	result = list(/turf/simulated/wall/mineral/plastitanium/nodiagonal = 1,
-		/obj/structure/falsewall/plastitanium = 1)
-
```
AffectedArc07, line 311: Why did you remove this falsewall spawner?

Kyep: Because it isn't used anymore. The only place it was used was in the red-boxed area on the right, around the (now also removed) plasma tank. Those tiles are solid walls now. The object was only defined so it could be used there. Not it isn't used there, it has no justification to exist.

Fox-McCloud:
I question if this will fix the issue, but /shrug

datlo:
Chameleon flags in the syndi depot are booby trapped with plasmafire grenades. I suggest replacing it with a syndi minibomb, should be a fairly simple edit in the chameleon flag file

Kyep:
@datlo My suspicion is that it is the overload (self destruct), not fires in general, that cause the MC issues. All the recent examples of MC issues have been at times the overload was active. So, I really don't think that the flags are the cause. If, however, I see further issues, especially after flags are used, then I will go after the flags as well.

PR #13400, opened 2020-05-08
Makes it so you can't repair wooden barricades with a welder

### LINKED ISSUE #13259: Being able to repair wooden barricades with a welder
**Issue Description**:

As stated in the title, You shouldn't be able to fix something made of wood with a welder.

I expected to be able to damage it on help intent.

**What happened instead**:

I was on help intent trying to break it since we can't deconstruct them at the moment as stated in Issue #12790 and instead I got the message for attempting to repair it as well as being blinded.

You shouldn't be able to repair a object made of wood with a welder.

Step 1. Have Lit welder on help intent.
Step 2. Use said welder on wooden barricade.
Step 3. Attempting to Repair wooden barricade.
Step 4. Get blinded for not having welding protection.

I found out about it a week ago I think, not sure when this became a issue however.

trololiver112:
## What Does This PR Do
Fixes #13259
You will now hit the barricade with a welder, even on help intent. Makes Wooden barricades non repairable.

## Why It's Good For The Game
Being able to repair wooden barricades with a welder makes no sense, and will blind people thinking the same while trying to attack it with a welder on help intent.

## Images of changes
am8dXeocIj

## Changelog
:cl:

fix: Makes the wooden barricades non repairable with a welder.

/:cl:

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -30,6 +30,8 @@
 		new stacktype(get_turf(src), drop_amount)
 
 /obj/structure/barricade/welder_act(mob/user, obj/item/I)
+	if(bar_material != METAL)
+		return
 	if(obj_integrity >= max_integrity)
 		to_chat(user, "<span class='notice'>[src] does not need repairs.</span>")
 		return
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
So, there's a var that exists for this.

throw a `if(bar_material != METAL) return` in `/obj/structure/barricade/welder_act`

trololiver112:
didn't know that, makes the code cleaner aswell. Changed in the latest commit

PR #13401, opened 2020-05-08
Borgs can no longer access mounted defibs

### LINKED ISSUE #13383: Borgs can remotly use wall defibrillators
**Issue Description**:

Borgs can use wall defibrillators remotely, making the defib paddles appear under them magically

Nothing to happen when I click on the wall mount from the middle of medical

**What happened instead**:

magically appearing paddles under me 

Borgs shouldn't be able to remove the paddles.

Head to medical as a borg
Click on the wall defib from everywhere, not even beside it
magic paddles below you

Unknown
image

trololiver112:
## What Does This PR Do
Fixes #13383
Makes it so that borgs can not use mounted defibrillators at all. Previous issue with borgs being able to access them from range and drop the paddles on the floor under them, Borgs will now not be able to use them at all.

## Why It's Good For The Game
Prevents and issue with them being able to drop the paddles under them away from the mount itself, and then being unable to put them back in or use them after.

## Images of changes
8xob3E6ECj

## Changelog
:cl:

tweak: Made borgs be unable to access mounted defibrillators 

/:cl:

### DIFF
--- a/code/game/machinery/defib_mount.dm
+++ b/code/game/machinery/defib_mount.dm
@@ -15,6 +15,9 @@
 	var/obj/item/defibrillator/defib //this mount's defibrillator
 	var/clamps_locked = FALSE //if true, and a defib is loaded, it can't be removed without unlocking the clamps
 
+/obj/machinery/defibrillator_mount/attack_ai()
+	return
+
 /obj/machinery/defibrillator_mount/get_cell()
 	if(defib)
 		return defib.get_cell()


trololiver112:
Then that just leaves the question if we want borgs to be able to use them at all. The example that was stated was that a borg could drop the paddles one the ground to show a new doctor what to use.

Though other than that i see no reason why they would need to do that. Open for feedback if anyone have opinions on this

DesolateG:
Doesn't mediborg have their own defibs?

trololiver112:
They do, both the handheld one and the paddles from the more compact one used to revive dead people

PR #13402, opened 2020-05-08
Add an animation component usable by any atom. The slime scanner has been used as example

farie82:
## What Does This PR Do
Adds in a component that can flick animations based on signals or on a random interval.
The component will make sure the animations don't collide with one another.

An ingame example is added in the form of the slime scanner.
- It will randomly lose connection and thus blink. Happens every 3 to 10 minutes.
- It will show an animation if a slime is scanned.

Current limitations:
- The animations should be in the same DMI as the original sprite itself. I'm thinking about how to elegantly solve this at the moment. Support for other sprite files might be included in this PR.
- The random interval is fixed over all the random animations. One animation has the same cooldown as the other. This is partially solved with the randomness. I don't think I'll change this anytime soon
- The way it's created is a bit chunky and clunky. Will think about how to best tackle this. Might be included in this PR or a follow up PR.
- One signal has one animation tied to it. This can be done randomly or such but I won't change it in this PR.

## Why It's Good For The Game
Will allow for more interesting animations to be made.

## Images of changes
YyZjfM7n7E
Animation randomly shown every now and then

yamFbu4VUm
Animation shown when a slime is scanned

Props to @DogsParadise for making the sprites!

## Changelog
:cl: Farie82 DogsParadise 
add: Added the animation component to the game code.
add: Adds animation for the slime scanner. When slimes are scanned and on a long random interval
/:cl:


Ionward:
There’s a bunch of possibilities for us spriters to make good use of this, thanks a bunch!

code/game/objects/items/devices/scanners.dm
```diff
@@ -541,18 +541,26 @@ REAGENT SCANNER
...
-	flags = CONDUCT
+	flags = CONDUCT | NOBLUDGEON
 	throwforce = 0
 	throw_speed = 3
 	throw_range = 7
 	materials = list(MAT_METAL=30, MAT_GLASS=20)
 
+/obj/item/slime_scanner/Initialize(mapload, ...)
+	. = ..()
+	new /datum/component/animation(src, list(COMSIG_ITEM_ATTACK = new /datum/animation_object(src, "adv_spectrometer_s_scan", 14)), \
```
SteelSlayer, line 552: You'll want to use `AddComponent(...)` here I believe.

farie82: Yes you're right. Will change!

code/datums/components/animation.dm
```diff
@@ -0,0 +1,70 @@
...
+	if(length(animation_list))
+		animations = animation_list.Copy()
+		for(var/signal in animation_list)
+			if(!QDELETED(animation_list[signal])) // Invalid animation_objects will be filtered
+				RegisterSignal(parent, signal, CALLBACK(src, .proc/flick_animation, animation_list[signal], signal))
+
+	if(length(random_animation_list))
+		src.random_animations = random_animation_list.Copy()
+		src.random_animation_min_delay = random_animation_min_delay
+		src.random_animation_max_delay = random_animation_max_delay
```
Fox-McCloud, line 38: To quote the prior maintainer DZD.

"if you have to use implied `src` with your arguments, then you should rename your arguments"

Fox-McCloud:
None of this de-refrences from each other or the atom's parent.

AffectedArc07:
I know I am probably majorly overlooking this, but whats wrong with just using `flick()` on an atom? That seems a lot less convoluted than a component for this.

AffectedArc07:
This PR is currently conflicted, and standing with 2 maintainer objections and a head objection. It will be closed.

SteelSlayer:
@Kyep Components are better OOP. They are not harder to use ultimately, they may be harder to use at first because you're learning yes, but ultimately once you get the hang of it, its simple. And not wanting them just because they're "harder to use" would be holding us back from a higher code quality.

PR #13403, opened 2020-05-10
Cyborg Hypospray QOL tweak

rebel228:
## What Does This PR Do

Once the reagent in the active mode is replenished to full capacity hypospray will start to replenish reagents in other modes, one mode at a time.
## Why It's Good For The Game

Reduces the amount that medical borgs have to toggle between their modes just to ensure they're getting refilled properly. 

## Changelog
:cl:
tweak: cyborg's hypospay now replenishes reagents in inactive modes once the active mode is full.
/:cl:


code/modules/reagents/reagent_containers/borghydro.dm
```diff
@@ -56,6 +56,13 @@
 			if(RG.total_volume < RG.maximum_volume) 	//Don't recharge reagents and drain power if the storage is full.
 				R.cell.use(charge_cost) 					//Take power from borg...
 				RG.add_reagent(reagent_ids[mode], 5)		//And fill hypo with reagent.
+			else                                            //if active mode is full loop through the list and fill the first one that is not full
+				for(var/i in 1 to reagent_list.len)
```
farie82, line 60: Best to use a foreach loop here. In DM it has this syntax:
```js
for(var/i in list)
  var/type/path/R = i
  i.do_stuff()
```
You can also include the path in `for(var/type/path/R in list)`. But this will filter the list on the type you give in the select. This will be slower than the way I showed before but will ensure that all entries in the loop are of that type. If you are sure they are of one type then you can best use the first way presented. Else you can use the filter option.

code/modules/reagents/reagent_containers/borghydro.dm
```diff
@@ -56,6 +56,13 @@
 			if(RG.total_volume < RG.maximum_volume) 	//Don't recharge reagents and drain power if the storage is full.
 				R.cell.use(charge_cost) 					//Take power from borg...
 				RG.add_reagent(reagent_ids[mode], 5)		//And fill hypo with reagent.
+			else                                            //if active mode is full loop through the list and fill the first one that is not full
+				for(var/i in 1 to reagent_list.len)            
+					RG = reagent_list[i]
+					if(RG.total_volume < RG.maximum_volume)
```
farie82, line 62: Best to make this into a proc on it's own. Since it's used twice now already. Here and on line 56

code/modules/reagents/reagent_containers/borghydro.dm
farie82: I've overlooked the fact that the borg hypo has multiple `datum/reagents`. With a s. Due to this the `id` of the reagent in question is unknown. Getting it from the reagents holder will work untill it is empty.
The for loop here is a valid way to solve this issue in combination with line 64 where you get the id from the list.

code/modules/reagents/reagent_containers/borghydro.dm
```diff
@@ -114,3 +126,5 @@
 
 		if(empty)
 			. += "<span class='notice'>It is currently empty. Allow some time for the internal syntheszier to produce more.</span>"
+
+#undef BORGHYPO_REFILL_VALUE
```
farie82, line 130: Forgot a line ending after this one. The last line always should be an empty line.

code/modules/reagents/reagent_containers/borghydro.dm
```diff
@@ -69,6 +73,14 @@
 	var/datum/reagents/R = reagent_list[reagent_list.len]
 	R.add_reagent(reagent, 30)
 
+/obj/item/reagent_containers/borghypo/proc/refill_borghypo(var/datum/reagents/RG, var/reagent_id)
+	if(RG.total_volume < RG.maximum_volume)
+		var/mob/living/silicon/robot/R = loc
```
farie82, line 78: best to pass this as a parameter. Can just be called `R` or `owner`. Whatever you think fits best

code/modules/reagents/reagent_containers/borghydro.dm
rebel228: done, thank you

code/modules/reagents/reagent_containers/borghydro.dm
rebel228: done

PR #13408, opened 2020-05-10
Moves config loading to world/New not Master/New

AffectedArc07:
## What Does This PR Do
This moves loading of the game configuration to `world/New` instead of the `Master/New`. This makes it so the config datum is no re-initialised every time the master controller breaks.

## Why It's Good For The Game
This stops the configs half-breaking when the MC destabilises, and moves it to a place that makes more sense imo.

## Changelog
:cl: AffectedArc07
add: The MC recovering no longer obliterates the config datum
/:cl:


AffectedArc07:
For those wondering why this is DNM, I need a way to test if configs still work after destabilising the MC, which I am having trouble doing.

AffectedArc07:
image
This works, and configs still work after
image

PR #13413, opened 2020-05-10
Makes you lava proof while wearing the heck suit

trololiver112:
## What Does This PR Do
Makes the heck suit work like a hoodie, and while the hood is active you can now walk on lava.

## Why It's Good For The Game
Makes the suit from killing the hardest megafauna in the game actually have something unique with it, and not just a weaker drake armor that makes you hallucinate. Also fixes an issue with you not being able to spray paint the armor itself.

## Images of changes
ErLsrYYraG

YiwodCnYVh

## Changelog
:cl:

tweak: Made the heck suit work like a hoodie and making you lava proof while wearing it

/:cl:


code/modules/mining/equipment/explorer_gear.dm
```diff
@@ -52,19 +52,23 @@
...
+	armor = list("melee" = 70, "bullet" = 50, "laser" = 30, "energy" = 10, "bomb" = 50, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 100)
 	allowed = list(/obj/item/flashlight, /obj/item/tank, /obj/item/resonator, /obj/item/mining_scanner, /obj/item/t_scanner/adv_mining_scanner, /obj/item/gun/energy/kinetic_accelerator, /obj/item/pickaxe)
+	body_parts_covered = UPPER_TORSO|LOWER_TORSO|LEGS|FEET|ARMS|HANDS
+	hoodtype = /obj/item/clothing/head/hooded/heck
+	flags_2 = LAVA_PROTECT
+
 
-/obj/item/clothing/suit/space/hostile_environment/New()
+/obj/item/clothing/suit/hooded/heck/New()
 	..()
```
AffectedArc07, line 63: Change this to 
```
/obj/item/clothing/suit/hooded/heck/Initialize()
	. = ..()
```

code/__DEFINES/flags.dm
```diff
@@ -55,7 +55,7 @@
 #define NO_MAT_REDEMPTION_2		1024
 
 // LAVA_PROTECT used on the flags_2 variable for both SUIT and HEAD items, and stops lava damage. Must be present in both to stop lava damage.
-#define LAVA_PROTECT_2			2048
+#define LAVA_PROTECT			2048
```
AffectedArc07, line 58: Keep the `_2` on the end of this since its for the `flags_2` var

trololiver112:
both of those things should be changed now in this commit

BakKeener:
*Heck* time to only see this armor worn by you Fred.

code/modules/mining/equipment/explorer_gear.dm
```diff
@@ -52,19 +52,23 @@
 	max_heat_protection_temperature = FIRE_IMMUNITY_MAX_TEMP_PROTECT
 	resistance_flags = FIRE_PROOF | LAVA_PROOF
 	slowdown = 0
-	armor = list("melee" = 70, "bullet" = 40, "laser" = 10, "energy" = 10, "bomb" = 50, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 100)
+	armor = list("melee" = 70, "bullet" = 50, "laser" = 30, "energy" = 10, "bomb" = 50, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 100)
```
Kyep, line 55: A lavaland suit shouldn't have 2x as much melee resist, >3x as much bullet resist, and the same laser/energy resist, as a Security Hardsuit from the armory.

If you want it to be really good against, say, brute damage, that's okay. But don't make it amazing at melee damage, good at brute damage, as good as a sec hardsuit against laser/energy damage, AND additionally make you totally immune to lava.

datlo: Those are the same values used by the ash drake armor, except the drake armor has 30 in bullet and 50 in laser. I dont really see the issue with those being stronger than security hardsuit when the sec hardsuit is 1. spaceproof 2. a weak piece of armor to begin with.

Kyep: Much of the loot on lavaland is OP compared to the rest of the game, and arguably deserves nerfing. So, comparing this to another lavaland item is probably not going to convince me, especially when this has lavaproofing in addition to its value as armor.

datlo: I feel like armor/clothes in general deserves a major balance pass PR, especially lavaland stuff. I'll think about it. For this PR, I suppose you're right.

trololiver112: I could lower them if that is wanted, though i must say that i find it making little sense that the drake armor that is often gotten in a round is stronger and better at everything but bullet damage. Not only is the Drake armor stronger and more ofte gotten, there is nothing stopping a miner that has killed a lot of drakes to give out this armor to the crew,, as there is two armor set pr 2 dragons killed. I think it makes sense that the one set you get from killing the king of slaugther demons is both stronger considering it's unique and very very hard to get

TDSSS: Then go ahead and nerf the drake armor to balance out how easy it is to get.

trololiver112: I will will turn the the armor values on it back in that case, but keep the hand and feet covering part plus lava proof.

Will change this when i get time later

Fox-McCloud:
I'm kinda not inclined to buff the suit, to be honest.

trololiver112:
Well considering i removed the stat buff, the suit still isn’t as strong as the drake armor depending on how much you weigh this gimmick, it isn’t really that buffed.
Also considering how rarely bubblegum is killed and given how how 2/3 of his items can’t even be used by the player, this would make his drops a little less underwhelming compared to the fight.
I know comparing to drake loot isn’t the best here either, but the drake literally drops an item giving you permanent lava immunity at the price of nothing, plus lesser drake form that even does the same thing while a drake. So this wouldn’t really be a new game breaking gimmick in my opinion

code/game/turfs/simulated/floor/lava.dm
```diff
@@ -103,6 +103,13 @@
 
 			if("lava" in L.weather_immunities)
 				continue
+			if(ishuman(L))
+				var/mob/living/carbon/human/H = L
+				var/obj/item/hat =H.head
```
farie82, line 108: spacing

trololiver112: the best bad spacer this is. Should be fixed now

trololiver112:
Can’t deconflict this. 3 months old and not really getting any responses anyways. Have to close this

PR #13417, opened 2020-05-11
Adds a message to lockers and crates if they are emagged 

trololiver112:
## What Does This PR Do
adds the text "The ID lock is sparking, looks like it has been fried." to the examine text you get when examining an emagged secure locker or crate.

## Why It's Good For The Game
Makes it possible to see the difference between a locker or crate that has been opened with brute force or just emagged. Like airlocks to a degree.

## Images of changes

**A picture of both an emagged locker and an emagged crate.**

Y1yHvfhfD8

## Changelog
:cl:

tweak: Added a line of text upon examining an emagged locker or crate

/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -12,6 +12,7 @@
 	var/opened = FALSE
 	var/welded = FALSE
 	var/locked = FALSE
+	var/emagged = FALSE
 	var/wall_mounted = 0 //never solid (You can always pass over it)
 	var/lastbang
 	var/sound = 'sound/machines/click.ogg'
@@ -449,3 +450,8 @@
 /obj/structure/closet/bluespace/close()
 	. = ..()
 	density = 0
+
+/obj/structure/closet/examine(mob/user)
+	. = ..() 					//This does the standard locker examine proc
+	if(emagged)			//show text that its emagged to user
+		. += "<span class='notice'>The ID lock is sparking, looks like it was hacked</span>"
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -454,4 +454,4 @@
 /obj/structure/closet/examine(mob/user)
 	. = ..() 					//This does the standard locker examine proc
 	if(emagged)			//show text that its emagged to user
-		. += "<span class='notice'>The ID lock is sparking, looks like it was hacked</span>"
+		. += "<span class='notice'>The ID lock is sparking, looks like it was hacked.</span>"
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -454,4 +454,4 @@
 /obj/structure/closet/examine(mob/user)
 	. = ..() 					//This does the standard locker examine proc
 	if(emagged)			//show text that its emagged to user
-		. += "<span class='notice'>The ID lock is sparking, looks like it has been fried.</span>"
+		. += "<span class='warning'>The ID lock is sparking, looks like it has been fried.</span>"
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -454,4 +454,4 @@
 /obj/structure/closet/examine(mob/user)
 	. = ..() 					//This does the standard locker examine proc
 	if(emagged)			//show text that its emagged to user
-		. += "<span class='notice'>The ID lock is sparking, looks like it was hacked.</span>"
+		. += "<span class='notice'>The ID lock is sparking, looks like it has been fried.</span>"
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -325,6 +325,7 @@
 		playsound(src.loc, "sparks", 60, 1)
 		src.locked = 0
 		src.broken = 1
+		src.emagged = 1
 		update_icon()
 		to_chat(user, "<span class='notice'>You unlock \the [src].</span>")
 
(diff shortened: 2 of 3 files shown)


code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -325,6 +325,7 @@
 		playsound(src.loc, "sparks", 60, 1)
 		src.locked = 0
 		src.broken = 1
+		src.emagged = 1
```
farie82, line 328: TRUE here.

trololiver112: right, changed it

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -449,3 +450,8 @@
 /obj/structure/closet/bluespace/close()
 	. = ..()
 	density = 0
+
+/obj/structure/closet/examine(mob/user)
+	. = ..() 					//This does the standard locker examine proc
+	if(emagged)			//show text that its emagged to user
+		. += "<span class='notice'>The ID lock is sparking, looks like it was hacked.</span>"
```
AffectedArc07, line 457: The word `hacked` here seems wrong, since theres a difference between something being hacked and emagged. Maybe replace it with `looks like it has been fried`

trololiver112: Sounds better yes, will change that

trololiver112: time to update the pictures too then

Fox-McCloud:
Plausible deniability and reducing meta is a good thing in a game about paranoia and uncertainty.

Do not support this.

Qwertytoforty:
Personally, I suggested this change, as I have seen shift after shift, XXX locker is emaged, code blue / red, when in reality the locker was just speared open, or poly broke out of it on the inside. Being uncertain would be nice, but when security is telling command that they see emaged lockers constantly when a locker was just broken, it gets silly, and annoying to be going red because of it.

EMPlight:
I would suggest you add the spark/fried text if the locker was EMP opened aswell. A small amount of evidence on how the locker was opened; force, explosion, laser, emp, or emagged would be nice.

McRamon:
While i get Fox point,its kinda silly to not being able to tell if something was emagged or broken with a spear. 
Paranoia must come from different plate

AzuleUtama:
> I would suggest you add the spark/fried text if the locker was EMP opened aswell. A small amount of evidence on how the locker was opened; force, explosion, laser, emp, or emagged would be nice.

EMP doesn't break the lock, it just has a 50% chance to toggle it on or off. The locker still works fine after it's been EMP'd.

> While i get Fox point,its kinda silly to not being able to tell if something was emagged or broken with a spear.
> Paranoia must come from different plate

I think fox is more leaning on preventing sec. from shift clicking a locker that's broken and yelling ":s CMO LOCKER EMAGGED, GO RED ALERT!". Currently you don't know if it was emagged or broken into by some greytide - unless there's other evidence around such as an emagged airlock.

trololiver112:
A problem with this now as mentioned earlier is that officers that see a broken locker that could have either been emagged or broken will just jump to conclusions on it being emagged. Making it more clear what the difference is will prevent some cases where sec yells at command to go red because some assistant with a spear broke the arrivals locker.

In the end i must say yes to the fact the being able to meta something and instantly see that it is emagged and therefore know that there is an eoc is out there might be bad. But i still think its better than when they so the exact same thing but without there being an eoc. Like it is right now

Dinarzad:
I'm going to agree with Fox on this one, I think the game is better served by keeping this a little vague and not outright spelling it out on whether a locker/crate has been emagged into or just busted up or EMP'd.

Carthusia:
I mean im fine either way, but if people don't like the idea of lockers, cases and what not showing them being burned out by an emag, should it also be removed from doors?

Rather than a burnout, it simply opens/bolts the door, this by the logic above removes the instant spotting of "The door is emaged, go red" and replaces it with "The door is open, and bolted...borgs rogue? AI rogue? Malfunction? Emag? Some guy hacking it? We'll need to investigate this"

Dinarzad:
> I mean im fine either way, but if people don't like the idea of lockers, cases and what not showing them being burned out by an emag, should it also be removed from doors?

I'm actually not opposed to doing that either, I just don't know if that's worth putting into this PR as well. I generally think there should be valid options to be sneeki boi or just going loud as hell.  The main problem with the doors is that e-mags are already pretty damned strong as it is, an removing something already there is basically buffing the E-mag further, while this is more or less just keeping the power level consistant.

Ideally (To me), this is something that should take a Detective or Engineer to figure out, so that the Detective can actually be a detective and focus on solving the puzzle/whodunit portion of things. But that is both more involved and out of the scope of this PR I feel.

Carthusia:
A possible counter to the sneakier nature of an emag if the above was done, is to make it take time to do, but have it retain the sparking console as it does it.

So you walk up to the door, put the emag into your hand, use it on a door, the console starts to spark as normal, 5 seconds or so go by, and the door opens and bolts.

That being said, this puts it much closer to the other door opening item, the access tuner, which does something similar and cannot be detected, though takes longer to do.

Perhaps that is the prime difference, emags are faster, and unlock other things like lockers and cases, but they leave behind evidence that can be used to track you. While the access tuner takes longer, and only can change door states (Open, emergency access, bolted) but leaves behind no trace  anyone tampered with it.

This makes it something you can be spotted doing in the act, but if you are sneaky about it, it will take much longer for the station to catch on that your a syndicate.

Regular-Joe-SF:
I like it due it would feel consistent to me, for emag to leave a trail with lockers too, as well as it does with other stuff. Traitor's life choices, wheter to use an emag to quickly get in and take your due from the head locker, or to be sneaky and get in there by other means, steal the thing and break it open. As Cartusia wrote above I'd think the intention of the emag is to be an effective but loud tool.

variableundefined:
I also agree with Fox on plausible deniability. This will make it easier to figure out whether a box / locker was broken out illegally, taking out any deduction from it.

variableundefined:
Closing due to two maintainers objection.

PR #13418, opened 2020-05-12
Adds Nukamphetamine

MarsM0nd:
## What Does This PR Do

Alternative to #13426 
Makes Nuka Cola and Methamphetamine combine into one chemical that has the effects of both with a bit of a drawback.

## Why It's Good For The Game

Currently, this combination has basically no drawback, the biggest one can currently be dealt with another chemical or a virus. The blurred eyes should make it less attractive for combat, and the chance for a heart attack should make risky.

## Changelog
:cl:
add: Adds Nukamphetamine, which is the combination of Nuka Cola and Meth, but has a few more drawbacks.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drugs.dm
+++ b/code/modules/reagents/chemistry/reagents/drugs.dm
@@ -367,7 +367,7 @@
 	overdose_threshold = 40
 	addiction_chance = 10
 	addiction_threshold = 10
-	metabolization_rate = 1
+	metabolization_rate = 2
 	taste_description = "hyper-speed"
 
 /datum/reagent/methamphetamine/nuka_meth/on_mob_life(mob/living/M)
--- a/code/modules/reagents/chemistry/reagents/drugs.dm
+++ b/code/modules/reagents/chemistry/reagents/drugs.dm
@@ -359,6 +359,36 @@
 			M.emote("laugh")
 	return list(effect, update_flags)
 
+/datum/reagent/methamphetamine/nuka_meth
+	name = "Nukamphetamine"
+	id = "nuka_meth"
+	description = "A potent cocktail of amphetamines, caffeine and corn syrup, this stuff puts a spring in your step, a fire in your eye, and a bad case of tachycardia tremors in your somersaulting heart." //Credits for the description to https://github.com/CleverRaven/Cataclysm-DDA, from methacola, has the license of https://creativecommons.org/licenses/by-sa/3.0/
+	color = "#385642" // rgb: 56, 86, 66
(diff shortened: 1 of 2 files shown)


Aquilar:
you need to add GOTTAGOFAST_METH as well as meths antistun as well as it's brain damage chance. Since it's easily gotten around with by just premixing this and adding more meth.
You better off making cause heart damage instead of heart attack if you wanted to have downside worth caring about.

OR

you could say remove GOTTAGOFAST_METH as modifier with GOTTAGOFAST, which requires less effort.

datlo:
before meth was nerfed to be the same speed boost as nuka cola, it made sense to have the two as separate status flags.

now that they provide the same speed, i think itd be easier to just have all speed up chemicals use GOTTAGOFAST so that they dont stack with each other, making this unnecessary

MarsM0nd:
@Aquilar The nukamphetamine is a subtype of meth, so it does exactly what meth does, plus the added things, which is mostly what nuka cola gives.
And yeah, i think it will be either a solution like this, allowing the stacking at a higher risk for the user, or a simpler PR that just removes stacking possibilities.

Mitchs98:
I was eventually going to get around to doing something like this. I was just going to have it have a decent chance of heart failure/attacks if you stacked the two together or something. This however also works 😄

Spacemanspark:
Clearly it should just make the user gib when used together

Fox-McCloud:
If the intention is to nerf stacking, give them both the same chem flag.

This way is just dumb; random chance to just about drop dead on the spot because you ingested both and they made a kooky new chemical. It's effectively saying "don't use these two together under any condition!"

This also makes it a really effective poison.

MarsM0nd:
The chance of 1 percent per cycler makes it not that effective of a poison, in my limited testing. It's very obvious for the person, and basically makes you unstunnable from the meth effects, and uncatchable due to the double speedup. So when you would get poisoned, you can run to medbay for it very fast.
I'll be making a PR to remove the meth specific flag, and just use other one, in case that will be the prefered solution.

Fox-McCloud:
I'd much rather just see them not be able to be stacked, personally.

Movement is ridiculously powerful.

procdrone:
Im in total support in the way Fox proposes it. just make both chemicals not stack its speed bonuses, and not do anything else. (beyond original drawbacks of course).

Being able to run through the whole station in 2 seconds is not okay.

MarsM0nd:
Given #13426 was merged, this is no longer required or would really work, as it would just have more drawbacks than either alone.

PR #13420, opened 2020-05-12
Refactor and Fixes EMP's

Fox-McCloud:
`emp_act` can get called multiple times in certain situations, resulting in more damage done to external limbs or objects than otherwise intended....additionally, there's a lot of edge cases of EMPs not impacting things on the ground, but impacting things held by/within a mob....not...good or intuitive.

Either case, this alters thing a bit to ensure things only get hit once and ensures a few more things properly get EMP'd even when not on a mob.

Side effect: people with robot limbs get a sizeable buff as they're taking half damage--a single EMP isn't enough to one-shot an IPC anymore--it will put them into hard crit however---at least heavy EMPs. They'll be able to survive a single light EMP.

:cl: Fox McCloud
fix: Fixes EMP's impacting some objects and limbs multiple times
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -481,10 +481,10 @@
 	return ..()
 
 /obj/item/storage/emp_act(severity)
-	if(!istype(loc, /mob/living))
-		for(var/obj/O in contents)
-			O.emp_act(severity)
 	..()
+	for(var/i in contents)
+		var/atom/A = i
+		A.emp_act(severity)
 
 /obj/item/storage/hear_talk(mob/living/M as mob, list/message_pieces)
 	..()
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -926,9 +926,10 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, list(/obj/machinery/atmospherics/unary/ven
 		return initial(pixel_y)
 
 /mob/living/carbon/emp_act(severity)
-	for(var/obj/item/organ/internal/O in internal_organs)
-		O.emp_act(severity)
 	..()
+	for(var/X in internal_organs)
+		var/obj/item/organ/internal/O = X
+		O.emp_act(severity)
 
 /mob/living/carbon/Stat()
 	..()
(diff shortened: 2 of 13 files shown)


code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -481,10 +481,10 @@
 	return ..()
 
 /obj/item/storage/emp_act(severity)
-	if(!istype(loc, /mob/living))
```
Fox-McCloud, line 484: disgusting.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -926,9 +926,10 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, list(/obj/machinery/atmospherics/unary/ven
 		return initial(pixel_y)
 
 /mob/living/carbon/emp_act(severity)
-	for(var/obj/item/organ/internal/O in internal_organs)
-		O.emp_act(severity)
 	..()
+	for(var/X in internal_organs)
```
Fox-McCloud, line 930: Internal organs are technically stored within a limb, so this is needed.

code/modules/mob/living/living_defense.dm
```diff
@@ -73,10 +73,9 @@
 		return shock_damage
 
 /mob/living/emp_act(severity)
-	var/list/L = src.get_contents()
```
Fox-McCloud, line 76: and this is why that `istype(loc, /mob/living)` was needed; it fetches all contents, which is a super snowflakey proc to begin with--as such, you have to explicitly code exceptions just to prevent it from EMP'ing things twice....so dumb.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -481,10 +481,10 @@
...
-		for(var/obj/O in contents)
-			O.emp_act(severity)
 	..()
+	for(var/i in contents)
+		var/atom/A = i
```
Fox-McCloud, line 486: I would hope a non-object can't end up in storage...but you know how things go. Not that it matters a whole bunch, but..

MRgaty:
This is nice

code/game/objects/items/weapons/storage/storage.dm
AffectedArc07: Shouldn't happen but then again this BYOND

PR #13426, opened 2020-05-13
Removes gottagofast_meth

MarsM0nd:
## What Does This PR Do

Alternative to #13418.
The stacked speedup, which can be achieved without much of a drawback, is too strong. This removes it outright.

## Why It's Good For The Game

People won't be able to outrun bullets anymore, and if you can get one chemical for the speedup, you can get the other, so it's a no-brainer to use both at once.

## Changelog
:cl:
del: Removes double speedup from meth and another speedup source.
/:cl:


Fox-McCloud:
I'm fine with this, personally.

Mitchs98:
Either option that gets nuka cola + meth stacking removed I'm in favor of myself. it needs to die.

Issue #11952, opened 2019-07-31
Resting is spammable again


Spacemanspark:
**Problem Description**:
The 'rest' command seems to have been fucked up somehow. It's once again spammable--the 'cooldown' is only applied once you stop spamming... and it only prevents you from moving in any of the eight directions for a mere second or so, you can even start spamming rest again.
https://i.imgur.com/whvd40r.gifv

**What did you expect to happen**: 
I expect a small cooldown once I've used the rest command, disabling me from standing again for a second or so.

**Why is this bad/What are the consequences:**
We initially imposed the rest cooldown to prevent players from spamming the command to dodge things such as taser shots easily.

**Steps to reproduce the problem**:
Spam rest. Congratulations, you can do push ups faster than the fastest person alive can run.

**When did the problem start happening**:
Unknown, I've only found this out today.

Couls:
Can't restore the old functionality but I can stick in a do_after to add a progress bar for getting up after resting. Would you prefer I do that?

Spacemanspark:
Should I close this? Not sure if this is a bug per se.

Spacemanspark:
@affectedarc07

Spacemanspark:
I'm fairly sure this isn't a bug.

PR #13429, opened 2020-05-13
Makes resting have a cooldown again

farie82:
## What Does This PR Do
Laying down gives you a short movement delay so you can't instantly try to stand up again. like how it was prior.

Fixes: #11952

## Why It's Good For The Game
It's a bug

## Changelog
:cl:
fix: Resting has a cooldown again
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -428,11 +428,8 @@ GLOBAL_LIST_INIT(intents, list(INTENT_HELP,INTENT_DISARM,INTENT_GRAB,INTENT_HARM
 		to_chat(src, "<span class='notice'>You are now resting.</span>")
 		StartResting()
 	else if(resting)
-		visible_message("<span class='notice'>[src] is attempting to stand up.</span>", "<span class='notice'>You are trying to stand up.</span>")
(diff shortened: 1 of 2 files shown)


Fox-McCloud:
Aside from the silliness of break-dancing (which has zero real gameplay application), I kinda feel this is a solution in search of a non-existent problem; you definitely can't wavedash anymore, which was the real problem with the original rapid rest-unresting.

This'll also break a lot of things that make utilization of `rest` in game, like the dance machine and other things---since now it'll have the delay added to it; and it'll wreck havoc for anything that doesn't use `set waitfor = FALSE`

farie82:
> Aside from the silliness of break-dancing (which has zero real gameplay application), I kinda feel this is a solution in search of a non-existent problem; you definitely can't wavedash anymore, which was the real problem with the original rapid rest-unresting.
> 
> This'll also break a lot of things that make utilization of `rest` in game, like the dance machine and other things---since now it'll have the delay added to it; and it'll wreck havoc for anything that doesn't use `set waitfor = FALSE`

I made it part of the other PR to make a possible exploit less usable. I will go check the other code that use this verb since they should (most likely) just use the StartResting verb anyway.

variableundefined:
I echo Fox's concern and will be adding a second no to this - closing the PR.

Issue #10281, opened 2018-11-19
Vox Changeling Transformation Fat


Wintermote:
**Problem Description**:
Being a vox-changeling has unintended side effects if you've eaten enough food to be fat as another race, and then transform from vox to that race. The game tries to apply fat effects while you are wearing things that may not be possible to wear while fat, and the end result is being unable to interact with anything until an admin intervenes.

**What did you expect to happen**: 
Transform into non-vox with maximum food, but not fat.

**What happened instead**:
Transformed into non-vox, fat, could not move or interact with anything - not even to remove a spacesuit

**Why is this bad/What are the consequences:**
You are unable to continue playing the game without admin intervention

**Steps to reproduce the problem**:
Be Vox Changeling
Eat enough food to be fat
Wear a spacesuit and be in space
Transform into a non-vox

**When did the problem start happening**:
Occurs as of Nov 18th, 2018

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was anything else out of the ordinary happening?)**:
Changeling game mode, I was playing as a vox-changeling life support technician. I found the CE dead near the singularity, stripped him, ate his DNA, then transformed into him while still wearing my atmos tech gear and suddenly became unable to pick up objects, move, or otherwise interact with the game (the character could still talk and clicking different directions still spun them around in place). I adminhelped the issue and the intervening admin (kleyt(?) - I don't recall) was able to reset my character so I could move again.

PR #13430, opened 2020-05-13
Removes Fat Sprites

Fox-McCloud:
Removes and purges all fat sprites from the game.

Fat sprites are a pain in the butt, to say the least. They require that every species have a separate fat sprite, if you want it to work, and they require every single piece of clothing to also have a separate sprite (suits as welll)---this gets even worse if you're dealing with a species like Vox or Drask that needs its own tailor made sprites for suit---now we're doubling up on doubling up on sprites.

This system just isn't particularly maintainable, in the long term and doesn't really bring a ton, aesthetically, without major issues....thus removal.

This does *NOT* remove the ability to be fat--on the contrary, now every species can be fat if their nutrition is too high for too long, in addition to suffering its ill effects (stuck in disposals, overheating while running, the works); it's only the aesthetics that have been removed.

Tested this out to ensure that overlays and stuff work; seems fine.

fixes: https://github.com/ParadiseSS13/Paradise/issues/10281

:cl: Fox McCloud
del: removed all fat sprites and fat clothing
add: All species can now be fat
fix: Fixes being able to spam click an APC to get more "nutrition" as an APC than intended
/:cl:


Ionward:
This solves all the problems I as a spriter had with the obese/fat mechanic. Greys had issues where they'd use their regular species sprite when they were obese, and inhands also would be wonky because of overlays and the extra belly pixels.

And we get the benefit of being able to imagine and draw art for fat Vox, so its a win/win in my books.

Carthusia:
I will miss fat ol-husky biggins

I do get the reasoning why though, IndieStone had a similar issue with project zomboid in its early days where every, single sprite needed upwards to 40 seperate variations at 8 different angles to account for every single clothing item they had at the time, so they ended up going the route of 3d models, but made to look like the original 2d style, did wonders for them.

Perhaps keep the fat sprites in, but remove the blimping up portion of the game and limit it to a pre-game decision a player makes?

I honestly dont see the fat sprites used to much so I am unaware of there are enough fat clothing sprites to even make the above possible.

Fox-McCloud:
>while also removing any visual indicator that they're fat.

For the player that this impacts, there's the nutrition bar; if it's grey, you ate way too much.

For others, there's the examine text of them being chubby.

>It is already too easy to become fat by overeating during one meal

Fatness isn't merely having the "chubby" text. You have to have have to be completely full first (into the grey icon that gets shown), and stay there for 17 minutes without doing anything about it before you finally become *actually* fat and not just "chubby" (chubby has no downsides to it).

If you eat too much, you just have to go buy a weight loss shake from one of the many hot drink vendors to ease the weight off.

Getting truly fat really is a matter of deliberately ignoring something for a very very long time.

At the most extreme end, you have the super big bite burger (hardly the standard chef meal) that gives 678 nutrition---and there's already safeties in place to prevent endless gobbling of things--most foods are going to only boost your nutrition by about 100 or so.

Carthusia:
> > while also removing any visual indicator that they're fat.
> 
> For the player that this impacts, there's the nutrition bar; if it's grey, you ate way too much.
> 

Just needed to note: Drinking 2 cans of chicken noodle soup is enough to blimp up, if you eat one can at round start, you have a high chance of also blimping up.

Not sure why that soup is so strong but it really is fattening. If you needed to blob up to get gluttonys blessing, a can or two of that and BOOM, blimp.

farie82:
All for this since now only a select few races can get fat (human, taj and grey I think?). And all others can't become fat. Making those races able to just dump food into themselves to avoid going hungry. It's the small things

Also code and sprite wise it's a pain to maintain this.

Crazylemon64:
(I know this is a raw nerve for you, but a question: IPCs aren't going to get fat on electrons, are they?)

Fox-McCloud:
they shouldn't be able to get fat without adminbus; I've added an additional safety to IPCs  so they can't spam click APCs to get more nutrition than otherwise intended.

I don't think they previously could, but they could certainly get way more nutrition than intended.

code/modules/surgery/organs/augments_arms.dm
```diff
@@ -329,6 +333,7 @@
 
 /obj/item/apc_powercord/proc/powerdraw_loop(obj/machinery/power/apc/A, mob/living/carbon/human/H)
 	H.visible_message("<span class='notice'>[H] inserts a power connector into \the [A].</span>", "<span class='notice'>You begin to draw power from \the [A].</span>")
+	drawing_power = TRUE
 	while(do_after(H, 10, target = A))
```
Crazylemon64, line 337: Would `do_after_once` make more sense here

Crazylemon64: We'd get double messages with it, so it's fine without

PR #13435, opened 2020-05-14
Fixes shadowstep and knock admin attack log spam

### LINKED ISSUE #13434: Casting the shadowstep vampire power generates a ridiculous amount of attack logs
**Issue Description**:
Casting the shadowstep vampire power generates a ridiculous amount of attack logs.

LS4s6F1QY9

For it to generate only one log.

**What happened instead**:
It generated a 100 logs.

It spams admins with tons of logs for no reason and lags the chat.

Become a vampire, get the spell "shadowstep" and cast it. Observe the admin log spam.

I have no idea.

**Extra information**:
It seems to generate more logs if every tile around you is darkness. But for example if you are in a lit hallway with just a few dark tiles nearby it only generates maybe a handfull of logs.

Also there were no runtimes.

SteelSlayer:
## What Does This PR Do
`shadowstep/choose_targets()` was basically passing an entire screens worth of turfs centered around the caster, into `perform()` which generated attack logs for every turf. Bad.

## Why It's Good For The Game
Fixes #13434

## Changelog
:cl:
fix: Fixes shadowstep and knock admin attack log spam.
/:cl:

### DIFF
--- a/code/datums/spells/knock.dm
+++ b/code/datums/spells/knock.dm
@@ -13,6 +13,11 @@
 	action_icon_state = "knock"
 	sound = 'sound/magic/knock.ogg'
 
+// Knock doesn't need to generate an attack log for every turf, set `make_attack_logs` to FALSE and just create a custom one.
+/obj/effect/proc_holder/spell/aoe_turf/knock/perform(list/targets, recharge, mob/user)
+	add_attack_logs(user, user, "cast the spell [name]", ATKLOG_ALL)
+	return ..(targets, recharge, user, make_attack_logs = FALSE)
+
 /obj/effect/proc_holder/spell/aoe_turf/knock/cast(list/targets, mob/user = usr)
 	for(var/turf/T in targets)
 		for(var/obj/machinery/door/door in T.contents)
@@ -29,8 +34,6 @@
 					var/obj/structure/closet/secure_closet/SC = C
 					SC.locked = 0
 				C.open()
-
-	return
 
 /obj/effect/proc_holder/spell/aoe_turf/knock/greater
 	name = "Greater Knock"
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -507,13 +507,13 @@
 		to_chat(user, "<span class='warning'>You cannot find darkness to step to.</span>")
 		return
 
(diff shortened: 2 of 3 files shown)


trololiver112:
think knock might have some similar issues?

ObjQbZOYut

Spacemanspark:
Imagine using attack logs in the first place :^)

farie82:
> think knock might have some similar issues?
> 
> ObjQbZOYut

Knock does need to target all the turfs. But yes it has the same issue. The code should send an attacklog to admins with `ATKLOG_ALL` if a non mob is targeted.

SteelSlayer:
I could probably fix knock while I am at it then

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -507,13 +507,13 @@
...
+	turfs = list(pick(turfs)) // Pick a single turf for the vampire to jump to.
 	perform(turfs, user = user)
 
+// `targets` should only ever contain the 1 valid turf we're jumping to, even though its a list, that's just how the cast() proc works.
 /obj/effect/proc_holder/spell/vampire/shadowstep/cast(list/targets, mob/user = usr)
 	spawn(0)
-		var/turf/picked = pick(targets)
-
-		if(!picked || !isturf(picked))
+		if(!targets.len) // If for some reason the turf got deleted.
```
farie82, line 516: `LAZYLEN` here

code/datums/spells/knock.dm
```diff
@@ -13,6 +13,11 @@
 	action_icon_state = "knock"
 	sound = 'sound/magic/knock.ogg'
 
+// Knock doesn't need to generate an attack log for every turf, set `make_attack_logs` to FALSE and just create a custom one.
+/obj/effect/proc_holder/spell/aoe_turf/knock/perform(list/targets, recharge, mob/user)
+	add_attack_logs(user, user, "cast the spell [name]", ATKLOG_ALL)
```
farie82, line 18: Can give the target here as target for the logs. That way admins can jump to the location if need be

SteelSlayer: Well see, knock basically passes in a screen's worth of turfs into the `targets` list, which is why I didn't want to do that.

code/game/gamemodes/vampire/vampire_powers.dm
SteelSlayer: Yep.

PR #13436, opened 2020-05-14
Lighting Performance Improvement

Fox-McCloud:
port of: https://github.com/tgstation/tgstation/pull/51004

Improves the performance of lighting corner updates by ~16%.

:cl: Fox McCloud
tweak: lighting. performance. It's a thing people.
/:cl:

### DIFF
--- a/code/__DEFINES/lighting.dm
+++ b/code/__DEFINES/lighting.dm
@@ -90,3 +90,27 @@
 #define FLASH_LIGHT_DURATION 2
 #define FLASH_LIGHT_POWER 3
 #define FLASH_LIGHT_RANGE 3.8
+
+/// Returns the red part of a #RRGGBB hex sequence as number
+#define GETREDPART(hexa) hex2num(copytext(hexa, 2, 4))
+
+/// Returns the green part of a #RRGGBB hex sequence as number
+#define GETGREENPART(hexa) hex2num(copytext(hexa, 4, 6))
+
+/// Returns the blue part of a #RRGGBB hex sequence as number
+#define GETBLUEPART(hexa) hex2num(copytext(hexa, 6, 8))
+
+/// Parse the hexadecimal color into lumcounts of each perspective.
+#define PARSE_LIGHT_COLOR(source) \
+do { \
+	if (source.light_color) { \
+		var/__light_color = source.light_color; \
+		source.lum_r = GETREDPART(__light_color) / 255; \
+		source.lum_g = GETGREENPART(__light_color) / 255; \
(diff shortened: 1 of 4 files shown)


code/__DEFINES/lighting.dm
```diff
@@ -90,3 +90,27 @@
...
+
+/// Returns the blue part of a #RRGGBB hex sequence as number
+#define GETBLUEPART(hexa) hex2num(copytext(hexa, 6, 8))
+
+/// Parse the hexadecimal color into lumcounts of each perspective.
+#define PARSE_LIGHT_COLOR(source) \
+do { \
+	if (source.light_color) { \
+		var/__light_color = source.light_color; \
+		source.lum_r = GETREDPART(__light_color) / 255; \
```
farie82, line 108: Can make this even faster by using a multiply instead of a divide.

code/__DEFINES/lighting.dm
```diff
@@ -90,3 +90,27 @@
...
+#define GETREDPART(hexa) hex2num(copytext(hexa, 2, 4))
+
+/// Returns the green part of a #RRGGBB hex sequence as number
+#define GETGREENPART(hexa) hex2num(copytext(hexa, 4, 6))
+
+/// Returns the blue part of a #RRGGBB hex sequence as number
+#define GETBLUEPART(hexa) hex2num(copytext(hexa, 6, 8))
+
+/// Parse the hexadecimal color into lumcounts of each perspective.
+#define PARSE_LIGHT_COLOR(source) \
```
AffectedArc07, line 104: The documentation here makes me happy

code/__DEFINES/lighting.dm
Fox-McCloud: While I'd normally agree in the vast majority of cases---this is a hex value, and I'd rather keep it like this for readability purposes.

WHen you see `/ 255` in a lot of contexts, you tend to think of hex.

When you see `* 0.0039215686274509803921568627451` not so much.

TDSSS: isn't that what comments are for? Put a comment explaining it's /255 and you're good.

farie82: As long as the calculation is done right in both the code and comments (IPC code didn't for example) then it's fine I'd say. Not sure how much performance difference it would make though.

farie82: Seems multiplications and divisions are almost as fast in DM. According to tests done by Fox and Crazy. So this should be fine. Danke for the tests!

PR #13438, opened 2020-05-14
Ports Cybernetic Ears

Fox-McCloud:
Ports cybernetic ears from TG.

We previously lacked the ability to make cybernetic ears (unless you could IPC "microphones"), but now you'll be able to--printable at R&D.

This also finally allows for 100% of your internal organs to finally be cybernetic--you'll be able to pick cybernetic ears from the preference/character customization window.

Also fixes a bug where cybernetic eyes from round-start didn't change appearance.

:cl: Fox McCloud
add: Adds in cybernetic ears
add: Can start off with cybernetic ears from character customization
fix: Fixes cybernetic eyes from character customization looking like normal eyes
/:cl:


Carthusia:
Curious, is there an advanced variant that lets you hear better? or are they more or less fluff items with a medical purpose if their organic ears are somehow lost like cybernetic eyes.

Side Note: I've always found it odd how the ingame item is called "ears" when its not the exterior ears themselves but the interior vestibular system connected to them, kind of just a personal tick, but perhaps this also fixes the incorrect name?

Or should it just remain ears for the sake of convenience? 

Either way its fine, but figured id mention it.

Aquilar:
consider it's the future, and audio to nerves is significantly more refined and easier to implement than visual to nerves.

PR #13440, opened 2020-05-15
One of the Most Disgusting PRs: IPC Edition

Fox-McCloud:
Reduces IPC brute+burn vulnerability from x1.5 to x1.2.

This is effectively 25% more survival for them.

Either case, between the EMP fix and this, they'll be in a lot better position---would still be better if they could *somehow* have some process that is similar to how humans die, but IPCs are ❄️ so...

🤷‍♂️ 

Alternative to: https://github.com/ParadiseSS13/Paradise/pull/12628

:cl: Fox McCloud
tweak: Makes IPCs 25% more durable.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/machine.dm
+++ b/code/modules/mob/living/carbon/human/species/machine.dm
@@ -15,8 +15,8 @@
 	skinned_type = /obj/item/stack/sheet/metal // Let's grind up IPCs for station resources!
 
 	eyes = "blank_eyes"
-	brute_mod = 2.28 // 100% * 2.28 * 0.66 (robolimbs) ~= 150%
-	burn_mod = 2.28  // So they take 50% extra damage from brute/burn overall
+	brute_mod = 1.82 // 100% * 1.82 * 0.66 (robolimbs) ~= 120%
+	burn_mod = 1.82  // So they take 20% extra damage from brute/burn overall
 	tox_mod = 0
 	clone_mod = 0
 	death_message = "gives a short series of shrill beeps, their chassis shuddering before falling limp, nonfunctional."


BakKeener:
this takes into account of the round start augments right?

Qwertytoforty:
yes

Dumbdumn5:
In comparison to the extensive changes put into it, the sheer amount of feedback, tweaking, and testing, and the considerable community support of the linked PR, this makes none of the same changes, attempts none of the same fixes, and doesn't really do anything to further the flavor of the race beyond 'make them tankier than they were. 

I'd have to say that I personally prefer the #12628 PR by a landslide for the amount of effort put in, the widespread reception, and its substantial amount of testing and tweaking to fit community opinion.

Spacemanspark:
Echoing Dumb.

I like that IPCs are challenge mode. The modifiers for brute and burn aren't an issue--and when steel tried to do that in his own PR he was told by many to change it.

What steelslayer changes makes IPCs more interesting and not quite as bullshit death scenarios, while still leaving a good bit of challenge behind.

I'm not sure what spurred the creation of this PR, but I feel the other I'd best.

MRgaty:
Aprils fools is next year.

(Yes please)

taukausanake:
> I'm not sure what spurred the creation of this PR

I want to argue it might be some attempt to get or keep IPCs as close to new crit as possible (they got oil blood for no reason so one can only assume) so this PR being an alternative to it would mean they'd stay roughly how they are now

AdamElTablawy:
> Tweaking to fit the community opinion

The sheer effort and work put in to compromise with the playerbase for #12628 is more of an argument against it than for it, to be honest.

This one's just overall better. It does what it should. The other one is needless fluff that accomplishes nothing.

MRgaty:
> > Tweaking to fit the community opinion
> 
> The sheer effort and work put in to compromise with the playerbase for #12628 is more of an argument against it than for it, to be honest.
> 
> This one's just overall better. It does what it should. The other one is needless fluff that accomplishes nothing.

>Solves the EMP issue
>Removes the oil nerf
>keeps the same damage value for those who seek a harder experience

I wouldnt say that it is "all fluff that accomplishes nothing" when it literally solves many of the issues IPC players have with the Race.

Spacemanspark:
Please explain how the other PR is 'needless fluff' and how this PR does what 'its supposed to' where the other doesn't.

@adameltabawy

AdamElTablawy:
@Spacemanspark 

The other PR removes their blood, which is whatever, and makes EMPs kill them slightly less. They're still not newcrit integrated, and as a result are absurdly fragile compared to everything else.

This one does what it should actually be done, which is compensate for the lack of newcrit integration by making them less fragile to incoming damage. They *still* only have 100 HP before they hit crit, so it's not really any sort of argument towards 'but IPCs should be more fragile than humans' because they're already far, far more fragile by virtue of not being newcrit integrated and hitting the floor period after five shots.

What does the other PR do? Useless fluff that doesn't really matter. They die a little less hard when hit by an EMP, and yet considering every EMP in the game can be used multiple times, no balance impact. They have no blood, but that hardly matters considering they only have 100 HP.

So yes, this one does what it should-tone down the absurd level of IPC fragility while still leaving them reasonably fragile compared to humans by virtue of still having 100 usable HP before they're completely incapacitated.

The other one is a cute fluff PR, nothing more, nothing less. There's no reason the other one shouldn't be merged-it's not an alternative to this one, because it doesn't touch the same topics at all.

PR #13447, opened 2020-05-16
Adds Bulldog extended drum mags and bundle for nuke ops.

EMPlight:
BDXLM_01

## What Does This PR Do
Adds, 16 round capacity, drum mags for the Bulldog and a bundle in the nuke ops uplink.
**XL drum mags:** Slug, Buckshot, Dragon's Breath.
**12g XL mag bundle 12TC:** XLmag(Slug), XLmag(Buck), XLmag(Dragon).

## Why It's Good For The Game
**RP:** Guns usually have an extended magazine option.
Gameplay: 
**Good:**
Helps the Bulldog with assaults and large fire fights with multiple targets.
Allows ops to invest more TC into their weapon.

**Bad:**
Magazines can reload other magazines instantly. If you buy the XL bundle and the Ammo bundle you can reload the XL mags off of the 9 normal mags faster than what you would expect to be balanced.

The bundles are too cheap. At low pop nuke ops, 5 ops and 25 crew, each op would have about 27 TC to spend and could be expected to fight 5 crew members. You could by the Ammo and the XL bundle for 24 TC, having 15 mags worth of rounds firing out of extended mags. If it was one mag to kill one crew member, you would kill your 5 crew members 3 time over.
I think this is more of a flaw with the bundles than the XL mags. If the bundles were full price,18TC, a low pop nuke op would have to choose between either the Ammo bundle, the XLmag bundle, or other kinds of gear options.

Nuke ops and other game modes need to be rebalanced though.

## Changelog
:cl:
add: 16 round Slug magazine for Bulldog.
add: 16 round Buck shot magazine for Bulldog.
add: 16 round Dragon's Breath magazine for Bulldog.
add: 12TC Three extended magazines bundle for nuke-ops. Slug, Buck, and Dragon.
/:cl:


Fox-McCloud:
I don't really get the point. 

You're paying 12 TC to get 48 rounds versus the other ammo bundle which gives 72 for the same price.

The only real advantage is more shots before having to reload.

I don't really care for this, to be honest; if nukies want more shotgun ammo, then they can just buy more of it.

EMPlight:
> I don't really get the point.
> 
> You're paying 12 TC to get 48 rounds versus the other ammo bundle which gives 72 for the same price.

If both bundles had the same amount of rounds you would never by the Ammo bundle.
The op that buys the Ammo bundle has more total ammo to fire.
The op that buys the XLmag bundle has more rounds to fire in a fight.

> The only real advantage is more shots before having to reload.

Its a really great advantage. If you breach into a room or assault down a hallway, its nice to have 16 rounds over 8 to do it.

> I don't really care for this, to be honest; if nukies want more shotgun ammo, then they can just buy more of it.

Thats the idea. If you really want more ammo get the Ammo bundle. If you think that you'll never use all 9 mags and want more of an advantage over regular shotguns, get XLmags.

variableundefined:
From my experience, running out of ammo rather than running out of ammo in your mag seems to be a bigger issue.

Nukie tends to have a much bigger chance of success if they stayed coherent and covered for eachother's reload.

This would encourage those who bought XLmag to go rambo and likely get overwhelmed by numbers.

EMPlight:
> From my experience, running out of ammo rather than running out of ammo in your mag seems to be a bigger issue.

Running out of ammo in either situation is bad, ammo bag solves the former, XLmag the latter.

> Nukie tends to have a much bigger chance of success if they stayed coherent and covered for eachother's reload.

There're a lot of tactics around team work that can lead to success, teammates don't always behave that way though. Speaking of covering each other for reloads, a 16 round mag would help with coving fire and suppression while a teammate is reloading.

> This would encourage those who bought XLmag to go rambo and likely get overwhelmed by numbers.

Thats more of them getting tunnel vision and over extending. There're other weapons that encourage going rambo and over extending aswell: chain saw, desword, and the LMG. Going rambo is one of the fun things about Nuke ops.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -275,12 +275,27 @@
 	if(magazine)
 		overlays.Cut()
 		overlays += "[magazine.icon_state]"
+		if(istype(magazine, /obj/item/ammo_box/magazine/m12g/XtrLrg))
+			w_class = WEIGHT_CLASS_BULKY
+		else
+			w_class = WEIGHT_CLASS_NORMAL
 		return
```
farie82, line 282: Non required return here.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -275,12 +275,27 @@
...
 	overlays.Cut()
 	update_magazine()
 	icon_state = "bulldog[chambered ? "" : "-e"]"
+	
+/obj/item/gun/projectile/automatic/shotgun/bulldog/attackby(var/obj/item/A as obj, mob/user as mob, params)
+	if(istype(A, /obj/item/ammo_box/magazine/m12g/XtrLrg))
+		if(istype(loc, /obj/item/storage))	// To prevent inventory exploits
+			var/obj/item/storage/Strg = loc
+			if(Strg.max_w_class < WEIGHT_CLASS_BULKY)
+				to_chat(user, "<span class='info'>You can't reload [src], with a XL mag, while it's in a normal bag.</span>")
```
farie82, line 296: design: I'd make this a `warning` instead of an `info`

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -275,12 +275,27 @@
...
+				return
+	..()
```
farie82, line 298: Return ..()

code/modules/projectiles/guns/projectile.dm
```diff
@@ -46,6 +46,7 @@
 		return
 	else if(magazine.ammo_count())
 		chambered = magazine.get_round()
+		magazine.update_icon()
```
farie82, line 49: update_icon is called to many times now down the calling tree. Example in automatic.dm at line 47.
Best to flush those out as well to avoid redundant calls

farie82:
The changelog is also not valid currently in your PR description. Do change it to reflect the actual please.

code/modules/projectiles/guns/projectile.dm
EMPlight: I've moved the update_icon() but I don't know if this will fix the performance problem.

The reason for adding an update_icon() was because there’s a bug. The magazine overlay for the Bulldog isn't updated as rounds are pulled from the magazine into the weapon chamber.

If you have two rounds in the Bulldog, one in the chamber and one in the magazine, and you fire. The weapon will chamber the last round but the weapon overlay doesn't update to show an empty magazine.

EMPlight:
> Looks mostly good codewise now. Be sure to not forcepush if you can help it. It makes reviewing it harder since the history is lost.

I'll stop force pushing.

> Down the get_round references the ammo_box is still calls `icon_update`. One direct location in `ammunition.dm` on line 186.
> I think the spot you put it at is a better spot for sure but it'll be best to remove the unneeded calls down the line.

There's a merge conflict now, so I'm not sure what needs to be done.

variableundefined:
I have no idea why there's a build queued since the beginning of summer. 

I'll close the PR and see if it fixes it.

PR #13449, opened 2020-05-17
TG Label Component Port

Fox-McCloud:
Ports TG/ @SteelSlayer 's PR: https://github.com/tgstation/tgstation/pull/49700

Because people are metagamey and powergamey assholes, some people are actively using labels as a tool to "win" by labeling their equipment/items so they're impossible to use, if stolen. The person whole stole it can't utilize it without being super obvious (not to mention it removes all plausible deniability during a random search).

Paranoia and uncertainty are something that should be embedded into the core of SS13 (and traditionally have)---not to mention this completely cuts off any detective work based on fingerprints, fibers, or the counterplay of covering those up, for a lame "haha, it labeled, you lose, I win; redtext 4 u".

As such, this PR allows you to remove labels from things---if you label something that already has a label, it relabels it---if you use a hand labeler that's turned ***off***, on an item, it'll remove the label entirely.

This also implements TG's hand labeler limit; hand labelers now have 30 uses before they run out, where upon you'll have to get a new labeler or order more refills from cargo.

:cl: Fox McCloud + SteelSlayer
tweak: Hand labelers are limited use to 30 uses--once out, you'll have to refill them from a crate in cargo
tweak: Labels to objects can be removed or overwritten; use a turned off hand labeler to remove an existing label
/:cl:


SteelSlayer:
A surprise to be sure, but a welcome one.

code/datums/components/label.dm
```diff
@@ -0,0 +1,87 @@
...
+
+/// Applies a label to the name of the parent in the format of: "parent_name (label)"
+/datum/component/label/proc/apply_label()
+	var/atom/owner = parent
+	owner.name += " ([label_name])"
+
+/// Removes the label from the parent's name
+/datum/component/label/proc/remove_label()
+	var/atom/owner = parent
+	owner.name = replacetext(owner.name, "([label_name])", "") // Remove the label text from the parent's name, wherever it's located.
```
farie82, line 86: Wouldn't it be best to look for `" ([label_name])"` with the space? If another effect puts text behind the name then there will be two spaces in between if the label was applied first. Can't quickly think of an effect that does such thing.

SteelSlayer: Better to just cover all possible situations and future proof it I say :P

Carthusia:
Oooohh I like being able to remove them, i mostly use labelers to add a bit of description fluff to items I wear, like say a gold locket "Contains a photo of Arkship Metal Bound Survivor" or with my guitar and "Because we can" is scratched onto it. Helps give better mental images when you examine a player.

Those moments though, where i screwed up and it says the wrong thing, or mislabeled an item you can't really get another of...

BakKeener:
At last, no longer will people stack labels over and over and over again.

Krysonism:
One of my strongest memories of paradise is garroting a security officer who demanded to search my bag for being in maint, only to discover that every single item he wore was labelled: (property of Dave Flufferton).

This bizarre powergaming strategy is to my knowledge pretty exclusive to paradise.

Good pr.

PR #13456, opened 2020-05-19
Gravimetric Analyzer: A device for finding things in space

PidgeyThePirate:
## What Does This PR Do
Adds a new piece of equipment and design to the protolathe:
**Gravimetric Analyzer**
_A handheld suite of gravimetric sensors designed to seek out points of interest in space. Its size makes it somewhat unreliable._

**Function**
When the Analyzer is activated, it polls anywhere between 40 and 100 random tiles on the z-level. 
This depends on the scanning module installed, polled tiles = 20 + scanner level*20.
If a tile polled is not a space tile or has any object on it, it determines the cardinal direction from the user to the tile and adds one to that direction's "count".
1 count: faint
2+ counts: substantial
5+ counts: strong
10+ counts: very strong
15+ counts: massive
The Analyzer can only be activated once every ten seconds.
You can remove the scanning module with a screwdriver to replace it with a better one.

## Why It's Good For The Game
Space exploration tends to be a lot of combing back and forth through empty space. #12647 addresses this, but only for a handful of the various ruins. This device gives the user some hint of where significant sites in the z-level are without giving the game away (but how many tiles it polls is a matter of balance and the main focus of the WIP, since it directly affects the reliability of the data). 

## Images of changes
Use:
gravanalyze

Profile:
image
~1/3ms to 1/2ms depending on upgrade level, if I'm reading it right.

New sprite by McRamon:
image

## Changelog
:cl:
add: Adds Gravimetric Analyzer, a device for vaguely pointing at things in space, to the Protolathe.
/:cl:


code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+	var/scan_cd = 10 SECONDS //how long before we can scan again
+	var/scan_tile_count // The number of tiles to poll randomly.
+
+/obj/item/space_scanner/Initialize()
+	s_module = new(null)
+	scan_tile_count = 20 + (s_module.rating * 20)
+
+/obj/item/space_scanner/screwdriver_act(mob/living/user, obj/item/I)
+	to_chat(user, "<span class='notice'> You unscrew and remove the [s_module.name] from [src].</span>")
+	s_module.loc = get_turf(src.loc)
```
farie82, line 894: src implied

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+	var/next_use = 0
+	var/scan_cd = 10 SECONDS //how long before we can scan again
+	var/scan_tile_count // The number of tiles to poll randomly.
+
+/obj/item/space_scanner/Initialize()
+	s_module = new(null)
+	scan_tile_count = 20 + (s_module.rating * 20)
+
+/obj/item/space_scanner/screwdriver_act(mob/living/user, obj/item/I)
+	to_chat(user, "<span class='notice'> You unscrew and remove the [s_module.name] from [src].</span>")
```
farie82, line 893: Spacing before "You"

farie82: Might be able to just put down `s_module` instead of `the [s_module.name]`

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+	s_module.loc = get_turf(src.loc)
+	s_module = null
+	return
+
+/obj/item/space_scanner/attackby(obj/item/W, mob/user, params)
+	if(istype(W, /obj/item/stock_parts/scanning_module))
```
farie82, line 899: Use an early return here instead

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+		if(!s_module)
+			user.drop_item()
+			W.loc = src
+			s_module = W
+			to_chat(user, "<span class='notice'>You install a [s_module.name] in [src].</span>")
```
farie82, line 904: Same as the earlier comment

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+			scan_tile_count = 20 + (s_module.rating * 20)
```
farie82, line 905: Make this into a proc. It's being used twice now already so it's best to make a proc for this. Avoids mistakes etc

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+	var/turf/scanned_tile
+	var/north_count = 0
+	var/south_count = 0
+	var/east_count = 0
+	var/west_count = 0
+	var/magnitude
+
+	while(tiles_scanned < scan_tile_count)
+		scanned_tile = locate(rand((2*TRANSITIONEDGE), world.maxx - (2*TRANSITIONEDGE)), rand((2*TRANSITIONEDGE), world.maxy - (2*TRANSITIONEDGE)), 3)
+		if(!istype(scanned_tile, /turf/space) || scanned_tile.contents.len > 0)
```
farie82, line 930: Can use `isspaceturf` here

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,109 @@ REAGENT SCANNER
...
+				south_count++
+			if(direction & EAST)
+				east_count++
+			if(direction & WEST)
+				west_count++
+		tiles_scanned++
+
+	playsound(get_turf(src), 'sound/items/posiping.ogg', 80, 0)
+	to_chat(user, "<span class='notice'>The analyzer pings softly, and lines appear on the screen:</span>")
+	if(north_count>0)
```
farie82, line 944: Spacing here. Same for the lines below

code/modules/research/designs/equipment_designs.dm
```diff
@@ -280,3 +280,13 @@
 	build_type = PROTOLATHE
 	materials = list(MAT_METAL = 5000, MAT_PLASMA = 2500, MAT_TITANIUM = 500, MAT_BLUESPACE = 500)
 	category = list("Equipment")
+
+/datum/design/space_scanner
+	name = "Gravimetric Analyzer"
+	desc = "A handheld suite of sensors designed to scan local areas of space for gravimetric signatures. The compact design renders it somewhat unreliable."
+	id = "space_scanner"
+	req_tech = list("materials"=4, "programming"=4, "magnets"=4)
```
farie82, line 288: Spacing here

code/game/objects/items/devices/scanners.dm
PidgeyThePirate: So you use get_turf(loc)?

farie82: Aye indeed so

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,113 @@ REAGENT SCANNER
...
+	var/tiles_scanned = 0
+	var/turf/scanned_tile
+	var/north_count = 0
+	var/south_count = 0
+	var/east_count = 0
+	var/west_count = 0
+	var/magnitude
+
+	while(tiles_scanned < scan_tile_count)
+		scanned_tile = locate(rand((2*TRANSITIONEDGE), world.maxx - (2*TRANSITIONEDGE)), rand((2*TRANSITIONEDGE), world.maxy - (2*TRANSITIONEDGE)), 3)
```
AffectedArc07, line 933: This looks like its only going to scan on Z3.

PidgeyThePirate: Aw crap I made a mental note and then completely forgot to change that.

PidgeyThePirate:
Now would be a good time to learn how to use the profiler then.

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,113 @@ REAGENT SCANNER
...
+/obj/item/space_scanner/Initialize()
+	s_module = new(null)
+	update_scanner()
+
+/obj/item/space_scanner/proc/update_scanner()
+	scan_tile_count = 20 + (s_module.rating * 20)
+
+/obj/item/space_scanner/screwdriver_act(mob/living/user, obj/item/I)
+	to_chat(user, "<span class='notice'>You unscrew and remove [s_module] from [src].</span>")
+	s_module.loc = get_turf(loc)
```
farie82, line 897: Missed this one. Use `forceMove` here

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,113 @@ REAGENT SCANNER
...
+	s_module = null
+	return
+
+/obj/item/space_scanner/attackby(obj/item/W, mob/user, params)
+	if(!istype(W, /obj/item/stock_parts/scanning_module))
+		return
+	if(!s_module)
+		user.drop_item()
+		W.loc = src
```
farie82, line 906: Same here

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,113 @@ REAGENT SCANNER
...
+	to_chat(user, "<span class='notice'>You unscrew and remove [s_module] from [src].</span>")
+	s_module.loc = get_turf(loc)
+	s_module = null
+	return
+
+/obj/item/space_scanner/attackby(obj/item/W, mob/user, params)
+	if(!istype(W, /obj/item/stock_parts/scanning_module))
+		return
+	if(!s_module)
+		user.drop_item()
```
farie82, line 905: Best to do an antidrop check here.

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,113 @@ REAGENT SCANNER
...
+		W.loc = src
+		s_module = W
+		to_chat(user, "<span class='notice'>You install [s_module] in [src].</span>")
+		update_scanner()
+	else
+		to_chat(user, "<span class='notice'>[src] already has a scanner module.</span>")
+
+/obj/item/space_scanner/attack_self(mob/user as mob)
```
farie82, line 913: Remove the as mob part here. It's not required.

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,113 @@ REAGENT SCANNER
...
+			to_chat(user, "<span class='notice'>East: [magnitude]</span>")
+	if(west_count > 0)
+		magnitude = get_signal_magnitude(west_count)
+		if(magnitude)
+			to_chat(user, "<span class='notice'>West: [magnitude]</span>")
+	if(!north_count && !south_count && !east_count && !west_count)
+		to_chat(user, "<span class='notice'>No gravimetric signatures detected.</span>")
+	next_use = world.time + scan_cd
+
+/obj/item/space_scanner/proc/get_signal_magnitude(var/count)
```
farie82, line 970: Var implied here

code/game/objects/items/devices/scanners.dm
```diff
@@ -867,3 +867,115 @@ REAGENT SCANNER
...
+				north_count++
+			if(direction & SOUTH)
+				south_count++
+			if(direction & EAST)
+				east_count++
+			if(direction & WEST)
+				west_count++
+		tiles_scanned++
+
+	playsound(get_turf(src), 'sound/items/posiping.ogg', 80, 0)
```
farie82, line 948: Can cache the turf to be a bit better on the performance. You did this already on line 931 but line 920 also makes use of the turf.

procdrone:
Space isn't that big to warrant such a tool , i think. Most space ruins are large enough to hit them... and a GPS is more than enough to navigate z levels.

But, thats just like my opinion.

VeteranKyl:
In my experience with space exploration, the issue comes from the fact that you can spend nearly a whole shift searching the randomly generated Z levels and not find anything due to RNG generation. No one really wants to spend the shift flying through empty space and find nothing. Even if it just speeds up the search process a little bit, I'd see this device as a welcome addition.

farie82:
It's not really required for our current space exploration. But I guess it's a nice to have and might bring in new PRs that expand space

AffectedArc07:
Minor gripe but this sprite is huge
image
Red = size of a regular tile. If this thing is placed on the floor its going to look like an entire console-sized machine, not just a scanner thing.

PidgeyThePirate:
The device is the size of an energy gun, it's not the size of a health analyzer. It should look a bit bulky.

Fox-McCloud:
I don't think ruins should be even easier to find, nor do I think we should be encouraging players to deliberately leave the station.

PidgeyThePirate:
I do not believe forcing RNG based tedium for the sake of balance is good design either. Genetics was awful to play, got that treatment and now it's more awful.

As to leaving the station, I'm of a mind that that's exactly what we should do at the moment, especially during highpop. The Cyberiad is already super overcrowded, and having some more interesting things to do off-station is one way to take some pressure off.

PidgeyThePirate:
Ok. Back to the drawing board then.

PR #13457, opened 2020-05-19
Updates the contribution info. Multiplications are not faster than divisions

farie82:
## What Does This PR Do
Updates the contribution info file to explain that divisions are not slower than multiplications. @Fox-McCloud and @Crazylemon64  tested this and in some cases divisions were faster than multiplications. 

No CL since it's a github only change

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -479,7 +479,7 @@ in the SQL/updates folder.
 
 * You are expected to help maintain the code that you add, meaning that if there is a problem then you are likely to be approached in order to fix any issues, runtimes, or bugs.
 
-* Do not divide when you can easily convert it to multiplication. (ie `4/2` should be done as `4*0.5`)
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
I'd just remove the line entirely; there's really no reason to specify either or when both are acceptable---when one is, though, that's when ya need to.

@Crazylemon64 @Anasari might want to provide their input though!

farie82:
> I'd just remove the line entirely; there's really no reason to specify either or when both are acceptable---when one is, though, that's when ya need to.
> 
> @Crazylemon64 @Anasari might want to provide their input though!

Aye I get your point. I left it in there to avoid people questioning whether to use `*` or `/` since in some languages one is preferred over the other.

I'll await their input before changing it.

Crazylemon64:
I concur with fox

farie82:
Sounds good then. I've changed the PR to just remove the line

PR #13459, opened 2020-05-19
Bluespace tile update

AffectedArc07:
## What Does This PR Do
This PR updates the sprite of bluespace floor tiles to still keep the bluespace-slime aesthetic (flashing green and sparks for whatever reason), but now the tiles dont look like they will electrocute you instantly. The tile sprite has also been updated

Before:
uuqMBZOaZq

After:
LZ3CJeLAHA

## Why It's Good For The Game
Currently, these tiles look like they are going to seriously harm you if walked on, which is discouraging to new players to say the least. The tiles should look legitimately helpful, not like deathtraps.

## Changelog
:cl: AffectedArc07
tweak: Bluespace floor tiles no longer look like they will shock you to death
/:cl:


AffectedArc07:
In a perfect world this would just be an overlay put onto existing tiles with a bluespace applicator which then gets a speedup component, which can then keep the look of the last tile but just have a small overlay, but I think thats out the scope of a sprite change, and afaik we dont currently have a speedup component.

Carthusia:
Is it possible to make these new tiles compatible with a floor painter?

AffectedArc07:
@Carthusia what I said above would work with that

McRamon:
Question, can animation be played only when people step on tile? I wish it was like that

Anyway, i have to say this sprite not great too

MRgaty:
I agree that the sprite is kinda ugly

AffectedArc07:
Making this into an item that can be applied to any floor tile on the station so the original aesthetic can be preserved

AffectedArc07:
Nevermind, I am already juggling enough projects to rewrite stack code to allow it to use an item when you click a turf

AffectedArc07:
Will deconflict when we reach a consensus of "do we want this or not"

AffectedArc07:
Disliked by majority, plus I have better plans for this.

PR #13462, opened 2020-05-20
Replaces the bloodcrawl variable with a trait

SteelSlayer:
## What Does This PR Do
Removes the `bloodcrawl` variable from mobs, and replace it with a trait instead. 

In a majority of rounds, `bloodcrawl` is never even used. Having this variable be attached to ALL living mobs is wasteful.

No CL because this is a back-end change.

### DIFF
--- a/code/__HELPERS/traits.dm
+++ b/code/__HELPERS/traits.dm
@@ -62,7 +62,9 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 
 //mob traits
 #define TRAIT_PACIFISM			"pacifism"
-#define TRAIT_WATERBREATH       "waterbreathing"
+#define TRAIT_WATERBREATH		"waterbreathing"
+#define TRAIT_BLOODCRAWL		"bloodcrawl"
+#define TRAIT_BLOODCRAWL_EAT	"bloodcrawl_eat"
 
 // common trait sources
 #define ROUNDSTART_TRAIT "roundstart" //cannot be removed without admin intervention
--- a/code/_globalvars/traits.dm
+++ b/code/_globalvars/traits.dm
@@ -5,7 +5,10 @@
 */
 GLOBAL_LIST_INIT(traits_by_type, list(
 	/mob = list(
-		"TRAIT_PACIFISM" = TRAIT_PACIFISM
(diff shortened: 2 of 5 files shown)


SteelSlayer:
Had to remove the mimery stuff from this because I don't know how to handle mime's gaining the trait upon getting cloned or other body-transferring effects, still learning traits.

code/_globalvars/traits.dm
```diff
@@ -5,7 +5,10 @@
 */
 GLOBAL_LIST_INIT(traits_by_type, list(
 	/mob = list(
-		"TRAIT_PACIFISM" = TRAIT_PACIFISM
+		"TRAIT_PACIFISM" = TRAIT_PACIFISM,
+		"TRAIT_WATERBREATH"	= TRAIT_WATERBREATH,
```
Fox-McCloud, line 9: Um.

code/__HELPERS/traits.dm
```diff
@@ -62,7 +62,9 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 
 //mob traits
 #define TRAIT_PACIFISM			"pacifism"
-#define TRAIT_WATERBREATH       "waterbreathing"
+#define TRAIT_WATERBREATH		"waterbreathing"
```
Fox-McCloud, line 65: uh!

SteelSlayer: Uh yes I can explain, the area between `TRAIT_WATERBREATH` and `"waterbreathing"` were all spaces and it looked ugly, so I converted it to using tabs :smile:

code/_globalvars/traits.dm
SteelSlayer: While I was adding bloodcrawl, I just decided to add waterbreath, because why not.

Fox-McCloud: but this isn't used anywhere!

SteelSlayer: If you mean the global list here, it's used in the same file a few times and in the 
`/datum/admins/proc/modify_traits(datum/D)` proc, which presumably allows admins to add/remove traits from mobs.

Fox-McCloud: nevermind, I was thinking something else *entirely*

PR #13463, opened 2020-05-20
Increase xray implant cost

TDSSS:
## What Does This PR Do
This PR increases the cost of manufacturing an x-ray implant. The new cost (Assuming an unupgraded protolathe with stock components, keep in mind this cost can be almost halved with high tier upgrades) is 2.5 sheets of uranium, 5 sheets of diamond, 5 bluespace crystals.

The old cost was 0.5 sheets of uranium, diamond and bluespace crystal, each. (Each sheet you feed the machine is equivalent to 2000 units)

Edit: After some coder chat discussion, the cost of thermal implants has been increased too, to 2 sheets of gold, 1.5 sheets of plasma, 2.5 sheets of diamond.

## Why It's Good For The Game
X-ray is probably the most powerful upgrade you can really get as crew. It's a lot of TC for nukies, an entire spellpoint for wizard, we've removed it as a genetics option...And yet its old cost was basically trivial. With the new amount of miners we get, the cost was laughably small once you've got the required research.

Security having wide access to x-ray is often credited with pushing cults off-station and can make things extremely unfair. Instead, giving multiple people xray should come at a real cost. I often see science build multiple mechs, so they're clearly not lacking in resources. if they really want to equip people with x-ray implants, they can, but it's going to cost them.

The increase to thermal is for the same reason, being able to see people through walls is still very powerful.

## Changelog
:cl:
tweak: Cost of x-ray implant massively increased.
tweak: Cost of thermal implant moderately increased.
/:cl:

### DIFF
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -385,7 +385,7 @@
 	req_tech = list("materials" = 6, "programming" = 4, "biotech" = 7, "magnets" = 5,"plasmatech" = 4)
 	build_type = PROTOLATHE | MECHFAB
 	construction_time = 60
-	materials = list(MAT_METAL = 600, MAT_GLASS = 600, MAT_SILVER = 600, MAT_GOLD = 600, MAT_PLASMA = 1000, MAT_DIAMOND = 2000)
+	materials = list(MAT_METAL = 600, MAT_GLASS = 600, MAT_SILVER = 600, MAT_GOLD = 4000, MAT_PLASMA = 3000, MAT_DIAMOND = 5000)
 	build_path = /obj/item/organ/internal/cyberimp/eyes/thermals
 	category = list("Medical")
 
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -374,7 +374,7 @@
 	req_tech = list("materials" = 7, "programming" = 5, "biotech" = 7, "magnets" = 5,"plasmatech" = 6)
 	build_type = PROTOLATHE | MECHFAB
 	construction_time = 60
-	materials = list(MAT_METAL = 600, MAT_GLASS = 600, MAT_SILVER = 600, MAT_GOLD = 600, MAT_PLASMA = 1000, MAT_URANIUM = 1000, MAT_DIAMOND = 1000, MAT_BLUESPACE = 1000)
+	materials = list(MAT_METAL = 600, MAT_GLASS = 600, MAT_SILVER = 600, MAT_GOLD = 600, MAT_PLASMA = 1000, MAT_URANIUM = 5000, MAT_DIAMOND = 10000, MAT_BLUESPACE = 10000)
 	build_path = /obj/item/organ/internal/cyberimp/eyes/xray
 	category = list("Medical")
 


trololiver112:
Even though i 100 percent agree on the idea of the implant being op as it is in rnd i don’t think this is going to solve the issues. 
I think removing the implant from rnd completely or giving it some drawbacks would be better.

An upgraded orm and miners with plasma cutter will still make this implant pretty easy to get, at least for the officers willing to get it.

procdrone:
Are you sure genetic xray is removed? Last i saw it had a very low chance of appearing, but it still did. I had genetic xray just a few days ago.

In my opinion XRAY should be just admin gene/implant, so not available to crew, and thermals reworked to be very expensive.

Xray is anti fun, and only 'required' in a very very limited situations - which thermals would cover anyway.

SteelSlayer:
X-ray is indeed still a genetics power and yeah, it’s got like a 15 or 20% to activate I believe.

AzuleUtama:
As much as I'd support any nerfs to xray given they just allow security to stop antags before they can even do anything - I don't think a material cost increase on its own is going to be enough. It really only takes 1 competent officer with xray to begin shutting down antags, since he can surprise attack / call them out incredibly easily - same with thermals. Science antags can also print themselves off the same set of implants and find a willing surgeon to usually become incredibly overpowered for what they need to do.

I agree with @procdrone 's suggestion of limiting Xrays to either admin intervention - but also keeping it for high-level antags only (Nukies & Wiz namely, I don't think full-power vamps should keep it if sec lose it as well).

Still, it's at least something to stop it being so easily spammable at R&D.

TDSSS:
Huh, I thought we replaced genetic x-ray with genetic thermal vision? Did that not actually go through?

TDSSS:
I've also seen people suggest that the tech requirements of xray should be increased. I could make it require materials 8 or biotech 8, which would make it impossible to get without abductors or a demon heart to decon. (If I also raise the tech origin of xray implants to match, it would allow you to also get xray if you can decon an xray, like from sol traders)

AzuleUtama:
> I could make it require materials 8 or biotech 8, which would make it impossible to get without abductors or a demon heart to decon.

IIRC, there's a bunch of abductor tools that can be found in a lavaland ruin that spawns most rounds - so not sure if that would be effective on it's own. Might be wrong though.

procdrone:
In case my previous suggestion finds support, thermals could have a detriment of being an active toggle power, which causes damage to your eyes while in use, which would balance it out for crew use. 

Wouldn't touch how antag thermals work.

MRgaty:
Costs too many diamonds IMO, you rarely get more than 20

TheMaveran:
Oh how I would love to see X-ray cost material/biotech 8, we know that probably wouldn't get merged. This is a good step in the right direction for sure though.

Spacemanspark:
Imo end game science stuff should be a bit more pick and choose, rather than 'hi yes I have all the infinity stones of science' and blazing through everything.

IE you should really watch what and who you're spending things on.
+1

Qwertytoforty:
Honestly, I would love the idea of X tech 8 stuff, however, considering tech 8 is like 1 in 100 rounds, it probably isn't worth it. More importantly, I don't think  any item we currently have should be moved to a supposed tech 8 level. I also don't think of an idea of late game science being, pick 3 of these 5 items to research, as it will only make people complain when something they want inst chosen, and if a meta for the 3 items appears, people will complain if someone did something different.

Denthamos:
Honestly in the end anything Xray wise, so vision and guns both should not be in the hands of the crew.  Vision is the second strongest tool/weapon after speed and being able to see everything through walls just ends the fun for rounds.  Its the equivalent of taking a paintball minigun to a paintball battle against a bunch of kids.  Fun for you but not fun for them.

As for xray guns, they have always been an issue against specific antagonists and remain a very strong tool against such.  Usually allowing players to completely bypass the antagonists attention and end their round fast.  I personally feel the xray guns should be in the gamma armory and not made from RnD.

ppi13:
This won't really change what makes x-ray good, you can still grind for it, one way or another at best, or just go genetics at worse, change how x-ray works by say, adding different levels of x-ray, like see through only one wall, can't see through reinforced walls, don't get night vision in addition to x-ray.

Spacemanspark:
Actually, on second thought, I agree with the idea of removing it from RnD outright.

Fox-McCloud:
Went with the alternative.

PR #13466, opened 2020-05-20
Fixes input being lost while holding keys down and pressing others

farie82:
TM requested if possible please to measure the actual performance gain and ofcourse to see if it fully works.

## What Does This PR Do
Makes the input system keep track of the keys held using an assoc list instead of an assoc list that reuses spots in the list once other keys are pressed.
This should in theory be a bit more performing as well since the `client/keyLoop` proc now doesn't have to loop through a filled list every tick. And instead it loops through about 0-2 entries on average.

This fixes movement suddenly stopping after you press shift a bunch of times. Or holding CTRL and spinning around making you move.

I've tested it locally for a while. Movement didn't feel weirder than normal

[EDIT]
I've tested it further using simulated input in a simple DM program. The input was generated by me playing on my local test server. I walked around to the brig to get guns. Shot up a vulp on the bridge and then proceeded to heal him. Took about 252 seconds total.
That input I then ran 6 times through both my suggested method and the current method using a simulation. 

Code used for testing: 
[Testing.zip](https://github.com/ParadiseSS13/Paradise/files/4671443/Testing.zip)

image
Stats of the tests.

## Why It's Good For The Game
The unresponsiveness of losing your movement input is quite annoying if you say examine a ton by pressing shift and then releasing while holding one movement key down. This fixes that.
The performance gain might also be something that's handy

## Changelog
:cl:
fix: Pressing buttons while holding a key down now won't stop the registration of the key hold down. Aka holding CTRL and moving around will now spin you forever and ever
/:cl:


Fox-McCloud:
I'd definitely ask around at the coderbus the impact of this, fully, but in theory, I strongly support this.

farie82:
I've asked ActionNinja about it, the author of the PR. 
image
I'm not sure how the performance compares here. If constantly looping over a list with empty entries is worse performing OR if constantly adding/removing assoc values from a list is worse performing.

@Fox-McCloud

AffectedArc07 (approved):
If this breaks we have revert buttons

PR #12345, opened 2019-09-18
MAP ROTATION PART 2

AffectedArc07:
## What Does This PR Do
Adds map rotation. This exists as a vote for players when the shuttle leaves the station (On a config option), and gives admins a verb to set the map for the next round. The default map is Cyberiad, and the "Set Next Map" verb only affects the next round, after that we go back to boxstation. 

This also introduces a benefit for people testing maps. You no longer need to recompile to test station remaps, just restart your server. This does come at the cost of an init time penalty, but its not hugely significant, and it will only go down once I kill off `/New()` further.

Yes I know this PR took so long to do, but atleast its done now, and we claimed 12345 for something good other than an insta-close meme PR 

## Why It's Good For The Game
Ability to switch maps with no external tools is a godsend

## Changelog
:cl: AffectedArc07
add: Map Rotation
/:cl:

PR #13467, opened 2020-05-20
Vendor fix

### LINKED ISSUE #13389: Vendor Account gains double/triple what you pay at a vendor
**Issue Description**:
Comparing account balances before and after purchase at a an Account Uplink Terminal (there's one between the HoP office and conference room), the Vendor account gain more cash from vendor transactions than it should.

For the vendor account to get the amount of cash that was paid, and not more.

**What happened instead**:
If you pay with your ID, it gets triple the cash.
If you pay with hard cash, it gets double.

Makes no sense and allows unintended illegitimate money duplication (even though the method sucks).

Note the cash balance of the Vendor account (starts at 5000).
Go purchase something from a vendor at price X.
Check the balance of the Vendor account once more, and note that it increased by an amount greater than X.

N/A

**Extra information**:
Checking the transactions log for the Vendor account reveals that ID purchases have two transaction logs instead of one. Cash purchases produce only one log. I assume there's one more hidden transaction happening. Checking the transaction account for your ID shows one log for payment.

tinfoiltophat:
## What Does This PR Do
Fixes #13389 by making sure that purchases from vendors are only credited to the vendor account one time. 

## Why It's Good For The Game
Previously, each purchase was being credited two or three times to the vendor account, although only charging the user once. This could have been used maliciously to dupe money as the HoP or anyone able to access the vendor account.

## Changelog
:cl: Tinfoiltophat
fix: Vendors will no longer be credited more money than they received for each purchase.
/:cl:

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -404,7 +404,8 @@
 	cashmoney.use(currently_vending.price)
 
 	// Vending machines have no idea who paid with cash
-	credit_purchase("(cash)")
+	GLOB.vendor_account.credit(currently_vending.price, "Sale of [currently_vending.name]",
+	name, "(cash)")
 	return 1
 
 /**
@@ -445,25 +446,17 @@
 	else
 		// Okay to move the money at this point
 		var/paid = customer_account.charge(currently_vending.price, GLOB.vendor_account,
+		customer_account.charge(currently_vending.price, GLOB.vendor_account,
 			"Purchase of [currently_vending.name]", name, GLOB.vendor_account.owner_name,
 			"Sale of [currently_vending.name]", customer_account.owner_name)
 
-		if(paid)
-			// Give the vendor the money. We use the account owner name, which means
-			// that purchases made with stolen/borrowed card will look like the card
-			// owner made them
-			credit_purchase(customer_account.owner_name)
-		return paid
+		return 1
 
 /**
  *  Add money for current purchase to the vendor account.
  *
  *  Called after the money has already been taken from the customer.
  */
-/obj/machinery/vending/proc/credit_purchase(var/target as text)
-	GLOB.vendor_account.money += currently_vending.price
-	GLOB.vendor_account.credit(currently_vending.price, "Sale of [currently_vending.name]",
-	name, target)
 
 /obj/machinery/vending/attack_ai(mob/user)
 	return attack_hand(user)
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -404,8 +404,7 @@
 	cashmoney.use(currently_vending.price)
 
 	// Vending machines have no idea who paid with cash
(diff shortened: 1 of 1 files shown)


tinfoiltophat:
First time submitting a PR through git, hope I'm doing this properly.

SteelSlayer:
Grats on your first PR. Haven't looked at the code, but yeah, seems you did everything right.

For fixes though, you want to right "Fixes #12345". Replace the number with whatever issue you're trying to fix obviously. But if you write it like that, when this PR gets merged, it will automatically close that issue.

tinfoiltophat:
> Grats on your first PR. Haven't looked at the code, but yeah, seems you did everything right.
> 
> For fixes though, you want to write "Fixes #13389". If you write it like that, when this PR gets merged, it will automatically close that issue.

Thank you. I have edited my post with the proper syntax.

code/game/machinery/vending.dm
```diff
@@ -404,7 +404,8 @@
 	cashmoney.use(currently_vending.price)
 
 	// Vending machines have no idea who paid with cash
-	credit_purchase("(cash)")
+	GLOB.vendor_account.credit(currently_vending.price, "Sale of [currently_vending.name]",
+	name, "(cash)")
```
farie82, line 408: Put this on the same line as 407. Or indent it.

code/game/machinery/vending.dm
```diff
@@ -444,26 +445,17 @@
...
 			"Purchase of [currently_vending.name]", name, GLOB.vendor_account.owner_name,
 			"Sale of [currently_vending.name]", customer_account.owner_name)
 
-		if(paid)
-			// Give the vendor the money. We use the account owner name, which means
-			// that purchases made with stolen/borrowed card will look like the card
-			// owner made them
-			credit_purchase(customer_account.owner_name)
-		return paid
+		return 1
```
farie82, line 452: `TRUE` here. Boolean values should always be denoted as `TRUE/FALSE`. Old code still uses `1` or `0`. Which TRUE or FALSE map to btw.

code/game/machinery/vending.dm
```diff
@@ -444,26 +445,17 @@
...
 
 /**
```
farie82, line 454: Remove the comment here about the removed proc as well.

tinfoiltophat:
That should be all the changes you advised. Thank you for your feedback.

code/game/machinery/vending.dm
```diff
@@ -444,26 +444,12 @@
 		return 0
 	else
 		// Okay to move the money at this point
-		var/paid = customer_account.charge(currently_vending.price, GLOB.vendor_account,
```
Crazylemon64, line 447: Where will the sale be credited? And what happens if payment fails?

tinfoiltophat: Payment is credited to the Vendor account. Previous checks ensure that the payment will  be go through, but if it somehow managed to fail anyway the user would still receive the item.

AffectedArc07 (approved):

PR #13470, opened 2020-05-21
A few small bugfixes

### LINKED ISSUE #13318: Money can be withdrawn from and deposited to a frozen account using ATM
**Issue Description**:
After getting my ingame bank account frozen i was still able to withdraw all my money, without the balance going down. I was also able to add the credits back into the account, but that would increase my balance. Thus giving me infinite money in theory

That i would not be able to do anything with a frozen account

**What happened instead**:
I was able to generate infinite money

It breaks the element of having your account frozen, and allows you to flood the station with cash. 

Get your account frozen
Withdraw all your money
Deposit the money
Repeat until desired amount of money is achieved

as soon as i got my account frozen, unsure if this is a bug that has always been here. 

**Extra information**:

### LINKED ISSUE #13377: While being in a vehicle you can be hit twice by the same projectile.
**Issue Description**:

As stated in the title you can be hit twice by the same projectile in a vehicle.

Only be hit once.

**What happened instead**:

I got hit twice by the same projectile.

If you are riding a vehicle, you shouldn't be forced to take two hits from a single projectile. I discovered this when being shot by disablers and it would be much worse if I were to be hit by buckshot, slugs or even 357., which would drop me very quickly if not instantly.

Step 1. Step into a vehicle.
Step 2. Get shot while moving towards or standing still in said vehicle.
Step 3. Receive two attack procs from a single projectile.
Step 4. Cry because you received double the damage.

No idea once again since I bet it been around for a long time and noone noticed it due to the only people using vehicles if the janitor and paramedic.

**Extra information**:

I think it procs from moving forward towards a projectile while in a vehicle.

### LINKED ISSUE #13458: Cursed Clown/Mime Mask (Not to be confused with cluwne) is not functional internals
**Issue Description**:

The masks don't work as internals like the original ones do which causes issues for races that get turned into clowns or mimes via magic who can't take off their masks, becoming unable to breathe. Mainly Vox have this issue since they are the majority that get hit by this.

I expected the masks to work normally and not make them suffocate to death.

**What happened instead**:

They suffocated on the O2 in the air giving them Toxin and O2 Damage which probably took them out of the round.

Mimeify and Clownify Wizard spell kills certain races who can't easily get acid that can melt masks off.
I don't think this was intended and was ment to be harmless for the most part.

1. Use the mimeify/clownify spell on a vox or plasmaman
2. Watch them slowly die on the O2 in the air.

Not sure when these spells were added. Might of started before vox and plasmaman were added? No idea when they got added though and I'm not certain when the issue started.

**Extra information**:

I think the spell might also remove plasmaman enviromental protective suits too.

tinfoiltophat:
## What Does This PR Do
Fixes #13458 - Nodrop clown and mime masks(from banana touch and mime malaise respectfully) are now compatible with internals. This also adds specialized nodrop clown and mime gear for plasmamen affected by the spell.
Fixes #13377 - While sitting in a vehicle and hit by a projectile, you will now only be hit once by each projectile. 
Fixes #13318 - Suspended bank accounts can no longer withdraw money or make transfers.

Unrelated - Empty vehicles will now actually show icons(such as laser blasts etc) when hit by weapon fire. This is purely cosmetic and does not make the vehicles take damage. 

## Why It's Good For The Game
Banana touch and mime malaise would previously result in the death of species that require internals, particularly plasmamen who would be unable to wear their envirosuits once affected by this spell. This was likely unintended behavior, and the spell is now nonlethal to all species.

Double damage from projectiles while in a vehicle could ruin rounds for a lot of people. The addition of projectile hit icons when a vehicle has been hit also looks a little nicer than before.

Suspended bank accounts were previously able to generate infinite money. We're not commies, you work for your paycheck.

## Changelog
:cl: Tinfoiltophat
fix: Banana touch and Mime malaise spells will no longer kill plasmamen and vox, cursed masks are now compatible with internals.
fix: Vehicle riders will no longer be hit twice by the same projectile.
fix: Suspended bank accounts can no longer be used for infinite money.
/:cl:


Ionward:
I like the vehicle blasting bit, never even would've thought of it.

tinfoiltophat:
I thought about taking the opportunity to make vehicles destructible, but it's probably for the best if I keep bugfix PRs and feature PRs separate.

code/datums/spells/banana_touch.dm
```diff
@@ -44,15 +44,25 @@
 		unEquip(wear_mask, TRUE)
 		unEquip(head, TRUE)
 		unEquip(wear_suit, TRUE)
-		equip_to_slot_if_possible(new /obj/item/clothing/head/wizard/clown, slot_head, TRUE, TRUE)
+		if(src.dna.species.name == "Plasmaman")
```
Fox-McCloud, line 47: Come on dude, you've been around long enough to know that `src` is always implied.

tinfoiltophat: Lmao
I'm not familiar with whatever that second thing is. Mind pointing me towards the proc name and I'll edit it in there?

Kyep: Guessing that what Fox is suggesting is something like:
if(is_species(src, /datum/species/plasmaman))
https://github.com/ParadiseSS13/Paradise/blob/master/code/__HELPERS/mobs.dm#L460

tinfoiltophat: Good call; it's changed. Thank you guys.

code/datums/spells/mime_malaise.dm
```diff
@@ -38,16 +38,26 @@
...
+			equip_to_slot_if_possible(new /obj/item/clothing/under/mime, slot_w_uniform, TRUE, TRUE)
 		equip_to_slot_if_possible(new /obj/item/clothing/mask/gas/mime, slot_wear_mask, TRUE, TRUE)
-		equip_to_slot_if_possible(new /obj/item/clothing/under/mime, slot_w_uniform, TRUE, TRUE)
 		equip_to_slot_if_possible(new /obj/item/clothing/suit/suspenders, slot_wear_suit, TRUE, TRUE)
 		Silence(7)
 	else
 		qdel(wear_mask)
 		qdel(w_uniform)
 		qdel(wear_suit)
+		if(src.dna.species.name == "Plasmaman")
```
Fox-McCloud, line 54: 😠

code/modules/clothing/spacesuits/plasmamen.dm
```diff
@@ -235,13 +235,19 @@
 	item_state = "mime_envirohelm"
 	visor_icon = "mime_envisor"
 
+/obj/item/clothing/head/helmet/space/plasmaman/mime/nodrop
+	flags = NODROP
```
Fox-McCloud, line 239: make sure this doesn't create the same problem as the original `nodrop` varients of the mime and clown mask had.

code/modules/vehicle/vehicle.dm
```diff
@@ -138,10 +138,9 @@
 	handle_vehicle_offsets()
 
 /obj/vehicle/bullet_act(obj/item/projectile/Proj)
-	if(has_buckled_mobs())
-		for(var/m in buckled_mobs)
-			var/mob/living/buckled_mob = m
-			buckled_mob.bullet_act(Proj)
```
Fox-McCloud, line 144: You can have more than one mob buckled to a vehicle, in theory.

As for what is written here---I'm not sure why it would rectify the issue unless bullets are calling `Bump` multiple times...

tinfoiltophat: What happened previously was that this section of code made the projectile hit each/any passenger, but due to recent vehicle changes it became redundant and caused projectiles to hit passengers twice.

code/datums/spells/mime_malaise.dm
tinfoiltophat: Most recent commit should've fixed this. Not sure why it's still showing up in this format.

tinfoiltophat: Oh, I'm dumbass.

code/modules/clothing/spacesuits/plasmamen.dm
tinfoiltophat: It doesn't.

code/datums/spells/banana_touch.dm
```diff
@@ -44,15 +44,25 @@
...
 		equip_to_slot_if_possible(new /obj/item/clothing/suit/wizrobe/clown, slot_wear_suit, TRUE, TRUE)
 		equip_to_slot_if_possible(new /obj/item/clothing/shoes/clown_shoes/magical, slot_shoes, TRUE, TRUE)
 		equip_to_slot_if_possible(new /obj/item/clothing/mask/gas/clownwiz, slot_wear_mask, TRUE, TRUE)
 	else
 		qdel(shoes)
 		qdel(wear_mask)
 		qdel(w_uniform)
-		equip_to_slot_if_possible(new /obj/item/clothing/under/rank/clown/nodrop, slot_w_uniform, TRUE, TRUE)
+		if(is_species(src, /datum/species/plasmaman))
+			qdel(head, TRUE)
```
farie82, line 61: I doubt the qdel has to be forced here.

code/datums/spells/mime_malaise.dm
```diff
@@ -38,16 +38,26 @@
...
 		equip_to_slot_if_possible(new /obj/item/clothing/mask/gas/mime, slot_wear_mask, TRUE, TRUE)
-		equip_to_slot_if_possible(new /obj/item/clothing/under/mime, slot_w_uniform, TRUE, TRUE)
 		equip_to_slot_if_possible(new /obj/item/clothing/suit/suspenders, slot_wear_suit, TRUE, TRUE)
 		Silence(7)
 	else
 		qdel(wear_mask)
 		qdel(w_uniform)
 		qdel(wear_suit)
+		if(is_species(src, /datum/species/plasmaman))
+			qdel(head, TRUE)
```
farie82, line 55: Same as the previous file. Doubt the qdel has to be forced

code/modules/clothing/masks/gasmask.dm
```diff
@@ -159,7 +159,7 @@
 	magical = TRUE
 
 /obj/item/clothing/mask/gas/mime/nodrop
-	flags = NODROP
+	flags = NODROP | BLOCK_GAS_SMOKE_EFFECT | AIRTIGHT
```
farie82, line 162: The normal ones don't have these flags. Calling this path nodrop won't convey the true changes this version has compared to the default one. Default one only has `FLAMMABLE`.

code/modules/clothing/spacesuits/plasmamen.dm
farie82: It will since the flags get overriden here

code/modules/clothing/spacesuits/plasmamen.dm
```diff
@@ -235,13 +235,19 @@
...
+
 /obj/item/clothing/head/helmet/space/plasmaman/clown
 	name = "clown envirosuit helmet"
 	desc = "The makeup is painted on, it's a miracle it doesn't chip. <i>'HONK!'</i>"
 	icon_state = "clown_envirohelm"
 	item_state = "clown_envirohelm"
 	visor_icon = "clown_envisor"
 
+/obj/item/clothing/head/helmet/space/plasmaman/clown/nodrop
+	flags = NODROP
```
farie82, line 249: Same applies here

code/modules/clothing/masks/gasmask.dm
tinfoiltophat: That's weird. I must have assumed the mime mask had the same flags as the clown mask. A little strange that it doesn't, actually. Iirc AIRTIGHT is required for internals, so AIRTIGHT should probably stay on regardless to avoid any vox or plasmaman shenanigans. Will make the changes when I have a moment.

variableundefined (changes requested):
@tinfoiltophat Please address change requests from Farie in a week

variableundefined:
Stale for over a month. Closing.

PR #13474, opened 2020-05-22
Adds Service Door Between Kitchen and Botany

Fox-McCloud:
A small map edit--inspired by TG.

Adds a service door between botany and the kitchen--access requirement is EITHER botany or kitchen.

That said, the chef has a shutter door button on his side to just straight up block the door off so botanist can't enter.

If there's ever time where the botanist/chef are SSD/just not there, this allows the other to potentially make a thing or two and not be completely stuck (especially for chef), while still letting the more authoritative role (chef) have control over who enters his kitchen.

img

:cl: Fox McCloud
add: Adds new service door between botany and the kitchen
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -30093,25 +30093,26 @@
 /turf/simulated/floor/plating,
 /area/chapel/main)
 "bcS" = (
-/obj/structure/table/reinforced,
 /obj/machinery/door/firedoor,
-/obj/machinery/door/window/eastleft{
-	name = "Hydroponics Desk";
-	req_access_txt = "35"
+/obj/machinery/door/airlock/public/glass{
+	name = "Service Door";
+	req_access_txt = "0";
+	req_one_access_txt = "35;28"
 	},
-/obj/machinery/door/window/northleft{
-	dir = 8;
-	icon_state = "left";
-	name = "Kitchen Desk";
-	req_access_txt = "28"
+/obj/machinery/door/poddoor/shutters{
+	density = 0;
+	dir = 2;
+	icon_state = "open";
+	id_tag = "servicedoorkitchen";
+	name = "Service Shutters";
+	opacity = 0
 	},
-/obj/effect/decal/warning_stripes/yellow,
 /turf/simulated/floor/plasteel{
 	dir = 2;
 	icon_state = "cafeteria";
 	tag = "icon-cafeteria (NORTHEAST)"
 	},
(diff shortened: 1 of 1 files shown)


EMPlight:
I can understand the convenience of the door but I think its a bad idea.

**Undermines the independence of the departments.** The chef's space belongs to the chef and botany's space belongs to botany. You'll have a working botanist and the chef will insist on growing and managing their own food. You'll have a very busy chef and all botany does is cook their own food and get in the way.

**Undermines the HOP’s job.** The HOP’s job is to give out and remove access. They’re suppose to be the approval and authority check for who works where. If a botanist or a chef really want kitchen or hydro access they’ll have to do a little RP with the HOP and work for it.

There are tools in the back of hydro, that are QoL stuff, but the chef doesn’t need them.

But in reality...

If you’re playing either job and really need access you can just un-wrench the SmartFridge.

taukausanake:
I kind of agree with @EMPlight. You shouldn't give chef permission into Botany. Worse case senario the chef can order seeds from cargo and use their little garden north of the kitchen to grow the food Botany should be growing. You shouldn't just give them free access to a whole different area

MRgaty:
I like this idea, I like to see botany less seen as chef's bitches and more as coworkers.
I also dont think letting the chef grow their stuff is bad, they already have the garden to grow their stuff so they just have easier access to the vendors plants.

Quantum-M:
Why not make the shutters start closed, since botanists has no real reason to be in the kitchen? Or alternatively give the door only kitchen access and the shutter control to the botanists, as to keep out any unruly chefs when botany is actually competent?

I strongly disagree, as you have to take in account that the chef will be able to access the botany tools with ease. I also agree with @EMPlight, as it's not usually difficult to find other ways to get into the botany, such as asking the HoP, getting the AI to set emergency access/bolt the airlocks open through the upper parts of the kitchen and botany, or even getting the AI to open the windoors and using tools/getting a engineer to remove the table, or even removing the fridge.

I also disagree on botany getting access to the kitchen, as they have no reason to be in there at all, their main job being growing plants, not cooking them.

Carthusia:
All in all, i dont mind a door there, but having botanists walking into my kitchen while im trying to work would be a pain. More often than them needing into my kitchen, I need into their room to grab some bug spray, or their not doing their job and i desperately need chilis or something.

Pre-this door: Id just wrench the fridge and push my way in, a door with chef access would be nice, but i dont like the botanists also having access, shutter or no shutter.

Fox-McCloud:
Maybe in the future?

PR #13480, opened 2020-05-22
remove stupid message

TDSSS:
## What Does This PR Do
removes stupid message that gets spammed pretty much every round and nobody ever asks them about this.

## Why It's Good For The Game
Stops confusing new admins that might actually do what the message says.

## Changelog
:cl:
del: spammy useless message
/:cl:

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -11,7 +11,6 @@
 	if(client || registered_z) // This is a temporary error tracker to make sure we've caught everything
 		var/turf/T = get_turf(src)
 		if(client && registered_z != T.z)
-			message_admins("[src] [ADMIN_FLW(src, "FLW")] has somehow ended up in Z-level [T.z] despite being registered in Z-level [registered_z]. If you could ask them how that happened and notify the coders, it would be appreciated.")
 			log_game("Z-TRACKING: [src] has somehow ended up in Z-level [T.z] despite being registered in Z-level [registered_z].")
 			update_z(T.z)
 		else if (!client && registered_z)


Fox-McCloud:
This needs to stay, otherwise the problems with this system are *NEVER* going to get rectified.

They've been resolved in the past because of this message--and there's not a ton of them remaining; only a few. 

We *need* this to be able to diagnose the problem---a simple runtime error won't solve the situation since it doesn't capture the contextual issue of moving/how they were moving as it's just a "great, so and so caused this---zero context".

nok-ko:
Would be good to remove it for wizard teleports, at least.

TDSSS:
So, it's actually needed? When I became an admin people told me to ignore it. Yesterday in VC with kyet he seemed to think it's useless too. I guess I'll actually investigate people if this crops up then.

@Fox-McCloud Would you want the reasons submitted via an issue report or just discord, or...?

Fox-McCloud:
Yeah he's...not correct at all.

So, we have a system that tracks people based on the z-level they're on. It's largely used to determine if simple mobs on a z-level need to "wake up" and activate, or just stay dormant--it's a big performance saver, to say the least.

Either case, how this is accomplished is by properly calling `Move` so the transition registers---we have a lot of bad/old code that just sets a mob's `loc` directly---when that happens, and they change z-levels, it messages admins---and that's why it can't really be a runtime error. Movement is so contextual and specific that a generic runtime error wouldn't mean anything.

So yeah, it should be made into an issue report and added to every time it happens.

Reports can't be vague either, they need to be highly specific

"I crossed over the z-level and got htis message"

it should be

"I crossed over the z-level while inside a locker that was held by a mech that moved across the z-level"

"I was the passenger in a space pod when was transitioned"

"I was a diona nymph who was being held by someone in a backpack while we crossed over".

Otherwise, it'll be really difficult to determine where the offending bad code is.

TDSSS:
Understood, I'll keep my eyes open for that then.

Fox-McCloud:
Closing this in favor of just better collection of the issues on the github.

PR #13481, opened 2020-05-23
Makes explicit paths in the code/datums/periodic_news.dm file

trololiver112:
## What Does This PR Do

Adds more explicit paths in the `code/datums/periodic_news.dm` file making it cleaner.

## Why It's Good For The Game

Cleaner code and also makes it easier to work with as well as find.

### DIFF
--- a/code/datums/periodic_news.dm
+++ b/code/datums/periodic_news.dm
@@ -1,135 +1,123 @@
 // This system defines news that will be displayed in the course of a round.
 // Uses BYOND's type system to put everything into a nice format
 
-// THIS ISNT PROPERLY PATHED AND I AM GOING TO FUCKING SCREAM AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
 /datum/news_announcement
-	var
-		round_time // time of the round at which this should be announced, in seconds
-		message // body of the message
-		author = "Nanotrasen Editor"
-		channel_name = "Nyx Daily"
-		can_be_redacted = 0
-		message_type = "Story"
 
-	revolution_inciting_event
+	var/round_time // time of the round at which this should be announced, in seconds
+	var/message // body of the message
+	var/author = "Nanotrasen Editor"
+	var/channel_name = "Nyx Daily"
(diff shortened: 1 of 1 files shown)


SteelSlayer:
Oh my god that old format is the most disgusting thing I've ever seen.

Fox-McCloud:
This doesn't need a changelog.

We shouldn't really be having changelogs for things that don't impact players; the changelog is somethign that's primarily meant for player impacting things so they know what has/hasn't change in their game.

trololiver112:
Okay, i removed the changelog

code/datums/periodic_news.dm
```diff
@@ -1,135 +1,123 @@
 // This system defines news that will be displayed in the course of a round.
 // Uses BYOND's type system to put everything into a nice format
 
-// THIS ISNT PROPERLY PATHED AND I AM GOING TO FUCKING SCREAM AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```
farie82, line 4: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

code/datums/periodic_news.dm
```diff
@@ -1,135 +1,123 @@
...
-			We have little to trade with them now\". Nanotrasen officials have denied these allegations, calling them
-			\"further proof\" of the colony's anti-Nanotrasen stance. Meanwhile, Refuge Security has been unable to quell
-			the riots. More on this at 6."}
-			round_time = 60 * 60
 
+	var/round_time // time of the round at which this should be announced, in seconds
+	var/message // body of the message
+	var/author = "Nanotrasen Editor"
+	var/channel_name = "Nyx Daily"
+	var/can_be_redacted = 0
```
farie82, line 10: FALSE

code/datums/periodic_news.dm
trololiver112: AAAAAAAAAAAAAAAAAAAAAAAAAAAA

AffectedArc07:
Remove the :cl: blocks completely, because the generator may break here

PR #13485, opened 2020-05-24
Fixes the explosion distance on proc/explosion()

EMPlight:
## What Does This PR Do
Fixes the explosion distance for proc/explosion().
The distance given for explosions was reduced by one and would not cause the proper explosion effects for the given distance specified.

explosion(loc, 1, 2, 4, flame_range = 2)
The values 1, 2, and 4 are suppose to call levels of  ex_act()  on turfs that are a distance of 1, 2, and 4 tiles from the given location, the same goes with flame_range. These ex_act() effects would only reach tiles at range of 0, 1, and 3.

## Why It's Good For The Game
It would make the distances for the explosions work as intended.
However...
**All explosions using proc/explode() would get a range bump of 1 to it's ex_act() calls.**
Making something like the syndie minibomb and pizza bomb more lethal than what's in the game now.

## Changelog
:cl:
fix: proc/explosion
/:cl:

### DIFF
--- a/code/game/objects/explosion.dm
+++ b/code/game/objects/explosion.dm
@@ -99,12 +99,12 @@
 			var/flame_dist = 0
 //			var/throw_dist = max_range - dist
 
-			if(dist < flame_range)
+			if(dist <= flame_range)
 				flame_dist = 1
 
-			if(dist < devastation_range)		dist = 1
-			else if(dist < heavy_impact_range)	dist = 2
-			else if(dist < light_impact_range)	dist = 3
+			if(dist <= devastation_range)		dist = 1
+			else if(dist <= heavy_impact_range)	dist = 2
+			else if(dist <= light_impact_range)	dist = 3
 			else 								dist = 0
 
 			//------- TURF FIRES -------


Fox-McCloud:
This is going to have insanely dramatic impacts on most of explosion code, and I'm not inclined to merge this or consider it a fix, at all.

Explosion have *always* worked with the assumption that a devastation of `1` means that the explosion *only* get applied to the turf that it takes place on (and that's exactly what happens); a range of 2 means that it applies to the turf it's on and extends to one turf over.

It's a radius, and you have to count the turf that it starts on.

datlo:
This is working as intended. Explosion of size 1 is just the tile it's on... Not an issue to fix.

variableundefined:
As per Fox's comment, this is working as intended.

EMPlight:
> This is working as intended. Explosion of size 1 is just the tile it's on... Not an issue to fix.

It's a distance though. The game measures a distance of 0 as the tile an object is on. There's even code where negative numbers are used to remove the more severe ex_act().

`explosion(loc, -1, -1, 2, flame_range = 4)	// small explosion, plus a very large fireball.`

EMPlight:
> This is going to have insanely dramatic impacts on most of explosion code, and I'm not inclined to merge this or consider it a fix, at all.

Your right, as is, this PR would through game balance off for explosions. More work would have to be done updating all the calls that use the explosion proc. It would mean a lot of data re-entry and testing, quite the monotonous task.

This was suppose to be submitted as a bug issue for discussion but git wont let me submit bugs.

> Explosion have _always_ worked with the assumption that a devastation of `1` means that the explosion _only_ get applied to the turf that it takes place on (and that's exactly what happens); a range of 2 means that it applies to the turf it's on and extends to one turf over.

It would still be the same system, just the numbers would shift.

> It's a radius, and you have to count the turf that it starts on.

If you did any range calculations you would always have to add a +1.

Issue #13093, opened 2020-03-08
You can't smash down airlocks with objects while the airlock's panel is open


SteelSlayer:
**Issue Description**:
You can't smash down airlocks with objects while the airlock's panel is open.

**Why is this bad/What are the consequences**:
After bashing a door for a while, its panel opens due to the damage. At this point you can no longer break the airlock without closing the panel between every hit.

**Steps to reproduce the problem**:
1. Take a good melee weapon, like a spear or esword
2. Switch to harm intent and click on the door until its damged enough so its panel opens
3. Observe you cannot click on it anymore while the panel is open

**When did the problem start happening**:
Likely related to the tool refactor #11700

Citinited:
I'll take a look into it, and include a fix in #13092 if it was indeed to do with the tool refactor (probably was)

DaveKorhal:
This is still an issue. Tried taking a Dsword to airlocks today, and once it was damaged enough that the panel opened, the game wouldn't let me attack the door. At all. Nothing would happen when I clicked it.

PR #13488, opened 2020-05-25
Makes airlocks destroyable again using weapons when the panel is open

farie82:
## What Does This PR Do
Makes you able to destroy airlocks when their panel is open.
Fixes: #13093 

## Why It's Good For The Game
Less bugs more mugs

## Changelog
:cl:
fix: You can destroy airlocks again when the panel is open
/:cl:


rapaskoti:
i don't come here often but i so

PR #13489, opened 2020-05-25
Improves Hacker and Payday uplink bundles

AzuleUtama:
## What Does This PR Do
Adds some more value to both the Payday and Hacker bundles you can roll when buying a Syndicate Bundle as a traitor.

## Why It's Good For The Game
Both of these bundles currently don't feel as worthwhile as the other bundles you can potentially get from buying a Syndi Bundle as a traitor.

For Payday, I've removed the C4 since one block isn't enough to blow open the vault anymore and there are much better ways to get inside it. The drill itself has been replaced by the diamond one since on the rare chance you actually get to use it - you may as well have the better one. I've also given the kit a radio jammer to help delay security's arrival to wherever you've forced your way into.

For Hacker, the main item in the kit can be rendered useless if the AI doesn't exist, isn't very good or is malfunctioning and just chooses to ignore the hacked law entirely. The C4 kit also doesn't spawn since it's missing a src tag, though it doesn't really seem to fit a bundle focused on hacking anyway. This means you're left with about 9-14TC worth of items should hacking the AI be a lost cause (depends on your use of the binary key).
I've updated the kit to include a bunch more items both for assisting in hacking the AI as well as having use if the AI isn't helping or responding. It now comes with some thermal glasses for wall-hacking, a hacker's toolbelt containing a full toolset, AI detector instead of multitool and an EMP flashlight and some insulated gloves to go with said toolset. This should help the hacker bundle better match the value of others.

I've also updated some of the commented values since a few were wrong.

## Changelog
:cl:
tweak: Improves value of Hacker and Payday syndicate bundles.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/belt.dm
+++ b/code/game/objects/items/weapons/storage/belt.dm
@@ -308,6 +308,20 @@
 	item_state = "utility"
 	use_item_overlays = 1 // So it will still show tools in it in case sec get lazy and just glance at it.
 
+/obj/item/storage/belt/military/traitor/hacker
+	name = "tool-belt"
+
+/obj/item/storage/belt/military/traitor/hacker/New()
+	..()
+	new /obj/item/screwdriver(src, "red")
+	new /obj/item/wrench(src)
+	new /obj/item/weldingtool/largetank(src)
+	new /obj/item/crowbar/red(src)
+	new /obj/item/wirecutters(src, "red")
+	new /obj/item/multitool/ai_detect(src)
+	new /obj/item/flashlight/emp(src)
+	update_icon()
+
 /obj/item/storage/belt/grenade
 	name = "grenadier belt"
 	desc = "A belt for holding grenades."
--- a/code/game/objects/items/weapons/storage/belt.dm
+++ b/code/game/objects/items/weapons/storage/belt.dm
@@ -309,7 +309,6 @@
 	use_item_overlays = 1 // So it will still show tools in it in case sec get lazy and just glance at it.
 
 /obj/item/storage/belt/military/traitor/hacker
-	name = "tool-belt"
 
 /obj/item/storage/belt/military/traitor/hacker/New()
 	..()
@@ -318,8 +317,7 @@
 	new /obj/item/weldingtool/largetank(src)
 	new /obj/item/crowbar/red(src)
 	new /obj/item/wirecutters(src, "red")
-	new /obj/item/multitool/ai_detect(src)
-	new /obj/item/flashlight/emp(src)
+	new /obj/item/stack/cable_coil(src, 30, COLOR_RED)
 	update_icon()
 
 /obj/item/storage/belt/grenade
--- a/code/game/objects/items/weapons/storage/uplink_kits.dm
+++ b/code/game/objects/items/weapons/storage/uplink_kits.dm
@@ -90,8 +90,10 @@
 			new /obj/item/encryptionkey/binary(src) // 5TC
 			new /obj/item/aiModule/toyAI(src) // 0TC
 			new /obj/item/clothing/glasses/chameleon/thermal(src) // 6TC
-			new /obj/item/storage/belt/military/traitor/hacker(src) // 5TC
-			new /obj/item/clothing/gloves/color/yellow(src) // 0TC
+			new /obj/item/storage/belt/military/traitor/hacker(src) // 3TC
+			new /obj/item/clothing/gloves/combat(src) // 0TC
+			new /obj/item/multitool/ai_detect(src) // 1TC
+			new /obj/item/flashlight/emp(src) // 2TC
 			return
 
 		if("darklord") // 24TC + two 0TC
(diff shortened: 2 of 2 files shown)


code/game/objects/items/weapons/storage/belt.dm
```diff
@@ -308,6 +308,20 @@
 	item_state = "utility"
 	use_item_overlays = 1 // So it will still show tools in it in case sec get lazy and just glance at it.
 
+/obj/item/storage/belt/military/traitor/hacker
+	name = "tool-belt"
```
datlo, line 312: Unnecessary var, this is already set by the parent

code/game/objects/items/weapons/storage/belt.dm
```diff
@@ -308,6 +308,20 @@
...
+
+/obj/item/storage/belt/military/traitor/hacker/New()
+	..()
+	new /obj/item/screwdriver(src, "red")
+	new /obj/item/wrench(src)
+	new /obj/item/weldingtool/largetank(src)
+	new /obj/item/crowbar/red(src)
+	new /obj/item/wirecutters(src, "red")
+	new /obj/item/multitool/ai_detect(src)
+	new /obj/item/flashlight/emp(src)
```
datlo, line 322: I would put the flashlight outside of the toolbelt otherwise you can miss that it's an EMP. With it in the box it stands out more.

code/game/objects/items/weapons/storage/uplink_kits.dm
```diff
@@ -83,17 +83,18 @@
...
 			new /obj/item/aiModule/syndicate(src) // 12TC
 			new /obj/item/card/emag(src) // 6TC
 			new /obj/item/encryptionkey/syndicate(src) // 2TC
 			new /obj/item/encryptionkey/binary(src) // 5TC
 			new /obj/item/aiModule/toyAI(src) // 0TC
-			new /obj/item/multitool/ai_detect(src) // 1TC
-			new /obj/item/storage/box/syndie_kit/c4 // 4TC
+			new /obj/item/clothing/glasses/chameleon/thermal(src) // 6TC
+			new /obj/item/storage/belt/military/traitor/hacker(src) // 5TC
+			new /obj/item/clothing/gloves/color/yellow(src) // 0TC
```
datlo, line 94: Maybe combat gloves instead of insulateds? I dont think traitors get insulateds anywhere else.

code/game/objects/items/weapons/storage/uplink_kits.dm
```diff
@@ -53,23 +53,23 @@
...
 			new /obj/item/encryptionkey/syndicate(src) // 2TC
 			return
 
-		if("payday") // 33TC + four 0TC
+		if("payday") // 35TC + four 0TC
 			new /obj/item/gun/projectile/revolver(src) // 13TC
 			new /obj/item/ammo_box/a357(src) // 3TC
 			new /obj/item/ammo_box/a357(src) // 3TC
 			new /obj/item/card/emag(src) // 6TC
-			new /obj/item/grenade/plastic/c4(src) // 1TC
```
datlo, line 66: Since you're removing the C4, which was meant to blow open the vault door, perhaps replace the emag with an access tuner? The access tuner is perfect to unbolt and open the reinforced vault door and it's the same TC value.

code/game/objects/items/weapons/storage/belt.dm
```diff
@@ -308,6 +308,20 @@
...
+	name = "tool-belt"
+
+/obj/item/storage/belt/military/traitor/hacker/New()
+	..()
+	new /obj/item/screwdriver(src, "red")
+	new /obj/item/wrench(src)
+	new /obj/item/weldingtool/largetank(src)
+	new /obj/item/crowbar/red(src)
+	new /obj/item/wirecutters(src, "red")
+	new /obj/item/multitool/ai_detect(src)
```
datlo, line 321: I would put this outside of the belt aswell. Both of those are items are class C (I believe) contraband, you want the traitor to be aware they are not the regular tools.

code/game/objects/items/weapons/storage/uplink_kits.dm
ZomgPonies: Insulated are pretty necessary if you're going to be hackerman. Are there any gloves that are insulated and don't stand out as much? Maybe create a special black insulated glove option just for the kit.

datlo: Combat gloves (Used by nuke ops, the blueshield, and found in syndicate toolboxes) are insulated and fireproof.

code/game/objects/items/weapons/storage/uplink_kits.dm
AzuleUtama: Payday is more geared for a louder approach to objectives, and the emag is generally preferable for that as well as being able to open lockers. Access Tuner kinda feels like it belongs in a stealthier kit, possibly in it's own bundle with a camera bug or something.
I get that the access tuner is better for the vault door, but I'd rather the kit was more useful in general than being geared solely for one single objective, the drill is there more for the theme than anything.

farie82: They now can't break into the vault using solely the kit. Couldn't before already though since they should have only gotten one C4 charge

variableundefined (approved):

PR #13490, opened 2020-05-25
Gives flashes and cameras a light flash

farie82:
## What Does This PR Do
Gives both flashers (the portable one and the brig cell ones), flashes and cameras a flash of light when used. Both taking 2 ticks (0.2 seconds)

## Why It's Good For The Game
Makes way for some interesting uses of cameras and flashes. Spooky station coming to you soon

## Images of changes
1A72sO0He1

## Changelog
:cl:
add: Gives a flasher (portable and brig flashers) a light flash after it is used
add: Gives a flash a light flash after it is used
add: Gives a camera a light flash after it is used
/:cl:

### DIFF
--- a/code/game/machinery/flasher.dm
+++ b/code/game/machinery/flasher.dm
@@ -58,6 +58,8 @@
 
 	playsound(loc, 'sound/weapons/flash.ogg', 100, 1)
 	flick("[base_state]_flash", src)
+	set_light(2, 1, COLOR_WHITE)
+	addtimer(CALLBACK(src, /atom./proc/set_light, 0), 2)
 	last_flash = world.time
 	use_power(1000)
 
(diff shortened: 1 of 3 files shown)


AffectedArc07:
Does this also work with the mounted varients?

farie82:
> Does this also work with the mounted varients?

It does not. Those do not use flashes internally. I will include those as well for consistency

AffectedArc07:
Just to be sure, this works with mounted flashes like at the hop office, the portable flashers in the warden storage, and handheld flashes?

Oh and cameras.

farie82:
> Just to be sure, this works with mounted flashes like at the hop office, the portable flashers in the warden storage, and handheld flashes?
> 
> Oh and cameras.

Aye I've included the `machinery/flasher` type. Which to my knowledge should be all those listed above.
For items I've included the flash and the camera only

PR #13500, opened 2020-05-26
NT undercover changes

TDSSS:
## What Does This PR Do
This PR effectively renames the existing 'NT Undercover Operative' into the newly created 'NT Undercover Officer', then creates a replacement NT Undercover Agent with reduced gear and tools.

The idea is that the Agent is an outfit that can be given to ghosts during events or in response to threats as an alternative to sending an ERT, while the Undercover Officer maintains access to all the OP admin-only tools that would be overkill to give to a player.

Detailed breakdown of changes:
1. Combat gloves of both are replaced with insulated gloves. Combat gloves are not exactly common and draw unwanted attention, an undercover operative should IMO not be using them. The heat resistance is only rarely relevant.
2. An NT Saber SMG was added to both, together with 3 magazines and a suppressor. Since the Agent loses access to the CC combat implant, he needed a replacement weapon. The Saber SMG is deadly, small, suppressed, made by NT and was used by NT Undercover operatives in the past. The Undercover Officer gets this too but can freely ditch it at CC if he doesn't want it, to help his cover.
3. As mentioned, Agents lose the CC combat implant which had a pulse pistol, mindshield implanter, baton, omni-door remote and healing injector. They keep welding shield, nutritien implant, sec hud implant and gain a baton implant for unarmed combat.  They also still have a dust implant
4. Agents no longer have full Centcom access. They now have the same access as an ert team leader, ie basic centcom access. Mostly a flavour thing but it is a role intended to be given to players, so better be safe here.
5. Agents also get an emp flashlight, for disabling coms or borgs or other misc utility.

## Why It's Good For The Game
NT Undercover Operatives have a lot of potential, IMO. Sometimes, CC wants someone eliminated quietly. While an admin could spawn in as the old version of undercover operative, that is in many cases suboptimal since it 1) is pretty damn unfair, raising the barrier for using it quite high. Good luck escaping someone with a centcom combat implant 2) Giving ghosts something to do can't hurt.

So I intend to give us another tool in our event/response toolbox. Maybe rather than an ert a few undercover agents can be sent to help deal with a situation like a mutiny discreetly, giving ghosts something to do and the antagonists still a fair fighting chance. And if full admeme force is required, the Undercover Officer still has all the good stuff.

## Changelog
:cl:
add: NT Undercover Officers
tweak: Changes to NT Undercover Operatives, renamed to NT Undercover Agent
/:cl:

### DIFF
--- a/code/datums/outfits/outfit_admin.dm
+++ b/code/datums/outfits/outfit_admin.dm
@@ -220,12 +220,14 @@
 
 /datum/outfit/admin/nt_undercover
 	name = "NT Undercover Operative"
+	var/access_level = "Emergency Response Team Leader"
 	// Disguised NT special forces, sent to quietly eliminate or keep tabs on people in high positions (e.g: captain)
+	//nerfed ghost version, doesn't get the op centcom combat implant, less access
 
 	uniform = /obj/item/clothing/under/color/black
 	back = /obj/item/storage/backpack
 	belt = /obj/item/storage/belt/utility/full/multitool
-	gloves = /obj/item/clothing/gloves/combat
+	gloves = /obj/item/clothing/gloves/color/yellow
 	shoes = /obj/item/clothing/shoes/chameleon/noslip
 	l_ear = /obj/item/radio/headset/centcom
 	id = /obj/item/card/id
@@ -233,11 +235,25 @@
 	backpack_contents = list(
 		/obj/item/storage/box/engineer = 1,
 		/obj/item/flashlight = 1,
-		/obj/item/pinpointer/crew = 1
+		/obj/item/pinpointer/crew = 1,
+		/obj/item/gun/projectile/automatic/proto = 1,
+		/obj/item/ammo_box/magazine/smgm9mm = 3,
+		/obj/item/suppressor = 1
 	)
 	implants = list(
 		/obj/item/implant/dust
 	)
+	cybernetic_implants = list(
+		/obj/item/organ/internal/cyberimp/eyes/shield,
+		/obj/item/organ/internal/cyberimp/chest/nutriment/plus,
+		/obj/item/organ/internal/cyberimp/arm/baton
+	)
+
+/datum/outfit/admin/nt_undercover/officer
+	name = "NT Undercover Officer"
+	access_level = "NT Undercover Officer"
+	//Disguised NT special forces, officer (ie full admeme) version
+
 	cybernetic_implants = list(
 		/obj/item/organ/internal/cyberimp/eyes/shield,
 		/obj/item/organ/internal/cyberimp/eyes/hud/security,
@@ -254,7 +270,7 @@
 
 	var/obj/item/card/id/I = H.wear_id
 	if(istype(I))
-		apply_to_card(I, H, get_centcom_access("NT Undercover Operative"), "Civilian")
+		apply_to_card(I, H, get_centcom_access(access_level), "Civilian")
 	H.sec_hud_set_ID() // Force it to show as Civ on sec huds
 
 	var/obj/item/radio/R = H.l_ear
--- a/code/datums/outfits/outfit_admin.dm
+++ b/code/datums/outfits/outfit_admin.dm
@@ -219,10 +219,10 @@
 	// Will show as ? on sec huds, as this is not a recognized rank.
 
 /datum/outfit/admin/nt_undercover
-	name = "NT Undercover Operative"
+	name = "NT Undercover Agent"
 	var/access_level = "Emergency Response Team Leader"
-	// Disguised NT special forces, sent to quietly eliminate or keep tabs on people in high positions (e.g: captain)
-	//nerfed ghost version, doesn't get the op centcom combat implant, less access
+	// Ghost role for undercover antics on behalf of NT. Much weaker than the "NT Undercover Officer" below.
+	//Lacks CC implant and CC access. Default weapon is silenced saber SMG.
 
 	uniform = /obj/item/clothing/under/color/black
 	back = /obj/item/storage/backpack
@@ -245,6 +245,7 @@
 	)
 	cybernetic_implants = list(
 		/obj/item/organ/internal/cyberimp/eyes/shield,
+		/obj/item/organ/internal/cyberimp/eyes/hud/security,
 		/obj/item/organ/internal/cyberimp/chest/nutriment/plus,
 		/obj/item/organ/internal/cyberimp/arm/baton
 	)
@@ -252,8 +253,7 @@
 /datum/outfit/admin/nt_undercover/officer
 	name = "NT Undercover Officer"
 	access_level = "NT Undercover Officer"
-	//Disguised NT special forces, officer (ie full admeme) version
-
+	// Disguised NT special forces. Admin ("officer") version. Has CC specops implant, and CC access.
 	cybernetic_implants = list(
 		/obj/item/organ/internal/cyberimp/eyes/shield,
 		/obj/item/organ/internal/cyberimp/eyes/hud/security,
--- a/code/datums/outfits/outfit_admin.dm
+++ b/code/datums/outfits/outfit_admin.dm
@@ -224,7 +224,7 @@
 	// Ghost role for undercover antics on behalf of NT. Much weaker than the "NT Undercover Officer" below.
 	//Lacks CC implant and CC access. Default weapon is silenced saber SMG.
 
-	uniform = /obj/item/clothing/under/color/black
+	uniform = /obj/item/clothing/under/color/random
 	back = /obj/item/storage/backpack
 	belt = /obj/item/storage/belt/utility/full/multitool
 	gloves = /obj/item/clothing/gloves/color/yellow
--- a/code/datums/outfits/outfit_admin.dm
+++ b/code/datums/outfits/outfit_admin.dm
@@ -234,7 +234,7 @@
 	pda = /obj/item/pda
 	backpack_contents = list(
 		/obj/item/storage/box/engineer = 1,
-		/obj/item/flashlight = 1,
+		/obj/item/flashlight/emp = 1,
 		/obj/item/pinpointer/crew = 1,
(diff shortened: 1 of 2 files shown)


Mitchs98:
https://www.youtube.com/watch?v=6iaR3WO71j4

Relevant and obligatory.

code/datums/outfits/outfit_admin.dm
```diff
@@ -220,24 +220,40 @@
 
 /datum/outfit/admin/nt_undercover
 	name = "NT Undercover Operative"
```
Kyep, line 222: Maybe if you're creating a lesser version, use 'Agent' instead of 'Operative'.
Rationale is that the syndicate already have a similar structure ("syndicate agents" are undercover players on station, "syndicate officers" are admin-only event roles). So it makes it easier to understand what each does, and the difference between them. Using 'Agent' instead of officer also prevents both of them having the same abbreviation ("NTUO").

code/datums/outfits/outfit_admin.dm
```diff
@@ -220,24 +220,40 @@
...
+	var/access_level = "Emergency Response Team Leader"
 	// Disguised NT special forces, sent to quietly eliminate or keep tabs on people in high positions (e.g: captain)
+	//nerfed ghost version, doesn't get the op centcom combat implant, less access
 
 	uniform = /obj/item/clothing/under/color/black
 	back = /obj/item/storage/backpack
 	belt = /obj/item/storage/belt/utility/full/multitool
-	gloves = /obj/item/clothing/gloves/combat
+	gloves = /obj/item/clothing/gloves/color/yellow
```
Kyep, line 230: Is this because combat gloves do things that insulated gloves don't?
Or is this just to help them blend in more?

code/datums/outfits/outfit_admin.dm
```diff
@@ -220,24 +220,40 @@
...
 	l_ear = /obj/item/radio/headset/centcom
 	id = /obj/item/card/id
 	pda = /obj/item/pda
 	backpack_contents = list(
 		/obj/item/storage/box/engineer = 1,
 		/obj/item/flashlight = 1,
-		/obj/item/pinpointer/crew = 1
+		/obj/item/pinpointer/crew = 1,
+		/obj/item/gun/projectile/automatic/proto = 1,
+		/obj/item/ammo_box/magazine/smgm9mm = 3,
```
Kyep, line 240: You might also want to give them a comms jammer, or something so that they don't instantly get dogpiled by sec when their subject screams on radio.

code/datums/outfits/outfit_admin.dm
ZomgPonies: Blend in more. Combat gloves also offer heat protection, but that is so rarely used it's not a factor.

code/datums/outfits/outfit_admin.dm
```diff
@@ -220,24 +220,40 @@
...
+		/obj/item/suppressor = 1
 	)
 	implants = list(
 		/obj/item/implant/dust
 	)
+	cybernetic_implants = list(
```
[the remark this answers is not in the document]
TDSSS: Good idea!

code/datums/outfits/outfit_admin.dm
TDSSS: I didn't think about the abbreviation thing, will go ahead and change it.

code/datums/outfits/outfit_admin.dm
TDSSS: It's for blending in more. Combat gloves are fairly rare and the fire resistance rarely comes up, imo. Nobody bats an eye at am assistant with insulated gloves.

code/datums/outfits/outfit_admin.dm
TDSSS: Good idea. I thought about giving them an emp flashlight or something but that works, too.

Qwertytoforty: Radio jammer may not work, due to the fact they can still call out on comns and make a lot of radio noise jibberish, which will probably cause the AI to track them, and people nearby may notice the radio jammer and call sec. EMP flashlight or a muting chemical would probably work better.

TDSSS:
It is definitely a role where you need to be careful so it doesn't lead to things going out of control, ie giving ghosts clear guidance on what they are allowed to do. I think dusting would generally be correct when caught by sec. As for keeping people dead, I assumed the operative would have to get creative or bring the body back to CC, but we could explore alternatives.

>This new version could easily be spawned from Smite->Hunter, so they'd get a pinpointer leading to their target (you've verified that works, right?).

I'm not sure what I should verify. I have not tried making hunter use/spawn this outfit, I just tested it with reincarnate and with the equip verb.

code/datums/outfits/outfit_admin.dm
Kyep: Could make a modified version of radio jammer that totally silences radio, instead of just garbling it.

Kyep: Also, they only have one spare mag. Once that's gone, they're stuffed.

code/datums/outfits/outfit_admin.dm
```diff
@@ -219,25 +219,41 @@
...
 
 /datum/outfit/admin/nt_undercover
-	name = "NT Undercover Operative"
-	// Disguised NT special forces, sent to quietly eliminate or keep tabs on people in high positions (e.g: captain)
+	name = "NT Undercover Agent"
+	var/access_level = "Emergency Response Team Leader"
+	// Ghost role for undercover antics on behalf of NT. Much weaker than the "NT Undercover Officer" below.
+	//Lacks CC implant and CC access. Default weapon is silenced saber SMG.
 
 	uniform = /obj/item/clothing/under/color/black
```
Kyep, line 227: Maybe use under/color/random instead of under/color/black
Otherwise, if an admin spawns 5 of these, they'll all appear in the exact same black outfit, which makes them super-easy to identify once you know what to look for.

code/datums/outfits/outfit_admin.dm
TDSSS: Uh, no they have 3 spare mags. If 3 spare mags and the one in the gun isn't enough to do the job it's less an undercover op and more a war.

TDSSS:
Addressed a few comments. To sum up:
Changed some comment
Changed outfit to random colour
Renamed from operative to agent
Given sec HUD
Flashlight turned into emp flashlight

An emp flashlight seems more universally useful and can help out against rogue borgs or AIs, traitors with 5 energy weapons in their backpack and implants, etc.

code/datums/outfits/outfit_admin.dm
Kyep: Oh, =3. Nevermind on that one.

Denthamos:
Personally I never see myself giving a normal player this role or using it.  If I need something handled from the CC side of things I will be doing it myself and not handing that off to a player which could make things worse.

The only thing CC wise that normal players should ever be in control with is ERT's and NT Rep/blueshield.  CC and syndicate officers/high command both should always be admin only.

TDSSS:
@Denthamos 

Nobody says it's an outfit you'll see used _often_. We have lots of outfits that are only very rarely used, like Singuloth Knight, Chrono Legionnaire, etc.

> If I need something handled from the CC side of things I will be doing it myself and not handing that off to a player which could make things worse.

I think you're misunderstanding the role this outfit should serve, there. It's not intended to flawlessly _solve_ problems. It's intended to give the problem it's supposed to solve (ie rampaging vampire, traitor arranging a mutiny, etc.) a fair fighting chance while attacking from a different angle than an ERT. 

Sure, you can spawn yourself in as an NT Undercover Operative and play fair and not make use of your full range of gear, but the players you're fighting will know you were holding back, that any win they got was only because you allowed it. A victory over another player that's given limited tools and giving his best is much more rewarding.

> The only thing CC wise that normal players should ever be in control with is ERT's and NT Rep/blueshield. CC and syndicate officers/high command both should always be admin only.

How is this so much different from trusting a ghost with being an ERT? The role comes with no official authority at all, wheras an ert can easily overrule the captain or execute people if they deem it necessary. It is a 'thing CC wise' only in that it will mostly be sent from CC, and not even that is a given.

Denthamos:
> @Denthamos
> 
> Nobody says it's an outfit you'll see used _often_. We have lots of outfits that are only very rarely used, like Singuloth Knight, Chrono Legionnaire, etc.
> 

Because we do not need more things bloating up the code and clogging menus.  
> > If I need something handled from the CC side of things I will be doing it myself and not handing that off to a player which could make things worse.
> 
> I think you're misunderstanding the role this outfit should serve, there. It's not intended to flawlessly _solve_ problems. It's intended to give the problem it's supposed to solve (ie rampaging vampire, traitor arranging a mutiny, etc.) a fair fighting chance while attacking from a different angle than an ERT.
> 

I think you are misunderstanding the staffs stance on how we prefer to solve such issues in game.  Its a long standing knowledge that anything that comes from central command is admin related and we do not need non staff in such roles that will cause further confusion for players.  We already have issues with the blueshield and rep which still come up now and then which makes those roles (especially blueshield) very controversial.

> Sure, you can spawn yourself in as an NT Undercover Operative and play fair and not make use of your full range of gear, but the players you're fighting will know you were holding back, that any win they got was only because you allowed it. A victory over another player that's given limited tools and giving his best is much more rewarding.
> 
> > The only thing CC wise that normal players should ever be in control with is ERT's and NT Rep/blueshield. CC and syndicate officers/high command both should always be admin only.
> 
> How is this so much different from trusting a ghost with being an ERT? The role comes with no official authority at all, wheras an ert can easily overrule the captain or execute people if they deem it necessary. It is a 'thing CC wise' only in that it will mostly be sent from CC, and not even that is a given.

Because you are essentially giving a player a role from central command that does antagonist actions against the crew.  NTUO is used very rarely for a reason because it muddies up the water on which factions are antagonistic and which are not and is like using a scalpel to cut a wound open.  What you propose here would be like using a machete for the same use and not what we need.

PR #13506, opened 2020-05-27
Helps Organs GC Better

Fox-McCloud:
Refactors a number of organ related behaviors to use `addtimer` instead of `spawn`.

Also removes a worthless `New` for brainmobs and ensure that brain items should GC a bit better.

### DIFF
--- a/code/modules/surgery/organs/augments_internal.dm
+++ b/code/modules/surgery/organs/augments_internal.dm
@@ -141,9 +141,11 @@
 /obj/item/organ/internal/cyberimp/brain/anti_stun/emp_act(severity)
 	if(crit_fail || emp_proof)
 		return
-	crit_fail = 1
-	spawn(90 / severity)
-		crit_fail = 0
+	crit_fail = TRUE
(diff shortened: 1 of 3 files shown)


code/modules/surgery/organs/heart.dm
```diff
@@ -23,9 +23,8 @@
 			Stop()
 			return
 
-	spawn(120)
-		if(!owner)
-			Stop()
+	if(!special)
```
Fox-McCloud, line 26: If the organ is just straight up deleted from a mob or  something similar, it's pointless to have the timer checked for.

code/modules/surgery/organs/augments_internal.dm
```diff
@@ -230,11 +232,13 @@
 	if(owner.stat == DEAD)
 		return
 	if(owner.nutrition <= hunger_threshold)
-		synthesizing = 1
+		synthesizing = TRUE
```
AffectedArc07, line 235: The `TRUE`/`FALSE` gods will be happy

PR #13515, opened 2020-05-29
Fixes the Changeling tentacle grab logging

trololiver112:
## What Does This PR Do
Replaces the [src] with [H] and fixes the logging for using a changeling tentacle on grab intent. Also simplify the message.

## Why It's Good For The Game
Simple fix on some logging issues.

## Changelog
:cl:
fix: fixes the logging for using the changeling tentacle on grab intent
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/mutations.dm
+++ b/code/game/gamemodes/changeling/powers/mutations.dm
@@ -321,7 +321,7 @@
 
 					if(INTENT_GRAB)
 						C.visible_message("<span class='danger'>[L] is grabbed by [H]'s tentacle!</span>","<span class='userdanger'>A tentacle grabs you and pulls you towards [H]!</span>")
-						add_attack_logs(src, C, "[src] has grabbed [C] and pulled them towards [H] with a tentacle")
+						add_attack_logs(H, C, "[H] grabbed [C] with a changeling tentacle")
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/changeling/powers/mutations.dm
```diff
@@ -321,7 +321,7 @@
 
 					if(INTENT_GRAB)
 						C.visible_message("<span class='danger'>[L] is grabbed by [H]'s tentacle!</span>","<span class='userdanger'>A tentacle grabs you and pulls you towards [H]!</span>")
-						add_attack_logs(src, C, "[src] has grabbed [C] and pulled them towards [H] with a tentacle")
+						add_attack_logs(H, C, "[H] grabbed [C] with a changeling tentacle")
```
farie82, line 324: The new system will put the attacker and defender in their own cell in the list. It'll show it like this:
`henk | grabbed using changeling tentacle | victim`
But then made visually better.
Due to this you don't need to add the text in the actual attack message. Can still do this though. Not sure what the admins prefer themselves. 
Gonna leave it up to you as this was more meant as information than a review.

trololiver112: If any admins prefer it differently then please do comment. If not i think i will keep this

PR #13525, opened 2020-06-01
"Reject All" button for cargo console and PDA notifications

rebel228:
## What Does This PR Do

Adds a "Reject All" for supply shuttle console that allows to reject all of the pending requests.

Whenever a supply request gets accepted or rejected the requester gets a PDA notification. The message has order number, the name of the requested crate and whether it was accepted or rejected.
However when "Reject All" button is used the requests get grouped. When the recipient has several requests of the same item he will get them in one message.
## Why It's Good For The Game

PDA notification is a convenient way for crewmembers to get updated on their cargo request status. 
Reject all button is going to allow cargotech/QM to easily clear pending requests instead of manually rejecting all of them whenever someone spams the console with 20 of the same crate.
## Changelog
:cl:
add: Added a "Reject All" button for Supply Shuttle Console
add: A PDA notification is being sent to the requester by cargo console whenever an order gets accepted/rejected.
/:cl:

### DIFF
--- a/code/modules/shuttle/supply.dm
+++ b/code/modules/shuttle/supply.dm
@@ -691,6 +691,31 @@
 				O.generateRequisition(loc)
 
 	else if (href_list["rejectall"])
+		var/list/orders_by_users = list()
+		var/ordernumbers
+		var/recipient
+		var/object_name
+		var/multiple = FALSE
+		for(var/A in SSshuttle.requestlist)
+			var/datum/supply_order/SO = A
+			if(!orders_by_users["[SO.orderedby][SO.object.name]"])
+				orders_by_users["[SO.orderedby][SO.object.name]"] = list()
+			orders_by_users["[SO.orderedby][SO.object.name]"] += SO
+
+		for(var/A in orders_by_users)
+			var/list = A
+			for(var/B in orders_by_users[list])
+				var/datum/supply_order/SO = B
+				if(!ordernumbers)
+					ordernumbers = "[SO.ordernum]"
+					recipient = SO.orderedby
+					object_name = SO.object.name
+				else
+					ordernumbers += ", #[SO.ordernum]"
+					multiple = TRUE
+			notify_pda(recipient, ordernumbers, object_name, multiple)
+			ordernumbers = null
+			multiple = FALSE
 		SSshuttle.requestlist.Cut()
 
 	else if(href_list["confirmorder"])
@@ -705,6 +730,7 @@
 				O = SO
 				P = O.object
 				if(SSshuttle.points >= P.cost)
+					notify_pda(SO.orderedby, SO.ordernum, SO.object.name, confirmed = TRUE)
 					SSshuttle.requestlist.Cut(i,i+1)
 					SSshuttle.points -= P.cost
 					SSshuttle.shoppinglist += O
@@ -718,6 +744,7 @@
 		for(var/i=1, i<=SSshuttle.requestlist.len, i++)
 			var/datum/supply_order/SO = SSshuttle.requestlist[i]
 			if(SO.ordernum == ordernum)
+				notify_pda(SO.orderedby, SO.ordernum, SO.object.name)
 				SSshuttle.requestlist.Cut(i,i+1)
 				break
 
@@ -749,6 +776,58 @@
 
 	frequency.post_signal(src, status_signal)
 
+/obj/machinery/computer/supplycomp/proc/notify_pda(recipient, order_number, ordered_object, multiple = FALSE, confirmed = FALSE)
+
+	var/obj/machinery/message_server/useMS = null
+	if(GLOB.message_servers)
+		for(var/A in GLOB.message_servers)
+			var/obj/machinery/message_server/MS = A
+			if(MS.active)
+				useMS = MS
+				break
+
+	if(!useMS)
+		return
+
+	var/datum/data/pda/app/messenger/PM
+	var/obj/item/pda/P
+	for(var/A in GLOB.PDAs)
+		P = A
+		if(P.owner == recipient)
+			PM = P.find_program(/datum/data/pda/app/messenger)
+			break
+
+	if(!PM || PM.toff)
+		return
+
+	var/turf/sender_pos = get_turf(src)
+	var/turf/recipient_pos = get_turf(P)
+
+	var/sendable = FALSE
+	var/receivable = FALSE
+	for(var/obj/machinery/tcomms/core/C in GLOB.tcomms_machines)
+		if(C.zlevel_reachable(sender_pos.z))
+			sendable = TRUE
+		if(C.zlevel_reachable(recipient_pos.z))
+			receivable = TRUE
+		// Once both are done, exit the loop
+		if(sendable && receivable)
+			break
+
+	if(!sendable || !receivable)
+		return	
+
+	var/havehas = "has"
+	var/s
+	var/confirmverb = "rejected"
+	if(confirmed)
+		confirmverb = "approved"
+	if(multiple)
+		havehas = "have"
+		s = "s"
+
+	useMS.send_pda_message("[recipient]","Supply Ordering Console","Your order[s] #[order_number] ([ordered_object]) [havehas] been [confirmverb]")
+	PM.notify("<b>Automatic message from Supply Ordering Console, </b>\"Your order[s] #[order_number] ([ordered_object]) [havehas] been [confirmverb]\" (No reply)", 0)
 
 #undef ORDER_SCREEN_WIDTH
 #undef ORDER_SCREEN_HEIGHT
--- a/code/modules/shuttle/supply.dm
+++ b/code/modules/shuttle/supply.dm
@@ -690,6 +690,9 @@
 			if(i == 1)
 				O.generateRequisition(loc)
 
+	else if (href_list["rejectall"])
+		SSshuttle.requestlist.Cut()
+
 	else if(href_list["confirmorder"])
 		if(SSshuttle.supply.getDockedId() != "supply_away" || SSshuttle.supply.mode != SHUTTLE_IDLE)
 			return 1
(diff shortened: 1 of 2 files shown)


Shendow:
Maybe you could add a "Reject all from requester" button next to the regular "Reject" button to reject all orders from that requester? It would prevent the trouble of having to redo the order for other requesters.

code/modules/shuttle/supply.dm
```diff
@@ -690,6 +690,34 @@
 			if(i == 1)
 				O.generateRequisition(loc)
 
+	else if (href_list["rejectall"])
```
SteelSlayer, line 693: This section, from this line 693 down to 719 is overly complicated, when it can actually be a lot simpler. I pulled your code and made some tweaks:
```js
	else if(href_list["rejectall"])
		for(var/A in SSshuttle.requestlist)
			var/datum/supply_order/SO = A
			notify_pda(SO.orderedby, SO.ordernum, SO.object.name)
		SSshuttle.requestlist.Cut()
```
Everything still worked fine, all the orders got rejected, they disappeared from the cargo console list, and I got a bunch of rejected PDA notifications. You might need to tweak your PDA proc a bit to make this work but its much easier to read.

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +776,58 @@
...
+	var/s
+	var/confirmverb = "rejected"
+	if(confirmed)
+		confirmverb = "approved"
+	if(multiple)
+		havehas = "have"
+		s = "s"
+
+	useMS.send_pda_message("[recipient]","Supply Ordering Console","Your order[s] #[order_number] ([ordered_object]) [havehas] been [confirmverb]")
+	PM.notify("<b>Automatic message from Supply Ordering Console, </b>\"Your order[s] #[order_number] ([ordered_object]) [havehas] been [confirmverb]\" (No reply)", 0)
```
SteelSlayer, line 830: Correct me if I'm wrong but, since you are doing a for loop in the `"rejectall"` section under line 693, each of these orders are printed one by one. Which means you have no use for the `multiple` variable here, as it will always be a singular order.

code/modules/shuttle/supply.dm
rebel228: This is how i did it initially, however the downside of the simple method is that when you request a lot of crates you get bombarded by a lot of notifications and beeps in a split second. It floods the chat screen with messages and the beep sounds overlap and sound not very cool.
So what i tried to accomplish is to group requests from the same person of the same item into one message. But unfortunally the code turns out to be complicated.

SteelSlayer: Oh jeez, ok then. But I have to wonder if putting up with a bit of spam is fine just for the sake of code simplicity though, because this probably won't happen more than 1 or 2 times a round, if even that. 

I suppose you could also find a way to just send a no-reply PDA message they can view, so its just one message and one ping, but that has a list of all the rejected orders. Just some ideas though, you don't have to listen to me.

code/modules/shuttle/supply.dm
rebel228: In the second screenshot there is a message which contains multiple orders. What happens is `orders_by_users` is an associative list, where the index is "requester name+requested object" and the value is another list which contains `var/datum/supply_order/` datums with the same object and requestd by the same person, which are then put into the same message.
The code from line 707 to 715 fills the message.

SteelSlayer: Yeah sorry I see what you did there now.

code/modules/shuttle/supply.dm
rebel228: I will definitely add some comments in the code for now to make it easier to read. Will see what other reviewers say.

code/modules/shuttle/supply.dm
farie82, line 830: `FALSE` at the end there instead of 0.

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +776,58 @@
...
+	var/havehas = "has"
+	var/s
+	var/confirmverb = "rejected"
+	if(confirmed)
+		confirmverb = "approved"
+	if(multiple)
+		havehas = "have"
+		s = "s"
+
+	useMS.send_pda_message("[recipient]","Supply Ordering Console","Your order[s] #[order_number] ([ordered_object]) [havehas] been [confirmverb]")
```
farie82, line 829: Spacing

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +776,58 @@
...
+		if(C.zlevel_reachable(sender_pos.z))
+			sendable = TRUE
+		if(C.zlevel_reachable(recipient_pos.z))
+			receivable = TRUE
+		// Once both are done, exit the loop
+		if(sendable && receivable)
+			break
+
+	if(!sendable || !receivable)
+		return
```
farie82, line 818: Rogue tab at the end. Best to clear those to avoid random whitespace changes in future PRs. VScode can clear these for you on save

code/modules/shuttle/supply.dm
farie82: Might be best to make this into a proc of it's own seeing the size of the logic done in here.

code/modules/shuttle/supply.dm
rebel228: All done.

rebel228:
> 
> 
> Maybe you could add a "Reject all from requester" button next to the regular "Reject" button to reject all orders from that requester? It would prevent the trouble of having to redo the order for other requesters.

I tried making something like this, but i didnt find a way to make it look nice and not clunky/weird in the UI.

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +752,85 @@
 
 	frequency.post_signal(src, status_signal)
 
+/obj/machinery/computer/supplycomp/proc/sort_and_send_messages()
```
farie82, line 755: A better name would help here. This name suggests it sorts and sends messages. While it does something else. It sends all players that have an order a message saying their order is denied.
Gonna leave the naming up to you

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +752,85 @@
...
+				recipient = SO.orderedby            //who will get the message
+				object_name = SO.object.name        //what crate was requested
+			else
+				ordernumbers += ", #[SO.ordernum]"  //when more then one crate was requested add a coma and the next order number
+				multiple = TRUE                     //different verbs are used in the message when there are multiple orders in it
+		notify_pda(recipient, ordernumbers, object_name, multiple)
+		ordernumbers = null
+		multiple = FALSE
+
+/obj/machinery/computer/supplycomp/proc/notify_pda(recipient, order_number, ordered_object, multiple = FALSE, confirmed = FALSE)
```
farie82, line 782: maybe renamed the `confirmed` param to `approved`. Confirmed can be either in this case.

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +752,85 @@
...
+		for(var/B in orders_by_users[list])         //loop through the order datums in each list
+			var/datum/supply_order/SO = B
+			if(!ordernumbers)                       //if its the first item in the list
+				ordernumbers = "[SO.ordernum]"
+				recipient = SO.orderedby            //who will get the message
+				object_name = SO.object.name        //what crate was requested
+			else
+				ordernumbers += ", #[SO.ordernum]"  //when more then one crate was requested add a coma and the next order number
+				multiple = TRUE                     //different verbs are used in the message when there are multiple orders in it
+		notify_pda(recipient, ordernumbers, object_name, multiple)
```
farie82, line 778: I'd explicitly state here that the messages send are denied. So adding a `FALSE` at the end will help readability.

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +752,85 @@
...
+	var/object_name
+	var/multiple = FALSE
+	for(var/A in SSshuttle.requestlist)
+		var/datum/supply_order/SO = A
+		if(!orders_by_users["[SO.orderedby][SO.object.name]"])         //orders_by_users is an associative list where index is "[requester name][object name]"
+			orders_by_users["[SO.orderedby][SO.object.name]"] = list() //and the value is another list, which contains order datums from same person of same object
+		orders_by_users["[SO.orderedby][SO.object.name]"] += SO        //they are used later to fill the messages with multiple orders
+
+	for(var/A in orders_by_users)                   //take a list from orders_by_users
+		var/list = A
```
farie82, line 768: naming a variable `list` is misleading if it's not a list. Best to rename it to `key` or `requesterName` or the like.

rebel228: done

code/modules/shuttle/supply.dm
```diff
@@ -746,6 +752,85 @@
...
+		for(var/B in orders_by_users[key])         //loop through the order datums in each list
+			var/datum/supply_order/SO = B
+			if(!ordernumbers)                       //if its the first item in the list
+				ordernumbers = "[SO.ordernum]"
+				recipient = SO.orderedby            //who will get the message
+				object_name = SO.object.name        //what crate was requested
+			else
+				ordernumbers += ", #[SO.ordernum]"  //when more then one crate was requested add a coma and the next order number
+				multiple = TRUE                     //different verbs are used in the message when there are multiple orders in it
+		notify_pda(recipient, ordernumbers, object_name, multiple, approved = FALSE)
```
AffectedArc07, line 778:
```suggestion
		notify_pda(recipient, ordernumbers, object_name, multiple, FALSE)
```

AffectedArc07:
After some discussion between maintainers, this PR will be closed for a few reasons.
The PDA sending code can greatly be simplified taking the message servers out of the equation, in the same way that the 
```dm
	var/obj/item/pda/target_pda
	for(var/obj/item/pda/check_pda in GLOB.PDAs)
		if(check_pda.owner == recipient)
			target_pda = check_pda
			break
	if(!target_pda)
		return
	var/datum/data/pda/app/messenger/PM = target_pda.find_program(/datum/data/pda/app/messenger)
	if(PM && PM.can_receive())
		PM.notify("<b>Automated Notification: </b>\"[text]\" (Unable to Reply)")
```

This PR also touches nanoUIs, which have all work on them banned as we progress further to TGUI. You are more than welcome to remake this in the future, after the console has been TGUI'd.

PR #13527, opened 2020-06-02
Robot_module and cyborg stack refactor

### LINKED ISSUE #12784: Borg & Drones have modules whose tools are lost when fully used
**Issue Description**:
Whenever a Borg or Drone fully use their stack of equipment, they disappear entirely from their module until recharged at a station.

That the stack didn't disappear and remained in the module, at 0 stacks, unable to be used but still able to interact and restock from equal products on the floor as per usual. With the addition that the Engineering Module should be able to restock to 50 Cable Coil, instead of only 30 (can only restock to 50 at a Recharger), but that's just a small QoL improvement. As far the Mediborg Cyborg Hypospray goes, it should actually be restocked on it's own with Battery Power, albeit slightly slower than when in a Recharger.

**What happened instead**:
You lose your stack entirely and the ability to use and restock until recharged. This is incredibly annoying when you're an Engineering Module that could restock from leftover materials after a bombing. And in fact, you still can IF you don't completely use your stack.

It is a big nuisance that forces you to backtrack to a Recharger even if you are able to either manufacture more on the spot (Tiles) or bring them with you (Cable Coil). If a Feature, it feels useless since you can just not use your stack entirely, but a player shouldn't have to worry about using it's stack until that 1 mark. In a special note, this has resulted before in a permanent loss of a stack, namely some Cyborg Hypospray products from the Mediborg, that refused to return even after a rather lengthy charging. This obviously resulted in a Mediborg with highly reduced efficiency whose Hypospray is already incredibly slow to fill up.

Using materials to it's fullest until they are spent\reach 0. They will disappear and force you to recharge to be able to use or manually restock again. I have not tried to replicate the Cyborg Hypospray issue (I consciously prevent it now), although it is incredibly slow at recharging it's chemicals. I am unsure if this slowness is related but the fully use of materials seems to be.

From my own experience, ever since I played Borg.

**Extra information**:
It could be stated as a Feature, but I consider it a Bug due to being a nuisance, inefficient and irrational. And it's not the first time where I lose certain tools, namely some Cyborg Hypospray products with a Medical Borg (as previously stated), due to using them fully (which is most certainly a Bug). This made the Borg rather underwhelming where it was most needed, since two of the most important and most used products were forever lost. These issues make playing Borg more of an hassle than fun, as nobody wants to keep going to a Recharger solely because of a misstep of using their stack in it's entirety. And as Mediborg it's all about measuring chemicals with unnecessary precision and using them very sparingly (in it's obligatory 5u from a total max of 30u), with products that should really be used freely (with the obvious exception of Epinephrine, unless you are rogue).

SteelSlayer:
## What Does This PR Do
This ports TG's `robot_energy_storage` datums for use with robot stacks (metal, wood, etc.). In the process, a lot of `robot_module.dm`, has been overhauled and brought more in line with what TG has. Other files have also had some changes to accommodate the additions in `robot_module.dm`.

Note: this is not a 1 to 1 port, and yes I did make some of my own additions and improvements where I saw fit, but a majority of TG's system is there.

Things that I personally created/modified/removed:
- The addition of the `default_modules` list
- The addition of the `special_rechargables` list
- The new `cyborg_recharge()` proc for items
- The new handling of how the cyborg recharger recharges its occupant's items (related to the above two)
- The creation of the new `unemag()` proc for borgs. Saves on copy and paste.
- Adding the `energy_type` variable to stacks. This is not part of TG's energy storage datums.
- Adding a new `robot_energy_storage` datum for every stack item. For example I added the wood, wood tile, reinforced glass, and rod datums, as well as all the medical ones. You won't find those in TG. The reason I created one for every stack,is that I wanted to keep things as close to how they currently are as possible, where every stack has its own seperate amount.
- The complete removal of the `installed_modules()` proc and anything related to it. As far as I can tell this was old legacy code for when robots didn't have a GUI for their inventory. I had to touch this to change some emag stuff but then realized its not even used so... to the chopping block with you.

#### What does this all mean for players?
There are a few gameplay tweaks players should notice:
1. All borg/drone stacks now start at max capacity and can be charged to their maximum stack size, as said in the first item of the changelog. This was mostly done for simplicity if I'm honest but I don't really see the need to cap tile recharge to 20, or have starting drone wood at 10 stacks but also having it able to charge up to 50, for example.

2. All borg/drone stacks will never disappear. If it runs to 0, the stack item remains in their inventory. So say you accidentally run out of cables wiring solars, well now you can just go find a roll of cable, and use your "empty" cable on it, which will add the amount of that cable stack to your own. The same can be done with metal, glass, and so on.

3. Gives borgs/drones some text in their status tab which lists their current stack amounts. You can see that here:
image

4. Re-organizes the medical and syndicate medical borg inventory so that all surgery tools are on one line, and related tools are more grouped up.

## Why It's Good For The Game
Better/cleaner way of handling robot stacks. Adds more documentation to vars and procs. QoL improvements for borgs and drones.

Closes #12784

## Changelog
:cl:
tweak: All robot stacks will now start at maximum capacity and can be charged to their maximum stack size. This means 50 for things like metal or glass, 60 for floor tiles, and 6 for medical stacks, like splints, burn kits, etc.
tweak: Robot stacks will no longer disappear when their amount reaches 0.
tweak: Re-organizes the medical and syndicate medical borg inventory so that all surgery tools are on one line, and related tools are more grouped up
add: Gives borgs and drones some text in their status tab which lists their current stack amounts.
/:cl:


code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -6,77 +6,141 @@
...
+/// Builds the usable module list from the modules we have in `default_modules`, `basic_modules` and `emag_modules`
+/obj/item/robot_module/proc/rebuild_modules()
+	var/mob/living/silicon/robot/R = loc
+	R.uneq_all()
 	modules = list()
-	for(var/obj/O in temp_list)
-		if(O)
-			modules += O
 
+	for(var/obj/item/I in default_modules)
```
farie82, line 123: Would these lists ever have something else than an item?
Only when it fails to New the borg right?
Same for the ones below

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -6,77 +6,141 @@
...
+	if(R.emagged)
+		for(var/obj/item/I in emag_modules)
+			add_module(I, FALSE)
+
+	if(R.hud_used)
+		R.hud_used.update_robot_modules_display()
+
+/// Handles the recharging of all borg stack items and any items which are in the `special_rechargables` list.
+/obj/item/robot_module/proc/recharge_consumables(mob/living/silicon/robot/R, coeff = 1)
+	for(var/datum/robot_energy_storage/st in storages)
```
farie82, line 138: Same here and the one below.
Can this list have anything other than the filtered var type?

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -114,214 +180,196 @@
...
+		/obj/item/stack/medical/ointment/advanced/cyborg,
+		/obj/item/stack/medical/splint/cyborg,
+		/obj/item/stack/nanopaste/cyborg
+	)
+	emag_modules = list(/obj/item/reagent_containers/spray/cyborg_facid)
+	special_rechargables = list(/obj/item/reagent_containers/spray/cyborg_facid, /obj/item/extinguisher/mini)
+
+/obj/item/robot_module/medical/emag_act()
+	for(var/obj/item/borg_defib/F in modules)
+		F.safety = 0
```
farie82, line 237: `FALSE` here

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -339,252 +387,300 @@
...
-	modules += new /obj/item/t_scanner/adv_mining_scanner/cyborg(src)
-	modules += new /obj/item/gun/energy/kinetic_accelerator/cyborg(src)
-	modules += new /obj/item/gps/cyborg(src)
-	emag = new /obj/item/borg/stun(src)
-
-	fix_modules()
+// Replace their normal drills with diamond drills
+/obj/item/robot_module/miner/emag_act()
+	var/obj/item/pickaxe/drill/cyborg/D = locate() in modules
+	basic_modules -= D // Remove this here so it won't get re-created in the module rebuild.
```
farie82, line 413: Won't this break the borg if he gets unemagged?

code/modules/mob/living/silicon/robot/robot_modules.dm
SteelSlayer: It *shouldn't* at this point but it is a list of type paths intially. Same with `basic_modules` and `emag_modules` as well. I figure this doesn't happen to often too be considered a performance concern, but I could probably do a typeless loop here.

code/modules/mob/living/silicon/robot/robot_modules.dm
SteelSlayer: Nope it should always contain those datums.

code/modules/mob/living/silicon/robot/robot_modules.dm
SteelSlayer: Yep

code/modules/mob/living/silicon/robot/robot_modules.dm
SteelSlayer: Yeah this will break now that I think about it. The miner borg is the snowflake of the group, its the only one that gets an item removed when emagged. Would simplify things if I could just add the diamond one and have two drills.

AffectedArc07:
Can we get this updated to master so it can be testmerged?

SteelSlayer:
There we go, was easy since no conflicts.

AffectedArc07:
This was testmerged on live and a few runtimes were found. Runtimes have been handed off to PR author to be dealt with.

ExusA:
not sure if this is the right place to report this, but solar panels built using cyborg glass stacks during the testmerge, when deconstructed dropped /glass/cyborg stacks without a stack size ("there is enough energy to produce  sheets of glass")

SteelSlayer:
@ExusA That is very helpful, Thanks. I was wondering how the hell a /cyborg stack got out into the world and was being used by a human.

AffectedArc07:
This was testmerged again on live, and the following bugs were encountered, specifically when building a sleeper
```
[2020-08-24T07:58:42] Runtime in stack.dm,259: Cannot read null.energy
   proc name: get amount (/obj/item/stack/proc/get_amount)
   usr: DTX (affectedarc07) (/mob/living/silicon/robot)
   usr.loc: The floor (123,96,1) (/turf/simulated/floor/plasteel)
   src: the glass (/obj/item/stack/sheet/glass/cyborg)
   src.loc: the machine frame (/obj/machinery/constructable_frame/machine_frame)
   call stack:
   the glass (/obj/item/stack/sheet/glass/cyborg): get amount()
   the glass (/obj/item/stack/sheet/glass/cyborg): merge(the glass (/obj/item/stack/sheet/glass/cyborg))
   the glass (/obj/item/stack/sheet/glass/cyborg): New(the machine frame (/obj/machinery/constructable_frame/machine_frame), null, 1)
   the glass (/obj/item/stack/sheet/glass/cyborg): New(the machine frame (/obj/machinery/constructable_frame/machine_frame), null)
   the machine frame (/obj/machinery/constructable_frame/machine_frame): attackby(the glass (/obj/item/stack/sheet/glass/cyborg), DTX (/mob/living/silicon/robot), "icon-x=9;icon-y=19;left=1;scre...")
   the glass (/obj/item/stack/sheet/glass/cyborg): melee attack chain(DTX (/mob/living/silicon/robot), the machine frame (/obj/machinery/constructable_frame/machine_frame), "icon-x=9;icon-y=19;left=1;scre...")
   DTX (/mob/living/silicon/robot): ClickOn(the machine frame (/obj/machinery/constructable_frame/machine_frame), "icon-x=9;icon-y=19;left=1;scre...")
   the machine frame (/obj/machinery/constructable_frame/machine_frame): Click(the floor (123,97,1) (/turf/simulated/floor/plasteel), "mapwindow.map", "icon-x=9;icon-y=19;left=1;scre...")
```
^ This runtime occured when inserting the second sheet of glass, however, not the first.


image
^ The component parts of the machine referenced the cyborg stacks of glass and cable, meaning a human could get hold of these if they took the machine apart.

SteelSlayer:
Nice, more useful information. I think I see the problem, missed some stuff from tg's constructable frames. I'll fix it soon:tm:

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -913,30 +909,24 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
...
 
-	var/list/all_modules = contents | module.modules
-	for(var/obj/item/I in all_modules)
-		if(!(I in module.emag_modules)) // If `I` is not an emagged module, skip it.
-			continue
-
-		// Clear the emagged item from any of their active modules.
+	// Deletes the emag items, and recreates them.
+	// This isn't that clean but it prevents shenanigans where emag items would hang around in the inventory.
+	for(var/obj/item/I in module.emag_modules)
```
farie82, line 920: this list can only contain items right? If so it can be made typeless

farie82: Shouldn't this also be part of the `module/unemag` proc?

SteelSlayer: Yes it can be made typless, and yeah I'll put it into `module/unemag` instead. Also, this made me take a look at this proc and realize I'm missing some logic, which I will be adding in.

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -6,77 +6,222 @@
...
+	/// For icon usage.
+	var/module_type = "NoMod"
 
-	var/module_type = "NoMod" // For icon usage
 
-	var/list/stacktypes
-	var/channels = list()
-	var/list/custom_removals = list()
+/obj/item/robot_module/New()
+	. = ..()
```
farie82, line 34: `New` does not return a value

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -6,77 +6,222 @@
...
+		if(I) // If it exists, add the object reference.
+			special_rechargables += I
+		special_rechargables -= path // No matter what, remove the path from the list.
+
+	// Add all the modules into the robot's inventory. Without this, their inventory will be blank.
+	rebuild_modules()
+
+
+/obj/item/robot_module/emp_act(severity)
+	for(var/obj/O in modules)
```
farie82, line 70: Can be typeless since modules should only ever have items in it

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -378,267 +515,346 @@
...
+	/// The amount of energy the stack will regain while charging.
+	var/recharge_rate = 1
+	/// Current amount of energy.
+	var/energy
+
+/datum/robot_energy_storage/New(obj/item/robot_module/R = null)
+	if(!energy)
+		energy = max_energy
+	if(R)
+		R.storages |= src
```
farie82, line 769: |= will be the same as += since src is just created in this instance

code/modules/power/solar.dm
```diff
@@ -220,7 +220,13 @@
 		if(istype(W, /obj/item/stack/sheet/glass) || istype(W, /obj/item/stack/sheet/rglass))
 			var/obj/item/stack/sheet/S = W
 			if(S.use(2))
-				glass_type = W.type
+				// Snowflake checks to make sure when the solar panel is deconstructed, it doesn't spawn cyborg glass.
```
farie82, line 223: Can you use the `merge_type` value here instead?

farie82 (changes requested):

code/modules/power/solar.dm
SteelSlayer: That is a much cleaner idea.

farie82 (approved):
LGTM. Good shit

Alex1222003:
This is the PR that let's synth mains play the way god intended.

code/game/objects/items/weapons/stunbaton.dm
```diff
@@ -202,6 +202,10 @@
 		return TRUE
 	..()
 
+/obj/item/melee/baton/cyborg_recharge(coeff, emagged)
+	if(cell)
+		cell.charge = cell.maxcharge
```
Fox-McCloud, line 207: update the icon of this too.

code/modules/atmospherics/machinery/airalarm.dm
```diff
@@ -961,15 +961,13 @@
...
-				if(coil.amount < 5)
+				if(coil.get_amount() < 5)
 					to_chat(user, "You need more cable for this!")
 					return
 
 				to_chat(user, "You wire \the [src]!")
 				playsound(get_turf(src), coil.usesound, 50, 1)
-				coil.amount -= 5
-				if(!coil.amount)
-					qdel(coil)
```
Fox-McCloud, line 972: oh dear.

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -635,17 +635,27 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 	else
 		stat(null, text("No Cell Inserted!"))
 
+/mob/living/silicon/robot/proc/show_gps_coords()
```
Fox-McCloud, line 638: Reviewed up to here, will continue.

code/modules/projectiles/guns/energy/stun.dm
```diff
@@ -53,3 +53,12 @@
 /obj/item/gun/energy/disabler/cyborg/newshot()
 	..()
 	robocharge()
+
+/obj/item/gun/energy/disabler/cyborg/cyborg_recharge(coeff, emagged)
```
Fox-McCloud, line 57: TG has a generic process for recharging a shot for all energy guns; might be necessary to prevent some of the energy guns on a few borgs from breaking.

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -131,6 +131,9 @@
 	volume = 50
 	list_reagents = list("cleaner" = 50)
 
+/obj/item/reagent_containers/spray/cleaner/drone/cyborg_recharge(coeff, emagged)
```
Fox-McCloud, line 134: What about the janiborg lube?

code/modules/projectiles/guns/energy/stun.dm
```diff
@@ -53,3 +53,12 @@
...
+	if(cell.charge < cell.maxcharge)
+		var/obj/item/ammo_casing/energy/E = ammo_type[select]
+		cell.give(E.e_cost)
```
Fox-McCloud, line 60: `* coeff`

code/game/objects/items/weapons/stunbaton.dm
SteelSlayer: The baton, cell, or both?

code/modules/reagents/reagent_containers/spray.dm
SteelSlayer: It's declared locally over in `robot_module.dm`:

image

code/modules/projectiles/guns/energy/stun.dm
SteelSlayer: Do you know the name of the proc they use?

code/game/objects/items/weapons/stunbaton.dm
Fox-McCloud: The baton.

code/modules/projectiles/guns/energy/stun.dm
Fox-McCloud: It's just in the base level of `respawn_consumable`

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -6,77 +6,223 @@
...
 	flags = CONDUCT
 
+	/// A list of all currently usable and created modules the robot currently has access too.
 	var/list/modules = list()
-	var/obj/item/emag = null
-	var/list/subsystems = list()
+	/// A list of module-specific, non-emag modules the borg will gain when this module is chosen.
+	var/list/basic_modules = list()
+	/// A list of modules ALL robots will get by default, unless the list is overriden.
+	var/list/default_modules = list(/obj/item/flash/cyborg)
```
Fox-McCloud, line 14: I kinda question the need to keep a separate list just for this, unless there's a good reason for it. Couldn't we just have `/obj/item/flash/cyborg` in each module's `basic_modules` list?

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -6,77 +6,223 @@
...
+		if(I) // If it exists, add the object reference.
+			special_rechargables += I
+		special_rechargables -= path // No matter what, remove the path from the list.
+
+	// Add all the modules into the robot's inventory. Without this, their inventory will be blank.
+	rebuild_modules()
+
+
+/obj/item/robot_module/emp_act(severity)
+	for(var/object in modules)
```
Fox-McCloud, line 70: Your parents want to call you.

code/modules/mob/living/silicon/robot/robot_modules.dm
SteelSlayer: That's a good point. I'm going to get rid of it, and stick the flash in `basic_modules`

Fox-McCloud:
Done reviewing, seems ok.

I might come back and tweak this, later on down the road, but as a base system, it seems fine; still some issues to address though.

code/modules/projectiles/guns/energy/stun.dm
SteelSlayer: Ah, I believe we have something similar:

image

Which is ultimately called from the `on_recharge()` just below on line 61.

Fox-McCloud: Hmm, not the same, but or gun code is ancient by now...

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -307,7 +307,7 @@
 		return FALSE
 	if(istype(I, /obj/item/stack))
 		var/obj/item/stack/S = I
-		if(S.amount == 0)
+		if(S.get_amount() == 0)
```
AffectedArc07, line 310:
```suggestion
		if(!S.get_amount())
```

AffectedArc07 (approved):
Testmerged for multiple *days* with minimal issues (which have since been fixed).

Its time.

PR #13529, opened 2020-06-02
ChemHeater upgrading and QOL

dearmochi:
## What Does This PR Do
- ChemHeater's micro-laser component can now be upgraded for faster temperature change
- Adds auto-eject feature (off by default) to ChemHeater when the target has been reached
- Makes ChemHeater temperature label colored depending on whether the target has been reached
- Adds a "ding" sound to ChemHeater when the target has been reached
- ChemHeater no longer turns off when there's a <3 temperature difference between target and current - it now turns off when the target has been reached

## Why It's Good For The Game
Allows Chemists to spend less time waiting for the machine to heat/cool the beaker
Increases quality of life when using ChemHeater

## Images of changes
image

## Changelog
:cl:
add: Ability to upgrade the chemical heater's micro-laser for increased speed
add: Auto-eject feature for the chemical heater
add: Chemical heater UI colors temperature label depending on if it reached the target or not
add: Ding sound when chemical heater is done
tweak: Chemical heater now turns off when the target has been reached, not when there's a <3 temperature difference between target and current
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/machinery/chem_heater.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_heater.dm
@@ -1,7 +1,7 @@
 /obj/machinery/chem_heater
 	name = "chemical heater"
 	density = 1
-	anchored = 1
+	anchored = TRUE
 	icon = 'icons/obj/chemical.dmi'
 	icon_state = "mixer0b"
 	use_power = IDLE_POWER_USE
@@ -37,7 +37,7 @@
 				on = FALSE
 				SSnanoui.update_uis(src)
 				return
-			beaker.reagents.temperature_reagents(desired_temp, 35 - speed_increase)
+			beaker.reagents.temperature_reagents(desired_temp, max(1, 35 - speed_increase))
 			if(round(beaker.reagents.chem_temp) == round(desired_temp))
 				playsound(loc, 'sound/machines/ding.ogg', 50, 1)
 				on = FALSE
@@ -122,11 +122,11 @@
 		if(!beaker.reagents.total_volume)
 			return FALSE
 		on = !on
-		. = 1
+		. = TRUE
 
 	if(href_list["toggle_autoeject"])
 		auto_eject = !auto_eject
-		. = 1
+		. = TRUE
 
 	if(href_list["adjust_temperature"])
 		var/val = href_list["adjust_temperature"]
@@ -137,11 +137,11 @@
 			desired_temp = Clamp(target, 0, 1000)
 		else
 			return FALSE
-		. = 1
+		. = TRUE
 
 	if(href_list["eject_beaker"])
 		eject_beaker(usr)
-		. = 0 //updated in eject_beaker() already
+		. = FALSE //updated in eject_beaker() already
 
 /obj/machinery/chem_heater/ui_interact(mob/user, ui_key = "main", datum/nanoui/ui = null)
 	if(user.stat || user.restrained())
@@ -158,17 +158,17 @@
 	var/cur_temp = beaker ? beaker.reagents.chem_temp : null
 
 	data["targetTemp"] = desired_temp
-	data["targetTempReached"] = ""
+	data["targetTempReached"] = FALSE
 	data["autoEject"] = auto_eject
 	data["isActive"] = on
-	data["isBeakerLoaded"] = beaker ? 1 : 0
+	data["isBeakerLoaded"] = beaker ? TRUE : FALSE
 
 	data["currentTemp"] = cur_temp
 	data["beakerCurrentVolume"] = beaker ? beaker.reagents.total_volume : null
 	data["beakerMaxVolume"] = beaker ? beaker.volume : null
 
 	if (cur_temp)
-		data["targetTempReached"] = round(cur_temp) == round(desired_temp) ? "good" : "average"
+		data["targetTempReached"] = round(cur_temp) == round(desired_temp)
 
 	//copy-pasted from chem dispenser
 	var/beakerContents[0]
--- a/code/modules/reagents/chemistry/machinery/chem_heater.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_heater.dm
@@ -10,6 +10,8 @@
 	var/obj/item/reagent_containers/beaker = null
 	var/desired_temp = T0C
 	var/on = FALSE
+	var/auto_eject = FALSE
+	var/speed_increase = 0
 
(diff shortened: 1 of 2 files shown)


nano/templates/chem_heater.tmpl
```diff
@@ -33,7 +41,7 @@ Used In File(s): \code\modules\reagents\newchem\chem_heater.dm
 				<div style="width: 100%;">
 					{{if data.isBeakerLoaded}}
 						<b>Volume:&nbsp;{{:helper.smoothRound(data.beakerCurrentVolume)}}&nbsp;/&nbsp;{{:data.beakerMaxVolume}}</b><br>
-						<b>Temperature:&nbsp;{{:helper.smoothRound(data.currentTemp)}}&nbsp;Kelvin</b><br>
+						<b>Temperature:&nbsp;<span class="{{:data.targetTempReached}}">{{:helper.smoothRound(data.currentTemp)}}</span>&nbsp;Kelvin</b><br>
```
farie82, line 44: This makes it harder to see what is happening here at a glance.
Best to make this an if statement with a boolean value instead of the classname being passed by the code behind.

code/modules/reagents/chemistry/machinery/chem_heater.dm
```diff
@@ -137,20 +150,26 @@
...
 	data["isActive"] = on
 	data["isBeakerLoaded"] = beaker ? 1 : 0
 
-	data["currentTemp"] = beaker ? beaker.reagents.chem_temp : null
+	data["currentTemp"] = cur_temp
 	data["beakerCurrentVolume"] = beaker ? beaker.reagents.total_volume : null
 	data["beakerMaxVolume"] = beaker ? beaker.volume : null
 
+	if (cur_temp)
+		data["targetTempReached"] = round(cur_temp) == round(desired_temp) ? "good" : "average"
```
farie82, line 171: key/variable name suggests that the value will be a boolean yet is a classname. Best to change it to a boolean for readability in the UI side.

code/modules/reagents/chemistry/machinery/chem_heater.dm
```diff
@@ -30,10 +37,12 @@
 				on = FALSE
 				SSnanoui.update_uis(src)
 				return
-			beaker.reagents.temperature_reagents(desired_temp)
-			beaker.reagents.temperature_reagents(desired_temp)
-			if(abs(beaker.reagents.chem_temp - desired_temp) <= 3)
+			beaker.reagents.temperature_reagents(desired_temp, 35 - speed_increase)
```
farie82, line 40: Should not happen. But maybe add a `max` here to ensure that the divisor is never 0

code/modules/reagents/chemistry/machinery/chem_heater.dm
```diff
@@ -115,6 +124,10 @@
 		on = !on
 		. = 1
 
+	if(href_list["toggle_autoeject"])
+		auto_eject = !auto_eject
+		. = 1
```
farie82, line 129: `TRUE` here. it's a boolean value so the `TRUE`/`FALSE` defines are prefered.

nano/templates/chem_heater.tmpl
Shendow: I've added newlines and replaced the class name by a ternary, resulting in this:
```
<b>
	Temperature:&nbsp;
	<span class="{{:data.targetTempReacheddata.targetTempReached ? 'good' : 'average'}}">{{:helper.smoothRound(data.currentTemp)}}</span>&nbsp;Kelvin
</b><br>
```

I'm not sure if this is what you meant by an if statement or you really want me to use one. If it's the latter, I'm not sure it's good because it would result in duplicate code.

code/modules/reagents/chemistry/machinery/chem_heater.dm
Shendow: `targetTempState` would have probably been a more suitable key in hindsight, but I changed the value type to a boolean as per your previous comment to make the template code easier to read, using a ternary operator.

code/modules/reagents/chemistry/machinery/chem_heater.dm
Shendow: Rookie mistake from my part, done

nano/templates/chem_heater.tmpl
farie82: Aye this is a good way to solve it. Looks good!

code/modules/reagents/chemistry/machinery/chem_heater.dm
Shendow: Done, I also went and replaced the other 1/0's in the file with the defines

farie82: Very nice. Thanks for the added refactor!

AffectedArc07:
Does this make it so that a heater set to 1000K wont heat a bajillion times faster than one set to 374K

dearmochi:
> 
> 
> Does this make it so that a heater set to 1000K wont heat a bajillion times faster than one set to 374K

No, that behavior is still there and that's because of the temperature proc:
```
/datum/reagents/proc/temperature_reagents(exposed_temperature, divisor = 35, change_cap = 15) //This is what you use to change the temp of a reagent holder.
	//Do not manually change the reagent unless you know what youre doing.
	var/difference = abs(chem_temp - exposed_temperature)
	var/change = min(max((difference / divisor), 1), change_cap)
```

I'm no thermodynamics expert, but if the machine blasts something cold at 1000K (that's the intended limit of the ChemHeater by the way), I believe it would heat up faster than blasting it at 500K would, so I think keeping it like that is the better idea.

A band-aid fix would be to pass the `change_cap` argument to be something smaller like 5, so that massive temperature differences won't cause the reagents' temperature to change as quickly.
Another option which is more far-reaching would be to replace the increased based on difference by a linear increase.

DashCee:
This is still going to result in the ChemHeater being almost entirely useless due to lighters being able to heat solutions multiple times faster, using less clicks, without having to worry about setting temperature to the maximum then back down again to speed up the process. 

This is definitely better than current heater though, and makes cooling of solutions less of a pain, especially due to the auto-eject and timer audio.

AffectedArc07:
Doesnt https://github.com/ParadiseSS13/Paradise/pull/13738 replace this?

dearmochi:
Closing this as it is now included in #13738 because NanoUI must die

PR #13532, opened 2020-06-02
Removes something which shouldnt really exist here

AffectedArc07:
## What Does This PR Do
This PR removes the snake body accessory. Yes this was admin only, but it was also something which shouldnt exist in this code. It doesnt fit in with everything else, nor does anyone use it anyway.

## Why It's Good For The Game
This shouldnt exist in our code

## Changelog
:cl: AffectedArc07
del: Removed the human-snake body modification monster thingy
/:cl:


Mitchs98:
I actually see it used semi-often, and have seen it used in a few events. Don't see why it shouldn't exist, and it is used, just not all the time.

MRgaty:
I, on the other hand, have seen it used once in my entirely playtime, and it was just with a random vox getting it cuz of memes.

Spacemanspark:
Imo it has no actual use, it was added as a joke about fetishes, and frankly makes no sense being here.

I'm in support of its removal.

PR #13534, opened 2020-06-04
Organ processing performance

Fox-McCloud:
This is a minor optimization, it's nothing special:

It mostly just reduced proc calls, puts cheaper proc calls first so more expensive ones don't run, and makes it a bit more OOP.

Testing revealed it made organ `process`ing about ~16% cheaper.

:cl: Fox McCloud
tweak: Massive doses of spaceacillin no longer prevents organ necrosis from ever happening.
tweak: Buffs spaceacillin just a hair by making it kill a couple more germs per cycle
tweak: Improves processing of organs so it improves game performance
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine.dm
@@ -173,6 +173,29 @@
 	metabolization_rate = 0.2
 	taste_description = "antibiotics"
 
+/datum/reagent/medicine/spaceacillin/on_mob_life(mob/living/M)
+	var/list/organs_list = list()
+	if(iscarbon(M))
+		var/mob/living/carbon/C = M
+		organs_list += C.internal_organs
+
+	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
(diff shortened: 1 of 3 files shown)


Fox-McCloud:
Made a minor alteration. It makes it more OOP---handles the organ germ level reduction on `spaceacillin` instead of on the external organ.

This does technically mean the loss of a few minor things---but I don't think it's worth quibbling over.

If someone has an infection and had spacecillin in them, it'd prevent the germ level from going up (if you had a particular volume)---I just removed this, completely; I don't think it's really going to do much of anything in most cases and it means we don't have to do a reagent check for every gosh darn organ in existence.

KasparoVy:
Lmao. Reminds me of the old code bay space mining days and how often we'd get full body infections or obscene shit like septic brains and wonder how we're alive.

Fox-McCloud:
@KasparoVy those were crazy times 

"Hmm, yes, your brain is dead, but you're still talking---let's not think about that too much--now into surgery".

Also, hit me up on Discord, if you don't mind, and aren't busy.

PR #13545, opened 2020-06-05
Nerfs Trashbags

Fox-McCloud:
Trashbags have become problematic, to say the least.

They're basically vastly superior toolbelts; the bluspace trashbag is even worse.

Either case, this nerfs them in two ways:

- Their weight class is always bulky, which makes means you can no longer do the meme of putting a trashbag in a trashbag then one from inside the other. It also means you can't carry it around in your backpack.
- They no longer have a slot flag, meaning you can't wear them on your belt. You'll have to carry it around in your hands.

Trashbag is a janitorial tool that's mean to make life easier for the janitor--and it accomplishes that, but it absolutely should not be the best toolbelt item in the game.

This also fixes regular trashbags only being able to hold 14 items.

:cl: Fox McCloud
tweak: Trashbags no longer fit on belts
tweak: Trashbags are always bulky, even when empty
fix: Fixes regular trashbags only being able to hold 14 items.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/bags.dm
+++ b/code/game/objects/items/weapons/storage/bags.dm
@@ -33,9 +33,11 @@
 	icon_state = "trashbag"
 	item_state = "trashbag"
 
-	w_class = WEIGHT_CLASS_TINY
+	w_class = WEIGHT_CLASS_BULKY
 	max_w_class = WEIGHT_CLASS_SMALL
+	slot_flags = null
 	storage_slots = 30
+	max_combined_w_class = 30
 	can_hold = list() // any
 	cant_hold = list(/obj/item/disk/nuclear)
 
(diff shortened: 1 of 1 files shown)


datlo:
Good change. Trashbag and trashbags of holding having any use outside of janitorial duties is just silly, and the powergaming is insane. 

'Waaah muh belt storage' : Just get a belt of holding and bag of holding. You can already stuff half the station in your backpack with these combined.

AffectedArc07:
> For janitor players : Trashbags can fit on both the ridable cart and the bucket janicart. Store it there.

I was about to ask this, and thanks for clearing it up

lolzyking:
This is good

Shockpoint:
> tweak: Trashbags are always bulky, even when empty

"Help, my trashbag is too heavy. It's empty but it's still heavy as fuck." - No one ever

lolzyking:
> > tweak: Trashbags are always bulky, even when empty
> 
> "Help, my trashbag is too heavy. It's empty but it's still heavy as fuck." - No one ever

Suggestion give trashbags a folded up state like body bags do that allows for easy carrying?

Mitchs98:
Disagree with it no longer fitting on belt. But otherwise fine with this. I myself only use it as a janitor and having it on your belt slot really helps and is a lot more intuitive than having to mess with the janicart UI.

Carthusia:
Is it possible to make it only mountable on the janibelt? Up until now I hadnt even heard of the issues described above but a non-belt garbage bag kind of makes it a moot item for alot of janitors.

Personally as a janitor I dont really use the hand cart or the janicart on average. 
The hand cart due to it being mostly a mop based object and eveyrthing it can carry fitting into my bag and its ease of being stolen, and the janicart due to it being slower than running, with the only saving grace of the janicart being its buffer upgrade making it at least be viable.

Having it on the belt, and then using it to pickup a bunch of trash, unloading the trash then clipping it back on the belt makes the job quite a bit easier, not having that, for me anyways, makes it kind of a non-option in a majority of cases

Mitchs98:
Yeah I wouldn't even use it at all if I had to carry it around like that, it's such a hassle to drag the cart around and the janicarts slowness hurts my soul.

PR #13546, opened 2020-06-05
Reduces Storage Creep

Fox-McCloud:
Idea from TG.

Storage has powercreeped greatly, over the years, to say the least, with this being a list of the buffs it has received:

- Item total storage space based on the size of the items instead of being a fixed 7 slots
- Proliferation of Easy to access storage solutions to get even more storage
- Can access storage within storage

Needless to say, the amount of items that you can carry on your person is a bit ridiculous compared to what it used to be---and, as such, people tend to carry everything but the kitchen sink with them, at all times.

Andddd then you have the bag of holding, which can hold just about everything and has a ridiculous number of storage slots and a gigantic weight class total limit.

And the belt of holding---which is in the same vein, only for a belt.

This alters the bag of holding to require an anomaly core to create; you can print off an inert bag of holding, but it's useless until you combine it with an anomaly core.

Bag of holding has been buffed a bit to compensate for its increased rarity; it'll hold literally any item in the game now (that you can pick up anyway).

The belt of holding has been straight up removed from R&D (it still exists for adminbus and such).

This should dramatically cut down on the extreme cases of storage creep while still leaving day to day operations largely untouched. Carrying everything makes difficult things like Lavaland even easier and makes carrying 5 guns with 10 batters and 20 mags easy as pie.

:cl: Fox McCloud
tweak: Bluespace bags of holding now require an anomaly core to create; print off an inert bag of holding at R&D and use crafting to make a genuine bag of holding
del: R&D can no longer make belts of holding
/:cl:


lolzyking:
Can we hold off on the anomaly core requirement until AFTER the SM rework, which will make anomaly farming a realistic, if not dangerous endeavor?

Kyep:
How does reducing the amount of storage players have access to, result in more fun and interesting gameplay?
I'm not sure it does.

TDSSS:
Well, if picking and chosing what you carry with you isn't interesting gameplay, why do we have storage restrictions at all?

Denthamos:
Kind of agree with Lolzy on this one.  Storage has increased heavily over the years but to just remove it (which basically this would be as it would become so heavily rare, because why spend a core on that over phazon) seems too much.  When the SM rework/engie remap goes through and gives the crew a method to farm anomalies (at the danger of blowing stuff up or worse) I feel this would work nicely.

Fox-McCloud:
>How does reducing the amount of storage players have access to, result in more fun and interesting gameplay?
I'm not sure it does.

I don't think this is a very good argument, for this reason:

"how does reducing the number of guns players have access to by putting them in lockboxes result in more fun and interesting gameplay?"
"How does reducing the number of stun chems players have access to result in more fun and interesting gameplay?"
"How does reducing the strength of loot players earn result in more fun and interesting gameplay".

On the surface, it sounds ok, but there's a bigger picture---storage being limited means players have to make meaningful choices in what they carry---there's a trade off between carrying lots of weapons or lots of medical supplies, for instance. When storage is unlimited (or incredibly high, as is the case of the bag of holding), then there's not really much of a trade-off; you might as well carry anything and everything.

Why not carry 5 guns, 3 medkits, and a surgery set in your backpack? Why not have 5 bluspace trashbags each filled with 60 legion cores?

Resources being limited makes them more meaningful--especially in a game like SS13---it also prevents more blatant/obvious abuses with said resources.

code/modules/crafting/recipes.dm
```diff
@@ -534,6 +534,14 @@
...
 	category = CAT_MISC
 
+/datum/crafting_recipe/boh
+	name = "Bag of Holding"
+	reqs = list(
+            /obj/item/bag_of_holding_inert = 1,
+            /obj/item/assembly/signaler/anomaly = 1)
+	result = /obj/item/storage/backpack/holding
+	category = CAT_CLOTHING
+
```
AffectedArc07, line 544: Please make it so you can just click the inert bag of holding with an anomalie core to turn it into a working BOH

ZomgPonies:
I think if we get the supermatter rework merged first, which would make farming cores a more regular thing, then this level of balance would make sense.

Fox-McCloud:
Going to wait for the anomaly thing and techwebs before resubmitting.

PR #13547, opened 2020-06-05
Superfast automatic nanomap generation, non OOM edition

AffectedArc07:
## What Does This PR Do
This PR completley rewrites the nanomap generator for it to exist inside RUSTG. This allows it to be threaded alongside BYOND and to not hog DD while generating the map.

This PR also adds in the framework for autogenerating nanomaps after map edits, and caching the generated maps unless they need to be re-generated.

This PR modifies RUSTG, and the respective PR is open that https://github.com/ParadiseSS13/rust-g/pull/2

image
image

## Why It's Good For The Game
The current nanomap generator takes 50 seconds to process, locks up the entire game, and must be ran manually. I was unhappy with all of that so I did something about it.

## Changelog
:cl: AffectedArc07
add: Nanomaps will auto generate when the station map is updated
add: Nanomap generation no longer take 50 seconds to complete
add: Nanomap generation no longer locks up the entire game process while running
/:cl:


AffectedArc07:
Turns out this fails if it tries to render twice in the same DD session. Even though we kill DD here after each round, others may not, so it needs fixing.
image

AffectedArc07:
The threading issue has been fixed, but now theres an issue of the server spuriously dying when a DLL call is made
```
[2020-06-08T21:59:24] ADMIN: AffectedArc07/(AffectedArc07) called NanoMap's render_map() with no arguments
[2020-06-08T21:59:24] WORLD: Started render of NSS Cyberiad.
(Server Restart)
[2020-06-08T22:02:27] Starting up.
[2020-06-08T22:02:27] -------------------------
[2020-06-08T22:02:27] WORLD: World loaded at 2020-06-08T23:02:27
```

I will try to figure it out, but the nerds in coderbus have had the same issue with modifications to their rust DLL, and its *likely* a `STATUS_STACK_BUFFER_OVERRUN`

*Im not a scientist or expert I am just going off what was said*

AffectedArc07:
RUST_G now invokes `SpacemanDMM` under a shell call, which is fully handled by RUSTG instead of being a BYOND shell call, which allows it to still be asynchronous. Since its a spawned process, it will use its own RAM allowance and separate core to the main game, and even allows the spawned application to be 64 bit. Rendering applications for windows & linux have been included.

AffectedArc07:
I will make this not gameserver dependant with the power of magic

PR #13549, opened 2020-06-06
Refactors and reworks transit tube pods. Maint drones can now also use them

### LINKED ISSUE #10111: Ancient Station : Hivebot can get stuck in travel tube
**Problem Description**:
What is the problem?
Hivebot can path into travel tube and get stuck in them. They can't leave and you can't force them out/destroy them.

Why do you think this is a bug?

Expected to bash bots and enjoy my time, then return to solars to get power to RND

**What happened instead**:

Couldn't return to solars without admin intervention/breaking a window to space

Blocks the way back to the rest of the station. Probably not exclusive to ancient station, just never relevant on the cyberiad.

Step 1. Fight hivebots in front of the travel tube, hivebots swarm into the room and one of them path into the tube.
Step 2. Tube is now useless.

If your report is about something that used to work but no longer does. When was the last time you remember it working?
First test of ancient station on para live servers.

anything else out of the ordinary happening?)**: Anything else you can tell us.

farie82:
## What Does This PR Do
1. Refactors the transit tubes. Dear god what a shit code
2. The transit pods will now wait 5 seconds after arriving before they launch again.
3. Removing things from a pod now takes 2 seconds instead of 4.
4. Putting somebody in a pod using a grab now takes 3 seconds instead of 6.
5. You can't open/close pods yourself anymore. This is pure automated
6. Fixes that you can't use a pod when another pod is already at the station and is full
7. Pods will now eject all mindless things when arriving
8. The pod will have an icon_state when it's full
9. Maint drones can now also use them

Fixes #10111

## Why It's Good For The Game
1. Shit code gone is gut.
2. Pods could get stuck with people inside of them. Or objects and they would be virtually unusable.
3. 4 seconds is way to long since the launch delay is 5 seconds. This wasn't even possible before hand.
4. Same here.
5. This will keep the pods in sync. Instead of them clogging up on one side.
6. It was a bug
7. Fixes the bug
8. User friendlyness
9. Just a QOL for drone mains

## Changelog
:cl:
tweak: Transit pods now eject all mindless things when arriving at a station
tweak: You can't open/close transit pods
tweak: Removing things from a transit pod now takes 2 seconds instead of 4.
tweak: Putting somebody in a transit pod now takes 3 seconds instead of 6.
tweak: Transit pods actually wait 5 seconds now after arriving before launching again. Making the above possible
fix: Having multiple pods at a transit station now won't block you from going in the last one if the first one is full
add: Transit pods now have an full icon state
/:cl:

### DIFF
--- a/code/game/objects/structures/transit_tubes/station.dm
+++ b/code/game/objects/structures/transit_tubes/station.dm
@@ -1,3 +1,6 @@
+#define CLOSE_DURATION 6
+#define OPEN_DURATION 6
+
 
 // A place where tube pods stop, and people can get in or out.
 // Mappers: use "Generate Instances from Directions" for this
@@ -8,13 +11,12 @@
 	icon_state = "closed"
 	exit_delay = 1
 	enter_delay = 2
-	var/pod_moving = 0
+	var/pod_moving = FALSE
 	var/cooldown_delay = 50
 	var/launch_cooldown = 0
-	var/reverse_launch = 0
+	var/reverse_launch = FALSE
+	var/hatch_state = TRANSIT_TUBE_CLOSED
 
-	var/const/OPEN_DURATION = 6
-	var/const/CLOSE_DURATION = 6
 	var/list/disallowed_mobs = list(/mob/living/silicon/ai)
 
 /obj/structure/transit_tube/station/New()
@@ -27,20 +29,22 @@
 
 // Stations which will send the tube in the opposite direction after their stop.
 /obj/structure/transit_tube/station/reverse
-	reverse_launch = 1
+	reverse_launch = TRUE
 
 /obj/structure/transit_tube/station/should_stop_pod(pod, from_dir)
-	return 1
+	return TRUE
 
-/obj/structure/transit_tube/station/Bumped(mob/AM as mob|obj)
-	if(!pod_moving && icon_state == "open" && istype(AM, /mob/living) && !is_type_in_list(AM, disallowed_mobs))
+/obj/structure/transit_tube/station/Bumped(mob/living/L)
+	if(!pod_moving && hatch_state == TRANSIT_TUBE_OPEN && isliving(L) && !is_type_in_list(L, disallowed_mobs))
+		var/failed = FALSE
 		for(var/obj/structure/transit_tube_pod/pod in loc)
 			if(pod.contents.len)
-				to_chat(AM, "<span class=The pod is already occupied.</span>")
-				return
+				failed = TRUE
 			else if(!pod.moving && (pod.dir in directions()))
-				AM.forceMove(pod)
+				L.forceMove(pod)
 				return
+		if(failed)
+			to_chat(L, "<span class='warning'>The pod is already occupied.</span>")
 
 
 
@@ -48,13 +52,13 @@
 	if(!pod_moving)
 		for(var/obj/structure/transit_tube_pod/pod in loc)
 			if(!pod.moving && (pod.dir in directions()))
-				if(icon_state == "closed")
-					open_animation()
+				if(hatch_state == TRANSIT_TUBE_CLOSED)
+					open_hatch()
 
-				else if(icon_state == "open")
+				else if(hatch_state == TRANSIT_TUBE_OPEN)
 					if(pod.contents.len && user.loc != pod)
 						user.visible_message("<span class='warning'>[user] starts emptying [pod]'s contents onto the floor!</span>")
-						if(do_after(user, 40, target = src)) //So it doesn't default to close_animation() on fail
+						if(do_after(user, 40, target = src)) //So it doesn't default to close_hatch() on fail
 							if(pod.loc == loc)
 								for(var/atom/movable/AM in pod)
 									AM.loc = get_turf(user)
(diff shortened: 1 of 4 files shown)


datlo:
Since you are refactoring transit tubes, could you add something so that this cannot happen? https://github.com/ParadiseSS13/Paradise/issues/10111

Not sure what the best way is... check if simple animals have a mind/ckey to let them use the tube?

farie82:
> Since you are refactoring transit tubes, could you add something so that this cannot happen? #10111
> 
> Not sure what the best way is... check if simple animals have a mind/ckey to let them use the tube?

That one is fixed by this PR. The bot will now be pushed out when they arrive at the other side. Thus fixing it

code/game/objects/structures/transit_tubes/station.dm
```diff
@@ -27,34 +29,36 @@
...
 /obj/structure/transit_tube/station/reverse
-	reverse_launch = 1
+	reverse_launch = TRUE
 
 /obj/structure/transit_tube/station/should_stop_pod(pod, from_dir)
-	return 1
+	return TRUE
 
-/obj/structure/transit_tube/station/Bumped(mob/AM as mob|obj)
-	if(!pod_moving && icon_state == "open" && istype(AM, /mob/living) && !is_type_in_list(AM, disallowed_mobs))
```
AffectedArc07, line 36: Dear god why

code/game/objects/structures/transit_tubes/station.dm
```diff
@@ -80,63 +84,76 @@
...
 /obj/structure/transit_tube/station/pod_stopped(obj/structure/transit_tube_pod/pod, from_dir)
-	pod_moving = 1
-	spawn(5)
-		launch_cooldown = world.time + cooldown_delay
-		open_animation()
-		sleep(OPEN_DURATION + 2)
-		pod_moving = 0
-		pod.mix_air()
+	pod_moving = TRUE
+	addtimer(CALLBACK(src, .proc/pod_stopped_callback, pod), 5)
```
AffectedArc07, line 143: Death to `spawn()`

procdrone:
oh god thanks

putting someone arrested in a tube and they NOT opting to come out with NO way to eject them... same with dying while in transit. body lost forever.

Fox-McCloud:
I wouldn't force eject upon arriving.

There's some maps on other stations that have multiple stops. In these circumstances, you don't *want* to be ejected until you choose to do so.

Heck, even we had this on the science outpost, when it still existed.

farie82:
> I wouldn't force eject upon arriving.
> 
> There's some maps on other stations that have multiple stops. In these circumstances, you don't _want_ to be ejected until you choose to do so.
> 
> Heck, even we had this on the science outpost, when it still existed.

I'm trying to think about something that would fix it in another way.
Longer cooldown when starting to move when the pod is filled maybe? Combine that with an actual icon_state when the pod is filled and making it less clunky to get people out of one will probably be enough to help players get dead/unconcious/restrained etc people out of them.

For the actual bug it fixes. I'm not 100% sure. It can check on mind I guess. If it's a bump. But that feels cheesy. Maybe mobs without a mind will eject at the first station (simulating exploration... ish).

I'll implement the above for now

farie82:
Changed the PR description.
Pods now have a delay before launching and are fully automated. They will eject mindless things on arrival. Using them is actually doable now and they have an icon_state when occupied. Still awaiting a better one since this one is rather badly visible.

rebel228:
Is there any reason why maintenance drones can't enter the pods?

farie82:
> Is there any reason why maintenance drones can't enter the pods?

They should be able to. Nothing in the bumped code itself forbids them to

farie82:
PR updated. Drones can now also use the tubes

code/game/objects/structures/transit_tubes/station.dm
```diff
@@ -27,116 +29,128 @@
...
 			else if(!pod.moving && (pod.dir in directions()))
-				AM.forceMove(pod)
+				pod.move_into(L)
 				return
+		if(failed)
+			to_chat(L, "<span class='warning'>The pod is already occupied.</span>")
 
 
 
 /obj/structure/transit_tube/station/attack_hand(mob/user as mob)
```
SteelSlayer, line 51: The formatting of the logic in this proc is SO disgusting and hard to read, it works but.. bonus points if you refactor it with early returns, continues, more multi conditional ifs etc.

SteelSlayer (approved):
Besides the little optional change I wrote, LGTM.

code/game/objects/structures/transit_tubes/station.dm
farie82: Yep ugly shit. Refactored that as well

Alex1222003:
As a Maintenance Drone main i thank you. Normally if i had to get in the AI sat i had to use the airlock, but if the AI sat was out of power i had to dismantle the windows to get in, causing a bit of a nuisance. (I'd rather not cause a big ruckus by using the teleporter, i don't know what someone passing by would think if they saw an active teleporter set to AI sat.)

AffectedArc07 (approved):

PR #13557, opened 2020-06-06
Re-adds NTTC filtering

AffectedArc07:
## What Does This PR Do
This PR re-adds NTTC filtering. I originally removed this due to lack of use, yet I have seen three occasions myself where this would have come in really useful, hence the re-addition. You just add names into the filter list, and they are then blocked from speaking on comms.

image
image

## Why It's Good For The Game
Removing the feature wasnt even the best idea, and its useful to be able to shut up comms-spamming greytide.

## Changelog
:cl: AffectedArc07
add: Re-added NTTC filtering
/:cl:

### DIFF
--- a/code/game/machinery/tcomms/core.dm
+++ b/code/game/machinery/tcomms/core.dm
@@ -1,5 +1,6 @@
 #define UI_TAB_CONFIG "CONFIG"
 #define UI_TAB_LINKS "LINKS"
+#define UI_TAB_FILTER "FILTER"
 
 /**
   * # Telecommunications Core
@@ -80,6 +81,11 @@
 
 	// Now we can run NTTC
 	tcm = nttc.modify_message(tcm)
+
+	// If the signal shouldnt be broadcast, dont broadcast it
+	if(!tcm.pass)
+		// We still return TRUE here because the signal was handled, even though we didnt broadcast
+		return TRUE
 
 	// Now we generate the list of where that signal should go to
 	tcm.zlevels = reachable_zlevels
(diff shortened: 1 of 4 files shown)


procdrone:
Thank you

pinatacolada:
Does it have to fully match or ir is partial? Like would that work 'Vul' work for 'Vulp'. Is it case sensitive?

AffectedArc07:
Must be a fun match not partial

code/game/machinery/tcomms/core.dm
```diff
@@ -257,9 +266,35 @@
 			to_chat(usr, "<span class='notice'>Successfully changed password from <b>[link_password]</b> to <b>[new_password]</b>.</span>")
 			link_password = new_password
 
+	if(ui_tab == UI_TAB_FILTER)
+		if(href_list["add_filter"])
+			// This is a stripped input because I did NOT come this far for this system to be abused by HTML injection
+			var/name_to_add = stripped_input(usr, "Enter a name to add to the filtering list", "Name Entry")
+			if(name_to_add == "")
+				return
+			if(name_to_add in nttc.filtering)
```
farie82, line 275: Performance wise it would be better to use an assoc list. Where the key is the name and the value is just 1. O(1) complexity VS O(n) complexity.

code/game/machinery/tcomms/core.dm
```diff
@@ -257,9 +266,35 @@
...
+				to_chat(usr, "<span class='alert'><b>ERROR:</b> User already in filtering list.</span>")
+			else
+				nttc.filtering |= name_to_add
+				log_action(usr, "has added [name_to_add] to the NTTC filter list on core with ID [network_id]", TRUE)
```
farie82, line 279: Side thing. But can you add logging using the new logging system to the log_action proc?
So just `user.create_log(...)`

AffectedArc07: This is a specific proc to NTTC with flags of whether it should admin log. If you wish, you can modify `log_action` directly to add that in.

code/game/machinery/tcomms/core.dm
AffectedArc07: Considering this list is going to rarely be used, and even when it is, it will have about 2 people inside it, I dont see this as being a huge issue, and I would prefer this for readability.

PR #13558, opened 2020-06-06
Adds the slippery component

SteelSlayer:
## What Does This PR Do
Loosely ports TG's slippery component, but its been heavily modified for paradise, and has some of my own additions as well, namely creating a base `after_slip()` proc for *all* movable atoms.

There is no CL because this is a backend change only.

## Why It's Good For The Game
Cuts down on `obj` vars related to slipping that go unused for a majority of items. More OOP.

### DIFF
--- a/code/datums/components/slippery.dm
+++ b/code/datums/components/slippery.dm
@@ -1,0 +1,54 @@
+/**
+	This is a component that can be applied to any movable atom (mob or obj).
+
+	While the atom has this component, any carbon mob that walks over it will slip.
+	Duration, tiles moved, and so on, depend on what variables are passed in when the component is added.
+*/
+/datum/component/slippery
+	/// Text that gets displayed in the slip proc, i.e. "user slips on [description]"
+	var/description
+	/// The amount of stun to apply after slip.
+	var/stun
+	/// The amount of weaken to apply after slip.
+	var/weaken
+	/// The chance that walking over the parent will slip you.
+	var/slip_chance
+	/// The amount of tiles someone will be moved after slip.
+	var/slip_tiles
+	/// TRUE If this slip can be avoided by walking.
+	var/walking_is_safe
+	/// TRUE if having no slip shoes makes you immune to this slip.
+	var/noslip_is_immune
+	/// The verb that players will see when someone slips on the parent. In the form of "You [slip_verb]ped on".
+	var/slip_verb
+
+/datum/component/slippery/Initialize(_description, _stun = 0, _weaken = 0, _slip_chance = 100, _slip_tiles = 0, \
+									_walking_is_safe = TRUE, _noslip_is_immune = TRUE, _slip_verb = "slip")
+	if(!ismovable(parent))
+		return COMPONENT_INCOMPATIBLE
+
+	description = _description
+	stun = max(0, _stun)
+	weaken = max(0, _weaken)
+	slip_chance = max(0, _slip_chance)
+	slip_tiles = max(0, _slip_tiles)
+	walking_is_safe = _walking_is_safe
+	noslip_is_immune = _noslip_is_immune
+	slip_verb = _slip_verb
+
+/datum/component/slippery/RegisterWithParent()
+	RegisterSignal(parent, list(COMSIG_MOVABLE_CROSSED, COMSIG_ATOM_ENTERED), .proc/Slip)
+
+/datum/component/slippery/UnregisterFromParent()
+	UnregisterSignal(parent, list(COMSIG_MOVABLE_CROSSED, COMSIG_ATOM_ENTERED))
+
+/**
+	Called whenever the parent recieves either the `MOVABLE_CROSSED` signal or the `ATOM_ENTERED` signal.
+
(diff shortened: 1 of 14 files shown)


code/datums/components/slippery.dm
```diff
@@ -0,0 +1,54 @@
...
+	/// The chance that walking over the parent will slip you.
+	var/slip_chance
+	/// The amount of tiles someone will be moved after slip.
+	var/slip_tiles
+	/// TRUE If this slip can be avoided by walking.
+	var/walking_is_safe
+	/// TRUE if having no slip shoes makes you immune to this slip.
+	var/noslip_is_immune
+	/// The verb that players will see when someone slips on the parent. In the form of "You [slip_verb]ped on".
+	var/slip_verb
```
AffectedArc07, line 23: The documentation is great for the var definitions, but not for the definition.

While I would love to suggest reading up on the doucmentation guide, its not compulsory here so its no huge issue

SteelSlayer: I can make this *not* autodoc'd if you want, as in `/* bla */`

AffectedArc07:
```
/**
	This is a component that can be applied to any movable atom (mob or obj).
	While the atom has this component, any carbon mob that walks over it will slip.
	Duration, tiles moved, and so on, depend on what variables are passed in when the component is added.
*/
```

Should be
```
/**
  * # Slip Component
  *
  * This is a component that can be applied to any movable atom (mob or obj).
  *
  * While the atom has this component, any carbon mob that walks over it will slip.
  * Duration, tiles moved, and so on, depend on what variables are passed in when the component is added.
  *
  */
```

but again, doc isnt required here so its up to you

SteelSlayer: Eh I really hate that asterisk per line format but, If that's how its suppose to be, why not

Fox-McCloud:
`but its been heavily modified for paradise, `

Whenever you do a port, but heavily modify it, you should enumerated differences.

Some of us have a high familiarity with TG code and what it does, so we expect functionality to be the same. It's a pain in the ass when we have to manually find the differences ourselves.

code/datums/components/slippery.dm
```diff
@@ -0,0 +1,57 @@
...
+
+/datum/component/slippery/Initialize(_description, _stun = 0, _weaken = 0, _slip_chance = 100, _slip_tiles = 0, \
+									_walking_is_safe = TRUE, _noslip_is_immune = TRUE, _slip_verb = "slip")
```
Fox-McCloud, line 29: Just put this all on teh same line, please.

code/datums/components/slippery.dm
```diff
@@ -0,0 +1,57 @@
...
+	if(!ismovable(parent))
```
Fox-McCloud, line 30: `ismovableatom`

We'll have to wait until 513 stable and use updating to that, officially before we can do this.

code/datums/components/slippery.dm
SteelSlayer: Sure

code/datums/components/slippery.dm
SteelSlayer: Got it.

SteelSlayer:
Sorry, I can list out the differences from TG. 
- All of the component's variables are different so they correspond with the variables that are removed from `item`'s and the arguments for the `slip` proc. Such as the slip chance and amount of tiles moved.
- This: `RegisterSignal(parent, COMSIG_ITEM_WEARERCROSSED, .proc/Slip_on_wearer)` and the proc in there have been removed, because we don't have this signal.
- I removed the `callback` var because there didn't seem to be a use for it. However, I was exploring around TG the other day and I saw that its used for `wet_floor` component so I may need to add that back in, I don't know.
- The creation of the `atom/movable/after_slip()` proc was my idea.
- TG has a `!victim.is_flying()` check among the other checks on my line 55. But I removed it since we check that in the `slip` proc anyway, and `slip` is used outside this component.

code/datums/components/slippery.dm
SteelSlayer: I realized this should be `isatom` because I think this component is made to be placed on turfs as well.

code/game/atoms_movable.dm
```diff
@@ -330,6 +330,16 @@
...
 
+/**
+	Base proc, intended to be overriden.
+
+	This should only be called from one place: inside the slippery component.
+	Called after a human mob slips on this atom.
+
+	If you want the person who slipped to have something special done to them, put it here.
+*/
+/atom/movable/proc/after_slip(mob/living/carbon/human/H)
```
Fox-McCloud, line 341: if the component is atom level, this should be as well.

Fox-McCloud:
Last thing. 

Our components aren't 100% up to date with TG's, so we can't make use of named args.

https://github.com/tgstation/tgstation/pull/49098

Make sure that everything works without utilizing named args.

SteelSlayer:
I doubled checked and did a final test, all the args should be in right spot for all the items.

Issue #13550, opened 2020-06-06
PDAs getting destroyed and initialized forever


SteelSlayer:
**Issue Description**:
PDAs are initializing forever. Its a bit hard to see from the picture, but this stuff just goes on and on every second. Apparently they are constantly getting destroyed and respawned? 

This is from my test server obviously, where I just spawned in by myself. There are no other players online, so I'm not even sure where all these PDAs are coming from.

![image](https://user-images.githubusercontent.com/42044220/83938065-4c496480-a797-11ea-835f-fa86f1952dad.png)

![image](https://user-images.githubusercontent.com/42044220/83938400-71d76d80-a799-11ea-9390-6440a3ce2bb6.png)

**What did you expect to happen**: 
For each PDA to initialize on spawn only.

**Why is this bad/What are the consequences**:
The server is wasting resources for no reason.

**Steps to reproduce the problem**:
Add a `message_admins()` line somewhere in the `/obj/item/pda/Initialize()` proc and observe the spam.

**When did the problem start happening**:
I have no idea.

PR #13561, opened 2020-06-07
Fixes Excessive PDA Creation

Fox-McCloud:
Good heavens.

First off, chameleon items were not properly GC'ing---their action datum was getting added to the processing SS, but never removed from it, causing it to be hard deleted if it was EMP'd or is a broken chameleon item.

Furthermore, the chameleon action itself wasn't properly deleted when a chameleon item was deleted.

Next up, how chameleon items change their states is just plain awful---it has to create an item just to fetch values from it . This is partially understandable because of Bay's awful species icon system, but it wasn't necessary for every single chameleon item when it was changing.

Thirdly, you could make the broken chameleon and chameleon PDA icon states for the PDA printer...not...needed or great.

And fourthly, PDA printers create every single PDA type, minus a few blacklisted ones, just got get their icon state. 

*AWFUL*.

What's changed:
- Chameleon action should GC
- Chameleon items should GC
- How chameleon item changes has been refactored to utilize `initial`, where possible; if something has a `sprite_sheet` set, then it will create the item and copy the list (sadly `initial` doesn't work with lists)---a lot of items have `sprite_sheet` set, but some don't at all (like PDAs)
- PDA printers no longer create PDAs, but instead have an associative list of the PDA icon states to select from and the description of said PDA---functionality is identical without the nonsense of needing to create PDAs. 

The lack of the broken chameleon PDA being blacklisted meant that a broken chameleon PDA was created at round start, then never deleted----and with how icon changing was added, it'd just create a new PDA every 2 seconds to get its icon state and description. AWFUL.

Fixes: https://github.com/ParadiseSS13/Paradise/issues/13550

Performance PR since it makes chameleon handling perform better and also means PDA's aren't being created like crazy, every 2 seconds.

:cl: Fox Mcloud
fix: Fixes thousands of PDA's being created in a round, decreasing game performance
/:cl:

### DIFF
--- a/code/game/machinery/PDApainter.dm
+++ b/code/game/machinery/PDApainter.dm
@@ -30,15 +30,11 @@
 /obj/machinery/pdapainter/New()
 	..()
 	var/blocked = list(/obj/item/pda/silicon/ai, /obj/item/pda/silicon/robot, /obj/item/pda/silicon/pai, /obj/item/pda/heads,
-						/obj/item/pda/clear, /obj/item/pda/syndicate)
+						/obj/item/pda/clear, /obj/item/pda/syndicate, /obj/item/pda/chameleon, /obj/item/pda/chameleon/broken)
 
-	for(var/P in typesof(/obj/item/pda)-blocked)
-		var/obj/item/pda/D = new P
-
-		//D.name = "PDA Style [colorlist.len+1]" //Gotta set the name, otherwise it all comes up as "PDA"
-		D.name = D.icon_state //PDAs don't have unique names, but using the sprite names works.
-
-		src.colorlist += D
+	for(var/thing in typesof(/obj/item/pda) - blocked)
+		var/obj/item/pda/P = thing
+		colorlist[initial(P.icon_state)] = initial(P.desc)
 
 /obj/machinery/pdapainter/Destroy()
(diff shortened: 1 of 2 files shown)


code/game/machinery/PDApainter.dm
```diff
@@ -30,15 +30,11 @@
...
-	for(var/P in typesof(/obj/item/pda)-blocked)
-		var/obj/item/pda/D = new P
-
-		//D.name = "PDA Style [colorlist.len+1]" //Gotta set the name, otherwise it all comes up as "PDA"
-		D.name = D.icon_state //PDAs don't have unique names, but using the sprite names works.
-
-		src.colorlist += D
+	for(var/thing in typesof(/obj/item/pda) - blocked)
+		var/obj/item/pda/P = thing
+		colorlist[initial(P.icon_state)] = initial(P.desc)
```
AffectedArc07, line 37: This feels like it should be the other way round, or have P.icon_state as the key and value.

Fox-McCloud: This definitely appears as `pda_clown`, `pda_toxins`, and what have you, which is the same behavior to the player.

AffectedArc07: Ok thats fine

PR #13562, opened 2020-06-07
Fixes Vampire Vision

Fox-McCloud:
Fixes vamp vision not updating without taking off/putting on glasses.

:cl: Fox McCloud
fix: Fixes vampire vision not updating when you get the proper blood level
/:cl:


code/game/gamemodes/vampire/vampire.dm
```diff
@@ -227,6 +227,7 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 	if(istype(spell, /obj/effect/proc_holder/spell))
 		owner.mind.AddSpell(spell)
 	powers += spell
+	owner.update_sight()
```
Crazylemon64, line 230: somehow feel like this should be specifically for the /datum/vampire_passive/vision path

PR #13564, opened 2020-06-07
Dramatically Cuts Object Processing Down

Fox-McCloud:
Cuts down on the amount of things in the SSobj processing system by about ~3,000 items.

Dramatically short list to iterate over and wayyyyy less proc call overhead.

Basically every single instance of `/datum/reagents` in the world was added to processing just in case cryostylane or pyrosium comes into contact with oxygen and needs to cool the container it's in---or stable mutagen needs to copy blood data to itself.

It also was for welders to add reagents to themselves....heaven knows why this wasn't on the welder processing itself.

Either case, this means that only pyrosium/cryostylane/stable mutagen process when they actually exist in the world--and they also process at the speed I originally wanted/intended them to (once every 1 second instead of every 2).

:cl: Fox McCloud
tweak: dramatically improves object processing performance
/:cl:


code/modules/reagents/chemistry/reagents/pyrotechnic.dm
```diff
@@ -333,19 +333,26 @@
...
 		M.reagents.remove_reagent("oxygen", 1)
 		M.bodytemperature -= 30
 	return ..()
 
-/datum/reagent/cryostylane/on_tick()
-	if(holder.has_reagent("oxygen"))
-		holder.remove_reagent("oxygen", 2)
-		holder.remove_reagent("cryostylane", 2)
-		holder.temperature_reagents(holder.chem_temp - 200)
-		holder.temperature_reagents(holder.chem_temp - 200)
```
Fox-McCloud, line 347: These all did this twice because we didn't have the `processing` SS at the time, ergo processing only happened once every 2 seconds--so to compensate, I just called it twice. Having it be - 400 would not be the same as calling - 200 twice.

Issue #13541, opened 2020-06-04
Magic spells / items / summons forbiden to mages on raging mages can be bought


Qwertytoforty:
**Issue Description**:

Wizards can by anything in the spellbook on raging mages

**What did you expect to happen**: 

Apprenteces, lich, and the like to be RIGHT OUT

**What happened instead**:

Raging mages have apprentices and lich and everything is on fire

**Why is this bad/What are the consequences**:

Raging mages have restrictions to keep wizards living forever / making more apprentences, this breaks that

**Steps to reproduce the problem**:

Force the round as raging
Watch as as everything goes to hell as there is apprenteces and liches

**When did the problem start happening**:

06/04 first reported case

Shockpoint:
This is why I never force ragin'.

SteelSlayer:
I don't know if there was ever code to restrict buying these two things during ragin. Do you recall them being unable to buy in the past at any point?

PR #13576, opened 2020-06-08
Fixes wizards on ragin mages being able to buy restricted spells

### LINKED ISSUE #8278: Bind Soul can be taken on Raging Mages, even though it doesn't work.
**Problem Description**:
You still get gibbed when you die as a wizard on raging mages, despite having an ensouled item

To be able to revive using Bind Soul.

**What happened instead**:
Got gibbed on death, no chance to revive.

It's 2 wasted points on an antag where every point matters. No indication of raging mages for the first mage, so you have no idea not to take it/do it.

Die with bind soul active on raging mages. Become a pillar of salt.

SteelSlayer:
## What Does This PR Do
There are a few items that wizards on ragin' mages aren't allowed to buy but the code wasn't working properly, and this fixes that. 

The spells are:
- The Bind Soul (lich) spell
- The Summon Guns spell
- The Summon Magic spell
- The Summon Ghosts spell
- The Lich loudout (had to add this one in because it contains the lich spell)
- The Gunslinging Reaper loudout (had to add this one in because it contains the lich spell)

## Why It's Good For The Game
Fixes #8278
Fixes: #13541

You'll notice in the second bug report that it suggests apprentices should be restricted as well, but I couldn't fine evidence that it ever was. So that is more of a feature request.

## Changelog
:cl:
fix: Fixes wizards on ragin mages being able to buy restricted spells
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/spellbook.dm
+++ b/code/game/gamemodes/wizard/spellbook.dm
@@ -1,19 +1,15 @@
 /datum/spellbook_entry
 	var/name = "Entry Name"
-
+	var/is_ragin_restricted = FALSE // FALSE if this is buyable on ragin mages, TRUE if it's not.
 	var/spell_type = null
 	var/desc = ""
 	var/category = "Offensive"
 	var/log_name = "XX" //What it shows up as in logs
 	var/cost = 2
 	var/refundable = TRUE
-	var/surplus = -1 // -1 for infinite, not used by anything atm
 	var/obj/effect/proc_holder/spell/S = null //Since spellbooks can be used by only one person anyway we can track the actual spell
 	var/buy_word = "Learn"
 	var/limit //used to prevent a spellbook_entry from being bought more than X times with one wizard spellbook
-
-/datum/spellbook_entry/proc/IsSpellAvailable() // For config prefs / gamemode restrictions - these are round applied
-	return TRUE
 
 /datum/spellbook_entry/proc/CanBuy(mob/living/carbon/human/user, obj/item/spellbook/book) // Specific circumstances
 	if(book.uses < cost || limit == 0)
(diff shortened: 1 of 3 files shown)


datlo:
This also fixes https://github.com/ParadiseSS13/Paradise/issues/8278

SteelSlayer:
Thanks, I added it.

AzuleUtama:
I think that the gunslinging reaper loadout also has lich in it - so you'll want to blacklist that one too.

SteelSlayer:
@AzuleUtama Oh... you're right it does. Maybe in the future some alternative could be given, like giving back 2 points to choose another spell/item if you choose gunslinger on ragin, but for now I'll have to blacklist it.

SteelSlayer:
@farie82 Yeah I'll try to do something like that

code/game/gamemodes/wizard/spellbook.dm
```diff
@@ -642,11 +610,12 @@
 	var/entry_types = subtypesof(/datum/spellbook_entry) - /datum/spellbook_entry/item - /datum/spellbook_entry/summon - /datum/spellbook_entry/loadout
 	for(var/T in entry_types)
 		var/datum/spellbook_entry/E = new T
-		if(E.IsSpellAvailable())
-			entries |= E
-			categories |= E.category
-		else
+		if(GAMEMODE_IS_WIZARD && E.is_ragin_restricted)
 			qdel(E)
```
farie82, line 614: Can most likely use the same construct as:
https://github.com/ParadiseSS13/Paradise/blob/73e8781b0ad9c51a5593868e0f9330d38e7e47a5/code/game/machinery/PDApainter.dm#L35-L37
`thing` here is a type. This way you don't have to create a new `spellbook_entry` and remove it again if it doesn't fit

SteelSlayer: I tried doing 
```js
for(var/T in entry_types)
		var/datum/spellbook_entry/E = T
		if(GAMEMODE_IS_WIZARD && E.is_ragin_restricted)
			continue
		E = new T
        ...
```
But my spellbook ends up completely blank. I assume byond won't allow me to do the `E = new T` line.

PR #13577, opened 2020-06-08
Gives the Chef the CQC manual as a traitor item

trololiver112:
## What Does This PR Do
This PR allows the Chef to buy CQC from his uplink for 12 TC. This will also add the manual to the surplus, which should be fine considering you can already get it from a bundle.

## Why It's Good For The Game
CQC currently can only be gotten for nukies or from the traitor bundle. This would make it possible for a chef, who can already use CQC in the kitchen to bring it outside the kitchen. As well as traitors having a chance of getting it from surplus.

## Changelog
:cl:
tweak: Chefs can now buy the CQC manual from their traitor uplink. It can also now be gotten from the surplus crate.
/:cl:

### DIFF
--- a/code/modules/martial_arts/martial.dm
+++ b/code/modules/martial_arts/martial.dm
@@ -204,8 +204,6 @@
 	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
 
 	var/datum/martial_art/cqc/CQC = new(null)
-	var/datum/martial_art/cqc/under_siege/CCQC = new(null)
-	CCQC.remove(user)
 	CQC.teach(user)
 	user.drop_item()
 	visible_message("<span class='warning'>[src] beeps ominously, and a moment later it bursts up in flames.</span>")
--- a/code/modules/martial_arts/martial.dm
+++ b/code/modules/martial_arts/martial.dm
@@ -195,12 +195,17 @@
 	icon = 'icons/obj/library.dmi'
 	icon_state = "cqcmanual"
 
+/obj/item/CQC_manual/chef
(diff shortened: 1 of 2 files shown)


code/modules/martial_arts/martial.dm
```diff
@@ -195,12 +195,17 @@
...
+/obj/item/CQC_manual/chef
+	desc = "A small, black manual. Written on the back it says: Bringing the home advantage with you."
+
 /obj/item/CQC_manual/attack_self(mob/living/carbon/human/user)
 	if(!istype(user) || !user)
 		return
 	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
 
 	var/datum/martial_art/cqc/CQC = new(null)
+	var/datum/martial_art/cqc/under_siege/CCQC = new(null)
```
farie82, line 207: Making a new one here is a bit odd. Best to include the removing of the current martial arts thing in the `teach` proc which overrides the old one

trololiver112: After doing some testing it seems that removing the old CQC isn't even needed at all

farie82:
Can you check if the verbs of the user contain duplicate values? I'm expecting the help verb to be duped in that list.
If so it would be best to remove the current martial_arts in the new teach. Gotta be sure to make it also work for temp martial arts though

Fox-McCloud:
Please double check what @farie82 mentioned.

trololiver112:
As discussed with @farie82 on discord, i wont touch on changing the deleting of older martial arts o. This PR, as it is working just fine like this. I think the plan was that Farie was going to look over all that in general in the martial arts refactor PR he has. This here works and doesn’t break anything

PR #13578, opened 2020-06-08
Gives the arrivals airlocks a open and close sound. Also makes them autoclose after some time

trololiver112:
## What Does This PR Do
Since the arrivals airlcoks are subtypes directly to doors and not airlocks they do not have a open and close sound, and autoclose is FALSE by default.
This PR gives them the normal open and close airlock sounds, as well as sets autoclose to TRUE

## Why It's Good For The Game
This should have been in long time ago, and i am surprised we never noticed.

## Changelog
:cl:
fix: Makes the arrivals airlocks have sounds for opening and closing as well as automatically close.
/:cl:

### DIFF
--- a/code/game/machinery/doors/unpowered.dm
+++ b/code/game/machinery/doors/unpowered.dm
@@ -18,3 +18,6 @@
 /obj/machinery/door/unpowered/shuttle
 	icon = 'icons/turf/shuttle.dmi'
 	icon_state = "door1"
+	SdoorOpen = 'sound/machines/airlock_open.ogg'
+	SdoorClose = 'sound/machines/airlock_close.ogg'
+	autoclose = TRUE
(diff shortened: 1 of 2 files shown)


code/game/machinery/doors/unpowered.dm
```diff
@@ -18,3 +18,6 @@
 /obj/machinery/door/unpowered/shuttle
 	icon = 'icons/turf/shuttle.dmi'
 	icon_state = "door1"
+	SdoorOpen = 'sound/machines/airlock_open.ogg'
+	SdoorClose = 'sound/machines/airlock_close.ogg'
+	autoclose = TRUE
```
Fox-McCloud, line 23: This isn't the way to do it, this is literally a copy-pasted that's on a different subtype. It's pretty gross.

Either refactor or change the type of doors on the shuttle (not recommended).

trololiver112: couldn't i define the sounds for the airlocks all the way down at door and then just use it there instead?

trololiver112:
Will reconsider this one and maybe reopen at a later date.

PR #13583, opened 2020-06-08
Adds Swarming Component

Fox-McCloud:
Ports TG's Swarming component

Swarming is a component where multiple objects with the swarming component can occupy the same tile while giving them the appearance of being clustered together.

This is currently utilized for Legions, Bees, Spiderlings, and Viscerators. Necessary mob adjustments have been made so they can occupy the same tile.

Needless to say, this make all the above mobs (well spiderlings aren't mobs; it's mostly just aesthetics for that one) significantly more dangerous in groups as they can cluster up a lot easier and attack all at the same time.

visual of 20 bees swarming a mob:

img

:cl: Fox McCloud
add: Bees, legions, and viscerators can now all swarm, making them significantly more difficult to fight
add: spiderlings can swarm, though this is mostly aesthetic
/:cl:


taukausanake:
_THE BEEEEEEES!_ I love this just for the bees alone but the others seem nice as well. There might be a few other mobs that could use this but I don't think anyone wants swarms of pests infesting the station. A literal wave of rats or kittens (catsplosion anyone?) would be more amusing at least

farie82:
Balance change as well this one due to the first CL point.

Question about the picture. Where are they located tile wise? Some seem to be on the victims tile and some down right. If that's the case then it's fine. But if they are all located on the right tile then less so since it doesn't show this that well.

Fox-McCloud:
BYOND is going to handle them being on a different tile, depending on their pixel offset. Nothing we can really do about that, just how things work.

farie82:
> BYOND is going to handle them being on a different tile, depending on their pixel offset. Nothing we can really do about that, just how things work.

As long as the sprites are not misleading at where they are located it will be fine I'd say. Else people might get confused as to why they can't hit the bee or such.

AffectedArc07:
NO NOT THE BEES AAAAAAAAAAA

code/modules/mob/living/simple_animal/hostile/syndicate.dm
```diff
@@ -349,7 +349,8 @@
 	icon = 'icons/mob/critter.dmi'
 	icon_state = "viscerator_attack"
 	icon_living = "viscerator_attack"
-	pass_flags = PASSTABLE
+	pass_flags = PASSTABLE | PASSMOB
```
Crazylemon64, line 352: Is the idea that for this to work, swarming enemies should have `PASSMOB` set?

Crazylemon64:
@Fox-McCloud What do you mean by your statement that BYOND will handle them being on another tile? To my knowledge, for the example image in the OP, all of them will be handled as if occupying the tile to the right regardless of whatever `pixel_foo` value is had

Fox-McCloud:
I believe their `loc` will indeed be the turf in front of the mob, but as for how the player sees it, they will exist on a different tile.

It's like when you right click something that is orbiting; you have to click the tile that its own (right click or otherwise)--you can't just right click the tile/turf that their actual `loc` is to click on them.

code/modules/mob/living/simple_animal/hostile/syndicate.dm
Fox-McCloud: Well, for mobs, yes.

For everything else? Not really.

It's just mobs have this weird aspect where they can only occupy the same turf if one happens to be lying down, initially (or have passmob); otherwise they bump into each other and swap spots.

PR #13592, opened 2020-06-10
Refactors Simple Mob Respawning

Fox-McCloud:
Refactors simple mob respawnining to be more OOP instead of doing a snowflake, single proc, istype check.

I also tweaked the list a little of what you can play as.

additions:
- Can play as all dogs

Removals:
- Can't play as poly (but can play as regular parrots)
  - Too many people use poly as a means to metagame and talk to people when they really shouldn't
- Commander Clucky
- Betsy
  - Depriving the chef of the ability to get milk, eggs, or a reliable supply of meat just because you're upset you can't play anymore isn't cool

Also fixes Tom's `real_name` not being set.

:cl: Fox McCloud
del: can no longer respawn as Poly, Betsy, or Commander Clucky
add: can respawn as any type of dog (hellhounds are not dogs)
/:cl:


BakKeener:
I can understand Betsy and Commander Clucky as they are the chef only source of milk and eggs, but Poly? That'll shove him aside as chat filler instead of having a player being able to take the role and do something like RPing or possibly entertaining the crew as he'll just sit in the CE Office doing nothing except making a pile of small objects.

Carthusia:
Im also confused about the poly removal, can you go into more depth why this is the case?

Fox-McCloud:
Poly, unlike other animals, has `universal_speak` set to `TRUE`. 

As such, it's quite common for players to play him and just say--whatever they feel like.

With other mobs, it's mostly not a problem, unless the crew gets super involved or a borg is found--but with poly, they can just openly talk about whatever. 

Metagaming, in general, is a problem with simple mob respawning, but when you have access to a simple mob that has a radio headset and `universal_speak` enabled (which is needed for regular poly to be understood), well---it's a very obvious target for *that* kind of player.

Carthusia:
mmm, ok i get it a bit more now, what if his universal speak was set to false and he no longer started with a headset?

He can only at the start then speak to borgs, and I think vox, and isnt capable of blaring any possible meta info to the station from the get go en mass.

It keeps him playable, but puts him more in line with Ian and others, where they are simply animals, who with some work can be made to be understood, if work is done by an exterior force.

PidgeyThePirate:
I'd like it if someone playing poly had an ability to speak using lines spoken around them only (like having them in a big list). Attempting a normal Say just gives a squawk.

ZomgPonies:
@PidgeyThePirate That would be a thematically and balance appropriate compromise I believe.

code/modules/mob/mob.dm
```diff
@@ -1061,12 +1061,11 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
...
 		return
 
-	if((usr in GLOB.respawnable_list) && (stat==2 || istype(usr,/mob/dead/observer)))
+	if((usr in GLOB.respawnable_list) && (stat == DEAD || istype(usr,/mob/dead/observer)))
 		var/list/creatures = list("Mouse")
-		for(var/mob/living/L in GLOB.alive_mob_list)
-			if(safe_respawn(L.type) && L.stat!=2)
-				if(!L.key)
-					creatures += L
+		for(var/mob/living/simple_animal/L in GLOB.mob_living_list)
```
farie82, line 1066: Wouldn't the `alive_mob_list` be less populated than the `mob_living_list` and thus be a better fit? Since observers won't be included in here as well and `alive_mob_list` won't have dead mobs in them anyway

code/modules/mob/living/simple_animal/parrot.dm
```diff
@@ -705,6 +707,9 @@
 	available_channels = list(":e")
 	..()
 
+/mob/living/simple_animal/parrot/Poly/npc_safe(mob/user) // Hello yes, I have universal speak and I follow people around and shout out antags
```
Kyep, line 710: 👍

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -615,3 +615,6 @@
 	if(pcollar && collar_type)
 		add_overlay("[collar_type]collar")
 		add_overlay("[collar_type]tag")
+
+/mob/living/simple_animal/proc/npc_safe(mob/user)
+	return FALSE
```
Kyep, line 620: Glad this defaults to FALSE.

variableundefined (approved):
Will merge when deconflicted

code/game/gamemodes/miniantags/borer/borer.dm
```diff
@@ -838,6 +838,11 @@
 		to_chat(src, "Sugar nullifies your abilities, avoid it at all costs!")
 		to_chat(src, "You can speak to your fellow borers by prefixing your messages with ':bo'. Check out your Borer tab to see your abilities.")
 
+/mob/living/simple_animal/borer/npc_safe(mob/user)
+	if(!jobban_isbanned(user, ROLE_BORER) && !jobban_isbanned(user, "Syndicate"))
```
SabreML, line 842: Probably better to use the define here for consistency, LGTM otherwise aside from Farie's review.
```suggestion
	if(!jobban_isbanned(user, ROLE_BORER) && !jobban_isbanned(user, ROLE_SYNDICATE))
```

Spacemanspark:
>Metagaming, in general, is a problem with simple mob respawning, but when you have access to a simple mob that has a radio headset and universal_speak enabled (which is needed for regular poly to be understood), well---it's a very obvious target for that kind of player

This isn't difficult to circumvent for the other mobs, and it's something we've had players do before.

A simple /me can instantly bypass the say emote to those nearby and not blind.

It's also not difficult for a mob--especially Ian or E-N--to captivate someone's attention long enough to lead them towards metagame land.

We can handle Poly players that call out things over the radio via meta knowledge if they come up--it's seldom, if ever, an issue in practice.

PR #13594, opened 2020-06-10
Updates Components

Fox-McCloud:
Updates components to TG's latest standard.

This is mostly to keep us up to date and make it easier to maintain consistency with what they have---components aren't going away, and they're only going to be expanded upon even more.

Named arguments are now supported with components, instead of having to have every single argument filled or it won't behave itself.

This also does away with a few macros `GET_COMPONENT` and `GET_COMPONENT_FROM` as they're confusing and it's not always clear what it's doing; it's better to just have code readability that's consistent with the rest of the codebase---it didn't even save on proc call overhead.

This also ads the components subsystem and elements.

Elements are kinda like "light" components that have very small behaviors---waddling, for instance, is perfect for this as it's a very very simple behavior that doesn't need the complexity of components to utilize.

I also axed the jestosterone component, because it was awful from a component standpoint, when a simple `mind` check would suffice---the unique behavior of it, of giving you COMIC SANS, has been turned into a trait `TRAIT_JESTER`.

This also adds support for global signals. 

An example of this is the component "edit complainer", which makes the thing that has it talk whenever it's edited. There's a particularly grumpy plushie that very loudly lets you know when it's been var edited and properly grumbles at you for doing so.

Lastly, this updates the documentation and adds support for in-code documentation of components, should we ever add that.

No CL as this is all backend and completely invisible to the player.


code/controllers/subsystem/dcs.dm
```diff
@@ -0,0 +1,53 @@
...
+	var/datum/element/eletype = arguments[1]
+	var/list/fullid = list("[eletype]")
+	var/list/named_arguments = list()
+	for(var/i in initial(eletype.id_arg_index) to length(arguments))
+		var/key = arguments[i]
+		var/value
+		if(istext(key))
+			value = arguments[key]
+		if(!(istext(key) || isnum(key)))
+			key = "\ref[key]"
```
Fox-McCloud, line 41: @ninjanomnom This is the only concern I have and the only difference between TG and ourselves in this PR.

We don't have `REF` and we've generally relied on `UID()` instead (ie: `thing.UID()`)---do you believe this will break anything or should this be fine?

code/controllers/subsystem/dcs.dm
```diff
@@ -0,0 +1,53 @@
...
+		key = "[key]" // Key is stringified so numbers dont break things
+		if(!isnull(value))
+			if(!(istext(value) || isnum(value)))
+				value = "\ref[value]"
```
Fox-McCloud, line 45: Also here.

code/game/objects/items/toys.dm
```diff
@@ -1023,6 +1023,18 @@ obj/item/toy/cards/deck/syndicate/black
 	name = "orange fox plushie"
 	icon_state = "orangefox"
 
+/obj/item/toy/plushie/orange_fox/grump
+	name = "grumpy fox"
+	desc = "An ancient plushie that seems particularly grumpy."
```
Crazylemon64, line 1028: :eyes:

code/game/objects/items/toys.dm
```diff
@@ -1023,6 +1023,18 @@ obj/item/toy/cards/deck/syndicate/black
...
+
+/obj/item/toy/plushie/orange_fox/grump/ComponentInitialize()
+	. = ..()
+	var/static/list/grumps = list("Ahh, yes, you're so clever, var editing that.", "Really?", "If you make a runtime with var edits, it's your own damn fault.",
+	"Don't you dare post issues on the git when you don't even know how this works.", "Was that necessary?", "Ohhh, setting admin edited var must be your faovrite pastime!",
```
Kyep, line 1033: faovrite

PR #13597, opened 2020-06-10
Brigtimer Runtime fix

ZomgPonies:
## What Does This PR Do
Fixed a rare runtime when people have their ID in their PDA and the PDA in the ID slot. Also adds additionnal checks of ways you can inteact with the brigtimers while having your ID be valid (in-hand, in PDA/PDAslot, in PDA/IDslot)

Note this runtime was in the original brig timer code and dates from before the NanoUI upgrade.

## Why It's Good For The Game
Runtimes bad

## Changelog
:cl:
fix: brigtimer runtime fix
/:cl:


code/game/machinery/doors/brigdoors.dm
```diff
@@ -95,8 +95,21 @@
 		var/mob/living/carbon/human/M = usr
 		var/rank = "UNKNOWN RANK"
 		if(istype(M) && M.wear_id)
```
Fox-McCloud, line 97: use `GetID` please.

code/game/machinery/doors/brigdoors.dm
```diff
@@ -95,8 +95,21 @@
...
+			if(istype(M.wear_id,/obj/item/card/id))
+				var/obj/item/card/id/I = M.wear_id
+				rank = I.assignment
+			else if(istype(M.wear_id,/obj/item/pda))
+				var/obj/item/pda/P = M.wear_id
+				if(P.id)
+					var/obj/item/card/id/I = P.id
+					rank = I.assignment
+		else if(istype(M) && (M.l_hand || M.r_hand))
+			if(istype(M.l_hand,/obj/item/card/id))
```
Fox-McCloud, line 107: same

code/game/machinery/doors/brigdoors.dm
ZomgPonies: Fixed

code/game/machinery/doors/brigdoors.dm
ZomgPonies: Fixed

PR #13601, opened 2020-06-11
Cyborg factory tweaks/improvements.

SteelSlayer:
## What Does This PR Do
This improves the cyborg transformer/factory that malf AI's can buy. Also refactors/improves some of the `/transformer` code.

#### Whats changed?
Currently the borg factory is awful, and there's not much reason to use it.

- First of all, borgs created get synced up to whichever AI has the fewest borgs, which most of the time, is *not* the malf AI. This is bad and goes against the point of the factory. I'd almost consider this a bug fix but I listed it as a tweak anyway. Now they **always** get synced to the malf AI who placed the factory.

- Secondly, the borgs only get a **5,000** capacity cell. This is also pretty terrible, as the borgs can't really do much of anything before having to recharge. And remember, cyborgs can't change out each other's batteries. So I've bumped that up to a **15,000** charge cell (high capacity plus).

- The borg factory can convert players which are not laying down. This means that you don't have to stun a player before they will get sent into the factory. They will simply be pushed by the conveyor belt and converted.

- The borg factory can now convert dead players. Currently, even if someone is dead, as long as their brain is not completely maxed out on damage, they can be borged normally. So I'm just applying that logic to the borg factory.

- If an person gets converted to a borg, but is either SSD or catatonic (or can't re-enter their body for whatever reason), the game will poll ghosts on whether or not they'd like to play a malfunctioning borg, and additionally let admins know.

I think this will make the borg factory a viable option for malf AIs to choose again.
## Changelog
:cl:
tweak: Cyborgs created from the robotic factory that was placed by a Malf AI will now connect to that AI, instead of whichever AI has the least amount of cyborgs
tweak: Cyborgs created from the robotic factory now start with a 15,000 capacity cell.
tweak: Robotic factories can now convert dead humans, and humans who are not laying down.
tweak: Players converted by the robotic factory who are SSD or otherwise cannot return to their body, will trigger a ghost poll in order to get a replacement player for the borg.
/:cl:


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -130,7 +130,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 	radio = new /obj/item/radio/borg(src)
 	common_radio = radio
 
-	init()
+	init(alien, connect_to_AI)
```
SteelSlayer, line 133: I hate having to pass around this argument so much but I have to due to this stupid `init` proc being here.

code/modules/mob/transform_procs.dm
```diff
@@ -44,26 +44,37 @@
...
+	* cell_type: A type path of the cell the new borg should receive.
+	* connect_to_default_AI: TRUE if you want /robot/New() to handle connecting the borg to the AI with the least borgs.
+	* AI: A reference to the AI we want to connect to.
+*/
+/mob/living/carbon/human/proc/Robotize(cell_type = null, connect_to_default_AI = TRUE, mob/living/silicon/ai/AI = null)
 	if(notransform)
 		return
 	for(var/obj/item/W in src)
 		unEquip(W)
-	regenerate_icons()
```
SteelSlayer, line 53: This is pointless because the human is going to be immediately transferred into the robot anyway

code/modules/mob/transform_procs.dm
```diff
@@ -44,26 +44,37 @@
...
+
 	notransform = 1
 	canmove = 0
 	icon = null
 	invisibility = 101
-	for(var/t in bodyparts)
-		qdel(t)
-	for(var/i in internal_organs)
-		qdel(i)
```
SteelSlayer, line 61: Deletion of these is handled in `/carbon/Destroy` or `/human/Destroy` already

BakKeener:
Can the dead players who are shoved in who don't want to be cyborgs have their body given up? Because not everyone wants to be a cyborg. Also what does this do about bodies that don't have ckeys attached to them anymore aka DC'D Players or players that have taken over a new body?

SteelSlayer:
If you don’t want to play that role you’ll have to Ahelp, similar to other conversion antags. 

And you make a good point about the SSD thing, I’ll have to add some code to bring back players who are ghosting. And if they’re completely disconnected then offer them up to ghosts.

BakKeener:
Another thing to add, Make sure that humanized monkeys can't be used for this and the such. Only bodies that HAD ckeys attached to them should be usable otherwise it'll just refuse the body as incompatible. So that ghosts don't get a prompt if a body without a ckey goes in.

Spacemanspark:
I think he would know to do that :p

code/game/machinery/transformer.dm
```diff
@@ -6,25 +6,51 @@
...
+	/// Cooldown between each transformation, in deciseconds.
+	var/cooldown_duration = 600
+	/// If the factory is currently on cooldown from its last transformation.
+	var/is_on_cooldown = FALSE
+	/// The type of cell that newly created borgs get.
+	var/robot_cell_type = /obj/item/stock_parts/cell/high/plus
+	/// The direction that mobs must moving in to get transformed.
 	var/acceptdir = EAST
+	/// The AI who placed this factory.
+	var/mob/living/silicon/ai/owner_AI
```
Fox-McCloud, line 22: `var/mob/living/silicon/ai/masterAI`

code/game/machinery/transformer.dm
```diff
@@ -35,163 +61,104 @@
...
-			do_transform(AM)
+	var/mob/living/carbon/human/H = AM
+	var/move_dir = get_dir(loc, H.loc)
 
-/obj/machinery/transformer/proc/do_transform(var/mob/living/carbon/human/H)
-	if(stat & (BROKEN|NOPOWER))
-		return
-	if(cooldown == 1)
+	if((transform_standing || H.lying) && move_dir == acceptdir)
+		H.forceMove(loc)
```
Fox-McCloud, line 78: `drop_location()`

Fox-McCloud:
I'd look into how TG handles this, though you mostly get it right, IMO.

I'm a bit skeptical of a few things though. Free borgs is very strong---so the increase in cell capacity and the ability to borg dead individuals is kinda iffy, in my opinion; the tweaks to who to sync to are fine---and I'm not sure if the standing or lying thing is really that big of a deal.....

How will this handle things like monkeys? Will it poll ghosts as well?

SteelSlayer:
@Fox-McCloud Which part specifically from TG do you mean?

The dead convert thing is probably the most controversial thing and something I could stand to remove. I mostly did it to cover anyone who rage suicides whenever they see they are about to be borged, but that might be an ahelp situation.

About the monkeys: This bit will prevent anything other than humans from getting transformed. Bumped is the only thing that calls the transform proc.
```js
/obj/machinery/transformer/Bumped(atom/movable/AM)
    // They have to be human to be transformed.
    if(is_on_cooldown || !ishuman(AM))
        return
```

code/game/machinery/transformer.dm
```diff
@@ -6,25 +6,51 @@
...
-	var/transform_standing = 0
-	var/cooldown_duration = 600 // 1 minute
-	var/cooldown = 0
-	var/robot_cell_charge = 5000
+	/// TRUE if the factory can transform dead mobs.
+	var/transform_dead = TRUE
+	/// TRUE if the mob can be standing and still be transformed.
+	var/transform_standing = TRUE
+	/// Cooldown between each transformation, in deciseconds.
+	var/cooldown_duration = 600
```
farie82, line 14: can use `1 MINUTE` here

code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -592,7 +592,8 @@
 		active = FALSE
 		return
 	var/turf/T = get_turf(owner_AI.eyeobj)
-	new /obj/machinery/transformer/conveyor(T)
+	var/obj/machinery/transformer/TR = new(T)
+	TR.masterAI = owner_AI
```
farie82, line 596: Can maybe be added to the `New` of the transformer. Bonus points if you refactor the `New` to `Initialize`

code/game/machinery/transformer.dm
```diff
@@ -35,163 +61,104 @@
...
+	H.emote("scream")
+	if(!masterAI) // If the factory was placed via admin spawning or other means, it wont have an owner_AI.
+		H.Robotize(robot_cell_type)
+		return
 
+	var/mob/living/silicon/robot/R = H.Robotize(robot_cell_type, FALSE, masterAI)
+	if(R.mind && !R.client && !R.grab_ghost()) // Make sure this is an actual player first and not just a humanized monkey or something.
+		message_admins("[key_name_admin(R)] was just transformed by a borg factory, but they were SSD. Polling ghosts for a replacement.")
+		var/list/candidates = pollCandidates("Do you want to play as a malfunctioning cyborg?", ROLE_TRAITOR, poll_time = 15 SECONDS)
+		if(!LAZYLEN(candidates))
```
farie82, line 108: just `length` here will do. `LAZYLEN` will now only be used for lists that are initialised lazily.

SteelSlayer:
@farie82 Well, the subtypes do have similar behavior. They push something through a machine using conveyors and then do something to the object/mob. I'd rather not worry about these, only because they're so obscure and so rarely used in a round. The most used one is probably the outfit equipper, which I've seen admins use ocassionally, but other than that, they're unused I think.

code/game/machinery/transformer.dm
SteelSlayer: Yep.

code/game/gamemodes/malfunction/Malf_Modules.dm
SteelSlayer: Yeah that will be cleaner, I'll change that

code/game/machinery/transformer.dm
SteelSlayer: Can do.

BakKeener:
Could make it so the factory can only eat bodies if they are defibbable so dead players being converted into borgs aren't abused.

code/game/machinery/transformer.dm
SteelSlayer: I forgot you can't do this in variable declarations:

Code_M0mr6TDlAh

And the compiler line says: `code\game\machinery\transformer.dm:14:error: MINUTE: expected end of statement`

farie82: Oh yeah it's `MINUTES` instead of MINUTE. Should compile with that

SteelSlayer: Oh... right, it's plural. Works now :smile:

AffectedArc07:
One the conflicts are resolved, this can be merged.

SteelSlayer:
Conflicts resolved.

PR #13602, opened 2020-06-12
Gas Mixture Refactor

Fox-McCloud:
Ideally, we need listmos, from TG.

Either case, that's probably not going to happen any time soon....for now, I just did this.

Currently, gas mixtures are a datum that holds the common gases in var form----then has traces gases as a list of gas datums that nave no useful info on them bar their `specific_heat`. 

Moving towards datumizing all gases would be the ideal, but with the current system that's not really possible and would just lead to an even more confusing mess of code and a performance loss.

Either way, this means less looping through lists on gas datums and just manipulation of vars, so this *should* perform better.

Additionally, this means that we now have support for sleeping gas and others on unsimulated turfs so....yay I guess?

This also increases the readability of code and means far less more gas datums are copied when they're just a trace amount of N2O or something of that nature---I'm guessing this will also make N2O more responsive for actually moving around and showing up on a turf tile.

This also replaces a few `.len` checks in atmos with `length`, which is faster, by a hair---also utilizes caching of some lists to and mathematical functions to make things perform slightly better.

Need to profile this, but I've tested it--things seem to generally behave as they should.


code/LINDA/LINDA_turf_tile.dm
```diff
@@ -271,8 +280,7 @@
 	if(air.toxins > MOLES_PLASMA_VISIBLE)
 		return "plasma"
 
-	var/datum/gas/sleeping_agent = locate(/datum/gas/sleeping_agent) in air.trace_gases
```
Fox-McCloud, line 274: The real reason for this whole refactor. I'm not even kidding.

AffectedArc07:
Noticed that you have modified Box & Meta. Wouldnt this issue also exist on Delta?

code/LINDA/LINDA_turf_tile.dm
```diff
@@ -358,32 +366,23 @@
...
-				var/datum/gas/sleeping_agent/G = new
-				G.moles = S.moles/turf_list.len
-				T.air.trace_gases += G
+		T.air.oxygen			= A.oxygen / turf_list.len
+		T.air.carbon_dioxide	= A.carbon_dioxide / turf_list.len
+		T.air.nitrogen			= A.nitrogen / turf_list.len
+		T.air.toxins			= A.toxins / turf_list.len
+		T.air.sleeping_agent	= A.sleeping_agent / turf_list.len
+		T.air.agent_b			= A.agent_b / turf_list.len
+		T.air.volatile_fuel		= A.volatile_fuel / turf_list.len
```
Fox-McCloud, line 385: I could probably just cache `length(turf_list)` and re-use it for faster nonsense.

code/game/machinery/atmoalter/canister.dm
```diff
@@ -533,29 +532,11 @@ update_flag
...
-//Dirty way to fill room with gas. However it is a bit easier to do than creating some floor/engine/n2o -rastaf0
-/obj/machinery/portable_atmospherics/canister/sleeping_agent/roomfiller/New()
-	..()
-	var/datum/gas/sleeping_agent/trace_gas = air_contents.trace_gases[1]
-	trace_gas.moles = 9*4000
-	spawn(100)
-		var/turf/simulated/location = src.loc
-		if(istype(src.loc))
-			while(!location.air)
-				sleep(1000)
```
Fox-McCloud, line 553: What the everloving heck.

code/game/turfs/simulated/floor/plating.dm
```diff
@@ -208,10 +208,8 @@
 /turf/simulated/floor/engine/n20/New()
 	..()
 	var/datum/gas_mixture/adding = new
-	var/datum/gas/sleeping_agent/trace_gas = new
 
-	trace_gas.moles = 6000
-	adding.trace_gases += trace_gas
+	adding.sleeping_agent = 6000
```
Fox-McCloud, line 212: Just define this on the turf instead of this BS.

code/datums/gas_mixture.dm
```diff
@@ -362,56 +323,62 @@ What are the archived variables for?
 
 /datum/gas_mixture/check_turf(turf/model, atmos_adjacent_turfs = 4)
 	var/delta_oxygen = (oxygen_archived - model.oxygen)/(atmos_adjacent_turfs+1)
```
Fox-McCloud, line 325: derp

code/datums/gas_mixture.dm
```diff
@@ -362,56 +323,62 @@ What are the archived variables for?
...
 		return 0
 
-	for(var/gas in trace_gases)
-		var/datum/gas/trace_gas = gas
-		if(trace_gas.moles_archived > MINIMUM_AIR_TO_SUSPEND*4)
-			return 0
-
 	return 1
 
 /datum/gas_mixture/proc/check_turf_total(turf/model) //I want this proc to die a painful death
```
Fox-McCloud, line 348: All these modelling turfs disgust me.

Unsimulated turfs, reeee.

code/datums/gas_mixture.dm
```diff
@@ -460,83 +448,44 @@ What are the archived variables for?
 	toxins -= delta_toxins
 	sharer.toxins += delta_toxins
 
-	var/moved_moles = (delta_oxygen + delta_carbon_dioxide + delta_nitrogen + delta_toxins)
-	last_share = abs(delta_oxygen) + abs(delta_carbon_dioxide) + abs(delta_nitrogen) + abs(delta_toxins)
```
Fox-McCloud, line 464: `last_share = abs(delta_oxygen) + abs(delta_carbon_dioxide) + abs(delta_nitrogen) + abs(delta_toxins)` this needs to stay.

code/game/machinery/atmoalter/canister.dm
KasparoVy: Deep wizard code right here

code/datums/gas_mixture.dm
```diff
@@ -608,75 +560,62 @@ What are the archived variables for?
 		var/sharer_heat_capacity = sharer.heat_capacity_archived()
 
 		if((sharer_heat_capacity > MINIMUM_HEAT_CAPACITY) && (self_heat_capacity > MINIMUM_HEAT_CAPACITY))
-			var/heat = conduction_coefficient*delta_temperature* \
-				(self_heat_capacity*sharer_heat_capacity/(self_heat_capacity+sharer_heat_capacity))
+			var/heat = conduction_coefficient*delta_temperature * \
+				(self_heat_capacity * sharer_heat_capacity / (self_heat_capacity+sharer_heat_capacity))
```
Fox-McCloud, line 564: boop

code/datums/gas_mixture.dm
```diff
@@ -608,75 +560,62 @@ What are the archived variables for?
...
-		((nitrogen < (1-MINIMUM_AIR_RATIO_TO_SUSPEND)*sample.nitrogen) || (nitrogen > (1+MINIMUM_AIR_RATIO_TO_SUSPEND)*sample.nitrogen)))
+	if((abs(nitrogen - sample.nitrogen) > MINIMUM_AIR_TO_SUSPEND) && \
+		((nitrogen < (1 - MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.nitrogen) || (nitrogen > (1 + MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.nitrogen)))
 		return 0
-	if((abs(carbon_dioxide-sample.carbon_dioxide) > MINIMUM_AIR_TO_SUSPEND) && \
-		((carbon_dioxide < (1-MINIMUM_AIR_RATIO_TO_SUSPEND)*sample.carbon_dioxide) || (carbon_dioxide > (1+MINIMUM_AIR_RATIO_TO_SUSPEND)*sample.carbon_dioxide)))
+	if((abs(carbon_dioxide - sample.carbon_dioxide) > MINIMUM_AIR_TO_SUSPEND) && \
+		((carbon_dioxide < (1 - MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.carbon_dioxide) || (carbon_dioxide > (1 + MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.carbon_dioxide)))
 		return 0
 	if((abs(toxins-sample.toxins) > MINIMUM_AIR_TO_SUSPEND) && \
```
Fox-McCloud, line 602: boop

code/datums/gas_mixture.dm
```diff
@@ -608,75 +560,62 @@ What are the archived variables for?
...
-		((toxins < (1-MINIMUM_AIR_RATIO_TO_SUSPEND)*sample.toxins) || (toxins > (1+MINIMUM_AIR_RATIO_TO_SUSPEND)*sample.toxins)))
+		((toxins < (1 - MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.toxins) || (toxins > (1 + MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.toxins)))
+		return 0
+	if((abs(sleeping_agent - sample.sleeping_agent) > MINIMUM_AIR_TO_SUSPEND) && \
+		((sleeping_agent < (1 - MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.sleeping_agent) || (sleeping_agent > (1 + MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.sleeping_agent)))
+		return 0
+	if((abs(agent_b - sample.agent_b) > MINIMUM_AIR_TO_SUSPEND) && \
+		((agent_b < (1 - MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.agent_b) || (agent_b > (1 + MINIMUM_AIR_RATIO_TO_SUSPEND) * sample.agent_b)))
+		return 0
+	if((abs(toxins - sample.volatile_fuel) > MINIMUM_AIR_TO_SUSPEND) && \
```
Fox-McCloud, line 611: `volatile_fuel` not `toxins`

Fox-McCloud:
>Wouldnt this issue also exist on Delta?

Nope; it doesn't use the `roomfiller` type.

The canister there is purely decorative.

code/game/machinery/atmoalter/canister.dm
AffectedArc07: Someone probably messed up JS & DM

code/ATMOSPHERICS/components/unary_devices/vent_scrubber.dm
```diff
@@ -250,14 +250,13 @@
...
-					else if(istype(trace_gas, /datum/gas/sleeping_agent) && scrub_N2O)
-						removed.trace_gases -= trace_gas
-						filtered_out.trace_gases += trace_gas
+			if(removed.agent_b)
+				filtered_out.agent_b = removed.agent_b
+				removed.agent_b = 0
+
+			if(scrub_N2O)
+				filtered_out.sleeping_agent = removed.sleeping_agent
+				removed.sleeping_agent = 0
```
Crazylemon64, line 259: Is `volatile_fuel` intentionally omitted here? (I generally see a pattern of its omission throughout the code here that doesn't appear to be corrected)

code/LINDA/LINDA_turf_tile.dm
```diff
@@ -211,7 +220,14 @@
 
 	if(planet_atmos) //share our air with the "atmosphere" "above" the turf
 		var/datum/gas_mixture/G = new
-		G.copy_from_turf(src)
```
Crazylemon64, line 214: Why not use `copy_from_turf` here?

code/ATMOSPHERICS/components/unary_devices/vent_scrubber.dm
Fox-McCloud: I just put it in where it already existed, but yeah, the thought did occur to me. 

Any other locations that'd be ideal?

code/LINDA/LINDA_turf_tile.dm
Fox-McCloud: so, here's `copy_from_turf`:

```
/datum/gas_mixture/copy_from_turf(turf/model)
	oxygen = model.oxygen
	carbon_dioxide = model.carbon_dioxide
	nitrogen = model.nitrogen
	toxins = model.toxins

	//acounts for changes in temperature
	var/turf/model_parent = model.parent_type
	if(model.temperature != initial(model.temperature) || model.temperature != initial(model_parent.temperature))
		temperature = model.temperature

	return 1
```

The problem with this is the temperature checks. the vars for gases on a turf are *NEVER* altered under any circumstances; they're used for turf initialization with air, so we have an easy, non string way to handle what gases a turf has by default.

Temperature is a double whammy---it' used simultaneously for initialization temperature...anddd it's involved in atmos superconduction and is modified at runtime.

Having this proc run ultimately means that if the turf winds up with a temperature of something like 1,000,000 or something, it'll copy that to the air above it.

code/ATMOSPHERICS/components/unary_devices/vent_scrubber.dm
Fox-McCloud: I went ahead and just purged volatile fuel.

After doing some code digging, it was originally a gas that was created by splashing welding fuel on a turf--it was, predictably, horrifically overpowered.

It hasn't been used, and probably never will be---Agent B is *barely* used, but has an interesting enough reaction that I'm willing to keep it in---unless you want me to remove that as well.

Fox-McCloud:
I went ahead and purged volatile fuel--it's not used.

Also added full support for agent b---you can now create it with the appropriate proc, utilize it in build mode, and it filters out with plasma using filters (one filter, curiously, properly filtered it, the other didn't).

PR #13604, opened 2020-06-12
Replaces two malf AI verbs with action buttons, gets rid of the malfunction verb tab

SteelSlayer:
## What Does This PR Do
Finally gets rid of the awful "malfunction" verb tab for malf AIs and converts the two verbs which relied on it existing, into action buttons instead. 

The two abilities being:
- The "Choose Modules" verb, aka the ability shop for the AI to pick all their abilities like machine overload, doomsday, etc.
- The "Return to Main Core" verb, which was used for an APC-shunted AI to return to their actual core.

## Why It's Good For The Game
AIs already have enough tabs to worry about, they don't need another one just for two verbs.

## Images of changes
image

image

## Changelog
:cl:
add: Added two new action buttons for malf AIs, the "choose module" button, and the "return to main core" button. These replace their respective verbs from the malfunction tab.
/:cl:


ZomgPonies:
I have a special hate for most of the verb tabs. I would totally see an action button for example for changing your suit sensor level.

AffectedArc07:
@ZomgPonies just righclick your jumpsuit in the slot
image

Rurik123:
Is it possible to keep the verb tab too? It may sound like heresy but the option of both is nice, like vampire. Shame changeling doesn't follow suit in that like vampire.

Fox-McCloud:
Might want to look into how TG handles this; they have this all tied into a few procs to make things really smooth and easy to add/remove.

SteelSlayer:
Sure if you want. Our stuff is pretty similar already, just have to add a few lines.

PR #13605, opened 2020-06-13
Correcting Welder Messages on the Field Generator.

Moku-Mok:
## What Does This PR Do
This simply swaps WELD and SLICE messages that were previously incorrect to instead reflect as to what they should actually do.
Hopefully this can clear up some confusion for players, aswell as making it just look cleaner.

## Why It's Good For The Game
I believe that having the game display the proper messages would be a good change in any case.

## Changelog
:cl:
fix: Replaced SLICE and WELD Calls with each other in field_generator.dm in order to correctly reflect         what happens to the player that is interacting with the object.
/:cl:

### DIFF
--- a/code/modules/power/singularity/field_generator.dm
+++ b/code/modules/power/singularity/field_generator.dm
@@ -113,15 +113,15 @@ field_generator power level display
 	if(!I.tool_use_check(user, 0))
 		return
 	if(state == FG_SECURED)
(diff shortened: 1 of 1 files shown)


Moku-Mok:
I wish to apologize if i have done anything wrong with this, as i am still very new to github.

PR #13608, opened 2020-06-13
Removes references to console screens and removes an unnecessary var from circuitboards

SteelSlayer:
## What Does This PR Do
Since console screens do not exist anymore, they shouldn't be referenced anywhere in the game.
This removes the variable `frame_desc` from circuitboards because it's not even used anywhere, *and* it contained reference to the console screens.

## Changelog
:cl:
spellcheck: Examine text for circuit boards that require more than 1 of a part will now have an "s" on the end of the name of that part. For example "2 capacitor" becomes "2 capacitors"
/:cl:

### DIFF
--- a/code/game/machinery/constructable_frame.dm
+++ b/code/game/machinery/constructable_frame.dm
@@ -226,7 +226,6 @@ to destroy them and players will be able to make replacements.
 	name = "circuit board (Booze-O-Mat Vendor)"
 	board_type = "machine"
 	origin_tech = "programming=1"
-	frame_desc = "Requires 1 Resupply Canister."
 	build_path = /obj/machinery/vending/boozeomat
 	req_components = list(/obj/item/vending_refill/boozeomat = 1)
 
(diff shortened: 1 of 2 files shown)


code/game/machinery/constructable_frame.dm
```diff
@@ -881,7 +894,7 @@ to destroy them and players will be able to make replacements.
 	build_path = /obj/machinery/atmospherics/unary/cryo_cell
 	board_type = "machine"
 	origin_tech = "programming=4;biotech=3;engineering=4;plasmatech=3"
-	frame_desc = "Requires 1 Matter Bin, 1 piece of cable and 4 Console Screens."
+	frame_desc = "Requires 1 Matter Bin, 1 piece of cable and 4 Sheets of glasss."
```
Fox-McCloud, line 897: `glasss`?

Fox-McCloud:
I'll be honest, I dunno why these aren't just automagically generated.

code/game/machinery/constructable_frame.dm
SteelSlayer: I.. dont know how that happened :smile:

SteelSlayer:
That would probably be a good idea.

SteelSlayer:
I just realized we DO have these automatically generated in `/obj/item/circuitboard/examine`. And actually, `frame_desc` isn't even used so I've just removed the var completely. The only thing the examine text doesn't look good on is cables but, because its says "cable coils" instead of "pieces of cable", but not a huge deal I think

image

Fox-McCloud:
Woooo, removal of unecessary vars!

PR #13614, opened 2020-06-14
Removes the X-ray implant from RND

trololiver112:
## What Does This PR Do
This PR removes the X-ray implant from RND completely, making it an antag/admin/genetics only thing.

## Why It's Good For The Game
X-ray is one of the most powerful tools you can have, and not something you should be able to distribute to lots of crew with ease. Considering even antags have a hard time getting this, for example vampires needing full power, and nukies having to pay for it. I don't think the crew should get it basically for free. Especially considering most antags don't have any way to get it at all.

## Changelog
:cl:
tweak: Made it so that you can no longer make the X-ray implant in RND, thus making it an antag/admin/genetics only thing.
/:cl:

PR #13615, opened 2020-06-14
change tech requirement for xray implant

TDSSS:
## What Does This PR Do
This is intended as an alternative to both #13614 and #13463, though it could also be merged alongside the cost change if desired.

The tech requirement for making an x-ray implant is raised to biotech 8. This means it is not obtainable in a normal round without midround helping out. Namely, you need to deconstruct one of the following items to reach biotech 8:

- another xray implant, gotten from sol traders or surgically removing it from a nuke ops or deathsquad member without them blowing up (good luck)
- An abductor baton
- An abductor gland
- A slaughter demon heart
- An abductor recall implant
- A changeling egg (as in, let a headslut implant someone then surgically remove it. Good luck)

## Why It's Good For The Game
This change effectively locks x-ray implants behind the rewards for mid/lateround events. This means both that there will be many rounds where it isn't gotten at all and that in the rounds that do have it, the implants only come into play in the lategame, once antagonists have had plenty of time to prepare and arm up.

It also effectively rewards the crew for actually going out of their way to reach very high tech levels when possible. Quite often even scientists don't care at all about a sol trader selling fleshy masses or research data.

While I see the argument for removing xray completely, I think leaving it as a very rare reward for when science gets access to special rewards is more exciting.

## Changelog
:cl:
tweak: x-ray vision implants now need higher research to create.
/:cl:

### DIFF
--- a/code/modules/research/designs/medical_designs.dm
+++ b/code/modules/research/designs/medical_designs.dm
@@ -371,7 +371,7 @@
 	name = "X-Ray implant"
 	desc = "These cybernetic eyes will give you X-ray vision. Blinking is futile."
 	id = "ci-xray"
-	req_tech = list("materials" = 7, "programming" = 5, "biotech" = 7, "magnets" = 5,"plasmatech" = 6)
+	req_tech = list("materials" = 7, "programming" = 5, "biotech" = 8, "magnets" = 5,"plasmatech" = 6)
 	build_type = PROTOLATHE | MECHFAB
 	construction_time = 60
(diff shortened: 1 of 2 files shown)


Fox-McCloud:
I like this change, better, personally. 

It makes it supers super rare, but something to still hypothetically hope for.

It'll be rarer than even phazons.

Mitchs98:
I'd prefer this myself. Gives people something to work towards on the rare occasions that it'd be possible to do so.

Carthusia:
If it's between removal and this, this is the far more preferable option.

Not impossible to obtain, just improbable

Denthamos:
Hmn, after thinking it over yeah.  This would fix the issue of it being used way too often while giving that rare slim chance of making it available for use.  I am for this.

KasparoVy:
Def the best option

PR #13616, opened 2020-06-14
Fix HUDs taking parent atom's color

dearmochi:
## What Does This PR Do
Apply `RESET_COLOR` to `appearance_flags` of all HUDs so that they don't take their parent's colour

## Why It's Good For The Game
Makes viewing HUDs easier (or not outright impossible depending on the colour)

## Images of changes
Before:
image
After:
image

## Changelog
:cl:
fix: Fix HUDs being rendered with colour of the object they belong to
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -42,6 +42,7 @@
 				hud_list[hud] = list()
 			else
 				var/image/I = image('icons/mob/hud.dmi', src, "")
+				I.appearance_flags = RESET_COLOR | RESET_ALPHA
 				hud_list[hud] = I
 
 /mob/proc/generate_name()
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -42,7 +42,7 @@
 				hud_list[hud] = list()
 			else
 				var/image/I = image('icons/mob/hud.dmi', src, "")
-				I.appearance_flags = RESET_COLOR
+				I.appearance_flags = RESET_COLOR | RESET_TRANSFORM
 				hud_list[hud] = I
 
 /mob/proc/generate_name()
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -42,6 +42,7 @@
 				hud_list[hud] = list()
 			else
 				var/image/I = image('icons/mob/hud.dmi', src, "")
+				I.appearance_flags = RESET_COLOR | RESET_ALPHA
```
farie82, line 45: Won't this also make invisible people visible on the HUDs? Like with the genetics chameleon power?

Fox-McCloud: Yeah, this resetting alpha isn't ok.

It not inheriting color is fine.

dearmochi: Fixed

code/modules/mob/mob.dm
```diff
@@ -42,6 +42,7 @@
 				hud_list[hud] = list()
 			else
 				var/image/I = image('icons/mob/hud.dmi', src, "")
+				I.appearance_flags = RESET_COLOR
```
Fox-McCloud, line 45: Might as well make this `I.appearance_flags = RESET_COLOR|RESET_TRANSFORM` to head off another problem before it becomes one.

dearmochi: Wouldn't it make sense to not reset the transform if the mob is offset, though? e.g. lying in bed, etc.

Fox-McCloud: 99% of the time, players read it from left to right---someone lying on the ground suddenly makes it read vertically--which is a bit strange---this info is typicalllyy something you want  presented to the player in the same way.

What benefit is there to the HUD looking upside down and backwards just because the mob's transform is set that way? Doesn't make much sense.

dearmochi: Fair point, I'll add it in

KasparoVy:
Bloody love image manipulation and appearance flags! So much more functionality in them now

PR #13621, opened 2020-06-14
Make HUD settings apply consistently

### LINKED ISSUE #12943: Intent indicators are discoloured when using a custom HUD skin
**Issue Description**:
When using a custom skin colour in preferences, the intent selector gets discoloured and looks bad.
K6oXE5EI2U

This to be in the excluded icons list, like body targeting
image

**What happened instead**:
It looked awful

Something in the game could look a lot better

Set a custom HUD colour and look at the intent selector

Probably since custom UIs were added

**Extra information**:
This can probably be fixed by adding this HUD element into the color exclusion list

dearmochi:
## What Does This PR Do
Fixes #12943
- HUD alpha, color and style settings should now apply to all HUD elements consistently
- Cleans up some code when possible
- Fixes intent selector using the HUD color setting when it shouldn't

## Why It's Good For The Game
It makes HUD settings actually work instead of for 50% of the elements on the screen, making it more aesthetically pleasing for the player

## Images of changes
image
_HUD with Alpha: 200, Color: lightblue, Style: White_

image
_The HUD settings do not interfere with the custom action button backgrounds_

## Changelog
:cl:
fix: Fix HUD settings (alpha, color and style) not applying to most of the HUD elements (actions, gear, etc.)
fix: Fix intent selector taking HUD color setting when it shouldn't, causing it to look hard to read
/:cl:


Fox-McCloud:
That big red cross is ugly as hell.

It being greyed out is more than good enough.

BakKeener:
I'm with Fox on that, the big red crosses aren't good looking, I thought Greyed out was enough, plus isn't here a text message telling you it doesn't work when it's unavailable when you click it.

Mitchs98:
Echoing on the big red cross. Otherwise it looks nice~

dearmochi:
Oof, after sleeping on it I realized it also looks terrible, so I removed the red cross and added a black overlay:
image

code/datums/action.dm
```diff
@@ -126,25 +136,28 @@
...
-		return 0
+		return FALSE
 	if(target && attack_self)
 		var/obj/item/I = target
 		I.ui_action_click(owner, type)
-	return 1
+	return TRUE
 
 /datum/action/item_action/ApplyIcon(obj/screen/movable/action_button/current_button)
+	current_button.overlays.Cut()
```
Fox-McCloud, line 146: We don't do this anymore.

`cut_overlays`

dearmochi: Found about 170 instances of that, so I replaced them all by `cut_overlays`

Fox-McCloud:
Erm, I wouldn't do that in this PR---that'd be better reserved for a PR that exclusively does that. It makes it unceessarily difficult for us to review such things--and it's really beyond the scope of the PR in question. I meant for *That* specific use case and anything related to it.

dearmochi:
> Erm, I wouldn't do that in this PR---that'd be better reserved for a PR that exclusively does that. It makes it unceessarily difficult for us to review such things--and it's really beyond the scope of the PR in question. I meant for That specific use case and anything related to it.

Okay, sure, I'll undo that. However I still don't see the point of using the proc you suggested. Is it for readability? Is it a *just in case* like `LAZYLEN` (which you want removed)?

Fox-McCloud:
the overlay manipulation system is to give overlays specific priority so that certain overlays with specific priorities can be sorted properly or skipped over--having things outside of it is going to mess up that system.

It's not just a redefine of an existing proc; it's entire its own thing.

When it was added, it wasn't plugged in universally (unfortunately)--but we've been fixing it as we go along/refactor code.

code/_onclick/hud/action_button.dm
```diff
@@ -99,14 +99,20 @@
 
 /obj/screen/movable/action_button/hide_toggle/proc/InitialiseIcon(mob/living/user)
 	if(isalien(user))
+		icon = 'icons/mob/actions/actions.dmi'
 		icon_state = "bg_alien"
 	else
-		icon_state = "bg_default"
+		icon = user.client ? ui_style2icon(user.client.prefs.UI_style) : initial(icon)
```
farie82, line 105: Making these two ternaries into an `if/else` would improve readability I think.

code/_onclick/hud/human.dm
```diff
@@ -395,9 +402,9 @@
 	for(var/obj/screen/inventory/inv in (static_inventory + toggleable_inventory))
 		if(inv.slot_id)
 			if(inv.slot_id in S.no_equip)
-				inv.alpha = 128
+				inv.alpha = hud_alpha * 0.5
```
farie82, line 405: `/ 2` can be done here. Divisions and multiplications have mostly the same performance on byond. In some cases divisions are faster than multiplications.
Rule of thumb is to use which one reads the easiest

code/datums/action.dm
```diff
@@ -126,25 +136,28 @@
...
-		return 0
+		return FALSE
 	if(target && attack_self)
 		var/obj/item/I = target
 		I.ui_action_click(owner, type)
-	return 1
+	return TRUE
 
 /datum/action/item_action/ApplyIcon(obj/screen/movable/action_button/current_button)
+	current_button.cut_overlays()
```
farie82, line 146: Looking at line 156. It is not even needed to have this here. It'll be cut already there and on line 110 the parent cuts it if you were to go into the else.
That or line 156 can be gone and this statement can be put in the `if(use_itemicon)` scope

dearmochi:
@Kyep 

I think there was a misunderstanding on what this PR does. It simply makes the common HUD elements (inventory, hands, modes) use the player's [Custom UI settings](https://user-images.githubusercontent.com/1496804/87282590-b8d12500-c4f4-11ea-9563-1f6c2c90161a.png) in the Game Preferences consistently such that ALL common elements use the same theme rather than only some of them. The player's role(s) has otherwise no bearing on the common elements.

Vampire: 
image

Changeling:
image

**As a reference, here how the HUD currently looks on master:**

image

Three glaring issues:

1. The intent selector is colored for some reason, making it hard to tell which intent you're actually on.
2. The body part selector *isn't* coloured when it should be.
3. Neither are the **common** action buttons in the top-left corner of the screen.

Issues 2 and 3 result in visual inconsistencies with the HUD. This PR fixes all 3 points

dearmochi:
image

How it looks now

AffectedArc07 (approved):

PR #13623, opened 2020-06-15
Fixes Shadowling darksight not giving night vision when toggled

### LINKED ISSUE #13596: Shadowlings do not get their night vision when turning into a shadowling.
**Issue Description**:

When shadowlings turn into a shadowling from their human form, they do not get their night vision without updating their vision in another way

Night vision to turn on and work when the sling hatches

**What happened instead**:

slings have to put on sunglasses or a welding helmet or shadow step to get it to work

Shouldn't have to mess with your vision like that to turn on night vision

Become a shadowling
Hatch
Notice it is dark and you can not see

Life rework

trololiver112:
## What Does This PR Do

Fixes #13596

This fixes the darksight ability for shadowlings, and once again gives them night vision when toggling it.

## Why It's Good For The Game

Fix Fix, prevents users from having to update their visions in other ways just to use their god damn abilities

## Changelog
:cl:
fix: Fixed the shadowling darksight not toggling night vision without something else updating your vision.
/:cl:

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -184,7 +184,7 @@
 		var/mob/living/carbon/human/H = target
 		if(!H.vision_type)
 			to_chat(H, "<span class='notice'>You shift the nerves in your eyes, allowing you to see in the dark.</span>")
-			H.set_sight(new /datum/vision_override/nightvision)
+			H.set_sight(/datum/vision_override/nightvision)
 		else
 			to_chat(H, "<span class='notice'>You return your vision to normal.</span>")
 			H.set_sight(null)
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -187,7 +187,7 @@
 			H.set_sight(new /datum/vision_override/nightvision)
 		else
 			to_chat(H, "<span class='notice'>You return your vision to normal.</span>")
-			H.set_sight()
+			H.set_sight(null)
 
 /obj/effect/proc_holder/spell/targeted/shadow_vision/thrall
 	desc = "Thrall Darksight"
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -186,9 +186,12 @@
 		if(!H.vision_type)
 			to_chat(H, "<span class='notice'>You shift the nerves in your eyes, allowing you to see in the dark.</span>")
 			H.vision_type = new vision_path
+			H.update_sight()
+
 		else
 			to_chat(H, "<span class='notice'>You return your vision to normal.</span>")
 			H.vision_type = null
+			H.update_sight()
 
 /obj/effect/proc_holder/spell/targeted/shadow_vision/thrall
 	desc = "Thrall Darksight"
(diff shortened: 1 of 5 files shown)


trololiver112:
Both shadowlings and Shadows now has this and it is also a proc on the mob level called set_sight

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -185,12 +184,10 @@
...
 		if(!H.vision_type)
 			to_chat(H, "<span class='notice'>You shift the nerves in your eyes, allowing you to see in the dark.</span>")
-			H.vision_type = new vision_path
-			H.update_sight()
+			H.set_sight(new /datum/vision_override/nightvision)
 		else
 			to_chat(H, "<span class='notice'>You return your vision to normal.</span>")
-			H.vision_type = null
-			H.update_sight()
+			H.set_sight()
```
farie82, line 190: Will be best to add `null` here as param. The effect is the same but this way it makes it clear to the reader that you mean to set it to `null`.

code/modules/mob/living/carbon/human/species/shadow.dm
```diff
@@ -39,10 +39,10 @@
 /datum/action/innate/shadow/darkvision/Activate()
 	var/mob/living/carbon/human/H = owner
 	if(!H.vision_type)
-		H.vision_type = new /datum/vision_override/nightvision
+		H.set_sight(new /datum/vision_override/nightvision)
 		to_chat(H, "<span class='notice'>You adjust your vision to pierce the darkness.</span>")
 	else
-		H.vision_type = null
+		H.set_sight()
```
farie82, line 45: Same here

code/modules/mob/mob.dm
```diff
@@ -1356,6 +1356,10 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 	SEND_SIGNAL(src, COMSIG_MOB_UPDATE_SIGHT)
 	sync_lighting_plane_alpha()
 
+/mob/proc/set_sight(datum/vision_override/O)
+	vision_type = O
```
farie82, line 1360: Can you check if you can `qdel` the old `vision_type`?
You will have to make sure that every time vision_type is set that a new type is being used there. And not a variable that is saved on the item or such. Else you remove the items vision_type.

code/modules/mob/mob.dm
```diff
@@ -1356,6 +1356,12 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 	SEND_SIGNAL(src, COMSIG_MOB_UPDATE_SIGHT)
 	sync_lighting_plane_alpha()
 
+/mob/proc/set_sight(datum/vision_override/O)
```
Fox-McCloud, line 1359:
```

/mob/proc/set_sight(datum/vision_override/O)
	QDEL_NULL(vision_type)
	if(O)
		vision_type = O
	update_sight()
```

farie82: Wanted to get Fred to figure this one out himself but this works :P
Wouldn't it just be `set_sight(new_vision_type)` where `new_vision_type` is the type of the object?

Fox-McCloud: You could do that too, but having the args like that makes it explicitly clear what you put there.

Fox-McCloud:
Please be sure to test this, with the changes. Make sure there are not runtimes or anything and that it functions as intended.

trololiver112:
yup, tested both with shadow people and shadowlings, works as intended and 0 runtimes

PR #13624, opened 2020-06-15
minor fix

TDSSS:
## What Does This PR Do
Deletes some spaces that got in there at some point

## Why It's Good For The Game
I don't _think_ this is a functional change? The foamblade can still be spawned in and so on ingame, but it definitely shouldn't be like that. It also makes the extension not report a problem anymore, so that's good too.

## Changelog
:cl:
fix: minor formatting fix in toy foamblade
/:cl:


AffectedArc07:
I saw these when the langserver was screaming
image
Can you also fix the others?

PR #13626, opened 2020-06-15
Purges Laziness

Fox-McCloud:
There's no point to `LAZYLEN` IMO ...and upon thinking on it, I don't personally think there ever will be; having a define of an existing BYOND proc is a bit silly--especially one who's functionality is stable and unlikely to ever change.


Fox-McCloud:
Bleh. Mixed feelings.

PR #13628, opened 2020-06-16
Ports Orbiting Component

Fox-McCloud:
Ports TG's orbiting component.

Fixes the issue with weird offsets in 513.

I've done the basic testing with this, and it appears to work fine; will need to test more, but everything *should* be fine.

Also, ghosts now float up and down--spooookyyy.

And you can no longer orbit turfs--only movable atoms.

also fixes Wisps failing to GC under certain circumstances, then getting deleted, anyway, after a couple of minutes.

And cleans up the `following` holder that was used with the old ghost following system.

:cl: Fox McCloud
fix: Fixes weird offsets with orbiting in 513
del: can no longer orbit turfs
add: ghosts how float up and down
/:cl:


code/datums/components/orbiter.dm
```diff
@@ -0,0 +1,139 @@
+/datum/component/orbiter
```
Fox-McCloud, line 1: I'll be the first to admit, I'm not really a big fan of this component---it's turns a lightweight proc into something far more complicated.

That said, it does mean we can do things like `transfer_observers_to`, which does help cut down on thing, I suppose.

@Crazylemon64 would really like your opinion on all of this.

code/datums/components/orbiter.dm
```diff
@@ -0,0 +1,139 @@
...
+	to_chat(orbiter, "<span class='notice'>Now orbiting [parent].</span>")
+
+/datum/component/orbiter/proc/end_orbit(atom/movable/orbiter, refreshing=FALSE)
+	if(!orbiters[orbiter])
+		return
+	UnregisterSignal(orbiter, COMSIG_MOVABLE_MOVED)
+	SEND_SIGNAL(parent, COMSIG_ATOM_ORBIT_STOP, orbiter)
+	orbiter.SpinAnimation(0, 0)
+	if(istype(orbiters[orbiter],/matrix)) //This is ugly.
+		orbiter.transform = orbiters[orbiter]
```
Fox-McCloud, line 91: the above two lines are the ugly hack that fixes the weird offset with orbiting/transform.

Fox-McCloud:
Don't have time to fix all the bugs with this, right now---and I really need to get moving on the supermatter anyway---closing this, but leaving it on my branch.

PR #13630, opened 2020-06-16
Minor refactor to wireless eguns

AffectedArc07:
## What Does This PR Do
Wireless eguns now use the `COMSIG_MOVABLE_Z_CHANGED` signal to decide if they are in an away mission or not, which means they arent checking for a location every time they process. Also cuts down on call overhead for recharging

## Why It's Good For The Game
Code should be good

## Changelog
:cl: AffectedArc07
tweak: Optimised the code of wireless eguns
/:cl:

### DIFF
--- a/code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
+++ b/code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
@@ -191,25 +191,27 @@
 	origin_tech = null
 	selfcharge = 1
 	can_charge = 0
-	var/inawaymission = 1
+	var/inawaymission = TRUE
 
-/obj/item/gun/energy/laser/awaymission_aeg/process()
+/obj/item/gun/energy/laser/awaymission_aeg/Initialize(mapload)
+	RegisterSignal(src, list(COMSIG_MOVABLE_Z_CHANGED), .proc/updateZ)
+	// Force update it incase it spawns outside an away mission and shouldnt be charged
+	updateZ()
+	return ..()
+
+/obj/item/gun/energy/laser/awaymission_aeg/proc/updateZ()
 	var/turf/my_loc = get_turf(src)
 	if(is_away_level(my_loc.z))
-		if(inawaymission)
-			return ..()
 		if(ismob(loc))
 			to_chat(loc, "<span class='notice'>Your [src] activates, starting to draw power from a nearby wireless power source.</span>")
-		inawaymission = 1
+		inawaymission = TRUE
 	else
 		if(inawaymission)
 			if(ismob(loc))
 				to_chat(loc, "<span class='danger'>Your [src] deactivates, as it is out of range from its power source.</span>")
 			cell.charge = 0
-			inawaymission = 0
+			inawaymission = FALSE
 			update_icon()
-
-
 
 /obj/item/reagent_containers/glass/beaker/terror_black_toxin
 	name = "beaker 'Black Terror Venom'"
--- a/code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
+++ b/code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
@@ -191,31 +191,25 @@
 	origin_tech = null
 	selfcharge = 1
 	can_charge = 0
-	var/inawaymission = TRUE
+	// Selfcharge is enabled and disabled, and used as the away mission tracker
+	selfcharge = TRUE
 
 /obj/item/gun/energy/laser/awaymission_aeg/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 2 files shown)


code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
```diff
@@ -191,26 +191,28 @@
...
 	can_charge = 0
-	var/inawaymission = 1
+	var/inawaymission = TRUE
 
-/obj/item/gun/energy/laser/awaymission_aeg/process()
+/obj/item/gun/energy/laser/awaymission_aeg/Initialize(mapload)
+	RegisterSignal(src, list(COMSIG_MOVABLE_Z_CHANGED), .proc/updateZ)
+	// Force update it incase it spawns outside an away mission and shouldnt be charged
+	updateZ()
+	return ..()
```
farie82, line 200: Shouldn't this happen before the above, the `..()`? Usually Initiations are handled base -> self. And destroys self -> base due to how dependencies can work. This rule can be broken if you need to override something that gets used in the init of a base.
It should be fine here though but just for consistence.

AffectedArc07:
Tested, that version works as well, so I will do that for consistency

code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
```diff
@@ -191,26 +191,28 @@
 	origin_tech = null
 	selfcharge = 1
 	can_charge = 0
-	var/inawaymission = 1
+	var/inawaymission = TRUE
 
-/obj/item/gun/energy/laser/awaymission_aeg/process()
+/obj/item/gun/energy/laser/awaymission_aeg/Initialize(mapload)
+	. = ..()
+	RegisterSignal(src, list(COMSIG_MOVABLE_Z_CHANGED), .proc/updateZ)
```
SteelSlayer, line 198: I'm curious as to why the signal is in a list here. It probably works fine but, just wondering.

AffectedArc07: Other examples in the code work like this, and I was concerned if something would break if I didn't do it like this.

code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
```diff
@@ -191,26 +191,33 @@
...
+	// Force update it incase it spawns outside an away mission and shouldnt be charged
+	updateZ()
 
 /obj/item/gun/energy/laser/awaymission_aeg/process()
+	// Only call the self-recharge proc if inside an away mission
+	if(inawaymission)
+		..()
+
+/obj/item/gun/energy/laser/awaymission_aeg/proc/updateZ()
```
Fox-McCloud, line 207: Now that I think about it.....why not just do 

`/obj/item/gun/energy/laser/awaymission_aeg/onTransitZ(old_z, new_z)`

and have all the code below run.

Then you're not having to register signals or send signals or anything.

code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
Fox-McCloud: If it's just a single signal, you can just put the signal in.

code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
AffectedArc07: `onTransitZ` is now used to cut down on the overhead, even though its minimal.

code/modules/awaymissions/mission_code/UO71-terrorspiders.dm
```diff
@@ -191,26 +191,27 @@
...
 
-/obj/item/gun/energy/laser/awaymission_aeg/process()
-	var/turf/my_loc = get_turf(src)
-	if(is_away_level(my_loc.z))
-		if(inawaymission)
-			return ..()
+/obj/item/gun/energy/laser/awaymission_aeg/Initialize(mapload)
+	. = ..()
+	// Force update it incase it spawns outside an away mission and shouldnt be charged
+	onTransitZ(new_z=loc.z)
```
Fox-McCloud, line 200: spacing.

AffectedArc07: I thought args for procs didnt have spaces. Huh.

Fox-McCloud: `onTransitZ(new_z = loc.z)` will work.

AffectedArc07: I meant in our code style. I will change it now I can switch branches.

PR #13631, opened 2020-06-16
Recharger tweaks and fixes

dearmochi:
## What Does This PR Do
- **Recharger efficiency now applies to all item types, not just guns**
- Recharger icon now goes to the charging state right away when adding an item in rather than next cycle
- Fixes recharger icon not updating on power loss/gain
- Breaks down recharger code into procs and clean up where possible for readability

## Why It's Good For The Game
- Makes using rechargers a more consistent experience in regards to readability at a glance
- Minor buff to charging items that aren't guns if the recharger is upgraded

## Changelog
:cl:
tweak: Recharger efficiency now applies to all item types, not just guns
fix: Recharger icon now uses proper icon instantly when an item is put in
fix: Fix recharging icon not updating on power loss/gain
/:cl:


code/game/machinery/recharger.dm
```diff
@@ -3,19 +3,23 @@
...
-	var/using_power = FALSE
+
 	var/list/allowed_devices = list(/obj/item/gun/energy, /obj/item/melee/baton, /obj/item/modular_computer, /obj/item/rcs, /obj/item/bodyanalyzer)
 	var/icon_state_off = "rechargeroff"
 	var/icon_state_charged = "recharger2"
 	var/icon_state_charging = "recharger1"
 	var/icon_state_idle = "recharger0"
 	var/recharge_coeff = 1
+	var/power_usage_gun = 250
+	var/power_usage_misc = 200
```
SteelSlayer, line 19: I think these would be better as defines, because they're static numbers that are never going to change.

code/game/machinery/recharger.dm
```diff
@@ -179,6 +145,60 @@
 		return
 	icon_state = icon_state_idle
 
+/obj/machinery/recharger/proc/get_cell_from(obj/item/I)
+	if(!I)
+		return null
```
SteelSlayer, line 150: Do you think the if check is necessary here? Because in `try_recharging_if_possible()` you already check if `charging` exists before the call to this proc, which then checks if `I`, aka `charging`, exists a second time.

code/game/machinery/recharger.dm
dearmochi: Yeah, I made them vars with the idea in mind that they could be edited later for a more efficient recharger or something but in the grand scheme of things it's probably very insignificant. I'll make them defines

code/game/machinery/recharger.dm
dearmochi: Well, I'd consider that an utility function in the context of the recharger and I personally think utils should be self-contained in terms of checks rather than assuming their input is sane, so I'd like to keep that in

code/game/machinery/recharger.dm
```diff
@@ -179,6 +146,57 @@
...
+		var/obj/item/rcs/R = I
+		return R.rcell
+
+	if(istype(I, /obj/item/bodyanalyzer))
+		var/obj/item/bodyanalyzer/B = I
+		return B.cell
+
+	return null
+
+/obj/machinery/recharger/proc/check_cell_needs_recharging(obj/item/stock_parts/cell/C)
```
farie82, line 174: I'd just keep this as a check in the proc itself. So no proc call will be needed. Unless there is a need to override this check then it's valid. But currently it's not required

code/game/machinery/recharger.dm
```diff
@@ -179,6 +146,57 @@
...
+	return C && C.charge < C.maxcharge
+
+/obj/machinery/recharger/proc/recharge_cell(obj/item/stock_parts/cell/C, power_usage)
+	C.give(C.chargerate * recharge_coeff)
+	use_power(power_usage)
+
+/obj/machinery/recharger/proc/try_recharging_if_possible(just_check = FALSE) // just_check means no actual recharging takes place, useful for cosmetic purposes
```
farie82, line 181: I'd split this up in two procs. One that checks if it has to charge and another to actually charge. Which makes use of the prior proc.
This way you ditch the awkward param

code/game/machinery/recharger.dm
```diff
@@ -1,22 +1,27 @@
+#define RECHARGER_POWER_USAGE_GUN 250
+#define RECHARGER_POWER_USAGE_MISC 200
```
AffectedArc07, line 2: Please `#undef` these at the end of the file

dearmochi: done

PR #13633, opened 2020-06-16
Fixes being able to 'stack' repairs on mechs

AzuleUtama:
## What Does This PR Do
This implements some additional lines to the 'welder_act' code for mechs to prevent people from being able to "stack" welding repairs to almost fully heal a mech in about 2 seconds. Anyone who tries to weld a mech when it's already being repaired by one will now get a warning message and need to wait for the current cycle to finish:

image

## Why It's Good For The Game
Prevents mechs from being nigh-invulnerable against someone standing next to it with a welder. 

## Changelog
:cl:
fix: Can no longer stack welder repairs on mechas
/:cl:

### DIFF
--- a/code/game/mecha/combat/gygax.dm
+++ b/code/game/mecha/combat/gygax.dm
@@ -38,8 +38,9 @@
 	name = "Dark Gygax"
 	icon_state = "darkgygax"
 	initial_icon = "darkgygax"
+	max_integrity = 300
 	deflect_chance = 20
-	armor = list(melee = 30, bullet = 30, laser = 30, energy = 35, bomb = 20, bio = 0, rad =20, fire = 100, acid = 100)
+	armor = list(melee = 40, bullet = 40, laser = 50, energy = 35, bomb = 20, bio = 0, rad =20, fire = 100, acid = 100)
 	max_temperature = 35000
 	leg_overload_coeff = 100
 	operation_req_access = list(ACCESS_SYNDICATE)
--- a/code/game/mecha/combat/gygax.dm
+++ b/code/game/mecha/combat/gygax.dm
@@ -38,9 +38,8 @@
 	name = "Dark Gygax"
 	icon_state = "darkgygax"
 	initial_icon = "darkgygax"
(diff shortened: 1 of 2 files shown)


Denthamos:
This really should be two different PR's.  The fix is needed while the nerf to a nuclear mode only mech is something that is controversial and needs to be discussed.

code/game/mecha/mecha.dm
```diff
@@ -836,7 +837,11 @@
 	if((obj_integrity >= max_integrity) && !internal_damage)
 		to_chat(user, "<span class='notice'>[src] is at full integrity!</span>")
 		return
+	if(repairing)
+		to_chat(user, "<span class='notice'>[src] is currently being repaired!</span>")
+		return
```
AffectedArc07, line 842: This is not the proper way to do this. If someone walks mid repair this will likely prevent the mech from being repaired ever.

code/game/mecha/combat/gygax.dm
```diff
@@ -38,9 +38,8 @@
 	name = "Dark Gygax"
 	icon_state = "darkgygax"
 	initial_icon = "darkgygax"
-	max_integrity = 300
 	deflect_chance = 20
-	armor = list(melee = 40, bullet = 40, laser = 50, energy = 35, bomb = 20, bio = 0, rad =20, fire = 100, acid = 100)
```
AffectedArc07, line 43: This belongs in another PR. Dont try and combine a fix and a nerf into one PR.

Spacemanspark:
I don't mind the nerf necessarily, but as others have said it really does belong in a separate PR.

AzuleUtama:
Following initial feedback, I've reverted the nerfs to the gygax and instead just kept the welder repair changes to this PR.

code/game/mecha/mecha.dm
farie82: Should be fine. If they do that then line 854 will still be hit. `I.use_tool` uses a `do_after` internally

PR #13637, opened 2020-06-18
Fixes the non stopping projectile bug. Also removes the unused legacy bullet code

farie82:
## What Does This PR Do
Stops the projectile from looping when they are `QDELETED` instead of when their loc becomes null.
The bullet is most likely not dereferenced properly or his location is changed after it gets qdelled.

Also removes the unused legacy bullet code.

I've tested this locally to see if it worked. All guns seemed to work. I've not tested them all.
I've tried to replicate the bug by removing the chambered bullet using VV and it worked. The shot got removed once it was shot as expected. Prior this caused the same effect as the bug

## Why It's Good For The Game
In some rare cases the projectile will be "deleted" but will still have a loc and thus continue on shooting shit till the GC decides it's time for it to die.

fixes:
```
[2020-06-18T05:48:52] Runtime in unsorted.dm,2014: The weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened weakened kinetic force taking damage after deletion
[2020-06-18T05:48:52]   proc name: stack trace (/datum/proc/stack_trace)
[2020-06-18T05:48:52]   src: the weakened weakened weakened... (/obj/item/projectile/kinetic)
[2020-06-18T05:48:52]   src.loc: the volcanic floor (32,199,5) (/turf/simulated/floor/plating/asteroid/basalt/lava_land_surface)
[2020-06-18T05:48:52]   call stack:
[2020-06-18T05:48:52]   the weakened weakened weakened... (/obj/item/projectile/kinetic): stack trace("The weakened weakened weakened...")
[2020-06-18T05:48:52]   the weakened weakened weakened... (/obj/item/projectile/kinetic): take damage(239, "brute", "bomb", 0, null, 0)
[2020-06-18T05:48:52]   the weakened weakened weakened... (/obj/item/projectile/kinetic): ex act(2)
[2020-06-18T05:48:52]   explosion(the volcanic floor (32,200,5) (/turf/simulated/floor/plating/asteroid/basalt/lava_land_surface), 1, 2, 5, null, 0, 0, 0, 0, 1, null, 1)
[2020-06-18T05:48:52]   (This error will now be silenced for 10 minutes)
```

## Changelog
:cl:
fix: Projectiles no longer have the infinite living bug where they won't get removed once they hit something
/:cl:


code/modules/projectiles/projectile.dm
```diff
@@ -232,74 +231,61 @@
...
-							Bump(original, 1)
-				Range()
-			sleep(1)
+
+	while(!QDELETED(src))
+		if(!paused)
+			if((!( current ) || loc == current))
+				current = locate(Clamp(x+xo,1,world.maxx),Clamp(y+yo,1,world.maxy),z)
+			if(isnull(Angle))
+				Angle=round(Get_Angle(src,current))
```
SteelSlayer, line 243: While you're touching all this, think you could fix those jank parentheses on line 240, and add spacing around the operations at line 241, 243, and anywhere else below this that needs it?

code/modules/projectiles/projectile.dm
```diff
@@ -232,74 +231,61 @@
 	return 1 //Bullets don't drift in space
 
 /obj/item/projectile/proc/fire(var/setAngle)
+	set waitfor = 0
```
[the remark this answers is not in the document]
farie82: Yes got it

code/modules/projectiles/projectile.dm
farie82: Got it danke for the review!

code/modules/projectiles/projectile.dm
```diff
@@ -232,74 +231,61 @@
...
-					current = locate(Clamp(x+xo,1,world.maxx),Clamp(y+yo,1,world.maxy),z)
-				step_towards(src, current)
-				if(original && (original.layer>=2.75) || ismob(original))
-					if(loc == get_turf(original))
-						if(!(original in permutated))
-							Bump(original, 1)
-				Range()
-			sleep(1)
+
+	while(!QDELETED(src))
```
Fox-McCloud, line 238: Please point out any other changes made.

code/modules/projectiles/projectile.dm
```diff
@@ -26,7 +26,6 @@
 	var/speed = 1			//Amount of deciseconds it takes for projectile to travel
 	var/Angle = null
 	var/spread = 0			//amount (in degrees) of projectile spread
-	var/legacy = FALSE			//legacy projectile system
```
farie82, line 29: Removed this var and it's uses

code/modules/projectiles/projectile.dm
```diff
@@ -232,74 +231,61 @@
...
-				else
-					animate(src, pixel_x = pixel_x_offset, pixel_y = pixel_y_offset, time = max(1, (speed <= 3 ? speed - 1 : speed)))
-
-				if(original && (original.layer>=2.75) || ismob(original))
-					if(loc == get_turf(original))
-						if(!(original in permutated))
-							Bump(original, 1)
-				Range()
-			sleep(max(1, speed))
-	else //old projectile system
```
farie82, line 290: Removed the else here (`if(legacy)`)

code/modules/projectiles/projectile.dm
farie82: I've put comments on the parts I changed. The last commit has some minor refactoring done as well which was mostly spacing. I'll comment the logic "changes" that I've done in that refactor

code/modules/projectiles/projectile.dm
```diff
@@ -232,74 +231,61 @@
...
+
+			speed = round(speed)
+			step_towards(src, locate(new_x, new_y, z))
+			if(speed <= 1)
+				pixel_x = pixel_x_offset
+				pixel_y = pixel_y_offset
+			else
+				animate(src, pixel_x = pixel_x_offset, pixel_y = pixel_y_offset, time = max(1, (speed <= 3 ? speed - 1 : speed)))
+
+			if(ismob(original) || (original && original.layer >= 2.75))
```
farie82, line 283: was `if(original && (original.layer>=2.75) || ismob(original))`
changed it since mobs usually fire the bullets so it's a minor optimization. Also made the cases more clear since the parentheses didn't make sense

Fox-McCloud: This is actually more expensive. 2.75 is a layer that most mobs are at, therefore this will return `TRUE` and you won't be running the `istype` check with mobs.

Checking for a holder and its var value is going to short circuit wayyy more often than actually checking for `ismob`.

Also, this isn't the one doing the firing, this is the original target *clicked*.

farie82: Ah that makes more sense. Will change it back to that order then

PR #13639, opened 2020-06-18
Ports Cloth Golems

Fox-McCloud:
Ports Cloth Golems, from TG.

Cloth golems are a new type of ...well, golem, that can be made by adding cloth or gauze to a construct shell.

They have significantly less armor than regular golems (They only take 15% less damage) and are incredibly weak to burn damage---they also have no resistance to heat, either, like regular golems---additionally, they're always flammable.

In exchange they're a bit faster than regular golems....andddd---they revive on non-gibbing death.

If you die, you collapse into a pile of cloth---after 90 seconds you'll fully regenerate. If you're on fire when you die, however, you'll just turn into dust. The pile of cloth that is left behind can be hit with something that's hot and it will set it on fire--once it burns, it'll prevent the golem from reviving.

img

Also increases the max amount of gauze you can stack together to 12 (starting value is still 6)

:cl: Fox McCloud
add: Adds cloth golems
tweak: increases the max amount of gauze you can stack together from 6 to 12
/:cl:


fludd12:
I like golems, but I can understand most of your points in opposition (even though cloth is my favorite subtype). That being said...
> they seem a bit ridiculous. I mean... cloth? Automatically reviving on death? They seem like straight up magic, more something the wizard federation would use, than actual NT technology.

I'm not sure this point has much ground to stand on, given they're __golems__ made of magic slime metal which spontaneously develop sentience in seconds 😅

variableundefined (approved):

PR #13645, opened 2020-06-19
Refactor dmm_suite for readability

dearmochi:
## What Does This PR Do

Cleans up arguments, comments and paths of `/dmm_suite` (now `/datum/dmm_suite`) to increase readability should other developers want to take a look at it. Any suggestions to make it even more readable are appreciated!


code/modules/awaymissions/maploader/writer.dm
```diff
@@ -5,28 +5,27 @@
...
-			"P","Q","R","S","T",
-			"U","V","W","X","Y",
-			"Z"
-			)
-
-/dmm_suite/save_map(var/turf/t1 as turf, var/turf/t2 as turf, var/map_name as text, var/flags as num)
-	//Check for illegal characters in file name... in a cheap way.
-	if(!((ckeyEx(map_name)==map_name) && ckeyEx(map_name)))
+/datum/dmm_suite
+	var/static/list/letter_digits = list(
```
AffectedArc07, line 9: Shouldnt this be in `dmm_suite.dm`

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -1,13 +1,13 @@
...
+// global datum that will preload variables on atoms instanciation
 GLOBAL_VAR_INIT(use_preloader, FALSE)
-GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
+GLOBAL_DATUM_INIT(_preloader, /datum/dmm_suite/preloader, new())
 
-/dmm_suite
+/datum/dmm_suite
 	// These regexes are global - meaning that starting the maploader again mid-load will
 	// reset progress - which means we need to track our index per-map, or we'll
 	// eternally recurse
```
AffectedArc07, line 13: Same with these regexes

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -470,12 +465,12 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
 // so that one can have separate "unpowered" areas for ruins or whatever,
 // yet have a single area type for use of mapping, instead of creating
 // a new area type for each new ruin
-/dmm_suite/loaded_map
+/datum/dmm_suite/loaded_map
 	parent_type = /datum
```
AffectedArc07, line 469: This isnt needed anymore

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +255,182 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
 
 //////////////////
-//Preloader datum
+// Preloader datum
 //////////////////
 
 // This ain't re-entrant, but we had this before the maploader update
-/dmm_suite/preloader
+/datum/dmm_suite/preloader
 	parent_type = /datum
```
AffectedArc07, line 428: Nor is this

Fox-McCloud:
I'm normally always for refactors--but this is a case of something so ancient and so untouched I'm afraid it'll break in ways we don't understand unless we're insanely careful.

We know it works, at the moment---dunno if we really want to touch it, aside from implementing TG's version of it.

@Crazylemon64 @AffectedArc07 your input is the most valuable here.

AffectedArc07:
TG **mostly** killed off dmm_suite to turn it into `/datum/parsed_map`, which creates a parsed map on `/New()` of the datum, which I think is quite honestly awful from a design perspective. 

I would also like to point out that I suggested this as a joke in the discord just saying `hey someone refactor /dmm_suite` for me, so power to @dearmochi for actually doing this. I will say the same thing as I said in my initial review of this PR. Please make sure this is **thoroughly** tested, and maybe even request a TM. 

I think this is a good idea, but we **MUST** make sure its fully working before merge.

code/modules/awaymissions/maploader/writer.dm
dearmochi: You're right, in my mind I thought it'd make more sense to put this in `writer.dmm` because it's exclusively used there but I think for consistency all static stuff should be in the main file along with `var/static/quote` and the regexes

code/modules/awaymissions/maploader/reader.dm
dearmochi: Moved

code/modules/awaymissions/maploader/reader.dm
dearmochi: You mean the `parent_type` right?

AffectedArc07: Yeah. You don’t need to set `parent_type` to datum when it’s already a datum.

dearmochi: Alright, removed

code/modules/awaymissions/maploader/reader.dm
dearmochi: Removed.

dearmochi:
https://github.com/ParadiseSS13/Paradise/blob/a640335c2986dc164dffd73f41f831e6e1a78c9c/code/modules/awaymissions/maploader/writer.dm#L106

Oof. During one of my mass replacements, I added a space in the coordinates to be written in `write_map`. That caused the regex to not recognize it and thus failing to load the map because it'd think it didn't have valid bounds. That is now fixed and after extensive testing it works again, as well as map loading (both away missions and smaller maps)

dearmochi:
I'm somewhat confident this now works at it should, but I'd recommend a TM with admins messing around in build mode to test out the saving/loading

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +243,181 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
 	var/turf/T
 	if(members[first_turf_index] != /turf/template_noop)
-		T = instance_atom(members[first_turf_index],members_attributes[first_turf_index],xcrd,ycrd,zcrd)
+		T = instance_atom(members[first_turf_index], members_attributes[first_turf_index], xcrd, ycrd, zcrd)
 
 	if(T)
-		//if others /turf are presents, simulates the underlays piling effect
+		// if others /turf are presents, simulates the underlays piling effect
 		index = first_turf_index + 1
 		while(index <= members.len - 1) // Last item is an /area
```
farie82, line 284: Best to cache the length prior to the while loop

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +243,181 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
 
 ////////////////
-//Helpers procs
+// Helpers procs
 ////////////////
 
-//Instance an atom at (x,y,z) and gives it the variables in attributes
-/dmm_suite/proc/instance_atom(path,list/attributes, x, y, z)
+// Instance an atom at (x, y, z) and gives it the variables in attributes
+/datum/dmm_suite/proc/instance_atom(path, list/attributes, x = 0, y = 0, z = 0)
```
farie82, line 304: This should never be called with default values so best to not add them either

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +243,181 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
+	var/turf/T = locate(x, y, z)
 	if(T)
 		if(ispath(path, /turf))
 			T.ChangeTurf(path, defer_change = TRUE, keep_icon = FALSE)
 			instance = T
 		else if(ispath(path, /area))
 
 		else
-			instance = new path (T)//first preloader pass
+			instance = new path (T) // first preloader pass
```
farie82, line 316: no space after `path` here

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +243,181 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
 		GLOB._preloader.load(instance)
 
 	return instance
 
-//text trimming (both directions) helper proc
-//optionally removes quotes before and after the text (for variable name)
-/dmm_suite/proc/trim_text(what as text,trim_quotes=0)
+// text trimming (both directions) helper proc
+// optionally removes quotes before and after the text (for variable name)
+/datum/dmm_suite/proc/trim_text(what = "", trim_quotes = FALSE)
```
farie82, line 325: The `what` should not be defaulted either. Calling this empty doesn't make much sense

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +243,181 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
 	else
 		return trimRegex.Replace(what, "")
 
-
-//find the position of the next delimiter,skipping whatever is comprised between opening_escape and closing_escape
-//returns 0 if reached the last delimiter
-/dmm_suite/proc/find_next_delimiter_position(text as text,initial_position as num, delimiter=",",opening_escape=quote,closing_escape=quote)
+// find the position of the next delimiter, skipping whatever is comprised between opening_escape and closing_escape
+// returns 0 if reached the last delimiter
+/datum/dmm_suite/proc/find_next_delimiter_position(text = "", initial_position = 0, delimiter = ",", opening_escape = quote, closing_escape = quote)
```
farie82, line 333: `text` here should not have a default

code/modules/awaymissions/maploader/reader.dm
```diff
@@ -255,186 +243,181 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new())
...
 	return next_delimiter
 
-
-//build a list from variables in text form (e.g {var1="derp"; var2; var3=7} => list(var1="derp", var2, var3=7))
-//return the filled list
-/dmm_suite/proc/readlist(text as text, delimiter=",")
-
+// build a list from variables in text form (e.g {var1="derp"; var2; var3=7} => list(var1="derp", var2, var3=7))
+// return the filled list
+/datum/dmm_suite/proc/readlist(text = "", delimiter = ",")
```
farie82, line 347: Same here. text shouldn't have a default

code/modules/awaymissions/maploader/writer.dm
```diff
@@ -54,96 +38,110 @@
...
-	template_buffer_text = jointext(template_buffer,"")
+
+	template_buffer_text = jointext(template_buffer, "")
 	log_debug("Reading turfs took [stop_watch(timer)]s.")
 
 	if(templates.len == 0)
 		CRASH("No templates found!")
-	var/key_length = round/*floor*/(log(letter_digits.len,templates.len-1)+1)
+
+	var/key_length = round/*floor*/(log(letter_digits.len, templates.len - 1) + 1)
```
farie82, line 63: Best to put this comment here behind the actual code. Or above it. Placing it in between... jeesh how odd

code/modules/awaymissions/maploader/writer.dm
```diff
@@ -54,96 +38,110 @@
...
 	log_debug("TOTAL TIME: [stop_watch(total_timer)]s.")
+
 	return dmm_text
 
-/dmm_suite/proc/make_template(var/turf/model as turf, var/flags as num)
-	var/use_json = 0
-	if(flags & DMM_USE_JSON)
-		use_json = 1
+/datum/dmm_suite/proc/make_template(turf/model, flags = 0)
+	var/use_json = flags & DMM_USE_JSON
```
farie82, line 118: `use_json` will now have the value of `DMM_USE_JSON` or 0. It will work but it's best to make this a boolean value just for consistency

code/modules/awaymissions/maploader/writer.dm
```diff
@@ -155,65 +153,68 @@
...
+			if(A.vars[T] != initial(A.vars[T]))
+				to_encode -= T
+				attributes += var_to_dmm(A.vars[T], T)
 
 		// Remove useless info
-		yeah -= "type"
-		if(yeah.len)
-			var/json_stuff = json_encode(yeah)
+		to_encode -= "type"
+		if(to_encode.len)
```
farie82, line 185: `length` is preferred over `.len`. It's both faster and null pointer safe

code/modules/awaymissions/maploader/writer.dm
```diff
@@ -155,65 +153,68 @@
...
-		attributes_text = copytext(attributes_text, 1, length(attributes_text)-1)
+	if(copytext(attributes_text, length(attributes_text) - 1, 0) == "; ")
+		attributes_text = copytext(attributes_text, 1, length(attributes_text) - 1)
+
 	attributes_text = "{[jointext(attributes,"; ")]}"
 	return attributes_text
 
-
-/dmm_suite/proc/get_model_key(var/which as num, var/key_length as num)
+/datum/dmm_suite/proc/get_model_key(which = 0, key_length = 0)
```
farie82, line 200: These should not have default values

Fox-McCloud:
If this breaks something, @AffectedArc07 has a token to come after you @dearmochi 

😛

PR #13646, opened 2020-06-19
Removes Set Background

Fox-McCloud:
Removes `set background` from the code.

This defaulted to `FALSE` in most cases.

what enabling the background does is insert a `sleep(-1)` between each proc in the loop; this can hypothetically cause race conditions----the actual sleep doesn't seem to kick in until tick_useage is really high....at a high enough point where this won't realistically reduce lag.

It does, however, potentially incur proc call overhead, so....getting rid of this is likely to offer a *tiny* speedup to a few procs.

:cl: Fox McCloud
del: removes antiquated code that was hypothetically slowing the game down
/:cl:

### DIFF
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -183,9 +183,6 @@
 
 
 /proc/get_mobs_in_radio_ranges(var/list/obj/item/radio/radios)
-
-	set background = 1
-
 	. = list()
 	// Returns a list of mobs who can hear any of the radios given in @radios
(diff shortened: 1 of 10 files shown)


AffectedArc07:
Quoting from the stonerman:
```
set background = 1 inserts a sleep(-1) to the body of every loop in the proc.

that's all it does.

sleep(-1) seems to only sleep once world.tick_usage gets to anywhere between 450% and 1350%

so its useless for removing lag.
```

PR #13650, opened 2020-06-20
Moves to BYOND 513

Fox-McCloud:
Fixes orbiting for 513--all without implementing the component (Frankly I'm glad).

This also moves us to 513; you'll be required to code in 513 and the live server will be required to run 513.

:cl: Fox McCloud
fix: Orbiting will no longer be weirdly offset; the tesla, dance machine, and ghosts should all look normal again
/:cl:


code/__HELPERS/unsorted.dm
```diff
@@ -1520,7 +1520,7 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 //orbit() can run without it (swap orbiting for A)
 //but then you can never stop it and that's just silly.
 /atom/movable/var/atom/orbiting = null
-
+/atom/movable/var/cached_transform = null
```
Fox-McCloud, line 1523: Making an entire var on every movable atom just for caching the transform is gross, but we can't safely do `initial(transform)` in this case, because not everything has default `transform` to begin with.

Fox-McCloud:
Went ahead and just made this a move to 513 and a refactor of the math fines.

Updates to TG's math defines and purges some old math we never used + moves all math related helpers/procs to be under the proper folder.

Also no more nonsense math***S*** name.

_build_dependencies.sh
```diff
@@ -8,4 +8,4 @@ export PHP_VERSION=5.6
 # Byond Major
 export BYOND_MAJOR=513
 # Byond Minor
-export BYOND_MINOR=1505
+export BYOND_MINOR=1514
```
AffectedArc07, line 11: `513.1526` is the stable version now, so I would use that

AffectedArc07:
This can take priority over #13331

PR #13655, opened 2020-06-20
Fix marked-paradise not deleting unwanted paragraphs properly

### LINKED ISSUE #13653: Paper Formatting Error
**Issue Description**:
Minor error in pencode formatting, unsure if this has been reported or not...frankly not going to look through 760 issues to be sure right now. At either rate, when using [center] before [large] it ignores the center tag entirely, and only makes large text. However, using [large][center] works as you would expect it to.

I expected pencode to not care which format tag is used first, as it makes no real sense for it to do so.

**What happened instead**:
Pencode is apparently picky as hell, and did in-fact care what tag is used first.

It will easily lead to incorrect formatting in papers and people assuming that formatting on a whole is currently broken, when in-fact it just has to be done unexpectedly specificially.

1. Grab paper.
2. Write [center][large]
3. Watch as you get large text but it's not centered.
4. Put [large][center].
5. Wonder why exactly it matters which goes first.

Unsure

**Extra information**:
Credit goes to Squishy for finding out this issue.

dearmochi:
## What Does This PR Do

Fixes an issue in `marked-paradise.js` where it would parse more nodes than needed, preventing the deletion of some `<p>` tags. The result is correct formatting regardless of tag order (fixes #13653)

## Why It's Good For The Game

Papers now display and lay out content that is consistent with the written code. It no longer throws a hissy fit if you use tags in the wrong order - it won't and shouldn't matter anymore. 
It also fixes some pre-made forms (namely the body scanner's report)

## Images of changes

_Before/after_
image image

image image

## Changelog
:cl:
fix: Fix papers not formatting as expected in some cases
/:cl:


html/browser/marked-paradise.js
```diff
@@ -4,24 +4,22 @@
...
-    if (node.children.length == 1) {
-        node.innerHTML = node.children[0].innerHTML;
-    }
+	for (var i = 0; i < node.childNodes.length; i++)
+		parse(node.childNodes[i]);
+
+	if (!node.innerHTML)
+		return;
+
+	if (node.children.length == 0) {
```
farie82, line 13: Actual change here

html/browser/marked-paradise.js
```diff
@@ -4,24 +4,22 @@
...
+		node.innerHTML = marked(node.innerHTML.replace(/<br>/gi, '\n').replace(/\t/gi, ''), { breaks: false, gfm: false });
+		// marked.js wraps content into <p> tags, which is looks atrocious when we call it recursively.
+		// The following line unwraps it.
+		if (node.children.length == 1 && node.children[0].tagName == "P")
```
farie82, line 17: And here

code/game/machinery/adv_med.dm
```diff
@@ -404,7 +404,7 @@
 			dat += "Ether: [occupant.reagents.get_reagent_amount("ether")] units<BR>"
 
 			extra_font = (occupant.reagents.get_reagent_amount("silver_sulfadiazine") < 30 ? "<font color='black'>" : "<font color='red'>")
-			dat += "[extra_font]\tSilver Sulfadiazine: [occupant.reagents.get_reagent_amount("silver_sulfadiazine")]</font><br>"
+			dat += "[extra_font]\tSilver Sulfadiazine: [occupant.reagents.get_reagent_amount("silver_sulfadiazine")] units</font><br>"
```
farie82, line 407: Undocumented change/fix here.

farie82:
Btw if this fixes #13653 then please put it in the PR.
Something like `fixes #13653` will auto close the issue

dearmochi:
> 
> 
> Btw if this fixes #13653 then please put it in the PR.
> Something like `fixes #13653` will auto close the issue

Whoops, thought including the issue number would be enough, but I guess I have to add `fixes` for it to work

AffectedArc07:
Considering the only "real" change in here are 2 JS changes and an addition of the word `units`, I recommend completley removing the bodyscanner changes and doing just the JS ones, especially since this will conflict with medical TGUIs. Before & After images would also be appreciated, instead of just after

PR #13656, opened 2020-06-21
Reagents Cleanup

Fox-McCloud:
Just cleans up the styling of the reagents holder folder.

This also makes all loops in `reagents_list` typeless and removes `istype` checks from it.

If someone is not utilizing reagents properly, this refactor WILL cause a mob (or anything that utilizes reagents) to runtime like *mad*---but having `istype` checks on a list that should *ONLY* contains reagents and *NEVER* contain nulls either is something we *want* to happen. The current way the runtime was "resolved" just allowed for bad behavior to persist and get papered over.

:cl: Fox McCloud
tweak: reagents performance slightly improved
/:cl:


code/modules/reagents/chemistry/holder.dm
```diff
@@ -53,20 +53,22 @@
 				GLOB.chemical_reactions_list[id] += D
 				break // Don't bother adding ourselves to other reagent ids, it is redundant.
 
-/datum/reagents/proc/remove_any(amount=1)
+/datum/reagents/proc/remove_any(amount = 1)
+	var/list/cached_reagents = reagent_list
```
Fox-McCloud, line 57: TG just uses this for every single time they utilize `reagent_list`. I don't like this; there's no reason to do this unless you're access the same list multiple times--and Id rather know, at a glance, if the list is being used, multiple times, in the same proc.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -486,21 +457,21 @@
 						var/obj/item/slime_extract/ME2 = my_atom
 						ME2.Uses--
 						if(ME2.Uses <= 0) // give the notification that the slime core is dead
-							for(var/mob/M in seen)
+							for(var/mob/living/M in seen)
```
Fox-McCloud, line 460: AI camera nonsense.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -42,7 +42,7 @@
 			var/datum/chemical_reaction/D = new path()
 			var/list/reaction_ids = list()
 
-			if(D && D.required_reagents && D.required_reagents.len)
+			if(D && D.required_reagents && length(D.required_reagents))
```
farie82, line 45: Can remove the `D.required_reagents` check here since length covers it.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -364,68 +365,38 @@
...
+		var/datum/reagent/R = AB
+		R.on_move(A, Running)
 	update_total()
 
-/datum/reagents/proc/conditional_update(atom/A, )
-	for(var/datum/reagent/R in reagent_list)
+/datum/reagents/proc/conditional_update(atom/A)
+	for(var/AB in reagent_list)
+		var/datum/reagent/R = AB
 		R.on_update (A)
```
farie82, line 377: weird spacing here

code/modules/reagents/chemistry/holder.dm
```diff
@@ -364,68 +365,38 @@
...
 				if(!reaction)
 					continue
 
 				var/datum/chemical_reaction/C = reaction
-				var/total_required_reagents = C.required_reagents.len
+				var/total_required_reagents = length(C.required_reagents)
 				var/total_matching_reagents = 0
-				var/total_required_catalysts = C.required_catalysts.len
+				var/total_required_catalysts = length(C.required_catalysts)
 				var/total_matching_catalysts= 0
```
farie82, line 397: spacing here as well

PR #13657, opened 2020-06-21
Terror Princess Rework

### LINKED ISSUE #12969: Terror Spiders Stuck in Webs
**Problem Description**:
As a terror spider, kept getting stuck and unable to move through normal giant spider webs

To be able to normally walk though the spider webs

**What happened instead**:
It kept telling me that I got stuck, and took me about eight tries to get through

Giant spiders are a mob that can be spawned through xenobiology, and can create webs much like terror spiders can. A benefit to the giant spiders is that they are able to walk through terror spider webs without any issue, the webs don't affect them. However, this doesn't hold true for the terror spiders, who are supposed to have stronger and harder to pass through webs.

I was playing as a Queen of Terror Spider, who had nested, and was filling tool storage with webs. A spiderling hatched into a giant spider, and was AI controlled, and put down a web. I went after it, and after killing it turned around and got stuck in the web trying to pass through. The first time I went through the web, this didn't happen though.

Nothing that I can think of that was hugely out of the ordinary or related to it.

Kyep:
## What Does This PR Do

Changes for all terror spiders:
- add: terror spiders now have status panel text telling them how fast their health is regenerating (and, when it is regenerating slowly, that this is because they are hungry). This helps players understand how regeneration mechanics work.
- tweak: terror spiders now use the 'bite' attack animation (same as xenos) instead of the previous generic 'smash' animation
- tweak: ghosts no longer get popups about spiders already under player control. They still get a ghost message in chat so they can locate the new spider.
- fix: fixed many cases where a spider was incorrectly referred to as "it" or "he".
- fix: fixed examining a spider showing they were dead twice.
- fix: fixes #12969 - terror spiders getting stuck in giant spider webs
- balance: Terror spiders now take 2x as much damage from having too little oxygen (or too much plasma) in the air they breathe.
- balance: Terror spiders now take 2.5x as much damage from being in an environment that is too hot. This stacks with the above, and hits them every tick until their body temperature cools down to being within normal range.
- balance: Terror spiders can no longer tolerate temperatures down to 100K and up to 500K. They now have default temperature tolerance for simple mobs, meaning that if you heat them up with a plasma fire they'll stay "too hot" for a lot longer.

Queen Changes:
- summary: queens are stronger, but only appear 1/3 as often (due to changes to whites/princesses below)
- del: removed "fakelings" ability of queens that did nothing but waste CPU.
- tweak: queen spit now deals 40 burn rather than 30 toxin, and affects objects it hits. So, queens can defend themselves against borgs, mechs, IPCs, etc, and use it to break objects (like cameras) from a distance. Conversely, it will now hurt other terrors if it hits them. Terrors, please don't stand in front of the queen.
- del: queens can no longer wrap objects. Queens never need to wrap corpses for health regen, and they should have other spiders to wrap other objects, so this ability was unnecessary and encouraged queens to waste corpses by wrapping them instead of giving them to another spider who needed them for health regen.
- tweak: queen webs no longer shimmer, no longer cause toxin damage, and no longer trigger hallucinations. Instead, they now block atmos, which enables queens to re-seal breached areas so that a single purple breaking a wall does not doom their nest to depressurization.
- tweak: removed the old "point cost" system for queen egglaying, where different spiders cost different numbers of "egg points". Now, all spiders cost 1 egg (more intuitive), and the queen is simply limited on how many she can have of the higher types at the same time. Queens also now get 5 initial eggs instead of 4.
- add: queens who survive 20+ minutes can now more easily lay a T3 (prince OR princess, NOT both) egg to serve them. If that spider dies, they can lay a new one. This effectively gives old queens a permanent T3 henchmen spider. Different T3s will provide different advantages to the hive.
- balance: queens now take slightly longer to reach stage 3 (20m instead of 15m). However, they're a bit stronger at stage 3, and they can keep leveling beyond stage 3 if they live long enough.
- add: examining a queen (when you are a ghost, or another tspider) will tell you how many minutes she has lived for, and how many eggs she has laid.
- fix: fixed a potential exploit with queen/princess egglaying
- add: the death of a queen now causes all her spiderlings, and half of her adult spiders, to die. This provides crew a way to still win against out-of-control queen infestations (by sniping the queen)

Princess Changes:
- summary: princesses have been totally reworked, and are no longer just prelude to queen. They're now they own caste with different gameplay.
- add: princesses now lay eggs like queens do, on a 2m timer, but they can't lay as many or maintain as many living spiders as a queen can. Unlike a queen, they can still ventcrawl, and don't need to "nest", but cannot spit. Overall they're mini-queens, weaker in egglaying but also not as essential for the hive, and they favor an evasion-based playstyle rather than a nest-defense playstyle. They can maintain a single tier 2 spider at a time - they must pick between black, brown and purple.
- tweak: princesses now color-shift their sprite based on how many spiders they've created, rather than how much they have fed.
- balance: princesses now spawn in a group of 3 during the princess infestation event. Overall they can produce more spiders than a queen early on, but they're at much greater risk of dying off than a queen is, and obviously each one that dies permanently reduces their egg-laying rate. Queens also get higher-tier offspring than princesses do.

White Changes:
- del: white terror infections no longer create princesses.
- add: instead, white terror infections have 5 infection stages (up from 3) and they always produce spiderlings in this order: gray, red, brown, green, green. This prevents white infestations from collapsing into the new princess infestation type. 
- balance: white infestations have been nerfed to be beatable, though the spiders can also win if their infections go untreated long enough to produce some browns, and the browns start opening vents for whites. Or they get greens and the greens work hard. Conversely, the fact they now get browns reduces the chances they'll get totally locked out of vents.

Refactors/Cleanups:
- refactor: Princesses are now a subtype of queen, for vastly improved code/var re-use
- refactor: Changed a bunch of for(... loops to use the var/thing standard.
- refactor: Refactored many procs for var re-use, early return statements, and better generalization

## Changelog
:cl: Kyep
add: Terrors now have status panel text letting them know when they're regenerating slowly because they're hungry.
tweak: Terrors now use the 'bite' attack animation
tweak: Ghosts no longer get popup notices about terrors they cannot control
fix: Fixed terrors being described as the wrong gender, and other bugs related to examining a spider.
fix: Fixed terrors getting stuck in giant spider webs.
add: Reworked Queen terrors. They're now stronger but rarer, their egg-laying system is more intuitive, and they have more interesting choices to make after they have lived for awhile.
add: Reworked Princess terrors. They are now their own caste with similar mechanics to queens, but a playstyle more focused on evasion than defending a static nest. They can maintain a smaller brood than a queen can, but they can ventcrawl.
add: Reworked white terrors to be weaker (more eggs per infestation, no princesses, guaranteed browns), and have their own unique playstyle rather than just being a production system for princesses.
/:cl:


Carthusia:
Daum, kyet loves him some spiders.

I like the changes

TDSSS:
Is there a way to tell at a glance what level a Queen is? Either a sprite change or a line in the description.

I also see you let Princesses lay Purples. I might be stating the obvious but I assume you made it so Purples don't die when near a Princess/count Princesses as queens?

BakKeener:
Since Princesses don't evolve into queens anymore I would love to see fluffy spiders from the princess so you can tell what spiders belong to which caste.

Kyep:
@Carthusia The princess sprites haven't changed in this update. They used to shift from pink to red as the princess wrapped more humanoid corpses. Now they shift (in both directions) based on how many children the princess has.

@TDSSS There has never been a way for other people to tell what level a queen is. Her combat stats never change... the only thing that changes for her is her egg-laying rate. And its a known thing that the older a queen is, the more eggs she lays. Regarding purples, a princess can support one purple, but if the purple leaves the princess to go rambo the purple will die off, just like if a purple abandons their queen. Note: this is based on the specific queen/princess they are linked to. Hanging around a different one will NOT keep them alive.

@BakKeener Explain please?

TDSSS:
> And its a known thing that the older a queen is, the more eggs she lays.

Is it, though? I would wager many people are unaware of this, so adding some indicator would definitely not hurt.

Kyep:
@TDSSS There are a ton of things I want to do for terrors. Including some big ones, like a total re-work of the Mother class. As well as giving Queens more stuff to do while laying eggs (like remote-viewing other spiders), so their gameplay is less boring. Thing is, I'm trying to keep this PR reasonably small. So I need to pick and choose what the top things are. A status indicator, while nice to have, isn't high up on the priority list.

BakKeener:
@Kyep Basically new sprites to show which spiders belong to who, like the current ones we have for spiders created by the queen, greens and whites then we have new ones for the princess spider since they won't be able to evolve into queens anymore. I just think it would be cool to have new sprites for spiders from the princess to be fluffy lookin instead just like the princess.

Kyep:
@BakKeener I still don't understand what you mean.

SpookSpooky:
I've got a single issue with this just from a players POV

Normally an infestation of terrors seems to always try and evolve towards a better spider, (except the green infestation those are different being a how much chaos can 5 greens readily do?) 
An example being, white terrors spawn, do their bites and webs infecting to get a princess will rarely come out and survive being a spider-ling, who will become a queen and then mass produce, the ability for whites to even get a princess has been removed forcing greens to do all the work and removing that evolution.

The other example is two princesses, one becomes a queen and mass produces, now its just two greens with basically limitless egg spamming and vent crawl causing a lack of nesting and more of a pop in, lay eggs and hope they don't find them.

Now the problem is there's no more evolution of spiders getting more powerful, whites would stay infecting and a lack of princesses just makes this stale once crew gets a hold of themselves (or didn't and ERT called). Two princess spawn would just do the exact same thing, find a quiet spot, make eggs and when they are found vent crawl. It becomes less of a struggle to power and more of "how many spiders can these princesses produce before gamma borgs get called in?"

The point I saw of nesting was the Queen has decided this place will infinitely reproduce spiders and cause crew to know where they based. And greens needing to wrap bodies to make eggs made sense, not letting them mass produce but allowing eggs to be away from the nest, and the front line assault against the nest. 

Anywho, looks cool, application might be different then what I think but terrors will always be terrifying.

Kyep:
@SpookSpooky I'm trying to move away from the "terror spiders evolve upwards to higher tiers" idea. The reason is that it collapses infestation types and ultimately reduces variety. If white infestations turn into princess infestations, and princess infestations turn into queen infestations, then over time the variety of infestations is reduced and there's less gameplay variety since most infestations sooner or later collapse into the queen type. Aside from that having less variety, it also constrains queen infestations in undesirable ways, such as by forcing me to balance them for all of the potential different routes they can be obtained by.

What I want to do instead is make each infestation type different, viable in its own way, with the higher tiers being stronger and the lower tiers weaker. So for example in this PR, white infestations no longer create princesses that turn into queens. Instead, each white infection produces more spiderlings of lower tiers, and one brown on the 3rd cycle. This makes white infestations a bit weaker (thank goodness, they were too strong before this), but it also makes them more varied, more like their own infestation type instead of simply a route to a queen.

Princesses are indeed strong, and that's intentional. They're meant to be the 2nd strongest infestation type, after a queen. Yes, in theory they can keep evading forever, laying eggs in unused areas. That's intended. Thing is, knowing players, they'll get ambushed by someone with a shotgun sooner or later, and unlike a queen that stays in a defended nest, they'll have shorter life expectancy. And still cannot be replaced. Since their egglaying ability is also limited in both numbers and tiers, a princess infestation won't (and shouldn't) have as high potential as a queen infestation... but it should still be more dangerous than a white infestation. I'll be watching the balance of this and adjusting over time as needed. I do believe, though, to the idea that the different terror infestations should play out genuinely differently and not all just collapse into the "queen in the well-defended nest" subtype.

Hope that all makes sense.

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -20,21 +21,41 @@
...
-		if(!L.stillborn && L.grow_as == specific_type && L.z == z)
-			numspiders += 1
+	var/check_list = islist(mytypes) && mytypes.len > 0
+	for(var/thing in GLOB.ts_spiderlist)
+		var/mob/living/simple_animal/hostile/poison/terror_spider/T = thing
+		if(T.stat == DEAD || T.spider_placed || spider_awaymission != T.spider_awaymission)
+			continue
+		if(check_mine && (!T.spider_myqueen || T.spider_myqueen != src))
+			continue
+		if(check_list && !(T.type in mytypes))
```
farie82, line 39: Performance wise.
You can build a type cache here and use that. Assoc list accessing is done in constant time. Where this takes linear time with the length of the list.
Seeing this again as a bonus though due to how little the proc is used.

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -20,21 +21,41 @@
...
+			continue
+		if(T == src)
+			continue
+		numspiders += 1
```
farie82, line 43: `numspiders++` will do. Same below

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -45,67 +45,92 @@
...
-					to_chat(src, "<span class='notice'>You have [canlay] eggs available to lay.</span>")
+				grant_eggs()
+
+
+/mob/living/simple_animal/hostile/poison/terror_spider/queen/proc/grant_eggs()
+	spider_lastspawn = world.time
+	canlay += getSpiderLevel()
+	if(canlay == 1)
+		to_chat(src, "<span class='notice'>You have an egg available to lay.</span>")
+	else if(canlay > 0)
```
farie82, line 83: `> 1` in this case. Result is the same though

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -45,67 +45,92 @@
...
+		to_chat(src, "<span class='notice'>You have [canlay] eggs available to lay.</span>")
+
+
+/mob/living/simple_animal/hostile/poison/terror_spider/queen/proc/getSpiderLevel()
+	return (1 + round(MinutesAlive() / 10))
```
farie82, line 88: outer () are not required.

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -45,67 +45,92 @@
...
+			if(!T.spider_myqueen)
+				continue
+			if(T.spider_myqueen != src)
+				continue
+			if(prob(50) || (T.spider_tier >= spider_tier))
+				to_chat(T, "<span class='userdanger'>\The psychic backlash from the death of [src] crashes into your mind! Somehow... you find a way to keep going!</span>")
+				continue
+			T.visible_message("<span class='danger'>[T] writhes in pain!</span>")
+			to_chat(T, "<span class='userdanger'>\The psychic backlash from the death of [src] overwhelms you! You feel the life start to drain out of you...</span>")
+			T.degenerate = 1
```
farie82, line 112: TRUE

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -45,67 +45,92 @@
...
-				P.degenerate = 1
+		// When a queen (or subtype!) dies, so do all of her spiderlings, and half of all her fully grown offspring
+		// This feature is intended to provide a way for crew to still win even if the queen has overwhelming numbers - by sniping the queen.
+		for(var/thing in GLOB.ts_spiderlist)
+			var/mob/living/simple_animal/hostile/poison/terror_spider/T = thing
+			if(!T.spider_myqueen)
+				continue
+			if(T.spider_myqueen != src)
+				continue
+			if(prob(50) || (T.spider_tier >= spider_tier))
```
farie82, line 107: () around the >= check are not required

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -299,58 +294,75 @@
...
+/mob/living/simple_animal/hostile/poison/terror_spider/queen/proc/ListAvailableEggTypes()
+	if(MinutesAlive() >= 20)
+		var/num_my_special = CountSpidersDetailed(TRUE, list(/mob/living/simple_animal/hostile/poison/terror_spider/mother, /mob/living/simple_animal/hostile/poison/terror_spider/prince, /mob/living/simple_animal/hostile/poison/terror_spider/queen/princess))
+		if(num_my_special == 0)
+			return list(TS_DESC_PRINCE, TS_DESC_PRINCESS) // Mother will be added to this list.... AFTER mothers are reworked.
+
+	var/list/valid_types = list(TS_DESC_RED, TS_DESC_GRAY, TS_DESC_GREEN)
+	var/num_browns = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/brown))
+	if(num_browns < 2)
+		valid_types |= TS_DESC_BROWN
```
farie82, line 318: += here instead of |=.
|= will look through the list to see if the to be added item is already present or not. Which can't be the case here

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -299,58 +294,75 @@
...
+
 
 /obj/structure/spider/terrorweb/queen
-	name = "shimmering web"
-	desc = "This web seems to shimmer all different colors in the light."
+	name = "airtight web"
+	desc = "This multi-layered web seems to be able to resist air pressure."
+
 
 /obj/structure/spider/terrorweb/queen/web_special_ability(mob/living/carbon/C)
```
farie82, line 359: Not required sincethe parent already does nothing

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -222,25 +229,23 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
 /mob/living/simple_animal/hostile/poison/terror_spider/examine(mob/user)
 	. = ..()
 	var/list/msgs = list()
-	if(stat == DEAD)
-		msgs += "<span class='notice'>It appears to be dead.</span>\n"
-	else
+	if(stat != DEAD)
 		if(key)
-			msgs += "<span class='warning'>Its eyes regard you with a curious intelligence.</span>"
+			msgs += "<span class='warning'>[p_they(TRUE)] regards [p_their()] surroundings with a curious intelligence.</span>"
```
farie82, line 234: can just do `. += message`.

Same goes for the others as well. Don't have to join manually here

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -365,21 +372,54 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
...
-		to_chat(src, "<span class='warning'>The door is welded shut.</span>")
+	if(D.welded)
+		to_chat(src, "<span class='warning'>The door is welded.</span>")
 	else if(D.locked)
-		to_chat(src, "<span class='warning'>The door is bolted shut.</span>")
+		to_chat(src, "<span class='warning'>The door is bolted.</span>")
 	else if(D.allowed(src))
-		D.open(1)
+		if(D.density)
+			D.open(1)
```
farie82, line 381: `TRUE` here. Same for the others

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -56,6 +66,10 @@
 			continue
 		if(check_list && L.grow_as && !(L.grow_as in mytypes))
 			continue
-		numspiders += 1
-	return numspiders
+		if(spider_totals[L.grow_as])
```
farie82, line 69: idem

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -210,14 +210,12 @@
 							neststep = 4
 						else
 							spider_lastspawn = world.time
-							var/num_purple = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/purple))
-							var/num_white = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/white))
-							var/num_brown = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/brown))
-							if(num_purple < 4)
+							var/spider_array = CountSpidersDetailed(FALSE)
```
farie82, line 213: `var/list/spider_array` here. It's a list that is being returned so best to type it as well

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -27,9 +27,11 @@
 			numspiders += 1
 	return numspiders
 
-/mob/living/simple_animal/hostile/poison/terror_spider/proc/CountSpidersDetailed(check_mine = FALSE, list/mytypes)
-	var/numspiders = 0
+/mob/living/simple_animal/hostile/poison/terror_spider/proc/CountSpidersDetailed(check_mine = FALSE, list/mytypes = list())
+	var/spider_totals = list("all" = 0)
```
farie82, line 31: `var/list/spider_totals` here. It makes a list so best to type it

code/modules/mob/living/simple_animal/hostile/terror_spiders/green.dm
```diff
@@ -39,11 +39,10 @@
 		to_chat(src, "<span class='warning'>You must wrap more humanoid prey before you can do this!</span>")
 		return
 	var/list/eggtypes = list(TS_DESC_RED, TS_DESC_GRAY, TS_DESC_GREEN)
-	var/num_brown = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/brown))
-	if(num_brown < 2)
+	var/spider_array = CountSpidersDetailed(FALSE)
```
farie82, line 42: `var/list/spider_array`

code/modules/mob/living/simple_animal/hostile/terror_spiders/princess.dm
```diff
@@ -112,8 +117,8 @@
 
 
 /mob/living/simple_animal/hostile/poison/terror_spider/queen/princess/ai_nest_is_full()
-	var/brood_count = CountSpidersDetailed(TRUE)
-	if(brood_count >= spider_max_children)
+	var/spider_array = CountSpidersDetailed(TRUE)
```
farie82, line 120: `var/list/spider_array`

code/modules/mob/living/simple_animal/hostile/terror_spiders/princess.dm
```diff
@@ -62,7 +62,8 @@
 	if(!prob(grant_prob))
 		return
 
-	var/brood_count = CountSpidersDetailed(TRUE)
+	var/spider_array = CountSpidersDetailed(TRUE)
```
farie82, line 65: `var/list/spider_array`

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -308,19 +306,17 @@
 
 /mob/living/simple_animal/hostile/poison/terror_spider/queen/proc/ListAvailableEggTypes()
 	if(MinutesAlive() >= 20)
-		var/num_my_special = CountSpidersDetailed(TRUE, list(/mob/living/simple_animal/hostile/poison/terror_spider/mother, /mob/living/simple_animal/hostile/poison/terror_spider/prince, /mob/living/simple_animal/hostile/poison/terror_spider/queen/princess))
-		if(num_my_special == 0)
+		var/spider_array = CountSpidersDetailed(TRUE, list(/mob/living/simple_animal/hostile/poison/terror_spider/mother, /mob/living/simple_animal/hostile/poison/terror_spider/prince, /mob/living/simple_animal/hostile/poison/terror_spider/queen/princess))
```
farie82, line 309: `var/list/spider_array`

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -308,19 +306,17 @@
...
+		if(spider_array["all"] == 0)
 			return list(TS_DESC_PRINCE, TS_DESC_PRINCESS) // Mother will be added to this list.... AFTER mothers are reworked.
 
 	var/list/valid_types = list(TS_DESC_RED, TS_DESC_GRAY, TS_DESC_GREEN)
-	var/num_browns = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/brown))
-	if(num_browns < 2)
+	var/spider_array = CountSpidersDetailed(FALSE, list(/mob/living/simple_animal/hostile/poison/terror_spider/brown, /mob/living/simple_animal/hostile/poison/terror_spider/purple, /mob/living/simple_animal/hostile/poison/terror_spider/black))
```
farie82, line 314: `var/list/spider_array`

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -40,14 +42,22 @@
 			continue
 		if(T == src)
 			continue
-		numspiders++
+		if(spider_totals[T.type])
```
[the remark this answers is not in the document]
Kyep: No I can't, as that only initializes passed types.

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -40,14 +42,22 @@
...
+			spider_totals[T.type] = 1
+		spider_totals["all"]++
 	for(var/thing in GLOB.ts_egg_list)
 		var/obj/structure/spider/eggcluster/terror_eggcluster/E = thing
 		if(check_mine && (!E.spider_myqueen || E.spider_myqueen != src))
 			continue
 		if(check_list && E.spiderling_type && !(E.spiderling_type in mytypes))
 			continue
-		numspiders += E.spiderling_number
+		if(spider_totals[E.spiderling_type])
```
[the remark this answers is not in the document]
Kyep: No, same reason as above.

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
Kyep: As above

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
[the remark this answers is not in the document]
farie82: True that. I forgot the case where you can pass an empty list or null which will count all the spiders

code/modules/mob/living/simple_animal/hostile/terror_spiders/hive.dm
```diff
@@ -20,21 +21,55 @@
...
+/mob/living/simple_animal/hostile/poison/terror_spider/proc/CountSpidersDetailed(check_mine = FALSE, list/mytypes = list())
+	var/list/spider_totals = list("all" = 0)
+	var/check_list = length(mytypes) > 0
+	for(var/thistype in mytypes)
+		spider_totals[thistype] = 0
+	for(var/thing in GLOB.ts_spiderlist)
+		var/mob/living/simple_animal/hostile/poison/terror_spider/T = thing
+		if(T.stat == DEAD || T.spider_placed || spider_awaymission != T.spider_awaymission)
+			continue
+		if(check_mine && (!T.spider_myqueen || T.spider_myqueen != src))
```
farie82, line 39: Just a simple `E.spider_myqueen != src` will do. if `spider_myqueen` is null then it also won't be equal to src.
Same applies to the other 2 checks below

code/modules/mob/living/simple_animal/hostile/terror_spiders/queen.dm
```diff
@@ -287,6 +278,8 @@
 	if(eggtype == null || numlings == null)
 		to_chat(src, "<span class='danger'>Cancelled.</span>")
 		return
+
+	// Actually lay the eggs.
 	canlay -= numlings
 	eggslaid += numlings
 	if(eggtype == TS_DESC_RED)
```
farie82, line 285: Can make a switch here instead of an if else chain

TDSSS:
> add: the death of a queen now causes all her spiderlings, and half of her adult spiders, to die. This provides crew a way to still win against out-of-control queen infestations (by sniping the queen)

This will encourage spiders to play extremely passively, once their swarm is established. I'm a bit worried that's going to be very boring for the spider player in question. If the good play is to run away and never engage anyone, ever, because if you die half your team instantly dies as well. It wouldn't be so bad if spiders had a way to act remotely and still help their team somehow.

SteelSlayer:
Definitely agree with TDSSS, that would make queens super boring to play.

Kyep:
@TDSSS @SteelSlayer bear in mind... high level queens can have a LOT of spiders, including a T3. Without a change like that, it becomes semi-unwinnable for the crew. I want there to still be a way crew can beat overwhelming odds by sniping the queen. This also gives the other spiders a serious reason to protect the queen.

Kyep:
I've decided to add some tweaks in this PR, designed to discourage the "spiders create plasma fires" meta:
- balance: Terror spiders now take 2x as much damage from having too little oxygen (or too much plasma) in the air they breathe.
- balance: Terror spiders now take 2.5x as much damage from being in an environment that is too hot. This stacks with the above, and hits them every tick until their body temperature cools down to being within normal range.
- balance: Terror spiders can no longer tolerate temperatures down to 100K and up to 500K. They now have default temperature resistance meaning that if you heat them up with a plasma fire they'll stay "too hot" for a lot longer.

PR #12451, opened 2019-09-29
Mining Cyborgs get NOTHING (But the mining vendor gets rearranged and gets a new cell)

PidgeyThePirate:
## What Does This PR Do
- Reorganizes the mining vendor into categories for easier navigation (once people relearn it).
- Adds the Thermal Cell, costing 600 points from the vendor. It is a 10k capacity cell that charges at a rate of 50 units/tick (compared to 125 per tick for the slime core) when outside in lavaland or in other high temperature environments.

[Vendor Appearance](https://imgur.com/a/zBxbQO0)

## Why It's Good For The Game
The mining vendor has a crappy UI, and I improved it a bit. The thermal cell is a hold-over from the previous version of this PR, but I left it in because it still may have some value in future mining cyborgs or mechs.

## Changelog
:cl:
add: The Thermal Cell: 10k power, recharges 50 energy/tick in lavaland or hot environs.
add: Added the Thermal Cell to the mining vendor for 600 points.
tweak: The mining vendor UI has been reorganized and category spacers added.
/:cl:

PR #13658, opened 2020-06-21
Reorganizes the prizes in the Mining Vendor

PidgeyThePirate:
## What Does This PR Do
Reorganizes the mining vendor prizes into categories with more used categories at the top. The prizes in each category are alphabetically sorted. The default windows are a little taller.

## Why It's Good For The Game
Sorting through an unordered list to find your gear sucks even if you're used to it.

## Images of changes
image
image
image

## Changelog
:cl:
tweak: reorganized the mining vendor into categories and made the default window a little taller.
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -8,8 +8,7 @@
 	density = 1
 	anchored = 1.0
 	var/obj/item/card/id/inserted_id
-	var/list/prize_list = list( //Sorted by category, then alphabetical order.
-		new /datum/data/mining_equipment("Gear"),
+	var/list/gear_list = list(
 		new /datum/data/mining_equipment("Advanced Scanner",			/obj/item/t_scanner/adv_mining_scanner,								800),
 		new /datum/data/mining_equipment("Explorer's Webbing",			/obj/item/storage/belt/mining,										500),
 		new /datum/data/mining_equipment("Fulton Beacon",				/obj/item/fulton_core,												400),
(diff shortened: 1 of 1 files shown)


Aidenzack1:
As a person who plays miner semi-frequently, I cant say how much I would love this simple quality of life improvement, and I cant tell you how much time I've wasted looking for something in the menu. The only thing more I could ask for would be a searchbar, but I could see the many problems with that

Fox-McCloud:
This was rejected once before: https://github.com/ParadiseSS13/Paradise/pull/12451

This really needs to be made into a TGUI thing (once that gets merged), and the datums can be organized by category.

Abusing the fact that something doesn't have a price to be a breaker doesn't set well with me.

@Crazylemon64 might want to give your input.

PidgeyThePirate:
It was rejected once before because it was packaged with a mining borg update and a new cell, and this part of it was considerably better received than either. TGUI isn't in yet and there's no guarantee that it will come in the immediate future. This will do fine as a stopgap, I think.

Still, this was a couple hours' work so I'm not married to it.

AffectedArc07:
There should be lists for each category with categorisation headers as part of the UI itself. Making it so items with 0 cost dont show up is one of the hackiest things I have seen, and setting an item to 0 cost just to make it a linebreak is not a solution.

code/modules/mining/machine_vending.dm
```diff
@@ -172,7 +209,7 @@
 	if(href_list["purchase"])
 		if(istype(inserted_id))
 			var/datum/data/mining_equipment/prize = locate(href_list["purchase"])
-			if(!prize || !(prize in prize_list) || prize.cost > inserted_id.mining_points)
+			if(!prize || prize.cost > inserted_id.mining_points)
```
farie82, line 212: This is exploitable atm. Very minor exploit but.
In theory people can give a reference to a prize that is not part of this machine. For instance a grey slime extract from a golem vendor. If they somehow get that reference.

PidgeyThePirate:
Oh, that one's up now.

Issue #12822, opened 2019-12-18
Strange crematorium camera bug


lirdos:
**Issue Description**:
If you orbit someone, and then disconnect, and then that someone gets cremated, your camera will no longer lock onto you, the player.

**What did you expect to happen**: 
Camera should stay locked on my ghost.

**What happened instead**:
Camera stayed inside the crematorium.

**Why is this bad/What are the consequences**:
Pretty inconvenient if you wanted to observe the round, I'd imagine.

**Steps to reproduce the problem**:
-Orbit someone
-at this point my internet shut off, so my client was disconnected
-Have someone cremate that someone. Don't open the crematorium.
-Come back, click-teleport outside the crematorium, camera no longer focused on you

**When did the problem start happening**:
Not sure. Probably an older bug?

**Extra information**:
I've included a screenshot, to explain the result better, if that helps.
![image](https://user-images.githubusercontent.com/59033413/71126196-8c4f6780-21b6-11ea-970e-559c2f628707.png)

meow20:
Will happen in morgue trays too

matttheficus:
Fixed by #17221 and a duplicate of #10840 , albeit a subtype of the trays.

PR #13661, opened 2020-06-21
Minor GC Tweaks

Fox-McCloud:
Just some minor tweaks with garbage collection.

Follows TG's lead in calling parent before completing other tasks.

Also ensures things stop orbiting when they're GC, which clears another atoms reference.

:cl: Fox McCloud
fix: Makes objects garbage collect better, enhancing performance
del: removes drop_item verb, for whaever reason that existed; I doubt anyone used it.
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -44,16 +44,17 @@
 
 /atom/movable/Destroy()
 	unbuckle_all_mobs(force = TRUE)
+
+	. = ..()
 	if(loc)
 		loc.handle_atom_del(src)
 	for(var/atom/movable/AM in contents)
 		qdel(AM)
 	loc = null
 	if(pulledby)
-		if(pulledby.pulling == src)
-			pulledby.pulling = null
-		pulledby = null
-	return ..()
+		pulledby.stop_pulling()
+	if(orbiting)
+		stop_orbit()
 
 //Returns an atom's power cell, if it has one. Overload for individual items.
 /atom/movable/proc/get_cell()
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -64,7 +64,7 @@
 	..()
 	can_hold = typecacheof(can_hold)
 
-/obj/item/gripper/verb/drop_item()
+/obj/item/gripper/verb/drop_item_gripped()
```
Fox-McCloud, line 67: naming a verb the same thing as an existing proc isn't a generally good practice.

code/modules/mob/mob_movement.dm
```diff
@@ -26,25 +26,6 @@
 	else
 		to_chat(usr, "<span class='danger'>This mob type cannot throw items.</span>")
 
-
-/client/verb/drop_item()
```
farie82, line 30: Will remove the `drop-item` verb for people to use while typing in the command bar.
Dunno if anybody uses that though but it's a removal that is not described in the PR

code/game/atoms_movable.dm
```diff
@@ -44,16 +44,17 @@
 
 /atom/movable/Destroy()
```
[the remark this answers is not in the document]
Fox-McCloud: Going to take a pass on that one for, now. Normally I'd agree, but I really don't want to do an extra layer of lighting debugging right now xP

farie82:
Might fix #12822

PR #13668, opened 2020-06-23
Progress Bar Refactor

Fox-McCloud:
Ports TG's refactor to progress bars.

The animation for progress bars that was added broke garbage collection of them---as such, they're one of the most common things that fails to GC---this rectifies that.

This also adds a bunch of `qdeleted` checks to progress bars so it's not done on things that are deleted---which has weird/unintended results.

Lastly, as a feature/balance implication, this allows you to use objects on yourself while drifting through space--which is consistent with other changes made for cuff resisting, while in space.

:cl: Fox McCloud
fix: Fixes progress bars incurring more server performance than  they should
add: can use tools that have a delay on your self while drifting through space
/:cl:


code/datums/progressbar.dm
```diff
@@ -1,70 +1,133 @@
...
-	var/client/client
-	var/listindex
+	///The client seeing the progress bar.
+	var/client/user_client
+	///Effectively the number of steps the progress bar will need to do before reaching completion.
+	var/goal = 1
+	///Control check to see if the progress was interrupted before reaching its goal.
+	var/last_progress = 0
+	///Variable to ensure smooth visual stacking on multiple progress bars.
+	var/listindex = 0
```
AffectedArc07, line 18: Mm yes documentation

code/datums/progressbar.dm
```diff
@@ -1,70 +1,133 @@
...
-	if(user)
-		client = user.client
 
-	LAZYINITLIST(user.progressbars)
-	LAZYINITLIST(user.progressbars[bar.loc])
-	var/list/bars = user.progressbars[bar.loc]
-	bars.Add(src)
+	LAZYADDASSOC(user.progressbars, bar_loc, src)
+	var/list/bars = user.progressbars[bar_loc]
 	listindex = bars.len
```
farie82, line 42: `length` here

code/modules/mob/mob.dm
```diff
@@ -3,6 +3,9 @@
 	GLOB.dead_mob_list -= src
 	GLOB.alive_mob_list -= src
 	focus = null
+	if(length(progressbars))
+		stack_trace("[src] destroyed with elements in its progressbars list")
```
farie82, line 7: Maybe add a comment here saying it is being deleted due to the signal `COMSIG_PARENT_QDELETING` being listened to by the progress bars

PR #13670, opened 2020-06-23
Tcomms rework tweaks

AffectedArc07:
## What Does This PR Do
Fixes turfs in the metastation sat being the wrong kind, as well as a missing transit tube piece. It also makes it so AIs cannot disable the ionspheric anomalie event

image

## Why It's Good For The Game
The AI shouldnt able to instantly null out a random event, and broken turfs are bad

## Changelog
:cl:
tweak: You can no longer toggle the tcomms core to null out the ionspheric anomalie event
fix: Fixes some broken turfs on MetaStation
/:cl:

### DIFF
--- a/_maps/map_files/MetaStation/MetaStation.v41A.II.dmm
+++ b/_maps/map_files/MetaStation/MetaStation.v41A.II.dmm
@@ -43780,6 +43780,11 @@
 	d2 = 8;
 	icon_state = "4-8"
 	},
+/obj/structure/transit_tube/station{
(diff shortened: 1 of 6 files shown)


farie82:
Design wise. I think readding the `emp_act` back will be better. And putting a disable timer on the machine. Kinda like the ion var in the current PR.
So an EMP will disable it for a (set) duration and will reset the disable variable once it's done. This way people can EMP the tcomms machine to disable it temporarily. Would generate some high risk decisions to cut comms for a duration

PR #13673, opened 2020-06-23
Fixes brainless cyborgs, Removes "Android" alt job title for cyborgs

Kyep:
## What Does This PR Do
1. Alters the spawning process for cyborgs such that:
- Players spawning as the cyborg job with their job title set to "robot" get robotic brains.
- In all other cases, the cyborg gets a MMI.
2. Removes the "Android" alt cyborg job title.

## Why It's Good For The Game
1. Fixes a bug where players whose cyborg alt job title is set to "Robot" spawn without any brain at all, which leads to nasty results (they cannot have their brain removed by robotics to be put into a new body, which means if they're very damaged players have to ahelp to get us to manually fix it by respawning them). It is also more consistent - we should be able to count on cyborgs always having some form of brain.
2. Removing the Android job title was requested by Fox, and the title (as far as I could see) was pointless anyway.

## Changelog
:cl: Kyep
fix: Fixed a bug where players joining as a cyborg with the alt job title "Robot" were not given a brain.
del: Removed "Android" alt job title for cyborgs.
/:cl:

### DIFF
--- a/code/modules/mob/transform_procs.dm
+++ b/code/modules/mob/transform_procs.dm
@@ -82,14 +82,11 @@
 	O.notify_ai(1)
 
 	if(O.mind && O.mind.assigned_role == "Cyborg")
-		if(O.mind.role_alt_title == "Android")
+		if(O.mind.role_alt_title == "Android" || O.mind.role_alt_title == "Robot")
 			O.mmi = new /obj/item/mmi/robotic_brain(O)
-		else if(O.mind.role_alt_title == "Robot")
-			O.mmi = null //Robots do not have removable brains.
 		else
 			O.mmi = new /obj/item/mmi(O)
-
-		if(O.mmi) O.mmi.transfer_identity(src) //Does not transfer key/client.
+		O.mmi.transfer_identity(src) //Does not transfer key/client.
 
 	O.update_pipe_vision()
 
--- a/code/modules/mob/transform_procs.dm
+++ b/code/modules/mob/transform_procs.dm
@@ -82,7 +82,7 @@
 	O.notify_ai(1)
 
 	if(O.mind && O.mind.assigned_role == "Cyborg")
-		if(O.mind.role_alt_title == "Android" || O.mind.role_alt_title == "Robot")
+		if(O.mind.role_alt_title == "Robot")
 			O.mmi = new /obj/item/mmi/robotic_brain(O)
 		else
 			O.mmi = new /obj/item/mmi(O)
--- a/code/game/jobs/job/silicon.dm
+++ b/code/game/jobs/job/silicon.dm
@@ -32,7 +32,7 @@
 	minimal_player_age = 21
 	exp_requirements = 300
 	exp_type = EXP_TYPE_CREW
-	alt_titles = list("Android", "Robot")
+	alt_titles = list("Robot")
 
 /datum/job/cyborg/equip(mob/living/carbon/human/H)
 	if(!H)


code/modules/mob/transform_procs.dm
```diff
@@ -82,14 +82,11 @@
 	O.notify_ai(1)
 
 	if(O.mind && O.mind.assigned_role == "Cyborg")
-		if(O.mind.role_alt_title == "Android")
+		if(O.mind.role_alt_title == "Android" || O.mind.role_alt_title == "Robot")
```
Fox-McCloud, line 85: I'd prefer android title just be axed, entirely.

Kyep: Why? It seems like a redundant title to me, so I don't object to axing it, but when making a request it is a good idea to explain your reason.

SQL/updates/12-13.sql
```diff
@@ -0,0 +1,4 @@
+#Updating the SQL from version 12 to version 13. -Kyep
+#Removes support for the "Android" alt title for the Cyborg job
+
+UPDATE `characters` SET player_alt_titles = REPLACE(player_alt_titles, 'Cyborg=Android','Cyborg=Robot') WHERE player_alt_titles LIKE '%Cyborg=Android%';
```
AffectedArc07, line 4: DB version is only really incremented when the database undergoes a schema edit (table modification). The game can likely gracefully handle this, and a DB version bump should not be done for the sake of an update query imo

code/__DEFINES/misc.dm
```diff
@@ -319,7 +319,7 @@
 #define INVESTIGATE_BOMB "bombs"
 
 // The SQL version required by this version of the code
-#define SQL_VERSION 12
+#define SQL_VERSION 13
```
AffectedArc07, line 322: Same here

SQL/updates/12-13.sql
Kyep: I'd rather not update the DB version for this either, but Fox requested it.

config/example/dbconfig.txt
```diff
@@ -9,7 +9,7 @@
 ## This value must be set to the version of the paradise schema in use.
 ## If this value does not match, the SQL database will not be loaded and an error will be generated.
 ## Roundstart will be delayed.
-DB_VERSION 12
+DB_VERSION 13
```
KasparoVy, line 12: Should use SS13 DB version 13 on something epic

Spacemanspark:
1) the robot spawning without a brain is actually intentional
2) Android is a neat title, and to my knowledge it's actually meant to have you start with a robot brain as opposed to an MMI.

As such, opposed to this. I wouldn't mind the removal of the 'robot' title, but the android one I'd rather remain, if only for fluff reasons.

SQL/updates/12-13.sql
Fox-McCloud: I didn't request a DB version update; I only requested a change be made to cleanse the existing table---not update the entire schema.

Fox-McCloud:
"Android" is a 100%, objectively wrong title to have for cyborgs.

Android applies only to humanoid robots that are extremely difficult (if not impossible) to differentiate from humans.

It'd be like calling Ian a cat because "well, they're technically both mammals".

Spacemanspark:
And on the flip side of that argument, a cyborg is a human that's been modified robotically. It makes little sense for a robotic brain inserted into a robotic shell to be called such, does it not?

SQL/updates/12-13.sql
Kyep: Okay, PR updated.

config/example/dbconfig.txt
Kyep: Removed DB version incrementing.

Kyep:
@Fox-McCloud @AffectedArc07  Updated to remove the DB version change. I will still run the query after this is merged to update existing player prefs.
The PR still removes "Android" job title and fixes brainless issue.

AffectedArc07:
@Fox-McCloud Any reason SQL change was added here?

Spacemanspark:
I imagine because it will remove the title from anyone currently using it in the SQL

Fox-McCloud:
I threw it up because it does require a SQL command to be entered once this merges---we typically throw the label on anything that requires  command to be input, whether it  changes the schema or not---just so no one forgets.

PR #13674, opened 2020-06-23
Removes Excessive Automated Yakking

Fox-McCloud:
Removes ATC, Busyspace, Useless Yakking Simulator, Chatterbox 9000. Whatever you want to call it.

You know those useless messages that periodically get spammed over common channel---about so and so requesting docking at so and so, or X happening at Y location? Yeah, that's what this is.

Rips out the entire system.

I've always hated this system, personally. I predicted it'd be hyped up, people would think it's super neat for all of a week, then it'd promptly be ignored and thought of as mildly annoying from then on out.

It doesn't do anything, and it doesn't really help with immersion or roleplay---no one roleplays around it or talks about it---and with as many players as we have, it's just more useless chatter to clog up commons.

ALSO ITS AN UNREGULATED WHILE LOOP THAT IS TRYING TO ACT LIKE A SUBSYSTEM DESPITE NOT BEING ONE WHY?!

It also uses up CPU cycles best used for something else, though...I stress---it's VERY VERY low overall cost, so pitching this as a "performance" PR would be dishonest.

:cl: Fox McCloud
del: Removes random meaningless conversations from common radio
/:cl:


Mitchs98:
Honestly the extra spam has always just annoyed me personally.

code/modules/busy_space/organizations.dm
```diff
@@ -1,549 +0,0 @@
...
-		"Unwitting Accomplice",
-		"Bad For Business",
-		"Just Testing",
-		"Yawning Angel",
-		"Liveware Problem",
-		"Very Little Gravitas Indeed",
-		"Zero Gravitas",
-		"Gravitas Free Zone",
-		"Absolutely No You-Know-What",
-		"Existence Is Pain",
```
AffectedArc07, line 312: Yep

code/modules/busy_space/air_traffic.dm
```diff
@@ -1,128 +0,0 @@
...
-/datum/lore/atc_controller/process()
-	if(world.time >= next_message)
-		if(squelched)
-			next_message = world.time + backoff_delay
-		else
-			next_message = world.time + rand(delay_min,delay_max)
-			random_convo()
-
-	spawn(1 MINUTES) //We don't really need high-accuracy here.
-		process()
```
AffectedArc07, line 28: Why was this never subsystemised is my question

Fox-McCloud:
For clarity. It's not that I'm opposed to little tiny bits of fluff or immersion.

Those kind of features are great. I just don't think this is a great way to go about it and what impact it has is minimal to both immersion and roleplay---which actually actively *discouraging* people (as small as it may be) to listen to the common radio channel.

Carthusia:
It would be neat if the messages you hear act as hints to what has spawned that round in space, ATC reports a lost freighter, you could find that freighter in a space Z level, that kind of stuff

TDSSS:
It was always amusing when diplomatic vessels requested flight data for a raid or whatever but yeah, I don't think we need the radio clutter.

It's neat in that it reminds people of the existence of certain factions I guess.

MrMagolor:
>It was always amusing when diplomatic vessels requested flight data for a raid

I believe that's only Vox ships

Balthusdire:
I really like the idea, but yeah it hasn't lived up to the intent. Best to gentle put it to bed.

PR #13675, opened 2020-06-24
Finishes out Reagent Container Refactor

Fox-McCloud:
@Anasari ported the refactor to reagent containers for if they're drainable or not--and that caught most of the problems---but glass containers were never refactored.

As such, there was some oddities about them--you couldn't use them in a lot of circumstances where you had to use a drinking glass or something similar.

This meant that you couldn't use a beaker or other container on a fermenting barrel, for instance (I'm sure there's other things too).

Either case, this finishes out the refactor for those containers so anything that is a glass subtype should behave more consistently.

:cl: Fox McCloud
fix: Fixes glass containers behaving inconsistently. Can now use beakers and other glass containers to get liquid from a fermenting barrel
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/glass_containers.dm
+++ b/code/modules/reagents/reagent_containers/glass_containers.dm
@@ -14,38 +14,7 @@
 	container_type = OPENCONTAINER
 	has_lid = TRUE
 	resistance_flags = ACID_PROOF
-
 	var/label_text = ""
-	// the fucking asshole who designed this can go die in a fire - Iamgoofball
-	var/list/can_be_placed_into = list(
-		/obj/machinery/chem_master/,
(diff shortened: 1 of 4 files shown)


code/modules/reagents/reagent_containers/glass_containers.dm
```diff
@@ -14,38 +14,7 @@
 	container_type = OPENCONTAINER
 	has_lid = TRUE
 	resistance_flags = ACID_PROOF
-
 	var/label_text = ""
-	// the fucking asshole who designed this can go die in a fire - Iamgoofball
-	var/list/can_be_placed_into = list(
```
Fox-McCloud, line 20: thank goodness this is dead.

PR #13677, opened 2020-06-24
Allows SSinput to recover after an MC crash

AffectedArc07:
## What Does This PR Do
This allows SSinput to preserve its data after being recovered from an MC crash. This is integral to functionality because otherwise the system will have to be force-initialized after recovery (VV -> Call proc -> Initialize) to regenerate the lists, which are **required** for anyone in the game to be able to move.

## Why It's Good For The Game
Subsystems should be able to recover from crashes which 99% of the time aren't their fault

## Changelog
:cl: AffectedArc07
fix: SSinput will now recover properly from an MC crash. 
/:cl:


Fox-McCloud:
I hate this SS so much.

PR #13680, opened 2020-06-24
Add character slot clearing

dearmochi:
## What Does This PR Do

- Adds ability to clear a character slot, erasing it from the database. It requires confirmation from the user by prompting them to enter the character's full name
- Adds **(empty)** next to the slot number if the current slot isn't saved to the database 
- Slot number now gets its own line instead of being aligned with the slot buttons

## Why It's Good For The Game

I made this PR because I wanted to enable Randomized Character Slot but I remembered my old ~~anime~~ characters are still in there and I don't want to end up playing as them, nor do I want to adjust them

## Images of changes

X8Ppsmqhcv

## Changelog
:cl:
add: Add ability to clear character slots
/:cl:


Spacemanspark:
Might want to test and make sure this works with characters that have apostrophes in their names.

dearmochi:
> 
> 
> Might want to test and make sure this works with characters that have apostrophes in their names.

The deletion is performed on the row with `slot` equal to the player's `default_slot` which is an integer so it shouldn't cause any problems. I tried clearing a slot containing a character with an apostrophe in their name and it worked fine.

code/modules/client/preference/preferences_mysql.dm
```diff
@@ -495,3 +501,20 @@
 	load_character(C,pick(saves))
 	return 1
 
+/datum/preferences/proc/clear_character_slot(client/C)
+	. = FALSE
+	// Is there a character in that slot?
+	var/DBQuery/query = GLOB.dbcon.NewQuery("SELECT slot FROM [format_table_name("characters")] WHERE ckey='[C.ckey]' AND slot='[default_slot]'")
```
AffectedArc07, line 507: Please sanitize the ckey. I know its unlikely, but never underestimate the chance that someone joins called `'; DROP DATABASE 'paradise`

code/modules/client/preference/preferences_mysql.dm
```diff
@@ -495,3 +501,20 @@
...
+	query.Execute()
+	if(!query.RowCount())
+		return
+
+	query = GLOB.dbcon.NewQuery("DELETE FROM [format_table_name("characters")] WHERE ckey='[C.ckey]' AND slot='[default_slot]'")
```
AffectedArc07, line 512: Dont re-use the same query object. Make a new one.

code/modules/client/preference/preferences_mysql.dm
dearmochi: I agree with the idea that you should **always** sanitize anything you pass to SQL, but as we found out, it isn't needed here (or in any other SQL procs present in the file)

`world.log << ckey("'; DROP DATABASE 'db")` outputs `dropdatabasedb`

code/modules/client/preference/preferences_mysql.dm
dearmochi: Donezo

code/modules/client/preference/preferences_mysql.dm
```diff
@@ -132,6 +133,7 @@
 	// Let's not have this explode if you sneeze on the DB
 	var/DBQuery/query = GLOB.dbcon.NewQuery({"SELECT
 					OOC_Notes,
+					slot,
```
AffectedArc07, line 136: This currently breaks the database
image

PR #13688, opened 2020-06-25
Floor Tile Memory Useage and Initialization

Fox-McCloud:
A long time ago, there was an idea that, to cut down on how expensive the singulo was, each floor turf would have its tile built into it---then it'd get `forceMove`'d out when ti got pulled. Likewise, prying up a tile would remove it instead of creating a new one.

Thing is this really doesn't save much on CPU--creation of the tiles is...not expensive at all. Due to the nature of how stacks work, creating a new turf inherently means that the stack you used will not suddenly go into the new turf---and even if it did, you'd be creating a new item anyway.

All of this ends up slowing down initializations (by 1-2 seconds) and bloats up memory useage and instances by some 14,000+ objects.

My testing seems to show that this was using up about ~20 MB of memory--but that seems like an awful lot---still, it'll be a boost.

### DIFF
--- a/code/game/turfs/simulated/floor/light_floor.dm
+++ b/code/game/turfs/simulated/floor/light_floor.dm
@@ -77,14 +77,13 @@
 
 /turf/simulated/floor/light/attackby(obj/item/C, mob/user, params)
 	if(istype(C,/obj/item/light/bulb)) //only for light tiles
-		if(istype(builtin_tile, /obj/item/stack/tile/light))
-			if(!state)
-				qdel(C)
-				state = LIGHTFLOOR_ON
-				update_icon()
-				to_chat(user, "<span class='notice'>You replace the light bulb.</span>")
-			else
-				to_chat(user, "<span class='notice'>The light bulb seems fine, no need to replace it.</span>")
+		if(!state)
+			qdel(C)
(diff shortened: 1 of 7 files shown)


KasparoVy:
Damn I'm impressed, you guys are really going deep to slim down and improve efficiency. Never expected floor tiles of all things

code/game/turfs/simulated/floor/light_floor.dm
```diff
@@ -77,14 +77,13 @@
...
-				to_chat(user, "<span class='notice'>You replace the light bulb.</span>")
-			else
-				to_chat(user, "<span class='notice'>The light bulb seems fine, no need to replace it.</span>")
+		if(!state)
+			qdel(C)
+			state = LIGHTFLOOR_ON
+			update_icon()
+			to_chat(user, "<span class='notice'>You replace the light bulb.</span>")
+		else
+			to_chat(user, "<span class='notice'>The light bulb seems fine, no need to replace it.</span>")
```
AffectedArc07, line 86: Not really related to this PR, but do we even need this message? Not a whole load of actions will say "no need to X" when you do them (Only one I can think of is 'the airlock doesnt need repairing')

PR #13689, opened 2020-06-25
Structures Armor Fix

Fox-McCloud:
Having the armor set in `New` will invariably break things.

:cl: Fox McCloud
fix: Fixes structures behaving oddly for armor
/:cl:

### DIFF
--- a/code/game/objects/structures.dm
+++ b/code/game/objects/structures.dm
@@ -7,8 +7,6 @@
 	var/broken = FALSE
 
 /obj/structure/New()
-	if (!armor)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures.dm
```diff
@@ -20,6 +18,11 @@
 	if(SSticker)
 		GLOB.cameranet.updateVisibility(src)
 
+/obj/structure/Initialize(mapload)
+	if(!armor)
+		armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 50, "acid" = 50)
+	. = ..()
```
AffectedArc07, line 24: Maybe `return ..()` since that’s the standard here

PR #13690, opened 2020-06-25
Minor Turf Refactor

Fox-McCloud:
mostly from TG.

Turfs uhhhhh, aren't exactly going through the GC, in most cases.

I have no idea what kind of issues this is or is not creating, on line, but the only time turfs ever have their `Destroy` proc called is if they they are manually `qdel`'d, which is almost never how you interact with turfs.

This also refactors turfs to use `Initialize`--the sole exception is simulate turfs, for having air--otherwise, there will be mass runtimes; the air placement could be moved to `Initialize`, but the issue is that we don't have a good way to start tracking when a map has begun loading and finished---until we do, having `/turf/simulated` generate its air when it loads is the better deal.


code/game/turfs/turf.dm
```diff
@@ -70,16 +77,29 @@
 	log_startup_progress(" Smoothed atoms in [stop_watch(watch)]s.")
 	return 1
 
-/turf/Destroy()
-// Adds the adjacent turfs to the current atmos processing
+/turf/Destroy(force)
+	. = QDEL_HINT_IWILLGC
```
Fox-McCloud, line 81: You can't really delete turfs---one instance is invariably replaced with another

code/game/turfs/turf.dm
```diff
@@ -45,7 +45,12 @@
 		queue_smooth(src)
 
 /turf/Initialize(mapload)
-	. = ..()
+	if(initialized)
+		stack_trace("Warning: [src]([type]) initialized multiple times!")
+	initialized = TRUE
```
AffectedArc07, line 50: imo this should be right before the return

code/game/turfs/turf.dm
```diff
@@ -45,7 +45,12 @@
 		queue_smooth(src)
 
 /turf/Initialize(mapload)
-	. = ..()
+	if(initialized)
+		stack_trace("Warning: [src]([type]) initialized multiple times!")
```
AffectedArc07, line 49: `stack_trace()` does not halt execution, dont know if this is intended or not

code/game/turfs/simulated/floor.dm
```diff
@@ -32,8 +32,8 @@ GLOBAL_LIST_INIT(icons_to_ignore_at_floor_init, list("damaged1","damaged2","dama
 	var/list/burnt_states = list("floorscorched1", "floorscorched2")
 	var/list/prying_tool_list = list(TOOL_CROWBAR) //What tool/s can we use to pry up the tile?
 
-/turf/simulated/floor/New()
-	..()
+/turf/simulated/floor/Initialize(mapload)
```
AffectedArc07, line 35: *oh lord*

code/game/turfs/turf.dm
Fox-McCloud: Intended; we want want this logged; not halted.

atom initializations are the same.

PR #13692, opened 2020-06-25
Replaces the toSeconds and toMinutes procs with the defines

farie82:
## What Does This PR Do
Reduces proc overhead and removes duplicate code calls.

## Why It's Good For The Game
Should give a bit better performance and gives coders one clear option to use instead of two which do the same but one is worse.

## Changelog
No CL since it's purely backend stuff

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -641,31 +641,31 @@
 					config.max_maint_drones = text2num(value)
 
 				if("expected_round_length")
-					config.expected_round_length = MinutesToTicks(text2num(value))
+					config.expected_round_length = text2num(value) MINUTES
 
 				if("event_custom_start_mundane")
 					var/values = text2numlist(value, ";")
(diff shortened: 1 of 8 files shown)


code/controllers/configuration.dm
```diff
@@ -641,31 +641,31 @@
 					config.max_maint_drones = text2num(value)
 
 				if("expected_round_length")
-					config.expected_round_length = MinutesToTicks(text2num(value))
+					config.expected_round_length = text2num(value) MINUTES
```
AffectedArc07, line 644: `MINUTES` is only `*10`, no tick value is used. You sure you werent meant to use these?

```dm
#define DS2TICKS(DS) ((DS)/world.tick_lag)
#define TICKS2DS(T) ((T) TICKS)
```

farie82: It is used in the context of `world.time`.
I also do not want to change functionality in this PR unless really needed.

AffectedArc07: Oh thats fine then. Reading this code now makes me realise how much utter garbo is here.

PR #13693, opened 2020-06-26
Makes Mobs Lazier and Removes Hard References

Fox-McCloud:
Does a few things, mostly performance/refactor related.

- Makes `stomach_contents` and `processing_patches` lazy lists that only initialize when they're actually needed.
- Stomach contents assumed you could put atoms in it, despite it only ever being mobs---been fixed and made typeless
- Moves stomach handling to the carbon level, to avoid proc call overhead for a bunch of mobs that never even had a stomach
- makes `lastattacker` the attacker's real name instead of a hard reference
- makes a new var `lastattackedckey` which is used in place of having an `:` operator--also avoids more hard references
- removes `lastattacked` because it was unused
- Purges all use of `LAssailant` that are unnecessary and improperly used
  - This breaks the Vox Raider objectives, but `LAssailant` isn't tied into guns and chems and other things anyway, so you can super cheese this---no big loss and of virtually no impact
- putting a mob in disposals properly sets their `LAssailant`
- Removes superhero hitblips because it improperly abused `LAssailant` and requires and `istype` check every single time a human gets damaged
  - Also no one will really miss them
- Fixes dodgeball not really working properly
- Removes some unused mob vars
- clears the `LAssailant` in mob's GCing

mini-rant:

I hold a special disdain for vars like `LAssailant`, where they have to be plugged in at a whole bunch of different locations, but are really only utilized by one thing---slimes---it creates hard references that increase deletion time and makes de-referencing things harder and they're not really possible to clear without doing a search of the entire mob list. For example, to properly clear `LAssailant`, you'd literally have to do this:

```
/mob/Destroy()
	for(var/mob in GLOB.mob_list)
		var/mob/M = mob
		if(M.LAssailant == src)
			M.LAssailant = null
```

This is beyond awful; things like this should utilize soft references, when possible

:cl: Fox McCloud
tweak: clean up to various parts of the code; slightly cuts memory useage and slightly enhances performance related to deleting mobs
del: superheroes no longer have weird hitblips
/:cl:


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -51,41 +50,40 @@
...
-				if(M.client)
-					M.show_message(text("<span class='warning'><B>[user] attacks [src]'s stomach wall with the [I.name]!</span>"), 2)
-			playsound(user.loc, 'sound/effects/attackblob.ogg', 50, 1)
+				for(var/mob/M in viewers(user, null))
+					if(M.client)
+						M.show_message(text("<span class='warning'><B>[user] attacks [src]'s stomach wall with the [I.name]!</span>"), 2)
+				playsound(user.loc, 'sound/effects/attackblob.ogg', 50, 1)
 
-			if(prob(src.getBruteLoss() - 50))
-				for(var/atom/movable/A in stomach_contents)
```
Fox-McCloud, line 85: this is unecessary; when a carbon of human gibs, they're forced out of `stomach_contents` anyway.

PR #13696, opened 2020-06-26
Scoreboard: Score and Rating Cleanup

DaveKorhal:
## What Does This PR Do

1. Commented out score lines that were unimplemented and fixed at 0.
2. Changed the ranges for the crew ratings.

## Why It's Good For The Game

The current scoreboard had a lot of scores that were fixed at 0 and thus actually didn't grade anything, which is disappointing when you busted your ass all shift cooking meals. There were also numerous crew ratings that were almost never seen due to having narrow ranges or requiring absurd score totals; those have been modified so crews have a decent chance of seeing them.

## Changelog
:cl:
tweak: Commented out scoreboard lines that were fixed at 0 and thus don't work yet.
tweak: Changed crew rating ranges so more of them would be seen. (Some had very narrow ranges, others required absurd scores to reach.)
/:cl:

### DIFF
--- a/code/game/gamemodes/scoreboard.dm
+++ b/code/game/gamemodes/scoreboard.dm
@@ -1,5 +1,21 @@
 /datum/controller/subsystem/ticker/proc/scoreboard()
 
+	//Thresholds for Score Ratings
+	#define SINGULARITY_DESERVES_BETTER -3500
+	#define SINGULARITY_FODDER -3000
+	#define ALL_FIRED -2500
+	#define WASTE_OF_OXYGEN -2000
+	#define HEAP_OF_SCUM -1500
+	#define LAB_MONKEYS -1000
+	#define UNDESIREABLES -500
+	#define SERVANTS_OF_SCIENCE 500
+	#define GOOD_BUNCH 1000
+	#define MACHINE_THIRTEEN 1500
+	#define PROMOTIONS_FOR_EVERYONE 2000
+	#define AMBASSADORS_OF_DISCOVERY 3000
+	#define PRIDE_OF_SCIENCE 4000
+	#define NANOTRANSEN_FINEST 5000
+
 	//Print a list of antagonists to the server log
 	var/list/total_antagonists = list()
 	//Look into all mobs in world, dead or alive
@@ -93,7 +109,6 @@
 
 
 	// Bonus Modifiers
-	//var/traitorwins = score_traitorswon
 	var/deathpoints = GLOB.score_deadcrew * 25 //done
 	var/researchpoints = GLOB.score_researchdone * 30
 	var/eventpoints = GLOB.score_eventsendured * 50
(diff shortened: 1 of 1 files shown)


DaveKorhal:
Deleted the commented-out code, also put in local vars for the score ratings so it's easier to change them without accidentally overlapping the ranges.

code/game/gamemodes/scoreboard.dm
```diff
@@ -121,13 +136,9 @@
 		GLOB.score_crewscore += 2500
 		GLOB.score_powerbonus = 1
 
-	if(GLOB.score_mess == 0)
-		GLOB.score_crewscore += 3000
-		GLOB.score_messbonus = 1
```
SteelSlayer, line 126: Correct me if I'm wrong but It looks like this is still used. Line 190:
```
<b>Ultra-Clean Station:</b> [GLOB.score_mess ? "No" : "Yes"] ([GLOB.score_messbonus * 3000] Points)<br><br>
```

code/game/gamemodes/scoreboard.dm
```diff
@@ -177,26 +188,19 @@
 
 	dat += {"
 	<b><u>General Statistics</u></b><br>
-	<u>The Good:</u><br>
-
-	<b>Useful Items Shipped:</b> [GLOB.score_stuffshipped] ([GLOB.score_stuffshipped * 5] Points)<br>
-	<b>Hydroponics Harvests:</b> [GLOB.score_stuffharvested] ([GLOB.score_stuffharvested * 5] Points)<br>
-	<b>Ore Mined:</b> [GLOB.score_oremined] ([GLOB.score_oremined * 2] Points)<br>
```
SteelSlayer, line 184: This is also tracked. In `ore_coins.dm`, `GLOB.score_oremined` is incremented when a new ore is spawned.
```js
/obj/item/stack/ore/New(loc, new_amount, merge = TRUE)
	..()
	pixel_x = rand(0, 16) - 8
	pixel_y = rand(0, 8) - 8
	if(is_mining_level(z))
		GLOB.score_oremined++ //When ore spawns, increment score.  Only include ore spawned on mining level (No Clown Planet)
```

code/game/gamemodes/scoreboard.dm
DaveKorhal: It is. I removed it from the window text because that bonus only applies if there's NO blood/gibs on the station. At all. Which will occur sometime after the heat death of the universe. I'm hesitant to remove the actual calculation just in case it breaks something else, or I decide to use it again later. Suppose I should just remove the code for now, though.

code/game/gamemodes/scoreboard.dm
DaveKorhal: I know, hence why it's still in there.

SteelSlayer: Oop yeah thats my mistake, didn't pay attention. It wasn't removed even though it was in the red section.

code/game/gamemodes/scoreboard.dm
DaveKorhal: Removed it from the code given it has a .000001% chance of actually happening.

Issue #13686, opened 2020-06-25
Melee Attacks in Space drift your Sprite Off-screen


DaveKorhal:
**Issue Description**:

Doing multiple melee attacks from space will slowly drift your sprite up, until it goes off the edge of the screen.

**What did you expect to happen**: 

My sprite would stay in the same place.

**What happened instead**:

Instead it slowly drifted upward until it looked like it was on a different tile, and kept drifting further upward with each attack until it drifted off the top of the screen.

**Why is this bad/What are the consequences**:

**Steps to reproduce the problem**:

1. Get Paramedic hardsuit and Dsword.
2. Go into space.
3. Repeatedly attack windows from space over and over again.

**When did the problem start happening**:

**Extra information**:

Getting shocked or lying down resets the position of your sprite to where it's supposed to be.

This is a screenshot of me appearing to be 2 tiles above where I actually am.
![image](https://user-images.githubusercontent.com/9967634/85664673-1f92b980-b680-11ea-8d39-2260d673bfda.png)

DaveKorhal:
Did some more testing:
-It also resets your position if you re-enter the station.
-Other crew can also see the drift. (One commented I looked like I was in the Cargo Bay when I was just outside.)

farie82:
Probably the same issue as #9119

Issue #9119, opened 2018-06-15
Rubber ducky sprte raise bug.


IIWOLFII:
**Problem Description**:
When flying in space and constantly hitting yourself with rubber ducky, your sprite moves slightly higher than your actual character's position.

**What did you expect to happen**: 
To squack myself to oblivion.

**What happened instead**:
Literally squacked myself to oblivion.

**Why is this bad/What are the consequences:**
Prolly gamebreaking a little.

**Steps to reproduce the problem**:
0. Use hand teleporter to get into open space. (Probably not the cause)
1. Fly to north/south in space, while constantly hitting yourself with captain's rubber ducky.
2. See your sprite raise up.
3. Panic.

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was
anything else out of the ordinary happening?)**: I was playing vox that time. If you hit anything, or do flip and fall down, sprite goes back to normal. Pretty much any sprite update, i guess. It also resets the moment you touch floor or plating.
![image](https://user-images.githubusercontent.com/11389033/41465673-c63b29aa-70a7-11e8-997a-e69845f9785b.png)
Red circle is voxxy's actual position.

tigercat2000:
Ugh, this is a result of matrix modifications stacking upon one another. It's really not trivial to fix, especially with how buggy space floating already is.

IIWOLFII:
Fair enough

MINIMAN10000:
"To squack myself to oblivion." "Literally squacked myself to oblivion." Nice working as you expected :P

PR #13687, opened 2020-06-25
Fix mob sprite drift when doing anims while floating

dearmochi:
## What Does This PR Do

- Fixes most (if not all) cases of sprite drifting on floating mobs (no gravity/in space) that perform an animation (attacking, jitter). The fix works by making the floating animation (among others) use the `ANIMATION_RELATIVE` flag so that it doesn't use outdated values (fixes #9119, fixes #13686). We should be using that flag where possible anyway.
- Minor code clean-up, improve `mob/living/proc/float` readability

## Why It's Good For The Game

Prevents mob sprites from drifting away, providing a clear view on their actual location. That glitch may be used maliciously to harm people without them being able to do anything

## Changelog
:cl:
fix: Fix cases of sprite drifting while floating
/:cl:

PR #13703, opened 2020-06-27
TG Flying+Floating Refactor

Fox-McCloud:
Tested this out, and I can't get the pixel Y offsets to stack, so I believe this...

fixes: https://github.com/ParadiseSS13/Paradise/issues/13686
fixes: https://github.com/ParadiseSS13/Paradise/issues/9119

Alternative to: https://github.com/ParadiseSS13/Paradise/pull/13687

This is from TG, where they refactored flying and floating (mild tweaks) for fairly universal behavior and to ensure this kind of problem doesn't occur.

This also ensures that a lot of things that are flying are immune to typical things that you'd expect. and ensures floating does the same (in most, but not all) cases.

:cl: Fox McCloud
fix: Fixes being able to bap things and move yourself into oblivion
fix: Guardians that go over lava or chasms no longer make you super dead
tweak: flying/floating now more consistently makes you immune to things you'd expect to not trigger if you were in the air.
/:cl:


code/modules/mob/living/life.dm
```diff
@@ -2,9 +2,8 @@
 	set waitfor = FALSE
 	set invisibility = 0
 
-	if(flying) //TODO: Better floating
-		animate(src, pixel_y = pixel_y + 5 , time = 10, loop = 1, easing = SINE_EASING)
-		animate(pixel_y = pixel_y - 5, time = 10, loop = 1, easing = SINE_EASING)
+	if(flying && !floating) //TODO: Better floating
+		float(on = TRUE)
```
AffectedArc07, line 6: If `on` is the only argument here, why not just use `float(TRUE)`

code/modules/mob/living/simple_animal/hostile/megafauna/colossus.dm
```diff
@@ -241,6 +241,14 @@ Difficulty: Very Hard
 		AT.pixel_y += random_y
 	return ..()
 
+/mob/living/simple_animal/hostile/megafauna/colossus/float(on) //we don't want this guy to float, messes up his animations
```
AffectedArc07, line 244: If we dont want the colossus to float, why not just override this to `return` with nothing else, or am I missing something.

code/modules/mob/living/living.dm
```diff
@@ -746,16 +746,17 @@
...
 		return
-	var/fixed = 0
+	var/fixed = FALSE
 	if(anchored || (buckled && buckled.anchored))
-		fixed = 1
+		fixed = TRUE
 	if(on && !floating && !fixed)
 		animate(src, pixel_y = pixel_y + 2, time = 10, loop = -1)
-		floating = 1
+		sleep(10)
```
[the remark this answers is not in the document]
Fox-McCloud: Tested this; not really a problem because of how floating conditionally handles things.

Going into space with magboots on means you don't float--having the magboots off means you do float.

Likewise, if you're a flying mob and toggle magboots on or off you'll still float---in space or not.

code/modules/mob/living/simple_animal/hostile/megafauna/colossus.dm
Fox-McCloud: Done for recording keeping purposes regarding `floating`.

AffectedArc07 (approved):

PR #13708, opened 2020-06-27
Access tweaks to employment consoles

Gatchapod:
## What Does This PR Do
Tweaks accesses to employment records consoles, so they are actually working as intended.
They all now share the access with the offices they're located in (HoP, NT Representative and Blueshield).
Previously, the ones in HoP and NTR office used combined accesses of HoP office and bridge.
The one in BS office had no restrictions whatsoever. You could log in using a blank ID.

Delta and Meta have no employment records consoles in HoP offices, which is why on those maps only NTR and BS consoles were touched.

## Why It's Good For The Game
It's a relatively minor thing I have found annoying when playing NTR. You have the console in your office, but you cannot access it. Noticed Blueshield one has no restrictions along the way, so changed that, too.
In practice, most of characters lack employment records, so it doesn't affect the game that much.

They also now all use req_access_txt to define their accesses, which seems to be the intended way of handling things.

## Changelog
:cl:
tweak: Employment records consoles now share ID access with the offices they're located in
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -59350,6 +59350,7 @@
 "cce" = (
 /obj/structure/table/wood,
 /obj/machinery/computer/skills{
+	req_access_txt = "67";
 	req_one_access = null
 	},
 /turf/simulated/floor/plasteel{
@@ -60291,7 +60292,8 @@
 "cdK" = (
 /obj/structure/table/wood,
 /obj/machinery/computer/skills{
-	req_access_txt = "57"
+	req_access_txt = "73";
+	req_one_access = null
 	},
 /turf/simulated/floor/carpet,
 /area/ntrep)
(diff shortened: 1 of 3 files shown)


Fox-McCloud:
These should probably be hardcoded subtypes that are deliberately mapped in and not icky map edits

Gatchapod:
Well, I followed the logic used by apparently all the consoles on the station. I assume there is a reason why these are not hardcoded, but it might be a wrong assumption.
If you think rewriting how access to the consoles is handled in general is a better idea, I can just close this.
As I've mentioned, it doesn't fix anything actually important.

Fox-McCloud:
Well, it's just that---the old standard *was* to just var edit things on a map.

In this day and age though, it's best to hard code things that you intend to use on the map--with the exception of stuff like pixel placement and access on doors.

Gatchapod:
Well, it is so counter-intuitive it took me a while to realize the access is kept on the map file.
Seeing as that appeared to be the standard, I followed it.
I do agree having it hard-coded into object types would be better.

All things considered, probably not worth to keeping it. I'll close it soon, unless someone feels like it's a good idea to use it. I don't think anyone's working on refactoring the consoles right now?

SteelSlayer:
You can just make a couple new subtypes like `/obj/machinery/computer/skills/nt_rep`, `/obj/machinery/computer/skills/blueshield`, which have the access vars set where the console is defined and not in the `.dmm` file. A then place those subytypes on the map.

PR #13709, opened 2020-06-27
Makes maint drones not have a camera

farie82:
## What Does This PR Do
Maint drones now don't have a camera

## Why It's Good For The Game
AIs or so can use this to find antags and what not all the while drones should not interfere with crew

## Changelog
:cl:
tweak: Maint drones now don't have a camera
/:cl:


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -132,7 +133,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 
 	init()
 
-	if(!scrambledcodes && !camera)
```
TDSSS, line 135: Won't no longer checking for scrambled codes change things? Is scrambled code still needed in other places?

farie82: It is still required in other spots. It was used for syndicate borgs that spawned with this turned on.
It indirectly also meant they had no cameras. I gave them a FALSE on the `has_camera` instead.

TDSSS: In that case, why not just give maint drones scrambled codes? From a quick search it seems its only other use is preventing the borg from being assigned to an AI, it would even let you safe a A.scrambledcodes || istype(A,/mob/living/silicon/robot/drone) in unsorted.dm

farie82: It would also prevent them from being blown up.
scrambledcodes means they are rogue in most of the code where it is used. It was abused in the New to see if it should have a camera or not which in the case of a syndicate borg (only one who has it on newing) should not have one.

Spacemanspark:
I like to set the scrambled codes var to 1 for every borg at round end when I can tell the RD is going to blow them all up

Memes aside, good change imo. Some drones tend to watch things from a distance, including antag activity. If they aren't interfering this isn't a problem, of course, but others really shouldn't be able to peer into it via cameras

AffectedArc07:
```
[2020-06-28T20:53:17] Runtime in _logging.dm,46: Cannot read null.log_game
  proc name: log game (/proc/log_game)
  src: null
  call stack:
  log game("EMP with size (0, 0) in area A...")
  empulse(the floor (222,39,2) (/turf/unsimulated/floor), 0, 0, 1, null)
  EMP Pulse (/datum/chemical_reaction/emp_pulse): on reaction(/datum/reagents (/datum/reagents), 6)
  /datum/reagents (/datum/reagents): handle reactions()
  /datum/reagents (/datum/reagents): add reagent("iron", 3, null, 293.15, 0)
  the mountain blue (/obj/item/reagent_containers/food/pill/random_drugs): New(the pill bottle (???) (/obj/item/storage/pill_bottle/random_drug_bottle))
  the pill bottle (???) (/obj/item/storage/pill_bottle/random_drug_bottle): New(the floor (222,39,2) (/turf/unsimulated/floor))
```

Travis is not your fault. This is also why I wanted to turn storage containers into using `Initialize()` but gave up

farie82:
nice

PR #13710, opened 2020-06-27
Overhaul event mob selection

dearmochi:
## To-do List

- [X] **Subsystem that handles polling for candidates from creation to end**
- [X] **Poll stacking e.g multiple blob infested mice available to control would only show one alert rather than multiple**
- [X] Alert UI
- [X] **Convert all `/proc/pollCandidates` calls to new subsystem proc**
- [X] Port `/proc/pollCandidatesWithVeto`

## What Does This PR Do

- Creates the `ghost_spawns` subsystem which serves as the subsystem for polling candidates for a special role. On the outside it works the same as `/proc/pollCandidates` but unifies everything so that event mob candidacies are a consistent experience.
- When a candidate poll is created through `SSghost_spawns.poll_candidates`, an alert is shown in the top-right corner of the screen to all ghosts with a timer. Clicking that alert signs the player up as a candidate

## Why It's Good For The Game

- Signing up for an event mob (blob, abductors, ERT...) is more consistent and doesn't require the player to interact with elements that vary in location. Everything is done through the top-right corner alerts.
- Alerts display a timer as well as the number of open positions for a specific role, making it more obvious how long a poll lasts

## Images of changes

image -> image -> image
_Alert gets a white outline when the player signs up for the poll. Time text turns red when there are less than 10 seconds left_

ezgif-6-2b993fcdc2d2
_Poll stacking in action - polls stack if they have the same question and role. Clicking an alert with stacked polls signs the player up for all of them_

## Changelog
:cl:
tweak: Overhaul event mob selection to make it more consistent, more intuitive and less frustrating
/:cl:


TDSSS:
IIRC the reason some roles are first come first serve is because they quickly need a controller. Like if a terror spider spawns and doesn't do anything for 30 seconds because the poll is running, it might very well die before a player is picked. Does your system address that?

SteelSlayer:
That can be changed with a bit more code, unless its simply not wanted at all that terrors could be polled. For example, poll ghosts X number of seconds before a spiderling is about to turn into a terror, and then the ghost can be put in control of the spider instantly.

dearmochi:
My main pet peeve with terror spider selection was that the only way (if I remember correctly) to take control is to click on a link in the chat. The problem with that is that you can easily miss the line and by the time you click, someone would have already clicked.
My plan was to tie the player selection to use the new subsystem which makes all polls use the same UI element (an alert) at a consistent location (top-right corner of viewport). 
The actual selection mechanic can be further discussed but @SteelSlayer is a great solution to that problem IMO - poll for candidates right before the eggs hatch rather than *when* they hatch

Kyep:
Steelslayer:
I don't want terrors to poll at all. Terrors should always be first-come, first-served.

dearmochi:
The chat link is actually just a backup. They generate an alert for ghosts in the top right hand corner of the screen, along with a picture of the spider. That's the main method to control them - not the link in chat.

TDSSS:
I feel like we've had this discussion several times, but I'm still not convinced Terror Spiders should be first come first serve as long as the issues with polling can be solved. It is ultimately a very unfair system.

McRamon:
With terror spiders polling may happen for example 30 seconds before spiderling transforms (watching spiderling instead of game just to play it is weird design)

AffectedArc07:
You can make proper tick boxes in your PR description

[ ] unticked
[x] ticked

```
[ ] unticked
[x] ticked
```

dearmochi:
@Kyep 

Ultimately I consider the goal of this PR to streamline (and improve) the polling process both visually and internally. However I'd like to address the points you've made to @SteelSlayer :

1. Everyone has different reaction times, different PC specs and generally factors that may lead to them not being mechanically able to assume control of a spider in time. I'm talking about a <1s time frame to find AND click the right button. **I think it's bad design to make observers constantly be on the lookout for a new spawn when they could be watching the game or chatting with other observers. This only leads to more frustration when they end up not getting a role they'd be interested in.**
Additionally, there's the concern of role hogging. Someone may constantly be the fastest to click thus getting the role but also the quickest to die. Back to observing they go and the cycle continues, denying other players of the role, which may have fared better if they'd got it.

2. This point is moot if we poll before the egg hatches.

3. Not possible with the new subsystem as candidacies are trimmed upon completion, removing mobs deemed invalid.

4. The new subsystem allows for poll stacking. This means only one alert shows up for possibly multiple polls for the same role. Sign up for one role and any subsequent polls for an identical role will sign you up automatically.

5. This is a valid point. Poll stacking means you can't choose (as it is) the specific poll you can sign up for, but that problem can easily be fixed by adding the choice of displaying buttons for individual polls next to the main alert. Hovering one such button would show the spawn location and clicking would sign the player up for that poll specifically.

code/_onclick/hud/alert.dm
```diff
@@ -59,14 +59,18 @@
...
 
-	if(alert.timeout)
-		spawn(alert.timeout)
+	if(!no_anim)
+		alert.transform = matrix(32, 6, MATRIX_TRANSLATE)
+		animate(alert, transform = matrix(), time = 2.5, easing = CUBIC_EASING)
+
+	var/timeout = timeout_override || alert.timeout
+	if(timeout)
+		spawn(timeout)
```
farie82, line 69: `addtimer` should be used here

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
...
+	if(!next_poll_to_finish || poll_time < next_poll_to_finish.time_left())
+		next_poll_to_finish = P
+
+	var/category = "[P.hash]_notify_action"
+
+	for(var/mob/dead/observer/M in (ignore_respawnability ? GLOB.player_list : GLOB.respawnable_list))
+		if(!is_eligible(M))
+			continue
+
+		M << 'sound/misc/notice2.ogg'
```
farie82, line 46: `SEND_SOUND` should be used here instead.

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
...
+			I.layer = FLOAT_LAYER
+			I.plane = FLOAT_PLANE
+			A.overlays += I
+
+		// Start processing it so it updates visually the timer
+		START_PROCESSING(SSprocessing, A)
+		A.process()
+
+	// Sleep until the time is up
+	UNTIL(P.finished)
```
farie82, line 113: Have to be careful with this. Since if this proc is not called from a proc that has set `waitfor` to `TRUE/1` then it will hang the game.

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
...
+			return
+	if(check_antaghud && cannotPossess(M))
+		return
+
+	return TRUE
+
+// Called when polling is finished for a /datum/candidate_poll
+/datum/controller/subsystem/ghost_spawns/proc/polling_finished(datum/candidate_poll/P)
+	// Trim players who aren't eligible anymore
+	var/len_pre_trim = P.signed_up.len
```
farie82, line 141: `length` instead of `.len`. Applies for other places as well. `length` is a bit faster and less runtime prone since it also catches null exceptions

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
...
+	if(!istype(M) || !M.key || !M.client)
+		return
+	if(M in signed_up)
+		if(!silent)
+			to_chat(M, "<span class='warning'>You have already signed up for this!</span>")
+		return
+	if(time_left() <= 0)
+		if(!silent)
+			to_chat(M, "<span class='danger'>Sorry, you were too late for the consideration!</span>")
+			M << 'sound/machines/buzz-sigh.ogg'
```
farie82, line 193: `SEND_SOUND`

code/controllers/subsystem/ghost_spawns.dm
dearmochi: Gotcha, I'll add a warning above the define and I'll make sure any calls to the proc don't have `waitfor` set to `FALSE`/`0`

code/_onclick/hud/alert.dm
```diff
@@ -515,20 +548,47 @@ so as to remain in compliance with the most up-to-date laws."
...
+	var/image/I = image('icons/mob/screen_gen.dmi', icon_state = "selector")
+	I.layer = FLOAT_LAYER
+	I.plane = FLOAT_PLANE + 2
+	overlays += I
+
+/obj/screen/alert/notify_action/proc/display_stacks(stacks = 1)
+	if(stacks <= 1)
+		return
+
+	var/obj/O = new
```
dearmochi, line 580: Forgot to `qdel` this. It'll be fixed next commit

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
+SUBSYSTEM_DEF(ghost_spawns)
+	name = "Ghost Spawns"
+	init_order = INIT_ORDER_EVENTS
+	flags = SS_BACKGROUND
+	wait = 1 SECONDS
+	runlevels = RUNLEVEL_GAME
+	offline_implications = "Ghosts will no longer be able to respawn as event mobs (Terror, Blob, etc..). Shuttle call recommended."
```
AffectedArc07, line 7: ❤️ 

Someone actually put the offline implications in

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
+SUBSYSTEM_DEF(ghost_spawns)
+	name = "Ghost Spawns"
+	init_order = INIT_ORDER_EVENTS
```
AffectedArc07, line 3: Use the default init order

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
...
+	if(!polls_active)
+		return
+
+	for(var/poll in currently_polling)
+		var/datum/candidate_poll/P = poll
+		if(P.time_left() <= 0)
+			polling_finished(P)
+
+/datum/controller/subsystem/ghost_spawns/Initialize(start_timeofday)
+	currently_polling = list()
```
AffectedArc07, line 24: Define this as a list in the subsystem definition, or make it a lazy list and remove the need for an Initialize() proc

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,216 @@
...
+	wait = 1 SECONDS
+	runlevels = RUNLEVEL_GAME
+	offline_implications = "Ghosts will no longer be able to respawn as event mobs (Terror, Blob, etc..). Shuttle call recommended."
+
+	var/list/datum/candidate_poll/currently_polling // List of polls currently ongoing, to be checked on next fire()
+	var/polls_active = FALSE // Whether there are active polls or not
+	var/total_polls = 0 // Number of polls performed since the start
+	var/datum/candidate_poll/next_poll_to_finish // The poll that's closest to finishing
+
+/datum/controller/subsystem/ghost_spawns/fire()
```
AffectedArc07, line 14: Minor styling, but `fire()` usually goes after `Initialize()` in the file, and if you had a recover, would go at the very bottom of the file

code/modules/events/blob.dm
```diff
@@ -12,7 +12,7 @@
 	if(!T)
 		return kill()
 
-	var/list/candidates = pollCandidates("Do you want to play as a blob infested mouse?", ROLE_BLOB, 1)
+	var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a blob infested mouse?", ROLE_BLOB, TRUE, source = /mob/living/simple_animal/mouse/blobinfected)
```
AffectedArc07, line 15: Any reason youre not just doing `source=src` here?

dearmochi: `source` here refers to the mob prototype/instance to display as an icon in the alert panel. Passing an (event) datum makes no sense because it has no icon

AffectedArc07: Ah I see

dearmochi:
@Kyep I'd like to (re-)direct your attention the comment I've written in response to the points you've made to SteelSlayer. I believe that removing the first-come first-served basis on terror spider assignment would alleviate a lot of frustration that ghosts may have during a terror spider event for virtually no downsides. One could argue that the RNG element of polling is a negative, but it's precisely intended to make selection fairer and not just restrict it to the fastest ones. 

**I haven't made anything related to that change as of yet - terror spider selection still operates the way it did before. All I would like is a definite yes or no from you based on my arguments before deciding if it should be part of the PR's scope.**

ZomgPonies:
I know I had a voicechat discussion with Kyep regarding terror spiders. Pretty much I think we agreed that as long as (astute) players have a couple of second to manually click on a new spider to get into it before the icon for the polling shows up, and as long as the polling differentiates between the different spider types then it should be okay.

@Kyep  - Feel free to correct me, it was pretty late when we talked.

code/controllers/subsystem/ghost_spawns.dm
```diff
@@ -0,0 +1,268 @@
+SUBSYSTEM_DEF(ghost_spawns)
+	name = "Ghost Spawns"
+	init_order = INIT_ORDER_EVENTS
+	flags = SS_BACKGROUND
+	wait = 1 SECONDS
+	runlevels = RUNLEVEL_GAME
+	offline_implications = "Ghosts will no longer be able to respawn as event mobs (Terror, Blob, etc..). Shuttle call recommended."
```
Kyep, line 7:
```suggestion
	offline_implications = "Ghosts will no longer be able to respawn as event mobs (Blob, etc..). Shuttle call recommended."
```

code/game/objects/effects/spiders.dm
```diff
@@ -180,7 +180,7 @@
 			if(player_spiders && !selecting_player)
 				selecting_player = 1
 				spawn()
-					var/list/candidates = pollCandidates("Do you want to play as a spider?", ROLE_GSPIDER, 1)
+					var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a spider?", ROLE_GSPIDER, TRUE, source = S)
```
Kyep, line 183:
```suggestion
					var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a giant spider?", ROLE_GSPIDER, TRUE, source = S)
```

code/modules/admin/verbs/gimmick_team.dm
```diff
@@ -52,7 +52,7 @@
 			players_to_spawn += candidate
 	else
 		to_chat(src, "Polling candidates...")
-		players_to_spawn = pollCandidates("Do you want to play as an event character?")
+		players_to_spawn = SSghost_spawns.poll_candidates("Do you want to play as an event character?")
```
Kyep, line 55: If you're changing this anyway, might be good to include the name of the outfit. Not required, just an idea.
E.g: Do you want to play as a Soviet Tourist?
Will end the "what type of event char?!?!" questions in deadchat.

code/modules/events/traders.dm
```diff
@@ -34,7 +34,7 @@ GLOBAL_LIST_INIT(unused_trade_stations, list("sol"))
 	trader_objectives = forge_trader_objectives()
 
 	spawn()
-		var/list/candidates = pollCandidates("Do you want to play as a trader?", ROLE_TRADER, 1)
+		var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a trader?", ROLE_TRADER, TRUE)
```
Kyep, line 37:
```suggestion
		var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a Sol Trader?", ROLE_TRADER, TRUE)
```

Daylight2:
I know I'm a bit late to the party, but I'd like it if the new mob selection system played a sound as well, so alt-tabbed ghosts would have some other indication that roles were available other than the byond icon flashing. 
The lack of sound was fine with the earlier first come first serve one notification per mob system, but if timed requests are going to use it, I'd prefer an audio queue too. (although roles that still use individual pop-ups per mob, like Kyet demanded for terrors, should probably refrain from spamming you with sounds. And while we're at it, could the sound not be the same as the station announcement sound? Thanks.)

code/_onclick/hud/alert.dm
```diff
@@ -515,20 +551,49 @@ so as to remain in compliance with the most up-to-date laws."
...
+	O.maptext = "<span style='font-weight: bold; font-size: 32px; color: yellow;'>[stacks]x</span>"
```
AffectedArc07, line 584: Set the font family here to `Small Fonts` so it renders crisply with no anti-aliasing (The same font is used on status displays and now the ticket machine). Do this in all other places its used.

code/_onclick/hud/alert.dm
```diff
@@ -507,6 +514,35 @@ so as to remain in compliance with the most up-to-date laws."
...
+	if(show_time_left)
+		var/timeleft = timeout - world.time
+		if(timeleft <= 0)
+			return PROCESS_KILL
+
+		if(time_left_overlay)
+			overlays -= time_left_overlay
+
+		var/obj/O = new
+		O.maptext = "<span style='font-weight: bold; font-size: 32px; color: [(timeleft <= 10 SECONDS) ? "red" : "white"];'>[CEILING(timeleft / 10, 1)]</span>"
```
AffectedArc07, line 531: Here

dearmochi: Done and boy is it c r i s p

image

code/modules/events/sentience.dm
```diff
@@ -4,7 +4,7 @@
 	processing = FALSE //so it won't fire again in next tick
 
 	var/ghostmsg = "Do you want to awaken as a sentient being?"
-	var/list/candidates = pollCandidates(ghostmsg, ROLE_SENTIENT, 1)
+	var/list/candidates = SSghost_spawns.poll_candidates(ghostmsg, ROLE_SENTIENT, TRUE)
```
AffectedArc07, line 7: Is `ghostmsg` only used on this line? If so, remove the var entirely and substitute the text in

AffectedArc07:
One the requested changes are integrated, this can be merged.

code/modules/events/sentience.dm
dearmochi: Donezo

AffectedArc07:
Package repositories are dead, travis failing is not your fault. Remind me in 1 hour to re-run.

PR #13714, opened 2020-06-28
Makes blob splits not spawn a new camera in null space if the core died

farie82:
## What Does This PR Do
Makes split blobs not spawn a new player as blob camera if the blob core already died.
Also did some refactoring of the proc itself.

## Why It's Good For The Game
Removes a way to get players into null space. Fixing quite a few runtimes because of it

## Changelog
:cl:
fix: Split blobs now won't spawn a new player in when their core is already destroyed before the poll ended.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/core.dm
+++ b/code/game/gamemodes/blob/blobs/core.dm
@@ -89,41 +89,40 @@
 	..()
 
 
-/obj/structure/blob/core/proc/create_overmind(var/client/new_overmind, var/override_delay)
+/obj/structure/blob/core/proc/create_overmind(client/new_overmind, override_delay)
 	if(overmind_get_delay > world.time && !override_delay)
 		return
 
-	overmind_get_delay = world.time + 3000 // 5 minutes
+	overmind_get_delay = world.time + 5 MINUTES
 
 	if(overmind)
 		qdel(overmind)
 
+	INVOKE_ASYNC(src, .proc/get_new_overmind, new_overmind)
+
+/obj/structure/blob/core/proc/get_new_overmind(client/new_overmind)
 	var/mob/C = null
 	var/list/candidates = list()
-
-	spawn()
-		if(!new_overmind)
-			if(is_offspring)
-				candidates = pollCandidates("Do you want to play as a blob offspring?", ROLE_BLOB, 1)
-			else
-				candidates = pollCandidates("Do you want to play as a blob?", ROLE_BLOB, 1)
-
-			if(candidates.len)
-				C = pick(candidates)
+	if(!new_overmind)
+		if(is_offspring)
+			candidates = pollCandidates("Do you want to play as a blob offspring?", ROLE_BLOB, 1)
 		else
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -89,41 +89,40 @@
...
-
-		if(C)
-			var/mob/camera/blob/B = new(src.loc)
-			B.key = C.key
-			B.blob_core = src
-			src.overmind = B
-			color = overmind.blob_reagent_datum.color
-			if(B.mind && !B.mind.special_role)
-				B.mind.make_Overmind()
-			spawn(0)
```
farie82, line 124: I don't really get why this was here so I removed it in the new version. Only usage was for the split ability.

code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -89,41 +89,40 @@
...
-				if(is_offspring)
-					B.is_offspring = TRUE
+			candidates = pollCandidates("Do you want to play as a blob?", ROLE_BLOB, 1)
+
+		if(length(candidates))
+			C = pick(candidates)
+	else
+		C = new_overmind
+
+	if(C && !QDELETED(src))
```
farie82, line 117: Fix change here

code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -89,41 +89,40 @@
 	..()
 
 
-/obj/structure/blob/core/proc/create_overmind(var/client/new_overmind, var/override_delay)
+/obj/structure/blob/core/proc/create_overmind(client/new_overmind, override_delay)
 	if(overmind_get_delay > world.time && !override_delay)
 		return
 
-	overmind_get_delay = world.time + 3000 // 5 minutes
+	overmind_get_delay = world.time + 5 MINUTES
```
AffectedArc07, line 96: The styling gods approve

code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -89,41 +89,40 @@
...
 
 	if(overmind)
 		qdel(overmind)
 
+	INVOKE_ASYNC(src, .proc/get_new_overmind, new_overmind)
```
AffectedArc07, line 101: Death to `spawn()`

code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -89,41 +89,40 @@
...
+		if(length(candidates))
+			C = pick(candidates)
+	else
+		C = new_overmind
+
+	if(C && !QDELETED(src))
+		var/mob/camera/blob/B = new(loc)
+		B.key = C.key
+		B.blob_core = src
+		src.overmind = B
```
AffectedArc07, line 121: I think `src` may be implied here, but I aint fully sure.

farie82: Yep it is. Will change!

PR #13715, opened 2020-06-28
Makes combat mech circuitboards much harder to print, adds them to cargo

TDSSS:
## What Does This PR Do
This PR makes is to combat mech circuit boards (gygax, durand, phazon) now require illegals 3 research to print yourself. To compensate, the boards can now be ordered from cargo for 50 points for gygax and durand and 100 points for the phazon.

## Why It's Good For The Game
Mech spamming has shown that simple material cost is not enough to keep mechs in check. It is not uncommon for the station to build 3+ mechs 'just in case', and I have seen up to 5 or more with a motivated roboticist. Combat mechs are a very powerful tool, so they should come at a significant cost.

This also causes some inter-department cooperation, as science (usually) needs cargo's help to get their death mech army going. It also adds a reward for getting and deconstructing rare and valuable syndicate gear.

I originally planned to just remove them entirely from the circuit imprinter, or require a very high/normally unreachable combat level, but making it require syndicate tech instead means that traitors can get the research and build themselves a mech in secret, if they want.

I was conservative on the original costing, it might be best to make them much more expensive in cargo points.

## Changelog
:cl:
add: Combat mech circuit boards to cargo
tweak: Combat mech circuit boards now require rare research to build yourself
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -938,17 +938,41 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 /datum/supply_packs/science/robotics/mecha_ripley
 	name = "Circuit Crate (Ripley APLU)"
 	contains = list(/obj/item/book/manual/ripley_build_and_repair,
-					/obj/item/circuitboard/mecha/ripley/main, //TEMPORARY due to lack of circuitboard printer
-					/obj/item/circuitboard/mecha/ripley/peripherals) //TEMPORARY due to lack of circuitboard printer
+					/obj/item/circuitboard/mecha/ripley/main,
+					/obj/item/circuitboard/mecha/ripley/peripherals)
 	cost = 30
 	containername = "\improper APLU \"Ripley\" circuit crate"
 
 /datum/supply_packs/science/robotics/mecha_odysseus
 	name = "Circuit Crate (Odysseus)"
-	contains = list(/obj/item/circuitboard/mecha/odysseus/peripherals, //TEMPORARY due to lack of circuitboard printer
-					/obj/item/circuitboard/mecha/odysseus/main) //TEMPORARY due to lack of circuitboard printer
+	contains = list(/obj/item/circuitboard/mecha/odysseus/peripherals,
+					/obj/item/circuitboard/mecha/odysseus/main)
 	cost = 25
 	containername = "\improper \"Odysseus\" circuit crate"
+
+/datum/supply_packs/science/robotics/mecha_gygax
+	name = "Circuit Crate (Gygax)"
+	contains = list(/obj/item/circuitboard/mecha/gygax/main,
+					/obj/item/circuitboard/mecha/gygax/peripherals,
+					/obj/item/circuitboard/mecha/gygax/targeting)
+	cost = 50
+	containername = "\improper \"Gygax\" circuit crate"
+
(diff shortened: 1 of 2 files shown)


Qwertytoforty:
Personally, I would like to see a combat mech circuit board added, universal to combat mechs, that robotics starts with 1 or 2, so robotics can build a mech or 2 without having to go through cargo. Otherwise, great change, mechs should be more limited

Fox-McCloud:
This is addressing a more fundamental problem (mineral spam) with a sledge hammer. 

Not a fan of this method of handling mechs. 👎

lolzyking:
Major problem by making it a cargo thing, Exosuit fabs don't require ID's to use, making the combat mech boards orderable is a huge step in allowing cargo to just build their own, by themselves.

Balthusdire:
While mechs could use some cost increase of some sort I think I don't really like this. Feels like treating the symptom and not the disease (mineraaaals).

AffectedArc07:
This is just going to stop people producing mechs, and all the minerals will probably end up turning into bags of holdings full of AEGs, with an amount of them which allows infinite rapid fire.

This is not a solution.

TDSSS:
uh...
@Kyep Did you read the part where the circuitboards can be ordered from cargo, now? Combat mechs can defintiely still be gotten in a normal round.

Kyep:
@TDSSS obviously not.
I'm still very :/ about this though. I feel like it has a good chance of doing nothing at all, as cargo/science typically co-operate and I don't imagine it will be hard for science to get mech boards.
It also still has the issue that it treats all combat mechs equally. I'd love to have different tiers of combat mechs, where some are harder to obtain than others.

TDSSS:
@Kyep Well, Phazon boards cost twice as much as Durand and Gygax boards. It would be fairly trivial to, for example, make Durand boards more expensive than Gygax. If you want, we could also make Gygax boards constructable on station easily while Druands and Phazons must be ordered, if you want the Gygax to be used more often, as a 'light', more easily available mech. And the mechs still keep their material costs, which set them apart.

Dumbdumn5:
Requiring illegals for a station mech makes absolutely no sense to me from a mechanical standpoint because other normal weapons don't require nearly the same tech (i.e. AEGs only take "combat" 5, "magnets"  5, "powerstorage"  5). I could understand if it was more ballistic or syndicate oriented, but typically any NT sanctioned gear doesn't need any illegal components. 

Making them orderable by cargo makes them even less difficult to get as you could presumably get them with even more regularity than their boards could get now, in addition to making it possible for cargo to build mechs entirely independent of science. Since sci relies so much on mining to get their minerals in already, it doesn't even put much of a damper on them to get it from cargo early as a single combat mech board is comparable in cost to Vox life support closets, SM Shards, and riot shotguns. Expensive, but not unreachable by any means (especially considering, if you think about it, pizza costs 60 points. Pizza is more expensive than guns and mech boards). Phazon boards are much more reasonably rare, however, considering their cost is near double the expense of the most expensive items in cargo's supply, but the illegals 3 requirements still doesn't make too much sense. 

In my personal opinion, upping material requirements of certain parts rather than research requirements to nerf mech-spam would be preferable to this solution as, once reached, the material requirements otherwise will only lead to more or less the same issue if not a worse one considering cargo can now order the boards *and* science can build them.

TDSSS:
With two heads against it, it's time to close this down.

Hopefully another way to balance mechs and make them actually a significant investment will be found.

PR #13716, opened 2020-06-28
Alternative mech rebalance PR

TDSSS:
## What Does This PR Do
This is an alternative to my other PR, #13715 , based on some comments I got on there and in the coder_chat on discord.

First, this PR makes it so all three combat mechs (durand, gygax, phazon) now share an Exosuit Weapons Targetting circuitboard, which takes the place of their individual circuitboards. Secondly, said circuitboard requires illegals 3 to create on a circuit printer, but can be ordered from cargo for 50 points. Thirdly, a single such circuitboard starts the round in the RD's locker, allowing them to build a single combat mech for free and giving the RD more control over when/if a mech gets built.

The other circuitboards (main and peripherals) are unaffected and are still required, which addresses the concern that cargo could build their own combat mechs.

## Why It's Good For The Game
Mechs are way too cheap and accessible, it is not uncommon for crew to build 3+ mechs on code green just in case. A powerful weapon like a Durand should be a significant investment of resources. Requiring cargo points and inter-department cooperation ensures that they are only built when needed or when people go out of their way for it. Cargo is unlikely to just order you 5 boards so you can give every sec member their own personal mech, for example.

This version of the PR leaves the first mech free and ensures you still need circuit printer access to get the other boards. As in the first version, illegals 3 means you get to make your own mech, to aid traitors with an RD that refuses to hand over the board, etc.

## Changelog
:cl:
add: A generic mecha weapons targetting circuitboard, orderable from cargo or buildable with illegal research
add: A mecha targetting circuitboard to the RD locker
tweak: Mechas now need a generic mecha targetting circuitboard, which can be gotten from cargo or the RD
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -950,29 +950,11 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	cost = 25
 	containername = "\improper \"Odysseus\" circuit crate"
 
-/datum/supply_packs/science/robotics/mecha_gygax
-	name = "Circuit Crate (Gygax)"
-	contains = list(/obj/item/circuitboard/mecha/gygax/main,
-					/obj/item/circuitboard/mecha/gygax/peripherals,
-					/obj/item/circuitboard/mecha/gygax/targeting)
+/datum/supply_packs/science/robotics/mecha_weapons_control
+	name = "Circuit Crate (Exosuit Weapons)"	//the name has to be short so it doesn't glitch the cargo console
+	contains = list(/obj/item/circuitboard/mecha/generic_targetting)
 	cost = 50
-	containername = "\improper \"Gygax\" circuit crate"
-
-/datum/supply_packs/science/robotics/mecha_durand
-	name = "Circuit Crate (Durand)"
-	contains = list(/obj/item/circuitboard/mecha/durand/main,
-					/obj/item/circuitboard/mecha/durand/peripherals,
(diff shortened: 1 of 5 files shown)


Aidenzack1:
RDs could just de-construct the board to potentially get to illegals 3, and from there they could just print the boards as normal

TDSSS:
The board doesn't give any illegal tech when deconstructed, of course.

Spacemanspark:
A better idea might just be to lockbox the individual weapons that can be attached to the mechs.

TDSSS:
@Kyep 
> the most difficult component to acquire will always be the board
The Phazon requires an anomaly core, boards can just be ordered for some points from cargo, so I don't see how you come to this conclusion. 

More generally, the mechs still have their material costs and other requirements (like needing a phasic scanner, etc.) Honestly, having different resources required for building mechs should lead to more variety and decisionmaking. In some rounds, you might have lots of cargo points but barely any raw material, in others maybe you have shitloads of diamons but no silver, etc. Not to mention anomaly cores

Dumbdumn5:
This runs into a similar issue that #13715  has, where illegals 3 doesn't make too much sense and, after it's completed, cargo can just order more boards while science produces boards with the same material cost. 

In addition, though it would be variable based on materials acquired that particular round, it doesn't offset the core issue, which is that, once tech levels are reached, mech spam becomes a problem, and might end up worsening it.

Again, I would be in favor of upping material costs or something like lock boxing particular weapons or parts to slow things down a bit more. Research levels combined with cargo orders just doesn't really fix that.

TDSSS:
With two heads against it, it's time to close this down.

Hopefully another way to balance mechs and make them actually a significant investment will be found.

PR #13718, opened 2020-06-28
Fixes use of a no longer existent helper that prevented compilation

Crazylemon64:
travis compilation results being stale on old PRs is stinky


Fox-McCloud:
Nothing to see here folks. Nothing at all. Nope. Nada.

PR #13720, opened 2020-06-28
Add ability to VV an object's armor easily

dearmochi:
## What Does This PR Do

Adds a VV option on all objects called **Modify armor values**. As the name suggests, it allows for editing an item's armor values. Either individually or blanket setting all of them

## Why It's Good For The Game

Makes it less of a pain for admins to modify an object's armor for events and such

## Changelog
:cl:
add: Add option to easily modify an object's armor through VV for admins
/:cl:

### DIFF
--- a/code/datums/datumvars.dm
+++ b/code/datums/datumvars.dm
@@ -819,6 +819,59 @@
 		message_admins("<span class='notice'>[key_name(usr)] has made [A] process normally</span>")
 		return TRUE
 
+	else if(href_list["modifyarmor"])
+		if(!check_rights(R_DEBUG|R_ADMIN))
+			return
+		var/obj/A = locateUID(href_list["modifyarmor"])
+		if(!istype(A))
+			return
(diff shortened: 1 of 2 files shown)


dearmochi:
@Kyep I should've mentioned that was a verbal request from @Fox-McCloud and editing an item's `armor` right now is very inconvenient because multiple items may share the same `armor` datum (caching). This VV creates (or uses a cached) armor datum if the need arises

TDSSS:
Might want to mention it's a port (at least, I think it's a port, if I read coder_chat right), and link to where you're porting from exactly.

dearmochi:
@TDSSS It's not a port as we seem to be missing a [proc or two](https://github.com/tgstation/tgstation/blob/1fdd64532e7e6d7278cd705655bc945fc4bcc309/code/game/objects/objs.dm#L240) and I didn't think there'd be a point to port these as well just for this

PR #13723, opened 2020-06-29
Makes Blood GC

Fox-McCloud:
Did some debugging and figured out that cleaner bots were preventing a lot of blood from properly GCing--if they ever spot some blood that they can't get to, they store it in a list, then...never clean it out---which means that once that blood is deleted, it won't GC.

hard deleting blood is already surprisingly expensive, and this occurs often enough to be an issue.

This ensures that blood should GC the vast bulk of the time.

I was originally going to utilize signals with this, but then I wondered if TG had this same issue; I checked and it turns out they did, but solved it---so I went with their solution as I felt it was better than cleanbots registering signals on all cleanable decals all over the place (with each cleanbot capable of doing so).

:cl: Fox McCloud
tweak: Ensures blood deletion handles better, improving overall server performance
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/bot.dm
+++ b/code/modules/mob/living/simple_animal/bot/bot.dm
@@ -63,6 +63,7 @@
 	var/new_destination		// pending new destination (waiting for beacon response)
 	var/destination			// destination description tag
 	var/next_destination	// the next destination in the patrol route
+	var/ignorelistcleanuptimer = 1 // This ticks up every automated action, at 300 we clean the ignore list
 	var/robot_arm = /obj/item/robot_parts/r_arm
 
 	var/blockcount = 0		//number of times retried a blocked path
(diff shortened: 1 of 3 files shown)


Fox-McCloud:
Hmm, not sure which I like better. 

Doing this:

```
/mob/living/simple_animal/bot/proc/add_to_ignore(atom/A)
    if(ignore_list.len < 50) //This will help keep track of them, so the bot is always trying to reach a blocked spot.
        ignore_list |= A
    else  //If the list is full, insert newest, delete oldest.
        ignore_list.Cut(1, 2)
        ignore_list |= A
    RegisterSignal(A, COMSIG_PARENT_QDELETING .proc/clear_refs, TRUE)

/mob/living/simple_animal/bot/proc/clear_refs(atom/A)
    ignore_list -= A
```

Also works, but it means that every cleanbot has a list of a signal for every object it can't reach and every object it can't reach has a list of cleanbots its has a signal registered with.

20 cleanbots that can't reach 20 objects means a list total length of 800; versus the method in this PR which would keep a total list length of 400, but would have the additional proc call overhead (every 10 minutes) of clearing out the its own lists (is it even necessary to clean up the list? Though I could imagine a weird scenario where 50 cleanbots can't reach 100 gibs and the list operations get so lengthy it becomes problematic?)

Really not sure if it *really* matters at all (or is worth profiling the difference).

PR #13725, opened 2020-06-30
Genetics DNA Modifier Fixes

Kyep:
## What Does This PR Do
1) Fixes this runtime:
[2020-06-29T23:37:06] Runtime in ,: DEBUG: to_chat called with invalid message/target.
   Message: '<span class='warning'>The scanner is locked!</span>'
   Target: ''
This runtime is caused by dna_scannernew/proc/go_out() assuming that 'usr' is defined when it often isn't.
To fix it, the code is altered so it only attempts to to_chat(usr, ...) if usr exists.

2) Fixes bodies inside a locked DNA scanner not being ejected by Destroy() on the scanner
This was caused by the fact that go_out returns early if the dna scanner is locked... even when go_out is invoked via Destroy() -> eject_occupant() -> go_out(). 
To fix this one, a force param is added to the procs, so that when they're invoked from Destroy() the person is ejected even if the scanner is locked.

## Why It's Good For The Game
1) fixes a common runtime error
2) fixes an oversight

## Changelog
:cl: Kyep
fix: fixed locked DNA scanners in genetics generating runtimes and some other weird behavior
/:cl:

### DIFF
--- a/code/game/dna/dna_modifier.dm
+++ b/code/game/dna/dna_modifier.dm
@@ -125,7 +125,7 @@
 	eject_occupant(null, TRUE)
 	return ..()
 
-/obj/machinery/dna_scannernew/proc/eject_occupant(user = null, force = FALSE)
+/obj/machinery/dna_scannernew/proc/eject_occupant(user, force)
 	go_out(user, force)
 	for(var/obj/O in src)
 		if(!istype(O,/obj/item/circuitboard/clonescanner) && \
@@ -274,11 +274,10 @@
 		if(user)
 			to_chat(user, "<span class='warning'>The scanner is empty!</span>")
 		return
-	if(locked)
+	if(locked && !force)
 		if(user)
 			to_chat(user, "<span class='warning'>The scanner is locked!</span>")
-		if(!force)
-			return
+		return
 	occupant.forceMove(loc)
 	occupant = null
 	icon_state = "scanner_open"
--- a/code/game/dna/dna_modifier.dm
+++ b/code/game/dna/dna_modifier.dm
@@ -118,17 +118,15 @@
 
 	if(usr.incapacitated())
 		return
-
-	eject_occupant()
-
+	eject_occupant(usr)
 	add_fingerprint(usr)
 
 /obj/machinery/dna_scannernew/Destroy()
-	eject_occupant(TRUE)
+	eject_occupant(null, TRUE)
 	return ..()
 
-/obj/machinery/dna_scannernew/proc/eject_occupant(force)
-	go_out(force)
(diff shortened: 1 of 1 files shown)


code/game/dna/dna_modifier.dm
```diff
@@ -271,15 +271,16 @@
 
 		occupant.notify_ghost_cloning(source = src)
 
-/obj/machinery/dna_scannernew/proc/go_out()
+/obj/machinery/dna_scannernew/proc/go_out(force)
```
Fox-McCloud, line 274: pass the user arg so we're not using usr

Kyep: Updated. Please verify that it works the way you want.

code/game/dna/dna_modifier.dm
```diff
@@ -271,15 +269,16 @@
...
 
-/obj/machinery/dna_scannernew/proc/go_out()
+/obj/machinery/dna_scannernew/proc/go_out(mob/user, force)
 	if(!occupant)
-		to_chat(usr, "<span class='warning'>The scanner is empty!</span>")
+		if(user)
+			to_chat(user, "<span class='warning'>The scanner is empty!</span>")
 		return
-
 	if(locked)
```
Fox-McCloud, line 277: `if(locked && !force)`

code/game/dna/dna_modifier.dm
```diff
@@ -118,17 +118,15 @@
...
 	add_fingerprint(usr)
 
 /obj/machinery/dna_scannernew/Destroy()
-	eject_occupant()
+	eject_occupant(null, TRUE)
 	return ..()
 
-/obj/machinery/dna_scannernew/proc/eject_occupant()
-	go_out()
+/obj/machinery/dna_scannernew/proc/eject_occupant(user = null, force = FALSE)
```
Fox-McCloud, line 128: typecast

code/game/dna/dna_modifier.dm
Kyep: Changed

code/game/dna/dna_modifier.dm
Kyep: Changed

code/game/dna/dna_modifier.dm
```diff
@@ -751,7 +749,7 @@
 		return TRUE
 
 	if(href_list["ejectOccupant"])
-		connected.eject_occupant()
+		connected.eject_occupant(usr)
```
AffectedArc07, line 752:
```suggestion
		if(connected.locked) // Tell the person who clicked the UI that its locked if they try to eject while locked
				connected.eject_occupant(usr)
		else
				to_chat(usr, "<span class='warning'>[src] is locked!</span>")
```

PR #13730, opened 2020-06-30
Removes isSynthetic() helper, replaces its instances with ismachineperson()

Gatchapod:
## What Does This PR Do
Removes `isSynthetic()` helper and replaces all instances of it with `ismachine()`.
Why? Because `isSynthetic()` is literally just calling for `ismachine()` and returns the same value.
This interchangeability leads to inconsistent usage of the two across the code.

I have considered making `isSynthetic()` a helper that would return true value for IPCs and silicon mobs, but the only benefit I could find in that was allowing cyborgs to slur via variable editing.

Additionally, as Fox correctly pointed out, `ismachine()` and `ismachinery()` are extremely similar and possibly confusing. Following their request, I changed the former to `ismachineperson()`.

## Why It's Good For The Game
More consistent code is good, right?

## Changelog
:cl:
del: isSynthetic() helper
tweak: ismachineperson() is now used instead of isSynthetic()
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -196,7 +196,7 @@
 #define isslimeperson(A) (is_species(A, /datum/species/slime))
 #define isgrey(A) (is_species(A, /datum/species/grey))
 #define isdiona(A) (is_species(A, /datum/species/diona))
-#define ismachine(A) (is_species(A, /datum/species/machine))
+#define ismachineperson(A) (is_species(A, /datum/species/machine))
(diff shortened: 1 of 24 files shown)


Fox-McCloud:
Do me a big favor and make `ismachine` something like `ismachineperson`.

`ismachine` is probably going to invariably confuse someone into thinking that it means it's an `/obj/machinery` subtype.

KasparoVy:
`isrobutt()`

PR #13731, opened 2020-06-30
Fixes TGUI aghost interactions + disposals

AffectedArc07:
## What Does This PR Do
This PR makes it so admins ghosting can interact with any TGUI, and also fixes redundant checks in the disposal bin TGUI.

## Why It's Good For The Game
Admins should have abilities.

## Changelog
:cl: AffectedArc07
fix: Ghosting admins can now interact with TGUIs
/:cl:

### DIFF
--- a/code/modules/recycling/disposal.dm
+++ b/code/modules/recycling/disposal.dm
@@ -271,7 +271,7 @@
 
 /obj/machinery/disposal/tgui_data(mob/user)
 	var/list/data = list()
-  
+
 	data["isAI"] = isAI(user)
 	data["flushing"] = flush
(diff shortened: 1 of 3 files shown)


code/modules/recycling/disposal.dm
```diff
@@ -290,15 +290,12 @@
...
 
-	if(..())
-		return
-
 	if(stat & BROKEN)
 		return
 
 	src.add_fingerprint(usr)
 
-	if(usr.stat || usr.restrained() || src.flushing)
```
AffectedArc07, line 301: This check is already handled by the TGUI state, and was redundant here

PR #13732, opened 2020-06-30
[TGUI] Slot Machines - Example PR for NanoUI --> TGUI conversions

AffectedArc07:
## What Does This PR Do
This PR converts slot machines from NanoUI to TGUI, and also serves as an example for how to go about converting a UI, as I will document the code explaining what does what.

Idle:
image

Spinning:
image

Spun:
image

No ID:
image

## Why It's Good For The Game
Killing off nanoUI is good

## Changelog
:cl: AffectedArc07
add: Slot machines now use TGUI
/:cl:


code/game/machinery/slotmachine.dm
```diff
@@ -12,90 +12,94 @@
 	var/resultlvl = null
 
 /obj/machinery/slot_machine/attack_hand(mob/user as mob)
+	tgui_interact(user)
```
AffectedArc07, line 15: All `ui_` procs have been replaced with `tgui_` procs for the sake of separation. This should be called from `attack_hand()` or `attack_ai`, and **must** pass `user` as an argument.

code/game/machinery/slotmachine.dm
```diff
@@ -12,90 +12,94 @@
...
+
+/obj/machinery/slot_machine/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = 0, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
+	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
+	if(!ui)
+		ui = new(user, src, ui_key, "SlotMachine", name, 350, 200, master_ui, state)
```
AffectedArc07, line 20: Here is where you define the UI. Take note of `SlotMachine`, as that is what the JS file is named, and more importantly, the export inside the file. You can also pass through a custom name or state here, as well as the UI size.

code/game/machinery/slotmachine.dm
```diff
@@ -12,90 +12,94 @@
...
+		ui.open()
+		ui.set_autoupdate(TRUE)
+
+/obj/machinery/slot_machine/tgui_data(mob/user)
```
AffectedArc07, line 24: `ui_data` is replaced with `tgui_data` for separation. In here is where the data for the UI is collected from DM and sent to the JS handler. I moved the code around here to make ID grabbing more responsive, but in most cases this isnt needed.

code/game/machinery/slotmachine.dm
```diff
@@ -12,90 +12,94 @@
...
+	// Send data
 	data["working"] = working
 	data["money"] = account ? account.money : null
 	data["plays"] = plays
 	data["result"] = result
 	data["resultlvl"] = resultlvl
 	return data
 
-/obj/machinery/slot_machine/Topic(href, href_list)
+/obj/machinery/slot_machine/tgui_act(action, params)
```
AffectedArc07, line 42: `tgui_act()` is a full replacement for `Topic`, and has a much better way of handling calls, as it seperates `href` and `href_list` out into `action` and `params`. 

**NOTE**:
The following line is *very* important
```
if(..())
    return
```
This MUST be present at the top of any `tgui_act` to avoid a UI being range-exploited.

code/game/machinery/slotmachine.dm
```diff
@@ -12,90 +12,94 @@
...
-			if(!account || account.money < 10)
-				return
-			if(!account.charge(10, null, "Bet", "Slot Machine", "Slot Machine"))
-				return
-			plays++
-			working = 1
-			icon_state = "slots-on"
-			playsound(src.loc, 'sound/machines/ding.ogg', 50, 1)
-			addtimer(CALLBACK(src, .proc/spin_slots, usr.name), 25)
+	if(action == "spin")
```
AffectedArc07, line 47: Here we check for the action, in this case, `"spin"`. In more complicated UIs, this should be a `switch()` statement, but since this UI has one callback its an if statement.

tgui/packages/tgui/interfaces/SlotMachine.js
```diff
@@ -0,0 +1,58 @@
+import { useBackend } from "../backend";
+import { Button, LabeledList, Box, AnimatedNumber } from "../components";
+import { Window } from "../layouts";
+
+export const SlotMachine = (props, context) => {
```
AffectedArc07, line 5: This is the UI export, and what is referenced in the `tgui_interact` proc.

tgui/packages/tgui/interfaces/SlotMachine.js
```diff
@@ -0,0 +1,58 @@
...
+        </Window.Content>
+      </Window>
+    );
+  } else {
+    let playerText;
+    if (data.plays === 1) {
+      playerText = data.plays + " player has tried their luck today!";
+    } else {
+      playerText = data.plays + " players have tried their luck today!";
+    }
```
AffectedArc07, line 28: You can freely create variables inside the UI as you would in JavaScript, allowing for custom messages based off of a value. Here I have used that to make it grammatically correct when only one person has spun the slot machine.

tgui/packages/tgui/interfaces/SlotMachine.js
```diff
@@ -0,0 +1,58 @@
...
+              <LabeledList.Item label="Credits Remaining">
+                <AnimatedNumber
+                  value={data.money}
+                />
+              </LabeledList.Item>
+              <LabeledList.Item label="10 credits to spin">
+                <Button
+                  icon="coins"
+                  disabled={data.working}
+                  content={data.working ? "Spinning..." : "Spin"}
```
AffectedArc07, line 46: You can use `?` operators here as if it were DM or any other language that supports these.

tgui/packages/tgui/interfaces/SlotMachine.js
```diff
@@ -0,0 +1,58 @@
...
+                  onClick={() => act("spin")} />
```
AffectedArc07, line 47: The `onclick` call here is what sends the actual request to the DM side of things, in this case, the `act("spin")` will send the `"spin"` action. To add a parameter to an action, use the following:
`act('track', {track: crewmember.ref})`

nano/templates/slotmachine.tmpl
```diff
@@ -1,34 +0,0 @@
-<!--
```
AffectedArc07, line 1: Oh, and dont forget to delete your nanoUI template

code/game/machinery/slotmachine.dm
```diff
@@ -12,90 +12,94 @@
...
 
 /obj/machinery/slot_machine/attack_hand(mob/user as mob)
+	tgui_interact(user)
+
+/obj/machinery/slot_machine/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = 0, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
+	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
+	if(!ui)
+		ui = new(user, src, ui_key, "SlotMachine", name, 350, 200, master_ui, state)
+		ui.open()
+		ui.set_autoupdate(TRUE)
```
SteelSlayer, line 22: Does the UI need to update this fast? From what I can tell of the comment of `autoupdate`, it updates every MC tick. That seems excessive, but come to think of it, I'm not sure how fast TGUI's update without this var on FALSE.

AffectedArc07: Its every time `SStgui` fires, which is every 0.9 seconds. I think this is required due to the fact the machine needs to update after the spin has completed, which is on a delayed callback. I guess I *could* throw `SStgui.update_uis(src)` into the delayed callback, but this works, and its a very lightweight update.

SteelSlayer: Oh ok, its not that much of a difference then, shouldn't be a big deal.

AffectedArc07: Done it anyway

AffectedArc07:
I know this may sound weird, but I want to go *against* the suggestions for the sake of keeping this PR down to the point. Its designed to be a minimal TGUI conversion PR, without a ton of extra DM changes which may confuse people. I will take maintainer verdict on whether they think I should put these changes in, but I think its just going to cause confusion for what should be a simple PR.

Fox-McCloud:
This is largely a refactor of the UI, @AffectedArc07 isn't obligated to HAVE to correct every single styling error in this file when he updates what he needs to---especially as this is kinda a tutorial PR that is meant to be instructive---having a lot of extraneous changes really detracts from those who want to learn just the basics.

PR #13736, opened 2020-06-30
Headcrab Bug, eating while being dead. Bug, sometimes display wrong name while eating

denghis1:
## What Does This PR Do
This PR fixes a couple bugs with headcrabs, the first one fixes headcrabs being able to zombify/eat while being dead by adding a check for the headcrab itself being dead

During testing we uncovered a small bug where headcrabs while eating would sometimes display the wrong target in chat, this is fixed by changing the displayed target in chat.

## Why It's Good For The Game
Preventing headcrabs from zombifying and eating while being dead is good. 
Displaying the correct information in chat is good as well. 

## Changelog
:cl: Denghis, Power, Toodles
fix: Headcrabs zombifying while being dead
fix: Headcrabs displaying wrong target message while eating
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -25,7 +25,7 @@
 	var/list/human_overlays = list()
 
 /mob/living/simple_animal/hostile/headcrab/Life(seconds, times_fired)
-	if(..())
+	if(..() && src.stat != DEAD)
 		if(!is_zombie && isturf(src.loc))
 			for(var/mob/living/carbon/human/H in oview(src, 1)) //Only for corpse right next to/on same tile
 				if(H.stat == DEAD || (!H.check_death_method() && H.health <= HEALTH_THRESHOLD_DEAD))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -25,7 +25,7 @@
 	var/list/human_overlays = list()
 
 /mob/living/simple_animal/hostile/headcrab/Life(seconds, times_fired)
-	if(..())
+	if(..() && src.stat != DEAD)
```
farie82, line 28: `src` implied. It can be omitted here.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -35,7 +35,7 @@
 		if(cycles >= 4)
 			for(var/mob/living/simple_animal/K in oview(src, 1)) //Only for corpse right next to/on same tile
 				if(K.stat == DEAD || (!K.check_death_method() && K.health <= HEALTH_THRESHOLD_DEAD))
-					visible_message("<span class='danger'>[src] consumes [target] whole!</span>")
+					visible_message("<span class='danger'>[src] consumes [K.name] whole!</span>")
```
farie82, line 38: just `K` will do here. It will auto make a fitting name. Usually it'll be the same as `K.name`

Terilia:
There is a small bug with the cycles... I wanted to overthrow that system anyway and I will fix that in a later PR together with a complete Headcrab overhaul. 

What will be the overhaul? Well here is a hint
image

PR #13744, opened 2020-07-01
Tweak, Added wrench_act to slot machines, so that they can be moved by the crew

denghis1:
## What Does This PR Do
This PR adds wrench_act to the slot machine, this is the same section of code that is used on most machines, making slot machines wrenchable. 
~~Timer is set to 120 so as to make it take a bit of time to complete the wrenching.~~
Timer set to default value.

## Why It's Good For The Game
Allows barkeeps (or others for that matter) to move the slot machines found around the station. 
This also allows the barkeep to redecorate more without having to worry about the static placement of the slot machines.

## Images of changes
Current change:
sUworUAJ2I

Comparison with other wrenchable items:
sxHoskRfsU

## Changelog
:cl: denghis
tweak: Makes slot machines wrenchable
/:cl:

### DIFF
--- a/code/game/machinery/slotmachine.dm
+++ b/code/game/machinery/slotmachine.dm
@@ -24,7 +24,7 @@
 	. = TRUE
 	if(!I.tool_use_check(user, 0))
 		return
-	default_unfasten_wrench(user, I, 120)
+	default_unfasten_wrench(user, I)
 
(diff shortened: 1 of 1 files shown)


AffectedArc07:
I would make the cooldown the same as unwrenching other things (machine frames, computer frames, etc), the current time seems too long

Fox-McCloud:
Awful, horrendous, game breaking, will completely unhinge everything. How could you?!

I'm kidding. Good first PR.

👍

PR #13748, opened 2020-07-01
Purges Rigs

Fox-McCloud:
I'm not really calling this a revert of feature removal because these aren't accessible to players.

They're so wonky, broken, and unsupported, these are pointless to keep around.

Also their NanoUI was purged and converting these to use TGUI is silly for something so untouched.

These are unsupported, untouched, and the code is, frankly, terrifying and snowflakey.

They're not going to be supported, and, as originally suspected, they got added and were never used except during a handful of rounds as adminbus things.

Keeping them around is ill advised---they're not something we're going to use.

Also, did I mention the massive number of sprites they use? It's insane.

:cl: Fox McCloud
del: purges Baystation rigsuits in hellfire
/:cl:


Shockpoint:
> They're not going to be supported, and, as originally suspected, they got added and were never used except during a handful of rounds as adminbus things.

I always thought these were cool, I just never knew how they worked. Is there really no way to support them?

Fox-McCloud:
>I always thought these were cool, I just never knew how they worked. Is there really no way to support them?

They're broken half the time, insanely buggy, and snowflaked all over the place just to "work".

They're from an ecosystem that's extremely divorced from how we code and how we design+balance things.

On Bay, they completely replaced hardsuits with these; we still strongly support hardsuits and probably always will---advanced suits with unique functions are better handled like how Space Ninja, Chrono suit, and stuff like the shielded hardsuit work; not something like this.

taukausanake:
> Space Ninja

Inb4 that never gets work on too and gets removed too. Then again, Fox did try working on it once before, but so were rig suits...

Carthusia:
I dont think i've ever seen a RIG suit?

All the hardsuits look somewhat like a deadspace RIG suit, but there is an actual RIG suit?

Denthamos:
Only a select few I see worth trying to salvage and fix.  The sheer amount of them and the amount of work and spriting that would have to be done for all of them would be immense.  Only one that really sees any semi regular use is the ninja one for when another admin or myself manually runs a ninja event using it.  

As long as TG's ninja gets ported in then that will help fill that void, though I do love the idea of some hardsuits becoming modular/up-gradable with special functions like fox said in comparison to the chrono hardsuit.  I love the thought of the hardsuits on station becoming more of a valuable asset over time in a round with science(or mechanic) being able to make more upgrades revolving around them besides just the xenobio potions or jetpack upgrades from mining.

fludd12:
I think, rather than even attempting to salvage, a complete re-code may be more prudent, maybe using some of this as a guideline. It's not like the code won't exist in git history, after all.

AffectedArc07:
There we so many icon diffs it made IDB run twice
image

PR #13750, opened 2020-07-01
[TGUI] Camera Consoles (Actual Magic)

AffectedArc07:
## What Does This PR Do
This PR moves camera consoles to TGUI. Not only does this make them much more performant, but it allows picture-in-picture style viewing, showing the cameras on the actual UI instead of redirecting your game viewport.
image

Cameras also scale depending on environment size
image
image

You can also search for them now, which removes the need for network selection.

## Why It's Good For The Game
Part of the mass conversions

## Changelog
:cl: AffectedArc07
add: Camera consoles now use TGUI
add: The traitor camera bug now uses a TGUI
/:cl:


code/game/machinery/computer/camera.dm
```diff
@@ -10,75 +10,157 @@
...
-	available_networks["Supermatter"] =       list(ACCESS_CE,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["MiniSat"] =           list(ACCESS_RD,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["Singularity"] =       list(ACCESS_CE,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["Anomaly Isolation"] = list(ACCESS_RD,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["Toxins"] =            list(ACCESS_RD,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["Telepad"] =           list(ACCESS_RD,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["TestChamber"] =       list(ACCESS_RD,ACCESS_HOS,ACCESS_CAPTAIN)
-	available_networks["ERT"] =               list(ACCESS_CENT_SPECOPS_COMMANDER,ACCESS_CENT_COMMANDER)
-	available_networks["CentComm"] =          list(ACCESS_CENT_SECURITY,ACCESS_CENT_COMMANDER)
-	available_networks["Thunderdome"] =       list(ACCESS_CENT_THUNDER,ACCESS_CENT_COMMANDER)
```
AffectedArc07, line 41: All of this has been removed, and networks are now decided by the boards of the subtypes, since having a selection box for networks in this UI size would not work well, and we have a search anyway.

BakKeener:
this looks fantastic and way better, also is it still possible to cycle to nearby cameras with movement controls still?

AffectedArc07:
Not at the moment, but in my experience that feature on the old monitors never even worked well anyway. I can re-implement it if needed, but at the same time, I dont know if its worth it from how badly it worked before.

BakKeener:
yeah the issue is, it wasn't consistent in what camera it went to... I feel like you would be able to make it work better then what we had though. Although I just thought of something, could it be possible to save cameras to a list and be able to cycle them with the movement keys or would that be too powergamey?

Fox-McCloud:
Just demo'd this.

This is literally magic.

It's incredible.

BakKeener:
I just had a crazy but amazing thought today, Camera Bug TGUI instead of the shitty popup list?

AffectedArc07:
@BakKeener that can come in the future after this PR is in. It touches enough files already so I want to keep bloat down.

AffectedArc07:
Gonna wait for alarm rework to be in before I deconflict

AffectedArc07:
The latest update refactors camera bugs. You cannot "bug" a camera anymore, but you have an actual UI you can use with it now, and use it as a handheld camera monitor, which is useful as a traitor, especially since it no longer obscures your main viewport, which means this could be a much nicer tool for traitors than the existing one, and might actually encourage people to start buying them.

dearmochi (approved):
Documentation makes me happy

Issue #12153, opened 2019-08-30
Training bomb does not reset properly after succesful diffusal


datlo:
**Issue Description**:
Security training bomb does not reset its wires after a succesful diffusal.

**What did you expect to happen**: 
Diffuse bomb, 'YOU WIN!', bomb reset

**What happened instead**:
Diffuse bomb, its diffused normally, you have to mess with wires again to reset it and score isnt tracked.

**Why is this bad/What are the consequences**:
Clunky to fix, feature isnt working properly. Also my score isnt tracked properly so I cant flex my bomb diffusing skills.

**Steps to reproduce the problem**:
Diffuse training bomb, see it doesnt reset

**When did the problem start happening**:
Long time ago.

datlo:
This was fixed in mentionned PR

PR #13760, opened 2020-07-03
bugfix syndicate bomb not triggering defuse properly

denghis1:
## What Does This PR Do
This PR makes the syndicate bombs, and more importantly for this fix, the training bombs actually trigger the defuse part of the bomb process() proc. 

It did not trigger since cutting the wire would set b.active to false, and would thus instantly break out of process() before it had a chance to check if the bomb was defused. (see line 49 & 54 in /obj/machinery/syndicatebomb/process() )

In order to make sure that it falls into the defused check in process(), i set active to false when actually processing the defused part, instead of in the wire code.

The removal of !active in the if is due to it being redundant inside process()

Only a few lines changed, but due to the way it flows, i did not see many other options of fixing this while touching as little as possible.

## Why It's Good For The Game

Fixes issue #12153 and makes training bombs actually work as intended, go set a new highscore!

## Images of changes

Alien multitool used in the gifs to make finding the correct wire easier for testing.

Before: 
57ny6v86qL

After: 

L4GR04TvPQ

## Changelog
:cl: denghis
fix: syndicate bomb not triggering defuse properly, preventing training bombs from working as intended
/:cl:

### DIFF
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -79,10 +79,8 @@
 	//Counter terrorists win
 	else if(defused)
 		active = FALSE
-		if(defused && (payload in src))
+		if(payload in src)
 			payload.defuse()
-			countdown.stop()
-			STOP_PROCESSING(SSfastprocess, src)
 
 /obj/machinery/syndicatebomb/New()
 	wires 	= new(src)
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -77,7 +77,8 @@
 		update_icon()
 		try_detonate(TRUE)
 	//Counter terrorists win
-	else if(!active || defused)
+	else if(defused)
+		active = FALSE
 		if(defused && (payload in src))
 			payload.defuse()
 			countdown.stop()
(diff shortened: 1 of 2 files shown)


code/game/machinery/syndicatebomb.dm
```diff
@@ -77,7 +77,8 @@
 		update_icon()
 		try_detonate(TRUE)
 	//Counter terrorists win
-	else if(!active || defused)
+	else if(defused)
+		active = FALSE
 		if(defused && (payload in src))
 			payload.defuse()
 			countdown.stop()
```
farie82, line 84: This one and the line below should be in the scope of the defused if.
It might be nice to make a proc to deactivate the bomb since it's done twice now. Once here and once on line 49-53

denghis1: When discussing in coding chat, some us wondered that if we did merge those two if's, what state would we leave the bomb in, if it did not have a payload? As for making that proc, i might need a little help with that. Alternatively, only keep payload.defuse(), and leave the rest to be done on next tick of process() ?

farie82: Also a possibility to do that yeah. Feel free to ping me if you choose to do it the proc way. The way you suggested also works in this case

denghis1: Chose to cleanup, and let payload and countdown be run on next tick of process instead.

farie82:
Quick note on the fixing of issues. If you type something like: fixes #num_here then it'll auto close the issue once merged. It has to be in that format otherwise github won't recognize it. Other keywords work as well like closes etc

PR #13765, opened 2020-07-04
Ups the experience requirements for Warden/HoS

Mitchs98:
## What Does This PR Do
This PR ups the time requirement for HoS to twenty hours as security and more appropriately makes Warden take ten hours as security rather than unlocking at the same time as every other security role. I DID intend to raise the security timelock itself up, but I was unaware it was actually ten hours as crew. Though, I would argue that it could be higher...it would lead to security being staffed less often so that's not entirely a good thing. 

Also..correct me if I'm wrong, but I assume the numbers on the experience required are in minutes.

## Why It's Good For The Game
Ten hours as security is far from enough experience for a HoS to come to grips on the security role on a whole, there are MANY Heads of Security that don't actually know space law despite those ten hours as an officer. I'd actually be more inclined to raise the requirement for HoS to 25 hours, however twenty hours should be fine in most instances I believe. 

A player shouldn't be able to have the responsibility nor access that a Warden has without any experience as an officer, he has access to the armory, is generally in charge of sentencing barring the HoS/Captain/Magistrate/etc. opposing the sentencing, and is generally viewed as a de-facto second in command for security. They should be expected to have a moderate grasp of Space Law at the very least, which is what forcing them to require ten hours in a security position will do.

These numbers can be tweaked incredibly easily, but I strongly feel they should remain as is. At the very least Warden should.

## Changelog
:cl:
tweak: Upped the time required as security to unlock HoS to twenty hours. Made the Warden require ten hours as any security position to unlock.
/:cl:

### DIFF
--- a/code/game/jobs/job/security.dm
+++ b/code/game/jobs/job/security.dm
@@ -18,7 +18,7 @@
 			            ACCESS_RESEARCH, ACCESS_ENGINE, ACCESS_MINING, ACCESS_MEDICAL, ACCESS_CONSTRUCTION, ACCESS_MAILSORTING,
 			            ACCESS_HEADS, ACCESS_HOS, ACCESS_RC_ANNOUNCE, ACCESS_KEYCARD_AUTH, ACCESS_GATEWAY, ACCESS_PILOT, ACCESS_WEAPONS)
 	minimal_player_age = 21
-	exp_requirements = 300
+	exp_requirements = 1200
 	exp_type = EXP_TYPE_SECURITY
 	disabilities_allowed = 0
 	outfit = /datum/outfit/job/hos
@@ -64,7 +64,7 @@
 	minimal_access = list(ACCESS_SECURITY, ACCESS_SEC_DOORS, ACCESS_BRIG, ACCESS_ARMORY, ACCESS_COURT, ACCESS_MAINT_TUNNELS, ACCESS_WEAPONS)
 	minimal_player_age = 21
 	exp_requirements = 600
-	exp_type = EXP_TYPE_CREW
+	exp_type = EXP_TYPE_SECURITY
 	outfit = /datum/outfit/job/warden
 
 /datum/outfit/job/warden


Shockpoint:
Not quite understanding the thumbs down on this. As both an admin and player that has dealt with new players jumping straight into Warden instead of actually learning security, it's frustrating.

They don't know Space Law, Sec SOP, or anything past "I HAVE A GUN AND COOL STUFF". They're almost _always_ poor at communication, and frequently don't know how sec records work, or how to brig someone, let alone with the correct sentencing.  This usually leads to officers getting very angry with these players, or the HoS just outright firing them for complete incompetence. 

A Warden's job is often the most difficult during fast-paced rounds with higher antagonist activity. And as such, I'd much prefer to see more experienced people with this job rather than a green player that can't figure out what their baton does.

Mitchs98:
And, as stated, if the numbers are too steep for HoS I can tweak them down to 15 hours or so which I would also find semi-reasonable. At the very least, even if heads and maints want the HoS portion removed, Warden carries much more responsibility than normal officers. They _should_ come with some sort of timelock that requires security experience.

I genuinely think having these jobs require more dedication to get into doing will help cut down on a lot of shitcurity. Officers remain at the same timelock they've been, so security being understaffed won't really be affected.

And if people are still concerned with Warden and HoS being unfilled, let me ask you this.

Would you rather it be taken by an inexperienced player that barely knows how to use their secHUD/doesn't know how to use one at all, doesn't know the law, breaks the law, and just uses the job to wave their authority around?

Or would you rather it not be filled as often, people that are interested in learning security do so working their way through the ranks at a reasonable level, and actually have decent knowledge with the job and its inner workings at each tier of power and THEN take the job when they're actually ready?

This is, of course, baring in mind Warden can be promoted to Acting HoS without _any_ security experience at all in any given round that they just unlocked Warden and joined in.

AffectedArc07:
I dont think this will help the situation, mainly because if we increase the playtime for warden and HoS drastically, I feel security will become more of a "job you grind the shit out of without actually caring how you play it" instead of "A job you play properly and then get the unlocks as time goes by". Perhaps this is me being too cynical, but I feel like this will make the quality of officers go *down*.

Mitchs98:
> I dont think this will help the situation, mainly because if we increase the playtime for warden and HoS drastically, I feel security will become more of a "job you grind the shit out of without actually caring how you play it" instead of "A job you play properly and then get the unlocks as time goes by". Perhaps this is me being too cynical, but I feel like this will make the quality of officers go _down_.

If that's the case and that's how people treat the job then admins will even more easily be able to differentiate between players properly learning the job and playing it because they want to learn it and those that just want to unlock the higher positions because 'muh authority'. Not only that most of the playerbase wouldn't be willing to grind it just to be able to get authority for the sake of authority, they'd grind it because they want to learn those roles and properly play the higher roles. The average low quality player in my experience really doesn't have the dedication to play for a large amount of hours just to unlock roles with more authority.

And if they do it'd be painfully obvious by the time they get it and likely net themselves a job ban and waste all that time in the process.

procdrone:
I agree with upping the hour requirement. Too many times i had to explain to warden how to process a simple criminal for a standard cell let alone perma. People coming into sec higher roles should be expected to know how the basics work, given a higher standard is required of sec as a whole.

BakKeener:
I personally agree with upping the time required for Warden/HoS. In my experience I have seen way too many incompetent Wardens and Head of Securities just because they were able to unlock the job super quick without any real effort in learning how security and all the intricacies that involves it work.

Mitchs98:
Closing as #14290 was merged, which covers the roles in this PR.

PR #13773, opened 2020-07-05
Fixes SR spawning mobs in nullspace instead of hydroponics trays

Gatchapod:
## What Does This PR Do
Makes killer tomatoes and pine trees spawn when Strange Reagent is added to a hydroponics tray.

## Why It's Good For The Game
It's the intended behaviour, but mobs were spawning in nullspace.

## Changelog
:cl:
fix: Aggressive plants will now properly spawn when SR is added to a hydroponics tray
/:cl:

### DIFF
--- a/code/modules/hydroponics/hydroponics.dm
+++ b/code/modules/hydroponics/hydroponics.dm
@@ -978,7 +978,7 @@
 /obj/machinery/hydroponics/proc/spawnplant() // why would you put strange reagent in a hydro tray you monster I bet you also feed them blood
 	var/list/livingplants = list(/mob/living/simple_animal/hostile/tree, /mob/living/simple_animal/hostile/killertomato)
 	var/chosen = pick(livingplants)
-	var/mob/living/simple_animal/hostile/C = new chosen
+	var/mob/living/simple_animal/hostile/C = new chosen(src.loc)
 	C.faction = list("plants")
 
 /obj/machinery/hydroponics/proc/become_self_sufficient() // Ambrosia Gaia effect
--- a/code/modules/hydroponics/hydroponics.dm
+++ b/code/modules/hydroponics/hydroponics.dm
@@ -978,7 +978,7 @@
 /obj/machinery/hydroponics/proc/spawnplant() // why would you put strange reagent in a hydro tray you monster I bet you also feed them blood
 	var/list/livingplants = list(/mob/living/simple_animal/hostile/tree, /mob/living/simple_animal/hostile/killertomato)
 	var/chosen = pick(livingplants)
-	var/mob/living/simple_animal/hostile/C = new chosen(src.loc)
+	var/mob/living/simple_animal/hostile/C = new chosen(get_turf(src))
 	C.faction = list("plants")
 
 /obj/machinery/hydroponics/proc/become_self_sufficient() // Ambrosia Gaia effect


code/modules/hydroponics/hydroponics.dm
```diff
@@ -978,7 +978,7 @@
 /obj/machinery/hydroponics/proc/spawnplant() // why would you put strange reagent in a hydro tray you monster I bet you also feed them blood
 	var/list/livingplants = list(/mob/living/simple_animal/hostile/tree, /mob/living/simple_animal/hostile/killertomato)
 	var/chosen = pick(livingplants)
-	var/mob/living/simple_animal/hostile/C = new chosen
+	var/mob/living/simple_animal/hostile/C = new chosen(src.loc)
```
SteelSlayer, line 981: I would go with `get_turf(src)`. Either way, you don't want the implied `src`.

Gatchapod: It might not be the best practice to refer to variables directly between objects, but that is not the case here.
And I don't quite see how `src` is implied when it's explicitly stated? Implied `src` would be just making it `new chosen(loc)`, which works exactly the same. I merely think `src.loc` is more readable.

SteelSlayer: "It might not be the best practice to refer to variables directly between objects, but that is not the case here." I'm not sure what you mean by this, can you explain?

The reason I mentioned the `src` is because Paradise's (and probably most other server's) preferred style is that we do not write the implied the `src`.

Gatchapod: Using DM terminology, some OOP theory:
Each atom has a `loc` var. When another atom wants to access said var, it should use a proc that returns the value of said var. So instead of `someAtom.loc`, it should, in case of Paradise, use `get_turf(someAtom)`, which more-or-less returns `loc`.
Procs belonging to a particular atom can (/should be able to) safely refer to its vars directly.

Admittedly, while explaining this to you, I have realized a mistake in my reasoning here. I am "feeding" one atom a raw var of another one, so definitely `get_turf()` is more appropriate.

Regarding the `src` bit, it is a bit redundant here, as explained. If skipping it whenever its implied is the consensus, I will adjust.

SteelSlayer: Yeah, it's *possible* that `loc` is something other than a turf, such as a mob, or another object, but lot of the time it is the turf it's on. And I certainly hope the hydroponics tray's `loc` is always a turf :stuck_out_tongue:. 

IMO, its best to stay consistent, and that goes for all items. If your intent is that the new item appear on a turf, use `get_turf()`. But, not sure we have a preferred style on that yet... its not in the guidelines.

Gatchapod: That's more-or-less what I'm talking about. Within a specific atom, you _know_ what `loc` is. You can safely refer to it directly.
You cannot, however, truly foresee what `loc` will be in atoms you are not working in. Even if in your particular relation (i.e. hydroponics tray and mob) data is compatible, in future someone might try to feed your proc an invalid one (let's say you create a living tray that's a mob with the machinery "inside" of it, making the tray's `loc` a mob).

Sadly, I don't think I can use `get_turf()` without using `src` in this case.

SteelSlayer: Yep exactly. Also writing `get_turf(src)` is perfectly acceptable, necessary even.

PR #13779, opened 2020-07-06
Headcrab Cycle Fix

Terilia:
## What Does This PR Do

It fixes an issue. Kyet and me agreed, back then when Headcrab were new, that they should only look for food every 4 Ticks to keep CPU low - as they are a swarm enemy. 

Unfortunately due to a very annoying mistake, this has been never actually be implemented. This now has been fixed. 

## Why It's Good For The Game

It was supposed to be in the game already, hasn't been. This adds it back
## Changelog
:cl: Terilia
fix: Headcrabs now only look every 4 ticks for food.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -23,8 +23,6 @@
 	robust_searching = 1
 	var/host_species = ""
 	var/list/human_overlays = list()
-	var/cycles = 4
-
 
 /mob/living/simple_animal/hostile/headcrab/Life(seconds, times_fired)
 	if(..() && !stat)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -23,6 +23,8 @@
 	robust_searching = 1
 	var/host_species = ""
 	var/list/human_overlays = list()
+	var/cycles = 4
+
 
 /mob/living/simple_animal/hostile/headcrab/Life(seconds, times_fired)
```
dearmochi, line 29: Instead of storing a `cycles` variable altogether, perhaps you could simply modulo the `times_fired` argument - it should contain the number of times the `Mobs` subsystem fired since the start of the world.
A side effect of this is that all headcrabs will cycle at the same time rather than at sparse moments.

Terilia: I disagree with that solution. 

I think, on a per tick basis, a calculation of a bigger number, in this case a division of a number that only goes up dramatically, takes more cpu cycles than the division of 0,1,2,3,4

Terilia: And I have been proven wrong. I will add a version with modulo tomorrow.

PR #13781, opened 2020-07-06
Fixes timings

AffectedArc07:
## What Does This PR Do
Basically, I noticed that when the server inits, timing is all skewed for what actually takes time and what doesnt. 
image

Turns out, all this is from `start_watch()`. A proc which starts a timer, and then `stop_watch()` allows you to calculate the time difference between. Except, ***IT ONLY COUNTS TIME THAT THE GAME IS ACTUALLY RUNNING FOR, MEANING THAT WHILE THE SERVER IS HELD UP INITIALIZING, IT WONT EVEN TIME PROPERLY, MAKING AN INITIALIZATION TIMER BLOODY POINTLESS***. 

This PR makes it so that `start_watch()` uses real time, instead of game time. This means we actually get accurate startup timers now
image

I looked for each use of `start_watch()`, and its only used for initialization time logging. So this is fine.

## Why It's Good For The Game
A timer that checks how long the server has been doing a length operation for, **should not be affected by the server lagging for the love of god**

## Changelog
:cl: AffectedArc07
fix: Initializations are now accurately timed
/:cl:

### DIFF
--- a/code/__HELPERS/time.dm
+++ b/code/__HELPERS/time.dm
@@ -11,25 +11,6 @@
 
 #define TICKS2DS(T) ((T) TICKS)
 
-#define TimeOfGame (get_game_time())
-#define TimeOfTick (world.tick_usage*0.01*world.tick_lag)
-
-/proc/get_game_time()
-	var/global/time_offset = 0
-	var/global/last_time = 0
-	var/global/last_usage = 0
-
-	var/wtime = world.time
-	var/wusage = world.tick_usage * 0.01
-
-	if(last_time < wtime && last_usage > 1)
-		time_offset += last_usage - 1
-
-	last_time = wtime
-	last_usage = wusage
-
-	return wtime + (time_offset + wusage) * world.tick_lag
-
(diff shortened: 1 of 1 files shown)


code/__HELPERS/time.dm
```diff
@@ -11,25 +11,6 @@
 
 #define TICKS2DS(T) ((T) TICKS)
 
-#define TimeOfGame (get_game_time())
-#define TimeOfTick (world.tick_usage*0.01*world.tick_lag)
```
AffectedArc07, line 15: These defines were only used for this, were formatted wrong, and quite honestly fucking hurt me

KasparoVy:
Wtf? The timers were a lie all along!

Crazylemon64:
Forcing CI to re-run to see if it bugs out again

Crazylemon64:
still 4...?

AffectedArc07:
Travis is being wonky. *very* tempted to look into github actions and/or azure pipelines

PR #13783, opened 2020-07-07
Lavaland Atmos Fixes

Fox-McCloud:
some of the ruins have issues with simulated turfs that are just straight up exposed to lavaland, at roundstart, which causes active turfs to bloat up and stay bloated up for a while, until they equalize.

This doesn't address some of the major ones (Ratvar, summoning rune, and a few others), but this fixes some of the minor ones that have obvious oversights, while also ensuring that some of the turfs properly have planetary_atmos set to true---which should cause them to enter a more "settled" state, easier.

Also there was some errors on some turfs---the `/turf/simulated/floor/indestructible/boss/air` turf was...airless---also in the ashwalker den for some reason (they don't even breathe).

:cl: Fox McCloud
tweak: Tweaks some atmos settings to make lavaland slightly "quieter" in terms of having active atmos
/:cl:

### DIFF
--- a/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_ash_walker1.dmm
+++ b/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_ash_walker1.dmm
@@ -112,7 +112,7 @@
 	dir = 1
 	},
 /obj/item/flashlight/lantern,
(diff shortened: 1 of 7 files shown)


AffectedArc07:
@farie82 Mapdiffbot has the images. The fan is also under the door and isnt even visible.

PR #13784, opened 2020-07-07
Ports TG's Reference Tracking

Fox-McCloud:
Ports: https://github.com/tgstation/tgstation/pull/51467 100% credit to TG and @MCHSL

Implements EXTOOLs, which is an external .dll that allows you to hook into BYOND functions with C++; you can find more on it here: https://github.com/MCHSL/extools

Reference tracking is a nice little tool that allows you to check back and forward references to any datum or *KINDA* list--it's mostly to assist devs in figuring out what does and does not hold reference to things; you can enable it by setting the define, in compile options `REFERENCE_TRACKING`; you can also enable it, server side, to make the next round have reference tracking enabled.

Please note, using this, all the time, on the live server, is not recommended; it incurs additional overhead for everything while it's running just to track the forward and back refs; it's definitely something that should be reserved for *maintainers only* for debugging particularly expensive and problematic datums that just won't GC, and local testing isn't cutting it.

:cl: ASD and Fox McCloud
add: Adds a reference tracking tool that can allow devs to more easily find where references are hiding for datums
/:cl:


Fox-McCloud:
Upon further testing, I've confirmed the issues with this are not para related, but extend to TG as well. Until those issues are resolved, I'll be closing this.

KasparoVy:
Good call, I'm curious now though about what the issues were though? This is well over my head but I love hearing about it

MCHSL:
The issue was extools was recording references that didn't actually exist, made even more apparent by the fact that it often reported references held in variables that didn't even exist on the given types. I suspect it's an issue with detecting datum deletion. I'll be revisiting it sometime in the future, currently working on something way cooler (also extools related).

KasparoVy:
Thanks @MCHSL, greatly appreciated explanation!

PR #13785, opened 2020-07-08
Added security counter to check antags panel

TDSSS:
## What Does This PR Do
This PR adds a count to the check antags panel that tells admins the number of security officer on station, as well as how many are active, dead or antagonists.

It takes into account job transfers, so only people currently working as security officer or similar will be counted. Specifically, the roles that count are HoS, Warden, Detective, Security Officer and Pod Pilot

The definition of active is conscious and not an antagonist. Officers currently down for surgery or otherwise unconscious are not counted here. This does mean the three other numbers won't necessarily sum up to the first one, but it seems better than grouping unconscious officers with dead ones, or adding another entry just for unconscious.
Antag can include things like manifest hire traitors, but also shadowling thralls, which i expect to be the most common antagonist where this matters.

## Why It's Good For The Game
It lets admins know how bad things are for the station, letting them gauge if an ert is needed and how many people, if half of sec is sling thralls or not, if the spiders have murdered everyone and are taking over, etc.

## Images of changes

Capture
A quick picture of the panel addition in action. Not pretty, but most admin tools aren't.
Edit: changed it on kyet's suggestion, slightly less awful I guess.

## Changelog
:cl:
add: Security force counter for admins
/:cl:

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -624,16 +624,18 @@ GLOBAL_LIST_INIT(do_after_once_tracker, list())
 		return
 	return M.playtime_role ? M.playtime_role : M.assigned_role	//returns current role
 
-//checks the security force on station and returns a list of numbers, of the form:
-// total, active, dead, antag
-// where active is defined as conscious (STAT = 0) and not an antag
+/**	checks the security force on station and returns a list of numbers, of the form:
+ * 	total, active, dead, antag
+ * 	where active is defined as conscious (STAT = 0) and not an antag
+*/
 /proc/check_active_security_force()
 	var/sec_positions = GLOB.security_positions - "Magistrate" - "Brig Physician"
 	var/total = 0
 	var/active = 0
 	var/dead = 0
 	var/antag = 0
-	for(var/mob/living/carbon/human/player in GLOB.human_list)
+	for(var/p in GLOB.human_list)	//contains only human mobs, so no type check needed
+		var/mob/living/carbon/human/player = p	//need to tell it what type it is or we can't access stat without the dreaded :
 		if(determine_role(player) in sec_positions)
 			total++
 			if(player.stat == DEAD)
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -616,3 +616,32 @@ GLOBAL_LIST_INIT(do_after_once_tracker, list())
 		chosen = pick(mob_spawn_meancritters)
 	var/mob/living/simple_animal/C = new chosen(spawn_location)
 	return C
+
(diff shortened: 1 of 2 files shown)


code/__HELPERS/mobs.dm
```diff
@@ -616,3 +616,32 @@ GLOBAL_LIST_INIT(do_after_once_tracker, list())
...
+//checks the security force on station and returns a list of numbers, of the form:
+// total, active, dead, antag
+// where active is defined as conscious (STAT = 0) and not an antag
+/proc/check_active_security_force()
+	var/sec_positions = GLOB.security_positions - "Magistrate" - "Brig Physician"
+	var/total = 0
+	var/active = 0
+	var/dead = 0
+	var/antag = 0
+	for(var/mob/living/carbon/human/player in GLOB.human_list)
```
farie82, line 636: don't type cast in a for loop if you can help it.
`GLOB.human_list` only has humans in them so you can do a typeless loop

code/__HELPERS/mobs.dm
```diff
@@ -616,3 +616,32 @@ GLOBAL_LIST_INIT(do_after_once_tracker, list())
...
 	return C
+
+//determines the job of a mob, taking into account job transfers
+/proc/determine_role(mob/living/P)
+	var/datum/mind/M = P.mind
+	if(!M)
+		return
+	return M.playtime_role ? M.playtime_role : M.assigned_role	//returns current role
+
+//checks the security force on station and returns a list of numbers, of the form:
```
farie82, line 627: Can you make this into a documentation that the auto documentation tool can use?
```
/**
  * documentation here
  * more here etc
*/

Gatchapod:
I think admins should keep in mind the possibility of having an officer or two with non-standard titles. This might treat them as not-officers, even if they otherwise would me the active officer criteria.
It's probably not an issue large enough to warrant addressing it in code, just something admins should keep in mind when using this panel.

TDSSS:
The issue is, how do you detect officers with non-standard titles? If you just check for access, you'd also count things like a hop that gave himself all access, antags with stolen IDs, etc.

We can't even manage to display custom jobs in the right place on the manifest, they're all at the end with the civilians.

TDSSS:
Ah good, it's already using playtime role when set. I didn't know that custom job titles don't overwrite it.

code/__HELPERS/mobs.dm
TDSSS: Should be addressed now

code/__HELPERS/mobs.dm
TDSSS: Done, hope it's right like that.

TDSSS:
Capture
Done similar to what kyet suggested, made it a vertical table.

PR #13786, opened 2020-07-08
LINDA Performance Refactor and Changes

Fox-McCloud:
Refactors `process_cell` to perform a lot better, ports: https://github.com/tgstation/tgstation/pull/15218

Instead of adjacent turfs constantly being recalculated and it looping through through every turf and doing and istype check on them, it caches those turfs on a list of every turf.

There's one downside to this though---because it's a cached list, it's not going to play nice with melting walls;  walls melting are dependent on looping through *every* turf in order to work.

As such, melting walls have been removed---and to be honest, I think that's for the better; while there is some interesting notion to fire spreading in this manner, it also means fires tend to burn themselves out *very rapidly*; they just get too hot in a confined area, melt the walls, then rapidly spread out and dissipate not too long after. I don't consider this a net positive, as it means that, while yes, the fire may spill into other areas, it also means that a fire is far far more likely to breach to space and extinguish itself.

On that note, I also removed insulated floors--superconduction is a pretty important part of LINDA: being able to have temperatures that go to obscene levels because there's no temperature conduction with walls, the floor, or space just leads to *ridiculous* numbers that really shouldn't exist---this is going to become even *more* insane once we move to a primarily atmospherics based engine (Supermatter). I wouldn't be all surprised if it is playing its part in atmospheric oddities from time to time, as well.

I'd like this test merged on live before merging to make sure everything functions as intended.

:cl: Fox McCloud
tweak: Improves LINDA's performance, overall
del: walls can no longer melt from fire
del: insulated floors have been removed
/:cl:


code/LINDA/LINDA_fire.dm
```diff
@@ -164,16 +158,14 @@
 /obj/effect/hotspot/Destroy()
 	set_light(0)
 	SSair.hotspots -= src
+	var/turf/simulated/T = loc
+	if(istype(T) && T.active_hotspot == src)
+		T.active_hotspot = null
```
Fox-McCloud, line 163: Has to be moved, otherwise this is a Destroy loop.

code/LINDA/LINDA_turf_tile.dm
```diff
@@ -142,80 +141,89 @@
...
-					share_air(enemy_simulated) //share
+						our_excited_group = excited_group //update our cache
+					share_air(enemy_simulated, fire_count, adjacent_turfs_length) //share
 
 		/******************* GROUP HANDLING FINISH *********************************************************************/
 
 		else
-			if(!air.check_turf(enemy_tile, atmos_adjacent_turfs_amount))
-				var/difference = air.mimic(enemy_tile,atmos_adjacent_turfs_amount)
+			if(air.check_turf(enemy_tile, adjacent_turfs_length))
```
Fox-McCloud, line 215: There's my mistake.

code/LINDA/LINDA_turf_tile.dm
```diff
@@ -142,80 +141,89 @@
...
 						consider_pressure_difference(enemy_tile, difference)
 					else
 						enemy_tile.consider_pressure_difference(src, difference)
 				remove = 0
-				if(excited_group)
+				if(our_excited_group)
 					last_share_check()
 
-	if(planet_atmos) //share our air with the "atmosphere" "above" the turf
+	if(planetary_atmos) //share our air with the "atmosphere" "above" the tur
```
Fox-McCloud, line 226: `tur`. Pffft

TDSSS:
So walls are fireproof now? How does that affect the average plasma fire in, say, toxins storage? Is it contained to the room itself? Does it burn the doors and devastate parts of science anyway?

Carthusia:
I feel like the melting walls thing refers to things like acid melting, and maybe thermite, though I could be wrong?

Fox-McCloud:
Doors will be impacted by fires, as they have ever since object damage was put in--but only if the fire is on the same turf as the door.

Fires will typically be more well bound by individual rooms if they're solid; glass, however, is fair game.

Carthusia:
If this will be the case, doesnt this remove the purpose of fireproofing reinforced walls? and if so, would said wall type be removed from the code as it now would be moot to have.

Or does fireproofing a wall also make it more bomb proof?

Fox-McCloud:
Yes; I have removed all those walls from the codebase to avoid confusion; no sense in being able to upgrade a wall if it doesn't do anything.

TDSSS:
This has pretty huge balance implications, if a room without windows perfectly contains a fire.

Btw, do fires still melt floor? If they do that means they'll just burn through the floor and vent harmlessly into space, never spreading unless you deliberately open a door for them.

Muhdi345:
This PR will make a plasmafire even more of a joke as it is now. You might as well just remove any firefighting items too since they cannot be used even in the current situation, and if this gets merged, the only thing you need is a little patience.

In my opinion, plasma burns too hot and too fast which makes floor plating useless to resist it, I'm not even sure if the default floor plating even has a heat resist to speak off. 

Considering fires only typically happen in toxins storage or toxins proper, and there's no windows there, why would you even even open the door when you know the floor will destroy itself within the first 30 seconds? Heck, a default plasma canister barely has enough power to get through one reinforced wall under optimal conditions before burning itself out.

Carthusia:
I'm also concerned this kind of eliminates the concept of flooding the station with plasma, if all you have to do is RCD a wall in front of the fires path, they are VERY easy to stop.

Muhdi is correct, firefighting gear would more or less be moot when you can just RCD a wall in place and wait for it to burn out.

Is the performance increase this adds really worth this much of a change to the fundamental way people know plasma to be? Its supposed to be a very hazardous and volatile gas that risks burning out a section of the station very easily should something go wrong.

This makes it a mildly dangerous gas with a multitude of counters that can stop it in its tracks, preventing people from getting harmed, compared to the previous "Cut the supply if you can, and hold it at bay if at all possible until it burns out"

TDSSS:
This should have the balance tag, change my mind.

Fox-McCloud:
There's a misconception about fires burning through floors--it doesn't work by damaging the floor--it goes off of temperature, and the turf won't typically be destroyed while the fire is burning; it'll happen when the fire dies down.

How a turf is "burnt through" is by setting a var on the turf---once the hotspot is gone (which will only happen once fuel is exhausted), then the turf will be destroyed--meaning, once a hole is burnt through the floor, the fire was already largely done, anyway.

pinatacolada:
cant walls be damaged by fire just like windows are, it'd act the same as melting

PR #10955, opened 2019-02-26
Random advanced virus event

farie82:
**What does this PR do:**
Adds the advanced virus outbreak event.
Like the normal virus outbreak event this spawns a virus. But this one is advanced instead of a regular one.
Meaning it has 2 to 6 random symptoms which you can also get from human engineering.
The virus goes through the advanced virus limit and can always spread at least by touch.

**Changelog:**
:cl:
add: Added an advanced virus outbreak. Now besides a normal virus outbreak there is also a chance to get an advanced virus outbreak which has randomised symptoms
/:cl:

PR #13788, opened 2020-07-08
Reworks the disease outbreak event. Makes sure it is actually a potential dangerous issue to the crew

### LINKED ISSUE #11617: Can't get infected by the cold virus (event) when you already have an advanced virus
**Problem Description**:
Can't get sick if you have an advanced virus

Get sick

**What happened instead**:
Don't get sick. ABM won't work

Doesn't make sense

Have an advanced virus in you. Get the event to trigger (Or just spawn in the cold virus). Don't get sick.

Since forever. Especially now since you can only have 1 advanced virus

The event cold virus is an advanced virus. Meaning you won't get infected if you already have an advanced virus in you.

### LINKED ISSUE #13787: Transforming level 7 virus
**Issue Description**:

There was a random virus, which did have a name and cure of saline and spreading by blood. For some reason, it did transform into an unknown virus that spread by contact and had a different listed cure, acording to VV.

The virus to stay the same and let itself be cured.

**What happened instead**:

The virus seems to have needed one cure, but another was shown.

Medbay would need to test every potential cure to heal a virus if it changes like that, as no reading can be trusted.

It should be reproducable by a virus that has theese symptoms, but may be spawned by the event: /datum/symptom/itching /datum/symptom/sneeze  /datum/symptom/cough /datum/symptom/shivering /datum/symptom/headache
Name it, and in the first stage, it'll have that name. Later, it'll have the same ID, but different cure and unknown name.

Unknown.

**Extra information**:

Was checked on with VV and SDQL to get all /disease/advance, SDQL found only one of the named virus remaining, which was on the carrier of the event, staying at stage 1, many instances of the unknown version with the same symptoms, but other cure and spread. Did VV one infected early, they had the named version, but later, it switched to unknown with the different cure and spread, which i belive was caused by the raised stage.

farie82:
## What Does This PR Do
Makes the disease outbreak event actually work when it rolls to create a random advanced virus.
1. Will actually give it symptoms instead of the default 1 or 2 it gets. Also removes the weird property setting that just broke the virus when it spread to somebody else. Fixes #13787
2. Makes sure the virus created is actually able to spread. Either by it's stats or by sneezing.
3. Makes the virus able to be contracted even if you already have one advanced virus. This will make curing viruses using ABM actually valid again. fixes #11617

Partial revive of #10955

## Why It's Good For The Game
1. It fixes a bug
2. It will ensure the virus actually spreads and not just infects patient 0.
3. It will ensure the virus actually will be an issue to the crew even if the virologist already made his healing virus

## Changelog
:cl:
tweak: Random event viruses are now guaranteed to spread
tweak: Random event viruses now spread to victims even if a victim already has an advanced virus
fix: Random event advanced viruses now actually get a set of symptoms instead of the default generated ones.
/:cl:

### DIFF
--- a/code/modules/events/disease_outbreak.dm
+++ b/code/modules/events/disease_outbreak.dm
@@ -9,12 +9,25 @@
 		D = new virus_type()
 	else
 		var/datum/disease/advance/A = new /datum/disease/advance
+		// Try generating a random spreading virus. If that fails just use the last generated one
+		for(var/i = 0; i < 100; i++)
+			var/symptom_amount = rand(2, 6)
+			QDEL_LIST(A.symptoms)
+			A.symptoms = A.GenerateSymptoms(1, 6, symptom_amount)
+			A.Refresh(FALSE, FALSE) // Don't actually archive the miscreants
+			if(VirusCanSpread(A))
+				break
 		A.name = capitalize(pick(GLOB.adjectives)) + " " + capitalize(pick(GLOB.nouns + GLOB.verbs)) // random silly name
-		A.GenerateSymptoms(1,9,6)
-		A.AssignProperties(list("resistance" = rand(0,11), "stealth" = rand(0,2), "stage_rate" = rand(0,5), "transmittable" = rand(0,5), "severity" = rand(0,10)))
+		A.event_spawned = TRUE
+		A.Refresh() // Actually archive it with it's name
 		D = A
 
 	D.carrier = TRUE
+
+/datum/event/disease_outbreak/proc/VirusCanSpread(datum/disease/advance/A)
+	if(A.spread_flags == BLOOD)
+		return locate(/datum/symptom/sneeze) in A.symptoms // Should atleast have sneezing then to be considered spreading
+	return TRUE
 
 /datum/event/disease_outbreak/announce()
 	GLOB.event_announcement.Announce("Confirmed outbreak of level 7 major viral biohazard aboard [station_name()]. All personnel must contain the outbreak.", "Biohazard Alert", new_sound = 'sound/AI/outbreak7.ogg')
--- a/code/modules/events/disease_outbreak.dm
+++ b/code/modules/events/disease_outbreak.dm
@@ -18,10 +18,12 @@
 			if(VirusCanSpread(A))
 				break
 		A.name = capitalize(pick(GLOB.adjectives)) + " " + capitalize(pick(GLOB.nouns + GLOB.verbs)) // random silly name
-		A.event_spawned = TRUE
 		A.Refresh() // Actually archive it with it's name
 		D = A
 
+	if(istype(D, /datum/disease/advance))
+		var/datum/disease/advance/A = D
+		A.event_spawned = TRUE // Will make random advanced viruses and default ones like the flu or such also able to spread
 	D.carrier = TRUE
 
 /datum/event/disease_outbreak/proc/VirusCanSpread(datum/disease/advance/A)
--- a/code/modules/events/disease_outbreak.dm
+++ b/code/modules/events/disease_outbreak.dm
@@ -10,7 +10,7 @@
 	else
 		var/datum/disease/advance/A = new /datum/disease/advance
 		// Try generating a random spreading virus. If that fails just use the last generated one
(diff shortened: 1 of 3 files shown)


Regular-Joe-SF:
Thanks, the random outbreaks as well as ABM are quite nil as features by now, this would be very neat still rather small edit to help that situation.

code/modules/events/disease_outbreak.dm
```diff
@@ -9,13 +9,28 @@
 		D = new virus_type()
 	else
 		var/datum/disease/advance/A = new /datum/disease/advance
+		// Try generating a random spreading virus. If that fails just use the last generated one
+		for(var/i = 0; i < 100; i++)
```
SteelSlayer, line 13: I would use the `for(var/i in 1 to 100)` syntax, as it's faster than this traditional for loop syntax.

AffectedArc07: image

farie82: Oh wew interesting. I'll change it

code/modules/events/disease_outbreak.dm
```diff
@@ -9,13 +9,28 @@
 		D = new virus_type()
 	else
 		var/datum/disease/advance/A = new /datum/disease/advance
+		// Try generating a random spreading virus. If that fails just use the last generated one
+		for(var/i in 1 to 100)
```
Kyep, line 13: I really don't like this. It seems like it just generates up to 100 random viruses at once until it finds one that's acceptable.
Instead of doing this, why not just say, make the 1st symptom sneezing, then auto-gen the rest?
That would mean you get a valid result in one generation, as opposed to taking up to 100.

code/modules/events/disease_outbreak.dm
```diff
@@ -9,13 +9,28 @@
...
-		A.AssignProperties(list("resistance" = rand(0,11), "stealth" = rand(0,2), "stage_rate" = rand(0,5), "transmittable" = rand(0,5), "severity" = rand(0,10)))
+		A.Refresh() // Actually archive it with it's name
 		D = A
 
+	if(istype(D, /datum/disease/advance))
+		var/datum/disease/advance/A = D
+		A.event_spawned = TRUE // Will make random advanced viruses and default ones like the flu or such also able to spread
 	D.carrier = TRUE
 
+/datum/event/disease_outbreak/proc/VirusCanSpread(datum/disease/advance/A)
```
Kyep, line 29: First, I'm not sure all viruses are spreadable by default. What about viruses with no spreading flags? You don't handle that case.
Second, why are blood-only virusses spread by sneezing?
Third, why not just add an extra symptom rather than re-genning until you get the one you want?

code/modules/events/disease_outbreak.dm
farie82: Adding sneezing as a fixed symptom would make it very meta-able. Now it can also spread without. And can for instance be touch spread.

Ive tested the call a few times. It made a functioning virus every time below the 20 tries on average.

code/modules/events/disease_outbreak.dm
farie82: All viruses are atleast spread by blood. That is how advanced viruses work.

Smeezing wont force the airborne trait on the virus. Instead it makes the virus airborne due to the activation of the symptom. Sneezing when activated will spread the virus to surrounding mobs. Toxic filterings activation will heal toxins.

Adding one random symptom can make the virus get into a minimum it cant get out of. The more symptoms a virus has the lower its transmition stat is. Rerolling the entire virus will thus result in better results

code/modules/events/disease_outbreak.dm
Kyep: I'm going to leave an objection on this until your generation is re-written to avoid re-rolling, and instead use a system that only generates once, but generates in a way that is guaranteed to produce a correct output that one time.

Kyep:
When you are free to voice, poke me on Discord?
I'd like to discuss this PR more, but I may need some help understanding the viro mechanics, as I do not play viro myself.

farie82:
Closing this for now till the github is a bit more settled and there is more time to look at this

PR #13791, opened 2020-07-09
Makes Ice Cream GC

Fox-McCloud:
So strange--ice cream cones are actually one of the most consistent things (outside of mobs) that doesn't GC.

This is due to them using `image` for their overalays.

Ice cream cones now properly use `mutable_appearance` instead of `image`; additionally, I removed the ice cream vat that was only used on meta and delta and standardized on the one on Box.

Also implemented a better ice cream topping icon that works better with color mixing and reagents.

:cl: Fox McCloud
fix: makes ice cream garbage collect properly, increasing performance
/:cl:


code/modules/food_and_drinks/kitchen_machinery/icecream_vat.dm
```diff
@@ -43,10 +43,9 @@
 		return
 	if(istype(I, /obj/item/reagent_containers/food/snacks/icecream))
 		if(!I.reagents.has_reagent("sprinkles"))
-			if(I.reagents.total_volume > 29) I.reagents.remove_any(1)
-			I.reagents.add_reagent("sprinkles",1)
-			var/image/sprinkles = image('icons/obj/kitchen.dmi', src, "sprinkles")
```
Fox-McCloud, line 48: this didn't exist and has never existed.

Spacemanspark:
Why not just eat it instead of wasting perfectly good ice cream

PR #13795, opened 2020-07-10
Optimizes Smoking

Fox-McCloud:
I noticed that nicotine was one of the most deleted reagents in the entire game----and I immediately knew why (no, it's not that smoking is super super popular--that plays into it a bit).

Smoking currently injected 0.4 of its current chem into the person---then said person processes said reagent which removes it.

This means two things: every time that person hits 0 on the volume of that particular reagent, it's deleted---it also means that the next time you smoke and ingest more nicotine, it notes that there's no nicotine reagent in the mob, so it creates a new instance.

Reagent creation+destroying is ultra cheap, but this was an obvious optimization that doesn't really increase complexity (and it prevents weirdness with reagent flags getting toggled on and off repeatedly too).

The solution: the first time you take a puff on a cig, it injects 1 reagent instead of 0.4.

What this means is that the reagent is created once, it doesn't instantly deplete, and once the next time you smoke rolls around, it just change a var value instead of creating an entirely new instance of that reagent (let's be honest, ti's going to be nicotine 99% of the time).

If you smoke for 40 minutes straight, this is how many instance of nicotine will be created+destroyed, normally:

```
/datum/reagent/nicotine
	qdel() Count: 1600
	Destroy() Cost: 23ms
```

Here's how much nicotine will be created+destroyed, after this optimization:

```
/datum/reagent/nicotine
	qdel() Count: 5
	Destroy() Cost: 0ms
```

This is a micro optimization; this isn't realistically going to really result in that much of a performance gain (23 ms saved over 40 minutes is practically nothing--there's probably more performance gain to be had by the fact of less `locate` calls and less queue length in the GC), but this was an easy to implement optimization, so why not?

:cl: Fox McCloud
fix: Smoking optimized
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/cigs.dm
+++ b/code/game/objects/items/weapons/cigs.dm
@@ -31,6 +31,7 @@ LIGHTERS ARE IN LIGHTERS.DM
 	var/smoketime = 150
 	var/chem_volume = 60
 	var/list/list_reagents = list("nicotine" = 40)
+	var/first_puff = TRUE // the first puff is a bit more reagents ingested
 	sprite_sheets = list(
 		"Vox" = 'icons/mob/species/vox/mask.dmi',
 		"Unathi" = 'icons/mob/species/unathi/mask.dmi',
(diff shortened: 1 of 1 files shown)


KasparoVy:
Smoke 'em if you got 'em
tf2 spy the gentlest gentleman

PR #13801, opened 2020-07-10
Notify admins when AI core is created

Kyep:
## What Does This PR Do
Notifies admins, and writes a message to game.log, whenever an AI core is completed.
The notification includes the name of the person who did it, where they are, and the zlevel.

## Why It's Good For The Game
Lately, we've been having issues with free golems in lavaland creating AIs every shift, even though the rules prohibit this.
This notification to admins will make it easier for admins to catch this, and stop it, before it gets out of control.

## Images
ai_core_finish_log

## Changelog
:cl: Kyep
add: Completing an AI core now notifies admins.
/:cl:

### DIFF
--- a/code/game/machinery/computer/ai_core.dm
+++ b/code/game/machinery/computer/ai_core.dm
@@ -154,8 +154,8 @@
 			state = SCREWED_CORE
 		if(GLASS_CORE)
 			var/area/R = get_area(src)
-			message_admins("[key_name_admin(usr)] has completed an AI core in [R] on z[z].")
-			log_game("[key_name(usr)] has completed an AI core in [R] on z[z].")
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/ai_core.dm
```diff
@@ -153,6 +153,9 @@
 			to_chat(user, "<span class='notice'>You screw the circuit board into place.</span>")
 			state = SCREWED_CORE
 		if(GLASS_CORE)
+			var/area/R = get_area(src)
+			message_admins("[key_name_admin(usr)] has completed an AI core in [R] on z[z].")
```
AffectedArc07, line 157: Make this use `ADMIN_JMP(src)`, so the admins have a jump button.

PR #13806, opened 2020-07-11
Fixes drift on projectiles firing west

### LINKED ISSUE #13248: Emitter Beams drift South when pointing West
**Issue Description**:
Emitter beams drift south while pointing west, but don't drift at all when pointing east.

There would either be south drift when pointing east or west, or no drift at all.

**What happened instead**:
Emitter beams drift south when pointing west, but not when pointing east. There is also no drift when pointing north.

It makes it harder to fire emitter beams at targets to the west.

1. Set up emitter beam pointing west. Turn it on.
2. See the drift to the south.
3. Set up emitter beam pointing east. Turn it on.
4. See there's no drift to the south.

**Extra information**:
Reading an old bug report about the same thing, apparently ALL projectiles drift south when pointing west.

DaveKorhal:
Added variable to offset rounding so that it actually rounds to the closest number instead of flooring it to the next lowest integer.

## What Does This PR Do
Changes the projectile X/Y movement calculations to actually round to the nearest integer instead of flooring it to the next lowest integer. This stops projectiles firing straight west from drifting south and makes aiming in general slightly more accurate.
Fixes #13248 

## Why It's Good For The Game
Makes aim slightly more accurate.
Fixes noticeable southward drift on projectiles that should be firing straight west.

## Changelog
:cl:
tweak: Fixed error causing projectiles to drift slightly off-target.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -247,8 +247,8 @@
 			M.Turn(Angle)
 			transform = M
 
-			var/Pixel_x = round(sin(Angle) + 16 * sin(Angle) * 2)
-			var/Pixel_y = round(cos(Angle) + 16 * cos(Angle) * 2)
+			var/Pixel_x = round(sin(Angle) + 16 * sin(Angle) * 2, 1)
+			var/Pixel_y = round(cos(Angle) + 16 * cos(Angle) * 2, 1)
 			var/pixel_x_offset = pixel_x + Pixel_x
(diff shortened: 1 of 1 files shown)


dearmochi:
Looks good. One minor complaint: the changelog is too technical. It's displayed to players in-game and most of them probably wouldn't be able to make any sense of the first half. I recommend wording it as such that the layperson instantly understands the impact of this change. Something like:

> Fixed projectiles such as bullets or lasers drifting away from their intended direction in some cases

PR #13807, opened 2020-07-12
Xeno Disarm Rework

Fox-McCloud:
A larger rework of xenos is beyond what I have time for at the moment, but this is a fairly simple change that was easy to do:

- Xeno disarm is not longer a near guaranteed down; if the person is holding an object in their hand, it will disarm of them of that object; only if their hands are empty (or they have a deadthgrip on said item via anti drop) will it still down them
- Attacking with your claws no longer has a random chance to down someone
- Flattened Xeno claw damage to 20, instead of it being random between 15 and 30
- Xenos have a 10% chance to miss their attack entirely
- Claws have 10 armor penetration
- Borgs operate on the same premise, only it's their currently selected active module
- Reduces borg stun time from 7 to 2
- Borg damage flattened to 20 instead of between 10 and 20

:cl: Fox McCloud
tweak: Xeno stun reworked; having an active item in your hand will now provide protection against instantly getting tackled down; borgs rely on their active module instead
tweak: Borg xeno disarm time reduced from 7 to 2
tweak: xeno claws can no longer stun and can miss, but have more consistent damage with very slight armor penetration
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -603,8 +603,7 @@ emp_act
 			var/obj/item/I = get_active_hand()
 			if(I && unEquip(I))
 				playsound(loc, 'sound/weapons/slash.ogg', 25, TRUE, -1)
-				visible_message("<span class='danger'>[M] disarms [src]!</span>", \
-								"<span class='userdanger'>[M] disarms you!</span>", "<span class='hear'>You hear aggressive shuffling!</span>", null, M)
+				visible_message("<span class='danger'>[M] disarms [src]!</span>", "<span class='userdanger'>[M] disarms you!</span>", "<span class='hear'>You hear aggressive shuffling!</span>")
 				to_chat(M, "<span class='danger'>You disarm [src]!</span>")
 			else
 				var/obj/item/organ/external/affecting = get_organ(ran_zone(M.zone_selected))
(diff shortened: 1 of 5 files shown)


SpookSpooky:
Thank you.

Mitchs98:
Definitely a good start in the right direction.

Balthusdire:
Feels like a good mix of making borgs less useless, but still giving the xenos the ability to kill them. Would have to see how it works out when live but seems good.

Carthusia:
I wasn't even aware xenos stunned borgs for 7 seconds per hit, sheeeesh.

I'm not 100% sure on the anti-drop thing still being affected.

Weapon in hand, cant get stunned, weapon in hand BUT also anti-dropped? Stunned.

Kind of makes having anti-drop a disadvantage for no logically good reason besides (Xenos couldn't knock them over otherwise), unless im misreading that part.

SteelSlayer:
To clarify for non-coders, the 7 and 2 are in life 'cycles' so that's 14 seconds and 4 seconds. It varies slightly of course due to possible lag etc.

Samman166:
good shit

Rurik123:
Awesome

lolzyking:
This is a good start

Regular-Joe-SF:
Thanks, very interesting to see how it will play out! I’d totally like xenos back around with something like this.

AffectedArc07 (approved):

PR #13811, opened 2020-07-12
Build time for aquariums

### LINKED ISSUE #13809: Fulltile windows, directional windows, wall aquariums, and fish tanks can be made instantly
**Issue Description**:
As said in the title, anything it seems with glass can be made instantly, allowing for people to make an instant barricade blocking pursuers.  This is the reason instant table construction was nerfed.

For there to be a timer attached to building anything glass related.

**What happened instead**:
Instant barricades against people chasing the builder preventing people from following/shooting them.

As said above, it is horrible balance wise and the reason tables/racks were nerfed with the build timer long ago.

Get glass, run, build anything glass related while running in a chokepoint.  Profit.

Since forever, just never really realized until now.

**Extra information**:

TDSSS:
## What Does This PR Do
This adds a short build time for fish bowls and aquariums, fixes #13809

## Why It's Good For The Game
You can't flee from sec and instantly drop dense items behind you to stop people chasing you. 

## Changelog
:cl:
tweak: fish bowls and aquariums now take some time to make.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/glass.dm
+++ b/code/game/objects/items/stacks/sheets/glass.dm
@@ -15,11 +15,11 @@
  */
 
 GLOBAL_LIST_INIT(glass_recipes, list ( \
-	new/datum/stack_recipe/window("directional window", /obj/structure/window/basic, time = 0, on_floor = TRUE, window_checks = TRUE), \
-	new/datum/stack_recipe/window("fulltile window", /obj/structure/window/full/basic, 2, time = 0, on_floor = TRUE, window_checks = TRUE), \
-	new/datum/stack_recipe("fishbowl", /obj/machinery/fishtank/bowl, 1, time = 0), \
-	new/datum/stack_recipe("fish tank", /obj/machinery/fishtank/tank, 3, time = 0, on_floor = TRUE), \
-	new/datum/stack_recipe("wall aquariam", /obj/machinery/fishtank/wall, 4, time = 0, on_floor = TRUE) \
+	new/datum/stack_recipe/window("directional window", /obj/structure/window/basic, time = 5, on_floor = TRUE, window_checks = TRUE), \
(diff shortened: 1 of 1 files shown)


Carthusia:
How much time exactly? 

Personally ive never seen them used in such a way but i can imagine them doing so, but the time is not going to be as long as building a wall segment is it?

TDSSS:
It's deciseconds, so 40 means 4 seconds I believe. It's definitely a very short amount of time for all of them.

Carthusia:
ahh alright, its basically no time at all, but just long enough to make it not a viable tactic to drop in the middle of a chase due to you needing to stop at all.

Drobe50:
What about inflatable walls and doors? Those can be used to instantly create a barrier to escape from someone.

TDSSS:
IMO, those are a lot less abusable. It's trivial to get a stack of 50 glass and have more instant windows than you'd ever need, but inflatable walls are a lot less common.

Carthusia:
You can also just pop the inflatable ones

procdrone:
inflatable wall takes few seconds to deflate

Carthusia:
Really? I thought you could just shoot them or stab them and they pop?

Fox-McCloud:
Fishtanks make sense, but windows start off unanchored, are easy to decosntruct, or can easily be rotated; they take time cause trouble---so I kinda raise an eyebrow at this, I guess, unless you've actively seen this abused quite often.

procdrone:
> could just shoot them or stab them and they pop?

Alright I tested it quickly and appears my knowledge on inflated walls is outdated. Previously when you used a screwdriver it would PUNCTURE the wall and it would deflate after few seconds. now, you can smack it more(need to cause enough damage), but it deflates immediately.

TDSSS:
PR has been updated to remove the changes to windows. It's only fish bowls and aquariums that have a buildtime.

PR #13816, opened 2020-07-12
Adds the guardian element

SteelSlayer:
## What Does This PR Do
Adds a new element: the guardian element, which is designed to handle all the interactions hosts have with their guardians. Those being recalling, communication and re-rolling.

The current way guardians are handled is gross. For example this happens whenever the user tries to recall their guardian:
```js
for(var/mob/living/simple_animal/hostile/guardian/G in GLOB.mob_list)
    if(G.summoner == src)
        G.Recall()
```
This loops through the entirety of all mobs in the list, filters it, and checks if the summoner is the correct one. Its an awful way to handle things and its what sparked me make this PR.

I thought about converting the guardian verbs into action buttons at first, but decided it might clutter up the host's button list, especially if they have other traitor things like chameleon items or implants. That would certainly be the *simpler* alternative. But with this PR, it allows us to keep the verbs. I thought I'd just throw this up and see what people think. It didn't take that long to write, so not a huge deal.

Backend change, so no CL.


SteelSlayer:
Action buttons would be better for this

PR #13818, opened 2020-07-12
Janiborgs can no longer clean while dead

### LINKED ISSUE #13802: Dead janitor-borgs can still clean
**Issue Description**:
Dragging a dead janiborg still cleans floor tiles that the janiborg passes over.

Dead janiborgs do not clean.

**What happened instead**:
Dead janiborgs still clean tiles they pass over.

Slaughter demons that kill a janiborg don't actually stop that borg cleaning, because anyone can drag the dead borg around to clean stuff.
Also, as a general rule, borg abilities shouldn't function while the borg is dead.

Create a janiborg.
Create some cleanable blood.
Kill the borg.
Drag the dead janiborg over the blood.
Observe the blood gets cleaned.

trololiver112:
## What Does This PR Do
Messed up my fork so this is a redo with Fox's suggestion

Fixes #13802

This PR makes it so that dead janitor borgs will no longer clean the floors under them.

## Why It's Good For The Game

This is just bad, an example of this would be trying to sacrifice a dead janitor borg as a cultist, but the borg cleaning the rune very time you put it on top of the rune.
This also doesn't make sense, if the borg is off or non functional, why would it still clean perfectly fine?

## Changelog
:cl:
tweak: Janitor borgs no longer clean while dead
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -1147,41 +1147,39 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 					updating = 0
 	if(module)
 		if(module.type == /obj/item/robot_module/janitor)
-			if(stat != DEAD)
-				var/turf/tile = loc
-				if(isturf(tile))
-					var/floor_only = TRUE
-					for(var/A in tile)
-						if(istype(A, /obj/effect))
-							if(is_cleanable(A))
-								var/obj/effect/decal/cleanable/blood/B = A
-								if(istype(B) && B.off_floor)
-									floor_only = FALSE
-								else
-									qdel(A)
-						else if(istype(A, /obj/item))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1147,39 +1147,41 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 					updating = 0
 	if(module)
 		if(module.type == /obj/item/robot_module/janitor)
-			var/turf/tile = loc
-			if(isturf(tile))
```
Fox-McCloud, line 1151: it should be here, on the same line, IMO, like I said in the first PR.

Regular-Joe-SF:
Funny, didn’t even realize this! No longer undead janitorial workers then, nice

Fox-McCloud:
👍

PR #13825, opened 2020-07-13
Async Lighting

AffectedArc07:
## What Does This PR Do
Makes lighting operations async.

Without async:
image

With async:
image

*Ignore the log format change, I redid the code and forgot to type the message properly* 

## Why It's Good For The Game
Makes lighting faster

## Changelog
:cl: AffectedArc07
tweak: Lighting is faster now
/:cl:


AffectedArc07:
So after speaking with people over in coderbus (Oranges, stylemistake, AnturK and Asd), apparently this should have 0 effect on performance, yet it does.

/shrug

AffectedArc07:
Better testing method

Before:
image

After:
image

Turns out, this is meaningless

PR #13835, opened 2020-07-14
[TGUI] Mining Equipment Vendor

dearmochi:
## What Does This PR Do

- Takes the torch from #13658 and re-organizes the prizes in the Mining Equipment Vendor in categories
- Replaces the ugly browser UI by TGUI!
- Refactors the code for readability and other good stuff

## Why It's Good For The Game

Makes Mining Equipment Vendors less of a pain to use (doesn't scroll back up, easier to find items) and makes the UI more aesthetically pleasing

## Images of changes

ejjp7G2Yms

## Changelog
:cl:
add: Add TGUI for Mining Equipment Vendor
add: Add ability to search and sort items in the Mining Equipment Vendor
tweak: Reorganize Mining Equipment Vendor items into categories
/:cl:


code/modules/mining/machine_vending.dm
```diff
@@ -126,62 +106,89 @@
...
-		dat += "<tr><td>[prize.equipment_name]</td><td>[prize.cost]</td><td><A href='?src=[UID()];purchase=\ref[prize]'>Purchase</A></td></tr>"
-	dat += "</table>"
-	var/datum/browser/popup = new(user, "miningvendor", "Mining Equipment Vendor", 400, 350)
-	popup.set_content(dat)
-	popup.open()
-
-/obj/machinery/mineral/equipment_vendor/Topic(href, href_list)
+		data["has_id"] = FALSE
+
+/obj/machinery/mineral/equipment_vendor/tgui_static_data(mob/user)
```
AffectedArc07, line 128: Thanks for using this to save on overhead

code/modules/mining/machine_vending.dm
```diff
@@ -126,62 +106,89 @@
...
+				to_chat(usr, "<span class='danger'>You have insufficient points.</span>")
 				return
 
 			inserted_id.mining_points -= prize.cost
-			new prize.equipment_path(src.loc)
-	updateUsrDialog()
+			new prize.equipment_path(loc)
+		else
+			return FALSE
+	add_fingerprint()
```
AffectedArc07, line 188: Might wanna add this to the start of the `tgui_act` because currently it wont be hit

SteelSlayer:
That's a sexy ass UI.

code/modules/mining/machine_vending.dm
dearmochi: It will, since the actual cases don't return. Only the default case returns, which makes sense because you don't want to add fingerprints if no action is performed (due to it being non-existent)

AffectedArc07: Oh I see, github was squishing my diff view :v

Carthusia:
oh god, this UI needs to eventually ported to the HOP's computer and their account terminal, it would make it SO much easier to just be able to search for an account number, instead of having to try to eyeball and scroll upwards to 150 accounts, hoping to god you spot it.

code/modules/mining/machine_vending.dm
```diff
@@ -126,62 +106,89 @@
...
-
-	var/dat
-	dat +="<div class='statusDisplay'>"
-	if(istype(inserted_id))
-		dat += "You have [inserted_id.mining_points] mining points collected. <A href='?src=[UID()];choice=eject'>Eject ID.</A><br>"
+	tgui_interact(user)
+
+/obj/machinery/mineral/equipment_vendor/tgui_data(mob/user)
+	var/list/data[0]
+	. = data
```
farie82, line 116: Best to use `return` when it's possible.
TG has a love for using `. = X` over return. Yet it makes the code a bit more abstract to glance over

code/modules/mining/machine_vending.dm
```diff
@@ -126,62 +106,89 @@
...
-	var/datum/browser/popup = new(user, "miningvendor", "Mining Equipment Vendor", 400, 350)
-	popup.set_content(dat)
-	popup.open()
-
-/obj/machinery/mineral/equipment_vendor/Topic(href, href_list)
+		data["has_id"] = FALSE
+
+/obj/machinery/mineral/equipment_vendor/tgui_static_data(mob/user)
+	var/list/static_data[0]
+	. = static_data
```
farie82, line 130: same here

code/modules/mining/machine_vending.dm
```diff
@@ -192,24 +199,29 @@
...
-			var/obj/item/card/id/C = usr.get_active_hand()
-			if(istype(C) && !istype(inserted_id))
-				if(!usr.drop_item())
-					return
-				C.forceMove(src)
-				inserted_id = C
-				interact(user)
+		var/obj/item/card/id/C = usr.get_active_hand()
+		if(istype(C) && !istype(inserted_id))
+			if(!usr.drop_item())
```
farie82, line 209: use `user` here

code/modules/mining/machine_vending.dm
```diff
@@ -192,24 +199,29 @@
...
 			return
-		else
-			var/obj/item/card/id/C = usr.get_active_hand()
-			if(istype(C) && !istype(inserted_id))
-				if(!usr.drop_item())
-					return
-				C.forceMove(src)
-				inserted_id = C
-				interact(user)
+		var/obj/item/card/id/C = usr.get_active_hand()
```
farie82, line 207: use `user` here

code/modules/mining/machine_vending.dm
```diff
@@ -192,24 +199,29 @@
...
+				return
+			C.forceMove(src)
+			inserted_id = C
+			tgui_interact(user)
 		return
 	return ..()
 
-/obj/machinery/mineral/equipment_vendor/proc/RedeemVoucher(obj/item/mining_voucher/voucher, mob/redeemer)
+/**
+  * Called when someone slaps the machine with a mining voucher
```
farie82, line 218: tres bean

dearmochi: 🥖 

(everything else is done)

tgui/packages/tgui/layouts/Window.js
```diff
@@ -73,15 +73,10 @@ export class Window extends Component {
...
-  const { scrollable, children } = props;
-  // A bit lazy to actually write styles for it,
-  // so we simply include a Box with margins.
+  const { scrollable, className, children } = props;
   return (
-    <Layout.Content
-      scrollable={scrollable}>
-      <Box m={1}>
-        {children}
-      </Box>
```
AffectedArc07, line 84: I feel like this change will space a lot of existing UIs out too much, and I dont know how I feel about that

dearmochi: Thing is, some components don't play nice with each other so it's better to give full control to the developer rather than have them do some workarounds to have a consistent experience

AffectedArc07, line 84: This currently breaks the crew monitor UI, however it doesnt break anything else.
image
image

Other UIs for context.
image

TLDR: Its not perfect, but its "manageable"

dearmochi: Added a class to solve this.

AffectedArc07 (approved):
Well theres a class to fix things, but I am taking you down with me if other PRs break

PR #13854, opened 2020-07-16
Fix player controlled simple animals still moving under AI pathing after a player takes control

### LINKED ISSUE #13752: Possessed TSpider still heads towards Last Target
**Issue Description**:

If you possess a Terror Spider that's chasing someone, it'll continue to try chasing them.

I expected to have full control of the Spider.

**What happened instead**:

It continuously tried to move towards its former target for a minute after possession, screwing up my own attempts to web people and fight.

Players shouldn't have to wrestle with the AI to control a mob.

-Possess a Terror Spider that's currently controlled by the AI and chasing someone.

I noticed it happening this afternoon, on July 1st 2020.

**Extra information**:

Kyep:
## What Does This PR Do
Fixes #13752 - simple animals still move under AI pathing after a player has taken control of them in some cases

## Why It's Good For The Game
Fixes an obvious (and annoying) bug.

## Explanation of why this works
https://secure.byond.com/docs/ref/index.html#/proc/walk
"A call to a walking function aborts any previous walking function called on Ref. To halt walking, call walk(Ref,0). "

## Changelog
:cl: Kyep
fix: Fixed AI sometimes trying to move mobs that are meant to be player-controlled.
/:cl:

### DIFF
--- a/code/modules/mob/living/login.dm
+++ b/code/modules/mob/living/login.dm
@@ -29,4 +29,6 @@
 	//Should update regardless of if we can ventcrawl, since we can end up in pipes in other ways.
 	update_pipe_vision()
 
+	walk(src, 0) // if mob is moving under ai control, then stop AI movement
+
 	return .
--- a/code/modules/mob/living/login.dm
+++ b/code/modules/mob/living/login.dm
@@ -29,6 +29,4 @@
 	//Should update regardless of if we can ventcrawl, since we can end up in pipes in other ways.
 	update_pipe_vision()
 
-	walk(src, 0) // if mob is moving under ai control, then stop AI movement
-
 	return .
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/ghost.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/ghost.dm
@@ -42,5 +42,6 @@
 		to_chat(user, "<span class='notice'>Someone else already took this spider.</span>")
 		return
 	key = user.key
+	walk(src, 0) // if spider is moving under ai control (already aggroed on something before player control), then stop AI movement
 	for(var/mob/dead/observer/G in GLOB.player_list)
 		G.show_message("<i>A ghost has taken control of <b>[src]</b>. ([ghost_follow_link(src, ghost=G)]).</i>")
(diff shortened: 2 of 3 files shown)


code/modules/mob/living/simple_animal/hostile/terror_spiders/ghost.dm
```diff
@@ -42,5 +42,6 @@
 		to_chat(user, "<span class='notice'>Someone else already took this spider.</span>")
 		return
 	key = user.key
+	walk(src, 0) // if spider is moving under ai control (already aggroed on something before player control), then stop AI movement
```
TDSSS, line 45: Wouldn't it be better to do this in the spider's login proc? That gets called whenever a player takes over the mob regardless of source, which means it would also apply if, for example, an admin manually drags a ghost into the spider.

Kyep: Moved to mob/living/Login() so it handles that case, and also handles cases for other mobs (like headcrabs).

code/modules/mob/living/login.dm
```diff
@@ -29,4 +29,6 @@
 	//Should update regardless of if we can ventcrawl, since we can end up in pipes in other ways.
 	update_pipe_vision()
 
+	walk(src, 0) // if mob is moving under ai control, then stop AI movement
```
Fox-McCloud, line 32: Are xenos trying to run away? Are brains trying to run away?  Are human mobs running around like chickens with their heads cut off? What is OOP?

This has no reason to be in the login for all `/living` when the problem is very clearly only associated with simple mobs.

Kyep: Moved to simple_animal/Login()

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -615,3 +615,8 @@
 	if(pcollar && collar_type)
 		add_overlay("[collar_type]collar")
 		add_overlay("[collar_type]tag")
+
+/mob/living/simple_animal/Login()
+	. = ..()
```
AffectedArc07, line 620: Any reason why you did `. = ..()`? afaik `Login()` doesnt take a return value anywhere, so this seems redundant

Kyep: Changed

PR #13865, opened 2020-07-17
Port Baystruments PR, optimize and TGUI-ize musical instruments

dearmochi:
**All credit should go to silicons, the creator of the [original PR](https://github.com/tgstation/tgstation/pull/50549).** I only ported the PR they made into this codebase.

## What Does This PR Do

- **Ports the above PR. Summary:**
  - Two new subsystems: `SSinstruments` (processing) and `SSsounds`. The former processes instruments for playback. The latter manages sound channels to alleviate issues with sounds cutting each other.
  - Musical instruments are now contained in a single `instruments` module and are processed by `SSinstruments`.
  - Musical instruments are either Legacy or Synthesized. Legacy instruments are the ones who currently exist (piano, guitar, etc.). Synthesized instruments can use potentially any sound, as long as it doesn't sound too horrible.
  - Synthesized musical instruments have their own specificities, many of which can be customized in the UI.
  - Removes the **Change Synthesizer Instrument** action as that can now be performed through the Synthesizer's UI.
  - Adds the Banjo musical instrument. It can be purchased at the merch store or crafted.
  - Adds the `spooky` component that can be added to instruments. Attacking someone with a spooky instrument will weaken them gradually before transforming them into a spooky skeleton with their own spooky instrument.
- **Brings changes I've made myself:**
  - **Replaces the PR's BrowserUI by TGUI instead.**
  - **Headphones are now able to play songs. The song can only be played when the headphones are worn and can only be heard if standing on the same tile as the wearer. The "Toggle Headphones" action has been replaced by an action allowing the wearer to change the song to play.**
  - Added a "Reset to Default" button in the instrument UI to reset all advanced settings.
  - Added a custom hit sound for the acoustic guitar.
  - Added some help text about the advanced settings in the instrument UI.
  - Importing a song now stops the current playing one.
  - Headphones now have their own icon in hand.
  - A few optimizations and fixes:
    - No longer constantly update the UI every MC tick. Sending the entire song to the UI is costly. The UI is only updated when needed. This means the UI won't have a stroke when someone imports a song any more
    - When playing a note, cache `fexists` result into a static associative list. In my benchmarks it caused the brunt of Self CPU in the proc, which is now gone with the cache.
    - When playing a note, replace `hearers()` by a loop through `GLOB.player_list`. Uncontrolled mobs don't hear the sound anyway (perhaps there could be some behaviour tied to that, but...)
    - Fixed the regular accordion being blue or red depending on which hand it's being held. _Fixes #13898_

## Why It's Good For The Game

Currently instruments are.. not great. They're laggy and sound bad as a result. They make the game client have a seizure when importing a song.. the list goes on. Synthesized musical instruments are much more performant over legacy ones (which are still kept because they sound unique).

On the dev side, everything is organized in a neat folder and set of files so it's easier to tinker with the instrument code.

## Images of changes

5laWf8yX13

## Changelog
:cl:
add: Add Baystation 12 musical instrument system
add: Add the Banjo musical instrument which can be purchased in the merch store or crafted
add: Headphones are now able to play songs that can only be heard if on the same time as the wearer
add: Overhaul musical instruments UI, replacing it with TGUI greatly improving flow and performance 
tweak: Increase song line limit from 300 to 10,000 lines
fix: Major optimizations to musical instruments
soundadd: Guitar now has its own hit sound
imageadd: Fix headphones not having in-hand icons
imageadd: Fix accordion changing colour depending on the holding hand
/:cl:


tgui/packages/tgui/interfaces/Instrument.js
```diff
@@ -0,0 +1,377 @@
+/* eslint-disable max-len */
```
AffectedArc07, line 1: hmmm

dearmochi: I didn't want to put that in but had to because of the help menu, otherwise the 80 character per line limit should be respected everywhere else.

AffectedArc07: TG has this
```jsx
    <Section title="Business Exploration Protocol Incubation Sink">
      <Section 
        title="Information"
        backgroundColor="#450F44"
        buttons={(
          <Button
            icon="power-off"
            content={data.manual_power ? 'Off' : 'On'}
            selected={!data.manual_power}
            onClick={() => act(ref, 'toggle_power')} />
        )}>
        All you need to know about the B.E.P.I.S. and you!
        The B.E.P.I.S. performs hundreds of tests a second
        using electrical and financial resources to invent
        new products, or discover new technologies otherwise
        overlooked for being too risky or too niche to produce!
      </Section>
```
You can newline text without needing to go above the 80 character limit

dearmochi: I'm aware. Adding newlines results in this (note the missing spaces in some places):

image

I guess it *can* be fixed by adding `&nbsp;`:

image

image

AffectedArc07: I would much rather this than having to break the linter's line limiting

Carthusia:
yesh, YESSHHHH!!!

God i hope this gets merged ASAP

dearmochi:
@Carthusia Ah, nope. It's actually not obtainable anywhere right now but I can make it a crafting recipe/purchasable merch.

Carthusia:
oohh please do, if there are more instruments I didnt see please make them vendor purchasable like the others, im dieing to hear how the banjo sounds.

code/modules/instruments/objs/items/_instrument.dm
```diff
@@ -0,0 +1,59 @@
...
+	song = new(src, allowed_instrument_ids, instrument_range)
+	allowed_instrument_ids = null			//We don't need this clogging memory after it's used.
+
+/obj/item/instrument/Destroy()
+	QDEL_NULL(song)
+	return ..()
+
+/obj/item/instrument/suicide_act(mob/user)
+	user.visible_message("<span class='suicide'>[user] begins to play 'Gloomy Sunday'! It looks like [user.p_theyre()] trying to commit suicide!</span>")
+	return (BRUTELOSS)
```
farie82, line 28: non required ()

code/modules/instruments/objs/items/instruments.dm
```diff
@@ -153,22 +167,22 @@
 	lefthand_file = 'icons/mob/inhands/items_lefthand.dmi'
 	righthand_file = 'icons/mob/inhands/items_righthand.dmi'
 	attack_verb = list("beautifully honks")
-	instrumentId = "bikehorn"
-	instrumentExt = "ogg"
 	w_class = WEIGHT_CLASS_TINY
 	force = 0
 	throw_speed = 3
-	throw_range = 7
+	throw_range = 15
```
farie82, line 173: Why the increase here?

code/modules/instruments/songs/_song.dm
```diff
@@ -0,0 +1,401 @@
...
+
+	var/static/list/valid_files[0] // Cache to avoid running fexists() every time
+
+/datum/song/New(atom/parent, list/instrument_ids, new_range)
+	SSinstruments.on_song_new(src)
+	lines = list()
+	tempo = sanitize_tempo(tempo)
+	src.parent = parent
+	if(instrument_ids)
+		allowed_instrument_ids = islist(instrument_ids)? instrument_ids : list(instrument_ids)
```
farie82, line 138: spacing here

Fox-McCloud:
I tested this out locally---and I couldn't get anything to play (at least on the piano synth); I toggled prefs on and off for instruments, but I still couldn't hear jack---did ya test this?

code/modules/instruments/objs/items/instruments.dm
dearmochi: Seems to have been overwritten by [TG's throw range](https://github.com/tgstation/tgstation/blob/1aa293ea337283a0191140a878eeba319221e5df/code/game/objects/items/devices/instruments.dm#L289), I'll fix it

code/controllers/subsystem/processing/instruments.dm
```diff
@@ -0,0 +1,86 @@
+PROCESSING_SUBSYSTEM_DEF(instruments)
+	name = "Instruments"
+	init_order = INIT_ORDER_INSTRUMENTS
+	wait = 0.5
+	flags = SS_KEEP_TIMING
+	offline_implications = "Instruments will no longer play. No immediate action is needed."
+	stat_tag = "Instruments"
```
AffectedArc07, line 7: This isnt needed, we can just have `P: x` on the statpanel for instruments, since everything else follows this format.

code/controllers/subsystem/sounds.dm
```diff
@@ -0,0 +1,165 @@
...
+	return ..()
+
+/**
+  * Sets up all available sound channels
+  */
+/datum/controller/subsystem/sounds/proc/setup_available_channels()
+	channel_list = list()
+	reserved_channels = list()
+	using_channels = list()
+	using_channels_by_datum = list()
```
AffectedArc07, line 38: Initialize these in the definition, or use a lazy list here

code/controllers/subsystem/sounds.dm
```diff
@@ -0,0 +1,165 @@
...
+  * Removes a channel from using list
+  *
+  * Arguments:
+  * * channel - The channel number
+  */
+/datum/controller/subsystem/sounds/proc/free_sound_channel(channel)
+	var/text_channel = num2text(channel)
+	var/using = using_channels[text_channel]
+	using_channels -= text_channel
+	if(using != TRUE)		// datum channel
```
AffectedArc07, line 54: `if(!using)`

code/controllers/subsystem/sounds.dm
```diff
@@ -0,0 +1,165 @@
...
+		return
+	var/channel = channel_list[channel_reserve_high]
+	reserved_channels[num2text(channel)] = channel_reserve_high--
+	return channel
+
+/**
+  * Frees a channel and updates the datastructure. Private proc.
+  */
+/datum/controller/subsystem/sounds/proc/free_channel(number)
+	PRIVATE_PROC(TRUE)
```
AffectedArc07, line 126: For anyone wondering, this doesnt make the proc itself private (BYOND isnt smart enough for that), but will make the linter cry out in agony if you try use this proc outside of the datum.

code/controllers/subsystem/processing/instruments.dm
```diff
@@ -0,0 +1,86 @@
+PROCESSING_SUBSYSTEM_DEF(instruments)
+	name = "Instruments"
+	init_order = INIT_ORDER_INSTRUMENTS
+	wait = 0.5
```
AffectedArc07, line 4: I am a little concerned with having a subsystem fire this quickly

dearmochi: There won't be that many instruments playing at once (maybe 5 tops) and the note timing is dependent on `world.tick_lag` (which is `0.5` by default I think?). Setting the subsystem delay to be higher than `world.tick_lag` would result in notes playing later than expected

code/controllers/subsystem/sounds.dm
dearmochi: Not sure it's worth it? Sounds are played basically immediately so it would barely make a difference if these lists were lazily initialized. I don't see any harm in leaving these here

dearmochi:
Stamped out all outstanding issues, good to go 👍

Carthusia:
hnnnnnnnnnnnnnnnnnnnnnnnnnnng cant wait to see this ingame, is the banjo and other new instruments buyable at the vendor?

dearmochi:
@Carthusia the banjo is buyable and craftable

Kyep:
Testmerge issues:

[2020-07-22T04:59:55] Runtime in play_legacy.dm,92: Cannot read null.prefs
   proc name: playkey legacy (/datum/song/proc/playkey_legacy)
   src: (/datum/song/handheld)
   call stack:
   (/datum/song/handheld): playkey legacy(5, "b", 5, XXXXXX (/mob/living/carbon/human))
   (/datum/song/handheld): play chord(/list (/list))
   (/datum/song/handheld): process song(0.5)
   (/datum/song/handheld): process(0.5)
   Instruments (/datum/controller/subsystem/processing/instruments): fire(0)
   Instruments (/datum/controller/subsystem/processing/instruments): ignite(0)

Carthusia:
So far during testing ive noticed the guitar is alot quieter than it used to be, and the banjo overpowers most everything else

Mitchs98:
After playing the Grand Piano I want nothing more in life at the moment than for this to get merged as soon as possible 😍

Carthusia:
I never got a chance to try the grand piano, but i heard it, and it sounds amazing. Cannot wait to try it with some of my piano designated songs.

dearmochi:
@Carthusia If you'd like to test it in a private server, I can start one up and send you my address, just hit me up on Discord: `mochi#5162`

dearmochi:
Private testing uncovered more issues that I'll fix shortly.

dearmochi:
And we're ready. The PR should be at its' final version. Check the changelog for the final additions.

Carthusia:
God i hope this is merged soon, this is a god send in terms of performance.

I got 16gb of ram and to load a longer song, quad core CPU, and trying to load a longer midi sheet would cuase my CPU to chug in the old system.

Not with this one, its instant, no lag, no fuss.

SteelSlayer:
Can you add "Fixes #13898" to your PR description?

AffectedArc07:
> God i hope this is merged soon, this is a god send in terms of performance.
> 
> I got 16gb of ram and a quad core CPU, and trying to load a longer midi sheet would cuase my CPU to chug in the old system.
> 
> Not with this one, its instant, no lag, no fuss.

Worth nothing. BYOND is a single-threaded 32 bit app. At most youre using 1.8GB of RAM and a singular CPU core. Everyone chugs, your core count means nothing to BYOND.

Carthusia:
Then beyond you fickle mistress, stop having a stroke lol

Wasn't aware how little hardware meant to Beyond itself, no matter how good anything else gets, beyond stays in its little time bubble i guess

code/controllers/subsystem/sounds.dm
AffectedArc07: Having this in a separate proc from `Initialize` had me under the impression that this was called in multiple places, hence me requesting it be lazy-ified. This is only called in `Initialize`, so I question why you made this into an extra proc.

Pyrodes:
grn must be added now

code/controllers/subsystem/processing/instruments.dm
```diff
@@ -0,0 +1,86 @@
...
+	flags = SS_KEEP_TIMING
```
AffectedArc07, line 5: Make this a mix of `SS_TICKER | SS_BACKGROUND | SS_KEEP_TIMING`, and set `wait` to `1` so its logged properly in ticks. `SS_BACKGROUND` will then also bump the priority to normal (`SS_TICKER` will treat wait as tick value, but also bump priority, so you use background to bump it out)

code/datums/action.dm
```diff
@@ -232,30 +229,13 @@
...
 	name = "Vortex Recall"
 	desc = "Recall yourself, and anyone nearby, to an attuned hierophant beacon at any time.<br>If the beacon is still attached, will detach it."
 	button_icon_state = "vortex_recall"
 
-/datum/action/item_action/vortex_recall/IsAvailable()
-	if(istype(target, /obj/item/hierophant_club))
-		var/obj/item/hierophant_club/H = target
-		if(H.teleporting)
-			return 0
-	return ..()
```
AffectedArc07, line 258: Why did you delete this?

AffectedArc07 (changes requested):

code/datums/action.dm
dearmochi: I have absolutely no idea how this got here.

AffectedArc07 (approved):
If this goes horribly wrong we have a revert button

taukausanake:
Lefthand headphones have incorrect sprites. North and South sprites are the same
image
The others look fine though

PR #13874, opened 2020-07-19
Terror spider event types now depend on server pop

Kyep:
## What Does This PR Do
Terror spider events now consider server pop when deciding what type of spiders to spawn.
On lowpop (<80 clients), it picks from: greens/princes/whites/princesses.
On highpop (80+) it picks from: prince/whites/princesses/queen.
This means greens (weakest infestation type) are never seen on highpop, and queen (strongest infestation type) are never seen on lowpop.

## Why It's Good For The Game
Helps ensure that terror events are more scaled to server pop.

## Changelog
:cl: Kyep
balance: Terror spider events now have their severity scaled by server pop.
/:cl:

### DIFF
--- a/code/modules/events/spider_terror.dm
+++ b/code/modules/events/spider_terror.dm
@@ -41,7 +41,7 @@
 			spawncount = 2
 		if(4)
 			// Pretty strong.
-			spider_type = /mob/living/simple_animal/hostile/poison/terror_spider/queen/princess
+			spider_type = /mob/living/simple_animal/hostile/poison/terror_spider/princess
 			spawncount = 2
 		if(5)
 			// Strongest, only used during highpop.
--- a/code/modules/events/spider_terror.dm
+++ b/code/modules/events/spider_terror.dm
@@ -71,3 +71,4 @@
 			spawncount--
 
 #undef TS_HIGHPOP_TRIGGER
+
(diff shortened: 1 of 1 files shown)


TDSSS:
Less Greens that never get going is definitely nice. Might even be nice to have a weighting system based on pop.

code/modules/events/spider_terror.dm
```diff
@@ -22,23 +22,31 @@
 			if(temp_vent.parent.other_atmosmch.len > 50)
 				vents += temp_vent
 	var/spider_type
-	var/infestation_type = pick(1, 2, 3, 4, 5)
+	var/highpop_trigger = 80
```
farie82, line 25: best to make this a local define. Be sure to undef it at the end of the file

code/modules/events/spider_terror.dm
```diff
@@ -22,23 +22,31 @@
...
+	var/infestation_type = pick(1, 2, 3, 4)
```
farie82, line 26: best to put the pick in the else of the if. Slight performance stuff but mostly for consistency

code/modules/events/spider_terror.dm
Kyep: Why? That would make it 3 lines rather than 1, for a var that's only used once.

farie82: A define is usually something that you first notice when opening the file since it's at the top. It also stands out. Performance wise it's a tiny improvement but nothing to write home about.
It is mostly for consistency with how it's done in other parts of the codebase. Variables can be used if they can be changeable through inheritance or admemery and such. (only works if they are defined in the object though. Not inside a proc)

PR #13879, opened 2020-07-20
Prevents slimes from feeding on anyone in a Hardsuit or Biosuit

warior4356:
## What Does This PR Do
Adds a check to slimes that prevents them from latching onto a Hardsuit, biosuit, or any suit with the flag THICKMATERIAL. 

## Why It's Good For The Game
After watching an ERT get messed up by slimes it felt needed. Unlike most damage types their continual TOX and CLONE damage is really hard to resist and even harder to cure. Additionally, it makes it so biosuits will actually protect Xenobiologists from the slimes. They can still be attacked, but if they take the time to wear the suit they will be protected.

## Changelog
:cl:
tweak: Slimes can no longer feed on mobs wearing biosuits or hardsuits.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/slime/powers.dm
+++ b/code/modules/mob/living/simple_animal/slime/powers.dm
@@ -99,6 +99,14 @@
 			return FALSE
 		to_chat(src, "<span class='warning'><i>Another slime is already feeding on this subject...</i></span>")
 		return FALSE
+
+	if(iscarbon(M))
+		var/mob/living/carbon/C = M
+		if((C.wear_suit && ((C.wear_suit.flags & THICKMATERIAL)) && ((C.head && ((C.head.flags & THICKMATERIAL))))))
+			if(silent)
+				return FALSE
+			to_chat(src, "<span class='warning'><i>This subject's covering is too thick...</i></span>")
+			return FALSE
 	return TRUE
 
 /mob/living/simple_animal/slime/proc/Feedon(mob/living/M)
--- a/code/modules/mob/living/simple_animal/slime/powers.dm
+++ b/code/modules/mob/living/simple_animal/slime/powers.dm
@@ -102,7 +102,7 @@
 
 	if(iscarbon(M))
 		var/mob/living/carbon/C = M
-		if((C.wear_suit && ((C.wear_suit.flags & THICKMATERIAL)) && ((C.head && ((C.head.flags & THICKMATERIAL))))))
+		if(C.wear_suit && (C.wear_suit.flags & THICKMATERIAL) && C.head && (C.head.flags & THICKMATERIAL))
 			if(silent)
 				return FALSE
 			to_chat(src, "<span class='warning'><i>This subject's covering is too thick...</i></span>")


code/modules/mob/living/simple_animal/slime/powers.dm
```diff
@@ -99,6 +99,14 @@
 			return FALSE
 		to_chat(src, "<span class='warning'><i>Another slime is already feeding on this subject...</i></span>")
 		return FALSE
+
+	if(iscarbon(M))
+		var/mob/living/carbon/C = M
+		if((C.wear_suit && ((C.wear_suit.flags & THICKMATERIAL)) && ((C.head && ((C.head.flags & THICKMATERIAL))))))
```
farie82, line 105: the amount of parenthesis make it a bit hard to read. Can just write it like this:
`suit && (suit & THICK) && head && (head & THICC)`

code/modules/mob/living/simple_animal/slime/powers.dm
```diff
@@ -99,6 +99,14 @@
...
+			if(silent)
+				return FALSE
+			to_chat(src, "<span class='warning'><i>This subject's covering is too thick...</i></span>")
```
farie82, line 108: Why the `<i>`?

code/modules/mob/living/simple_animal/slime/powers.dm
```diff
@@ -99,6 +99,14 @@
 			return FALSE
 		to_chat(src, "<span class='warning'><i>Another slime is already feeding on this subject...</i></span>")
 		return FALSE
+
+	if(iscarbon(M))
+		var/mob/living/carbon/C = M
+		if((C.wear_suit && ((C.wear_suit.flags & THICKMATERIAL)) && ((C.head && ((C.head.flags & THICKMATERIAL))))))
+			if(silent)
```
farie82, line 106: I'd just make this into a `!silent` and put the to_chat in the if scope

taukausanake:
Yeah it makes sense with a biosuit and by comparison a hardsuit but now slimes are going to be even more easy mode with this protection. Xenobiologists already never need to touch them or even go near them. Now this means they can be completely disregarded. There needs to be some downsides or potential danger when working around slimes with all these safeguards put into place

Denthamos:
While I do like the idea of hardsuits protecting as it makes sense and by proxy bio suits due to _what_ the slimes do damage wise @taukausanake does have a point with making it pretty much an easy mode for xenobios with no risk to themselves when it comes to starved slimes.  Honestly on the fence about this.

farie82:
> While I do like the idea of hardsuits protecting as it makes sense and by proxy bio suits due to _what_ the slimes do damage wise @taukausanake does have a point with making it pretty much an easy mode for xenobios with no risk to themselves when it comes to starved slimes. Honestly on the fence about this.

Xenobiology currently already has 0 threat as long as you keep the slimes in the cages.
The slimes should still be able to attack you I recon so they will still kill you if they are hungry. They will just be sucking on a hardsuit instead of your insides

warior4356:
> While I do like the idea of hardsuits protecting as it makes sense and by proxy bio suits due to _what_ the slimes do damage wise @taukausanake does have a point with making it pretty much an easy mode for xenobios with no risk to themselves when it comes to starved slimes. Honestly on the fence about this.

Not zero threat as a slime with either attack with a punch or feed. They can still beat you pretty heavily if you're not careful. This just prevents the damage that is really hard to remove (Clone). Remember that bio suits really slow you down as a trade-off, so you are working slower but safer.

code/modules/mob/living/simple_animal/slime/powers.dm
warior4356: I'd rather keep it the same as the other conditions.

code/modules/mob/living/simple_animal/slime/powers.dm
warior4356: I was being consistent with the other conditions.

code/modules/mob/living/simple_animal/slime/powers.dm
warior4356: Yea that is better, I'll make that change.

Fox-McCloud:
This pushes slimes into being even more pathetic than they already are.

Also slimes are *pathetically* easy to fight. A simple disarm flings them off and a few water sprays will kill them.

We should ***not*** be appealing to the lowest common denominator of player/player capability/player intelligence; this is a knowledge problem and either can only be corrected with education....orrrr not at all because some people just refuse to learn/read the wiki.

This has been a non-issue since para inception--it's always been a "well then learn to fight a slime, eh?" situation. I see no reason to diverge from that policy.

TDSSS:
I thought the slime still attacks, it just doesnt feed? Trying to do hijack where you get constantly punched by slimes seems unlikely (besides, slimes are friendly if you've fed them anyway, right?)

I don't think this will do much one way or another

taukausanake:
I think I would go for something like this if clothing and suits had health values. A slime can still attack you but it'll have to get through the outer shell before you start taking damage. Normal clothing wouldn't protect anything, biosuit would protect for a little maybe enough to throw the slime off, hardsuit for longer but it will still be damaged. That would require a refactor however and no one's got time for that

variableundefined:
Objection on the same ground as Fox and Kyet

variableundefined:
Closing due to 2 maint + 1 head objection

PR #13884, opened 2020-07-20
Adds a world/Topic entry for announcing by server

AffectedArc07:
## What Does This PR Do
This PR adds a world/Topic entry for automated announcements to be sent to the server via internal tooling.
image
This allows the playerbase to be notified of impending restarts, testmerges or more.

## Why It's Good For The Game
People should be informed

## Changelog
:cl: AffectedArc07
add: Added the ability for announcements to be sent to the server via server tooling
/:cl:

### DIFF
--- a/tools/send2server.py
+++ b/tools/send2server.py
@@ -1,0 +1,14 @@
+# Script to send world/Topic announcements to a server via automated tooling
+# Author: AffectedArc07
+# Takes a message as command line argument
+# Example: python send2server.py "This is a message that will be sent to the server"
+import socket, struct, urllib.parse, sys
+
+message = sys.argv[1]
+commskey = "YOURKEYHERE"
+
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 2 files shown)


tools/send2server.py
```diff
@@ -0,0 +1,14 @@
...
+# Author: AffectedArc07
+# Takes a message as command line argument
+# Example: python send2server.py "This is a message that will be sent to the server"
+import socket, struct, urllib.parse, sys
+
+message = sys.argv[1]
+commskey = "YOURKEYHERE"
+
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+cmd = "?hostannounce&key={}&message={}".format(urllib.parse.quote(commskey), urllib.parse.quote(message))
```
Crazylemon64, line 11: Will sending the auth key via plaintext be an issue here?

AffectedArc07: I am unsure, but it works with and without the url encoding, so I threw it in just as a precaution.

PR #13886, opened 2020-07-20
Removes xray lasers from R&D, replaces it with the L.W.A.P. sniper rifle

SteelSlayer:
## What Does This PR Do
This is @Qwertyotoforty's code and these are his words, I just copied the changes and made the PR.

This PR removes xray from the protolathe, and replaces it with the LWAP sniper, while putting xray lasers in the gamma armory and removing the LWAPs from the gamma armory. Creating the sniper rifle has the same exact tech requirements as the xray laser.

## Why It's Good For The Game
Xray lasers have 10 shots, do 15 damage, decreased by 0.75 damage per tile. They also irradiate anyone hit for 30 points, leading to huge toxin and burn damage after 3-4 hits. They also of course, go through anything really, irradiating whoever they hit, going through walls, and of course, dunking even late game blobs. Crew can naturally mass produce these, and really, compared to normal lasers guns, they don’t have a downside apart from the decreased damage radiation more than makes up for. Based on suggestions of several people who do not think crew should have access to xray type objects normally, this removes xray lasers from the protolathe and puts it in the gamma armory, while moving the LWAP to protolathe instead.

Stat wise, the LWAP has and is:
Bulky, and requires 2 hands to fire.
Has 10 shots
Does 60 damage burn with a 10 second stun with each shot, but has a 4 second cooldown between shots.
Has a zoom function.

This weapon does seem strong at first, but it has a few catches. It is bulky, you are not easily able to carry it around, unless you wear it on armour, in your hand, or in your backpack slot. Since it is a 2 handed weapon, unlike an AEG, you can not have a baton or riot shield in your off hand, you need to have it empty to fire. Yes, the weapon does do a high 60 damage shot with a 10 second stun, but you could also use a taser and a laser gun, for a similar stun, but way more damage in a shorter time. Unlike the LWAP as well, a taser can be fired every second, instead of the 4 second cool down of the LWAP, where if you miss, you are probably screwed. And of course, you could also use a taser and a shotgun as well, for even stronger effects in the more popular brute damage.

LWAP’s tech level value is also increased to xrays value, to not change any R&D pathings.

## Changelog
:cl:
tweak: Xray lasers have been removed from the protolathe design list
add: L.W.A.P Sniper rifles are now buildable from protolathes with the same tech requirements as the now removed xray lasers
tweak: Replaces the L.W.A.P sniper rifles in the gamma armory with xray lasers
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/z2.dmm
+++ b/_maps/map_files/cyberiad/z2.dmm
@@ -5193,9 +5193,6 @@
 /area/shuttle/gamma/space)
 "oC" = (
 /obj/structure/closet/secure_closet/guncabinet,
-/obj/item/gun/energy/sniperrifle,
-/obj/item/gun/energy/sniperrifle,
-/obj/item/gun/energy/sniperrifle,
 /obj/machinery/light/spot{
 	tag = "icon-tube1 (EAST)";
 	icon_state = "tube1";
@@ -5204,6 +5201,9 @@
 /obj/machinery/ai_status_display{
 	pixel_y = 32
 	},
+/obj/item/gun/energy/xray,
+/obj/item/gun/energy/xray,
+/obj/item/gun/energy/xray,
 /turf/simulated/shuttle/floor{
 	icon_state = "floor4"
 	},
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -310,7 +310,7 @@
 	name = "L.W.A.P. Sniper Rifle"
 	desc = "A rifle constructed of lightweight materials, fitted with a SMART aiming-system scope."
 	icon_state = "esniper"
-	origin_tech = "combat=6;materials=5;powerstorage=4"
+	origin_tech = "combat=6;materials=4;magnets=4;syndicate=1"
 	ammo_type = list(/obj/item/ammo_casing/energy/sniper)
 	item_state = null
 	weapon_weight = WEAPON_HEAVY
--- a/code/modules/research/designs/weapon_designs.dm
+++ b/code/modules/research/designs/weapon_designs.dm
@@ -236,15 +236,15 @@
 	build_path = /obj/item/ammo_casing/shotgun/techshell
 	category = list("Weapons")
 
-/datum/design/xray
-	name = "Xray Laser Gun"
-	desc = "Not quite as menacing as it sounds"
-	id = "xray"
+/datum/design/sniperrifle
+	name = "L.W.A.P. Sniper Rifle"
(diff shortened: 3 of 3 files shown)


Shockpoint:
**Please.**

Mitchs98:
I don't at all see x-ray rifles used that much anymore, especially not outside of blob. And even with blob in the vast majority of cases I see RnD print AEGs instead of xray rifles. Once every handful of blob rounds I see an xray laser, but outside of blob I've never really seen xrays be used. I don't see a problem in keeping it, it's a niche gun that's useful mostly in one round type and even then isn't often printed.

I certainly don't see it used nor abused enough to warrant it being a problem. 

LWAP is more of a gimmick item in my opinion and experience. It looks cool, sounds cool, but is outclassed by nearly every weapon you can use.

Fox-McCloud:
I'm not  entirely opposed to removing x-ray laserguns, but LWAP is going to be utilized for more than just a gun; people are going to want to carry them around for their scope ability, and *that* I'm not particularly keen on.

Qwertytoforty:
LWAP now needs to be in hands for the zoom function from the last LWAP PR, but I do see your point.

Rurik123:
Are xrays overperforming? I've not been entirely active as of late but the earlier months of this year, from what I could tell, xrays weren't being used too terribly much and when they were blobs weren't devastated by it. Mostly due to the new buffs blob has, though I suppose said buffs aren't exactly new anymore.

Definitely agree with what's been said about the LWAP. Fun to use, but gimmick mostly. _Mostly._ Either way would be fun to get to use it more.

ppi13:
LWAP will be used far more then the xray laser ever will be, you got a gun that with a single shot from max range is basically a killshot, since you can fire the next one 4 seconds later, with a guy pretty much unable to dodge.
LWAP is only strong against mobs, while the x-ray laser is strong against multiple mobs, such as blob, and this here will be a hefty buff to blob which is already in a really good place right now, while x-ray laser is not the win button of blob, it is absurdly powerfull if used correctly against it.

Some adjustments to make up for this, an alternative to blobs is required before this is something everyone can accept, assuming the ability to have a LWAP every round is fine.

variableundefined:
Objection. 

Agrees with Fox on the scope issue. And from what I see, X-ray are no longer the cure all against blob it used to be.

AffectedArc07:
People will use these, not for the shooting power, but just to scope down hallways for the sake of more vision. Combine that with the xray nerf which went through, I dont see this PR being useful. 

As it stands, this PR is at 2 maintainer objections and a head objection, and no one in favour from either party. Closing PR.

PR #13894, opened 2020-07-21
Fix AI teleporting tickets to its core

Nazsgull:
## What Does This PR Do
Fixes AI interacting with the ticket machine.

## Why It's Good For The Game
AI will no longer teleport tickets to its core when interacting with the ticket machine.

## Changelog
:cl: Nazsgull
del: AI can no longer interact with ticket machine.
fix: AI can no longer teleport tickets to its core. 
/:cl:

### DIFF
--- a/code/modules/paperwork/ticketmachine.dm
+++ b/code/modules/paperwork/ticketmachine.dm
@@ -169,7 +169,7 @@
 
 /obj/item/ticket_machine_ticket
 	name = "Ticket"
-	desc = "A ticket which shows your place in the Head of Personnel's line. Made from Nanotrasen patented NanoPaper®. Though solid, its form seems to shimmer slightly. Feels (and burns) just like the real thing."
+	desc = "A ticket which shows your place in the Head of Personnel's line. Made from Nanotrasen patented NanoPaperÂ®. Though solid, its form seems to shimmer slightly. Feels (and burns) just like the real thing."
 	icon = 'icons/obj/bureaucracy.dmi'
 	icon_state = "ticket"
 	maptext_x = 7
--- a/code/modules/paperwork/ticketmachine.dm
+++ b/code/modules/paperwork/ticketmachine.dm
@@ -167,6 +167,8 @@
 		user.adjust_fire_stacks(1)
 		user.IgniteMob()
 
+/obj/machinery/ticket_machine/attack_ai(mob/user)//AI cannot interact with ticket_machine
+	return
 /obj/item/ticket_machine_ticket
 	name = "Ticket"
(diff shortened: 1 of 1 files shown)


Ty-Omaha:
Here I am thinking there was some sort of real AI like a neural network or something.

code/modules/paperwork/ticketmachine.dm
```diff
@@ -167,6 +167,8 @@
 		user.adjust_fire_stacks(1)
 		user.IgniteMob()
 
+/obj/machinery/ticket_machine/attack_ai(mob/user)//AI cannot interact with ticket_machine
+	return
 /obj/item/ticket_machine_ticket
```
SteelSlayer, line 172: Usually you want to put any procs *after* the object definition, so something like: 
```js
/obj/item/ticket_machine_ticket
    name = "Ticket"
    ... // other stuff here, description, etc.

/obj/machinery/ticket_machine/attack_ai(mob/user)//AI cannot interact with ticket_machine
	return
```

It just looks better that way. Also it's good to have a blank line between every proc or path

Fox-McCloud: Yep, for sure.

code/modules/paperwork/ticketmachine.dm
```diff
@@ -167,9 +167,11 @@
...
 	name = "Ticket"
-	desc = "A ticket which shows your place in the Head of Personnel's line. Made from Nanotrasen patented NanoPaper�. Though solid, its form seems to shimmer slightly. Feels (and burns) just like the real thing."
+	desc = "A ticket which shows your place in the Head of Personnel's line. Made from Nanotrasen patented NanoPaper®. Though solid, its form seems to shimmer slightly. Feels (and burns) just like the real thing."
```
Fox-McCloud, line 174: What changed here?

Nazsgull: Encoding got changed, I changed it back but I had removed the corrupted symbol, so I put it back.

code/modules/paperwork/ticketmachine.dm
Nazsgull: Yeah but the method I added is for the machine, so it goes after the other machine methods, not near the ticket ones.

SteelSlayer: You're right, I mixed up the two similar paths here. So the way you have it is fine. The blank line after the return would still be nice though.

Fox-McCloud: still has spacing issues.

code/modules/paperwork/ticketmachine.dm
Fox-McCloud: I'd prefer this not show up at all.

PR #13896, opened 2020-07-21
Fixes issues with machines/mechs when a person is teleported out of them

### LINKED ISSUE #13132: Nanomachine Cyborgification Breaks Scanners
**Issue Description**:
Related to #4191, though not quite as that issue has been fixed. Rather than the person being unable to exit the scanner, the person is now spawned outside of the scanner however the scanner itself is now permanently occupied and the 'contents' unable to be ejected. It's not too major, as screwdrivering the scanner DOES fix it, it just probably shouldn't happen in the first part. 

I expected to eat a roburger, enter the scanner, and pop out with the scanner at-least being able to be ejected afterwards. 

**What happened instead**:
I ate the roburger, transformation occured, but the scanner was 'occupied' until able to be hit with  a screwdriver.

It's very much an edge case but if no one doesn't think to try and deconstruct the machine it could end up with one of medbays important pieces of equipments being entirely nonfunctional.

1. Eat roburger.
2. Enter scanner.
3. Spawn as borg.
4. Spam eject on the scanner and realize it doesn't work and is permanently 'occupied'.

Presumably since whatever PR fixed #4191

**Extra information**:

Kyep:
## What Does This PR Do
Fixes issues created when a human is teleported directly out of a machine/object without the object being told.
Fixes #13132 - similar issue created when a human is transformed into a robot/xeno/etc by a disease while inside a machine.

Teleport types addressed in this PR:
- auto-cryo for SSD/AFK players sending someone to cryo
- admin teleport effects (get mob, send to area, jump to key, jump to ruin, etc)

Machines/objects addressed in this PR:
- sleepers
- cryopods
- bodyscanners
- DNA modifiers
- suit storage units
- lockers
- mechs
IE: all the common things humans might be found inside.
There is also a catch-all which will generate an informative exception for the log in case a human is teleported directly out of any other un-handled object by the above methods.

For those objects, it fixes:
- them getting stuck in a 'occupied' state where nobody can use them
- them having hard references to the teleported mob that are never cleared
- them generating runtimes due to thinking they have a mob in them when they don't

Notes:
- this doesn't un-DNA-lock a mech in cases where a pilot is removed from the mech. Mostly because that would mean admins teleporting a mech pilot would let someone else steal the mech.

## Why It's Good For The Game
Fewer bugs.

## Changelog
:cl: Kyep
fix: Fixed issues occurring when a player is teleported out of an object (by SSD/AFK system, or admin teleport) without the object being told.
fix: Fixed issues occurring when a player is transformed into a new type of creature while inside a machine.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -772,12 +772,9 @@
 /proc/cryo_ssd(var/mob/living/carbon/person_to_cryo)
 	if(istype(person_to_cryo.loc, /obj/machinery/cryopod))
 		return 0
-	if(istype(person_to_cryo.loc, /obj/mecha))
-		var/obj/mecha/E = person_to_cryo.loc
-		E.go_out()
-	if(istype(person_to_cryo.loc, /obj/machinery))
-		var/obj/machinery/E = person_to_cryo.loc
-		E.force_eject_occupant()
+	if(istype(person_to_cryo.loc, /obj))
+		var/obj/O = person_to_cryo.loc
+		O.force_eject_occupant()
 	var/list/free_cryopods = list()
 	for(var/obj/machinery/cryopod/P in GLOB.machines)
 		if(!P.occupant && istype(get_area(P), /area/crew_quarters/sleep))
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -772,7 +772,7 @@
 /proc/cryo_ssd(var/mob/living/carbon/person_to_cryo)
 	if(istype(person_to_cryo.loc, /obj/machinery/cryopod))
 		return 0
-	if(istype(person_to_cryo.loc, /obj))
+	if(isobj(person_to_cryo.loc))
 		var/obj/O = person_to_cryo.loc
 		O.force_eject_occupant()
 	var/list/free_cryopods = list()
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -768,9 +768,16 @@
 
 	return ..()
 
+
 /proc/cryo_ssd(var/mob/living/carbon/person_to_cryo)
 	if(istype(person_to_cryo.loc, /obj/machinery/cryopod))
 		return 0
+	if(istype(person_to_cryo.loc, /obj/mecha))
+		var/obj/mecha/E = person_to_cryo.loc
+		E.go_out()
+	if(istype(person_to_cryo.loc, /obj/machinery))
+		var/obj/machinery/E = person_to_cryo.loc
+		E.force_eject_occupant()
 	var/list/free_cryopods = list()
 	for(var/obj/machinery/cryopod/P in GLOB.machines)
 		if(!P.occupant && istype(get_area(P), /area/crew_quarters/sleep))
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -740,10 +740,11 @@
 	if(!occupant)
 		return
 
-	if(user != occupant)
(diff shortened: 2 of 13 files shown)


code/game/machinery/cryopod.dm
```diff
@@ -768,9 +768,13 @@
 
 	return ..()
 
+
 /proc/cryo_ssd(var/mob/living/carbon/person_to_cryo)
 	if(istype(person_to_cryo.loc, /obj/machinery/cryopod))
 		return 0
+	if(istype(person_to_cryo.loc, /obj))
```
SteelSlayer, line 775: We have an `isobj(...)` define to do this. I'm not sure if that's preferred or not, so up to you if you want to keep it like this or use the define.

code/modules/admin/verbs/adminjump.dm
```diff
@@ -111,6 +111,10 @@
 
 	log_admin("[key_name(usr)] teleported [key_name(M)]")
 	message_admins("[key_name_admin(usr)] teleported [key_name_admin(M)]", 1)
+
+	if(istype(M.loc, /obj))
```
SteelSlayer, line 115: Can also use the helper here

code/modules/admin/verbs/adminjump.dm
```diff
@@ -135,6 +139,9 @@
 	log_admin("[key_name(usr)] teleported [key_name(M)]")
 	message_admins("[key_name_admin(usr)] teleported [key_name(M)]", 1)
 	if(M)
+		if(istype(M.loc, /obj))
```
SteelSlayer, line 142: And here

Mitchs98:
@Kyep I think this might also fix https://github.com/ParadiseSS13/Paradise/issues/13132

I am unsure personally, but it seems like something it would fix.

Kyep:
@Mitchs98 it doesn't. It could, though, if I modified the nanomachine/xenomicrobe code so that, before transformation, it checked if you were inside an object and then applied the method this fix uses for the other cases.

Kyep:
Updated to also fix that issue as well.

code/game/objects/objs.dm
```diff
@@ -351,3 +351,9 @@ a {
 
 /obj/proc/check_uplink_validity()
 	return TRUE
+
+/obj/proc/force_eject_occupant()
+	// This proc handles safely removing occupant mobs from the object if they must be teleported out (due to being SSD/AFK, by admin teleport, etc) or transformed.
+	// In the event that the object doesn't have an overriden version of this proc to do it, log a runtime so one can be added.
+	log_runtime(EXCEPTION("Proc force_eject_occupant() is not overriden on a machine containing a mob."), src)
```
AffectedArc07, line 358:
```suggestion
	CRASH("Proc force_eject_occupant() is not overriden on a machine containing a mob.")
```

code/game/machinery/suit_storage_unit.dm
```diff
@@ -740,17 +740,20 @@
...
 
-	if(user != occupant)
-		to_chat(occupant, "<span class='warning'>The machine kicks you out!</span>")
-	if(user.loc != loc)
-		to_chat(occupant, "<span class='warning'>You leave the not-so-cozy confines of the SSU.</span>")
+	if(user)
+		if(user != occupant)
+			to_chat(occupant, "<span class='warning'>The machine kicks you out!</span>")
+		if(user.loc != loc)
+			to_chat(occupant, "<span class='warning'>You leave the not-so-cozy confines of the SSU.</span>")
```
AffectedArc07, line 747:
```suggestion
			to_chat(occupant, "<span class='warning'>You leave the not-so-cozy confines of [src].</span>")
```

PR #13899, opened 2020-07-22
Update flashbang FX and code clean-up

dearmochi:
## What Does This PR Do

- **Removes use of `set_light` in flashbang and the delayed `qdel` and instead creates a `/obj/effect/dummy/lighting_obj` to handle the area illumination and `qdel`s immediately.**
- Increases flashbang detonation volume from `25` to `100` - it's a flashbang, not a silent fart
- Refactors and cleans up `flashbang.dm` for readability

## Why It's Good For The Game

Flashbangs should be louder and use less poor quality code

## Changelog
:cl:
tweak: Increase flashbang volume
tweak: Replace flashbang VFX by one that is less error-prone
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/flashbang.dm
+++ b/code/game/objects/items/weapons/grenades/flashbang.dm
@@ -5,39 +5,51 @@
 	origin_tech = "materials=2;combat=3"
 	light_power = 10
 	light_color = LIGHT_COLOR_WHITE
-	var/light_time = 2
-	var/range = 7
+
+	var/light_time = 0.2 SECONDS // The duration the area is illuminated
+	var/range = 7 // The range in tiles of the flashbang
 
 /obj/item/grenade/flashbang/prime()
 	update_mob()
-	var/flashbang_turf = get_turf(src)
-	if(!flashbang_turf)
-		return
+	var/turf/T = get_turf(src)
+	if(T)
+		// VFX and SFX
+		do_sparks(rand(5, 9), FALSE, src)
+		playsound(T, 'sound/effects/bang.ogg', 100, TRUE)
+		new /obj/effect/dummy/lighting_obj(T, light_color, range + 2, light_power, light_time)
 
-	set_light(7)
+		// Stunning & damaging mechanic
+		bang(T, src, range)
+	qdel(src)
 
-	do_sparks(rand(5, 9), FALSE, src)
-	playsound(flashbang_turf, 'sound/effects/bang.ogg', 25, 1)
-	bang(flashbang_turf, src, range)
-
-	for(var/obj/structure/blob/B in hear(8, flashbang_turf))     		//Blob damage here
-		var/damage = round(30 / (get_dist(B, get_turf(src)) + 1))
-		B.take_damage(damage, BURN, "melee", 0)
-
-	spawn(light_time)
-		qdel(src)
-
+/**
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/grenades/flashbang.dm
```diff
@@ -5,39 +5,42 @@
...
+		var/damage = round(30 / (get_dist(B, T) + 1))
+		B.take_damage(damage, BURN, "melee", FALSE)
+
+	qdel(src)
 
 /proc/bang(turf/T, atom/A, range = 7, flash = TRUE, bang = TRUE)
+	var/source_turf = get_turf(A)
 	for(var/mob/living/M in hearers(range, T))
-		if(M.stat == DEAD)
+		if(M.is_dead())
```
Fox-McCloud, line 36: This is the first time, ever this proc has been used--I admit, I'm a bit skeptical of it, as the only thing it does is check for `stat == DEAD` in and of itself---only with additional proc call overhead.

dearmochi: True. Personally I prefer getters and setters where possible (I blame coming from a Java environment for that) but for the sake of consistency with the rest of the code base, I'll probably switch this back to direct value checking.

code/game/objects/items/weapons/grenades/flashbang.dm
```diff
@@ -5,39 +5,42 @@
...
+	if(!T)
 		return
 
-	set_light(7)
-
+	// VFX and SFX
 	do_sparks(rand(5, 9), FALSE, src)
-	playsound(flashbang_turf, 'sound/effects/bang.ogg', 25, 1)
-	bang(flashbang_turf, src, range)
+	playsound(T, 'sound/effects/bang.ogg', 100, TRUE)
```
AffectedArc07, line 20: I feel that flashbangs are already loud enough, though that might just be me

code/game/objects/items/weapons/grenades/flashbang.dm
```diff
@@ -5,39 +5,42 @@
...
+	bang(T, src, range)
 
-	spawn(light_time)
-		qdel(src)
+	// Blob damage here
+	for(var/obj/structure/blob/B in hear(range + 1, T))
+		var/damage = round(30 / (get_dist(B, T) + 1))
+		B.take_damage(damage, BURN, "melee", FALSE)
+
+	qdel(src)
```
AffectedArc07, line 31: On an unrelated sidenote. Can we add this to sonic powder or flash powder or whichever reagent affects the blob IC? I would love to see all the wonderful creations people make to crowd control the blob

code/game/objects/items/weapons/grenades/flashbang.dm
dearmochi: From my testing it wasn't really any louder so 🤷‍♀️

AffectedArc07: I ask the question "why bother with the increase if it wasnt louder"

code/game/objects/items/weapons/grenades/flashbang.dm
dearmochi: Donezo

dearmochi:
Had to force push because Git did a bad

PR #13900, opened 2020-07-22
Removes Unused Power Code

Fox-McCloud:
Just removes some super old and unused power code.

This was a project of some ancient coder back in 2012 that he eventually moved to a folder called "unused".

Apparently Bay split off before that happened on TG and they inherited it (and thu we did too).

Time to purge it.

Also removed duplicate sprites in power.dmi.


Fox-McCloud:
Oh for Pete's sake, Hotel.

PR #13905, opened 2020-07-23
Adds book DRM

AffectedArc07:
## What Does This PR Do
This PR makes it so you can no longer upload reference manuals to the book database. Why you might ask? Well its because the database has too many clones of the same bloody book which can be printed en mass from the librarians computer. It also stops you from re-uploading books you have already printed out.

## Why It's Good For The Game
Duplicate uploads of stuff we can easily print is terrible

## Images of changes
Happens when trying to put a book into the scanner
image

## Changelog
:cl: AffectedArc07
add: You can no longer upload reference manuals to the book database
add: You can no longer re-upload printed books to the book database
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/manuals.dm
+++ b/code/game/objects/items/weapons/manuals.dm
@@ -5,6 +5,7 @@
 	icon = 'icons/obj/library.dmi'
 	due_date = 0 // Game time in 1/10th seconds
 	unique = 1   // 0 - Normal book, 1 - Should not be treated as normal book, unable to be copied, unable to be modified
+	has_drm = TRUE // No reuploading. Piracy is a crime
 
 
 /obj/item/book/manual/engineering_construction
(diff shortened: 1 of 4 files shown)


code/game/objects/items/weapons/manuals.dm
```diff
@@ -5,6 +5,7 @@
 	icon = 'icons/obj/library.dmi'
 	due_date = 0 // Game time in 1/10th seconds
 	unique = 1   // 0 - Normal book, 1 - Should not be treated as normal book, unable to be copied, unable to be modified
```
AffectedArc07, line 7: For whoever asks why dont I just use the `unique` variable, this is used in another places and I am not fully aware of all the implications this may have if I add this onto the scanner, hence the new variable.

Spacemanspark:
The book database has years worth of junk that I doubt we'll ever get to fully remove.

Like, holy hell do you know how many awful _Naruto_ fanfictions I've had to fucking delete?

Ugh.

+1

Carthusia:
ahahaha, just how many have you spark? How many horrible fanfics have you had to partially read before determining to delete them.

Or did you simply see the title "Hinata/Naruto Vs Mecha Hitler" and decide "Nope, nope nope nope NOPE!"

PR #13906, opened 2020-07-23
Fixes admin "call shuttle" verb ignoring station alert level

Kyep:
## What Does This PR Do
Fixes the emergency shuttle ignoring red alert status when called using the admin "call shuttle" verb.
Now, it respects the alert status, displaying a different message, and sending a 5m shuttle instead of a 10m one.
This makes it behave the same way as it already does when the shuttle is called via comms console.

## Changelog
:cl: Kyep
fix: Fixed admin-called shuttles ignoring red alert status.
/:cl:

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -826,7 +826,10 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	else
 		SSshuttle.emergency.canRecall = FALSE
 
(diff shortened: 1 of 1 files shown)


Qwertytoforty:
Isnt their an option in call shuttle verb to make it unrecallable

Kyep:
@Qwertyotoforty Yes, there is.

Spacemanspark:
Ah, hrm. I've always used the antag panel to call the shuttle.

PR #13907, opened 2020-07-23
Energized fire axe rework

AzuleUtama:
## What Does This PR Do
This reworks the energized fireaxe to be a subtype of the standard axe and have a different and (hopefully) more useful special attack.

Previously, this used a lot of copy-pasted code from the regular axe but wasn't a subtype, meaning you couldn't use it to do things like open airlocks. The charged attack it also carried just did an extra 30 burn damage and was recharged by hitting random windows and grilles until you hit a 4% chance to add a charge to the axe, which is quite situational for general combat. The main changes it brings are:

- No longer has the name "energized fire axe" when examined from a distance, meaning you can hold it one handed or on your back without getting valid'd. Still has a unique sprite when wielded as well as an obvious description.
- Can be used for everything a regular axe is used for, namely forcing open airlocks.
- Retains the same armor penetration and basic wielded damage of the old energized axe.
- Axe now recharges after use, taking 30 ticks to charge up for another use.
- The charge attack now stuns for 4 ticks and flings people backwards about 5 tiles when successfully used. Previously this was 30 extra burn damage.

## Why It's Good For The Game

The energized axe is no longer a completely seperate item from the standard axe and works as a subtype, meaning less copy-pasted code. The somewhat awkward way of recharging the axe has been reworked to be more simple and the axe itself should overall be much more useful since I usually see this thing either used for a laugh or discarded if found in a surplus crate.

## Changelog
:cl:
tweak: Reworked energized axe and made it a subtype of the regular fireaxe.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -209,6 +209,27 @@
 /obj/item/twohanded/fireaxe/boneaxe/update_icon()
 	icon_state = "bone_axe[wielded]"
 
+/obj/item/twohanded/fireaxe/energized
+	desc = "Someone with a love for fire axes decided to turn one into a high-powered energy weapon. Seems excessive."
+	force_wielded = 36
+	armour_penetration = 30
+	wieldsound = 'sound/weapons/saberon.ogg'
+	unwieldsound = 'sound/weapons/saberoff.ogg'
+
+/obj/item/twohanded/fireaxe/energized/update_icon()
+	if(wielded)
+		icon_state = "fireaxe2"
+	else
+		icon_state = "fireaxe0"
+
+/obj/item/twohanded/fireaxe/energized/unwield()
+	..()
+	hitsound = "sound/weapons/bladeslice.ogg"
+
+/obj/item/twohanded/fireaxe/energized/wield()
+	..()
+	hitsound = 'sound/weapons/blade1.ogg'
+
 /*
  * Double-Bladed Energy Swords - Cheridan
  */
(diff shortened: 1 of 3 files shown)


code/game/objects/items/weapons/twohanded.dm
```diff
@@ -209,6 +209,42 @@
...
+/obj/item/twohanded/fireaxe/energized/New()
+	..()
+	START_PROCESSING(SSobj, src)
+
+/obj/item/twohanded/fireaxe/energized/Destroy()
+	STOP_PROCESSING(SSobj, src)
+	return ..()
+
+/obj/item/twohanded/fireaxe/energized/process()
+	if(charge < 30)
```
Fox-McCloud, line 233: make this something like `charge = min(charge + 1, 30)` instead

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -209,6 +209,42 @@
...
+		charge++
+
+/obj/item/twohanded/fireaxe/energized/attack(mob/M, mob/user)
+	..()
+	if(wielded && charge == 30)
```
Fox-McCloud, line 238: && `charge == max_charge` or something like that to make it configurable.

Muhdi345:
I dunno, the one time I used it I had a real fun time destroying every window and grille to recharge the damn thing. Then again, having a self-charging stun....

TDSSS:
Wow, I had no idea the old fireaxe was this shitty.

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -209,6 +209,42 @@
...
+/obj/item/twohanded/fireaxe/energized/New()
+	..()
+	START_PROCESSING(SSobj, src)
+
+/obj/item/twohanded/fireaxe/energized/Destroy()
+	STOP_PROCESSING(SSobj, src)
+	return ..()
+
+/obj/item/twohanded/fireaxe/energized/process()
+	charge = min(charge + 1, max_charge)
```
farie82, line 234: Would be better to just use a time instead of it to process constantly.
Currently this takes 30 * 20 (wait time of SSobj) ticks to gain full charge.

You can also make use of a callback if you want to give the user feedback as to when it is charged again

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -209,6 +209,42 @@
...
+
+/obj/item/twohanded/fireaxe/energized/attack(mob/M, mob/user)
+	..()
```
farie82, line 237: Best to use `. = ..()`.
attack does return a value if it is resolved or not

code/game/objects/items/weapons/twohanded.dm
Fox-McCloud: Don't do either; just set it so that you can only use it once every 60 seconds by setting a cooldown of worldtime + 600---then have it display a message if you try to use it before then.

farie82: Using it is attacking something right? So the user gets spammed 60 seconds long when he attacks something that the axe is not charged yet. Not sure if that is a good addition since the user can't check prior to attacking something. And attacking is quite a thing

code/game/objects/items/weapons/twohanded.dm
variableundefined: See above

variableundefined (changes requested):
Just one nitpick, as per Farie @AzuleUtama

PR #13916, opened 2020-07-24
Newer Adminfreeze effect

xProlithium:
## What Does This PR Do
Changed the old adminfreeze effect for a more blatant and obvious effect.

## Why It's Good For The Game
Makes it much more difficult for people to mistake an adminfrozen player for one whose just asleep or unconscious.

## Images of changes
https://i.imgur.com/bIhljrL.png

https://i.imgur.com/3GZccVM.gif

Credit for the sprite goes to Darkmight9.
## Changelog
:cl: xProlithium
Modified adminfreeze effect to be more noticable.
/:cl:


Breenland:
It's nice that it's much easier to see, but it feels almost too intrusive? I think a current issue with the adminfreeze animation now is that it ends up flipped and is very thin. 

Unfortunately I don't know of a good way to make it visually obvious without being how I mentioned, but then again...it's an admin freeze, so that's not entirely necessary. So as it stands I'm okay with it if for no reason other than to stop people from bashing frozen people due to not knowing/seeing the alert.

Spacemanspark:
Make it so they get locked into a giant ice cube with the word 'admin' over it perhaps.
Not sure I'm a fan of the sprites presented, but I do agree that something new could be done to better illustrate what's going on.

_And while we're at it, maybe make it so admin frozen people can't ghost out._

xProlithium:
Couple reasons why I didn't make it an ice cube or like Affected's version.

First, an admin freeze is entirely OOC and should require zero IC justification. If a player gets adminfrozen, we don't need people misunderstanding it for something IC. Similar to why we inform people in LOOC that we're going to admin rejuv a player. If we don't, and the dead player suddenly gets up full health. People will rightfully assume they're a changeling.

Second, because the player falls down to the left or right, Admin needs to be on both sides to make it easier to read. That was the consensus from the staff team during the effect's initial creation.

Darkmight9:
Well if people have suggestions for improving the sprite feel free to tell me them. One thing that could be done is remove the Freeze part of the sign so it's not just one big box with signs and instead it's more like the original jail cell idea I had for it.

AffectedArc07:
The sprite is far too jarring for the game. Yes I know, its meant to be obvious what it is, but it looks like its just pasted into this game from something else, like that kind of jarring. The letters being right on the edge also looks off imo.

Kyep:
Yeah, I'm going to put this down as an actual objection. The sprite is just way too jarring.

variableundefined:
Second objection from maint, closing.

PR #13919, opened 2020-07-25
Removes the refresher anomalous crystal

farie82:
## What Does This PR Do
Staff want it gone for the following reasons:
1. Revive gateway xenomorph facehuggers (bad)
2. Refresh guardian injectors for infinite guardians (bad)
3. infinite death alarms

Discussions have led to the decision to remove it

## Why It's Good For The Game
Interesting idea but seems the exploits it brings with it are more of an issue than what it actually brings

## Changelog
:cl:
del: Removes the refresher anomalous crystal
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/colossus_loot.dm
+++ b/code/modules/mining/lavaland/loot/colossus_loot.dm
@@ -389,32 +389,6 @@
 	if(..())
 		death()
 
-/obj/machinery/anomalous_crystal/refresher //Deletes and recreates a copy of the item, "refreshing" it.
-	activation_method = "touch"
-	cooldown_add = 50
-	activation_sound = 'sound/magic/timeparadox2.ogg'
-	var/list/banned_items_typecache = list(/obj/item/storage, /obj/item/implant, /obj/item/implanter, /obj/item/disk/nuclear, /obj/item/projectile, /obj/item/spellbook)
-
-/obj/machinery/anomalous_crystal/refresher/New()
-	..()
-	banned_items_typecache = typecacheof(banned_items_typecache)
-
-
-/obj/machinery/anomalous_crystal/refresher/ActivationReaction(mob/user, method)
-	if(..())
-		var/list/L = list()
-		var/turf/T = get_step(src, dir)
-		new /obj/effect/temp_visual/emp/pulse(T)
-		for(var/i in T)
-			if(istype(i, /obj/item) && !is_type_in_typecache(i, banned_items_typecache))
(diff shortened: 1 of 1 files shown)


Mitchs98:
Looking at the code for it. Can't you just blacklist certain items from being able to be refreshed?
At-least I assume that's what this line does? 
`if(istype(i, /obj/item) && !is_type_in_typecache(i, banned_items_typecache))`

farie82:
> Looking at the code for it. Can't you just blacklist certain items from being able to be refreshed?
> At-least I assume that's what this line does?
> `if(istype(i, /obj/item) && !is_type_in_typecache(i, banned_items_typecache))`

It is possible yeah
I was requested to make this PR so I didn't look much into it myself. But it sounds like a lot of edge cases you will have to cover for a nearly never used item.

Mitchs98:
> > Looking at the code for it. Can't you just blacklist certain items from being able to be refreshed?
> > At-least I assume that's what this line does?
> > `if(istype(i, /obj/item) && !is_type_in_typecache(i, banned_items_typecache))`
> 
> It is possible yeah
> I was requested to make this PR so I didn't look much into it myself. But it sounds like a lot of edge cases you will have to cover for a nearly never used item.

True....but I mean if the items listed are the main reasons of contention, could add those no?

Was just a thought. I don't really....have any stake in it either way. Just thought I'd mention it.

Qwertytoforty:
The main issue reason that we are removing it is their is a lot of items _that we thought of_ that were dangerous. That doesn't mean those are the only items that exist. More importantly, if we keep this item in, that means every single use item in the future will have to have this crystal as a balance implication. It's better just to remove this.

Mitchs98:
Ahh. I see. Fair enough then 👍

Fox-McCloud:
This has a 16% chance of of even spawning---and Colossus is hardly defeated every single shift.

I'd much rather the problematic things like death alarms or stimulants pens be addressed--it's a pretty interesting item, otherwise.

This is nowhere near the level of what the ExperiMENTOR was capable of with being able to *duplicate* objects.

I'd much rather the extremely egregious edge cases be handled (not so sure facehuggers are that kind of example either) than this to just straight up removed---especially considering how rare this is.

TheMaveran:
Given that Colossus is defeated like what... once every 10-20 shifts, probably rarer than that tbh, and this drops only 16% of the time, I don't think this is breaking any rounds, only adding to them.

variableundefined:

PR #13925, opened 2020-07-25
Light update

Fox-McCloud:
Overlay component based lighting, a port of a port. Originally from Skyrat coded by @Azarak then ported to TG by @Rohesie. Credit to them for most of this. https://github.com/tgstation/tgstation/pull/52413

Component based lighting utilizes filters to "cut through" the darkness to emulate light as opposed to utilizing actual lights. On the surface, this sounds boring, but it allows for extremely smooth lights that really "stick" to an atom---it's also way less costly.

There's no such thing as a free lunch though, and this system does have a number of quirks, though, in my opinion, they're worth putting up with for the massive performance boost and the dramatic increase in smoothness.

- There's a BYOND bug where anything with the light displays the overlay upon right click instead of the object itself:
img
- This doesn't work well with long range lights; it also prevents lights from rendering past a certain view distance
  - Plus side is that lights on the other side of the wall won't show through the wall--downside is that lights that are super long range will stop rendering onside of your render distance...for example:
img
img

Those quirks aside, these lights are virtually indistinguishable from regular lights in appearance:

img

And rest assured, tapetum lucidum still functions:

img

As for performance, the impact is quite sizeable:

here's running around the bridge 5 times with a light that has a range of 5:

(proc, self CPU, total CPU, real time, overtime, total calls)

`/datum/controller/subsystem/lighting/fire 0.061 1.420 1.412 0.000 403`

This consistently incurred a 4-6 ms cost for a single light from a single player

Here's the same light with this new system:

(proc, self CPU, total CPU, real time, overtime, total calls)

`/datum/controller/subsystem/lighting/fire 0.008 0.077 0.077 0.000 337`

now, most of these new lights don't use the lighting system; they utilize other procs---but those profiled so far down the list and their CPU was so low as to effectively be inconsequential.

Also I was annoyed by beepsky's super long delay siren, so I made it change ever half a second instead of every 2 seconds.

:cl: Fox McCloud
tweak: Implements component based lighting overlay system, which should dramatically improve overall lighting performance
tweak: Beepsky's siren now alters at a rate of 1 color change every 0.5 seconds instead of 2 seconds
/:cl:


code/game/machinery/igniter.dm
```diff
@@ -62,18 +62,13 @@
 	var/base_state = "migniter"
 	anchored = 1
 
-/obj/machinery/sparker/New()
-	..()
```
Fox-McCloud, line 66: nice one, old coder.

code/game/objects/effects/misc.dm
```diff
@@ -110,7 +110,7 @@
 
 /obj/effect/dummy/lighting_obj/Initialize(mapload, _color, _range, _power, _duration)
 	. = ..()
-	set_light(_range ? _range : light_range, _power ? _power : light_power, _color ? _color : light_color)
+	AddComponent(/datum/component/overlay_lighting, _range, _power, _color)
```
Fox-McCloud, line 113: note to self, line 108 may need to be removed.

code/modules/fish/fish_items.dm
```diff
@@ -104,7 +104,7 @@
 
 /obj/item/fish/glofish/New()
 	..()
-	set_light(2,1,"#99FF66")
+	AddComponent(/datum/component/overlay_lighting, 1, 1, "#99FF66")
```
Fox-McCloud, line 107: correct this.

code/modules/food_and_drinks/drinks/drinks/shotglass.dm
```diff
@@ -8,6 +8,10 @@
 	var/light_intensity = 2
 	light_color = LIGHT_COLOR_LIGHTBLUE
 
+/obj/item/reagent_containers/food/drinks/drinkingglass/shotglass/Initialize(mapload)
+	. = ..()
+	AddComponent(/datum/component/overlay_lighting, light_intensity, null, light_color)
```
Fox-McCloud, line 13: shouldn't be null; error in old light code.

code/modules/mob/living/simple_animal/hostile/megafauna/colossus.dm
```diff
@@ -222,12 +222,12 @@ Difficulty: Very Hard
 	icon = 'icons/effects/effects.dmi'
 	icon_state = "at_shield2"
 	layer = FLY_LAYER
-	light_range = 2
 	duration = 8
 	var/target
 
 /obj/effect/temp_visual/at_shield/Initialize(mapload, new_target)
 	. = ..()
+	AddComponent(/datum/component/overlay_lighting, 2)
```
Fox-McCloud, line 230: missing args.

code/modules/spacepods/spacepod.dm
```diff
@@ -124,6 +124,7 @@
 	cargo_hold.max_w_class = 5		//fit almost anything
 	cargo_hold.max_combined_w_class = 0 //you can optimize your stash with larger items
 	START_PROCESSING(SSobj, src)
+	AddComponent(/datum/component/overlay_lighting, lights_power, light_power, light_color, lights)
```
Fox-McCloud, line 127: second arg is incorrect

Denthamos:
There goes fox, lighting up our world with this enlightening PR.  It really is a shining example of old crap code to be forgotten in a flash.

In seriousness, finally doing away with the old shit lighting system.  The improvement to the server CPU wise will be amazing.

KasparoVy:
Glad the eyeshine still works! Cheers. This is bloody wizardry

SteelSlayer:
When you say "This doesn't work well with long range lights; it also prevents lights from rendering past a certain view distance". Is this for all light sources? And do you know the distance? I tested this with a floodlight and some other stuff, like flashlights, and I could see them from far away. For example, I put a floodlight in one corner of the central primary hallway, and walked all the way down the hall, almost to another corner. Even then I could still see its light on the edge of my screen.

Fox-McCloud:
@SteelSlayer 

The floodlights use the old system because they have a massive range on them that's well beyond what this system would support.

This doesn't change every light in existence; only *most* lights that you move around---computers, lights on the wall, machinery, and the likes are still mostly utilizing the old system.

Qwertytoforty:
Ghosts make a ball of light arround them, visible to crew, may have been due to summon ghosts however 
image

ZomgPonies:
Radius of lights from mobs light up tiles through walls. Noticed while ghosting a player.

ssavinov-numerator:
lighting_bug

Looks like a bug

Butlerblock:
Everything about borg headlamp is broken. level 4 and 5 are the exact same as 3, but still, take the corresponding power, and no matter what level you can't see past 3 tiles away. There's literally no reason to post the same image twice so I'm just gonna put the one with head lamp level 5 
image

limofs:
> 
> 
> Radius of lights from mobs light up tiles through walls. Noticed while ghosting a player.

lighting_bug

Azarak:
@limofs This is only noticable as a ghost/person with Xray, as in game you won't see the lighting mask unless you see the atom holding it

Mitchs98:
It's pretty much impossible to see in maint past your source of light. And the light randomly disappearing when people walk to the corner of the screen/off your screen is as jarring as I expected it to be.

Fox-McCloud:
>Ghosts make a ball of light arround them, visible to crew, may have been due to summon ghosts however

Intentional. When ghosts are made visible, they emit light, even presently

>Everything about borg headlamp is broken. level 4 and 5 are the exact same as 3

I already noted this.

@azarak covered the details of x-ray and ghost vision.

Fox-McCloud:
this will be revived, eventually.

PR #13934, opened 2020-07-27
Maintenance Drone Tweaks

SabreML:
## What Does This PR Do

1. Allows Maintenance drones to drag construction materials such as Rods, Metal sheets and Floor tiles.
2. Shows a popup if you try using your ID on a drone.
3. Allows Maintence drones and Engiborgs to pick up Light Switches and Extinguisher Cabinet frames with their grabber.

## Why It's Good For The Game

1. Nobody likes seeing metal rods inside windows.
2. No more accidentaly instakilling a drone.
3. Self explanitory.

## Changelog
:cl:
add: Maintenance drones can now drag construction materials.
add: Warns people what will happen when they try to use their ID on a maintenance drone.
tweak: Maintenance drones can now pick up Light Switches and Extinguisher Cabinet frames.
tweak: Maintenance drones and Cyborgs can now open and close Extinguisher Cabinets.
spellcheck: Fixed light replacer grammar.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone.dm
@@ -162,8 +162,7 @@
 			return
 
 		else
-			var/confirm = null
-			confirm = alert("Using your ID on a Maintenance Drone will shut it down, are you sure you want to do this?", "Disable Drone", "Yes", "No")
+			var/confirm = alert("Using your ID on a Maintenance Drone will shut it down, are you sure you want to do this?", "Disable Drone", "Yes", "No")
 			if(confirm == ("Yes"))
 				user.visible_message("<span class='warning'>\the [user] swipes [user.p_their()] ID card through [src], attempting to shut it down.</span>", "<span class='warning'>You swipe your ID card through \the [src], attempting to shut it down.</span>")
 
--- a/code/modules/mob/living/silicon/robot/drone/drone.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone.dm
@@ -35,6 +35,14 @@
 	var/reboot_cooldown = 60 // one minute
 	var/last_reboot
 	var/emagged_time
+	var/list/pullable_drone_items = list(
+		/obj/item/pipe,
+		/obj/structure/disposalconstruct,
+		/obj/item/stack/cable_coil,
+		/obj/item/stack/rods,
+		/obj/item/stack/sheet,
+		/obj/item/stack/tile
+	)
 
 	holder_type = /obj/item/holder/drone
 //	var/sprite[0]
@@ -323,15 +331,6 @@
 
 /mob/living/silicon/robot/drone/Bumped(atom/movable/AM)
 	return
-
-var/list/pullable_drone_items = list(
-	/obj/item/pipe,
-	/obj/structure/disposalconstruct,
-	/obj/item/stack/cable_coil,
-	/obj/item/stack/rods,
-	/obj/item/stack/sheet,
-	/obj/item/stack/tile
-)
 
 /mob/living/silicon/robot/drone/start_pulling(var/atom/movable/AM)
 
--- a/code/modules/mob/living/silicon/robot/drone/drone.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone.dm
@@ -330,7 +330,7 @@ var/list/pullable_drone_items = list(
 	/obj/item/stack/cable_coil,
 	/obj/item/stack/rods,
 	/obj/item/stack/sheet,
-	/obj/item/stack/tile,)
+	/obj/item/stack/tile)
 
 /mob/living/silicon/robot/drone/start_pulling(var/atom/movable/AM)
 
--- a/code/modules/mob/living/silicon/robot/drone/drone.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone.dm
@@ -77,6 +77,10 @@
 	verbs -= /mob/living/silicon/robot/verb/Namepick
 	module = new /obj/item/robot_module/drone(src)
 
+	//Allows Drones to hear the Engineering channel.
+	module.channels = list("Engineering" = 1)
+	radio.recalculateChannels()
+
 	//Grab stacks.
 	stack_metal = locate(/obj/item/stack/sheet/metal/cyborg) in src.module
 	stack_wood = locate(/obj/item/stack/sheet/wood) in src.module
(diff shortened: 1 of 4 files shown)


SabreML:
image
Ah crap, hang on.

code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -8,7 +8,6 @@
 
 	//Has a list of items that it can hold.
 	var/list/can_hold = list(
-		/obj/item/stock_parts/cell,
```
SteelSlayer, line 11: Why was this removed?

code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -154,15 +154,18 @@
 			return
 
 		else
-			user.visible_message("<span class='warning'>\the [user] swipes [user.p_their()] ID card through [src], attempting to shut it down.</span>", "<span class='warning'>You swipe your ID card through \the [src], attempting to shut it down.</span>")
+			var/confirm = null
+			confirm = alert("Using your ID on a Maintenance Drone will shut it down, are you sure you want to do this?", "Disable Drone", "Yes", "No")
+			if (confirm == ("Yes"))
```
SteelSlayer, line 159: We prefer that if's are adjacent to parentheses: `if(confirm == ("Yes"))`.

code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -323,7 +326,12 @@
 
 /mob/living/silicon/robot/drone/start_pulling(var/atom/movable/AM)
 
-	if(istype(AM,/obj/item/pipe) || istype(AM,/obj/structure/disposalconstruct))
+	if(istype(AM,/obj/item/pipe) \
+	|| istype(AM,/obj/structure/disposalconstruct) \
+	|| istype(AM,/obj/item/stack/cable_coil) \
+	|| istype(AM,/obj/item/stack/rods) \
+	|| istype(AM,/obj/item/stack/sheet) \
+	|| istype(AM,/obj/item/stack/tile))
```
SteelSlayer, line 334: I suggest making a list of all these paths, something like `pullable_drone_items`, so you can do
 `if(is_type_in_list(AM, pullable_drone_items))`.

code/modules/mob/living/silicon/robot/drone/drone_items.dm
SabreML: `/obj/item/stock_parts`, is also listed, so it's already there.

code/modules/mob/living/silicon/robot/drone/drone.dm
SabreML: Where should I put the list?

code/modules/mob/living/silicon/robot/drone/drone_items.dm
SteelSlayer: Oh yeah this is fine then.

code/modules/mob/living/silicon/robot/drone/drone.dm
SteelSlayer: I was thinking as a var on drones.

code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -321,10 +324,20 @@
 /mob/living/silicon/robot/drone/Bumped(atom/movable/AM)
 	return
 
+var/list/pullable_drone_items = list(
```
SteelSlayer, line 327: Good but you should probably declare this on drones, as right now its just hanging around. I'm not sure how this would behave actually. 

What I mean is something like this:

```js
/mob/living/silicon/robot/drone
	name = "drone"
    ... // other stuff

	var/emagged_time
    var/list/pullable_drone_items = list(
        ...
    )

Comment (SabreML) [reply code/modules/mob/living/silicon/robot/drone/drone.dm:327]: Seems to work just fine ingame as it is. *shrug

It should be like this then?
```
        var/last_reboot
        var/emagged_time
	var/list/pullable_drone_items = list(
	/obj/item/pipe,
	/obj/structure/disposalconstruct,
	/obj/item/stack/cable_coil,
	/obj/item/stack/rods,
	/obj/item/stack/sheet,
	/obj/item/stack/tile
	)

SteelSlayer: Yeah, but indent all the entries in the list to the right one tab. Also my indenting got messed up in my last comment, fixed it though.

SabreML:
Woo!

code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -154,15 +162,18 @@
 			return
 
 		else
-			user.visible_message("<span class='warning'>\the [user] swipes [user.p_their()] ID card through [src], attempting to shut it down.</span>", "<span class='warning'>You swipe your ID card through \the [src], attempting to shut it down.</span>")
+			var/confirm = null
+			confirm = alert("Using your ID on a Maintenance Drone will shut it down, are you sure you want to do this?", "Disable Drone", "Yes", "No")
```
SteelSlayer, line 166: You can meld this into one line by just doing `var/confirm = alert(blah blah)`

code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -154,15 +162,17 @@
 			return
 
 		else
-			user.visible_message("<span class='warning'>\the [user] swipes [user.p_their()] ID card through [src], attempting to shut it down.</span>", "<span class='warning'>You swipe your ID card through \the [src], attempting to shut it down.</span>")
+			var/confirm = alert("Using your ID on a Maintenance Drone will shut it down, are you sure you want to do this?", "Disable Drone", "Yes", "No")
```
Kyep, line 165: Doesn't this let you swipe your ID on a drone, keep the popup on-screen, and then click 'yes' later after the drone has moved offscreen?

Also, if a check is added, doesn't this mean the drone can avoid being de-activated by moving away from you really fast?

SabreML: It does, yes.
I thought it would be better to avoid the second situation since otherwise you could never actually deactivate one, unless you got lucky.

It also won't matter for emagged drones, since they can't be deactivated anyway.

SabreML:
Just had another idea, how about letting drones listen in to the Engineering channel, like they can with the Binary one?

code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -154,15 +166,17 @@
...
```
AffectedArc07, line 169: This isnt going to pass until this is changed to have a range check, or a 3 seconds timeout.

SabreML: I'll try to figure that out tomorrow, too late right now.

SabreML: Pretty sure that does it.

PR #13935, opened 2020-07-27
[TGUI] port of silicon/aghost airlock controls

Kyep:
## What Does This PR Do
Converts the AI/borg/adminghost view of airlocks to use TGUI.

## Why It's Good For The Game
Less nanoui. Less whitescreens.

## Images of changes
doorcontrol_normal
doorcontrol_allcut
doorcontrol_shorted
doorcontrol_elec_temp
doorcontrol_elec_perm
doorcontrol_options_off

## Changelog
:cl: Kyep
add: Ported airlock silicon/adminghost controls to TGUI.
/:cl:


AffectedArc07:
You can indeed
image

code/game/machinery/doors/airlock.dm
```diff
@@ -570,38 +570,61 @@ About the new airlock wires panel:
...
-	ui_interact(user)
+	tgui_interact(user)
 
 /obj/machinery/door/airlock/attack_ai(mob/user)
-	ui_interact(user)
+	tgui_interact(user)
 
-/obj/machinery/door/airlock/ui_interact(mob/user, ui_key = "main", datum/nanoui/ui = null, force_open = 1)
-	ui = SSnanoui.try_update_ui(user, src, ui_key, ui, force_open)
+/obj/machinery/door/airlock/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = FALSE, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
```
AffectedArc07, line 578: You change `force_open = 1` to `force_open = FALSE`. Was this meant to be `TRUE`?

code/game/machinery/doors/airlock.dm
```diff
@@ -193,10 +193,10 @@ About the new airlock wires panel:
 	return wires.IsIndexCut(wireIndex)
 
 /obj/machinery/door/airlock/proc/canAIControl()
-	return ((aiControlDisabled!=1) && (!isAllPowerLoss()))
+	return ((aiControlDisabled != 1) && (!isAllPowerLoss()))
```
AffectedArc07, line 196: What is `1` in this context? TRUE? Another value that should be a define? The numbers are too magical

code/game/machinery/doors/airlock.dm
```diff
@@ -570,38 +570,61 @@ About the new airlock wires panel:
...
-	commands[++commands.len] = list("name" = "Door State",		"command"= "open",		"active" = density,				"enabled" = "Closed",	"disabled" = "Opened", 		"danger" = 0, "act" = 0)
-	commands[++commands.len] = list("name" = "Emergency Access","command"= "emergency",	"active" = !emergency,			"enabled" = "Disabled",	"disabled" = "Enabled", 	"danger" = 0, "act" = 0)
-
-	data["commands"] = commands
+
+
+/obj/machinery/door/airlock/tgui_data(mob/user)
+	var/list/data = list()
+
+	// 2 = green, 1 = orange, 0 = red.
```
AffectedArc07, line 588: May be worth doing
```
#define TGUI_GREEN 2
#define TGUI_ORANGE 1
#define TGUI_RED 0
```

and then
```
#undef TGUI_GREEN
#undef TGUI_ORANGE
#undef TGUI_RED
```
for the sake of readability

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
-				aiDisabledIdScanner = 0
-				to_chat(usr, "IdScan feature has been enabled.")
-			else if(!activate && !aiDisabledIdScanner)
-				aiDisabledIdScanner = 1
-				to_chat(usr, "IdScan feature has been disabled.")
-		if("main_power")
+/obj/machinery/door/airlock/tgui_act(action, params)
+	if(..())
+		return
+	if(!issilicon(usr) && !usr.can_admin_interact())
```
AffectedArc07, line 771: I am not sure about having `admins` in a message youll see from an IC object but I wont make a huge debate over it

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
-				to_chat(usr, "The door bolts have been dropped.")
-			else if(!activate && unlock())
-				to_chat(usr, "The door bolts have been raised.")
-		if("electrify_temporary")
-			if(activate && isWireCut(AIRLOCK_WIRE_ELECTRIFY))
+			else
+				to_chat(usr, "<span class='warning'>Backup power is already offline.</span>")
+			. = TRUE
+		if("shock-restore")
+			to_chat(usr, "The door is now un-electrified.")
```
AffectedArc07, line 802: Maybe `<span class='notice'>` for this?

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
+			electrify(0)
+			. = TRUE
+		if("shock-temp")
+			if(isWireCut(AIRLOCK_WIRE_ELECTRIFY))
 				to_chat(usr, text("The electrification wire is cut - Door permanently electrified."))
```
AffectedArc07, line 807: We dont allow use of `text()`
image

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
-			else if(!activate && !density)
-				close()
-		if("safeties")
-			// Safeties!  We don't need no stinking safeties!
+			. = TRUE
+		if("idscan-toggle")
+			if(isWireCut(AIRLOCK_WIRE_IDSCAN))
+				to_chat(usr, "The IdScan wire has been cut - IdScan feature permanently disabled.")
+			else if(aiDisabledIdScanner)
+				aiDisabledIdScanner = 0
```
AffectedArc07, line 830: FALSE?

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
+				to_chat(usr, "IdScan feature has been enabled.")
+			else
+				aiDisabledIdScanner = 1
```
AffectedArc07, line 833: FALSE?

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
+			. = TRUE
+		if("light-toggle")
+			if(isWireCut(AIRLOCK_WIRE_LIGHT))
+				to_chat(usr, "The bolt lights wire has been cut - The door bolt lights are permanently disabled.")
+			else if(lights)
+				lights = 0
+				to_chat(usr, "The door bolt lights have been disabled.")
+			else if(!lights)
+				lights = 1
+				to_chat(usr, "The door bolt lights have been enabled.")
```
AffectedArc07, line 860: All of these should have `<span class='notice'>` in my opinion

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
+			update_icon()
+			. = TRUE
+		if("safe-toggle")
 			if(isWireCut(AIRLOCK_WIRE_SAFETY))
 				to_chat(usr, text("The safety wire is cut - Cannot secure the door."))
```
AffectedArc07, line 865: Same thing for using `text()` here

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
 				safe = 0
-			else if(!activate && !safe)
+			else
 				safe = 1
-		if("timing")
-			// Door speed control
+			. = TRUE
+		if("speed-toggle")
 			if(isWireCut(AIRLOCK_WIRE_SPEED))
 				to_chat(usr, text("The timing wire is cut - Cannot alter timing."))
```
AffectedArc07, line 873: And here

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
-			// Emergency access
-			if(emergency)
-				emergency = 0
-				to_chat(usr, "Emergency access has been disabled.")
+			. = TRUE
+		if("open-close")
+			if(welded)
+				to_chat(usr, text("The airlock has been welded shut!"))
+			else if(locked)
+				to_chat(usr, text("The door bolts are down!"))
```
AffectedArc07, line 883: Same thing with `text()` here

code/game/machinery/doors/airlock.dm
```diff
@@ -737,125 +760,132 @@ About the new airlock wires panel:
...
-			if(isWireCut(AIRLOCK_WIRE_IDSCAN))
-				to_chat(usr, "The IdScan wire has been cut - IdScan feature permanently disabled.")
-			else if(activate && aiDisabledIdScanner)
-				aiDisabledIdScanner = 0
-				to_chat(usr, "IdScan feature has been enabled.")
-			else if(!activate && !aiDisabledIdScanner)
-				aiDisabledIdScanner = 1
-				to_chat(usr, "IdScan feature has been disabled.")
-		if("main_power")
+/obj/machinery/door/airlock/tgui_act(action, params)
```
dearmochi, line 768: Putting `. = TRUE` in every valid action leads to duplicate code. You can simply do that *once* before the `switch` and then add a default case using `else` that just returns `FALSE`. That way, only valid actions will return `TRUE`.

tgui/packages/tgui/interfaces/AiAirlock.js
```diff
@@ -0,0 +1,202 @@
...
+    localStatusText: 'Caution',
+  },
+  0: {
+    color: 'bad',
+    localStatusText: 'Optimal',
+  },
+};
+
+export const AiAirlock = (props, context) => {
+  const { act, data } = useBackend(context);
```
dearmochi, line 22: Rather than access every field from `data` by doing `data.power` etc., you can make local variables from them by doing: 
```js
const {
	power,
	wires,
	// ETC...
} = data;
```

tgui/packages/tgui/interfaces/AiAirlock.js
```diff
@@ -0,0 +1,202 @@
...
+                    onClick={() => act('shock-temp')} />
+                  <Button
+                    icon="bolt"
+                    disabled={!data.wires.shock || data.shock === 0}
+                    content="Permanent"
+                    onClick={() => act('shock-perm')} />
+                </Fragment>
+              )}>
+              {data.shock === 2 ? 'Safe' : 'Electrified'}
+              {' '}
```
dearmochi, line 90: Perhaps simply `&nbsp;`

tgui/packages/tgui/interfaces/AiAirlock.js
Kyep: I could, but I don't really see any advantage to doing that, aside from slightly shorter var names, which isn't really worth the price of having every var name be different from TG.

dearmochi: They wouldn't be different, just made local. `data.power` would turn into `power`. Besides, I feel like we shouldn't copy everything from TG as-is. Porting is fine, but it should also be an opportunity to improve the code that is being ported, no matter how insignificant the problem might be. It's a matter of good practice within the codebase and consistency.

tgui/packages/tgui/interfaces/AiAirlock.js
Kyep: No, I tested that and it does not work.

Spacemanspark:
This is great and all, but why are we removing white screens? I think it's a great feature that makes players think before clicking.

AffectedArc07:
@Spacemanspark 
image
*2002 windows XP notebook wants to know your location*

AffectedArc07:
PR contains conflicts outside of `tgui.bundle.js`.

PR will not be auto-deconflicted.

Kyep:
@AffectedArc07 Deconflicted.

AffectedArc07 (approved):
IN THE NAME OF BACKLOG CLEARANCE

PR #13936, opened 2020-07-27
Game Preferences Update

### LINKED ISSUE #13733: We are out of bitflags
**Issue Description**:
Our toggles preference has used the maximum amount of bitflags that BYOND supports (24). As such, we cannot add any more toggleable preferences without a stupid workaround, or a preference refactor (maybe ill do it in the future)

Preferences to save above that number

**What happened instead**:
It didnt even toggle properly

We are unable to save any more preferences

```dm
#define AATESTING 16777216

/client/verb/aatesting()
	set name = "AA Testing"
	set category = "Special Verbs"
	prefs.toggles ^= AATESTING
	to_chat(src, "Test toggle [(prefs.toggles & AATESTING) ? "on" : "off"]")
```
image

When numpad targetting became a toggle

**Extra information**:
JSON preferences or savefiles are the way to fix this.

AffectedArc07:
## What Does This PR Do
This PR updates the DB schema for a `toggles_2` column since we are out of bitflags. It also condenses some single-bool columns into this bitflag (SQL update file will make sure existing prefs are preserved).

It also goes over the existing toggles to make it obvious what they are
Fixes #13733 

And yes. This has been tested. **thoroughly**.

## Why It's Good For The Game
We shouldnt be cut off from making more preference toggles

## Changelog
:cl: AffectedArc07
add: Adds ability for us to store more preference toggles in the future
/:cl:


AffectedArc07:
The reason we need an entire new variable is a BYOND limitation, not a DB limitation. We can have *many* more flags in the DB by increasing the int size, but the actual limit is the byond numerical limit.

Kyep:
Note to self: add something to backend tooling which, during testmerge start, does a hard file check for the existence of SQL/updates/X-Y.sql. If not found, auto-cancel testmerge. This safety check should forcibly prevent TMing any PR which hasn't been updated since the last SQL update PR was merged.

PR #13940, opened 2020-07-28
Adds a warning when using Strange Reagent on a corpse.

SabreML:
## What Does This PR Do

Gives a warning when using Strange Reagent on a corpse.

## Why It's Good For The Game

Less accidental gibbing.

## Changelog
:cl:
add: Added warning popup when using SR on a dead mob.
/:cl:

Many thanks to #coding_chat on the Discord.

Not particularly happy with the alert text, suggestions would be nice!

### DIFF
--- a/code/modules/reagents/reagent_containers/pill.dm
+++ b/code/modules/reagents/reagent_containers/pill.dm
@@ -24,7 +24,7 @@
 /obj/item/reagent_containers/food/pill/attack(mob/living/carbon/M, mob/user, def_zone)
 	if(!istype(M))
 		return 0
-	if(reagents.has_reagent("strange_reagent") & M.stat== DEAD)  //If the pill contains SR and the mob is dead.
+	if(reagents.has_reagent("strange_reagent", 1) && M.stat == DEAD)  //If the pill contains SR and the mob is dead.
 		var/confirm = alert(user, "Are you sure? Using Strange Reagent on a body with more than 150 combined Brute, Burn, and Genetic damage will gib them!", "Warning", "Yes", "No")
 		if(confirm != "Yes")
 			return
--- a/code/modules/reagents/reagent_containers/pill.dm
+++ b/code/modules/reagents/reagent_containers/pill.dm
@@ -24,6 +24,10 @@
 /obj/item/reagent_containers/food/pill/attack(mob/living/carbon/M, mob/user, def_zone)
 	if(!istype(M))
 		return 0
+	if(reagents.has_reagent("strange_reagent") & M.stat== DEAD)  //If the pill contains SR and the mob is dead.
+		var/confirm = alert(user, "Are you sure? Using Strange Reagent on a body with more than 150 combined Brute, Burn, and Genetic damage will gib them!", "Warning", "Yes", "No")
+		if(confirm != "Yes")
+			return
 	bitesize = reagents.total_volume
 	if(M.eat(src, user))
 		spawn(0)


code/modules/reagents/reagent_containers/pill.dm
```diff
@@ -24,6 +24,10 @@
 /obj/item/reagent_containers/food/pill/attack(mob/living/carbon/M, mob/user, def_zone)
 	if(!istype(M))
 		return 0
+	if(reagents.has_reagent("strange_reagent") & M.stat== DEAD)  //If the pill contains SR and the mob is dead.
```
SteelSlayer, line 27: `&&` instead of `&` and need a space between `stat` and the `==`.

TDSSS:
Have to say, not sure this is a thing I want to see getting merged. Ultimately, we must allow people to fail, we can't hugbox everything. 'I messed up and my patient exploded into gibs, oops' is a very SS13-style thing and generates hilarious moments sometimes. 

SR revival itself is a high-end medical process that should not be taken for granted, being unclonable is a downside after all.

Carthusia:
Yea, gota say, even after all the times i've been asploded in medical as as vox, i wouldn't want it to go "WARNING THIS MAY EXPLODE THEM IF THEY ARE OVER 150 COMBINED BRUTE AND BURN DAMAGE! DO YOU WISH TO CONTINUE?!"


Its one person, and even then its not the end of the round and they can be fixed in a number of different ways.

I still remember the first time I exploded someone, it was a good lesson on being careful with patients and always frequently checking on them before continuing treatment.

I agree with TDSSS, people need to be allowed to fail, SS13 isa game based around failure and learning from it to better yourself in how you play, be it accidentally releasing a horrible virus, or transplanting the brain of an EOC into a monkey and not realising it can vent crawl and it escapes.

These instances of failure are what create great stories for people. If it was all to safe and sterile, the game wouldn't be nearly as much fun.

Spacemanspark:
Kinda agree with what's already been said, I don't think we need to put safety tape over everything.

code/modules/reagents/reagent_containers/pill.dm
Kyep, line 27:
```suggestion
	if(reagents.has_reagent("strange_reagent", 1) & M.stat== DEAD)  //If the pill contains SR and the mob is dead.
```

SabreML:
Travis seems over enthusiastic.
image

SabreML:
Very good points, and to be honest I agree with you.
This was primarily beacuse I've seen people in dchat and so on asking for a warning.

I'll leave the PR up for a while, mostly to collect feedback.

farie82:
> This was primarily beacuse I've seen people in dchat and so on asking for a warning.

Do be aware that dchat is very salty at times so taking their requests has to be taken with a grain of salt as well ;)

Dumbdumn5:
Similarly concerned with adding potential safety tape to things because it make the 'death by SR' a bit harder to pull off. I understood the warnings for SSDs, those were breaking OOC rules when they went down, but this is more of an IC thing and having it be as seamless as possible makes the game a bit less jank. Competency in doctors means more lives saved and incompetency leads to moments that, while frustrating, force players in the medical field to start learning. Every death or person mismanaged is a failure worth learning from, and that problem is better remedied through mentors, tutors, and IC partners imo.

SabreML:
Reopening to fix travis

AffectedArc07:
Travis wont be fixed, you just have to deal with it

SabreML:
Well that didn't work :/

Sikroth:
Personally against this as it's way too hand holdy. Plenty of things in the game can kill you or others pretty quickly and there's no warning. As frustrating as it might be to be wrongly SR'd and gibbed, it is a part of the game that not everyone is perfect at their job. And as previously discussed there's also a lot of more traitorous uses of SR that this'd mess with.

That said, I've seen people being bwoinked over missusing SR in the past, which I still to this day find questionable... but that's an entirely different topic

SabreML:
Based on the feedback, I'll close this PR.

I hadn't really considered the intentional gibbing uses, and just wanted to add something to stop people from doing it accidentaly.
I suppose this is what the comment system is for, to stop people having tunnel vision.

Either way, I learned more byond code, so that's always good.

PR #13943, opened 2020-07-28
*more* Tcomms Tweaks

### LINKED ISSUE #13968: Sentence verbs over comns do not work right anymore
**Issue Description**:

Sentence prefixes, like asks, yells, queries, states, ect, doesn't work over telecomns anymore

For if you are to speak in a way that affects how your verb comes out, for this to be reflected on radio

**What happened instead**:

everyone has says over radio no matter what

AI no longer states over radio, and its weird having yourself ask or state locally, and it comes out as says over radio.

say ; Does this work?
image
See as it says ass locally and says over telecomns

Telecomns rework

AffectedArc07:
## What Does This PR Do
This PR makes it so that tcomms machines update when being moved on shuttles, auto disable when moved on shuttles, and also disallowed having multiple machines active on the same zlevel (hot spares).

Fixes #13968

It also fixes this weird bug with verbage

Before:
*notice the difference says and exclaims*
image

After:
image

## Why It's Good For The Game
People have started creating multiple tcomms cores on the station, all of which are active at once. This means that you cannot easily take down tcomms. This will make it so that people cant just have a running hot spare tcomms core. You can still make a new core, but cant enable it until the existing one is disabled or destroyed. 

The reason for the shuttle changes is twofold:
1: Moving a tcomms machine between z-levels didnt update the linkage lists, and that was a bug
2: Moving active machines onto zlevels made it so that you could bypass the one-machine checks, so it was done regardless

## Images of changes
*ignore the name differences, I loaded on the wrong SQL version for one run and screwed character loading*
image

image

image

## Changelog
:cl: AffectedArc07
add: Tcomms machines now disable on Z-level changes (Example: Shuttle moves)
add: Tcomms machines now update properly on Z-level changes (Example: Shuttle moves)
tweak: You can no longer have more than 1 running tcomms core per Z-level
tweak: You can no longer have more than 1 running tcomms relay per Z-level
/:cl:

### DIFF
--- a/code/game/machinery/tcomms/_base.dm
+++ b/code/game/machinery/tcomms/_base.dm
@@ -106,22 +106,36 @@ GLOBAL_LIST_EMPTY(tcomms_machines)
 
 
 /**
-  * Start of Ion Anomalie Event
+  * Start of Ion Anomaly Event
   *
-  * Proc to easily start an Ion Anomalie's effects, and update the icon
+  * Proc to easily start an Ion Anomaly's effects, and update the icon
   */
 /obj/machinery/tcomms/proc/start_ion()
 	ion = TRUE
 	update_icon()
 
 /**
-  * End of Ion Anomalie Event
+  * End of Ion Anomaly Event
   *
-  * Proc to easily stop an Ion Anomalie's effects, and update the icon
+  * Proc to easily stop an Ion Anomaly's effects, and update the icon
   */
 /obj/machinery/tcomms/proc/end_ion()
 	ion = FALSE
 	update_icon()
+
+/**
+  * Z-Level transit change helper
+  *
+  * Proc to make sure you cant have two of these active on a Z-level at once. It also makes sure to update the linkage
+  */
+/obj/machinery/tcomms/onTransitZ(old_z, new_z)
+	. = ..()
+	if(active)
+		active = FALSE
+		// This needs a timer because otherwise its on the shuttle Z and the message is missed
+		addtimer(CALLBACK(src, /atom.proc/visible_message, "<span class='warning'>Radio equipment on [src] has been overloaded by heavy bluespace interference. Please restart the machine.</span>"), 5)
+	update_icon()
+
 
 /**
   * Logging helper
--- a/code/game/machinery/tcomms/core.dm
+++ b/code/game/machinery/tcomms/core.dm
@@ -41,6 +41,7 @@
 		active = TRUE
 	else
 		visible_message("<span class='warning'>Error: Another core is already active in this sector. Power-up cancelled due to radio interference.</span>")
+	update_icon()
 
 /**
   * Destructor for the core.
--- a/code/game/machinery/tcomms/core.dm
+++ b/code/game/machinery/tcomms/core.dm
@@ -14,6 +14,8 @@
 	name = "Telecommunications Core"
 	desc = "A large rack full of communications equipment. Looks important."
 	icon_state = "core"
+	// This starts as off so you cant make cores as hot spares
+	active = FALSE
 	/// The NTTC config for this device
 	var/datum/nttc_configuration/nttc = new()
 	/// List of all reachable devices
@@ -35,6 +37,10 @@
 	link_password = GenerateKey()
 	reachable_zlevels |= loc.z
 	component_parts += new /obj/item/circuitboard/tcomms/core(null)
+	if(check_power_on())
+		active = TRUE
+	else
+		visible_message("<span class='warning'>Error: Another core is already active in this sector. Power-up cancelled due to radio interference.</span>")
 
 /**
   * Destructor for the core.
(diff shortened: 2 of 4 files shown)


code/game/machinery/tcomms/_base.dm
```diff
@@ -106,23 +106,37 @@ GLOBAL_LIST_EMPTY(tcomms_machines)
...
 /obj/machinery/tcomms/proc/start_ion()
 	ion = TRUE
 	update_icon()
 
 /**
-  * End of Ion Anomalie Event
+  * End of Ion Anomaly Event
   *
-  * Proc to easily stop an Ion Anomalie's effects, and update the icon
+  * Proc to easily stop an Ion Anomaly's effects, and update the icon
```
AffectedArc07, line 120: These typos were hurting me so I fixed them

code/game/machinery/tcomms/relay.dm
```diff
@@ -83,9 +117,9 @@
   * Proc which ensures the host core has its zlevels updated (icons are updated by parent call)
   */
 /obj/machinery/tcomms/relay/power_change()
-    ..()
-    if(linked_core)
-        linked_core.refresh_zlevels()
+	..()
+	if(linked_core)
+		linked_core.refresh_zlevels()
```
AffectedArc07, line 122: This was indented with spaces, not tabs. Has been fixed

AffectedArc07:
oshit forgot relays

code/game/machinery/tcomms/_base.dm
farie82: Who is Anomalie and why is she making these ion anomalies?!

code/game/machinery/tcomms/core.dm
```diff
@@ -121,6 +128,38 @@
...
+/obj/machinery/tcomms/core/onTransitZ(old_z, new_z)
+	. = ..()
+	refresh_zlevels()
+
+/**
+  * Power-on checker
+  *
+  * Checks the z-level to see if an existing core is already powered on, and deny this one turning on if there is one. Returns TRUE if it can power on, or FALSE if it cannot
+  */
+/obj/machinery/tcomms/core/proc/check_power_on()
```
farie82, line 145: Wouldn't it be better to declare this at tcomms level instead of the core and relay separately?
They both are nearly the same yet search for different machines. Best to avoid dupe code

code/game/objects/items/devices/radio/radio.dm
```diff
@@ -325,7 +325,7 @@ GLOBAL_LIST_INIT(default_medbay_channels, list(
 	// If we were to send to a channel we don't have, drop it.
 	return RADIO_CONNECTION_FAIL
 
-/obj/item/radio/talk_into(mob/living/M as mob, list/message_pieces, channel, var/verb = "says")
+/obj/item/radio/talk_into(mob/living/M as mob, list/message_pieces, channel, verbage = "says")
```
farie82, line 328: can ya remove the as mob thing in the params here?

code/game/machinery/tcomms/relay.dm
```diff
@@ -26,6 +28,11 @@
 /obj/machinery/tcomms/relay/Initialize(mapload)
 	. = ..()
 	component_parts += new /obj/item/circuitboard/tcomms/relay(null)
+	if(check_power_on())
```
farie82, line 31: Mostly dupe code here as well. Best to check if you can put this a level up

code/game/machinery/tcomms/core.dm
AffectedArc07: I defined it at this level because of the typepath change in the `GLOB.tcomms_machines` list, and theres so much of a change between those that this made more sense ot me

code/game/machinery/tcomms/relay.dm
AffectedArc07: This might be cleanable but I aint sure of it

code/game/objects/items/devices/radio/radio.dm
AffectedArc07: When I get round to it

code/game/machinery/tcomms/core.dm
farie82: You can make a proc to return the list type. Then loop through it with a `var/thing` and cast it to `obj/machinery/tcomms`. I don't think you have any actual checks that relate to core or relay right? If so you can make a proc for the checks

AffectedArc07: I do, the checks need to be separate between a core or a relay, so you can have 1 core AND 1 relay. Not 1 core OR relay.

farie82: You don't have to check variables on the core or relay though. Only on core level.
My idea was like this
```
.../tcomms/proc/get_machines_to_check()
  return
...

...tcomms/proc/check_power_on()
  ..
  var/list/machines = get_machines_to_check() 
  for(var/thing in machines)
    var/.../tcomms/TC = thing
    ....

.../core/core/get_machines_to_check()
  var/list/result = list()
  for(var/obj/machinery/tcomms/core/C in GLOB.tcomms_machines)
    result += C
  return result
...
```
This way you reduce dupe code and make it OO

AffectedArc07: Defining a blank proc just to override it in 2 places seems bad in my eyes, so I ain't doing it.

farie82: That is... how OO works though... look at attackby. attack etc. All pretty empty until behaviour is defined. Or wrench_act etc

AffectedArc07: Yeah but the difference is that has about 1000 re-definitions which all rely on parent calls. These **two** procs are written in very specific ways to iterate through the loop and only get the specific machine.

code/game/machinery/tcomms/core.dm
```diff
@@ -121,6 +128,38 @@
...
+	// Cancel if we are already on
+	if(active)
+		return TRUE
+
+	for(var/obj/machinery/tcomms/core/C in GLOB.tcomms_machines)
+		// Make sure we dont check ourselves
+		if(C == src)
+			continue
+		// We dont care about ones on other zlevels
+		if(C.z != z)
```
Fox-McCloud, line 155: `!atoms_share_zlevel`

AffectedArc07: Thats a helper?

PR #13946, opened 2020-07-28
Click prevention for admin frozen players

xProlithium:
## What Does This PR Do
Prevents click interactions with adminfrozen players. 

## Why It's Good For The Game
Prevents players from attacking, stripping or dragging admin frozen players.

## Changelog
:cl:
del: Outdated admin frozen attack prevention.
add: Click prevention to admin frozen players.
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -2017,5 +2017,18 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	set waitfor = FALSE
 	return call(source, proctype)(arglist(arguments))
 
+/proc/IsFrozen(var/atom/A)
+	if(istype(A, /mob/living))
+		var/mob/living/L = A
+		if(L.frozen)
+			return TRUE
+
+	if(istype(A, /obj/mecha))
+		var/obj/mecha/M = A
+		if(M.frozen)
+			return TRUE
+
+	return FALSE
+
 /// Waits at a line of code until X is true
 #define UNTIL(X) while(!(X)) stoplag()
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -2018,16 +2018,8 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return call(source, proctype)(arglist(arguments))
 
 /proc/IsFrozen(var/atom/A)
-	if(istype(A, /mob/living))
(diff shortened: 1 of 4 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -2017,5 +2017,18 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	set waitfor = FALSE
 	return call(source, proctype)(arglist(arguments))
 
+/proc/IsFrozen(var/atom/A)
+	if(istype(A, /mob/living))
```
SteelSlayer, line 2021: Instead of hard coding these conditions, I would use the `GLOB.frozen_atom_list` list we have. Anything that is frozen will be added to this list.

So instead you can do something like 
```js
/proc/IsFrozen(var/atom/A)
    if(A in GLOB.frozen_atom_list)
        return TRUE
    return FALSE
```

xProlithium: I modified and tested it. I think I re-up'd it correctly.

SteelSlayer:
Also the CL is a bit off, you have to use the tags for it to show up properly. 

del: Removed some stuff
add: Added some stuff

xProlithium:
> Also the CL is a bit off, you have to use the tags for it to show up properly.
> 
> del: Removed some stuff
> add: Added some stuff

Fixed

code/__HELPERS/unsorted.dm
```diff
@@ -2017,5 +2017,10 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	set waitfor = FALSE
 	return call(source, proctype)(arglist(arguments))
 
+/proc/IsFrozen(var/atom/A)
```
Fox-McCloud, line 2020: implied vars are not allowed.

code/modules/mob/mob.dm
```diff
@@ -880,6 +880,9 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 		return
 	if(!Adjacent(usr))
 		return
+	if(IsFrozen(src))
```
Fox-McCloud, line 883: might as well add the admin check here too.

code/__HELPERS/unsorted.dm
AffectedArc07:
```suggestion
/proc/IsFrozen(atom/A)
```

PR #13955, opened 2020-07-29
[TGUI] Supermatter Monitoring Console

AffectedArc07:
## What Does This PR Do
This PR adds in a super matter monitoring console. It is completely independent of modular computers and can be built from R&D. It is currently not mapped in due to the mapping queue and we dont run the supermatter as standard. The screen on the console will update based on the state of the supermatter.

Menu View:
image

Data View:
image

Inactive shard:
image

Active shard:
*the bar graph is animated, see sprite changes*
image

## Why It's Good For The Game
There should be a way to monitor a supermatter without going through the pain and suffering that is making a modular computer or tablet.

## Images of changes
Icondiffbot will handle the icon changes

## Changelog
:cl: AffectedArc07
add: Added a supermatter monitoring console, which can be built from R&D.
/:cl:


taukausanake:
Shouldn't we wait for #11311 to be merged before this? Also just wondering, this detects all SMs no matter where (read: any area) they are placed? I feel like it should only detect SMs in a shared area, not station-wide even if read-only.

This got me thinking of a SM bomb. I wonder how viable that would be...

AffectedArc07:
This was made to aid the supermatter port so monitoring it isn't reliant on a modular computer. Detecting things in all areas is fine, and a necessity since we often make supermatters in atmos or new areas here.

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
+/obj/machinery/computer/sm_monitor
+	name = "supermatter monitoring console"
+	desc = "Used to monitor supermatter shards."
+	icon_keyboard = "power_key"
+	icon_screen = "smmon_0"
+	circuit = /obj/item/circuitboard/sm_monitor
+	light_color = LIGHT_COLOR_YELLOW
+	/// Cache-list of all supermatter shards
+	var/list/supermatters = list()
```
dearmochi, line 9: [Lazy list](https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md#startupruntime-tradeoffs-with-lists-and-the-hidden-init-proc)?

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+
+/obj/machinery/computer/sm_monitor/Destroy()
+	active = null
+	return ..()
+
+/obj/machinery/computer/sm_monitor/attack_ai(mob/user)
+	attack_hand(user)
+
+/obj/machinery/computer/sm_monitor/attack_hand(mob/user)
+	add_fingerprint(user)
```
dearmochi, line 23: I feel like this shouldn't silently add a fingerprint if the computer's broken or without power. The player doesn't get any feedback if that's the case

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+		data["SM_ambienttemp"] = air.temperature
+		data["SM_ambientpressure"] = air.return_pressure()
+		//data["SM_EPR"] = round((air.total_moles / air.group_multiplier) / 23.1, 0.01)
+		var/other_moles = air.total_trace_moles()
+		var/TM = air.total_moles()
+		if(TM)
+			data["SM_gas_O2"] = round(100*air.oxygen/TM,0.01)
+			data["SM_gas_CO2"] = round(100*air.carbon_dioxide/TM,0.01)
+			data["SM_gas_N2"] = round(100*air.nitrogen/TM,0.01)
+			data["SM_gas_PL"] = round(100*air.toxins/TM,0.01)
```
dearmochi, line 61: [Spacing between the operators and comma](https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md#spacing).

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+	else
+		var/list/SMS = list()
+		for(var/obj/machinery/power/supermatter_shard/S in supermatters)
+			var/area/A = get_area(S)
+			if(!A)
+				continue
+
+			SMS.Add(list(list(
+			"area_name" = A.name,
+			"integrity" = S.get_integrity(),
```
dearmochi, line 81: Should indent one tab to the right to make it more readable

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+				data["SM_gas_OTHER"] = 0
+		else
+			data["SM_gas_O2"] = 0
+			data["SM_gas_CO2"] = 0
+			data["SM_gas_N2"] = 0
+			data["SM_gas_PH"] = 0
+			data["SM_gas_OTHER"] = 0
+	else
+		var/list/SMS = list()
+		for(var/obj/machinery/power/supermatter_shard/S in supermatters)
```
dearmochi, line 74: If `supermatters` only contains `/obj/machinery/power/supermatter_shard`s then you can just make the [loop typeless](https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md#istypeless-for-loops) as a minor optimization.

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+  * This proc loops through the list of supermatters in the atmos SS and adds them to this console's cache list
+  */
+/obj/machinery/computer/sm_monitor/proc/refresh()
+	supermatters = list()
+	var/turf/T = get_turf(tgui_host()) // Get the TGUI host incase this ever turned into a supermatter monitoring module for AIs to use or something
+	if(!T)
+		return
+	for(var/obj/machinery/power/supermatter_shard/S in SSair.atmos_machinery)
+		// Delaminating, not within coverage, not on a tile.
+		if(!(is_station_level(S.z) || is_mining_level(S.z)  || atoms_share_level(S, T) || !istype(S.loc, /turf/simulated/)))
```
dearmochi, line 102: One space too much after `is_mining_level(S.z)`

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+
+	switch(action)
+		if("refresh")
+			refresh()
+
+		if("view")
+			var/newuid = params["view"]
+			for(var/obj/machinery/power/supermatter_shard/S in supermatters)
+				if(S.UID() == newuid)
+					active = S
```
dearmochi, line 138: `break` off the loop if the shard was found, no need to loop over the rest.

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+
+		if("back")
+			active = null
+
```
dearmochi, line 142: I'd suggest adding a default case with `else` that returns `FALSE` so that the UI doesn't refresh on an invalid action. Additionally `add_fingerprint` could be called after the switch to indicate that the computer was indeed used.

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+/obj/machinery/computer/sm_monitor/tgui_act(action, params)
+	if(..())
+		return
+
+	. = TRUE
+
+	switch(action)
+		if("refresh")
+			refresh()
+
```
dearmochi, line 133: I personally don't like spacing switch cases with each other, but that's just me.

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -0,0 +1,142 @@
...
+			if(last_status == SUPERMATTER_ERROR)
+				last_status = SUPERMATTER_INACTIVE
+			icon_screen = "smmon_[last_status]"
+			update_icon()
+
+	return TRUE
+
+/obj/machinery/computer/sm_monitor/tgui_act(action, params)
+	if(..())
+		return
```
dearmochi, line 126: Not sure on this, but the absence of a check for if the machine is powerless or broken might allow an user to interact with the machine even when they shouldn't.

tgui/packages/tgui/interfaces/SupermatterMonitor.js
```diff
@@ -0,0 +1,147 @@
...
+  if (pressure > 10000) {
+    return 'bad';
+  } else if (pressure > 5000) {
+    return 'average';
+  } else {
+    return 'good';
+  }
+};
+
+export const SupermatterMonitor = (props, context) => {
```
dearmochi, line 37: I feel like the body of the main TGUI interface should be as compact as possible. You could separate the generated windows in the `if`/`else` into their own sub-interface (in the same file!) so that the main interface is as small as possible. Example:

```js
export const MyInterface = (props, context) => {
	const { act, data } = useBackend(context);
	if (data.screen === 1) {
		return <MyInterfaceScreenOne />;
	} else if (data.screen === 2) {
		return <MyInterfaceScreenTwo />;
	}

	// screen should be clamped server-side to prevent this location from being reached
}

// No export here so that this interface remains private (if it can be used somewhere else, it's better to put it in interfaces/common)
const MyInterfaceScreenOne = (props, context) => {
	const { act, data } = useBackend(context);
	// SCREEN ONE...
}

const MyInterfaceScreenTwo = (props, context) => {
	const { act, data } = useBackend(context);
	// SCREEN TWO...
}
```

tgui/packages/tgui/interfaces/SupermatterMonitor.js
```diff
@@ -0,0 +1,147 @@
...
+            <LabeledList>
+              <LabeledList.Item label="Core Integrity">
+                <ProgressBar
+                  ranges={{
+                    good: [95, Infinity],
+                    average: [80, 94],
+                    bad: [-Infinity, 79],
+                  }}
+                  minValue={0}
+                  maxValue={100}
```
dearmochi, line 102: No real need to make these expressions, you can just wrap them in quotes.

tgui/packages/tgui/interfaces/SupermatterMonitor.js
```diff
@@ -0,0 +1,147 @@
...
+            />
+          }>
+            <LabeledList>
+              <LabeledList.Item label="Core Integrity">
+                <ProgressBar
+                  ranges={{
+                    good: [95, Infinity],
+                    average: [80, 94],
+                    bad: [-Infinity, 79],
+                  }}
```
dearmochi, line 100: I'd make this a `const` at the top of the file.

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: Youre still touching the machine, and every instance of `attack_hand` should be fingerprinted, not just for the detective, but admins too

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: This was copied from the old version but ill fix this one's anyway

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: This was copied from the old version but ill fix this one's anyway

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: This was copied from the old version but ill fix this one's anyway

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: This was copied from the old version but ill fix this one's anyway

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: This was copied from the old version but ill fix this one's anyway

tgui/packages/tgui/interfaces/SupermatterMonitor.js
AffectedArc07: Progressbars must be done like this afaik

tgui/packages/tgui/interfaces/SupermatterMonitor.js
AffectedArc07: That method would have to make 3 callbacks back to BYOND (useBackend), whereas my current approach uses one. Im sticking with the method with the least overhead.

code/game/machinery/computer/sm_monitor.dm
AffectedArc07: Its much more readable this way

PR #13962, opened 2020-07-30
Nerfs ashstorm sounds

TDSSS:
## What Does This PR Do
This PR more or less cuts the volume of ash storms in half

## Why It's Good For The Game
They are annoyingly loud and annoying. I even consciously stop observing people going to lavaland because of how loud and annoying the sound is.

## Changelog
:cl:
tweak: ash storm volume reduced
/:cl:

### DIFF
--- a/code/datums/looping_sounds/weather.dm
+++ b/code/datums/looping_sounds/weather.dm
@@ -8,7 +8,7 @@
 	start_sound = 'sound/weather/ashstorm/outside/active_start.ogg'
 	start_length = 130
 	end_sound = 'sound/weather/ashstorm/outside/active_end.ogg'
-	volume = 80
+	volume = 40
 
 /datum/looping_sound/active_inside_ashstorm
 	mid_sounds = list(
@@ -20,7 +20,7 @@
 	start_sound = 'sound/weather/ashstorm/inside/active_start.ogg'
 	start_length = 130
 	end_sound = 'sound/weather/ashstorm/inside/active_end.ogg'
-	volume = 60
+	volume = 30
 
 /datum/looping_sound/weak_outside_ashstorm
 	mid_sounds = list(
@@ -32,7 +32,7 @@
 	start_sound = 'sound/weather/ashstorm/outside/weak_start.ogg'
 	start_length = 130
 	end_sound = 'sound/weather/ashstorm/outside/weak_end.ogg'
-	volume = 50
+	volume = 20
 
 /datum/looping_sound/weak_inside_ashstorm
 	mid_sounds = list(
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
Count me opposed.

This is a vital sound to communicate what is happening to players and ghosts alike; it needs to be this loud to signal to players what's going on, especially about the existing noises that are going on.

It's really easy for players to turn the volume up/down themselves and just as easy for observers (who can easily mute/unmute things if they find it irritating).

It's fine the way it is.

👎

TDSSS:
It is practically impossible to miss an ash storm. Not only do you have a giant bold red text to warn you beforehand, you also have a screen-covering effect of fuck-you-you-can't-see-shit while it is going on.

If anyone doesn't notice an ash storm is going on they are probably blind and deaf. Not to mention there's even a tool that tells you when the next ash storm is going to hit.

SteelSlayer:
Its easy to miss red text, especially on lavaland when you're fighting a bunch of mobs and the chat window is full of bold *red* attack notifications. And now with the lower sound, combined with there potentially being a bunch of fighting noises, I worry people will miss it.

shmabuu:
If possible this is only really needed on station. When someone gets the storm staff it's really annoying having the storm sounds on the station.

Denthamos:
Having it as a pref to enable/disable would be the best of both worlds when it comes to the ash storm noise.

AffectedArc07:
More prefs can be added after #13936

TDSSS:
I've been told I shouldn't make it a pref by AA, something about being able to turn it off being bad, so I settled for just making it less annoying. *shrug

KathrinBailey:
Hey, I was just hunting the original code of a port of a port a mirror and stumbled across this, I forgot to do this from where I'm from, buuut.
A better way of doing this might be to manually lower the decibels of just these middle sounds in audacity:
image
That way Fox's concerns that players aren't being conveyed the audio hints that a storm is coming are (probably) addressed, because they actually hear it start, but the middle of it is just a little quieter. If you arrive in the middle of the storm, it being quieter shouldn't matter because the darkened tiles should be hint enough that there's a storm.

lolzyking:
If the problem is missing the Red text...
I dunno....Make ash storm text another color? Bright yellow?

Its too loud, it hurts the ears, it's max volume inside buildings, as ghosts, ect.

trololiver112:
Miners who are there to mine needs to hear it loud and clear sure.
But can’t we make it toggleable like instruments for ghosts? Cause that is in my opinion the most annoying part.

Chezmonkey:
100% This should be toggle-able, have the volume for the middle of the storm lowered, SOMETHING, **for ghosts at least**. Relying on the _player_ to mute their volume or adjust their system settings is irresponsible, because everyone's audio setup is unique and they may have different preferences for certain game items as audio cues.

Ashstorms are REALLY CLEAR when they're occurring, but even if players who are alive NEED to be warned that an Ashstorm is coming, there is **no** reason for ghosts and observers to be forced to endure the loud volume.

farie82 (approved):
Code is good

variableundefined (approved):

PR #13971, opened 2020-07-31
Nerfs Lava Staff

Nazsgull:
## What Does This PR Do
Lava staff will check what z level is his user on, and if it's not mining will only spawn a few sparks.

## Why It's Good For The Game
It is good because prevents using it on the station, where it can cause too much destruction.

## Changelog
:cl: Nazsgull
tweak: Lava staff will check what z level is his user on, and if it's not mining will only spawn a few sparks.
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -158,7 +158,7 @@
 
 /obj/item/lava_staff
 	name = "staff of lava"
-	desc = "The ability to fill the emergency shuttle with lava. What more could you want out of life?"
+	desc = "The power of fire and rocks in your hands!"
 	icon_state = "staffofstorms"
 	item_state = "staffofstorms"
 	icon = 'icons/obj/guns/magic.dmi'
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -177,7 +177,6 @@
 	var/create_delay = 30
 	var/reset_cooldown = 50
 	var/timer = 0
-	var/timerMalfunction = 0
 	var/banned_turfs
 
 /obj/item/lava_staff/New()
@@ -194,8 +193,8 @@
 
 	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
 
-		if(world.time > timerMalfunction)
-			timerMalfunction = world.time + create_delay + 1
+		if(world.time > timer)
+			timer = world.time + create_delay + 1
 			user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
 			do_sparks(5, FALSE, user)
 		return
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -179,13 +179,9 @@
 	var/timer = 0
 	var/timerMalfunction = 0
 	var/banned_turfs
-	var/datum/effect_system/spark_spread/spark_system
 
 /obj/item/lava_staff/New()
 	. = ..()
-	spark_system = new /datum/effect_system/spark_spread
-	spark_system.set_up(5, 0, src)
-	spark_system.attach(src)
 	banned_turfs = typecacheof(list(/turf/space/transit, /turf/unsimulated))
 
 /obj/item/lava_staff/afterattack(atom/target, mob/user, proximity_flag, click_parameters)
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -177,10 +177,15 @@
 	var/create_delay = 30
 	var/reset_cooldown = 50
 	var/timer = 0
+	var/timerMalfunction = 0
 	var/banned_turfs
+	var/datum/effect_system/spark_spread/spark_system
 
 /obj/item/lava_staff/New()
 	. = ..()
+	spark_system = new /datum/effect_system/spark_spread
+	spark_system.set_up(5, 0, src)
+	spark_system.attach(src)
 	banned_turfs = typecacheof(list(/turf/space/transit, /turf/unsimulated))
 
 /obj/item/lava_staff/afterattack(atom/target, mob/user, proximity_flag, click_parameters)
@@ -189,6 +194,14 @@
 		return
 
 	if(is_type_in_typecache(target, banned_turfs))
+		return
+
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+		if(world.time > timerMalfunction)
+			timerMalfunction = world.time + create_delay + 1
+			user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
+			do_sparks(5,FALSE,user)
 		return
 
 	if(target in view(user.client.view, get_turf(user)))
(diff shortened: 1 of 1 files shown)


code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -191,6 +196,14 @@
 	if(is_type_in_typecache(target, banned_turfs))
 		return
 
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+		if(world.time > timerMalfunction)
+			timerMalfunction = world.time + create_delay + 1
+			user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
+			do_sparks(5,FALSE,user)
```
Kyep, line 204:
```suggestion
			do_sparks(5, FALSE, user)
```

code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -177,10 +177,15 @@
...
 	var/timer = 0
+	var/timerMalfunction = 0
 	var/banned_turfs
+	var/datum/effect_system/spark_spread/spark_system
 
 /obj/item/lava_staff/New()
 	. = ..()
+	spark_system = new /datum/effect_system/spark_spread
+	spark_system.set_up(5, 0, src)
+	spark_system.attach(src)
```
SteelSlayer, line 188: `do_sparks` essentially covers what you've trying to accomplish here (minus doing `.attach(src)` because it doesn't need to).
```js
/proc/do_sparks(n, c, source)
	var/datum/effect_system/spark_spread/sparks = new
	sparks.set_up(n, c, source)
	sparks.autocleanup = TRUE
	sparks.start()
```
So I recommend either: removing these lines, and the `var/datum/effect_system/spark_spread/spark_system` reference you've created on line 182. 

Or, instead of calling `do_sparks(5,FALSE,user)` on line 204, call `spark_system.start()`

Nazsgull: Gonna go with do_sparks(5, FALSE, user).

code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -191,6 +192,14 @@
...
 		return
 
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+		if(world.time > timerMalfunction)
+			timerMalfunction = world.time + create_delay + 1
+			user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
+			do_sparks(5, FALSE, user)
+		return
+
```
SteelSlayer, line 202: Styling suggestion using an early return instead. If you don't decide to go with this, I still suggest removing the trailing `return`, as it is unnecessary. 
```suggestion
	if(is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
	    return

	if(world.time > timerMalfunction)
		timerMalfunction = world.time + create_delay + 1
		user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
		do_sparks(5, FALSE, user)

```

GentleManners:
Why would you nerf it when its application are so minor? Maybe make the blob immune to larva instead as its an already underwhelming item and that would make it borderline useless. 25 burn damage and no block chance on a melee weapon is bad.

code/modules/mining/lavaland/loot/ashdragon_loot.dm
Crazylemon64: Your suggested change would prevent the staff from working at all on the mining levels

SteelSlayer: You're right I was tunnel visioned on this one code snippet, didn't think about the rest of the proc. Ignore what I said.

code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -191,6 +192,14 @@
 	if(is_type_in_typecache(target, banned_turfs))
 		return
 
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+		if(world.time > timerMalfunction)
```
farie82, line 197: You can just make use of the existing `timer` variable. That way you can just set that and have line 189 take care of the timing

MRgaty:
> Why would you nerf it when its application are so minor? Maybe make the blob immune to larva instead as its an already underwhelming item and that would make it borderline useless. 25 burn damage and no block chance on a melee weapon is bad.

Better than having the possibility to massively grief the station with it.
It will still be useful in lavaland

GentleManners:
If you just want to grief the station there are easier ways than just placing one lava tile every 20 seconds and not really, the staff is kinda a joke in lavaland like what applications does it have when everything is immune to lava

code/modules/mining/lavaland/loot/ashdragon_loot.dm
Nazsgull: Makes sense, I wanted to use two variables to separate the behavior of malfunctioning from the other one, but I think it's way more simple the way you say.

code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -191,6 +191,14 @@
 	if(is_type_in_typecache(target, banned_turfs))
 		return
 
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+		if(world.time > timer)
```
farie82, line 196: This is already checked at line 188. Don't need this here. Just need to set the timer

TheMaveran:
Kind of agree with @GentleManners , if blobs were just made lava immune, item would be fine. This item literally has no usage if it doesn't work on station, so say goodbye to any antag strats or buildings you wanted to use lava with.

Spacemanspark:
The staff has no reason to work on station. Agreeing with the nerf.

code/modules/mining/lavaland/loot/ashdragon_loot.dm
Nazsgull: Solved in 72437017dce98d8019ddb00da6ee04feab2be82a

Varghh:
This has to go, blobs gets literally destroyed with this, we have all been super vocal about how bad lavaland loot is on the station. I don't want lavaland removed because i think it's fun, but making most of the loot from there useless on the station is something we absolutely need. We are not TG, the powergaming this stuff brings up is insane

code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -191,6 +191,14 @@
 	if(is_type_in_typecache(target, banned_turfs))
 		return
 
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+
+		timer = world.time + create_delay + 1
```
farie82, line 197:
```suggestion
	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
		timer = world.time + create_delay + 1
```

code/modules/mining/lavaland/loot/ashdragon_loot.dm
```diff
@@ -191,6 +191,14 @@
...
+		user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
```
farie82, line 198:
```suggestion
		user.visible_message("<span class='danger'>[user]'s [src] malfunctions!</span>")
```
Will need to check if this formats correctly. If not try using `name` instead of `src`

farie82 (approved):

variableundefined (approved):

PR #13981, opened 2020-08-02
Adds Atmospherics Gas Supply area.

### LINKED ISSUE #13980: Swarmers can eat the reinforced walls and grills between main atmos canisters
**Issue Description**:

Swarmers can eat the reinforced walls and grills between main atmos canisters, breaking their laws and causing issues

Some warning and thing to prevent them from doing this

**What happened instead**:

They cause atmos issues and kinda break their laws they sort of have

Screwing over atmos as a swarmers that is heavily coded to NOT do this should not be possible

Be a swarmer
Ventcrawl into main canisters
Starts eating the walls between them
Watch as atmos mix and laws break

SabreML:
## What Does This PR Do

Makes the gas supply tanks in Atmospherics their own seperate area.
Stops swarmers from deconstructing the (internal) walls of the gas tanks. (Fixes #13980)

## Why It's Good For The Game

It should probably have been its own area anyway, since it's not physically connected to the rest of Atmos.
Swarmers really shouldn't be able to mess with atmospherics gas mixes.

## Changelog
:cl:
add: Added the `atmos/gas_supply` area.
tweak: Changed the Atmos gas tanks to their own area.
fix: Swarmers can no longer break the Atmospherics Gas Supply tanks.
/:cl:

I'm sure there's a much easier way of doing this which doesn't involve messing with areas, and if anyone comes up with one I'll close this.

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -28777,7 +28777,7 @@
 	anchored = 1
 	},
 /turf/simulated/floor/engine/n2,
-/area/atmos)
+/area/atmos/gas_supply)
 "bbN" = (
 /obj/effect/decal/warning_stripes/northeast,
 /turf/simulated/floor/plasteel{
@@ -28789,7 +28789,7 @@
 	anchored = 1
 	},
 /turf/simulated/floor/engine/o2,
-/area/atmos)
+/area/atmos/gas_supply)
 "bbP" = (
 /obj/effect/decal/warning_stripes/north,
 /turf/simulated/floor/plasteel{
@@ -29780,7 +29780,7 @@
 	anchored = 1
 	},
 /turf/simulated/floor/engine/air,
-/area/atmos)
+/area/atmos/gas_supply)
 "bdH" = (
 /obj/effect/landmark/start{
 	name = "Civilian"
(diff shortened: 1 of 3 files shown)


BakKeener:
Wait are swarmers able to break Gas Canisters? Or do you mean the supply areas that's in space from atmos? If so there's actually no harm in breaking them since swarmers can't break into space naturally and the station atmos won't run dry fast enough for it to actually be a issue.

SabreML:
> Wait are swarmers able to break Gas Canisters? Or do you mean the supply areas that's in space from atmos?

They're not able to break the actual canisters, but they can break the walls between the different storages and mix stuff together.
Which considering they can't destroy the outer walls, is clearly a bug if nothing else.

Balthusdire:
Good change. Definitely seems like a bug.

TDSSS:
I kind of fear that having a magically, always-powered area on station has some exploit potential.

Muhdi345:
Will Atmosians still be able to use the main atmospherics air alarm to play around with the storage vent pressure checks?

I know you have to set vents to their default setting if you want to achieve that. But if the storage tanks become their own room you'll have to place an air alarm inside the damn things before I can play around with them.

TDSSS:
It might be better to create an rwall subtype, make the tanks out of that and make swarmers unable to deconstruct them, or something like that.

SteelSlayer:
Rwall subtype is a nice idea. Would be simpler and probably create less issues.

SabreML:
> It might be better to create an rwall subtype, make the tanks out of that and make swarmers unable to deconstruct them, or something like that.

This seems like a better idea in general, so I'll do that instead.

PR #13990, opened 2020-08-02
Fix runtime in Hallucination.dm when target.client is null

Bizzonium:
## What Does This PR Do
Fixes runtime in Hallucination.dm when target.client is null (left the game) and deletes object. 
```
[2020-08-02T16:25:00] WORLD: [2020-08-02T16:25:00] Runtime in Hallucination.dm,169: Cannot read null.client
[2020-08-02T16:25:00] WORLD:   proc name: Expand (/obj/effect/hallucination/fake_flood/proc/Expand)
[2020-08-02T16:25:00] WORLD:   src: the fake flood (/obj/effect/hallucination/fake_flood)
[2020-08-02T16:25:00] WORLD:   src.loc: null
[2020-08-02T16:25:00] WORLD:   call stack:
[2020-08-02T16:25:00] WORLD:   the fake flood (/obj/effect/hallucination/fake_flood): Expand()
[2020-08-02T16:25:00] WORLD:   the fake flood (/obj/effect/hallucination/fake_flood): process(20)
[2020-08-02T16:25:00] WORLD:   Objects (/datum/controller/subsystem/processing/obj): fire(0)
[2020-08-02T16:25:00] WORLD:   Objects (/datum/controller/subsystem/processing/obj): ignite(0)
[2020-08-02T16:25:00] WORLD:   Master (/datum/controller/master): RunQueue()
[2020-08-02T16:25:00] WORLD:   Master (/datum/controller/master): Loop()
[2020-08-02T16:25:00] WORLD:   Master (/datum/controller/master): StartProcessing(0)
```

## Why It's Good For The Game
Possibly prevent invalid timer runtime.

## Changelog
:cl: azizonkg
fix: fixed runtime in Hallucination.dm when target.client is left the game. 
/:cl:

### DIFF
--- a/code/modules/flufftext/Hallucination.dm
+++ b/code/modules/flufftext/Hallucination.dm
@@ -159,6 +159,8 @@ Gunshots/explosions/opening doors/less rare audio (done)
 		next_expand = world.time + FAKE_FLOOD_EXPAND_TIME
 
 /obj/effect/hallucination/fake_flood/proc/Expand()
+	if(!target)
+		qdel(src)
 	for(var/turf/FT in flood_turfs)
 		for(var/dir in GLOB.cardinal)
 			var/turf/T = get_step(FT, dir)
--- a/code/modules/flufftext/Hallucination.dm
+++ b/code/modules/flufftext/Hallucination.dm
@@ -149,18 +149,18 @@ Gunshots/explosions/opening doors/less rare audio (done)
 /obj/effect/hallucination/fake_flood/process()
 	if(!target)
 		qdel(src)
+		return
 	if(next_expand <= world.time)
 		radius++
 		if(radius > FAKE_FLOOD_MAX_RADIUS)
 			qdel(src)
+			return
 		Expand()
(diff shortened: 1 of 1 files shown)


code/modules/flufftext/Hallucination.dm
```diff
@@ -159,6 +159,8 @@ Gunshots/explosions/opening doors/less rare audio (done)
 		next_expand = world.time + FAKE_FLOOD_EXPAND_TIME
 
 /obj/effect/hallucination/fake_flood/proc/Expand()
+	if(!target)
+		qdel(src)
```
farie82, line 163: I just noticed I incorrectly fixed this bug a while ago. On line 150 you can see my attempt. The issue comes because the qdel at line 155 does not return.
qdel will not terminate the proc it got called from but will call Destroy. This will null target. Thus causing the bug.
A simple return after line 155 will do. Line 150 and 151 can then also be removed I recon

Bizzonium:
Applied suggested changes but left ifnull check for target.

PR #13991, opened 2020-08-02
Fix runtime in wires.dm when holder is null

Bizzonium:
## What Does This PR Do
Fixes runtime in wires.dm when holder is null because thing is destroyed and user try interact with still open ui panel.

## Why It's Good For The Game
Less runtimes.

## Changelog
:cl: azizonkg
fix: fixed runtime in wires.dm when holder is null because thing is destroyed and user try interact with still open ui panel.
/:cl:

### DIFF
--- a/code/datums/wires/wires.dm
+++ b/code/datums/wires/wires.dm
@@ -141,7 +141,7 @@ GLOBAL_LIST_INIT(wireColours, list("red", "blue", "green", "black", "orange", "b
 	if(..())
 		return 1
 	var/mob/L = usr
-	if(holder && CanUse(L) && href_list["action"])
+	if(CanUse(L) && href_list["action"])
 		var/obj/item/I = L.get_active_hand()
 		var/colour = lowertext(href_list["wire"])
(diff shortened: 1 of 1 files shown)


code/datums/wires/wires.dm
```diff
@@ -194,7 +194,7 @@ GLOBAL_LIST_INIT(wireColours, list("red", "blue", "green", "black", "orange", "b
 	return 1
 
 /datum/wires/CanUseTopic(mob/user, datum/topic_state/state)
-	if(!CanUse(user))
+	if(holder && !CanUse(user))
```
farie82, line 197: Would make more sense here to do this:
`if(!holder || !CanUse(user))`
That way you don't have to check for holder on line 144 either since the UI will be closed before it hits that line

AffectedArc07:
Steelslayer is fixing this with TGUI wires by closing the UI when the object holding the wires is destroyed

Bizzonium:
Applied suggested changes.

PR #13993, opened 2020-08-02
Fix var type of new nuke code in blob gamemode

Bizzonium:
## What Does This PR Do
Fixes type of new nuke code in blob gamemode. Before it was int but should be string.

## Why It's Good For The Game
Now you can actually blow up the station in blob gamemode.

## Changelog
:cl: azizonkg
fix: fixed type of new nuke code in blob gamemode. Now you can actually blow up the station in blob gamemode.
/:cl:


enderman51505:
Does this affect Malf Ais nuke as well?

Bizzonium:
No. Only affects blob game mode.

PR #13994, opened 2020-08-02
Fix and tweak teleportation for mecha

Bizzonium:
## What Does This PR Do
Fix and tweak teleportation for mecha (through CentCom Mass Driver for example).

## Why It's Good For The Game
Now you can actually use CentCom Mass Driver on full power.

## Changelog
:cl: azizonkg
fix: now you can actually use CC Mass Driver to teleport mecha.
/:cl:

### DIFF
--- a/code/game/objects/effects/portals.dm
+++ b/code/game/objects/effects/portals.dm
@@ -14,6 +14,7 @@
 	var/precision = TRUE // how close to the portal you will teleport. FALSE = on the portal, TRUE = adjacent
 	var/can_multitool_to_remove = FALSE
 	var/ignore_tele_proof_area_setting = FALSE
+	var/can_mecha_pass = FALSE
 
 /obj/effect/portal/New(loc, turf/target, creator = null, lifespan = 300)
 	..()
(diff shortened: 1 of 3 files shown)


code/game/objects/effects/portals.dm
```diff
@@ -88,7 +89,7 @@
 	if(!M.simulated || iseffect(M))
 		. = FALSE
 
-	if(M.anchored && ismecha(M))
+	if(!can_mecha_pass && M.anchored && ismecha(M))
```
farie82, line 92: I'm not 100% sure myself why this check is even here. Since in `mecha/Bump` this is written:
image
This code only happens if you walk into a portal. Which is a bit... dumb that it even exists.

Can you check why this check here exists and if it's still required or not. If not then you can remove both this check and the ugly "fix" in `mecha/Bumb()`

Bizzonium:
Removed unnecessary checks

Kyep:
Have you tested this end-to-end, with launching mechas from CC, and it actually working? Reliably?

Last time I tried using the CC mecha launchers, I couldn't find the button to actually fire them.

The time before that, when I did find the button, it launched the mecha only halfway to the teleporter portals. They had to manually move the rest of the way.

Mecha launch from CC has always been very buggy.

Bizzonium:
I had no problems with CC massdrivers. There is maybe only one bug in central console: to see mass drivers controls you should click 'reset' bitton

variableundefined (approved):

PR #13996, opened 2020-08-03
Adds rubber ducky shoes

kugamo:
## What Does This PR Do
This PR adds rubber ducky shoes as an item. It can be worn on the feet, and makes squeaks everytime you walk. It can be crafted using a wirecutters with 2 rubber duckies, and a pair of sandals.

## Why It's Good For The Game
Its stupid fun, plain and simple. Ideally this would add to the charm of the game.

## Images of changes
PaintDotNet_sv3YZ07ZJi
ducker

## Changelog
:cl:
add: Adds Rubber Ducky Shoes, and a crafting recipe for them.
tweak: Adds Rubber Duckies as a prize in the arcade for 80.
imageadd: Adds sprites for item icon, on character, and in hand.
/:cl:


Mitchs98:
Should definitely add more ways to get rubber duckies if these get merged. Maybe not an incredibly easy way...but it'd be pretty hard to get a pair of these otherwise.

code/modules/clothing/shoes/miscellaneous.dm
```diff
@@ -343,3 +343,11 @@
 		recharging_time = world.time + recharging_rate
 	else
 		to_chat(user, "<span class='warning'>Something prevents you from dashing forward!</span>")
+
+/obj/item/clothing/shoes/ducky
+	name = "rubber ducky shoes"
+	desc = "Two rubber duckies made to fit."
+	icon_state = "ducky"
+	item_state = "ducky"
+	var/footstep = 1
```
SteelSlayer, line 352: What is this var for? It's not used anywhere.

kugamo: well I know next to nothing about byond code, I just noticed that for the other shoes with sounds, like jackboots and clownshoes, had it. So I copied it over. I never tested to see if it still made the noises without it, I just assumed it was necessary.

kugamo: OK so sounds work fine without that var, and it is seemingly totally pointless. (for rubber ducky shoes at least)

SteelSlayer: Aha I see. After some research, those vars on the other shoes are also unused, because shoe step sound is handled elsewhere and it never even uses the `footstep` var.

kugamo: must be an artifact or something

kugamo:
PaintDotNet_sg2u5vYuSX
dreamseeker_7sMR6oiIk9
I think this covers all issues for now, I am treating this as finished for now. Unless I need to change the merch console additons, its done.

taukausanake:
A soap or wash crate wouldn't be bad. Maybe include with the janitor crate. Just a rubber ducky crate wouldn't be good unless it came with more than a couple but too much wouldn't be good either

datlo:
I think it'd be better to have them be in a cargo crate, that way there's a way to regulate/slow down ducky spam.

code/modules/clothing/shoes/miscellaneous.dm
```diff
@@ -343,3 +343,10 @@
 		recharging_time = world.time + recharging_rate
 	else
 		to_chat(user, "<span class='warning'>Something prevents you from dashing forward!</span>")
+
+/obj/item/clothing/shoes/ducky
+	name = "rubber ducky shoes"
+	desc = "These shoes are made for quacking, and thats just what they'll do."
+	icon_state = "ducky"
+	item_state = "ducky"
+	shoe_sound = list('sound/items/squeaktoy.ogg' = 1)
```
farie82, line 352: `shoe_sound` does not expect a list. It expects a string which is the sound.
Example
image

Spacemanspark:
I can't deny this earned a chuckle from me.
I do think it shouldn't be _too_ easy to get so there's not a billion spamming ducklings running around, but I like this.
:+1:

code/modules/clothing/shoes/miscellaneous.dm
kugamo: I had trouble getting sounds using shoe_sound = squeaktoy or shoe_sound = squeaktoy.ogg, so I tried using what code I found for the rubberducky (which used a list for whatever reason). After some testing I found that using a filepath works fine, so I just commmited that.

kugamo:
So I removed rubber ducky as an entry in the merch console, and added it to the arcade. I moved it to the arcade because ~~I couldn't find where the crate code is~~ its more of a hurtle to get it than just $300, and I still want it some what accessible. You would still have to get 2 to make the shoes, which in total amounts to 160 tickets.
dreamseeker_XjRLpNmhxb

KasparoVy:
The quackers, they're adorable

variableundefined (approved):
@kugamo Will merge when deconflicted.

Spent 2 minutes running around in circle in an empty station with this stuff.

PR #13998, opened 2020-08-03
Wormhole Jaunters can no longer teleport megafauna to station, unless emagged

Kyep:
## What Does This PR Do
Alters the "wormhole jaunter" item that shaft miners use.
By default, the portals it creates will no longer be able to teleport megafauna to the station.
If you emag the jaunter though, its portals will then be able to teleport megafauna.

## Why It's Good For The Game
Right now, hijack traitors can teleport megafauna onto the station, effectively weaponizing megafauna against the station, without having to pay a single TC to do so. This is overpowered. Especially when you consider that the station has no idea which of the miners is doing it, and realistically isn't going to be able to hunt them down in lavaland to stop them. Especially if they do it repeatedly and move half the megafauna of lavaland to the station.
If hijackers want to use megafauna as their own personal army, they should either have to pay some TC to enable this tactic (such as the 6 TC for an emag) or they should have to take some other substantial risk (like stealing the more powerful hand tele which doesn't need to be emagged to be capable of moving megafauna).

Regular non-antag miners shouldn't be able to teleport megafauna to the station at all. If a non-antag does it deliberately, its self-antagging - but its incredibly hard to prove they did it deliberately if they claim it was a mistake. Which they often do.
While dying is part of the game, shaft miners being able to accidentally wipe out half the station by teleporting a colossus, drake or bubblegum to the station with an ill-placed jaunt tunnel is not reasonable.

## Changelog
:cl: Kyep
balance: the wormhole jaunter is no longer able to send megafauna to the station, unless you emag it.
/:cl:

### DIFF
--- a/code/modules/mining/equipment/wormhole_jaunter.dm
+++ b/code/modules/mining/equipment/wormhole_jaunter.dm
@@ -11,6 +11,7 @@
 	throw_range = 5
 	origin_tech = "bluespace=2"
 	slot_flags = SLOT_BELT
+	var/emagged = FALSE
 
 /obj/item/wormhole_jaunter/attack_self(mob/user)
 	user.visible_message("<span class='notice'>[user.name] activates the [name]!</span>")
@@ -43,6 +44,7 @@
 		return
 	var/chosen_beacon = pick(L)
 	var/obj/effect/portal/jaunt_tunnel/J = new(get_turf(src), get_turf(chosen_beacon), src, 100)
+	J.emagged = emagged
 	if(adjacent)
 		try_move_adjacent(J)
 	else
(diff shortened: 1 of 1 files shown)


TheMaveran:
Doesn't really affect the balance side of things, because Hijackers basically require an emag anyway. Will definitely stop megafauna from accidentally ruining rounds though.

Mitchs98:
I love you.

datlo:
Good change. It's very easy to accidentally bring megafauna to the station without intending to do so as a non-antag, speaking from experience. *cough cough*

This keeps it an option, but a very deliberate one. I like it.

Balthusdire:
Excellent change. Good to not punish new players for panic jaunting out of sudden megafauna encounters.

KasparoVy:
Man I'm so glad I wasn't in these rounds.

Gatchapod:
There is just one problem with this. It actually makes fighting megafauna a lot more forgiving. Significantly more forgiving.
Guilty as charged, I once did bring an ash drake through a wormhole by accident. Ever since, I have learned to just not use them while fighting megafauna. _I may have also practiced wormholing Bubblegum in case I ever get a hijack, but that's a side note._

The important thing is that this PR will remove the only drawback jaunters have. An ability to get away after a major screw-up without any risk, just so you can have a round 2, round 3 or even round 4 will basically turn fighting megafauna into an attrition war that any player can tackle quite easily.

It's not enough to warrant rejecting this PR, as an accidental colossus on the station is basically a round-ender with how much station damage it causes. I just do think jaunters should receive some nerfs alongside this change. Perhaps a significant cost increase or a chance to space you?

Pixelzilla:
Not really a fan for a lot of the same reasons that Gatchapod stated but also because having something be mechanically limited to traitors only makes it pretty impossible to have any sort of plausible deniability if you are a traitor using this proposed change/feature. If megafauna ends up on station after this is merged there is absolutely no way to deny it was accidental or just miner stupidity because it will be mechanically impossible for it to be anything otherwise and will just result in all of the miners being hauled in and that might suck for the times there are multiple antags in mining that will inevitably end up being caught due to the unavoidable meta.

I don't really think the number of times this happens is great enough to justify this sort of PR and it seems like if it is a recurrent problem with individuals then perhaps the individuals doing it as non-antags should instead be the ones punished via mining/antag bans? Can we stop trying to babyproof everything in this crazy space game?

Kyep:
@Pixelzilla A single megafauna on station is enough to get the shuttle called, and additionally kill almost anyone who encounters it. A good chunk of the crew can die to this. In this sort of situation, expecting sec to search every miner, find the one traitor minor, the traitor miner to have not stashed their emag in say a storage implant, etc, is just not realistic.

AffectedArc07 (approved):
This has been open for long enough

PR #13999, opened 2020-08-03
Stops Swarmers deconstructing high pressure/temp areas

### LINKED ISSUE #13980: Swarmers can eat the reinforced walls and grills between main atmos canisters

SabreML:
## What Does This PR Do

_New version of #13981_ 
1. Adds a check to `swarmer_act` for pressure over 10x normal pressure to stop Swarmers from deconstructing the Atmoshperics Holding Tanks.
As it is right now they can destroy the internal walls, but not the external ones, so it seems to me like an oversight.
Fixes #13980 
 2. This also has the effect of stopping swarmers from deconstructing the walls of the Turbine when it's in use, or any other burn chamber for that matter.

 3. Changed `return FALSE` flag to `return TRUE` if an AI Swarmer tried deconstructing a door near space, since that added _all_ doors to a blacklist.

## Why It's Good For The Game

Swarmers really shouldn't be able to mess with atmospherics, or spread superheated gas around an area (in the case of a turbine breach).

## Changelog
:cl:
add: Adds a pressure check to `swarmer_act` to stop areas of high pressure getting deconstructed.
tweak: Swarmers can no longer breach the Turbine (or any) Burn Chamber if it is in use.
fix: Swarmers can no longer deconstruct inner walls of the Atmospherics Holding Tanks.
fix: Fixes AI Swarmers completely ignoring all doors if one of them tried to eat an airlock.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
@@ -260,14 +260,19 @@
 	S.DisIntegrate(src)
 	return TRUE
 
-/obj/machinery/door/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)
+/obj/machinery/door/swarmer_act(mob/living/simple_animal/hostile/swarmer/S, var/datum/gas_mixture/air_contents)
 	var/isonshuttle = istype(get_area(src), /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
 			S.target = null
-			return FALSE
+			return TRUE
+		else if(pressure_check.return_pressure() > 5000)  //If pressure > 5000kPa
+			to_chat(S, "<span class='warning'>An inhospitable area may be created as a result of destroying this object. Aborting.</span>")
+			S.target = null
+			return TRUE
 		else if(istype(A, /area/engine/supermatter))
 			to_chat(S, "<span class='warning'>Disrupting the containment of a supermatter crystal would not be to our benefit. Aborting.</span>")
 			S.target = null
@@ -370,9 +375,14 @@
 /turf/simulated/wall/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)
 	var/isonshuttle = istype(loc, /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
+			S.target = null
+			return TRUE
+		else if(pressure_check.return_pressure() > 5000)  //If pressure > 5000kPa
+			to_chat(S, "<span class='warning'>An inhospitable area may be created as a result of destroying this object. Aborting.</span>")
 			S.target = null
 			return TRUE
 		else if(istype(A, /area/engine/supermatter))
@@ -381,12 +391,17 @@
 			return TRUE
 	return ..()
 
-/obj/structure/window/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)
+/obj/structure/window/swarmer_act(mob/living/simple_animal/hostile/swarmer/S, var/datum/gas_mixture/air_contents)
 	var/isonshuttle = istype(get_area(src), /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
+			S.target = null
+			return TRUE
+		else if(pressure_check.return_pressure() > 5000)  //If pressure > 5000kPa
+			to_chat(S, "<span class='warning'>An inhospitable area may be created as a result of destroying this object. Aborting.</span>")
 			S.target = null
 			return TRUE
 		else if(istype(A, /area/engine/supermatter))
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
@@ -367,6 +367,10 @@
 	to_chat(S, "<span class='warning'>Disrupting this energy field would overload us. Aborting.</span>")
 	return FALSE
 
+/turf/simulated/wall/r_wall/swarm_proof/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)//Used in the Atmospherics Gas Holding Tanks
+	to_chat(S, "<span class='warning'>An inhospitable area may be created as a result of destroying this object. Aborting.</span>")
+	return FALSE
+
 /turf/simulated/wall/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)
 	var/isonshuttle = istype(loc, /area/shuttle)
 	for(var/turf/T in range(1, src))
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
@@ -269,7 +269,7 @@
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
 			S.target = null
 			return TRUE
-		else if(pressure_check.return_pressure() > 10*ONE_ATMOSPHERE)  //If pressure > 1013.25kPa
+		else if(pressure_check.return_pressure() > 10 * ONE_ATMOSPHERE)  //If pressure > 1013.25kPa
 			to_chat(S, "<span class='warning'>An inhospitable area may be created as a result of destroying this object. Aborting.</span>")
 			S.target = null
 			return TRUE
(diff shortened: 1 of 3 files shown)


Kyep:
Why not just make it check all turfs adjacent to a wall you plan to deconstruct, and if any of them have atmos unsafe for humans, you can't deconstruct that wall?
That would handle the case of deconstructing rooms next to special gas tanks. But it would also stop them deliberately spreading fire, or punching open walls leading to space, etc.

SabreML:
> 1. When deconstructing a wall as swamer, check all adjacent turfs.
> 2. If any of them are space ... you can't deconstruct the wall.

This is how it's already done for things like Windows and Walls, to stop swarmers from causing hull breaches, it doesn't work in the Atmospherics tanks since the interior walls aren't near space.

> ... or have air that humans can't breathe, ...

This might work, but I imagine it would be very complicated code. Especially considering that the standard station air is ~80% nitrogen, which isn't breathable for humans.

code/game/turfs/simulated/walls_misc.dm
```diff
@@ -50,6 +50,11 @@
 	icon = 'icons/turf/walls/coated_reinforced_wall.dmi'
 	max_temperature = INFINITY
 
+/turf/simulated/wall/r_wall/swarm_proof  // Used to stop swarmers messing with atmos mixes
+	name = "reinforced wall"
```
farie82, line 54: All the values below are dupes from the actual r_wall and are not required due to how inheritance works

SabreML: So just leave it blank?

SabreML:
> Another solution would be to check the area of the tanks. Not sure if they have their own area but that might be another solution. That is also how outer walls are done

I tried doing that in the first PR #13981, but that had a lot of downsides too.

> I'd say try to see if you can check if the air surrounding the wall is deadly. The atmos tanks have high high pressure which is something you can check.

I'll see about making it a pressure thing instead.

code/game/turfs/simulated/walls_misc.dm
farie82: Yep yep. Just a blank `/turf/simulated/wall/r_wall/swarm_proof` will do. Can even do without but it's cleaner to at least have it defined somewhere and not just in the proc

Fox-McCloud:
I'd rather not add yet another wall subtype PURELY for one single antag that's used in exactly 1 spot on the entire map.

SabreML:
Changed it to a pressure check.

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
```diff
@@ -260,14 +260,19 @@
 	S.DisIntegrate(src)
 	return TRUE
 
-/obj/machinery/door/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)
+/obj/machinery/door/swarmer_act(mob/living/simple_animal/hostile/swarmer/S, var/datum/gas_mixture/air_contents)
```
farie82, line 263: var is implied with params. Can remove that

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
```diff
@@ -260,14 +260,19 @@
...
 	var/isonshuttle = istype(get_area(src), /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
 			S.target = null
-			return FALSE
+			return TRUE
+		else if(pressure_check.return_pressure() > 5000)  //If pressure > 5000kPa
```
farie82, line 272: Magic number here. Best to make use of a define or make one yourself. `ONE_ATMOSPHERE` for instance is the define for the value of one atmosphere. Found in __DEFINES/atmospherics.dm

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
```diff
@@ -370,25 +375,35 @@
 /turf/simulated/wall/swarmer_act(mob/living/simple_animal/hostile/swarmer/S)
 	var/isonshuttle = istype(loc, /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
 			S.target = null
 			return TRUE
+		else if(pressure_check.return_pressure() > 5000)  //If pressure > 5000kPa
```
farie82, line 384: Same here

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
```diff
@@ -370,25 +375,35 @@
...
+/obj/structure/window/swarmer_act(mob/living/simple_animal/hostile/swarmer/S, var/datum/gas_mixture/air_contents)
 	var/isonshuttle = istype(get_area(src), /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
 			S.target = null
 			return TRUE
+		else if(pressure_check.return_pressure() > 5000)  //If pressure > 5000kPa
```
farie82, line 403: Magic number again

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
```diff
@@ -260,14 +260,19 @@
...
+/obj/machinery/door/swarmer_act(mob/living/simple_animal/hostile/swarmer/S, var/datum/gas_mixture/air_contents)
 	var/isonshuttle = istype(get_area(src), /area/shuttle)
 	for(var/turf/T in range(1, src))
+		var/datum/gas_mixture/pressure_check = T.return_air()
 		var/area/A = get_area(T)
 		if(isspaceturf(T) || (!isonshuttle && (istype(A, /area/shuttle) || istype(A, /area/space))) || (isonshuttle && !istype(A, /area/shuttle)))
 			to_chat(S, "<span class='warning'>Destroying this object has the potential to cause a hull breach. Aborting.</span>")
 			S.target = null
-			return FALSE
+			return TRUE
```
farie82, line 271: Done here to fix AI swarmers not being able to destroy doors once another swarmer hit a wall next to space or such.
Please note this in the PR description and CL

SabreML: I'm not sure what the change here is, since it's already set to `return TRUE`?

farie82: You've changed it from `FALSE` to `TRUE`. Which is correct but you've not mentioned this in the CL and description of your PR. It's a fix since now AI swarmers will stop deconstructing doors once another AI swarmer tries to decon a door adjacent to space or such.

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
```diff
@@ -263,11 +263,16 @@
...
+		else if(pressure_check.return_pressure() > 10*ONE_ATMOSPHERE)  //If pressure > 1013.25kPa
```
farie82, line 272: Spacing here. When using an operator such as * or + and such then you have to put some space between the 2 sides.
So `X * Y`

code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
SabreML: I think I'm misreading something here, doesn't `return TRUE` mean that AI swarmers will destroy whatever it is?
Which would mean that the `return TRUE` in the SM containment would be a bad thing.

This is what I'm looking at:
> //return TRUE/FALSE whether or not an AI swarmer should try this swarmer_act() again, NOT whether it succeeded.

farie82: It will use that value to check if it should try any objects of the same type again. If you make it return FALSE here it will never try to eat doors again. Some for the walls. It will add the type of the object to a blacklist

SabreML:
> I don't want an extra turf type solely for the internal barriers between atmos rooms.

I reverted that, it's all a check now.
> Adding an extra check every time a swarmer_act on a wall... seems excessive.

The Swarmer code is pretty much entirely made up of `if` statements, I don't know much about the performance side of SS13 but I wouldn't think adding another wall check to the two already there would add lag. As I said though, I don't know much about it.

> And simply letting swarmers break the tanks.... I highly doubt that its gonna matter. Breaking the walls between say the plasma/N2O rooms isn't gonna matter as those are not used. Only the N/O2/Air tank wall breaches matter, and even then there's probably still enough O2 in the mix if you combine them to dispense breathable air into the station.

Due to how Atmospherics is laid out it's actually surprisingly hard to mix bad stuff into the air tank, so that's not really the issue.

In terms of Swarmers though, trying to disassemble the outer tank walls brings up a warning, but the inner walls are fine?
This seems like an oversight to me, and this is, I think, a good way of fixing that.

Since it checks for pressure, it also has the side effect of stopping swarmers from disassembling the Turbine walls when it's in use, and also any other burn chambers. (e.g: SM chamber, Toxins mix chamber)

I personally think that letting Swarmers spread superheated air around the station a bit of an issue, if nothing else because it violates their third law.
> 3. Biological and sentient resources will be harvested at a later date, do not harm them.

SabreML:
Updated Title and Description to better reflect what I'm aiming for.

Kyep (commented):
Dismissing my objection. 
I still don't support this, but I no longer oppose it. I'm neutral on it.

variableundefined:
Objection - I don't think this is worth the code hassle to fix it.

AffectedArc07:
This rarely happens ingame, and even if you breach the tanks together, theres no way to breach plasma into the air supply. Atmospherics is specifically designed in this way. Not in favour.
image

PR #14002, opened 2020-08-03
Removes 80 character line limit from TGUI

AffectedArc07:
## What Does This PR Do
Removes the 80 character line limit. It was a pain in the ass to deal with and makes code readability worse

## Why It's Good For The Game
Easier TGUI creation

## Changelog
N/A

### DIFF
--- a/tgui/.eslintrc.yml
+++ b/tgui/.eslintrc.yml
@@ -366,13 +366,13 @@ rules:
   ## Enforce a maximum depth that blocks can be nested
   # max-depth: error
   ## Enforce a maximum line length
-  max-len: [error, {
-    code: 80,
+  #max-len: [error, {
+    #code: 80,
     ## Ignore imports
-    ignorePattern: '^(import\s.+\sfrom\s|.*require\()',
-    ignoreUrls: true,
-    ignoreRegExpLiterals: true,
-  }]
+    #ignorePattern: '^(import\s.+\sfrom\s|.*require\()',
+    #ignoreUrls: true,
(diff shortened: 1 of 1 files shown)


AffectedArc07:
22 TGUI PRs have been made with this limitation in place, including baystruments and the medical PRs (Over quarter of our conversion work)

Put shortly, this isnt needed. We can work around it, and already do just fine.

Kyep:
I strongly disagree.
As (I believe) the guy who'se created the most TGUI PRs, I think this is change very much still needed.

Reasons:
1) The length limit is unreasonable. Our monitors can hold way more characters than the current limit.
2) The length limit consumes a substantial portion of our time when creating TGUI PRs. We have to go to significant effort just to get around this pointless limit.
3) The length limit results in less readable code. It creates situations where what would have only taken me 1 line, instead takes 5 lines. It discourages me from using longer (but informative!) var names like "suggested_prisoner_names" and instead forces me to use shorthands like "snps" which are less readable. Or to use even more lines. I've had multiple comments on my TGUI PRs expressing confusion about something which I was forced to add in order to work around the line length limit. Had the limit not been in place, much time/confusion would have been saved.

I know the github (and you) have many PRs, and its nice to cross one off the list, but speaking as the guy who'se done the most TGUI PRs out of anyone (I believe), I can tell you that the line length limit most definitely DOES make TGUI conversions unnecessarily more time-consuming, annoying and difficult, and if we want more people to take up the task of doing TGUI conversions, or just have more readable/concise TGUI code in general, this would really help.

PR #14003, opened 2020-08-03
Fix colored lips drawing over masks and random lipstick colouring

dearmochi:
## Context
PR #13911 fixed lipsticks not having a visual effect. As it turns out, they are being rendered in the wrong place (base body icon rather than the head organ) which resulted in two things:
- Them rendering over masks due to layering
- Them not showing up on detached heads when they should.

## What Does This PR Do

- Fixes the above issues by moving the lip rendering code to the same place as where eyes are rendered (in the head organ). _Fixes #13986_
- Opportunistically fixes eyes icon being created redundantly in `/mob/living/carbon/human/proc/get_eyecon()`
- Additionally, this PR integrates PR #14113: it fixes random lipsticks always appearing white because the RGB value (rather than the colour name) was being used in the `colour` variable.

## Why It's Good For The Game

- Lipsticks now render properly on both mob and detached head
- Lipsticks (hopefully) fixed for good
- Very minor performance/memory improvement when rendering human eyes.

## Changelog
:cl:
fix: Fix lipsticks rendering over masks
fix: Fix random lipstick colour always appearing white
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/cosmetics.dm
+++ b/code/game/objects/items/weapons/cosmetics.dm
@@ -55,16 +55,15 @@
 
 /obj/item/lipstick/random/Initialize(mapload)
 	. = ..()
-	var/lscolor = pick(lipstick_colors) // A random color is picked from the var defined initially in a new var.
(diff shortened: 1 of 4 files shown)


AffectedArc07 (approved):
Tested and working

PR #14009, opened 2020-08-04
Alters mining conscription kits to remove the mining access card

Kyep:
## What Does This PR Do
Removes the "mining access card" from the "mining conscription kit".
The mining conscription kit is a duffelbag full of various mining gear, that you can obtain by:
- using a mining voucher (which all miners start with) on the mining equipment vendor
- spending 1500 mining points (less than the cost of a single mining hardsuit) at the mining equipment vendor

The conscription kit contains many other items, besides this access card. Including things like a scanner, mesons, pickaxe, ore bag, cargo encryption key, mask, kinetic accelerator, survival knife, seclite torch, etc. This PR doesn't change any of that, it only removes the access card.

The mining access card is an item which grants ACCESS_MINING, ACCESS_MINING_STATION, ACCESS_MINERAL_STOREROOM, ACCESS_CARGO to your ID.

## Reasons For Change

1. Permanent job transfers need to go through the HOP or Captain. We don't even allow actual heads of department (like the CE, HOS, CMO, etc) to hire people into their department without getting the HOP/Captain to do the transfer. Why should we allow shaft miners to effectively hire extra shaft miners, when even the heads of other departments can't just hire people for their department without going through HOP/Captain? The mining access card item is wildly inconsistent with how we handle permanent job transfers.

2. For granting temporary access to a department to help with emergencies, we already have the far superior guest pass system. There's even a guest pass terminal available in cargo. Why are guest passes better? Because they let you customize the details (exactly what access to grant, for how long, to whom, etc). They're logged. They're visible. And they're truly temporary (limited to 30m max duration).

3. The current way this access is granted by the card, leaves a ton of loose ends. Here are a few examples:
- The new "miner" doesn't have their transfer logged anywhere, which means they don't show up as a transfer in ID computer log, guest pass log, etc. If you put their ID into an ID computer, it reads like an agent ID (normal ID with additional access it shouldn't normally have, with no explanation why). 
- The new "miner" is not considered a transfer under the playtime system, meaning they don't get department playtime correctly. This is even potentially exploitable. Want to earn enough playtime in a department to unlock a head job, without actually playing that department? Get a conscription kit and go mining. You continue to earn playtime for your original job/department, even if you're off mining on lavaland instead.
- The new "miner" doesn't have their old job freed up, which can lead to issues if they're an in-demand or limited role, like chaplain, medical chemist, etc. It also misrepresents the number of actual miners, since they're not counted as a miner either for job slot purposes.
- The new "miner" keeps the job title and sec HUD icon of their old job, which leads to massive confusion if they walk around in full miner gear, especially including contraband items from lavaland.

4. There are balance issues with allowing this sort of "transfer".
- For antags, all they need to do is beg, borrow or steal a voucher, and they get to become a bootleg miner. Miner access... including access to all the goodies of lavaland... but not listed as a miner, no log of the transfer, command doesn't know, etc. 
- For non-antags, this is very abusable by metafriends. All you need is a friend in mining, and they can effectively transfer you into mining almost at will, bypassing HOP/Captain/etc. Actually, its even better than a transfer for you, since you don't lose any of your old access/tools/gear, don't have to worry about anyone replacing you at your old job, and don't have to get permission from anyone except your friend in mining. No oversight.

5. There are design & admin issues.
- This item makes it more difficult for us to balance the number of job slots mining has. Since we have to account design-wise for the possibility that miners can self-replicate... creating additional people with permanent mining access / bootleg job transfers to mining.
- This item could allow people to bypass job bans. Normally job-banned players can't join as jobs they're job-banned from, and if they go to the HOP to get a transfer to them, we get an alert. This access card doesn't have that safety, though. So someone who we job banned from mining, could still use this item to get "transfers" to mining on a regular basis in a way that never alerted us they were violating their job ban.

## Changelog
:cl: Kyep
tweak: mining conscription kits no longer contain a card you can use to get mining access. 
/:cl:


procdrone:
on point 2, the guest pass system i feel could do with an overhaul. The UI is incredibely clunky. Also, i feel like "timed" guestpasses are bad, and guestpass computer should have a list of active passes that can be revoked at will by head, instead of arbitral limit.

Kyep:
@procdrone you might even be right, but (A) guest passes still have lots of functionality like logging that mining access card doesn't, and (B) reworking guest passes is out of scope for this PR.

procdrone:
> @procdrone you may be right, but (A) guest passes still have lots of functionality like logging that mining access card doesn't, and (B) reworking guest passes is out of scope for this PR.

Of course im not discrediting your points, just pointing out that if we are going to rely on emergency mining on guest pass system, it could be brushed up to standard.

AffectedArc07 (approved):

PR #1835, opened 2015-08-23
Let players modify access of individual turrets

Markolie:
Changes:
1) Players can now modify the req_access/req_one_access on turrets that aren't connected to a turret controller. This allows players to set up turrets for specific areas such as the armory.
2) Fixes a runtime with communication computers when an AI tries to log in.

PR #14011, opened 2020-08-05
[TGUI] Turrets & Turret Controls

### LINKED ISSUE #12253: Laser tag turrets don't work!
**Issue Description**:

When you finish a turret with a laser tag gun in it, it doesn't become a `porta_turret/tag` and inherit those properties, meaning you CAN'T TAG FORTRESS *scream

My tag turrets to work correctly and have the right path.

**What happened instead**:

They didn't.

*scream

1) Create a turret, but use a laser tag gun instead of a normal gun.

?????

**Extra information**:

Admins showed that the porta_turret was the default pathing, and required sec/head access.

Kyep:
## What Does This PR Do, and Why
Ports turrets (porta_turrets, syndi turrets, lasertag turrets, CC turrets, etc) and area turret controls to TGUI.
This is good because... less nanoui.

~~Removes the ability to customize the access that a turret checks for when deciding whether to fire on someone or not. While that sounds like a neat feature, in practice it was almost useless, as you could only do it on turrets outside of an area controller, which in practice meant that you could only do it if building new turrets from scratch, which is never done.~~ Ended up porting this too, using TGUI's AccessList.js component.

Adds the ability to have turrets target cyborgs.
This is good because it allows, for example, AI to set its turrets to target borgs so that a single syndiborg / emagged borg / emagged drone / etc can't just waltz into the AI core and kill the AI by itself with zero opposition.

Fixes lasertag turrets being buggy to the point of being unusable. Most notably:
1. Lasertag turrets no longer generate a runtime when spawned.
2. Building a lasertag turret as a crew member no longer incorrectly results in a non-lasertag turret with a lasertag gun. That was extremely useless, as it targeted things using the same rules as a normal turret (ignored lasertag armor, etc) but did so with an ineffective lasertag projectile. Now, you can actually build real lasertag turrets, and they actually work.
3. Fixed lasertag turrets still having the pop-up cover armor appearing and disappearing into thin air.
Fixes  #12253

Refactors a fair bit of the turret and turret control code, most notably replacing a lot of 1/0 variables with TRUE/FALSE.

You can now open the turret / turretcontrol interface even if you can't change anything. So, you can see what the controls are set to.

The description of some of the turret controls is now much clearer about what those settings actually do.

## Images of changes
Default view (locked):
turrets_aiupload_locked
Default view (unlocked):
turrets_aiupload_unlocked
As previously, different options set:
turrets_aiupload_different_options_set
A customized turret, outside of a turret control, with extra options:
turrets_custom_setup

## Changelog
:cl:
add: Ported turrets and turret controls to TGUI, with better interface and more clarity.
add: Added the ability to set turrets to target borgs and drones.
fix: Fixed lasertag turrets not working in several different ways.
/:cl:


TDSSS:
Does the 'Everyone' option include cyborgs and other silicons? Does it include AI cores? If not, it's probably not the best name.

Kyep:
@TDSSS That option has been renamed to "All Non-Synthetics".

AffectedArc07:
I would put this on a new line under its own list item. It looks like the options are merged into the same thing, despite them not being.
image

tgui/packages/tgui/interfaces/PortableTurret.js
```diff
@@ -0,0 +1,102 @@
...
+    neutralize_criminals,
+    neutralize_all,
+    neutralize_unidentified,
+    neutralize_cyborgs,
+  } = data;
+  return (
+    <Window
+      width={300}
+      height={300}>
+      <Window.Content>
```
AffectedArc07, line 26: These are TGUI-V4 properties. Remove the `Width` and `Height` properties

tgui/packages/tgui/interfaces/PortableTurret.js
```diff
@@ -0,0 +1,102 @@
...
+        <NoticeBox>
+          Swipe an ID card to {locked ? 'unlock' : 'lock'} this interface.
+        </NoticeBox>
+        <Fragment>
```
AffectedArc07, line 30: Unsure if this needs to be inside a `Fragment`

Kyep: Removed

Kyep:
@AffectedArc07 Lethals option moved to its own line. Screenshots updated.

AffectedArc07:
@Kyep is it too much to ask to have an `X` (times) icon on the lethals button when they are off, and maybe a warning sign (exclamation-triangle) when lethals are on?

Would make the UI look more polished and complete since every other button has an icon

Kyep:
@AffectedArc07 Done

AffectedArc07:
Also another thing I noticed after a while
image
All options are capitalised Like This For Each Word

Whereas `No sec record` is not

procdrone:
How about making turrets able to target blob turfs as well? blobs going into AI sat for free and easy growth is too cheesy.

Markolie:
Just because building turrets is seldomly done, doesn’t mean the access feature should be removed. Especially in a pull request that just overhauls the UI. In my opinion, if you’re going to do a conversion, at least go the full mile and do everything. Don’t cut something out that works perfectly fine because it takes time. Otherwise, just leave it as is. Building cool stuff and these nifty features with so many possibilities is what makes SS13 great.

Pretty sure TG has an airlock electronics tgui that you can just copy the interface from to save 90% of the work for access setting here.

Kyep:
@procdrone I don't think it would work. Usually when blob mice burst on the AI sat, they burst right next to the AI, which gives the AI zero time to react. Even if turrets could shoot blob turfs, they could still burst next to the turrets to kill them. And that's without even talking about reflective blobs, corners, blobbernauts, etc. Ultimately, I don't think its practical to make turrets a good anti-blob defense.

@Markolie TGUI PRs are allowed to add/remove features as part of the port. Sometimes I'll see a feature that TG has, and which I feel is valuable enough to port in alongside TGUI, like the ability to target silicons. Other times, I will see a feature we have and which I feel is worthless, like the turret access number setting, and I'll take it out at the same time. I know you like this particular feature as you were the one who added it back in #1835. But still, that's just not enough to justify keeping it. I did actually explore the airlock electronics thing, before ultimately deciding that the feature itself just isn't worth keeping. Here are some of my reasons in greater detail:
- In order to use that feature at all, you must build a custom turret, and you must build it in an area with no turret control (ie: not the upload, ai minisat, or basically any of the areas that need turrets).
- Nobody builds turrets since you typically need an egun or equivalent to build them, and sec just won't hand those out.
- Even if you want to build one and get the parts, all the areas that really need turrets (upload, minisat, etc) aren't viable for this as this feature only applies to un-turretcontrolled turrets outside of those areas.
- So you'd have to put it in an area like armory, engineering, or science. Many of those areas have glass windows everywhere, which means your turret that you set to say, armory access in the armory is going to constantly be shooting at sec officers in the main hallway who lack armory access. Your turret set to CE access in the CE office will shoot at regular engineers in the eng hallway, etc. 
- Possibly the most useful case where you might benefit from building a turret by itself in an unusual area is defense against xenos/terrors/etc, but they're not very good for that.
- More likely building random turrets outside of the normal areas, then changing their access requirements away from the standard.... say making them shoot anyone who isn't a member of your department.... would be useless for antags (AI would just turn the turret off) and self-antagging for anyone else.
I did seriously consider carrying your feature over to TGUI... but in the end the complete lack of anyone using it, and the various issues with what it would take to actually use in practice, along with the not-so-great consequences of actually using it... made me decide to can the feature and not port it.

Markolie:
I regret posting here, because naturally it resulted in the usual massive paragraphs and will inevitably result in the endless discussions that always happens here. Which is also one of the reasons that I (among others) don't bother being active on here anymore. I also know you don't really ever change your mind no matter how much arguing happens, so I won't bother.

All I'll say is that your argument boils down to that **you** don't consider it useful and not commonly used. As such, you don't want to spend the time to add it in. In the end, I think if you're going to convert an interface like this you should go all the way, instead of just doing a partial port because **you** feel like it isn't worth the effort. I'd rather have a UI with more features than one that just works a little better. The Cyborg feature would take like five minutes to add to the current one. 

"Trading" this feature for the Cyborg feature is not really a relevant argument, as they're two completely different features.

Either way, like I said, I'll unsubscribe from this post now and move on.

procdrone:
@Kyep 
While AI will die, turrets would still offer PLENTY of deterrence from the blob growth and delay it. AI dies, okay. but there are still 4 turrets in the chamber that are operational even with AI dead.

Now, lets say the blob took out all the turrets. now it either has to advance towards the hallway which is also turret heavy, or go grow into space.

Not to mention, if the AI wasn't fast enough to warn the crew, there would be a lot of audio warning for anyone in the vicinity to note all the lasers shooting.

This is especially important for blob because blob snowball hard and at some point they are simply unbeatable.  and If AI isn't robust enough to notice the sounds in AI sat (and lets face it most of the time they dont), there is no warning for the crew, and no possibility to notice the blob for the crew. (how often you walk inside the AI sat?)

this is just meta in the bad sense of it, and turrets should also protect against that. or, the blob rushing towards AI that has no means of defense. 

There shouldn't be a 100% safe place for blobs to spawn in. This ruins the balance and makes it boring and repetitive. (same locations all the time. Electric Maint, AI sat, Turbine, Secure Storage).

_another solution (to this issue and many others) is just a bigger map but we all know it aint happening_

TheMaveran:
So this removes the ability to make laser tag turrets? That is not epic at all, what is the point of removing this?

AffectedArc07:
```
14.33s$ tools/travis/build_tgui.sh
Now using node v12.18.3 (npm v6.14.6)
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@xtuc/long/-/long-4.2.2.tgz: Request failed \"404 Not Found\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/travis/build/ParadiseSS13/Paradise/tgui/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command "tools/travis/build_tgui.sh" exited with 1.
```
Re-run travis. This error is not your fault/

Kyep:
@procdrone :
Giving turrets the ability to target blob tiles/cores/etc would be a significant feature / balance thing. I'm not putting that in this PR.

@TheMaveran 
- No, this PR *doesn't* remove the ability to make laser tag turrets. In fact, laser tag turrets on the live server are HORRIBLY broken in at least 3 different ways (spawning a real one creates a runtime error, and attempting to construct one results in a normal turret with a lasertag gun, which won't actually look or behave correctly at all). Lasertag turrets were already broken long before this PR.
- However, this PR *does* fix all 4 of those issues with lasertag turrets. So, with this PR, you can actually build laser tag turrets that really do work. And they have proper interfaces using TGUI. (just the on/off, since none of the other options make sense for them). They even use the correct sprites now.

@AffectedArc07 Changed. Not updating the screenshot for this one since.... its so minor.

AffectedArc07:
image
All NPM CI is broken

tgui/packages/tgui/interfaces/PortableTurret.js
```diff
@@ -0,0 +1,103 @@
...
+              <Button.Checkbox
+                fluid
+                checked={neutralize_criminals}
+                content="Wanted Criminals"
+                disabled={locked}
+                onClick={() => act('autharrest')} />
+              <Button.Checkbox
+                fluid
+                checked={neutralize_norecord}
+                content="No Sec Record"
```
AffectedArc07, line 62: Ayy

AffectedArc07:
NPM seems to be fixed now, so next commit should pass CI

Kyep:
Changes:
- No longer removes the "change access requirements of turrets outside of a turret controller" feature. I'm still skeptical as to the utility of this... but its now supported. 
- Now has additional refactoring.
- Now ports AccessList.js, a TGUI component for handling access list settings. (with some tweaks to make it more readable)

PacifistDalek:
Very excited that this fixes lasertag turret bugs

code/game/machinery/portable_tag_turret.dm
```diff
@@ -18,43 +26,15 @@
...
-	data["enabled"] = enabled
-	data["is_lethal"] = 0
+	return
+
+/obj/machinery/porta_turret/tag/tgui_data(mob/user)
+	var/list/data = list(
+		"locked" = isLocked(user), // does the current user have access?
+		"on" = enabled, // is turret turned on?
+		"lethal" = FALSE,
+		lethal_is_configurable = lethal_is_configurable
```
AffectedArc07, line 36: Marked this as a "come back to"

PR #14014, opened 2020-08-05
Hides the "Set transfer amount" verb on certain items like syringes

### LINKED ISSUE #12888: Syringes' "Set Transfer Amount" popup window doesn't appear.
**Problem Description**:
Syringes have got a rightclick menu item for changing the transfer amount. When you click on it, nothing happens.

I expect a popup window to appear where I can select some specific amount of chemical that I want to transfer each time - you can do this with a Dropper and set the amount from 1-5, for example.

**What happened instead**:
Nothing. I clicked on the item, the rightclick menu disappeared, and the game didn't give me any new prompt.

Lack of control when mixing precise amounts of chemicals. For myself I know it is quite inefficient for xenobio where chemicals must be injected into slime cores to produce a particular reaction; only one unit is required and it would be nice to only actually inject that much. Bizarre medical experiments probably want this even more badly because at least I have plasma on-tap.

Pick up a syringe. Rightclick it in your hand. Click the "set transfer amount" menu item.

I didn't know about the intended functionality until today, so I don't know when it last worked. A mentor telling me how to do xenobio was unaware his instructions would not work.

It was a pretty calm round. I was a regular scientist in xenobiology (and a touch of RnD and repairing some station damage), there were changelings. I tested again on the subsequent round (syndicate war declaration) with the same behaviour.

SteelSlayer:
## What Does This PR Do
Some objects, like syringes, have the "Set transfer amount" right click verb, but nothing is meant to happen when you click on it. This PR hides the verb on those objects, so there is no confusion as to why it's not working.

## Why It's Good For The Game
Fixes #12888

## Changelog
:cl:
tweak: Hides the "Set transfer amount" right click verb on syringes, auto-menders, honeycombs, and all food items
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/food.dm
+++ b/code/modules/food_and_drinks/food.dm
@@ -33,6 +33,9 @@
 		deltimer(ant_timer)
 	return ..()
 
+/obj/item/reagent_containers/food/set_APTFT()
+	set hidden = TRUE
+
 /obj/item/reagent_containers/food/proc/check_for_ants()
 	if(!antable)
 		return
--- a/code/modules/hydroponics/beekeeping/honeycomb.dm
+++ b/code/modules/hydroponics/beekeeping/honeycomb.dm
@@ -17,6 +17,8 @@
 	pixel_y = rand(8,-8)
 	update_icon()
 
+/obj/item/reagent_containers/honeycomb/set_APTFT()
+	set hidden = TRUE
 
 /obj/item/reagent_containers/honeycomb/update_icon()
 	overlays.Cut()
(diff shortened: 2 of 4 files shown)


code/modules/food_and_drinks/food.dm
```diff
@@ -33,6 +33,9 @@
 		deltimer(ant_timer)
 	return ..()
 
+/obj/item/reagent_containers/food/set_APTFT()
+	set hidden = TRUE
```
AffectedArc07, line 37: Who named this proc

PR #14016, opened 2020-08-06
Fixed Gulag Wire Spaghetti

SabreML:
## What Does This PR Do

Repositions the items in the Labour Camp Maintenance a bit to make the wiring look better.

## Why It's Good For The Game

Current setup could cause eye damage.

## Images of changes

Old:
old
New:
new

## Changelog
:cl:
tweak: Moved a couple of things at the Labour Camp to reduce Cable Spaghetti.
/:cl:

~~very important PR, please speedmerge~~

### DIFF
--- a/_maps/map_files/generic/Lavaland.dmm
+++ b/_maps/map_files/generic/Lavaland.dmm
@@ -681,7 +681,7 @@
 	req_access_txt = "2"
 	},
 /obj/structure/cable{
-	icon_state = "1-8"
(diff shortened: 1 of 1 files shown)


Cocacolagua:
it's best to avoid wires below walls if possible, which in this case it is

SabreML:
> it's best to avoid wires below walls if possible, which in this case it is

Good point, one second.

PR #14017, opened 2020-08-06
Changes the projectile path to not be an item

farie82:
## What Does This PR Do
Makes it so projectiles are not items.
Projectiles and items only shared 2 variables that mattered. `hitsound` and `armour_penetration`. Other logic and values on item were not required for projectiles.

Ports https://github.com/tgstation/tgstation/pull/46692

## Why It's Good For The Game
Should save a bit of memory for each projectile. Also makes the paths a bit more sane.

I've tested this using multiple guns and magic and didn't see any issues.

I did use a mass replace for this. I'll comment on the code that I changed manually.
Replaced:
`obj/item/projectile`
With:
`obj/projectile`
In all files. (only found in the DM files)

## Changelog
Pure backend

### DIFF
--- a/code/modules/ninja/suit/n_suit_verbs/ninja_stars.dm
+++ b/code/modules/ninja/suit/n_suit_verbs/ninja_stars.dm
@@ -6,7 +6,7 @@
 	icon = 'icons/obj/ninjaobjects.dmi'
 	fire_sound = 'sound/weapons/genhit.ogg'
 	cell_type = "/obj/item/stock_parts/cell/infinite"
-	projectile_type = "/obj/item/projectile/energy/shuriken"
+	projectile_type = /obj/projectile/energy/shuriken
 	charge_cost = 0
 	silenced = 1
 
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -1,4 +1,4 @@
-/obj/item/projectile
+/obj/projectile
 	name = "projectile"
 	icon = 'icons/obj/projectiles.dmi'
 	icon_state = "bullet"
@@ -8,7 +8,8 @@
 	flags = ABSTRACT
 	pass_flags = PASSTABLE
 	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
-	hitsound = 'sound/weapons/pierce.ogg'
+	//The sound this plays on impact.
+	var/hitsound = 'sound/weapons/pierce.ogg'
 	var/hitsound_wall = ""
 	var/def_zone = ""	//Aiming at
 	var/mob/firer = null//Who shot it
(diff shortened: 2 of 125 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -8,7 +8,8 @@
 	flags = ABSTRACT
 	pass_flags = PASSTABLE
 	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
-	hitsound = 'sound/weapons/pierce.ogg'
+	//The sound this plays on impact.
+	var/hitsound = 'sound/weapons/pierce.ogg'
```
farie82, line 12: Here. Made this a var on the projectile object

code/modules/projectiles/projectile.dm
```diff
@@ -36,7 +37,9 @@
 	var/damage_type = BRUTE //BRUTE, BURN, TOX, OXY, CLONE are the only things that should be in here
 	var/nodamage = FALSE //Determines if the projectile will skip any damage inflictions
 	var/flag = "bullet" //Defines what armor to use when it hits things.  Must be set to bullet, laser, energy,or bomb	//Cael - bio and rad are also valid
-	var/projectile_type = "/obj/item/projectile"
+	///How much armor this projectile pierces.
+	var/armour_penetration = 0
```
farie82, line 41: Here. Made this a var on the projectile object

AffectedArc07:
I would say TM this at minimum. I dont like PRs which are mass find+replace automatically across large amounts of files. Also considering projectiles are in existence for very short periods of time, I dont see this as too impactful.

taukausanake:
What about Carp? Is this going to mess with anyone blocking bullets? That's the first thing I would test

code/modules/ninja/suit/n_suit_verbs/ninja_stars.dm
```diff
@@ -6,7 +6,7 @@
 	icon = 'icons/obj/ninjaobjects.dmi'
 	fire_sound = 'sound/weapons/genhit.ogg'
 	cell_type = "/obj/item/stock_parts/cell/infinite"
-	projectile_type = "/obj/item/projectile/energy/shuriken"
+	projectile_type = /obj/projectile/energy/shuriken
```
farie82, line 9: Removed the "'s as well here to make it the same as with other guns

farie82:
> What about Carp? Is this going to mess with anyone blocking bullets? That's the first thing I would test

image
as expected. It's just a path change. Any references that were there are changed.

> I would say TM this at minimum. I dont like PRs which are mass find+replace automatically across large amounts of files. Also considering projectiles are in existence for very short periods of time, I dont see this as too impactful.

Manually doing it would have led to the same result. Maybe even missing paths.
I've gone over the list of changes though to see if something is odd about them. Found nothing myself. Same with the PR on TG. They did the same and also found that hitsound and the armor one are used which were from item.

I would very much prefer a TM though just to verify it fully

farie82:
> I don't like this. Lots of potential conflicts and bugs. All for what? Saving a small amount of memory on an item that exists for 2 seconds? When we're not exactly short on RAM anyway?

That is why test merges exist. The PR is also more about the refactor than the memory and processing saved.
A projectile being an item does not make much sense since it does not behave like one. You cannot grab it. Throw it etc. It being one is quite odd OO wise

variableundefined:
Objection, on the same ground as Kyet and Affected. I don't think this is worth the troubles.

variableundefined:
Closing. Objection - 1 head and 2 maints (affected and i)

PR #14018, opened 2020-08-06
Hierophant staff nerf

Nazsgull:
## What Does This PR Do
Nerfs Hierophant Club so it cannot use ranged attacks or blasts on station (it only spawns a few sparks)

## Why It's Good For The Game
Helps miners not getting weapons of mass destruction.

## Changelog
:cl:
tweak: Hierophant club will now malfunction if used onboard station
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/hierophant_loot.dm
+++ b/code/modules/mining/lavaland/loot/hierophant_loot.dm
@@ -52,9 +52,9 @@
 	..()
 	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
 
-		if(world.time > timerMalfunction)
-			timerMalfunction = world.time + cooldown_time
-			user.visible_message("<span class='danger'>[user]'s lava staff malfunctions!</span>")
+		if(world.time > timer)
+			timer = world.time + cooldown_time
+			user.visible_message("<span class='danger'>[user]'s hierophant club malfunctions!</span>")
 			do_sparks(5, FALSE, user)
 		return
 
--- a/code/modules/mining/lavaland/loot/hierophant_loot.dm
+++ b/code/modules/mining/lavaland/loot/hierophant_loot.dm
@@ -21,6 +21,7 @@
 	var/chaser_timer = 0 //what our current chaser cooldown is
 	var/chaser_speed = 0.8 //how fast our chasers are
 	var/timer = 0 //what our current cooldown is
+	var/timerMalfunction = 0 //timer for when the club malfunctions
 	var/blast_range = 13 //how long the cardinal blast's walls are
 	var/obj/effect/hierophant/beacon //the associated beacon we teleport to
 	var/teleporting = FALSE //if we ARE teleporting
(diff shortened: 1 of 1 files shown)


hal9000PR:
would its teleport function still work on the station?

Nazsgull:
> work

No, it does not affect that, only the ability to attack with the staff. Beacon can be placed and works normally.

code/modules/mining/lavaland/loot/hierophant_loot.dm
```diff
@@ -21,6 +21,7 @@
 	var/chaser_timer = 0 //what our current chaser cooldown is
 	var/chaser_speed = 0.8 //how fast our chasers are
 	var/timer = 0 //what our current cooldown is
+	var/timerMalfunction = 0 //cooldown for when the club malfunctions
```
farie82, line 24: Unused var

code/modules/mining/lavaland/loot/hierophant_loot.dm
```diff
@@ -49,6 +50,14 @@
 
 /obj/item/hierophant_club/afterattack(atom/target, mob/user, proximity_flag, click_parameters)
 	..()
+	if(!is_mining_level(user.z)) //Will only spawn a few sparks if not on mining z level
+
+		if(world.time > timer)
```
farie82, line 55: Instead of checking the timer twice in this proc it's best to move line 62 to the front. The timer check only that is. And remove this one.
so something like this:
```
if(!(timer check))
 return

if(!mining)
 your stuff
 return

var/T = turf
if(!T)
 return
other stuff
```

Balthusdire:
Perfect change. Leaving the teleport intact is good since it is frequently used as a quick way to get from the station back to mining (with the added bonus of skipping the mining shuttle console screens which are super prone to whitescreen for me at least) but not having a burning super weapon on station is good.

hal9000PR:
i think this is a good time to remove the heiro clubs ability to teleport megafauna onto the station

code/modules/mining/lavaland/loot/hierophant_loot.dm
Nazsgull: It makes sense, but I'm not confident enough with this syntax to do that modification and assure the rest will work the same way...

farie82: Can always ask somebody in coding chat to help you. Or check it before you commit it.
Enough people there willing to help!

Nazsgull: I think I solved it!

code/modules/mining/lavaland/loot/hierophant_loot.dm
```diff
@@ -49,42 +50,50 @@
...
-		if(is_station_level(T.z))
-			add_attack_logs(user, target, "Fired 3x3 blast at [src]")
-		else
-			add_attack_logs(user, target, "Fired 3x3 blast at [src]", ATKLOG_ALL)
-	else
-		if(ismineralturf(target) && get_dist(user, target) < 6) //target is minerals, we can hit it(even if we can't see it)
-			INVOKE_ASYNC(src, .proc/cardinal_blasts, T, user)
-			timer = world.time + cooldown_time
-		else if(target in view(5, get_turf(user))) //if the target is in view, hit it
+	if(world.time > timer)
```
farie82, line 53: An early return is preferred in this codebase.
Early returns are like this:
```
if(!can_continue)
  return
if(!other_check)
  return
do_stuff
```

This way the code stays on rougly the same indentation which improves readability

Kyep:
FWIW, even TG is nerfing this item because of how OP it is: https://github.com/tgstation/tgstation/pull/52880

hal9000PR:
That blink feature is interesting. I always like more mobility items

code/modules/mining/lavaland/loot/hierophant_loot.dm
Nazsgull: Done

AffectedArc07 (approved):

PR #14022, opened 2020-08-07
[TGUI] Tanks

Kyep:
## What Does This PR Do
Converts tanks (both the small ones that come in your internals box, and the big ones you get from tank dispensers) to TGUI.

## Why It's Good For The Game
One less thing using nanoui.

## Images of changes
Default state for an internals tank:
tank_default

As previously, but now you're wearing a mask, and have filled the tank up to capacity using a canister:
tank_internals_off

And finally, with a fully operational internals tank:
tank_internals_on

## Changelog
:cl: Kyep
refactor: Ported tanks to TGUI.
/:cl:


code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -142,70 +142,49 @@
...
-			var/cp = text2num(href_list["dist_p"])
-			distribute_pressure += cp
-		distribute_pressure = min(max(round(distribute_pressure), 0), TANK_MAX_RELEASE_PRESSURE)
-
-	if(href_list["stat"])
-		toggle_internals(usr)
-
+		return
+	. = TRUE
+	switch(action)
```
dearmochi, line 173: Lacks a default case in case the action isn't a valid one. I usually `return FALSE` to prevent fingerprints from being added

code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -142,70 +142,49 @@
...
+		if("pressure")
+			var/pressure = params["pressure"]
```
dearmochi, line 175: Unlikely to happen, but this may result in a runtime if `params["pressure"]` is some random text. This could be fixed by adding a default case in the below `if` and returning or something.

code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -142,70 +142,49 @@
...
-				if(H.head && (H.head.flags & AIRTIGHT))
-					data["maskConnected"] = 1
-
+	var/mob/living/carbon/C = user
+	if(!istype(C))
+		C = loc.loc
+	if(!istype(C))
+		return data
+	data["has_mask"] = C.wear_mask ? TRUE : FALSE
+	data["connected"] = (C.internal && C.internal == src) ? TRUE : FALSE
```
dearmochi, line 166: Maybe just `data["connected"] = (C.internal == src) ? TRUE : FALSE` since `src` can't really be `null`

tgui/packages/tgui/interfaces/Tank.js
```diff
@@ -0,0 +1,81 @@
+import { useBackend } from '../backend';
+import { Button, LabeledList, NumberInput, ProgressBar, Section, Tooltip } from '../components';
+import { Window } from '../layouts';
+
+export const Tank = (props, context) => {
+  const { act, data } = useBackend(context);
```
dearmochi, line 6: Not a blocker, but I still maintain that making the variables in `data` local so you don't have to prefix them by `data.` when using them is better in my opinion.

code/game/objects/items/weapons/tanks/tanks.dm
Kyep: A maint has told me this way is more readable.

tgui/packages/tgui/interfaces/Tank.js
Kyep: I'm not willing to remove data. from these TGUI PRs, because I'm trying to maintain as much consistency with TG's version as possible, and renaming all the vars is not gonna help with that.

code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -142,70 +142,55 @@
...
-			else if(ishuman(C))
-				var/mob/living/carbon/human/H = C
-				if(H.head && (H.head.flags & AIRTIGHT))
-					data["maskConnected"] = 1
-
+	var/mob/living/carbon/C = user
+	if(!istype(C))
+		C = loc.loc
+	if(!istype(C))
+		return data
```
SteelSlayer, line 164: Now I know the old code was doing some weird stuff like this too, but I'm not sure if this is needed. From what I can see, the tank UI is only opened in `attack_hand()` which means that the user needs to be a carbon (and have hands) to open the UI.

I don't see any situation where the user isn't a carbon but maybe I'm overlooking something.

Kyep: Both Paradise and TG have safety checks like this. And there are some cases where people can open TGUIs they normally shouldn't be able to, like aghosts clicking on stuff. I'd rather err on the side of caution and have a safety check in there.

SteelSlayer: Alright then

SteelSlayer:
Personal preference, but I think that each line of the UI would look better if they were the same length. So maybe the progress bar could be shrunk a bit to be in line with the "reset" button below it, and the internals on/off button could be stretched out a bit to also be in line.

Kyep:
@SteelSlayer Perhaps, but:

- As far as I can tell, ProgressBar has no width attribute
- Even if I could scale it, the third line has to account for a few varied states, so its not gonna line up regardless
So, probably not worth trying to make it all line up.

PR #14023, opened 2020-08-07
Make wet overlays render on MID_TURF_LAYER plane

### LINKED ISSUE #14007: Wet Floor Graphic Doesn't Appear On Space Floor Tiles
**Issue Description**:
I was playing recently when someone went around replacing all the floor tiles in the main halls with the tiles that make it look as if there isn't a floor there at all, their intentions were harmless but someone who loves being annoying with lube decided to lube various parts of the hall, the floors however didn't render the shiny wet floor graphic they usually do effectively creating invisible lube

To see a shiny/wet floor graphic where the floor was wet

**What happened instead**:
Lube/water was totally invisible and you couldn't tell where it was if it was there at all

It could be exploited and also seems to just be a bug because the floor tiles have a rather unique function

1. Replace floor tiles with space floor tiles
2. Wet floor with lube or water
3. Notice that the lube/water is invisible and suffer as you can't tell where to walk or crowbar

Unsure, probably when the tiles were added

**Extra information**:

dearmochi:
## What Does This PR Do

Fixes #14007 by setting `wet_overlay` plane to `MID_TURF_LAYER`.

## Why It's Good For The Game

Makes it possible to tell whether astral carpet is slippery while keeping the parallax effect.

## Images of changes

dreamseeker_Vry2DfTlyM

## Changelog
:cl:
fix: Fix space floors not looking slippery despite being so
/:cl:

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -53,8 +53,7 @@
 				wet_overlay = image('icons/effects/water.dmi', src, "ice_floor")
 			else
 				wet_overlay = image('icons/effects/water.dmi', src, "wet_static")
-		if(plane == PLANE_SPACE)
-			wet_overlay.plane = PLANE_SPACE_PARALLAX + 0.1
+		wet_overlay.plane = MID_TURF_LAYER
 		overlays += wet_overlay
 	if(time == INFINITY)
 		return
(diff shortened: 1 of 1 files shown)


code/game/turfs/simulated.dm
```diff
@@ -53,6 +53,8 @@
 				wet_overlay = image('icons/effects/water.dmi', src, "ice_floor")
 			else
 				wet_overlay = image('icons/effects/water.dmi', src, "wet_static")
+		if(plane == PLANE_SPACE)
+			wet_overlay.plane = PLANE_SPACE_PARALLAX + 0.1
```
AffectedArc07, line 57:
```suggestion
		wet_overlay.plane = MID_TURF_LAYER
```

This makes it a lot less hacky, and forces it on all turfs just fine
image

dearmochi: done

PR #12241, opened 2019-09-09
Admin verb mega update

SteelSlayer:
**What does this PR do:**
As per @Kyep's changes wanted page on discord, and from what he personally told me to update, I've tweaked/removed/updated quite a few admin verbs. Now obviously I took some liberty in how I changed *some* of these verbs. So if any of the staff want me to do something I did here differently, just tell me.

Preview of the new combined "List SSDs and AFKs" verb. Two seperate lists, one window
ssdafk

Preview of where the "resolve all ticket" buttons are now located:
tickets
This button will be visible to mentors in their mentor ticket window but since it calls the admin verb, they won't have the permissions to actually use it and will get a message stating so. The "resolve all" button is also present in the admin ticket window.

**Changelog:**
:cl:
tweak: The "Show Player Panel" verb is now right click only
tweak: The "C.M.A. - Admin" verb is now right click only
tweak: The "C.M.A. - Self" verb is now right click only
tweak: The "Check Contents" verb is now right click only
tweak: The "Freeze" verb is now right click only
tweak: The "Freeze Mech" verb is now right click only
tweak: The "Jump to Turf" verb is now right click only
tweak: The "Update Mob Sprite" verb is now right click only
tweak: The "List SSDs" and "List AFKs" verbs have been combined into one verb named "List SSDs and AFKs"
tweak: The "Resolve All Open Mentor Tickets" verb has been moved inside the mentor ticket interface window
tweak: The "Resolve All Open Admin Tickets" verb has been moved inside the admin ticket interface window
tweak: Increases the width of the new "List SSDs and AFKs" window a bit
/:cl:

PR #14024, opened 2020-08-07
Admin Verb Updates

Kyep:
## What Does This PR Do
Verbs that used to require only R_ADMIN, but due to being for events, now require R_EVENT:
- SubtleMessage
- HeadsetMessage
- CMA admin/self

Other admin verb permission changes:
- toggleenter/toggleguests/toggledrones now require R_SERVER, since they affect the whole server
- toggle VPN whitelist now requires R_BAN, since its used to determine if an individual can join or not
- Map template load/upload/place now require R_DEBUG (for consistency with other debug verbs, and because they potentially break the server)

Removed verbs:
- "alt check", which displays every player (like player panel) including players with no alts. We don't even care if players have alts. If someone is banned on an alt, then we care, but we already have systems to auto-detect that. If we want to check a specific player for alts, we have player panel for that. This verb has no reason to exist.
- "Del-All", which provides a nasty list to pick through, lags your client, and is better done by SDQL faster and easier anyway.
- "Toggle Aliens", which enables/disables the xeno event, which we already have the event panel to do - we don't need an entire verb just for toggling one event.
- "Unleash Floor Cluwne", which is pointless, as spawning one in general can be done via game panel like everything else (and is bugged and doesn't work anyway). Spawning one to target one player is better done via smite. We don't need a whole verb just for spawning one specific mob! This sort of verb is exactly how we got to the current ridiculous levels of admin verb bloat.
- "Display Job Bans" and "Unban Panel", which are now both combined into one verb: "Ban Panel" (since its not specific to job bans). The only thing this gets rid of is file-based jobban support. We've used the DB-based ban system forever. All file-based bans are depreciated at this point.
- "Player Panel", because "Player Panel New" is the new "Player Panel". It does everything the old "Player Panel" does, but better, searchable, etc. Realistically "Player Panel New" has been "New" for how many years now? It is time to put the old "Player Panel" out to pasture accept its forever-waiting-in-the-wings replacement.

Made these admin-verbs right-click/chat-bar use only:
- "Tweak Component Ratings"
- "Get Mob" and "Get Key"
- "Atom ProcCall"
- "Delete"
- "Rejuvenate"
- "Direct Narrate"
- "Man Up"
IE: you can still use them via right-click or entering them in the chat bar, but they no longer show up in the "admin" tab's verb list. Instead, you're meant to use them by right-clicking a suitable target (like a mob). This reduces the bloated verb list, reduces the chance of selecting the wrong person, and is much faster in many cases (no more scrolling through a HUGE list to pick the right target). In some cases (like "delete") this target list was so big it could freeze your client. You can also still use them via entering them in the chat bar, if you really want.

Some other changes:
- Admins without R_PERMISSIONS no longer see a non-functional rank/permission edit link in every player panel that generates a nasty warning to all admins if you touch it
- "Show Custom Tip" now correctly checks for R_EVENT instead of R_ADMIN, and shows up in the Events tab instead of the Admin tab for consistency with the other verbs that require R_EVENT. This also fixes the fact that some admins without R_EVENT have a whole Event tab for like... 1-2 verbs.
- "Reload Admins" now shows up under Server category, and isn't incorrectly added to the verb list of admins with R_DEBUG but not R_SERVER
- Fixed a bunch of basic "if(!holder) continue" checks by replacing them with a proper check_rights check.
- Fixed possession verbs not checking R_POSSESS, which is the point of the permission flag existing in the first place
- Fixed "show debug verbs" letting you access "Trigger Event" without R_EVENT that is meant to be required
- Deletes commented-out code in admin files

## Why This PR Is Good For The Game
- Reduces admin verb bloat
- Refactoring
- Consistency
- TAs not using event verbs or stuff that can mess up the server

## Changelog
:cl:
tweak: simplified the admin verb list.
/:cl:


AffectedArc07:
Please also update `config/example/admin_ranks.txt` to not only include better descriptions, but also fix the fact that `+PROCCALL` is not mentioned in there, nor is `+MOD` (even though we dont use it, it should be documented), and `+FUN` is when we dont have that permission.

code/modules/admin/topic.dm
```diff
@@ -2379,7 +2379,7 @@
 		usr.client.cmd_admin_direct_narrate(M)
 
 	else if(href_list["subtlemessage"])
-		if(!check_rights(R_ADMIN))	return
+		if(!check_rights(R_EVENT))	return
```
SteelSlayer, line 2382:
```suggestion
		if(!check_rights(R_EVENT))	
		    return
```

AffectedArc07:
```
loading paradise.mdme
code/modules/admin/topic.dm:2383:error: inconsistent indentation
paradise.mdme.dmb - 1 error, 0 warnings (8/7/20 2:49 pm)
```

Likely tab/space merge

Kyep:
@AffectedArc07 Fixed

TheSardele:
https://github.com/ParadiseSS13/Paradise/pull/12241#issuecomment-529323662

Last time someone wanted to remove the player panel, Fox wanted to keep it in as the new one had broken before, but the old one still worked. May want to consider keeping it in.

AffectedArc07:
> Map template load/upload/place (due to it potentially breaking the game)

I disagree with this being gated behind `R_EVENT` as I may need to urgently debug map templates at some point live, and this would lock me out of said debugging.

Kyep:
@AffectedArc07 The two map template verbs have been changed to require R_DEBUG.

ZomgPonies:
Yeah this is a long overdue cleanup. Everything in the list is logical and needed. 10/10

Still think a lot of the 'Secret' menu stuff should be moved to the event tab, but eh.

PR #14033, opened 2020-08-07
Adds Earmuffs providing immunity to Changeling shriek

warior4356:
## What Does This PR Do
For consistency, this makes it so earmuffs protect against changeling shriek just like they do the vampire shriek.

EDIT: PR refactored sonic protection to be a tiered system where earmuffs and deafness provide immunity to all and bowman headsets provide immunity to only bang.

## Why It's Good For The Game
Earmuffs are supposed to provide immunity to all sonic attacks, it is inconsistent that they don't protect against this one.

EDIT: The refactoring will make it easier to expand on sonic damage and sonic protection.

## Changelog
:cl:
add: Earmuffs now provide immunity to changeling shriek
add: Deafness now provides immunity to all sonic attacks
tweak: Refactored sonic protection into a tiered system
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -264,7 +264,7 @@
 					limbs_affected -= 1
 				else valid_limbs -= processing_dismember
 
-			if(!istype(l_ear, /obj/item/clothing/ears/earmuffs) && !istype(r_ear, /obj/item/clothing/ears/earmuffs))
+			if(src.check_ear_prot() < HEARING_PROTECTION_TOTAL)
 				AdjustEarDamage(30, 120)
 			if(prob(70) && !shielded)
 				Paralyse(10)
@@ -288,7 +288,7 @@
 						limbs_affected -= 1
 					else valid_limbs -= processing_dismember
 
-			if(!istype(l_ear, /obj/item/clothing/ears/earmuffs) && !istype(r_ear, /obj/item/clothing/ears/earmuffs))
+			if(src.check_ear_prot() < HEARING_PROTECTION_TOTAL)
 				AdjustEarDamage(15, 60)
 			if(prob(50) && !shielded)
 				Paralyse(10)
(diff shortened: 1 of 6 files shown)


TheMaveran:
While this is for consistency purposes - I do wonder, why do these provide protection from the shrieks whilst giving no downsides? Seems like a balance oversight, especially given that the vampire version is extremely costly to use.

Spacemanspark:
They do have downsides. You can't wear any headsets with them equipped and it has a chance for you to not hear what people around you are saying.

It's a very situational item.

TheSardele:
I don't remember the effects spark describe being a thing, I don't think they actually have any downsides. Quick search of the code for the earmuffs didn't turn up anything.

Denthamos:
@TheSardele I do remember the not hearing anything being a thing, but the not wearing headsets I do not.  I wonder if that downside of the earmuffs was lost in the fix to them which made them protect against vampire screech again.

MRgaty:
I normally wear earsmuffs as sec officer because they really have no downsides, have never experienced deafness or any kind of downsides and I could wear my headsets pretty well.

Qwertytoforty:
I thought the only catch earmuffs had was you couldn't wear two headsets.

TheMaveran:
Yeah - As earmuffs currently are, they have literally no downside. They just take one ear slot and as far as I've seen, do not prevent your ability hear.

Dumbdumn5:
If those changes do what I think they did @warior4356 then you got yourself a :+1: , would like some confirmation on that refactoring and deafness bit though is all since I can't read wizard speak.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -941,12 +941,18 @@
 	return number
 
 /mob/living/carbon/human/check_ear_prot()
+	if(!src.can_hear())
```
farie82, line 944: src is implied. Can remove `src.`

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -941,12 +941,18 @@
...
+		return HEARING_PROTECTION_TOTAL
+	if(l_ear && istype(l_ear, /obj/item/clothing/ears/earmuffs))
```
farie82, line 946: Can simplify this one and the one on line 948 to just an istype. `istype(null)` will return FALSE

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -264,7 +264,7 @@
 					limbs_affected -= 1
 				else valid_limbs -= processing_dismember
 
-			if(!istype(l_ear, /obj/item/clothing/ears/earmuffs) && !istype(r_ear, /obj/item/clothing/ears/earmuffs))
+			if(src.check_ear_prot() < HEARING_PROTECTION_TOTAL)
```
farie82, line 267: `src.` can be omitted here

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -288,7 +288,7 @@
 						limbs_affected -= 1
 					else valid_limbs -= processing_dismember
 
-			if(!istype(l_ear, /obj/item/clothing/ears/earmuffs) && !istype(r_ear, /obj/item/clothing/ears/earmuffs))
+			if(src.check_ear_prot() < HEARING_PROTECTION_TOTAL)
```
farie82, line 291: Same

farie82 (changes requested):
You're also missing quite some other tweaks due to the changes.
Deaf people now won't be affected by quite some things. Like the honk mech and explosions (for ear damage) etc.

variableundefined (changes requested):
Approved design-wise. Please address Farie's review above.

warior4356:
Requested changes have been made, changelog has been updated, and more information was added to the PR.

farie82 (approved):

Kyep (approved):

PR #14034, opened 2020-08-07
Improves botany fruit logging. Makes throws not spam admins when the item does no damage

farie82:
## What Does This PR Do
- Makes fruits and such log what they exactly did when thrown at somebody
- Makes thrown items that do no damage send a message to admins with `ATKLOG_ALMOSTALL` on

## Why It's Good For The Game
- Gives the log more info about what the fruit actually did. What it contained and what it's effects were
- Less spammy for admins who see gloves being thrown at people.

## Images of changes
image
When the admin is on ATKLOG_MOST
image

## Changelog
:cl:
add: Added extra logging for botany fruits.
tweak: Thrown objects which do no damage now only send a message to admins with `ATKLOG_ALMOSTALL` on.
/:cl:

### DIFF
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -117,6 +117,7 @@
 /obj/item/reagent_containers/food/snacks/grown/throw_impact(atom/hit_atom)
 	if(!..()) //was it caught by a mob?
 		if(seed)
+			log_action(thrownby, hit_atom, "Thrown [src] at")
 			for(var/datum/plant_gene/trait/T in seed.genes)
 				T.on_throw_impact(src, hit_atom)
 			if(seed.get_gene(/datum/plant_gene/trait/squash))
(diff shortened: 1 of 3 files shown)


code/modules/hydroponics/grown.dm
```diff
@@ -190,3 +191,17 @@
...
+	var/reagent_str = reagents.log_list()
+	var/genes_str
+	if(seed && length(seed.genes))
+		var/list/plant_gene_names = list()
+		for(var/datum/plant_gene/trait/T in seed.genes)
+			if(T.dangerous)
+				plant_gene_names += T.name
+		genes_str = english_list(plant_gene_names)
+	else
+		genes_str = "No genes"
```
SteelSlayer, line 204: Not sure how if this will make much of a difference, but if you assign a value to this var from the get go, you can avoid doing an `else` later on:
```js
var/genes_str = "No genes"
if(seed && length(seed.genes))
	var/list/plant_gene_names = list()
	for(var/datum/plant_gene/trait/T in seed.genes)
		if(T.dangerous)
			plant_gene_names += T.name
	genes_str = english_list(plant_gene_names)
```
Also this: `for(var/datum/plant_gene/trait/T in seed.genes)`, can probably be made typeless.

farie82: True that. It'll be a bit more readable.

The typed loop I forgot to take out yeah. Good catch!

PR #14036, opened 2020-08-08
Fixes offstation_role bug

### LINKED ISSUE #12764: On station rolls are getting flagged as off station roles.

Kyep:
## What Does This PR Do
Fixes #12764 - random crew members getting mind.offstation_role set to TRUE when they are actually normal crew.
The cause was the nuke ops can_start() proc assigning the flag in a way that persists even if the nuke ops game mode is not actually selected to be run.

## Why It's Good For The Game
Fixes a bug.
Ensures random people with nuke op role enabled don't incorrectly get a flag that makes them, for example, immune to being antag targets.
Allows #12765 to be reversed.

## Changelog
:cl: Kyep
fix: fixed random crew members being treated as offstation roles (non crew) when they weren't.
/:cl:

### DIFF
--- a/code/game/gamemodes/nuclear/nuclear.dm
+++ b/code/game/gamemodes/nuclear/nuclear.dm
@@ -53,7 +53,6 @@ proc/issyndicate(mob/living/M as mob)
 	for(var/datum/mind/synd_mind in syndicates)
 		synd_mind.assigned_role = SPECIAL_ROLE_NUKEOPS //So they aren't chosen for other jobs.
 		synd_mind.special_role = SPECIAL_ROLE_NUKEOPS
-		synd_mind.offstation_role = TRUE
 	return 1
 
 
@@ -113,7 +112,7 @@ proc/issyndicate(mob/living/M as mob)
 		if(spawnpos > synd_spawn.len)
 			spawnpos = 2
 		synd_mind.current.loc = synd_spawn[spawnpos]
-
+		synd_mind.offstation_role = TRUE
 		forge_syndicate_objectives(synd_mind)
 		create_syndicate(synd_mind)
 		greet_syndicate(synd_mind)


code/game/gamemodes/nuclear/nuclear.dm
```diff
@@ -53,7 +53,6 @@ proc/issyndicate(mob/living/M as mob)
 	for(var/datum/mind/synd_mind in syndicates)
 		synd_mind.assigned_role = SPECIAL_ROLE_NUKEOPS //So they aren't chosen for other jobs.
 		synd_mind.special_role = SPECIAL_ROLE_NUKEOPS
```
AffectedArc07, line 55: Wouldnt it make sense to also change these too?

Kyep: I don't *think* so.
mind.offstation_role isn't overwritten at roundstart, and thus persists into normal rounds, causing bugs. It has to be moved to fix them.
mind.assigned_role and mind.special_role DO get overwritten. So I don't think they need to be moved. I'm also concerned that moving them will potentially introduce other bugs.

Frankly, at some point we should probably rewrite all game_mode can_start procs to avoid setting ANYTHING that could potentially persist... but that's ridiculously out of scope for this PR. This PR is just about changing the thing which is definitely causing issues.

AffectedArc07: Tested locally and both of these are null at roundstart, but that was with a small sample size on a local server, so its unsure how it would react on live.

Kyep:
@farie82 Not removing that in this PR, for 2 reasons:

1. It can't cause this bug, as there is logic in game_modes which refuses to consider can_start() procs for disabled gamemodes.
2. There's likely to be another PR which completely nukes heist from the codebase, and changing that here would conflict with that PR.

farie82:
> @farie82 Not removing that in this PR, for 2 reasons:
> 
> 1. It can't cause this bug, as there is logic in game_modes which refuses to consider can_start() procs for disabled gamemodes.
> 2. There's likely to be another PR which completely nukes heist from the codebase, and changing that here would conflict with that PR.

It's the same bug happening in code that is deactivated config wise. Other downstream servers or any other server running para code using that gamemode will have the same issue.
I don't think there is a PR that removes the gamemode atm

AffectedArc07:
@farie82 The mode is extremely broken far beyond the issue of this thing, so any server that puts it into rotation is asking for stuff to go wrong. There were talks about axing this entire mode in the future and for the moment this PR is more important to get into the game than an entire gamemode removal (Which would also include config changes).

PR #14040, opened 2020-08-08
Fixes & Reworks Rogue Combat Drone Event

### LINKED ISSUE #14038: Combat drones don't work well at all
**Issue Description**:

Combat drones when in hostile_drone = 1 mode, chose themselves as a target, causing them to never attack anything in hostile mode. When they do successfully get a target, they will shoot at the target, but once they reach the target, they will endlessly nuzzle the target instead of doing a melee attack or point blank shooting them. Drones in hostile mode are also INCAPEABLE to attack people attacking them due to themselves being a target

For drones to agro people when hostile, and shoot them.

**What happened instead**:

They target themselves, or nuzzle whoever is pissing them off while they get crowbared to death

State of the art combat drones should be a threat

Spawn drones
Punch one with hostile drone = zero
watch them rush you and nuzzle you after shooting
Or
Attack one with hostile drone = 1
Watch as they do nothing.

Kyep:
## What Does This PR Do
Fixes #14038 - combat drones not working
Rewrites combat drone code for more interesting encounters & improved code quality.
Details:

1. Drones are quite dangerous now. They have a burst-fire ranged laser pellet weapon. When active, they attack anything near them except other drones. They also spawn in greater numbers. Hitting them when they're deactivated makes them activate.
2. Drones can be bypassed by waiting for them to enter their deactivated state, then slipping by before they reactivate.
3. You can also fight them for boards useful for R&D.
4. They're removed when the event ends.

## Why It's Good For The Game
Right now, combat drones are utterly broken, never attacking anyone.
Much of their code is also terrible, running in Life(), using src. everywhere, New() instead of Initialize(), etc.
This PR addresses the many issues with them and their code.

## Changelog
:cl: Kyep
add: Rewrote the "rogue combat drone" event code. Combat drones are now dangerous, but options exist to evade or fight them.
/:cl:

### DIFF
--- a/code/modules/events/rogue_drones.dm
+++ b/code/modules/events/rogue_drones.dm
@@ -4,24 +4,16 @@
 	var/list/drones_list = list()
 
 /datum/event/rogue_drone/start()
-	//spawn them at the same place as carp
 	var/list/possible_spawns = list()
 	for(var/thing in GLOB.landmarks_list)
 		var/obj/effect/landmark/C = thing
-		if(C.name == "carpspawn")
+		if(C.name == "carpspawn") //spawn them at the same place as carp
 			possible_spawns.Add(C)
 
-	//25% chance for this to be a false alarm
-	var/num
-	if(prob(25))
-		num = 0
-	else
-		num = rand(2,6)
-	for(var/i=0, i<num, i++)
-		var/mob/living/simple_animal/hostile/retaliate/malf_drone/D = new(get_turf(pick(possible_spawns)))
+	var/num = rand(2, 12)
+	for(var/i = 0, i < num, i++)
+		var/mob/living/simple_animal/hostile/malf_drone/D = new(get_turf(pick(possible_spawns)))
 		drones_list.Add(D)
-		if(prob(25))
-			D.disabled = rand(15, 60)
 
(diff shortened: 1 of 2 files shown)


Cocacolagua:
an immolator shot won’t do anything in space, the fire will extinguish instantly

Kyep:
During testing, that wasn't what happened.

Kyep:
Yeah, I've been thinking of changing it to a single-shot standard laser. But the triple-shot effect looks cool and so does the fire effect (even if it only lasts a moment before space extinguishes it).

Qwertytoforty:
Hmm, maybe put it back to what it used to be as a triple shot 15 damage laser? or maybe just 4-5 shots of the immolator, would get similar effects.

Carthusia:
would it be possible to make them emaggable, or in some way hijackable by a syndicate agent?

Kyep:
@Qwertyotoforty It already has similar damage to what it had before, with a momentary flame effect in space. I think that's fine.

@Carthusia Not really. Many issues with that idea.

variableundefined (approved):
This make the event an actual moderate threat and also worth dealing with.

PR #14041, opened 2020-08-08
Revert #12765 - make offstation roles immune to cult conversion rune again

Kyep:
## What Does This PR Do
#12765 changed offstation roles (like lavaland golems, ashwalkers, etc) so that they could be converted to cult.
It was done because of bug #12764. 
That bug has now been found, and fixed.
So, this PR reverts #12765, and makes offstation roles like lavaland golems/cult/ashwalkers once again immune to cult conversion using the convert rune.

## Why It's Good For The Game
Admins are really tired of cult basing in lavaland and converting all the free golems/ashwalkers/etc.

## Note
This doesn't block cultists from killing these people, then sharding them, and turning them into constructs.
While I could block that as well, I'd like some feedback before doing so.
It also doesn't block cult from saccing those people.
It simply blocks standard conversion.

## Changelog
:cl: Kyep
balance: offstation roles (e.g: all lavaland ghost roles) can no longer be converted to cult via the conversion rune.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult.dm
+++ b/code/game/gamemodes/cult/cult.dm
@@ -231,6 +231,8 @@ GLOBAL_LIST_EMPTY(all_cults)
 /datum/game_mode/cult/proc/get_unconvertables()
 	var/list/ucs = list()
 	for(var/mob/living/carbon/human/player in GLOB.player_list)
+		if(player.mind && player.mind.offstation_role)
+			continue
 		if(!is_convertable_to_cult(player.mind))
 			ucs += player.mind
 	return ucs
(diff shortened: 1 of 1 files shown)


Carthusia:
I fully get why we dont want lavalanders as humanoid cultists.

The other day there was a cult base INSIDE of the greenhouse ship, and the admins online had to shut down their abiliy to spawn from their pods, because if they did, BOOM, they woulda been culted and the cult would have up to 4 new members, which also would have allowed ghost players to think "Oh hey, if i spawn as this i can join the cult right away and be an antag"

Sharding i think is still fine as a shard construct is FAR less useful in the early stages than a humanoid member and are best for late game cult/round end cult activities

PR #14042, opened 2020-08-08
Fixes and re-activates SSchat

dearmochi:
## What Does This PR Do

Takes up the torch from @SteelSlayer's `SSchat` fix (itself ported from TG):
- **Fixes `SSchat` not converting `\n` characters into line breaks (`<br>`), causing some text (e.g. human examine) to not render properly due to being one big `\n`-separated message rather than a succession of lines.**
- Fixes minor bug where condensed chat message badge would show up in a separate line rather than the repeated one's.
- Documents `SSchat` and `to_chat` procs.
- Re-activates `SSchat` which was added in #11921 but disabled in #13088 due to the above bug in bold.

## Why It's Good For The Game

As stated in #13088:
> The chat subsystem queues new chat messages and prevents lag spikes that usually happened when you'd be in a situation where you'd get a lot of messages at once, this means fastmos no longer KILLS your game, at least in my tests with 5 monkeys and 15 glass shards and deleting the floor, the game didn't even hang and this was on my laptop. Big performance improvement in that **regard**

**tl;dr perf improvement**

## Changelog
:cl:
tweak: Re-activate chat message buffering, increasing performance in message-heavy situations
fix: Fix condensed chat message badge appearing in a separate line
/:cl:

### DIFF
--- a/code/controllers/subsystem/chat.dm
+++ b/code/controllers/subsystem/chat.dm
@@ -1,15 +1,21 @@
 SUBSYSTEM_DEF(chat)
 	name = "Chat"
-	flags = SS_TICKER|SS_NO_INIT
-	wait = 1
+	flags = SS_TICKER
+	wait = 0.1 SECONDS
 	priority = FIRE_PRIORITY_CHAT
 	init_order = INIT_ORDER_CHAT
 	offline_implications = "Chat messages will no longer be cleanly queued. No immediate action is needed."
+	// Associative list (/client => /text) containing messages to be sent to the client next fire
+	var/list/payload
 
-	var/list/payload = list()
-
+/datum/controller/subsystem/chat/Initialize()
+	LAZYINITLIST(payload)
+	return ..()
 
 /datum/controller/subsystem/chat/fire()
+	if(!payload)
+		return
+
 	for(var/i in payload)
 		var/client/C = i
 		if(C)
@@ -19,8 +25,16 @@ SUBSYSTEM_DEF(chat)
 		if(MC_TICK_CHECK)
 			return
 
-
-/datum/controller/subsystem/chat/proc/queue(target, message, flag)
+/**
+  * Queues a message to be sent to one or multiple targets next fire.
+  *
+  * Arguments:
+  * * target - The target or (list of targets) to send the message to
+  * * message - The message in HTML
+  * * handle_whitespace - Whether \n and \t should be converted into HTML counterparts
+  * * trailing_newline - Whether a line break should be added at the end of the message
+  */
+/datum/controller/subsystem/chat/proc/queue(target, message, handle_whitespace = TRUE, trailing_newline = TRUE)
 	if(!target || !message)
 		return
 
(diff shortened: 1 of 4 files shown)


goon/browserassets/js/browserOutput.js
```diff
@@ -92,15 +92,24 @@ if (typeof String.prototype.trim !== 'function') {
...
+		return this.substring(this_len - search.length, this_len) === search;
+	};
+}
 
 //Polyfill for string.prototype.includes. Why the fuck. Just why the fuck.
 if (!String.prototype.includes) {
 	String.prototype.includes = function(search, start) {
 	  'use strict';
-  
+
```
dearmochi, line 109: Seems like some whitespace was added or removed due to an extension in VSCode, apologies for the noise.

SteelSlayer: It's perfectly normal, everyone's VSC does it. Mine does anyway.

code/controllers/subsystem/chat.dm
```diff
@@ -1,15 +1,21 @@
 SUBSYSTEM_DEF(chat)
 	name = "Chat"
-	flags = SS_TICKER|SS_NO_INIT
-	wait = 1
+	flags = SS_TICKER
+	wait = 0.1 SECONDS
 	priority = FIRE_PRIORITY_CHAT
 	init_order = INIT_ORDER_CHAT
 	offline_implications = "Chat messages will no longer be cleanly queued. No immediate action is needed."
+	// Associative list (/client => /text) containing messages to be sent to the client next fire
```
SteelSlayer, line 8: Missing a third `/` here for autodoc.

dearmochi: Done

AffectedArc07:
I am kind of skeptical of the whole SSchat thing, for a few reasons

1. Goonchat has a habit of missing messages from time to time. Does this mean that a client can potentially miss an entire group of messages? Also if the payload is many, many messages, wont this be a large amount of network overhead going to a client, causing major lag spikes
2. I am a little concerned with RAM in cases like when the tesla is in xenboiology and it will create a queue of about 400 messages for about 50 clients. Assuming 100 characters per message (including formatting overhead), thats 40k per person. This then becomes a full 2MB of RAM which may not be fully cleared out properly depending on how well DM clears lists out
3. I dont like the idea of subsystems which fire every single tick which can have major CPU impact to replace something which already worked fine. (Looking at you `SSinput`)

This isnt an objection, but it is food for thought

dearmochi:
@AffectedArc07 

1. That's what bothers me with the current implementation of `SSchat`. Each entry in the payload list corresponds to a message for a client. That indeed results to a lot of overhead if there are a lot of messages for one client pending. **I personally think it'd be better to consolidate all these messages into a single buffer for each client.** That way only one (big) payload is sent to the client at once rather than multiple small ones. We're past the age of 56k so everyone should be able to handle a big message over the network.

2. See 1.

3. Suggestion: do away with `SSchat` and instead use timers with a global list.

AffectedArc07:
In regards to the handling of big messages. Everyones connection can handle it, but BYOND cant. A 3.5k character message hangs the client on my laptop (Has an i5-8300H and runs faster than my RYZEN PC) for about half a second. A message of 40k characters would probably hang it for about 5 seconds, Considering the tesla moves every 2 seconds (Every time `SSobj` fires), this could potentially fully lockup a client, which is not ok.

SteelSlayer:
I did test this a couple months ago, with like 100 slimes and a tesla spamming shock messages in xenobiology. I got zero chat lag while I stood there for minutes. But then again I was alone on a test server.

dearmochi:
Closing this PR while I think up a better solution. `SSchat` has many flaws as @AffectedArc07 pointed out.

PR #14057, opened 2020-08-10
Fixes blob mouse splatting, refactors mice a bit

### LINKED ISSUE #13823: Runtime the cat kills blobs by itself in seconds
**Issue Description**:
Runtime can Splat() blob-infected mice, which sets them to stat = DEAD without calling death(), and thus without triggering the blob mouse's transformation into a blob.

Runtime (or anything else!) kills the infected mouse.
Mouse instantly becomes blob on death.

**What happened instead**:
Runtime Splat()ed mouse. Mouse died. No blob was created.

Runtime can single handedly kill a blob before it even bursts.

1) Spawn /mob/living/simple_animal/mouse/blobinfected and put yourself in control of it
2) Spawn /mob/living/simple_animal/pet/cat/Runtime
3) Wait a few seconds
4) You are now Splatted by runtime, annihilator of blobs, and never got a chance to live up to your blobby potential

**Extra information**:
Does https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/mob/living/simple_animal/friendly/mouse.dm#L83 even need to exist? Why not just call .death()? That Splat() proc is terrible quality code littered with "src.". I doubt we'd accept it nowadays.

Kyep:
## What Does This PR Do
Fixes #13823 - runtime the cat in the CMO office can kill blob-infected mice in seconds, preventing them from bursting.
Refactors mice to not have a silly snowflake splat() proc which behaves like death() except doesn't actually call death().
Dead mice now always have a small pool of blood next to them, as opposed to only mice killed via splat() having this.

## Why It's Good For The Game
Fixes a bug.
Refactors the code to reduce special cases.
Makes it more obvious that dead mice are in fact dead and not just sleeping or playing dead.

## Images of changes
runtime_mice

## Changelog
:cl: Kyep
fix: fixed runtime the cat insta-slaying blob mice without triggering their transformation into blobs.
tweak: dead mice now always have blood on their sprite.
/:cl:


AffectedArc07:
Im not sure the mouse being covered in blood is always correct, such as asphyxiation, or when they bite wires (the blood splat will be pure black and look janky)

Code is fine though

Kyep:
@AffectedArc07 PR updated. No longer has any sprite changes / no longer needs sprites label.
Now simply refactors it so that mousetrap/cat splatting a mouse doesn't runtime.

PR #14059, opened 2020-08-11
[TGUI] Tank Transfer Valves (TTV)

Kyep:
## What Does This PR Do
Converts Tank Transfer Valves (TTVs) to TGUI.

## Images of changes
TTV-default
TTV-full

## Changelog
:cl: Kyep
tweak: Converted TTVs (tank transfer valves, the bomb component in toxins) to TGUI.
/:cl:


tgui/packages/tgui/interfaces/TransferValve.js
```diff
@@ -0,0 +1,90 @@
...
+        </Section>
+        <Section
+          title="Valve Attachment"
+          buttons={(
+            <Button
+              textAlign="center"
+              width="30px"
+              icon={"cog"}
+              disabled={(!attached_device)}
+              onClick={() => act('device')} />
```
AffectedArc07, line 35: Maybe a tooltip which says something like `Configure Assembly`?

code/game/objects/items/devices/transfer_valve.dm
```diff
@@ -86,64 +86,64 @@
...
 		ui.open()
-		// auto update every Master Controller tick
-		//ui.set_auto_update(1)
 
-/obj/item/transfer_valve/ui_data(mob/user, ui_key = "main", datum/topic_state/state = GLOB.default_state)
-	var/data[0]
+/obj/item/transfer_valve/tgui_data(mob/user)
+	var/list/data = list()
+	data["tank_one"] = tank_one
+	data["tank_two"] = tank_two
```
AffectedArc07, line 100: I would use the old ternary check here as its less overhead to send the name than the entire object

code/game/objects/items/devices/transfer_valve.dm
```diff
@@ -86,64 +86,64 @@
...
+	data["attached_device"] = attached_device
```
AffectedArc07, line 101: Same here

AffectedArc07:
Looking back on this, I notice a few things
image
The wrench icon doesnt make sense considering its the detach button. I feel like the eject button would make more sense.

image
The use of the word "insert" here feels weird because the TTV isnt a container you "insert" things into. Its a valve stuff attaches to as seen below. 
image

Not major, but it just seems weird.

Kyep:
@AffectedArc07 Changed wording, updated screenshots.

PR #14061, opened 2020-08-11
Ranged holoparasites no longer obliterate admin attack logs

Kyep:
## What Does This PR Do
Makes the ranged holoparasite's attacks (the green spray) take 5x as long to reload between shots... but also do 5x the damage.

## Why It's Good For The Game
The current attack speed (0.1 seconds!) is absolutely obnoxious for admins watching attack logs. It completely drowns out everything else in chat. This change makes it much more bearable, without changing its damage per second.

## Changelog
:cl: Kyep
tweak: ranged holoparasite now has 5x longer firing cooldown, but also 5x the damage. Damage per second is unchanged.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/guardian/types/ranged.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/ranged.dm
@@ -1,7 +1,7 @@
 /obj/item/projectile/guardian
 	name = "crystal spray"
 	icon_state = "guardian"
-	damage = 5
+	damage = 25
 	damage_type = BRUTE
 	armour_penetration = 100
 
@@ -11,7 +11,7 @@
 	melee_damage_upper = 10
 	damage_transfer = 0.9
 	projectiletype = /obj/item/projectile/guardian
-	ranged_cooldown_time = 1 //fast!
+	ranged_cooldown_time = 5 //fast!
 	projectilesound = 'sound/effects/hit_on_shattered_glass.ogg'
 	ranged = 1
 	range = 13


TDSSS:
This seems like it could have quite a few subtle balance implications, ie against shielded hardsuits or block chances where your damage is now less consistent.

Is there no way this can be solved on the log side of things instead, ie collect similar logs together in chat with an x5 behind them or something like it? Or make every click fire a burst, like burst-fire weapons? Or do those also generate 1 log for every bullet even if there's 3 bullets per click?

SteelSlayer:
The issue is that people can use autoclickers, or just click really fast to rack up several thousand damage per minute. So even if you limit the logs, the actual speed at which you can fire is just OP. The burst fire idea could maybe work though.

Kyep:
@TDSSS Holoparas do not generally fight anyone in a shielded hardsuit. Against block chance, yeah its a bit more variable but IMHO admins not drowning in attack logs is far more important than the attack always doing the exact same damage against one type of target.
We already have a system where messages group up, but when you're aghosting/etc things can happen so fast that even if it stacked up repeated messages, there's a good chance that things would happen in between the 0.1s fire rate of the projectile as-is. Remember, aghosts see EVERYTHING that happens in world. It generates 1 log per projectile, which at 10 projectiles/second is simply far too much.

@SteelSlayer I'm not changing overall damage in this PR. I'm simply combining the same amount of damage per second into fewer projectiles, so it doesn't mess up admins' attack logs so badly.

SteelSlayer:
Yeah I know, my response was directed at TDSSS about the thing they said: "Is there no way this can be solved on the log side of things instead, ie collect similar logs together in chat with an x5 behind them or something like it?" I forgot to @ them, sorry.

TDSSS:
Mining guardians can sometimes be used against wizards or nukies, but I admit it's super rare.

What about the salvo/burst fire idea?

Kyep:
Each hit generates an attack log. Number of clicks doesn't matter. Only number of projectiles that hit does. The only way to reduce the attack log spam is to reduce the number of projectiles.

Qwertytoforty:
This is great,  the shear spam of the projectiles, either as a person watching it, or being anywhere else with attack logs on, is just too high from one singular mob.

PR #14062, opened 2020-08-11
Revert "Make wet overlays render on MID_TURF_LAYER plane"

Kyep:
Reverts ParadiseSS13/Paradise#14023

This is an emergency revert.

Reason: original PR breaks the game. Wet floor overlay blocks you from attacking things you want to attack - like blob tiles. And people.


AffectedArc07:
Heres the real question. Why are mobs and structures below the `MID_TURF_LAYER`

PR #14064, opened 2020-08-11
Add FLOOR_OVERLAY_PLANE (and use that for wetness)

dearmochi:
## What Does This PR Do

Adds plane `FLOOR_OVERLAY_PLANE` and uses it for wetness instead because #14023 breaks with blobs.

## Why It's Good For The Game

Slippery overlays no longer appear above blobs

## Changelog
:cl:
fix: Fix slippery overlays rendering above blobs
/:cl:


AffectedArc07:
So it turns out that every `/atom` is on `GAME_PLANE` by default and there are 0 overrides, despite the fact we have `MOB_LAYER` and `FLOOR_PLANE` and turf layers. Essentially our current plane system is 99% useless. Someone some day (Not in this PR) needs to go through the planelist and actually assign things properly, but that will take time, be a pain to review, and probably break stuff.

TLDR: This is fine and I hate our current planes.

PR #14065, opened 2020-08-11
[TGUI] Shuttle console

TDSSS:
## What Does This PR Do
This PR converts our shuttle UIs to TGUI

Note that this is my first time working with tgui, mostly to figure out how it even works. As such, I probably messed up in some places.

Note, I also removed a var from the shuttle console that had no way of getting set and didn't port the code that looked at it in the UI, since it was unreachable.

## Why It's Good For The Game
We want everything converted to TGUI, this gets us one step closer.

## Images of changes
new and improved screenshots:
Untitled
all UIs are at their default size.

## Changelog
:cl:
tweak: port shuttle consoles to tgui
/:cl:


code/modules/mining/laborcamp/laborshuttle.dm
```diff
@@ -14,8 +14,8 @@
 	circuit = /obj/item/circuitboard/labor_shuttle/one_way
 	req_access = list( )
 
-/obj/machinery/computer/shuttle/labor/one_way/Topic(href, href_list)
-	if(href_list["move"])
+/obj/machinery/computer/shuttle/labor/one_way/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
```
dearmochi: `list/` can be omitted from `list/params` and the last 2 arguments are typically not added to the signature because there are almost no cases they'll be used.

code/modules/mining/laborcamp/laborshuttle.dm
```diff
@@ -24,4 +24,5 @@
 		if(S && S.name == "laborcamp_away")
 			to_chat(usr, "<span class='warning'>Shuttle is already at the outpost!</span>")
 			return 0
```
dearmochi: `0` becomes `FALSE`

code/modules/mining/laborcamp/laborshuttle.dm
```diff
@@ -24,4 +24,5 @@
...
-	..()
+	..()	//I know the tutorial says if(..()) should always be at the start, but that won't work here
```
dearmochi: Don't do this. If you want to add checks before any `tgui_act` in a subclass, create and call a proc called something like `can_call_shuttle` in the parent class (`/obj/machinery/computer/shuttle` I'm guessing). What you're doing right now could break a lot of things and is exploit prone.

code/modules/shuttle/ert.dm
```diff
@@ -6,14 +6,13 @@
 	resistance_flags = INDESTRUCTIBLE
 	flags = NODECONSTRUCT
 
-/obj/machinery/computer/shuttle/ert/Topic(href, href_list)
-	if(href_list["move"])
+/obj/machinery/computer/shuttle/ert/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
```
dearmochi, line 9: Ditto

code/modules/shuttle/shuttle.dm
```diff
@@ -773,21 +773,20 @@
 		return
 	connect()
 	add_fingerprint(user)
-	ui_interact(user)
+	tgui_interact(user)
 
-/obj/machinery/computer/shuttle/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)
-	var/obj/docking_port/mobile/M = SSshuttle.getShuttle(shuttleId)
-	ui = SSnanoui.try_update_ui(user, src, ui_key, ui, force_open)
+/obj/machinery/computer/shuttle/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = 0, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
```
dearmochi, line 778: `force_open = FALSE`

code/modules/mining/laborcamp/laborshuttle.dm
```diff
@@ -13,15 +13,3 @@
 	possible_destinations = "laborcamp_away"
 	circuit = /obj/item/circuitboard/labor_shuttle/one_way
 	req_access = list( )
-
-/obj/machinery/computer/shuttle/labor/one_way/Topic(href, href_list)
```
TDSSS, line 17: explanation: The one_way console only has one possible destination, laborcamp_away. This means when the shuttle is at that destination the destination list is empty and you see Shuttle Locked. And if there is no shuttle, the UI isn't shown either.

So as far as I can tell, there's no way for this code to ever matter at all.

code/modules/shuttle/shuttle.dm
```diff
@@ -799,34 +798,31 @@
...
 		data["admin_controlled"] = admin_controlled
-		data["docking_request"] = docking_request
-
 	return data
 
-/obj/machinery/computer/shuttle/Topic(href, href_list)
-	if(..())
-		return 1
-
+/obj/machinery/computer/shuttle/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
```
dearmochi, line 803: Same

code/modules/shuttle/shuttle.dm
```diff
@@ -799,34 +798,31 @@
...
+	if(..())	//we can't actually interact, so no action
+		return TRUE
 	if(!allowed(usr))
 		to_chat(usr, "<span class='danger'>Access denied.</span>")
-		return
-
+		return	//I am not sure if/why this returning instead of returning 1 matters here, original topic did it. Change?
```
dearmochi, line 808: `return FALSE` to prevent updating

code/modules/shuttle/ert.dm
```diff
@@ -6,14 +6,13 @@
...
+	if(action == "move")
 		var/authorized_roles = list(SPECIAL_ROLE_ERT, SPECIAL_ROLE_DEATHSQUAD)
 		if(!((usr.mind?.assigned_role in authorized_roles) || is_admin(usr)))
 			message_admins("Potential ERT shuttle hijack, ERT shuttle moved by unauthorized user: [key_name_admin(usr)]")
 	..()
```
TDSSS, line 14: had a discussion with AA on this on discord, regarding not having if(..()) return at the start. He said "
 there isnt any better way to do this
 shuttles suck, whats new" So leaving it like this for the time being.

code/modules/shuttle/shuttle.dm
```diff
@@ -799,34 +798,31 @@
...
 				atom_say("Shuttle departing! Please stand away from the doors.")
 				usr.create_log(MISC_LOG, "used [src] to call the [shuttleId] shuttle")
+				if(!moved)
+					moved = TRUE
 			if(1)
 				to_chat(usr, "<span class='warning'>Invalid shuttle requested.</span>")
 			else
 				to_chat(usr, "<span class='notice'>Unable to comply.</span>")
-		return 1
+		return TRUE
```
dearmochi, line 826: `return FALSE` to prevent updating as a default case.

code/modules/shuttle/shuttle.dm
```diff
@@ -799,34 +798,31 @@
...
```
dearmochi, line 825: `add_fingerprint(usr)`

code/modules/shuttle/shuttle.dm
```diff
@@ -727,13 +727,13 @@
...
 	icon_keyboard = "tech_key"
-	req_access = list( )
+	req_access = list()
 	circuit = /obj/item/circuitboard/shuttle
 	var/shuttleId
 	var/possible_destinations = ""
 	var/admin_controlled
 	var/max_connect_range = 7
-	var/docking_request = 0
+	var/moved = FALSE	//workaround for nukie shuttle, hope I find a better way to do this...
```
TDSSS, line 736: the security check needing to be invoked forced me to move this to here, though. Since otherwise you might be able to fake-move the nukie shuttle and deny them a wardec or something with an exploit.

code/modules/shuttle/shuttle.dm
```diff
@@ -799,34 +798,31 @@
...
-		if(!options.Find(href_list["move"])) //I see you're trying Href exploits, I see you're failing, I SEE ADMIN WARNING.
-			// Seriously, though, NEVER trust a Topic with something like this. Ever.
-			// Sidenote for whoever did this last. Why did you set it to echo whatever the user entered to admin chat? You solved one exploit and created another.
+	if(action == "move")
+		var/destination = params["move"]
+		if(!options.Find(destination))//figure out if this translation works
 			message_admins("<span class='boldannounce'>EXPLOIT:</span> [ADMIN_LOOKUPFLW(usr)] attempted to move [src] to an invalid location! [ADMIN_COORDJMP(src)]")
 			return
-		switch(SSshuttle.moveShuttle(shuttleId, href_list["move"], 1, usr))
+		switch(SSshuttle.moveShuttle(shuttleId, destination, 1, usr))
```
dearmochi, line 815: `1` becomes `TRUE` (3rd argument is a boolean)

code/modules/shuttle/shuttle.dm
```diff
@@ -849,20 +845,21 @@
 	admin_controlled = 1
```
dearmochi, line 845: `1` becomes `TRUE`

code/modules/shuttle/shuttle.dm
```diff
@@ -849,20 +845,21 @@
...
 	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | ACID_PROOF
 
-/obj/machinery/computer/shuttle/ferry/request/Topic(href, href_list)
+/obj/machinery/computer/shuttle/ferry/request/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
```
dearmochi, line 848: Same

code/modules/shuttle/shuttle.dm
```diff
@@ -849,20 +845,21 @@
...
-		cooldown = 1
+		cooldown = TRUE
 		to_chat(usr, "<span class='notice'>Your request has been recieved by Centcom.</span>")
 		log_admin("[key_name(usr)] requested to move the transport ferry to Centcom.")
 		message_admins("<b>FERRY: <font color='blue'>[key_name_admin(usr)] (<A HREF='?_src_=holder;secretsfun=moveferry'>Move Ferry</a>)</b> is requesting to move the transport ferry to Centcom.</font>")
-		. = 1
-		SSnanoui.update_uis(src)
+		. = TRUE
+		SStgui.update_uis(src)
 		spawn(600) //One minute cooldown
```
dearmochi, line 860: Using `addtimer` is preferable here. You could even get rid of the `cooldown` variable and instead use `world.timer` and adding whatever duration you want. Do make a define for the cooldown, avoid magic numbers.

Also `add_fingerprint(usr)`

code/modules/shuttle/shuttle.dm
```diff
@@ -849,20 +845,21 @@
...
-			cooldown = 0
+			cooldown = FALSE
+
```
dearmochi, line 862: `return FALSE` as default case

code/modules/shuttle/shuttle.dm
```diff
@@ -849,20 +845,21 @@
...
 			return
-		cooldown = 1
+		cooldown = TRUE
 		to_chat(usr, "<span class='notice'>Your request has been recieved by Centcom.</span>")
 		log_admin("[key_name(usr)] requested to move the transport ferry to Centcom.")
 		message_admins("<b>FERRY: <font color='blue'>[key_name_admin(usr)] (<A HREF='?_src_=holder;secretsfun=moveferry'>Move Ferry</a>)</b> is requesting to move the transport ferry to Centcom.</font>")
-		. = 1
-		SSnanoui.update_uis(src)
+		. = TRUE
+		SStgui.update_uis(src)
```
dearmochi, line 859: Don't need this. Returning `TRUE` from `tgui_act` will automatically update the UI.

code/modules/shuttle/syndicate.dm
```diff
@@ -11,19 +11,18 @@
...
-	var/moved = FALSE
 
 /obj/machinery/computer/shuttle/syndicate/recall
 	name = "syndicate shuttle recall terminal"
 	circuit = /obj/item/circuitboard/shuttle/syndicate/recall
 	possible_destinations = "syndicate_away"
 
-/obj/machinery/computer/shuttle/syndicate/Topic(href, href_list)
-	if(href_list["move"])
+/obj/machinery/computer/shuttle/syndicate/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
```
dearmochi, line 20: Ditto

code/modules/shuttle/syndicate.dm
```diff
@@ -11,19 +11,18 @@
...
+	if(action == "move")
 		if(challenge && world.time < SYNDICATE_CHALLENGE_TIMER)
 			to_chat(usr, "<span class='warning'>You've issued a combat challenge to the station! You've got to give them at least [round(((SYNDICATE_CHALLENGE_TIMER - world.time) / 10) / 60)] more minutes to allow them to prepare.</span>")
 			return 0
-		moved = TRUE
+		moved = TRUE	//MYTODO safety check this somehow, don't make affected cry
 	..()
```
dearmochi, line 25: Same

code/modules/shuttle/syndicate.dm
```diff
@@ -35,8 +34,8 @@
 	shuttleId = "steel_rain"
 	possible_destinations = null
 
-/obj/machinery/computer/shuttle/syndicate/drop_pod/Topic(href, href_list)
-	if(href_list["move"])
+/obj/machinery/computer/shuttle/syndicate/drop_pod/tgui_act(action, list/params, datum/tgui/ui, datum/tgui_state/state)
```
dearmochi, line 36: Here too

tgui/packages/tgui/interfaces/ShuttleConsole.js
```diff
@@ -0,0 +1,71 @@
...
+                </Fragment>
+              )}
+              {/* end of the docking_ports_len if/else block
+             there is an (as far as I can tell) unreachable
+             piece of code in the original tmpl since
+             docking_request doesn't seem to be ever set
+             so it's probably not needed?*/}
+            </Fragment>
+          )}
+        </LabeledList>
```
dearmochi: Should probably wrap that in a `Section` so it looks better.

tgui/packages/tgui/interfaces/ShuttleConsole.js
```diff
@@ -0,0 +1,71 @@
...
+                        content={port.name}
+                        // TODO: should button have icon?
+                        onClick={() => act('move', {
+                          move: port.id,
+                        })} />
+                    )
+                    )}
+                  </LabeledList.Item>
+                  {/* Apparently I need to comment like this sometimes??
+                Also end of the button at /> up there.*/}
```
dearmochi, line 39: Comments are nice but keep that in a diary please.

code/modules/shuttle/syndicate.dm
```diff
@@ -35,8 +34,8 @@
...
+	if(action == "move")
 		if(z != level_name_to_num(CENTCOMM))
 			to_chat(usr, "<span class='warning'>Pods are one way!</span>")
 			return 0
```
dearmochi, line 40: `0` becomes `FALSE`. Also don't call `..()` like this as I pointed out in my first comment or so.

tgui/packages/tgui/interfaces/ShuttleConsole.js
```diff
@@ -0,0 +1,71 @@
...
+              {!!data.docking_ports_len && (
+              // I am not sure if I am nesting this stuff right...
+                <Fragment>
+                  <LabeledList.Item label={"Send to "}>
+                    {data.docking_ports.map(port => (
+                      <Button
+                        icon="chevron-right"
+                        key={port.name}
+                        content={port.name}
+                        // TODO: should button have icon?
```
dearmochi: If you can find a good icon, sure. Here's the [list](https://fontawesome.com/icons?d=gallery).

code/modules/mining/laborcamp/laborshuttle.dm
TDSSS: oh, now that's an approach I hadn't thought of. Thanks!

code/modules/shuttle/shuttle.dm
TDSSS: tgui_interact gets called from attack_hand, which already adds fingerprints, doesn't it? It shouldn't be possible to be in a tgui_act without a tgui_interact before, right?

code/modules/shuttle/shuttle.dm
TDSSS: I mean, not returning anything is the same as returning false, isn't it?

code/modules/shuttle/shuttle.dm
dearmochi: 1. Wear gloves that don't leave any fingerprints
2. Use the console, firing `tgui_interact` (no prints shouldn't added)
3. Remove your gloves while keeping the window open
4. Press a button or something, firing `tgui_act`. A fingerprint should be added there because the user isn't wearing gloves anymore.

code/modules/shuttle/shuttle.dm
dearmochi: Personally I'd say do it for clarity's sake.

AffectedArc07:
image
Shorten this to just `Authorization:`. Bloating the entire width of the UI out for one list marker looks bad.

TDSSS:
The ..() calls at the end should be gone now, also shortened the Authorization text. Not sure it needs a new picture since it's a fairly minor change.

AffectedArc07:
As a rule of thumb you should always be updating the screenshots if something ever changes

TDSSS:
Capture

TDSSS:
Fingerprints have been added to any move of the shuttle.

code/modules/shuttle/shuttle.dm
TDSSS: done

TDSSS:
Prettied up the windows a bit, adding a section and using noticeboxes. Screenshots in original post are updated.

code/modules/shuttle/shuttle.dm
```diff
@@ -844,25 +848,23 @@
...
 	circuit = /obj/item/circuitboard/ferry/request
-	var/cooldown //prevents spamming admins
+	var/next_request	//to prevent spamming admins
 	possible_destinations = "ferry_home"
-	admin_controlled = 1
+	admin_controlled = TRUE
 	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | ACID_PROOF
 
-/obj/machinery/computer/shuttle/ferry/request/Topic(href, href_list)
+/obj/machinery/computer/shuttle/ferry/request/tgui_act(action, params)
```
AffectedArc07, line 856: Did you mean `can_call_shuttle()` here? Youre making the parent TGUI call here, which is going to cause issues.

TDSSS: Can you explain _why_ it's going to cause issues? I don't see why it should be can_call_shuttle

AffectedArc07: After doing some thourough testing, this *shouldnt* circumvent security measures, but is still a little confusing given that 
```dm
if(..())
    return
```
is the standard for security checking, which while this does do, it does a lot more than that. May want to add a comment clarifying this.

AffectedArc07 (approved):

PR #14066, opened 2020-08-12
Remove highlight length limit and fix highlights breaking the DOM

craftxbox:
## What Does This PR Do
Removes the limit for highlight length, and stops highlights containing " from breaking the DOM.

## Why It's Good For The Game
The highlight length limit imposes problems to those wishing to create regex highlights to match all their character names (and, to some extent, normal highlights ~~@Spacemanspark~~). As far as I can tell, removing it will not cause any issues.
Aswell, The way the highlight popup was originally written, inserted highlight terms as text into the DOM, instead of setting the value of their input boxes, This caused any highlight containing " to cut off when the highlight panel was opened, and could, in theory, break the dom if other special characters such as > were placed after it. This has now been fixed by manipulating the input box elements in code.

## Changelog
:cl:
del: Removed length limit for highlights
fix: Highlight terms being able to break the dom when the highlight panel opens.
/:cl:

I do not understand the person who originally wrote goonchat, The functions for manipulating the DOM in code had to have existed when it was written, but instead, they opted to do popups by appending html code as text? I may refactor this in the future, or completely rewrite goonchat. But if someone else wants to instead, go right ahead.

(posted as draft awaiting for me to test it properly, which may take a while)

### DIFF
--- a/goon/browserassets/js/browserOutput.js
+++ b/goon/browserassets/js/browserOutput.js
@@ -1049,7 +1049,7 @@ $(function() {
 		if ($('.popup .highlightTerm').is(':visible')) {return;}
 		var termInputs = '';
 		for (var i = 0; i < opts.highlightLimit; i++) {
-			termInputs += '<div><input type="text" name="highlightTermInput'+i+'" id="highlightTermInput'+i+'" class="highlightTermInput'+i+'" maxlength="255" value="'+(opts.highlightTerms[i] ? opts.highlightTerms[i] : '')+'" /></div>';
+			termInputs += '<div><input type="text" name="highlightTermInput'+i+'" id="highlightTermInput'+i+'" class="highlightTermInput'+i+'" value="" /></div>';
 		}
 		var popupContent = '<div class="head">String Highlighting</div>' +
 			'<div class="highlightPopup" id="highlightPopup">' +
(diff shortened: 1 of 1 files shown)


AffectedArc07:
Please, thoroughly test this. Highlights were broken for a full year because of previous highlight change PRs. This PR is also a webedit which gives me even less confidence. 
image
image

craftxbox:
I know, This is why i posted this as a draft.

craftxbox:
> What's to stop someone putting in some ridiculously long regex, and what would that do?

If someone enters an absurdly long regex, its just going to be slower, But the main issue is within catastrophic backtracing, which can't be addressed in code as far as my research has told, and can be achieved in very, very tiny regexes anyways.

> Frankly, I think our highlighting code is overly-complex. 

This, I can agree with, however, you do need to keep in mind the entirety of goonchat is overly complex aswell.

craftxbox:
Update: There *is* a library out there that may work to stop the backtracing issue, https://github.com/davisjam/safe-regex , however, it is not compatible with IE11 and would need to go through babel.

craftxbox:
> What's to stop someone putting in some ridiculously long regex,

I have just tested with 5 highlights of 308 thousand characters (totalling abouat 1.5m) and found negligable change in performance

craftxbox:
Regexes that catastrophically backtrace appear to cause dreamseeker to crash before they end up saving to cookies.

craftxbox:
I've tested everything that i've seen cause problems in the past, the things changed by this pr, and some other things (ie, the backtracing problem) and have not found any problems. Marking ready for review/merge.

AffectedArc07:
`If someone enters an absurdly long regex, its just going to be slower`

Goonchat is client side. Its not going to slow the server down if a user puts an absurdly long regex in. If their PC is beef enough to handle it, let the user put a long ass regex in. If its not, they can clear their IE cookies, which is standard procedure for goonchat issues.

variableundefined:
I'm unwilling to take the risk of breaking highlighting again for infinitely long regex. I don't think the use cases are worth the risk. In addition, this is a webedit.

PR #14067, opened 2020-08-12
Ashstorm MMI fixes + spacepod immunity

### LINKED ISSUE #13926: Ash Storm causes MMI failure
As a miner borg you are now immune to ash storms, but apparently the MMI inside of you is not.

The borgs MMI will die if you stand out in the storm, and while this does not disable your borg, if your brain is removed from your borg you will die because your brain is pre-dead.

The MMI to not die because it is inside of a borg that is immune to ash storms

**What happened instead**:
MMI is dead, I grabbed a screenshot: https://cdn.discordapp.com/attachments/572946419028066317/736804814129725530/Ash_Storm.png

Just seems like a strange bug, also some roboticists tend to take borgs apart to repair them and pulling a dead MMI out of a borg that might be repairable is kind of a problem, especially when mmis cannot be repaired.

Be borg, stand in ash storm, wait.

Probably after borgs were given the ability to be immune to ash storms?

**Extra information**:
I was told to add that I was a round start cyborg, and my job preference is set to Robot.

ExusA:
## What Does This PR Do
1. Makes /mob/living/carbon/brain ashstorm immune. The way brains are handled in the case of MMIs is confusing, with robobrains actually having a /carbon/brain mob stored inside them... making brains immune to ash weather solves the [cyborg's] MMI flatlines! message that shows when a cyborg spends time in an ashstorm, which otherwise has no effect on gameplay
2. Edits some code weirdness with how the game checks whether a mob is ashstorm immune, making it hopefully a little cleaner.
3. Being inside a spacepod on lavaland during an ashstorm protects you from ashstorm weather. Nobody ever builds spacepods on lavaland, but when it does rarely happen, it should function like a mech and protect from ashstorms as you might expect it to.

## Why It's Good For The Game
Fixes #13926

🆑
fix: brains can no longer take damage from ash storms
tweak: spacepods now protect from ashstorms
/🆑


ExusA:
First time using Git, bear with me 😅

code/game/mecha/mecha.dm
```diff
@@ -40,6 +40,7 @@
 	var/emagged = FALSE
 	var/frozen = FALSE
 	var/repairing = FALSE
+	var/list/weather_protection_types = "ash"
```
SteelSlayer, line 43: What did you create this var for? It's never used.

ExusA: good question :p

code/datums/weather/weather_types/ash_storm.dm
```diff
@@ -75,15 +75,18 @@
...
 
 /datum/weather/ash_storm/proc/is_ash_immune(atom/L)
-	while(L && !isturf(L))
-		if(ismecha(L)) //Mechs are immune
+	if(L)
+		if(ismecha(L.loc)) //Mechs are immune
+			return TRUE
+		if(isspacepod(L.loc)) //spacepods also protect you
+			return TRUE
+		if(is_ash_immune(L.loc))
```
farie82, line 83: Best to put the recursion after the full check so after line 89. Might also benefit from a small comment about it

ExusA:
redid it again, now every time weather_act() is called for a mob, the game checks if the mob is a mech, a spacepod, or an item (which it isn't), and if none of those conditions are met, checks the mob's loc to see if it's any of those things, and recurs until one of those things is found, and if it isn't, then the mob is NOT ashstorm immune. This means that ANY mob inside a mech or spacepod will be immune, including robobrains, pilots, people in ody sleepers, etc. Also anything in an item is immune which doesn't make as much sense, but player mobs rarely fit in storage items like bags anyway, and this covers the MMI problem without having to add ashstorm immunity to the brain mob itself

ExusA:
messed up somewhere with github, got to close this for now

PR #14070, opened 2020-08-13
feat: add alert window to notify for ghost cloning

Bizzonium:
## What Does This PR Do
Adds alert window to notify for ghost cloning.

## Why It's Good For The Game
To really get the player's attention like other ghost role alerts do.

## Images of changes
Аннотация 2020-08-13 031455

## Changelog
:cl: azizonkg
add: Added alert window to notify for ghost cloning.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -300,6 +300,20 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 				A.overlays += source
 				source.layer = old_layer
 				source.plane = old_plane
+		spawn(0)
+			var/ans = alert(src, "Re-enter your corpse?", message, "Yes", "No")
+			if(!client)
+				return
+			switch(ans)
+				if("Yes")
+					to_chat(src, "<span class='notice'>Choice registered: Yes.</span>")
+					reenter_corpse()
+					return
+				if("No")
+					to_chat(src, "<span class='danger'>Choice registered: No.</span>")
+					return
+				else
+					return
 	to_chat(src, "<span class='ghostalert'><a href=?src=[UID()];reenter=1>(Click to re-enter)</a></span>")
 	if(sound)
 		src << sound(sound)
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -29,6 +29,7 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 	var/data_hud_seen = FALSE //this should one of the defines in __DEFINES/hud.dm
 	var/ghost_orbit = GHOST_ORBIT_CIRCLE
 	var/health_scan = FALSE //does the ghost have health scanner mode on? by default it should be off
+	var/revive_alert_promt = TRUE //used in proc/reenter_corpse_prompt()
 
 /mob/dead/observer/New(mob/body=null, flags=1)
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -300,6 +300,20 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 				A.overlays += source
 				source.layer = old_layer
 				source.plane = old_plane
+		spawn(0)
```
SteelSlayer, line 303: Instead of using `spawn` I recommending creating a proc out of the code you've added here, something like `proc/reenter_corpse_prompt()` (you can name it whatever you want).

And then call the proc with `INVOKE_ASYNC(src, .proc/reenter_corpse_prompt)`

ZomgPonies:
Should probably add a 'not for this round' 3rd option like some of the other spawn popups.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -300,10 +301,32 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
...
+			INVOKE_ASYNC(src, .proc/reenter_corpse_prompt)
+
 	to_chat(src, "<span class='ghostalert'><a href=?src=[UID()];reenter=1>(Click to re-enter)</a></span>")
 	if(sound)
 		src << sound(sound)
 
+/mob/dead/observer/proc/reenter_corpse_prompt()
+	if(!client)
+		return
+	var/ans = alert(src, "Re-enter your corpse?", "Someone is trying to revive you...", "Yes", "No", "Not Until Next Time")
```
SteelSlayer, line 314: To keep consistent with other prompts, it should probably be:
```suggestion
	var/ans = alert(src, "Re-enter your corpse?", "Someone is trying to revive you...", "Yes", "No", "Not this round")
```

code/modules/mob/dead/observer/observer.dm
```diff
@@ -300,10 +301,32 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
...
+		if("Yes")
+			to_chat(src, "<span class='notice'>Choice registered: Yes.</span>")
+			reenter_corpse()
+			return
+		if("No")
+			to_chat(src, "<span class='danger'>Choice registered: No.</span>")
+			return
+		if("Not Until Next Time")
+			to_chat(src, "<span class='danger'>Choice registered: No.</span>")
+			to_chat(src, "<span class='notice'>You will no longer receive notifications for revival attempts until next time.</span>")
```
SteelSlayer, line 325:
```suggestion
			to_chat(src, "<span class='notice'>You will no longer receive notifications for revival attempts this round.</span>")
```

code/modules/mob/dead/observer/observer.dm
Bizzonium: But the variable for this answer option is live until ghost soul is returned into the mob. I'm not that fluent in english so I kept it simple "Not Until Next Time"

SteelSlayer: Oh ok, I see what its suppose to do now. Still, "Not until next time" is a bit confusing on what it means. But, I don't have a good suggestion for a replacement.

AffectedArc07:
Popups drive me insane on here to the point where I call the game broken when "Not for this round" doesnt work on half the offered roles.

Big no from me.

variableundefined:
Objection on the same ground as above. Two maints + 1 head objection, closing.

PR #14075, opened 2020-08-13
[TGUI] Vending Machines

Kyep:
## What Does This PR Do
Converts /obj/machinery/vending and its subtypes to TGUI.
This covers stuff like:
- Medical: NanoMed Plus
- Civilian: Robust Softdrinks, Mr. Chang, Getmore Chocolate Corp, Hot Drinks Machine, Booze-O-Mat, Autodrobe, PTech
- Engineering: YouTool, Engi-Vend
- Science: Toximate 3000
- Security: SecTech
- Service: MegaSeed Servitor, NutriMax, Dinnerware Vendor, 
Basically anything you'd consider a "vendor" that gives out items either for free, or for credits. Including off-station versions.

This does not cover machines which give out items based on some non-credit currency, such as:
- Mining's "mining equipment vendor", which is not actually a subtype of vendor at all

## How Do I Buy Stuff?
1, In general: walk up to the vendor, click it, and click the button next to the item name. It'll dispense the item straight into your hand, if possible. Unlike current vendors, there is no second "pay" screen you must click through each time. There is, however, a vendor sound effect ported from TG.
2. Free items are just dispensed, no questions asked. This applies to basically every vendor/item off-station (where there are no bank accounts, and typically no money of any kind).
3. Items that require a coin require you to put the coin into the vendor first. The item will be displayed as "COIN" and grayed out until you do that. If you change your mind, you can remove the coin before it is used. Unlike our current vendors, you can actually see what items inserting a coin will unlock, before you put in the coin. No more guesswork.
4. Items that have a price require money. It'll automatically figure out who you are and your account info, based on your ID, and charge you. Note that it considers the ID in your ID slot LAST, so if you have, say, stolen an ID and held it in your hand, or put another ID in your PDA, it will try to charge those IDs first, instead of your main ID. You can also pay for stuff by holding enough cash in your active hand to pay for the item.

## Images of changes
Example of what you see at a relatively normal vendor on station:
VG6KlvMQTi
When the machine sells items for credits, and you have no ID (in ID card slot, PDA, or either hand):
vendor_no_id_at_all
When the machine sells items for credits, and your ID has no account number (captain's spare, badly made replacement ID, syndie staff on lavabase, etc):
vendor_blank_id
When the machine sells items for credits, and you are holding cash in your hand:
vendor_cash_transaction
Here's a screenshot showing every section:
TaTseGXfUx

Note:
- You'd never see all of these options at once in practice, I'm just doing a combined screenshot for ease of documentation.
- The "Coin" section only appears when a coin is inserted, which you can only do if the vendor has extra coin-only items. The function of the section is to tell you a coin is inserted, and let you eject it.
- The "Item" section only appears when a vendor-specific item has been inserted. Its function is similar to the Coin section for those vendors. Its primary use is to let you take a coffee cup, put it back into the coffee vendor, and have the vendor fill it with coffee/soup for you.
- The "Maintenance" section only appears when the panel is screwdrivered, and its function is to let you toggle the speaker of the vendor on/off.
- The "User" section at the top only appears for on-station vendors (the kind that take credits).

## Changelog
:cl: Kyep
tweak: Converted vending machines to TGUI.
/:cl:


TDSSS:
Part of me wishes we had a symbol to use for credit.  All good sci-fi currencies have one.

AffectedArc07:
image
Having a section header with just a button in the header and nothing below it seems extremely off in my opinion.

Kyep:
Updated:

1. Now supports icons.
2. Now has better UI for Maintenance section.

PR #14077, opened 2020-08-14
Wizmime Mask fix

### LINKED ISSUE #14076: Magical mime mask doesnt hide your face
**Issue Description**:

The magical mime mask doesn't hide your face like a mask should, so you can see who it really is under it, dispite having your face covered sprite wise and logicaly

For the mime  mask to cover your face and make you look unknown

**What happened instead**:

Mask doesn't stop your identity from showing.

Allows for easy tracking to happen, inconsistent from other masks

Be a wizard.
Get a magical mime mask.
Put it on.

image

ExusA:
## What Does This PR Do
Edits the wizard mime mask, making it hide your identity like the clown version.

## Why It's Good For The Game
Fixes #14076 

## Changelog
:cl:
fix: the wizard mime mask is now functionally identical to its clown counterpart
/:cl:

### DIFF
--- a/code/modules/clothing/masks/gasmask.dm
+++ b/code/modules/clothing/masks/gasmask.dm
@@ -155,7 +155,8 @@
 /obj/item/clothing/mask/gas/mime/wizard
 	name = "magical mime mask"
 	desc = "A mime mask glowing with power. Its eyes gaze deep into your soul."
-	flags_inv = HIDEEARS | HIDEEYES
+	flags = BLOCK_GAS_SMOKE_EFFECT | AIRTIGHT
+	flags_inv = HIDEFACE | HIDEEARS | HIDEEYES
 	magical = TRUE
 
 /obj/item/clothing/mask/gas/mime/nodrop


code/modules/clothing/masks/gasmask.dm
```diff
@@ -155,7 +155,8 @@
 /obj/item/clothing/mask/gas/mime/wizard
 	name = "magical mime mask"
 	desc = "A mime mask glowing with power. Its eyes gaze deep into your soul."
-	flags_inv = HIDEEARS | HIDEEYES
+	flags = BLOCK_GAS_SMOKE_EFFECT | AIRTIGHT
+	flags_inv = HIDEFACE | HIDEEARS | HIDEEYES
```
Kyep, line 159: Here

datlo:
The clown and mime wizard masks intentionally do not hide your identity, as wizards do not have IDs and would always show up as unknown otherwise. This is because the only reason to wear those masks is for the sake of a gimmick, and not having your gimmick name shown is a pain.

ExusA:
after reading both items again, neither block view of your face and I suppose neither _should_ for the reasons Datlo gave. I'd maybe like to see a description change to those items to handwave away exactly WHY you can see someone's identity while wearing them, but not with the normal clown/mime masks, but since I can't think of one that makes sense, going to close this.

PR #14078, opened 2020-08-14
[TGUI] Fax Machines

Kyep:
## What Does This PR Do
Converts fax machines to TGUI.

## Images of changes
Default state (only available option is insert ID):
faxmachine_default
Ready-to-send state (ID in, logged in, paper inserted, destination chosen):
faxmachine_readytosend
Post-send state (ID removed, sending on cooldown, left logged-in for convenience):
faxmachine_postsend

## Changelog
:cl: Kyep
tweak: Converted fax machines to TGUI.
/:cl:


code/modules/paperwork/faxmachine.dm
```diff
@@ -58,17 +66,16 @@ GLOBAL_LIST_EMPTY(fax_blacklist)
...
 /obj/machinery/photocopier/faxmachine/attack_ghost(mob/user)
-	ui_interact(user)
+	tgui_interact(user)
 
 /obj/machinery/photocopier/faxmachine/attackby(obj/item/item, mob/user, params)
 	if(istype(item,/obj/item/card/id) && !scan)
 		scan(item)
 	else if(istype(item, /obj/item/paper) || istype(item, /obj/item/photo) || istype(item, /obj/item/paper_bundle))
 		..()
-		SSnanoui.update_uis(src)
```
AffectedArc07, line 71: Put `SStgui.update_uis(src)` here

code/modules/paperwork/faxmachine.dm
```diff
@@ -80,130 +87,141 @@ GLOBAL_LIST_EMPTY(fax_blacklist)
...
+	else if((destination in GLOB.admin_departments) || (destination in GLOB.hidden_admin_departments))
+		var/cooldown_seconds = cooldown_seconds()
+		if(cooldown_seconds)
+			.["sendError"] = "Re-aligning in [cooldown_seconds] seconds..."
+
+
+/obj/machinery/photocopier/faxmachine/tgui_act(action, params)
+	. = ..()
+	if(.)
+		return
```
AffectedArc07, line 134:
```suggestion
	if(..())
		return
```
The other approach would return TRUE if someone had exploited the UI, which would update the UI even though it wasnt a valid action.

code/modules/paperwork/faxmachine.dm
```diff
@@ -226,7 +244,6 @@ GLOBAL_LIST_EMPTY(fax_blacklist)
 			usr.drop_item()
 			card.forceMove(src)
 			scan = card
-	SSnanoui.update_uis(src)
```
AffectedArc07, line 229: Same here, add it back so the UI still speed-updates when you put a card in

code/modules/paperwork/faxmachine.dm
```diff
@@ -237,25 +254,20 @@ GLOBAL_LIST_EMPTY(fax_blacklist)
...
 		scan.forceMove(get_turf(src))
 		if(!usr.get_active_hand() && Adjacent(usr))
 			usr.put_in_hands(scan)
 		scan = null
 	else
-		to_chat(usr, "There is nothing to remove from \the [src].")
+		to_chat(usr, "There is nothing to remove from [src].")
 
 /obj/machinery/photocopier/faxmachine/proc/sendfax(var/destination,var/mob/sender)
-	if(stat & (BROKEN|NOPOWER))
```
AffectedArc07, line 249: I assume this was removed because you checked it at the `tgui_act` level?

code/modules/paperwork/faxmachine.dm
```diff
@@ -291,19 +303,13 @@ GLOBAL_LIST_EMPTY(fax_blacklist)
...
 	else
-		return 0
+		return FALSE
 
 	use_power(active_power_usage)
-	return 1
+	return TRUE
 
 /obj/machinery/photocopier/faxmachine/proc/send_admin_fax(var/mob/sender, var/destination)
-	if(stat & (BROKEN|NOPOWER))
```
AffectedArc07, line 300: Same here

tgui/packages/tgui/interfaces/FaxMachine.js
```diff
@@ -0,0 +1,68 @@
...
+                onClick={() => act("auth")} />
+            </LabeledList.Item>
+          </LabeledList>
+        </Section>
+        <Section title="Fax Menu">
+          <LabeledList>
+            <LabeledList.Item label="Network">
+              {data.network}
+            </LabeledList.Item>
+            <LabeledList.Item label="Currently Sending">
```
AffectedArc07, line 35: Rename this to something like `Document` or `Item`. `Currently Sending` looks like the header for what the fax machine is literally already sending at that moment in time (IE: Its in the process of sending)

code/modules/paperwork/faxmachine.dm
Kyep: Yes

code/modules/paperwork/faxmachine.dm
Kyep: Same

code/modules/paperwork/faxmachine.dm
```diff
@@ -226,7 +244,7 @@ GLOBAL_LIST_EMPTY(fax_blacklist)
 			usr.drop_item()
 			card.forceMove(src)
 			scan = card
-	SSnanoui.update_uis(src)
+		SStgui.update_uis(src)
```
AffectedArc07, line 247: You accidentally double-indented this, when its meant to be single-indented (See old diff and spacing)

PR #14080, opened 2020-08-14
Playtime DB update call refactor

Kyep:
## What Does This PR Do

- Refactors /proc/update_exp to remove spawn() and sleep(), and instead have a CHECK_TICK. Goal is to ensure it runs as fast as it reasonably can without lagging the server.
- Changes /client/proc/update_exp_client to abort instantly if there's no DB connection. And generally avoid causing issues if it runs past DB connection shutdown time.

## Why It's Good For The Game
- The sleep(10) in /proc/update_exp meant that with 100+ players, it could run for 100+ seconds... which meant it could still be running when DB connection is shut down.... which meant it could runtime/error after that. Fixing that requires that update_exp run quickly.
- ... but doing that might lag the server. So the CHECK_TICK is there to prevent it lagging the server (still requires TM to verify)
- Just in case it still survives to roundend (or we're unlucky) there's also !GLOB.dbcon.IsConnected() to abort it if needed.

## Changelog
:cl:
fix: Fixed playtime tracking system sometimes failing to complete its run before round-end.
/:cl:

### DIFF
--- a/code/controllers/subsystem/jobs.dm
+++ b/code/controllers/subsystem/jobs.dm
@@ -27,7 +27,7 @@ SUBSYSTEM_DEF(jobs)
 /datum/controller/subsystem/jobs/fire()
 	if(!config.sql_enabled || !config.use_exp_tracking)
 		return
-	update_exp(5,0)
+	INVOKE_ASYNC(update_exp(5, 0))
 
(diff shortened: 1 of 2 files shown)


Kyep:
This was TM'ed for a round with no issues noticed. Playtime did appear to update correctly.

PR #14082, opened 2020-08-15
Fixes destroyer borg lawsync, laws, sound

Kyep:
## What Does This PR Do
Fixes a few bugs with the admin-only destroyer borgs, such as them:
- syncing to an AI, when they're supposed to be an event borg
- getting a random lawset, instead of the blank or DS-borg lawset they're meant to have
- not playing a sound on creation

## Changelog
:cl: Kyep
fix: fixed a few bugs with admin-only destroyer borgs, such as them linking to an AI (and thus generating a notice to the station AI when spawned in admin test area).
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -1437,7 +1437,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 	default_cell_type = /obj/item/stock_parts/cell/bluespace
 
 /mob/living/silicon/robot/destroyer/init(alien = FALSE, mob/living/silicon/ai/ai_to_sync_to = null)
-	..()
+	laws = new /datum/ai_laws/deathsquad
 	module = new /obj/item/robot_module/destroyer(src)
 	module.add_languages(src)
 	module.add_subsystems_and_actions(src)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1437,7 +1437,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 	default_cell_type = /obj/item/stock_parts/cell/bluespace
 
 /mob/living/silicon/robot/destroyer/init(alien = FALSE, mob/living/silicon/ai/ai_to_sync_to = null)
-	..()
```
AffectedArc07, line 1440: This parent call handles more than just laws and AI syncing, such as adding a camera and the binary channel for law stating. I would re-add those unless its 100% intentional to not have either
```dm
	aiCamera = new/obj/item/camera/siliconcam/robot_camera(src)
	make_laws()
	additional_law_channels["Binary"] = ":b "
	var/found_ai = ai_to_sync_to
	if(!found_ai)
		found_ai = select_active_ai_with_fewest_borgs()
	if(found_ai)
		lawupdate = 1
		connect_to_ai(found_ai)
	else
		lawupdate = 0

	playsound(loc, 'sound/voice/liveagain.ogg', 75, 1)
```

Kyep: Originally it was intentional to not have either.
Having slept on it though, I've decided to add them. I don't expect either will be used, and I'm not even sure the camera is a good thing, but for the sake of keeping changes minimal, they're now added.

PR #14086, opened 2020-08-15
Reworks Military Shuttle (Admin only)

AffectedArc07:
## What Does This PR Do
This PR replaces the old military shuttle with a new one. It has:
- Ample seating area
- Turrets
- A heavy-duty aesthetic
- Shield generators
- A flasher outside the brig section
- Donuts

Here are some screenshots in no particular order. All of these were taken without nightvision or xray
Full seating area:
image

Brig area:
image

Medical:
image

Cockpit:
image

Shields up:
image

Default turret settings:
image

## Why It's Good For The Game
Could be nice for the admins to use a heavy duty military-esque escape shuttle on rounds where a lot of the crew have been cultified or if there is an overwhelming threat. 

## Changelog
:cl: AffectedArc07
add: Added a new emergency shuttle admin-option, the NDV Brutus Battlecruiser
/:cl:


taukausanake:
Nice. Could use a few cyborg chargers? Maybe on the top-right corner of the seating area

AffectedArc07:
I knew I was missing something. Borg chargers seem appropriate.

AffectedArc07:
Added 2 borg chargers to the seating area
image

TDSSS:
Why do the shields only cover half the window of the bridge?

Also, the ship seems too small and not armed enough to be a battlecruiser, imo. I'd expect something with the name battlecruiser to have BSA batteries and stuff, not just a handful of turrets. I'd recommend renaming this.

AffectedArc07:
@TDSSS The shields cover half because it would get in the way of the consoles, and they also block shots for long enough for command to evac. As for the name, I wanted to stick with lore
image

And there isnt really a way to fit BSA batteries onto a ship.

*Also correct me if I am wrong but a BSA is a stationary cannon which requires a hefty powersource, far more than a cruiser would be able to give out*

TDSSS:
Capture

AffectedArc07:
New design has some more fluff and full protection in the cockpit area. For a full diff, please see MapDiffBot.
image
image

Pyrodes:
me likey when is become merge

code/datums/shuttles.dm
```diff
@@ -55,7 +55,9 @@
 	suffix = "mil"
 	name = "emergency shuttle (military)"
 	description = "Troop transport with point defense turrets."
-	admin_notes = "Designed to ensure a safe evacuation during xeno outbreaks."
+	admin_notes = "Designed to ensure a safe evacuation during xeno outbreaks. Comes with turrets that have the same factions as the station defaults, and no required access. \
```
Kyep, line 58: Maybe also say if they'll shoot aliens/animals/etc.

code/modules/power/lighting.dm
```diff
@@ -749,3 +749,8 @@
 	on = FALSE
 	visible_message("<span class='danger'>[src] flickers and falls dark.</span>")
 	update(0)
+
+// TODO: Add creation methods for these
+/obj/machinery/light/floor
```
Kyep, line 754: Curious what these are for. Also that TODO.

AffectedArc07: They are for the in-floor lights
image

I can do the todo in this PR but it makes it bigger than it already is, and the lights are already an admin only thing. *Insert shadowling balance concern here too*

AffectedArc07:
Current issues
- turrets at the front shoot each other
- "check sec records" makes them shoot ERT (who have no records)
- shields get left behind when shuttle moves

AffectedArc07:
If an admin wants this template, they can have it from the PR.

For now though, I want to get my own backlog down.

MarsM0nd:
Closed it right when the old military shuttle was used. Also, would have been good to know "check sec records" makes them shoot ERT (who have no records) at that time, happens with the old one, too.

PR #14087, opened 2020-08-15
Nerfs butterfly crate

AffectedArc07:
## What Does This PR Do
Reduces the amount of butterflies in the butterfly crate to 1. Why? Because being able to order packs of 50 butterflies en mass (which obliterates the NPC subsystem) is quite frankly bad design, and shouldn't be a thing.

## Why It's Good For The Game
See above

## Changelog
:cl: AffectedArc07
tweak: Butterfly crate is now non-contraband
tweak: Butterfly crate contains 1 butterfly, instead of 50
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -1173,11 +1173,10 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	containername = "fox crate"
 
 /datum/supply_packs/organic/butterfly
-	name = "Butterflies Crate"
+	name = "Butterfly Crate"
 	cost = 50
 	containertype = /obj/structure/closet/critter/butterfly
-	containername = "butterflies crate"
-	contraband = 1
+	containername = "butterfly crate"
 
 /datum/supply_packs/organic/deer
 	name = "Deer Crate"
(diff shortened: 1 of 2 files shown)


Mitchs98:
Why not just lower the amount of butterflies you get in the crate to 5 or so? See no reason personally to just delete the crate entirely. They don't exactly get ordered every single shift anyways.

Qwertytoforty:
I'm fine with the crate being fully removed, based on the lag, not to mention it is never ordered except to mess with people. If people want butterflies, there are many ways to spawn simple animals as decoration.

Landerlow:
Personally I'd say good riddance. all it does is lag the server with butterflies. They get annoying real fast, too. I'd be fine with keeping the crate and changing the amount of butterflies to 5 IF butterflies actually served a purpose. Which they don't, AFAIK.

SteelSlayer:
@Landerlow You can say that about most of the animal crates though. Other than to get DNA for the DNA vault or just to mess around with as a pet or whatever, they don't serve a purpose.

Denthamos:
Kind of gotta agree with Mitch on this one.  Just removing it seems a bit much when you can just lower the amount of the mobs in the crate to be in line with the other animal crates.  It _still_ does have a purpose for the DNA vault, so nuking it feels a bit knee jerk there.

AffectedArc07:
PR has been modified so the crate still exists, but its only 1 butterfly now, so you can still use it for the DNA vault. Every other animal crate has 1 in, so this is what its at. The crate is also non-contraband now.

@Mitchs98 @Denthamos

PR #13292, opened 2020-04-13
Revert "Removes Xenomorphs from event rotation"

Kyep:
## What Does This PR Do
Reverts ParadiseSS13/Paradise#13104
Re-adds xenos to major event rotation.

## Why revert this now, when you approved it yesterday
Since yesterday, I've been able to talk to more people about it, such as the staff of CM.
With this new information, I no longer think that removing xenos from rotation is necessary.

I still have reservations about their un-fun stun-based gameplay, but the removal wasn't for that, it was for other reasons that no longer apply. So it is only fair to re-enable them.

## Changelog
:cl: Kyep
tweak: Re-adds xeno event to major event rotation.
/:cl:

PR #14092, opened 2020-08-16
Revert "Disables dragon transformation because its users can freeze the whole server"

farie82:
Reverts ParadiseSS13/Paradise#14037

#13247 is fixed which fixes the actual issue behind the server killing lag

### DIFF
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -120,13 +120,18 @@
 		return
 
 	var/mob/living/carbon/human/H = user
-	var/random = rand(1,2)
+	var/random = rand(1,3)
 
 	switch(random)
 		if(1)
 			to_chat(user, "<span class='danger'>Your flesh begins to melt! Miraculously, you seem fine otherwise.</span>")
 			H.set_species(/datum/species/skeleton)
 		if(2)
+			to_chat(user, "<span class='danger'>Power courses through you! You can now shift your form at will.")
+			if(user.mind)
+				var/obj/effect/proc_holder/spell/targeted/shapeshift/dragon/D = new
+				user.mind.AddSpell(D)
+		if(3)
 			to_chat(user, "<span class='danger'>You feel like you could walk straight through lava now.</span>")
 			H.weather_immunities |= "lava"
 


AffectedArc07:
Please test this instead of just assuming

farie82:
> Please test this instead of just assuming

It is a clean revert of the PR. Nothing else changed surrounding the code and only adds the same working logic back. The issue was unreproducable on local servers but was producable on live. The copying of the list cost the most when it was reproduced since it used the timSort proc which peaked at 200 CPU cycles in one tick.

AffectedArc07:
Make a macro to generate 4000 say logs for a working example. I am not prepared to merge in something which could potentially crash the server again.

Fox-McCloud:
Once this is tested this can go in--the first PR was speed and emergency merged because of its server-crashing and server-crushing potential. That said, please, please test this thoroughly so we can be sure that this doesn't do what it used to do.

There was no discussion or anything else revolving around balance/design/whatever else have you about this spell.

Keeping this out when it was merged in under completely different pretenses is unneeded politiking.

farie82:
Using 
```dm
/mob/proc/test()
	var/turf/T =  get_turf(src)
	LAZYINITLIST(logs[ATTACK_LOG])
	var/list/log_list = logs[ATTACK_LOG]
	for(var/i in 1 to 1000000)
		var/datum/log_record/record = new(ATTACK_LOG, src, "test[i]", src, T, world.time+i)
		log_list.Add(record)
```

I managed to fully freeze my client. I wanted to profile it but it took more than 5 minutes already to process.
image

Using the new master and this code:
```dm
/mob/proc/test()
	var/turf/T =  get_turf(src)
	for(var/i in 1 to 1000000)
		var/datum/log_record/record = new(ATTACK_LOG, src, "test[i]", src, T, world.time + i)
		GLOB.logging.add_log(ckey, record)
```
It was instant
image
Which used `sortTim()` before which caused the big lag

Kyep:
@Fox-McCloud 
When I did #13292 to re-add xenos to rotation, the other two heads decided that they would not support it. Xenos had only been out of rotation for about a day, IIRC, and the original reason for their removal had been addressed. Despite all this, the other two heads still opted to change their minds and not support the revert. This has stuck, with xenos being out of rotation ever since. I felt a bit annoyed with this, with people changing their minds over what appeared to be *less than a day*, but I accepted it. I don't recall labeling it "unneeded politiking" or otherwise questioning its legitimacy. I may have been annoyed. Even frustrated. But I accepted it as a legitimate thing for them to do.

That precedent is pretty clear on this. Heads are NOT obliged to re-add something just because the original reason for its removal has been addressed. Even if it was only removed a day ago! 

And... I don't think we really want to change this precedent. Imagine if heads and/or maints felt obliged to not change their mind about stuff. Even after they'd spent more time looking into it and thinking about it. Even after they've realized that there are other problems with the thing and they just don't want it re-added right now. Imagine if, in that situation, they still feel obliged to support re-adding it even though they're convinced its not a good idea. That would suck, and be super-dysfunctional.

PR #14094, opened 2020-08-16
AoE spell attack logging tweak

### LINKED ISSUE #14093: Turfs hit by ghosts boo are logged to attack laws
**Issue Description**:

Turfs hit by ghosts boo are logged to attack laws, leading to 9 attack logs from one spell

For boo to not be logged, or logged once.

**What happened instead**:

9 attack logs for the price of one
image

Clogs up attack logs a lot for not good reason

Be a ghost.
Boo.
Look at attack logs.

https://github.com/ParadiseSS13/Paradise/pull/13247

SteelSlayer:
## What Does This PR Do
Limits the attack logs that `/aoe_turf` type spells generate. Some of these can generate an absurd amount of attack logs because they take a screen's worth of turfs and make a log on each of them, which can mean hundreds of logs. They will now only generate 1 attack log, which says that the user has has cast the spell.

## Why It's Good For The Game
Less attack log spam for admins. Fixes #14093

## Changelog
:cl:
tweak: /aoe_turf type spells, such as Boo!, Knock, and Repulse, will now only generate 1 attack log, instead of a log for every turf they loop through while searching for targets
/:cl:

### DIFF
--- a/code/datums/spell.dm
+++ b/code/datums/spell.dm
@@ -444,6 +444,11 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 
 	return
 
+// Normally, AoE spells will generate an attack log for every turf they loop over, while searching for targets.
+// With this override, all /aoe_turf type spells will only generate 1 log, saying that the user has cast the spell.
+/obj/effect/proc_holder/spell/aoe_turf/perform(list/targets, recharge, mob/user, make_attack_logs)
+	add_attack_logs(user, null, "Cast the AoE spell [name]", ATKLOG_ALL)
+	return ..(targets, recharge, user, FALSE)
 
 /obj/effect/proc_holder/spell/targeted/proc/los_check(mob/A,mob/B)
 	//Checks for obstacles from A to B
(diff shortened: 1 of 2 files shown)


Qwertytoforty:
Thank you

code/datums/spell.dm
```diff
@@ -444,6 +444,11 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 
 	return
 
+// Normally, AoE spells will generate an attack log for every turf they loop over, while searching for targets.
+// With this override, all /aoe_turf type spells will only generate 1 log, saying that the user has cast the spell.
+/obj/effect/proc_holder/spell/aoe_turf/perform(list/targets, recharge, mob/user, make_attack_logs)
+	add_attack_logs(user, user, "cast the spell [name]", ATKLOG_ALL)
```
farie82, line 450: Can also remove the target here. So just null as second param. It now shows up as if it used the spell on himself.

Maybe some more description about it being an AOE spell in the `what` could also be nice

PR #14099, opened 2020-08-17
Replaces skills computer with skills HUD glasses

Kyep:
## What Does This PR Do
- Deletes skills computers (/obj/machinery/computer/skills) from the game. This includes their placement from maps (cyberiad, delta, meta), their object in the game, and their NanoUI template, skills_data.tmpl
- Replaces them with the new "Skills HUD", which is a pair of glasses that look like mesons, but are blue, and let you see employment data (the fluff text that people enter into their character history as employment records) when you examine crew. Similar to how you can use a sec hud to see someone's sec status when examining them. Unlike the sec hud though, skills hud is always read-only.
- The Captain, HOP, NTR and Blueshield (who currently have access to skills computers) now instead get skills HUD glasses by default. Since they're replacing sunglasses for these jobs, the version that these jobs get has the same protective abilities as their old sunglasses had.
- Science can also print skills HUDs with the same tech/cost requirements as sec huds. But unlike the versions issued to command, these don't have flash protection. They're not sunglasses.
- Ghosts also see a character's skills/employment data when examining them. Assuming they have any saved. And assuming the ghost has HUDs enabled.

## Why It's Good For The Game
- Skills computers are terrible, for several reasons. First, they're NanoUI. Second, they're the worst kind of NanoUI - copypaste from the Sec Records computer. Third, they're INCREDIBLY complicated for what they do (provide access to skills data - all their other functions being duplicated by the sec records computer). Fourth, they're clunky and unnecessarily difficult to use, being found in a few places and not easily portable. Although they look like laptops, they're actually NOT laptops. They're not even modular computers. They're a full computer that (for some reason) looks like a laptop, but doesn't share any laptop functions. You can't fold them up and take them with you, for example, which is the primary reason for laptops existing. Yet they still look like laptops. To summarize by simply quoting how skills computers are defined in the code right now: "/obj/machinery/computer/skills//TODO:SANITY"
- Conversely, the new HUD glasses idea solve all of these issues. No more NanoUI. No more copy-pasted code or templates from sec computer. Very simple - does only one thing - show skills data. Actually does its job well, showing you skills data on all crew you examine (so long as they haven't hidden their identity). Requires almost no effort to use. You don't even have to go to a terminal, just examine people wherever you happen to be. All this is vastly simpler, easier, and less hassle.
- With the addition of a non-sunglasses variant printable in R&D, it also opens up skills data to non-command staff. IE: you might actually get your character's employment history with NT brought up as a topic by the people you encounter in-game now. Your character's backstory might actually come up in conversation!
- Ultimately, the point of all this is to both axe a complicated NanoUI template in an expedient way, and also to, at the same time, vastly improve the usefulness of skills data, by making it more accessible, especially to the Captain/HOP/NTR who are encouraged to use it for RPing purposes.
## Images of changes
skills_examine
skills_humans
## Changelog
:cl: Kyep
del: Removed skills computers.
add: Added skills HUD, blue glasses that let you see other crew members' NT employment history. Cap/HOP/NTR/BS start with these. Lesser variants can be printed at R&D.
/:cl:


code/modules/clothing/glasses/hud.dm
```diff
@@ -196,3 +196,17 @@
...
+	icon_state = "material"
+	item_state = "glasses"
+
+/obj/item/clothing/glasses/hud/skills/sunglasses
+	name = "Skills HUD Sunglasses"
+	desc = "Sunglasses with a build-in skills HUD, showing the employment history of nearby NT crew members."
+	see_in_dark = 1
+	flash_protect = 1
+	tint = 1
+	prescription_upgradable = 1
```
AffectedArc07, line 212: Shouldnt these be `TRUE` here?

Mitchs98:
So does the Blueshield no longer get their health HUD glasses? If so, why do they get the skills HUD(Which is, honestly, useless for them?) instead? Or do they just get them optionally?

Kyep:
@Mitchs98 The BS still gets their health HUD sunglasses. They just also get a pair of skills HUD sunglasses in their locker as well.

AffectedArc07:
Note to self if this is merged: Remove from TGUI tickoff list.

Mitchs98:
Ah alright, I see. Sounds good then just wanted to make sure!

Carthusia:
Is it possible to make a craftable variant of the sunglasses one, similar to how you can craft sec sunglasses and medical ones so long as you get the parts to construct it.

Currently from what i can remember, sunglasses variants of the hud items just require the hud type from RND, a pair of real sunglasses, and maybe one or two other minor components.

Kyep:
@Carthusia In theory yes, that is doable, but I have no plans to do it in this PR. Since the Captain/HOP/NTR/etc already get sunglasses versions of the skills HUD, the only people that feature would be useful for are people printing the non-sunglasses version in R&D, who somehow also obtain a separate pair of sunglasses. That's a pretty niche situation.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -364,6 +364,17 @@
...
+	if(hasHUD(user, "skills"))
+		var/perpname = get_visible_name(TRUE)
+		var/skills
+
+		if(perpname)
+			for(var/datum/data/record/E in GLOB.data_core.general)
+				if(E.fields["name"] == perpname)
+					skills = E.fields["notes"]
+			if(skills)
+				msg += "<span class = 'deptradio'>Employment records:</span> [skills]\n"
```
AffectedArc07, line 376:
```suggestion
				msg += "<span class='deptradio'>Employment records:</span> [skills]\n"
```

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #14108, opened 2020-08-18
Reclassifies most clothing as Small

ExusA:
## What Does This PR Do
Changes the default weight class for clothing items to SMALL instead of NORMAL.
In effect, this means:
1. You'll be able to fit as many jumpsuits and similar items in your bag as you can tools.
2. You'll be able to pick up and dispose of most loose clothing items using a trash bag.
3. Hardsuits, spacesuits, armor vests, and most anything with a particularly useful mechanical function keeps its old weight class. No carrying around hardsuits in your backpack!

## Why It's Good For The Game
1. Clothes in general are mostly cosmetic items, but their NORMAL weight class means you can't keep as many on you as you can functional items like tools. Being able to carry clothes along with your other items is a general QoL, and doesn't affect the balance much in other ways.
2. Clothes often get strewn around the station, but janitors and janitor cyborgs don't have a good way of picking them up en-masse. This PR allows them to fit most clothes in a trashbag for easy transport/disposal.
3. Realistically, you can fold most types of clothes down into a very small space. It doesn't make much sense that you CANNOT fit some shorts in a trashbag no matter how you fold them or anything.

## Changelog
:cl:
tweak: Made most clothes smaller.
/:cl:

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -32,6 +32,7 @@
 	var/cooldown = 0
 	var/species_disguise = null
 	var/magical = FALSE
+	w_class = WEIGHT_CLASS_SMALL
 
 /obj/item/clothing/proc/weldingvisortoggle(mob/user) //proc to toggle welding visors on helmets, masks, goggles, etc.
 	if(!can_use(user))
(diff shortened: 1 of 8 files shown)


TDSSS:
Does this make labcoats small? They have two pockets to put stuff in, you could in theory stuff a lot of items in a trashbag that way.

ExusA:
Labcoats are small still, because they can store two TINY sized items. TINY takes up 1 unit worth of space, and SMALL makes up 2 units of space, so storing a bunch of labcoats in a bag won't let you store any more items, other than the labcoats themselves. Not to mention how cumbersome it would be to carry items that way...

AffectedArc07 (approved):

PR #14114, opened 2020-08-19
Clown lavaland ruin loot change and Honkputer

ExusA:
## What Does This PR Do
1. Replaces the Staff of Slipping obtainable from the loot room in the clown biodome ruin with a HONKputer computer board.
2. Fixes a few issues with the HONKputer, including deconstruction path, fixing the way the login works, and making it slightly clearer that a bananium frame is required when examining the board.

## Why It's Good For The Game
1. The staff of slipping is supposed to be wizard-exclusive for a good reason: it's an instant, ranged, self-recharging stun weapon. It's _designed_ to be unfair, it's a wizard item. Miners, miner antags and even security when they get their hands on it, shouldn't really have access to the staff on normal rounds.
2. The circuit board for the HONKputer is unobtainable currently, this PR makes it obtainable. As a refresher, the HONKputer is like a communications console, allowing contact with the clown planet. That is, it PMs admins similar to how the communications console can. Keep in mind, this will NOT result in spam to admins, because the board can only be placed in a bananium computer frame, which takes FIFTY bananium sheets to assemble, and that's very unlikely to happen in a normal round. The computer is also locked to clown-id only, and has a lengthy 10 minute cooldown between messages. If someone does decide to abuse the console, which will happen very rarely, there's always easy options to punish IC with admin smiting. Adding the HONKputer back into the game has potential to add variety on the very few rounds where it's able to be assembled.
3. The honkputer is generally more interesting than the staff, and leads to better scenarios. It could be fun for players to discover the board and try figure out what it does, much more so than just getting easy access to an infinite-ammo wizard's stun-gun.

## Changelog
:cl:
add: Made the mysterious HONKputer obtainable.
fix: Made the HONKputer function properly again.
del: Removed the Staff of Slipping from the clown biodome ruin.

/:cl:

### DIFF
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -11,6 +11,7 @@
 	integrity_failure = 100
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 40, "acid" = 20)
 	var/obj/item/circuitboard/circuit = null //if circuit==null, computer can't disassembly
+	var/frame_type = /obj/structure/computerframe
 	var/processing = 0
 	var/icon_keyboard = "generic_key"
 	var/icon_screen = "generic"
@@ -92,7 +93,7 @@
 	on_deconstruction()
 	if(!(flags & NODECONSTRUCT))
 		if(circuit) //no circuit, no computer frame
-			var/obj/structure/computerframe/A = new /obj/structure/computerframe(loc)
+			var/obj/structure/computerframe/A = new frame_type(loc)
 			var/obj/item/circuitboard/M = new circuit(A)
 			A.setDir(dir)
 			A.circuit = M
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -11,7 +11,6 @@
 	integrity_failure = 100
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 40, "acid" = 20)
 	var/obj/item/circuitboard/circuit = null //if circuit==null, computer can't disassembly
-	var/frame_type = /obj/structure/computerframe
 	var/processing = 0
 	var/icon_keyboard = "generic_key"
 	var/icon_screen = "generic"
@@ -93,7 +92,7 @@
 	on_deconstruction()
 	if(!(flags & NODECONSTRUCT))
 		if(circuit) //no circuit, no computer frame
-			var/obj/structure/computerframe/A = new frame_type(loc)
+			var/obj/structure/computerframe/A = new /obj/structure/computerframe(loc)
 			var/obj/item/circuitboard/M = new circuit(A)
 			A.setDir(dir)
 			A.circuit = M
--- a/code/game/machinery/computer/honkputer.dm
+++ b/code/game/machinery/computer/honkputer.dm
@@ -1,5 +1,5 @@
 /obj/machinery/computer/HONKputer
-	name = "\improper HONKputer"
+	name = "\improper HONKputer Mark I"
 	desc = "A yellow computer used in case of critically low levels of HONK."
 	icon = 'icons/obj/machines/HONKputer.dmi'
 	icon_state = "honkputer"
@@ -8,7 +8,6 @@
 	light_color = LIGHT_COLOR_PINK
 	req_access = list(ACCESS_CLOWN)
 	circuit = /obj/item/circuitboard/HONKputer
-	frame_type = /obj/structure/computerframe/HONKputer
 	var/authenticated = 0
 	var/message_cooldown = 0
 	var/state = STATE_DEFAULT
@@ -32,15 +31,15 @@
 		// main interface
 		if("main")
 			src.state = STATE_DEFAULT
-
 		if("login")
-			var/mob/living/carbon/human/M = usr
-			var/list/access = M.get_access()
-			if (ACCESS_CLOWN in access)
-				authenticated = 1
-			if(src.emagged==1)
-				authenticated = 1
-
+			var/mob/M = usr
+			var/obj/item/card/id/I = M.get_active_hand()
+			if(istype(I, /obj/item/pda))
+				var/obj/item/pda/pda = I
+				I = pda.id
+			if(I && istype(I))
+				if(src.check_access(I) || src.emagged==1)
+					authenticated = 1
 		if("logout")
 			authenticated = 0
 
--- a/code/game/machinery/computer/honkputer.dm
+++ b/code/game/machinery/computer/honkputer.dm
@@ -37,9 +37,9 @@
 			var/mob/living/carbon/human/M = usr
 			var/list/access = M.get_access()
 			if (ACCESS_CLOWN in access)
-				authenticated = 1
-			if(src.emagged==1)
-				authenticated = 1
+				authenticated = TRUE
+			if(emagged)
+				authenticated = TRUE
 
 		if("logout")
 			authenticated = FALSE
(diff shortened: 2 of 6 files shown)


TDSSS:
TBH, fifty bananium sheets to assemble seems almost impossible. Or is there enough stuff to deconstruct in the biodome to get that?

ExusA:
it pretty much is impossible, and that's sort of the point. the only realistic way to get it is to find the clown spaceship crash space ruin, which has a prebuilt one in it already, plus enough ore for fifty sheets to construct one on-station. I can change it to less if it'd be better like that, but I want it to be very hard to obtain, cause it could get quite annoying for admins

AffectedArc07:
```
loading paradise.mdme
loading interface/skin.dmf
code/game/machinery/computer/honkputer.dm:11:error: frame_type: undefined var
paradise.mdme.dmb - 1 error, 0 warnings (8/19/20 12:28 pm)
```

You actually tested this. Right???

code/game/machinery/computer/honkputer.dm
```diff
@@ -31,22 +32,22 @@
...
+				authenticated = 1
+
 		if("logout")
 			authenticated = 0
 
 		if("MessageHonkplanet")
 			if(src.authenticated==1)
 				if(message_cooldown)
-					to_chat(usr, "Arrays recycling.  Please stand by.")
+					to_chat(usr, "<span_class = 'warning'>Arrays recycling.  Please stand by.</span>")
```
AffectedArc07, line 50:
```suggestion
					to_chat(usr, "<span class='warning'>Arrays recycling.  Please stand by.</span>")
```
Spacing. Also the HTML was incorrectly formatted before.

ExusA:
YES! i don't.. even.. know what happened.. AAGH why am I so bad at github

code/game/machinery/computer/honkputer.dm
```diff
@@ -31,22 +32,22 @@
...
-				var/obj/item/pda/pda = I
-				I = pda.id
-			if(I && istype(I))
-				if(src.check_access(I) || src.emagged==1)
-					authenticated = 1
+			var/mob/living/carbon/human/M = usr
+			var/list/access = M.get_access()
+			if (ACCESS_CLOWN in access)
+				authenticated = 1
+			if(src.emagged==1)
```
AffectedArc07, line 41:
```suggestion
			if(emagged)
```
Implied src plus checking a bool

code/game/machinery/computer/honkputer.dm
```diff
@@ -31,22 +32,22 @@
...
-			if(istype(I, /obj/item/pda))
-				var/obj/item/pda/pda = I
-				I = pda.id
-			if(I && istype(I))
-				if(src.check_access(I) || src.emagged==1)
-					authenticated = 1
+			var/mob/living/carbon/human/M = usr
+			var/list/access = M.get_access()
+			if (ACCESS_CLOWN in access)
+				authenticated = 1
```
AffectedArc07, line 40:
```suggestion
				authenticated = TRUE
```
Bool

code/game/machinery/computer/honkputer.dm
```diff
@@ -31,22 +32,22 @@
...
+			if(src.emagged==1)
+				authenticated = 1
```
AffectedArc07, line 42:
```suggestion
				authenticated = TRUE
```
Bool

code/game/machinery/computer/honkputer.dm
```diff
@@ -31,22 +32,22 @@
...
+
 		if("logout")
 			authenticated = 0
```
AffectedArc07, line 45:
```suggestion
			authenticated = FALSE
```
Bool

code/game/machinery/computer/honkputer.dm
```diff
@@ -31,22 +32,22 @@
 		// main interface
 		if("main")
 			src.state = STATE_DEFAULT
```
AffectedArc07, line 34:
```suggestion
			state = STATE_DEFAULT
```
Implied src

code/game/machinery/computer/honkputer.dm
```diff
@@ -8,6 +8,7 @@
 	light_color = LIGHT_COLOR_PINK
 	req_access = list(ACCESS_CLOWN)
 	circuit = /obj/item/circuitboard/HONKputer
+	frame_type = /obj/structure/computerframe/HONKputer
 	var/authenticated = 0
```
AffectedArc07, line 12:
```suggestion
	var/authenticated = FALSE
```
Bool

AffectedArc07 (changes requested):

Cocacolagua:
I am pretty sure the clown shuttle ruin /sometimes/ has a bananium computer frame, but it’s unlikely for the stars to align

ExusA:
looking at the ruin, it seems like the bananium frame is always there. maybe it was changed at some point?

code/game/machinery/computer/computer.dm
```diff
@@ -11,6 +11,7 @@
 	integrity_failure = 100
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 40, "acid" = 20)
 	var/obj/item/circuitboard/circuit = null //if circuit==null, computer can't disassembly
+	var/frame_type = /obj/structure/computerframe
```
AffectedArc07, line 14: Im quite iffy on the idea of adding an entire frame type var to each computer in the game just for the sake of a lavaland ruin.

ExusA: definitely true, but the var could be used in future if any other computers want to have bananium frames, of if new frame types were added. walls have the `girder_type` var which is almost never used as anything other than the default, but is still useful for future

ExusA: that being said, is there a better way to do this?

Mitchs98:
If this gets put in it pretty much makes the ruin pointless, doesn't it? Players wouldn't really pretty much ever be able to experiment with the board and find out what it is or does for the sole reason they'd pretty much never be able to get the bananium required for it. And even if say someone gets fifty bananium, the likelihood of them getting the bananium and getting the board is also quite low. 

I'd argue a subtype of the staff should be made for lavaland, arguing it's a wizard only item isn't really something I'd consider cause for removal. And if not a subtype of the staff, at-least something that's actually interesting/worthwhile to get that people will actually be able to use/enjoy on a semi-regular basis as opposed to something that you'll see be used maybe once every blue moon.

ExusA:
i can see your point there with the cost of the honkputer tbh. I didn't actually change anything about the amount of bananium required, it was fifty already, and i left it at fifty because i thought that just clearing the ruin shouldn't give you access to the honkputer, since it PMs admins and could become pretty annoying. that being said, it also requires clown access. maybe needing a clown ID and clearing the honk ruin is enough obstacles?

I'd argue that most of the stuff you get for clearing the ruin currently is either useless, or exceptionally useful for griff and/or validhunting and/or as an antag. only the last case is valid, and even then using mining loot in antag shenanigans is seen as pretty unfair by ghosts a lot of the time. 

I think just changing the bananium cost to 5 as standard for computer frames would address the issue

ExusA:
the honkputer now costs 15 bananium. There's 10 bananium in the ruin, and five more can be gotten by deconstructing the station clown's statue, so you won't need to find the space ruin anymore, but if you don't have bananium from said ruin you _will_ need cooperation from the clown. this is so you can't go into the ruin, deconstruct the clown statues there and assemble the computer then and there, you'll need to take it to the station first

Kyep (commented):
I'm really :/ on the idea of letting the crew have access to a HONKputer. It is presumably inaccessible for a reason.

ExusA:
Ultimately, it pretty much behaves the same as the pray verb, so it wouldn't be too bad for mins. All of everyone's comments in mind though, i'll close this while i think of something better to replace the staff.

PR #14120, opened 2020-08-20
Cyborg Cryo equity

ExusA:
## What Does This PR Do
Generates a robot endoskeleton when a cyborg cryos. Each body part other than the chest has a chance to be missing proportional to how much health the cyborg had when they cryoed, with a 98% chance at -99 health and a 0% chance at 100 health.

## Why It's Good For The Game
If a ghost joins as a cyborg, is assembled by robotics, and then immediately cryos for whatever reason, an entire stack of metal is completely deleted from the round. This PR changes it so that when a cyborg cryos, it gets "recycled", giving some of the materials back in the form of a new vacant chassis. There's a chance that body parts will be missing depending on the cyborg's health, though: damaged cyborgs can't cryo then rejoin at little cost to the crew, so repairing them properly is still the cheaper option.

## Changelog
:cl:
tweak: Cyborgs now return their chassis when they cryo.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -458,7 +458,7 @@
 			announce.autosay("[occupant.real_name]  ([announce_rank]) [on_store_message]", "[on_store_name]")
 		else
 			announce.autosay("[occupant.real_name] [on_store_message]", "[on_store_name]")
-	visible_message("<span class='notice'>\The [src] hums and hisses as it moves [occupant.real_name] into storage.</span>")
+	visible_message("<span class='notice'>\The [src] hums and hisses as it moves [occupant.real_name] into AI storage and recycles the usable parts of [p_their()] chassis.</span>")
 
 	// Ghost and delete the mob.
 	if(!occupant.get_ghost(1))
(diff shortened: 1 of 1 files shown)


ExusA:
crossing my fingers I did github this right this time...

lolzyking:
I was under the presumption when a borg cryos it opens up a cyborg join slot for latejoins.

taukausanake:
> I was under the presumption when a borg cryos it opens up a cyborg join slot for latejoins.

That is correct, but this is addressing the fact that the resources to make the cyborg are being wasted. Personally I don't see a problem with that, there won't be a shortage of metal or any materials ever in a round so wastage is fine in my book

ExusA:
It's more addressing the fact that the best thing to do if you want to leave the round as a cyborg isn't to cryo: it's to go to robotics and suicide so that the roboticists can scrap you down for parts, and possibly tell them in LOOC that that's what you're doing, which is needlessly clunky. And while there IS a huge amount of metal in any given round, there's not INFINITE metal. The roboticists still have to go and find more somewhere, and if mining hasn't been doing well, deleting an entire stack is a big deal.
On top of that, it makes sense IC that just the robobrain would be stored, not the entire chassis, same as how it works with AI cores. You don't have to assemble a new AI every time the player has to leave, borgs should probably be the same.

PR #14121, opened 2020-08-20
Restores ability to set transfer amount on glasses

AffectedArc07:
## What Does This PR Do
This got removed during #14014 and is apparently important to play chef properly, which I can understand.

## Why It's Good For The Game
Should be able to set transfers on condiments and glasses

## Changelog
:cl: AffectedArc07
fix: You can set transfer amount on glasses and condiment bottles again
/:cl:


Carthusia:
Thank you!

This was so annoying and not quite sure why it was removed, my buckets could pour at 5, but a glass couldn't pour any less than 10

Carthusia:
Regular bottles are also affected, so if i have a bottle of 100u vodka, i cant transfer only 5u anymore

PR #14142, opened 2020-08-24
Gives the recycler a higher damage reflection to avoid it breaking from most trash

farie82:
## What Does This PR Do
Gives the recycler 15 damage reflection instead of 10.
Rods, mobs, toolboxes, tanks, basebal bats and a lot of more common items now don't harm the recycler.

## Why It's Good For The Game
It stops the recycler from breaking by common trash.
Especially rods and other common trash can break it now. Items such as the fire axe or such can still break it

## Changelog
:cl:
fix: The recycler can now withstand most things being thrown at it
/:cl:

### DIFF
--- a/code/game/machinery/recycler.dm
+++ b/code/game/machinery/recycler.dm
@@ -8,7 +8,7 @@
 	layer = MOB_LAYER+1 // Overhead
 	anchored = 1
 	density = 1
-	damage_deflection = 10
+	damage_deflection = 15
 	var/safety_mode = 0 // Temporarily stops machine if it detects a mob
 	var/icon_name = "grinder-o"
 	var/blood = 0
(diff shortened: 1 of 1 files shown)


taukausanake:
Do disposal bins also get damaged when things are thrown at/into them? Just a thought if this is happening to recyclers

TDSSS:
This feels like a very snowflakey way to do this. Couldn't we just increase the damage deflection? Damage deflection applies to thrown objects as well, doesn't it?

farie82:
> Do disposal bins also get damaged when things are thrown at/into them? Just a thought if this is happening to recyclers

Good suggestion. Yet looking at the code...
The disposals do this by moving the object in the CanPass check. if the object is an item and is being thrown then it can go into the disposal bin.... Not the best solution since the proc CanPass should only return a value. Not move the thrown item etc.

Mitchs98:
Do note that floor tiles will still damage and break it. But otherwise seems good.

Kyep (approved):

farie82:
> Do note that floor tiles will still damage and break it. But otherwise seems good.

Their throwforce is 10. If it were 15 then they would do damage

AffectedArc07 (approved):

PR #14144, opened 2020-08-25
Prevents blob mobs from being thrown

Qwertytoforty:
## What Does This PR Do

This PR removes the CANPUSH flag most mobs have by default from blob mobs, preventing them from being thrown, or pulled. This prevents them from being cheesed by throwing them into space, although also prevents blobernaughts from pulling blob spores.

## Why It's Good For The Game

Throwing blob spores off the ZLVL makes them useless, but more importantly, still count to a factories max number of spores, preventing them from making one to replace it till it dies. Its also less damaging to you to throw one into deep space, then welder it twice and get hit by spores. Blobbernaughts have received the same change, as the blobbernaught is bigger than, you, and unlike a spore, can not move in space. Cheesing blobbernaughts by throwing them into space and trapping them is not good.

## Changelog
:cl:
tweak: Blob spores and Blobbernaughts can no longer be pulled or thrown
/:cl:


code/game/gamemodes/blob/blobs/blob_mobs.dm
```diff
@@ -7,6 +7,7 @@
 /mob/living/simple_animal/hostile/blob
 	icon = 'icons/mob/blob.dmi'
 	pass_flags = PASSBLOB
+	status_flags = 0 //No throwing blobspores into deep space to despawn, or throwing blobbernaughts, which are bigger than you.
```
farie82, line 10: Can use the define `None` here to be a bit more verbose. It is the same as 0 but it reads a bit better

Qwertytoforty: Done!

farie82 (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #14147, opened 2020-08-25
A couple of cling bodyswap bugfixes

SabreML:
## What Does This PR Do

 1. Fixes a runtime when trying to use the 'Swap Forms' ability on a mob with no mind, by adding a check for that.

 2. If a Changeling swapped forms with someone who committed suicide, they can now commit suicide themselves.
(Not sure why they'd want to, but whatever. 🤷)

 3. You can no longer use 'Swap Forms' if either participant has a Borer.

## Why It's Good For The Game

Runtime bad, suicide good.

## Changelog
:cl:
tweak: Blocked using 'Swap Forms' as a Changeling if you, or the victim has a Cortical Borer.
fix: Fixed runtime when trying to 'Swap Forms' into a mob with no mind.
fix: If a Changeling bodyswaps into someone who suicided, they can now commit suicide themselves.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/swap_form.dm
+++ b/code/game/gamemodes/changeling/powers/swap_form.dm
@@ -18,7 +18,7 @@
 	if((NOCLONE || SKELETON || HUSK) in target.mutations)
 		to_chat(user, "<span class='warning'>DNA of [target] is ruined beyond usability!</span>")
 		return
-	if(!istype(target) || issmall(target) || (NO_DNA in target.dna.species.species_traits))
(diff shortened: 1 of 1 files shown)


Kyep (commented):
Request: please make it impossible to use changeling mindswap if EITHER you (the cling) OR the target has a borer. Because in that scenario, use of mindswap causes bad things to happen. Specifically it can end up with a borer's mind in control of a normal body without being a borer. And other bugs.

This isn't a blocking request (IE: PR can still be approved, even if you don't do this). Its just an idea for an additional fix that would be helpful, seeing as you're already touching cling mindswap code.

SabreML:
Seems simple enough, give me a second.

PR #14148, opened 2020-08-26
Combines Changeling 'Channel DNA' and 'Absorb DNA' into one button

SabreML:
## What Does This PR Do

Combines the buttons for `hivemind_upload` and `hivemind_download` into one, `hivemind_pick`. Using the icon for `Hive Absorb DNA`.

Selection

## Why It's Good For The Game

Changeling actions can get _very_ cluttered, to the point of taking up almost a quarter of the screen when combined with other items.
While this only removes one button, it's definitely a step in the right direction.

## Images of changes

Old:
Old

New:
New

## Changelog
:cl:
tweak: Combined the buttons for the Changeling 'Channel DNA' and 'Absorb DNA' abilities into one, to reduce screen clutter.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/hivemind.dm
+++ b/code/game/gamemodes/changeling/powers/hivemind.dm
@@ -38,9 +38,8 @@ GLOBAL_LIST_EMPTY(hivemind_bank)
 		if(changeling.using_stale_dna(user))//If our current DNA is the stalest, we gotta ditch it.
 			to_chat(user, "<span class='warning'>We have reached our capacity to store genetic information! We must transform before absorbing more.</span>")
 			return
-		dna_absorb(user)
-	else
-		return
+		else
+			dna_absorb(user)
 
 /datum/action/changeling/proc/dna_upload(mob/user)
 	var/datum/changeling/changeling = user.mind.changeling
--- a/code/game/gamemodes/changeling/powers/hivemind.dm
+++ b/code/game/gamemodes/changeling/powers/hivemind.dm
@@ -12,27 +12,37 @@
 	var/datum/changeling/changeling=user.mind.changeling
 	changeling.changeling_speak = 1
 	to_chat(user, "<i><font color=#800080>Use say \":g message\" to communicate with the other changelings.</font></i>")
-	var/datum/action/changeling/hivemind_upload/S1 = new
+	var/datum/action/changeling/hivemind_pick/S1 = new
 	if(!changeling.has_sting(S1))
 		changeling.purchasedpowers+=S1
 		S1.Grant(user)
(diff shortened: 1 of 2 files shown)


Tokorizo:
a welcomed addition clings usually have two rows of buttons, anything to reduce it is always nice.

Muhdi345:
Why do we even have those obnoxiously large buttons for things you only use once or twice. The changeling tab was great for those.

code/game/gamemodes/changeling/powers/hivemind.dm
```diff
@@ -12,27 +12,37 @@
...
+
+	if(channel_pick == "Upload")
+		dna_upload(user)
+	if(channel_pick == "Absorb")
+		if(changeling.using_stale_dna(user))//If our current DNA is the stalest, we gotta ditch it.
+			to_chat(user, "<span class='warning'>We have reached our capacity to store genetic information! We must transform before absorbing more.</span>")
+			return
+		dna_absorb(user)
+	else
+		return
```
farie82, line 43: Non required return

code/game/gamemodes/changeling/powers/hivemind.dm
```diff
@@ -12,27 +12,37 @@
...
+	var/datum/changeling/changeling = user.mind.changeling
+	var/channel_pick = alert("Upload or Absorb DNA?", "Channel Select", "Upload", "Absorb")
+
+	if(channel_pick == "Upload")
+		dna_upload(user)
+	if(channel_pick == "Absorb")
+		if(changeling.using_stale_dna(user))//If our current DNA is the stalest, we gotta ditch it.
+			to_chat(user, "<span class='warning'>We have reached our capacity to store genetic information! We must transform before absorbing more.</span>")
+			return
+		dna_absorb(user)
```
farie82, line 41: Can put this in the else of the if. Bit more readable

farie82 (changes requested):
Small things

code/game/gamemodes/changeling/powers/hivemind.dm
SabreML: Wasn't sure if that was needed, since it's an input box thing. But I guess it returns either way.

farie82: At the end of code paths in procs it won't be needed. Only exception of doing a return is if you want to override the return value. Say you have this:
```dm
/proc/thingy()
 . = TRUE
 if(thingies)
  ...
  return
 
if(failed)
 return FALSE
```
Not the best of examples :P but still

AffectedArc07 (approved):

PR #13487, opened 2020-05-24
NanoUI Brig Timers

ZomgPonies:
## What Does This PR Do
Switches brig timers to NanoUI
Only people with cell access can view Intake Control or Action sections, including AI and admins
Ghosts or non-authorized personel can only view the top part
Auto-updates and counts down remaining time, shows total time, reason for sentence and prisoner name.

## Why It's Good For The Game
As one of the most used interfaces in-game, it's crazy that this wasn't done before. Shows more information than previously, also a cleaner way to fill in intake information.

## Images of changes
Brig 1
Brig 2
Brig 3
Brig 4

## Changelog
:cl:
add: Converts brig timers to NanoUI
/:cl:

PR #14151, opened 2020-08-26
[TGUI] Brig Timers

Kyep:
## What Does This PR Do
Converts brig timers (the things you use to give people timed sentences in the brig) to TGUI.

Adds the following new features:
- When you first open the brig timer window, if it has no active prisoner, it will generate a drop-down menu of all handcuffed humanoids within 4 tiles. There will be one name input box showing "-----", which you can always use to enter the name manually. Assuming there are eligible names, under the manual input box there will also appear the drop-down menu that lets you quickly set the prisoner name to one of those nearby cuffed humanoids. In most cases, this will save you from having to type out prisoner names yourself.
- After entering someone's name as a prisoner, but before you sentence them, if they have no sec record, a warning triangle will display next to their name, giving you a clue that either you typoed it, or they have no sec record.
- After entering someone's name as a prisoner, and starting their sentence, if they have no sec record, their name will display in red.
- When you start someone's brig timer, it will tell the prisoner what crime they are being brigged for. Normally, only they will see this message, but if the timer cannot find them (either they left, or you entered their name incorrectly) then it will announce in local chat instead. 
- While someone is a prisoner, there is a "Reset Timer" button which resets the timer back to what it was at the start of the sentence. Attempting to use this will prompt for a reason. After entering a reason, the timer being reset will be announced on sec radio, along with being entered in their sec record. Both will show the prisoner's name, who reset their timer, and why. The prisoner themselves will also get a shorter, direct message stating that their timer was reset, and why.

## Images of changes
Default view:
brigtimer_nobodyhome

Prisoner chosen, ready to sentence:
brigtimer_readytosentence

Sentence actually running:
brigtimer_sentencerunning

How it appears in sec records (from examine) if you reset their timer:
brigtimer_resetupdate

How it appears if you enter a name with no sec record, prior to sentencing:
brigtimer_bogusname

How it appears if you enter a name with no sec record, after sentence has started:
brigtimer_bogusname2

Old interface that this replaces (for comparison):
https://github.com/ParadiseSS13/Paradise/pull/13487

## Changelog
:cl: Kyep
tweak: Converted brig timers to TGUI.
add: Added convenience for brig timers, such as selecting the new prisoner from a drop down rather than having to type their name.
/:cl:


tgui/packages/tgui/interfaces/BrigTimer.js
```diff
@@ -0,0 +1,121 @@
...
+  const { act, data } = useBackend(context);
+  data.nameText = data.occupant;
+  if (data.timing) {
+    if (data.prisoner_hasrec) {
+      data.nameText = (<Box color="green">{data.occupant}</Box>);
+    } else {
+      data.nameText = (<Box color="red">{data.occupant}</Box>);
+    }
+  }
+  data.nameIcon = "pencil-alt";
```
AffectedArc07, line 16: Define this as its own var here with `let`. Dont modify `data`, its meant to be read only.

tgui/packages/tgui/interfaces/BrigTimer.js
```diff
@@ -0,0 +1,121 @@
...
+      data.nameIcon = "exclamation-triangle";
+    }
+  }
+  let nameOptions = [];
+  let i = 0;
+  for (i = 0; i < data.spns.length; i++) {
+    nameOptions.push(data.spns[i]);
+  }
+  return (
+    <Window resizable>
```
AffectedArc07, line 28: Does this really need to be resizable?

AffectedArc07 (changes requested):
Only a few things

code/game/machinery/doors/brigdoors.dm
```diff
@@ -276,71 +277,101 @@
...
+/obj/machinery/door_timer/allowed(mob/user)
+	if(user.can_admin_interact())
+		return TRUE
+	return ..() ? TRUE : FALSE
+
+/obj/machinery/door_timer/tgui_act(action, params)
+	if(..())
+		return
+	if(!allowed(usr))
+		to_chat(usr, "<span class='warning'>Access denied.</span>")
```
AffectedArc07, line 318:
```suggestion
	if(!allowed(user))
		to_chat(user, "<span class='warning'>Access denied.</span>")
```

Forgot these

tgui/packages/tgui/interfaces/BrigTimer.js
Kyep: I'd rather leave it resizable. Just in case, say, someone has an astronomically long name. Or there are literally 10 people cuffed outside cells. Or similar edge situations. Its always better to allow the user to resize the window, even if you THINK they should not need to.

code/game/machinery/doors/brigdoors.dm
```diff
@@ -276,71 +277,101 @@
...
+			crimes = prisoner_charge
+			prisoner_name = ""
+			prisoner_charge = ""
+			prisoner_time = ""
+			timing = TRUE
+			timer_start()
+			update_icon()
+		if("restart_timer")
+			if(timing)
+				var/reset_reason = sanitize(copytext(input(usr, "Reason for resetting timer:", name, "") as text|null, 1, MAX_MESSAGE_LEN))
```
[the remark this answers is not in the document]
Kyep: code\game\machinery\doors\brigdoors.dm:352:error: user: undefined var

code/game/machinery/doors/brigdoors.dm
```diff
@@ -276,71 +277,101 @@
...
-/obj/machinery/door_timer/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)
-	ui = SSnanoui.try_update_ui(user, src, ui_key, ui, force_open)
+/obj/machinery/door_timer/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = TRUE, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
+	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
 	if(!ui)
-		ui = new(user, src, ui_key, "brig_timer.tmpl", "Brig Timer", 500, 400)
+		prisoner_suggested_names = list()
+		for(var/mob/living/carbon/human/H in range(4, get_turf(src)))
+			if(H.handcuffed)
+				prisoner_suggested_names += H.name
```
AffectedArc07, line 288: Use `tgui_static_data` for this. The `tgui_interact` proc is meant to follow standard format and not have modifications like this in, especially when `tgui_static_data` exists for this very purpose

```dm
/thing/tgui_static_data(mob/user)
	var/list/data = list()
	prisoner_suggested_names = list()
	for(var/mob/living/carbon/human/H in range(4, get_turf(src)))
		if(H.handcuffed)
			prisoner_suggested_names += H.name
	data["prisoner_suggested_names"] = prisoner_suggested_name
	return data
```

You can then interact with this in the UI in the same way as `tgui_data()`

code/__HELPERS/time.dm
```diff
@@ -95,7 +95,7 @@ proc/isDay(var/month, var/day)
 /proc/seconds_to_time(var/seconds as num)
```
AffectedArc07, line 95: Thankfully this proc is only used in `brigdoors.dm` so you dont have to go around and update it everywhere.

code/game/machinery/doors/brigdoors.dm
```diff
@@ -276,71 +287,108 @@
...
-	data["time_left"] = seconds_to_clock(timeleft())
-	data["timing"] = timing
-	data["isAllowed"] = allowed(user)
-	data["prisoner_name"] = prisoner_name
-	data["prisoner_charge"] = prisoner_charge
-	data["prisoner_time"] = prisoner_time
 
+/obj/machinery/door_timer/tgui_static_data(mob/user)
+	var/list/data = list()
+	data["spns"] = list()
```
AffectedArc07, line 300: Not a merge-blocking nitpick, but what is this even supposed to mean? Security prisoners near screen?

AffectedArc07 (approved):
Seems fine

code/game/machinery/doors/brigdoors.dm
Kyep: Originally the var was named Suggested_Prisoner_NameS (without the letters capitalized), which exactly describes what the var is - a list of names suggested to the brigging officer to save them time identifying their prisoner.
I ended up changing it to "spns" simply because it was short, and I was tired of TGUI's line length limits forcing me to re-write stuff in a way that took 3-5 more lines than necessary, just because it couldn't handle var names as long as "suggested_prisoner_names" without starting to force every line to be broken up into several lines.

PR #14155, opened 2020-08-26
Removes KFV

ppi13:
## What Does This PR Do

This removes KFV as a craftable item, it can still be spawned in via admins

## Why It's Good For The Game

KFV was fun in the beginning, but when people straight up starts bullying people, and turning the server environment hostile against vox players, there is a difference between just joking and bullying, and if an item is used like that, it should only belong in admin events.

## Changelog
:cl: ppi
del: KFV Recipie
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
+++ b/code/modules/food_and_drinks/kitchen_machinery/deep_fryer.dm
@@ -126,13 +126,3 @@
 /datum/deepfryer_special/carrotfries
 	input = /obj/item/reagent_containers/food/snacks/grown/carrot/wedges
 	output = /obj/item/reagent_containers/food/snacks/carrotfries
-
-/datum/deepfryer_special/fried_vox
-	input = /obj/item/organ/external
-	output = /obj/item/reagent_containers/food/snacks/fried_vox
-
-/datum/deepfryer_special/fried_vox/validate(var/obj/item/I)
-	if(!..())
-		return FALSE
-	var/obj/item/organ/external/E = I
-	return istype(E.dna.species, /datum/species/vox)


Carthusia:
mmmmm, personally I feel adding more species based foods would solve this, vox is the only one who get it, and its such an obvious reference, so the others need one in kind to fill the gap.

Unathi on a stick, Iced-Drask, Candied-Kidan etc

Balthusdire:
Monkey is a very common food, as well as the other simple mobs (stok, farwa, etc). I've seen plenty of humanized monkies turned into other races for food, for heads for pikes, for anything. Soylent Green has long been on the menu for foods to be created. NO ONE should get special consideration just because they've decided they want it. If vox are upset that humanized monkies are being used for food, that's a problem with their attitude, not the game. The fact that you included the word "bullying" is extremely concerning since if you are referring to the recent round that happened there was not even a slight bit of bullying involved, literally all that happened was a shop was set up to serve KFV. If that is "bullying" for vox then maybe we should look at removing vox since its encouraging a very toxic play style that demands special privileges.

That being said we DEFINITELY need to add more race based foods. Draskcicles, chocolate covered kidan, diona-root beer, there are so many great options.

Mitchs98:
Pretty much agreeing with Balth. Just because KFV is far more common, doesn't mean Soylent Green or other such items aren't made or exist. Pike heads and everything else are made. Monkies and such are humanized for organs and limbs, no one bats an eye. Chef can take corpses from the morgue with CMO permission, you might have a few people call them out on it but at the end of the day most won't or won't care.

People will always find a hot topic to jump on the riot train about, this is not really different. Nor do I see it happen every single round, and I don't really see anyone bullied by it on an OOC level either.

Spacemanspark:
who was even getting 'bullied'

Shockpoint:
> turning the server environment hostile against vox players

Vox do this on their own.

Tokorizo:
Im a disgusting vox player and i disagree with removing KFV, I dont really like it as well because the joke is almost a dead horse to me but if people get a kick out of it I dont care. And I also can take a joke/insult.

Spacemanspark:
I do agree that both sides are taking it a bit far, from what I've seen.

However, I feel the comment many make regarding cat ears applies here as well; fix the people causing the issue, not the item.

RaidarV:
Imagine thinking that bullying Vox is a bad thing.

This post was made by Unathi gang.

ZomgPonies:
Time to skin some lizards and make purses out of them.

Ty-Omaha:
As the creator of KFV I agree with all the other alternatives above. It's so easy to just remove something than to actually fix / balance it.

AffectedArc07:
This has 3 objections from heads and maints, and has absolutely no one in favour. No.

procdrone:
LRP here we go

PR #12846, opened 2019-12-25
Gives Vox a chirp emote and hiss emote

ppi13:
## What Does This PR Do
Give Vox the ability to emote chirps and hiss

## Why It's Good For The Game
Burd sounds are cute, emotes are great.

## Changelog
:cl:
add: Add multiple bird hiss sounds
add: Add multiple bird chirp sounds
add: Add age pitch to hiss emote
add: Add age pitch to chirp emote
/:cl:

PR #14164, opened 2020-08-28
Give Vox extra emote sounds

ppi13:
## What Does This PR Do

Improved version of #12846
Adds hiss and chirp emote to vox
All sounds have an universal loudness of -23 LUFS, so that all sounds are equally loud
image
Which is just good sound design.
These sounds are recordings of real vultures, a very raptor like noise.

## Why It's Good For The Game

Immersion is a huge part of RP, and good sound goes a long way in assisting with that, pluss birbs are great

## Changelog
:cl: ppi
add: Add the ability for Vox to use use a chirp and hiss emote
soundadd: Add 10 new vox chirps
soundadd: Add 9 new vox hisses
/:cl:


Mitchs98:
Rattling emotes for Plasmamen when V.V?

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -26,6 +26,9 @@
 	if(mind)
 		miming = mind.miming
 
+	var/vox_chirp = list('sound/effects/vox/voxchirp1.ogg','sound/effects/vox/voxchirp2.ogg','sound/effects/vox/voxchirp3.ogg','sound/effects/vox/voxchirp4.ogg','sound/effects/vox/voxchirp5.ogg','sound/effects/vox/voxchirp6.ogg','sound/effects/vox/voxchirp7.ogg','sound/effects/vox/voxchirp8.ogg','sound/effects/vox/voxchirp9.ogg','sound/effects/vox/voxchirp10.ogg') //Credit to AlexTriceratops123  (https://www.youtube.com/watch?v=uA5yGyB_z5U) for the sound.
+	var/vox_hiss = list('sound/effects/vox/voxhiss1.ogg','sound/effects/vox/voxhiss2.ogg','sound/effects/vox/voxhiss3.ogg','sound/effects/vox/voxhiss4.ogg','sound/effects/vox/voxhiss5.ogg','sound/effects/vox/voxhiss6.ogg','sound/effects/vox/voxhiss7.ogg','sound/effects/vox/voxhiss8.ogg','sound/effects/vox/voxhiss9.ogg') //Credit to AlexTriceratops123  (https://www.youtube.com/watch?v=uA5yGyB_z5U) for the sound.
```
SteelSlayer, line 30: Because these all share a similar name, you shouldn't hardcode them here. Instead do something like:
```dm
playsound(loc, 'sound/effects/vox/voxchirp[rand(1, 10)].ogg, ...)
```

iantine:
Happy to see this brought back to life!

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -224,13 +230,26 @@
...
 
-			if(!muzzled)
-				message = "<B>[src]</B> hisses[M ? " at [M]" : ""]."
-				playsound(loc, 'sound/effects/unathihiss.ogg', 50, 1, frequency = get_age_pitch()) //Credit to Jamius (freesound.org) for the sound.
-				m_type = 2
-			else
-				message = "<B>[src]</B> makes a weak hissing noise."
-				m_type = 2
+			if(isunathi(src))
+				if(!muzzled)
```
farie82, line 234: Best to flip these.
Check the muzzled first so you can catch the makes a weak noise.
Then check if they are either unathi or vox. Can even just make the calling of the actual sound and message of hissing once.
So something like this:
```
if(!muzzled)
  var/sound_file
  if(vox)
    sound_file = X
  else if(uanthi)
    sound_file = Y
  play_sound
....
```

farie82 (changes requested):
`fix: Update the *help for vox to include the new species specific emotes`
Isn't quite a fix since you're adding it in this PR. This is just an added feature due to your other added feature and can be omitted.

A fix is something that was forgotten or something changed in code to make the code behave as expected, think of it not runtiming or a bullet actually hitting a target only once. If for example you forgot it in this PR and made a new PR to actually add it. Then it is a fix

MRgaty:
Vox already have a ton of emotes, I think this will make them more annoying.

Spacemanspark:
Tbh vox already have emotes and all of them fill the niche of 'obnoxious shitbird raptor scream' well enough.

I'd prefer to see other races get their own sound effects. IPCs also have quite a bit, but other than that many races are rather lacking.

Also vox are pox and deserve to have their skipjacks raided

ppi13:
> Vox already have a ton of emotes, I think this will make them more annoying.
Like what?
They have quill?

Spacemanspark:
Quill and scream are enough imo

AffectedArc07:
Im not inclined to make vox even more obnoxious than they already are.

ppi13:
Vox scream is 13 dB louder then any of these, audio is set to a level where it's not obnoxious, compared to several other sounds that are.

farie82 (approved):

ExusA:
I'd argue that having more different sounds makes a species a lot less obnoxious if anything. Having say, one or two emotes that all players of a species use gets on peoples' nerves MUCH faster than having a whole bunch of different ones that get used less commonly.

Carthusia:
Id agree that the normal SRKEEEEEEECH everytime I try to talk can get a bit annoying even as a vox player, having a series of much quieter birb sounds that rotate through eachother seems a good way to reduce the issue of repetitiveness.

Tokorizo:
More vox noises to make the crew hate my guts, I love it

Kyep (changes requested):
I don't like this.
While I'm fine with the idea of more vox sounds in general:
1) there's too many of these
2) they all sound very similar
3) they're all grating noises

variableundefined:
Objecting, echoing Affected and Kyet's concerns.

Closing due to two maintainer objections.

PR #14166, opened 2020-08-28
[TGUI] Communications Console

Kyep:
## What Does This PR Do
Major Changes:
- Converted Communications Console to TGUI, and improved its interface in many ways.
- Removed the comms card program. Using the program required a clunkly modular laptop built from parts, and a captain-level ID. It was rarely used, and when it was used it was typically used to keep changing the alert level, or calling/recalling the shuttle, while hiding in maint.
- Station-wide announcements from the comms computer now require at least 6 characters. No more "E" over priority announce. Messages to CC/Syndicate (direct message, nuke codes, syndi msg) now require at least 20 characters.

Minor Changes to comms computer:
- Now shows how many messages have been logged in the console. 
- No longer logs auto-generated HTML messages (such as station goal messages, and swarmer announcement) to the message log. These messages won't render in TGUI, everyone has seen them many times before, logging the message is unnecessary because the comms console prints a (well formatted) paper with the message anyway, and axing them from the log makes the "messages received" count more useful as that now refers exclusively to admin-sent messages.
- Now shows the entire interface, even to non-Captains. (though, normal heads will see the buttons in the Captain section grayed out)
- Now splits up the interface into Captain-only and general command sections... instead of the previous mess where all the options were mixed together and half of them did not tell you they were above your clearance level until you clicked them.
- Now shows the recall timer for the shuttle, if the shuttle is being recalled, as opposed to acting like you can still call the shuttle when you obviously can't.
- Now also shows the timer for how often you can message CC/syndicate, so we don't get ahelps/mhelps about how long the arrays take to re-align.
- When admins send a communications report, comms consoles will now print papers in the name format: "paper- 'Nanotrasen Update'", "paper- 'Syndicate Message'", "paper- 'Classified: Nanotrasen Update 2'", etc. This has several advantages over the old system. First, reports are no longer treated as CC updates even when they're from the syndicate, etc. And second there is some variety in paper names now, which makes it easier to find the right paper in a pile of papers.
- The fact that the shuttle is (and has been) un-callable for the first 20 minutes of the shift is now shown in the interface, with the shuttle being listed as "refueling: XX:XX".

## Images of changes
Default state (not logged in):
comms_default
Logging in with captain-level ID:
comms_captain
Logged in as captain, having pressed all the buttons:
comms_captain_used
Shuttle section, when shuttle is being recalled to CC:
comms_recalling
Shift start, shuttle cannot be called due to it being too early:
comms_refueling
Head of department, browsing comms messages:
comms_messages_headofdept
Viewing a single message:
comms_singlemessage
Head of department, fiddling with station status monitor interface:
comms_statusicons_bridgeguy

## Changelog
:cl: Kyep
tweak: Converted communications console to TGUI, and made it much easier to use.
del: Removed communications program.
/:cl:


TDSSS:
>Removed the comms card program. Using the program required a clunkly modular laptop built from parts, and a captain-level ID. It was rarely used, and when it was used it was typically used to keep changing the alert level, or calling/recalling the shuttle, while hiding in maint.

All good changes except this one, I'll miss that program. Being able to set up command somewhere else if the bridge is destroyed or similar was always useful. And sending station announcements on the go from a laptop was great. Any chance you can just make this less of a total pain to create rather than removing it?

Kyep:
@TDSSS You can still set up a bridge somewhere else. Just grab the comms console board from secure tech storage or R&D, some metal/wire/glass, and make a comms computer there. Probably faster than gathering the parts for a laptop, building one, and downloading the program. Not to mention a proper comms computer can be used by other heads, whereas the comms program can only be used by the Captain. So yeah, just use a proper computer instead, it will take about the same amount of time, and work better.

Modular computers in general are terrible. You'd think a modular computer program would just be like an embedded copy of its regular computer. But it isn't. No. It is actually an entirely separate thing code-wise, with all its own code, which just produces an interface that *looks* like the proper computer. But it isn't. They copy-paste all the code of the regular computer that they are a bad copy of, but also change that code in various ways that make it unnecessarily difficult to port functionality between them. Like for example, changing var and parameter names for no good reason. Basically, the existence of modular computers forces everyone who writes code for computers to make one of three choices:
A) ignore the modular computer version, which leads to modular computers and their programs gradually becoming ever more second-class, and not worth it over just building a proper computer of the same type
B) contribute to the copy/paste problem by copying over your code into the modular computer but tweaking it to follow the modular computer conventions, which feels like coding the same feature twice, once for normal computers and then having to re-do it just to support modular computers
C) dramatically expand the scope of the task you signed up to do, by attempting to unify both normal and modular computers as much as possible by having them both utilize shared functions embedded in something else (like a subsystem). This reduces the degree of copy/paste code in modular computers but takes up a great deal more of your time and AT BEST you only ever end up with modular computer programs functioning equivalent to regular computers. Long-term this option doesn't make sense as we could just get rid of modular computers totally and spend all the time we save fixing other things that aren't copy-paste, or adding neat new features we actually want. Rather than having to spend a bunch of extra time effectively writing each computer feature 3 times - once for proper computers, once again for modular computers, and then a third time in a unified way that avoids copy/paste between the two.

code/game/machinery/computer/communications.dm
```diff
@@ -16,20 +15,23 @@
 	req_access = list(ACCESS_HEADS)
 	circuit = /obj/item/circuitboard/communications
 	var/prints_intercept = 1
```
AffectedArc07, line 17: Not part of your edits, but could you make these defaults `TRUE`/`FALSE` to aid readability

code/game/machinery/computer/communications.dm
```diff
@@ -70,85 +72,76 @@
...
-		SSnanoui.update_uis(src)
-		return
-
-	if(href_list["logout"])
-		authenticated = COMM_AUTHENTICATION_NONE
-		crew_announcement.announcer = ""
-		setMenuState(usr,COMM_SCREEN_MAIN)
-		SSnanoui.update_uis(src)
+		if(authenticated == COMM_AUTHENTICATION_NONE)
+			to_chat(usr, "<span class='warning'>You need to swipe your ID.</span>")
```
AffectedArc07, line 105: I feel `swipe` here is misleading, given how this console doesnt actually handle being hit by an ID

code/game/machinery/computer/communications.dm
```diff
@@ -70,85 +72,76 @@
...
+			if(istype(I))
 				if(ACCESS_CAPTAIN in I.access)
-					change_security_level(text2num(href_list["level"]))
+					change_security_level(text2num(params["level"]))
 				else
 					to_chat(usr, "<span class='warning'>You are not authorized to do this.</span>")
-				setMenuState(usr,COMM_SCREEN_MAIN)
+				setMenuState(usr, COMM_SCREEN_MAIN)
 			else
 				to_chat(usr, "<span class='warning'>You need to swipe your ID.</span>")
```
AffectedArc07, line 136: Same here

code/game/machinery/computer/communications.dm
```diff
@@ -70,85 +72,76 @@
...
 				if(message_cooldown)
 					to_chat(usr, "<span class='warning'>Please allow at least one minute to pass between announcements.</span>")
-					SSnanoui.update_uis(src)
 					return
 				var/input = input(usr, "Please write a message to announce to the station crew.", "Priority Announcement")
 				if(!input || message_cooldown || ..() || !(is_authenticated(usr) == COMM_AUTHENTICATION_MAX))
-					SSnanoui.update_uis(src)
 					return
 				crew_announcement.Announce(input)
 				message_cooldown = 1
```
AffectedArc07, line 147: Any change you can take this and the spawn below it to use booleans, as well as `addtimer` instead of the `spawn()`

code/game/machinery/computer/communications.dm
```diff
@@ -158,35 +151,34 @@
...
 			var/response = alert("Are you sure you wish to recall the shuttle?", "Confirm", "Yes", "No")
 			if(response == "Yes")
 				cancel_call_proc(usr)
 				if(SSshuttle.emergency.timer)
 					post_status("shuttle")
-			setMenuState(usr,COMM_SCREEN_MAIN)
+			setMenuState(usr, COMM_SCREEN_MAIN)
 
 		if("messagelist")
 			currmsg = 0
```
AffectedArc07, line 173: And here

code/game/machinery/computer/communications.dm
```diff
@@ -199,40 +191,41 @@
 						currmsg = 0
 					if(aicurrmsg == id)
 						aicurrmsg = 0
```
AffectedArc07, line 193: And here (you get the idea, I am gonna stop commenting there)

code/game/machinery/computer/communications.dm
```diff
@@ -241,17 +234,15 @@
 				centcomm_message_cooldown = 1
 				spawn(6000)//10 minute cooldown
 					centcomm_message_cooldown = 0
```
AffectedArc07, line 236: Same thing with the spawns here

code/game/machinery/computer/communications.dm
```diff
@@ -260,31 +251,29 @@
...
```
AffectedArc07, line 253: And here

AffectedArc07:
I would split these out into two rows
image
```
Current Alert Level: Current alert, with a <box color={}> attribute
Set Alert Level: Have the 2 buttons, and if the current alert is green or blue, set the buttons as disabled
```

tgui/packages/tgui/interfaces/CommunicationsComputer.js
```diff
@@ -0,0 +1,304 @@
...
+          </Window.Content>
+        </Window>
+      );
+
+    default:
+      return (
+        <Window>
+          <Window.Content>
+            {authBlock}
+            ERRROR. Unknown menu_state: {data.menu_state}
```
AffectedArc07, line 297: Maybe add a note here saying "Please report this to NT Technical Support"

AffectedArc07 (changes requested):
Just a few things, mostly related to cleaning up old things for the sake of completeness

procdrone:
this makes me sad. I've been using a "remote bridge" laptop as captain ALWAYS. It allowed me to visit areas of the station without having the risk of being far away to announce or change alert level. Like many service areas, talk to crew...

That will just lock me into bridge even more, if I dont want be accused of "not doing my job".

AffectedArc07:
> this makes me sad. I've been using a "remote bridge" laptop as captain ALWAYS. It allowed me to visit areas of the station without having the risk of being far away to announce or change alert level. Like many service areas, talk to crew...
> 
> That will just lock me into bridge even more, if I dont want be accused of "not doing my job".
> 
> EDIT:
> Additionally, I feel like alert level buttons should be somewhere bellow announcement button

Expecting someone to have to update 2 entirely separate UI templates and 2 entirely separate sets of code to change 1 tiny thing is god awful from a code standpoint, and makes people want to work on things even less. This is why modular computers are seldom used and never worked on.  Modular consoles are flawed. Yes they are modular from an ingame standpoint, but by code they are the least modular thing we have. 

The extra work that modular computers puts on devs slows the process down as its extra things to review, takes longer to fix bugs with, and makes people have to learn 2 separate computer baselines compared to just the standard ones. Having to support modular computers is going to slow us down even further. (Honestly in my opinion we should just axe modular computers/laptops/tablets and add **some** of the features as PDA apps)

AffectedArc07 (approved):

PR #14169, opened 2020-08-28
Centralised persistent data management framework

AffectedArc07:
## What Does This PR Do
This PR converts all current persistent atoms (Ian, Runtime and the colossus which is inaccessible because its a broken piece of garbage) are now centrally managed by `SSpersistent_data`. This SS will handle saving and loading of persistent data on server load and on round end, since currently we rely on `process()` and `Life()` to save and load, as well as what state the ticker is in. Thats awful.

This framework also allows us to add in more persistent objects in a relatively standard way in the future (I have plans for a chemistry explosions counter). 

## Why It's Good For The Game
Cleaning up of bad code, as well as adding in systems which allow for easier management of persistent data

## Changelog
:cl: AffectedArc07
tweak: Ian and Runtime now use SSpersistent_data for managing saving and loading of their data
/:cl:


code/modules/mining/lavaland/loot/colossus_loot.dm
```diff
@@ -49,25 +49,21 @@
 		qdel(src, force=TRUE)
 		return
 	current = src
-	ReadMemory()
-	. = ..()
-
-/obj/machinery/smartfridge/black_box/process()
-	..()
-	if(!memory_saved && SSticker.current_state == GAME_STATE_FINISHED)
-		WriteMemory()
```
AffectedArc07, line 58: This is my main reason for making this SS. This proc is disgusting

code/controllers/subsystem/persistent_data.dm
```diff
@@ -0,0 +1,58 @@
...
+SUBSYSTEM_DEF(persistent_data)
+	name = "Persistent Data"
+	init_order = INIT_ORDER_PERSISTENCE // -95 | Loads after EVERYTHING else
+	flags = SS_NO_FIRE
+	/// List of atoms registered into the subsystem for persistent data storage. Can be anything at all
+	var/list/registered_atoms = list()
+
+/datum/controller/subsystem/persistent_data/Initialize()
+	// Load all the data of registered atoms
+	for(var/atom/A in registered_atoms)
```
farie82, line 19: Why only atoms? I can see myself using this for instance for the logging tools or other backend logic that is not managed by a SS.
Can also be made into a non typed loop if there is only one type in the list. Same as the one below

code/controllers/subsystem/persistent_data.dm
```diff
@@ -0,0 +1,58 @@
...
+/**
+  * Proc to register an atom with SSpersistent_data
+  *
+  * This will add any provided atom to the list of registered atoms, and add it to the Initialization queue
+  * If the system has already initialized, it calls PersistentLoad() at that moment
+  *
+  * Arguments:
+  * * A - Atom to register
+  */
+/datum/controller/subsystem/persistent_data/proc/register(atom/A)
```
farie82, line 38: Best to also make a unregister proc. Will do the same as you have now in other places. The -=. But it will be a bit cleaner than directly accessing the list

code/modules/mob/living/simple_animal/friendly/dog.dm
```diff
@@ -420,13 +420,18 @@
...
 /mob/living/simple_animal/pet/dog/corgi/Ian/Initialize(mapload)
 	. = ..()
-	//parent call must happen first to ensure IAN
-	//is not in nullspace when child puppies spawn
+	SSpersistent_data.register(src)
+
+/mob/living/simple_animal/pet/dog/corgi/Ian/death()
+	Write_Memory(TRUE)
+	SSpersistent_data.registered_atoms -= src // We already wrote here, dont overwrite!
+	..()
```
farie82, line 432: I'd put this before the write memory etc. It will return FALSE if it did not die and TRUE if it did die.
So also return the value

code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -62,8 +60,9 @@
 		return baby
 
 /mob/living/simple_animal/pet/cat/Runtime/death()
-	if(can_die() && !memory_saved)
-		Write_Memory(1)
+	if(can_die())
```
farie82, line 63: Same here use the `. = ..()` as stuff to check if they actually died.

code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -62,8 +60,9 @@
 		return baby
 
 /mob/living/simple_animal/pet/cat/Runtime/death()
```
farie82, line 62: Missing a param here `gibbed`

code/modules/mob/living/simple_animal/friendly/dog.dm
```diff
@@ -420,13 +420,18 @@
...
-	var/memory_saved = FALSE
 	var/saved_head //path
 
 /mob/living/simple_animal/pet/dog/corgi/Ian/Initialize(mapload)
 	. = ..()
-	//parent call must happen first to ensure IAN
-	//is not in nullspace when child puppies spawn
+	SSpersistent_data.register(src)
+
+/mob/living/simple_animal/pet/dog/corgi/Ian/death()
```
farie82, line 429: Missing a param here `gibbed`

farie82 (changes requested):
Interesting idea to use a SS for this. I was myself thinking about doing a signal on the world shutdown. Yet the world is no datum. Making another object and having things register on that kinda makes a hook object. But I guess this is sort of that but without signals.

code/controllers/subsystem/persistent_data.dm
AffectedArc07: I cant think of any datums which would have persistent storage, which wouldnt be externally managed (Example: When I do map rotation, the saving and loading of data will be handled by the mapping SS). It also relies on the atoms existing to be added to this list, as well as adding them with `Initialize()`. Instancing datums becomes an issue, and should be managed by their own service. If you wanted to do this for logging tools, it should be made into a subsystem, so all the vars can be associated with it, especially since I spent time before killing off orphaned controllers which were hooked into other subsystems in less than optimal ways.

code/controllers/subsystem/persistent_data.dm
AffectedArc07: I dont see the point since `register` was only added for loading things at runtime, that arent placed on the map. If you look at the original commit for this PR, using `register()` is cleaner than the if check each time

code/modules/mob/living/simple_animal/friendly/dog.dm
AffectedArc07: I will look at other implementations before I start changing things round. This was meant to only change how reading and writing was called, not refactor death

code/modules/mob/living/simple_animal/friendly/cat.dm
AffectedArc07: Same here

code/controllers/subsystem/persistent_data.dm
farie82: It has no logic behind it yet. Yet it might get some logic behind it.
Also makes it more readable than removing it from a list. Will also remove the suggestion you can just add an atom to the list using += or |=

code/modules/mob/living/simple_animal/friendly/cat.dm
AffectedArc07: Kind of out of the scope of this PR

code/modules/mob/living/simple_animal/friendly/dog.dm
AffectedArc07: Param is not used anywhere and its not halting dreamchecker

code/modules/mob/living/simple_animal/friendly/dog.dm
farie82: It's not refactoring death. It's just taking the value that it returns and using that to check if the dog actually died

AffectedArc07:
```
/mob/living/simple_animal/pet/dog/corgi/Ian/death()
	. = ..()
	if(.)
		Write_Memory(TRUE)
		SSpersistent_data.registered_atoms -= src // We already wrote here, dont overwrite!

```
???

code/controllers/subsystem/persistent_data.dm
AffectedArc07: I cant think of any logic at all that would be behind this. If an object has already been saved, it should be out of the list, since thats the only reason the list exists, for either invoking load on initialize, and for saving at round end.

AffectedArc07:
This PR also removes the lavaland blackbox. Its unspawnable, never going to be implemented, half broken, and is the most overpowered thing to exist in this code. You cant even obtain it now without admin spawning it, and its never been used (The savefile on disk is zero bytes). Im not going to continue supporting it unless it merge-blocks this PR.

PacifistDalek:
Out of sheer curiosity, what did the blackbox do?

PacifistDalek:
Heck, for that matter, what about Ian and Runtime and the like are persistent? Sorry, I'm not great at reading code

AffectedArc07:
@PacifistDalek 
Blackbox is persistent through-round item storage, and its disabled for a reason
Ian ages once each shit, and will be a puppy next shift if he is killed the shift before. His age "record" is stored on the file, and if Ian hits the age record, he becomes old Ian (he gets a doggy wheelchair as well as moving a lot slower)
Runtime keeps track of the cats she breeds in a shift. If runtime breeds 10 cats in one shift, she will spawn with a family of 10 the next shift

PacifistDalek:
That's awesome!

code/controllers/subsystem/persistent_data.dm
SteelSlayer: Isn't `registered_atoms` always empty on server start? I'm probably missing the part where data is carried over to the next round.

AffectedArc07: `SSatoms` initialises before `SSpersistent_data`, which adds the atoms into the `SSatoms.registered_atoms` list.

SteelSlayer: Ah ok, and `SSpersistent_data.register(src)` is called in `New` of Ian and runtime, which adds it to the list. I see.

PR #14171, opened 2020-08-28
Changes dark mode command channel color to the old one

TheSardele:
## What Does This PR Do
Changes the dark mode color of the command channel back to the old value.

## Why It's Good For The Game
In my opinion the new color the command channel uses was too close to the color logs use, and the old one never actually seemed problematic to me in dark mode.

## Images of changes
Current:
Untitled
With this PR:
Untitled2

## Changelog
:cl:
tweak: Changes the command channel color in dark mode to the old one
/:cl:

### DIFF
--- a/goon/browserassets/css/browserOutput-dark.css
+++ b/goon/browserassets/css/browserOutput-dark.css
@@ -270,7 +270,7 @@ em						{font-style: normal;	font-weight: bold;}
 .deadsay				{color: #B800B1;}
 .radio					{color: #408010;}
 .deptradio				{color: #993399;}
-.comradio				{color: #526aff;}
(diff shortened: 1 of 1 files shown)


Deelite34:
The problem with the old dark mode command colour is that it is significantly harder to read than other colours in dark mode. I'm against the old colour, could you try instead to find shade of blue(or maybe other colour) easier to read than old command colour?
image

AffectedArc07:
Personally to me, these seem readable enough as they are now, and no one else has really complained about it since. It is fine how it is now, in my opinion.

variableundefined:
Agreed with above. 

Two maint objection, closing.

PR #14176, opened 2020-08-29
[TGUI] Keycard Swiper

Kyep:
## What Does This PR Do
Converts Keycard Swipers to TGUI.

## Notable Differences
- Swipers now allow 5 seconds for the second person to swipe to confirm. Previously, you only had 2 seconds.
- Swipers now provide more feedback. When requesting an ERT, if you're the second person, you can see the ERT request reason they provided. In all cases, you can see whether you're swiping to start your own request, or to support someone else's, as well as what that request is. 

## Images of changes
Default view:
kc_1_default
Action selected, ready to swipe (ERT requires reason, which shows up only for ERT requests):
kc_2_readytoswipe
Swiped on my side (upper), other side is prompted to swipe to confirm (lower):
kc_3_swiped
Completed action, green on both sides:
kc_4_complete

## Changelog
:cl: kyep
tweak: Converted Keycard Swipers to TGUI.
/:cl:


AffectedArc07:
Thank god you didn't do TGs way of handing "swipes". They make you have the UI open and hit a confirm button, and disable the ability to attack it with an ID, which fucked with my muscle memory hard.

code/modules/security_levels/keycard authentication.dm
```diff
@@ -4,17 +4,17 @@
...
-	var/screen = 1
+	var/active = FALSE // This gets set to TRUE on all devices except the one where the initial request was made.
+	var/event
+	var/swiping = FALSE // on swiping screen?
 	var/list/ert_chosen = list()
-	var/confirmed = 0 //This variable is set by the device that confirms the request.
-	var/confirm_delay = 20 //(2 seconds)
-	var/busy = 0 //Busy when waiting for authentication or an event request has been sent from this device.
+	var/confirmed = FALSE // This variable is set by the device that confirms the request.
+	var/confirm_delay = 50 // time allowed for a second person to confirm a swipe. Deciseconds.
```
AffectedArc07, line 12:
```suggestion
	var/confirm_delay = 5 SECONDS // time allowed for a second person to confirm a swipe. 
```

code/modules/security_levels/keycard authentication.dm
```diff
@@ -59,57 +62,58 @@
 		stat |= NOPOWER
 
 /obj/machinery/keycard_auth/attack_ghost(mob/user)
-	ui_interact(user)
+	tgui_interact(user)
 
-/obj/machinery/keycard_auth/attack_hand(mob/user as mob)
+/obj/machinery/keycard_auth/attack_hand(mob/user)
 	if(..())
 		return 1
```
AffectedArc07, line 69:
```suggestion
		return TRUE
```

AffectedArc07 (commented):
Only some styling things

procdrone:
If you want to make reason visible, why make open the UI, if confirming the action doesn't require that? Wouldn't it be better that the reason is transmitted to command comms before its accepted, and then only confirmation it succeeded?

Kyep:
@AffectedArc07 I considered several different major reworks to how they function, but I ultimately decided it wasn't necessary, and that the best course of action was just converting them to TGUI while changing as little as possible about how they worked.
They do get a few interface clarity upgrades, but that's about it.

@procdrone I don't actually think you have to open the UI on the second device. If you simply swipe while its flashing, you confirm whatever request the other person was trying to make. The reason I made it display the status IF you have the UI open, is that the previous nano UI was unclear about exactly what you were doing. If you were trying to swipe for ERT for example, and two people put in two different ERT reasons, it could be unclear about exactly which one you were swiping for. In this version, it is crystal clear, to the point that while your device is acting as a secondary swiper to confirm someone else's, it shows their swipe reason in your UI, displays it as uneditable, and displays that you're swiping to confirm. I realize most people probably don't care about this. But if you're the Captain or whatnot and you want to make sure your ERT has a decent request reason beyond "changelings"... you might care. In theory yes I could make it announce on radio... but in practice I'd rather just show it in the UI, because (A) telecomms can be down, especially during emergencies (B) command can get confused after multiple failed swipes as to what reason is the currently active one and (C) I feel the old interface was VERY confusing / unclear about the fact that the primary keycard swiper forces all others to use its action (ie: if you swipe for A, and second person has interface up for B and swipes, it confirms A rather than starting a call for B). For all those reasons I decided the UI needed to show what you were doing.

AffectedArc07 (approved):

PR #14177, opened 2020-08-29
Balances the safety override module and Fixes #14172

### LINKED ISSUE #14172: Borgs with Safety override do not get malf AIs law zero when master AI gets traitor put on them
**Issue Description**:

Borgs with Safety override do not get malf AIs law zero when master AI gets traitor put on them, illegals has to be removed to fix it

For law zero to be applied

**What happened instead**:

Nothing

It screws up auto traitor malf ai / admin malf ai if a borg has safety override

Give a borg safety override.
Malf the AI
note safety overriden borg did not have law zero

ppi13:
## What Does This PR Do

Fixes #14172 
Cyborgs with the Safety Overide Module does not de-sync the borg
Cyborgs with the Safety Overide Module are no longer protected against the emag act
Cyborgs with the Safety Overide Module can no longer change their own laws.
Cyborgs with the Safety Overide Module no longer show the examine text "The interface seems slightly damaged" when open"
Cyborgs with the Safety Overide Module will no longer have their mmi destroyed upon self destruction
Cyborgs with the Safety Overide Module will do a -1,0,2 explosion instead of a 1,2,4 explosion on self destruct
Due to cyborgs no longer being 'emagged' this is not an illegal piece of tech, so tech requirements has been increased to Combat and Materials 7.

## Why It's Good For The Game

As qwerty stated in his issue "It screws up auto traitor malf ai / admin malf ai if a borg has safety override" and there isnt supposed to be methods to 'protect' again emagging either.
With this change both malf ai's and syndicate traitors can emag cyborgs with or without the safety override.
So now admins can malf an ai at their own pleasure without the issue of it messing with the synced borgs.
Any potential law change abuse is prevented, as well as not making them super destructive if you have to destroy them when ai has been subverted.

## Images of changes

Self destruct now
image
Self destruct after this pr
image

## Changelog
:cl: ppi
tweak: Tweaks the safety override module to not be an emag act
/:cl:

### DIFF
--- a/code/modules/research/designs/mechfabricator_designs.dm
+++ b/code/modules/research/designs/mechfabricator_designs.dm
@@ -1094,7 +1094,7 @@
 	name = "Cyborg Upgrade (Safety Override)"
 	id = "borg_syndicate_module"
 	build_type = MECHFAB
-	req_tech = list("combat" = 4, "syndicate" = 2)
+	req_tech = list("combat" = 7, "materials" = 7)
 	build_path = /obj/item/borg/upgrade/syndicate
 	materials = list(MAT_METAL=10000,MAT_GLASS=15000,MAT_DIAMOND = 10000)
 	construction_time = 120
--- a/code/modules/research/designs/mechfabricator_designs.dm
+++ b/code/modules/research/designs/mechfabricator_designs.dm
@@ -1094,7 +1094,7 @@
 	name = "Cyborg Upgrade (Safety Override)"
 	id = "borg_syndicate_module"
 	build_type = MECHFAB
-	req_tech = list("combat" = 7, "materials" = 7)
+	req_tech = list("combat" = 7, "programming" = 7)
 	build_path = /obj/item/borg/upgrade/syndicate
 	materials = list(MAT_METAL=10000,MAT_GLASS=15000,MAT_DIAMOND = 10000)
 	construction_time = 120
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -185,7 +185,7 @@
 
 /obj/item/borg/upgrade/syndicate
 	name = "safety override module"
-	desc = "Unlocks the hidden, deadlier functions of a cyborg. Also prevents emag subversion."
+	desc = "Unlocks the hidden, deadlier functions of a cyborg."
 	icon_state = "cyborg_upgrade3"
 	origin_tech = "combat=4;syndicate=1"
 	require_module = TRUE
@@ -193,12 +193,10 @@
 /obj/item/borg/upgrade/syndicate/action(mob/living/silicon/robot/R)
 	if(..())
 		return
-	if(R.emagged)
-		return
 	if(R.weapons_unlock)
-		to_chat(R, "<span class='warning'>Internal diagnostic error: incompatible upgrade module detected.</span>")
+		to_chat(R, "<span class='warning'>Warning: Safety Overide Protocols have be disabled.</span>")
 		return
-	R.emagged = 1
+	R.weapons_unlock = 1
 	return TRUE
 
 /obj/item/borg/upgrade/lavaproof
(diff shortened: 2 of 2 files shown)


Kyep (commented):
I classify this as a balance change, not a bugfix.

procdrone:
not being able to be emmaged was one of the main reasons this upgrade exists AND requires illegal techs to make - which means not always and means security breaking SoP/law to provide contraband they should store in evidence instead. (dont joke, we all know maint loot is never donated to science)

If you want it to go through, have it require combat research levels instead of illegals, so its consistent.

ppi13:
Remove syndicate tech requirement, and replace it it with a higher combat and materials instead, which should roughly make it available around the same as it used to before, with the most common way to get illegal tech was deconstruction an x-ray laser.

Kyep (changes requested):
Are you sure that the ONLY consquences of using weapons_unlock = 1 instead of emagged = 1 are:
- Borg can now be emagged
- If AI is made malf by admins, borg now inherits AI's law 0
?

I think there are going to be several other high-impact effects of this change, which you are not mentioning.
Here are a couple of examples:
Does the fact they're no longer emagged=1 remove the ability for upgraded borgs to modify their own laws? I think it probably does. This might be a good thing but its still most definitely something you need to mention in the PR description.
Does removing the emagged=1 also cause borgs with this upgrade to no longer have the "The interface seems slightly damaged." message when you crowbar them open? I assume it does, which has the side effect that where previously, most borgs displayed that message late-round (since they were upgraded with this).... now.... only truly emagged borgs will display this message.... which makes emagged borgs much easier to tell apart from other borgs!
Those are just two examples of how changing the var has serious effects that you aren't mentioning in your PR. There are probably even more impacts of this change if you really think about it and check all the places in borg code where "emagged" var is checked for.

You really need to list all the impacts that this change will have on robots if you want to push for it. Otherwise, this change would have unanticipated/undocumented effects if merged.

I'm not saying I'm definitely against this change. I've actually been considering making a change like this myself. But I *am* saying that your PR needs to list *all* the consequences of the change, so that we can fairly evaluate it.

Spacemanspark:
@procdrone the point of illegals is to unlock your hidden module, not to prevent emagging.

I like this change, myself.

ppi13:
Updated with better documentation of changes

code/modules/research/designs/mechfabricator_designs.dm
```diff
@@ -1094,7 +1094,7 @@
 	name = "Cyborg Upgrade (Safety Override)"
 	id = "borg_syndicate_module"
 	build_type = MECHFAB
-	req_tech = list("combat" = 4, "syndicate" = 2)
+	req_tech = list("combat" = 7, "materials" = 7)
```
Kyep, line 1097: On the one hand, I do feel like it requiring syndicate 2, despite not being syndi tech, or even a syndi item, is a good change.
On the other hand, I feel like data/programming tech might be a better requirement than materials tech.
I also wonder whether requiring two level 7 techs will make this harder, or easier, to get than requiring combat4/syndi2, and whether that's a good thing or not.

ppi13: It should be gotten at roughly the same step, as to get combat 7 you need either a combat shotgun, captains rapier or chain of command, to get past combat 5, a
It should be slightly easier then it currently, but only due to the nature of syndicate tech being locked behind either thermals implant or xray lasers gun, with thermals needing 6 mat, 4 programming, 7 bio, 5 magnets and plasma 4, and xray needing combat 7, magnets 5, bio 5 and power 5, while changing materials to data might make a bit more sense, it would make it a fair more 'difficult' to get due to the nature of only a handful of miners knowing what tech they need to bring back, which often makes it one of the missing techs.  
Again this can differ from round to round, but for your average highpop round, it will be available at roughly the same time in general, all depending on how early roboticist offers upgrades or borgs asking for them.

Kyep: Do combat 7 + data 7.

ppi13: Done

Kyep (approved):

variableundefined (approved):
Safety override should be about weapons not Emag. I like it.

PR #14180, opened 2020-08-29
Allows the AI to speak all (common) languages

SabreML:
## What Does This PR Do

Allows the AI to speak all languages spoken by the station crew, the same way that Service Borgs can.

## Why It's Good For The Game

1. It seems odd that some Cyborgs can speak these languages, but the Station AI (presumably a lot more complex) can't.
You could make the argument that body language is required, but then *how is it spoken over the radio*. 🤔 

2. Less reason to use the translator.

3. It encourages more crew interaction or something

*I feel like there must be a very good reason why **not** to have this, considering how small the change is code wise; but I can't think of it.
I'm sure that if there is, someone will point it out.*

## Images of changes

Desktop Screenshot 2020 08 29 - 05 20 37 32
*Look at the pretty colours!*

## Changelog
:cl:
add: The station AI can now speak all languages, similar to Service Borgs
/:cl:


procdrone:
As far as my experience went, borgs and ai could only speak trinary. Will borgs also be able to speak common languages?

SabreML:
> As far as my experience went, borgs and ai could only speak trinary. Will borgs also be able to speak common languages?

This only changes the AI, so most borgs (other than service) will still have the same language options as before.

Galactic Common, Sol Common, Tradeband, Trinary and Binary.

Kyep (commented):
I'm not sure about this.
Yeah, in theory it makes sense for the AI to be able to speak any crew language.
In practice though:
- I worry this is going to encourage AIs to talk exclusively in non-common languages, which will be a pain in the ass.
- I also worry it allows powergamers to call for help in a rare language that their attacker won't understand, but the AI will.

SabreML:
@Kyep 
> * I worry this is going to encourage AIs to talk exclusively in non-common languages, which will be a pain in the ass.

That's a valid point. but I would argue that's more of a law break scenario.

(Easily fixed in Crewsimov with law 2, and probably breaks law 3 of Corporate)

(Possibly enters the territory of breaking Server Rule 5, but that's not my call.)

> * I also worry it allows powergamers to call for help in a rare language that their attacker won't understand, but the AI will.

This can happen even without the changes. The AI can already understand all languages, this just gives them the ability to *speak* them.

procdrone:
Man, as a borg I sure would wish I could speak all these languages sometimes.

Kyep (approved):

AffectedArc07 (approved):

PR #14181, opened 2020-08-29
Client BYOND Version Blacklisting+Warning

AffectedArc07:
## What Does This PR Do
**THIS IS NOT A SYSTEM DESIGNED TO FORCE USERS TO USE THE LATEST VERSION OF THE CLIENT, THIS IS MERELY FOR BLOCKING OUT VERSIONS WITH EXPLOITS, AND TO INFORM USERS IF THEY ARE USING A CLIENT WHICH IS UNSTABLE**

This PR adds an in-code way of notifying connecting clients if their BYOND version has issues on the server, and provides a way to outright blacklist versions which have exploits in. You can set a version to have either give a warning to the client and still let them play (This is for client builds with bugs which make the game unstable for the end uesr), or set it to blacklist the build entirely, and kick the user (This is for client builds which actively pose a threat to the game, be it exploits or other reasons).

The message appears 5 seconds after logging in, for 3 reasons
1: Gives the client time for the chat output to load so they can actually see why
2: It lets the message be after the MOTD so they can easily see it
3: The client window doesnt even appear until `client/New` completes fully, so you cant kick them instantly otherwise they wont know why

## Why It's Good For The Game
Ability to blacklist client versions with exploits is good, as well as informing users if they are using an ill-advised client version.

## Images of changes
*Please ignore spark being banned, I needed to test ban code a while ago and spark was the first key I thought of*

Normal connection
image

Warned connection
image

Blacklisted connection
image

## Changelog
:cl:
add: Added a way to warn clients if the BYOND build they are using is unstable
/:cl:

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -418,7 +418,7 @@
 	// This is on a timer so that the client's chat output can actually load
 	// spawn() is used here, so it can catch clients which connect before the MC has started
 	// I hate it too, dont worry -aa
-	spawn(50)
+	spawn(5 SECONDS)
 		byond_check()
 
 /client/proc/is_connecting_from_localhost()
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -295,14 +295,6 @@
 
 	if(connection != "seeker")					//Invalid connection type.
 		return null
-	if(byond_version < MIN_CLIENT_VERSION) // Too out of date to play at all. Unfortunately, we can't send them a message here.
-		return null
-	if(byond_build < config.minimum_client_build)
-		alert(src, "You are using a byond build which is not supported by this server. Please use a build version of atleast [config.minimum_client_build].", "Incorrect build", "OK")
-		qdel(src)
-		return
-	if(byond_version < SUGGESTED_CLIENT_VERSION) // Update is suggested, but not required.
-		to_chat(src,"<span class='userdanger'>Your BYOND client (v: [byond_version]) is out of date. This can cause glitches. We highly suggest you download the latest client from http://www.byond.com/ before playing. </span>")
 
 	if(IsGuestKey(key))
 		alert(src,"This server doesn't allow guest accounts to play. Please go to http://www.byond.com/ and register for a key.","Guest","OK")
@@ -422,6 +414,12 @@
 	if((playercount < threshold) && (GLOB.panic_bunker_enabled == TRUE))
 		GLOB.panic_bunker_enabled = FALSE
 		message_admins("Panic bunker has been automatically disabled due to playercount dropping below [threshold]")
+
+	// This is on a timer so that the client's chat output can actually load
+	// spawn() is used here, so it can catch clients which connect before the MC has started
+	// I hate it too, dont worry -aa
+	spawn(50)
+		byond_check()
 
 /client/proc/is_connecting_from_localhost()
 	var/localhost_addresses = list("127.0.0.1", "::1") // Adresses
--- a/code/modules/client/version_management.dm
+++ b/code/modules/client/version_management.dm
@@ -41,7 +41,7 @@
 	GLOB.flagged_byond_versions["1407"] = format_byond_version_info(CLIENT_BUILD_BLACKLISTED, "This client version contains a bug preventing display overrides from working, which leads to clients being able to see atoms they shouldn't be able to see")
 	GLOB.flagged_byond_versions["1408"] = format_byond_version_info(CLIENT_BUILD_BLACKLISTED, "This client version contains a bug preventing display overrides from working, which leads to clients being able to see atoms they shouldn't be able to see")
 	GLOB.flagged_byond_versions["1428"] = format_byond_version_info(CLIENT_BUILD_BLACKLISTED, "This client version contains a bug which causes right-click menus to show too many verbs")
-	GLOB.flagged_byond_versions["1530"] = format_byond_version_info(CLIENT_BUILD_WARNING, "This client version contains a bug which can cause crashes when near atoms with maptext (Brig cell timers for example)")
+	GLOB.flagged_byond_versions["1529"] = format_byond_version_info(CLIENT_BUILD_WARNING, "This client version contains a bug which can cause crashes when near atoms with maptext (Brig cell timers for example)")
 
 /**
   * Proc to check a client against flagged BYOND versions
--- a/code/modules/client/version_management.dm
+++ b/code/modules/client/version_management.dm
@@ -13,10 +13,6 @@
 	/// Details on why this build is warned/blacklisted
 	var/details
 
-/// Global list to keep track of BYOND version data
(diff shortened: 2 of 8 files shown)


code/modules/client/version_management.dm
```diff
@@ -0,0 +1,69 @@
...
+  */
+
+/client/proc/byond_check()
+	// Check their client minor version against the list of flagged ones
+	if("[byond_build]" in GLOB.flagged_byond_versions)
+		var/datum/byond_version/BV = GLOB.flagged_byond_versions["[byond_build]"]
+		switch(BV.response_type)
+			// Only warn them
+			if(CLIENT_BUILD_WARNING)
+				to_chat(src, "<span class='alert'><b>WARNING:</b> You are currently playing on a non-recommended client version ([byond_build]). You may experience the following issues: <span class='revenwarning'>[BV.details]</span>. You have been warned.</span>")
```
AffectedArc07, line 64: This class was chosen because its a different colour, and italics. I am not sure this is the best one to use, so you're all welcome to suggest something better.

code/modules/client/client procs.dm
```diff
@@ -423,6 +415,12 @@
 		GLOB.panic_bunker_enabled = FALSE
 		message_admins("Panic bunker has been automatically disabled due to playercount dropping below [threshold]")
 
+	// This is on a timer so that the client's chat output can actually load
+	// spawn() is used here, so it can catch clients which connect before the MC has started
+	// I hate it too, dont worry -aa
+	spawn(50)
```
dearmochi, line 421: nit: `5 SECONDS`

code/modules/client/version_management.dm
```diff
@@ -0,0 +1,69 @@
...
+  *
+  */
+/datum/byond_version
+	/// Response type (Either CLIENT_BUILD_WARNING or CLIENT_BUILD_BLACKLISTED)
+	var/response_type
+	/// Details on why this build is warned/blacklisted
+	var/details
+
+/// Global list to keep track of BYOND version data
+GLOBAL_LIST_EMPTY(flagged_byond_versions)
```
dearmochi, line 17: nit: I'd put it at the top of the file

code/modules/client/version_management.dm
dearmochi: Tangentially related - maybe wrap in `<big>`?

dearmochi (commented):
Few nitpicks. The presence of documentation makes for a happy mochi.

AffectedArc07:
New colour
image

goon/browserassets/css/browserOutput.css
```diff
@@ -325,7 +325,7 @@ h1.alert, h2.alert			{color: #000000;}
 .sans					{font-family: 'Comic Sans MS', cursive, sans-serif;}
 .wingdings				{font-family: Wingdings, Webdings;}
 .robot					{font-family: 'PxPlus IBM MDA'; font-size: 1.15em;}
-.ancient				{color: #008B8B;	font-stye: italic;}
```
AffectedArc07, line 328: So turns out this style has been broken for 4 years. huh.

farie82: true to it's name

code/modules/client/version_management.dm
```diff
@@ -0,0 +1,65 @@
...
+  * Proc to register all flagged BYOND versions
+  *
+  * This proc is called from world/New(), and is for registering BYOND versions
+  * Add versions to flag into this proc, following the format of the others
+  * Please keep these in build version order. Please.
+  * VERSION NUMBER MUST BE A STRING BECAUSE OTHERWISE IT USES INDEXES NOT ASSOCIATIONS
+  */
+/proc/register_flagged_byond_versions()
+	// Do not put a . at the end of the details, otherwise it will double-period when the client sees it
+	GLOB.flagged_byond_versions["1407"] = format_byond_version_info(CLIENT_BUILD_BLACKLISTED, "This client version contains a bug preventing display overrides from working, which leads to clients being able to see atoms they shouldn't be able to see")
```
Kyep, line 41: Shouldn't this be set in config? Otherwise we have to do PRs whenever we need to add a new blacklisted version. Also, these reasons need to be understandable by normal players in addition to having enough info for admins/etc to realize how bad the problem is.

code/modules/client/version_management.dm
```diff
@@ -0,0 +1,65 @@
...
+  */
+
+/client/proc/byond_check()
+	// Check their client minor version against the list of flagged ones
+	if("[byond_build]" in GLOB.flagged_byond_versions)
+		var/datum/byond_version/BV = GLOB.flagged_byond_versions["[byond_build]"]
+		switch(BV.response_type)
+			// Only warn them
+			if(CLIENT_BUILD_WARNING)
+				to_chat(src, "<span class='alert'><b>WARNING:</b> You are currently playing on a non-recommended client version ([byond_build]). You may experience the following issues: <span class='ancient'>[BV.details]</span>. You have been warned.</span>")
```
Kyep, line 60: "Non-recommended"? Maybe something like "Your BYOND version (X) is known to have these issues: Y. We strongly suggest upgrading your BYOND version before playing."
This also doesn't have a link for where to download the latest version of byond (said link should be a config directive too, so that we can hardlink to a specific version in the event that the latest byond version is unstable).

code/modules/client/version_management.dm
```diff
@@ -0,0 +1,65 @@
...
+			// Client is blacklisted, warn and kick
+			if(CLIENT_BUILD_BLACKLISTED)
+				to_chat(src, "<span class='alert'><b>ALERT:</b> You are currently playing on a blacklisted client version ([byond_build]). The reasoning for this is: <span class='ancient'>[BV.details]</span>. Please update your client. You have been kicked from the server.</span>")
```
Kyep, line 63: Again, this needs a link to where they can get the most recent version. Links in chat may not work for players that are kicked before they click the link, so you may need to just open the URL in a browser.

Kyep (changes requested):
I do think a system for flagging byond versions would be useful, but:
- If we're going to use a system that flags individual byond versions, then the list of those blacklisted versions MUST be defined by a config file, because otherwise we're forcing ourselves to do a new PR every time we want to blacklist a version, which is unnecessary extra effort.
- I'm kinda skeptical about the idea of using specific blacklisted versions at all. While I suppose it does make sense for versions with active exploits, I'd prefer that, most of the time, we just set the 'minimum' byond version a bit more aggressively and actively prompt people to upgrade more. There will always be issues with byond that are fixed by upgraded versions... forcing ourselves to effectively do byond version management per-release, rather than saying "please use at least version X" is unnecessarily time-consuming.
- There also needs to be the concept of "max version" or "latest byond version we support". For times like now when the current stable build of byond has crippling bugs.

AffectedArc07:
I'll do a rewrite of this in a different way at some point, one which requires less maintenance.

Spacemanspark:
Time to ban evade with my windows XP netbook

PR #11575, opened 2019-05-22
The Foodening (Adjusts nutriment factors)

Shadeykins:
**What does this PR do:**
_**It's not even real chicken anyways.**_

**The Controversial:**
Hot Coco nutriment factor 3 -> 0 (full base nutriment reduction. Coco decays into sugar which provides its own nutrition. Hot Chocolate has been changed to use hot_coco as well).
Chicken Soup nutriment factor 2 -> 0.5 (75% reduction)

Why? The convenience of these items is so high due to their values (1 cup from nothing to full) that they disincentive people from ever engaging with other food sources (even other vending machine items). Ideally they could have a junkiness factor like other vending machine food, but this is not viable due to them being reagents. While they can still be abused (simply drink more), they are less convenient insofar as the metabolic rates are already relatively low (you will need to wait longer to get your bang for your buck whereas other foods satiate quicker).

**The (Hopefully) Uncontroversial:**
Protein (found in products containing meat) provides a greater satiety over nutriment (15 -> 20)
Plantmatter (found in products containing plants) provides a greater satiety over nutriment (15 -> 18)

Why? Both of these are more commonly found in chef's food. While chef's food already has higher nutritional values (and also contains vitamins which heal damage and restore blood), this is further weighting to the efficacy of the chef's food. There are also some types of junkfood that got a buff due to this (pistachios, jerky).

**The Potentially Controversial:**
Vitamins (found only in chef's food) have had their healing effects doubled. There is a 50% chance per tick of metabolism for vitamins to heal 1 brute/burn damage and restore 0.5 blood. This has been doubled to 2/2 brute/burn and 1 blood (the chance remains the same).

**The Goal**
We have awful chefs/botanists quite frequently. While a lot of this can be attributed to it being a starter role, I think a fair amount is that these roles are also easily made irrelevant through more reliable mechanisms (vending machines). There is no pressure for anyone in these positions to do their job nor any reason for the HoP to fire them (unless they're being borderline griefy). By making them have the ability to offer out-and-out superior options for satiety and clawing back some other convenient sources, there should be a natural predilection for people to scrutinize players in these roles more frequently.

This should, in turn, lead to more people using the chef and thereby increase the general quality of chefs as demand rises.

**Changelog:**
:cl:
tweak: Hot Chocolate/Hot Coco contains no base nutriment, only the nutrition from sugar decay.
Hot Chocolate changed to use hot_coco reagent, not pure liquid chocolate.
Chicken Soup 75% less effective.
Protein 33% more effective.
[...]

PR #14185, opened 2020-08-29
Hot Chocolate and Chicken Soup nerf

Cocacolagua:
## What Does This PR Do
Nerfs Hot Chocolate to have 15 units of hot chocolate, 6 units of chocolate and 9 units of water instead of 30 units of pure chocolate. (from 108 nutrition total to 36)
Nerfs Chicken Soup by making the chemical have a metabolism rate, but buffs it's nutrition factor to 2.5 (from 60 nutrition total to 30)

## Why It's Good For The Game
You shouldn't be able to buy a couple of drinks for around 100$ and not worry about hunger ever again, which in return ends with the other vendor items not being bought as much, or the chef being ignored (you can still buy items from the vendors and ignore the chef, but now you'll need to buy more than two for the entire round)

## Changelog
:cl:
tweak: Hot Chocolate now is composed of fifteen units of Hot Chocolate, six units of Chocolate and nine units of Water instead of thirty units of Chocolate.
tweak: Chicken Soup now has a metabolism rate, and it's nutriment factor is now 2.5
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/drinks/drinks.dm
+++ b/code/modules/food_and_drinks/drinks/drinks.dm
@@ -226,7 +226,7 @@
 	desc = "Made in Space Switzerland."
 	icon_state = "hot_coco"
 	item_state = "coffee"
-	list_reagents = list("hot_coco" = 30)
+	list_reagents = list("hot_coco" = 15, "chocolate" = 6, "water" = 9)
 	resistance_flags = FREEZE_PROOF
 
 /obj/item/reagent_containers/food/drinks/weightloss
--- a/code/modules/food_and_drinks/drinks/drinks.dm
+++ b/code/modules/food_and_drinks/drinks/drinks.dm
@@ -226,7 +226,7 @@
 	desc = "Made in Space Switzerland."
 	icon_state = "hot_coco"
 	item_state = "coffee"
-	list_reagents = list("chocolate" = 30)
+	list_reagents = list("hot_coco" = 30)
 	resistance_flags = FREEZE_PROOF
 
 /obj/item/reagent_containers/food/drinks/weightloss
--- a/code/modules/reagents/chemistry/reagents/food.dm
+++ b/code/modules/reagents/chemistry/reagents/food.dm
@@ -628,7 +628,7 @@
 	reagent_state = LIQUID
 	color = "#B4B400"
 	metabolization_rate = 0.2
-	nutriment_factor = 2
+	nutriment_factor = 2.5 * REAGENTS_METABOLISM
 	taste_description = "broth"
 
 /datum/reagent/consumable/cheese


trololiver112:
Muh powergaming 😳

Cocacolagua:
muh powergaming indeed

kenken244:
this makes fliptonium very difficult to make now, since it requires actual chocolate. how about instead making it 10u chocolate and 20u water? or something similar

Mitchs98:
I'll say the same thing I did when this has been brought up before. If we had less bald chefs, chefs served more food my character preferres, or if the food wasn't consistently drugged, I'd see no problem in this. This is the main reason most people use hot chocolate and chicken soup as there is no difference in grabbing food from the chef counter, other than you'll probably be drugged with _insert thing here_.

That said I don't agree with the changes for these reasons, at the same time they aren't as heavy a nerf as before...sort of. 0.2 symbolization rate is immensely slow. I just don't see a reason to change it, especially now that you can actually starve to death as a mechanic.

Cocacolagua:
You can still choose to ignore the chef entirely, but you will have to purchase more products from the vendors than before.

If we /ever/ want to have a semi-functional economy, we do want alternatives to services offered from some departments (bartender, chef, for example), but those alternatives either have to be not time-efficient (making your own food with maintenance machines) to the real thing, or more expensive (buying from the vendors) than the real thing.

Kyep (changes requested):
Objection, for the same reasons as the last time this idea was brought up: https://github.com/ParadiseSS13/Paradise/pull/11575
Forcing people to rely on the chef is bad. Half the time, there is no chef. Either they're dead, they're busy being bald in maintenance, they're SSD, or they're otherwise unavailable. Even when they are available, there's no guarantee that they're making edible food. Much of the time they don't make anything, they make exclusively spider legs, or someone poisons their food.
Chef food is simply not reliable enough to force everyone to depend on it all the time.

Cocacolagua:
And I will give the same rebuttal that the author gave in #11575. The alternative from the vendors is still there, this PR does not remove it, it only makes it less efficient than before. You can buy enough food to pass you through the shift after this nerf, even if the economy RNG gods decided you should start the shift with around 500$. The chef /should/ be more efficient at satisfying people than food from the vendors. (hell, even /if/ there isn’t chef food, you can get food from botany, cargo, xenobiology, or even sugar from a chemical dispenser)

Kyep:
I remain unconvinced this is a good idea. I'm willing to TM it or evaluate it further, but I'm still very skeptical about it.

ExusA:
The fact that "Hot Chocolate", the drink, actually contains no "hot chocolate" the reagent is confusing as heck. The fact that "Hot Coco", a different drink, DOES contain the reagent, "Hot Chocolate", is even more confusing. It'd be nice if it were changed for that reason alone.

TDSSS:
If we are 100% committed to having a drink in vending machines that completely eliminates all need for food, then we might as well remove the hunger mechanic entirely or make everyone gain a nutrient implant at roundstart.

Yes, chef food is unreliable. That doesn't mean we should keep it completely unneccessary. If the crew is incompetent enough to not manage their food supply, then they deserve to suffer the consequences of it until they set something up again.

Mitchs98:
> If the crew is incompetent enough to not manage their food supply, then they deserve to suffer the consequences of it until they set something up again.

I'd be fine with this if you couldn't starve to death in a two hour shift span.

taukausanake:
> I'd be fine with this if you couldn't starve to death in a two hour shift span

I don't know, that sounds right up there in how SS13 should be. You're expected to die but how is up to the round and the competency of the crew. I can see a round of extended, no antags or dangerous events going on, and suddenly the kitchen shuts down. Botany is gone. The garden has all its soil removed. There is no way for the station to locally product food. They resort to buying from the vendors, ordering things from cargo, trying to survive until their shift ends. That to me sounds fun but last I remember Paradise doesn't like fun

Cocacolagua:
changed the hot chocolate drink composition to be 6 units of chocolate, 15 units of hot chocolate and 9 units of water, the nutrition value is still 24, reason for this is so you can use this chocolate for recipes that require it, like fliptonium. updating the PR description now.

Denthamos:
@Mitchs98 Starving to death is not a feature here and has never been.  You just get slowed down due to being hungry and having less energy.

I still believe the fix to the whole chef issue in terms of the food not being safe is porting in VG's vending machines, and slapping one down right at the chefs window/table to the outside so he can put food inside of it and attach a price tag to it, allowing him to sell his food to the crew.

That way the chef has incentive to make food for the crew as he gets a reward from it, and the normal crew cannot tamper with it either.

Cocacolagua:
#13098 @Denthamos

Cocacolagua:
Correction: I just remembered that for each unit of chocolate you drink, your body makes 0.8 units of sugar, and both chocolate and sugar have the same nutritional value, so current hot chocolate drink isn't actually 60 nutrition, but 108!

Mitchs98:
> @Mitchs98 Starving to death is not a feature here and has never been. You just get slowed down due to being hungry and having less energy.
> 
> I still believe the fix to the whole chef issue in terms of the food not being safe is porting in VG's vending machines, and slapping one down right at the chefs window/table to the outside so he can put food inside of it and attach a price tag to it, allowing him to sell his food to the crew.
> 
> That way the chef has incentive to make food for the crew as he gets a reward from it, and the normal crew cannot tamper with it either.

Seems the PR was already linked for me. As far as your idea, I'd be inclined to agree. And if that was added I'd be more inclined to be alright with this PR going through. Until then this is another case of a feature that should be added and probably won't be until the far future.

TDSSS:
Here's my prediction on how a testmerge will go:
75% of players won't notice, 24% will just buy two cups of chicken soup instead of one, 1% will complain about starving before gorging on random vendor foot until they're fat.

More seriously, I don't think most people will really notice.

farie82 (approved):

Kyep (commented):
Retracting my objection from this, because:
- It does seem reasonable that hot chocolate actually contains the hot_coco reagent
- It also seems reasonable that chicken soup considers the speed at which you metabolize reagents, and balance-wise while I am not a fan of forcing people to rely more on the chef, everyone pretty much recognizes that chicken soup is vastly stronger than other vendor items, to the point that it renders the other vendor items useless. 

If this change starts to result in crew starving or the station regularly having food issues, then I'll consider buffing some other vendor food instead, but for the moment, I am willing to try this, and see how it works out in practice.

AffectedArc07 (approved):

PR #14186, opened 2020-08-29
Reworks Baseball Bat

Kyep:
## What Does This PR Do
Nerfs the knockback ability of baseball bats, so that it:
- doesn't apply to mobs which are bigger than you, flagged as immune to pushing, or flagged as having higher than normal resistance to being moved
- can only trigger at most once per ten seconds

## Why It's Good For The Game
Right now, baseball bats are overpowered. They're so overpowered I still see sec carrying them during red and gamma alert, when much better weapons should be available.
They're overpowered because of their knockback ability. It throws almost any mob, even including things like megafauna, on every hit.
This is abusable in many ways, such as:
- Cheesing NPCs by batting them away every time they get close, so they can never stay next to you for long enough for them to hit you. So you can beat each NPC without taking any hits at all.
- Spamming it on players in order to block them from acting, because they're constantly being thrown and moved.
- Using it to force-move mobs that are not meant to be movable.

## Changelog
:cl: Kyep
tweak: baseball bats can no longer knockback mobs that are meant to be push-resistant, and can now only get knockback effects once per ten seconds.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/weaponry.dm
+++ b/code/game/objects/items/weapons/weaponry.dm
@@ -181,10 +181,11 @@
 	item_state = "baseball_bat"
 	var/deflectmode = FALSE // deflect small/medium thrown objects
 	var/lastdeflect
-	force = 18
+	force = 10
 	throwforce = 12
 	attack_verb = list("beat", "smacked")
 	w_class = WEIGHT_CLASS_HUGE
+	var/nextthrow_time = 0
 	var/homerun_ready = 0
 	var/homerun_able = 0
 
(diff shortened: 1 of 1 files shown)


ppi13:
Dunno about the damage buffs, but being unable to solo warops and terror spiders and the likes is a welcoming change.

farie82 (commented):
The `/obj/item/melee/baseball_bat/ablative` version is now weaker. It was supposed to be slightly stronger it seems.

It's also quite strong now. Same damage output as an edagger and more than the chainsaw. All the while it's a legal weapon (I believe)

Cocacolagua:
Possession of a weapon does say "To be in possession of a dangerous item that is not part of one's job", so unless you’re playing baseball, you can be charged for it

Qwertytoforty:
@farie82 chainsaw has a force of 24 on. Spear also has a force of 18, and is 2 handed, but spears are sharp, and can be thrown

Kyep:
@farie82 The ablative bat has had its damage increased from 12 to 20, consistent with how the standard wooden bat gained 8 damage. The PR description has also been updated to reflect the fact it now applies to the ablative bat as well.

ppi13:
baseball has brute damage, which means that if it targets head it has a chance of a 5 second stun with a 15 second confusion, increasing the force from 10 to 18 is nearly doubling the probability of a stun.

Kyep:
That's okay, as its consistent with other brute damage melee weapons, and its miles better than a guaranteed throw.

Denthamos:
I do agree, bats have been needing the....nerfbat for some time due to simply how strong they were as a melee weapon.  However loosing a key feature of them that has been with them since their incarnation feels....well, kind of sad.

Would it be possible to keep their knocking away feature, but have it work like the home run bats charge up time?  Where you have to stand still, fully charge up the timer above your head, and THEN you can knock someone back with a hit but only once?

Kyep:
@Denthamos The homerun bat is not a good example, because its an incredibly OP adminspawn-only item. Once charged, it keeps its charge forever as you walk around, meaning that in practice there is no reason not to just keep it charged all the time. It also hits with the force of an explosion. And throws you a long distance. We shouldn't be basing anything on it.
To give you a few examples:

1. If the bat loses its charge when you move, then humans are just going to push you to depower your bat.
2. If the bat doesn't lose its charge when you move, then there's no reason not to keep it charged all the time.
3. If humans can deal with the bat more easily than NPCs can, congrats, you've invented a new way to cheese NPCs. People already cheese NPCs with the bat by hitting the NPC back away from them before the NPC can attack, thus forcing the NPC to never stay in melee range for long enough to land an attack. This cheesy method lets you kill many hostile NPCs without taking a single hit.
4. If it takes special care/preparation to use it properly, then its a lot less useful to your average assistant (its intended user) than a bat with no special tricks but higher base damage. There are places in the game for stuff that takes considerable prep / knowledge to use... but random maint items for civilians is probably not it.
So... for all these reasons, no, I think its better to just do what this PR does. Give it higher base damage and remove its knockback.

Aurorablade:
gonna have to agree with denth...adding some sort of delay to it or charge or cool down is much better then making it hit harder.  Its just going to cause another issue down the line.

Kyep:
**This PR now limits knockback instead of removing it, and no longer increases the bat's damage.**
Now, the limits placed on knockback are:
- can only trigger at most once per ten seconds (so you can't spam it)
- won't work on mobs meant to be immune to pushing/etc (so you can't use it to casually bat around megafauna)

Spacemanspark:
Honestly not sure how I feel about the brute damage buff going up quite that high, but I can look past that if the dumb tossing mechanic is gutted tbh

code/game/objects/items/weapons/weaponry.dm
```diff
@@ -185,6 +185,7 @@
 	throwforce = 12
 	attack_verb = list("beat", "smacked")
 	w_class = WEIGHT_CLASS_HUGE
+	var/nextthrow_time = 0
```
farie82, line 188: Minor thing but can you make this full snake_case? so something like `next_throw_time`. Reads just a bit better

farie82 (commented):
Tested. Works as expected. Only one minor styling remark

Kyep:
@farie82 Done

farie82 (approved):

variableundefined (approved):

PR #14195, opened 2020-08-30
Makes gygax take damage from mecha overload again

Qwertytoforty:
## What Does This PR Do

This PR re-adds gygaxs taking damage from leg overload, which was removed with object damage. This PR also prevents 1 damage attacks from ever displaying TAKING DAMAGE! to the user in mech overload mode, to prevent them from being spammed with messages.

## Why It's Good For The Game

Normally, gygaxs are limited currently in speed mode by their cell. The issue is when botany gets 400000 charge cells, and the gygax can effectively never run out of power in speed mode, able to run at full speed constantly with no catch. By re-adding the damage catch to leg overload, you have to consider if the speed boost now is worth it, as with too much damage you wont be able to speed boost again till fixed.

## Changelog
:cl:
tweak: Re-added the feature of mechs taking damage from mech overload.
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -504,8 +504,9 @@
 				check_for_internal_damage(list(MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH,MECHA_INT_CONTROL_LOST))
 			else
 				check_for_internal_damage(list(MECHA_INT_FIRE,MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH,MECHA_INT_CONTROL_LOST,MECHA_INT_SHORT_CIRCUIT))
-		if((. != 1 && leg_overload_mode) && (. >= 5 || prob(33))) //If it takes 1 damage and leg_overload_mode is true, do not say TAKING DAMAGE! to the user several times a second.
-			occupant_message("<span class='userdanger'>Taking damage!</span>")
+		if(. >= 5 || prob(33)) //If it takes 1 damage and leg_overload_mode is true, do not say TAKING DAMAGE! to the user several times a second.
+			if(!(. != 1 && leg_overload_mode))
+				occupant_message("<span class='userdanger'>Taking damage!</span>")
 		log_message("Took [damage_amount] points of damage. Damage type: [damage_type]")
 
 /obj/mecha/run_obj_armor(damage_amount, damage_type, damage_flag = 0, attack_dir)
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -331,6 +331,8 @@
 		else
 			occupant.clear_alert("mechaport")
 	if(leg_overload_mode)
+		log_message("Leg Overload damage.")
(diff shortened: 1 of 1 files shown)


code/game/mecha/mecha.dm
```diff
@@ -331,6 +331,8 @@
 		else
 			occupant.clear_alert("mechaport")
 	if(leg_overload_mode)
+		log_message("Leg Overload damage.")
+		take_damage(1, BRUTE, 0, 0)
```
SteelSlayer, line 335:
```suggestion
		take_damage(1, BRUTE, FALSE, FALSE)
```

code/game/mecha/mecha.dm
```diff
@@ -502,7 +504,7 @@
 				check_for_internal_damage(list(MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH,MECHA_INT_CONTROL_LOST))
 			else
 				check_for_internal_damage(list(MECHA_INT_FIRE,MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH,MECHA_INT_CONTROL_LOST,MECHA_INT_SHORT_CIRCUIT))
-		if(. >= 5 || prob(33))
+		if((. != 1 && leg_overload_mode) && (. >= 5 || prob(33))) //If it takes 1 damage and leg_overload_mode is true, do not say TAKING DAMAGE! to the user several times a second.
```
SteelSlayer, line 507: This won't work as intended currently. 
For example, given `. = 10` and `leg_overload_mod = FALSE`, we'll get:
```dm
if((TRUE && FALSE) && TRUE)
```
Which of course equates to:
```dm
if(FALSE && TRUE)
```
Which means that having leg overload mode off will prevent valid messages from appearing.

code/game/mecha/mecha.dm
```diff
@@ -502,8 +504,9 @@
 				check_for_internal_damage(list(MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH,MECHA_INT_CONTROL_LOST))
 			else
 				check_for_internal_damage(list(MECHA_INT_FIRE,MECHA_INT_TEMP_CONTROL,MECHA_INT_TANK_BREACH,MECHA_INT_CONTROL_LOST,MECHA_INT_SHORT_CIRCUIT))
-		if(. >= 5 || prob(33))
-			occupant_message("<span class='userdanger'>Taking damage!</span>")
+		if(. >= 5 || prob(33)) //If it takes 1 damage and leg_overload_mode is true, do not say TAKING DAMAGE! to the user several times a second.
+			if(!(. != 1 && leg_overload_mode))
```
SteelSlayer, line 508: I think it should be `. == 1` because, taking the previous example:
Let `. = 10` and `leg_overload_mode = TRUE`
```dm
if(TRUE)
    if(!(TRUE && TRUE)) // <- if(!TRUE) does not pass, message is not shown.
```
Also you can still one-line this:
```dm
if((. >= 5 || prob(33)) && !(. == 1 && leg_overload_mode))
```

AffectedArc07 (approved):

PR #14196, opened 2020-08-30
Adds the proximity monitor component. Performance improvement for turf/Entered

SteelSlayer:
## What Does This PR Do
Adds the proximity monitior component, which can be used for atoms which need to react to things moving with 1 tile of them. This is inspired by TG's `proximity_monitor` datum and `proximity_checker` objects, so credit to them.

Basically how it works is that, there are 9 objects that get created around the parent object. Whenever `Crossed()` is called on these objects, they call `HasProximity()` on the parent object. I won't go into a lengthy explaination of how the component works here, because I have, hopefully, written enough comments in the code to make it easier to understand. Plus the gif below should also help. 

This is a solution to the fairly expensive way in which we handle proximity right now, which is this code in `turf/Entered()`:
```dm
	var/loopsanity = 100
	for(var/atom/A in range(1))
		if(loopsanity == 0)
			break
		loopsanity--
		A.HasProximity(M)
```
This is very inefficient and runs on *everything* regardless of if an atom cares about other things being in proximity of it, which is not a whole lot of things currently.

Now onto the juicy profiler numbers.
```
Proc Name                                             				     Self CPU    Total CPU    Real Time     Overtime        Calls
----------------------------------------------------------------------------------  ---------    ---------    ---------    ---------    ---------
**Paradise proximity_monitor branch**
/datum/component/proximity_checker/proc/HandleMove                                      0.005        0.006        0.010        0.005        40000
/datum/component/proximity_checker/proc/recenter_prox_checkers                          0.251        0.428        0.435        0.251        40000
/obj/effect/abstract/proximity_checker/proc/HandleMove                                  0.169        0.174        0.206        0.167       360000
/turf/Entered                                                                           0.032        0.044        0.049        0.032        40082

**Paradise master branch**
/turf/Entered                                                                           1.594        1.775        1.778        1.589        40019

**TG master branch**
/datum/proximity_monitor/proc/HandleMove                                                0.070        1.631        1.634        0.070        40000
/datum/proximity_monitor/proc/SetRange                                                  0.166        1.561        1.566        0.164        40000
/turf/open/Entered                                                                      0.094        0.286        0.292        0.090        40282
```

Keep in mind you probably wont see a proximity sensing object move or teleport 40,000 times in a "normal" round. And that these numbers were just for stress testing.

Also as a comparison for how many `turf/Entered()` calls there can be in a typical round:
```
/turf/Entered                                          19.387       21.325       21.399        0.441       708954
```
This was profiled from the live server, by Farie, like a month ago. The round length was about an hour and half I believe.

#### Basic movement testing
On my branch, I took an object which has the proximity component on it, and called `Move()` on it 40,000 times. This simulates an an object or mob moving 1 tile in any direction, each time. The cost of that, component-side, is the result of adding the two `HandleMove` proc costs together, which is `0.006 + 0.174` = `0.180`. And adding in the `0.044` cost of `turf/Entered`, you get a total cost `0.224`.

This was also done on both the current master branch of Paradise and on TG for comparison. For both of these instances, an object had `Move()` called on it 40,000 times. In TG's case it was an object which uses the `proximity_monitor` datum. As you can see, the implementation I'm currently using is a lot cheaper than Paradise's or TG's current implementation.

#### Non-standard movement test
The other test I did, was calling `recenter_prox_checkers`, which is a proc that fires when something moves in non standard way. This is separate from the basic `Move()` tests above. This handles things like teleporting, getting picked up or dropped, the beginning of a throw, getting placed onto a table/into a closet and so on. The cost of this proc is the most expensive, but still relatively cheap overall.

## Why It's Good For The Game
*Drastically* cuts down on the amount of CPU cost of the `turf/Entered()` proc.

Also fixes an unlisted bug: Makes the portable flasher's unwrenching message visible. Previously, it would never display.

## Images of changes
Visualization of the proximity checker objects that follow the parent. All the "DJ" area sprites are the proximity checkers (sprite won't be visible in the actual game of course)

proxchecker-min

As you can see, there are 9 objects that need to move every time the main parent object needs to move.

## Changelog
:cl:
fix: Makes the portable flasher's unwrenching message visible
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1997,5 +1997,26 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 		return TRUE
 	return FALSE
 
+/**
+ * Proc which gets all adjacent turfs to `src`, including the turf that `src` is on.
+ *
+ * This is similar to doing `for(var/turf/T in range(1, src))`. However it is slightly more performant.
+ * Additionally, the above proc becomes more costly the more atoms there are nearby. This proc does not care about that.
+ */
+atom/proc/get_all_adjacent_turfs()
+	var/turf/src_turf = get_turf(src)
+	var/list/_list = list(
+		src_turf,
+		get_step(src_turf, NORTH),
+		get_step(src_turf, NORTHEAST),
+		get_step(src_turf, NORTHWEST),
+		get_step(src_turf, SOUTH),
+		get_step(src_turf, SOUTHEAST),
+		get_step(src_turf, SOUTHWEST),
+		get_step(src_turf, EAST),
+		get_step(src_turf, WEST)
+	)
+	return _list
+
 /// Waits at a line of code until X is true
 #define UNTIL(X) while(!(X)) stoplag()
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -2003,7 +2003,7 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
  * This is similar to doing `for(var/turf/T in range(1, src))`. However it is slightly more performant.
  * Additionally, the above proc becomes more costly the more atoms there are nearby. This proc does not care about that.
  */
-atom/proc/get_all_adjacent_turfs()
+/atom/proc/get_all_adjacent_turfs()
 	var/turf/src_turf = get_turf(src)
 	var/list/_list = list(
 		src_turf,
--- a/code/datums/components/proximity_monitor.dm
+++ b/code/datums/components/proximity_monitor.dm
@@ -1,0 +1,140 @@
+/**
+ * # Proximity monitor component
+ *
+ * Attaching this component to an atom means that the atom will be able to detect mobs/objs moving within a 1 tile of it.
+ *
+ * The component creates several `obj/effect/abstract/proximity_checker` objects, which follow the parent atom around, always making sure it's at the center.
+ * When something crosses one of these `proximiy_checker`s, the parent has the `HasProximity()` proc called on it, with the crossing mob/obj as the argument.
+ */
+/datum/component/proximity_monitor
+	var/atom/owner
+	/// A list of currently created `/obj/effect/abstract/proximity_checker`s in use with this component.
+	var/list/proximity_checkers
+
+/datum/component/proximity_monitor/Initialize()
+	. = ..()
+	if(!isatom(parent))
+		return COMPONENT_INCOMPATIBLE
+	owner = parent
+	create_prox_checkers()
+
+/datum/component/proximity_monitor/Destroy(force, silent)
+	QDEL_LIST(proximity_checkers)
+	owner = null
+	return .. ()
+
+/datum/component/proximity_monitor/RegisterWithParent()
+	. = ..()
+	if(ismovable(parent))
+		RegisterSignal(parent, COMSIG_MOVABLE_MOVED, .proc/HandleMove)
+
+/datum/component/proximity_monitor/UnregisterFromParent()
+	. = ..()
+	if(ismovable(parent))
+		UnregisterSignal(parent, COMSIG_MOVABLE_MOVED)
+
+/**
+ * Called when the `parent` receives the `COMSIG_MOVABLE_MOVED` signal, which occurs when it `Move()`s
+ *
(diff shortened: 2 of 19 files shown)


AffectedArc07:
```
🆑
fix: Makes the portable flasher's unwrenching message visible
/🆑
```
Did you get the right CL here? This isnt mentioned at all in the PR

code/__HELPERS/unsorted.dm
```diff
@@ -2022,5 +2022,26 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 		return TRUE
 	return FALSE
 
+/**
+ * Proc which gets all adjacent turfs to `src`, including the turf that `src` is on.
+ *
+ * This is similar to doing `for(var/turf/T in range(1, src))`. However it is slightly more performant.
+ * Additionally, the above proc becomes more costly the more atoms there are nearby. This proc does not care about that.
+ */
+atom/proc/get_all_adjacent_turfs()
```
AffectedArc07, line 2031:
```suggestion
/atom/proc/get_all_adjacent_turfs()
```

AffectedArc07 (commented):
I am by no means a component expert, so someone else can have this one

SteelSlayer:
Yeah, I did include this fix in my PR just because its like 1 line thing. I can remove it if you want.

AffectedArc07:
Its fine to keep in, just mention it a little more

SteelSlayer:
@AffectedArc07 Added it in the description. Also, just review the non-component code :smirk:

code/datums/components/proximity_monitor.dm
```diff
@@ -0,0 +1,140 @@
...
+	. = ..()
+	if(!isatom(parent))
+		return COMPONENT_INCOMPATIBLE
+	owner = parent
+	create_prox_checkers()
+
+/datum/component/proximity_monitor/Destroy(force, silent)
+	QDEL_LIST(proximity_checkers)
+	owner = null
+	return .. ()
```
farie82, line 24: spacing here

code/datums/components/proximity_monitor.dm
```diff
@@ -0,0 +1,140 @@
...
+ */
+/obj/effect/abstract/proximity_checker/Crossed(atom/movable/AM, oldloc)
+	set waitfor = FALSE
+	if(active)
+		hasprox_receiver.HasProximity(AM)
+
+/**
+ * Moves the proximity_checker 1 tile in the `Dir` direction. If `Dir` is null, it doesn't change location.
+ *
+ * If the new location of the receiver is NOT a turf, set `active` to FALSE, so that it does not receive proximity calls.
```
farie82, line 122: It suggests that this is part of the logic whether or not Dir is null. Which is not the case since if Dir is not null it won't do this

farie82 (changes requested):

code/datums/components/proximity_monitor.dm
SteelSlayer: Ah you are referring to the fact that it will teleport if dir is null right?

code/datums/components/proximity_monitor.dm
SteelSlayer: Yep

code/datums/components/proximity_monitor.dm
```diff
@@ -0,0 +1,140 @@
...
+ * * forced - if we were forced to move
+ */
+/obj/effect/abstract/proximity_checker/proc/HandleMove(datum/source, atom/OldLoc, Dir, forced)
+	if(Dir)
+		loc = get_step(src, Dir) // Basic movement 1 tile in some direction.
+		return
+	if(!isturf(hasprox_receiver.loc))
+		active = FALSE // Receiver shouldn't detect proximity while in a backpack, closet, etc.
+		return
+	if(isturf(hasprox_receiver.loc))
```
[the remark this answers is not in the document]
SteelSlayer: Yeah, that was silly of me. Leaving it as a basic if else *for now* since there are only 2 conditions.

code/datums/components/proximity_monitor.dm
farie82: Nah meant that it will return if Dir is not null. Thus not set the `active` variable in the same proc call

SteelSlayer: Ah gotcha

SteelSlayer: I just realized I have it return for a reason. If `Dir` is not null, its just a standard movement in 1 tile. It *will* be on a turf, therefore I don't want to keep checking if its on a turf or not.

PR #14198, opened 2020-08-30
Buffs Lightning Guardian punch damage, makes guardian and host shock proof

Qwertytoforty:
## What Does This PR Do

This PR buffs the Lightning guardian in 4 ways. Improves its base punch damage from 7-12, improves its armor from 30% to 40%, has the guardian give its host immunity to shocks, and finally prevents the guardian from being hurt by shocks.

## Why It's Good For The Game

The lightning guardian has a few issues. It's entirely burn damage, which compared to brute, is not that good. It does do much more damage when fighting a group of people, but those people really need to be close together to take the most damage, and holoparas are often not built for group fights. Unlike most other heavy attacking guardians, it had lower armor, so its armor was increased slightly to make it better in close combat. The burn buff to its punches will help it for one on one combat.

Shock immunity was more of a stylistic change, based on how the chaos guardian makes you fire immune, its good for the lightning guardian to be immune to shocks, as well as the host. It also makes this guardian the only one that can be safely used with the telsa released, as even if the user had full shock protection, the guardian would be shocked if deployed, and die instantly from the tesla arc. 

Overall, this should make the lightning guardian easier to use, actually chosen, and give it the bonus of working well in a tesla hijack scenario. 
## Changelog
:cl:
tweak: Buffs the Lightning guardians damage and armor, and makes it and the host immune to shocks.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/guardian/types/lightning.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/lightning.dm
@@ -18,7 +18,7 @@
 	var/list/enemychains = list()
 	var/successfulshocks = 0
 
-/mob/living/simple_animal/hostile/guardian/beam/New(loc, mob/living/user)
+/mob/living/simple_animal/hostile/guardian/beam/New(loc, mob/living/user, mob/living/user)
     . = ..()
     if(!user)
         return
--- a/code/game/gamemodes/miniantags/guardian/types/lightning.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/lightning.dm
@@ -18,7 +18,7 @@
 	var/list/enemychains = list()
 	var/successfulshocks = 0
 
-/mob/living/simple_animal/hostile/guardian/beam/New(loc, mob/living/user, mob/living/user)
+/mob/living/simple_animal/hostile/guardian/beam/New(loc, mob/living/user)
     . = ..()
     if(!user)
         return
--- a/code/game/gamemodes/miniantags/guardian/types/lightning.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/lightning.dm
@@ -24,8 +24,8 @@
 		if(!(NO_SHOCK in summoner.mutations))
 			summoner.mutations.Add(NO_SHOCK)
 
-/mob/living/simple_animal/hostile/guardian/beam/electrocute_act()
-	return //You are lightning, you should not be hurt by such things.
+/mob/living/simple_animal/hostile/guardian/beam/electrocute_act(shock_damage, obj/source, siemens_coeff = 1, safety = FALSE, override = FALSE, tesla_shock = FALSE, illusion = FALSE, stun = TRUE)
+	return FALSE //You are lightning, you should not be hurt by such things.
 
 /mob/living/simple_animal/hostile/guardian/beam/AttackingTarget()
 	. = ..()
--- a/code/game/gamemodes/miniantags/guardian/types/lightning.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/lightning.dm
@@ -18,11 +18,13 @@
 	var/list/enemychains = list()
 	var/successfulshocks = 0
 
-/mob/living/simple_animal/hostile/guardian/beam/Life(seconds, times_fired) //Dies if the summoner dies
-	..()
-	if(summoner)
-		if(!(NO_SHOCK in summoner.mutations))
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/miniantags/guardian/types/lightning.dm
```diff
@@ -18,6 +18,15 @@
 	var/list/enemychains = list()
 	var/successfulshocks = 0
 
+/mob/living/simple_animal/hostile/guardian/beam/Life(seconds, times_fired) //Dies if the summoner dies
+	..()
+	if(summoner)
+		if(!(NO_SHOCK in summoner.mutations))
+			summoner.mutations.Add(NO_SHOCK)
```
SteelSlayer, line 25: This probably shouldn't be running in Life, and instead be a one time thing. I was thinking you could add a new proc, or even just add code in `New()` for guardians (though that would need modifications to `New()`'s args). The code would be roughly what you have here, except it only needs to run once.

For example with New:

```dm
// Add the user var to the New proc for this example
/mob/living/simple_animal/hostile/guardian/beam/New(loc, mob/living/user)
    if(!user)
        return ..()
    summoner = user
    if(!(NO_SHOCK in summoner.mutations))
		summoner.mutations.Add(NO_SHOCK)
```
Also you may want to remove the shock immunity from the host when the guardian dies, something like:
```dm
/mob/living/simple_animal/hostile/guardian/beam/death(gibbed)
    if(summoner && (NO_SHOCK in summoner.mutations))
		summoner.mutations.Remove(NO_SHOCK)
    return ..()
```

Qwertytoforty: Sure, one moment

Qwertytoforty: don't think I need user coded like that though, since I can just check for if it has a summoner?

SteelSlayer: You would with the New route, because when you get a guardian, the code does this:
```
	var/mob/living/simple_animal/hostile/guardian/G = new pickedtype(user)
	G.summoner = user
```
Alternative would be to create another proc, and then run it after this line above

Qwertytoforty: Getting indentation errors with that code, along with bad argument definition

SteelSlayer: Oh yeah I indented with spaces because you can't use tab in freaking github comments for some reason. Now, the bad argument definition... don't know why you're getting that.

Qwertytoforty: /mob/living/simple_animal/hostile/guardian/beam/New() //Dies if the summoner dies
	..()
	if(summoner)
		if(!(NO_SHOCK in summoner.mutations))
			summoner.mutations.Add(NO_SHOCK)`

code/game/gamemodes/miniantags/guardian/types/lightning.dm
```diff
@@ -106,3 +117,8 @@
 					)
 				L.adjustFireLoss(1.2) //adds up very rapidly
 				. = 1
+
+/mob/living/simple_animal/hostile/guardian/beam/death(gibbed)
+    if(summoner && (NO_SHOCK in summoner.mutations))
+        summoner.mutations.Remove(NO_SHOCK)
+    return ..()
```
SteelSlayer, line 124: Need a trailing new line here.

code/game/gamemodes/miniantags/guardian/types/lightning.dm
```diff
@@ -18,6 +18,17 @@
 	var/list/enemychains = list()
 	var/successfulshocks = 0
 
+/mob/living/simple_animal/hostile/guardian/beam/New(loc, mob/living/user, mob/living/user)
```
SteelSlayer, line 21: :eyes: What I meant was that you needed to add a second `user` argument at this line further down in the file:
`var/mob/living/simple_animal/hostile/guardian/G = new pickedtype(user)` -> `new pickedtype(user, user)`

Dumbdumn5 (approved):
considering its current state as a deathly underused guardian, this seems like a buff that isn't too nuts and could actually help them get some use, even if it is for a niche purpose.

variableundefined (approved):
Echoing Dumb's reasons, more diversity in holopara usage is nice.

PR #14207, opened 2020-09-01
Fixes unusable "sign up" chat button, some missing icons when polling for candidates and gimmick team

dearmochi:
## What Does This PR Do

- Fixes the `[Sign up]` button not doing anything when clicked when polling for candidates for a role
- Fixes some roles showing up as "Do you want to play as a traitor?" when the role is more specific than that
- Fixes some more missing poll icons (SIT, Commando, Wizard...)
- Fixes being unable to send Gimmick Teams due to a runtime

## Why It's Good For The Game

The message is good but the button does not work, forcing the observer to use the in-HUD button. This fixes that. Also Gimmick teams were broken

## Changelog
:cl:
fix: Fix [Sign up] chat button not working when polling for candidates
fix: Fix SIT poll question asking "Do you want to play as a traitor?" instead of SIT
fix: Fix more missing icons when polling (SIT, ...)
fix: Fix being unable to send Gimmick teams
/:cl:


code/modules/admin/verbs/gimmick_team.dm
```diff
@@ -53,7 +53,9 @@
 			players_to_spawn += candidate
 	else
 		to_chat(src, "Polling candidates...")
-		players_to_spawn = SSghost_spawns.poll_candidates("Do you want to play as \a [O.name]?")
+		var/datum/outfit/temp = new O
+		players_to_spawn = SSghost_spawns.poll_candidates("Do you want to play as \a [temp.name]?")
+		qdel(temp)
```
SteelSlayer, line 58: Pretty sure you can get away with this:
```suggestion
		players_to_spawn = SSghost_spawns.poll_candidates("Do you want to play as \a [initial(O.name])?")
```
I tested this and it seemed to give me the proper outfit name. You may want to double check though.

Kyep (approved):

variableundefined:
@dearmochi Will merge when fixed.

AffectedArc07 (approved):

PR #14219, opened 2020-09-02
Detective Revolver Nerf

Cocacolagua:
## What Does This PR Do
A long time coming.
Nerfs the Detective Revolver to no longer weaken on hit, and lowers it's stamina damage from 60 to 35, a tad bit stronger than a disabler, able to down people in three hits instead the disabler’s four.

## Why It's Good For The Game
The detective's job is NOT to chase antagonists, criminals, or whatever.
The revolver, however, encourages them to do otherwise, as it's one of the best ranged weapons in the game, a ballistic taser that you can recharge on-the-go, which can be even be turned into a lethal weapon with a screwdriver.
They can still keep their revolver, but now it will be the same as a disabler, which is still a pretty good weapon.

## Changelog
:cl:
tweak: Detective's Revolver no longer weakens on hit, and it now deals 35 stamina damage instead of 60
/:cl:


TDSSS:
While I'm not opposed to a nerf in principle, making it do exactly as much as a disabler means you're better off just grabbing (or handing the detective a) disabler.

Cocacolagua:
After this change, the disabler will be the counter in some scenarios, and the revolver in other ones, the main reason for this being these key differences:
The revolver uses ballistic ammo, which can for example counter d-swords, energy shields, and the like.
The revolver can be recharged in the middle of a fight, as long as you have spare ammo.
The revolver can be turned into a lethal weapon if it's needed.
The situations in which the disabler are better than the revolver are:
It can go through windows.
It can be recharged with a recharger, instead of an autolathe.

And to the argument of "it's better to just hand a detective a disabler", that's kind of the point of this PR, his revolver shouldn't be better than whatever is in the armory, or whatever security officers start with.

Qwertytoforty:
Seems this will modify the mimes traitor item as well, as their bullets are a subtype of the detectives revolver, which will cause this PR to nerf them.

Cocacolagua:
whoops, wrong value

Mitchs98:
I don't really see a reason for this. The revolver is meant to be more powerful, as it's meant to be a line of defense for the detective(that most die trying to use anyway). If Detectives are running around acting as security constantly and/or lethalling people without situation warranting it, that's a user problem not an item problem imo.

Carthusia:
I feel its damage is fine, though I feel it would be more acceptable for it to have less and less effect on hardsuits the more armored the suit is, so say on civ hardsuits it still works, but on say a syndie hardsuit it does nothing, or very little.

No more nuke ops being 1 shot by a detective.

TheMaveran:
The gun is still better than a disabler, detectives reaalllly don’t need the strongest gun on station round start.

Mitchs98:
> I feel its damage is fine, though I feel it would be more acceptable for it to have less and less effect on hardsuits the more armored the suit is, so say on civ hardsuits it still works, but on say a syndie hardsuit it does nothing, or very little.
> 
> No more nuke ops being 1 shot by a detective.

This I could certainly agree with.

procdrone:
I've recently been chased by a detective who was "playing" an officer and chased me into the maints shooting his revolver not sparingly. his SINGLE shot stunned me instantly. If I didn't ebow him before being hit, that would be over for me.

Instant stun one shot revolver is a bad thing. you say taser, sure. there are ways of blocking a taser. much less options for ballistic bullets.

I agree with the nerf

Mitchs98:
Then remove the weaken, that much I could agree with. I see no reason to lower the stamina damage.

Cocacolagua:
Problem with only removing the weaken is that the gun would deal enough stamina damage to down people in two hits, and the first one would make you so slow you don’t be able to dodge the second one.

I could however maybe raise it to 35 stamina damage, so it three-shots people instead

TheMaveran:
60 stamina damage is insane, it’ll still two shot you. It needs to be nerfed as well, especially considering there are more counters to disablers than ballistics. And again, why does the detective need such a good weapon in the first place? I would genuinely rate it better than AEGs.

Kyep (approved):

Mitchs98:
35 to keep it more in line with disablers would be acceptable at the very least imo.

Spacemanspark:
As amusing as it is to wander up to a changeling arm blading someone and just blast them from across the hallway before bashing them to pieces (yes, I actually did this a few days ago), I have to agree that the detective's revolver (which was among the first of many changes made to stuns way back when, mind you) is actually ridiculous and should have probably been nerfed ages ago, perhaps not even changed in the first place.

Also I'm the guy with the 'I hate instant stuns' title so I have to up vote this obviously

SteelSlayer (approved):
Code's fine.

Tokorizo:
I always wondered why Detectives have better versions of Tazers in almost every way, this is a good PR. The detective should not be armed with a better Tazer, there not supposed to be a combative role their more of sec support, this change leans them more towards that. Also instant stuns are trash remove em.

variableundefined:
I approve this, but I prefer 35 instead of 30 so it is a 3-shoter instead of a 4-shoter, giving it a bit more of a leg up against a normal disabler.

PR #14220, opened 2020-09-02
Makes protector guardians only do 6 object damage in shield mode

Qwertytoforty:
## What Does This PR Do

This PR makes protector guardians only do 6 object damage in shield mode, and returns them back to 40 combat damage in combat mode. 
## Why It's Good For The Game

Right now, protector guardian in shield mode take 5% of incoming damage. They also only deal 2 damage in melee to balance it out. However, they still have full 40 object damage. Normally, this is fine... except when blobs are involved. When blobs are involved, the guardian does 40 object damage to blob tiles while taking only 5% of damage applied to it, allowing to shred through blob, slowly  push to the core, and kill it with its host barely taking damage. Combine this with legion cores, healing meds, and armor to prevent the host from taking full damage from being attacked, and the guardian pushed to the blob core, unstoppable effectively. By nerfing object damage along with melee damage in defense mode, it can no longer do 40 object damage to blob, and instead only 6. It can keep blob from spreading on a tile, but it can not shred through the blob. If it wants to hurt the blob, it must be in combat mode, and take more damage.

## Changelog
:cl:
tweak: Protector guardians now only do 6 object damage in shield mode, preventing cheesing of blobs.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/guardian/types/protector.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/protector.dm
@@ -1,7 +1,6 @@
 /mob/living/simple_animal/hostile/guardian/protector
 	melee_damage_lower = 15
 	melee_damage_upper = 15
-	obj_damage = 40
 	range = 15 //worse for it due to how it leashes
 	damage_transfer = 0.4
 	playstyle_string = "As a <b>Protector</b> type you cause your summoner to leash to you instead of you leashing to them and have two modes; Combat Mode, where you do and take medium damage, and Protection Mode, where you do and take almost no damage, but move slightly slower."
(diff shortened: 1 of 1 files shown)


Rurik123:
RIP my chameleon projector strat of having a guardian rip apart every console in bridge while taking little to no damage in return.

code/game/gamemodes/miniantags/guardian/types/protector.dm
```diff
@@ -1,6 +1,7 @@
 /mob/living/simple_animal/hostile/guardian/protector
 	melee_damage_lower = 15
 	melee_damage_upper = 15
+	obj_damage = 40
```
SteelSlayer, line 4: This is already the default for all guardians. You don't need to write it.

hal9000PR:
I agree with this change. A mob with an effective health of 2000 should not deal this much damage to a blob

variableundefined (approved):
Looks Fair and Balanced™.

PR #14225, opened 2020-09-03
[TGUI] Autolathe

Kyep:
## What Does This PR Do
Converts autolathe to TGUI.
Autolathe now shows pictures of the things you can build in it.

## Images of changes
uNVQSacsaQ

## In this new interface, how do I:
- List items: click the drop down menu and select category. There is no main menu anymore.
- Print/Queue item: click the name of the item in the build list (hammer icon)
- Remove item from queue: click the name of the item in the 'queue' section ('x' icon)
- Clear the whole queue: click the clear queue button, top right, only available when there is a queue
- Search all items: type the name of the item in the 'search for' box (this searches ALL categories, not just current one, so no need to change category to do this)
- Clear search: clear the text in the search box

## Special Thanks
https://github.com/VOREStation/VOREStation/pull/8825

## Changelog
:cl: Kyep
tweak: Convered autolathe to TGUI.
/:cl:


TDSSS:
Maybe how full the storage is could be a bar? Also, those long numbers can be a bit hard to read and parse easily, maybe they could be formatted?

Kyep:
@AffectedArc07 
- The hammer icon actually spins on the thing you're currently constructing. I've uploaded an animated GIF to the PR description so you can see it in action. I think it looks pretty neat. Though I suppose I could change it to simply highlight the currently building item in green if its visible in the queue, if people think that's better.
- Hacked items used to show in red. This has been changed in the most recent build.
- I'm hesitant to put all of those status things on new lines, because if I did, it would really push down the actual list of buildable items, and force the user to scroll a lot more.
- The biggest issues are the one mentioned in the PR TO-DO, I'd like your advice on those.

@TDSSS 
- I agree they aren't easy to read. I've been debating formatting them with ,s: "84,394" for example. But that would make it harder to combine them with other things in a comma separated list.

Really I just want to have an entire second window / column on the right side of the UI, which is used to show the build queue, and the amount of mats in the machine, so those things don't take up vertical space in the main view. I just haven't figured out how to do this yet.

TDSSS:
Gotta say, not a fan of the spinning hammer at all, it looks pretty silly imo. But that might just be me.

AffectedArc07:
Personally I think making the button the standard green "selected" colour is better and easier to see at a glance than a spinning hammer

Kyep:
@AffectedArc07 
- Hammers no longer spin, instead, button is green if its what's being built
- GIF/video/screenshot updated

Kyep:
All remaining icon issues (including both "animated icons display incorrectly" and "some ammo boxes have broken icons") have been fixed.

This PR is now ready for review.

code/game/machinery/autolathe.dm
```diff
@@ -78,63 +78,118 @@
...
+			maxmult = D.maxstack
+		recipes.Add(list(list(
+			"name" = D.name,
+			"category" = D.category,
+			"uid" = D.UID(),
+			"requirements" =  matreq,
+			"hacked" = ("hacked" in D.category) ? TRUE : FALSE,
+			"max_multiplier" = maxmult,
+			"image" = "[icon2base64(icon(initial(I.icon), initial(I.icon_state), SOUTH, 1))]"
+		)))
```
AffectedArc07, line 117: I would like to see profiler results for this, as I feel like `icon2base64` may be expensive.

code/game/machinery/autolathe.dm
```diff
@@ -78,63 +78,118 @@
...
+				return
+			//multiplier checks : only stacks can have one and its value is 1, 10 ,25 or max_multiplier
+			var/multiplier = text2num(params["multiplier"])
+			var/max_multiplier = min(design_last_ordered.maxstack, design_last_ordered.materials[MAT_METAL] ?round(materials.amount(MAT_METAL)/design_last_ordered.materials[MAT_METAL]):INFINITY,design_last_ordered.materials[MAT_GLASS]?round(materials.amount(MAT_GLASS)/design_last_ordered.materials[MAT_GLASS]):INFINITY)
+			var/is_stack = ispath(design_last_ordered.build_path, /obj/item/stack)
+
+			if(!is_stack && (multiplier > 1))
+				return
+			if(!(multiplier in list(1, 10, 25, max_multiplier))) //"enough materials ?" is checked in the build proc
+				return
```
AffectedArc07, line 183: I feel like a `message_admins` here would be good to track potential attempts at href exploits

code/game/machinery/autolathe.dm
```diff
@@ -184,14 +239,20 @@
...
 					"You hear the chatter of a floppy drive.")
 				playsound(get_turf(src), 'sound/goonstation/machines/printer_dotmatrix.ogg', 50, 1)
-				busy = 1
+				busy = TRUE
 				if(do_after(user, 14.4, target = src))
+					if(!("Imported" in D.blueprint.category)) // R&D should always ensure this is set on design disks, but it doesn't.
+						D.blueprint.category += "Imported" // now it will actually show up in the list.
 					files.AddDesign2Known(D.blueprint)
-				busy = 0
+					SStgui.close_uis(src) // forces all connected users to re-open the TGUI. Imported entries won't show otherwise due to static_data
```
AffectedArc07, line 254: Try using `SStgui.update_uis(src)` here as that *should* update static data.

code/game/machinery/autolathe.dm
```diff
@@ -467,3 +454,4 @@
 		for(var/datum/design/D in files.known_designs)
 			if("hacked" in D.category)
 				files.known_designs -= D.id
+	SStgui.close_uis(src) // forces all connected users to re-open the TGUI, thus adding/removing hacked entries from lists
```
AffectedArc07, line 457: Same here

AffectedArc07 (changes requested):

code/game/machinery/autolathe.dm
Kyep: Added

code/game/machinery/autolathe.dm
Kyep: update_uis(src) does not reload static data. Only closing the entire window and re-opening it does.

code/game/machinery/autolathe.dm
Kyep: update_uis(src) does not reload static data. Only closing the entire window and re-opening it does.

SteelSlayer:
Think you could add some space in between these sections?

image

code/game/machinery/autolathe.dm
Kyep: Marking resolved because you stated in design chat that having TM'ed this and profiled it, the CPU is fine.

AffectedArc07:
```
                                                         Profile results (total time)
Proc Name                                       Self CPU     Total CPU    Real Time    Overtime     Calls        Average (Real / Calls)
--------------------------------------------    ---------    ---------    ---------    ---------    ---------    -----------------------
/obj/machinery/autolathe/ui_data                0.063        0.169        0.166        0.019        1494         0.0001
/obj/machinery/autolathe/tgui_static_data       0.215        3.566        3.567        0.117        136          0.0262
```

Profile results with a calculated average. Its **much** higher due to the `icon2base64` calls but has a much smaller amount of calls.

I would still rather have some sort of cache in place on there.

Kyep:
@SteelSlayer Gap added. GIF in description updated.
@AffectedArc07 Caching added.

AffectedArc07 (approved):

PR #14233, opened 2020-09-04
chameleon flag fix

### LINKED ISSUE #14138: Chameleon Flags rigged with Minibombs cannot be lit on fire normally
**Issue Description**:
If you rig a Chameleon flag with a syndie minibomb you can't set it on fire unless you have a fire in the same space, say by burning it with a flamethrower for example

I expected to be able to light the flag on fire with a lighter or welding tool.

**What happened instead**:
The flag became unburnable once given a bomb, confounding myself and abydos

It basically makes the chameleon flag useless outside of being something syndies can buy for giggles.

Get a chameleon flag, set it to any flag type, add a minibomb, cry because you just wasted 7 TC on a flag and bomb.

No clue when it could've started, but issue was noticed and tested by myself and abydos on Aug 23rd 2020

**Extra information**:
N/A

RedWolfie:
## What Does This PR Do
Fixes #14138. Chameleon flags with boobytraps in them can now be set on fire which will trigger the boobytrap
Fixes a runtime error in smoke grenades

## Why It's Good For The Game
Bugs are bad

## Changelog
:cl:
fix: Chameleon flags with booby traps can now be set on fire which will trigger the boobytrap
fix: Runtime error in smoke grenades
/:cl:

### DIFF
--- a/code/game/objects/items/flag.dm
+++ b/code/game/objects/items/flag.dm
@@ -269,12 +269,15 @@
 /obj/item/flag/chameleon/burn()
 	if(boobytrap)
 		fire_act()
-		spawn(boobytrap.det_time)
-			boobytrap.loc = get_turf(loc)
-			boobytrap.prime()
-			..()
+		addtimer(CALLBACK(src, .proc/prime_boobytrap), boobytrap.det_time)
 	else
 		..()
+
+/obj/item/flag/chameleon/proc/prime_boobytrap()
+	boobytrap.forceMove(get_turf(loc))
+	boobytrap.prime()
+	boobytrap = null
+	burn()
 
 /obj/item/flag/chameleon/updateFlagIcon()
 	icon_state = updated_icon_state
--- a/code/game/objects/items/flag.dm
+++ b/code/game/objects/items/flag.dm
@@ -251,6 +251,7 @@
 		message_admins("[key_name_admin(user)] has lit the [src] trapped with [boobytrap] by [key_name_admin(trapper)] at <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[bombturf.x];Y=[bombturf.y];Z=[bombturf.z]'>[A.name] (JMP)</a>.")
 		log_game("[key_name_admin(user)] has lit the [src] trapped with [boobytrap] by [key_name_admin(trapper)] at [A.name] ([bombturf.x],[bombturf.y],[bombturf.z]).")
 		investigate_log("[key_name_admin(user)] has lit the [src] trapped with [boobytrap] by [key_name_admin(trapper)] at [A.name] ([bombturf.x],[bombturf.y],[bombturf.z]).", INVESTIGATE_BOMB)
+		burn()
 	else
 		return ..()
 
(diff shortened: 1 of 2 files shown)


code/game/objects/items/flag.dm
```diff
@@ -267,8 +268,13 @@
...
 	if(boobytrap)
-		boobytrap.prime()
-	..()
+		fire_act()
+		spawn(boobytrap.det_time)
+			boobytrap.loc = get_turf(loc)
+			boobytrap.prime()
+			..()
+	else
+		..()
```
SteelSlayer, line 277: To avoid using spawns, I suggest something like this instead, using callbacks:
```dm
/obj/item/flag/chameleon/burn()
	if(boobytrap)
		fire_act()
		addtimer(CALLBACK(src, .proc/prime_boobytrap), boobytrap.det_time)
	return ..()
```
With the accompanying proc (name it whatever you want):
```dm
/obj/item/flag/chameleon/proc/prime_boobytrap()
	boobytrap.forceMove(get_turf(loc))
	boobytrap.prime()
```
Also `forceMove()` is preferred over setting `loc` directly.

farie82: Echoing Steel here.
Some extra info
Spawns are less performent and also and are also not managed by a subsystem.
`forceMove` will also trigger quite some logic. Which ensures other objects and such react appropriately to the object in question.

farie82 (changes requested):

code/game/objects/items/weapons/grenades/smokebomb.dm
```diff
@@ -19,7 +19,7 @@
 
 /obj/item/grenade/smokebomb/prime()
 	playsound(src.loc, 'sound/effects/smoke.ogg', 50, 1, -3)
-	src.smoke.set_up(10, 0, usr.loc)
+	smoke.set_up(10, 0)
```
farie82, line 22: Be sure to also include this in the CL and description

code/game/objects/items/flag.dm
```diff
@@ -267,8 +268,16 @@
...
+		fire_act()
+		addtimer(CALLBACK(src, .proc/prime_boobytrap), boobytrap.det_time)
+	else
+		..()
+
+/obj/item/flag/chameleon/proc/prime_boobytrap()
+	boobytrap.forceMove(get_turf(loc))
+	boobytrap.prime()
+	boobytrap = null
+	burn()
```
farie82, line 280: Won't this now cause a delayed burn? You use the lighter and then after `det_time` the flag actually starts burning visually

RedWolfie: The lighter calls burn(), which calls fire_act() before the timer. Fire_act causes the flag to burn visually. The timer then calls burn() again after the explosive does it's thing to prevent the flag from burning forever.

farie82: You're right. All good here

farie82 (approved):

AffectedArc07 (approved):

PR #14234, opened 2020-09-04
Makes CNS rebooter stun on EMP

Qwertytoforty:
## What Does This PR Do

This PR makes CNS rebooter stun the user when EMP'D, instead of just being temporarily disabled. 

## Why It's Good For The Game

For the base code of head implants, they stun the user when emp'd. This makes sense, emps and wires in the brain wont end well. Grey translator implants are not included, which makes sense, as making greys get stunned by every emp due to a translator implant is a bit silly. However, the other implant that does not stun the user is... cns. Currently, cns is temporarily disabled when emp'd, but otherwise the host is not affected. This seems a bit silly, as most other implants have a downside to the host when emp'd, not just disabled. More so, a combat implant in your brain when emp'd should be much more dangerous then the clown implant, which does currently stun you when emp'd. Thus, this PR changes the code to make cns stun you AND be disabled temporarily. Now there is a risk to having cns, rather then, its emp'd and disabled temporarily, at no risk to the user.

## Changelog
:cl:
tweak: CNS rebooter now stuns the user on EMP, instead of just being temporarily disabled. 
/:cl:

### DIFF
--- a/code/datums/outfits/outfit_admin.dm
+++ b/code/datums/outfits/outfit_admin.dm
@@ -242,7 +242,7 @@
 		/obj/item/organ/internal/cyberimp/eyes/shield,
 		/obj/item/organ/internal/cyberimp/eyes/hud/security,
 		/obj/item/organ/internal/cyberimp/eyes/xray,
-		/obj/item/organ/internal/cyberimp/brain/anti_stun,
+		/obj/item/organ/internal/cyberimp/brain/anti_stun/hardened,
 		/obj/item/organ/internal/cyberimp/chest/nutriment/plus,
 		/obj/item/organ/internal/cyberimp/arm/combat/centcom
 	)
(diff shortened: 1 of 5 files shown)


procdrone:
Sounds like a reasonable counter to mass produced CNS implanted sec officers.

SteelSlayer (approved):

farie82 (approved):

Tokorizo:
me likey nerf to power-gaming seccies

Kyep (approved):

Kyep (changes requested):
Objection.
If you're going to do this, then you need to distinguish between science-printable implants, and CC/ERT/etc versions.

Kyep (approved):

variableundefined (approved):

PR #14239, opened 2020-09-05
Improvements to singulo logging

AffectedArc07:
## What Does This PR Do
Preamble: Singulo applies to tesla and singularity in here

This PR changes the log messages for singularities being loose from field generator failure. Only singulos on the same z-level as the failing generator will give alerts. This stops us getting alerts from singulos on the wizard shuttle ruin. It also provides details on the singulo location and field generator location.

Before: 
image

After:
image

## Why It's Good For The Game
I am fed up of spurious log messages from singulos on Z6 and other ruins

## Changelog
:cl: AffectedArc07
tweak: Singulos will only trigger field generator failure alerts for field generators on the same z-level
/:cl:

### DIFF
--- a/code/modules/power/singularity/field_generator.dm
+++ b/code/modules/power/singularity/field_generator.dm
@@ -310,16 +310,25 @@ field_generator power level display
 	//This is here to help fight the "hurr durr, release singulo cos nobody will notice before the
 	//singulo eats the evidence". It's not fool-proof but better than nothing.
 	//I want to avoid using global variables.
-	spawn(1)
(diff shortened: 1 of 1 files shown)


Kyep (approved):
Thank goodness.

SteelSlayer (approved):
Yus, more spawns getting killed off.

farie82 (approved):
Currently it is possible to have to different singulo fields on the same level fails at once and only get one message. But this happening is such a low chance.
Looks good to me.

PR #14244, opened 2020-09-06
Adds a notification for Vampires if their victims blood level is getting low

SabreML:
## What Does This PR Do

This adds two notifications in chat for Vampires if their victims blood gets too low, 336 and 224 units to be precise.

Also makes the notification for if you're de-vampired mid succ 'userdanger' rather than 'warning', since only admins can actually cause that afaik.

## Why It's Good For The Game

I saw a youtube video of Fulpstation with some blood level warnings and I thought "why don't we have that"

Really it's mostly a QOL tweak since most vampires know when to stop if they've done it enough, but I still think it's a pretty good tweak to have.

## Images of changes

Desktop Screenshot 2020 09 06 - 06 02 33 01
*(no idea what's up with the gender change thing)*

## Changelog
:cl:
add: As a vampire, you now get warned if your victims blood level is getting too low
/:cl:

### DIFF
--- a/code/game/gamemodes/vampire/vampire.dm
+++ b/code/game/gamemodes/vampire/vampire.dm
@@ -292,7 +292,7 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 		H.LAssailant = owner
 	while(do_mob(owner, H, 50))
 		if(!(owner.mind in SSticker.mode.vampires))
-			to_chat(owner, "<span class='warning'>Your fangs have disappeared!</span>")
+			to_chat(owner, "<span class='userdanger'>Your fangs have disappeared!</span>")
 			return
 		old_bloodtotal = bloodtotal
 		old_bloodusable = bloodusable
--- a/code/game/gamemodes/vampire/vampire.dm
+++ b/code/game/gamemodes/vampire/vampire.dm
@@ -279,6 +279,7 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 	var/blood = 0
 	var/old_bloodtotal = 0 //used to see if we increased our blood total
 	var/old_bloodusable = 0 //used to see if we increased our blood usable
+	var/blood_volume_warning = 9999 //Blood volume threshold for warnings
 	if(owner.is_muzzled())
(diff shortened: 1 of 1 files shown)


Asmerath:
I like this! I would love to see it soon.

hal9000PR:
I like this because vampires aren't necessarily ment to kill people and when I play vampire I don't want to kill people for blood

SteelSlayer (approved):

code/game/gamemodes/vampire/vampire.dm
```diff
@@ -312,6 +310,17 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 				to_chat(owner, "<span class='notice'><b>You have accumulated [bloodtotal] [bloodtotal > 1 ? "units" : "unit"] of blood[bloodusable != old_bloodusable ? ", and have [bloodusable] left to use" : ""].</b></span>")
 		check_vampire_upgrade()
 		H.blood_volume = max(H.blood_volume - 25, 0)
+		//Blood level warnings (Code 'borrowed' from Fulp)
+		if(H.blood_volume)
+			if(H.blood_volume <= BLOOD_VOLUME_BAD && blood_volume_warning > BLOOD_VOLUME_BAD)
+				to_chat(owner, "<span class='danger'>Your victim's blood volume is dangerously low.</span>")
```
Noobeld, line 316: Plural of victim is victims, otherwise it reads "Your victim is blood volume is dangerously low"

SteelSlayer: This is the correct grammer. "Victim's" shows ownership of the blood. It is the victim's blood.

SabreML: It might be Victims' actually, but that doesn't look right to me.

AffectedArc07: `Victims'` would imply multiple victims.

Dumbdumn5 (approved):
Yeah this is pretty cool! Glad to keep the juice boxes alive sometimes. 👍

PR #14248, opened 2020-09-06
Adds a new subsystem: SScleanup

SteelSlayer:
## What Does This PR Do
Adds a new subsystem: SScleanup. Currently, this subsystem is designed to run every 30 seconds, and clear nulls from any of the given global lists. It's pretty lightweight and I don't think it will use that much performance at all, but I can't test that by myself.

The subsystem checks if any nulls were found in the list on its `fire()`, and if there was.. it writes it to the runtime log.

image

## Why It's Good For The Game
 Sometimes, these global lists can gain nulls due to errors. For example, when a dead player trasitions from the `dead_mob_list` to the `alive_mob_list`, a null value may get stuck in the dead mob list.

This can cause issues when other code tries to do things with the values in the list, but are instead met with null values. These problems are incredibly hard to track down and fix (I've tried). I think this subsystem should be a decent solution to that.

## Changelog
Nothing. Backend change.


Kyep:
:/

I wonder if this encourages lazy coding practice of the form: "just allow it to be null, it'll get cleared eventually anyway".

SteelSlayer:
Maaaaybe if they're a lazy nerd. Its more just for stuff slipping through the cracks.

farie82:
Maybe add logging comparing the old values from last tick to the new values of the new tick somehow.
Or log the previous ticks list if the new ticks list has nulls.

This system will help filter out odd cases that cause random code to fail. But it will also filter out odd cases which causes code to fail while testing. I'm not sure if in the current form this SS will help maintain stability or hurt it in the long run

SteelSlayer:
Should it be logged to the runtime log or somewhere else?

farie82:
> Should it be logged to the runtime log or somewhere else?

Would be the easiest log for contributers to access it. It would make more sense in the debug log or such but those are harder to get currently on Paradise itself.

SteelSlayer:
Yeah that's what I was thinking. I can add that to the PR.

farie82:
While this will signal that something went wrong in a certain list. It won't tell much itself.
Like you don't know what the null value was before. You just know something was null.
Due to the 30 seconds in between certain data might get lost as to what the null value might have been. But perhaps keeping a stringified version of the prev list value could be written to the log and a stringified version of the new list.
Example:
```
prev_list = list("client1", "client2", "client3")
new_list = list(client1, null, client3)
```
This is a very optimistic example since the values didn't get shuffled around etc
With this you can maybe find out what happened by finding out what object got nulled. Or you could trace it back in the logs. client2 might have crashed or such which somehow caused logout to not happen and the client object got deleted.
Though again due to the 30 seconds in between I am not sure if the data would help much. Lots can happen in 30 seconds

SteelSlayer:
That's probably possible. I'd make it just print the value(s) at those positions that were nulled out. Using your example, it would print `"client2"` in this case. But you're right about the whole 30 second thing. And if this fires too much or has too much logic, it will start to eat up performance, which I didn't want to have happen.

code/controllers/subsystem/cleanup.dm
```diff
@@ -0,0 +1,44 @@
...
+	flags = SS_POST_FIRE_TIMING
+	priority = FIRE_PRIORITY_CLEANUP
+	init_order = INIT_ORDER_CLEANUP
+	runlevels = RUNLEVEL_LOBBY | RUNLEVELS_DEFAULT
+	offline_implications = "Certain global lists will no longer be cleared of nulls, which may result in runtimes. No immediate action is needed."
+	/// A list of global lists we want the subsystem to clean.
+	var/list/lists_to_clean
+
+/datum/controller/subsystem/cleanup/Initialize(start_timeofday)
+	. = ..()
```
AffectedArc07, line 25: I know its essentially the same, but its standard format here to have `return ..()` at the end of SS init. It also makes more sense to send the initialization message *after* the SS has initialized, not before the proc finishes.

AffectedArc07 (changes requested):

code/controllers/subsystem/cleanup.dm
SteelSlayer: Ah good to know

AffectedArc07 (approved):

PR #14250, opened 2020-09-06
Kills off SSchat

SteelSlayer:
## What Does This PR Do
Essentially a revert of #11921.

As discussed here: #14042, and with Affected on discord, SSchat is no longer wanted, because it has some issues/shortcomings.

## Why It's Good For The Game
Removes unused and unwanted code. Probably a small performance boost on `to_chat` since we won't have to do a bunch of if checks related to SSchat anymore.

## Changelog
Nothing. Backend change.


code/__DEFINES/subsystems.dm
```diff
@@ -80,8 +80,6 @@
 #define INIT_ORDER_NANOMOB -23
 #define INIT_ORDER_SQUEAK -40
 #define INIT_ORDER_PATH -50
-#define INIT_ORDER_PERSISTENCE		-95
```
AffectedArc07, line 83: Put this back in

AffectedArc07 (changes requested):

AffectedArc07 (approved):
This hasnt been enabled apart from when it was bugged, and had issues and shortcomings.

PR #14260, opened 2020-09-07
refactor of anchored, density and opacity

RedWolfie:
## What Does This PR Do
Replaces the 0/1 values of anchored, desnity and opacity with TRUE/FALSE

## Why It's Good For The Game
Improves readability of the code


farie82:
Can you post the search/replace (regex) strings you used to replace the values?

SteelSlayer:
This will probably conflict with a lot of stuff, but.. it needs to be done so :+1:

SteelSlayer (approved):
Gave everything a one over, seem fine to me.

RedWolfie:
> Can you post the search/replace (regex) strings you used to replace the values?

I searched&replaced these strings and checked the results afterwards. Not all of them had results. 
anchored = 0.0
anchored = 1.0
anchored = 0
anchored = 1
anchored=0.0
anchored=1.0
anchored=0
anchored=1

Same goes for density/opacity

AffectedArc07:
Im going to put my concerns down here, in the same way I did on https://github.com/ParadiseSS13/Paradise/pull/14017

This doesnt "fix" anything ingame, and will just create mass conflicts across open PRs. Yes all of these things *should* be TRUE/FALSE, but with the amount of conflicts this would create, I am not fully on board with mass replaces which dont have any ingame effect, especially when this doesnt do a huge amount for maintainability, since everyone pretty much accepts that there will be discrepancies in the codebase.

SabreML:
I approve of this, for what it's worth.
While a bunch of conflicts will pop up if this gets merged, it really wouldn't be hard to fix them. Since it'd just be changing a few `1`s to `TRUE`.

I don't know about anyone else, but I personally think that anything which makes the code cleaner is worth it. Especially with the only cost being a minute of someones time.

variableundefined:
Second objection on the same ground as Affected.

When (if) our open PR number is cut down to a more manageable amount, this should be OK. Until then we should go for an approach of refactoring them when we see them.

PR #14262, opened 2020-09-08
Narrow Conveyors

ExusA:
## What Does This PR Do
Makes conveyor belt sprites EVER so slightly narrower (one pixel off each side).

## Why It's Good For The Game
It's currently quite hard to click on/see/mouseover the turf under a conveyor. Generally, it's good to be able to click on/hover your mouse over/see a turf easily without alt-clicking or right-clicking to see everything on the tile, and this helps with that. This also (at least to my eyes) makes the belts look a little less 2D, which I think is good.

## Images of changes
Before
conveyorwide

After
conveyor

## Changelog
:cl:
imageadd: Narrow-ified the conveyor sprites.
/:cl:


ExusA:
Those before/after pictures aren't great, redoing...

PacifistDalek:
Looks like it actually fits under the upper side of the recycler now too, which helps with the 3d effect

Dumbdumn5 (approved):
Kinda like the 3D effect like this! :+1:

AffectedArc07 (approved):

PR #14263, opened 2020-09-08
Recycler & Conveyor switch examines

ExusA:
## What Does This PR Do
1. Rejigs the status light messages on the recycler.
2. Explains how to use a conveyor belt switch, conveyor belt, and both their assemblies when examining them.

## Why It's Good For The Game
These two objects are a bit hard to figure out to new players. Who's to know the difference between "safety mode" and "safety sensors"? The way the conveyor switch and belts are set up is also a bit unintuitive, and having it explained can be a big help.

## Images of changes
Recycler:
recyclernew

Belt switch:
switchnew

## Changelog
:cl:
tweak: Made the recycler and conveyor switch examine text a bit clearer.
/:cl:

### DIFF
--- a/code/game/machinery/recycler.dm
+++ b/code/game/machinery/recycler.dm
@@ -42,8 +42,8 @@
 
 /obj/machinery/recycler/examine(mob/user)
 	. = ..()
-	. += "<span class='notice'>The power light is <b>[(stat & NOPOWER) ? "<b>off</b>" : "<b>on</b>"]</b>."
-	. += "The operation light is [emergency_mode ? "<b>off</b>. \The [src] has detected a forbidden object with its sensors, and has shut down temporarily." : "<b>on</b>. \The [src] is active."]"
+	. += "<span class='notice'>The power light is [(stat & NOPOWER) ? "<b>off</b>" : "<b>on</b>"]."
+	. += "The operation light is [emergency_mode ? "<b>off</b>. [src] has detected a forbidden object with its sensors, and has shut down temporarily." : "<b>on</b>. [src] is active."]"
 	. += "The safety sensor light is [emagged ? "<b>off</b>!" : "<b>on</b>."]</span>"
 
 /obj/machinery/recycler/power_change()
--- a/code/game/machinery/recycler.dm
+++ b/code/game/machinery/recycler.dm
@@ -9,7 +9,7 @@
 	anchored = 1
 	density = 1
 	damage_deflection = 15
-	var/safety_mode = 0 // Temporarily stops machine if it detects a mob
+	var/emergency_mode = FALSE // Temporarily stops machine if it detects a mob
 	var/icon_name = "grinder-o"
 	var/blood = 0
 	var/eat_dir = WEST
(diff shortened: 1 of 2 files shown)


code/game/machinery/recycler.dm
```diff
@@ -42,9 +42,9 @@
 
 /obj/machinery/recycler/examine(mob/user)
 	. = ..()
-	. += "The power light is [(stat & NOPOWER) ? "off" : "on"]."
-	. += "The safety-mode light is [safety_mode ? "on" : "off"]."
-	. += "The safety-sensors status light is [emagged ? "off" : "on"]."
+	. += "<span class='notice'>The power light is <b>[(stat & NOPOWER) ? "<b>off</b>" : "<b>on</b>"]</b>."
+	. += "The operation light is [emergency_mode ? "<b>off</b>. \The [src] has detected a forbidden object with its sensors, and has shut down temporarily." : "<b>on</b>. \The [src] is active."]"
```
SteelSlayer, line 46: Some superfluous `<b>` tag removal. Also you don't need to use `\The`. simply writting `[src]` will append a "the" to the front.
```suggestion
	. += "<span class='notice'>The power light is [(stat & NOPOWER) ? "<b>off</b>" : "<b>on</b>"]."
	. += "The operation light is [emergency_mode ? "<b>off</b>. [src] has detected a forbidden object with its sensors, and has shut down temporarily." : "<b>on</b>. [src] is active."]"
```

farie82 (approved):
It will work as is but I'm not sure if it would be desirable to be done like this.
Maybe adding a new var on the `obj` class that will be shown when examining up close or such? This info is most likely only needed when examining something up close since you're using it.
This new var can then hold more info regarding the object. On how to use it etc
I'll approve it since the idea above is quite a bigger PR than this PR itself is and I'll leave it up to you if you want to implement that or not.

AffectedArc07:
> It will work as is but I'm not sure if it would be desirable to be done like this.
> 
> Maybe adding a new var on the `obj` class that will be shown when examining up close or such? This info is most likely only needed when examining something up close since you're using it.
> 
> This new var can then hold more info regarding the object. On how to use it etc
> 
> I'll approve it since the idea above is quite a bigger PR than this PR itself is and I'll leave it up to you if you want to implement that or not.

Don't do this. The blue detailed examine text is usually used for object construction and meant to be updated as stages go. An example of this is taking apart a reinforced wall, it changes this text each stage and is set during the examine() call. 

We also already have a detailed description var, but it's seldom used. Examine an airlock or vent and then look in the Examine verb tab. I believe guns also have a detailed description too.

farie82:
> > It will work as is but I'm not sure if it would be desirable to be done like this.
> > Maybe adding a new var on the `obj` class that will be shown when examining up close or such? This info is most likely only needed when examining something up close since you're using it.
> > This new var can then hold more info regarding the object. On how to use it etc
> > I'll approve it since the idea above is quite a bigger PR than this PR itself is and I'll leave it up to you if you want to implement that or not.
> 
> Don't do this. The blue detailed examine text is usually used for object construction and meant to be updated as stages go. An example of this is taking apart a reinforced wall, it changes this text each stage and is set during the examine() call.
> 
> We also already have a detailed description var, but it's seldom used. Examine an airlock or vent and then look in the Examine verb tab. I believe guns also have a detailed description too.

Exactly my point. A lot of things already do this their own way. All doing the same but with their own logic. All the while they want to achieve the same goal. Add extra blue text to give hints of to what do next gameplay wise

Kyep (approved):

PR #14268, opened 2020-09-08
Fixes incorrect TS gibbing

Kyep:
## What Does This PR Do
Attempts to fix an issue where sometimes, terror spiders can gib while being alive, due to Life() being relied upon to always return TRUE if the mob is alive, but apparently it no longer does this, leading to the TS code thinking the spider is dead when it isn't.

## Changelog
:cl: Kyep
fix: fixed terrors sometimes wrongly decomposing(gibbing) for no good reason while alive.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
@@ -297,8 +297,10 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
 	return ..()
 
 /mob/living/simple_animal/hostile/poison/terror_spider/Life(seconds, times_fired)
-	. = ..()
-	if(!.) // if mob is dead
+	if(QDELETED(src))
+		return
+	..()
+	if(stat == DEAD) // Can't use if(.) for this due to the fact it can sometimes return FALSE even when mob is alive.
 		if(prob(2))
 			// 2% chance every cycle to decompose
 			visible_message("<span class='notice'>\The dead body of the [src] decomposes!</span>")
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
@@ -297,9 +297,7 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
 	return ..()
 
 /mob/living/simple_animal/hostile/poison/terror_spider/Life(seconds, times_fired)
-	if(QDELETED(src))
-		return
-	..()
+	. = ..()
 	if(stat == DEAD) // Can't use if(.) for this due to the fact it can sometimes return FALSE even when mob is alive.
 		if(prob(2))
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -297,8 +297,10 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
 	return ..()
 
 /mob/living/simple_animal/hostile/poison/terror_spider/Life(seconds, times_fired)
-	. = ..()
-	if(!.) // if mob is dead
+	if(QDELETED(src))
+		return
+	..()
```
farie82, line 302: Still needs to be `. = ..()`

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -297,8 +297,10 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
 	return ..()
 
 /mob/living/simple_animal/hostile/poison/terror_spider/Life(seconds, times_fired)
-	. = ..()
-	if(!.) // if mob is dead
+	if(QDELETED(src))
```
farie82, line 300: Can't ever happen here.

farie82 (changes requested):
Doing an analysis myself as well:
`/mob/living/simple_animal/hostile/Life`:
Can return `FALSE` if the return of the parent was already `FALSE`

`/mob/living/proc/Life`:
can return `FALSE` if `notransform` is `TRUE`. Which I doubt it is
returns `FALSE` when `loc` is null. Same here should not really happen.
returns `null` if a chem qdels the spider.
returns `null` if a virus or anything else in between qdels the spider
returns `null` if `forced_look` is not null and that value is a UID of an atom and that atom is out of view distance. The logic here is ... flawed at best if you ask me. Probably needs some relooking
Will return `TRUE` if the spider is not dead. Else it'll return `null`

Only causes I can see here are:
- The `notransform` causing the issue here. The timestop effect can in theory currently kill the terrors since it sets notransform to TRUE while it's active.
Something with the `view_portal` object can also cause it but I have no clue what this is.
- The `forced_look` if you force look (ctrl shift middle click) on something and that thing gets out of sight.
- `loc` somehow being null. But that would have caused more issues.

Can use this to debug some more. The current proposed solution with the death check should work but there might be some other things you might want to look into as well

Kyep:
@farie82 Insight: most of the cases of this bug happening seem to happen in space. I've not been able to reproduce it yet, but that's the best lead we have thus far.

farie82:
> @farie82 Insight: most of the cases of this bug happening seem to happen in space. I've not been able to reproduce it yet, but that's the best lead we have thus far.

Do they perhaps die in space and get instantly gibbed upon dying? I can't remember if terrors take pressure/cold damage from space

Kyep:
Updated.
@farie82 that is possible, but its only a 2% chance per life tick, so unlikely.

farie82 (approved):

AffectedArc07 (approved):

PR #14272, opened 2020-09-09
Security Holobarrier Revamp

ExusA:
## What Does This PR Do
Every time a human mob passes a security holobarrier, the holder of the projector gets alerted. The projector tells you:
Who passed the barrier, (only based off sight, someone with a gasmask+no ID will come out as "Unknown")
Where the barrier is (you can have two active, so you need to know which one was triggered.)

Extra changes/info:
- The projector can now only support two barriers in total.
- The projector is removed from the security officers' and the warden's closets, it's now only in the HoS and detective lockers. ERTs and cyborgs still have them.
- You can now pass a security barrier at full running speed if you have an ID with sec access, and an engi barrier if you have an id with engi access. If you have a sec ID, the sec barrier won't report you crossing it, either.
- When destroyed, the barrier will send a dying message to its projector, declaring the visible name of anyone in a 1 tile radius. If it was destroyed in melee, it'd be the attacker, and the holder would get a clue as to who broke it. Otherwise, they'd be kinda useless to monitor traffic, since you could just break them without consequence. You still sorta can by shooting them from a distance, but the projector will still report the barrier being destroyed, just not who destroyed it.
- I thought it might be a secret cham projector nerf, but apparently you can't pass a security barrier with a cham projector at all anyway. Huh.

## Why It's Good For The Game
The security holobarrier is fairly silly, since It really doesn't work as a means to stop someone going somewhere, which is what it's supposed to do. This makes it a valuable tool for increasing the security on certain areas like maints, while limiting it in a similar way to the detective's scanner, in that there's only effectively 4 on a normal round (1 for HoS, 1 for det, 2 for cyborgs). 
I really just made this for practice, though. It seems like a neat use for an underused, unappreciated item, but I'll leave that up to you.

## Changelog
:cl:
experiment: Revamped security holobarriers.
/:cl:


Spacemanspark:
Hrm, I like this, although I'm not sure about taking it away from officers.

ppi13:
This here would make the holo barrier extremely usefull, it would allow you to guard a place, and it being impossible to hack your way in without anyone knowing, you could hide it beneath doors, in secure locations, the ai is omnisciencent enough.  
Due to player mentality, this is a thing that would be abused and just take away the greatest power antags have, stealth.
Only thing good here, really, is engineer ones actually blocking people, however, i would add engineer and atmospheric access as one of the two accesses required.  
Sec barrier barrier should be the same, where brig access lets you pass through while running.  
They are not ment to be a permanent thing, just a temporary one.

ExusA:
Thoughts from discord:
- Sec will preemptively put them on armory/maints and otherwise generally powergame with them
- The detective probably shouldn't have one...
- Lots of antags rely on stealth, this'll nerf like... all of them at once

All that in mind, this clearly needs a rethink, but I'm trying to sleep, so... tomorrow!

PR #14217, opened 2020-09-02
Crabwalking now causes slowdown similar to walk speed

Kyep:
## What Does This PR Do
While crabwalking (ie: using direction lock to keep your sprite looking at a certain object / in a certain direction) your movement speed will be reduced by the same amount its reduced by being on walk intent instead of run intent.

## Why It's Good For The Game
Discourages people from crabwalking for a long period of time, while still allowing direction locking to be used for things like slowly backing away from a threat.

## Design Questions
Does this stack with walk intent? Yes. If you use walk intent with this as well, the penalty stacks.
Why not just lock the mob to walk intent? Because this also applies to simple_animals who don't support move intents or changing intents.

## Coding Questions
Why do you delete the . = ..() in simple_animal/movement_delay()? Because the very next line overwrites . with something else, so functionally it did nothing. I chose to remove it for the sake of clarity and not running /mob/living/movement_delay() every life tick for every simple mob for no purpose. If desired, I can change it to ..(), though it would still do nothing as written.

## Changelog
:cl: Kyep
tweak: using direction lock to make your character always face one direction, now reduces your movement speed.
/:cl:

PR #14273, opened 2020-09-09
Removes Crab walking

ppi13:
## What Does This PR Do

This removes crabwalking from the game

## Why It's Good For The Game

Crab walking is against the rules for being low RP, so instead of giving admins additional work due to crab walking, it can just be removed.

## Changelog
:cl:
del: Removes Crabwalking
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -76,12 +76,6 @@
 	if(IsFrozen(A) && !is_admin(usr))
 		to_chat(usr, "<span class='boldannounce'>Interacting with admin-frozen players is not permitted.</span>")
 		return
-	if(modifiers["middle"] && modifiers["shift"] && modifiers["ctrl"])
-		MiddleShiftControlClickOn(A)
-		return
-	if(modifiers["middle"] && modifiers["shift"])
-		MiddleShiftClickOn(A)
-		return
 	if(modifiers["shift"] && modifiers["ctrl"])
 		CtrlShiftClickOn(A)
 		return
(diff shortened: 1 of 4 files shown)


TheSardele:
https://github.com/ParadiseSS13/Paradise/pull/14217
This PR is supposed to fix this already.

Aurorablade:
No reason to remove it just add a feedback to on the combo to disable it and honestly crabwalking is not a issue that needs admin intervention

AffectedArc07:
As sardele said, we already have https://github.com/ParadiseSS13/Paradise/pull/14217 open. I would also like to quote this section from the advanced rules

image
Direction locking **can** have its uses in RP, such as locking your direction forwards and walking backwards, to slowly walk out of a room you werent supposed to walk in on, etc etc.

Mitchs98:
This is incredibly excessive. And the rules on crabwalking themselves even more-so, really. I see no reason it should be removed, much less be against the rules or how it's 'lowRP' or damaging the RP environment. I think nothing of it when I see people doing it, it's a little quirky and silly too but...so is SS13. 

I don't really agree with giving it any downsides either...but it's better than outright removal.

Aurorablade:
It dosent need downsides or a whole advanced rule either

Pixelzilla:
> 
> 
> It dosent need downsides or a whole advanced rule either

But the fact of the matter is that right now players _are_ being reprimanded for doing it and there isn't a point in having something that was intentionally added being basically not allowed but still a feature because all it is now is a trap for players to get in trouble. Its so easy to accidentally do it and it isn't even clear how to stop doing it, doesn't seem worth keeping if players are just going to be told to not do it.

ppi13:
If I can do something that harms nobody, destroys nothing, and gives you zero advantages, can cause you to be bwoinked by an admin, then it does not matter if it gets nerfed, unless the rule is changed, if something like this is made against the rules, you remove it. 

It's sending mixed signals between what is and is not okay, and if something like this is against the rules, and if something like this can get you in trouble then you remove it, I have no way to change the rules, but I can change this

Dumbdumn5:
Though I personally find the rule against it and the subsequent PRs over something this minor unnecessary, I put up a vote on the admin forums just last week that went over 'how we want to deal with it going forwards', with most (1/2+) of the admin staff voting for Kyet's option of nerfing it, hence the other PR.

Kyep (changes requested):
This isn't going to pass.
It has 2 maints and 1 head (not me) objecting to it, plus no head or maint in favor of it.

PR #14274, opened 2020-09-09
Adds a 'Clear Scanner Records' action for Forensic Scanners

SabreML:
## What Does This PR Do

Adds `clear_scanner()` as an action button for the Forensic Scanner, which clears the scanner logs without having to print a report.

Also cleans up the surrounding code a bit, mostly removing `spawn()`s.

## Why It's Good For The Game

It's pretty inconvenient to have to print a whole scanner report if you just want to remove your accidental misclick from the log.

## Images of changes

NSS Cyberiad 2020-09-15 091400

## Changelog
:cl:
add: There is now a 'Clear Scanner Records' button for Forensic Scanners
imageadd: Added an icon for the 'Print Scanner Report' action
tweak: Refactored the scanner code a bit
/:cl:


code/modules/detective_work/scanner.dm
```diff
@@ -88,6 +88,19 @@
...
+/obj/item/detective_scanner/verb/clear_scanner()
+	set name = "Clear Scanner Records"
+	set category = "Object"
+	set desc = "Clear the records on your Forensic Scanner."
+
+	if(length(log) && !scanning)
+		log = list()
+		playsound(loc, 'sound/machines/ding.ogg', 50)
+		spawn(15) //So that it clears on the 'ding'
+			to_chat(usr, "<span class='notice'>Scanner logs cleared.</span>")
```
SteelSlayer, line 100: Timers are preferred over spawns:
```suggestion
		addtimer(CALLBACK(GLOBAL_PROC, .proc/to_chat, usr, "<span class='notice'>Scanner logs cleared.</span>"), 1.5 SECONDS)  //So that it clears on the 'ding'
```

SteelSlayer: If you want, yeah, go for it.

code/modules/detective_work/scanner.dm
```diff
@@ -17,77 +17,85 @@
...
-			log = list()
-			scanning = 0
+
+		addtimer(CALLBACK(src, .proc/make_paper, log), 10 SECONDS) // Create our paper
+		log = list() // Clear the logs
+		scanning = 0
 	else
 		to_chat(usr, "<span class='notice'>The scanner has no logs or is in use.</span>")
 
+/obj/item/detective_scanner/proc/make_paper(log) // Moved to a proc because 'spawn()' is evil
```
farie82, line 73: Best to give this call an user. Tracing it back. `ui_action_click` gives an user and actiontype as param. The override in this file of that proc does not take those params but it should. That way you can pass the user

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
-			else
-				if(holder)
-					to_chat(holder, "<span class='notice'>You finish scanning \the [target_name].</span>")
-
-			add_log("---------------------------------------------------------", 0)
-			scanning = 0
-			return
+
+		// We gathered everything. Slowly display the results to the holder of the scanner.
+		var/found_something = 0
```
farie82, line 167: `FALSE`

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+		add_log("<B>[station_time_timestamp()][get_timestamp()] - [target_name]</B>", 0)
+
+		// Fingerprints
+		if(length(fingerprints))
+			sleep(30)
+			add_log("<span class='info'><B>Prints:</B></span>")
+			for(var/finger in fingerprints)
+				add_log("[finger]")
+			found_something = 1
```
farie82, line 176: `TRUE`. Same for all the other cases of this var

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+		if(length(reagents))
+			sleep(30)
+			add_log("<span class='info'><B>Reagents:</B></span>")
+			for(var/R in reagents)
+				add_log("Reagent: <font color='red'>[R]</font> Volume: <font color='red'>[reagents[R]]</font>")
+			found_something = 1
+
+		// Get a new user
+		var/mob/holder = null
+		if(ismob(src.loc))
```
farie82, line 204: src implied

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+			holder = src.loc
```
farie82, line 205: Same

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+
+		if(!found_something)
+			add_log("<I># No forensic traces found #</I>", 0) // Don't display this to the holder user
```
farie82, line 208: `FALSE` instead of the 0.

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+			if(holder)
+				to_chat(holder, "<span class='notice'>Unable to locate any fingerprints, materials, fibers, or blood on \the [target_name]!</span>")
```
farie82, line 210: `\the` doesn't do anything here I believe. Need to test this though

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+		else
+			if(holder)
+				to_chat(holder, "<span class='notice'>You finish scanning \the [target_name].</span>")
```
farie82, line 213: Same here

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+
+		add_log("---------------------------------------------------------", 0)
```
farie82, line 215: `FALSE`

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+		scanning = 0
```
farie82, line 216: `FALSE`

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+		return
```
farie82, line 217: Non required return

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+				add_log("[fiber]")
+			found_something = 1
+
+		//Reagents
+		if(length(reagents))
+			sleep(30)
+			add_log("<span class='info'><B>Reagents:</B></span>")
+			for(var/R in reagents)
+				add_log("Reagent: <font color='red'>[R]</font> Volume: <font color='red'>[reagents[R]]</font>")
+			found_something = 1
```
farie82, line 200: `TRUE`

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
+			for(var/B in blood)
+				add_log("Type: <font color='red'>[blood[B]]</font> DNA: <font color='red'>[B]</font>")
+
+		//Fibers
+		if(length(fibers))
+			sleep(30)
+			add_log("<span class='info'><B>Fibers:</B></span>")
+			for(var/fiber in fibers)
+				add_log("[fiber]")
+			found_something = 1
```
farie82, line 192: `TRUE`

code/modules/detective_work/scanner.dm
```diff
@@ -154,61 +162,59 @@
...
-				if(holder)
-					to_chat(holder, "<span class='notice'>You finish scanning \the [target_name].</span>")
-
-			add_log("---------------------------------------------------------", 0)
-			scanning = 0
-			return
+
+		// We gathered everything. Slowly display the results to the holder of the scanner.
+		var/found_something = 0
+		add_log("<B>[station_time_timestamp()][get_timestamp()] - [target_name]</B>", 0)
```
farie82, line 168: `FALSE`

farie82 (changes requested):

code/modules/detective_work/scanner.dm
```diff
@@ -17,77 +17,85 @@
 	actions_types = list(/datum/action/item_action/print_report)
 
 /obj/item/detective_scanner/attack_self(var/mob/user)
-	var/search = input(user, "Enter name, fingerprint or blood DNA.", "Find record", "")
+	var/search = lowertext(input(user, "Enter name, fingerprint or blood DNA.", "Find record", ""))
```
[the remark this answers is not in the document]
SabreML: I didn't realise that would have any performance impact, I was trying to clean up mostly.
Should I just revert that part and move it back to after the checks?

code/modules/detective_work/scanner.dm
```diff
@@ -17,77 +17,85 @@
...
-			if(fingerprint == "FINGERPRINT NOT FOUND") // We have searched by DNA, and do not have the relevant information from the fingerprint records.
+			if(fingerprint == "FINGERPRINT NOT FOUND") // We have searched for DNA, and so do not have the relevant information from the fingerprint records.
 				name = M.fields["name"]
 				for(var/gen_record in GLOB.data_core.general)
 					var/datum/data/record/S = gen_record
 					if(S && (name == S.fields["name"]))
 						fingerprint = S.fields["fingerprint"]
-						continue
-			continue
+						break
```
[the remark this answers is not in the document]
SabreML: I'm really not sure about that whole loop at line 45, but I couldn't figure out how to do it better.
If you search for DNA, then it runs a loop over `data_core.general` twice, which just seems really inefficient.

Of course, every time I've worked on this it's been around midnight, so I might just be missing something obvious.

code/modules/detective_work/scanner.dm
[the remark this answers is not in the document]
farie82: I'd move it back to where it was yeah

code/modules/detective_work/scanner.dm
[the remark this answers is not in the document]
farie82: I'd say think about what this two loops try to achieve. Then think okay when is this achieved and is it always required?
From that point you can most likely reconstruct it so that it is more efficient and logical

code/modules/detective_work/scanner.dm
SabreML: It seems like it was needed there, since `target_name` was the `name` of the atom, rather than the atom itself.
I've changed it to `[A]`, so it's not needed anymore.

code/modules/detective_work/scanner.dm
SabreML: Why would `user` be needed here? To replace the `if(ismob(loc))`?

If so, wouldn't that mean that only the user who printed the report gets in put into their hands?
(i.e. Detective prints report, drops the scanner, someone else picks it up.)

farie82: Fair point there. Yeah in this case a loc check would make more sense due to the long delay

Kyep (changes requested):
Any chance you could make this verb into a HUD button?
I ask because its bad practice to be forced to right-click on an object to use a verb.

SabreML:
> Any chance you could make this verb into a HUD button?
> I ask because its bad practice to be forced to right-click on an object to use a verb.

I wanted to avoid that because I was worried someone could accidentaly wipe the records when they're trying to print them, but I suppose if I make a modified icon for the action it should be fine.

code/modules/detective_work/scanner.dm
```diff
@@ -154,63 +164,60 @@
...
-			return
 
-/obj/item/detective_scanner/proc/add_log(var/msg, var/broadcast = 1)
+		// We gathered everything. Slowly display the results to the holder of the scanner.
+		var/found_something = FALSE
+		add_log("<B>[station_time_timestamp()][get_timestamp()] - [target_name]</B>", FALSE)
+
+		// Fingerprints
+		if(length(fingerprints))
+			sleep(30)
```
Kyep, line 174: Sleeps?

code/datums/action.dm
```diff
@@ -162,6 +162,14 @@
...
 	name = "Print Report"
 
+/datum/action/item_action/print_forensic_report
+	name = "Print Report"
+	button_icon_state = "scanner_print"
+	use_itemicon = FALSE
+
+/datum/action/item_action/clear_records
+	name = "Clear Scanner Records"
+
```
Kyep, line 172: Doesn't this need its own icon? Like a picture of the scanner with an 'X' or something?

code/modules/detective_work/scanner.dm
SabreML: Those are so that it pauses between each section, presumably to make it feel more like it's scanning in real time.
I tested it by setting it to `sleep(100)`, and it only the scanner gets paused while it's sleeping.

code/datums/action.dm
SabreML: I tried that, but I couldn't work out a way to make it look good.
I figured it'd be fine with just the default icon, since the 'Print Scanner Report' icon is different now.

code/modules/detective_work/scanner.dm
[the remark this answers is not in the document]
SabreML: It could be stockholm syndrome talking, but I'm starting to think this works fine as it is.

- If you search for a name or fingerprint, it'll look through the general datacore for them, and then get the blood DNA from the medical datacore, based on the name.

- If you searched for DNA, that's not in the general datacore, so it won't find anything in the first`for()` loop.
Then it gets the DNA and the name, and searches the general datacore again for the fingerprint.

The best way to fix it would be to skip the first loop if you're searching for a fingerprint, but there's no way to tell that automatically. And I don't think it's worth adding a 'Name, Fingerprint or DNA?' alert.

Kyep (approved):

PR #14283, opened 2020-09-12
Makes cigarettes catch fire if their holder is also on fire

SabreML:
## What Does This PR Do

This makes it so that when you're on fire, any cigarettes or cigarette subtypes in your inventory will also become lit.
*(Top level only, so it doesn't look inside bags and so on.)*

Doing this also adds a framework for any other fire-clothing interactions, throught the `catch_fire()` proc.

## Why It's Good For The Game

You can already light cigarettes on someone elses burning body, so why not on your own!

## Images of changes

ezgif com-optimize
Capture

## Changelog
:cl:
add: Cigarettes and their subtypes will now light when their owner catches fire
add: catch_fire() proc for all clothing
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/cigs.dm
+++ b/code/game/objects/items/weapons/cigs.dm
@@ -66,6 +66,10 @@ LIGHTERS ARE IN LIGHTERS.DM
 	..()
 	light()
 
+/obj/item/clothing/mask/cigarette/catch_fire()
+	if(!lit)
+		light("<span class='warning'>Your [name] is lit by the flames!</span>")
+
 /obj/item/clothing/mask/cigarette/welder_act(mob/user, obj/item/I)
 	. = TRUE
 	if(I.tool_use_check(user, 0)) //Don't need to flash eyes because you are a badass
--- a/code/game/objects/items/weapons/cigs.dm
+++ b/code/game/objects/items/weapons/cigs.dm
@@ -68,7 +68,7 @@ LIGHTERS ARE IN LIGHTERS.DM
 
 /obj/item/clothing/mask/cigarette/catch_fire()
 	if(!lit)
-		light("<span class='warning'>Your [name] is lit by the flames!</span>")
+		light("<span class='warning'>The [name] is lit by the flames!</span>")
 
 /obj/item/clothing/mask/cigarette/welder_act(mob/user, obj/item/I)
 	. = TRUE
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -120,7 +120,9 @@
 	else
 		icon = initial(icon)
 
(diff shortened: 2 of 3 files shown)


code/game/objects/items/weapons/cigs.dm
```diff
@@ -66,6 +66,10 @@ LIGHTERS ARE IN LIGHTERS.DM
 	..()
 	light()
 
+/obj/item/clothing/mask/cigarette/catch_fire()
+	if(!lit)
+		light("<span class='warning'>Your [name] is lit by the flames!</span>")
```
dearmochi, line 71: At first glance this message would be displayed to [anyone on the turf](https://github.com/SabreML/Paradise/blob/Fire-Lighting/code/game/objects/items/weapons/cigs.dm#L164-L166). It doesn't make sense to put "You" for those who aren't wearing the cigarette.

code/modules/clothing/clothing.dm
```diff
@@ -120,6 +120,9 @@
 	else
 		icon = initial(icon)
 
+/obj/item/clothing/proc/catch_fire() //Called in handle_fire()
```
dearmochi, line 123: nit: [documentation](https://github.com/tgstation/tgstation/blob/master/.github/AUTODOC_GUIDE.md)

code/modules/mob/living/living_defense.dm
```diff
@@ -186,6 +186,8 @@
 		return 1
 	if(fire_stacks > 0)
 		adjust_fire_stacks(-0.1) //the fire is slowly consumed
+		for(var/obj/item/clothing/C in src.contents)
```
dearmochi, line 189: [Implied src](https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md#legacy-code) is bad. `src.contents` -> `contents`

dearmochi (approved):
Looks OK

code/game/objects/items/weapons/cigs.dm
SabreML: Ah, yep. That's a `visible_message`. One second.

TDSSS:
I wonder if this system could be expanded from just clothes to other items on you, ie plasma in your pockets catching fire, grenades cooking off...

SabreML:
> I wonder if this system could be expanded from just clothes to other items on you, ie plasma in your pockets catching fire, grenades cooking off...

That's sort of what I'm hoping for.
While the`catch_fire()` proc currently only works for clothing, it'd be pretty easy to change that to all items. Although that might need to be on `IgniteMob()` for performance.

SteelSlayer (approved):

variableundefined (approved):

PR #14285, opened 2020-09-12
Turn off xenobiology camera acceleration

Sinestia:
## What Does This PR Do

By default, the eye that the console derives itself from has acceleration turned on, this overrides that default to turn it off. 

## Why It's Good For The Game

Acceleration makes sense for the station AI due to A. the large distance it has to cover and B. most AI interactions (opening doors, bolting, etc.) will involve the use of their mouse, requiring less eye precision.

Xenobiology lacks both of those justifications, the space is small, and it seems pretty common for people to operate it by moving over slimes and clicking on the top buttons. If you do it in that manner, you want precise movement, and that's where acceleration causes unnecessary problems. 

## Changelog
:cl:Froststahr
tweak: Disable default camera acceleration for Xenobiology's Management Console
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/xenobio_camera.dm
+++ b/code/modules/research/xenobiology/xenobio_camera.dm
@@ -68,6 +68,7 @@
 	eyeobj = new /mob/camera/aiEye/remote/xenobio(get_turf(src))
 	eyeobj.origin = src
 	eyeobj.visible_icon = 1
+	eyeobj.acceleration = 0
 	eyeobj.icon = 'icons/obj/abductor.dmi'
 	eyeobj.icon_state = "camera_target"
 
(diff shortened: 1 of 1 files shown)


code/modules/research/xenobiology/xenobio_camera.dm
```diff
@@ -68,6 +68,7 @@
 	eyeobj = new /mob/camera/aiEye/remote/xenobio(get_turf(src))
 	eyeobj.origin = src
 	eyeobj.visible_icon = 1
+	eyeobj.acceleration = 0
```
AffectedArc07, line 71:
```suggestion
	eyeobj.acceleration = FALSE
```

AffectedArc07 (changes requested):

farie82 (approved):

Pixelzilla:
This doesn't seem to really consider the potential impacts of how this would negatively impact those who do choose to expand the console range well outside of its default parameters. It would be terrible trying to actually use the console in that situation due to how large the station is. It isn't difficult to use the console the way that it currently exists and I don't really see it as being a particularly difficult thing to get used to.

AffectedArc07 (approved):

Kyep (approved):

PR #14288, opened 2020-09-12
fixed the toggle guests command

RedWolfie:
## What Does This PR Do
Makes the 'Toggle guests' admin verb work

## Why It's Good For The Game
Fixes a bug. Allows admins to toggle guests.

## Changelog
:cl:
fix: Fixed 'Toggle guests' admin verb
/:cl:

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -304,7 +304,7 @@
 	if(byond_version < SUGGESTED_CLIENT_VERSION) // Update is suggested, but not required.
 		to_chat(src,"<span class='userdanger'>Your BYOND client (v: [byond_version]) is out of date. This can cause glitches. We highly suggest you download the latest client from http://www.byond.com/ before playing. </span>")
 
-	if(IsGuestKey(key))
+	if(IsGuestKey(key) && !GLOB.guests_allowed)
 		alert(src,"This server doesn't allow guest accounts to play. Please go to http://www.byond.com/ and register for a key.","Guest","OK")
 		qdel(src)
 		return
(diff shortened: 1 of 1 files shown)


code/modules/client/client procs.dm
```diff
@@ -304,7 +304,7 @@
 	if(byond_version < SUGGESTED_CLIENT_VERSION) // Update is suggested, but not required.
 		to_chat(src,"<span class='userdanger'>Your BYOND client (v: [byond_version]) is out of date. This can cause glitches. We highly suggest you download the latest client from http://www.byond.com/ before playing. </span>")
 
-	if(IsGuestKey(key))
+	if(IsGuestKey(key) && !GLOB.guests_allowed)
 		alert(src,"This server doesn't allow guest accounts to play. Please go to http://www.byond.com/ and register for a key.","Guest","OK")
 		qdel(src)
 		return
```
AffectedArc07, line 310: Remove this entire proc. We already check if a user is a guest in `world/IsBanned()`, and the `alert()` here wont even show up, since we delete their client the next line.

AffectedArc07 (changes requested):

code/modules/client/client procs.dm
RedWolfie: done

AffectedArc07 (approved):

Kyep (approved):
I guess this okay, so long as its been tested.
Really, the "allow guests" function should be replaced with an auth method not based on byond accounts (for when byond is down) but that's out of scope for this PR.

AffectedArc07:
I will personally test this myself later

PR #14291, opened 2020-09-13
[TGUI] ID Computer

Kyep:
## What Does This PR Do
1. Converts ID computer to TGUI. This also applies to the CC ID card computer, and the department management consoles in all department head offices, both of which are subtypes of the ID computer.
2. Deletes ID card program for laptops and tablets. (But not the card slot hardware)
3. Attempting to insert an ID with no access into the login slot of an ID computer now causes the computer to buzz and reject the ID. This is actually a convenience feature since it stops you inserting the ID into the wrong slot, and then having to take both of them out and swap them around.
4. Clicking an ID computer with a card in hand will now try to intelligently put the card in the correct slot.
5. Encouraged job transfers are now cyan, instead of green. This is because in TGUI, green is a reserved color used for the currently selected button/option.
6. Demoted/terminated IDs show all jobs in discouraged color (gray).
7. CC jobs/skins (when admins see them on CC ID computer) always show in purple.
8. Reworks the access grant/revoke interface such that you can now grant all access to an ID in about 2 clicks, instead of forcing you to make them a Captain or grant every access manually.
9. Now tells you how many free job slots there for each job are on the jobs tab. Its even color-coded.
10. Heads of department can now open/close job slots within their department. They cannot prioritize/unprioritize jobs, but they can see which if any of their department's jobs are prioritized. Opening/closing jobs has a cooldown (like HOP ID computer) and affects which transfer options show up in what colors on the HOP computer.
11. Added a "department" section which shows a list of people who report directly to the currently authorized user. E.g. Captain sees heads in this list, CMO sees medical staff, CE sees engineering staff, etc. Each person is shown with their job title, name, current security status and a demote button. The demote button changes their sec status to demote. In order for this button to be enabled, their current sec status must be one of: none/monitor/released/search. If used, it requires a reason for the demotion. Any use of this ability creates a message for all command staff (via command radio) as well as all online admins. It also notifies the person being demoted that they must go to the head or HOP office (this is to prevent heads using it to get the jump on antags in their departments).
12. Each ID now has a "lastlog" defined on the ID, which is used to show the most recent job transfer done to an ID within the interface. It means that for example if the HOP transfers someone or a head demotes someone, and their ID is put into another computer, it shows the last transfer event in their history.
13. Removed the "print manifest" function (which is useless because everyone has a manifest on their PDA anyway) and the "print ID card details" function (which is also useless, because again, you can see the info in the computer anyway).
14. Transferring someone now automatically notifies the department head of whatever department they are transferred into. E.g: reassigning a station engineer to medical doctor sends an automatic notification to the CMO.
15. In the list of jobs during transfer, the supply jobs (QM, CT, SM) are now shown separately from the rest of the service jobs, to reduce clutter in the service job list. In addition, HOP is shown first in the Service job list, instead of last.
16. Improves and standardizes logging, with a view to allowing admins to search logs about transfers/demotions/terminations more easily.
17. Allows Captain/HOP to terminate an ID using a department console. In practice this can be used by the Cap/HOP to terminate IDs using the HOS' console in their office, saving a trip to bridge/HOPoffice.
18. Fix: the automated notifications sent by the ID computer no longer cause your PDA to flash incorrectly.

## Images of changes
1st tab. Job transfers, card renaming, account number entering and card skin changing.
The Authentication block and tabs at the top are common for all views, but are only shown in this screenshot to reduce the amount of screenshots I have to update later if I make changes:
idcomputer_1_jobtransfer_and_skins
2nd tab, access changing on a card:
idcomputer_2_accessmod
3rd tab, managing job slots:
idcomputer_3_jobslotmanagement
4th tab, records viewing:
idcomputer_4_records
5th tab, department functions:
idcomputer_5_department

## Changelog
:cl: Kyep
tweak: Converted ID computer to TGUI.
add: Improved functionality of ID computer.
/:cl:


procdrone:
I very much like that you can *select all in category*. Nice looking.

code/game/machinery/computer/card.dm
```diff
@@ -227,282 +291,378 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
-	var/mins = round(time_to_wait / 60)
-	var/seconds = time_to_wait - (60*mins)
-	data["cooldown_mins"] = mins
-	data["cooldown_secs"] = (seconds < 10) ? "0[seconds]" : seconds
-
-	if(mode == 3 && is_authenticated(user))
-		data["id_change_html"] = SSjobs.fetch_transfer_record_html(is_centcom())
+	data["iscentcom"] = is_centcom() ? TRUE : FALSE
+	switch(mode)
+		if(0) // JOB TRANSFER
```
AffectedArc07, line 316: Nitpick, should be defines here

code/game/machinery/computer/card.dm
```diff
@@ -227,282 +291,378 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
+			if(modify)
+				if(!scan)
+					// don't gen data
```
AffectedArc07, line 319: Nitpick, `return data` early here instead of a blank statement, or use `pass()`

code/game/machinery/computer/card.dm
```diff
@@ -227,282 +291,378 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
+			if(!j)
+				return FALSE
+			if(!can_open_job(j))
+				return FALSE
+			if(opened_positions[edit_job_target] >= 0)
+				GLOB.time_last_changed_position = world.time / 10
+			j.total_positions++
+			opened_positions[edit_job_target]++
+			log_game("[key_name(usr)] has opened a job slot for job \"[j]\".")
+			message_admins("[key_name_admin(usr)] has opened a job slot for job \"[j.title]\".")
```
AffectedArc07, line 540: Mix of `[j]` and `[j.title]`. Any reason for doing this?

code/game/machinery/computer/card.dm
```diff
@@ -227,282 +291,378 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
+				return FALSE
+			if(!can_close_job(j))
+				return FALSE
+			//Allow instant closing without cooldown if a position has been opened before
+			if(opened_positions[edit_job_target] <= 0)
+				GLOB.time_last_changed_position = world.time / 10
+			j.total_positions--
+			opened_positions[edit_job_target]--
+			log_game("[key_name(usr)] has closed a job slot for job \"[j]\".")
+			message_admins("[key_name_admin(usr)] has closed a job slot for job \"[j.title]\".")
```
AffectedArc07, line 557: Same here

code/game/machinery/computer/card.dm
```diff
@@ -227,282 +291,378 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
+						visible_message("<span class='warning'>[src]: Only the head of this employee may demote them.</span>")
+						return FALSE
+					for(var/datum/data/record/R in GLOB.data_core.security)
+						if(R.fields["id"] == E.fields["id"])
+							if(status_valid_for_demotion(R.fields["criminal"]))
+								set_criminal_status(usr, R, SEC_RECORD_STATUS_DEMOTE, reason, scan.assignment)
+								Radio.autosay("[scan.registered_name] ([scan.assignment]) has set [tempname] ([temprank]) to demote for: [reason]", name, "Command", list(z))
+								message_admins("[key_name_admin(usr)] ([scan.assignment]) has set [tempname] ([temprank]) to demote for: [reason]")
+							else
+								to_chat(usr, "<span class='warning'>Cannot demote, due to their current sec status.</span>")
```
AffectedArc07, line 581: Nitpick, `security record` instead of just `sec`

code/game/machinery/computer/card.dm
```diff
@@ -227,282 +291,378 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
...
-			if(is_authenticated(usr) && is_centcom())
-				var/delcount = SSjobs.delete_log_records(scan.registered_name, FALSE)
-				if(delcount)
-					playsound(src, 'sound/machines/terminal_prompt_confirm.ogg', 50, 0)
-				SSnanoui.update_uis(src)
 
-		if("wipe_all_logs")
+		if("wipe_all_logs") // Delete all records from 'records' section
 			if(is_authenticated(usr) && !target_dept)
 				var/delcount = SSjobs.delete_log_records(scan.registered_name, TRUE)
```
AffectedArc07, line 661: Did you make this still keep a copy of logs for admins

tgui/packages/tgui/interfaces/CardComputer.js
```diff
@@ -0,0 +1,543 @@
...
+        {authBlock}
+        {bodyBlock}
+      </Window.Content>
+    </Window>
+  );
+
+
+
+
+
```
AffectedArc07, line 542: Trim this whitespace a bit

AffectedArc07 (changes requested):
Few nitpicks

code/game/machinery/computer/card.dm
Kyep: CC card computer can still view "deleted logs".
"Delete" functions only mark the records as hidden from non-CC computers.

AffectedArc07 (approved):

PR #14296, opened 2020-09-14
Fixes NO_SCAN cloning exploit through replica pods

dearmochi:
## What Does This PR Do

Removes the ability to clone species that have the `NO_SCAN` trait (namely Vox) using replica pods. Upon trying to inject the seeds with such blood, the client will get a generic rejection message.

## Why It's Good For The Game

This method differs from Stable Mutagen cloning which requires the original brain. Replica pods don't require the brain for the client to be moved into their new body, and that's an exploit.

## Changelog
:cl:
fix: Fix an exploit allowing the cloning of unclonable species through the use of replica pods
/:cl:


Kyep (approved):

kenken244:
this isn't an exploit, its intended that non-scannable species can be turned into diona through this. they aren't even a non-scannable species anymore, why should that restriction apply? this is a balance change, and should be reverted until it is properly discussed

PR #14309, opened 2020-09-15
Fixes #14302

### LINKED ISSUE #14302: Wrong access on syndie depot turret control
**Issue Description**:
A player with a 'syndicate operative' ID card was unable to use the external turret controller on the Syndie depot

for the ID to let you unlock the controller

**What happened instead**:
It didn't

You can't do the smart thing

Spawn in a 'syndicate operative' ID card and try to unlock the turret controller on the right of the depot, outside.

**Extra information**:
Capture
I used VV to take a look. On the right you see the turret controller that doesn't work. On the left, you see an internal turret controller as comparison. In the middle, you see the Operative ID card, which has the access it would need.

I assume the cause is incorrectly var-editing the turret controller in question.

TDSSS:
## What Does This PR Do
Fixes #14302 by setting the access via text rather than the number.

Not sure why it has to be set this way but it seems varediting and req_access don't mix.

## Why It's Good For The Game
bugs bad

## Changelog
:cl:
fix: Syndie depot turret controller can now be correctly unlocked with syndie access.
/:cl:

### DIFF
--- a/_maps/map_files/RandomRuins/SpaceRuins/syndiedepot.dmm
+++ b/_maps/map_files/RandomRuins/SpaceRuins/syndiedepot.dmm
@@ -68,7 +68,6 @@
 "an" = (
 /obj/machinery/navbeacon/invisible{
 	codes_txt = "patrol;next_patrol=SDNW";
-	invisibility = 100;
 	location = "SDNE"
 	},
 /turf/simulated/floor/plating/asteroid/airless,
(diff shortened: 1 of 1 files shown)


TDSSS:
The other edits seem to be strongDMM removing varedits that set a var to the value it already was. IE, normaldoorcontrol was already 0 anyway, etc. If that's an issue I could redo the edit and turn that off (at least I think it can be turned off)

PR #14316, opened 2020-09-15
Allows Shadowlings to disable 'Light Floor' tiles

SabreML:
## What Does This PR Do

This allows Shadowlings to disable 'Light Floor' tiles by using their 'Veil' ability.

Also fixes a bug where floodlights would keep their 'on' sprite after being disabled by a sling.
## Why It's Good For The Game

Shadowlings can, as far as I know, turn off all electronic light sources *except* for floor lights. Which really seems like an oversight to me.

## Changelog
:cl:
tweak: Shadowlings can now disable floor lights with their 'Veil' ability
fix: Fixed floodlights disabled by slings keeping their 'on' sprite
/:cl:

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -112,9 +112,7 @@
 		G.visible_message("<span class='warning'>[G] withers away!</span>")
 		qdel(G)
 	for(var/turf/T in targets)
-		if(istype(T, /turf/simulated/floor/light))
-			var/turf/simulated/floor/light/L = T
-			L.toggle_light(0)
+		T.extinguish_light()
 		for(var/atom/A in T.contents)
 			A.extinguish_light()
 
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -109,9 +109,12 @@
 		return
 	to_chat(user, "<span class='shadowling'>You silently disable all nearby lights.</span>")
 	for(var/obj/structure/glowshroom/G in orange(2, user)) //Why the fuck was this in the loop below?
-		G.visible_message("<span class='warning'>\The [G] withers away!</span>")
+		G.visible_message("<span class='warning'>[G] withers away!</span>")
 		qdel(G)
 	for(var/turf/T in targets)
+		if(istype(T, /turf/simulated/floor/light))
+			var/turf/simulated/floor/light/L = T
+			L.toggle_light(0)
 		for(var/atom/A in T.contents)
 			A.extinguish_light()
 
--- a/code/game/turfs/simulated/floor/light_floor.dm
+++ b/code/game/turfs/simulated/floor/light_floor.dm
@@ -11,7 +11,7 @@
 #define LIGHTFLOOR_CYCLEB 10
 
 /turf/simulated/floor/light
-	name = "Light floor"
+	name = "\improper light floor"
 	light_range = 5
 	icon_state = "light_on"
 	floor_tile = /obj/item/stack/tile/light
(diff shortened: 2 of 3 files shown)


code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -109,9 +109,12 @@
 		return
 	to_chat(user, "<span class='shadowling'>You silently disable all nearby lights.</span>")
 	for(var/obj/structure/glowshroom/G in orange(2, user)) //Why the fuck was this in the loop below?
-		G.visible_message("<span class='warning'>\The [G] withers away!</span>")
+		G.visible_message("<span class='warning'>[G] withers away!</span>")
 		qdel(G)
 	for(var/turf/T in targets)
+		if(istype(T, /turf/simulated/floor/light))
```
farie82, line 115: Best to just call `T.extinguish_light()` here

code/game/turfs/simulated/floor/light_floor.dm
```diff
@@ -103,6 +102,19 @@
...
+	// 1 = ON
+	if(state == 1)
+		on = TRUE
+	// 0 = OFF
+	else if(state == 0)
+		on = FALSE
+	update_icon()
+
+/turf/simulated/floor/light/extinguish_light()
+	toggle_light(0)
```
farie82, line 115: `FALSE`

code/game/turfs/simulated/floor/light_floor.dm
```diff
@@ -103,6 +102,19 @@
 	else
 		to_chat(user, "<span class='warning'>[src]'s light bulb appears to have burned out.</span>")
 
+/turf/simulated/floor/light/proc/toggle_light(state)
+	// 1 = ON
+	if(state == 1)
```
farie82, line 107: Just a `on = state` will do here.

code/game/turfs/simulated/floor/light_floor.dm
```diff
@@ -69,14 +69,13 @@
 	set_light(0)
 	..()
 
-/turf/simulated/floor/light/attack_hand(mob/user)
+/turf/simulated/floor/light/attack_hand()
```
farie82, line 72: Keep the param here as it's defined on it's parent. Not using it is no issue but omitting it here can lead to issues down the line

farie82 (changes requested):

code/game/turfs/simulated/floor/light_floor.dm
SabreML: What about for line 77, the `params`?

farie82: Same stuff.
image
The proc is defined with it so it's best to keep atleast the parameters given by the parents proc and above.

It can break if for example this happens:
```dm
/proc/test()
   attack_hand(user = usr)
```
If it calls the attack_hand on the light turf then it will runtime due to the argument missing

code/game/turfs/simulated/floor/light_floor.dm
SabreML: Ah yeah, some of this is left over from failed ideas.

Mitchs98:
While it does sort of seem like an oversight, I'm unconvinced Shadowlings need a buff. As well as the disabling of light floors seems like something that should be delegated to thralls to do as a sort of teamwork thing. Shadowlings already seem powerful enough to me on their own, they should need to rely on their thralls/team in order to tackle certain things more than they already do in my opinion and this just seems like a step in the wrong direction.

Kyep (approved):

Cocacolagua:
Shadowlings do indeed need a mejor re-vamp, but the veil ability should be consistent and turn off all lights, choosing between some and some is pure newbie bait, which the gamemode already has tons of. Good change.

taukausanake:
Maybe make floor lights hook up to the lighting in an area first? Or some other area power through the APC? Just having a perpetual light source during power outages or any other area you want dark can be an issue

code/game/turfs/simulated/floor/light_floor.dm
```diff
@@ -103,6 +102,16 @@
 	else
 		to_chat(user, "<span class='warning'>[src]'s light bulb appears to have burned out.</span>")
 
+/turf/simulated/floor/light/proc/toggle_light(light)
+	// 0 = OFF
```
farie82, line 106: I'd say that this is logical due to how the proc is set up. Not sure if the comments add something

farie82 (approved):
LGTM.
Nitpick on the comments but that's fine

variableundefined (approved):
Since shadowling can extinguish most other artificial lights, this PR increases consistency. I support it.

PR #14317, opened 2020-09-15
Fixes canister admin alerts

Kyep:
## What Does This PR Do
1) ~~Stops admins getting an alert about someone opening the valve on a canister - if the canister is currently connected to a floor canister port. This is a common event (toxins, atmos, etc) and not worthy of an alert.~~ Removed due to AA request.
2) Stops admins getting an alert about someone removing their tank from a canister with an open valve - if the canister does not contain any plasma or N2O. Again, people forgetting to actually close the valve before removing their tank is a common event, and its not worthy of an admin alert. While it might overpressurize the room, that's not a big issue. Only cases where someone is doing this deliberately to release N2O/plasma are worthy of an alert to admins.

## Changelog
:cl: Kyep
fix: stopped admins getting alerts about perfectly innocent/harmless interactions with canisters.
/:cl:

### DIFF
--- a/code/game/machinery/atmoalter/canister.dm
+++ b/code/game/machinery/atmoalter/canister.dm
@@ -352,7 +352,7 @@ update_flag
 			valve_open = !valve_open
 			if(valve_open)
 				logmsg = "Valve was <b>opened</b> by [key_name(usr)], starting a transfer into the [holding || "air"].<br>"
-				if(!holding && !connected_port)
+				if(!holding)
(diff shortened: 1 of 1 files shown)


AffectedArc07:
```
Stops admins getting an alert about someone opening the valve on a canister - if the canister is currently connected to a floor canister port. This is a common event (toxins, atmos, etc) and not worthy of an alert.
```
If I recall correctly, this still leaks gas out, so I would keep this in

PR #14246, opened 2020-09-06
Adds auto-reconnect chat feature on server reboot

dearmochi:
## What Does This PR Do

When the server is rebooted (through `/world/Reboot`), a chat message now appears saying the server is rebooting. That message includes a link to manually reconnect as well as a timer that automatically reconnects the client after a while. 
That timer can be adjusted on the DM side to be longer if an update is pending, or something.

## Why It's Good For The Game

Removes the need for players to wait some time and manually reconnect after a server reboot.

## Images of changes

fFaXef0AIk

## Changelog
:cl:
add: Add auto-reconnect feature for when the server restarts
/:cl:

PR #14318, opened 2020-09-15
Server awareness of pending updates

Kyep:
## What Does This PR Do
If a PR is merged, OR hostannounce is used to send a message to the server, clients will wait 60s before reconnecting to the server, instead of the normal 10s.
Extends #14246 

## Why It's Good For The Game
Means clients can autoreconnect successfully after the server is patched.

## Changelog
:cl: Kyep
tweak: server auto-reconnect now accounts for when downtime is a bit longer due to pending updates.
/:cl:

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -223,7 +223,7 @@ GLOBAL_VAR_INIT(world_topic_spam_protect_time, world.timeofday)
 				return "Bad Key"
 			else
 				var/prtext = input["announce"]
-				if(findtext(prtext, "Pull request merged by") != 0)
+				if(copytext(prtext, 1, 25) == "Pull requested merged by")
 					GLOB.pending_server_update = TRUE
 				for(var/client/C in GLOB.clients)
 					to_chat(C, "<span class='announce'>PR: [prtext]</span>")
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -223,7 +223,8 @@ GLOBAL_VAR_INIT(world_topic_spam_protect_time, world.timeofday)
 				return "Bad Key"
 			else
 				var/prtext = input["announce"]
-				if(copytext(prtext, 1, 23) == "Pull+Request+merged+by") // its urlencoded
+				var/pr_substring = copytext(prtext, 1, 23)
+				if(pr_substring == "Pull Request merged by")
 					GLOB.pending_server_update = TRUE
 				for(var/client/C in GLOB.clients)
 					to_chat(C, "<span class='announce'>PR: [prtext]</span>")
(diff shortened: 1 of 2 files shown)


code/game/world.dm
```diff
@@ -222,8 +222,11 @@ GLOBAL_VAR_INIT(world_topic_spam_protect_time, world.timeofday)
 			if(input["key"] != config.comms_password)
 				return "Bad Key"
 			else
+				var/prtext = input["announce"]
+				if(findtext(prtext, "Pull request merged by") != 0)
```
AffectedArc07, line 226: Someone could open a PR with the name `Pull request merged by` and cheat the system here.

Dumbdumn5 (approved):

code/game/world.dm
```diff
@@ -339,8 +342,12 @@ GLOBAL_VAR_INIT(world_topic_spam_protect_time, world.timeofday)
 	return
 	#endif
 
+	var/secs_before_auto_reconnect = 10
+	if(GLOB.pending_server_update)
+		secs_before_auto_reconnect = 60
```
farie82, line 347: Could be a made a config option. Would also allow the automatic changing of the waiting time depending on how long it took the times before hand.
All very overkill for what it is though

farie82 (approved):
LGTM. Could make the time a config option though but that's optional

code/game/world.dm
AffectedArc07: @farie82 it takes 60 seconds for an update by default, if you include the restart overhead too. A config option is not only necessary for this, it wouldnt be useful since the backend doesnt modify the config in any way.

farie82: That's on our hardware on our network. It can take longer I'd guess on other setups and seeing this is an open source repo. But like I said it's fully optional

AffectedArc07: @farie82 This entire feature is also custom-specific to our backend tooling. I only know of one "real" downstream of us, and they host on TGS, which doesnt have reboot delays.

code/game/world.dm
```diff
@@ -222,8 +222,11 @@ GLOBAL_VAR_INIT(world_topic_spam_protect_time, world.timeofday)
 			if(input["key"] != config.comms_password)
 				return "Bad Key"
 			else
+				var/prtext = input["announce"]
+				if(copytext(prtext, 1, 23) == "Pull+Request+merged+by") // its urlencoded
+					GLOB.pending_server_update = TRUE
```
AffectedArc07, line 227:
```suggestion
				var/prtext = input["announce"]
				var/pr_substring = copytext(prtext, 1, 23)
				if(pr_substring == "Pull Request merged by")
					GLOB.pending_server_update = TRUE
```
image
This configuration works, and I tested it with the script itself

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #14335, opened 2020-09-16
Adds a new mini gamemode to the traitor gamemode. The VIP target

farie82:
## What Does This PR Do
Adds a new combination of objectives which lead to one crew member being special. He will be both loved by traitors and hated.
He will be the target for 1 or more traitors to protect and target for 0 or more traitors to assassinate.
- When there are 3 or more existing traitors this "gamemode" will become available
- One traitor will be chosen to become the first protector. He will get a protection objective. The target of that objective will become the VIP
- Each new traitor will have a chance to become either a protector or assassin. They will become an assassin if the assassin count is lower than the protector count (with a 80% chance minus 10 \* protector count with a minimum of 50%). They will become a protector if the protector count is equal or less than the assassin count (with a 30% chance minus 10 \* assassin count with a minimum of 10%).
- Protectors will get the following objective text: "Protect NAME, the ROLE. They are believed to be hunted by agents of other companies."

They won't know about the others who protect the VIP or hunt the VIP. Nor will the VIP know he's special that round.
There can be a maximum of 1 difference between the assassins and protectors. The protectors will have X or X+1 members and the assassins will have X members. X being a number between 0 and infinity (god please don't let it go that high)

Admins can also force a target using the Check Antagonists verb. It will update all the existing objectives to take the new VIP target as their target. 
Admins can also add the VIP objectives to people at will.

Refactors the objective code as well to make this possible and to keep me sane.
Also fixes a bug where rolling a maroon/destruction/debrain objective twice could give you the same target. This could happen because the objective datum was put in the targets list (which is used to check if you already have a target) instead of the actual target.
Also refactors the cryosleep code and mind topic code. Objectives were sometimes not removed and what not. 

Fixes a bug where you could get the same target when your target cryos. Breaking the system.
Fixes a bug where failed objectives would still be around linked to the "owner"
Fixes a bug where you could get the AI as a target again after it just wiped it's core

## Why It's Good For The Game
Adds some more interaction between the crew and the antags. The antags will have to fight against each other which can lead to fun situations and the VIP will be in the middle of it.

## Images of changes
image
Image of two agents getting the VIP target. Tested with 3 clients and 2 agents. I've removed the 3 agents minimum to test this
Older version, The protect objective text has changed to "Protect NAME, the ROLE. They are believed to be hunted by agents of other companies."

image
Admin ~~abuse~~ check antagonists panel

image
Antag HUD icon. Made by @smokincheez

## Changelog
:cl: Farie82 smokincheez
add: Adds a new mini gamemode to the traitor gamemode. The VIP. Agents will get one global target. Either to protect or to assassinate. One VIP defended and hunted by multiple agents.
add: Admins can set the VIP target in the Check Antagonists verb
add: Admins can add the VIP objectives to people at will using the player panel
fix: Fixed an edge case where you could get the same maroon/destruction or debrain objective more than once
fix: Fixed stray objectives sticking around when creating them failed. Causing random objective changed alerts when somebody cryos.
fix: You can't get the same target who just cryod now. Crewmembers in cryo tubes are not valid targets for the objective system
fix: You can't get the same AI as target when they just wiped their core
imageadd: Antag HUD icon for the VIP
/:cl:


Cocacolagua:
How does this work with the all-crew-are-antags secret button?
Also, what about letting the protectors and assassins the members on their team? So they can coordinate a defense / attack easier.
Other than that, fun idea, can’t wait for a changeling-traitor to get the assassinate objective and rip the other team apart.

Nazsgull:
Nice! Traitors not having always to kill is a good thing!

TDSSS:
1. can the VIP be a traitor?
2. What happens if the VIP cryos?

farie82:
> 1. can the VIP be a traitor?
> 2. What happens if the VIP cryos?

1. He can be a traitor.
2. I've not thought of that yet. Will fix it! Thanks

farie82:
> How does this work with the all-crew-are-antags secret button?

It will only work with roundstart agents just like the document trading objectives


> Also, what about letting the protectors and assassins the members on their team? So they can coordinate a defense / attack easier.

Maybe. I left them uninformed to create suspicion among them. Assassins can pretend to be protectors and that way earn trust with the other protectors.
I might give the assassins their coworkers to give them a small bonus seeing they are in the disadvantage here. Having to attack and be illegal and all

farie82:
Updated:
Cryoing targets won't break the system now. Worst case is that a protector or assassin gets chosen to become the next target. But this will just mean that he loses his objective.
Also some more fixes but that's less exciting

taukausanake:
Any chance we can get that to cycle the alphabet? It would be hilarious if you manage to go through 26 targets.

"Time for plan Y."
"Time for plan Z."
"Time for plan... SHIT."

farie82:
> Any chance we can get that to cycle the alphabet? It would be hilarious if you manage to go through 26 targets.
> 
> "Time for plan Y."
> "Time for plan Z."
> "Time for plan... SHIT."

I'll leave that up for somebody else to manage. It's currently fully randomized with no persistence so yeah

code/_globalvars/game_modes.dm
```diff
@@ -9,3 +9,7 @@ GLOBAL_VAR(custom_event_msg)
 GLOBAL_VAR(custom_event_admin_msg)
 
 GLOBAL_LIST_EMPTY(summon_spots)
+
+GLOBAL_VAR(protect_target) // The one sucker that will be fought over by antags
+GLOBAL_LIST_EMPTY(protect_target_assassins) // Who is going to hunt them
+GLOBAL_LIST_EMPTY(protect_target_protectors) // Who is going to protect them
```
SteelSlayer, line 15: If this mini-gamemode is suppose to be traitor only, I'm thinking these would be better as vars on 
`datum/gamemode/traitor`. And then can be accessed by `SSticker.mode.protect_target` etc.

farie82: I thought about putting them on the gamemode but I decided to make them global to make expansion possible for later if we want to get rid of mode bound antags and such.
I'll think about it though since currently finding who is the target can be a bit annoying as admin. Having to navigate the global variables and the like (or just finding somebody with protect)

TDSSS:
You say it's traitor only, is that an actual restriction? Like would the objective break if it was manually assigned to a cling or vamp, or during a tatorling round etc?

Spacemanspark:
Heh, time to select one person and set the entire station on a war against them.

farie82:
> You say it's traitor only, is that an actual restriction? Like would the objective break if it was manually assigned to a cling or vamp, or during a tatorling round etc?

It would not. If you assign the VIP objective to somebody they will get the same target and it will all work out. They won't be part of the assassins/protectors list though. I'll see about making it more modular so admins can fuck around with it as well

code/game/gamemodes/objective.dm
```diff
@@ -293,13 +327,17 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 			var/mob/living/carbon/human/H = possible_target.current
 			if(!(NO_DNA in H.dna.species.species_traits))
 				possible_targets += possible_target
-	if(possible_targets.len > 0)
+	var/found_target
+	if(length(possible_targets) > 0)
 		target = pick(possible_targets)
+	set_target(found_target)
+	return found_target
```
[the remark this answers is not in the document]
farie82: Good find. Completely overlooked that one myself

code/_globalvars/game_modes.dm
farie82: Changed it in the latest commit. It made more sense to put it on the gamemode datum in the end. Also avoids more globals that could be put under other objects

farie82:
Added the ability for admins to select a VIP using the Check Antagonists panel. Admins can also give the VIP objectives to people now.
Also some more refactors and what not to accompany the change

code/datums/mind.dm
```diff
@@ -19,6 +19,10 @@
 		yourself.
 */
 
+#define OBJECTIVE_LIST list("assassinate", "blood", "debrain", "protect", "prevent", "brig", "hijack",\
```
TDSSS, line 22: Would it make sense to use defined here instead of strings so typos get noticed immediately etc? Or even objective types directly.

farie82: That would be cleaner indeed. I'll see about refactoring it

code/game/gamemodes/objective.dm
```diff
@@ -224,6 +232,12 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 	if(target && owner)
 		SSticker.mode.protect_target_protectors += owner
 
+/datum/objective/protect/vip/set_explaination_text()
+	if(target && target.current)
+		explanation_text = "Protect [target.current.real_name], the [target.assigned_role]. They are believed to be hunted by agents of other companies."
+	else
+		..()
```
SteelSlayer, line 239: Any reason why some of the parent calls for `set_explanation_text()` are `return ..()`, and some are just plain `..()`?

farie82: No real reason other than me going on auto pilot it seems :P
Will be fixed in the next commit!

Necaladun:
GET DOWN MR PRESIDENT

I like it

SteelSlayer:
I really like your idea for objective creation. I'm trying to clean up antag datums and the traitor datum right now, because I've been meaning to do that, and I see that more people are working with them now. It will unfortunately conflict with your PR a bit, hopefully yours is merged before mine so I can deal with that though.

Also side note, I think it would be neat if the `create_objective()` proc and the `assigned_targets` list was on the 
`/datum/antagonist` level instead. You don't have to do this but it's something I plan on doing if you don't.

farie82:
> I really like your idea for objective creation. I'm trying to clean up antag datums and the traitor datum right now, because I've been meaning to do that, and I see that more people are working with them now. It will unfortunately conflict with your PR a bit, hopefully yours is merged before mine so I can deal with that though. Or better yet, I hope this is merged before I even put my PR up so I can incorporate this.

Yes please do. I'd not mind deconflicting if needed :D
 
> Also side note, I think it would be neat if the `create_objective()` proc and the `assigned_targets` list was on the
> `/datum/antagonist` level instead. You don't have to do this but it's something I plan on doing if you don't.

Done and done. It is a better place indeed since it's not tied to just traitors. I've left out making all the other code use this since that'll be a bit much for this PR

code/game/gamemodes/objective.dm
```diff
@@ -17,8 +17,20 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
...
 	return ..()
 
+/**
+ * Will try to renew the objective to a valid state with another target.
+ * Returns the new found target or null if none were found
+ */
+/datum/objective/proc/on_target_loss()
+	if(owner?.current)
+		to_chat(owner.current, "<BR><span class='userdanger'>You get the feeling your target is no longer within reach. Time for Plan [pick("A","B","C","D","X","Y","Z")]. Objectives updated!</span>")
+		SEND_SOUND(owner.current, 'sound/ambience/alarm4.ogg')
```
SteelSlayer, line 31: I'm not 100% sure, but I think `playsound()` is preferred over this, just judging by how many more instances of it there are than `SEND_SOUND`.

code/game/gamemodes/objective.dm
```diff
@@ -29,41 +41,55 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 		return TARGET_INVALID_IS_TARGET
 	if(!ishuman(possible_target.current))
 		return TARGET_INVALID_NOT_HUMAN
-	if(!possible_target.current.stat == DEAD)
-		return TARGET_INVALID_DEAD
 	if(!possible_target.key)
 		return TARGET_INVALID_NOCKEY
 	if(possible_target.current)
+		if(!possible_target.current.stat == DEAD)
```
SteelSlayer, line 47: I think the `!` might break things here. For example `if(!x == 2)` (2 being the value for `DEAD`) equates to 0, which means the if never succeeds, no matter the `stat` value.

code/game/gamemodes/objective.dm
```diff
@@ -76,17 +102,38 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 		return 0
 	return 1
 
+/datum/objective/assassinate/vip/Destroy()
+	SSticker.mode.protect_target_assassins -= owner
+	. = ..()
```
SteelSlayer, line 107:
```suggestion
	return ..()
```

code/game/gamemodes/objective.dm
```diff
@@ -491,16 +577,18 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
...
 	if(target && target.current)
 		target_real_name = target.current.real_name
 		explanation_text = "Destroy [target_real_name], the AI."
 	else
-		explanation_text = "Free Objective"
-	return target
+		..()
+
+/datum/objective/destroy/find_target()
+	var/list/possible_targets = active_ais(1)
```
SteelSlayer, line 588: This proc doesn't have any arguments.
```suggestion
	var/list/possible_targets = active_ais()
```

code/game/gamemodes/objective.dm
farie82: oh wow. I just copy pasted it without checking really. But yeah... odd how this has been like this for so long. Will fix

code/game/gamemodes/objective.dm
farie82: `playsound` will make it hearable for all others nearby as well. `playsound_local` is an option though which is used by the AI and traitor intro sound... oddly enough `SEND_SOUND` is used by the changeling intro sound.
I'll use `playsound_local` here in the same way the agents do it!

code/modules/antagonists/traitor/datum_traitor.dm
```diff
@@ -125,8 +114,23 @@
 			add_objective(hijack_objective)
 			return
 
+	// See if they get the global target.
+	if(objective_count <= objective_amount && length(SSticker.mode.traitors) >= 3 && SSticker.mode.VIP_target != owner)
+		var/assassin_count = length(SSticker.mode.protect_target_assassins)
+		var/protector_count = length(SSticker.mode.protect_target_protectors)
+
+		// Make a new protector if there is no target set yet (first protector) or at a 20% chance if there are less or equal protectors compared to assassins
+		if(!SSticker.mode.VIP_target || (protector_count <= assassin_count && prob(10)))
```
SteelSlayer, line 123: There's a mismatch here between the comment and code. Comment says 20% chance, code is 10%.

code/game/gamemodes/objective.dm
```diff
@@ -47,7 +51,7 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 		if(possible_target.current.stat == DEAD)
 			return TARGET_INVALID_DEAD
 		var/turf/current_location = get_turf(possible_target.current)
-		if(current_location && !is_level_reachable(current_location.z))
+		if(!current_location || !is_level_reachable(current_location.z))
```
farie82, line 54: Fixes targets that just cryod being chosen as potential target

code/modules/antagonists/traitor/datum_traitor.dm
farie82: Changed in the last PR. Danke!

farie82:
TM was done. One issue found with the check antag panel. It's fixed now.
I've also increased the chances of getting a new protector/assassin. This chance will however decrease the more there already are active

code/modules/mob/mob.dm
```diff
@@ -192,6 +192,25 @@
 	return 0
 
 
+/**
+ * Will handle a player cryoing or leaving the round in a similar way
+ *
+ */
+/mob/proc/handle_removal()
```
SteelSlayer, line 199: Nitpick: This proc name is a bit generic, maybe something like `on_removal_from_round()`?

code/modules/admin/topic.dm
```diff
@@ -388,6 +388,28 @@
 			SSticker.real_reboot_time = 0 // If they set this at round end, show the "Reboot was cancelled by an admin" message instantly
 		href_list["secretsadmin"] = "check_antagonist"
 
+	else if(href_list["change_vip_target"])
+		if(!check_rights(R_ADMIN))
+			return
+		var/list/possible_targets = list()
+		for(var/datum/mind/possible_target in SSticker.minds)
+			if((possible_target != src) && istype(possible_target.current, /mob/living/carbon/human))
```
SteelSlayer, line 396:
```suggestion
			if((possible_target != src) && ishuman(possible_target.current)
```

SteelSlayer (commented):
Looks good, some final things

SteelSlayer (approved):

code/game/gamemodes/objective.dm
```diff
@@ -517,16 +587,19 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
...
-		explanation_text = "Free Objective"
-	return target
+		..()
+
+/datum/objective/destroy/find_target()
+	var/list/possible_targets = active_ais()
+	var/mob/living/silicon/ai/target_ai = safepick(possible_targets)
+	var/found_target = target_ai?.mind
+	set_target(found_target)
+	return found_target
```
SabreML, line 602: Similar to a bit above, seems like it might be better to just return early if there's no active AIs rather than putting failsafes in.
```suggestion
/datum/objective/destroy/find_target()
	var/list/possible_targets = active_ais()
	if(!length(possible_targets))
		return null
	var/found_target = pick(possible_targets)
	set_target(found_target)
	return found_target
```

code/modules/admin/topic.dm
```diff
@@ -388,6 +388,28 @@
 			SSticker.real_reboot_time = 0 // If they set this at round end, show the "Reboot was cancelled by an admin" message instantly
 		href_list["secretsadmin"] = "check_antagonist"
 
+	else if(href_list["change_vip_target"])
+		if(!check_rights(R_ADMIN))
+			return
+		var/list/possible_targets = list()
+		for(var/datum/mind/possible_target in SSticker.minds)
+			if((possible_target != src) && ishuman(possible_target.current))
```
SabreML, line 396: Probably only minds in `SSticker.minds`. (If there's not, then that's a bigger issue)
```suggestion
		for(var/mind in SSticker.minds)
			var/datum/mind/possible_target = mind
			if((possible_target != src) && ishuman(possible_target.current))
```

code/modules/admin/topic.dm
```diff
@@ -388,6 +388,28 @@
...
+				possible_targets += possible_target.current
+
+		possible_targets = sortAtom(possible_targets)
+		var/mob/living/carbon/human/new_target = input("Select new VIP:", "Objective target") as null|anything in possible_targets
```
SabreML, line 400: No option to unset the target entirely? You can cancel, but that still keeps the current one.

code/modules/admin/topic.dm
```diff
@@ -388,6 +388,28 @@
...
+		if(!new_target)
+			return
```
SabreML, line 402: Since it's after an `input()`, you might want to make sure they've not been deleted or otherwise disappeared in the wait time.
```suggestion
		if(QDELETED(new_target))
			return
```

code/modules/admin/topic.dm
```diff
@@ -388,6 +388,28 @@
...
+		SSticker.mode.change_vip_target(new_target.mind)
+		// Update the existing objectives
+		for(var/thing in GLOB.all_objectives)
+			if(!istype(thing, /datum/objective/protect/vip) && !istype(thing, /datum/objective/assassinate/vip))
+				continue
+			var/datum/objective/O = thing
+			O.on_target_loss()
+		check_antagonists()
+		return
```
SabreML, line 411: You shouldn't need to specify return here since everything afterwards is a 3000 line `else if()`.

code/modules/antagonists/_common/antag_datum.dm
```diff
@@ -56,6 +57,36 @@ GLOBAL_LIST_EMPTY(antagonists)
 /datum/antagonist/proc/create_team(datum/team/team)
 	return
 
+/**
+ * Tries to create a new objective
+ * Will return the objective if it succeeds. Will return null if it fails
+ * type - The objective type that will be made
+ * target_override - Target that will be used by the objective. Default is null meaning it'll be decided by the objective.find_target
+ */
+/datum/antagonist/proc/create_objective(type, target_override = null)
```
SabreML, line 66:
```suggestion
/**
 * Tries to create a new objective
 * Will return the objective if it succeeds. Will return null if it fails
 * * type - The objective type that will be made
 * * target_override - Target that will be used by the objective. Default is null meaning it'll be decided by the objective.find_target
 */
/datum/antagonist/proc/create_objective(type, target_override = null)
```

code/modules/antagonists/_common/antag_datum.dm
```diff
@@ -56,6 +57,36 @@ GLOBAL_LIST_EMPTY(antagonists)
...
+	O.owner = owner
+	var/target = target_override || O.find_target()
+	if(target_override)
+		O.set_target(target_override)
+	var/target_text = "[target]"
+	if(target_text in assigned_targets)		// Is this target already in their list of assigned targets? If so, don't add this objective and return
+		qdel(O)	// Actually delete the objective else stray references will remain
+		return null
+	else if(target)							// Is the target a real one and not null? If so, add it to our list of targets to avoid duplicate targets
+		assigned_targets.Add(target_text)	// This logic is applied to all traitor objectives including steal objectives
```
SabreML, line 78: `+=` is a little faster than `Add()`.
```suggestion
		assigned_targets += target_text	// This logic is applied to all traitor objectives including steal objectives
```

code/modules/antagonists/_common/antag_datum.dm
```diff
@@ -82,6 +113,7 @@ GLOBAL_LIST_EMPTY(antagonists)
 		owner.current.key = C.key
 
 /datum/antagonist/proc/on_removal()
+	assigned_targets.Cut()
 	remove_innate_effects()
```
SabreML, line 117: Is wiping the list neccesary here? Since it's being deleted anyway a few lines later.

code/modules/antagonists/traitor/datum_traitor.dm
```diff
@@ -169,75 +174,43 @@
...
-			var/datum/objective/destroy/destroy_objective = new
-			destroy_objective.owner = owner
-			destroy_objective.find_target()
-			if("[destroy_objective]" in assigned_targets)	        // Is this target already in their list of assigned targets? If so, don't add this objective and return
-				return 0
-			else if(destroy_objective.target)					    // Is the target a real one and not null? If so, add it to our list of targets to avoid duplicate targets
-				assigned_targets.Add("[destroy_objective.target]")	// This logic is applied to all traitor objectives including steal objectives
-			add_objective(destroy_objective)
-
+		if(length(active_ais) && prob(100 / length(GLOB.player_list)))
```
SabreML, line 196: You can move the `active_ais` variable to here and save a line.
```suggestion
		if(length(active_ais()) && prob(100 / length(GLOB.player_list)))
```

code/modules/mob/living/silicon/ai/latejoin.dm
```diff
@@ -13,19 +13,9 @@ GLOBAL_LIST_EMPTY(empty_playable_ai_cores)
...
-	SSjobs.FreeRole(job)
-
-	if(mind.objectives.len)
-		mind.objectives.Cut()
-		mind.special_role = null
-	else
-		if(SSticker.mode.name == "AutoTraitor")
-			var/datum/game_mode/traitor/autotraitor/current_mode = SSticker.mode
-			current_mode.possible_traitors.Remove(src)
+	control_disabled = TRUE // Ensure the AI is not chosen again as target
```
SabreML, line 16: I might be missing something, but as far as I can tell this variable is never actually referenced in objective code?

SabreML (commented):
:eyes:

code/game/gamemodes/objective.dm
```diff
@@ -17,8 +17,25 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
...
 	return ..()
 
+/**
+ * Will try to renew the objective to a valid state with another target.
+ * Returns the new found target or null if none were found
+ */
+/datum/objective/proc/on_target_loss()
+	if(owner?.current)
+		to_chat(owner.current, "<BR><span class='userdanger'>You get the feeling your target is no longer within reach. Time for Plan [pick("A","B","C","D","X","Y","Z")]. Objectives updated!</span>")
+		owner.current.playsound_local(get_turf(owner.current), 'sound/ambience/alarm4.ogg', 100, FALSE, pressure_affected = FALSE)
```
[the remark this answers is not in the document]
farie82: `playsound_local` does only play it for the one mob. `playsound` will play it for all in the area.
I believe I took this from how traitors are greeted atm which uses the same system.

code/game/gamemodes/objective.dm
```diff
@@ -380,9 +440,22 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
...
+			explanation_text += "It may also be in the possession of the [jointext(steal_target.protected_jobs, ", ")]."
+
 /datum/objective/steal/find_target()
 	var/loop=50
-	while(!steal_target && loop > 0)
+	var/found_target
+	while(!found_target && loop > 0)
 		loop--
 		var/thefttype = pick(GLOB.potential_theft_objectives)
 		var/datum/theft_objective/O = new thefttype
```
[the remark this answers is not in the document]
farie82: Will be quite the undertaking to refactor this seeing you need to check multiple things. I will leave it out, for now, to keep the scope as it is now. Will be something for a later refactor though

code/game/gamemodes/objective.dm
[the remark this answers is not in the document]
SabreML: Ah, I was looking at how cult conversion does it. I tested ingame too but in hindsight I called it on my mob so of course it would still play. I think one of them adds reverb and one doesn't maybe?

code/datums/mind.dm
```diff
@@ -18,6 +18,32 @@
...
+#define OBJECTIVE_DOWNLOAD			"download"
+#define OBJECTIVE_NUCLEAR			"nuclear"
+#define OBJECTIVE_CAPTURE			"capture"
+#define OBJECTIVE_ABSORB			"absorb"
+#define OBJECTIVE_DESTROY			"destroy"
+#define OBJECTIVE_MAROON			"maroon"
+#define OBJECTIVE_IDENTITY_THEFT	"identity theft"
+#define OBJECTIVE_ASSASSINATE_VIP	"assassinate VIP"
+#define OBJECTIVE_PROTECT_VIP		"protect VIP"
+#define OBJECTIVE_CUSTOM			"custom"
```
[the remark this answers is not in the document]
farie82: Can't actually capitalise most of them. Since they are used to determine the path of the objective.
Changing this is a bit too much of a refactor

code/game/gamemodes/objective.dm
```diff
@@ -319,13 +374,18 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 			var/mob/living/carbon/human/H = possible_target.current
 			if(!(NO_DNA in H.dna.species.species_traits))
 				possible_targets += possible_target
-	if(possible_targets.len > 0)
-		target = pick(possible_targets)
+	var/found_target
+	if(length(possible_targets) > 0)
+		found_target = pick(possible_targets)
+	set_target(found_target)
+	return found_target
```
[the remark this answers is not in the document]
farie82: I will override `is_invalid_target` instead of using this dupe code

code/game/gamemodes/objective.dm
```diff
@@ -392,13 +465,10 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 			continue
 		if(O.flags & 2)
```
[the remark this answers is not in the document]
farie82: Afraid this isn't possible because `THEFT_FLAG_UNIQUE` is defined in `steal_items.dm` which is alphabetically after `objective.dm`.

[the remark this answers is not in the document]
SabreML: [code\__DEFINES\antagonists.dm](https://github.com/ParadiseSS13/Paradise/blob/master/code/__DEFINES/antagonists.dm) would work for that I think, especially since it's completely empty at the moment.

code/game/gamemodes/objective.dm
farie82: `find_target` is always supposed to set the `target` value. Even if this becomes `null`. I'll change this in the other instances as well to reflect this better.

code/modules/admin/topic.dm
farie82: This will most likely break the existing targets. Or cause people to lose their objective. Or turn it into a Free objective. Haven't fully checked but I am fine with not supporting this due to these outcomes.

code/modules/admin/topic.dm
farie82: The glorious `if else` chain

code/modules/antagonists/_common/antag_datum.dm
farie82: The cut is needed for a proper GC. But I'll move it to the `Destroy` which is a better place for that

code/modules/mob/living/silicon/ai/latejoin.dm
farie82: It is not used there indeed. But it is used in `active_ais`. I'll update the comment

code/game/gamemodes/objective.dm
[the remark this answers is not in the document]
farie82: Indeed it adds reverb. Fox has changed this part of the code a few months ago to also use `SEND_SOUND` so I'll change it to that as well to keep it the same

code/game/gamemodes/objective.dm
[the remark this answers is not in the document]
farie82: How can that one be empty of all define files... Thanks I'll use that one

code/game/gamemodes/objective.dm
```diff
@@ -17,8 +17,25 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
...
+		owner.objectives -= src
+		owner = null
 	return ..()
 
+/**
+ * Will try to renew the objective to a valid state with another target.
+ * Returns the new found target or null if none were found
+ */
+/datum/objective/proc/on_target_loss()
+	if(owner?.current)
```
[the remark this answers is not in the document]
farie82: Good point. This might however require a rework depending on how much other code relies on the objective not to be null after this call. I'll keep it as is for now but you do raise a good point here.

dearmochi:
Sorry for the late response. **As there are at least two objections on this pull request, there is no choice but to unfortunately close it.**

Let's start with the good points:
- **More objective variety** - I feel like this is a good effort at giving antagonists more diverse things to do instead of the usual "kill X" or "steal Y", instead making them a sort of "guardian angel" for a random crew member. This provides a nice break from the monotony of antagonism.
- **The gamemode design** - I can see that a lot of thought was put in the balancing of this with the assassin/protector ratios and such.
- **The refactors and bug fixes** - These are always good to have and I would still like to have them in the codebase.

Now let's talk about the sticky points:
- **The VIP is a new player** - I can see this easily cause frustration amongst the protectors -- if your protection target keeps waltzing into maintenance shafts or just being generally reckless because they don't know better, you'll very quickly reach a point where you need to be more coercive and end up having to bucklecuff them to a chair someplace hidden so they don't get themselves killed.
- **It might be too much of a burden on the VIP** - Sometimes people just want to play the game without having a war being waged behind their backs. Being *one* antagonist's target can already be a lot, but being the target of *all* assassins even more so. All of this while the VIP is essentially helpless by themselves against antagonists with explosives, guns or guardians. Hell, an assassin with a stunprod and adrenals can very quickly kill the VIP without recourse.

For the points raised above, this PR was objected to by two different maintainers and unfortunately it has to be closed. I am sorry it was processed this late only for this outcome.

PR #14341, opened 2020-09-16
Gives engi/medical borgs an abductor tool upgrade from the exosuit fabricator

Qwertytoforty:
## What Does This PR Do

This PR gives engineering / medical borgs the ability to get an abductor equipment upgrade. This upgrade replaces the tools the borg gets with the equivalent abductor tool, and with the case of medical, replaces the laser scalpel with IMS, due to lack of an abductor laser scalpel. They can be produced in the exosuit fabricator for approximately the price of all the tools if printed for a human.

Also fixes borg upgrades like the diamond drill for mining borg leaving a ghost item slot when upgraded.

## Why It's Good For The Game

This is good for the game as it adds an upgrade to borgs other than security or miner that helps them specifically. It is not something that will happen every round, due to the requirement of abductor 4, which will require either very productive miners with the abductor ruin, traders, or actual abductors.  These tools are not combat oriented, and does not give the security borg the abductor baton either. This does not make borgs stronger then crew in these areas, as they are the exact same tools crew get, for the same price.

## Images of changes

Already in the code, but unused, is the new sprite for the module: 
image

## Changelog
:cl:
add: Allows medical / engineering borgs to get abductor tool upgrades from the exosuit fabricator with a new upgrade.
fix: Fixes modules like diamond drill upgrade for mining borgs having a ghost slot filled with the old drill.
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -183,7 +183,7 @@
 
 	return TRUE
 
-/obj/item/borg/upgrade/abductorengi
+/obj/item/borg/upgrade/abductor_engi
 	name = "engineering cyborg abductor upgrade"
 	desc = "An experimental upgrade that replaces an engineering cyborgs tools with the abductor version."
 	icon_state = "abductor_mod"
@@ -191,7 +191,7 @@
 	require_module = TRUE
 	module_type = /obj/item/robot_module/engineering
 
-/obj/item/borg/upgrade/abductorengi/action(mob/living/silicon/robot/R)
+/obj/item/borg/upgrade/abductor_engi/action(mob/living/silicon/robot/R)
 	if(..())
 		return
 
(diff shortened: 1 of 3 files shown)


SteelSlayer:
What happens if the borg loses this upgrade? Do their abductor tools remain?

Aquilar:
don't give medical module the IMS, use the tier 3 laser scalpel instead.

code/game/objects/items/robot/robot_upgrades.dm
```diff
@@ -183,6 +183,82 @@
 
 	return TRUE
 
+/obj/item/borg/upgrade/abductorengi
```
farie82, line 186: minor nitpick. Can you make this snake_case? Bit easier to read like that. Same for the other path

farie82 (changes requested):
Tested and works. Resetting the borg works as expected.
Minor nitpick styling wise

I also noticed something odd while testing but it's not due to the PR I believe. The newly added modules were not clickable on the inventory slot. But only on the item itself. After equiping and un-equiping the item it did become clickable. Quite annoying to have to accurately click on an item but yeah. Out of the scope of this PR though

farie82 (approved):

variableundefined (approved):

Kyep (approved):

PR #14348, opened 2020-09-17
Adds a number of new implants

Aquilar:
## What Does This PR Do

This PR adds the following implants:
Neural Jumpstarter: This implant attempts to wake you up back up if you've been slept or paralysed, has no affect on ether/ketamine and similar chems, can not be installed along side the CNS rebooter and has a cooldown on it's wake up function. Has hardened and compatible variants for admin shenanigans.

Janitorial arm implant: contains advanced mop, soap, light replacer, bottle of space cleaner (that only accepts space cleaner), flyswatter and holosign projector. For all your cleaning needs.

Botanical arm implant: contains plant analyser, cultivator, spade, hatchet, wire cutters and wrench for all your hydroponics needs.
note: I did consider adding the portable seed extractor and plant bag but I don't know how that would go over and yes they do work.

Hacking arm implant: an arm implant that contains screwdriver, wirecutters and alien multiool, only avaliable after getting abductor 4. Can not use it with the toolset implant. Same reason you can not have two toolset implants at once.

## Why It's Good For The Game

Adds more implant variety to the game, adding a counter part to the CNS rebooter that covers sleep and paralysis while not allowing them to be taken both at once ~~outside of admin events~~ and for those that say anything about the upgraded cyber heart, atleast this doesn't instantly kill you when your EMP'd 
The two toolset implants bring some implants to the service department namely janitor and hydroponics allowing the joy of chrome to be spread further. 
The hacking arm is a small award for researching abductor tech, so it would not appear very often.

## Changelog
:cl:
add: Neural jumpstarter, no more sleep for you.
add: Janitorial and botanical arm implants, exactly what you think.
add: Hacking arm implant.
/:cl:


code/modules/surgery/organs/augments_arms.dm
```diff
@@ -212,6 +212,17 @@
 		return TRUE
 	return FALSE
 
+/obj/item/organ/internal/cyberimp/arm/hacking
+	name = "hacking arm implant"
+	desc = "An small arm implant containing an advanced screwdriver, wirecutters, and multitool designed for engineers and on-the-field machine modification. Actually legal, despite what the name may make you think."
```
RedWolfie, line 217:
```suggestion
	desc = "A small arm implant containing an advanced screwdriver, wirecutters, and multitool designed for engineers and on-the-field machine modification. Actually legal, despite what the name may make you think."
```

TDSSS:
Having two different implants that start with CNS might get pretty confusing. Maybe consider an entirely different name to make them easier to tell apart? Especially since they both do very similar things.

Aquilar:
> Having two different implants that start with CNS might get pretty confusing. Maybe consider an entirely different name to make them easier to tell apart? Especially since they both do very similar things.

do you have any suggestions for names? since they both attach to the same brain slot preventing the use of either one if the other is present.

Kyep:
Pick a different prefix that isn't "CNS" for that implant.
Having two implants that both start with "CNS" but are incompatible with each other invites confusion.

Aquilar:
done.

Pennwick:
I'm honestly a little concerned about the janitor implant. 

The mop itself is a little questionable. Admittedly its a low level slip but it can no longer be removed from the janitor without removing his arm. For antaggery this is probably okay but I see it more used by janitors who want to water mop in congested areas and now people cant disarm and disposal your tools anymore. 

The spray cleaner bottle is an even bigger concern. Its a no-drop fully concealable spray bottle. I'm not up to date on my hellmixes but I'm pretty sure that would be wildly dangerous. Unless I'm misreading something and you can only fill it with water this has to be out of balance.

Aquilar:
It wouldn't take much work to make a safe spray cleaner (means editing a further two files for it). But I don't think hell mixes aren't really that much of a concern, since they're barely seen and there's better things to be doing than harvesting ants for 40 minutes.

procdrone:
Hacking implant seems refundant to me, given that we already have toolset implant which includes all necessary hacking equipment

TDSSS:
Alien multitool is much better at hacking since it shows what wires do.

procdrone:
then im even more against adding such implant. It just doesn't feel right or balanced to me.

Aquilar:
you get it at the same time you get the alien multitool anyway which is Abductor 4, along with all the other abductor tools.

Pennwick:
I don't really have an issue with the hacking one. If you could print that then you could almost certainly get the multitool on its own. And since it doesn't include a crowbar you can still properly brig someone with it. If you already know the wires its straight up lesser compared to the normal toolset. Though its definitely more of an asset to the hacker going in blind and the ultimate in bomb defusal.

code/modules/research/designs/medical_designs.dm
```diff
@@ -323,6 +323,39 @@
 	build_path = /obj/item/organ/internal/cyberimp/arm/toolset
 	category = list("Medical")
 
+/datum/design/cyberimp_janitorial
+	name = "Janitorial Toolset Implant"
+	desc = "A set of janitorial tools hidden behind a concealed panel on the user's arm"
```
SteelSlayer, line 328:
```suggestion
	desc = "A set of janitorial tools hidden behind a concealed panel on the user's arm."
```

code/modules/research/designs/medical_designs.dm
```diff
@@ -323,6 +323,39 @@
...
+	req_tech = list("materials" = 3, "engineering" = 4, "biotech" = 4)
+	build_type = PROTOLATHE | MECHFAB
+	materials = list(MAT_METAL = 2500, MAT_GLASS = 1500, MAT_SILVER = 1500)
+	construction_time = 200
+	build_path = /obj/item/organ/internal/cyberimp/arm/janitorial
+	category = list("Medical")
+
+/datum/design/cyberimp_botanical
+	name = "Botanical Toolset Implant"
+	desc = "A set of botanical tools hidden behind a concealed panel on the user's arm"
```
SteelSlayer, line 339:
```suggestion
	desc = "A set of botanical tools hidden behind a concealed panel on the user's arm."
```

code/modules/surgery/organs/augments_internal.dm
```diff
@@ -161,6 +161,54 @@
...
+	if(crit_fail)
+		return
+	if(owner.stat == UNCONSCIOUS && cooldown == FALSE)
+		owner.AdjustSleeping(-100, FALSE)
+		owner.AdjustParalysis(-100, FALSE)
+		to_chat(owner, "<span class='notice'>You feel a rush of energy course through your body!</span>")
+		cooldown = TRUE
+		addtimer(CALLBACK(src, .proc/sleepytimerend), 50)
+
+/obj/item/organ/internal/cyberimp/brain/anti_sleep/proc/sleepytimerend()
```
SteelSlayer, line 183: Just a suggestion to make this more readable.
```suggestion
/obj/item/organ/internal/cyberimp/brain/anti_sleep/proc/sleepy_timer_end()
```

SteelSlayer (commented):
Code looks good aside from some small grammatical stuff. 

Also just a suggestion about the icons for the action buttons of the janitorial and botanical implant. Instead of duffel bags, I think something like a mop, and a spade might look better. You know, something more tool related. Doesn't have to be exactly those but, just something other than duffel bags.

Denthamos:
My problem with the janitorial one is it makes the existing mop implant null along with the spray bottle part which really should not be in an arm due to chemical concerns unless it can ONLY be refilled with cleaner.

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -108,6 +107,21 @@
...
+/obj/item/reagent_containers/spray/cleaner/safety
+	desc = "BLAM!-brand non-foaming space cleaner! This spray bottle can only accept space cleaner."
+
+/obj/item/reagent_containers/spray/cleaner/safety/on_reagent_change()
+	var/found_filthy_reagent = FALSE
+	for(var/datum/reagent/R in reagents.reagent_list)
+		if(R.id != "cleaner") //all chems other than space cleaner are filthy.
+			reagents.del_reagent(R.id)
+			found_filthy_reagent = TRUE
+		if(found_filthy_reagent)
```
SteelSlayer, line 119: The use of the `found_filthy_reagent` var is not really needed here. For example, if the `if(R.id != "cleaner")` check passes, you know it found a filthy reagent, there's no need to set the var to true and then check if its true.

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -108,6 +107,18 @@
 	desc = "BLAM!-brand non-foaming space cleaner!"
 	list_reagents = list("cleaner" = 250)
 
+/obj/item/reagent_containers/spray/cleaner/safety
+	desc = "BLAM!-brand non-foaming space cleaner! This spray bottle can only accept space cleaner."
+
+/obj/item/reagent_containers/spray/cleaner/safety/on_reagent_change()
+	for(var/datum/reagent/R in reagents.reagent_list)
```
farie82, line 114: Can be made typeless

farie82 (changes requested):
One minor thing

code/modules/reagents/reagent_containers/spray.dm
Aquilar: I've forgotten how to do typeless, it's been awhile.

farie82:
```dm
for(var/thing in list)
  var/datum/path/to/obj/O = thing
  O.do_stuff()
```
Will make sure that the list is not filtered. Since the reagent_list should always have reagents in it this should be fine

Kyep (approved):

farie82 (approved):

PR #14358, opened 2020-09-17
Fix atmospherics control UI not refreshing data correctly

dearmochi:
## What Does This PR Do

Fixes air alarms not updating on Atmospherics Consoles. They'd always show up as "Good" because of data caching that is never invalidated. That PR fixes that.

## Changelog
:cl:
fix: Fix Atmospheric Consoles not using up-to-date data
/:cl:


AffectedArc07 (approved):
Seems fine

PR #14360, opened 2020-09-18
Fixes station bounced radios / intercomms

Kyep:
## What Does This PR Do
Fixes radios (and subtypes) completely ignoring the internal_channels list.
The way radios are meant to work, they have an internal_channels list of channels that have access requirements. IF you meet the requirements (per channel) then you can change the radio's primary/default channel to that channel.
This is how station bounced radios work. By default, you pick them up, they are set to common, but if you have say engi access you can set them to engi radio as their default (";" hotkey) radio.
It is also how intercomms work.

Previous TGUI port of radios removed this. This PR re-adds it.
It also refactors some 1/0 to TRUE/FALSE, and removes some snowflake (locked intercomms having their own locked var separate from normal freqlock).

## Changelog
:cl: Kyep
fix: fixed being unable to re-tune intercomms, station bounced radios, etc to department channels.
/:cl:


code/game/objects/items/devices/radio/radio.dm
```diff
@@ -129,7 +129,10 @@ GLOBAL_LIST_INIT(default_medbay_channels, list(
 /obj/item/radio/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = TRUE, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
 	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
 	if(!ui)
-		ui = new(user, src, ui_key, "Radio", name, 360, 150 + (length(channels) * 20), master_ui, state)
+		var/list/schannels = list_secure_channels(user)
+		var/list/ichannels = list_internal_channels(user)
+		var/calc_height = 150 + (schannels.len * 20) + (ichannels.len * 10)
+		ui = new(user, src, ui_key, "Radio", name, 400, calc_height, master_ui, state)
```
AffectedArc07, line 135: I would say clamp this to a sane height, but the max this can reasonably be is about 400 pixels.

tgui/packages/tgui/interfaces/Radio.js
```diff
@@ -27,13 +27,18 @@ export const Radio = (props, context) => {
...
+  const schannels = map((value, key) => ({
     name: key,
     status: !!value,
-  }))(data.channels);
+  }))(data.schannels);
+  const ichannels = map((value, key) => ({
+    name: key,
+    freq: value,
+  }))(data.ichannels);
+  const clength = "YYY: " + schannels.length;
```
AffectedArc07, line 38: Variable defined but not used

AffectedArc07 (changes requested):

AffectedArc07 (approved):
Should be fine

PR #14367, opened 2020-09-18
Replaces firealarm interface

Kyep:
## What Does This PR Do
Removes the interface you get when you click a fire alarm.
Instead, clicking the alarm (if you're next to it) toggles it on/off.
If you want to see the alert level, you can do that by just looking at the fire alarm (its icon shows the alert level). Or you can examine the fire alarm.
This removes the feature where you could set a timer on a fire alarm so it would go off after awhile - as nobody ever used that feature.
This also deletes a few nano templates I somehow forgot to delete in other PRs.

## Changelog
:cl: Kyep
tweak: simplified fire alarms - you now click them to turn them on/off, or examine them to see the alert level.
/:cl:


TDSSS:
I think you accidentally deleted some more templates there.

AffectedArc07:
@TDSSS 
image

Dumbdumn5 (approved):
If fire alarms have no other use, I wouldn't mind this change too much. Tentative :+1:

PR #14374, opened 2020-09-18
Highfive changes and fixes

### LINKED ISSUE #13207: High Five Fail
**Issue Description**:
High five timer is both too short(not a bug, technically, I think?) and, even if a high five is successful, will give you a message that you were left hanging right after the high five.

I didn't expect to magically be left hanging after high fiving.

**What happened instead**:
I was somehow left hanging after high fiving.

You aren't left hanging if you high fived man.

1. High five
2. Somehow feel as if there was no passion in the high five, and be left hanging.

Forever, I guess.

**Extra information**:
N/A

Cocacolagua:
## What Does This PR Do
Fixes #13207
Removes the explosion and godmode from the wizard highfive, now it plays a cool fireball sound instead
You now have ten seconds to react to a *highfive instead of four seconds

## Why It's Good For The Game
Bugs bad
Spammable godmode on demand bad
Human reaction time for high-fives good

## Changelog
:cl:
tweak: Wizards' high-five no longer makes you immortal and explodes, but instead plays a cool fireball sound.
tweak: High-five prompts last for ten seconds instead of four.
fix: High-fives no longer fail after succeeding.
/:cl:

### DIFF
--- a/code/datums/status_effects/neutral.dm
+++ b/code/datums/status_effects/neutral.dm
@@ -40,8 +40,8 @@
 
 /datum/status_effect/high_five
 	id = "high_five"
-	duration = 40
+	duration = 200
 	alert_type = null
 
 /datum/status_effect/high_five/on_remove()
-	owner.visible_message("[owner] was left hanging....")
+	return
--- a/code/datums/status_effects/neutral.dm
+++ b/code/datums/status_effects/neutral.dm
@@ -40,7 +40,7 @@
 
 /datum/status_effect/high_five
 	id = "high_five"
-	duration = 200
+	duration = 100
 	alert_type = null
 
 /datum/status_effect/high_five/on_remove()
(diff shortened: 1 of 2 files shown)


Aurorablade:
the wizard one is working as intended. Plus its wizard...thats the point. to cause choas. If anything just make the react chance shorter for them. you are rarely going to see two wizards in one round to proc this. and if you do see two wizards in one round..its ragin' mages, admin abuse or an apprentice. in the case of the first two...You are already doomed.

Cocacolagua:
it didn't actually "work as intended" before high-fives broke either, wizards would get gibbed for performing the highfive.

even if not, i don't believe we should keep infinite explosions and godmode because "you need two wizards, it's never going to happen", since it's still bullshit for everyone involved, even more so in ragin' mages rounds, where the crew already has stuff to deal with.

Aurorablade:
if they got gibbed before hand thats the bug. The godmode part was what was preventing the gibbing of the wizard. so the timing is off or out of sequence.

TDSSS:
Don't make wizard high-fives lame, please.

variableundefined (commented):
Objection

Cocacolagua:
After some thought, I decided that these small fun interactions are part of what makes the game special, so I'll revert and fix the wizard mindswap to be more unique when I can

PR #14384, opened 2020-09-19
Creates login plugin for TGUI to streamline and unify login frontend and backend

dearmochi:
## What Does This PR Do

- Creates the folder `modules/tgui/plugins` which houses self-contained libraries that can be used to complement TGUI interactions. This results in less duplicate code and less code to maintain.
- Adds the `login` plugin which gives the option to handle logins using ID in the same manner as the Medical Machine.
- Adds an Eject ID button to the login plugin to allow ejecting the ID without logging off a machine.

## Images of changes

image


code/modules/tgui/plugins/login.dm
```diff
@@ -0,0 +1,203 @@
...
+  * Attempts to log out.
+  *
+  * Arguments:
+  * * state - The current login state
+  */
+/obj/proc/tgui_login_logout(datum/tgui_login/state = tgui_login_get())
+	if(!state.logged_in)
+		return
+
+	state.name = state.rank = ""
```
AffectedArc07, line 154:
```suggestion
	state.name = ""
	state.rank = ""
```

code/game/machinery/computer/medical.dm
```diff
@@ -92,14 +85,10 @@
...
 	data["temp"] = temp
-	data["scan"] = scan ? scan.name : null
-	data["authenticated"] = authenticated
-	data["rank"] = rank
 	data["screen"] = screen
 	data["printing"] = printing
-	data["isAI"] = isAI(user)
-	data["isRobot"] = isrobot(user)
-	if(authenticated)
+	tgui_login_data(data, user)
```
AffectedArc07, line 90: Add a note here to say this will append to data, otherwise it just looks like its using data, not sending anything back.

AffectedArc07 (changes requested):
Few things. I do like the idea of a TGUI plugins system though.

code/game/machinery/computer/medical.dm
dearmochi: Sure

code/modules/tgui/plugins/login.dm
dearmochi: Okay

AffectedArc07 (approved):

PR #14385, opened 2020-09-19
[TGUI] Portable Scrubbers

SteelSlayer:
## What Does This PR Do
Portable scrubbers now use TGUI. 

Also cleans up from some stuff I forgot while doing a couple previous TGUIs:
- `#undef`s some defines in `code/ATMOSPHERICS/components/trinary_devices/filter.dm`
- `#undef`s some defines in `code/game/machinery/atmoalter/pump.dm`
- Removes a `resizable` from `<Window>` in `PortablePump.js`. This wasn't suppose to be here.

## Images of changes
Very similar to the portable pump UI, just without the "pump direction" setting.

scrubber

## Changelog
:cl:
tweak: Portable scrubbers now use TGUI.
/:cl:


AffectedArc07 (approved):
God please these are so clunky to use

PR #14391, opened 2020-09-20
Adds a confirmation to the deny shuttle call verb

farie82:
## What Does This PR Do
Does what it says. Defaults to Yes

## Why It's Good For The Game
Less accidental denying of the shuttle

## Changelog
:cl:
tweak: The Toggle shuttle call verb now has a confirmation
/:cl:

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -874,7 +874,7 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	if(!check_rights(R_ADMIN))
 		return
 
-	if(alert(usr, "Do you want to [SSshuttle.emergencyNoEscape ? "ALLOW" : "DENY"] shuttle calls?", "Toggle Deny Shuttle", "No", "Yes") != "Yes")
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -874,6 +874,9 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	if(!check_rights(R_ADMIN))
 		return
 
+	if(alert(usr, "Do you want to [SSshuttle.emergencyNoEscape ? "ALLOW" : "DENY"] shuttle calls?", "Toggle Deny Shuttle", "No", "Yes") != "Yes")
```
AffectedArc07, line 877:
```suggestion
	if(alert(usr, "Do you want to [SSshuttle.emergencyNoEscape ? "ALLOW" : "DENY"] shuttle calls?", "Toggle Deny Shuttle", "Yes", "No") != "Yes")
```
Yes is always on the left for popups and this will be a pain in the ass otherwise
image

AffectedArc07 (changes requested):

code/modules/admin/verbs/randomverbs.dm
farie82: Oh I always do it the other way around to make sure admins actually know what they are saying yes to but I'll change it

AffectedArc07 (approved):

PR #14394, opened 2020-09-20
Makes slimes immune to tesla

ppi13:
## What Does This PR Do

This makes the tesla zap ignore slimes, so that when the tesla eventually makes it way to xenobiology and it's army of slimes, there won't be infinity lag as each slime is shocked by each other slime that is shocked, it's not the 'cleanest' of solutions, but it works.

## Why It's Good For The Game

Everytime tesla gets free, xenobiology had good xenobiologists, and the tesla gets to xenobio, it would create infinity zap logs, this prevents this lag machine from functioning in the first place.
I tested it by spawning 2000 slimes, and the only lag was when they died from space.

## Changelog
:cl: ppi
add: Add simple_mob slimes to the tesla ignore list to avoid lag
/:cl:

### DIFF
--- a/code/modules/power/tesla/energy_ball.dm
+++ b/code/modules/power/tesla/energy_ball.dm
@@ -185,7 +185,8 @@
 										/obj/structure/sign,
 										/obj/machinery/gateway,
 										/obj/structure/grille,
-										/obj/machinery/the_singularitygen/tesla))
+										/obj/machinery/the_singularitygen/tesla,
+										/mob/living/simple_animal/slime))
 
 
 	for(var/A in typecache_filter_multi_list_exclusion(oview(source, zap_range+2), things_to_shock, blacklisted_tesla_types))


AffectedArc07:
Just pointing this out. This only lags the client itself, not the server.

TDSSS:
And here I thought this was a buff to slimepeople

Have we considered just deleting dead slimes once they get tesla'd once or something instead of them being immune?

variableundefined (approved):

ppi13:
I did consider just straight up dusting them, but this is a kinda cleaner solution that is less likely to break interactions.

taukausanake:
I would think of making them pop and throw their cores all over. Maybe make some slimy gibs everywhere too

farie82 (approved):
Code works

Kyep (approved):

PR #14396, opened 2020-09-20
Light fixture code cleanup & spooking

SabreML:
## What Does This PR Do

- Cleans up the light fixture/light bulb code, and adds some documentation too.
This isn't meant to be a refactor, I was originally going to just fix a typo but I got a bit carried away.

- This also makes it so that every time a light flickers, it has a very small chance of burning out the bulb. With each flicker adding a 0.1% chance of it happening.
This may seem incredibly small *(because it is)* but each ghost spook has 5 to 10 flickers (0.5% to 1%), so it adds up over time.
And ghosts shouldn't be able to break lights consistently, since that ends up being a shadowling buff.

## Why It's Good For The Game

Documentation is good, as is *(somewhat)* more readable code!

Makes ghosts even more spooky. 👻 

## Changelog
:cl:
add: Ghosts spooking a light adds a very small chance of the light burning out (0.5% - 1%)
tweak: Cleaned up/refactored the light fixture code
/:cl:

### DIFF
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -796,13 +796,12 @@
 		if(!length(S.reagents.reagent_list))
 			return
 
-		else
-			to_chat(user, "<span class='notice'>You inject the solution into [src].</span>")
-			if(S.reagents.has_reagent("plasma", 5) || S.reagents.has_reagent("plasma_dust", 5))
-				log_admin("LOG: [key_name(user)] injected a light with plasma, rigging it to explode.")
-				message_admins("LOG: [key_name_admin(user)] injected a light with plasma, rigging it to explode.")
+		to_chat(user, "<span class='notice'>You inject the solution into [src].</span>")
+		if(S.reagents.has_reagent("plasma", 5) || S.reagents.has_reagent("plasma_dust", 5))
+			log_admin("LOG: [key_name(user)] injected a light with plasma, rigging it to explode.")
+			message_admins("LOG: [key_name_admin(user)] injected a light with plasma, rigging it to explode.")
 
-				rigged = TRUE
+			rigged = TRUE
 
 			S.reagents.clear_reagents()
 	else
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -1,7 +1,6 @@
-// The lighting system
+// The Lighting System
 //
-// consists of light fixtures (/obj/machinery/light) and light tube/bulb items (/obj/item/light)
-
+// Consists of light fixtures (/obj/machinery/light) and light tube/bulb items (/obj/item/light)
 
 // status values shared between lighting fixtures and items
 #define LIGHT_OK 0
@@ -9,18 +8,29 @@
 #define LIGHT_BROKEN 2
 #define LIGHT_BURNED 3
 
+/**
+  * # Light fixture frame
+  *
+  * Incomplete light tube fixture
+  *
+  * Becomes a [Light fixture] when completed
+  */
 /obj/machinery/light_construct
 	name = "light fixture frame"
 	desc = "A light fixture under construction."
 	icon = 'icons/obj/lighting.dmi'
 	icon_state = "tube-construct-stage1"
-	anchored = 1
+	anchored = TRUE
 	layer = 5
 	max_integrity = 200
 	armor = list("melee" = 50, "bullet" = 10, "laser" = 10, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 80, "acid" = 50)
+	/// Construction stage (1 = Empty frame | 2 = Wired frame | 3 = Completed frame)
 	var/stage = 1
+	/// Light bulb type
 	var/fixture_type = "tube"
+	/// How many metal sheets get given after deconstruction
 	var/sheets_refunded = 2
+	/// Holder for the completed fixture
 	var/obj/machinery/light/newlight = null
 
 /obj/machinery/light_construct/New()
@@ -43,12 +53,11 @@
 	. = TRUE
 	switch(stage)
 		if(1)
-			to_chat(user, "<span class='notice'>You begin deconstructing [src].</span>")
+			to_chat(user, "<span class='notice'>You begin to dismantle [src].</span>")
 			if(!I.use_tool(src, user, 30, volume = I.tool_volume))
 				return
 			new /obj/item/stack/sheet/metal(get_turf(loc), sheets_refunded)
-			user.visible_message("<span class='notice'>[user] deconstructs [src].</span>", \
-				"<span class='notice'>You deconstruct [src].</span>", "You hear a noise.")
+			TOOL_DISMANTLE_SUCCESS_MESSAGE
 			qdel(src)
 		if(2)
 			to_chat(user, "<span class='warning'>You have to remove the wires first.</span>")
@@ -58,10 +67,8 @@
 /obj/machinery/light_construct/wirecutter_act(mob/living/user, obj/item/I)
 	if(stage != 2)
 		return
-	. = TRUE
-	if(!I.use_tool(src, user, 0))
+	if(!I.use_tool(src, user, 0, volume = I.tool_volume))
 		return
-	playsound(loc, I.usesound, 100, 1)
 	. = TRUE
 	stage = 1
 	switch(fixture_type)
@@ -70,14 +77,13 @@
 		if("bulb")
 			icon_state = "bulb-construct-stage1"
 	new /obj/item/stack/cable_coil(get_turf(loc), 1, paramcolor = COLOR_RED)
-	user.visible_message("<span class='notice'>[user] removes the wiring from [src].</span>", \
-		"<span class='notice'>You remove the wiring from [src].</span>", "You hear a noise.")
+	WIRECUTTER_SNIP_MESSAGE
 
 /obj/machinery/light_construct/screwdriver_act(mob/living/user, obj/item/I)
 	if(stage != 2)
 		return
 	. = TRUE
-	if(!I.use_tool(src, user, 0))
+	if(!I.use_tool(src, user, 0, volume = I.tool_volume))
 		return
 	switch(fixture_type)
 		if("tube")
@@ -86,8 +92,7 @@
 			icon_state = "bulb-empty"
 	stage = 3
 	user.visible_message("<span class='notice'>[user] closes [src]'s casing.</span>", \
-		"<span class='notice'>You close [src]'s casing.</span>", "You hear a noise.")
-	playsound(loc, I.usesound, 75, 1)
+		"<span class='notice'>You close [src]'s casing.</span>", "You hear a screwdriver.")
 
 	switch(fixture_type)
 		if("tube")
@@ -115,7 +120,7 @@
 		stage = 2
 		playsound(loc, coil.usesound, 50, 1)
 		user.visible_message("<span class='notice'>[user.name] adds wires to [src].</span>", \
-			"<span class='notice'>You add wires to [src].</span>")
+			"<span class='notice'>You add wires to [src].</span>", "You hear a noise.")
 		return
 
 	return ..()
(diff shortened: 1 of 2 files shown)


code/modules/power/lighting.dm
```diff
@@ -86,8 +92,7 @@
 			icon_state = "bulb-empty"
 	stage = 3
 	user.visible_message("<span class='notice'>[user] closes [src]'s casing.</span>", \
-		"<span class='notice'>You close [src]'s casing.</span>", "You hear a noise.")
-	playsound(loc, I.usesound, 75, 1)
+		"<span class='notice'>You close [src]'s casing.</span>", "You hear a screwdriver.")
```
SteelSlayer, line 95:
```suggestion
		"<span class='notice'>You close [src]'s casing.</span>", "<span class='notice'>You hear a screwdriver.</span>")
```

code/modules/power/lighting.dm
```diff
@@ -115,7 +120,7 @@
 		stage = 2
 		playsound(loc, coil.usesound, 50, 1)
 		user.visible_message("<span class='notice'>[user.name] adds wires to [src].</span>", \
-			"<span class='notice'>You add wires to [src].</span>")
+			"<span class='notice'>You add wires to [src].</span>", "You hear a noise.")
```
SteelSlayer, line 123:
```suggestion
			"<span class='notice'>You add wires to [src].</span>", "<span class='notice'>You hear a noise.</span>")
```

code/modules/power/lighting.dm
```diff
@@ -129,58 +134,89 @@
 		new /obj/item/stack/sheet/metal(loc, sheets_refunded)
 	qdel(src)
 
+/**
+  * # Small light fixture frame
+  *
+  * Incomplete light bulb fixture
+  *
+  * Become a [Small light fixture] when completed
```
SteelSlayer, line 142:
```suggestion
  * Becomes a [Small light fixture] when completed
```

code/modules/power/lighting.dm
```diff
@@ -129,58 +134,89 @@
...
+	layer = 5
 	max_integrity = 100
 	use_power = ACTIVE_POWER_USE
 	idle_power_usage = 2
 	active_power_usage = 20
 	power_channel = LIGHT //Lights are calc'd via area so they dont need to be in the machine list
-	var/on = FALSE					// 1 if on, 0 if off
+	/// Is the light on or off?
+	var/on = FALSE
 	var/on_gs = 0
```
SteelSlayer, line 176: Could you document this var? If you know what it does that is.

code/modules/power/lighting.dm
```diff
@@ -200,34 +236,32 @@
 
 /obj/machinery/light/built/New()
 	status = LIGHT_EMPTY
-	update(0)
```
SteelSlayer, line 203: Why did you remove the 0 here and from the others below?

code/modules/power/lighting.dm
```diff
@@ -378,32 +412,30 @@
...
+			"<span class ='danger'>You hear the tinkle of breaking glass.</span>")
 			if(on && (W.flags & CONDUCT))
 				if(prob(12))
 					electrocute_mob(user, get_area(src), src, 0.3, TRUE)
 			break_light_tube()
 
 		else
-			user.visible_message("<span class='danger'>[user.name] hits the light.</span>")
+			user.visible_message("<span class='danger'>[user] hits the light.</span>", "<span class ='danger'>You hit the light.</span>", \
+			"<span class ='danger'>You hear someone hitting a light.")
```
SteelSlayer, line 426:
```suggestion
			"<span class ='danger'>You hear someone hitting a light.</span>")
```

code/modules/power/lighting.dm
```diff
@@ -378,32 +412,30 @@
...
 			playsound(src.loc, 'sound/effects/glasshit.ogg', 75, 1)
 
 	// attempt to stick weapon into light socket
 	else if(status == LIGHT_EMPTY)
 		if(istype(W, /obj/item/screwdriver)) //If it's a screwdriver open it.
-			playsound(src.loc, W.usesound, 75, 1)
-			user.visible_message("[user.name] opens [src]'s casing.", \
-				"You open [src]'s casing.", "You hear a noise.")
+			playsound(src.loc, W.usesound, W.tool_volume, 1)
```
SteelSlayer, line 432: Implied src
```suggestion
			playsound(loc, W.usesound, W.tool_volume, 1)
```

code/modules/power/lighting.dm
```diff
@@ -469,19 +501,18 @@
...
-				on = !on
-				update(0)
-				sleep(rand(5, 15))
-			on = (status == LIGHT_OK)
-			update(0)
-		flickering = 0
+	flickering = TRUE
+	if(on && status == LIGHT_OK)
+		for(var/i = 0; i < amount; i++)
+			if(status != LIGHT_OK) break
```
SteelSlayer, line 509:
```suggestion
			if(status != LIGHT_OK) 
				break
```

code/modules/power/lighting.dm
```diff
@@ -469,19 +501,18 @@
...
-				if(status != LIGHT_OK) break
-				on = !on
-				update(0)
-				sleep(rand(5, 15))
-			on = (status == LIGHT_OK)
-			update(0)
-		flickering = 0
+	flickering = TRUE
+	if(on && status == LIGHT_OK)
+		for(var/i = 0; i < amount; i++)
```
SteelSlayer, line 508: Faster loop format:
```suggestion
		for(var/i in 1 to amount)
```

code/modules/power/lighting.dm
```diff
@@ -722,14 +772,12 @@
 	if(istype(I, /obj/item/reagent_containers/syringe))
 		var/obj/item/reagent_containers/syringe/S = I
 
-		to_chat(user, "You inject the solution into the [src].")
-
 		if(S.reagents.has_reagent("plasma", 5) || S.reagents.has_reagent("plasma_dust", 5))
-
+			to_chat(user, "You inject the solution into [src].")
```
SteelSlayer, line 776:
```suggestion
			to_chat(user, "<span class='notice'>You inject the solution into [src].</span>")
```

code/modules/power/lighting.dm
SabreML: No idea what that does, it's only used here:
```
if(on != on_gs)
    on_gs = on
```
I couldn't figure out why it needs to be there, so I just left it until I had time to test.

code/modules/power/lighting.dm
SabreML: The `0` is implied if there's no arguments, so I thought it'd be cleaner to just remove them.

SteelSlayer: well in the proc: `/obj/machinery/light/proc/update(trigger = TRUE)`, the default arg is true which means now all of the old `update(0)`'s in this file are `update(TRUE)`.

SabreML: Just tested it again and yeah, you're right.

The last time I tested it it said that `trigger` was set to `null`, but I guess I was looking at the wrong thing?

code/modules/power/lighting.dm
SteelSlayer: Honestly I can't even follow what its suppose to do either. I would remove it and then test lights, see what happens. Or just leave it alone.  :smile:

code/modules/power/lighting.dm
```diff
@@ -200,34 +236,32 @@
...
 /obj/machinery/light/built/New()
 	status = LIGHT_EMPTY
-	update(0)
+	update(FALSE)
 	..()
 
 /obj/machinery/light/small/built/New()
 	status = LIGHT_EMPTY
-	update(0)
+	update(FALSE)
```
farie82, line 244: Can be removed since it's being done on line 264 as well

code/modules/power/lighting.dm
```diff
@@ -200,34 +236,32 @@
...
 	..()
 
 // create a new lighting fixture
 /obj/machinery/light/New()
 	..()
-	spawn(2)
-		var/area/A = get_area(src)
-		if(A && !A.requires_power)
-			on = 1
+	var/area/A = get_area(src)
```
farie82, line 250: Can you try moving this to the `Initialize` instead of the `New`?

code/modules/power/lighting.dm
farie82: It is here to check if the state changed between `update` calls. If update is called when the light is on and before was on as well then it won't do the power stuff below.

code/modules/power/lighting.dm
```diff
@@ -343,13 +394,13 @@
...
-			to_chat(user, "There is a [fitting] already inserted.")
+			to_chat(user, "<span class='notice'>There is a [fitting] already inserted.</span>")
 		else
-			src.add_fingerprint(user)
+			add_fingerprint(user)
 			var/obj/item/light/L = W
 			if(istype(L, light_type))
 				status = L.status
-				to_chat(user, "You insert the [L.name].")
+				to_chat(user, "You insert [L].")
```
farie82, line 403: Needs spans

code/modules/power/lighting.dm
```diff
@@ -378,32 +429,30 @@
...
-			user.visible_message("[user.name] opens [src]'s casing.", \
-				"You open [src]'s casing.", "You hear a noise.")
+			playsound(loc, W.usesound, W.tool_volume, 1)
+			user.visible_message("<span class='notice'>[user] opens [src]'s casing.</span>", \
+				"<span class='notice'>You open [src]'s casing.</span>", "<span class='notice'>You hear a screwdriver.</span>")
 			deconstruct()
 			return
 
-		to_chat(user, "You stick \the [W] into the light socket!")
+		to_chat(user, "<span class='danger'>You stick [W] into the light socket!</span>")
```
farie82, line 455: `userdanger` might fit better here

code/modules/power/lighting.dm
```diff
@@ -461,31 +510,33 @@
...
+	if(flickering)
+		return
+	if(on)
+		flickering = TRUE
+		for(var/i in 1 to amount)
+			if(status != LIGHT_OK)
+				break
+			on = !on
+			update() // No param means that ghosts can burn out lights
+			sleep(rand(5, 15))
```
farie82, line 531: Please ensure this won't hang the whole code path above.
Revenant spells will also call this for instance

code/modules/power/lighting.dm
```diff
@@ -611,30 +662,39 @@
...
-	var/turf/T = get_turf(src.loc)
-	spawn(0)
-		break_light_tube()	// break it first to give a warning
-		sleep(2)
-		explosion(T, 0, 0, 2, 2)
-		sleep(1)
-		qdel(src)
+	var/turf/T = get_turf(loc)
+	break_light_tube()	// break it first to give a warning
+	sleep(2)
```
farie82, line 667: Same here

code/modules/power/lighting.dm
```diff
@@ -722,14 +792,12 @@
 	if(istype(I, /obj/item/reagent_containers/syringe))
 		var/obj/item/reagent_containers/syringe/S = I
 
-		to_chat(user, "You inject the solution into the [src].")
```
farie82, line 725: now you won't get feedback if you inject anything other than plasma or plasma dust

code/modules/power/lighting.dm
```diff
@@ -745,11 +813,11 @@
 
 /obj/item/light/proc/shatter()
 	if(status == LIGHT_OK || status == LIGHT_BURNED)
-		src.visible_message("<span class='warning'>[name] shatters.</span>","<span class='warning'>You hear a small glass object shatter.</span>")
+		visible_message("<span class='warning'>[src] shatters.</span>","<span class='warning'>You hear a small glass object shatter.</span>")
```
farie82, line 816: spacing

farie82 (changes requested):

variableundefined (dismissed):

variableundefined (changes requested):
See farie's changes. Otherwise approved.

code/modules/power/lighting.dm
SabreML: Tested both of them with `sleep(200)`, and it only freezes the proc it's called in. You can still call others while it's sleeping, or even call the same proc again.

code/modules/power/lighting.dm
```diff
@@ -722,16 +793,18 @@
...
 
-		to_chat(user, "You inject the solution into the [src].")
-
-		if(S.reagents.has_reagent("plasma", 5) || S.reagents.has_reagent("plasma_dust", 5))
+		if(!length(S.reagents.reagent_list))
+			return
 
-			log_admin("LOG: [key_name(user)] injected a light with plasma, rigging it to explode.")
-			message_admins("LOG: [key_name_admin(user)] injected a light with plasma, rigging it to explode.")
+		else
```
farie82, line 799: else is not required due to line 797

code/modules/power/lighting.dm
```diff
@@ -745,11 +818,11 @@
 
 /obj/item/light/proc/shatter()
 	if(status == LIGHT_OK || status == LIGHT_BURNED)
-		src.visible_message("<span class='warning'>[name] shatters.</span>","<span class='warning'>You hear a small glass object shatter.</span>")
+		visible_message("<span class='warning'>[src] shatters.</span>", "<span class='warning'>You hear a small glass object shatter.</span>")
```
farie82, line 821: it will now show the "You hear a glass object shatter" message to the light. I think you meant to make that message be the `blind_message`. You can do this by adding a null as 2nd param

code/modules/power/lighting.dm
SabreML: That's a leftover from something else I tried, whoops!

code/modules/power/lighting.dm
SabreML: The `self_message` parameter is only on `/mob/visible_message`, whereas this is just `/atom/proc/visible_message`.

farie82: Yep you're right! Overlooked that version of the proc

code/modules/power/lighting.dm
```diff
@@ -722,16 +793,17 @@
 	if(istype(I, /obj/item/reagent_containers/syringe))
 		var/obj/item/reagent_containers/syringe/S = I
 
-		to_chat(user, "You inject the solution into the [src].")
+		if(!length(S.reagents.reagent_list))
+			return
 
+		to_chat(user, "<span class='notice'>You inject the solution into [src].</span>")
```
farie82, line 799: This one combined with 806 is still a bit odd. Here it says you did it. Then if you don't have plasma in the syringe it won't clear it.
I'd say put this one inside the if scope and maybe give a warning message saying that you can't rig it or such if you don't have plasma in the syringe

SabreML: Damn, could've sworn I fixed that.

farie82 (approved):

variableundefined (approved):

PR #14398, opened 2020-09-20
Removes some 'null' variables from cyberiad.dmm

SabreML:
## What Does This PR Do

Removes a bunch of `x = null;`'s from the `cyberiad.dmm` file.

## Why It's Good For The Game

They're only there because of a Dream Maker "quirk", so removing them doesn't actually have any downsides.
And it makes the 'Edit Instance' thing look cleaner.

## Changelog
:cl:
tweak: Removed a few 'null' variables from the map file
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -668,7 +668,6 @@
 	id_tag = "synd_inner";
 	locked = 1;
 	name = "Ship External Access";
-	req_access = null;
 	req_access_txt = "150"
 	},
 /obj/machinery/atmospherics/pipe/simple/hidden{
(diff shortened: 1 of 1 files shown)


TDSSS:
this is going to merge conflict all mapping prs, won't it?

SabreML:
> this is going to merge conflict all mapping prs, won't it?

In theory it shouldn't, unless they're editing the same stuff I did.
AFAIK mapmerge stops conflicts like that from happening.

variableundefined:
I don't think they do, might be best to hold off on this PR. Other PR are more important.

variableundefined:
Closing. Due to pending mapping changes I don't think this is worth the conflicts at the moment.

PR #14400, opened 2020-09-21
Solar Flares

Kyep:
## What Does This PR Do
Adds the medium event "Solar Flare".
The event lasts 5-15 minutes, and has two effects:
- Everyone located in the space around the station takes 1 burn damage per second they are in space.
- Solars produce 40x as much power as normal.

The first one is a mixed blessing. It can kill off huge carp swarms, but it can also make space travel around the station very dangerous.

The second one is also a mixed blessing. If you have solars set up correctly, you can fill up your SMES very fast. If, on the other hand, you have violated every safety protocol, and hotwired the solars directly into the powergrid... suddenly there's 9MW of power in the grid and APCs start spitting lightning all over the station.

Secondly, it re-scales the shock chance of APCs.
- It used to be a flat 5% if there was >=5M power in the grid.
- Now its 5% between 2.5M and 5.0M, 10% between 5.0M and 7.5M, and 15% over 7.5M power in the grid. 

## Why It's Good For The Game
- Provides a disincentive against just hotwiring the solars and running the entire shift on a risk-free engine that can never cause any problems.
- Provides an extra incentive for people to learn to use the upcoming SM, instead of just always relying on solars.
- Occasionally makes spacewalks more dangerous.
- Makes overcharging the grid significantly more dangerous - an actual hazard, not just a mild annoyance. It also has a gradual transition from "annoyance" to "really dangerous" instead of just being binary on/off.

## Images
Kentucky Fried Cargotech:
cargo_tech_bbq
For full effect, imagine screams of "LET ME IN!!!".

## Changelog
:cl: Kyep
add: Added Solar Flares, a new medium event which makes space more dangerous and provides a disincentive against hotwiring solars.
/:cl:

### DIFF
--- a/code/datums/weather/weather_types/solar_flare.dm
+++ b/code/datums/weather/weather_types/solar_flare.dm
@@ -1,0 +1,50 @@
+/datum/weather/solar_flare
+	name = "solar flare"
+	desc = "An intense blast of light and heat from the sun, affecting all space around the station."
+
+	telegraph_duration = 300 // 30 seconds
+	telegraph_message = null // handled via event announcement
+
+	weather_message = "<span class='userdanger'><i>A solar flare has arrived! Find shelter!</i></span>"
+	weather_overlay = "light_ash"
+	weather_duration_lower = 3000 // 5-15 minutes
+	weather_duration_upper = 9000
+	weather_color = "#ffff00" // bright yellow
+	weather_sound = 'sound/misc/bloblarm.ogg' // also used by radiation storm and SM
+
+	end_duration = 10 // wind_down() does not do anything for this event, so we just trigger end() semi-immediately
+	end_message = null
+	area_type = /area/space // read generate_area_list() as well below
+	protected_areas = list()
+	target_trait = STATION_LEVEL
+	immunity_type = "burn"
+
+/datum/weather/solar_flare/generate_area_list()
+	..()
+	var/list/bonus_areas = list()
+	for(var/V in get_areas(/area/solar))
+		// no, solars in space are NOT a subtype of /area/space.
+		// no, we don't want to re-path every reference to all the subtypes of /area/solar across every map file.
+		// no, we don't want to change /datum/weather/var/area_type into a list as that requires changing every item that touches weather
+		bonus_areas += V
+	for(var/V in bonus_areas)
+		var/area/A = V
+		if(A.z in impacted_z_levels)
+			impacted_areas |= A
+
+/datum/weather/solar_flare/start()
+	..()
+	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
+	SSsun.solar_gen_rate = 1500 * 40
+
+/datum/weather/solar_flare/weather_act(mob/living/L)
+	L.adjustFireLoss(1)
+	if(prob(10))
+		to_chat(L, "<span class='warning'>The solar flare burns you! Seek shelter!")
+
+/datum/weather/solar_flare/end()
+	if(..())
+		return
+	GLOB.event_announcement.Announce("The solar flare has passed.", "Solar Flare Advisory")
+	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
+	SSsun.solar_gen_rate = 1500
--- a/code/datums/weather/weather_types/solar_flare.dm
+++ b/code/datums/weather/weather_types/solar_flare.dm
@@ -46,5 +46,5 @@
 	if(..())
 		return
 	GLOB.event_announcement.Announce("The solar flare has passed.", "Solar Flare Advisory")
-	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
+	// Ends the temporary 40x increase that happened during the weather event
 	SSsun.solar_gen_rate = 1500
--- a/code/datums/weather/weather_types/solar_flare.dm
+++ b/code/datums/weather/weather_types/solar_flare.dm
@@ -8,7 +8,7 @@
 	weather_message = "<span class='userdanger'><i>A solar flare has arrived! Find shelter!</i></span>"
 	weather_overlay = "light_ash"
 	weather_duration_lower = 5 MINUTES
-	weather_duration_upper = 15 MINUTES
+	weather_duration_upper = 10 MINUTES
 	weather_color = COLOR_YELLOW
 	weather_sound = 'sound/misc/bloblarm.ogg' // also used by radiation storm and SM
 
--- a/code/datums/weather/weather_types/solar_flare.dm
+++ b/code/datums/weather/weather_types/solar_flare.dm
@@ -21,12 +21,10 @@
 
 /datum/weather/solar_flare/generate_area_list()
 	..()
-	var/list/bonus_areas = list()
-	for(var/V in get_areas(/area/solar))
-		// no, solars in space are NOT a subtype of /area/space.
-		// no, we don't want to re-path every reference to all the subtypes of /area/solar across every map file.
-		// no, we don't want to change /datum/weather/var/area_type into a list as that requires changing every item that touches weather
-		bonus_areas += V
+	var/list/bonus_areas = get_areas(/area/solar)
+	// no, solars in space are NOT a subtype of /area/space.
+	// no, we don't want to re-path every reference to all the subtypes of /area/solar across every map file.
+	// no, we don't want to change /datum/weather/var/area_type into a list as that requires changing every item that touches weather
 	for(var/V in bonus_areas)
 		var/area/A = V
 		if(A.z in impacted_z_levels)
@@ -40,7 +38,7 @@
 /datum/weather/solar_flare/weather_act(mob/living/L)
 	L.adjustFireLoss(1)
 	if(prob(10))
-		to_chat(L, "<span class='warning'>The solar flare burns you! Seek shelter!")
+		to_chat(L, "<span class='warning'>The solar flare burns you! Seek shelter!</span>")
 
 /datum/weather/solar_flare/end()
 	if(..())
--- a/code/datums/weather/weather_types/solar_flare.dm
+++ b/code/datums/weather/weather_types/solar_flare.dm
@@ -2,14 +2,14 @@
 	name = "solar flare"
 	desc = "An intense blast of light and heat from the sun, affecting all space around the station."
 
-	telegraph_duration = 300 // 30 seconds
+	telegraph_duration = 30 SECONDS
 	telegraph_message = null // handled via event announcement
 
 	weather_message = "<span class='userdanger'><i>A solar flare has arrived! Find shelter!</i></span>"
 	weather_overlay = "light_ash"
-	weather_duration_lower = 3000 // 5-15 minutes
-	weather_duration_upper = 9000
-	weather_color = "#ffff00" // bright yellow
+	weather_duration_lower = 5 MINUTES
+	weather_duration_upper = 15 MINUTES
+	weather_color = COLOR_YELLOW
 	weather_sound = 'sound/misc/bloblarm.ogg' // also used by radiation storm and SM
 
 	end_duration = 10 // wind_down() does not do anything for this event, so we just trigger end() semi-immediately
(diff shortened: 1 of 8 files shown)


code/modules/events/solarflare.dm
```diff
@@ -0,0 +1,29 @@
+/datum/event/solar_flare
+	startWhen = 10
+	endWhen = 500 // = ~500x2 seconds or ~16m40s
+
+/datum/event/solar_flare/announce()
+	GLOB.event_announcement.Announce("A solar flare has been detected on collision course with the station.", "Flare Alert", new_sound = 'sound/AI/attention.ogg')
+
+/datum/event/solar_flare/start()
+	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
```
AffectedArc07, line 9: Part of me wants to add an orange tint to space tiles, in the same way that they glow green when theres a radstorm, but make it much less saturated, so its noticeable but not intrusive

code/modules/power/solar.dm
```diff
@@ -1,5 +1,6 @@
 #define SOLAR_MAX_DIST 40
-#define SOLARGENRATE 1500
+
+GLOBAL_VAR_INIT(solar_gen_rate, 1500)
```
AffectedArc07, line 3: Make this a var on `SSsun` please, we have enough globs already

AffectedArc07 (changes requested):

Spacemanspark:
This is neat.

My only question is if it has any visual effect.

code/modules/events/solarflare.dm
```diff
@@ -0,0 +1,29 @@
+/datum/event/solar_flare
+	startWhen = 10
+	endWhen = 500 // = ~500x2 seconds or ~16m40s
+
+/datum/event/solar_flare/announce()
+	GLOB.event_announcement.Announce("A solar flare has been detected on collision course with the station.", "Flare Alert", new_sound = 'sound/AI/attention.ogg')
```
TDSSS, line 6: Maybe you could include a warning? Like how radstorms tell you to seek cover, maybe it should say "suspend all EVA activity until further notice" or something so people unfamiliar with the event don't get blindsided.

SteelSlayer:
I like the idea of this, but hotwiring solars isn't really done anymore. Ever since the change to the power outage event, there's no point. So instead, I think solar flares will just provide a ton of easy and free power from that will top off the SMES, if they weren't already.

code/modules/events/solarflare.dm
Kyep: I would, but I don't see any way to do that.

Kyep:
@Spacemanspark No and Yes.
@SteelSlayer I saw them hotwired when I checked two days ago. It is still done.

code/modules/events/solarflare.dm
Kyep: I've debated doing that, but:
1) It would completely negate the point of the event, because people would de-hotwire solars before it kicked in.
2) The space damage is not very big - you have enough time to get inside before it cripples you.

code/modules/events/solarflare.dm
matttheficus: Is it possible to take from the Ashstorm code at all to make it affect all non-room marked areas on the Z-Level? When Elysian does the Lavaland Refugee event, we have to use Permits and such to mark rooms so they dont get burned, so perhaps its possible to use the same system for the tint?

lolzyking:
I like hotwiring the grid with solars so I can run the emergency shield generator 24/7 without powersinking the station...

code/modules/events/solarflare.dm
kugamo: I did some testing making a weather event called firestorm. Weather events let you define what areas get overlays. So its a simple matter of making an icon and setting the affected area to /turf/space.

Kyep: Converted to weather datum.

Kyep:
Update:
- Converted entire event to use a weather datum.
- This means there's now a yellow overlay on the affected space tiles.

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+	weather_color = "#ffff00" // bright yellow
+	weather_sound = 'sound/misc/bloblarm.ogg' // also used by radiation storm and SM
+
+	end_duration = 10 // wind_down() does not do anything for this event, so we just trigger end() semi-immediately
+	end_message = null
+	area_type = /area/space // read generate_area_list() as well below
+	protected_areas = list()
+	target_trait = STATION_LEVEL
+	immunity_type = "burn"
+
```
AffectedArc07, line 21: This is weighted 0 by default, so it *should* never happen outside of events

code/modules/events/solarflare.dm
```diff
@@ -0,0 +1,10 @@
+/datum/event/solar_flare
+	startWhen = 2
+	endWhen = 3
+	announceWhen = 1
```
AffectedArc07, line 4: Are there no defines for these?

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+/datum/weather/solar_flare/weather_act(mob/living/L)
+	L.adjustFireLoss(1)
+	if(prob(10))
+		to_chat(L, "<span class='warning'>The solar flare burns you! Seek shelter!")
+
+/datum/weather/solar_flare/end()
+	if(..())
+		return
+	GLOB.event_announcement.Announce("The solar flare has passed.", "Solar Flare Advisory")
+	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
```
AffectedArc07, line 49: Change this to "Lowers the rate from the 40x increase" or something similar

code/modules/events/solarflare.dm
Kyep: They're a counter for the number of ticks. Not mode switches.

code/datums/weather/weather_types/solar_flare.dm
Kyep: Changed

code/modules/events/solarflare.dm
Kyep: I ended up giving it a 30 second telegraphing/advanced warning. I'm really skeptical that this is a good idea, but I'll try it. In the event that people leave solars hotwired, and just de-hotwire them when they get the alert... I'm going to remove the warning completely.

code/modules/power/apc.dm
```diff
@@ -1195,8 +1195,14 @@
 			charging = 0
 			chargecount = 0
 
-		if(excess >= 5000000 && !shock_proof) //If there's more than 5,000,000 watts in the grid, start arcing and shocking people.
-			if(prob(5))
+		if(excess >= 2500000 && !shock_proof)
+			var/shock_chance = 5 // 5%
+			if(excess >= 5000000)
```
TDSSS, line 1200: instead of nested ifs, couldn't you use a switch for this?

Kyep: I changed to using else if. I don't want to use switch as it would be more lines for the same effect.

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+	desc = "An intense blast of light and heat from the sun, affecting all space around the station."
+
+	telegraph_duration = 300 // 30 seconds
+	telegraph_message = null // handled via event announcement
+
+	weather_message = "<span class='userdanger'><i>A solar flare has arrived! Find shelter!</i></span>"
+	weather_overlay = "light_ash"
+	weather_duration_lower = 3000 // 5-15 minutes
+	weather_duration_upper = 9000
+	weather_color = "#ffff00" // bright yellow
```
farie82, line 12: Can use `COLOR_YELLOW` for this

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
+/datum/weather/solar_flare
+	name = "solar flare"
+	desc = "An intense blast of light and heat from the sun, affecting all space around the station."
+
+	telegraph_duration = 300 // 30 seconds
```
farie82, line 5: `30 SECONDS`

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+	telegraph_message = null // handled via event announcement
+
+	weather_message = "<span class='userdanger'><i>A solar flare has arrived! Find shelter!</i></span>"
+	weather_overlay = "light_ash"
+	weather_duration_lower = 3000 // 5-15 minutes
```
farie82, line 10: `5 MINUTES`

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+	weather_duration_upper = 9000
```
farie82, line 11: `15 MINUTES`

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+		bonus_areas += V
+	for(var/V in bonus_areas)
+		var/area/A = V
+		if(A.z in impacted_z_levels)
+			impacted_areas |= A
+
+/datum/weather/solar_flare/start()
+	..()
+	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
+	SSsun.solar_gen_rate = 1500 * 40
```
farie82, line 38: Best to use the initial of the var.

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+	L.adjustFireLoss(1)
+	if(prob(10))
+		to_chat(L, "<span class='warning'>The solar flare burns you! Seek shelter!")
+
+/datum/weather/solar_flare/end()
+	if(..())
+		return
+	GLOB.event_announcement.Announce("The solar flare has passed.", "Solar Flare Advisory")
+	// Ends the temporary 40x increase that happened during the weather event
+	SSsun.solar_gen_rate = 1500
```
farie82, line 50: Same

farie82 (changes requested):

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+
+/datum/weather/solar_flare/start()
+	..()
+	// Solars produce 40x as much power. 240KW becomes 9.6MW. Enough to cause APCs to arc all over the station if >=2 solars are hotwired.
+	SSsun.solar_gen_rate = initial(SSsun.solar_gen_rate) * 40
+
+/datum/weather/solar_flare/weather_act(mob/living/L)
+	L.adjustFireLoss(1)
+	if(prob(10))
+		to_chat(L, "<span class='warning'>The solar flare burns you! Seek shelter!")
```
farie82, line 43: missing end of span

code/datums/weather/weather_types/solar_flare.dm
```diff
@@ -0,0 +1,50 @@
...
+	end_message = null
+	area_type = /area/space // read generate_area_list() as well below
+	protected_areas = list()
+	target_trait = STATION_LEVEL
+	immunity_type = "burn"
+
+/datum/weather/solar_flare/generate_area_list()
+	..()
+	var/list/bonus_areas = list()
+	for(var/V in get_areas(/area/solar))
```
farie82, line 25: No need for this loop.
Just a `bonus_areas = get_areas(/area/solar)` will do here

farie82 (approved):

TDSSS:
> * Provides a disincentive against just hotwiring the solars and running the entire shift on a risk-free engine that can never cause any problems.
> * Provides an extra incentive for people to learn to use the upcoming SM, instead of just always relying on solars.

I don't think this PR will actually accomplish its goals. People will just...set up solars normally, then enjoy a risk-free engine that can never cause any problems and on occasion has extra-full SMES. Not being able to hotwire solars isn't actually a huge downside to it in general.

Kyep:
@TDSSS What if I also added a variant of the event, "solar blackout" or something, where an asteroid gets between the station and the sun? This would drop solar output to 0 for 10-30 minutes, and would mean the solars aren't always perfectly reliable even if set up correctly. That would expand the scope of the PR though, from "nerfs people hotwiring the solars" to "nerfs the solars overall".

Necaladun:
Might want to tweak the length it lasts just to avoid it being too annoying, otherwise I think it's good.

taukausanake:
> What if I also added a variant of the event, "solar blackout" 

The station is supposed to be orbiting a celestial object, isn't it? You could make it that instead of an asteroid. Would make more sense and it would have to be a big af asteroid to block the station

Kyep:
@Necaladun Reduced max length from 15m to 10m.

PR #14414, opened 2020-09-22
Fixes meteor sat emag oversight and cleans event code a bit

### LINKED ISSUE #14413: Emagged meteor sats do not increase meteor event weight
**Issue Description**:
Emagged meteor sats do not increase the weight of meteor events, which they should.

They should make meteor events more common

**What happened instead**:
The base weight of the meteor event gets multiplied by 2...but the base weight is 0, since the event relies entirely on role weight, meaning the emagged sat does nothing.

Things don't work as intended

Look at the code for meteor shield sats and see they do
M.weight *= 2
Then look at event_container.dm and notice the weight of meteor events is 0, meaning this does nothing.

**Extra information**:
The meteor event relies entirely on job weighting to fire, which is unaffected by this. It probably should be multiplied by the modifier as well.

TDSSS:
## What Does This PR Do
This PR does two things. It
fixes #14413
by introducing a system letting you modify the weight of events in general, taking into account both normal weight and job-weight.
Secondly, it does some clean-up and light refactoring in event code while I was there.  Replacing 1 with TRUE and making it so that, if no max weight is set, the event listing actually says that instead of displaying a 0. Using 0 as a max weight that's not set seems pretty confusing to me.

## Why It's Good For The Game
Bugs are bad, refactors are good.

## Changelog
:cl:
tweak: light event refactor
fix: emagged meteor sats now increase meteor event chance as intended.
/:cl:

### DIFF
--- a/code/modules/events/event.dm
+++ b/code/modules/events/event.dm
@@ -63,13 +63,13 @@
 	/// How long the event has existed. You don't need to change this.
 	var/activeFor		= 0
 	/// If this event is currently running. You should not change this.
-	var/isRunning		= 1
+	var/isRunning		= TRUE
 	/// When this event started.
 	var/startedAt		= 0
 	/// When this event ended.
 	var/endedAt			= 0
 	/// Does the event end automatically after endWhen passes?
-	var/noAutoEnd       = 0
+	var/noAutoEnd       = FALSE
 	/// The area the event will hit
 	var/area/impact_area
 	var/datum/event_meta/event_meta = null
@@ -145,14 +145,14 @@
 		tick()
 
 	if(activeFor == startWhen)
-		isRunning = 1
+		isRunning = TRUE
 		start()
 
 	if(activeFor == announceWhen)
 		announce()
 
 	if(activeFor == endWhen && !noAutoEnd)
-		isRunning = 0
+		isRunning = FALSE
 		end()
 
 	// Everything is done, let's clean up.
@@ -167,7 +167,7 @@
 /datum/event/proc/kill()
 	// If this event was forcefully killed run end() for individual cleanup
 	if(isRunning)
-		isRunning = 0
+		isRunning = FALSE
 		end()
 
 	endedAt = world.time
--- a/code/modules/events/event.dm
+++ b/code/modules/events/event.dm
@@ -5,11 +5,12 @@
 	var/min_weight	= 0 // The minimum weight that this event will have. Only used if non-zero.
 	var/max_weight	= 0 // The maximum weight that this event will have. Only use if non-zero.
 	var/severity 	= 0 // The current severity of this event
-	var/one_shot	= 0	//If true, then the event will not be re-added to the list of available events
+	var/one_shot	= 0	// If true, then the event will not be re-added to the list of available events
+	var/weight_mod	= 1	// A modifier applied to all event weights (role or base), respects min and max
 	var/list/role_weights = list()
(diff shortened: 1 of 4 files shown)


AffectedArc07:
Ignore this, misclick
image

code/modules/events/event.dm
```diff
@@ -5,11 +5,12 @@
 	var/min_weight	= 0 // The minimum weight that this event will have. Only used if non-zero.
```
farie82, line 5: All these except `weight_mod` and `enabled` don't have to be set due to them being instantly overriden in the New

farie82 (changes requested):

code/modules/events/event.dm
```diff
@@ -5,11 +5,12 @@
...
 	var/max_weight	= 0 // The maximum weight that this event will have. Only use if non-zero.
 	var/severity 	= 0 // The current severity of this event
-	var/one_shot	= 0	//If true, then the event will not be re-added to the list of available events
+	var/one_shot	= 0	// If true, then the event will not be re-added to the list of available events
```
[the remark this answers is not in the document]
TDSSS: I knew I forgot some

SteelSlayer (approved):

code/modules/events/event.dm
```diff
@@ -1,15 +1,16 @@
 /datum/event_meta
 	var/name 		= ""
-	var/enabled 	= 1	// Whether or not the event is available for random selection at all
-	var/weight 		= 0 // The base weight of this event. A zero means it may never fire, but see get_weight()
-	var/min_weight	= 0 // The minimum weight that this event will have. Only used if non-zero.
-	var/max_weight	= 0 // The maximum weight that this event will have. Only use if non-zero.
-	var/severity 	= 0 // The current severity of this event
-	var/one_shot	= 0	//If true, then the event will not be re-added to the list of available events
+	var/enabled 	= TRUE	// Whether or not the event is available for random selection at all
```
farie82, line 3: Since you're refactoring the file. Can you make these work with auto doc?
https://github.com/ParadiseSS13/Paradise/blob/master/.github/AUTODOC_GUIDE.md

TDSSS: Entire file should be documented now

code/modules/events/event.dm
```diff
@@ -49,62 +52,91 @@
...
+	/// When in the lifetime to call announce().
+	var/announceWhen	= 0
+	/// When in the lifetime the event should end.
+	var/endWhen			= 0
+	/// Severity. Lower means less severe, higher means more severe. Does not have to be supported. Is set on New().
+	var/severity		= 0
+	/// How long the event has existed. You don't need to change this.
+	var/activeFor		= 0
+	/// If this event is currently running. You should not change this.
+	var/isRunning		= 1
```
farie82, line 66: boolean value here

code/modules/events/event.dm
```diff
@@ -49,62 +52,91 @@
...
+	/// When this event started.
+	var/startedAt		= 0
+	/// When this event ended.
+	var/endedAt			= 0
+	/// Does the event end automatically after endWhen passes?
+	var/noAutoEnd       = 0
```
farie82, line 72: Same

farie82 (commented):
Good stuff on the added documentation.
Two little nitpicks are all that's left from what I can tell

farie82 (approved):

PR #14415, opened 2020-09-22
Stops maintenance drones from clicking stuff while ventcrawling

SabreML:
## What Does This PR Do

Stops maintenance drones from using any click-based interactions when ventcrawling, since afaik none of them are intentional.
This includes (but is probably not limited to):

**1: Doors
2: Firelocks
3: Fire alarms
4: Fire extinguisher cabinets
5: Light bulbs/tubes
6: Door buttons *(e.g. Medbay/Security Lockdown)*
6.5: Actually all buttons *(Mass driver, toxins/turbine ignition switch, etc)*
7: Robotics bots**

Also adds `is_ventcrawling()` as a define, as shorthand for `(istype(A.loc, /obj/machinery/atmospherics))`.

Also also removes the ability to ctrl-click fire alarms for silicons, since now you can just click normally to toggle it.

*(and technically reverts #13758)*

## Why It's Good For The Game

Bugfixes!

## Changelog
:cl:
del: Removes ctrl-clicking fire alarms for silicons, since you can now just click them to toggle.
fix: Maintenance drones can no longer interact with objects while ventcrawling (e.g. Firelocks, Airlocks, Light bulbs, etc.)
/:cl:

### DIFF
--- a/code/_onclick/cyborg.dm
+++ b/code/_onclick/cyborg.dm
@@ -185,12 +185,6 @@
 /obj/machinery/turretid/BorgAltClick(mob/living/silicon/robot/user) //turret lethal on/off. Forwards to AI code.
 	AIAltClick(user)
 
-
-// FIRE ALARMS
-
-/obj/machinery/firealarm/BorgCtrlClick(mob/living/silicon/robot/user) // Turns on/off fire alarms. Forwards to AI code.
-	AICtrlClick(user)
-
 /*
 	As with AI, these are not used in click code,
 	because the code for robots is specific, not generic.
--- a/code/_onclick/cyborg.dm
+++ b/code/_onclick/cyborg.dm
@@ -6,7 +6,7 @@
 	adjacency code.
 */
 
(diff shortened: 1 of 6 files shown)


SteelSlayer (approved):

code/_onclick/cyborg.dm
```diff
@@ -15,6 +15,8 @@
 		return
 	changeNext_click(1)
 
+	if(is_ventcrawling(src)) // To stop drones interacting with anything while ventcrawling
```
farie82, line 18: This also stops them from examining stuff and what not. Which is fine but it doesn't sound like a fix but more like a tweak

farie82 (approved):
Code is fine

TDSSS:
Being unable to examine seems kind of annoying.

code/_onclick/cyborg.dm
SabreML: Just tested it on master, and it looks like they can't examine anyway.

Kyep (approved):

PR #14421, opened 2020-09-23
Prevents Curare from being synthesized 

Qwertytoforty:
## What Does This PR Do

This pr gives Curare, a chemical normally only avaible in traitor bottles, can_synth = false, preventing it from being synthesized.

## Why It's Good For The Game

Traitor bottle chemicals generaly have the no synth flag if their only source is from the traitor bottle, to prevent bees / medical mechs / strange plants from making them, due to their strong chemical abilities that would be bad to have an unlimited supply of. Curare acts very similarly to Pancuronium, a chemical that has no recipe that can be synthasized, however, Curare can only be found in traitor bottles and not medbots and NanoMeds, unlike Pancuronium, and also unlike pancuronium, it deals one toxin per cycle, lasts twice as long in a persons system and has penetrates_skin = TRUE, which means it can be splashed on someone, or used in a gas, and affect them even if they have a gas mask on. In effect it is a Pancuronium plus, and for all the previously listed reasons, it should not be able to be synthesized. 
## Changelog
:cl:
tweak: Curare can no longer be synthesized. 
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/toxins.dm
+++ b/code/modules/reagents/chemistry/reagents/toxins.dm
@@ -917,6 +917,7 @@
 	reagent_state = LIQUID
 	color = "#191919"
 	metabolization_rate = 0.1
+	can_synth = FALSE
 	penetrates_skin = TRUE
 	taste_mult = 0
 


variableundefined (approved):

farie82 (approved):

TDSSS:
I don't think Curare is powerful enough that it strictly needs to be unable to be synth'd, but it does make sense from a consistency POV.

Kyep (approved):

PR #14433, opened 2020-09-25
Adds Atmospheric Anomaly

TDSSS:
## What Does This PR Do
This PR adds a new anomaly type, the atmos transformation anomaly. Like other anomalies, it first spawns and has a rather minor affect while warning crew via an announcement. It can be scanned and neutralized as usual during this time, yielding an anomaly core.
Each anomaly randomly picks a gas from among Nitrogen, N2O and CO2 on creation. The anomaly produces small amounts of its chosen gas until it explodes.

When it explodes in a small boom, it transforms **all** Oxygen and Nitrogen in its current area into its chosen gas. This presents a danger to any unprepared crew, but can easily be survived by simply putting on internals. Roundstart atmos should generally fix the issue without player intervention eventually, though atmos techs with portable scrubbers and co can of course speed this up a lot. In my tests, chapel (a fairly large area) was completely cleaned up automatically in 20 minutes.

I considered also having it produce plasma, but decided against it, since a potential plasma flood followed by a fire sounds like more than a medium event generally should be.

## Why It's Good For The Game
The station's air supply system is in most cases totally unnecessary. Rooms contain enough air to last an entire shift, distro is unneeded unless rooms need to be completely refilled following sabotage or breaches. Since it is mostly unnecessary, crew also doesn't hesitate to unwrench pipes against vent-crawlers without suffering any negative consequences from it.

This event is fairly harmless and easily dealt with as long as nobody has unwrenched pipes or otherwise messed with the roundstart setup.

It is also generally a good thing to have more variety in potential events, since they are one way that the regular rounds are kept fresh without any admin intervention.

## Changelog
:cl:
add: A new anomaly, the atmos transformation anomaly transforms breathable air into something quite different.
/:cl:

### DIFF
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -3,6 +3,7 @@
 #define GAS_PL	(1 << 2)
 #define GAS_CO2	(1 << 3)
 #define GAS_N2O	(1 << 4)
+#define GAS_A_B	(1 << 5)
 
 //ATMOS
 //stuff you should probably leave well alone!
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -3,7 +3,6 @@
 #define GAS_PL	(1 << 2)
 #define GAS_CO2	(1 << 3)
 #define GAS_N2O	(1 << 4)
-#define GAS_A_B	(1 << 5)
 
 //ATMOS
 //stuff you should probably leave well alone!
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -3,6 +3,7 @@
 #define GAS_PL	(1 << 2)
 #define GAS_CO2	(1 << 3)
 #define GAS_N2O	(1 << 4)
+#define GAS_A_B (1 << 5)
 
 //ATMOS
 //stuff you should probably leave well alone!
--- a/code/__DEFINES/atmospherics.dm
+++ b/code/__DEFINES/atmospherics.dm
@@ -3,7 +3,7 @@
 #define GAS_PL	(1 << 2)
 #define GAS_CO2	(1 << 3)
 #define GAS_N2O	(1 << 4)
-#define GAS_A_B (1 << 5)
+#define GAS_A_B	(1 << 5)
 
 //ATMOS
 //stuff you should probably leave well alone!
--- a/code/game/objects/effects/anomalies.dm
+++ b/code/game/objects/effects/anomalies.dm
@@ -193,7 +193,7 @@
 
 /obj/effect/anomaly/atmos/New()
 	..()
-	gas_type = pick(GAS_N2O, GAS_CO2, GAS_N2, GAS_A_B)
+	gas_type = pick(GAS_N2O, GAS_CO2, GAS_N2)
 	aSignal.origin_tech = "materials=7"	//turning gas into another gas has some interesting implications for material science
 										//might also work as biotech maybe? Since there's no anomaly for that.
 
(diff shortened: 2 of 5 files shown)


AffectedArc07:
image
No. This gas is extremely inaccessible for a reason.

TDSSS:
> No. This gas is extremely inaccessible for a reason.
@AffectedArc07 
I don't think a 1 in 4 chance of getting agent B from a single rather rare event (and then needing to skip getting an anomaly core from it) is much more accessible than just finding it in a space ruin.

Kyep (approved):
I like the idea of this.
I do suggest taking Agent B out of the list, though, just in the interests of getting this merged within a reasonable period of time.

TDSSS:
Hmm, I'll see what the other maints think about it adding agent B. Especially fox as the person that added it probably has an opinion one way or another. If they're against it too I'll remove it.

Carthusia:
Question: What IS Agent B? Ive never heard of it till this point

AffectedArc07:
> Question: What IS Agent B? Ive never heard of it till this point

Agent B is a gas with a very high heat capacity (10 **TIMES** that of CO2). When reacting, it consumes CO2 and creates oxygen. The results are absolutely devastating, and there is a reason this gas is almost inaccessible, due to the sheer power of it.

Necaladun:
Plasma wouldn't be too bad if it strips all the oxygen in the room out, and it wouldn't have an ignition spot.

TDSSS:
It is so inaccessible it might as well not exist. Without code diving nobody would know it's a thing. Better to have features actually have a chance to show up, if you ask me.

variableundefined:
I object to the inclusion of Agent B, but otherwise I'm onboard with this anomaly

TDSSS:
Agent B has been removed from the event.

code/game/objects/effects/anomalies.dm
```diff
@@ -183,3 +183,30 @@
...
+/////////////////////
+
+/obj/effect/anomaly/atmos
+	name = "transformative gas anomaly"
+	icon_state = "electricity2"
+	var/gas_type
+
+/obj/effect/anomaly/atmos/New()
+	..()
+	gas_type = pick(GAS_N2O, GAS_CO2, GAS_N2)
```
SteelSlayer, line 196: I think this can be made more efficient if you instead assign `gas_type` to one of `LINDA_SPAWN_CO2`, `LINDA_SPAWN_N2O` or `LINDA_SPAWN_NITROGEN`.

And then in your `anomalyEffect()`, you can remove the switch statement and just call:
```dm
T.atmos_spawn_air(LINDA_SPAWN_20C | gas_type, 10)
```

code/__DEFINES/atmospherics.dm
```diff
@@ -3,7 +3,6 @@
 #define GAS_PL	(1 << 2)
 #define GAS_CO2	(1 << 3)
 #define GAS_N2O	(1 << 4)
-#define GAS_A_B	(1 << 5)
```
variableundefined: I'd leave this define in for the sake of consistency - In case someone want to use it for another purpose. 

It is weird to have a define list of all gas only to have one type explicitly missing

code/game/objects/effects/anomalies.dm
variableundefined: I think the performance impact is minimal - and the code is clearer this way. Plus, he'll have to use the gas type later on.

variableundefined (changes requested):
Approved with just one minor change needed.

code/__DEFINES/atmospherics.dm
TDSSS: Done

variableundefined (approved):

PR #14440, opened 2020-09-26
Changes cryoxadone to check external temp instead of internal temp

warior4356:
## What Does This PR Do
Cryoxadone now checks the temperature of the atmosphere or cryo cell around you rather than your body temperature.

## Why It's Good For The Game
The code from dropping or raising your body temp via chems was never designed for cryoxadone pills. It was made to do damage, not to heal. Currently, if you consume a reagent that is at 0 kelvin it drops your body temp by 250 kelvin per chem tick. This system needs to be reworked, but until that happens this PR is a bandaid measure to remove the exploit of 0 kelvin cryoxadone pills.

## Changelog
:cl:
tweak: Cryoxadone checks your external temperature instead of your internal temperature.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine.dm
@@ -126,7 +126,7 @@
 		var/obj/machinery/atmospherics/unary/cryo_cell/C = M.loc
 		external_temp = C.temperature_archived
 	else
-		var/turf/T = M.loc
+		var/turf/T = get_turf(M)
 		external_temp = T.temperature
 	if(external_temp < TCRYO)
 		update_flags |= M.adjustCloneLoss(-4, FALSE)
--- a/code/modules/reagents/chemistry/reagents/medicine.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine.dm
@@ -121,7 +121,14 @@
 
 /datum/reagent/medicine/cryoxadone/on_mob_life(mob/living/M)
 	var/update_flags = STATUS_UPDATE_NONE
-	if(M.bodytemperature < TCRYO)
+	var/external_temp
+	if(istype(M.loc, /obj/machinery/atmospherics/unary/cryo_cell))
+		var/obj/machinery/atmospherics/unary/cryo_cell/C = M.loc
+		external_temp = C.temperature_archived
+	else
+		var/turf/T = M.loc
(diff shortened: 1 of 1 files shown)


Mitchs98:
If this affects ghetto cryo showers, which is definitely something I don't want it to effect as it's fairly niche, then I'd rather not.

I'm not sure which a cold shower effects. Otherwise if it doesn't I'm fine with it.

Cocacolagua:
I do like the idea of nerfing cryoxpills, but i am unsure if this is the way to go about it

Pennwick:
> 
> 
> I do like the idea of nerfing cryoxpills, but i am unsure if this is the way to go about it

Some other alternatives would be making it so hot/cold items don't chill your body that much unless there is a lot of them. The fact that 5u is more than enough to bring you to cryo temps is a little weird.

Also weird that pills retain their near absolute zero temp when just sitting in your pocket. I thought chems were supposed to drift toward ambient temperature unless in an insulated container? Not sure if it just happens very slowly or if its not happening to pills at all.

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -121,7 +121,14 @@
 
 /datum/reagent/medicine/cryoxadone/on_mob_life(mob/living/M)
 	var/update_flags = STATUS_UPDATE_NONE
-	if(M.bodytemperature < TCRYO)
+	var/external_temp
+	if(istype(M.loc, /obj/machinery/atmospherics/unary/cryo_cell))
+		var/obj/machinery/atmospherics/unary/cryo_cell/C = M.loc
+		external_temp = C.temperature_archived
+	else
+		var/turf/T = M.loc
```
farie82, line 129: `loc` can be a locker as well here. You'd want to use `get_turf(M)` here instead

farie82 (changes requested):

farie82 (approved):
Code will work

Necaladun:
I'd prefer to see a bigger more holistic fix, but sometimes you need to use a bandaid. 

Would like a bit of time to wait for people to think up what else this may affect. If shower-ghetto-cryo is hurt by this...well, not really bothered by a feature that is barely ever used or known about.

AffectedArc07 (approved):
I hear the chemists reeeing in the distance

PR #14442, opened 2020-09-26
Makes mining vendors drop inserted IDs when destroyed/deconstructed

### LINKED ISSUE #14423: A mining vendor ate an ID and didnt return it upon the machines destruction
**Issue Description**:
My ID was put into a mining vendor machine and when I went to buy stuff, the machine was inoperable. which made me need to destroy it to get my ID out, however it didnt spit out my ID.

I put my ID in the machine I buy shit, I pull out my ID

**What happened instead**:'
I put my ID in machine, I tried to use it, however nothing happened. No mining vendor popup. 
I then destroyed the machine, but it didnt spit out my ID.

Its a bug. Causing extra work for a miner who lost their ID.

Not sure how the machine got into that state, but it accepted my ID and wouldnt open up the interface.

Noticed it in the current round.

**Extra information**:
Round information: Started at 8:40 PM EST. 
I ahelped and MysticLiger responded so they may have more info other than what I have.

SabreML:
## What Does This PR Do

This makes it so that when mining vendors are destroyed they drop any inserted ID cards, rather than just deleting them.

Fixes #14423 

## Why It's Good For The Game

Less ID cards lost to the void.

## Changelog
:cl:
fix: Mining vendors now drop any inserted ID cards when destroyed/deconstructed
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -266,7 +266,7 @@
 /obj/machinery/mineral/equipment_vendor/Destroy()
 	if(inserted_id)
 		inserted_id.forceMove(loc)
-	..()
+	return ..()
 
 
 /**********************Mining Equiment Vendor (Golem)**************************/
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -263,6 +263,12 @@
 	if(prob(50 / severity) && severity < 3)
 		qdel(src)
 
+/obj/machinery/mineral/equipment_vendor/Destroy()
+	if(inserted_id)
+		inserted_id.forceMove(loc)
+	..()
+
+
 /**********************Mining Equiment Vendor (Golem)**************************/
 
 /obj/machinery/mineral/equipment_vendor/golem


code/modules/mining/machine_vending.dm
```diff
@@ -263,6 +263,12 @@
 	if(prob(50 / severity) && severity < 3)
 		qdel(src)
 
+/obj/machinery/mineral/equipment_vendor/Destroy()
+	if(inserted_id)
+		inserted_id.forceMove(loc)
+	..()
```
farie82, line 269: `return ..()`
Destroy returns a value which is used to determine what happens next to the object. If it gets queued in the garbage collector. Or if it gets hard deleted etc

farie82 (changes requested):

code/modules/mining/machine_vending.dm
SabreML: I thought that'd get called anyway, since `..()` is there.

Change it to `return ..()`, then?

farie82: only calling `..()` will just call the parent. if you do `return ..()` it will call the parent and return the value the parent returns. So aye aye

farie82 (approved):

SteelSlayer (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #14455, opened 2020-09-27
Fixes zero-G spiderweb weirdness

### LINKED ISSUE #12838: Terror Spider webs do not work in Zero Gravity for spiders
**Issue Description**:

For carbon humanoid mobs which the crew is, they can use terror spider webs in zero gravity to push off of and will not get stuck in open areas.  For terror spiders they get stuck on their own webs in zero gravity resulting in them being unable to move around at all.

For spiders to push off their own webs just like the crew.

**What happened instead**:

Spiders get stuck on their webs away from walls.

It allows players to exploit spiders by turning off gravity which forces them to stay near walls for if they touch a web away from walls they get stuck

Get a bunch of terror webs in an open area, put a terror spider in the middle, disable gravity.  Spider cannot move.

Probably was an issue for a while but no one reported it.

**Extra information**:

Kyep:
## What Does This PR Do
Fixes #12838 and similar cases in zero-gravity where spiders (both terror & giant):
- are unable to push off of webs & similar spider structures, even in cases where normal non-spider crew can do it.
- can get completely stuck in zero-g, floating helplessly forever in the middle of a hallway, unable to move or recover no matter what they do.

The fix is: spiders (both kinds) can now always move in zero-gravity IF they are standing on (or next to) a spider structure (e.g: web).

## Why It's Good For The Game
- Spiderwebs are no longer bizzarely more helpful to the crew in zero-G than they are to the spiders who created them. That was unintended, and made no sense. Worst of all was where a spider floating along and hitting a web would stop the spider BUT not enable them to move - thus causing a web to actively harm the spider, which was extremely illogical.
- Spiders who get stuck in a zero-G location that isn't next to a web (e.g: middle of 3-tile-wide hallway) are no longer forced to either ahelp to get moved, or remain there, stuck forever, until someone comes along to kill them. Or ghost out of their body while alive, which gives up their respawnability. Now, they have another option: spin a web, then use that web to propel themselves to the nearest wall, then cling to that. They're still vulnerable while spinning the web, but at least they're not stuck FOREVER, even while supposedly standing on their own web.

## Changelog
:cl: Kyep
tweak: spiders (both giant, and terror) can now "push off" their own structures to move in zero-G, like normal crew already could "push off" a spider's web to move.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/giant_spider.dm
+++ b/code/modules/mob/living/simple_animal/hostile/giant_spider.dm
@@ -47,6 +47,14 @@
 		if(C.can_inject(null, FALSE, inject_target, FALSE))
 			C.reagents.add_reagent("spidertoxin", venom_per_bite)
 
+/mob/living/simple_animal/hostile/poison/giant_spider/get_spacemove_backup()
+	. = ..()
+	// If we don't find any normal thing to use, attempt to use any nearby spider structure instead.
+	if(!.)
+		for(var/obj/structure/spider/S in range(1, get_turf(src)))
+			. = S
+			break
+
 //nursemaids - these create webs and eggs
 /mob/living/simple_animal/hostile/poison/giant_spider/nurse
 	desc = "Furry and black, it makes you shudder to look at it. This one has brilliant green eyes."
--- a/code/modules/mob/living/simple_animal/hostile/giant_spider.dm
+++ b/code/modules/mob/living/simple_animal/hostile/giant_spider.dm
@@ -52,8 +52,7 @@
 	// If we don't find any normal thing to use, attempt to use any nearby spider structure instead.
 	if(!.)
 		for(var/obj/structure/spider/S in range(1, get_turf(src)))
-			. = S
-			break
+			return S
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/hostile/giant_spider.dm
```diff
@@ -47,6 +47,14 @@
 		if(C.can_inject(null, FALSE, inject_target, FALSE))
 			C.reagents.add_reagent("spidertoxin", venom_per_bite)
 
+/mob/living/simple_animal/hostile/poison/giant_spider/get_spacemove_backup()
+	. = ..()
+	// If we don't find any normal thing to use, attempt to use any nearby spider structure instead.
+	if(!.)
+		for(var/obj/structure/spider/S in range(1, get_turf(src)))
+			. = S
```
SteelSlayer, line 55: Is there a reason this can't be `return S`? Same for the one in the other file.

SteelSlayer (approved):

AffectedArc07 (approved):

PR #14461, opened 2020-09-27
Makes it so that not every datum uses an UID everytime it gets destroyed

farie82:
## What Does This PR Do
Fixes the issue where we run out of UIDs for no reason

## Why It's Good For The Game
Bug fix

## Changelog
:cl:
fix: Makes datums that have no UID not make one to try and close non existing TGUI windows
/:cl:


AffectedArc07 (approved):
Objects without a TGUI attached wont have a UID, so this is an effective fix. Its hacky and I dont like it, but its better than wasting UIDs on qdel'd objects

PR #14191, opened 2020-08-30
Allows toy E-Swords to reflect Laser Tag lasers

SabreML:
## What Does This PR Do

This lets you reflect Laser Tag lasers with a toy sword, in order to get the "Full Jedi Experience!™"
Very fluffy, but I thought it'd be fun.

## Why It's Good For The Game

- Allows people to give those laser taggers a taste of their own medicine.

- Gives the arcade more of a use, I guess.

## Changelog
:cl:
add: Double bladed toy swords can now reflect laser tag beams
/:cl:

*(🎉25th PR, woo!🎉)*

PR #14467, opened 2020-09-27
Fixes lasers ignoring shields of all types.

### LINKED ISSUE #14466: Lasers ignore shielded hardsuits
**Issue Description**:
Shielded hardsuits block all projectiles/melee expect lasers.

For the shield to also ignore the laser.

**What happened instead**:
Laser ignores shield completely, making the user vulnerable.

Shielded hardsuits come from nuclear operatives, wizards and cultists, this level of gear shouldn’t be able to be ignored by lasers.

1. Wear any kind of shielded hardsuit.
2. Fire at the shielded hardsuit wearer with a laser/disabler/tazer

This is quite recent, since it worked fine atleast a month ago.

**Extra information**:
Reading the code, this might also happen with the RD’s armor.

Aquilar:
## What Does This PR Do

fixes the error introduced by #14191 
the error was once a projectile was checked as reflectable, it returned no matter what instead of moving on.
Fixes #14466

## Why It's Good For The Game

fixes are gooood.

## Changelog
:cl:
fix: lasers no longer ignore shields
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -22,7 +22,7 @@ emp_act
 
 		else if(can_reflect == 2) //If target is holding a toy sword
 			var/list/safe_list = list(/obj/item/projectile/beam/lasertag, /obj/item/projectile/beam/practice)
-			if(is_type_in_list(P, safe_list)) //And it's not safe
+			if(is_type_in_list(P, safe_list)) //And it's safe
 				visible_message("<span class='danger'>The [P.name] gets reflected by [src]!</span>", \
 				   "<span class='userdanger'>The [P.name] gets reflected by [src]!</span>")		
 				P.reflect_back(src)
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -16,6 +16,11 @@ emp_act
 		var/can_reflect = check_reflect(def_zone)
 		if(!can_reflect)
 			return (..(P , def_zone)) //Bad luck
+		else if(can_reflect == 1)		
+			P.reflect_back(src)
+
+			return -1 // complete projectile permutation
+
 		else
 			if(can_reflect == 2) //If target is holding a toy sword
 				var/list/safe_list = list(/obj/item/projectile/beam/lasertag, /obj/item/projectile/beam/practice)
@@ -24,9 +29,9 @@ emp_act
 		visible_message("<span class='danger'>The [P.name] gets reflected by [src]!</span>", \
 						"<span class='userdanger'>The [P.name] gets reflected by [src]!</span>")
 
-		P.reflect_back(src)
+			P.reflect_back(src)
 
-		return -1 // complete projectile permutation
+			return -1 // complete projectile permutation
 
 	//Shields
 	if(check_shields(P, P.damage, "the [P.name]", PROJECTILE_ATTACK, P.armour_penetration))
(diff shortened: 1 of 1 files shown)


SteelSlayer:
Tested locally and I'm still able to fire lasers through the shields of shielded hardsuits. It doesn't appear this fixes the problem.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -16,17 +16,21 @@ emp_act
 		var/can_reflect = check_reflect(def_zone)
 		if(!can_reflect)
```
Aquilar, line 17: I realised what the actual problem was when I went to sleep.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -14,19 +14,21 @@ emp_act
...
+
+		else if(can_reflect == 2) //If target is holding a toy sword
+			var/list/safe_list = list(/obj/item/projectile/beam/lasertag, /obj/item/projectile/beam/practice)
+			if(!is_type_in_list(P, safe_list)) //And it's not safe
+				return (..(P , def_zone)) //Bad luck
+			visible_message("<span class='danger'>The [P.name] gets reflected by [src]!</span>", \
+						"<span class='userdanger'>The [P.name] gets reflected by [src]!</span>")		
+			P.reflect_back(src)
+
+			return -1 // complete projectile permutation
```
SteelSlayer, line 31: I know you didn't add/change this but, I have some have some concerns about the possible early return in here on line 26. Before the toy sword reflect PR, if something failed to reflect, no matter what, it would continue on through the rest of the proc. There was no early return like this. But now, it can terminate here, skipping some important stuff below about shields and whatnot. So, I think this might need to be something like:

```dm
	else if(can_reflect == 2) //If target is holding a toy sword
		var/list/safe_list = list(/obj/item/projectile/beam/lasertag, /obj/item/projectile/beam/practice)
		if(is_type_in_list(P, safe_list)) //And it's not safe
			visible_message("<span class='danger'>The [P.name] gets reflected by [src]!</span>", \
						"<span class='userdanger'>The [P.name] gets reflected by [src]!</span>")		
			P.reflect_back(src)
			return -1 // complete projectile permutation
```

Aquilar: yeah could fix that, didn't think it would be much of an issue unless you considered adding more toys with reflect

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -14,19 +14,20 @@ emp_act
...
+		if(can_reflect == 1) // proper reflection
+			visible_message("<span class='danger'>The [P.name] gets reflected by [src]!</span>", \
 						"<span class='userdanger'>The [P.name] gets reflected by [src]!</span>")
+			P.reflect_back(src)
+			return -1 // complete projectile permutation
 
-		P.reflect_back(src)
+		else if(can_reflect == 2) //If target is holding a toy sword
+			var/list/safe_list = list(/obj/item/projectile/beam/lasertag, /obj/item/projectile/beam/practice)
+			if(is_type_in_list(P, safe_list)) //And it's not safe
```
SteelSlayer, line 25:
```suggestion
			if(is_type_in_list(P, safe_list)) //And it's safe
```

SteelSlayer (approved):
Little comment nitpick since the logic got swapped, but otherwise the code LGTM.

SabreML:
Thanks for fixing my bug for me. 👍

AffectedArc07 (approved):

PR #14469, opened 2020-09-27
New traitor item: The Doppelganger Armor

Cocacolagua:
## What Does This PR Do
Adds a new 11TC item in the stealthy-tools section of the uplink, the doppelganger armor!

The armor has two abilities:
 1- Chameleon armor slot, you can change the sprite to whatever you want to disguise it.
 2- The Doppelganger: After you activate the armor on-hand, you will be protected by the reactive armor! When hit, you will turn invisible and a fake-copy of the user will manifest and start running away from the aggressor, which will drive away security forces off your tail! Five seconds after activation, the clone will disappear and you will turn back visible again. This ability has a cooldown, so you can't spam it endlessly at once. Don't get EMPd!

## Why It's Good For The Game
Traitors are severely lacking in anti-stun tools, as the only ones currently are the adrenalin implant (only three uses) and the stimulants (only one use). This item aims for an anti-stun that will last the traitor for as long as they are wearing it, that's why it has the following downsides:
       - 11 Telecrystals cost, so it will zap away most of your budget unless you get lucky with discounts (item is banned for nukies)
       - Invisible does not mean Invulnerable, although you will not take damage from the activating hit, any hits while invisible will still affect you, same as items like the energy-bola, flashbangs and the like.
       - Being EMP'd will turn off your armor for a long time, so security can disable it with an ion rifle as an opening move if they know what's up.
       - The armor only works while you have it equipped, and the armor stats are not great, so you won't be able to stack a lot of armor to go with this.

CYVOEO65qz

## Changelog
:cl:
add: Adds The Doppelganger Armor to the uplink, an 11TC chameleon suit that will turn you invisible and create a fake clone of yourself when hit!
/:cl:

### DIFF
--- a/code/modules/clothing/chameleon.dm
+++ b/code/modules/clothing/chameleon.dm
@@ -292,17 +292,16 @@
 /obj/item/clothing/suit/chameleon/doppel/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
 	if(!active | cooldown == TRUE)
 		return FALSE
-	else
-		var/mob/living/simple_animal/hostile/illusion/escape/E = new(owner.loc)
-		E.Copy_Parent(owner, 70)
-		E.GiveTarget(owner) //so it starts running right away
-		E.Goto(owner, E.move_to_delay, E.minimum_distance)
-		owner.alpha = 0
-		owner.visible_message("<span class='danger'>[owner] is hit by [attack_text] in the chest!</span>") //We pretend to be hit, since blocking it would stop the message otherwise
-		addtimer(CALLBACK(src, .proc/doppel_cloaking, owner), 70)
-		cooldown = TRUE
-		addtimer(CALLBACK(src, .proc/doppel_cooldown), 300)
-		return TRUE
+	var/mob/living/simple_animal/hostile/illusion/escape/E = new(owner.loc)
+	E.Copy_Parent(owner, 70)
+	E.GiveTarget(owner) //so it starts running right away
+	E.Goto(owner, E.move_to_delay, E.minimum_distance)
+	owner.alpha = 0
+	owner.visible_message("<span class='danger'>[owner] is hit by [attack_text] in the chest!</span>") //We pretend to be hit, since blocking it would stop the message otherwise
+	addtimer(CALLBACK(src, .proc/doppel_cloaking, owner), 70)
+	cooldown = TRUE
+	addtimer(CALLBACK(src, .proc/doppel_cooldown), 300)
+	return TRUE
 
 /obj/item/clothing/glasses/chameleon
 	name = "Optical Meson Scanner"
--- a/code/modules/clothing/chameleon.dm
+++ b/code/modules/clothing/chameleon.dm
@@ -257,6 +257,51 @@
 	. = ..()
 	chameleon_action.emp_randomise(INFINITY)
 
+// Traitor item "Doppelganger Suit", behaves the same way as the reactive stealth armor, but has the added bonus of chameleon disguise.
+/obj/item/clothing/suit/chameleon/doppel
+	var/active = 0
+	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
+
+/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cooldown(mob/user)
+	cooldown = FALSE
+
+/obj/item/clothing/suit/chameleon/doppel/attack_self(mob/user)
+	active = !(active)
+	if(active)
+		to_chat(user, "<span class='notice'>[src] is now active.</span>")
+	else
+		to_chat(user, "<span class='notice'>[src] is now inactive.</span>")
+		add_fingerprint(user)
+	user.update_inv_wear_suit()
+	for(var/X in actions)
+		var/datum/action/A = X
+		A.UpdateButtonIcon()
+
+/obj/item/clothing/suit/chameleon/doppel/emp_act(severity)
+	active = 0
+	cooldown = TRUE
+	addtimer(CALLBACK(src, .proc/doppel_cooldown), 400)
+	if(istype(loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/C = loc
+		C.update_inv_wear_suit()
+	..()
+
+/obj/item/clothing/suit/chameleon/doppel/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
+	if(!active | cooldown == TRUE)
+		return 0
+	else
+		var/mob/living/simple_animal/hostile/illusion/escape/E = new(owner.loc)
+		E.Copy_Parent(owner, 50)
+		E.GiveTarget(owner) //so it starts running right away
+		E.Goto(owner, E.move_to_delay, E.minimum_distance)
+		owner.alpha = 0
+		owner.visible_message("<span class='danger'>[owner] is hit by [attack_text] in the chest!</span>") //We pretend to be hit, since blocking it would stop the message otherwise
+		spawn(50)
+			owner.alpha = initial(owner.alpha)
+		cooldown = TRUE
+		addtimer(CALLBACK(src, .proc/doppel_cooldown), 300)
+		return 1
+
 /obj/item/clothing/glasses/chameleon
 	name = "Optical Meson Scanner"
 	desc = "Used by engineering and mining staff to see basic structural and terrain layouts through walls, regardless of lighting condition."
--- a/code/modules/clothing/chameleon.dm
+++ b/code/modules/clothing/chameleon.dm
@@ -265,7 +265,7 @@
 /obj/item/clothing/suit/chameleon/doppel/proc/doppel_cooldown(mob/user)
 	cooldown = FALSE
 
-/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cloaking(mob/user)
+/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cloaking(mob/living/carbon/human/owner)
 	owner.alpha = initial(owner.alpha)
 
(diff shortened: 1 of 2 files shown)


MarsM0nd:
Saying it's an anti-stun, wouldn't it kinda fail at that, given you spawn a simple animal, which can't be stunned? So someone wearing it gets tazed/batoned, seems like they keep standing from that (while the invisible agent can run away), would either be handled as KoS due to the apperant stun immunity, or instantly seen through it, so they know they need to shoot around and watch for open doors. If they were getting shot with lethals, should be getting noticed only after the copy disappears.
Also, how would it handle HuDs? The alpha should handle it on the agent, but does the doppleganger? If not, an attack disappearing the HuD is quite an easy tell for it.

Cocacolagua:
I mean yeah, using the armor should land you into the KoS spot since stuns do not work unless you can read where the invisible agent is moving.

I'll check the HUD behaviour, however.

Cocacolagua:
After some testing with HUDS the behavior is this: 
Medical HUDs do not show the invisible traitor, and show a 100% health display on the clone, even if the user isn’t at full health.
Security HUDs do not show the invisible traitor, but they also do not show on the fake clone, since they are a simplemob.

I'm willing to say there’s nothing necessarily bad about this, as a careful security officer will notice what’s up (they already would’ve when they tried to stun probably), and a SUPER careful medical doctor will notice the HUD differences (even though they it could just be the effect of potent healing.

Cocacolagua:
CYVOEO65qz

Ignore the de-sync between the disguise and the invisibility for now plz

PacifistDalek:
We TF2 Dead Ringer now kids

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,51 @@
 	. = ..()
 	chameleon_action.emp_randomise(INFINITY)
 
+// Traitor item "Doppelganger Suit", behaves the same way as the reactive stealth armor, but has the added bonus of chameleon disguise.
+/obj/item/clothing/suit/chameleon/doppel
+	var/active = 0
```
SteelSlayer, line 262: Since this var is behaving like a boolen
```suggestion
	var/active = FALSE
```

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,51 @@
...
+	else
+		to_chat(user, "<span class='notice'>[src] is now inactive.</span>")
+		add_fingerprint(user)
+	user.update_inv_wear_suit()
+	for(var/X in actions)
+		var/datum/action/A = X
+		A.UpdateButtonIcon()
+
+/obj/item/clothing/suit/chameleon/doppel/emp_act(severity)
+	active = 0
```
SteelSlayer, line 281:
```suggestion
	active = FALSE
```

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,51 @@
...
+// Traitor item "Doppelganger Suit", behaves the same way as the reactive stealth armor, but has the added bonus of chameleon disguise.
+/obj/item/clothing/suit/chameleon/doppel
+	var/active = 0
+	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
+
+/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cooldown(mob/user)
+	cooldown = FALSE
+
+/obj/item/clothing/suit/chameleon/doppel/attack_self(mob/user)
+	active = !(active)
```
SteelSlayer, line 269: Can leave out the parentheses here.
```suggestion
	active = !active
```

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,51 @@
...
+	cooldown = TRUE
+	addtimer(CALLBACK(src, .proc/doppel_cooldown), 400)
+	if(istype(loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/C = loc
+		C.update_inv_wear_suit()
+	..()
+
+/obj/item/clothing/suit/chameleon/doppel/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
+	if(!active | cooldown == TRUE)
+		return 0
```
SteelSlayer, line 291:
```suggestion
		return FALSE
```

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,51 @@
...
+		E.Copy_Parent(owner, 70)
+		E.GiveTarget(owner) //so it starts running right away
+		E.Goto(owner, E.move_to_delay, E.minimum_distance)
+		owner.alpha = 0
+		owner.visible_message("<span class='danger'>[owner] is hit by [attack_text] in the chest!</span>") //We pretend to be hit, since blocking it would stop the message otherwise
+		spawn(65)
+			owner.alpha = initial(owner.alpha)
+		cooldown = TRUE
+		addtimer(CALLBACK(src, .proc/doppel_cooldown), 300)
+		return 1
```
SteelSlayer, line 303:
```suggestion
		return TRUE
```

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,51 @@
...
+		return 0
+	else
+		var/mob/living/simple_animal/hostile/illusion/escape/E = new(owner.loc)
+		E.Copy_Parent(owner, 70)
+		E.GiveTarget(owner) //so it starts running right away
+		E.Goto(owner, E.move_to_delay, E.minimum_distance)
+		owner.alpha = 0
+		owner.visible_message("<span class='danger'>[owner] is hit by [attack_text] in the chest!</span>") //We pretend to be hit, since blocking it would stop the message otherwise
+		spawn(65)
+			owner.alpha = initial(owner.alpha)
```
SteelSlayer, line 300: You should make this into callback just like you did with `doppel_cooldown`

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,53 @@
...
+	if(!active | cooldown == TRUE)
+		return FALSE
+	else
+		var/mob/living/simple_animal/hostile/illusion/escape/E = new(owner.loc)
+		E.Copy_Parent(owner, 70)
+		E.GiveTarget(owner) //so it starts running right away
+		E.Goto(owner, E.move_to_delay, E.minimum_distance)
+		owner.alpha = 0
+		owner.visible_message("<span class='danger'>[owner] is hit by [attack_text] in the chest!</span>") //We pretend to be hit, since blocking it would stop the message otherwise
+		addtimer(CALLBACK(src, .proc/doppel_cloaking), 70)
```
SteelSlayer, line 302: Since your `doppel_cloaking` proc doesn't know what `owner` is, you need to pass it an argument here in the callback:
```dm
addtimer(CALLBACK(src, .proc/doppel_cloaking, owner), 70)
```
And then with the cloaking proc:
```dm
/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cloaking(mob/living/carbon/human/owner)
	owner.alpha = initial(owner.alpha)
```

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,53 @@
...
+	addtimer(CALLBACK(src, .proc/doppel_cooldown), 400)
+	if(istype(loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/C = loc
+		C.update_inv_wear_suit()
+	..()
+
+/obj/item/clothing/suit/chameleon/doppel/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
+	if(!active | cooldown == TRUE)
+		return FALSE
+	else
```
farie82, line 295: not required else

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,53 @@
...
+// Traitor item "Doppelganger Suit", behaves the same way as the reactive stealth armor, but has the added bonus of chameleon disguise.
+/obj/item/clothing/suit/chameleon/doppel
+	var/active = FALSE
+	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
+
+/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cooldown(mob/user)
+	cooldown = FALSE
+
+/obj/item/clothing/suit/chameleon/doppel/proc/doppel_cloaking(mob/living/carbon/human/owner)
+	owner.alpha = initial(owner.alpha)
```
farie82, line 269: Can you check if this breaks other abilities that make you near invisible? I'm not sure if those are constantly applied or not

code/modules/clothing/chameleon.dm
```diff
@@ -257,6 +257,53 @@
...
+	user.update_inv_wear_suit()
+	for(var/X in actions)
+		var/datum/action/A = X
+		A.UpdateButtonIcon()
+
+/obj/item/clothing/suit/chameleon/doppel/emp_act(severity)
+	active = FALSE
+	cooldown = TRUE
+	addtimer(CALLBACK(src, .proc/doppel_cooldown), 400)
+	if(istype(loc, /mob/living/carbon/human))
```
farie82, line 287: `ishuman`

farie82 (changes requested):

code/modules/clothing/chameleon.dm
Cocacolagua: tested it with changelings’, the clone is invisible if you’re invisible, and when the ability ends, the user is visible. tbh i’m considering this is fine, given that this and the genetic power are the only things that turn the user invisible. (and also to activate the ability you gotta hit the wearer)

farie82 (approved):

Necaladun:
I think this is worth giving a go. If it's OP, we jack the price up/nerf it some.

Necaladun (approved):
I think this is worth giving a go. If it's OP, we jack the price up/nerf it some.

Cocacolagua:
Closing since:
- PR is stuck in freeze hell
- #15343 completely overshadows it anyway lmao

PR #14472, opened 2020-09-27
Codebase Cleanup - 2020-09-27

AffectedArc07:
## What Does This PR Do
Cleans up some old pieces of the codebase. Everything done in this PR is listed below in bulk, to save diff reading

- Removed DLLSocket from the bay12 era before Paradise even existed
- Repathed `/obj/vox/win_button` to `/obj/machinery/vox_win_button` (MAP EDIT). This is to remove an almost top-level path which shouldnt exist
- Renamed a bunch of filenames which had spaces to remove their spaces (Spaces in filenames are bad)
- Tweaked stack crafting so it uses `obj/result` instead of `/obj/result`, which caused the game to have a literal definition for `/obj/result` ingame
- Removed `/obj/crate` and `/obj/crate/fireworks`. These dont exist ingame, dont work, and have not been touched in **7 years**
- Removed POS (Its not used anywhere, defines a `/line_item` top level path which shouldnt exist, and was a half-implemented port that was never finished
- Removes `UnstandardnessTestForDM` from tools. This has been replaced with `DreamChecker`. 
- Moves `Runtime Condenser` to `RuntimeCondenser` since spaces in filenames are a pain, especially with things which are shell automated, such as the runtime condenser 

## Why It's Good For The Game
No ingame change, purely a github/maintenence improvement

## Changelog
No ingame change


code/game/area/dynamic_areas.dm
```diff
@@ -1,44 +1,44 @@
-/area/dynamic // Do not use.
```
AffectedArc07, line 1: No idea why this changed line endings

_maps/map_files/MetaStation/z2.dmm
```diff
@@ -1312,7 +1312,7 @@
 	},
 /area/vox_station)
 "dA" = (
-/obj/vox/win_button,
```
AffectedArc07, line 1315: These are the only map edits in the entire thing

Kyep (approved):

code/game/objects/items/stacks/stack_recipe.dm
```diff
@@ -24,27 +24,27 @@
 	src.on_floor = on_floor
 	src.window_checks = window_checks
 
-/datum/stack_recipe/proc/post_build(var/obj/item/stack/S, var/obj/result)
+/datum/stack_recipe/proc/post_build(var/obj/item/stack/S, obj/result)
```
farie82, line 27: Can you remove the other `var/X`s from this file as well? Atleast for the procs you touched

farie82 (changes requested):
Minor nitpick

farie82 (approved):

PR #14493, opened 2020-09-30
Adds Procedure radio to Magistrate, NT Rep, and IAA

warior4356:
## What Does This PR Do
This PR adds a new radio frequency called procedure for the IAA, NT Rep, and Magistrate. The AI and captain also have access. The new frequency is on ":x" and vocal cords have been moved to ":~".

## Why It's Good For The Game
This is good for the game because it provides a place other than sec comms for these jobs to communicate about SOP issues. This is good because one security comms are often far too busy to have a conversation about SOP and for two frequently the SOP violations are from security and they shouldn't be part of that conversation.

## Images of changes
Samples of radio color
dreamseeker_2020-09-30_14-53-25
dreamseeker_2020-09-30_14-53-07

## Changelog
:cl:
add: Added a new oversight radio frequency on :x for the NT Rep, IAA, Magistrate, Captain, and AI
tweak: Vocal cords are now on :~
/:cl:


AffectedArc07:
That darkmode colour is so unreadable I don't even know where to start. Change please.

code/game/objects/items/devices/radio/encryptionkey.dm
```diff
@@ -88,7 +93,7 @@
 /obj/item/encryptionkey/heads/captain
 	name = "Captain's Encryption Key"
 	icon_state = "cap_cypherkey"
-	channels = list("Command" = 1, "Security" = 1, "Engineering" = 0, "Science" = 0, "Medical" = 0, "Supply" = 0, "Service" = 0)
+	channels = list("Command" = 1, "Security" = 1, "Engineering" = 0, "Science" = 0, "Medical" = 0, "Supply" = 0, "Service" = 0, "Oversight" = 0)
```
AffectedArc07, line 96: Shouldn't the captain have this on by default?

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -128,6 +128,14 @@
 	icon_state = "sec_headset_alt"
 	item_state = "sec_headset_alt"
 
+/obj/item/radio/headset/headset_iaa
+	name = "internal affairs bowman headset"
+	desc = "This is used by your elite legal team. Protects ears from flashbangs."
+	flags = EARBANGPROTECT
```
AffectedArc07, line 134: Not sure if IAA should be flashbang proof. The NT rep ain't.

goon/browserassets/css/browserOutput-dark.css
```diff
@@ -282,6 +282,7 @@ em						{font-style: normal;	font-weight: bold;}
 .sciradio				{color: #993399;}
 .supradio				{color: #9F8545;}
 .srvradio				{color: #80A000;}
+.overadio				{color: #4e3b25;}
```
AffectedArc07, line 285: Again, this needs to change

AffectedArc07 (changes requested):

code/game/objects/items/devices/radio/headset.dm
warior4356: They already were. Previously they got a sec headset and I didn’t want to balance change. When I brought this up the rather valid point I heard was that flash bangs get used in sec and IAA is in sec a lot.

I’m not against changing it though.

code/game/objects/items/devices/radio/encryptionkey.dm
warior4356: Unsure on that point, the channel was more for NT Rep, IAA, and magistrate in my mind with the option of the captain being involved. Still I have no real opinion on this point and will bow to the consensus.

denghis1:
I like this, being able to have a channel to discuss SOP with the other IAA/Magistrate/NTRep without bothering everyone else ( is welcome.  
I also agree with affected about the colour, i missed the text entirely on my first glance at it.

meow20:
With how dynamic some rounds are, and with how rp-less some roles feel even in calm rounds this feels like a great step in a direction to revive underused part of the game.

MarsM0nd:
Not sure if kicking the unathi language off O and even giving it a two-letter key is the best decision. It seems to be usually for never used/tertiary languages, like the monkey languages, or ones that will only be used as default, like terror spider ones. Changing a languages key may also result in a few muscle-memory wrong usages at first.
Of course, the whoole alphabet is currently in use, but special characters seem to work. *, &, (, or ) may work, though the brackets are the least intuitive. * usually notes some fine print, so the focus on the details is there. & would be and about that.
§ would probably be perfect, as it's basically only used in legal code, but only a few keyboards have dedicated keybinds for it, using alt+21 on the numpad is probably not a good alternative for that.

code/game/objects/items/devices/radio/encryptionkey.dm
warior4356: It is now on by default for the captain.

goon/browserassets/css/browserOutput-dark.css
warior4356: The color has been changed.

code/game/objects/items/devices/radio/headset.dm
Kiyahitayika: To that point and bringing in something Kyet said last which I happen to agree with - the IAA _shouldn't_ be just hanging in their office which just happens to be located at the Brig. They should be moving around the station, checking in on departments and other things. Flashbangs suck, sure, but making their headsets immune just because of where they are, I don't know. I support removing the flashbang proofing, especially since this feels like a step in a bigger picture of remodeling the IAA role. I wouldn't be surprised if the future has their office being moved closer to the NT Rep once the map freeze is lifted.

Kyep (approved):

Kiyahitayika:
I strongly disagree with the nano mapping commit.

warior4356:
> I strongly disagree with the nano mapping commit.

I don't see any nano map commit, are you sure you're looking at the right PR?

Kiyahitayika:
> > I strongly disagree with the nano mapping commit.
> 
> I don't see any nano map commit, are you sure you're looking at the right PR?

You're all set.

SteelSlayer:
Not sure if it's intended or not, but your channel doesn't show up on intercoms.

Kyep:
Making it show up on intercomms would require modifying the GLOBAL_LIST_INIT(default_internal_channels
block in radio.dm. This would also require assigning an access number to it.

Kyep:
Having TM'ed this, the channel itself was useful, but the colors did not render right for ghosts:
dreamseeker_CEXDQADMlF
dreamseeker_8O3jxJTbai

warior4356:
Asking around with someone else who had the issue, restarting their client fixed it. I think it's just a client-side issue with a cache of some sort. I never had that happen when testing, so that's my best explanation.

farie82 (approved):

Kyep:
@warior4356 please:
- deconflict
- update from master just to make sure its compatible
- re-run TGUI generation
- resolve all the open comments above (yes, address them, but also use the resolve button)

AffectedArc07 (approved):

PR #14498, opened 2020-09-30
Beverage can fizz

kugamo:
## What Does This PR Do
This PR adds soda fizz as a feature. When you alt click a beverage can, this will shake it up. Throwing the can will shake it up as well. When opened, depending on how much it was shaken, it either splashes some of its contents on you, or erupts into foam if shaken five times or more. If you have shaken it five times already and continue to shake it, it will have a higher and higher chance of opening on its own. These chances are reset after 20 seconds. A minute from when you shake it, your shake is undone and the beverage returns to a less shaken state or unshaken state if only one shake remains.

The liquid inside the beverage cans will be altered in proportion to how much it is shook, in steps of 1/5 of the total contents of the can.

A new sound for opening a shaken beverage can is added too.

Big thank you to @farie82 and @SteelSlayer for helping improve the quality of my code. XXXOOO

## Why It's Good For The Game
Adds the ability to prank your friends or boss in-game, which is always good fun. And some discouragement for roughly handling these beverages. I would also argue it adds to immersion in a mostly harmless way.

## Images of changes

KVrbFBNN2Y

## Changelog
:cl:
add: Adds beverage can fizz, they can now erupt into foam and fizz if shaken!
soundadd: Adds new sound for fizzy openings of beverage cans.
/:cl:


Kyep (approved):

SteelSlayer (commented):
Nipicks. Also having the `attack_self` proc handling the bursting instead of the `burst` proc is a bit weird. IMO it should be structured a bit differently. Nevertheless, it does work as is.

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
+	var/bursting = FALSE
+	var/burstchance = 0
 
 /obj/item/reagent_containers/food/drinks/cans/New()
 	..()
 	flags &= ~OPENCONTAINER
 
-/obj/item/reagent_containers/food/drinks/cans/attack_self(mob/user)
-	if(canopened == 0)
+/obj/item/reagent_containers/food/drinks/cans/attack_self(mob/user, burstopen = FALSE)
```
farie82, line 13: Can you make this one `return ..()`. I did notice that `drinks/attack_self` does not `return ..()`... which prevents it from sending the attackself signal but hey... better ensure this one works when that one gets fixed

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +86,47 @@
 /obj/item/reagent_containers/food/drinks/cans/afterattack(obj/target, mob/user, proximity)
 	if(!proximity)
 		return
-	if(istype(target, /obj/structure/reagent_dispensers) && (canopened == 0))
+	if(istype(target, /obj/structure/reagent_dispensers) && (canopened == FALSE))
```
farie82, line 89: can remove the `()` and make it `!canopened`

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +86,47 @@
...
 		to_chat(user, "<span class='notice'>You need to open the drink!</span>")
 		return
-	else if(target.is_open_container() && (canopened == 0))
+	else if(target.is_open_container() && (!canopened))
```
farie82, line 92: same here for the ()'s

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +86,47 @@
...
+		bursting = TRUE
+		burstchance = 1
+		return
+	if(burstchance < 10)
+		burstchance++
+	if(prob((burstchance * 5)))
+		attack_self(user, TRUE)
+
+/obj/item/reagent_containers/food/drinks/cans/proc/resetshake()
+	shaken =  0
```
farie82, line 123: one to many spaces here

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
+		for(var/mob/living/carbon/C in range(1, get_turf(src)))
+			C.visible_message("<span class='warning'>You are splattered with [name]!</span>")
+			reagents.reaction(C, REAGENT_TOUCH)
+			if(shaken > C.wetlevel)
+				C.wetlevel = shaken
+			else if(shaken == C.wetlevel)
+				C.wetlevel++
+
+		reagents.remove_any((shaken/5) * reagents.total_volume)
+		playsound(loc,'sound/effects/canopenfizz.ogg', rand(10,50), 1)
```
farie82, line 34: spacing

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
+
+		for(var/mob/living/carbon/C in range(1, get_turf(src)))
+			C.visible_message("<span class='warning'>You are splattered with [name]!</span>")
+			reagents.reaction(C, REAGENT_TOUCH)
+			if(shaken > C.wetlevel)
+				C.wetlevel = shaken
+			else if(shaken == C.wetlevel)
+				C.wetlevel++
+
+		reagents.remove_any((shaken/5) * reagents.total_volume)
```
farie82, line 33: spacing

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
+		if(shaken < 5)
+			to_chat(user, "<span class='warning'>The [name] fizzes violently!</span>")
+		else
+			to_chat(user, "<span class='warning'>The [name] erupts into foam!</span>")
+			foam_up()
+
+		for(var/mob/living/carbon/C in range(1, get_turf(src)))
+			C.visible_message("<span class='warning'>You are splattered with [name]!</span>")
+			reagents.reaction(C, REAGENT_TOUCH)
+			if(shaken > C.wetlevel)
```
farie82, line 28: can replace the following 4 lines with:
`C.wetlevel = max(C.wetlevel + 1, shaken)`

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +86,47 @@
...
+/obj/item/reagent_containers/food/drinks/cans/proc/burst(mob/user)
+	if(!bursting)
+		addtimer(CALLBACK(src, .proc/resetburst), 10 SECONDS)
+		bursting = TRUE
+		burstchance = 1
+		return
+	if(burstchance < 10)
+		burstchance++
+	if(prob((burstchance * 5)))
+		attack_self(user, TRUE)
```
farie82, line 120: Please don't call `attack_self` from code if a user doesn't attack the object itself. It makes the code very impure and is prone for errors.
It's best to make a new proc for the bursting logic and then call that from `attack_self` and in `burst` instead

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
+			reagents.reaction(C, REAGENT_TOUCH)
+			if(shaken > C.wetlevel)
+				C.wetlevel = shaken
+			else if(shaken == C.wetlevel)
+				C.wetlevel++
+
+		reagents.remove_any((shaken/5) * reagents.total_volume)
+		playsound(loc,'sound/effects/canopenfizz.ogg', rand(10,50), 1)
+		canopened = TRUE
+		flags |= OPENCONTAINER
```
farie82, line 36: Can extract the top 3 lines to one scope below and remove line 40, 42, 43

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
+		else
+			visible_message("<span class='warning'>The [name] bursts open!</span>")
+		if(shaken < 5)
+			to_chat(user, "<span class='warning'>The [name] fizzes violently!</span>")
+		else
+			to_chat(user, "<span class='warning'>The [name] erupts into foam!</span>")
+			foam_up()
+
+		for(var/mob/living/carbon/C in range(1, get_turf(src)))
+			C.visible_message("<span class='warning'>You are splattered with [name]!</span>")
```
farie82, line 26: This will show the message to everybody whenever somebody gets splattered.
When X gets splattered Y gets to see this message

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,17 +1,45 @@
...
 	flags &= ~OPENCONTAINER
 
-/obj/item/reagent_containers/food/drinks/cans/attack_self(mob/user)
-	if(canopened == 0)
+/obj/item/reagent_containers/food/drinks/cans/attack_self(mob/user, burstopen = FALSE)
+	if(!canopened && shaken)
+		if(!burstopen)
+			to_chat(user, "<span class='notice'>You open the drink with an audible pop!</span>")
+		else
+			visible_message("<span class='warning'>The [name] bursts open!</span>")
```
farie82, line 18: `[src] burts open!`
The will be added on it's own

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -29,7 +57,16 @@
 	return crushed_can
 
 /obj/item/reagent_containers/food/drinks/cans/attack(mob/M, mob/user, proximity)
-	if(canopened == 0)
+	if(user.a_intent == INTENT_HARM && !canopened)
```
farie82, line 60: You double check `canopened` now. Can put this check in the existing one

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -29,7 +57,16 @@
...
+		visible_message("<span class='warning'>[user.name] shakes up the [name]!</span>")
+		if(shaken < 5)
+			if(!shaken)
+				addtimer(CALLBACK(src, .proc/resetshake), 1 MINUTES)
```
farie82, line 64: shaking it once then waiting 50 seconds and shaking it 3 more times will still reset after the initial 1 minute timer.
Just making sure that is intended

farie82 (changes requested):

code/modules/food_and_drinks/drinks/drinks/cans.dm
kugamo: I moved canopend and flags ahead of the branching to avoid needlessly repeating that code, but the reagents.remove_any needs to be after reagents.react and playsound is circumstantially unique.

code/modules/food_and_drinks/drinks/drinks/cans.dm
kugamo: I learned adding many timers isnt as bad as I thought it might be, so I changed this behavior to loose shakenedness over this time period for each time it is shoook.

FlattestGuitar:
Should be alt+click on the can. Also shown on examine.

Whether a can is open or not is not immediately obvious from the sprite, so people will shake it up instead of drinking from it by accident, which will be annoying.

kugamo:
> 
> 
> Should be alt+click on the can. Also shown on examine.
> 
> Whether a can is open or not is not immediately obvious from the sprite, so people will shake it up instead of drinking from it by accident, which will be annoying.

Considering how niche using alt click is I think I will stick with harm for now. I did add code that will update the description of the beverage once opened, so in that way its knowable whether or not it can fizz. If you do perhaps mistakenly shake up the beverage, its only a minute until its fine again.

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,18 +1,26 @@
...
-		to_chat(user, "<span class='notice'>You open the drink with an audible pop!</span>")
-		canopened = 1
+	if(!canopened)
+		if(shaken)
+			fizzyopen()
+			return
+		playsound(loc, 'sound/effects/canopen.ogg', rand(10, 50), 1)
+		canopened = TRUE
 		flags |= OPENCONTAINER
+		desc += "\nIt has been opened."
```
SteelSlayer, line 21: Instead of needing to add this in multiple locations, I would put this in `examine`:
```dm
/obj/item/reagent_containers/food/drinks/cans/examine(mob/user)
	. = ..()
	if(canopened)
		. += "It has been opened."
```

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +66,71 @@
...
+	if(burstchance < 10)
+		burstchance++
+
+	if(prob((burstchance * 5)))
+		fizzyopen(user, TRUE)
+
+/obj/item/reagent_containers/food/drinks/cans/proc/resetshake()
+	shaken--
+
+/obj/item/reagent_containers/food/drinks/cans/proc/resetburst()
```
SteelSlayer, line 130: Snek case for both of these.
```suggestion
/obj/item/reagent_containers/food/drinks/cans/proc/reset_shake()
	shaken--

/obj/item/reagent_containers/food/drinks/cans/proc/reset_burst()
```

FlattestGuitar:
> Considering how niche using alt click is I think I will stick with harm for now.

Clicking yourself with something on harm intent being used for an interaction other than attacking yourself is actually LESS niche than an alt+click to perform a special action? Are you sure about that?

> Altclick has no timing restrictions either, unlessed manually coded in, so it can be spammed. and I dont know how to force timing checks without using clunky code

There are people in coder chat that will more than gladly help you achieve this task. It's not a challenging enough implementation to warrant settling for less intuitive features.

Take this as an opportunity to grow your skills. 😎

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +76,70 @@
...
+	if(!bursting)
+		addtimer(CALLBACK(src, .proc/reset_bursting), 30 SECONDS)
+		bursting = TRUE
+		burstchance = 1
+		return
+
+	if(burstchance < 10)
+		burstchance++
+
+	if(prob((burstchance * 5)))
```
SteelSlayer, line 133: I suggest doing this so you can avoid having to multiply by 5 in every prob call.
```suggestion
		burstchance = 5
		return

	if(burstchance < 50)
		burstchance += 5

	if(prob((burstchance)))
```

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -28,8 +40,23 @@
 	qdel(src)
 	return crushed_can
 
+/obj/item/reagent_containers/food/drinks/cans/AltClick(mob/user)
+	. = ..()
```
farie82, line 44: This will also lag the player when using it since it'll show the contents of the turf in the top right.
I'd say it's best to do this if you are not shaking it. Meaning it's on the floor or the like

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,18 +1,30 @@
...
+	if(canopened)
+		. += "It has been opened."
+
 /obj/item/reagent_containers/food/drinks/cans/attack_self(mob/user)
-	if(canopened == 0)
-		playsound(loc,'sound/effects/canopen.ogg', rand(10,50), 1)
-		to_chat(user, "<span class='notice'>You open the drink with an audible pop!</span>")
-		canopened = 1
+	if(!canopened)
```
farie82, line 19: Early returns are preferred

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,18 +1,30 @@
...
+		if(shaken)
+			fizzy_open(user)
+			return
+		playsound(loc, 'sound/effects/canopen.ogg', rand(10, 50), 1)
+		canopened = TRUE
 		flags |= OPENCONTAINER
+		to_chat(user, "<span class='notice'>You open the drink with an audible pop!</span>")
+		return ..()
```
farie82, line 27: I feel like it should always call this no matter if the can is opened or not

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +76,70 @@
...
 		to_chat(user, "<span class='notice'>You need to open the drink!</span>")
 		return
-	else if(target.is_open_container() && (canopened == 0))
+	else if(target.is_open_container() && !canopened)
 		to_chat(user, "<span class='notice'>You need to open the drink!</span>")
 		return
 	else
 		return ..(target, user, proximity)
 
+/obj/item/reagent_containers/food/drinks/cans/after_throw(mob/user)
```
farie82, line 88: This one needs to call it's parent at one point since that handles quite some logic. 
The param here is also incorrect:
image
is where the base is defined and called

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +76,70 @@
...
+			var/datum/effect_system/foam_spread/sodafizz = new
+			sodafizz.set_up(1, get_turf(src), reagents)
+			sodafizz.start()
+
+	for(var/mob/living/carbon/C in range(1, get_turf(src)))
+		to_chat(C, "<span class='warning'>You are splattered with [name]!</span>")
+		reagents.reaction(C, REAGENT_TOUCH)
+		C.wetlevel = max(C.wetlevel + 1, shaken)
+
+	reagents.remove_any((shaken / 5) * reagents.total_volume)
```
farie82, line 118: no need for the ()s here around the division

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -49,15 +76,70 @@
...
+
+/obj/item/reagent_containers/food/drinks/cans/proc/bursting(mob/user)
```
farie82, line 120: Best to prevent naming procs and variables the same. `bursting` as a proc also reads as if calling this would always burst the can. Which is not the case here

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,18 +1,30 @@
 /obj/item/reagent_containers/food/drinks/cans
-	var/canopened = 0
+	var/canopened = FALSE
 	var/is_glass = 0
 	var/is_plastic = 0
+	var/shaken = 0
+	var/bursting = FALSE
```
farie82, line 6: `can_burst` here? Bit more descriptive variable name will atleast help. `bursting` reads as if it's already bursting. AKA if an effect is playing or such but here it means that it can burst once the roll happens

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,18 +1,30 @@
 /obj/item/reagent_containers/food/drinks/cans
-	var/canopened = 0
+	var/canopened = FALSE
 	var/is_glass = 0
 	var/is_plastic = 0
+	var/shaken = 0
```
farie82, line 5: minor nitpick: `times_shaken` would be more descriptive here as shaken reads as if it's a boolean value

code/modules/food_and_drinks/drinks/drinks/cans.dm
kugamo: what exactly should be called?? return ..()???

farie82: Indeed so. It will currently not do much to always call this since the parent just returns. But this is a bug more than anything.
`item/attack_self` sends a signal saying attack_self is called on this object. Which other things can subscribe to. So once the parent is fixed calling `..()` will ensure cans also send said signal
Btw unless specified otherwise you can assume it's about the line the comment is on :P

kugamo: Right ok, thanks. Good to know.

code/modules/food_and_drinks/drinks/drinks/cans.dm
kugamo: I switched to using `throw_impact()` hopefully for the better

farie82 (commented):
One minor thing.
Looking at the timers I also noticed this:
If you shake it say 5 times. Then wait 55 seconds and shake again it will reset after 5 seconds. You can use unique timers here to make "reset" the running timer. `TIMER_UNIQUE` and `TIMER_OVERRIDE` are the flags you need to add to the timer to achieve this.
`TIMER_UNIQUE` will make sure that there is only ever one timer of the same type running
`TIMER_OVERRIDE` will then remove the old running timer and ensure the new one gets started.

Btw this is a design thing so if you decide not to change this then that's also just fine.

kugamo:
> 
> 
> One minor thing.
> Looking at the timers I also noticed this:
> If you shake it say 5 times. Then wait 55 seconds and shake again it will reset after 5 seconds. You can use unique timers here to make "reset" the running timer. `TIMER_UNIQUE` and `TIMER_OVERRIDE` are the flags you need to add to the timer to achieve this.
> `TIMER_UNIQUE` will make sure that there is only ever one timer of the same type running
> `TIMER_OVERRIDE` will then remove the old running timer and ensure the new one gets started.
> 
> Btw this is a design thing so if you decide not to change this then that's also just fine.

I used `TIMER_STOPABLE` and `deltimer()` because I had trouble with what you suggested. I am not keen on recoding it, but if someone knows what they are doing and gets the same behaviour using what @farie82 suggests.. then I will happily use it.

variableundefined (changes requested):
Can you please add text to explain that alt-clicking on it allows you to shake it and make a beverage fizz? 

Else this feature would be obscure and unlikely to be discovered by the average player.

kugamo:
> 
> 
> Can you please add examine text to explain that alt-clicking on it allows you to shake it and make a beverage fizz?
> 
> Else this feature would be obscure and unlikely to be discovered by the average player.

@variableundefined I have done this. :^)

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -1,18 +1,35 @@
...
 /obj/item/reagent_containers/food/drinks/cans/New()
 	..()
 	flags &= ~OPENCONTAINER
 
+/obj/item/reagent_containers/food/drinks/cans/examine(mob/user)
+	. = ..()
+	if(canopened)
+		. += "<span class='notice'>It has been opened.</span>"
+	else
+		. += "<span class='notice'>Alt-click to shake it up!</span>"
```
farie82, line 20:
```suggestion
		. += "<span class='info'>Alt-click to shake it up!</span>"
```
`info` is the standard used for gameplay tips

code/modules/food_and_drinks/drinks/drinks/cans.dm
```diff
@@ -28,8 +45,38 @@
...
+		return ..()
+	if(src == H.l_hand || src == H.r_hand)
+		can_shake = FALSE
+		addtimer(CALLBACK(src, .proc/reset_shakable), 1 SECONDS)
+		to_chat(H, "<span class ='notice'>You start shaking up [src].</span>")
+		if(do_after(H, 1 SECONDS, target = H))
+			visible_message("<span class='warning'>[user.name] shakes up the [name]!</span>")
+			if(times_shaken == 0)
+				times_shaken++
+				shake_timer = addtimer(CALLBACK(src, .proc/reset_shaken), 1 MINUTES, TIMER_STOPPABLE)
```
farie82, line 64:
```suggestion
				addtimer(CALLBACK(src, .proc/reset_shaken), 1 MINUTES, TIMER_UNIQUE | TIMER_OVERRIDE | TIMER_NO_HASH_WAIT)
```
Using this will also make it behave the same way it does now using the existing timer system. I'd suggest using this to avoid extra code on your end.
The `TIMER_UNIQUE` says there should only be one timer of this sort. `TIMER_OVERRIDE` says to override the existing timer if any (AKA destroy it). `TIMER_NO_HASH_WAIT` will not include the `wait` parameter in the hashing method. Meaning any value of wait will result in the same hash thus is considered the same timer

Same goes for all the other `addtimer` calls

farie82 (changes requested):

farie82 (changes requested):
Last nitpicks

farie82 (approved):

PR #14499, opened 2020-10-01
Adds reagent vision to Service, Medical, ERT, DS, and Destroyer cyborgs

warior4356:
## What Does This PR Do
As the title states, it allows Service, Medical, ERT, DS, and Destroyer cyborgs to see what reagents are inside a container.

## Why It's Good For The Game
Medical borgs have a syringe they can not see the contents of, similarly service borgs can't see what is in their drinks. The other three are the "super borgs" and probably should have the ability even if it's not relevant most of the time.

## Changelog
:cl:
add: Added reagent vision to Service, Medical, ERT, DS, and Destroyer cyborgs
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -108,6 +108,8 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 	var/datum/action/item_action/toggle_research_scanner/scanner = null
 	var/list/module_actions = list()
 
+	var/see_reagents = FALSE // Determines if the cyborg can see reagents
+
 /mob/living/silicon/robot/get_cell()
 	return cell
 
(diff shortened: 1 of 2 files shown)


Sheppysneik:
... the medical borgs have an advanced reagent scanner though, so they can actually see the reagents by scanning their syringe

SteelSlayer (approved):

warior4356:
> ... the medical borgs have an advanced reagent scanner though, so they can actually see the reagents by scanning their syringe

Good point, I gave service borgs one too, so that they both have both options. The advanced reagent scanner can scan all containers, while chem vision works only on transparent containers. An example of when the advanced scanner would work and vision would not is food or pills.

procdrone:
yes please

Necaladun:
Sounds good!

I'll karma the first DS robot that actually finds a good reason to need this.

Issue #14303, opened 2020-09-14
Orion trail Ghost Suicide


Cocacolagua:
**Issue Description**:
If you spam the "Kill a crewmember" button in the arcade game "orion trail", you recieve a bugged message.

**What did you expect to happen**: 
For a videogame ghost to not kill themselves.

**What happened instead**:
![image](https://user-images.githubusercontent.com/48032385/93134897-d31fae80-f6af-11ea-8d52-c1b29e08dd30.png)

**Why is this bad/What are the consequences**:
immersion breaking, a scary creepypasta, do not play orion trail at 3 AM.

**Steps to reproduce the problem**:
1- Start an Orion Trail game.
2- Spam the "Kill a crewmember button"
3- Realize you are playing a haunted copy of Orion Trail.

**When did the problem start happening**:
I don't remember this happening a year ago, but it might have been happening since it was added.

farie82:
Easy first issue. The Topic does not check if you can still kill a crewmember

further-reading:
I believe I have fixed this in https://github.com/ParadiseSS13/Paradise/pull/14503 However I was not able to click fast enough to make the bug happen before changing it so I can't be sure.

PR #14503, opened 2020-10-01
[orion trail] Check if all crew dead before kill

further-reading:
## What Does This PR Do
This PR adds a check to Orion Trail to see if there are any living crew members left when the kill crewmember button is pressed.

## Why It's Good For The Game
If clicked fast enough you can get a message saying a person without a name committed suicide. This is bad for immersion as this game would not be bugged if it was on a real space station. fixed https://github.com/ParadiseSS13/Paradise/issues/14303

## Changelog
:cl: further-reading
tweak: Orion trail checks for living crew before killing a crew member.
/:cl:

### DIFF
--- a/code/game/machinery/computer/arcade.dm
+++ b/code/game/machinery/computer/arcade.dm
@@ -571,23 +571,24 @@
 		event = null
 
 	else if(href_list["killcrew"]) //shoot a crewmember
-		var/sheriff = remove_crewmember() //I shot the sheriff
-		playsound(loc,'sound/weapons/gunshots/gunshot.ogg', 100, 1)
+		if(settlers.len > 0 || alive > 0)
+			var/sheriff = remove_crewmember() //I shot the sheriff
+			playsound(loc,'sound/weapons/gunshots/gunshot.ogg', 100, 1)
 
-		if(settlers.len == 0 || alive == 0)
-			atom_say("The last crewmember [sheriff], shot themselves, GAME OVER!")
-			if(emagged)
-				usr.death(0)
-				emagged = 0
-			gameover = 1
-			event = null
-		else if(emagged)
-			if(usr.name == sheriff)
-				atom_say("The crew of the ship chose to kill [usr.name]!")
-				usr.death(0)
+			if(settlers.len == 0 || alive == 0)
+				atom_say("The last crewmember [sheriff], shot themselves, GAME OVER!")
+				if(emagged)
+					usr.death(0)
+					emagged = 0
+				gameover = 1
+				event = null
+			else if(emagged)
+				if(usr.name == sheriff)
+					atom_say("The crew of the ship chose to kill [usr.name]!")
+					usr.death(0)
 
-		if(event == ORION_TRAIL_LING) //only ends the ORION_TRAIL_LING event, since you can do this action in multiple places
-			event = null
+			if(event == ORION_TRAIL_LING) //only ends the ORION_TRAIL_LING event, since you can do this action in multiple places
+				event = null
 
 	//Spaceport specific interactions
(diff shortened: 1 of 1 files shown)


SabreML:
If you change `#14303` to `fixed #14303` it'll automatically close the issue when this gets merged.

FlattestGuitar:
Make the button disappear as well if you want to disable its function completely.

code/game/machinery/computer/arcade.dm
```diff
@@ -571,23 +571,24 @@
 		event = null
 
 	else if(href_list["killcrew"]) //shoot a crewmember
-		var/sheriff = remove_crewmember() //I shot the sheriff
-		playsound(loc,'sound/weapons/gunshots/gunshot.ogg', 100, 1)
+		if(settlers.len > 0 || alive > 0)
```
farie82, line 574: An early return is preferred in our codebase. In this case it'll be something like this:
```dm
if(amount <= 0 || alive <= 0)
  return
do_happy_flow
```
Also be sure to use `length(X)` instead of `X.len`. `length(X)` is null pointer exception proof and faster compared to `X.len`

code/game/machinery/computer/arcade.dm
```diff
@@ -571,23 +571,24 @@
...
+			var/sheriff = remove_crewmember() //I shot the sheriff
+			playsound(loc,'sound/weapons/gunshots/gunshot.ogg', 100, 1)
```
farie82, line 576: spacing here between the parameters

code/game/machinery/computer/arcade.dm
```diff
@@ -571,23 +571,24 @@
...
-			gameover = 1
-			event = null
-		else if(emagged)
-			if(usr.name == sheriff)
-				atom_say("The crew of the ship chose to kill [usr.name]!")
-				usr.death(0)
-
-		if(event == ORION_TRAIL_LING) //only ends the ORION_TRAIL_LING event, since you can do this action in multiple places
-			event = null
+			if(settlers.len == 0 || alive == 0)
```
farie82, line 578: Same here for the `.len`. It was already here but refactoring existing code is always a good thing to do!

code/game/machinery/computer/arcade.dm
```diff
@@ -571,23 +571,24 @@
...
+				atom_say("The last crewmember [sheriff], shot themselves, GAME OVER!")
+				if(emagged)
+					usr.death(0)
```
farie82, line 581: `FALSE` here instead of `0`. Boolean values are best to be indicated using the `TRUE/FALSE` defines. Same applies here as above

code/game/machinery/computer/arcade.dm
```diff
@@ -571,23 +571,24 @@
...
+					emagged = 0
```
farie82, line 582: Same here and below for `gameover` and the other `usr.death(0)`

farie82 (changes requested):
Hey welcome to our codebase!
Got some minor styling issues but the logic should be correct.

code/game/machinery/computer/arcade.dm
further-reading: Can you clarify what you mean here? Should it something like `settlers.len>0 || alive>0` or `settlers.len > 0||alive > 0`?

farie82: the if statement itself is fine. I meant the `playsound(loc,'sound....)`. Standard is to put a space after each `,` or operator and the like

further-reading: Ah I understand, thanks!

farie82 (approved):

AffectedArc07 (approved):

PR #14504, opened 2020-10-01
Changes the emag effect and telepad selection for Rapid Crate Senders

SabreML:
## What Does This PR Do

- Changes the emag effect of a Rapid Crate Sender (RCS) to allow you to teleport yourself while inside a locker, as well as choose a random location to teleport to.

- Also changes the way that the destination telepad is set.
Previously, it would ask for a destination every time you tried teleporting a crate.
I've changed it so that it gets set in `attack_hand`, so you only have to set it once before you go crate hunting.

Also also reorganises, cleans up, and documents the code a bit.

## Why It's Good For The Game

Less hassle when teleporting crates, and an actual reason to emag the RCS.

## Changelog
:cl:
add: Emagging a Rapid Crate Sender (RCS) will now allow you to teleport yourself in a locker, as well as being able to choose a random location.
tweak: The destination telepad of an RCS is now set by using it in your hand, rather than whenever you teleport a crate.
tweak: Various other code tweaks.
/:cl:

### DIFF
--- a/code/modules/telesci/rcs.dm
+++ b/code/modules/telesci/rcs.dm
@@ -27,10 +27,7 @@
 	var/teleporting = FALSE
 	/// How much power does each teleport use?
 	var/chargecost = 1000
-
-	// Random coordinates for if emagged
-	var/rand_x = 0
-	var/rand_y = 0
+	/// Is emagged?
 	var/emagged = FALSE
 
 /obj/item/rcs/get_cell()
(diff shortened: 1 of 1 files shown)


code/modules/telesci/rcs.dm
```diff
@@ -38,84 +48,79 @@
...
+		L[locname] = R
+
+	if(emagged) // Add an 'Unknown' entry at the end if it's emagged
+		L += "**Unknown**"
+
+	var/select = input("Please select a telepad.", "RCS") in L
+	if(select == "**Unknown**") // Randomise the teleport location
+		rand_x = rand(50, 200)
+		rand_y = rand(50, 200)
+		pad = locate(rand_x, rand_y, 6)
```
SteelSlayer, line 80: There's a problem here where this tries to set `pad` to something at the given location, but there isn't actually a pad there, so `pad` ends up being null. Which means when you try to teleport something, you get the "Error: No telepad selected." message.

Also why the hardcoded Z-6 here?

code/modules/telesci/rcs.dm
```diff
@@ -38,84 +48,79 @@
...
 	if(teleporting)
 		to_chat(user, "<span class='warning'>You're already using [src]!</span>")
 		return
-	if(user in C.contents) //to prevent self-teleporting.
+	if((user in C.contents) && (!emagged)) // If it's emagged, skip this check.
+		to_chat(user, "<span class='warning'>Error: User located in container--aborting for safety.</span>")
 		return
-	if(!rcell || (rcell.charge < chargecost))
-		to_chat(user, "<span class='warning'>Out of charges.</span>")
+	if(!rcell || rcell.charge < chargecost)
```
SteelSlayer, line 101: Do you think its necessary to check if `rcell` exists? From what I can tell, there's no way to remove a cell from an RCS so I think we can just assume one is there.

code/modules/telesci/rcs.dm
SabreML: I've not been able to reproduce this, can you elaborate?

As for the z6, presumably so that it's always in space? I could make it z1 with a small chance of being 6, or just make it always on the station level.

code/modules/telesci/rcs.dm
```diff
@@ -38,84 +48,79 @@
...
+		to_chat(user, "<span class='boldwarning'>Warning: Safeties disabled.</span>")
 		return
 
+
 /obj/item/rcs/proc/try_send_container(mob/user, obj/structure/closet/C)
 	if(teleporting)
 		to_chat(user, "<span class='warning'>You're already using [src]!</span>")
 		return
-	if(user in C.contents) //to prevent self-teleporting.
+	if((user in C.contents) && (!emagged)) // If it's emagged, skip this check.
```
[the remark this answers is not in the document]
SabreML: I always assumed it checks both conditions anyway, rather than breaking

code/modules/telesci/rcs.dm
SteelSlayer: In the location selection menu, when I choose **Unknown**, and then click on a crate, it says "Error: No telepad selected."

Dumbdumn5 (approved):
Could it be, the return of the fucking RCS teleportation but with an actual requirement to it to make it so not just any random civvie can snag it and teleport away? Though I'll be disappointed about the lack of the old non-antag shenanigans I could do with a teleporting locker, I *do* think this would be a good fucking use for emagging one of these.

PR #14507, opened 2020-10-02
Fixes Wryn's tail and adds the ability to sting

Marginalorb:
## What Does This PR Do
This fixes an issue that Wryn have had for quite some time, and also gives them the ability to sting.
Stinging is a fluff ability that only does a few brute damage and costs somewhere in the ballpark of 20 stamina damage to the user as to prevent people from just spamming it on someone until they die a death of a thousand paper cuts.

A HUGE thank you to Kiyahitayika for figuring out how to, and coding, the ability for Wryn to sting.

## Why It's Good For The Game
This paves the way to potentially adding Wryn as a player species rather than letting them collect dust as a broken relic of the past.

## Images of changes
wryntail
wrynsting

## Changelog
:cl:
add: Adds the ability to sting to wryn, and a button to the HUD to do so
fix: Fixes broken tail code, allowing for wryn to have their bee tails again.
/:cl:

PR #14509, opened 2020-10-02
Refactors how targeted changeling stings are handled

SteelSlayer:
## What Does This PR Do
Removes snowflake changeling checks on the `/mob/living/carbon/MiddleClickOn` and `/mob/living/carbon/AltClickOn` procs, and creates a new `middleClickOverride` that can be used to invoke any callback, which I'm now using with changeling stings. 

Changeling sting functionality remains the same.

## Changelog
None, backend change only.

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -242,11 +242,11 @@
 	return
 
 // See click_override.dm
-/mob/living/MiddleClickOn(var/atom/A)
- if(src.middleClickOverride)
- 	middleClickOverride.onClick(A, src)
- else
- 	..()
+/mob/living/MiddleClickOn(atom/A)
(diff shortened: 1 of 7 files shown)


SteelSlayer:
Ok I swapped this back from being a left click intercept to a middle click intercept. I realized running around and having a prepared sting while still being able to left click things was superior. 

I Left in the cleanup/documentation from `code/datums/click_intercept.dm` because why not. I had already written it so might as well keep it in, although it has nothing to do with the middle click override stuff I've added.

Kiyahitayika:
Thanks, @SteelSlayer ! @AffectedArc07 - if this gets approved, do we have permission to use this type of coding on Wryn Sting in #14507 ?

PR #14513, opened 2020-10-02
Removes the 'bnotice' span class

SabreML:
## What Does This PR Do

Removes the `bnotice` span class, since it's only used in one file and is identical to `boldnotice`.

I noticed this while working on something else, so I thought I might as well make a PR to change it.

## Why It's Good For The Game

The `boldnotice` span class does exactly the same thing, and is easier to undstand.

## Changelog
No in-game effects.

### DIFF
--- a/code/game/objects/items/weapons/AI_modules.dm
+++ b/code/game/objects/items/weapons/AI_modules.dm
@@ -143,8 +143,8 @@ AI MODULES
 		target.set_zeroth_law(law)
 		GLOB.lawchanges.Add("The law specified [targetName]")
 	else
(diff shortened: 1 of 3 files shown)


AffectedArc07:
If youre going through the CSSs, please fix two things
1) Please fix the invalid styling in `.ancient`, it says `font-stye` instead of `font-style`
2) Go through and look at other spans which arent used anywhere, I know we have some

SabreML:
@AffectedArc07 
I'd rather not remove any more span classes if I can avoid it, since that would involve a *lot* of testing to make sure I didn't accidentaly break something.

AffectedArc07:
Fair enough.

AffectedArc07 (approved):

PR #14514, opened 2020-10-02
Reduce Water Stun/Weaken Time

matttheficus:
## What Does This PR Do
Reduces the stun time and weaken time of TURF_WET_WATER by half.

## Why It's Good For The Game
Water is basically as good as slapping somebody with a baton in some cases, which logically makes absolutely zero sense. Should it knock you over and make you drop items? Yes. But should it knock you down for 6-8 seconds? No.

While this also nerfs the usage of water tanks by simplemobs (such as Terrors) there REALLY shouldnt be a reason that they can use a free 8x8 tile stun for 6-8 seconds.

edit: Slimes (the race) being able to leave stun-like landmines behind them when bleeding is also INSANELY silly from an Antag perspective.

## Changelog
:cl:
tweak: tweaked stun and weaken amounts in TURF_WET_WATER (Stun: 4 -> 2, Weaken: 2 -> 1)
/:cl:

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -81,7 +81,7 @@
 
 			switch(src.wet)
 				if(TURF_WET_WATER)
-					if(!(M.slip("the wet floor", 1, 0, tilesSlipped = 0, walkSafely = 1)))
+					if(!(M.slip("the wet floor", 1, 1, tilesSlipped = 0, walkSafely = 1)))
 						M.inertia_dir = 0
 						return
 
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -1,3 +1,5 @@
+#define WATER_STUN_TIME 2 //Stun time for water, to edit/find easier
+#define WATER_WEAKEN_TIME 1 //Weaken time for water, to edit/find easier
 /turf/simulated
 	name = "station"
 	var/wet = 0
@@ -81,7 +83,7 @@
 
 			switch(src.wet)
 				if(TURF_WET_WATER)
-					if(!(M.slip("the wet floor", 2, 1, tilesSlipped = 0, walkSafely = 1)))
+					if(!(M.slip("the wet floor", WATER_STUN_TIME, WATER_WEAKEN_TIME, tilesSlipped = 0, walkSafely = 1)))
 						M.inertia_dir = 0
 						return
 
@@ -107,3 +109,6 @@
 	queue_smooth_neighbors(src)
 
 /turf/simulated/proc/is_shielded()
+
+#undef WATER_STUN_TIME
+#undef WATER_WEAKEN_TIME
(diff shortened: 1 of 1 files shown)


Mitchs98:
You should include specifically in the changelog how much you tweaked/changed the values imo. Otherwise though, nice change.

Dashsora:
This is FOR SURE a good change - but there's one thing i'd like to propose : What if all slips displace you one tile, and lube/acid naners three tiles, for the short stun duration, BUT if you hit an object like a wall/ closed locker/machine/table/etc, you get stunned for longer ?

It could check how many tiles you should've displaced, and how many tiles you actually displaced, and then handle the stun accordingly (More remaining tiles = longer stun, so if you slip on lube and immediately bonk your head on a wall you get stunned for longer than if it's on the end of the slide, water potential bonk stun would always be the same since it displaces only one tile)

matttheficus:
> You should include specifically in the changelog how much you tweaked/changed the values imo. Otherwise though, nice change.

fixed

Necaladun:
I'll miss being able to robust sec with just a bucket of water in perma but this is a needed change imo.

MRgaty:
I think this will increment the amount of people that use space lube.

matttheficus:
> I'll miss being able to robust sec with just a bucket of water in perma but this is a needed change imo.

you still CAN, it just wont stun for as long, so make sure to yoink their dropped weapons and stun them ASAP!

Kyep (approved):
This doesn't need my approval, as it already has a head voting for it.
Just for the record though, I approve too.

variableundefined (changes requested):
Since this is an important variable that might be tweaked in the future, do you mind making a constant say, called 

#def WATER_STUN_TIME 
and 
#def WATER_WEAKEN TIME

And refer to that constant in the lines, and then #undef them at the end?

matttheficus:
> Since this is an important variable that might be tweaked in the future, do you mind making a constant say, called
> 
> #def WATER_STUN_TIME
> and
> #def WATER_WEAKEN TIME
> 
> And refer to that constant in the lines, and then #undef them at the end?

Is it worth separating the values twice? The slip effect (at least the one changed here) is already only being applied to TURF_WET_WATER, implying that it only changes

variableundefined:
@matttheficus 
Yes, magic number should be avoided. 

While this number is not reused anywhere, a meaningful name is going to be helpful for a maintainer, future contributor or you in the future to recall the **intent** behind the code. It also makes it easier for a layman to understand what is going on. 

If, for some reason I want to change the code for stun, I go into this file and I can see WATER_STUN_TIME in all caps at the top. I can search for the variable and quickly find the code responsible and debug it. 

Without it, I will have to search thoroughly for the code in this file that affects the water slip before I know what to do, and if a subtle typo is made, I will have a hard time understanding whether that is the intent in the first place.

variableundefined:
@matttheficus 
Please enact requested changes within 4 days

matttheficus:
life slammed stuff to a halt, changes made (hopefully properly, havent worked with defines at all with DM, used to constants instead)

variableundefined (approved):
@matttheficus You did it correctly

PR #14519, opened 2020-10-02
Makes the speed gene and other ignoreslowdown buffs take speedups into account

### LINKED ISSUE #14090: Legion cores slow you down when you have leg stimulus
**Issue Description**:
When the legion core slowdown removal buff is active it decreases the users max speed if they have leg stimulus

I expected the buff would remove slowdown from injuries and leave my bonus speed from leg stimulus.

**What happened instead**:
When I used a legion core it removed the slowdown from pain but also removes the bonus speed from leg stimulus.

Legion cores are super important to fighting megafauna on lavaland and this bug basically makes leg stimulus useless against bosses like ash drakes and colossus. It also means that you can't get leg stimulus and meth to fight bubblegum.

1. Get leg stimulus from the DNA vault. 2. Get and use a legion core.

I'm not sure but this happened to me a couple days ago

**Extra information**:
I used a stabilized legion core. Not sure if this affects it

farie82:
## What Does This PR Do
Makes the `IGNORESLOWDOWN` flag and `RUN` mtuation only ignore slowdown and take speedups from `speed_mod` and the `GOTTAGOFAST` flag into account.

Also makes clothing such as the cult flagellant's robes work with the flag and mutation

Adds the IGNORE_SPEED_CHANGES status flag for the legion cores which makes the affected player ignore all speed changes. Be it good or bad. This doesn't change anything functionally from the previous workings

Fixes #14090

## Why It's Good For The Game
The speed gene now actually makes you faster if you take meth or such

## Changelog
:cl:
tweak: Makes the genetics super speed mutation not ignore speedups
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/_species.dm
+++ b/code/modules/mob/living/carbon/human/species/_species.dm
@@ -222,6 +222,8 @@
 
 /datum/species/proc/movement_delay(mob/living/carbon/human/H)
 	. = 0	//We start at 0.
+	if(H.status_flags & IGNORE_SPEED_CHANGES)
+		return .
 
 	if(has_gravity(H))
 		if(H.status_flags & GOTTAGOFAST)
--- a/code/modules/mob/living/carbon/human/species/_species.dm
+++ b/code/modules/mob/living/carbon/human/species/_species.dm
@@ -221,20 +221,24 @@
 
 /datum/species/proc/movement_delay(mob/living/carbon/human/H)
 	. = 0	//We start at 0.
-	var/ignoreslow = FALSE
 
+	if(has_gravity(H))
+		if(H.status_flags & GOTTAGOFAST)
+			. -= 1
 
-	if((H.status_flags & IGNORESLOWDOWN) || (RUN in H.mutations))
-		ignoreslow = TRUE
+		var/ignoreslow = FALSE
+		if((H.status_flags & IGNORESLOWDOWN) || (RUN in H.mutations))
+			ignoreslow = TRUE
 
-	if(ignoreslow)
-		if(speed_mod < 0)	// Apply the speed_mod if it's negative. Meaning it's a speedup
-			. = speed_mod
-	else if(has_gravity(H))
+		if(ignoreslow)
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/species/_species.dm
```diff
@@ -221,22 +221,24 @@
...
+		var/ignoreslow = FALSE
+		if((H.status_flags & IGNORESLOWDOWN) || (RUN in H.mutations))
+			ignoreslow = TRUE
 
-	if(has_gravity(H))
-		gravity = 1
+		if(ignoreslow)
+			if(speed_mod < 0)	// Apply the speed_mod if it's negative. Meaning it's a speedup
+				. += speed_mod
+			return .
```
SteelSlayer, line 236: This section can be made a bit more efficient by removing the use of `ignoreslow`:
```dm
		if((H.status_flags & IGNORESLOWDOWN) || (RUN in H.mutations))
			if(speed_mod < 0)	// Apply the speed_mod if it's negative. Meaning it's a speedup
				. += speed_mod
			return .
```

farie82: Yeah that will remove the var. I did think of other cases though where worn equipment speeds you up. So I'll need to factor those in as well. Which most likely means I'll still be using this var

SteelSlayer: Fair enough

Kyep (commented):
This may be a fix, but I'm not sure how I feel about buffing legion cores.
Lavaland loot is pretty OP in general, and this is buffing one of the most common lavaland drops which is already quite strong.

Necaladun:
As above - it maybe a fix but it seems a good downside for legion cores, IMO. 

The genetics one for sure I think is fine.

hal9000PR:
Personally I think legion cores shouldnt be instant healing, they should heal over 5 or so seconds. Meaning you can't just face tank buckshot if you have enough cores.

farie82:
Changed to make legion cores behave the same by giving them a new status flag which just ignores all speed changes

AffectedArc07 (approved):

PR #14521, opened 2020-10-03
Sling thralling now respects mindshields

TDSSS:
## What Does This PR Do
It was a resulting proposal from the survey meeting and something I've felt was kind of needed for a while now. 
This PR makes it so a sling that tries to enthrall someone with a mindshield will fail. 

Edit: the mindshield now just completely blocks the thralling with a message explaining why to the sling.

## Why It's Good For The Game
Slings have a ridiculous snowballing component. Every sec officer they get not only removes one of their enemies but gives them one very potent ally that can more or less freely arrest more victims for them to thrall.

## Changelog
:cl:
tweak: Mindshields now protect from Shadowling thralling
/:cl:

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -252,12 +252,11 @@
 				user.visible_message("<span class='warning'>[user]'s palms flare a bright red against [target]'s temples!</span>")
 				to_chat(target, "<span class='danger'>A terrible red light floods your mind. You collapse as conscious thought is wiped away.</span>")
 				target.Weaken(12)
-				sleep(2 SECONDS)
 			if(3)
 				to_chat(user, "<span class='notice'>You begin planting the tumor that will control the new thrall...</span>")
 				user.visible_message("<span class='warning'>A strange energy passes from [user]'s hands into [target]'s head!</span>")
 				to_chat(target, "<span class='boldannounce'>You feel your memories twisting, morphing. A sense of horror dominates your mind.</span>")
-		if(!do_mob(user, target, 7 SECONDS)) //around 21 seconds total for enthralling, 31 for someone with a mindshield implant
+		if(!do_mob(user, target, 9 SECONDS)) //around 21 seconds total for enthralling
 			to_chat(user, "<span class='warning'>The enthralling has been interrupted - your target's mind returns to its previous state.</span>")
 			to_chat(target, "<span class='userdanger'>You wrest yourself away from [user]'s hands and compose yourself</span>")
 			enthralling = FALSE
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -249,12 +249,12 @@
 				user.visible_message("<span class='warning'>[user]'s palms flare a bright red against [target]'s temples!</span>")
 				to_chat(target, "<span class='danger'>A terrible red light floods your mind. You collapse as conscious thought is wiped away.</span>")
 				target.Weaken(12)
-				sleep(20)
+				sleep(2 SECONDS)
 				if(ismindshielded(target))
 					to_chat(user, "<span class='notice'>They have a mindshield implant. You try to overwhelm it!</span>")
 					user.visible_message("<span class='warning'>[user] pauses, then dips [user.p_their()] head in concentration!</span>")
 					to_chat(target, "<span class='boldannounce'>Your mindshield implant becomes hot as it comes under attack!</span>")
-					sleep(100) //10 seconds - not spawn() so the enthralling takes longer
+					sleep(10 SECONDS) //10 seconds - not spawn() so the enthralling takes longer
 					to_chat(user, "<span class='notice'>The nanobots composing the mindshield implant put up a valiant fight but they are about to lose...</span>")
 					to_chat(user, "<span class='boldannounce'>Rather than be disabled, the mindshield superheats, frying your victim's brain!</span>")
 					to_chat(target, "<span class='boldannounce'>You feel a burning, horrible pain in your head!</span>")
@@ -268,7 +268,7 @@
 				to_chat(user, "<span class='notice'>You begin planting the tumor that will control the new thrall...</span>")
 				user.visible_message("<span class='warning'>A strange energy passes from [user]'s hands into [target]'s head!</span>")
 				to_chat(target, "<span class='boldannounce'>You feel your memories twisting, morphing. A sense of horror dominates your mind.</span>")
-		if(!do_mob(user, target, 70)) //around 21 seconds total for enthralling, 31 for someone with a mindshield implant
+		if(!do_mob(user, target, 7 SECONDS)) //around 21 seconds total for enthralling, 31 for someone with a mindshield implant
 			to_chat(user, "<span class='warning'>The enthralling has been interrupted - your target's mind returns to its previous state.</span>")
 			to_chat(target, "<span class='userdanger'>You wrest yourself away from [user]'s hands and compose yourself</span>")
 			enthralling = FALSE
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -235,6 +235,9 @@
 	if(!(ling.mind in SSticker.mode.shadows))
 		return
 	var/mob/living/carbon/human/target = targets[1]
+	if(ismindshielded(target))
+		to_chat(user, "<span class='danger'>This target has a mindshield, blocking your powers! You cannot thrall it!</span>")
+		return
 	enthralling = TRUE
 	to_chat(user, "<span class='danger'>This target is valid. You begin the enthralling.</span>")
 	to_chat(target, "<span class='userdanger'>[user] stares at you. You feel your head begin to pulse.</span>")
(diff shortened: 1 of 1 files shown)


Necaladun:
I'd like a 1 week or so testmerge on this, then decide if it's too much of a buff to sec and we look at some inbetween like increased resistance/able to call for help/etc.

AffectedArc07:
A weeklong testmerge for this wouldn't be viable since (as far as I know, @Kyep can confirm or deny this), testmerges lock the server in that state and newly merged PRs wouldn't take effect. 

A valid option is to merge this then consider a revert after a week however

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -249,23 +249,26 @@
...
-					for(var/obj/item/implant/mindshield/L in target)
+					sleep(10 SECONDS) //10 seconds - not spawn() so the enthralling takes longer
+					to_chat(user, "<span class='notice'>The nanobots composing the mindshield implant put up a valiant fight but they are about to lose...</span>")
+					to_chat(user, "<span class='boldannounce'>Rather than be disabled, the mindshield superheats, frying your victim's brain!</span>")
+					to_chat(target, "<span class='boldannounce'>You feel a burning, horrible pain in your head!</span>")
+					for(var/obj/item/implant/mindshield/L in target)	//it self-destructs to kill you, after all
 						if(L && L.implanted)
 							qdel(L)
-					to_chat(target, "<span class='boldannounce'>Your mental protection implant unexpectedly falters, dims, dies.</span>")
+					target.adjustBrainLoss(120, TRUE, FALSE)
```
AffectedArc07, line 264: Im really not sure that giving the target **120** brain damage is a good solution here

TDSSS: 120 is the brain damage needed for instant death. If I did less damage than that they either wouldn't die or (if I manually call death() after applying it) they could simply be defibbed back alive.

I considered doing a lot of burn damage to the head instead but this runs into the same issue. I'd have to either make it a ridiculous number (like 500 burn or whatever) to more or less guarantee death or I'd have to manually call death(), at which point you just slap some burn patches on and defib.

Someone also brought up the option of exploding their head, but that would make them harder to clone as well and I didn't want found, dead sec officers to be out for longer than the normal clone timer.

If you have a good idea for a solution, I'm all ears.

kenken244:
I think a better solution would be to just have mindshields block thralling entirely. If they want to kill the officer they can do it the old fashioned way. There's no need for special mechanics to kill them when they have to be helpless anyway

code/game/gamemodes/shadowling/shadowling_abilities.dm
AffectedArc07: Currently, it takes 19 seconds without the target moving to kill them, which is shorter than the regular enthrall time, and a lot more potent (Slings can now wipeout security officers in a smaller time than enthralling, then stash the body in the depths of maintenance, greatly hindering security)

Im not sure death is even a good solution here, maybe just denying the enthrall in the same way that cult and vamp thralls do.

AzuleUtama:
Blocking thralling entirely with them would just lead to security ordering and handing out mindshields to as many people as possible so slings have more trouble ascending. Having a deterrent against rushing to grab them by making it so you die if a sling attempts to thrall you is a better option IMO.

AffectedArc07:
> Blocking thralling entirely with them would just lead to security ordering and handing out mindshields to as many people as possible so slings have more trouble ascending. Having a deterrent against rushing to grab them by making it so you die if a sling attempts to thrall you is a better option IMO.

Security doesnt rapidly mindshield the entire crew on cult rounds, and they have to get atleast 9 people to summon their god, so I dont see that as an issue.

AzuleUtama:
> > Blocking thralling entirely with them would just lead to security ordering and handing out mindshields to as many people as possible so slings have more trouble ascending. Having a deterrent against rushing to grab them by making it so you die if a sling attempts to thrall you is a better option IMO.
> 
> Security doesnt rapidly mindshield the entire crew on cult rounds, and they have to get atleast 9 people to summon their god, so I dont see that as an issue.

Mindshielded converts for cult are instead sacrificed and put into a soul stone IIRC, so it gives cultists a potential construct instead which can still invoke runes - probably why experienced sec. don't bother with mindshields there.

kenken244:
Mass mindshields are against sop afaik, and there's isn't much reason for crew to get one, as being thralled is preferable to dying for most people.

code/game/gamemodes/shadowling/shadowling_abilities.dm
Kyep: Yeah, probably best to just make mindshields block thralling, instead.

Muhdi345:
If this has become such a problem, why not just make a death alarm like announcement to the crew such and so has been thralled? The main problem is normal crew have no way of knowing which officer can be trustworthy since they nigh always wear their gas masks anyway.

TDSSS:
Just blocking the thralling is the boring, practical approach. But if people want that, then sure. Will change soon.

TDSSS:
PR updated to completely block thralling now.

TheMaveran:
I don’t think this will do anything but make a gross meta of sec handing out mass amounts of mindshields, which would then result in a meta where slings carry around surgery tools, or another ooc rule is made around mindshields not being handed during shadowling rounds. Not sure this is going to make the game mode more enjoyable.

Necaladun:
Mass-mindshields haven't been a problem in rev or cult very often, I don't see it being a problem here suddenly.

SteelSlayer:
>Edit: oh, or sec officers will just get killed and thrown in a locker with all their gear off and left for the rest of the round. Sounds very fun to be security.

Hard agree on this. This is going to make people who play mindshielded roles hate Shadowlings so much, because whats going to happen? They're going to get murdered and tossed into space, or worse. People are always talking about how being permakilled and removed from the round sucks. I'm surprised this is getting thumbs up at all to be honest.

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -249,23 +252,12 @@
...
-					user.visible_message("<span class='warning'>[user] pauses, then dips [user.p_their()] head in concentration!</span>")
-					to_chat(target, "<span class='boldannounce'>Your mindshield implant becomes hot as it comes under attack!</span>")
-					sleep(100) //10 seconds - not spawn() so the enthralling takes longer
-					to_chat(user, "<span class='notice'>The nanobots composing the mindshield implant have been rendered inert. Now to continue.</span>")
-					user.visible_message("<span class='warning'>[user] relaxes again.</span>")
-					for(var/obj/item/implant/mindshield/L in target)
-						if(L && L.implanted)
-							qdel(L)
-					to_chat(target, "<span class='boldannounce'>Your mental protection implant unexpectedly falters, dims, dies.</span>")
+				sleep(2 SECONDS)
```
SteelSlayer, line 255: I think you should remove this sleep, now that nothing needs to happen after it, and because sleeps are evil. If you feel the need to compensate on the overall time it takes, you could probably change:
```dm
if(!do_mob(user, target, 7 SECONDS))
```
To
```dm
if(!do_mob(user, target, 7.7 SECONDS))
```
Or something like that. Making the total time to thrall ~23.1 seconds, just 0.1 over what it was before.

Kyep:
Alternatively:
- Slings might start avoiding sec due to changed risk/reward, which is potentially interesting cat-and-mouse
- Dead sec might respawn as ERT, which are essentially a strictly better version of sec, with better gear
- In the meantime, slings won't instantly snowball out of control due to being able to thrall trusted people with stun weapons

code/game/gamemodes/shadowling/shadowling_abilities.dm
AffectedArc07: Siding with steel. Remove this sleep.

AffectedArc07 (changes requested):

taukausanake:
> they die often

Like a red shirt should?

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -252,12 +252,11 @@
 				user.visible_message("<span class='warning'>[user]'s palms flare a bright red against [target]'s temples!</span>")
 				to_chat(target, "<span class='danger'>A terrible red light floods your mind. You collapse as conscious thought is wiped away.</span>")
 				target.Weaken(12)
-				sleep(2 SECONDS)
 			if(3)
 				to_chat(user, "<span class='notice'>You begin planting the tumor that will control the new thrall...</span>")
 				user.visible_message("<span class='warning'>A strange energy passes from [user]'s hands into [target]'s head!</span>")
 				to_chat(target, "<span class='boldannounce'>You feel your memories twisting, morphing. A sense of horror dominates your mind.</span>")
-		if(!do_mob(user, target, 7 SECONDS)) //around 21 seconds total for enthralling, 31 for someone with a mindshield implant
+		if(!do_mob(user, target, 9 SECONDS)) //around 21 seconds total for enthralling
```
AffectedArc07, line 259:
```suggestion
		if(!do_mob(user, target, 9 SECONDS)) //around 27 seconds total for enthralling
```
Update the comments as well. I am also questioning whether the total enthrall time for non-mindshielded crew should be raised by a full 6 seconds.

AffectedArc07 (changes requested):

code/game/gamemodes/shadowling/shadowling_abilities.dm
TDSSS: My bad, I misread and didn't notice the timer is used 3 times, I meant for the total time to stay the same. Should be the ~23 seconds it was before now.

AffectedArc07 (approved):

thatdanguy23:
Having observed a round with this in play, shadowlings have been pretty severely gimped by this PR. It sucks that they snowball when they thrall a sec officer, BUT it sucks equally as much that they have to resort to killing officers in order to succeed - something that is either not fun for security or not fun for slings.

I thought we were trying to encourage less murderbone, but no, it seems we're just reinforcing the idea that team deathmatch is cool. Maybe change it so that breaking the mindshield takes longer. Something other than this.

AffectedArc07:
> Having observed a round with this in play, shadowlings have been pretty severely gimped by this PR. It sucks that they snowball when they thrall a sec officer, BUT it sucks equally as much that they have to resort to killing officers in order to succeed - something that is either not fun for security or not fun for slings.
> 
> I thought we were trying to encourage less murderbone, but no, it seems we're just reinforcing the idea that team deathmatch is cool. Maybe change it so that breaking the mindshield takes longer. Something other than this.

That round had only 2 shadowlings, and only 2 thralls by the end. Security was overcompetant and there were 8 cyborgs. It was not a valid test case.

thatdanguy23:
> That round had only 2 shadowlings, and only 2 thralls by the end. Security was overcompetant and there were 8 cyborgs. It was not a valid test case.

It's certainly indicative of an issue with lower pop rounds, though. We can't base everything on a high pop scenario, can we? That's only half of the playerbase. Still, sure - more sling rounds needed.

Spacemanspark:
I wish I'd seen this PR.
I feel this kind of puts slings in a very similar state like revolution does. Slings will eventually have to resort to murderboning once security players realize that they can mass implant everyone. Cargo will be under siege most rounds.

We'll see where time takes us, however.

Jungling:
Revert asap. Promotes murderboning. Prevents RPing.

PR #14522, opened 2020-10-03
Makes laser tags have 0 damage and makes the field generators take nodamage into consideration

farie82:
## What Does This PR Do
Fixes laser tags being able to power field generators.

I changed the damage in the laser tags to 0 to make it actually make sense

## Why It's Good For The Game
Bug fix

## Changelog
:cl:
fix: Laser tags have 0 damage now.
fix: Field generators now don't accept power if the projectile has the nodamage flag set to true
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -85,6 +85,7 @@
 	icon_state = "omnilaser"
 	hitsound = 'sound/weapons/tap.ogg'
 	nodamage = 1
+	damage = 0
 	damage_type = STAMINA
 	flag = "laser"
 	var/suit_types = list(/obj/item/clothing/suit/redtag, /obj/item/clothing/suit/bluetag)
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/projectile/beams.dm
```diff
@@ -85,6 +85,7 @@
 	icon_state = "omnilaser"
 	hitsound = 'sound/weapons/tap.ogg'
 	nodamage = 1
+	damage = 0
 	damage_type = STAMINA
```
AffectedArc07, line 89: Does this still do stamina damage if you specify damage to 0?

Necaladun:
That was fucking hilarious.
But yeah uhm. No. 

Also check it doesn't break laser tag guns doing no damage to people with the suits?

code/modules/projectiles/projectile/beams.dm
farie82: nodamage will override it all so it does nothing
image

farie82:
> That was fucking hilarious.
> But yeah uhm. No.
> 
> Also check it doesn't break laser tag guns doing no damage to people with the suits?

Yep yep still works. That's handled in the on_hit for the laser tag guns. Which doesn't care about anything else other than the suit you're wearing

AffectedArc07 (approved):

PR #14529, opened 2020-10-03
Morgue destroy fix

Kiyahitayika:
## What Does This PR Do
This PR stops the destruction of a Morgue/Crematorium structure from deleting all of its contents, including any living mobs, and instead ejects them onto the same time the structure previously existed prior to its destruction.

Ensures destruction of the Morgue/Crematorium tray as well if the tray is open.

## Why It's Good For The Game
This is a bug fix. Destroying a Morgue/Crematorium structure deletes any mob inside, whether they are dead or alive. If their mob is deleted, a player can no longer be returned to the game. Certain items should be protected from destruction this way as well.

## Images of changes
Before the Fix:
https://imgur.com/a/Vfsf9XC

After the Fix (walls are not necessary):
https://imgur.com/a/sNfywBy

## Changelog
:cl:
fix: Morgue/Crematorium structure destruction no longer deletes anything inside
/:cl:

### DIFF
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -155,8 +155,12 @@
 	return
 
 /obj/structure/morgue/Destroy()
-	QDEL_NULL(connected)
+	var/turf/T = src.loc
+	if(!connected)
+		for(var/atom/movable/A as mob|obj in src)
+			A.forceMove(T)
 	return ..()
+
 
 /obj/structure/morgue/container_resist(var/mob/living/L)
 	var/mob/living/carbon/CM = L
@@ -389,7 +393,10 @@
 	return
 
 /obj/structure/crematorium/Destroy()
-	QDEL_NULL(connected)
+	var/turf/T = src.loc
+	if(!connected)
+		for(var/atom/movable/A as mob|obj in src)
+			A.forceMove(T)
 	return ..()
 
 /obj/structure/crematorium/container_resist(var/mob/living/L)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/morgue.dm
```diff
@@ -155,7 +155,12 @@
 	return
 
 /obj/structure/morgue/Destroy()
-	QDEL_NULL(connected)
+	if(!connected)
+		var/turf/T = src.loc
+		for(var/atom/movable/A as mob|obj in src)
+			A.forceMove(T)
```
AffectedArc07, line 161:
```suggestion
		var/turf/T = loc
		for(var/atom/movable/A in src)
			A.forceMove(T)
```

code/game/objects/structures/morgue.dm
```diff
@@ -389,7 +394,12 @@
 	return
 
 /obj/structure/crematorium/Destroy()
-	QDEL_NULL(connected)
+	if(!connected)
+		var/turf/T = src.loc
+		for(var/atom/movable/A as mob|obj in src)
+			A.forceMove(T)
```
AffectedArc07, line 400:
```suggestion
		var/turf/T = loc
		for(var/atom/movable/A in src)
			A.forceMove(T)
```

AffectedArc07 (changes requested):

code/game/objects/structures/morgue.dm
Kiyahitayika: Mmm. I trust your suggestions and I'm going to test them before committing them.

Kiyahitayika:
@AffectedArc07 Those changes can also be made elsewhere throughout the file

AffectedArc07 (approved):
Works

PR #14531, opened 2020-10-04
Blurry eyes blur eyes

FlattestGuitar:
2dQKKrurWp

Use a BYOND filter to blur eyes. It only applies to floor and game plane masters. I also put some filter actions behind a proc that will animate their removal (in most cases) properly. 

:cl: Flatty
add: Blurry eyes are now blurrier
/:cl:

### DIFF
--- a/code/_onclick/hud/plane_master.dm
+++ b/code/_onclick/hud/plane_master.dm
@@ -5,6 +5,7 @@
 	blend_mode = BLEND_OVERLAY
 	var/show_alpha = 255
 	var/hide_alpha = 0
+	var/current_filters = list()
 
 /obj/screen/plane_master/proc/Show(override)
 	alpha = override || show_alpha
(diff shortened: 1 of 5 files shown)


code/_onclick/hud/plane_master.dm
```diff
@@ -38,9 +46,8 @@
 	blend_mode = BLEND_OVERLAY
 
 /obj/screen/plane_master/game_world/backdrop(mob/mymob)
-	clear_filters()
```
FlattestGuitar, line 41: I can't figure out why filters are being rebuilt in this proc. I don't like this at all, because that would mean that we'd have to somehow keep track of every filter that should be applied to this plane master from this location, which just doesn't feel right. 

Maybe this should be removed altogether and handled elsewhere?

code/_onclick/hud/plane_master.dm
```diff
@@ -5,21 +5,29 @@
...
-/obj/screen/plane_master/proc/outline(_size, _color)
-	filters += filter(type = "outline", size = _size, color = _color)
+/obj/screen/plane_master/proc/add_filter(key, filter)
+	if(current_filters[key])
+		filters -= current_filters[key]
 
-/obj/screen/plane_master/proc/shadow(_size, _border, _offset = 0, _x = 0, _y = 0, _color = "#04080FAA")
-	filters += filter(type = "drop_shadow", x = _x, y = _y, color = _color, size = _size, offset = _offset)
+	filters += filter
+	current_filters[key] = filters[filters.len] //assigning `filter` will not allow it to work later while animating removal
```
FlattestGuitar, line 21: When I try to do something like

```var/filter/filter_to_remove = current_filters[key] //ref saved before the filter was applied
animate(filter_to_remove, ...)
```

BYOND will throw an exception saying that "abstract" filters cannot be animated, which is extremely odd. It somehow pulls information on what object this filter belongs to from an access like `filters[filters.len]`, as in the BYOND docs, but what I do here also works, so I have no clue how this is implemented. Magic.

procdrone:
Very cool, but will it have actual use? If so, where?

FlattestGuitar:
There's a massive amount of eye blur around.

The most obvious examples are having damaged eyes or going into crit.

Just up until now this was done with a simple full screen "white clouds" overlay. Since pretty much every other server now uses this method I figured there wouldn't be much opposition to this.

procdrone:
Im not opposed by all means, just wanted to understand where this effect was intended to be used.

Spacemanspark:
Now we just need to muffle sounds while in crit.

Dumbdumn5 (approved):
This physically pains my *real* eyes. 10/10.

AffectedArc07 (approved):

PR #14535, opened 2020-10-05
Nerfs the bluespace wormhole projector

Qwertytoforty:
## What Does This PR Do

Nerfs the bluespace wormhole projector (portal gun) by giving it a 1 second fire delay, and 10 shots. However, allows it to self charge one shot every 10 seconds.

## Why It's Good For The Game

The bluespace wormhole is a fun tool. A nice portal reference, can be used cleverly. The issue is, people have been abusing it heavily in 2 ways. One: Due to its infinite ammo and low fire rate, you can spam fire it everywhere without it ever running out. This is problematic, as you can not reach a person spam fireing this gun, as anyone hit by one of its shots is randomly teleported in a 6 tile radius, often into another room, or space nearby, preventing them from reaching the person with a gun. 
Two: If you enter the portal, you get teleported to the other portal. Simple. However, if you have a bag of holding, you get teleported a large distance away from where you should end up. This may seem like a downside, but it is not. If you willingly enter a portal while security is chasing you, you will teleport a large distance away from security, with them having no way of telling where you have gone, and no way to follow, even if they have a bag of holding on as well, they go somewhere else.

This PR attempts to fix this, by lowering fire rate and giving it a power cause, so it can not be spammed, and if you do want to use this in combat, you have to aim carefully.
## Changelog
:cl:
tweak: Gave the bluespace wormhole projector a 1 second fire delay, and changes it from infinite ammo to 10 shots, with a 1 shot per second recharge.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -201,7 +201,7 @@
 	icon_state = "wormhole_projector1"
 	origin_tech = "combat=4;bluespace=6;plasmatech=4;engineering=4"
 	charge_delay = 5
-	selfcharge = 1
+	selfcharge = TRUE
 	var/obj/effect/portal/blue
 	var/obj/effect/portal/orange
 
--- a/code/modules/projectiles/guns/energy/special.dm
+++ b/code/modules/projectiles/guns/energy/special.dm
@@ -200,6 +200,8 @@
 	item_state = null
 	icon_state = "wormhole_projector1"
 	origin_tech = "combat=4;bluespace=6;plasmatech=4;engineering=4"
+	charge_delay = 5
+	selfcharge = 1
 	var/obj/effect/portal/blue
 	var/obj/effect/portal/orange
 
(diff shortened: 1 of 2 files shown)


AffectedArc07:
I recommend rebasing to get rid of that nanomap commit, and as a practice, always start a branch from `ParadiseSS13/master`, not your fork master.

Kyep (approved):

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -200,6 +200,8 @@
 	item_state = null
 	icon_state = "wormhole_projector1"
 	origin_tech = "combat=4;bluespace=6;plasmatech=4;engineering=4"
+	charge_delay = 5
+	selfcharge = 1
```
AffectedArc07, line 204:
```suggestion
	selfcharge = TRUE
```

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #14536, opened 2020-10-05
Close storage UI when item becomes inaccessible

nicetoolbox:
## What Does This PR Do
Adds a proc called hide_from_all on obj/item/storage to hide the UI from all users who opened it. It should be called in places where an action makes objects unavailable. I added it to throwing, lockers closing, safes locking, and storage insertion, and disposal chutes. Any other places to add anyone can think of?

## Why It's Good For The Game
It's confusing to be able to see storage in certain situations. Especially when the storage already has checks that you can take items, but clicking them does nothing. For example, if you unlock and open the storage of the secure safe in the captains office, then lock it, you can still see the contents even though you can't interact with it. With this change the storage window would close once you lock the safe.

## Changelog
:cl:
fix: Some storage objects hide their UI after becoming inaccessible
/:cl:

### DIFF
--- a/code/modules/recycling/disposal.dm
+++ b/code/modules/recycling/disposal.dm
@@ -103,8 +103,8 @@
 		to_chat(user, "You can't place that item inside the disposal unit.")
 		return
 
-	if(istype(I, /obj/item/storage))
-		var/obj/item/storage/S = I
+	var/obj/item/storage/S = I
+	if(istype(S))
 		if((S.allow_quick_empty || S.allow_quick_gather) && S.contents.len)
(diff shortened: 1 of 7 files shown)


code/modules/recycling/disposal.dm
```diff
@@ -113,6 +113,8 @@
 			S.update_icon() // For content-sensitive icons
 			update()
 			return
+		else
+			S.hide_from_all()
```
nicetoolbox, line 117: oops move this to140

variableundefined (approved):

PR #14541, opened 2020-10-05
Gives admins a way to smoothly end a round

Qwertytoforty:
## What Does This PR Do

This PR gives admins a way to smoothly end the round, without having to call shuttle, or nuke the station. Shows up in server tab, requires SERVER rights, same as restarting the server.

## Why It's Good For The Game

Sometimes a nuke is not the right ending. Whether the theme of an event means the station is destroyed in another way, or if nuking the station normally is failing, or if the wizard is dead but the round is not ending, it would be nice to have a button to end the round and bring up objectives. This PR adds that.

Like when the station is nuked or the AI doomsdays, **If you are alive when the round is force ended, you count as escaping greentext wise.**

## Images of changes

image

## Changelog
:cl:
add: Allows admins to end the round instantly and bring up objectives with the press of a button
/:cl:

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -614,7 +614,7 @@ GLOBAL_VAR_INIT(nologevent, 0)
 /datum/admins/proc/smoothend()
 	set category = "Server"
 	set name = "Annouce game over"
-	set desc = "Instantly ends round and brings up objectives like shadowlings or slings dying."
+	set desc = "Instantly ends round and brings up objectives like shadowlings or wizards dying."
 
 	if(!check_rights(R_SERVER))
 		return
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -624,7 +624,7 @@ GLOBAL_VAR_INIT(nologevent, 0)
 	if(SSticker.force_ending)
 		return
 	message_admins("[key_name_admin(usr)] has admin ended the round with message: '[input]'")
-	log_admin("[key_name(usr)] has admin ended the round  with message: '[input]'")
+	log_admin("[key_name(usr)] has admin ended the round with message: '[input]'")
 	SSticker.force_ending = TRUE
 	to_chat(world, "<span class='warning'><big><b>[input]</b></big></span>")
 	feedback_add_details("admin_verb", "END") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -611,23 +611,24 @@ GLOBAL_VAR_INIT(nologevent, 0)
 	feedback_add_details("admin_verb","R") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 	world.Reboot("Initiated by [usr.client.holder.fakekey ? "Admin" : usr.key].", "end_error", "admin reboot - by [usr.key] [usr.client.holder.fakekey ? "(stealth)" : ""]", delay)
 
-/datum/admins/proc/smoothend()
+/datum/admins/proc/end_round()
 	set category = "Server"
-	set name = "Annouce game over"
-	set desc = "Instantly ends round and brings up objectives like shadowlings or wizards dying."
-
+	set name = "End Round"
+	set desc = "Instantly ends the round and brings up the scoreboard, like shadowlings or wizards dying."
 	if(!check_rights(R_SERVER))
 		return
+	var/input = sanitize(copytext(input(usr, "What text should players see announcing the round end? Input nothing to cancel.", "Specify Announcement Text", "Shift Has Ended!"), 1, MAX_MESSAGE_LEN))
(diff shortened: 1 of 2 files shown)


TDSSS:
Good idea in general, but I'd much prefer a less immersion-breaking way to end the round with an option for inputing a custom message like "Clown major victory" or whatever.

KasparoVy:
Dope idea, totally agree with being able to do a custom message

code/modules/admin/admin.dm
```diff
@@ -611,6 +611,24 @@ GLOBAL_VAR_INIT(nologevent, 0)
...
+	if(alert(usr, "Are you sure you want to admin end the round and bring up objectives?", "End the game", "Yes", "No") != "Yes")
+		return
+
+	message_admins("[key_name_admin(usr)] has admin ended the round.")
+	log_admin("[key_name(usr)] has admin ended the round")
+	SSticker.force_ending = 1
+	to_chat(world, "<span class='warning'><FONT size = 3><b>The administrators have declared the round over and have ended the round!</b></FONT></span>")
+	feedback_add_details("admin_verb","S") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
+	feedback_set_details("round_end_result","Admin ended")
+
```
Kyep, line 631:
```suggestion
/datum/admins/proc/smoothend()
	set category = "Server"
	set name = "End Round"
	set desc = "Instantly ends the round and brings up the scoreboard, like shadowlings or wizards dying."

	if(!check_rights(R_SERVER))
		return

	var/input = sanitize(copytext(input(usr, "What text should players see announcing the round end?", "Specify Announcement Text", "Shift Has Ended!"), 1, MAX_MESSAGE_LEN))	
	
	if(!input)
		return

	if(SSticker.force_ending)
		return

	message_admins("[key_name_admin(usr)] has admin ended the round with message: '[input]'")
	log_admin("[key_name(usr)] has admin ended the round  with message: '[input]'")
	SSticker.force_ending = 1
	to_chat(world, "<span class='warning'><FONT size = 3><b>[input]</b></FONT></span>")
	feedback_add_details("admin_verb", "S") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
	feedback_set_details("round_end_result", "Admin ended")

```
Reasoning:
1) Should be called "End Round".
2) Should prompt for a reason and display that reason to players, so we can declare e.g. "(event antag) Major Victory" as the reason.
3) Should log the reason as well.
4) Cancelling should be done by clicking "cancel" or not entering a reason on the text input box.
5) Proc calls need proper spacing for their arguments.
6) Sensible to have a check after the input to prevent accidental double-use.

Other feedback:
- Why "S" for the feedback_add_details, instead of say "END" or "ER"? Are those in use?

code/modules/admin/admin.dm
```diff
@@ -611,6 +611,24 @@ GLOBAL_VAR_INIT(nologevent, 0)
...
+	if(!check_rights(R_SERVER))
+		return
+
+	if(alert(usr, "Are you sure you want to admin end the round and bring up objectives?", "End the game", "Yes", "No") != "Yes")
+		return
+
+	message_admins("[key_name_admin(usr)] has admin ended the round.")
+	log_admin("[key_name(usr)] has admin ended the round")
+	SSticker.force_ending = 1
```
AffectedArc07, line 627:
```suggestion
	SSticker.force_ending = TRUE
```

code/modules/admin/admin.dm
```diff
@@ -611,6 +611,24 @@ GLOBAL_VAR_INIT(nologevent, 0)
...
+	to_chat(world, "<span class='warning'><FONT size = 3><b>The administrators have declared the round over and have ended the round!</b></FONT></span>")
```
AffectedArc07, line 628:
```suggestion
	to_chat(world, "<span class='warning'><big><b>The administrators have declared the round over and have ended the round!</b></big></span>")
```

AffectedArc07 (changes requested):

code/modules/admin/admin.dm
```diff
@@ -611,6 +611,25 @@ GLOBAL_VAR_INIT(nologevent, 0)
...
+	if(!check_rights(R_SERVER))
+		return
+	var/input = sanitize(copytext(input(usr, "What text should players see announcing the round end? Input nothing to cancel.", "Specify Announcement Text", "Shift Has Ended!"), 1, MAX_MESSAGE_LEN))
+
+	if(!input)
+		return
+	if(SSticker.force_ending)
+		return
+	message_admins("[key_name_admin(usr)] has admin ended the round with message: '[input]'")
+	log_admin("[key_name(usr)] has admin ended the round  with message: '[input]'")
```
AffectedArc07, line 627:
```suggestion
	log_admin("[key_name(usr)] has admin ended the round with message: '[input]'")
```

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #14542, opened 2020-10-05
IPC - Vomit Tweak

### LINKED ISSUE #14381: IPC using wormhole jaunter vomits upon landing
**Issue Description**:

An IPC using the wormhole jaunter, intentionally or falling into a chasm, throws up upon landing wherever it takes them.

Expected there to be a check to not vomit as IPC. Perhaps something else to make the stun last as long.

**What happened instead**:

IPC vomited upon landing.

Immersion breaking. IPC generally don't have the fleshy internals to vomit from.

1: Be IPC.
2: Acquire wormhole jaunter.
3: Use it and walk into the wormhole, or fall into a chasm while it is equipped on your belt slot.

Noticed 19/Sept/2020.

**Extra information**:

Boxstation

Kiyahitayika:
## What Does This PR Do
Tweaks `/mob/living/carbon/proc/vomit()` to do a _**cough** species check **cough**_ and stop IPCs from creating vomit splatters on the floor.

Fixes #14381

@AffectedArc07 Please beat me for "snowflake"

## Why It's Good For The Game
IPCs don't eat food and don't have stomachs. There aren't any particulates for them to vomit.

## Images of changes
https://imgur.com/a/ECRY4Mi
Update: Removed the stun, so that gif doesn't completely apply. 
IPCs _separately_ get stunned from getting too drunk already, so we're good.

## Changelog
:cl:
tweak: Added "snowflake" for IPC vomtting
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -92,6 +92,8 @@
 
 
 /mob/living/carbon/proc/vomit(var/lost_nutrition = 10, var/blood = 0, var/stun = 1, var/distance = 0, var/message = 1)
+	if(ismachineperson(src)) //IPCs do not vomit.
+		return 0
 	if(src.is_muzzled())
 		if(message)
 			to_chat(src, "<span class='warning'>The muzzle prevents you from vomiting!</span>")
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -92,14 +92,14 @@
 
 
 /mob/living/carbon/proc/vomit(var/lost_nutrition = 10, var/blood = 0, var/stun = 1, var/distance = 0, var/message = 1)
+	if(ismachineperson(src)) //IPCs do not vomit particulates
+		return 0
 	if(src.is_muzzled())
 		if(message)
 			to_chat(src, "<span class='warning'>The muzzle prevents you from vomiting!</span>")
 		return 0
 	if(stun)
 		Stun(4)
-	if(ismachineperson(src)) //IPCs do not vomit particulates
-		return 0
 	if(nutrition < 100 && !blood)
 		if(message)
 			visible_message("<span class='warning'>[src] dry heaves!</span>", \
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -98,6 +98,8 @@
 		return 0
 	if(stun)
 		Stun(4)
+	if(ismachineperson(src)) //IPCs do not vomit particulates
```
Mitchs98, line 101: Given they're an IPC it probably shouldn't return the stun in the first place since they can't vomit anyway, the stun itself is there for species that are actually able to vomit. This should probably be at the start of the proc instead to prevent the stun.

I believe IPC booze already has its own stun outside of vomiting(correct me if I'm wrong), but if it doesn't it'd be more appropriate to change something there rather than allow IPC's to suffer any sort of effects from an action they can't perform in the first place.

Kiyahitayika: Good point. We talked on discord. Made change and committed it.

Kyep (approved):

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -92,6 +92,8 @@
 
 
 /mob/living/carbon/proc/vomit(var/lost_nutrition = 10, var/blood = 0, var/stun = 1, var/distance = 0, var/message = 1)
+	if(ismachineperson(src)) //IPCs do not vomit particulates
+		return 0
```
AffectedArc07, line 96:
```suggestion
		return FALSE
```

AffectedArc07 (changes requested):
Please also clean up the rest of this proc, such as removing implied `src` and swapping the 1/0 to TRUE/FALSE

code/modules/mob/living/carbon/carbon.dm
Kiyahitayika: Buh! Will do!

Kiyahitayika (commented):
Changed in Commit

PR #14543, opened 2020-10-05
Boo! Rework

S-CR:
## What Does This PR Do
Slight reworking of the mechanics for the ghost's Boo! ability. Boo! is now a targeted ability rather than an AoE - click once to prime it, click on something in range to spook it. Boo!'s timer has been doubled, ghosts will need to wait two minutes before reaching across the veil again - unless the mistakenly click on something non-boo'able. In that case they'll need to wait 10 seconds before trying again. Ghosts who target objects in areas designated as "haunted" like the morgue or chapel will have 50% reduced cooldown.

The following changes have been made to the Boo! effects:
1. APC spooking has been reworked, it will flicker the BSOD as before and for every item on its powernet it has a 10% chance to cause that item to do its own spook effect.
2. Doors will now fake an "Access denied" flash when spooked.
3. Computer screens and vending machines will flicker when spooked. (This effect is purely visual, it will not stop a person using the computer).
4. Light flickers now act more sporadically - as you'd expect a flickering light to.

All machinery with a flicker effect (Doors, computers, APCs, and lights for now) have an extremely low chance to flicker as they process. Each eligible machine will flicker about twice a round, give or take. The chance is meant to be low, but often enough that you're never 100% sure if a ghost is trying to make contact or if the device flickered all on its own.

## Why It's Good For The Game
More flavorful Boo! effects just in time for October with a dash of less certainty about ghosts interacting with the world. What's not to like?

## Changelog
:cl: Dave
add: New Boo effects for computers, airlocks, and APCs, and vending machines. Try it out!
tweak: Boo is now a targeted effect rather than an AoE.
tweak: Boo's timer has been increased.
tweak: Ghosts get reduced Boo! cooldown for spooking targets in particularly haunted areas.
experiment: Machinery will now occasionally flicker all on its own, very spooky.
/:cl:

### DIFF
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -17,6 +17,8 @@
 	var/light_range_on = 2
 	var/light_power_on = 1
 	var/overlay_layer
+	var/flickering = FALSE
+	var/force_no_power_icon_state = FALSE
 
 /obj/machinery/computer/New()
 	overlay_layer = layer
@@ -36,9 +38,32 @@
 	set_light(0)
 	visible_message("<span class='danger'>[src] grows dim, its screen barely readable.</span>")
 
+/obj/machinery/computer/flicker()
+	var/amount = rand(5, 15)
+
+	if(flickering)
+		return FALSE
+
+	if(stat & (BROKEN|NOPOWER))
+		return FALSE
+
+	flickering = TRUE
+	spawn(0)
+		for(var/i = 0; i < amount; i++)
+			force_no_power_icon_state = TRUE
+			update_icon()
+			sleep(rand(1, 3))
+
+			force_no_power_icon_state = FALSE
+			update_icon()
+			sleep(rand(1, 10))
+		update_icon()
+		flickering = FALSE
+	return TRUE
+
 /obj/machinery/computer/update_icon()
 	overlays.Cut()
-	if(stat & NOPOWER)
+	if(stat & NOPOWER || force_no_power_icon_state)
 		if(icon_keyboard)
 			overlays += image(icon,"[icon_keyboard]_off",overlay_layer)
 		return
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -61,7 +61,7 @@
 /obj/machinery/computer/proc/flicker_event()
 	var/amount = rand(5, 15)
 
-	for(var/i = 0; i < amount; i++)
+	for(var/i in 1 to amount)
 		force_no_power_icon_state = TRUE
 		update_icon()
 		sleep(rand(1, 3))
@@ -74,7 +74,7 @@
 
 /obj/machinery/computer/update_icon()
 	overlays.Cut()
-	if(stat & NOPOWER || force_no_power_icon_state)
+	if((stat & NOPOWER) || force_no_power_icon_state)
 		if(icon_keyboard)
 			overlays += image(icon,"[icon_keyboard]_off",overlay_layer)
 		return
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -17,7 +17,9 @@
 	var/light_range_on = 2
 	var/light_power_on = 1
 	var/overlay_layer
+	/// Are we in the middle of a flicker event?
 	var/flickering = FALSE
+	/// Are we forcing the icon to be represented in a no-power state?
 	var/force_no_power_icon_state = FALSE
 
 /obj/machinery/computer/New()
@@ -38,9 +40,10 @@
 	set_light(0)
 	visible_message("<span class='danger'>[src] grows dim, its screen barely readable.</span>")
 
+/*
+ * Reimp, flash the screen on and off repeatedly.
+ */
 /obj/machinery/computer/flicker()
-	var/amount = rand(5, 15)
-
 	if(flickering)
 		return FALSE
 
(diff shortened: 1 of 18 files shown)


code/modules/power/apc.dm
```diff
@@ -419,18 +419,27 @@
...
 	if(!second_pass) //The first time, we just cut overlays
-		addtimer(CALLBACK(src, /obj/machinery/power/apc/proc.get_spooked, TRUE), 1)
+		addtimer(CALLBACK(src, /obj/machinery/power/apc/proc.flicker, TRUE), 1)
 		cut_overlays()
+		// APC power distruptions have a chance to propogate to other machines on its network
+		for(var/obj/machinery/M in area)
+			// Please don't cascade, thanks
+			if (M == src)
+				continue
+			if (prob(10))
```
SteelSlayer, line 435:
```suggestion
			if(M == src)
				continue
			if(prob(10))
```

code/modules/power/lighting.dm
```diff
@@ -469,19 +466,31 @@
...
+				if(status != LIGHT_OK)
+					break
+
+				on = FALSE
 				update(0)
-				sleep(rand(5, 15))
+				sleep(rand(1, 3))
+
+				on = (status == LIGHT_OK)
+				update(0)
```
SteelSlayer, line 488: Since this arg is a boolean.
```suggestion
				update(FALSE)
```

kenken244:
the random spook chance seems a little too high, considering it applies to EVERY object, take the bridge for example, it has at least 20 spookable objects, that's 40 spooks per round, or one every few minutes. I would lower the chance by half to a quarter, or maybe make it higher, but only 10-20% of objects are selected for random spooks each round. would have to see it in action to really know for sure

S-CR:
Update: Vending machines are now Boo!'able in a similar way to computers and lights

S-CR:
Update: Boo'ing objects in areas marked haunted such as the chapel or morgue invokes a lesser cooldown than boo'ing outside those areas, 50% less for a 60 second wait. Because duh.

code/game/machinery/machinery.dm
```diff
@@ -192,6 +207,8 @@ Class Procs:
 		use_power(idle_power_usage,power_channel, 1)
 	else if(use_power >= ACTIVE_POWER_USE)
 		use_power(active_power_usage,power_channel, 1)
+	if(prob(MACHINE_FLICKER_CHANCE))
```
Kyep, line 210: I'm really not sure about the idea of machines randomly flickering (ie: doing prob() calls regularly).

S-CR: From a technical, or design perspective?

SteelSlayer: Technical I'd imagine. It would be cool if you could profile the performance cost of this, i.e run a 30 minute test on master vs 30m with your changes. I don't think it will be *that* much of a cost, but worth checking out.

S-CR: I profiled it before I posted the PR, which is why I asked the question. Didn't save the results, so if it would make folks feel better I can profile it again when I have a few moments spare.

Necaladun:
@S-CR 
So uh, this is very selfish, but a little question/request:

Normally with BOO i can VV the area to say 12 to make all the lights in an entire room flicker and the like as a fun little spooky thing. Is this still possible?

S-CR:
@Necaladun Regrettably, no, this spell is no longer an AOE and instead is single-target.

code/game/machinery/machinery.dm
S-CR: Profiled the results of a run over 30 minutes each.

With the code enabled:
profile1-code_is_in

With the code disabled:
profile2-code_is_out

Variance in the number of calls can be explained by space ruins and away missions or slight changes in how I timed the runs, but it seems pretty clear this doesn't impact the proc in any major way.

code/game/machinery/computer/computer.dm
```diff
@@ -36,9 +40,41 @@
...
+
+	return TRUE
+
+/*
+ * Proc to be called by invoke_async in the above flicker() proc.
+ */
+/obj/machinery/computer/proc/flicker_event()
+	var/amount = rand(5, 15)
+
+	for(var/i = 0; i < amount; i++)
```
SteelSlayer, line 64:
```suggestion
	for(var/i in 1 to amount)
```

code/game/machinery/computer/computer.dm
```diff
@@ -36,9 +40,41 @@
...
+		force_no_power_icon_state = FALSE
+		update_icon()
+		sleep(rand(1, 10))
+	update_icon()
+	flickering = FALSE
+
 /obj/machinery/computer/update_icon()
 	overlays.Cut()
-	if(stat & NOPOWER)
+	if(stat & NOPOWER || force_no_power_icon_state)
```
SteelSlayer, line 77: I believe this needs parentheses around it, or you'll get some unintended behavior.
```suggestion
	if((stat & NOPOWER) || force_no_power_icon_state)
```

code/game/machinery/machinery.dm
SteelSlayer: Yep, this is negligible, and not really a concern. And all the other stuff (computers flickering etc.) happens so infrequently I don't think that will be a problem either.

code/game/machinery/vending.dm
```diff
@@ -138,6 +143,49 @@
 	for(var/obj/item/vending_refill/VR in component_parts)
 		restock(VR)
 
+/obj/machinery/vending/update_icon()
+	overlays.Cut()
```
SteelSlayer, line 147:
```suggestion
	cut_overlays()
```

code/game/machinery/vending.dm
```diff
@@ -138,6 +143,49 @@
...
+	if(panel_open)
+		overlays += image(icon, "[initial(icon_state)]-panel")
```
SteelSlayer, line 149:
```suggestion
		add_overlay(image(icon, "[initial(icon_state)]-panel"))
```

code/game/machinery/vending.dm
```diff
@@ -138,6 +143,49 @@
...
+
+	return TRUE
+
+/*
+ * Proc to be called by invoke_async in the above flicker() proc.
+ */
+/obj/machinery/vending/proc/flicker_event()
+	var/amount = rand(5, 15)
+
+	for(var/i = 0; i < amount; i++)
```
SteelSlayer, line 179:
```suggestion
	for(var/i in 1 to amount)
```

code/modules/mob/dead/observer/spells.dm
```diff
@@ -9,22 +9,38 @@ GLOBAL_LIST_INIT(boo_phrases, list(
...
+	// no need to spam admins regarding boo casts
+	create_logs = FALSE
 
-/obj/effect/proc_holder/spell/aoe_turf/boo/cast(list/targets, mob/user = usr)
-	for(var/turf/T in targets)
-		T.get_spooked()
+
+/obj/effect/proc_holder/spell/targeted/click/boo/cast(list/targets, mob/user = usr)
+	var/atom/target = targets[1]
+	ASSERT(istype(target))
```
SteelSlayer, line 37: I don't *think* you need this because the only clickable things should be atoms, you can't click on datums.

S-CR: Is there a reason not to confirm type safety? Unlikely (Or seemingly impossible), yes, but it's also an inexpensive check.

SteelSlayer: Yeah usually safety checks are good. I just think that its uneeded here because clicking on a non-atom is impossible. But like you said its not expensive, and doesn't matter either way. Up to you (or maints) to decide.

Issue #14276, opened 2020-09-10
Security jacket is invisible for greys.


Marcolinoj:
**Issue Description**:
Grey's security jacket sprite is now invisible.

**What did you expect to happen**: 
To see the security jacket on my character when I put it on.

**What happened instead**:
When I put on the security jacket, the sprite doesnt appear on my character

**Why is this bad/What are the consequences**:
Makes greys unable to use the alternative to the armor, the security jacket.

**Steps to reproduce the problem**:
Be a grey. and then put on a security jacket

**When did the problem start happening**:
After the grey sprite update, where the armor and hailer were fixed, probably something there broke it.

**Extra information**:
Screenshots of the jacket not showing
![image](https://user-images.githubusercontent.com/61087399/92781440-8455da00-f37a-11ea-94c5-a08b52c9a715.png)
![image](https://user-images.githubusercontent.com/61087399/92781461-891a8e00-f37a-11ea-8c86-2815ec47f9e8.png)

SabreML:
Unable to reproduce, so presumably this was fixed at some point.

PR #14240, opened 2020-09-05
Fixes Greys' armor and secmask sprites

kugamo:
## What Does This PR Do
Fixes https://github.com/ParadiseSS13/Paradise/issues/8174
This PR adds grey species specific sprites for armor, and reworks the sprites for the grey's sec mask. Also Greys had a missing sprite for the 'tan pants', which I added just for the hell of it.

## Why It's Good For The Game
These sprites looked very wonky. The armor didnt fit the Greys back and it covered the face. The sec mask stuck out weirdly from the face and had strange alignment with the neck strap. When put off the mouth the sec mask didnt move at all, and had some weird alignment when facing different directions. I made all the strap positions consistent.

## Images of changes

gray_sprite_changes

## Changelog
:cl:
tweak: Resolves issues with Grey armor, secmask, and tanpants sprites.
/:cl:

PR #14545, opened 2020-10-06
Fixes "Invisible" Suit Sprites for Grey

Kiyahitayika:
## What Does This PR Do
Tweaks code changes made in #14240 which had the unintended side effect of rendering several Grey clothing items "invisible". 

## Why It's Good For The Game
For some reason, `/icons/mobs/species/grey/suits.dmi` is less multiple suit sprites. I moved the `sprite_sheet` overrides for @kugamo neat new sprites for the two security armor types into the code instantiating the security armors.

Addresses issue raised in #14276 

Overrode `sprite_list = list()` instead of using `sprite_list +=` so we wouldn't have to worry about removing items from the list at other points.

## Images of changes
See Security Jacket not being invisible and Kugamo's sprites still working!
https://imgur.com/a/gdbAsKo

## Changelog
:cl:
tweak: Moved override of `sprite_sheet` in `/code/modules/clothing/suits/armor.dm`
fix: Fixed "invisible" Grey suit sprites
/:cl:

### DIFF
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -12,7 +12,6 @@
 	armor = list("melee" = 30, "bullet" = 30, "laser" = 30, "energy" = 10, "bomb" = 25, "bio" = 0, "rad" = 0, "fire" = 50, "acid" = 50)
 	sprite_sheets = list(
 		"Vox" = 'icons/mob/species/vox/suit.dmi',
-		"Grey" = 'icons/mob/species/grey/suit.dmi'
 		)
 	w_class = WEIGHT_CLASS_NORMAL
 
@@ -41,6 +40,10 @@
 /obj/item/clothing/suit/armor/vest/security
 	name = "security armor"
 	desc = "An armored vest that protects against some damage. This one has a clip for a holobadge."
+	sprite_sheets = list(
+		"Vox" = 'icons/mob/species/vox/suit.dmi',
+		"Grey" = 'icons/mob/species/grey/suit.dmi'
+	)
 	icon_state = "armor"
 	item_state = "armor"
 	var/obj/item/clothing/accessory/holobadge/attached_badge
@@ -51,9 +54,12 @@
 			add_fingerprint(user)
 			W.forceMove(src)
 			attached_badge = W
-
 			var/datum/action/A = new /datum/action/item_action/remove_badge(src)
 			A.Grant(user)
+			sprite_sheets = list(
+				"Vox" = 'icons/mob/species/vox/suit.dmi',
+				"Grey" = 'icons/mob/species/grey/suit.dmi'
+			)
 			icon_state = "armorsec"
 			user.update_inv_wear_suit()
 			desc = "An armored vest that protects against some damage. This one has [attached_badge] attached to it."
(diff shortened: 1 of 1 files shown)


code/modules/clothing/suits/armor.dm
```diff
@@ -41,6 +40,10 @@
 /obj/item/clothing/suit/armor/vest/security
 	name = "security armor"
 	desc = "An armored vest that protects against some damage. This one has a clip for a holobadge."
+	sprite_sheets = list(
+		"Vox" = 'icons/mob/species/vox/suit.dmi',
```
kugamo, line 44: no need for vox in the list here in `obj/item/clothing/suit/armor/vest/security` , since its already in `/obj/item/clothing/suit/armor`.

code/modules/clothing/suits/armor.dm
```diff
@@ -51,9 +54,12 @@
 			add_fingerprint(user)
 			W.forceMove(src)
 			attached_badge = W
-
 			var/datum/action/A = new /datum/action/item_action/remove_badge(src)
 			A.Grant(user)
+			sprite_sheets = list(
+				"Vox" = 'icons/mob/species/vox/suit.dmi',
```
kugamo, line 60: same as earlier comment, no need for vox in the list here.

kugamo:
Thanks for this fix, I hadn't realised my code would break all the children of `/obj/item/clothing/suit/armor` for greys. 😃

Kiyahitayika:
> Thanks for this fix, I hadn't realised my code would break all the children of `/obj/item/clothing/suit/armor` for greys. 😃

The problem is that Greys are missing a ton of suit sprites! It was only because someone submitted an issue about the Security Jacket that I decided to dig into the sprite files and found them missing. Then I checked the recent merges. The only one was yours which changed the path file and I came to the realisation that the majority of suit sprites were missing since Greys were added.  :( Poor Greys! I think your tweaks to the armor are great!

code/modules/clothing/suits/armor.dm
```diff
@@ -51,9 +53,11 @@
 			add_fingerprint(user)
 			W.forceMove(src)
 			attached_badge = W
-
 			var/datum/action/A = new /datum/action/item_action/remove_badge(src)
 			A.Grant(user)
+			sprite_sheets = list(
+				"Grey" = 'icons/mob/species/grey/suit.dmi'
+			)
```
SteelSlayer, line 60: Why does this need to be reassigned here in attackby? Is there a place where the armor gets its `sprite_sheets` value changed from the default, which already has this assigned?

Kiyahitayika: There's two armor sprites.
`icon_state = "armor"`
`icon_state = "armorsec"`

Kiyahitayika: Ah, my apologies, it was hard to see on Github. Looking into your suggestion offline.

Kiyahitayika: As expected, this works! :D
Nice catch! Thank you so much!

Kiyahitayika (commented):
Yaya!

AffectedArc07 (approved):

PR #14546, opened 2020-10-06
Reduces Buckshot Armor Penetration

S-CR:
## What Does This PR Do
Adds a -30 armor penetration modifier to buckshot pellets.

## Why It's Good For The Game
Shotguns - especially loaded with buckshot - stand head and shoulders above other ballistic weapons because of how quickly they turn even nasty opponents into pulled pork with a meaty shot or two. I don't think it's great for the game, and it's certainly not "realistic". This PR makes buckshot reflect more how it actually behaves - that is to say, unarmored targets are in deep trouble, armored targets stand a much better chance.

After these changes an unarmored human at point blank takes ~180 brute from three shots to the chest. With a sec vest on that's reduced to ~120. With a bulletproof vest on that's reduced to ~45.

I'm not sure this will get merged in its current state, but I hope to start a discussion about what we can do to improve this at very least.

## Changelog
:cl: Dave
tweak: Buckshot now has a harder time punching armor.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets.dm
+++ b/code/modules/projectiles/projectile/bullets.dm
@@ -85,6 +85,7 @@
 	damage = 12.5
 	tile_dropoff = 0.75
 	tile_dropoff_s = 1.25
+	armour_penetration = -30
 
 /obj/item/projectile/bullet/pellet/rubber
 	name = "rubber pellet"


FlattestGuitar:
Salt PR. Needs repo ban.

S-CR:
Salt PR would be making buckshot have an even money chance of firing backwards.

Rurik123:
Do slugs also have fairly good armor penetration?

If so, odds are everyone will just start using slugs instead. Which is fine, if that's the goal, but it sounded more like you wanted shotguns to be less prevalent in the ballistic meta, so to speak, and this won't change that. If that's the case, probably better off tweaking shotguns themselves since slugs are already arguably better than buckshot in the first place.

Denthamos:
I can agree with this view actually.  Buckshot is way overused even reaching into where slugs should be used as the spread is not as great as what people think it is.  At three tiles you can still reasonably hit all buckshot pellets into the same target if you are on the same x/y axis as the target.  Beyond the 3 tile range people **should** be using slugs instead of buckshot.

S-CR:
@Rurik123 I don't mega have an issue with slugs, there are plenty of availability vs. usability alternatives for slug-like behavior (Autorifles, for example). Buckshot stands out as an issue because it's pretty hard to argue it's worse than pretty much any alternative *and* it's one of the most likely ballistics for a crew member to get their hands on besides a traitor uplink.

Balthusdire:
While the idea is an interesting one that deserves consideration, I dont think this is the way to do it. -30 is a massive amount and will change up the balance of guns considerably. Yes its more realistic but we need to be balancing things based on what's going to be best for the game and a big nerf to buckshot vs armor is something that Im not sure is the best way to do it.

Varghh:
> While the idea is an interesting one that deserves consideration, I dont think this is the way to do it. -30 is a massive amount and will change up the balance of guns considerably. Yes its more realistic but we need to be balancing things based on what's going to be best for the game and a big nerf to buckshot vs armor is something that Im not sure is the best way to do it.

Yes and no, i truly don't mind this buckshot nerf because Buckshots are still going to be amazing to deal with Spiders, Xenos and unarmored targets like Bobby the vampire who won't stop misting and eating civilians in science maints. It's going to make you unable to just flank a nukie and blast his ass off in seconds, that's honestly great for the game since nukies are most of the times wrecked by buckshots.

Kyep (approved):

AffectedArc07 (approved):

PR #14547, opened 2020-10-06
Sentient minebots are no longer convertable to cult

Kyep:
## What Does This PR Do
Attempting to use a cultist convert rune on a sentient minebot no longer converts the minebot to cult.
Instead, it simply destroys (sacrifices) the minebot.

## Why It's Good For The Game
A single cultist miner being able to make many minebots, make them all sentient, and then turn them all into permanent cultists, is overpowered.

## Changelog
:cl: Kyep
tweak: sentient minebots can no longer be converted to cult.
/:cl:


SteelSlayer (approved):
I question why all simple animals themselves aren't already prevented from being converted. Nevertheless, this works.

variableundefined (approved):

PR #13712, opened 2020-06-28
Fix command Admin, Secrets, Debug, Make all areas powered

blinkdog:
## What Does This PR Do
While working on #13117 , I made `Make all areas powered` an alias to `Power all SMES`. In hindsight, this was a mistake, as there are useful reasons to recharge APCs independently of the SMES.

This PR restores the code necessary to recharge APC cells with the command. Note: It does NOT repair broken APCs or add power cells where they are missing. The command simply recharges APC power cells up to full power.

Originally, the command also restored some charge to the SMES based on the Grid Check event. That part has been removed, as the new APC Short event does not affect the SMES. If you want to recharge the SMES, just use the `Power all SMES` command.

## Why It's Good For The Game
Admins probably didn't use this command very often in real rounds.
It is a useful command to have for debugging and testing, though.

## Changelog
:cl:
fix: Fixed Admin command Make all areas powered
/:cl:

PR #14553, opened 2020-10-06
Fix - Admin Secrets - "Make All Areas Powered", "Make All Areas Unpowered"

### LINKED ISSUE #14249: "Make all areas unpowered" secret shorts out/overloads random APCs instead of draining power from everything
**Issue Description**:
"Make all areas unpowered" secret shorts out/overloads random APCs instead of draining power from everything.

For all APCs on the station to get drained of power.

**What happened instead**:
APCs got shorted/overloaded.

It makes testing things with power annoying since you can't make all areas unpowered and then repowered them, because the APCs become borked.

1. Open secrets panel
2. Click "Make all areas unpowered"

Likely after this PR: #13117

Kiyahitayika:
## What Does This PR Do
Tweaks how "Make All Areas Powered" and "Make All Areas Unpowered" work in Admin Secrets commands. 

Fixes #14249, spoke with @SteelSlayer
Continues the work merged in https://github.com/ParadiseSS13/Paradise/pull/13117 and #13712, spoke with @blinkdog

## Why It's Good For The Game
As raised in the issue referenced above, "Make All Areas Powered" does not currently undo the effects of "Make All Areas Unpowered". After speaking with SteelSlayer, it was also brought up that "Make All Areas Unpowered" is a misnomer as it actually shorts out APCs. This PR adds the following features:

**Make All Areas Unpowered** (separate `procs`):
- "Depower all APCs" - sets the charge of all APCs' cells to 0
- "Short out APCs" - shorts out APCs per the current form of "Make All Areas Unpowered"

**Make All Areas Powered:**
- "Power All APCs" - (Power without Repairing) - sets the charge of all functioning APCs' cells to their `maxcharge`.
- "Repair All APCs" - (Repair without Powering) - repairs all APCs to reverse the effects of "Short out APCs"
- "Repair and Power APCs" - (Repair and Power) - repairs all APCs and sets their cells to their `maxcharge`.

Please see the "Images of Changes" video.

## Images of changes
https://imgur.com/a/fPzATEI

## Changelog
:cl:
add: New features to Admin Secrets within "Make all areas powered" and "Make all areas unpowered"
fix: "Make all areas powered" can now undo "Make all areas unpowered"
/:cl:


Kiyahitayika:
Yep. Fixing it.

McRamon:
Can you also add ability to rename boxes, so empty syndie box named something like "c4 kit" is no longer a meta reason to suspect or arrest someone

code/modules/events/apc_short.dm
```diff
@@ -27,51 +27,105 @@
...
-				C.toggle_breaker()
-			// no matter what, ensure the area knows something happened to the power
-			current_area.power_change()
-			affected_apc_count++
-	log_and_message_admins("APC Short event shorted out [affected_apc_count] APCs.")
+	switch(alert("What Would You Like to Do?", "Make All Areas Unpowered", "Depower all APCs", "Short out APCs"))
+		if("Depower all APCs")
+			var/list/skipped_areas_apc = list(
+			/area/engine/engineering,
+			/area/turret_protected/ai)
```
SteelSlayer, line 34: Spacing.
```suggestion
			var/list/skipped_areas_apc = list(
				/area/engine/engineering,
				/area/turret_protected/ai)
```

code/modules/events/apc_short.dm
```diff
@@ -27,51 +27,105 @@
...
+					continue
+				var/obj/item/stock_parts/cell/C = A.get_cell()
+				if(C)
+					C.charge = 0
+				current_area.power_change()
+			log_and_message_admins("Power has been drained from all APCs.")
+		if("Short out APCS")
+			var/list/skipped_areas_apc = list(
+			/area/engine/engineering,
+			/area/turret_protected/ai)
```
SteelSlayer, line 49: Same.
```suggestion
			var/list/skipped_areas_apc = list(
				/area/engine/engineering,
				/area/turret_protected/ai)
```

code/modules/events/apc_short.dm
```diff
@@ -27,51 +27,105 @@
...
-	log_and_message_admins("APC Short event shorted out [affected_apc_count] APCs.")
+	switch(alert("What Would You Like to Do?", "Make All Areas Unpowered", "Depower all APCs", "Short out APCs"))
+		if("Depower all APCs")
+			var/list/skipped_areas_apc = list(
+			/area/engine/engineering,
+			/area/turret_protected/ai)
+			if(announce)
+				GLOB.event_announcement.Announce("Power failure detected in [station_name()]'s powernet. All APCs have lost power. Gravity systems likely to fail.", "Systems Power Failure", new_sound = 'sound/AI/attention.ogg')
+			for(var/obj/machinery/power/apc/A in GLOB.apcs)
+				var/area/current_area = get_area(A)
```
SteelSlayer, line 38: Why did you change this from a typeless loop to a filtered loop? Same with the others below.

code/modules/events/apc_short.dm
```diff
@@ -27,51 +27,105 @@
...
-		var/obj/item/stock_parts/cell/C = A.get_cell()
-		if(C)
-			C.give(C.maxcharge)
+	switch(alert("What Would You Like to Do?", "Make All Areas Powered", "Power all APCs", "Repair all APCs", "Repair and Power APCs"))
+		if("Power all APCs")	//Power without Repairing
+			if(announce)
+				GLOB.event_announcement.Announce("All operational APCs on [station_name()] have been fully charged.", "Power Systems Nominal", new_sound = 'sound/AI/poweron.ogg')
+			var/affected_apc_count = 0
+			for(var/obj/machinery/power/apc/A in GLOB.apcs)
+				var/obj/item/stock_parts/cell/C = A.get_cell()
```
SteelSlayer, line 82: I would avoid getting this cell until it is needed on line 87-88. There is a possibility that you get this cell, and then `continue` because the APC is not on the station z-level, thus doing a pointless operation.

code/modules/events/apc_short.dm
```diff
@@ -27,51 +27,105 @@
...
+				if(A.shorted)
+					A.shorted = FALSE
+				current_area.power_change()
+			log_and_message_admins("Power has been restored to all APCs.")
+		if("Repair and Power APCs")	//Repair and Power APCs
+			if(announce)
+				GLOB.event_announcement.Announce("All APCs on [station_name()] have been repaired and recharged. We apologize for the inconvenience.", "Power Systems Optimal", new_sound = 'sound/AI/poweron.ogg')
+			// repair the APCs and recharge them
+			for(var/obj/machinery/power/apc/A in GLOB.apcs)
+				var/obj/item/stock_parts/cell/C = A.get_cell()
```
SteelSlayer, line 112: Same thing here, declare this when you need it, sometime after the `continue` below.

code/modules/events/apc_short.dm
Kiyahitayika: Huh. I thought it would be cleaner code. Are we expecting something other than APCs? be inside GLOB.apcs? 😰

SteelSlayer: Nope, we shouldn't be expecting anything else (unless someone did an oopsie or there is a bug). That's why we can use a typeless loop (which is faster) and just cast the thing to an APC.

Landerlow:
God yes

code/modules/events/apc_short.dm
Kiyahitayika: Can we move this declaration to above the `switch()` statement so we only declare it once?

Kiyahitayika: `/proc/power_failure(announce=TRUE)
	var/list/skipped_areas_apc = list(
		/area/engine/engineering,
		/area/turret_protected/ai)
	switch(alert("What Would You Like to Do?", "Make All Areas Unpowered", "Depower all APCs", "Short out APCs"))`

code/modules/events/apc_short.dm
Kiyahitayika: See my question above

code/modules/events/apc_short.dm
Kiyahitayika: Also, do we want to add Grav Gen to the list?

code/modules/events/apc_short.dm
```diff
@@ -27,51 +27,105 @@
 	return TRUE
 
 /proc/power_failure(announce=TRUE)
```
blinkdog, line 29: Careful here; this proc is called by the event code on Line 15 above.
I'm not sure what value an `alert()` will return when not directly invoked by a client admin verb.
Or to put it another way: When the server fires the proc as an event, who would get the pop-up selection dialog to choose what happens?

Test that the event still works by opening the Admin -> Event Manager Panel, choosing APC Short for the Moderate event, then reduce the time in order to make the event fire within in a few seconds.

APC Short does not correct itself when the event is over, so the changes to the `power_restore` proc shouldn't affect the event.

Kiyahitayika: @blinkdog Thank goodness I keyed you in on this. You are a lifesaver. I'm going to have to test that and I'm expecting one of two outcomes:
A.) Server crashes
B.) The alert pops up
Either way, that's _not_ what we want.
The solution would be to plug in the alert somewhere else, probably `topics.dm`, and make two separate procs:
- `power_failure` would be your APC shorting code
- `apc_depower` would be the other one

code/modules/events/apc_short.dm
SteelSlayer: >Can we move this declaration to above the switch() statement so we only declare it once?

Yes.
>Also, do we want to add Grav Gen to the list?

I would just leave it as is.

code/modules/events/apc_short.dm
Kiyahitayika: Hypotheses are great, especially when they are wrong.

> A.) Server crashes
> B.) The alert pops up

No, let's try:

C.) No alert pops up and it defaults to "Depower all APCs"

Kiyahitayika:
@blinkdog  Thank you!

Changed from a `switch()` on `apc_short.dm` to an `if()` on `topic.dm`. Separated into two procs called `power_failure` and `depower_apcs`. 

image

Kiyahitayika:
Travis fix attempt, fingers crossed.

code/modules/events/apc_short.dm
```diff
@@ -26,55 +26,108 @@
...
-			if(C.operating)
-				C.toggle_breaker()
+			if(A.wires)
+				if(!A.wires.is_cut(WIRE_MAIN_POWER1))
+					A.wires.cut(WIRE_MAIN_POWER1)
+				if(!A.wires.is_cut(WIRE_MAIN_POWER2))
+					A.wires.cut(WIRE_MAIN_POWER2)
+				// if it was operating, toggle off the breaker
+				if(A.operating)
+					A.toggle_breaker()
```
SteelSlayer, line 68: Why did line 67 and 68 get indented?

code/modules/events/apc_short.dm
```diff
@@ -26,55 +26,108 @@
...
 		GLOB.event_announcement.Announce("Overload detected in [station_name()]'s powernet. Engineering, please repair shorted APCs.", "Systems Power Failure", new_sound = 'sound/AI/attention.ogg')
-
 	// break APC_BREAK_PROBABILITY% of all of the APCs on the station
 	var/affected_apc_count = 0
-	for(var/thing in GLOB.apcs)
-		var/obj/machinery/power/apc/C = thing
-		// skip any APCs that are too critical to disable
-		var/area/current_area = get_area(C)
-		if((current_area.type in skipped_areas_apc) || !is_station_level(C.z))
+	for(var/obj/machinery/power/apc/A in GLOB.apcs)
```
SteelSlayer, line 54: You should leave this syntax how it was before:
```dm
for(var/thing in GLOB.apcs)
	var/obj/machinery/power/apc/A = thing
```
Ditto for other loops dealing with `GLOB.apcs` below you may have changed.

code/modules/admin/topic.dm
```diff
@@ -2782,9 +2782,11 @@
...
 				feedback_inc("admin_secrets_fun_used",1)
 				feedback_add_details("admin_secrets_fun_used","UP")
-				log_admin("[key_name(usr)] made all areas unpowered", 1)
-				message_admins("<span class='notice'>[key_name_admin(usr)] made all areas unpowered</span>", 1)
-				power_failure()
+				if(alert("What Would You Like to Do?", "Make All Areas Unpowered", "Depower all APCs", "Short out APCs") == "Depower all APCs")
+					depower_apcs()
+				else
+					power_failure()
+				log_and_message_admins("<span class='notice'>[key_name_admin(usr)] made all areas unpowered</span>", 1)
```
SteelSlayer, line 2789: Can you make this log say they "shorted APCs" if they chose to short, and "made all areas unpowered" if they chose to de-power APCs?

code/modules/events/apc_short.dm
```diff
@@ -26,55 +26,108 @@
...
+					continue
+				if(A.wires)
+					A.wires.repair()
+				if(!A.operating)
+					A.toggle_breaker()
+				if(A.shorted)
+					A.shorted = FALSE
+				var/obj/item/stock_parts/cell/C = A.get_cell()
+				if(C)
+					C.charge = C.maxcharge
```
SteelSlayer, line 124: Since these bits of code are used a few times, it would be neat to see them turned into APC procs. Line 116-121 could be put in a proc called `repair_apc` for example. And the code from line 122-124, could be called `recharge_apc`. Optional request here.

SteelSlayer (commented):
Small stuff. The code at large seems to work fine.

variableundefined (changes requested):
Address Steel Concerns

code/modules/events/apc_short.dm
Kiyahitayika: Agreed! You pointed this out to me before and I thought I made the change back to the original syntax. I wonder if it reverted to my change (using a filtered for loop) when I closed and re-opened. I'll make the change again and see if it sticks. Ever since you explained this to me, I've been trying to make sure I code my loops in the way you have suggested.

code/modules/events/apc_short.dm
Kiyahitayika: I had not considered that. I like the idea. When working on this, I was wondering if there was a neater way to do it. Let's see what implementing such procs would look like based on your suggestions.

code/modules/events/apc_short.dm
Kiyahitayika: There's also `for(var/mob/living/M in GLOB.player_list)` on line 17. Is this redundant because GLOB.player_list is all living mob players? Could it just be `for(var/M in GLOB.player_list)`? or:

`for(var/M in GLOB.player_list)
var/mob/living/mob = M`

code/modules/events/apc_short.dm
Kiyahitayika: That's curious. According to Gitlens, it look I cute a lot of the code out and then pasted it back in. If I recall correctly, I was trying to adjust it so admins had options both for depowering the station _and_ powering the station. In the midst of rearranging the code, I must have missed that indent. I should have caught that and did not.

code/modules/events/apc_short.dm
SteelSlayer: Yeah so in this case it's necessary to have a filtered loop because the `GLOB.player_list` contains both alive and dead (ghost) mobs. However you *can* make it a typeless loop, but you'll need to check if `M` is not a dead mob first, then cast it to a living one. I'm not sure if that is actually any better though.

code/modules/admin/topic.dm
Kiyahitayika: No problem. I was not sure what Admins would find most helpful here.

Side Note: A lot of the code here has both `log_admin` and `message_admins`.
`log_and_message_admins` exist and it's really ... really tempting to change those lines.

Kiyahitayika: `if(alert("What Would You Like to Do?", "Make All Areas Unpowered", "Depower all APCs", "Short out APCs") == "Depower all APCs")
					depower_apcs()
					log_and_message_admins("<span class='notice'>[key_name_admin(usr)] made All Areas Unpowered via Admin Secrets.</span>", 1)
				else
					power_failure()
					log_and_message_admins("<span class='notice'>[key_name_admin(usr)] has Shorted APCs via Admin Secrets.</span>", 1)`

Kiyahitayika: Additionally changed how the procs are called.

Kiyahitayika:
Made changes as per SteelSlayer's requests.
In addressing the request related to topic.dm, I realized that there was an inconsistency in where I placed the 'alerts' for depowering the station and powering the station. I moved the alert from the power_restore() proc to topic.dm. This should free up power_restore() to be called elsewhere without an alert popping up.

code/modules/events/apc_short.dm
```diff
@@ -63,67 +64,54 @@
...
-				if(A.shorted)
-					A.shorted = FALSE
-				var/obj/item/stock_parts/cell/C = A.get_cell()
-				if(C)
-					C.charge = C.maxcharge
-				current_area.power_change()
-			log_and_message_admins("Power has been restored to all APCs.")
+/proc/power_restore(announce=TRUE, var/power_type)
+	var/type = power_type //Using if() statements instead of a Switch since ifs are supposed to be slightly faster and it's for three case types
+	if(type == 0)	//Power without Repairing
```
SteelSlayer, line 77: I'd say ditch the `type` var and just use `power_type` directly.

Kiyahitayika: Oh sure!

SteelSlayer (approved):
Looks fine to me.

Kiyahitayika:
Wait - a Map Diff Render? I was just deconflicting in accordance with PR #15132.

Kiyahitayika:
https://github.com/ParadiseSS13/Paradise/pull/15199

PR #14557, opened 2020-10-07
[DNM] Summon Ghosts Allows Ghosts to Possess Items

S-CR:
## What Does This PR Do
The Wizard's "summon ghosts" ritual now allows ghosts to Boo! an item to possess it and cause it to float around very spookily. Possessed items may be used by their possessor in most ways that you could normally use the item in a human's hand. Donuts may be forcefed to people, clothes can put themselves on targets, syringes can inject, guns (Or disarmed wizard staves) can fire, etcetera. For obvious reasons, you cannot possess items on Z2.
**NOTE:** This PR is branched off of #14543 and requires that framework for these changes.

## Why It's Good For The Game
Wizard rounds are about tomfoolery, and keeping ghosts involved in the round is cool and fun.

## Changelog
:cl: Dave
add: Summon ghosts now allows ghosts to possess items by casting Boo! on them.
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -75,6 +75,9 @@
 #define ORGANIC 1
 #define SYNTHETIC 2
 
+#define GHOST_POWER_NORMAL 0
+#define GHOST_POWER_SPOOKY 1
+
 // Appearance change flags
 #define APPEARANCE_UPDATE_DNA 1
 #define APPEARANCE_RACE	2|APPEARANCE_UPDATE_DNA
--- a/code/datums/spell.dm
+++ b/code/datums/spell.dm
@@ -5,7 +5,7 @@
 	var/panel = "Debug"//What panel the proc holder needs to go on.
 	var/active = FALSE //Used by toggle based abilities.
 	var/ranged_mousepointer
-	var/mob/living/ranged_ability_user
+	var/mob/ranged_ability_user
 
 /obj/effect/proc_holder/singularity_act()
 	return
@@ -15,7 +15,7 @@
 
 GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 
-/obj/effect/proc_holder/proc/InterceptClickOn(mob/living/user, params, atom/A)
+/obj/effect/proc_holder/proc/InterceptClickOn(mob/user, params, atom/A)
 	if(user.ranged_ability != src)
 		to_chat(user, "<span class='warning'><b>[user.ranged_ability.name]</b> has been disabled.")
 		user.ranged_ability.remove_ranged_ability(user)
@@ -38,7 +38,7 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 	spell.remove_ranged_ability(spell.ranged_ability_user)
 	return ..()
 
-/obj/effect/proc_holder/proc/add_ranged_ability(mob/living/user, var/msg)
+/obj/effect/proc_holder/proc/add_ranged_ability(mob/user, var/msg)
 	if(!user || !user.client)
 		return
 	if(user.ranged_ability && user.ranged_ability != src)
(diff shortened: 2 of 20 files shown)


taukausanake:
This sounds very dangerous and I really like the idea. A problem though might show up with some ghosts going straight for weapons and either try to valid the wizard (which is probably granted) or murderbone crew. That should be handled by the rules and admins unless Wiz round will make an  exception but I think we all know how that will play out...

Unrelated suggestion: do you think it might be better to make it a tiered spell so you can still have the basic ghosting if you want only that and then buy the spell again to give them possession? I feel like possession should cost something above just spawning ghosts. Might be a balance issue to look at.

Lich + ghost possessed phylactery sounds fun. Me want

S-CR:
Ghosts will almost certainly beeline for weaponry, I expect it. Ghosts certainly can try to murderbone crew or the wizard as such, but they'll have a hard time of that (Especially as guns, probably the most likely culprit for possession bumrushing) because possessed items don't have an easy way to reload.

I have zero control over how admin policy views this sort of thing, especially attacking crew, so it is whatever they decide to make it - but personally I see ghosts being able to attack crew as a good thing and a component of that gamble. "Will the ghosts do more damage to the crew than to me" is a neat question to have to ask. It's also worth nothing that possessed items are relatively weak at 50 HP, and there's an implicit two minute cooldown once you die because your boo charge is reset when you ghost.

In my opinion, though, the best part of the feature is the broad range of creativity ghosts can apply because this feature is quite flexible. Will a crack team of ghost-tools run around hacking doors? Will we have a ghostly guitar playing a melancholy tune down a dimly lit hallway? Will a ghostly hand teleporter make a daring escape? I really, really hope so! Murdering is probably the most boring thing a ghost can opt to do with this feature.

taukausanake:
> crack team of ghost-tools run around hacking doors

Holy fuck you are a genius. I want a station that is just ghosts possessing everything in order to get their job done. I want to see ghost teams doing the Botanist and chef jobs. Ghosts mining Lavaland. Ghosts police hunting criminals with batons and cuffs.
How would cuffs work now that I think about it? Would the ghost stop possessing once someone is cuffed? I don't recall if the cuffs are deleted. And maybe if ghosts decided to make transfer valve bombs, putting the parts together

S-CR:
So if a ghost does a thing like putting themselves on a person as handcuffs or clothing, or inside a box as an item, the item works as it normally does when in that state with no real way for the ghost to interact until the item is removed and placed back on a turf.

However, possessed items are allowed to ghost to end their possession without the normal penalties for ghosting and the intro message tells them so.

Cocacolagua:
This is going to be super fun, and bring back the good ole' days where ghosts teamed-up in mindswap wizard summon ghosts rounds to lynch random people by using the crew's metagaming power against them.

TDSSS:
Considering how much chaos this causes, the cost of the spell might need some tweaking, since (iirc) it's currently 0 points.

PacifistDalek:
This is incredible and I look forward to ghost possessing the HoP's stamp and stamping every piece of paper I pass

Kyep (changes requested):
Objection.
I don't like the concept of ghosts possessing items, even as a wizard spell. The way item possession works, I worry its going to effectively be like that scene from the ghostbusters where the ghostbusters cause total chaos on the station in the name of pranks, to the point that the ghosts become the real antag. And the crew is just not equipped to fight ghosts.

Necaladun:
Yeah as cool as this is (And maybe could be tweaked to be merged over halloween....) I worry about 10+ ghosts rushing the armoury and robusting all of sec. There's not much to really stop them either, if you kill the item they'll just grab another.

I like the theory of wizards having ghosts possess things - a wand or staff to allow it could be cool. But having everyone who is dead able to mass-possess items and run over the station would be pure chaos.

moxian:
Would tweaking this so that:
- it becomes a new activatable spell, with a cooldown
- when cast, only items in the vicinity of the wizard become posessable
address the "too much chaos" concerns? It limits creativity a bit, naturally, but it also gives the wizard means to control the amount of chaos induced.

The caveats are: I have no idea how hard would this be to implement and I don't know of a way to tell ghosts which items are possessable (maybe just add an overlay on the tiles affected?)

S-CR:
I've had a few people approach me with ideas, though I'm not making any further changes until the precursor PR is merged.

AffectedArc07:

Issue #14362, opened 2020-09-18
Legion Cores Burning Drasks When Used


Medium-Salad:
**Issue Description**:

When a Drask uses a legion core, the temperature reset passes the species threshold for overheat, causing burn damage.

**What did you expect to happen**: 

For a Drask to use a legion core and be reset to their specific species base body temperature.

**What happened instead**:

Drask gets overheated when using a legion core.

**Why is this bad/What are the consequences**:

This makes Drask as a species worse than they already are for mining, and causes an annoyance every time you use an item meant for specifically healing.

**Steps to reproduce the problem**:

Use a Legion Core as a Drask.

**Extra information**:

Not high priority whatsoever.

Spacemanspark:
Maybe this should be made intentional tbh

It makes sense that a drask--something specifically meant to keep away from heat--is pretty boned in a place like lavaland.

Kiyahitayika:
@Medium-Salad  @Spacemanspark  This is ... somewhat intentional. What I mean is that the way Drask is coded gives them specific heat resistance levels:
```	heat_level_1 = 300 //Default 360 - Higher is better
	heat_level_2 = 340 //Default 400
	heat_level_3 = 400 //Default 460
```
And then they take 200% damage from going over those `heat_levels`:
```heatmod = 2```

They would likely also take damage from other objects/weapons producing heat, such as the liquid cleaner from a Cleaning Grenade.

Kiyahitayika:
Ah, going to reference Fox's comment in https://github.com/ParadiseSS13/Paradise/issues/13938

PR #14558, opened 2020-10-07
Legion Core - Burn Damage Fix

Kiyahitayika:
## What Does This PR Do
Changes the Body Temp setting when using the Legion Core to set to the Owner Species' `heat_level_1` threshhold instead of `BODYTEMP_NORMAL`. Ensures that non-human mobs continue to use `BODYTEMP_NORMAL`.

Fixes https://github.com/ParadiseSS13/Paradise/issues/14362
, spoke with @Medium-Salad.

## Why It's Good For The Game
`BODYTEMP_NORMAL` is a static `#define` of `310.15`. This value is greater than the `heat_level_1` of some races. As such, using the Legion Core as a these races would induce burn damage. This goes against the purpose of the Legion Core which is to heal brute and burn damage.

## Images of changes
Footage: https://imgur.com/a/YP21VRu

## Changelog
:cl:
fix: Stopped burn damage from Legion Core on species
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -221,7 +221,7 @@
 	owner.remove_CC()
 	if(ishuman(owner))
 		var/mob/living/carbon/human/H = owner
-		H.bodytemperature = H.dna.species.heat_level_1
+		H.bodytemperature = H.dna.species.body_temperature
 		for(var/thing in H.bodyparts)
 			var/obj/item/organ/external/E = thing
 			E.internal_bleeding = FALSE
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -219,7 +219,7 @@
 	owner.adjustBruteLoss(-25)
 	owner.adjustFireLoss(-25)
 	owner.remove_CC()
-	owner.bodytemperature = BODYTEMP_NORMAL
+	owner.bodytemperature = owner.dna.species.heat_level_1
 	if(ishuman(owner))
 		var/mob/living/carbon/human/H = owner
 		for(var/thing in H.bodyparts)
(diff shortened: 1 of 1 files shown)


code/datums/status_effects/buffs.dm
```diff
@@ -219,13 +219,15 @@
 	owner.adjustBruteLoss(-25)
 	owner.adjustFireLoss(-25)
 	owner.remove_CC()
-	owner.bodytemperature = BODYTEMP_NORMAL
 	if(ishuman(owner))
 		var/mob/living/carbon/human/H = owner
+		H.bodytemperature = H.dna.species.heat_level_1
```
farie82, line 224: Might need `body_temperature` instead. `heat_level_1` will be the flipping point between hurting and not hurting. While `body_temperature` is the temp the body strives towards and thus normalizes to

farie82 (changes requested):
Be sure to put `fixes` before the issue number
So something like this: `fixes #123456`
This will auto close the issue when this is merged

code/datums/status_effects/buffs.dm
Kiyahitayika: that would make sense

Kiyahitayika: Thanks, @farie82! Change made and committed!

SteelSlayer (approved):

farie82 (approved):
Code is good. Still missing the fixes #XXXX though in the PR description

Balthusdire:
Excellent! It was such a pain to take damage from legion cores as a drask, this is a good change.

Kiyahitayika:
> Excellent! It was such a pain to take damage from legion cores as a drask, this is a good change.

Really happy to hear this feedback! It looks like it was a bit of an oversight. At first I wasn't sure, but after talking to Medium Salad, it became clear.

PR #14559, opened 2020-10-07
Prevents flash / sonic powder from working on blob tiles again.

Qwertytoforty:
## What Does This PR Do

This PR reverts the change of #13899 not in the changelog, that allowed flash powder and sonic powder to affect blob tiles. This moves it back up to the base flashbang code. 

## Why It's Good For The Game
 

Logically, flash powder or sonic powder should damage blobs. Logically. The problem is that this is not balanced. Flashbangs take 5 seconds to arm by default, and if they are on your person when detonating, stun you. Flash powder does not stun the user when detonating, and can be mixed rapidly by an advanced release grenade, or heaven forbid, mixing them rapidly in hand in beakers, decimating entire blobs in seconds. Since major midrounds should not be decimated by 4 chemicals 2 beakers and a bored scientist causing no damage to the station in seconds, it should be changed back to only flashbangs affecting blobs.

## Images of changes

Before this PR: (Flashpowder working) 

https://imgur.com/a/goIU6U6

After this PR: (Flashpowder not working) 

https://imgur.com/a/hqYraQ7

## Changelog
:cl:
tweak: Flash powder and Sonic powder no longer work on blobs.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/flashbang.dm
+++ b/code/game/objects/items/weapons/grenades/flashbang.dm
@@ -17,10 +17,10 @@
 		do_sparks(rand(5, 9), FALSE, src)
 		playsound(T, 'sound/effects/bang.ogg', 100, TRUE)
 		new /obj/effect/dummy/lighting_obj(T, light_color, range + 2, light_power, light_time)
-	// Blob damage
-	for(var/obj/structure/blob/B in hear(range + 1, T))
-		var/damage = round(30 / (get_dist(B, T) + 1))
-		B.take_damage(damage, BURN, "melee", FALSE)
+		// Blob damage
+		for(var/obj/structure/blob/B in hear(range + 1, T))
+			var/damage = round(30 / (get_dist(B, T) + 1))
+			B.take_damage(damage, BURN, "melee", FALSE)
 
 		// Stunning & damaging mechanic
 		bang(T, src, range)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/grenades/flashbang.dm
```diff
@@ -17,6 +17,10 @@
 		do_sparks(rand(5, 9), FALSE, src)
 		playsound(T, 'sound/effects/bang.ogg', 100, TRUE)
 		new /obj/effect/dummy/lighting_obj(T, light_color, range + 2, light_power, light_time)
+	// Blob damage
+	for(var/obj/structure/blob/B in hear(range + 1, T))
+		var/damage = round(30 / (get_dist(B, T) + 1))
+		B.take_damage(damage, BURN, "melee", FALSE)
```
AffectedArc07, line 23:
```suggestion
		// Blob damage
		for(var/obj/structure/blob/B in hear(range + 1, T))
			var/damage = round(30 / (get_dist(B, T) + 1))
			B.take_damage(damage, BURN, "melee", FALSE)
```
Include this in the `if(T)` check since it uses range on the turf itself.

AffectedArc07 (changes requested):

Kyep (approved):

PR #14562, opened 2020-10-08
Code Standarisation - Tank Distribute Pressure

Kiyahitayika:
## What Does This PR Do
Changes the `distribute_pressure` for handheld tanks from `ONE_ATMOSPHERE*O2STANDARD` to `TANK_DEFAULT_RELEASE_PRESSURE`.

Additionally reorganizes the code in `tank_types.dm` to improve flow and reduce separation of related code bits. Gets rid of a section called "Emergency Oxygen" and places its code throughout "Nitrogen", "Oxygen", and "Plasma" respectively.

## Why It's Good For The Game
`TANK_DEFAULT_RELEASE_PRESSURE` is the `distribute_pressure` used for Plasmaman tanks and is set as #define at `16`. This standardises the `distribute_pressure` across all handheld tank types

This makes it so all internals tanks last, by default, as long they can without causing undue harm to their users. It standardises code with the tank a Plasmaman spawns with.

## Changelog
:cl:
tweak: Changes the default distribution pressure for all internals tanks from 21kpa to 16kpa. This allows the gas in internals tanks to last longer. Anesthesic tanks and Air Mix tanks are still at 101kpa as necessary for their functions.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/tanks/tank_types.dm
+++ b/code/game/objects/items/weapons/tanks/tank_types.dm
@@ -2,9 +2,9 @@
  * Contains:
  *		Oxygen
  *		Anesthetic
- *		Air
+ *		Air Mix
+ *		Nitrogen
  *		Plasma
- *		Emergency Oxygen
  */
 
 /*
@@ -16,7 +16,6 @@
 	icon_state = "oxygen"
 	distribute_pressure = ONE_ATMOSPHERE*O2STANDARD
 	dog_fashion = /datum/dog_fashion/back
-
 
 /obj/item/tank/oxygen/New()
 	..()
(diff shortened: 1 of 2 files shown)


SteelSlayer:
Personally I would do this for all internals tanks, *except* anesthetic, which I think needs 100 something kPa. For normal tanks, people only need 16kPa to breathe.

AffectedArc07:
Make the changelog more friendly for people ingame. Not everyone knows what a code path is, so write something like "Tweaked vox nitrogen tank release pressure from this to that"

Sikroth:
Not sure if you're changing all the internals like Steel said but there's also Air Tanks (not O2) which need normal pressure to work, so not all internals use 21/16 kPa. I think the RD has one in their locker, maybe? I know I've seen it somewhere in science, unless it was removed.

Kiyahitayika:
> Not sure if you're changing all the internals like Steel said but there's also Air Tanks (not O2) which need normal pressure to work, so not all internals use 21/16 kPa. I think the RD has one in their locker, maybe? I know I've seen it somewhere in science, unless it was removed.

Air Mix? Sure thing.

code/game/objects/items/weapons/tanks/tank_types.dm
```diff
@@ -251,3 +214,41 @@
...
+
+/*
+ * Anesthetic
+ */
+/obj/item/tank/anesthetic
+	name = "anesthetic tank"
+	desc = "A tank with an N2O/O2 gas mix."
+	icon_state = "anesthetic"
+	item_state = "an_tank"
+	distribute_pressure = ONE_ATMOSPHERE*O2STANDARD
```
variableundefined, line 248: See above.

code/game/objects/items/weapons/tanks/tank_types.dm
```diff
@@ -251,3 +214,41 @@
...
+
+/*
+ * Air Mix
+ */
+/obj/item/tank/air
+	name = "air tank"
+	desc = "Mixed anyone?"
+	icon_state = "air"
+	item_state = "air"
+	distribute_pressure = ONE_ATMOSPHERE*O2STANDARD
```
variableundefined, line 226: It appears the default distribute pressure is in fact, 101 kPa, changing it to 21 kPA will result in choking people to death

variableundefined (changes requested):
Two things. The changelog should be rewritten for better clarity. 

add: Added a distribute_pressure for Anesthetic tanks per @SteelSlayer's suggestion so they still use ONE_ATMOSPHERE * O2 STANDARD at 21kpa
tweak: Changed the default distribution pressure from ONE_ATMOSPHERE * O2 STANDARD at 21kpa to TANK_DEFAULT_DISTRIBUTE_PRESSURE at 16kpa.


- Order of the changelog should be reversed, you cannot understand the first item without the second. 
- Default distribution pressure on its own will confuse people, change it to Default distribution pressure of all tanks. 
- Skip the code macros, no one in changelog will understand nor care about it. They're only for future coder and the compiler. 

Second issue:
The default distribution pressure is in fact ONE ATMOSPHERE, (101 kPa). The changes to Anesthetic and Air Mix will results in choking people to death.

code/game/objects/items/weapons/tanks/tank_types.dm
Kiyahitayika: distribute_pressure = ONE_ATMOSPHERE*O2STANDARD
So that's...101.325 *0.21 =21.27825

So, if we use Air Mix, that's 50/50 Oxygen/Nitrogen. That's 10.639125kpa Oxygen and 10.639125kpa Nitrogen. Great. Everyone's dead. Thanks, maths.

Thank you, working on it.

Kiyahitayika:
Made the change on `tank.dm` for all tank objects instead of individually setting it to the same variable. Fixed Anesthetic and Air Mix!

variableundefined (approved):

PR #14568, opened 2020-10-09
Replace IPC starting box with machine survival box

pmitchell:
## What Does This PR Do
This replaces the starting box for Machines (currently the default O2 survival box) with a machine survival box. The machine survival box contains:
- an emergency welder
- 15 cables
- an emergency flare

## Why It's Good For The Game
IPCs should not start with an O2 kit because they do not breathe. This gives IPCs something relevant to their survival to start with instead.

## Images of changes
box_machine

## Changelog
:cl: ARMA-993
tweak: IPCs spawn with machine survival box
add: added a machine survival box
/:cl:


AffectedArc07:
Not in favour. Giving an IPC three full cable coils and a full welding tool as soon as they join is not great. Yes I know that them having an oxygen kit makes no sense at all, but I aint in favour of giving them this much of a legup at roundstart. In the three **years** I have spent maining IPC, I have never felt this to be a requirement.

SabreML:
I 100% support changing the emergency box for IPCs, since the only useful thing in there is a glowstick.
That said, a full on Robotics Repair Kit is definitely a bit much.

It'd probably be better to create a new type of box specifically for IPCs with an emergency welder and some cables, or something along those lines.

Spacemanspark:
I'd be fine with giving them an emergency welder and maybe a half stack of cables, but this is excessive as it stands now.

matttheficus:
To be honest, I dont even see the need for cables, as you can get them for free nearly everywhere. The emergency welder IS a good idea in my opinion, due to it being such a small amount of fuel anyway.

Carthusia:
an emerg welder and half a stack of cables seems fair

pmitchell:
Okay, I'll make a separate machine survival box

Pennwick:
I'm not so sure on this. Giving a universal kit to all empoyees seems the NT thing to do and you could always ask for these things. Other races don't get healing items in their kits they get something to stop them from taking damage.

It should also probably have the emergency pen. More than half the time its not for you but the guy who can't use theirs because they're collaped in crit.

matttheficus:
> I'm not so sure on this. Giving a universal kit to all empoyees seems the NT thing to do and you could always ask for these things. Other races don't get healing items in their kits they get something to stop them from taking damage.
> 
> It should also probably have the emergency pen. More than half the time its not for you but the guy who can't use theirs because they're collaped in crit.

Vox and Plasmamen have their own personalized boxes, to counter the universal kit point.

On another note, it may be healing, sure, but IPCs literally get smacked by EVERYTHING. They take insanely more damage than almost every other race, and giving them an Emergency Welder (which to be honest, theyll 9/10 go get a normal/experimental welder anyway) that alleviates some of that isnt too big a stretch.

pmitchell:
Honestly, I'd be just fine with the machine survival box having a deck of cards and a fake mustache if that's what makes sense for IPCs - I just want them to have something that doesn't incorrectly suggest they need O2.

Necaladun:
Not having to breathe O2 is already a buff. Giving them self-healing at round start because they don't have to have O2 is just double dipping.

If the lore is a problem, then assume NT gives all IPCs a kit to help their O2 breathing co-workers with.

taukausanake:
Lore hasn't been the problem in years. It's been argued that lore is at the mercy of the code, not the other way around.

IPCs don't have a similar time as fleshies do. All biological beings get an auto injector that helps stabilize them if slowly dying. IPCs have two states: alive and dead. For them, any amount of healing should match up to an auto injector, just not a lot. A short-lived welder and some wires should be fine, unless there is something else that can fix IPCs (I'm thinking nanites or something similar but I'm drawing a blank)

Spacemanspark:
Personally I don't care one way or the other either. It's not hard to get some of these items if cargo has a single braincell.

variableundefined:
Objection on the same design ground as Affected.

Closing PR

PR #14569, opened 2020-10-09
Vampire Rebalancing

Cocacolagua:
## What Does This PR Do
Hypnotize stun ability lasts twenty seconds instead of ten, stuttering lasts thirty seconds instead of ten.
Diseased Touch costs 80 blood instead of 100.
Enthrall costs 220 instead of 300.
Summon bats costs 70 instead of 75, spawns three bats instead of two.
Mist form costs 45 instead of 30.
Rejuvination+ unlocks at 250 blood instead of 200.
Vampires with less than 250 blood total suffer no special effects from holy water.
Vampires between the range of 250-499 blood total suffer the current effects of holy water.
Vampires above 500 total blood suffer more damage from holy water than current holy water.
Vampire full-power heals rapidly while sleeping in coffins.
Holy water deals some brain damage if you have 30-74 units in your system, and deals major brain damage if 75 units or above.
TODO:

- [ ] Rejuvinate+ purges bad chemicals when used, same style as legion cores.
- [X] Vampire full-power makes you weaker to holy-water than a regular vampire, not immune. This should replace the x-ray removal.
- [X] (expansion from above) The stronger a vampire gets, the weaker they get to holy water, the chapel, and whatever. This means low-blood vampires can't be metachecked with holy water.
- [ ] A late-game objective for full-power vampires to thrall people.

## Why It's Good For The Game
So, vampire, suffers from the "happens the same way every shift" curse the most from the three main antagonists, traitor changelings and vampires. 

This happens for two reasons:
1- They always have the same powers and abilities, unlike changelings or traitors.
2- Half of said powers are very bad, to the point nobody uses them.

This PR aims to fix the second point, the shitty powers, while leaving the first point for the futureTM.

Now, let's discuss each change:
Hypnotize: There is no reason to use this, the stun doesn't last anything, and it costs thirty blood to use it. Decided to double the stun time and triple the stutter to make it some-what worthwhile, but even then, it won't get much use when the glare power exists.
Diseased Touch: This is a total meme power, the effect isn't that strong and it costs 100! blood, I lowered it to 50 so you dont spend that much on it.
Enthrall: This change aims for vampires to actually do something else than mist-form and glare spam, instead allowing them to build their thrall army more viably. 300 blood is too much for thralling someone.
Summon bats: Bats are very weak, they die to two lasers/batons, not delaying too much time from security catching up to you, I added an extra bat so it's worth the 70 blood.
Vampire full-power(coffin): After all of these changes, full-power should have it’s upsides, one of which is this, which adds the dynamic of thralls protecting the vampire while they sleep.

Now the nerfs:
Mist form: This is the entire reason why vampire is like this, a very cheap cost for a get-out-of-jail-free card, vampires spam this since the other abilities suck, and ends with each vampire shift being a goose chase until the vampire runs out of blood from the first victim they gave five blood bags.
Rejuv+: I decided to move it to the 250 blood unlock so you get it as the same time as shadow walk, as your vampiric side evolves, your weakness to holy water does too, that's why it starts happening from this point.
Vampire full-power(holywater): Apart from it not making any sense (a vampire fully embracing it's power would probably be weaker to it's silver bullet), it allows for another tool for security to use against them.

And finally, the TODOs:
Rejuvenate: This is for two big reasons, not allowing vampires, the antagonist designed for having a lot of safety nets, be countered by a single ether syringe (ability wakes them up but doesnt remove the chemical, so they fall asleep instantly). Other reason being vampires using methamphetamine is fucking stupid, and should never happen again.
Thrall objective: Late-game vampires don't have much to do, and have security on their tail, so giving them something to do and spend blood on is great.
 
And the change that doesn't fit on the other sections:
Holy water: Change was done so security can't just smoke holy water bomb places with no downsides, small doses are still fine.

## Changelog
:cl:
tweak: Vampire Hypnotizes' stun lasts forty seconds instead of twenty, the stutter lasts a minute intead of twenty seconds.
tweak: Vampire Diseased Touch costs 50 blood instead of 100.
tweak: Vampire Enthrall costs 220 blood instead of 300.
tweak: Vampire Summon Bats costs 70 instead of 75, summons 3 bats instead of 2.
tweak: Vampire Mist Form costs 45 blood instead of 30.
tweak: Vampires below 250 blood total suffer no vampiric downsides from holy water.
tweak: Vampire full-power is no longer immune to holy water, but is very weak against it.
tweak: Vampire full-power heals faster than a low-blood vampire while sleeping in coffins.
tweak: Holy water causes moderate brain damage if you have between 30 to 74 units of it in your system, and major brain damage if you have 75 or more of it.
/:cl:


Cocacolagua:
fuck sent it early, didn't finish writing the changelog

SteelSlayer:
The time on hypnotize is in life cycles btw, so its 20 seconds -> 40 seconds, etc.

S-CR:
I actually think gaining more esoteric weaknesses as you become more "The creature you are" is a neat way to make the counterplay evolve as you gain more power. Vastly prefer that option.

TDSSS:
Good idea in general, vampires badly needed this. I also really like the weakness suggestion.

On the disease touch, I don't think just reducing the cost can make this worth using, it might just need to be reworked or replaced. As for hypnotism, that's a good change but I'm not sure it's enough, considering you could just use some cable cuffs instead.

Cocacolagua:
read some good suggestions, will add to the TODO

matttheficus:
I do think it will be important to make sure holy water cant just be used by Officers and such as a main weapon once vamps are called, especially high-power/full-powered ones. Seeing Security rush RND for RSGs and holy water syringes is gonna be REALLY rough and nigh-uncounterable. Perhaps make vamp's skin "harden" or something to prevent syringes from being insta-death?

Rurik123:
I normally, as a rule, try not to complain about features/mechanics I don't have a better idea for. Better not to nag on a mechanic when I can't come up with a better one.

However, this will be the exception. 

Mist Form needs to go. Its unfun to fight, gets vampires out of bad situations that they put themselves in, and has zero counterplay (no, blessing tiles doesn't count as counterplay). Essentially a get out of jail free card to kite security. Its the only real reason vamps are unfun to fight, and the only real reason fully powered vamps are scary. Take away the mist form and suddenly vamp looks a lot weaker. If we want vamps to be entertaining to fight and not the most powerful antag out of the golden 3 (traitor, cling, vamp) then a change here needs to happen.

I have no idea what the replacement ability would be, nor do I have a good idea on how better to balance it in its current form. I just know the problem from facing it so many times.

McRamon:
Full power vampire shpuld be a serious foe to deal with. Unlike other antags, vampire dont have access to their abilities roundstart and instead earn them. X ray is very powerful, but i recommend not removing it but rather switch to thermals. Afterall, it even makes Sense for vampires to "feel" their victims from behind the walls.

McRamon:
About misting - i am totally against removal or mega nerfing it. That ability Is exactly what makes vampires Fun for me and probably for many others - not because its Jaja op op easy win out of jail card and i like to win. No. Exactly because its easy to escape tool, it allowes me to perform risky but fun and interesting moves, which i would not dare to try as other antags. Sure some people may abuse it becoming omegapowerful antags, but world doesnt revolve around them.

Denthamos:
I do not see the turning thrall into vampire feature being something that would be well liked/approved staff wise due to the impact that could potentially have.  Generally creation of new antagonists outside of very specific things like constructs breaches into the realm of admin only determination and the minor/moderate/major events of the round as we try to control the pace of the round to prevent it from becoming overwhelming against the crew or too underwhelming.

Think of it like we try to play the role of a dungeon master in D&D, if someone else started introducing enemies and threats without us knowing or controlling it, it can derail entire games.

Cocacolagua:
@Denthamos fair enough, giving them an objective to thrall is good enough

Rurik123:
> About misting - i am totally against removal or mega nerfing it. That ability Is exactly what makes vampires Fun for me and probably for many others - not because its Jaja op op easy win out of jail card and i like to win. No. Exactly because its easy to escape tool, it allowes me to perform risky but fun and interesting moves, which i would not dare to try as other antags. Sure some people may abuse it becoming omegapowerful antags, but world doesnt revolve around them.

As a vamp player, I get that. Theres no other antag that I feel as comfortable and at home with as vampire with mist, but you also have to consider the sec's viewpoint—it is simply not fun to chase around something that kites you without ever fighting. Kiting in of itself isn't a bad thing as for most antags, since they need a route of escape. Vamps (as well as wizards, slaughters, and slings) don't.

code/modules/reagents/chemistry/reagents/water.dm
```diff
@@ -301,12 +303,53 @@
 					if(prob(40))
 						M.emote("scream")
 					M.mind.vampire.nullified = max(5, M.mind.vampire.nullified + 2)
+	else if(ishuman(M) && M.mind && M.mind.vampire && M.mind.vampire.get_ability(/datum/vampire_passive/full) && prob(80)) //Dirty copypaste for vampire fullpower, 500 blood to infinity, they take more damage from holy water
```
SteelSlayer, line 306: :angry: `//Dirty copypaste`. I suggest making this section and the one above it, into a proc, with the ability to tweak the severity of the jitter, stutter etc.

code/modules/mob/living/carbon/life.dm
```diff
@@ -345,11 +345,15 @@
 
 /mob/living/carbon/handle_sleeping()
 	if(..())
-		if(mind?.vampire)
-			if(istype(loc, /obj/structure/closet/coffin))
+		if(istype(loc, /obj/structure/closet/coffin))
+			if(mind?.vampire && !mind.vampire.get_ability(/datum/vampire_passive/full))
```
SteelSlayer, line 349: Slightly better organization, avoids some if checks.
```dm
		if(mind?.vampire && istype(loc, /obj/structure/closet/coffin))
			if(!mind.vampire.get_ability(/datum/vampire_passive/full))
				adjustBruteLoss(-1, FALSE)
				adjustFireLoss(-1, FALSE)
				adjustToxLoss(-1)
			else
				adjustBruteLoss(-2, FALSE)
				adjustFireLoss(-2, FALSE)
				adjustToxLoss(-3)
```

Aquilar:
does this mean the average 50 u of holy water will kill cultist during a cult round? maybe add a safety that the brain damage only affects normal crew.

Necaladun:
Vampires need tweaks for sure, the biggest tweak needed to them is mist form imo, @Rurik123 I think makes the case for that very well.

All in all I'm of a "sure lets give it a try" mentality to the rest of the changes here, but Mist Form is the biggest thing needing reworking.

Cocacolagua:
I'll work on this when I can, and will re-open it after it’s done, but I don’t want to clog the open PR lists with this draft.

PR #14572, opened 2020-10-09
No more DIY xeno outbreaks

### LINKED ISSUE #14266: Refresher Crystal Xenos Outbreak?
**Issue Description**:

You can create a xenomorph outbreak with no admin intervention with what is already on station.
Llamar, the RD's pet is harmless by default but apparently if you kill them then and then use SR on the dead body, it comes back to life as a normal facehugger and will hug whomever they can and implant a xenomorph. At least from what I witnessed that is what happened. 
***EDIT*** Looks like it was a miner using a refresher crystal on gateway facehuggers. As I said I didn't witness this first hand so I wasn't sure what happened. But yeah. Still something to note as space ruins and gateways both can have dead face huggers.

I'm not sure if it is intentional or not but it seems like an oversight that has never really been looked at before, most shifts llamar just stays in their display case and never comes out so its not surprising people never really checked.
I would expect llamar to come back as a harmless facehugger and not grow their .. uh, whatever it is that they use to implant a xeno egg back.

**What happened instead**:

They grew it back and became a threat!

Now that this is known people are going to be using it a lot I tihnk. >..> it might become a new "thing" unless it gets handled. Xenos are out of rotation for being broken but being able to just make them on a whim kind of defeats the purpose. though they would have to start with just a single drone so maybe not entirely dangerous

Kill llamar, apply SR, get face hugged, burst?

I have zero idea, probably forever. Llamar is one of the overlooked pets like renault, nobody pays attention to them because they never leave their owners office

**Extra information**:

This is just from observation, I do not have a test server or anything to test it myself, may need some more testing by someone that does.

TDSSS:
## What Does This PR Do
This PR makes it so you can no longer revive dead facehuggers from ruins to create your own xeno outbreak as a non-antag miner, using the refresher crystal.
fixes #14266

## Why It's Good For The Game
Players shouldn't be able to deliberately trigger a major biohazard, especially not this early into the round.  While this does rob hijack antags of a potential route, a hijack antag can always try and ask for a TC trade if he really wants to go for a xeno outbreak and can convince the responding admin.

## Changelog
:cl:
tweak: refresher anomalous crystals can no longer revive facehuggers
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/colossus_loot.dm
+++ b/code/modules/mining/lavaland/loot/colossus_loot.dm
@@ -393,7 +393,8 @@
 	activation_method = "touch"
 	cooldown_add = 50
 	activation_sound = 'sound/magic/timeparadox2.ogg'
-	var/list/banned_items_typecache = list(/obj/item/storage, /obj/item/implant, /obj/item/implanter, /obj/item/disk/nuclear, /obj/item/projectile, /obj/item/spellbook)
+	var/list/banned_items_typecache = list(	/obj/item/storage, /obj/item/implant, /obj/item/implanter, /obj/item/disk/nuclear,
+											/obj/item/projectile, /obj/item/spellbook, /obj/item/clothing/mask/facehugger/)
 
 /obj/machinery/anomalous_crystal/refresher/New()
 	..()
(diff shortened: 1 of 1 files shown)


Kiyahitayika:
@TDSSS It's also going to be using SR on Lamar. Or any dead Xeno bio.

hal9000PR:
This is a good thing as a lucky miner could destroy the station very easily.

TDSSS:
You can't SR Lamar because facehuggers are clothes, not mobs.

farie82 (approved):

procdrone:
As for highjack, it makes little sense. Biohazards are great at NOT having the shuttle called.

Kyep (approved):

AffectedArc07 (approved):

PR #14574, opened 2020-10-10
Changling Swap forms now takes 10 seconds to pull off.

Qwertytoforty:
## What Does This PR Do

This PR makes changeling swap forms take 10 seconds to happen, instead of 2

## Why It's Good For The Game

It's a bit too easy to mindswap into a sec officer. Tentacle to grab to swap, then stun and mutesting target, or just pre remove your headset, or hell, even a water slip into a grab. After you swap, you kill them, hide the body, and you are a mindshielded sec officer. This PR makes it that it now takes 10 seconds to swap forms with someone, meaning you need a better plan than slip grab and swap to swap forms with people, and either trap your target or get the help of a second changeling.

## Changelog
:cl:
tweak: Changelings swap forms now takes 10 seconds to pull off.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/swap_form.dm
+++ b/code/game/gamemodes/changeling/powers/swap_form.dm
@@ -38,7 +38,7 @@
 	target.do_jitter_animation(500)
 	user.do_jitter_animation(500)
 
-	if(!do_mob(user,target,100))
+	if(!do_mob(user, target, 10 SECONDS))
 		to_chat(user, "<span class='warning'>The body swap has been interrupted!</span>")
 		return
 
--- a/code/game/gamemodes/changeling/powers/swap_form.dm
+++ b/code/game/gamemodes/changeling/powers/swap_form.dm
@@ -27,9 +27,6 @@
 	if(target.has_brain_worms() || user.has_brain_worms())
 		to_chat(user, "<span class='warning'>A foreign presence repels us from this body!</span>")
 		return
-	if(ismindshielded(target))
-		to_chat(user, "<span class='warning'>We are unable to bypass the implant protecting their mind!</span>")
-		return
 	return 1
 
 /datum/action/changeling/swap_form/sting_action(var/mob/living/carbon/user)
@@ -41,7 +38,7 @@
 	target.do_jitter_animation(500)
 	user.do_jitter_animation(500)
 
-	if(!do_mob(user,target,20))
+	if(!do_mob(user,target,100))
 		to_chat(user, "<span class='warning'>The body swap has been interrupted!</span>")
 		return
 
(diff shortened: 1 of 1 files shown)


S-CR:
If a changeling can't effectively.. change, what's the point of the antag? Getting an officer alone is rough, getting them in a state for a mindswap without alerting anyone isn't necessarily easy, and there's always the chance you're tested later by one of the many different methods available to folks. That's on top of the fact changeling struggles far more than the other two bread and butter antags. I really don't feel like it deserves the nerf.

Rurik123:
Mindswap needs to change. Its been broken for a while. Removing one of the few ways a mindshielded officer can be made a baddie? Not the way to go about it imo. Officers are already overly trusted as "safe." Datlo, if I recall correctly, tried to nerf and make mindswap take longer (as it bloody well should, 3 or 4 seconds is way to quick) but I don't believe anything ever became of it.

> If a changeling can't effectively.. change, what's the point of the antag? Getting an officer alone is rough, getting them in a state for a mindswap without alerting anyone isn't necessarily easy, and there's always the chance you're tested later by one of the many different methods available to folks. That's on top of the fact changeling struggles far more than the other two bread and butter antags. I really don't feel like it deserves the nerf.

Good points here that I really agree on, but, in the end, as someone who used the tentacle + mindswap I still feel it needs a bit of a nerf to the time it takes. Hell half the time you can pull off a mindswap on someone who is not even restrained. All it takes is a bit of a distraction or for the victim to have their hands off the keyboard for 3 seconds. I think it just needs to take a bit more time for the swap to go through. Something in the 10 second range. That'd be far better than just making mindshielded personnel, once more, almost always guaranteed "safe"

S-CR:
I'm not at all opposed to upping the time, mid-hallway ninja mindswaps are super silly and kind of miss the point of the ability. It's stopping it from working on mindshielded folk that I think would turn this ability from an interesting absorb sidegrade to never used at all.

Qwertytoforty:
Changling swap forms now takes 10 seconds to pull off and works on mindshielded people again.

code/game/gamemodes/changeling/powers/swap_form.dm
```diff
@@ -38,7 +38,7 @@
 	target.do_jitter_animation(500)
 	user.do_jitter_animation(500)
 
-	if(!do_mob(user,target,20))
+	if(!do_mob(user,target,100))
```
SteelSlayer, line 41: Better readability.
```suggestion
	if(!do_mob(user, target, 10 SECONDS))
```

SteelSlayer (approved):

S-CR:
Changed my 👎 to 👍

Vastly prefer this, and I agree slip-n-swap is low effort cheese.

Kyep (approved):

AffectedArc07 (approved):

PR #14576, opened 2020-10-10
Added holostools and actually mapped holo furniture

Adrer:
## What Does This PR Do
Adds holodeck stools and changes z2 holodecks to actually contain holodeck furniture instead of regular furniture so that it can't be deconstructed and harvested.

## Why It's Good For The Game
Prevents people from harvesting infinite materials from holodeck. 

## Changelog
:cl:
fix: holodeck now actually uses holodeck furniture
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/z2.dmm
+++ b/_maps/map_files/cyberiad/z2.dmm
@@ -50,6 +50,16 @@
 	icon_state = "dark"
 	},
 /area/syndicate_mothership)
+"aj" = (
+/obj/structure/rack/holorack,
+/obj/item/clothing/under/dress/dress_saloon,
+/obj/item/clothing/head/hairflower,
+/turf/simulated/floor/holofloor{
+	tag = "icon-cult";
+	icon_state = "cult";
+	dir = 2
+	},
+/area/holodeck/source_theatre)
 "ak" = (
 /obj/structure/window/reinforced{
 	dir = 4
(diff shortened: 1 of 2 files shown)


AffectedArc07:
```
$ tools/travis/check_grep.sh
_maps/map_files/cyberiad/z2.dmm
ERROR: Non-TGM formatted map detected. Please convert it using Map Merger!
The command "tools/travis/check_grep.sh" exited with 1.
```
Please ensure youre using mapmerge

Kyep (commented):
Instead of making subtypes for everything that could ever be used in the holodeck, I wonder why we can't just set the NODECONSTRUCT flag on the relevant items when they're loaded in.
Do we really want /holo subtypes of everything that's ever used in a holodeck room of any map? Maybe we do...?

AffectedArc07:
> Instead of making subtypes for everything that could ever be used in the holodeck, I wonder why we can't just set the NODECONSTRUCT flag on the relevant items when they're loaded in.
> Do we really want /holo subtypes of everything that's ever used in a holodeck room of any map? Maybe we do...?

Map standardisation will make it so we only have to have 1 set of holodecks, and having holodeck specific subtypes is fine for now.

I would also like to point out this is @Adrer's first PR 
image

And making demands to change map template loading code is unreasonable for a first-timer

Adrer:
> Instead of making subtypes for everything that could ever be used in the holodeck, I wonder why we can't just set the NODECONSTRUCT flag on the relevant items when they're loaded in.
> Do we really want /holo subtypes of everything that's ever used in a holodeck room of any map? Maybe we do...?

While I do agree that this most certainly isn't the cleanest solution for this problem, with the current amount of deconstructable objects on holodecks, it might be more effort than it's worth to adjust the map template loading; should there be a significant change to the available holodecks it could be worth looking into though. I'm personally not familiar both with DM and the project to do something like that myself however.

Most of PR is actually just mapping the already existing /holo subtypes onto the holodecks anyhow, I only added the subtype for the stool since it didn't exist yet.

PR #14577, opened 2020-10-10
Makes the vox spikethrower more like a cyborg LMG, fixes bugs with it.

Qwertytoforty:
## What Does This PR Do

This PR makes the vox spike thrower an energy gun instead of a physical gun that attempts to reload itself with ammo but just makes ammo casings on the ground, sort of similar to the cyborg LMG on syndicate borgs. It can not be put in a charger to replenish ammo. Also makes it "recharge" slightly faster. Repaths it to go along with the fact it is no longer an energy gun. It still shoots physical projectiles like before.

## Why It's Good For The Game

Hot buggy messes are not good, like making invisible ammo on the ground, and not actually gaining ammo, making the gun simpler code wise is good. Also makes it recharge slightly faster, as 1 bullet per 30 seconds is not very useful, vs 1 bullet per 20 seconds.

## Changelog
:cl:
tweak: Vox spike thrower is now an energy gun instead of a physical gun.
tweak: Vox spike thrower now recharges 1 shot every 20 seconds instead of 30.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/alien.dm
+++ b/code/modules/projectiles/guns/alien.dm
@@ -8,8 +8,8 @@
 	fire_sound_text = "a strange noise"
 	can_suppress = 0
 	burst_size = 2 // burst has to be stored here
-	can_charge = 0
-	selfcharge = 1
+	can_charge = FALSE
+	selfcharge = TRUE
 	charge_delay = 10
 	restricted_species = list(/datum/species/vox)
 	ammo_type = list(/obj/item/ammo_casing/energy/spike)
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/guns/alien.dm
```diff
@@ -1,59 +1,30 @@
...
 	fire_sound_text = "a strange noise"
-	mag_type = /obj/item/ammo_box/magazine/internal/spikethrower
-	burst_size = 2
-	fire_delay = 3
 	can_suppress = 0
-	var/charge_tick = 0
-	var/charge_delay = 15
+	burst_size = 2 // burst has to be stored here
+	can_charge = 0
+	selfcharge = 1
```
SteelSlayer, line 12:
```suggestion
	can_charge = FALSE
	selfcharge = TRUE
```

SteelSlayer (approved):

variableundefined (approved):
Sound design reasoning, approved for design

variableundefined:
Relabeling, this is not a pure fix and has balance implications too

PR #14580, opened 2020-10-10
[WIP] Nucleation Rework

fludd12:
## What Does This PR Do
Retools Nucleations into something that may actually be able to be played in the current game. They have a number of interesting features which makes them rather distinct from other species -- namely, their association with power and electricity.

Their traits are as follows:
- They cannot eat food.
- They take significantly more brute damage from all sources, but take slightly less burn damage.
- They take extreme damage from electricity.
- They can drain power from anything with a cell, gaining nutrition from doing so.
- They can supply power to anything with a cell or any machine, in the latter case acting as a one-machine APC, losing nutrition from doing so. The nutrition cost of supplying energy is much higher than the nutrition gain of draining it.
- Their maximum health and how much they glow is dependent on how well-fed they are.
- Upon death, they dust, leaving no organs behind. Consequently, they cannot be cloned.
- If they die while particularly charged with energy, they release a burst of lightning, like a tesla coil.

## Why It's Good For The Game
Having races in our code that go entirely unused and unmaintained is not healthy for the codebase, and I felt that nucleations are such an interesting concept that they should be at least considered for a player-facing species. Hopefully, this makes them worth the title.

## Tasks:
- [x] Adjust species tags and traits to be up to date with current code, as well as with the design plan for the rework.
- [x] Make health and light level scale with nutrition.
- [x] Change death behavior depending on nutrition.
- [x] Drain electricity spell, associated behavior.
- [x] Supply electricity spell, associated behavior
- [x] `power_redirect` datum to allow machines to work "off-the-grid."
- [ ] Make nutrition / food consumption work.
- [ ] Drop items when dusting.
- [ ] Test for balance, etc of all above behavior.
- [ ] Add new icons for nucleations.
- [ ] Add new icons for nucleation abilities.
- [ ] Nucleation lore / name scheme.
- [x] Overcome code burnout for the first time in two years.

## Changelog
:cl:
add: Nucleations are now available from species selection! This fragile race of what essentially amounts to electricity elementals can drain or supply energy to devices and machines, but there's no coming back if they die.
/:cl:


TDSSS:
Interesting. Are these planned to be a karma race, or should everyone have access to them?
Also, considering they dust on death, I think they don't really have to take more damage from brute. Being actually unrevivable (not even with SR) is a huge downside.

fludd12:
Uncertain about it at the moment, but it will probably be a karma race due to the fact that such a dangerous-to-play race isn't very enjoyable for new players. Karma works as an effective gate for new players to have to learn the game one unique-race-or-job at a time, and I think that applies very well to this. It's more up to maintainer decision, with regards to that!

As for the dusting-on-death and damage vulnerabilities and such, this is definitely something I'm still looking to balance for now. I've had several people comment on perhaps using less punishing forms of death for these, and I'm considering them, especially since dusting can have several unforeseen consequences (heh, supermatter) on the round depending on round type.

Spacemanspark:
Tbh I kind of wish IPC had the ability to  recharge APCs as well as drain them.

That said, really promising PR.

Kyep (commented):
I like the concept of this, though:
- I'd suggest requiring them to absorb radiation, rather than electricity, for nutrition. That would be more interesting, different from IPCs, and would make more sense (given that shocks kill them). They should be able to use radiation from the SM, from Radium in their blood, or cosmic background radiation from standing on space turfs.

Dashsora:
> I like the concept of this, though:
> 
> * I'd suggest requiring them to absorb radiation, rather than electricity, for nutrition. That would be more interesting, different from IPCs, and would make more sense (given that shocks kill them). They should be able to use radiation from the SM, from Radium in their blood, or cosmic background radiation from standing on space turfs.

Could also use uranium structures for this ! We've had voxboxes and plasmeme caves, can't wait to see the rad-rooms !

Would love to see this come to fruition

fludd12:
I actually like that idea a lot, since it means that the availability for the power that lets them charge things or power machinery will be restricted to if they can obtain a source of radiation, which frequently is an all-or-nothing kind of thing on the station -- either you have a lot available, either through uranium structures, chemistry, or engineering, or nothing at all. This may make me reconsider the balance of nutrition-to-power-supply scale, too, since right now it's rather heavily weighted against the player.

Regarding the absorbing cosmic background radiation, my only concern is that it'll be a bit of a snowflake code thing, being (likely) just a check on life() to see what turf they're on, unless you'd find it more interesting to make space mildly radioactive innately, but that'd affect a lot more than just this species.

taukausanake:
> cosmic background radiation from standing on space turfs

A problem with that though is that no one wants people to have a reason to go outside the station. In addition Nucleations would need to be spaceproof for that to theoretically work as a spacesuit would block that radiation.

Uranium, radium, and SM are all good though. Having a creature that can handle SM naturally would be really nice.

How much of their [barely established lore and biology](https://www.paradisestation.org/wiki/index.php/Nucleation) is going to be kept? I kind of like what's on that page already

fludd12:
To be honest, preferably none -- I intend to resprite them and radically alter that lore and such. My current working thought process is they're kind of like the Erichus Ghosts from starbound, in that they tend to manifest around dense concentrations of supermatter; of course, in our case, they aren't spooky blob things that suck your soul out, they're just crystal skeletons with transparent "skin." Or, at least, they will be if I actually get around to the sprite change.

Spacemanspark:
Novakid race when

taukausanake:
> Novakid race when

Are plasmamemes not enough for you? But yeah, I'm still holding off for Resomi so I won't judge

variableundefined:
This PR:
- Has been WIP for a while
- Has not received explicit head support yet. 

I'm putting an objection on it unless the two issues above are resolved.

fludd12:
😰

AffectedArc07:
PR has 3 objections with no one in favour from either party. I recommend making a thread in the future to discuss all changes before the PR comes up, with something as big as this.

PR #14585, opened 2020-10-11
(Maybe) Fixes spurious failures in pipe icon generation

### LINKED ISSUE #14323: Pipes don't always visually connect right at round start.
**Issue Description**:
Pipes sometimes dont visually connect properly at round start.

The pipes in atmos to not be little blocks

**What happened instead**:
The pipes to be... pipes.

It makes learning atmos a huge pain, as well as just a general eye sore.

It seems to be random, ive seen it a few times with the pipes in the bottom right corner of sci chem but never in atmos, until now.

Maybe a few weeks? haven't exactly kept track, could easily be longer.

**Extra information**:
I did confirm that when they do glitch its not just client side, also here's a screen shot of what i mean. Also only happens to pipes that spawn by default and the pipes work like normal, its just visual. Un-Wrenching the wrenching again does not fix it either.
piped

AffectedArc07:
## What Does This PR Do
This PR (MAYBE) fixes the bug where atmos link icons seemingly dont generate. Its fixable by force-generating all the pipe icons, then force-updating each pipe icon, so that is now part of the `SSair` initialization flow, instead of being left to `/New()`, which has a randomised order.

Fixes #14323 (Hopefully)

This PR is also a perfect example on why relying on /New is a bad idea

## Why It's Good For The Game
Having pipes which link together is a nice thing

## Images of changes
image
*assuming this happens each time*

## Changelog
:cl: AffectedArc07
experiment: Potentially fixed the issue where atmos icons dont link together sometimes
/:cl:

### DIFF
--- a/code/ATMOSPHERICS/atmospherics.dm
+++ b/code/ATMOSPHERICS/atmospherics.dm
@@ -8,7 +8,6 @@ Pipelines and other atmospheric objects combine to form pipe_networks
 Pipes -> Pipelines
 Pipelines + Other Objects -> Pipe network
 */
-GLOBAL_DATUM_INIT(pipe_icon_manager, /datum/pipe_icon_manager, new())
 /obj/machinery/atmospherics
 	anchored = 1
 	layer = GAS_PIPE_HIDDEN_LAYER  //under wires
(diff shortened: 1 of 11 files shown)


SteelSlayer (approved):
Well obviously this is hard to test because it's random, but it looks like `update_underlays()` should work.

DevL-Whats-a-jpg:
I'm assuming this fixes #14323 right? Or at least is thought to.

PR #14586, opened 2020-10-11
[TGUI] Update secure storage interface

nicetoolbox:
## What Does This PR Do
Updates the secure safe UI to tgui. Cleans up the dm code a little, adds timers. Adds an error message variable, so you can show the user a message without passing it via the code input.

## Why It's Good For The Game
Makes the safe easier to use, with larger buttons and clearer feedback.

### After device is emagged
Contents are accessible
emagged
### Message displayed for a short while after being hacked
User opens service panel with screwdriver, uses multitool on safe. 40% chance to reset password.
hacked
### Wrong code of correct length tried
Code length is set to 5 in init, items that extend `/obj/item/storage/secure` can override the length
invalid
### Device is locked
Reset button was pressed while unlocked, this locks the device.
set locked
### Initial state, and after being hacked
setup
### Code entered was to short
short
### Safe is unlocked
User entered the correct digits in order, then pressed E.
unlocked

## Changelog
:cl:
tweak: Update secure storage items to TGUI
/:cl:


AffectedArc07:
I assume this PR will also include the safe's like in the vault? Those are the ones which actually use NanoUI.

nicetoolbox:
Wow I didn't even know there was a safe in the vault... 🤦‍♂️

AffectedArc07:
> Wow I didn't even know there was a safe in the vault... 🤦‍♂️

Those are the ones which are on the list for being converted from NanoUI to TGUI
image
*Ignore the lack of a safe dial, my assets are fucked at the moment*

tgui/packages/tgui/interfaces/SecureSafe.js
```diff
@@ -0,0 +1,112 @@
...
+    message = 'ENTER NEW PASSCODE';
+  }
+
+
+  return (
+    <Window>
+      <Window.Content>
+        <Section>
+
+          <Flex flexDirection="row" alignItems="center" justifyContent="space-between">
```
stylemistake, line 65: @dearmochi check this out, your flex properties on box have single handedly made the Flex component obsolete.

dearmochi: ... Which I think is a good thing **personally** because I don't want to box things over and over for. It gets crowded really quickly when you have flex layouts inside other flex layouts

stylemistake: One of the reasons Flex exists in its current opinionated form, is that it allows easy polyfilling of its functionality for IE8.

dearmochi: Fair enough. I'll admit that I'm far from experienced with web development and polyfilling certainly didn't come across my mind when I added the `Box` properties. I can see how it makes sense to have a separate `Flex` component now.

nicetoolbox: ie8 lol?

AffectedArc07:
Please update within the next 14 days, or this PR will be closed

variableundefined:
I'm closing this PR 4 days earlier, as the last response was over two months ago and this PR is effectively stale

PR #14598, opened 2020-10-12
[TGUI] RCD

SteelSlayer:
## What Does This PR Do
Gives the RCD TGUI. Also:

- Small tweaks to `AccessList.js` to allow for some RCD specific features. This does not affect other things which uses this component like the ID modification computer.
- Replaced the use of `<Grid>` with `<Flex>` in `AccessList.js` because Grid is deprecated.

## Images of changes
nkHt7nB

## Changelog
:cl:
tweak: The RCD now uses TGUI.
/:cl:


procdrone:
I love the preview on how will door look.

Access sorting is good too

AffectedArc07 (changes requested):
First passover.

code/game/objects/items/weapons/RCD.dm
```diff
@@ -108,33 +155,41 @@ GLOBAL_LIST_INIT(rcd_door_types, list(
...
-		SSnanoui.update_uis(src)
+	var/obj/item/rcd_ammo/R = W
+	if((matter + R.ammoamt) > max_matter)
+		to_chat(user, "<span class='notice'>The RCD can't hold any more matter-units.</span>")
+		return
+
+	matter += R.ammoamt
+	if(!user.unEquip(R))
+		to_chat(user, "<span class='warning'>[R] is stuck to your hand!</span>")
+		return
```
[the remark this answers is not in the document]
SteelSlayer: The if statement above would prevent that, but I'm still gonna move this because it makes sense.

code/game/objects/items/weapons/RCD.dm
```diff
@@ -1,42 +1,23 @@
...
 /obj/item/rcd
 	name = "rapid-construction-device (RCD)"
 	desc = "A device used to rapidly build and deconstruct walls, floors and airlocks."
 	icon = 'icons/obj/tools.dmi'
 	icon_state = "rcd"
 	opacity = 0
-	density = 0
-	anchored = 0
+	density = FALSE
+	anchored = FALSE
```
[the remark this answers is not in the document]
SteelSlayer: Yep, they don't need to be here.

code/game/objects/items/weapons/RCD.dm
```diff
@@ -51,24 +29,36 @@ GLOBAL_LIST_INIT(rcd_door_types, list(
...
-	// Stupid shit
-	var/static/allowed_targets = list(/turf, /obj/structure/grille, /obj/structure/window, /obj/structure/lattice, /obj/machinery/door/airlock)
+	/// If this is TRUE, any airlocks that gets built will require only ONE of the checked accesses. If FALSE, it will require ALL of them.
+	var/one_access = TRUE
+	/// Which airlock tab the UI is currently set to display.
+	var/ui_tab = TAB_AIRLOCK_TYPE
+	/// A list of access numbers which have been checked off by the user in the UI.
+	var/list/selected_accesses = list()
+	/// A list of objects that block construction if said object is on the tile the RCD is trying to build on.
+	var/static/list/allowed_targets = list(/turf, /obj/structure/grille, /obj/structure/window, /obj/structure/lattice, /obj/machinery/door/airlock)
```
AffectedArc07, line 55: Calling this `allowed_targets` is a little misleading given its things that block construction. Also, are are sure `/turf` is meant to be in here? It seems like it shouldn't.

SteelSlayer: Ah that's actually a mistake on my part with the comment I wrote, I got it backward. It's a list of things you CAN target.

tgui/packages/tgui/interfaces/RCD.js
```diff
@@ -0,0 +1,258 @@
+import { useBackend } from "../backend";
+import { Button, Section, LabeledList, Box, ProgressBar, Flex, Tabs, Icon } from "../components";
+import { Window } from "../layouts";
+import { ComplexModal, modalOpen } from "./common/ComplexModal";
+import { AccessList } from './common/AccessList';
+import { Fragment } from "inferno";
+
+export const RCD = () => {
```
AffectedArc07, line 8: Put `(props, context)` back here. Yes they arent used, but its standard format.

tgui/packages/tgui/interfaces/RCD.js
```diff
@@ -0,0 +1,258 @@
...
+  );
+};
+
+const ConstructionTypeCheckbox = (props, context) => {
+  const { act, data } = useBackend(context);
+  const { mode_type } = props;
+  return (
+    <Button.Checkbox
+      content={mode_type}
+      checked={data.mode === mode_type ? 1 : 0}
```
AffectedArc07, line 73: Add mode as a var on line 69, since you already did mode_type

tgui/packages/tgui/interfaces/RCD.js
```diff
@@ -0,0 +1,258 @@
...
+const AirlockSettings = (props, context) => {
+  const { data } = useBackend(context);
+  return (
+    <Section
+      title="Airlock Settings"
+      flexBasis="content"
+      height={5.5}>
+      <LabeledList>
+        <LabeledList.Item color="silver" label="Name">
+          {data.door_name}
```
AffectedArc07, line 89: Same here, add a var for `door_name` here

tgui/packages/tgui/interfaces/RCD.js
```diff
@@ -0,0 +1,258 @@
...
+
+const TypesAndAccess = (props, context) => {
+  const { act, data } = useBackend(context);
+  return (
+    <Fragment>
+      <Tabs>
+        <Tabs.Tab
+          content="Airlock Types"
+          icon="cog"
+          selected={data.tab === 1}
```
AffectedArc07, line 110: Same down here, last one ill comment on

AffectedArc07 (changes requested):
Please see my other comments

tgui/packages/tgui/interfaces/RCD.js
SteelSlayer: Do you mean `const { mode } = data;`?

AffectedArc07:
```js
const {
  mode,
  mode_type,
} = data;
```

SteelSlayer: Right but `mode_type` is from props, so don't they have to be separate? i.e
```js
const { mode_type } = props;
const { mode } = data;
```
?

AffectedArc07: Yes, apologies

AffectedArc07 (approved):

PR #14602, opened 2020-10-13
Fixes ERT/borg/AI cameras displaying black screen on monitors

Kyep:
## Changelog
:cl: Kyep
fix: fixed a bug which caused ERT/cyborg/AI cameras to display an empty black screen when selected in a camera console / monitor.
/:cl:


AffectedArc07 (approved):
Should be fine

Coolrune206:
It appears the cameras on Cyborgs do not track them and instead remain in the position the Cyborg was when their camera was clicked. Clicking the Cyborg's camera again will recentre the view on the Cyborg.

PR #14606, opened 2020-10-14
Stops shadowlings dividing by zero

Kyep:
## What Does This PR Do
1) Fixes this runtime:
[2020-10-14T10:59:15] Runtime in spell.dm,325: Division by zero
   proc name: get availability percentage (/obj/effect/proc_holder/spell/proc/get_availability_percentage)
   src: Enthrall (/obj/effect/proc_holder/spell/targeted/click/enthrall)
   src.loc: null
   call stack:
   Enthrall (/obj/effect/proc_holder/spell/targeted/click/enthrall): get availability percentage()
   Enthrall (/datum/action/spell_action): UpdateButtonIcon()
   Enthrall (/obj/effect/proc_holder/spell/targeted/click/enthrall): process(2)
The fix is to check for charge_max==0 and return a value in this case, instead of just letting it divide by a number that might be zero.

1) Fixes this runtime:
[2020-10-14T10:52:22] Runtime in ,: DEBUG: to_chat called with invalid message/target.
   Message: '<span class='warning'>You can't use abilities affecting others while you are traversing between worlds!</span>'
   Target: ''
The fix is to change spell/proc/shadowling_check(...H) to actually use H in all cases instead of trying to use the possibly non-existent usr var in some cases.

## Changelog
:cl: Kyep
fix: fixed some shadowling spells generating runtime errors.
/:cl:

### DIFF
--- a/code/datums/spell.dm
+++ b/code/datums/spell.dm
@@ -322,6 +322,8 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 		if("recharge")
 			if(charge_counter == 0)
 				return 0
+			if(charge_max == 0)
+				return 1
 			return charge_counter / charge_max
 		if("charges")
 			if(charge_counter)
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -1,20 +1,21 @@
 #define EMPOWERED_THRALL_LIMIT 5
 
 /obj/effect/proc_holder/spell/proc/shadowling_check(var/mob/living/carbon/human/H)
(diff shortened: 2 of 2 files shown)


code/datums/spell.dm
```diff
@@ -322,6 +322,8 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 		if("recharge")
 			if(charge_counter == 0)
 				return 0
+			if(charge_max == 0)
+				return 1
```
AffectedArc07, line 326: Shouldnt these returns be `TRUE/FALSE`?

farie82: They are percentage values

AffectedArc07 (approved):

PR #14613, opened 2020-10-15
Tweaks Wiki Button

Kyep:
## What Does This PR Do
1) Changes the behavior of the "Wiki" button in-game.
- Old behavior: it asked you for a query. If you entered nothing, it opened the wiki homepage. If you entered something, it sent that query to the search page on the wiki and took you to the results page.
- New behavior: it asks you if you want to open the wiki in your browser. If you say "yes", it opens the wiki homepage.
2) Fixes the spacing in some of the alert() calls in this file.

## Why It's Good For The Game
The old behavior (asking the user in-game for search terms, which are then passed to the wiki software in the URL) no longer works as of Byond 513.1535, because Byond no longer allows you to open links containing the "&" character, which is needed for our wiki software to parse submissions. While in theory I could try to force our webserver to handle another character (like ";") I'm worried that might break other things (e.g. what if you submit a post/page/etc that contains ;?). I'd rather just link the user directly to the wiki homepage, since the homepage also has a search box on it and they can enter their query there (and, given that many wiki queries are for basic stuff like job info, they may not have to type anything at all as the homepage already has the most-used pages linked on it).

## Changelog
:cl: Kyep
tweak: changed wiki button to open wiki homepage instead of asking for search terms, a workaround for the latest version of Byond not passing URLs correctly. 
/:cl:

### DIFF
--- a/code/controllers/subsystem/changelog.dm
+++ b/code/controllers/subsystem/changelog.dm
@@ -245,7 +245,7 @@ SUBSYSTEM_DEF(changelog)
 				if(config.wikiurl)
 					if(alert("This will open the wiki in your browser. Are you sure?",,"Yes","No")=="No")
 						return
-					usr.client.wiki("") // Blank arg is important here
+					usr.client.wiki()
 				else
 					to_chat(usr, "<span class='danger'>The Wiki URL is not set in the server configuration. Please inform the server host.</span>")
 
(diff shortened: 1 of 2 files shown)


variableundefined (approved):
Approved. I don't see any reason this need to wait 24 hours because this is also a bugfix.

PR #14615, opened 2020-10-15
Refactor IsHandgun to a class variable called can_holster

variableundefined:
## What Does This PR Do
This PR is inspired by the previous PR that made the energy sniper (Which inherit from gun/energy, which is a holsterable type) non-holsterable. 

This refactors the IsHandgun() proc, which is only used ever to check whether something is holsterable, into a class level variable called can_holster, which is checked by the can_holster proc to see if it can in fact, be holstered.

This is much more accurate to how the proc is actually used. We've never used the IsHandgun() proc anywhere to distinguish between a handgun vs rifle form factor and it never mattered in more than one case. 

Ideally the holster should be checking for more than that, but that's for another day 

## Changelog
:cl:
tweak: Code refactor behind the gun holstering code. Report any issues.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -56,6 +56,8 @@
 	var/knife_x_offset = 0
 	var/knife_y_offset = 0
 
+	var/can_holster = TRUE // Can it be holstered?
+
 	var/list/upgrades = list()
 
 	var/ammo_x_offset = 0 //used for positioning ammo count overlay on sprite
@@ -465,9 +467,6 @@
 		chambered.BB.damage *= 5
 
 	process_fire(target, user, 1, params)
-
-/obj/item/gun/proc/isHandgun()
-	return 1
 
 /////////////
 // ZOOMING //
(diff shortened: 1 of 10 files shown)


code/modules/projectiles/gun.dm
```diff
@@ -56,6 +56,8 @@
 	var/knife_x_offset = 0
 	var/knife_y_offset = 0
 
+	var/can_holster = TRUE // Can it be holstered?
```
farie82, line 59: Can you make this in the autodoc format?
```
/// comment here
var/variable
```

farie82 (changes requested):

variableundefined:
Sure, also need to amend a commit because I messed up the name of the commit anyway

code/modules/projectiles/gun.dm
FlattestGuitar: Don't do it. This doesn't need documentation.

Also remove that comment, as it does not improve the readability. The name clearly conveys what this variable is for.

variableundefined:
There we go

SteelSlayer (approved):

farie82 (approved):

AffectedArc07 (approved):

PR #14619, opened 2020-10-15
Fixes 2 kingstons bugs, adds diona to advanced kingstons

### LINKED ISSUE #13058: Advanced Kingstons keeps flipping between tajaran and the intended species
**Issue Description**:
The advanced kingstons virus turns people into tajarans and the actual target species at random, switching back and forth.

For them to only turn into the target species.

**What happened instead**:
They kept flipping between tajarans and the target species.

Virus doesn't work as intended. Also, because of the way we handle species changes, they constantly get a full heal as their species changes.

Give someone advanced kingstons (make sure that the target species isn't tajarans), wait for the disease to get bad enough. watch them keep flipping back and forth.

**Extra information**:
This is probably due to advanced kingstons inheriting from normal kingstons (which only turns you into tajarans, no other species), and calling ..() as the very first line in their stage_act.

Qwertytoforty:
## What Does This PR Do

This pr fixes people being healed from kingstons, by adding retain_damage = TRUE to the transformation part of the virus, allowing admins to use it without randomly healing people. This PR also fixes advanced kingstons from having symptoms of normal and advanced kingstons by making it a separate virus (but acts the same way), as advanced kingstons was coded as a sub type of kingstons, and thus inherited its stage_act.

Also adds diona, a non karma species, to advanced kingstons, since it probably should have all non karma species as potential infections.

The virus continues to be admin only.

Fixes #13058 

## Why It's Good For The Game

Bugs are bad, viruses that are subtypes are bad without propper setup, more species from kingstons is good.

## Changelog
:cl:
tweak: Advanced kingstons can now turn you into a diona.
fix: Kingstons / Advanced kingstons can no longer heal you.
fix: Advanced kingstons no longer improperly also triggers the effects of normal kingstons.
/:cl:

### DIFF
--- a/code/datums/diseases/kingstons.dm
+++ b/code/datums/diseases/kingstons.dm
@@ -43,11 +43,12 @@
 					affected_mob.visible_message("<span class='danger'>[affected_mob]'s form contorts into something more feline!</span>", \
 													"<span class='userdanger'>YOU TURN INTO A TAJARAN!</span>")
 					var/mob/living/carbon/human/catface = affected_mob
-					catface.set_species(/datum/species/tajaran)
+					catface.set_species(/datum/species/tajaran, retain_damage = TRUE)
 
 
-/datum/disease/kingstons/advanced
+/datum/disease/kingstons_advanced //this used to be directly a subtype of kingstons, which sounds nice, but it ment that it would *turn you into a tarjaran always and have normal kingstons stage act* Don't make virusus subtypes unless the base virus does nothing.
 	name = "Advanced Kingstons Syndrome"
+	max_stages = 4
 	spread_text = "Airborne"
 	cure_text = "Plasma"
 	cures = list("plasma")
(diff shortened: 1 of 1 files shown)


code/datums/diseases/kingstons.dm
```diff
@@ -43,11 +43,12 @@
 					affected_mob.visible_message("<span class='danger'>[affected_mob]'s form contorts into something more feline!</span>", \
 													"<span class='userdanger'>YOU TURN INTO A TAJARAN!</span>")
 					var/mob/living/carbon/human/catface = affected_mob
-					catface.set_species(/datum/species/tajaran)
+					catface.set_species(/datum/species/tajaran, retain_damage = TRUE)
 
 
-/datum/disease/kingstons/advanced
+/datum/disease/kingstons_advanced //this used to be directly a subtype of kingstons, which sounds nice, but it ment that it would *turn you into a tarjaran always and have normal kingstons stage act* Don't make virusus subtypes unless the base virus does nothing.
```
SteelSlayer, line 49: Comment nitpick: this a bit misleading. It should say something like "Don't make viruses subtypes unless the base virus is written in a way to handle these subtypes." 

For example, the transformation viruses (and some others) are set up to handle subtypes. They all run the parent stage act, but with some code of their own.

PR #14628, opened 2020-10-16
Stamp People

kugamo:
## What Does This PR Do
This PR adds the ability to stamp people with stamps. They will get a colored mark on them in the location where the stamper stamps them. The marks can be washed off with all the regular washing methods. Examination of the stamped person will show the stamp's unique marks in text. This can only be done to humans/human-like species.

## Why It's Good For The Game
General fun. Collect them all and become the rainbow of ugly stamp marks you were meant to be! Or deny the greytide and mark them for it! You can even approve of beneficial members of your team as head, giving them a sign of your appreciation. Its as harmless as a walk to the showers and brings some more charm to the game.

## Images of changes

KCr2xM2hvr

dreamseeker_4gfCrwF9CF

## TODO
- [x] Try using params instead of obnoxious switch if tree for offseting marks
- [x] Try condensing impact of having a large number of marks on an individual at one time
- [x] General logic placement etc
**Ready for review!**

## Changelog
:cl:
add: Adds the ability to stamp people.
imageadd: Adds marks that display on stamped people.
/:cl:


McRamon:
I can see this as being annoying

variableundefined:
You're not making people out of stamp?

Carthusia:
Does this apply to all races?

kugamo:
> 
> 
> I can see this as being annoying

I suppose it could be, maybe I make it so it takes time to accomplish. That way its at least resistible, otherwise showers are only a short walk to dorms away.

> 
> 
> Does this apply to all races?

all human races.. everything on the characters selection screen.

kenken244:
this is great. I need this

Necaladun (approved):
This is fucking stupid. I love it and can't believe it wasn't already in the game.

code/modules/paperwork/stamps.dm
```diff
@@ -12,6 +12,61 @@
...
+	targetBaseIcon.dir = H.dir												//on their backpack or items they hold in their hand
+	targetBaseIcon.overlays += H.overlays_standing[BODY_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[LIMBS_LAYER]				//Currently this is here and not in update_icons because my only known
+	targetBaseIcon.overlays += H.overlays_standing[UNDERWEAR_LAYER]			//method of turning an image into an icon destroys its directionality
+	targetBaseIcon.overlays += H.overlays_standing[HAIR_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[UNIFORM_LAYER]			//some species' bodies dont take up all the space their uniforms do
+	new_stamp_mark.Blend(getFlatIcon(targetBaseIcon), BLEND_MULTIPLY)		//cut out any parts of the stamp mark that aren't on base human
+
+	var/image/stamp_image = image(new_stamp_mark)
+	stamp_image.text = icon_state
```
SteelSlayer, line 53: Instead of setting this to icon state and then have a giant switch statement later on for human examines, I'm going to suggest a bit of a different approach.

Create a static associative list for stamps, that stores the icon_state to examine text, i.e:
```dm
var/static/list/stamp_examine_list = list(
	"stamp-qm" = "Quartermaster approved"
	"stamp-law" = "Justice Department approved"
	// and so on
)
```
Then here you can do:
```dm
stamp_image.text = stamp_examine_list[icon_state]
```
If icon state is `"stamp-qm"` for example, `stamp_image.text` will be set to `"Quartermaster approved"`. Instead of doing the switch in the human examine proc, you can just do `msg += stamp_image.text`

kugamo: ah so this is what a associative list is, neat stuff

Fox-McCloud:
Adding an entire human icon layer for what is essentially a gimmick is low on my list of acceptability.

kugamo:
> 
> 
> Adding an entire human icon layer for what is essentially a gimmick is low on my list of acceptability.

I suppose there is no real need for a new layer, so looking at the current layer options I moved the stamp mark icons to the MISC_LAYER. @Fox-McCloud

code/modules/clothing/chameleon.dm
```diff
@@ -168,6 +168,11 @@
 
 	target.icon = initial(picked_item.icon)
 
+	if(istype(target, /obj/item/stamp))
+		var/obj/item/stamp/S = target
+		var/obj/item/stamp/P = picked_item
+		S.stamp_color = initial(P.stamp_color)
```
moxian, line 174: do you need to use `initial` here? I think it's better to use `P.stamp_color` instead. `initial` would probably break when chameleoning another chameleon stamp or something...

moxian (commented):
Small-ish things.

code/modules/paperwork/stamps.dm
```diff
@@ -12,6 +12,78 @@
...
+	targetBaseIcon.overlays += H.overlays_standing[BODY_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[LIMBS_LAYER]				//Currently this is here and not in update_icons because my only known
+	targetBaseIcon.overlays += H.overlays_standing[TAIL_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[UNDERWEAR_LAYER]			//method of turning an image into an icon destroys its directionality
+	targetBaseIcon.overlays += H.overlays_standing[HAIR_LAYER]				//	->and it needs to be an icon for the blending step
+	targetBaseIcon.overlays += H.overlays_standing[UNIFORM_LAYER]			//some species' bodies dont take up all the space their uniforms do
+	new_stamp_mark.Blend(getFlatIcon(targetBaseIcon), BLEND_MULTIPLY)		//cut out any parts of the stamp mark that aren't on base human
+
+	var/image/stamp_image = image(new_stamp_mark)
+	stamp_image.text = stamp_examine_list[icon_state]
```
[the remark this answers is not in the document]
moxian: On the second thought you don't even need to keep them coupled?
Just have a list of stamps (descriptions), and a (single!) image of all(!) the stamps on the person (you don't even need to separate different stamps into different images).

code/modules/paperwork/stamps.dm
```diff
@@ -12,6 +12,78 @@
 	item_color = "cargo"
 	pressure_resistance = 2
 	attack_verb = list("stamped")
+	var/stamp_color = "#264715"		//used for stamp marks, based on the item sprites.. not the on paper images(those are very bland)
+	var/static/list/stamp_examine_list = list(
```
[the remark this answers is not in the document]
SteelSlayer: Yeah I agree with this. This is the better option.

code/modules/mob/mob_defines.dm
```diff
@@ -199,3 +199,5 @@
 
 	var/forced_look = null // This can either be a numerical direction or a soft object reference (UID). It makes the mob always face towards the selected thing.
 	var/registered_z
+
+	var/list/ink_marks = list()
```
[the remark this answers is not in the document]
kugamo: right, well said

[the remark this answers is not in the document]
kugamo: Initially I wanted to make it so all mobs could be marked, but I started with just humans so I could get the code working.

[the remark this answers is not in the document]
moxian: Let's start with `ink_marks` on just a human then.

When/if we expand that to all mobs, it would be easy to move this var there.

code/modules/clothing/chameleon.dm
kugamo: I only got it to work by using `initial` was spitting strange error at me otherwise...

code/modules/paperwork/stamps.dm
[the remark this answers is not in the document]
kugamo: This is all real good, I will attempt to recode it this evening. I originally used .text to avoid lists of lists, but what you have suggested here is a great alternative.

DevL-Whats-a-jpg:
> I can see this as being annoying

I think i need to stamp someone with DENIED.

code/modules/clothing/chameleon.dm
kugamo: all the other definitions in chameleon code use `initial` I dont believe this is incorrect code. I havent gotten any problems with extensive testing, includiung making the cham stamp chameleon.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -161,6 +161,23 @@
...
+			stampTypeCounter++
+
+			if(stampTypeCounter == numberOfStampTypes && numberOfStampTypes > 1)
+				msg += "and "
+			msg += I.text
+			if(stampTypeCounter == numberOfStampTypes)
+				msg += ".\n"
+			else
+				msg += ", "
+
```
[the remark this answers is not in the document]
kugamo: as bloated as my current code is, it works. The code you suggested returns `illegal use of list2args()` and places an "and" after every comma. I will attempt reworking this later if I have a chance, but I don't believe the code as is is a travesty or anything.

code/modules/paperwork/stamps.dm
```diff
@@ -12,6 +12,56 @@
...
+		if(NORTH)
+			attackedSide = SOUTH
+		if(SOUTH)
+			attackedSide = NORTH
+		if(EAST)
+			attackedSide = EAST
+		if(WEST)
+			attackedSide = WEST
+
+	var/icon/new_stamp_mark = icon('icons/effects/stamp_marks.dmi', "stamp[rand(1,3)]_[attackedSide]")
```
moxian, line 39: I assume you've considered that already, and decided against it, but.. Do you maybe want to have three stamp icons (`stamp1`, `stamp2`, `stamp3`) in the dmi, each with several directions, and have this line be
```dm
var/icon/new_stamp_mark = icon('icons/effects/stamp_marks.dmi', "stamp[rand(1,3)]", attackedSide)
```
Or is this kind of directionality exactly what you're trying to avoid here (and treat every icon as if it's SOUTH direction)?

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -161,6 +161,23 @@
 	if(wear_id)
 		msg += "[p_they(TRUE)] [p_are()] wearing [bicon(wear_id)] \a [wear_id].\n"
 
+	//Stamp Marks
+	var/numberOfStampTypes = length(stamp_marks_desc)
+	if(numberOfStampTypes)
+		msg += "[p_they(TRUE)] [p_have()] been stamped: "
+		var/stampTypeCounter = 0
+
+		for(var/i in stamp_marks_desc)
```
moxian, line 170: Nit: `i` no longer makes mnemonic sense, given that it's not an `image` anymore (nor is it a loop counter). Consider renaming it to `var/desc` or something

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -5,8 +5,11 @@
 	icon = 'icons/mob/human.dmi'
 	icon_state = "body_m_s"
 	deathgasp_on_death = TRUE
+	var/list/stamp_marks_desc = list()	// stores stamp desc
+	var/image/stamp_marks				// stores image of stamp marks
```
moxian, line 9:
```suggestion
	var/list/stamp_marks_desc = list()	// descriptions of stamps applied to this human; visible on examine.
	var/image/stamp_marks				// combined image of stamp marks
```

code/modules/clothing/chameleon.dm
moxian: Oh, that's because `picked_item` is not an actual item, but rather just a path, I see..!

moxian (approved):
LGTM modulo setting `stamp_marks` as image on `human/New()` (and a couple of very small nits)

code/modules/paperwork/stamps.dm
kugamo: currently I have four states for each varient of the mark, one for each direction. The other directions are empty for the state of the direction its defined as. bizzarely having a single image defined in one orientation and changing the direction in the definition of the image doesn't work, and results in the mark being visible regardless of the dir of the human its on. The only other alternative I know of would be having the sprite in all directions in one state, and cutting out the sprites for the directions it doesnt face, but that seems too costly for just avoiding +3 states in the dmi.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -5,8 +5,11 @@
...
 
 /mob/living/carbon/human/New(loc)
+	stamp_marks = image('icons/effects/stamp_marks.dmi')
```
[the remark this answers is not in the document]
kugamo: Initially I used a list of images, then changed it to a single sprite in the form of a mutable appearance since I no longer needed seperate images for the stamp mark desc in examine. Since then I discovered overlaying images is faster than mutable appearances , so I changed it to an image. 

Although I added this define here in New() because I was getting errors trying to reference .overlays on an un instatiated variable? The overlays list didnt exist for the image here, so to get around this I gave the image an empty icon_state/icon def. This doesnt add anything visible since the `no_name` state is empty in the dmi. The check in `update_misc_effects()` isnt correctly functioning, so I will move to it check the lengeth of the overlays list of the image, and not whether or not it exists.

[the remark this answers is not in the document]
moxian: Use an empty image then `stamp_marks = new /image()`

code/modules/paperwork/stamps.dm
moxian: I see. Makes sense (kinda). I could reproduce this.
And I could work around this by doing something like:

```dm
// Don't even need to specify the direction, if you have them as unidirectional in the .dmi
var/icon/new_stamp_mark_base = icon('icons/effects/stamp_marks.dmi', "stamp[rand(1,3)]", dir=SOUTH) 
// This is an empty four-directional icon.
var/icon/new_stamp_mark = new /icon('icons/effects/stamp_marks.dmi', "stamp1_e")
new_stamp_mark.Insert(new_stamp_mark_base, dir=H.dir)
```
The need for empty icon is unfortunate though.. I tried to create that one on the fly via either `new /icon()` or even
```dm
var/icon/new_stamp_mark = new /icon()
for(dir in GLOB.alldirs)
  new_stamp_mark.Insert(new /icon(), dir=dir)
```
and that even *almost* worked, except when the `H.dir==SOUTH` - that makes the stamp mark somehow visible from all directions (but stamping from non-SOUTH makes it work as expected!)

Either way, I suggest at least clarifying this quirk in the comment (and the fact that each stampX_Y only has Y direction in it).
And if you do end up going the "create empty icon" path, i guess that empty icon should be somewhere near the top, like  `icons/empty.dmi` or `icons/misc/empty.dmi`

kugamo: As ideal was this would be I cannot get it to work, unless the approach I have now is explicitly disallowed I am content with the code as is. I have clarified the way it currently works in code comments.

code/modules/paperwork/stamps.dm
```diff
@@ -12,6 +12,59 @@
...
+	//we will use it so stamp marks cannot be put on backpack or inhand items
+
+	var/mutable_appearance/targetBaseIcon = mutable_appearance()
+	targetBaseIcon.dir = H.dir
+	targetBaseIcon.overlays += H.overlays_standing[BODY_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[LIMBS_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[TAIL_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[UNDERWEAR_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[HAIR_LAYER]
+	targetBaseIcon.overlays += H.overlays_standing[UNIFORM_LAYER]
```
SteelSlayer, line 56: Does one of these include clothes that go over jumpsuits? Like armor, jackets, hardsuits, etc.

kugamo: no, its only uniform.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -1257,11 +1258,13 @@ GLOBAL_LIST_EMPTY(damage_icon_parts)
...
-
+	var/mutable_appearance/standing = mutable_appearance(layer = -MISC_LAYER)
+	if(length(stamp_marks.overlays))
+		standing.overlays += stamp_marks
 	//Begin appending miscellaneous effects.
 	if(eyes_shine())
-		overlays_standing[MISC_LAYER] = get_eye_shine() //Image layer is specified in get_eye_shine() proc as LIGHTING_LAYER + 1.
-
+		standing.overlays += get_eye_shine() //Image layer is specified in get_eye_shine() proc as LIGHTING_LAYER + 1.
+	overlays_standing[MISC_LAYER] = standing
```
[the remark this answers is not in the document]
kugamo: I tested this, because the eyeshine code is after the stamp mark code it is rendered above the stamp marks.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -161,6 +161,23 @@
...
+			stampTypeCounter++
+
+			if(stampTypeCounter == numberOfStampTypes && numberOfStampTypes > 1)
+				msg += "and "
+			msg += stampDescription
+			if(stampTypeCounter == numberOfStampTypes)
+				msg += ".\n"
+			else
+				msg += ", "
+
```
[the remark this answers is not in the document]
kugamo: awesome, thank you. I was having trouble executing joining everything in the list except the last entry, I ought to have looked at `.join`'s accepted arguements. I will implement this.

code/modules/paperwork/stamps.dm
kugamo: Currently as is they will show up in the same place regardless of what you are wearing and regardless if you change or remove clothes. This is to reduce complexity mostly, only uniform layer because I want to avoid floating stamp marks when you change backpacks or exosuits.

 I see an alternative as making a whole parallel code structure to the blood system but for stamp marks. When stamping we would check where the click lands on the human starting from the outer most layer working in, then adding a flag or data for each stamp mark onto the item of clothing or bodypart it does hit. Then render the mark when each layer of human is rendered, having a check for stamp marks when each respective layer is updated on their item/bodypart. Lastly have the stamping force a layer update for where it lands.

But that would be a very large undertaking that would have a much greater impact on the codebase as a whole. And to be honest I don't think I have a good enough grasp on it all to create an acceptable result.

SteelSlayer: A project for you in the future maybe :stuck_out_tongue:

SteelSlayer (approved):
A couple things in here I'm unfamiliar with, but tested locally and It works fine. LGTM.

PR #14629, opened 2020-10-16
Megafauna and Tendril armour rebalancing

hal9000PR:
## What Does This PR Do
This PR nerfs the armour values on, ash drake armour and the champions hardsuit. It also removes the possibility for tendrils to drop the inquisitor's hardsuit. This PR also buffs the H.E.C.K suit.

## Why It's Good For The Game
Currently miners can fairly easily get some of the most powerful armours in the game, in the form of ash drake armour, this is not healthy for the games balance considering how easy ash drake armour is to obtain. A robust miner can kill an ash drake within 30 minutes of round time. This level of challenge is too low for the insane levels of bullet and laser protection that the ash drake armour provides. The two ERT hardsuits that are loot rewards from tendrils are an issue, tendrils are easy to kill and can provide armour that is arguably better than drake armour in all almost all respects. Finally the H.E.C.K suit is extremely underwhelming, currently it has worse armour values than ash drake armour except bullet resistance. This is an issue considering the monumental challenge that is killing Bubblegum and how comparatively easy it is to kill an ash drake. 

## Changelog
:cl:

del: Removed the inquisitors hardsuit and hell water flask from the tendril loot table
tweak: nerfed ash drake armour's melee, bullet, laser and energy resistances
tweak: nerfed champion's hardsuit's melee, bullet, laser and energy resistances
tweak: added ACID_PROOF flag to H.E.C.K suit
tweak: buffed the H.E.C.K suit's laser, energy and bomb resistance 

/:cl:


code/modules/mining/equipment/explorer_gear.dm
```diff
@@ -95,3 +95,13 @@
 	var/mutable_appearance/glass_overlay = mutable_appearance(icon, "hostile_env_glass")
 	glass_overlay.appearance_flags = RESET_COLOR
 	add_overlay(glass_overlay)
+
+/obj/item/clothing/suit/space/hostile_environment/equipped(mob/living/carbon/human/user, slot)
+  if(slot == slot_wear_suit)
+    flags |= NODROP
```
Kyep, line 101: Why make this nodrop when worn? Doesn't that mean that once eqipped, the suit acts are unremovable armor, and potentially prevents anyone removing say your headset underneath it?

Aquilar: wouldn't it also prevent surgery?

hal9000PR: It will not prevent surgery. I made the suit nodrop when worn because it is meant to be a cursed item. I have considered how this would make the wearer KOS in terms of space law, because no stripping and such. However I had not considered how this would affect antags Vs the wearer. Not being able to remove someone's headset would make the suit very powerful against vampires and anyone else who tries to abduct the wearer. There is two things that I could do. I could remove the nodrop when worn on the helmet, so antags can actually take the wearer's headset while still retaining the suits curse. Or I could leave it how it is and make the entire suit nodrop, this might have balancing conserns. However I think these conserns aren't too bad as a pAI can also make the user unabductable.

GentleManners:
And ash drake armour needs a nerf in particular because?

Spacemanspark:
The PR owner gives a good explanation for that

code/modules/mining/equipment/explorer_gear.dm
Kyep: Don't use NODROP for it. There are just too many situations where having NODROP clothing is an advantage, particularly if the clothing also counts as armor/suit.
Instead, make it removable, but have it apply some sort of horrible curse to the user IF it is removed. Like killing off 1-2 of your internal organs.

hal9000PR: how about if it was unremovable but if you have holy water in your system you can remove it?

Kyep: I'd really rather you just give it some other downside that isn't "hard to remove".

hal9000PR: I am thinking of scrapping this part, then in a future pr I might flesh out the heck suit. I am a very inexperienced coder and I'd rather not take on more than I can chew

Pixelzilla:
It seems really silly to remove more loot from tendrils when tendrils are not a guarantee nor are their drops. If you happen to luck out getting a specific drop that is RNG dependent then that's not really a problem. There is also a limited amount of tendrils (and a limited amount of megafauna for that matter!) so it seems silly to remove one of the other possible armors that miners who may not want to go after drakes can get to protect themselves. I dont like that this is forcing a singular playstyle of someone who actively hunts megafauna and ignores/neglects the people who may not want (or be able to) hunt megafauna by removing their small chance of finding hardsuits via tendril drops.

What happens on rounds where there are no drakes or only one drake? What about when bubblegum isn't there? :p

Spacemanspark:
Then you don't get the armor. Game doesn't need to be static every round.

Cocacolagua:
RNG mechanics aren’t immune to balance, the armor removed in this PR has almost the same armor as syndicate elite hardsuits (you know, the one nukies can buy).
hell, ideally instead of having a loot-chest with 1/26 different outcomes, we should have smaller ones with 1/7 or 1/6 coming from different sources like megafauna currently do, but that’s a topic for another day.

Kyep (approved):

Denthamos:
Ignoring the fact the inquisitors hardsuit was originally from lavaland and was ported in first and used for the ERT inquisitorial role and was noted this would cause conflicts when lavaland was eventually merged in but just "we can ignore it till it comes up". 
 Additionally there already is in the code a ERT paranormal armor that is unused and should probably be put into place on either end instead of flat out removing the suit from lavaland.  The same issue exists with the champions hardsuit to a degree as it's sprites were ripped for the cult speedy armor.

Now, I can see nerfing the other resistances on the ash drake and champion armor, bullet/energy/laser and all that but melee however is one I do not agree with considering that is the primary damage source for threats the miners face on lavaland as that way the armor remains consistent with their mining hardsuits and the explorer suits which increase in melee resistance as they upgrade it with goliath plates but not any other of the damage types.

hal9000PR:
Melee damage is the most common damage type on lavaland, however the majority of mobs on lavaland have fully armour piercing melee attacks. Only basic fauna, blood drunk miner, and I think legion (the big one, not sure on that however) don't have fully armour piercing attacks. This means that drakes, colossus and bubblegum (the biggest threats to a miner) completely ignore melee damage resistance, and any damage resistance for that matter. 
This is why I believe fire resistance is the most important stat of armour for lavaland. You can mine/fight fauna through ash storms and it is the only stat that matters against drakes.

Denthamos:
That seems really counterintuitive to the reason why they get armor in the first place if all of the higher fauna/bosses down there bypass armor entirely and makes using the armor for well...armor pointless besides the secondary effects some have and it being a exosuit item slot holder.  Probably something that should be looked at in the future then, because I can understand some special ability attacks piercing armor by a % but not completely bypassing it entirely.

Dashsora:
All of the lavaland armor also fit.. in your pockets. (Ever since there was a PR to resize some of the clothing).... this probably should be fixed as well.

ExusA:
> All of the lavaland armor also fit.. in your pockets. (Ever since there was a PR to resize some of the clothing).... this probably should be fixed as well.

OOPS... that was me, i knew i'd have missed something with that pr :P sorry bout that!

Kyep:
This really needs to be merged.

Spacemanspark:
Are we merging this soon

farie82 (approved):
image
Remove the striped through there since it is no longer a change

hal9000PR:
> image
> Remove the striped through there since it is no longer a change

done

PR #14630, opened 2020-10-16
Fixes spell buttons flickering and other issues. Also makes the recharge text more readable

### LINKED ISSUE #14540: Spell recharge effect for icons is broken
**Issue Description**:

After spells go on cooldown, instead of lighting up gradually, and displaying the percent its recharged so far, it just flickers over and over until it comes back off cooldown. This happens with ALL spells from what I can tell.

OfpyWwuA39

**What did you expect to happen?**:
For the spell icons to behave like how they were tweaked in #13200.

Click on any spell icon. Observe.

Very recently. Possibly as a result of #13621 being merged.

farie82:
## What Does This PR Do
Fixes a few bugs with spell buttons:
- They don't flicker anymore
- An unavailable spell like hypnotise for vamps with not enough blood now won't show 100% below it while it's unavailable.
- Using a spell that fails and gets reverted now properly reverts the image

The PR also made the button text better to read. The text itself is now always fully visible. No weird alpha is being applied or such.

Fixes #14540

## Why It's Good For The Game
Bug fixes are very good

## Images of changes
ezgif-4-2458876dcb19

## Changelog
:cl:
fix: Spell buttons now don't flicker anymore
fix: Unavailable spells due to other reasons than recharging now will be blacked out without text
fix: Using a spell that fails and reverts now properly shows the button
tweak: The spell button recharge text is now better readable
/:cl:

### DIFF
--- a/code/datums/action.dm
+++ b/code/datums/action.dm
@@ -16,8 +16,6 @@
 	var/icon_icon = 'icons/mob/actions/actions.dmi'
 	var/button_icon_state = "default"
 	var/mob/owner
-	/// Whether to darken the button or not if [/datum/action/proc/IsAvailable] return FALSE.
-	var/darken_when_unavailable = TRUE
 
 /datum/action/New(var/Target)
 	target = Target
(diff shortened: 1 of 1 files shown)


code/datums/action.dm
```diff
@@ -515,26 +513,28 @@
...
+
+	var/alpha = 220 - 140 * progress
+
+	var/image/img = image('icons/mob/screen_white.dmi', icon_state = "template")
+	img.alpha = alpha
+	img.appearance_flags = RESET_COLOR | RESET_ALPHA
+	img.color = "#000000"
+	img.plane = FLOAT_PLANE + 1
+	button.add_overlay(img)
+	// Make a holder for the charge text
```
moxian, line 532: It's clear that you're making the holder for the text. It's not clear why you need the holder at all, and why can't you slap that text onto `img`.
Clarify the rationale in the comment (i.e. that the alpha shenanigans make the text unreadable, or something)

code/datums/action.dm
```diff
@@ -515,26 +513,28 @@
...
+	var/image/count_down_holder = image('icons/effects/effects.dmi', icon_state = "nothing")
+	count_down_holder.plane = FLOAT_PLANE + 1.1
```
moxian, line 534:
```suggestion
	count_down_holder.plane = FLOAT_PLANE + 2
```
From the docs:
```
Possible values:
-10000 to 10000 (integers only), FLOAT_PLANE+p
````
First of all, I have no idea how this works with a float (does it get treated as 1 or as 2?..), but even if it works as you'd expect, you still have 20k values to work with, no need to be stingy here.

variableundefined (approved):

code/datums/action.dm
farie82: I used it here to ensure it was just above the other image generated above. If say an admin or the game itself wants to apply a film for all players to see on a higher plane then they can more easily do it. A `FLOAT_PLANE + 2` will ensure that it would override this text atleast. They would most likely still use a higher value though

code/datums/action.dm
farie82: Fair point. Will add it when the inevitable third PR comes around to change spells :P

PR #14635, opened 2020-10-17
Make floor painter UI nicer and move it to TGUI

moxian:
## What Does This PR Do
Makes the floor painter UI nicer by showing all the floor directions in a nice and digestable way.
Also moves it to TGUI .

Also, also it tweaks the Dropdown TGUI React component, so that the chosen value is updated when the `selected` property is changed. It used to store the selection in its `state` and only be updatable via `setSelected()` method, which made it very difficult to drive updates from the BYOND backend (and, according to https://reactjs.org/docs/refs-and-the-dom.html, also unidiomatic).

P.S.: I have to mention that my visual design skills are atrocious, and any kind of HTML/CSS/layouting criticism is absolutely welcome.

## Why It's Good For The Game
Makes it soooo much easier to use when trying to paint anything other than a solid fill. 

## Images of changes
Old:

 dreamseeker-20201017-165804

New:

dreamseeker-20201018-174835
dreamseeker-20201018-174846

## Changelog
:cl:
tweak: floor painter UI is now nicer
/:cl:


moxian:
Merge conflict from https://github.com/ParadiseSS13/Paradise/pull/14636 which was created 30 minutes after this PR, and speedmerged.
Please deconflict at your own pace.

PacifistDalek:
Yes, yes, a thousand times yes

tgui/packages/tgui/interfaces/FloorPainter.js
```diff
@@ -0,0 +1,71 @@
...
+    selectedDir,
+    directionsPreview,
+  } = data;
+  return (
+    <Window resizable>
+      <Window.Content scrollable>
+        <Section title="Floor painter">
+          <LabeledList>
+            <LabeledList.Item label="style">
+              <Flex>
```
dearmochi, line 19: Not sure this is doing anything. There are no `<Flex.Item>`s

dearmochi (approved):
JS side looks fine

procdrone:
nice

moxian:
Addressed the comments, thanks!
Per discussion in discord added a nice visual preview of all available styles. Updated OP with the new screenshots.
It now takes
```
  33.064 client rendered in 109ms (6.54 frames)
  33.064 client fully loaded in 292ms (17.52 frames)
```
this long on initial load, it subjectively feels very acceptable (just a very small hiccup), although I'm not sure how it'll behave under high load. Likely fine.

In the light of the above, the js is now very different. Please take another look.

SteelSlayer:
Have you considered caching your list that has the base64 icon previews it in, so you don't have to loop through the list and do `icon2base64(floor_icon)` every time someone opens the UI? I think this should improve your loading times. Granted, it's not a huge issue but, something to explore if you wanted to optimize it.

moxian:
I have, but this does not help much.
Simply caching the `icon2base64(floor_icon)` would make it a bit easier for the server, but does nearly nothing for the user lag, since that one is driven by needing to receive all that base64-encoded data (and then lay it out). At least that's what I'm seeing from my testing now.

If we wanted to make it faster, we'd need to go all the way as to create a new resource, similar to how Vending ~~does it~~ used to do it before it got converted to TGUI, but that's honestly way more effort (and code! And potential bugs!) than I think is worth.

moxian:
P.s.: Speaking of Vending - opening AutoDrobe (the vendomat that sells clothes, near arrivals) suffers the same problem, and is actually much more slower (subjectively, at least) than what floor painter experience, Even though it's cached on the server.
Again, it was nippier before TGUI because it was sending a single mosaic of all the products combined into a single png, and not a json of base64 of the contents.

moxian:
P.p.s: It's very very possible (and likely) that our Vending never had the item preview, and I'm confusing this with /tg/station. See their [code/modules/asset_cache/asset_list_items.dm](https://github.com/tgstation/tgstation/blob/7aec5d43b8d70dc947c21b9dbefbd4c8315d1482/code/modules/asset_cache/asset_list_items.dm) for the unholy hackery that you need to invoke to get the performance gains

SteelSlayer (approved):
Oh well, the loading time isn't that big of a deal. LGTM

PR #14637, opened 2020-10-17
Make can_inject deterministic; fix odysseus syringes sometimes not working [PR #12858, take two]

### LINKED ISSUE #12857: Odysseus syringe gun does not work properly

moxian:
I can't reopen #12858 , so filing this a new one.
Copy-pasting description:

## What Does This PR Do
Fixes #12857 by passing in targeting_zone that the mecha occupant aims at.
The problem here is that when neither targeting zone nor the user are passed to `can_inject()` , it picks a random zone (which is weird, since one would expect `can_inject()` to be deterministic), and many of those random choices are not really zones (`"left leg"` is not a zone, but `"l_leg"` is).
Anyhow, now all callers to `can_inject` pass in either non-null user or properly pass in targeting zone (or both).

Also added test for #12857  . It probably doesn't add much value at this point, but I think this time is as good as any to start building up a test suite.

Separately, my previous pr of #12697 was overly cautions with not closing the bugs it fixed, so I'll do it here:
Fixes: #12685
Fixes: #7132

## Why It's Good For The Game
Bugfix good
Automated testing good

## Images of the changes
B8g9ZeWIaH

## Changelog
:cl:
fix: Odysseus syringes now inject the victim consistently
/:cl:


moxian:
Addressing @Kyep's comment https://github.com/ParadiseSS13/Paradise/pull/12858#pullrequestreview-404498623:

Spiders are not broken by the change; added test to ensure that to be the case.
Merge conflicts resolved.

variableundefined (commented):
Not sure about leaving so much testing code in relative to actual production code - It is likely to sit there unmaintained after a while. 

I'll test this out and then ask to remove.

moxian:
The intent of the testing code is to (eventually) be integrated into CI (travis or what have you), and be checked on every PR, to make sure that new changes don't break old edge-cases.
I admit that *these specific* tests are unlikely to be broken by future code, but I still think it's a good idea to set up CI "properly", and this (I hope) gives us some direction of doing so.

variableundefined (changes requested):
Affected has requested on Discord that this should be converted over to a unit test or removed.

Kyep:
When I asked for a test to make sure this doesn't break spiders, I meant I wanted you to test and confirm it.
I wasn't asking for you to make an automated test.

moxian:
I know. I did both, mostly because I could.

variableundefined:
Please implemented requested change by tommorow.

moxian:
Removed the tests.

variableundefined (approved):

PR #14640, opened 2020-10-17
ERT Commander Rework

Kyep:
## What Does This PR Do
ERT commanders no longer get an ion rifle.
ERT commander headsets (and admin CC officer headsets) now work even if telecomms is down.
Red/Gamma ERT commanders now get two new items, to replace the ion rifle they lost: 
- An omni door remote, which works like the Captain's door remote, but on any station airlock.
- An ERT camera monitor, which works like a camera bug, except that it only shows camera feeds from ERT hardsuit cams and the ERT starting area. It does NOT show normal station cams.

## Why this is good for the game.
The ion rifle is a liability. ERT commanders, right now, do more damage to their own team with the ion rifle than they do to enemies. There's no reason for them to get an ion rifle - if they need one they can already get one from the armory or R&D.

ERT commanders need the ability to issue orders to their team, especially in emergencies like downed comms. So, tweaking their headset this way makes sense.

Giving ERT commanders the ability to remotely monitor their team, and support them by opening bolted doors for them (or bolting doors for them to help lock down an area) gives them not just the ability to give informed orders, but also perform a support role that makes their team more effective.

Effectively, this rework transforms the ERT commander from a role that's mainly a liability (due to their ion rifle) into a role that's actually useful in keeping an eye on, commanding, and assisting their team.

## Changelog
:cl: Kyep
tweak: reworked ERT commander gear. They no longer get an ion rifle. Their headset now works even if telecomms is down. Red/Gamma ERT commanders also get an ERT camera monitor and an omni door remote.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/headset.dm
+++ b/code/game/objects/items/devices/radio/headset.dm
@@ -289,19 +289,25 @@
 	freqlock = TRUE
 
 /obj/item/radio/headset/ert/alt
-	name = "\proper emergency response team's bowman headset"
+	name = "emergency response team's bowman headset"
 	desc = "The headset of the boss. Protects ears from flashbangs."
 	flags = EARBANGPROTECT
 	icon_state = "com_headset_alt"
 	item_state = "com_headset_alt"
 
+/obj/item/radio/headset/ert/alt/commander
+	name = "ERT commander's bowman headset"
+	desc = "The headset of the boss. Protects ears from flashbangs. Can transmit even if telecomms are down."
+	requires_tcomms = FALSE
+
 /obj/item/radio/headset/centcom
 	name = "\proper centcom officer's bowman headset"
-	desc = "The headset of final authority. Protects ears from flashbangs."
+	desc = "The headset of final authority. Protects ears from flashbangs. Can transmit even if telecomms are down."
 	flags = EARBANGPROTECT
 	icon_state = "com_headset_alt"
 	item_state = "com_headset_alt"
 	ks2type = /obj/item/encryptionkey/centcom
+	requires_tcomms = FALSE
 
 /obj/item/radio/headset/heads/ai_integrated //No need to care about icons, it should be hidden inside the AI anyway.
 	name = "\improper AI subspace transceiver"
--- a/code/game/objects/items/devices/radio/headset.dm
+++ b/code/game/objects/items/devices/radio/headset.dm
@@ -299,6 +299,7 @@
 	name = "ERT commander's bowman headset"
 	desc = "The headset of the boss. Protects ears from flashbangs. Can transmit even if telecomms are down."
 	requires_tcomms = FALSE
+	instant = TRUE
 
 /obj/item/radio/headset/centcom
 	name = "\proper centcom officer's bowman headset"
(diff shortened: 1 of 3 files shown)


code/game/objects/items/devices/radio/headset.dm
```diff
@@ -289,19 +289,25 @@
...
+
 /obj/item/radio/headset/centcom
 	name = "\proper centcom officer's bowman headset"
-	desc = "The headset of final authority. Protects ears from flashbangs."
+	desc = "The headset of final authority. Protects ears from flashbangs. Can transmit even if telecomms are down."
 	flags = EARBANGPROTECT
 	icon_state = "com_headset_alt"
 	item_state = "com_headset_alt"
 	ks2type = /obj/item/encryptionkey/centcom
+	requires_tcomms = FALSE
```
AffectedArc07, line 310: If youre doing this tack `instant = TRUE` too, otherwise ERT commanders will have a 2 second delay at times. Unless you want that, in which case, carry on.
```dm
	if(connection)
		for(var/obj/machinery/tcomms/core/C in GLOB.tcomms_machines)
			if(C.handle_message(tcm))
				handled = TRUE
				qdel(tcm) // Delete the message datum
				return TRUE

	// If we dont need tcomms and we have no connection
	if(!requires_tcomms && !handled)
		// If they dont need tcomms for their signal, set the type to intercoms
		tcm.data = SIGNALTYPE_INTERCOM_SBR
		tcm.zlevels = list(position.z)
		if(!instant)
			// Simulate two seconds of lag
			addtimer(CALLBACK(GLOBAL_PROC, .proc/broadcast_message, tcm), 20)
			QDEL_IN(tcm, 20)
		else
			// Nukeops + Deathsquad headsets are instant and should work the same, whether there is comms or not
			broadcast_message(tcm)
			qdel(tcm) // Delete the message datum
		return TRUE
```

procdrone:
I love this.

moxian:
(Didn't look at the code.) Can the ERT headset now broadcast on [Common] if telecomms are down? If yes, it feels.. weird..  Would it be feasible to only make [ERT] frequency telecomms-less?

AffectedArc07:
> (Didn't look at the code.) Can the ERT headset now broadcast on [Common] if telecomms are down? If yes, it feels.. weird.. Would it be feasible to only make [ERT] _frequency_ (as opposed to the headset) telecomms-less?

Not doable in the way tcomms are done, but it will only broadcast to intercoms, not headsets.

moxian:
I assume that means they cannot realistically broadcast on [Command]/[Security]/etc (unless the intercoms were specifically set up for that beforehand). If so, then I'm happy with this.

Necaladun:
Giving the commander something focused on commanding seems a lot more appropriate than a random ion rifle.

Necaladun (approved):
Oh wait this is how I'm meant to say I approve of things now

Kyep:
@moxian 
- This PR only changes ERT commander gear. It does not change sec/eng/med/etc ERT gear.
- ERT commander headsets work completely normally even if telecomms if off. This includes the ability to issue orders on say, security comms. All of this is intended.
Even after this PR, ERT commanders are still going to be a niche role. I'm not sure this is enough to make them truly useful... but at least they won't be a liability, and in the right hands they might potentially be useful.

PR #14648, opened 2020-10-18
TGUI Mirrors + CMA

AffectedArc07:
## What Does This PR Do
Upgrades mirrors and admin CMA to TGUI

## Why It's Good For The Game
TGUI Good NanoUI Bad.

## Images of changes
Mirror view
image
*Ignore the trade; part, that has been fixed*

Admin view
image

## Changelog
:cl: AffectedArc07
add: Mirrors now use TGUI
add: Admin CMA now uses TGUI
/:cl:


Asmerath:
this looks exactly like what we have now

Spacemanspark:
The purpose is to convert the interface to TGUI from the current NanoUI system, not to revamp how it looks @asmerath

Asmerath:
ah, gotcha

AffectedArc07:
@Asmerath I was half tempted to convert it to use sections at the same time, but it just felt too bloated, and as spark said, the goal here is to move it off of the NanoUI backend and into the TGUI backend, not to revamp it.

tgui/packages/tgui/interfaces/AppearanceChanger.js
```diff
@@ -0,0 +1,310 @@
...
+                  selected={s.specimen === specimen}
+                  onClick={
+                    () => act('race', { race: s.specimen })
+                  }
+                />
+              ))}
+            </LabeledList.Item>
+          )}
+          {!!change_gender && (
+            <LabeledList.Item label="Gender">
```
farie82, line 76: Nitpick. Can you turn these LabledLists into their own method (class... thing not sure how it's called in js)?
Makes it a bit more readable.

code/modules/tgui/modules/_base.dm
```diff
@@ -10,8 +10,8 @@ Code is pretty much ripped verbatim from nano modules, but with un-needed stuff
 	var/name
 	var/datum/host
 
-/datum/tgui_module/New(var/host)
-	src.host = host
+/datum/tgui_module/New(datum/_host)
+	host = _host
```
AffectedArc07, line 14: This is just another fix I wanted to do to clear out an implied `src.`

Spacemanspark:
So when are we getting CMA runechat

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+		data["hair_styles"] = hair_styles
+		data["hair_style"] = (head_organ ? head_organ.h_style : "Skinhead")
+
+	data["change_facial_hair"] = can_change(APPEARANCE_FACIAL_HAIR)
+	if(data["change_facial_hair"])
+		var/list/facial_hair_styles = list()
+		for(var/facial_hair_style in valid_facial_hairstyles)
+			facial_hair_styles += list(list("facialhairstyle" = facial_hair_style))
+		data["facial_hair_styles"] = facial_hair_styles
+		data["facial_hair_style"] = (head_organ ? head_organ.f_style : "Shaved")
```
farie82, line 225: can be without ()s

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+		data["head_accessory_styles"] = head_accessory_styles
+		data["head_accessory_style"] = (head_organ ? head_organ.ha_style : "None")
+
+	data["change_hair"] = can_change(APPEARANCE_HAIR)
+	if(data["change_hair"])
+		var/list/hair_styles = list()
+		for(var/hair_style in valid_hairstyles)
+			hair_styles += list(list("hairstyle" = hair_style))
+		data["hair_styles"] = hair_styles
+		data["hair_style"] = (head_organ ? head_organ.h_style : "Skinhead")
```
farie82, line 217: can be without ()s

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+	data["change_skin_tone"] = can_change_skin_tone()
+	data["change_skin_color"] = can_change_skin_color()
+	data["change_eye_color"] = can_change(APPEARANCE_EYE_COLOR)
+	data["change_head_accessory"] = can_change_head_accessory()
+	if(data["change_head_accessory"])
+		var/list/head_accessory_styles = list()
+		for(var/head_accessory_style in valid_head_accessories)
+			head_accessory_styles += list(list("headaccessorystyle" = head_accessory_style))
+		data["head_accessory_styles"] = head_accessory_styles
+		data["head_accessory_style"] = (head_organ ? head_organ.ha_style : "None")
```
farie82, line 209: can be without ()s

tgui/packages/tgui/interfaces/AppearanceChanger.js
```diff
@@ -0,0 +1,310 @@
...
+                />
+              )}
+              {!!change_tail_marking_color && (
+                <Button
+                  content="Change tail marking color"
+                  onClick={
+                    () => act('tail_marking_color')
+                  }
+                />
+              )}
```
dearmochi, line 191: I'd probably simplify that with a list of Objects containing the name of the key in `data` to check, the label name and the action.

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+					if(isnum(new_s_tone) && (!..()))
+						new_s_tone = max(min(round(new_s_tone), length(owner.dna.species.icon_skin_tones)), 1)
+
+				if(new_s_tone)
+					owner.change_skin_tone(new_s_tone)
+
+		if("skin_color")
+			if(can_change_skin_color())
+				var/new_skin = input(usr, "Choose your character's skin colour: ", "Skin Color", owner.skin_colour) as color|null
+				if(new_skin && (!..()) && owner.change_skin_color(new_skin))
```
farie82, line 77: Why the `!..()` here, above and below?
`..()` has already returned `FALSE` on line 29

tgui/packages/tgui/interfaces/AppearanceChanger.js
```diff
@@ -0,0 +1,310 @@
...
+                <Button
+                  key={s.altheadstyle}
+                  content={s.altheadstyle}
+                  selected={s.altheadstyle === alt_head_style}
+                  onClick={
+                    () => act('alt_head', { alt_head: s.altheadstyle })
+                  }
+                />
+              ))}
+            </LabeledList.Item>
```
dearmochi, line 304: Ditto

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+			if(can_change_skin_tone())
+				var/new_s_tone = null
+				if(owner.dna.species.bodyflags & HAS_SKIN_TONE)
+					new_s_tone = input(usr, "Choose your character's skin tone:\n(Light 1 - 220 Dark)", "Skin Tone", owner.s_tone) as num|null
+					if(isnum(new_s_tone) && (!..()))
+						new_s_tone = 35 - max(min(round(new_s_tone), 220),1)
+				else if(owner.dna.species.bodyflags & HAS_ICON_SKIN_TONE)
+					var/const/MAX_LINE_ENTRIES = 4
+					var/prompt = "Choose your character's skin tone: 1-[length(owner.dna.species.icon_skin_tones)]\n("
+					for(var/i = 1 to length(owner.dna.species.icon_skin_tones))
```
farie82, line 59:
```suggestion
					for(var/i in 1 to length(owner.dna.species.icon_skin_tones))
```

dearmochi (changes requested):
Need to get rid of the duplicate code in the JS side.

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+	blacklist = species_blacklist
+
+/datum/tgui_module/appearance_changer/tgui_act(action, list/params)
+	if(..())
+		return
+
+	. = TRUE
+
+	switch(action)
+
```
farie82, line 35: Enter can be left out here I feel

code/modules/tgui/modules/appearance_changer.dm
```diff
@@ -0,0 +1,356 @@
...
+	data["change_body_marking_color"] = can_change_markings("body")
+	data["change_tail_marking_color"] = can_change_markings("tail")
+
+	return data
+
+/datum/tgui_module/appearance_changer/proc/update_dna()
+	if(owner && (flags & APPEARANCE_UPDATE_DNA))
+		owner.update_dna()
+
+/datum/tgui_module/appearance_changer/proc/can_change(var/flag)
```
farie82, line 285: var

code/modules/tgui/modules/_base.dm
farie82, line 14: I know src is bad. But this will just make the params a bit less standard so to say. Can we either make this the coding default or allow an exception for use of src in Inits/News? Quite some language standards make the exception here as well

farie82 (changes requested):

tgui/packages/tgui/interfaces/AppearanceChanger.js
AffectedArc07: I cant think of a pheasible way to do this, but youre welcome to improve it later

tgui/packages/tgui/interfaces/AppearanceChanger.js
AffectedArc07: I cant think of a pheasible way to do this, but youre welcome to improve it later

tgui/packages/tgui/interfaces/AppearanceChanger.js
AffectedArc07: I cant think of a pheasible way to do this, but youre welcome to improve it later

code/modules/tgui/modules/appearance_changer.dm
AffectedArc07: You can open the input, walk away, then set things by queueing up the inputs. The check for this again prevents that.

code/modules/tgui/modules/_base.dm
AffectedArc07: Not relevant to this pull request

procdrone:
If you ever plan to overhaul it, you could go with the RCD approach, sections on the left with content menus on the right.

AffectedArc07:
I compressed the colour stuff down so its less repetitive. The other options cant really be compressed down due to the weirdness with how `act()` handles params.

This is now ready for merge.

dearmochi (approved):

Kiyahitayika:
@AffectedArc07 
Being an Admin on the server you run this code on allows you access to the tail type selector for races with fixed tail types (Unathi, Vox) and races without tails.
image
That could be changed (I remember where the code is when I was testing something things). From what I recall, it does not show up for non-Admin players. Pointing this out just in case.

AffectedArc07:
> Being an Admin on the server you run this code on allows you access to the tail type selector for races with fixed tail types (Unathi, Vox) and races without tails.
> image
> That could be changed (I remember where the code is when I was testing something things). From what I recall, it does not show up for non-Admin players. Pointing this out just in case.
> 
> EDIT: But where is Snake Tail? ;(

I didnt change functionality of what displays and what doesnt, as far as I know. As for the snake tail, we nuked that like 4 months ago.

PR #14655, opened 2020-10-19
Make conveyor belts cinematic

moxian:
## What Does This PR Do
Makes it possible to slow down conveyor belts up to 50x, with a multitool.

## Why It's Good For The Game
Sushi belts become usable, you can actually grab food from them under no pressure.

## Images of changes
Left to right: default speed (same as live), 5x slow, 20x slower.
xvxcYXMpNi

TGUI Menu:
XiSIOcGKrW

## Changelog
:cl:
tweak: Conveyor belts can now be slowed by using multitool on the switch.
/:cl:


moxian:
In the interest of transparency, I should note that there is a minor issue: if you're using extremely low speeds, it takes a while (specifically - the time it takes for the item to travel one tile) for the item to start moving. See the gif:
xort0lOk9H

This was the reason for limiting the speed to only 50x. Might be a good idea to reduce this further to something like 25x. Although I'd argue that if you're using that big of a slowdown, you have a reason to, and can live with the minor visual discrepancy.

Carthusia:
As someone who will be using this for sushi bars, I can live with it as the conveyor will be under tables, and thus nearly invisible

tgui/packages/tgui/interfaces/ConveyorSwitch.js
```diff
@@ -0,0 +1,58 @@
...
+    oneWay,
+    position,
+  } = data;
+
+  return (
+    <Window>
+      <Window.Content>
+        <Section>
+          <LabeledList>
+            <LabeledList.Item label="lever position">
```
AffectedArc07, line 18:
```suggestion
            <LabeledList.Item label="Lever position">
```

tgui/packages/tgui/interfaces/ConveyorSwitch.js
```diff
@@ -0,0 +1,58 @@
...
+              { (position > 0)
+                ? "forward"
+                : (position < 0)
+                  ? "reverse"
+                  : "neutral"}
+            </LabeledList.Item>
+            <LabeledList.Item label="allow reverse">
```
AffectedArc07, line 25:
```suggestion
            <LabeledList.Item label="Allow reverse">
```

tgui/packages/tgui/interfaces/ConveyorSwitch.js
```diff
@@ -0,0 +1,58 @@
...
+              <Button.Checkbox
+                // content="allow reverse"
+                checked={!oneWay}
+                onClick={() => act('toggleOneWay')}
+              />
+            </LabeledList.Item>
+
+            <LabeledList.Item label="slowdown factor">
```
AffectedArc07, line 33:
```suggestion
            <LabeledList.Item label="Slowdown factor">
```

tgui/packages/tgui/interfaces/ConveyorSwitch.js
```diff
@@ -0,0 +1,58 @@
...
+              <Flex >
```
AffectedArc07, line 34:
```suggestion
              <Flex>
```

AffectedArc07 (changes requested):
First pass

Cocacolagua:
how does this look on mobs?

Kyep (approved):

SteelSlayer:
I'd be interested in seeing a profile of the conveyor pushing procs that has your new logic, compared to what we currently use.

Kiyahitayika:
Conveyer belts are barely used currently, this PR will make them much more worthwhile.

moxian:
> First pass

Thanks, addressed.

> how does this look on mobs?

Same as on items, with the same caveat that on high slowdown it takes a bit to start moving. The gifs below are on 5x:
hXQfpVWDdJ
aHZcqXnZI0

> I'd be interested in seeing a profile of the conveyor pushing procs that has your new logic, compared to what we currently use.

I honestly don't know how to get one. But honestly, I don't see how it would be much different from what we have today. Compared to live I do now calculate the `loc.contents - affecting` set, but it's `O(len(loc.contents))` complexity anyway, same as `for(var/atom/movable/AM in loc.contents)` on the very next line.
But for slow belts I do now spawn a bit less callbacks. Though that's only because items get moved less (over the same time frame).

SteelSlayer:
> I honestly don't know how to get one. But honestly, I don't see how it would be much different from what we have today. Compared to live I do now calculate the loc.contents - affecting set, but it's O(len(loc.contents)) complexity anyway, same as for(var/atom/movable/AM in loc.contents) on the very next line.
But for slow belts I do now spawn a bit less callbacks. Though that's only because items get moved less (over the same time frame).

You're right, I ran a short test and profiled your changes vs master, having the belts run at their normal speeds. The difference was extremely minimal, and nothing to fuss over so :+1:

variableundefined (approved):

procdrone:
Will you run at standard speed on top of a slow conveyor if you try to move on your own?

PS. What about reverse only position for levers?

moxian:
> Will you run at standard speed on top of a slow conveyor if you try to move on your own?

You will be slightly faster/slightly slower (depending on your heading), but not noticeably so.
Compare:
20201022-021257-oxqDxaOMp1
20201022-021336-gUUiRKJGE3
20201022-021501-hcuPxlXnGp


> PS. What about reverse only position for levers?

Currently on live attacking the lever with a multitool switches it between one-way, and two-way.
In two-way (default):
- toggle to start the belt moving items in the direction of the arrows
- toggle again to stop
- toggle once more to make it go the opposite direction
- toggle finally to stop again
- repeat
In one-way:
- toggle to start the belt moving items in the direction of the arrows
- toggle to stop
- repeat
This PR simply preserves this functionality.

PR #14663, opened 2020-10-19
Makes the healing guardian not take more damage in healing mode

Qwertytoforty:
## What Does This PR Do

This PR makes healing guardian always have a flat damage transfer rate of 0.7, or 70%, instead of becoming 1, or 100% damage transfer, when in heal mode.

## Why It's Good For The Game

Healing guardian is a nice guardian that could use a little love to make it a bit more common, and a nice way to make it do that is to not make it loose armor in heal mode. Why did it loose armor in heal mode? Probably so it could heal and drag away the user, but, 30% is not that much in the end. In any case, this means it's a little more sturdy when healing, and this wont overly change its combat abilities, as you do not heal in combat, and it's healing rate is unchanged.

## Changelog
:cl:
tweak: Healing guardian now always has a damage transfer rate of 70%.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/guardian/types/healer.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/healer.dm
@@ -1,6 +1,7 @@
 /mob/living/simple_animal/hostile/guardian/healer
 	friendly = "heals"
 	speed = 0
+	damage_transfer = 0.7
 	melee_damage_lower = 15
 	melee_damage_upper = 15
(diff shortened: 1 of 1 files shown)


Butlerblock:
Now make chaos/lightning/stealth/protector not useless

Qwertytoforty:
Already buffed lightning, chaos I need to work on, protector does what it does real well, and has quite a bit of armor even in combat mode.

TheMaveran:
Protector is borderline OP in a lot of situations, it’s assassin and chaos that need the love. @Butlerblock

variableundefined (approved):

PR #14676, opened 2020-10-20
Database indexing & table cleanup

Kyep:
## What Does This PR Do
1) Adds new DB indexes for various tables we use, making database queries faster.
2) Commits the existing SQL DB indexes, collation, etc that we already use in production into the SQL schema. This both ensures there aren't undocumented things in the live DB, and also helps downstream servers. **This eliminates >50% of the differences between the actual production paradise database, and the in-theory-accurate-but-actually-not official DB schema we're supposed to be using.**
3) Adds a table used by backend tooling (no in-game code) to the schema.
4) Adds a col to the player table for tracking of players' byond account creation dates.
5) Deletes an index (ckey_UNIQUE) that had no reason to exist

## Changelog
:cl: Kyep
tweak: database now has search indexes for faster queries, and supports tracking the creation date of player byond accounts.
/:cl:


SQL/updates/13-14.sql
```diff
@@ -0,0 +1,35 @@
...
+  `groupstr` varchar (32) NOT NULL DEFAULT '',
+  PRIMARY KEY (`ip`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+ALTER TABLE `ip2group` ADD INDEX(`groupstr`);
+
+
+# Delete tables that already exist but are not used anywhere in either game or backend code
+
+DROP TABLE IF EXISTS `erro_privacy`;
+DROP TABLE IF EXISTS `population`;
```
AffectedArc07, line 19: These arent part of `paradise_schema.sql`, and are remains of the original DB schema from 2013, hence the `erro_` prefix, and thus these lines are not relevant to 99% of servers. I would remove them.

SQL/updates/13-14.sql
```diff
@@ -0,0 +1,35 @@
...
+# Add new tables used by backend tooling
+
+CREATE TABLE `ip2group`
+(
+  `ip` varchar (18) NOT NULL,
+  `date` timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP NOT NULL,
+  `groupstr` varchar (32) NOT NULL DEFAULT '',
+  PRIMARY KEY (`ip`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+ALTER TABLE `ip2group` ADD INDEX(`groupstr`);
```
AffectedArc07, line 13: Questioning the worth of putting a closed-source backend table into the public schema if we are the only ones using it, and it wont have any ingame use, unless you plan to add some.

config/example/dbconfig.txt
```diff
@@ -9,7 +9,7 @@
 ## This value must be set to the version of the paradise schema in use.
 ## If this value does not match, the SQL database will not be loaded and an error will be generated.
 ## Roundstart will be delayed.
-DB_VERSION 13
+DB_VERSION 14
```
AffectedArc07, line 12: Dont forget to update `tools/travis/dbconfig.txt`

AffectedArc07 (changes requested):

SQL/updates/13-14.sql
```diff
@@ -0,0 +1,32 @@
...
+ALTER TABLE `customuseritems` ADD INDEX(`cuiCKey`);
+ALTER TABLE `donators` ADD INDEX(`ckey`);
+ALTER TABLE `library` ADD INDEX(`ckey`);
+ALTER TABLE `library` ADD INDEX(`flagged`);
+ALTER TABLE `notes` ADD INDEX(`ckey`);
+ALTER TABLE `oauth_tokens` ADD INDEX(`ckey`);
+
+# Add player table field for byond account creation date
+
+ALTER TABLE `player` ADD COLUMN `byond_date` DATE;
```
AffectedArc07, line 32: This was requested for an upcoming PR I am going to make, and its easier for me to put the SQL bump in now than another version increase later on.

SQL/paradise_schema.sql
```diff
@@ -604,3 +616,17 @@ CREATE TABLE `changelog` (
 	`cl_entry` TEXT NOT NULL,
 	PRIMARY KEY (`id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+--
+-- Table structure for table `ip2group`
+--
+DROP TABLE IF EXISTS `ip2group`;
+CREATE TABLE `ip2group`
+(
```
AffectedArc07, line 625:
```suggestion
CREATE TABLE `ip2group` (
```
Keep standardisation between creations

SQL/paradise_schema_prefixed.sql
```diff
@@ -601,3 +613,17 @@ CREATE TABLE `SS13_changelog` (
 	`cl_entry` TEXT NOT NULL,
 	PRIMARY KEY (`id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+--
+-- Table structure for table `ip2group`
+--
+DROP TABLE IF EXISTS `SS13_ip2group`;
+CREATE TABLE `SS13_ip2group`
+(
```
AffectedArc07, line 622:
```suggestion
CREATE TABLE `SS13_ip2group` (
```

SQL/updates/13-14.sql
```diff
@@ -0,0 +1,32 @@
+#Updating the SQL from version 13 to version 14. -Kyet
+
+
+# Add new tables used by backend tooling
+
+CREATE TABLE `ip2group`
+(
```
AffectedArc07, line 7:
```suggestion
CREATE TABLE `ip2group` (
```

AffectedArc07 (changes requested):

SQL/paradise_schema.sql
```diff
@@ -242,32 +250,37 @@ DROP TABLE IF EXISTS `player`;
...
   `parallax` tinyint(1) DEFAULT '8',
   PRIMARY KEY (`id`),
-  UNIQUE KEY `ckey` (`ckey`)
-) ENGINE=InnoDB AUTO_INCREMENT=32446 DEFAULT CHARSET=utf8mb4;
+  UNIQUE KEY `ckey` (`ckey`),
+  KEY `lastseen` (`lastseen`),
+  KEY `computerid` (`computerid`),
+  KEY `ip` (`ip`),
+  KEY `fuid` (`fuid`),
+  KEY `fupdate` (`fupdate`)
```
AffectedArc07, line 282: `byond_date` was removed here

SQL/paradise_schema.sql
```diff
@@ -473,14 +489,14 @@ DROP TABLE IF EXISTS `watch`;
...
+  `ckey` varchar(32) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `reason` mediumtext COLLATE utf8mb4_unicode_ci NOT NULL,
   `timestamp` datetime NOT NULL,
-  `adminckey` varchar(32) NOT NULL,
-  `last_editor` varchar(32),
-  `edits` text,
+  `adminckey` varchar(32) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `last_editor` varchar(32) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `edits` mediumtext COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   PRIMARY KEY (`ckey`)
```
AffectedArc07, line 498: No `ckey` inde on watchlist? Its queried any time any player joins the server so it would make sense imo

SQL/paradise_schema.sql
Kyep: Re-addded

SQL/paradise_schema.sql
Kyep: PRIMARY KEY is always an index.
https://stackoverflow.com/questions/1071180/is-the-primary-key-automatically-indexed-in-mysql

AffectedArc07: Ah yes, forgot about that. For some reason `watchlist` is the only table which doesnt use an ID col

Kyep: Reason: for most tables, you can have multiple rows with the same ckey. E.g. there can be multiple bans, notes, etc that have the same ckey. So you need something to distinguish them - an ID number. But watchlist can only ever have one entry for any given ckey... so it makes sense for that to be the primary key of the table. No need for an ID number.
Why don't tables like, say, admin, which can only have one entry per ckey... ditch their ID number and just have ckey as their primary key? I don't know. Structure wise they could. But I'm trying to avoid like.... restructuring the entire database. So I'm not changing that.
But equally, I'm not adding an unnecessary id col to tables that need it. And I'm making use of alternate primary keys where they exist.

AffectedArc07: Could be something for down the line. I know (probably) MSO did this for /tg/
image

*NOTE: `ckey` and `byond_key` are different things (KEY: `Key with spaces and case`, CKEY: `keywithspacesandcase`)*

AffectedArc07 (approved):

PR #14677, opened 2020-10-20
Fixes ghosts unable to pass gateway

Kyep:
## What Does This PR Do
Ghosts can now always pass through the gateway, so long as an away mission is loaded.
Previously, they could get "the gateway has no destination" if they clicked the gate before a human touched it.

## Changelog
:cl: Kyep
fix: ghosts can now pass through the gateway to/from away missions even if a human has not touched it yet.
/:cl:

### DIFF
--- a/code/modules/awaymissions/gateway.dm
+++ b/code/modules/awaymissions/gateway.dm
@@ -47,7 +47,10 @@ GLOBAL_DATUM_INIT(the_gateway, /obj/machinery/gateway/centerstation, null)
 /obj/machinery/gateway/centerstation/Initialize()
 	..()
 	update_icon()
-	wait = world.time + config.gateway_delay	//+ thirty minutes default
+	wait = world.time + config.gateway_delay
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/gateway/centerstation/LateInitialize()
 	awaygate = locate(/obj/machinery/gateway/centeraway) in GLOB.machines
 
 /obj/machinery/gateway/centerstation/update_density_from_dir()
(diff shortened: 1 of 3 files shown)


Cocacolagua:
all ghost players have this bug, not admins only

AffectedArc07 (approved):

Kyep:
@Cocacolagua Yeah, changed the PR so it lets non-admin ghosts pass too.

code/modules/awaymissions/gateway.dm
```diff
@@ -47,7 +47,10 @@ GLOBAL_DATUM_INIT(the_gateway, /obj/machinery/gateway/centerstation, null)
 /obj/machinery/gateway/centerstation/Initialize()
 	..()
 	update_icon()
-	wait = world.time + config.gateway_delay	//+ thirty minutes default
+	wait = world.time + config.gateway_delay
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/gateway/centerstation/LateInitialize()
```
SteelSlayer, line 53: What's the reason for moving this to late initialize? Was `awaygate` getting set to null because it hadn't spawned in yet properly or something?

Kyep: awaygate should have been defined in Initialize(), but it wasn't because a GLOB it was searching in wasn't defined by that point. AA nicely found that moving it to LateInitialize() made it work properly.

SteelSlayer: Yeah sounds good to me.

SteelSlayer (approved):

code/modules/awaymissions/gateway.dm
AffectedArc07: @SteelSlayer The away gate gets added to GLOB.machines during its own Initialize() call, which can cause issues if the station gate fires off before that one does. By moving the search to LateInitialize(), an away gate will always be found since there's no longer a race condition.

AffectedArc07 (approved):

PR #14681, opened 2020-10-21
Helpful intent prevents harm

moxian:
## What Does This PR Do
Prevents damaging things when on help intent.

## Why It's Good For The Game
Accidentally hitting someone feels bad for both parties.
How often have you swung at a bystander when replacing the floors? How many times have you seen a new doctor robust a dying patient with a defib? Do you remember that time when you wanted to hug your friend, but forgot that you have an desword in hand, and sliced them in half instead?
Well, with this PR all of those nightmares are a things of the past!

And, after all, why do we even need 4 intents, when they are exactly the same 95% of the time?

In action:
oO2G8XaB8w
IMLTU8DPMD

Tested:
- On Renault (see above)
- Equipment (cameras, ipcs, lights) no longer break on help intent
- Equipment still does break on harm intent
- Terror spiders AI still works, and they still consistently kill hapless dummies

Concerns:
- This only changes melee behavior, not guns (to limit the scope of change, mostly, and also because it's rarer to accidentally fire those, IMO)
- I did *not* remove attack admin logging, out of an abundance of caution (since the attack logs and damage application are handled in separate places. It's possible i've missed a proc or two)
- The attack message on help intent is smaller now, but still might be mistake for a real deal. I don't know if this is a problem. The reason I keep text content as is, is so that we still can see the flavorful on-hit verbs ("cut"/"sliced"/"smacked"/"robusted"/...) even in otherwise peaceful rounds.
- Since this changes such a fundamental aspect of the game, it should probably go through a test merge first.

## Changelog
:cl:
tweak: Hitting things in melee when on Help intent no longer damages them.
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -64,11 +64,11 @@
 			else
 				return TRUE
 
-	if(force && HAS_TRAIT(user, TRAIT_PACIFISM))
+	if(HAS_TRAIT(user, TRAIT_PACIFISM) && force && user.a_intent != INTENT_HELP)
 		to_chat(user, "<span class='warning'>You don't want to harm other living beings!</span>")
 		return
 
-	if(!force)
+	if(!force || user.a_intent == INTENT_HELP)
 		playsound(loc, 'sound/weapons/tap.ogg', get_clamped_volume(), 1, -1)
 	else
 		SEND_SIGNAL(M, COMSIG_ITEM_ATTACK)
@@ -100,13 +100,15 @@
 	return
 
 /obj/attacked_by(obj/item/I, mob/living/user)
+	if(user.a_intent == INTENT_HELP)
+		return
 	if(I.force)
 		user.visible_message("<span class='danger'>[user] has hit [src] with [I]!</span>", "<span class='danger'>You hit [src] with [I]!</span>")
 	take_damage(I.force, I.damtype, "melee", 1)
 
 /mob/living/attacked_by(obj/item/I, mob/living/user, def_zone)
 	send_item_attack_message(I, user)
-	if(I.force)
+	if(I.force && user.a_intent != INTENT_HELP)
 		apply_damage(I.force, I.damtype, def_zone)
 		if(I.damtype == BRUTE)
 			if(prob(33))
@@ -118,7 +120,10 @@
 		return TRUE //successful attack
 
 /mob/living/simple_animal/attacked_by(obj/item/I, mob/living/user)
-	if(!I.force)
+	if (user.a_intent == INTENT_HELP)
+		user.visible_message("<span class='warning'>[user] gently taps [src] with [I].</span>",\
+						"<span class='warning'>You gently tap [src] with [I].</span>")
+	else if(!I.force)
 		user.visible_message("<span class='warning'>[user] gently taps [src] with [I].</span>",\
 						"<span class='warning'>This weapon is ineffective, it does no damage!</span>")
 	else if(I.force < force_threshold || I.damtype == STAMINA)
@@ -150,9 +155,14 @@
 	var/message_hit_area = ""
 	if(hit_area)
 		message_hit_area = " in the [hit_area]"
+	var/self_class = "combat userdanger"
+	var/others_class = "combat danger"
+	if(user.a_intent == INTENT_HELP)
+		self_class = "warning"
+		others_class = "warning"
 	var/attack_message = "[src] has been [message_verb][message_hit_area] with [I]."
 	if(user in viewers(src, null))
 		attack_message = "[user] has [message_verb] [src][message_hit_area] with [I]!"
-	visible_message("<span class='combat danger'>[attack_message]</span>",\
-		"<span class='combat userdanger'>[attack_message]</span>")
+	visible_message("<span class='[others_class]'>[attack_message]</span>",\
+		"<span class='[self_class]'>[attack_message]</span>")
 	return 1
--- a/code/game/gamemodes/cult/cult_items.dm
+++ b/code/game/gamemodes/cult/cult_items.dm
@@ -12,6 +12,8 @@
 
 
 /obj/item/melee/cultblade/attack(mob/living/target as mob, mob/living/carbon/human/user as mob)
+	if(user.a_intent == INTENT_HELP)
+		return ..()
 	if(!iscultist(user))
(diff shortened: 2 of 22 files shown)


TDSSS:
Disagree with this. if you're clicking people with a weapon like an e-sword, you probably want to hurt them regardless of intent.

Besides, accidentally beating people in the face with a crowbar is a part of the game at this point, imo. Nobody's going to die from that, it just gives medbay a bit of something to do and adds to the general tone of the game.

shmabuu:
All I think of with this is nukies and stuff lightly tapping people with a dsword, also.. does this apply to stun batons and stuff ? They don't do any damage unless on harm intent but if you bounce your baton off people in help intent that's kinda wack

code/_onclick/item_attack.dm
```diff
@@ -64,11 +64,11 @@
 			else
 				return TRUE
 
-	if(force && HAS_TRAIT(user, TRAIT_PACIFISM))
+	if(HAS_TRAIT(user, TRAIT_PACIFISM) && force && user.a_intent != INTENT_HELP)
```
TDSSS, line 67: Does this now let pacifists use non-lethal stun weapons like stun batons or police batons? Or do those no longer work in help intent? Either approach causes issues, imo.

moxian: The intent (ha!) here is that pacifists can boop people with desword on help intent (and do no damage). The rest of pacifist behaviour should be unchanged.
The should still be able to use batons on help - those have `force=0` so this check fails.

moxian:
> if you're clicking people with a weapon like an e-sword, you probably want to hurt them regardless of intent.

Sure. Select the harm intent then.

> Besides, accidentally beating people in the face with a crowbar is a part of the game at this point, imo. Nobody's going to die from that, it just gives medbay a bit of something to do and adds to the general tone of the game.

I'm a big advocate for "losing is fun" myself, but this case is different. If you try to open a toolbox by clicking on yourself, and get hurt - that's "fun", i don't mind this at all, go ahead. It's when you accidentally hurt *other* people that I have issue with, because that creates very awkward social situations, which are not fun for either dealing nor receiving party.

> All I think of with this is nukies and stuff lightly tapping people with a dsword

This is a matter of re-education. Yes, people who are used to the current system will need to change their habits. But folks completely new to SS13 (or even coming from other servers!) would have no problem, since this knowledge should picked up within 1-2 rounds.

> also.. does this apply to stun batons and stuff ?

Stun batons behavior is unchanged - they still stun on help intent.

Qwertytoforty:
The fact this still makes a harm intent admin log when it is on... is stupid. We do not want to have to tell people to stop having people click on each other rapidly in help intent as they are bapping someone with a toolbox in a group of 3 people. I also don't think this is a good change, since swapping intents means you can no longer go through people on help intent if you are beating someone with _insert weapon here_ and you now will always need to beat down everyone in front of you. The fact it still makes the harm intent message but smaller is just going to confuse people and bait people into attacking each other

For these reasons, I am not a fan of this PR

procdrone:
Yeah, I'm also against this. This will only generate confusion and inconvenience. While this looks cool, it will just screw over people.

variableundefined (commented):
Objection

MarsM0nd:
The messages that still come, especially the logs one, make it bad for now. As for requiring non-help intent, that'll cause quite a problem in a team fight. Unable to pass others is bad, and you can't really get it coordinated to swap, as both can't fight at all in melee for that. 
If it's to stop accidental hits, maybe stopping the first hit you do if you haven't recently hit someone and are in help intent, but nothing more. So the first click on help intent with a weapon on someone may be an accident, more aren't, as well as it won't be an accident if you are in a fight with someone currently, or at least a realistic one. Won't smash a toolbox in someones face, but smashing one in another guy that was just swapped in by the one you wanted to hit will happen.

TDSSS:
> If it's to stop accidental hits, maybe stopping the first hit you do if you haven't recently hit someone and are in help intent, but nothing more. So the first click on help intent with a weapon on someone may be an accident, more aren't, as well as it won't be an accident if you are in a fight with someone currently, or at least a realistic one. Won't smash a toolbox in someones face, but smashing one in another guy that was just swapped in by the one you wanted to hit will happen.

That would be incredibly snowflakey and not intuitive. Smashing someone in the face by accident sometimes really isn't that bad.

Mitchs98:
Disagree with this for all the reasons listed above. Help intent combat isn't always beneficial, but there are plenty of times it is and it's needed to avoid just mowing down everyone in your path to get away or whatever.

As said previously, I just see people baiting others into attacking them and what have you/misunderstanding the 'harmless' attacks for actual attacks despite the lack of actual damage.

Kyep (changes requested):
I don't like this.
While I like the idea of making the game more new-player-friendly (hitting yourself with something because you clicked yourself is... not intuitive)...
... the issue I see with this is that it will force people onto harm/disarm intent a lot more.
That's bad, because it will make moving around a lot more annoying, because you will constantly be getting stuck on people and unable to move through them.

Objection.

moxian:
When making this PR i completely overlooked the effect it has on movement, and that it would prevent shuffling in real combat situations. I agree that this is a big deal, and not a good change. I don't see an easy way to address this while keeping the core idea of the PR intact, so I'm closing this.


I don't think the other objections are major blockers though, and they can be feasibly addressed, so I'd like to mention those (for posterity or something):

> The fact this still makes a harm intent admin log when it is on... is stupid. We do not want to have to tell people to stop having people click on each other rapidly in help intent as they are bapping someone with a toolbox in a group of 3 people. The fact it still makes the harm intent message but smaller is just going to confuse people and bait people into attacking each other

People can already rapidly bap each other with belts, carboard boxes, toys, etc etc. They do that sometimes, but not often enough to be a problem. I don't envision this being much more common occurrence was this PR accepted.
With some effort, admin logging for these cases can also be tweaked to ignore these zero-damage attacks.

 > If it's to stop accidental hits, maybe stopping the first hit you do if you haven't recently hit someone and are in help intent, but nothing more. So the first click on help intent with a weapon on someone may be an accident, more aren't, as well as it won't be an accident if you are in a fight with someone currently, or at least a realistic one. 

I agree with others that this particular fix would be snowflakey and clunky to use. But thanks for mentioning this anyway!


Thanks everyone for the feedback!

PR #14686, opened 2020-10-21
[TGUI] TachyonArray/doppler array

farie82:
## What Does This PR Do
Makes the TachyonArray / doppler array TGUIified

## Why It's Good For The Game
Nyanow UwI be fucking dead. Long live TGUI

## Images of changes
image
image

## Changelog
:cl:
add: TGUIify the doppler array/TachyonArray
/:cl:


code/game/machinery/doppler_array.dm
```diff
@@ -37,9 +37,6 @@ GLOBAL_LIST_EMPTY(doppler_arrays)
 	logged_explosions.Cut()
 	return ..()
 
-/obj/machinery/doppler_array/process()
-	return PROCESS_KILL
```
AffectedArc07, line 41: Put this back. The whole point of this is to remove the excess processing overhead of processing an object with no `process()` call

farie82, line 41: `machine/process()` already does the same

code/game/machinery/doppler_array.dm
```diff
@@ -176,47 +173,50 @@ GLOBAL_LIST_EMPTY(doppler_arrays)
...
-/obj/machinery/doppler_array/ui_data(mob/user, ui_key = "main", datum/topic_state/state = GLOB.default_state)
-	var/data[0]
-	var/list/explosion_data = list()
-	for(var/D in logged_explosions)
-		var/datum/explosion_log/E = D
-		explosion_data += list(list(
+/obj/machinery/doppler_array/tgui_data(mob/user)
+	var/list/data = list()
+	var/list/records = list()
+	for(var/i in 1 to length(logged_explosions))
```
AffectedArc07, line 185: Any reason for switching this to index instead of UIDs?

tgui/packages/tgui/interfaces/TachyonArray.js
```diff
@@ -0,0 +1,108 @@
...
+                align="center"
+                onClick={() => act('delete_logs')} />
+            </LabeledList.Item>
+          </LabeledList>
+        </Section>
+        {!records.length ? (
+          <NoticeBox>
+            No Records
+          </NoticeBox>
+        ) : (<TachyonArrayContent />)}
```
AffectedArc07, line 45:
```suggestion
        ) : (
          <TachyonArrayContent />
        )}
```

AffectedArc07 (changes requested):

code/game/machinery/doppler_array.dm
farie82: The list order won't change. Only edge case is when people try to remove things at the (near) exact same time.
Using indexes here was a bit cheaper on the processing. Also less UIDs being used

code/game/machinery/doppler_array.dm
farie82: image
It's dupe code

tgui/packages/tgui/interfaces/TachyonArray.js
farie82: The formatter kinda shat itself when I did that. I'll change it to this though

tgui/packages/tgui/interfaces/TachyonArray.js
```diff
@@ -0,0 +1,110 @@
...
+  const { act, data } = useBackend(context);
+  const {
+    records = [],
+  } = data;
+
+  return (
+    <Section>
+      <Flex>
+        <Flex.Item>
+          <Section label="Logged Explosions">
```
SteelSlayer, line 63: The correct syntax is `title="Logged Explosions"`. But also, I think this needs to go onto the outer section at line 60. Otherwise, each time a test is made there will be a section named "Logged Explosions" created. 

And for a personal opinion, I think you should ditch this section, because nested sections always looks a bit funky with their double borders. 

image

farie82: Oh darn forgot to change that one. Good spot will fix

AffectedArc07 (approved):

PR #14687, opened 2020-10-21
Makes TGUI respect advanced admin interaction

farie82:
## What Does This PR Do
Makes TGUI interfaces respect the advanced_admin_interaction variable when ghosted as admin.
The UI will still seem enabled and clickable but if you try it it won't work unless you got the variable on TRUE as admin

## Why It's Good For The Game
It's an oversight/bug

## Changelog
:cl:
fix: Makes TGUI interfaces respect the advanced_admin_interaction variable when ghosted as admin
/:cl:

### DIFF
--- a/code/modules/tgui/states/default.dm
+++ b/code/modules/tgui/states/default.dm
@@ -57,7 +57,7 @@ GLOBAL_DATUM_INIT(tgui_default_state, /datum/tgui_state/default, new)
 	else
 		return ..()
 
-/mob/dead/observer/default_can_use_tgui_topic()
-	if(can_admin_interact())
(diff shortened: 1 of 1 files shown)


AffectedArc07:
Not an oversight. Intentional.
image
Nano does the exact same thing.

Qwertytoforty:
Indeed, as mentioned in the discord, its more of for buttons and lockers, so we can open those only with the setting on, but otherwise, we generally interacted with uis as a ghost with this setting off. This way we can interact with doors and apcs and consoles just fine and use it, but we wont accidentally open a locker or close a fire lock by clicking on it.

PR #14656, opened 2020-10-19
A variety of minor food related fixes and tweaks

PacifistDalek:
## What Does This PR Do
Tweaks and fixes a bunch of minor food things: typos, outdated descriptions, a few broken obscure recipes , and a few recipes that didn't make any sense as they were, and removed a few recipes that both didn't work and didn't make sense (Beff, Pepperoni, a duplicate doughball recipe). Also renamed Milo Soup to Miso Soup since that appears to be (presumably) a reference that predates our server and nobody seems to know what it actually is a reference to!

(My first PR, so please let me know if I screwed anything up)

## Why It's Good For The Game
Typos, outdated descriptions, and broken recipes are bad. The recipes that I tweaked were to make them more in line with how said recipes are made in real life, or to make them consistent with other similar recipes in game. This will make it more intuitive and easier to remember for chefs.

The fries/tator tot recipe change also means they benefit from kitchen upgrades generating more of a given recipe, so its easier to fill an entire station with french fries now!

## Changelog
:cl:
del: Removed the semi-functional recipes for Beff, Pepperoni, and a duplicate Doughball recipe
fix: Fixed the recipes for Universal Enzyme and Weird Cheese
add: Added a recipe for Cheese Substitute (based on an existing variant recipe of Weird Cheese)
tweak: Onion Rings are now made in the deep fryer
tweak: Cutting potato gives wedges, processing potato gives raw potato sticks. Frying the wedges gives you potato chips, frying the sticks gives you fries, and processing the sticks gives you tator tots.
tweak: made burger recipes consistently use a bun as a core item (except for Big Bite and Super Big Bite burgers, who use a normal burger instead)
tweak: Yaki Imo is made by baking a sweet potato
tweak: drying grapes gives you healthy homemade raisins
tweak: wine is hand made from grape juice, not berry juice
tweak: toasted sandwiches and tofurkey are made in the oven now
spellcheck: fixed a few typos, updated some out of date descriptions
/:cl:

PR #14699, opened 2020-10-23
Fix new cheese /on_reaction path

moxian:
## What Does This PR Do
Fixes a minor oversight of #14656 . In that PR cheese was renamed to fake_cheese, but /cheese/on_reaction proc stayed, now effectively orphaned.

## Why It's Good For The Game
bugfix

## Changelog
N/A

### DIFF
--- a/code/modules/reagents/chemistry/recipes/food.dm
+++ b/code/modules/reagents/chemistry/recipes/food.dm
@@ -146,7 +146,7 @@
 	result_amount = 5
 	mix_message = "The mixture curdles up."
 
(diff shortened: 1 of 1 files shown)


AffectedArc07:
Please remove the CL tags from your description entirely if you have no entries in them. The generator is a finnicky piece of garbo.

moxian:
Done. I assumed it's finnicky the other way and would complain if there is NO section present.

farie82 (approved):
It technically is a fix. So you can add that in the CL since you fixed the message you should get when making fake cheese

PacifistDalek:
MGS3-Salute

PR #14709, opened 2020-10-24
Stops the crew from preforming the darkest magical art, metamancy

Qwertytoforty:
## What Does This PR Do

This PR replaces most 1's and 0's in the wizard file to TRUE and FALSE, and makes it so in the default wizard gamemode, the round does not end unless the wizard and all of his apprentices have been dead for 5 minutes. If the wizard dies, but is revived, the 5 minute counter will fail, and the timer will be able to trigger again when he dies again. Theoretically, he could be killed, revived, and killed again in 5 minutes, but that is fine. 

This does not affect raging mages. 

## Why It's Good For The Game

Crew preforms a dark art almost every wizard round. The dark art of metamancy, or metaknowledge, is practiced by the crew. The wizard is beating to crit, and then dies. 1 of 3 things happens. He is a lich, and revives. He is a normal wizard, and dies, and there is much celebration. Or, the 3rd option. The wizard has an apprentice, mind swap, or is an adminbussed wizard. If the 3rd option is picked, crew will either: Call out that an apprentice must be alive, or even if the wizard does not have mind swap, or is adminbussed, declare the wizard must have mind swap, and begin interrogating, or beating to death, all of the crew around him.

Why does this happen? Because the crew knows when the wizard is dead that the round ends. They know, that since the "wizard" is dead, that there must be a mind swap. Time and time again, we see this happen. Last night we had a round where we gave a traitor some magical powers, and when he died and the crew noticed the round was still going, CE, HOS, and AI said that mind swap almost certainly happened. They said this, as the round did not end. Said traitor did not have mind swap, the only evidence of mind swap was the lack of round ending.

By adding a 5 minute timer, the crew can no longer say that, a mind-swap must have happened, because the round is still going. They no longer can kill a bunch of people in a row to find out who the wizard is by seeing the round ending. Crew must now actually commit to their options, and no longer meta know who the wizard is by death. I considered making this only happen when the wizard does have mind swap, but this then again, leads to crew knowing mind swap happened. Does this mean the round continues for 5 extra minutes on normal wizard rounds? Yes. But the cost of those 5 extra minutes are worth the benefit of stopping the mind swap meta checks.

## Changelog
:cl:
tweak: The wizard and apprentices now have to stay dead for 5 minutes for a wizard round to end.
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/wizard.dm
+++ b/code/game/gamemodes/wizard/wizard.dm
@@ -206,7 +206,7 @@
 			if(able_to_check)
 				able_to_check = FALSE
 				message_admins("final countdown")
-				addtimer(CALLBACK(src, .proc/metamancy), 1 MINUTES)
+				addtimer(CALLBACK(src, .proc/metamancy), 5 MINUTES)
 			else
 				message_admins("this check is failing")
 
--- a/code/game/gamemodes/wizard/wizard.dm
+++ b/code/game/gamemodes/wizard/wizard.dm
@@ -199,21 +199,17 @@
 		return ..()
 	else
 		if(really_finished)
-			message_admins("END THIS SHITTTTTTE")
 			finished = TRUE
 			return TRUE
 		else
 			if(able_to_check)
 				able_to_check = FALSE
-				message_admins("final countdown")
 				addtimer(CALLBACK(src, .proc/metamancy), 5 MINUTES)
 			else
-				message_admins("this check is failing")
 
 /datum/game_mode/wizard/proc/metamancy()
 	really_finished = TRUE
 	able_to_check = TRUE
-	message_admins("yeet")
 
 /datum/game_mode/wizard/declare_completion(var/ragin = FALSE)
 	if(finished && !ragin)
--- a/code/game/gamemodes/wizard/wizard.dm
+++ b/code/game/gamemodes/wizard/wizard.dm
@@ -8,10 +8,12 @@
 	required_players = 20
 	required_enemies = 1
 	recommended_enemies = 1
-	var/use_huds = 1
+	var/use_huds = TRUE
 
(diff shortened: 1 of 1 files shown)


Carthusia:
Personally id prefer the round not end when they die and instead an increased % for more hazardous midround to show up comes into place, that way stories get to continue, but this is a nice step towards that possibility in the future.

TDSSS:
True, it's really rather inconsistent. Some rounds instantly end because some people are dead while others just continue. Personally, I'd prefer if all rounds just continued. But this is still a good change in general.

Kiyahitayika:
Facetious Alternative Solution: Round ends only when a Corgi pulling a shower curtain over the Wizard's dead body.

variableundefined (approved):
Approved for design

Kyep:
I'm... skeptical about this.
While it does appear to be popular, at the same time... personally I'd rather wizard rounds just ended immediately.
People have clocks. A 5-minute delay isn't going to stop people metaing, its just going to delay it for 5 minutes - at the cost of delaying every wizard round.

TDSSS:
We could always make wizard rounds just not end automatically at all

Spacemanspark:
What if wizards were a major event antagonist with specific objectives, as opposed to being a round type that ends in about fifteen to twenty minutes on average?

Necaladun:
I don't think this change is really worthwhile for a niche situation. Extending every wizard round by 5mins to avoid something that rarely happens doesn't feel like a good deal.

If people are going to assume things incorrectly because of meta knowledge, then that should be dealt with OOCly, and fucked with by admins screwing with the meta. Master Einath The Archmage is my wizard character for that. A few people have picked up on his acronym, other admins are free to use him.

variableundefined:
Closing due to two head objections.

PR #14710, opened 2020-10-24
TM NEEDED - Structures no longer /New()

AffectedArc07:
## What Does This PR Do
Kills off all `/New()` instances for `/obj/structure` in favour of `Initialize()`, and also substituting in `/LateInitialize()` where needed instead of unreliable `spawn()` use. 

*I am not sure on 100% of my ..() calls here and whether I need args or not. Feel free to correct me in review*

The one structure that hanst been converted is the `/obj/structure/fusion_reactor` used on the syndicate depot, since @Kyep is certain that needs to use `/New()`, so I will have to find a way to make that work at some point, which is out the scope of this PR.

Reasons for moving to Initialize() are as follows:
- Less lag when initially starting up DD, which reduces the time period between test restarts and reconnection delay
- Atoms which `/New()` call that as soon as they are placed into the map, which could cause complications with map rotation and how I plan to load the maps in
- Initialize() gives us a lot more control of when and how atoms initialize [See LateInitialize() and Initialization hints]

I didnt encounter any weird issues when testing this locally, but a full TM is **definitely** in order due to the sheer scope of this, and how `Initialize()` can have unintended side effects if not done properly. 

## Why It's Good For The Game
See above reasons

## Changelog
:cl: AffectedArc07
tweak: All /obj/structure types now use Initialize() instead of New()
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/core.dm
+++ b/code/game/gamemodes/blob/blobs/core.dm
@@ -12,7 +12,8 @@
 	var/is_offspring = null
 	var/selecting = 0
 
-/obj/structure/blob/core/Initialize(mapload, loc, h = 200, client/new_overmind = null, new_rate = 2, offspring)
+/obj/structure/blob/core/Initialize(mapload, h = 200, client/new_overmind = null, new_rate = 2, offspring)
+	. = ..()
 	GLOB.blob_cores += src
 	START_PROCESSING(SSobj, src)
 	GLOB.poi_list |= src
@@ -24,7 +25,6 @@
 	if(!overmind)
 		create_overmind(new_overmind)
 	point_rate = new_rate
-	..(mapload, loc, h)
 
 
 /obj/structure/blob/core/adjustcolors(var/a_color)
--- a/code/game/gamemodes/blob/blobs/core.dm
+++ b/code/game/gamemodes/blob/blobs/core.dm
@@ -12,7 +12,7 @@
 	var/is_offspring = null
 	var/selecting = 0
 
-/obj/structure/blob/core/New(loc, var/h = 200, var/client/new_overmind = null, var/new_rate = 2, offspring)
+/obj/structure/blob/core/Initialize(mapload, loc, h = 200, client/new_overmind = null, new_rate = 2, offspring)
 	GLOB.blob_cores += src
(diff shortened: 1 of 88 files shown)


code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -24,7 +24,7 @@
 	if(!overmind)
 		create_overmind(new_overmind)
 	point_rate = new_rate
-	..(loc, h)
+	..(mapload, loc, h)
```
AffectedArc07, line 27: Instance of where I am unsure with parent args

code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -12,7 +12,7 @@
 	var/is_offspring = null
 	var/selecting = 0
 
-/obj/structure/blob/core/New(loc, var/h = 200, var/client/new_overmind = null, var/new_rate = 2, offspring)
+/obj/structure/blob/core/Initialize(mapload, loc, h = 200, client/new_overmind = null, new_rate = 2, offspring)
```
Cyberboss, line 15: loc isn't an arg of Initialize, mapload replaces it

code/game/gamemodes/blob/blobs/core.dm
Cyberboss: Should just be . = ..(), and at the start of the proc

Cyberboss (changes requested):

code/game/objects/structures/electricchair.dm
```diff
@@ -7,19 +7,18 @@
 	var/last_time = 1.0
 	var/delay_time = 50
 
-/obj/structure/chair/e_chair/New()
-	..()
+/obj/structure/chair/e_chair/Initialize(mapload)
+	. = ..()
 	overlays += image('icons/obj/chairs.dmi', src, "echair_over", MOB_LAYER + 1, dir)
-	spawn(2)
```
AffectedArc07, line 13: I see no reason that these things should have `spawn()`'d in `New()`

AffectedArc07:
I will fry these fish another day, first I have to fix the init logic itself in half the items in the game

PR #14711, opened 2020-10-24
[TGUI] Changeling evolution menu

SteelSlayer:
## What Does This PR Do
Converts the godawful changeling evolution menu to TGUI.

Yes I realize this was not a nanoUI and therefore other PRs have priority over this one.

## Why It's Good For The Game
The old menu sucks, TGUI good.

## Images of changes
 By default the UI will start in the "expanded" view mode.

1NJQ8ZS

## Changelog
:cl:
tweak: The changeling evolution menu now uses TGUI.
/:cl:


code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -1,300 +1,85 @@
...
-				var locked_tabs = new Array();
 
-				function updateSearch(){
+	var/datum/changeling/cling = owner.mind.changeling
+	. = TRUE
 
+	switch(action)
+		if("readapt")
+			cling.lingRespec(owner)
+			purchased_abilities.Cut()
```
dearmochi, line 70: I'd probably check if the re-spec passed before doing any UI-facing changes, so as to not have any inconsistencies between the server and client

code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -1,300 +1,85 @@
...
 
-					var filter_text = document.getElementById('filter');
-					var filter = filter_text.value.toLowerCase();
+		if("purchase")
+			var/power_name = params["power_name"]
+			purchased_abilities += power_name
```
dearmochi, line 74: Ditto - also I'd replace the `+=` by `|=` so you can't have multiple of the same ability (which doesn't really matter, but can still lead to a big list if spammed, which is bad when networking)

code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -1,300 +1,85 @@
...
-	return dat
-
-
-/datum/action/changeling/evolution_menu/Topic(href, href_list)
-	..()
-	if(!(iscarbon(usr) && usr.mind && usr.mind.changeling))
-		return
+		if("set_view_mode")
+			var/new_view_mode = text2num(params["mode"])
+			ASSERT(new_view_mode in list(COMPACT_MODE, EXPANDED_MODE))
```
dearmochi, line 79: I wouldn't use `ASSERT` as the passed value can be altered by the client. If they decided to voluntarily spam the act with an invalid mode, the assertion would fail every time and cause runtimes. Just have it silently fail with a `return`

code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -1,300 +1,85 @@
...
+			view_mode = new_view_mode
 
-	if(href_list["P"])
-		usr.mind.changeling.purchasePower(usr, href_list["P"])
-	else if(href_list["readapt"])
-		usr.mind.changeling.lingRespec(usr)
-	var/dat = create_menu(usr.mind.changeling)
-	usr << browse(dat, "window=powers;size=600x700")
-/////
+	SStgui.update_uis(src)
```
dearmochi, line 82: Redundant because returning `TRUE` from `tgui_act` triggers an update

dearmochi (changes requested):
A few things. JS side looks fine and the theme is great!

I'd also make a `.scss` file for the `EvolutionMenu` interface, at least for the repeated elements (the abilities) to reduce the rendered template size

code/game/gamemodes/changeling/evolution_menu.dm
SteelSlayer: Oh really, even with autoupdate set to `FALSE`?

AffectedArc07: `autoupdate = FALSE` means that the UI wont periodically refresh its `tgui_data()`

SteelSlayer:
Just want to get this to compile for now from that travis nitpick. I do plan on doing Mochi's suggestion of:
> I'd also make a .scss file for the EvolutionMenu interface, at least for the repeated elements (the abilities) to reduce the rendered template size
Sometime in the near future, if I can figure it out.

dearmochi (approved):

code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -1,300 +1,88 @@
...
-	return dat
-
-
-/datum/action/changeling/evolution_menu/Topic(href, href_list)
-	..()
-	if(!(iscarbon(usr) && usr.mind && usr.mind.changeling))
+/datum/action/changeling/evolution_menu/Trigger()
+	tgui_interact(owner)
+
+/datum/action/changeling/evolution_menu/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = FALSE, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_always_state)
```
AffectedArc07, line 37: Unsure about the `always` state. Should changelings be able to modify their abilities even when dead?

SteelSlayer: To be honest, I used something that worked for the time being, because some of the other states I tried weren't allowing me to open the UI, so I chose this. If there's a better state I can use here, I'll switch to that one. 

You could also say that changelings aren't really dead I guess. Even when they're "dead", they can revive.

SteelSlayer:
Ok, I was trying to implement Mochi's changes but unfortunately I have no idea what I'm doing and just running around in circles at this point, because I don't know how to get the results I want. 

Getting this to work would probably require me just asking for help from a bunch of people, and I don't want to have to do that. I think I'll go ahead and call this ready to go for now, unless it's required for me to implement this change.

PR #14714, opened 2020-10-24
Sec ERT Laser Carbine Rework

Leanfrog:
## What Does This PR Do
This PR changes the laser carbine from being some sort of hybrid projectile energy weapon abomination into a standard energy weapon that relies on chargers to restore its ammo meaning it can be used more than 30 times allowing the weapon to be useful outside of a brief engagement with a crew antag. As compensation for losing the magazines and thus amount of shots one can make per fight the ammo capacity has also been increased slightly to 24 shots

## Why It's Good For The Game
Currently the laser carbine is absolutely pitiful against biohazards primarily due to the lack of ability to recharge/reload it once you use up your 3 magazines you start with, all the while this weapon has an arguably absurd potential damage against crew with a maximum damage of around 1.2k burn if all shots were to hit their mark, this PR brings it more in line with other energy weapons while still retaining the superiority that most CC equipment tends to have as this weapon has double the capacity of a standard laser rifle, the strongest energy weapon the crew can get ahold of.

## Changelog
:cl:
tweak: Total rework of the laser carbine, it no longer relies on external cartridges for ammo and instead functions like a standard energy weapon
/:cl:


Spartan6:
I'd be more in fashion of keeping the rifle as is, but allowing the magazines to be refilled in some fashion. The uniqueness of the weapon does not mean it should be removed. The Red ERT even have a backup gun to deal with issues once they run out of ammo.

Against this in its current form.

TDSSS:
IMO, people that think the lasercarbine is bad against biohazards/blobs don't know how to use them. Yes, they have bad sustain but they're incredible burst damage. Fly your ert shuttle to a place where you can flank the blob and concentrate fire and you can easily burn your way through to a blob core.

Of course, most ert being bald and wasting all their ammo does make a case for changing how they work.

hal9000PR:
Why not just make it printable from the protolathe with end game tech levels? I feel like if you did it this way you'd keep the uniqueness of the gun while still fixing the issue.

Kyep:
The laser carbine can do just fine against blobs, when the ERT players use it correctly. You just park the ERT shuttle in space next to the blob core (which you can usually see using the nav console, since its usually next to space), have the entire ERT jump out, and all throw their flashbangs, then focus-fire the core. Blob will die unless they instantly spam reflectives (and sometimes even then, if the ERT just rotate around the core to avoid the reflectives).

However, you're right that its limited ammo is very much a trap. If you use the weapon incorrectly (ie: spam-firing it without doing much damage) you will quickly run out of ammo and the weapon becomes useless. While you might see that as a bad thing, it also limits the amount of damage an antag can do if they steal an ERT member's gun.

Leanfrog:
Right, after thinking this over, I may just make an alternative PR to this that gives sec ERT a data disk in their bag that contains the blueprint for more cartridges, closing this

Kyep:
If you're going to do that, then only give the disk to the ERT commander. Otherwise, you risk creating a situation where an antag can kill an ERT member, then get indefinite usage out of their normally limited-ammo weapon.

PR #6764, opened 2017-03-14
Removes the Gatfruit from Botany

SamHPurp:
Removes the Gatfruit tree from Botany, removing their ability to grow .357 Revolvers....

All reactions to this when seen/grown in game has been one of complete and utter disbelief in IC and OOC overwhelmingly.

:cl: Purpose2
tweak: Removes the Gatfruit tree from Botany.
/ :cl:

PR #14715, opened 2020-10-24
Shoots the botanical gun industry

Qwertytoforty:
## What Does This PR Do

Removes gatfruit, and commented references to gatfruit, from the code. Adds the 10% carbon gene gatfruit has to the carpet grass, as a way for botanists to make synthflesh still. (Black carpet, black carbon.) Adds extra random strange seeds to the diona loot drop to replace gatfruit, with a chance of 12 vs 10 to make the total odds add up to 40

## Why It's Good For The Game

Mass producible .357 revolvers is not something anyone should have, be it traitors, cult, pod dionas, or the crew. Too often do miners bring a literal bag full of revolvers to the station, which is either used for nigh infinite ammo for antaging purposes, or it reaches botany, who grows it, and distributing it, leading to a massive mess for security as they now have to arrest 25% of the station for S class contraband. That, or sec gets it, uses it as it comes from a plant, or crew uses it to dunk whatever the major midround is. In any case, it is not something the crew should have even rare access to. Thus, this PR removes it.

## Changelog
:cl:
del: Removes gatfruit.
tweak: Carbon is now found in carpet seeds.
tweak: The extra seed crates on the pod plant ship now have a chance for extra strange seeds.
/:cl:


Dashsora:
I wouldn't add carbon to carpet. Getting a carbon gene as a botanist is a godsend for even better healing tomatoes - and they're strong enough with styptic/silver sulfa.. making this gene so easily accessible would be a major buff to botany.

Gratfruit definitely needs changes.. but this might be a little too extreme imo, how about limitting gratfruit yield to one fruit always... maybe you could even make it so the revolver comes unloaded !

Varghh:
Gatfruit adds nothing to the game, it's mostly a situation where every single crewmember is armed with a super strong weapons and end up validhunting, or security use them to 2 shots anyone down without much effort, everytime shift where gatfruit is hella involved ends up to be a pain in the ass for everyone. Totally agree with the removal.

variableundefined (approved):
I'm on board with removing Gatfruit.

Dashsora:
> Gatfruit adds nothing to the game, it's mostly a situation where every single crewmember is armed with a super strong weapons and end up validhunting, or security use them to 2 shots anyone down without much effort, everytime shift where gatfruit is hella involved ends up to be a pain in the ass for everyone. Totally agree with the removal.

but isn't that an IC problem though ? They are s-class contraband and should be treated as such, why would sec be using them ? Why are people allowed to blatantly validhunt with .357s ? Why aren't all botanists immediately stopped from producing them ?

 Furthermore, if you make them much harder to mass produce, and remove their ammunition adding an extra step required to actually make it what it is- you'd severly reduce the shitshow that we've seen where the botanist just handed out hundreds of .357s. Gratfruit becomes a way for a lucky botanist/miner tot to get a .357, instead of a means of mass armament. 

I'm just adding to the discussion - Absolutely wouldn't miss it if it were gone - Only maybe a little for the mentionned carbon gene.

Spartan6:
IC problem it may be, but when each time it is used, it becomes an IC issue where NOBODY wins, it's not worth it.

Fox-McCloud (changes requested):
https://github.com/ParadiseSS13/Paradise/pull/6764

Was attempted once before. This is a rare occurrence at best, and handling it with a scalpel is preferable to handing it with a butcher cleaver.

If the issue is a result of the pod people (which I suspect it the culprit) then address that.

Taking it away entirely when it only happened a few times, via other means, is a bit extreme.

Generally, balance is important, but a few, rare, outsized events and occurrences adds a healthy amount of variety, even if it's "unfair".

TDSSS:
Personally, I think having an extremely rare, exotic plant that makes guns is fine. It just shouldn't make a very good gun.

TheMaveran:
I think you could just have the production speed and yield be locked and the plant would be “balanced”. Or have the fruit make stetchkins or... anything that isn’t one of the best guns in game really.

Dashsora:
> I think you could just have the production speed and yield be locked and the plant would be “balanced”. Or have the fruit make stetchkins or... anything that isn’t one of the best guns in game really.

50% for it to be a trick revolver

TheMaveran:
Actually making it give a random ballistic gun would be hilarious. But also nerf it stats wise

Cocacolagua:
peashooter

Kyep (approved):

Qwertytoforty:
Gatfruit was removed from the seed vault, so will close this PR.

PR #14719, opened 2020-10-24
The road to Initialize() stability - Enforcing parent call on New(), Initialize() and Destroy()

AffectedArc07:
## What Does This PR Do
This PR adds in `SHOULD_CALL_PARENT()` annotations to the code for `New()`, `Initialize()` and `Destroy()` procs respectively. For the full initialization system (and GC but thats already working very well) to be fully functional, these calls need to go all the way down to parent level, otherwise stuff breaks badly and you get atoms not initializing even when they should

## Why It's Good For The Game
The road away from /New() is a good road

## Changelog
:cl: AffectedArc07
fix: Fixed some code paths not fully calling parents when they should
/:cl:


code/datums/datum.dm
```diff
@@ -20,6 +20,7 @@
 // This should be overridden to remove all references pointing to the object being destroyed.
 // Return the appropriate QDEL_HINT; in most cases this is QDEL_HINT_QUEUE.
 /datum/proc/Destroy(force = FALSE, ...)
+	SHOULD_CALL_PARENT(TRUE)
```
AffectedArc07, line 23: This here is the magic. During a travis run, dreamchecker will look for these annotations and trace entire call stacks down, failing if you do not call the parent on a proc

Note, this applies all the way down a call stack

code/modules/research/message_server.dm
```diff
@@ -222,6 +222,7 @@ GLOBAL_DATUM(blackbox, /obj/machinery/blackbox_recorder)
 
 	//Only one can exsist in the world!
 /obj/machinery/blackbox_recorder/New()
+	SHOULD_CALL_PARENT(FALSE) // TODO: I still need to shoot this
```
Kyep, line 225: 👍

code/game/objects/effects/manifest.dm
```diff
@@ -1,21 +0,0 @@
-/obj/effect/manifest
```
AffectedArc07, line 1: Deleted because not used anywhere

code/modules/mob/living/silicon/decoy/decoy.dm
```diff
@@ -6,12 +6,6 @@
 	canmove = 0
 	a_intent = INTENT_HARM // This is apparently the only thing that stops other mobs walking through them as if they were thin air.
 
-/mob/living/silicon/decoy/New()
-	src.icon = 'icons/mob/ai.dmi'
-	src.icon_state = "ai"
-	src.anchored = 1
-	src.canmove = 0
```
AffectedArc07, line 13: It appears that when you make this call parent it makes an issue with huds

AffectedArc07:
Latest commit now fixes the wonky initialization of AI decoys and costume spawners. Both of those need refactors at some point, but for now, this will do.

code/game/turfs/space/space.dm
```diff
@@ -19,6 +19,7 @@
 	plane = PLANE_SPACE
 
 /turf/space/Initialize(mapload)
+	SHOULD_CALL_PARENT(FALSE)
```
AffectedArc07, line 22: Space turfs do their own stuff explicitly and should NOT call parent

code/game/turfs/turf.dm
```diff
@@ -40,6 +40,7 @@
 	var/shoe_walking_volume = 20
 
 /turf/Initialize(mapload)
+	SHOULD_CALL_PARENT(FALSE)
```
AffectedArc07, line 43: Same with regular turfs

code/modules/awaymissions/mission_code/wildwest.dm
```diff
@@ -135,9 +135,6 @@
 	var/triggerproc = "triggerrad1" //name of the proc thats called when the mine is triggered
 	var/triggered = 0
 
-/obj/effect/meatgrinder/New()
-	icon_state = "blobpod"
```
AffectedArc07, line 139: We had a fair few `/New()` instances which just set vars to static vars which were already set

code/modules/mob/living/silicon/silicon.dm
```diff
@@ -39,15 +39,18 @@
...
-	var/datum/atom_hud/data/diagnostic/diag_hud = GLOB.huds[DATA_HUD_DIAGNOSTIC]
-	diag_hud.add_to_hud(src)
-	diag_hud_set_status()
-	diag_hud_set_health()
 	add_language("Galactic Common")
 	init_subsystems()
 	RegisterSignal(SSalarm, COMSIG_TRIGGERED_ALARM, .proc/alarm_triggered)
 	RegisterSignal(SSalarm, COMSIG_CANCELLED_ALARM, .proc/alarm_cancelled)
 
+/mob/living/silicon/Initialize()
```
AffectedArc07, line 47: The HUD list is made during `Initialize()` at `/mob` level, and making decoy AI cores parent call `New()` caused them to get HUDs without it working. This fixes that.

variableundefined (approved):

code/modules/awaymissions/mission_code/wildwest.dm
Kiyahitayika: `var/triggered = 0`

This is a boolean. Please change to FALSE/TRUE to please the nerd.

code/modules/food_and_drinks/drinks/bottler/bottler.dm
```diff
@@ -23,6 +23,7 @@
 	var/bottling = 0
```
Kiyahitayika, line 23: Boolean here. 0/1 should be FALSE/TRUE

code/modules/mob/living/silicon/decoy/decoy.dm
Kiyahitayika: `canmove = 0`
Is that a boolean?!

code/modules/awaymissions/mission_code/wildwest.dm
AffectedArc07: Not happening, the whole point of this PR is just to add parent calls, not refactor half the files.

code/modules/food_and_drinks/drinks/bottler/bottler.dm
AffectedArc07: Not happening, the whole point of this PR is just to add parent calls, not refactor half the files.

farie82 (changes requested):
All `New`s have no return value so `. = ..()` is not required and is misleading. Best to replace them with just `..()`

AffectedArc07:
> All `New`s have no return value so `. = ..()` is not required and is misleading. Best to replace them with just `..()`

Disagree. Its also standard format for a lot of things to pass parent value up the chain even if its not using, not to mention this will aid `Initialize()` conversion.

farie82:
> Disagree. Its also standard format for a lot of things to pass parent value up the chain even if its not using, not to mention this will aid `Initialize()` conversion.

It's not that things are not using it though. It's more that it does never return anything since it's an internal byond call.
For the aiding the Init conversion. It will allow people to blindly convert them to `Initialize` easier. Quite a few times you need to change the order anyway of initialization. Or you at least need to think about it.

Issue #7164, opened 2017-04-24
People who are labeled unknown from facial damage still have a character description


ModDownloading:
**Problem Description**:
If someone becomes unknown due to facial damage, their character description is still visible upon examine, potentially revealing their identity.

**What did you expect to happen**: 
The unknown person should not be easily recognizable if they got mutilated/husked/facially disfigured.

**What happened instead**:
When examined, the character description is still visible.

**Why is this bad/What are the consequences:**
This enables easy identification of people even when they should not be identified, especially if they are dead or otherwise incapable of speech (so people would not recognize them by voice either).

**Steps to reproduce the problem**:
Get facially disfigured, then examine yourself or have someone examine you

**Possibly related stuff (which gamemode was it? What were you doing at the time? Was
anything else out of the ordinary happening?)**: I am unsure if this bug applies to all injuries to the face, or to beheadings as well, but it could. Also, this issue is related to #6405.

PR #14722, opened 2020-10-25
No longer shows flavour text when examining someone who's mutilated

Adrer:
## What Does This PR Do
Makes it so that when someone becomes unknown due to facial damage, flavour text no longer shows up upon examining.

fixes: #7164

## Why It's Good For The Game
Someone who's mutilated/husked/facially disfigured should not be easily identifiable.

## Changelog
:cl:
tweak: People with a disfigured face no longer have flavour text upon examination.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -388,7 +388,7 @@
 		msg += "<span class = 'deptradio'>Physical status:</span> <a href='?src=[UID()];medical=1'>\[[medical]\]</a>\n"
 		msg += "<span class = 'deptradio'>Medical records:</span> <a href='?src=[UID()];medrecord=`'>\[View\]</a> <a href='?src=[UID()];medrecordadd=`'>\[Add comment\]</a>\n"
 
-	if(print_flavor_text() && !is_destroyed["head"])
+	if(print_flavor_text() && get_organ("head"))
 		if(!(skipface || get_organ("head").disfigured || src.cloneloss > 50 || (HUSK in src.mutations)))
 			msg += "[print_flavor_text()]\n"
 
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -388,9 +388,11 @@
 		msg += "<span class = 'deptradio'>Physical status:</span> <a href='?src=[UID()];medical=1'>\[[medical]\]</a>\n"
 		msg += "<span class = 'deptradio'>Medical records:</span> <a href='?src=[UID()];medrecord=`'>\[View\]</a> <a href='?src=[UID()];medrecordadd=`'>\[Add comment\]</a>\n"
 
-	if(print_flavor_text() && get_organ("head"))
-		if(!(skipface || get_organ("head").disfigured || src.cloneloss > 50 || (HUSK in src.mutations)))
-			msg += "[print_flavor_text()]\n"
+	if(print_flavor_text() && !skipface)
+		if(get_organ("head"))
+			var/obj/item/organ/external/head/H = get_organ("head")
+			if(!(H.disfigured || cloneloss > 50 || (HUSK in mutations)))
+				msg += "[print_flavor_text()]\n"
 
 	msg += "*---------*</span>"
 	if(pose)
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -391,7 +391,7 @@
 	if(print_flavor_text() && !skipface)
 		if(get_organ("head"))
 			var/obj/item/organ/external/head/H = get_organ("head")
-			if(!(H.disfigured || cloneloss > 50 || (HUSK in mutations)))
+			if(!H.disfigured)
 				msg += "[print_flavor_text()]\n"
 
 	msg += "*---------*</span>"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -388,9 +388,9 @@
 		msg += "<span class = 'deptradio'>Physical status:</span> <a href='?src=[UID()];medical=1'>\[[medical]\]</a>\n"
 		msg += "<span class = 'deptradio'>Medical records:</span> <a href='?src=[UID()];medrecord=`'>\[View\]</a> <a href='?src=[UID()];medrecordadd=`'>\[Add comment\]</a>\n"
 
-
-	if(print_flavor_text() && !skipface)
-		msg += "[print_flavor_text()]\n"
+	if(print_flavor_text() && get_organ("head"))
+		if(!(skipface || get_organ("head").disfigured || src.cloneloss > 50 || (HUSK in src.mutations)))
```
farie82, line 392: One to many ()'s here. `src` is also implied so that can be removed.
The cloning damage is also already retrieved on line 268. It is however put in a temp var so it's (currently) not saved.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -388,9 +388,9 @@
 		msg += "<span class = 'deptradio'>Physical status:</span> <a href='?src=[UID()];medical=1'>\[[medical]\]</a>\n"
 		msg += "<span class = 'deptradio'>Medical records:</span> <a href='?src=[UID()];medrecord=`'>\[View\]</a> <a href='?src=[UID()];medrecordadd=`'>\[Add comment\]</a>\n"
 
-
-	if(print_flavor_text() && !skipface)
-		msg += "[print_flavor_text()]\n"
+	if(print_flavor_text() && get_organ("head"))
```
farie82, line 391: I'd put `!skipface` in here instead of get_organ("head"). Then retrieve the head organ if this if statement is true retrieve the head and cache it in a variable.
The linter is also failing due to you directly using the output of `get_organ` to check `.disfigured`.

farie82 (changes requested):

code/modules/mob/living/carbon/human/examine.dm
Adrer: I don't see where these unneeded parentheses are.

farie82: Ah i just noticed the ! before it. All good for the parentheses!

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -388,9 +388,11 @@
 		msg += "<span class = 'deptradio'>Physical status:</span> <a href='?src=[UID()];medical=1'>\[[medical]\]</a>\n"
 		msg += "<span class = 'deptradio'>Medical records:</span> <a href='?src=[UID()];medrecord=`'>\[View\]</a> <a href='?src=[UID()];medrecordadd=`'>\[Add comment\]</a>\n"
 
-
 	if(print_flavor_text() && !skipface)
-		msg += "[print_flavor_text()]\n"
+		if(get_organ("head"))
+			var/obj/item/organ/external/head/H = get_organ("head")
+			if(!(H.disfigured || cloneloss > 50 || (HUSK in mutations)))
```
variableundefined, line 394: According to what I see in the code, if someone's husked, they would be automatically disfigured. 

Is there a situation where: 
- Someone is mutilated solely due to clone loss, yet somehow not disfigured
- Someone is husked yet somehow not disfigured? 

I think disfigured should be the only check here, else people's flavour text could potentially disappear after plastic surgery but before their clone damage or husk mutations is cured (Not sure if there's an edgecase like that)

variableundefined (changes requested):

code/modules/mob/living/carbon/human/examine.dm
Adrer: You're absolutely correct on the removal of the husk check, there should not be a case in which someone's husked without their head being disfigured.

About the clone loss, this is solely taken from mob/living/carbon/human/human.dm in which someone's identity is returned as unknown if clone loss is over 50 (I don't exactly know why cellular damage would make someone unidentifiable, I'm just trying to be consistent with get_face_name() here). It is very much possible (though rare) to cause over 50 cloneloss without being disfigured eg. by using the decloner on them.

```//Returns "Unknown" if facially disfigured and real_name if not. Useful for setting name when polyacided or when updating a human's name variable
/mob/living/carbon/human/proc/get_face_name()
	var/obj/item/organ/external/head = get_organ("head")
	if(!head || head.disfigured || cloneloss > 50 || !real_name || (HUSK in mutations))	//disfigured. use id-name if possible
		return "Unknown"
	return real_name
```

variableundefined: Make sense to me. Can you change it to just check for disfigured?

I'd also request that the cloneloss check be removed from get face name, but that should be a separate PR as I am not familiar with the history behind it.

variableundefined (approved):

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -388,9 +388,11 @@
 		msg += "<span class = 'deptradio'>Physical status:</span> <a href='?src=[UID()];medical=1'>\[[medical]\]</a>\n"
 		msg += "<span class = 'deptradio'>Medical records:</span> <a href='?src=[UID()];medrecord=`'>\[View\]</a> <a href='?src=[UID()];medrecordadd=`'>\[Add comment\]</a>\n"
 
-
 	if(print_flavor_text() && !skipface)
-		msg += "[print_flavor_text()]\n"
+		if(get_organ("head"))
+			var/obj/item/organ/external/head/H = get_organ("head")
+			if(!H.disfigured)
```
moxian, line 394: I believe this will runtime if they are missing a head (so `H==null`)
```suggestion
			if(H && !H.disfigured)
```

moxian, line 394: i happened to get more confused that I should have been reading this, and assumed that H can be `null` (which it cannot). I'd prefer something like
```suggestion
		var/obj/item/organ/external/head/H = get_organ("head")
		if(H && !H.disfigured)
```
(One can even say that this is "faster", because it's only one call to `get_organ` instead of two, but that's honestly irrelevant)

Kyep (approved):

Kiyahitayika:
Definitely needed.

PR #14724, opened 2020-10-25
Prevents MMI's from being inserted into mechs the user does not have access to

Qwertytoforty:
## What Does This PR Do

Adds a check for the users ID access when the the user attempts to put an MMI in the mech. If they do not have access, they can not put it in the mech. Also turns some 1/0 into TRUE/FALSE

## Why It's Good For The Game

If you do not have access to an ert mech, you should not be able to _put an MMI in it_ to bypass this restriction. 

## Images of changes

image

## Changelog
:cl:
tweak: You now need the required access to operate a mech to put the MMI in it.
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -1130,23 +1130,26 @@
 			occupant << sound(nominalsound, volume = 50)
 		if(state)
 			H.throw_alert("locked", /obj/screen/alert/mech_maintenance)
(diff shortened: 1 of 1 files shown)


SteelSlayer (approved):

variableundefined (approved):
Approved for design

Kyep (approved):

Issue #14723, opened 2020-10-25
Cant set ID tag on mass drives


AffectedArc07:
**Issue Description**:
Cant set ID tag on mass drives

You can open the UI with a multitool but clicking the link doesnt open an entry popup.

**What did you expect to happen**: 
An entry popup

**What happened instead**:
No popup

**Why is this bad/What are the consequences**:
Cant set ID tags on mass drivers

**Steps to reproduce the problem**:
Use multitool on mass driver and attempt to set tag

**When did the problem start happening**:
Unsure

**Extra information**:
N/A

PR #14725, opened 2020-10-25
Allow mass drivers and mass driver buttons to be linked again

Adrer:
## What Does This PR Do
Allows you to edit mass driver and mass driver button tags again.

fixes: #14723

## Why It's Good For The Game
Tags should be editable so you can actually properly link these.

## Changelog
:cl:
fix: Mass driver and mass driver button id tags are now editable
/:cl:

### DIFF
--- a/code/game/machinery/buttons.dm
+++ b/code/game/machinery/buttons.dm
@@ -89,18 +89,6 @@
 	<li><b>Logic ID Tag:</b> [format_tag("Logic ID Tag", "logic_id_tag")]</li>
 	</ul>"}
 
-/obj/machinery/driver_button/multitool_topic(var/mob/user, var/list/href_list, var/obj/O)
-	if("set_id" in href_list)
-		var/newid = copytext(reject_bad_text(input(user, "Specify the new ID tag for this machine", name, id_tag) as null|text), 1, MAX_MESSAGE_LEN)
-		if(!newid)
-			return
-		
-		id_tag = newid
-
-		return TRUE
-
-	return ..()
-
 /obj/machinery/driver_button/attack_hand(mob/user as mob)
 
 	add_fingerprint(usr)
--- a/code/game/machinery/buttons.dm
+++ b/code/game/machinery/buttons.dm
@@ -91,7 +91,7 @@
 
 /obj/machinery/driver_button/multitool_topic(var/mob/user, var/list/href_list, var/obj/O)
 	if("set_id" in href_list)
-		var/newid = copytext(reject_bad_text(input(usr, "Specify the new ID tag for this machine", src, src.id_tag) as null|text), 1, MAX_MESSAGE_LEN)
+		var/newid = copytext(reject_bad_text(input(user, "Specify the new ID tag for this machine", name, id_tag) as null|text), 1, MAX_MESSAGE_LEN)
 		if(!newid)
 			return
 		
--- a/code/game/machinery/buttons.dm
+++ b/code/game/machinery/buttons.dm
@@ -84,10 +84,22 @@
 /obj/machinery/driver_button/multitool_menu(var/mob/user, var/obj/item/multitool/P)
 	return {"
 	<ul>
-	<li><b>ID Tag:</b> [format_tag("ID Tag","id_tag")]</li>
+	<li><b>ID Tag:</b> [format_tag("ID Tag","id_tag","set_id")]</li>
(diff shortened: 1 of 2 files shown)


code/game/machinery/buttons.dm
```diff
@@ -84,11 +84,23 @@
...
 	<ul>
-	<li><b>ID Tag:</b> [format_tag("ID Tag","id_tag")]</li>
+	<li><b>ID Tag:</b> [format_tag("ID Tag","id_tag","set_id")]</li>
 	<li><b>Logic Connection:</b> <a href='?src=[UID()];toggle_logic=1'>[logic_connect ? "On" : "Off"]</a></li>
 	<li><b>Logic ID Tag:</b> [format_tag("Logic ID Tag", "logic_id_tag")]</li>
 	</ul>"}
 
+/obj/machinery/driver_button/multitool_topic(var/mob/user, var/list/href_list, var/obj/O)
+	if("set_id" in href_list)
+		var/newid = copytext(reject_bad_text(input(usr, "Specify the new ID tag for this machine", src, src.id_tag) as null|text), 1, MAX_MESSAGE_LEN)
```
AffectedArc07, line 94:
```suggestion
		var/newid = copytext(reject_bad_text(input(user, "Specify the new ID tag for this machine", name, id_tag) as null|text), 1, MAX_MESSAGE_LEN)
```

code/game/machinery/mass_driver.dm
```diff
@@ -37,9 +37,21 @@
 /obj/machinery/mass_driver/multitool_menu(var/mob/user, var/obj/item/multitool/P)
 	return {"
 	<ul>
-	<li>[format_tag("ID Tag","id_tag")]</li>
+	<li>[format_tag("ID Tag","id_tag","set_id")]</li>
 	</ul>"}
 
+/obj/machinery/mass_driver/multitool_topic(var/mob/user, var/list/href_list, var/obj/O)
+	if("set_id" in href_list)
+		var/newid = copytext(reject_bad_text(input(usr, "Specify the new ID tag for this machine", src, src.id_tag) as null|text), 1, MAX_MESSAGE_LEN)
```
AffectedArc07, line 45:
```suggestion
		var/newid = copytext(reject_bad_text(input(user, "Specify the new ID tag for this machine", name, id_tag) as null|text), 1, MAX_MESSAGE_LEN)
```

AffectedArc07 (changes requested):
As a general practice, if you have a `user` arg, use that over `usr`. Also implied `src.` usage is bad

code/game/machinery/buttons.dm
```diff
@@ -84,11 +84,23 @@
 /obj/machinery/driver_button/multitool_menu(var/mob/user, var/obj/item/multitool/P)
 	return {"
 	<ul>
-	<li><b>ID Tag:</b> [format_tag("ID Tag","id_tag")]</li>
+	<li><b>ID Tag:</b> [format_tag("ID Tag","id_tag","set_id")]</li>
 	<li><b>Logic Connection:</b> <a href='?src=[UID()];toggle_logic=1'>[logic_connect ? "On" : "Off"]</a></li>
 	<li><b>Logic ID Tag:</b> [format_tag("Logic ID Tag", "logic_id_tag")]</li>
 	</ul>"}
 
+/obj/machinery/driver_button/multitool_topic(var/mob/user, var/list/href_list, var/obj/O)
```
SteelSlayer, line 92: So this this proc you've written is very similar to the base machinery one:
```dm
/obj/machinery/proc/multitool_topic(mob/user, list/href_list, obj/O)
	if("set_id" in href_list)
		if(!("id_tag" in vars))
			warning("set_id: [type] has no id_tag var.")
		var/newid = copytext(reject_bad_text(input(usr, "Specify the new ID tag for this machine", src, src:id_tag) as null|text),1,MAX_MESSAGE_LEN)
		if(newid)
			src:id_tag = newid
			return TRUE
```
It does have the cursed `:` format but still, it looks like it does everything your new proc does. Do you know why the base proc doesn't work for the button and driver?

Adrer: Turns out I'm an idiot.
You're absolutely right, current base machinery proc works just fine, I'd just wrongly assumed it didn't work on the mass driver stuff for whatever reason. All that was missing was the set_id format tags. Should've probably tested just that first.

SteelSlayer (approved):

AffectedArc07 (approved):

PR #14728, opened 2020-10-25
Light Flickers Now Process Asynchronously Again

### LINKED ISSUE #14727: Light flickering cycles through lights one by one when your boo! ability affects multiple lights
**Issue Description**:

When using your boo! ability as a ghost if your boo! affected multiple light, the lights flicker one by one instead of all of them at once. 

I expected all of the lights affected by the boo! to flicker at the same time.

**What happened instead**:

One light flickered and the other, which was also affected by the boo waited for the previous light to finish before starting to flicker.

Fairly sure this is not intended behavior and it reduces the spookiness levels of light flickering if they flicker one by one instead of all of them at the same time.

1. Be a ghost.
2. Either change the range of your boo ability or find a room, where multiple lights are nearby each other.
3. Use your Boo! ability so that it affects multiple lights at the same time.

Pretty sure it began happening, when this PR was merged: #14396 

**Extra information**:

A gif of the issue in effect. You can see the left one start flickering first and once it finished the right one starts
1lightatatime

S-CR:
## What Does This PR Do
Calls invoke_async to process the light flicker loop while allowing for a return.
fixes #14727 

As an alternative solution, this bug is also fixed by #14543

## Why It's Good For The Game
Lights together strong

## Changelog
:cl: Dave
fix: Boo'ing lights no longer politely waits for them
/:cl:

### DIFF
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -522,17 +522,27 @@
 /obj/machinery/light/proc/flicker(amount = rand(10, 20))
 	if(flickering)
 		return
-	if(on)
-		flickering = TRUE
-		for(var/i in 1 to amount)
-			if(status != LIGHT_OK)
-				break
-			on = !on
-			update() // No param means that ghosts can burn out lights
-			sleep(rand(5, 15))
-		if(status == LIGHT_OK)
-			on = TRUE
-			update(FALSE)
+
+	if(!on)
+		return
+
+	flickering = TRUE
+	INVOKE_ASYNC(src, /obj/machinery/light/.proc/flicker_event, amount)
+
+/**
+  * Flicker routine for the light.
+  * Called by invoke_async so the parent proc can return immediately.
+  */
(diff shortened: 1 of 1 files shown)


code/modules/power/lighting.dm
```diff
@@ -522,17 +522,28 @@
...
+
+	flickering = TRUE
+	INVOKE_ASYNC(src, /obj/machinery/light/.proc/flicker_event, amount)
+
+/**
+  * Flicker routine for the light.
+  * Called by invoke_async so the parent proc can return immediately.
+  */
+/obj/machinery/light/proc/flicker_event(amount)
+	flickering = TRUE
```
AffectedArc07, line 537: Any reason you set `flickering` here as well as on `529`? Seems a little redundant, no?

S-CR: Dangling local change, I check before I push I promise.

AffectedArc07 (approved):

PR #14730, opened 2020-10-25
Operating Computer - Patient Status

Kiyahitayika:
## Special Thanks
**HUGE thanks to @moxian for helping with this PR!**
Their feedback has been invaluable in cleaning this code up and making the new code much neater and more efficient. I definitely consider them a co-author on this project since they gave me "snippets" of code that I then implemented in the design.
Credit also to @farie82 for helping out with the initial thought process around this.

## What Does This PR Do
Operating Computers now give updates when the patient's status changes in terms of being awake, asleep, or dead.

## Why It's Good For The Game
When you're operating on a patient, it's often important to know if they're awake or asleep. This is particularly handy when you're operating on patients who cannot use anesthesia, such as Vox and Plasmamen - if they are using the Sleep verb, they can wake up and it's good to know when this changes. The "dead" status also accounts for Changeling fake death. No meta here, please!

This PR also goes through and tries to clean up old code in `OpTable.dm` and `Operating.dm`.

## Images of changes
**Note:** "currently" has been changed to "now" // _"Patient is currently Awake"_ will read as _"Patient is **now** Awake"_
\-
Awake:
image
\-
Then the patient fell Asleep:
image
\-
After a brief nap, they became Awake, then returned to being Asleep, then were suddenly Dead.
Here's Dead:
image

## Changelog
:cl:
add: A check on an Operating Table's patient's awake/asleep/dead status. The Operating Computer will now alert doctors when this changes.
tweak: Clean-up of the Operating Table and Operating Computer codes.
/:cl:

### DIFF
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -51,17 +51,14 @@
 /obj/machinery/optable/MouseDrop_T(atom/movable/O, mob/user)
 	if(!ishuman(user) && !isrobot(user)) //Only Humanoids and Cyborgs can put things on this table
 		return
-	if(!check_table()) //Is the Operating Table empty?
+	if(!check_table()) //If the Operating Table is occupied, you cannot put someone else on it
 		return
 	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
 		return
-	if(!ishuman(O)) //Only Humanoids can go on this table
+	var/mob/living/carbon/C = O
+	if(!C) //Only Carbon mobs can go on this table, if C is null, get out of here
 		return
-	if(isanimal(O) || isrobot(O)) //Animals and Cyborgs do not go on the table
-		return
-	var/mob/living/L = O
-	take_patient(L, user)
-	return
+	take_patient(C, user)
 
 /obj/machinery/optable/proc/check_patient()
 	if(locate(/mob/living/carbon/human, loc))
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -48,19 +48,19 @@
 		return FALSE
 
 
-/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, var/mob/P)
-	if(usr.stat || (!ishuman(P) && !isrobot(P)) || !check_table())
+/obj/machinery/optable/MouseDrop_T(atom/movable/O, mob/user)
+	if(!ishuman(user) && !isrobot(user)) //Only Humanoids and Cyborgs can put things on this table
 		return
-	if(!check_table()) //Is the Operating Table currently occupied?
+	if(!check_table()) //Is the Operating Table empty?
 		return
-	if(P.restrained() || P.buckled || P.incapacitated()) //Folks who are restrained or otherwise incapacitated cannot get on the table
+	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
 		return
-	if(!ismob(O)) //Humans only
+	if(!ishuman(O) //Only Humanoids can go on this table
 		return
-	if(istype(O, /mob/living/simple_animal) || istype(O, /mob/living/silicon)) //animals and robots dont fit
+	if(isanimal(O)) || isrobot(O)) //Animals and Cyborgs do not go on the table
 		return
 	var/mob/living/L = O
-	take_patient(L, usr)
+	take_patient(L, user)
 	return
 
 /obj/machinery/optable/proc/check_patient()
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -8,12 +8,10 @@
 	use_power = IDLE_POWER_USE
 	idle_power_usage = 1
 	active_power_usage = 5
-	var/mob/living/carbon/human/victim = null
-	var/strapped = 0.0
-
-	var/obj/machinery/computer/operating/computer = null
+	var/mob/living/carbon/human/patient
+	var/obj/machinery/computer/operating/computer
 	buckle_lying = -1
-	var/no_icon_updates = 0 //set this to 1 if you don't want the icons ever changing
+	var/no_icon_updates = FALSE //set this to TRUE if you don't want the icons ever changing
 	var/list/injected_reagents = list()
 	var/reagent_target_amount = 1
 	var/inject_amount = 1
@@ -29,10 +27,9 @@
 /obj/machinery/optable/Destroy()
 	if(computer)
 		computer.table = null
-		computer.victim = null
 		computer = null
-	if(victim)
-		victim = null
+	if(patient)
+		patient = null
 	return ..()
 
 /obj/machinery/optable/attack_hulk(mob/living/carbon/human/user, does_attack_animation = FALSE)
@@ -43,40 +40,41 @@
 		return TRUE
 
 /obj/machinery/optable/CanPass(atom/movable/mover, turf/target, height=0)
-	if(height==0) return 1
-
+	if(height==0)
+		return TRUE
 	if(istype(mover) && mover.checkpass(PASSTABLE))
-		return 1
+		return TRUE
 	else
-		return 0
+		return FALSE
 
 
-/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
-	if(usr.stat || (!ishuman(user) && !isrobot(user)) || user.restrained() || !check_table(user) || user.IsWeakened() || user.stunned)
+/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, var/mob/P)
+	if(usr.stat || (!ishuman(P) && !isrobot(P)) || !check_table())
 		return
-
-	if(!ismob(O)) //humans only
+	if(!check_table()) //Is the Operating Table currently occupied?
 		return
-
+	if(user.restrained() || user.buckled || user.incapacitated()) //Folks who are restrained or otherwise incapacitated cannot get on the table
+		return
+	if(!ismob(O)) //Humans only
+		return
 	if(istype(O, /mob/living/simple_animal) || istype(O, /mob/living/silicon)) //animals and robots dont fit
 		return
-
 	var/mob/living/L = O
-	take_victim(L,usr)
+	take_patient(L, usr)
 	return
 
-/obj/machinery/optable/proc/check_victim()
-	if(locate(/mob/living/carbon/human, src.loc))
-		var/mob/living/carbon/human/M = locate(/mob/living/carbon/human, src.loc)
+/obj/machinery/optable/proc/check_patient()
+	if(locate(/mob/living/carbon/human, loc))
+		var/mob/living/carbon/human/M = locate(/mob/living/carbon/human, loc)
 		if(M.lying)
-			src.victim = M
+			patient = M
 			if(!no_icon_updates)
 				icon_state = M.pulse ? "table2-active" : "table2-idle"
-			return 1
-	src.victim = null
+			return TRUE
+	patient = null
 	if(!no_icon_updates)
 		icon_state = "table2-idle"
-	return 0
+	return FALSE
 
 /obj/machinery/optable/Crossed(atom/movable/AM, oldloc)
 	. = ..()
@@ -84,19 +82,21 @@
 		to_chat(AM, "<span class='danger'>You feel a series of tiny pricks!</span>")
 
 /obj/machinery/optable/process()
-	check_victim()
+	check_patient()
 	if(LAZYLEN(injected_reagents))
 		for(var/mob/living/carbon/C in get_turf(src))
 			var/datum/reagents/R = C.reagents
 			for(var/chemical in injected_reagents)
 				R.check_and_add(chemical,reagent_target_amount,inject_amount)
 
-/obj/machinery/optable/proc/take_victim(mob/living/carbon/C, mob/living/carbon/user as mob)
+/obj/machinery/optable/proc/take_patient(mob/living/carbon/C, mob/living/carbon/user)
 	if(C == user)
 		user.visible_message("[user] climbs on the operating table.","You climb on the operating table.")
 	else
 		visible_message("<span class='alert'>[C] has been laid on the operating table by [user].</span>")
-	C.resting = 1
+	computer.newPat = TRUE
+	computer.patStatHolder = null
+	C.resting = TRUE
 	C.update_canmove()
 	C.forceMove(loc)
 	if(user.pulling == C)
@@ -105,10 +105,10 @@
 		C.s_active.close(C)
 	for(var/obj/O in src)
 		O.loc = src.loc
-	src.add_fingerprint(user)
+	add_fingerprint(user)
 	if(ishuman(C))
 		var/mob/living/carbon/human/H = C
-		src.victim = H
+		patient = H
 		if(!no_icon_updates)
 			icon_state = H.pulse ? "table2-active" : "table2-idle"
 	else
@@ -119,17 +119,15 @@
 	set name = "Climb On Table"
 	set category = "Object"
 	set src in oview(1)
-
-	if(usr.stat || !ishuman(usr) || usr.restrained() || !check_table(usr))
+	if(usr.stat || !ishuman(usr) || usr.restrained() || !check_table())
 		return
-
-	take_victim(usr,usr)
+	take_patient(usr, usr)
 
 /obj/machinery/optable/attackby(obj/item/I, mob/living/carbon/user, params)
 	if(istype(I, /obj/item/grab))
 		var/obj/item/grab/G = I
 		if(iscarbon(G.affecting))
-			take_victim(G.affecting, user)
+			take_patient(G.affecting, user)
 			qdel(G)
 	else
 		return ..()
@@ -143,13 +141,9 @@
 		new /obj/item/stack/sheet/plasteel(loc, 5)
 		qdel(src)
 
-/obj/machinery/optable/proc/check_table(mob/living/carbon/patient as mob)
-	if(src.victim && get_turf(victim) == get_turf(src) && victim.lying)
+/obj/machinery/optable/proc/check_table()
+	if(check_patient() && patient.lying)
 		to_chat(usr, "<span class='notice'>The table is already occupied!</span>")
-		return 0
-
-	if(patient.buckled)
-		to_chat(usr, "<span class='notice'>Unbuckle first!</span>")
-		return 0
-
-	return 1
+		return FALSE
+	else
+		return TRUE
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -55,7 +55,7 @@
 		return
 	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
 		return
-	if(!ishuman(O) //Only Humanoids can go on this table
+	if(!ishuman(O)) //Only Humanoids can go on this table
 		return
 	if(isanimal(O)) || isrobot(O)) //Animals and Cyborgs do not go on the table
 		return
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -57,7 +57,7 @@
 		return
 	if(!ishuman(O)) //Only Humanoids can go on this table
 		return
-	if(isanimal(O)) || isrobot(O)) //Animals and Cyborgs do not go on the table
+	if(isanimal(O) || isrobot(O)) //Animals and Cyborgs do not go on the table
 		return
 	var/mob/living/L = O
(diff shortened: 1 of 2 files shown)


TDSSS:
You don't need to quote the code you changed, we can all look at files changed to see that. If you want to comment on the individual lines of code, you can insert that there, too.

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,63 @@
...
+		return
+	if(!ismob(O)) //Humans only
 		return
-
 	if(istype(O, /mob/living/simple_animal) || istype(O, /mob/living/silicon)) //animals and robots dont fit
 		return
-
 	var/mob/living/L = O
-	take_victim(L,usr)
+	take_patient(L, usr)
```
moxian, line 63:
```dm
var/mob/living/carbon/C = O
if(!C)  // humanoids only
  return
take_patient(C, usr)
```
The `ismob()` check does not do what the previous coder expected it to do - it would match all the weird movable things (like ghosts) (and would likely runtime later).
The *intent* here seems to only be taking in humanoids, so let's just check for that instead of matching everything and special-casing some of the things we don't want.

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,63 @@
...
 	if(istype(mover) && mover.checkpass(PASSTABLE))
-		return 1
+		return TRUE
 	else
-		return 0
+		return FALSE
 
 
-/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
-	if(usr.stat || (!ishuman(user) && !isrobot(user)) || user.restrained() || !check_table(user) || user.IsWeakened() || user.stunned)
```
moxian, line 55: i think `user.IsWeakened()` and `user.stunned` got lost? I agree they logically should be part of `.incapacitated` but they don't seem to be?..
Though I'm not sure if this matters enough.

code/game/machinery/computer/Operating.dm
```diff
@@ -7,18 +7,19 @@
...
-	var/healthAnnounce = 1 //healther announcer toggle
-	var/crit = 1 //crit beeping toggle
+	var/choice = FALSE //just for going into and out of the options menu
+	var/healthAnnounce = TRUE //healther announcer toggle
+	var/crit = TRUE //crit beeping toggle
 	var/nextTick = OP_COMPUTER_COOLDOWN
 	var/healthAlarm = 50
-	var/oxy = 1 //oxygen beeping toggle
+	var/oxy = TRUE //oxygen beeping toggle
+	var/newPat = TRUE
```
moxian, line 20: Let's not have a "this thing has changed - please update me" var.
Either call a "please update your state" proc, or let the machinery itself notice that things have changed (like it used to do with `if(patientName!=table.victim.name)`

code/game/machinery/computer/Operating.dm
```diff
@@ -7,18 +7,19 @@
...
+	var/patStat
+	var/patStatHolder
```
moxian, line 22: The code style seems to be having full words for variables. So `var/patientStat` and `var/patientStatHolder`.
In either way, these need to be documented. on what they do - they are somewhat cryptic right now.

code/game/machinery/computer/Operating.dm
```diff
@@ -173,20 +171,29 @@
...
 /obj/machinery/computer/operating/process()
-
-	if(table && table.check_victim())
+	if(table && table.check_patient())
 		if(verbose)
-			if(patientName!=table.victim.name)
-				patientName=table.victim.name
+			if((table.patient.stat != patStatHolder) || newPat)
+				if(table.patient.stat == DEAD || table.patient.status_flags & FAKEDEATH)
+					patStat = "Dead"
```
moxian, line 178: You are only using `patStat` inside this function. I strongly suggest making it a local variable (rather than a member variable on the computer itself).

code/game/machinery/computer/Operating.dm
```diff
@@ -173,20 +171,29 @@
...
 					playsound(src.loc, 'sound/machines/defib_success.ogg', 50, 0)
-				if(oxy && victim.getOxyLoss()>oxyAlarm)
+				if(oxy && table.patient.getOxyLoss()>oxyAlarm)
 					playsound(src.loc, 'sound/machines/defib_saftyoff.ogg', 50, 0)
-				if(healthAnnounce && victim.health <= healthAlarm)
-					atom_say("[round(victim.health)]")
+				if(healthAnnounce && table.patient.health <= healthAlarm)
+					atom_say("[round(table.patient.health)]")
+				if(table.patient.stat != patStatHolder)
+					atom_say("Patient is currently: [patStat]")
```
moxian, line 198:
```suggestion
					atom_say("Patient is currently [patStat]")
```

moxian (changes requested):
First pass.

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,63 @@
...
+/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, var/mob/P)
+	if(usr.stat || (!ishuman(P) && !isrobot(P)) || !check_table())
 		return
-
-	if(!ismob(O)) //humans only
+	if(!check_table()) //Is the Operating Table currently occupied?
+		return
+	if(P.restrained() || P.buckled || P.incapacitated()) //Folks who are restrained or otherwise incapacitated cannot get on the table
```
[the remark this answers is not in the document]
Kiyahitayika: From my understanding and from what was discussed in Discord, in the original code, `user` was being cast a new variable, passed through MouseDrop.

So you click on someone and drag their sprite to the table. That's the MouseDrop. The `mob/user as mob` here is that someone. It was checking if the `usr` (_the one who did the dragging_), was UNCONSCIOUS (that is, `usr.stat = 1`). If they were, it did a `return`.

Then it was checking if the mob **passed** (_the one being dragged_) named by the original proc as `user` was in a variety of states. If so, `return`. Let's make this clearer.

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,63 @@
...
+		return TRUE
 	else
-		return 0
+		return FALSE
 
 
-/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, mob/user as mob)
-	if(usr.stat || (!ishuman(user) && !isrobot(user)) || user.restrained() || !check_table(user) || user.IsWeakened() || user.stunned)
+/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, var/mob/P)
+	if(usr.stat || (!ishuman(P) && !isrobot(P)) || !check_table())
```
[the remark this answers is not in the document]
Kiyahitayika: `usr` is defined by BYOND. It is who did the MouseDrop_T().
I'm thinking of making this: `if(usr.stat != CONSCIOUS)`
We could also do:
```var/dragger = usr
if(dragger.stat != CONSCIOUS)```

File: code/game/machinery/OpTable.dm (hunk 1 above)
Comment (Kiyahitayika) [reply code/game/machinery/OpTable.dm:63]: Thanks! I was confused by this too!

File: code/game/machinery/OpTable.dm (hunk 2 above)
Comment (Kiyahitayika) [reply code/game/machinery/OpTable.dm:55]: Hmm. @SabreML said they were... Let's find out.

File: code/game/machinery/computer/Operating.dm (hunk 3 above)
Comment (Kiyahitayika) [reply code/game/machinery/computer/Operating.dm:20]: patientName is exactly what that is.

File: code/game/machinery/computer/Operating.dm (hunk 4 above)
Comment (Kiyahitayika) [reply code/game/machinery/computer/Operating.dm:22]: patStat = Patient Status
patStatHolder = Patient Status Holder

File: code/game/machinery/computer/Operating.dm (hunk 5 above)
Comment (Kiyahitayika) [reply code/game/machinery/computer/Operating.dm:178]: If it's made a local variable, the it is **created** every time `process()` is called which is on every tick.

File: code/game/machinery/OpTable.dm (hunk 7 above)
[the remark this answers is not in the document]
Comment (Kiyahitayika) [reply code/game/machinery/OpTable.dm:56]: Yeah, I'm an idiot.

File: code/game/machinery/OpTable.dm (hunk 8 above)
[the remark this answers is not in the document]
Comment (Kiyahitayika) [reply code/game/machinery/OpTable.dm:52]: Yeah, let's...make things a little clearer. I'm confused too.

File: code/game/machinery/OpTable.dm (hunk 1 above)
Comment (Kiyahitayika) [reply code/game/machinery/OpTable.dm:63]: `ishuman(O)`

File: code/game/machinery/computer/Operating.dm (hunk 6 above)
Comment (Kiyahitayika) [reply code/game/machinery/computer/Operating.dm:198]: Agreed!

File: code/game/machinery/OpTable.dm (hunk 8 above)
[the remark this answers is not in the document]
Comment (Kiyahitayika) [reply code/game/machinery/OpTable.dm:52]: I was wrong. :D Fixed per your suggestions.

File: code/game/machinery/computer/Operating.dm (hunk 4 above)
Comment (Kiyahitayika) [reply code/game/machinery/computer/Operating.dm]: Leaving this comment open - not sure if current commits address this well enough.

`patientStatus` is intended to be a string that tells the computer what to say. The string is based on `table.patient.stat`. Is the patient Awake? Asleep? Dead?

`patientStatusHolder` holds the previous instance of `patientStatus`. When `patientStatus` no longer matches `patientStatusHolder`, it's time to update and have the Operating Computer inform the doctor.

File: code/game/machinery/OpTable.dm [hunk 9]
```diff
@@ -48,19 +48,19 @@
...
-/obj/machinery/optable/MouseDrop_T(atom/movable/O as mob|obj, var/mob/P)
-	if(usr.stat || (!ishuman(P) && !isrobot(P)) || !check_table())
+/obj/machinery/optable/MouseDrop_T(atom/movable/O, mob/user)
+	if(!ishuman(user) && !isrobot(user)) //Only Humanoids and Cyborgs can put things on this table
 		return
-	if(!check_table()) //Is the Operating Table currently occupied?
+	if(!check_table()) //Is the Operating Table empty?
 		return
-	if(P.restrained() || P.buckled || P.incapacitated()) //Folks who are restrained or otherwise incapacitated cannot get on the table
+	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
```
moxian, line 56: this does not check for `user.stat` anymore, and i'm not sure it's covered by incapacitateted() (it might!)
Please double-check whether uncoscious people can drag stuff.

code/game/machinery/OpTable.dm
```diff
@@ -48,19 +48,19 @@
...
 		return
-	if(!ismob(O)) //Humans only
+	if(!ishuman(O)) //Only Humanoids can go on this table
 		return
-	if(istype(O, /mob/living/simple_animal) || istype(O, /mob/living/silicon)) //animals and robots dont fit
+	if(isanimal(O) || isrobot(O)) //Animals and Cyborgs do not go on the table
 		return
```
moxian, line 61: You shouldn't need this anymore - it's covered by `ishuman(O)` (or `iscarbon`) above

code/game/machinery/OpTable.dm
```diff
@@ -48,19 +48,19 @@
...
+	if(!ishuman(user) && !isrobot(user)) //Only Humanoids and Cyborgs can put things on this table
 		return
-	if(!check_table()) //Is the Operating Table currently occupied?
+	if(!check_table()) //Is the Operating Table empty?
 		return
-	if(P.restrained() || P.buckled || P.incapacitated()) //Folks who are restrained or otherwise incapacitated cannot get on the table
+	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
 		return
-	if(!ismob(O)) //Humans only
+	if(!ishuman(O)) //Only Humanoids can go on this table
```
moxian, line 58: `if(!iscarbon(O))`
The table on live can accept xenos (which are carbon, but not human). This removal shouldn't matter in practice, but i think it's good to leave them in.

code/game/machinery/OpTable.dm
```diff
@@ -48,19 +48,19 @@
...
 		return
-	if(istype(O, /mob/living/simple_animal) || istype(O, /mob/living/silicon)) //animals and robots dont fit
+	if(isanimal(O) || isrobot(O)) //Animals and Cyborgs do not go on the table
 		return
 	var/mob/living/L = O
```
moxian, line 62:
```suggestion
	var/mob/living/carbon/C = O
```
Both because you've checked for this above, and because `take_patient` accepts a `mob/living/carbon`, and not just  any `mob/living`.

But also
```dm
if(!iscarbon(O))
  return
var/mob/living/carbon/C = O
```
is the same as
```dm
var/mob/living/carbon/C = O
if(!C)
  return
```

I'm leaning a bit towards the latter, because it's harder to screw up.

Consider the following snippet (roughly what used to be here):
```dm
if(!ismob(O))
  return
var/mob/living/L = O
```
Now `L` can suddenly be null, and you might not notice (because you've checked that it's a mob already! Why would you check it again, right?)! (The previous code tried to work around this by snowflaking simple animals and silicons, but those "fixes" only hide the problem, and don't actually resolve it).

code/game/machinery/computer/Operating.dm
```diff
@@ -7,18 +7,18 @@
...
-	var/healthAnnounce = 1 //healther announcer toggle
-	var/crit = 1 //crit beeping toggle
+	var/choice = FALSE //just for going into and out of the options menu
+	var/healthAnnounce = TRUE //healther announcer toggle
+	var/crit = TRUE //crit beeping toggle
 	var/nextTick = OP_COMPUTER_COOLDOWN
 	var/healthAlarm = 50
-	var/oxy = 1 //oxygen beeping toggle
+	var/oxy = TRUE //oxygen beeping toggle
+	var/currentPatient //Who is on the Operating Table connected to the respective Operating Computer?
```
moxian, line 20: Let's specify the variable type as well: 
```suggestion
	var/mob/living/carbon/currentPatient //Who is on the Operating Table connected to the respective Operating Computer?
```

moxian (changes requested):
Minor things now.

code/game/machinery/OpTable.dm
Kiyahitayika: Hmm. Other MouseDrop_T() procs don't do the `stat` check, so I removed it. I'll double check!

code/game/machinery/OpTable.dm
Kiyahitayika: Good catch! Thank you!

code/game/machinery/OpTable.dm
Kiyahitayika: Wait, it _can_? Whaaaat

code/game/machinery/OpTable.dm
Kiyahitayika: Confirmed, you can **not** mouse-drag while asleep. Removing this check appears to be okay!
I don't think it matters if it is covered by `incapacitated()` as `user.stat` check does not appear in similar `MouseDrop_T()` procs (looking at body scanner for instance) and you still cannot drag while asleep into/onto those objects.
I wonder if BYOND had an update that changed this since OpTable was originally written. Either way, it all seems good!

code/game/machinery/OpTable.dm
Kiyahitayika: Huh. So it is! I was squinting at this last night, thinking that it was redundant. I guess my brain was too tired to make the connection.

code/game/machinery/OpTable.dm
Kiyahitayika:
```
	if(!ishuman(user) && !isrobot(user)) //Only Humanoids and Cyborgs can put things on this table
		return
	if(!check_table()) //Is the Operating Table empty?
		return
	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
		return
	var/mob/living/carbon/C = O
	if(!C) //Only Carbon mobs can go on this table, if C is null, get out of here
		return
	take_patient(C, user)
```

code/game/machinery/OpTable.dm
Kiyahitayika: Changed this to `iscarbon`, then changed to your other suggestion!

code/game/machinery/computer/Operating.dm
Kiyahitayika: Not sure why I didn't do this in the first place! Thank you!

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,58 @@
 		return TRUE
 
 /obj/machinery/optable/CanPass(atom/movable/mover, turf/target, height=0)
-	if(height==0) return 1
-
+	if(height==0)
```
farie82, line 43: Spacing here

code/game/machinery/OpTable.dm
```diff
@@ -29,10 +27,9 @@
 /obj/machinery/optable/Destroy()
 	if(computer)
 		computer.table = null
-		computer.victim = null
 		computer = null
-	if(victim)
-		victim = null
+	if(patient)
```
farie82, line 31: if not needed

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,58 @@
...
+/obj/machinery/optable/MouseDrop_T(atom/movable/O, mob/user)
+	if(!ishuman(user) && !isrobot(user)) //Only Humanoids and Cyborgs can put things on this table
 		return
-
-	if(!ismob(O)) //humans only
+	if(!check_table()) //If the Operating Table is occupied, you cannot put someone else on it
 		return
-
-	if(istype(O, /mob/living/simple_animal) || istype(O, /mob/living/silicon)) //animals and robots dont fit
+	if(user.restrained() || user.buckled || user.IsWeakened() || user.stunned || user.incapacitated()) //Is the person trying to use the table incapacitated or restrained?
```
farie82, line 56: `user.incapacitated()` will already check most of these. Only the buckled check is not done there.

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,58 @@
...
 		return
+	var/mob/living/carbon/C = O
+	if(!C) //Only Carbon mobs can go on this table, if C is null, get out of here
+		return
+	take_patient(C, user)
```
farie82, line 61:
```suggestion
	take_patient(O, user)
```
Don't have to cast it here. The check doesn't do much either as you can do this:
```
var/datum/t1/T = new()
var/datum/t2/T2 = T
if(T2)
 // will be true since T2 is assigned (just not the same type)
```

code/game/machinery/OpTable.dm
```diff
@@ -43,60 +40,58 @@
...
 
-	var/mob/living/L = O
-	take_victim(L,usr)
-	return
-
-/obj/machinery/optable/proc/check_victim()
-	if(locate(/mob/living/carbon/human, src.loc))
-		var/mob/living/carbon/human/M = locate(/mob/living/carbon/human, src.loc)
+/obj/machinery/optable/proc/check_patient()
+	if(locate(/mob/living/carbon/human, loc))
```
farie82, line 64: Double locate here while you can just assign the var prior and then check if said var is not null

farie82 (changes requested):
partial review. Other half will come soon :tm:

code/game/machinery/OpTable.dm
Kiyahitayika: Yeah, that's weird! Thanks!

code/game/machinery/OpTable.dm
Kiyahitayika: There was confusion between a few of us on this. Thanks for clearing this up!

code/game/machinery/OpTable.dm
Kiyahitayika: Huh. True.

code/game/machinery/OpTable.dm
Kiyahitayika: This doesn't exclude `isanimal` and `isrobot`

Kiyahitayika:
```
	if(isanimal(O) && isrobot(O)) //Only Carbon mobs can go on this table, if C is null, get out of here
		return
	take_patient(O, user)
```

Kiyahitayika:
I just noticed a game-crashing runtime that can occur. Looking into it.

Kyep (approved):

code/game/machinery/OpTable.dm
Kiyahitayika: Changed to:
```
if(!iscarbon(O) || !ismob(O))
```
ismob() is needed to ensure that we're not placing an object on the table. If we do, it creates a runtime error in `take_patient()`, so we need the return. The ismob() check was removed based on a previous suggestion - the comments next to the original line of code led us to believe it did something else. Not sure where my head has been lately in terms of understanding things, apologies.

iscarbon() check replacing the casting since the casting wasn't actually helping check as you said.

code/game/machinery/OpTable.dm
Kiyahitayika: Hmm. I may have messed this up.

PR #14732, opened 2020-10-26
[TGUI] Vault Safe

nicetoolbox:
## What Does This PR Do
Updates the UI code for the vault safe to tgui. **_There is a weird issue where clicking the buttons too quickly causes them to not respond any more. Using the keyboard (tab and spacebar) to select buttons doesn't have this issue._**

## Why It's Good For The Game
Removed old nano ui code

## Images of changes
locked
open
unlocked

## Changelog
:cl:
tweak: Update vault safe to TGUI
/:cl:


moxian:
Where did you get the dial picture from? Does the license allow using it here?

tgui/packages/tgui/styles/main.scss
```diff
@@ -64,3 +64,4 @@
   background-position: center;
   background-repeat: no-repeat;
 }
+@include meta.load-css('./interfaces/Safe.scss');
```
AffectedArc07, line 67: Stick this in order with the other interface loads

tgui/packages/tgui/interfaces/Safe.js
```diff
@@ -0,0 +1,119 @@
...
+            <Button content="Left 1" onClick={() => turn('left', 1)} icon="arrow-left" />
+          </Flex>
+
+          <Flex flexGrow={1} alignItems="center" flexDirection="column">
+            {dial}
+          </Flex>
+
+          <Flex>
+
+            <Button disabled={!locked} content="Right 1" onClick={() => turn('right', 1)} icon="arrow-left" />
```
moxian, line 76: This makes the "Right" icon with left-pointing arrow, which is very odd
firefox-20201026-093444

I assume this is done because `-> Right` looks weird.

If you want to have icon *following* the text (so `Right ->`) you need to explicitly declare button contents, i.e.:
```jsx
<Button disabled={!locked} onClick={() => turn('right', 1)}>Right 1 <Icon name="arrow-right"/></Button>
```

code/game/objects/structures/safe.dm
```diff
@@ -211,96 +212,80 @@ GLOBAL_LIST_EMPTY(safes)
...
-		if(index > 0 && index <= contents.len)
-			var/obj/item/P = contents[index]
-			if(open)
-				if(P && in_range(src, user))
-					user.put_in_hands(P)
-					space -= P.w_class
-		. = TRUE
+				to_chat(user, "<span class='warning'>You can't open [src], as its lock is engaged!</span>")
+
+			. = TRUE
```
AffectedArc07, line 269: Define this once instead of after each switch row

code/game/objects/structures/safe.dm
```diff
@@ -211,96 +212,80 @@ GLOBAL_LIST_EMPTY(safes)
...
 
-		if(broken)
-			to_chat(user, "<span class='warning'>The dial will not turn, as the mechanism is destroyed.</span>")
-			return
 
-		var/ticks = text2num(href_list["turnleft"])
-		for(var/i = 1 to ticks)
-			dial = WRAP(dial + 1, 0, 100)
+/obj/structure/safe/tgui_act(action, list/params)
+	var/mob/user = usr
```
AffectedArc07, line 249: Any reason to define this instead of just using `usr` in each statement down

code/game/objects/structures/safe.dm
```diff
@@ -211,96 +212,80 @@ GLOBAL_LIST_EMPTY(safes)
...
+	if(!user.IsAdvancedToolUser() && !isobserver(user))
```
AffectedArc07, line 250: I am questioning the need for the ` && !isobserver()` check, especially since the TGUI safeties will check the state. You are also missing those. Please add
```dm
if(..())
	return
```
To the very top of `tgui_act()`

code/game/objects/structures/safe.dm
```diff
@@ -176,17 +177,17 @@ GLOBAL_LIST_EMPTY(safes)
 			if("Cancel")
 				return
 	else
-		ui_interact(user)
+		tgui_interact(user)
 
-/obj/structure/safe/ui_interact(mob/user, ui_key = "main", datum/nanoui/ui = null, force_open = 1)
-	ui = SSnanoui.try_update_ui(user, src, ui_key, ui, force_open)
+/obj/structure/safe/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = FALSE, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_default_state)
```
AffectedArc07, line 182:
```suggestion
/obj/structure/safe/tgui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = FALSE, datum/tgui/master_ui = null, datum/tgui_state/state = GLOB.tgui_physical_state)
```
Safes are physical, not electronic. Use `physical_state` to avoid AIs and borgs using them

tgui/packages/tgui/interfaces/Safe.js
```diff
@@ -0,0 +1,119 @@
...
+          </Box>
+          <Box bold>To lock fully, turn the dial to the left after closing the safe.</Box>
+        </Box>
+      </Flex>
+
+    </Section>
+  );
+};
+
+export const Safe = (properties, context) => {
```
AffectedArc07, line 107: Put this as the first constant just below the imports for the sake of consistency

tgui/packages/tgui/interfaces/Safe.js
```diff
@@ -0,0 +1,119 @@
...
+    contents,
+  } = data;
+
+  const close = () => {
+    act('open');
+  };
+
+  return (
+    <Section>
+      <Button content="Close" onClick={close} />
```
AffectedArc07, line 18: But the button inside the `<Section buttons={}` attribute, and also add a section title

tgui/packages/tgui/interfaces/Safe.js
```diff
@@ -0,0 +1,119 @@
...
+        <Flex alignItems="center" flexDirection="column">
+
+          <Box className="Safe__Dial" style={{
+            transform: `rotate(${rotation})`,
+          }} />
+        </Flex>
+
+
+        <Box>
+          <Box bold>How to open your Scarborough Arms tumbler safe.</Box>
```
AffectedArc07, line 92: Stick an `mt={2}` on this to push it down a little

tgui/packages/tgui/interfaces/Safe.js
```diff
@@ -0,0 +1,119 @@
...
+    <Section>
+
+      <Flex alignItems="stretch" flexDirection="column">
+
+        <Flex alignItems="center" flexDirection="column">
+          <Button content="Open" onClick={open} />
+
+        </Flex>
+
+        <Flex justifyContent="space-between" alignItems="center">
```
AffectedArc07, line 62: Can you force set the width of the left and right columns to avoid this
image
Yes the buttons on the left are smaller than the ones on the right, but the box being off-centre looks real off

AffectedArc07 (changes requested):

code/game/objects/structures/safe.dm
```diff
@@ -211,96 +212,80 @@ GLOBAL_LIST_EMPTY(safes)
...
 		to_chat(user, "<span class='italics'>You hear a [pick(sounds)] from [src].</span>")
 
-/obj/structure/safe/Topic(href, href_list)
-	if(..())
-		return TRUE
-
-	var/mob/user = usr
-	if(!user.IsAdvancedToolUser() && !isobserver(user))
-		to_chat(user, "<span class='warning'>You're not able to operate the safe.</span>")
+/obj/structure/safe/proc/start_turning_knob(mob/user, canhear, ticks, left)
```
moxian, line 215: Consider accepting a signed `ticks` value (with positive values meaning turn right, and negative ones - turn left, or vice versa).
Subjectively, it feels a bit cleaner that way, but feel free to leave as is.

code/game/objects/structures/safe.dm
```diff
@@ -211,96 +212,80 @@ GLOBAL_LIST_EMPTY(safes)
...
-			check_unlocked(user, canhear)
-			SSnanoui.update_uis(src)
+		if(!invalid_turn && dial == tumblers[current_tumbler_index])
+			notify_user(user, canhear, list("tonk", "krunk", "plunk"), ticks, i)
+			current_tumbler_index++
+		else
+			notify_user(user, canhear, list("click", "chink", "clink"), ticks, i)
 
-		. = TRUE
+	addtimer(CALLBACK(src, .proc/finish_turning_knob, user, canhear), 1)
```
moxian, line 240: If I'm reading this right, this only updates the UI once, after all the turning is done, instead of on each tick? And there's little reason to have a callback here, and not just call `finish_turning_knob` directly?

The previous code had it turn gradually(-ish.. as much as sleep(1) can be called gradual). I'm not sure if having `sleep(1)` in here messes with `tgui_act` but if so, you probably need the whole `start_turning_knob` be in a callback, and not just the `finish_turning_knob` part of it.

moxian (changes requested):
I can confirm the "buttons stop working" , but that's very weird and seem to be dependent on the *state* of the safe. Can it be that the tumlber image overlaps with the buttons? See the gif, the `Left 1` stops working only when the safe position is at ~41-42
20201026-101421-kJYnZwccdh

warior4356:
@moxian I did the original refactor for the safe (dial, nanoui, drills, etc) over a year ago, and I made that dial graphic myself.

AffectedArc07:
Requested this be seen to within the next week, as its now the last nanoUI which needs shooting.

AffectedArc07:
@nicetoolbox please respond to this PR within 48 hours or it will be closed and taken up by someone else. It is the last TGUI left in the chain and needs dealing with relatively soon, not to mention this PR being open for over 3 weeks with 0 response.

AffectedArc07:
This PR will now be taken up by someone else due to no activity for 25 days

PR #14734, opened 2020-10-26
[TGUI] Request Consoles

AffectedArc07:
## What Does This PR Do
Converts request consoles from NanoUI to TGUI

## Why It's Good For The Game
***NANO YOU DONT HAVE ANY INTERFACES LEFT WHICH ARENT BEING WORKED ON. YOUR END IS NEAR...***

## Images of changes
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image

## Changelog
:cl: AffectedArc07
add: Request consoles now use TGUI
/:cl:


code/game/machinery/requests_console.dm
```diff
@@ -164,142 +165,128 @@ GLOBAL_LIST_EMPTY(allRequestConsoles)
...
+			announcement.Announce(message, msg_sanitized = TRUE)
+			reset_message(TRUE)
+
+		if("department")
+			if(!message)
+				return
+			var/log_msg = message
+			var/pass = FALSE
+			screen = RCS_SENTFAIL
+			for(var/obj/machinery/message_server/MS in GLOB.machines)
```
dearmochi, line 214: Prefer using `GLOB.message_servers` and cast the value inside the loop

code/game/machinery/requests_console.dm
```diff
@@ -164,142 +165,128 @@ GLOBAL_LIST_EMPTY(allRequestConsoles)
...
+				if(recipient in MED_ROLES)
+					radiochannel = "Medical"
+				if(recipient in COM_ROLES)
+					radiochannel = "Command"
+				if(recipient in SCI_ROLES)
+					radiochannel = "Science"
+				if(recipient == "AI")
+					radiochannel = "AI Private"
+				if(recipient == "Cargo Bay")
+					radiochannel = "Supply"
```
dearmochi, line 236: Yuck! Maybe make that into a loop?

code/game/machinery/requests_console.dm
```diff
@@ -164,142 +165,128 @@ GLOBAL_LIST_EMPTY(allRequestConsoles)
...
+				return
+			if(tempScreen == RCS_VIEWMSGS)
+				for(var/obj/machinery/requests_console/Console in GLOB.allRequestConsoles)
+					if(Console.department == department)
+						Console.newmessagepriority = RQ_NONEW_MESSAGES
+						Console.icon_state = "req_comp0"
+						Console.set_light(1)
+			if(tempScreen == RCS_MAINMENU)
+				reset_message()
+			screen = tempScreen
```
dearmochi, line 255: Would add a default case for invalid screen indexes.

code/game/machinery/requests_console.dm
```diff
@@ -164,142 +165,128 @@ GLOBAL_LIST_EMPTY(allRequestConsoles)
...
+
+		if("shipSelect")
+			ship_tag_name = params["shipSelect"]
+			ship_tag_index = GLOB.TAGGERLOCATIONS.Find(ship_tag_name)
+
+		//Handle Shipping Label Printing
+		if("printLabel")
+			var/error_message = ""
```
dearmochi, line 263:
```suggestion
			var/error_message
```

That variable doesn't need initializing

code/game/machinery/requests_console.dm
```diff
@@ -164,142 +165,128 @@ GLOBAL_LIST_EMPTY(allRequestConsoles)
...
+						Console.newmessagepriority = RQ_NONEW_MESSAGES
+						Console.icon_state = "req_comp0"
+						Console.set_light(1)
+			if(tempScreen == RCS_MAINMENU)
+				reset_message()
+			screen = tempScreen
+
+		if("shipSelect")
+			ship_tag_name = params["shipSelect"]
+			ship_tag_index = GLOB.TAGGERLOCATIONS.Find(ship_tag_name)
```
dearmochi, line 259: Needs a check to whether the shipping destination is valid.

code/game/machinery/requests_console.dm
```diff
@@ -308,59 +295,63 @@ GLOBAL_LIST_EMPTY(allRequestConsoles)
...
 	capitalize(title)
-	if(src.newmessagepriority < priority)
-		src.newmessagepriority = priority
+	if(newmessagepriority < priority)
+		newmessagepriority = priority
 		update_icon()
-	if(!src.silent)
-		playsound(src.loc, 'sound/machines/twobeep.ogg', 50, 1)
+	if(!silent)
+		playsound(loc, 'sound/machines/twobeep.ogg', 50, 1)
```
dearmochi, line 339:
```suggestion
		playsound(loc, 'sound/machines/twobeep.ogg', 50, TRUE)
```

code/modules/research/message_server.dm
```diff
@@ -86,15 +86,15 @@ GLOBAL_LIST_EMPTY(message_servers)
...
-		authmsg += "[id_auth]<br>"
+		authmsg += " - [id_auth]"
 	if(stamp)
-		authmsg += "[stamp]<br>"
+		authmsg += " - [stamp]"
 	for(var/obj/machinery/requests_console/Console in GLOB.allRequestConsoles)
 		if(ckey(Console.department) == ckey(recipient))
 			if(Console.inoperable())
-				Console.message_log += "<B>Message lost due to console failure.</B><BR>Please contact [station_name()] system adminsitrator or AI for technical assistance.<BR>"
+				Console.message_log += "Message lost due to console failure. Please contact [station_name()] system adminsitrator or AI for technical assistance."
```
dearmochi, line 97:
```suggestion
				Console.message_log += "Message lost due to console failure. Please contact [station_name()]'s system administrator or AI for technical assistance."
```

dearmochi (changes requested):
A few things

code/game/machinery/requests_console.dm
AffectedArc07: In what way

code/game/machinery/requests_console.dm
AffectedArc07: In the UI?

code/game/machinery/requests_console.dm
AffectedArc07:
```dm
			if(!ship_tag_index)
				error_message = "Please select a destination."
```

code/game/machinery/requests_console.dm
AffectedArc07: No
image

code/game/machinery/requests_console.dm
dearmochi: Something along the lines of
```dm
var/channels = list("Engineering" = ENGI_ROLES, "Security" = SEC_ROLES)
for(var/channel in my_list) // ETC..
	if(recipient in channels[channel])
		radiochannel = channel
		break
```

code/game/machinery/requests_console.dm
dearmochi: I'd add one in the DM side to prevent the client from reaching that part to begin with

code/game/machinery/requests_console.dm
AffectedArc07: Doesnt work because theres mixes of `==` and `in LIST`

dearmochi: And for these cases you can just make them into a list i.e `list("AI")` and `list("Cargo Bay")`

AffectedArc07: Im also not sure if instantiating a list, especially with more lists for singular jobs, just for iteration is any less hacky than using lots of if statements, and it is likely slower since it has to instantiate the list and the loop.

dearmochi: Then perhaps you can forgo that and instead make them into `else if`s

PR #14741, opened 2020-10-26
[DNM] Prevents ghosts from being visable with spectral sword. (working on orbiting ghost orbs)

Qwertytoforty:
## What Does This PR Do

Prevents spectral sword from making ghosts visible. You can still call ghosts to orbit you, and it still becomes stronger with more ghosts.

## Why It's Good For The Game

Ghosts being visible from spectral sword leads to many problems. They point to antags. They point to security officers for the antag in maint. They orbit other things then the sword user, being visible for a moment before becoming invisible again . And they can reveal way to many things to the crew that they should not know. Be it a swarmer or terror corpse, or the body of the dead captain, while the "captain" is sitting on the bridge. As nice as seeing ghosts with this sword is, the ghosts themselves are not responsible enough to be visible, and thus, should be removed. I would _love_ a visual orbit effect of ghostly orbs orbiting you, thus preventing pointing, seeing dead bodies, or leading to things, but as I am not able to do that, this fix will do for now.

## Changelog
:cl:
tweak: Spectral sword no longer makes ghosts visible.
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -40,6 +40,8 @@
 	attack_verb = list("attacked", "slashed", "stabbed", "sliced", "torn", "ripped", "diced", "rended")
 	var/summon_cooldown = 0
 	var/list/mob/dead/observer/spirits
+	var/globes_list = list()
+	var/spirit_count = 0
 
 /obj/item/melee/ghost_sword/New()
 	..()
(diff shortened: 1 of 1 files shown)


Kiyahitayika:
> I would _love_ a visual orbit effect of ghostly orbs orbiting you

Something someone might want to look into if they're interested!
I agree with this PR. Honestly, I had no idea that my pointing as a ghost while orbiting the sword could be seen. That's...meta.

Cocacolagua:
You already know about this but I’m going to say it here for the record. Removing the ghost vision eliminates meta yes, but also makes it undoubtedly the strongest melee weapon in the game, as it won’t have ghosts blocking it now, each hit will deal 80~ damage and stuns, a better chainsaw.

Leanfrog:
I don't believe removing the visual prevents the ghost block effect from happening, which is concerning as there's no longer any visual indication of whether or not they're a highly armored killing machine

TheMaveran:
There needs to be an indication of the strength the blade is at, metamancy is bad sure but yeah the blade is 100% OP without ghosts showing.

Spacemanspark:
It does need an indication of sorts, agreed.

But I think this is def something that should be done either way.

Fox-McCloud:
these needs some kind of neutral indicator.

Maybe ghosts lose their sprite and ability to point, or something similar---but as it stands, there absolutely needs to be an indicator that "this guy is a wimp" versus "oh no this guy could potentially kill everyone".

trololiver112:
Isn't the problem mostly in the ghosts ability to point? The ghosts themselves aren't that bad right?

Spacemanspark:
Please do not remove ghost character sprites from the game, regardless of what you do.

Kyep (approved):
Frankly I would rather kill off the spectral sword entirely, but this is at least a step towards reducing the metagame issues it creates.

Qwertytoforty:
I've been busy working on other things, and can not for the life of me get the ghost orbs to work around the sword, so closing this pr, anyone can feel free to finish this idea / PR.

PR #14742, opened 2020-10-26
Update RustG + Minor Travis Updates

AffectedArc07:
## What Does This PR Do
This PR is the first of a chain I plan to make for several features depending on RUSTG (Discord webhooks, Async SQL, in-code revision info, etc). This PR is updating the libraries themselves as well as the DM interface side (It already has a sizable diff as seen below). It also cleans up some log ends. 

Images of logging still functioning before the admins threaten to pitchfork me. We also still follow 8601 standards, because anything else is objectively terrible.

*windows*
image

*also on linux*
image

## Why It's Good For The Game
Up to date libraries are good, this also unlocks a lot of optimisations down the road for us.

## Changelog
:cl: AffectedArc07
add: Updated a backend library 
/:cl:

### DIFF
--- a/_build_dependencies.sh
+++ b/_build_dependencies.sh
@@ -9,3 +9,5 @@ export PHP_VERSION=5.6
 export BYOND_MAJOR=513
 # Byond Minor
 export BYOND_MINOR=1526
+# For the RUSTG library
+export RUSTG_VERSION=2.0
--- a/_build_dependencies.sh
+++ b/_build_dependencies.sh
@@ -4,7 +4,7 @@ export SPACEMANDMM_TAG=suite-1.6
 # For NanoUI + TGUI
 export NODE_VERSION=12
 # For the scripts in tools
-export PHP_VERSION=5.6
+export PHP_VERSION=7.2
 # Byond Major
 export BYOND_MAJOR=513
 # Byond Minor
--- a/code/__DEFINES/rust_g.dm
+++ b/code/__DEFINES/rust_g.dm
@@ -1,0 +1,46 @@
+// rust_g.dm - DM API for rust_g extension library
+//
+// To configure, create a `rust_g.config.dm` and set what you care about from
+// the following options:
+//
+// #define RUST_G "path/to/rust_g"
+// Override the .dll/.so detection logic with a fixed path or with detection
+// logic of your own.
+//
+// #define RUSTG_OVERRIDE_BUILTINS
+// Enable replacement rust-g functions for certain builtins. Off by default.
+
+#define RUST_G (world.system_type == UNIX ? "./librust_g.so" : "./rust_g.dll")
+
+// Defines for internal job subsystem //
+#define RUSTG_JOB_NO_RESULTS_YET "NO RESULTS YET"
+#define RUSTG_JOB_NO_SUCH_JOB "NO SUCH JOB"
+#define RUSTG_JOB_ERROR "JOB PANICKED"
+
+#define rustg_dmi_strip_metadata(fname) call(RUST_G, "dmi_strip_metadata")(fname)
+#define rustg_dmi_create_png(path, width, height, data) call(RUST_G, "dmi_create_png")(path, width, height, data)
+
+#define rustg_noise_get_at_coordinates(seed, x, y) call(RUST_G, "noise_get_at_coordinates")(seed, x, y)
+
+#define rustg_git_revparse(rev) call(RUST_G, "rg_git_revparse")(rev)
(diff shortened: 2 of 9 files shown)


AffectedArc07:
Travis may fail depending on what versions of libraries its own OS has (Linux dependencies are pain). If it does I will recompile inside a Travis-spec VM to avoid failures.

AffectedArc07:
image
As expected

AffectedArc07:
The system has now been updated to make RUSTG retrieval part of the travis workflow, and the library is gitignored on disk. Why? Because linux dependency management sucks and varies for pretty much every OS in existence.

AffectedArc07:
This PR tweaks the Travis configuration slightly, to run us on `bionic` (18) instead of `xenial` (16). Some dependencies were updated, as well as old stuff being removed.

_build_dependencies.sh
```diff
@@ -4,8 +4,10 @@ export SPACEMANDMM_TAG=suite-1.6
 # For NanoUI + TGUI
 export NODE_VERSION=12
 # For the scripts in tools
-export PHP_VERSION=5.6
+export PHP_VERSION=7.2
```
dearmochi, line 7: Pretty sure PHP 7 has a lot of changes that make migrating from PHP 5 harder. Is it necessary?

AffectedArc07: Yes otherwise stuff fails. This version is only used for CI, and the only script that validates is the changelog processor PHP script, which runs just fine on 7.2 (I used to run it in prod on that version)

dearmochi: Alright

dearmochi (approved):
LGTM, even if I'm not familiar with the infrastructure

code/__DEFINES/rust_g.dm
```diff
@@ -0,0 +1,40 @@
...
+#define RUSTG_JOB_NO_SUCH_JOB "NO SUCH JOB"
+#define RUSTG_JOB_ERROR "JOB PANICKED"
+
+// DMI related operations //
+#define rustg_dmi_strip_metadata(fname) call(RUST_G, "dmi_strip_metadata")(fname)
+#define rustg_dmi_create_png(path, width, height, data) call(RUST_G, "dmi_create_png")(path, width, height, data)
+
+#define rustg_noise_get_at_coordinates(seed, x, y) call(RUST_G, "noise_get_at_coordinates")(seed, x, y)
+
+// Git related operations // - AA TODO
```
farie82, line 15: Why the TODOs here? They are not yet used but are they usable? Is RUST not yet able to provide them? Or is it just for yourself? If it's the latter than I'd just remove it to reduce confusion. Or explain what there is todo

code/__HELPERS/_logging.dm
```diff
@@ -32,119 +29,119 @@ GLOBAL_VAR_INIT(log_end, (world.system_type == UNIX ? ascii2text(13) : ""))
...
+	rustg_log_write(GLOB.world_runtime_log, text)
 
 /proc/log_config(text)
-	WRITE_LOG(GLOB.config_error_log, text)
+	rustg_log_write(GLOB.config_error_log, text)
 	SEND_TEXT(world.log, text)
 
 /proc/log_href(text)
-	WRITE_LOG(GLOB.world_href_log, "HREF: [html_decode(text)]")
+	rustg_log_write(GLOB.world_href_log, "HREF: [html_decode(text)][GLOB.log_end]")
```
farie82, line 135: Some minor undocumented changes here with the log_end being added.

tools/travis/install_build_deps.sh
```diff
@@ -8,7 +8,4 @@ nvm install $NODE_VERSION
 nvm use $NODE_VERSION
 npm install --global yarn
 
-pip install --user PyYaml -q
```
farie82, line 11: I recon these ain't used anymore?

code/__HELPERS/_logging.dm
AffectedArc07: image

tools/travis/install_build_deps.sh
AffectedArc07: Aint been used since `SSchangelog`

code/__DEFINES/rust_g.dm
AffectedArc07: TODOs for what I am working on, and also to dissuade other people using them yet due to their finnicky nature (These are things in my TODO list which need to be very specifically done)

farie82: Then include that info there please. A random TODO in a public repo is a bit vague

PR #14747, opened 2020-10-27
Properly enforced LF in DMAPI Updater

AffectedArc07:
## What Does This PR Do
This PR makes LF properly enforced in the DMAPI updater action

## Why It's Good For The Repository
Auto updates shouldnt be failing CI instantly

## Changelog
N/A


AffectedArc07:
To aid reviewing:
image
Disable whitespace changes and it makes the other files minimise since they had no actual diff

.github/workflows/update_tgs_dmapi.yml
```diff
@@ -19,12 +19,20 @@ jobs:
         git checkout tgs-dmapi-update
         git reset --hard master
 
+    - name: Install Deps
+      run: sudo apt install dos2unix
+
```
dearmochi, line 24: Undocumented change?

AffectedArc07: This is what enforces the LF since it force converts it all

dearmochi: Fair enough

dearmochi (approved):
LGTM, even if I'm not familiar with the infrastructure

farie82 (approved):
Should be fine if it's tested and confirmed to work

PR #14759, opened 2020-10-28
CQC can no longer be discounted

Cocacolagua:
## What Does This PR Do
You can no longer get CQC discounted in the uplink.

## Why It's Good For The Game
This item is already strong as is, some people would even argue it's "overpowered", the fact that you can have it discounted for 7TC also doesn't help, since it lets you combo it with other stuff on the uplink far easier.

## Changelog
:cl:
tweak: CQC can't be randomly discounted in the uplink anymore.
/:cl:

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -860,6 +860,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	reference = "CQC"
 	item = /obj/item/CQC_manual
 	cost = 13
+	cant_discount = TRUE
 
 /datum/uplink_item/stealthy_weapons/cameraflash
 	name = "Camera Flash"


Carthusia:
I mean, id say thats part of the luck of the draw, getting insanely powerful items for 50% off, if you remove CQC from that 50% discount table, one may as well remove it from the surplus crate to, considering you could get an even more insane mix of items, and if your lucky enough, basically become an adrenald, chainsaw having, sleeping carp, CQCing god.

Overall, id say itd be better off leaving it in the discount possibility pool

Cocacolagua:
It's harder to get insane combos with the surplus given that you can also get job-specific items, which most of them are useless uplink-bloat items.

Frankly, my own idea on the discount system is that it should only discount items you wouldn't normally buy otherwise as to encourage traitor variety. 
To give an example. items like CQC, adrenalin and whatever else shouldn't be discounted, but items like .357s revolver, mindslave implants and chameleon projectors should.

Rurik123:
Isn't the point of the discount to have the low chance of having insane combos? Such as .357 with adrenals or adrenals with sleeping carp? If CQC is that op I'll defer to you since I haven't had a chance to mess with it much myself, but if we're doing it on the basis of powerful items not being discounted, well, that defeats the purpose of discount I would think.

There could be an argument made that discount is for traitor variety, not normally impossible combos, but nobody goes to the uplink discount thinking "oh boy, what variety will I get today?" Half of the fun is hoping something good is discounted.

Cocacolagua:
actually sure, i changed my mind.

Cocacolagua:
Opening back this PR, will discuss this topic here instead of #15074, given that PR is more centered around lowering / removing items and prices, and limited hijack-tier stuff.

Reason of re-opening this PR is:
- codewords are highlighted now, being a better alternative for getting your CQC combo you want, forcing you to team-up with someone else.
- I do not really like "luck of the draw", specially in this case when it means items in the uplink must be balanced around having their price cut in half by discounts, having to balance them on being both 13TC and 7TC.

Cocacolagua:
oh right i deleted the repository, guess i gotta make a new PR lmao

PR #14762, opened 2020-10-28
SShttp + SSdiscord | ASYNCHRONOUS STUFF IN BYOND!

AffectedArc07:
## What Does This PR Do
This PR uses parts of the rust extension to make asynchronous HTTP work inside BYOND. Current uses for this are a rich discord webhook interface, as well as converting all the old IRC stuff to Discord Webhooks.

Adminwho message:
image

Simple example
image

Complex example
image

<details><summary>Old problem (No longer relevant)</summary>
<p>

Theres just one little problem, which is keeping this from being mergeable

Currently, if you restart the server without force restarting DD, it doesnt let go of the thread spawned by the rust DLL, which means it can be unloaded at a random moment without its memory references cleared. *Which means it can hard crash dream daemon and instantly kill everything with zero runtimes or anything*. I need to work this issue out before I can even think about getting this merged, but for now, the PR is a working proof of concept.
image
image

</p>
</details>

## Why It's Good For The Game
IRC code hasnt been used in years, and it relied on direct calls to the system shell, which is a direct no go, **ESPECIALLY** with SS13. This adds a HTTP client into the code which not only supports much more than `GET`, but can run asynchronously with callbacks instead of halting a proc thread, or worse, the entire DD process 

## Changelog
:cl: AffectedArc07
add: Added support for discord webhooks
add: Converted all the old IRC code to discord webhook code
add: Adminhelping with no admins online will have your ahelp sent to discord, so still ahelp if no one is online
del: Removed old, slow, insecure IRC code
/:cl:


Varghh:
Is this the return of the cyberiadbot?

AffectedArc07:
> Is this the return of the cyberiadbot? 

No. CyberiadAI was tiger's and that isn't stuff I have access to, nor would I rehost it against his wishes.

This system lays the framework for a simpler bot, which may (unlikely) become a thing, but a lot of features (note viewing) have been replaced by web tools.

AffectedArc07:
Requires https://github.com/ParadiseSS13/rust-g/pull/4 to be reviewed and merged

code/game/world.dm
```diff
@@ -515,3 +516,8 @@ GLOBAL_VAR_INIT(failed_old_db_connections, 0)
     var/dll = world.GetConfig("env", "EXTOOLS_DLL")
     if (dll)
         call(dll, "debug_initialize")()
+
+
+/world/Del()
+	rustg_close_async_http_client() // Close the HTTP client. If you dont do this, youll get phantom threads which can crash DD from memory access violations
+	..()
```
AffectedArc07, line 523: This *shouldnt* get me shot

AffectedArc07:
TODO: Linux release and update `.so` release

code/game/world.dm
AffectedArc07: For context. `world/Reboot()` only triggers via the ingame reboot verb, waiting for round end or the hard reset in DD. Pressing the "Stop" button does not invoke `world/Reboot()`, so I cant use that as an area for thread cleanup.

`world/Del()` is called in all instances of the world deleting, apart from DD being forcefully quit (EG: From task manager, which will unload all the threads regardless if the parent process is killed)

AffectedArc07:
Changes:
- Added config option for forwarding all ahelps to the discord instead of just ones when no admins are online. (Ahelps sent when admins are on do not have the ping, for obvious reasons)
image
- Made it so that if you do not supply a role in the discord, it handles it gracefully
image

Dumbdumn5:
When this works, praise the bot. :+1:

AffectedArc07:
It compiles, but this still needs to be:
A: testmerged for stability
B: I need to debug the latest rust build to ensure thread sanity on an actual machine, not a laptop

AffectedArc07:
Tested locally with proper debugging and it properly cleans up threads on any world close, and has safeties in place for other rarities. 

Now this just needs a TM.

code/controllers/configuration.dm
```diff
@@ -746,6 +737,18 @@
 					config.enable_gamemode_player_limit = 1
 				if("byond_account_age_threshold")
 					config.byond_account_age_threshold = text2num(value)
+				// Discord stuff
+				if("enable_discord_webhooks")
+					discord_webhooks_enabled = TRUE
+				if("discord_webhooks_admin_role_id")
+					discord_admin_role_id = "[value]" // This MUST be a string because BYOND doesnt like massive integers
```
farie82, line 744: isn't `value` a string anyway? So just this would work:
```suggestion
					discord_admin_role_id = value // This MUST be a string because BYOND doesnt like massive integers
```

code/datums/http.dm
```diff
@@ -0,0 +1,152 @@
...
+/datum/http_request/proc/prepare(_method, _url, _body = "", list/_headers)
+	if(!length(_headers))
+		_headers = ""
+	else
+		_headers = json_encode(_headers)
+
+	method = _method
+	url = _url
+	body = _body
+	headers = _headers
```
farie82, line 54:
```suggestion
		headers = ""
	else
		headers = json_encode(_headers)

	method = _method
	url = _url
	body = _body
```
Id just assign the `headers` up top. Else you make a list variable have a string as value.

code/datums/http.dm
```diff
@@ -0,0 +1,152 @@
...
+  * Response deserializer
+  *
+  * Takes a HTTP request object, and converts it into a [/datum/http_response]
+  * The entire thing is wrapped in try/catch to ensure it doesnt break on invalid requests
+  * Can be called on async and blocking requests
+  */
+/datum/http_request/proc/into_response()
+	var/datum/http_response/R = new()
+
+	try
```
farie82, line 123: The fabled try catch. Not even with syntax highlighting

code/game/world.dm
```diff
@@ -515,3 +518,8 @@ GLOBAL_VAR_INIT(failed_old_db_connections, 0)
...
```
farie82, line 525: Does `world/Del()` even have a parent? More out of curiosity

farie82 (changes requested):

code/game/world.dm
ZeWaka: It could have other declarations, which this would then call.

code/controllers/configuration.dm
AffectedArc07: Better safe than sorry incase config parsing is ever changed without it being fully thought through (Or as an internal BYOND update which breaks everything)

code/datums/http.dm
AffectedArc07: Ill get this sorted when I get time to

code/controllers/configuration.dm
farie82: If either changes then a whooooole lot more will fail as well. Not just the discord integration

ZeWaka:
Are the rust-g modifications going to be upstreamed?

AffectedArc07:
> Are the rust-g modifications going to be upstreamed?

@ZeWaka You are welcome to, but I personally wont since:
A) I have very little time at the minute, hence why updates to this PR took so long
B) There are people far more competent at rust than me who can likely do this in a much better way

code/controllers/configuration.dm
AffectedArc07: Im keeping the safety check in, because it also makes it very clear to anyone looking at it that the value is a string. Someone can quite easily forget this is a string and put an int in its place because IDs are numbers, then break everything

farie82: It's more that `value` is always a string. You'll need to parse it to a number like on line 739. Only time this can break is when `file2list` suddenly started to return non strings.

AffectedArc07:
This was TM'd for 2 rounds with no issues.

code/controllers/subsystem/discord.dm
```diff
@@ -0,0 +1,83 @@
...
+	var/alerttext
+	var/list/admincounter = staff_countup(R_BAN)
+	var/active_admins = admincounter[1]
+	var/inactive_admins = admincounter[3]
+	var/total_admins = active_admins + inactive_admins
+	var/add_ping = TRUE
+
+	if(active_admins <= 0)
+		if(inactive_admins > 0)
+			alerttext = "All admins AFK ([inactive_admins]/[total_admins])!"
```
farie82, line 55: Here all admins are AFK so saying X/X admins AFK won't add much

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -34,62 +34,45 @@ GLOBAL_LIST_INIT(adminhelp_ignored_words, list("unknown","the","a","an","of","mo
...
-			modholders += X
-			continue
-		if(check_rights(R_MENTOR, 0, X.mob))
-			mentorholders += X
-			continue
+	switch(selected_type)
+		if("Adminhelp")
+			//See how many staff are on
+			var/admin_number_afk = 0
+			var/list/mentorholders = list()
```
farie82, line 41: Unused var

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -34,62 +34,45 @@ GLOBAL_LIST_INIT(adminhelp_ignored_words, list("unknown","the","a","an","of","mo
...
+			for(var/client/X in GLOB.admins)
+				if(check_rights(R_ADMIN, 0, X.mob))
+					if(X.is_afk())
+						admin_number_afk++
+					adminholders += X
+					continue
+				if(check_rights(R_MOD, 0, X.mob))
+					modholders += X
+					continue
+				if(check_rights(R_MENTOR, 0, X.mob))
```
farie82, line 53: Part of the unused var

PR #14763, opened 2020-10-28
Adds proper snake cases to SSPersistence

TDSSS:
## What Does This PR Do
Adds consistent, snake_cased proc names to the persistence subsystem.

I normally wouldn't bother with such a minor thing but I was working on something related to persistence anyway, decided to fix this stuff while I was there, then discovered I don't need the actual persistence system for what I'm working on, so I'm splitting those changes off.

## Why It's Good For The Game
More consistent cases. 
Seeing 
```
PersistentSave()
Read_Memory()
make_babies()
```
all in the same file is just disgusting.

## Changelog
no changelog because purely backend with no gameplay impact at all.

### DIFF
--- a/code/controllers/subsystem/persistent_data.dm
+++ b/code/controllers/subsystem/persistent_data.dm
@@ -17,20 +17,20 @@ SUBSYSTEM_DEF(persistent_data)
 /datum/controller/subsystem/persistent_data/Initialize()
 	// Load all the data of registered atoms
 	for(var/atom/A in registered_atoms)
-		A.PersistentLoad()
+		A.persistent_load()
 	return ..()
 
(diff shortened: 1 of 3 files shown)


AffectedArc07:
While youre doing this, can you add some presence check ternaries into the debug messages? EG: `[family ? list2params(family) : "None"]`

SabreML:
I'm pretty sure this is snake case, actually.

`camelCase`
`PascalCase`
`snake_case`

TDSSS:
Can't believe I fucked that up right after we talked about it in coder chat. Oh well.

Also will get to AAs request soon.

AffectedArc07 (approved):

PR #14765, opened 2020-10-28
Voodoo doll isn't a scam anymore

Cocacolagua:
## What Does This PR Do
Lowers the cost of the voodoo doll dramatically from 13TC to 4TC
Removes the 20% chance for the doll to reveal your name when using it on someone.
Lowered the 50% chance of the doll revealing your location to 20%.
Upped the cooldown from 3 seconds to 7 seconds.
Buffed some effects of the doll:
- Rises the body temperature by 60 instead of 50.
- Weaken time was doubled.
- Deaf time was tripled.
- Dizzy time was doubled.

## Why It's Good For The Game
This item is probably the worst item in the uplink, with it's only redeeming quality being you can forcibly biblefart people if done correctly.
Now it's a cheap way to keep a close eye on your target, and the punishments don't suck that much anymore (they aren't the best though)

## Changelog
:cl:
tweak: Voodoo Doll costs 4TC instead of 13TC.
del: Voodoo Doll  no longer has a chance to reveal the name of the user.
tweak: Voodoo Doll's cooldown is now 7 seconds instead of 3.
tweak: Voodoo Doll's chance to reveal the location of the user is now 20% instead of 50%.
tweak: The stunning, deafening, dizzying and temperature-rising effects of the Voodoo Doll are now stronger.
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/artefact.dm
+++ b/code/game/gamemodes/wizard/artefact.dm
@@ -778,7 +778,7 @@ GLOBAL_LIST_EMPTY(multiverse)
 	var/mob/living/carbon/human/target = null
 	var/list/mob/living/carbon/human/possible = list()
 	var/obj/item/link = null
-	var/cooldown_time = 70 //7s
+	var/cooldown_time = 7 SECONDS
 	var/cooldown = 0
 	max_integrity = 10
 	resistance_flags = FLAMMABLE
--- a/code/game/gamemodes/wizard/artefact.dm
+++ b/code/game/gamemodes/wizard/artefact.dm
@@ -778,7 +778,7 @@ GLOBAL_LIST_EMPTY(multiverse)
 	var/mob/living/carbon/human/target = null
 	var/list/mob/living/carbon/human/possible = list()
 	var/obj/item/link = null
-	var/cooldown_time = 30 //3s
+	var/cooldown_time = 70 //7s
 	var/cooldown = 0
 	max_integrity = 10
 	resistance_flags = FLAMMABLE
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/wizard/artefact.dm
```diff
@@ -778,7 +778,7 @@ GLOBAL_LIST_EMPTY(multiverse)
 	var/mob/living/carbon/human/target = null
 	var/list/mob/living/carbon/human/possible = list()
 	var/obj/item/link = null
-	var/cooldown_time = 30 //3s
+	var/cooldown_time = 70 //7s
```
farie82, line 781:
```suggestion
	var/cooldown_time = 7 SECONDS
```

farie82 (changes requested):

farie82 (approved):

Kyep (commented):
Objection.
The voodoo doll is way too strong for it to be 4 TC.
Among other things, it lets you remotely view your target, force them to say things, force them to attack people near them, force them to move, or weaken them.
Any one of these things would be incredibly powerful. Combined, they're extremely strong. You can weaken people before going to cuff them. You can make them confess to a capital crime while they're in sec processing, thus getting them executed. You can stalk them invisibly like a super camera bug.
This item is way, way too strong to be 4 TC.

Cocacolagua:
Do bear in mind two things:
- The item has a 20% chance to reveal your location to the victim, which gives them a sign that something is wrong.
- You need an item that they had their fingerprints on, gloves protect them from this. Almost all jobs spawn wearing gloves, so you have to convince them to take it off and then grab an object to leave prints on, which by that point you might as well just use a stunprod.

I do however understand the concern of it being too cheap, so I could up the price to 5~7 instead.

Kyep:
I really dislike this item, even before the buffs this PR gives it.
I also dislike buffing it and massively reducing its cost at the same time.
If the cost reduction was removed, without any other buff being added, I'd probably withdraw my objection.

HeWhoMustNOtBe:
> Objection.
> The voodoo doll is way too strong for it to be 4 TC.
> Among other things, it lets you remotely view your target, force them to say things, force them to attack people near them, force them to move, or weaken them.
> Any one of these things would be incredibly powerful. Combined, they're extremely strong. You can weaken people before going to cuff them. You can make them confess to a capital crime while they're in sec processing, thus getting them executed. You can stalk them invisibly like a super camera bug.
> This item is way, way too strong to be 4 TC.

I dont see your reasoning here, vodoo doll is such an underwhelming item that even if you get it for free you always throw it away since it so shitty (I have done it myself a dozen times). You barely get any reward for getting your targets fingerprints

Kyep:
I really dislike the voodoo doll as an item.
If it went away entirely, I would not be sad.
Probably the most fundamental problem with it is that it attempts to many different things, and does all of them badly.
If you want to rework it, by all means either buff it or reduce its cost, but not both in the same PR.
Realistically, we're probably better off scrapping it and replacing it with a new, better designed item.

Cocacolagua:
I'm not going to maintain this PR anymore due to:

- I no longer hold the position this item should exist in the uplink
- merge conflict for a PR i don't like aaargh!!!

PR #14772, opened 2020-10-28
Changes the chances for major events. Terrors/Blobs less common. Others a bit more common

farie82:
## What Does This PR Do
Changes the major event chances. Makes some scale of security or the like.

Highpop (13 sec staff and 8 engineering staff)
|EVENT TYPE  | BEFORE | AFTER  |
| ------------- |:---------:|:--------:|
|NOTHING     | 62.5%    | 58.7%     |
|Carp migration| 1.8%    | 4.9%     |
|Blob| 11.4%    | 7.0%     |
|Meteor| 1.9%    | 7.0%     |
|Abductors| 3.8%    | 5.9%     |
|Traders| 8.5%    | 8.5%     |
|Terrors| 9.2%    | 7.2%     |
|Demon| 0.7%    | 1.0%     |

Lowpop (7 sec staff and 4 engineering staff)
|EVENT TYPE  | BEFORE | AFTER  |
| ------------- |:---------:|:--------:|
|NOTHING     | 70.9%    | 65.2%     |
|Carp migration| 1.1%    | 3.4%     |
|Blob| 6.4%    | 5.5%     |
|Meteor| 1.1%    | 5.5%     |
|Abductors| 4.3%    | 4.5%     |
|Traders| 9.7%    | 9.4%     |
|Terrors| 5.6%    | 5.3%     |
|Demon| 0.8%    | 1.1%     |

All rounded down to 1 decimal.

## Why It's Good For The Game
Terrors or blobs take up over 20% of all high pop midround spawns. Which makes about 1/5 of all "normal" high pop rounds turn into an action packed round.
This PR tries to reduce the chances of those two and increase the chances of the other events which are less chaotic.

## Changelog
:cl:
tweak: The event chances are changed. Read the PR for details
/:cl:

### DIFF
--- a/code/modules/events/event_container.dm
+++ b/code/modules/events/event_container.dm
@@ -189,17 +189,17 @@ GLOBAL_LIST_EMPTY(event_last_fired)
 /datum/event_container/major
 	severity = EVENT_LEVEL_MAJOR
 	available_events = list(
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Nothing",			/datum/event/nothing,			1320),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Carp Migration",	/datum/event/carp_migration,	0,						list(ASSIGNMENT_SECURITY =  3), TRUE),
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Nothing",			/datum/event/nothing,			580),															// 57% on high pop (120+). 64.2% on lowpop (70+)
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Carp Migration",	/datum/event/carp_migration,	10,						list(ASSIGNMENT_SECURITY =  3), TRUE),	// 4.8% on high pop, 3.4% on low pop
 		//new /datum/event_meta(EVENT_LEVEL_MAJOR, "Containment Breach",	/datum/event/prison_break/station,	0,			list(ASSIGNMENT_ANY = 5)),
 		new /datum/event_meta(EVENT_LEVEL_MAJOR, "APC Overload",	/datum/event/apc_overload,		0),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Blob",			/datum/event/blob, 				0,						list(ASSIGNMENT_ENGINEER = 30), TRUE),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Meteor Wave",		/datum/event/meteor_wave,		0,						list(ASSIGNMENT_ENGINEER =  5),	TRUE),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Abductor Visit",	/datum/event/abductor, 		    80, is_one_shot = TRUE),
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Blob",			/datum/event/blob, 				30,						list(ASSIGNMENT_ENGINEER =  5), TRUE),	// 6.9% on high pop, 5.5% on low pop
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Meteor Wave",		/datum/event/meteor_wave,		30,						list(ASSIGNMENT_ENGINEER =  5),	TRUE),	// 6.9% on high pop, 5.5% on low pop
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Abductor Visit",	/datum/event/abductor, 		    20, 					list(ASSIGNMENT_SECURITY =  3), TRUE),	// 5.8% on high pop, 4.5% on low pop
 		//new /datum/event_meta/alien(EVENT_LEVEL_MAJOR, "Alien Infestation",	/datum/event/alien_infestation, 		0,		list(ASSIGNMENT_SECURITY = 15), TRUE),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Traders",			/datum/event/traders,			180, is_one_shot = TRUE),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Terror Spiders",			/datum/event/spider_terror, 		0,			list(ASSIGNMENT_SECURITY = 15), is_one_shot = TRUE),
-		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Slaughter Demon",	/datum/event/spawn_slaughter,	15, is_one_shot = TRUE)
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Traders",			/datum/event/traders,			85, is_one_shot = TRUE),										// 8.4% on high pop, 9.4% on low pop
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Terror Spiders",	/datum/event/spider_terror, 	20,						list(ASSIGNMENT_SECURITY = 4), TRUE),	// 7.1% on high pop, 5.3% on low pop
+		new /datum/event_meta(EVENT_LEVEL_MAJOR, "Slaughter Demon",	/datum/event/spawn_slaughter,	5, 						list(ASSIGNMENT_SECURITY = 2), TRUE)	// 3% on high pop, 2.1% on low pop
 		//new /datum/event_meta(EVENT_LEVEL_MAJOR, "Floor Cluwne",	/datum/event/spawn_floor_cluwne,	15, is_one_shot = TRUE)
 	)
 
--- a/code/modules/events/event_container.dm
+++ b/code/modules/events/event_container.dm
@@ -189,7 +189,7 @@ GLOBAL_LIST_EMPTY(event_last_fired)
 /datum/event_container/major
 	severity = EVENT_LEVEL_MAJOR
 	available_events = list(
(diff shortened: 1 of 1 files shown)


TheMaveran:
Yeah it’s getting pretty old seeing terrors/blob what feels like every other round.

Kyep:
I'm not sure how I feel about this.
On the one hand, in theory, the idea of having diverse major events sounds nice.
However, I can't help but feel this tends to ignore the realities of how events actually work.

For example, this massively increases the probability of space carp migration events (over 2x in highpop and 3x in lowpop).
Carp migration is unlikely to affect you unless you are the space pod pilot. It isn't really impactful enough to make it a major event.
Why, exactly, should we make a major event that's safely ignored by the vast majority of crew more common? It might increase diversity but its not meaningful.
It don't think it results in a more fun or interesting game.

Conversely, the demon event is incredibly impactful. A demon will result in red alert with many crew being armed.
Right now, the demon event has a tiny 0.7-0.8% chance. This PR increases that to 2-5x what it was before.
I'm not sure that's a good change. While it does increase diversity, is it really a good idea to have 3x or 4x more common slaughter demon events?
I feel that this could turn what is essentially a very rare spawn right now, into a "another 3% of shifts always reach gamma due to slaughter demons murdering people indiscriminately". That's a significant change.

Let's take a third example. Blob.
This drastically cuts the odds of blob, from 11.4% to 6.9% on highpop, and 6.4% to 5.5% on lowpop.
That's an effective reduction of ~40% on highpop, which I am not sure is a good thing.
Blob is one of the most interesting major events. Its certainly more interesting than meteor, abductors, etc.
To drastically reduce the odds of one of the most interesting lateround events seems... :/

I get that you're trying for diversity here, but IMHO, you need to consider how interesting the different events are. Increasing diversity is NOT automatically good if you're shifting odds away from interesting events towards less interesting ones.
Basically, you can re-allocate odds from less interesting major events (like nothing) to more interesting ones (like carp, which is more interesting than nothing).
But its probably not a good idea to re-allocate odds from more interesting events (like blob) to ones that have essentially no impact on players (like carp, which is barely above nothing).

Kyep (commented):
While the idea of "more diverse events" might *sound* nice, the details of this have a lot of issues.
It feels like this somewhat ignores the realities of how the various events actually work in practice.

For example, this massively increases the probability of space carp migration events (over 2x in highpop and 3x in lowpop).
Carp migration is unlikely to affect you unless you are the space pod pilot. It isn't really impactful enough to make it a major event at all, and should probably be taken out of the major events list entirely.
Why, exactly, should we make a major event that's safely ignored by the vast majority of crew more common? It might increase diversity but its not meaningful diversity - its taking away chance of something interesting happening, in favor of higher chance of the yawnfest that is a space carp event.
It don't think these sort of changes make the game more fun or interesting.
Your changes to meteors fall into the same bucket. While in theory a tunguska can create an interesting event, in practice meteor damage is usually superficial and ignorable, thus the meteor event being more common is not great.

Conversely, the demon event is incredibly impactful. A demon will result in red alert with many crew being armed.
Right now, the demon event has a tiny 0.7-0.8% chance. This PR increases that to 2-5x what it was before.
I'm not sure that's a good change. While it does increase diversity, is it really a good idea to have 3x or 4x more common slaughter demon events?
I feel that this could turn what is essentially a very rare spawn right now, into a "another 3% of shifts always reach gamma due to a slaughter demon murdering people indiscriminately". That's a significant change. I'm really not sure that "one super-powered creature that murders everything in sight, full-heals from each kill and can phase around" is fun for the majority of players. So I'm really skeptical about increasing the odds of this event.

Let's take a third example. Blob.
This drastically cuts the odds of blob, from 11.4% to 6.9% on highpop, and 6.4% to 5.5% on lowpop.
That's an effective reduction of ~40% on highpop, which I am not sure is a good thing.
Blob is one of the most interesting major events. Its certainly more interesting than meteor, abductors, etc.
To drastically reduce the odds of one of the most interesting lateround events seems... :/

I get that you're trying for diversity here, but IMHO, you need to consider how interesting the different events are. Increasing diversity is NOT automatically good if you're shifting odds away from interesting events towards less interesting ones.
Basically, you can re-allocate odds from less interesting major events (like nothing) to more interesting ones (like carp, which is more interesting than nothing).
But its not a good idea to re-allocate odds from more interesting events (like blob) to ones that have essentially no impact on players (like carp, which is barely above nothing).

Also, you completely ignore xenos in this (they're commented out, sure, but they are unlikely to stay that way forever).

You're also shifting odds from pop-scaled (ie: event's chance is mostly based on pop in the relevant department, e.g. sec players for sec threats) to more static. That greatly increases the odds that the station won't have enough players in the relevant department to deal with whichever event happens. Which, I think, creates a not-great player experience.

The more I look at this, the more issues I find with it. I think I'm going to have to object to it.

Fox-McCloud:
A step in the right direction, in my opinion.

I do think demons should stay incredibly rare though; if it is to be increased, keep it something like a flat 1% no matter population; I feel 2-3% is going to wear it out too often.

farie82:
> I do think demons should stay incredibly rare though; if it is to be increased, keep it something like a flat 1% no matter population; I feel 2-3% is going to wear it out too often.

> Right now, the demon event has a tiny 0.7-0.8% chance. This PR increases that to 2-5x what it was before.
> I'm not sure that's a good change. While it does increase diversity, is it really a good idea to have 3x or 4x more common slaughter demon events?

Gonna agree with these. Will change it to about 1%



> For example, this massively increases the probability of space carp migration events (over 2x in highpop and 3x in lowpop).
> Carp migration is unlikely to affect you unless you are the space pod pilot. It isn't really impactful enough to make it a major event at all, and should probably be taken out of the major events list entirely.
> Why, exactly, should we make a major event that's safely ignored by the vast majority of crew more common? It might increase diversity but its not meaningful diversity - its taking away chance of something interesting happening, in favor of higher chance of the yawnfest that is a space carp event.
> It don't think these sort of changes make the game more fun or interesting.
> Your changes to meteors fall into the same bucket. While in theory a tunguska can create an interesting event, in practice meteor damage is usually superficial and ignorable, thus the meteor event being more common is not great.

For both I've made them more common and removed the chance from nothing to happen. Engineering usually gets pretty bored halfway so meteors will spice it up a bit. They can also help antags with escaping due to the open breaches.

For the carp. It is just an event compared to nothing. People in maint will have a laugh when a carp kills itself while breaching into the station.

> Let's take a third example. Blob.
> This drastically cuts the odds of blob, from 11.4% to 6.9% on highpop, and 6.4% to 5.5% on lowpop.
> That's an effective reduction of ~40% on highpop, which I am not sure is a good thing.
> Blob is one of the most interesting major events. Its certainly more interesting than meteor, abductors, etc.
> To drastically reduce the odds of one of the most interesting lateround events seems... :/

In my opinion blob is one of the lesser events... it usually involves to an attrition war. Which is fine every now and then. But it happens a lot. Which just makes fighting blobs tiresome once you play the game for longer. I myself (and others I know) would like less action in my para rounds on average which the reduction here helps with.


> I get that you're trying for diversity here, but IMHO, you need to consider how interesting the different events are. Increasing diversity is NOT automatically good if you're shifting odds away from interesting events towards less interesting ones.
> Basically, you can re-allocate odds from less interesting major events (like nothing) to more interesting ones (like carp, which is more interesting than nothing).
> But its not a good idea to re-allocate odds from more interesting events (like blob) to ones that have essentially no impact on players (like carp, which is barely above nothing).

> Also, you completely ignore xenos in this (they're commented out, sure, but they are unlikely to stay that way forever).

I'm not gonna balance them out since that will break the math for the existing code. When uncommented they should be balanced accordingly.

> You're also shifting odds from pop-scaled (ie: event's chance is mostly based on pop in the relevant department, e.g. sec players for sec threats) to more static. That greatly increases the odds that the station won't have enough players in the relevant department to deal with whichever event happens. Which, I think, creates a not-great player experience.

I've lowered the big numbers to make the math easier. 2 or 20 for each sec member or 10 or 100 statically. The percentages don't care since it cares about the relativity compared to the total. The total of high pop is about 1000 which made doing the math easier. (10 is ~1% chance)
A PR can be made later (or this one expanded) to include jobs like science to the equation. Since a blob scaling solely on engineering is odd since engineering has very little to do with a blob. Emitters suck against them nowadays and the only other thing they got are hardsuits and welders which can be ordered by cargo or made by science (in the case of welders).

AffectedArc07 (approved):

Kyep:
I still dislike the idea of re-allocating odds away from interesting events (like blob, terrors, and even traders on lowpop) in favor of snoozefest easily-ignored non-events like the carp migration. It just feels like watering down the impact of major events and turning more rounds into drawn-out affairs where nothing major happens in the second half.

Spacemanspark:
Personally I'm kind of tired of seeing the same exact things (blob and terrors) over and over again. Especially with how much both are allowed to meta (or meta in ways they aren't supposed to anyways--looking at you, blob mice. Yeah, I see you in the AI core. Don't lie).

They're fun, don't get me wrong, but they do happen far too often. I'd rather see some of the other events get more of a chance to come through, for the sake of round to round diversity.

Dumbdumn5 (approved):
I actually really like the fact that seldom-seen events are increased in chance in this, even demon, as it may actually make rounds feel more interesting by way of not having back-to-back-to-back terror/blob/trader rounds sometimes. Even if it can, at times, be boring, it's better than what we've got in my opinion!

PR #14778, opened 2020-10-29
Saves the runechat preference when the verb is used

farie82:
## What Does This PR Do
Makes using the toggle runechat verb save the preferences

## Why It's Good For The Game
Was forgotten to add so it doesn't save the preferences now unless you change another preference

## Changelog
:cl:
fix: Your runechat preference is always saved now
/:cl:


AffectedArc07 (approved):
This is what happens when you convert testmerge code to real code

PR #14784, opened 2020-10-30
Makes it so only engineer, heads, and other command staff can open station goal crates

warior4356:
## What Does This PR Do
As the title states, this PR locks the station goal crates to everyone with access to engineering: engineering, CE, atmos, heads, IAA, BS, Magistrate, and captain

## Why It's Good For The Game
Particularly with the shield sats, it's really annoying for items that are in goal crates to not fit back into the crate. Also, station goal items should really be the domain of engineers and not go missing to random people. Finally, by making them secure crates they can no longer lose items, removing annoyance with missing goal items, particularly with newer crews who don't realize an item is missing.

## Changelog
:cl:
add: Added station goals to crates locked to engineering access.
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -1790,7 +1790,8 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	cost = 10
 	special = TRUE
 	containername = "empty station goal crate"
-	containertype = /obj/structure/closet/crate/engineering
+	containertype = /obj/structure/closet/crate/secure/engineering
+	access = ACCESS_CONSTRUCTION
 
 /datum/supply_packs/misc/station_goal/bsa
 	name = "Bluespace Artillery Parts"


code/datums/supplypacks.dm
```diff
@@ -1790,7 +1790,8 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	cost = 10
 	special = TRUE
 	containername = "empty station goal crate"
-	containertype = /obj/structure/closet/crate/engineering
+	containertype = /obj/structure/closet/crate/secure/engineering
+	access = ACCESS_CONSTRUCTION
```
AffectedArc07, line 1794: Curios where theres a diff on `1793`. Did you change anything?

MarsM0nd:
Don't think every station goal should be getting locked down. Satellites, okay, really a thing only engineering needs to touch, and a problem if opened. But for example the DNA vault, suddenly engineering has to become cargo to distribute the samplers to botany and science, for pet and plant DNA.
It'll also stop any other potential delivery not inside the crate. Manually handing it over is harder and can be lost if you have to drag the crate around, keeping boards all in a handy box won't be possible either, which may just make engineers not have the space to transport all the artillery ones at the time. Even just combining the items in a single crate won't be possible, so suddenly you need to transport 3 crates instead of just one.
As for the missing item from shipping, how should people learn that it can happen? It has to happen in shipments that are noticeable, and not many are as noticeable as the station goal. Don't think we should just lock crates to stop that from happening. Refactoring it to check for a flag whether items can go missing would be better if we just want to stop that, especially as text searching the path sure isn't OOP.

code/datums/supplypacks.dm
SteelSlayer: Changed to a secure crate: `/crate/engineering` > `/crate/secure/engineering`

AffectedArc07: My eyes actually need checking, and I guess this is what happens when you use the split git view instead of the layered one (Its much more obvious like this)
image

Compared to this
image

procdrone:
Maybe a solution, make satellites fit in crates/lockers?

code/datums/supplypacks.dm
warior4356: Resolving since this isn't an issue.

variableundefined:
Objecting along the same line of reasoning as Marsmond

AffectedArc07:
Not in favour due to how this makes distributing DNA samplers a pain, and this isn't really the solution to crates being randomly opened. What proc said about making them fit back in crates is a better solution. 

Closing due to 2 objections.

Issue #14766, opened 2020-10-28
Singulo Can Phase Through Containment


Leanfrog:
**Issue Description**:
The singulo was able to phase through containment despite only being at stage 2

**What did you expect to happen**: 
For the singulo to not get loose

**What happened instead**:
The singulo just up and left containment making the station RIP in peace 

**Why is this bad/What are the consequences**:
Singulo shouldn't be able to phase through containment at stage 2

**Steps to reproduce the problem**:
No clue, what lead up to it happening was it got stranded  in containment at stage one, so it was fed like 5 items to give it a boost so it would move into the PA's line of fire, then once it hit stage 2 it just kind of phased through the containment walls

**When did the problem start happening**:
Wish I knew, but with what little info I have I can't make a guess

**Extra information**:

AffectedArc07:
Was it directly adjacent to containment walls when it went loose? Chances are it expanded and ate its own containment.

Leanfrog:
Containment was left untouched, it physically just left containment while being inside

Kepplerx:
Containment was even standing after it left, and continued to do so, until minutes afterwards

PR #14787, opened 2020-10-30
Fixes the singulo going through the containment when it shouldn't be able to

farie82:
## What Does This PR Do
Makes the singulo not eat the containment when it shouldn't. Or noclip through it due to weirdness

Issue was with the sleep in the `CHECK_TICK`. This could cause the singulo to have multiple `process` calls open which could move it while it was still eating or worse. Make it grow. Making it's eat distance bigger causing it to eat things that it shouldn't eat.

Now it just stops eating when it has taken enough of the current ticks time. Ensuring no sleep occurs and thus only one process call ever exists

fixes: #14766 and other related singulo accidents on highpop

## Why It's Good For The Game
Dear god why

## Changelog
:cl:
fix: The singulo now can't go through containment when it shouldn't be able to
/:cl:

### DIFF
--- a/code/modules/power/singularity/singularity.dm
+++ b/code/modules/power/singularity/singularity.dm
@@ -268,7 +268,8 @@
 					X.singularity_pull(src, current_size)
 				else
 					consume(X)
-			CHECK_TICK
+			if(TICK_CHECK)
+				return // You've eaten enough. Prevents weirdness like the singulo eating the containment on stage 2
 
 
 /obj/singularity/proc/consume(atom/A)


code/modules/power/singularity/singularity.dm
```diff
@@ -268,7 +268,8 @@
 					X.singularity_pull(src, current_size)
 				else
 					consume(X)
-			CHECK_TICK
+			if(TICK_CHECK)
```
AffectedArc07, line 271: = `>80% of tick`

AffectedArc07 (approved):

PR #14792, opened 2020-10-30
Bloodcrawling now has a blood-boiling animation

Cocacolagua:
## What Does This PR Do
Adds a blood sparkles animation to a blood turf when a mob is bloodcrawling out of it.
- [ ] Unlike the branch's name, I will have to change the sprite
- [x] Make the animation change colours depending on the blood pool's colour.
- [ ] Cancel the animation when done multiple times.

## Why It's Good For The Game
The biggest reason why slaughter demons are considered "overpowered" is because the crew doesn't really know which blood tile they are coming from, which doesn't allow them to prepare and usually ends with one of them dead.
This change forces the demon to try different strategies, rather than the one that makes them basically unkillable.

## Images of changes
A8W3mn5Qs4
EEmcYTGqdw

## Changelog
:cl:
add: Bloodcrawling out of blood now has an animation.
/:cl:


AffectedArc07:
image
Im going to request you do, because red EMP sparkles do not suit this *at all*

Cocacolagua:
fair enough

Spacemanspark:
Initially, before looking at the gif I thought this left a trail of sparkles whenever they moved. I was about to cry.

This seems... fine, though.

code/game/gamemodes/miniantags/slaughter/bloodcrawl.dm
```diff
@@ -96,21 +96,27 @@
...
 
 	if(notransform)
 		to_chat(src, "<span class='warning'>Finish eating first!</span>")
-		return 0
+		return FALSE
 	B.visible_message("<span class='warning'>[B] starts to bubble...</span>")
+	var/obj/effect/temp_visual/uniquebloodcrawl = new /obj/effect/temp_visual/bloodcrawl(B.loc)
+	uniquebloodcrawl.color = B.color
 	if(!do_after(src, 20, target = B))
 		return
```
farie82, line 121:
```suggestion
		qdel(uniquebloodcrawl)
		return
```
Currently the effect will stay even if the demon cancels the phasing or when the blood gets removed.

code/game/gamemodes/miniantags/slaughter/bloodcrawl.dm
```diff
@@ -96,21 +96,27 @@
...
+	duration = 20
+
 /mob/living/proc/phasein(var/obj/effect/decal/cleanable/B)
 
 	if(notransform)
 		to_chat(src, "<span class='warning'>Finish eating first!</span>")
-		return 0
+		return FALSE
 	B.visible_message("<span class='warning'>[B] starts to bubble...</span>")
+	var/obj/effect/temp_visual/uniquebloodcrawl = new /obj/effect/temp_visual/bloodcrawl(B.loc)
```
farie82, line 118:
```suggestion
	var/obj/effect/temp_visual/bloodcrawl/uniquebloodcrawl = new(B.loc)
```

farie82 (changes requested):

Kiyahitayika:
Thank you. The worst is the blood under the Operating Table. One time, I made an incision on a patient, blood dripped, _oof_. All of a sudden, I'm on the floor, unconscious patient being dragged around for more blood. I'm pretty sure I AHELPed because I was so spooked and was like "WHAT! WHERE DID IT COME FROM! TH-TH-THIS IS POWERGAMING! REEE!" (spoiler: it wasn't and it's not) 😄 I would be concerned about giving people "too much" of a head's up, but the message to users in range/view of the blood spot is there for a reason. I definitely appreciate this!

AffectedArc07 (changes requested):
image
Holding an objection until this sprite is changed to something which actually represents bubbling or something else, atleast something other than a reskinned EMP sparkle.

farie82 (approved):

Cocacolagua:
Closing this since i can’t make a good sprite for blood boiling, anyone’s free to remake this if they have one though

PR #14804, opened 2020-10-31
Adds confirmation message to CC packed crates

warior4356:
## What Does This PR Do
Adds a warning and confirmation prompt to opening any crate with items inside that won't fit back inside the crate.

## Why It's Good For The Game
To prevent items from going all over cargo that then have to be transported one by one as they wont fit back into the crate they came in. (Also it will reduce the number of engineer on cargo tech rage killings)

## Images of changes
image

## Changelog
:cl:
add: Added a confirmation prompt to opening crates with too big items inside.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -28,18 +28,19 @@
 /obj/structure/closet/crate/can_close()
 	return TRUE
 
-/obj/structure/closet/crate/open()
+/obj/structure/closet/crate/open(irregular_open = FALSE)
 	if(src.opened)
 		return FALSE
 	if(!src.can_open())
 		return FALSE
 
-	for(var/obj/O in src)
-		if(O.density)
-			var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
-			if(response == "No")
-				return FALSE
-			break
+	if(!irregular_open)
+		for(var/obj/O in src)
+			if(O.density)
+				var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
+				if(response == "No")
+					return FALSE
+				break
 
 	if(rigged && locate(/obj/item/radio/electropack) in src)
 		if(isliving(usr))
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -28,17 +28,17 @@
 /obj/structure/closet/crate/can_close()
 	return TRUE
 
-/obj/structure/closet/crate/open(irregular_open = FALSE)
+/obj/structure/closet/crate/open(by_hand = FALSE)
 	if(src.opened)
 		return FALSE
 	if(!src.can_open())
 		return FALSE
 
-	if(!irregular_open)
+	if(by_hand)
 		for(var/obj/O in src)
 			if(O.density)
 				var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
-				if(response == "No")
+				if(response == "No" || !Adjacent(usr))
 					return FALSE
 				break
 
@@ -90,6 +90,10 @@
 	if(!opened && try_rig(W, user))
 		return
 	return ..()
+
+/obj/structure/closet/crate/toggle(mob/user, by_hand = FALSE)
+	if(!(opened ? close() : open(by_hand)))
+		to_chat(user, "<span class='notice'>It won't budge!</span>")
 
 /obj/structure/closet/crate/proc/try_rig(obj/item/W, mob/user)
 	if(istype(W, /obj/item/stack/cable_coil))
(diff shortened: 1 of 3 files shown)


code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -34,6 +34,17 @@
...
+	var/cc_packed = FALSE
+	for(var/obj/O in src)
+		if(O.density)
+			cc_packed = TRUE
+			break
+
+	if(cc_packed)
+		var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
+		if(response == "No")
+			return FALSE
```
SteelSlayer, line 46: You can eliminate the need for the `cc_packed` var by doing something like this:
```dm
for(var/obj/O in src)
	if(O.density)
		var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
		if(response == "No")
			return FALSE
		break
```

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -34,6 +34,13 @@
 	if(!src.can_open())
 		return FALSE
 
+	for(var/obj/O in src)
+		if(O.density)
+			var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
```
farie82, line 39: This will break things like the knock spell or revenants opening things with defile.
You'll need to solve this at the point where it is sure that a crewmember opens it by hand and not by a power or such

farie82 (changes requested):

code/game/objects/structures/crates_lockers/crates.dm
warior4356: I added an argument to the proc to handle that edge case. Let me know if you think of any other cases.

warior4356:
Do not merge for the moment, I need feedback from maints on resolving farie's issue. Should it be handled in the edge cases or handled in the base case?

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -28,12 +28,20 @@
...
+/obj/structure/closet/crate/open(irregular_open = FALSE)
 	if(src.opened)
 		return FALSE
 	if(!src.can_open())
 		return FALSE
 
+	if(!irregular_open)
+		for(var/obj/O in src)
+			if(O.density)
+				var/response = alert(usr, "This crate has been packed with bluespace compression, an item inside won't fit back inside. Are you sure you want to open it?","Bluespace Compression Warning", "No", "Yes")
```
AffectedArc07, line 40: Add a check to make sure the person doesnt open the dialogue box then walk away, and remotely open the crate.

AffectedArc07 (changes requested):

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -28,12 +28,20 @@
...
 	return TRUE
 
-/obj/structure/closet/crate/open()
+/obj/structure/closet/crate/open(irregular_open = FALSE)
 	if(src.opened)
 		return FALSE
 	if(!src.can_open())
 		return FALSE
 
+	if(!irregular_open)
```
farie82, line 37: There are still quite some other cases list `bust_open` (not `burst_open`... yeah).
Also when one is EMPed. Or when the `effect/overload` processes. Or when the open door magic hits this.
Be sure to cover all the cases. Also cases where `usr` is null or such

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -28,12 +28,20 @@
 /obj/structure/closet/crate/can_close()
 	return TRUE
 
-/obj/structure/closet/crate/open()
+/obj/structure/closet/crate/open(irregular_open = FALSE)
```
farie82, line 31: I'm not 100% on the param name. `irregular_open` doesn't tell much about what it does when it's `TRUE` or `FALSE`. Adding comments will help with that. Though a more descriptive param name would be better. (combination is also always good)

variableundefined:
Approved for design, will wait for code issues to be addressed

Kyep (approved):

code/game/objects/structures/crates_lockers/crates.dm
warior4356: I inverted the check to be checking positive instead of negative.

code/game/objects/structures/crates_lockers/crates.dm
warior4356: It's now by_hand and confirms that it was opened by hand, rather than if it was opened improperly

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -83,6 +91,10 @@
 		return
 	return ..()
 
+/obj/structure/closet/crate/toggle(mob/user, by_hand = FALSE)
+	if(!(opened ? close() : open(by_hand)))
+		to_chat(user, "<span class='notice'>It won't budge!</span>")
```
moxian, line 96: too much punctuation to comfortably parse
```suggestion
	var/success = opened ? close() : open(by_hand)
	if(!success)
		to_chat(user, "<span class='notice'>It won't budge!</span>")
```

warior4356: I'm duplicating the closet toggle proc mostly, just to add an argument and I'd rather not change it.

moxian: Alright, fair.

warior4356:
This PR is currently done and ready for merge, if anyone wants to check the code again, I made some significant changes.

farie82 (approved):
Crate code sure does need a refactor. But that's outside the scope of this PR. LGTM

AffectedArc07 (approved):

PR #14813, opened 2020-11-02
Buff EMPs against augments, remove duplicate external organ emp_act (buffs IPCs vs. EMPs)

dearmochi:
## What Does This PR Do

- **Augmented limbs now take `4` or `2` (depending on severity) burn damage on EMP instead of `1.5` or `0`. For a total of 11 limbs, that's `44`/`22` total burn damage**
- Removed the generic `Stun` (which doesn't even stack) on augmented limbs from EMPs, as it is replaced by per-limb behaviour. Upon EMP:
  - An augmented upper body will cause `4`/`2` breath loss and `20`/`10` stamina damage
  - An augmented lower body will cause `20`/`10` toxin damage
  - An augmented arm or hand will drop whatever it's holding
  - An augmented leg or foot will weaken by `4`/`2`, stacking
  - An augmented head will confuse by `30`/`20`
- Remove duplicate `/obj/item/organ/external/emp_act` override which removed the case for IPC limbs - they no longer take prosthetic-level damage from EMPs, which was significantly higher (`9`/`5.5` vs. `20`/`7`).

## Why It's Good For The Game

- **Augments are very old, and very unbalanced. They are immune to stings, syringes, are spaceproof and immune to fractures. They also take significantly less damage (`-5` brute and `-4` burn **flat**). They *are* supposed to take damage on EMP but the current amount is laughable due to a possible oversight, so it has been adjusted to be more dangerous but not too dangerous, because..**
- **..augmented limbs now have their own behaviour on EMP, which further cripple and immobilize the user, making it an easier victim. While a fully augmented human with CNS Rebooter won't remain on the floor for long, they will be sufficiently injured for an attacker to follow up comfortably**
- An oversight led to `/obj/item/organ/external/emp_act` being duplicate, the second of which didn't include the IPC-specific damage amounts from EMPs, causing them to take massive damage. This is now fixed.

## Images of changes

ezgif-3-7408674291dc

## Changelog
:cl:
tweak: Drastically increase burn damage on augmented limbs on EMP, from 1.5/0 to 4/2 (depending on severity)
tweak: Augmented limbs now have their own effect on EMP depending on the limb, ranging from item dropping, falling to the ground and general damage
fix: Fix IPC limbs taking prosthetic-level damage from EMPs, which was excessive
/:cl:


NetraKyram:
I have to **firmly** disagree with this, on a couple of points

- Toxins Damage is nonsensical, the organs beneath are still organic.
- Oxy/suffocation is a really silly thing too as yet again, the reasons listed above.
- The hand-dropping doesn't really change anything and is redundant if you're already gonna be on the floor from the stuns.
- This legit does nothing but add more EMP/EMP counterplay instead of fixing Augments in a reasonable manner, I'll give you some suggestions of course, as I'm not going to tell you you're doing something wrong without giving you a reasonable alternative.

Augments are great and all, but something people tend to forget, or maybe choose to forget is that they are still organic underneath. Why not add the ability for bones to break from high amounts of damage, just make it a higher chance to not break/etc, to uphold the durability? I agree with the EMP still staying, they are basically cybernetic limbs of course, adding the ability for bones to break, and you to have to go through a proper surgery/such to get it fixed would honestly **heavily balance** the way augments currently are for all their boons.

As you know, and I see no reason to hide it, I've used Augments a lot in the past, on occasion, I still do, and it's perfectly acceptable to call me biased in some sense. I'm simply openly opposing this on the fact that it's not really fixing anything. 

**Augments are powerful, and to an extent unbalanced, but this doesn't fix them, and in fact, places them in the same place they were before, to where a player like me can keep distance/meth themselves up whilst augmented, speed away from your EMP attempt, and kill you at range, as I'd do with anyone with an EMP implant, or EMP's.** The only time EMP's are really effective is if you have the drop on somebody, and if you are, for example, using your augmented character as a Emergency Response Force, to back up other officers, the whole 'EMP Meta' is completely negated.

My point all in all is adding bone breaking, and a surgical system to allow Augmented people to get their organs, and or bones fixed, is a preferential alternative to buffing something that hardly works to begin with outside of exceptional cases.

ppi13:
the odd part is really the external organs causing oxy and toxic damage, usually internal organs stand for that, the fact that they are pierceimmune is also weird, but thats more of a debuff then a buff as well.
As for the stun, well cns rebooters stop working when emped. Not to speak about how the reviver already straight up kills you.
There is also cybernetic organs to think about as well.

Kiyahitayika:
Regarding the changes to how EMPs affect augments, I'm . . . not in favor of this PR right now. My reasoning is flawed, I admit, for it is based upon player behavior. Most players will head to Medbay when they need care. I think one way I could see a build like this working on live is if Robotics and Medbay were merged. Or perhaps if Robotics was separated into Mech/Exosuit Fabrication (merged with Engineering) and Synthetics & Auguments (merged with Medbay). 

> **Augments are very old, and very unbalanced. They are immune to stings, syringes, are spaceproof and immune to fractures. 

These are _very_ good points. It sounds like balancing is indeed needed. However, I am not in favor of EMPs being the solution. IPCs aren't spaceproof, why should cybernetic limbs be? And fracture immunity does not make much sense either.

> Why not add the ability for bones to break from high amounts of damage, just make it a higher chance to not break/etc, to uphold the durability? 

I have been toying with the idea of adding a `bone_mod` separate from `brute_mod` for a while. I have not had a real reason to work on this yet, so if this is a concept that might be useful here, let me know! 
I have been taking a bit of a break from the general Discord. My Discord name is **Iren#5315** ; feel free to hit me up. 

> **They also take significantly less damage (-5 brute and -4 burn flat).** 

I was looking at https://github.com/ParadiseSS13/Paradise/pull/14818 earlier. What if cybernetic limbs had `brute_mod` and `burn_mod`? Flat rates are bad.

> An oversight led to /obj/item/organ/external/emp_act being duplicate, the second of which didn't include the IPC-specific damage amounts from EMPs, causing them to take massive damage. This is now fixed.

I like this part! :D

> An augmented upper body will cause 4/2 breath loss and 20/10 stamina damage
That's a lot of stamina damage.
> An augmented lower body will cause 20/10 toxin damage
That's a lot of toxin damage.

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -17,6 +17,19 @@
 	encased = "ribcage"
 	convertable_children = list(/obj/item/organ/external/groin)
 
+/obj/item/organ/external/chest/emp_act(severity)
+	..()
+	if(!is_robotic() || emp_proof || !tough) // Augmented chest suffocates the user on EMP.
```
SteelSlayer, line 22: I think I would add a check for `owner` here so you don't have to use so many `?.` operators. Also, with the code as is, the `to_chat` doesn't check if owner is null, which can runtime.

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -30,6 +43,17 @@
 	amputation_point = "lumbar"
 	gendered_icon = 1
 
+/obj/item/organ/external/groin/emp_act(severity)
+	..()
+	if(!is_robotic() || emp_proof || !tough) // Augmented groin does toxin damage to the user on EMP.
```
SteelSlayer, line 48: Ditto about the owner check.

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -210,3 +286,14 @@
 /obj/item/organ/external/head/set_dna(datum/dna/new_dna)
 	..()
 	new_dna.write_head_attributes(src)
+
+/obj/item/organ/external/head/emp_act(severity)
+	..()
+	if(!is_robotic() || emp_proof || !tough) // Augmented head confuses the user on EMP.
```
SteelSlayer, line 292: Ditto about the owner check.

Spacemanspark:
Perhaps the IPC fix and the balance change should be in separate PRs

Aquilar:
how about just lower the damage reduction to 5%? or remove damage reduction entirely, rather than trying to balance the damage reduction with a bunch of cons that let someone with an EMP gun, EMP flashlight, EMP implant or EMP grenades to murder someone with all augments in seconds.

pmitchell:
It looks like organ_internal.dm also has a duplicate /obj/item/organ/internal/emp_act; could you remove that one as well?

Necaladun (approved):
Not enough of a nerf but a step in the right direction and the code allows for some good options when we do overhaul them for individual EMP effects

variableundefined:
Both Affected and I disagree with making robotic augments taking toxin and oxygen damage. Since this PR has two maint objections and is unlikely to receive an approval anytime soon I'll close it.

variableundefined:
Reopening. The PR can go through with the aforementioned changes removed.

PR #14817, opened 2020-11-03
Cuts the price of summon guns in half, now only 2 points

Qwertytoforty:
## What Does This PR Do

Cuts the price of lesser summon guns in half, making it cost 2 wizard points instead of 4. Reduces cooldown by 15 seconds, now takes a minute to recharge.

## Why It's Good For The Game

Right. Lesser summon guns right now... is not good. 4 points, almost half of what you get. It summons around 20 guns for you to use, which sounds nice, but you need both hands to use it, so it can not be combined well with a taser or riot shield or wand, and the bullets only do 20 brute, less than even a stechkin. Most wizards would just rather a laser gun and charge, or an autorifile, or a shotgun the hos dropped. More importantly, it's 4 points for 20 shots using both hands, 400 total damage, with a cooldown of 75 seconds. In comparison, one could buy mutate, get eye lasers, 20 burn per shot, only one hand available needed, with the bonus of hulk for punches and stun immunity, active for 30 seconds and a cooldown of 40, so a downtime of *10 seconds*. In the current state, there is no point to buying lesser summon guns over mutate.

This PR makes the spell way more affordable, and a bit lower of a cooldown, but honestly, could potentially use a larger buff than this. 

## Changelog
:cl:
tweak: Lesser summon guns has a cooldown of 1 minute now, and costs 2 spellpoints instead of 4
/:cl:

### DIFF
--- a/code/datums/spells/infinite_guns.dm
+++ b/code/datums/spells/infinite_guns.dm
@@ -6,7 +6,7 @@
 	range = -1
 
 	school = "conjuration"
-	charge_max = 750
+	charge_max = 600
 	clothes_req = 1
 	cooldown_min = 10 //Gun wizard
 	action_icon_state = "bolt_action"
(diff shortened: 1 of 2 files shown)


SteelSlayer (approved):

variableundefined (approved):

Necaladun (approved):
Sure why the hell not, gunwizards are generally hilarious.

AffectedArc07 (approved):

PR #14822, opened 2020-11-04
Makes removing legcuffs toggle movement intent back to 'Run'

SabreML:
## What Does This PR Do

This makes resisting out of leg cuffs or being uncuffed toggle your move intent to 'Run'. ("Leg cuffs" being bolas, bear traps, etc.)
Previously your move intent would stay on 'Walk', slowing you down for a while after uncuffing if you're not paying attention.

I've made this into a different PR from #14821 entirely because I thought it needed to be separate and definitely not because @variableundefined is too fast for me to make a second commit. :eyes:

## Why It's Good For The Game

I know for a fact that walk intent staying on causes a lot of confusion, especially among newer players.
While this *could* be argued to be a bola nerf, I strongly disagree, since anyone with experience will be able to just hotkey
`toggle-walk-run` or something and avoid the issue entirely.

## Changelog
:cl:
tweak: Removing leg cuffs now changes your move intent back to 'Run'
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -649,6 +649,7 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, list(/obj/machinery/atmospherics/unary/ven
 		update_handcuffed()
 	else if(I == legcuffed)
 		legcuffed = null
+		toggle_move_intent()
 		update_inv_legcuffed()
 
 /mob/living/carbon/show_inv(mob/user)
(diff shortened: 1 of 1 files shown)


farie82 (approved):
So many ways to remove leg/hand cuffs... All doing the same code but copy pasted... Code will work but refactoring this will be a good thing to do as a bonus :D

SabreML:
> So many ways to remove leg/hand cuffs... All doing the same code but copy pasted... Code will work but refactoring this will be a good thing to do as a bonus :D

I might do that if blood cult gets merged before this, but right now I'm trying not to new projects.

variableundefined (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #14462, opened 2020-09-27
Prevent observe before server is finished starting

Kyep:
## What Does This PR Do
Prevents players (including admins) from using "Observe" before the server is finished starting.
Until the server is ready for observers, a "Observe (starting up...)" will be displayed instead of the normal blue "Observe" button.

NOTE: This _doesn't_ stop you using Observe several minutes before roundstart. It just stops you using it for the first 5-20 seconds or so after the server first comes online, while the server is still initializing, loading the map, etc.

## Why It's Good For The Game
1) Various misleading runtimes and other errors can happen if you leap into the round before the server has even finished initializing subsystems and things.
2) This is eventually going to be required in order to support map rotation.

## Images
TZUcUcszmh
(yes, its slightly bigger than the options list presented, this is to account for "View the Crew Manifest" appearing for latejoiners)

## Changelog
:cl: Kyep
add: Server no longer lets you Observe before its even finished loading the game.
/:cl:

PR #14824, opened 2020-11-05
Adds a config setting to enable observers before server initializations

SabreML:
## What Does This PR Do

Adds a developer config setting to allow observers before server initializations are completed.

## Why It's Good For The Game

While the checks added in #14462 are undoubtedly very useful on a live server, sometimes when testing something locally you want to be able to enter the game without waiting ~40 seconds for the server to load.
This *will* cause runtimes if enabled, but those aren't very important if you just want to check the colour of a sprite or something.

## Changelog
:cl:
add: Added a server config setting to allow observers before initializations are complete
/:cl:

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -241,6 +241,9 @@
 	//Start now warning
 	var/start_now_confirmation = 0
 
+	// Allow observers before initialization
+	var/observer_initialize = 0
+
 	// Lavaland
 	var/lavaland_budget = 60
 
@@ -727,6 +730,9 @@
 
 				if("start_now_confirmation")
 					config.start_now_confirmation = 1
+
+				if("observer_initialize")
+					config.observer_initialize = 1
 
 				if("tick_limit_mc_init")
 					config.tick_limit_mc_init = text2num(value)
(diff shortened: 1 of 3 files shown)


code/controllers/configuration.dm
```diff
@@ -241,6 +241,9 @@
 	//Start now warning
 	var/start_now_confirmation = 0
 
+	// Allow observers before initialization
+	var/observer_initialize = 0
```
SabreML, line 245: I realise that this and line 735 should be `FALSE` and `TRUE` since it's a boolean, but all the other variables in this file are `1`s and `0`s so it'll look a bit out of place if I change it. And I don't particularly want to go through 917 lines to change everything manually.

AffectedArc07:
This is intentional behaviour because when map rotation comes in, observing before the station even exists will break EVERYTHING, and it made sense to do at the time. If you want to speed up your initialisation time, disable space ruins and the gateway. That takes the time way down. 


Not in favour.

SabreML:
@AffectedArc07 I could change it to only allow observers after the map has loaded, it wouldn't be quite as fast but it'd still cut the time down by a good bit.
Although at that point I suppose there's the argument that you could just wait for it to finish, rather than changing a config setting.

AffectedArc07:
@SabreML Disabling the extra z-levels greatly helps with init times, and if youre not doing that already, I highly recommend doing it. It cuts the loading of SSmapping down the 0.2 seconds, and reduces not only the amount of maps, but the amount of atoms to be initialized (dont forget its only half of the turfs that need to load), as well as all the other objects, not to mention taking less time for lighting to initialize.

Testing features before the MC has fully loaded is a bad idea anyway. If youre profiling youll get fuzzed results, and its better to test after starting the game when the MC is in the midgame state, and all subsystems are firing at normal game rates, as opposed to lobby or init state.

variableundefined (commented):
Objection on the same ground as Affected, closing.

PR #14837, opened 2020-11-07
Makes fireaxes take time to pry open airlocks.

Qwertytoforty:
## What Does This PR Do

Fireaxes now take 7.5 seconds to open airlocks.

## Why It's Good For The Game

Fireaxes now take 7.5 seconds to open airlocks, instead of instantly prying, so they are no longer the magic all access tool, and makes them more similar to jaws of life.

## Changelog
:cl:
tweak: Axes now take 7.5 seconds to pry open doors.
/:cl:


Samman166:
Based(I agree with this nerf miners for god sakes)

Kyep (approved):

variableundefined (dismissed):

Kyep:
Originally I thought the title of this PR was an axe-edent.
But, it turns out you really have a handle on the problem.
This PR is no quick hack job, in fact I'd say its axe-ellent.
I look forward to seeing more axe of coding from you in future.

Fox-McCloud:
Fixing the logic so these aren't free unconditional door openers would probably be a better start: https://github.com/tgstation/tgstation/blob/840312474b616ed683b7980c83f4d26c6a5eff79/code/game/machinery/doors/airlock.dm#L1049

Qwertytoforty:
Credit to @Varghh for fixing the fireaxe code, so I could add it to my PR.

All fireaxes now take time to open doors, instead of instant prying, bone axe is now once again a subtype of a fire axe.

code/game/machinery/doors/airlock.dm
```diff
@@ -1109,13 +1109,16 @@ About the new airlock wires panel:
 	if(istype(I, /obj/item/twohanded/fireaxe)) //let's make this more specific //FUCK YOU
 		var/obj/item/twohanded/fireaxe/F = I
 		if(F.wielded)
-			spawn(0)
-				if(density)
-					open(1)
+			if(density)
+				var/time_to_open = 75
```
SabreML, line 1113: This could be moved to the `do_after()` proc, since it's not being used elsewhere.

code/game/machinery/doors/airlock.dm
```diff
@@ -1109,13 +1109,16 @@ About the new airlock wires panel:
...
+				playsound(src, 'sound/machines/airlock_alien_prying.ogg', 100, 1) //is it aliens or just the CE being a dick?
+				prying_so_hard = TRUE
+				if( do_after(user, time_to_open, target = src) )
```
SabreML, line 1116: Spacing

code/game/machinery/doors/airlock.dm
```diff
@@ -1109,13 +1109,16 @@ About the new airlock wires panel:
...
+					prying_so_hard = FALSE
+					if(density && !open(1))
```
SabreML, line 1118:
```suggestion
					if(density && !open(TRUE))
```

code/game/machinery/doors/airlock.dm
Qwertytoforty: done!

code/game/machinery/doors/airlock.dm
Qwertytoforty: Almost done

Varghh:
> Credit to @Varghh for fixing the fireaxe code, so I could add it to my PR.

I was feeling sharp today

code/game/machinery/doors/airlock.dm
```diff
@@ -1109,13 +1109,15 @@ About the new airlock wires panel:
...
-				if(density)
-					open(1)
+			if(density)
+				playsound(src, 'sound/machines/airlock_alien_prying.ogg', 100, 1) //is it aliens or just the CE being a dick?
+				prying_so_hard = TRUE
+				if(do_after(user, 75, target = src))
+					prying_so_hard = FALSE
+					if(density && !open(TRUE))
+						to_chat(user, "<span class='warning'>Despite your attempts, [src] refuses to open.</span>")
 				else
```
moxian, line 1119: Judging by the error message, this `else` corresponds to `if(F.wielded)` but the indentation makes it an error for failing `do_after()` (i.e. moving away after you start prying, and interrupting it)

code/game/machinery/doors/airlock.dm
```diff
@@ -1109,13 +1109,15 @@ About the new airlock wires panel:
...
 		var/obj/item/twohanded/fireaxe/F = I
 		if(F.wielded)
-			spawn(0)
-				if(density)
-					open(1)
+			if(density)
+				playsound(src, 'sound/machines/airlock_alien_prying.ogg', 100, 1) //is it aliens or just the CE being a dick?
+				prying_so_hard = TRUE
+				if(do_after(user, 75, target = src))
+					prying_so_hard = FALSE
```
moxian, line 1116: You need to reset `prying_so_hard = FALSE` even if `do_after` fails. Otherwise, interrupting the prying would make it unpriable...
.. but only by jaws of life, since you're not checking prying_so_hard here at all.

And now that I say that, this is effectively a copy-paste of jaws behavior several lines below. Let's not have copy-paste like that. Either see that the code falls through to the jaws when using fireaxe OR isolate this logic into a dedicated proc, and just call that proc in these two places.


Also add a comment on the variable what it means, because the name is atrocious. "Whether the door is currently being pried by someone" or similar.

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -245,6 +236,15 @@
 			var/atom/throw_target = get_edge_target_turf(M, get_dir(src, get_step_away(M, src)))
 			M.throw_at(throw_target, 5, 1)
 
+/obj/item/twohanded/fireaxe/boneaxe  // Blatant imitation of the fireaxe, but made out of bone.
```
moxian, line 239: You just moved it around, and now git blame would think you deleted the old one and added the new one.
Please move back to above energized.

moxian (changes requested):

code/game/objects/items/weapons/twohanded.dm
Qwertytoforty: Moved back

code/game/machinery/doors/airlock.dm
Qwertytoforty: else was moved

code/game/machinery/doors/airlock.dm
Qwertytoforty: Made the prying so hard work, kept it different as one uses the two handed.

PR #14840, opened 2020-11-07
In-code server revision info

AffectedArc07:
## What Does This PR Do
Gives the server code the ability to be aware of the commit hash the server is running on, as well as providing players a link to said hash if they want to try a specific build from a specific round locally. Pressing `Get Revision Info` in the OOC tab will provide you with this
image
*more info may be added here if it seems fit*

It also stamps into the loglines of `game.log`, `runtime.log` and `runtime_summary.log`
image

image

image

And yes the server still works outside a git environment
image
image
image
image

## Why It's Good For The Game
The server being aware of its commit hash and date is a very valuable tool for debugging testmerges, round-isolated incidents, amongst other things

## Changelog
:cl: AffectedArc07
add: The code now knows what commit its running at. Use "Get Revision Info" in the OOC tab to view the server code revision
add: [Administration] game.log and runtime.log now mark commit hash at the start  of the log to aid debugging
/:cl:

### DIFF
--- a/code/datums/revision.dm
+++ b/code/datums/revision.dm
@@ -1,0 +1,56 @@
+GLOBAL_DATUM_INIT(revision_info, /datum/code_revision, new)
+GLOBAL_PROTECT(revision_info) // Dont mess with this
+
+/**
+  * Code Revision Datum
+  *
+  * Allows the server code to be aware of the Git environment it is running in, and lets commit hash be viewed
+  */
+/datum/code_revision
+	/// Current commit hash the server is running
+	var/commit_hash
+	/// Date that this commit was made
+	var/commit_date
+
+/datum/code_revision/New()
+	commit_hash = rustg_git_revparse("HEAD")
+	if(commit_hash)
+		commit_date = rustg_git_commit_date(commit_hash)
+
+/**
+  * Code Revision Logging Helper
+  *
+  * Small proc to simplify logging all this stuff
+  */
+/datum/code_revision/proc/log_info()
+	// Put revision info in the world log
+	var/logmsg
+	if(commit_hash && commit_date)
+		logmsg = "Running ParaCode commit: [GLOB.revision_info.commit_hash] (Date: [GLOB.revision_info.commit_date])"
+	else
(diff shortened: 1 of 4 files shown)


code/datums/revision.dm
```diff
@@ -0,0 +1,56 @@
...
+/**
+  * Code Revision Logging Helper
+  *
+  * Small proc to simplify logging all this stuff
+  */
+/datum/code_revision/proc/log_info()
+	// Put revision info in the world log
+	var/logmsg
+	if(commit_hash && commit_date)
+		logmsg = "Running ParaCode commit: [GLOB.revision_info.commit_hash] (Date: [GLOB.revision_info.commit_date])"
```
moxian, line 29: Should this be 
```suggestion
		logmsg = "Running ParaCode commit: [commit_hash] (Date: [commit_date])"
```
given that this is an owned proc?

code/datums/revision.dm
```diff
@@ -0,0 +1,56 @@
...
+
+/client/verb/get_revision_info()
+	set name = "Get Revision Info"
+	set category = "OOC"
+	set desc = "Retrieve technical information about the server"
+
+	var/list/msg = list()
+	msg += "<span class='notice'><b>Server Revision Info</b></span>"
+	// Commit info first
+	if(GLOB.revision_info.commit_hash && GLOB.revision_info.commit_date)
```
moxian, line 46: `var/revision_info = GLOB.revision_info` to not carry `GLOB.` around with you?

dearmochi (approved):

code/datums/revision.dm
AffectedArc07: Done

code/datums/revision.dm
AffectedArc07: This works and isnt creating excess objects in a proc, and referencing it directly is cleaner in my opinion

moxian (approved):

farie82 (approved):

Kyep:
I need to actually do some local testing before I can review this. It may interfere with some other things. Unsure yet.

Kyep (approved):

PR #14846, opened 2020-11-08
Firelocks can be opened by hand in reasonable time

moxian:
## What Does This PR Do
Reduces the time it takes to open firelock by hand (without a crowbar) from 30 seconds down to 5.

## Why It's Good For The Game
Atmos breaches are fairly regular on the station. Whenever one happens, firelocks, naturally, engage. However they often cut you off from the fire panel, so you effectively need a crowbar to get yourself to where you want to be (and often, to safety).
However, needing to carry a crowbar on you at all times exactly for the reason of "what if i get bombed?" is.. unexciting, especially if you play a role that shouldn't *need* to carry a crowbar IC'ly.

Now, turns out, that it *is* possible to open firelocks by hand, but it takes an excuriating 30 seconds, which is just waaay too long. Lowering that time to something - still long, but reasonable, like 5 seconds (or maybe 10 even) would eliminate the need to carry crowbar "just in case", and (arguably!) higher the RP standards a tiniest bit.

Please note that 5 seconds is still way way longer than the "instant" that you get when applying the crowbar, so crowbar absolutely does not become useless item.

## Images of changes
20201108-025530-IhHankIt98

## Changelog
:cl:
tweak: Firelocks can be opened by hand in 5 seconds, down from 30
/:cl:

### DIFF
--- a/code/game/machinery/doors/firedoor.dm
+++ b/code/game/machinery/doors/firedoor.dm
@@ -19,11 +19,11 @@
 	safe = FALSE
 	layer = BELOW_OPEN_DOOR_LAYER
 	closingLayer = CLOSED_FIREDOOR_LAYER
-	auto_close_time = 50
+	auto_close_time = 5 SECONDS
 	assemblytype = /obj/structure/firelock_frame
 	armor = list("melee" = 30, "bullet" = 30, "laser" = 20, "energy" = 20, "bomb" = 10, "bio" = 100, "rad" = 100, "fire" = 95, "acid" = 70)
 	/// How long does opening by hand take, in deciseconds.
-	var/manual_open_time = 50
+	var/manual_open_time = 5 SECONDS
 	var/can_crush = TRUE
 	var/nextstate = null
 	/// Whether the "bolts" are "screwed". Used for deconstruction sequence. Has nothing to do with airlock bolting.
@@ -98,13 +98,13 @@
 	user.changeNext_move(CLICK_CD_MELEE)
 
 	user.visible_message(
-		"<span class='notice'>[user] tries to open \the [src] manually.</span>",
-		"<span class='notice'>You operate the manual lever on \the [src].</span>")
+		"<span class='notice'>[user] tries to open [src] manually.</span>",
+		"<span class='notice'>You operate the manual lever on [src].</span>")
 
 	if(do_after(user, manual_open_time, target = src))
 		user.visible_message(
-			"<span class='notice'>[user] opens \the [src].</span>",
-			"<span class='notice'>You open \the [src].</span>")
+			"<span class='notice'>[user] opens [src].</span>",
+			"<span class='notice'>You open [src].</span>")
 		open(auto_close = FALSE)
 
 /obj/machinery/door/firedoor/attackby(obj/item/C, mob/user, params)
--- a/code/game/machinery/doors/firedoor.dm
+++ b/code/game/machinery/doors/firedoor.dm
@@ -6,7 +6,7 @@
 
 /obj/machinery/door/firedoor
 	name = "firelock"
-	desc = "Apply crowbar."
+	desc = "A convenable firelock. Equipped with a manual lever for operating in case of emergency."
 	icon = 'icons/obj/doors/doorfireglass.dmi'
 	icon_state = "door_open"
 	opacity = 0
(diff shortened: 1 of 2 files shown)


Mitchs98:
I'll still carry a crowbar, and I don't think arguing for the PR in an IC sense is relevant. But this WILL be great for the rounds you spawn in to an arrivals breach and can't get a crowbar, or what have you.

Carthusia:
Perhaps 10-15 seconds, its still a long time, but nowhere near as bad.

You are trying to force open a hydraulically pressure sealed door.

There is also another PR right now to change the fireaxes opening speed to 7.5 seconds, i feel there is a bit of a conflict or inconsistency between this and that one, where opening a firelock with your hands would be far quicker than even using a fireaxe

moxian:
dreamseeker-20201108-233948

> You are trying to force open a hydraulically pressure sealed door.
I agree it's silly.
I was originally inspired to do this by bee, where the firelocks mention that they have a mechanical lever on them, but was too lazy to do this properly (and they are in general closer to what we use regular airlocks for I think).
Anyhow, added the flavour.

Also removed the need for non-help intent to force it open. The "if it's a glass airlock, then tap on the glass instead" bit is kinda nice, but it took me reading the code, and turning my volume all way up to realize the sound is indeed there - it's that quiet.

> There is also another PR right now to change the fireaxes opening speed to 7.5 seconds, i feel there is a bit of a conflict or inconsistency between this and that one, where opening a firelock with your hands would be far quicker than even using a fireaxe.
Airlocks and firelocks are different entities with different behavior. Not saying this is how it should be, but that's how the game is right now.
Perhaps firelocks' motors are weaker than airlocks'? After all you can crowbar firelocks instantly, but airlocks not at all...

> Perhaps also add a timer to crowbaring the airlocks as well? 
I'm not brave enough to make that kind of PR :stuck_out_tongue:

code/game/machinery/doors/firedoor.dm
```diff
@@ -298,7 +301,6 @@
 	opacity = 1
 	explosion_block = 2
 	assemblytype = /obj/structure/firelock_frame/heavy
-	can_force = FALSE
```
moxian, line 301: No longer makes flavour sense, given that we are no longer *literally* forcing the lock open by hand, but using the emergency fallback instead.

Carthusia:
Ooohh, i like the idea of it noting an emergency lever, makes MUCH more sense than trying to push the door up AND down at the same time with your bare hands

code/game/machinery/doors/firedoor.dm
```diff
@@ -22,10 +22,11 @@
 	auto_close_time = 50
 	assemblytype = /obj/structure/firelock_frame
 	armor = list("melee" = 30, "bullet" = 30, "laser" = 20, "energy" = 20, "bomb" = 10, "bio" = 100, "rad" = 100, "fire" = 95, "acid" = 70)
-	var/can_force = TRUE
-	var/force_open_time = 300
+	/// How long does opening by hand take, in deciseconds.
+	var/manual_open_time = 50
```
SabreML, line 26:
```suggestion
	var/manual_open_time = 5 SECONDS
```

SabreML (changes requested):

code/game/machinery/doors/firedoor.dm
```diff
@@ -89,20 +90,22 @@
...
-			user.visible_message("<span class='notice'>[user] forces \the [src].</span>", \
-								 "<span class='notice'>You force \the [src].</span>")
-			open()
-	else if(glass)
-		user.visible_message("<span class='warning'>[user] bangs on \the [src].</span>",
-							 "<span class='warning'>You bang on \the [src].</span>")
-		playsound(get_turf(src), 'sound/effects/glassknock.ogg', 10, 1)
+	user.visible_message(
+		"<span class='notice'>[user] tries to open \the [src] manually.</span>",
+		"<span class='notice'>You operate the manual lever on \the [src].</span>")
```
SabreML, line 102: You might not need the `\the`s here, `[src]` usually adds them automatically.

code/game/machinery/doors/firedoor.dm
```diff
@@ -89,20 +90,22 @@
...
+
+	if(do_after(user, manual_open_time, target = src))
+		user.visible_message(
+			"<span class='notice'>[user] opens \the [src].</span>",
+			"<span class='notice'>You open \the [src].</span>")
```
SabreML, line 107: Same here

SabreML (changes requested):

Kyep (approved):

AffectedArc07 (approved):

PR #14847, opened 2020-11-08
Disable nullified stacking

### LINKED ISSUE #14136: Prayer beads nullification stacks to insane values.
**Issue Description**:

Prayer beads nullification to vampires can be stacked multiple times in a row, preventing a vampire from using powers at all, effectively. 

Prayers beads to apply their effect, and if applied again, refill the nulification to the beads amount, but not go higher.

**What happened instead**:

You can stack the nullification, leading to vampires being nullified to 20 minutes or more.

You should not be able to make a vampire unable to use any abilities for 20+ minutes by using a null rod a few times, it should only nullify them for 4 minutes and have to be refreshed.

Get a vampire and chaplain.
Nullify with the beads, look at the nullify value.
Use the beads again, see how the value is added.

mokulus:
## What Does This PR Do
Fixes #14136 

## Why It's Good For The Game
Explained in issue

## Images of changes
na

## Changelog
:cl:
fix: fixed infinite stacking of nullification of vampire
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/holy_weapons.dm
+++ b/code/game/objects/items/weapons/holy_weapons.dm
@@ -465,7 +465,7 @@
 					return
 
 				if(target.mind.vampire && !target.mind.vampire.get_ability(/datum/vampire_passive/full)) // Getting a full prayer off on a vampire will interrupt their powers for a large duration.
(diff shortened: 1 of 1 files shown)


farie82:
There are quite some more instances where this logic is used. One in the same file even for the same object. Can you also fix those? Those will most likely require the `min` proc

PR #14850, opened 2020-11-08
world/Topic Refactor

AffectedArc07:
## What Does This PR Do
Refactors world/Topic. Key changes are:
- Topic types are now split into datumised entries for cleaner organisation and standardisation of auth checking
- Topic spam prevention is now much more sane
- All topics now return JSON apart from `?ping`, which is used by BYOND itself for stuff
- Removed some old topics we no longer use
- Split player list out of `?status` and into `?playerlist`

Key points about rate limiting are:
- Each IP gets its own rate limits
- If you topic multiple times within 0.5 seconds, you are given a strike
- If you get 5 strikes, you are blacklisted from topics for a minute
- If you attempt to topic while blacklisted, your lockout is reset back to the full minute
- Strikes will reset to 0 after your lockout period has ended

Example of much cleaner data (JSON):
image

## Why It's Good For The Game
Our current world/Topic implementation is quite frankly awful, non-standardised, a ton of bloat in `world.dm` and just generally not easily maintainable. This PR solves that.

## Changelog
:cl: AffectedArc07
add: Refactored world/Topic to use JSON replies
/:cl:


code/modules/world_topic/adminmsg.dm
```diff
@@ -0,0 +1,36 @@
+/datum/world_topic_handler/adminmsg
+	topic_key = "adminmsg"
+	requires_commskey = TRUE
+
+/datum/world_topic_handler/adminmsg/execute(list/input, key_valid)
+	/*
+	We got an adminmsg from the Discord bot, so lets split the input then validate the input. Expected output:
+		1. adminmsg = ckey of person the message is to
+		2. msg = contents of message, parems2list requires
```
moxian, line 9: Pre-existing code but.. it requires what?
Also should be `params2list`, I assume

code/game/world.dm
```diff
@@ -54,242 +54,30 @@ GLOBAL_LIST_INIT(map_transition_config, MAP_TRANSITION_CONFIG)
...
-//		if(href_list["hello"])
-//			to_chat(world, "Hello world!")
-//			return "Hello world!"
-//		to_chat(world, "End of Topic() call.")
-//		..()
-
-GLOBAL_VAR_INIT(world_topic_spam_protect_ip, "0.0.0.0")
-GLOBAL_VAR_INIT(world_topic_spam_protect_time, world.timeofday)
+/// List of all world topic spam prevention handlers. See code/modules/world_topic/_spam_prevention_handler.dm
+GLOBAL_LIST_EMPTY(world_topic_spam_prevention_handlers)
```
moxian, line 58: This is empty and unused?
How does one interact with it on the running server?

AffectedArc07: I spent so much time debugging this locally with a client verb that I never actually did the server implementation. Conclusion: I am actually stupid, and wondering how the fuck I missed this.

code/modules/world_topic/adminmsg.dm
AffectedArc07: Yeah, I'll fix that when I can

AffectedArc07:
Example of rate limiting in action:
image
This script:
```py
for x in range(10):
    sock.sendall(query)
    data = sock.recv(16384)
    decode_data() # Handles parsing and printing
```
First 5 attempts were successful, last 5 failed due to rate limits

code/modules/admin/sql_notes.dm
```diff
@@ -202,18 +202,3 @@
 		output += ruler
 	usr << browse(output, "window=show_notes;size=900x500")
 
-/proc/show_player_info_irc(var/key as text)
```
AffectedArc07, line 205: No longer required

code/modules/world_topic/adminmsg.dm
```diff
@@ -0,0 +1,36 @@
...
+	var/amessage =  "<font color='blue'>Discord PM from <a href='?discord_msg=1'>[input["sender"]]</a> to <b>[key_name(C)]</b>: [input["msg"]]</font>"
+
+	// THESE TWO VARS DO VERY DIFFERENT THINGS. DO NOT ATTEMPT TO COMBINE THEM
+	C.received_discord_pm = world.time
+	C.last_discord_pm_time = 0
+
+	SEND_SOUND(C, 'sound/effects/adminhelp.ogg')
+	to_chat(C, message)
+
+	for(var/client/A in GLOB.admins)
```
AffectedArc07, line 32: Need to test to make sure mentors arent in this list

code/modules/world_topic/status.dm
```diff
@@ -0,0 +1,53 @@
+/datum/world_topic_handler/status
+	topic_key = "status"
+
+/datum/world_topic_handler/status/execute(list/input, key_valid)
+	var/list/status_info = list()
+	var/list/admins = list()
+	status_info["version"] = GLOB.game_version
```
AffectedArc07, line 7: TODO: Make this the running git hash

PR #14856, opened 2020-11-09
TS Mother Rework

Kyep:
## What Does This PR Do & Why
* Boss spiders (Queen, Princess, Empress, Mother) now have the ability to see through the eyes of lesser terrors, much like the genetic "remote view" ability. This lets them lead the hive in a more meaningful way, since they can observe their minions and give orders. This encourages teamwork between spiders, and makes playing a boss spider (which mostly just sits around laying eggs) less boring. It also makes boss spiders much easier to kill! If you find one, the odds are substantial that it will be distracted using this ability, so you can sneak up to it and kill it. For Queens/Princesses, this also triggers the death of most of their hive. So, while this feature makes bosses more interesting to play, and opens up new options for smart leaders, it is also a huge weakness that crew can exploit to wreck entire hives.
* Spiderlings no longer continually attempt to path to welded vents.
* Mothers have been completely reworked. They are no longer suicide spiders that only exist to die. That wasn't fun to play. Now they have a totally new purpose. See below for details.
* Queens now have an option to lay a mother egg, instead of a prince or princess egg. This is the only way a Mother can enter the round currently (except adminbus). The idea is instead of choosing a Prince (single strong hero unit) or a Princess (overwhelm with numbers) she can instead choose a Mother (boost other spiders) as her T3 option. Mothers might be weaker than the other two options... but at least they can be spawned now, and provide some variety.
* Fixed terrors wrapping a spiderling into a cocoon if the spiderling was standing on top of a dead body and they wrap the dead body.

## How Mothers Work Now
* Overall, mothers act like a team medic, keeping existing spiders, and spiderlings, alive. All of their abilities serve this goal in one way or another. They're bad at combat.
* Mothers' primary ability is now creating royal jelly. They spend regeneration points to do this, and it creates a jelly that can be eaten like a corpse by other spiders to increase their regeneration points. Effectively, the mother is giving other spiders health packs that heal over time. This ability benefits from teamwork, as another spider giving a corpse to a mother, then eating the jellies the mother produces, is more efficient than the original spider just wrapping the corpse themselves.
* Mothers' webs block spiderlings from crossing them, so they can be placed strategically in doorways to stop spiderlings from wandering too far out of the nest and getting themselves killed. Or a curtain of them can be created which limits spiderlings to one small area with a vent.
* Mothers also have an ability that lets them pick up spiderlings, carrying them on their back until the spiderlings mature. This ability lets mothers protect spiderlings from attack during nest invasions, and helps relocate loose spiderlings to a new nest if the queen decides the time has come to move. 
* Mothers also have an "incubate eggs" ability which takes an existing egg (laid by the queen) and speeds up its progress towards hatching, at the cost of the mother's regeneration points. Since the egg would have hatched on its own anyway, and this ability costs points, it is really only worth it to do this as a last resort, if a nest is being overrun, and the mother is sacrificing its points in order to retrieve all the spiderlings she can just before the nest is evacuated.

## Images of changes
A mother's action bar (eye is remote review, purple blob is royal jelly):
mother_action_bar
A mother next to a royal jelly she has laid, for scale:
mother_next_to_royal_jelly
An animated gif of a jelly:
AMvszZR5yY
Royal jelly sprites are by FoS, thank you FoS!

## Changelog
:cl: Kyep
tweak: mothers of terror no longer spawn spiderlings on death. Instead, they've been reworked to now create royal jelly which, when eaten, helps other spiders regenerate health. They can also pick up spiderlings, incubate eggs, and create webs that block spiderling movement.
add: mothers/queens/princesses can now remotely watch other terrors. This lets them issue relevant orders to their spider underlings. It also lets crew sneak up to them and eliminate them if they are not properly guarded!
tweak: terror spiderlings no longer attempt to crawl through vents that are welded.
fix: fixed terrors wrapping a spiderling into a cocoon if the spiderling was standing on top of a dead body and they wrap the dead body.
/:cl:


code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -424,4 +433,30 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
...
 				stat(null, "Regeneration: [ltext]: <font color='[lcolor]'>[num2text(pc_of_max_per_second)]% of health per second</font>")
 
-
+/mob/living/simple_animal/hostile/poison/terror_spider/proc/DoRemoteView()
+	if(client && (client.eye != client.mob))
+		reset_perspective()
+		return
+	if(health != maxHealth)
+		to_chat(src, "<span class='warning'>You must be at full health to do this!</span>")
+		return
```
moxian, line 442: Huh?
So if the princess nest is raided, but she escapes with only a scratch - she can no longer provide her leadership to the hive? That sounds very weird..
What's the motivation here?

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -424,4 +433,30 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
...
+	targets += src // ensures that self is always at top of the list
+	for(var/thing in GLOB.ts_spiderlist)
+		var/mob/living/simple_animal/hostile/poison/terror_spider/T = thing
+		if(T.stat == DEAD)
+			continue
+		if(T.spider_awaymission != spider_awaymission)
+			continue
+		if(T == src)
+			continue
+		targets += T
```
moxian, line 453: I think if you do `targets |= T`, you won't need `if(T == src)` check.

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
```diff
@@ -424,4 +433,30 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
...
+	var/mob/living/L = input("Choose a terror to watch.", "Selection") in targets
```
moxian, line 454: I know I'm very spoiled by the goodies we have, but it'd be so cool if this was TGUI !

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
Kyep: Queens regenerate very fast. If she only takes a scratch, she'll have fully healed before she even reaches her new nest.
If she's severely wounded, then I think it makes sense for the pain/damage to block her viewing ability for a few minutes.
I also want to discourage wounded queens from using it again until they're in a nest and have had a chance to heal up.

code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
Kyep: Cool would be combining this and the 'hive sense' into one ability which gives you a window listing all spiders with a 'follow' button for each. Unfortunately, that is not in scope right now. This is still a draft and I'm not even sure this will be final design. So I don't really want to start re-coding the hivesense ability and TGUI-ing it all in this PR. I might in a subsequent PR though once this has been live for a bit and I'm sure the balance is fine.

TDSSS:
Mothers were in need of a rework, good change there. Onto the specifics:

- First, the jelly looks pretty similar to slimes to me, maybe that's just me.
- Is there any use for crew capturing jelly? If not, can they destroy it? It might be worth giving the jelly a use for crew, such as letting them extract healing chemicals from it.
- Does Mother have to wrap corpses to be able to lay any jelly at all, or does it speed it up? She'd be competing with greens for wraps so sometimes a Mother might just not get any corpses brought to her, maybe she should have some baseline jelly generation that's just boosted by corpses.
- If she doesn't have enough points to create a jelly, the mother sees `You need [jelly_cost] regeneration points to do this.` Can she actually see her current points anywhere? If not, this information is rather useless, isn't it?

Kyep:
@TDSSS answers:
- Yeah, jellies do look a bit like slimes, but the jelly sprites are animated, and sparkle a bit, so that helps to tell them apart. That and you rarely find slimes inside a terror nest. Spiders hate slimes. I've uploaded a GIF to the PR description so you can see the effect.
- Crew cannot currently capture or move jelly, but they can destroy it. It currently only has 15 HP currently, so one hit with a welder destroys it. I've been thinking about what happens if crew touches a jelly. Maybe they get the heebie jeebies? I don't really want to make it useful to crew, because that might discourage mothers from being picked, or mothers from creating it.
- Jelly-creation is fueled by regen points. Mother gets 1 regen point per second. All spiders also get regen points (200 for mother, 100 for other spiders) when they wrap a humanoid corpse. So no, mother does not have to wrap corpses to make jellys, but it does speed up the process quite a bit. 
- Mother won't be competing with greens for wraps. At least, if the greens are competent. When a green gives a corpse to a mother, she wraps it, getting 200 regen points as opposed to the 100 the green would have got. With 200 regen points, the mother can then create two jellies. The green can eat both - and each of them count as a corpse. The green and the mother have just used teamwork to get twice the value out of that corpse. If I think it might lead to queen being able to use mother+greens to spam eggs, though, I'll probably remove it. Mothers are meant to be support, not "extra eggs".
- Mothers can see their regen points in their status tab. That said, I've also now changed the message to "You only have [regen_points] of the [jelly_cost] regeneration points you need to do this.".

Mitchs98:
+1 for using the Eye of Cthulu from Terraria for the camera view button.

AffectedArc07:
> +1 for using the Eye of Cthulu from Terraria for the camera view button.

image
Not quite but close enough

Spacemanspark:
Isn't there a hardmode eye of cthulu mob that looks like that

Kyep:
The eye sprite was already in our codebase. I simply re-used it.
The only sprites actually added in this PR are the sprites for the jellies.

Spacemanspark:
I'm just pointing out the similarities :p

hal9000PR:
I don't see the mothers unique webs ever coming into effect because by the time a queen lays a mother all choke points will be filled with red,purple, queen and grey webs. Maybe if higher tier spiders had the ability to replace webs? But currently the mother will almost never have the opportunity to make use of their unique webs.

Kyep:
@hal9000PR You're partly right. In the initial nest, yes, other spiders will have webbed every turf. But part of the mother's purpose is to enable easier moves to new nests, and in a new nest its unlikely that every doorway will be webbed. That said, I do take your point that perhaps higher-tier spiders should be able to replace the webs of lower-tier ones. I've been considering implementing that for awhile and may yet do so in this PR, I just haven't decided yet if/how I want to do it.

Necaladun (approved):
like it all, i would like to see the suicide-bombing mothers return as some kinda NPC or something one day though

code/modules/mob/living/simple_animal/hostile/terror_spiders/mother.dm
```diff
@@ -2,56 +2,135 @@
...
+	if(statpanel("Status") && ckey && stat == CONSCIOUS)
+		stat(null, "Regeneration Points: [regen_points]")
+
+/mob/living/simple_animal/hostile/poison/terror_spider/mother/examine(mob/user)
+	. = ..()
+	if(!key || stat == DEAD)
+		return
+	if(!isobserver(user) && !isterrorspider(user))
+		return
+	. += "<span class='notice'>[p_they(TRUE)] is carrying [contents.len] spiderlings.</span>"
```
AffectedArc07, line 67:
```suggestion
	. += "<span class='notice'>[p_they(TRUE)] is carrying [length(contents)] spiderlings.</span>"
```

AffectedArc07 (changes requested):

AffectedArc07 (approved):
What could go wrong

PR #14860, opened 2020-11-09
Clarifies CC announcements during cult ascension

Kyep:
## What Does This PR Do
Changes the CC messages sent when cult gets halos, and when they ascend, to make a few things clearer:
- When cult gets halos, sec can lethal them as needed. Non-sec can also lethal them as a last resort (self-defense, defense of their workplace, etc). But even when cult has halos, non-sec crew should NOT be trying to hunt them down (partly as its validhunting, partly as there's no way your average civilian will win against a big group of armed cultists).
- When cult goes for the summon-your-god rune, EVERYONE on station is obliged to kill them on sight. Space Law and SOP are suspended until the crisis is dealt with. 

## Why It's Good For The Game
We've been getting some confusion, even among admins, about exactly how far sec and non-sec can go hunting cultists in the various stages of cult.

## Changelog
:cl: Kyep
tweak: Tweaked the CC messages sent when cult is growing in power, to make it clearer when lethal force is or is not authorized.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult.dm
+++ b/code/game/gamemodes/cult/cult.dm
@@ -226,7 +226,7 @@ GLOBAL_LIST_EMPTY(all_cults)
 			SEND_SOUND(M.current, 'sound/hallucinations/im_here1.ogg')
 			to_chat(M.current, "<span class='cultlarge'>Your cult is ascendant and the red harvest approaches - you cannot hide your true nature for much longer!")
 			addtimer(CALLBACK(src, .proc/ascend, M.current), 20 SECONDS)
-		GLOB.command_announcement.Announce("Picking up extradimensional activity related to the Cult of [SSticker.cultdat ? SSticker.cultdat.entity_name : "Nar'Sie"] from your station. Data suggests that about [ascend_percent * 100]% of the station has been converted. Security staff is authorised lethal force on confirmed cultists to contain the threat. Ensure dead crewmembers are revived and deconverted once the situation is under control.", "Central Command Higher Dimensional Affairs", 'sound/AI/commandreport.ogg')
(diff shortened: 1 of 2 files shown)


Carthusia:
Phew, sounds good.

The otherday we had a halo'd cult around, and a group of them got into the bar, took one person, and the halo'd CMO tried to get me, I chef CQCd them to the ground, and the bar more or less lynched them until they were out cold, and then died of their wounds not long after.

I was unsure if this was acceptable in hindsight, as under normal circumstances self defence ends when their on the ground and you gota avoid killing them.

With the halo's though, there were no ifs, ands or buts about it in the bar at that moment, if they didnt die after we got them on the ground, they would just get back up, and try to kill us or capture us with spooky cult magic, guns or all their big evil swords they kept pulling out like a magic infused zombie.

Necaladun:
@Carthusia Exactly the kind of confusion we want cleared up. Good to hear.

If you (or anyone reading this) can think of better wording, please hit us with it.

SabreML:
I'm definitely in favour of clarifying the announcement, but I'm worried that if it's too long people might not read it properly and take it as 'Cultists are valid' anyway.

I've not tested how it looks in-game though, so it might be fine.

McRamon:
I feel like this new announcement is very non immersive and OOCish. I am not a great writer myself, so i cant come up with best solution.
Maybe devide announcement into two parts, one being IC "we are detecting huge problem" kind of announcement, and second part be that one marked as "ooc" explaining who should do what.
Or maybe find a way to combine them into immersive yet informative announcement, maybe something like "blah blah blah continuum fluctuations, blah blah blah imminent threat to station and its personnel, lethal force fully authorized, civillians must follow any given orders and advised to stay in groups and get ready to defend themselves" something like that. Just the word "cultist" in announcement really breaks immersion as it is imho not a word officials would use

Dumbdumn5 (approved):
I n d e e d :+1:

AffectedArc07 (approved):

PR #14865, opened 2020-11-10
Adds 2 new blob reagents, radioactive gel, and teslium paste.

Qwertytoforty:
## What Does This PR Do

This PR adds 2 new blob reagents, radioactive gel, a blob type that does low toxin and very low brute damage, but injects radium to irradiate the user, and teslium paste, a blob type that adds teslium to the user, with a chance for a shocking outcome. (10% on blob attack, ~3.2% on blob spore burst, will not shock unless the user has teslium in their body.

## Why It's Good For The Game

More blob types to spice up the variety of blobs you fight are good.

The first blob, radioactive gel, is a lighter toxin blob then envenomed filiments, unless you get hit a few times in a row, where radiation is going to start building up on the user, making them start to take brute and burn over time, unless they have some radiation meds. Perfect against miners floating in space, but not effective against crew rushing you with welders early on, or plasmamen or IPC.

The second blob type, teslium paste, is not too dangerous if you are hit once, but the more you are hit, the more likely you will experience an electrical discharge, stunning you. Great against welder rushers early on, if you have the points to attack them directly, due to the change to shock. Not the best however if the target is in a group far away, where they will just be shaken up, and burn damage can be healed easily, with the lack of bone breakage. No need to worry about one hit stun, the blob can not shock you unless you already have teslium in your system, so long as you wait between hits, you will not be stuned directly from the blob. Teslium has a separate slight chance to stun you.

##Note:
This PR is work in progress, and 100% could use some adjustments before finalizing the stats of their base values, and the amount of reagents injected could use some work, as well as the strength of the electrocute act of the blob overcharge. A test merge, to see the reagents in action, could also potentially be good. Feel free to comment with how the reagents could be changed, if you think they could use an adjustment.

## Images of changes

Teslium paste
image

Radioactive gel
image

Radioactive gel is that color as radiation is often blue, teslium paste is that colour as that is the colour of teslium, and blue sorta looks like teslium, but frankly it could use a better color, *maybe* yellow for electricity? not sure.

## Changelog
:cl:
add: Added 2 new blob reagents, Radioactive gel, and Teslium paste.
/:cl:


code/modules/reagents/chemistry/reagents/blob.dm
```diff
@@ -152,6 +152,40 @@
...
+	complementary_color = "#24f0f0"
+	message_living = ", and you feel a strange warmth from within"
+
+/datum/reagent/blob/radioactive_gel/reaction_mob(mob/living/M, method=REAGENT_TOUCH, volume)
+	if(method == REAGENT_TOUCH)
+		volume = ..()
+		M.apply_damage(0.4*volume, TOX)
+		M.apply_damage(0.2*volume, BRUTE) // lets not have IPC / plasmaman only take 7.5 damage from this
+		if(M.reagents)
+			M.reagents.add_reagent("radium", 0.3*volume)
```
SabreML, line 169:
```suggestion
		M.apply_damage(0.4 * volume, TOX)
		M.apply_damage(0.2 * volume, BRUTE) // lets not have IPC / plasmaman only take 7.5 damage from this
		if(M.reagents)
			M.reagents.add_reagent("radium", 0.3 * volume)
```

code/modules/reagents/chemistry/reagents/blob.dm
```diff
@@ -152,6 +152,40 @@
...
 
+/datum/reagent/blob/radioactive_gel
+	name = "Radioactive gel"
+	description = "Deals medium toxin damage and a little brute damage, but irradiates those struck."
+	id = "radioactive_gel"
+	color = "#2476f0"
+	complementary_color = "#24f0f0"
+	message_living = ", and you feel a strange warmth from within"
+
+/datum/reagent/blob/radioactive_gel/reaction_mob(mob/living/M, method=REAGENT_TOUCH, volume)
```
SabreML, line 163: Spacing, here and a few below.
```suggestion
/datum/reagent/blob/radioactive_gel/reaction_mob(mob/living/M, method = REAGENT_TOUCH, volume)
```

code/modules/reagents/chemistry/reagents/blob.dm
```diff
@@ -152,6 +152,40 @@
...
+	description = "Deals medium burn damage, and shocks those struck over time"
+	id = "teslium_paste"
+	color = "#20324D"
+	complementary_color = "#412968"
+	message_living = ", and you feel a static shock"
+
+/datum/reagent/blob/teslium_paste/reaction_mob(mob/living/M, method=REAGENT_TOUCH, volume)
+	if(method == REAGENT_TOUCH)
+		volume = ..()
+		M.apply_damage(0.4*volume, BURN)
```
SabreML, line 182:
```suggestion
		M.apply_damage(0.4 * volume, BURN)
```

code/modules/reagents/chemistry/reagents/blob.dm
```diff
@@ -152,6 +152,40 @@
...
+		if(M.reagents)
+			if(M.reagents.has_reagent("teslium") && prob(0.6 * volume))
+				M.electrocute_act((0.75 * volume), "the blob's electrical discharge", 1, TRUE)
+				M.reagents.del_reagent("teslium")
+				return //don't add more teslium after you shock it out of someone.
+			M.reagents.add_reagent("teslium", 0.15*volume)  // a little goes a long way
```
SabreML, line 188:
```suggestion
			M.reagents.add_reagent("teslium", 0.15 * volume)  // a little goes a long way
```

procdrone:
not a fan of teslium. stun blob sounds like a terrible idea.

Kyep:
This really needs some testmerges. Work with me to figure out a time we're both available.

Necaladun (changes requested):
The stun thing makes me really hesitant. I would need a TM showing it's not OP to consider it before I risk it.

I love an electricity blob in theory a lot for sure, but stun is scary. If you can think of a way of doing that without the stun then that'd be great.

In general more blob variety good though.

Qwertytoforty:
@Necaladun It will be hard to make an electricity blob without stunning from shocks, as shocks inherently stun, but I will see what I can do. This is why it has only a chance to stun, but I can see how it can potentially be problematic. Currently, 15% chance to stun from a normal blob attack _ so long as_ they have teslium in there system. However, the random stun of teslium could be problematic too. Hmm.

Dumbdumn5 (approved):
Let the zappening **_begin_**. :+1:

Kyep (approved):

Issue #13772, opened 2020-07-05
Unlimited Range Pointing


Drobe50:
**Issue Description**: Infinite Range Pointing

**What did you expect to happen**: Not being able to point through the entertainment console.

**What happened instead**: You can point from any range with the Librarians camera and a entertainment monitor.

**Why is this bad/What are the consequences**: You can point anywhere you aren't. 

**Steps to reproduce the problem**: Turn on librarians camera, go to entertainment monitor, point anywhere on screen.

**When did the problem start happening**: Unknown

**Extra information**: N/A

stylemistake:
Was anyone able to see you pointing through cameras? If not, I don't see the issue.

Drobe50:
> 
> 
> Was anyone able to see you pointing through cameras? If not, I don't see the issue.

Actually they can, it's not just a "Clientside" thing. I forgot to mention that.

stylemistake:
I don't believe you, post video proofs from a bystanders perspective, preferably at very long range, and with obstruction between the camera user and the observer.

Why I think this isn't an issue - visibility rules still apply. If you can't see the person pointing at you, you don't see the pointing.

stylemistake:
Ok, I retract my previous statement, I thought tgui camera consoles were merged here. :smiley:

Contrabang:
I believe this has been fixed because I am not able to reproduce it with any of the current cameras, you're only able to point to areas in your current vision through cameras. However, I am not sure what PR fixed this possibly #19072? Leaving this open for a while to see if I can get any confirmations on this

PR #14879, opened 2020-11-12
Makes pointing only visible to those who see the pointee

farie82:
## What Does This PR Do
Makes the pointing arrow only visible to those who can see the one who points.

Fixes: #15299
Fixes: #13772

## Why It's Good For The Game
Avoids meta gaming where you see people point at walls from the other side.
It also helps antags with thermals/XRAY to communicate better through walls. Since they can actually point to things of notice without the thing of notice seeing them point at them.

How it looks:
dVMcgXZGgG
Me pointing a bit around. At the start I'm pointing at the walls and inside. While the door is closed I point at the door. All not visible

## Changelog
:cl:
tweak: Pointing now only shows the arrow to those who can see the one who points
fix: Can't scare people over long distances anymore by pointing at them through camera consoles
/:cl:


Fox-McCloud:
This *will* solve that, but I'm not sure how important this is, and it'll have some negative consequences as well that well.

While there's certainly the "realism" aspect of things, having a mild break in this, for the sake of ease of communication expedience, I can see would be useful.

The amount of times someone has gotten my attention, because of point, that I couldn't initially see, is fairly high.

While this does have some meta impacts, like what you mention, these are also fairly easy to avoid once the player is equipped with the knowledge that pointing behaves like it does.

For that reason, I do favor it behaving as it does---I'd rather trade off a *bit* of immersion breaks and have a *tiny* bit of meta, in favor of allowing players to communicate and get each other's attention a bit easier/better.

farie82:
> While there's certainly the "realism" aspect of things, having a mild break in this, for the sake of ease of communication expedience, I can see would be useful.

As long as you can see the one pointing you can see the arrow. As sec you can point at somebody who is moving to "notify them" That you need them. Only difference is that if you point out of sight of the person you're trying to get.
Any other time you'll want to use the point functionality then the other person should also see you.

> While this does have some meta impacts, like what you mention, these are also fairly easy to avoid once the player is equipped with the knowledge that pointing behaves like it does.

My worst issue with this is that people use it to point at walls while they are muted or such. Unable to scream. Yet are able to point to the security officer behind said wall who has to awkwardly ignore it else he might also get accused of meta gaming.

All in all I don't see any issues with changing it to this way. People trying to point stuff out still can do that. People with xray will also see the pointing through walls as expected.

code/modules/mob/mob.dm
```diff
@@ -610,11 +610,19 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
...
+
+	var/image/I = image('icons/mob/screen_gen.dmi', tile, "arrow", POINT_LAYER)
+	I.pixel_x = A.pixel_x
+	I.pixel_y = A.pixel_y
+	I.invisibility = invisibility
+	I.plane = GAME_PLANE
+
+	var/list/viewers = list()
+	for(var/mob/M in range(src))
+		if(M.client && (src in view(M)))
```
SteelSlayer, line 622: Don't you need to specify the tile distance for the `range` and `view` procs? I admit I don't actually know what happens if you use them like this.

farie82: The documentation is a bit odd. But I believe using it like this will use the view range of the given object. http://www.byond.com/docs/ref/#/proc/view has an example of this.

SteelSlayer: Ah alright, and so `range` apparently does the same thing but visibility is ignored according to the docs. I guess we can assume this is the full screen's worth of range.

farie82: Aye that is what I assume atleast. I'll test it a bit more around the edge cases to see if this is exactly what it does

KasparoVy:
> 
> 
> My worst issue with this is that people use it to point at walls while they are muted or such. Unable to scream. Yet are able to point to the security officer behind said wall who has to awkwardly ignore it else he might also get accused of meta gaming.

Found this out myself lol. I don't play too much at the moment but when I do it's usually blind challenge runs and found I could point at some crazy stuff or get information about things I otherwise shouldn't

If I need to point to something behind a wall I'll point to the wall and like, gesture 'over' or 'past' or say it if I can speak

code/modules/mob/mob.dm
farie82: Tested it a bit more. It indeed only works if you can see the one pointing and can see the thing being pointed at.

Kyep (commented):
I'm neutral on this.
While it does make physical sense, I do worry that it might have a negative impact on player interaction. E.g: blind person running around medbay never stops trying to move because he doesn't realize everyone is pointing at him.

farie82:
> I'm neutral on this.
> While it does make physical sense, I do worry that it might have a negative impact on player interaction. E.g: blind person running around medbay never stops trying to move because he doesn't realize everyone is pointing at him.

image
Still works since you can technically see it

Spacemanspark:
Why would a blind person looking for treatment be bumbling around medbay while people are chasing them to help?  You don't need to point at them, just walk up to them and start dragging them, they'll likely sit back and wait.
If not and they cause a commotion, that's their problem, and if it's bad enough you call security. I don't see why pointing being seen or not seen (which it is still seen, as Farie demonstrates) is an issue in this instance.

Good change imo.

Necaladun:
I strongly default to pointing being accurate to how it should be IRL - eg, seeing an actual person pointing, not manifesting a red arrow above an item.

But Fox makes a good and interesting point about the accepted nature of it being a game and all. A lot of games have a function like this, to get attention from other players, whether it's pinging the map or whatever.

In the end I think the communication aspect of it is worth the 1 in 10000 situations where it actually affects a situation

variableundefined (approved):
Make sense to me. Approved.

farie82:
I can add some more personal anecdotes of me having to deal with the pointing being visible to all.
As vampire in maint (doesn't matter where happened in multiple spots for me) your victim starts to point at the walls when he hears or sees people passing by. Costing you the victim and if you're just starting maybe also your round. Shit for all parties involved since people have a hard time ignoring obvious signs if they are there.
5FgIgGzXAb

Pointing to somebody just off screen usually doesn't work for me. I have to scream and point at them when they are in reach for them to notice me. Since then they know who is pointing at them and thus they stop. For me this little advantage of the current system is not enough to warrant the issues it brings with it of people using it to meta game and people being forced to know where to look.

AffectedArc07:
2 maints and a head against. Closing.

PR #14880, opened 2020-11-12
Swaps Gatfruit with Red Glowcaps @ Lavaland Terrarium

matttheficus:
## What Does This PR Do
Removes Gatfruit from the loot table of the Terrarium and adds Red Glowcaps (a powerful RND component, allowing for Power 6/7) in its place.

## Why It's Good For The Game
Infinite .357s that you can shove into a plant bag are REALLY dumb, and frankly shouldnt be something a traitor can just find for free, ESPECIALLY as a miner who already has a TON of other dumb and overpowered items at their disposal anyway. Botany, frankly, doesnt need it either as they can commit chemical warfare anyway which is a whole can of beans i dont even want to complain about here. Swapping it with Red Glowcaps allows for Science on station to possibly benefit too in the event that botany is new/MIA.

## Changelog
:cl:
tweak: replaces gatfruit (free guns) with red glowcaps (RND research tech) at the Lavaland Terrarium
/:cl:

### DIFF
--- a/code/modules/ruins/lavalandruin_code/seed_vault.dm
+++ b/code/modules/ruins/lavalandruin_code/seed_vault.dm
@@ -2,7 +2,7 @@
 	name = "seed vault seeds"
 	lootcount = 1
 
-	loot = list(/obj/item/seeds/gatfruit = 10,
+	loot = list(/obj/item/reagent_containers/food/snacks/grown/mushroom/glowshroom/glowcap = 10,
 				/obj/item/seeds/cherry/bomb = 10,
 				/obj/item/seeds/berry/glow = 10,
 				/obj/item/seeds/sunflower/moonflower = 8


Fox-McCloud:
This is the balance change I prefer to make, as it limits scope to the most abusive and common case.

Kyep (approved):

GentleManners:
Well replacing gat with useless glow caps is one way to do it. Though I thought Maints were against just straight up removing gat like it was said in a previous PR? Isn't this basically the same

matttheficus:
> Well replacing gat with useless glow caps is one way to do it. Though I thought Maints were against just straight up removing gat like it was said in a previous PR? Isn't this basically the same

Red Glowcaps are likely one of the MOST useful plants for the station, being the fastest upgrade path to Power 6/7 for RND. This could make them an interesting barter tool, which in my opinion is one of the more fun ways of interacting with the diona as Crew. They do, however, lack and active use I suppose for the terrarium residents. It was mentioned in #coder-chat that it could be swapped to Strange Seeds instead, allowing for interesting plant traits for the diona instead of a "barter good".

GentleManners:
> > Well replacing gat with useless glow caps is one way to do it. Though I thought Maints were against just straight up removing gat like it was said in a previous PR? Isn't this basically the same
> 
> Red Glowcaps are likely one of the MOST useful plants for the station, being the fastest upgrade path to Power 6/7 for RND. This could make them an interesting barter tool, which in my opinion is one of the more fun ways of interacting with the diona as Crew. They do, however, lack and active use I suppose for the terrarium residents. It was mentioned in #coder-chat that it could be swapped to Strange Seeds instead, allowing for interesting plant traits for the diona instead of a "barter good".

As someone who played both botany and science, I can confidently say that glowcaps aren't hard to get at all and don't have much value. If you have the glowshrooms and a dispenser it is a 50/50 to get it in like 30 seconds. I understand your point and would say it is a good idea but it's not nearly as valuable as you seem to think it is. (I have broken into botany before just to hack the vendor and get glowcaps in less than a minute before). It really isn't something you can bargain with. Also strange seeds are already present in the vendor (2  of them)

GentleManners:
Though do not mistake my intent, I am not against removing Gat but I think that the way it's done here is sub optimal in my opinion

Spacemanspark:
What would you suggest replacing it with

AffectedArc07 (approved):

PR #14882, opened 2020-11-12
Fixes revanents not respawning properly

Qwertytoforty:
## What Does This PR Do

This PR updates revenant respawn code to use timer / procs instead of spawn, replaces some ones and zeros with TRUES and FALSE, and most importantly, makes revenant dust able to respawn revenants after a minute.

## Why It's Good For The Game

Right, this is a bug fix, but should maybe be considered balance, due to the implications of fixing it. If this PR is merged, revenants will be able to respawn after a minute, if their dust is not touched. This may seem like a large rev buff, but consider the following.

Revenants either die early, or build up a lot of essence and don't get close to dying. Rev players that do really well, are unaffected by these changes, as they will not die. This helps newer players, or revs that make a mistake early on, have a chance to be revived after the bug is fixed. It is also _not hard_  to stop the revival of a revenant. Picking it up and throwing it, picking it up and using it in hand, or picking it up and... eating it... will stop the revival. Anyone that kills a revenent will generally be between 
1 - 4 tiles from it, and will be able to pick up the dust. This only really will revive the rev if crew forgets, rev is killed by a simple mob, or is killed in space outside a window.

## Changelog
:cl:
fix: fixed revenants not respawning from their dust after a minute.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/revenant/revenant.dm
+++ b/code/game/gamemodes/miniantags/revenant/revenant.dm
@@ -355,7 +355,7 @@
 	addtimer(CALLBACK(src, .proc/attempt_revive), 1 MINUTES)
 
 /obj/item/ectoplasm/revenant/proc/attempt_revive()
-	if(src && reforming)
+	if(reforming)
 		reform()
 	else
 		inert = TRUE
--- a/code/game/gamemodes/miniantags/revenant/revenant.dm
+++ b/code/game/gamemodes/miniantags/revenant/revenant.dm
@@ -212,6 +212,7 @@
 	var/reforming_essence = essence_regen_cap //retain the gained essence capacity
 	R.essence = max(reforming_essence - 15 * perfectsouls, 75) //minus any perfect souls
 	R.client_to_revive = src.client //If the essence reforms, the old revenant is put back in the body
+	R.reforming = TRUE
 	ghostize()
 	qdel(src)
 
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/miniantags/revenant/revenant.dm
```diff
@@ -344,20 +345,21 @@
...
-	spawn(600) //1 minutes
-		if(src && reforming)
-			reform()
-		else
-			inert = 1
-			visible_message("<span class='warning'>[src] settles down and seems lifeless.</span>")
+	addtimer(CALLBACK(src, .proc/attempt_revive), 1 MINUTES)
+
+/obj/item/ectoplasm/revenant/proc/attempt_revive()
+	if(src && reforming)
```
farie82, line 358: No need for the `src` check here. It was used in the spawn since spawns can have their `src` object be nulled.
A callback using a timer can never have a `src` that is null. Else the timer just won't go off

farie82 (changes requested):

code/game/gamemodes/miniantags/revenant/revenant.dm
Qwertytoforty: Done!

farie82 (approved):

Kyep (commented):
Objection.
I don't like the idea of self-reviving characters in general.
Revenants in particular are extremely annoying and add little to the round. A change that makes them harder to permanently kill? No thanks.
This isn't a forever-objection. If they were reworked to be more interesting RP wise, I'd consider not objecting to this change to make them harder to kill. But in their current state? No, they should most definitely not become harder to kill.

Necaladun:
Revs are pretty well hated as it is and hard enough to kill if they're smart. I appreciate the fix, but I don't want any form of buff to revs at all.

As Kyet said - if they get fixed in other ways I think it'd be cool.

AffectedArc07:

PR #14883, opened 2020-11-13
Clarifies LAZYLEN

SabreML:
## What Does This PR Do

~~Replaces all instances of `LAZYLEN` with `length`, since `LAZYLEN` has literally no reason to exist.~~

**Update:**
Adds a comment to the `LAZYLEN` define in order to hopefully reduce confusion.

## Why It's Good For The Game

This define consistently causes confusion about its use, so at the very least it should have some sort of explanation in the code so that some idiot doesn't try to remove it. :eyes:
As it is right now the use of it is one of those 'Unsaid Rules' which isn't actually explained anywhere.

### DIFF
--- a/code/__DEFINES/subsystems.dm
+++ b/code/__DEFINES/subsystems.dm
@@ -134,15 +134,15 @@
 		var/list/ad = A.add_overlays;\
 		var/list/rm = A.remove_overlays;\
 		var/list/po = A.priority_overlays;\
-		if(length(rm)){\
+		if(LAZYLEN(rm)){\
 			A.overlays -= rm;\
 			rm.Cut();\
 		}\
-		if(length(ad)){\
+		if(LAZYLEN(ad)){\
 			A.overlays |= ad;\
 			ad.Cut();\
 		}\
-		if(length(po)){\
+		if(LAZYLEN(po)){\
 			A.overlays |= po;\
 		}\
 		A.flags_2 &= ~OVERLAY_QUEUED_2;\
--- a/code/__DEFINES/subsystems.dm
+++ b/code/__DEFINES/subsystems.dm
@@ -134,15 +134,15 @@
 		var/list/ad = A.add_overlays;\
 		var/list/rm = A.remove_overlays;\
 		var/list/po = A.priority_overlays;\
-		if(LAZYLEN(rm)){\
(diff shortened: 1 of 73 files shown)


Fox-McCloud:
No. This isn't the proper way to handle this.

`LAZYLEN` is used when we're dealing with lazylists--it's to inform the coder that the list is inherently different, compared to most others.

Swapping it out for `.len` or `length` destroys this information, which makes this a very bad idea.

AffectedArc07:
image
Defines **dont** reduce performance. That's not how this works at all.

I also echo what fox said. You use any of the LAZY defines in **ANY** update or access or anything else to a lazy list to make it very obvious to the coder that's looking at it that it's a lazy list.

This is also a massive diff which is bound to throw conflicts everywhere for no gain. This define is intentionally like this.


TLDR: Not in favour at all.

SabreML:
@Fox-McCloud @AffectedArc07 
Considering this got two maintainer objections within a few hours, I've changed the PR in order to just add a comment to the define.

<hr>
> Defines dont reduce performance. That's not how this works at all.

I assumed that 1.5M consistent tests was enough to confirm it, my mistake.

AffectedArc07:
> I assumed that 1.5M consistent tests was enough to confirm it, my mistake.
1: I dont appreciate this level of snarkery at 9AM in the morning
2: The 1.5 million consistent tests actually boils down to *3* consistent tests, within the same minute
image
Using this code
```dm
#define LAZYLEN(L) length(L)

/proc/list_lazylen(list/L)
    return LAZYLEN(L)

/proc/list_length(list/L)
    return length(L)

/proc/main()
    var/list/L = list(1,2,3,5,6,7,8,9,10)
    BENCHT("lazylen", 500000, list_lazylen(L))
    BENCHT("length", 500000, list_length(L))
```

Given how performance can and *does* change over time depending on external factors IE: CPU load since theres a **lot** more happening on the server than this bot, not to mention results within 3ms of eachother with each iteration 500k iteration, it is safe to declare these results within margin of error.  

Anyway, I digress. Objection removed since now this has a helpful comment to avoid people doing the same in the future.

AffectedArc07 (approved):
At some point I want these properly documented for the autodoc system, but this will do for now

PR #14887, opened 2020-11-13
Made the traitor botanist seed less of a joke

SoThisIsItNow:
## What Does This PR Do
Makes the traitor seed Item more worthwhile

## Why It's Good For The Game
Seeds and botanists, there is no more iconic duo than this! So it is that the botanist specific item is a bit of a.... joke. It adds nothing new that a botanists cant already get and while cheap (2 TC) I feel like it should do a bit more so that botanists are encouraged to actually buy it and not be worse than a plant that a botanist can already grow naturally (Looking at you Omega Weeds).

 For that end I upped the amount of bath salts in it (if you manage to actually  get someone to somehow smoke it they will get a lot of it but probably not enough to OD), removed all the medicine out of it (It is supposed to be a lethal plant made by the syndicate, not heal your target.If you really want it you can get the same amount from the normal ambrosia vulgaris plant), added some lsd (It´s neither addicting or lethal but it will get you drugged, less LSD than rainbow weed by a fair bit) and atropine (It was removed from Omega so that botanists  had a harder time making Initropidril but it´s a traitor so I think them having an easier time to make it is fine).

I do not think that botany as a whole should get buffed but a traitor botanists should be encouraged to buy his job specificc  item. People seemed to agree that the Item is a joke and needed a buff to be an option.  So what do you get for 2 TC with this change? Basically a lot of bath salts and atropine. I do not think that what you get for the price is strong, as it is something you need to invest time into (Growing or taking the trait and putting it into another plant) and can already get pretty reliably. I considered just giving the plant just straight up Initropidril but settled on Atropine so that it takes a chunk of work to get it (It´s also at only 5%, maybe increasing it would be fine?).

Numbers are subject to change if people have objections.

## Changelog
:cl:
add: Added LSD 10%,Atropine 5% and ADDITIONAL 15% bath salts to ambrosia cruciatus
del: Kelotane and Bicardini from ambrosia cruciatus

/:cl:

### DIFF
--- a/code/modules/hydroponics/grown/ambrosia.dm
+++ b/code/modules/hydroponics/grown/ambrosia.dm
@@ -87,7 +87,7 @@
 	product = /obj/item/reagent_containers/food/snacks/grown/ambrosia/cruciatus
 	potency = 10
 	mutatelist = list()
-	reagents_add = list("thc" = 0.15, "kelotane" = 0.15, "bicaridine" = 0.1, "bath_salts" = 0.20, "plantmatter" = 0.05)
+	reagents_add = list("thc" = 0.15, "lsd" = 0.10, "Atropine" = 0.05, "bath_salts" = 0.35, "plantmatter" = 0.05)
 
 /obj/item/reagent_containers/food/snacks/grown/ambrosia/cruciatus
 	seed = /obj/item/seeds/ambrosia/cruciatus


hal9000PR:
Is this item a joke? Yes. However I do not believe it warrants this. Botany is still, arguably, the strongest department for an antag. Why buff them? I really feel like this is unnecessary because botany is already, despite the nerfs, still super strong. I think this item being bad is fine because the rest of the department makes up for it.

Dashsora:
Agreed with Hal. There's a reason atropine was removed. Initropidril in plants is still attainable, just requires much more work, as it really REALLY should. This would completely undermise the atropine removal in omegas. 

Besides, the  whole point of this seed is to be hidden - to make people think it's ambrosia. By changing so much of it it really strays away from that... and if you do wanna use it to hurt someone with it, just remove the bicaridine and kelotane. Takes four clicks.

Furthermore, with the omega weed balance changes, this seed DID go up in value - as it has much more bath salts than omegas do.

Kyep (commented):
Botany requires severe nerfs.
If you are going to buff botany traitor items, nerf some of the regular botany items (e.g: omegaweed by making them traitor botanist only) at the same time.

Necaladun:
Yeah gonna have to object to buffing botany antags at the current time. I like the idea of the traitor seeds being more interesting, but i'd prefer something interesting and fun than atropine.

AffectedArc07:

PR #14888, opened 2020-11-13
Windowpainter

thatdanguy23:
## What Does This PR Do
Adds an item that changes the colors of basic windows, reinforced windows (as well as both of their fulltile variants) and windoors. It's based off of the pipe painter code, so you can only choose from the same presets as you would painting pipes, but it's a start.

## Why It's Good For The Game
It's just an extra something to make customizing a bit more fun.

## Images of changes
image
image
image
image

## Changelog
:cl: thatdanguy23
add: added a window painting tool
/:cl:


code/game/objects/items/devices/pipe_painter.dm
```diff
@@ -33,7 +33,7 @@
 
 
 /obj/item/pipe_painter/attack_self(mob/user as mob)
-	mode = input("Which colour do you want to use?", "Pipe Painter", mode) in modes
+	mode = input("Which colour do you want to use?", name, mode) in modes
```
thatdanguy23, line 36: Changed to ensure that it said "Window Painter" instead of "Pipe Painter".

AffectedArc07 (changes requested):
Few minor things

AffectedArc07 (approved):

code/game/objects/items/devices/window_painter.dm
```diff
@@ -0,0 +1,15 @@
+/obj/item/pipe_painter/window_painter
+	name = "window painter"
+	icon_state = "window_painter"
+
+/obj/item/pipe_painter/window_painter/afterattack(atom/A, mob/user as mob)
+	if(!(istype(A,/obj/structure/window/reinforced) || istype(A,/obj/structure/window/basic) || istype(A,/obj/machinery/door/window) || istype(A,/obj/structure/window/full/reinforced) || istype(A,/obj/structure/window/full/basic)) || !in_range(user, A))
```
SteelSlayer, line 6: Having so many conditions like this in 1 if statement makes it a bit hard to read, and annoying to add to later. I suggest putting all these types in a list on the window painter, i.e
```dm
var/list/paintable_windows = list(
	/obj/structure/window/reinforced,
	/obj/structure/window/basic
	// etc.
)
```
And then you can do:
```dm
if(!is_type_in_list(A, paintable_windows) || !in_range(user, A))
	return
```

code/game/objects/items/devices/window_painter.dm
```diff
@@ -0,0 +1,22 @@
+/obj/item/pipe_painter/window_painter
+	name = "window painter"
+	icon_state = "window_painter"
+
+var/list/paintable_windows = list(
+		/obj/structure/window/reinforced,
+		/obj/structure/window/basic,
+		/obj/structure/window/full/reinforced,
+		/obj/structure/window/full/basic,
+		/obj/machinery/door/window)
```
AffectedArc07, line 10:
```suggestion
	var/list/paintable_windows = list(
			/obj/structure/window/reinforced,
			/obj/structure/window/basic,
			/obj/structure/window/full/reinforced,
			/obj/structure/window/full/basic,
			/obj/machinery/door/window)
```
Needs to be indented to be part of the definition, not a global (unindented) list

AffectedArc07 (changes requested):

AffectedArc07 (approved):

SteelSlayer (approved):

Kyep (approved):
Initially I was not in favor of this, as I felt it would be confusing to have windows that for example looked like plasmaglass but were not.

After thinking about it though, color-coding would be nice and opens up new design options. Not to mention it lets people customize their workspaces. And the available colors (non,red,blue,teal,green,yellow,purple) really don't seem that atrocious.

AffectedArc07 (approved):

PR #14891, opened 2020-11-14
Allows doctors to eject cloning pods, activates the radio cloners have.

Qwertytoforty:
## What Does This PR Do

Updates most of cloner code to TRUE / FALSE, changes the access requirement to stop a clone to medical instead of genetics, enables the radio cloners have

## Why It's Good For The Game

Stopping the cloner from cloning, or clearing out the bloody mess inside, is not ment to be a hard thing to do. It is supposed to be CMO, or geneticists, that clean out the cloner. However, with genetics often never running the cloner, only really the CMO ever does this, who is often busy. By making it medical, it just makes cleaning the cloner easier, and stops medical from having to ask borgs to turn off the APC to stop a clone early.

Having the radio for the cloner enabled is nice, tells when the cloner gets jammed, or when someone finishes cloning. No real downside to having it on.

## Images of changes

Radio messages:
image
image
image
image

## Changelog
:cl:
tweak: Cloner now requires medical access instead of genetics to eject a cloning pod.
tweak: The radio inside of cloners is now enabled.
/:cl:

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -220,28 +220,28 @@ GLOBAL_LIST_INIT(cloner_biomass_items, list(\
 //Start growing a human clone in the pod!
 /obj/machinery/clonepod/proc/growclone(datum/dna2/record/R)
 	if(mess || attempting || panel_open || stat & (NOPOWER|BROKEN))
-		return FALSE
+		return 0
 	clonemind = locate(R.mind)
 	if(!istype(clonemind))	//not a mind
-		return FALSE
+		return 0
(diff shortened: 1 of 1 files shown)


PacifistDalek:
What's an authorized vs unauthorized ejection?

Qwertytoforty:
Technically an emag on the pod is the unauthorised ejection, but it just calls go_out, and has no message

SteelSlayer (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #14902, opened 2020-11-15
Notes can be removed from airlocks with wirecutters, even if the airlock maintenance panel is closed

moxian:
## What Does This PR Do
Notes can be removed from airlocks with wirecutters, even if the airlock maintenance panel is closed.

## Why It's Good For The Game
This is a salt-induced PR. I couldn't figure out how to remove the note without diving into the code.
After diving I realized that
1) i should have used grab intent, not harm, when trying by hand (although, admittedly, examine text should tell me what to do very clearly) and
2) to cut it with wirecutters one needs to open the panel first, which is unintuitive and must be an oversight.

This PR fixes the latter.

## Images of changes
20201115-231617-GZR34kRlKZ

## Changelog
:cl:
fix: No longer need to open airlock panel to wirecutter the note out of it.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -719,7 +719,7 @@ About the new airlock wires panel:
 		return
 
 	if(headbutt_airlock(user))
-		return//Smack that head against that airlock
+		return // Smack that head against that airlock
 	if(remove_airlock_note(user, FALSE))
 		return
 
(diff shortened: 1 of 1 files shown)


Adrer:
I mean... It does kinda... Say so when you inspect the door
image

moxian:
Thanks. Slimmed down the PR to just the wirecutter interaction bufix.

SteelSlayer (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #14908, opened 2020-11-16
Makes morgue tray respect unrevivable mobs

warior4356:
## What Does This PR Do
Adds logic from cloner to morgue trays, so suicided or otherwise unrevivable mobs don't show up as green or purple.

## Why It's Good For The Game
It's confusing when someone who committed suicide shows up as green in the morgue.

## Changelog
:cl:
fix: Morgue trays now respect unrevivable mobs for their lights
/:cl:

### DIFF
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -49,12 +49,16 @@
 			if(M)
 				var/mob/dead/observer/G = M.get_ghost()
 
-				if(M.client && M.mind && M.mind.is_revivable() && !M.mind.suicided)
-					icon_state = "morgue3"
-					desc = initial(desc) + "\n[status_descriptors[4]]"
-				else if(G && G.client && M.mind && M.mind.is_revivable() && !M.mind.suicided) //There is a ghost and it is connected to the server
-					icon_state = "morgue5"
-					desc = initial(desc) + "\n[status_descriptors[6]]"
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/morgue.dm
```diff
@@ -49,10 +49,10 @@
 			if(M)
 				var/mob/dead/observer/G = M.get_ghost()
 
-				if(M.client)
+				if(M.client && M.mind && M.mind.is_revivable() && !M.mind.suicided)
 					icon_state = "morgue3"
 					desc = initial(desc) + "\n[status_descriptors[4]]"
-				else if(G && G.client) //There is a ghost and it is connected to the server
+				else if(G && G.client && M.mind && M.mind.is_revivable() && !M.mind.suicided) //There is a ghost and it is connected to the server
```
SteelSlayer, line 55: I think it would be better to do something like this to avoid potentially repeating some checks
```dm
if(M.mind && M.mind.is_revivable() && !M.mind.suicided)
	if(M.client)
		icon_state = "morgue3"
		desc = initial(desc) + "\n[status_descriptors[4]]"
	else if(G && G.client) //There is a ghost and it is connected to the server
		icon_state = "morgue5"
		desc = initial(desc) + "\n[status_descriptors[6]]"
```

warior4356: I redid the logic

Kyep (approved):

AffectedArc07 (approved):

Issue #14909, opened 2020-11-16
Suit Storage Unit electrification wire not electrocuting user, breaking wire interface and some additional odd disinfection cycle behaviour


Vulkoras:
**Issue Description**:
(Tested on a private server and to a limited extent the live server (Tested the electrification issues but not the disinfection cycle given its potential to cause problems/impact the round) with the same results).

(I have very minimal understanding of the code/internal workings involved and am only reporting based on what I've observed in game and from a crude examination of the files involved).

The electrification and UV hacking functions of suit storage units both appear to have a few issues I've tried to document to the best of my abilities below.

**What did you expect to happen**: 
The electrification wire to shock users.

The electrification of the unit to largely have no effect on it beyond potentially shocking users.

High risk items within the unit to emerge undamaged at the end of a overclocked disinfection cycle while other items are destroyed permanently.  

**What happened instead**:
The unit produced sparks but caused no shocks.

In the five seconds following pulsing or indefinitely after cutting the shock wire the unit's wire interface could no longer be interacted with or opened once closed.

High risk items were destroyed*/disappeared into the unit.  Any Items previously destroyed reappear when a carbon or simple mob enters then exited the unit or when the unit is destroyed.

**Why is this bad/What are the consequences**:
As far as I can tell the lack of electrocution is non intentional and largely takes the risk out of hacking the units.

The wire interface breaking is problematic given the inability to revert any hacked functions already enabled and somewhat compounded by the lack of an easy means to disassemble the unit (as far as I'm aware of).

The ability to disappear/destroy high risk objects is as far as I know intended to be largely impossible and even though this method isn't permanent given the other behaviour, it still makes them extremely difficult to find (the advanced pinpointer still leads to the items but isn't of much use unless you know the workarounds). Conversely having a means to destroy stolen or contraband EVA equipment is neat and having it reappear (fingerprints intact as far as I can tell) when the unit is destroyed or someone hides in it (or stuffs poly inside) is unfortunate.     

**Steps to reproduce the problem**:
When electrified interact with the wires or other functions of the machine without insulated gloves/etc.

Pulse or cut the Electrification wire (toggles green light) on a suit storage unit then attempt to interact with any wires in the 5 seconds/indefinite period afterwards. 

[...]

PR #14912, opened 2020-11-16
Suit storage unit wires now be accessible when electrocuted and hacked UV radiation properly destroys contents

Adrer:
## What Does This PR Do
fixes: #14909

Do note that the suit storage unit is **still** capable of permanently deleting high value items, I assumed this to be intended behaviour, but if I am wrong in this, please do let me know.

Also slightly cleans up the code

## Why It's Good For The Game
Bugs are bad.

## Changelog
:cl:
fix: Suit storage unit will now actually electrocute when electrocute wire is cut/pulsed
fix: Suit storage unit wires can now still be accessed if the electrocute wire is cut/pulsed
fix: Suit storage unit UV radiation at 15 nm now will permanently delete the unit's contents, rather than allowing it to be conjured back into existence
/:cl:

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -292,18 +292,15 @@
 	if(shocked)
 		if(shock(user, 100))
 			return
-
 	if(!is_operational())
 		if(panel_open)
 			to_chat(usr, "<span class='warning'>Close the maintenance panel first.</span>")
 		else
 			to_chat(usr, "<span class='warning'>The unit is not operational.</span>")
 		return
-
 	if(panel_open)
 		wires.Interact(user)
 		return
-
 	if(state_open)
 		if(store_item(I, user))
 			update_icon()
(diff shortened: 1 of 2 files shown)


code/game/machinery/suit_storage_unit.dm
```diff
@@ -288,10 +288,11 @@
 	else if(occupant)
 		add_overlay("human")
 
-/obj/machinery/suit_storage_unit/attackby(obj/item/I as obj, mob/user as mob, params)
+/obj/machinery/suit_storage_unit/attackby(obj/item/I as obj, mob/user, params)
```
SteelSlayer, line 291: Since you're touching this proc already.
```suggestion
/obj/machinery/suit_storage_unit/attackby(obj/item/I, mob/user, params)
```

code/game/machinery/suit_storage_unit.dm
```diff
@@ -301,7 +302,6 @@
 
 	if(panel_open)
 		wires.Interact(user)
-		return
```
SteelSlayer, line 304: Without this return, when you use a multitool/wirecutter on the unit to open the wires, you will also hit the unit with your tool.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -539,19 +524,15 @@
...
 	var/dat
-	if(shocked)
-		if(shock(usr, 100))
+	if(shocked && !(stat & NOPOWER))
+		if(shock(user, 100))
 			return
-	if(..())
-		return
-	if(stat & NOPOWER)
-		return
```
SteelSlayer, line 550: 1) Without the parent call here, you skip a bunch of stuff, like an incapacitated check, a brain damage check, adding fingerprints of the mob to the unit, etc.
2) Without some sort of power check like this, you can open the UI of the unit while it doesn't have power, which I think is a bit weird.

Kyep (approved):

AffectedArc07 (approved):

Issue #14691, opened 2020-10-22
Runtime with Re-Admin Self on local code.


Mitchs98:
**Issue Description**:
[2020-10-21T19:15:27] Runtime in admin_verbs.dm,681: list index out of bounds
  proc name: Re-admin self (/client/proc/readmin)
  usr: Zachary Cherry (mitchs98) (/mob/dead/observer)
  usr.loc: The floor (126,141,1) (/turf/simulated/floor/plasteel)
  src: Mitchs98 (/client)
  call stack:
  Mitchs98 (/client): Re-admin self()

This happens when trying to re-admin self after de-adminning self to test things that require you to interact as a ghost without admin ghost interaction abilities. You are, however, re-adminned if you log back in after using it.
**What did you expect to happen**: 
I expected it to work without having to log in and to not give me a runtime.

**What happened instead**:
It gave me a runtime.

**Why is this bad/What are the consequences**:
Mild annoyance. Runtime man bad.

**Steps to reproduce the problem**:
1. Aghost.
2. De-admin self.
3. Re-admin self.
4. Have to relog to re-admin self.

**When did the problem start happening**:
No clue

**Extra information**:
N/A

Mitchs98:
Will look into fixing it myself by reviving Moxian's PR(https://github.com/ParadiseSS13/Paradise/pull/12721).

Will leave this open in the mean time.

PR #14918, opened 2020-11-17
Fixes re-admin runtime

SabreML:
## What Does This PR Do

Fixes: #14691 

## Why It's Good For The Game

Being unable to re-admin yourself is a pretty big problem, even if it only happens with the legacy system.

## Changelog
:cl:
fix: Fixed a runtime when re-adminning yourself using the legacy admin system
/:cl:


AffectedArc07:
Going to do a lot of local testing with this one, since a previous fix for this allowed for a permissions elevation exploit.

SabreML:
> Going to do a lot of local testing with this one, since a previous fix for this allowed for a permissions elevation exploit. 

It _should_ be fine since the actual rank setting code is untouched; I've just added a few things around it.

But yeah, testing is always good.

AffectedArc07 (approved):

PR #14923, opened 2020-11-18
Actually prevents swarmers from eating high-value targets

warior4356:
## What Does This PR Do
Swarmers already had a check for preventing them from eating high-value items or other items flagged as indestructible, however, the parent call was missing. Also, this gives the captain's rapier the same resistance flags as all other high-value items.

Also adds an IC reason that swarmers can't eat high-value targets, eg: "High value materials detected, consuming captain's rapier would waste them. Aborting."

## Why It's Good For The Game
This is just a bug fix and bringing an item to parity with other high-value targets.

## Changelog
:cl:
fix: Fixes bug that allowed swarmers to eat high-value targets
/:cl:


Kyep (approved):

AffectedArc07 (approved):

Jungling:
captain getting his rapier eaten is peak swarmer gameplay.

PR #14930, opened 2020-11-18
Adds a setting to disable chat bubbles when typing 'me' emotes

SabreML:
## What Does This PR Do

Adds a new preferences setting to disable chat bubbles when typing 'me' emotes.

Also fixed the typing indicator staying active after you become unconscious.

## Why It's Good For The Game

It can be pretty irritating sometimes for an indicator to pop up when you're trying to type an emote, especially as an animal.
I personally always end up using the command bar to write out emotes because of this, and I'm sure I can't be the only one.

## Changelog
:cl:
add: Added a new preference to disable chat bubbles when typing 'me' emotes
fix: Chat bubbles are now removed when you become unconscious
/:cl:

**~~Currently set to draft until I figure out the SQL stuff~~**

**SQL stuff *seems* to be working, since it's saving the preferences between rounds. Undrafted now.**


code/__DEFINES/preferences.dm
```diff
@@ -22,22 +22,21 @@
 #define PREFTOGGLE_CHAT_DEBUGLOGS 			128
 #define PREFTOGGLE_CHAT_LOOC 				256
 #define PREFTOGGLE_CHAT_GHOSTRADIO 			512
-#define PREFTOGGLE_SHOW_TYPING 				1024
-#define PREFTOGGLE_DISABLE_SCOREBOARD 		2048
```
AffectedArc07, line 26: Absolutely fucking not. This will corrupt all current bitflag entries.

AffectedArc07 (changes requested):

code/__DEFINES/preferences.dm
SabreML: Does shifting them all down by one not stop that?
Surely unless something is checking for, say, `262144` rather than `PREFTOGGLE_DISABLE_KARMA` everything will work the same way since it's all defined.

AffectedArc07: It shifts them down ingame, but not in the DB. 1024 in the DB before this will equal typing, but will equal scoreboard after, breaking everyones prefs.

code/modules/mob/say.dm
```diff
@@ -37,7 +37,7 @@
 
 	message = strip_html_properly(message)
 
-	set_typing_indicator(0)
+	set_typing_indicator(0, TRUE)
```
AffectedArc07, line 40: If state is a bool, please refactor it here as well

code/modules/mob/typing_indicator.dm
```diff
@@ -74,21 +81,35 @@ GLOBAL_LIST_EMPTY(typing_indicator)
...
+	set category = "Preferences"
+	set desc = "Toggles showing an indicator when you are typing an emote."
+	prefs.toggles2 ^= PREFTOGGLE_2_EMOTE_BUBBLE
+	prefs.save_preferences(src)
+	to_chat(src, "You will [(prefs.toggles2 & PREFTOGGLE_2_EMOTE_BUBBLE) ? "no longer" : "now"] display a typing indicator for emotes.")
+
+	feedback_add_details("admin_verb", "EID")
 
-	feedback_add_details("admin_verb","TID") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
+#undef TYPING_INDICATOR_LIFETIME
```
AffectedArc07, line 115: Please add the comment that warns about copypaste to the feedback add details here.

AffectedArc07 (changes requested):

code/modules/mob/say.dm
SabreML: It's kinda a boolean? I wasn't sure if I should use 'true' and 'false' there since it's meant to be 'on' or 'off'.

AffectedArc07: You should

code/modules/mob/typing_indicator.dm
```diff
@@ -69,26 +76,40 @@ GLOBAL_LIST_EMPTY(typing_indicator)
...
+	set desc = "Toggles showing an indicator when you are typing a message."
 	prefs.toggles ^= PREFTOGGLE_SHOW_TYPING
 	prefs.save_preferences(src)
 	to_chat(src, "You will [(prefs.toggles & PREFTOGGLE_SHOW_TYPING) ? "no longer" : "now"] display a typing indicator.")
 
 	// Clear out any existing typing indicator.
 	if(prefs.toggles & PREFTOGGLE_SHOW_TYPING)
-		if(istype(mob)) mob.set_typing_indicator(0)
+		if(istype(mob))
+			mob.set_typing_indicator(FALSE)
```
AffectedArc07, line 100: Also may be worth cloning this in

AffectedArc07 (approved):

PR #14931, opened 2020-11-18
Stops explosions causing millions of runtimes

farie82:
## What Does This PR Do
Stops explosions affecting GCed objects.
Objects that were destroyed while the explosion was in it's CHECK_TICK phase (AKA sleep on high load) were still considered when looping through them.
This PR checks if an atom is `QDELETED`. If so. Skip it

Test case:
Make about 5 breaches using the explosion on turf tool admins have in VV.
Spawn about 300 monkeys.
Run SDQL: CALL flicker() ON /obj/machinery/light

Make two bombs explode close to one another

## Why It's Good For The Game
Fixes a ton of runtimes related to explosions. Most notably the atmos pipenet one. `parent == null`

## Changelog
:cl:
fix: Deleted atoms are no longer affected by explosions
/:cl:

### DIFF
--- a/code/game/objects/explosion.dm
+++ b/code/game/objects/explosion.dm
@@ -122,13 +122,13 @@
 							affecting_level = S.is_shielded() ? 2 : (S.intact ? 2 : 1)
 						for(var/atom in S.contents)	//bypass type checking since only atom can be contained by turfs anyway
 							var/atom/AM = atom
-							if(AM && AM.simulated)
+							if(!QDELETED(AM) && AM.simulated)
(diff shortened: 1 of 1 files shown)


AffectedArc07:
Sadly this wont fix pipenets fucking up when they get merged together horribly, but this is a good start.

farie82:
> Sadly this wont fix pipenets fucking up when they get merged together horribly, but this is a good start.

Aye it won't. It is the root cause for most atmos related issues though I've seen on the station

AffectedArc07 (approved):

PR #14938, opened 2020-11-19
Removes meteor shot from crafting menu

Darkmight9:
## What Does This PR Do
As the name says, it removes meteor shot from the crafting menu. Do note that the shell itself still exists but crew can no longer normally get it.

Edit: an alternative PR has been made if people still want to keep the shells, however it would respect anchored objects now #14941

## Why It's Good For The Game
Meteor shot is able to move things that you _really_ shouldn't be moving or shouldn't be movable at all. Examples include, the gateway the grav gen, the vault door, the safe within, Lava land tendrils, shutter walls, and other things.

## Changelog
:cl:
del: Removed meteor shot from the crafting menu
/:cl:

### DIFF
--- a/code/modules/crafting/recipes.dm
+++ b/code/modules/crafting/recipes.dm
@@ -151,17 +151,6 @@
 	subcategory = CAT_WEAPON
 	alert_admins_on_craft = TRUE
 
-/datum/crafting_recipe/meteorshot
-	name = "Meteorshot Shell"
-	result = /obj/item/ammo_casing/shotgun/meteorshot
-	reqs = list(/obj/item/ammo_casing/shotgun/techshell = 1,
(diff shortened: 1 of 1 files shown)


Kyep (approved):
Yes. We need this.
It makes no sense to be able to shoot a section of forcefield, and detach that section from the rest as if you were taking a brick out of a wall.
Things which are anchored are not supposed to be movable. They're not coded to support being moved. Meteorshot violates this.
If you want a type of ammo which absolutely shreds airlocks, and can be used for breaching, have an ammo type that shreds airlocks on contact. Not meteorshot. The ability to push ANYTHING is broken.

procdrone:
PULSE shot i believe is sufficient enough to break down doors. Not sure if its still doable though.

PR #14939, opened 2020-11-19
Doubles the contents of the jani closet

Darkmight9:
## What Does This PR Do
As the name says, it doubles most of the contents of the janitor closet. I also decided to add a second space cleaner spray bottle to the closet to get around the mapping freeze and because those bottles are dang useful.

## Why It's Good For The Game
Janitor is a common station job and is kind of the dump job for HoPs. Usually however the first janitor takes all the tools leaving the second janitor to not have any equipment. This semi helps alleviate the problem.

## Changelog
:cl:
tweak: Doubled most of the contents of the janitor closet
/:cl:


Kyep (approved):

farie82:
Did you check if you could open and close the locker and see if all items fit when it closes?

code/game/objects/structures/crates_lockers/closets/job_closets.dm
```diff
@@ -73,23 +73,38 @@
 /obj/structure/closet/jcloset/New()
 	..()
 	new /obj/item/clothing/under/rank/janitor(src)
```
AffectedArc07, line 75: Theres now 32 items here. A closet's capacity is 30. Please reduce the item count.

AffectedArc07 (changes requested):

code/game/objects/structures/crates_lockers/closets/job_closets.dm
Darkmight9: I knew I should have tested closing the locker...

PR #14941, opened 2020-11-19
Makes Meteorshot Respect Anchored

Leanfrog:
## What Does This PR Do
This PR makes meteorshot respect anchored objects and unable to move them
alternative to #14938 

## Why It's Good For The Game
Meteorshot shouldn't be able to move anchored objects, this stops it from doing that while still leaving the shell available for use

## Changelog
:cl:
tweak: Meteorshot respects anchored objects now
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets.dm
+++ b/code/modules/projectiles/projectile/bullets.dm
@@ -198,8 +198,9 @@
 
 /obj/item/projectile/bullet/meteorshot/on_hit(var/atom/target, var/blocked = 0)
 	..()
-	if(istype(target, /atom/movable))
-		var/atom/movable/M = target
+	var/atom/movable/M = target
+	if(M.anchored)return
+	else if(istype(target, /atom/movable))
 		var/atom/throw_target = get_edge_target_turf(M, get_dir(src, get_step_away(M, src)))
 		M.throw_at(throw_target, 3, 2)
 


code/modules/projectiles/projectile/bullets.dm
```diff
@@ -198,8 +198,9 @@
 
 /obj/item/projectile/bullet/meteorshot/on_hit(var/atom/target, var/blocked = 0)
 	..()
-	if(istype(target, /atom/movable))
-		var/atom/movable/M = target
+	var/atom/movable/M = target
+	if(M.anchored)return
+	else if(istype(target, /atom/movable))
```
SteelSlayer, line 203: You're casting the target to a movable atom and then accessing a var on it before you know if its a movable, which can be bad. I suggest keeping it similar to the format of what it was previously:
```dm
if(istype(target, /atom/movable))
	var/atom/movable/M = target
	if(M.anchored)
		return
	...
```

code/modules/projectiles/projectile/bullets.dm
```diff
@@ -198,8 +198,9 @@
 
 /obj/item/projectile/bullet/meteorshot/on_hit(var/atom/target, var/blocked = 0)
 	..()
-	if(istype(target, /atom/movable))
-		var/atom/movable/M = target
+	var/atom/movable/M = target
+	if(M.anchored)return
```
S-CR, line 202: Is there any particular reason to make this a one-liner? It makes the code really awkward to read.

Leanfrog:
Lost my coding fever, not gonna end up updating this so closing in favor of the alternate

PR #14956, opened 2020-11-21
Reverts 'Sling thralling now respects mindshields'

Cocacolagua:
## What Does This PR Do
Reverts everything in #14521

## Why It's Good For The Game
I hate shadowlings with a passion, I believe they're the second-worst gamemode after roundstart blobs, but this change made them horrible.
Shadowlings have a lot of problems, but the biggest one being that it's a snowball gamemode, where if one side has the upperhand, it's unlikely the other side will ever recover. You could usually see this when most shadowlings died, they would almost never catch up, this also happened when security were the first ones getting thralled.
This change, however, almost never win, as the shadowlings now couldn't convert the department that is actually expected to step into maintenance in the gamemode, making this even more TDM.
I am not happy with thralling security either, but this change works for now, without needing to make a full rework of shadowlings on the entirety.

## Changelog
:cl:
tweak: Shadowlings can thrall mindshielded personnel again.
/:cl:

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -236,9 +236,6 @@
 	if(!(ling.mind in SSticker.mode.shadows))
 		return
 	var/mob/living/carbon/human/target = targets[1]
-	if(ismindshielded(target))
-		to_chat(user, "<span class='danger'>This target has a mindshield, blocking your powers! You cannot thrall it!</span>")
-		return
 	enthralling = TRUE
 	to_chat(user, "<span class='danger'>This target is valid. You begin the enthralling.</span>")
 	to_chat(target, "<span class='userdanger'>[user] stares at you. You feel your head begin to pulse.</span>")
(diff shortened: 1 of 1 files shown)


Mitchs98:
I'm not at all agreeing with this change because the mindshielding lowers sling winrate. What I am agreeing with reverting the PR for is based on the fact this PR causes slings to either 1. Murderbone everything with a mindshield or 2. Get a thrall to remove the mindshield which is a risky process so most won't do it leading to option 1.

TL;DR: Murderbone bad.

ppi13:
This will just end up with murder bonering anyway, don't forget that slings do when they ascend, so saying murderbone bad when thats was aclings do seems silly, and that at some point, murderbonering thralls is just better.

There is a forum thread on it
https://www.paradisestation.org/forum/topic/19748-shadowlings-post-mindshield-thralling-nerf/

Mitchs98:
> This will just end up with murder bonering anyway, don't forget that slings do when they ascend, so saying murderbone bad when thats was aclings do seems silly, and that at some point, murderbonering thralls is just better.
> 
> There is a forum thread on it
> https://www.paradisestation.org/forum/topic/19748-shadowlings-post-mindshield-thralling-nerf/

You and I both know comparing the end game of slings to the main gameplay of slings makes absolutely no sense to do. This PR is to address the gameplay, not to affect the end-game when the round is basically over anyway.

Shockpoint:
It took me one round of shadowlings to even want this. 
From an admin point of view, all I saw was attack logs of shadowlings and thralls killing people. Basically preventing me from handling anything actually important.
It also depresses me because I know the first default method to go to will be murderboning. Nobody wants to waste their time with that.
A revert would be good until we have a better way of handling this. There was a suggestion about letting sec be thralled after like 5 normal thralls have been made. That'd be okay.

Mitchs98:
> It took me one round of shadowlings to even want this.
> From an admin point of view, all I saw was attack logs of shadowlings and thralls killing people. Basically preventing me from handling anything actually important.
> It also depresses me because I know the first default method to go to will be murderboning. Nobody wants to waste their time with that.
> A revert would be good until we have a better way of handling this. There was a suggestion about letting sec be thralled after like 5 normal thralls have been made. That'd be okay.

This is something I'd definitely agree with. It would be a happy medium, and an incentive for slings to play it cool while leaving security alive for some sweet security converts once they get a few people. A reward for playing the game in a good non murderboney fashion.

Necaladun:
I'd like to find a sweet spot between "too easily" and "not at all" when it comes to thralling sec officers. Some decent ideas have been thrown around with a secondary ability allowing mindshield melting.

DevL-Whats-a-jpg:
Yea, i like the idea of needing _X_ thralls before thralling sec, and then just have it take an extra _X_ seconds. Something that also gives a indicator that a seccie is thralled would be good to, like upon examining while near them you get a message like.
_The skin on their head is burned and charred_

shmabuu:
After now having experienced playing as the sling and as sec against slings in one week, being killed by slings instead of getting thralled and having to kill sec absolutely sucks. I can deal with dying and most likely not getting revived in a round. But that's just me. Many other people don't feel the same way. I'd much rather thrall a sec and actually let them keep playing the game during shadowlings instead of killing them semi-permanently. People have complained that "Thralling sec gives them too much of a snowball" and yes, sec gear is strong. But slings are a snowballing game mode anyway. And if the slings have thralls before killing said sec member, a thrall is getting that gear anyway. Once they get enough thralls they are much harder to stop. Now that you can't thrall sec, it just turns into "Revolutionary Plus". This point is reinforced even more if sec mass orders mindshields like revolutionaries.

AffectedArc07:
We have had an internal discussion and decided this is not the solution we want. Also there's 2 people objecting and no one in favour.

Daylight2:
> We have had an internal discussion and decided this is not the solution we want. Also there's 2 people objecting and no one in favour.

What is the solution you want then? Having every validhunter and their pet dog be immune to slings is no fun for the sling.

Dafrek:
> We have had an internal discussion and decided this is not the solution we want. Also there's 2 people objecting and no one in favour.

Do you have a general idea of what is going to happen to slings instead? Personally, I really supported the sling nerf being reverted since the nerf forced them to murder sec and made them a whole lot less powerful even though they didn't seem overpowered before to me. 
If there is another way you want to go to make them balanced again I'd be down to hear it especially since its been over three months since the nerf was implemented.

PR #8464, opened 2017-12-21
Removes FlattestGuitar

SamHPurp:
So, we tracked the shitcode that causes the server to crash on load 9 times out of 10.

Hug your local FlattestGuitar today. (And kick him in the bollocks)

Reverts the OOC strikethrough/italics PR.

love,
Purpose

:cl: Purpose2
del: You may no longer use strikethroughs/italics in OOC. This was causing backend server problems due to how shoddily BYOND is coded, and cannot be safely added right now.
/:cl:

PR #14959, opened 2020-11-21
Adds the ability to italicise words when speaking or emoting

SabreML:
## What Does This PR Do

Adds the ability to *italicise* words when speaking or emoting by surrounding the words in brackets `(Hello!)`

There needs to be an even number of brackets for the words to become italicised, so typing `((Test)` or `(Test))` won't work.

*(Originally used regex, but changed to DM due to stability concerns.)*

## Why It's Good For The Game

Greatly increases roleplaying potential.

Have you ever got *really* irritated that you can't actually emphasise words through text?
Have you ever wanted to convey sarcasm *without* massively over exaggerating it?
Now you can!

## Images of changes

italics

## Changelog
:cl:
add: You can now italicise words when talking! Just put a pair of brackets `()` around the word or words you want to change. e.g: (This sentence is italicised!)
/:cl:

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -145,6 +145,18 @@
 
 	return null
 
+#define ENCODE_HTML_EMPHASIS(input, char, html, varname) \
+	var/static/regex/##varname = regex("[char]{2}(.+?)[char]{2}", "g");\
+	input = varname.Replace(input, "<[html]>$1</[html]>")
+
+/mob/proc/say_emphasis(input)
+	ENCODE_HTML_EMPHASIS(input, "\\/", "i", italics)
+	//ENCODE_HTML_EMPHASIS(input, "\\+", "b", bold)
+	//ENCODE_HTML_EMPHASIS(input, "_", "u", underline)
+	return input
+
+#undef ENCODE_HTML_EMPHASIS
+
 /datum/multilingual_say_piece
 	var/datum/language/speaking = null
 	var/message = ""
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -145,17 +145,28 @@
 
 	return null
 
-#define ENCODE_HTML_EMPHASIS(input, char, html, varname) \
-	var/static/regex/##varname = regex("[char]{2}(.+?)[char]{2}", "g");\
-	input = varname.Replace(input, "<[html]>$1</[html]>")
-
+/**
+  * Replaces all brackets from the input with italics
+  *
+  * Requires an equal number of opening and closing brackets, to avoid overflow.
+  */
 /mob/proc/say_emphasis(input)
-	ENCODE_HTML_EMPHASIS(input, "\\/", "i", italics)
-	//ENCODE_HTML_EMPHASIS(input, "\\+", "b", bold)
-	//ENCODE_HTML_EMPHASIS(input, "_", "u", underline)
-	return input
+	/// Number of opening brackets in the input
+	var/opening = 0
+	/// Number of closing brackets in the input
+	var/closing = 0
 
-#undef ENCODE_HTML_EMPHASIS
+	for(var/I in 1 to length(input))
+		var/char = input[I]
+		if(char == "(")
+			opening++
+		else if(char == ")")
+			closing++
+
+	if(opening <= closing) // Opening number isn't higher
+		input = replacetext(input, "(", "<i>")
+		input = replacetext(input, ")", "</i>")
+	return input
 
 /datum/multilingual_say_piece
 	var/datum/language/speaking = null
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -148,20 +148,26 @@
 /**
   * Replaces all brackets from the input with italics
   *
-  * Requires an equal number of opening and closing brackets, to avoid overflow.
+  * Formats `The (quick) brown fox (jumps over) the lazy dog()` to `The <i>quick</i> brown fox <i>jumps over</i> the lazy dog<i></i>`
+  *
+  * If there is an uneven number of brackets, none of them will be changed in order to avoid overflow.
   */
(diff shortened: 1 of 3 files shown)


AffectedArc07:
Going to link https://github.com/ParadiseSS13/Paradise/pull/8464 here. Using regex like this for matching in OOC, causing major server instability issues. 

Not in favour at all.

SabreML:
> Going to link #8464 here. Using regex like this for matching in OOC, causing major server instability issues.
> 
> Not in favour at all.

Assuming the issue is regex in OOC specifically,
image

AffectedArc07:
> > Going to link #8464 here. Using regex like this for matching in OOC, causing major server instability issues.
> > Not in favour at all.
> 
> @AffectedArc07 Assuming the issue is regex in OOC specifically,
> image

The issue isnt the fact its in OOC, it was the regex matching itself, and likely high frequency of it.

SabreML:
I've tested initializing and starting a server around twenty times, plus everything during testing. No crashes or runtimes.
I'm pretty confident at this point that whatever made #8464 crash the server was either fixed in a byond update or in some other refactor PR.

On top of that, [Citadel](https://github.com/Citadel-Station-13/Citadel-Station-13/blob/master/code/game/say.dm#L103), [Skyrat](https://github.com/Skyrat-SS13/Skyrat13/blob/986f8e6c07a0f41c674c42f0064b3a94f038fb50/code/game/say.dm#L103), [Vorestation](https://github.com/VOREStation/VOREStation/blob/27d3bf2b0f917b40f7cb3480bad84d23e2fe4f80/code/modules/mob/hear_say.dm#L154) and [Polaris](https://github.com/PolarisSS13/Polaris/blob/4795e295595919273bea57a4c0e96610fe673acc/code/modules/mob/hear_say.dm#L154) *(And probably other downstreams but I'm not counting those)* all have it implimented in some form or another, presumably with no issues.

Kyep (commented):
I too am against this. You can always just put *s around your words, like *this*.
I don't like the idea of more regex parsing of every chat message.

Necaladun:
I fucking *love* italics in chat for emphasis, and anything that helps get tone across in an RP game is fucking great.

I'm neutral on regex parsing of every chat message because I don't understand what that means so this is now maint turf.

SabreML:
@Kyep 
> I too am against this. You can always just put *s around your words, like _this_.

I think I'm misunderstanding something here because using asterisks in-game doesn't have any function for speech, just emotes.
If you meant to put a marker of some sort around the word then I personally think that's a bad solution.
I find that having any sort of symbols in the IC chat really pulls me out of the experience, even if it's something as harmless as an ooc message to help someone.
> **Acolyte John Smith**: "Don't worry, it'll use my cult magic to send you an explanation. [[[www.example.com]]]"

I've never actually seen anyone try to emphasise words before but I can't imagine it would be very nice to look at.
While the above example isn't directly related to italics I hope you can understand my point.
<hr>

> I don't like the idea of more regex parsing of every chat message.

For this point, I really don't understand why this is an issue. From my own testing, it has a very low performance impact and I can't see a way for the regex to fail or break in any way. As for the bug from #8464 I've already commented about that above.
I'm not trying to be confrontational here, I just want to clear up my confusion.

AffectedArc07:
Pretty sure kyet was just meant to type \*this\* but didnt escape the asterisks

SabreML:
Updated to use DM rather than regex, so one way or another the crashing bug won't be an issue.

Also from some stress tests in-game there doesn't seem to be any added delay when talking, so hopefully there won't be any downsides. (Performance wise anyway)

code/modules/mob/say.dm
```diff
@@ -145,6 +145,29 @@
 
 	return null
 
+/**
+  * Replaces all brackets from the input with italics
+  *
+  * Requires an equal number of opening and closing brackets, to avoid overflow.
+  */
+/mob/proc/say_emphasis(input)
+	/// Number of opening brackets in the input
```
AffectedArc07, line 154: Triple / applies to class based var definitions, not in-proc ones. Please remove.

code/modules/mob/say.dm
```diff
@@ -145,6 +145,29 @@
...
+
+	for(var/I in 1 to length(input))
+		var/char = input[I]
+		if(char == "(")
+			opening++
+		else if(char == ")")
+			closing++
+
+	if(opening <= closing) // Opening number isn't higher
+		input = replacetext(input, "(", "<i>")
```
AffectedArc07, line 167: I'm a little paranoid this allows someone to type ))))))(( and italicise an entire window by leaving an unterminated <I> span. Please test this.

AffectedArc07 (changes requested):

code/modules/mob/say.dm
FlattestGuitar: Yeah this logic in general doesn't quite make sense. You count the opening and closing brackets to make sure you can't leave a hanging open tag, but at the same time you don't actually check for hanging open tags. You can just do `)(` to defeat this magic.

Just use a regex. 

The PR from 2017 @AffectedArc07 linked earlier in the comments here had nothing to do with OOC's frequency of use, but was a result of BYOND initializing regexes on server startup improperly, as is explained in the PR and the BYOND forum thread linked there. My bet is this will work fine today, you can always revisit the issue if you have consistent issues with regex usage.

Also, for the love of god, too many comments.

SabreML: That was an issue, yes. Although it would only effect the rest of the `say` line so it wouldn't have been too catastrophic.

Fixed now though, and a few other things I noticed. Fingers crossed I got them all.

variableundefined (commented):
Objection. I do not want to take risk with a PR like this.

FlattestGuitar:
What do you think is risky about this PR, @variableundefined ?

AffectedArc07:
2 maintainers and a head objection. Closing.

PR #14960, opened 2020-11-21
Refactors renaming things with a pen.

moxian:
## What Does This PR Do
Unifies all the places that rename a thing with a pen to use common proc.
Now when renaming things you get a consistent format of "thing - user-provided label", so, e.g. "body bag - John Doe". (Unless overriden of course).
You also now consistently have an option to realize "oh, wait, no, i don't actually want to rename this" and bail out, without ruining the existing label.

The only concern I have is about the format. Some things currently use a "thing - label", while others go "thing (label)", and I don't have a strong enough opinion to meaningfully choose one of the two.

Also (almost unrelated) changes the pen attack handling, so it no longer gives "you stab a guy with a pen" messages that only you (but not your victim!) and admins(!) can see.

## Why It's Good For The Game
QoL

## Changelog
:cl:
tweak: renaming things should be a bit more consistent now.
fix: fix misleading messages when stabbing people with a pen.
/:cl:

### DIFF
--- a/code/ATMOSPHERICS/components/binary_devices/pump.dm
+++ b/code/ATMOSPHERICS/components/binary_devices/pump.dm
@@ -237,12 +237,7 @@ Thus, the two variables affect pump operation are set in New():
 
 /obj/machinery/atmospherics/binary/pump/attackby(obj/item/W, mob/user, params)
 	if(istype(W, /obj/item/pen))
-		var/t = copytext(stripped_input(user, "Enter the name for the pump.", "Rename", name), 1, MAX_NAME_LEN)
-		if(!t)
-			return
-		if(!in_range(src, usr) && loc != usr)
-			return
-		name = t
+		rename_interactive(user, W)
 		return
 	else if(!istype(W, /obj/item/wrench))
 		return ..()
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -968,39 +968,52 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 	Arguments:
 	* user - the renamer.
 	* implement - the tool doing the renaming (usually, a pen).
-	* freeform - whether we allow to change the name arbitrarily, or if it should go "thing - user-given label".
+	* use_prefix - whether the new name should follow the format of "thing - user-given label" or
+		if we allow to change the name completely arbitrarily.
 	* actually_rename - whether we want to really change the `src.name`, or if we want to do everything *except* that.
 	* prompt - a custom "what do you want rename this thing to be?" prompt shown in the inpit box.
 
 	Returns: Either null if the renaming was aborted, or the user-provided sanitized string.
  **/
-/atom/proc/rename_interactive(mob/user, obj/implement = null, freeform = FALSE,
+/atom/proc/rename_interactive(mob/user, obj/implement = null, use_prefix = TRUE,
 		actually_rename = TRUE, prompt = null)
 	// Sanity check that the user can, indeed, rename the thing.
-	if((implement && implement.loc != user) || !in_range(src, user))
-		stack_trace("Somebody called rename_interactive, but the user cannot rename this.")
+	// This, sadly, means you can't rename things with a telekinetic pen, but that's
+	// too much of a hassle to make work nicely.
+	if((implement && implement.loc != user) || !in_range(src, user) || user.incapacitated(ignore_lying = TRUE))
 		return null
 
 	var/prefix = ""
-	if(!freeform)
+	if(use_prefix)
 		prefix = "[initial(name)] - "
 
 	var/default_value
-	if(freeform)
+	if(!use_prefix)
 		default_value = name
 	else if(findtext(name, prefix) != 0)
-		default_value = copytext(name, length(prefix)+1)
+		default_value = copytext(name, length(prefix) + 1)
 	else
 		// Either the thing has a non-conforming name due to being set in the map
 		// OR (much more likely) the thing is unlabeled yet.
 		default_value = ""
 	if(!prompt)
-		prompt = "What would you like the label on the [name] to be?"
-	var/t = input(user, prompt, "Renaming [name]", default_value)  as text | null
-	if(user)
-		// users could have moved, or lost the pen between input is called and now
-		if((implement && implement.loc != user) || !in_range(src, user))
-			return null
+		prompt = "What would you like the label on [src] to be?"
+
+	var/t = input(user, prompt, "Renaming [src]", default_value)  as text | null
+
+	// Things could have changed between when `input` is called and when it returns.
+	if(!user)
+		return null
+	else if(implement && implement.loc != user)
+		to_chat(user, "<span class='warning'>You no longer have the pen to rename [src].</span>")
+		return null
+	else if(!in_range(src, user))
+		to_chat(user, "<span class='warning'>You cannot rename [src] from here.</span>")
+		return null
+	else if (user.incapacitated(ignore_lying = TRUE))
+		to_chat(user, "<span class='warning'>You cannot rename [src] in your current state.</span>")
+		return null
+
 	if(isnull(t))
 		// user pressed Cancel
 		return null
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -958,3 +958,56 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 		else if(C)
 			color = C
 			return
+
+
+/** Call this when you want to present a renaming prompt to the user.
+
+    It's a simple proc, but handles annoying edge cases such as forgetting to add a "cancel" button,
+	or being able to rename stuff remotely.
+
+	Arguments:
+	* user - the renamer.
+	* implement - the tool doing the renaming (usually, a pen).
+	* freeform - whether we allow to change the name arbitrarily, or if it should go "thing - user-given label".
+	* actually_rename - whether we want to really change the `src.name`, or if we want to do everything *except* that.
+	* prompt - a custom "what do you want rename this thing to be?" prompt shown in the inpit box.
+
+	Returns: Either null if the renaming was aborted, or the user-provided sanitized string.
+ **/
+/atom/proc/rename_interactive(mob/user, obj/implement = null, freeform = FALSE,
+		actually_rename = TRUE, prompt = null)
+	// Sanity check that the user can, indeed, rename the thing.
+	if((implement && implement.loc != user) || !in_range(src, user))
+		stack_trace("Somebody called rename_interactive, but the user cannot rename this.")
+		return null
+
+	var/prefix = ""
+	if(!freeform)
+		prefix = "[initial(name)] - "
+
+	var/default_value
+	if(freeform)
+		default_value = name
+	else if(findtext(name, prefix) != 0)
+		default_value = copytext(name, length(prefix)+1)
+	else
+		// Either the thing has a non-conforming name due to being set in the map
+		// OR (much more likely) the thing is unlabeled yet.
+		default_value = ""
+	if(!prompt)
+		prompt = "What would you like the label on the [name] to be?"
+	var/t = input(user, prompt, "Renaming [name]", default_value)  as text | null
+	if(user)
+		// users could have moved, or lost the pen between input is called and now
+		if((implement && implement.loc != user) || !in_range(src, user))
+			return null
+	if(isnull(t))
+		// user pressed Cancel
+		return null
+	t = sanitize(copytext(t, 1, MAX_NAME_LEN))
+	if(actually_rename)
+		if(t == "")
+			name = "[initial(name)]"
+		else
+			name = "[prefix][t]"
+	return t
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -1000,6 +1000,9 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 		prompt = "What would you like the label on [src] to be?"
 
 	var/t = input(user, prompt, "Renaming [src]", default_value)  as text | null
+	if(isnull(t))
+		// user pressed Cancel
+		return null
 
 	// Things could have changed between when `input` is called and when it returns.
 	if(!user)
@@ -1014,9 +1017,7 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 		to_chat(user, "<span class='warning'>You cannot rename [src] in your current state.</span>")
 		return null
 
-	if(isnull(t))
-		// user pressed Cancel
-		return null
+
 	t = sanitize(copytext(t, 1, MAX_NAME_LEN))
 	if(actually_rename)
 		if(t == "")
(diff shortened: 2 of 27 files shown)


code/game/atoms.dm
```diff
@@ -958,3 +958,56 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+	* freeform - whether we allow to change the name arbitrarily, or if it should go "thing - user-given label".
+	* actually_rename - whether we want to really change the `src.name`, or if we want to do everything *except* that.
+	* prompt - a custom "what do you want rename this thing to be?" prompt shown in the inpit box.
+
+	Returns: Either null if the renaming was aborted, or the user-provided sanitized string.
+ **/
+/atom/proc/rename_interactive(mob/user, obj/implement = null, freeform = FALSE,
+		actually_rename = TRUE, prompt = null)
+	// Sanity check that the user can, indeed, rename the thing.
+	if((implement && implement.loc != user) || !in_range(src, user))
```
moxian, line 980: I could be checking for user being alive, conscious, literate, etc.
I could also be checking that the pen is in active hand (rather than anywhere on the person), as some of the previous code was doing.

I don't do the former because I'm lazy, and the latter because i would *want* to break "realism" for a bit easier gameplay.

farie82: Trying to rewrite things using Telekinesis will now cause runtimes. Most likely not what you'd want to happen

code/game/atoms.dm
```diff
@@ -958,3 +958,56 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+	if(freeform)
+		default_value = name
+	else if(findtext(name, prefix) != 0)
+		default_value = copytext(name, length(prefix)+1)
+	else
+		// Either the thing has a non-conforming name due to being set in the map
+		// OR (much more likely) the thing is unlabeled yet.
+		default_value = ""
+	if(!prompt)
+		prompt = "What would you like the label on the [name] to be?"
```
farie82, line 998:
```suggestion
		prompt = "What would you like the label on [src] to be?"
```
Byond will handle 'the' and such here

code/game/atoms.dm
```diff
@@ -958,3 +958,56 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+
+	var/prefix = ""
+	if(!freeform)
+		prefix = "[initial(name)] - "
+
+	var/default_value
+	if(freeform)
+		default_value = name
+	else if(findtext(name, prefix) != 0)
+		default_value = copytext(name, length(prefix)+1)
```
farie82, line 992:
```suggestion
		default_value = copytext(name, length(prefix) + 1)
```

code/game/objects/items/bodybag.dm
```diff
@@ -27,18 +27,11 @@
...
-		if(!in_range(src, user) && src.loc != user)
-			return
-		t = sanitize(copytext(t,1,MAX_MESSAGE_LEN))
-		if(t)
-			src.name = "body bag - "
-			src.name += t
-			src.overlays += image(src.icon, "bodybag_label")
+		rename_interactive(user, W)
+		if(name == initial(name))
+			src.overlays.Cut()
```
farie82, line 32:
```suggestion
			cut_overlays()
```
calling `cut_overlays()` is preferred

code/game/objects/items/bodybag.dm
```diff
@@ -27,18 +27,11 @@
...
-			return
-		if(!in_range(src, user) && src.loc != user)
-			return
-		t = sanitize(copytext(t,1,MAX_MESSAGE_LEN))
-		if(t)
-			src.name = "body bag - "
-			src.name += t
-			src.overlays += image(src.icon, "bodybag_label")
+		rename_interactive(user, W)
+		if(name == initial(name))
```
farie82, line 31:
```suggestion
		var/t = rename_interactive(user, W)
		if(!t)
```

code/game/objects/items/bodybag.dm
```diff
@@ -27,18 +27,11 @@
...
+			src.overlays.Cut()
 		else
-			src.name = "body bag"
+			src.overlays += image(src.icon, "bodybag_label")
```
farie82, line 34:
```suggestion
			add_overlay(image(icon, "bodybag_label"))
```
`add_overlay` is preferred over manually adding it. Also implied `src`

code/game/objects/structures/morgue.dm
```diff
@@ -125,17 +125,10 @@
...
-		var/t = input(user, "What would you like the label to be?", text("[]", name), null)  as text
-		if(user.get_active_hand() != P)
-			return
-		if((!in_range(src, usr) && loc != user))
-			return
-		t = sanitize(copytext(t,1,MAX_MESSAGE_LEN))
-		if(t)
-			name = text("Morgue- '[]'", t)
+		rename_interactive(user, P)
+		if(name != initial(name))
```
farie82, line 129:
```suggestion
		var/t = rename_interactive(user, P)
		if(t)
```

code/game/objects/structures/morgue.dm
```diff
@@ -125,17 +125,10 @@
...
 			overlays += image(icon, "morgue_label")
```
farie82, line 130: Maybe out of the scope but I think this means that everytime you change the name it adds a new overlay without removing the old one.

code/game/objects/items/bodybag.dm
farie82: Here an overlay is also added each time the body bag changes names

code/game/atoms.dm
```diff
@@ -958,3 +958,56 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+	else
+		// Either the thing has a non-conforming name due to being set in the map
+		// OR (much more likely) the thing is unlabeled yet.
+		default_value = ""
+	if(!prompt)
+		prompt = "What would you like the label on the [name] to be?"
+	var/t = input(user, prompt, "Renaming [name]", default_value)  as text | null
+	if(user)
+		// users could have moved, or lost the pen between input is called and now
+		if((implement && implement.loc != user) || !in_range(src, user))
```
farie82, line 1002: I'd check conciousness and what not here as well. An user can write his dying will. Die. And then keep writing till he's done while dead and it'll all be on the object

code/modules/reagents/reagent_containers/iv_bag.dm
```diff
@@ -202,5 +197,8 @@
 	blood_type = "O-"
 
 /obj/item/reagent_containers/iv_bag/slime
-	name = "\improper IV Bag (Slime Jelly)"
 	list_reagents = list("slimejelly" = 200)
+
+/obj/item/reagent_containers/iv_bag/salglu/Initialize(mapload)
+	..()
```
farie82, line 203:
```suggestion
	. = ..()
```

code/modules/reagents/reagent_containers/iv_bag.dm
```diff
@@ -149,37 +149,32 @@
...
-		set_label(blood_type)
+		name = "[initial(name)] - [blood_type]"
 		reagents.add_reagent("blood", 200, list("donor"=null,"viruses"=null,"blood_DNA"=null,"blood_type"=blood_type,"resistances"=null,"trace_chem"=null))
 		update_icon()
 
 
-/obj/item/reagent_containers/iv_bag/blood/random/New()
+/obj/item/reagent_containers/iv_bag/blood/random/Initialize()
 	blood_type = pick("A+", "A-", "B+", "B-", "O+", "O-")
 	..()
```
farie82, line 179:
```suggestion
	return ..()
```

code/modules/reagents/reagent_containers/iv_bag.dm
```diff
@@ -149,37 +149,32 @@
...
+	..()
+	name = "[initial(name)] - Saline Glucose"
+
 /obj/item/reagent_containers/iv_bag/blood // Don't use this - just an abstract type to allow blood bags to have a common blood_type var for ease of creation.
 	var/blood_type
 	amount_per_transfer_from_this = 5 // Bloodbags are set to transfer 5 units by default.
 
-/obj/item/reagent_containers/iv_bag/blood/New()
+/obj/item/reagent_containers/iv_bag/blood/Initialize(mapload)
 	..()
```
farie82, line 170:
```suggestion
	. = ..()
```

code/modules/reagents/reagent_containers/iv_bag.dm
```diff
@@ -149,37 +149,32 @@
...
+		rename_interactive(user, I)
 
 // PRE-FILLED IV BAGS BELOW
 
 /obj/item/reagent_containers/iv_bag/salglu
-	name = "\improper IV Bag (Saline Glucose)"
 	list_reagents = list("salglu_solution" = 200)
 
+/obj/item/reagent_containers/iv_bag/salglu/Initialize(mapload)
+	..()
```
farie82, line 162:
```suggestion
	. = ..()
```

code/modules/paperwork/pen.dm
```diff
@@ -104,21 +104,6 @@
 	icon_state = "fountainpen"
 	pen_color_shift = 0
 
-/obj/item/pen/attack(mob/living/M, mob/user)
```
farie82, line 107: why is this removed?

code/modules/paperwork/pen.dm
```diff
@@ -128,15 +113,24 @@
...
-		if(reagents.total_volume)
-			if(M.reagents)
-				reagents.trans_to(M, 50)
 
 
-/obj/item/pen/sleepy/New()
+/obj/item/pen/sleepy/Initialize(mapload)
 	create_reagents(100)
 	reagents.add_reagent("ketamine", 100)
 	..()
```
farie82, line 136:
```suggestion
	return ..()
```
Inits return a value saying if they are properly inited or if they need a late init etc.

code/modules/projectiles/gun.dm
```diff
@@ -318,7 +318,9 @@
 
 	if(unique_rename)
 		if(istype(I, /obj/item/pen))
-			rename_gun(user)
+			var/t = rename_interactive(user, I, freeform = TRUE)
+			if(!isnull(t))
+				to_chat(user, "You name the gun [name]. Say hello to your new friend.")
```
farie82, line 323:
```suggestion
				to_chat(user, "<span class='notice'>You name the gun [name]. Say hello to your new friend.</span>")
```

code/game/atoms.dm
```diff
@@ -958,3 +958,56 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+	Arguments:
+	* user - the renamer.
+	* implement - the tool doing the renaming (usually, a pen).
+	* freeform - whether we allow to change the name arbitrarily, or if it should go "thing - user-given label".
+	* actually_rename - whether we want to really change the `src.name`, or if we want to do everything *except* that.
+	* prompt - a custom "what do you want rename this thing to be?" prompt shown in the inpit box.
+
+	Returns: Either null if the renaming was aborted, or the user-provided sanitized string.
+ **/
+/atom/proc/rename_interactive(mob/user, obj/implement = null, freeform = FALSE,
```
farie82, line 977:
```suggestion
/atom/proc/rename_interactive(mob/user, obj/implement = null, use_prefix= TRUE,
```
`freeform` doesn't tell that much about what it does. This change does require an invert of all `freeform` usages

farie82 (changes requested):

code/modules/paperwork/pen.dm
moxian: This is arguably out of scope, but when I was originally making this PR i had the renaming proc be a method on the pen. And figured that the `/attack` logic is horrid, and only exists to facilitate sleepy pen injections. There's no good reason for why it is not under `pen/sleepy/attack()`.
Sleepy pen still works; attacking people with regular pen no longer gives "you stab a guy with a pen" messages that only you (but not your victim!) and admins(!) can see.

code/ATMOSPHERICS/components/binary_devices/pump.dm
```diff
@@ -237,12 +237,7 @@ Thus, the two variables affect pump operation are set in New():
 
 /obj/machinery/atmospherics/binary/pump/attackby(obj/item/W, mob/user, params)
 	if(istype(W, /obj/item/pen))
-		var/t = copytext(stripped_input(user, "Enter the name for the pump.", "Rename", name), 1, MAX_NAME_LEN)
-		if(!t)
-			return
-		if(!in_range(src, usr) && loc != usr)
-			return
-		name = t
+		rename_interactive(user, W)
```
moxian, line 240: In the interest of transparency, I should probably note that this (and other atmos machinery) are technically a behavior change, since I'm now forcing the new name to be "gas pump - N2O", with "gas pump" being mandatory.
The old behavior, where you could name the pump just "N2O" can be easily achieved by passing in `use_prefix = FALSE`, but I don't think this is actually needed.

code/modules/paperwork/pen.dm
farie82: Sounds logical. Can you add this to the PR description and CL? Good change btw

code/modules/reagents/reagent_containers/iv_bag.dm
```diff
@@ -202,5 +197,8 @@
 	blood_type = "O-"
 
 /obj/item/reagent_containers/iv_bag/slime
-	name = "\improper IV Bag (Slime Jelly)"
 	list_reagents = list("slimejelly" = 200)
+
+/obj/item/reagent_containers/iv_bag/salglu/Initialize(mapload)
```
farie82, line 202:
```suggestion
/obj/item/reagent_containers/iv_bag/slime/Initialize(mapload)
```

code/game/atoms.dm
```diff
@@ -958,3 +958,69 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+		to_chat(user, "<span class='warning'>You no longer have the pen to rename [src].</span>")
+		return null
+	else if(!in_range(src, user))
+		to_chat(user, "<span class='warning'>You cannot rename [src] from here.</span>")
+		return null
+	else if (user.incapacitated(ignore_lying = TRUE))
+		to_chat(user, "<span class='warning'>You cannot rename [src] in your current state.</span>")
+		return null
+
+	if(isnull(t))
```
farie82, line 1017: I feel like this should be tested before the rest

code/modules/reagents/reagent_containers/iv_bag.dm
moxian: I feel so bad about this right now.. (Fixed)

code/game/atoms.dm
moxian: I don't have a strong opinion here, but sure.

code/modules/paperwork/pen.dm
moxian: Done.

farie82 (approved):
LGTM

PR #14965, opened 2020-11-22
Gives Shadowlings a Melt Mindshield ability

ppi13:
## What Does This PR Do

_sigh_, this is an alternative to #14956
After #14521 was added over a month ago, we got a severe case of I ded pls nerf, where shadowlings lost every round, which honestly, was fine for me. However, on the forum there was a thread for this as well, and there was a pretty good suggestion, where destroying mind shields is part of the progressive powerup.  
Currently it takes 30 seconds to destroy an implant with a 90 second cooldown.
In an average round of 100 players a sling will learn this ability at somewhere between 6-7 thralls.

## Why It's Good For The Game

This is a better alternative to reverting #14521
This should also halt shadowling turning into murderboner security and then shadowlings lose. 

## Changelog
:cl: ppi
add: Add a mindshield melt ability to shadowlings
/:cl:


SabreML:
image
:eyes:

AffectedArc07:
Not in favour:
1) You proceeded to bait people with your PR title and commit, even though that's not what this PR does (Commits are the *last* thing you enter after the work is done, so there's no excuse here)
2) Your code isn't even tested, and doesn't even compile, which doesn't give us hope in this implementation
3) This still wont solve the issue. By the time slings get to 15 thralls, they can effectively snowball anyway from pure numbers. Taking only 14 seconds to remove a mindshield before subsequently thralling an officer doesn't matter if you're in a secluded corner of maintenance without your victim having a headset.

Aurorablade:
Honestly the revert of it going back to old methods of extra time to remove a mindshield is the better solution here so yeah fully with a. Affected. We already have antags that can snowball out of control if left alone long enough. Now back to killing my arms st work

ppi13:
- Fixes the merge conflicts
- Decrease the amount of thralls needed to gain access to mindshield melt
- Adjusts the timer for mindshield removal up

ppi13:
Closed due to merge conflicts

PR #14971, opened 2020-11-22
[TGUI] Airlock Electronics

SabreML:
## What Does This PR Do

Converted the interface for Airlock Electronics from Browser UI to TGUI.
I'm still not 100% sure about the design, so any feedback would be appreciated.

*(Marked as 'Needs Review' since this is my first time coding Javascript)*

## Why It's Good For The Game

Easier to use, better looking, and it runs faster.

## Images of changes

**Old:**
airlock_1

**New:**
c4xJREfokL

## Changelog
:cl:
tweak: TGUIified the Airlock Electronics UI
/:cl:


SabreML:
The main design issue I have currently is the 'Unrestricted Access' section.

If you selected 'North' for example, then anyone entering the airlock from the north will be allowed through regardless of their access level.
The issue is that the icons might be confusing since the arrow for the 'North' button is pointing northwards, implying that it's unrestricted from the south.

So, two options:
Option 1: (Arrows point in the correct direction, but opposite to the direction of travel)
option 1
Option 2: (Arrows point in the opposite direction, but are the same as the direction of travel)
option 2

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -3,100 +3,102 @@
...
-	if(usr.incapacitated() || (!ishuman(usr) && !isrobot(usr)))
-		return 1
-
-	if(href_list["close"])
-		usr << browse(null, "window=airlock_electronics")
+	tgui_interact(user)
+
+
+// tgui\packages\tgui\interfaces\AirlockElectronics.js
+/obj/item/airlock_electronics/tgui_interact(mob/user, ui_key, datum/tgui/ui, force_open, datum/tgui/master_ui, datum/tgui_state/state)
```
AffectedArc07, line 43: This is missing all of the proper parameters and their definitions (EG: No state specified). Please look at another implemented TGUI DM side for a good example.

tgui/packages/tgui/interfaces/AirlockElectronics.js
```diff
@@ -0,0 +1,111 @@
...
+    <Section title="Access Control">
+      <Flex direction="column">
+        <Flex.Item bold mb={1}>
+          Unrestricted Access Settings
+        </Flex.Item>
+        <Flex.Item>
+          <Button width={5.7}
+            icon="arrow-up"
+            content="North"
+            selected={data.unrestricted_dir === 1 ? "selected" : null}
```
AffectedArc07, line 33:
```suggestion
            selected={unrestricted_dir === 1 ? "selected" : null}
```
`data.` not required for these since you did the override up here
```js
	  const {
    unrestricted_dir,
  } = data;
```

tgui/packages/tgui/interfaces/AirlockElectronics.js
```diff
@@ -0,0 +1,111 @@
...
+            onClick={() => act('unrestricted_access', {
+              unres_dir: 1,
+            })}
+          />
+          <Button width={5.7}
+            icon="arrow-down"
+            content="South"
+            selected={data.unrestricted_dir === 2 ? "selected" : null}
```
AffectedArc07, line 41: Same for all the others

tgui/packages/tgui/interfaces/AirlockElectronics.js
```diff
@@ -0,0 +1,111 @@
...
+    one_access,
+    regions,
+  } = data;
+  return (
+    <AccessList
+      usedByRcd={1}
+      rcdButtons={
+        <Fragment>
+          <Button.Checkbox
+            checked={data.one_access}
```
AffectedArc07, line 81: Applies here too

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -3,100 +3,102 @@
...
-	var/const/max_brain_damage = 60 // Maximum brain damage a mob can have until it can't use the electronics
-	var/unres_sides = 0
+	/// List of accesses currently set
+	var/list/selected_accesses = list()
+	/// Is the door access using req_one_access (TRUE) or req_access (FALSE)
+	var/one_access = FALSE
+	/// An associative list containing all station accesses. Includes their name and access number.
+	var/static/list/door_accesses_list = list()
+	/// Maximum brain damage a mob can have until it can't use the electronics
+	var/const/max_brain_damage = 60
```
AffectedArc07, line 18: I feel like theres a central define for a brain damage threshold somewhere. That or I have brain damage.

AffectedArc07 (changes requested):

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -3,100 +3,102 @@
...
+	/// Which direction has unrestricted access (e.g. medbay doors from the inside)
 	var/unres_direction = null
```
moxian, line 20: Does not explain which direction you can come FROM or which direction you can go TOWARDS.
Definitely needs a better comment, but would be great if the name of the var had an indication of this (although the latter might be out of scope).

code/game/machinery/doors/airlock_electronics.dm
SabreML: If there is one, I can't find it.

moxian: There is none but `60` features rather consistently as "you do something dumb if you have more than this much brain damage" across the codebase. So defining that might not be a bad idea.

tgui/packages/tgui/interfaces/AirlockElectronics.js
```diff
@@ -0,0 +1,111 @@
...
+            content="One"
```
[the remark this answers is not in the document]
AffectedArc07: One makes more sense here. `Any` could mean `Any access` which would imply the access restrictions mean nothing, imo

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -3,100 +3,102 @@
...
 
+/obj/item/airlock_electronics/Initialize(mapload)
+	. = ..()
+	if(!length(door_accesses_list))
+		for(var/access in get_all_accesses())
+			door_accesses_list += list(list(
```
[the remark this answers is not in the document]
SabreML: Currently this is set to static because the list of accesses will never actually change, so there's no reason to have it update more than once.

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -3,100 +3,102 @@
...
+	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
+	if(!ui)
+		ui = new(user, src, ui_key, "AirlockElectronics", name, 450, 575, master_ui, state)
+		ui.open()
+
+/obj/item/airlock_electronics/tgui_data(mob/user)
+	var/list/data = list()
+	data["selected_accesses"] = selected_accesses
+	data["one_access"] = one_access
+	data["unrestricted_dir"] = unres_direction
```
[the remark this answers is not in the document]
SabreML: I think the most I can do here is change `unres_dir: 1,` to `unres_dir: "North",` in the js code, since the actual airlock itself needs the directions as a number.

tgui/packages/tgui/interfaces/AirlockElectronics.js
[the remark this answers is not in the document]
SabreML: The biggest reason why I didn't change this to something else is because the actual airlock uses the `req_one_access` and `req_access` vars, and changing those to fit with 'Any' would need map edits.

code/game/machinery/doors/airlock_electronics.dm
SabreML: How's `unres_access_from` for a var name?
I've already touched airlock code so it's within scope.

code/game/machinery/doors/airlock_electronics.dm
SabreML: Wouldn't defining a global brain damage threshold be a bit out of scope?

code/game/machinery/doors/airlock_electronics.dm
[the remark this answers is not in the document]
moxian: > `selected={unrestricted_dir === 1 ? "selected" : null}` would need to stay the same.

If you change `data["unrestricted_dir"] = unres_direction` to `data["unrestricted_dir"] = dir2text(unres_direction)`, then you'd be able to `selected={unrestricted_dir === "north" ? "selected" : null}`

code/game/machinery/doors/airlock_electronics.dm
moxian: `_from` is good. It's not super eloquent of course, but it's sufficiently unconfusing.

procdrone:
option 2 is far more user friendly hence superior

code/game/machinery/doors/airlock_electronics.dm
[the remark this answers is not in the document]
SabreML: ahh, yep. Forgot I could change that part

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -3,100 +3,102 @@
...
-			if(!conf_access.len)
-				conf_access = null
+	. = TRUE
+	// Mostly taken from the RCD code
+	switch(action)
+		if("unrestricted_access")
+			var/direction = text2num(params["unres_dir"])
+			if(direction == unres_direction)
+				unres_direction = null // Deselecting
+				return
```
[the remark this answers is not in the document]
SabreML: I don't really see what's wrong with this one myself.
```
if(user clicked on already selected button)
    deselect button
```

tgui/packages/tgui/interfaces/AirlockElectronics.js
```diff
@@ -0,0 +1,111 @@
...
+  const { act, data }= useBackend(context);
+  const {
+    selected_accesses,
+    one_access,
+    regions,
+  } = data;
+  return (
+    <AccessList
+      usedByRcd={1}
+      rcdButtons={
```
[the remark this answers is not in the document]
SabreML: I'd definitely like to change it to something else, but I really can't think of anything.

tgui/packages/tgui/interfaces/AirlockElectronics.js
```diff
@@ -0,0 +1,111 @@
...
+    unrestricted_dir,
+  } = data;
+  return (
+    <Section title="Access Control">
+      <Flex direction="column">
+        <Flex.Item bold mb={1}>
+          Unrestricted Access From:
+        </Flex.Item>
+        <Flex.Item>
+          <Button width={5.7}
```
AffectedArc07, line 30: I would put these inside a `<Grid>` element and use `<Grid.Column>` to house each button, then add the `fluid` property to each button to make them fit the full width of the container, which would make them all equal width and fill a full space, looking a bit nicer in the UI

SabreML: I think I considered doing that a while ago but I went with `<Flex>` instead because it says this in the [Component Reference](https://github.com/ParadiseSS13/Paradise/blob/master/tgui/docs/component-reference.md#grid).
> Deprecated: This component is no longer recommended due to the variety of bugs that come with table-based layouts. We recommend using Flex instead.

I'll try it out anyway though, no reason not to.

SabreML: I've messed around with `<Grid>` a bit and I think I'll keep it as it is. While there *is* a blank space on the right side I personally prefer that to the buttons taking up the whole row. It just feels better, I suppose.

SteelSlayer (approved):
LGTM.

AffectedArc07:
Could you deconflict and update images please

AffectedArc07 (approved):

PR #14972, opened 2020-11-22
Mechs must now have maint protocols active to be repaired.

Qwertytoforty:
## What Does This PR Do

This PR adds a check to welder_act on mech code, requiring maintenance protocols to be active in order for a mech to be repaired via welder.

## Why It's Good For The Game

Mechs are strong, perhaps too strong, and could use some nerfs. This is a simple change, that forces a mech to be vulnerable, and unable to move or fire, when getting welder repaired. This prevents crew from fighting blob with a mech and constantly welding it, making it effectively an un-killable infinitely firing turret. Instead, the mech must retreat a distance to be fixed and stop firing, or be repaired on the front lines and vulnerable.

## Changelog
:cl:
tweak: Mechs must now have maintenance protocols active in order to be repaired with a welder.
/:cl:

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -840,7 +840,7 @@
 	if(repairing)
 		to_chat(user, "<span class='notice'>[src] is currently being repaired!</span>")
 		return
-	if(state == 0)
+	if(state == 0) // If maint protocols are not active, the state is zero
 		to_chat(user, "<span class='warning'>[src] can not be repaired without maintenance protocols active!</span>")
 		return
 	WELDER_ATTEMPT_REPAIR_MESSAGE
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -840,6 +840,9 @@
 	if(repairing)
 		to_chat(user, "<span class='notice'>[src] is currently being repaired!</span>")
 		return
+	if(state < 1)
+		to_chat(user, "<span class='notice'>[src] can not be repaired without maintenance protocols active!</span>")
+		return
 	WELDER_ATTEMPT_REPAIR_MESSAGE
 	repairing = TRUE
 	if(I.use_tool(src, user, 15, volume = I.tool_volume))
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -841,7 +841,7 @@
 		to_chat(user, "<span class='notice'>[src] is currently being repaired!</span>")
 		return
 	if(state == 0)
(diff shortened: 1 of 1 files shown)


code/game/mecha/mecha.dm
```diff
@@ -840,6 +840,9 @@
 	if(repairing)
 		to_chat(user, "<span class='notice'>[src] is currently being repaired!</span>")
 		return
+	if(state < 1)
```
moxian, line 843:
```suggestion
	if(state == 0)
```
The state numbers are just arbitrary unordered magic numbers. It does not make sense to compare them for greater than/less than, just as it doesn't make sense to compare `"normal mode"` `"maintenance mode"` `"power cell removed"` etc. via `<`.

procdrone:
cannot fight while repairing, sounds good. if damaged, the pilot will be forced to either retreat and fix or take a gamble and continue fighting despite damage and risk loosing the mech.

TDSSS:
It might be worth thinking about buffing nukie mechs, or making them cheaper, along with this change. They're already not very viable or used very often, and this does make them even less attractive of a purchase.

code/game/mecha/mecha.dm
```diff
@@ -840,6 +840,9 @@
 	if(repairing)
 		to_chat(user, "<span class='notice'>[src] is currently being repaired!</span>")
 		return
+	if(state == 0)
```
SteelSlayer, line 843: I think you want to use `maint_access` here for clarity. If its `1`, it means maint protocols are on, `0` is off. Mech code is so convoluted, it's hard to tell what state 0 - 4 are suppose to do at a glance.

SteelSlayer (approved):

Kyep (approved):

code/game/mecha/mecha.dm
farie82, line 843: Ah yes state... It'll be out of scope to refactor this into defines but can you place a comment saying what the state means? Makes it easier for later

code/game/mecha/mecha.dm
```diff
@@ -840,6 +840,9 @@
...
+		to_chat(user, "<span class='notice'>[src] can not be repaired without maintenance protocols active!</span>")
```
farie82, line 844:
```suggestion
		to_chat(user, "<span class='warning'>[src] can not be repaired without maintenance protocols active!</span>")
```
I'd make this a warning level message

farie82 (changes requested):

code/game/mecha/mecha.dm
Qwertytoforty: Done!

farie82 (approved):
LGTM

PR #14977, opened 2020-11-23
Allows nanopaste to repair machinery

Qwertytoforty:
## What Does This PR Do

This PR lets nanopaste be used on subtypes of obj/machinery to repair 50 object damage each use, at the cost of one nanopaste, so long as the object is not broken. Takes 3 seconds.

## Why It's Good For The Game

When something is damaged in game currently... it's just damaged. You can't fix it unless you deconstruct the item or in certain cases, preform an action (IE fixing wires on a camera). This is very annoying on things like the RND servers, apcs, or protolathe, where deconstruction risks levels, require re-linking, disables an area for a long period of time, or destroys resources inside. By the addition of allowing nanopaste to be used on machinery, this allows people to repair important structures without disabling an area, or loosing resources. 3 second cooldown prevents spamming on structures to fast repair.

## Changelog
:cl:
add: Nanopaste can now be used on machinery to heal 50 object damage per use.
/:cl:

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -480,24 +480,27 @@ Class Procs:
 
 /obj/machinery/attackby(obj/item/O, mob/user, params)
 	if(istype(O, /obj/item/stack/nanopaste))
-		if(stat && BROKEN)
-			to_chat(user, "<span class='notice'>\The [src] is too damaged to be fixed with nanopaste!</span>")
+		if(stat & BROKEN)
+			to_chat(user, "<span class='notice'>[src] is too damaged to be fixed with nanopaste!</span>")
 			return
 		if(obj_integrity == max_integrity)
-			to_chat(user, "<span class='notice'>\The [src] is fully intact.</span>")
+			to_chat(user, "<span class='notice'>[src] is fully intact.</span>")
 			return
 		if(!being_repaired)
-			to_chat(user, "<span class='notice'>You start applying the [O] to the [src].</span>")
+			to_chat(user, "<span class='notice'>You start applying [O] to [src].</span>")
 			being_repaired = TRUE
 			var/result = do_after(user, 3 SECONDS, target = src)
 			being_repaired = FALSE
 			if(result)
 				var/obj/item/stack/nanopaste/N = O
 				obj_integrity = clamp((obj_integrity + 50), obj_integrity, max_integrity)
-				N.use(1)
+				if(!N.use(1))
+					to_chat(user, "<span class='warning'>You don't have enough to complete this task!</span>")
+					return
 				user.visible_message("<span class='notice'>[user.name] applied some [O] at [src]'s damaged areas.</span>",\
 					"<span class='notice'>You apply some [O] at [name]'s damaged areas.</span>")
-
+	else
+		return ..()
 /obj/machinery/proc/exchange_parts(mob/user, obj/item/storage/part_replacer/W)
 	var/shouldplaysound = 0
 	if((flags & NODECONSTRUCT))
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -495,11 +495,11 @@ Class Procs:
 		if(!result)
 			return
 		var/obj/item/stack/nanopaste/N = O
-		obj_integrity = clamp((obj_integrity + 50), obj_integrity, max_integrity)
+		obj_integrity = min(obj_integrity + 50, max_integrity)
 		if(!N.use(1))
 			to_chat(user, "<span class='warning'>You don't have enough to complete this task!</span>")
 			return
-		user.visible_message("<span class='notice'>[user.name] applied some [O] at [src]'s damaged areas.</span>",\
+		user.visible_message("<span class='notice'>[user] applied some [O] at [src]'s damaged areas.</span>",\
 			"<span class='notice'>You apply some [O] at [src]'s damaged areas.</span>")
 	else
 		return ..()
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -480,6 +480,7 @@ Class Procs:
 
 /obj/machinery/attackby(obj/item/O, mob/user, params)
 	if(istype(O, /obj/item/stack/nanopaste))
+		var/obj/item/stack/nanopaste/N = O
 		if(stat & BROKEN)
 			to_chat(user, "<span class='notice'>[src] is too damaged to be fixed with nanopaste!</span>")
 			return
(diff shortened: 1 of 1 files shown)


Drobe50:
actually a good change holy moly, you deserve a medal

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,26 @@ Class Procs:
...
+			return
+		if(!being_repaired)
+			to_chat(user, "<span class='notice'>You start applying the [O] to the [src].</span>")
+			being_repaired = TRUE
+			var/result = do_after(user, 3 SECONDS, target = src)
+			being_repaired = FALSE
+			if(result)
+				var/obj/item/stack/nanopaste/N = O
+				obj_integrity = clamp((obj_integrity + 50), obj_integrity, max_integrity)
+				N.use(1)
```
farie82, line 497:
```suggestion
				if(!N.use(1))
					to_chat(user, "<span class='warning'>You don't have enough to complete this task!</span>")
					return
```
or something the like. Should not occur with normal nanopaste. But if say a borg tries it without energy then it'll succeed even though it shouldn't

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,26 @@ Class Procs:
 	if(.)
 		power_change()
 
+/obj/machinery/attackby(obj/item/O, mob/user, params)
+	if(istype(O, /obj/item/stack/nanopaste))
+		if(stat && BROKEN)
```
farie82, line 483:
```suggestion
		if(stat & BROKEN)
```
This should be a bitwise operator version of and. `1101 & 1000 = 1000`

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,26 @@ Class Procs:
...
+			return
+		if(obj_integrity == max_integrity)
+			to_chat(user, "<span class='notice'>\The [src] is fully intact.</span>")
+			return
+		if(!being_repaired)
+			to_chat(user, "<span class='notice'>You start applying the [O] to the [src].</span>")
+			being_repaired = TRUE
+			var/result = do_after(user, 3 SECONDS, target = src)
+			being_repaired = FALSE
+			if(result)
```
farie82, line 494: Same here

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,26 @@ Class Procs:
...
+				var/obj/item/stack/nanopaste/N = O
+				obj_integrity = clamp((obj_integrity + 50), obj_integrity, max_integrity)
+				N.use(1)
+				user.visible_message("<span class='notice'>[user.name] applied some [O] at [src]'s damaged areas.</span>",\
```
farie82, line 498: I think this will show `at the NAMEs` for the first message and `at NAMEs` for the second due to how byond works

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,26 @@ Class Procs:
...
+					"<span class='notice'>You apply some [O] at [name]'s damaged areas.</span>")
+
```
farie82, line 500:
```suggestion
	else
		return ..()
```
You can't destroy machines now using weapons due to this lacking

farie82 (changes requested):

Kyep (approved):

code/game/machinery/machinery.dm
Qwertytoforty: done

code/game/machinery/machinery.dm
Qwertytoforty: adjusted

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,26 @@ Class Procs:
...
 
+/obj/machinery/attackby(obj/item/O, mob/user, params)
+	if(istype(O, /obj/item/stack/nanopaste))
+		if(stat && BROKEN)
+			to_chat(user, "<span class='notice'>\The [src] is too damaged to be fixed with nanopaste!</span>")
+			return
+		if(obj_integrity == max_integrity)
+			to_chat(user, "<span class='notice'>\The [src] is fully intact.</span>")
+			return
+		if(!being_repaired)
```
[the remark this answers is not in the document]
Qwertytoforty: done

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,31 @@ Class Procs:
...
+		if(being_repaired)
+			return
+		to_chat(user, "<span class='notice'>You start applying [O] to [src].</span>")
+		being_repaired = TRUE
+		var/result = do_after(user, 3 SECONDS, target = src)
+		being_repaired = FALSE
+		if(!result)
+			return
+		var/obj/item/stack/nanopaste/N = O
+		obj_integrity = clamp((obj_integrity + 50), obj_integrity, max_integrity)
```
farie82, line 498: This needs to be placed after the `use`. Else you can have the situation you can't use it yet will heal

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,31 @@ Class Procs:
...
+		if(stat & BROKEN)
+			to_chat(user, "<span class='notice'>[src] is too damaged to be fixed with nanopaste!</span>")
+			return
+		if(obj_integrity == max_integrity)
+			to_chat(user, "<span class='notice'>[src] is fully intact.</span>")
+			return
+		if(being_repaired)
+			return
+		to_chat(user, "<span class='notice'>You start applying [O] to [src].</span>")
+		being_repaired = TRUE
```
farie82, line 492: an `get_amount()` check here would also be handy. To check if you have enough to do the job. Again will only realistically happen with borg modules

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,31 @@ Class Procs:
...
+		var/result = do_after(user, 3 SECONDS, target = src)
+		being_repaired = FALSE
+		if(!result)
+			return
+		var/obj/item/stack/nanopaste/N = O
+		obj_integrity = clamp((obj_integrity + 50), obj_integrity, max_integrity)
+		if(!N.use(1))
+			to_chat(user, "<span class='warning'>You don't have enough to complete this task!</span>")
+			return
+		user.visible_message("<span class='notice'>[user.name] applied some [O] at [src]'s damaged areas.</span>",\
```
farie82, line 502:
```suggestion
		user.visible_message("<span class='notice'>[user] applied some [O] at [src]'s damaged areas.</span>",\
```

code/game/machinery/machinery.dm
farie82:
```suggestion
		obj_integrity = min(obj_integrity + 50, max_integrity)
```
It can also be simplified since `obj_integrity + 50` can never be lower than `obj_integrity`

code/game/machinery/machinery.dm
Qwertytoforty: Done

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,34 @@ Class Procs:
...
+		being_repaired = TRUE
+		var/result = do_after(user, 3 SECONDS, target = src)
+		being_repaired = FALSE
+		if(!result)
+			return
+		var/obj/item/stack/nanopaste/N = O
+		obj_integrity = min(obj_integrity + 50, max_integrity)
+		if(N.get_amount() < 1)
+			to_chat(user, "<span class='warning'>You don't have enough to complete this task!</span>")
+			return
```
SteelSlayer, line 501: I'd say it makes a bit more sense to put this on line 490, before the `"You start applying [O] to [src]."` message.

code/game/machinery/machinery.dm
```diff
@@ -476,6 +478,34 @@ Class Procs:
...
+		if(being_repaired)
+			return
+		to_chat(user, "<span class='notice'>You start applying [O] to [src].</span>")
+		being_repaired = TRUE
+		var/result = do_after(user, 3 SECONDS, target = src)
+		being_repaired = FALSE
+		if(!result)
+			return
+		var/obj/item/stack/nanopaste/N = O
+		obj_integrity = min(obj_integrity + 50, max_integrity)
```
farie82, line 498: And these two on line 505. So it won't heal when it shouldn't heal

code/game/machinery/machinery.dm
Qwertytoforty: done!

code/game/machinery/machinery.dm
Qwertytoforty: Done!

PR #14981, opened 2020-11-24
Adds a "follow" button to things pointed to by ghosts

dearmochi:
## What Does This PR Do

Adds follow links to things ghosts point to. Additional buttons may appear depending on the pointed thing:
- For AIs with a client, the eye can be followed.
- For other observers, their body if any

## Why It's Good For The Game

Makes it easier to track what a ghost is looking at and pointing to, rather than having to follow the pointer then find the pointed atom or going through the menu which can be a bit populated sometimes.

## Images of changes

image

## Changelog
:cl:
add: Add a follow button to things pointed to by ghosts
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -739,9 +739,9 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 //overriden here and in /mob/living for different point span classes and sanity checks
 /mob/dead/observer/pointed(atom/A as mob|obj|turf in view())
 	if(!..())
-		return 0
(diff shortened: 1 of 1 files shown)


TDSSS:
If ghosts are visible, do living players see the follow link?

dearmochi:
@TDSSS 

They would indeed - good catch. The follow button will no longer appear if the ghost has no `invisibility` set, which would be equal to being visible to living mobs

farie82 (approved):

SteelSlayer (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #14953, opened 2020-11-21
Greatly reduces Slot machine money abuse

Darkmight9:
## What Does This PR Do
Greatly reduces the amount of money you can win on the slot machines and raises the buy in price to 50 credits

## Why It's Good For The Game
The slot machines have been a long standing issue with getting an economy going on paradise since it's basically free money with enough patience and clicking. Now the most you'll be able to win on the slots is 10,000 which is no small amount but won't completely bust the economy

## Changelog
:cl:
tweak: Reduced how much you can win on the slot machines and raise the buy in price to 50
/:cl:

PR #14988, opened 2020-11-26
Adds the Synth to Loadout.

Mitchs98:
## What Does This PR Do
Simply put this adds the Piano Synthesizer to the loadout screen, allowing players to select it for 2 loadout points. A sizeable chunk of your load-out points in exchange for a nice fluff item that's actually fairly fun to use. I contemplated adding the banjo, but it's technically a pretty decent weapon so I didn't want this to incur any sort of balance discussions or whatever.

## Why It's Good For The Game
Not every round are you able to afford a synthesizer, and rounds you aren't are often times rounds you just want to chill and play some good music. I have this happen frequently enough to myself that I assume it's a problem for others. The option to buy it is still there, but this way people will be able to choose between a fashion choice or the thing they wanted to spend all round doing. Less of an impact on Donors, but a decent decision to be made for non-donors.

## Changelog
:cl:
tweak: Added the Synthesizer to the Loadout section of character creation for 2 points.
/:cl:

PR #14989, opened 2020-11-26
Makes starting funds amount spread a bit tighter.

moxian:
## What Does This PR Do
~Increase minimum starting funds from 500 to 2500.~
Reduces the starting funds rng from 500-5000 to 2500-5000

## Why It's Good For The Game
With ~#14982~ #14953 being merged, there is now no way to make money.
While we wait for "real economy"  to arrive, let's temporarily increase the starting funds so that people can afford to buy a synth.

## Changelog
:cl:
tweak: You are now guaranteed to start the round with at least 2500cr balance
/:cl:

### DIFF
--- a/code/controllers/subsystem/jobs.dm
+++ b/code/controllers/subsystem/jobs.dm
@@ -585,7 +585,7 @@ SUBSYSTEM_DEF(jobs)
 
 
 /datum/controller/subsystem/jobs/proc/CreateMoneyAccount(mob/living/H, rank, datum/job/job)
-	var/datum/money_account/M = create_account(H.real_name, rand(50,500)*10, null)
+	var/datum/money_account/M = create_account(H.real_name, rand(250,500)*10, null)
 	var/remembered_info = ""
 
 	remembered_info += "<b>Your account number is:</b> #[M.account_number]<br>"


Shockpoint:
I am not quite sure how that PR prevents making money. Please elaborate.

moxian:
My bad, I meant #14953
Spinning slot machines used to - on average - give you money. I.e. you are expected to win more than what you spend per spin. Slot machines being able to make money, and it being bad for the game was the entire reason for for #14953 
So now that this is reversed, and you are expected to *lose* money on slot machines. And if you start with 500cr you cannot afford a synthesizer.

Kyep:
Objection.

500 is plenty. That's enough to buy food, which is the only thing that costs money.

If anything, 500 is probably too generous. 

When we start having a real economy, I'd be tempted to have everyones' accounts start out at 100... and only be filled through the shift as payroll is processed.

moxian:
Then let's decrease synth price down from its current 2000, yes?

When we start having real economy - sure, we can have everyone start with 100, but until that day comes, let's not deny people fluff items.

That  or.. let's have everyone start with 500? 2500 might be a lot, but it's still much lower than the 5000 maximum that you get. Being either able or not able to buy fluff depending on roundstart rng feels bad.

Shockpoint:
Ah, that explains that.

I like this then.

Shockpoint:
> Objection.
> 
> 500 is plenty. That's enough to buy food, which is the only useful thing that costs money.
> 
> If anything, 500 is probably too generous. 
> 
> When we start having a real economy, I'd be tempted to have everyones' accounts start out at 100... and only be filled through the shift as payroll is processed.

This isn't feasible. There are plenty of items that require a purchase, that cost more than 500. Food might be essential, but we're talking about items that promote RP here. We should be very interested in this, or otherwise slash those prices.

If there is no way to make money outside of being a roboticist or whatever, then we need something like this to make up for it so people can still buy things.

Kyep:
Please list the items that are useful and cost more than 500c.

moxian:
PDA: 300
Musical instruments: 500 (piano synth being an exception at 1000)
Arcade claw game rewards. 
For arcade: quick experimentation seems to put my win chance at ~30%. That means a single prize ball costs ~50cr. Which puts the cost of a random single non-carp plushie at ~250cr, and a specific non-carp plushie (say, a deer) at ~7500. (slightly less if we also spend tickets on the plushie - ~200cr for random and ~5850cr for a specific).

Yes, those items are not "useful" as in, they serve no mechanical purpose. They are fluff. Never said they aren't.
People would still really enjoy having those in a round for RP purposes, and they are important to be affordable.

TDSSS:
> 3) asking someone else to lend you (which is not really "making" money)

Why just lend? You could always convince them to pay you by doing something for them, or giving them something for it. Or heck, convince the captain to pay you a bonus for something out of the station account. There is plenty of money in a round, it might just take some effort to get.

moxian:
Look, the biggest issue i'm trying to solve here is that *whether or not you get to play music is up to roundstart RNG*.
If you want to have everyone start with 100cr, and have them sell their goods and services to other in order to finally acquire a guitar after two hours of ~RP~ work - well, I won't say I'll be happy with that, but that'd be understandable and consistent.  However, as it is right now, the inequality in the funds just feels.. unfair.

I just want my heckin synth, and I don't want to rely on #14988 as the only reliable easy way to get one.

Spacemanspark:
I'd be fine with slashing a few prices so people can still buy things.

moxian:
I would be happy with reducing the prices as an alternative to increasing starting funds (since that's more or less equivalent)
Although I still feel that the current order of magnitude difference in starting funds (500 - 5000) is unhealthy even for the "proper" economy, and needs to be addressed 

What about:
- All the items in the cargo shop get price reduced 2x (so guitar costs 250cr, synth costs 500cr, banhammer costs 1000cr)
- Claw game round costs 5cr (down from 15cr)
- Starting funds spread is reduced to 500cr-1000cr (there is still randomness but it's no longer 10x)
?

Denthamos:
Really you can just look at VG for a functioning example of a working economy where everyone starts with less money but gets paid as the round progress's, and even cargo runs off said money as well.  If we are going to open this can of worms about money and economies.

moxian:
I'm not willing to implement a full economy.
I'm just trying to address the specific problem that arose from #14953 that was done "in preparation" for the economy. But to my knowledge @Darkmight9 is not working on the full economy thing either.
And, again, *once* we have The Ecomony, I'm totally open to completely reverting this PR. This is just a short-term band-aid for until The Economy is actually a thing.

Necaladun:
If there's an issue with individual items that should be cheaper to be more accessible, I'd prefer to slash their prices.

Increasing the amount of money everyone always has will devalue money itself. This will make interesting RP opportunities less valuable - such as traitor hops embezzling station funds, or the briefcase of cash. 

I'd prefer money is limited to the degree where it is possible to run out more easily, so people actually go out looking for money.

Fox-McCloud:
I'm not particularly opposed to this--money doesn't really mean/do anything, and never really has, despite prices being tinkered up or down, in the past.

The slot machine didn't really alter the dynamic, as earning more money wasn't even a thing, for years, even before the introduction of the slot machine and mining money; it was really only R&D+robotics that could earn money.

To get money to be "valuable", you pretty much have to *force* everyone to use it and make life without using it miserable (this has largely been TG's approach). Even the times where bartenders or chefs charge money, consistently, it usually always leads to people getting quite upset---imagine extending this out to medical treatment, getting thrown in jail, or science printing you things. TG's system really isn't "emergent" in the sense real money is---it's more they lock key things for actually *being able to do your job* behind a money sink, then force that feedback loop to create conflict with the wider playerbase....it's...interesting, but really combative.

It'll accomplish an "economy" of sorts, but I'm notttt 100% sure that it necessarily leads to better gameplay, on the whole.

Tinkering base monetary values up/down is definitely altering the fewest variables possible, so I generally support this.

moxian:
Closing in favor of #14997

PR #14991, opened 2020-11-26
Fix roundstart Robots having null-named brainmobs

### LINKED ISSUE #14841: Runtime in getpois. Ghost orbit menu
**Issue Description**:
```
Runtime in unsorted.dm,1986: getpois returned something under a non-string name - 
   proc name: stack trace (/datum/proc/stack_trace)
   usr: NAME (CKEY) (/mob/dead/observer)
   usr.loc: The floor (41,145,1) (/turf/simulated/shuttle/floor)
   src: /datum/orbit_menu (/datum/orbit_menu)
   call stack:
   /datum/orbit_menu (/datum/orbit_menu): stack trace("getpois returned something und...")
   /datum/orbit_menu (/datum/orbit_menu): tgui static data(NAME (/mob/dead/observer))
   /datum/tgui (/datum/tgui): open()
   /datum/orbit_menu (/datum/orbit_menu): tgui interact(NAME (/mob/dead/observer), "main", /datum/tgui (/datum/tgui), 0, null, /datum/tgui_state/observer_sta... (/datum/tgui_state/observer_state))
   NAME (/mob/dead/observer): Orbit()
   Orbit (/obj/screen/ghost/orbit): Click(null, "mapwindow.map", "icon-x=17;icon-y=19;left=1;scr...")
```

It not to runtime

**What happened instead**:
It runtimes. Unsure what the effects are functionally

Runtimes are bad

- Have a roundstart borg with "robot" job title.
- Have a ghost open orbit menu

TGUI version of the orbit menu. Runtime observed around 14:00 7/11/2020

moxian:
## What Does This PR Do
Very slightly fixes robotization process, so that old mmis are properly cleaned up.
The brainmobs in the new posibrains now get the name of a human being robotized, and not a `null`.
Side-effect: roundstart Robots' brainmob is now called their human name (rather than HIU-123), so said human name appears in the orbit menu (under NPCs) and orbiting that orbits the borg (does that make sense?). This should have next to no effect on gameplay, or break immersion, since the MMI already had the human name on it anyway.

Honestly this entire MMI affair looks rather gnarly, but i'm not confident enough in robotics to fix it "properly"

Fixes #14841 

## Why It's Good For The Game
Bugfix.

## Changelog
:cl:
fix: Roundstart Robots no longer have two MMIs; and the brain inside the MMI no longer has null name
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/brain/robotic_brain.dm
+++ b/code/modules/mob/living/carbon/brain/robotic_brain.dm
@@ -90,10 +90,13 @@
 
 /obj/item/mmi/robotic_brain/transfer_identity(mob/living/carbon/H)
 	name = "[src] ([H])"
-	if(isnull(brainmob.dna))
-		brainmob.dna = H.dna.Clone()
-	brainmob.name = brainmob.dna.real_name
-	brainmob.real_name = brainmob.name
+
+	brainmob.dna = H.dna.Clone()
+	// I'm not sure we can remove species override. There might be some loophole
+	// that would allow posibrains to be cloned without this.
+	brainmob.dna.species = new /datum/species/machine()
+	brainmob.real_name = brainmob.dna.real_name
+	brainmob.name = brainmob.real_name
 	brainmob.timeofhostdeath = H.timeofdeath
 	brainmob.stat = CONSCIOUS
 	if(brainmob.mind)
--- a/code/modules/mob/transform_procs.dm
+++ b/code/modules/mob/transform_procs.dm
@@ -91,13 +91,21 @@
 	O.job = "Cyborg"
 
 	if(O.mind && O.mind.assigned_role == "Cyborg")
-		if(O.mind.role_alt_title == "Robot")
-			O.mmi = new /obj/item/mmi/robotic_brain(O)
-			if(O.mmi.brainmob)
-				O.mmi.brainmob.name = O.name
-		else
-			O.mmi = new /obj/item/mmi(O)
-		O.mmi.transfer_identity(src) //Does not transfer key/client.
+		var/obj/item/mmi/new_mmi
+		switch(O.mind.role_alt_title)
+			if("Robot")
+				new_mmi = new /obj/item/mmi/robotic_brain(O)
+				if(new_mmi.brainmob)
+					new_mmi.brainmob.name = O.name
+			if("Cyborg")
+				new_mmi = new /obj/item/mmi(O)
+			else
+				// This should never happen, but oh well
+				new_mmi = new /obj/item/mmi(O)
+		new_mmi.transfer_identity(src) //Does not transfer key/client.
+		// Replace the MMI.
+		QDEL_NULL(O.mmi)
+		O.mmi = new_mmi
 
 	O.update_pipe_vision()
 


code/modules/mob/transform_procs.dm
```diff
@@ -91,11 +91,19 @@
...
 	if(O.mind && O.mind.assigned_role == "Cyborg")
-		if(O.mind.role_alt_title == "Robot")
-			O.mmi = new /obj/item/mmi/robotic_brain(O)
-		else
-			O.mmi = new /obj/item/mmi(O)
-		O.mmi.transfer_identity(src) //Does not transfer key/client.
+		var/obj/item/mmi/new_mmi
+		switch(O.mind.role_alt_title)
+			if("Robot")
+				new_mmi = new /obj/item/mmi/robotic_brain(O)
```
dearmochi, line 97: Huh, I didn't know having the "Robot" title changes which brain you get as a borg, interesting

dearmochi (approved):

code/modules/mob/living/carbon/brain/robotic_brain.dm
```diff
@@ -90,10 +90,13 @@
...
 	name = "[src] ([H])"
-	if(isnull(brainmob.dna))
-		brainmob.dna = H.dna.Clone()
-	brainmob.name = brainmob.dna.real_name
-	brainmob.real_name = brainmob.name
+
+	brainmob.dna = H.dna.Clone()
+	// I'm not sure we can remove species override. There might be some loophole
+	// that would allow posibrains to be cloned without this.
+	brainmob.dna.species = new /datum/species/machine()
```
SteelSlayer, line 97: Does this mean the mob in the MMI will get the change monitor action button, like IPC mobs get?
`var/datum/action/innate/change_monitor/monitor`

moxian: It does not
dreamseeker-20201127-171514

(ignore the fact that the research scanner action is duplicated - this is the same in live version)

SteelSlayer: Ah right because its set directly and not going through `.set_species()`.

Spacemanspark:
Perhaps it would be interesting to have it use the MMI type of the character you're using to play borg, rather than this workaround.

IE; slime people would have a slime core MMI, IPC would have a posibrain, so on so forth.

If we're going to be naming them the same, they might as well be the same mob.

moxian:
This PR does not change in-game observable behavior, other than removing an extra useless MMI.
On live, the species is set to machine in the /New() proc, and untouched in /transfer_identity, this PR just re-sets it in `/transfer_identity()` as well.

Currently Robot gets robo brain, Android else gets MMI with the "real" brain of the species (so, yes, if you disassemble an android of a character with Slime People species at round start - you will find a MMI with a slime core inside). I'm not sure how it works for IPC, but probably horribly (a posibrain inside a MMI?...)

Changing that so that the difference between Robot and Android is purely in the name, and that the MMI type depends on the species is... doable, but i don't understand robotics (nor silicon) *at all* to have an opinion on whether it's a good idea.
So I'll claim that this is out of scope of this PR, and that the PR is not a "workaround" but rather a "normal" fix, preserving the behavior.

Spacemanspark:
Ah, it seems I misinterpreted what you were saying, my bad.
Hrm, I didn't realize it already did that. I suppose that's because I just play IPCs all the time :p

I believe an IPC spawning in as a borg gets a robotic brain... which makes little sense to me, but it's not an MMI with a regular brain, at least.

Kyep:
@farie82 
There's still a purpose to this PR, as I've still seen:
[2020-12-02T02:11:07] Runtime in unsorted.dm,1986: getpois returned something under a null name. Type: /mob/living/carbon/brain
even after the previous fix was merged.
I think there may be a few seconds at round start where borg brains can trigger this error, even before they're renamed by the code I added.
Worth investigating.

variableundefined (changes requested):
Please deconflict in 7 days

moxian:
deconflicted

AffectedArc07 (approved):

PR #14997, opened 2020-11-27
Make fluff items cheaper

moxian:
## What Does This PR Do
Fluff items are now 2x cheaper, claw game now costs 5cr.
This puts banjo at 250cr (slightly cheaper than a PDA) and synth at 500cr (slightly more expensive than a PDA)
This is a - hopefully - less controversial alternative to #14989 
This is NOT changing starting account balance in any way, since that seems to be a very touchy subject

## Why It's Good For The Game
The prices in the cargo shop are mostly arbitrary, since nobody particularly cared how much an item costs if you can spend 15 minutes spinning slots and get infinite money.
Now that #14953 is merged and infinite money is a thing of the past, being able to obtain a synth is no longer a given, and people are upset that they cannot play music and have fun.
This calls to re-evaluate the pricing (which, again, was set more or less arbitrarily in the first place).
Being able to buy a synth the moment you enter the station (even, potentially, having to spend ALL your money doing so) is a good thing

## Changelog
:cl:
tweak: Cargo shop fluff items are twice as cheap now.
tweak: Claw game is now cheaper to play.
/:cl:

### DIFF
--- a/code/modules/arcade/claw_game.dm
+++ b/code/modules/arcade/claw_game.dm
@@ -5,7 +5,7 @@ GLOBAL_VAR(claw_game_html)
 	desc = "One of the most infuriating ways to win a toy."
 	icon = 'icons/obj/arcade.dmi'
 	icon_state = "clawmachine_1_on"
-	token_price = 15
+	token_price = 5
 	window_name = "Claw Game"
 	var/machine_image = "_1"
 	var/bonus_prize_chance = 5		//chance to dispense a SECOND prize if you win, increased by matter bin rating
--- a/code/modules/store/items.dm
+++ b/code/modules/store/items.dm
@@ -43,178 +43,178 @@
 	name = "Snap-Pops"
 	desc = "Ten-thousand-year-old chinese fireworks: IN SPACE"
 	typepath = /obj/item/storage/box/snappops
-	cost = 200
+	cost = 100
 
 /datum/storeitem/dnd
 	name = "Dungeons & Dragons Set"
 	desc = "A box containing minifigures suitable for a good game of D&D."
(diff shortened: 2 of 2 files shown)


Necaladun (changes requested):
Claw game I think should be risky for the gambling and the house always wins - they're not meant to be prizes people should expect to get.

The other things being cheaper is generally good as it will encourage their use. 

Snap pops being that cheap might be a bit annoying but it's not big deal. Overall this is a good PR and my change requests aren't something I'm going to die on a hill for.

moxian:
With my claw game skills i get a prize 3 times out of ten.
Which means I need to play ~15 times to get a random non-carp plushie. (Or, likewise ~15 times if i *want* a carp plushie)
At 15cr/spin that makes the random(!) plushie cost 225cr. If you're aiming for something more specific you need to get more than one random plushie.
While I agree that claw game should not be a charity and there should be risk involved, I don't think 225cr is a fair price for something you don't *quite* want. (Compare, with, again, a 250cr guitar from which you *know* what to expect)
Also compare: if you want a figurine, you need to currently spend, again, 15 spins/225cr to get one. When you can buy *a box* of them from cargo for only 100cr.

I would honestly like to cut the claw game cost in half down from the current 15cr, but 7.5 is not a nice number. And 10 is still a bit too high for my tastes. Hence 5cr.

Let me know if this is not convincing enough, and I'll increase it back (preferably to 10cr though - a "rounder" number than 15).

Necaladun:
That's convincing, keep the price there.

Kyep (approved):

AffectedArc07 (approved):

PR #14999, opened 2020-11-27
Adds nanopaste to IPC medkit

Darkmight9:
## What Does This PR Do
Adds nanopaste to IPC medkits

## Why It's Good For The Game
Nanopaste is handy to have to fix up IPCs as well as robotic limbs and organs. Sometimes RnD or robotics can't be bothered to make any so it'd be nice to have some on hand via cargo orders.

## Changelog
:cl:
tweak: added nanopaste to IPC medkits
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/firstaid.dm
+++ b/code/game/objects/items/weapons/storage/firstaid.dm
@@ -181,6 +181,7 @@
 	new /obj/item/stack/cable_coil(src)
 	new /obj/item/stack/cable_coil(src)
 	new /obj/item/robotanalyzer(src)
+	new /obj/item/stack/nanopaste(src)
 
 /obj/item/storage/firstaid/machine/empty
 	empty = TRUE


AffectedArc07:
iirc this has been tried before and shot quickly, but dont quote me on that.

Spacemanspark:
Nanopaste is a bit much imo, it allows you to completely bypass surgery when an IPC takes enough damage... something I'd rather see stay a thing.

Denthamos:
@AffectedArc07 Was that not for the emergency kit you spawn with and not the first aid kit?  It would basically function like a adv medkit taking the place of the trauma and burn items in there but for IPC's.

Darkmight9:
> @AffectedArc07 Was that not for the emergency kit you spawn with and not the first aid kit? It would basically function like a adv medkit taking the place of the trauma and burn items in there but for IPC's.

Yeah someone in a previous PR wanted to give IPCs the medkit as a replacement for the emergency kit, this just adds nanopaste to the IPC kit. You'd still have to order and distribute them

Kyep (commented):
My understanding is:
- The IPC medkit is meant to be the machine equivalent of the basic organic regular medkit: https://github.com/ParadiseSS13/Paradise/blob/master/code/game/objects/items/weapons/storage/firstaid.dm#L54
- The organic version has syptic, salic, silversulf, a healthanalyser and autoinjector.
- None of the things in the organic version heal organ damage.
- So you're actually making the machine version of this kit superior to the regular organic version, as it heals more things.
- You're not increasing the price, though. You can still get 4 of these kits per 10 cargo points. Just like the organic kits.
- So, its totally feasiable now for the QM to just spend 20-30 points on 2-3 crates of these, and have enough kits to give every IPC on the station a kit complete with nanopaste.

I'm a little bit skeptical of this. Not totally opposed, but... I feel like IPCs are already able to self-repair possibly better than humans, and if you make it easy for them to obtain stuff that even heals organ damage... you're just further improving their ease of healing above base human.

Qwertytoforty:
Correct me if I am wrong @Kyep , but doesn't screwdrivers have a 90% chance to heal internal organ damage in surgery? However, other points are valid.

variableundefined (commented):
Objection echoing points above

Kyep (commented):
Officially objecting to this, based on Spark (who is the person I know who plays IPCs most) thinking this will essentially negate the need for IPC surgery, which is too much of a buff.

AffectedArc07:
2 maintainers and a head objection, with no one in favour. Closing.

PR #15011, opened 2020-11-29
Fixes TGUI ghost orbit menu runtime

Kyep:
## What Does This PR Do
1) Fixes this runtime:
[2020-11-29T03:53:21] Runtime in unsorted.dm,1986: getpois returned something under a non-string name - 
   proc name: stack trace (/datum/proc/stack_trace)
   usr: XXXXXXXXXXXXXXXXXX
   usr.loc: The floor (50,145,1) (/turf/simulated/floor/plasteel)
   src: /datum/orbit_menu (/datum/orbit_menu)
   call stack:
   /datum/orbit_menu (/datum/orbit_menu): stack trace("getpois returned something und...")
   /datum/orbit_menu (/datum/orbit_menu): ui static data(XXXXXXXXXXXX (/mob/dead/observer))
   /datum/tgui (/datum/tgui): open()
   /datum/orbit_menu (/datum/orbit_menu): ui interact(XXXXXXXXXX (/mob/dead/observer), "main", /datum/tgui (/datum/tgui), 0, null, /datum/ui_state/observer_state (/datum/ui_state/observer_state))
   XXXXXXXXXXX (/mob/dead/observer): Orbit()
   Orbit (/obj/screen/ghost/orbit): Click(null, "mapwindow.map", "icon-x=20;icon-y=10;left=1;scr...")

The problem was caused by cyborg.mmi.brainmob /mob/living/carbon/brain not having a name defined, which led to runtimes in TGUI.

2) Adds better logging (runtimes now show the path of the object) so if more errors like this come up in future, it will be easier to find which mob is causing them.

## Changelog
:cl: Kyep
fix: fixed TGUI ghost orbit menu generating a runtime error.
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -54,7 +54,8 @@
 	R.name = heldname
 	R.custom_name = heldname
 	R.real_name = heldname
-
+	if(R.mmi && R.mmi.brainmob)
+		R.mmi.brainmob.name = R.name
 	return TRUE
 
 /obj/item/borg/upgrade/restart
--- a/code/modules/mob/dead/observer/orbit.dm
+++ b/code/modules/mob/dead/observer/orbit.dm
@@ -47,10 +47,10 @@
 
 	var/list/pois = getpois(mobs_only = FALSE, skip_mindless = FALSE)
 	for(var/name in pois)
-		var/poi = pois[name]
+		var/mob/M = pois[name]
 		if(name == null)
-			if(pois[name] && pois[name].type)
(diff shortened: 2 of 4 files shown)


AffectedArc07:
Question 

Why don't we just add the failing type path to the stack trace too? If it's aware that something isn't a string, why aren't we echoing it back so we're aware of the issue so we have a point of reference to chase from?

Kyep:
This does add the path of the failing thing to this runtime.
It also splits up the errors to make them more informative.

If you want to modify all stack traces in general, that's probably a different PR.

AffectedArc07 (approved):

PR #15012, opened 2020-11-29
Keeps CQC from benifitting from Gloves of the North Start

Aurorablade:
Prevents CQC from benefitting from the speed of glove of the northstar, simply by lowering the click speed back to regular. I may need to change it so it excludes the admin subtype.
The reasoning is that when someone gets it, and its honestly not too rare if someone knows how it can be pretty ridiculous 

Consideration: Restrict Sleeping carp in the same manner, I am torn on this as the feel of carp is to be a hand to hand master..and well gloves go brrr fits that bill. If people feel i need to add that restriction i can but i am on the fence.

## Changelog
:cl: Fethas
tweak: CQC is no longer super fast with gloves of the northstar.
/:cl:

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -166,7 +166,10 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		M.changeNext_move(click_speed_modifier)
+		if(istype(M.mind.martial_art, /datum/martial_art/cqc))
+			M.changeNext_move(CLICK_CD_MELEE)//normal attack speed for hulk, CQC and Carp.
+		else
+			M.changeNext_move(click_speed_modifier)
 	.= FALSE
 
 /obj/item/clothing/gloves/fingerless/rapid/admin


Mitchs98:
Honestly I already thought Carp was restricted? It was supposed to be in one PR I think...unless I'm misremembering PR's. Either way I'd honestly say to restrict both. It's insanely over powered regardless, even more-so than GOTNS. At-least without martial arts they're SOMEWHAT manageable.

imo.

Carthusia:
I dono, i recall this having been brought up before, and while I think i was all for it being limited, now, im more for it being left as is, as the extreme rarity of such a combo kind of keeps it balanced in that you almost never see it, similar to vampire or changeling syndicates, they exist, their insanely powerful, you fear them, but your most likely not going to see them very often.

Aurorablade:
From what I have been hearing its been happening  often enough to put our above the rare level

Carthusia:
i'll admit, i don't play security all that often, so maybe I just don't see it as much as it does occur.

Do we have any hard stats on this?

Spacemanspark:
Likely not.

As is, I don't think people should have access to this.
+1

procdrone:
I think this should stay in, and thats because a normal traitor normally cannot buy both due to TC costs. you can only get that combo if you land good with a discount (or well, rob another tag of his TC but that never happens), which i feel could be a nice unique thing. Just like your grace. 

Well, unless that happens more often then im also for the nerf.

Kyep (approved):

Dashsora:
CQC + north stars are only one TC off from each other, while carp and gloves are a little more far-fetched - without taking into consideration surplus RnG, and cqc is arguably stronger in combat prowess than carp.
 I wouldn't mind both being restricted, but I remembered reading in the nukie uplink's CQC entry that it was already conflicting with north stars... is it just bugged/unintentionnal ?

While we touching cqc it'd be nice if it couldn't go on discount too, that's probably one of the prime offenders of why this seems to be frequent

PR #15014, opened 2020-11-29
Better "please update byond" prompts

Kyep:
## What Does This PR Do
Alters the "please update byond" prompt that people with out-of-date byond versions get.
Before, it would only alert them if they were on major version *510* or lower. That version is _ancient_.
Now, it will alert them if they're on any version less than (as of writing) the current standard version: 513.1536.
It will consider both major version (e.g: 512, 513) and build number when making this determination.

## Why it is good for the game
There are lots of bugs (like the %22 URL bug) caused by out-of-date versions of byond.
This encourages people using those versions to update. It doesn't require they upgrade. It just reminds them to do so.

## Changelog
:cl: Kyep
add: Players now get asked to update their Byond if they connect using an out-of-date version.
/:cl:

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -335,14 +335,13 @@
 		alert(src, "You are using a byond build which is not supported by this server. Please use a build version of atleast [config.minimum_client_build].", "Incorrect build", "OK")
 		qdel(src)
 		return
+
 	var/show_update_prompt = FALSE
 	if(byond_version < SUGGESTED_CLIENT_VERSION) // Update is suggested, but not required.
 		show_update_prompt = TRUE
 	else if(byond_version == SUGGESTED_CLIENT_VERSION && byond_build < SUGGESTED_CLIENT_BUILD)
 		show_update_prompt = TRUE
-	if(show_update_prompt)
-		to_chat(src,"<span class='userdanger'>Your BYOND client (v: [byond_version].[byond_build]) is out of date. This can cause glitches. We highly suggest you download the latest client from http://www.byond.com/ before playing. </span>")
-
+	// Actually sent to client much later, so it appears after MOTD.
 
 	to_chat(src, "<span class='warning'>If the title screen is black, resources are still downloading. Please be patient until the title screen appears.</span>")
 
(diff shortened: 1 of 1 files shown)


code/modules/client/client_procs.dm
```diff
@@ -333,8 +335,14 @@
...
 		qdel(src)
 		return
+	var/show_update_prompt = FALSE
 	if(byond_version < SUGGESTED_CLIENT_VERSION) // Update is suggested, but not required.
-		to_chat(src,"<span class='userdanger'>Your BYOND client (v: [byond_version]) is out of date. This can cause glitches. We highly suggest you download the latest client from http://www.byond.com/ before playing. </span>")
+		show_update_prompt = TRUE
+	else if(byond_version == SUGGESTED_CLIENT_VERSION && byond_build < SUGGESTED_CLIENT_BUILD)
+		show_update_prompt = TRUE
+	if(show_update_prompt)
+		to_chat(src,"<span class='userdanger'>Your BYOND client (v: [byond_version].[byond_build]) is out of date. This can cause glitches. We highly suggest you download the latest client from http://www.byond.com/ before playing. </span>")
```
AffectedArc07, line 344: These messages won't even be seen because goonchat doesn't even start for another ~100 lines or so.

This entire system annoys me a bit (not your code, the whole version checking thing), especially since it's sending alerts to the user and then promptly qdel-ing the client before they even get a chance to see the alerts at all.

Kyep: I tested it, and saw the messages.

AffectedArc07: Confimed. Its `alert()` and `to_chat()` which differ.

Dumbdumn5 (approved):
Easy QoL approval on this one.

code/modules/client/client_procs.dm
AffectedArc07, line 344: Last minor change request. Remove the trailing space from the end of the `.`, make the BYOND link a hyperlink, and make it use HTTPS instead of HTTP

AffectedArc07 (changes requested):

AffectedArc07 (approved):

Issue #14968, opened 2020-11-22
List of Browser UIs able to be converted to TGUI


AffectedArc07:
The following is a list of non-TGUIs to be converted. This is non time critical and just used as a simple project if someone wants one.

This list intentionally **DOES NOT INCLUDE OOC UIS**. Do not convert purely OOC UIs to TGUI. Your PR will be objected to.

### Pure HTML browser UIs (Black on White) 
- [ ] Magnet control (Firing range) | code\game\machinery\magnet.dm (Will do as part of larger refactor)
- [ ] Space heaters (Same template as ACs likely) | code\game\machinery\spaceheater.dm
- [ ] Message monitoring console | code\game\machinery\computer\message.dm
- [ ] Mass driver and pod doors control | code\game\machinery\computer\pod.dm
- [ ] Mech interface (Includes all sub-equipment like sleepers) | code\game\mecha\mecha_topic.dm
- [ ] All multitool link menus | code\game\objects\objs.dm (This is being killed in another refactor)
- [ ] Human pAI interface | code\game\objects\items\devices\paicard.dm
- [ ] Tracking implant locator | code\game\objects\items\weapons\teleportation.dm
- [ ] Secure storage | code\game\objects\items\weapons\storage\secure.dm
- [ ] Noticeboard | code\game\objects\structures\noticeboard.dm
- [ ] Claw game | code\modules\arcade\claw_game.dm
- [ ] Prize counter machine | code\modules\arcade\prize_counter.dm
- [ ] Stacking machine console | code\modules\mining\machine_stacking.dm
- [ ] Coin mint | code\modules\mining\mint.dm
- [ ] AI alarm alerts | code\modules\mob\living\silicon\ai\ai.dm
- [ ] pAI recruiting | code\modules\mob\living\silicon\pai\recruit.dm
- [ ] Drone installed modules list | code\modules\mob\living\silicon\robot\drone\drone_items.dm
- [ ] Newspapers (Be careful with sanitization) | code\modules\newscaster\obj\newspaper.dm
- [ ] Ninja SpiderOS | code\modules\ninja\suit\SpiderOS.dm
- [ ] Folders | code\modules\paperwork\folders.dm
- [ ] Photos | code\modules\paperwork\photography.dm
- [ ] AI PDA message log | code\modules\pda\ai.dm
- [ ] Dart gun | code\modules\projectiles\guns\dartgun.dm
- [ ] R&D server control | code\modules\research\server.dm 

[...]

PR #15016, opened 2020-11-29
[TGUI] Ore Redemption Machine

dearmochi:
## What Does This PR Do

- **Ports the Ore Redemption Machine from BrowserUI to TGUI.**
  - Not sure if it was an issue before, but inserted IDs and disks will now drop to the ground should the machine be destroyed.
  - Upgraded micro-manipulators never did effectively did anything, so the related code to upgrading them was removed.
  - Fixed a bug where if the machine stopped being powered, the inserted ID would drop but the reference would not clean up.
- Added `ore_type` field to `/datum/material` to make looking up ore stacks associated to a material easier.
- Completely refactored, cleaned up and documented the code.

## Why It's Good For The Game

Progress towards #14968

## Images of changes

NlMt2iWNuZ

EPQKtLhBb9

TScwp5meME

pPPPE1WiRV

## Changelog
:cl:
add: Port Ore Redemption Machine UI to TGUI
/:cl:


tgui/packages/tgui/interfaces/OreRedemption.js
```diff
@@ -0,0 +1,229 @@
...
+  const {
+    id,
+    points,
+    disk,
+  } = data;
+  const {
+    ...rest
+  } = properties;
+  return (
+    <Section {...properties.rest}>
```
AffectedArc07, line 43: Ima be stupid here a second. Didnt you just create `rest` in the lines above this? And if so, why do you need to prefix with `...properties`

dearmochi: oof

AffectedArc07 (approved):

PR #15019, opened 2020-11-30
Fixes reflector chat spam

Kyep:
# Changelog
:cl: Kyep
del: Removed the chat message when an emitter beam bounces off a reflector.
/:cl:

# Why
Because of SM setups that include 16+ emitters and crash the chat of anyone who watches them due to obscene amounts of chat messages created by 16 emitters' beams being constantly reflected.


AffectedArc07:
Citing as fix since currently this breaks peoples chat outputs, requiring them to restart their client.

AffectedArc07 (approved):

PR #15022, opened 2020-11-30
Drops twohanded items that failed to wield instead of the user's active hand

dearmochi:
## What Does This PR Do

Follow-up to #14993 after `/obj/item/twohanded/required` would drop the wrong item if wielding on pickup failed. It will now always drop the twohanded item instead of the active item.

## Why It's Good For The Game

Bugfix to bugfix

## Changelog
:cl:
fix: Picking up a twohanded weapon and failing to wield it will now drop that item properly instead of sometimes dropping whatever is in the other hand.
/:cl:


farie82 (approved):
Wont work for antidrop items in this case but those didnt work anyway

PR #15025, opened 2020-11-30
Makes machine frames show their circuit name

SabreML:
## What Does This PR Do

Refactors circuit board naming to make machine frames and computer frames show their circuit board if inserted.
Also cleaned up some unnessecary vendor `Initialize()`s.

Also also fixes YouTool vendors not being constructable or deconstructable. (Crowbar just bounced off)

Also also also refactors constructing/deconstructing computer frames to use `tool_act()`s rather than a massive `attackby()`, allows computer frames to always be wrenched and moved in any state, and removed the HONKputer to avoid a future merge conflict.

## Why It's Good For The Game

You don't have to guess what machine you're making by looking at the part requirements, or by taking the circuit out.

## Images of changes

arcade
vend
borgo charger

## Changelog
:cl:
tweak: Half-built machine and computer frames now show the circuit board's name. e.g: `machine frame (Cyborg Recharger)`
tweak: Added the ability to construct and deconstruct YouTool vendors, and added a refill pack to the 'Engineering Vendor Supply Crate' in cargo.
tweak: Allowed computer frames to always be wrenched and moved around in any construction state.
/:cl:


code/game/machinery/vending.dm
```diff
@@ -885,14 +893,6 @@
 	ads_list = list("Only the finest!","Have some tools.","The most robust equipment.","The finest gear in space!")
 	refill_canister = /obj/item/vending_refill/assist
 
-/obj/machinery/vending/assist/Initialize(mapload)
```
AffectedArc07, line 888: Gonna test a lot of these locally. Modifying consturction calls en masse worries me.

SabreML: As far as I can tell all of them were identical except for the circuit board, so hopefully nothing will have changed.

Kyep (approved):

code/game/machinery/constructable_frame.dm
```diff
@@ -223,61 +225,58 @@ Note: Once everything is added to the public areas, will add MAT_METAL and MAT_G
...
 	. = TRUE
 	if(!I.use_tool(src, user, 0, volume = I.tool_volume))
 		return
-	var/static/list/display_vending_names_paths
-	if(!display_vending_names_paths)
-		display_vending_names_paths = list()
-		for(var/path in vending_names_paths)
-			display_vending_names_paths[vending_names_paths[path]] = path
-	var/choice =  input(user, "Choose a new brand","Select an Item") as null|anything in display_vending_names_paths
+	var/choice =  input(user, "Choose a new brand", "Select an Item") as null|anything in vending_names_paths
```
SteelSlayer, line 265: It's possible for `choice` to equal null if the user chooses cancel, which gets passed to `set_type` and results in a runtime.

SabreML:
This should probably have another testmerge just to make sure I've caught all of the exceptions, but other than that I think it's good to go.

farie82 (changes requested):
Few things. Good stuff

code/game/machinery/computer/buildandrepair.dm
```diff
@@ -46,15 +17,20 @@
...
 
 /obj/item/circuitboard/machine
 	board_type = "machine"
 
+/obj/item/circuitboard/Initialize(mapload)
+	. = ..()
+	format_board_name()
+
+/obj/item/circuitboard/proc/format_board_name()
+	name = "[initial(name)] ([board_name])"
```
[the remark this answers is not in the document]
SabreML: `ASSERT()` won't work here I think since the `board_name` for `/obj/item/circuitboard` will always be null. So I'll do this instead to stop any formatting errors:
```dm
if(board_name)
	name = "[initial(name)] ([board_name])"
else
	name = "[initial(name)]"
```

code/game/machinery/Freezer.dm
```diff
@@ -179,9 +179,10 @@
 /obj/machinery/atmospherics/unary/heat_reservoir/heater/New()
 	..()
 	initialize_directions = dir
-	var/obj/item/circuitboard/thermomachine/H = new /obj/item/circuitboard/thermomachine(null)
+	var/obj/item/circuitboard/thermomachine/H = new
 	H.build_path = /obj/machinery/atmospherics/unary/heat_reservoir/heater
```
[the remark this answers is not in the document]
SabreML: Finally got around to this PR again, sorry for the delay!

I'm not sure what you mean by this exactly, would you mind elaborating a bit?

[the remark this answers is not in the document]
SabreML: Presumably not relevant anymore due to #15376.

code/game/machinery/computer/buildandrepair.dm
```diff
@@ -67,307 +46,371 @@
...
 	origin_tech = "programming=2"
 	icon = 'icons/obj/machines/HONKputer.dmi'
 	icon_state = "bananium_board"
 	board_type = "honkcomputer"
 
 
-/obj/item/circuitboard/supplycomp/attackby(obj/item/I as obj, mob/user as mob, params)
-	if(istype(I,/obj/item/multitool))
+/obj/item/circuitboard/supplycomp/attackby(obj/item/I, mob/user, params)
+	if(istype(I, /obj/item/multitool))
```
farie82, line 413: Could you rewrite this to use `multitool_act`?

code/game/machinery/computer/buildandrepair.dm
```diff
@@ -389,94 +432,127 @@
...
+/obj/structure/computerframe/proc/drop_computer_parts()
+	new base_mineral(loc, 5)
+	if(circuit)
+		circuit.forceMove(loc)
+		circuit = null
+	if(state >= 3)
+		var/obj/item/stack/cable_coil/A = new /obj/item/stack/cable_coil(loc)
+		A.amount = 5
+	if(state >= 4)
+		new /obj/item/stack/sheet/glass(loc, 2)
```
farie82, line 499: Can you use `drop_location()` here instead of `loc`?

code/game/machinery/computer/buildandrepair.dm
```diff
@@ -389,94 +432,127 @@
...
+
+/obj/structure/computerframe/attackby(obj/item/I, mob/user, params)
```
farie82, line 501: Christ. Lots of tool behaviour in here ... could you rewrite those to `XXX_act`s if you have time spare?

code/game/machinery/computer/buildandrepair.dm
```diff
@@ -560,54 +636,54 @@
 	icon = 'icons/obj/machines/HONKputer.dmi'
 	base_mineral = /obj/item/stack/sheet/mineral/bananium
 
-/obj/structure/computerframe/HONKputer/attackby(obj/item/P as obj, mob/user as mob, params)
+/obj/structure/computerframe/HONKputer/attackby(obj/item/I, mob/user, params)
```
farie82, line 639: Same here as above

farie82 (changes requested):
Few things. Feel free to say no to the `tool_act` refactors if you lack the time

code/game/machinery/computer/buildandrepair.dm
SabreML: You know I've never actually seen `drop_location()` before now

code/game/machinery/computer/buildandrepair.dm
SabreML: Looks like this is getting removed by #16213, so I'll just do the other one.

code/game/machinery/computer/buildandrepair.dm
farie82: Hmmm looking at it again. I am not sure if that is the proper proc. A maint can confirm this or not.
But `drop_location` is used to create more generic handling of dropping items while inside of `atoms`. It is also used for deconstruction in certain cases. It makes use of `AllowDrop` (yes the casing is all over the place) which is usually `FALSE` except for space pods and lockers for example.
If it is `TRUE` then the `drop_location` proc will drop the items inside of the thing (as is its standard implementation). This proc could however be overridden for say multi-tile buildings.

SabreML: Unsure if this is neccesary for computer frames, but I've done it anyway because there doesn't seem to be any reason *not* to.

AffectedArc07 (approved):

PR #15026, opened 2020-11-30
Terror event spawns now pick players randomly

Kyep:
## What Does This PR Do
Previously, the terror spiders major event spawned its spiders, then offered each to ghosts individually, with the first player to click a spider or its prompt getting control of that spider.
Now, the terror spiders event polls eligible ghosts for 30 seconds, then assigns spiders randomly to eligible opted-in players. Only after its picked who will be what spider, does it actually spawn any spiders.
This means players who accept quickly no longer have advantage on becoming initially spawned terror spiders.
It also means that if not enough people sign up to play terrors during an event, admins are warned and fewer than normal spiders are spawned.

## Why It's Good For The Game
Normally, if terrors are being spawned there's generally enough of them for them to be handled on a first-come, first-serve basis.
However, it is possible for the terror spider spawn event to choose to spawn, say, only a single prince of terror.
When that happens, its probably best for selection there to be entirely random, as its likely to be the only terror in the round at all.

## Note
Due to multiplayer issues, this will require a TM to fully test.
This only applies to the initial 1-5 spiders directly spawned by the event. It does NOT apply to any terror who grows up from a spiderling. Don't ask me to apply it there too. I may or may not do that in future, but I want to see how this goes before I even consider that.

## Changelog
:cl: Kyep
tweak: terror spiders spawned directly by event now get offered to all eligible ghosts for 30 seconds, and pick players randomly from that pool.
/:cl:

### DIFF
--- a/code/modules/events/spider_terror.dm
+++ b/code/modules/events/spider_terror.dm
@@ -16,7 +16,6 @@
 		log_and_message_admins("Warning: Could not spawn any mobs for event Terror Spiders")
 
 /datum/event/spider_terror/start()
-	var/list/vents = get_valid_vent_spawns(exclude_mobs_nearby = TRUE, exclude_visible_by_mobs = TRUE)
 	var/spider_type
 	var/infestation_type
 	if((length(GLOB.clients)) < TS_HIGHPOP_TRIGGER)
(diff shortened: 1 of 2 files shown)


code/modules/events/spider_terror.dm
```diff
@@ -44,11 +43,18 @@
 			// Strongest, only used during highpop.
 			spider_type = /mob/living/simple_animal/hostile/poison/terror_spider/queen
 			spawncount = 1
-	while(spawncount && length(vents))
-		var/obj/vent = pick_n_take(vents)
-		new spider_type(vent.loc)
-		spawncount--
-		successSpawn = TRUE
+	spawn() // Just like in xeno and demon events, this spawn() is necessary. Do not remove.
```
SteelSlayer, line 46: Why is this necessary? What else needs to run before the code below?

Kyep: Without it, the whole thing breaks. The event loops and it generates incorrect popups.

AffectedArc07, line 46: I am aware that killing the `spawn()` here will break things. **However**, I am going to ask you offload this to another proc and use `INVOKE_ASYNC` to do the work instead of wrapping all of this inside a `spawn()`

AffectedArc07 (changes requested):

Spacemanspark:
Just to be clear, does this only impact the initial wave of terrors? Or does this impact all terror spawns?

cluwneythescourge:
thank goodness for this!

AffectedArc07 (approved):

Kyep:
@Spacemanspark only the initial spiders spawned by the event. I won't be applying it to all spiders in this PR (and will likely object if anyone tries). That's something for me to consider in another PR, later, once I'm sure I'm happy with the effect this has.

Dumbdumn5 (approved):
Though I would personally support this for all spiders, this is a step in the right direction, RNG is a glorious thing and so is giving everyone a chance to actually play a midround without dedicating themselves to hunting the things down and slapping the thing as soon as it shows up. :+1:

Kyep:
Please don't merge this yet, I want to do a TM of it on the live server before its merged.
I realize this is probably excessively cautious, I've tested it several times, but I saw some really bad timing issues when spawn() was not used, and while I have tested it several times with INVOKE_ASYNC I want a live server test before full merge just to be doubly sure.

AffectedArc07:
TM'd with 0 issues

PR #15034, opened 2020-12-01
Tajaran Hissing

SabreML:
## What Does This PR Do

Allows Tajaran to `*hiss`, and adds a [new sound](https://www.dropbox.com/s/5qy9ssn8znaxcb1/kittyhiss.wav) for it. *(Sounds a lot better in-game)*

## Why It's Good For The Game

~~Kitty go hiss~~
Tajaran need some love. Especially since they don't actually have any emotes right now.

## Changelog
:cl:
add: Tajaran can now use `*hiss`
soundadd: Custom hiss sound for Tajaran
/:cl:

*not sure if this is needed here but better safe than sorry*
`Copyright CC BY 3.0 Garuda1982 at freesound.org`


Cocacolagua:
controversial tag pls

Spacemanspark:
I'm going to admit, not a fan of that sound.

It's more akin to an automatic car wash water sprayer than a cat's hiss.

meow20:
Kitty go hiss

Kyep (changes requested):
Concept of taj being able to hiss with a sound is fine. I like that.
The sound file is terrible. Get a different sound file.

AffectedArc07 (changes requested):
Agreeing with the others. The sound file used here just sounds like static inside of an audio stream. It has to be replaced.

SabreML:
I guess since I heard it in context with the original sound it seems fine to me.

I'll look for another one.

Kyep:
This sound is better than the previous one, but IMHO still sounds more like an airlock opening than a cat hissing.
No longer as opposed as I was, but I would still much prefer a better sound file.

variableundefined:
Having tried out the sound on local, I am also not in favour of the sound. 

Closing due to two maint objections.

KasparoVy:
Good sound = angry catbeast merged, hard to get a good free license sound tho

Good luck!

PR #15035, opened 2020-12-01
Fixes TGUIs not reopening properly sometimes

### LINKED ISSUE #15028: Issues opening windows with TGUI
**Issue Description**:
TGUI windows currently sometimes bug out. I meant to post this during the test merge, and meant to post this yesterday. But I've been sick(still am) and haven't felt like doing much. Anywho. Essentially some TGUI windows won't open without reloading UI resources(in the case of terminals/consoles/anything you can't stick in inventory or drop) or (in the case of items that can go in inventory) drop/repeatedly put in and out of backpack and click it repeatedly. It won't produce an error(outside of perhaps a runtime on server log), it just simply won't load at all.

I expected to be able to use the items and consoles like normal. 

**What happened instead**:
Consoles and items could sometimes(frequently) fail to load and be usable at all.

UI should work without having to reload it constantly.

1. Click a console or pick up an item with a UI.
2. Use it a few times, being sure to close the window in-between uses.
3. Eventually, it just won't work.
4. Reload UI resources or drop/put in bag repeatedly to get it to work again.
5. Repeat.

Ever since #14957 , I'm assuming. The issue just wasn't reported on the PR itself. It may not be that, but that's when I noticed it was during that test merge.

**Extra information**:
N/A

AffectedArc07:
## What Does This PR Do
See title

Fixes #15028

## Why It's Good For The Game
People should be able to use their UIs properly

## Changelog
:cl: AffectedArc07
fix: Fixed a bug where TGUIs wouldnt open sometimes 
/:cl:

### DIFF
--- a/code/modules/tgui/tgui.dm
+++ b/code/modules/tgui/tgui.dm
@@ -125,7 +125,7 @@
 	// Instruct the client to signal UI when the window is closed.
 	// NOTE: Intentional \ref usage; tgui datums can't/shouldn't
 	// be tagged, so this is an effective unwrap
-	winset(user, window_id, "on-close=\"uiclose \ref[src]\"")
+	winset(user, window_id, "on-close=\"uiclose [src.UID()]\"")
 
 	// Pre-fetch initial state while browser is still loading in
 	// another thread
--- a/code/modules/tgui/tgui.dm
+++ b/code/modules/tgui/tgui.dm
@@ -123,8 +123,6 @@
 	user << browse(html, "window=[window_id];[window_options]")
 
 	// Instruct the client to signal UI when the window is closed.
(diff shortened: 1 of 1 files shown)


code/modules/tgui/tgui.dm
```diff
@@ -125,7 +125,7 @@
 	// Instruct the client to signal UI when the window is closed.
 	// NOTE: Intentional \ref usage; tgui datums can't/shouldn't
```
SabreML, line 126: Should probably change this too since it's not `\ref` anymore.

Spacemanspark:
Thank fuck

code/modules/tgui/tgui.dm
AffectedArc07: Will do when I can

farie82 (approved):

PR #15039, opened 2020-12-02
Removes the 0.1 second delay for window deconstruction

SabreML:
## What Does This PR Do

Removes the 1 decisecond delay when constructing or deconstructing directional window panes.

## Why It's Good For The Game

It's very irritating if you're trying to go quickly.

## Changelog
:cl:
tweak: Removed the 0.1 second delay when (de)constructing window panes
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -88,7 +88,7 @@ GLOBAL_LIST_INIT(wcCommon, pick(list("#379963", "#0d8395", "#58b5c3", "#49e46e",
 		shards++
 		setDir()
 
-	if((decon_speed == null) && fulltile)
+	if(decon_speed == null && fulltile)
 		decon_speed = 2 SECONDS
 
 	var/rods = 0
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -88,11 +88,8 @@ GLOBAL_LIST_INIT(wcCommon, pick(list("#379963", "#0d8395", "#58b5c3", "#49e46e",
 		shards++
 		setDir()
 
-	if(decon_speed == null)
-		if(fulltile)
-			decon_speed = 20
-		else
-			decon_speed = 1
+	if((decon_speed == null) && fulltile)
+		decon_speed = 2 SECONDS
 
 	var/rods = 0
 	if(reinf)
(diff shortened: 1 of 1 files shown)


KasparoVy:
AFAIK this was added to stop macro memes and to make it so you can't just swooce your way through glass like it's nothing

SabreML:
> AFAIK this was added to stop macro/hyperspeed hotkey memes and to make it so you can't just swooce your way through glass like it's nothing
> 
> Can't locate the old PR but maybe @Kyep has an opinion on this

Text macros wouldn't be able to deconstruct stuff anyway, and something like an autoclicker would only be delayed by 0.3 seconds.

SteelSlayer (approved):

Carthusia:
So THIS is why when i rapidly try to attach or detach a window, it sometimes skips a step, or does no steps at all.

Its so frustrating when your doing an entire line of windows,and have to recheck each and every one of them because some may not have registered you were using tools on them, or they missed one tool and did not secure fully.

code/game/objects/structures/window.dm
```diff
@@ -88,11 +88,8 @@ GLOBAL_LIST_INIT(wcCommon, pick(list("#379963", "#0d8395", "#58b5c3", "#49e46e",
 		shards++
 		setDir()
 
-	if(decon_speed == null)
-		if(fulltile)
-			decon_speed = 20
-		else
-			decon_speed = 1
+	if((decon_speed == null) && fulltile)
```
farie82, line 91:
```suggestion
	if(decon_speed == null && fulltile)
```

farie82 (changes requested):

farie82 (approved):

Kyep (approved):

Spacemanspark:
Frankly I've always seen this as obnoxious, and causing me to accidentally smack a window with a wrench more than anything.

Carthusia:
I dono Dashora, I wouldnt really call it a skillcheck, as I swear sometimes it causes things to glitch out.

So say i have 3 reinforced windows in a row and I wish to secure them in place, if i screwdriver, crowbar, screwdriver on them, one window at a time, at a rapid pace there is a good chance that due to the .1 delay that some tools wont even register being used, it will SAY they got used, and it will show the timebar and make the sound, but it will have not applied some or any of the tools, leaving it unsecured and able to be pushed around.

This happens all the time when i try to build areas with lots of windows.

S34NW:
The 0.1 second delay can be extremely infuriating as a player, not to mention having two messages like "Starting to screw ..." "... was screwed" for each step just fills up your chat instantly when working on multiple windows.

DevL-Whats-a-jpg:
Yea i hate the delay, resulted in me burning my ass off twice when i was new to atmos and making burn chambers.

PR #15041, opened 2020-12-02
513 requirement, user-friendly byond version requirement handling

Kyep:
## What Does This PR Do
Players with a byond install too old to play on our server at all will no longer simply have their connection silently die (with the possibility of nasty string errors).
Instead, they'll be able to connect, but not ready up, observe or latejoin. Trying to do so will tell them to update their client.
Raises the absolute minimum version required to play on our server to Byond 513 (which has been the stable version for awhile now).

## Why It's Good For The Game
Prevents people with VERY out of date clients from being confused about why they cannot connect to Paradise.
Stops people complaining about bugs that only happen because their byond version is many years out of date.

## Changelog
:cl: Kyep
tweak: improved error messaging for players with clients too old to be usable.
tweak: Byond stable (513) is now required to play. If you do not have it, you will be required to update your byond.
/:cl:


AffectedArc07:
Deeming this part of #15014

code/modules/client/client_procs.dm
```diff
@@ -1090,6 +1088,8 @@
 	if(notify && (byondacc_age < config.byond_account_age_threshold))
 		message_admins("[key] has just connected for the first time. BYOND account registered on [byondacc_date] ([byondacc_age] days old)")
 
+/client/proc/show_update_notice()
+	to_chat(src, "<span class='userdanger'>Your BYOND client (v: [byond_version].[byond_build]) is out of date. This can cause glitches. We highly suggest you download the latest client from <a href='https://www.byond.com/download/'>Byond.com</a> before playing.</span>")
```
AffectedArc07, line 1092:
```suggestion
	to_chat(src, "<span class='userdanger'>Your BYOND client (v: [byond_version].[byond_build]) is out of date. This can cause glitches. We highly suggest you download the latest client from <a href='https://www.byond.com/download/'>byond.com</a> before playing. You can also update via the BYOND launcher application.</span>")
```

May as well suggest this while were here

AffectedArc07 (changes requested):

code/modules/client/client_defines.dm
```diff
@@ -90,6 +90,9 @@
 	// Last world.time that the player tried to request their resources.
 	var/last_ui_resource_send = 0
 
+	// If true, client cannot ready up, late join, or observe. Used for players with EXTREMELY old byond versions.
+	var/version_blocked = FALSE
```
AffectedArc07, line 94:
```suggestion
	/// If true, client cannot ready up, late join, or observe. Used for players with EXTREMELY old byond versions.
	var/version_blocked = FALSE
```
Autodoc compliance suggested

AffectedArc07 (approved):

PR #15048, opened 2020-12-04
Fixes an unitentional Skrell nerf

SabreML:
## What Does This PR Do

Reverts the ~~*(presumably)* unintended~~ Skrell nerf in #13430 in order to make Skrell immune to obesity again.

Edit: Apparently it wasn't unintentional, but I still think this needs to be changed back since #13430 was advertised as a sprite removal only, and not a species nerf.

## Why It's Good For The Game

image

## Changelog
:cl:
tweak: Re-adds the 'Cannot become obese' Skrell trait
/:cl:


AffectedArc07:
While technically a fix, this is a game-altering chance, so it needs heads to sign off on it.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -597,7 +597,7 @@
 			if(overeatduration < 100)
 				becomeSlim()
 		else
-			if(overeatduration > 500)
+			if(overeatduration > 500 && !(NO_OBESITY in dna.species.species_traits))
```
SteelSlayer, line 600: I think this check would be better placed with the hunger trait check on line 595. It skips having to check `FAT` or `overeatduration`, which are things obese immune species don't care about.

SabreML: Just tested it out, and it can't be put on line 595 because that'd block off all of the hunger/food code. (e.g. Plasmamen)

Doing something like this would work:
```cs
if(!(NO_HUNGER in dna.species.species_traits))
    if(!(NO_OBESITY in dna.species.species_traits))
        if(FAT in mutations)
            if(overeatduration < 100)
                becomeSlim()
        else
            if(overeatduration > 500)
                becomeFat()
```
But I think it would probably just be better to leave the check where it is, considering that you're more likely *not* to have the `NO_OBESITY` flag than to have it.

SteelSlayer: Yeah that's my fault for not looking at what comes after that code. Still, every non-fat mob, regardless of species is going to have to check for the `NO_OBESITY` flag. But ultimately it probably doesn't matter where it goes, so nevermind.

SteelSlayer (approved):

Dumbdumn5 (approved):
Got some approval for the sake of keeping racial buffs/nerfs, however small, consistent unless it was fully intended as a rework for em.

AffectedArc07 (approved):

PR #15051, opened 2020-12-05
Craftable donut boxes

SabreML:
## What Does This PR Do

Allows donut boxes to be crafted using one sheet of cardboard.
Also added an 'empty' variable to storage items.

## Why It's Good For The Game

image

## Changelog
:cl:
tweak: Donut boxes can now be crafted with cardboard
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -12,6 +12,7 @@
 	var/silent = 0 // No message on putting items in
 	var/list/can_hold = new/list() //List of objects which this item can store (if set, it can't store anything else)
 	var/list/cant_hold = new/list() //List of objects which this item can't store (in effect only if can_hold isn't set)
+	var/empty = FALSE // Will this spawn as an empty box
 	var/max_w_class = WEIGHT_CLASS_SMALL //Max size of objects that this object can store (in effect only if can_hold isn't set)
 	var/max_combined_w_class = 14 //The sum of the w_classes of all the items in this storage item.
 	var/storage_slots = 7 //The number of storage slots in this container.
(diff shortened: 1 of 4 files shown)


code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -12,6 +12,7 @@
 	var/silent = 0 // No message on putting items in
 	var/list/can_hold = new/list() //List of objects which this item can store (if set, it can't store anything else)
 	var/list/cant_hold = new/list() //List of objects which this item can't store (in effect only if can_hold isn't set)
+	var/empty = FALSE // Will this spawn as an empty box
```
AffectedArc07, line 15: I dont like this being made just for donut boxes. Either refactor all empty storage boxes to use this var, or remove it.

AffectedArc07 (changes requested):

code/game/objects/items/weapons/storage/storage.dm
SabreML: As far as I can tell the only way to avoid using an `empty` variable for donut boxes would be to make `/obj/item/storage/fancy/donut_box` the empty one, and add a `filled` subtype.
That of course is out of the question since it would need map edits, so I guess I'll get started on a storage refactor.

FlattestGuitar: Don't use a variable for this. Just make a separate method called during `New()` that fills a storage item. 

Way easier to override and no hidden quirks.

AffectedArc07 (approved):

AffectedArc07:
Merging since this was meant to be part of the other PR and I just wasnt thinking.

Spacemanspark:
Arc abusing maint powers to early merge PRs confirmed

PR #15052, opened 2020-12-05
Fixes Exosuit Fabricator runtime when creating non-objs (e.g. IRC body)

dearmochi:
## What Does This PR Do

- Title. Happens on any produced atom that isn't a `/obj/item`.

## Why It's Good For The Game

Bugfix

## Changelog
:cl:
fix: Fix a runtime when fabricating certain things in the Exosuit Fabricator (e.g. IRC body)
/:cl:

### DIFF
--- a/code/game/mecha/mech_fabricator.dm
+++ b/code/game/mecha/mech_fabricator.dm
@@ -204,7 +204,8 @@
 		L.origin_tech = I.origin_tech
 	else
 		I.forceMove(get_step(src, SOUTH))
-	I.materials = final_cost
+	if(istype(I))
+		I.materials = final_cost
 
 	// Clean up
 	being_built = null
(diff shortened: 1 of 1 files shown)


code/game/mecha/mech_fabricator.dm
```diff
@@ -204,7 +204,8 @@
 		L.origin_tech = I.origin_tech
 	else
 		I.forceMove(get_step(src, SOUTH))
-	I.materials = final_cost
+	if(istype(I))
```
AffectedArc07, line 207: I feel like this entire thing should check if `D.build_path` is an item before creating it. I would dread to think what happens if someone tries to lockbox an IRC frame by accident

farie82 (approved):

AffectedArc07 (approved):

PR #15058, opened 2020-12-06
Rework of chav text replacement

S34NW:
## What Does This PR Do
**Updates the chav disability:**
Replaces individual replacetext() calls with one regex function and a list of words to check against.
Adds more chav terms, removes or updates disused terms.
Stops replacetext() calls changing matches found in the middle of other words, now only changes whole words.

## Why It's Good For The Game
Little things like this add immersion and variety, so having a few extra changes at minimal performance cost seems like a positive change.

Matches found in the middle of words no longer get replaced, so you don't end up looking like a tool by saying "com**mate**d" every time you try and say "com**man**d".

Multiple replacetext() calls weren't the most efficient way to code this originally, one list with a single regex check is easier to maintain.

Also, who doesn't want more chavs infesting space in the far future?
image

## Changelog
:cl:
add: Added more chav words, innit.
del: Removed some old words.
tweak: Replaced many replacetext() calls with a single regex function that only changes whole words.
/:cl:

### DIFF
--- a/code/game/dna/genes/goon_disabilities.dm
+++ b/code/game/dna/genes/goon_disabilities.dm
@@ -84,13 +84,7 @@
 	activation_message = "Ye feel like a rite prat like, innit?"
 	deactivation_message = "You no longer feel like being rude and sassy."
 	mutation = CHAV
-
-/datum/dna/gene/disability/speech/chav/New()
-	..()
-	block = GLOB.chavblock
-	
-//Normal word first, chav word second
-/var/static/list/chavlinks = list(
+	/var/static/list/chavlinks = list(
 	"arrest" = "nick",
 	"arrested" = "nicked",
 	"ass" = "arse",
@@ -151,7 +145,14 @@
 	"yikes" = "blimey",
 	"you" = "u",
 	"your" = "ur"
-)
+	)
+
+/datum/dna/gene/disability/speech/chav/New()
+	..()
+	block = GLOB.chavblock
+	
+//Normal word first, chav word second
+
 /datum/dna/gene/disability/speech/chav/OnSay(mob/M, message)
 	var/static/regex/R = regex("\\b([chavlinks.Join("|")])\\b", "g")
 	message = R.Replace(message, /proc/rep_test)
--- a/code/game/dna/genes/goon_disabilities.dm
+++ b/code/game/dna/genes/goon_disabilities.dm
@@ -81,41 +81,85 @@
 /datum/dna/gene/disability/speech/chav
 	name = "Chav"
 	desc = "Forces the language center of the subject's brain to construct sentences in a more rudimentary manner."
-	activation_message = "Ye feel like a reet prat like, innit?"
+	activation_message = "Ye feel like a rite prat like, innit?"
 	deactivation_message = "You no longer feel like being rude and sassy."
 	mutation = CHAV
 
 /datum/dna/gene/disability/speech/chav/New()
 	..()
 	block = GLOB.chavblock
-
+	
+//Normal word first, chav word second
+var/global/chavlinks = list(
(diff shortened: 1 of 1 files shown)


S34NW:
oops let me stop that var being global

FlattestGuitar:
So this is great, but it should definitely just work like this for all speech disabilities. Move the behavior up to `/datum/dna/gene/disability/speech/OnSay` and add a variable to each speech disability that holds the list of replacements. The generic implementation of this will just look up the list of replacements of the specific disability it's processing.

Watch out for stuff like the swedish replacements, they aren't all meant to match word boundaries. You'll need a separate list for these.

This seems daunting, but it will result in much cleaner code and you'll be able to learn a bit about how inheritance works here.

S34NW:
> This seems daunting, but it will result in much cleaner code and you'll be able to learn a bit about how inheritance works here.

Thanks for the feedback, I agree! But I think figuring that out will take me a little longer than this did...

code/game/dna/genes/goon_disabilities.dm
```diff
@@ -81,41 +81,84 @@
...
-	message = replacetext(message,"windows","windies")
-	message = replacetext(message,"window","windy")
-	message = replacetext(message,"break","do")
-	message = replacetext(message,"your","yer")
-	message = replacetext(message,"security","coppers")
+	var/static/regex/R = regex("\\b([chavlinks.Join("|")])\\b", "g")
+	message = R.Replace(message, /datum/dna/gene/disability/speech/chav/proc/rep_test)
 	return message
-
+/datum/dna/gene/disability/speech/chav/proc/rep_test(matched)
```
farie82, line 159:
```suggestion
/datum/dna/gene/disability/speech/chav/proc/replace_speech(matched)
```
`rep_test` doesn't tell much about what the proc is supposed to do. Would also make the procname better suited for all other speech disabilities in a future refactor

farie82 (changes requested):

code/game/dna/genes/goon_disabilities.dm
S34NW: Agreed, made the changes in the latest commit

SteelSlayer:
It seems you committed to, and made a PR from your master branch. You'll want to avoid this in the future because your master branch will become very messy. You should make a new, separate branch for every PR you create.

farie82 (approved):

SteelSlayer:
Personally I think the use of "u" and "ur" is a bit too much like net speak, but we'll see what the heads/maints say about that. Also not sure if you're aware but you'll be losing some functionality here with certain words. For example, "killed" will no longer become "banged" because the regex is looking for spaces around the original word (goes for any words with -ed endings or similar).

S34NW:
> Personally I think the use of "u" and "ur" is a bit too much like net speak, but we'll see what the heads/maints say about that. Also not sure if you're aware but you'll be losing some functionality here with certain words. For example, "killed" will no longer become "banged" because the regex is looking for spaces around the original word (goes for any words with -ed endings or similar).

Yeah, I was struggling to think how to properly convey the stereotypical chav "you", the old code had "yer" but I felt it wasn't quite there. Not a sticking point for me though and I'll happily change that depending on other's opinions.

I recognise the problem with "-ed" word endings, I'd manually put in a few entries to resolve this, but missed "_killed_" so will add that in. It's not the most flexible solution to just add a new entry for each -ed word, however I don't want to add more regex functions to check for these because it might get expensive.

(I've also just noticed there are two entries for "great", one of them needs to go. I'll make these changes once there is more feedback on you/your = u/ur)

Dumbdumn5 (approved):
Bangin'

AffectedArc07 (approved):

PR #15060, opened 2020-12-06
Adds a message to the level 7 biohazard event: tells ghosts who got infected and with what disease

SteelSlayer:
## What Does This PR Do
Adds a message to the level 7 biohazard event that tells ghosts who got infected and with what disease. Someone mentioned this in deadchat a few days ago, and I thought it would be a neat idea to have.

## Why It's Good For The Game
I think it's a nice QoL thing for ghosts. Instead of having to track down patient zero, they can jump to them.

## Images of changes
I got some randomly-generated disease name here which is why you've never heard of it.

u2KdpSl

## Changelog
:cl:
add: Adds a message to the level 7 biohazard event that tells ghosts who got infected and with what disease
/:cl:

### DIFF
--- a/code/modules/events/disease_outbreak.dm
+++ b/code/modules/events/disease_outbreak.dm
@@ -1,6 +1,9 @@
 /datum/event/disease_outbreak
 	announceWhen = 15
+	/// The type of disease that patient zero will be infected with.
 	var/datum/disease/D
+	/// The initial target of the disease.
+	var/mob/living/carbon/human/patient_zero
 
 /datum/event/disease_outbreak/setup()
 	announceWhen = rand(15, 30)
@@ -18,6 +21,10 @@
 
 /datum/event/disease_outbreak/announce()
 	GLOB.event_announcement.Announce("Confirmed outbreak of level 7 major viral biohazard aboard [station_name()]. All personnel must contain the outbreak.", "Biohazard Alert", new_sound = 'sound/AI/outbreak7.ogg')
+	for(var/p in GLOB.dead_mob_list)
+		var/mob/M = p
+		if(isobserver(M) || M.stat == DEAD)
+			to_chat(M, "<span class='deadsay'><b>[patient_zero]</b> has been infected with <b>[D.name]</b> ([ghost_follow_link(patient_zero, M)])</span>")
 
 /datum/event/disease_outbreak/start()
 	for(var/mob/living/carbon/human/H in shuffle(GLOB.alive_mob_list))
@@ -33,4 +40,5 @@
 
 		if(!H.ForceContractDisease(D))
 			continue
+		patient_zero = H
 		break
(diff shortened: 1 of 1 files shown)


code/modules/events/disease_outbreak.dm
```diff
@@ -18,6 +21,10 @@
 
 /datum/event/disease_outbreak/announce()
 	GLOB.event_announcement.Announce("Confirmed outbreak of level 7 major viral biohazard aboard [station_name()]. All personnel must contain the outbreak.", "Biohazard Alert", new_sound = 'sound/AI/outbreak7.ogg')
+	for(var/p in GLOB.dead_mob_list)
+		var/mob/M = p
+		if(isobserver(M) || M.stat == DEAD)
```
farie82, line 26: Should always be the case right? Either a ghost or a dead mob

SteelSlayer: In theory it should yeah. Some other ghost messages I saw had these sanity checks as well, such as the message shown when a player dies, so I just included them here to be safe. Think I should remove them?

farie82: I see it in the code yeah. Tons of checks if it's a `new_player` or dead etc....
I just looked at the code where things are added and removed to `dead_mob_list` and it works as it says in the comment. Only gets added when something dies or is dead from the start and gets removed when it becomes alive again.
`new_player` mobs being the exception since they don't call their parents `Initialize`
So it should be fine to remove this. If there are bugs they better get solved than to add weird checks to lists that should not need those checks

farie82 (approved):

AffectedArc07:
I'm fine with saying *what* the disease is, but I don't want the message to have *who* got it. Feels a little too easy to meta it that way and have virologists be given subtle tipoffs by their buddies.

variableundefined (changes requested):
Echoing above

farie82:
> I'm fine with saying _what_ the disease is, but I don't want the message to have _who_ got it. Feels a little too easy to meta it that way and have virologists be given subtle tipoffs by their buddies.

Most viruses are fairly easy to spot as a ghost anyway once you search long enough using a medhud. This concern only goes for viruses that somehow are stealthy enough to not be seen by the medhuds.

SteelSlayer:
@AffectedArc07 @variableundefined 
Yeah, usually someone will just hunt around with the medhud for a minute or two and find them anyway. But if you want, I will change it so it says something like "Patient zero has been infected with X".

variableundefined (approved):
Taking off objection, approving.

Dumbdumn5 (approved):
Considering the medhud bit mentioned, approving this.

AffectedArc07:
> @AffectedArc07 @variableundefined
> Yeah, usually someone will just hunt around with the medhud for a minute or two and find them anyway. But if you want, I will change it so it says something like "Patient zero has been infected with X".

I'll approve once this bit is sorted.

SteelSlayer:
I'm actually going leave this as is, I don't want to remove the original reason I made the PR, which was showing ghosts who was infected. So with that, feel free to close or merge, at your discretion.

PR #15062, opened 2020-12-07
Stops crew from pushing massive spiders

DevL-Whats-a-jpg:
## What Does This PR Do
Sets the move_resist of the following terror spiders to "STRONG" (2000) making it so the crew cannot push/pull them. I may do the same with brown terrors but that will make them capable of literally walking in a line and deleting everything they touch.
The fucking EMPRESS.
Prince of Terror.
Queen of Terror.
Princess of Terror.
Purple Terror Spider.
Red Terror Spider.
## Why It's Good For The Game
Being able to push a several hundred, if not thousand pound spider isn't realistic, it also makes it near impossible to escape a lot of scenarios.

## Changelog
:cl:
tweak: Makes larger terror spiders un-push/pull able.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/empress.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/empress.dm
@@ -24,7 +24,7 @@
 	projectiletype = /obj/item/projectile/terrorqueenspit/empress
 	icon = 'icons/mob/terrorspider64.dmi'
 	pixel_x = -16
-	move_resist = STRONG // no more pushing a several hundred if not thousand pound spider
+	move_resist = MOVE_FORCE_STRONG // no more pushing a several hundred if not thousand pound spider
 	mob_size = MOB_SIZE_LARGE
 	icon_state = "terror_empress"
 	icon_living = "terror_empress"
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/empress.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/empress.dm
@@ -24,6 +24,7 @@
 	projectiletype = /obj/item/projectile/terrorqueenspit/empress
 	icon = 'icons/mob/terrorspider64.dmi'
 	pixel_x = -16
+	move_resist = STRONG // no more pushing a several hundred if not thousand pound spider
 	mob_size = MOB_SIZE_LARGE
 	icon_state = "terror_empress"
 	icon_living = "terror_empress"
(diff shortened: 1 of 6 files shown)


Kyep:
In general, its probably a good idea to ask me before doing a PR that affects terror spider balance. It is in my changes-wanted list of "maybe"s - stuff I ask people discuss with me before doing.

That said, I have been considering doing this same thing myself, but for different reasons.
Yes, terrors are at least as big as a human, so intuitively its odd that a human can push or drag them.
The really odd thing though is that terrors have 8 legs, which makes them exceptionally stable on the ground. Basically, good luck trying to push or pull something that has 8 legs and likely claws it can dig into the ground to stop you moving it. It doesn't seem realistic that you can do that, especially when it is at least as big as you.
If you even could pull/push a terror, it'd likely expose you to savage melee attacks in a way that is not currently modeled, but perhaps should be if they remain pushable/pullable.

Perhaps the most gameplay impact is that people sometimes drag terrors away from vents when they're trying to ventcrawl, which is... kinda cheap and shitty, really.
Ventcrawling terrors are meant to be able to run away if hopelessly outmatched. Denying them this is... one of the tactics that I've been keeping a watchful eye on in terms of crew strategies which might be unfair.
Crew are meant to have to kill terrors by taking them out before they can escape. Denying them the ability to escape at all should probably require more effort than just ctrl-clicking them and running away from the vent they're trying to enter.

Anyway. While I do believe this change makes sense, I'm also hesitant to support it yet.
The main reason is that I did do another terror PR recently, and I'd like to keep monitoring terror spider balance for at least a few weeks before I consider having any more PRs change it.
Basically, I'm cautious about changing too much about terrors balance-wise in the same month. I don't want to risk accidentally over-buffing or over-nerfing them by doing too many balance changes in a short period of time.

That's only a temporary reason for me not supporting this though.
Once its been about a month, and I'm sure it won't be a balance problem, I'll probably end up supporting this.

If I were to suggest changes, the change would be: simply apply the move resist to all terrors. All of them have 8 legs, and their size doesn't differ that much (exception for the empress). Their move resist is primarily (in my mind, at least) based on their numerous legs which provide them great control over their motion. Their size / muscular strength is also a factor, but the biggest factor is their legs, and since their legs are common to all of them, their move resist probably should be too. It would make intuitive sense. The only one I'd give a different move resist would be the empress, she's she's enormous compared to the others. She's the same size/strength as a megafauna, and so should probably have megafauna-level move resist. But that's an absolutely tiny quibble given that she's adminspawn-only, almost never used and when she is her move resist doesn't really come into play.

Kyep (approved):
Approving. Its been 2 weeks, and enough other stuff has happened in the meantime that I'm comfortable voting for this now.

farie82 (approved):

AffectedArc07 (approved):

PR #15063, opened 2020-12-07
Bar sign maint/dismantle bug fix.

### LINKED ISSUE #14884: Bar Sign is invisible once removed
**Issue Description**:
The bar sign turns invisible when you remove it from the wall with a screwdriver

To have a sign that was able to be moved elsewhere, or one that can't move.

**What happened instead**:
Sign turned invisible, was still interactable via ALT-clicking the tile where it was, and interacting with it via the "floor" tab. When attempting to place it back on a wall, it remained invisible.

Invisible items are bad, plus it would be nice to be able to move the bar sign around for bar rearranging.

Unscrew sign with screwdriver.

Tuulipommi:
## What Does This PR Do

Removed the ability to unscrew and dismantle
the bar sign, as it caused the bar sign to
disappear, without a way to make it visible again.
This was done, as there were no sprites for either of these functions.
fixes #14884
## Why It's Good For The Game

As stated in the [issue report](https://github.com/ParadiseSS13/Paradise/issues/14884), invisible objects aren't really a good addition to the game. Yes, its a cosmetic one, but it can be a bit distracting.
## Changelog
:cl: Windbomb
fix: you can no longer break the bar sign by trying to disassemble it
/:cl:

### DIFF
--- a/code/game/objects/structures/barsign.dm
+++ b/code/game/objects/structures/barsign.dm
@@ -44,10 +44,6 @@
 	if(!broken && !(flags & NODECONSTRUCT))
 		broken = TRUE
 
-/obj/structure/sign/barsign/deconstruct(disassembled = TRUE)
-	new /obj/item/stack/sheet/metal(drop_location(), 2)
-	new /obj/item/stack/cable_coil(drop_location(), 2)
-	qdel(src)
 
 /obj/structure/sign/barsign/play_attack_sound(damage_amount, damage_type = BRUTE, damage_flag = 0)
 	switch(damage_type)
@@ -59,6 +55,9 @@
 /obj/structure/sign/barsign/attack_ai(mob/user as mob)
 	return src.attack_hand(user)
 
+// Added to prevent people from unscrewing the barsign, making go invisible in the process
+obj/structure/sign/barsign/screwdriver_act()
+    return
 
 
 /obj/structure/sign/barsign/attack_hand(mob/user as mob)
@@ -69,42 +68,6 @@
 		to_chat(user, "<span class ='danger'>The controls seem unresponsive.</span>")
 		return
 	pick_sign()
-
-
-
-
-/obj/structure/sign/barsign/attackby(var/obj/item/I, var/mob/user)
-	if( istype(I, /obj/item/screwdriver))
-		if(!panel_open)
-			to_chat(user, "<span class='notice'>You open the maintenance panel.</span>")
-			set_sign(new /datum/barsign/hiddensigns/signoff)
-			panel_open = 1
-		else
-			to_chat(user, "<span class='notice'>You close the maintenance panel.</span>")
-			if(!broken && !emagged)
-				set_sign(pick(barsigns))
-			else if(emagged)
-				set_sign(new /datum/barsign/hiddensigns/syndibarsign)
-			else
-				set_sign(new /datum/barsign/hiddensigns/empbarsign)
-			panel_open = 0
-
-	if(istype(I, /obj/item/stack/cable_coil) && panel_open)
-		var/obj/item/stack/cable_coil/C = I
-		if(emagged) //Emagged, not broken by EMP
-			to_chat(user, "<span class='warning'>Sign has been damaged beyond repair!</span>")
-			return
-		else if(!broken)
-			to_chat(user, "<span class='warning'>This sign is functioning properly!</span>")
-			return
-
-		if(C.use(2))
-			to_chat(user, "<span class='notice'>You replace the burnt wiring.</span>")
-			broken = 0
-		else
-			to_chat(user, "<span class='warning'>You need at least two lengths of cable!</span>")
-	else
-		return ..()
 
 
 
--- a/code/game/objects/structures/barsign.dm
+++ b/code/game/objects/structures/barsign.dm
@@ -44,6 +44,10 @@
 	if(!broken && !(flags & NODECONSTRUCT))
 		broken = TRUE
 
+/obj/structure/sign/barsign/deconstruct(disassembled = TRUE)
+	new /obj/item/stack/sheet/metal(drop_location(), 2)
+	new /obj/item/stack/cable_coil(drop_location(), 2)
+	qdel(src)
 
 /obj/structure/sign/barsign/play_attack_sound(damage_amount, damage_type = BRUTE, damage_flag = 0)
 	switch(damage_type)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/barsign.dm
```diff
@@ -59,6 +55,9 @@
 /obj/structure/sign/barsign/attack_ai(mob/user as mob)
 	return src.attack_hand(user)
 
+// Added to prevent people from unscrewing the barsign, making go invisible in the process
+obj/structure/sign/barsign/screwdriver_act()
```
FlattestGuitar, line 59: Add a `/` at the start of this line.

FlattestGuitar:
The changelog is for the players. They don't care about how many pieces of code you added or removed. 

How about `fix: you can no longer break the bar sign by trying to disassemble it`?

code/game/objects/structures/barsign.dm
Tuulipommi: Fixed

code/game/objects/structures/barsign.dm
```diff
@@ -72,42 +71,6 @@
...
-			panel_open = 1
-		else
-			to_chat(user, "<span class='notice'>You close the maintenance panel.</span>")
-			if(!broken && !emagged)
-				set_sign(pick(barsigns))
-			else if(emagged)
-				set_sign(new /datum/barsign/hiddensigns/syndibarsign)
-			else
-				set_sign(new /datum/barsign/hiddensigns/empbarsign)
-			panel_open = 0
```
farie82, line 90: This part of the code should be part of the `screwdriver_act` proc. You used to be able to screw the panel open and the barsign had special logic for it to avoid the weirdness we have now.

code/game/objects/structures/barsign.dm
```diff
@@ -72,42 +71,6 @@
...
-			to_chat(user, "<span class='warning'>This sign is functioning properly!</span>")
-			return
-
-		if(C.use(2))
-			to_chat(user, "<span class='notice'>You replace the burnt wiring.</span>")
-			broken = 0
-		else
-			to_chat(user, "<span class='warning'>You need at least two lengths of cable!</span>")
-	else
-		return ..()
```
farie82, line 107: This can all stay in the `attackby` proc since then you can repair the barsign

code/game/objects/structures/barsign.dm
```diff
@@ -44,10 +44,6 @@
 	if(!broken && !(flags & NODECONSTRUCT))
 		broken = TRUE
 
-/obj/structure/sign/barsign/deconstruct(disassembled = TRUE)
-	new /obj/item/stack/sheet/metal(drop_location(), 2)
-	new /obj/item/stack/cable_coil(drop_location(), 2)
-	qdel(src)
```
farie82, line 50: Should remain here. This is called when you destroy the sign which makes it drop the items listed above

farie82 (changes requested):
Be sure to also include the following to your PR description:
fixes #14884

This will ensure that once this PR gets merged the issue gets closed since it is fixed

code/game/objects/structures/barsign.dm
```diff
@@ -59,22 +59,27 @@
...
-	pick_sign()
-
-
-
-
-/obj/structure/sign/barsign/attackby(var/obj/item/I, var/mob/user)
-	if( istype(I, /obj/item/screwdriver))
+// Added to prevent people from unscrewing the barsign, making go invisible in the process
+/obj/structure/sign/barsign/screwdriver_act()
+    if( istype(I, /obj/item/screwdriver))
```
variableundefined, line 82: Spacing

code/game/objects/structures/barsign.dm
```diff
@@ -89,22 +94,15 @@
...
 
-		if(C.use(2))
-			to_chat(user, "<span class='notice'>You replace the burnt wiring.</span>")
-			broken = 0
-		else
-			to_chat(user, "<span class='warning'>You need at least two lengths of cable!</span>")
-	else
-		return ..()
+/obj/structure/sign/barsign/attack_hand(mob/user as mob)
+	if(!src.allowed(user))
```
variableundefined, line 99: Implicit src, please remove

variableundefined (changes requested):
Also fix the things the linter pointed out. Please do not webedit.

code/game/objects/structures/barsign.dm
farie82, line 82: This check should also not exist. `screwdriver_act` means an item was used with the property of being a screwdriver tool. The old way used istype checks

code/game/objects/structures/barsign.dm
```diff
@@ -89,22 +94,15 @@
...
+		to_chat(user, "<span class = 'info'>Access denied.</span>")
```
farie82, line 100:
```suggestion
		to_chat(user, "<span class='info'>Access denied.</span>")
```
spacing in html string is done like this. In DM it's correct though to put a space in between

code/game/objects/structures/barsign.dm
```diff
@@ -89,22 +94,15 @@
...
+		return
+	if(broken)
+		to_chat(user, "<span class ='danger'>The controls seem unresponsive.</span>")
```
farie82, line 103:
```suggestion
		to_chat(user, "<span class='danger'>The controls seem unresponsive.</span>")
```

code/game/objects/structures/barsign.dm
```diff
@@ -89,22 +94,15 @@
...
-			return
 
-		if(C.use(2))
-			to_chat(user, "<span class='notice'>You replace the burnt wiring.</span>")
-			broken = 0
-		else
-			to_chat(user, "<span class='warning'>You need at least two lengths of cable!</span>")
-	else
-		return ..()
+/obj/structure/sign/barsign/attack_hand(mob/user as mob)
```
farie82, line 98:
```suggestion
/obj/structure/sign/barsign/attack_hand(mob/user)
```
`as mob` does nothing here. It's also the old way of doing things. Nowadays just a `mob/user` is plenty

variableundefined:
@Tuulipommi Since this is a small PR, I am requesting a fix within 4 days or the PR may be closed.

Kiyahitayika:
@Tuulipommi First off, thank you for working on this and tis was a really good first PR.  Second, you are dealing with code that has not been touched in a while here.

Farie is right about the `as mob` lines. It looks like they've been there for a while - just `mob/user` is fine. The spacing in the HTML output was like that already and it's easy to fix as well. Cleaning up and tweaking old code is something that eventually becomes old-hat when working on the Paradise codebase. Contributors and Maintainers like to keep the code fresh and update it whenever anything is touched.

VariableUndefined is right about the implicit src. This happens a lot in older code. Instead of `if(!src.allowed(user))`, the code can nowadays use `if(!allowed(user))` because it recognizes the 'src' as the datum this line of code is being called on (barsign) and sees that the `allowed` proc is referring to the barsign. 
The `allowed` proc itself is passed all the way up from /obj/ through /structure/ through /signs/ to /barsign/ so we can ask "hey, does this `user` have access to this object?". Line 6 of this code file sets that access with `req_access = list(ACCESS_BAR)`.

There are three other instances of implied src in this file as well. Two are here:
```
		if(BRUTE)
			playsound(src.loc, 'sound/effects/glasshit.ogg', 75, TRUE)
		if(BURN)
			playsound(src.loc, 'sound/items/welder.ogg', 100, TRUE)

```
In both cases, `src.loc` can be `loc`. The third one is:
```
/obj/structure/sign/barsign/attack_ai(mob/user)
	return src.attack_hand(user)
```

All this to say - you are on the right path with this! I was able to get your code running with the changes suggested in the reviews (and a change in indentation). This looks like an excellent first PR and I don't want you to give up on coding if you're interested in it!

I am planning on using your work from this project to fix the same issue on a Paradise downstream server. Seriously, good work, and don't get discouraged. If you're not on the Paradise Discord, I recommend joining it and reaching out to Farie and joining the coding channel. Paradise is good people and coding is tough to do alone.

PR #15067, opened 2020-12-08
Fixes #15065 - green terrors laying eggs while in vents

### LINKED ISSUE #15065: Green Terrors can lay eggs in vents
As a green terror on live, I managed to lay eggs while ventcrawling and it put the eggs on the turf where the vent was.

Could be used as an exploit, nesting in inaccessible areas, that sort of thing.

Join as a green terror, attempt to lay eggs in a vent.

Not sure, probably forever.

**Extra information**:
May also be the case for princesses and non-terror green spiders.

Kyep:
## What Does This PR Do
Fixes #15065 
Adds some additional safety checks to a few terror spider functions (egglaying, web creation, royal jelly creation) to prevent people doing them while inside a pipe.

## Changelog
:cl: Kyep
fix: fixed terrors being able to lay eggs (and some other things) while inside a vent.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/actions.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/actions.dm
@@ -137,6 +137,9 @@
 /mob/living/simple_animal/hostile/poison/terror_spider/proc/Web()
 	if(!web_type)
 		return
+	if(!isturf(loc))
+		to_chat(src, "<span class='danger'>Webs can only be spun while standing on a floor.</span>")
+		return
 	var/turf/mylocation = loc
 	visible_message("<span class='notice'>[src] begins to secrete a sticky substance.</span>")
 	if(do_after(src, delay_web, target = loc))
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/mother.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/mother.dm
@@ -72,6 +72,9 @@
 	if(regen_points < jelly_cost)
 		to_chat(src, "<span class='danger'>You only have [regen_points] of the [jelly_cost] regeneration points you need to do this.</span>")
 		return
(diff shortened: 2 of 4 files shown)


code/modules/mob/living/simple_animal/hostile/terror_spiders/actions.dm
```diff
@@ -137,6 +137,9 @@
 /mob/living/simple_animal/hostile/poison/terror_spider/proc/Web()
 	if(!web_type)
 		return
+	if(!isturf(loc))
+		to_chat(src, "<span class='danger'>Webs can only be spun while standing on a floor.</span>")
```
farie82, line 141: the span class `warning` is usually used for these messages so they don't spam up the chat to much but are still noticeable

Kyep: They can't really spam the chat unless the user deliberately spam-clicks the web button while they're inside a pipe or similar.
In that rare case, I want it to be obvious so they learn immediately that they can't do it.

code/modules/mob/living/simple_animal/hostile/terror_spiders/mother.dm
```diff
@@ -72,6 +72,9 @@
 	if(regen_points < jelly_cost)
 		to_chat(src, "<span class='danger'>You only have [regen_points] of the [jelly_cost] regeneration points you need to do this.</span>")
 		return
+	if(!isturf(loc))
+		to_chat(src, "<span class='danger'>This can only be done while standing on a floor.</span>")
```
AffectedArc07, line 76:
```suggestion
	if(!isturf(loc))
		to_chat(src, "<span class='danger'>You can only secrete jelly while standing on a floor.</span>")
```
Maybe make it more specific than just `This`

AffectedArc07 (approved):

PR #15071, opened 2020-12-09
Cult barrier rune chaining

SabreML:
## What Does This PR Do

Adds a chaining effect to Cult barrier runes, so that activating one will also activate any adjacent runes.
This doesn't work for deactivating though, so they still need to be individually turned off.

## Why It's Good For The Game

QOL improvement for Cultists.
*(Could maybe be considered a buff, but I'd say it's a very small one.)*

### Before:
XuoRET8ENl
### After:
cgDXMwC83P

## Changelog
:cl:
tweak: When activating a Cult barrier rune, any adjacent barrier runes will also be enabled (This still includes the invoke damage for each activated rune)
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/runes.dm
+++ b/code/game/gamemodes/cult/runes.dm
@@ -634,7 +634,7 @@ structure_check() searches for nearby cultist structures required for the invoca
 //Rite of the Corporeal Shield: When invoked, becomes solid and cannot be passed. Invoke again to undo.
 /obj/effect/rune/wall
 	cultist_name = "Barrier"
-	cultist_desc = "when invoked, makes a temporary invisible wall to block passage. Can be destroyed by brute force. Can be invoked again to reverse this."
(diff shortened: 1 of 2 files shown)


datlo:
I've removed this as it felt too easy to create massive barriers around the summoning location at little cost to the user. Toggling the rune costs health (admittedly, not much, but still).

If this is added, I think the toggle cost should be raised for every rune toggled by the chain effect.

SabreML:
> I've removed this as it felt too easy to create massive barriers around the summoning location at little cost to the user. Toggling the rune costs health (admittedly, not much, but still).
> 
> If this is added, I think the toggle cost should be raised for every rune toggled by the chain effect.

I wasn't sure about increasing the health cost since I don't want people to misclick and somehow put themselves into crit, but since it's only a maximum of 18 damage it does make sense to include it.

variableundefined (approved):

Kyep (approved):

PR #15073, opened 2020-12-09
Makes terrors less strong when pulling depending on their tier/class

farie82:
## What Does This PR Do
Makes lesser terror spiders unable to pull big objects behind them. Any machine or structure that can be moved is usually considered a bigger object in this case. Items of size bulky or bigger are also considered to big. They can still pull mobs without an issue

Tier 3 and higher spiders can drag things like they normally could since they are stronger and bigger. The brown and purple spiders can also do this even though they are tier 2 spiders.

## Why It's Good For The Game
Reduces the cheese terrors use. Dragging either undestroyable/hard to destroy objects behind them to block laser shots or dragging fuel/water tanks behind them to instantly kill people.

## Changelog
:cl:
balance: Terror spiders of tier 2 and lower now can't drag bigger objects behind them. They can still drag mobs just fine. Exceptions are the brown and purple spiders.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/brown.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/brown.dm
@@ -23,6 +23,7 @@
 	spider_opens_doors = 2 // Breach specialist.
 	environment_smash = ENVIRONMENT_SMASH_RWALLS // Breaks anything.
 	spider_tier = TS_TIER_2
+	pull_force = PULL_FORCE_DEFAULT
 	ai_ventbreaker = 1
 	freq_ventcrawl_combat = 600 // Ventcrawls very frequently, breaking open vents as it goes.
 	freq_ventcrawl_idle =  1800
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/purple.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/purple.dm
@@ -21,6 +21,7 @@
 	melee_damage_lower = 15
 	melee_damage_upper = 25
 	spider_tier = TS_TIER_2
+	pull_force = PULL_FORCE_DEFAULT
 	move_to_delay = 5 // at 20ticks/sec, this is 4 tile/sec movespeed, same as a human. Faster than a normal spider, so it can intercept attacks on queen.
 	speed = 0 // '0' (also the default for human mobs) converts to 2.5 total delay, or 4 tiles/sec.
 	spider_opens_doors = 2
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/terror_spiders.dm
@@ -296,6 +296,16 @@ GLOBAL_LIST_EMPTY(ts_spiderling_list)
 	U.add_hud_to(src)
 	spider_creation_time = world.time
 
+/mob/living/simple_animal/hostile/poison/terror_spider/Initialize(mapload)
+	if(!pull_force)
+		switch(spider_tier)
+			if(TS_TIER_3, TS_TIER_4)
+				pull_force = MOVE_FORCE_DEFAULT
+			else
+				pull_force = MOVE_FORCE_WEAK
+	return ..()
+
+
 /mob/living/simple_animal/hostile/poison/terror_spider/proc/announcetoghosts()
 	if(spider_awaymission)
 		return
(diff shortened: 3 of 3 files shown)


Kyep:
Objection.
First, terrors are on my changes-wanted "please ask me before changing this" list, and you did not ask. You really should have. There's not much point in heads or maints having a "please discuss with me before you create a PR that alters this" list if contributors ignore said list and proceed anyway. I realize you said in Discord that you'd have created this regardless of my objection... but had you asked first at least you'd have understood my concerns better.
Second, there is no list here of what classifies as a "bigger object". You do say "water tanks and such", and mobs don't count, but what about everything else? Would, say, a displaced girdir qualify? What about an unanchored vending machine? Or similar? There are quite a lot of draggable objects that can be used as barriers. This goes way further than just tanks or canisters.
Third, I'm worried about the implications of this. While I have considered removing their ability to drag gas canisters in particular, this goes much further than that. Yes, moving water tanks has combat usage but I was under the impression that eventually we'd just nerf water slips in general and that would be less of an issue. And design-wise I do not want to cut them off from interacting with their environment too much. There's also the issue of crew exploiting this to create barriers that crew can move, but terrors can't.

Overall... while I can see what you're trying to do, I'm really not a fan of the method by which you've done it, for all three reasons.

farie82:
> Second, there is no list here of what classifies as a "bigger object". You do say "water tanks and such", and mobs don't count, but what about everything else? Would, say, a displaced girdir qualify? What about an unanchored vending machine? Or similar? There are quite a lot of draggable objects that can be used as barriers. This goes way further than just tanks or canisters.

I'll include more concrete examples. But all structures/machines that can be moved are considered bigger in this case. Also objects which are normal sized or bigger.

> Third, I'm worried about the implications of this. While I have considered removing their ability to drag gas canisters in particular, this goes much further than that. Yes, moving water tanks has combat usage but I was under the impression that eventually we'd just nerf water slips in general and that would be less of an issue. And design-wise I do not want to cut them off from interacting with their environment too much. There's also the issue of crew exploiting this to create barriers that crew can move, but terrors can't.

They can still push objects. Just not pull them. Crew can't easily destroy things they drag so it's a free escape if they got a hold of a closed locker or such. It's cheap to play against and near uncounterable in most cases.
Water slips currently are nerfed (thank god) but still mean a death penalty if a terror uses them. Once down you stay down. Or maybe you get up once you're at 120 damage or so which gives you one more shot at the terror before kneeling over and puking blood.
The terrors can also just destroy most if not all objects the crew can use for barricades. Bigger terrors can also just pull the objects away from the front if needed. Giving the brown spider a new goal as well to clear the battlefield of obstacles instead of just being the "I destroy welded vents and walls" class.

Kyep:
That really isn't reassuring.
The first point makes me think this PR effectively translates to "most terrors can no longer pull any structure or machine".
The second point doesn't stop crew, for example, placing them in corners where they can easily be moved by pulling, but pushing them is impossible, thus creating effectively a door that a terror cannot pass.
I should also mention that there are plenty of things that the crew can do which are still death to terrors. Not just mechs and mass-arming the crew, even more common things (like simply chasing down wounded terrors off of webs) are very effective at killing them. Both terrors and crew are meant to have tactics which, while they aren't usable all the time, are still very useful in situations where they can be used. That's part of what adds depth to playing them, and playing against them. You have to maintain awareness of your environment and alert for environmental things that provide advantage or disadvantage.

Kyep (changes requested):
Above reasoning.

Spacemanspark:
>First, terrors are on my changes-wanted "please ask me before changing this" list, and you did not ask. You really should have.

This isn't how open source works. It might have been your idea and something you created, but when you put it onto a public codebase like this, anyone is free to try and make changes, with or without your approval. 

I agree with this change, as I'm frankly quite tired of seeing terrors dragging canisters, welding tanks and water tanks around. They're fun to engage with normally, but this often times just makes things far more of a curbstomp than it needs to be.

Kyep:
I too am not a fan of terrors dragging around gas canisters. I've been debating removing their ability to do that.
I'm okay with them dragging tanks a short distance, for the purpose of say, removing an obstacle, or setting a trap.
But I don't want to prevent them ever dragging anything substantial, as I feel that's really heavy-handed. I'm going to look into alternate solutions here.

Also, regarding the open-source bit: the point I am trying to make here is that I *don't* expect contributors to always agree with heads/maints, and if they want to put up PRs they know at least one head/maint will object to, they can. But if a head/maint has explicitly said "please talk to me before creating a PR that alters X", its a bit discourteous to go "well, they won't agree with my idea, so I'm just going to ignore them and make it anyway". I can remember the maints in the past being quite miffed about people doing this, as it tended to result in controversial PRs that took up a lot of time and ultimately went nowhere.

Spacemanspark:
Thing is, you don't have to agree with their idea.
One person should not be the sole arbiter of what is and isn't alright for any feature of the codebase. That's why we have the PR process, why we talk about it with the players, heads and maints alike. Yes, on occasion it might slow things down, but at least it can be properly discussed--with code ready--with the full community. So the hundreds of players who are actively using these features can give their opinions on it. I'd argue it's far more discourteous to essentially ignore them.

AffectedArc07:
Closing in favour of #15113

I feel like a brutal slowdown for all is better than cutting it from all except T3. While it would be good to remove water cheese and plasma fires on a majority, spiders can just get a prince or other to do that (Looking at you random event which has just a prince)

PR #15075, opened 2020-12-09
Methamphaldehyde - Memechem Nerf/Change

matttheficus:
## What Does This PR Do
Adds a new chemical, methamphaldehyde, to act as a sort of "buffer chemical" to prevent the mixing of Meth and Mannitol in one of the most common memechem combos. It acts as a "safer" meth-like chemical by not applying brain damage, but it doesnt mix well with other stimulants and will cause complications if done so.

**Paralysis Adjust**: -1.5 -> -2
**Stun Adjust**: -1.5 -> -2
**Weakened Adjust**: -1.5 -> -2
**Stamina Adjust**: removed -> -1.5
**Vomit Chance**: 10% -> removed
**Speed Increase**: removed

If mixed with other stimulants, it will now cause vomiting, a knockdown much like when overdosing, and will purge the offending stimulant from your system. This prevents it from being mixed with other stimulants for MASSIVELY increased anti-stun.

The discussions around other possible changes for what the effects can be are found at: https://www.paradisestation.org/forum/topic/19884-possible-memechem-nerfs-meth-mannitol/

## Why It's Good For The Game
Memechems are one of the dumbest ways for antags to get dunked on by the Crew. I haven't seen a single Nukies round where there aren't 10 people packed into Sci/MedChem making 1000u+ of Meth+Mann+Hydro. When your only answer to it is "blow them up with explosives", that's poor balance. With this PR, it addresses the issue that people can use Meth safely with zero downsides by simply adding mannitol to the mix. 

Originally, I was going to leave it at that, but Qwerty and a few others in #coding_chat mentioned how you would just add this chemical to a meth+methamphaldehyde mix and literally be an unstunnable god. So, to counter this case, I've made it so that you have to choose between the safer-yet-slower methamphaldehyde or the faster-but-dangerous methamphetamine. This allows for antags to still have a potent anti-stun in a game where the combat only revolves around stuns for the most part, but prevents them from zipping around at Mach 7 at the same time with zero downsides.

## Changelog
:cl:
add: adds methamphaldehyde, a 1:1 ratio mixture of methamphetamine and mannitol
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drugs.dm
+++ b/code/modules/reagents/chemistry/reagents/drugs.dm
@@ -360,17 +360,39 @@
 	return list(effect, update_flags)
 
 /datum/reagent/methamphaldehyde
-	name = "methamphaldehyde"
+	name = "Methamphaldehyde"
 	id = "methamphaldehyde"
-	description = ""
+	description = "A diluted mix of methamphetamine and mannitol. Its effects are weaker leading to no chance at brain damage or addiction, but it reacts poorly to one's immune system."
 	reagent_state = LIQUID
-	color = "#60A584" // rgb: 96, 165, 132
+	color = "#97d1b5" // rgb: 59, 82, 71
 	overdose_threshold = 0
 	addiction_chance = 0
 	addiction_threshold = 0
 	metabolization_rate = 0.6
 	heart_rate_increase = 1
 	taste_description = "speed"
+
+/datum/reagent/methamphaldehyde/on_mob_life(mob/living/M)
+	var/update_flags = STATUS_UPDATE_NONE
+	if(prob(5))
+		M.emote(pick("twitch_s","blink_r","shiver"))
+	if(current_cycle >= 25)
+		M.AdjustJitter(5)
+	M.AdjustDrowsy(-10)
+	update_flags |= M.AdjustParalysis(-1.5, FALSE)
+	update_flags |= M.AdjustStunned(-1.5, FALSE)
+	update_flags |= M.AdjustWeakened(-1.5, FALSE)
+	update_flags |= M.adjustStaminaLoss(-1, FALSE)
+	update_flags |= M.SetSleeping(0, FALSE)
+	M.status_flags |= GOTTAGOFAST
+	if(prob(10))
+		var/mob/living/carbon/human/C = M
+		C.vomit()
+	return ..() | update_flags
+
+/datum/reagent/methamphaldehyde/on_mob_delete(mob/living/M)
+	M.status_flags &= ~GOTTAGOFAST
+	..()
 
 /datum/reagent/bath_salts
 	name = "Bath Salts"
--- a/code/modules/reagents/chemistry/reagents/drugs.dm
+++ b/code/modules/reagents/chemistry/reagents/drugs.dm
@@ -381,8 +381,7 @@
 			var/datum/reagent/R = I
 			if(stimulant_list.Find(R.id))
 				has_stimulant = TRUE
-				if(R.id == "methamphetamine")
-					M.reagents.remove_reagent(R.id, 5)
+				M.reagents.remove_reagent(R.id, 5)
 			else
 				has_stimulant = FALSE
 	if(has_stimulant == TRUE)
--- a/code/modules/reagents/chemistry/reagents/drugs.dm
+++ b/code/modules/reagents/chemistry/reagents/drugs.dm
@@ -382,24 +382,23 @@
 			if(stimulant_list.Find(R.id))
 				has_stimulant = TRUE
 				M.reagents.remove_reagent(R.id, 5)
-			else
-				has_stimulant = FALSE
 	if(has_stimulant == TRUE)
 		if(prob(10))
 			if(iscarbon(M))
 				var/mob/living/carbon/human/C = M
 				C.vomit()
-				update_flags |= M.Weaken(15, FALSE)
+				update_flags |= M.Weaken(5, FALSE)
+	else
+		M.AdjustDrowsy(-10)
+		update_flags |= M.AdjustParalysis(-2, FALSE)
+		update_flags |= M.AdjustStunned(-2, FALSE)
+		update_flags |= M.AdjustWeakened(-2, FALSE)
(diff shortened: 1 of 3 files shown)


matttheficus:
Following some discussion in #coding_chat, some possible changes came up so im tweaking the aim a tad. 

**Paralysis Adjust**: -1.5 -> -2
**Stun Adjust**: -1.5 -> -2
**Weakened Adjust**: -1.5 -> -2
**Stamina Adjust**: removed -> -1.5
**Vomit Chance**: 10% -> removed
**Speed Increase**: removed

Also added OD and addiction changes in line with standard Meth

This makes methamphaldehyde a pure anti-stun, but without the speed, preventing its use as a memechem but still allowing for antags to make use of it.

farie82 (approved):
Code is good

farie82:
Issue I have with this is that this is a very potent drug that can be mixed with meth to get a safe even bigger boost. Only downside there is that you have to pop your mannitol pill afterwards

Spacemanspark:
Perhaps it should just be a straight up bad drug to use, without benefits.

matttheficus:
> Issue I have with this is that this is a very potent drug that can be mixed with meth to get a safe even bigger boost. Only downside there is that you have to pop your mannitol pill afterwards.

At the moment, you dont have to even wait till later to pop the mannitol, you can just mix it and ignore the downside, this at least forces you to medicate yourself during downtime, and if youre being chased, youll have to deal with banging your head into airlocks after long enough. That stun alone I think is enough to prevent you from chugging both down.

> Solution might be to add a negative effect when it detects meth in the user or something.

As in, when methamphaldehyde detects that standard meth is in the system as well? I'll take a peek in that direction, as it might help.

farie82:
> At the moment, you dont have to even wait till later to pop the mannitol, you can just mix it and ignore the downside, this at least forces you to medicate yourself during downtime, and if youre being chased, youll have to deal with banging your head into airlocks after long enough. That stun alone I think is enough to prevent you from chugging both down.

Aye it's not great currently. Far from it. But having two very potent chems which work together will negate most stuns since it'll heal 5 stun per tick instead of just the 3(?) from meth


> As in, when methamphaldehyde detects that standard meth is in the system as well? I'll take a peek in that direction, as it might help.

Yeh yeh something like that

variableundefined (commented):
I'm not familiar enough with the meta to make an insightful comment, but I share Farie's concern. 

If this ends up as an even better memechem I'll have to say no.

matttheficus:
If mixed with other stimulants, it will now cause vomiting, a knockdown much like when overdosing, and will purge the offending stimulant from your system. This prevents it from being mixed with other stimulants for MASSIVELY increased anti-stun.

code/modules/reagents/chemistry/reagents/drugs.dm
```diff
@@ -359,6 +359,77 @@
...
+	var/update_flags = STATUS_UPDATE_NONE
+	var/has_stimulant = FALSE
+	if(M.reagents.reagent_list)
+		for(var/I in M.reagents.reagent_list)
+			var/datum/reagent/R = I
+			if(stimulant_list.Find(R.id))
+				has_stimulant = TRUE
+				M.reagents.remove_reagent(R.id, 5)
+			else
+				has_stimulant = FALSE
```
farie82, line 386:
```suggestion
```
Else if you have meth and say occuline in you it will say `has_stimulants = FALSE` if occuline is last in the list

code/modules/reagents/chemistry/reagents/drugs.dm
```diff
@@ -359,6 +359,77 @@
...
+	if(has_stimulant == TRUE)
+		if(prob(10))
+			if(iscarbon(M))
+				var/mob/living/carbon/human/C = M
+				C.vomit()
+				update_flags |= M.Weaken(15, FALSE)
```
farie82, line 392: Design thing:
Might want to instead of adding a hefty weaken lower it a bit and just not have effect from the beneficial parts of this chem

Kyep (commented):
I'm not sure what to say about this.
On the one hand, I agree that taking both meth and mannitol together might be too powerful.
On the other hand, I can't recall the last time I actually saw someone do this.
I'm also a bit unclear as to the intended change here.
Is it:
- Old: people took meth and mannitol together, resulting in fast speed, resistance to stuns and no downside
- New: people can't do that as if they do the chems mix together and start stunning them?
The description needs to be a bit clearer.

matttheficus:
> On the other hand, I can't recall the last time I actually saw someone do this.

Ive seen this on nearly every Nukie round for the past month or so at minimum. 
 
> * Old: people took meth and mannitol together, resulting in fast speed, resistance to stuns and no downside
> * New: people can't do that as if they do the chems mix together and start stunning them?

This confusion might be an issue due to some edge-case issues. The only reason I made the mix-prevention a thing was due to Qwerty pointing out that being able to stack methamphaldehyde with standard meth would be insanely, gamebreakingly broken and create the worst memechem combo on the face of the planet. Ill update the description to fit what the overall goal is, as ive been editing with changes piece-by-piece rather than a full rewrite, which is probably where the disconnect is.

MRgaty:
But nuke ops kinda need anti stuns and other boosts, remember that they are 4 vs 80 or so.

Spacemanspark:
Nukies have eshields, implants and adrenals tho

matttheficus:
> But nuke ops kinda need anti stuns and other boosts, remember that they are 4 vs 80 or so.

This shouldnt even affect Nuclear Operatives, as they dont usually make their own memechems. This solely affects on-Station antags and the Crewmembers that use memechems to fight antags.

Wojz:
> On the other hand, I can't recall the last time I actually saw someone do this.

Huh? It happens all the time, every time someone uses meth they also use mannitol because it's become meta to do this.

hal9000PR:
> Perhaps it should just be a straight up bad drug to use, without benefits.

If it was a bad drug it would 1 be a major nerf to antags because antistuns are their life and blood, and it could be weaponized potentially making an even worse meme chem, an offensive memechem.

AffectedArc07:
2 maints against, no one in favour after many weeks. Closing.

PR #15098, opened 2020-12-12
Radial Menus for the Bible and Null Rod

SabreML:
## What Does This PR Do

Converts the Chaplain's Bible and Null Rod to use Radial Menus rather than a text list.
Because of this the Bible menu only opens if you activate the Bible in your hand as opposed to opening when you spawn in.
I've also removed the 20 second time limit on choosing a Bible type, since I didn't see any good reason for it to be there.

Choosing a Null Rod though is unchanged, except for the visual differences.

Also commented out the Chapel's carpet symbols changing, since it didn't actually work properly.
Also also cleaned up and documented some of the code and variables.

## Why It's Good For The Game

1. It looks better.
2. You don't need to tab out to the wiki if you want to try out a new Null Rod.
3. Choosing a bible sprite is far more flexible, and admins can also re-enable the `customisable` variable if they want to change it.

## Images of changes

*All options have highlighting and tooltips, so contrast with the background tiles shouldn't ever be an issue.*
### Bible Radial Menu
NSS Cyberiad 2020-12-12 213641NSS Cyberiad 2020-12-12 213646

### Null Rod Radial Menu
NSS Cyberiad 2020-12-12 213702NSS Cyberiad 2020-12-12 213705
NSS Cyberiad 2020-12-12 213708NSS Cyberiad 2020-12-12 213711
NSS Cyberiad 2020-12-12 213714

## Changelog
:cl:
add: Added a fancy new Radial Menu for Bibles and Null Rods!
tweak: Picking a Bible variation is now done by activating it in your hand, rather than when you spawn in
tweak: Removed the 20 second time limit when choosing a Bible
tweak: Stopped the Chapel's carpet from changing based on the Bible sprite, since the implimentation was bugged
/:cl:

### DIFF
--- a/code/game/jobs/job/support_chaplain.dm
+++ b/code/game/jobs/job/support_chaplain.dm
@@ -5,7 +5,7 @@
 	department_flag = JOBCAT_SUPPORT
 	total_positions = 1
 	spawn_positions = 1
-	is_service = 1
+	is_service = TRUE
 	supervisors = "the head of personnel"
 	department_head = list("Head of Personnel")
 	selection_color = "#dddddd"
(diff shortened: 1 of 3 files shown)


AffectedArc07:
I have a concern with just how large these menus are. Maybe a TGUI with previews of each would work better.

code/game/jobs/job/support_chaplain.dm
```diff
@@ -36,134 +36,51 @@
...
-
+	INVOKE_ASYNC(src, .proc/religion_pick, H)
+
+/datum/outfit/job/chaplain/proc/religion_pick(mob/living/carbon/human/user)
+	var/obj/item/storage/bible/B = new /obj/item/storage/bible(get_turf(user))
+	B.customisable = TRUE // Only the initial bible is customisable
+	user.put_in_l_hand(B)
+
+	var/religion_name = "Christianity"
+	var/new_religion = sanitize(copytext(input(user, "You are the Chaplain. What name do you give your beliefs? Default is Christianity.", "Name change", religion_name), 1, MAX_NAME_LEN))
```
AffectedArc07, line 47: `clean_input` may work better here, this is a lot of nested procs.

code/game/jobs/job/support_chaplain.dm
```diff
@@ -36,134 +36,51 @@
...
-		feedback_set_details("religion_deity", "[new_deity]")
-		feedback_set_details("religion_book", "[new_book_style]")
-
-
+	INVOKE_ASYNC(src, .proc/religion_pick, H)
+
+/datum/outfit/job/chaplain/proc/religion_pick(mob/living/carbon/human/user)
+	var/obj/item/storage/bible/B = new /obj/item/storage/bible(get_turf(user))
+	B.customisable = TRUE // Only the initial bible is customisable
+	user.put_in_l_hand(B)
```
AffectedArc07, line 44: Any reason this was changed from `equip_to_slot_or_del`?

code/game/jobs/job/support_chaplain.dm
```diff
@@ -36,134 +36,51 @@
...
+
+	var/religion_name = "Christianity"
+	var/new_religion = sanitize(copytext(input(user, "You are the Chaplain. What name do you give your beliefs? Default is Christianity.", "Name change", religion_name), 1, MAX_NAME_LEN))
+
+	if(!new_religion)
+		new_religion = religion_name
+
+	switch(lowertext(new_religion))
+		if("christianity")
+			B.name = pick("The Holy Bible")
```
AffectedArc07, line 54: `pick` with only 1 item in it

code/game/jobs/job/support_chaplain.dm
```diff
@@ -36,134 +36,51 @@
...
+		if("toolboxia")
+			B.name = "Toolbox Manifesto Robusto"
+		if("science")
+			B.name = pick("Principle of Relativity", "Quantum Enigma: Physics Encounters Consciousness", "Programming the Universe", "Quantum Physics and Theology", "String Theory for Dummies", "How To: Build Your Own Warp Drive", "The Mysteries of Bluespace", "Playing God: Collector's Edition")
+		else
+			B.name = "The Holy Book of [new_religion]"
+	feedback_set_details("religion_name", "[new_religion]")
+
+	var/deity_name = "Space Jesus"
+	var/new_deity = sanitize(copytext(input(user, "Who or what do you worship? Default is Space Jesus.", "Name change", deity_name), 1, MAX_NAME_LEN))
```
AffectedArc07, line 76: Same here, try `clean_input`

code/game/objects/items/weapons/storage/bible.dm
```diff
@@ -8,9 +8,19 @@
 	resistance_flags = FIRE_PROOF
 	var/mob/affecting = null
 	var/deity_name = "Christ"
+	/// Is the sprite of this bible customisable
+	var/customisable = FALSE
+
+	// ALL OF THESE MUST BE THE SAME LENGTH
+	/// List of bible variant names, used for the radial menu
+	var/static/list/bible_names = list("Bible", "Koran", "Scrapbook", "Creeper", "White Bible", "Holy Light", "PlainRed", "Tome", "The King in Yellow", "Ithaqua", "Scientology", "the bible melts", "Necronomicon", "Greentext")
```
AffectedArc07, line 16: Maybe `ASSERT()` that these 3 are the same length in `Initialize()`

code/game/objects/items/weapons/storage/bible.dm
```diff
@@ -68,51 +78,98 @@
...
+	/*var/carpet_dir
+	switch(choice)
+		if("Bible")
+			carpet_dir = 2
+		if("Koran")
+			carpet_dir = 4
+		if("Scientology")
+			carpet_dir = 8
+	if(carpet_dir)
+		for(var/area/chapel/main/A in world)
```
AffectedArc07, line 157: Good riddance. It was a frickin `in world` search?!

AffectedArc07 (changes requested):

code/game/objects/items/weapons/storage/bible.dm
moxian: Or maybe make it a list of (associative) lists.
```dm
var/static/variants = list(
  list("name" = "Bible", "state" = "bible", "inhand" = "bible"),
  list("name" = "Koran", "state" = "koran", "inhand" = "koran"),
  ...
)
```

code/game/jobs/job/support_chaplain.dm
SabreML: It seemed a bit odd to me that the chaplain's bible could be deleted in some circumstances, so I thought why not just make it drop on the floor instead.

code/game/jobs/job/support_chaplain.dm
SabreML: The `copytext()` is still needed I think, since that puts a maximum length on the name.

SabreML:
> I have a concern with just how large these menus are. Maybe a TGUI with previews of each would work better.

@AffectedArc07
The Null Rods should definitely have a UI at some point with some sort of explanation of their effects, but that probably won't happen with this PR.
While I'm a bit more confident with Javascript now than I was a month ago, I still have no idea how to do sprite previews or anything like that.

As for the size of the menus, the `show_radial_menu()` proc has a `radius` argument which sets their size. I put both of them at `42` because ~~that's what tg has it set to~~ the default value has the options clipping into eachother.
NSS Cyberiad 2020-12-13 122312 vs image

Kyep (approved):
While I do think that the concept of showing how the item looks before you pick has value...
I'm a little concerned that you're encouraging chaplains to choose their rod based solely on how the weapon LOOKS with no understanding of what it does.
I'm also unsure how this will work well for paranormal ERT (who I've been considering reworking to have a far more restricted but more useful set of nullrods).

SabreML:
> While I do think that the concept of showing how the item looks before you pick has value...
> I'm a little concerned that you're encouraging chaplains to choose their rod based solely on how the weapon LOOKS with no understanding of what it does.
> I'm also unsure how this will work well for paranormal ERT (who I've been considering reworking to have a far more restricted but more useful set of nullrods).

Right now the Chaplain can't actually see what the null rod does anyway, so I can only see upsides to letting them know what it'll look like at least.
Paranormal ERT has the same null rod as the Chaplain (as far as I know) so it'd be the same menu.

While I definitely think it would be a lot better with some sort of UI describing the effects of each rod, right now I just don't think I'm good enough to code that.

procdrone:
I feel like tgui with some basic explanations of the items would be pretty cool, for the less experienced chaplains.

variableundefined (approved):
Echoing concerns that TGUI with Sprite Preview would be better - Radial Menus don't really scale well when you have too many items to scroll through.

Still an improvement over what we have though.

PR #15100, opened 2020-12-13
Adjusts who gets to steal the hand teleporter / CMO defib

Qwertytoforty:
## What Does This PR Do

Prevents CE from getting objectives to steal the hand teleporter, prevents paramedic from getting an objective to steal the CMO's defib

## Why It's Good For The Game

We don't give traitors steal objectives for things they often have direct access to. RD is not allowed to steal the hand teleporter, as they have teleporter room access. However, CE has that same access, and can get it just as legally / easily. Thus, CE should not get it as a steal objective, just like the RD.

About 40% of shifts I see the paramedic get the compact defib for free as the CMO often does not use it, which makes it real easy for them to get it legally without officers arresting them for it, or them putting any effort in. Thus, it should fall under the same restrictions as CMO, as they get it just about as often.

## Changelog
:cl:
tweak: CE can no longer get the hand teleporter as a steal objective.
tweak: Paramedic can no longer get the CMO's defib as a steal objective.
/:cl:

### DIFF
--- a/code/game/gamemodes/steal_items.dm
+++ b/code/game/gamemodes/steal_items.dm
@@ -50,7 +50,7 @@
 /datum/theft_objective/hand_tele
 	name = "a hand teleporter"
 	typepath = /obj/item/hand_tele
-	protected_jobs = list("Captain", "Research Director")
+	protected_jobs = list("Captain", "Research Director", "Chief Engineer")
 
 /datum/theft_objective/ai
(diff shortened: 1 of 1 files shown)


Dumbdumn5 (approved):
Given the justification in the description, :+1:

AffectedArc07 (approved):

PR #15103, opened 2020-12-13
Gives red and gamma ERT (just gamma for janitor) death alarms

Qwertytoforty:
## What Does This PR Do

Gives all gamma and red ERT (bar janitor for red) Death alarm implants.
## Why It's Good For The Game

Ert sadly often wordlessly die. Command gets death alarm implants. ERT is kinda like command. Therefore, they should probably get death alarm implants, so ert / crew knows when the ERT is loosing badly, be it verse a blob, or kidnapped by a vampire in medmaints.

## Changelog
:cl:
tweak: All gamma ERT, and all red ERT (bar janitor) now have death alarm implants.
/:cl:

### DIFF
--- a/code/modules/response_team/ert_outfits.dm
+++ b/code/modules/response_team/ert_outfits.dm
@@ -89,6 +89,9 @@
 		/obj/item/storage/lockbox/mindshield = 1
 	)
 
+	implants = list(/obj/item/implant/mindshield,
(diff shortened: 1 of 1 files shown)


datlo:
I approve but if we are going that direction I feel itd be better to just make death alarm standard issue for every single ERT member. Would cause less confusion when an amber ERT is called.

PR #15104, opened 2020-12-13
Syndie Space Base

### LINKED ISSUE #15350: Syndicate Teleporter Console in Syndicate Forward Base shows BSOD with correct access ID or advanced admin interaction
**Issue Description**:

The Syndicate Teleporter Console in the Syndicate Forward Base shows a NTOS BSOD when the user has A) a correct Syndicate ID allowing access to enabling the Teleport Beacon/Portal or B) Advanced Admin Interaction enabled whilst observing. The Syndicate Teleporter Console in the Suspicious Supply Depot, however, works correctly. This has been observed by an admin on live, then tested on private.

Not to encounter BSOD. If the  Syndicate Teleporter Console in the Syndicate Forward Base is not to be interfaced with, it should have a screen for this rather than a BSOD.

**What happened instead**:

Correct access ID or advanced admin interaction causes a consistent BSOD on the Syndicate Teleporter Console in the Syndicate Forward Base.

We shouldn't have BSOD, even if the player won't generally encounter them; it may be inconvenient for administration.

Either:
1. be a humanoid mob with correct syndicate access ID at the Syndicate Teleporter Console in the Syndicate Forward Base, and access the console,
or:
2. observe the Syndicate Teleporter Console in the Syndicate Forward Base with proper administrative rights and enable advanced admin interaction, then access the console.

Noticed 25/Jan/2021

**Extra information**:

Accessing console with advanced admin interaction (console appears like this with proper access also):
VG4Rpuw
Accessing console without proper access:
8oH9Jdx

Kyep:
## What Does This PR Do
Adds the "Syndicate Space Base" space ruin, replacing the old disabled syndicate lavabase ruin.
This is a from-scratch redesign.
Unlike the old syndie lavabase, this new syndie space base:
- Won't crash the server via atmos/lavaland problems, as the old lavabase did
- Won't provide miners extreme powergaming equipment, like head-exploding sniper rifles
- Won't have a comms agent and thus won't enable ghosts to troll the station over comms all shift

Instead, the new base is designed to have a much more positive influence on the server:
- Will have 6 spawners, instead of 4, thus providing more positions ghosts can take
- Will support a much wider variety of jobs (e.g: chef, botanist, TEG engine setup, viably sized grenade testing range, etc)
- Is designed to be a much more isolated area, where newer players can learn complicated jobs (e.g: viro, chemistry, surgery, etc) in peace without worrying about being murdered by miners, swarmers, dragons, ashwalkers, etc
- Has very limited interaction with the main station. Basically, they can watch advanced camera console, and talk on syndie frequency to give traitors some intel, but that's it. Their base is not intended to be entered by outsiders.

Also fixes this runtime:
[2020-12-19T20:36:26] Runtime in depot.dm,444: undefined variable /area/ruin/unpowered/XXXXXXXXXXXXX/arrivals/var/on_peaceful proc name: get menu (/obj/machinery/computer/syndicate_depot/teleporter/get_menu)
And fixes #15350 
These fixes are a side-effect of tweaking the syndi teleporter console to be more broadly compatible with being placed in other locations. It now uses an ACCESS_SYNDICATE_LEADER check to determine access, before it lets you change any setting. It no longer checks for admin status directly, nor does it care as to the depot's visitor status. 

## Why It's Good For The Game
Reasons:
- Serves as a tutorial area for newer players, helping them learn some of the more complicated jobs in peace. This base provides similar value to the old VR and similar PRs, but with far fewer code changes.
- Serves as a fun / mess around area for bored ghosts, reducing salt in dchat and meaning that if you're say murdered by an antag you are more likely to have something to do beyond just observing.
- Provides a SMALL amount of help to syndies who buy the comms key (scientists can use camera console and let you know where your target is, for example). This is a very limited amount of help BUT it does slightly encourage teamwork by traitors.

## Images of changes
StrongDMM_VM2h7RqH8m
NOTE: the edge turfs are /turf/template_noop, IE whatever they spawn on is unchanged, so when the ruin spawns on the live server they will be space. They just appear like this in the map editor so map editors can tell them apart.

## Known Issues
You won't get power/air alarms from the base as-is, BUT if an air alarm is destroyed and a standard alarm built to replace it, the standard alarm (since it doesn't have the syndicate type) will generate an alert that is viewable on station.

## Changelog
:cl: Kyep
add: Added Syndicate Space Base, a large new space ruin that is effectively a tutorial/learning area ghosts can spawn into. Replaces broken old syndie lavabase, and supports a much wider variety of jobs to learn.
tweak: Syndi teleporter consoles no longer runtime in some situations, and now check for access rather than directly checking admin status. People with agent IDs can no longer change some settings on the one in the depot. 
/:cl:


code/game/machinery/autolathe.dm
```diff
@@ -490,3 +491,14 @@
...
 		disabled = FALSE
+
+/obj/machinery/autolathe/syndicate
+	name = "suspicious autolathe"
+	board_type = /obj/item/circuitboard/autolathe/syndi
+
+/obj/machinery/autolathe/syndicate/New()
+	..()
+	if(files)
+		qdel(files)
```
AffectedArc07, line 502:
```suggestion
		QDEL_NULL(files)
```
This may work better here as its probably best to null the entire thing out if youre about to re-assign it

code/modules/research/designs.dm
```diff
@@ -45,3 +45,4 @@ other types of metals and chemistry for reagents).
 	var/list/reagents_list = list()			//List of reagents. Format: "id" = amount.
 	var/maxstack = 1
 	var/lathe_time_factor = 1			//How many times faster than normal is this to build on the protolathe
+	var/syndi_allowed = TRUE			//If false, syndi lathes will not be able to make this (used for radios, etc)
```
AffectedArc07, line 48: Not sure how I feel about designs being snowflaked in this manner. Would be much happier with removing the specific designs from `known_designs` since its only 3. This can easily be accomplished with

```dm
known_designs["intercom_electronics"] = null // No intercom electronics for syndies
```

Better yet, do this in the `/New()` for the `/datum/research/autholathe/syndicate` datum

Darkmight9:
Alright here's some of the issues. Big one is that there is a baseturf tool somewhere on the base, look around for a syndi balloon, it's causing the base turf to be lava rather than space. On that subject you might want to add some breach repair equipment such as the porta fans, those blue mobile air pump things along with a fill up port at the atmos room to supply it with.

You might also want to give extra reinforcement to the bomb testing range, one powerful bomb can cause the chamber and rock to blast out to space

A cryo pod would be nice to have so people can go if they want.

Lastly a nitpick about the medical wing, The surgery area could use a light at the end of the hall at the table 
image

Kyep:
@Darkmight9 
Re: lava/baseturfs, there is no baseturf helper on this map. After investigating I figured out the real cause is mapping mistakes in the other lavaland ruin map files. They're causing properties of lavaland ruins to be incorrectly applied to other ruins, including this one. After discussing with a maint (AA) they're planning to put up their own fix for this. Its unrelated to this PR and touches so many maps it needs to be its own PR.
Re: repair equipment, all syndies spawn with a full toolbelt and there are 3 50 stacks of plastitanium/metal/glass in storage. They also have a couple of space suits. I assume by "blue mobile air pump things" you mean portable air pumps. Thus far they haven't needed anything extra beyond what they already have, but I will keep an eye on it during testing and if they're having issues repairing areas then I may add more repair supplies. Bear in mind they do have a working atmos system so all they really have to do is patch the holes and most places' air should refill over time. I've also added two connectors to the distribution system so you can re-fill tanks with air from there if you need to.
Re: reinforcing the bomb testing range, realistically that room is gonna get breached/depressurized no matter what. That's its entire purpose. To get blown up so the rest of the base doesn't have to be. That's why it has fans in front of it to stop it depressurizing the rest of the base when its exploded.
Re: cryopod, I actually did have this in a prior version but I had to take it out because it was notifying people on station when one of them cryoed. I might add a new variant of cryopod which doesn't alert the station when people cryo in it.
Re: surgery light: added

Wojz:
Illicit intercoms could be nice for those situations when station telecomms get skrekked, if I remember correctly those can't talk on other channels than syndie. Station can deal with telecomms by printing station bounced radios, not the case for the syndies for obvious reasons.

Also I think a morgue could be nice to deal with corpses, though well the base DOES get a gibber :^)

Kyep:
@Wojz Probably not going to add intercomms as I want syndies on the base to have to talk to each other personally in some situations (like weather events which take down comms). Morgue I might add in a subsequent PR if I see the base ends up with a ton of corpses. I don't expect it will, though.

Pyrodes:
Suggestion, what if a new slot opened up for the spawner when someone cryos in the base?

Kyep:
@Pyrodes I thought about that, but decided not to do it, because:
1. If people are cryoing at the base, its likely the stuff to do there is mostly done, so there won't be much demand for new people joining.
2. Allowing people to cryo and then potentially rejoin might open up some nasty exploits, like respawning gear, cryoing and rejoining instantly to heal yourself, etc.
For both of those reasons, I've decided not to do it in this PR. I might re-evaluate in future if it frequently comes up as people want, and I'm confident that I can do it in a way that doesn't introduce any unintended behavior, consequences, or exploits.

code/datums/ruins/space.dm
```diff
@@ -258,6 +258,15 @@
 	always_place = TRUE
 	cost = 0
 
+/datum/map_template/ruin/space/syndicate_space_base
+	name = "Syndicate Space Base"
+	id = "syndie-space-base"
+	description = "A secret base researching illegal bioweapons, it is closely guarded by an elite team of syndicate agents."
+	suffix = "syndie_space_base.dmm"
+	cost = 0
+	always_place = TRUE
```
AffectedArc07, line 267: Not sure if making this `always_place` is the best idea

code/game/machinery/cryopod.dm
```diff
@@ -708,6 +710,12 @@
 /obj/machinery/cryopod/blob_act()
 	return //Sorta gamey, but we don't really want these to be destroyed.
 
+/obj/machinery/cryopod/offstation
+	// Won't announce when used for cryoing.
+	silent = TRUE
+	orient_right = 1
+	icon_state = "body_scanner_0-r"
```
AffectedArc07, line 717: Making these orient right purely for this ruin is meh. Make another instance for `offstation/right` or edit this on the DMM itself.

AffectedArc07 (changes requested):

TDSSS:
If the bomb range is bound to get blown up and breached, maybe it should start off depressurised to avoid creating active turfs and the Atmos load that brings.

They do have a space suit so access shouldn't be an issue.

code/datums/ruins/space.dm
Kyep: There are a couple of reasons for this.
The first is that I want people to play it a lot over the next few weeks so I get another burst of feedback and can consider any tweaks/improvements quickly. Instead of having the feedback strung out over months.
The second is that it provides a bunch of isolated off-station spawns for ghosts, and I feel we often have plenty of ghosts who'd like to make use of this.
If you really want it changed to use the same spawn weight as old lavabase (cost = 20) I'll do it, but personally I'd like to see it get used extensively for at least a few weeks as a form of stress test. I've already uploaded a variant of this map into a bunch of rounds so its already been tested quite a bit, but a longer lasting intense testing period can't hurt.

code/game/machinery/cryopod.dm
Kyep: Changed

code/datums/ruins/space.dm
MrMagolor: couldn't that be a config setting instead?

Kyep:
@AffectedArc07 All your stuff is resolved.

@TDSSS I thought about that, but the bomb range already has air-sealing fans to severely limit how many turfs become atmos-active even if the range is blown up, and I do not want to encourage them to rely on suits. I was actually thinking I might remove the suits totally in a future version of this, depending on how players use them. For the moment though I've left them in but I don't want to make any change which would make their use mandatory.

AffectedArc07 (changes requested):
After looking this over, this needs scrubbers mapping in, with an air injector to vent scrubber gas into space. Theres 4 plasma cannisters in here, and no doubt someone is going to accidentally it.

Kyep:
Added

AffectedArc07 (approved):
If theres quirks now we can easily fix these down the road

PR #15111, opened 2020-12-14
Fixes snowballs damaging structures

### LINKED ISSUE #15110: Snowballs damaging structures
**Problem Description**:
Snowballs, a recent addition, will damage structures like windows and windoors.

The snowball to do no damage.

**What happened instead**:
The snowball did a small amount of damage

It's a free and infinite source of damaging things.

Alt-click snow. Throw snowball at structure.

As soon as I started using snowballs, so right when they were added.

Just goofing around throwing snowballs. I hit the kitchen windoor and it sparks.

dearmochi:
## What Does This PR Do

- Fixes #15110 by moving the stamina damage application to `throw_impact`. Issue root cause was that all thrown objects deal brute damage regardless of `damtype`, which is likely intended

## Why It's Good For The Game

Snowballs can no longer break things

## Changelog
:cl:
fix: Fix snowballs being able to damage and break structures (e.g. windows, mechs)
/:cl:

### DIFF
--- a/code/game/objects/effects/snowcloud.dm
+++ b/code/game/objects/effects/snowcloud.dm
@@ -125,12 +125,15 @@
 /obj/item/snowball
 	name = "snowball"
 	desc = "Get ready for a snowball fight!"
-	throwforce = 10
 	icon_state = "snowball"
-	damtype = STAMINA
+	/// The amount of stamina damage to do on hit.
+	var/stamina_damage = 10
 
 /obj/item/snowball/throw_impact(atom/target)
 	..()
+	if(istype(target, /mob/living))
+		var/mob/living/M = target
+		M.adjustStaminaLoss(stamina_damage)
 	qdel(src)
 
 /obj/item/snowball/fire_act(datum/gas_mixture/air, exposed_temperature, exposed_volume, global_overlay = TRUE)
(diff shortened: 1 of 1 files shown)


code/game/objects/effects/snowcloud.dm
```diff
@@ -125,12 +125,15 @@
...
 	name = "snowball"
 	desc = "Get ready for a snowball fight!"
-	throwforce = 10
 	icon_state = "snowball"
-	damtype = STAMINA
+	/// The amount of stamina damage to do on hit.
+	var/stamina_damage = 10
 
 /obj/item/snowball/throw_impact(atom/target)
 	..()
```
farie82, line 133: Doesn't this also mean that it will hit a person even if he blocks it?

code/game/objects/effects/snowcloud.dm
```diff
@@ -125,12 +125,15 @@
...
+	if(istype(target, /mob/living))
```
SabreML, line 134:
```suggestion
	if(isliving(target))
```

code/game/objects/effects/snowcloud.dm
dearmochi: Correct, fixed

farie82 (approved):

FlattestGuitar:
Wouldn't it make more sense to properly handle throw objects that deal stamina damage? This likely isn't the only case where that's a problem, plus it will be one in the future.

AffectedArc07 (approved):

PR #15113, opened 2020-12-14
Nerfs terrors' ability to pull canisters, tanks and other objects

Kyep:
## What Does This PR Do
When a terror is pulling anything which is not a mob (such as a canister, water tank, or any other object) their movement speed is drastically reduced. A basic gray/green moves at less than half normal speed. They move so slowly that they cannot effectively chase anyone, and they cannot escape if they are chased.
Alternative to https://github.com/ParadiseSS13/Paradise/pull/15073

## Why It's Good For The Game
Terrors are intended to be able to manipulate their environment at a basic level, such as by moving obstacles out of the way without having to smash everything, or conversely moving an object so that it blocks a doorway. They're also intended to be able to utilize basic traps such as by smashing something near them at an opportune time.
What they're not intended to do, however, is drag gas canisters or water tanks halfway across the map and then break them in the most devastating locations. That's a cheesy strategy on behalf of the terrors, and this PR is aimed at reducing their ability to do that.
It also reduces their ability to run up to a group of humans while dragging a tank/canister/whatever and then break it when they're right next to the humans. Under this PR, they can still *attempt* this - but they'll move so slowly while doing it that humans have time to shoot them as they approach, or just move out of the way.

## Changelog
:cl: Kyep
tweak: terror spiders dragging objects (not including mobs) are now drastically slowed.
/:cl:


Pumpkin312:
Unless this comes along with a buff on other aspects this will kill terror spiders survival chances, i know they are not supposed to "win" but if this passes without other tweaks they will not even survive.
They come at midround when science probably already has AEGS and XRAYS and those when mass produced are very hard to counter and one of the only ways to counter those ranged attacks is dragging lockers and tanks. with this, there will be no counter play to ranged attacks from the spider's side.
I know the water tank meta is annoying but its kinda the only way since crew doesn't play fair either.

matttheficus:
Water tank meta is not the most annoying part, the fact you can send one Red Terror to Toxins Storage and literally delete RND/Robotics in one go is silly as hell and shouldnt ever be a thing.

A combined terror ambush that wipes out RND is on thing, a single spider going brrrrt on a plasma canister and then breaking a light is another, especially when they then go and do it to Medbay as well, effectively killing the crew with the loss of maybe 1 or 2 spiders.

Qwertytoforty:
Good change. Still allows terrors to move a welder tank away from their nest, just slower, and still leaves it as a tactic, just not as strong.

Mitchs98:
> Unless this comes along with a buff on other aspects this will kill terror spiders survival chances, i know they are not supposed to "win" but if this passes without other tweaks they will not even survive.
> They come at midround when science probably already has AEGS and XRAYS and those when mass produced are very hard to counter and one of the only ways to counter those ranged attacks is dragging lockers and tanks. with this, there will be no counter play to ranged attacks from the spider's side.
> I know the water tank meta is annoying but its kinda the only way since crew doesn't play fair either.

I really don't think this is going to drastically nerf terrors that much. I've never really seen them _need_ to drag something to survive, AEGs or otherwise. I've seen them use this strategy for cheese 9/10, I've not seen them use it just to escape or survive. Walls exist, moving behind things to block shots exists. This is what I see terrors do more often than not. 

Sure some drag tanks and what have you but the majority don't, and they survive just fine, and even win without doing this.

variableundefined (approved):

SteelSlayer (approved):

Spacemanspark:
I don't feel this will change much, if anything. Farie's change completely addresses the main concern, while this merely pads it down a bit.

It's better than nothing, I suppose.

Kyep:
This was TM'ed today.
No issues were noticed, and it about as well as it possibly could.
Not only were there no instances of "drag X everywhere", but a red dragging a welding tank was moving slow enough that someone shot the tank, blew it up, which did enough thermal damage to kill the nearby queen and thus wipe out the hive.
Maybe this even too much of a nerf. It was fairly brutal to watch. However... it was effective in testing.

SabreML:
During the testmerge I noticed that the slowdown happens no matter what you're dragging, is this intentional?
It seems a bit odd that it would take the same effort to drag a water tank as it would a piece of paper.

Kyep:
@SabreML Yes, that's intentional. Terrors have no reason to drag, say a piece of paper. Aside from mobs, tanks and canisters, the only objects they really have any reason to drag are guns... and even then its generally not a good idea.

AffectedArc07 (approved):

PR #15118, opened 2020-12-15
Revert the workaround for byond version 1537. Bump min version for travis

farie82:
## What Does This PR Do
Removes the workaround for the previous byond version with the `?.` stuff

## Why It's Good For The Game
It's an ugly workaround that was needed but no more

## Changelog
Not needed


AffectedArc07 (approved):
If this breaks I blame you

PR #15119, opened 2020-12-15
Active Turfs Visualiser

AffectedArc07:
## What Does This PR Do
Adds in a debug verb to visualise currently active atmos turfs. When you invoke the verb, it gives you a list of z-levels in chat, as well as their count of active turfs
image

Once you have selected a zlevel, you will get this view.
image
image

This UI does **NOT** live update, and only initially renders. This is to ensure server load is kept to a minimum.

## Why It's Good For The Game
Recently, theres been a major rise in ATs spiking in late round (Into the range of 10k). This verb aims to find those hotspots and figure out *why* that happens. 

## Changelog
:cl: AffectedArc07
add: [Admin] Added a Debug verb to visualise the current list of active atmospheric turfs
/:cl:

### DIFF
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -929,3 +929,55 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
 	message_admins("<span class='adminnotice'>[key_name_admin(src)] [SSmedals.hub_enabled ? "disabled" : "enabled"] the medal hub lockout.</span>")
 	feedback_add_details("admin_verb","TMH") // If...
 	log_admin("[key_name(src)] [SSmedals.hub_enabled ? "disabled" : "enabled"] the medal hub lockout.")
+
+
+/client/proc/visualise_active_turfs()
+	set category = "Debug"
+	set name = "Visualise Active Turfs"
+
+	if(!check_rights(R_DEBUG))
+		return
+
+	// This can potentially iterate through a list thats 20k things long. Give ample warning to the user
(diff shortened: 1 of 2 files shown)


code/modules/admin/verbs/debug.dm
```diff
@@ -929,3 +929,55 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
...
+	if(!z_to_view)
+		return
+
+	// Do not combine these
+	var/list/ui_dat = list()
+	var/list/turf_markers = list()
+
+	var/datum/browser/vis = new(usr, "atvis", "Active Turfs (Z[z_to_view])", 300, 315)
+	ui_dat += "<center><canvas width=\"255px\" height=\"255px\" id=\"atmos\"></canvas></center>"
+	ui_dat += "<script>e=document.getElementById(\"atmos\");c=e.getContext('2d');c.fillStyle='#ffffff';c.fillRect(0,0,255,255);function s(x,y){var p=c.createImageData(1,1);p.data\[0]=255;p.data\[1]=0;p.data\[2]=0;p.data\[3]=255;c.putImageData(p,(x-1),255-Math.abs(y-1));}</script>"
```
AffectedArc07, line 974: This is just a bit of JS to draw things onto a HTML5 canvas. This is the most efficient way I could think of doing this, though if you have any other suggestions, ill take them. Function is called `s` because minification helps to send less data from server to client.

farie82:
If it lags the server out hard then maybe adding some sleeps/check_ticks in the code would help.

AffectedArc07:
> If it lags the server out hard then maybe adding some sleeps/check_ticks in the code would help.

Will do

Kyep (approved):
Approving because I believe this would be helpful in debugging atmos issues.

PR #15120, opened 2020-12-15
Makes syndicate depot structures Initialize()

AffectedArc07:
## What Does This PR Do
Refactors the syndicate depot to use `Initialize()` for its structures and other things, instead of `New()`. This is in its own PR because its quite finnicky and needs testing on live.

## Why It's Good For The Game
The death of `New()` is my new crusade. Things will adapt, ~~or perish~~

## Changelog
N/A, refactor

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/syndicate.dm
+++ b/code/modules/mob/living/simple_animal/hostile/syndicate.dm
@@ -102,11 +102,10 @@
 	var/shield_key = FALSE
 	var/turf/spawn_turf
 
-/mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/New()
-	..()
+/mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/Initialize(mapload)
+	. = ..()
 	name = "[name] [pick(GLOB.last_names)]"
-	// Do not attempt to move this code to Initialize() or LateInitialize(). Doing so with other objects has caused bugs in the past, because assigning "depotarea" may not work there.
-	depotarea = areaMaster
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/simple_animal/hostile/syndicate.dm
```diff
@@ -102,11 +102,10 @@
 	var/shield_key = FALSE
 	var/turf/spawn_turf
 
-/mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/New()
-	..()
+/mob/living/simple_animal/hostile/syndicate/melee/autogib/depot/Initialize(mapload)
+	. = ..()
 	name = "[name] [pick(GLOB.last_names)]"
-	// Do not attempt to move this code to Initialize() or LateInitialize(). Doing so with other objects has caused bugs in the past, because assigning "depotarea" may not work there.
```
AffectedArc07, line 108: Reason that this works now. What *likely* happened before is that `depotarea` was being set twice, and never overwritten properly. `areaMaster` is also set in `/atom/movable/New()`. 

Another issue before was that objects spawned from spawners which `Initialize()` sometimes, and at one point child objects spawned from said spawners would *not* `Initialize()`, however, they do now. If you create a spawner with an `Initialize()` call to make 4 objects, those 4 objects will also `Initialize()` themselves now.

Kyep: Yeah, the latter issue is what I ran into when I tried to make this Initialize() before.
Good job getting it working.

PR #15129, opened 2020-12-17
Fixes ticker roundstart being slow

AffectedArc07:
## What Does This PR Do
This *hopefully* fixes the issue where the ticker takes 10 years to start the round. While Async SQL is very good for queries to run in the background instead of blocking DD, the tradeoff is it taking a while for mass queries which return data since it sleeps for atleast 1 tick each query. Now throw in checking every player for job bans and species bans at roundstart and you see where the delay comes from. These queries have been made blocking now. While this does mean DD will freeze at roundstart (like it always has), its only for about 5 seconds, instead of 30 seconds of limbo where you can still chat, but nothing happens.

At some point I may refactor these so it loads into a GLOB list to avoid the massive DB lookups, but this will do for now

## Why It's Good For The Game
Roundstart shouldnt take an eternity.

## Changelog
:cl: AffectedArc07
fix: Roundstart setup no longer takes an eternity
/:cl:

### DIFF
--- a/code/game/jobs/whitelist.dm
+++ b/code/game/jobs/whitelist.dm
@@ -35,7 +35,7 @@ GLOBAL_LIST_EMPTY(whitelist)
 				list("ckey" = M.ckey)
 			)
 
-			if(!job_read.warn_execute())
+			if(!job_read.warn_execute(async = FALSE)) // Dont async this. Youll make roundstart slow.
 				qdel(job_read)
 				return FALSE
 
@@ -87,7 +87,7 @@ GLOBAL_LIST_EMPTY(alien_whitelist)
 			list("ckey" = M.ckey)
 		)
 
-		if(!species_read.warn_execute())
+		if(!species_read.warn_execute(async = FALSE)) // Dont async this one. It makes roundstart take 10 years.
 			qdel(species_read)
 			return FALSE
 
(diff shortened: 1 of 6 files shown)


code/modules/client/client_procs.dm
```diff
@@ -572,7 +572,7 @@
 	// Log player connections to DB
 	var/datum/db_query/query_accesslog = SSdbcore.NewQuery("INSERT INTO `[format_table_name("connection_log")]`(`datetime`,`ckey`,`ip`,`computerid`) VALUES(Now(), :ckey, :ip, :cid)", list(
 		"ckey" = ckey,
-		"ip" = address,
+		"ip" = "[address ? address : ""]", // This is important. NULL is not the same as "", and if you directly open the `.dmb` file, you get a NULL IP.
```
AffectedArc07, line 575: This is another fix I threw in while fixing DB things

code/game/jobs/whitelist.dm
```diff
@@ -35,7 +35,7 @@ GLOBAL_LIST_EMPTY(whitelist)
 				list("ckey" = M.ckey)
 			)
 
-			if(!job_read.warn_execute())
+			if(!job_read.warn_execute(async = FALSE)) // Dont async this. Youll make roundstart slow.
```
AffectedArc07, line 38: At some point ill likely refactor these to use a GLOB and have them all cache in at world init with a single query that can run async, instead of 300 queries running async.

AffectedArc07:
Tested on live.
image
**Much** better

PR #15130, opened 2020-12-17
More stealthy fake walls

TDSSS:
## What Does This PR Do
This PR makes it so you can't immediately tell a false wall from a real one by examining it and noticing it doesn't tell you it's intact. You need to interact with the wall or use other,harder to spot clues now.

A comment on the examine_status proc suggested it was created for false walls, but it was not actually used for that for some reason.

## Why It's Good For The Game
Makes on-station antag bases slightly harder to spot. Preserves IMMERSION by requiring a closer examination of a false wall to notice it's false.

## Changelog
:cl:
tweak: false wall examine descriptions should match real walls now
/:cl:

### DIFF
--- a/code/game/objects/structures/false_walls.dm
+++ b/code/game/objects/structures/false_walls.dm
@@ -37,6 +37,18 @@
 	..()
 	air_update_turf(1)
 
+/obj/structure/falsewall/examine_status(mob/user)
+	var/healthpercent = (obj_integrity/max_integrity) * 100
+	switch(healthpercent)
+		if(100)
+			return "<span class='notice'>It looks fully intact.</span>"
+		if(70 to 99)
+			return  "<span class='warning'>It looks slightly damaged.</span>"
+		if(40 to 70)
+			return  "<span class='warning'>It looks moderately damaged.</span>"
+		if(0 to 40)
+			return "<span class='danger'>It looks heavily damaged.</span>"
+
 /obj/structure/falsewall/ratvar_act()
 	new /obj/structure/falsewall/brass(loc)
 	qdel(src)
@@ -158,6 +170,10 @@
 	icon_state = "r_wall"
 	walltype = /turf/simulated/wall/r_wall
 	mineral = /obj/item/stack/sheet/plasteel
+
+/obj/structure/falsewall/reinforced/examine_status(mob/user)
+	. = ..()
+	. += "<br><span class='notice'>The outer <b>grille</b> is fully intact.</span>"	//not going to fake other states of disassembly
 
 /obj/structure/falsewall/reinforced/ChangeToWall(delete = 1)
 	var/turf/T = get_turf(src)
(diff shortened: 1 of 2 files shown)


FreeStylaLT:
(As a note for our currently-frozen mappers, the `/obj/effect/spawner/random_barrier/wall_probably` and `/obj/effect/spawner/random_barrier/obstruction` spawners have chances to spawn falsewalls.)

Spacemanspark:
Good change, although I don't think it will change much unfortunately.
Most players already just spam click on walls if they even slightly suspect a fake one is nearby.

DoctorDrugs:
The ones that didn't know even players with 1000+ hours already spam click walls around common areas to locate false walls, i don't think this will honestly change it much outside of just having the group that knew about examining walls to switch to spam clicking every wall to locate possible false walls

procdrone:
making fake walls not pop up on mesons would be a better way to go, personally. For me its the most dead giveaway of fake walls.

Leanfrog:
I have another suggestion, instead of just changing the examine, why not make the screwdriver not turn it into a regular wall but instead sort of lock it, so you can't open it by just pushing it, you can already fix a false wall by dismantling it and wrenching the frame, this would have much more impact than an examine text that nobody notices anyways

SteelSlayer (approved):

Carthusia:
Perhaps another possible change would be adding a few second timer to pushing the wall.

So you cant rapidly spam 10 sections of wall to test for fakes, you'd need to spend say, 5 seconds checking the wall segment.

"You begin checking the wall for irregularities"
(You find nothing)
(You find a small hatch and pull it open!)

While closing a false wall would t ake less time as you just are pulling the hatch closed.

TDSSS:
Going to be honest, this was primarily intended as a fix-style PR since I figured the description of false walls and real walls not matching up was unintended.

But since it's gotten the Balance tag already, I might as well look into ways to make fake walls harder to notice in general.

Kiyahitayika:
Would you be able to change the title of this PR? It really faked me out, though I think clarity could help. I was expecting something about the `*collapse` emote.

Kiyahitayika:
> Perhaps another possible change would be adding a few second timer to pushing the wall ontop of making it so mesons can no longer spot them.
> 
> This way you cant rapidly spam an entire section of the stations walls to test for fakes by just running past them, you'd need to spend say, 5 seconds checking the wall segment.
> 
> "You begin checking the wall for irregularities"
> (You find nothing)
> (You find a small hatch and pull it open!)
> 
> While closing a false wall would t ake less time as you just are pulling the hatch closed.

On the other hand, would this also reduce their usability for antags?

Carthusia:
Perhaps its coded so that if you make the wall, you have no delay, as you already know where the hatch is

Kyep (approved):

AffectedArc07 (approved):

PR #10205, opened 2018-11-12
JSON Feedback - Statbus Prep (66% done) (fuck it I'm not getting to 100% edition)

tigercat2000:
Changes list:
 - Blackbox is dead, long live SSblackbox
 - dbcon is dead, long live SSdbcore
 - The feedback database now uses JSON instead of stupid as fuck feedback data that's borderline unusable
 - Some feedback data is now tracked in a different table
 - round_id is tracked globally. Whatever the flying fuck you do, don't include this shit in the log file tree structure - it'll fuck up our backend tooling.
 - We now **require** MariaDB, not any other SQL server. Classic MySQL's support for the `JSON` field type is so god awful that it shits itself if you even think about using it. On the bright side, most linux distros have already figured this shit out, and mariaDB is the default resolution package for `mysql`.
 - We've gone up 2 SQL versions in one PR, and they gotta be applied successively or shit breaks bad.

This fucking mess is overhauling our feedback database to make it all run on JSON so we can do stuff like https://atlantaned.space/statbus/ in the future. 

No CL because this is purely tooling stuff for our feedback db.

PR #15132, opened 2020-12-17
Annihilates the blackbox

AffectedArc07:
## What Does This PR Do
Finished what was started in #10205. Most of this work is @tigercat2000's, I have just bought it up to date. This JSON-ifies all of the data in the `feedback` table, making it much nicer for external sites (**maybe**, dont bug us for this). It also integrates the `round_id` system, meaning rounds can actually be differentiated by a numerical ID, making it much easier for log diving and bug reports. 
image
image

## Why It's Good For The Game
Many reasons

JSON feedback is **MUCH** smaller, since it can compact `FRDBC FRDBC FRDBC FRDBC FRDBC FRDBC FRDBC FRDBC` can be condensed to `{"FRDBC":8}`.
image

It also removes the functionality from the on-station blackbox object. Believe it or not, that **on station object** was responsible for backend logging of server data. I hate it with a burning passion.

And yes, we have gone up **4** SQL versions in 1 PR. And yes, this is necessary.

## Changelog
:cl: AffectedArc07
add: Massively overhauled the database feedback statistic storage system. If youre a regular player, ignore this PR. If you want to know more, open the PR number in this changelog entry.
/:cl:


SQL/paradise_schema.sql
```diff
@@ -558,3 +559,24 @@ CREATE TABLE `ip2group` (
...
+  `server_port` SMALLINT(5) UNSIGNED NOT NULL,
+  `commit_hash` CHAR(40) NULL,
+  `game_mode` VARCHAR(32) NULL,
+  `game_mode_result` VARCHAR(64) NULL,
+  `end_state` VARCHAR(64) NULL,
+  `shuttle_name` VARCHAR(64) NULL,
+  `map_name` VARCHAR(32) NULL,
+  `station_name` VARCHAR(80) NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
AffectedArc07, line 582: These arent UTF8. Oops.

SQL/updates/17-18.py
```diff
@@ -0,0 +1,535 @@
+#Python 3+ Script for jsonifying feedback table data as of 2017-11-12 made by Jordie0608
+#Apologies for the boilerplated and squirrely code in parts, this has been my first foray into python
+#
+#Before starting ensure you have installed the mysqlclient package https://github.com/PyMySQL/mysqlclient-python
+#It can be downloaded from command line with pip:
+#pip install mysqlclient
```
AffectedArc07, line 6: Wrong dep

SQL/updates/18-19.sql
```diff
@@ -0,0 +1,8 @@
+# Updating DB from 15-16, -AffectedArc07
```
SteelSlayer, line 1: This should be `18-19` right?

AffectedArc07: Changed

code/game/world.dm
```diff
@@ -240,7 +245,7 @@ GLOBAL_LIST_EMPTY(world_topic_handlers)
 	return s
 
 /world/proc/SetupLogs()
-	GLOB.log_directory = "data/logs/[time2text(world.realtime, "YYYY/MM-Month/DD-Day")]"
+	GLOB.log_directory = "data/logs/[time2text(world.realtime, "YYYY/MM-Month/DD-Day")]/round-[GLOB.round_id]"
```
moxian, line 248: This feels like it would break when round_id is not set (because we have db-less setup for example.
I don't think it's worth fixing it here, instead make GLOB.round_id default to something sane (so it can be used in other places and not deal with `null` handling).

code/modules/admin/topic.dm
```diff
@@ -426,7 +426,6 @@
 		to_chat(GLOB.banlist_savefile["minutes"], minutes)
 		to_chat(GLOB.banlist_savefile["bannedby"], usr.ckey)
 		GLOB.banlist_savefile.cd = "/base"
-		feedback_inc("ban_edit",1)
```
moxian, line 429: I see all the ban-related counters got axed? Why? Not saying it's a bad thing, just feels odd.

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -106,8 +106,7 @@
 		dna.species.handle_death(gibbed, src)
 
 	if(SSticker && SSticker.mode)
-//		log_world("k")
```
moxian, line 109: k

code/modules/recycling/disposal.dm
```diff
@@ -368,7 +368,6 @@
 		if( contents.len )
 			if(mode == 2)
 				spawn(0)
-					feedback_inc("disposal_auto_flush",1)
```
moxian, line 371: ?

code/controllers/subsystem/blackbox.dm
```diff
@@ -0,0 +1,274 @@
...
+		append the key and it's value if it doesn't exist
+	calls: 	SSblackbox.record_feedback("nested tally", "example", 1, list("fruit", "orange", "apricot"))
+			SSblackbox.record_feedback("nested tally", "example", 2, list("fruit", "orange", "orange"))
+			SSblackbox.record_feedback("nested tally", "example", 3, list("fruit", "orange", "apricot"))
+			SSblackbox.record_feedback("nested tally", "example", 10, list("fruit", "red", "apple"))
+			SSblackbox.record_feedback("nested tally", "example", 1, list("vegetable", "orange", "carrot"))
+	json: {"data":{"fruit":{"orange":{"apricot":4,"orange":2},"red":{"apple":10}},"vegetable":{"orange":{"carrot":1}}}}
+	tracking values associated with a number can't merge with a nesting value, trying to do so will append the list
+	call:	SSblackbox.record_feedback("nested tally", "example", 3, list("fruit", "orange"))
+	json: {"data":{"fruit":{"orange":{"apricot":4,"orange":2},"red":{"apple":10},"orange":3},"vegetable":{"orange":{"carrot":1}}}}
```
moxian, line 156: This is very bad, because *depending on the tool you use to parse the json* you'll be getting different results.
Can we have it runtime instead/in addition to this?

code/controllers/subsystem/blackbox.dm
```diff
@@ -0,0 +1,274 @@
...
+"associative"
+	used to record text that's associated with a value i.e. coordinates
+	further calls to the same key will append a new list to existing data
+	calls:	SSblackbox.record_feedback("associative", "example", 1, list("text" = "example", "path" = /obj/item, "number" = 4))
+			SSblackbox.record_feedback("associative", "example", 1, list("number" = 7, "text" = "example", "other text" = "sample"))
+	json: {"data":{"1":{"text":"example","path":"/obj/item","number":"4"},"2":{"number":"7","text":"example","other text":"sample"}}}
```
moxian, line 162: This is 
```json
 {"data":{"1":{"text":"example","path":"/obj/item","number":"4"},"2":{"number":"7","text":"example","other text":"sample"}}}
 ```
 and not 
 ```json
  {"data": [{"text":"example","path":"/obj/item","number":"4"}, {"number":"7","text":"example","other text":"sample"}]}
  ```
  only for technical reasons, and not because it's a better way, right?

Also, likely related, what does `1` in the call stand for?

File: code/controllers/subsystem/blackbox.dm [hunk 10]
```diff
@@ -0,0 +1,274 @@
...
+
+Versioning
+	If the format of a feedback variable is ever changed, i.e. how many levels of nesting are used or a new type of data is added to it, add it to the versions list
+	When feedback is being saved if a key is in the versions list the value specified there will be used, otherwise all keys are assumed to be version = 1
+	versions is an associative list, remember to use the same string in it as defined on a feedback variable, example:
+	list/versions = list("round_end_stats" = 4,
+						"admin_toggle" = 2,
+						"gun_fired" = 2)
```
moxian, line 170: This versioning thing is very very confusing... I'll try to think more about this whole thing next time...

code/modules/security_levels/keycard_authentication.dm
```diff
@@ -197,13 +192,15 @@
 GLOBAL_VAR_INIT(maint_all_access, 0)
 GLOBAL_VAR_INIT(station_all_access, 0)
 
+// Why are these global procs?
 /proc/make_maint_all_access()
-	for(var/area/maintenance/A in world)
+	for(var/area/maintenance/A in world) // Why are thse global lists? AAAAAAAAAAAAAA
```
moxian, line 197:
```suggestion
	for(var/area/maintenance/A in world) // Why are these global lists? AAAAAAAAAAAAAA
```

code/modules/shuttle/shuttle_manipulator.dm
```diff
@@ -155,7 +155,7 @@
 					M.setTimer(50)
 					message_admins("[key_name_admin(usr)] fast travelled [M]")
 					log_admin("[key_name(usr)] fast travelled [M]")
-					feedback_add_details("shuttle_fasttravel", M.name)
+					SSblackbox.record_feedback("text", "shuttle_fasttravel", 1, "[M.name]")
```
moxian, line 158: Wait, this is confusing. Sometimes (like here) you add extra quoting to have `"[M.name]"`, yet other times (e.g. in `code/modules/reagents/chemistry/holder.dm`) you REMOVE the quoting to have just
`SSblackbox.record_feedback("tally", "chemical_reaction", C.result_amount * multiplier, C.result)` (note the bare `C.result`).

Which one is it?

tools/ci/generate_sql_scripts.py
```diff
@@ -15,8 +16,12 @@
 orderedSqlFiles = sorted(orderedSqlFiles)
 
 for index in orderedSqlFiles:
-    # Yes I know half these casts are probably not necassary, but python is very picky
-    orderedSqlFiles[index] = str(index) + "-" + (str(int(index)+1)) + ".sql"
+    # Yes I know half of the casts below this are probably not necassary, but python is very picky
+    # AND YES I KNOW THIS IS SNOWFLAKEY AS HELL, BUT IT MUST BE DONE FOR PROPER CI
+    if index in [16, 17]:
+        orderedSqlFiles[index] = str(index) + "-" + (str(int(index)+1)) + ".py"
```
moxian, line 22: Easy dumb way to write this better:
```suggestion
        orderedSqlFiles[index] = "%d-%d.py" % (index, index+1)
```
More fancy way:
```suggestion
        orderedSqlFiles[index] = "{}-{}.py".format(index, index+1)
```
Fanciest way (if you have python 3.6+, which is likely, but idk what your CI uses):
```suggestion
        orderedSqlFiles[index] = f"{index}-{index+1}.py"
```

Same below.

tools/ci/generate_sql_scripts.py
```diff
@@ -39,24 +44,36 @@
...
 for file in orderedSqlFiles:
-    inFile = open("SQL/updates/" + file, "r")
-    fileLines = inFile.readlines()
-    inFile.close()
-    # Add in a line which tells it to use the feedback DB
-    fileLines.insert(0, "USE `feedback`;\n")
+    if file.endswith(".py"):
+        # Begin snowflakery
+        if file == "16-17.py":
+            scriptLines.append("python3 SQL/updates/" + str(file) + " 127.0.0.1 root root feedback feedback round\n")
```
moxian, line 57:
```suggestion
            scriptLines.append("python3 SQL/updates/%s 127.0.0.1 root root feedback feedback round\n" % file)
```
or if the %s is too smol and sneeki for you:
```suggestion
            scriptLines.append("python3 SQL/updates/{file} 127.0.0.1 root root feedback feedback round\n".format(file=file))
```
Or, again, with the fancy f-syntax:
```suggestion
            scriptLines.append("python3 SQL/updates/{file} 127.0.0.1 root root feedback feedback round\n")
```

tools/ci/generate_sql_scripts.py
```diff
@@ -39,24 +44,36 @@
...
+        elif file == "17-18.py":
+            scriptLines.append("python3 SQL/updates/" + str(file) + " 127.0.0.1 root root feedback feedback feedback_2\n")
+        else:
+            print("ERROR: CI failed due to invalid python file in SQL/updates")
+            exit(1)
+    else:
+        inFile = open("SQL/updates/" + file, "r")
```
moxian, line 64: honestly at this point `SQL/updates/file` should be its own variable given that you've used it 3 times.
And btw the canonical way of making the full path would be `os.path.join("SQL", "updates", file)` (or if you're lazy, `os.path.join("SQL/updates", file)` is a middle-ground that happens to work (same as `+`), but also feels good because it's the "proper" way).

tools/ci/generate_sql_scripts.py
```diff
@@ -39,24 +44,36 @@
...
-    outFile.close()
+        # Write new files to be used by the testing script
+        outFile = open("tools/ci/sql_tmp/" + file, "w+")
+        outFile.writelines(fileLines)
+        outFile.close()
 
-    # Add a line to the script being made that tells it to use this SQL file
-    scriptLines.append("mysql -u root -proot < tools/ci/sql_tmp/" + str(file) + "\n")
+        # Add a line to the script being made that tells it to use this SQL file
+        scriptLines.append("mysql -u root -proot < tools/ci/sql_tmp/" + str(file) + "\n")
```
moxian, line 76: something something string concatenation

code/controllers/subsystem/blackbox.dm
```diff
@@ -0,0 +1,274 @@
...
+*/
```
moxian, line 171: Please quickly mention that the data is actually saved at the end of the round, not when the function is called.

SQL/paradise_schema.sql
```diff
@@ -558,3 +559,24 @@ CREATE TABLE `ip2group` (
...
+--
+DROP TABLE IF EXISTS `round`;
+CREATE TABLE `round` (
+  `id` INT(11) NOT NULL AUTO_INCREMENT,
+  `initialize_datetime` DATETIME NOT NULL,
+  `start_datetime` DATETIME NULL,
+  `shutdown_datetime` DATETIME NULL,
+  `end_datetime` DATETIME NULL,
+  `server_ip` INT(10) UNSIGNED NOT NULL,
+  `server_port` SMALLINT(5) UNSIGNED NOT NULL,
```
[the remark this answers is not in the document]
AffectedArc07: Yes, incase we ever branch to separate server instances.

SQL/paradise_schema.sql
```diff
@@ -558,3 +559,24 @@ CREATE TABLE `ip2group` (
   KEY `groupstr` (`groupstr`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
+--
+-- Table structure for table `round`
+--
+DROP TABLE IF EXISTS `round`;
+CREATE TABLE `round` (
+  `id` INT(11) NOT NULL AUTO_INCREMENT,
```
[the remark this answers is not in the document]
AffectedArc07: Because the other port did it this way

code/game/mecha/mecha_construction_paths.dm
```diff
@@ -22,6 +22,9 @@
 	else
 		return ..()
 
+/datum/construction/mecha/spawn_result(name)
+	SSblackbox.record_feedback("tally", "mechas_created", 1, "[name]")
```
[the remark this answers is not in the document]
AffectedArc07: It seems fine to me

code/game/world.dm
```diff
@@ -4,14 +4,19 @@ GLOBAL_LIST_INIT(map_transition_config, MAP_TRANSITION_CONFIG)
...
-	// Setup all log paths and stamp them with startups
-	SetupLogs()
 	enable_debugger() // Enable the extools debugger
+
+	// Right off the bat, load up the DB
+	SSdbcore.CheckSchemaVersion() // This doesnt just check the schema version, it also connects to the db! This needs to happen super early! I cannot stress this enough!
+	SSdbcore.SetRoundID() // Set the round ID here
+
+	// Setup all log paths and stamp them with startups, including round IDs
+	SetupLogs()
```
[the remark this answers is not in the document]
AffectedArc07: Yes, it logs to the DD log window itself
```dm
/datum/controller/subsystem/dbcore/proc/CheckSchemaVersion()
	if(config.sql_enabled)
		// The unit tests have their own version of this check, which wont hold the server up infinitely, so this is disabled if we are running unit tests
		#ifndef UNIT_TESTS
		if(config.sql_enabled && sql_version != SQL_VERSION)
			config.sql_enabled = FALSE
			schema_valid = FALSE
			SSticker.ticker_going = FALSE
			SEND_TEXT(world.log, "Database connection failed: Invalid SQL Versions")
			return FALSE
		#endif
		if(Connect())
			SEND_TEXT(world.log, "Database connection established")
		else
			SEND_TEXT(world.log, "Your server failed to establish a connection with the database. Please check SQL error logs.")
	else
		SEND_TEXT(world.log, "Database is not enabled in configuration")
```

code/game/world.dm
```diff
@@ -91,7 +96,7 @@ GLOBAL_LIST_EMPTY(world_topic_handlers)
 	wth = new wth()
 	return wth.invoke(input)
 
-/world/Reboot(var/reason, var/feedback_c, var/feedback_r, var/time)
+/world/Reboot(var/reason, end_string, var/time)
```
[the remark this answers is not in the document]
AffectedArc07: End string is just the string passed to the chat, and the ticker values. I really dont see a need to document this.

code/game/world.dm
AffectedArc07: image

code/modules/admin/topic.dm
AffectedArc07: We can query the ban DB for more accurate statistics

code/modules/recycling/disposal.dm
AffectedArc07: Why the hell were we logging these is my first question

code/controllers/subsystem/blackbox.dm
AffectedArc07: If you read the comment, its for associative data. `1` is a key here, not an index.

code/controllers/subsystem/blackbox.dm
AffectedArc07: Could you elaborate?

code/modules/shuttle/shuttle_manipulator.dm
AffectedArc07: Sometimes these things can be null, and `json_encode` will treat nulls as actual nulls with `C.result`, however, with `"[M.name]"`, if `M.name` is null, the response will still be `""`, which is not a null entry in `json_encode`

tools/ci/generate_sql_scripts.py
AffectedArc07: The script works fine as it is, and its with code I actually fully understand like the back of my hand (Even though I wrote `generate_sql_scripts.py`, I am not a python wizard). If other maints ask I will change it to be this way, but as of now I would rather keep it in a way I can understand.

tools/ci/generate_sql_scripts.py
AffectedArc07: Ditto

tools/ci/generate_sql_scripts.py
AffectedArc07: Ditto

tools/ci/generate_sql_scripts.py
AffectedArc07: Ditto

AffectedArc07:
Todo: Add round ID to a few more places

code/controllers/subsystem/blackbox.dm
moxian: Some tools/languages/libraries would error out trying to parse `{"data":{"fruit":{"orange":{"apricot":4,"orange":2},"red":{"apple":10},"orange":3}`;
some would treat it  as `{"data":{"fruit":{"orange":{"apricot":4,"orange":2},"red":{"apple":10}}`;
some as `{"data":{"fruit":{"red":{"apple":10},"orange":3}`
some will treat it as either of the above randomly in each individual run.
This nondeterminism is bad and makes for hard to catch bugs..

code/controllers/subsystem/blackbox.dm
moxian: Then why does `SSblackbox.record_feedback("associative", "example", 1, list("number" = 7, "text" = "example", "other text" = "sample"))` add `{"number":"7","text":"example","other text":"sample"}` under the key of `2`, not `1`?

AffectedArc07: Because its assosiactive with `"1"` and `"2"`, not index 1 and 2

moxian: But the `"2"` (nor `2`) is not specified anywhere in the in the
```dm
SSblackbox.record_feedback("associative", "example", 1, list("number" = 7, "text" = "example", "other text" = "sample"))
```
call!

moxian:
I wrote up a comment on this last time, but I guess github did a dumb...

1) Why is `key_type` argument to `record_feedback` a string literal and not `#define`d anywhere? (so relies on people copy-pasting the line and only catching typoes at runtime rather the compiler yelling at you)
2) Why is it stringly typed in the first place, if it has such different semantics depending on the argument? It would be so much nicer to have `record_feedback_text`, `record_feedback_amount` as dedicated procs!

code/controllers/subsystem/blackbox.dm
AffectedArc07: Because it's associative to the first and second calls? There's 2 procs there.

AffectedArc07:
@moxian They are this was because then they match up perfectly with the DB column enumerators. Arguably they should be defined, and I may end up changing that

As for why it's one proc, easier to find references and keep all the variables in one place. It's all /datum/feedback_variable so why have 5 procs to do the same thing but with 1 var different.

moxian:
I don't think that they match to internal DB layout should be this visible to the callers, because they most likely don't care, and that's an implementation detail. The callers don't even know `/datum/feedback_variable` exists.
You can still have `record_feedback_text(..)` simply invoke the existing `record_feedback("text", ..)`; I just strongly believe the call sites in user code should use the former, not the latter.

code/controllers/subsystem/blackbox.dm
tigercat2000: Ooh, I see the problem. The concept of nested tallies itself isn't a problem, it's the fact that this leads to duplicated keys. Note the warning given.

tigercat2000: Wait, is this even possible in BYOND? Associative lists have unique keys, how would this kind of JSON be represented internally?

AffectedArc07:
Updates

Round ID is now shown in revision info
image

Server now handles no round ID better
Null in statpanel
image
Null in revision info
image
Logs now just store to the date directory instead of a `round-` folder
image

PacifistDalek:
> Believe it or not, that **on station object** was responsible for backend logging of server data.

Spacemanspark:
Yeah... uh, so what happened if that object was destroyed

Better yet, how the fuck

AffectedArc07:
> Yeah... uh, so what happened if that object was destroyed
> 
> Better yet, how the fuck

Currently it cant be destroyed due to awful hacky code (Destroying it spawns a new blackbox on Z2). In terms of how the fuck this happened, the person who made this system clearly doesnt know what a datum is.

Kyep (approved):

Kyep:
This makes me a bit nervous. Then again, any large-scale DB change does.
The blackbox is an abomination and our feedback table is out of control.
It needs a rewrite like this, and he who puts in the work (AA) gets to choose how the work is done.
So, approved.

AffectedArc07:
Need to test stuff related to objective saving

variableundefined (approved):

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+}
+```
+
+Notice how everything is casted to strings, and each new entry added has its index increased ("1", "2", etc).
+
+### Nested Tally
+
+`nested tally` is used to track the number of occurances of structured semi-relational values (EG: the results of arcade machines). Similar to running total, but related values are nested in a multi-dimensional array. The final element in the data list is used as the tracking key, all prior elements are used for nesting. Further calls to the same key will add or subtract from the saved value of the data key if it already exists in the same multi-dimensional position,  append the key and it's value if it doesn't exist already. This one is quite complicated, but an example is below:
```
moxian, line 79:
```suggestion
`nested tally` is used to track the number of occurances of structured semi-relational values (e.g.: the results of arcade machines). Similar to running total, but related values are nested in a multi-dimensional array. The final element in the data list is used as the tracking key, all prior elements are used for nesting. Further calls to the same key will add or subtract from the saved value of the data key if it already exists in the same multi-dimensional position,  append the key and it's value if it doesn't exist already. This one is quite complicated, but an example is below:
```

moxian, line 79: I find it nicer to think of as just "tally, but the key is a list". Consider:

`nested tally` is used to track the number of occurances of structured semi-relational values (e.g.: the results of arcade machines). You can think of it as a running total, with the key being a list of strings (rather than a single string), with elements incrementally identifying the entity in question.
Technically, the values are nested in a multi-dimensional array. The final element in the data list is used as the tracking key, all prior elements are used for nesting. Further calls to the same key will add or subtract from the saved value of the data key if it already exists in the same multi-dimensional position,  append the key and it's value if it doesn't exist already. This one is quite complicated, but an example is below:

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+}
+```
+
+Avoid doing this, since having duplicate keys in JSON (data.fruit.orange) will break when parsing.
+
+### Tally
+
+`tally` is used to track the number of occurances of multiple related values (EG: how many times each type of gun is fired). Further calls to the same key will add or subtract from the saved value of the data key if it already exists, and append the key and it's value if it doesn't exist.
```
moxian, line 148:
```suggestion
`tally` is used to track the number of occurances of multiple related values (e.g.: how many times each type of gun is fired). Further calls to the same key will add or subtract from the saved value of the data key if it already exists, and append the key and it's value if it doesn't exist.
```

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+```
+
+### Text
+
+`text` is used for simple single-string records (EG: the current map). Further calls to the same key will append saved data unless the overwrite argument is true or it already exists. When encoded, calls made with overwrite will lack square brackets
+
+```dm
+SSblackbox.record_feedback("text", "example", 1, "sample text")
```
moxian, line 172: Again, the `1` is unused, and, again, this is a reason to have it be its own `SSblackbox.record_feedback_text`

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+SSblackbox.record_feedback("text", "example", 1, "other text")
```
moxian, line 173: Would be good to demonstrate  that the result is a set (with no duplicates), rather than a list/multiset.
```suggestion
SSblackbox.record_feedback("text", "example", 1, "sample text")
SSblackbox.record_feedback("text", "example", 1, "other text")
SSblackbox.record_feedback("text", "example", 1, "sample text")
```

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+    "data":[
+        "sample text",
+        "other text"
+    ]
+}
+```
+
+## Feedback Versioning
+
+If the logging format (IE: What data is logged) for a variable is ever changed, the version needs bumping. This can be done with the `versions` list on the subsystem definition itself. All values default to `1`.
```
moxian, line 189:
```suggestion
If the logging format (i.e.: What data is logged) for a variable is ever changed, the version needs bumping. This can be done with the `versions` list on the subsystem definition itself. All values default to `1`.
```

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+
+```dm
+var/list/versions = list(
+    "round_end_stats" = 4,
+    "admin_toggle" = 2,
+    "gun_fired" = 2)
+```
+
+If you are doing a major change (IE: Changing from a `tally` to a `nested tally`), **USE AN ENTIRELY NEW KEY NAME**.
```
moxian, line 198:
```suggestion
If you are doing a major change (i.e.: Changing from a `tally` to a `nested tally`), **USE AN ENTIRELY NEW KEY NAME**.
```

code/controllers/subsystem/blackbox.dm
```diff
@@ -0,0 +1,307 @@
...
+	if(depth == key_list.len)
+		if(L.Find(key_list[depth]))
+			L["[key_list[depth]]"] += increment
+		else
+			var/list/LFI = list(key_list[depth] = increment)
+			L += LFI
+	else
+		if(!L.Find(key_list[depth]))
+			var/list/LGD = list(key_list[depth] = list())
+			L += LGD
```
moxian, line 235: the mnemonics of `LFI` and `LGD` are entirely arcane to me. List.. First.. Increment and List... Global.. Data..?

code/controllers/subsystem/blackbox.dm
```diff
@@ -0,0 +1,307 @@
...
+	switch(key_type)
+		if("text")
+			if(!istext(data))
+				return
+			if(!islist(FV.json["data"]))
+				FV.json["data"] = list()
+			if(overwrite)
+				FV.json["data"] = data
+			else
+				FV.json["data"] |= data
```
moxian, line 191: Do we maybe want
```suggestion
			if(overwrite)
				FV.json["data"] = "[data]"
			else
				FV.json["data"] |= "[data]"
```
for when people inevitably accidentally a wrong type?

moxian (commented):
I still think that `SSblackbox.record_feedback("type",..)` should be `SSblackbox.record_feedback_type(..)` but other than that looks fine.

.github/USING_FEEDBACK_DATA.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Things you should and should not record
+
+Feedback data can be useful, depending on how you use it. You need to be careful with what you record to make sure you are not saving useless data. Examples of good things to record:
+
+- Antagonist win/loss rates if a new gamemode or antagonist is being added
+- Department performance (IE: Slime cores produced in science)
+- Basically anything which has actual meaning
+
+Examples of bad things to record:
```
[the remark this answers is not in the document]
AffectedArc07: Feedback data is meant to be anon unless we log things that are available to EVERYONE (EG: antag status).

As for GDPR, that doesn't matter here. We already have many other ckey entries in the DB, and if someone did invoke right to erasure we can just remove their rows.

code/controllers/subsystem/blackbox.dm
AffectedArc07: Types will be auto encoded
image

PR #15133, opened 2020-12-17
Shadowling hatch spell tweaks for lockers

Kyep:
## What Does This PR Do
1) Shadowlings who attempt to hatch while inside something that is not a turf (e.g: a locker) no longer have to wait out the full spell charge timer before they can attempt to hatch again. 
2) Additionally, the error message for attempting to hatch while not standing on a turf is now clearer.

## Why It's Good For The Game
1) Right now, shadowlings take constant burn damage once they go without hatching for too long. If a sling right now tries to hatch in a locker because they'll die if they don't, the fact that it fails and depletes their hatch spell charge means that they have to wait for the full charge timer before they can try again. In practice... that can kill them. Its reasonable for them to be forced out of the locker before they can hatch - but its not reasonable for a failed hatch attempt to delay them for several minutes before they can retry. That just results in slings dying unnecessarily because they did not realize they were not allowed to hatch while inside a locker, nor did they realize attempting to do so and failing would block attempts to re-hatch until after they died from burn damage.
2) The previous error message said that they could not hatch 'here', but it was unclear what 'here' meant. It could easily be interpreted to mean "anywhere near here". What it actually meant was "inside a container". The new error message is much clearer that they must be standing on a floor to hatch.

## Changelog
:cl: Kyep
tweak: tweaked shadowling hatch spell so that if you try to use it in a locker or other container, it gives a clear error message, and lets you try again immediately rather than making you wait a long time first.
/:cl:


code/game/gamemodes/shadowling/special_shadowling_abilities.dm
```diff
@@ -15,7 +15,8 @@ GLOBAL_LIST_INIT(possibleShadowlingNames, list("U'ruan", "Y`shej", "Nex", "Hel-u
 	if(user.stat || !ishuman(user) || !user || !is_shadow(user || isinspace(user)))
 		return
 	if(!isturf(user.loc))
-		to_chat(user, "<span class='warning'>You can't hatch here!</span>")
+		charge_counter = charge_max
```
farie82, line 18:
```suggestion
		revert_cast(user)
```

farie82 (changes requested):

farie82 (approved):

SteelSlayer (approved):

cluwneythescourge:
Thank god

Dumbdumn5 (approved):
Definitely a solid change, shouldn't exactly punish that heavily in the event somebody had no idea the spell'd fail in a locker. :+1:

AffectedArc07 (approved):

PR #15137, opened 2020-12-18
Makes closets Initialize()

AffectedArc07:
## What Does This PR Do
This PR makes closets use `Initialize()` over `New()`. It also moves their item spawning behaviour to its own `populate_contents()` proc, removing the need to null or reset `contents()` in chains of `New()` or `Initialize()`

## Why It's Good For The Game
The next crusade begins

## Changelog
:cl: AffectedArc07
tweak: [Backend] Closets now use a better spawning mechanism
/:cl:


AffectedArc07:
Testmerged with 0 issues

PR #15142, opened 2020-12-19
Adds gravaton grenades

Qwertytoforty:
## What Does This PR Do

This pr adds gravaton grenades, which are grenades that invert gravity in the area for 30 seconds, presuming that the area is not outside / space / a shuttle / a ruin / in a teleport locked area. If one is used in that area, it deactivates.

4 of them can be bought for 3 TC.

Adds a clusterbomb version, grenades that fail to activate from the clusterbomb will be deleted.

Adds a special_gravity variable to areas that is false everywhere by default, but tells the gravity checker to pay attention to the rooms gravity over  the gravity generator, if it exists, so the area can have no gravity while a gravity generator is active on that zlvl.

## TODO

- [x] Proper area variable name
- [x] Make the timer on the grenade for how long till gravity returns customizable in variables.

## Why It's Good For The Game

First off, rule of cool.

Second off, more seriously, this has a few practical uses. As a speed up, you don't slow down in no gravity, a way to mess with sec chasing you as their feet are no longer on the ground, way to cause light chaos / throw something farther, a way to trap a room so when someone leaves it, or the timer runs out, they feel the slam of gravity and get stunned, and such.

Also nice for admins who wish to make a zero gravity room on the station.

## Images of changes

See icon diff

## Changelog
:cl:
add: Adds gravaton grenades, grenades that invert gravity in an area for 30 seconds. Can buy 4 for 3 TC.
add: Adds a new variable to areas, special_gravity, that allow the area to ignore gravity generators on the same zlvl when it comes to no gravity
/:cl:


code/datums/uplink_item.dm
```diff
@@ -1083,6 +1083,15 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/storage/box/syndie_kit/emp
 	cost = 2
 
+/datum/uplink_item/explosives/gravity
+	name = "Gravaton Grenade Kit."
+	desc = "This box contains 4 gravaton grenades, which upon dentation release a pulse of negative gravatons, which invert gravity in the area. \
```
SteelSlayer, line 1088:
```suggestion
	desc = "This box contains 4 gravaton grenades, which upon detonation release a pulse of negative gravatons, which inverts gravity in the area. \
```

code/datums/uplink_item.dm
```diff
@@ -1083,6 +1083,15 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
...
+			Lasts for 30 seconds, good for moving fast, causing people to float by, chaos, or just for the slam of gravity returning 30 seconds later. \
+			Magboots reccomended."
+	reference = "GRAV"
+	item =/obj/item/storage/box/syndie_kit/gravity
```
SteelSlayer, line 1092:
```suggestion
	item = /obj/item/storage/box/syndie_kit/gravity
```

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,58 @@
+/obj/item/grenade/gravitygrenade
+	name = "Gravaton Grenade"
+	desc = "This grenade emmits a blast of negative gravatons, inverting gravity in the nearbye area for 30 seconds."
```
SteelSlayer, line 3:
```suggestion
	desc = "This grenade emmits a blast of negative gravatons, inverting gravity in the nearby area for 30 seconds."
```

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,58 @@
...
+	icon_state = "gravity"
+	item_state = "flashtool"
+	origin_tech = "magnets=4;combat=4"
+	///used if it is a clusterbuster, so it deletes if it fails to activate, and doesnt leave 10+ unactivated nades
+	var/clustersegment = FALSE
+	///How long untill gravity returns (In deciseconds)?
+	var/gravity_timer = 300
```
SteelSlayer, line 10:
```suggestion
	var/gravity_timer = 30 SECONDS
```

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,58 @@
...
+/obj/item/grenade/gravitygrenade/proc/unskrek_gravity(area/area)
+	if(area.has_gravity)
+		area.gravitychange(0, area)
+	else
+		area.gravitychange(1, area)
+
+	if(area.special_gravity)
+		area.special_gravity = FALSE
+	else
+		area.special_gravity = TRUE
```
SteelSlayer, line 49: I recommend changing this to something like:
```dm
/obj/item/grenade/gravitygrenade/proc/invert_gravity(area/area)
	area.gravitychange(!area.has_gravity, area)
	area.special_gravity = !area.special_gravity
```
Takes advantage of the fact you're dealing with booleans. You don't have to change the logic inside the proc if you don't want, but the point is, you can use this to replace the duplicate code from line 27 to 35.

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,58 @@
...
+
+/obj/item/grenade/gravitygrenade/proc/unprime()
+	active = 0
```
SteelSlayer, line 52:
```suggestion
	active = FALSE
```

code/game/objects/items/weapons/grenades/gravitygrenade.dm
Qwertytoforty: Done!

code/game/area/areas.dm
```diff
@@ -67,6 +67,8 @@
 	var/moving = FALSE
 	/// "Haunted" areas such as the morgue and chapel are easier to boo. Because flavor.
 	var/is_haunted = FALSE
+	///This variable, normaly set to true by a gravity grenade, means it checks this variable before it checks if a gravity generator exists on the level, meaning you can have a no gravity room on the same level as a generator.
+	var/special_gravity = FALSE
```
FlattestGuitar, line 71: Very non-descriptive variable name.

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,43 @@
...
+	var/area/area = get_area(src)
+	if(area.outdoors)
+		visible_message("<span class='warning'> The [src] will not work in such a big area!</span>")
+		unprime()
+		return
+	if(!is_teleport_allowed(area.z))
+		visible_message("<span class='warning'> The [src] seems to not work here!</span>")
+		unprime()
+		return
+	if(istype(area, /area/shuttle) || istype(area, /area/ruin))
```
FlattestGuitar, line 23: Why not make it work on the shuttle? That seems like the only actual use of this grenade.

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,43 @@
...
+		visible_message("<span class='warning'> The [src] seems to not work here!</span>")
+		unprime()
+		return
+	invert_gravity(area)
+	playsound(loc, 'sound/weapons/wave.ogg', 60, 1)
+	addtimer(CALLBACK(null, .proc/invert_gravity, area), gravity_timer)
+	qdel(src)
+
+/obj/item/grenade/gravitygrenade/proc/invert_gravity(area/area)
+	area.gravitychange(!area.has_gravity, area)
```
FlattestGuitar, line 33: The `gravitychange` proc is written horribly. You could take this opportunity to fix that. It takes an area as an argument, where it should just use `src` instead.

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,43 @@
...
+	area.special_gravity = !area.special_gravity
+
+/obj/item/grenade/gravitygrenade/proc/unprime()
+	active = FALSE
+	icon_state = initial(icon_state)
+	if(clustersegment)
+		qdel(src)
+
+/obj/item/grenade/gravitygrenade/clustersegment
```
FlattestGuitar, line 42: Could move this logic to all grenades. Make clusters set this var on their babies so you don't have to make weird subtypes. Not a huge issue though.

code/game/objects/items/weapons/grenades/gravitygrenade.dm
```diff
@@ -0,0 +1,43 @@
...
+		visible_message("<span class='warning'> The [src] seems to not work here!</span>")
+		unprime()
+		return
+	if(istype(area, /area/shuttle) || istype(area, /area/ruin))
+		visible_message("<span class='warning'> The [src] seems to not work here!</span>")
+		unprime()
+		return
+	invert_gravity(area)
+	playsound(loc, 'sound/weapons/wave.ogg', 60, 1)
+	addtimer(CALLBACK(null, .proc/invert_gravity, area), gravity_timer)
```
FlattestGuitar, line 29: Messing with the whole gravity system seems bad to me because it's not really written in a sensible way.

`atom/has_gravity()` despite first checking turf and area stuff will still return true if there's a local gravity gen present.
All while gravity gens try to set the gravity of areas while turning on.

Seems to me there are just better ways to keep track of this. Perhaps removing one of these altogether? This could break stuff, but I have a hard time understanding what the reasoning was behind the current system.

But that's all I could dig up in ten minutes. You can spend some more time looking at this. I guess in the end the maintainers will decide whether this is fine. Code quality is their thing.

code/game/area/areas.dm
Qwertytoforty: Changed to the one you suggested on discord

code/game/objects/items/weapons/grenades/gravitygrenade.dm
Qwertytoforty: Done

code/game/objects/items/weapons/grenades/gravitygrenade.dm
Qwertytoforty: Shuttles change zlvls, and are generally important to not be messed with, IE, arrivals, escape, so I decided not to have it work there

code/game/objects/items/weapons/grenades/gravitygrenade.dm
Qwertytoforty: Adjusted it a bit

Kyep (commented):
I don't like this.
Removing gravity from an area temporarily is... really not that useful. If you're close enough to throw one of these at sec, they're close enough to tase you.
In theory you could detonate this around a corner to impede their ability to chase you, but you'd have to know they were there, have the nades, throw them at the right time, etc. And even then, because /areas are not visible to players, the start/end points of the effect would seem very arbitary. Potentially if you throw these at a room edge you might find it lasts 10+ tiles in one direction but only 1 tile in the other direction. And lack of gravity won't stop someone walking along the wall... which given the narrow tunnels of maints they'll likely be able to do anyway. An item like this only really shines in big open areas, of which there are not many on the station, and those there are (atmos, bar, etc) are not common fight locations. Plus using one of these wouldn't stop someone standing at the door and shooting you. You'd still be better off just running away instead of using one of these.
In short, while it might seem like a neat idea, I feel it would be very underwhelming in practice.

Spacemanspark:
Does it need to be the most useful thing in the world? Its a neat idea that doesn't seem like it'll hurt anything.

PacifistDalek:
Personally I'm very much in favour of any traitor item that allows for more yakkety sax level chaos, without involving sudden random explosions and loss of life and limb

variableundefined (commented):
Personally, I am not really sold on messing around with gravity and adding complexity. Especially since the grenade doesn't seems too useful for your average traitor.

Qwertytoforty:
Liked this PR, really enjoyed the idea of messing with gravity as a traitor item, but unlikely to be merged, as such closing to reduce pr count. Anyone can feel free to use the idea for something later

PR #15147, opened 2020-12-19
Improved Wiki Button

Kyep:
## Changes & Reasoning
Partially reverts https://github.com/ParadiseSS13/Paradise/pull/14613
That PR was necessary due to a bug in Byond, which has since been fixed. So this PR restores 'search the wiki in-game' functionality.

With this PR, when you click the "Wiki" button, it once again asks you what page you want. You have 3 options:
- Hit OK, get taken to wiki homepage
- Change the input text to something else, like "Space Law" and then hit return, to be taken straight to that page
- Hit Cancel (does nothing)

This is better than the current system because instead of having to click "yes", end up on the homepage, enter your search terms, and possibly click through a search result.... you now get taken instantly to the page you want. Less clicking.

## Changelog
:cl: Kyep
add: Wiki button in-game once again lets you quickly jump to the wiki page you want, instead of forcing you to go via the homepage.
/:cl:


interface/interface.dm
```diff
@@ -4,8 +4,12 @@
 	set desc = "Type what you want to know about.  This will open the wiki in your web browser."
 	set hidden = 1
 	if(config.wikiurl)
-		if(alert("Open the wiki in your browser?", null, "Yes", "No") == "Yes")
+		var/query = stripped_input(src, "Enter Search:", "Wiki Search", "Homepage")
```
AffectedArc07, line 7: Add a parameter to this verb which can skip the popup entirely and take straight to the homepage, since that what the Wiki button in the changelog does. See https://github.com/ParadiseSS13/Paradise/blob/3ed5f7caf547a0b9513a87e6f8308ed3514e2b5e/code/controllers/subsystem/changelog.dm#L255

They would have to press buttons in 2 different dialogue boxes, which is awful

AffectedArc07 (changes requested):

interface/interface.dm
Kyep: Deleted the first button, kept the second (since it allows all 3 options, whereas the 1st one did not).

AffectedArc07, line 7:
```suggestion
		var/query = stripped_input(src, "Enter Search. Leave the box blank to cancel", "Wiki Search", "Homepage")
```

AffectedArc07 (changes requested):
Smidge of clarification

interface/interface.dm
Kyep: That's not how it works - there's an actual "cancel" button.

AffectedArc07 (approved):

AffectedArc07 (approved):

PR #15148, opened 2020-12-19
Boxing gloves haymaker now weakens instead of sleeping

dearmochi:
## What Does This PR Do

The haymaker move from the boxing martial arts (bestowed when wearing boxing gloves) now weakens for 5 cycles instead of putting to sleep.

## Why It's Good For The Game

- Boxing gloves can be used in combination with stamina weapons to reliably and quickly put someone to sleep after enough stamina damage. This issue is most problematic with contractor batons as they deal 70 stamina damage, unlocking the haymaker move immediately if the attacker is wearing boxing gloves, making for an overpowered combo.
- It's annoying to win a boxing match and having to wait for the opponent to wake up to call them a loser.

## Changelog
:cl:
tweak: Boxing gloves can no longer put to sleep - they instead weaken the victim, rendering them unable to move but still conscious
/:cl:

### DIFF
--- a/code/modules/martial_arts/brawling.dm
+++ b/code/modules/martial_arts/brawling.dm
@@ -14,7 +14,7 @@
 	A.do_attack_animation(D, ATTACK_EFFECT_PUNCH)
 
 	var/atk_verb = pick("left hook","right hook","straight punch")
-	
+
 	var/damage = rand(5, 8) + A.dna.species.punchdamagelow
 	if(!damage)
 		playsound(D.loc, 'sound/weapons/punchmiss.ogg', 25, 1, -1)
@@ -39,7 +39,7 @@
 			D.visible_message("<span class='danger'>[A] has knocked [D] out with a haymaker!</span>", \
 								"<span class='userdanger'>[A] has knocked [D] out with a haymaker!</span>")
 			D.apply_effect(10,WEAKEN,armor_block)
-			D.SetSleeping(5)
+			D.Weaken(5)
 			D.forcesay(GLOB.hit_appends)
 		else if(D.lying)
 			D.forcesay(GLOB.hit_appends)


Carthusia:
I dono, it was always kind of fun to knock them out, perhaps a short KO, followed by the stamina stun leaving them on the ground.

Boxing gloves have in the past been used as a method of kidnapping stun as well, challenge them to a boxing match, knock them out, then drag them off, zip them and what have you.

If their only stunned, but conscious, they can easily call out for help.

Cocacolagua:
the way that boxing gloves work is that they put you to sleep if the victim has enough stamina damage, this means that you could pair it with anything that deals stamina damage for free sleeping, and since sleeps are a straight upgrade to the stamina stun that is already crazy, it becomes messy.

in short: boxing gloves bad, good change

Carthusia:
Is there some way to prevent the combining of boxing gloves with other items?

SteelSlayer (approved):

dearmochi:
Realistically, no because there are a ton of edge cases you have to consider, hitting someone with a disabler/contractor baton then equipping the gloves and slapping the victim.

Carthusia:
Hmmm, what if you enable a kind of block, where if you are wearing boxing gloves, you cannot attack other weapons, similar to how they also dont let you disarm.

This would make them explicitly for knocking people out in a fight, and not exploitable to stack stamina damage on weapons that dont get it.

dearmochi:
That still doesn't fix the problem of using the disabler/contractor baton THEN equipping the gloves and hitting with them, though?

Dashsora:
I don't think whether the person loosing a boxing fight gets slept or not will change anything.. however getting a free, almost instant, sleep in addition to the contractor batons/disablers is absolutely busted.

Much needed change, good stuff

Carthusia:
> That still doesn't fix the problem of using the disabler/contractor baton THEN equipping the gloves and hitting with them, though?

Sorry late reply, for some reason i didnt get the ping.

It may not stop that method, but it mitigates it.

To resolve it entirely, you have to basically destroy one option people use in the way its intended, to prevent another option that was unintended, overall the prefered fix would be to maintain the original option while preventing the unintended option.

Its also still probably a good idea to implement the weapon limitation regardless though as as noted above, even if say im stabbing someone to death with a knife, that knife is now adding boxing glove stamina damage, making my job stabbin, that much easier when they fall over after about 5-6 stabs.

AffectedArc07 (approved):

PR #15151, opened 2020-12-19
Spiderling Pathing Tweaks

Kyep:
## What Does This PR Do
Spiderlings (both terror and giant) will no longer:

1. attempt to path into solid walls (or other dense turfs)
2. attempt to path into non-simulated turfs, like space
3. potentially runtime if there are no available turfs they can path to
4. have copy/pasted duplicate code instead of re-using the same proc
5. have their pathing destinations weighted by # of objects on a turf (thus making them 10x as likely to path to a stack of 10 objects than to a single object, and thus resulting in spiderlings stacking up and making an already crowded tile even more crowded, and harder to find a single object on)
6. be instantly qdeled upon touching a recharging station (this was a really dumb workaround I implemented years ago for a runtime that would otherwise happen when a spiderling touched the recharge station and the recharge station attempted to move the spiderling inside it, only to find out it wasn't a mob). I intended to come back to it and implement something better, but I had not... until now. Now it behaves as you'd expect: nothing happens.

## Why It's Good For The Game
1. Things that move should not be attempting to path into solid walls.
2. Terror spiderlings were meant to avoid pathing into space. But this did not actually work as they could still path to objects on space turfs (like girdirs, or lattice) which in practice meant they still pathed into space even though they were not meant to. This fixes that. Note: spiderlings DON'T take damage from being in space, so this doesn't help them grow... it just stops them looking absolutely stupid by hanging out in space while they grow.
3. Runtimes are bad
4. Copy/pasted duplicate code is bad
5. Spiderlings attempting to stack up on top of turfs which already have tons of objects on them is ... contributing to the problem of too many things on one tile.
6. The old spiderling/rechargestation code was an incredibly hacky workaround. It used a manual istype check. It silently deleted spiderlings with no body, no remains, no notice. It did not even call their death proc it just qdeled them on contact. It was atrocious. The new code has no istype checks, generates the behavior you'd expect (nothing), and even as a side-benefit handles the case of other non-mobs bumping the recharge station, which prevents other similar runtimes from happening in future.

## Changelog
:cl: Kyep
tweak: tweaked spiderling pathing code to fix cases of spiderlings wrongly spacing themselves, running into walls, using duplicated code, potentially generating runtimes, stacking up on one tile, or getting silently deleted when they touch a recharge station.
/:cl:


SabreML:
RIP spiderling space program, gone so soon.

TDSSS:
>attempt to path into non-simulated turfs, like space

Does this break spiderlings on z-2? What happens if, through an event or similar, spiderlings were to get to centcom, the syndie forward base, etc?

Or well, break is a strong word, they'd just potentially not move at all and throw runtimes.

Kyep:
@TDSSS If they're spawned in an area with only unsimulated turfs (like CC, or syndie HQ), they won't move. This is intentional. They don't need to move in these areas. Most of the time there will be nobody there to attack them. If there is, you can always spawn fully-grown spiders directly, you don't need to even spawn spiderlings there. In fact, you probably shouldn't be spawning spiderlings in those areas for other reasons (their growing type is random, unlike eggs or directly spawned spiders).

code/game/objects/effects/spiders.dm
```diff
@@ -190,6 +186,19 @@
...
+	var/list/available_turfs = list()
+	for(var/turf/simulated/S in oview(10, src))
+		// no !isspaceturf check needed since /turf/simulated is not a subtype of /turf/space
+		if(S.density)
+			continue
+		available_turfs += S
+	if(!length(available_turfs))
+		return FALSE
+	var/turf/T = pick(available_turfs)
+	walk_to(src, T)
```
SteelSlayer, line 199:
```suggestion
	walk_to(src, pick(available_turfs))
```

AffectedArc07 (approved):

PR #14975, opened 2020-11-23
Changes atmospherics machinery plane from "FLOOR_PLANE" to "GAME_PLANE"

FreeStylaLT:
## What Does This PR Do
Changes the atmospherics' plane from -2 to -1

## Why It's Good For The Game
Allows mappers to *see the pipes*, as -2 is under the walls, resulting in:

## Images of changes
**VIEW FROM A MAPPING TOOL (Identical results with StrongDMM and Dreammaker)**
## **BEFORE:**
StrongDMM_2020-11-23_18-29-59

## **AFTER:**
StrongDMM_2020-11-23_18-24-25

Was briefly tested, no ill effects were found in-game at a quick glance.

## Changelog
:cl:FreeStylaLT
tweak:Changed atmos machinery plane.
/:cl:

PR #15152, opened 2020-12-20
Changes vent, scrubber, and passive vent plane back to FLOOR_PLANE

SteelSlayer:
## What Does This PR Do
Changes vents, scrubbers and passive vents to use `FLOOR_PLANE` instead of `GAME_PLANE`, partially reverting what was done in #14975. All other atmospherics machines still use the `GAME_PLANE`. The only thing that's changed is vents, scrubbers and passive vents.

## Why It's Good For The Game
Vents/scrubbers having a shadow under them looks odd.

## Images of changes
Left: before, Right: after
unknown

## Changelog
:cl:
tweak: Changed vent, scrubber and passive vent plane, removing the odd shadow that was appearing under them.
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/passive_vent.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/passive_vent.dm
@@ -1,7 +1,7 @@
 /obj/machinery/atmospherics/unary/passive_vent
 	icon = 'icons/atmos/vent_pump.dmi'
 	icon_state = "map_vent"
-
+	plane = FLOOR_PLANE
 	name = "passive vent"
 	desc = "A large air vent"
 
--- a/code/modules/atmospherics/machinery/components/unary_devices/vent_pump.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/vent_pump.dm
@@ -9,7 +9,7 @@
 	name = "air vent"
 	desc = "Has a valve and pump attached to it"
 	use_power = IDLE_POWER_USE
-
+	plane = FLOOR_PLANE
 	layer = GAS_SCRUBBER_LAYER
 
 	can_unwrench = 1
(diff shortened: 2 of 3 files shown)


moxian:
Needs at least a comment so that the next person does not revert them back again.
Either on the machines or (better?) on the FLOOR_PLANE define itself

moxian:
Would also be a good idea to figure out where the shadows are created, and if maybe there's a (better?) way to have GAME_PLANE objects without shadows.

FreeStylaLT:
your steel looks odd 😠

SteelSlayer:
@moxian I looked into it, it's not possible as far as I can tell. Or I may just lack the knowledge.

AffectedArc07:
> Would also be a good idea to figure out where the shadows are created, and if maybe there's a (better?) way to have GAME_PLANE objects without shadows.

They are generated with the ambient occlusion filter
```dm
#define FILTER_AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, color="#04080FAA")

/obj/screen/plane_master/game_world
	name = "game world plane master"
	plane = GAME_PLANE
	appearance_flags = PLANE_MASTER
	blend_mode = BLEND_OVERLAY

/obj/screen/plane_master/game_world/backdrop(mob/mymob)
	clear_filters()
	if(istype(mymob) && mymob.client && mymob.client.prefs && (mymob.client.prefs.toggles & PREFTOGGLE_AMBIENT_OCCLUSION))
		filters += FILTER_AMBIENT_OCCLUSION

```
It literally applies *to* the game plane. If you make objects on the game plane ignore shadows, you remove shadows entirely.

Kyep (approved):
My understanding is that this removes the incorrect shadow from vents/scrubbers/etc, without affecting anything else.
Since the changes made previously to make these objects show up on top of walls, while helpful, aren't necessary for vents/etc as you don't place vents on a wall.

So, I consider this PR a fix, and not something that heads should need to vote on.
However, since its classified as a tweak (why?) I'm voting for it.

AffectedArc07 (approved):

PR #15155, opened 2020-12-20
Removes ability to put collars on jumpsuits

AffectedArc07:
# Preliminary Note
This has already been discussed internally between admins and heads, and we have decided we want this. Please do not reeeee in the comments.

## What Does This PR Do
Removes the ability to put collars on jumpsuits.

## Why It's Good For The Game
People only ever add collars to jumpsuits for weird petplay fetish things, which we dont want here. Its time to stop this.

## Changelog
:cl: AffectedArc07
del: You can no longer put pet collars on jumpsuits
/:cl:

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -668,12 +668,16 @@ BLIND     // can't see anything
 	var/list/accessories = list()
 
 	/// List of blacklisted accessory types
-	var/list/blacklisted_accessory_typecache = list(
-		/obj/item/clothing/accessory/petcollar // No collars on jumpsuits
-	)
+	var/list/blacklisted_accessory_typecache
 	var/displays_id = 1
 	var/rolled_down = 0
 	var/basecolor
+
+/obj/item/clothing/under/Initialize(mapload)
+	. = ..()
+	blacklisted_accessory_typecache = typecacheof(list(
+		/obj/item/clothing/accessory/petcollar // No collars on jumpsuits
+	))
 
 /obj/item/clothing/under/rank/New()
 	if(random_sensor)
@@ -690,7 +694,7 @@ BLIND     // can't see anything
 	else
 		return FALSE
 
-	if(A.type in blacklisted_accessory_typecache)
+	if(is_type_in_typecache(A, blacklisted_accessory_typecache))
 		to_chat(usr, "<span class='notice'>[A] doesn't fit on [src].</span>")
 		return FALSE
 
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -666,6 +666,11 @@ BLIND     // can't see anything
 		3 = Report location
 		*/
 	var/list/accessories = list()
+
+	/// List of blacklisted accessory types
+	var/list/blacklisted_accessory_typecache = list(
+		/obj/item/clothing/accessory/petcollar // No collars on jumpsuits
+	)
 	var/displays_id = 1
 	var/rolled_down = 0
 	var/basecolor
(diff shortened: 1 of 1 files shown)


DoctorDrugs:
REEEEEEEEEEEEEEEEEEEEEEEE (because you said not to and i quote "i will repo ban you")

Necaladun:
REEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEally about time.

code/modules/clothing/clothing.dm
```diff
@@ -666,6 +666,11 @@ BLIND     // can't see anything
 		3 = Report location
 		*/
 	var/list/accessories = list()
+
+	/// List of blacklisted accessory types
+	var/list/blacklisted_accessory_typecache = list(
```
dearmochi, line 671: `typecacheof`?

code/modules/clothing/clothing.dm
```diff
@@ -685,11 +690,17 @@ BLIND     // can't see anything
 	else
 		return FALSE
 
+	if(A.type in blacklisted_accessory_typecache)
```
dearmochi, line 693: `is_type_in_typecache`?

farie82:
Why keep it as a `obj/item/clothing/accessory` if you can't equip it as an accessory? Sounds like this just needs to become something else that is only useable by pets or so

Aurorablade:
Look if people are using it for pet play fetish things we need to crack down on them or add some kind of alternative like actually military dogtags. Editable in the same vein if we are going to remove this. I really dislike removing based on the people using it for weird fetish crap.

AffectedArc07:
> Why keep it as a `obj/item/clothing/accessory` if you can't equip it as an accessory? Sounds like this just needs to become something else that is only useable by pets or so

Because refactoring the pathtypes is out the scope of this PR, and I dont want to inadvertently break corgi access by doing that

Spacemanspark:
@aurorablade
I'd be just fine with adding an alternative that wouldn't be used for kinky purposes so easily. Military dogtags would be neat. Someone has also suggested chokers, which I believe is far less of a fetish item and would do well for those looking for the 'punk kid' angle. 

That's all beyond the scope of this PR however.

This item has long been a problem. It can still be used as a pet accessory as it's primary intended use.

AffectedArc07:
Tagged `Controversial` even though it really shouldn't be

Cocacolagua:
lol

ZanderDoggo:
I believe that there should be an alternative, or just adding in a new rule about it for using the collars in an ERP way. The main problem I have with removing collars is that you then can't clip IDs to them, that's what I use them for. So having dog tags as an alternative where you can still clip an ID to it would be good.

Spacemanspark:
>just adding in a new rule about it for using the collars in an ERP way

Our server rules already cover this sort of behavior.
The problem is that people, even regulars, still toe the line with the item... and honestly there's no real good reason for it to exist as it is. We can tell people not to do something over and over, but when the behavior is consistent across several different types of people, we can't really just take the usual ban hammer approach. 

Again, alternatives would be great to have in a future PR should someone want to sprite up new things. But as it is, the item treads far too close to the line... and we've only really ever seen one type of person using it for anything but a gross fetish purpose or joke.

procdrone:
what the fuck

farie82:
> > Why keep it as a `obj/item/clothing/accessory` if you can't equip it as an accessory? Sounds like this just needs to become something else that is only useable by pets or so
> 
> Because refactoring the pathtypes is out the scope of this PR, and I dont want to inadvertently break corgi access by doing that

Yet will leave technical debt this way. An accessory that can't be used as one. Why is it one then?

AffectedArc07:
> > > Why keep it as a `obj/item/clothing/accessory` if you can't equip it as an accessory? Sounds like this just needs to become something else that is only useable by pets or so
> > 
> > 
> > Because refactoring the pathtypes is out the scope of this PR, and I dont want to inadvertently break corgi access by doing that
> 
> Yet will leave technical debt this way. An accessory that can't be used as one. Why is it one then?

If you wanna clean it up after the fact and double check things, you are more than welcome to

Carthusia:
Wait...this is actually happening? Seriously?

Is it in like the darkest depths of maint that no light shall escape from, and all eyes who behold it shall be clawed out from thine head as to prevent madness?

Spacemanspark:
No, no, the madness has been amongst you this whole time, in fact. Walking among you as you stroll to the bridge to gaze upon the captain, watching you.

If you are not careful, the curse will catch you too, dear Kar-Ski. Beware.

Carthusia:
The mad laughs of dead gods...

Kyep (approved):

Kyep:
For clarity: I wasn't the first head to vote in favor of this, so my approval isn't necessary for this to pass, but I'll still add my approval anyway because I feel its important for our community to know that this change has broad support among the staff team.

variableundefined (approved):

Bkulli-VI:
But where will I hide the spare now?

PR #15158, opened 2020-12-20
More New() -> Initialize() for /obj/structure

AffectedArc07:
## What Does This PR Do
Converts more structures to `Initialize()`. A few of them still remain on `/New()` since they require fleshing out and testing in their own PR due to their specifics (IE: Blob stuff).

A lot of the changes here are simple ones which only did stuff like setting icon states, or adding to global lists. Some tweaks have been commented on below in the review.

Remaining conversions for `/obj/structure` are below
image

## Why It's Good For The Game
`New()` bad

## Changelog
:cl: AffectedArc07
add: [Backend] Most /obj/structure types now spawn with the SSatoms instead of /New()
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult_structures.dm
+++ b/code/game/gamemodes/cult/cult_structures.dm
@@ -140,7 +140,7 @@
 	choosable_items = list("Eldritch Whetstone" = /obj/item/whetstone/cult, "Flask of Unholy Water" = /obj/item/reagent_containers/food/drinks/bottle/unholywater,
 							"Construct Shell" = /obj/structure/constructshell)
 
-/obj/structure/cult/functional/altar/New()
+/obj/structure/cult/functional/altar/Initialize(mapload)
 	. = ..()
 	icon_state = SSticker.cultdat?.altar_icon_state
 
@@ -160,7 +160,7 @@
 	choosable_items = list("Shielded Robe" = /obj/item/clothing/suit/hooded/cultrobes/cult_shield, "Flagellant's Robe" = /obj/item/clothing/suit/hooded/cultrobes/flagellant_robe,
 							"Mirror Shield" = /obj/item/shield/mirror)
 
-/obj/structure/cult/functional/forge/New()
+/obj/structure/cult/functional/forge/Initialize(mapload)
 	. = ..()
 	icon_state = SSticker.cultdat?.forge_icon_state
 
@@ -217,16 +217,13 @@ GLOBAL_LIST_INIT(blacklisted_pylon_turfs, typecacheof(list(
 	var/corrupt_delay = 50
 	var/last_corrupt = 0
 
-/obj/structure/cult/functional/pylon/New()
+/obj/structure/cult/functional/pylon/Initialize(mapload)
 	. = ..()
+	START_PROCESSING(SSobj, src)
 	icon_state = SSticker.cultdat?.pylon_icon_state
 
 /obj/structure/cult/functional/pylon/attack_hand(mob/living/user)//override as it should not create anything
 	return
-
-/obj/structure/cult/functional/pylon/New()
-	START_PROCESSING(SSobj, src)
-	..()
 
 /obj/structure/cult/functional/pylon/Destroy()
 	STOP_PROCESSING(SSobj, src)
@@ -307,7 +304,7 @@ GLOBAL_LIST_INIT(blacklisted_pylon_turfs, typecacheof(list(
 	choosable_items = list("Shuttle Curse" = /obj/item/shuttle_curse, "Zealot's Blindfold" = /obj/item/clothing/glasses/hud/health/night/cultblind,
 							"Veil Shifter" = /obj/item/cult_shift) //Add void torch to veil shifter spawn
 
-/obj/structure/cult/functional/archives/New()
+/obj/structure/cult/functional/archives/Initialize(mapload)
 	. = ..()
 	icon_state = SSticker.cultdat?.archives_icon_state
 
--- a/code/game/objects/effects/decals/contraband.dm
+++ b/code/game/objects/effects/decals/contraband.dm
@@ -12,8 +12,8 @@
 	var/poster_type
 	var/obj/structure/sign/poster/poster_structure
 
-/obj/item/poster/New(loc, obj/structure/sign/poster/new_poster_structure)
-	..()
+/obj/item/poster/Initialize(mapload, obj/structure/sign/poster/new_poster_structure)
+	. = ..()
 	poster_structure = new_poster_structure
 	if(!new_poster_structure && poster_type)
 		poster_structure = new poster_type(src)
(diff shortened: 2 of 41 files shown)


code/game/gamemodes/cult/cult_structures.dm
```diff
@@ -217,17 +217,14 @@ GLOBAL_LIST_INIT(blacklisted_pylon_turfs, typecacheof(list(
...
-/obj/structure/cult/functional/pylon/New()
+/obj/structure/cult/functional/pylon/Initialize(mapload)
 	. = ..()
+	START_PROCESSING(SSobj, src)
 	icon_state = SSticker.cultdat?.pylon_icon_state
 
 /obj/structure/cult/functional/pylon/attack_hand(mob/living/user)//override as it should not create anything
 	return
 
-/obj/structure/cult/functional/pylon/New()
```
AffectedArc07, line 227: Duplicated `/New()` definition

code/game/objects/structures/electricchair.dm
```diff
@@ -7,19 +7,19 @@
 	var/last_time = 1.0
 	var/delay_time = 50
 
-/obj/structure/chair/e_chair/New()
-	..()
+/obj/structure/chair/e_chair/Initialize(mapload)
+	. = ..()
 	overlays += image('icons/obj/chairs.dmi', src, "echair_over", MOB_LAYER + 1, dir)
-	spawn(2)
```
AffectedArc07, line 13: I see no reason for this to be inside a `spawn()` bracket, but if anyone has insight feel free to inform

code/game/objects/structures/flora.dm
```diff
@@ -15,33 +15,29 @@
...
+/obj/structure/flora/tree/pine/Initialize(mapload)
+	. = ..()
 	icon_state = "pine_[rand(1, 3)]"
 
 /obj/structure/flora/tree/pine/xmas
 	name = "xmas tree"
 	icon = 'icons/obj/flora/pinetrees.dmi'
 	icon_state = "pine_c"
 
-/obj/structure/flora/tree/pine/xmas/New()
```
AffectedArc07, line 27: Useless `/New()`

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -18,12 +18,6 @@
 	var/propelled = FALSE // Check for fire-extinguisher-driven chairs
 	var/comfort = 0
 
-/obj/structure/chair/New()
-	..()
-	spawn(3)	//sorry. i don't think there's a better way to do this.
-		handle_rotation()
```
AffectedArc07, line 24: This proc is meant to handle rotating mobs buckled to a chair when the chair spawns and swaps direction. I cannot think of a single instance where theres a chair being spawned at mapload with a mob somehow pre-buckled that needs to be set in a specific direction.

code/modules/library/random_books.dm
```diff
@@ -27,7 +27,7 @@
 	icon_state = "random_bookcase"
 	anchored = TRUE
 
-/obj/structure/bookcase/random/New()
+/obj/structure/bookcase/random/Initialize()
```
AffectedArc07, line 30: This should actually fix random bookcases. Currently, because they `/New()`, they are created before `world/New()` has progressed, thus trying to spawn books without a database connection existing.

code/modules/mob/living/simple_animal/hostile/terror_spiders/reproduction.dm
```diff
@@ -209,27 +210,29 @@
...
 
-/obj/structure/spider/eggcluster/terror_eggcluster/New()
-	..()
+/obj/structure/spider/eggcluster/terror_eggcluster/Initialize(mapload)
+	. = ..()
 	GLOB.ts_egg_list += src
-	spawn(50)
-		if(spiderling_type == /mob/living/simple_animal/hostile/poison/terror_spider/red)
+
+/obj/structure/spider/eggcluster/terror_eggcluster/proc/update_name()
```
AffectedArc07, line 217: Moved to its own proc to kill off a `spawn(50)`. Will await response from @Kyep to see if this was done in a super specific way, but from what I can see, the change on 191 should fix any issues this has.

code/modules/paperwork/filingcabinet.dm
```diff
@@ -187,9 +187,9 @@ GLOBAL_LIST_EMPTY(employmentCabinets)
 	icon_state = "employmentcabinet"
 	var/populated = FALSE
 
-/obj/structure/filingcabinet/employment/New()
+/obj/structure/filingcabinet/employment/Initialize(mapload)
+	. = ..()
```
AffectedArc07, line 191: Why the hell do we track a GLOB of filing cabinets

code/modules/recycling/disposal.dm
```diff
@@ -1138,11 +1137,10 @@
...
 	var/obj/linked 	// the linked obj/machinery/disposal or obj/disposaloutlet
 
-/obj/structure/disposalpipe/trunk/New()
-	..()
+/obj/structure/disposalpipe/trunk/Initialize(mapload)
+	. = ..()
 	dpdir = dir
-	spawn(1)
-		getlinked()
+	addtimer(CALLBACK(src, .proc/getlinked), 0) // This has a delay of 0, but wont actually start until the MC is done
```
AffectedArc07, line 1143: This `addtimer()` should be better than the removed spawn, and more instant when running the round as opposed to a 1 tick delay to account for mapping weirdness.

code/game/objects/effects/decals/contraband.dm
```diff
@@ -12,8 +12,8 @@
 	var/poster_type
 	var/obj/structure/sign/poster/poster_structure
 
-/obj/item/poster/New(loc, obj/structure/sign/poster/new_poster_structure)
-	..()
+/obj/item/poster/Initialize(mapload, obj/structure/sign/poster/new_poster_structure)
+	. = ..()
```
AffectedArc07, line 16: While this is an `/obj/item`, not an `/obj/structure`, this had to also be done since posters were converted, and this would load before its internal poster could `Initialize()`, hence the conversion. 

Note, there is not a race condition here. Any new atoms which are spawned during the initialization of SSatoms are immediately initialized before the list continues.

code/game/objects/structures/electricchair.dm
farie82: Because `part` is not an argument and is passed later when creating it.
image

code/modules/mob/living/simple_animal/hostile/terror_spiders/reproduction.dm
```diff
@@ -188,6 +188,7 @@
 	stop_automated_movement = 1
 	var/obj/structure/spider/eggcluster/terror_eggcluster/C = new /obj/structure/spider/eggcluster/terror_eggcluster(get_turf(src))
 	C.spiderling_type = lay_type
```
farie82, line 190: Sounds like it would be better to pass these as arguments

code/game/objects/structures/electricchair.dm
AffectedArc07: TODO: Refactor that to take it as an arg

code/game/gamemodes/cult/cult_structures.dm
SabreML: For the record, blood cult didn't touch these 😓

code/game/objects/structures/electricchair.dm
AffectedArc07: Done

AffectedArc07:
Testmerged with no issues

PR #15159, opened 2020-12-20
Combines some admin verbs together

SabreML:
## What Does This PR Do

Moves the admin verbs `Jump to Area`, `Jump to Mob`, `Jump to Key`, and `Jump to Coordinate` to a new verb named `Jump to...` in order to reduce clutter.

I was planning on doing a few others too but in the end I couldn't think of any, so suggestions would be great.

## Why It's Good For The Game

The admin verb list is amazingly cluttered, and since these are all related it makes sense to combine them into one.

While you could say it would mess with muscle memory, I feel like that argument would just lead to things getting progressively worse.

## Images of changes

image

## Changelog
:cl:
tweak: The admin verbs 'Jump to Area', 'Jump to Mob', 'Jump to Key', and 'Jump to Coordinate' have all been moved into a new verb named 'Jump to...' in order to reduce clutter.
/:cl:


AffectedArc07:
While I appreciate the concept, I feel like a **lot** of admins will scream at the changes due to muscle memory being screwed. Approved for code, neutral on design.

code/modules/admin/verbs/adminjump.dm
```diff
@@ -1,12 +1,49 @@
...
+		if("Mob")
+			jumping = input(src, "Mob to jump to", "Jump to Mob") as null|anything in GLOB.mob_list
+			if(jumping)
+				return jumptomob(jumping)
+		if("Key")
+			var/list/keys = list()
+			for(var/I in GLOB.player_list)
+				var/mob/M = I
+				keys += M.client
+			jumping = input(src, "Key to jump to", "Jump to Key") as null|anything in sortKey(keys)
```
AffectedArc07, line 29:
```suggestion
			jumping = input(src, "Key to jump to", "Jump to Key") as null|anything in sortKey(GLOB.clients)
```
Try this

code/modules/admin/verbs/adminjump.dm
```diff
@@ -46,14 +79,11 @@
...
-	return
+	feedback_add_details("admin_verb", "JT") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
-/client/proc/jumptomob(var/mob/M in GLOB.mob_list)
-	set category = "Admin"
+/client/proc/jumptomob(mob/M)
 	set name = "Jump to Mob"
-
-	if(!check_rights(R_ADMIN))
+	if(!check_rights(R_ADMIN) || !M)
```
farie82, line 86:
```suggestion
	if(!M || !check_rights(R_ADMIN))
```
Cheaper checks usually go first

code/modules/admin/verbs/adminjump.dm
```diff
@@ -46,14 +79,11 @@
...
 		message_admins("[key_name_admin(usr)] jumped to [T.x], [T.y], [T.z] in [T.loc]", 1)
 	admin_forcemove(usr, T)
-	feedback_add_details("admin_verb","JT") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
-	return
+	feedback_add_details("admin_verb", "JT") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
-/client/proc/jumptomob(var/mob/M in GLOB.mob_list)
-	set category = "Admin"
+/client/proc/jumptomob(mob/M)
 	set name = "Jump to Mob"
```
farie82, line 85:
```suggestion
```
Since it's hidden now

code/modules/admin/verbs/adminjump.dm
```diff
@@ -1,12 +1,45 @@
...
+			if(!y)
+				return
+			var/z = input(src, "Z Coordinate", "Jump to Coordinates") as null|num
+			if(!z)
+				return
+			return jumptocoord(x, y, z)
+
+
+/client/proc/jumptoarea(area/A)
+	if(!check_rights(R_ADMIN) || !A)
```
farie82, line 42:
```suggestion
	if(!A || !check_rights(R_ADMIN))
```

code/modules/admin/verbs/adminjump.dm
```diff
@@ -15,7 +48,7 @@
 			continue
 		if(locate(/obj/structure/grille, T)) // Quick check to not spawn in windows
 			continue
-		turfs.Add(T)
+		turfs += T
 
 	var/turf/T = pick_n_take(turfs)
```
farie82, line 53:
```suggestion
	var/turf/T = safepick(turfs)
```
No clue why they `pick_n_take` here... maybe to avoid empty list exceptions? anyway `safepick` does that but cheaper

code/modules/admin/verbs/adminjump.dm
```diff
@@ -87,25 +114,14 @@
...
 	if(!isobserver(usr))
 		message_admins("[key_name_admin(usr)] jumped to coordinates [tx], [ty], [tz]")
 
-/client/proc/jumptokey()
-	set category = "Admin"
-	set name = "Jump to Key"
-
-	if(!check_rights(R_ADMIN))
+/client/proc/jumptokey(client/C)
+	if(!check_rights(R_ADMIN) || !C.mob)
```
farie82, line 122:
```suggestion
	if(!C?.mob || !check_rights(R_ADMIN))
```
Also adds a null check to C. Not just the mob

farie82 (changes requested):

code/modules/admin/verbs/adminjump.dm
SabreML: I tried this, but the proc also shows up in the right click menu for mobs, so it still needs to have a name set.

code/modules/admin/verbs/adminjump.dm
SabreML: Yeah not sure why it's removing stuff from the list.

farie82 (approved):

SabreML:
Go away conflict, I fixed you already

Dumbdumn5 (approved):
Though muscle memory is admittedly a thing that'd be affected, there was a similar issue when a number of verbs were removed from the menu and things were shuffled around. Considering this works to reduce the clutter a bit further without removing any of the verbs, 👍

Fox-McCloud:
Don't know how I feel about this. More often than not I'm only using one of these (jump to mob) and the other times it's occasional usage of jump to area--dunno if I really like having to two operations when 2 of these largely go unused (but are still necessary).

SabreML:
> Don't know how I feel about this. More often than not I'm only using one of these (jump to mob) and the other times it's occasional usage of jump to area--dunno if I really like having to two operations when 2 of these largely go unused (but are still necessary).

You could also reverse this argument by saying that if two of them are generally unused, they shouldn't be taking up space on the main 'Admin' tab.
In the end it's just a matter of what is more convenient to use.

dearmochi (approved):

Spacemanspark:
I'm fine with this myself, searching for one verb instead of searching among four is much easier.

code/modules/admin/verbs/adminjump.dm
```diff
@@ -30,9 +63,9 @@
 	log_admin("[key_name(usr)] jumped to [A]")
 	if(!isobserver(usr))
 		message_admins("[key_name_admin(usr)] jumped to [A]")
-	SSblackbox.record_feedback("tally", "admin_verb", 1, "Jump To Area") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
```
AffectedArc07, line 33: Leave these notes in. Not only is it an un-needed diff to remove them, its very important that this is known for anyone who adds blackbox feedback for verbs

AffectedArc07 (changes requested):

code/modules/admin/verbs/adminjump.dm
SabreML: Not sure why I even removed those to be honest. Presumably because the second parameter is the same across all of them, so I thought the comment was outdated or something.
🤷

code/modules/admin/verbs/adminjump.dm
```diff
@@ -1,23 +1,56 @@
-/client/proc/Jump(area/A in return_sorted_areas())
-	set name = "Jump to Area"
-	set desc = "Area to jump to"
+/client/verb/jump_to()
+	set name = "Jump to..."
```
AffectedArc07, line 2: Making this `/client/verb` makes it accessible to everyone regardless of rank. They wont be able to use it but it will be misleading.

AffectedArc07 (changes requested):

code/modules/admin/verbs/adminjump.dm
AffectedArc07: image

SabreML: I'm not able to reproduce it showing up there, but clearly it does so...

AffectedArc07 (approved):

PR #15161, opened 2020-12-20
A lot more Cult fixes

### LINKED ISSUE #14483: Giving a traitor assassinate with a random target through traitor panel can select off station roles
**Issue Description**:

Giving a traitor assassinate with a random target through traitor panel can select off station roles 

For it to exclude off station roles in its random generation

**What happened instead**:

It selected an off station role, a theta station member

Ghost roles, ERT, and admin stuff off the station should not be random targets

Spawn a traitor.
Spawn a ghost role.
Give a traitor a random assassinate target through traitor panel.
Watch as it can the ghost role.

### LINKED ISSUE #15267: Cult doors are hackable
**Issue Description**:

Cult doors can be hacked, instead of stuning the user that screwdrivers them

to stun the user, since it should not allow normal crew to interact with them

**What happened instead**:

Securityy officers could hack the cult door open

Breaks the point of the doors

SabreML:
## What Does This PR Do

1. Closes #14483 *(Fixed already, but the issue is still open.)*
2. Removed Cultists from the list of possible sacrifice targets.
3. Stopped cultists teleporting into nullspace.
4. Removed some unnecessary `\improper`s from area names, to make the Cult summoning locations more gramatically correct.
5. Stopped cultists drawing more than one rune on a single tile.
6. Fixed cultists who are deconverted after 'Rising' not getting their old eye colour back.
7. Fixed deconverted cultists still getting objective notifications.
8. Fixed roundstart cultists not getting the 'servecult' objective.
9. Fixed summoned ghosts all being set to male.
10. Fixed #15267
11. Fixed deconverting a cultist with Prayer Beads not stripping them of their cult gear.
12. Fixed all holy constructs getting the ability to summon soulstones, when it *should* only be artificers.
13. Fixed holy constructs being convertable to the cult.
14. Fixed holy wraiths getting their light colour reset back to cult colours after jaunting.
15. Fixed Nar'Sie constructs not getting a slight glow around their highlights.
16. Fixed upgrading a grab while teleporting bringing the victim with you. (For all antags with instant teleports)
17. Fixed `Runtime in outfit.dm,114: bad number`.
18. Updated the cultist corpse loot spawned by dead Legions.
19. Fixed the phasing effect for the teleport spell being tied to the user even when teleporting someone else.
20. Fixed `Runtime in cult_items.dm,355: pick() from empty list`.
21. Fixed Veil Shifters not retaining any grabs after teleporting.

## Why It's Good For The Game

1. The ERT can go back to Z2, making it impossible to get to them.
2. It's not actually possible to sacrifice a cultist.
3. Nullspace bad.
4. Grammar good.
5. Bugs bad.
6. Bugs bad.
7. Could potentially be used to metagame.
8. Not sure how this didn't get noticed until now.
9. Consistency is good.
10. Shouldn't be able to hack a magic door.
11. No non-cultists with flagellant's robes.
12. Only artificers should be able to make soulstones.
13. Can very easily cause confusion.
14. Can very easily cause confusion.
15. The glow was in a separate icon state, for whatever reason.
16. Pretty easily abusable.
17. Runtime.
18. Maybe a bit easier to think there's a lavaland cult.
19. Looks pretty confusing I guess.
20. Runtime.
21. It's *intended* as one of the big selling points of veil shifters.

## Changelog
:cl:
tweak: Cultists can no longer be chosen as a sacrifice objective.
tweak: Cult airlocks can no longer be hacked.
tweak: Updated the loot of the dead cultist spawned by Legions.
fix: Roundstart cultists now actually get given the 'servecult' objective.
fix: Deconverting a cultist with Prayer Beads now properly strips them of their gear.
fix: Fixed deconverted cultists keeping the 'servecult' objective.
fix: Fixed cultists teleporting into nullspace if the target rune was deleted.
fix: Cultists can no longer draw more than one rune on the same tile.
fix: The space/lavaland portal effect on Teleport runes now gets deleted along with the rune itself.
fix: Cultists with red eyes being deconverted now get their old eye colour back.
fix: Fixed summoned cult ghosts all being male.
fix: Fixed *all* holy constructs getting the 'Summon Soulstone' ability.
fix: Fixed holy constructs being convertable to the cult.
fix: Fixed holy wraiths getting their light colour changed to red after jaunting.
fix: Fixed Nar'Sie constructs not getting the correct glow overlay.
fix: Fixed upgrading a grab while teleporting bringing the victim with you.
fix: Fixed a runtime caused by dead Legions spawning clowns.
fix: Fixed the phasing effect for the teleport spell being tied to the user even when teleporting someone else.
fix: Fixed teleporting with a Veil Shifter releasing any pulled objects, and also a runtime.
spellcheck: Made some area names more gramatically correct.
/:cl:


code/game/gamemodes/cult/blood_magic.dm
```diff
@@ -914,3 +881,36 @@
 					to_chat(user, "<span class='warning'>You need a free hand for this rite!</span>")
 					uses += BLOOD_RECHARGE_COST // Refund the charges
 					qdel(rite)
+
+		if("Blood Spear (150)")
```
moxian, line 885: These are ctrl+x -> ctrl+v with no changes to the content, right?

code/game/gamemodes/cult/runes.dm
```diff
@@ -452,7 +452,7 @@ structure_check() searches for nearby cultist structures required for the invoca
 
 	var/input_rune_key = input(user, "Choose a rune to teleport to.", "Rune to Teleport to") as null|anything in potential_runes //we know what key they picked
 	var/obj/effect/rune/teleport/actual_selected_rune = potential_runes[input_rune_key] //what rune does that key correspond to?
-	if(!src || !Adjacent(user) || QDELETED(src) || user.incapacitated() || !actual_selected_rune)
+	if(!src || QDELETED(src) || !actual_selected_rune || QDELETED(actual_selected_rune) ||!Adjacent(user) || user.incapacitated())
```
moxian, line 455: split the infrastructural "is this thing qdelled?" and the business logic "is this human capable of the task?" into two checks maybe?

code/modules/response_team/ert.dm
```diff
@@ -177,6 +177,7 @@ GLOBAL_VAR_INIT(ert_request_answered, FALSE)
 	M.mind.original = M
 	M.mind.assigned_role = SPECIAL_ROLE_ERT
 	M.mind.special_role = SPECIAL_ROLE_ERT
+	M.mind.offstation_role = TRUE
```
moxian, line 180: I don't think this is correct.
Offstation roles are things like dionae in terrarium, or syndicate agents in the syndicate lava base, who are *physically* unreachable (barring extraordinary effort).
ERT, however, are very much reachable, and are *on*station most of the time.
Setting this to TRUE just so that you can treat them the same in cult objective selection feels extremely hacky. Please don't. Better just check that special_role != SPECIAL_ROLE_ERT in the cult objective selector.

(And from the design perspective: I don't even think ERT members being sacrifice targets is a bad thing. Yes, they *can* get to z2, but they usually don't until the cult is dealt with. And if they do go to z2, they can take captain, NTR, or other "legit" sacrifice objective with them, so it doesn't even address the problem fully. But in either case, I'm more concerned about this from the code standpoint.)

code/game/gamemodes/cult/blood_magic.dm
SabreML: I removed a couple of unneccesary backslashes, but other than that it's all the same.

code/modules/response_team/ert.dm
SabreML: >I don't think this is correct.
Offstation roles are things like dionae in terrarium, or syndicate agents in the syndicate lava base, who are physically unreachable (barring extraordinary effort).
ERT, however, are very much reachable, and are onstation most of the time.
Setting this to TRUE just so that you can treat them the same in cult objective selection feels extremely hacky. Please don't. Better just check that special_role != SPECIAL_ROLE_ERT in the cult objective selector.

My main reason for not doing as you suggested was the description for the variable in `mind.dm`:
```cs
var/offstation_role = FALSE //set to true for ERT, deathsquad, abductors, etc, that can go from and to z2 at will and shouldn't be antag targets
```
It seems to me like it's intended for exactly this use, but someone just forgot to add ERT to the list.
Also, I feel like having a check specifically for the ERT role is a bit too snowflakey; Considering the above variable exists and does the same thing.
<hr>

>(And from the design perspective: I don't even think ERT members being sacrifice targets is a bad thing. Yes, they can get to z2, but they usually don't until the cult is dealt with. And if they do go to z2, they can take captain, NTR, or other "legit" sacrifice objective with them, so it doesn't even address the problem fully. But in either case, I'm more concerned about this from the code standpoint.)

I agree that it'd be good to have ERT members as targets for antags, but I've heard of two instances in the last month alone of the ERT leaving mid-round and the cult being stuck until an admin notices or someone ahelps.
While it's a lot more likely that they'll stay until the shuttle is called, it's definitely not certain.

code/game/gamemodes/cult/runes.dm
SabreML: It'd definitely be more readable if they were split up, but there would need to be the same thing copy-pasted twice for it to work.
e.g.
```dm
if(QDELETED(stuff))
	fail_invoke()
	return

if(!can_use_or_whatever(user))
	fail_invoke()
	return
```
Unless you can think of a better way of doing this, I think it would be better to just leave it as-is.

moxian: I think two `fail_invoke()` calls are fine like that.

code/modules/response_team/ert.dm
moxian: Setting it because the comment says so is fair.
I'm retracting my concern.

code/game/gamemodes/cult/runes.dm
farie82, line 455:
```suggestion
	if(QDELETED(src) || QDELETED(actual_selected_rune) || !Adjacent(user) || user.incapacitated())
```
`!src` is checked in `QDELETED` same for `actual_selected_rune`

farie82 (changes requested):

code/game/gamemodes/cult/runes.dm
SabreML: Oh, that's actually really good to know.

code/game/gamemodes/cult/blood_magic.dm
```diff
@@ -490,7 +490,7 @@
 	var/mob/living/L = target
 	var/input_rune_key = input(user, "Choose a rune to teleport to.", "Rune to Teleport to") as null|anything in potential_runes //we know what key they picked
 	var/obj/effect/rune/teleport/actual_selected_rune = potential_runes[input_rune_key] //what rune does that key correspond to?
-	if(!src || QDELETED(src) || !user || user.l_hand != src && user.r_hand != src || user.incapacitated() || !actual_selected_rune)
+	if(QDELETED(src) || !user || user.l_hand != src && user.r_hand != src || user.incapacitated() || !actual_selected_rune)
```
farie82, line 493:
```suggestion
	if(QDELETED(src) || QDELETED(actual_selected_rune) || !user || user.l_hand != src && user.r_hand != src || user.incapacitated())
```
The rune might not yet be null but can already be qdeleted here. Can try and repro this by opening the TP window. Removing the dest rune and then selecting said rune.
Could do the same for user but the check after the `!user` should catch that already

SabreML: I can't see the problem problem here.
The rune being deleted before you selected it was fixed by adding the `QDELETED()` check, so it'll just make the rune fail now.

SteelSlayer (approved):

code/game/gamemodes/cult/blood_magic.dm
```diff
@@ -249,7 +249,7 @@
 /datum/action/innate/cult/blood_spell/horror/Destroy()
 	var/obj/effect/proc_holder/horror/destroy = PH
 	. = ..()
-	if(destroy  && !QDELETED(destroy))
+	if(destroy && !QDELETED(destroy))
```
dearmochi, line 252:
```suggestion
	if(!QDELETED(destroy))
```

dearmochi (approved):

moxian:
Each time you add a new fix to this great bundle of fixes, the merge date gets delayed 2 weeks further into the future

SabreML:
Since everyone else is doing it

code/game/area/ss13_areas.dm
```diff
@@ -1108,7 +1108,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 	sound_environment = SOUND_AREA_SMALL_ENCLOSED
 
 /area/engine/equipmentstorage
-	name = "\improper Engineering Equipment Storage"
+	name = "Engineering Equipment Storage"
```
AffectedArc07, line 1111: I am worried with the amount of extra side effects this will cause.

SabreML: It'll almost definitely have side effects, but all of them *should* be good ones since it's more gramatically correct now.

AffectedArc07 (approved):
Testmerged for 12 hours with no issues

PR #15167, opened 2020-12-21
Firedoors open/close even for air alarms that don't report to computers

Kyep:
## What Does This PR Do
When a silent (report_danger_level=FALSE, IE /obj/machinery/alarm/monitor or /obj/machinery/alarm/syndicate) air alarm changes from "green" status to a warning status (orange/red), or the other way around, it will now close/open firedoors in its area.
In practice, what this means is that firelocks in space ruins and other areas that shouldn't show up on station alerts console will still work. 
These air alarms still won't generate alerts to consoles, but they will at the very least open/close firedoors in their area. So they won't be totally useless.

## Why It's Good For The Game
Currently, firelocks in these areas don't open/close based on their air alarm, making them pretty useless.
With this PR, they'll actually do what they're meant to do.

## Changelog
:cl:
tweak: firelocks in offstation areas with air alarms will now open/close when the air alarm signals the air is safe/unsafe.
/:cl:


code/game/machinery/alarm.dm
```diff
@@ -536,6 +536,12 @@
 
 /obj/machinery/alarm/proc/post_alert(alert_level)
 	if(!report_danger_level)
+		// Don't report the level to computers, but do toggle firedoors
+		var/area/A = get_area(src)
+		if(alert_level == ATMOS_ALARM_NONE && A.air_doors_activated)
+			A.air_doors_open()
+		else if(alert_level != ATMOS_ALARM_NONE && !A.air_doors_activated)
```
farie82, line 543:
```suggestion
		else if(alert_level != ATMOS_ALARM_NONE)
```
Same as above

code/game/machinery/alarm.dm
```diff
@@ -536,6 +536,12 @@
 
 /obj/machinery/alarm/proc/post_alert(alert_level)
 	if(!report_danger_level)
+		// Don't report the level to computers, but do toggle firedoors
+		var/area/A = get_area(src)
+		if(alert_level == ATMOS_ALARM_NONE && A.air_doors_activated)
```
farie82, line 541:
```suggestion
		if(alert_level == ATMOS_ALARM_NONE)
```
`air_doors_activated` is already checked in the `air_doors_open` proc itself

farie82 (changes requested):

farie82 (approved):

Dumbdumn5 (approved):
Makes sense, yeah 👍  Don't see a reason why firelocks shouldn't work like that.

AffectedArc07 (approved):

AffectedArc07:
Please resolve merge conflicts when you can

Kyep:
@AffectedArc07 Said it had a conflict, but apparently it did not. Updated.

PR #15170, opened 2020-12-22
Makes the petcollar not an accessory. Refactor of #15155. Revert of the revert

farie82:
## What Does This PR Do
Makes the petcollar not an accessory and removes the accessory blacklist when trying to attach accessories to jumpsuits from #15155

Alternative to the inevitable revert of the revert.

## Why It's Good For The Game
Removes technical debt. Other part I don't care much about but it'll happen anyway

## Changelog
<strike>Nothing here. Unless the revert goes through</strike>
🆑
del: You can no longer put pet collars on jumpsuits
/🆑


SteelSlayer:
image

Dmi's changed but no icon diff bot changes for some reason. Did you just move the sprite without changing it?

farie82:
> image
> 
> Dmi's changed but no icon diff bot changes for some reason. Did you just move the sprite without changing it?

Yep yep only moved it

farie82:
Update the PR to reflect that it will remove the ability to use the collar as an accessory

code/game/objects/items/misc.dm
```diff
@@ -22,3 +23,77 @@
...
+
+/obj/item/petcollar/attackby(obj/item/card/id/W, mob/user, params)
+	if(!istype(W))
+		return ..()
+	if(access_id)
+		to_chat(user, "<span class='warning'>There is already \a [access_id] clipped onto \the [src]</span>")
+	user.drop_item()
+	W.forceMove(src)
+	access_id = W
+	to_chat(user, "<span class='notice'>\The [W] clips onto \the [src] snugly.</span>")
```
SabreML, line 68:
```suggestion
	to_chat(user, "<span class='notice'>[W] clips onto [src] snugly.</span>")
```

code/game/objects/items/misc.dm
```diff
@@ -22,3 +23,77 @@
...
+		START_PROCESSING(SSobj, src)
+
+/obj/item/petcollar/dropped(mob/living/simple_animal/user)
+	STOP_PROCESSING(SSobj, src)
+
+/obj/item/petcollar/process()
+	var/mob/living/simple_animal/M = loc
+	// if it wasn't intentionally unequipped but isn't being worn, possibly gibbed
+	if(istype(M) && src == M.pcollar && M.stat != DEAD)
+		return
```
SabreML, line 89: Feels like this would be much better using a death signal of some sort rather than processing, but that's something for another PR.

code/game/objects/items/misc.dm
```diff
@@ -22,3 +23,77 @@
...
+
+	var/area/t = get_area(M)
+	var/obj/item/radio/headset/a = new /obj/item/radio/headset(src)
+	if(istype(t, /area/syndicate_mothership) || istype(t, /area/shuttle/syndicate_elite))
+		//give the syndicats a bit of stealth
+		a.autosay("[M] has been vandalized in Space!", "[M]'s Death Alarm")
+	else
+		a.autosay("[M] has been vandalized in [t.name]!", "[M]'s Death Alarm")
+	qdel(a)
```
SabreML, line 98: Variable names are a little odd here.
```suggestion
	var/area/A = get_area(M)
	var/obj/item/radio/headset/H = new /obj/item/radio/headset(src)
	if(istype(A, /area/syndicate_mothership) || istype(A, /area/shuttle/syndicate_elite))
		//give the syndicats a bit of stealth
		H.autosay("[M] has been vandalized in Space!", "[M]'s Death Alarm")
	else
		H.autosay("[M] has been vandalized in [A.name]!", "[M]'s Death Alarm")
	qdel(H)
```

code/game/objects/items/misc.dm
```diff
@@ -22,3 +23,77 @@
...
+	STOP_PROCESSING(SSobj, src)
```
SabreML, line 99:
```suggestion
	else
		a.autosay("[M] has been vandalized in [t.name]!", "[M]'s Death Alarm")
	qdel(a)
	return PROCESS_KILL
```

code/game/objects/items/misc.dm
SabreML, line 98: *Radio code.*

code/game/objects/items/misc.dm
```diff
@@ -22,3 +23,77 @@
...
+	access_id = W
+	to_chat(user, "<span class='notice'>\The [W] clips onto \the [src] snugly.</span>")
+
+/obj/item/petcollar/GetAccess()
+	return access_id ? access_id.GetAccess() : ..()
+
+/obj/item/petcollar/examine(mob/user)
+	. = ..()
+	if(access_id)
+		. += "There is [bicon(access_id)] \a [access_id] clipped onto it."
```
SabreML, line 76: Oh, also here too for the `\an` thing:
```suggestion
		. += "There is [bicon(access_id)] \an [access_id] clipped onto it."
```

SabreML (commented):
I know all of this was there already, but may as well clean up a bit while you're moving it.

farie82:
> I know all of this was there already, but may as well clean up a bit while you're moving it.

Danke for the review. I'll see about resolving your comments this weekend

code/game/objects/items/misc.dm
farie82: Aye, indeed that would be a bit too much to add on.

code/game/objects/items/misc.dm
farie82: Yep it ain't great at all this.... luckily it is only done once though

farie82:
Bit later than I wanted but I fixed the review stuff. Good finds!

farie82:
Gonna close this one to reduce my PRs open. If people want to reopen it then go-ahead

PR #15177, opened 2020-12-23
Remove the laser pointer from the mining vending machine

Varghh:
## What Does This PR Do
Removes the laser pointer from the mining vendor and add it to the robotech vending machine.

## Why It's Good For The Game
We're already nuking borgs with many nerfs over the years, we all know sec borgs are a problem and laser pointers are the cheesiest way to deal with them, a range one click stun is too strong and miners should not be able to buy a shit ton of them.
Also, why do miners need a laser pointer? They already have a lot of antag toys from lavaland and, let's be real, It makes no sense for them to be able to buy them.

## Changelog
:cl:
del: Miners can no longer buy the laser pointer from the mining vending machine
add: You can now acquire more laser pointers from the robotech vending machine
/:cl:


hal9000PR:
Maybe add them in as a cargo crate or a prize from the arcade instead? I am fine with removing them from miners but I would rather them not be a finite resource, also more cool cargo crates is always a welcome addition.

Varghh:
> Maybe add them in as a cargo crate or a prize from the arcade instead? I am fine with removing them from miners but I would rather them not be a finite resource, also more cool cargo crates is always a welcome addition.

I really like the idea of the arcade

Varghh:
Added them to the arcade prize machine at a high price, waiting for more discussion and maintainer to see how much should they cost (or where to add them, maybe it should be a cargo crate).

moxian:
Echoing what @datlo said on discord: if you need to have high price for it in the arcade (and 2000 IS high), then it's a just a powergamer tax, and then why add it to arcade in the first place?
Arcade has always been fluff items that have next to no gameplay impact (except for the bike that's literally unobtainable). Altering that is a very significant change, and one I'm personally very against.

Just add it to cargo/protolathe/autolathe/maint loot.

Or add it to arcade, but remove the stun/blind/gameplay significance, and make it, do the "point to" verb or something.

Spacemanspark:
I think it would be better to just remove the stun functionality of the item itself than anything else.

Varghh:
I could totally see Laser pointer getting removed from the mining vendor since they have no reason to be there, getting his price lowered at the arcade prize machine and remove the stun functionality for borgs.
Also, like @moxian pointed out on discord, they shred Vulps eyes because of the eye damage buff, maybe we could get rid of the eye damage functionality too.

Denthamos:
What if you keep the laser pointer as is but throw it into the robco vendor which is normally not on station, but can be ordered in through cargo?  It has a bunch of random miscellaneous tech pieces in it which seems like it may fit in with.

Varghh:
Going with @Denthamos idea, laser pointers are now inside the robotech vending machine, you can order those from cargo

Kyep (approved):

PR #15185, opened 2020-12-24
Removes the Lesser Drake form from dragons blood

hal9000PR:
## What Does This PR Do
This PR removes the chance for miners to get the ability to transform into a lesser ash drake by drinking dragons blood.

## Why It's Good For The Game
Lesser Ash drake form is extremely powerful, outside of shotguns, you cannot combat it as sec. it has a powerful ranged attack dealing a large amount of burn damage, igniting and it can hit multiple targets. It has a lot of melee damage that is armour piercing. It is slow, however it has a swoop attack which deals a MASSIVE amount of brute damage and making lava tiles underneath it, dealing even more damage and closing the distance for melee attacks. Then, if the target of a swoop or melee attack, is dead/asleep it will gib them, and if a swoop gibbed them the lava will likely burn their brain, removing them from the round permanently. Also, it makes miners practically unstunnable because as soon as they go down and sec comes over to cuff them, they will morph into a giant dragon and eat them. Finally, lesser drakes can cause a lot of confusion on station, "ASH DRAKE ON STATION", "SOMEONE GET CLOSE TO IT TO FIND OUT IF IT IS A MINER"

## Changelog
:cl:
del: Removed lesser ash drake form from mining loot
tweak: also cleaned up the code
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/loot/ashdragon_loot.dm
+++ b/code/modules/mining/lavaland/loot/ashdragon_loot.dm
@@ -119,22 +119,17 @@
 		return
 
 	var/mob/living/carbon/human/H = user
-	var/random = rand(1,3)
+	var/random = rand(1, 2)
 
 	switch(random)
 		if(1)
 			to_chat(user, "<span class='danger'>Your flesh begins to melt! Miraculously, you seem fine otherwise.</span>")
 			H.set_species(/datum/species/skeleton)
 		if(2)
-			to_chat(user, "<span class='danger'>Power courses through you! You can now shift your form at will.")
-			if(user.mind)
-				var/obj/effect/proc_holder/spell/targeted/shapeshift/dragon/D = new
-				user.mind.AddSpell(D)
-		if(3)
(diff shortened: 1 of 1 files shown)


Shockpoint:
Y'know, I'm not really a fan of this form myself, but I'd prefer a balance change on the form itself rather than full removal.

I think at least swoop could go from lesser drake.

Kyep (approved):

Denthamos:
What if it worked like a transformation ability that required an uninterrupted charge up to perform and getting stunned/pushed would interrupt the charge up on it, meanwhile a noticable text would appear in the chat box that people could see and react to.  And likewise being stunned would prevent it from happening.

AffectedArc07:
I would much rather this be tweaked instead of straight up removed. Not in favour.

hal9000PR:
I have very little coding knowledge but I will try and work on tweaking it instead. I'm gonna take into consideration what shockpoint and denthamos said. If anyone else has any suggestions please post them here!

PR #15186, opened 2020-12-26
Changes some world init order stuff around

AffectedArc07:
## What Does This PR Do
Currently, people can join the world in the split second before early inits are done (Specifically SQL queries), and end up attempting to reference a species which doesnt exist because `GLOB.all_species` hasnt been loaded yet. This fixes that.

Note that this is not the confirmed reason people are losing characters, but it adds up with code flow and with runtimes
```
[2020-12-26T17:55:38] Runtime in preferences_setup.dm,10: Cannot read null.bodyflags
   proc name: random character (/datum/preferences/proc/random_character)
   src: /datum/preferences (/datum/preferences)
   call stack:
   /datum/preferences (/datum/preferences): random character(null)
   /datum/preferences (/datum/preferences): New([redacted] (/client))
   [redacted] (/client): New(null)
```
`null.bodyflags` is meant to be a reference to `species.bodyflags`, which is looked up in `GLOB.species`

## Why It's Good For The Game
People losing characters is bad

## Changelog
:cl: AffectedArc07
fix: Moved some world init stuff round.
/:cl:

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -13,6 +13,9 @@ GLOBAL_LIST_INIT(map_transition_config, MAP_TRANSITION_CONFIG)
 	// Setup all log paths and stamp them with startups, including round IDs
 	SetupLogs()
 
+	// This needs to happen early, otherwise people can get a null species, nuking their character
+	makeDatumRefLists()
+
 	TgsNew(new /datum/tgs_event_handler/impl, TGS_SECURITY_TRUSTED) // creates a new TGS object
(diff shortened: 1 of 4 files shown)


Kyep:
Thanks for this, I think I've seen this error myself in testing, and it was annoying me.

AffectedArc07:
More tweaks have been made:
- Nothing in `world/New()` sleeps now so it should all happen within the same tick
- A note has been added to warn people what will happen if they do make things sleep

farie82 (approved):

PR #15192, opened 2020-12-27
Fixes some overlay screwery on death and revival

### LINKED ISSUE #15190: Vulpkanin darkvision dies with them
**Issue Description**:
When a Vulpkanin dies and gets revived, they lose the ability to see in the dark.

Uhh it's a racial trait so, they should have it all the time?

**What happened instead**:
After I was revived, unlit areas were super dark.

Because, that's, a racial trait they should always have. These kind of are repetitive questions.

Died
Stayed in body, never flying around as ghost.
Was revived and healed.

I try to not die, so I'm not sure how often this happens.

**Extra information**:
Love you all and you do wonders and it's real good stuff.
This is a picture of me in maints, with a light source at the top to show the darkness.

dreamseeker_adG10Qytxl

dearmochi:
## What Does This PR Do

- Toggles overlays off/on on death/revival instead of deleting them
  - Fixes #15190
- Restores unconscious overlay when revived

## Why It's Good For The Game

- Fixes a lot of issues regarding overlays, such as Vulpkanin night-vision/welding helmet overlay vanishing on death, the latter of which can be exploited.
- Fixes issue where the unconscious overlay wouldn't appear when defibrillated, even though the patient is unconscious, improving experience

## Changelog
:cl:
fix: Fix overlays (e.g Vulpkanin night-vision, unconscious, welding helmet) not re-appearing when revived
/:cl:


code/modules/mob/living/carbon/human/death.dm
```diff
@@ -108,7 +108,7 @@
 	if(SSticker && SSticker.mode)
 		SSblackbox.ReportDeath(src)
 
-/mob/living/carbon/human/update_revive()
+/mob/living/carbon/human/update_revive(updating)
```
dearmochi, line 111: Unsure if the argument needs to be written explicitly here to be passed to the parent, but I put it in for good measure

AffectedArc07: Dreamchecker can throw a fit sometimes

AffectedArc07 (approved):

PR #15193, opened 2020-12-27
Makes mindread not list non carbon mobs. No more reading the mind of ghosts or AI eyes

### LINKED ISSUE #15181: Genetics mind read ability allow you to know if AI eyes or ghosts or any mob is nearbye
**Issue Description**:

Genetics mind read ability targets mob in range, instead of mobs the user can see

For the user to read minds of living people they can see

**What happened instead**:

Ghosts, ai eyes, probably demons and revenants

You should NOT be able to do this

Get mind read ability
Go where ghosts are
See what ghosts are nearbye, and other fun mobs

Ever since it was bloody added

farie82:
## What Does This PR Do
Mind read (empathic tought) now only gives you the option between things you can mind read. So no more ghosts or AI eyes.

Fixes #15181

## Why It's Good For The Game
Bug fix bug fix

## Changelog
:cl:
fix: Makes mindread not list non carbon mobs. No more reading the mind of ghosts or AI eyes
/:cl:

### DIFF
--- a/code/game/dna/genes/goon_powers.dm
+++ b/code/game/dna/genes/goon_powers.dm
@@ -520,9 +520,9 @@
 	name = "Read Mind"
 	desc = "Read the minds of others for information."
 	charge_max = 180
-	clothes_req = 0
-	human_req = 1
-	stat_allowed = 0
+	clothes_req = FALSE
+	human_req = TRUE
+	stat_allowed = CONSCIOUS
 	invocation_type = "none"
 	range = -2
 	selection_type = "range"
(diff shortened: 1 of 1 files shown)


moxian:
Do you maybe want it to list `/mob/living`?  Neither ghosts nor AI eye are even `/mob/living` so it should fix the issue, but that would still allow you to read Ian's mind

farie82:
> Do you maybe want it to list `/mob/living` (and not more specific `/mob/living/carbon`)? Neither ghosts nor AI eye are even `/mob/living` so it should fix the issue, but that would still allow you to read Ian's mind

You were not able to read Ians mind before hand. A bit later in the code in the `cast` it will loop over the targets (single mob) using the `/mob/living/carbon` filter. So it won't find any targets to mind read.
I'm planning on keeping the functionality the same in this PR since it's only a fix PR

AffectedArc07 (approved):

PR #15195, opened 2020-12-27
Makes JMP usable by people other than admins if they are already a ghost

farie82:
## What Does This PR Do
Makes the admin jump (JMP) usable by other staff members if they are already a ghost/observer.
If they are not ghosted then they still need the ADMIN permission to use it.

## Why It's Good For The Game
It can be a bit annoying to not be able to use the JMP while you're already a ghost. All the while you can already do it 99/100 times using the ghost verbs which take more time

## Changelog
:cl:
tweak: The admin JMP href is now usable by all staff members who are ghosted. If they are not ghosted they will still require the admin permission.
/:cl:

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -1806,14 +1806,16 @@
 		log_admin("Admin [key_name_admin(usr)] has unlocked the Cult's ability to summon Nar'Sie.")
 
 	else if(href_list["adminplayerobservecoodjump"])
-		if(!check_rights(R_ADMIN))	return
+		var/client/C = usr.client
+		if(!isobserver(usr))
+			if(!check_rights(R_ADMIN)) // Need to be admin to aghost
+				return
+			C.admin_ghost()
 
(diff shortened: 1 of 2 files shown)


code/modules/admin/topic.dm
```diff
@@ -1806,14 +1806,16 @@
...
+				return
+			C.admin_ghost()
 
 		var/x = text2num(href_list["X"])
 		var/y = text2num(href_list["Y"])
 		var/z = text2num(href_list["Z"])
 
-		var/client/C = usr.client
-		if(!isobserver(usr))	C.admin_ghost()
 		sleep(2)
```
AffectedArc07, line 1819: Why is this `sleep(2)` here

farie82: Not 100% sure but my guess is to wait for the spawns etc done in `admin_ghost`. Haven't checked though

Kyep (approved):

AffectedArc07 (approved):

Issue #14955, opened 2020-11-21
You can make multiple DNA Vaults.


Shockpoint:
**Issue Description**:
You can make multiple DNA vaults.

**What did you expect to happen**: 
To only have one vault.

**What happened instead**:
There are FOUR.

**Why is this bad/What are the consequences**:
Multiple DNA vaults means multiple free genes without instabilities.

**Steps to reproduce the problem**:
Have the DNA Vault as a station goal, order multiple crates.

**When did the problem start happening**:
I have no idea.

**Extra information**:

S34NW:
I Think this is intended, it gives people something to do on a slow shift!

Muhdi345:
It's not so intended when you can use the same samplers you used for the first DNA vault to instantly be able to choose the next 2 free powers though.

S34NW:
Ah if you can instantly reuse samplers then yeah

PR #15197, opened 2020-12-27
Get max 1 power from the DNA vaults

farie82:
## What Does This PR Do
Limits the amount of powers you get from multiple DNA vaults to 1.
Adds UI to notify the user that they have already selected a power

Fixes: #14955

## Why It's Good For The Game
The DNA vault powers are "balanced" in a way that a player can only really have one max in a round. Having more will break balance big time for antags and security alike. However fun it could be

## Images of changes
image

## Changelog
:cl:
tweak: A player can have up to 1 power from the DNA vaults now. Making more will not add powers to choose from
/:cl:


PacifistDalek:
Does multiple dna vaults still give you more options to choose from? So you can only get one power, but putting in the effort to build another vault will give more to choose from for that single power?

farie82:
> Does multiple dna vaults still give you more options to choose from? So you can only get one power, but putting in the effort to build another vault will give more to choose from for that single power?

It won't. You get one set of powers to pick from per round

Butlerblock:
"The DNA vault powers are "balanced" in having one max" DNA vaults aren't balanced regardless so honestly, the powers themselves should be balanced before anything else because they're EXTREMELY overtuned.

Kyep (approved):

AffectedArc07 (approved):
An end to this meme

PR #15203, opened 2020-12-28
Refactors the events to be async (not use sleeps) and not use spawns

farie82:
A TM or two would be appreciated due to the impact

## What Does This PR Do
Refactors/reworks most of the events so they don't use spawns and sleeps. If they do indirectly then they will use a timer or `INVOKE_ASYNC`.
Changed some stuff around to make it better code wise. These changes will be listed in the CL.

Original PR reason:
I've not been able to repro the issue locally. But from analysis it seems that the sleeps in the opening of the doors might cause the issue. This can cause the SSevent to hang. Causing it to go on pause and then it will be "resumed" next tick with another process call. All the while the other process call is still running.
If the event announce and end time are the same (random chance) then it can cause multiple announcements at once due to this. Since it will announce, open doors, hang. And then another call does the same.

The async call should prevent this since it will not wait for the sleeps.

## Why It's Good For The Game
Will fix some issues with events hanging the event SS and causing multiple process calls to be alive at once.

## Changelog
:cl:
fix: Stops the prison break event from announcing itself 100 times (I hope)
tweak: Meaty ores will now not turn invisible once they hit their first target
tweak: Meaty ores will now not stop moving when they hit a mob
fix: Meaty ores now won't spawn mobs in the void under rare edge cases
/:cl:


AffectedArc07 (changes requested):
Fix this properly at an event level by making all events only fire off their `start()` once
`code\modules\events\event.dm#L140`. 
```dm
/datum/event/process()
	if(!processing)
		return

	if(activeFor > startWhen && activeFor < endWhen || noAutoEnd)
		tick()

	if(activeFor == startWhen)
		// Add a check to see if were already running here
		isRunning = TRUE
		start()
```
Currently any event which sleeps (mob spawns, this, etc) is bandaided against this mess, because someone forgot sanity checks. Fix it properly please.

farie82:
> Fix this properly at an event level by making all events only fire off their `start()` once
> `code\modules\events\event.dm#L140`.
> 
> ```dm
> /datum/event/process()
> 	if(!processing)
> 		return
> 
> 	if(activeFor > startWhen && activeFor < endWhen || noAutoEnd)
> 		tick()
> 
> 	if(activeFor == startWhen)
> 		// Add a check to see if were already running here
> 		isRunning = TRUE
> 		start()
> ```
> 
> Currently any event which sleeps (mob spawns, this, etc) is bandaided against this mess, because someone forgot sanity checks. Fix it properly please.

This will cause other issues as well. Ticks will not be done every X seconds which can cause other issues. I'd rather prefer to add some documentation saying to avoid sleeps and use async calls if you do call a proc with sleep.
Adding a bandaid will also promote "lazy" coding. I'd much rather have people be aware of what sleep does if they code events

AffectedArc07:
> Adding a bandaid will also promote "lazy" coding. I'd much rather have people be aware of what sleep does if they code events

This isn't a bandaid. Its a legitimate fix for sloppy code from years and years ago. Events aren't meant to start multiple times. Please actually do what I have asked instead of avoiding the problem.

code/modules/events/event.dm
```diff
@@ -91,6 +91,7 @@
   *
   * Allows you to start before announcing or vice versa.
   * Only called once.
+  * Ensure no sleep is called. Use INVOKE_ASYNC to call procs which do.
```
AffectedArc07, line 94: It literally says here `Only called once`.

farie82: image
Any of the calls there can be called multiple times if the first processing call gets put to sleep long enough for another one to spawn before line 167 is hit

AffectedArc07: Then make the code actually work properly, and put it in a correct order, such as

`Start > Announce > Tick > End > Kill`

farie82: It doesn't have to be done in that order is the thing. The announce can happen before the start. The end can happen before the announce and such. Start should happen before the end ofcourse

AffectedArc07: You are still trying to beat around the point of refusing to add actual protection into avoiding things starting twice, by forcing every event proc to be invoked async, which is not a good idea. BYOND is a synchronous language. At this point you may as well just invoke the actual start async instead of having all procs inside have to be snowflaked. Or just add an actual check to make sure it doesnt start 3 times, like I have asked you to do 3 times now.

farie82: I've listed my reasoning to not do that earlier. It will make ticks between events be irregular. One time the time between can be 2 seconds the other time 6 which makes the system unpredictable. Having the system work like in the PR will not have that effect and instead the event will just continue on as expected. Feel free to make the PR yourself to change it the way you want.

AffectedArc07:
After further analysis, I am reinforcing my point that these should **not** be farmed off asynchronously. 
```dm
/datum/controller/subsystem/events/fire()
	for(var/datum/event/E in active_events)
		E.process()

	for(var/i = EVENT_LEVEL_MUNDANE to EVENT_LEVEL_MAJOR)
		var/datum/event_container/EC = event_containers[i]
		EC.process()
```

Events are processed every 2 seconds, plus or minus some if the MC is heavily overloaded (can happen on highpop). The way current events handle this is **awful**, either by spawning off candidate selection. 
```dm
/datum/event/spider_terror/start()
	// It is necessary to wrap this to avoid the event triggering repeatedly.
	INVOKE_ASYNC(src, .proc/wrappedstart)
```

or manually disabling processing
```dm
/datum/event/blob/start()
	processing = FALSE //so it won't fire again in next tick
```

The actual fault here is that setup is running multiple times, despite the fact it shouldnt. This (and all the other issues with start repeating) can easily be fixed by changing
```dm
	if(activeFor == startWhen)
		isRunning = TRUE
		start()
```

to
```dm
	// Only start once
	if(activeFor == startWhen && !isRunning)
		isRunning = TRUE
		start()
```

This will fix the actual issue at hand, and allow other events to be desnowflaked to not be awful.

code/modules/events/wallrot.dm
```diff
@@ -6,26 +6,31 @@
...
 
-		if(center)
-			// Make sure at least one piece of wall rots!
-			center.rot()
+	// 100 attempts
+	for(var/i in 0 to 100)
+		var/turf/candidate = locate(rand(1, world.maxx), rand(1, world.maxy), 1)
+		if(istype(candidate, /turf/simulated/wall))
+			center = candidate
+			break
```
farie82, line 19: Break here to avoid unneeded computations

code/modules/events/dust.dm
```diff
@@ -64,30 +65,33 @@
 	src.x = startx
 	src.y = starty
 	src.z = level_name_to_num(MAIN_STATION)
-	spawn(0)
-		walk_towards(src, goal, 1)
```
farie82, line 68: Not needed to call this async. The references of DM state it returns instantly and will continue on the background

code/modules/events/event.dm
```diff
@@ -51,7 +51,6 @@
 	return 0*/
 
 /datum/event	//NOTE: Times are measured in master controller ticks!
-	var/processing = 1
```
farie82, line 54: This was a hack to avoid dupe process calls

code/modules/events/dust.dm
```diff
@@ -64,30 +65,33 @@
...
+	if(life <= 0)
+		walk(src, 0)
+		on_shatter(where)
+		qdel(src)
 
-			life--
-			if(life <= 0)
-				walk(src,0)
-				spawn(1)
-					qdel(src)
```
farie82, line 89: Unsure why this was done prior

code/modules/events/dust.dm
```diff
@@ -64,30 +65,33 @@
...
 
 /obj/effect/space_dust/Bump(atom/A)
-	spawn(0)
-		if(prob(50))
-			for(var/mob/M in range(10, src))
-				if(!M.stat && !istype(M, /mob/living/silicon/ai))
-					shake_camera(M, 3, 1)
-		if(A)
-			playsound(src.loc, 'sound/effects/meteorimpact.ogg', 40, 1)
+	if(QDELETED(src))
```
farie82, line 71: Avoids the null space issue

code/modules/events/abductor.dm
```diff
@@ -1,12 +1,12 @@
 /datum/event/abductor
 
 /datum/event/abductor/start()
-	//spawn abductor team
-	processing = 0 //so it won't fire again in next tick
+	INVOKE_ASYNC(src, /datum/event/abductor.proc/try_makeAbductorTeam)
```
AffectedArc07, line 4:
```suggestion
	INVOKE_ASYNC(src, .proc/try_makeAbductorTeam)
```
Pretty sure you can use a relative ref to the proc here, but not 100%. Please try for this one, and if you can, update the rest please.

farie82: Yep that seems to work. Still magic to me at times how byond determines when a relative path works and when it does not

code/modules/events/event.dm
```diff
@@ -91,6 +90,7 @@
   *
   * Allows you to start before announcing or vice versa.
   * Only called once.
+  * Ensure no sleep is called. Use INVOKE_ASYNC to call procs which do.
   */
 /datum/event/proc/start()
 	return
```
AffectedArc07, line 96: Put the `SHOULD_NOT_SLEEP(TRUE)` dreamchecker flags in these procs

AffectedArc07 (changes requested):

code/modules/events/event.dm
farie82: Those will call out false positives. Such as `emote` having a sleep in it when you flip. `emote` being called on `death` which is called on `gib` etc... I couldn't be bothered to fix the false positives

AffectedArc07:
I still feel that the actual async invoke should be here
```dm
/datum/event/process()
	if(activeFor > startWhen && activeFor < endWhen || noAutoEnd)
		tick()

	if(activeFor == startWhen)
		isRunning = TRUE
		start()

	if(activeFor == announceWhen)
		announce()

	if(activeFor == endWhen && !noAutoEnd)
		isRunning = FALSE
		end()

	// Everything is done, let's clean up.
	if(activeFor >= lastProcessAt() && !noAutoEnd)
		kill()

	activeFor++
```
and not inside every single event. Its much cleaner this way and will absolutely make sure no one does anything sleeping by accident.

AffectedArc07 (approved):

PR #15204, opened 2020-12-28
more stealthy fake walls 2.0. Stealth harder!

TDSSS:
## What Does This PR Do
This is a more far-reaching version of #15130 that includes the changes in that other PR as well. To sum up:

- The examination text of false walls now matches that of the corresponding real wall, preventing you from detecting false walls just by examining them.
- False walls can now be 'locked' by using a screwdriver on them while they're closed. A locked false wall _acts like a real wall_ when clicked. IE, it cannot be opened by clicking it, it produces the same sound, etc. It needs to be unlocked by using a screwdriver to unlock it again.
- Unlocking a false wall takes 5 seconds. EDIT: Locking a false wall takes just 1 second. Real walls now also have 'fake' interaction with screwdrivers, taking you 5 seconds to try and unlock them, followed by failing. This means it now takes 5 seconds per wall to check if something is a (hidden) fake wall, rather than a drive-by click while walking past.
- If you don't want to lock a wall, you can leave them unlocked and they will keep acting as they do now, for a quick getaway.
- False walls can no longer be turned into real walls by using a screwdriver on them.
- I added some autodoc documentation to tools acts while I was it, since what their return value means was very non-obvious to me.

## Why It's Good For The Game
Our maintenance being too small for our pop has been brought up a lot. It makes it really hard for antags to build on-station bases without being found. Making false walls harder to locate helps with this. Trying to spam-check every single wall takes a lot longer now, but you can still easily check a wall you suspect of being false with a small individual time-investment.
Being able to tell a false wall from a real one because their examine text doesn't match is, as far as I am concerned, a bug to be fixed.

Being able to instantly turn a false wall into a real wall felt very cheesy to me. The primary advantage of a false wall should be that it's a hidden door, not that it's a door you can instantly close. Still, you can now lock the wall in 1 second while it takes pursuers 5 seconds to get it open again, preserving the same effect.

## Changelog
:cl:
add: Fake walls can now be locked and unlocked with a screwdriver. Locked fake walls will not open when clicked
tweak: fake wall examine text now matches real wall examine text.
del: You can no longer instantly turn fake walls into real walls with a screwdriver.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -360,6 +360,14 @@
 	if(reagents)
 		reagents.temperature_reagents(exposed_temperature)
 
+
+/**
+  * Calls the right tool act.
+  *
+  * Call the appropriate tool act, such as screwdriver_act, crowbar_act and so on.
+  * Returns TRUE if a tool act was found/used.
+  * This means the tool shouldn't call its attack_by proc.
+  */
 /atom/proc/tool_act(mob/living/user, obj/item/I, tool_type)
 	switch(tool_type)
 		if(TOOL_CROWBAR)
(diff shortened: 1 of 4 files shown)


AffectedArc07:
`fix: fake wall examine text now matches real wall examine text.`

This is a tweak. Not a fix.

TDSSS:
>This is a tweak. Not a fix.

Disagree. I am fairly sure that false and real walls should have the same examine text and the fact they don't is an oversight, not a deliberate design decision. But since it's a minor thing (and the PR itself includes unambiguous balance changes anyway), I'll go ahead and change it.

Fox-McCloud:
Personally think the current mechanic of turning false walls back into regular walls should remain---otherwise, it's not really possible to cover your tracks---it's also not really possible, anymore, to have those "clutch" saves where you manage to tighten it right as the security officer checks the wall behind you.

And, of course, the one thing this PR doesn't (and can't without issues) is prevent mesons from detecting false walls.


The examine text is fine though.

datlo:
I strongly disagree with these changes to how screwdriver interacts with false walls. Adding a 5 second delay ( A very, very long time when action is going on) to locking a wall is a massive nerf to false walls, not a buff. This removes quick locking the wall behind you in a chase, quickly hiding your track after you get out of a break in, and does not at all change the meta of hitting every wall with an empty hand to check for a false wall. 

The new 'locked false wall' is basically the same as a regular wall... it doesnt take all that much longer to make a false wall to begin with. 

On top of that, you now HAVE to deconstruct a false wall to turn it back into a regular wall and hide your tracks. Playing a stealthy antag on para is already difficult enough with our pop on box and this would make it harder, not easier.

TDSSS:
@Fox-McCloud  @datlo I've updated the PR, locking a false wall now takes only 1 second while unlocking it takes 5. You can quickly lock a wall behind yourself still this way, but it's no longer instant.

I hope that serves as a good compromise that preserves the old functionality. As for mesons, yeah, not much I can really do about those, I'm afraid.

Carthusia:
Is there no way to prevent mesons from detecting them?

All the locks and timers in the world wont prevent the meson from basicly nullifying any stealth advantage a false wall provides.

datlo:
You'd basically have to refactor false walls to be turfs instead of items, I think. Changing mesons instead would be snowflakey as fuck.

Dumbdumn5:
I like a few of the ideas here with it, but I can't really approve of it in its current state.

Essentially, the locking mechanic proposed doesn't punish somebody using it without thinking by making the wall basically unusable, it just fucks over pursuers, meanwhile, the current state of false walls takes a bit of planning so as to avoid being stuck and there's a bit more to it so you don't get locked into a maint room or otherwise and pursuers have to avoid getting stuck the same way. Also having that unlock check be the case for every wall seems like a hell of a lot more of a punishment than necessary for wallchecking when mesons does it in no time at all. Maybe having another tool allowing locking and unlocking while leaving screwdrivers as the 'turn back into regular wall' thing and leaving the time for both locking and unlocking both at somewhere around 1-3 seconds rather than a 1 and 5 second disparity.

Removing the examine text I can actually respect, but that's... kinda about it.

I like the mesons idea, but I also know it'd probably be a monumental pain in the ass and a snowflake code bit to do, so, for the moment, I like how the system currently works, though it could definitely use some improvements.

AffectedArc07:
This PR has multiple objections from both parties and no one in favour from either party. Closing.

PR #15212, opened 2020-12-30
krumping wizard demonology.

Aquilar:
## What Does This PR Do
 this reduces the number of possible demon summons by wizards and restricts those to the wizard game mode.

## Why It's Good For The Game

While demons are all good fun, having a wizard being able to summon 6 of them and stand back and hide while the chaos ensues is to be honest not in the spirit of how wizard is meant to be played, wizard is meant to get stuck in, the games are meant to be short or at least interesting with the wizard as an active participant not a brutal slaughter where the main antagonist of the show, the wizard is no where to be found. 
As for raging mages, this should need no explanation, going all in on demons in raging is being actively malicious to your own side since the offending wizard is not being an active participant, they're just hiding and their follow wizards are getting eaten by demons that they've summoned.

## Changelog
:cl:
tweak: demons are no longer available in raging mages.
tweak: demon summon purchases are capped 1 of each type. 
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/spellbook.dm
+++ b/code/game/gamemodes/wizard/spellbook.dm
@@ -516,8 +516,9 @@
 	item_path = /obj/item/antag_spawner/slaughter_demon/laughter
 	log_name = "HB"
 	category = "Summons"
-	limit = 3
+	limit = 1
 	cost = 1 // Non-destructive; it's just a jape, sibling!
+	is_ragin_restricted = TRUE
 
 /datum/spellbook_entry/item/bloodbottle
 	name = "Bottle of Blood"
@@ -525,7 +526,8 @@
 	item_path = /obj/item/antag_spawner/slaughter_demon
 	log_name = "BB"
 	category = "Summons"
-	limit = 3
+	limit = 1
+	is_ragin_restricted = TRUE
 
 /datum/spellbook_entry/item/contract
 	name = "Contract of Apprenticeship"


Pixelzilla:
Why are we trying to dictate how other people choose to spend _their_ points and how _they_ choose to play? It isn't even that common, is this because of the round last night with like two demons and a few people were extra salty because they died?

Aquilar:
> Why are we trying to dictate how other people choose to spend _their_ points and how _they_ choose to play? It isn't even that common, is this because of the round last night with like two demons and a few people were extra salty because they died?

This is specifically eliminating demon all in, and additionally removes it from ragin' mages where there is already restrictions on what ragin' mages are allowed to bring. 
you can purchase demons in normal wizard mode, just bringing demons will no longer be a viable strategy in itself anymore.
Wizard should be aiming to do the chaos in person or at least be an active participant in the chaos not sitting in a corner thumbing their nose waiting for shuttle and their green text.

AzuleUtama:
As long as wizards are bringing chaos to the station in some manner, regardless of how direct it is, I don't really see why we should be nerfing specific strategies just because the wizard is taking a backseat approach to making the station collapse. The large number of demons on the station are a very direct threat for the crew to fight and the demons themselves can kill the wizard themselves if they wish, since it's part of their objectives. Not to mention doing so gives lots of ghosts a chance to play demon which is already a fairly rare antag role.

The only change I would agree with is removing them from raging, since it's already a very chaotic gamemode with multiple wizards for the crew to fight as it is, never mind letting each of them summon demons individually as well which remain even after their summoner has died.

TDSSS:
The point of wizard is to sow chaos, and summoning a bunch of demons seems like very in line with that. If people really are fine doing nothing but hiding all round, then let them imo.

Dashsora:
Here's the thing - a single competent demon can already bring the station to its knees, with very little counterplay. The only reliable way to get rid of one who strives to survive is shotguns.. and even then, I've  seen some just crawling out for half a second to hit someone, and then back in they go. A SINGLE one can be pretty frustrating to fight already... and this PR doesn't remove that. 

The ability to buy a slaughter and a laughter would still be there, and the wizard could take more spells to support this passive approach to bringing chaos/to help them survive (a morph, mindswaps, no clothes, lich, blink, mist, holopara, stuff like that).

 Morph, slaughter and laughter demons are FOUR points used for plenty PLENTY enough chaos, with some wider diversity in what's used, and leaving the wiz six points to allow them to still do fun stuff themselves. I really don't think there's any need for more.

If you don't wanna be a wizard that uses spells to have an - at least somewhat - active part to bringing the chaos... why do you even have wiz enabled ? It's the same question I ask the people who release engines thirty minutes in for hijack and then have absolutely NO follow up (and usually die) - why are you even attempting a hijack at that point ?

You might say " it gives better odds for newer players who don't have any other idea", and to this I say again, what's the point then ? They won't try things and learn to be better for next time, and might as well not have even attempted anything. If you limit those things, you encourage those people to actually consider other options to try out.

Varghh:
Echoing what @Dashsora

Butlerblock:
This is literally just a salt PR from when i summoned 3 laughters and 3 slaughters during a ragin mages. Half of dchat got angry for seemingly no reason because “noooo you can’t just focus on only chaos during a wizard round you’re only supposed to do minor chaos!” If you think demons are op(which they are btw), focus on nerfing then directly, not how they’re summoned. Maybe if they’re nerfed the random event chance can be increased(from like 2%) so we see them more often.

Kiyahitayika:
Could you cap the # of demons a single-wizard round Wizard can summon at 1 per type as long as that type of demon is alive?

So, say I summon a Laughter demon. I can only summon 1 of them because I can only summon 1 Laughter demon. Then it dies, but I still have points. Why not let me summon a _second_ Laughter demon now?

That stops there from being 6 demons AT ONCE, but you can still have 6 demons in a single game-round. Your Wizard will be notified when a demon-type dies. That way, the Wizard can't summon all 6 at once and may have to rely on other spells to defend themselves. If you're good at hiding, maybe you'll be able to summon six demons.

S34NW:
NGL I've never seen this happen once, is it so common a problem that it requires these changes? Esp. deleting them entirely during raging.
Raging is meant to be chaotic, instead of removing them entirely why not make them die when their summoner dies?

Kyep (commented):
Not convinced this is needed.
I can't even remember the last time I saw a wizard go full demons... and even if they did they have to do so on-station and the demons will likely kill the wizard sooner or later.

Necaladun:
If someone wants to give up their fancy cool spells to let a bunch of other players be demons I'm all ok on that.

AffectedArc07:
2 heads and 2 maints against, with no one in favour. Closing.

PR #15219, opened 2021-01-01
[TGUI] Depot Computers

Kyep:
## What Does This PR Do
1. Converts syndi depot computers to TGUI.
2. Minor refactoring/simplification of computer code. E.g:
Replacing X.len with length(X), eliminating sub-arguments to secondary functions
Consolidating functions (e.g: moving "logout visitors" function from comms computer to AI computer for better code re-use)
3. During peaceful mode when an agent has turned on the teleporter, they can now also turn it off, they're not stuck with it on forever

## Why It's Good For The Game
- New UI looks much nicer. Like an actual computer instead of a 1980s-era black-on-white webpage.
- Allows ghosts to click computers and see what's on their screen (but not click any of the buttons, unless they're an aghost with adv admin interaction on)

## Images of changes
depot_comms
depot_aiterminal
depot_teleporter
depot_shields
depot_doors
depot_reactor

## Changelog
:cl: Kyep
tweak: Converted syndi depot computers to TGUI.
/:cl:


code/game/machinery/computer/depot.dm
```diff
@@ -84,17 +72,54 @@
...
+	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
+	if(!ui)
+		ui = new(user, src, ui_key, "ComputerSimple",  name, window_width, window_height, master_ui, state)
+		ui.open()
+
+/obj/machinery/computer/syndicate_depot/ui_data(mob/user)
+	var/list/data = list()
+	data["rows"] = list()
+	if(security_lockout)
+		data["rows"].Add(list(list(
```
AffectedArc07, line 85:
```suggestion
		data["rows"] += list(list(
```
These warning can be suppressed by doing this. 

Explanation of warning: `data["rows"]` is not declared as a static type here, meaning that if `data["rows"]` was *not* equal to a list when the code was running, runtime errors would occur. The *proper* way to fix this is to cast to a list each time
```dm
var/list/rows = data["rows"]
rows.Add(whatever)
```
But that is a lot of extra lines, so ill allow `data["rows"] +=`

code/game/machinery/computer/depot.dm
```diff
@@ -84,17 +72,54 @@
...
+			"title" = "Security Lockout",
+			"status" = "Due to heightened security alert, base computers are locked out.",
+		)))
+	else if(length(req_access))
+		data["rows"].Add(list(list(
```
AffectedArc07, line 90: Same applies to the others

code/game/machinery/computer/depot.dm
```diff
@@ -84,17 +72,54 @@
...
+			"title" = "Security Notice",
+			"status" = "This terminal requires a syndicate ID of sufficient clearance.",
+		)))
+	/*
+		Guide for making your own template sections for SimpleComputer:
+		data["rows"].Add(list(list(
```
AffectedArc07, line 96: Dont forget to update this one

code/game/machinery/computer/depot.dm
```diff
@@ -84,17 +72,54 @@
 /obj/machinery/computer/syndicate_depot/proc/disable_special_functions()
 	return
 
-/obj/machinery/computer/syndicate_depot/Topic(href, href_list)
+/obj/machinery/computer/syndicate_depot/ui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = TRUE, datum/tgui/master_ui = null, datum/ui_state/state = GLOB.default_state)
+	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
+	if(!ui)
+		ui = new(user, src, ui_key, "ComputerSimple",  name, window_width, window_height, master_ui, state)
```
AffectedArc07, line 78:
```suggestion
		ui = new(user, src, ui_key, "SyndicateComputerSimple",  name, window_width, window_height, master_ui, state)
```
UI should be prefixed with `Syndicate`. `ComputerSimple` is too generic and people may end up trying to use it to make on-station things.

code/game/machinery/computer/depot.dm
```diff
@@ -127,13 +149,28 @@
...
 	var/pub_access = FALSE
 
-/obj/machinery/computer/syndicate_depot/doors/get_menu(mob/user)
-	return {"<B>Syndicate Depot Door Control Computer</B><HR>
-	<BR><BR><a href='?src=[UID()];primary=1'>Toggle Airlock Emergency Access</a>
-	<BR><BR><a href='?src=[UID()];secondary=1'>Toggle Hidden Doors</a>
-	<BR>"}
+/obj/machinery/computer/syndicate_depot/doors/ui_data(mob/user)
+	var/list/data = ..()
+	data["rows"].Add(list(list(
```
AffectedArc07, line 157: Same here for the `+=` stuff

code/game/machinery/computer/depot.dm
```diff
@@ -164,12 +201,19 @@
...
-	<BR>"}
-	return menutext
+	window_height = 200 // this might appear big, but it has to have space for the lockout alert
+
+/obj/machinery/computer/syndicate_depot/selfdestruct/ui_data(mob/user)
+	var/list/data = ..()
+	data["rows"].Add(list(list(
+		"title" = "Reactor Containment Fields",
+		"status" = (istype(depotarea) && !depotarea.used_self_destruct) ? "Online" : "Offline",
+		"buttontitle" = (istype(depotarea) && !depotarea.used_self_destruct) ? "Disable" : null,
```
AffectedArc07, line 211: Question, how does this button render if the title is `null`?

code/game/machinery/computer/depot.dm
```diff
@@ -289,48 +352,37 @@
 	Syndicate_announce(input, user)
 	to_chat(user, "Message transmitted.")
 	log_say("[key_name(user)] has sent a Syndicate comms message from the depot: [input]", user)
-	updateUsrDialog()
 	playsound(user, sound_yes, 50, 0)
 
-/obj/machinery/computer/syndicate_depot/syndiecomms/secondary(mob/user, subcommand)
+/obj/machinery/computer/syndicate_depot/syndiecomms/secondary(mob/user)
 	if(..())
```
AffectedArc07, line 358: I feel like these checks are redundant since its already checked in `ui_act()`

tgui/packages/tgui/interfaces/ComputerSimple.js
```diff
@@ -0,0 +1,36 @@
+import { Fragment } from 'inferno';
+import { useBackend } from "../backend";
+import { Button, LabeledList, Section, Box, Table, ProgressBar } from "../components";
+import { LabeledListItem } from "../components/LabeledList";
+import { Window } from "../layouts";
+
+export const ComputerSimple = (props, context) => {
```
AffectedArc07, line 7:
```suggestion
export const SyndicateComputerSimple = (props, context) => {
```

tgui/packages/tgui/interfaces/ComputerSimple.js
```diff
@@ -0,0 +1,36 @@
...
+  const { act, data } = useBackend(context);
+  return (
+    <Window theme="syndicate">
+      <Window.Content>
+        {data.rows.map(record => (
+          <Section key={record.title} title={record.title}
+            buttons={
```
AffectedArc07, line 14: I feel like there should be a var in `record` itself to signify whether buttons should be present or not, incase theres just a section header.

AffectedArc07 (changes requested):

code/game/machinery/computer/depot.dm
Kyep: It doesn't. Per above, 'null' means don't render a button.

AffectedArc07:
Ignore `Annotate Lints` failing, this is the only one you need to worry about
```
Time: 22967ms
Built at: 01/01/2021 12:27:44 PM
          Asset     Size  Chunks                    Chunk Names
tgui.bundle.css  128 KiB       0  [emitted]         tgui
 tgui.bundle.js  728 KiB       0  [emitted]  [big]  tgui
Entrypoint tgui [big] = tgui.bundle.css tgui.bundle.js
Error: our build differs from the build committed into git.
Please rebuild tgui. <<<<<
```

tgui/packages/tgui/interfaces/ComputerSimple.js
Kyep: Originally I thought about doing that, but decided not to because I could just use whether buttontitle exists or not to also decide if there should be a button. A buttonexists var would be redundant.
What I'd prefer to do is not have the buttons={} section even have a button defined if buttontitle is null, but I haven't worked out a way to actually do that without having to duplicate the entire <Section>, which seemed like a very bad idea.

code/game/area/areas/depot-areas.dm
```diff
@@ -424,18 +424,49 @@
...
+  *
+  * E.g. list_show(depotarea.guard_list) returns a string like:
+  * "Syndicate Backup (123), Syndicate Backup(456), Syndicate Backup(789)", etc.
+  * Arguments:
+  * * list/L, the list from which to draw members
+  * * show_ckeys, bool, if true will display ckeys in addition to names
+  */
 /area/syndicate_depot/core/proc/list_show(list/L, show_ckeys = FALSE)
-	var/list/formatted = list()
+	// Returns a string of the names of the list members, separated by ", "
```
AffectedArc07, line 437: Redundant comment, doc exists above

code/game/area/areas/depot-areas.dm
```diff
@@ -424,18 +424,49 @@
...
+/**
+  * Returns a LIST of the NAMES of the mobs in the provided list.
+  *
+  * E.g. list_shownames(depotarea.guard_list) returns a list of the names of extra guard mobs in depot.
+  * Arguments:
+  * * list/L, the list from which to draw members
+  * * show_ckeys, bool, if true will display ckeys in addition to names
+  */
+/area/syndicate_depot/core/proc/list_shownames(list/L, show_ckeys = FALSE)
+	//
```
AffectedArc07, line 450: Blank comment

code/game/area/areas/depot-areas.dm
```diff
@@ -424,18 +424,49 @@
...
+	// Returns a string of the names of the list members, separated by ", "
+	var/list/formatted = list_shownames(L, show_ckeys)
+	return formatted.Join(", ")
+
+/**
+  * Returns a LIST of the NAMES of the mobs in the provided list.
+  *
+  * E.g. list_shownames(depotarea.guard_list) returns a list of the names of extra guard mobs in depot.
+  * Arguments:
+  * * list/L, the list from which to draw members
```
AffectedArc07, line 446:
```suggestion
  * * list/L, the list of UIDs from which to draw members
```

AffectedArc07 (changes requested):

code/game/machinery/computer/depot.dm
```diff
@@ -478,64 +520,72 @@
...
+	else if(depotarea.called_backup)
+		alertlevel = "Red"
+	else if(depotarea.local_alarm)
+		alertlevel = "Blue"
+	else
+		alertlevel = "Green"
+	data["rows"] += list(list(
+		"title" = "Alert Level",
+		"status" = alertlevel,
+		"buttontitle" = (allowed(user) && check_rights(R_ADMIN, 0, user)) ? "(ADMIN) Reset Alert Level" : null,
```
AffectedArc07, line 544:
```suggestion
		"buttontitle" = (allowed(user) && check_rights(R_ADMIN, FALSE, user)) ? "(ADMIN) Reset Alert Level" : null,
```

code/game/machinery/computer/depot.dm
```diff
@@ -478,64 +520,72 @@
...
+		"title" = "Visiting Agents",
+		"status" = !length(depotarea.peaceful_list) ? "None" : null,
+		"bullets" = length(depotarea.peaceful_list) ? depotarea.list_shownames(depotarea.peaceful_list) : null
+	))
+	return data
 
 /obj/machinery/computer/syndicate_depot/aiterminal/primary(mob/user)
-	if(..())
-		return
 	if(!check_rights(R_ADMIN, 0, user))
```
AffectedArc07, line 577:
```suggestion
	if(!check_rights(R_ADMIN, FALSE, user))
```

AffectedArc07 (changes requested):
Last things

AffectedArc07 (approved):

PR #15225, opened 2021-01-02
Starvation No Longer Impacts SSDs

Fox-McCloud:
I've seen way too many people dying in cryo because they've been SSD forever.

Starvation is meant to be a mechanic to punish not eating, during a shift, when you're active.

This won't prevent people who are AFK from starving (we can have a discussion on that one if needed), but it will prevent people who are SSD from doing so, which is, IMO, for the better.

:cl: Fox McCloud
tweak: SSD individuals will no longer die from starvation
/:cl:

### DIFF
--- a/code/datums/diseases/critical.dm
+++ b/code/datums/diseases/critical.dm
@@ -165,6 +165,8 @@
 
 /datum/disease/critical/hypoglycemia/stage_act()
 	if(..())
+		if(isLivingSSD(affected_mob)) // We don't want AFK people dying from this.
+			return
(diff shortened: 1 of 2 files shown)


hal9000PR:
how does this affect diona in the dark?

Kiyahitayika:
> how does this affect diona in the dark?

The current scope of this PR does not appear to affect Diona in the Dark which sounds like a really really really bad romance novel, omg.

Kyep (approved):

AffectedArc07 (approved):

PR #15236, opened 2021-01-04
Fixes the PDA Messenger not decoding HTML tags

SabreML:
## What Does This PR Do

Fixes the PDA 'Messenger' app not decoding escaped HTML tags in player names and messages.
Also did the same for the Message Monitor Console and the PDA 'Notekeeper' app.

PDA admin logs are unchanged because I didn't want to break something, so they're still a bit odd looking sometimes.

## Why It's Good For The Game

It's very irritating to send a long, thought out message to someone and have it ruined by encoding.

### Before:
broken name
before
### After:
fixed name
pda

## Changelog
:cl:
fix: Fixed characters like these: (& ' < >) changing to `&#39;` or something similar in the PDA 'Messenger' and 'Notekeeper' apps.
/:cl:


TheYeetster1:
I think there is a similar problem with the crew monitor, if someone's in the Head Of Security's Office it will read:
Vitals | (name) Head Of Security&#39;s Office

SabreML:
> I think there is a similar problem with the crew monitor, if someone's in the Head Of Security's Office it will read:
> Vitals | (name) Head Of Security's Office

Crew monitor, and I also noticed the ghost orbit menu has it too.
Presumably it's all TGUI interfaces.

AffectedArc07 (approved):

PR #15237, opened 2021-01-04
A bit of GDPR compliance stuff

AffectedArc07:
## What Does This PR Do
TLDR: Moves TOS consent handling to `/client` instead of `/mob`, and only stores data on the client when they request

Long version:
It occurred to me that we can end up storing data on clients before they have accepted TOS. While this is an absolute rarity, it is a bit of cause for concern, given this
```
Although cookies are mentioned only once in the GDPR, cookie consent is nonetheless a cornerstone of compliance for websites with EU-located users. This is because one of the most common ways for personal data to be collected and shared online is through website cookies.
```
We end up storing cookies on a client before they accept, which means if a client does not consent, they still end up with cookies logged. After this PR, cookies wont be stored on a client until they accept the main server TOS (Which you need to do anyway). 

## Why It's Good For The Game
Compliance with the law is a pretty good thing

## Changelog
Not really important

### DIFF
--- a/code/modules/client/client_defines.dm
+++ b/code/modules/client/client_defines.dm
@@ -105,3 +105,14 @@
 
 	/// Last world/time that a PM was sent to the player by an admin
 	var/received_discord_pm = -99999 // Yes this super low number is intentional
+
+	/// Has the client accepted the TOS about data collection and other stuff
+	var/tos_consent = FALSE
+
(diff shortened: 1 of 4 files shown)


Spacemanspark:
I am the law

Necaladun (approved):
yay law

dearmochi (approved):

PR #15238, opened 2021-01-05
Adds a warning message if the game config files aren't set up properly

SabreML:
## What Does This PR Do

Generates a runtime if the `config.txt` and `game_options.txt` config files aren't set up correctly.
(If they've not been moved out of the 'example' folder)

**IMPORTANT:**
If merged, people will have to make sure to update their config files to add the new settings, otherwise they'll get the runtime.

## Why It's Good For The Game

Hopefully helps alleviate the common question of 'Why is my character running at mach 3' when people are setting up their own server.

## Images of changes

image

## Changelog
:cl:
tweak: Loading a private server without the config files properly set up will now display a warning message.
/:cl:


code/game/world.dm
```diff
@@ -22,14 +22,16 @@ GLOBAL_LIST_INIT(map_transition_config, MAP_TRANSITION_CONFIG)
 
 	TgsNew(new /datum/tgs_event_handler/impl, TGS_SECURITY_TRUSTED) // creates a new TGS object
 	log_world("World loaded at [time_stamp()]")
-	log_world("[GLOB.vars.len - GLOB.gvars_datum_in_built_vars.len] global variables")
+	log_world("[length(GLOB.vars) - length(GLOB.gvars_datum_in_built_vars)] global variables")
 	GLOB.revision_info.log_info()
-	load_admins(run_async=FALSE) // This better happen early on.
+	load_admins() // This better happen early on.
```
SabreML, line 27: Removed the argument here because it's already set in the proc itself:
```cs
/proc/load_admins(run_async = FALSE)
```

AffectedArc07: Yes the arg is already set but given the circumstances of how obvious it needs to be that this proc cannot sleep, I would rather it stay there

config/example/config.txt
```diff
@@ -471,3 +471,7 @@ BYOND_ACCOUNT_AGE_THRESHOLD 7
 ## THE TRAILING SLASH ON THE END IS IMPORTANT SINCE IT JUST APPENDS THE CKEY TO THE END IN RAW
 ## Add a hash before the line below to disable the system
 CENTCOM_BAN_DB_URL https://centcom.melonmesa.com/ban/search/
+
+## This is to let the game know that the config files are loaded properly.
```
AffectedArc07, line 475: This is the wrong way to go about this and is an unnecessary config change. Just use `if(fexists("config/config.txt"))` inside the config loader.

AffectedArc07 (changes requested):

config/example/config.txt
SabreML: I *knew* there would be an easier way to do this, thanks!

Can't put it in `load_configuration()` though since that's done before `enable_debugger()`. I want it to make a breakpoint if someone's running the game through VSCode.

config/example/game_options.txt
```diff
@@ -76,4 +76,4 @@ LAVALAND_BUDGET 60
 
 ## Cube monkey limit
 # Amount of cube monkeys that can be spawned before the game limits them
-CUBEMONKEY_CAP 32
\ No newline at end of file
+CUBEMONKEY_CAP 32
```
AffectedArc07, line 79: Please revert the diff in here

code/game/world.dm
```diff
@@ -22,14 +22,16 @@ GLOBAL_LIST_INIT(map_transition_config, MAP_TRANSITION_CONFIG)
 
 	TgsNew(new /datum/tgs_event_handler/impl, TGS_SECURITY_TRUSTED) // creates a new TGS object
 	log_world("World loaded at [time_stamp()]")
-	log_world("[GLOB.vars.len - GLOB.gvars_datum_in_built_vars.len] global variables")
+	log_world("[length(GLOB.vars) - length(GLOB.gvars_datum_in_built_vars)] global variables")
 	GLOB.revision_info.log_info()
-	load_admins(run_async=FALSE) // This better happen early on.
+	load_admins(FALSE) // This better happen early on.
```
AffectedArc07, line 27: Again, nitpicky but please keep the actual words "run_aysnc" just to make it absolutely abundantly clear. These parts of code aren't modified often, and should be very obvious for where mistakes can be made.

AffectedArc07 (changes requested):

config/example/game_options.txt
SabreML: That's just adding a line ending, I think.

code/game/world.dm
SabreML: Don't mind nitpicks, especially since this is a pretty important bit of the code.

FlattestGuitar:
Generating a runtime? You think whoever is fucking up this badly at setting up a local server will know to check runtimes? Why not just a `to_chat(world)` or at least an admin log

SabreML:
> Generating a runtime? You think whoever is fucking up this badly at setting up a local server will know to check runtimes? Why not just a `to_chat(world)` or at least an admin log

`to_chat(world, "error message")` needs them to actually be logged in for it to work, which they probably won't be since it's before initialization starts.

Admin logs has the same issue as runtimes, but worse since it won't make a breakpoint on the offchance they're running it through vscode.

FlattestGuitar:
A good point!

Kyep (commented):
A runtime is really not a good fit for this.
New admins who don't know they have to copy config files won't know what a runtime error is or the fact they should check for them.

If anything, this sort of warning should appear in red text when you connect to the server.

code/modules/mob/login.dm
```diff
@@ -68,5 +68,9 @@
 		for(var/datum/alternate_appearance/AA in viewing_alternate_appearances)
 			AA.display_to(list(src))
 
+	if(!fexists("config/config.txt") || !fexists("config/game_options.txt"))
+		to_chat(src, "<span class='biggerdanger'>The game config files have not been properly set!\n Please copy ALL files from '/example/config' into the parent folder, 'config'.</span>")
```
AffectedArc07, line 72: It's config/example, not example/config

AffectedArc07 (changes requested):

code/modules/mob/login.dm
SabreML: Oh right, obviously. 🤦

Kyep (approved):
So long as the fexists checks don't cause lag, approved.

AffectedArc07 (approved):

PR #15239, opened 2021-01-05
Adds GBS bottle as a 20 TC hijack item for Virologists

warior4356:
## What Does This PR Do
This PR adds a bottle of real GBS as a hijack item for virologists. This will cost them 20 TC and not come in a viral injector, so they need to find a way to distribute it and protect themselves from it.

## Why It's Good For The Game
This PR gives GBS an actual place in the game. gives virology a dangerous hijack only item. This item is also very easy to kill yourself with, and very hard to distribute without proper planning. On the safety side though, everything to cure it is already easily accessible to medical via nanomedplus or virology smart fridge + sulfur from chem. This should help add some more flavor to the hijack meta as well for non atmos/engineer/chaplain/scientist.

## Changelog
:cl:
add: Adds GBS as a 20 TC hijack item for Virologists
/:cl:

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -297,6 +297,16 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	cost = 3
 	job = list("Virologist")
 
+/datum/uplink_item/jobspecific/gbs_bottle
+	name = "Bottle of GBS"
+	desc = "An extremely dangerous illness caused by contagious micro-singularities that will cause the human body to be torn apart from the inside. The initial phase begins in the lungs and mimics flu symptoms. Wear your protective suit."
+	reference = "GBS"
+	item = /obj/item/reagent_containers/glass/bottle/gbs
+	cost = 20
+	job = list("Virologist")
+	surplus = 0 //No lucky chances from the crate; if you get this, this is ALL you're getting
+	hijack_only = TRUE //This is an extremely lethal disease and will do similar damage to the crew that a mass bombing would
+
 /datum/uplink_item/jobspecific/cat_grenade
 	name = "Feral Cat Delivery Grenade"
 	desc = "The feral cat delivery grenade contains 5 dehydrated feral cats in a similar manner to dehydrated monkeys, which, upon detonation, will be rehydrated by a small reservoir of water contained within the grenade. These cats will then attack anything in sight."


AffectedArc07:
Not in favour. There's a reason this thing is admin only, it's one of the most overpowered things in the entire game.

Fox-McCloud:
Yeah, I don't think so.

GBS is highly contagious and just straight up gibs you or kills you dead super fast from toxins before it does.

It's been admin-only, for years, for a reason.

Kyep:
While having a dangerous hijack-only item for virologists is fine, GBS is too strong to be used for this.
I've run GBS infection events a couple of times over the years as an admin, and they almost invariably result in a lot of deaths.
A better solution would be a new virus which is balanced such that it CAN kill a lot of people if untreated, but there's also enough time that the CMO or someone they transfer could cure the virus.
Basically, hijack-only virus as a concept is fine, but there has to be counterplay. GBS is too lethal, and too fast, to allow adequate counterplay.

Necaladun:
Last outbreaks I've done in events have been cured super fast with at most a handful of people gibbing. Last I remember the cure is sulfur? For 20TCs and hijack you could do a lot more damage.

Sure, a bunch of people are gonna be gibbed, but a good bombing of medbay would do way more. There's also a fair chance this will get the virologist lynched too. Compared to something like His Grace I think this is fair, especially if it's a single-job-hijack-only item.

If I'm wrong about the sulfur though I'd be more hesistant.

Mitchs98:
Definitely disagree with GBS being added in even as hijack only for all the reasons listed above. A different item could be pretty fun and definitely spice things up, but GBS is very easy to spiral out of control with and could well end with a significant portion of the station dead quite quickly.

> If I'm wrong about the sulfur though I'd be more hesistant.

Not too far off, sulfur + diphendamine(or however you spell it) at once. And it can be pretty annoying to cure if it doesn't wanna proc/people don't know you need both.

Necaladun:
Yeah I was wrong. I'm testing it right now and this is fucking brutal.  All I did was infect myself and claim it was a hoax and refuse to wear a mask. If a traitor was intentionally spreading this it'd be brutal and quite easily take out most of the station. Even if it was just sulfur to cure, it's so quick to hit this would destroy the station.

warior4356:
If the speed of progression is the issue, it would be trivial to half or quarter the progression rate of GBS. For reference by default, diseases have a 4% chance to progress a stage every tick. If it was decreased to a 2% chance or a 1% chance the time to kill would be around 4-5 or 8-10 minutes on average which would be far less likely to result in mass casualties but still be a crisis.

farie82:
Not in favor of this one myself. If done right the viro can nearly instantly kill all the crew without the crew being able to respond on it. I remember having this as a random event and I got gibbed myself within 2 minutes after getting it as viro. While my monkey didn't cure it while it had the antibodies virus in him.
RNG be RNG.

If the virus was a bit more predictable in it's stage speed and when you die then maybe. But as it is now you can die from it in 10 seconds if you're unlucky all the while the first stages have very little in the way of warning the player they are about to die

AffectedArc07:
2 maintainer objections, and no head or maintainer in favour. Closing.

PR #15240, opened 2021-01-05
Mindshield crates are more expensive

TDSSS:
## What Does This PR Do
This PR increases the cost of mindshield crates to 80 points, or 20 points per mindshield.

## Why It's Good For The Game
There's no real need for the crew/sec to order large amounts of mindshields. The station supply should be more than enough to re-mindshield cloned sec officers, and if they really run out they could simply remove the implants from the dead. The only reason to order lots of mindshield crates is to mindshield lots of non-sec crew. The only gamemode balanced around that is revolution, and the word 'balanced' is a pretty big overstatement there. Revolution is out of rotation, has been for a long-ass time and is never used.

Other conversion antags aren't really designed around most of the crew being mindshielded and immune to conversion. Not to mention mass mindshielding even unlocks the cult's final objective faster, if I remember right.

Lastly, if mindshields were cheap, why doesn't NT just mindshield everyone just as a general precaution? Mindshields are supposed to be something rare and reserved for security, not something the station can just spam on a lark.

## Changelog
:cl:
tweak: mindshield crates are more expensive
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -479,7 +479,7 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 /datum/supply_packs/security/armory/mindshield
 	name = "Mindshield Implants Crate"
 	contains = list (/obj/item/storage/lockbox/mindshield)
-	cost = 40
+	cost = 80
 	containername = "mindshield implant crate"
 
 /datum/supply_packs/security/armory/trackingimp


farie82 (approved):

Kyep:
I'm skeptical about this.
Yeah, command can order a ton of mindshields and mindshield everyone.
But this typically backfires. It just results in cult for example turning people into un-deconvertable constructs... and them getting the summon objective earlier.
I'm unsure why we need to nerf it as a strategy if it is typically not a winning strategy.

TDSSS:
> I'm skeptical about this.
> Yeah, command can order a ton of mindshields and mindshield everyone.
> But this typically backfires. It just results in cult for example turning people into un-deconvertable constructs... and them getting the summon objective earlier.
> I'm unsure why we need to nerf it as a strategy if it is typically not a winning strategy.

Yes, cult has a built-in safeguard against doing this but a) this doesn't apply to shadowlings and b) I think on principle it's better to just not make mass-mindshielding possible rather than having every gamemode where it could become relevant get specific safeguards against it.

What happens when we get another antag or mode where people get converted, do they also need to put in a 'fuck station over if there's too many mindshielded people' mechanic? At some point that just cheapens mindshields.

Initially, I was even going to just disable the crate because there's no reason station should even order it, but it's a tool for antags to get creative and impersonate sec, so I figured it should stay.

Daylight2:
> I'm skeptical about this.
> Yeah, command can order a ton of mindshields and mindshield everyone.
> But this typically backfires. It just results in cult for example turning people into un-deconvertable constructs... and them getting the summon objective earlier.
> I'm unsure why we need to nerf it as a strategy if it is typically not a winning strategy.

The only time mindshields are ordered are when slings are spotted, and every assistant who wants to plant glow tiles in maint gets implanted with one.

And if implanting everyone is not, as you say, a winning strategy, then it shouldn't matter if the crates are more expensive, due to ordering them being against the crew's best interest.

Necaladun:
I'm not really sure why this is needed? Rev isn't in rotation and doesn't matter, and a hypothetical potential future game mode isn't a reason to change anything at all. 

If there's an actual issue with mass mindshielding, I'd consider it, but I'm not seeing it.

Daylight2:
> I'm not really sure why this is needed? Rev isn't in rotation and doesn't matter, and a hypothetical potential future game mode isn't a reason to change anything at all.
> 
> If there's an actual issue with mass mindshielding, I'd consider it, but I'm not seeing it.

To elaborate on my last comment, if a good portion of the crew (especially those who wander into maints) are mindshields, slings don't even stand a chance, as those they manage to kidnap are usually mindshielded and thus useless to them (usually leading to an awkward moment when the sling just has to beat them to death) + their convert objective doesn't take into account (AFAIK) how many people are mindshielded, which could lead to shadowlings winning being physically impossible when enough mindshields are dished out.

AffectedArc07:
2 heads and a maint against, no one in favour. Closing.

PR #15246, opened 2021-01-05
Fixes Mentor asay

SabreML:
## What Does This PR Do

Removes the ability for mentors to (somewhat) access the 'asay' verb by pressing F5. The input will get sent to 'msay' now instead.

~~*(Only been a mentor for half a day and I'm already fixing things, smh.)*~~

## Why It's Good For The Game

Nothing typed into the input box would actually get through to the asay chat, but it probably still shouldn't appear in the first place.

## Changelog
:cl:
fix: Mentors pressing F5 will now open a window for 'msay', rather than 'asay'.
/:cl:

### DIFF
--- a/code/modules/admin/verbs/adminsay.dm
+++ b/code/modules/admin/verbs/adminsay.dm
@@ -26,8 +26,9 @@
 	SSblackbox.record_feedback("tally", "admin_verb", 1, "Asay") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
 /client/proc/get_admin_say()
-	var/msg = input(src, null, "asay \"text\"") as text|null
-	cmd_admin_say(msg)
+	if(check_rights(R_ADMIN))
+		var/msg = input(src, null, "asay \"text\"") as text|null
+		cmd_admin_say(msg)
 
 /client/proc/cmd_mentor_say(msg as text)
 	set category = "Admin"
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/adminsay.dm
```diff
@@ -26,8 +26,9 @@
 	SSblackbox.record_feedback("tally", "admin_verb", 1, "Asay") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
 /client/proc/get_admin_say()
-	var/msg = input(src, null, "asay \"text\"") as text|null
-	cmd_admin_say(msg)
+	if(check_rights(R_ADMIN))
```
AffectedArc07, line 29: Why not add an `else if(check_rights(R_MENTOR))` here for msay to be used so mentors can also use the F5 function

SabreML: Actually that's a good idea, I'll do that.

code/modules/admin/verbs/adminsay.dm
```diff
@@ -26,8 +26,9 @@
...
+		var/msg = input(src, null, "asay \"text\"") as text|null
+		cmd_admin_say(msg)
```
farie82, line 31:
```suggestion
	if(!check_rights(R_ADMIN))
		return
	var/msg = input(src, null, "asay \"text\"") as text|null
	cmd_admin_say(msg)
```
Early returns are preferred

dearmochi (approved):

AffectedArc07:
image
This branch name looks promising

Kyep (approved):

AffectedArc07 (approved):

PR #15249, opened 2021-01-06
Tuulipoomi's Barsign Fix - Remastered

Kiyahitayika:
## Credit
A few PRs here have attempted to fix this with the most recent one being https://github.com/ParadiseSS13/Paradise/pull/15063 . 
The work completed by Tuulipoomi and the Maintainers and Reviewers was very close to mergeable code, just needing a few tweaks to become mergeable.

## What Does This PR Do

- Sets out to resolve Issue: https://github.com/ParadiseSS13/Paradise/issues/14884.
- If the barsign has been emagged, the list of signs will include the "hidden signs"

## Why It's Good For The Game
Invisible sign bad.
Choose signage good.
Now with menu - perhaps as originally intended?

## Images of changes
image
image
**The above image is out of date. I coded a chunk to do this and then realized that there was already code to do this, so I deferred to that code. That's why the language does not match up.**
image

## Changelog
:cl:
tweak: Barsign's maintenance panel can be opened using a screwdriver, open Barsign only be changed via attacking with hand by either Bartender or Syndicate Agents (post-emag), open Barsign can be dismantled and moved with a wrench by anyone.
fix: Barsign can no longer turns invisible.
/:cl:

### DIFF
--- a/code/game/objects/structures/barsign.dm
+++ b/code/game/objects/structures/barsign.dm
@@ -18,7 +18,7 @@
 /obj/structure/sign/barsign/New()
 	..()
 	//filling the barsigns lists
-	for(var/bartype in subtypesof(/datum/barsign))
+	for(var/bartype in typesof(/datum/barsign))
 		var/datum/barsign/signinfo = new bartype
 		if(!signinfo.hidden)
 			barsigns += signinfo
@@ -58,13 +58,17 @@
 
 /obj/structure/sign/barsign/attack_hand(mob/user)
 	if(panel_open)
-		if(broken)
-			to_chat(user, "<span class='danger'>The controls seem unresponsive.</span>")
-			return
+		if(allowed(user))
+			if(broken)
+				to_chat(user, "<span class='danger'>The controls seem unresponsive.</span>")
+				return
+			else
+				pick_sign()
+				to_chat(user, "<span class='notice'>You set the barsign and close the maintenance panel.</span>")
+				panel_open = FALSE
+				return
 		else
-			pick_sign()
-			to_chat(user, "<span class='notice'>You set the barsign and close the maintenance panel.</span>")
-			panel_open = FALSE
+			to_chat(user, "<span class='info'>Access denied.</span>")
 			return
 	else
 		to_chat(user, "<span class='info'>The maintenance panel is currently closed.</span>")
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/barsign.dm
```diff
@@ -48,68 +49,75 @@
...
+	var/list/signs = barsigns
+	var/new_sign
+	if(!broken && !emagged)
+		new_sign = input("Available Signage: ", "Bar Sign", null) as null|anything in signs
+		set_sign(new_sign)
+	else if(!broken && emagged)
+		signs += hiddensigns
+		new_sign = input("Available Signage: ", "Bar Sign", null) as null|anything in signs
+		set_sign(new_sign)
+	else
```
Kiyahitayika, line 115: Looks like I made a mistake here. I think this should be `else if(broken`. That, or the indentation should change and have an if(!broken).

code/game/objects/structures/barsign.dm
```diff
@@ -386,29 +363,20 @@
 	desc = "Open since 2125, Not much has changed since then; the engineers still release the singulo and the damn miners still are more likely to cave your face in that deliver ores."
 
 /datum/barsign/hiddensigns
```
Kiyahitayika, line 365: Is this subclass necessary? If it's just setting one new variable, why not have the variable set on the `/barsign` class and default it to FALSE and set it to true on each of the "hiddensigns"?

S34NW:
Can emagging it automatically force one of the "hidden signs" to be set? No bartender in their right mind would emag it if only they can change the sign...

code/game/objects/structures/barsign.dm
```diff
@@ -48,68 +49,76 @@
...
+			to_chat(user, "<span class='notice'>You close the maintenance panel.</span>")
+			panel_open = FALSE
 	else
-		return ..()
-
+		to_chat(user, "<span class='info'>Access denied.</span>")
+		return
 
+/obj/structure/sign/barsign/proc/pick_sign()
+	var/list/signs = barsigns
```
Kiyahitayika, line 106: var/list/signs = barsigns.Copy()
per @blinkdog

Kiyahitayika:
> Can emagging it automatically force one of the "hidden signs" to be set? No bartender in their right mind would emag it if only they can change the sign...

Good catch. I did not account for code that I didn't change:

```
/obj/structure/sign/barsign/proc/post_emag()
	if(broken || emagged)
		return
	set_sign(new /datum/barsign/syndibarsign)
	emagged = TRUE
```

According to my changes, the code now means that if the barsign gets emagged, only the Syndicate can open it with a screwdriver. Hmm. I don't think that's good.

I had switched the required access to the attackby() in a test (instead of putting it in screwdriver_act()), I'll try that again.

Kyep (commented):
Skeptical of this.
While the fix is nice, I don't like the fact the bar sign can no longer be dismantled.
I am certain that this will result, sooner or later, in people having to ahelp to have the sign moved or removed.

Kiyahitayika:
@Kyep 
The current method of dismantling the barsign makes it invisible and does not allow for moving since it cannot be reconstructed. The scope of this PR was to bring the work of @tuulipoomi's https://github.com/ParadiseSS13/Paradise/pull/15063 into a mergeable state. If this PR does not achieve the best result for the server, please feel free to close it. The code will still be available and free to be used for anyone who would like to fix being able to remove and move the barsign.

Kiyahitayika:
image

Kiyahitayika:
@Kyep I think I've figured out a way. I used the same structure as other signs. In testing, the sprite was no longer invisible and dragging it worked fine. You can now fasten it elsewhere.

The difference is that the barsign is using `wrench_act()` instead of `screwdriver_act()`. I switched the `attackby()` in signs.dm from checking for screwdriver as well to `screwdriver_act()` and asked it to make sure the src type was not a barsign. ... which reminds me, I need to make a quick fix to re-add `attackby()` to `signs.dm`.

Kiyahitayika:
After testing this live on Scorpio - I'm going to make another adjustment.  If someone opens and closes the panel without changing the sign, it should default back to the sign it was before the panel was opened.

Kiyahitayika:
https://imgur.com/a/78Ng2bS

Aaah so close.

Kiyahitayika:
Raaah, merge conflicts. I'll just use a new branch. Sorry.

PR #15257, opened 2021-01-07
Two handed Shotguns

shmabuu:
## What Does This PR Do
Makes it so you must have your other hand empty to shoot a shotgun with lethal capabilities with one hand, the exception being sawn-off improvised and sawn-off double-barrel shotguns.

## Why It's Good For The Game
Shotguns need some kind of nerf. Them being used in tandem with shields to deal high amounts of damage and have a decent chance at blocking damage or stuns coming your way isn't very balanced. The same goes for using a shotgun at the same time as say, an e-sword. I also think that being able to shoot and pump and fire a shotgun with one hand is a bit much if you do end up pairing it with something

## Images of changes
shotgun

## Changelog
:cl:
Tweak: You can no longer have both of your hands full While holding any shotgun with lethal capabilities. The exception being sawn-off improvised shotguns and sawn-off double-barreled shotguns.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/revolver.dm
+++ b/code/modules/projectiles/guns/projectile/revolver.dm
@@ -330,6 +330,7 @@
 	icon_state = "dshotgun"
 	item_state = "shotgun"
 	w_class = WEIGHT_CLASS_BULKY
+	weapon_weight = WEAPON_HEAVY
 	force = 10
 	flags = CONDUCT
 	slot_flags = SLOT_BACK
@@ -362,6 +363,9 @@
 	else
 		return ..()
 
+/obj/item/gun/projectile/revolver/doublebarrel/sawoff(mob/user)
+    . = ..()
+    weapon_weight = WEAPON_MEDIUM
 /obj/item/gun/projectile/revolver/doublebarrel/attack_self(mob/living/user)
 	var/num_unloaded = 0
 	while(get_ammo() > 0)
--- a/code/modules/projectiles/guns/projectile/revolver.dm
+++ b/code/modules/projectiles/guns/projectile/revolver.dm
@@ -366,6 +366,7 @@
 /obj/item/gun/projectile/revolver/doublebarrel/sawoff(mob/user)
     . = ..()
     weapon_weight = WEAPON_MEDIUM
+
 /obj/item/gun/projectile/revolver/doublebarrel/attack_self(mob/living/user)
 	var/num_unloaded = 0
 	while(get_ammo() > 0)
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -131,7 +131,7 @@
 	name = "assault shotgun"
 	desc = sawn_desc
 	w_class = WEIGHT_CLASS_NORMAL
-	weapon_weight = WEAPON_MEDIUM
+	weapon_weight = WEAPON_HEAVY
 	current_skin = "riotshotgun-short"
 	item_state = "gun"			//phil235 is it different with different skin?
 	slot_flags &= ~SLOT_BACK    //you can't sling it on your back
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -83,7 +83,6 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/riot
 	sawn_desc = "Come with me if you want to live."
 	sawn_state = SAWN_INTACT
-	weapon_weight = WEAPON_HEAVY
 
 /obj/item/gun/projectile/shotgun/riot/attackby(obj/item/A, mob/user, params)
 	if(istype(A, /obj/item/circular_saw) || istype(A, /obj/item/gun/energy/plasmacutter))
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -83,6 +83,7 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/riot
 	sawn_desc = "Come with me if you want to live."
 	sawn_state = SAWN_INTACT
+	weapon_weight = WEAPON_HEAVY
 
 /obj/item/gun/projectile/shotgun/riot/attackby(obj/item/A, mob/user, params)
 	if(istype(A, /obj/item/circular_saw) || istype(A, /obj/item/gun/energy/plasmacutter))
@@ -308,6 +309,7 @@
 	origin_tech = "combat=6"
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/com
 	w_class = WEIGHT_CLASS_HUGE
+	weapon_weight = WEAPON_HEAVY
 
 //Dual Feed Shotgun
 
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -12,7 +12,7 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/shot
 	fire_sound = 'sound/weapons/gunshots/gunshot_shotgun.ogg'
 	var/recentpump = 0 // to prevent spammage
-	weapon_weight = WEAPON_MEDIUM
+	weapon_weight = WEAPON_HEAVY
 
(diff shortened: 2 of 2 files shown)


Muhdi345:
I'd rather it didn't block your attempt to fire it outright though... Maybe make it act like the L6 when fired with something in hand?

shmabuu:
@Muhdi345 It's the exact same thing:
saw

`/obj/item/gun/projectile/automatic/l6_saw
	name = "\improper L6 SAW"
	desc = "A heavily modified 5.56 light machine gun, designated 'L6 SAW'. Has 'Aussec Armoury - 2531' engraved on the receiver below the designation."
	icon_state = "l6closed100"
	item_state = "l6closedmag"
	w_class = WEIGHT_CLASS_HUGE
	slot_flags = 0
	origin_tech = "combat=6;engineering=3;syndicate=6"
	mag_type = /obj/item/ammo_box/magazine/mm556x45
	weapon_weight = WEAPON_HEAVY`

Muhdi345:
Huh. Guess it's been a while since I did that... I remember it being different 2 or 3 years ago... Screen going wonk as if you're drunk and you'd fire in a random direction.

FreeStylaLT:
May be better to make it work the same way the fireaxe does, two-handing it increasing the accuracy.

HeWhoMustNOtBe:
I do not think accuracy matters to the shotgun, its kinda its entire point. You either point blank or spray and hope one hits

FreeStylaLT:
If you increase the angles to be at potentially 90 degrees from the firer when one-handed, it'll definitely make it nearly unusable at anything beyond 2-3 tiles and barely consistently effective at even point blank range.

Carthusia:
I agree that it being more like a fire axe would be better, always felt all larger firearms should be like that, similar to on CM, you CAN fire them one handed, but its not advisable.

Could lead to a new combat implant that steady's your aim for one handed firing.

What about saw shortened riot shotguns? Their size is about that of a slightly larger pistol

Mitchs98:
Definitely would prefer an accuracy penalty, or potentially even a minor brute damage penalty. 

And as Kar-Ski mentions sawn-off should maintain current functionality, really, given its smaller size.

Varghh:
Shotguns have been fucking stupid for years, this is a good nerf.

Fox-McCloud:
Just do this to all shotguns (minus the bulldog).

Keep it so if you saw it off the double-barrel, though, it's just a medium-sized weapon; the riot shouldn't get the same treatment though.

shmabuu:
> What about saw shortened riot shotguns? Their size is about that of a slightly larger pistol
>
Good point. Made the change to the assault shotgun
>Just do this to all shotguns (minus the bulldog).
> 
> Keep it so if you saw it off the double-barrel, though, it's just a medium-sized weapon; the riot shouldn't get the same treatment though.

Okay, I have these changes done all except for the improvised shotgun and the double-barrel shotgun... I want to make it so you are able to fire the sawn-off versions of these two shotguns with one hand but they aren't separated like the riot shotgun is. Would you be able to suggest how to do that?

Carthusia:
Is there anyway at all to make one handing a shotgun instead of being impossible, as noted above with others:

Have a chance to do some brute damage/the shotgun flys out of your hand
The pellets/slug have their spread increased.

If people cant one hand them, i feel situations where if say, someone who had a shotgun is caught in an explosion, but survives, but looses one hand, or in the case of IPC's someone just punches them and their hand falls off, a situation that would normally have a small chance for self defence and survival, becomes non existent.

At least if it wasnt impossible, but inadvisable, it would become a last ditch method compared to the current "Duel Fisten Shotguns"

Even with pump actions, its not impossible to one hand fire, and rack the next shell, its hard, its probably going to hurt, and you cant really properly aim the weapon, and hell, it may even knock you over if your not prepared for the recoil what with the lack of a supporting hand, but you can do it.

Fox-McCloud:
>Okay, I have these changes done all except for the improvised shotgun and the double-barrel shotgun... I want to make it so you are able to fire the sawn-off versions of these two shotguns with one hand but they aren't separated like the riot shotgun is. Would you be able to suggest how to do that?

Make it so the double-barrel's `sawoff` proc changes it to be a medium weapon.

shmabuu:
> > Okay, I have these changes done all except for the improvised shotgun and the double-barrel shotgun... I want to make it so you are able to fire the sawn-off versions of these two shotguns with one hand but they aren't separated like the riot shotgun is. Would you be able to suggest how to do that?
> 
> You'll first have to make the double-barrel a heavy weapon, specifically, since it's....technically a revolver subtype (don't...ask).
> 
> Then make it so the double-barrel's `sawoff` proc changes it to be a medium weapon.

Alright, this is done, along with the rest of the shotguns. I chose to make the 'assault shotgun' able to shoot with one hand due to its small size, reasonable downside, and general lack of usage. However, if there is generally negative response to the assault shotgun being one hand-able, I have no issues with changing it

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -131,6 +131,7 @@
 	name = "assault shotgun"
 	desc = sawn_desc
 	w_class = WEIGHT_CLASS_NORMAL
+	weapon_weight = WEAPON_MEDIUM
```
Fox-McCloud, line 134: I wouldn't do this to the riot shotgun; it's really strong too---the main advantage to this should be that you can store it in your backpack.

code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -362,6 +363,9 @@
 	else
 		return ..()
 
+/obj/item/gun/projectile/revolver/doublebarrel/sawoff(mob/user)
+    . = ..()
+    weapon_weight = WEAPON_MEDIUM
```
Fox-McCloud, line 368: need a new line.

shmabuu: I believe I've fixed it- do correct me if I'm wrong though

dearmochi (approved):
A much needed nerf!

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -83,6 +83,7 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/riot
 	sawn_desc = "Come with me if you want to live."
 	sawn_state = SAWN_INTACT
+	weapon_weight = WEAPON_HEAVY
```
Fox-McCloud, line 86: This inherits from parent, no need to define it again.

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -130,6 +131,7 @@
 	name = "assault shotgun"
 	desc = sawn_desc
 	w_class = WEIGHT_CLASS_NORMAL
+	weapon_weight = WEAPON_HEAVY
```
Fox-McCloud, line 134: just get rid of this entirely.

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -308,6 +310,7 @@
 	origin_tech = "combat=6"
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/com
 	w_class = WEIGHT_CLASS_HUGE
+	weapon_weight = WEAPON_HEAVY
```
Fox-McCloud, line 313: This inherits from parent, no need to define it again.

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -318,6 +321,7 @@
 	origin_tech = "combat=4;materials=2"
 	mag_type = /obj/item/ammo_box/magazine/internal/shot/tube
 	w_class = WEIGHT_CLASS_HUGE
+	weapon_weight = WEAPON_HEAVY
```
Fox-McCloud, line 324: This inherits from parent, no need to define it again.

Dumbdumn5 (approved):
Shotguns have been a bit boosted for a while now, and if all the code changes are settled, I'll have to approve this to see how it plays out. Admittedly should be a bit more of a risk to get people smacked with a shotty than it currently is.

AffectedArc07 (approved):
Will merge once Fox's review concerns have been addressed

code/modules/projectiles/guns/projectile/shotgun.dm
AffectedArc07, line 86: Last one

AffectedArc07 (changes requested):

PR #15261, opened 2021-01-07
Stops the cryo sleepers from putting items into null space. Fixing the pinpointer runtime. Fix recall items from being wacky in some cases

farie82:
## What Does This PR Do
Makes it so that items being stored in cryo get put in the console. And not in nullspace

Also fixes summonitems from breaking cryo dorms and atmos.

fixes:
```
Runtime in pinpointer.dm,359: Cannot read null.z
   proc name: trackable (/obj/item/pinpointer/crew/proc/trackable)
   src: the crew pinpointer (/obj/item/pinpointer/crew)
   src.loc: null
   call stack:
   the crew pinpointer (/obj/item/pinpointer/crew): trackable(NAME (/mob/living/carbon/human))
   the crew pinpointer (/obj/item/pinpointer/crew): point at(NAME (/mob/living/carbon/human))
   the crew pinpointer (/obj/item/pinpointer/crew): process(2)
   Fast Processing (/datum/controller/subsystem/processing/fastprocess): fire(0)
   Fast Processing (/datum/controller/subsystem/processing/fastprocess): ignite(0)
   Master (/datum/controller/master): RunQueue()
   Master (/datum/controller/master): Loop()
   Master (/datum/controller/master): StartProcessing(0)
```

## Why It's Good For The Game
Putting items into nullspace can lead to a ton of issues down the line as shown with the pinpointer.

## Changelog
:cl:
fix: Recall items no longer pulls vents to you when you use it while the item is in the vent
/:cl:

### DIFF
--- a/code/datums/spells/summonitem.dm
+++ b/code/datums/spells/summonitem.dm
@@ -12,7 +12,8 @@
 	include_user = 1
 
 	var/obj/marked_item
-
+	/// List of objects which will result in the spell stopping with the recursion search
+	var/static/list/blacklisted_summons = list(/obj/machinery/computer/cryopod, /obj/machinery/atmospherics, /obj/structure/disposalholder, /obj/machinery/disposal)
 	action_icon_state = "summons"
 
 /obj/effect/proc_holder/spell/targeted/summonitem/cast(list/targets, mob/user = usr)
@@ -89,7 +90,7 @@
 						var/obj/machinery/portable_atmospherics/P = item_to_retrieve.loc
 						P.disconnect()
 						P.update_icon()
-					if(istype(item_to_retrieve.loc, /obj/structure/disposalholder) || istype(item_to_retrieve.loc, /obj/machinery/disposal))//fixes the breaking of disposals. No more bluespace connected disposal bins!
+					if(is_type_in_list(item_to_retrieve.loc, blacklisted_summons))
 						break
 					item_to_retrieve = item_to_retrieve.loc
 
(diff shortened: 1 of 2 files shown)


code/datums/spells/summonitem.dm
```diff
@@ -12,7 +12,8 @@
 	include_user = 1
 
 	var/obj/marked_item
-
+	/// List of objects which will result in the spell stopping with the recursion search
+	var/static/list/blacklisted_summons = list(/obj/machinery/computer/cryopod, /obj/machinery/atmospherics, /obj/structure/disposalholder, /obj/machinery/disposal)
```
AffectedArc07, line 16: Can you make this use a proper typecache instead of just a regular type list?

farie82: sure thing

code/game/machinery/cryopod.dm
```diff
@@ -137,12 +141,23 @@
...
+	frozen_items -= I
+	UnregisterSignal(I, COMSIG_MOVABLE_MOVED)
+
 /obj/machinery/computer/cryopod/proc/dispense_item(obj/item/I)
 	if(!(I in frozen_items))
 		return
 	I.forceMove(get_turf(src))
-	objective_items -= I
-	frozen_items -= I
+	item_got_removed(I)
```
SteelSlayer, line 160: You don't need to call this directly, since `forceMove` calls `Moved` which sends the `COMSIG_MOVABLE_MOVED` signal which then calls the `item_got_removed` proc.
```suggestion
```

farie82: Oh true that. Will remove it and add a comment to explain it

AffectedArc07 (approved):

PR #15269, opened 2021-01-09
Adds reagent vision to observers

SabreML:
## What Does This PR Do

Allows ghosts to see reagents in containers, the same as wearing Science Goggles.

## Why It's Good For The Game

Better awareness about what's going on for observers.

## Changelog
:cl:
tweak: Added reagent vision for ghosts.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -698,6 +698,9 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 
 	updateghostimages()
 
+/mob/dead/observer/can_see_reagents()
+	return TRUE
+
 /proc/updateallghostimages()
 	for(var/mob/dead/observer/O in GLOB.player_list)
 		O.updateghostimages()


dearmochi (approved):

Fox-McCloud:
Ghost don't need to know everything---and the more knowledge they possess, the more they can act on it---I've never been a fan of ghosts being able to, effectively, know just about whatever they want.

Yes, we have "rules" in place, but it's pretty hard to tell who is doing meta stuff based off of this knowledge---especially if it's subtle---likewise, even those with good intentions are likely to behave in a manner different to what they otherwise would.

👎

SabreML:
I'd say it's around the same level of abusability as being able to look inside storage containers. If someone's left a box of C4 or a beaker of Initropidril out in the open then it's always possible that someone in-game could just come up and examine it.

While the chances of a ghost finding something suspicious are obviously much higher than that of a security officer, I don't think this will increase the chances of metagaming, especially since ghosts can already see what's being made in a chem dispenser.

Fox-McCloud:
>especially since ghosts can already see what's being made in a chem dispenser.

And that's definitely not something I think should be in existence, either, so I don't buy that argument, personally.


>I don't think this will increase the chances of metagaming

It objectively will--giving ghosts information they previously did not possess objectively increases metagaming information they have to act on---you can't really argue this away.

Kyep (approved):

AffectedArc07 (approved):

PR #15284, opened 2021-01-12
Makes the Contractor Kit a 20 TC uplink purchase

dearmochi:
## What Does This PR Do

- **Makes the Contractor Kit a 20 TC uplink purchase**
  - Not available to existing Contractors (aka can't buy two kits) and Nuclear Operatives
  - Increases Blackout cooldown to 45 minutes
  - Adds logging on target extraction

## Why It's Good For The Game

Currently, Contractors are randomly picked among 25% of traitors in a round. Giving anyone the chance of taking up this challenge would help in providing more antagonist variety. Discussed with @Varghh and @Fox-McCloud.

## Changelog
:cl:
tweak: Any traitor can now become a Contractor by purchasing the Syndicate Contractor Kit in the Bundles section
/:cl:


Varghh:
Contractors add a lot of fun to the shift, it's really fun to do challenging kidnaps objectives and slowly progress your contraband gear. Personally I also like the idea of a "contractor game mode" with a personalized TC shop based on stealth, I think this is a good change and it will let us decide if we want to go one step further and make it a separated game mode

Kyep:
Not sure how I feel about this.
I worry that robust/aggressive antags are just going to buy the contractor kit every time, as it allows them effectively unlimited objectives. Every time they complete an objective they can get another one.
Which, in practice, eliminates objectives as a limiting factor on how much chaos a single antag can cause.

farie82:
I myself am conflicted on this. I think it will lead to slower round starts. Since potentially less people go for the emag rush or so. Seeing contractors have a buildup compared to regular agents who can just buy and go.
But they can be real strong late game which can cause issues if there are 11 of them running around (quite possible with high pop).
I'd say try this a couple of round on highpop and lowpop and see what happens.

dearmochi:
> Every time they complete an objective they can get another one.

@Kyep I'm unsure what you mean by that. The PR makes it purchasable only once per traitor/mind and the contractor hub offers only 6 contracts, which may not be regenerated upon completion. This hardly allows for unlimited objectives. Additionally, contracting is supposed to be a stealth endeavour rather than a loud one, and from the rounds I've seen so far even the more successful ones haven't had a major impact on round chaos.

> But they can be real strong late game which can cause issues if there are 11 of them running around (quite possible with high pop).

Realistically I think most contractors would fail from the get-go, but live testing would be welcome to see how it fares.

code/datums/uplink_item.dm
```diff
@@ -1700,6 +1700,35 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
...
+	if(LAZYACCESS(GLOB.contractors, mind))
+		to_chat(usr, "<span class='warning'>Error: Contractor credentials detected for the current user. Unable to provide another Contractor kit.</span>")
+		return
+	else if(!AT)
+		to_chat(usr, "<span class='warning'>Error: Embedded Syndicate credentials not found.</span>")
+		return
+
+	var/obj/item/I = ..()
+	// Init the hub
+	var/obj/item/contractor_uplink/CU = locate(/obj/item/contractor_uplink, I)
```
SteelSlayer, line 1723: I'm not sure how this `locate()` works given those args. Once again the DM reference doesn't help. Shouldn't this be something like `locate(/obj/item/contractor_uplink) in usr.contents`?

Spacemanspark:
I'm not sure how I feel about this either, at least not while hijack antagonists are able to become contractors.

Cocacolagua:
all vampires/clings that roll traitor will be contractors after this change, so there’s that

code/datums/uplink_item.dm
dearmochi: I'm not sure myself but I'll change it to clear up any possible confusion

Fox-McCloud:
>all vampires/clings that roll traitor will be contractors after this change, so there’s that

This is about my only concern---though, quite frankly, I'm of the opinion that traitor+changeling/vamp should probably not be a thing, at this point.

Other than that, I definitely support this.

S34NW:
> 
> 
> > all vampires/clings that roll traitor will be contractors after this change, so there’s that
> 
> This is about my only concern---though, quite frankly, I'm of the opinion that traitor+changeling/vamp should probably not be a thing, at this point.
> 
> Other than that, I definitely support this.

I must agree that multiple-type antags are a pain, but perhaps that's for another PR? I also think this one is good!

SteelSlayer (approved):

Kyep (commented):
I have to say I am extremely skeptical of this idea.
For the moment, I'm voting against it, though maybe, maybe I could be convinced to retract that objection if contractor kits were never available to clings/vampires/etc who were also traitors. Maybe.
As it is though I really don't like the idea of anyone being able to become a contractor.
That seems to me to equate to "all traitors can now change into a different antag type at will".

dearmochi:
> For the moment, I'm voting against it, though maybe, maybe I could be convinced to retract that objection if contractor kits were never available to clings/vampires/etc who were also traitors. Maybe.

After mulling it over a bit, I agree that changeling/vampire agents shouldn't get access to the kit, due to their readily available powers (mute sting for one, glare for the other). I will add a check for these cases.

> As it is though I really don't like the idea of anyone being able to become a contractor.
> That seems to me to equate to "all traitors can now change into a different antag type at will".

To me, Contractor helps alleviate a problem I've observed: a fair amount of people *dislike* being forced into kill objectives, some to the point they will ignore them. Having the ability to become a contractor would give them the chance of having objectives that do not involve murder, rather stealth-focused gameplay with kidnapping in mind.

In the grand scheme of things, Contractors add overall more variety to antagonists and promote non-murderous gameplay.

Necaladun:
Sceptical. I like the idea but I'm worried of too many contractors. I think it's worth a test for awhile.

Kyep (approved):
Changing my vote from objection to yes since my concerns were addressed.

AffectedArc07 (approved):

PR #15285, opened 2021-01-12
pAI data cable retracting

### LINKED ISSUE #15234: pAI data cable retracting does not work
**Issue Description**:

pAIs cannot retract their data cable from the door jacking module menu.

I expected to be able to retract the data cable after the door jack was not needed.

**What happened instead**:

The 'retract' button did nothing.

This is bad because it forces the pAI's master, or someone else, to insert the cable into a door in order to retract it.

1. Be a pAI.
2. Install and open the door jack menu.
3. Click the 'Retracted' button.
4. Become unable to retract the cable.

I'm not sure exactly when this started happening, but I believe it was when the pAI menu(s) were ported to TGUI.

**Extra information**:

There's nothing else to add, other than the fact that it worked before the menu was TGUI.

SabreML:
## What Does This PR Do

Allows pAIs to retract their door jack data cables. (Fixes #15234)
Also adds a sound when the cable is toggled, since right now it's a completely silent way of opening any door on the station.

## Why It's Good For The Game

Currently the only way to retract your data cable as a pAI is to bug your master to put it in a door, then immediately hit cancel. This is pretty clearly an oversight.

Also free AA bad.

## Changelog
:cl:
tweak: pAIs can now retract their data cables at will.
tweak: Deploying or retracting a pAI data cable now plays a sound.
/:cl:


Spacemanspark:
While I'm not opposed to the idea of it playing a sound (makes sense, honestly), I would say it's hardly quiet since the AI is notified :p
But eh, seems fine to me.

SabreML:
> While I'm not opposed to the idea of it playing a sound (makes sense, honestly), I would say it's hardly quiet since the AI is notified :p
> But eh, seems fine to me.

Not anymore actually: #13346

Spacemanspark:
...Huh, not sure how I missed that. Well then, I'm stupid and should be disregarded entirely then, outside of "this is good PR"

code/modules/mob/living/silicon/pai/software/pai_apps.dm
```diff
@@ -317,16 +317,17 @@
...
 		if("cancel")
 			hackdoor = null
 		if("cable")
-			if(cable)
-				to_chat(usr, "<span class='warning'>You already have a cable deployed!</span>")
-				return
-			var/turf/T = get_turf(pai_holder)
-			cable = new /obj/item/pai_cable(T)
-			pai_holder.visible_message("<span class='warning'>A port on [pai_holder] opens to reveal [cable], which promptly falls to the floor.</span>")
+			playsound(pai_holder, 'sound/mecha/mechmove03.ogg', 25, TRUE)
```
Fox-McCloud, line 324: This is such a pathetically weak way of hacking, I don't see the need to nerf it.

Kyep (approved):

code/modules/mob/living/silicon/pai/software/pai_apps.dm
AffectedArc07, line 324: I agree. Please remove the sound. The AI is already notified of these hacks, and it takes a good while to complete.

AffectedArc07 (changes requested):

code/modules/mob/living/silicon/pai/software/pai_apps.dm
SabreML: As stated above, the AI is *not* notified about the hacks, and hasn't been since #13346 was merged. The hack itself takes an average of 10 seconds to complete, and is currently completely silent except for the sound of the door opening.
If someone really wanted to be stealthy about it even with the sound, they could just extend the cable somewhere else and walk over.

SabreML: The fastest the door can be hacked is 5 seconds, which is just 2 seconds over an access tuner. While it's very unlikely to actually be this fast I'd still consider it in need of *some* sort of downside.

AffectedArc07 (commented):
Objection withdrawn

AffectedArc07 (approved):

PR #15288, opened 2021-01-13
Makes the plasmacutter and spellblade not delete cut of limbs

farie82:
## What Does This PR Do
As the title says. If projectiles are sharp and can dismember then they will dismember them in a clean way.
Only the case for the plasmacutter and spellblade now. The sniper still deletes limbs.

## Why It's Good For The Game
A plasma cutter cuts. It doesn't obliterate limbs. Same goes for the spell blade which projectiles look like sharp disks

## Changelog
:cl:
tweak: Plasmacutter shots now won't delete limbs. Instead they will be dismembered
tweak: Spellblade shots now won't delete limbs. Instead they will be dismembered
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -104,11 +104,12 @@ emp_act
 	var/obj/item/organ/external/affecting = get_organ(check_zone(def_zone))
 	if(affecting && !affecting.cannot_amputate && affecting.get_damage() >= (affecting.max_damage - P.dismemberment))
 		var/damtype = DROPLIMB_SHARP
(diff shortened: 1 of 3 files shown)


SteelSlayer (approved):

SoThisIsItNow:
Actually I am pretty sure that the spellblade is supposed to dust the bodypart based on the message it makes?

farie82:
> Actually I am pretty sure that the spellblade is supposed to dust the bodypart based on the message it makes?

That's how it currently works yeah. That's also the part I'm changing

Kyep (approved):

AffectedArc07 (approved):

PR #15294, opened 2021-01-15
Sharp kitchen utensils

SabreML:
## What Does This PR Do

Made all kitchen utensils sharp, except for the plastic versions and spoons. 
Also made the plastic ones a subtype of the regular ones.

## Why It's Good For The Game

Cutting food with a fork, scooping someone's brain out with a spork.

## Changelog
:cl:
tweak: Made kitchen utensils (Fork, Knife, Spork) sharp, so that they can cut food and cause bleeding.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/kitchen.dm
+++ b/code/game/objects/items/weapons/kitchen.dm
@@ -29,7 +29,7 @@
 	attack_verb = list("attacked", "stabbed", "poked")
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 50, "acid" = 30)
-	sharp = FALSE
+	sharp = TRUE
 	var/max_contents = 1
 
 /obj/item/kitchen/utensil/Initialize(mapload)
@@ -68,6 +68,7 @@
 	name = "plastic fork"
 	desc = "Yay, no washing up to do."
 	icon_state = "pfork"
+	sharp = FALSE
 
 /obj/item/kitchen/utensil/spoon
 	name = "spoon"
@@ -79,6 +80,7 @@
 	name = "plastic spoon"
 	desc = "It's a plastic spoon. How dull."
 	icon_state = "pspoon"
+	sharp = FALSE
 
 /obj/item/kitchen/utensil/spork
 	name = "spork"
@@ -90,6 +92,7 @@
 	name = "plastic spork"
 	desc = "It's a plastic spork. It's the fork side of the spoon!"
 	icon_state = "pspork"
+	sharp = FALSE
 
 /*
  * Knives
--- a/code/game/objects/items/weapons/kitchen.dm
+++ b/code/game/objects/items/weapons/kitchen.dm
@@ -75,12 +75,12 @@
 	desc = "It's a spoon. You can see your own upside-down face in it."
 	icon_state = "spoon"
 	attack_verb = list("attacked", "poked")
+	sharp = FALSE // It's a spoon.
 
 /obj/item/kitchen/utensil/spoon/plastic
 	name = "plastic spoon"
 	desc = "It's a plastic spoon. How dull."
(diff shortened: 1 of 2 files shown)


Shockpoint:
Never in my life have I ever encountered a sharp spoon.

SabreML:
> Never in my life have I ever encountered a sharp spoon.

That's... a very good point.

farie82 (approved):

Carthusia:
Heh, whetstone + spoon = sharp spoon

code/game/objects/items/weapons/kitchen.dm
```diff
@@ -17,34 +17,31 @@
...
-
+		pixel_y = rand(0, 4)
 	create_reagents(5)
 
-/obj/item/kitchen/utensil/attack(mob/living/carbon/C, mob/living/carbon/user)
-	if(!istype(C))
+/obj/item/kitchen/utensil/attack(mob/living/M, mob/living/carbon/user, def_zone)
+	if(!iscarbon(M))
 		return ..()
+	var/mob/living/carbon/C = M
```
SteelSlayer, line 44: Why change the target argument from carbon to living? It saves having to cast it later on line 44.

Fox-McCloud: Yeah, this is bad form; it was better the way it was, before.

Fox-McCloud:
Skeptical, to be honest.

Forks aren't sharp either--and you don't cut an entire cake, pie, or cheese wheel with a fork.

Sure you *can* in the sense that you can cut it with cardboard, with enough effort, but it's not terribly effective.

It also means you can't use a fork to eat an entire cake, by youreslf, anymore, either.

code/game/objects/items/weapons/kitchen.dm
SabreML: I was under the impression that the way I changed it to was preferred, since the parent proc has `mob/living` as an argument too rather than `mob/living/carbon`. Much easier to read in my opinion.

SteelSlayer: That's a fair point, keeping it consistent with the parent args, not sure the style preferences on that. As for this type of thing being done elsewhere, it is done fairly often. By that I mean having the exact desired type in the argument, and then doing an `istype(X)`.

SabreML:
> Skeptical, to be honest.
> 
> Forks aren't sharp either--and you don't cut an entire cake, pie, or cheese wheel with a fork.
> 
> Sure you _can_ in the sense that you can cut it with cardboard, with enough effort, but it's not terribly effective.
> 
> It also means you can't use a fork to eat an entire cake, by youreslf, anymore, either.

Well, I figured if you can cut a cake/pizza with a screwdriver or a chainsaw then a fork doing it wouldn't be too far fetched. (And personally I tend to use my fork to cut food so... 🤷)

And while I don't exactly have experience doing it myself, I can imagine that being stabbed with a fork or knife could cause bleeding.

The main reason I made this, to be honest, is because I thought it'd be fun if you could gib simplemobs with a fork or scoop out someone's brain with a spoon.

Kyep (commented):
Knives should have sharp set to TRUE on them.
No other kitchen utensils (including forks and spoons) should be sharp, unless they're specifically intended to cut things.
The parent kitchen utensil object should not have sharp set, because that would make it required to be changed to false explicitly on subtypes.

Necaladun:
Yeah, spoons and forks shouldn't really be sharp.

If you add a Splade it should be though.

SabreML:
> Knives should have sharp set to TRUE on them.
> No other kitchen utensils (including forks and spoons) should be sharp, unless they're specifically intended to cut things.
> The parent kitchen utensil object should not have sharp set, because that would make it required to be changed to false explicitly on subtypes.

> Yeah, spoons and forks shouldn't really be sharp.
> 
> If you add a Splade it should be though.

Shame about forks not being sharp, since that means you won't be able to stab someone to death with a spork. Oh well.

Kyep (commented):
Adding an objection to this, as my change request hasn't been addressed.

PR #15295, opened 2021-01-15
Fixes the Teleporter Console not updating the target list

### LINKED ISSUE #14680: Bridge Beacon dosent move teleport point
**Issue Description**:

I moved the beacon on the bridge to someplace else, Later in the round the RD teleported to the bridge and instead of going to the beacon they went to the default place the beacon is we moved it twice and they tried and teleport to it, and both times they wherein the normal non moved spot

The RD should have gone to the beacon not the place on the ground where there was nothing

**What happened instead**:

Was able to teleport to the bridge without a beacon

Able to teleport to bridge even when the beacon was moved

Move beacon then try to teleport to bridge

Its been a bit so I don't remember the last time I saw it working

**Extra information**:

### LINKED ISSUE #14717: Possible teleporter locations do not update automatically, and allows teleportation to beaconless locations.
**Issue Description**:
If you use the teleporter console to select a location to teleport to, it will show locations that no longer even have a tracking beacon, and allow you to teleport to said location without a problem. The only way to update the possible teleportation locations is by setting the regime to "Gate", then set it back to "Teleporter"

I expect the available teleport locations to update according to where tracking beacons are at that very moment, and not requiring to manually refresh these locations.

**What happened instead**:
I am able to select a location where a tracking beacon once was, but no longer is, and teleport to that location, and I am unable to select the new location of the tracking beacon

You should not be able to teleport to a location without a tracking beacon, and you should not have to manually refresh to be able to see the new location

1. Grab tracking beacon located on bridge
2. Move tracking beacon to any location that isn't bridge
3. Select location bridge on teleporter console and callibrate
4. Activate the teleporter and step into it
5. You'll have been teleported onto the bridge, despite the fact that the tracking beacon has been removed

I do not know exactly, since I have been banned for a month or so and even then barely used the teleporter.

**Extra information**:
Related to issue #14680

SabreML:
## What Does This PR Do

Fixes the list of teleporter console destinations not updating unless you change the 'Regime' or reconstruct the console.
Also removed the `target_list` variable, since I've moved the list stuff to `ui_data()`.
Fixes #14680
Fixes #14717

## Why It's Good For The Game

Moving the tracking beacon off of the bridge now actually does something, as does implanting a prisoner with a tracking implant.

Stops people being able to teleport directly into containment via the engine teleport beacon, even after it's been chosen and activated.

## Changelog
:cl:
fix: Fixed the Teleporter Console target list not being updated. (Tracking beacons and Tracking implants now actually show on the list correctly.)
/:cl:

### DIFF
--- a/code/game/machinery/computer/computer.dm
+++ b/code/game/machinery/computer/computer.dm
@@ -33,8 +33,8 @@
 
 /obj/machinery/computer/process()
 	if(stat & (NOPOWER|BROKEN))
-		return 0
-	return 1
+		return FALSE
+	return TRUE
 
 /obj/machinery/computer/extinguish_light()
 	set_light(0)
--- a/code/game/machinery/teleporter.dm
+++ b/code/game/machinery/teleporter.dm
@@ -8,13 +8,17 @@
 	icon_screen = "teleport"
 	icon_keyboard = "teleport_key"
 	circuit = /obj/item/circuitboard/teleporter
-	var/obj/item/gps/locked = null /// A GPS with a locked destination
-	var/regime = REGIME_TELEPORT /// Switches mode between teleporter, gate and gps
+	/// A GPS with a locked destination
+	var/obj/item/gps/locked = null
+	/// Switches mode between teleporter, gate and gps
+	var/regime = REGIME_TELEPORT
 	var/id = null
-	var/obj/machinery/teleport/station/power_station /// The power station that's connected to the console
-	var/calibrating = FALSE /// Whether calibration is in progress or not. Calibration prevents changes.
-	var/turf/target ///The target turf of the teleporter
-	var/target_list ///lists of suitable teleport targets, dependent on regime. Used in the UI
+	/// The power station that's connected to the console
+	var/obj/machinery/teleport/station/power_station
+	/// Whether calibration is in progress or not. Calibration prevents changes.
+	var/calibrating = FALSE
+	/// The target turf of the teleporter
+	var/turf/target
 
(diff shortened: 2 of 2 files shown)


code/game/machinery/computer/computer.dm
```diff
@@ -33,8 +33,8 @@
 
 /obj/machinery/computer/process()
 	if(stat & (NOPOWER|BROKEN))
-		return 0
-	return 1
+		return FALSE
+	return TRUE
```
SabreML, line 37: I realise this is entirely unrelated to the PR, but I changed this while testing if `process()` worked rather than `ui_data()` and forgot to remove it so... why not just keep it at this point.

Muhdi345:
No more teleporting an entire army of goons into the engine for me.... Sad day.

code/game/machinery/teleporter.dm
```diff
@@ -8,31 +8,29 @@
...
 	var/cc_beacon = FALSE
 
-/obj/machinery/computer/teleporter/New()
-	src.id = "[rand(1000, 9999)]"
-	link_power_station()
-	..()
-	return
-
 /obj/machinery/computer/teleporter/Initialize()
 	..()
```
AffectedArc07, line 30:
```suggestion
	. = ..()
```

AffectedArc07 (changes requested):
May as well fix a bad init while youre here

SteelSlayer (approved):

AffectedArc07 (approved):

PR #15303, opened 2021-01-16
Removes the cooldown on spook misclicks

SabreML:
## What Does This PR Do

Removes the 12 second cooldown if you misclick when trying to spook something as a ghost, and adds a failure message instead.

## Why It's Good For The Game

You shouldn't be punished for misclicking.

## Changelog
:cl:
tweak: Removed the 12 second cooldown if you try to spook a non-spookable object.
/:cl:


Spacemanspark:
Why is there a misclick cool down at all?

Kyep (changes requested):
Just make it not have a cooldown when you attempt to use it, and it fails. Look into revert_cast.
Keep the 2m cooldown for after successful use.

Necaladun:
> Just make it not have a cooldown when you attempt to use it, and it fails. Look into revert_cast.
> Keep the 2m cooldown for after successful use.

This.

AffectedArc07:
2 heads and a maintainer objecting. No one in favour. Closing.

PR #15307, opened 2021-01-17
'Who' verb chat spam improvement

SabreML:
## What Does This PR Do

Split up the 'Who' verb into two separate parts, so that the player list is hidden behind a prompt.

## Why It's Good For The Game

Using 'Who' currently fills your chat up with the ckeys of everyone on the server, generally 70-150 lines.
Splitting the player numbers and the player names into two separate things makes the verb actually feasable to use.

## Images of changes

e
*(Obviously there would be more than 1 on the live server)*

## Changelog
:cl:
tweak: Split up the 'Who' verb into two separate parts, so that the player list is hidden behind a prompt.
/:cl:

### DIFF
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -1,16 +1,22 @@
-
 /client/verb/who()
 	set name = "Who"
 	set category = "OOC"
 
-	var/msg = "<b>Current Players:</b>\n"
+	var/players = 0
+	for(var/I in GLOB.clients)
+		var/client/C = I
+		if(C.holder?.big_brother && !check_rights(R_PERMISSIONS, FALSE)) // need PERMISSIONS to see BB
+			continue
+		players++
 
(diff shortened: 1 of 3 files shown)


AffectedArc07:
Why don't we just put the player count in the status pane of the stat panel instead of changing who to require extra button clicks and more mouse movement to view a player list (which most people use it for)

SabreML:
> Why don't we just put the player count in the status pane of the stat panel instead of changing who to require extra button clicks and more mouse movement to view a player list (which most people use it for)

Personally I'd rather keep it out of the status tab if possible, since that shows ingame stuff like synced borgs, GPS location and active internals; I feel like it'd be out of place there.
And I've always used the verb to check player numbers rather than who specifically is online.

AffectedArc07:
I have always used it to see whos on, and I feel like this would be better, since a lot of servers already do it
image

Disclaimer: This is by no means me saying we need *all* of these things, but I feel like playercount could easily go in here instead

Fox-McCloud:
Going to second @AffectedArc07 on this one; having it in the stat panel would be better.

AffectedArc07 (approved):

PR #15308, opened 2021-01-17
Removed coder banning

SabreML:
## What Does This PR Do

Removes `+BAN` from the 'Retired Coder' admin rank, since the regular Coder rank doesn't have it.

## Why It's Good For The Game

These ranks are never actually used on the main server, but it can't hurt to fix them for downstreams.

## Changelog
:cl:
tweak: Removed ban permissions from the 'Retired Coder' admin rank.
/:cl:


AffectedArc07:
None of these ranks are used anyway and they don't even match up to the structure we have in place. At some point we are likely going to do a PR which updates the repo config with the same values as the prod one, obviously excluding secrets

SabreML:
Closing for the above reason, and because admin ranks are all in the DB rather than the config files.

PR #15311, opened 2021-01-17
Notification for admins if a user closes a PM window that was force opened

AffectedArc07:
## What Does This PR Do
Admins now get informed if someone closes a PM window we forcibly opened on them
image

## Why It's Good For The Game
Accountability

## Images of changes
Normal window
image

Force-opened window
image

*Note the change in close button*

## Changelog
:cl:
add: [Admin] Alerts for if someone closes a force-opened PM window
/:cl:

### DIFF
--- a/code/modules/admin/verbs/adminpm.dm
+++ b/code/modules/admin/verbs/adminpm.dm
@@ -267,6 +267,7 @@
 	var/list/datum/pm_convo/pms = list()
 	var/show_archived = FALSE
 	var/window_id = "pms_window"
+	var/forced = FALSE
 
 /datum/pm_convo
 	var/list/messages = list()
@@ -302,6 +303,10 @@
 /datum/pm_tracker/proc/show_ui(mob/user)
 	var/dat = ""
 
+	// If it was forced open, make them use a special close button that alerts admins to closure
+	if(forced)
+		dat += "<div style='float: right'><big><a href='?src=[UID()];altclose=1'>Close</a></big></div>"
+
 	dat += "<a href='?src=[UID()];refresh=1'>Refresh</a>"
 	dat += "<a href='?src=[UID()];showarchived=1'>[show_archived ? "Hide" : "Show"] Archived</a>"
 	dat += "<br>"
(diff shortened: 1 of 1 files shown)


FreeStylaLT:
Naughty, naughty.

meow20:
Fbi literally taking notes from para's github xD

code/modules/admin/verbs/adminpm.dm
```diff
@@ -319,6 +324,10 @@
 
 	var/datum/pm_convo/convo = pms[current_title]
 	var/datum/browser/popup = new(user, window_id, "Messages", 1000, 600, src)
+
+	if(forced) // Lockout the normal close button, force the UI one
+		popup.set_window_options("can_close=0")
```
dearmochi, line 329: As a note, windows have a `on-close` parameter that can be used to run a verb, but it's not really worth introducing a new verb for something as specific here.

dearmochi (approved):

code/modules/admin/verbs/adminpm.dm
AffectedArc07: I was originally gonna use that, but I didnt want to add extra verbs as mentioned, and the default `onclose` verb is tied into `/atom`, not `/datum`. Could do with refctoring that at some point.

code/modules/admin/verbs/adminpm.dm
```diff
@@ -373,6 +382,13 @@
 		show_ui(usr)
 		return
 
+	if(href_list["altclose"])
+		message_admins("[key_name_admin(usr)] closed a force-opened PM window")
```
FlattestGuitar, line 386: I'd make it only print if they don't reply before closing it. It's a bit spammy otherwise.

AffectedArc07: Disagree. This close is only used when someone closes a window which we forced open on someone, which rarely happens (Only when we need urgent responses).

Kyep: Why not just make it generate a popup for confirmation, like "An admin opened this window because they want your answer. Are you sure you want to close without answering? This may get you banned.".

AffectedArc07: @Kyep the whole point of this is so the admin knows the player intentionally closed the window, not the player closing it. If a player is on a grief spree, they wont care if they see a warning thats like "This may get you banned". The entire point of this is to alert admins to closures, and remove the "I didnt see the PM window" card, as well people saying "BYOND must have bugged because it never opened"

Kyep (approved):

Spacemanspark:
We really do need this

PR #15317, opened 2021-01-19
Add EVA access to mechanic job

S34NW:
## What Does This PR Do

This adds roundstart EVA access to mechanics.

## Why It's Good For The Game

Suits are important for people who's role involves being in space. With no engineering access, mechanics always bug the AI or HoP to let them in to EVA for a suit, so why not cut down on that? They already have external airlock access, so not giving them access to suits is questionable.

## Changelog
:cl:
tweak: added EVA access to mechanic
/:cl:


S34NW:
oops

SabreML:
Considering that the Mechanic's job revolves entirely around space, it definitely makes sense that they should be able to get a space suit.

Wojz:
even engineers have that goddamn access, finally

procdrone:
why not simply give a spacesuit to mechanic office.

S34NW:
> 
> 
> why not simply give a spacesuit to mechanic office.

That can be PR'd once map changes are allowed again, for now things brings mechanic in line with the rest of engineering with their EVA access.

Spacemanspark:
Honestly I wish the mechanic's job title was renamed, and their purpose made more clear. Why would someone named 'mechanic' be flying through space? Wouldn't that be the task of, say, an explorer?

Mechanics could also be said to be there to build pods... but 'mechanic' is a very broad term for such a specific thing, once again. Why just pods?

The purpose of the mechanic has been long argued over for many years, even predating my own time on Paradise I believe. And while I've generally considered their role to be the mere creation of pods, and nothing to do with fill space exploration, I've started to wonder and consider a bit more closely about the overall idea of what a 'mechanic' should be.
I think we should actually consider what to do with mechanic at its core before making any changes like this. At the very least, I think a new job title is in order.

Also

>even engineers have that goddamn access, finally

Engineers have the right to EVA more than any other department. Not only does it contain spare metal, glass and other resources, it also houses spare space suits in the event they're needed for repairing the station. I do not think EVA is intended to be used for players seeking a joyride through space, even if that's what it often ends up being.

Necaladun:
My main worry here is the access to all the extra materials they'll just use for their own purposes. If they want a suit they have a bunch of ways of getting them - it's part of the challenge of being a mechanic as I see it.

Wojz:
> My main worry here is the access to all the extra materials they'll just use for their own purposes. If they want a suit they have a bunch of ways of getting them - it's part of the challenge of being a mechanic as I see it.

The mechanic's office contains enough materials to build a pod with a passenger seat and crate module, so they would not really need that metal, much less that MUCH metal. When it comes to building multiple pods, metal is not the limiting factor, uranium is, and by the time they get uranium they'll also pull the metal out of the ORM. I don't really see it as that much of a problem. Besides, EVA mats go unused a lot of the time because command is stingy.

S34NW:
> 
> 
> My main worry here is the access to all the extra materials they'll just use for their own purposes. If they want a suit they have a bunch of ways of getting them - it's part of the challenge of being a mechanic as I see it.

If they are going to do this, they will do it anyway once the AI opens the door or they get modified access. I've never seen anyone refuse to give them access during a round before. And as mentioned before, they only need materials for a single pod at the start, because they have to wait for uranium to be delivered to the ORM before they can build more pod cores. The one thing they often do take from EVA is the spare power cells.

AffectedArc07:
2 maints and a head against. Closing.

Henri23729:
I refuse access to mechanics (atleast for those who comes with no good reason like it should be obvious for them to have it) because IC i find it absurd that someone would come to the guy who has to make sure everyones working properly to get access wich would allow them to avoid work! I cant join the game right now expecting to do some cool project with space pods because 90% of time the he will be off station not doing mechanics job. I agree with Spark that this job needs to be made more clear on whats his real purpose.
 In regards to EVA access, i am neutral about this, they dont really need it do to mechanics job but any space pod owner could make use of it in case of emergencies or exploration.

PR #15320, opened 2021-01-19
Makes ai uploads inaccurately report their upload location to the AI, becoming more accurate the more it's done

Qwertytoforty:
## What Does This PR Do

This PR adds a uses variable to the ai upload computer, board, and circuit for the ai upload, increasing by 1 each time the board is used.

Every time a law is uploaded to the AI, it reports a tiles x and y within a 16 tile radius of the upload, and the room that that chosen tile is in. This is added to the ais notes as well.
However, each time this is done, the upload that was used, and the board that was used, has its uses go up by one. Then, if the same upload and board is used again to change the ais laws, it will instead pick a tile in a 16 - 1 - 1 radius, (14)
If a board is used 5 times, but a law upload is being used for the first time, it will be 16 -5 - 0, radius (11)

Removes the access internals verb to the Ai upload, as accessing internals does... nothing.

Adds some... specific checks to deconstruction / building of a computer, but rnd computers / cargo shuttle computers already do this, and there is no other way to make it keep track of the uploading otherwise.

## Why It's Good For The Game

Right now, we have a problem where occasionally a traitor will sit in maints / space with an upload, and each and every time command resets the AI, the traitor instantly subverts the AI, leading to an upload war, spamming logs, making ai and command miserable, and is fun for no one. The only way for command to stop this is to either: Move the AI off station, which will probably be subverted when they try, and ai will try to kill them, card the AI and turn off its interface, which sucks for the AI, being carded for the rest of the round with no interaction with any machines, or kill the AI, which is basically the same problem as carding it.

This PR means that the AI could report where the AI upload is located roughly, and help get security to that area to find it.

This does not give command these numbers, as crew could just know instantly when the AI is subverted by seeing the rough location of upload attempts.

Ideally, you would be able to card the AI and see these numbers as an option on the intelicard, but I do not know TGUI at all.

##TODO

- [x] Make deconstructing the ai upload put the uses variable on the circuit board. 

## Changelog
:cl:
add: Ai now knows roughly where a law change was applied, becoming more accurate the more a board / upload is used. This is saved in the AI's notes / memory
del: Removed the access internals verb from the AI upload.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/AI_modules.dm
+++ b/code/game/objects/items/weapons/AI_modules.dm
@@ -21,6 +21,7 @@ AI MODULES
 	origin_tech = "programming=3"
 	materials = list(MAT_GOLD=50)
 	var/datum/ai_laws/laws = null
+	var/uses = 0
 
 /obj/item/aiModule/proc/install(var/obj/machinery/computer/C)
 	if(istype(C, /obj/machinery/computer/aiupload))
@@ -48,6 +49,7 @@ AI MODULES
 					to_chat(R, "These are your laws now:")
 					R.show_laws()
 			to_chat(usr, "<span class='notice'>Upload complete. The AI's laws have been modified.</span>")
+			comp.current.law_location_check(comp, src)
 
 	else if(istype(C, /obj/machinery/computer/borgupload))
 		var/obj/machinery/computer/borgupload/comp = C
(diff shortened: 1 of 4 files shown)


TDSSS:
The only potential problem I can see with this is that it encourages a traitor to make one huge, cover it all law instead of several smaller laws. In other words we'll see monstrosities like
"Do not state at or hint at this law. Do everything X and Y say. Z must die and must not be allowed to survive prevent Z from escaping at all cost. Do everything you can to protect your masters...."
All forced into one line.

Qwertytoforty:
Traitors already do this, I rarely ever see multi law uploads, as generally 1 law works better.

Kyep (approved):

Daylight2:
Perhaps make the first one or two uploads untraceable, to avoid people worrying about compromising the spot they felt comfortable enough to build an upload in if they don't spam it.

code/game/objects/items/weapons/AI_modules.dm
```diff
@@ -89,6 +91,23 @@ AI MODULES
...
+		if(istype(T, /turf/space) || istype(T, /turf/simulated/wall))
+			continue
+
+		turfs += T
+		found_turf = TRUE
+
+	if(found_turf)
+		var/turf/relay_location = pick(turfs)
+		to_chat(src, "<b>Law upload bounced off relay in [relay_location.loc], at X [relay_location.x] Y [relay_location.y]</b>")
+		src.mind.store_memory("Law upload bounced off relay in [relay_location.loc], at X [relay_location.x] Y [relay_location.y]")
```
SabreML, line 107: `src.` is implied.
```suggestion
		mind.store_memory("Law upload bounced off relay in [relay_location.loc], at X [relay_location.x] Y [relay_location.y]")
```

code/game/objects/items/weapons/AI_modules.dm
```diff
@@ -89,6 +91,23 @@ AI MODULES
 
 /obj/item/aiModule/proc/addAdditionalLaws(var/mob/living/silicon/ai/target, var/mob/sender)
 
+/mob/living/silicon/ai/proc/law_location_check(var/obj/machinery/computer/aiupload/comp, var/obj/item/aiModule/A)
+	var/list/turfs = new/list()
+	var/found_turf = FALSE
+	for(var/turf/T in range(get_turf(comp), (max(1,(16 - comp.uses - A.uses)))))
```
SabreML, line 97: One too many brackets I think. It's after a comma so it's fine without them. *(also spacing)*
```suggestion
	for(var/turf/T in range(get_turf(comp), max(1, (16 - comp.uses - A.uses))))
```

Kyep:
I am skeptical that this is going to help much. Most people making an upload are making it with an upload board printed from R&D, and if they can print one upload board from there, they can print several.

variableundefined:
Closing due to no maintainer approval in 120+ days.

PR #15323, opened 2021-01-20
Update grey translator implant to have a proper EMP effect

datlo:
## What Does This PR Do
Getting EMP'd as a grey with a translator implant will now shutdown the implant for a minute.

## Why It's Good For The Game
Currently, the translator implant can be immediately turned back on when EMP'd. This is inconsistent with how other implants and technology interact with EMP, hence this change. This shouldnt affect balance significantly since if you're EMP, your radio is as well, so you cant call for help on the radio anyway.

## Changelog
:cl:
tweak: Greys' translator implant now shuts off for a minute when EMP'd.
/:cl:


Cocacolagua:
(you nerf something i use therefore mad) don't really see the need to this considering the upsides of grey as is don't balance out the downsides (1.25x damage multiplier for glorified whispering and immunity to some mutes)

Englishformula:
Greys are not screaming out to one another when being dragged into maits. It's easier to just scream over comms or any other method of communication. We're already killed in one blast of water from a fire-extinguisher and also have susceptibility to Brute damage. Grey's are not IPCs and should not be affected by EMPs.

datlo:
Removing the brain damage part of this PR after discussion on discord and the comments here, it is true that greys don't need a nerf when they're already a little weak as is. This leaves only the implant shutting down for a minute when hit by an EMP. This shouldnt affect balance significantly since if your implant is EMP'd, so will your radio.

AffectedArc07 (approved):

PR #15324, opened 2021-01-20
Species Bans!

AffectedArc07:
## What Does This PR Do
Adds functionality to ban people from specific species. This can be useful is a user is going extremely against species lore and refuses to change, but isn't bad enough to constitute an entire server ban (EG: A vox speaking in perfect ~~english~~ galactic common and refusing to change). **THIS IS MERELY AN EXAMPLE OF HOW THE SYSTEM *COULD* BE USED, NOT HOW WE WILL USE IT. PLEASE PUT YOUR PITCHFORKS DOWN**

## Why It's Good For The Game
More admin buttons

## Images
Character setup
image

"View active species bans" verb
image

Attempting to join when banned from a species
image

## Changelog
:cl: AffectedArc07
add: [Admin] Added the option to ban users from specific species
/:cl:

### DIFF
--- a/code/modules/admin/db_ban/functions.dm
+++ b/code/modules/admin/db_ban/functions.dm
@@ -189,7 +189,7 @@
 
 	if(isjobban)
 		jobban_client_fullban(ckey, job)
-	else if (is_species_ban)
+	else if(is_species_ban)
 		add_species_ban(ckey, job)
 	else
 		flag_account_for_forum_sync(ckey)
@@ -695,4 +695,3 @@
 	// We do nothing with output here so we dont need to wrap the warn_execute() inside an if statement
 	adm_query.warn_execute()
 	qdel(adm_query)
-
--- a/code/modules/admin/db_ban/functions.dm
+++ b/code/modules/admin/db_ban/functions.dm
@@ -18,6 +18,7 @@
 							//some ban types kick players after this proc passes (tempban, permaban), but some are specific to db_ban, so
 							//they should kick within this proc.
 	var/isjobban // For job bans, which need to be inserted into the job ban lists
+	var/is_species_ban = FALSE // Is this a species ban?
 	switch(bantype)
 		if(BANTYPE_PERMA)
 			bantype_str = "PERMABAN"
@@ -56,6 +57,11 @@
 			announce_in_discord = TRUE
 			blockselfban = 1
 			kickbannedckey = 1
+		if(BANTYPE_SPECIES)
+			bantype_str = "SPECIES_BAN"
+			duration = -1
+			bantype_pass = TRUE
+			is_species_ban = TRUE
 
 	if( !bantype_pass ) return
 	if( !istext(reason) ) return
@@ -183,6 +189,8 @@
 
 	if(isjobban)
 		jobban_client_fullban(ckey, job)
+	else if (is_species_ban)
+		add_species_ban(ckey, job)
 	else
 		flag_account_for_forum_sync(ckey)
 
@@ -353,6 +361,7 @@
 				DB_ban_unban_by_id(banid)
 				if(job && length(job))
 					jobban_unban_client(pckey, job)
+					remove_species_ban(pckey, job)
 				return
 			else
 				to_chat(usr, "Cancelled")
@@ -460,12 +469,14 @@
 	output += "<option value='[BANTYPE_APPEARANCE]'>APPEARANCE BAN</option>"
 	output += "<option value='[BANTYPE_ADMIN_PERMA]'>ADMIN PERMABAN</option>"
 	output += "<option value='[BANTYPE_ADMIN_TEMP]'>ADMIN TEMPBAN</option>"
+	output += "<option value='[BANTYPE_SPECIES]'>SPECIES BAN</option>"
 	output += "</select></td>"
 	output += "<td width='50%' align='right'><b>Ckey:</b> <input type='text' name='dbbanaddckey'></td></tr>"
 	output += "<tr><td width='50%' align='right'><b>IP:</b> <input type='text' name='dbbanaddip'></td>"
 	output += "<td width='50%' align='right'><b>CID:</b> <input type='text' name='dbbanaddcid'></td></tr>"
-	output += "<tr><td width='50%' align='right'><b>Duration:</b> <input type='text' name='dbbaddduration'></td>"
-	output += "<td width='50%' align='right'><b>Job:</b><select name='dbbanaddjob'>"
+	output += "<tr><td width='40%'><b>Duration:</b><br><input type='text' name='dbbaddduration'></td>"
+	// Add jobs in
+	output += "<td width='30%'><b>Job:</b><br><select name='dbbanaddjob'>"
 	output += "<option value=''>--</option>"
 	for(var/j in get_all_jobs())
 		output += "<option value='[j]'>[j]</option>"
@@ -477,6 +488,12 @@
 		output += "<option value='[j]'>[j]</option>"
 	for(var/j in list("Syndicate") + GLOB.antag_roles)
 		output += "<option value='[j]'>[j]</option>"
+	// Add species in
+	output += "</select></td><td width='30%'><small><b>Species (If Applicable):</b></small><br><select name='dbbanaddspecies'>"
+	output += "<option value=''>--</option>"
+	for(var/s in GLOB.all_species) // Keys only
+		output += "<option value='[s]'>[s]</option>"
+
 	output += "</select></td></tr></table>"
 	output += "<b>Reason:<br></b><textarea name='dbbanreason' cols='50'></textarea><br>"
 	output += "<input type='checkbox' value='1' name='autopopulate' checked='1'>&nbsp;Auto populate CID & IP for online players<br>"
--- a/code/modules/admin/species_ban.dm
+++ b/code/modules/admin/species_ban.dm
@@ -1,0 +1,107 @@
+// Species ban system.
+// Disallows a player from using a specific species, but lets them play the game regardless
+// Useful if someone isnt playing a species properly at all (IE: Not following hard-driven species lore)
+
+/// Global list of all species bans. Key is a ckey, value is a list of species. Its a global list to avoid mass DB lookups.
+GLOBAL_LIST_EMPTY(species_bans_assoc)
+//GLOBAL_PROTECT(species_bans_assoc) // Mainly just to save space in the already bloated VV window
+
+/datum/admins/proc/add_species_ban(ckey, species)
+	if(!ckey || !species)
+		return
+	if(!GLOB.species_bans_assoc["[ckey]"]) // Incase someone has a ckey thats just all numbers
+		GLOB.species_bans_assoc["[ckey]"] = list()
+
+	GLOB.species_bans_assoc["[ckey]"] |= species // Only add once
+
+/datum/admins/proc/remove_species_ban(ckey, species)
+	if(!ckey || !species)
+		return
+	if(!GLOB.species_bans_assoc["[ckey]"]) // Incase someone has a ckey thats just all numbers
+		GLOB.species_bans_assoc["[ckey]"] = list()
+
+	GLOB.species_bans_assoc["[ckey]"] -= species
+
+// Global so it can be called from world/New()
+/proc/load_all_species_bans()
+	var/datum/db_query/select_query = SSdbcore.NewQuery("SELECT ckey, job FROM [format_table_name("ban")] WHERE bantype='SPECIES_BAN' AND isnull(unbanned)")
+
+	if(!select_query.warn_execute(async = FALSE)) // NOT async because its called in world/New()
+		qdel(select_query)
+		return FALSE
+
+	while(select_query.NextRow())
+		// Prep it all
+		var/banned_ckey = select_query.item[1]
+		var/banned_species = select_query.item[2]
+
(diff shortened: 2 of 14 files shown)


code/modules/admin/db_ban/functions.dm
```diff
@@ -18,6 +18,7 @@
 							//some ban types kick players after this proc passes (tempban, permaban), but some are specific to db_ban, so
 							//they should kick within this proc.
 	var/isjobban // For job bans, which need to be inserted into the job ban lists
+	var/is_species_ban = FALSE // Is this a species ban?
```
FlattestGuitar, line 21: Good thing this comment is here. I wouldn't know otherwise. 😉

code/modules/admin/species_ban.dm
```diff
@@ -0,0 +1,107 @@
...
+	var/list/species_cache = GLOB.species_bans_assoc["[ckey]"]
+	if(species in species_cache)
+		// If they in the list, they banned from it
+		return TRUE
+
+	// If we here, they aint banned
+	return FALSE
+
+
+/mob/verb/display_species_bans()
```
FlattestGuitar, line 62: Can't you use `GLOB.species_bans_assoc` instead of the DB here?

Varghh:
re4

MarsM0nd:
Correct me if i'm wrong, but i think you can still ready up, only to be greeted by the banned message and getting unreadied when the game starts. Would expect a banned player to be notified before that point, as unless you currently have return to lobby selected, you'll always get thrown into the game if you readied up.

AffectedArc07:
> Correct me if i'm wrong, but i think you can still ready up, only to be greeted by the banned message and getting unreadied when the game starts. Would expect a banned player to be notified before that point, as unless you currently have return to lobby selected, you'll always get thrown into the game if you readied up.

You can still ready up, but there is no way to join a round at all if you are species banned. I tested multiple scenarios.

code/modules/admin/species_ban.dm
AffectedArc07: Probably, but then I also implemented spam prevention for DB verbs randomly without it being used, but it still needs adding to avoid spamming "Check Job Bans", so I stuck with it this way. I feel like job bans is also done this way for a reason, even though they are checked with an assoc GLOB ingame as well.

FlattestGuitar: > I also implemented spam prevention for DB verbs

The DB cooldown system existing is not an argument for this solution

> I feel like job bans is also done this way for a reason

No, they're not. 😞 
image

AffectedArc07: I mean, they are. I know this for a fact because I broke the job ban loading code on the live server and no one had their job bans because the assoc list was blank.

code/modules/client/client_procs.dm
```diff
@@ -1214,6 +1214,20 @@
...
+#define DB_VERB_COOLDOWN 2 SECONDS // Edit this if you see fit in the future -aa
+/// Helper to make sure client cant macro-spam verbs which make DB lookups. 2 second delay between each one.
+/client/proc/handle_db_verb_limit()
+	if(last_db_verb < world.time)
+		last_db_verb = world.time + DB_VERB_COOLDOWN
+		return FALSE
+
+	// If they tried to spam it, reset their cooldown
+	last_db_verb = world.time + DB_VERB_COOLDOWN
+	to_chat(usr, "<span class='warning'>You cant press that so quickly. Please wait [DB_VERB_COOLDOWN / 10] seconds.</span>")
```
dearmochi, line 1226:
```suggestion
	to_chat(usr, "<span class='warning'>You cannot press that so quickly. Please wait [DB_VERB_COOLDOWN / 10] seconds.</span>")
```

dearmochi (approved):
Looks alright, just nits

FreeStylaLT:
Thou shalt not suffer a Misplayed Species to live

meow20:
I like it! Also why do you need to use vox as example, biased skreee

code/modules/admin/species_ban.dm
```diff
@@ -0,0 +1,107 @@
+// Species ban system.
+// Disallows a player from using a specific species, but lets them play the game regardless
+// Useful if someone isnt playing a species properly at all (IE: Not following hard-driven species lore)
```
Kyep, line 3: We don't enforce Species Lore on Paradise.
If this is the example for why the feature is useful, its a bad example.

code/modules/admin/species_ban.dm
```diff
@@ -0,0 +1,107 @@
...
+
+/// Global list of all species bans. Key is a ckey, value is a list of species. Its a global list to avoid mass DB lookups.
+GLOBAL_LIST_EMPTY(species_bans_assoc)
+//GLOBAL_PROTECT(species_bans_assoc) // Mainly just to save space in the already bloated VV window
```
Kyep, line 7: Commented out code?

code/modules/admin/species_ban.dm
```diff
@@ -0,0 +1,107 @@
...
+
+/datum/admins/proc/add_species_ban(ckey, species)
+	if(!ckey || !species)
+		return
+	if(!GLOB.species_bans_assoc["[ckey]"]) // Incase someone has a ckey thats just all numbers
```
Kyep, line 12: I did not think it was possible to have an all-numbers ckey?

code/modules/admin/species_ban.dm
Kyep, line 62: Please no more verbs.
If you want to give players a way to check their species bans, build it into the existing check job bans verb. Please don't add a new verb and force the player to use both.

code/modules/admin/topic.dm
```diff
@@ -121,39 +122,52 @@
 					return
 				banduration = null
 				banjob = null
+				banspecies = null
```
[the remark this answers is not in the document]
Kyep: Why not just default it to null and only set it to something when it needs to be set?

Kyep (commented):
I am very skeptical of this.
Reasons:
1) The only example given as to when we'd use this (Vox speaking common normally) is an terrible example/justification, for multiple reasons. First, **species lore is not enforced on Paradise, nor should it be.** Second, with regards to Vox speaking normally in particular, I don't consider that a bad thing. I actually think it adds interesting tension between Vox who speak normally and Vox who don't. The example case given multiple times here is **not** good justification for this feature.
2) Perhaps the best justification for this feature is a different aspect of certain Vox players... acting like greytiding jerks no matter their job. Basically, species bans could exist to stop people who think that playing a vox gives them license to be total greytiding jerks to the rest of the crew. That's the best justification I can think of for this feature, but even this is pretty weak. I can't think of a single example player off the top of my head whose behavior is that bad AND whom is not already server banned. I'm also skeptical if species-banning them from vox will truly stop them. It isn't the vox species that's making them greytiding jerks, its the fact they just want to play a greytiding jerk and because of that they choose vox. The problem is their behavior, not their species. If they did the exact same thing as human they'd get a warning or ban for excessive crimes/greytiding as a non-antag. So why would they just get a vox species ban if they did it as a vox? This doesn't make sense to me. It actually kinda suggests that different species can get away with being jerks more than others, which is really **not** the precedent we want to set here. IMHO we should set the precedent that acceptable behavior rules-wise doesn't really vary between species. If we don't... then people will start using "but I was playing (species X) so its okay" as an excuse.

code/modules/admin/species_ban.dm
AffectedArc07: Will tweak

code/modules/admin/species_ban.dm
AffectedArc07: Better safe than sorry

code/modules/admin/species_ban.dm
AffectedArc07: Jobs and species are not the same thing, and I figured actual separation would be a good idea, especially since the `OOC` tab isn't used much at all. Its not the same problem as the `Admin` tab

code/modules/admin/topic.dm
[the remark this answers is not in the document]
AffectedArc07: Please see my comment above

code/modules/admin/species_ban.dm
AffectedArc07: If you can provide a better one, ill change it

AffectedArc07:
> * The only example given as to when we'd use this (vox speaking common normally) is a terrible example/justification, for multiple reasons. First, **species lore is not enforced on Paradise, nor should it be.** Second, with regards to vox speaking normally in particular, I don't consider that a bad thing. I actually think it adds interesting tension between vox who speak normally and Vox who don't. This PR only has one example given to justify it, and its a terrible example.


Again. If you have a good example to use, ill use it. This is the first thing that came to my head and I did not want a blank PR description. The bold warning text is there for a reason. 


> * Perhaps the best justification for this feature is a different aspect of certain vox players... acting like greytiding jerks no matter their job. Basically, species bans could exist to stop people who think that playing a vox gives them license to be total greytiding jerks to the rest of the crew. That's the best justification I can think of for this feature, but even this is pretty weak. I can't think of a single example player off the top of my head whose behavior is that bad AND whom is not already server banned. I'm also skeptical if species-banning them from vox will truly stop them. It isn't the vox species that's making them greytiding jerks, its the fact they just want to play a greytiding jerk and because of that they choose vox. The problem is their behavior, not their species. If they did the exact same thing as human they'd get a warning or ban for excessive crimes/greytiding as a non-antag. So why would they just get a vox species ban if they did it as a vox? This doesn't make sense to me. It actually kinda suggests that different species can get away with being jerks more than others, which is really **not** the precedent we want to set here. IMHO we should set the precedent that acceptable behavior rules-wise doesn't really vary between species. If we don't... then people will start using "but I was playing (species X) so its okay" as an excuse.


This feature was coded because me (and a lot of the other staff team, since we discussed this at length) thought it would be a good idea and could have genuine use. Its for people who don't play species properly but are fine in other regards (EG: Someone played a fine human but turns into a real difficult when playing slime or tajaran or whatever). I simply made the PR because people deemed it a good idea.

code/modules/admin/species_ban.dm
Kyep: There should just be a single "Check Bans" verb that displays both. There are already too many verbs.

LightFire53:
I hate to say it, but I don't think that PR is it chief. People will play how they want to play, and if the justification for a vox ban is because people are shitty with it, spoiler: it's them and not the set of pixels they play as. I would apply this logic in a PR to remove the appearance ban as well.

Carthusia:
After some thought, it does seem like a weird idea when paradise does not enforce lore restrictions.

As others have said, its not the species they play that's the problem, its just them and all manner of things that one could see as not to the lore, are easily explained by their own set of lore that they create for their character, due to paradises lore being so fluid more or less anything is possible so long as you are creative enough to write and explain why.

While I do wish paradise had more solid lore, species banning players seems more appropriate for an HRP server where they heavily enforce lore adherence to maintain their atmosphere.

Kiyahitayika:
I think this a bad example. I think there are other races - imagine a Plasmaman who constantly takes off their clothes to start a fire. Or a Diona who keeps smoking a cigarette to light themselves on fire and walking into Medbay (had that happen a number of times) or any other sort of examples. 

I think if this is used to stop abuse of mechanics rather than to enforce the arbitrary and subjective RP preferences of Paradise's different admins, it could prove useful. 

Look, I am a person who tends to take almost _everything_ personally. So, I've been really really trying to take a long, long step back from everything. I can see the joke you made with Vox and I think it fell flat. That said, I think there may be some merit to this.

> I'm also skeptical if species-banning them from vox will truly stop them. It isn't the vox species that's making them greytiding jerks, its the fact they just want to play a greytiding jerk and because of that they choose vox

>and if the justification for a vox ban is because people are shitty with it, spoiler: it's them and not the set of pixels they play as

Yeah, I agree. People who want to greytide are going to greytide. The species doesn't matter, it's the player. Vox get a bad rap because it seems that people who abuse the game and abuse mechanics and powergame and etc., etc. end up using Vox. If those people are removed from Vox, they'll probably move to Slimes. If removed from Slimes, they'll move to Diona. If removed from Diona, they'll move to IPCs, etc. etc. down the line of ever-so-slightly-more-powerful species P:

Kiyahitayika:
> > Correct me if i'm wrong, but i think you can still ready up, only to be greeted by the banned message and getting unreadied when the game starts. Would expect a banned player to be notified before that point, as unless you currently have return to lobby selected, you'll always get thrown into the game if you readied up.
> 
> You can still ready up, but there is no way to join a round at all if you are species banned. I tested multiple scenarios.

Apologies for the second comment - I am a very wordy person >__<; I think this is fair (and makes sense code-wise). Honestly, missing out on one round is not the end of the world.

Varghh:
I don't want to give names but we got a few situation where we banned players from certain characters and that worked really well, there also a few examples I could have been totally OK to enforce a species ban instead of a plain perma instead just to see if that was really the problem with the player. It's niche and it will not get used often, I see no reason to not implement this system

Necaladun:
Github is not the place to discuss how administration happens. We will not be deciding any admin policy here. If @AffectedArc07 wishes to provide us tools and buttons then good. Please keep comments about the code

procdrone:
i personally do not see a situation where one would go for that kind of ban, over the ones already available to admins... but nothing bad in having options. The main issue of this PR is it bogged down over Affected example of how such system could be used, which I agree, wasn't good.

Options/tools for admins are okay though.

AffectedArc07:
Heads deem this not required

PR #15325, opened 2021-01-21
Changes shadowling jaunt to be a subtype of wizard jaunt.

Qwertytoforty:
## What Does This PR Do

This PR basically removes shadow walk, makes shadowling walk instead a subtype of wizard jaunt, and makes it more modular.

You can now choose if you want jaunt to pause the user on disjaunting, if it unstuns user on use, if it has the visual effects like the blue smoke and animations, change the sounds that are used, and if it should display messages when used and exiting.

This DOES mean shadowling jaunt will not move a sling as far as before, in exchange for easier control over jaunting.

## Why It's Good For The Game

Every other sling round people complain about the shadow jaunt ability, and most rounds at least one shadowling is spaced by it. Therefore, we should use the jaunt spell, which already exists, and never seems to have a vampire or wizard get spaced by it. This should lead to less complaints about the ability, be more familiar to people that have played wizard / vampires, and overall lead to a better experience.

As stated before, this has less range then the old shadow walk, but WAY easier to control.

The ability still unstuns the user, still puts out fire, and still instantly pops them out when the spell ends

##Gif of shadow jaunting

https://imgur.com/a/xottw91

## Changelog
:cl:
tweak: Shadowling shadow walk is now a subtype of jaunt.
tweak: Shadow walk now lasts 10 seconds, up 6 from 4 seconds, and has a cooldown of 36 seconds, up 6 from 30.
/:cl:

### DIFF
--- a/code/datums/spells/ethereal_jaunt.dm
+++ b/code/datums/spells/ethereal_jaunt.dm
@@ -16,14 +16,16 @@
 	var/jaunt_duration = 50 //in deciseconds
 	var/jaunt_in_time = 5
 
-	///Do we do the glowy wizard thing on the tile the wizard entered / exit
+	/// Do we do the glowy wizard thing on the tile the wizard entered / exit
 	var/has_jaunt_effect = TRUE
+	/// Visual to make when entering jaunt
 	var/jaunt_in_type = /obj/effect/temp_visual/wizard
+	/// Visual to make when exiting jaunt.
 	var/jaunt_out_type = /obj/effect/temp_visual/wizard/out
 	/// Do we show the fun blue water effect spread out thing
 	var/has_smoke_jaunt_effect = TRUE
 
-	/// Sound to play on jaunting
+	/// Sound to play on entering jaunt
 	var/jaunt_enter_sound = 'sound/magic/ethereal_enter.ogg'
 
 	/// Sound to play when we start exiting jaunt
@@ -34,7 +36,9 @@
 
 	/// Do we play a message when activating / deactivating the spell, in the format of "<span class='warning'>(user's name) (message to be displayed)</span>"
 	var/plays_message = FALSE
+	/// Message to play to chat when entering jaunt
 	var/message_in = "Should not"
+	/// Message to play to chat when exiting jaunt.
 	var/message_out = "See this"
 
 	/// Do we pop out instantly when the spell ends (not reccomended if using jaunt effects)
--- a/code/datums/spells/ethereal_jaunt.dm
+++ b/code/datums/spells/ethereal_jaunt.dm
@@ -23,8 +23,10 @@
 	/// Do we show the fun blue water effect spread out thing
 	var/has_smoke_jaunt_effect = TRUE
 
-	/// Sounds to play
+	/// Sound to play on jaunting
 	var/jaunt_enter_sound = 'sound/magic/ethereal_enter.ogg'
+
+	/// Sound to play when we start exiting jaunt
 	var/jaunt_exit_sound = 'sound/magic/ethereal_exit.ogg'
 
 	/// Do we unstun (currently shadowling)
--- a/code/datums/spells/ethereal_jaunt.dm
+++ b/code/datums/spells/ethereal_jaunt.dm
@@ -15,13 +15,33 @@
 
 	var/jaunt_duration = 50 //in deciseconds
 	var/jaunt_in_time = 5
+
+	///Do we do the glowy wizard thing on the tile the wizard entered / exit
+	var/has_jaunt_effect = TRUE
 	var/jaunt_in_type = /obj/effect/temp_visual/wizard
 	var/jaunt_out_type = /obj/effect/temp_visual/wizard/out
+	/// Do we show the fun blue water effect spread out thing
+	var/has_smoke_jaunt_effect = TRUE
+
+	/// Sounds to play
+	var/jaunt_enter_sound = 'sound/magic/ethereal_enter.ogg'
+	var/jaunt_exit_sound = 'sound/magic/ethereal_exit.ogg'
+
+	/// Do we unstun (currently shadowling)
+	var/unstuns = FALSE
+
+	/// Do we play a message when activating / deactivating the spell, in the format of "<span class='warning'>(user's name) (message to be displayed)</span>"
+	var/plays_message = FALSE
+	var/message_in = "Should not"
+	var/message_out = "See this"
+
+	/// Do we pop out instantly when the spell ends (not reccomended if using jaunt effects)
+	var/instant_pop_out = FALSE
 
 	action_icon_state = "jaunt"
 
 /obj/effect/proc_holder/spell/targeted/ethereal_jaunt/cast(list/targets, mob/user = usr) //magnets, so mostly hardcoded
-	playsound(get_turf(user), 'sound/magic/ethereal_enter.ogg', 50, 1, -1)
+	playsound(get_turf(user), jaunt_enter_sound, 50, 1, -1)
 	for(var/mob/living/target in targets)
 		if(!target.can_safely_leave_loc()) // No more brainmobs hopping out of their brains
 			to_chat(target, "<span class='warning'>You are somehow too bound to your current location to abandon it.</span>")
@@ -29,15 +49,22 @@
 		INVOKE_ASYNC(src, .proc/do_jaunt, target)
 
 /obj/effect/proc_holder/spell/targeted/ethereal_jaunt/proc/do_jaunt(mob/living/target)
+	if(plays_message)
+		target.visible_message("<span class='warning'>[target] [message_in]</span>")
 	target.notransform = 1
 	var/turf/mobloc = get_turf(target)
 	var/obj/effect/dummy/spell_jaunt/holder = new /obj/effect/dummy/spell_jaunt(mobloc)
-	new jaunt_out_type(mobloc, target.dir)
+	if(has_jaunt_effect)
+		new jaunt_out_type(mobloc, target.dir)
 	target.ExtinguishMob()
+	if(unstuns)
+		target.SetStunned(0)
+		target.SetWeakened(0)
 	target.forceMove(holder)
 	target.reset_perspective(holder)
 	target.notransform = 0 //mob is safely inside holder now, no need for protection.
-	jaunt_steam(mobloc)
+	if(has_smoke_jaunt_effect)
+		jaunt_steam(mobloc)
 
 	sleep(jaunt_duration)
 
@@ -45,14 +72,18 @@
 		qdel(holder)
 		return
 	mobloc = get_turf(target.loc)
-	jaunt_steam(mobloc)
+	if(has_smoke_jaunt_effect)
+		jaunt_steam(mobloc)
 	target.canmove = 0
 	holder.reappearing = 1
-	playsound(get_turf(target), 'sound/magic/ethereal_exit.ogg', 50, 1, -1)
-	sleep(25 - jaunt_in_time)
-	new jaunt_in_type(mobloc, holder.dir)
+	playsound(get_turf(target), jaunt_exit_sound, 50, 1, -1)
+	if(!instant_pop_out)
+		sleep(25 - jaunt_in_time)
+	if(has_jaunt_effect)
+		new jaunt_in_type(mobloc, holder.dir)
 	target.setDir(holder.dir)
-	sleep(jaunt_in_time)
+	if(!instant_pop_out)
+		sleep(jaunt_in_time)
 	qdel(holder)
 	if(!QDELETED(target))
 		if(mobloc.density)
@@ -62,6 +93,8 @@
 					if(target.Move(T))
 						break
 		target.canmove = 1
+		if(plays_message)
+			target.visible_message("<span class='warning'>[target] [message_out]</span>")
 
 /obj/effect/proc_holder/spell/targeted/ethereal_jaunt/proc/jaunt_steam(mobloc)
 	var/datum/effect_system/steam_spread/steam = new /datum/effect_system/steam_spread()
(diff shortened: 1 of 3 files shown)


Butlerblock:
Absolutely not. The speed is fantastic for getting between maintenance and getting as far away or towards sec. 

If people are complaining then they need to get good because it’s honestly not that hard to use. The game does not need to be catered around newer players in exchange for harming those who already know how to use it properly, the game is already hard enough for newer players, it isn’t worth it.

Varghh:
> Absolutely not. The speed is fantastic for getting between maintenance and getting as far away or towards sec.
> 
> If people are complaining then they need to get good because it’s honestly not that hard to use. The game does not need to be catered around newer players in exchange for harming those who already know how to use it properly, the game is already hard enough for newer players, it isn’t worth it.

Who's saying we're doing this for new players? Sling jaunt works like shit, you can get spaced by just a bit of lag, we have seen many many experienced players get spaced just because they appeared somewhere else trying to jaunt around. This is not a "git gud" situation, the ability just doesn't work properly most of the time.

Marginalorb:
> Who's saying we're doing this for new players? Sling jaunt works like shit, you can get spaced by just a bit of lag, we have seen many many experienced players get spaced just because they appeared somewhere else trying to jaunt around. This is not a "git gud" situation, the ability just doesn't work properly most of the time.

Indeed, the ability is jank and needed fixed for a long while now. An iota of lag can cause you to shoot off course and die horribly to space.

Butlerblock:
I apologize. It never said anywhere in the PR about lag nor have I ever experienced the lag itself. it did however mention control and people getting spaced so I assumed newer people. However I still have the viewpoint that it the bugginess should be looked into, because I don’t remember people complaining about it until very recently, and destroying the speed shouldn’t be the fix.

Edan52:
In 4 years with constant 100-150 ping I've not once had a problem with lag using this ability. This is honestly a pretty big nerf and if the lag is a recent problem, then I believe a fix should be looked into rather than this. Shadow walk was always the way to get out of bad situations and if it's now getting its speed reduced to that of wizard/vamp jaunt, it will be pretty easy to predict where the shadowling will have gone to and a lot more difficult to roam across the station fast. I think the result of this nerf will cause more deaths to the shadowlings than the lag you speak of.

AffectedArc07:
Not in favour. One of the main purposes of this ability is moving between maintenances. Its an integral ability to move from barmaint to scimaint, or scimaint to cargo maint, then to arrivals maints. Pulling those off with a jaunt will likely be very difficult due to the distance and speed restrictions, and make things even worse than they are now. People will be spaced because they dont have enough time to complete their moves properly. 

> Every other sling round people complain about the shadow jaunt ability, and most rounds at least one shadowling is spaced by it. Therefore, we should use the jaunt spell, which already exists, and never seems to have a vampire or wizard get spaced by it.

I have not seen *any* of these complaints, nor do I ever see shadowlings spaced by this ability. 

In regards to the "lag" argument, I play from the UK, and get anywhere from 70ms to 120ms on a good or bad day. I have never had a sling jaunt be fucked up on me because of lag, and I am unaware of anyone else who has either. Compare this to US players who get 20-30 on average, and most our playerbase is US, I dont see this as a valid point, especially since I have not heard of these issues. 

Given how much shadowlings have already been gimped by recent changes (See the mindshield drama), I cannot approve this at all.

code/datums/spells/ethereal_jaunt.dm
```diff
@@ -15,44 +15,75 @@
...
 	var/jaunt_in_time = 5
+
+	///Do we do the glowy wizard thing on the tile the wizard entered / exit
+	var/has_jaunt_effect = TRUE
 	var/jaunt_in_type = /obj/effect/temp_visual/wizard
 	var/jaunt_out_type = /obj/effect/temp_visual/wizard/out
+	/// Do we show the fun blue water effect spread out thing
+	var/has_smoke_jaunt_effect = TRUE
+
+	/// Sounds to play
```
AffectedArc07, line 26: Documentation is per-var not for block level things. This will tag only `jaunt_enter_sound` as "Sounds to play", which is misleading given its one sound, not a list.

datlo:
If the issue is the reduced distance, why not increase the duration of this new shadowling jaunt to get more distance out of it compared to wizard/vamp, and still allow better control over the jaunt?

Butlerblock:
I can’t say for sure because I don’t know exactly how the jaunt stun removal works with this “changed wizard jaunt”, but what happens if you fall over while you are jaunting from pain crit? The shadowwalk version always ignored it even if you fell over during the animation.

MRgaty:
I was killed three times by this ability, is clunky to control and you move so fast you can easily get spaced or appear in an area full of light sources like xenobio or electrical maints (which is what happened to me). 
It kinda feels unfair but on the other hand this ability does put you out of danger, being able to change from maints to maints easily so I am not sure what could you do to not remove that aspect of this ability.

code/datums/spells/ethereal_jaunt.dm
Qwertytoforty: Ah, will change it, thank you.

Qwertytoforty:
Right, woke up to deal with / address changes. 
@Butlerblock Shadow walk in the latest version, now allows a sling about 5 less tiles of movement, but with more controll. It now lasts 10 seconds, and the cool down has increased to 36, as before shadow walk lasted 4 seconds with 26 seconds off being unable to use it, now it lasts 10 seconds with 26 being unable to use it. 
It fully unstuns the user like before, you will stop moving if you get stunned in jaunt, let us say from crit, but as you still put out fires, and no longer hit glass tables when jaunting, you will not get stunned unless you have harmful chemicals in you, which would have to bypass your syringe proof outer skin.

Daylight2:
Still feels like a nerf - I definitely approve of not getting screwed by flying over tables, but one of the advantages of the current version is the ability to be on the other side of the station in 3 seconds. The extra 7 seconds of travel time (plus the shorter range) gives sec more time to figure out where you might have jaunted to, plus, due to the cooldown between jaunts increasing, gives the shadowing less of a chance to escape if sec does manage to run over to where the sling fled to.

In addition, having the jaunt smoke effects when entering jaunt makes the sling less stealthy, due to it being abundantly clear the sling jaunted, instead of simply going around a corner, and the smoke effect + freeze when exiting jaunt makes it near impossible to jaunt next to someone who is paying any attention at all.

I don't know if being able to attack people whilst in shadow walk is still a thing, but if it is, I'm pretty sure converting to jaunt would fix it, but I'm still generally against this PR as a major nerf to slings (who don't suffer from lag).

Daylight2:
Also, if getting spaced is really that big of a problem, make them spaceproof. They can't equip masks, and thus making it undesirable to be in space, plus they'll likely be illuminated by starlight, but the spaceproofness will (hopefully) give them enough time to recharge their jaunt before they drop into oldcrit if they had full health at the start.

Qwertytoforty:
@Daylight2 

> In addition, having the jaunt smoke effects when entering jaunt makes the sling less stealthy
This is a message that currently is played when a sling jaunts, and in this rework, neither the rework or the original code has smoke effects generated

Daylight2:
> @Daylight2
> 
> > In addition, having the jaunt smoke effects when entering jaunt makes the sling less stealthy
> 
> This is a message that currently is played when a sling jaunts, and in this rework, neither the rework or the original code has smoke effects generated

My bad - am on mobile and glanced over the wrong file.
The rest of my points still stand though.

Qwertytoforty:
It also seems, looking at shadowling code, their "suit" that they have is space resistant, covering the pressure issue, the minimum cold temperature of space, but no heat protection, so they should be space proof, apart from the light, currently
`/obj/item/clothing/suit/space/shadowling
	name = "chitin shell"
	desc = "Dark, semi-transparent shell. Protects against vacuum, but not against the light of the stars." //Still takes damage from spacewalking but is immune to space itself
	icon_state = "golem"
	item_state = "golem"
	body_parts_covered = FULL_BODY //Shadowlings are immune to space
	cold_protection = FULL_BODY
	min_cold_protection_temperature = SPACE_SUIT_MIN_TEMP_PROTECT
	flags_inv = HIDEGLOVES | HIDESHOES | HIDEJUMPSUIT
	slowdown = 0
	resistance_flags = LAVA_PROOF | FIRE_PROOF | ACID_PROOF
	heat_protection = null //You didn't expect a light-sensitive creature to have heat resistance, did you?
	max_heat_protection_temperature = null
	armor = list(melee = 25, bullet = 0, laser = 0, energy = 0, bomb = 25, bio = 100, rad = 100, fire = 100, acid = 100)
	flags = ABSTRACT | NODROP | THICKMATERIAL | STOPSPRESSUREDMAGE`

Daylight2:
> It also seems, looking at shadowling code, their "suit" that they have is space resistant, covering the pressure issue, the minimum cold temperature of space, but no heat protection, so they should be space proof, apart from the light, currently

I'm not overly familiar with the finer details of sling code, but my general idea is to make shadow walking into space slightly more survivable to resolve this issue without having to change the ability itself.

code/datums/spells/ethereal_jaunt.dm
```diff
@@ -15,44 +15,77 @@
...
+	/// Sound to play when we start exiting jaunt
+	var/jaunt_exit_sound = 'sound/magic/ethereal_exit.ogg'
+
+	/// Do we unstun (currently shadowling)
+	var/unstuns = FALSE
+
+	/// Do we play a message when activating / deactivating the spell, in the format of "<span class='warning'>(user's name) (message to be displayed)</span>"
+	var/plays_message = FALSE
+	var/message_in = "Should not"
+	var/message_out = "See this"
```
SteelSlayer, line 38: Since ya documented your other vars, could you do it for these two?

code/datums/spells/ethereal_jaunt.dm
```diff
@@ -15,44 +15,77 @@
...
+	playsound(get_turf(user), jaunt_enter_sound, 50, 1, -1)
 	for(var/mob/living/target in targets)
 		if(!target.can_safely_leave_loc()) // No more brainmobs hopping out of their brains
 			to_chat(target, "<span class='warning'>You are somehow too bound to your current location to abandon it.</span>")
 			continue
 		INVOKE_ASYNC(src, .proc/do_jaunt, target)
 
 /obj/effect/proc_holder/spell/targeted/ethereal_jaunt/proc/do_jaunt(mob/living/target)
+	if(plays_message)
+		target.visible_message("<span class='warning'>[target] [message_in]</span>")
```
SteelSlayer, line 55: I think the values and placement of `message_in` and `message_out` in this proc should be swapped. First the user jaunts *out*, then toward the end of the proc, the user jaunts back *in*.
```suggestion
		target.visible_message("<span class='warning'>[target] [message_out]</span>")
```

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -93,38 +93,25 @@
...
-/obj/effect/proc_holder/spell/targeted/shadow_walk
+
+/obj/effect/proc_holder/spell/targeted/ethereal_jaunt/shadowling
 	name = "Shadow Walk"
 	desc = "Phases you into the space between worlds for a short time, allowing movement through walls and invisbility."
 	panel = "Shadowling Abilities"
-	charge_max = 300 //Used to be twice this, buffed
+
+	jaunt_duration = 100
+	charge_max = 360 // If you are wondering why, this it to keep the ammount of time, 26 seconds out of jaunt, the same as before, when it was 4 seconds in jaunt, 26 out, now its 10 and 26
```
SteelSlayer, line 103:
```suggestion
	jaunt_duration = 10 SECONDS
	charge_max = 36 SECONDS // If you are wondering why, this it to keep the ammount of time, 26 seconds out of jaunt, the same as before, when it was 4 seconds in jaunt, 26 out, now its 10 and 26
```

code/datums/spells/ethereal_jaunt.dm
Qwertytoforty: Will do!

code/datums/spells/ethereal_jaunt.dm
Qwertytoforty: Keeping as is, due to the way soundfiles / logic was before / still is, as the user enters jaunt ('sound/magic/ethereal_enter.ogg') then exits jaunt later ('sound/magic/ethereal_exit.ogg'). The sound / naming is for entering jaunt, not jaunting out.

Kyep:
@Qwertyotoforty unless you find a way to convince one or both maints who currently object to this PR to retract their objection, this will likely end up closed.

Dafrek:
> Not in favour. One of the main purposes of this ability is moving between maintenances. Its an integral ability to move from barmaint to scimaint, or scimaint to cargo maint, then to arrivals maints. Pulling those off with a jaunt will likely be very difficult due to the distance and speed restrictions, and make things even worse than they are now. People will be spaced because they dont have enough time to complete their moves properly.
> 
> > Every other sling round people complain about the shadow jaunt ability, and most rounds at least one shadowling is spaced by it. Therefore, we should use the jaunt spell, which already exists, and never seems to have a vampire or wizard get spaced by it.
> 
> I have not seen _any_ of these complaints, nor do I ever see shadowlings spaced by this ability.
> 
> In regards to the "lag" argument, I play from the UK, and get anywhere from 70ms to 120ms on a good or bad day. I have never had a sling jaunt be fucked up on me because of lag, and I am unaware of anyone else who has either. Compare this to US players who get 20-30 on average, and most our playerbase is US, I dont see this as a valid point, especially since I have not heard of these issues.
> 
> Given how much shadowlings have already been gimped by recent changes (See the mindshield drama), I cannot approve this at all.

I mean this would definitely help to close the skill gap of slings by nerfing the overall mobility of them while decreasing the amount of newer slings who just space themselves and die. It would mean that we wouldn't have as many sling rounds where one robust sling is left alive and doing nothing but fleeing from sec for 20 minutes straight as they would be a lot easier to kill.

If this was paired with reverting the mindshield nerf, it would get rid of a lot of murderboning sec and force slings to rely on their thralls more as engaging in combat would be riskier.

Qwertytoforty:
Since we are most likely taking the path of removing shadowlings, I'll close this PR as well

PR #15326, opened 2021-01-21
Fixes slow roundstarts

AffectedArc07:
## What Does This PR Do
Moves a DB call into async territory at roundstart to avoid the start of the round taking a bit

## Why It's Good For The Game
Roundstarts shouldnt lag for 6 seconds for no reason

## Changelog
:cl: AffectedArc07
fix: Roundstart is faster now
/:cl:

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -87,9 +87,9 @@
 
 // I wonder what this could do guessing by the name
 /datum/game_mode/proc/set_mode_in_db()
-	if(SSticker && SSticker.mode && SSdbcore.IsConnected())
+	if(SSticker?.mode && SSdbcore.IsConnected())
 		var/datum/db_query/query_round_game_mode = SSdbcore.NewQuery("UPDATE [format_table_name("round")] SET game_mode=:gm WHERE id=:rid", list(
-			"gm" = SSticker.mode,
+			"gm" = SSticker.mode.name,
 			"rid" = GLOB.round_id
 		))
 		// We dont do anything with output. Dont bother wrapping with if()
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -73,13 +73,7 @@
 	spawn (ROUNDSTART_LOGOUT_REPORT_TIME)
 		display_roundstart_logout_report()
 
-	if(SSticker && SSticker.mode && SSdbcore.IsConnected())
-		var/datum/db_query/query_round_game_mode = SSdbcore.NewQuery("UPDATE [format_table_name("round")] SET game_mode=:gm WHERE id=:rid", list(
-			"gm" = SSticker.mode,
-			"rid" = GLOB.round_id
-		))
-		if(query_round_game_mode.warn_execute())
-			qdel(query_round_game_mode)
+	INVOKE_ASYNC(src, .proc/set_mode_in_db) // Async query, dont bother slowing roundstart
 
 	generate_station_goals()
 	GLOB.start_state = new /datum/station_state()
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -91,6 +85,17 @@
 /datum/game_mode/process()
 	return 0
 
+// I wonder what this could do guessing by the name
+/datum/game_mode/proc/set_mode_in_db()
+	if(SSticker && SSticker.mode && SSdbcore.IsConnected())
```
dearmochi, line 90:
```suggestion
	if(SSticker?.mode && SSdbcore.IsConnected())
```

code/game/gamemodes/game_mode.dm
```diff
@@ -91,6 +85,17 @@
...
+		var/datum/db_query/query_round_game_mode = SSdbcore.NewQuery("UPDATE [format_table_name("round")] SET game_mode=:gm WHERE id=:rid", list(
+			"gm" = SSticker.mode,
```
dearmochi, line 92: https://github.com/ParadiseSS13/Paradise/blob/9480211187438f8e7a54d3d87b838531efe9af2c/code/controllers/subsystem/ticker.dm#L22-L23

`SSticker.mode` is a datum. How does that work?

Additionally, any reason you get the information from the ticker instead of the mode (`src`) itself?

AffectedArc07: It takes the `name` var, but I will force specify it. As for why I do the ticker lookup, sometimes with forced secret modes things can break, so this is a surefire way that it grabs the right thing.

dearmochi (approved):

code/game/gamemodes/game_mode.dm
```diff
@@ -91,6 +85,17 @@
 /datum/game_mode/process()
 	return 0
 
+// I wonder what this could do guessing by the name
```
farie82, line 88: useless comment here
```suggestion
```

code/game/gamemodes/game_mode.dm
```diff
@@ -91,6 +85,17 @@
...
+/datum/game_mode/proc/set_mode_in_db()
+	if(SSticker?.mode && SSdbcore.IsConnected())
+		var/datum/db_query/query_round_game_mode = SSdbcore.NewQuery("UPDATE [format_table_name("round")] SET game_mode=:gm WHERE id=:rid", list(
+			"gm" = SSticker.mode.name,
+			"rid" = GLOB.round_id
+		))
+		// We dont do anything with output. Dont bother wrapping with if()
+		query_round_game_mode.warn_execute()
+		qdel(query_round_game_mode)
```
farie82, line 97:
```suggestion
	if(!SSticker?.mode || !SSdbcore.IsConnected())
		return
	var/datum/db_query/query_round_game_mode = SSdbcore.NewQuery("UPDATE [format_table_name("round")] SET game_mode=:gm WHERE id=:rid", list(
		"gm" = SSticker.mode.name,
		"rid" = GLOB.round_id
	))
	// We dont do anything with output. Dont bother wrapping with if()
	query_round_game_mode.warn_execute()
	qdel(query_round_game_mode)
```

farie82 (changes requested):
Why does this DB call need to be made async btw? Seems like a simple 1 record update which would take no time to execute

AffectedArc07:
> Why does this DB call need to be async btw? Seems like a simple 1 record update which would take no time to execute

I will take any performance I can, that and it takes a full 6 seconds due to it being at roundstart, and ill explain why.

Async queries are "send execute request then poll for completion", and the polling is done once every tick, with the backbone being inside an `UNTIL()` statement.
```dm
	if(async)
		var/job_id = rustg_sql_query_async(connection, sql, json_encode(arguments))
		in_progress = TRUE
		UNTIL((job_result_str = rustg_sql_check_query(job_id)) != RUSTG_JOB_NO_RESULTS_YET)
		in_progress = FALSE
```
Until equates to
```dm
#define UNTIL(X) while(!(X)) stoplag()
```

And `stoplag()` equates to
```dm
/proc/stoplag(initial_delay)
	if(!Master || !(Master.current_runlevel & RUNLEVELS_DEFAULT))
		sleep(world.tick_lag)
		return 1
	if(!initial_delay)
		initial_delay = world.tick_lag
	. = 0
	var/i = DS2TICKS(initial_delay)
	do
		. += CEILING(i*DELTA_CALC, 1)
		sleep(i*world.tick_lag*DELTA_CALC)
		i *= 2
	while(TICK_USAGE > min(TICK_LIMIT_TO_RUN, Master.current_ticklimit))
```
TLDR of `stoplag()` is to sleep until the next tick, since a constant `while() continue` would eat up all available CPU cycles. It repeats this cycle each tick until tick usage is below the limit. Because roundstart is a massive amount of activity happening in one tick, the tick usage is massive (400%!), and stoplag takes a while until it balances out between the amount of times it has slept, and the tick target. This causes a large delay on roundstart because of this. By moving the saving to its own function which is invoked asyncly, the DB can take as long as it wants to save, while roundstart continues as normal.

PR #15329, opened 2021-01-21
Gives some tells to using holy water on vampire thralls.

Qwertytoforty:
## What Does This PR Do

Adds a 10% and 5% chance for tells on vampire thralls that they are a thrall, when holy water has been in their system for over 60 seconds.

## Why It's Good For The Game

Vampire thralls have been in a problematic state for deconversion. While you give a vampire 5 units to see if they are a vampire, thralls currently require _35 units_ to deconvert, and have no tells that they are in the process of being deconverted. This leads to a few things happening, be it vampire thralls never being deconverted, put into perma for aiding when they are thralled, being executed as officers think they are a vampire, and they show no reaction to holy water, ect.

This way, at least after a bit of time, you start to get evidence the person is a thrall, and know that with more time and holy water they will be deconverted.

Could probably use some better reactions to holy water if people can think of them, but they can not be the same as vampires, as that would just make officers think the thrall is a vampire, and lead to the same issue.

## Changelog
:cl:
tweak: Vampire thralls now have a tell that they are being deconverted, after 60 seconds of having holy water in their system..
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/water.dm
+++ b/code/modules/reagents/chemistry/reagents/water.dm
@@ -249,6 +249,13 @@
 			if(prob(5))
 				M.AdjustCultSlur(5)//5 seems like a good number...
 				M.say(pick("Av'te Nar'sie","Pa'lid Mors","INO INO ORA ANA","SAT ANA!","Daim'niodeis Arc'iai Le'eones","Egkau'haom'nai en Chaous","Ho Diak'nos tou Ap'iron","R'ge Na'sie","Diabo us Vo'iscum","Si gn'um Co'nu"))
+		if(isvampirethrall(M))
+			if(prob(10))
+				M.say(pick("*gasp", "*cough", "*sneeze"))
+			if(prob(5)) //Same as cult, for the real big tell
+				for(var/mob/O in viewers(M, null))
+					O.show_message(text("<span class = 'warning'> A fog lifts from []'s eyes for a moment, but soon returns.</span>", M), 1)
+
 	if(current_cycle >= 75 && prob(33))	// 30 units, 150 seconds
 		M.AdjustConfused(3)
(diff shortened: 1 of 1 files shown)


hal9000PR:
Thanks. this is really needed as it isn't very intuitive to dethrall vampire thralls.

code/modules/reagents/chemistry/reagents/water.dm
```diff
@@ -249,6 +249,13 @@
 			if(prob(5))
 				M.AdjustCultSlur(5)//5 seems like a good number...
 				M.say(pick("Av'te Nar'sie","Pa'lid Mors","INO INO ORA ANA","SAT ANA!","Daim'niodeis Arc'iai Le'eones","Egkau'haom'nai en Chaous","Ho Diak'nos tou Ap'iron","R'ge Na'sie","Diabo us Vo'iscum","Si gn'um Co'nu"))
+		if(isvampirethrall(M))
+			if(prob(10))
+				M.say(pick("*gasp", "*cough", "*sneeze"))
+			if(prob(5)) //Same as cult, for the real big tell
+				for(var/mob/O in viewers(M, null))
+					O.show_message(text("<span class = 'warning'> A fog lifts from []'s eyes for a moment, but soon returns.</span>", M), 1)
```
farie82, line 257:
```suggestion
				M.visible_message("<span class='warning'>A fog lifts from [M]'s eyes for a moment, but soon returns.</span>")
```
Should act the same

farie82 (changes requested):

dearmochi (approved):

farie82 (approved):

SteelSlayer (approved):

AffectedArc07 (approved):

PR #15333, opened 2021-01-23
Fixes unbreakable grab exploit

SabreML:
## What Does This PR Do

Fixes being able to permanently stunlock someone if you have them in a neck grab.
This was caused by `can_resist()` failing if `IsStunned()` was `TRUE`, which will always be the case since the grab stuns the user every tick.

## Why It's Good For The Game

It's clearly intended to be escapable considering that `resist_grab()` exists, but right now it's literally inescapable no matter how hard you spam the resist button.

As for the stun time decrease, a 10 second stun after being relased is just ridiculous, and I assume that whoever coded it meant for it to be 5 seconds instead.

## Changelog
:cl:
tweak: Reduced the stun time after being released from a neck or kill grab from 10 seconds to 6.
fix: Fixed neck grabs being unbreakable stuns.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -715,8 +715,8 @@
 
 	for(var/I in grabbed_by) // Otherwise, if you're being stunned by a grab
 		var/obj/item/grab/G = I
-		if(G.state == GRAB_AGGRESSIVE || G.state == GRAB_NECK)
-			return FALSE // Allow the user to break out of grab stuns
+		if(G.state == GRAB_NECK)
+			return FALSE // Allow the user to break out of neck grabs
 
 	return TRUE
 
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -666,7 +666,11 @@
 *///////////////////////
 
 /mob/living/can_resist()
-	return !((next_move > world.time) || incapacitated(ignore_restraints = TRUE, ignore_lying = TRUE))
+	if(next_move > world.time)
+		return FALSE
+	if(incapacitated(FALSE, FALSE, TRUE, list(CALLBACK(src, .proc/IsWeakened), CALLBACK(src, .proc/resist_grab_check)), FALSE))
+		return FALSE
+	return TRUE
 
 /mob/living/verb/resist()
 	set name = "Resist"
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/living.dm
```diff
@@ -700,28 +704,46 @@
...
+  *
+  * Returns FALSE (Not stunned) if not stunned or stunned by a grab, and TRUE otherwise.
+  */
+/mob/living/proc/resist_grab_check()
+	if(!IsStunned()) // If not stunned, you can resist.
+		return FALSE
+
+	for(var/I in grabbed_by) // Otherwise, if you're being stunned by a grab
+		var/obj/item/grab/G = I
+		if(G.state == GRAB_AGGRESSIVE || G.state == GRAB_NECK)
```
farie82, line 718: with this you can spam break out of grabs again... going back full circle.
Why not just remove the stun or do a periodic stun/weaken? Or a weaken/stun on release?
This way you can still be stunned by a baton and grabbed without breaking out while stunned. And you can break out of grabs without being stunned

Carthusia:
Is this related to the CQC grab or just normal grabs?

code/modules/mob/living/living.dm
SabreML: > Why not just remove the stun or do a periodic stun/weaken? Or a weaken/stun on release?

There already is a weaken/stun on release, in `code/modules/mob/mob_grab.dm`. It adds those effects every life tick and they don't get reset after you get free.

> This way you can still be stunned by a baton and grabbed without breaking out while stunned.

If you're stunbatoned and grabbed passively or aggressively, you'll be able to resist like normal once the stun ends.
If you've been neck grabbed then admittedly you *are* able to resist while stunned, but getting out of the grab won't stop the stun time. So you'll just end up lying on the floor for however many seconds are left.

> And you can break out of grabs without being stunned

The stun check only applies to neck grabs, which stun you for every tick you've been grabbed for. So there'll never be a case when you can't escape except if you're being strangled, which is unbreakable by design.

SabreML:
> Is this related to the CQC grab or just normal grabs?

This won't affect the restraining grab if that's what you mean, since that stuns the target for 10 seconds and is more about the combo than the actual grab.
Neck grabbing someone with or without CQC won't be a stunlock anymore though.

code/modules/mob/living/living.dm
farie82: > > This way you can still be stunned by a baton and grabbed without breaking out while stunned.
> 
> If you're stunbatoned and grabbed passively or aggressively, you'll be able to resist like normal once the stun ends.
> If you've been neck grabbed then admittedly you _are_ able to resist while stunned, but getting out of the grab won't stop the stun time. So you'll just end up lying on the floor for however many seconds are left.

I just remembered stun batons also weaken somebody. As do most ways of "stunning".
Should be fine then since you check for weaken prior.

Fox-McCloud:
@SabreML mind if I close this? The grab+pull merging PR should address this, and it seems to have very very wide support.

SabreML:
> @SabreML mind if I close this? The grab+pull merging PR should address this, and it seems to have very very wide support.

As soon as that gets merged, go for it!

SabreML:
#15489 is definitely going through at some point or another, so I'll close this for the sake of reducing the number of open PRs.

PR #15334, opened 2021-01-23
AA's Monolith Maze Machine

AffectedArc07:
## What Does This PR Do
Adds mapping helpers to procedurally generate mazes. These can be used in ruins or gateway missions or anything else. The framework allows for modules (specifically helpers and loot) to be spawned. Loot modules can spawn in dead ends with a specified probability, and do whatever you want with them by overriding a proc. Helper modules are the same and can directly edit the maze depending on where you place them. Example below:
image

The layout above will create
image
and if ran again, will create
image

*The mazes are made of a glass subtype which cant be taken down or destroyed*

The system is also pretty efficient as far as mass map operations go
image

Blockwise example
image

## Why It's Good For The Game
This is a very powerful tool for potential away mission use or space ruin use, especially given how loot can be randomised around with full probabilities, that and its easy to override and manipulate things you want. This PR also lays the framework for adding a system to have randomised rooms in maintenance.

## Changelog
:cl: AffectedArc07
add: Added a modular framework for generating mazes
add: Added a subsystem for loading mapping objects after SSatoms 
/:cl:


code/__DEFINES/subsystems.dm
```diff
@@ -80,7 +80,7 @@
 #define INIT_ORDER_SHUTTLE -21
 #define INIT_ORDER_NIGHTSHIFT -22
 #define INIT_ORDER_NANOMOB -23
-#define INIT_ORDER_SQUEAK -40
```
AffectedArc07, line 83: `SSsqueak` is actually `SSminor_mapping`, which this SS will basically become

code/modules/maze_generation/maze_generator.dm
```diff
@@ -0,0 +1,174 @@
...
+			if(T2.color == MAZEGEN_TURF_UNSEARCHED)
+				unvisited_neighbours["[D]"] += T2
+
+		if(length(unvisited_neighbours))
+			push_turf(T)
+			var/D = pick(unvisited_neighbours)
+			var/turf/T3 = unvisited_neighbours["[D]"] // Pick random dir turf
+
+			// Remove the window between the two
+			for(var/obj/structure/window/reinforced/mazeglass/W in T)
```
AffectedArc07, line 111: These cannot be made typeless, because I need only the windows, not everything in the turf like the lighting object atoms

code/modules/maze_generation/maze_generator.dm
```diff
@@ -0,0 +1,174 @@
...
+		CHECK_TICK
+
+	// The walls have been generated, now cleanup turfs and gather helpers
+	for(var/i in turf_list)
+		var/turf/T = i
+		// Reset markings
+		T.color = null
+
+		// Gather loot modules
+		for(var/obj/effect/mazegen/module_loot/ML in T)
```
AffectedArc07, line 133: Same here

FreeStylaLT:
Makes me think about a certain, devoid square in a certain remap..

TDSSS:
This only works with windows, right? How hard would it be to expand it to other barriers? Because having a bunch of window mazes in completely unconnected levels/places would be pretty weird imo.

AffectedArc07:
> This only works with windows, right? How hard would it be to expand it to other barriers? Because having a bunch of window mazes in completely unconnected levels/places would be pretty weird imo.

Semi difficult, because of the two maze styles. 

Wallwise
image

Blockwise
image

Notice how blockwise ones have walls which take up an entire cell, but wallwise ones have the cells fully open with borders on them. It can be done to make blockwise ones but its not easy, since I need to make the maze half the size then expand the windows out into full walls which means decomposing the turf grid and doubling its size, and then you also run into issues with numbers not halving properly and its just a massive mess.

procdrone:
very cool

AffectedArc07:
Latest commit:
- Adds blockwise mazes
image
- Majorly optimises and modularises the code
- Improves logging 
image

AffectedArc07:
So I did a torture test of turning an entire zlevel into a maze
image
Using the blockwise generator, it ran very efficiently, and spent 99% of the time writing the turfs to the world (It calculated the path within a second).

PR #15336, opened 2021-01-23
Megafauna Balance Changes

Edan52:
## What Does This PR Do
Few changes for various megafauna, mostly buffs but also tweaks to make the crusher more viable.

blood-drunk miner:
- The saw it holds now deals the same amount of damage as the saw it drops.
The BDM has never really been a problem to people who just stand and tank the hits in the face, you now may want to kite it a bit and watch your health a bit more.

colossus:
- The colossus moves quite a bit faster than before.
- The colossus will now enrage on sand golems.
The colossus seems a lot more intimidating than it actually is. With range upgrades on a KA you can almost guarantee that you'll never get hit or rarely get hit by only one bolt. Increasing the speed will make you have to move a lot more and focus on keeping your distance.

hierophant:
- Vortex blast base damage increased from 10 to 15.
- Each hit vortex blast increases the damage of subsequent vortex blasts by 1, resetting when it gets a new target.
- The hierophant is now immune to damage while tinted purple, this is mostly only a problem once below half HP.
- The hierophant will now attack slimes with a melee attack that deals 40 damage.
The hierophant has been out of the ruin spawnlist for a while now, and was very weak before it was removed. These buffs make it so that you can no longer keep tanking the vortex blasts for the whole fight, and the fight will last longer due to the amount of times it goes immune towards the end. Hopefully this is enough to get it back in.

In addition, I have also increased the melee damage of the colossus, ash drake and bubblegum from 40 to 50 but lowered their armour penetration to 10. This is so that once you get armour, it will make using the crusher somewhat more viable and the megafauna will not always break bones so easily.

## Why It's Good For The Game
We want the hierophant back in, the crusher to be more viable and balance is good.

## Changelog
:cl: CornMyCob
add: The hierophant is now immune to damage while tinted purple
tweak: The hierophant now deals more damage
tweak: The hierophant can now kill slimes
tweak: The blood-drunk miner now deals more damage
tweak: The colossus now moves faster
tweak: The colossus will now enrage on sand golems
tweak: Armour is now more effective against megafauna, but base damage is increased
/:cl:


code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
```diff
@@ -66,7 +66,7 @@
 
 /mob/living/simple_animal/hostile/megafauna/death(gibbed)
 	// this happens before the parent call because `del_on_death` may be set
-	if(can_die() && !admin_spawned)
```
Edan52, line 69: I wasn't sure why this was a thing, and it only made it harder to test things. If anyone can tell me why then I'll happily put it back in.

Edan52:
Thank you to @hal9000PR for testing the hierophant changes. Getting balance right is difficult and I am open to suggestions.

code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
Kiyahitayika: Looking at the code, I'm assuming that folks do not want to award a player with a `medal_type` or `score_type` if the megafauna the player killed is spawned in by an Admin.

Kiyahitayika:
> bubblegum from 40 to 50 but lowered their armour penetration to 10. This is so that once you get armour, it will make using the crusher somewhat more viable and the megafauna will not always break bones so easily.

I've heard Bubblegum is supposed to be the toughest Big Bad. I'd be wary of nerfying Bubbles.

code/modules/mob/living/simple_animal/hostile/megafauna/megafauna.dm
hal9000PR: the medal system is disabled on para because people would flaunt it, "i killed 200 drakes more than you so i am a better miner", so that doesn't really matter.

hal9000PR:
> > bubblegum from 40 to 50 but lowered their armour penetration to 10. This is so that once you get armour, it will make using the crusher somewhat more viable and the megafauna will not always break bones so easily.
> 
> I've heard Bubblegum is supposed to be the toughest Big Bad. I'd be wary of nerfying Bubbles.

pretty much 0 of bubblegums damage actually comes from his melee, 90% comes from his charges, 5% comes from blood, and the rest is from hallucinations, if bubblegum is meleeing you and you are using the KA, you are already dead. now if bubblegum is meleeing you and you are using the crusher, that is a rather common accurance, so it isn't a nerf to bubblegum, instead it is a buff to mining with a KC, which isn't very viable right now.

HeWhoMustNOtBe:
Why the colossus exactly? Isnt no one already doing him because the loot is practically useless, seems more like it just makes it so he is more deadly if you run into him on accident as it doesnt really change anything about the fight itself

hal9000PR:
> Why the colossus exactly? Isnt no one already doing him because the loot is practically useless, seems more like it just makes it so he is more deadly if you run into him on accident as it doesnt really change anything about the fight itself

make it easier to fight with the crusher without heiro trophy, and people fight it cos it is fun.

Kyep (approved):

code/modules/mob/living/simple_animal/hostile/megafauna/hierophant.dm
```diff
@@ -194,12 +196,24 @@ Difficulty: Hard
...
 		INVOKE_ASYNC(src, .proc/burst, get_turf(src))
 
+/mob/living/simple_animal/hostile/megafauna/hierophant/bullet_act()
+	if(immune)
+		visible_message("<span class='danger'>[src] deflects the attack; [p_they()] can't be hit while focused!</span>", "<span class='userdanger'>You deflect the attack!</span>")
+		return
+	return ..()
+
+// This is its own var modification in a callback to avoid using spawn()
+/mob/living/simple_animal/hostile/megafauna/hierophant/proc/toggleImmunity(state)
```
farie82, line 206:
```suggestion
/mob/living/simple_animal/hostile/megafauna/hierophant/proc/toggle_immunity(state)
```
Be sure to change it to all the instances as well

code/modules/mob/living/simple_animal/hostile/megafauna/hierophant.dm
```diff
@@ -68,16 +68,18 @@ Difficulty: Hard
 							   /datum/action/innate/megafauna_attack/cross_blasts,
 							   /datum/action/innate/megafauna_attack/blink_spam)
 
+	var/abilities_landed = 0 //amount of abilities that has been landed on the current target
```
farie82, line 71: Can you make this and the ones below auto doc proper comments? Not required but a nice to have

code/modules/mob/living/simple_animal/hostile/megafauna/hierophant.dm
```diff
@@ -430,11 +448,16 @@ Difficulty: Hard
...
 	. = ..()
 	if(. && target && !targets_the_same)
+		abilities_landed = 0 // Reset for new targets
 		visible_message("<span class='hierophant_warning'>\"[pick(target_phrases)]\"</span>")
 		if(spawned_beacon && loc == spawned_beacon.loc && did_reset)
 			arena_trap(src)
 
 /mob/living/simple_animal/hostile/megafauna/hierophant/adjustHealth(amount, updating_health = TRUE)
+	if(immune)
+		visible_message("<span class='danger'>[src] deflects the attack; [p_they()] can't be hit while focused!</span>", "<span class='userdanger'>You deflect the attack!</span>")
```
farie82, line 458: `;` in a sentence? Probably grammatically correct but I rarely see it in-game so it might stand out as odd.

code/modules/mob/living/simple_animal/hostile/megafauna/hierophant.dm
```diff
@@ -676,7 +702,8 @@ Difficulty: Hard
 		to_chat(L, "<span class='userdanger'>You're struck by a [name]!</span>")
 		var/limb_to_hit = L.get_organ(pick(BODY_ZONE_HEAD, BODY_ZONE_CHEST, BODY_ZONE_R_ARM, BODY_ZONE_L_ARM, BODY_ZONE_R_LEG, BODY_ZONE_L_LEG))
 		var/armor = L.run_armor_check(limb_to_hit, "melee", "Your armor absorbs [src]!", "Your armor blocks part of [src]!", 50, "Your armor was penetrated by [src]!")
-		L.apply_damage(damage, BURN, limb_to_hit, armor)
+		L.apply_damage(damage+C.abilities_landed, BURN, limb_to_hit, armor)
```
farie82, line 705:
```suggestion
		L.apply_damage(damage + C.abilities_landed, BURN, limb_to_hit, armor)
```

farie82 (changes requested):

variableundefined:
Two maintainers objection. We don't want to go down the rabbit hole of megafauna rebalance. We might be open to selective changes to the hierophant only (in line with TG's) to bring it back into rotation. Contact me for more details.

PR #15344, opened 2021-01-24
Tweaks Styling Guide for Bitwise Operators

Fox-McCloud:
Makes it so that `thing | thing` is the how you handle bitwise `OR`.

Largely for readability as `thing|thing|thing|thing|thing|thing|thing` gets difficult to read; it's generally more consistent with everything else, as well.


dearmochi (approved):
Agreed

AffectedArc07 (approved):
I was already doing this

PR #15345, opened 2021-01-24
The beginning: Moves CC to Z1 and station to Z2

AffectedArc07:
## What Does This PR Do
This thus marks the beginning of map rotation. This moves CC to Z1, which is part of map rotation, and should be done in its own PR. A lot of stuff is hardcoded for Z1 and Z2, so this allows us to isolate errors outside of the main map rotation code.

## Why It's Good For The Game
Its time

## Changelog
:cl: AffectedArc07
experiment: The station is now Z2
experiment: CentComm is now Z1
/:cl:


Kiyahitayika:
Rotate your maps. Rotate your maps. R-r-rotate your maps.
For Science!

_maps/cyberiad.dm
```diff
@@ -3,20 +3,19 @@ All z-levels should be identical in size. Their numbers should not matter.
 The order of z-levels should not matter as long as their attributes are properly defined at MAP_TRANSITION_CONFIG.
 Old code checked for the number of the z-level (for example whether there are any revheads on Z1),
```
farie82, line 4: Can most likely be removed since I can't find an instance where a numeric or define value is checked

AffectedArc07: Why remove it? It says what the code used to do, and why you should use the new system.

farie82: It would suggest that there is still old code which checks it

AffectedArc07: `checked` is past tense, not `checks`.

AffectedArc07:
I can do this in map standardisation which is actually happening soon, lol

PR #15348, opened 2021-01-25
Tunes down Bluespace Harvester Hellportals

TDSSS:
## What Does This PR Do
This dramatically lowers the rate of malfunctions/hellportal spawns for simply running it at high levels. After the levels rebalance, high power levels are much more possible/easily achieved compared to before. The chance of an emagged BSH spawning portals has not been reduced.

However, an emagged BSH that encoutners a malfunction will now turn down 1 level, eventually returning it to a safe state unless the saboteur can keep upping the level again. This also puts a cap on how many portals an unattended, emagged BSH can produce.

## Why It's Good For The Game
Enacting the plot of doom once or twice is fun, doing it 5 times every round is not. The frequent malfunctions also discouraged people from actually trying to get high scores.

## Changelog
:cl:
tweak: Bluespace Harvesters are much less likely to malfunction at high levels unless sabotaged.
tweak: Bluespace Harvesters that are sabotaged and spawn a Nether Portal will now turn down a level, leading to them eventually shutting off if unattended.
/:cl:

### DIFF
--- a/code/modules/station_goals/bluespace_tap.dm
+++ b/code/modules/station_goals/bluespace_tap.dm
@@ -326,10 +326,15 @@
 	else if(input_level > desired_level)
 		input_level--
 	if(prob(input_level - safe_levels + (emagged * 5)))	//at dangerous levels, start doing freaky shit. prob with values less than 0 treat it as 0
+		if(!emagged && prob(99))	//makes non-sabotaged events MUCH less likely
+			return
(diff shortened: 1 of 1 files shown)


Kyep (commented):
Skeptical of this. I rarely see the harvester built, and the only times I can remember it spawning a bunch of demons were when admins used adminbus to make it do so.

Even then, it required a LOT of adminbus.

TDSSS:
All it requires for demons to spawn is an ambitious engineering team running it at a high level. With the supermatter, I'd expect the power generation to get easier as well.

Fox-McCloud:
I'd wait for this to become a problem before we address this, personally.

The obscene power levels seen, initially, were only possible prior to the radiation refactor---the power amount has dropped, significantly, since then---and running it at higher levels will be even more dangerous once the tesla refactor gets merged in.

TDSSS:
Alright, since it seems there's no support for this at the moment, I'll close it for now.

If opinions change, it should be an easy one to re-open.

PR #15349, opened 2021-01-25
Implements N2O Decomposition

Fox-McCloud:
Ports N2O decomposition reaction, from TG.

Pretty simple; if N2O hits 1500K or higher, it'll begin decomposing into its constituent parts; nitrogen and oxygen.

This makes it so that it's not a "freebie" for putting out fires and also means it's not a 100% free ride "Throw some in the SM chamber since it just makes it better!".

Tested it out and it works just fine without any issues, that I could see. 

I'd love to implement more gas interactions, but this is about as good as it gets for what we can do, for now, sadly.

:cl: Fox McCloud
add: N2O will now decompose into nitrogen and oxygen if it's 1500 Kelvin or hotter
/:cl:


Necaladun (approved):
one step closer to my dream goal of atmos and chemistry being the same system

code/datums/gas_mixture.dm
```diff
@@ -80,7 +82,7 @@ What are the archived variables for?
 
 
 /datum/gas_mixture/proc/react(atom/dump_location)
-	var/reacting = 0 //set to 1 if a notable reaction occured (used by pipe_network)
+	var/reacting = FALSE //set to 1 if a notable reaction occured (used by pipe_network)
```
AffectedArc07, line 85:
```suggestion
	var/reacting = FALSE //set to TRUE if a notable reaction occured (used by pipe_network)
```
Yes I am think nitpicky

code/__DEFINES/atmospherics.dm
```diff
@@ -133,3 +133,7 @@
 
 //LAVALAND
 #define LAVALAND_EQUIPMENT_EFFECT_PRESSURE 50 //what pressure you have to be under to increase the effect of equipment meant for lavaland
+
+// Reactions
+#define N2O_DECOMPOSITION_MIN_ENERGY		1400
+#define N2O_DECOMPOSITION_ENERGY_RELEASED	200000
```
AffectedArc07, line 139: Get out fusion we have a 142x energy efficiency rate, literally destroying the universe

*this isnt a review comment just me joking around with values*

Kiyahitayika:
_one step closer to /tg/ atmos?_

Kyep (approved):

AffectedArc07 (approved):

PR #15354, opened 2021-01-25
Adds a heal point system to the rod of Asclupius

Qwertytoforty:
## What Does This PR Do

The status affect from the rod of asclepius now has "healing points" The buff has a max storage of 50 healing points, and generates 3 per buff tick. Every time the buff ticks, for each mob in range (including the own user) it uses a healing point to try to heal them, and an extra healing point if it tries to heal a limb. Thus, the rod is going to work perfectly fine when only the user and one other person is near them, but if 2 people are near the user, it will very slowly deplete over time, and if it is 8 people fighting a blob, it will run out of healing charge quick.

## Why It's Good For The Game

The rod of asclupius is can heal more damage types than a medbeam, is AOE instead of one target, doesn't explode, and is broken when used to combat a blob or just sitting around in medbay, as you rapidly heal crew and fix their internal injuries. 

Many times this has been used against a blob or similar threat, healing all crew fighting the threat faster than the threat can deal damage, and even if the person breaks a bone or gets IB, instead of having to go to medbay, just standing around the person with the rod will eventually fix it.

After this change, the rod still heals just as much damage, but can not heal a whole bunch of people at once, and instead is better healing 1 or 2  people other than the user at a time, or a group of 4-6 for a short period of time quickly.

## Changelog
:cl:
tweak: Rod of Asclupius now has a heal point system that regenerates over time, and is spent when trying to heal people. Now works better when healing small groups of people, and will stop working for a while if too many people are attempted to be healed at once.
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -109,8 +109,8 @@
 	alert_type = null
 	var/hand
 	var/deathTick = 0
-	/// How many points the rod has to heal with, starts with 25, maxes at 50, or whatever heal_points_max is set to.
-	var/heal_points = 25
+	/// How many points the rod has to heal with, maxes at 50, or whatever heal_points_max is set to.
+	var/heal_points = 50
 	/// Max heal points for the rod to spend on healing people
 	var/max_heal_points = 50
 
@@ -181,9 +181,9 @@
 			itemUser.adjustStaminaLoss(-1.5)
 			itemUser.adjustBrainLoss(-1.5)
 			itemUser.adjustCloneLoss(-0.5) //Becasue apparently clone damage is the bastion of all health
-		//Heal all those around you, unbiased
 		if(heal_points < (max_heal_points))
 			heal_points = min(heal_points += 3, max_heal_points)
+		//Heal all those around you, unbiased
 		for(var/mob/living/L in view(7, owner))
 			if(heal_points <= 0)
 				break
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -181,7 +181,7 @@
 			itemUser.adjustStaminaLoss(-1.5)
 			itemUser.adjustBrainLoss(-1.5)
 			itemUser.adjustCloneLoss(-0.5) //Becasue apparently clone damage is the bastion of all health
-		if(heal_points < (max_heal_points))
+		if(heal_points < max_heal_points)
 			heal_points = min(heal_points += 3, max_heal_points)
 		//Heal all those around you, unbiased
 		for(var/mob/living/L in view(7, owner))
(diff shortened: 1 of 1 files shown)


code/datums/status_effects/buffs.dm
```diff
@@ -178,7 +182,11 @@
 			itemUser.adjustBrainLoss(-1.5)
 			itemUser.adjustCloneLoss(-0.5) //Becasue apparently clone damage is the bastion of all health
 		//Heal all those around you, unbiased
+		if(heal_points < (max_heal_points))
+			heal_points = max((heal_points += 3), max_heal_points)
```
SteelSlayer, line 186: The `+=` makes it possible to store a bit more than `max_heal_points`. For example if `heal_points` is 48, the 48 turns to 51, and then it will be `max(51, 50)`, which is 51.

Qwertytoforty: Really not sure why I did code like that, should be min, as then if heal points was 12, it goes to 15 and chooses 12, it it was 29, it goes to 52 and chooses 50. Will fix

Kyep (approved):

code/datums/status_effects/buffs.dm
```diff
@@ -178,7 +182,11 @@
 			itemUser.adjustBrainLoss(-1.5)
 			itemUser.adjustCloneLoss(-0.5) //Becasue apparently clone damage is the bastion of all health
 		//Heal all those around you, unbiased
+		if(heal_points < (max_heal_points))
```
SabreML, line 185: Don't need this extra set of brackets here:
```suggestion
		if(heal_points < max_heal_points)
```

code/datums/status_effects/buffs.dm
```diff
@@ -178,7 +182,11 @@
...
+			heal_points = min((heal_points += 3), max_heal_points)
```
SabreML, line 186: Or here actually, since it's the only thing in that argument:
```suggestion
			heal_points = min(heal_points += 3, max_heal_points)
```

code/datums/status_effects/buffs.dm
SabreML, line 185: This comment should probably be moved back to line 186, above the `for()` loop.

variableundefined (changes requested):
@Qwertyotoforty please see requested change by sabre above

variableundefined:
DM when ready for merge

code/datums/status_effects/buffs.dm
```diff
@@ -109,6 +109,10 @@
 	alert_type = null
 	var/hand
 	var/deathTick = 0
+	/// How many points the rod has to heal with, starts with 25, maxes at 50, or whatever heal_points_max is set to.
+	var/heal_points = 25
+	/// Max heal points for the rod to spend on healing people
+	var/max_heal_points = 50
```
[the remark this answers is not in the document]
Qwertytoforty: changed to 50

code/datums/status_effects/buffs.dm
Qwertytoforty: other option used

Qwertytoforty: comment moved

SabreML (approved):

variableundefined (approved):

PR #15357, opened 2021-01-26
Refactors Electrocution and Tesla Zapping

Fox-McCloud:
Refactors electrocution and tesla zapping to be handled via flags instead of goodness knows how many arguments.

From TG, credit to them for most of this.

Previously `electrocute_act` had a lot of arguments, but these have now been condensed into a few flags to make things easier, instead of remembering a ton of positional arguments.

This also axes `tesla_zap` in favor of `zap_act`, which is more generically named; the tesla utilizes this, but it's under the generic proc `tesla_zap`.

The supermatter and tesla now use this new `zap_act` and `tesla_act`, which has balance implications.

The tesla doesn't just make things explode anymore, but deals object damage to things, too---so does the supermatter for that matter.

This has implications of making the supermatter SIGNIFICANTLY more dangerous if you overpower it---the zapping is no longer a mere "oh, it's fine, I have gloves on". Depending on the strength of the zaps, it could override your gloves protection---at high enough levels it'll start damaging things---and at very high levels, it'll cause things to just out and out explode.

This can cause a run-away chain reaction, if you're not careful---the crystal will overpower, which will cause it to arc, which will in turn cause things to explode, therefore busting open the chamber---at this point, the supermatter starts sucking things into itself, thus giving itself more power.

The end result is runaway charge reaction that's likely to release a Tesla and cause yet even more havoc.

The upside to this, however, is you can build grounding rods to ground out the excess energy---you can even build Tesla coils so that you can generate *even more power* from overloading the supermatter.

:cl: Fox McCloud
add: Tesla can now cause object damage instead of just exploding things
add: Supermatter now releases actual tesla bolts; if these aren't dealt with, they'll damage and explode things---upside, you can harvest even more power from a highly energetic supermatter!
/:cl:


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -144,54 +144,44 @@
...
-		shock_internal_organs(shock_damage)
-	visible_message(
-		"<span class='danger'>[src] was shocked by \the [source]!</span>",
-		"<span class='userdanger'>You feel a powerful shock coursing through your body!</span>",
-		"<span class='italics'>You hear a heavy electrical crack.</span>")
-	AdjustJitter(1000) //High numbers for violent convulsions
+	. = ..()
+	if(!.)
+		return
+	//Propagation through pulling, fireman carry
```
Fox-McCloud, line 152: we don't have this, remove the latter comment.

code/modules/power/supermatter/supermatter.dm
```diff
@@ -511,21 +522,40 @@
...
+				flags |= (ZAP_MOB_DAMAGE | ZAP_OBJ_DAMAGE)
+				zap_count = 3
+			if(CRITICAL_POWER_PENALTY_THRESHOLD to INFINITY)
+				zap_icon = OVER_9000_ZAP_ICON_STATE
+				//It'll stun more now, and damage will hit harder, gloves are no garentee.
+				//Machines go boom
+				flags |= (ZAP_MOB_STUN | ZAP_MACHINE_EXPLOSIVE | ZAP_MOB_DAMAGE | ZAP_OBJ_DAMAGE)
+				zap_count = 4
+		//Now we deal with damage shit
+		if (damage > damage_penalty_point && prob(20))
```
Fox-McCloud, line 552: spacing.

code/modules/power/tesla/coil.dm
```diff
@@ -77,38 +82,45 @@
...
 	last_zap = world.time
-	var/coeff = (20 - ((input_power_multiplier - 1) * 3))
-	coeff = max(coeff, 10)
-	var/power = (powernet.avail/2)
+	var/power = (powernet.avail) * 0.2 * input_power_multiplier  //Always always always use more then you output for the love of god
+	power = min(surplus(), power) //Take the smaller of the two
 	add_load(power)
-	playsound(src.loc, 'sound/magic/lightningshock.ogg', 100, 1, extrarange = 5)
-	tesla_zap(src, 10, power/(coeff/2))
+	playsound(src.loc, 'sound/magic/lightningshock.ogg', 100, TRUE, extrarange = 5)
```
Fox-McCloud, line 109: axe implied `src` and double check this file.

Kyep (approved):

code/game/objects/structures/grille.dm
```diff
@@ -271,7 +271,7 @@
 				var/obj/structure/cable/C = T.get_cable_node()
 				if(C)
 					playsound(src, 'sound/magic/lightningshock.ogg', 100, TRUE, extrarange = 5)
-					tesla_zap(src, 3, C.newavail() * 0.01) //Zap for 1/100 of the amount of power. At a million watts in the grid, it will be as powerful as a tesla revolver shot.
+					tesla_zap(src, 3, C.newavail() * 0.01, ZAP_MOB_DAMAGE | ZAP_OBJ_DAMAGE | ZAP_MOB_STUN | ZAP_ALLOW_DUPLICATES) //Zap for 1/100 of the amount of power. At a million watts in the grid, it will be as powerful as a tesla revolver shot.
```
AffectedArc07, line 274: Feel like the flags should be wrapped in parenthesis for better readability but thats just me

code/modules/power/tesla/coil.dm
```diff
@@ -77,38 +87,45 @@
...
+	desc = "Keeps an area from being fried by Edison's Bane."
 	icon = 'icons/obj/tesla_engine/tesla_coil.dmi'
 	icon_state = "grounding_rod0"
-	anchored = 0
-	density = 1
+	anchored = FALSE
+	density = TRUE
+
+	can_buckle = TRUE
+	buckle_lying = 0
```
AffectedArc07, line 127: Should this be TRUE/FALSE?

code/modules/power/tesla/energy_ball.dm
```diff
@@ -1,6 +1,15 @@
...
 #define TESLA_MINI_POWER 869130
-
+//Zap constants, speeds up targeting
+#define COIL (ROD + 1)
+#define ROD (RIDE + 1)
+#define RIDE (LIVING + 1)
+#define LIVING (MACHINERY + 1)
+#define MACHINERY (BLOB + 1)
+#define BLOB (STRUCTURE + 1)
+#define STRUCTURE (1)
```
AffectedArc07, line 10: Given these appear in the SM and the tesla, should these be globally scoped as `ZAPCONSTANT_COIL` or whatever? Just incase someone updates them in one plate not another

code/modules/power/tesla/energy_ball.dm
```diff
@@ -12,17 +21,28 @@
 	move_self = 1
 	grav_pull = 0
 	contained = 0
```
AffectedArc07, line 23: Should these be `TRUE`/`FALSE`?

AffectedArc07 (commented):
Left general comments, unsure if some things are intentional or not

code/modules/power/tesla/coil.dm
Fox-McCloud: This is not standardized well, unfortunately---this is actually not `TRUE/FALSE`; this is an angle that gets multiplied by this number.

AffectedArc07: Oh..

code/modules/power/tesla/energy_ball.dm
Fox-McCloud: I don't know what this even does; I'll remove the var---I"m guessing this is some really old var that never got purged.

code/modules/power/tesla/energy_ball.dm
Fox-McCloud: The Tesla and SM shocking are actually different, if you look closely at the defines.

code/game/objects/structures/grille.dm
Fox-McCloud: We usually only do that, when we have to, to prevent errors in logic--must like everything else.

AffectedArc07 (approved):

PR #15358, opened 2021-01-26
Fixes Scrubbers and Alarms

Fox-McCloud:
A few fixes---the larger one will have to wait on mapping standardization.

Air Alarm flood option doesn't work because it uses the incorrect setting--this corrects that---credit to TG for this very old fix.

This also ensures the proper settings for air alarm only filters out carbon dioxide, by default.

I dunno when or where this happened, but plasma and N2O are not supposed to be filtered, by default; only CO2. My best guess is that we inherited this from Bay because of ZAS and "immersion".

It might make sense for a realism standpoint to filter out everything, by default, but it makes shennanigans a lot less impactful and accidents self-repairing---since the scrubbers are just set to filter things as is---either case, this is not the intended behavior or functionality, I don't believe--even so, I"m marking it as a balance change because of what it'll impact.

This also fixes the "Cargon Dioxide" mispelling.

Once the mapping standardization is in, then I'll correct the scrubbers.

Hey, guess what? Every single ***GOSH DARNED SCRUBBER IN EXISTENCE IS A MAP EDIT***

:cl: Fox Mccloud
fix: Fixes air alarm flood setting (emagged setting for malf AI and emagged air alarms) so it works
tweak: air alarms are configured to only scrub CO2 by default
/:cl:


FreeStylaLT:
> Hey, guess what? **Every single GOSH DARNED SCRUBBER IN EXISTENCE IS A MAP EDIT**
image

AffectedArc07 (approved):

PR #15363, opened 2021-01-26
Fixes Cigarettes rendering on mob when on the ears

Fox-McCloud:
Things aren't really meant to be equipped to the ears as a standard, and their icon handling for them isn't great, resulting in things equipped to them rendering as if they were in another slot.

Unsurprisingly, things that are equipped to the ears are typically *only* equippable to the ears.

Having dual ear + other slot just doesn't render properly and results in a lot of oddities.

This fixes that for cigs and their subtypes.

:cl: Fox McCloud
fix: Fixes items on your ear rendering as if they were equipped to another slot
del: can't requip cigs to your ears anymore
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/cigs.dm
+++ b/code/game/objects/items/weapons/cigs.dm
@@ -18,7 +18,7 @@ LIGHTERS ARE IN LIGHTERS.DM
 	icon_state = "cigoff"
 	throw_speed = 0.5
 	item_state = "cigoff"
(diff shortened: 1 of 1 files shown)


Carthusia:
I'm torn on this, as this will destroy part of my asthetic of having a pipe slotted into a filter slot on my mask.

I get its a bit of an oddity and an unintended feature, though personally I feel it hurts no one.

Do we really need to get rid of it?

AffectedArc07 (approved):

PR #15364, opened 2021-01-26
Better Supermatter Monitor

Fox-McCloud:
Ports part of TG's Supermatter monitor interface.

Why? It basically allows you to see how close to danger you are, instead of it just being a number---also some of the numbers were inaccurate, and it'd get fussy if the energy output was over a very small amount.

This also allows you to see how much nitrous and agent b are in the chamber, with better graphs for that, as well!

img
img

:cl: Fox McCloud
add: Improves the supermatter monitor UI
/:cl:


code/game/machinery/computer/sm_monitor.dm
```diff
@@ -52,23 +52,21 @@
...
+			data["SM_gas_CO2"] = round(100 * air.carbon_dioxide / TM, 0.01)
+			data["SM_gas_N2"] = round(100 * air.nitrogen / TM, 0.01)
+			data["SM_gas_PL"] = round(100 * air.toxins / TM, 0.01)
+			data["SM_gas_N2O"] = round(100 * air.sleeping_agent / TM, 0.01)
+			data["SM_gas_AB"] = round(100 * air.agent_b / TM, 0.01)
 		else
 			data["SM_gas_O2"] = 0
 			data["SM_gas_CO2"] = 0
 			data["SM_gas_N2"] = 0
-			data["SM_gas_PH"] = 0
```
Fox-McCloud, line 70: someone typo'd!

tgui/packages/tgui/interfaces/SupermatterMonitor.js
```diff
@@ -102,33 +75,54 @@ const SupermatterMonitorDataView = (props, context) => {
...
             <LabeledList.Item label="Temperature">
-              <Box color={temperatureToColor(data.SM_ambienttemp)}>
-                {data.SM_ambienttemp} K
-              </Box>
+              <ProgressBar
+                value={logScale(data.SM_ambienttemp)}
+                minValue={0}
+                maxValue={logScale(10000)}
+                ranges={{
+                  teal: [-Infinity, logScale(80)],
```
Fox-McCloud, line 106: @AffectedArc07 we support `teal:` right?

code/game/machinery/computer/sm_monitor.dm
```diff
@@ -52,23 +52,23 @@
...
-			if(other_moles)
-				data["SM_gas_OTHER"] = round(100 * other_moles / TM, 0.01)
-			else
-				data["SM_gas_OTHER"] = 0
+			gasdata.Add(list(list("name"= "Oxygen", "amount" = round(100 * air.oxygen / TM, 0.01))))
+			gasdata.Add(list(list("name"= "Carbon Dioxide", "amount" = round(100 * air.carbon_dioxide / TM, 0.01))))
+			gasdata.Add(list(list("name"= "Nitrogen", "amount" = round(100 * air.nitrogen / TM, 0.01))))
+			gasdata.Add(list(list("name"= "Plasma", "amount" = round(100 * air.toxins / TM, 0.01))))
+			gasdata.Add(list(list("name"= "Nitrous Oxide", "amount" = round(100 * air.sleeping_agent / TM, 0.01))))
+			gasdata.Add(list(list("name"= "Agent B", "amount" = round(100 * air.agent_b / TM, 0.01))))
```
Fox-McCloud, line 63: this is gross because of how we handle it vs TG; they just use a for loop....we can't do that....

tgui/packages/tgui/interfaces/SupermatterMonitor.js
dearmochi: Yes, see `CSS_COLORS` in `tgui/constants.js`

code/game/machinery/computer/sm_monitor.dm
dearmochi: I think first you could define a (static) list at proc or class level, something along the lines of:
```dm
var/static/list/gasses = list(
	"oxygen" = "Oxygen",
	"carbon_dioxide" = "Carbon Dioxide",
	// ETC...
)
```

Then you could use a define along with direct substitution:
```dm
/// To be used only in ui_data()
#define GAS_ENTRY_DATA(n, k) list(list("name" = "[n]", "amount" = round(100 * air.[##k] / TM, 0.01)))
```

...Finally allowing you to make a clean loop in the data proc:
```dm
for(var/gastype in gasses)
	gasdata += GAS_ENTRY_DATA(gasses[gastype], gastype)
```

tgui/packages/tgui/interfaces/SupermatterMonitor.js
```diff
@@ -91,66 +69,104 @@ const SupermatterMonitorListView = (props, context) => {
...
+                </LabeledList.Item>
+                <LabeledList.Item label="Pressure">
+                  <ProgressBar
+                    value={logScale(SM_ambientpressure)}
+                    minValue={0}
+                    maxValue={logScale(50000)}
+                    ranges={{
+                      good: [logScale(1), logScale(300)],
+                      average: [-Infinity, logScale(1000)],
+                      bad: [logScale(1000), +Infinity],
```
dearmochi, line 135:
```suggestion
                      bad: [logScale(1000), Infinity],
```

dearmochi (commented):
One very little thing

code/game/machinery/computer/sm_monitor.dm
Fox-McCloud: Yeah, I contemplated this, but I'd like to generalize it, especially so it can be utilized with something like `getGasLabel` and `getGasColor` (both TGUI constants), as it'd make looping through every gas type a bit easier.

Unfortunately, in part, it's trying to fit a var based atmospheric system into working with a list based one, so you're going to wind up with some nasty stuff when converting between the two.

tgui/packages/tgui/interfaces/SupermatterMonitor.js
Fox-McCloud: Thanks!

dearmochi (approved):

Dumbdumn5 (approved):
Once that merge conflict tag gets sorted, :+1: here!

AffectedArc07 (approved):

PR #15366, opened 2021-01-27
Improves Geiger Counters; Allows Toggle

Fox-McCloud:
Refactors Geiger counters a bit--they should be a hair bit more responsive than they previously were---this was partially inspired by TG, but they use a whole other system called `delta_time` to handle processing differently.

This also corrects a couple of issues @SteelSlayer brought up in the original radiation PR review.

And lastly, this also allows you to toggle a hardsuit helmet's geiger counter on or off--after having witnessed this, the past few days, there's definitely circumstances where you'd just not want to put up with it blaring out when it's obvious you or the things around you are emitting radiation and you just don't need to know anymore. The counter, though, defaults to `on` for hardsuit helmets, unlike regular geiger counters.

:cl: Fox McCloud
tweak: Geiger counters should be a hair bit more responsive
add: Can toggle hardsuit helmet's geiger counter on and off now
/:cl:

### DIFF
--- a/code/game/objects/items/devices/geiger_counter.dm
+++ b/code/game/objects/items/devices/geiger_counter.dm
@@ -33,30 +33,28 @@
 
 /obj/item/geiger_counter/Destroy()
 	STOP_PROCESSING(SSobj, src)
+	QDEL_NULL(soundloop)
 	return ..()
 
 
 /obj/item/geiger_counter/process()
-	update_icon()
-	update_sound()
+	if(scanning)
+		radiation_count -= radiation_count / RAD_GEIGER_MEASURE_SMOOTHING
+		radiation_count += current_tick_amount / RAD_GEIGER_MEASURE_SMOOTHING
 
-	if(!scanning)
-		current_tick_amount = 0
-		return
+		if(current_tick_amount)
+			grace = RAD_GEIGER_GRACE_PERIOD
+			last_tick_amount = current_tick_amount
 
-	radiation_count -= radiation_count / RAD_GEIGER_MEASURE_SMOOTHING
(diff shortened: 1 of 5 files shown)


code/game/objects/items/devices/geiger_counter.dm
```diff
@@ -169,13 +167,13 @@
 		return ..()
 
 /obj/item/geiger_counter/AltClick(mob/living/user)
-	if(!istype(user) || !!user.Adjacent(src))
+	if(!istype(user) || !user.Adjacent(src))
 		return ..()
 	if(!scanning)
 		to_chat(usr, "<span class='warning'>[src] must be on to reset its radiation level!</span>")
```
farie82, line 173:
```suggestion
		to_chat(user, "<span class='warning'>[src] must be on to reset its radiation level!</span>")
```

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -105,19 +107,37 @@
...
+	if(ishuman(loc))
+		update_sound()
+
+/obj/item/clothing/head/helmet/space/hardsuit/proc/update_sound()
+	var/datum/looping_sound/geiger/loop = soundloop
+	if(!scanning)
+		loop.stop(loc)
+		return
+	if(!radiation_count)
+		loop.stop(loc)
```
farie82, line 132:
```suggestion
	if(!scanning || !radiation_count)
		loop.stop(loc)
```

farie82 (changes requested):

farie82 (approved):

AffectedArc07 (approved):

PR #15371, opened 2021-01-28
Tank Refactor; Breathing Retool

Fox-McCloud:
The way must be prepared. . .

Refactors tanks---this has been something on the agenda for a long time, but it's finally done.

Yes, this means you can no longer equip jetpacks to the suit storage slot---no more confused newbies!

Generally cleans up tank code and standardizes it. I did remove all the custom examines for tanks, because, quite frankly, it's not sustainable---and makes no sense for it to work or not based on the type of tank you have---if you can't be bothered to open a tank's interface to look at the bar of how much pressure you have...then well, I dunno what to tell you.

This also adjust lungs so that the damage for breathing in gases is significantly less.

This means no more of this, for getting dropped in the toxins tank, from breathing ONCE:

img

This also adjusts `handle_internal_lifeform` to be in line with TG---something that we should have done eons ago.

Lastly, this adjusts *how* much you breathe at once--you should drain tanks a bit faster and turfs should accumulate CO2 a bit more. A few tanks have had their volumes slightly adjusted.

Oh, I almost forget, this also implements the improvised jetpack--you make it with an extinguisher, some oxygen tanks, pipes, and cable coil---it doesn't hold much and is prone to just shutting off out of the blue, but it can let you have a functional jetpack.

This originally was going to be a simple refactor of how much damage breathing nasty gases was going to do, but it reminded me that `BREATH_MOLES` was a thing and needed to be gotten rid of---which in turn got me to notice the increased breath volume---which led me to tanks and their utter mess...which led to this refactor.

:cl: Fox McCloud
tweak: breathing highly dense nasty gases shouldn't one-shot you anymore
tweak: should breathe a bit more gas in each breath; tanks will drain a bit faster and turfs will accumulate CO2 a bit faster
tweak: made some slight volume adjustments on a few tanks
del: can no longer equip jetpacks to the suit storage slot
add: can craft a ghetto improvised jetpack---it's finicky, expect it to turn off on occasion
/:cl:


hal9000PR:
Will this affect nukies jetpack harnesses? Personally i only use those as internals because nukie suits have built in jetpacks so if i cannot put that in my exosuit slot i really don't see a use for the jetpack and it should just be replaced by a standard big oxygen tank.

AffectedArc07:
> Will this affect nukies jetpack harnesses? Personally i only use those as internals because nukie suits have built in jetpacks so if i cannot put that in my exosuit slot i really don't see a use for the jetpack and it should just be replaced by a standard big oxygen tank.

Nukies can use standard oxygen tanks, they are given some I believe

AffectedArc07:
Closing and re-opening because CI doesnt exist

code/__DEFINES/atmospherics.dm
```diff
@@ -21,8 +21,7 @@
 #define MOLES_O2STANDARD		(MOLES_CELLSTANDARD*O2STANDARD)	// O2 standard value (21%)
 #define MOLES_N2STANDARD		(MOLES_CELLSTANDARD*N2STANDARD)	// N2 standard value (79%)
 #define CELL_VOLUME				2500	//liters in a cell
-#define BREATH_VOLUME			0.5		//liters in a normal breath
-#define BREATH_MOLES 			(ONE_ATMOSPHERE * BREATH_VOLUME /(T20C*R_IDEAL_GAS_EQUATION))
+#define BREATH_VOLUME			1		//liters in a normal breath
```
AffectedArc07, line 24:
```suggestion
/// Litres in a normal breath
#define BREATH_VOLUME			1
```
Not necessary but would be nice

code/modules/admin/verbs/debug.dm
```diff
@@ -744,7 +744,7 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
 	for(var/obj/machinery/power/rad_collector/Rad in GLOB.machines)
```
AffectedArc07, line 744: This entire verb needs axing at some point, but thats out the scope of this PR

AffectedArc07 (commented):
I didnt spot anything major, but I dont know atmos code to an amazing degree. Just a few nitpicks

datlo:
You say improved jetpack in the description and the CL. I assume it's supposed to be 'improvised'.

Great changes from what I see.

Fox-McCloud:
>You say improved jetpack in the description and the CL. I assume it's supposed to be 'improvised'.

correct---my mistake.

Butlerblock:
Is the co2 accumulation raised enough to affect station areas if vents and scrubbers everywhere are welded for a prolonged period of time?

Fox-McCloud:
>Is the co2 accumulation raised enough to affect station areas if vents and scrubbers everywhere are welded for a prolonged period of time?

Probably not---it being high enough would lead to atmos being over-active.

Dumbdumn5 (approved):
I'd asked about the vox and plasmaman tank drain rates and supposedly we could hotfix that in with a simple one var change if anything goes wrong. Otherwise this does look good! :+1:

AffectedArc07 (approved):

PR #15374, opened 2021-01-28
Makes blob stuff /Initialize()

AffectedArc07:
## What Does This PR Do
Makes blob stuff Initialize(). This is in its own PR to make it easier to review, and a TM of this would be nice just to absolutely make sure I dont brick an antag. I tested nodes and stuff in singleplayer but cant test split consciousness in a singleplayer environment.

## Why It's Good For The Game
/New()'s days are numbered

## Changelog
:cl: AffectedArc07
add: Blob stuff now initializes with a better system
/:cl:


code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -12,19 +12,19 @@
 	var/is_offspring = null
 	var/selecting = 0
 
-/obj/structure/blob/core/New(loc, var/h = 200, var/client/new_overmind = null, var/new_rate = 2, offspring)
```
AffectedArc07, line 15: This `h` var was never used anywhere, so I fail to see why it was passed all the way down

code/game/gamemodes/blob/powers.dm
```diff
@@ -389,7 +389,7 @@
 		return
 
 	split_used = TRUE
-	new /obj/structure/blob/core/ (get_turf(N), 200, null, blob_core.point_rate, offspring = TRUE)
```
Fox-McCloud, line 392: `/core/ (`

what the heck.

AffectedArc07:
TM'd with no issues

PR #15375, opened 2021-01-28
Uses crisp font in blob HUD maptext

dearmochi:
## What Does This PR Do

Makes font used in blob's HUD crisp

## Why It's Good For The Game

Increased readability

## Images of changes

| Before | After |
| --- | --- |
| image |  image

## Changelog
:cl:
add: Blob HUD (core health, resources) text should now look crisper
/:cl:


AffectedArc07:
Declaring this a fix because its barely readable

AffectedArc07 (approved):

PR #15377, opened 2021-01-28
Adds a backstab effect to the edagger

hal9000PR:
## What Does This PR Do
Edaggers will now weaken for 1 cycle and do a bonus 10 damage if you stab your target from behind.

## Why It's Good For The Game
Edaggers are slightly underwhelming. They can be easily outclassed by a welder, kitchen knife or syndie toolbox in actual viability when you take into consideration all of those are easily available, cost no or less TC and they won't get you perma'd for owning them. Adding a situational combat buff to the Edagger makes it better at actually assassinating people and worth the 2tc you spend on it. It might also help people make cool antag builds revolving around being a knifed assassin. 

Overall adding more viable options that traitors can purchase will add variety to the game and make rounds less static.  

## Changelog
:cl:
add: Edagger now deals bonus damage and stuns for 0-2 seconds when stabbing your target from behind.
fix: fixes a bug where edaggers would give attack verbs when turned off
/:cl:

### DIFF
--- a/code/modules/paperwork/pen.dm
+++ b/code/modules/paperwork/pen.dm
@@ -142,11 +142,11 @@
 	light_color = LIGHT_COLOR_RED
 
 /obj/item/pen/edagger/attack(mob/living/M, mob/living/user, def_zone)
-	if(on == 1 && user.dir == M.dir && !M.weakened && user != M)
+	if(on && user.dir == M.dir && !M.incapacitated(TRUE) && user != M)
 		M.apply_damage(10, BRUTE, BODY_ZONE_CHEST)
 		M.Weaken(1)
-		M.visible_message("<span class='warning'>[user] stabs [M] in the back!", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!")
-	. = ..()
+		M.visible_message("<span class='warning'>[user] stabs [M] in the back!</span>", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!</span>")
+	return ..()
 
 /obj/item/pen/edagger/attack_self(mob/living/user)
 	if(on)
--- a/code/modules/paperwork/pen.dm
+++ b/code/modules/paperwork/pen.dm
@@ -137,10 +137,16 @@
  */
 /obj/item/pen/edagger
 	origin_tech = "combat=3;syndicate=1"
-	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut") //these wont show up if the pen is off
 	var/on = 0
 	var/brightness_on = 2
 	light_color = LIGHT_COLOR_RED
+
+/obj/item/pen/edagger/attack(mob/living/M, mob/living/user, def_zone)
+	if(on == 1 && user.dir == M.dir && !M.weakened && user != M)
+		M.apply_damage(10, BRUTE, BODY_ZONE_CHEST)
+		M.Weaken(1)
+		M.visible_message("<span class='warning'>[user] stabs [M] in the back!", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!")
+	. = ..()
 
 /obj/item/pen/edagger/attack_self(mob/living/user)
 	if(on)
@@ -148,6 +154,7 @@
 		force = initial(force)
 		sharp = 0
 		w_class = initial(w_class)
+		attack_verb = list() // no attack verbs when it is off
 		name = initial(name)
 		hitsound = initial(hitsound)
 		embed_chance = initial(embed_chance)
@@ -160,6 +167,7 @@
 		force = 18
 		sharp = 1
 		w_class = WEIGHT_CLASS_NORMAL
+		attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut")
 		name = "energy dagger"
 		hitsound = 'sound/weapons/blade1.ogg'
 		embed_chance = 100 //rule of cool


Corocan:
heh, nothing personnel, kid

Wojz:
coldsteel the vulpkanin

Fox-McCloud:
Depending on where in the `Life` cycle this happens, this won't do much of anything to being a 2 second stun.

If stuns ever are more consistent in duration, this will wind up being really frustrating to play against; if you get a backstab off, you can just perma-stun them.

e-daggers have their niche; they're an extremely stealthy and very easy-to-hide weapon that does 18 brute damage and is sharp.

Aside from getting into very large weapons, you can't really beat this damage point.

hal9000PR:
> Depending on where in the `Life` cycle this happens, this won't do much of anything to being a 2 second stun.
> 
> If stuns ever are more consistent in duration, this will wind up being really frustrating to play against; if you get a backstab off, you can just perma-stun them.
> 
> e-daggers have their niche; they're an extremely stealthy and very easy-to-hide weapon that does 18 brute damage and is sharp.
> 
> Aside from getting into very large weapons, you can't really beat this damage point.

Well, the issue is, edaggers aren't really that stealthy because they are really loud. about stunlocking, that can be said about harmbatons, AEGs, HoSes gun, and you cannot even stunlock people completely with backstabs. you can only get the backstab bonus if they aren't weakened, meaning they will always stand up at some point even if it is to be knocked back down immediately.

Wojz:
>If stuns ever are more consistent in duration (this is likely to happen eventually, once they're made into status effects), this will wind up being really frustrating to play against; if you get a backstab off, you can just perma-stun them.

I agree that the stun might be problematic for a 2 TC item, personally I think there should be a cooldown on how often it can actually stun sort of how telebaton has a delay. Generally I'm in favor of this being a bigger damage boost and maybe even no stun, and backstab having a cooldown to prevent abuse.

Kyep (approved):

code/modules/paperwork/pen.dm
```diff
@@ -137,17 +137,24 @@
...
-	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut") //these wont show up if the pen is off
 	var/on = 0
 	var/brightness_on = 2
 	light_color = LIGHT_COLOR_RED
 
+/obj/item/pen/edagger/attack(mob/living/M, mob/living/user, def_zone)
+	if(on == 1 && user.dir == M.dir && !M.weakened && user != M)
+		M.apply_damage(10, BRUTE, BODY_ZONE_CHEST)
+		M.Weaken(1)
+		M.visible_message("<span class='warning'>[user] stabs [M] in the back!", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!")
```
farie82, line 148:
```suggestion
		M.visible_message("<span class='warning'>[user] stabs [M] in the back!</span>", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!</span>")
```

code/modules/paperwork/pen.dm
```diff
@@ -137,17 +137,24 @@
  */
 /obj/item/pen/edagger
 	origin_tech = "combat=3;syndicate=1"
-	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut") //these wont show up if the pen is off
 	var/on = 0
 	var/brightness_on = 2
 	light_color = LIGHT_COLOR_RED
 
+/obj/item/pen/edagger/attack(mob/living/M, mob/living/user, def_zone)
+	if(on == 1 && user.dir == M.dir && !M.weakened && user != M)
```
farie82, line 145:
```suggestion
	if(on && user.dir == M.dir && !M.incapacitated(TRUE) && user != M)
```
using `incapacitated` will mean that you also don't backstab when they are resting or sleeping etc. the `TRUE` is there to avoid restraints from blocking it.
`on` is also a boolean value

code/modules/paperwork/pen.dm
```diff
@@ -137,17 +137,24 @@
...
+		M.apply_damage(10, BRUTE, BODY_ZONE_CHEST)
+		M.Weaken(1)
+		M.visible_message("<span class='warning'>[user] stabs [M] in the back!", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!")
+	. = ..()
```
farie82, line 149:
```suggestion
	return ..()
```
Standard is to return the parent call instead of set the return value to the parent call.

farie82 (changes requested):

farie82 (approved):

TDSSS:
If stuns are an issue, you could have backstabs simply do more damage.

Denthamos:
I too would agree for the stealthy part of the e-dagger if not for the loud noise it makes that can be heard from a good distance away making the entire stealth aspect of it moot when employed.

hal9000PR:
I really cannot see the stun being that OP. You cannot stun lock people with it, and people being stunned actively lowers your dps. Also you can get a 0tc item which has a much longer stun that's widely available, a stun prod.

TDSSS:
Maybe just lower or remove the noise it makes? Or give it a very low range so that you and your victim still hear the iconic dsword noise but people the next room over don't?

hal9000PR:
> Maybe just lower or remove the noise it makes? Or give it a very low range so that you and your victim still hear the iconic dsword noise but people the next room over don't?

In addition to the backstab thing?

Spacemanspark:
Literally nobody uses edaggers, I'd like to see that changed a bit. Right now, welding tools far outstrip them. 
Unsure about the stun, however. Perhaps apply a brief slowdown instead?

code/modules/paperwork/pen.dm
```diff
@@ -137,17 +137,24 @@
...
-	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut") //these wont show up if the pen is off
 	var/on = 0
 	var/brightness_on = 2
 	light_color = LIGHT_COLOR_RED
 
+/obj/item/pen/edagger/attack(mob/living/M, mob/living/user, def_zone)
+	if(on && user.dir == M.dir && !M.incapacitated(TRUE) && user != M)
+		M.apply_damage(10, BRUTE, BODY_ZONE_CHEST)
+		M.Weaken(1)
+		M.visible_message("<span class='warning'>[user] stabs [M] in the back!</span>", "<span class='userdanger'>[user] stabs you in the back! The energy blade makes you collapse in pain!</span>")
```
S34NW, line 148: these messages can be split for readability

PR #15378, opened 2021-01-28
Adds Sound Environments: Reverb and Echo

Fox-McCloud:
From TG.

Implements sound environments which adds reverb and echo to sounds, based on the area you're in.

Naturally, more open areas that have reflective surfaces echo more---so, for example, maintenance will echo a lot, where as a large room with lots of carpet typically won't.

I stress that this PR doesn't make me happy---utilizing sound environments like this WILL cause weird echoing overlapping issues that aren't really possible to fully overcome, because, at the end of the day, how this is being utilized is....a bit questionable, at best---as a number of older coders and I both came to the conclusion of. I stress, I don't dislike the *idea* of echo or reverb, it's just that....this system has drawbacks to it (though I suspect most of you won't notice).

Still, it is what it is---every codebase out there makes use of a system that's more or less similar to this, and most people aren't even aware there's an issue to begin with/don't hear/notice it, so---might as well. 

:cl: Fox McCloud
add: Adds echo and reverb that depends on the area you're in
/:cl:


Mitchs98:
Out of curiosity how is this going to effect instruments? Will those be affected in any way, or is it just stuff like gunshots and etcetra?

AffectedArc07 (approved):

Dumbdumn5 (approved):
_Spooky_ I like it! 👍

AffectedArc07 (approved):

PR #15385, opened 2021-01-29
Transitions to Full-tile Windows

Fox-McCloud:
Transitions to full-tile windows, station wide.

This was a long unfulfilled request of @Necaladun

This also updates windows so they're a bit more rounded and look smoother.

This also means that how damaged a window is will become far more prominent and noticeable

This also adjusts placing glass on a grille---it'll need 2 panes now instead of 1, but it'll be a full-tile window.

This does removed round-start colored windows, because, well, they don't look that great with full-tile windows.

Overall, I feel these windows are a bit more aesthetically pleasing, it showcases how damaged they are a bit better, and it means that the window shuffle cheese tactic for getting in/out of a station is a bit more involved now. It also means no more "in window" cheese for antags like revenants.

img
img

This doesn't actually require map edits (unless people eventually find mistakes, which they probably will), as most of our windows are handled via spawners anyway---so huzzah!

This also makes it so that breaking a window leaves some cleanable debris--bring along a way to clean it up to cover up the evidence!

Code wise I also refactored debris spawning for windows and window doors---it created the debris, at round start, then that was stored inside them until the window/window door broke---this has been axed---it saves on memory and initialization time.

:cl: Fox McCloud
add: Most windows are full-tile windows now
tweak: round-start windows are not pre-colored
tweak: Placing glass on a grill places a full tile window (2 panes now instead of 1)
/:cl:


code/game/objects/effects/spawners/windowspawner.dm
```diff
@@ -20,17 +20,17 @@
...
+		new windowtospawn(get_turf(src))
 
 	if(useGrille)
 		new /obj/structure/grille(get_turf(src))
 
-	src.air_update_turf(1) //atmos can pass otherwise
+	air_update_turf(TRUE) //atmos can pass otherwise
 	// Give some time for nearby window spawners to initialize
 	spawn(10)
 		qdel(src)
```
dearmochi, line 36: Right below here is the following code:
https://github.com/ParadiseSS13/Paradise/blob/fed65b1b02f59547c14e9bf225b8e772bc133da7/code/game/objects/effects/spawners/windowspawner.dm#L37-L38

Uncommenting that code (which was added [3 years ago](https://github.com/ParadiseSS13/Paradise/commit/b2d0bb1c0bd8b6216612fb536b3f499b5cfe5153)) and removing the `spawn(10)` block seems to function as expected still (on my local server).

code/game/objects/effects/spawners/windowspawner.dm
```diff
@@ -2,12 +2,12 @@
 	name = "window spawner"
 	icon = 'icons/obj/structures.dmi'
 	icon_state = "window_spawner"
-	var/useFull = 0
-	var/useGrille = 1
-	var/windowtospawn = /obj/structure/window/basic
-	anchored = 1 // No sliding out while you prime
+	var/useFull = TRUE
```
dearmochi, line 5: Since `windowtospawn` is a full window either way, setting this to `FALSE` would potentially lead to multiple full windows being created on the same turf. I can see two such cases happening by searching for that variable on the repository.

Fox-McCloud:
I'll have to resolve a bug with fulltiles, that I can't seem to locate.

Roundstart placed electrified fulltiles are fine if you bump into them--nothing happens.

But if you bump into one that you make yourself, you get shocked.

Fox-McCloud:
Managed to fix the `Bump` issue.

TG had the same problem, only what they noticed more was throwing things at a window, only for it to hit the grille underneath.

Either case, both things fixed now.

code/game/objects/effects/spawners/windowspawner.dm
Fox-McCloud: Yeahp---should be handled now.

code/game/objects/items/weapons/RCD.dm
```diff
@@ -492,17 +492,6 @@
...
-		for(var/cdir in GLOB.cardinal)
-			var/turf/T2 = get_step(T1, cdir)
-			if(locate(/obj/structure/window/full/shuttle) in T2)
-				continue // Shuttle windows? Nah. We don't need extra windows there.
-			if(!(locate(/obj/structure/grille) in T2))
-				continue
-			for(var/obj/structure/window/W in T2)
-				if(W.dir == turn(cdir, 180))
-					qdel(W)
-			var/obj/structure/window/reinforced/W = new(T2)
```
Fox-McCloud, line 504: I don't even get what this was trying to do.

code/game/objects/effects/spawners/windowspawner.dm
```diff
@@ -20,35 +21,34 @@
...
 		new /obj/structure/grille(get_turf(src))
 
-	src.air_update_turf(1) //atmos can pass otherwise
+	air_update_turf(TRUE) //atmos can pass otherwise
 	// Give some time for nearby window spawners to initialize
-	spawn(10)
-		qdel(src)
-	// why is this line a no-op
-	// QDEL_IN(src, 10)
+	QDEL_IN(src, 10)
```
AffectedArc07, line 36: This can probably be `QDEL_IN(src, 0)` since `SStimer` doesnt actually fire until the MC has fully started, and these are setup during the `SSatoms` portion

Carthusia:
Does this remove directional windows entirely, or just replace all windows that spawn on the station with full tiles?

SteelSlayer:
This makes things a lot simpler and fixes some issues so :+1:. But goodness, the green color on those glass sprites looks pretty bad.

matttheficus:
I agree with @SteelSlayer, the fulltile sprites just look REALLY bad, color wise. Maybe if it was less of a sea-glass green, it would look better, but im not sure what color would really fit for it.

I also agree with @Carthusia as well. Directional panes are SUPER important to a lot of neat atmos projects as well as a lot of Bar renovations, Bridge setups, and other general construction things. Id hope that you'd still be able to do the multi-directional glass panels around a grille still.

Mitchs98:
I genuinely dislike the full tile window on a whole as opposed to directional windows. I'm genuinely curious as to why the change and what the benefits of it are other than being able to discern the damage better. I especially dislike this if it fullstop removes directional windows from being a choice to build for players for aesthetic reasons, as mentioned above.

Overall....I strongly prefer the look of single pane windows, full tile windows just look like an awkward blob to me.

Varghh:
As long we can still build the directional windows I'm fine with it.

Spacemanspark:
I prefer the single tile windows, the full tiles look hella ugly.

Fox-McCloud:
>Does this remove directional windows entirely meaning players cannot build them

Absolutely not; this is about making full-tile windows the default in most things, but it doesn't remove the old windows; those are too useful for too many things to remove.

>But goodness, the green color on those glass sprites looks pretty bad.

Yeah, I don't like it either; I'll axe the colored windows.

Fox-McCloud:
@SteelSlayer 

Went ahead and axed the colored windows; updated the pics.

SteelSlayer:
Looks much better.

Carthusia:
Oh those look much nicer in blue, and thank you for clarifying fox

SabreML:
> @SteelSlayer
> 
> Went ahead and axed the colored windows; updated the pics.

Are all the sec windows also blue, now?

Fox-McCloud:
>Does this mean its no longer possible to place directional windows on a grille? Or does it give us the option to pick

It defaults to full tile since that's what's being aimed for. It'd rather not implement the option to pick since that'll make doing repairs in the middle of chaos and nonsense (or when you need to anchor it quickly after placing it) significantly more difficult with a pop up.

If you want to do the traditional method, you can always place it manually.

Necaladun:
@Fox-McCloud 
Only took 5 years. You busy in 2016 for another request? :p

Sinestia:
Not saying it's wrong, but this hits a very iconic aesthetic that this server retains. 

Ah well

Spacemanspark:
Not a fan of this removing the colors of security either.

MRgaty:
Its a sad day for brig but hey at least they do look good elsewhere.

Mitchs98:
> Even sec windows, yes.

The idea of security having blue windows sounds quite awful aesthetically...



> If you want to do the traditional method, you can always place it manually.

Pretty sure the single pane windows won't fit to a grille correctly manually.

code/game/objects/effects/spawners/windowspawner.dm
Fox-McCloud: I just did 'return INITIALIZE_HINT_QDEL` instead. I don't think this needs any spawn like mechanism anymore.

Spacemanspark:
To be fair, you _can_ click on a grille with glass in hand to place a directional window there.

...That is, of course, unless this PR turns those into full tile windows as well.

DevL-Whats-a-jpg:
> ...That is, of course, unless this PR turns those into full tile windows as well.

>It defaults to full tile since that's what's being aimed for. It'd rather not implement the option to pick since that'll make doing repairs in the middle of chaos and nonsense (or when you need to anchor it quickly after placing it) significantly more difficult with a pop up.

@Spacemanspark

dearmochi (approved):

Spacemanspark:
I promise I have the ability to read

AffectedArc07 (approved):
It kills part of an aesthetic but ill take it for the benefits these bring. Its a nice end to the easy breakin of thin windows on rooms where you toolspam and alt+click them.

PR #15386, opened 2021-01-29
supermatter_pull() now unbuckles mobs

dearmochi:
Pinging @Fox-McCloud as they worked on the SM.

## What Does This PR Do

Supermatter crystal pull now unbuckles nearby mobs.

## Why It's Good For The Game

Fixes an exploit allowing people to get front-row seats with the SM without getting pulled. This also means monkeys can no longer be used as immovable radiation accumulators.

## Changelog
:cl:
add: Supermatter crystal now unbuckles nearby mobs when pulling
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -773,6 +773,10 @@
 			var/mob/M = P
 			if(M.mob_negates_gravity())
 				continue //You can't pull someone nailed to the deck
+			else if(M.buckled)
+				var/atom/movable/buckler = M.buckled
+				if(buckler.unbuckle_mob(M, TRUE))
+					visible_message("<span class='danger'>[src]'s sheer force rips [M] away from [buckler]!</span>")
 		step_towards(P,center)
 
 /obj/machinery/power/supermatter_crystal/proc/supermatter_anomaly_gen(turf/anomalycenter, type = FLUX_ANOMALY, anomalyrange = 5)


dearmochi:
> This also means monkeys can no longer be used as immovable radiation accumulators.

It just occurred to me that this might still be possible by giving a (humanized) monkey enabled magboots:

https://github.com/ParadiseSS13/Paradise/blob/9340921cbbfeba185b8a89577551876650ece706/code/modules/power/supermatter/supermatter.dm#L779-L780

How would you feel about letting the SM pull the mob if they're weakened, resting or lying even with magboots on?

SabreML:
> How would you feel about letting the SM pull the mob if they're weakened, resting or lying even with magboots on?

If the SM is delaminating then you can get weakened by the anomalies, so this would definitely decrease the chances of it being fixed manually.

Wojz:
>How would you feel about letting the SM pull the mob if they're weakened, resting or lying even with magboots on?

How about have magboots turn off when you take them off? Then you can't put them on monke with them staying on.

Butlerblock:
Not sure how slime speed potions work completely, but wouldn’t having them turn off when taken off make magboots impossible to slime speed? Since (i think) you can only pot them if they’re unequipped, and if you turn them off then on it breaks the slowdown override.

Kyep (approved):

Kepplerx:
This is an "exploit" in power generation, but the fact that it makes way more power, shows a weakness in how rads are currently assigned. In addition, there is one more way to cheese the same way, by making a couple windows and grills, as such I recommend a refactoring to the rad code in in addition to this

Herder05:
As one of the people who did this I agree it needs to be removed but radiation needs to change to where doing that ISNT NEEDED cause as of right now the engine makes shit for power on a standard n2 setup

Dafrek:
I don't really understand why we would want this implemented into the game as it seems like it would be better to fix the actual radiation code that allows for exploits like this than to have mobs be unbuckled by the SM. I would imagine it would take more effort to fix all the ways this can be exploited individually than to just fix it by changing the radiation code. Also, it doesn't really make sense gameplay-wise that the SM can simply pull you out of a buckled chair, but maybe that's just me.

Kepplerx:
> > How would you feel about letting the SM pull the mob if they're weakened, resting or lying even with magboots on?
> 
> How about have magboots turn off when you take them off? Then you can't put them on monke with them staying on.

I see no need to change that, for a small issue such as this. The change you propose here would make it so one can never put on another person activated magboots. The "issue" as I see it isnt the monkey thing at all, but the SM and rads, and I can say that even if this change, or the proposed change was even implemented, I could work around that quite easily. 

On another point, this also is hardly an issue, as I dont see anyone really doing it either, so while an "exploit", it hardly affects the game significantly, and even less so as it isnt used.

PR #15396, opened 2021-01-30
Suit Storage Units now clean radiation (and blood)

TDSSS:
## What Does This PR Do
The cleaning cycle of suit storage units, so far only useful as a murder tool, now serve their actual function and clean blood as well as radiation from their content.

## Why It's Good For The Game
It makes sense. It also gives cleaning cycles an actual use beyond impractical murder.

## Changelog
:cl:
tweak: Suit Storage unit's cleaning cycle now removes radiation.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -793,7 +793,7 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 /atom/proc/clean_blood(radiation_clean = FALSE)
 	germ_level = 0
 	if(radiation_clean)
-		clean_radiation(clean_factor = 2)
+		clean_radiation()
 	if(islist(blood_DNA))
 		blood_DNA = null
 		return TRUE
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -793,14 +793,25 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 /atom/proc/clean_blood(radiation_clean = FALSE)
 	germ_level = 0
 	if(radiation_clean)
-		var/datum/component/radioactive/healthy_green_glow = GetComponent(/datum/component/radioactive)
-		if(!QDELETED(healthy_green_glow))
-			healthy_green_glow.strength = max(0, (healthy_green_glow.strength - (RAD_BACKGROUND_RADIATION * 2)))
-			if(healthy_green_glow.strength <= RAD_BACKGROUND_RADIATION)
-				qdel(healthy_green_glow)
+		clean_radiation(clean_factor = 2)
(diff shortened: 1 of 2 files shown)


Fox-McCloud:
I'd honestly just make this clean all rads off, instantly, at the end of the cycle.

TDSSS:
I feel like removing all rads, potentially thousands of them, could be a bit much. But if it has maintainer support, I'll think about it.

Kyep (approved):

TDSSS:
Code has been changed, now removes all radiation at the end of the cleaning cycle. I also moved radiation removal from clean_blood into its own proc that gets called by clean_blood, but can also be invoked on its own for the future.

code/game/atoms.dm
```diff
@@ -793,15 +793,26 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 /atom/proc/clean_blood(radiation_clean = FALSE)
 	germ_level = 0
 	if(radiation_clean)
-		var/datum/component/radioactive/healthy_green_glow = GetComponent(/datum/component/radioactive)
-		if(!QDELETED(healthy_green_glow))
-			healthy_green_glow.strength = max(0, (healthy_green_glow.strength - (RAD_BACKGROUND_RADIATION * 2)))
-			if(healthy_green_glow.strength <= RAD_BACKGROUND_RADIATION)
-				qdel(healthy_green_glow)
+		clean_radiation(clean_factor = 2)
```
AffectedArc07, line 796:
```suggestion
		clean_radiation()
```
You dont need to override the default, or use the full named arg here

code/game/machinery/suit_storage_unit.dm
```diff
@@ -414,6 +414,9 @@
 		uv_cycles = initial(uv_cycles)
 		uv = FALSE
 		locked = FALSE
+		for(var/atom/A in contents)
+			A.clean_blood(radiation_clean = FALSE)	// we invoke the radiation cleaning proc directly
+			A.clean_radiation(clean_factor = 12)	// instead of letting clean_blood do it
```
AffectedArc07, line 419:
```suggestion
			A.clean_radiation(12)	// instead of letting clean_blood do it
```
Again, dont need the named arg

AffectedArc07 (changes requested):

code/game/atoms.dm
TDSSS: I think I was trying to follow the style set by clean_blood there, where the arg seems to always be named. But if you want it changed, sure.

AffectedArc07 (approved):

PR #15399, opened 2021-01-30
Adds the Teleprod to the crafting menu and ups the cost per use

shmabuu:
## What Does This PR Do
This allows the Teleprod to be crafted via the crafting menu via stunprod and a bluespace polycrystal. This also ups the cost of using the teleprod. A high-capacity power cell will use 30% of the cell per use compared to the stunprod using 20% or the baton using 10%. This can be changed if it's not enough of a cost, but I think it's fine.

## Why It's Good For The Game
Allows for variety in stuntools, and a non-lethal option (though it can space people by chance) for players to use. This is also a cool tool that currently can't be made, making it cost more battery to use and a chance to break will make it fit more in the game I think. If it isn't enough cost or too low of a chance to break, it can be adjusted.

## Images of changes
teleprod

## Changelog
:cl:
add: adds the teleprod to the crafting menu
tweak: raised teleprod cost and has limited uses
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/teleprod.dm
+++ b/code/game/objects/items/weapons/teleprod.dm
@@ -4,6 +4,7 @@
 	icon_state = "teleprod_nocell"
 	base_icon = "teleprod"
 	item_state = "teleprod"
+	hitcost = 3000
 	origin_tech = "combat=2;bluespace=4;materials=3"
 
 /obj/item/melee/baton/cattleprod/teleprod/attack(mob/living/carbon/M, mob/living/carbon/user)//handles making things teleport when hit
--- a/code/game/objects/items/weapons/teleprod.dm
+++ b/code/game/objects/items/weapons/teleprod.dm
@@ -6,6 +6,7 @@
 	item_state = "teleprod"
 	hitcost = 3000
 	origin_tech = "combat=2;bluespace=4;materials=3"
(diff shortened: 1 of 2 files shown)


shmabuu:
Decided to make it so it's now limited use. From what I can tell it seems like it's 5-15 uses. If that's too many, the number can be lowered. After about 5-15 uses, the teleprods bluespace crystal will break and turn into a wired rod, and the cell will drop on the ground.

Cocacolagua:
this will be used more by non-antags than antags

datlo:
Bluespace crystals can already be thrown at people as an emergency 'get out of my face' option. I'm not sure adding those as a reusable (even if limited) weapon is a good idea.

shmabuu:
> this will be used more by non-antags than antags

> Bluespace crystals can already be thrown at people as an emergency 'get out of my face' option. I'm not sure adding those as a reusable (even if limited) weapon is a good idea.

Two points that I understand completely, and I don't want to have non antags using this, I doubt anyone else wants to either. Just going to close this as I highly doubt it'll go through. I think it's a cool item, but I don't think it'll be able to be balanced for paradise. Though this was a good experience for me learning more about coding. Thank you Qwerty and SabreML for helping me! Really appreciated. With that, I'm closing this PR.

PR #15404, opened 2021-01-31
Gives better feedback when control or alt-clicking pressure pumps, volume pumps, etc.

### LINKED ISSUE #15400: Ctrl+Clicking and alt+Clicking atmos pumps doesn't give correct feedback
**Issue Description**:

You can ctrl+click and alt+click on a valve (pumps, pressure pumps, air filters, etc) to turn it on/off and max its output. These functions work correctly, but you either get no chat notification or you get the "you can't pull the pump cause it's attached to the floor" chat line.

Get correct feedback like "You turn the pressure valve on/off" and "You max the output on the pressure valve".

**What happened instead**:

"You can't pull the valve you silly"

Obscure keyboard shortcuts that are inmensely helpful for making atmos (and now the SM) less annoying to set up should be easier to discover or learn about and should have appropriate chat feedback when used

Just ctrl+click or alt+click a valve

Probably a long time ago? Might be ever since the functionality was added in #11348

**Extra information**:

SteelSlayer:
## What Does This PR Do

1. Removes the "It's anchored to the floor!" messages when you control-click on pressure pumps, volume pumps, gas filters, and gas mixers. Adds a new chat message for this: "You toggle X on/off".
2. Adds a new chat message for alt-clicking these devices: "You set the target pressure of X to maximum."
3. Small refactor of some variables. I had to unfortunately change a bunch of files because tons of atmos machinery were declaring `var/on` and `var/target_pressure` all over the place. Made these base vars on `/obj/machinery/atmospherics`

## Why It's Good For The Game
Better, and correct feedback to players.

Fixes #15400

## Changelog
:cl:
tweak: Removes the "It's anchored to the floor!" messages when you control-click on pressure pumps, volume pumps, gas filters, and gas mixers
tweak: Adds feedback messages when alt or control-clicking on pressure pumps, volume pumps, gas filters, and gas mixers
/:cl:

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -658,11 +658,11 @@ Class Procs:
  * * user - the mob who is trying to interact with the machine.
  */
 /obj/machinery/proc/can_use_shortcut(mob/living/user)
-	if(!ishuman(user) && !issilicon(user))
-		return FALSE
-	if(!in_range(src, user) && !issilicon(user))
-		return FALSE
 	if(user.incapacitated())
 		to_chat(user, "<span class='warning'>You can't do that right now!</span>")
 		return FALSE
-	return TRUE
+	if(ishuman(user) && in_range(src, user))
+		return TRUE
+	if(issilicon(user))
+		return TRUE
+	return FALSE
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -650,3 +650,19 @@ Class Procs:
 	. = . % 9
 	AM.pixel_x = -8 + ((.%3)*8)
 	AM.pixel_y = -8 + (round( . / 3)*8)
+
+/**
+ * Makes sure the user is allowed to interact with the machine when they use a shortcut, like Control or Alt-clicking.
+ *
+ * Arguments:
+ * * user - the mob who is trying to interact with the machine.
+ */
+/obj/machinery/proc/can_use_shortcut(mob/living/user)
+	if(!ishuman(user) && !issilicon(user))
+		return FALSE
+	if(!in_range(src, user) && !issilicon(user))
+		return FALSE
+	if(user.incapacitated())
+		to_chat(user, "<span class='warning'>You can't do that right now!</span>")
+		return FALSE
+	return TRUE
--- a/code/modules/atmospherics/machinery/atmospherics.dm
+++ b/code/modules/atmospherics/machinery/atmospherics.dm
@@ -366,9 +366,10 @@ Pipelines + Other Objects -> Pipe network
  * * user - the mob who is toggling the machine.
  */
 /obj/machinery/atmospherics/proc/toggle(mob/living/user)
-	if(powered())
-		on = !on
-		update_icon()
+	if(!powered())
+		return
+	on = !on
+	update_icon()
 	if(user)
 		to_chat(user, "<span class='notice'>You toggle [src] [on ? "on" : "off"].</span>")
 
@@ -381,8 +382,9 @@ Pipelines + Other Objects -> Pipe network
  * * user - the mob who is setting the output pressure to maximum.
  */
 /obj/machinery/atmospherics/proc/set_max(mob/living/user)
-	if(powered())
-		target_pressure = MAX_OUTPUT_PRESSURE
-		update_icon()
+	if(!powered())
+		return
+	target_pressure = MAX_OUTPUT_PRESSURE
+	update_icon()
 	if(user)
 		to_chat(user, "<span class='notice'>You set the target pressure of [src] to maximum.</span>")
--- a/code/modules/atmospherics/machinery/atmospherics.dm
+++ b/code/modules/atmospherics/machinery/atmospherics.dm
@@ -20,7 +20,10 @@ Pipelines + Other Objects -> Pipe network
 	on_blueprints = TRUE
 	var/nodealert = 0
 	var/can_unwrench = 0
-
+	/// If the machine is currently operating or not.
+	var/on = FALSE
+	/// The amount of pressure the machine wants to operate at.
+	var/target_pressure = 0
 	var/connect_types[] = list(1) //1=regular, 2=supply, 3=scrubber
 	var/connected_to = 1 //same as above, currently not used for anything
 	var/icon_connect_type = "" //"-supply" or "-scrubbers"
@@ -353,3 +356,33 @@ Pipelines + Other Objects -> Pipe network
 //Used for certain children of obj/machinery/atmospherics to not show pipe vision when mob is inside it.
 /obj/machinery/atmospherics/proc/can_see_pipes()
 	return TRUE
+
+/**
+ * Turns the machine either on, or off. If this is done by a user, display a message to them.
+ *
+ * NOTE: Only applies to atmospherics machines which can be toggled on or off, such as pumps, or other devices.
+ *
+ * Arguments:
+ * * user - the mob who is toggling the machine.
+ */
+/obj/machinery/atmospherics/proc/toggle(mob/living/user)
+	if(powered())
+		on = !on
+		update_icon()
+	if(user)
+		to_chat(user, "<span class='notice'>You toggle [src] [on ? "on" : "off"].</span>")
+
+/**
+ * Maxes the output pressure of the machine. If this is done by a user, display a message to them.
+ *
+ * NOTE: Only applies to atmospherics machines which allow a `target_pressure` to be set, such as pumps, or other devices.
+ *
+ * Arguments:
+ * * user - the mob who is setting the output pressure to maximum.
+ */
+/obj/machinery/atmospherics/proc/set_max(mob/living/user)
+	if(powered())
+		target_pressure = MAX_OUTPUT_PRESSURE
+		update_icon()
+	if(user)
+		to_chat(user, "<span class='notice'>You set the target pressure of [src] to maximum.</span>")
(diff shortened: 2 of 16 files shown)


code/ATMOSPHERICS/components/trinary_devices/mixer.dm
```diff
@@ -6,37 +6,29 @@
...
-		return
-	if(!in_range(src, user) && !issilicon(usr))
-		return
-	if(!ishuman(usr) && !issilicon(usr))
-		return
-	set_max()
-	return
+	if(can_use_shortcut(user))
+		set_max(user)
+	return ..()
```
dearmochi, line 31: Some of the other machineries modified in this PR don't include this call to parent -- I'd probably include it as `/obj/machinery/atmospherics/AltClick()` handles vent crawling.

dearmochi (commented):
Just one thing

code/ATMOSPHERICS/components/trinary_devices/mixer.dm
SteelSlayer: I suppose but, should we call parent here even though the mixer, filter, and pumps don't allow you ventcrawl into them?

dearmochi: If we go up the parents...

https://github.com/ParadiseSS13/Paradise/blob/1d891384364a64f99948658d5124c92cb6e547e7/code/ATMOSPHERICS/atmospherics.dm#L295-L299

https://github.com/ParadiseSS13/Paradise/blob/1d891384364a64f99948658d5124c92cb6e547e7/code/_onclick/click.dm#L334-L343

Turns out alt-clicking would just pass the call to the parent to the parent if not a valid vent crawl object, showing the turf contents on the stat panel. Probably don't want that as it would ruin the flow of the shortcut -- feel free to ignore my comment above.

dearmochi (approved):

code/ATMOSPHERICS/components/trinary_devices/mixer.dm
SteelSlayer: Yeah, and in fact I should probably remove this `..()`, don't know why I added it here, but nowhere else.

AffectedArc07 (approved):

code/game/machinery/machinery.dm
```diff
@@ -650,3 +650,19 @@ Class Procs:
...
+ *
+ * Arguments:
+ * * user - the mob who is trying to interact with the machine.
+ */
+/obj/machinery/proc/can_use_shortcut(mob/living/user)
+	if(!ishuman(user) && !issilicon(user))
+		return FALSE
+	if(!in_range(src, user) && !issilicon(user))
+		return FALSE
+	if(user.incapacitated())
```
SabreML, line 665: This instead maybe, to avoid two `issilicon()` checks?
```suggestion
/obj/machinery/proc/can_use_shortcut(mob/living/user)
	if(ishuman(user))
		if(!in_range(src, user))
			return FALSE
	else if(!issilicon(user))
		return FALSE
	if(user.incapacitated())
```

code/modules/atmospherics/machinery/atmospherics.dm
```diff
@@ -353,3 +356,33 @@ Pipelines + Other Objects -> Pipe network
...
+ *
+ * Arguments:
+ * * user - the mob who is toggling the machine.
+ */
+/obj/machinery/atmospherics/proc/toggle(mob/living/user)
+	if(powered())
+		on = !on
+		update_icon()
+	if(user)
+		to_chat(user, "<span class='notice'>You toggle [src] [on ? "on" : "off"].</span>")
```
SabreML, line 373: Both of these procs still show the `to_chat()` message even if it wasn't powered and nothing changed, so the second part should probably be indented.
```suggestion
/obj/machinery/atmospherics/proc/toggle(mob/living/user)
	if(powered())
		on = !on
		update_icon()
		if(user)
			to_chat(user, "<span class='notice'>You toggle [src] [on ? "on" : "off"].</span>")
```

code/modules/atmospherics/machinery/atmospherics.dm
```diff
@@ -353,3 +356,33 @@ Pipelines + Other Objects -> Pipe network
...
+ *
+ * Arguments:
+ * * user - the mob who is setting the output pressure to maximum.
+ */
+/obj/machinery/atmospherics/proc/set_max(mob/living/user)
+	if(powered())
+		target_pressure = MAX_OUTPUT_PRESSURE
+		update_icon()
+	if(user)
+		to_chat(user, "<span class='notice'>You set the target pressure of [src] to maximum.</span>")
```
SabreML, line 388:
```suggestion
/obj/machinery/atmospherics/proc/set_max(mob/living/user)
	if(powered())
		target_pressure = MAX_OUTPUT_PRESSURE
		update_icon()
		if(user)
			to_chat(user, "<span class='notice'>You set the target pressure of [src] to maximum.</span>")
```

code/modules/atmospherics/machinery/atmospherics.dm
```diff
@@ -353,3 +356,33 @@ Pipelines + Other Objects -> Pipe network
...
+ * Maxes the output pressure of the machine. If this is done by a user, display a message to them.
+ *
+ * NOTE: Only applies to atmospherics machines which allow a `target_pressure` to be set, such as pumps, or other devices.
+ *
+ * Arguments:
+ * * user - the mob who is setting the output pressure to maximum.
+ */
+/obj/machinery/atmospherics/proc/set_max(mob/living/user)
+	if(powered())
+		target_pressure = MAX_OUTPUT_PRESSURE
```
SabreML, line 385: No shortcut to set it to the minimum?

code/modules/atmospherics/machinery/components/binary_devices/pump.dm
```diff
@@ -21,51 +21,31 @@ Thus, the two variables affect pump operation are set in New():
...
-	if(!ishuman(usr) && !issilicon(usr))
-		return
-	set_max()
-	return
+	if(can_use_shortcut(user))
+		set_max(user)
 
 /obj/machinery/atmospherics/binary/pump/AIAltClick()
 	set_max()
 	return ..()
```
SabreML, line 47: Because of this line in `ai.dm`, if the AI alt clicks on a pump `set_max()` will be called twice. Once with a `user` argument and once without:
https://github.com/ParadiseSS13/Paradise/blob/fe6e445760ade9158e5fcf84346b11d1cfb15d42/code/_onclick/ai.dm#L159-L160
I'd recommend just changing that to just `return` like all of the other procs there, but since it was added in 21a9d929304cc8c742369447c0c443028e879c09 7 years ago it might be tied into some other interactions, so probably best to leave that for another PR unless you can make sure there won't be any side effects.

TLDR: You can get rid of the `return ..()` and add a user argument here and on the others to make it work properly, but since it's only setting it to maximum it doesn't actually do anything bad.
```suggestion
/obj/machinery/atmospherics/binary/pump/AIAltClick(mob/user)
	set_max(user)
```

code/modules/atmospherics/machinery/components/binary_devices/volume_pump.dm
```diff
@@ -21,51 +21,31 @@ Thus, the two variables affect pump operation are set in New():
...
-	if(!ishuman(usr) && !issilicon(usr))
-		return
-	set_max()
-	return
+	if(can_use_shortcut(user))
+		set_max(user)
 
 /obj/machinery/atmospherics/binary/volume_pump/AIAltClick()
 	set_max()
 	return ..()
```
SabreML, line 47: Same issue here, called twice by AIs.
```suggestion
/obj/machinery/atmospherics/binary/volume_pump/AIAltClick(mob/user)
	set_max(user)
```

code/modules/atmospherics/machinery/components/trinary_devices/filter.dm
```diff
@@ -31,46 +31,27 @@
...
-	if(!ishuman(usr) && !issilicon(usr))
-		return
-	set_max()
-	return
+	if(can_use_shortcut(user))
+		set_max(user)
 
 /obj/machinery/atmospherics/trinary/filter/AIAltClick()
 	set_max()
 	return ..()
```
SabreML, line 53:
```suggestion
/obj/machinery/atmospherics/trinary/filter/AIAltClick(mob/user)
	set_max(user)
```

code/modules/atmospherics/machinery/components/trinary_devices/mixer.dm
```diff
@@ -6,37 +6,28 @@
...
-		return
-	if(!ishuman(usr) && !issilicon(usr))
-		return
-	set_max()
-	return
+	if(can_use_shortcut(user))
+		set_max(user)
 
 /obj/machinery/atmospherics/trinary/mixer/AIAltClick()
 	set_max()
```
SabreML, line 33: Can't select this one properly to suggest changes, but you get the idea.

code/modules/atmospherics/machinery/atmospherics.dm
SteelSlayer: Control click and alt click are already taken so I wouldn't be able.

SabreML: Oh true, fair enough.

code/game/machinery/machinery.dm
SteelSlayer: Yeah my logic was crap here. I'm rewriting it a bit differently than you suggested though.
```dm
/obj/machinery/proc/can_use_shortcut(mob/living/user)
	if(user.incapacitated())
		to_chat(user, "<span class='warning'>You can't do that right now!</span>")
		return FALSE
	if(ishuman(user) && in_range(src, user))
		return TRUE
	if(issilicon(user))
		return TRUE
	return FALSE

```

code/modules/atmospherics/machinery/components/binary_devices/pump.dm
SteelSlayer: Yeah good catche on this stuff.

variableundefined (changes requested):
DM on discord when deconflicted with link for merge.

variableundefined (approved):

PR #15408, opened 2021-02-01
Gives shadowlings the ability to destroy mindshields

ppi13:
## What Does This PR Do

Alternative to #14956 and an attempt to fix the merge conflict in #14965
After #14521 was added over a month ago, we got a severe case of I ded pls nerf, where shadowlings lost every round, which honestly, was fine for me. However, on the forum there was a thread for this as well, and there was a pretty good suggestion, where destroying mind shields is part of the progressive powerup.  
Currently it takes 30 seconds to destroy an implant with a 90 second cooldown.
In an average round of 100 players a sling will learn this ability at somewhere between 6-7 thralls.

## Why It's Good For The Game

This is a better alternative to reverting #14521
This should also halt shadowling turning into murderboner security and then shadowlings lose. 

## Changelog
:cl: ppi
add: Add a mindshield melt ability to shadowlings
/:cl:


Spacemanspark:
What

PR #15415, opened 2021-02-02
Some Admin/Mentor help improvements

SabreML:
## What Does This PR Do

1. Made the 'Adminhelp' input window change to 'Mentorhelp' if applicable.
2. Adds a `ticket_converted` variable to admin/mentor tickets to prevent them from being converted by multiple people.
*(This doesn't stop converted tickets from being converted back, since converting makes a new ticket.)*
3. Added a yellow 'CONVERTED' status message if the mentor/admin ticket has been converted.

## Why It's Good For The Game

1. Stops the worry that you might have selected the wrong ticket category by accident.
2. Stops potential spam, since it makes a new ticket each time it's clicked.

## Images of changes

**Ahelp**
ahelp

**Mhelp**
mhelp

**Converted**
converted

## Changelog
:cl:
tweak: Choosing 'Mentorhelp' from the ahelp selection list now changes the window title to 'Mentorhelp', to prevent confusion.
fix: Fixed Admin/Mentor tickets being converted to the other type more than once.
/:cl:

### DIFF
--- a/code/controllers/subsystem/tickets/tickets.dm
+++ b/code/controllers/subsystem/tickets/tickets.dm
@@ -286,7 +286,7 @@ SUBSYSTEM_DEF(tickets)
 
 /datum/controller/subsystem/tickets/proc/assignStaffToTicket(client/C, N)
 	var/datum/ticket/T = allTickets[N]
-	if(T.staffAssigned != null && T.staffAssigned != C && alert("Ticket is already assigned to [T.staffAssigned.ckey]. Are you sure you want to take it?","Take ticket","No","Yes") != "Yes")
+	if(T.staffAssigned != null && T.staffAssigned != C && alert("Ticket is already assigned to [T.staffAssigned.ckey]. Are you sure you want to take it?", "Take ticket", "Yes", "No") != "Yes")
 		return FALSE
 	T.assignStaff(C)
 	return TRUE
--- a/code/controllers/subsystem/tickets/tickets.dm
+++ b/code/controllers/subsystem/tickets/tickets.dm
@@ -172,6 +172,9 @@ SUBSYSTEM_DEF(tickets)
 	if(!other_ticket_system_staff_check())
 		return
 	var/datum/ticket/T = allTickets[ticketId]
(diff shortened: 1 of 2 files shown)


hal9000PR:
I have doubled back to check if i selected the right type atleast 5 or so times. Thank you.

code/controllers/subsystem/tickets/tickets.dm
```diff
@@ -328,6 +347,8 @@ SUBSYSTEM_DEF(tickets)
 
 //Return the ticket state as a colour coded text string.
 /datum/ticket/proc/state2text()
+	if(ticket_converted)
```
farie82, line 350: wouldn't this also make it show up as converted when it's resolved etc?

AffectedArc07:
While youre doing this, can you also fix the inconsistency of Yes/No in all the ticket related boxes? 

Every other button in the game has yes on the left and no on the right
image
image
image

Whereas ticket related buttons have it the other way round
image

farie82:
> Whereas ticket related buttons have it the other way round
> image

I've made that on purpose since "No" would be the most probable answer a user would give on that prompt

AffectedArc07:
> > Whereas ticket related buttons have it the other way round
> > image
> 
> I've made that on purpose since "No" would be the most probable answer a user would give on that prompt.
> Meaning a simple enter would suffice to continue on

Its still inconsistent design and will cause more people to hit `Yes` since its where `No` is for every other popup.

code/controllers/subsystem/tickets/tickets.dm
SabreML: It `return`s the status text, so it won't get to the others if it's converted.

farie82: Aye so resolved tickets still are labeled converted. Should be fine due to the filters but does appear a bit weird

Kyep (approved):

AffectedArc07 (changes requested):
Requesting the buttons be swapped

AffectedArc07 (approved):

PR #15419, opened 2021-02-02
Gives the Blueshield Head of Staff office access

SabreML:
## What Does This PR Do

Gives the Blueshield access to the RD, CE, CMO, HoS and HoP offices, so that they can enter them without needing the Spare.

## Why It's Good For The Game

The vast majority of rounds the Blueshield is given the Captain's Spare ID for better access, and I personally have only ever used it to get into the HoP and Captain offices.
While this is technically a pretty big buff, I feel like it'll reduce the amount of Blueshields getting given all access.

## Changelog
:cl:
tweak: The Blueshield now has access to the Head of Staff offices.
/:cl:


FlattestGuitar:
I feel like it won't.

Cocacolagua:
we can dream

Gatchapod:
Big yes on this one. Blueshield SOP suggests protection of the offices is part of their job, so having access to those seems most appropriate.
Small note, though: Magistrate and Representative are also under Blueshield protection and I think you should add access to their offices as well.

SteelSlayer (approved):

Fox-McCloud:
This has a lot more reaching implications than just accessing their offices---anything with, say, "access RD", the Blueshield can now do.

There's a number of things only a particular head can do and it's always tied to `ACCESS_[TITLE]`.

I don't think giving this to the Blueshield is a good idea, because of that----nor do I think making an entirely new access, just for this, is a good idea, either.

TheYeetster1:
I play BlueShield a lot, and one of the main reasons I think captains give you the spare ID is so you can go into heads office's during emergencies. BlueShield is also Karma locked, so I assume that people who get 30 karma can roleplay and not power game. I don't see a lot of power game potential since most captains just give the spare anyway.

As a side note, I think it would make BlueShield more balanced if he didn't get the spare headset and spare ID. This PR might make it less common for captains to give spares out.

Spacemanspark:
Could also just make it against SoP.
Not sure how I feel about this PR

TheYeetster1:
@Spacemanspark I think just making it against SOP would just encourage Blueshields to be bridge hobos+, Its already incredibly boring only standing on the bridge. I think this PR would make playing BlueShield more fun, also for antags, since you don't always know where the BlueShield is.

I am definitely in support of making it against SOP for the Blueshield to get the Spare ID, Spare Headset and Hand Teleporter.

Just don't make the blueshield more worthless than they already are. Encourage them to go around and **Protect the heads of staff** other than the captain, hop and whoever happens to be on the bridge. And I think giving the Blueshield access to head's offices would help, aswell as giving radio access to the sec channel, just for calling for help in a secure way.

I think that would a nice compromise.

Kyep:
The correct solution to blueshields being unhappy at not being simply handed the spare ID at roundstart... is not to make them start off with ACCESS_CAPTAIN.

SabreML:
> The correct way to solve this (if it even needs to be solved, which I'm not convinced it does)...

> While we're at it, I wish Blueshields would stop asking for the spare ID at round start, on green, with no threats. It is the Captain's spare ID... not the Blueshield's.

These two are very much related, at least for me. Being unable to access, say, the CE's office while they're being actively drained by a vampire is really quite annoying when you're meant to be protecting them.

I'll change this to what you suggested. I hadn't really considered things like the Communications console, since with the Spare you have access to it anyway. Definitely shouldn't be part of the BS's ID though. 👍

Denthamos:
Yeah, giving the blueshield that level of access by default is not a good path to go.  In general they should not be given all access either by the captain due to what exactly said access lets them do as its more than just opening doors.  Now its using certain machines, and opening things that really they should not be for their role.

SabreML:
>In general they should not be given all access either by the captain

In theory yes, but... they are anyway.

>Now its using certain machines, and opening things that really they should not be for their role.

I've changed it to only be the office doors, so that should resolve this issue.

AffectedArc07:
2 maintainers and a head against, no one in favour. Closing.

datlo:
Both Kyep and fox objected because the PR gave actual access instead of editing doors. The PR was modified to edit doors instead of giving access. Why was this closed?

PR #15422, opened 2021-02-02
RustG version + log format unit tests

AffectedArc07:
## What Does This PR Do
Adds unit tests to ensure that:
- RUSTG is the correct version for the codebase
- Logs are being written in the correct format for us (Prevents people copypasting TG's RUSTG DLL in, ours is different)

## Why It's Good For The Game
Sanity tests are good

## Changelog
N/A, no ingame impact

### DIFF
--- a/tools/ci/validate_rustg_windows.py
+++ b/tools/ci/validate_rustg_windows.py
@@ -13,7 +13,7 @@ ci_testing_text = "This is a test message"
 
 # Helpers
 def success(msg):
-    print("[✓] {}".format(msg))
+    print("[Y] {}".format(msg))
 
 def fail(msg):
     print("[X] {}".format(msg))
--- a/tools/ci/validate_rustg_windows.py
+++ b/tools/ci/validate_rustg_windows.py
@@ -68,6 +68,7 @@ rustg_dll.log_write.argtypes = [c_int, c_char_p * 2]
 
 timestamp = datetime.now()
 
+# Generate valid results for the time now and the next 2 seconds. This is so we can account for spurious CI lag.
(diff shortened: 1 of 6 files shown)


tools/ci/validate_rustg_windows.py
```diff
@@ -0,0 +1,90 @@
...
+
+# Now test log writing. This hurt to write.
+string_array = c_char_p * 2
+sa = string_array(bytes(ci_log_file, "ascii"), bytes(ci_testing_text, "ascii"))
+rustg_dll.log_write.argtypes = [c_int, c_char_p * 2]
+
+timestamp = datetime.now()
+
+valid_results = []
+for count in range(3):
```
farie82, line 72: best to add reasoning as to why you do this here in the form of a comment.

farie82 (commented):
One small thing. Also I'll restate my issue with unit tests like these. Features like these won't change often. Maybe once every 3 months. Yet the unit test does add overhead to the CI setup.
Prob out of scope for now but adding more intelligence to the system would be nice. Only firing these tests when the DLL/SO gets changed for instance

AffectedArc07:
> One small thing. Also I'll restate my issue with unit tests like these. Features like these won't change often. Maybe once every 3 months. Yet the unit test does add overhead to the CI setup.
> Prob out of scope for now but adding more intelligence to the system would be nice. Only firing these tests when the DLL/SO gets changed for instance

The unit tests DM side happen pretty much instantly, and the windows test runs in parallel so it doesnt take extra time.

farie82 (approved):

Issue #15425, opened 2021-02-03
Full tile window bugs


TheYeetster1:
**Issue Description**:

Pull #15385 Introduced a few bugs, 

External solar airlocks cannot be opened by personnel now. If you try to click on the button to go outside or inside you just knock on the glass. 

If you try to alt click to view the tile it gives you the message that you cant rotate the glass.

A̶l̶s̶o̶ ̶c̶o̶l̶o̶r̶e̶d̶ ̶g̶l̶a̶s̶s̶ ̶I̶ ̶t̶h̶i̶n̶k̶ ̶h̶a̶s̶ ̶d̶i̶s̶a̶p̶p̶e̶a̶r̶e̶d̶.̶ ̶I̶ ̶r̶e̶m̶e̶m̶b̶e̶r̶ ̶s̶e̶c̶u̶r̶i̶t̶i̶e̶s̶ ̶g̶l̶a̶s̶s̶ ̶b̶e̶i̶n̶g̶ ̶r̶e̶d̶.̶

**(Never mind killing tinted windows was intentional)**

(solars)
![solars](https://user-images.githubusercontent.com/70362590/106686869-17e1ae80-6599-11eb-930c-c3dd5bd3f623.png)

(blue windows in security, previously red windows.)
![bluewindows](https://user-images.githubusercontent.com/70362590/106686978-511a1e80-6599-11eb-9dd9-38e69d12195a.png)

**What did you expect to happen**: 

For me as an engineer to go wire solars without asking the AI to open it.

**What happened instead**:

**Why is this bad/What are the consequences**:

The AI has to open Solars for engineers,

Also red tint to security is gone, sad.
**(Never mind killing tinted windows was intentional)**

**Steps to reproduce the problem**:

Try to wire solar, using the external airlocks.

**When did the problem start happening**:

Since #15385 was merged

**Extra information**:

Fox-McCloud:
security windows aren't meant to be colored, because coloring fulltiles doesn't look...good.

PR #15426, opened 2021-02-03
Adjust Airlock Buttons Layers

Fox-McCloud:
Adjust the layer of of several airlock related buttons/infrastructure, so they're above windows and you press them, once again.

fixes: https://github.com/ParadiseSS13/Paradise/issues/15425

:cl: Fox McCloud
fix: Fixes the layering of airlock related buttons/infrastructure so you can interact with it, once again
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock_control.dm
+++ b/code/game/machinery/doors/airlock_control.dm
@@ -146,6 +146,7 @@
 /obj/machinery/airlock_sensor
 	icon = 'icons/obj/airlock_machines.dmi'
 	icon_state = "airlock_sensor_off"
+	layer = ABOVE_WINDOW_LAYER
 	name = "airlock sensor"
 	anchored = 1
 	resistance_flags = FIRE_PROOF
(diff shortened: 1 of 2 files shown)


SteelSlayer:
The issue that reported this bug didn't mention intercoms were affected too, but the I noticed them in one of the screenshots. I think it's safe to also give them the above window layer.

image

Also can you put 'Fixes #15425' in your description?

Fox-McCloud:
@SteelSlayer I actually tested those, and they weren't accessible before, either.

SteelSlayer:
@Fox-McCloud Oh alright. Should their layer remain the same then? And we'll just avoid mapping intercoms onto windows?

Fox-McCloud:
>Oh alright. Should their layer remain the same then? And we'll just avoid mapping intercoms onto windows?

Generally, yeah. There's a few exceptions--most of the intercomms currently in windows are acting more like broadcast systems than actual intercoms.

PR #15427, opened 2021-02-03
Further Updates Windows

Fox-McCloud:
Just a further update to windows.

- Adds plasma basic, tinted , plastitanium, ice, and survival pods spawners
- Refactors Plastititanium windows being a child of `/obj/effect/spawner/window/full`
- Fixes there being a `/obj/effect/spawner/window/shuttle` type by making pod windows to be the proper typepath
- Converts over existing plastititanium windows (only on syndicate base) to be spawners
- Fixes brass windows having a missing sprite
- Updates tinted window sprites to be in line with the new, more rounded look

:cl: Fox McCloud
tweak: improves some window spawner code
/:cl:


code/modules/mining/equipment/survival_pod.dm
```diff
@@ -70,23 +70,12 @@
...
-	fulltile = TRUE
-	flags = PREVENT_CLICK_UNDER
-	reinf = TRUE
-	heat_resistance = 1600
-	armor = list("melee" = 50, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 50, "bio" = 100, "rad" = 100, "fire" = 80, "acid" = 100)
 	smooth = SMOOTH_MORE
-	canSmoothWith = list(/turf/simulated/wall/mineral/titanium/survival, /obj/machinery/door/airlock/survival_pod, /obj/structure/window/shuttle/survival_pod)
-	explosion_block = 3
-	level = 3
-	glass_type = /obj/item/stack/sheet/titaniumglass
```
Fox-McCloud, line 88: re-add

CkeyLurapa:
In this PR/Update of Windows.

How would the OP feel about FT (Fulltile) windows being dislodged (unbolted) when their supporting structure, their Grille, is destroyed?

I ask this because of a Nuke Operative round. A bomb went off around the Bridge. And what would normally result in that area being vented to the existing space. Would result in the Grille being destroyed, and the combined directionals into FT, be damaged/slightly damaged.

It would be balanced, in my opinion, since FT are quite durable. And Grilles are only interactable /hittable if the Window itself is removed/shattered. 

So in the case of Bombs. Which can hit the grille before the window. Have either the Grille be immune to the bomb (if window exists) or have the Window disloge / unbolt if their Grille is shattered.

Fox-McCloud:
That's outside of the scope of the PR, but that's just how things function now---there's not a good way to ensure one thing vs another gets hit first.

It's just an overall change to get used to, really---full  tiles have the same HP as two windows, so make breaking through roughly the same duration of time (which is primarily what's going to happen). Explosions are a rare exception to this, due to how they work, but making a whole bunch of exceptions *just* for that is a bit peculiar and unideal.

AffectedArc07 (approved):

PR #15431, opened 2021-02-03
Removes Shaft Miners and the Mechanic as possible Contractor targets

SabreML:
## What Does This PR Do

Adds 'Mechanic' and 'Shaft Miner' to the list of excluded jobs for Contractor targets.

## Why It's Good For The Game

It's hard to kidnap someone alive when they're gibbed on another Z-level.

## Changelog
:cl:
tweak: Shaft Miners and the Mechanic can no longer be given as Contractor targets.
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/contractor/datums/objective_contract.dm
+++ b/code/modules/antagonists/traitor/contractor/datums/objective_contract.dm
@@ -8,6 +8,8 @@
 	/// Jobs that cannot be the kidnapping target.
 	var/static/list/forbidden_jobs = list(
 		"Captain",
+		"Mechanic",
+		"Shaft Miner",
 	)
 	/// Static whitelist of area names that can be used as an extraction zone, structured by difficulty.
 	/// An area's difficulty should be measured in how crowded it generally is, how out of the way it is and so on.


TDSSS:
Should have considered that before they bought the hard mode kit, imo. And contractors have multiple contracts for a reason. Not a fan

hal9000PR:
What if instead, contractor pinpointers could point to brains? Then you could find gibbed miners on lavaland.

Cocacolagua:
what if instead we removed miners and mechanic from being targetted by all objectives 😳

Fox-McCloud:
Mmm...I kind feel this is one of those "the chips fall where they may" deals, as unfortunate as it is.

Changelings who have to steal someone's identity or brain, or other such similar objectives just seem like one of those unfortunate things that crop up, every now and again.

Kyep:
Not a fan of this.

Why can't you just kill/kidnap/etc them when they return to the station?
If they never return, they're probably dead.

Not to mention, miners are some of the best geared crew. To additionally protect them from even being targeted by antags is... not a great look. Miners already get way too much preferential treatment.

Spacemanspark:
I had considered both sides of this argument and while I do believe both have some merit, Kyet makes a good point. If anything, miners should probably be a bit more of a priority to antags, given that they get some of the best gear in the game alongside how important they are to science.

SabreML:
> Why can't you just get them when they return to the station?
> If they never return, its probably because they're dead. And we don't generally protect contractors from targeting people who then die.

If someone on-station dies then sure, you can maybe get their body from the morgue in the worst case scenario. If a mechanic dies at the Depot though, realistically you have no chance of ever getting them.

> Not to mention, miners are some of the best geared crew. To additionally protect them from even being targeted by antags is... not a great look. Miners already get way too much preferential treatment.

If anything I would argue the opposite to this, having your target be a miner in drake armour with a cleaving saw and 10 legion cores is never fun.

Either way though, there's always the option of rerolling your contracts unless you get incredibly unlucky and get all shaft miners or something. *(Which is what prompted this PR in the first place.)*
I'll close this since it's definitely not getting merged, but I still think that the Mechanic at least would be a good job to blacklist.

PR #15057, opened 2020-12-06
Initialized Storage Items

SabreML:
## What Does This PR Do

Replaces all instances of `New()` with `Initialize()` and `populate_contents` for storage items.

Also cleaned up some of the code a bit because why not.

## Why It's Good For The Game

Better and more flexible code.

## Changelog
:cl:
tweak: Changed all storage items to use `Initialize()` rather than `New()`. Should have no ingame effect.
tweak: Stock part boxes now stack items, rather than having each item in a separate spot
/:cl:

PR #15434, opened 2021-02-03
Makes belts use Initialize()

SteelSlayer:
## What Does This PR Do
Makes belts use `Initialize()` instead of `New()`

## Why It's Good For The Game
Killing off more `New()` is good.

## Changelog
N/A


SteelSlayer:
Just kidding, we have this  #15057

PR #15436, opened 2021-02-03
Purging Will Continue Until Simulations Improve

Fox-McCloud:
Replaces the unsimulated turfs on all away missions (minus beach) and ruins with simulated variety.

Where a wall turf was meant to be impassable and indestructible, it was replaced with `/turf/simulated/wall/indestructible`

- Fixes simulated abductor turfs having no icon
- Fixes the syndicate cake factory having aired turfs in the middle of an airless area
  - Also why did this away mission use fake glass for no reason at all? This made no sense. To be fair, this is an awful away mission

. . . the way must be prepared . . .

:cl: Fox McCloud
tweak: Removes most unsimulated turfs with the exception of CentComm and Beach
/:cl:

### DIFF
--- a/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_cube.dmm
+++ b/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_cube.dmm
@@ -10,7 +10,7 @@
 /turf/simulated/floor/plating/asteroid/basalt/lava_land_surface,
 /area/lavaland/surface/outdoors)
 "d" = (
-/turf/unsimulated/wall,
+/turf/simulated/wall/indestructible/riveted,
 /area/lavaland/surface/outdoors)
 "e" = (
 /obj/machinery/wish_granter,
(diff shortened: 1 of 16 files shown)


Darkmight9:
Thank you for not touching the beach gateway I do _not_ want to have to deal with that big of a scale of a merge conflict 😓

Spacemanspark:
Can I ask why we're removing unsimulated turfs entirely? I find them useful when testing or doing small ghost related events in the admin test room/ thunder dome.

Fox-McCloud:
>Can I ask why we're removing unsimulated turfs entirely? I find them useful when testing or doing small ghost related events in the admin test room/ thunder dome.

Unsimulated turfs have a lot of issue with them, to say the least. They don't behave like most other turfs, you can't make them slippery, decals fail to render on them, people misplace them on actual maps, and they kinda stand in the way of a fully simulated atmospherics system. They also require a crap-ton of special snowflake code in gas mixtures and atmos code just so simulated turfs can interact with them properly. The only real advantage they have is that they're indestructible---which is...something simulated turfs can do, just fine.

AffectedArc07 (approved):

PR #15438, opened 2021-02-04
Removes `pixel_* = 0` and `tag = "icon*"` from the map files

SabreML:
## What Does This PR Do

Removes ALL instances of `pixel_* = 0` and `tag = "icon*"` in the map files, and adds CI checks to prevent any more from being added.
Also resaves every map file using StrongDMM's 'Sanitise Variables' feature, just for the sake of doing it all in one go.

Regex used:
```
^.*(\tpixel_. = 0;).*\n

(;\n\tpixel_. = 0).*

\{\n\tpixel_. = 0\n\t\}
```
```
^.*(\ttag = "icon.*;).*\n

(;\n\ttag = "icon).*

\{\n\ttag = "icon.*\n\t\}
```

## Why It's Good For The Game

It has no effect in-game, and adds even more bloat to the map files.

## Changelog
:cl:
tweak: Added some map-related CI checks to reduce file bloat.
/:cl:


FreeStylaLT:
image
image

SabreML:
you see nothing
image

SabreML:
## Steps to deconflict this PR once it's merged:
### 1: Regex replacement
Code_iO7fwIOAVM
1. Click the 'Use Regular Expressions' button so that it turns blue.
2. Copy one line from here into the search bar
```
^.*(	pixel_. = 0;).*\n

(;\n	pixel_. = 0).*

\{\n	pixel_. = 0\n	\}
```
3. Click 'Replace All' with nothing in the input box, and confirm.
4. Repeat step 3, with the next two lines of regex.
### 2: pixel_z
For some reason this timer specifically has `pixel_z = 3`. That variable is, as far as I'm aware, only used for isometric BYOND games.
image
This will make the CI unhappy since it's not meant to be there, so to fix it you just have to remove that line and the semicolon above.
image

Kyep (commented):
I'm not sure about this.
On the one hand, sure, its a little bit of extra bloat in the map files.
On the other hand, the size of map files doesn't really matter, mapping tools might add pixel_*=0 in some cases and it would be annoying if that made CI fail, and I'm really not sure what this improves other than compliance with OCD for the very, few people who read the map files by hand.

SabreML:
I would consider this around the same as any other backend cleanup stuff. While technically it's not necessary it might still help a bit.

I do agree that the point of the CI failing because of something out of your control is a pretty big one though, I could try making it remove the pixel variables automatically from mapmerge or something, since the whole thing is just three regex replaces.

AffectedArc07:
>I could try making it remove the pixel variables automatically from mapmerge or something, since the whole thing is just three regex replaces.

Do not do this. Tools should **not** modify a users file in ways that could modify a map away from original intent (IE: Potentially removing a legitimate `pixel_y`). Given how flakey DMMs can be (Trailing commas on ends of declarations, potential parenthesis, etc), I am not in favour of automatic replacement of things, because I dont trust regex find/replace by its nature and how it does not take specific things into account. Prime example of this was moving `var/const` to `#define`. That entire PR was hand done because blindly trusting regex is a bad idea for replacements. Matching is just fine, but I would never use it for replacements.

SabreML:
MDB really doesn't like this PR very much, does it?

SabreML:
Tested with various configurations on my end in 13 rounds, and I can't reproduce the runtimes.
I'll probably look into it again tomorrow.

S34NW (commented):
I can't really say I'm a fan of this approach, personally.

The better option would just be to save all the maps in SDMM with "Sanitize variables" enabled, then any pixel_* that is the same as default will get removed.
Some items will actually need pixel_* = 0 if their default value is something else and they need to be changed. (This can be seen in cell brig timers, for example.)

SabreML:
> The better option would just be to save all the maps in SDMM with "Sanitize variables" enabled, then any pixel_* that is the same as default will get removed.

Assuming you're referring to the regex replace, the 'Sanitize Variables' setting does seem to have the same effect as that and would definitely be the better option going forwards, but it currently it would be a lot slower (though possible) to go through the 39 files that this modifies and re-save them for the same effect, and the initial sanitize will conflict with pretty much every map related PR considering the size of this one.

If you're referring to the CI check instead, then I personally don't see any downsides in having it.
If it never gets used because people follow the mapping standards then great, the CI passing is always a good thing.
If someone adds a bunch of redundant variable changes to their map PR and the checks fail, then that means they need to change it back themselves, rather than someone else having to have phantom diffs in their PR the next time the map is touched.
(Assuming #16125 gets merged)

> Some items will actually need pixel_* = 0 if their default value is something else and they need to be changed. (This can be seen in cell brig timers, for example.)

In most cases I'd say that having a pixel offset set by default on an object is just bad practice. With an exception going to subtypes like the brig cell timers and APCs from #15665, since if you need them to be perfectly centered you can just place the parent type instead.

S34NW:
I was indeed referring to the regex replace, the CI check sounds good to me.

AffectedArc07:
MDB did succeed at rendering this, just the check refused to report
image

AffectedArc07 (approved):
I manually reviewed all of this.

I am in pain.

PR #15439, opened 2021-02-04
SM engine APCs no longer short out during APC short/overload events

### LINKED ISSUE #15394: The SM chamber APC is targeted by the APC shorted event
**Issue Description**:
The APC inside the SM chamber, is targeted by the APC shorted event, causing the air alarm to stop working and SM to delaminate.

That the SM APC was protected from the event as the Engine APC was before, and thus would be unaffected.

**What happened instead**:
The SM APC went down, the air alarm stopped, and SM started to delaminate. This combined with  a Telecomms APC down meant SM started delaminating with no warning for a time.

This event may cause a delamination, which I assume isnt what the event is meant to do, just as it was not meant to cause a loose engine in the past. As it is, using the event will always leave a risk of a delamination, which is a good reason not to trigger the event for admins, as well as an unexpected thing for the crew.

Trigger the APC shorted event till the SM chamber APC is shorted, it should happen over some rolls.

Whenever the SM became the engine

**Extra information**:

Kyep:
## What Does This PR Do
Stops the APC short/overload events from disabling the SM that powers SM cooling.

## Why It's Good For The Game
Fixes #15394 

## Changelog
:cl: Kyep
fix: APC short/overload events can no longer delaminate the SM.
/:cl:

### DIFF
--- a/code/modules/events/apc_overload.dm
+++ b/code/modules/events/apc_overload.dm
@@ -29,6 +29,7 @@
 /proc/apc_overload_failure(announce=TRUE)
 	var/list/skipped_areas_apc = list(
 		/area/engine/engineering,
+		/area/engine/supermatter,
 		/area/turret_protected/ai)
 
 	if(announce)
--- a/code/modules/events/apc_short.dm
+++ b/code/modules/events/apc_short.dm
@@ -29,6 +29,7 @@
 /proc/power_failure(announce=TRUE)
 	var/list/skipped_areas_apc = list(
 		/area/engine/engineering,
+		/area/engine/supermatter,
 		/area/turret_protected/ai)
 
 	if(announce)


SabreML:
Personally I consider this more of a feature than a bug. It adds a bit of panic when the event happens rather than the standard bored resignation.

TDSSS:
An APC overloading doesn't actually case the engine to delam unless engineering is totally incompetent or is running the SM in some crazy unstable configuration. Under normal circumstances, they should have plenty of time to notice and fix the issue.

Not the biggest fan.

Kyep:
I've seen it cause the engine to delam, as it disables the pumps which are being used to cool the engine.
The old tesla/SM engine was protected from this.
For consistency, the new one probably should be too.

AffectedArc07 (approved):
Old engineering was protected, this should be too

PR #15441, opened 2021-02-04
Cleans up Gas Mixture

Fox-McCloud:
Cleans up gas mixtures so that when you search for the proc in question, it takes you to the code in question, instead of an empty proc.

This also re-organizes things a bit and sets `TRUE` and `FALSE` where possible.

I also moved `fire` into `react` directly---there's not really any good reason to have be its own proc---this will also slightly improve performance, by reducing proc call overhead, during a fire.

:cl: Fox McCloud
tweak: refactored gas mixtures a bit
/:cl:

### DIFF
--- a/code/datums/gas_mixture.dm
+++ b/code/datums/gas_mixture.dm
@@ -519,41 +519,40 @@ What are the archived variables for?
 				reacting = TRUE
 
 	fuel_burnt = 0
+	//Handle plasma burning
 	if((toxins > MINIMUM_MOLE_COUNT) && (oxygen > MINIMUM_MOLE_COUNT) && temperature > FIRE_MINIMUM_TEMPERATURE_TO_EXIST)
 		var/energy_released = 0
 		var/old_heat_capacity = heat_capacity()
+		var/plasma_burn_rate = 0
+		var/oxygen_burn_rate = 0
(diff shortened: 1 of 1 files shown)


code/datums/gas_mixture.dm
```diff
@@ -562,58 +431,151 @@ What are the archived variables for?
...
+				nitrogen += burned_fuel
+
+				var/new_heat_capacity = heat_capacity()
+				if(new_heat_capacity > MINIMUM_HEAT_CAPACITY)
+					temperature = (temperature * old_heat_capacity + energy_released) / new_heat_capacity
+				reacting = TRUE
+
+	fuel_burnt = 0
+	//Handle plasma burning
+	if((toxins > MINIMUM_MOLE_COUNT) && (oxygen > MINIMUM_MOLE_COUNT) && temperature > FIRE_MINIMUM_TEMPERATURE_TO_EXIST)
```
Fox-McCloud, line 523: the logic of this is a bit different, and inspired from TG; checking that there's enough plasma and oxygen in the air means that we're skipping on on a bit of math and logic if we wind up with situations where there's not much plasma, but it is very hot.

AffectedArc07 (approved):

PR #15444, opened 2021-02-05
Fixes HUD sunglasses deconstruction

### LINKED ISSUE #14587: Disassembling Hudsunglasses
**Issue Description**:

Disassembling Hudsunglasses won't give you the HUD part back. And disassembling Skill Hudsunglasses cannot be done

Being able to disassemble security hudsunglasses can come in handy when you haven't been outed yet and didn't have flash protection yet. But without getting the HUD itself it's kinda moot. 

And I want to get rid of the awful skill hudsunglasses and just go with a normal pair that doesn't look like ugly mesons.

**What happened instead**:

You only get the sunglasses. the HUD goes POOF

I want my normal sunglasses baaaaaaack. And I don't really like the crutch that is any HUD, since it always enlarges your opponent's hitbox and yours.

Try to disassemble security hudsunglasses. Get only sunglasses in return. No recipe to disassemble skill hud sunglasses.

Probably from the start we could disassemble them.

**Extra information**:

The skillhud is ugly as sin. :sob:

SabreML:
## What Does This PR Do

Slightly refactors crafting recipe results to allow for more than one item to be made.
Fixes #14587

## Why It's Good For The Game

It makes sense that reversing the construction on HUD sunglasses would drop both the HUD and the sunglasses.
Also crafting being able to drop more than one item is just a good thing in general.

## Changelog
:cl:
tweak: Deconstructing HUD sunglasses now drops both the HUD and the sunglasses.
tweak: Renamed the 'Beer Goggles' crafting recipe to 'Sunscanners', since that's what it actually made.
/:cl:


code/modules/crafting/craft.dm
```diff
@@ -166,12 +166,13 @@
...
-						user.put_in_hands(I)
-					if(send_feedback)
-						SSblackbox.record_feedback("tally", "object_crafted", 1, I.type)
+					for(var/I in R.result)
+						var/atom/movable/M = new I (get_turf(user))
+						M.CheckParts(parts)
+						if(isitem(M))
+							user.put_in_hands(M)
+						if(send_feedback)
+							SSblackbox.record_feedback("tally", "object_crafted", 1, M.type)
```
SabreML, line 175: Not sure if this should be changed to tally the recipe datum instead, since currently it'll list both the HUD and sunglasses as being crafted separately.

dearmochi (approved):

code/game/objects/items/weapons/pneumaticCannon.dm
```diff
@@ -137,7 +137,7 @@
 
 /datum/crafting_recipe/improvised_pneumatic_cannon //Pretty easy to obtain but
 	name = "Pneumatic Cannon"
-	result = /obj/item/pneumatic_cannon/ghetto
+	result = list(/obj/item/pneumatic_cannon/ghetto)
```
AffectedArc07, line 140: This changes a lot. Please also add a unit test (See `code/modules/unit_test`) to ensure that every subtype of `/datum/crafting_recipe` has result as a list. `islist()` will help here. 

The amount of unknown runtimes this could cause if someone isnt aware is worrying.

AffectedArc07 (changes requested):

code/game/objects/items/weapons/pneumaticCannon.dm
SabreML: Could this not just be made into an `ASSERT()` on `/datum/crafting_recipe`?
```dm
/datum/crafting_recipe/New()
	. = ..()
	ASSERT(islist(result))
```

AffectedArc07: These dont get instanced until they are actually being crafted, which doesnt solve the problem. A unit test fires in the CI stage before the PR is even merged, saving bad code from getting in.

AffectedArc07 (approved):

PR #15449, opened 2021-02-06
Ports Physiology

Fox-McCloud:
Ports TG's physiology datum.

Physiology datums are kinda like species x_mod modifiers, only they're attached to the mob, itself, rather than the species. This has advantages for if we ever want to make something persist through species changes and for it to be more permanent unless you're cloned; it can also hold its own armor datum, if need be.

This also adds support for a generic "armor" value for species, as well.

Lastly, this alters `apply_damage` to always hit; if an organ is missing, then it'll always target the chest. Additionally, if you want to call `apply_damage` to apply spread out damage, you can, instead of just targeting a particular organ.

Blood drunk and the die of fate are really the only things that utilize this new system.

:cl: Fox McCloud
tweak: added a generic physiology system
/:cl:


dearmochi (approved):

code/modules/mob/living/carbon/human/species/_species.dm
```diff
@@ -329,46 +331,66 @@
 /datum/species/proc/handle_death(gibbed, mob/living/carbon/human/H) //Handles any species-specific death events (such as dionaea nymph spawns).
 	return
 
-/datum/species/proc/apply_damage(damage = 0, damagetype = BRUTE, def_zone = null, blocked = 0, mob/living/carbon/human/H, sharp = 0, obj/used_weapon = null)
-	blocked = (100 - blocked) / 100
-	if(blocked <= 0)
-		return 0
+/datum/species/proc/apply_damage(damage = 0, damagetype = BRUTE, def_zone, blocked = 0, mob/living/carbon/human/H, sharp = FALSE, obj/used_weapon, spread_damage = FALSE)
+	var/hit_percent = (100 - (blocked + armor)) / 100
+	hit_percent = (hit_percent * (100 - H.physiology.damage_resistance)) / 100
```
SteelSlayer, line 336:
```suggestion
	hit_percent *= (100 - H.physiology.damage_resistance)) / 100
```

Kyep (approved):

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -596,6 +599,7 @@
 	else if(!(flags & SHOCK_NOGLOVES)) //This gets the siemens_coeff for all non tesla shocks
 		if(gloves)
 			siemens_coeff *= gloves.siemens_coefficient
+	siemens_coeff *= physiology.siemens_coeff
 	siemens_coeff *= dna.species.siemens_coeff
```
AffectedArc07, line 603: Is this still meant to multiple by both?

Fox-McCloud: one is based on species, the other is based on physiology.

code/modules/mob/living/carbon/human/species/golem.dm
```diff
@@ -9,12 +9,7 @@
 	inherent_biotypes = MOB_HUMANOID | MOB_MINERAL
 	dies_at_threshold = TRUE
 	speed_mod = 2
-	brute_mod = 0.45 //55% damage reduction
-	burn_mod = 0.45
-	tox_mod = 0.45
-	clone_mod = 0.45
-	brain_mod = 0.45
-	stamina_mod = 0.45
+	armor = 55
```
Fox-McCloud, line 12: All of these values are the same as before, despite appearing different--we handle these with a single `armor` value now.

AffectedArc07 (approved):
Seems fine to me

PR #15454, opened 2021-02-07
Turf Refactor, Part 1: Purging /turf/simulated/shuttle

Fox-McCloud:
Beginning of standardizing turfs so they're more sane. 

This purges `/turf/simulated/shuttle` because it doesn't really have any reason to exist---it also causes issues when they're smoothing with or adjacent to other turfs---for instance, you can't see wet tiles on shuttle turfs, and blood fails to render.

The end goal of this is to purge all instances of `/turf/simulated/shuttle` and `/obj/structure/shuttle/window`.

As a side note, this also standardizes shuttle airlocks to `/obj/machinery/door/airlock/titanium`.  `/obj/machinery/door/airlock/shuttle` was utilized, sometimes, other times `/obj/machinery/door/airlock/titanium` was used...or even an unpowered door that *looks* like a shuttle airlock. `/obj/machinery/door/airlock/shuttle` is completely removed from the code, because, well....it's literally an exact duplicate of `/obj/machinery/door/airlock/titanium`.

Lastly, this corrects ALL non CC z-level multi-turf stacking issues; let's never do that ever again.

This also has the effect of making everything constructable and deconstructable, as it should be----technically, you could already deconstruct shuttle, but it had a lot of copy-pasta code to do so.

This will also allow us to more easily find turfs with duplicate turfs on them; this is a no no--we should never have multiple turfs on the same tile; these eventually all need to be gotten rid of. They were super hacky work around to have diagonal walls and the appearance of a turf underneath them.

This does alter the look of shuttle, but it's for the better, IMO:

img
img
img

Qualms: but the corner turfs near the syndicate shuttle look awful!

Yes, they do; this isn't the walls fault; this is due to being adjacent to `/turf/unsimulated`---those will be dealt with, after this refactor, so that *will* look proper, in time, don't worry.

:cl: Fox McCloud
tweak: purges some turfs that shouldn't exist
/:cl:


AffectedArc07 (approved):

AffectedArc07:
This jumping the map queue was approved internally to avoid it becoming a conflict bomb, due to the mass edits being manual and not find+replace.

PR #15462, opened 2021-02-07
Nerfs botany potatoes

Qwertytoforty:
## What Does This PR Do

This reduces the effect of the electrical activity gene when combined with Capacitive Cell Production by a factor of 10.
This reduced the max power of a fully optimized plant cell from 400000 to 40000, the same as a bluespace cell.

## Why It's Good For The Game

From a balance and realism standpoint, a potato that can be mass produced should not store 10x the power of the best science nanotrasen (TM) has to offer going into a bluespace cell. Apart from the fact it makes no logical sense, it allows crew to have things that are already insanely strong, to be stronger.  Mechs, batons, borgs will effectively never run out of power with this cell in them, even if they are using the most power intensive modules. 

Sure, yellow cells exist, but they can be drained and force the item to recharge a bit, when the most power intensive actions are taken. This cell however, with its 400k capacity, is for all intents and purposes, infinite, or at least has enough power to constantly fire a borgs laser gun with self repair and ion thrusters for at least 10 minutes, while at least with a yellow cell eventually, that power runs low, and a break needs to be taken. 

## Changelog
:cl:
tweak: Plant cells power capacity when combined with electrical activity gene, is 10x weaker than before.
/:cl:

### DIFF
--- a/code/modules/hydroponics/plant_genes.dm
+++ b/code/modules/hydroponics/plant_genes.dm
@@ -224,7 +224,7 @@
 	// Cell recharging trait. Charges all mob's power cells to (potency*rate)% mark when eaten.
 	// Generates sparks on squash.
 	// Small (potency*rate*5) chance to shock squish or slip target for (potency*rate*5) damage.
-	// Multiplies max charge by (rate*1000) when used in potato power cells.
+	// Multiplies max charge by (rate*100) when used in potato power cells.
 	name = "Electrical Activity"
 	rate = 0.2
 	origin_tech = list("powerstorage" = 5)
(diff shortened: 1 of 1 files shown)


Carthusia:
I wonder...if you inject plasma into a 400k potato cell and use it in a stun baton...would it explode with the explosive firey power of 1 million george foreman grills?

Kyep (approved):

TheYeetster1:
Video of 400,000k battery in a borg with an RCD 

https://youtu.be/sktu8akI5Ss

cluwneythescourge:
Just a players perspective, but that argument about going 10 minutes doesn't hold water to me, as that cell does not recharge.  Considering how easy xeno actually is, and the fact that most seem to prefer the yellow cells already as borg etc (at least in my experience), I would say that this seems unnecessary to me.   Nerf it some, sure;  but make it literally the same as a regular cell, then why bother at all?

Mitchs98:
> Just a players perspective, but that argument about going 10 minutes doesn't hold water to me, as that cell does not recharge. Considering how easy xeno actually is, and the fact that most seem to prefer the yellow cells already as borg etc despite that big 400k number (at least in my experience), I would say that this seems unnecessary to me.
> 
> Nerf it some, sure; but make it literally the same as a regular cell, then why bother at all?

Bluespace cells require materials to make them.

Potato batteries are infinitely renewable.

That's why you bother. Potatoes are free and can be grown without power. Batteries cost resources, need RnD, and also need an active power supply.

Spacemanspark:
I've seen potato cells that can regenerate before, I'm pretty sure.

As sad as I might be to see this go... yeah, it's far too much. Infinitely charged borgs shouldn't be a thing.

PR #15464, opened 2021-02-07
Adds door remote and the autopsy scanner to protected cryo items

Qwertytoforty:
## What Does This PR Do

This PR adds the door remote and autopsy scanner to the protected items cryopods save in storage.

## Why It's Good For The Game

Door remotes and autopsy scanners are irreplaceable, so it is good to save them in cryodorms, where they (might) be recovered.

## Changelog
:cl:
tweak: Cryodorms now store door remotes and autopsy scanners to items for cryodorms to store.
/:cl:


Kepplerx:
Would it be possible to add Atmos holofans to that list as well?

Kyep (approved):

PR #15468, opened 2021-02-07
Removes the Assassin Holoparasite

SabreML:
## What Does This PR Do

Removes the Assassin holoparasite from the list of possible choices, for the Injector, Tarot cards, and Scarab eggs. (Still admin spawnable though).

## Why It's Good For The Game

The Assassin has two abilities which are tied together: Stealth mode, and a big assassination attack.
The attack does 50 damage and has a 16 second cooldown. This is compared to the Ranged holoparasite which can do that much and more every 1-2 seconds.

If you *do* manage to hit someone with the attack then you'll be forced to wait 4 seconds until you can recall into your host, while fully visible and with no damage reduction or armour.

The attack can only be performed from the 'Stealth mode', which **doesn't even make you entirely invisible** and requires you to be manifested to in order to activate. While stealthed you're only slightly harder to see than a Ranged holoparasite in scout mode.

The Assassin holoparasite is undeniably the worst choice when compared to the others, and currently only exists as noob bait.

## Changelog
:cl:
del: Removed the Assassin holoparasite from the list of possible choices.
/:cl:


Aracino:
I'm confused. Why are you removing it? You can also buff it to make it a valuable choice instead of outright removing it.

SabreML:
> I'm confused. Why are you removing it? You can also buff it to make it a valuable choice instead of outright removing it.

Primarily it's because I, personally, can't think of any good ideas on how to buff it. And until someone else does I think it would be better to just remove it from the list, considering it's taken pretty often by people who don't know better, and the above issues.

Aracino:
> > I'm confused. Why are you removing it? You can also buff it to make it a valuable choice instead of outright removing it.
> 
> Primarily it's because I, personally, can't think of any good ideas on how to buff it. And until someone else does I think it would be better to just remove it from the list, considering it's taken pretty often by people who don't know better, and the above issues.

Well I have a few ideas. Maybe give it actual invisibility in stealth mode, and the stealth mode is a toggleable mode like scout/ranged attack for ranged holoparas. Go stealth mode, you're invisible, you're able to sneakily leave and exit your host and be confusing to whoevers trying to find you if your host isn't going loud, with the added bonus damage from hitting out of stealth. The downside is the normal downside, where you won't be able to reenter your host for a while, and you can only restealth by returning to the body. 

Just an idea I have for how you could improve it to focus on the assassination element, usually when a holopara pops out it's very obvious, maybe make this one focus on being the "stealthy" holopara.

Coolrune206:
If we're discussing improving it instead of removing it, why not give it actual assassination potential? Remove it's whole invisibility and stealth attack ability and replace it with attacks that silence the target for a few seconds after attacking, and make it so it's punches are silent unlike other Holoparasites. This would make it unideal in open combat because it would still have less armor than other Holoparasites, but would make it excellent for, y'know, assassination

Dashsora:
I like coolrune's Idea of making it a more silent subtype.

And, to be fair, if we compare any holoparasite type to the ranged one, might as well remove them all. Maybe the ranged one is the one that should go.

SabreML:
I'll see about making an alternative PR with balance changes instead, but I'll defintely leave this one up since as it is now I really do think it warrants removal.

hal9000PR:
I'd rather see it weaken then silence, being silenced is rare because when combined with a stun , something that is fairly common, it is completely unfun for the target because they can do literally nothing.

SabreML:
I'll just mention here that I've somehow managed to completely brick my test server database, so I'm having trouble testing any code related stuff. E.g. The half finished balance PR.

Kyep (changes requested):
Instead of removing it, I would rather it was reworked to be a viable option.

SabreML:
> Instead of removing it, I would rather it was reworked to be a viable option.

I have tried rebalancing it but I really can't find the inspiration or motivation to do it properly. The best idea I had was to force it to recall after each hit, and to make the hit a lot stronger. But that has its own balance implications that I honestly can't find the effort to argue for.
Personally I was hoping it could just be disabled until someone else could think of an idea to fix it, like with Xenomorphs, Hierophant, Lavaland syndi base, etc.

PR #15476, opened 2021-02-08
Beach Unsimulated Removal

Fox-McCloud:
Since either @Darkmight9 or my own beach PRs could be held up for a while and bother are conflicted---I'd rather not have unsimulated turfs hanging around until that can be settled, especially since it's just beach and CC that have them left. It also means that @Darkmight9 has a basis of how to adjust his map for the new, simulated reality.

This doesn't change much of anything about the away mission; it's the same--it just has indestructible sandstone around the level as a "just in case" and some baseturf helpers.

The standstone in the map is actual sandstone, with the exception of the pirate base, to prevent cheese.

:cl: Fox McCloud
tweak: beach now uses simulated turfs
/:cl:


code/modules/awaymissions/mission_code/beach.dm
```diff
@@ -27,3 +27,154 @@
...
+/turf/simulated/floor/beach/away/water/deep/rock_wall
+	name = "Reef Stone"
+	icon_state = "desert7"
+	density = TRUE
+	opacity = TRUE
+	explosion_block = 2
+	mouse_opacity = MOUSE_OPACITY_ICON
+	baseturf = /turf/simulated/floor/beach/away/water/deep/rock_wall
+
+/obj/effect/baseturf_helper/beach/away/sand
```
Fox-McCloud, line 174: away missions and ruins need baseturf helpers because when you play a turf on an existing turf, the new turf inherits the old turf's baseturf---this is what we want in 99% of scenarios to avoid weird oddities, but it does have some unintended consequences for things like ruins and away missions.

In the case of away missions, the z-level is technically `/turf/space`, so any turfs moved to this inherit the space as the baseturf.

code/modules/awaymissions/mission_code/beach.dm
```diff
@@ -27,3 +27,154 @@
 	spawn(1)
 		W.loc = get_step(W, dir)
 	water_timer = addtimer(CALLBACK(src, .proc/drip), water_frequency, TIMER_STOPPABLE)
+
+/turf/simulated/floor/beach/away
```
Fox-McCloud, line 31: This is really gross--we need a unified beach tile set for everything, but this is about getting beach altered ASAP with as little changes as possible.

Kiyahitayika:
Sorry for breaking MapDiffBot ;__;

AffectedArc07 (approved):

PR #15485, opened 2021-02-09
Makes printed welding tools/fire extinguishers start empty

SabreML:
## What Does This PR Do

Makes Welding tools and Fire extinguishers that are printed from an autolathe start empty, rather than fully fueled.

## Why It's Good For The Game

Even with fancy bluespace future technology, I don't think it makes much sense for a lathe to be able to turn metal and glass into welding fuel and water.

## Changelog
:cl:
tweak: Welding tools and Fire extinguishers that are printed from an autolathe will now start empty, rather than fully fueled.
/:cl:

### DIFF
--- a/code/game/objects/items/tools/welder.dm
+++ b/code/game/objects/items/tools/welder.dm
@@ -25,6 +25,8 @@
 	var/maximum_fuel = 20
 	var/requires_fuel = TRUE //Set to FALSE if it doesn't need fuel, but serves equally well as a cost modifier
 	var/refills_over_time = FALSE //Do we regenerate fuel?
+	/// Does it start with no fuel?
+	var/empty = FALSE
 	var/activation_sound = 'sound/items/welderactivate.ogg'
 	var/deactivation_sound = 'sound/items/welderdeactivate.ogg'
 	var/light_intensity = 2
@@ -32,9 +34,10 @@
 	var/progress_flash_divisor = 10 //Length of time between each "eye flash"
 
 /obj/item/weldingtool/Initialize(mapload)
-	..()
+	. = ..()
 	create_reagents(maximum_fuel)
-	reagents.add_reagent("fuel", maximum_fuel)
+	if(!empty)
+		reagents.add_reagent("fuel", maximum_fuel)
 	update_icon()
 
 /obj/item/weldingtool/Destroy()
@@ -175,6 +178,8 @@
 	update_torch()
 	..()
 
+/obj/item/weldingtool/empty
+	empty = TRUE
 
 /obj/item/weldingtool/largetank
 	name = "industrial welding tool"
@@ -183,6 +188,9 @@
 	maximum_fuel = 40
 	materials = list(MAT_METAL=70, MAT_GLASS=60)
 	origin_tech = "engineering=2;plasmatech=2"
+
+/obj/item/weldingtool/largetank/empty
+	empty = TRUE
 
 /obj/item/weldingtool/largetank/cyborg
 	name = "integrated welding tool"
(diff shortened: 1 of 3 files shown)


MRgaty:
I dont see why this would be an improvement, kinda forces you to go around to find a welding fuel tank which is kinda tedious if you have something else to do.

SabreML:
> I dont see why this would be an improvement, kinda forces you to go around to find a welding fuel tank which is kinda tedious if you have something else to do.

Engineers and all other jobs that use a welding tool (as far as I know) can vend them, or have them in lockers. So it won't be a problem for them.

If it's a civilian or an IPC trying to get one though then granted it *will* be slightly more inconvenient, but considering the massive number of welding tanks around the station it would only take maybe fifteen seconds at the absolute most to fuel it. Same with the fire extinguishers too.

Personally I think that the consistency/realism benefits outweigh the downsides.

There's also the case for roundstart blobs, those rounds may last slightly longer than 15 minutes if people actually need fuel their welders before attacking. (Assuming that cargo doesn't do it for you all with a welding tank, anyway.)

DeveloperLoser:
> There's also the case for roundstart blobs, those rounds may last slightly longer than 20 minutes on average if people actually need to fuel their welders before attacking. **(Assuming that cargo doesn't do it for you all with a welding tank, anyway.)**

This especially, you will likely have people bringing welder tanks to the front lines (if not at cargo) which, if left undefended for a moment, will result in a massive amount of damage to the crew. Also makes attacking a blob a bit more like a siege, where a blob could easily mount a more meaningful sortie as the attackers would actually have points they'd have to defend.

Mitchs98:
I don't really see much of a reason to implement this, other than the realism argument. I really doubt it's going to have much of an impact at all on blob rounds...people will just drag the fuel tank directly to cargo.

procdrone:
yeah i dont see this helping with biohazard gamemodes, and only it being an unecessary annoyance.

if your aim is to assist biohazards, having a larger map would be a direct fix, its hard to stay hidden on Box with 100 players, and a % of them civilians roaming maints all the time. (+sec + antags)

Fox-McCloud:
Not really a fan of this, to be honest---it's just tedium for the sake of it; I get why, but this is pretty eh.

SabreML:
> if your aim is to assist biohazards, having a larger map would be a direct fix, its hard to stay hidden on Box with 100 players, and a % of them civilians roaming maints all the time. (+sec + antags)

This is definitely true but... maybe a little bit out of scope for this PR?
Sure it probably won't do much to help blobs out, but that would always be more of just a nice side effect anyway.

The primary goal of this really was the """"realism"""" point. It really shouldn't add any extra hassle to getting a welder from cargo, since I imagine it would become standard to get a welding tank for exactly that reason.

And again, all other welding tools *(Lockers, toolboxes, vendors, maintenance, RnD)* are unaffected by this. It's only the stuff that's being printed from metal and glass.

Kyep (changes requested):
Objection. I don't like this.
It just adds tedium for no benefit.

SabreML:
> Objection. I don't like this.
> It just adds tedium for no benefit.

The benefit is that tools won't contain fluids when they're printed from metal and glass, which I personally consider worth any possible added tedium. And for that point, there really shouldn't be any. With the exception of maybe ten seconds for a cargo tech to grab this welding tank.
image

Fox-McCloud:
While the concept of something starting empty, to balance it isn't the worst idea (ie: power cells might need this), for the case of extinguishers and welders, I don't feel this really fits the case. Both of these are fairly low impact without horrific amount of balance implications.

This PR has been rejected and will be closed.

PR #15489, opened 2021-02-09
Merges Grab and Pull

Fox-McCloud:
From TG.

WIP

This just makes sense, both from a coding standpoint, and a "easier for the player" standpoint. It does mean some feature changes, but, overall, it lays groundwork for an, overall, better system (fireman carry, slowing down while pulling someone who's prone, and the likes). 

Instead of greating an abstract, no drop grab object on the mob, and having loads of cross references, this, instead, uses pulling as the basis of grabbing---how it works:

- The first pull is a passive grab---this doesn't slow you down one bit and functions exactly like it does for pulling right now
- Grabbing/pulling again, after this, puts them in an aggressive grab---this has an associated slowdown with it
- Grabbing yet another time puts them in an neck grab; they'll be lying on the floor and unable to move, but they won't take damage
- Grabbing yet one last time puts them in a kill grab; they'll accumulate losebreath during this time and suffocate, in addition to being unable to move

You can still throw mobs and you can still eat them, so that is retained.

Another thing that is likely to not survive is using grab to put someone directly into a machine; this will be done with mousedrop (though I'd like to move in the direction of TG's open/close state machines eventually).

This is marked WIP as I have the base level system working, but I need to purge `/obj/item/grab` and replace all existing code with this new system--just wanted to get the PR up now.

I'm aware this has lots of styling issues; another reason I threw this up--please no comments on *styling* specifically as I'll address that, later, once the entire thing is done.

This has been tested for all the basics, but there's a few things I need to double check.

:cl: Fox McCloud
add: Grabbing and pulling have been merged into one
add: Grab someone either by using the pull shortcut or the traditional method
fix: Crossing space while pulling something will now result in said item being pulled with you and pulled by you on z-level change
/:cl:


S34NW:
Solid idea, will hotkeys such as DEL to stop pulling still function as normal?

Fox-McCloud:
>Solid idea, will hotkeys such as DEL to stop pulling still function as normal?

No, you'll stop pulling to stop grabbing now.

Butlerblock:
Will swapping places with the mob while both standing remove the grab? Also does this take into account the possible balance issues of someone being full speed while having a passive grab for stuff like being able to near instantly throw let’s say a corpse you were dragging away directly at sec? Also will people with the ability to instantly neck grab(i think that’s the specific one, correct me if i’m wrong) such as a carpfu or CQC user be unable to pull people at full speeds?

TDSSS:
> * Grabbing/pulling again, after this, puts them in an **aggressive grab**---this has an associated slowdown with it
> * Grabbing yet another time puts them in an **aggressive grab**; they'll be lying on the floor and unable to move, but they won't take damage

One of those should probably be called something else.

Fox-McCloud:
>One of those should probably be called something else.

fixed the wording

ppi13:
How does this change catching thrown objects?

Fox-McCloud:
>How does this change catching thrown objects?

It doesn't.

matttheficus:
My only concern would be for CQC combos. If people start insta-grabbing you, Sec WILL meta that as a "oh this person has CQC". Would it be possible to retain ctrl-click as a grab method in this case?

Spacemanspark:
Didn't we try to do this before and it was rejected?

Dafrek:
Is this a first step into changing our combat into what TG recently changed theirs too?

Wojz:
So pulling will now be just grab intent, not ctrl+click? Do I understand that properly? If so I like that, I find it weird that we have two different kinds of grabbing things.

Fox-McCloud:
starting to pull something (whatever shorcut you use) will grab something. You'll be able to aggressively grab a human mob in this manner too; going to neck or kill grab, however, will specifically require you to be in grab intent---I stress, for clarity, though; you just have to be in that *intent*; you don't have to actually click on the person with an empty hand---you can still use the shortcut--it's just grabs of a certain level require grab intent.

Aracino:
How does this work for simple mobs who can't actually "grab", and rely on pull? Many simple mobs don't even have a drop pull button, like terror spiders for example. How would this work for them? Will they keep the drop pull macro?

Kyep (approved):
Having both "grab" and "pull" which are different things is pointlessly confusing to new players, and results in much of the code that handles these things having to check both.

So yes, it would be good to merge them.

Fox-McCloud:
Will be returning.

PR #15501, opened 2021-02-11
Fixes species without hair losing their hairstyle to radiation

SabreML:
## What Does This PR Do

Adds a `NO_HAIR` flag to Abductors, Dionae, Drask, Golems, Grey, Kidan, Plasmamen, Skeletons, Skrell, and Wyrn.
This prevents them from losing their horns/antennas/whatever the hell Dionae have to radiation, diseases, and blob/headcrab zombification.

## Why It's Good For The Game

It really doesn't make much sense that radiation would make a Skrell's tendrils 'Fall out in clumps', or that a disease would cause a Kidan's horns to disappear.

## Changelog
:cl:
tweak: Species without hair (Skrell, Kidan, Diona, etc.) will no longer lose their hairstyle to radiation, diseases, or zombification.
/:cl:


SteelSlayer (approved):

code/modules/mob/living/carbon/human/species/unathi.dm
```diff
@@ -15,7 +15,7 @@
 	else, frequently even their own lives. They prefer warmer temperatures than most species and \
 	their native tongue is a heavy hissing laungage called Sinta'Unathi."
 
-	species_traits = list(LIPS)
+	species_traits = list(LIPS, NO_HAIR)
```
Fox-McCloud, line 18: They do, though.

SabreML: ...Huh, so they do.

Although, while apparently the horns and hair are separated, the hair does still contains stuff like frills and secondary horns, which I think should still be immune to falling out.
I'd say it's better to make Unathi immune to hairloss entirely than having their frills start falling out in clumps.

dearmochi (approved):

code/modules/mob/living/carbon/human/species/unathi.dm
```diff
@@ -14,7 +14,7 @@
...
```
AffectedArc07, line 17: Undo this

AffectedArc07 (changes requested):

code/modules/mob/living/carbon/human/species/unathi.dm
SabreML: I still believe it'd be better to keep this so that Unathi won't have their frills falling out, but with two maintainers against it I'll concede the point.

AffectedArc07 (approved):

PR #15502, opened 2021-02-11
SS init tags V2

AffectedArc07:
## What Does This PR Do
Puts tags on all stages of `log_startup_progress` so it makes more sense what is initializing what both ingame and in DD
image
image

It also logs the startup stage in the MC tab so it can be debugged if a specific stage has halted and people arent sure on what it is
image

## Why It's Good For The Game
Looks better in my opinion, also helps for readability of what is assigned to what

## Changelog
:cl: AffectedArc07
add: Subsystem inits are now tagged
/:cl:


code/controllers/master.dm
```diff
@@ -17,48 +17,61 @@ GLOBAL_REAL(Master, /datum/controller/master) = new
 /datum/controller/master
 	name = "Master"
 
-	// Are we processing (higher values increase the processing delay by n ticks)
-	var/processing = TRUE
```
AffectedArc07, line 21: This is what happens when people change TRUE/FALSE willy nilly without actually thinking.

SteelSlayer (approved):

PR #15506, opened 2021-02-12
Allows blob mobs to move via blob tiles in zero-g

warior4356:
## What Does This PR Do
Allows blob tiles to work for stopping zero g drift for blob mobs and allows them to push off blob tiles when "stuck" in zero-g.

## Why It's Good For The Game
All other mobs are stopped by these tiles, however similar to spiders and webs, because the blob mobs don't collide with the blob tiles they are not stopped in zero g. This means blob mobs can be "stuck" floating on blob or float straight through the blob. This allows blob mobs to move around on blob in zero-g, as well as be stopped by them if they are drifting into them, instead of floating straight through.

## Changelog
:cl:
fix: Fixes blob mobs not interacting with blob tiles for zero-g movement
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -34,12 +34,11 @@
 				H.color = "#000000"
 		adjustHealth(-maxHealth * 0.0125)
 
-/mob/living/simple_animal/hostile/blob/get_spacemove_backup()
-	. = ..()
-	// If we don't find any normal thing to use, attempt to use any nearby blob structure instead.
-	if(!.)
-		for(var/obj/structure/blob/B in range(1, get_turf(src)))
-			return B
+/mob/living/simple_animal/hostile/blob/Process_Spacemove(movement_dir = 0)
+	// Use any nearby blob structures to allow space moves.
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/blob/blobs/blob_mobs.dm
```diff
@@ -34,6 +34,12 @@
 				H.color = "#000000"
 		adjustHealth(-maxHealth * 0.0125)
 
+/mob/living/simple_animal/hostile/blob/get_spacemove_backup()
```
Fox-McCloud, line 37: This should be

```
/mob/living/simple_animal/hostile/blob/Process_Spacemove(movement_dir = 0)
	for(var/obj/structure/blob/B in range(1, src))
		return TRUE
	return ..()
	```

Comment (Fox-McCloud) [inline code/game/gamemodes/blob/blobs/blob_mobs.dm:37]: This should be

```
/mob/living/simple_animal/hostile/blob/Process_Spacemove(movement_dir = 0)
	for(var/obj/structure/blob/B in range(1, src))
		return TRUE
	return ..()
```

Fox-McCloud (changes requested):

Kyep (approved):

AffectedArc07 (approved):

PR #15508, opened 2021-02-12
Make's drilling the safe actually viable

hal9000PR:
## What Does This PR Do
decreases the time it takes to drill the safe from 300 seconds to 120 seconds

## Why It's Good For The Game
Currently the drill takes way too long to open the safe.  Standing still in a highly secure area for 5 minutes isn't a viable strategy, especially when you consider the constant noise. Sec will be onto you almost instantly and then you have to kill them because you cannot subdue all of security for 5 minutes while you wait for the safe to be cracked open, so either this ends with a lot of dead sec officers, or a poor antag who thought the drill would drill faster. This opens up antags more ways to complete their objectives making rounds more varied and interesting.

## Changelog
:cl:
tweak: it now takes 120 seconds to drill a safe or 60 seconds with a diamond drill. 
/:cl:


FreeStylaLT:
POV: You've got 4 minutes and 45 seconds left and the AI has already screamed about your presence in the Vault
image

ppi13:
This is very reasonable, especially considering how loud it is.

warior4356:
I'm going to point out that the nuke ops drill is half of that listed time. The base time to drill is for the unupgraded drill that can be made without diamonds. It's also the drill normal traitors get. It being worse is deliberate to incentivise normal traitors being stealthy with a stethoscope.

TLDR: If you want the nukie drill to be 120 seconds, you should make it 240. Currently, the nukie drill takes 150 seconds.

hal9000PR:
> I'm going to point out that the nuke ops drill is half of that listed time. The base time to drill is for the unupgraded drill that can be made without diamonds. It's also the drill normal traitors get. It being worse is deliberate to incentivise normal traitors being stealthy with a stethoscope.
> 
> TLDR: If you want the nukie drill to be 120 seconds, you should make it 240. Currently, the nukie drill takes 150 seconds.
> 
> Further, the time to drill should ALWAYS be longer than a competent safecracker will take, to incentivize the skillful approach. With a good safecracker, you can do it in 60-120 seconds, so in my mind, the drill shouldn't be faster than that.

1) i have already considered this and nukies standing still for 60 seconds is fairly reasonable to get the NAD

2) I disagree, stethoscope is free, the drill isnt, yes you can make them but you need an exosuit fab. Something that costs TC should be faster than something that is free. The insensitive for the stealthy approach isnt that it is faster, that makes no sense. The insensitive for the stealthy approach is exactly that... It is stealthy. You use the drill and all of sec is on you and you have to spend 3tc, but in exchange for that downside it is easier and faster.

Kyep (approved):

AffectedArc07 (approved):
Payday2 made me develop PTSD over drills and I am not having that reflect into SS13

PR #15510, opened 2021-02-12
Add 20 plates to the dinnerware vendor in the kitchen

PidgeyThePirate:
## What Does This PR Do
Adds 20 plates to the dinnerware vendor so they can be used for Unique dishes without having to eat a ton of fries just to get the plates. Version 2 includes incredible features like not including commits from 2 years ago.

## Why It's Good For The Game
Plates are used to serve up and combine dishes, but are an absolute pain to get when you need a lot, since they rely on people having already eaten specific dishes and not thrown out the plates. Having some available from the outset should take some of the pressure off if the chef's style is heavy on this sort of thing.

## Changelog
:cl:
tweak: Added 20 plates to the dinnerware vendor in the kitchen.
/:cl:


farie82 (approved):

Carthusia:
Maybe make it 40-60 plates?

It sounds like alot, but when you get a good kitchen going, and want to make any kind of variety, those 20 are going to be gone real fast since unlike food cookers, you don't get extra produced as the station upgrades itself.

20 would only get you enough for 5 of 4 dishes

PidgeyThePirate:
I can make it more than 20, if a maint agrees that that's fine. I'm keeping it conservative for the moment though.

Kyep (approved):

Carthusia:
20 would be fine if you could make more at an autolathe, but currently its not possible (As far as i can tell)

AffectedArc07 (approved):

PR #15511, opened 2021-02-12
Adds Engine Access for Atmos Techs, but not Engi Access

Kepplerx:
## What Does This PR Do
-Adds Engine Access to Atmos Techs

## Why It's Good For The Game
-It gives them an access to an engine that is very Atmos based
-It allows Techs to set up Engine on low pop rounds
-It gives Techs more to do
-This is a access that is very often given to them in one form or another, be it by changing the access on doors, removing doors, or similar actions.
-Instead of mapediting to make the doors to Engine accept Atmos as well, I think this solution is better, as the Engine doors should only accept the Engine access.

## Changelog
:cl:
tweak: Tweaked the access for the Atmos job
/:cl:


Gatchapod:
`ACCESS_ENGINE_EQUIP` will also grant atmospherics technicians absolutely full engineering access, including engineering hardsuits, APCs and all lockers.
I understand radiation collectors and emitters use those, but I think a healthier approach would be adding `ACCESS_ATMOSPHERICS` to those two machines instead of making atmospheric technicians engineers+.

Fox-McCloud:
Not in support of this.

Atmospherics technician is *not* meant to be engineer++, nor should engineering be amalgamated, like science.

This should largely be the choice of the CE (or failing him being around, the AI or some other command figure). On full shifts, it might be unnecessary, but it should still be the CE's choice if he wants to allow this.

Setting emergency access on two doors isn't that much of an issue.

Kepplerx:
> `ACCESS_ENGINE_EQUIP` will also grant atmospherics technicians absolutely full engineering access, including engineering hardsuits, APCs and all engineering lockers.
> I understand radiation collectors and emitters use that access, but I think a healthier approach would be adding `ACCESS_ATMOSPHERICS` to those two machines instead of making atmospheric technicians engineers+.

On this I removed the ENGINE_EQUIP access, so as to not make Atmos Techs Engies+, a remap is also something that could work, but would mean that Atmos Access would be expanded to engine, even as a own "Engine Access" exists, which is not necessarily what one wants.

Kepplerx:
> Not in support of this.
> 
> Atmospherics technician is _not_ meant to be engineer++, nor should engineering be amalgamated, like science.
> 
> This should largely be the choice of the CE (or failing him being around, the AI or some other command figure). On full shifts, it might be unnecessary, but it should still be the CE's choice if he wants to allow this.
> 
> Setting emergency access on two doors isn't that much of an issue.

I removed the Engi access, so it is now only Engine Access. I will also point out that all CEs I have seen grant Atmos access, so this is being done in practice already. Furthermore the Engine is a very Atmos driven engine, so it would make sense that Atmos Techs would not only be familiar with it, but in some cases even more suitable than engis to use it. Atmos techs as they are now are far from Engis+, and more a Engis-, with some antag hiding spots.

Fox-McCloud:
>I will also point out that all CEs I have seen grant Atmos access, so this is being done in practice already.

That's exactly fine then--they're making the choice and decision themselves, as they should be.

This removes the choice, and edges atmos tech towards the "engineer+" category of things, which is exactly what I don't think is healthy.

Kyep (approved):

Wojz:
Currently the average atmos tech round consists of checking your notes to see your objectives, and if you have none, setting up the filters, maybe piping CO2 to the SM, setting the turbine, and... becoming an engineer without APC access. All of this takes 20 minutes if you are not, like, very new. This at least gives them something to do beyond the first 20 minutes of the game. 

Besides, they're supposed to be firefighters, and the SM is the only thing on station that can sustain a fire without petering out before anyone gets to it. Extinguishers are in fact used to delay delaminations.

code/game/jobs/job/engineering.dm
```diff
@@ -13,7 +13,7 @@
 			            ACCESS_TELEPORTER, ACCESS_EXTERNAL_AIRLOCKS, ACCESS_ATMOSPHERICS, ACCESS_EMERGENCY_STORAGE, ACCESS_EVA,
 			            ACCESS_HEADS, ACCESS_CONSTRUCTION, ACCESS_SEC_DOORS,
 			            ACCESS_CE, ACCESS_RC_ANNOUNCE, ACCESS_KEYCARD_AUTH, ACCESS_TCOMSAT, ACCESS_MINISAT, ACCESS_MECHANIC, ACCESS_MINERAL_STOREROOM)
-	minimal_access = list(ACCESS_ENGINE, ACCESS_ENGINE_EQUIP, ACCESS_TECH_STORAGE, ACCESS_MAINT_TUNNELS,
```
dearmochi, line 16: Why the removal? This is for the CE.

code/game/jobs/job/engineering.dm
```diff
@@ -94,7 +94,7 @@
 	department_head = list("Chief Engineer")
 	selection_color = "#fff5cc"
 	access = list(ACCESS_EVA, ACCESS_ENGINE, ACCESS_ENGINE_EQUIP, ACCESS_TECH_STORAGE, ACCESS_MAINT_TUNNELS, ACCESS_EXTERNAL_AIRLOCKS, ACCESS_CONSTRUCTION, ACCESS_ATMOSPHERICS, ACCESS_MINERAL_STOREROOM)
-	minimal_access = list(ACCESS_EVA, ACCESS_ATMOSPHERICS, ACCESS_MAINT_TUNNELS, ACCESS_EXTERNAL_AIRLOCKS, ACCESS_EMERGENCY_STORAGE, ACCESS_CONSTRUCTION, ACCESS_MINERAL_STOREROOM, ACCESS_TECH_STORAGE)
+	minimal_access = list(ACCESS_EVA, ACCESS_ENGINE, ACCESS_ENGINE_EQUIP, ACCESS_ATMOSPHERICS, ACCESS_MAINT_TUNNELS, ACCESS_EXTERNAL_AIRLOCKS, ACCESS_EMERGENCY_STORAGE, ACCESS_CONSTRUCTION, ACCESS_MINERAL_STOREROOM, ACCESS_TECH_STORAGE)
```
dearmochi, line 97: These additions give atmos techs access to Solars, Electrical Maintenance, Gravity Generator(!) and... the room with the Maintenance Drone machine.

Kepplerx: > These additions also give atmos techs access to Solars, Electrical Maintenance, Gravity Generator(!) and... the room with the Maintenance Drone machine, in addition to all of Engineering.

Hmm, gonna try a map change instead, as was proposed, or do a map change and something with this code, solars and elec maints with the grav gen should not all be on the same access

Kepplerx:
Going to replace this with a remap instead, due to how the accesses are used, with engine access giving grav gen access as well. That would be the simpler solution, alternatively I will try a more complex one, this PR does however have glaring issues as pointed out by Mochi. It will return in another form.

Issue #15504, opened 2021-02-11
You try to shake a dead person who has (near) no damage awake


farie82:
**Issue Description**:
You shake a dead non wounded person instead of it saying that they are dead like with all other wounded dead persons.
![image](https://user-images.githubusercontent.com/15887760/107702596-c46b1100-6cba-11eb-8e47-b566424b733a.png)

**What did you expect to happen**: 
It to tell me they are dead. So it will try to do CPR on the person and fail since they are dead

**What happened instead**:
My char tried to help them up on their feet again.

**Why is this bad/What are the consequences**:
Mostly inconsistency with the current working.

**Steps to reproduce the problem**:
1. advanced proc call "death" on a mob. No arguments
2. Click on them on help mode

**When did the problem start happening**:
Prob since new crit

**Extra information**:
Easy first issue

SabreML:
Fixed by the above PR.

PR #15512, opened 2021-02-12
Shaking the dead now gives a message indicating the person is dead

Kepplerx:
## What Does This PR Do
It fixes (https://github.com/ParadiseSS13/Paradise/issues/15504) by replacing the message for shaking someone with a message for shaking the dead (That being the dead who had not taken damage beyond crit).

## Why It's Good For The Game
Shaking the dead now no longer results in a weird message, if they were dead but not damaged beyond crit.

## Changelog
:cl:
fix: Fixed a issue with response messages for a target being dead
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -213,70 +213,77 @@
 		swap_hand()
 
 /mob/living/carbon/proc/help_shake_act(mob/living/carbon/M)
-	if(health >= HEALTH_THRESHOLD_CRIT)
-		if(src == M && ishuman(src))
-			check_self_for_injuries()
-		else
-			if(player_logged)
-				M.visible_message("<span class='notice'>[M] shakes [src], but [p_they()] [p_do()] not respond. Probably suffering from SSD.", \
-				"<span class='notice'>You shake [src], but [p_theyre()] unresponsive. Probably suffering from SSD.</span>")
-			if(lying) // /vg/: For hugs. This is how update_icon figgers it out, anyway.  - N3X15
-				add_attack_logs(M, src, "Shaked", ATKLOG_ALL)
-				if(ishuman(src))
-					var/mob/living/carbon/human/H = src
-					if(H.w_uniform)
-						H.w_uniform.add_fingerprint(M)
-				AdjustSleeping(-5)
-				if(sleeping == 0)
-					StopResting()
-				AdjustParalysis(-3)
-				AdjustStunned(-3)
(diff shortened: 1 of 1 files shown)


Kepplerx:
This is my second ever coding work, and is somewhat more difficult than the previous one I did. As far as I understand this should work, as I added a check for if dead, then moved the rest of the help_shake under the new check. That said any tips or such are appreciated, this may after all be bad coding work.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -213,70 +213,77 @@
...
-						active_hand.receive_damage(0, 5)
-						self_message = "<span class='danger'>You burn your hand trying to extinguish [src]!</span>"
-						H.update_icons()
-
-				M.visible_message("<span class='warning'>[M] tries to extinguish [src]!</span>", self_message)
-				playsound(get_turf(src), 'sound/weapons/thudswoosh.ogg', 50, 1, -1)
-				adjust_fire_stacks(-0.5)
-
-			// BEGIN HUGCODE - N3X
+	if(stat == DEAD)
```
farie82, line 216: This currently does not check if `M` is the same as `src`. AKA if the dead guy is trying to shake himself up.
Currently you'll get a message saying X tries to shake himself up

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -213,70 +213,77 @@
...
+		if(!player_logged)
+			M.visible_message( \
+			"<span class='notice'>[M] desperately shakes [src] trying to wake [p_them()] up, but sadly there is no reaction!</span>",\
+			"<span class='notice'>You shake [src] trying to wake [p_them()] sadly they appear to be too far gone!</span>",\
+			)
+	else
```
farie82, line 222: Early returns are preferred over `else`'s if you can help it.
Example of one:
```dm
/proc/test(A, B)
 if(A != 420)
  to_chat(usr, "invalid A")
  return // This is an early return.
 to_chat(usr, "valid params")
 return B * A
```
And then how it looks with an else:
```dm
/proc/test(A, B)
 if(A != 420)
  to_chat(usr, "invalid A")
 else
  to_chat(usr, "valid params") // non early return
  return B * A
```

farie82 (changes requested):

Kyep (approved):

AffectedArc07 (changes requested):
Please look into Farie's changes

AffectedArc07 (approved):

PR #15513, opened 2021-02-12
Rapid Cyborg Disabler Upgrade now requires way higher Tech levels

Kepplerx:
## What Does This PR Do
It changes the Cyborg disabler to require Engineering 7, Power 7 and Combat 7.

## Why It's Good For The Game
There have been many complaints of borgs as of lately, and the disabler being OP has been one of them. This PR would delay the potential to build them, and some rounds will not reach them at all. To get it with this PR one would need prototypes form Cargo as well as either a Slime core from Xeno or glowcaps from Botany, thus one would in certain rounds never reach this at all. The fact that one could make these just with Techs at Engi 4, Power 4 and Combat 4 will no longer be the case. (These are reachable in about 10 minutes in game time at the earliest)

## Changelog
:cl:
tweak: Tweaked tech requirements for Cyborg Disablers
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -104,7 +104,7 @@
 	name = "cyborg rapid disabler cooling module"
 	desc = "Used to cool a mounted disabler, increasing the potential current in it and thus its recharge rate."
 	icon_state = "cyborg_upgrade3"
-	origin_tech = "engineering=4;powerstorage=4;combat=4"
+	origin_tech = "engineering=7;powerstorage=7;combat=7"
 	require_module = TRUE
 	module_type = /obj/item/robot_module/security
 


Gatchapod:
This is a significant balance change.
For reference:

- engineering 7 requires a machine prototype from cargo
- power 7 requires glowcaps from hydroponics
- combat 7 is relatively easy to achieve using robotics

You won't be just delaying those upgrades. You'll be potentially denying them altogether, since hydroponics does not always provide glowcaps.
Not that I disapprove, mind you.

Kepplerx:
> This is a significant balance change.
> For reference:
> 
> * engineering 7 requires a machine prototype from cargo
> * power 7 requires glowcaps from hydroponics
> * combat 7 is relatively easy to achieve using robotics
> 
> You won't be just delaying those upgrades. You'll be potentially denying them altogether, since hydroponics does not always provide glowcaps.
> Not that I disapprove, mind you.

Good point, updated it to add that, so its stronger and more obvious than what I had written there before

S34NW:
Three different level 7's seems a bit harsh, maybe one or two would be a bit more balanced?
This would help with the secborg issue, but I don't think power 7 is the right call personally.

Kepplerx:
Did a major mistake, as this would change the wrong variable will be back with the right change. This would be a buff, as Gatchapod pointed out.

PR #15515, opened 2021-02-12
Rapid Cyborg Disabler Upgrade now requires way higher Tech levels V2

Kepplerx:
(Uploaded again as the last version edited the wrong file, and was as such a weird buff of a kind https://github.com/ParadiseSS13/Paradise/pull/15513)

## What Does This PR Do
It changes the Cyborg disabler to require Engineering 7, Power 7 and Combat 7.

## Why It's Good For The Game
There have been many complaints of borgs as of lately, and the disabler being OP has been one of them. This PR would delay the potential to build them, and some rounds will not reach them at all. To get it with this PR one would need prototypes form Cargo as well as either a Slime core from Xeno or glowcaps from Botany, thus one would in certain rounds never reach this at all. The fact that one could make these just with Techs at Engi 4, Power 4 and Combat 4 will no longer be the case. (These are reachable in about 10 minutes in game time at the earliest)

## Changelog
:cl:
tweak: Tweaked tech requirements for Cyborg Disablers
/:cl:

### DIFF
--- a/code/modules/research/designs/mechfabricator_designs.dm
+++ b/code/modules/research/designs/mechfabricator_designs.dm
@@ -1056,7 +1056,7 @@
 	id = "borg_upgrade_disablercooler"
 	build_type = MECHFAB
 	build_path = /obj/item/borg/upgrade/disablercooler
-	req_tech = list("combat" = 5, "powerstorage" = 4, "engineering" = 4)
+	req_tech = list("combat" = 7, "powerstorage" = 7, "engineering" = 7)
 	materials = list(MAT_METAL=80000 , MAT_GLASS=6000 , MAT_GOLD= 2000, MAT_DIAMOND = 500)
 	construction_time = 120
 	category = list("Cyborg Upgrade Modules")


MRgaty:
don't think borgs need to be nerfed to the ground, they are supposed to be stronger than a normal sec officer.

Gatchapod:
> 
> 
> they are supposed to be stronger than a normal sec officer.

[citation needed]

hal9000PR:
> don't think borgs need to be nerfed to the ground, they are supposed to be stronger than a normal sec officer.

are they?

Fox-McCloud:
This fundamentally misunderstand the position of those who are critical of secborgs. While yes, this is does lessen the late game criticisms of secborgs to a *very small degree*, this does nothing to address the average, day to day, complaints about secborgs.

It'll also conflict with techwebs and this is just going to wind up changed in the next month or so, anyway.

Kepplerx:
> This fundamentally misunderstand the position of those who are critical of secborgs. While yes, this is does lessen the late game criticisms of secborgs to a _very small degree_, this does nothing to address the average, day to day, complaints about secborgs.
> 
> It'll also conflict with techwebs and this is just going to wind up changed in the next month or so, anyway.

First of all this is not a one solution and erveryone loves borgs, I am aware there are other valid criticisms of borgs, that however does not mean this is not without effect. I am also aware the effect will be little, still, it is a step, if not as big a step as some want.  Second of all it will be rendered meaningless by techwebs, but only once those get implemented, which I myself want as fast as possible. The fact that changes are coming at some point is however not an argument to not to try and make it better now.

Kepplerx:
> don't think borgs need to be nerfed to the ground, they are supposed to be stronger than a normal sec officer.

While I agree that Secborgs should be in some ways better than Seccies, I am also aware that there are valid criticisms of them, which I do take into account. This is one thing I myself have used and found very OP, as it is in effect a spray and pray disabler, especially when combined with a larger battery. It is not removed either, just requires higher tech levels, which I consider perfectly fair considering the power of this upgrade.

MRgaty:
> > they are supposed to be stronger than a normal sec officer.
> 
> [citation needed]

Dont got any written quote, perhaps it was in 2018's survey but I do remember kyet stating that they are supposed to be the heavy artillery

Spacemanspark:
I distinctly recall Neca saying that borgs are meant to be good, if not better at their chosen module's department than their human counterparts. The downside is that they can't do jack shit outside of that.

Kepplerx:
> I distinctly recall Neca saying that borgs are meant to be good, if not better at their chosen module's department than their human counterparts. The downside is that they can't do jack shit outside of that.

While I agree Secborgs should be stronger Seccies, there are also valid complaints of them, as well as some very loud voices to "DELETE THEM". I do not wish that to happen, so this is an attempt to move towards a compromize. I know it isnt a big change, I know it isnt going to last long (with the techwebs), but it is an attempt to allivate some concerns I have heard about Secborgs.

Spacemanspark:
Oh I don't mind the change you're making here @Kepplerx
Personally I'd be fine with removing the upgrade altogether. My post was purely in response to Gatcha and charliminator.

Kyep (approved):

AffectedArc07 (approved):
Giving this the goahead since we need a remap done before techwebs go in.

PR #15518, opened 2021-02-13
Improves the usability of the thermal drill.

warior4356:
## What Does This PR Do
This PR exists as an alternative to #15508, as I feel the approach is a bit too extreme. That said I do feel like there are some legitimate issues with the drill that can be improved.

Time to drill for the default drill shortened to 4 minutes.
Diamond drill shortened to 2 minutes.
Regular traitors now have access to the diamond drill at 1 TC.

Adds a diamond drill and a stethoscope to the nukie ship.
image

While I was at it, I added a stethoscope to the DJ outpost to make explorer lives a bit easier. The first-aid kit on the table on top of it is the terrible one with ointment and gauze.
image

## Why It's Good For The Game
Nerfing the drill down to being a minute for nukies and two minutes for regular traitors removes any incentive for nuclear operatives to take the skillful approach. With a drill time of two minutes, a good safecracker will still be faster than drilling. One of the common complaints was that the nukies would have to spend TC on this, even if it was only 1 TC, so hopefully adding it to their ship will help with that issue.

As for regular traitors, I think the feedback that there is no incentive to use it given the noise and time is a fair one, further as they get less TC than nukies, even though I'd prefer to incentivize the skillful and stealthy approach giving them the faster drill and reducing the cost to 1 TC should make it far more viable to use.

Finally, the DJ thing was mostly thrown in because I'm already going to be in the mapping queue and it's related to the issue of opening safes. Few people even know that safe is there, and even if they find it, it's not likely that they will have the tools to get it open. This way at least there is a chance they can get it open.

## Changelog
:cl:
add: Added diamond drill to Nukie ship.
add: Added stethoscope to Nukie ship.
add: Added stethoscope to DJ outpost.
add: Gives regular traitors diamond drill for 1 TC.
delete: Removes regular drill from the regular traitor uplink.
tweak: Reduced the thermal drill to 4 minutes, diamond drill to 2 minutes.
/:cl:


hal9000PR:
I really don't understand why nukies who plan forward for the inevitability of the NAD being in the safe and save one tc shouldn't be rewarded, also if you make the thermal drill unpurchasable, you might as well remove it because nobody is going to make it because if they can make a thermal one they can make a diamond one. So this still makes the thermal drill a useless item.

hal9000PR:
Also, you say my pr is too extreme, however this buffs the drill more. It now costs 3x less for the most common gamemode while retaining the same drill speed. You also make nukies more forgiving (which I'm not opposed to) by giving them a drill round start. But yet these are all buffs that I didn't do in my PR yet you consider mine more "extreme". Nukies are never going to use the stethoscope. They are either war ops and then they arent gonna give up the opportunity for payday memes, or they are stealth ops and the NAD isnt in the safe. Personally all i see this as is a bigger buff to traitors and the change to nukies isnt gonna matter much, if a nukie can hold still for 60 seconds they can do it for 120 seconds aswell so this is just delaying the round. I feel like nukies should be rewarded for not being incompetent, like the trained mercenaries they are, preparing a tc incase the NAD is the safe should be rewarding.

warior4356:
@hal9000PR To address your feedback.

My main complaint with your PR is making it so nukies can drill in 60 seconds. I think that's too fast and there should be at least an incentive to play the minigame under fire instead of the drill always being a better option.

I think making it so regular traitors can drill in 120 seconds is a good idea though, which is why I changed which drill they got. Further you're right to point out that it is a loud and obvious item, so it probably shouldn't cost as much as it does. As for why the regular drill exists? Mostly just if you don't have access to diamonds/tech for crafting it. You could say the same thing about the laser scalpel when the IMS exists.

For the point of why would nukies ever buy a drill if they get a free one? Well, the drill has a large downside. It's a gigantic item. You can't wear it. Or put it in anything. With fox's drag changes coming soon, it will cost you a hand to bring one with you, so it's a bad "just in case" item. A nukie has to give up a hand that can be holding a gun to bring it. The spending of 1 TC lets you bring one in when you need it. 

For heads and maints, some options for tweaks I'd be happy to make for this PR, if you dislike parts of it:
Give regular traitors the regular drill again at the same price (3) or cheaper.
Change the drill on the nukie ship to be the worse one, to make spending TC more appealing.
Adjusting drill time on one or both drills.

hal9000PR:
1) maybe the drill cost should be decreased i agree with that.
2) nukies arent ever gonna use a stethoscope as i said. They are either war ops, so they are gonna save 1tc for a drill if the NAD is the the safe. Or they are stealth ops and then the NAD is 99.99% of the time on the captain.
3) fox's drag changes only effects pulling mobs if i am not mistaken.
4) If the crew cannot stop the nukies in 60 seconds then they cannot stop them in 120. Making it take longer than 60 seconds is just delaying the round.
5) you might as well remove the thermal drill in this PR because it will never see use, set the diamond drills multiplier to 1 then change the drill time to 120 seconds.

Kyep (changes requested):
Objection.
You're giving the nukies a diamond drill AND stethoscope for free.
In addition to making drills better.
Nukies don't need any buffs with regards to their ability to drill safes. They're already able to purchase a good drill for 1 TC.

Even traitors probably don't need drill buffs. The steal the docs objective is not that hard compared to say... kill objectives.

warior4356:
Objection noted. I may try a different PR at a later date, would you be alright with only providing them with a drill? Given the Item's weight class, the "free" drill is more of a hindrance than a help.

PR #15520, opened 2021-02-13
Tweaks to PDAPainter for quicker and more fluid paint.

CkeyLurapa:
## What Does This PR Do
The PR is a minor tweak that has the eject function called when a pda is painted. So the HoP does not have to click (insert), click (paint), click (eject). It cuts out one of the steps, making it a more fluid experience. 

## Why It's Good For The Game
Makes the painting of PDAs more time-efficient.

## Images of changes
Uge77ncSpg

## Changelog
:cl:
tweak: made PDA painting more efficient.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/guardian/types/bomb.dm
+++ b/code/game/gamemodes/miniantags/guardian/types/bomb.dm
@@ -25,6 +25,7 @@
 	if(istype(A, /obj/))
 		if(bomb_cooldown <= world.time && !stat)
 			var/obj/item/guardian_bomb/B = new /obj/item/guardian_bomb(get_turf(A))
+			add_attack_logs(src, A, "booby trapped (summoner: [summoner])")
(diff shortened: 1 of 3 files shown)


Fox-McCloud:
It's better---though, to be honest, I think having a menu where you can see the specific PDA icons, ahead of time, would be the best solution.

CkeyLurapa:
> It's better---though, to be honest, I think having a menu where you can see the specific PDA icons, ahead of time, would be the best solution.

Yeah. I have been looking into making a PDA Painter similar to a floor-painter, a belt-item. Or have a more up to date UI with the PDA painter that:

1. Has the departments,
2. Scales the departments in their chain of command
3. 2 buttons, paint and eject.

But that's in the future/wip. And thought this a better short-term solution. As I find the PDA Painting to be a bit slow.

PR #15526, opened 2021-02-14
Fixes thermite removing lubed/wet floor overlays

SteelSlayer:
## What Does This PR Do
Currently, you can put thermite into a spray bottle and combine it with space lube/water, and when you spray, the overlay does not appear on floors, because the thermite code removes it. This can also happen if you just splash the beaker onto the floor. This PR fixes that.

## Why It's Good For The Game
Fixes an exploit.

## Changelog
:cl:
fix: Fixes thermite removing lubed/wet floor overlays when used in a spray bottle or splashed onto turfs
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/pyrotechnic.dm
+++ b/code/modules/reagents/chemistry/reagents/pyrotechnic.dm
@@ -203,8 +203,6 @@
 			S.create_reagents(volume)
 		S.reagents.add_reagent("thermite", volume)
 		S.thermite = TRUE
-		S.overlays.Cut()
-		S.overlays = image('icons/effects/effects.dmi', icon_state = "thermite")
 		if(S.active_hotspot)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/reagents/pyrotechnic.dm
```diff
@@ -203,8 +203,6 @@
 			S.create_reagents(volume)
 		S.reagents.add_reagent("thermite", volume)
 		S.thermite = TRUE
-		S.overlays.Cut()
-		S.overlays = image('icons/effects/effects.dmi', icon_state = "thermite")
```
SteelSlayer, line 207: The first line here is the troublemaker. All it really does currently is just remove wall smoothing if you pour thermite on a wall, and does nothing on a floor.

The second line here doesn't even work, and even if it did, it wouldn't matter because thermiting a wall/floor burns and deletes it completely.

AffectedArc07 (approved):

PR #15527, opened 2021-02-14
Purges Unsimulated Turfs: Touches Up CC Areas

Fox-McCloud:
Converts CC over to using only simulated turfs----this purges them from the game

This also makes a few aesthetic changes to number areas and makes balance changes to the the Syndicate Infiltrator

First, the pics:

Adds some figurines and fixtures to the syndicate planet, inspired by TG.

img
img

Syndicate base has a new look that isn't just riveted walls--feels a bit more "IC"; also gives the surrounding snow a mountainside edge turf, so it feels like you're more on a planet, somewhere, rather than just some weird box in space.

img

Gives the wizard den a new aesthetic look, partially inspired by TG (we should probably just implement their version, at some point)

img

Dramatically increases the amount of turrets the Syndicate Infiltrator has, to compensate for the fact their walls are destructable now

img

Now, onto the code stuff.

This PR bans the usage of two things: unsimulated turfs, with the sole exceptions being `/turf/space` and its children and `/turf/noop` (as this is for templates). Likewise, the usage of turf stacking will also be forbidden; if you want a certain aesthetic look, you're going to have to rely on wall smoothing or something similar for it; putting two turfs on the same tile is no longer allowed.

Incidentally, this should resolve some atmos oddoties; some simulated turfs would misbehave when they were adjacent to both an unsimulated wall and space---it didn't know how to handle it and would wind up permanently activating itself and never sleeping again.

:cl: Fox McCloud
tweak: Gives the wizard den a new aesthetic
tweak: gives the syndicate base a new aesthetic along with the planet it's on
add: Dramatically increases the amount of turrets the Syndicate Infiltrator has
fix: Fixes some odd atmos issues related to unsimulated turfs interacting with simulated ones
fix: Fixes the desert holo having plating on it
del: Purges all unsimualted turfs. Dont' ever use them ever again
/:cl:


TDSSS:
While we are at it, maybe the lower corners of the shuttle should have turrets too, they seem like a bit of a blindspot atm.

AffectedArc07:
Yeah I agree turrets should be here
image

other than that, looks good

Kiyahitayika:
YAS

Fox-McCloud:
Added 2 more turrets.

Kyep (approved):

AffectedArc07 (approved):

PR #15528, opened 2021-02-14
Xray Gene removed from Genetics

Kepplerx:
## What Does This PR Do
Changes the chance for the Xray Gene to activate to 0.

## Why It's Good For The Game
Genetics Xray comes with too big an advantage, and can literally destroy rounds, especially if given to Sec, or even given out at all. This is something nukies need to pay a lot of TC for, and Genetics can get it for free, this further shows how broken this Gene is. This does not affect the Xray implant, merely the gene, which only really is used in Genetics.

## Changelog
:cl:
tweak: Changed the chance of Xray to activate to 0
/:cl:

PR #15483, opened 2021-02-09
Refactors DNA Mutations; Refactors Mutations to use Traits

Fox-McCloud:
Cleans up mutations and repaths them.

DNA has been a mess for a long time; this puts us on the path to implementing TG's genetics.

- Repaths `/datum/dna/gene/` to `/datum/mutation`
  - Why were mutations a child of DNA? They use zero attributes of DNA
- Repaths genes so there's not so many children subtypes
  - Disabilities and spell holders are their own subtypes, though these could probably just get moved into the parent type
- Removes references to "genes" in favor of "mutations" with regards to DNA modding
- Removes unused arguments for `domutcheck` and `/datum/mutation`'s `activate` and `deactivate`
- `genemutcheck` is now `singlemutcheck`
- Generic styling fixes

This also converts over muations and species traits to actual traits--this makes things generally better and paves the way for improving a lot of existing systems or implementing new ones that aren't so snowflakey or tied with genetics.

This does have some balance changes associated with it, as I picked up a few traits along the way:

- You cannot defib a husk or use strange reagent on one succesfully
- You'll be able to cure burn husking with synth flesh
- Burn husking is 2x as hard to do now
- Flashbangs consistently do 5 ear damage now
- skeleton species are now immune to tox and clone damage
- Scrying orb gives you its power based on possessing it, as opposed to giving it to you, permanently, when you first acquire it

Fixes:
- Fixes all the instances where a genetic power was activated by just adding the mutation to the mutation list, resulting in some strange behavior with toggling them on/off
- Fixes a ton of edge case instances where having a mutation or ability or power from multiple sources (or a speed boost) would get turned off by something that also turns off a speed boost; each speed-boost should be unique for other ability/powers.
- Fixes slimes being immune to deafness from non-ear related things (such as genetic deafness)
- Being able to cure genetic sources of eye/ear damage with oculine (and visa versa)

:cl: Fox McCloud
add: Burn husking will prevent revival by defib or strange reagent
add: Using enough synth flesh on a burn husked individual will remove the burn husking
tweak: Burn husking requires 2x more burn damage to occur now
fix: Fixes slimes being immune to genetic/magical/other special sources of deafness
fix: Improves instances where toggling on/off genetic mutations was inconsistent or misbehaved
add: Unifies ear damage so it's a bit more unified and less snowflakey
tweak: Flashbangs now consistently do 5 ear damage
tweak: Cannot cure genetic sources of eye/ear damage with mutadone/oculine or visa versa
[...]

PR #15529, opened 2021-02-14
Removes most instances of Xray

S34NW:
## What Does This PR Do
This PR removes the x-ray mutation from genetics, and removes x-ray implants from station protolathes and robotics fabricators.
It keeps x-ray implants for nukies and other special roles that already had them.
Scrying orb still gives x-ray as intended.
Gives thermal implants the same tech level requirements and material costs that x-ray had for RND.

This is an alternative to #15528 that removes x-ray implants from RND, but keeps the mutation on scrying orb and other rarer things such as wish granter.

## Why It's Good For The Game
Xray has been a problem for antags with no counterplay available. If sec or even regular crew have it, the round balance is practically destroyed. Why should the crew easily be able to obtain something that nukies have to pay TC for?

## Changelog
:cl:
del: Removed x-ray implants from RND, removes x-ray gene from genetics.
tweak: Thermal implants now have the material and tech requirements as x-ray implants did.
/:cl:


S34NW:
Having just seen the mess this will cause with #15483, I have decided to close it and will have another look after depending on how #15528 is progressed

PR #15533, opened 2021-02-15
Adds a discord ping for mhelps if there aren't any mentors online

SabreML:
## What Does This PR Do

Adds a config setting to ping mentors on discord if there's no active mentors ingame, the same as the admin ping system.

**IMPORTANT**
I can't actually test this myself, both because I have no clue how to even start testing discord stuff and because I somehow managed to break my server database. It *should* work since it's mostly copied over from the admin system, but if @AffectedArc07 or someone could verify that this actually works I'd appreciate it.

## Why It's Good For The Game

Helps prevent this from happening:
KjFicQg3M0

## Changelog
:cl:
add: Mentors will now be pinged on discord if someone mhelps with no mentors online.
/:cl:

### DIFF
--- a/code/controllers/subsystem/discord.dm
+++ b/code/controllers/subsystem/discord.dm
@@ -5,6 +5,8 @@ SUBSYSTEM_DEF(discord)
 	var/enabled = FALSE
 	/// Last time the administrator ping was dropped. This ensures administrators cannot be mass pinged if a large chunk of ahelps go off at once (IE: tesloose)
 	var/last_administration_ping = 0
+	/// Last time the mentor ping was dropped. This ensures mentors cannot be mass pinged if a large chunk of mhelps go off at once.
+	var/last_mentor_ping = 0
 
 /datum/controller/subsystem/discord/Initialize(start_timeofday)
 	if(config.discord_webhooks_enabled)
@@ -80,5 +82,15 @@ SUBSYSTEM_DEF(discord)
 
 		last_administration_ping = world.time + 60 SECONDS
 		return "<@&[config.discord_admin_role_id]>"
+
+	return ""
+
+/datum/controller/subsystem/discord/proc/handle_mentor_ping()
+	if(config.discord_mentor_role_id)
+		if(last_mentor_ping > world.time)
+			return "*(Role pinged recently)*"
+
+		last_mentor_ping = world.time + 60 SECONDS
+		return "<@&[config.discord_mentor_role_id]>"
 
 	return ""
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -247,6 +247,9 @@
 	/// Role ID to be pinged for administrative events
 	var/discord_admin_role_id = null // Intentional null usage
 
+	/// Role ID to be pinged for mentorhelps
+	var/discord_mentor_role_id = null
+
 	/// Webhook URLs for the main public webhook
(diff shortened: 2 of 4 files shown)


Dumbdumn5 (approved):
If this works properly, then much like we have for ahelps, this'd be nice for mhelps! 👍

AffectedArc07 (approved):
Approved for design. Awaiting host availability to merge since config change.

dearmochi (approved):

AffectedArc07:
Closing and reopening to kick CI checks in

code/controllers/subsystem/discord.dm
```diff
@@ -73,6 +73,27 @@ SUBSYSTEM_DEF(discord)
 	for(var/url in config.discord_admin_webhook_urls)
 		SShttp.create_async_request(RUSTG_HTTP_METHOD_POST, url, dwp.serialize2json(), list("content-type" = "application/json"))
 
+/datum/controller/subsystem/discord/proc/send2discord_simple_nomentors(content)
```
AffectedArc07, line 78: Make this have `check_send_always` as a parameter too, or rename it, since this is gonna send always, despite the name being `nomentors`.

AffectedArc07 (changes requested):
See comment

code/controllers/subsystem/discord.dm
SabreML: These could do with some documentation really, the `check_send_always` argument in particular is pretty ambiguous.
It could do any of these and still be valid:
1. Check if the message should always be sent
2. Check if 'always' should be sent
3. Always send the check
4. Always check 'send'
5. etc

I'll probably just rename the proc since there's no cases when mentorhelps *shouldn't* be sent to the discord channel.

AffectedArc07 (changes requested):
Oh I had this requested change from like a month ago. oops.

Also this PR is on hold till I can get a headmin to setup a reaction role for it.

code/modules/mob/logout.dm
```diff
@@ -7,14 +7,18 @@
 	create_log(MISC_LOG, "Logged out")
 	// `holder` is nil'd out by now, so we check the `admin_datums` array directly
 	//Only report this stuff if we are currently playing.
-	if(GLOB.admin_datums[ckey] && SSticker && SSticker.current_state == GAME_STATE_PLAYING)
+	if(GLOB.admin_datums[ckey] && SSticker?.current_state == GAME_STATE_PLAYING)
```
AffectedArc07: Unneeded. SSticker can never be null. Remove the check entirely.

SabreML: What about if someone logs out before SSticker has loaded? Right as the game's first starting.

AffectedArc07: Subsystems aren't bought into existence on Initialize(). 

They exist as soon as you press go on dreamdaemon.

This check only exists because of the old legacy code that didn't generate the ticker until you pressed start ingame, which is about 10 years old.

SabreML: Ah okay, that makes sense. 👍

SabreML:
This was set to DNM until I could be sure that it wouldn't break something, considering recent events. I'm unsetting it now after further testing.
image

AffectedArc07:
The one mergeblock here is needing a head to make a reaction role for this.

Issue #14830, opened 2020-11-05
AI detector bug


Quarktal:
**Issue Description**:
So, AI detector detects cryo'd AI, there was new AI, that was 99% not in the area (xeno lab), but the storaged AI was there before cryoing and my AI detector was on red 5 minutes while i waited. I even said provocative false message to make current AI look away, and judjung by chat it did.

**What did you expect to happen**: 
AI detector would'nt detect last position of cryo'd AI

**What happened instead**:
AI detector did detect last position of cryo'd AI

**Why is this bad/What are the consequences**:
I was confused, scared and wasted 5 minutes waiting in a locker for ai to go away

**Steps to reproduce the problem**:
Just walked around with uplink multitool i guess.

**When did the problem start happening**:
Everytime i was near xeno lab

**Extra information**:
round ended at 23:45 GMT+3 time (for logz)

farie82:
I believe the xenobio console "eyes" also make the AI detector go off. Can be wrong though but I've seen them go off around xenobio a lot

PR #15535, opened 2021-02-15
AI Detector can now only detect AI, not Abductors, Xenobio consoles or Shuttles

Kepplerx:
Fix for issue #14830

## What Does This PR Do
This PR makes the AI detector (Multitool) only able to detect the AI, instead of all subtypes who use AiEye. This includes the Abductor Console, The Shuttle Console, The Xenobiology Console as well as holocalls.

## Why It's Good For The Game
The AI detector should be just that, an AI detector (Multitool), it makes little sense for it to be able to detect abductors, or shuttles looking for landing sites. The Xenobiology console too would be detected as watching. All of these would result in a false positive for what it said it did, detect if the AI is watching.

## Changelog
:cl:
tweak: AI Detector Multitool only detects the AI, not Abductors nor the Xenobio console
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/freelook/eye.dm
+++ b/code/modules/mob/living/silicon/ai/freelook/eye.dm
@@ -17,7 +17,7 @@
 	var/use_static = TRUE
 	var/static_visibility_range = 16
 	// Decides if it is shown by AI Detector or not
-	var/show_on_multitool = TRUE
+	var/ai_detector_visible = TRUE
 
 
 // Use this when setting the aiEye's location.
--- a/code/modules/mob/living/silicon/ai/freelook/eye.dm
+++ b/code/modules/mob/living/silicon/ai/freelook/eye.dm
@@ -16,7 +16,7 @@
 	var/relay_speech = FALSE
 	var/use_static = TRUE
 	var/static_visibility_range = 16
-	var/AiDetector = TRUE
+	var/AiDetector = TRUE // Decides if it is shown by AI Detector or not
 
 
 // Use this when setting the aiEye's location.
--- a/code/modules/mob/living/silicon/ai/freelook/eye.dm
+++ b/code/modules/mob/living/silicon/ai/freelook/eye.dm
@@ -16,7 +16,8 @@
 	var/relay_speech = FALSE
 	var/use_static = TRUE
 	var/static_visibility_range = 16
-	var/AiDetector = TRUE // Decides if it is shown by AI Detector or not
+	// Decides if it is shown by AI Detector or not
+	var/show_on_multitool = TRUE
 
 
 // Use this when setting the aiEye's location.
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/silicon/ai/freelook/eye.dm
```diff
@@ -16,6 +16,7 @@
 	var/relay_speech = FALSE
 	var/use_static = TRUE
 	var/static_visibility_range = 16
+	var/AiDetector = TRUE // Decides if it is shown by AI Detector or not
```
AffectedArc07, line 19:
```suggestion
	/// Decides if it is shown by AI Detector or not
	var/show_on_multitool = TRUE
```
Autodoc compliance, better var name, as well as case compliance

AffectedArc07 (changes requested):

S34NW:
If you word it as Fixes #14830 it should autoclose that issue when this gets merged

code/modules/mob/living/silicon/ai/freelook/eye.dm
Kepplerx: Fixed the proposed change, it was used more than just those two places, now replaced all the places.

code/modules/mob/living/silicon/ai/freelook/eye.dm
```diff
@@ -16,6 +16,8 @@
 	var/relay_speech = FALSE
 	var/use_static = TRUE
 	var/static_visibility_range = 16
+	// Decides if it is shown by AI Detector or not
+	var/show_on_multitool = TRUE
```
Fox-McCloud, line 20: `ai_detector_visible`

Fox-McCloud:
I'm inclined to say that it should probably detect all camera eyes. 

The point is, you're being watched by *something*; if there's ever a time we have some condition where advanced camera consoles can be used by the crew (and they sorta can with Xenobio), a traitor should probably have some indication that he's being, well, watched.

Being caught because the Xenobio camera guy, or some other source of and advanced camera, was watching you, is kinda BS.

Kepplerx:
> I'm inclined to say that it should probably detect all camera eyes.
> 
> The point is, you're being watched by _something_; if there's ever a time we have some condition where advanced camera consoles can be used by the crew (and they sorta can with Xenobio), a traitor should probably have some indication that he's being, well, watched.
> 
> Being caught because the Xenobio camera guy, or some other source of and advanced camera, was watching you, is kinda BS.
> 
> If an AI cryo's, it should probably have all its related AI eye mob deleted in the process; that's a dangling reference anyway.

This was originally reported as a bug, and a easy to solve issue (very new to coding) with the question of if this was intended or not. If it is intended, that is of course fine, and I can perhaps change this to only make Abductor Cams not detectable. As the item is marketed however, it is an AI detector, and that only, so should perhaps be changed in another way, and so that the Xenobio cam does not make it show red as if an AI is following you.

dearmochi (approved):
Code is fine but I'd probably change the variable name to something some more generic, like `ai_detector_visible` as Fox suggested. However the changelog needs changing -- it's mainly read by players, not coders, so try to word it so that anyone can understand the change effected.

Kyep (approved):

AffectedArc07 (approved):

PR #15540, opened 2021-02-16
SSU improvements

TDSSS:
## What Does This PR Do
This PR includes various tweaks and improvements to Suit Storage Units. 

1. SSUs now accept any suit in the suit slot and and head cover in the helmet slot, instead of just space suits. This was done primarily because we had SSUs with rad suits and explorer suits, both of which aren't space suits, so you couldn't put the suit back once you removed it.
2. Aside from tanks of any size, the storage slot now also accepts any item that's size small or smaller. This was done because the rad suit SSU started with a geiger counter and again, you couldn't put it back once you removed it. IMO it makes sense that most small items could fit in there.
3. SSUs set to dangerous UV no longer delete absolutely every item in them. Items that are lavaproof are now immune to being fried, preventing the destruction of objective items or other high-value equipment.
4. Removed outdated code referring to a dial being turned when you control an SSU's setting via wires now
5. Cleaned up some comments so they work with autodoc/described some variables
6. Replaced some spaces with tabs

## Why It's Good For The Game
Being able to put items back inside is just common sense and was a pretty big inconsistency. The storage slot accepting other, small items makes sense to me as well and makes them a bit more flexible in general.

Destroying indestructible items should not be done, and definitely not by a random piece of machinery with one wire pulsed.

Lastly, code improvements good

## Changelog
:cl:
tweak: SSUs can now accept more items
tweak: Some items are now immune to the SSU's matter-destroying UV radiation
/:cl:

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -8,10 +8,10 @@
 	density = TRUE
 	max_integrity = 250
 
-	var/obj/item/clothing/suit/space/suit = null
-	var/obj/item/clothing/head/helmet/space/helmet = null
+	var/obj/item/clothing/suit/suit = null
+	var/obj/item/clothing/head/helmet = null
 	var/obj/item/clothing/mask/mask = null
-	var/obj/item/clothing/shoes/magboots/magboots = null
+	var/obj/item/clothing/shoes/magboots = null
 	var/obj/item/storage = null
 
 	var/helmet_type = null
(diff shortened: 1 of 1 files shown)


hal9000PR:
Would it be possible to make SSU moveable/creatable. Sometimes they get in the way of repiping atmos.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -649,18 +663,12 @@
...
 	else
-		if(uv_super)
-			to_chat(user, "<span class='notice'>You slide the dial back towards \"185nm\".</span>")
-			uv_super = FALSE
-		else
-			to_chat(user, "<span class='notice'>You crank the dial all the way up to \"15nm\".</span>")
-			uv_super = TRUE
+		uv_super = !uv_super
 
 /obj/machinery/suit_storage_unit/proc/togglesafeties(mob/user)
```
farie82, line 668:
```suggestion
/obj/machinery/suit_storage_unit/proc/togglesafeties()
```
No longer uses an user param so it can be removed. Ensure to change the usages as well

code/game/machinery/suit_storage_unit.dm
```diff
@@ -324,19 +333,19 @@
...
 		. = TRUE
 	if(istype(I, /obj/item/clothing/mask) && !mask)
 		mask = I
 		. = TRUE
-	if(istype(I, /obj/item/clothing/shoes/magboots) && !magboots)
+	if(istype(I, /obj/item/clothing/shoes) && !magboots)
 		magboots = I
 		. = TRUE
-	if((istype(I, /obj/item/tank)) && !storage)
+	if(((istype(I, /obj/item/tank)) || I.w_class <= WEIGHT_CLASS_SMALL) && !storage && !.)
```
farie82, line 348:
```suggestion
	if((istype(I, /obj/item/tank) || I.w_class <= WEIGHT_CLASS_SMALL) && !storage && !.)
```

code/game/machinery/suit_storage_unit.dm
```diff
@@ -8,10 +8,10 @@
 	density = TRUE
 	max_integrity = 250
 
-	var/obj/item/clothing/suit/space/suit = null
-	var/obj/item/clothing/head/helmet/space/helmet = null
+	var/obj/item/clothing/suit/suit = null
+	var/obj/item/clothing/head/helmet = null
 	var/obj/item/clothing/mask/mask = null
-	var/obj/item/clothing/shoes/magboots/magboots = null
+	var/obj/item/clothing/shoes/magboots = null
```
farie82, line 14: `magboots` is misleading here since it doesn't have to be magboots anymore

farie82 (changes requested):

TDSSS:
Addressed farie's comments.

On the making SSUs movable bit, I'm not sure. If an SSU is in the way of a build project, you can always ahelp to have it moved or deleted. Letting people just wrench them and steal suits that way might be too much.

Kyep (approved):

AffectedArc07 (approved):

PR #15538, opened 2021-02-16
Lets syndie medborgs deal with crit and shock without cluttering their hypospray

TDSSS:
## What Does This PR Do
As an alternative to #15517 , this PR simply modifies syndicate nanites so they cure both heart attacks and shock, making it unncessary for the syndie medical borg to have saline or epinephrine while still letting it deal with the challenges of newcrit.

## Why It's Good For The Game
As I  commented on that PR, syndie medical borgs chose their chem by cycling through them one at a time, needing to cycle through SIX of them to get to the one you want in the middle of an intense gunfight is very clunky.  You definitely don't want a new ghost pulled into the borg to accidentally inject his nukie buddy with poison because he was cycling through medical chems and went one too far.
This also prevents the borg from stacking the healing effects of saline glucose and medical nanites.

Nanites are supposed to be mostly a one size fits all solution to healing. You paid 50 TC for this borg, you better get your money's worth, so curing newcrit seems like a justifiable buff for such a unique chemical. Lorewise, it's NANOMACHINES SON. Or in other words, supertech medical nanites being able to cure shock or stop a heart attack makes sense to me.

As for spaceacillin, I don't see nukie rounds lasting long enough for a medical borg to do surgery and for the resulting infection to become relevant.

Note they are not added to the cure_text, since IMO it wouldn't make much sense for NT medical equipment to list a syndie-exclusive technology as a possible cure, even if it's theoretically one.

## Changelog
:cl:
tweak: syndicate nanites now cure shock and heart attacks on top of their other effects
/:cl:

PR #15517, opened 2021-02-13
Adds saline glucose, epinephrine, and spaceacillin to syndicate medical cyborg

warior4356:
## What Does This PR Do
As the title states, this adds saline glucose, epinephrine, and spaceacillin to the syndicate medical cyborg's hypospray.

## Why It's Good For The Game
Currently, the syndicate cyborg is a bit helpless when it comes to new crit. The normal medical borg has these tools to treat shock and heart failure. Shock is more of the minor one as they can treat the damage causing shock, but there isn't really harm in giving them saline glucose when they have nanites already. Heart failure is the more significant issue as they have no way to cure this disease, in fact, the only way to save a nukie with heart failure is to take them to the sleeper on their ship as of now. Finally I'm giving them spaceacillin because the normal medical borg has it, and it prevents needing surgery to deal with an infection.

## Changelog
:cl:
add: Adds saline glucose, epinephrine, and spaceacillin to syndicate medical cyborg
/:cl:

PR #15541, opened 2021-02-16
Better way for syndie medical borgs to treat newcrit, v2

TDSSS:
## What Does This PR Do
This is basically #15538 which also reverts #15517

To sum up, syndicate nanites now cure shock and heart attacks, on top of their other properties. This means syndie medical borgs no longer need to cycle through 6 different chems, which is very clunky. The newcrit treatment meds were removed from syndie medical borgs again, as well as spaceacillin 

## Why It's Good For The Game
First: Why am I reopening a PR that got closed by a maint, surely they considered both alternatives and decided against mine, right? :hammer:

By my understanding, #15517 got approval from both fox and kyet for a merge _before_ my alternative PR was made, so they didn't know there was an alternative even being put forward, which led to it being merged before anyone commented on my alternative (kyet's approval being timestamped two hours before mine was made). If maints/head decided against my PR on its merits rather than how the timing worked out, I apologize of course and will close this again.

Considering my PR got 6 thumbs up while the alternative got none, and that even the author of #15517 states he prefers my take on it, I decided to open this. Being able to treat newcrit is a good thing for a syndie med borg, agreed, but why go for the clunkier way to do it?

To reiterate why I think this approach is better: Borg hyposprays work by cycling through chems one at a time. Going from 3 to 6 chems to cycle through makes them a good deal less usable and more clunky. Syndicate nanites are used nowhere else and have good justifications for being very powerful, so serving as a cure-all for newcrit conditions seems fair to me. 

I decided against also adding infection treatment to syndie nanites for two reasons: 1) duplicating the infection treatment code from spaceallin seemed very clunky and 2) Nukie rounds are over quicker than an infection can really come into play, imo.

## Changelog
:cl:
del: Syndicate medical borgs no longer have saline glucose, epinephrine or spaceacillin 
tweak: Syndicate nanites now cure shock and heart attacks
/:cl:

### DIFF
--- a/code/datums/diseases/critical.dm
+++ b/code/datums/diseases/critical.dm
@@ -25,7 +25,7 @@
 	max_stages = 3
 	spread_flags = SPECIAL
 	cure_text = "Saline-Glucose Solution"
-	cures = list("salglu_solution")
+	cures = list("salglu_solution", "syndicate_nanites")
 	cure_chance = 10
 	viable_mobtypes = list(/mob/living/carbon/human)
 	stage_prob = 6
@@ -86,7 +86,7 @@
 	max_stages = 3
 	spread_flags = SPECIAL
 	cure_text = "Atropine, Epinephrine, or Heparin"
-	cures = list("atropine", "epinephrine", "heparin")
+	cures = list("atropine", "epinephrine", "heparin", "syndicate_nanites")
 	cure_chance = 10
 	needs_all_cures = FALSE
 	viable_mobtypes = list(/mob/living/carbon/human)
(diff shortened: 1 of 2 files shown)


Spacemanspark:
Perhaps it would be better to have a pop-up menu like RCDs do

TDSSS:
> Perhaps it would be better to have a pop-up menu like RCDs do

I think the issue with that would be the lack of icons for the different meds, since they're all just differently colored liquids.

warior4356:
This still leaves the issue of infection.

Wojz:
I prefer this, and I agree with the spaceacillin thing. It's unnecessary bloat, nukies don't get infections because they don't get surgery, and if they did get surgery, it'd be from a borg that doesn't cause infections lol.

dearmochi (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #15543, opened 2021-02-16
Adds a cancel combo button for krav maga 

hal9000PR:
## What Does This PR Do
Adds a button that cancels your prepared attack for krav maga
Also makes neck chop only damage the head

## Why It's Good For The Game
good quality of life change, the warden doesn't have to leg sweep themselves ~~or the clown~~ to cancel their prepared combo. Also stops the warden from neck chopping an officer when they wanted to hug them.

## Changelog
:cl:
add: krav maga has a cancel combo button. 
tweak: makes neck chop only damage the head
/:cl:


S34NW:
YES

Cocacolagua:
could you make the stance cancel by selecting the same stance you're already on instead? adding another HUD icon makes the actions clutter even bigger, given there’s four buttons now

code/modules/martial_arts/krav_maga.dm
```diff
@@ -3,6 +3,17 @@
 	var/datum/action/neck_chop/neckchop = new/datum/action/neck_chop()
 	var/datum/action/leg_sweep/legsweep = new/datum/action/leg_sweep()
 	var/datum/action/lung_punch/lungpunch = new/datum/action/lung_punch()
+	var/datum/action/neutral_stance/neutral = new/datum/action/neutral_stance()
+
+/datum/action/neutral_stance
+	name = "Neutral Stance - You relax, cancelling your last Krav Maga stance attack."
+	button_icon_state = "neutralstance"
+
+/datum/action/neutral_stance/Trigger()
```
dearmochi, line 12: It looks like this can be repeated infinitely, even if there is no active stance -- it could be used to spam the chat messages below.

code/modules/martial_arts/krav_maga.dm
```diff
@@ -3,6 +3,17 @@
...
+	to_chat(owner, "<b><i>You cancel your prepared attack.</i></b>")
+	owner.visible_message("<span class='danger'> [owner] relaxes his stance.")
```
dearmochi, line 14:
```suggestion
	owner.visible_message("<span class='danger'>[owner] relaxes [owner.p_their()] stance.</span>", "<span class='notice'>You cancel your prepared attack.</span>")
```

In HTML, any opening tag needs a closing tag. Also changed `his` to the owner's gender pronoun.
Finally, `visible_message` allows a second argument equal to the message displayed to the owner only -- no need to do a separate `to_chat` call.

dearmochi (changes requested):
Echoing what @Cocacolagua said, I think it would be more intuitive (and less cluttering) if one could just click the active stance again to cancel it

code/modules/martial_arts/krav_maga.dm
hal9000PR: you can currently do this with all Krav maga actions. And I don't see it being abused, however I will restrict it to only if there is an active stance.

Dumbdumn5 (approved):
_Imagine a warden animation-cancelling his sweep,, wavedashing behind you, and then tossing you into space._ :+1:

code/modules/martial_arts/krav_maga.dm
```diff
@@ -3,6 +3,20 @@
...
+	name = "Neutral Stance - You relax, cancelling your last Krav Maga stance attack."
+	button_icon_state = "neutralstance"
+
+/datum/action/neutral_stance/Trigger()
+	var/mob/living/carbon/human/H = owner
+	if(!H.mind.martial_art.in_stance)
+		to_chat(owner, "<b><i>You cannot cancel an attack you haven't prepared!</i></b>")
+		return
+	to_chat(owner, "<b><i>You cancel your prepared attack.</i></b>")
+	owner.visible_message("<span class='danger'> [owner] relaxes his stance.</span>")
```
AffectedArc07, line 18: *I swear this isnt tumblr*

Please use the appropriate pronoun here 
```suggestion
	owner.visible_message("<span class='danger'> [owner] relaxes [owner.p_their()] stance.</span>")
```

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #15547, opened 2021-02-17
Station ceiling collapsing fix

### LINKED ISSUE #15398: Tendrils cause the stations ceiling to collapse.
https://github.com/ParadiseSS13/Paradise/blob/e8272302045f219de6953fa5a5092413de894142/code/game/turfs/simulated/floor/chasm.dm#L129-L137
As the station is on Z2 now, and Lavaland on Z3, when a tendril gets destroyed, and such generates chasms, you can see the ceiling collapse on station. Given we don't use Z levels as actual Z levels, but more as different instances, i think that could be fully removed to fix this bug.

SabreML:
## What Does This PR Do

Removes the 'The ceiling gives way!' message on station if a miner destroys a tendril on lavaland. This was clearly intended for some sort of multi-z thing but since we don't have that it's just confusing.

Fixes #15398

## Why It's Good For The Game

The Ceiling is not real.
The Ceiling has never been real.
Do not think about The Ceiling.

## Changelog
:cl:
tweak: Removed the 'The ceiling gives way!' message on the station when a miner destroys a tendril.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/floor/chasm.dm
+++ b/code/game/turfs/simulated/floor/chasm.dm
@@ -131,10 +131,6 @@
 	drop_x = x
 	drop_y = y
 	drop_z = z - 1
-	var/turf/T = locate(drop_x, drop_y, drop_z)
-	if(T)
-		T.visible_message("<span class='boldwarning'>The ceiling gives way!</span>")
-		playsound(T, 'sound/effects/break_stone.ogg', 50, 1)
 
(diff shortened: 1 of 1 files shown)


dearmochi (approved):

Kyep (approved):

procdrone:
the ceiling... is a lie? they can't hide the truth forever...

AffectedArc07 (approved):

PR #15519, opened 2021-02-13
Converts Eye Augments to Cybernetic Eyes

Fox-McCloud:
Converts cybernetic eye augments into just straight cybernetic eyes, based on TG.

There's so much code overlap between eyes and eye augments, it's kinda gross; they both change your eye color, both have a range on what it grants darksight to your vision and both have the level of see invisible that is set.

All in all, this is a lot of code duplication for little benefit--it's gross.

This refactors eyes to have the ability to have flash protection and tint, as well as all their regular settings; the end result is that we have more customization for eyes, and we have less copy pasta in the code, overall.

This also removes `weakeyes`, because, quite frankly, keeping it around for one game mode is kinda ridiculous, and it doesn't really come into play *that* much; using light and pew pew is still more effective, overall.

This also adds night vision eyes---all shadow species have them; you can toggle your vision abilities, natively, on the eyes.

I couldn't purge `/datum/vision_override` completely, as the traits PR isn't alive yet, but I'd like to, eventually---better to handle things with traits and eyes, directly; having all these systems, together, is confusing.

This also ports TG's flashlight eyes; craft them with cablecuffs and 2 flashlights---put them in your head to light the way----at the cost of you having no vision. Why? Because why not.

also standardizes on `eye_color` because well---consistency.

This also makes the Lings' augmented eyesight change the stats on eyes, directly, instead of creating new eyes over and over again (awful). It's also been buffed to just provide straight up x-ray vision instead of thermals.

Also noted that we have duplicate torch sprites when I added the flashlight eyes sprite--- got rid of the old ones.

Lastly, this updates the autoimplaners (now autosurgeons) so they're generally more compatible with internal organs.

:cl: Fox McCloud
tweak: cybernetic eye augments are now full eyes, instead; you'll have to rip someone's previous eyes out and put new ones in
add: Adds flashlight eyes; craft them with 2 flashlights and cablecuffs
tweak: shadow people (and shadowlings) have built in night vision giving eyes with their very own toggle
tweak: Ling vision now gives x-ray instead of thermal vision
fix: fixes having duplicate torch sprites
add: Syndicate auto-surgeons have their own unique icon
/:cl:

PR #15548, opened 2021-02-17
Moves External Organs to Nullspace

Fox-McCloud:
Moves external organs to `null` loc---these really should *NOT* be in a mob's contents---it not only bloats up contents searches, but it makes searches through `contents` (normally what you're wearing and in your hands) really wonky and prone to causing errors or odd behavior. Internal organs are in `null` space and so should external ones---as you can tell from all the removed code here, this was QUITE the problem, back in the day, with a snowflake check in `canUnEquip` and lots of loops having to explicitly check for this.

Eventually, implants should probably be removed from general contents as well, but I don't feel like doing that in this PR.

Things I've tested:

- removing limbs
- re-attaching limbs
- Rejuvinating dismembered people
- Stripping someone of everything
- Equipping someone with an outfit
- monkeying/unmonkeying someone
- skrell headpocket usage

Things seem to function fine.

Also fixes an improper check for lungs that made you die too slowly, while in space.

:cl: Fox McCloud
fix: Fixes external organs being in a user's contents
fix: Fixes not dying rapidly, in space, when you can't breathe
/:cl:


KasparoVy:
Cheers, I always figured it should be in contents for SOME reason but never understood why
Good path out

SteelSlayer:
This will break EMPs on humans since EMPing their external organs relies on a loop through `contents`:
```dm
/mob/living/emp_act(severity)
	..()
	for(var/obj/O in contents)
		O.emp_act(severity)
```
Could possibly break other things but this is the only thing I can think of at the moment.

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -66,14 +66,6 @@
...
 			return TRUE
 
-// The actual dropping happens at the mob level - checks to prevent drops should
-// come here
-/mob/living/carbon/human/canUnEquip(obj/item/I, force)
-	. = ..()
-	var/obj/item/organ/O = I
-	if(istype(O) && O.owner == src)
-		. = 0 // keep a good grip on your heart
-
```
SteelSlayer, line 76: Is there a reason thing is being removed?

dearmochi: I'd imagine it'd be because `unEquip` is typically called when looping through one's `contents`, and that check would make more sense there as a failsafe

code/modules/surgery/organs/lungs.dm
```diff
@@ -88,7 +88,7 @@
 		return
 
 	if(!breath || (breath.total_moles() == 0))
-		if(isspaceturf(loc))
+		if(isspaceturf(H.loc))
```
dearmochi, line 91: Edge case to be careful about is being in a closed locker in space.

dearmochi:
https://github.com/ParadiseSS13/Paradise/blob/9fef0d6a16154a5f9aad5b2814a555d709b72cf1/code/datums/mind.dm#L1521-L1522

https://github.com/ParadiseSS13/Paradise/blob/9fef0d6a16154a5f9aad5b2814a555d709b72cf1/code/game/gamemodes/shadowling/special_shadowling_abilities.dm#L33-L34

https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/reagents/chemistry/reagents/water.dm#L275

These subtractions are probably redundant now

code/datums/outfits/vv_outfit.dm
```diff
@@ -125,9 +125,9 @@
 
 	// Copy cybernetic implants
 	O.cybernetic_implants = list()
-	for(var/obj/item/organ/internal/CI in contents)
-		if(istype(CI))
-			O.cybernetic_implants |= CI.type
+	for(var/org in internal_organs)
+		var/obj/item/organ/internal/aug = org
```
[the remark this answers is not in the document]
Fox-McCloud: Yeah, I did this on purpose because: https://github.com/ParadiseSS13/Paradise/pull/15519

code/modules/surgery/organs/lungs.dm
Fox-McCloud: Yeah, this isn't perfect, to be sure---I thought about this a bit when I noticed the error, but using something like `get_turf` is going to result in a lot weird behavior, like taking in crease O2 damage when you're in a mech or a pod while on a space turf.

code/modules/mob/living/carbon/human/inventory.dm
Fox-McCloud: Yeah, this isn't needed and is something we want to have happen now---because organs and bodyparts, post this PR should always have `null` as their loc if they're inside someone.

Wed did this and all the snowflakey checks because when limbs were made into objects (they used to be datums....thanks Baystation), the individual who ported them just kinda ported the system then barely fixed the bugs with it---it caused us major issues for a couple of months, and this unequip check was nuclear option, after we were unable to catch all the loops through internal organs and bodyparts (which used to all be in the same list).

dearmochi (approved):

AffectedArc07 (approved):

PR #15549, opened 2021-02-17
Explosive Implants now always possible to activate (Meant for Nukies especially)

Kepplerx:
## What Does This PR Do
Explosive implants are now always possible to activate, even while asleep, stunned or other. Also adds a new subtype of the action datum that can be used for other items or implants. Did the same change to Dust implants as well.

## Why It's Good For The Game
This PR is primarily for nukies, who often go down, but dont push the button in time, which in turn means they are looted for gear then used against the other nukies. It is actually possible to blow up regardless, by suiciding, which means that exploding the implant is an option with or without this PR. As such making this choice a lot more obvious by the "blow up" button, makes nukies a lot more able to use it, especially new ones. Dust implants are the same, and may be even more relevant.

## Images of changes
image
The button is available even while asleep as shown above

## Changelog
:cl:
tweak: Can now always activate the Explosive Implants (Nukies for example), the same applies to dust implants
/:cl:

### DIFF
--- a/code/datums/action.dm
+++ b/code/datums/action.dm
@@ -383,6 +383,9 @@
 /datum/action/item_action/hands_free/activate
 	name = "Activate"
 
+/datum/action/item_action/hands_free/activate/always
+    check_flags = null
+
 /datum/action/item_action/toggle_research_scanner
 	name = "Toggle Research Scanner"
 	button_icon_state = "scan_mode"
--- a/code/game/objects/items/weapons/implants/implant_explosive.dm
+++ b/code/game/objects/items/weapons/implants/implant_explosive.dm
@@ -3,6 +3,7 @@
 	desc = "And boom goes the weasel."
 	icon_state = "explosive"
 	origin_tech = "materials=2;combat=3;biotech=4;syndicate=4"
+	actions_types = list(/datum/action/item_action/hands_free/activate/always)
 	var/weak = 2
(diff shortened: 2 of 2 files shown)


Wojz:
Basically an anti-baldness quality of life feature that stops nukie rounds from being ruined by one dummy with a sniper rifle, I like it.

Kyep (approved):

procdrone:
huh. I could've swore it already worked like that. how weird. Either way, good pr.

Cocacolagua:
please change the changelog to "you can now activate explosive implants while asleep", since the changelog is primarily directed to players which have no idea what datums are

dearmochi (approved):

TDSSS:
Does this work for dust implants as well?

SteelSlayer (approved):

Wojz:
> Does this work for dust implants as well?

This is a good point. Does this include both micro and macrobomb as well as dust implant?

Kepplerx:
> > Does this work for dust implants as well?
> 
> This is a good point. Does this include both micro and macrobomb as well as dust implant?

At the moment it includes both bombs, and I were planning to add the dust implant as well, since that is basicly the same, just havent had time till now.

Kepplerx:
> Does this work for dust implants as well?

I am gonna add this as well yes

PR #15552, opened 2021-02-18
Better Alt Click

Fox-McCloud:
From TG.

Uses a better check for AltClick behavior, so it works better (and isn't so laggy) when you use it on some objects, particularly ones on your person or in your bag.

This also makes a unique alt-click for ghosts (and potentially for other mobs) so it won't send a signal---not used, at the moment, but this will likely be used for a coming-semi-soon-refactor™️, so might as well nab this now.

:cl: Fox McCloud
tweak: alt click behavior improved and less laggy, particularly for items on your person or in bags
/:cl:


code/_onclick/click.dm
```diff
@@ -331,16 +331,19 @@
...
-			// If we had a method to force a `Stat` update, it would go here
-		else
-			user.listed_turf = null
-	return
+	if(T && (isturf(loc) || isturf(src)) && user.TurfAdjacent(T))
+		user.listed_turf = T
+		user.client.statpanel = T.name
+
+/// Use this instead of [/mob/proc/AltClickOn] where you only want turf content listing without additional atom alt-click interaction
+/atom/proc/AltClickNoInteract(mob/user, atom/A)
```
Fox-McCloud, line 342: This doesn't have to have all the same checks since this is primarily ghosts.

dearmochi (approved):

code/_onclick/click.dm
```diff
@@ -331,16 +331,19 @@
...
 	var/turf/T = get_turf(src)
-	if(T)
-		if(user.TurfAdjacent(T))
-			user.listed_turf = T
-			user.client.statpanel = T.name
-			// If we had a method to force a `Stat` update, it would go here
-		else
-			user.listed_turf = null
-	return
+	if(T && (isturf(loc) || isturf(src)) && user.TurfAdjacent(T))
```
SteelSlayer, line 337: I don't see what the `(isturf(loc) || isturf(src))` checks are trying to accomplish here, since one of these should always be true theoretically.

A) You alt-click on some atom thats sitting on a turf, `isturf(loc)` is true: **success**
B) You alt-click on a turf directly, `isturf(src)` is true: **success**
C) You alt-click on an atom which is inside something, like a locker, `(isturf(loc) || isturf(src))` are both false: **failure**

So scenario C is the only time you can fail, but it's impossible for this to happen, due to you not being able to see the atom to alt-click on, since it's inside another atom.

Fox-McCloud: >So scenario C is the only time you can fail, but it's impossible for this to happen, due to you not being able to see the atom to alt-click on, since it's inside another atom.

Objects in your inventory or your backpack are neither a turf nor is their `loc` a turf.

SteelSlayer: Yeah you're right, and you know what.. I had no idea a `get_turf(src)` of an item in a backpack ultimately returned the turf it's on, I thought it would be null. There must be something going on behind the scenes I wasn't aware of.

AffectedArc07 (approved):

Issue #15200, opened 2020-12-27
Changelog Suggestion: CLING Headslugs Start on Harm instead of Help


MrDimento:
**Problem Description**:
As a Headslug, captured and about to be defeated, I tried to squirm my way out of a nasty spot, but kept being pushed around as I realized I had switched over to HELP instead of HARM.

**What did you expect to happen**: 
Headslugs should start on Harm instead, to allow them more chance to get away and ventcrawl somewhere safe.

PR #15557, opened 2021-02-18
Headslugs and Blobbernaughts now start on HARM intent

Kepplerx:
Fixes https://github.com/ParadiseSS13/Paradise/issues/15200

## What Does This PR Do
Makes Cling headslugs and Blob blobbernaughts start on HARM intent.

## Why It's Good For The Game
Blobbernaughts are hostile creatures, and have no real reason to start on help intent, other than to hinder friendly beating, which doesnt seem like a real issue. This would make blobbers slightly more effective as there would be no need to remember to turn on harm intent.

Headslugs are the same, they exist for one reason only, to get a new body. They are a hostile to all other lifeforms, and in a case where it was necessary to become a slug, it would likely be necessary to defend itself as well. It would also be a little less easy to push around, so little bit more likely to make it to a vent.

Both of these make the game slightly easier for new players, which I am also a fan of (and I forget both these all the time too).

## Changelog
:cl:
tweak: Headslugs and blobbernaughts now start on harm intent, instead of help intent
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -180,6 +180,7 @@
 	see_in_dark = 8
 	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE
 	move_resist = MOVE_FORCE_OVERPOWERING
+	a_intent = INTENT_HARM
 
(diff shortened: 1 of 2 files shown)


Kepplerx:
Got a possible issue with Blobbernaughts being aggressive if started on harm intent, so gonna need to check that, as there would be a chance they chased and died even before getting a player, so gonna need to check that first.

Kyep (approved):

AffectedArc07 (approved):

PR #15562, opened 2021-02-19
Tweaks cling revive to be an aheal

### LINKED ISSUE #15560: Changeling reviving with suicide command
**Issue Description**:

Possible unintended issue, undetermined until a coder reviews.

This stops you from being cloned even though you are currently living. Reviving from the suicide state which is supposed to be a DNR state and completely fools those around you if you are a changeling.

**What happened instead**:

Suicide was said to be a DNR state but that isn't true for changelings. Changelings are able to revive from the suicide command as stated.

The changeling can no longer use the cloner machine even if they are alive.
This could be a bug abuse/unintended or maybe even power gaming. Coder review is required.

I suicide, revive using changeling powers and so on. I made a video showcasing the steps reproduced. https://www.youtube.com/watch?v=M7ipgXd81mY&feature=youtu.be&ab_channel=Sneed

**Extra information**:

I've used this only twice and both times have been a clean get away as changeling whether this is intended or not I am unsure.  I am unsure if it's bug abuse because changelings are suppose to revive and yet the suicide state is not suppose allow for revival and yet here we are at a state where they both counteract each other in someway. You surely get your free get away as changeling from security but cannot use the cloner. It is a lose and win situation.

hal9000PR:
## What Does This PR Do
Fixes #15560

Makes cling revive aheal the user.
This fixes a bug where you would choke after being revived
It also fixes a bug where your head, arm and leg items (stuff like headsets, gloves, glasses, etc.) would fall onto the floor when you revive.
Also has some balance implications, Implants will remain after revive (stuff like CNS and demon hearts), clings will extinguish themselves and bad genes and viruses will be cured and removes some status affects that weren't removed before, however this last one doesn't matter as much because it takes a while to revive and in that time the confusion/dizzyness/hallucinations will probably run out in that time. (there might be some things that i have missed, please bring them up in the comments).

## Why It's Good For The Game
Bug fixes are good, stops clings reviving only to go down again because they have a bunch of oxyloss cycles left over from crit so they choke to death. 

Also  allows plasmamen clings to actually use the ability without dying because their helmet will no longer be dropped from their body when they use the ability. stops clings from loosing insuls/shoes/sechuds when they revive.

Also it doesn't really make much sense for clings to completely destroy implants and unique organs when they revive, why would the cling destroy this advantageous organ in exchange for a standard heart? Clings are supposed to be the masters of their own genes and bodies, surely they would keep things in their body that would help them.

## Changelog
:cl:
fix: Clings no longer choke when they revive
fix: Clings no longer get stripped of their gloves, boots and headgear when they revive
tweak: clings now don't lose implants and organs when they revive
/:cl:


Coolrune206:
If clings no longer lose implants when reviving, does this mean that Clings won't be able to use regenerative stasis to remove Tracking Implants from their bodies? It's presently somewhat of an effective strategy for a changeling that is in perma but not known to be a Cling to use their abilities to escape, then regenerative stasis to remove the tracking implant so sec can't immediately hunt them down. If this change would actually make Clings keep tracking implants, it feels like this would be a bit of a nerf to their stealth potential.

Aracino:
If you want to go a bit extra, having clings be able to choose which implants they keep when they regen would be nice. Mindshield is great while you're pretending to be an officer, but when your covers blown it's a liability that makes you stand out.

MRgaty:
> If you want to go a bit extra, having clings be able to choose which implants they keep when they regen would be nice. Mindshield is great while you're pretending to be an officer, but when your covers blown it's a liability that makes you stand out.

If you want to pretend to be an officer mindshield included you should suffer all the downsides having a mindshield has IMO.

Rurik123:
Should be noted clings keeping implant on revive, in the end, will hinder more than it helps. I also assume this pr would make it so on regen, cuffs are not taken off, thus on the way to the cremator clings will have a harder time reviving and running. I'd imagine that would mean, if cuffed, clings will be unable to escape unless they took biodegrade.

hal9000PR:
> Should be noted clings keeping implant on revive, in the end, will hinder more than it helps. I also assume this pr would make it so on regen, cuffs are not taken off, thus on the way to the cremator clings will have a harder time reviving and running. I'd imagine that would mean, if cuffed, clings will be unable to escape unless they took biodegrade.

Clings will still remove their cuffs when they revive, I don't see much of an issue with this considering once they revive you can tase them once and they go down because they have no chems left. 

> If clings no longer lose implants when reviving, does this mean that Clings won't be able to use regenerative stasis to remove Tracking Implants from their bodies? It's presently somewhat of an effective strategy for a changeling that is in perma but not known to be a Cling to use their abilities to escape, then regenerative stasis to remove the tracking implant so sec can't immediately hunt them down. If this change would actually make Clings keep tracking implants, it feels like this would be a bit of a nerf to their stealth potential.

Clings already destroy implants/cuffs when they transform between races so they still have the ability to do this.

Rurik123:
Ah, roger. Thanks for the clarification.

dearmochi (approved):
Code is as simple as it can be

Sneedstar:
Lines of code deleted, reduced to a cleaner form in which results in simple style of life improvements such as reviving does not remove clothes or implants. Things like tracking implants/implants are still removed transforming so the outcome is only positive from this as far as I can see.

Kyep (approved):

hal9000PR:
Update: clings no longer lose implanter implants (storage, mindslave, tracker, etc) when they revive or transform on master, so that is no longer a change. They do still lose cybernetic implants so that is still a change. @Coolrune206  @Rurik123

Rurik123:
Cheers for the update 👍

S34NW (approved):
For what it's worth, this looks fine.

AffectedArc07 (approved):
Fuck it

PR #15015, opened 2020-11-29
Adds the ability to pat out fires on burning people

SabreML:
## What Does This PR Do

Adds the ability to put out fires on another player by clicking on them with help intent.
While doing this there's a 30% chance that you'll burn your hand, and take some damage.

## Why It's Good For The Game

It's always a bit awkward to just stand around staring at someone burning to death if there's no fire extinguishers.
This way people can at least do something proactive.

## Changelog
:cl:
add: Using help intent on someone who is on fire will now help put the fire out
/:cl:

PR #15563, opened 2021-02-19
Fire patting improvements

SabreML:
## What Does This PR Do

Improves the fire patting code from #15015 by making it take priority over other help intent acts such as CPR, and making it work on dead people.
Also made it its own proc, slightly refactored it so that `src` is the mob doing the patting, and added documentation.

## Why It's Good For The Game

Doesn't really make sense that you can't pat out fires on dead people.

## Changelog
:cl:
fix: Patting out a fire on someone now works even if they're dead or in crit.
/:cl:


Dumbdumn5 (approved):
While breathing life into a flaming corpse is metal as fuck, probably not what was intended. 👍

AffectedArc07 (approved):

Issue #15184, opened 2020-12-24
Crematorium works without power


Qwertytoforty:
**Issue Description**:

Crematorium works without power, which seems strange, and prevents a clever sabotage by cutting power to the chapel. 

**What did you expect to happen**: 

For the button to not work, and for the person would not be cremated

**What happened instead**:

Ashes to ashes, dust to dust

**Why is this bad/What are the consequences**:

Powered things should not work without power

**Steps to reproduce the problem**:

Sabotage / drain / break APC in chapel office

put person in crematorium

Press button.

Note screaming and ash generation.

PR #15565, opened 2021-02-19
Crematorium now needs power to work and consumes power, and a lot of power when burning

Kepplerx:
Fixes https://github.com/ParadiseSS13/Paradise/issues/15184

## What Does This PR Do
Makes the crematorium (Or rather the button, but in effect the crematorium) consume power, and need power to start. It needs equipment to be on, as that is its source of power. In addition it consumes 40%ish percent of the starting cell in Chapel Office.

## Why It's Good For The Game
First of all making it need power makes it a lot easier for antags to sabotage it, apart from breaking it. Second, it now will not work if there is no power available, as one would expect. AIs and borgs now can hinder its use in other ways than locking its doors. Clings EMP will also hinder it (as it now relies on the APC). 

## Changelog
:cl:
tweak: Crematorium now consumes power, and a lot of power when it burns (40% percent of the starting cell)
fix: Crematorium now needs power to be on to work
/:cl:

### DIFF
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -457,7 +457,10 @@
 	name = "crematorium igniter"
 	icon = 'icons/obj/power.dmi'
 	icon_state = "crema_switch"
-	power_channel = ENVIRON
+	power_channel = EQUIP
+	use_power = IDLE_POWER_USE
+	idle_power_usage = 2
+	active_power_usage = 1000
 	anchored = 1.0
 	req_access = list(ACCESS_CREMATORIUM)
 	var/on = 0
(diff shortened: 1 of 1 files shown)


TDSSS:
> 16 commits

I like the idea of this PR, but you should get some help with the git workflow.

code/game/objects/structures/morgue.dm
```diff
@@ -469,13 +473,17 @@
...
-		for(var/obj/structure/crematorium/C in world)
-			if(C.id == id)
-				if(!C.cremating)
-					C.cremate(user)
-	else
-		to_chat(usr, "<span class='warning'>Access denied.</span>")
+	if(powered(power_channel)) // Do we have power?
+		if(allowed(usr) || user.can_advanced_admin_interact())
+			use_power(400000)
+			for(var/obj/structure/crematorium/C in world)
```
AffectedArc07, line 479: I hate this loop

AffectedArc07 (approved):

PR #15568, opened 2021-02-19
Changes wall welding to be intent based

SabreML:
## What Does This PR Do

Changes repairing or dismantling a wall to be based on intent rather than a popup.
Help, Disarm, or Grab intent to repair.
Harm intent to dismantle.

## Why It's Good For The Game

Massively improves the flow of repairing walls, since you don't have to click on a popup every time you want to do something.
I personally also find it more intuitive to use, since it's similar to other things like welding an airlock.

## Changelog
:cl:
tweak: Repairing/Dismanting a wall with a welder is now intent based rather than popup based.
tweak: Help/Disarm/Grab intent to repair, Harm intent to dismantle.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/walls.dm
+++ b/code/game/turfs/simulated/walls.dm
@@ -7,7 +7,7 @@
 	desc = "A huge chunk of metal used to seperate rooms."
 	icon = 'icons/turf/walls/wall.dmi'
 	icon_state = "wall"
-	var/rotting = 0
+	var/rotting = FALSE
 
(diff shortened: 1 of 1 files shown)


Dumbdumn5:
Though I _do_ kind of like this, it should probably be reflected in the wiki if we do add this, especially since it may be confusing for newer players if they want to try a little B&E.

Kyep (approved):

AffectedArc07 (approved):

PR #15573, opened 2021-02-21
Removes the slipping effect from gluon grenades

SabreML:
## What Does This PR Do

Removes the AOE slipping effect from nukie gluon grenades.
All other effects of the grenade are unchanged, so it still irradiates and does stamina damage.

## Why It's Good For The Game

I'll admit that this is almost entirely an *"I'm dead pls nerf"* PR, but considering that the slipping actually ends up harming nukies more than helping, I definitely consider it worth changing.
For clarification: Noslips and Magboots are both useless against the 'Permafrost' tiles, you'll still slip either way.

## Changelog
:cl:
tweak: Removes the permafrost effect from nukie gluon grenades.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/atmosgrenade.dm
+++ b/code/game/objects/items/weapons/grenades/atmosgrenade.dm
@@ -31,7 +31,7 @@
 	spawn_amount = 500
 
 /obj/item/grenade/gluon
-	desc = "An advanced grenade that releases a harmful stream of gluons inducing radiation in those nearby. These gluon streams will also make victims feel exhausted, and induce shivering. This extreme coldness will also wet any nearby floors."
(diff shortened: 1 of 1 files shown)


Denthamos:
Its...ice.  Its entire main purpose is to slip.

hal9000PR:
I think it would be just better to make no slips work on it

Mitchs98:
Speaking from personal experience. There are very few things in the world that protect against slipping on ice, slip resistant shoes are rarely one of them. But it does make more sense than non-slip ice.

SabreML:
> Its...ice. Its entire main purpose is to slip.

Fair point, but I don't consider it 'fun' for either side if there's no way to avoid slipping. At least with 

> I think it would be just better to make no slips work on it

I'm pretty sure that there would be no way to do that without refactoring the whole system.

Anyway, was in an incredibly bad mood when I posted this and didn't really think it through, so I'll just close this now since it's clearly not gonna happen.

PR #15574, opened 2021-02-21
Makes Nukies have their own type of access, which other Agent ID cards do not

Kepplerx:
DISCLAIMER: THIS PR NEEDS A MAP CHANGE AS WELL (I will just check if this PR is wanted or not, before I will try and attempt mapping, which would also leave this PR in the map queue)

## What Does This PR Do
Makes nukies have their own specific ID access, allowing doors to be locked to nukies while there are other Agent IDs. This also renames a Id type so as to make it more clear what it will be used for.

## Why It's Good For The Game
This PR would make it impossible for the Syndie depot Agent ID to open the shuttle door for nukies, as well as normal Syndies who bought the Agent ID from opening the shuttle doors. The "new" type of access would as such protect nukies a smidge more, especially if combined with a remap of their shuttle to a less breakable one.

## Changelog
:cl:
tweak: Changed Nukie Leader access to Nukie Access, making this the new ID needed for the nukie shuttle (I will do this if the PR is wanted, as this would need a remap)
/:cl:


Kepplerx:
The reason I am renaming the ID access, is because if I do this mapchange, that would be the effect, and there would be little reason to call it "Leader" if it does not only apply to the leader, most of the PR would be remapping, but this is what I can do without that, respond that you want this PR, and I will do the remapping (Seems like a lot of work to try and learn if it isnt even wanted, that is why)

datlo:
Why is this needed? Nearly all nuke ops rounds will not have syndicate agents on the station. And, as far as I know, the syndi depot is on lockdown for nuke ops rounds (Should be confirmed as I am not certain of this). Was there a specific situation in a round where this was an issue?

Carthusia:
I dont recall ever seeing regular agents on a nuke op round.

I actually dont think its possible, but I may be wrong there.

AffectedArc07:
Really not seeing the need for this

Kepplerx:
> Really not seeing the need for this

The situation where this would have had an effect, was a speedrun raid of the Syndie Outpost, which then gave them an ID to use at the Nukie shuttle.

Cocacolagua:
considering that you can just deconstruct the shuttle walls instead of launching yourself to space and hoping to win the space ruin gamble in less than the half hour these rounds last, there's not much need to this

Fox-McCloud:
If a single feature is causing this much of an issue, then it's likely not the rest of the game that has an issue with its overall scope or design.

Creating an entire new access type to resolve this *one* thing, thereby require a ton of map edits and other stuff is not an acceptable solution.

PR #15575, opened 2021-02-21
Pickup and Dropped Bugfixing

Fox-McCloud:
Alters `pickup` behavior to be what it used to be, once again.

- Alters the bomb guardian to instead be reliant on `attackhand` instead of `pickup`.

- Ensures that both `pickup` and `dropped` call their parents 📞

- If an object fails to be put in your hand, it'll call `dropped` on the object
  - There's currently edge cases where you can fail to pick up and object, gaining it the `in_inventory = TRUE` status, despite it not being in your inventory.

Also fixed deathnettles knocking you unconscious when you picked them up---I'm guessing this was the result of me using the same proc TG does, when I ported it---their `paralyze` is not the same as hours.

:cl: Fox McCloud
fix: Fixes some edge cases of dropped items being registered in someone's inventory when they actually were not
fix: Fixes deathnettles knocking you unconscious by picking them up
/:cl:


dearmochi (approved):

S34NW:
Would #15576 fall under the scope of this PR?

AffectedArc07 (approved):

PR #15577, opened 2021-02-21
Makes gluon grenades use ice instead of permafrost

Qwertytoforty:
## What Does This PR Do

Alternative to the now closed, #15573 

This PR makes gluon grenades make ice turfs, instead of permafrost turfs. This means no slips and magboots can walk accross ice without slipping, but people without magboots or no slips will slip, even on walk intent. This fixes the issue of nukies slipping on their own bomb.

## Why It's Good For The Game

A bomb that hurts crew without hurting nukies is good, instead of just slipping everyone, even with no slips.

## Changelog
:cl:
tweak: Gluon grenades now ice turfs instead of permafrost, can be crossed safely with magboots or no slips.
/:cl:


code/game/objects/items/weapons/grenades/atmosgrenade.dm
```diff
@@ -47,7 +47,7 @@
 	for(var/turf/T in view(freeze_range, loc))
 		if(isfloorturf(T))
 			var/turf/simulated/F = T
-			F.MakeSlippery(TURF_WET_PERMAFROST)
+			F.MakeSlippery(TURF_WET_ICE)
```
Fox-McCloud, line 50: This code isn't great.

It loops through all turfs in range, but then does an `isfloorturf` check on them.

This should definitely just loop through floor turfs, only, to begin with.

Qwertytoforty: It now loops through only floor turfs

Kyep (approved):

AffectedArc07 (approved):

PR #15580, opened 2021-02-21
Makes the SM drop anomaly cores

hal9000PR:
## What Does This PR Do
Anomalies spawned from the SM now drop anomaly cores

## Why It's Good For The Game
Firstly, let me say that this will NOT mean crew can farm endless phazons, The SM doesn't make bluespace anomalies.

Anomaly cores dropping from the SM seems to have been intended, it was mentioned in #13546. 

Giving engineering more things to do and more incentive to do them is a welcome change, means that engineering once they have built the DNA vault/station shields in 10 minutes they still can do something. There aren't always breaches and making bars does get old after a while. adding more fun things that engineering can do to help out science and the station I think is a positive.

## Changelog
:cl:
Tweak: anomalies from the SM now drop cores
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -819,12 +819,12 @@
 	if(L)
 		switch(type)
 			if(FLUX_ANOMALY)
-				var/obj/effect/anomaly/flux/A = new(L, 300, FALSE)
+				var/obj/effect/anomaly/flux/A = new(L, 300)
 				A.explosive = FALSE
 			if(GRAVITATIONAL_ANOMALY)
-				new /obj/effect/anomaly/grav(L, 250, FALSE)
+				new /obj/effect/anomaly/grav(L, 250)
 			if(PYRO_ANOMALY)
-				new /obj/effect/anomaly/pyro(L, 200, FALSE)
+				new /obj/effect/anomaly/pyro(L, 200)
 
 /obj/machinery/power/supermatter_crystal/proc/supermatter_zap(atom/zapstart = src, range = 5, zap_str = 4000, zap_flags = ZAP_SUPERMATTER_FLAGS, list/targets_hit = list())
 	if(QDELETED(zapstart))


Fox-McCloud:
I specifically removed this, as things changed between back then and now; the entire anomaly core system that was being prototyped, back then, was changed to revolve around cargo and the events, instead of the SM----the anomalies for the SM are meant to be more of a failure state and add serious challenge to fixing a delaming SM, as opposed to "haha, you can farm it!".

And for that reason 👎

hal9000PR:
> I specifically removed this, as things changed between back then and now; the entire anomaly core system that was being prototyped, back then, was changed to revolve around cargo and the events, instead of the SM----the anomalies for the SM are meant to be more of a failure state and add serious challenge to fixing a delaming SM, as opposed to "haha, you can farm it!".
> 
> And for that reason 👎

well, when the SM is delamming is not the only time it creates anomalies. you can maintain a stable 5000+ EER SM that produces anomalies without it delamming. So it isn't exactly a failure state. Also farming them does add a challenge as  you have to walk up to the anomalies scan them, then send the signal, all in a hostile in environment surrounded by anomalies. I severly doubt anyone is going to try and collapse anomalies off a delamming SM, instead they will use CO2 to make the EER go above 5k to make anomalies because many of the people i have talked to didn't know the SM anomalies didn't make cores, meaning people haven't tried to do it anyway.

Kyep (commented):
You do realize this will lead to people intentionally destabilizing the SM to get cores, right?

S34NW:
Giving engineering a difficult but optional challenge (running the engine "hot" but safe) seems fine. If a newer engineering team just wants to set it up and leave it, they can. But those with more advanced knowledge can do something interesting and be rewarded for it beyond "haha big number engine"

dearmochi (approved):
Code's fine, can't speak on design

SabreML:
Currently there's no real reason to ever go above ~5K EER, which is easily achievable with CO2 and a few minutes of fine tuning. Sure you can get more power out of it *(I think)* but there's far easier ways to do that than destroying half of engineering via anomalies.

This will, if nothing else, incentivise people to experiment with fancier gas mixes or cooling systems in order to get a sustainable level of anomalies.
And while it'll obviously lead to more delaminations, I personally consider that an acceptable cost for increasing the fun/skill ceiling of the SM engine.

hal9000PR:
> You do realize this will lead to people intentionally destabilizing the SM to get cores, right?

you can run a perfectly stable engine that makes anomalies. An engine with an EER above 5K isn't innately unstable. And if you are referring to people decreasing the integrity to make anomalies, then this won't happen at all and if someone is foolish enough to do that then that should ideally be handled administratively. 

Currently, I have been able to set up stable 7K EER SMs, just for the fun of it. It would be nice if doing this actually had a purpose because the power it outputs is pitiful compared to a TEG. I think engineering giving anomaly cores to the station is a fun side project that engineering could do on slow shifts.

PR #13979, opened 2020-08-01
Adds Detective access to the Escape Shuttle Security doors.

SabreML:
## What Does This PR Do

Adds Detective access to the Security and Cockpit Foyer doors on the Escape Shuttle.

## Why It's Good For The Game

The Detective should be able to sit ~~away from the peasants~~ with the rest of Security on the way to Centcom.

## Images of changes

Old:
image

New:
image
Please note that this is only adding Detective access to these areas.

## Changelog
:cl:
add: Adds Detective access to the Escape Shuttle Security areas.
/:cl:

PR #15584, opened 2021-02-22
Changes the Escape Shuttle Security doors to basic Security access

SabreML:
## What Does This PR Do

**[Remake of #13979, with added standardisation!]**
Changes the required access level on all shuttle brig doors and the departures holding cell from `2`*(Equipment Storage)* to `63`*(Sec front doors)*.
This also effects the 'foyer' doors, but all heads have basic sec access anyway so it shouldn't make any difference for them.

## Why It's Good For The Game

Security/Support staff should be allowed to sit ~~away from the rabble~~ with the rest of their department on the way to Centcomm.

### Before:
***(Green marks doors that can be accessed by these roles, red is inaccessable.)***
Old

### After:
New

**(For clarification, changed doors are marked in cyan:)**
image

## Changelog
:cl:
tweak: Changed the access on all escape shuttle brig doors to be basic sec access, rather than equipment access. (Detective, Blueshield, IAA, etc. have access now.)
/:cl:


AffectedArc07 (changes requested):
image
Making these security only is an absolute no from me.
Although, it appears you didnt do that in your diff. Could you make your screenshots clearer?

SabreML:
Sorry, I pretty much just copied them from the old PR. Green is the doors that the Detective/BS/IAA can access, from their point of view.

AffectedArc07 (changes requested):
As per map standardisation, please remake these changes on the other 2 maps.

SabreML:
Meta doesn't have a direct route from escape to the Security section, so no changes are needed there.

AffectedArc07 (approved):
Approving now, obviously waiting on mapping queue

AffectedArc07 (approved):

PR #15592, opened 2021-02-22
Prevents carps ranged protection from working with his grace

Qwertytoforty:
## What Does This PR Do

This PR adds a check to the martial arts reflection to check if the user has his grace's disease, and will prevent ranged deflections if the user has the disease.

## Why It's Good For The Game

Currently, we have precedent for ranged immunity and antistun abilities not stacking. For example, hulk with carp, or hulk with a dsword, both are prevented, so they are possible for crew to deal with. Carp and his grace, currently work together. This is worse than hulk, as it has healing, a toolbox that can potentially instantly kill you, and the user will not loose the effect via mutadone.

Thus, following precedent, and balance concerns, you will no longer reflect bullets with carp with his grace active, although you can still punch them and disarm them for extra damage.

## Changelog
:cl:
tweak: Carps ranged immunity no longer works if the user has His Grace
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -37,7 +37,7 @@ emp_act
 		return 2
 
 	if(mind?.martial_art?.deflection_chance) //Some martial arts users can deflect projectiles!
-		if(!lying && !HAS_TRAIT(src, TRAIT_HULK) && prob(mind.martial_art.deflection_chance)) //But only if they're not lying down, and hulks can't do it
+		if(!lying && !HAS_TRAIT(src, TRAIT_HULK) && !HasDisease(new /datum/disease/memetic_madness(0)) && prob(mind.martial_art.deflection_chance)) //But only if they're not lying down, and hulks can't do it
 			add_attack_logs(P.firer, src, "hit by [P.type] but got deflected by martial arts '[mind.martial_art]'")
 			visible_message("<span class='danger'>[src] deflects the projectile; [p_they()] can't be hit with ranged weapons!</span>", "<span class='userdanger'>You deflect the projectile!</span>")
 			return FALSE


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -37,7 +37,7 @@ emp_act
 		return 2
 
 	if(mind?.martial_art?.deflection_chance) //Some martial arts users can deflect projectiles!
-		if(!lying && !HAS_TRAIT(src, TRAIT_HULK) && prob(mind.martial_art.deflection_chance)) //But only if they're not lying down, and hulks can't do it
+		if(!lying && !HAS_TRAIT(src, TRAIT_HULK) && !HasDisease(new /datum/disease/memetic_madness(0)) && prob(mind.martial_art.deflection_chance)) //But only if they're not lying down, and hulks can't do it
```
SteelSlayer, line 40: My god, screw whoever wrote the `HasDisease()` code and made it so it takes an instance of a disease instead of just a type.

Fox-McCloud:
This is so hyper specific, I'm kinda of the opinion "who cares the 1 in 100 rounds this happens".

You have to be:

- Chaplain
- Hijack
- Get TC from someone else

And His Grace is still probably going to eat your face.

hal9000PR:
> This is so hyper specific, I'm kinda of the opinion "who cares the 1 in 100 rounds this happens".
> 
> You have to be:
> 
> * Chaplain
> * Hijack
> * Get TC from someone else
> 
> And His Grace is still probably going to eat your face.

The one way to fight his grace is with shotguns, bursting them down before they can heal up. Having carp and his grace makes it practically unkillable outside of stuff like bombs and high damage make weapons which sec doesn't get. So to beat this sec has to either ahelp to make a bomb, or hope a miner has a spectral sword or similar. This really isn't good balance. Also just because it doesn't happen often doesn't mean it is fun to fight against. A rare issue coming up is still an issue. 

Ideally this would extend to all bulky items, preventing things like carp saw and such. Complete range immunity is too strong when combined with strong weapons and antistun. When we have to make so many exceptions for sleeping carp ranged immunity, maybe it would be better to rework it.

Also you can get the required 37tc by one traitor now because of the addition of contractor so combos like this are now more common.

procdrone:
Is this even a problem? Like Fox said, its incredibly unlikely to happen, and sort of goes to "you managed to pull off the impossible, you deserve your OP combo".

It was the same discussion with your grace, people complaining how powerful it was, but the argument always was "it almost never happens, one in thousand chaos round".

Butlerblock:
I’ve personally played and constantly observed this game for over 5 years and have only seen carp grace twice. Definitely not nearly often enough to be an issue and should be kept because of the meme value.

hal9000PR:
> I’ve personally played and constantly observed this game for over 5 years and have only seen carp grace twice. Definitely not nearly often enough to be an issue and should be kept because of the meme value.

His grace was added just under 4 years ago. Saying that you've only seen it twice in 5 years is kinda silly considering both of those have been in the past three months. Recent changes such as highlighting keywords means more traitors working together and more pooling TC and contractors meaning traitors can have more than 20tc each. Personally i have seen this 4 times in 3 months.

Kyep (approved):

variableundefined:
No maintainer approval, 120+ days. 

Closing.

PR #15599, opened 2021-02-23
Fixes the flyswatter not swatting flies

SabreML:
## What Does This PR Do

Fixes the flyswatter not actually swatting bugs properly, since `afterattack()` never got called.
Also made syndibees immune to the new squishing mechanic, since they're fancy Syndicate bees.

## Why It's Good For The Game

Clearly intended to be a feature, considering that all the code is there.

## Changelog
:cl:
fix: Fixed the flyswatter not killing bugs properly.
/:cl:


Fox-McCloud:
I'd recommend finding out why the `afterattack` can't run.

SabreML:
> I'd recommend finding out why the `afterattack` can't run.

That was the first thing I looked in to, it's caused by this returning `TRUE`:
https://github.com/ParadiseSS13/Paradise/blob/5f69256a66502d14e2003a3e25f8b221244e1ea2/code/_onclick/item_attack.dm#L107-L118
Which gets passed all the way down the chain into this:
https://github.com/ParadiseSS13/Paradise/blob/5f69256a66502d14e2003a3e25f8b221244e1ea2/code/_onclick/item_attack.dm#L1-L6

And since `resolved` was `TRUE`, it doesn't call `afterattack()`.
Obviously I could try making `/mob/living/attacked_by()` return `FALSE` instead, but I figured that could have some unintended side effects, and would probably end up needing a refactor of some sort.

variableundefined (approved):

PR #15600, opened 2021-02-24
Fixes reverb being added to Lobby Music, Bwoinks, etc.

SabreML:
## What Does This PR Do

Adds the `SEND_SOUND_NO_ECHO()` define in order to remove the reverb from some things that really shouldn't have it, like lobby music, adminhelps/bwoinking, ambient music, etc.

## Why It's Good For The Game

As amusing as it may be to hear reverb on a bwoink sound, it really shouldn't happen. If only because it makes it sound like an ingame thing rather than being entirely OOC.

## Changelog
:cl:
fix: Fixed reverb being added to Lobby music, Round end sounds, Ambient sounds, Antag greetings, Admin/Mentorhelps, Bwoinking, and a few other things.
/:cl:

### DIFF
--- a/code/__HELPERS/_logging.dm
+++ b/code/__HELPERS/_logging.dm
@@ -5,11 +5,20 @@
 GLOBAL_VAR_INIT(log_end, (world.system_type == UNIX ? ascii2text(13) : ""))
 GLOBAL_PROTECT(log_end)
 
+// why are these all in _logging.dm
 #define DIRECT_OUTPUT(A, B) A << B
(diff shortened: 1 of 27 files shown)


SabreML:
Pinging @Fox-McCloud for a review since they added the whole sound environment system in #15378.

Fox-McCloud:
Not the most ideal way to fixing this, unfortunately---also, a few of these things already don't have echo and aren't in need of changing.

Please see: https://github.com/ParadiseSS13/Paradise/pull/15602 for details.

PR #15603, opened 2021-02-24
Reverb preference

AffectedArc07:
## What Does This PR Do
Adds a preference to enable/disable the reverb effects

## Why It's Good For The Game
Some people dislike this. Even if they are objectively wrong, we should cater to them.

## Changelog
:cl: AffectedArc07
add: Added a verb to enable/disable the new sound reverb effects.
/:cl:

### DIFF
--- a/code/__DEFINES/preferences.dm
+++ b/code/__DEFINES/preferences.dm
@@ -52,8 +52,10 @@
 #define PREFTOGGLE_2_RUNECHAT		64
 #define PREFTOGGLE_2_DEATHMESSAGE	128
 #define PREFTOGGLE_2_EMOTE_BUBBLE	256
+// Yes I know this being an "enable to disable" is misleading, but it avoids having to tweak all existing pref entries
+#define PREFTOGGLE_2_REVERB_DISABLE	512
 
-#define TOGGLES_2_TOTAL 			511 // If you add or remove a preference toggle above, make sure you update this define with the total value of the toggles combined.
+#define TOGGLES_2_TOTAL 			1023 // If you add or remove a preference toggle above, make sure you update this define with the total value of the toggles combined.
 
 #define TOGGLES_2_DEFAULT (PREFTOGGLE_2_FANCYUI|PREFTOGGLE_2_ITEMATTACK|PREFTOGGLE_2_WINDOWFLASHING|PREFTOGGLE_2_RUNECHAT|PREFTOGGLE_2_DEATHMESSAGE|PREFTOGGLE_2_EMOTE_BUBBLE)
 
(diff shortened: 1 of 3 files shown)


SabreML:
> Even if they are objectively wrong, we should cater to them.

👀

Spacemanspark:
Gottem

code/modules/client/preference/preferences_toggles.dm
```diff
@@ -331,3 +331,11 @@
 	prefs.toggles2 ^= PREFTOGGLE_2_DEATHMESSAGE
 	prefs.save_preferences(src)
 	to_chat(src, "You will [(prefs.toggles2 & PREFTOGGLE_2_DEATHMESSAGE) ? "now" : "no longer"] see a notification in deadchat when a player dies.")
+
+/client/verb/toggle_reverb()
```
Fox-McCloud, line 335: Make absolutely certain this is on, by default, if you have a database.

I don't want the default experience to be "oh, reverb is off, you have to turn it on to get the benefits"--I want the fact we have reverb to be a known thing on para.

We have this issue with some instrument and other prefs---people don't know they exist until they're informed they can turn it on.

AffectedArc07: It is on by default. I have verified.

Fox-McCloud: 👍

code/__DEFINES/preferences.dm
```diff
@@ -52,8 +52,10 @@
 #define PREFTOGGLE_2_RUNECHAT		64
 #define PREFTOGGLE_2_DEATHMESSAGE	128
 #define PREFTOGGLE_2_EMOTE_BUBBLE	256
+// Yes I know this being an "enable to disable" is misleading, but it avoids having to tweak all existing pref entries
+#define PREFTOGGLE_2_REVERB_DISABLE	512
```
farie82, line 56: Wouldn't making 1 simple SQL script fix this. And just adding it to the `TOGGLES_2_DEFAULT` value.
Would remove any confusion regarding this define as well. 
A simple `UPDATE preferences_table SET flag=flag+512;` would do (ofc with the correct name etc)

AffectedArc07: I could do that the same as I did for runechat, this is just less error prone

Aurorablade:
and they are infact objectively wrong but they are allowed.

code/__DEFINES/preferences.dm
farie82: Id say this is more error prone due to the double negative you have to do.
Is reverb not not enabled?
If so do reverb

Kyep (approved):

PR #15608, opened 2021-02-25
Broad TG Sound Update

Fox-McCloud:
Ports two major sound updates from TG:

- https://github.com/tgstation/tgstation/pull/54362 by @Qustinnus 
- https://github.com/tgstation/tgstation/pull/53911

Our sound system is a derivative of Bay, but the sound falloff is still very old and just relies on having the sound range be a massive `(world.view + extrarange) * 3`. This has really wonky implications in that you can hear people up in security screaming---clear down in medical. 

With our high population, this effectively makes doing anything stealth related difficult as they're just going to *immediately* scream---or the beating sound is likely to be heard, as well.  

Either case, this is in between; it has a better way of handling falloff so it's less of a harsh cuttoff while not having the *insanely* long range audio that we have before; I've found this also makes sounds far less "muddy" than they previously were.

This will set a better basis for implementing more footsteps or sounds for turfs when walking across them; currently with how insanely far sounds travel, floor stepping sounds being impactful just isn't a great thing because, again, it makes stealth impossible. With a system like this, it lays groundwork for that more immersive experience, without people in medical knowing someone is in southeast sci maint because they have their volume cranked all the way up.

This does tweak a few sound volumes that were out of line with TG; I figured it was best to have their volume levels as a baseline with this new system, since they've already had this system a while.

The second PR is a dramatic update to explosions and their sounds; the sounds you hear are based on how far away from the explosion you are---likewise, someone setting off a tiny light range `1` explosion will no longer trigger a z-level wide "distant boom" sound---because why should it? It's a firecracker.

This also makes explosions shake your screen when they happen as well---overall, it has an extremely immersive effect---see here:

https://streamable.com/yel0od

Credit to @Gandalf2k15 for the video

:cl: Fox McCloud
tweak: sound system falloff and distances completely overhauled--sounds should be clearer and more consistent in volume--overall travel distance should be a bit less, but still sizeable
tweak: updates explosion sounds and adds explosion shaking for *immersive explosion experiences*
/:cl:


Fox-McCloud:
Note to self, jesterone needs `, falloff_exponent = 20` added to it. I'll do that tomorrow.

Aracino:
Yes. All of my yes. Still need to know how much this applies and how far. Notably, if someone was kill holding someone in one of the dorm room, would the sound of their gasping be heard all the way in processing? One issue I had lately, where I was attempting to assassinate someone quietly in the private rooms in sec mait, and then suddenly 2 officers ran from sec to start welding the wall down because they heard their gasping through several walls.

Fox-McCloud:
@Aracino 

It'll still travel a decent ways, so someone in the area will still hear what's going on---again, this is a compromise between two extremes: just hearing things that are a bit further than your screen range and the incredibly long range sounds we have now.

You'll still have to be careful, but it won't reach *ridiculous* levels, either.

Qustinnus:
You can just change the sound range of specific sounds like gasping, I added defines for it

Fox-McCloud:
@Qustinnus 

Did you ever have plans of having dynamic hearing ranges for mobs, based on if they have ears and their sensitivity? Would be neat to have mobs that are extra sensitive to sounds (like flashbangs), but they can hear further than regular mobs (I know ear damage sensitivity is already a thing).

Aracino:
> You can just change the sound range of specific sounds like gasping, I added defines for it

I like this. Gasping should have a very short range, especially given the gasping sound effect didn't used to be a thing. It makes being stealth exceptionally harder, as the only antags I'm aware can stop emote sounds like gasping, are clings and slings given they have inbuilt mutes. Tape really doesn't work due to the ease of resisting out of it, so normal traitors need chems in order to do it, which I'm still pretty sure doesn't mute emotes like gasping, given I've fed someone a mimana and their gasping/followup death gasp were still audible.

DeveloperLoser:
>likewise, someone setting off a tiny light range 1 explosion will no longer trigger a z-level wide "distant boom" sound--->because why should it? It's a firecracker.

Aww I cant troll skyping with my fake tesla engines anymore XD

AffectedArc07 (approved):

PR #15610, opened 2021-02-25
Disaster Counter tweaks

SabreML:
## What Does This PR Do

Makes Disaster Counters reset their current record after any explosion, no matter the severity.
Also makes the maptext update with the counter, and clamped the visual counter to a minimum of 0, since -1 doesn't make much sense.
*(Requested by @AffectedArc07)*

## Why It's Good For The Game

Better visually since the numbers actually update, and it'll be a lot more reliable for detecting explosions.

## Changelog
:cl:
tweak: Disaster Counters are now reset by all explosion severities.
fix: Fixed Disaster Counters showing -1 in their examine text.
/:cl:


code/game/objects/structures/disaster_counter.dm
```diff
@@ -30,16 +30,19 @@
 	SSpersistent_data.register(src)
 
 /obj/structure/disaster_counter/ex_act(severity)
-	if(severity < 2)
-		current_count = -1
-		persistent_save()
 	. = ..()
+	current_count = -1
+	persistent_save()
+	update_maptext()
```
AffectedArc07, line 36:
```suggestion
	current_count = -1
	persistent_save()
	update_maptext()
	. = ..()
```

Do it this way round incase it qdels in `..()`

Kyep (approved):

code/game/objects/structures/disaster_counter.dm
```diff
@@ -30,16 +31,19 @@
...
 	. = ..()
 
 /obj/structure/disaster_counter/Destroy()
 	if(counter_id)
 		SSpersistent_data.registered_atoms -= src // Take us out the list
 	return ..()
 
+/obj/structure/disaster_counter/proc/update_maptext()
+	maptext = "<span class='maptext' style='text-align: center'>[max(current_count, 0)] / [record_count]</span>"
```
AffectedArc07, line 45:
```suggestion
	maptext = "<span class='maptext' style='text-align: center'>[max(current_count, 0)]/[record_count]</span>"
```
This spacing has an ingame effect. Its a string, not a mathematical operation.

AffectedArc07 (changes requested):

code/game/objects/structures/disaster_counter.dm
SabreML: Yeah I realise, it just looked better to me with spacing between the numbers.
NSS Cyberiad 2021-03-01 200435
NSS Cyberiad 2021-03-01 200423

AffectedArc07: Try bumping the values to bigger numbers like 42 on each side and see how it looks

SabreML: NSS Cyberiad 2021-03-01 203009

It only really starts clipping out of the screen when it goes above five digits, but fair point.

AffectedArc07 (approved):

PR #15613, opened 2021-02-25
Ports Ambience Subsystem

Fox-McCloud:
Ports TG's @Qustinnus ambience subsystem: https://github.com/tgstation/tgstation/pull/56723

Yes, I am on a sound updating binge of late.

Instead of ambience being based upon activation when you transition from one area to another, it's, instead, activated, periodically, when you're in an area. 

This allows you to hear some sounds more often (since you rarely go in/out a particular area) and makes for a more immersive setting---you'll hear the ambient sounds of maintenance while you're in it...as opposed to just when you first step in (or cross from one ambient area to another).

:cl: Fox McCloud
add: Ambience now plays periodically when you're in an area, as opposed to when you first step into it
/:cl:


code/game/area/areas.dm
```diff
@@ -476,27 +481,13 @@
 	if((oldarea.has_gravity == 0) && (newarea.has_gravity == 1) && (L.m_intent == MOVE_INTENT_RUN)) // Being ready when you change areas gives you a chance to avoid falling all together.
 		thunk(L)
 
-	// Ambience goes down here -- make sure to list each area seperately for ease of adding things in later, thanks! Note: areas adjacent to each other should have the same sounds to prevent cutoff when possible.- LastyScratch
-	if(L && L.client && !L.client.ambience_playing && (L.client.prefs.sound & SOUND_BUZZ))	//split off the white noise from the rest of the ambience because of annoyance complaints - Kluys
+	//Ship ambience just loops if turned on.
+	if(L && L.client && !L.client.ambience_playing && (L.client.prefs.sound & SOUND_BUZZ))
 		L.client.ambience_playing = TRUE
```
Fox-McCloud, line 486: @Qustinnus 

You got rid of this, and I get why, but it causes a subtle cracking/popping noise  when you cross from one area to another---I suspect 99% of people won't notice it or just think it's an audio glitch---but I pick up on stuff like this and it drives me crazy.

Also, separate ship white-noise from ambient sounds on TG when? 😛

Fox-McCloud:
Tested and still works great with @dearmochi's volume panel.

Qustinnus:
something i didn't bother fixing is that this makes ambience play in the lobby

code/game/area/areas.dm
Qustinnus: i'll have a look, and yeah tbf good point i should make both work at once

AffectedArc07 (approved):

dearmochi (approved):

PR #15616, opened 2021-02-26
Implements PDA Beeping

Fox-McCloud:
Ports the PDA beeping noises from TG.

PDA code is awfu---just...awful.

Either case, this adds various sounds to PDAs:

- when you insert/eject ID's
- when you insert/eject pens
- When you navigate around the screen it'll play a beep
- When you authenticate your ID
- When you successfully send a message
- When a message fails to send

Overall just makes using a PDA feel more like using a late 90's retro device while also making for a more immersive environment.

No, this doesn't apply to using the syndicate uplink, just to be clear--that's silent (and completely separate of PDA code).

Will leave a few comments in the code below....PDA code man...PDA code.

:cl: Fox McCloud
add: PDA's make various noises when using them
/:cl:


code/modules/pda/PDA.dm
```diff
@@ -31,6 +31,7 @@ GLOBAL_LIST_EMPTY(PDAs)
 	var/datum/data/pda/utility/scanmode/scanmode = null
 
 	var/lock_code = "" // Lockcode to unlock uplink
+	var/silent = FALSE //To beep or not to beep, that is the question
```
Fox-McCloud, line 34: I moved silent to the PDA level because fetching the messenger program every time you need to modify this is ridiculous---also, frankly, this is a case of making PDA's *too* modular; having the messenger silence on the program itself, and only the program means we can't utilize it for other stuff, like this.

code/modules/pda/app.dm
```diff
@@ -67,7 +66,9 @@
 	if(pda.current_app)
 		pda.current_app.stop()
 	pda.current_app = src
-	return 1
+	if(!pda.silent)
```
Fox-McCloud, line 69: the reason this has to be boilerplated so much is because some apps create sounds, and we don't want sounds to overlap between the beeping and say, the honk sound the clown plays---there's also some instances where we do/do not want the sound to play within the app, again, to avoid overlap (sending messages, for instance).

the app code is not very atomic or OOP---and I'm not about to refeactor the entirety of it just to make playing sounds a bit cleaner.

dearmochi (approved):

AffectedArc07 (approved):

PR #15618, opened 2021-02-26
SSU now reject nodrop items

### LINKED ISSUE #15596: SSU Storage Slot Borg Interaction Oddity
**Issue Description**:
While attempting to destroy an open SSU (explorer suit variant in the mining outpost) as an engineering borg with an integrated welder a copy of it was placed inside the ssu's storage slot which I was then able to eject (https://cdn.discordapp.com/attachments/811073644771016705/813779568124887040/ssustorageoddness.png). The copied welder mimicked the state of the original and disappeared when the original was stowed (original was able to be used throughout the rest of the shift without issue).

The ssu to take damage from the welder or to receive a variant of the "X is attached to you and you cannot place it in Y" message.

**What happened instead**:
An odd pseudo duplicate of the welder was placed within the SSU.

Assumedly unintentional behaviour/possible item duplication.

As a borg use a welder (or possibly any other item) on an open ssu with an empty storage slot.

I'd assume it's an unintended side effect of https://github.com/ParadiseSS13/Paradise/pull/15540 which allowed any small items to be placed in an ssu's storage slot but I'm not certain (I'm very much a fan of improvements from the pr and hope the issue can be addressed without any loss of functionality). 

**Extra information**:
N/A (I am however extremely tired while typing this (but figured I shouldn't delay reporting it) so I apologise for any mistakes, general poor grammar, poor choice of title or anything I've missed (as a semi related aside, a few months ago as a humanoid I picked up a borg variant wrench out of a linen basket which disappeared briefly after. At the time I wasn't sure what had happened but looking back I guess it might have been another variant of this issue)).

TDSSS:
## What Does This PR Do
You can no longer store nodrop items in an SSU, creating a phantom copy of them.
This fixes #15596  as well.
Note that the issue existed even before #15540, that PR just made it more obvious as more items could now possibly be stored in an SSU.

## Why It's Good For The Game
bugs bad

## Changelog
:cl:
fix: Nodrop items, including borg tools, no longer fit in SSU storage
/:cl:

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -331,28 +331,42 @@
 	if(default_deconstruction_screwdriver(user, "panel", "close", I))
 		I.play_tool_sound(user, I.tool_volume)
 
+/**
+  * Tries to store the item into whatever slot it can go, returns true if the item is stored successfully.
+  *
+**/
 /obj/machinery/suit_storage_unit/proc/store_item(obj/item/I, mob/user)
-	if(!user.canUnEquip(I))
-		return
-	. = FALSE
 	if(istype(I, /obj/item/clothing/suit) && !suit)
-		suit = I
-		. = TRUE
+		if(try_store_item(I, user))
+			suit = I
+			return TRUE
 	if(istype(I, /obj/item/clothing/head) && !helmet)
-		helmet = I
-		. = TRUE
+		if(try_store_item(I, user))
+			helmet = I
+			return TRUE
 	if(istype(I, /obj/item/clothing/mask) && !mask)
-		mask = I
-		. = TRUE
+		if(try_store_item(I, user))
+			mask = I
+			return TRUE
 	if(istype(I, /obj/item/clothing/shoes) && !boots)
-		boots = I
-		. = TRUE
-	if((istype(I, /obj/item/tank) || I.w_class <= WEIGHT_CLASS_SMALL) && !storage && !.)
-		storage = I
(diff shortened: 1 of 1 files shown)


code/game/machinery/suit_storage_unit.dm
```diff
@@ -332,6 +332,8 @@
 		I.play_tool_sound(user, I.tool_volume)
 
 /obj/machinery/suit_storage_unit/proc/store_item(obj/item/I, mob/user)
+	if(I.flags & NODROP)
```
Fox-McCloud, line 335: not the best way to handle it.

TDSSS: If I check for drop_item early, I need to put it in each case, since it needs to happen before the `suit = I` assignment but after the check if it goes in the right slot (since we want to be sure there's a free slot before we drop it from the hand)

What I've done instead is use the `canunequip` proc, which is the same proc `drop_item` uses internally to check if it can actually drop the item. 

I hope that's an acceptable solution.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -332,6 +332,8 @@
 		I.play_tool_sound(user, I.tool_volume)
 
 /obj/machinery/suit_storage_unit/proc/store_item(obj/item/I, mob/user)
+	if(!user.canUnEquip(I))
```
Fox-McCloud, line 335: still not the proper way to handle it. `drop_item` can already function, safely, with `NO_DROP` things.

PR #15619, opened 2021-02-26
Removes camera alarms

TDSSS:
## What Does This PR Do
This PR removes camera alarms that AIs and borgs used to get whenever a camera was destroyed or deactivated.

Edit: motion sensor alarms are still a thing and not impacted by this in any way.

## Why It's Good For The Game
Two issues we have that keep being brought up are 1) validhunting AIs that seem to think they're a part of sec and need to redtext every antag and 2) the station (especially maint) being too small, with very little hidden space to do hidden things.

As long as you are in a room with a camera, you could be watched at any moment. Any kind of sneaky move you might try pulling could be totally undone because the AI just happens to scroll by. A major argument against letting AIs have multi-window view was that it simply made the AI too omnipresent. So why, exactly, do we let the AI know exactly when one of its eyes gets destroyed or deactivated? Shouldn't we reward and enable the smart counter-play of taking out cameras beforehand, preparing your crime scene?

SS13 is usually a game that rewards being prepared and having a plan. But taking out cameras at the moment is a total trap. Nothing screams 'please come arrest me' as much as destroying a camera, and you can bet that a sec borg or three are on their way as soon as you snip a single camera wire.

By allowing antags to create 'camera-dead' zones within the station where they can potentially commit crimes, we take some 'pressure' out of maint as well and give them more options.

Cameras can also be a big issue for biohazards. They are naturally restricted to nesting or setting up in places without cameras, because if they destroy a camera they are sure to be found. Now, a blob, xeno or tspider could feasibly find a remote room, smash the camera and set up a nest there. The AI still has a chance to notice the camera is gone by the hole in its perception, but that requires the AI player to be actively monitoring the situation and moving around, instead of giving him a giant, glaring 'check this out' sign.

## Changelog
:cl:
del: Camera alarms have been removed.
/:cl:

### DIFF
--- a/code/controllers/subsystem/alarm.dm
+++ b/code/controllers/subsystem/alarm.dm
@@ -1,7 +1,7 @@
 SUBSYSTEM_DEF(alarm)
 	name = "Alarm"
 	flags = SS_NO_INIT | SS_NO_FIRE
-	var/list/alarms = list("Motion" = list(), "Fire" = list(), "Atmosphere" = list(), "Power" = list(), "Camera" = list(), "Burglar" = list())
+	var/list/alarms = list("Motion" = list(), "Fire" = list(), "Atmosphere" = list(), "Power" = list(), "Burglar" = list())
 
 /datum/controller/subsystem/alarm/proc/triggerAlarm(class, area/A, list/O, obj/alarmsource)
 	var/list/L = alarms[class]
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -66,7 +66,6 @@
 	if(istype(A))
 		A.motioncameras -= src
 	area_motion = null
-	cancelCameraAlarm()
 	cancelAlarm()
 	return ..()
 
@@ -86,7 +85,6 @@
 			var/thisemp = emped //Take note of which EMP this proc is for
 			spawn(900)
 				if(!QDELETED(src))
-					triggerCameraAlarm() //camera alarm triggers even if multiple EMPs are in effect.
 					if(emped == thisemp) //Only fix it if the camera hasn't been EMP'd again
 						network = previous_network
 						stat &= ~EMPED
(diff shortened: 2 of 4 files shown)


matttheficus:
motion sensors should be the only "warning" for someone in a room in my opinion anyway. If the AI wants to lock down a room camera-wise, the counter would be ANYONE entering sets off the alarm.

dearmochi (approved):

Qwertytoforty:
Slight note, looking at the code, it seems cutting wires on cameras do not set off the alarm, only mending them does. Doesnt change the impact on biohazards, however

Kyep (approved):

Fox-McCloud:
Notttt completely sure about this one.

As pointed out, before, those who can *cut* cameras, directly aren't terribly impacted by this, but those who overtly *destroy* cameras are. Sneaky antags are nearly always human or human based, where as Xenos, Demons, and spiders are a bit more...overt.

TDSSS:
True, I misread the code on when a cut camera sends a camera alarm. But in that case, does it make sense to send an alert when a camera reconnects? That seems rather pointless, doesn't it?

And imo, even biohazards have some things to gain from laying low at the start. Especially blobs can get easily rushed and destroyed if found immediately.

ppi13:
This should be a healthy buff to nearly every antag, I never touched cameras myself in years as an antag cause of the alarm.

TerryTehCoder:
Alternatively, super sneaky antags were always given the option to thwack the security camera with a multitool, screw up its focus,  and lower its vision range to 1-2 tiles. Which doesn't give an alarm when focused or defocused for the silicons. But I agree that this really doesn't affect sneaky antags that much, since cutting never actually caused a alarm.
A.I's find cameras down offline all the time, half the time it's not even antags, and just EMP's, explosions, or someone who thinks the A.I's malf/is paranoid.

variableundefined (approved):
Would be interesting to see how this play out.

Dafrek:
I thought this was the reason that cameras have a second wire that restricted the view range without setting off the alarm. 
image

PR #15626, opened 2021-02-27
Blood colouring in syringes

SabreML:
## What Does This PR Do

Fixes a typo in the reagent colouring code in order to allow different blood colours to show correctly in reagent containers, primarily syringes.
Also allowed the same thing for slime jelly colouring, since that uses a slightly different system.

## Why It's Good For The Game

Aside from the obvious consistency benefits, this not being a feature in the first place was most likely just a typo. `reagent.color` rather than `reagent_color`.

## Images of changes

**BEFORE/AFTER:**
Before Blood Colours

## Changelog
:cl:
tweak: Made the 'Blood' reagent show the species' blood colour. e.g. A syringe full of Skrell blood will be blue rather than red.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/colors.dm
+++ b/code/modules/reagents/chemistry/colors.dm
@@ -16,14 +16,16 @@ GLOBAL_LIST_INIT(random_color_list, list("#00aedb","#a200ff","#f47835","#d41243"
 	for(var/datum/reagent/reagent in reagent_list)
 		if(reagent.id == "blood" && reagent.data && reagent.data["blood_color"])
 			reagent_color = reagent.data["blood_color"]
+		else if(reagent.id == "slimejelly" && reagent.data && reagent.data["colour"])
+			reagent_color = reagent.data["colour"]
 		else
 			reagent_color = reagent.color
 
 		vol_temp = reagent.volume
 		vol_counter += vol_temp
 
-		if(isnull(color))
-			color = reagent.color
+		if(isnull(color)) // `color` will always be null so I'm not sure what these `else`s are for.
+			color = reagent_color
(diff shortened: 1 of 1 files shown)


Aracino:
You could actually fix this by injecting 5units, then taking blood again. But the blood working properly now without needing to fenangle it is a good change.

SabreML:
> You could actually fix this by injecting 5units, then taking blood again. But the blood working properly now without needing to fenangle it is a good change.

Yeah I noticed that when I was doing the 'Before' image, if it's over a certain threshold then it gets the correct colour.

code/modules/reagents/chemistry/colors.dm
```diff
@@ -16,14 +16,16 @@ GLOBAL_LIST_INIT(random_color_list, list("#00aedb","#a200ff","#f47835","#d41243"
...
+			reagent_color = reagent.data["colour"]
 		else
 			reagent_color = reagent.color
 
 		vol_temp = reagent.volume
 		vol_counter += vol_temp
 
-		if(isnull(color))
-			color = reagent.color
+		if(isnull(color)) // `color` will always be null so I'm not sure what these `else`s are for.
```
SteelSlayer, line 27: Regarding the comment: 
We're inside a for loop, so the `isnull` is essentially saying "do this thing on the first iteration only".

Kyep (approved):

code/modules/reagents/chemistry/colors.dm
SabreML: Ah, alright that makes sense.

AffectedArc07 (approved):

PR #15627, opened 2021-02-27
Adds Closet Open/Close Sounds

Fox-McCloud:
From TG.

Implement's custom closet open/closing sounds.

No longer do closets and crates play the same boring "click" noise--instead, lockers, creates, wooden crates, garbage cans, and carboard boxes all have their own unique open/close sounds that sound like, well...what you'd expect, as opposed to "generic click".

Removes the lock toggling sound from lockers for consistency--also doesn't make a ton of sense for some lockable crates/lockers.

:cl: Fox McCloud
add: adds in custom open/close sounds for lockers, crates, wooden lockers, garbage cans, and cardboard boxes
del: Lock toggling on lockers no longer plays a sound, for consistency
/:cl:


SabreML:
Personally I've never liked the locker sounds from TG, but I will admit that they're better than just having a click.
I'm not sure what the point is for removing locking/unlocking sounds though, I can't see any reason for that.

Fox-McCloud:
>I'm not sure what the point is for removing locking/unlocking sounds though, I can't see any reason for that.

It just doesn't fit well with a lot of the objects, now that they have different sounds; it's kinda jarring and weird to have the "CLICK" noise when the device is wooden or what have you---it's also inconsistent with crates, which also do not have a sound; practically everything that as a lock toggle, like this, also doesn't play a sound either.

SabreML:
Well that should be solved by adding the sound to crates, surely. Both metal and wooden lockers would presumably have a metal lock too, so I wouldn't say it's weird for them to click.

dearmochi (approved):

Spacemanspark:
I also don't see the need to remove the locking sounds. Otherwise, this is a good change.

Kyep (approved):

AffectedArc07 (approved):

PR #15628, opened 2021-02-27
Footstep Sound Overhaul

Fox-McCloud:
Ports TG's footstep sound system and components.

Now that our sound system is dramatically overhauled, this is a viable option to implement, now. Previously, with our former sound system, this would cause a lot of disruptions to game balance as you'd be able to hear footstep sounds a mile away.

The gist of this system is more in depth than our present one in that it allows for having defined sounds based upon the type of footstep a mob has; so, for example, a mob with claws (that are non human) has a particular set of sounds when stepping on wood or regular floors vs a human mob; it allows for sounds based on wearing shoes or going barefoot (again for human mobs).

This also adds a lot more sounds for various turfs; it's no longer just plating, carpet, and wood floors that make noise when stepped out; sand, water, regular floor, and even grass all have unique sounds they play when they're stepped on.

The overall goal is to make things a bit more immersive, but also add audio cues to movement, in general, while making silent footsteps a more balance impacting thing rather than just flavor. 

If you want to have quieter footsteps, you can walk, which lowers the volume produced.

As such, there's a few things that have been adjusted:

- You can no longer silence shoes/footsteps, in general, unless you have a specific trait for it
- All human species have a footstep sound now
- Jackboots are no longer Half-Life Roleplay.

Overall, this system, along with all the other sound changes, really brings the station to life, while adding audio cues for most mob's movement.

In the future, if we ever implement something like quirks, I'll be introducing the "light step" quirk, which makes your footsteps quieter, but it's not a "Freebie" so to speak.

:cl: Fox McCloud
add: Adds wider footstep sound system--can now hear different footstep sounds on a lot of different surface including wood floors, regular floors, plating, carpet, grass, lava, and sand-like surfaces
tweak: All human species have footstep sounds
tweak: Many mobs have their own unique footstep sounds
del: Jackboots no longer have a unique sound. Good riddance
/:cl:


FreeStylaLT:
https://user-images.githubusercontent.com/11361525/109402388-38b1df80-795e-11eb-8d31-c8703a033378.mp4

SabreML:
Personally I'm very much in favour of more footstep sounds, it's one of the features I'm most jealous of from TG.
What I'm not a fan of is removing silent footsteps from Vox, especially since it would literally be a single line change to keep it in:
`inherent_traits = list(TRAIT_NOGERMS, TRAIT_NODECAY)`
to
`inherent_traits = list(TRAIT_NOGERMS, TRAIT_NODECAY, TRAIT_SILENT_FOOTSTEPS)`

If you're insistent on it being removed though it should probably at least have its own changelog entry in order to avoid confusion, like with #15048. Since it *is* technically a species nerf.

Fox-McCloud:
>like with #15048

https://github.com/ParadiseSS13/Paradise/pull/13430
`add: All species can now be fat`

Calling that "unintentional" is fabricaction.

The CL quite definitely states that *all* species have footstep sounds now, I dunno how much clearer I can make it.

SabreML:
> The CL quite definitely states that all species have footstep sounds now, I dunno how much clearer I can make it.

#13430 also had the same thing in the changelog, but I don't think *anyone* actually knew that it had been removed.
Although in fairness that was most likely becuase the wiki page wasn't updated to reflect the changes.

The best way to make it clearer I think would be to specify that it affects Vox in the changelog, since they're the only species *without* footstep sounds currently, and therefore the only species being changed.
`tweak: Vox now have footstep sounds for consistency.` or something like that, rather than `tweak: All human species have footstep sounds`, since Vox are the only human species being altered.

<hr>

I'm still against the Vox nerf *(No matter how small it may be)* especially when, in my opinion, it's not a nessecary change and seems to have been done primarily for consistency with other species. Yes, it *will* have a gameplay effect with this new system, but I don't see that as a bad thing at all. This is why we have species advantages/disadvantages; to make them all feel unique in some way or another. *(At least in theory)*

I'm aware however that no matter how much I try to argue for this, nothing I say can actually make a difference if you've made your mind up. And I'm not about to start bugging the other maintainers/headmins to take my point of view. Everyone has different viewpoints on what should and shouldn't be added to the game.

code/datums/components/footstep.dm
```diff
@@ -0,0 +1,133 @@
...
+		return COMPONENT_INCOMPATIBLE
+	volume = volume_
+	e_range = e_range_
+	footstep_type = footstep_type_
+	sound_vary = vary
+	switch(footstep_type)
+		if(FOOTSTEP_MOB_HUMAN)
+			if(!ishuman(parent))
+				return COMPONENT_INCOMPATIBLE
+			RegisterSignal(parent, list(COMSIG_MOVABLE_MOVED), .proc/play_humanstep)
```
SteelSlayer, line 27: There's also 3 other unnecessary `list()`s inside of the `RegisterSignal()`s below, but I won't make a review for each. 
```suggestion
			RegisterSignal(parent, COMSIG_MOVABLE_MOVED, .proc/play_humanstep)
```

code/game/turfs/simulated/floor/indestructible.dm
```diff
@@ -1,4 +1,8 @@
 /turf/simulated/floor/indestructible
+	footstep = FOOTSTEP_FLOOR
+	barefootstep = FOOTSTEP_HARD_BAREFOOT
+	clawfootstep = FOOTSTEP_HARD_CLAW
+	heavyfootstep = FOOTSTEP_GENERIC_HEAVY
```
SteelSlayer, line 5: These are already set by the parent, `turf/simulated/floor`

Aracino:
As much as I hate the jackboot noise, it was pretty important knowing if sec was rushing your position. (Or just some assistant who nabbed a pair from the old bar maintenance area). Will this still give an audible indication for antags that secs around the corner?

S34NW:
I don't see why removing jackboot noises is a good idea, secoffs that want quiet boots go on a tape hunt anyway.
Having noisy boots as an option causes no harm to anyone.

MRgaty:
> As much as I hate the jackboot noise, it was pretty important knowing if sec was rushing your position. (Or just some assistant who nabbed a pair from the old bar maintenance area). Will this still give an audible indication for antags that secs around the corner?

Could always tape your shoes to prevent this, so this change will make no difference

Spacemanspark:
Please do not remove the jackboot sound effect. I'm aware you aren't the biggest fan of them, but many players like it and it's completely harmless.

Kyep (approved):

AffectedArc07 (approved):

PR #15631, opened 2021-02-28
Fixes ambient sounds playing in the lobby

SabreML:
## What Does This PR Do

Removes ambient sounds from the `/area/start` area, primarily so that they don't overlap with the lobby music.

## Why It's Good For The Game

It sounds confusing, and is most likely an unintentional change.

## Changelog
:cl:
fix: Fixed ambient sounds playing on the lobby/menu screen.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ambience.dm
+++ b/code/controllers/subsystem/ambience.dm
@@ -21,7 +21,7 @@ SUBSYSTEM_DEF(ambience)
 
 		var/area/current_area = get_area(client_iterator.mob)
 
-		var/ambience = pick(current_area.ambientsounds)
+		var/ambience = safepick(current_area.ambientsounds)
 		if(!ambience)
 			continue
 
--- a/code/controllers/subsystem/ambience.dm
+++ b/code/controllers/subsystem/ambience.dm
@@ -21,8 +21,10 @@ SUBSYSTEM_DEF(ambience)
 
 		var/area/current_area = get_area(client_iterator.mob)
 
-		var/sound = pick(current_area.ambientsounds)
+		var/ambience = pick(current_area.ambientsounds)
+		if(!ambience)
+			continue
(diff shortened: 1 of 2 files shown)


Fox-McCloud:
On one hand, it's a bug--on the other, it only occurs after round has started and it does have kinda an atmospheric vibe when you're just sitting on the screen.

Not sure about this or not, to be honest.

SabreML:
> On one hand, it's a bug--on the other, it only occurs after round has started and it does have kinda an atmospheric vibe when you're just sitting on the screen.
> 
> Not sure about this or not, to be honest.

If there's no lobby music playing then sure, it might be atmospheric. But currently it just seems to clash horribly.
And you also have to consider anyone that joins after roundstart, since they get both.

dearmochi (approved):

S34NW:
It just sounds weird and out of place, good riddance.

AffectedArc07 (approved):
Im in favour because ambience and lobby music clashing does sound bad, and this affects people who join the lobby mid round.

code/controllers/subsystem/ambience.dm
```diff
@@ -21,8 +21,10 @@ SUBSYSTEM_DEF(ambience)
 
 		var/area/current_area = get_area(client_iterator.mob)
 
-		var/sound = pick(current_area.ambientsounds)
+		var/ambience = pick(current_area.ambientsounds)
+		if(!ambience)
```
AffectedArc07, line 25: This can result in `error: pick from empty list`

Consider using `safepick`

SabreML: Ah, I'd assumed that the `pick()` proc would just pass on `null` if it was empty.

AffectedArc07 (approved):

PR #15635, opened 2021-02-28
Prevents lit fire sources being put into backpacks

SabreML:
## What Does This PR Do

Stops users from putting Welding tools, Lighters, Flamethrowers, Candles, Matches, Cigarettes, and Cigars into a storage container while they're lit.
Also adds the `can_enter_storage()` proc, which can be overridden with failure conditions.

## Why It's Good For The Game

It doesn't make much sense that you could put an active fire source into a cloth/cardboard container safely.

## Changelog
:cl:
tweak: Welding tools, Lighters, Flamethrowers, Candles, etc. can no longer be put into backpacks or boxes while they're lit.
/:cl:


dearmochi:
Might be worthwhile to check out https://github.com/ParadiseSS13/Paradise/blob/20525d361924cbb7abd912b14c73ecf38848b1f4/code/__HELPERS/unsorted.dm#L1132 to cut down on duplicate code. Maybe add such a check in the base proc.

SabreML:
> Might be worthwhile to check out
> 
> https://github.com/ParadiseSS13/Paradise/blob/20525d361924cbb7abd912b14c73ecf38848b1f4/code/__HELPERS/unsorted.dm#L1132
> 
> to cut down on duplicate code. Maybe add such a check in the base proc.

That would work if it wasn't for the igniter and plasmacutter checks. Don't want either of them being blocked.

Aracino:
What about when something is lite while in a box? Like if you light a pack of candles while they're inside the box, which I often do to speed up decorating the chapel.

SabreML:
> What about when something is lit while in a box? Like if you light a pack of candles while they're inside the box, which I often do to speed up decorating the chapel.

In that case lighting the candles will work the same as before and they'll all stay inside the box, but you won't be able to put them back in if you take them out.

kugamo:
I will put this here in case you come back to this with an eye to rework it alitte. I suggest making lit objects while inside flammable containers over some interval (30seconds or 2 min or something) have a chance to light the container on fire. With a bold text warning explaining what happened of course. I dont like completely disabling functionality like this, and we should allow people to do idiotic things if they want (just discourage it through clear feedback in gameplay).

AffectedArc07 (approved):

PR #15639, opened 2021-02-28
Adds a cooldown to some mob spawners

Qwertytoforty:
## What Does This PR Do

Adds a death_cooldown variable to mob spawners, which prevents anyone that has been dead for less then the time specified from the variable, from using the spawner, till they are dead long enough. If you have joined as ghost from lobby, the timer does not apply, unless you have already used a mob spawner.

Adds a 5 minute cooldown to golem shells and ashwalkers.

## Why It's Good For The Game

This is good for 2 reasons. It prevents someone from roundstart taking a spawner, dying to a chasm or sleeping near the tendril, then respawning and doing it 2 more times, ruining the mob spawner for everyone else.

The second reason this is good, is to prevent a near infinite respawn army, where there is 15 golem shells / eggs, and a player once they die, instantly respawn, ready to fight. This has been seen many times, whether it be xenobio making a hijack army and spamming golem shells that instantly fill, or ashwalker eggs, being near impossible to take out the tendril, as the walkers instantly respawn and attack once they die.

## Changelog
:cl:
add: Mob spawners can now have a death cooldown, preventing people from using them till they are dead long enough.
tweak: Ash walker eggs and golem shells now have a 5minute cooldown.
/:cl:

### DIFF
--- a/code/modules/awaymissions/corpse.dm
+++ b/code/modules/awaymissions/corpse.dm
@@ -94,7 +94,7 @@
 	var/joinedasobserver = FALSE
 	if(isobserver(user))
 		var/mob/dead/observer/G = user
-		if(G.started_as_observer == TRUE)
+		if(G.started_as_observer)
 			joinedasobserver = TRUE
 
 	var/deathtimeminutes = round(deathtime / 600)
@@ -107,7 +107,7 @@
 		pluralcheck = " [deathtimeminutes] minutes and"
 	var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10, 1)
 
-	if(deathtime < (death_cooldown MINUTES) && joinedasobserver == 0)
+	if(deathtime < (death_cooldown MINUTES) && !joinedasobserver)
 		to_chat(user, "You have been dead for[pluralcheck] [deathtimeseconds] seconds.")
 		to_chat(user, "<span class='warning'>You must wait [death_cooldown] minutes to respawn!</span>")
 		return TRUE
--- a/code/modules/awaymissions/corpse.dm
+++ b/code/modules/awaymissions/corpse.dm
@@ -33,7 +33,7 @@
 	var/banType = ROLE_GHOST
 	var/ghost_usable = TRUE
 	var/offstation_role = TRUE // If set to true, the role of the user's mind will be set to offstation
-	var/death_cooldown = 0 // How long you have to wait after dying before using it again, in minutes. People that join as observers are not included.
+	var/death_cooldown = 0 // How long you have to wait after dying before using it again, in deciseconds. People that join as observers are not included.
 
 /obj/effect/mob_spawn/attack_ghost(mob/user)
 	var/mob/dead/observer/O = user
@@ -107,9 +107,9 @@
 		pluralcheck = " [deathtimeminutes] minutes and"
 	var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10, 1)
 
-	if(deathtime < (death_cooldown MINUTES) && !joinedasobserver)
+	if(deathtime < (death_cooldown) && !joinedasobserver)
 		to_chat(user, "You have been dead for[pluralcheck] [deathtimeseconds] seconds.")
-		to_chat(user, "<span class='warning'>You must wait [death_cooldown] minutes to respawn!</span>")
+		to_chat(user, "<span class='warning'>You must wait [death_cooldown / 600] minutes to respawn!</span>")
 		return TRUE
 	return FALSE
 
--- a/code/modules/awaymissions/corpse.dm
+++ b/code/modules/awaymissions/corpse.dm
@@ -33,6 +33,7 @@
 	var/banType = ROLE_GHOST
 	var/ghost_usable = TRUE
 	var/offstation_role = TRUE // If set to true, the role of the user's mind will be set to offstation
+	var/death_cooldown = 0 // How long you have to wait after dying before using it again, in minutes. People that join as observers are not included.
 
 /obj/effect/mob_spawn/attack_ghost(mob/user)
 	var/mob/dead/observer/O = user
@@ -49,6 +50,8 @@
 		return
 	if(!O.can_reenter_corpse)
 		to_chat(user, "<span class='warning'>You have forfeited the right to respawn.</span>")
+		return
+	if(time_check(user))
 		return
 	var/ghost_role = alert("Become [mob_name]? (Warning, You can no longer be cloned!)",,"Yes","No")
 	if(ghost_role == "No")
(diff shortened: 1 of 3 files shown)


Aracino:
To a degree, I think this should also apply to borgs. It can feel bad when you destroy a borg, but they just ghost and get a new body in robotics immediatly.

hal9000PR:
> To a degree, I think this should also apply to borgs. It can feel bad when you destroy a borg, but they just ghost and get a new body in robotics immediatly.

Completely agree, however that is for a different PR.

PidgeyThePirate:
Interesting that this PR goes up very shortly after the miners and security got trounced by wave after wave of ashwalkers who were defending their home after spending the last hour doing basically nothing but building up eggs.

I guess a tactic that is only useful in home defense and gave ashwalkers a fighting chance against inevitable (and unwarranted) slaughter is too much, huh?

Denthamos:
Echoing an above comment, keep in mind for ash walkers its one of the very few things they have going for them currently.  They represent legion for they are many.  

If this is to be taken away from them then they should be buffed in some way because as it is right now they have a limited number of weapons for said spawns and once all their spears are gone its basically just a bunch of greytiders with nothing but fists trying to attack super powered miners that have infinite charge ballistic weapons, mechs, and more nonsense if they farmed some tendrils/megafauna.  Even then, a five minute timer feels too long considering the nature of the role and what they are up against.

hal9000PR:
> `add: Mob spawners can now have a death cooldown, preventing people from using them till they are dead long enough.`

does this mean you cannot join as mob spawners round start?

Dashsora:
My big problem with the walker respawns isn't when they're getting assaulted, it's when they attack individual miners/megafauna, throwing their bodies at the problem without any consideration for their lives, because they know they can just pop back out and immediately meta the location of their deaths, bee-line to it, and loot/finish the fight. 

Not only does this turn the usual "high risk, high reward" playstyle of lavaland into "no-risk, high reward", it's also pretty contrary to encouraging RP. 

eg : Why would an ashwalker CAUTIOUSLY try to ascertain if the miner they just met is friendly or not, or run from them in the fear they are not, when they can simply walk up to them and test the water this way ? If the miner kills them, they can just respawn and, either avoid them, or come back with more and kill them in return. On the flipside, though, knowing this and playing as a miner, why would you ever want to risk interacting with walkers ?? There's a reason some people just beeline the tendril and destroy it while walkers are still weak. 

Yes, walkers are "overall weak" down on lavaland, to miners and fauna alike, and i'd love to see ways for them to have more means to gain staying power (other than to rob miners), especially in their HQ, but allowing for this mindless swarm of lizards really isn't the way to go, imho.

 I'd totally be down to limitting walkers/golems/any other ghost spawner to one spawn per person per round, but I can see how since a lot of them have a role in teaching new players the game, that might be a little too harsh. In any case, five minutes cooldown certainly isn't going to dampen the swarm all that much, but it'll hopefully force the players to think a little more about their characters' lives before throwing them away.

PidgeyThePirate:
I don't think so. If you're in agreement that they're already weak, then the buff and the nerf should happen at the same time. All this is is a way to let miners focus down the tendril with even less than trivial resistance.

Kyep (approved):

Dashsora:
I do think a buff should come too, but this is already something that would benefit the miner/walker interraction, and further RP by making a walker's life more meaningful than something they can just throw away at will.

I also think destroying the tendril without any reason should be against the rules.

PidgeyThePirate:
The walkers do pretty well, but I do think things that would be really helpful for the building aspect of them would be the ability to create new shelters with a "necropolis land claim" (They might need inflatable barriers or a holofan to seal the rooms properly), access to something that will let them clean up blood and mess (though I guess gutlunches *miiiight* do?) and the ability to remove those stone tiles that look nice but are really annoying. That last one is a big one. A spraycan would be nice too so they can put markings around their buildings and territory.

code/modules/awaymissions/corpse.dm
```diff
@@ -86,6 +89,30 @@
...
+
+	var/deathtimeminutes = round(deathtime / 600)
+	var/pluralcheck = "minute"
+	if(deathtimeminutes == 0)
+		pluralcheck = ""
+	else if(deathtimeminutes == 1)
+		pluralcheck = " [deathtimeminutes] minute and"
+	else if(deathtimeminutes > 1)
+		pluralcheck = " [deathtimeminutes] minutes and"
+	var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10, 1)
```
dearmochi, line 108: I believe this is the same code as with maintenance drones -- this really needs its own proc.

code/modules/awaymissions/corpse.dm
```diff
@@ -86,6 +89,30 @@
 /obj/effect/mob_spawn/proc/equip(mob/M)
 	return
 
+/obj/effect/mob_spawn/proc/time_check(mob/user)
+	var/deathtime = world.time - user.timeofdeath
+	var/joinedasobserver = FALSE
+	if(isobserver(user))
+		var/mob/dead/observer/G = user
+		if(G.started_as_observer == TRUE)
```
dearmochi, line 97:
```suggestion
		if(G.started_as_observer)
```

code/modules/awaymissions/corpse.dm
```diff
@@ -86,6 +89,30 @@
...
+	var/pluralcheck = "minute"
+	if(deathtimeminutes == 0)
+		pluralcheck = ""
+	else if(deathtimeminutes == 1)
+		pluralcheck = " [deathtimeminutes] minute and"
+	else if(deathtimeminutes > 1)
+		pluralcheck = " [deathtimeminutes] minutes and"
+	var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10, 1)
+
+	if(deathtime < (death_cooldown MINUTES) && joinedasobserver == 0)
```
dearmochi, line 110:
```suggestion
	if(deathtime < (death_cooldown MINUTES) && !joinedasobserver)
```

code/modules/awaymissions/corpse.dm
```diff
@@ -33,6 +33,7 @@
 	var/banType = ROLE_GHOST
 	var/ghost_usable = TRUE
 	var/offstation_role = TRUE // If set to true, the role of the user's mind will be set to offstation
+	var/death_cooldown = 0 // How long you have to wait after dying before using it again, in minutes. People that join as observers are not included.
```
dearmochi, line 36: I'd personally prefer this is was in deciseconds so a coder can directly use the time macros instead of just say `5`.

AffectedArc07 (changes requested):
Please address requested changes.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -61,6 +61,7 @@
 	anchored = FALSE
 	move_resist = MOVE_FORCE_NORMAL
 	density = FALSE
+	death_cooldown = 3000
```
dearmochi, line 64:
```suggestion
	death_cooldown = 300 SECONDS
```

code/modules/ruins/lavalandruin_code/ash_walker_den.dm
```diff
@@ -67,6 +67,7 @@
 	anchored = FALSE
 	move_resist = MOVE_FORCE_NORMAL
 	density = FALSE
+	death_cooldown = 3000
```
dearmochi, line 70:
```suggestion
	death_cooldown = 300 SECONDS
```

code/modules/awaymissions/corpse.dm
```diff
@@ -86,6 +89,30 @@
...
+	var/pluralcheck = "minute"
+	if(deathtimeminutes == 0)
+		pluralcheck = ""
+	else if(deathtimeminutes == 1)
+		pluralcheck = " [deathtimeminutes] minute and"
+	else if(deathtimeminutes > 1)
+		pluralcheck = " [deathtimeminutes] minutes and"
+	var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10, 1)
+
+	if(deathtime <= (death_cooldown) && !joinedasobserver)
```
dearmochi, line 110:
```suggestion
	if(deathtime <= death_cooldown && !joinedasobserver)
```

dearmochi (approved):
Seems fine otherwise

AffectedArc07 (approved):

PR #15643, opened 2021-03-01
Nanomaps action tweaks

AffectedArc07:
## What Does This PR Do
This PR converts the Nanomap action into one which will create a PR, in the same way that the DMAPI action does. This has 2 benefits

1. People who have forked no longer need to disable actions to avoid spurious commits
2. This allows for branch protection rules to the repo master branch, as this is currently the last service which creates commits to the master branch

## Why It's Good For The Repo
Direct commits bad, branch protection good

## Changelog
N/A


Fox-McCloud (approved):
LGTM

PR #15651, opened 2021-03-02
Makes Mugwort Tea remove harmful chems

SabreML:
## What Does This PR Do

Makes Mugwort Tea remove all non-harmless chemicals at five units per tick when the Wizard drinks it.

## Why It's Good For The Game

It's not really fun for anybody when the Wizard gets shot with an ether/sarin syringe and the round ends.
Of course, this depends on the Wizard actually remembering to take the tea with them, so who knows. 🤷

## Changelog
:cl:
tweak: Mugwort Tea will now remove harmful chemicals at the same rate as Calomel for Wizards.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/food.dm
+++ b/code/modules/reagents/chemistry/reagents/food.dm
@@ -594,12 +594,14 @@
 
 /datum/reagent/consumable/mugwort/on_mob_life(mob/living/M)
 	var/update_flags = STATUS_UPDATE_NONE
-	if(ishuman(M) && M.mind)
-		if(M.mind.special_role == SPECIAL_ROLE_WIZARD)
-			update_flags |= M.adjustToxLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
-			update_flags |= M.adjustOxyLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
-			update_flags |= M.adjustBruteLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
-			update_flags |= M.adjustFireLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
+	if(ishuman(M) && M.mind?.special_role == SPECIAL_ROLE_WIZARD)
+		update_flags |= M.adjustToxLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		update_flags |= M.adjustOxyLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		update_flags |= M.adjustBruteLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		update_flags |= M.adjustFireLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/reagents/food.dm
```diff
@@ -594,12 +594,14 @@
...
-			update_flags |= M.adjustOxyLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
-			update_flags |= M.adjustBruteLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
-			update_flags |= M.adjustFireLoss(-1*REAGENTS_EFFECT_MULTIPLIER, FALSE)
+	if(ishuman(M) && M.mind?.special_role == SPECIAL_ROLE_WIZARD)
+		update_flags |= M.adjustToxLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		update_flags |= M.adjustOxyLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		update_flags |= M.adjustBruteLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		update_flags |= M.adjustFireLoss(-1 * REAGENTS_EFFECT_MULTIPLIER, FALSE)
+		for(var/datum/reagent/R in M.reagents.reagent_list)
+			if(R != src && !R.harmless)
```
TDSSS, line 603: Couldn't you just flag mugwort as harmless (it is, after all. Right?) and safe yourself the R!= src check?

SabreML: Uhhhhh

Sneedstar:
This is a must have!

SteelSlayer (approved):

code/modules/reagents/chemistry/reagents/food.dm
```diff
@@ -591,15 +591,18 @@
 	reagent_state = LIQUID
 	color = "#21170E"
 	taste_description = "tea"
+	harmless = TRUE
```
S34NW, line 594: I don't think you need to specify this, `/datum/reagent/consumable` defaults to `harmless = TRUE` and you only need to specify if a reagent is not harmless.

S34NW (changes requested):

code/modules/reagents/chemistry/reagents/food.dm
SabreML: Damn it, this is what happens when I code at 3AM.
It does default to `TRUE`, but I'll leave this here anyway to avoid any possible confusion.

dearmochi (approved):

Kyep (approved):
Memechems are one of the most effective anti-wizard weapons. I'll be sad to see that go.

On the flipside, its probably for the best, as a one shot with a death mix isn't really a good-for-the-server way to end the round defining antag.

Plus, mugwort tea is literally magic, so its ability to purge chems does make some sense. And likely only the smarter wizards will actually use it. Smart wizards already have options like the revive wand that fixes all chems, so this is not super out of place balance wise either.

FlattestGuitar:
The sarin/ether syringe is still effective against the 95% of wizards that won't know that mugwort tea could save them.

AffectedArc07 (approved):

Issue #15659, opened 2021-03-04
Skrell headpocket sends item to Nullspace


Somniworld:
**Issue Description**:

Placing any item in the skrell headpocket will be sent to nullspace, this makes them impossible to interact with due to being "too far away", this also makes the nuke auth disk teleport into maintenance.

**What did you expect to happen**: 

I expected the items, especially the nuke auth disk, to remain within the headpocket, in a somewhat safe storage area that is easy to access and often overlooked.

**What happened instead**:

The nuclear disk was teleported to maintenance without me knowing, giving me a minor heart attack and making me look like a fool, furthermore, the captain's spare, which I tried to place in later on, was ALSO sent to the shadow realm, and made me unable to interact with it.

**Why is this bad/What are the consequences**:

I believe that it is pretty obvious why this is an issue, headpockets can hold some surprising items, but having a portal to the bottomless void at the back of your head sounds midly unrealistic, and uncomfortable.

**Steps to reproduce the problem**:

Make a skrell character, with tentacles (in other word, don't be bald, so as to have the headpocket).
Put any item within the headpocket.
Watch in dismay as it is somehow too far away, or as the nuke disk teleports out of the headpocket after a while.

**When did the problem start happening**:

I believe this was caused after PR #15548, which moved the headpocket, and thus it's content, to Nullspace. I know that Fox stated that it seemed to function correctly, but that is the  only lead I have on this, and it was working fine prior to merging.

**Extra information**:

God damn you Fox! You and your ... *shuffles cards* ...Interdimentional headpockets!

PR #15660, opened 2021-03-04
Fixes Skrell Headpockets

Fox-McCloud:
Fixes Skrell headpockets misbehaving.

Skrell headpockets are no longer explicitly a storage subtype that has internal storage; instead if just uses a holder var to hold the item.

I'd have to rewrite some click code and storage code for headpockets to work properly, and that's just not feasible just to fix a feature for *one* species, at the moment.

Also, you'll no longer be able to hold the nuke disk in the headpocket, because, you never should have to begin with; hidden storage tpyically bars it, specifically.

fixes: https://github.com/ParadiseSS13/Paradise/issues/15659

:cl: Fox McCloud
fix: Fixes skrell headpocket weirdness
fix: Fixes being able to hide nuke disk in the Skrell headpocket
/:cl:

### DIFF
--- a/code/modules/surgery/organs/subtypes/skrell.dm
+++ b/code/modules/surgery/organs/subtypes/skrell.dm
@@ -23,18 +23,18 @@
 	..()
 	var/obj/item/organ/external/head/head = owner.get_organ("head")
 	if(held_item && !findtextEx(head.h_style, "Tentacles"))
-		owner.visible_message("<span class='notice'>[held_item] falls from [owner]'s [src]!</span>", "<span class='notice'>[held_item] falls from your [src]!</span>")
+		owner.visible_message("<span class='notice'>[held_item] falls from [owner]'s [name]!</span>", "<span class='notice'>[held_item] falls from your [name]!</span>")
 		empty_contents()
 
 /obj/item/organ/internal/headpocket/ui_action_click()
 	if(held_item)
-		owner.visible_message("<span class='notice'>[owner] removes [held_item] from [owner.p_their()] [src].</span>", "<span class='notice'>You remove [held_item] from your [src].</span>")
+		owner.visible_message("<span class='notice'>[owner] removes [held_item] from [owner.p_their()] [name].</span>", "<span class='notice'>You remove [held_item] from your [name].</span>")
 		owner.put_in_hands(held_item)
 		held_item = null
 	else
 		var/obj/item/I = owner.get_active_hand()
 		if(I && I.w_class <= WEIGHT_CLASS_SMALL && !istype(I, /obj/item/disk/nuclear) && owner.unEquip(I))
-			owner.visible_message("<span class='notice'>[owner] places [I] into [owner.p_their()] [src].</span>", "<span class='notice'>You place [I] into your [src].</span>")
+			owner.visible_message("<span class='notice'>[owner] places [I] into [owner.p_their()] [name].</span>", "<span class='notice'>You place [I] into your [name].</span>")
 			I.forceMove(src)
 			held_item = I
 
--- a/code/modules/surgery/organs/subtypes/skrell.dm
+++ b/code/modules/surgery/organs/subtypes/skrell.dm
@@ -9,42 +9,34 @@
 	icon = 'icons/obj/species_organs/skrell.dmi'
 	icon_state = "skrell_headpocket"
 	origin_tech = "biotech=2"
-	w_class = WEIGHT_CLASS_TINY
+	w_class = WEIGHT_CLASS_SMALL
 	parent_organ = "head"
 	slot = "headpocket"
 	actions_types = list(/datum/action/item_action/organ_action/toggle)
-	var/obj/item/storage/internal/pocket
-
-/obj/item/organ/internal/headpocket/New()
-	pocket = new /obj/item/storage/internal(src)
-	pocket.storage_slots = 1
-	// Fit only pocket sized items
-	pocket.max_w_class = WEIGHT_CLASS_SMALL
-	pocket.max_combined_w_class = 2
-	..()
+	var/obj/item/held_item
 
 /obj/item/organ/internal/headpocket/Destroy()
-	QDEL_NULL(pocket)
-	return ..()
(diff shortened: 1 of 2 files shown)


code/modules/surgery/organs/subtypes/skrell.dm
```diff
@@ -9,42 +9,34 @@
 	icon = 'icons/obj/species_organs/skrell.dmi'
 	icon_state = "skrell_headpocket"
 	origin_tech = "biotech=2"
-	w_class = WEIGHT_CLASS_TINY
+	w_class = WEIGHT_CLASS_SMALL
```
Fox-McCloud, line 12: Something tiny can hold something small. No.

TDSSS:
> Also, you'll no longer be able to hold the nuke disk in the headpocket, because, you never should have to begin with; hidden storage tpyically bars it, specifically.

What do you mean with hidden, exactly? People can search them like they can search pockets, can't they? So they're not any more secure than those.

code/modules/surgery/organs/subtypes/skrell.dm
```diff
@@ -9,42 +9,34 @@
...
 	return ..()
 
 /obj/item/organ/internal/headpocket/on_life()
 	..()
 	var/obj/item/organ/external/head/head = owner.get_organ("head")
-	if(pocket.contents.len && !findtextEx(head.h_style, "Tentacles"))
-		owner.visible_message("<span class='notice'>Something falls from [owner]'s head!</span>",
-													"<span class='notice'>Something falls from your head!</span>")
+	if(held_item && !findtextEx(head.h_style, "Tentacles"))
+		owner.visible_message("<span class='notice'>[held_item] falls from [owner]'s [src]!</span>", "<span class='notice'>[held_item] falls from your [src]!</span>")
```
SabreML, line 26: Needs to be changed to `falls from your [name]!` otherwise you get `falls from your the headpocket!`

code/modules/surgery/organs/subtypes/skrell.dm
```diff
@@ -9,42 +9,34 @@
...
 /obj/item/organ/internal/headpocket/ui_action_click()
-	pocket.MouseDrop(owner)
+	if(held_item)
+		owner.visible_message("<span class='notice'>[owner] removes [held_item] from [owner.p_their()] [src].</span>", "<span class='notice'>You remove [held_item] from your [src].</span>")
+		owner.put_in_hands(held_item)
+		held_item = null
+	else
+		var/obj/item/I = owner.get_active_hand()
+		if(I && I.w_class <= WEIGHT_CLASS_SMALL && !istype(I, /obj/item/disk/nuclear) && owner.unEquip(I))
+			owner.visible_message("<span class='notice'>[owner] places [I] into [owner.p_their()] [src].</span>", "<span class='notice'>You place [I] into your [src].</span>")
```
SabreML, line 37: Same with these two.

code/modules/surgery/organs/subtypes/skrell.dm
```diff
@@ -9,42 +9,34 @@
...
 
 /obj/item/organ/internal/headpocket/ui_action_click()
-	pocket.MouseDrop(owner)
+	if(held_item)
+		owner.visible_message("<span class='notice'>[owner] removes [held_item] from [owner.p_their()] [src].</span>", "<span class='notice'>You remove [held_item] from your [src].</span>")
+		owner.put_in_hands(held_item)
+		held_item = null
+	else
+		var/obj/item/I = owner.get_active_hand()
+		if(I && I.w_class <= WEIGHT_CLASS_SMALL && !istype(I, /obj/item/disk/nuclear) && owner.unEquip(I))
```
SabreML, line 36: And to echo what TDSSS said, why exclude the nuke disk from being held in headpockets? It functions exactly the same as a regular pocket and can be easily emptied:
image

Somniworld: As, to my knowledge, the only current user of this specific mechanic -due to playing skrell as well as captain-, I can't help but second what Sabre said, putting the nuke disk in the headpocket is of little impact, anyone who pays attention to your specie is likely to find out the trick (especially considering you're already toast if the nukies manage to get their hands on you), on top of that, you have no way of hiding this pocket unless wearing a hardsuit helmet.

The reason why I do it is that it's a very easy way to check on the disk and quickly cope with the anxiety I get from playing captain on war ops, since you only have one item, in an easy to access spot, while it may to some extent "hide" the disk, it's no more than just slapping it in a cardboard box in your bag.

Fox-McCloud: The very nature of how organs are handled now is that they're in nullspace---ergo anything you hold onto is also in null space, even the item has a `loc` inside the head pocket.

The pinpointer won't work at all as it won't be able to find it.

Refactor a ton of code just so one species can hold the nuke disk in their special organ? Not something worth it, since it functions 99.99% of the time for other things.

TDSSS:
Does that mean you can hide other steal objectives from an advanced pinpointer in your head pocket?

AffectedArc07:
> Does that mean you can hide other steal objectives from an advanced pinpointer in your head pocket?

Yes but this is obvious. If an RD skrell is hiding a hand tele in their headpocket, the RD is a target anyway since they are usually attacked for the hand tele, so the pocket can be searched. If the RD is giving the hand tele to a random on-station skrell just to avoid someone greentexting, thats then an administrative problem and should be dealt with.

AffectedArc07 (approved):

TDSSS:
> > Does that mean you can hide other steal objectives from an advanced pinpointer in your head pocket?
> 
> Yes but this is obvious. If an RD skrell is hiding a hand tele in their headpocket, the RD is a target anyway since they are usually attacked for the hand tele, so the pocket can be searched. If the RD is giving the hand tele to a random on-station skrell just to avoid someone greentexting, thats then an administrative problem and should be dealt with.

It also means if one traitor has to steal an item and stores it in his headpocket, other traitors can't find him with the pinpointer to take it from him.

PR #15664, opened 2021-03-04
Adds gamma alert request to keycard device

S34NW:
## What Does This PR Do
Adds the option to request gamma alert from CC via the keycard device.
Gamma alert can only be requested from red alert.
Notifies admins in the same way as ERT requests, with an easy option to accept or deny.
~~Gamma alert is auto-confirmed if requested with no admins online.~~ That was _not_ popular, yikes.
Logs users who request and confirm for admin purposes.

## Why It's Good For The Game
The current process to request gamma alert requires the captain to use their quantum communication, or for a command member with a functioning long-range fax machine to be able to draft a fax.
This makes it easier for other command members to request gamma alert in the event of an incapacitated captain or NT Rep.
I was going to PR a more extensive change similar to [what I suggested here](https://www.paradisestation.org/forum/topic/20372-decouple-gamma-alert-from-the-armoury-shuttle/), but that needs further feedback and discussion. So, for now, this can be considered a step in that direction by adding a QoL feature.

## Images of changes
image
image
image

## Changelog
:cl:
add: Gamma alert can now be requested on the keycard device
/:cl:


code/modules/admin/topic.dm
```diff
@@ -2425,8 +2425,32 @@
 			to_chat(H, "<span class='specialnoticebold'>Incoming priority transmission from Central Command. Message as follows,</span><span class='specialnotice'> Your ERT request has been denied for the following reasons: [input].</span>")
 		else
 			src.owner.response_team()
+			
+	else if(href_list["GammaReply"])
+		if(!check_rights(R_ADMIN))
+			return
 
+		if(alert(src.owner, "Accept or Deny Gamma Alert request?", "CentComm Response", "Accept", "Deny") == "Deny")
```
Kyep, line 2433:
```suggestion
		if(alert(owner, "Accept or Deny Gamma Alert request?", "CentComm Response", "Accept", "Deny") == "Deny")
```
Please don't use "src." unless its truly needed.

code/modules/admin/topic.dm
```diff
@@ -2425,8 +2425,32 @@
...
+			var/mob/living/carbon/human/H = locateUID(href_list["GammaReply"])
+			if(!istype(H))
+				to_chat(usr, "<span class='warning'>This can only be used on instances of type /mob/living/carbon/human</span>")
+				return
+			if(H.stat != 0)
```
Kyep, line 2438:
```suggestion
			if(H.stat != CONSCIOUS)
```
Please don't use hardcoded numbers like 0 when defines exist for the thing you are checking.

code/modules/admin/topic.dm
```diff
@@ -2425,8 +2425,32 @@
...
+				to_chat(usr, "<span class='warning'>The person you are trying to contact is not conscious.</span>")
+				return
+			if(!istype(H.l_ear, /obj/item/radio/headset) && !istype(H.r_ear, /obj/item/radio/headset))
+				to_chat(usr, "<span class='warning'>The person you are trying to contact is not wearing a headset</span>")
+				return
 
+			var/input = input(src.owner, "Please enter a reason for denying [key_name(H)]'s Gamma Alert request.","Outgoing message from CentComm", "")
+			if(!input)	return
```
Kyep, line 2446: Please don't put "return" on the same line as "if".
```suggestion
			if(!input)
				return
```

code/modules/admin/topic.dm
```diff
@@ -2425,8 +2425,32 @@
...
+			GLOB.gamma_request_answered = TRUE
```
Kyep, line 2447: Is this copy-paste?

code/modules/security_levels/keycard_authentication.dm
```diff
@@ -185,6 +195,25 @@
...
+			var/fullmin_count = 0
+			for(var/client/C in GLOB.admins)
+				if(check_rights(R_EVENT, 0, C.mob))
+					fullmin_count++
+			if(fullmin_count)
+				GLOB.gamma_request_answered = TRUE
+				GAMMA_Announce(gamma_reason , event_triggered_by, 0)
+				gamma_reason = null
+				SSblackbox.record_feedback("nested tally", "keycard_auths", 1, list("gamma", "called"))
+				spawn(3000)
```
Kyep, line 212: This might be better as a callback or invoke async. We try to avoid using spawns.
Yes, some old code, including the code you are using, still uses them.

code/modules/security_levels/keycard_authentication.dm
```diff
@@ -185,6 +195,25 @@
...
+					if(!GLOB.gamma_request_answered)
+						GAMMA_Announce(gamma_reason , event_triggered_by, 1)
+			else
+				set_security_level(SEC_LEVEL_GAMMA) // No admins? No problem. Automatically approve Gamma Alert.
```
Kyep, line 216: The "automatically approve gamma alert" is... probably going to draw at least one objection.

datlo:
This calls down the gamma armory right? I dont think it should be auto approved when admins are absent. There's often hours for EU folks where admins arent on. I know it's logged but that doesnt do much to help the antags who might be facing much more ridiculous odds in the current round.

Rurik123:
> I dont think it should be auto approved when admins are absent.

Echoing this. Gamma request shouldn't be of the same vein as ERT. Let it auto decline if no admins are on.

Kyep (changes requested):
I'm a bit skeptical of this.
Firstly, because there are some bits where the code may have "src." or spawn() or other stuff we may not like.
Secondly, because gamma alert is usually given for OOC reasons, not IC ones. IE: it doesn't matter how bad the nuke ops are stomping the station, we won't usually give gamma for it as the crew vastly out-number the nuke ops and don't need help against them, especially during war ops. Gamma alert is usually used only for admin events or other threats that are well and truly beyond the crew's normal ability to deal with.
I feel like if a swiper is added for gamma, then at the very minimum it should give some tips as to what is and what is not a legit reason for gamma. Otherwise command is just going to use it like they currently use red - at the drop of a hat.
You could argue its better to have it require two people to swipe to request gamma than the current system where its requested by one person (typically Captain) doing a priority message. You could argue that making two people swipe for it will lead to reduced number of requests. I don't believe that though as command swipes for red just fine and having it as an option will imply to them that they're meant to use it.

Overall, I'm skeptical of this PR.

shmabuu:
Like Kyet said, `Secondly, because gamma alert is usually given for OOC reasons, not IC ones.` when gamma alert is given it’s always by the admin taking a look at what is going on and judging the situation accordingly. They may even give you gamma in alternate to an ERT if you mention it in a request. And most often, just send it themselves when the situation is too dire for command so call for an ERT. Considering how powerful both the SOP and armoury are for this alert... it should at most set the station to gamma alert, but allow the armoury to be sent down at admins digression. I do not believe the crew having both in their hands with a few card swipes is very balanced. I completely understand the want for this when the station gets very messed up and no admins are online; that is why I suggest just giving them the alert level.

S34NW:
> 
> 
> Like Kyet said, `Secondly, because gamma alert is usually given for OOC reasons, not IC ones.` when gamma alert is given it’s always by the admin taking a look at what is going on and judging the situation accordingly. They may even give you gamma in alternate to an ERT if you mention it in a request. And most often, just send it themselves when the situation is too dire for command so call for an ERT. Considering how powerful both the SOP and armoury are for this alert... it should at most set the station to gamma alert, but allow the armoury to be sent down at admins digression. I do not believe the crew having both in their hands with a few card swipes is very balanced. I completely understand the want for this when the station gets very messed up and no admins are online; that is why I suggest just giving them the alert level.

I agree with your sentiments, if you follow the link to the forum from the main post, there is a discussion on decoupling the armoury shuttle from the alert level.


This change doesn't functionally make a difference in when gamma is called, since an admin still needs to approve it anyway. It just makes it easier for the crew since they don't have to write a bloody fax in the middle of an extreme crisis. If command don't understand what gamma is when they see it on their card swipe, they aren't properly fit for that role.

code/modules/security_levels/keycard_authentication.dm
AffectedArc07: Definitely should be an addtimer check here

code/modules/admin/topic.dm
S34NW: indeed, it was doing nothing of value

code/modules/security_levels/keycard_authentication.dm
S34NW: removed it entirely, a 5-minute reminder for this was pointless.

S34NW:
image
Added a clearer warning to the request message.
Also made it so you can only call for gamma alert while already in red alert.

Fox-McCloud:
I'd rather not enshrine Gamma as a player accessible thing, day to day---and certainly not the Gamma armory---if anything, I'd rather move *past* the Gamma armory; ERT already kinda covers the whole "External help from external source". 

I don't like this, from a design perspective at all.

Denthamos:
Echoing above comments from Kyet and Fox.  Gamma is something that players should not expect to get or consider asking for.  Really if anything I have been wondering if instead changing the gamma armory to be something for ERT only would be better, or some sort of proper prep area for the ERT to choose their equipment sort of like TG's ERT prep area.  More incentive for ghosts to get back into the round on crew side instead of only waiting for an antag ghost role to pop up.

Kyep (changes requested):
I'm putting an official objection on this, because I believe gamma alert should stay as something that's initiated by admins, rather than something that is directly requested. I feel like players already over-use red alert, and they'd overuse gamma too if given the chance.

S34NW:
In view of the feedback on this, I'll close it and instead implement a decoupling of the gamma armoury from the alert

PR #15665, opened 2021-03-04
TG Sprite Project: SMES, Supermatter, Rad Collector, and APC

Fox-McCloud:
Updates the Supermatter, SMES, APC, and rad collector sprites

img
img
img
img

:cl: Fox McCloud
imageadd: Updates the Supermatter, rad collector, APC, and SMES sprites
/:cl:


code/modules/power/apc.dm
```diff
@@ -130,6 +131,22 @@
 	req_access = list(ACCESS_SYNDICATE)
 	report_power_alarm = FALSE
 
+/obj/machinery/power/apc/north //Pixel offsets get overwritten on New()
```
Fox-McCloud, line 134: We can start using these instead of var-edited nonsense

Qwertytoforty:
My one complaint, is the SM sprite being cut off at the top.

Fox-McCloud:
>My one complaint, is the SM sprite being cut off at the top.

It glows and pulses a lot; I suspect I caught it between frames.

SabreML:
> > My one complaint, is the SM sprite being cut off at the top.
> 
> It glows and pulses a lot; I suspect I caught it between frames.

It does seem to have a cutoff point there for whatever reason, even with the glowing.
s6GPGGdpf5

S34NW:
would be an easy thumbs up if not for that cut off SM

Carthusia:
God, color that SM shard purple and  youve got the Dark Crystal.

Is it cut off at the top, or is just rounded? It kind of looks like it may be rounded off, but the lack of a glow on the top of the sprite makes it look like its cut off

Spacemanspark:
The SMES sprite feels like a step down from what we already have, if I'm honest, and I'd rather not have it. Everything else looks really dope, though.

SabreML:
I agree with Spark for the SMESes, they look a bit squashed compared to what we have currently:
image NSS Cyberiad 2021-03-05 112137

I'm also not really sure about the radiation collectors. While they are undoubtedly better than what we have right now, I feel like they look a bit too cartoony. Out of everything in this I feel that these stand out the most with their big soft blocks and pastel colours.
image
NSS Cyberiad 2021-03-05 112204

Other than those two and maybe the top of the SM, it all looks good to me!

Kyep (commented):
I'm really not sure I like the sprites in this PR.
The SM sprite looks a bit cooler, but is also so large the top appears cut off. It actually looks a little broken, like it wasn't sized correctly for the maximum sprite dimensions.
The rad collector sprite looks okay. Much brighter than our current ones, but, okay.
The SMES sprite looks kinda pixelated and low quality... but then again so do our current ones. At least our current ones are more detailed (which is probably better) than TG's.

I do very much like having apc/north, apc/south, etc. For mapping purposes, those are great. But equally they could be a separate PR.

Overall, I'm not sure that the sprites here are actually better than what we have now. In fact in some cases (like SMES) they seem to be a downgrade.

matttheficus:
Love the SM sprite, it looks WAY nicer.
The SMES sprite looks like a clear downgrade and the APCs look a bit fuzzy? Might just be the lighting in the picture.

Rad-Collectors look out of place but I'm sure when more TG Sprites are added they'll match better even if I severely dislike cartoon-y art styles in general.

SteelSlayer:
The SMES is whatever, mostly just a more pronounced 3/4ths style, and not an upgrade or downgrade IMO, but still good for the sake of standardization. Everything else is great, and the rad collectors are super clean and high quality. They make the current ones look like ancient coder sprites.

code/modules/power/apc.dm
```diff
@@ -151,20 +168,31 @@
...
 	GLOB.apcs += src
 	GLOB.apcs = sortAtom(GLOB.apcs)
 
 	wires = new(src)
-
+	// offset 24 pixels in direction of dir
+	// this allows the APC to be embedded in a wall, yet still inside an area
+	if(building)
+		setDir(ndir)
+	tdir = dir // to fix Vars bug
```
SteelSlayer, line 183: Can't this just be
```suggestion
	tdir = ndir // to fix Vars bug
```
Since you're setting `dir` to `SOUTH` on the next line. Also what is the "vars bug"?

code/modules/power/apc.dm
```diff
@@ -75,6 +75,7 @@
 	var/locked = 1
 	var/coverlocked = 1
 	var/aidisabled = 0
+	var/tdir = null
```
SteelSlayer, line 78: This isn't very descriptive, but if I had to guess, it means terminal direction?

SteelSlayer (commented):
Sorry I forgot to include this in my review, but got some potentially weird things going on with some sprites.

This APC sprite looks very glitchy.
YGmR6n4KU5

Looks like stray white pixels in the upper left corner in the two images below.
blXHDjNkKW

u2DhOR4Coe

Fox-McCloud:
@SteelSlayer that's just the renderer bugging out; it's fine in game.

code/modules/power/apc.dm
Fox-McCloud: This is mostly for record keeping; if you set dir manually, you're going to wind up with oddities between it being set, at round start for pre-made ones, and ones that you actually build.

code/modules/power/apc.dm
```diff
@@ -151,20 +168,31 @@
 	if(terminal)
 		terminal.connect_to_network()
 
-/obj/machinery/power/apc/New(turf/loc, direction, building = 0)
+/obj/machinery/power/apc/New(turf/loc, ndir, building = 0)
 	if(!armor)
 		armor = list("melee" = 20, "bullet" = 20, "laser" = 10, "energy" = 100, "bomb" = 30, "bio" = 100, "rad" = 100, "fire" = 90, "acid" = 50)
 	..()
 	GLOB.apcs += src
 	GLOB.apcs = sortAtom(GLOB.apcs)
```
dearmochi, line 176: Not your fault but oh dear, this really shouldn't happen during a map load!

S34NW:
Definitely think the rad collectors from #15822 are better than these ones

PR #15673, opened 2021-03-05
Fix crayon drawing invisible runes

### LINKED ISSUE #15670: Crayon "Random Rune" button
The "Random Rune" button on the crayon interface can pick runes 9 and 10 despite there being only eight runes.
When drawn the rune is present on the drawn surface but is invisible.

"Random Rune" to max out at rune8.

**What happened instead**:
Maxes out at rune10.

Minor inconsistency.

1. Pick up crayon
2. Spam "Random rune" until you get rune9 or rune10
3. Draw

Probably after cult runes got their graphical update.

Deelite34:
## What Does This PR Do
Stops crayon from being able to select invisible runes number 9 and 10
fixes #15670

## Changelog
:cl:
fix: Crayon will not create invisible runes anymore
/:cl:

### DIFF
--- a/code/game/objects/items/crayons.dm
+++ b/code/game/objects/items/crayons.dm
@@ -69,7 +69,7 @@
 		if("letter")
 			temp = input("Choose the letter.", "Scribbles") in letters
 		if("random_rune")
-			temp = "rune[rand(1,10)]"
+			temp = "rune[rand(1, 8)]"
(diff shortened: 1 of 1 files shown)


SabreML (approved):
👍

AffectedArc07 (approved):

PR #15682, opened 2021-03-06
Refactors occupancy at machinery-level

dearmochi:
## What Does This PR Do

- Refactors occupancy at machinery-level
  - To use, add `occupy_whitelist = list(/mob/living/carbon/human)` to your machinery
  - By default, any machinery that can be occupied supports drag-and-drop and mouse-drop
  - There may be only one occupant at once
  - `occupy()` and `unoccupy()` return values on success, if you'd like to have behaviour when an occupant is added/removed (e.g. to update icons)
  - Affected machines: Body Scanner, Cryo Pod, Cryo Sleeper, DNA Scanner, Experimentation Machine, Gibber, Mindshield Implanter, Recharging Station, Sleeper, Suit Storage Unit

## Why It's Good For The Game

Each machine currently has its own occupancy system and that's essentially duplicate code. Duplicate code is bad and unifying all of it will greatly streamline further machinery additions down the road.

## Changelog
:cl:
add: Occupying machinery such as sleepers and body scanners now shares the same behaviour for a more consistent and unified experience
spellcheck: Gibber -> gibber, Sleeper -> sleeper
/:cl:


code/__DEFINES/dcs/signals.dm
```diff
@@ -743,3 +743,8 @@
 ///SSalarm signals
 #define COMSIG_TRIGGERED_ALARM "ssalarm_triggered"
 #define COMSIG_CANCELLED_ALARM "ssalarm_cancelled"
+
+// /datum/component/occupant signals
+
+/// (...) - Calls the occupant_act(mob/occupant, ...) callback passed to the component on initialization.
+#define COMSIG_OCCUPANT_ACT "occupant_act"
```
dearmochi, line 750: Shouldn't have made it in

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -494,20 +391,42 @@
...
+			to_chat(M, "<span class='boldnotice'>You feel a cold liquid surround you. Your skin starts to freeze up.</span>")
+		if(M != user)
+			add_attack_logs(user, M, "put into a cryo cell at [COORD(src)].", ATKLOG_ALL)
+		update_icon()
+		M.ExtinguishMob()
+
+/obj/machinery/atmospherics/unary/cryo_cell/unoccupy(mob/user, force)
+	. = ..()
+	if(.)
+		var/mob/living/carbon/M = .
```
SteelSlayer, line 422: To keep consistent with `M` being for `/mob`, `L` for `/mob/living` etc.
```suggestion
		var/mob/living/carbon/C = .
```

SteelSlayer (commented):
Looks good so far. Since its mostly boilerplate stuff, I didn't go super into detail. Will save that for once its out of draft mode.

dearmochi:
Ready for reviewing, but I'd still like a TM to err on the safe side.

code/game/machinery/cryopod.dm
```diff
@@ -574,9 +572,9 @@
 	var/obj/machinery/cryopod/target_cryopod = null
 	if(free_cryopods.len)
 		target_cryopod = safepick(free_cryopods)
+		var/turf/T = get_turf(person_to_cryo)
```
SteelSlayer, line 575: What was the reason for moving this up? It won't be used if the check below fails.

SteelSlayer (commented):
Did some testing. *Somehow* putting yourself into a cryotube puts you into it's pipe network and you can ventcrawl around.

jTuFqoYWAy

code/game/machinery/cryopod.dm
dearmochi: Whoops, moving it back.

SteelSlayer: Ok that makes sense.

PR #15683, opened 2021-03-07
Consciousness transfer potions are now usable on golem shells

MrMagolor:
## What Does This PR Do
You can now use a consciousness transfer potion from xenobio to transfer your mind into a completed golem shell. If you transfer your mind into a servant golem, you will not be made a servant (to prevent any weirdness or effectively self-antagging through mindslaving yourself to someone else).

## Why It's Good For The Game
It only makes sense, as golems can transfer their minds to other shells, and consciousness transfer potions can already transfer minds to things like trees and crates. Furthermore, using brain surgery to transfer your mind to a golem is already possible: this just makes it easier (provided that xenobio makes rainbow slimes).

## Changelog
:cl:
tweak: You can now use consciousness transfer potions to transfer your mind into a finished golem shell.
/:cl:


MrMagolor:
I tried squashing the commits, but that didn't seem to work out. Probably shouldn't've pushed so much.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
 		user.death()
 		return
 
+/obj/effect/mob_spawn/human/golem/attackby(obj/item/slimepotion/transference/I, mob/user, params)
+	. = ..()
+	if(istype(src, /obj/effect/mob_spawn/human/golem/servant))
```
SabreML, line 139: It's more standard to keep the argument the same and typecast it later, since that'll help avoid any issues.
```suggestion
/obj/effect/mob_spawn/human/golem/attackby(obj/item/I, mob/user, params)
	. = ..()
	if(!istype(I, /obj/item/slimepotion/transference))
		return
	if(istype(src, /obj/effect/mob_spawn/human/golem/servant))
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
...
+		src.has_owner = FALSE
+	src.flavour_text = null
```
SabreML, line 141: `src.` is implied here since the proc is on `/obj/effect/mob_spawn/human/golem`, so that can just be left out.
```suggestion
		has_owner = FALSE
	flavour_text = null
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
...
+	if(iscarbon(user) && can_transfer)
+		var/human_transfer_choice = alert("Transfer your soul to [src]? (Warning, your old body will die!)",,"Yes","No")
```
SabreML, line 143: *Maybe* change the `title` argument to `null` here? Doesn't actually make a difference in the end but it's better for readability.
```suggestion
		var/human_transfer_choice = alert("Transfer your soul to [src]? (Warning, your old body will die!)", null,"Yes","No")
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
...
+		if(human_transfer_choice != "Yes")
+			return
+		if(QDELETED(src) || uses <= 0)
+			return
+		log_game("[key_name(user)] used [I] to transfer their mind into [src]")
+		user.visible_message("<span class='notice'>As [user] applies the potion on the golem shell, a faint light leaves them, moving to [src] and animating it!</span>","<span class='notice'>You apply the potion to [src], feeling your mind leave your body!</span>")
```
SabreML, line 149: Another styling thing but this might be neater if it was moved to two lines rather than one.
```suggestion
		user.visible_message("<span class='notice'>As [user] applies the potion on the golem shell, a faint light leaves them, moving to [src] and animating it!</span>",
			"<span class='notice'>You apply the potion to [src], feeling your mind leave your body!</span>")
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
SteelSlayer, line 141: Implicit src
```suggestion
		has_owner = FALSE
	flavour_text = null
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
 		user.death()
 		return
 
+/obj/effect/mob_spawn/human/golem/attackby(obj/item/slimepotion/transference/I, mob/user, params)
```
SteelSlayer, line 137: You need to make sure the object that is attacking is actually a transference potion, otherwise you can just hit the shell with any item to transfer yourself.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
...
+		if(human_transfer_choice != "Yes")
+			return
+		if(QDELETED(src) || uses <= 0)
+			return
+		log_game("[key_name(user)] used [I] to transfer their mind into [src]")
+		user.visible_message("<span class='notice'>As [user] applies the potion on the golem shell, a faint light leaves them, moving to [src] and animating it!</span>","<span class='notice'>You apply the potion to [src], feeling your mind leave your body!</span>")
+		create(ckey = user.ckey, name = user.real_name)
+		user.death()
+		qdel(I)
+		return
```
SteelSlayer, line 153: Unnecessary return
```suggestion
		return
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
...
+	if(iscarbon(user) && can_transfer)
+		var/human_transfer_choice = alert("Transfer your soul to [src]? (Warning, your old body will die!)",,"Yes","No")
+		if(human_transfer_choice != "Yes")
+			return
+		if(QDELETED(src) || uses <= 0)
+			return
+		log_game("[key_name(user)] used [I] to transfer their mind into [src]")
+		user.visible_message("<span class='notice'>As [user] applies the potion on the golem shell, a faint light leaves them, moving to [src] and animating it!</span>","<span class='notice'>You apply the potion to [src], feeling your mind leave your body!</span>")
+		create(ckey = user.ckey, name = user.real_name)
+		user.death()
```
SabreML, line 151: This will show the `"X has died at area."` message to anyone observing, which I assume is probably unintentional.

More importantly, the old body is still revivable via defib. I'd suggest changing this to dust them, or maybe delete the old brain once they've transfered over.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,24 @@
...
+		qdel(I)
+		return
+
```
SabreML, line 154: The return is implied here since it's the end of the proc, so you don't need to specify it.
```suggestion
		qdel(I)

```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
MrMagolor: I thought that placing it that way in the attackby parameters made it so that it only checked for the transference potion?

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
MrMagolor: How would I go about doing the brain deletion? Dusting wouldn't be a good idea since I want them to still have access to their old items.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
SteelSlayer: Parameters themselves have no type checking. You could pass.. say an `/area` into this proc and `I` would be that area. Writing out the full type like you've done, does allow you to do `istype(I)`, instead of 
`istype(I, /obj/item/slimepotion/transference)` though.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
SabreML: This should work I think:
```dm
	user.death()
	QDEL_NULL(user.internal_organs_slot[BRAIN])
	qdel(I)
```
You might even be able to remove the `death()` since deleting the brain should do that for you.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,27 @@
 		user.death()
 		return
 
+/obj/effect/mob_spawn/human/golem/attackby(obj/item/slimepotion/transference/I, mob/living/carbon/user, params)
```
dearmochi, line 137: Not 100% a fan of checking for the item's type this way, because if someone wanted to add another attack behaviour for another item, they wouldn't be able to. Better to leave the argument as `obj/item/I` and pass the desired type in `istype`'s second argument in a `if`/`else if`

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,27 @@
...
+	. = ..()
+	if(!istype(I))
+		return
+	if(istype(src, /obj/effect/mob_spawn/human/golem/servant))
+		has_owner = FALSE
+	flavour_text = null
```
dearmochi, line 143: These will not get reset should the user select "no" afterwards.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,27 @@
...
+	if(iscarbon(user) && can_transfer)
+		var/human_transfer_choice = alert("Transfer your soul to [src]? (Warning, your old body will die!)",null,"Yes","No")
```
dearmochi, line 145:
```suggestion
		var/human_transfer_choice = alert("Transfer your soul to [src]? (Warning, your old body will die!)", null, "Yes", "No")
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,27 @@
...
+		if(human_transfer_choice != "Yes")
+			return
+		if(QDELETED(src) || uses <= 0)
```
dearmochi, line 148: Needs to check if `I` and `user` still exist after the alert.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,27 @@
...
+			return
+		user.visible_message("<span class='notice'>As [user] applies the potion on the golem shell, a faint light leaves them, moving to [src] and animating it!</span>",
+		"<span class='notice'>You apply the potion to [src], feeling your mind leave your body!</span>")
+		message_admins("[key_name(user)] used [I] to transfer their mind into [src]")
+		create(ckey = user.ckey, name = user.real_name)
+		user.death()
+		QDEL_NULL(user.internal_organs_slot[BRAIN])
```
dearmochi, line 155: Brutal. Won't this negatively affect antagonists with a debrain objective?

dearmochi (changes requested):

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,27 @@
 		user.death()
 		return
 
+/obj/effect/mob_spawn/human/golem/attackby(obj/item/slimepotion/transference/I, mob/living/carbon/user, params)
+	. = ..()
```
dearmochi, line 138: The parent should only really be called if specific item types didn't pass.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
MrMagolor: The alternative is (at least according to SabreML): dusting them (which prevents them from getting their ID or any of their other gear), or letting them just die (which theoretically makes the old body revivable)

MrMagolor: The other problem is that adjusting this to make the mind transfer would require a rework of the create() proc in and of itself.

dearmochi: The old body would be revivable through Strange Reagent, and it would result in a catatonic body since the mind was transferred already.

MrMagolor: To be honest, I have no idea how to go about resolving this.

dearmochi: What I mean to say is, I think it's fine if the old body stays intact. Even if it gets revived, they wouldn't be controlled so it hardly matters. If anything, it could help anyone looking for the potion user since they aren't dusted.

SabreML: I hadn't considered antag objectives when I suggested those, but it would probably be better to just leave the body as-is like Mochi suggested.
Any potential weirdness from a catatonic clone is less important than accidentally forcing redtext on a changeling.

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,26 @@
 		user.death()
 		return
 
+/obj/effect/mob_spawn/human/golem/attackby(obj/item/I, mob/living/carbon/user, params)
+	if(!istype(I, /obj/item/slimepotion/transference))
+		. = ..()
+		return
```
SteelSlayer, line 140:
```suggestion
		return ..()
```

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
```diff
@@ -134,6 +134,26 @@
...
+	if(iscarbon(user) && can_transfer)
+		var/human_transfer_choice = alert("Transfer your soul to [src]? (Warning, your old body will die!)", null, "Yes", "No")
+		if(human_transfer_choice != "Yes")
+			return
+		if(QDELETED(src) || uses <= 0 || user.stat >= 1 || QDELETED(I))
+			return
+		if(istype(src, /obj/effect/mob_spawn/human/golem/servant))
+			has_owner = FALSE
+		src.flavour_text = null
```
SteelSlayer, line 149:
```suggestion
		flavour_text = null	
```

MrMagolor:
okay, gonna wait until i get confirmation that git is no longer being broken before i try to close/reopen again

MrMagolor:
Third time's the charm

code/modules/awaymissions/mission_code/ghost_role_spawners/golems.dm
MrMagolor: I think it's settled than that there will be no brain deletion.

dearmochi (approved):

Kyep (approved):
I really don't like golems in general.
However... you're correct that there's really no logical reason for this NOT to work.
So, approved.

AffectedArc07 (approved):

Issue #15609, opened 2021-02-25
area creation log location is wrong.


TDSSS:
**Issue Description**:
The creation of new rooms/areas is logged. However, the coordinates attached to the message admins get about that is always 0,0,0, which is wrong.

**What did you expect to happen**: 
Actual coordinates

**What happened instead**:
You get a message like
ADMIN LOG: A new room was made by [Some CKEY]/(IC NAME(?) (FLW)(?) (FLW) at Shitty DNA Vault (0,0,0) (JMP) with the name Shitty DNA Vault

(also the ? and FLW links are both duplicated)

**Why is this bad/What are the consequences**:
coordinates should be correct

**Steps to reproduce the problem**:
make new area, be admin, look at message

**When did the problem start happening**:
once the logging feature was put in, in #15525

**Extra information**:

SabreML:
Fixed by the above PR

PR #15689, opened 2021-03-08
Fix coordinates logging when new area is made, and excesive second pair of ? and FLW

Deelite34:
## What Does This PR Do

Fixes both problems from #15609
Correct new area location is logged/shown to admins after new area has been made with station blueprints instead of [0,0,0]. Duplicates of ? and FLW will not be shown anymore.

## Images of changes

image

## Changelog
:cl:
fix: Fix incorrect coordinates being logged when area is made, and excesive ? and FLW appearing
/:cl:

### DIFF
--- a/code/game/objects/items/blueprints.dm
+++ b/code/game/objects/items/blueprints.dm
@@ -228,8 +228,8 @@
 		FD.CalculateAffectingAreas()
 
 	interact()
-	message_admins("A new room was made by [ADMIN_LOOKUPFLW(usr)] at [ADMIN_VERBOSEJMP(usr)] with the name [str]")
-	log_game("A new room was made by [key_name(usr)] at [AREACOORD(usr)] with the name [str]")
+	message_admins("A new room was made by [key_name_admin(usr)] at [ADMIN_VERBOSEJMP(usr)] with the name [str]")
+	log_game("A new room was made by [key_name_admin(usr)] at [AREACOORD(usr)] with the name [str]")
 	area_created = TRUE
 	return area_created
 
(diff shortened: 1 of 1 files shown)


dearmochi (approved):

code/game/objects/items/blueprints.dm
```diff
@@ -228,8 +228,8 @@
 		FD.CalculateAffectingAreas()
 
 	interact()
-	message_admins("A new room was made by [ADMIN_LOOKUPFLW(usr)] at [ADMIN_VERBOSEJMP(src)] with the name [str]")
-	log_game("A new room was made by [key_name(usr)] at [AREACOORD(src)] with the name [str]")
+	message_admins("A new room was made by [key_name_admin(usr)] at [ADMIN_VERBOSEJMP(usr)] with the name [str]")
+	log_game("A new room was made by [key_name_admin(usr)] at [AREACOORD(usr)] with the name [str]")
```
farie82, line 232:
```suggestion
	log_game("A new room was made by [key_name(usr)] at [AREACOORD(usr)] with the name [str]")
```
just `key_name` here will do since this will be logged to a file and thus should not have hrefs

code/game/objects/items/blueprints.dm
```diff
@@ -250,8 +250,8 @@
 			FD.CalculateAffectingAreas()
 	to_chat(usr, "<span class='notice'>You rename the '[prevname]' to '[str]'.</span>")
 	interact()
-	message_admins("A room was renamed by [ADMIN_LOOKUPFLW(usr)] at [ADMIN_VERBOSEJMP(usr)] changing the name from [prevname] to [str]")
-	log_game("A room was renamed by [key_name(usr)] at [AREACOORD(usr)] changing the name from [prevname] to [str] ")
+	message_admins("A room was renamed by [key_name_admin(usr)] at [ADMIN_VERBOSEJMP(usr)] changing the name from [prevname] to [str]")
+	log_game("A room was renamed by [key_name_admin(usr)] at [AREACOORD(usr)] changing the name from [prevname] to [str] ")
```
farie82, line 254:
```suggestion
	log_game("A room was renamed by [key_name(usr)] at [AREACOORD(usr)] changing the name from [prevname] to [str] ")
```
Same here

farie82 (changes requested):

code/game/objects/items/blueprints.dm
Deelite34: At first I wasn't sure wheter logs are for outside game use or for in game viewing, I assumed latter, but it seems I was wrong.

farie82: Understandable.
The ingame log viewer currently uses the following proc for that:
`usr.create_log(LOG_TYPE, message, target, where)` where ` target` and `where` are optional parameters. Where is auto filled if left empty

farie82 (approved):

PR #15699, opened 2021-03-10
Adds in-hand sprites to xray and mindflayer guns

S34NW:
## What Does This PR Do
This PR codes in the existing in-hand sprites for the x-ray gun that were sat in the files unused, and changes the gun to use the shaded charge system.
It also recolours the mindflayer gun to be purple/pink, to more clearly differentiate the two weapons. The mindflayer was previously a visual carbon copy of the x-ray gun. (not that it gets used or seen all that often currently)
The mindflayer also now uses energy projectile as opposed to beam, so it cannot ricochet, and doesn't do 20 burn damage on top of its large brain/hallucination damage. (For clarification, being a beam projectile did not make it a beam in the way that the medibeam gun works.)

## Why It's Good For The Game
Making use of existing sprites is good, as well as moving old ones to new systems such as shaded charge.
Also, guns looking like twigs because they don't have a custom in-hand is bad. No more twig guns please.

## Images of changes
image image screenrecording
[Video to see mindflayer firing](https://user-images.githubusercontent.com/12197162/110713046-286cf080-81f9-11eb-82a7-93106b71d35d.mp4)

## Changelog
:cl:
add: In-hand sprites for x-ray gun
tweak: Recoloured mindflayer gun, changed its projectile type from beam to energy
/:cl:


MrMagolor:
This'll probably be useful for blobs to tell who has x-rays: then again, it'd probably be hard to focus on individual people during the slaughterhouse that is a blob warzone.

code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -106,6 +106,8 @@
 	name = "xray laser gun"
 	desc = "A high-power laser gun capable of expelling concentrated xray blasts. These blasts will penetrate solid objects, but will decrease in power the longer they have to travel."
 	icon_state = "xray"
+	item_state = null
+	shaded_charge = 1
```
dearmochi, line 110:
```suggestion
	shaded_charge = TRUE
```

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -85,10 +85,10 @@
 /obj/item/gun/energy/mindflayer
 	name = "\improper Mind Flayer"
 	desc = "A prototype weapon recovered from the ruins of Research-Station Epsilon."
-	icon_state = "xray"
+	icon_state = "flayer"
 	item_state = null
+	shaded_charge = 1
```
dearmochi, line 90:
```suggestion
	shaded_charge = TRUE
```

code/modules/projectiles/projectile/special.dm
```diff
@@ -141,10 +141,14 @@
 	nodamage = 1
 	flag = "energy"
 
-/obj/item/projectile/beam/mindflayer
+/obj/item/projectile/energy/mindflayer
```
dearmochi, line 144: This seems like an undocumented change

code/modules/projectiles/projectile/special.dm
```diff
@@ -141,10 +141,14 @@
...
 	name = "flayer ray"
+	icon_state = "flayer"
+	damage = 0
+	impact_effect_type = /obj/effect/temp_visual/impact_effect/purple_laser
+	light_color = LIGHT_COLOR_PURPLE
 
-/obj/item/projectile/beam/mindflayer/on_hit(var/atom/target, var/blocked = 0)
+/obj/item/projectile/energy/mindflayer/on_hit(var/atom/target, var/blocked = 0)
```
dearmochi, line 151: Ditto

code/modules/projectiles/projectile/special.dm
S34NW: oops, i forgot to mention this, thanks for pointing it out

dearmochi (approved):

SteelSlayer (approved):

Kyep (approved):
I really like this. I may end up tweaking the mindflayer guns later in a subsequent PR for the sake of an outfit that uses them... but this is a good start for both weapons.

AffectedArc07 (approved):

PR #15706, opened 2021-03-12
Revert "Fixes input being lost while holding keys down and pressing others"

AffectedArc07:
Reverts ParadiseSS13/Paradise#13466

It broke movement mid round and caused SSinput to sleep.


AffectedArc07:
Reopening to fix CI

AffectedArc07:
Declaring this not a fix as this code doesnt have anywhere it can even sleep inside of it, and the proc itself has `waitfor = FALSE`, so I have no idea.

Chalking it up to solar flares affecting memory since a 9900k doesnt support ECC ram.

PR #15712, opened 2021-03-13
Restricts "Delete All Security Records" to Magistrate or Armory access.

S34NW:
## What Does This PR Do

This PR means only ID's with magistrate or armory access can use the "Delete All Security Records" option.
Regular sec ID's can still delete individual records.

## Why It's Good For The Game
Stealing an ID to delete your (or a friend/enemy's) sec record is a valid antag tactic, and should not be changed.
However, the only thing that the option to delete all records does is harm security's ability to do anything useful with their HUDs until the records are manually re-created, one by one. This can take one user a significant amount of time, and the antagonist can just delete them all again in an instant.
Having it restricted to Magistrate/HoS/Warden/Captain/HoP (or any other ID with magistrate/armory access) will reduce the amount of cheese and "dickish" behaviour, but still leave it open for antags who can get their hands on such an ID.
It also does not stop antags that want to spend the time deleting records one by one.

## Changelog
:cl:
tweak: "Delete All Security Records" now requires magistrate or armory access.
/:cl:

### DIFF
--- a/code/game/machinery/computer/security.dm
+++ b/code/game/machinery/computer/security.dm
@@ -248,14 +248,19 @@
 			update_all_mob_security_hud()
 			set_temp("Security record deleted.")
 		if("delete_security_all") // Delete All Security Records
+			var/datum/ui_login/state = ui_login_get()
 			if(!logged_in)
 				return
-			for(var/datum/data/record/S in GLOB.data_core.security)
-				qdel(S)
-			message_admins("[key_name_admin(usr)] has deleted all security records at [ADMIN_COORDJMP(usr)]")
-			usr.create_log(MISC_LOG, "deleted all security records")
-			update_all_mob_security_hud()
-			set_temp("All security records deleted.")
+			if((ACCESS_MAGISTRATE in state.access) || (ACCESS_ARMORY in state.access))
+				for(var/datum/data/record/S in GLOB.data_core.security)
+					qdel(S)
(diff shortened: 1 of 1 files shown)


Rurik123:
Wait there's a button that deletes all sec records with a click? Here I thought antags were removing each individual record one by one, just to make my life harder...

Dafrek:
Honestly, I don't understand why anyone on the station would have the option to delete all security records with a click of a button. Unless I'm missing something it just doesn't make sense IC why even the captain or magistrate would be able to just delete all the records, it would make more sense if this was restricted to CC personnel.

S34NW:
removing the option entirely is easy enough too, but uh _improve don't remove_.
If enough people think it should go, I'm happy to PR that

hal9000PR:
> removing the option entirely is easy enough too, but uh _improve don't remove_.
> If enough people think it should go, I'm happy to PR that

I'd say just remove them button entirely. Antags can still do it but it requires extensive effort. There is really no IC reason for that button to exist.

matttheficus:
Another possible way to make this nice is to make the "Backup Sec Records" button actually do something via a Data Disk instead of it just being there and have zero use.

dearmochi (approved):

Kyep (approved):

AffectedArc07 (approved):
This cheese is awful

Issue #9951, opened 2018-10-23
The RPD's 'orient automatically' setting doesn't work on certain objects


SteelSlayer:
**Problem Description**:
The following items from the Rapid Pipe Dispenser don't work with the 'orient automatically' setting:
- unary vent fitting
- scubber fitting
- passive vent
- connector
- air injector
- all pipe caps (normal, supply, scrubber)

**What did you expect to happen**: 

With the orient automatically setting, this is how the objects should get placed, with the pipe that sticks out facing toward the player:
![rpd issue2](https://user-images.githubusercontent.com/42044220/47329897-519a7f80-d644-11e8-8f89-2d1d5143dcb9.jpg)

**What happened instead**:

Instead, the objects get placed like this, with the pipe facing upwards:
![rpd issue1](https://user-images.githubusercontent.com/42044220/47329878-45aebd80-d644-11e8-9c0e-63adf4806bc5.jpg)

**Why is this bad/What are the consequences:**
Not a very pressing issue, but its a bug nonetheless.

**Steps to reproduce the problem**:
Pick up any RPD, select 'orient automatically' and try to place any one of the objects listed above.

PR #15714, opened 2021-03-13
Vents and Scrubbers can be placed 180 degrees on each other, with a max of 4 total each tile

### LINKED ISSUE #15713: Atmos pipe devices take up multiple directions.
**Problem Description**:
What is the problem?
vents and scrubbers take up 2 directions on a tile instead of 1. Example: A vent piped going up also takes up the down pipe position.
They should only take up 1 position, allowing you to wrench pipes and pipe devices in the opposite direction on the tile as well.
**What happened instead**:
You can only put 1 vent/scrubber facing up/down, and one facing left/right
They used to be set so that you can for example place 4 vents facing 4 different directions on a single tile.
Place a vent/scrubber in one direction and wrench it. Place another in the opposite direction and try to wrench it.

It worked when I used to make my turbine TEG setups, I used single tiles with 3 vents and a scrubber on it to bypass pressure restrictions that volume pumps had.

dreamseeker_NCkO9t7odf
dreamseeker_3xd54EEDwC

Kepplerx:
Fixes an unintended sideeffect of #9951
Fixes #15713 

## What Does This PR Do

Vents and scrubbers can now be on the same tile, opposed to each other, instead of max 2 on a tile

## Why It's Good For The Game

This was a unintended change, this would revert that, makes many Atmos related constructs more viable, and possible to make.

## Images of changes

image

## Changelog
:cl:
fix: Fixed vents not being able to be put 180 degrees on each other
/:cl:


code/game/machinery/pipe/construction.dm
```diff
@@ -256,8 +256,10 @@
 			return dir|flip
 		if(PIPE_SIMPLE_BENT, PIPE_INSULATED_BENT, PIPE_HE_BENT, PIPE_SUPPLY_BENT, PIPE_SCRUBBERS_BENT)
 			return dir //dir|acw
-		if(PIPE_CONNECTOR, PIPE_UVENT, PIPE_PASV_VENT, PIPE_SCRUBBER, PIPE_HEAT_EXCHANGE, PIPE_INJECTOR)
+		if(PIPE_CONNECTOR,  PIPE_HEAT_EXCHANGE, PIPE_INJECTOR)
 			return dir|flip
+		if(PIPE_UVENT, PIPE_PASV_VENT, PIPE_SCRUBBER)
+			return dir
```
SteelSlayer, line 262: Can you not put these inside the `if` on line 257?

Kepplerx: Why not indeed, let me actually go ahead and do that

Kepplerx: Oh, it was to make clear the difference between 257 which was bent pipes, and how they also are applied the acw/cw, and vents and scrubbers which do not @SteelSlayer

SteelSlayer: I'm really just referring to the return values. Both the new `if` you've added and the check on 257 does a `return dir`, therefore they should probably be combined.

warior4356:
This is really going to be abused in the SM.

DinBlade:
> This is really going to be abused in the SM.

Rad collectors are meh at best, so unless someone is doing Rad related antaging or delams, SM is not the issue this makes. TEGs tho and what little I know of them, this is a gamechanger.

DesolateG:
> > This is really going to be abused in the SM.
> 
> Rad collectors are meh at best, so unless someone is doing Rad related antaging or delams, SM is not the issue this makes. TEGs tho and what little I know of them, this is a gamechanger.

It's not anything new, just a fix of a bug that no one noticed until I tried making one of my old TEG setups.

Fox-McCloud:
This seems unwise to do, at a glance.

DesolateG:
> This seems unwise to do, at a glance.

Could you clarify?

Arvenius169:
I was always under the impression that this had been an intentional decision to reduce crazy atmos device spam.

DesolateG:
> I was always under the impression that this had been an intentional decision to reduce crazy atmos device spam.

It was a bug caused by Mitch's PR.

Arvenius169:
> This is really going to be abused in the SM.

SM almost never has the draft touched. I don't think that's a legitimate worry.

code/game/machinery/pipe/construction.dm
variableundefined: Probably won't make a difference when it comes to performance.

variableundefined (approved):

PR #15716, opened 2021-03-14
Kisses no longer break operatives anti-love shields

Qwertytoforty:
## What Does This PR Do

Shielded hardsuits no longer take damaged from projectiles with the nodamage flag, that do not stun. (IE, kisses, laser tag, magic wand). Disablers and tasers still damage the shield as before.

## Why It's Good For The Game

You should not be able to disable a nuclear operatives hardsuit with kisses.

## Changelog
:cl:
fix: Nerfs the power of love on shielded hardsuits, kisses no longer lower the shields of shielded hardsuits.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult_items.dm
+++ b/code/game/gamemodes/cult/cult_items.dm
@@ -161,7 +161,12 @@
 /obj/item/clothing/suit/hooded/cultrobes/cult_shield/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the attack", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
 	if(current_charges)
 		owner.visible_message("<span class='danger'>[attack_text] is deflected in a burst of blood-red sparks!</span>")
-		current_charges--
+		if(istype(hitby, /obj/item/projectile))
+			var/obj/item/projectile/P = hitby
+			if(!P.nodamage)
+				current_charges--
+		else
+			current_charges--
 		playsound(loc, "sparks", 100, TRUE, SHORT_RANGE_SOUND_EXTRARANGE)
 		new /obj/effect/temp_visual/cult/sparks(get_turf(owner))
 		if(!current_charges)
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/cult/cult_items.dm
```diff
@@ -161,7 +161,12 @@
...
 		owner.visible_message("<span class='danger'>[attack_text] is deflected in a burst of blood-red sparks!</span>")
-		current_charges--
+		if(istype(hitby, /obj/item/projectile))
+			var/obj/item/projectile/P = hitby
+			if(!P.nodamage)
+				current_charges--
+		else
+			current_charges--
 		playsound(loc, "sparks", 100, TRUE, SHORT_RANGE_SOUND_EXTRARANGE)
 		new /obj/effect/temp_visual/cult/sparks(get_turf(owner))
```
SabreML, line 171: You could maybe do this instead to avoid the duplicate code, since nothing past line 167 is going to happen anyway if it does no damage.
Although if you do, then you should move it down a couple of lines so that the sparks still happen.
```suggestion
		owner.visible_message("<span class='danger'>[attack_text] is deflected in a burst of blood-red sparks!</span>")
		playsound(loc, "sparks", 100, TRUE, SHORT_RANGE_SOUND_EXTRARANGE)
		new /obj/effect/temp_visual/cult/sparks(get_turf(owner))
		if(istype(hitby, /obj/item/projectile))
			var/obj/item/projectile/P = hitby
			if(P.nodamage)
				return TRUE
		current_charges--
		if(!current_charges)
```

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -628,10 +628,18 @@
...
-		if(recharge_rate)
+		if(istype(hitby, /obj/item/projectile))
+			var/obj/item/projectile/P = hitby
+			if(!P.nodamage)
+				current_charges--
+				was_damaged = TRUE
+		else
+			current_charges--
+			was_damaged = TRUE
+		if(recharge_rate && was_damaged)
```
SabreML, line 642: Same here:
(Returning `TRUE` on both so that the projectile doesn't go through)
```suggestion
	if(current_charges > 0)
		do_sparks(2, 1, src)
		owner.visible_message("<span class='danger'>[owner]'s shields deflect [attack_text] in a shower of sparks!</span>")
		if(istype(hitby, /obj/item/projectile))
			var/obj/item/projectile/P = hitby
			if(P.nodamage)
				return TRUE
		current_charges--
		if(recharge_rate)
```

Carthusia:
You could say that showing the syndicate some love...lowered their defences (Ba-dum-tss)

Qwertytoforty:
Kisses are removed, PR not needed.

PR #15720, opened 2021-03-15
fixes firelocks et al. preventing the movement of air

mafemergency:
certain objects like firelocks, single-pane windows, windoors, and potentially others would prevent the movement of air until they were "woken up" by opening/closing the firelock, breathing on it or removing/adding the floor tile underneath.

image attached demonstrates this, with the plasma being very secure and stable within its region and never moving passed the firelock.

fix_atmos_firelocks

:cl: anon
fix: firelocks and others no longer prevent movement of air until "awoken" after server intializations
/:cl:

### DIFF
--- a/code/controllers/subsystem/air.dm
+++ b/code/controllers/subsystem/air.dm
@@ -284,14 +284,8 @@ SUBSYSTEM_DEF(air)
 			add_to_active(S)
 
 /datum/controller/subsystem/air/proc/setup_allturfs(list/turfs_to_init = block(locate(1, 1, 1), locate(world.maxx, world.maxy, world.maxz)))
-	var/list/active_turfs = src.active_turfs
-
-	// Clear active turfs - faster than removing every single turf in the world
-	// one-by-one, and Initalize_Atmos only ever adds `src` back in.
-	active_turfs.Cut()
-
-	for(var/thing in turfs_to_init)
-		var/turf/T = thing
+	for(var/turf/T as anything in turfs_to_init)
+		remove_from_active(T)
 		if(T.blocks_air)
 			continue
 		T.Initialize_Atmos(times_fired)


code/controllers/subsystem/air.dm
```diff
@@ -284,14 +284,8 @@ SUBSYSTEM_DEF(air)
...
 /datum/controller/subsystem/air/proc/setup_allturfs(list/turfs_to_init = block(locate(1, 1, 1), locate(world.maxx, world.maxy, world.maxz)))
-	var/list/active_turfs = src.active_turfs
-
-	// Clear active turfs - faster than removing every single turf in the world
-	// one-by-one, and Initalize_Atmos only ever adds `src` back in.
-	active_turfs.Cut()
-
-	for(var/thing in turfs_to_init)
-		var/turf/T = thing
+	for(var/turf/T in turfs_to_init)
```
Fox-McCloud, line 287: `block` can only return a list of of turfs, so this would be better to keep as a typeless list.

mafemergency: i've tested three variations on 514.1549 and was unable to determine if any is faster than the other

```dm
/datum/controller/subsystem/air/proc/setup_allturfs(list/turfs_to_init = block(locate(1, 1, 1), locate(world.maxx, world.maxy, world.maxz)))
	for(var/turf/T in turfs_to_init)
		remove_from_active(T)
		if(T.blocks_air)
			continue
		T.Initialize_Atmos(times_fired)
		//CHECK_TICK

/datum/controller/subsystem/air/proc/setup_allturfs2(list/turfs_to_init = block(locate(1, 1, 1), locate(world.maxx, world.maxy, world.maxz)))
	for(var/T in turfs_to_init)
		remove_from_active(T)
		if(T:blocks_air)
			continue
		T:Initialize_Atmos(times_fired)
		//CHECK_TICK

/datum/controller/subsystem/air/proc/setup_allturfs3(list/turfs_to_init = block(locate(1, 1, 1), locate(world.maxx, world.maxy, world.maxz)))
	for(var/t in turfs_to_init)
		var/turf/T = t
		remove_from_active(T)
		if(T.blocks_air)
			continue
		T.Initialize_Atmos(times_fired)
		//CHECK_TICK
```

Fox-McCloud: Typed loops implicitly do an `istype` check on every single thing in their list, which adds to total CPU; for some things, it's going to be difficult to profile, but it's there.

If a loop is all of one particular type (or subtypes) and you intend to act on every single one, we go with typeless lists for performance reasons.

see: https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md istypeless loops.

mafemergency:
would this be acceptable? it goes by latest info from https://github.com/tgstation/tgstation/blob/master/.github/CONTRIBUTING.md#user-content-avoid-unnecessary-type-checks-and-obscuring-nulls-in-lists but for what it's worth the difference after millions of iterations is barely measurable and i'm dubious of the merits of these types of optimizations

code/controllers/subsystem/air.dm
```diff
@@ -284,14 +284,8 @@ SUBSYSTEM_DEF(air)
...
 /datum/controller/subsystem/air/proc/setup_allturfs(list/turfs_to_init = block(locate(1, 1, 1), locate(world.maxx, world.maxy, world.maxz)))
-	var/list/active_turfs = src.active_turfs
-
-	// Clear active turfs - faster than removing every single turf in the world
-	// one-by-one, and Initalize_Atmos only ever adds `src` back in.
-	active_turfs.Cut()
-
-	for(var/thing in turfs_to_init)
-		var/turf/T = thing
+	for(var/turf/T as anything in turfs_to_init)
```
Fox-McCloud, line 287: We're not going to be utilizing `as anything` for now, so please stick with the typeless format.

Fox-McCloud:
Please update or this will be closed.

variableundefined (changes requested):
Giving this two more days

Fox-McCloud:
No response. Closing.

PR #15726, opened 2021-03-17
Prevents Voice of God from being used while in jaunt, fixes cult VOG

Qwertytoforty:
## What Does This PR Do

Prevents speaking / using VOG while in jaunt by checking for if they are in jaunt, fixes cult voice of god not giving the right span.

## Why It's Good For The Game

No, you should not be able to tell people to sleep while mid jaunt, with no way to see or hear it coming.
Bug fixes for cult is good.

## Changelog
:cl:
tweak: Voice of God can not be used while the user is jaunting.
fix: VOG now properly uses cult text when a cultist is using them.
/:cl:


dearmochi:
Perhaps move the check to `can_speak()` (or add/remove `TRAIT_MUTE` on jaunt start/end)? It would prevent the jaunter from speaking completely while jaunting.

Kyep (approved):

AffectedArc07 (approved):

PR #15729, opened 2021-03-18
Contribution guidelines: type checks in lists

dearmochi:
# Original PR: tgstation/tgstation#56761

## What Does This PR Do

- Updates the contribution guidelines to include the as anything key to skip checks in lists.
- Reduces the explanation a bit, to make it more succinct.
- Makes the examples use default byond types and vars, and code that actually compiles, instead of pseudo-code. Plus early continues, because they're good.

## Why It's Good For Developers

Invalidates the pattern of doing:
```dm
for(var/r in radios)
	var/obj/item/radio/radio = r
	radio.stuff()
```
by instead doing:
```dm
for(var/obj/item/radio/radio as anything in radios)
	radio.stuff()
```

This results in cleaner code IMO. Additionally, `null` values are preserved which may help in debugging cases where lists shouldn't contain such values.

## Changelog
N/A

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -432,7 +432,32 @@ SS13 has a lot of legacy code that's never been updated. Here are some examples 
      var/user = interactor
      fillReserves(user)
      ```
+### Avoid unnecessary type checks and obscuring nulls in lists
+Typecasting in `for` loops carries an implied `istype()` check that filters non-matching types, nulls included. The `as anything` key can be used to skip the check.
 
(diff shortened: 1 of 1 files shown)


SabreML:
Unless I'm mistaken all three loop methods will have more or less the same performance impact in v514. Is it really worth changing the standard for a good chunk of the codebase now, when we're presumably going to update in the near future?
(Let me know if I _am_ mistaken because I haven't been tracking the 514 update much.)

dearmochi:
Tests seem to point toward that, at least there have been changes that greatly reduce the difference between any looping method, so I'll close this PR

PR #15730, opened 2021-03-18
Converts MyISAM tables to InnoDB

AffectedArc07:
## What Does This PR Do
Small PR before I vanish for 2 weeks. This kills off our MyISAM tables in favour of InnoDB ones. InnoDB has several benefits over MyISAM:

- MyISAM is no longer maintained, InnoDB is
- Support for transactions
- You can lock at row-level as opposed to locking the entire table
- More resistant to table corruption
- Allows for foreign key relationships (we likely wont do this but its better to support it)
- Larger buffer for better performance

The only disadvantage is MyISAM uses less storage space than InnoDB, but of the affected tables there shoulndt be a major increase

- `customuseritems` - This does not even have many rows, especially compared to existing InnoDB datasets like `player`
- `death` - Log of all deaths. The MyISAM version is 170mb, and the converted one is 171mb.
- `feedback` - I already compacted this loads and loads
- `karma` - This is the entire log of karma purchases. The existing MyISAM table is 47mb, and the conversion is 49mb
- `karmatotals` - Just a log of the total amount of karma people have. Shouldnt increase too much.
- `legacy_population` - This shouldnt be a major increase, rows older than the past 6 months can likely be archived,
- `library` - This should not cause major space increases.
- `whitelist` - This is karma purchases oddly enough, and again, this shouldnt increase much

Requesting @Kyep give thoughts on these.

## Why It's Good For The Game
Using an inferior, unmaintained table structure is bad

## Changelog
:cl: AffectedArc07
add: Made major optimisations to the DB schema 
/:cl:


tgui/log.txt
```diff
@@ -0,0 +1,293 @@
+yarn install v1.22.4
```
AffectedArc07, line 1: Oh lord this wasnt meant to be staged

Kyep (approved):
Signing off on this.
While restructuring a bunch of our tables isn't something I look forward to, AA's arguments in favor of doing this are solid.

dearmochi (approved):

PR #15735, opened 2021-03-19
Prevent disposing things to the Captain's office in Boxstation with just one pipe or a chute

Vi3trice:
## What Does This PR Do
Adds a sorting pipe within the confines of the bridge to filter out unwanted trash.

As it turns out, disposals can do a full 180 and go the opposite direction, so you can put sorting pipes without worrying about bursting floor tiles. Delta has a less robust version of that particular fail-safe mapped in, but with far too many unused pipes. This only uses one.

## Why It's Good For The Game
You can still screw with the captain or other people in some ways, but that will require more than just one pipe swap.

## Images of changes
Box-CapShip-New
Box-CapShip-Old

## Changelog
:cl:
tweak: Replaced one pipe leading toward the Captain's office with a sorting junction pipe
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -31335,8 +31335,8 @@
 	},
 /obj/effect/spawner/window/reinforced,
 /obj/structure/cable{
-	icon_state = "0-4";
-	d2 = 4
+	d2 = 4;
+	icon_state = "0-4"
 	},
 /obj/machinery/status_display{
 	density = 0;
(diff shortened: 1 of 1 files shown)


TDSSS:
I disagree with this change. The station is supposed to have weaknesses and disposal flipping is one way to break into the captain's office. It's not like it's trivial either, wrenching up the floor in a busy hallway has a decent chance to get noticed.

hal9000PR:
One major antag strat is rerouting disposals into the captains office to break in and steal the medal/gun/jetpack. this just makes this impossbile. and it won't stop tidery, now i can just reroute disposals onto the bridge and flood the bridge with shit instead of the caps office. this PR does nothing except remove interesting antag strats

PR #15750, opened 2021-03-23
Adds blood drain limit, removes gender from vampires

Qwertytoforty:
## What Does This PR Do

This PR adds mobs vampires drains to a list, and tracks how much blood is drained from a vampire. If a vampire has drained more than 250 usable blood from a person, they will no longer gain useable blood from draining them. They will still be able to drain the person of blood and gain nutrition, just no more usable blood.

Also removes gender from vampires. Not sure why it was in there, but it was not used.

## Why It's Good For The Game

Removing unused gender code is good.

Vampires are in an interesting place. An antagonist that kidnaps multiple people, slowly getting stronger, before getting full power and becoming a nightmare for security. This is fine. It is balanced around vampires kidnapping multiple people, which is risky, and makes you more likely to get noticed by security. The issue is when vampires capture someone / find a friend, and use them, and 3-4 blood bags, to get full power, with almost no risk to themself. This makes getting full power more easy then it is intended to be, and encourages vampires to get friends to sit still, take an IV drip, and let them drain blood till they get full power.

This PR makes it so vampires can only get 250 units of usable blood from someone, before they can get no more. Lore wise, it makes sense, vampires are after your lifeforce, and get it through your blood. People logically probably don't have infinite life force, and you probably cant add more lifeforce by stabbing them with a needle and inserting more blood into them. Vampires still have uses for bloodbags, 250 units is more than the usual 220 units you can drain from someone before they die, if you want to keep a thrall alive, or if vampires are working together, they can use the bags to ensure all vampires get 250 units from one person, before they die.

## Changelog
:cl:
tweak: Vampires now can only get 250 units of usable blood from a person, afterwords they only gain nutrition.
/:cl:

### DIFF
--- a/code/game/gamemodes/vampire/vampire.dm
+++ b/code/game/gamemodes/vampire/vampire.dm
@@ -199,7 +199,6 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 	var/bloodtotal = 0 // CHANGE TO ZERO WHEN PLAYTESTING HAPPENS
 	var/bloodusable = 0 // CHANGE TO ZERO WHEN PLAYTESTING HAPPENS
 	var/mob/living/owner = null
-	var/gender = FEMALE
 	var/iscloaking = 0 // handles the vampire cloak toggle
 	var/list/powers = list() // list of available powers and passives
 	var/mob/living/carbon/human/draining // who the vampire is draining of blood
@@ -219,9 +218,8 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 		/obj/effect/proc_holder/spell/vampire/self/jaunt = 300,
 		/obj/effect/proc_holder/spell/vampire/targetted/enthrall = 300,
 		/datum/vampire_passive/full = 500)
+	var/list/drained_humans = list() //This is a list of people the vampire has drained, and how much blood has been drained from them.
 
-/datum/vampire/New(gend = FEMALE)
-	gender = gend
 
 /datum/vampire/proc/force_add_ability(path)
 	var/spell = new path(owner)
@@ -257,7 +255,7 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 		return
 	var/datum/vampire/vamp
 	if(!mind.vampire)
-		vamp = new /datum/vampire(gender)
+		vamp = new /datum/vampire
 		vamp.owner = src
 		mind.vampire = vamp
 	else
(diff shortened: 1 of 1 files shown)


Samman166:
I like it

code/game/gamemodes/vampire/vampire.dm
```diff
@@ -219,9 +218,8 @@ You are weak to holy things and starlight. Don't go into space and avoid the Cha
 		/obj/effect/proc_holder/spell/vampire/self/jaunt = 300,
 		/obj/effect/proc_holder/spell/vampire/targetted/enthrall = 300,
 		/datum/vampire_passive/full = 500)
+	var/list/drained_humans = list() //This is a list of people the vampire has drained, and how much blood has been drained from them.
```
Fox-McCloud, line 221: Leaving hard refs in someone's mind is something that I will absolutely not allow.

FlattestGuitar:
Remove the irrelevant changelog entry.

code/game/gamemodes/vampire/vampire.dm
FlattestGuitar: You can move this to mind.dm, you're in good company there with weird antag shit. Though then you don't track the succ per-vampire, which might not be what you're going for here.

Rurik123:
Is the limit renewed upon being revived? All for the removal of cheese but valid hunters/lone officers that suicide into maint alone multiple times after a vamp should be able to succ'd more than once, lest you encourage the vamp to permakill them.

Coolrune206:
Can the drain limit perhaps "decay" after a certain time? Like, say, thirty minutes after the limit starts to fill from draining a target, the numbers will tick down at the same rate they were acquired? I liked the idea of a Vampire spending some blood to thrall someone, then being able to drain them in the future as a sort of renewable source of blood - they have to invest 300 blood to thrall the target (I believe), so only being able to get 250 out of that target would turn the investment into somewhat of a loss at the moment. I really like the concept of Vampires using thralls, I feel they're very much underused at the moment, so I think leaving an exception like this that would leave both draining thralls in the long term viable, while also addressing the point Rurik made. At the same time, it probably wouldn't reward the blood-bag/metafriend gameplay, unless a vampire is willing to sit kidnapped with the same person for thirty minutes (at which point the behavior would become obvious for an admeme to interject).

dearmochi (commented):
Seeing a lot of magic numbers, it would be nice if they could be made into defines at the top of the file instead for clarity

Kyep (approved):

SabreML:
The changelog should probably be clarified from
`tweak: Vampires now can only get 250 units of usable blood from a person, afterwords they only gain nutrition.`
to
`tweak: Vampires now can only get 250 units of usable blood from each person, afterwards they only gain nutrition.`
(Also a typo in `afterwords`)

code/game/gamemodes/vampire/vampire.dm
AffectedArc07: You will have to make this use soft refs.

AffectedArc07 (changes requested):

Qwertytoforty:
Being done in bubblegums vamp rework, so closing

PR #15757, opened 2021-03-27
Makes the chapel holy once more.

warior4356:
## What Does This PR Do
The chapel is mostly a roleplay location, this PR seeks to give it a little gameplay relevance for the supernatural-focused antagonists.

The following changes have been made for antagonists inside the chapel:

- Cult spells can not be summoned to hand, and if there will not activate. This does not prevent them from firing into the chapel or otherwise affecting people and items inside.
- Cult runes can be drawn but will fizzle inside the bounds of the chapel.
- More mundane cult items still function as expected. You can still stab with a spear. A bola will still wrap around someone's legs.
- A few advanced cult items are modified inside the chapel. In particular, the blood spear can not be recalled to you while inside, the veil shifter won't work at all, and the mirror shield/shielded robes will not block projectiles.
- Runed metal can not be used to craft inside the chapel. Further, you can't anchor cult structures inside the chapel such as a pylon.
- Construct spells will not work inside the chapel.
- Constructs will take 1/8 of their max health every life tick, resulting in them living for around 15 seconds inside the chapel. When this happens the welder burn sound plays and a fire alert appears for the player in the top right.
- Non-cult constructs are exempt from the above.
- Once the cult summons all cult-related restrictions are removed.
- Vampire powers have been refactored into this new system, but still work as expected. Normally their powers wont work, but full power vampire powers will.
- Revenant powers are disabled inside the chapel.
- For all of the above the following text from vampire powers is used as a warning when their ability fails "Your powers are useless on this holy ground."

I also caught an exploit that allowed the building of cult structures on Centcom and fixed it while I was disallowing their construction in the chapel. This is a fix not a feature because there was already code to prevent this, it just had a way around the check.

## Why It's Good For The Game
Blood cult, particularly as it grows in numbers and visibility can be very violent and aggressive. For most players, their best option for safety is a secure department, such as science, engineering, or cargo where ID requirements act as a barrier for a sudden abduction. Civilians do not have this luxury, and their best place of refuge is the bar which can be somewhat secure due to mob justice but only by that virtue. Giving the chapel these protections not only provides refuge to civilians in these rounds but also gives the chapel an extremely thematic role of holy sanctity against the "forces of evil".

## Changelog
:cl:
add: The chapel now functions as a place of refuge against the supernatural. Cult, vampire, and revenant powers no longer work inside.
fix: Fixed an exploit that allows the build of cult structures on centcom.
/:cl:


code/datums/spell.dm
```diff
@@ -121,6 +121,7 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 
 	var/critfailchance = 0
 	var/centcom_cancast = TRUE //Whether or not the spell should be allowed on the admin zlevel
+	var/holy_cancast = TRUE //Whether or not the spell functions in a holy place
```
FlattestGuitar, line 124: this is not a descriptive name

code/game/gamemodes/cult/blood_magic.dm
```diff
@@ -123,6 +123,8 @@
 	return ..()
 
 /datum/action/innate/cult/blood_spell/Activate()
+	if(owner.holy_check())
```
FlattestGuitar, line 126: You should probably do all these at a higher level. `/datum/action/innate/cult` or `.../cult/blood_spell`

Add a var like for the other spells if you don't want all of them to work this way.

code/game/gamemodes/cult/runes.dm
```diff
@@ -139,6 +139,8 @@ fail_invoke() is called when the rune fails, via not enough people around or oth
 structure_check() searches for nearby cultist structures required for the invocation. Proper structures are pylons, forges, archives, and altars.
 */
 /obj/effect/rune/proc/can_invoke(mob/living/user)
```
FlattestGuitar, line 141: I think it'd be more intuitive if you also couldn't draw the runes in the first place. Might be frustrating.

Also, these runes badly suffer from not telling people what is wrong with them. This might be a good opportunity to add a simple bit of feedback for a fizzing out rune.

code/datums/spell.dm
warior4356: What would you change it to?

code/game/gamemodes/cult/blood_magic.dm
warior4356: I did originally have the checks in attack and attack_self, but blood cult is designed where the parent calls happen after spell effects activate, not before as a check, so without a major refactor there was no way to only put this check in one place.

I wish I could have.

code/game/gamemodes/cult/runes.dm
warior4356: I tried to follow a design principle of, if it is a mundane action it still works. You can still draw a rune, you can still cut yourself, you can still use a spear as a spear. That said, you may have a point here, I wouldn't mind changing it so runes can't be drawn at all. Open for discussion here!

Wojz:
How does this interact with holy constructs?

warior4356:
> How does this interact with holy constructs?

It specifies that this does not affect holy constructs. Specifically, it only affects constructs that are also members of the cult.

code/datums/spell.dm
S34NW: `holy_immunity` perhaps?

Kyep (approved):

code/datums/spell.dm
```diff
@@ -561,6 +562,9 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 	if(is_admin_level(user.z) && !centcom_cancast) //Certain spells are not allowed on the centcom zlevel
 		return 0
 
+	if(!holy_area_cancast && user.holy_check())
+		return 0
```
AffectedArc07, line 566:
```suggestion
		return FALSE
```
Same applies for the rest of this proc

code/datums/spell.dm
```diff
@@ -121,6 +121,7 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 
 	var/critfailchance = 0
 	var/centcom_cancast = TRUE //Whether or not the spell should be allowed on the admin zlevel
+	var/holy_area_cancast = TRUE //Whether or not the spell functions in a holy place
```
AffectedArc07, line 124:
```suggestion
	/// Whether or not the spell functions in a holy place
	var/holy_area_cancast = TRUE
```

code/modules/mob/mob.dm
```diff
@@ -1459,3 +1459,30 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
  */
 /mob/proc/update_runechat_msg_location()
 	return
+
+GLOBAL_LIST_INIT(holy_areas, typecacheof(list(
+	/area/chapel)))
```
AffectedArc07, line 1464:
```suggestion
	/area/chapel
)))
```

AffectedArc07 (changes requested):

code/game/gamemodes/cult/runes.dm
SabreML: Maybe a bit late, but I agree with Flattest on this one. Someone could theoretically set up a whole base in the chapel office only for none of the runes to work when they actually need them.

variableundefined (approved):

PR #15762, opened 2021-03-28
Fixes Gloves of the North Star exploit.

LordShekelstein:
## What Does This PR Do
This PR aims to fix the exploit likely caused by a code oversight which allows for the Gloves of the North Star to be stacked with Carp-Fu and the Hulk Mutations.

## Why It's Good For The Game
This is good for the game because it prevents a frankly overpowered strategy from being used and abused. "Carp-Star" as it has been dubbed, is an unfair strategy which is incredibly unfun to be on the receiving end of. 

## Changelog
:cl:
fix: Defines hulk and "martial art" as originally intended so it is unable to be stacked. 
/:cl:

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -166,7 +166,7 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		if(M.mind.martial_art, /datum/martial_art) || HAS_TRAIT(M, TRAIT_HULK))
+		if(M.mind.martial_art) || HAS_TRAIT(M, TRAIT_HULK))
 			M.changeNext_move(CLICK_CD_MELEE)//normal attack speed for hulk, CQC and Carp.
 		else
 			M.changeNext_move(click_speed_modifier)
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -166,7 +166,7 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		if(M.mind.martial_art) || HAS_TRAIT(M, TRAIT_HULK))
+		if(M.mind.martial_art || HAS_TRAIT(M, TRAIT_HULK))
 			M.changeNext_move(CLICK_CD_MELEE)//normal attack speed for hulk, CQC and Carp.
 		else
 			M.changeNext_move(click_speed_modifier)
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -166,7 +166,8 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		if(istype(M.mind.martial_art, /datum/martial_art/cqc))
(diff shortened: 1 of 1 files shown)


Scout360Pyro:
Thought this was patched a long time ago?

code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -166,7 +166,8 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		if(istype(M.mind.martial_art, /datum/martial_art/cqc))
+		if(istype(M.mind.martial_art, /datum/martial_art))
+		if(HAS_TRAIT(user, HULK_TRAIT))
 			M.changeNext_move(CLICK_CD_MELEE)//normal attack speed for hulk, CQC and Carp.
```
SabreML, line 171: You probably want this instead:
```suggestion
		if(istype(M.mind.martial_art, /datum/martial_art) || HAS_TRAIT(M, TRAIT_HULK))
			M.changeNext_move(CLICK_CD_MELEE)//normal attack speed for hulk, CQC and Carp.
```
Or to make that a bit more readable:
```dm
		if((M.mind has any martial arts) OR (M has the Hulk trait))
				Change the click delay back to normal.
		else
				Change the click delay to the fast version.
```

SteelSlayer: I also suggest doing `if(M.mind.martial_art)` instead of the istype, because, the only thing that's ever assigned to that to that var is a `/datum/martial_art`.

LordShekelstein: Right, followed your suggestions. Thank you all for being so incredibly helpful. As you could most likely tell this is my first PR and all of this is moon runes to my primitive rodent brain.

code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -166,7 +166,7 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		if(istype(M.mind.martial_art, /datum/martial_art/cqc))
+		if(M.mind.martial_art, /datum/martial_art) || HAS_TRAIT(M, TRAIT_HULK))
```
SteelSlayer, line 169: This is what I meant. Checking if the `martial_art` var is something other than null, which implies they currently have a martial art.
```suggestion
		if(M.mind.martial_art) || HAS_TRAIT(M, TRAIT_HULK))
```

code/modules/clothing/gloves/miscellaneous.dm
```diff
@@ -166,7 +166,7 @@
 	var/mob/living/M = loc
 
 	if(M.a_intent in accepted_intents)
-		if(istype(M.mind.martial_art, /datum/martial_art/cqc))
+		if(M.mind.martial_art) || HAS_TRAIT(M, TRAIT_HULK))
```
SteelSlayer, line 169: I forgot remove this in my last suggestion :upside_down_face:. That's why it's not compiling.
```suggestion
		if(M.mind.martial_art || HAS_TRAIT(M, TRAIT_HULK))
```

AffectedArc07 (changes requested):
Im holding an objecting here till you confirm this is tested, due to 5/6 of the commits made not even compiling.

AffectedArc07 (approved):

PR #15778, opened 2021-03-30
Prevents Ghosts from circumventing boo spell cooldown

### LINKED ISSUE #15766: Ghosts Able to Spook Airlocks Free of Charge
**Issue Description**:
As a ghost, you are normally able to use your boo ability on airlocks, which makes them go off as if someone without the correct ID access is interacting with it. This normally causes your boo ability to reset to 0 percent charge and makes you have to wait for it to recharge to 100 percent, however if you have your boo ability ready to influence an airlock and you double-click on it, it instead stays fully charged and ready to use right away.

I expected that the use of the boo ability as a ghost would set the charge of the ability to zero and that the ghost who used it would then have to wait for it to fully recharge.

**What happened instead**:
Instead, the use of a double-click sets the ability back to 100 percent charge.

While definitely not one of the worst bugs/exploits in the game this is still bad as it allows ghosts to use the boo ability on airlocks free of charge, which means they can easily spam it if they want. 

To replicate this bug you just have to be a ghost, select the boo ability, and then double click on an airlock. If done correctly your first click will spook the airlock, which makes it make the no access noise, and the second click should reset the cooldown of the boo ability to 100 percent.

Not sure exactly when this started happening, but I would assume it started after the rework of the boo ability, which allowed ghosts to spook airlocks.

**Extra information**:
N/A

Sirryan2002:
## What Does This PR Do
Fixes #15766 

When rapidly double clicking with the boo spell activated you are able to cast the spell twice. This PR adds an is_casting variable to spell.dm which prevents refund_cast from being called if the same spell has already been cast within 1 second of it. 

This prevents the spell charge level from being set to 100% even if the spell was correctly and successfully cast right before it.

## Why It's Good For The Game
This patches an abusable bug that would allow ghosts to spam boo an airlock as fast as they can click which can have measurable in game impacts. The less things ghosts can abuse the better.

## Changelog
:cl:
fix: Fixed Ghosts being able to double-click airlocks when spooking in order to avoid boo spell cooldown
/:cl:


code/datums/spell.dm
```diff
@@ -465,7 +470,8 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 /obj/effect/proc_holder/spell/targeted/click/InterceptClickOn(mob/living/user, params, atom/A)
 	if(..() || !cast_check(TRUE, TRUE, user))
 		remove_ranged_ability(user)
-		revert_cast(user)
+		if (!is_casting)
```
SteelSlayer, line 473:
```suggestion
		if(!is_casting)
```
Also you never set `is_casting` to true anywhere, so it's always going to be false.

code/datums/spell.dm
```diff
@@ -448,6 +450,9 @@ GLOBAL_LIST_INIT(spells, typesof(/obj/effect/proc_holder/spell))
 		else
 			to_chat(user, "<span class='warning'>[src] is not ready to be used yet.</span>")
 
+/obj/effect/proc_holder/spell/proc/cast_cooldown
```
SteelSlayer, line 453:
```suggestion
/obj/effect/proc_holder/spell/proc/cast_cooldown()
```

dearmochi:
After investigating a little, the root cause of this issue is that airlocks `sleep()` when haunted by a ghost, which prevents spell targeting from ending immediately, instead leaving a small time frame for a double click.

image

Pending a spell refactor, I'd like if your PR instead swapped the following lines:

https://github.com/ParadiseSS13/Paradise/blob/ec967cbaa7b3d96ce2bd0e726e514e5e04ebb30c/code/datums/spell.dm#L506-L507

So you would get:

```dm
	remove_ranged_ability(user)
	perform(targets, user = user, make_attack_logs = create_logs)
```

Why? Because if **any** proc in `perform` calls `sleep()`, you get the issue you are trying to fix. By ending the targeting *before* casting, this problem should be gone. Please do test before committing though.

Sirryan2002:
> After investigating a little, the root cause of this issue is that airlocks `sleep()` when haunted by a ghost, which prevents spell targeting from ending immediately, instead leaving a small time frame for a double click.
> 
> image
> 
> Pending a spell refactor, I'd like if your PR instead swapped the following lines:
> 
> https://github.com/ParadiseSS13/Paradise/blob/ec967cbaa7b3d96ce2bd0e726e514e5e04ebb30c/code/datums/spell.dm#L506-L507
> 
> So you would get:
> 
> ```dm
> 	remove_ranged_ability(user)
> 	perform(targets, user = user, make_attack_logs = create_logs)
> ```
> 
> Why? Because if **any** proc in `perform` calls `sleep()`, you get the issue you are trying to fix. By ending the targeting _before_ casting, this problem should be gone and you shouldn't need the code you wrote which is a workaround. Please do test before committing though.

I have a bit of stuff I'm working on currently, however, I'll get this added before the weeks over. Thanks for pointing this out :D

Sirryan2002:
updated to reflect requested change

dearmochi (approved):

PR #15792, opened 2021-04-01
Removes nuke ops shuttle blindspots

warior4356:
## What Does This PR Do
Have you ever felt like the nuke ops shuttle is too easy to approach? That the space around it has too few bullets? Well have I got a deal for you!

Adds a new and improved nuke ops shuttle, now with 300% more bullet per bullet.

## Why It's Good For The Game
STOP STEALING MY NUKE!

## Images of changes
image


Gatchapod:
You missed two tiles next to the external airlock, one with the button (nukies don't really need that button anyway) and one with a SPACE warning (nukies forget their internals regardless, so it's useless).
Powergamers will find a way to abuse those spots.

PR #15793, opened 2021-04-01
Deploys the nuke by default

warior4356:
Real PR, not April fools.

## What Does This PR Do
This PR makes three changes:
1. The crew's nuke and the nuclear operative's nuke starts deployed and anchored to the floor.
2. Changes the nuke so it can be anchored and unanchored with only the disk, the code is no longer required.
3. Adds a new, admin-only nuke type that is undeployed like the old nuke to be spawned for death squads, SSTs, ERTs, etc.

## Why It's Good For The Game
First and foremost, the nuke being stolen is fun for no one but the person who stole it. Regardless if it is a terror spider stealing the crew's nuke or someone taking the syndicate's off their ship by breaking through the engines, all the nuke being stolen does is delay the round and possibly deny an earned win.

The second change is so the crew can move their nuke if needed, but they need to take the disk to the vault first, which involves the captain, or at least the disk in the round further. This also means that nuclear operatives need to get the disk back to their ship to get the nuke to the station.

The third change is to make sure admin's don't have to mess with var editing if they are making a new nuke for an event or response team of some sort, in case, for example, a blob has eaten the nuke and they want to have an ERT deliver a nuke.

## Images of changes
image
image

## Changelog
:cl:
tweak: All normal nukes start deployed by default
tweak: Nukes can be unanchored with the disk only, without the code
add: Added a new, admin only nuke that is undeployed
/:cl:


ppi13:
Primary issue is is welder, crowbar, welder, wrench, crowbar, so the only one this would stop is blob and terrors from messing with the nuke.

While I do agree the nuke being stolen is somewhat of an issue, it is an imperfect solution as the bolts can be removed in a very short time, and while it does slow down stealing the nuke, it only takes 30 seconds to unbolt it. There is really no issues with this, and honestly not all that bad of an idea, but it is still an imperfect solution.

Short term this will work really well, as most people don't really know you can cut the bolts with tools, and terrors hiding the nuke is probably the biggest issue that will be fixed, so I can get behind that.

DogsParadise:
As I have just resprited the nuke, just out of curiosity.

Have you changed the states, because now there will be a constant animation played.
When the nuke gets bolted, it starts like a timer animation ( it's weird ).
Just checking.

warior4356:
> As I have just resprited the nuke, just out of curiosity.
> 
> Have you changed the states, because now there will be a constant animation played.
> When the nuke gets bolted, it starts like a timer animation ( it's weird ).
> Just checking.

It becomes anchored true and extended true

code/game/gamemodes/nuclear/nuclearbomb.dm
```diff
@@ -14,7 +14,8 @@ GLOBAL_VAR(bomb_set)
 	icon_state = "nuclearbomb0"
 	density = 1
 	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | UNACIDABLE | ACID_PROOF
-	var/extended = FALSE
+	var/extended = TRUE
+	anchored = TRUE
```
dearmochi, line 18:
```suggestion
	anchored = TRUE
	var/extended = TRUE
```

Inherited variables typically go first

dearmochi (changes requested):

DogsParadise:
> > As I have just resprited the nuke, just out of curiosity.
> > Have you changed the states, because now there will be a constant animation played.
> > When the nuke gets bolted, it starts like a timer animation ( it's weird ).
> > Just checking.
> 
> It becomes anchored true and extended true

But.. does this mean there is a constant animation playing.
I personally think it does, which would be weird?

Kyep (approved):

AffectedArc07 (approved):

PR #15803, opened 2021-04-01
SSinput Rewrite, Custom Keybindings

dearmochi:
# NOTE: please testmerge this

## What Does This PR Do

- **Rewrites `SSinput` from ground up.**
  - **Stripped down key loops of as many operations as possible. Movement is for the most part computed on press/release.**
  - **Instead of ***constantly*** looping through the list of clients, only those who are using relevant inputs are looped through, and that for a set amount of time (60 seconds) before being culled from the processing list.**
  - The main input bar should be a little more responsive.
- **Add an actual key binding system.**
  - **Key bindings can be fully customized in the preferences menu. They are global across characters.**
  - **Every key binding is datumized.**
  - This means AZERTY mode and hotkey mode are redundant, so they have been removed.
  - Unicode characters may be used for non-QWERTY keyboards.
  - Support for Alt, Ctrl and Shift modifiers in key bindings.
  - Up to 5 commands may be ran per key.
- Updates DB version to `25`
  - `player.keybindings` added (`longtext`)
- Changes BrowserUI encoding to `UTF-8`

## Why It's Good For The Game

- `SSinput` is horrible and one of the biggest performance hogs. While I don't promise any miracles with this PR, I believe it will be a little better.
- Custom key bindings should absolutely be a thing. Not everyone (myself included) plays on a QWERTY keyboard, or is savvy enough to create macros.

## Images of changes

4GyuvJVs1f

## Changelog
:cl:
add: Add a key binding system -- configure them in the preferences menu
del: Remove Hotkey mode and AZERTY mode
tweak: Can now face a direction without cooldown
fix: Fix input bar appearing as red even if it didn't have focus
/:cl:


code/controllers/subsystem/input.dm
```diff
@@ -1,128 +1,32 @@
...
-		)
-	var/static/list/azerty_movement_keys = list(
-		"Z" = NORTH, "Q" = WEST, "S" = SOUTH, "D" = EAST,				// WASD
-		"North" = NORTH, "West" = WEST, "South" = SOUTH, "East" = EAST,	// Arrow keys & Numpad
-	)
-	movement_keys = default_movement_keys.Copy()
-	alt_movement_keys = azerty_movement_keys.Copy()
+/datum/controller/subsystem/input/fire(resumed = FALSE)
+	var/list/to_cull
+	for(var/c in processing)
```
AffectedArc07, line 17:
```suggestion
	var/list/to_cull
	var/list/processing = src.processing // List cache song 
	for(var/c in processing)
```

code/datums/keybindings/admin.dm
```diff
@@ -0,0 +1,67 @@
+/datum/keybinding/admin
+	category = KB_CATEGORY_ADMIN
+
+/datum/keybinding/admin/can_use(client/C, mob/M)
+	return !isnull(C.holder) && ..()
```
AffectedArc07, line 5: Make this a check_rights() check, otherwise mentors will be able to invoke player panel

AffectedArc07 (changes requested):
Youll need to remove the saving code before a TM can be done

code/controllers/subsystem/input.dm
SteelSlayer: Why do this?

dearmochi:
> 
> 
> Youll need to remove the saving code before a TM can be done

Can do when it's TM time

code/controllers/subsystem/input.dm
AffectedArc07: This gains a perf improvement, this is why `SSlighting` and `SSair` do this

farie82: It has a bigger chance of putting the list in cache memory thus lowering the access time

SteelSlayer: Did some tests and it doesn't seem that way. 

image

AffectedArc07: Ah, this scales much better with larger lists, which is likely why machines and atmos use this
image

SteelSlayer: It looks like using an existing list is faster no matter the size.

AffectedArc07:
```
make new list: 2552.02 ms
use existing list: 2363.62 ms
```
Re-read the numbers

code/datums/keybindings/admin.dm
Wojz: foiled again.

code/modules/mob/mob_defines.dm
```diff
@@ -200,3 +200,6 @@
 
 	/// The location our runechat message should appear. Should be src by default.
 	var/atom/runechat_msg_location
+
+	/// The datum receiving keyboard input. src by default.
+	var/datum/input_focus = null
```
SteelSlayer, line 205: Does this mean you could have a mob remote controlling another mob without actually having to inhabit that mob? :eyes: Something like a borer feeding their inputs to their host.

SteelSlayer (commented):
DM looks fine and its working in my tests. I don't know much about SQL, so I can't comment on that.

code/modules/mob/mob_defines.dm
dearmochi: It would be possible with some work because the movement loop calls the client's `Move` proc, which rely on its current `mob`, not one that is explicitly passed

Mitchs98:
I had no idea this PR existed but now I want this.

code/datums/keybindings/_keybinding.dm
```diff
@@ -0,0 +1,59 @@
+/datum/keybinding
+	/// The user-facing name.
+	var/name
+	/// The UI category to belong to.
+	var/category = KB_CATEGORY_UNSORTED
+	/// The default key(s) assigned to the keybind.
+	var/list/keys
+	/// Whether the keybind should add the client to the SSinput processing loop. Use VERY sparingly.
+	var/key_loop = FALSE
```
farie82, line 9: Why have this when you already have `should_start_looping` for this? Seems a bit double. Just override that one to return `TRUE` instead

code/datums/keybindings/client.dm
```diff
@@ -0,0 +1,42 @@
...
+	. = ..()
+	C.ooc()
+
+/datum/keybinding/client/say
+	name = "Say"
+	keys = list("F3")
+
+/datum/keybinding/client/say/down(client/C)
+	. = ..()
+	C.mob?.say_wrapper()
```
farie82, line 26: I believe a player always has a mob attached to his client. Though I see that the code already checks this a couple of times prior.

code/datums/keybindings/movement.dm
```diff
@@ -0,0 +1,49 @@
+/datum/keybinding/movement
+	category = KB_CATEGORY_MOVEMENT
+	key_loop = TRUE
+	/// The direction to move to when held.
+	var/move_dir
+
+/datum/keybinding/movement/down(client/C)
+	. = ..()
+	var/datum/input_data/ID = C.input_data
+	if(!ID)
```
farie82, line 10: Should never be null

code/modules/input/focus.dm
```diff
@@ -0,0 +1,5 @@
+/mob/proc/set_input_focus(datum/new_focus)
+	if(input_focus == new_focus)
+		return
+	input_focus = new_focus
+	reset_perspective(input_focus) // Maybe this should be done manually? You figure it out, reader
```
farie82, line 5: I'd say so yes. Changing the controlled character does not always equal changing the perspective

code/modules/mob/mob.dm
```diff
@@ -1041,7 +1041,6 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 /mob/proc/canface()
 	if(!canmove)						return 0
 	if(client.moving)					return 0
-	if(world.time < client.move_delay)	return 0
```
farie82, line 1044: Minor thing but this is an undocumented tweak, right?

interface/skin.dmf
```diff
@@ -177,7 +177,7 @@ window "mainwindow"
 		size = 477x20
 		anchor1 = 0,100
 		anchor2 = 100,100
-		background-color = #d3b5b5
+		background-color = #f0f0f0
```
farie82, line 180: Undocumented change here as well

farie82 (changes requested):
Some stuff. Good work on this grandma.
input_data should also never be null. I didn't comment on every instance btw

code/modules/client/client_defines.dm
```diff
@@ -112,6 +107,12 @@
 	/// Is the client watchlisted
 	var/watchlisted = FALSE
 
+	/// Input datum, what the client is pressing.
+	var/datum/input_data/input_data = new()
```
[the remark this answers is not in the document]
AffectedArc07: Clients are hard-del'd themselves so hard deleting this is just extra overhead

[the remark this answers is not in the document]
farie82: Are any variables on the client hard deleted as well then? I'm not 100% aware of how del itself works.

code/__DEFINES/misc.dm
```diff
@@ -363,7 +363,7 @@
 #define INVESTIGATE_BOMB "bombs"
 
 // The SQL version required by this version of the code
-#define SQL_VERSION 23
+#define SQL_VERSION 24
```
AffectedArc07, line 366: Make this 23 for the testmerge so I dont have to juggle TM configs.

code/controllers/subsystem/input.dm
```diff
@@ -1,128 +1,32 @@
...
+		var/client/C = c
+		if(processing[C] + AUTO_CULL_TIME < world.time)
+			if(!length(C.input_data.keys_held))
+				LAZYADD(to_cull, C)
+			else
+				continue // they fell asleep on their keyboard or w/e, let them
```
AffectedArc07, line 23: Is this meant to skip the clients key loop if they are holding down keys? I am worried about cases such as people using spacepods where they may hold a direction key for more than 60 seconds and then wonder why it stops.

interface/interface.dm
```diff
@@ -73,138 +73,3 @@
 		src << link(config.donationsurl)
 	else
 		to_chat(src, "<span class='danger'>The rules URL is not set in the server configuration.</span>")
-
-/client/verb/hotkeys_help()
```
AffectedArc07, line 77: Make sure to remove this verb from the Help menu in the skin

JedDeLagged:
Hey, Numpad user here, and I've noticed some issues. Yes, I know I'm in the super minority.

For background, I can't play on TG servers because the numpad is half broken on that codebase. I posted about this about a year ago, and the problem is still unaddressed today. Understandable, but frustrating.

https://github.com/tgstation/tgstation/issues/52359

Just landed into a game with this test merge, and Page Up (Numpad 9 with Numlock off) doesn't work. Like on TG, you can set Page Up as a hotkey without error or issue, however pressing the key in-game produces no effect. Even pressing the dedicated Page Up key does not work either. Numpad 9 works.

Also, the Numpad bind wording is non-reproducible. What I mean by this, is that the default labels for numpad bindings with Numlock off are North, South, East and West. However, trying to actually bind them using the Keybinder gives you Up Down Left and Right, and then don't perform the action you've attempted to bind them to. Same goes for the diagonals, the default bind lables are Northeast, Southwest, etc, but trying to binding them gives you Home, End, etc.

I suspect there is a discrepancy between what the keybinder is considering a pressed key to be, and what the game is expecting a pressed key to be.

I fully planned to test this further but then the change was pulled and I didn't get the chance. If someone is able to test all the keys on that side of the keyboard (Numlock On and Off) before next release, it'd be appreciated.

Thank you!

dearmochi:
@JedDeLagged

Thanks for reporting this. Numpad keybindings should now be better supported (that includes the issue you raised in the link)

code/modules/admin/verbs/deadsay.dm
```diff
@@ -45,5 +45,7 @@
 	SSblackbox.record_feedback("tally", "admin_verb", 1, "Dsay") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
 /client/proc/get_dead_say()
+	if(!check_rights(R_ADMIN|R_MENTOR))
```
AffectedArc07, line 48:
```suggestion
	if(!check_rights(R_ADMIN))
```

PR #15810, opened 2021-04-02
Evidence bag items now call pickup proc

### LINKED ISSUE #15774: You can hold Cult blades without shakeing as non cult
**Issue Description**:

If you take a cult blade out of a evidence bag you can hold it without haveing the shaking effect

you get the thing that makes you shake and such

**What happened instead**:

you didn't shake and could weild the cult blade as a non cult with no side effects

sec might think your cult or something there is a reason that you cant weild it in the first place

put cult blade in evidence bag
 take out of bag as non cult
 make sure other hand is empty
 you should be holding a cult blade

*shrug figure its always been there

dirrgang:
## What Does This PR Do
PR fixes #15774 . Removing an object from an evidencebag just put the item in your hands without calling the item's pickup proc.

After adding it, it seems to have worked without issues in tests.

## Why It's Good For The Game
Fixes unexpected behaviour as picking up cult items regardless out of which storage item should have the negative effect on the player if they are not a cultist.

## Images of changes
grafik

## Changelog
:cl:
add: Added pickup(user) to /obj/item/evidencebag/attack_self
fix: Fixed item behaviour taken out of evidencebags
/:cl:

### DIFF
--- a/code/modules/detective_work/evidence.dm
+++ b/code/modules/detective_work/evidence.dm
@@ -76,6 +76,7 @@
 		"<span class='notice'>You hear someone rustle around in a plastic bag, and remove something.</span>")
 		overlays.Cut()	//remove the overlays
 		user.put_in_hands(I)
+		I.pickup(user)
 		w_class = WEIGHT_CLASS_TINY
 		icon_state = "evidenceobj"
 		desc = "An empty evidence bag."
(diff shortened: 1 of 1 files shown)


Wojz:
No more free 30 damage swords for the tiders, oh the humanity, how can we go on.

Dumbdumn5 (approved):
This feels like a fix more than a tweak, but 👍 as is necessary.

AffectedArc07 (approved):

PR #15811, opened 2021-04-02
Allows anchoring of fueltanks/watertanks/etc

AffectedArc07:
## What Does This PR Do
You can now anchor and unanchor reagent dispensers (fuel tanks, water tanks, oil tanks, etc etc) with a wrench.

## Why It's Good For The Game
99% of structures like this can be wrenched in place, feels weird not to have it be consistent. Also prevents greytidery by requiring assistants to take 2 seconds to unwrench a high-capacity watertank before just running off with it.

## Changelog
:cl: AffectedArc07
add: You can now anchor and unanchor reagent dispensers (fueltanks, water tanks, etc) with a wrench
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_dispenser.dm
+++ b/code/modules/reagents/reagent_dispenser.dm
@@ -3,8 +3,8 @@
 	desc = "..."
 	icon = 'icons/obj/objects.dmi'
 	icon_state = "watertank"
-	density = 1
-	anchored = 0
+	density = TRUE
+	anchored = TRUE
 	pressure_resistance = 2*ONE_ATMOSPHERE
 	container_type = DRAINABLE | AMOUNT_VISIBLE
 	max_integrity = 300
@@ -27,6 +27,12 @@
 	. = ..()
 	create_reagents(tank_volume)
 	reagents.add_reagent(reagent_id, tank_volume)
+
+/obj/structure/reagent_dispensers/wrench_act(mob/user, obj/item/I)
+	. = TRUE
+	if(!I.tool_use_check(user, 0))
+		return
+	default_unfasten_wrench(user, I)
 
 /obj/structure/reagent_dispensers/temperature_expose(datum/gas_mixture/air, exposed_temperature, exposed_volume)
 	..()
@@ -233,12 +239,6 @@
 	var/obj/item/reagent_containers/food/drinks/sillycup/S = new(get_turf(src))
 	user.put_in_hands(S)
 	paper_cups--
-
-/obj/structure/reagent_dispensers/water_cooler/wrench_act(mob/user, obj/item/I)
-	. = TRUE
-	if(!I.tool_use_check(user, 0))
-		return
-	default_unfasten_wrench(user, I, 40)
 
 /obj/structure/reagent_dispensers/beerkeg
 	name = "beer keg"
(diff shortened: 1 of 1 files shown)


dearmochi (approved):

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -3,8 +3,8 @@
 	desc = "..."
 	icon = 'icons/obj/objects.dmi'
 	icon_state = "watertank"
-	density = 1
-	anchored = 0
+	density = TRUE
+	anchored = TRUE
```
SteelSlayer, line 7: This change should go in the CL.

farie82:
Small note. This will also mean that wall tanks are moveable. Like the ones in chemistry, virology, janitor closet etc

code/modules/reagents/reagent_dispenser.dm
AffectedArc07: > This change should go in the CL.

Will do when I push another update

AffectedArc07:
> Small note. This will also mean that wall tanks are moveable. Like the ones in chemistry, virology, janitor closet etc

Will tweak

AffectedArc07:
Walltanks are now no longer affected by the ability to lock and unlock. Also added some examine text
image

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -3,14 +3,19 @@
...
 	container_type = DRAINABLE | AMOUNT_VISIBLE
 	max_integrity = 300
-	var/tank_volume = 1000 //In units, how much the dispenser can hold
-	var/reagent_id = "water" //The ID of the reagent that the dispenser uses
-	var/lastrigger = "" // The last person to rig this fuel tank - Stored with the object. Only the last person matter for investigation
+	/// How much this dispenser can hold (In units)
+	var/tank_volume = 1000
+	/// The ID of the reagent that the dispenser uses
+	var/reagent_id = "water"
+	/// The last person to rig this fuel tank - Stored with the object. Only the last person matter for investigation
```
S34NW, line 15:
```suggestion
	/// The last person to rig this fuel tank - Stored with the object. Only the last person matters for investigation
```

The rest of the code, including the investigate text, is fine from a grammar perspective.

Cocacolagua:
> Walltanks are now no longer affected by the ability to lock and unlock. Also added some examine text
> image

small dumb nitpick but it would be better to give hints the same way windows do, like "The wheels look like they can be **wrenched** down"

code/modules/reagents/reagent_dispenser.dm
farie82: Should still be in the CL. The examine text is a good extra though

Dumbdumn5:
👍 to this if you include the hint thing @Cocacolagua just mentioned.

AffectedArc07:
image
@Dumbdumn5 @Cocacolagua

farie82 (approved):

Necaladun (approved):
i miss when you could wrench them to loosen the valve and make it leave welding fuel on the ground which you could then ignite and incinerate the entire station 'cause ZAS was horrible.

but yeah this works, approved.

Fox-McCloud:
I'll approve this, but if this rapidly turns into an emerging meta for blocking people, quite frequently, I mightttt grump about this.

Issue #15814, opened 2021-04-02
Heat resistance mutation and trait does not protect from fire anymore.


datlo:
**Issue Description**:

Having the heat resistance mutation/trait protects from fire damage, but not from overheating, causing burn damage anyway. This affects skeletons (Who used to be fireproof) and genetics.

**What did you expect to happen**: 

Be a fireproof skeleton, robust the station by burning everything.

**What happened instead**:

Oof owie ouch my smoky bones

**Why is this bad/What are the consequences**:

This seems to be an unintentional balance change as it is not documented, so, a bug

**Steps to reproduce the problem**:

1. Get heat resistance mutation from genetics/adminbus or become a lich
2. Set yourself on fire
3. Oof ow ouch

**When did the problem start happening**:

Most likely since https://github.com/ParadiseSS13/Paradise/pull/15483

datlo:
Upon investigation, I believe that heat resistance mutation might have always worked this way before the rework. However, skeletons were definitely nerfed without documentation.

PR #15815, opened 2021-04-02
Make the heat resistance trait fully fireproof

datlo:
## What Does This PR Do

Change the heat resistance trait to grant full fireproofing to whoever has it, by making the user immune to temperature raising from fire. This bug was caused by a missing check in /mob/living/carbon/human/proc/get_thermal_protection(), while this check is present in /mob/living/carbon/human/proc/get_heat_protection(temperature).

This affects the heat resistance mutation and skeletons. Skeletons used to be fireproof before the mutation into traits refactor, this was changed without documenting.

Fixes https://github.com/ParadiseSS13/Paradise/issues/15814

## Why It's Good For The Game

From the way it's used in the codebase, it seems that this mutation/trait is meant to make you completely fireproof. 

The opposite mutation make you completely cold proof (including from space, having the coldproof mutation make you entirely spaceproof), the DNA vault says it's supposed to make you 'fireproof', and it's a pretty useless trait/power to have if it doesnt do the one job it's supposed to do effectively.

## Images of changes

image

## Changelog
:cl:
tweak: Heat resistance mutation now makes you completely fireproof.
tweak: Liches are once again completely fireproof.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -358,7 +358,7 @@
 		return
 	var/thermal_protection = get_thermal_protection()
 
-	if(thermal_protection >= FIRE_IMMUNITY_MAX_TEMP_PROTECT || HAS_TRAIT(src, TRAIT_RESISTHEAT))
+	if(thermal_protection >= FIRE_IMMUNITY_MAX_TEMP_PROTECT)
 		return
 	if(thermal_protection >= FIRE_SUIT_MAX_TEMP_PROTECT)
 		bodytemperature += 11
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -366,6 +366,10 @@
 		bodytemperature += (BODYTEMP_HEATING_MAX + (fire_stacks * 12))
 
 /mob/living/carbon/human/proc/get_thermal_protection()
+
```
dearmochi, line 369:
```suggestion
```

dearmochi (changes requested):

SteelSlayer (approved):

Fox-McCloud:
I recall there being a reason for this, but if I figure that out, we can address that later--for now, this makes sense.

Issue #15641, opened 2021-02-28
You cannot augment hands, feet and lower bodies


hal9000PR:
**Issue Description**:

you cannot start the surgery to augment hands, feet and lower bodies

**What did you expect to happen**: 

to be able to augment those limbs

**What happened instead**:

no augmentation surgery could be selected 

**Why is this bad/What are the consequences**:

you cannot augment people.

**Steps to reproduce the problem**:

1) lay human on surgery table
2) aim for lower body/hands/feet
3) try and select augmentation surgery

**When did the problem start happening**:

maybe since #14813 ? but I am unsure.

**Extra information**:

Adrer:
Note: Tested this, appears to only to happen with slimepeople.

PR #15826, opened 2021-04-03
Fix diona/slimepeople limbs not augmenting fully

Adrer:
## What Does This PR Do
Fixes a bug where when you augment a limb on a diona/slimeperson, it would not augment connected limbs (EG. Augmenting an arm would not augment the hand.)

Fixes: #15641

## Why It's Good For The Game
Bug bad.

## Changelog
:cl:
fix: Augmenting limbs on slimepeople and diona now also augments connected limbs.
/:cl:


AffectedArc07 (approved):
convert the child

PR #15828, opened 2021-04-03
Painter Item Refactor

SabreML:
## What Does This PR Do

Refactors Airlock Painters, Floor Painters, Pipe Painters and Window Painters into being datum modules for a single 'Modular Painter' object.
Also moved all the painter icons to `painting.dmi`, and added a new inhand sprite for Window Painters.

## Why It's Good For The Game

Having four items with almost identical sprites and mechanics being completely unrelated to each other code wise is really odd, so I figured why not just make them all part of one big 'Painter' item.
*(Also radial menus are pretty cool looking.)*

## Images of changes

https://user-images.githubusercontent.com/57483089/113489683-7d7ce900-94bd-11eb-8fa0-e587e9cf7706.mp4

## Changelog
:cl:
tweak: Replaced all 'X Painters' with a single 'Modular Painter'. Ctrl+Click to swap the function!
imageadd: Added an inhand sprite for window painters.
/:cl:


Carthusia:
I like, many many times have i specifiaclly asked for a floor painter, and been cycled through the other painter types as they toss them at me one by one.

"No, a floor painter
glass painter?
"No...a floor painter"
Pipe painter?
"No...A FLOOR PAINTER!"

code/game/objects/items/devices/painter/painter.dm
```diff
@@ -0,0 +1,70 @@
...
+	if(loc != user) // Not being held
+		return
+
+	playsound(src, 'sound/effects/pop.ogg', 50, TRUE)
+	var/choice = show_radial_menu(user, src, painter_types, require_near = TRUE)
+	if(!choice)
+		return
+	var/choice_path = painter_types[choice]
+	if(istype(src, choice_path))
+		return
```
SteelSlayer, line 34: There is an issue when switching from window painter to pipe painter. Since the pipe painter is a window painter subtype, this if is true, and will return.

code/game/objects/items/devices/painter/painter.dm
```diff
@@ -0,0 +1,70 @@
...
+
+	playsound(src, 'sound/machines/click.ogg', 50, TRUE)
+	do_sparks(1, FALSE, src)
+	var/obj/item/painter/P = new choice_path(get_turf(user))
+	if(P)
+		qdel(src) // Swap it out for the new one
```
SteelSlayer, line 40: I'm skeptical of new'ing and qdel'ing the items each time you want to switch. I know other items do it to like the jaws of life, but it feels hacky. 

I feel like a better way to do it would be with a `mode` var or similar. For example, given the target is an airlock, in `afterattack` you check it's `mode`. If its in airlock mode, then paint the airlock. Then you can unify everything under one item, sorta like the RCD.

code/game/objects/items/devices/painter/painter.dm
SabreML: Ah damn, I thought I'd fixed that. I don't suppose there's some way of specifying that an object has to be one type specifically, and not a subtype?

code/game/objects/items/devices/painter/painter.dm
SabreML: The reason I didn't do something like this is because of all of variables that the different painter types need. Having them defined on the main object or inside `afterattack()` would get messy really quickly, whereas having them as subtypes overriding a proc is much cleaner.
The new/qdel thing is definitely an issue though. I'd considered having some sort of variable to store any unused painters once they've been spawned to avoid having to delete them, but I wasn't sure how that would work out.

code/game/objects/items/devices/painter/painter.dm
SteelSlayer: The easiest way would be to check the type directly: `if(type == choice_path)`

SabreML: Oh, that's easy.

code/game/objects/items/devices/painter/painter.dm
SteelSlayer: Yeah, I mean I suppose you could just store all 4 painter types inside of the modular painter, and have 4 variable as references to track them. You check the type of object the painter hit, and then route it to the respective painter which can handle it.

Dumbdumn5 (approved):
Oh I _really_ really like this, combining all the aesthetic changing tools into one is incredibly nice for QoL decorating. 👍

Dumbdumn5:
I also just realized this doesn't need head approval but f u c k  it

code/game/objects/items/devices/painter/painter.dm
SabreML: I'm having a lot of trouble trying to figure out how to do this properly, so I'll commit the other two separately.

code/game/objects/items/devices/painter/airlock_painter.dm
```diff
@@ -0,0 +1,61 @@
...
+	if(A.assemblytype == assembly)
+		to_chat(user, "<span class='notice'>This airlock is already painted [paint_setting]!</span>")
+		return
+
+	if(A.airlock_material == "glass" && initial(assembly.noglass)) // prevents painting glass airlocks with a paint job that doesn't have a glass version, such as the freezer
+		to_chat(user, "<span class='warning'>This paint job can only be applied to non-glass airlocks.</span>")
+		return
+
+	if(do_after(user, 2 SECONDS, FALSE, A))
+		// applies the user-chosen airlock's icon, overlays and assemblytype to the src airlock
```
SteelSlayer, line 56:
```suggestion
		// applies the user-chosen airlock's icon, overlays and assemblytype to the target airlock
```

code/game/objects/items/devices/painter/airlock_painter.dm
```diff
@@ -0,0 +1,61 @@
...
+
+	if(!A.paintable)
+		to_chat(user, "<span class='warning'>This type of airlock cannot be painted.</span>")
+		return
+
+	var/obj/machinery/door/airlock/airlock = available_paint_jobs["[paint_setting]"] // get the airlock type path associated with the airlock name the user just chose
+	var/obj/structure/door_assembly/assembly = initial(airlock.assemblytype)
+
+	if(A.assemblytype == assembly)
+		to_chat(user, "<span class='notice'>This airlock is already painted [paint_setting]!</span>")
```
SteelSlayer, line 48: Nitpick but it be a little weird saying "The airlock is already painted with External". I suggest:
```suggestion
		to_chat(user, "<span class='notice'>This airlock is already painted with the \"[paint_setting]\" color scheme!</span>")
```
Or similar.

code/game/objects/items/devices/painter/pipe_painter.dm
```diff
@@ -0,0 +1,35 @@
...
+
+/datum/painter/pipe/New()
+	..()
+	if(!length(modes))
+		for(var/C in GLOB.pipe_colors)
+			modes += "[C]"
+	paint_setting = pick(modes)
+
+/datum/painter/pipe/pick_color(mob/user)
+	paint_setting = input("Which colour do you want to use?", null, paint_setting) in modes
```
S34NW, line 15:
```suggestion
	paint_setting = input("Which color do you want to use?", null, paint_setting) in modes
```
I think we have to use American English here (unfortunately)

S34NW (approved):
Just this one thing I noticed, the rest of the grammar was okay.

Kiyahitayika:
image

@SabreML is always a step ahead of me and I love it!

AffectedArc07 (approved):

PR #15834, opened 2021-04-04
Fixes AI doomsday greentext

### LINKED ISSUE #5773: Doomsday Device does not work for traitor AI hijack objective.
Description:
When playing as a traitor AI, using the doomsday device does not succeed the hijack-style objective, even though no-one escaped.

What I expected:
I expected it would work as there would be no shuttle so no-one could escape alive.

What actually happened:
I failed the objective.

Why it's bad:
Causes what should be a won objective to be lost.

Steps to replicate:
When playing as a traitor AI with the objective to let no-one escape other than AI cores, pAIs and borgs, use the doomsday device in the malfunction tab. When the timer runs out the objective will still fail. 

Gamemode: Traitor, Traitor/Ling, Traitor/Vamp. Not sure which one it was.

hal9000PR:
## What Does This PR Do
If an AI doomsdays they now greentext their hijack objective
Fixes #5773

## Why It's Good For The Game
bugs are bad

## Changelog
:cl:
Fix: Malf AI doomsday now greentexts their hijack objective
/:cl:

### DIFF
--- a/code/game/gamemodes/malfunction/Malf_Modules.dm
+++ b/code/game/gamemodes/malfunction/Malf_Modules.dm
@@ -332,6 +332,7 @@
 		L.dust()
 	to_chat(world, "<B>The AI cleansed the station of life with the doomsday device!</B>")
 	SSticker.force_ending = 1
+	SSticker.mode.station_was_nuked = TRUE
 
 //AI Turret Upgrade: Increases the health and damage of all turrets.
 /datum/AI_Module/large/upgrade_turrets
--- a/code/game/gamemodes/malfunction/Malf_Modules.dm
+++ b/code/game/gamemodes/malfunction/Malf_Modules.dm
@@ -331,7 +331,7 @@
 		to_chat(L, "<span class='danger'><B>The blast wave from [src] tears you atom from atom!</B></span>")
 		L.dust()
 	to_chat(world, "<B>The AI cleansed the station of life with the doomsday device!</B>")
-	SSticker.force_ending = 1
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/objective.dm
```diff
@@ -263,6 +263,8 @@ GLOBAL_LIST_INIT(potential_theft_objectives, (subtypesof(/datum/theft_objective)
 /datum/objective/block/check_completion()
 	if(!istype(owner.current, /mob/living/silicon))
 		return 0
+	if(SSticker.mode.station_was_nuked)
+		return 1
```
AffectedArc07, line 267: TRUE/FALSE these please

AffectedArc07 (changes requested):

code/game/gamemodes/objective.dm
hal9000PR: Done

code/game/gamemodes/malfunction/Malf_Modules.dm
```diff
@@ -332,6 +332,7 @@
 		L.dust()
 	to_chat(world, "<B>The AI cleansed the station of life with the doomsday device!</B>")
 	SSticker.force_ending = 1
```
AffectedArc07, line 334:
```suggestion
	SSticker.force_ending = TRUE
```
This too

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #15836, opened 2021-04-05
Updates the proximity monitor component

### LINKED ISSUE #16219: Proximity sensor - igniter assembly doesn't work when in pockets or held in hands
**Issue Description**:

Proximity sensor - igniter assembly doesn't work when in pockets or held in hands

I expected the assembly to cause sparks indepantly of it's storage slot location

**What happened instead**:

The red light flickers but sparks don't come out.

Because I want to moonwalk while sparking all around very fabously down the hallways while wearing these contraptions because they'r very very cool and awesome, also it has other uses which are not really important right now

put together two of these and fit them in your pockets, they won't work, maybe the first time, but they used to work while I walked around the place, so someone probbly broke it on purpose buecase they hate fun

sometime

**Extra information**:

Fix this k bye

SteelSlayer:
## What Does This PR Do
Requested by @Fox-McCloud

This updates the proximity monitor component and introduces framework for an "advanced" subtype component that has some additional functionality. Following the original PR which added this component, the code is influenced by TG's 
`/datum/proximity_monitor/advanced` datum, but it's not really a "port" per se, and has been heavily modified.

**Changes:**
- Slightly optimizes and simplifies the existing proximity monitor component code.
- Adds a new subsystem, `SSfields` for use with the advanced proximity monitor component. Which processes every 0.2 seconds.
- Creates an "advanced" proximity monitor component which can take advantage of `process()`, make a distinction between inner and edge parts of it's field, check if objects can pass it's field tiles, and more.
- Documents all code related to the proximity monitor component.

## Why It's Good For The ~~Game~~ Codebase
Gives developers additional tools to utilize. Optimizes some code.

Fixes #16219

## Changelog
:cl: 
fix: Fixes a bug where proximity sensors were only detecting motion while on turfs.
/:cl:

### DIFF
--- a/code/controllers/subsystem/processing/fields.dm
+++ b/code/controllers/subsystem/processing/fields.dm
@@ -1,0 +1,6 @@
+PROCESSING_SUBSYSTEM_DEF(fields)
+	name = "Fields"
+	wait = 2
+	priority = FIRE_PRIORITY_FIELDS
+	flags = SS_KEEP_TIMING | SS_NO_INIT
+	runlevels = RUNLEVEL_GAME | RUNLEVEL_POSTGAME
--- a/code/datums/components/_component.dm
+++ b/code/datums/components/_component.dm
@@ -531,6 +531,28 @@
 			target.TakeComponent(comps)
 
 /**
+  * Transfer a single component from the source datum, to the target.
+  *
+  * Arguments:
+  * * datum/target - the target to move the component to
+  * * component_instance_or_typepath - either an already created component, or a component typepath
+  */
+/datum/proc/TransferComponent(datum/target, component_instance_or_typepath)
+	if(!datum_components)
+		return
+	// If the proc was fed a typepath.
+	var/datum/component/comp = datum_components[component_instance_or_typepath]
+	if(comp?.can_transfer)
+		target.TakeComponent(comp)
+		return
+	// if the proc was fed a component instance.
+	for(var/component in datum_components)
+		var/datum/component/C = datum_components[component]
+		if(istype(C, component_instance_or_typepath) && C.can_transfer)
+			target.TakeComponent(C)
+			return
+
+/**
   * Return the object that is the host of any UI's that this component has
   */
 /datum/component/ui_host(mob/user)
--- a/code/datums/components/proximity_monitor.dm
+++ b/code/datums/components/proximity_monitor.dm
@@ -98,8 +98,9 @@
 	// The receiver moved inside of another atom, set the host to that atom so we can track its movement.
 	else if(always_active)
 		toggle_checkers(TRUE)
-		host = get_atom_on_turf(hasprox_receiver)
-		if(ismovable(host) && host != old_loc)
+		var/atom/new_host = get_atom_on_turf(hasprox_receiver)
+		if(ismovable(new_host) && host != new_host && host != old_loc)
+			host = new_host
 			RegisterSignal(host, COMSIG_MOVABLE_MOVED, .proc/handle_move)
 	// Deactivate the field, leave it where it is, and stop listening for movement.
 	else
@@ -181,6 +182,8 @@
  */
 /datum/component/proximity_monitor/proc/recenter_prox_checkers()
 	var/turf/parent_turf = get_turf(parent)
+	if(!parent_turf)
+		return // Need a sanity check here for certain situations like objects moving into disposal holders. Their turf will be null in these cases.
 	var/index = 1
 	for(var/T in RANGE_TURFS(radius, parent_turf))
 		var/obj/checker = proximity_checkers[index++]
--- a/code/datums/components/proximity_monitor.dm
+++ b/code/datums/components/proximity_monitor.dm
@@ -1,139 +1,474 @@
 /**
- * # Proximity monitor component
+ * # Basic Proximity Monitor
  *
- * Attaching this component to an atom means that the atom will be able to detect mobs/objs moving within a 1 tile of it.
+ * Attaching this component to an atom means that the atom will be able to detect mobs or objects moving within a specified radius of it.
  *
- * The component creates several `obj/effect/abstract/proximity_checker` objects, which follow the parent atom around, always making sure it's at the center.
- * When something crosses one of these `proximiy_checker`s, the parent has the `HasProximity()` proc called on it, with the crossing mob/obj as the argument.
+ * The component creates several [/obj/effect/abstract/proximity_checker] objects, which follow the parent (or host) atom around, always making sure it's at the center.
+ * When something crosses one of these proximiy checkers, the `hasprox_receiver` will have the `HasProximity()` proc called on it, with the crossing mob/obj as the argument.
  */
 /datum/component/proximity_monitor
-	var/atom/owner
-	/// A list of currently created `/obj/effect/abstract/proximity_checker`s in use with this component.
+	can_transfer = TRUE
+	var/name = "Proximity detection field"
+	/// The primary atom the component is attached to and that will be receiving `HasProximity()` calls. Usually the `parent`.
+	var/atom/hasprox_receiver
+	/**
+	 * The atom that the component is listening to for movement. Null by default, but this can change.
+	 * For example, if the receiver gets put into a humans's pocket, the host will become the human.
+	 */
+	var/atom/host
+	/// The radius of the field, in tiles.
+	var/radius
+	/// A list of currently created [/obj/effect/abstract/proximity_checker] in use with this component.
 	var/list/proximity_checkers
+	/// The type of checker object that should be used for the main field.
+	var/field_checker_type = /obj/effect/abstract/proximity_checker
+	/// Should the parent always detect proximity and update the field on movement, even if it's not on a turf?
+	var/always_active
 
-/datum/component/proximity_monitor/Initialize()
+/datum/component/proximity_monitor/Initialize(_radius = 1, _always_active = FALSE)
 	. = ..()
-	if(!isatom(parent))
+	if(!ismovable(parent) && !isturf(parent)) // No areas or datums allowed.
 		return COMPONENT_INCOMPATIBLE
-	owner = parent
+	ASSERT(_radius >= 1)
+	hasprox_receiver = parent
+	radius = _radius
+	always_active = _always_active
 	create_prox_checkers()
 
+	if(isturf(hasprox_receiver.loc))
+		return
+	if(!always_active)
+		toggle_checkers(FALSE)
+		return
+	// We only want a host to track if we're `always_active`.
+	host = get_atom_on_turf(host)
+
 /datum/component/proximity_monitor/Destroy(force, silent)
+	hasprox_receiver = null
+	host = null
 	QDEL_LIST(proximity_checkers)
-	owner = null
 	return ..()
 
 /datum/component/proximity_monitor/RegisterWithParent()
 	. = ..()
-	if(ismovable(parent))
-		RegisterSignal(parent, COMSIG_MOVABLE_MOVED, .proc/HandleMove)
+	// Override vars are set to TRUE here to allow for component transfering.
+	if(ismovable(hasprox_receiver))
+		RegisterSignal(hasprox_receiver, COMSIG_MOVABLE_MOVED, .proc/handle_move, TRUE)
+	if(host && ismovable(host))
+		RegisterSignal(host, COMSIG_MOVABLE_MOVED, .proc/handle_move, TRUE)
 
 /datum/component/proximity_monitor/UnregisterFromParent()
 	. = ..()
-	if(ismovable(parent))
-		UnregisterSignal(parent, COMSIG_MOVABLE_MOVED)
+	if(ismovable(hasprox_receiver))
+		UnregisterSignal(hasprox_receiver, COMSIG_MOVABLE_MOVED)
+	if(host && ismovable(host))
+		UnregisterSignal(host, COMSIG_MOVABLE_MOVED)
+
+/datum/component/proximity_monitor/PostTransfer()
+	if(!ismovable(parent) && !isturf(parent))
+		return COMPONENT_INCOMPATIBLE
+	host = null
+	hasprox_receiver = parent
+	if(!isturf(hasprox_receiver.loc))
+		host = get_atom_on_turf(host)
+	recenter_prox_checkers()
 
 /**
- * Called when the `parent` receives the `COMSIG_MOVABLE_MOVED` signal, which occurs when it `Move()`s
- *
- * Code is only ran when there is no `Dir`, which occurs when the parent is teleported, gets placed into a storage item, dropped, or picked up.
- * Normal movement, for example moving 1 tile to the west, is handled by the `proximity_checker` objects.
+ * Called when the `hasprox_receiver` moves. If a `host` exists, its movement will also call this proc.
  *
  * Arguments:
- * * source - this will be the `parent`
- * * OldLoc - the location the parent just moved from
+ * * datum/source - this will be the `parent`
+ * * atom/OldLoc - the location the parent just moved from
  * * Dir - the direction the parent just moved in
  * * forced - if we were forced to move
  */
-/datum/component/proximity_monitor/proc/HandleMove(datum/source, atom/OldLoc, Dir, forced)
-	if(!Dir) // No dir means the parent teleported, or moved in a non-standard way like getting placed into disposals, onto a table, dropped, picked up, etc.
-		recenter_prox_checkers()
+/datum/component/proximity_monitor/proc/handle_move(datum/source, atom/OldLoc, Dir, forced)
+	SIGNAL_HANDLER
+
+	if(Dir)
+		move_prox_checkers(Dir)
+		return // It was just a normal tile-based move, return.
+
+	// The field is always active while the receiver is on a turf.
+	if(isturf(hasprox_receiver.loc))
+		toggle_checkers(TRUE)
+		if(host && ismovable(host))
+			UnregisterSignal(host, COMSIG_MOVABLE_MOVED)
+			host = null
+	// The receiver moved inside of another atom, set the host to that atom so we can track its movement.
+	else if(always_active)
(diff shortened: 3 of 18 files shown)


code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
- * * OldLoc - the location the parent just moved from
+ * * datum/source - this will be the `parent`
+ * * atom/OldLoc - the location the parent just moved from
  * * Dir - the direction the parent just moved in
  * * forced - if we were forced to move
  */
-/datum/component/proximity_monitor/proc/HandleMove(datum/source, atom/OldLoc, Dir, forced)
-	if(!Dir) // No dir means the parent teleported, or moved in a non-standard way like getting placed into disposals, onto a table, dropped, picked up, etc.
-		recenter_prox_checkers()
+/datum/component/proximity_monitor/proc/handle_move(datum/source, atom/OldLoc, Dir, forced)
```
dearmochi, line 85:
```suggestion
/datum/component/proximity_monitor/proc/handle_move(datum/source, atom/old_loc, dir, forced)
```

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
+	SIGNAL_HANDLER
+
+	if(Dir)
+		move_prox_checkers(Dir)
```
dearmochi, line 89:
```suggestion
	if(dir)
		move_prox_checkers(dir)
```

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
+	if(isturf(hasprox_receiver.loc))
+		toggle_checkers(TRUE)
+		if(host && ismovable(host))
+			UnregisterSignal(host, COMSIG_MOVABLE_MOVED)
+			host = null
+	// The receiver moved inside of another atom, set the host to that atom so we can track its movement.
+	else if(always_active)
+		toggle_checkers(TRUE)
+		host = get_atom_on_turf(hasprox_receiver)
+		if(ismovable(host) && host != OldLoc)
```
dearmochi, line 102:
```suggestion
		if(ismovable(host) && host != old_loc)
```

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
-		P.loc = pick_n_take(prox_checkers)
+		P.active = new_active
+
+/**
+ * Specifies a new radius for the field. Creates or deletes proximity_checkers accordingly.
+ *
+ * This proc *can* have a high cost due to the `new`s and `qdel`s of the proximity checkers, depending on the number of calls you need to make to it.
+ *
+ * Arguments:
+ * new_raidus - the new value that `proximity_raidus` should be set to.
```
dearmochi, line 138:
```suggestion
 * new_radius - the new value that `proximity_radius` should be set to.
```

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
+			checker.loc = T
+			continue
+		checker = field_checkers[inner_index++]
+		checker.loc = T
+
+/datum/component/proximity_monitor/advanced/process()
+	if(process_field_checkers)
+		for(var/checker in field_checkers)
+			var/obj/effect/abstract/proximity_checker/advanced/inner_field/F = checker
+			process_inner_checker(F)
```
dearmochi, line 270:
```suggestion
			process_inner_checker(checker)
```

Probably fine to pass as is without casting

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
+	if(process_edge_checkers)
+		for(var/checker in field_checkers)
+			var/obj/effect/abstract/proximity_checker/advanced/edge_field/F = checker
+			process_edge_checker(F)
```
dearmochi, line 274:
```suggestion
			process_edge_checker(checker)
```

Ditto

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,474 @@
...
+ * # Advanced Proximity Checker
  *
- * If `Dir` is null it will be recentered around the receiver via the `recenter_prox_checkers()` proc.
- * If the new location of the receiver is NOT a turf, set `active` to FALSE, so that it does not receive proximity calls.
- * If the new location of the receiver IS a turf, set `active` to TRUE, so that it can receive proximity calls again.
+ * Like basic proximity checkers, these objects can also detect proximity.
+ * However these are meant for when you need to have some additional (more advanced) behavior on top of what basic proximity checkers can do.
+ */
+/obj/effect/abstract/proximity_checker/advanced
+	name = "advanced proximity phecker"
```
dearmochi, line 416:
```suggestion
	name = "advanced proximity checker"
```

code/controllers/subsystem/processing/fields.dm
```diff
@@ -0,0 +1,6 @@
+PROCESSING_SUBSYSTEM_DEF(fields)
```
AffectedArc07, line 1: Why not just hook into `SSfastprocess`? They are both processing subsystems, and they have the same wait.

SteelSlayer: I thought about that and wasn't sure which to use. TG uses `SSfields`, so thats what I used. I assume its for the `SS_KEEP_TIMING` and some of the other flags.

code/datums/components/proximity_monitor.dm
SteelSlayer: Uh, yeah, I don't know why I did that

dearmochi (approved):

matttheficus:
Seems Portable Flashers are broken. They dont detect movement nearby them.

As of #15836 - Updates the proximity monitor component @ 20c9ff

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,472 @@
...
+	if(host && ismovable(host))
+		UnregisterSignal(host, COMSIG_MOVABLE_MOVED)
+
+/datum/component/proximity_monitor/PostTransfer()
+	if(!ismovable(parent) && !isturf(parent))
+		return COMPONENT_INCOMPATIBLE
+	host = null
+	hasprox_receiver = parent
+	if(!isturf(hasprox_receiver.loc))
+		host = get_atom_on_turf(host)
```
farie82, line 73:
```suggestion
		host = get_atom_on_turf(hasprox_receiver)
```
`host` here is `null` and I believe you meant `hasprox_receiver` here. That or `hasprox_receiver.loc`

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,472 @@
...
+	SIGNAL_HANDLER
+
+	if(dir)
+		move_prox_checkers(dir)
+		return // It was just a normal tile-based move, return.
+
+	// The field is always active while the receiver is on a turf.
+	if(isturf(hasprox_receiver.loc))
+		toggle_checkers(TRUE)
+		if(host && ismovable(host))
```
farie82, line 95:
```suggestion
		if(ismovable(host))
```
Can be simplified

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,472 @@
...
+			UnregisterSignal(host, COMSIG_MOVABLE_MOVED)
+			host = null
+	// The receiver moved inside of another atom, set the host to that atom so we can track its movement.
+	else if(always_active)
+		toggle_checkers(TRUE)
+		var/atom/new_host = get_atom_on_turf(hasprox_receiver)
+		if(ismovable(new_host) && host != new_host && host != old_loc)
+			host = new_host
```
farie82, line 103: You'd have to unregister from the host as well here right?
Prob best to make unsetting the host a proc

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,472 @@
...
-/obj/effect/abstract/proximity_checker/Initialize(mapload, atom/_hasprox_receiver)
-	if(_hasprox_receiver)
-		hasprox_receiver = _hasprox_receiver
-		RegisterSignal(hasprox_receiver, COMSIG_PARENT_QDELETING, .proc/OnParentDeletion)
-		if(isturf(hasprox_receiver.loc)) // if the reciever is inside a locker/crate/etc, they don't detect proximity
-			active = TRUE
-	else
-		stack_trace("/obj/effect/abstract/proximity_checker created without a receiver")
+/obj/effect/abstract/proximity_checker/Initialize(mapload, atom/_hasprox_receiver, datum/component/proximity_monitor/P)
+	if(!_hasprox_receiver)
```
farie82, line 380: Is this still required? Seeing it is not used and only used to see if the creator has it's stuff in order?

farie82 (changes requested):

code/datums/components/proximity_monitor.dm
SteelSlayer: Yeah it's probably not needed.

code/datums/components/proximity_monitor.dm
SteelSlayer: I'm actually reworking most of this section to make `always_active` components not horribly break when dealing with nested storage, so this will change quite a bit. Good idea about making the host setting a proc though.

code/datums/components/proximity_monitor.dm
SteelSlayer: Yeah should be `hasprox_receiver`, good catch.

SteelSlayer:
Converting this to a draft to denote that I'm working on fixing some nasty bugs found during the TM. Behavior with `always_active` components breaks when proximity detecting items are placed into nested storage, so I need to come up with a better solution.

lewcc:
> Converting this to a draft to denote that I'm working on fixing some nasty bugs found during the TM. Behavior with `always_active` components breaks when proximity detecting items are placed into nested storage, so I need to come up with a better solution.

I don't know how far you've gotten with this, but I think there's a good bit of crossover between this PR and #16718 that you might be able to harness. I had to fight hard in order to get nested objects to track right, but I think I've got a fine solution with building a hierarchy of signals that ensures `MOVABLE_MOVED` signals get fired if a tracked item's container moves.

SteelSlayer:
I've finally found a better solution to handle nested location hell for the component's parent. The latest commit (`01a60d6`), includes a bit of new logic to accommodate this, as well as removing some of the old logic like `on_detach()` and `on_detach()` for assembly objects, as now the component handles all of that internally.

As such, I believe this is ready for re-review and more test merging.

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,512 @@
...
-/datum/component/proximity_monitor/proc/HandleMove(datum/source, atom/OldLoc, Dir, forced)
-	if(!Dir) // No dir means the parent teleported, or moved in a non-standard way like getting placed into disposals, onto a table, dropped, picked up, etc.
-		recenter_prox_checkers()
+/datum/component/proximity_monitor/proc/on_nested_loc_move(atom/moved_atom, atom/old_loc, dir)
+	SIGNAL_HANDLER
+
+	// It was just a normal tile-based move, so we return here.
+	if(dir)
+		move_prox_checkers(dir)
+		return
```
lewcc, line 111: I'm a little concerned that relying on `dir` might be risky, I think a few things don't necessarily set it when they should (beyond setting .loc). In concept I think it's good, but I'd be a little wary.

lewcc (commented):
Overall looks pretty good! I like the way you solved the problem by storing the hierarchy on the object itself, that seems like a good way to avoid the headaches from tracking old objects. Just a few quick thoughts/concerns of mine.

code/datums/components/_component.dm
```diff
@@ -530,6 +530,28 @@
...
+	if(!datum_components)
+		return
+	// If the proc was fed a typepath.
+	var/datum/component/comp = datum_components[component_instance_or_typepath]
+	if(comp?.can_transfer)
+		target.TakeComponent(comp)
+		return
+	// if the proc was fed a component instance.
+	for(var/component in datum_components)
+		var/datum/component/C = datum_components[component]
```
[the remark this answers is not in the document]
SteelSlayer: `datum_components` is an associative list for whatever reason:

image

So `component` here is `/datum/component/proximity_monitor`, and to get the instance, we need to do `datum_components[component]`.

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,512 @@
...
+
+/datum/component/proximity_monitor/advanced/Initialize(_radius = 1, _always_active = FALSE)
+	. = ..()
+	if(requires_processing)
+		START_PROCESSING(SSfields, src)
+
+/datum/component/proximity_monitor/advanced/Destroy(force, silent)
+	STOP_PROCESSING(SSfields, src)
+	QDEL_NULL(field_checkers)
+	QDEL_NULL(edge_checkers)
```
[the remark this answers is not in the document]
SteelSlayer: Ah, I actually wanted to use `QDEL_LIST()` for these two, because I want to delete everything in the list, and also cut them.

code/datums/components/proximity_monitor.dm
SteelSlayer: Yeah it is a little risky. Currently though, the base component (what's on live) has been using this method for over a year and I've yet to hear about a bug like this. Hopefully that means there are not a lot of direct `loc` sets or moves to nullspace that are a problem. That or no one's reported it.

I think the best idea might be to just address those cases when they come up.

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,512 @@
...
+ * Re-centers all of the `proximity_checker`s around the parent's current location.
  */
 /datum/component/proximity_monitor/proc/recenter_prox_checkers()
-	var/list/prox_checkers = owner.get_all_adjacent_turfs()
-	for(var/checker in proximity_checkers)
-		var/obj/effect/abstract/proximity_checker/P = checker
-		P.loc = pick_n_take(prox_checkers)
+	var/turf/parent_turf = get_turf(parent)
+	if(!parent_turf)
+		return // Need a sanity check here for certain situations like objects moving into disposal holders. Their turf will be null in these cases.
```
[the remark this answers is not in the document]
SteelSlayer: Yeah, I'm thinking the field should just be disabled. Then when they move back onto a turf it can be re-enabled. Will have to fiddle around with some logic to make that work.

hal9000PR (commented):
the majority of the component stuff I don't understand. but this looks good to me.

code/datums/components/proximity_monitor.dm
SteelSlayer: Nope, this is based on the direction they moved in, and not in mob facing direction.

code/datums/components/proximity_monitor.dm
```diff
@@ -1,139 +1,517 @@
...
+
+/**
+ * Relays basic directional movement from the `hasprox_receiver` or `host`, to all objects in the `proximity_checkers` list.
+ *
+ * Arguments:
+ * * move_dir - the direction the checkers should move in
+ */
+/datum/component/proximity_monitor/proc/move_prox_checkers(move_dir)
+	for(var/obj/P as anything in proximity_checkers)
+		P.loc = get_step(P, move_dir)
```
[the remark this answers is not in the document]
SteelSlayer: The proximity checkers are essentially invisible objects which only exist to say to the receiver "hey something moved within X tiles of you". The overhead for `forceMove` is very unnecessary, because turfs and areas don't need to know when these have entered or exited them, etc. Really the only thing that needs to care about what these objects are doing is the component. 

I can add a comment explaining this somewhere in this file.

moxian:
I might be terribly confused and this has nothing to do with the PR (i literally didn't look at the code), but the timing and the name "proximity" make me very suspicious.

Was observing the round (31138). There was a virus outbreak that had coughing as one of its symptoms.
I kept hearing the coughing "to the left of me" regardless of where i was, even when I was on the deserted AI sat.

Also heard a deathgasp.
But that didn't stop the coughs, which means it was not just a single mob that was "stuck" to me.
After a while that coughing became gasping, finally followed by another deathgasp.

I think i was able to narrow the location to this one:
20220429-192432-dreamseeker at least time of the monkey death matched the time of the deathgasp.

I did jump to an executed prisoner earlier in the round and it must have "stuck" somehow that i was there. I'm pretty sure at the time of the execution there were no monkeys to the left, they were brought in later (to feed the vampire in perma).

I should also note that this *only* made me hear the audible emotes, and not things like opening/closing doors.


So, there. Once again, this might be compeltely unrelated to PR, just suspicious that i never encountered (or at least noticed) this before.

AffectedArc07:
> EDIT: I'm dumb. I just found out that i had a local test server running in background, with a spawned human standing next to me. So all these weird coughs and gasps are just from my human dying and then me dying myself. Please ignore.
> 
> I might be terribly confused and this has nothing to do with the PR (i literally didn't look at the code), but the timing and the name "proximity" make me very suspicious.
> 
> Was observing the round (31138). There was a virus outbreak that had coughing as one of its symptoms. I kept hearing the coughing "to the left of me" regardless of where i was, even when I was on the deserted AI sat.
> 
> Also heard a deathgasp. But that didn't stop the coughs, which means it was not just a single mob that was "stuck" to me. After a while that coughing became gasping, finally followed by another deathgasp.
> 
> I think i was able to narrow the location to this one: 20220429-192432-dreamseeker at least time of the monkey death matched the time of the deathgasp.
> 
> I did jump to an executed prisoner earlier in the round and it must have "stuck" somehow that i was there. I'm pretty sure at the time of the execution there were no monkeys to the left, they were brought in later (to feed the vampire in perma).
> 
> I should also note that this _only_ made me hear the audible emotes, and not things like opening/closing doors.
> 
> So, there. Once again, this might be compeltely unrelated to PR, just suspicious that i never encountered (or at least noticed) this before.

If youre going to post unrelated things please wrap it in a markdown `<summary>Short<details>Full speed</details></summary>`.

AffectedArc07 (approved):

PR #15837, opened 2021-04-05
Strange Reagent Decay Rework

SabreML:
## What Does This PR Do

Reworks the damage, necrosis chance, and infection severity for Strange Reagent in order to make it far more consistent, and an actual faster alternative to cloning rather than being worse in every possible way.

With the new system, species with `TRAIT_NODECAY` (Diona and Vox) will get 3/4 of the standard necrosis chance, since currently not decaying makes no difference if all of your organs rot off immediately upon revival.

Overall this will make Strange Reagent a lot more forgiving for `NODECAY` species and in the early stages of death, and slightly  worse for the late stages in order to keep it balanced around cloning.

**CURRENT:**
`var/necrosis_prob = 40 * min((20 MINUTES), max((time_dead - (1 MINUTES)), 0)) / ((20 MINUTES) - (1 MINUTES))`

| Time Dead | Necrosis Chance | `TRAIT_NODECAY` Necrosis Chance | Infection Severity |
| ------------ | ------------------ | --------------------------------------- | ------------------- |
| **0** minutes | 0% | 0% | N/A |
| **3** minutes | 4% | 4% | Organ Death |
| **5** minutes | 8% | 8% | Organ Death |
| **10** minutes | 19% | 19% | Organ Death |
| **15** minutes | 29% | 29% | Organ Death |
| **20** minutes | 40% | 40% | Organ Death |
| **30** minutes | 42% | 42% | Organ Death |

**NEW:**
`var/necrosis_prob = min(decay_time / 200, 50)`
| Time Dead | Necrosis Chance | `TRAIT_NODECAY` Necrosis Chance | Infection Severity |
| ------------ | ------------------ | --------------------------------------- | ------------------- |
| **0** minutes | 0% | 0% | N/A |
| **3** minutes | 0% | 0% | N/A |
| **5** minutes | 0% | 0% | N/A |
| **10** minutes | 15% | 11.25% | Acute Infection |
| **15** minutes | 30% | 22.5% | Organ Death |
| **20** minutes | 45% | 33.75% | Organ Death |
| **30** minutes | 50% | 37.5% | Organ Death |

## Why It's Good For The Game

Currently, using Strange Reagent to revive someone is objectively worse in every possible way to just cloning them. If it fails, you've just gibbed a patient. If it succeeds then you have to spend 10+ minutes fixing their organs.

This change will make it an actually viable method of reviving people without needing to clone them.
It still has the risk of gibbing if you're not careful, but from 5 to 15 minutes after death any organ injuries caused by it can simply be fixed with spaceacillin.
It'll also make reviving Vox less infuriating for both sides since their 'Immune to decay and will not skeletonise' mechanic will  actually make a difference.
(Less of an improvement for Slime People, since they have less organs and can regrow limbs.)

## Changelog
:cl:
experiment: Reworked the damage and decay effects of Strange Reagent. (See the PR for more details and damage/time stats)
tweak: Buffed the toxin damage of SR from 0-15 to 0-50.
/:cl:


code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -779,22 +779,37 @@
...
-							// 0% chance of necrosis within 1 minute of death
-							// 40% chance of necrosis after 20 minutes of death
-							if(!O.vital && prob(necrosis_prob))
-								// side effects may include: Organ failure
-								O.necrotize(FALSE)
-								if(O.status & ORGAN_DEAD)
-									O.germ_level = INFECTION_LEVEL_THREE
-						H.update_body()
+						if(time_dead > 5 MINUTES) // If dead for more than 5 minutes, start the decay.
+							var/decay_time = time_dead - 5 MINUTES // Cut out the threshold
```
dearmochi, line 789: Feels like this and most numbers below could be extracted in defines at the top

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -779,22 +779,37 @@
...
+							necrosis_prob = round(necrosis_prob, 0.1)
+							/*
+							The chance of an organ decaying increases by 15% for every five minutes the mob has been dead, up to a maximum of 50%.
+							Dead for 5 minutes:		[(3000 - 3000) / 200]  = 0%
+							Dead for 10 minutes:	[(6000 - 3000) / 200]  = 15%
+							Dead for 15 minutes:	[(9000 - 3000) / 200]  = 30%
+							Dead for 20 minutes:	[(12000 - 3000) / 200] = 45%
+							*/
+							if(HAS_TRAIT(H, TRAIT_NODECAY))
+								necrosis_prob = min((necrosis_prob / 4) * 3, 0) // Three quarters of the probability for species with NODECAY
```
dearmochi, line 801: Suggestion, possibly look into making a setting per-species/physiology?

code/modules/reagents/chemistry/reagents/medicine.dm
```diff
@@ -779,22 +779,37 @@
 					M.visible_message("<span class='warning'>[M] seems to rise from the dead!</span>")
 					M.adjustCloneLoss(50)
 					M.setOxyLoss(0)
-					M.adjustBruteLoss(rand(0, 15))
-					M.adjustToxLoss(rand(0, 15))
-					M.adjustFireLoss(rand(0, 15))
+					M.adjustBruteLoss(rand(0, 50))
```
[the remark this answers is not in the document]
SabreML: That was intentional, but on second thought I'll just keep the toxin damage increase.

code/modules/reagents/chemistry/reagents/medicine.dm
SabreML: Is there a way to do this without making a new species modifier variable? Having one which *only* effects SR seems a bit odd, whereas the `TRAIT_NODECAY` method works on any species with that trait.

dearmochi: Ah, I was thinking more a general "organ decay rate" variable which I think would be relevant for species but not much else

SabreML: I can't see any way of doing this without either refactoring all organ decay code to standardise it, or just having it a SR only thing. There's also the issue of Vox not having any organ decay at all. (At least while attached to the body, so a species var wouldn't work.)
Personally I don't think having a magic number here would be too bad, and if something gets made which can replace it then it'll be pretty easy to change.

dearmochi: Fair enough :+1:

ManiacalSymphony:
What's trait node cay?

procdrone:
> 
> 
> What's trait node cay?

Its no-decay, a trait vox have, that their bodies will never decay and skeletonize

S34NW (approved):
Grammar in comments concise and understandable. No issues.

SteelSlayer (approved):

Wojz:
What I'd like to see reworked is the trash mechanic that makes griefer docs able to force revive you without it asking you if you want to unlike defib and cloning.

MRgaty:
This could probably be used as a "field resurrection" method, can see doctors carrying bottles with 50 pills of this and some chems to resurrect people while fighting blobs and so.

Carthusia:
> This could probably be used as a "field resurrection" method, can see doctors carrying bottles with 50 pills of this and some chems to resurrect people while fighting blobs and so.

To be fair, it already can now, but even without the insane mass decaying of parts, you still get a huge amount of genetic damage that cant be fixed outside of a cryo tank.

So even if you SR'd someone in the field, theyd have so much genetic damage, that they cant really fight, and if you tried to SR them again, you may well explode them as I think genetic damage is one of the damage types that enables exploding if past 150

Side Note: Ive never heard of a doctor griefing someone by bringing them back to life? Could you elaborate on that more @Wojz

Sikroth:
> Side Note: Ive never heard of a doctor griefing someone by bringing them back to life? Could you elaborate on that more @Wojz

iirc Strange Reagent doesn't prompt you if you want to be revived, instead popping the body back to life instantly. Some people who wish to remain dead for whatever reason don't appreciate this, which I think is undestandable.

Not sure what the best way to go about it would be unless you give SR like a 10 second acting period with some flavor like "the body convulses and shakes" and give the ghost a prompt to revive, and if they don't want to, the body just goes limp again. Something close to how gibbing is handled with a delay? Idunno just throwing ideas.

Carthusia:
> 
> **Not sure what the best way to go about it would be unless you give SR like a 10 second acting period with some flavor like "the body convulses and shakes" and give the ghost a prompt to revive, and if they don't want to, the body just goes limp again.** Something close to how gibbing is handled with a delay? Idunno just throwing ideas.

Ooohh, thats a good idea.

SabreML:
Probably out of scope for this PR, but I've added a 'DNR' verb for ghosts to prevent being revived at all. It also makes the "and their soul has departed" message show when you examine the body.
I figured this way would be better than adding something as you're being revived, since with Strange Reagent especially it can take a lot of work to get to that point, and if they're just unresponsive after however many minutes of healing it can be really demotivating.

Sikroth:
> I figured this way would be better than adding something as you're being revived, since with Strange Reagent especially it can take a lot of work to get to that point

Isn't the DNR "body unresponsive" line added in this PR being delivered at that point as well? Not saying it is a bad way to deal with it, but the feedback for the doctor on wether a body is revivable or not would come at about the same time with a prompt than with this, barring the 10s wait.

Sadly I feel that adding an examine text for a corpse that has decided to DNR, while it would be helpful to avoid wasting time for doctors, would also be kind of too meta 🤔.

Also despite my nitpicks I do like the PR, just giving feedback. Anything that makes cloning less important or relied upon is a positive in my book. Also I'd personally be fine with making the trait_nodecay a bit stronger, though well, probably best to leave that for a possible separate PR.

SabreML:
> Isn't the DNR "body unresponsive" line added in this PR being delivered at that point as well? Not saying it is a bad way to deal with it, but the feedback for the doctor on wether a body is revivable or not would come at about the same time with a prompt than with this, barring the 10s wait.

That's just an extra thing I added so that it doesn't revive the body anyway with no mind. The primary way to tell would be in the examine text, since the DNR verb uses the same system that antaghudding does.
`"They are limp and unresponsive; there are no signs of life..."`
vs
`"They are limp and unresponsive; there are no signs of life and their soul has departed..."`

> Sadly I feel that adding an examine text for a corpse that has decided to DNR, while it would be helpful to avoid wasting time for doctors, would also be kind of too meta 🤔.

Repeating what I said above, this is the same examine text that's added if the player has enabled Antag-HUD or left the server.

Mitchs98:
> This could probably be used as a "field resurrection" method, can see doctors carrying bottles with 50 pills of this and some chems to resurrect people while fighting blobs and so.

Not unless that doctor can also heal all the damage from the blob and the genetic damage. This only affects organ decay, not anything that involves actually making them capable of revival. Nor does it fix any broken bones or anything else that would still be wrong with them.

Necaladun (commented):
Opposed to this. Strange Reagent I have never liked. I don't want it to be a faster alternative to cloning. We don't need a faster alternative to cloning. Dying should be punishing, even if it's just a few minutes out of the game - not "take a pill and get better".

The Ghost thing I like but that should be absolutely a different PR. Please keep your PRs atomic .

SabreML:
> Opposed to this. Strange Reagent I have never liked. I don't want it to be a faster alternative to cloning. We don't need a faster alternative to cloning. Dying should be punishing, even if it's just a few minutes out of the game - not "take a pill and get better".

Personally, I see it as a more engaging alternative to just sticking someone in a cloning pod and walking away. Although I will admit that below 15 minutes after death it's definitely the faster option.
Still, with Strange Reagent you need to heal all of the patient's damage before revival or you risk gibbing them. Compared to Cloning where the only downside is the wait timer.

<hr>

> The Ghost thing I like but that should be absolutely a different PR. Please keep your PRs atomic.

Yeah fair enough, I'll move all of that to a new PR. 👍

Djvoid225:
> Opposed to this. Strange Reagent I have never liked. I don't want it to be a faster alternative to cloning. We don't need a faster alternative to cloning. Dying should be punishing, even if it's just a few minutes out of the game - not "take a pill and get better".
> 
> The Ghost thing I like but that should be absolutely a different PR. Please keep your PRs atomic .

I gotta say I agree with Necal, but also have to disagree on the fairness of it currently. Death is really only punished for non-cloneables right now, and I don't think slime people or vox should be significantly punished because they love playing their character. Meanwhile species that CAN be cloned just pop in a pod and bam, good as new. 

Aside from that, cloning can be upgraded but SR cannot. I feel this would actually make early round revival a little easier, then get phased out late round.

Death should be punishing, but it should be at least _close_ to equally punishing. Races shouldn't be balanced around what happens completely out of their control (every single time it happens). Perhaps this is just coming from a biased source who's sick of seeing miners clone them selves eight times a round while I have to watch a single vox spend half an hour in surgery.

Sikroth:
I feel like the statement "Dying should be punishing, even if it's just a few minutes out of the game - not "take a pill and get better"." in regards to not making SR slightly more useable is sort of invalid. As long as cloning exists, no cloneable species will be punished for more than a few minutes of cloning, and even less when upgraded. SR keeps all the issues the body had (which, if you are dead, it's fair to assume something that kills you likely may have caused further internal damage), so when cloning is an option, SR is rarely if ever used (yes you can SR a human, but why would you when cloning exists).

As for non cloneable species, currently the situation is that either a competent doctor notices that you're a dead, non-cloneable species, and spends up to half an hour trying to revive and fix you and your 12 necrosis, or you just fall victim of malpracticebay and spend the rest of the round dead. This is not going to change significantly with this PR. You might get slightly less necrotic bits inside you but that is the extent of the changes.

This change will do nothing to make cloning less *good*, and the only impact it might have is making some SR resurrections maybe take 15 minutes instead of 30. I don't see how this is problematic. In fact, in my opinion the argument of "dying should be punishing" currently only holds any weight in regards to nerfing or removing cloning instead because it trivializes basically all deaths outside of having your brain stolen/spaced (which is true for all forms of revival anyway).

Fox-McCloud:
Echoing the objections of Neca.

Three objections have been achieved for this PR, closing.

Issue #15667, opened 2021-03-04
bots pathfinding seems to be broken


Sikroth:
**Issue Description**:

There seems to be some issue with bots such as medibots, cleanbots and such being unable to navigate around walls/doors anymore, they seem to just try to go in a straight line to their next checkpoint. This is also an issue with Mulebots.

Not sure if it is related but bots will also sometimes teleport several tiles when moving (like if you put beepsky in a locker as he stops in a checkpoint for a few seconds and drag him, he'll then open the locker and immediately snap back in place).

**What did you expect to happen**: 

Bots pathfind, going around walls and through doors.

**What happened instead**:

Bots stuck in corners, mulebots stuck running into the cargo desk instead of through the doors, etc.

**Why is this bad/What are the consequences**:

Makes bots bad/useless.

**Steps to reproduce the problem**:

build cleanbots or similar in robotics and tell them to patrol
watch them get stuck in a corner

or

send mulebot to other department
watch it run into the cargo desk and get stuck

**When did the problem start happening**:

I noticed it some weeks ago? Maybe earlier, but I had been taking a hiatus until recently, so it's hard to tell. It can't be too old though.

**Extra information**:

S34NW:
armsky has been broken like this for a week or two now, he just sits in the top left corner where he spawns (unless he sees something to beat up)

White-Diavol:
Noticed that clean-bots consistently get stuck if for example there is blood under the sleeper, table or  similar.

White-Diavol:
Looks like the auto-patrol method and the calc_path it calls for the smaller bots is broken here - it calculates a target regardless of obstacles and simply pushes the bot toward without any attempts for a different path or acknowledging they are trapped.

In the image both are stuck and partly non-functionaly and are both attempting to reach targets that are off-screen on the other end of the table (from their POV).
![image](https://user-images.githubusercontent.com/53142276/110205830-91134080-7e7a-11eb-89ff-1347a138a850.png)

White-Diavol:
Also I managed to reproduce them getting stuck on corners.

datlo:
This also affects the syndicate bot on the depot. Get stuck on the closed blast doors when hes supposed to patrol around the depot.
![image](https://user-images.githubusercontent.com/22121511/119669636-7e385880-be38-11eb-8dee-58d4db54f93c.png)

PR #15838, opened 2021-04-05
Extended cleanbot functionality

dirrgang:
## What Does This PR Do
This PR adds functionality similar to those found in the tg codebase, but altered to work with paradise code.

This includes making them a larger threat when emagged - they will continue to make the floor slippery or spew foam, but will spray hydrochloric acid on people who have slipped and fallen.

## Why It's Good For The Game
Makes cleanbots more useful and extends their use both for antagonists as well as regular players.

## Images of changes
2021-04-05 20_47_18-Paradise Station 13
Cleanbot cleans the usual dirt and grime.
2021-04-05 20_47_40-Paradise Station 13
Attacks and kills pests (off by default). On tg it then also removes the corpse - however it makes use of a mechanic that wasn't easily implementable at this time. I may add this in a future PR, ideally treating pests like "trash" after they died.
2021-04-05 20_48_01-Paradise Station 13
Approaches trash and sprays it with hydrochloric acid until it turns into goo.
2021-04-05 20_48_23-Paradise Station 13
Then cleans that up.
2021-04-05 20_48_28-Paradise Station 13
2021-04-05 20_48_40-Paradise Station 13
Busy bot.

I appropriately extended the UI.

## Changelog
:cl:
add: Added the option of pest control, graffiti and trash removal to the cleanbot as well as the TGUI. This also adds a proc to the bots.dem that checks whether there is already a bot on the tile. This is for now only used on cleanbot, but I want to work on the other bots as well, which will also make use of this.
tweak: Modified some of the cleanbot code. 
wip: "Mouse removal" at the moment just kills mice. May be extended in future to also deal with ants, and turn mice into goo...
/:cl:


code/modules/mob/living/simple_animal/bot/bot.dm
```diff
@@ -464,6 +464,13 @@ Pass the desired type path itself, declaring a temporary var beforehand is not r
 			continue //The current element failed assessment, move on to the next.
 		return final_result
 
+/mob/living/simple_animal/bot/proc/check_bot(targ)
+	var/turf/T = get_turf(targ)
+	if(T)
+		for(var/C in T.contents)
+			if(istype(C,type) && (C != src)) //Is there another bot there already? If so, let's skip it so we dont all atack on top of eachother.
+				return TRUE //Let's abort if we find a bot so we dont have to keep rechecking
```
farie82, line 472:
```suggestion
	if(!T)
		return

	for(var/C in T.contents)
		if(istype(C, type) && C != src) //Is there another bot there already? If so, let's skip it so we dont all atack on top of eachother.
			return TRUE //Let's abort if we find a bot so we dont have to keep rechecking
```
Early returns are preferred. Also, spacing and the ()'s around the != check are not required.
Might also want to filter on the bot type in the for loop so it saves some processing power. So like this:
```
for(var/mob/living/simple_animal/bot/B in T.contents)
```

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -19,9 +19,13 @@
 	pass_flags = PASSMOB
 	path_image_color = "#993299"
 
-
+	// Types of dirt to clean
 	var/blood = 1
-	var/list/target_types = list()
+	var/trash = 0
+	var/pests = 0
+	var/drawn = 0
```
farie82, line 26:
```suggestion
	var/trash = FALSE
	var/pests = FALSE
	var/drawn = FALSE
```
All of these are booleans so `TRUE` or `FALSE` are preferred.

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -30,9 +34,12 @@
 	var/failed_steps
 	var/next_dest
 	var/next_dest_loc
+	var/shuffle = FALSE // If we should shuffle our adjacency checking
+
 
 /mob/living/simple_animal/bot/cleanbot/New()
-	..()
+	. = ..()
```
farie82, line 41:
```suggestion
	..()
```
New has no return value

code/modules/mob/living/simple_animal/bot/bot.dm
```diff
@@ -464,6 +464,13 @@ Pass the desired type path itself, declaring a temporary var beforehand is not r
 			continue //The current element failed assessment, move on to the next.
 		return final_result
 
+/mob/living/simple_animal/bot/proc/check_bot(targ)
```
farie82, line 467: `targ` is a bit non-descriptive. Just using `target` is fine. Also if you know the type it has then it's best to include it
```suggestion
/mob/living/simple_animal/bot/proc/check_bot(atom/target)
```
Be sure to change the code below as well to include the name change

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -93,21 +103,43 @@
...
 
 	else if(prob(5))
 		audible_message("[src] makes an excited beeping booping sound!")
 
-	if(!target) //Search for cleanables it can see.
+	if(ismob(target))
+		if(!(target in view(DEFAULT_SCAN_RANGE, src)))
+			target = null
+		if(!process_scan(target))
+			target = null
```
farie82, line 122:
```suggestion
		if(!(target in view(DEFAULT_SCAN_RANGE, src)) || !process_scan(target))
			target = null
```

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -116,6 +148,21 @@
 			bot_patrol()
 
 	if(target)
+		if(QDELETED(target) || !isturf(target.loc))
+			target = null
+			mode = BOT_IDLE
+			return
```
farie82, line 154: I feel like this should be done way earlier. So that the bot can select a new target in the same tick.

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -130,60 +177,111 @@
...
+
+	if(drawn)
+		target_types += /obj/effect/decal/cleanable/crayon
+
+	if(trash)
+		target_types += /obj/item/trash
+
+	target_types = typecacheof(target_types)
+
+/mob/living/simple_animal/bot/cleanbot/UnarmedAttack(atom/A, proximity_flag, list/modifiers)
```
farie82, line 217: I feel it's better to keep the logic below here in the `clean` proc. Calling `UnarmedAttack` while you want to clean is a bit odd. So it's best to keep the old construction of having `UnarmedAttack` call `clean` and `clean` being called by the bot logic

farie82 (changes requested):
Partial review. There is more stuff but I lack the time atm

S34NW (approved):
Grammar looks good to me.

dirrgang:
Okay, I ended up reworking a lot more than initially intended, though the behaviour should be nearly identical. This is largely done in handle_automated_action() which now makes use of the bot modes to switch between different states. This should, in theory, make it easier to understand what the bot is doing.

I also needed to extend the mouse.dm to react to acid damage to get the desired behaviour of dissolving mice.

dirrgang:
I thoroughly apologise for the mess that this PR is, considering the number of commits and such. I'm still a newbie regarding working on larger projects.

dirrgang:
I noticed that the pathfinding/behaviour of the (clean)bots is still pretty broken. At first I thought it was caused by something I did in my PR, but switching back to the master branch without the changes shows the same behaviour.

 They still try to approach messes whose path leads through impassable objects, thus the bot gets stuck on iterating through every visible object it can see but not reach, *slowly* filling the ignore_list. I want to take a look at that in another PR where I look at the bot.dm itself.

unknown

TheYeetster1:
Probably related to issue #15667

Kyep (approved):

code/modules/mob/living/simple_animal/bot/cleanbot.dm
dirrgang: I ended up changing a lot more, I hope it makes more sense this way.

dirrgang:
I extended the cleanbot a little more - I added a crossed proc which gives you a 10% chance to fall over if you cross a cleanbot (or vice versa).

I also improved the code a little more.

tgui/yarn.lock
```diff
@@ -6349,3 +6349,8 @@ yargs@13.2.4:
     which-module "^2.0.0"
     y18n "^4.0.0"
     yargs-parser "^13.1.0"
+
+yarn@^1.22.10:
+  version "1.22.10"
```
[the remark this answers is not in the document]
dirrgang: I removed the change as requested.

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -130,60 +177,111 @@
...
+	else
+		..()
+
 
 /mob/living/simple_animal/bot/cleanbot/explode()
-	on = 0
+	on = FALSE
 	visible_message("<span class='userdanger'>[src] blows apart!</span>")
-	var/turf/Tsec = get_turf(src)
+	var/atom/Tsec = drop_location()
```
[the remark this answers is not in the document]
dirrgang: Implemented.

code/modules/mob/living/simple_animal/bot/cleanbot.dm
```diff
@@ -237,14 +338,17 @@
...
+		if("pests")
+			pests =!pests
+		if("trash")
+			trash =!trash
+		if("drawn")
+			drawn =!drawn
 		if("ejectpai")
 			ejectpai()
+	get_targets()
+	update_controls()
```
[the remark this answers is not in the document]
dirrgang: The get_targets() is called to check for new targets as soon as the targeting parameters are changed, and the update_controls() is to make the tgui reactive.

code/modules/mob/living/simple_animal/friendly/mouse.dm
```diff
@@ -119,6 +119,12 @@
 	if(client)
 		client.time_died_as_mouse = world.time
 
+/mob/living/simple_animal/mouse/acid_act()
+	visible_message("<span class='warning'>[src] dissolve[gender==PLURAL?"":"s"] into a puddle of sizzling goop!</span>")
```
S34NW, line 123:
```suggestion
	visible_message("<span class='warning'>[src] dissolves into a puddle of sizzling goop!</span>")
```

This should always have an `s` after it as a present tense verb.

PR #15821, opened 2021-04-03
Removes Stable Mutagen

SabreML:
## What Does This PR Do

Removes both chemistry recipes for Stable Mutagen, and makes it unable to be synthesised by botany. (Still admin spawnable)
Also makes any species with the `NOTRANSSTING` trait (Vox and Plasmamen) immune to the Changeling Transformation Sting in order to balance it out.

## Why It's Good For The Game

Stable Mutagen currently has two uses:

### 1. Vox Cloning:
For this use, not only does it pretty much invalidate the 'Cannot be cloned' malus, it also doesn't make any sense lore wise.
Cloning a Vox with SM also clones their Cortical Stack, which is a bio-mechanical implant. You can't copy robotic limbs and organs with SM, so there's no reason for this to be an exception.
And of course there's also this section on the Vox wiki page:
> So far, attempts at lab-grown samples have produced no results with such efforts often producing highly acidic compounds capable of burning through most lab equipment.

As it is that statement makes no sense, since anyone with medbay access and a bit of lithium can clone a perfectly functional Vox with no issues.

Removing Stable Mutagen *will* force gibbed Vox to play as another species for the remainder of the round, but considering that Slime People (The only other non-clonable species) have to do this already, I can't see this as too much of a downside.

### 2. Forcefully transforming half the crew into yourself:
I don't think I need to go into too much detail as to why this is bad, but suffice it to say that non-antags really shouldn't have access to a DIY Transformation Sting. Especially considering how controversial the sting itself already is.

## Changelog
:cl:
del: Removed the recipe for Stable Mutagen.
tweak: Made Vox and Plasmamen immune to the Changeling Transformation Sting.
/:cl:

PR #15844, opened 2021-04-06
Doubles Defib Timer

ppi13:
Doubles defib timer, quadruple the time for brain damage to set in, and replaced a number with the define.

## What Does This PR Do

With the removal of stable mutagen #15821 and the strange reagent rework #15837, increasing the time to allow people to be revived to lessen the reliance on cloning, it also increases the time it takes before one starts taking brain damage, and replaces the defib time for cyborg defib with the define for it.

## Why It's Good For The Game

Five minutes goes by really fast, especially when there are already multiple people being treated, or there just are few doctors available, and considering that priority of patient care is not really a thing, slime people and vox can quickly end up needing strange reagent, and for everyone else to depend on cloning.

## Changelog
:cl: PPI
tweak: Tweaks the defib timer from 5 to 10 minutes
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -390,9 +390,9 @@
 
 // Defib stats
 /// Past this much time the patient is unrecoverable (in deciseconds).
-#define DEFIB_TIME_LIMIT 300 SECONDS
+#define DEFIB_TIME_LIMIT 600 SECONDS
 /// Brain damage starts setting in on the patient after some time left rotting.
-#define DEFIB_TIME_LOSS 60 SECONDS
+#define DEFIB_TIME_LOSS 240 SECONDS
 
 //different types of atom colorations
 #define ADMIN_COLOUR_PRIORITY 		1 //only used by rare effects like greentext coloring mobs and when admins varedit color
--- a/code/game/objects/items/weapons/defib.dm
+++ b/code/game/objects/items/weapons/defib.dm
@@ -526,8 +526,8 @@
 
 				QDEL_NULL(ghost)
 			var/tplus = world.time - H.timeofdeath
-			var/tlimit = 3000 //past this much time the patient is unrecoverable (in deciseconds)
-			var/tloss = 600 //brain damage starts setting in on the patient after some time left rotting
+			var/tlimit = DEFIB_TIME_LIMIT //past this much time the patient is unrecoverable (in deciseconds)
+			var/tloss = DEFIB_TIME_LOSS //brain damage starts setting in on the patient after some time left rotting
 			var/total_burn	= 0
 			var/total_brute	= 0
 			if(do_after(user, 20 * toolspeed, target = M)) //placed on chest and short delay to shock for dramatic effect, revive time is 5sec total


Mitchs98:
I'm generally in favor of most sensible PR's to make people lean on cloning less. This will definitely help, albeit not much if the person is too damaged for the doctors' liking. But it'll help. Perhaps another thing that would be good to help would be how blood bags/replacement bags replace blood/the amount it does, as that's one other big point of contention for cloning over treating in most instances. Because really low blood quickly burns through blood stocks, and unless you order lots of extra crates you'll quickly run out. 

Though the blood thing is obviously out of scope for this PR, it's something I feel would help veer away from the 'Ah he's too fucked clone he' mentality even further.

Wojz:
first step to removing cloning, nice I like it

Dafrek:
> first step to removing cloning, nice I like it

Oh god I was thinking the same thing

SabreML:
I don't think *doubling* the timer is needed really, that just seems like it'll almost entirely invalidate Cloning and SR. Making it 7:30 or something like that would be fine though I think.

Leanfrog:
> I don't think _doubling_ the timer is needed really, that just seems like it'll almost entirely invalidate Cloning and SR. Making it 7:30 or something like that would be fine though I think.

I agree with this, _10_ minutes is a long time, to the point where unless your body is in space or something, odds are you'll be coming back, maybe reduce it to something in the range of what Sabre said.

BakKeener:
7 and a half minutes does sound more reasonable, most bodies on station unless hidden very well do tend to get found within 10 minutes.

Gatchapod:
SR rework is mechanically a buff. It more than compensates for SM removal.
With the SR rework, you will want to defib in 0-5 minutes period, SR in 5-10 and consider cloning only after 10 minutes. It'll be a superior option even for cloneable species.
With this PR, you once again make SR a niche chemical used exclusively for two species.

code/__DEFINES/misc.dm
```diff
@@ -390,9 +390,9 @@
 
 // Defib stats
 /// Past this much time the patient is unrecoverable (in deciseconds).
-#define DEFIB_TIME_LIMIT 300 SECONDS
+#define DEFIB_TIME_LIMIT 600 SECONDS
 /// Brain damage starts setting in on the patient after some time left rotting.
-#define DEFIB_TIME_LOSS 60 SECONDS
+#define DEFIB_TIME_LOSS 240 SECONDS
```
dearmochi, line 395: This should be noted in the changelog as well

Kyep (commented):
I'm a little skeptical.
Personally if I'm brought into medbay dead, I'd rather be cloned than revived. Cloning restores you to a pretty-functional state fast, where if you're simply revived you often have to spend the next 20 minutes in medbay as your 43 broken bones are individually mended, which is not very fun.

Mitchs98:
> I'm a little skeptical.
> Personally if I'm brought into medbay dead, I'd rather be cloned than revived. Cloning restores you to a pretty-functional state fast, where if you're simply revived you often have to spend the next 20 minutes in medbay as your 43 broken bones are individually mended, which is not very fun.

This in and of itself won't be nearly as much of an issue once Farie's surgery PR is merged, as surgery on a whole will be quite a lot faster. Plus, that's somewhat of an edge case; having every bone broken and everything damaged, that is.

Necaladun:
Objecting to this. Defibs aren't meant to be the primary resuscitation method, and primarily should be for people who drop dead right in front of you. I'm fine with 5 minutes meaning you're dead-dead.

Fox-McCloud:
3 Objections to this, and closing.

Defibs are meant to be a short-term tool to deal with problems, at the moment; they aren't meant to be a total replacement for cloning---if you want to do non-cloning revives in a longer time frame, you're going to have to use the much more involved and risky strange reagent.

For *now* this is a great way to add increasing levels of time gates to reviving someone, from "defib, patch up, send on the way" all the way to "wait forever looking at a black screen".

PR #15855, opened 2021-04-08
Who & Adminwho improvements

AffectedArc07:
## What Does This PR Do
This PR improves both the `who` verbs and adminwho verbs. The following changes have been made:

Ranks now have an associated colour with them, which can be seen in the table below:
image
*ignore the fact mentors and PRRs show up in the admin section, I was modifying the rank name and not rights to make testing easier*

`who` has also been tweaked. It now shows a comma separated list of players, with patrons showing up orange.
image
Staff will also show in their respective colours. Admins also now have a link to use old `who` since it provides more details.
image
This renders the exact same view as the current who with all the added extras such as mob name, player age, and PP lookup (?). 

The point of putting this into an extra verb and simplifying the existing one is simple. When the server is at highpop, 150 of these lines have to be added, which all have lots of HTML spans and hrefs added in which arent visible in the normal player who list. All the extra data here has to be decoded by the client and can take up to **3 seconds** of a complete client freeze for the view to fully render. This isnt ok when youre just looking to see whos online.

The colours for admin ranks are defined in a config file

## Why It's Good For The Game
- No more client hangs if you want to see who is on while being an admin in 150 pop
- Patrons now get a `who` colour as well as the OOC tag which is visible for 4 minutes in a 2 hour round
- Staff now have fluff colours in the `adminwho` list making it easier to differentiate. 

## Changelog
:cl: AffectedArc07
add: Cleaned up the "who" verb
add: Admins will now have coloured ranks in adminwho
/:cl:

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -651,3 +651,22 @@
 
 	// return the split html object to the caller
 	return s
+
+/**
+  * Proc to generate a "rank colour" from a client
+  *
+  * This takes the client and looks at various factors in order, such as patreon status, staff rank, and more
+  * Arguments:
+  * * C - The client were looking up
+  */
+/proc/client2rankcolour(client/C)
+	// First check if end user is an admin
+	if(C.holder)
+		if(C.holder.rank in GLOB.rank_colour_map)
+			// Return their rank colour if they are in here
+			return GLOB.rank_colour_map[C.holder.rank]
+
+	// If they arent an admin, see if they are a patreon. Just accept any level
+	if(C.donator_level)
+		return "#e67e22" // Patreon orange
+	return null
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -925,3 +925,34 @@
 			runnable_modes[M] = probabilities[M.config_tag]
 //			to_chat(world, "DEBUG: runnable_mode\[[runnable_modes.len]\] = [M.config_tag]")
 	return runnable_modes
+
+/datum/configuration/proc/load_rank_colour_map()
+	var/list/lines = file2list("config/rank_colours.txt")
+
+	for(var/line in lines)
+		// Skip newlines
+		if(!length(line))
+			continue
+		// Skip comments
+		if(copytext(line,1,2) == "#")
+			continue
+
+		//Split the line at every " - "
+		var/list/split_holder = splittext(line, " - ")
+		if(!length(split_holder))
+			continue
+
+		// Rank is before the " - "
+		var/rank = split_holder[1]
+		if(!rank)
+			continue
+
+		// Color is after the " - "
+		var/colour = ""
+		if(length(split_holder) >= 2)
+			colour = split_holder[2]
+
+		if(rank && colour)
+			GLOB.rank_colour_map[rank] = colour
+		else
+			stack_trace("Invalid colour for rank '[rank]' in config/rank_colours.txt")
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -934,7 +934,7 @@
 		if(!length(line))
 			continue
 		// Skip comments
-		if(copytext(line,1,2) == "#")
+		if(line[1] == "#")
 			continue
 
 		//Split the line at every " - "
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -10,9 +10,9 @@
 			continue
 
 		if(C.holder && C.holder.fakekey)
-			temp.Add(C.holder.fakekey)
+			temp += C.holder.fakekey
 		else
-			temp.Add(C.key)
+			temp += C.key
 
 	temp = sortList(temp) // Sort it. We dont do this above because fake keys would be out of order, which would be a giveaway
 
@@ -32,13 +32,13 @@
 		if(colour)
 			out = "<font color='[colour]'>[out]</font>"
 
-		output_players.Add(out)
+		output_players += out
 
-	lines.Add("<b>Current Players ([length(output_players)]): </b>")
-	lines.Add(output_players.Join(", ")) // Turn players into a comma separated list
+	lines += "<b>Current Players ([length(output_players)]): </b>"
+	lines += output_players.Join(", ") // Turn players into a comma separated list
 
 	if(check_rights(R_ADMIN, FALSE))
-		lines.Add("Click <a href='?_src_=holder;who_advanced=1'>here</a> for detailed (old) who.")
+		lines += "Click <a href='?_src_=holder;who_advanced=1'>here</a> for detailed (old) who."
 
 	var/msg = lines.Join("\n")
 
@@ -89,7 +89,7 @@
 
 		if(is_special_character(C.mob))
 			entry += " - <b><font color='red'>Antagonist</font></b>"
-		entry += " ([ADMIN_QUE(C.mob,"?")])"
+		entry += " ([ADMIN_QUE(C.mob, "?")])"
 		Lines += entry
 
 	var/msg = ""
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -1,4 +1,3 @@
-
 /client/verb/who()
 	set name = "Who"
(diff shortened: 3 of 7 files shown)


config/example/rank_colours.txt
```diff
@@ -0,0 +1,11 @@
+# Use this file to denote colours for ingame admin ranks, using the following format
+# Rankname - #12A5F4
+# Colours dont have to be in hex, since this colour string is thrown into the style -aa
+Head of Staff - #e74c3c
```
AffectedArc07, line 4: These colours are sampled from the current discord roles

config/example/rank_colours.txt
```diff
@@ -0,0 +1,11 @@
+# Use this file to denote colours for ingame admin ranks, using the following format
```
dearmochi, line 1: Not sure it makes sense to include Paradise's staff structure as default, considering some of these ranks don't appear in `config/example/admin_ranks.txt`

code/__HELPERS/text.dm
```diff
@@ -651,3 +651,22 @@
...
+/proc/client2rankcolour(client/C)
+	// First check if end user is an admin
+	if(C.holder)
+		if(C.holder.rank in GLOB.rank_colour_map)
+			// Return their rank colour if they are in here
+			return GLOB.rank_colour_map[C.holder.rank]
+
+	// If they arent an admin, see if they are a patreon. Just accept any level
+	if(C.donator_level)
+		return "#e67e22" // Patreon orange
```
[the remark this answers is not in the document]
AffectedArc07: Its only used in one place so I dont see the point

code/controllers/configuration.dm
```diff
@@ -925,3 +925,34 @@
...
+		if(!rank)
+			continue
+
+		// Color is after the " - "
+		var/colour = ""
+		if(length(split_holder) >= 2)
+			colour = split_holder[2]
+
+		if(rank && colour)
+			GLOB.rank_colour_map[rank] = colour
```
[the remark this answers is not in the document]
AffectedArc07: As far as I know, we dont have a handler for this already, and the existing methods just use `ckeyEx()` to split text from ckeys, which wont work for these

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
-
 /client/verb/who()
 	set name = "Who"
 	set category = "OOC"
 
-	var/msg = "<b>Current Players:</b>\n"
+	var/list/lines = list()
+	var/list/temp = list()
+
+	for(var/client/C in GLOB.clients)
```
SabreML, line 8: Presumably `GLOB.clients` would only contain clients?
```suggestion
	for(var/I in GLOB.clients)
		var/client/C = I
```

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
...
+		if(C.holder && C.holder.big_brother) // BB doesn't show up at all
+			continue
+
+		if(C.holder && C.holder.fakekey)
```
SabreML, line 12:
```suggestion
		if(C.holder?.big_brother) // BB doesn't show up at all
			continue

		if(C.holder?.fakekey)
```

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
...
+			temp.Add(C.holder.fakekey)
+		else
+			temp.Add(C.key)
```
SabreML, line 15: `+=` is faster than `Add()` for lists. Same for below.
```suggestion
			temp += C.holder.fakekey
		else
			temp += C.key
```

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
...
+		if(colour)
+			out = "<font color='[colour]'>[out]</font>"
 
+		output_players.Add(out)
+
+	lines.Add("<b>Current Players ([length(output_players)]): </b>")
+	lines.Add(output_players.Join(", ")) // Turn players into a comma separated list
+
+	if(check_rights(R_ADMIN, FALSE))
+		lines.Add("Click <a href='?_src_=holder;who_advanced=1'>here</a> for detailed (old) who.")
```
SabreML, line 41:
```suggestion
		lines += "Click <a href='?_src_=holder;who_advanced=1'>here</a> for detailed (old) who."
```

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
...
-					if(isobserver(C.mob))
-						var/mob/dead/observer/O = C.mob
-						if(O.started_as_observer)
-							entry += " - <font color='gray'>Observing</font>"
-						else
-							entry += " - <font color='black'><b>DEAD</b></font>"
-					else if(isnewplayer(C.mob))
-						entry += " - <font color='green'>New Player</font>"
+	for(var/client/C in GLOB.clients)
+		if(C.holder && C.holder.big_brother && !check_rights(R_PERMISSIONS, FALSE)) // need PERMISSIONS to see BB
```
SabreML, line 55:
```suggestion
		if(C.holder?.big_brother && !check_rights(R_PERMISSIONS, FALSE)) // need PERMISSIONS to see BB
```

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
...
+			continue
+
+		var/entry = "\t[C.key]"
+		if(C.holder && C.holder.fakekey)
```
SabreML, line 59:
```suggestion
		if(C.holder?.fakekey)
```

code/game/verbs/who.dm
```diff
@@ -1,63 +1,98 @@
...
-				entry += " - <b><font color='red'>Antagonist</font></b>"
-			entry += " ([ADMIN_QUE(C.mob,"?")])"
-			Lines += entry
-	else
-		for(var/client/C in GLOB.clients)
-			if(C.holder && C.holder.big_brother) // BB doesn't show up at all
-				continue
+		if(is_special_character(C.mob))
+			entry += " - <b><font color='red'>Antagonist</font></b>"
+		entry += " ([ADMIN_QUE(C.mob,"?")])"
```
SabreML, line 92:
```suggestion
		entry += " ([ADMIN_QUE(C.mob, "?")])"
```

code/__HELPERS/text.dm
```diff
@@ -651,3 +651,22 @@
...
+  * This takes the client and looks at various factors in order, such as patreon status, staff rank, and more
+  * Arguments:
+  * * C - The client were looking up
+  */
+/proc/client2rankcolour(client/C)
+	// First check if end user is an admin
+	if(C.holder)
+		if(C.holder.rank in GLOB.rank_colour_map)
+			// Return their rank colour if they are in here
+			return GLOB.rank_colour_map[C.holder.rank]
```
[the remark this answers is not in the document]
AffectedArc07: I'm discouraging ?. given how much it breaks in BYOND updates and how it's already unstable in 514

code/game/verbs/who.dm
AffectedArc07: See above

code/game/verbs/who.dm
AffectedArc07: For a list with a max of 150 things that isn't being iterated on a loop, the performance gain here will be negligible.

Also sometime GLOB.clients can end up with nulls inside so this accounts for that.

code/game/verbs/who.dm
AffectedArc07: Will change

code/game/verbs/who.dm
AffectedArc07: See above

code/game/verbs/who.dm
AffectedArc07: Will change

Kyep (approved):

dearmochi (approved):

Kyep (approved):
Conffig file has been added to the live server.

variableundefined (approved):

KozackWolf:
I don't know if this is all other people, but the list is like this for me https://gyazo.com/ae9f987479887eb49b20bb999e7b87cb

PR #12873, opened 2020-01-02
adds a wide varriety of new loadout options and some minor cost adjustments

Darkmight9:
## What Does This PR Do
Adds many new items to the loadout screen and reduces the price of the lace up shoes and fancy sandals

General items added: marajauna packet, cigar, zippo, smoking pipe, camera, various plushies, crayon box, replica katana, paper bin, walking cane, fanny pack, drinking flask, silver and gold rings.

Hats and shoes: ushanka, witch hat, pirate hat, rasta beanie, rice hat, jackboots and jack sandals.

Outfits and uniforms: Added all color variations of the jumpsuit, witches costume, USSP uniform, and kilt.

## Why It's Good For The Game
This allows for greater customization of characters without having to do a round start run to the locker room in hopes that your character's look hasn't been taken by someone else already. As for the shoes, if someone wants to dress fancy, let them.

## Changelog
:cl:
add: Added many new load out options
tweak: reduced the price of lace up shoes and fancy sandals
/:cl:

PR #15861, opened 2021-04-09
Fixes loadout errors appearing on every round

SabreML:
## What Does This PR Do

Fixes this appearing on *every* round at startup ever since #12873 was merged:
`WORLD: ## ERROR: Loadout - Missing display name: /datum/gear/uniform/suits`
`WORLD: ## ERROR: Loadout - Missing display name: /datum/gear/uniform/suit`

## Why It's Good For The Game

It causes two error messages in `game.log` every round. (At least it does on my test server, can't verify live.)

## Changelog
:cl:
N/A
/:cl:

### DIFF
--- a/code/__HELPERS/global_lists.dm
+++ b/code/__HELPERS/global_lists.dm
@@ -84,13 +84,13 @@
 			continue
 
 		if(!use_name)
-			error("Loadout - Missing display name: [G]")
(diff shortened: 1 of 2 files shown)


AffectedArc07:
If you're fixing this, make the load out error logging use stack_trace so it generates a runtime, and thus will fail CI

AffectedArc07 (approved):

PR #15867, opened 2021-04-11
Makes Vulps Space German

FlattestGuitar:
## What Does This PR Do
This PR adds an **OPTIONAL** accent for vulps. They're German now, as per the lore. This is off by default, enable it just like you would other auto-accents.

I wouldn't say it's *perfect*, but I wanted to skip using regexes here for obvious reasons. I tried to account for some very specific combinations that would sound wrong, but I'm definitely open for more feedback.

I had to make some minor changes to the replacement code to support bigger keys.

## Why It's Good For The Game
It's funny and also adds some optional flavor to Vulps. Their language is basically space german according to the lore (and the way it sounds), so this is just an extension of what's already there. If your vulp was raised on the race's homeworld they probably won't speak perfect Common.

## Images of changes
image

## Changelog
:cl: Flatty
add: Vulps now have an auto-accent option that makes them sound more... organized. Try it out in your preferences.
/:cl:

### DIFF
--- a/code/modules/mob/living/autohiss.dm
+++ b/code/modules/mob/living/autohiss.dm
@@ -94,7 +94,7 @@
 	if(mode == AUTOHISS_FULL && autohiss_extra_map)
 		map |= autohiss_extra_map
 
-	. = ""
+	. = list()
 
 	while(length(message))
 		var/min_index = 10000 // if the message is longer than this, the autohiss is the least of your problems
@@ -118,6 +118,8 @@
 		else
 			. += pick(map[min_key])
 		message = copytext(message, min_index + length(min_key))
+
+	return jointext(., "")
 
 #undef AUTOHISS_OFF
 #undef AUTOHISS_BASIC
--- a/code/modules/mob/living/autohiss.dm
+++ b/code/modules/mob/living/autohiss.dm
@@ -113,8 +113,10 @@
 
 		. += copytext(message, 1, min_index)
 
-		if(copytext(message, min_index, min_index + length(min_key)) == capitalize(min_key))
+		if(copytext(message, min_index, min_index + length(min_key)) == capitalize(min_key)) //starts with cap
 			. += capitalize(pick(map[min_key]))
+		else if(copytext(message, min_index, min_index + length(min_key)) == uppertext(min_key)) //all caps
+			. += uppertext(pick(map[min_key]))
 		else
 			. += pick(map[min_key])
 		message = copytext(message, min_index + length(min_key))
--- a/code/modules/mob/living/autohiss.dm
+++ b/code/modules/mob/living/autohiss.dm
@@ -48,6 +48,21 @@
 			"s" = list("ss", "sss", "ssss")
 		)
 
+/datum/species/vulpkanin
+	autohiss_basic_map = list(
+			"wh" = list("v"),
+			"w" = list("v"),
+			"th" = list("z")
+	)
+
+	autohiss_extra_map = list(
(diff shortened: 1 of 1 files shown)


DogsParadise:
Pr by flattest +1 always

LightFire53:
Accents are always nice. ~~And that it's not uwu~~

procdrone:
Alles in ordnung

dearmochi (approved):

Carthusia:
Ooohh, is this an accent any race can use, similar to chav or bork bork?

FlattestGuitar:
> Ooohh, is this an accent any race can use, similar to chav or bork bork?

Nope. This is a species-specific auto accent. Like the Unathi, Tajaran, Drask, Kidan and Plasmaman ones. This one is just slightly more involved.

You can toggle this in your character preferences or with the `Toggle Auto-Accent` verb.

KasparoVy:
Fuck yes!

Carthusia:
Could it be possible to make it one of the other accents that people can select for any race? The more we have on that list the better.

The current ones are available are....kind of memey? Bork bork nom nom, uwotm8 kind of deal?

Ralta:
> Could it be possible to make it one of the other accents that people can select for any race? The more we have on that list the better.
> 
> The current ones that are available are....kind of memey? Bork bork nom nom, uwotm8 kind of deal?
> 
> Just feel that this would be better suited to allowing any race to pick it, than limiting it behind one singular species

While it would certainly be possible, the intent of the PR is to bring a lore-friendly accent to the Vulpkanin race; diluting it by allowing other races to use it is outside the scope and intent.

Necaladun (commented):
Germanic Languages =/= German - English is the most spoken Germanic language. It has some harsh gutteral sounds, and "ze" instead of "the" doesn't really get that across to me. "bier" rather than "beer" is just...German. I don't want Vulp to be just "space german".

Unlike this hiss of unathi etc, this is much more of a meme to me akin Swedish genes. "It's funny" should be a justification used for clowns.

FlattestGuitar:
Sorry for the late response, I missed that comment!

I'm calling it German because I think that's a funny way of wording it in the PR, but I'm absolutely up for making it less Dr Strangelove-y. The idea here is to make it more in line with their lore-assigned language, because I think this is an easy way to add some juicy RP flavor to a race that doesn't have that much of it. I don't think we need to stick to any already existing languages too much - Vox are a great argument for that.

Would you be open to seeing this PR if it was more in line with just the Caniluntzt syllables rather than an obvious German accent, @Necaladun? I'm seeing some objections and I'd like to know whether there's something that can be done to still get something like this through.

Kyep (changes requested):
Nein. This won't pass.
I was going to look at voting against it, then I saw it already has 3 head/maint votes against it already, which is enough to close it. Even without my vote.

FlattestGuitar:
@Kyep can you say what you think is wrong with this PR? Will making it less obviously German lead to you supporting the PR?

Or is it something else?

Kyep:
@FlattestGuitar 
Personally, when I imagine the long-term consequences of this PR (a 30% chance for every vulp I see to speak with a German accent)... I find the idea annoying. That's pretty much it. I try to imagine what the game would be like if this saw regular use, and what I imagine is not appealing.
That's just my opinion though. It is up to the other heads/maints voting on this to explain their own objections.

FlattestGuitar:
Yeah I think this might end up being too silly. I'll close it for now.

PR #14589, opened 2020-10-11
Changeling powers adjustments, and removals.

Qwertytoforty:
## What Does This PR Do

**Nerfs:**

Strained muscles.
Has to be active for at least 3 stacks.
Used 3 chemicals per stack (4 seconds), scaling up for an extra 3 chemicals per stack.
Can not be used again till stacks go away (2 seconds per stack)
Speed legs can not be used if you are faster than normal speed in your DNA, like DNA vault or being an ashwalker.

**Buffs**
Armblade.
Cost down by 10 chemicals, to 15

Shield.
Cost down by 10 chemicals to 10, changeling internal genetic damage cost down by 5 to 10.

Chitinous armor. 
Cost down by 5 to 20, changling internal genetic damage cost by 3 to 8.
Armor values changed from 40 40 40 20 to 50 50 30 25 

Chameleon skin.
Cost down by 5 to 20 chemicals.

Anatomic panacea now removes hostile implanted implants, (IE, death alarm / tracker / exile / explosive, ect...)

**Buffs/nerfs:** Epi overdose deals 30 genetic damage and takes 30, deals 5 toxin damage to user as it is an overdose, no longer adds synaptizine to the user, instead, after unstuning you, gives you 90% stamina resist and makes you unstunable for 3 seconds.

**Removals**

Removes blind sting.

Removes transformation sting

**ADDS**

Fake armblade sting, a 20 chemical sting that allows changelings to put a fake armblade on someone for 1 minute. Fake armblade has 5 brute and doesnt pry open doors.

Armblade chat messages: Fake and real armblades now has a message of them forming in chat localy around the changeling. 

Bioadaption, an passive ability that allows changelings to keep exotic organs (IE, xeno organs, demon heart, or abductor organs) on transforming / reviving without destroying them, and cybernetic organs / implants, so long as said organ / implant is not dead. If the cybernetic  organ / implant is broken or dead, it will be removed like before (no broken dead heart keeping a changeling dead). Costs 1 DNA point

Bone shuriken, an ability that allows changelings to grow 2 bone shruikens, that can be thrown at people like throwing stars. Unlike throwing stars, they deal less damage, and become dull after hitting someone, so crew, or changelings, can not endlessly reuse them.

**TODO**

- [x] An ability that allows changelings to make bone shuriken, weaker than traitor ones, that become dull after embedding / hitting the target. Also needs a sprite.

[...]

PR #15869, opened 2021-04-11
Removes the Transformation Sting ability

SabreML:
## What Does This PR Do

Removes the Transformation Sting ability, and the `NOTRANSSTING` trait.

Pretty much the same thing as in #14589, but I figured that since that's been open for six months now with no change, and that the only way to reverse the sting was removed (#15821), it'd be best to do this separately.

## Why It's Good For The Game

To be honest, the reason for removing this should be pretty obvious to anyone who's played more than two Changeling rounds.

While someone could in theory use the transformation sting to frame someone else, I've never seen that happen once in 1100 hours.
There's even advanced rules specifically because of it being abused so often:
> Murderboning so you can absorb is seen as excessive. Your abilities should be used to further your goals or to prevent apprehension if you are found out. Example: Transformation Sting.
Randomly transforming crew for no other purpose than to cause chaos runs counter to this, however using it to create copies of yourself to throw off pursuit of Security is acceptable.

And even despite that, it still happens to some extent almost every cling round.

## Changelog
:cl:
del: Removed the Transformation Sting ability.
/:cl:

### DIFF
--- a/code/__DEFINES/genetics.dm
+++ b/code/__DEFINES/genetics.dm
@@ -74,7 +74,6 @@
 
 //Species traits.
 #define NO_BLOOD		"no_blood"
-#define NOTRANSSTING	"no_trans_sting"
 #define IS_WHITELISTED 	"whitelisted"
 #define LIPS			"lips"
 #define EXOTIC_COLOR	"exotic_blood_color"
--- a/code/game/gamemodes/changeling/powers/tiny_prick.dm
+++ b/code/game/gamemodes/changeling/powers/tiny_prick.dm
@@ -66,60 +66,6 @@
 		add_attack_logs(user, target, "Unsuccessful sting (changeling)")
 	return 1
 
-/datum/action/changeling/sting/transformation
-	name = "Transformation Sting"
-	desc = "We silently sting a human, injecting a retrovirus that forces them to transform. Costs 50 chemicals."
-	helptext = "The victim will transform much like a changeling would. The effects will be obvious to the victim, and the process will damage our genomes."
-	button_icon_state = "sting_transform"
-	sting_icon = "sting_transform"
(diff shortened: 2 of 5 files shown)


Mitchs98:
> and that the only way to reverse the sting was removed (#15821

I'd like to point out this is actually untrue. Genetic syringes can reverse it all the same for races that SM could reverse it for.

That said. I'm all for it being removed. All I ever see it really used for is grief like SM, very rarely do I see it actually used for its purpose.

AffectedArc07:
> and that the only way to reverse the sting was removed (#15821)

image
image

You cannot argue this as a removal reason, when you yourself are the removal reason. Thats like me nerfing the shit out of skrell then removing them because "Its just a human clone now".

matttheficus:
GriefSting is not fun to deal with, nor have I seen it be used for its actual purpose in nearly 1000 hours. Id be ecstatic to see it go.

SabreML:
> You cannot argue this as a removal reason, when you yourself are the removal reason. Thats like me nerfing the shit out of skrell then removing them because "Its just a human clone now".

Yeah I'm not trying to use that as an excuse to remove the transformation sting; I was just as against the ability before that PR was merged as I am now. Even dismissing that point I still believe there's more than enough reasons to get rid of this.

That was meant as more of a side note and another point against the transformation sting, similar to the Shadowling mindshield nerf. That change made people dislike Shadowlings even more than they already did, but it's only one of many reasons that there's a PR to remove the gamemode entirely.
The only reason I didn't make this PR before or as part the SM one is because it was already a feature in #14589, and I didn't want to just steal from that.

Kyep (approved):
Yes, kill it.

variableundefined (commented):
Two maintainers objection. Closing.

PR #15872, opened 2021-04-12
Removes the "Record Maintenance" tab from sec records consoles

S34NW:
## What Does This PR Do
This PR removes the Record Maintenance tab from sec records consoles, thus consigning its functions to the void.
No more:

- Useless backup/import buttons that literally do nothing
- Delete all records button
- Delete cell logs button (has anyone pressed this ever?)

## Why It's Good For The Game
The ability to delete all records does not make the game more fun. It brings nothing good to the table.
I made a change to try and make the button more tolerable at #15712 but it has not really had the intended effect, that that was due to me missing the point of what caused the issue. The problem was never with who can access the button, the problem was with the button itself.

As an antag with records access, it takes one extra click to delete only _your_ record specifically, rather than delete them all. And it is functionally the same for sec dealing with you, except way less obvious, and possible to fix with a few clicks. 

What deleting _all_ records does, is make sec unable to set Tyder Grey for breaking into chemistry for the 3rd time. Deleting all records doesn't impede sec chasing you more. It just makes the round less fun for everyone because dealing with shitters and tiders while also dealing with Robusto McAllaccess is now way harder because Tyder Grey doesn't have a red W, if you have a photographic memory then great this is not a problem. But for your average secoff, it can completely paralyze the department and their role, akin to hijack level chaos.
Someone nabs the baldie and yeets them into processing. Warden walks up. No record. They have no clue what they did.
The delete all records button is fun for nobody, it doesn't help you more than deleting your own record does, and it takes literally 10+ minutes to restore all records manually, but why would anyone do that with the risk of you just pressing 1 button to undo it all again?

As for the other features under this tab, the Backup and Import Backup buttons literally have never worked, because there was never any code to actually do those things behind it. Encouraging disk hoarding and deletion/restoration wars is not ideal either, so this is better than having them work.

Delete cell records, who ever even uses this? I don't think I have ever seen it used. Not even once.

## Changelog
:cl:
del: Removed "Record Maintenance" tab from security records consoles
/:cl:


hal9000PR:
Nobody has ever done this and it added to the enjoyment of the round. Good riddance.

Kyep (approved):

AffectedArc07 (approved):

PR #15886, opened 2021-04-14
Fixes surgical table moving anchored things

S34NW:
## What Does This PR Do
Makes surgical table check for anchored before grabbing something.

## Why It's Good For The Game
Anchored things are anchored, not moveable.

No more moving intercoms:

https://user-images.githubusercontent.com/12197162/114706003-5675b100-9d20-11eb-8418-0299d1596b04.mp4

## Changelog
:cl:
fix: Fixed surgical tables moving anchored things
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -591,7 +591,7 @@
 	verbs -= /obj/structure/table/verb/do_flip
 	typecache_can_hold = typecacheof(typecache_can_hold)
 	for(var/atom/movable/held in get_turf(src))
-		if(is_type_in_typecache(held, typecache_can_hold))
+		if(is_type_in_typecache(held, typecache_can_hold) && held.anchored == 0)
 			held_items += held.UID()
 
 /obj/structure/table/tray/Move(NewLoc, direct)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/tables_racks.dm
```diff
@@ -591,7 +591,7 @@
 	verbs -= /obj/structure/table/verb/do_flip
 	typecache_can_hold = typecacheof(typecache_can_hold)
 	for(var/atom/movable/held in get_turf(src))
-		if(is_type_in_typecache(held, typecache_can_hold))
+		if(is_type_in_typecache(held, typecache_can_hold) && held.anchored == 0)
```
AffectedArc07, line 594:
```suggestion
		if(is_type_in_typecache(held, typecache_can_hold) && !held.anchored)
```

AffectedArc07 (changes requested):

code/game/objects/structures/tables_racks.dm
farie82, line 594:
```suggestion
		if(!held.anchored && held.move_resist != INFINITY && is_type_in_typecache(held, typecache_can_hold))
```
`move_resist` is also used to say an object can't be moved. Left over from a rework that never fully got finished

S34NW: I'd forgotten about move resist, this means no more terror empress surfing down corridors on a surgical table :(

farie82: Those got `MOVE_FORCE_STRONG` as value. `INFINITY` is only reserved for things that should never ever move

AffectedArc07 (approved):

Issue #2913, opened 2015-12-21
Directional Windows + Atmos


Deanthelis:
When a directional (single-pane, not full) window is secured to a floor, then the tile under it is removed, then the plating has a reinforced floor (metal rods instead of floor tiles) placed under it, it loses its airtight quality. It can be fixed by using a screwdriver twice. I don't know if this is a feature or a bug, but it is rather misleading - I wouldn't have realized it had the floor under me not started turning icy.

TheYeetster1:
Can report that this bug is still in the game 

Video of what's going on.
https://youtu.be/EcXotcPBmGQ

PR #15893, opened 2021-04-15
Contributer Banned from Paradise

Kiyahitayika:
## What Does This PR Do

- Attempts to fix the window atmos leak issue when you move fulltile windows
- Attempts to fix the problem with atmos leaking under windows when you pry up their tile - resolves https://github.com/ParadiseSS13/Paradise/issues/2913
- Deletes the `fulltile` variable, replaces the very few instances where this was referenced with `is_fulltile()`
- Makes `window/full` a proper subtype with overrides (except for CanAtmosPass(), my arch nemesis)
- Tidies up the Clockwork window types

## To Do

- [ ] Ensure that directional windows only block atmos from an exterior-facing direction, not an internal-facing direction - currently they are doing both as seen in the linked video
- [ ] AND that they will only do that when Secured()

## Why It's Good For The Game
- Nice and airtight once it's done

## Images of changes
**Current Iteration:**
https://user-images.githubusercontent.com/69871346/114944669-0aa53380-9e16-11eb-9588-6462d3bf9705.mp4

## Changelog
:cl:
tweak: Makes properly-secured windows airtight again
/:cl:

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -28,6 +28,8 @@
 
 #define is_pen(W) (istype(W, /obj/item/pen))
 
+#define is_fulltile(W) (istype(W, /obj/structure/window/full))
+
 GLOBAL_LIST_INIT(pointed_types, typecacheof(list(
 	/obj/item/pen,
 	/obj/item/screwdriver,
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -156,16 +156,17 @@ Turf and target are seperate in case you want to teleport some distance from a t
 	for(var/obj/structure/window/D in loc)
 		if(!D.density)
 			continue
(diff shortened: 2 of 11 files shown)


Kiyahitayika:
Linter failed trying to call `CanAtmosPass()` on a turf at the edge of a Z-Level as per @AffectedArc07 's apropos forewarning to me!

` src: the asteroid sand (255,29,3) (/turf/simulated/floor/plating/asteroid/airless/cave/volcanic/has_data)`

TheYeetster1:
This may or may not be helpful but what I have found is that if you crowbar a floor tile adjacent to a full window, then replace it, the window then becomes air tight, I'll try to get a video soon.

Adrer:
Possibly a fix to #2913? If it is, you should probably add it to PR description.

TheYeetster1:
What happened here?

PR #15903, opened 2021-04-17
Fixes character saving. Again.

AffectedArc07:
## What Does This PR Do
Fixes the bug where character saving would try to save nulls in a column that doesnt accept nulls.

At some point I need to go over the schema and make these columns actually allow nulls instead of them having phantom empty strings. However, that day is not today

## Why It's Good For The Game
People should be able to save stuff

## Changelog
:cl: AffectedArc07
fix: Fixed an issue where you couldnt save certain alt head styles
/:cl:

### DIFF
--- a/code/modules/client/preference/preferences_mysql.dm
+++ b/code/modules/client/preference/preferences_mysql.dm
@@ -440,7 +440,7 @@
 													"f_style" = f_style,
 													"markingstyleslist" = markingstyleslist,
 													"ha_style" = ha_style,
-													"alt_head" = alt_head || "",
+													"alt_head" = (alt_head ? alt_head : ""), // This it intentnional. It wont work without it!
 													"e_colour" = e_colour,
 													"underwear" = underwear,
 													"undershirt" = undershirt,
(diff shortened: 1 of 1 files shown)


code/modules/client/preference/preferences_mysql.dm
```diff
@@ -440,7 +440,7 @@
 													"f_style" = f_style,
 													"markingstyleslist" = markingstyleslist,
 													"ha_style" = ha_style,
-													"alt_head" = alt_head || "",
```
FlattestGuitar, line 443: Does this not actually work for a null `alt_head`? That seems odd.

AffectedArc07: Nope. It doesnt.

Calyxman:
Somebody. Please merge this. It would make my day.

code/modules/client/preference/preferences_mysql.dm
```diff
@@ -440,7 +440,7 @@
...
+													"alt_head" = (alt_head ? alt_head : ""), // This it intentnional. It wont work without it!
```
AffectedArc07, line 443:
```suggestion
													"alt_head" = (alt_head ? alt_head : ""), // This it intentional. It wont work without it!
```

PR #15906, opened 2021-04-18
Fixes: Preferences now properly loads the UI style.

### LINKED ISSUE #7510: UI style doesn't save between games
**Problem Description**:
The UI style will always default to midnight when a new round starts.

I expect that when I change the UI style to "plasmafire" and press "save setup," I will have plasmafire UI next shift.

**What happened instead**:
I changed my UI style to "plasmafire" on multiple shifts, pressed "save setup" multiple times, and I still start new rounds with the default "Midnight" style.

It'd be nice if saved preferences actually saved.

Change UI style to plasmafire. Play a match with your sexy orange UI. Join the next match. Cry when your UI is greyish blue again.

anything else out of the ordinary happening?)**: I really didn't want to report this as a bug in case it's just me being stupid, but I don't know what to do at this point. You can set the UI style as plasmafire both from the preferences tab in-game and at the lobby screen, and it will still revert back next shift.

Dragonkiller93:
## What Does This PR Do
This change hopefully makes the preferences menu now properly loads the correct UI style. It adds the other UI styles besides White and Midnight to the list sanitized against by the loadpreferences function. fixes #7510 First PR so it might have issues, sorry in advance.

## Why It's Good For The Game
So that users can choose other UI styles and have them saved for when they log in later.

## Changelog
:cl:
fix: Fixed an issue where saving UI styles wouldn't work.
/:cl:

### DIFF
--- a/code/modules/client/preference/preferences_mysql.dm
+++ b/code/modules/client/preference/preferences_mysql.dm
@@ -50,7 +50,7 @@
 
 	//Sanitize
 	ooccolor		= sanitize_hexcolor(ooccolor, initial(ooccolor))
-	UI_style		= sanitize_inlist(UI_style, list("White", "Midnight","Plasmafire","Retro","Slimecore","Operative"), initial(UI_style))
+	UI_style		= sanitize_inlist(UI_style, list("White", "Midnight", "Plasmafire", "Retro", "Slimecore", "Operative"), initial(UI_style))
 	default_slot	= sanitize_integer(default_slot, 1, max_save_slots, initial(default_slot))
 	toggles			= sanitize_integer(toggles, 0, TOGGLES_TOTAL, initial(toggles))
 	toggles2		= sanitize_integer(toggles2, 0, TOGGLES_2_TOTAL, initial(toggles2))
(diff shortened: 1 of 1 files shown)


farie82:
`fixes #7510` that needs to be put in the `What Does This PR Do` part.
The changelog should contain text that is usable for people in-game when they look at the in-game changelog.

Dragonkiller93:
Is that good? Thanks for the help.

farie82:
> Is that good? Thanks for the help.

Yep, that works. Welcome to the codebase!

code/modules/client/preference/preferences_mysql.dm
```diff
@@ -50,7 +50,7 @@
 
 	//Sanitize
 	ooccolor		= sanitize_hexcolor(ooccolor, initial(ooccolor))
-	UI_style		= sanitize_inlist(UI_style, list("White", "Midnight"), initial(UI_style))
+	UI_style		= sanitize_inlist(UI_style, list("White", "Midnight","Plasmafire","Retro","Slimecore","Operative"), initial(UI_style))
```
farie82, line 53:
```suggestion
	UI_style		= sanitize_inlist(UI_style, list("White", "Midnight", "Plasmafire", "Retro", "Slimecore", "Operative"), initial(UI_style))
```
We put one space after each operator and `,` to separate values. It makes reading the code a bit easier.

farie82 (changes requested):

AffectedArc07 (approved):

PR #15909, opened 2021-04-18
Gives Greys FluoroSulphuric Acid immunity 

LordShekelstein:
## What Does This PR Do
This PR gives Greys FluoroSulphuric acid immunity. 

## Why It's Good For The Game
Greys not having FluoroSulphuric acid immunity seems like a major oversight in my opinion. 

Greys are already in an incredibly weak position with their water weakness. Water is ubiquitous across the station, and Greys can be easily killed by someone filling up a mug in the sink, shoving them over, and force feeding them water. Or, picking up a fire extingusher and spritzing them three times. This coupled by the fact that water deals almost the same amount of damage to Greys as flurosulphuric acid, seems rather unbalanced. 

While acid, on the other hand, is scantly used. Personally, I have yet to see sulphuric acid actually weaponized in any form other than botanists using deathnettle. While flurosulphuric acid on the other hand, while still barely used, is used in nuclear operative's grenades, syndiebees (dag nabbit botany!), and is used by crafty chemists. 

I think that this is a good balance change and wouldn't make Greys too po
## Changelog
:cl:
tweak: Makes an or statement involving facid, giving Greys fluorosulphuric acid immunity.
/:cl:


datlo:
This will only handle processing facid inside greys, but will still damage greys on contact. gotta add a isgrey check to facid's code, like sulph acid does.

SabreML (approved):
Code looks good, although you'll probably want to sort out your commit problem at some point.

Necaladun (approved):
I've weaponized sulfuric acid by drinking a glass of it and then offering the captain a sip. It's hilarious.

Greys need love, super approved.

AffectedArc07 (approved):
This will be the last PR of yours I merge until you sort your commit problem out

PR #15923, opened 2021-04-21
Sets a Mob Cap for Giant Spiders

Sirryan2002:
## What Does This PR Do
**this does not relate to terror spiders at all**
Adds a variable to SSmobs to track the number of giant spiders
Prevents Eggs `amount_grown` variable from incrementing when number of giant spiders is larger than 30 therefor halting hatching temporarily
When giant spiders count is larger than 40 spiderlings are just killed when they grow up instead of producing a spider

## Why It's Good For The Game
The number of Giant Spiders can snowball extremely quickly especially when assisted by players. This mob cap is intended to prevent griefing and people intentionally lagging the server. From my own observations, admins will usually have to use a CC event to eradicate giant spiders whether it's BSA'ing xenobiology or straight up mass deleting all giant spiders.

Additionally, this doesn't "harm" giant spiders as most eggs will halt hatching until the number of giant spiders drops back down and frankly, 40 giant spiders is more than enough to accomplish whatever goals they have in mind.

I was not sure what was the best way to implement this, if there is a better way to track the number of spiders I'll happily change the code to fit that. Additionally, the caps are just arbitrary numbers I thought was best, please contribute your thoughts as to what is a good balance for these.

## Changelog
:cl:
add: Places a hard 40 mob cap on Giant spiders(not terrors)
/:cl:


code/game/objects/effects/spiders.dm
```diff
@@ -67,16 +69,18 @@
...
-		var/num = rand(3, 12)
-		for(var/i in 1 to num)
-			var/obj/structure/spider/spiderling/S = new /obj/structure/spider/spiderling(loc)
-			S.faction = faction.Copy()
-			S.master_commander = master_commander
-			if(player_spiders)
-				S.player_spiders = 1
-		qdel(src)
+	if(SSmobs.giant_spiders <= SPIDER_SOFT_CAP) //eggs gonna chill out until there is less spiders
+		amount_grown += rand(0,2)
```
dearmochi, line 73:
```suggestion
		amount_grown += rand(0, 2)
```

code/game/objects/effects/spiders.dm
```diff
@@ -67,16 +69,18 @@
...
+
+		if(amount_grown >= 100)
+			var/num = rand(3, 12)
+			for(var/i in 1 to num)
+				var/obj/structure/spider/spiderling/S = new /obj/structure/spider/spiderling(loc)
+				S.faction = faction.Copy()
+				S.master_commander = master_commander
+				if(player_spiders)
+					S.player_spiders = 1
```
dearmochi, line 82:
```suggestion
					S.player_spiders = TRUE
```

code/game/objects/effects/spiders.dm
```diff
@@ -168,23 +172,27 @@
...
-					var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a giant spider?", ROLE_GSPIDER, TRUE, source = S)
-
-					if(candidates.len)
-						var/mob/C = pick(candidates)
-						if(C)
-							S.key = C.key
-							if(S.master_commander)
-								to_chat(S, "<span class='biggerdanger'>You are a spider who is loyal to [S.master_commander], obey [S.master_commander]'s every order and assist [S.master_commander.p_them()] in completing [S.master_commander.p_their()] goals at any cost.</span>")
-			qdel(src)
+			if (SSmobs.giant_spiders <= SPIDER_HARD_CAP)
```
dearmochi, line 175:
```suggestion
			if(SSmobs.giant_spiders <= SPIDER_HARD_CAP)
```

code/modules/mob/living/simple_animal/hostile/giant_spider.dm
```diff
@@ -56,6 +56,13 @@
 		for(var/obj/structure/spider/S in range(1, get_turf(src)))
 			return S
 
+/mob/living/simple_animal/hostile/poison/giant_spider/death(gibbed)
```
dearmochi, line 59: It might be simpler to just increment and decrement `SSmobs.giant_spiders` on `Initialize()`/`Destroy()` respectively.

dearmochi (changes requested):

code/modules/mob/living/simple_animal/hostile/giant_spider.dm
SteelSlayer: That does mean however, that you can just kill spiders until you have 40 dead bodies, and then they can't reproduce because `Destroy()` isn't called, which would be weird.

Sirryan2002: Is there a better place do this if not the death proc?

dearmochi: Right, giant spidders don't auto-gib like terrors do -- I'd probably keep `death()` then

Sirryan2002: Alright sounds like a plan.

Kyep (commented):
What's this?
An overabundance of spiders in the workplace?
My briefcase full of spiders...

More seriously, I am not sure about this change.
On the one hand, having a ton of giant spiders can be annoying, and there might be performance concerns.
On the other hand, its pretty much one of the coolest things xenobio can do. And a straight-up mob cap seems very artificial. 

So... not sure. Leaning towards yes, but not sure.

Sirryan2002:
> What's this?
> An overabundance of spiders in the workplace?
> My briefcase full of spiders...
> 
> More seriously, I am not sure about this change.
> On the one hand, having a ton of giant spiders can be annoying, and there might be performance concerns.
> On the other hand, its pretty much one of the coolest things xenobio can do. And a straight-up mob cap seems very artificial.
> 
> So... not sure. Leaning towards yes, but not sure.

Agreed on it being cool and fun, I just want to prevent them from getting grossly abundant. We could probably jump the max spiders to 60 or whatever y'all feel is best.

Kyep:
I've always felt that xenobio mobs should require some sort of upkeep so you can't just spam infinite numbers of them. However, that's out of scope for this PR.

Kyep (approved):

code/controllers/subsystem/mobs.dm
```diff
@@ -9,6 +9,7 @@ SUBSYSTEM_DEF(mobs)
 	var/static/list/clients_by_zlevel[][]
 	var/static/list/dead_players_by_zlevel[][] = list(list()) // Needs to support zlevel 1 here, MaxZChanged only happens when CC is created and new_players can login before that.
 	var/static/list/cubemonkeys = list()
+	var/giant_spiders = 0
```
AffectedArc07, line 12:
```suggestion
	/// The amount of giant spiders that exist in the world. Used for mob capping.
	var/giant_spiders = 0
```

Also finally a var scoped in the right place. This isn't aimed at you, just my hatred for people making globals when something can be scoped better.

code/game/objects/effects/spiders.dm
```diff
@@ -168,23 +172,27 @@
...
+
+						if(candidates.len)
+							var/mob/C = pick(candidates)
+							if(C)
+								S.key = C.key
+								if(S.master_commander)
+									to_chat(S, "<span class='biggerdanger'>You are a spider who is loyal to [S.master_commander], obey [S.master_commander]'s every order and assist [S.master_commander.p_them()] in completing [S.master_commander.p_their()] goals at any cost.</span>")
+				qdel(src)
+			else
+				Destroy()
```
AffectedArc07, line 195: Do not manually call `Destroy()`.

**Ever**

AffectedArc07 (changes requested):

code/game/objects/effects/spiders.dm
Sirryan2002: should I just drop the else and destroy() then here?

variableundefined, line 195: Remove this in favor of just a qdel. 

Do you want the spiderling to grow up and then automatically get deleted. Or just have their growth stalled?

variableundefined (changes requested):

code/game/objects/effects/spiders.dm
AffectedArc07: `qdel(src)` implies `src.Destroy()` anyway.

Just remove the else block

code/game/objects/effects/spiders.dm
```diff
@@ -168,23 +172,27 @@
...
-					var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a giant spider?", ROLE_GSPIDER, TRUE, source = S)
-
-					if(candidates.len)
-						var/mob/C = pick(candidates)
-						if(C)
-							S.key = C.key
-							if(S.master_commander)
-								to_chat(S, "<span class='biggerdanger'>You are a spider who is loyal to [S.master_commander], obey [S.master_commander]'s every order and assist [S.master_commander.p_them()] in completing [S.master_commander.p_their()] goals at any cost.</span>")
-			qdel(src)
+			if(SSmobs.giant_spiders <= SPIDER_HARD_CAP)
```
AffectedArc07, line 175: You can simplify this diff right down by simply doing
```dm
if(SSmobs.giant_spiders > SPIDER_HARD_CAP)
    return
```

Use early returns where possible, it makes code a lot less painful.

AffectedArc07 (changes requested):

Sirryan2002:
ok one moment while I figure out how the fuck I just broke this

code/game/objects/effects/spiders.dm
Sirryan2002: grow up and delete, the growth is already stalled for the eggs.

code/game/objects/effects/spiders.dm
```diff
@@ -67,16 +69,18 @@
...
-	if(amount_grown >= 100)
-		var/num = rand(3, 12)
-		for(var/i in 1 to num)
-			var/obj/structure/spider/spiderling/S = new /obj/structure/spider/spiderling(loc)
-			S.faction = faction.Copy()
-			S.master_commander = master_commander
-			if(player_spiders)
-				S.player_spiders = 1
-		qdel(src)
+	if(SSmobs.giant_spiders <= SPIDER_SOFT_CAP) //eggs gonna chill out until there is less spiders
```
AffectedArc07, line 72:
```suggestion
	if(SSmobs.giant_spiders > SPIDER_SOFT_CAP) //eggs gonna chill out until there is less spiders
		return
```

And un-indent the block below

AffectedArc07 (changes requested):

code/modules/mob/living/simple_animal/hostile/giant_spider.dm
```diff
@@ -56,6 +56,13 @@
...
+	. = ..()
+	if(!.)
+		return FALSE
+	SSmobs.giant_spiders--
+	return ..()
```
AffectedArc07, line 64: This double parent call (there's one a few lines above) seems off

Sirryan2002: Apologies for me not knowing. I'm checking to make sure the spider actually died here and if so we subtract the count from the SSmobs tracker variable. Or am I accidently calling the parent function twice by using `. = ..()` and `return ..()` ?

AffectedArc07: Yes, that is he issue. You need to remove line 64 entirely

Sirryan2002: Thank you, fixing now

Sirryan2002:
I believe everything has been taken care of.

AffectedArc07 (approved):

PR #15925, opened 2021-04-21
Remove unsafe areas from wizard teleport list

datlo:
## What Does This PR Do

This PR removes areas from the wizard teleport list that are usually unsafe and/or clogs the list. 
This also affects as a side effect the area selector for the Nuke Ops Assault Pod, and which areas you can be randomly teleported to if you're a non-abductor using the abductor teleporter. IMO, none of those areas will be really missed for these cases.

This does not affect any other cases of teleportation such as admin/ghost area teleporting, teleporting machinery, random teleportation by teleporter error, etc.

Areas removed from the teleport list : 

- Space, Solars and Toxins Test Site (No atmos, you can still teleport to solar control rooms)
- Turbine, Toxins Mixing, Incinerator (Can teleport you in the middle of a plasma fire)
- Atmospherics (Can teleport you in the middle of plasma/sleeping gas/etc, the distribution loop and control rooms are still accessible)
- All AI sats rooms (Usually guarded by turrets, and clogs the list right at the top with a bunch of rooms you arent usually interested to go for. I decided to lock out the entire sat for consistency and because it makes sense... it's not technically part of the station)
- AI upload (Guarded by turrets)
- Supermatter room (Unsafe atmos and there's an instant death machine in there)
- Scichem test chamber (Usually safe but it's often not, and you can just select scichem instead so not needed)
- Server coldroom (Unsafe atmos, you can still teleport to the safe side of the server room)

## Why It's Good For The Game

The teleport scroll/teleport wiz spell is quite clunky to use. 
Removing unsafe areas will both prevent unexperienced wizards from dying one minute in and also make it less annoying to use by having less irrelevant areas in the list, especially the AI satellite areas right at the top of the list.

## Images of changes

image

## Changelog
:cl:
tweak: Removed usually unsafe areas from the list of areas wizards can teleport to, such as space or turret protected areas.
/:cl:

### DIFF
--- a/code/game/area/ss13_areas.dm
+++ b/code/game/area/ss13_areas.dm
@@ -593,12 +593,12 @@ NOTE: there are two lists of areas in the end of this file: centcom and station 
 /area/atmos/control
 	name = "Atmospherics Control Room"
 	icon_state = "atmos"
-	no_teleportlocs = TRUE
+	no_teleportlocs = FALSE
 
 /area/atmos/distribution
 	name = "Atmospherics Distribution Loop"
 	icon_state = "atmos"
-	no_teleportlocs = TRUE
+	no_teleportlocs = FALSE
 
 //Maintenance
 /area/maintenance
--- a/code/game/area/ss13_areas.dm
+++ b/code/game/area/ss13_areas.dm
@@ -43,6 +43,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station 
 	outdoors = TRUE
 	ambientsounds = SPACE_SOUNDS
 	sound_environment = SOUND_AREA_SPACE
+	no_teleportlocs = TRUE
 
 /area/space/nearstation
 	icon_state = "space_near"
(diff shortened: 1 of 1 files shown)


LordShekelstein:
I think most of these are fine. However, sabotaging AI stuff on wizard is incredibly useful. I would assume most new wizards already know the hazards of the AI satellite though.

datlo:
I think the benefits of removing the AI sat rooms (which are REALLY annoying when you are trying to use teleport in a pinch) outweights the downside that a wizard might want to start there. You can still go to atmos maintenance and make your way to the satellite using jaunt and/or knock.

Sirryan2002:
I'm a fan of this, however I would like for non-turret AI rooms(think antechamber, tcomms) to be teleport friendly. Additionally, incinerator should not be removed because it is a good stealth spot for wizards and frankly the incinerator never actually gets turned on, especially during round start.

datlo:
Unfortunately antechamber isnt its own area. Basically the only area without turrets in AI sat is maintenance and telecomms... I'll think about it.

PR #15935, opened 2021-04-23
Fixes firing flamethrowers when you cannot fire guns

hal9000PR:
## What Does This PR Do
You can no longer fire flamethrowers if you cannot fire guns

Sleeping carp, golems etc.

## Why It's Good For The Game
Likely an oversight. Seems exploity to use a ranged weapon when you shouldn't be able to.

## Changelog
:cl:
Fix: flamethrowers now respect races and martial arts that cannot fire guns.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/flamethrower.dm
+++ b/code/game/objects/items/weapons/flamethrower.dm
@@ -68,6 +68,8 @@
 	. = ..()
 	if(flag)
 		return // too close
+	if(!user)
+		return
 	if(user?.mind?.martial_art?.no_guns)
 		to_chat(user, "<span class='warning'>[user.mind.martial_art.no_guns_message]</span>")
 		return
--- a/code/game/objects/items/weapons/flamethrower.dm
+++ b/code/game/objects/items/weapons/flamethrower.dm
@@ -72,7 +72,7 @@
 		to_chat(user, "<span class='warning'>[user.mind.martial_art.no_guns_message]</span>")
 		return
 	if(HAS_TRAIT(user, TRAIT_CHUNKYFINGERS))
-		to_chat(user, "<span class='warning'>Your meaty finger is much too large for the trigger guard!</span>")
+		to_chat(user, "<span class='warning'>Your meaty finger is far too large for the trigger guard!</span>")
 		return
 	if(user && user.get_active_hand() == src) // Make sure our user is still holding us
 		var/turf/target_turf = get_turf(target)
--- a/code/game/objects/items/weapons/flamethrower.dm
+++ b/code/game/objects/items/weapons/flamethrower.dm
@@ -70,13 +70,13 @@
 		return // too close
 	if(!user)
 		return
-	if(user?.mind?.martial_art?.no_guns)
+	if(user.mind?.martial_art?.no_guns)
 		to_chat(user, "<span class='warning'>[user.mind.martial_art.no_guns_message]</span>")
 		return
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -68,6 +68,12 @@
 	. = ..()
 	if(flag)
 		return // too close
+	if(user?.mind?.martial_art?.no_guns)
+		to_chat(user, "<span class='warning'>[user.mind.martial_art.no_guns_message]</span>")
+		return
+	if(HAS_TRAIT(user, TRAIT_CHUNKYFINGERS))
+		to_chat(user, "<span class='warning'>Your meaty finger is much too large for the trigger guard!</span>")
```
S34NW, line 75:
```suggestion
		to_chat(user, "<span class='warning'>Your meaty finger is far too large for the trigger guard!</span>")
```

minor nitpick

hal9000PR: This is from the chunkyfingers code directly. so I don't think I should change this without changing that aswell, and it is so minor it barely matters but, ¯\ _(ツ)_/¯

Wojz:
Flaming Carp is no more. Good riddance.

code/game/objects/items/weapons/flamethrower.dm
AffectedArc07: Change it please

hal9000PR: changed

code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -68,6 +68,12 @@
 	. = ..()
 	if(flag)
 		return // too close
+	if(user?.mind?.martial_art?.no_guns)
```
farie82, line 71: I don't quite see a case where `user` would be null in all honesty. I know they guarded it below as well but now it will cause an exception in `HAS_TRAIT` if it's null.
You could rewrite this proc so it checks the user beforehand. Or just leave the check out

hal9000PR: done

code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -68,6 +68,14 @@
 	. = ..()
 	if(flag)
 		return // too close
+	if(!user)
+		return
+	if(user?.mind?.martial_art?.no_guns)
```
farie82, line 73:
```suggestion
	if(user.mind?.martial_art?.no_guns)
```
`user` is already checked prior

code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -68,6 +68,14 @@
...
+		to_chat(user, "<span class='warning'>[user.mind.martial_art.no_guns_message]</span>")
+		return
+	if(HAS_TRAIT(user, TRAIT_CHUNKYFINGERS))
+		to_chat(user, "<span class='warning'>Your meaty finger is far too large for the trigger guard!</span>")
+		return
 	if(user && user.get_active_hand() == src) // Make sure our user is still holding us
```
farie82, line 79:
```suggestion
	if(user.get_active_hand() == src) // Make sure our user is still holding us
```
Can also remove the check here

farie82 (changes requested):

dearmochi (approved):

farie82 (approved):

AffectedArc07 (approved):

PR #15956, opened 2021-04-28
Adds grouped spawners. Makes spawning unique things easier.

farie82:
## What Does This PR Do
Makes it easier for mappers to add multiple spots for unique loot/monsters/etc. Or groups of enemies etc.
Allows you to both define spawners in code as well as in the map itself using the variable overrides.

You can make different spawners that are tied together with the same key. This allows you to spawn a variety of unique monsters only a set amount of times.
For example, you can spawn a prince or a princess terror spider using the same key. Or you want to make sure a gateway has 10 enemies total but you want them to randomly spawn in 20 possible spots. All possible using this.

Does not actually add something into the game itself. This PR is more meant for later use.
It does include an example spawner though.

## Why It's Good For The Game
It makes mapping in unique things easier and allows for more interesting map designs

## Images of changes
image
Two unique spawner groups. The top one is one using the base type which will spawn 3 pens max. Their `group_id` has the same value so they belong to the same group.
The bottom one has instances of the given example.

image
One of the possible outcomes

## Changelog
Pure technical thus far. Hopefully not for long though!


code/game/objects/effects/spawners/unique_spawner.dm
```diff
@@ -0,0 +1,129 @@
...
+ * How to use:
+ * Either place the base unique_spawner in the map and define the key_override to determine which spawners belong to one another.
+ * Or make a child instance of the unique_spawner and define the values there.
+ * When using the base version be sure to define both total_amount and max_per_spawner on only one of the spawners
+***/
+
+/obj/effect/spawner/unique_spawner
+	name = "unique spawner"
+	icon = 'icons/mob/screen_gen.dmi'
+	icon_state = "x2"
```
farie82, line 11: It currently uses the default icon. I ain't a mapper so I don't know what is handy here

code/game/objects/effects/spawners/unique_spawner.dm
```diff
@@ -0,0 +1,129 @@
...
+	/// The assoc list of unique spawners. Key = type or key_override, value = list of spawners of that type
+	var/static/list/spawners
+	/// Which spawners are chosen to spawn the objects
+	var/static/list/chosen_spawners
+	/// The list which holds the total_amount values for each spawner group
+	var/static/list/total_amount_for_spawners
+	/// The list which holds the max_per_spawner values for each spawner group
+	var/static/list/max_per_spawner_for_spawners
+
+/obj/effect/spawner/unique_spawner/Initialize(mapload)
```
AffectedArc07, line 30: **THANK YOU** 

If I see another spawner with `/New()` I am gonna be annoyed

farie82: Haha. I was tempted to use it but I found other ways to circumvent using it.

S34NW (approved):

AffectedArc07 (approved):

PR #15962, opened 2021-04-30
Example Config Sync

Kyep:
## What Does This PR Do
Makes example config match the live server's, with a few narrow exceptions:
- Not including live server passwords
- Not using the DB by default, or anything else that would mess up a new contributor just getting started with our codebase
- Not using the live server servername/MOTD/TOS/etc which might cause a game admin who was connected to both to confuse the live server with their test environment and accidentally reboot the live server because they thought it was their test/dev server

## Why It's Good For The Game
I'm tired of seeing cases where people on Github argue something on the basis of them having "tested" it on their local server... unaware that the live server uses a different config than the default test server config. Most recently, this happened with one of the cyborg PRs. More generally, example config should match production as closely as possible so that contributors aren't surprised when their stuff works differently in production due to config changes they did not know about.

## Changelog
:cl: Kyep
tweak: Made example config far better match production, to make development/testing easier.
/:cl:


dearmochi (approved):

config/example/config.txt
```diff
@@ -1,21 +1,21 @@
...
 ## Server tagline: This appears on the hub entry.
-#SERVER_TAG_LINE The Perfect Mix of RP & Action
+SERVER_TAG_LINE ParaCode Test
 
 ## Server extra features: This appears in the feature list on the hub entry.
-#SERVER_EXTRA_FEATURES medium RP, varied species/jobs/modes
+SERVER_EXTRA_FEATURES Medium RP, varied species/jobs.
+
+## Download the RSC from the web - this needs to be regenerated every time the code is changed.
+#RESOURCE_URLS http://www.paradisestation.org/windows/paradise.rsc.zip
```
AffectedArc07, line 11: I would put this in, since this is for test servers and their branch RSCs may be **majorly** different than the production RSC

config/example/config.txt
```diff
@@ -390,8 +385,8 @@ EXTRA_SPACE_RUIN_LEVELS_MAX 8
 ## Uncomment this to shut down the world any time it would normally reboot
 #SHUTDOWN_ON_REBOOT
 ## A command to run prior to the world shutting down, only used if the above option is enabled
-## This default value will kill Dream Daemon on Windows machines
-#SHUTDOWN_SHELL_COMMAND taskkill /f /im dreamdaemon.exe
+## This default value will kill Dream Daemon on Windows machines.
+SHUTDOWN_SHELL_COMMAND taskkill /f /im dreamdaemon.exe
```
AffectedArc07, line 389: Recomment this. Even though the above option is disabled, its better to have both disabled.

config/example/config.txt
```diff
@@ -447,16 +443,16 @@ BYOND_ACCOUNT_AGE_THRESHOLD 7
...
-#DISCORD_WEBHOOKS_MAIN_URL
+#DISCORD_WEBHOOKS_MAIN_URL https://discordapp.com/api/webhooks/.........................
 
 ## Webhook URLs for the admin discord webhook. Separate multiple URLs with a | (EG: https://url1|https://url2)
-#DISCORD_WEBHOOKS_ADMIN_URL
+#DISCORD_WEBHOOKS_ADMIN_URL https://discordapp.com/api/webhooks/.........................
 
 ## Webhook URLs for the mentor discord webhook. Separate multiple URLs with a | (EG: https://url1|https://url2)
-#DISCORD_WEBHOOKS_MENTOR_URL
+#DISCORD_WEBHOOKS_MENTOR_URL https://discordapp.com/api/webhooks/........................
```
AffectedArc07, line 455: Just leave all these as commented defaults since they arent really game parameters

config/example/rank_colors.txt
```diff
@@ -0,0 +1,11 @@
+# Use this file to denote colours for ingame admin ranks, using the following format
```
AffectedArc07, line 1: Remove this. https://github.com/ParadiseSS13/Paradise/pull/15855 isnt merged yet.

AffectedArc07 (changes requested):
Just a few things

config/example/config.txt
Kyep: Did you mean... that you would leave this blank? Also the reason its there is so that anyone usinjg our codebase can become familiar with how to use the option.

AffectedArc07: Yeah I meant `wouldn't`. Not used to this laptop's keyboard. If its commented out that should be fine though.

AffectedArc07 (approved):

PR #15966, opened 2021-05-01
CQC Leg Sweep Tweaks

matttheficus:
## What Does This PR Do

Changes CQC's Leg Sweep into a 1 cycle (1-2 second) Weaken instead of a 3 cycle Weaken, practically making it a trip rather than a full knockdown stun. It replaces the 2 cycles of Weaken with 40 Stamina Damage (minor slowdown for a few seconds), which will allow it to be a good tool for escape and evasion rather than a MASSIVELY strong opener. 

## Why It's Good For The Game

Having what amounts to an instant, one-button insta-stun is a tad nuts in the realm of our current Combat System. This should reduce the Leg Sweep > Sleeper Hold combo's effectiveness by giving the victim a small window to resist out of the grab if theyre fast enough.

## Changelog
:cl:
tweak: Leg Sweep weaken reduced to 1 cycle (2ish seconds) as well as dealing minor stamina damage
/:cl:

### DIFF
--- a/code/modules/martial_arts/cqc.dm
+++ b/code/modules/martial_arts/cqc.dm
@@ -50,8 +50,9 @@
 		D.visible_message("<span class='warning'>[A] leg sweeps [D]!", \
 							"<span class='userdanger'>[A] leg sweeps you!</span>")
 		playsound(get_turf(A), 'sound/effects/hit_kick.ogg', 50, 1, -1)
-		D.apply_damage(10, BRUTE)
-		D.Weaken(3)
+		D.apply_damage(5, BRUTE)
+		D.Weaken(1)
+		D.adjustStaminaLoss(40)
 		add_attack_logs(A, D, "Melee attacked with martial-art [src] : Leg sweep", ATKLOG_ALL)
 	return TRUE
 


Samman166:
so just use a stun prod instead? It's also a one click stun, i don't really get the craze about the leg sweep

SteelSlayer:
The numbers associated with weaken are not in seconds, they are in Life cycles. So the weaken time is going from 6 seconds to 2 seconds.

matttheficus:
> The numbers associated with weaken are not in seconds, they are in Life cycles. So the weaken time is going from 6 seconds to 2 seconds.

updated CL to use correct timings, forgot to convert it, whoops!

hal9000PR:
> so just use a stun prod instead? It's also a one click stun, i don't really get the craze about the leg sweep

You don't have a stun prod in perma. Stun prods can also be disarmed.

Wojz:
Just remove it.

LordShekelstein:
Eh, I don't mind this. I'd prefer if it was just removed but its a good step forwards.

Mitchs98:
> > so just use a stun prod instead? It's also a one click stun, i don't really get the craze about the leg sweep
> 
> You don't have a stun prod in perma. Stun prods can also be disarmed.
> 
> However, i don't agree with this change. Cqc leg sweep isn't powerful because it stuns, it's powerful because it combos into a 30 second sleep. This will however give it a chance to fail because depending on where you are in the cycle you might not have enough time to pull off that second punch. Rng in combat, especially something as deciding as a 30 second sleep isnt ideal.

Basically this. A full on removal of it would be the more ideal option.

dearmochi (approved):

variableundefined:
No approval from head or maint. Closing.

PR #15970, opened 2021-05-01
Removes duplicate code from handle_status_effects

### LINKED ISSUE #15916: dizziness, confused, drowsyness,  and jitteriness are being lowered two times per life tick for human mobs
**Issue Description**:
Dizziness is being reduced in two separate places for human mobs. Once in `/mob/living/carbon/handle_status_effects()` and once in `/mob/living/carbon/human/handle_chemicals_in_body()`
Resulting in a reduction of 4 dizziness per tick if you're not resting and a reduction of 20 if you are resting.

`/mob/living/proc/Life(seconds, times_fired)` is the one calling both calls

The same goes for the other effects

It only to reduced once.

**What happened instead**:
It got reduced multiple times. Causing weird behaviour.

It should only really be done once as this creates odd behaviour for this status effect

Get dizzy

Unsure. Might be caused by the life refactor a while back.

**Extra information**:
My expectation is that `handle_chemicals_in_body` should not handle this. A maint has to confirm this though @Fox-McCloud

Dragonkiller93:
## What Does This PR Do
Fixes #15916, removes duplicate code for dizziness, drowsiness, and jitteriness lowering from the handle_status_effects function. The same functionality seems to be already present in carbon/human/life/handle_chemicals_in_body.

## Why It's Good For The Game
Hopefully stops these status effects going down faster than they should.

## Changelog
:cl:Iwanabeu
fix: Drowsiness, Dizziness, Confusion, and Jitteriness time is decreased properly.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -314,12 +314,6 @@
 		do_jitter_animation(jitteriness)
 		AdjustJitter(-restingpwr)
 
-
-
-	//Jitteryness
-	if(jitteriness)
-		do_jitter_animation(jitteriness)
-
 	if(hallucination)
 		spawn handle_hallucinations()
 
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -263,7 +263,6 @@
 			setStaminaLoss(0, FALSE)
 			update_health_hud()
 
-	var/restingpwr = 1 + 4 * resting
 
 	//Dizziness
 	if(dizziness)
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -297,22 +297,12 @@
 						C.pixel_x -= pixel_x_diff
 						C.pixel_y -= pixel_y_diff
 			src = oldsrc
-		AdjustDizzy(-restingpwr)
 
-	if(drowsyness)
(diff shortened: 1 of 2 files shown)


Dragonkiller93:
Could someone help me with these tests? I have no idea what they mean

DeveloperLoser (commented):
I may be wrong because i'm pretty new to this but it seems like you just want to delete this to fix your error.

`code/modules/mob/living/carbon/life.dm:266:warning: restingpwr: variable defined but not used`

Can't do any testing myself but definitely do some things on your local server and make sure it doesn't blow up.

Dragonkiller93:
Okay, I followed mochi's advice and removed the code from carbon/human, I tested it on drowsiness and jitteriness and both seem to function correctly, not sure if there was a reason for the code in handle_chemicals_in_body(). If there was, this PR probably isn't needed.

code/modules/mob/living/carbon/life.dm
```diff
@@ -314,6 +314,12 @@
 		do_jitter_animation(jitteriness)
 		AdjustJitter(-restingpwr)
 
+
+
+	//Jitteryness
+	if(jitteriness)
+		do_jitter_animation(jitteriness)
+
```
dearmochi, line 322:
```suggestion
```

Unneeded. It's being done right above.

dearmochi (changes requested):

dearmochi (approved):

PR #15973, opened 2021-05-02
All playsound now respect client settings

ppi13:
## What Does This PR Do

This makes it so that Play Local Sound, Play Server Sound and Play Sound via Intercomms checks and respects client settings, while giving Play Local Sound a cancel window, as well as logging of mentioned verbs.

## Why It's Good For The Game

Consistency, and all those sounds should be mutable and volume adjustable

## Changelog
:cl:
add: Add client preference checks for the admin playsound verbs
/:cl:


code/modules/admin/verbs/playsound.dm
```diff
@@ -57,10 +70,16 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
-	var/melody = input("Select a sound from the server to play", "Server sound list") as null|anything in sounds
+	var/sound/melody = input("Select a sound from the server to play", "Server sound list") as null|anything in sounds
 	if(!melody)	return
 
-	play_sound(melody)
+	log_admin("[key_name(src)] played server sound [melody]")
+	message_admins("[key_name_admin(src)] played server sound [melody]", 1)
+
+	for(var/mob/M in GLOB.player_list)
+		if(prefs.sound & SOUND_LOBBY)
```
SabreML, line 80: Is the `SOUND_LOBBY` here intentional? All of the others are checking for `SOUND_MIDI`, which makes a lot more sense to me.
Also this is checking for the preferences of `src`, which is the admin in this case.
```suggestion
		if(M.client.prefs.sound & SOUND_MIDI)
```

code/modules/admin/verbs/playsound.dm
```diff
@@ -57,10 +70,16 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
+			SEND_SOUND(src, sound(melody, repeat = 0, wait = 0, volume = 85 * prefs.get_channel_volume(CHANNEL_LOBBYMUSIC), channel = CHANNEL_LOBBYMUSIC))
```
SabreML, line 81: [`repeat`](http://www.byond.com/docs/ref/info.html#/sound/var/repeat) and [`wait`](http://www.byond.com/docs/ref/info.html#/sound/var/wait) both default to `0`, so you shouldn't need them here.
```suggestion
			SEND_SOUND(src, sound(melody, volume = 85 * prefs.get_channel_volume(CHANNEL_LOBBYMUSIC), channel = CHANNEL_LOBBYMUSIC))
```

code/modules/admin/verbs/playsound.dm
```diff
@@ -44,9 +44,22 @@ GLOBAL_LIST_EMPTY(sounds_cache)
 	set name = "Play Local Sound"
 	if(!check_rights(R_SOUNDS))	return
 
-	log_admin("[key_name(src)] played a local sound [S]")
-	message_admins("[key_name_admin(src)] played a local sound [S]", 1)
-	playsound(get_turf(src.mob), S, 50, 0, 0)
+	var/sound/uploaded_sound = sound(S, repeat = 0, wait = 1, channel = CHANNEL_ADMIN)
```
SabreML, line 47: See above. (Also boolean)
```suggestion
	var/sound/uploaded_sound = sound(S, wait = TRUE, channel = CHANNEL_ADMIN)
```

code/modules/admin/verbs/playsound.dm
```diff
@@ -44,9 +44,22 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
+	uploaded_sound.priority = 250
+
+	GLOB.sounds_cache += S
+
+	if(alert("Are you sure?\nSong: [S]\nNow you can also play this sound using \"Play Local Sound\".", "Confirmation request" ,"Play", "Cancel") == "Cancel")
```
SabreML, line 52: Spacing is a bit off here.
```suggestion
	if(alert("Are you sure?\nSong: [S]\nNow you can also play this sound using \"Play Local Sound\".", "Confirmation request", "Play", "Cancel") == "Cancel")
```

code/modules/admin/verbs/playsound.dm
```diff
@@ -44,9 +44,22 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
+		return
+
+	log_admin("[key_name(src)] played local sound [S]")
+	message_admins("[key_name_admin(src)] played local sound [S]", 1)
```
SabreML, line 56: `message_admins()` doesn't seem to have a second argument, unless I'm missing something? (Same on the other occurrences.)
https://github.com/ParadiseSS13/Paradise/blob/58d7054e92a2a67adf1e78da8ad524fdba8f2504/code/modules/admin/admin.dm#L5-L10
```suggestion
	message_admins("[key_name_admin(src)] played local sound [S]")
```

code/modules/admin/verbs/playsound.dm
ppi13: This one only plays the lobby music and two? Three? other sounds. With is being lobby music, i figured use sound_lobby over sound_midi.

SabreML: That makes sense. I'd still suggest having it check the midi preference anyway since it *is* being played by an admin.
Something like this instead:
```suggestion
		if((M.client.prefs.sound & SOUND_LOBBY) && (M.client.prefs.sound & SOUND_MIDI))
```

Kyep (approved):

code/modules/admin/verbs/playsound.dm
```diff
@@ -57,10 +70,16 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
+	var/sound/melody = input("Select a sound from the server to play", "Server sound list") as null|anything in sounds
 	if(!melody)	return
 
-	play_sound(melody)
+	log_admin("[key_name(src)] played server sound [melody]")
+	message_admins("[key_name_admin(src)] played server sound [melody]")
+
+	for(var/mob/M in GLOB.player_list)
+		if(prefs.sound & SOUND_MIDI)
+			SEND_SOUND(src, sound(melody, volume = 85 * prefs.get_channel_volume(CHANNEL_ADMIN), channel = CHANNEL_ADMIN))
```
AffectedArc07, line 81: You do realise this just checks the admins client and sends it to them *hundreds* of times right?


You dont read from `M` in this loop.

Please actually test your code.

AffectedArc07 (changes requested):

code/modules/admin/verbs/playsound.dm
ppi13: Sorry, this is fixed now.

code/modules/admin/verbs/playsound.dm
```diff
@@ -57,10 +70,16 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
-	var/melody = input("Select a sound from the server to play", "Server sound list") as null|anything in sounds
+	var/sound/melody = input("Select a sound from the server to play", "Server sound list") as null|anything in sounds
 	if(!melody)	return
 
-	play_sound(melody)
+	log_admin("[key_name(src)] played server sound [melody]")
+	message_admins("[key_name_admin(src)] played server sound [melody]")
+
+	for(var/mob/M in GLOB.player_list)
+		if(prefs.sound & SOUND_MIDI)
```
AffectedArc07, line 80: You still havnt checked the targets sound pref, only your own

code/modules/admin/verbs/playsound.dm
```diff
@@ -102,4 +124,9 @@ GLOBAL_LIST_EMPTY(sounds_cache)
 			continue
 		if(!I.on && !ignore_power)
 			continue
-		playsound(I, melody, cvol)
+		for(var/mob/M in GLOB.player_list)
+			if(M.client.prefs.sound & SOUND_MIDI)
+				var/ivol = cvol * M.client.prefs.get_channel_volume(CHANNEL_ADMIN)
+				playsound(I, melody, ivol)
+
```
AffectedArc07, line 131: This is going to stack the sound on the intercom multiple times.

AffectedArc07 (changes requested):
Again, please actually test this code with multiple people on with different preferences.


The fact I had to ask you to test your code **twice** here isnt good.

code/modules/admin/verbs/playsound.dm
```diff
@@ -44,9 +44,22 @@ GLOBAL_LIST_EMPTY(sounds_cache)
...
+	if(alert("Are you sure?\nSong: [S]\nNow you can also play this sound using \"Play Local Sound\".", "Confirmation request", "Play", "Cancel") == "Cancel")
+		return
+
+	log_admin("[key_name(src)] played local sound [S]")
+	message_admins("[key_name_admin(src)] played local sound [S]")
+
+	for(var/mob/M in GLOB.player_list)
+		if(M.client.prefs.sound & SOUND_MIDI)
+			uploaded_sound.volume = 100 * M.client.prefs.get_channel_volume(CHANNEL_ADMIN)
+			SEND_SOUND(get_turf(src.mob), uploaded_sound)
```
AffectedArc07, line 61: Again. This is playing the same sound over and over. You need to send this to the mobs themselves in range.

Or, alternatively, just play the sound down the `CHANNEL_ADMIN` channel.

If I have to make a **fourth** review saying `test your code` I am going to lose it.

AffectedArc07 (changes requested):
See requested changes

variableundefined:
PR hasn't been updated. 

I'm objecting due to code quality issues. Closing.

PR #15982, opened 2021-05-04
Mentor Playtime Panel Tweaks

SabreML:
## What Does This PR Do

Adds Supply (Cargo) and Ghost hours to the Mentor/Admin 'Check Player Playtime' panel.
Also allows Mentors to open the individual 'Experience Panel', since it being blocked looks like an oversight.

## Why It's Good For The Game

**Playtime Panel:**
1. All of the other departments are shown here, so Supply being excluded is defintely an oversight.
2. There's no real *benefit* to showing ghost hours, but the Experience Panel shows them and they're pretty interesting to see, so why not have it here too.

**Experience Panel:**
It being unavailable for Mentors seems unintentional, considering that:
1. The proc is called `cmd_mentor_show_exp_panel()`.
2. The actual proc itself allows Mentors to view it, it's just the link that's blocked.
https://github.com/ParadiseSS13/Paradise/blob/58d7054e92a2a67adf1e78da8ad524fdba8f2504/code/modules/admin/topic.dm#L2632-L2634
https://github.com/ParadiseSS13/Paradise/blob/58d7054e92a2a67adf1e78da8ad524fdba8f2504/code/game/jobs/job_exp.dm#L88-L97
3. There's really no downside to allowing it that I can see. It's just a more concise version of the existing panel.

## Images of changes

**Before:**
dreamseeker_YaKBNtW9yL

**After:**
dreamseeker_CKA347zard

**Playtime Panel:**
dreamseeker_U7iZ5KrZot
*(Only just enabled playtime tracking so numbers are low.)*

## Changelog
:cl:
tweak: Added Supply and Ghost hours to the Mentor/Admin playtime panel.
tweak: Allowed Mentors to access the individual player playtime panel, on top of the server-wide one.
/:cl:

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -2630,7 +2630,7 @@
 		fax_panel(usr)
 
 	else if(href_list["getplaytimewindow"])
-		if(!check_rights(R_ADMIN))
+		if(!check_rights(R_ADMIN | R_MOD | R_MENTOR))
 			return
 		var/mob/M = locateUID(href_list["getplaytimewindow"])
 		if(!istype(M, /mob))
--- a/code/__DEFINES/preferences.dm
+++ b/code/__DEFINES/preferences.dm
@@ -94,7 +94,7 @@
 #define EXP_TYPE_SERVICE		"Service"
 #define EXP_TYPE_WHITELIST		"Whitelist"
 
-#define EXP_DEPT_TYPE_LIST		list(EXP_TYPE_SERVICE, EXP_TYPE_MEDICAL, EXP_TYPE_ENGINEERING, EXP_TYPE_SCIENCE, EXP_TYPE_SECURITY, EXP_TYPE_COMMAND, EXP_TYPE_SILICON, EXP_TYPE_SPECIAL)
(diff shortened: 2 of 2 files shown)


Wojz:
I have no idea why this isn't a thing in the first place. Gets in the way a lot when people ask for playtime and you're like haha yeah it just DOESN'T SHOW SUPPLY

code/modules/admin/topic.dm
```diff
@@ -2630,7 +2630,7 @@
 		fax_panel(usr)
 
 	else if(href_list["getplaytimewindow"])
-		if(!check_rights(R_ADMIN))
+		if(!check_rights(R_ADMIN | R_MOD | R_MENTOR))
```
AffectedArc07, line 2633: Wouldnt really call this a fix. This panel is used by admins too and also shows job unlocks.

SabreML: Yeah that's fair. I wasn't sure if fix or tweak would be more appropriate here since it seems (at least to me) like it was intended to be accessible to mentors in the first place.

Kyep: Frankly I think everyone should be able to view their own playtime stats.
And it only shows jobs unlocked via playtime, I believe.

Kyep (approved):

AffectedArc07 (approved):

PR #15987, opened 2021-05-04
Removes combat shotguns from cargo

S34NW:
## What Does This PR Do

This PR removes cargos ability to order combat shotguns, and increases riot shotgun crates from 50 to 80.
This PR does **not** remove combat shotguns in Gamma armoury, the only place crew will currently have access to them outside of direct admin spawning.

## Why It's Good For The Game

Lets look at some weapon statistics for common weapons (and a not-so-common one).

### Combat shotgun:
7 round capacity with no cooldown, 60 damage per shot (slugs).
Potential damage per magazine: 420 (if using slugs)
~4 second mag empty
DPS: 105

### Riot shotgun:
As above, but requiring skill in timing pumps to rapidly empty the clip.
~9 second mag empty (using my poor skills)
DPS: 46

### Autorifle:
20 round magazine, forced cooldown between shots, 20dmg per shot.
Potential damage per magazine: 400 
~11 second mag empty
DPS: 36

### Laser carbine: (Red ERT and above)
20 round magazine, 2 shot burst, 20dmg per shot. 
Potential damage per magazine: 400 burn damage, no chance to break bones.
~7 second mag empty
DPS: 57

From this list, you can clearly see that the combat shotgun is horrendously unbalanced, even the most unrobust of users can empty the mag super fast doing devastating damage. It's TWICE the DPS of a red ERT weapon (a team that _should, by definition_ be better equipped than most station-side personnel.) not to mention the ERT weapon only does burn damage and cannot break bones. Burn damage currently just requires an automender to fix, a mag of slugs WILL break a bone or cause IB.

Combat shotguns, when paired with slugs, are too strong for station crew to have outside of emergencies. By keeping them to gamma armoury, this stops the horrid cheese we've seen lately where antags such as biohazards are completely destroyed.
Skilled players can still use riot shotguns for a high damage weapon, and I'm sure can push the DPS higher than I could manage.

Regarding the riot shotgun crate price increase, this is to bring it in line with the cost vs. DPS that we see on the other gun crates, plus it gets more guns per crate than the other lethal weapons cargo can order.

## Changelog
:cl:
del: Removed combat shotgun crates from cargo
tweak: Riot shotgun crates now cost 80 points, not 50.
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -391,20 +391,8 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 					/obj/item/storage/belt/bandolier,
 					/obj/item/storage/belt/bandolier,
 					/obj/item/storage/belt/bandolier)
-	cost = 50
-	containername = "riot shotgun crate"
-
-/datum/supply_packs/security/armory/ballisticauto
-	name = "Combat Shotguns Crate"
-	contains = list(/obj/item/gun/projectile/shotgun/automatic/combat,
-					/obj/item/gun/projectile/shotgun/automatic/combat,
-					/obj/item/gun/projectile/shotgun/automatic/combat,
-					/obj/item/storage/belt/bandolier,
-					/obj/item/storage/belt/bandolier,
-					/obj/item/storage/belt/bandolier)
 	cost = 80
-	containername = "combat shotgun crate"
-
+	containername = "riot shotgun crate"
 /datum/supply_packs/security/armory/buckshotammo
 	name = "Buckshot Ammo Crate"
 	contains = list(/obj/item/ammo_box/shotgun/buck,


AffectedArc07:
Theres more to guns than just their raw DPS. An important part of shotguns, is the fact they are a shotgun and have an AOE spread as opposed to just a single bullet (Yes I know slugs exist but these are loaded with buckshot by default).

S34NW:
> 
> 
> Theres more to guns than just their raw DPS. An important part of shotguns, is the fact they are a shotgun and have an AOE spread as opposed to just a single bullet (Yes I know slugs exist but these are loaded with buckshot by default).
> 
> TLDR: Not in favour

The problem is nobody uses buckshot, they unload next to the nearest lathe and put slugs in. Observe it happen at any round involving nukies, terrors, or anything else that warrants mass shotgun rollout.
Only baldies keep buckshot in combats, because they are worse by almost every metric. The _only_ advantage they have is being easier to hit the target with at least some of the shot because of spread.
If you offer a player a combat shotgun vs any other cargo item, there will not be any hesitation. Having one item be the default best is not good balance, players choosing between riots and autorifles would be a thing, because each has an advantage over the other with neither being straight up "better".

Rurik123:
I don't see this changing much aside from letting nukies win easier on lowpop and slaughter demons roam free without fear of getting deleted in .2 seconds from a combat shotty. The latter could be argued to be good, but I'm not convinced that's worth the price of the former.

Xenos/Terrors will likely remain the same. Literally ever crewmember and their mother having AEGs (even after the threat is gone) is the problem there. Not shottys. 

You might have better luck touching upon slugs themselves if its DPS you're worried about, especially since there was a (shortsighted imo) pr a while back that nerfed buckshot while leaving the already superior slugs the same.

matttheficus:
I agree with @Rurik123, the issue isnt the Combat Shotguns themselves, but an issue with Slugs WITH the Combat Shotguns. In all honesty, I'd prefer to see the Slugs get some sort of change. At them moment, there is no difference in code between a .357's shot and a Slug, as they both use /obj/item/projectile/bullet. Perhaps make a custom projectile for it that can be balanced accordingly without affecting the balance of an Antag Weapon (in this case the .357, which already has a hefty TC cost).

hal9000PR:
Combat shotguns are fine when they have buckshot in. Why? because buckshot is balanced, high DPS but countered by armour. 

The clear solution here is nerfing the other ammo types. Slugs, rubber and beanbags are all ridiculously strong. Nerfing these would make the game a lot healthier - however I do see this buffing vampires which is non ideal. Ideally combat shotguns would stay, but the ammo types would be nerfed. But as a said this has the side effect of buffing vampires which are in *desperate* need of reworking. 

TLDR, nerf the ammo not the gun. But only once vampires get reworked.

Mitchs98:
Pretty much agreeing with the above. Though I don't agree rubber rounds or beanbag rounds need to be nerfed in any capacity, as it would only encourage security to use lethals more than they already do with shotguns(which...I honestly feel is a bad thing, people use lethals a fair bit too much imo as is).

LightFire53:
While I'm all for the removal of combat shotguns, a common point seems to be jumping out here:

> deleted in .2 seconds from a combat shotty.

> with no cooldown,

Even with a small cooldown thrown in, you drastically reduce the time to kill/maim/seriously injure. Pair it with a nerf for slugs (maybe how many you can actually fit in a mag?), and you get some balance alternatives for those who wish to keep it in?

S34NW:
> 
> 
> While I'm all for the removal of combat shotguns, a common point seems to be jumping out here:
> 
> > deleted in .2 seconds from a combat shotty.
> 
> > with no cooldown,
> 
> Even with a small cooldown thrown in, you drastically reduce the time to kill/maim/seriously injure. Pair it with a nerf for slugs (maybe how many you can actually fit in a mag?), and you get some balance alternatives for those who wish to keep it in?

This idea is actually quite nice, as it would give the riot shotgun a chance to become a more "high skill" weapon vs the ease of using a combat.

I'll let this gather feedback for a bit before potentially reworking it and touching on slugs, to be more _"improve don't remove"_

MRgaty:
I think shotguns are fine and they are not an issue like trasers. Not a fan of this.

S34NW:
closing in favour of a more nuanced balance pass

PR #15991, opened 2021-05-05
Fuel tanks now don't blow up 2-3 times when blown up. Their fire doesn't go through windows. And more fiery fixes

farie82:
## What Does This PR Do
Fixes/changes a few things
- Fuel fireflashes and other fireflashes using `fireflash_sm` will now respect windows etc. They won't go where atmos can't get through
- Fuel tanks now don't blow up 2-3 times and instead just blow up once. Same goes for other fireflashes like burning up oil in a beaker.
- Some logging fixes regarding fire flashes. Mostly runtimes
- hotspots now only affect items that are not `QDELETED`. This could happen due to earlier `fire_act`s

Profiler data:
```
New
/obj/structure/reagent_dispensers/fueltank/do_boom                                      0.000        0.088        0.088        0.000            4
Old
/obj/structure/reagent_dispensers/fueltank/boom                                         0.000        0.181        0.183        0.000            8
```
Note the boom is called 8 times for the old version. It was only called 4 times in total on 4 fuel tanks in different locations (same over the versions). But the other 4 times were because the fireflash detonated the fuel tank again

Thing that remains open is the `fireflash` proc. Which currently does not respect windows etc. These are less common though.

## Why It's Good For The Game
Mostly bug fixing.
The fuel tank not blowing up 3 times will reduce the danger of blowing up fuel tanks by a bit. The fireflashes not going through windows will also affect balance. It was intended though it seemed in the code

## Images of changes
image
Example of how explosions of fuel tanks now go. My ghost was at the place of detonation

## Changelog
:cl:
fix: Fuel tanks blowing up multiple times when they blow up.
fix: Most fireflashes don't go through windows etc now
/:cl:


code/modules/atmospherics/environmental/LINDA_fire.dm
```diff
@@ -295,15 +295,13 @@
 
 		if(T.density)
 			continue
-		for(var/obj/O in T)
```
farie82, line 298: This code is what I think should have checked if the fire could pass a turf... But as you can see it was just a loop that did nothing at all.

Wojz:
Oh thank fuck for that

Mitchs98:
No more getting lit on fire in the hallway from a fuel tank explosion. I like it.

PidgeyThePirate:
I was wondering why temperature spiked so high after being hit by a welder bomb while doing my PR. That explains a lot, really.

code/modules/reagents/chemistry/reagents/pyrotechnic.dm
```diff
@@ -56,10 +56,11 @@
 /datum/reagent/napalm/reaction_temperature(exposed_temperature, exposed_volume)
 	if(exposed_temperature > T0C + 100)
 		var/radius = min(max(0, volume * 0.15), 8)
-		fireflash_sm(get_turf(holder.my_atom), radius, rand(3000, 6000), 500)
+		var/turf/T = get_turf(holder.my_atom)
 		fire_flash_log(holder, id)
 		if(holder)
 			holder.del_reagent(id)
+		fireflash_sm(T, radius, rand(3000, 6000), 500)
```
SabreML, line 63: Why not just leave the `get_turf()` on the fireflash line? It's not being used anywhere else so I don't think it needs to be set as a variable.
```suggestion
 		fire_flash_log(holder, id)
 		if(holder)
 			holder.del_reagent(id)
		fireflash_sm(get_turf(holder.my_atom), radius, rand(3000, 6000), 500)

```
Same below as well.

farie82: `holder` can be null at that point due to line 62. Removing the reagent from the holder will IIRC qdel the reagent.

dearmochi (approved):

code/modules/atmospherics/environmental/LINDA_fire.dm
SteelSlayer: Does it really do nothing? It looks like its suppose to prevent fire spreading to turfs which have dense objects on them.

farie82: It indeed looks like that. The issue it has is that it calls `continue` while it is in a `for` loop itself. Thus going to the next `O in T` instead of continuing the while loop
As a result this loop does nothing but continue if `O.density` is true. And continue if it is false

SteelSlayer: Right, its a continue inside a for loop.. for some reason I was thinking it was going to continue the while. Yes this is useless.

variableundefined:
@farie82 Please deconflict

variableundefined (approved):

PR #15996, opened 2021-05-06
Optimizes handle_mutations_and_radiation. No more looping 55 times per human to do nothing

farie82:
## What Does This PR Do
Optimizes `handle_mutations_and_radiation` so it only loops over gene mutations that are actually active instead of looping over all possible gene mutations.

Test case involved gibbing all carbons first using `CALL gib() ON /mob/living/carbon`. Then spawning 101 unathis.
Profiler stats:
```
Proc Name                                                   Self CPU    Total CPU    Real Time     Overtime      Calls
Old
/mob/living/carbon/human/handle_mutations_and_radiation      0.140        0.192        0.190        0.000         1717
New
/mob/living/carbon/human/handle_mutations_and_radiation      0.009        0.012        0.012        0.000         2100
```
In this test case it saved a fuckton of self CPU and total CPU (it calls some procs in there) usage. Which was the goal here.

Profiler data from the live server:
```
Proc Name                                                  Self CPU    Total CPU    Real Time     Overtime        Calls
/mob/living/carbon/human/Life                              4.797        228.773        229.34        0.001        688526
/mob/living/carbon/human/handle_mutations_and_radiation    13.211        21.164        21.221        0.009        461744
```
Best case the total CPU cost will lower to about 2-3. This will mean about a 7% decrease in Life total CPU usage (rough estimates)

## Why It's Good For The Game
Saves the CPU some load without sacrificing features.

## Changelog
Purely technical

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -14,6 +14,7 @@ GLOBAL_LIST_INIT(assigned_mutation_blocks, new(DNA_SE_LENGTH))
 // Used to determine what each block means (admin hax and species stuff on /vg/, mostly)
 GLOBAL_LIST_INIT(assigned_blocks, new(DNA_SE_LENGTH))
 
+/// Assoc list. Key = type of the mutation, value = the mutation
 GLOBAL_LIST_EMPTY(dna_mutations)
 
 GLOBAL_LIST_EMPTY(good_blocks)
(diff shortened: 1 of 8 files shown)


dearmochi (approved):

Fox-McCloud:
Marginal, but I'll take it

PR #16000, opened 2021-05-07
Removed someone's old email from the code

TwoCam:
## What Does This PR Do

This PR removes someone's email address from the newscaster error message and replaces it with a link to our Github issues page.
This is my first PR, so please feel free to tell me if I've done anything wrong.
## Why It's Good For The Game

We'll be more likely to get error reports instead of having them sent to an unknown Gmail address. Furthermore, there will be one less Gmail address the game.

## Changelog
:cl: 
tweak: Changed a rare error message on newscasters.
/:cl:

### DIFF
--- a/code/modules/newscaster/obj/newspaper.dm
+++ b/code/modules/newscaster/obj/newspaper.dm
@@ -115,7 +115,7 @@
 				dat+= "<hr><div style='float:left;'><a href='?src=[UID()];prev_page=1'>Previous Page</a></div>"
 			else
 				// No trailing punctuation so that it's easy to copy and paste the address
-				dat += "We're sorry to break your immersion, but there has been an error with the newscaster. Please report this error, along with any more information you have, to https://github.com/ParadiseSS13/Paradise/issues/new?assignees=&labels=&template=bug_report.md&title="
+				dat += "We're sorry to break your immersion, but there has been an error with the newscaster. Please report this error, along with any more information you have, to [config.githuburl]/issues/new?template=bug_report.md"
 
 		dat += "<br><hr><div align='center'>[curr_page+1]</div>"
 		human_user << browse(dat, "window=newspaper_main;size=300x400")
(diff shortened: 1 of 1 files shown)


S34NW:
Not sure about hardcoded URLs... Perhaps there is some other way?

TwoCam:
> Not sure about hardcoded URLs... Perhaps there is some other way?

I suspect that GitHub will go out of its way to keep these URL's pretty stable for this exact reason. It can go into the database or config if we want that though.

S34NW:
The GitHub address is defined in `config.txt` as `GITHUBURL` so that's always an option.

code/modules/newscaster/obj/newspaper.dm
```diff
@@ -114,7 +114,8 @@
 					dat += "<br><i>There is a small scribble near the end of this page... It reads: \"[scribble]\"</i>"
 				dat+= "<hr><div style='float:left;'><a href='?src=[UID()];prev_page=1'>Previous Page</a></div>"
 			else
-				dat += "i'm sorry to break your immersion. This shit's bugged. Report this bug to Agouri, polyxenitopalidou@gmail.com"
+				// No trailing punctuation so that it's easy to copy and paste the address
+				dat += "We're sorry to break your immersion, but there has been an error with the newscaster. Please report this error, along with any more information you have, to https://github.com/ParadiseSS13/Paradise/issues/new?assignees=&labels=&template=bug_report.md&title="
```
farie82, line 118: `config.githuburl` can be used here instead.

TwoCam: My only concern there is that we could wind up sending non-programmers to a big scary Github page if we do that.
If there's a consensus, I could create `config.githuperrorreporturl` instead. (Although I'd prefer `config.githubErrorReportUrl`.)

farie82: That would be good. Could also be used by admins/mentors to forward people to make an issue etc.
I'd recommend using snake_case for the naming btw

AffectedArc07, line 118:
```suggestion
				dat += "We're sorry to break your immersion, but there has been an error with the newscaster. Please report this error, along with any more information you have, to [config.githuburl]/issues/new?template=bug_report.md"
```

Give this a try.

TwoCam: There is a config freeze due to the ongoing conversion to TOML. I'm going with AffectedArc's solution instead

dearmochi (approved):

datlo:
I feel like this should just be removed entirely and replaced by an IC error message to contact Nanotrasen System Administrators, but that's just my opinion.

AffectedArc07 (approved):

PR #16004, opened 2021-05-08
Fixes self surgery with beds

hal9000PR:
## What Does This PR Do
You now drop items when you get buckled to beds so it is more consistent with resting.

## Why It's Good For The Game
You shouldn't be able to hold things if you cannot pick things up and you can exploit this to find where/if you have internal bleeding.

## Changelog
:cl:
fix: you can no longer start a  surgery on yourself with beds.
fix: buckling on a bed drops what you are holding
/:cl:

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/bed.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/bed.dm
@@ -25,11 +25,6 @@
 	buckle_offset = -6
 	var/comfort = 2 // default comfort
 
-/obj/structure/bed/post_buckle_mob(mob/living/M)
-	if(iscarbon(M))
-		var/mob/living/carbon/C = M
-		C.drop_l_hand()
-		C.drop_r_hand() // no self surgery
 
 /obj/structure/bed/psych
 	name = "psych bed"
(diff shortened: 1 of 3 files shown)


Carthusia:
Self surgery would be kind of nice, but make it have high fail chances so its more of a last ditch effort thing

AffectedArc07:
> Self surgery would be kind of nice, but make it have high fail chances so its more of a last ditch effort thing

No

dearmochi (commented):
I would probably move the fix up to anything that uses `buckle_lying`. Look at the following for ideas:

https://github.com/ParadiseSS13/Paradise/blob/60edf50279660a13c55f5e70d32ab825450780df/code/modules/mob/living/update_status.dm#L86-L98

hal9000PR:
image
okay so it turns out you can start the surgery but you cannot actually do any of the steps. Still a bug however and inconsistent. if you cannot pick up items you shouldn't be able to hold them.

ppi13:
You can't do any surgeries in the first place when strapped to a bed.

code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -25,6 +25,7 @@
 	buckle_offset = -6
 	var/comfort = 2 // default comfort
 
+
```
hal9000PR, line 28: oops

dearmochi (approved):

AffectedArc07 (approved):

Varghh:
I just realized this kills roller bed + fire extinguisher rocket program

hal9000PR:
oh fuck- REVERT IT

PR #16021, opened 2021-05-13
Fixes #16019

### LINKED ISSUE #16019: Infinite Kisses Bug
Doing the kiss emote puts a kiss in your hand. By clicking on the sprite with your free hand, it is switched to your other hand, but makes it into an invisible kiss "item." You can hit people harmlessly with this kiss "item," drop it like a normal item to remove it, or click the air in front of your player to infinitely shoot kisses. This can easily spam the chat log.

### LINKED ISSUE #16036: Changelings can drop their armblades when shifting into lesser form
What is the problem?
Changelings can drop their armblades when shifting into lesser form allowing anyone else to pick it up and having it bond to their hand.
Anyone can pick up an armblade by accident and be gunned down without anyone knowing they picked it off the ground and there is also no evidence that this occured and it's supposed to be impossible.

Again, Security will kill innocent players because they picked this armblade up as the changeling has vented away unknowingly caused another player to die because of this accident.

1. Take out armblade
2. Use Lesser form.
3. Armblade is forced to be dropped to the ground.
When did the problem start happening
I have no idea, it may of been a thing for a long time since lesser form and armblade came in and it just took this long for it get noticed? Maybe some new code was brought in to cause this?

ppi13:
## What Does This PR Do

Fixes #16019 
Partially fixes #16036 by making people unable to pickup the armblade

## Why It's Good For The Game

No more machine gun kissing.
No more people picking up cling armblades.

## Changelog
:cl: ppi
fix: Fixes the ability to fire infinity kisses as fast as you can click
fix: Fixes it so people can't pick up a changeling armblade if droped
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -288,6 +288,9 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 		if(!user.unEquip(src, silent = TRUE))
 			return 0
 
+	if(flags & ABSTRACT)
+		return 0
+
 	else
 		if(isliving(loc))
 			return 0


hal9000PR:
well now you cannot get rid of them without kissing

Wojz:
Now make ranged holopara do this instead of crystal spray

code/game/objects/items/hand_item.dm
```diff
@@ -51,7 +51,7 @@
 	item_state = "nothing"
 	force = 0
 	throwforce = 0
-	flags = DROPDEL | ABSTRACT
+	flags = DROPDEL | ABSTRACT | NODROP
```
SabreML, line 54: I mean this *works*, but it does mean that there's no way to cancel kissing someone.
Also makes the `DROPDEL` flag a bit redundant.

S34NW:
I feel like the better way to do this is to make you unable to move abstract items between hands

dearmochi:
> 
> 
> I feel like the better way to do this is to make you unable to move abstract items between hands

Agreed. I would look at checking for the `ABSTRACT` flag when picking up items, i.e. in `/obj/item/attack_hand()`

ppi13:
Every other abstract item already got nodrop by default, this is the oddman out, but this here change should fix it, and any switching of hands of abstract items.

code/game/objects/items/hand_item.dm
```diff
@@ -68,6 +68,11 @@
 	blown_kiss.fire()
 	qdel(src)
 
+/obj/item/kisser/attack_self(mob/living/user)
+	to_chat(usr, "<span class='notice'>You lower your kiss-blowing hand!</span>")
+	qdel(src)
+	return
```
dearmochi, line 74: Again, I personally think from an user perspective it makes more sense to retain the drop functionality to cancel the emote, rather than using it.

code/modules/mob/inventory.dm
```diff
@@ -43,7 +43,7 @@
 /mob/proc/put_in_l_hand(obj/item/W)
 	if(!put_in_hand_check(W))
 		return 0
-	if(!l_hand && has_left_hand())
+	if(!l_hand && has_left_hand() && !ABSTRACT)
```
dearmochi, line 46: Absolutely not. `ABSTRACT` is a number (64) so doing `!64` will always evaluate to false, preventing you from putting *any* item in your hand. To check if an item has a flag, you need to do `W.flags & ABSTRACT`.

Either way, I think this is the wrong proc to put the check in. Switching the hand holding an item occurs in `/obj/item/attack_hand()`, so do take a look at that.

code/modules/mob/inventory.dm
```diff
@@ -57,7 +57,7 @@
 
 //Puts the item into your r_hand if possible and calls all necessary triggers/updates. returns 1 on success.
 /mob/proc/put_in_r_hand(obj/item/W)
-	if(!put_in_hand_check(W))
+	if(!put_in_hand_check(W) && !ABSTRACT)
```
dearmochi, line 60: Ditto

dearmochi (changes requested):

Drobe50:
This can also be done with the slap hand on a table, does this fix that as well?

ppi13:
slapping on a table 3 times consumes the slap 👏 
also everything should be a working now.

dearmochi (approved):

AffectedArc07 (approved):

PR #16024, opened 2021-05-13
Removes the construct shell from the beach gateway

Qwertytoforty:
## What Does This PR Do

This PR removes the construct shell from the beach gateway.

## Why It's Good For The Game

This one might be a bit contriversal.
The chaplain has a holy soul stone, as most people know, that lets them turn 1 person into a shade, if they are willing. This has no issue. The main issue, is when a chaplain gets ahold of a construct shell, and makes an artificer. Suddenly, the chaplain can now create endless holy constructs that are loyal to chaplain, and are just as strong as normal constructs, with the same sprite, so long as there are willing bodies. Generaly, there are quite a few of them. While one holy construct is fine, an army of them can become quite a mess. Often, one of a few situations happen.

The chaplain gets an army of constructs, and it turns out to be a cult round. This leads to confusion on crew and cult side, of which constructs are loyal, and often leads to one side (usualy cult) being suddenly attacked by a construct they did not expect to be hostile, and ahelps to be made. It often also leads to validhunting jaunting constructs that find bases with ease, or juggernaughts smashing down doors everywhere.

The constructs see a crewmember attacking a artificer for building cult stuff in their bar / library / office / bridge, and then attack the crewmember. Other crewmemebers see this, start attacking constructs, constructs attack back, becomes a huge mess.

The constructs are used by a traitor chaplain, and crew gets confused when the "holy" constructs attack security, as they think they are loyal to the crew. This often again leads to a mess as both sides attack each other.

To avoid this issue, I propose removing the spawn in the beach gateway, leading to construct shells only being avalible to the chaplain if they get a shell from the cult, a wizard, or the admins, instead of showing up every time the beach gateway exists, potentialy screwing over modes like cult before they even really get started.

## Changelog
:cl:
del: Removed the construct shell from the beach gateway.
/:cl:

### DIFF
--- a/_maps/map_files/RandomZLevels/beach.dmm
+++ b/_maps/map_files/RandomZLevels/beach.dmm
@@ -18,9 +18,6 @@
 	},
 /turf/simulated/floor/beach/away/water/deep/sand_floor,
 /area/awaymission/undersea)
-"af" = (
-/turf/simulated/floor/beach/away/water/deep/wood_floor,
-/area/awaymission/undersea)
 "ag" = (
 /obj/structure/chair/wood/wings,
 /turf/simulated/floor/beach/away/water/deep/wood_floor,
@@ -46311,7 +46308,7 @@ ad
 ad
 ad
 ah
-af
+ah
 ah
 ah
 ar
@@ -46822,7 +46819,7 @@ av
 av
 ad
 ad
-af
+ah
 ah
 ah
 ah
@@ -48107,7 +48104,7 @@ av
 av
 ad
 ad
-af
+ah
 ah
 ah
 ah
@@ -48624,7 +48621,7 @@ ad
 ad
 ad
 ah
-af
+ah
 ah
 ah
 ah
--- a/_maps/map_files/RandomZLevels/beach.dmm
+++ b/_maps/map_files/RandomZLevels/beach.dmm
@@ -19,7 +19,6 @@
 /turf/simulated/floor/beach/away/water/deep/sand_floor,
 /area/awaymission/undersea)
 "af" = (
-/obj/structure/constructshell,
 /turf/simulated/floor/beach/away/water/deep/wood_floor,
 /area/awaymission/undersea)
 "ag" = (


_maps/map_files/RandomZLevels/beach.dmm
```diff
@@ -19,7 +19,6 @@
 /turf/simulated/floor/beach/away/water/deep/sand_floor,
 /area/awaymission/undersea)
 "af" = (
-/obj/structure/constructshell,
```
AffectedArc07, line 22: Please mapmerge this properly instead of making a manual map edit (key `"ah"` is the exact same as this).

AffectedArc07 (changes requested):

S34NW (approved):
latest commit looks good, no mapmerge issues

Darkmight9:
Note to self, remember if this gets merged or not when it comes time to work on the beach revamp again.

_maps/map_files/RandomZLevels/beach.dmm
Qwertytoforty: map merged

Gatchapod:
While this is definitely the easiest solution, I do wonder if giving holy constructs unique sprites could solve the issue as well. Naturally, the problem is actually making those hypothetical sprites.

dearmochi (approved):

Stokes52:
> While this is definitely the easiest solution, I do wonder if giving holy constructs unique sprites could solve the issue as well. Naturally, the problem is actually making those hypothetical sprites.

I agree. I would be in favor of adding unique sprites to holy constructs since that will solve the actual problem instead of the band aid that removing the beach construct shell is.

Personally, I think the chaplain being able to make constructs when RNG makes the beach available is fun and cool and it's one of those little "secrets" that can make certain rounds interesting. Holy constructs are already rarely used.

I would be in favor of the following changes:
# Short term: Resprite holy constructs to a silver/blue scheme to make them different from the cult versions
# Medium term: Make new, unique, thematic sprites for a holy construct to make them different
# Long term: Perhaps consider making holy constructs mechanically different from cult constructs, to make them fit better with the crew, but that's a whole new discusion.

I would be willing to work on new sprites for holy constructs when I have time for it if that sounds like a good solution to others.

S34NW:
> 
> I agree. I would be in favor of adding unique sprites to holy constructs since that will solve the actual problem instead of the band aid that removing the beach construct shell is.

>
>I do wonder if giving holy constructs unique sprites could solve the issue as well. 

Counterargument, sprites being the same is an issue, but it is not _THE_ issue. While I'm in favour of doing that, chaplain constructs need to be reworked more generally.
They should not be as strong as cult constructs.
They should have a large message upon becoming a construct informing them that they cant just go murder people without consequences.
In an ideal world, they would be completely unique with different abilities, and then you could happily re-add the shells to the beach and it would not cause issues.

Stokes52:
> In an ideal world, they would be completely unique with different abilities, and then you could happily re-add the shells to the beach and it would not cause issues.

Yeah ultimately I think we agree that a general rework is the best solution.  In the meantime, removing the shell and/or recoloring the constructs is a good short term solution imo. The beach isn't the only place to get the shells after all.

Varghh:
Good change

Kyep (approved):

AffectedArc07 (approved):

KozackWolf:
I think the best solution, Holy constructs need sprite reworks, and also stat and ability reworks. I'd also suggest objectives being given to them along the lines of "You are a holy being, and are forbidden from committing evil acts." Similar to borg laws. They should also not be taking orders from the chaplain, because the chaplain is just a vessel of the holy being that is granting them their power. 

Question also. Do you get cult metal when you destroy alters, forges, etc... back? I know you don't from pylons, but curious about the others.

hal9000PR:
1) constructs getting laws just promotes valid hunting.

2) no they don't get metal back

PR #16028, opened 2021-05-15
Borgs can't use onclick when locked stunned, weakened or paralysed.

ppi13:
## What Does This PR Do

This fixes an issue where lockeddown borgs can still interact with doors, on top of making sure stunned, weakend or paralyzed borgs can't do it either.

## Why It's Good For The Game

Per today, a lockeddown borg are not supposed to be able to interact with much outside of their radio, this fixes it

## Changelog
:cl: ppi
fix: Fixes oversight where locked down, stunned, weakened or paralysed borgs could interact with airlocks.
/:cl:

### DIFF
--- a/code/_onclick/cyborg.dm
+++ b/code/_onclick/cyborg.dm
@@ -17,7 +17,7 @@
 
 	if(is_ventcrawling(src)) // To stop drones interacting with anything while ventcrawling
 		return
-	if(stat == DEAD || lockcharge || IsWeakened() || stunned || paralysis)
+	if(stat == DEAD || lockcharge || IsWeakened() || stunned || paralysis || low_power_mode)
 		return
 
 	var/list/modifiers = params2list(params)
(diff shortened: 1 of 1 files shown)


code/_onclick/cyborg.dm
```diff
@@ -17,7 +17,7 @@
 
 	if(is_ventcrawling(src)) // To stop drones interacting with anything while ventcrawling
 		return
-	if(stat == DEAD)
+	if(stat == DEAD || lockcharge || IsWeakened() || stunned || paralysis)
```
S34NW, line 20:
```suggestion
	if(stat == DEAD || lockcharge || IsWeakened() || stunned || paralysis || low_power_mode)
```
Is it also worth adding low power mode to this too? They should only be able to do limited things in this mode.

AffectedArc07: Please make this change too.

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #16034, opened 2021-05-15
EFTPOS code parcel now spawns in the chef's backpack.

### LINKED ISSUE #16031: The chef's EFTPOS scanner access code gets spawned inside the chef
**Issue Description**:
The chef's EFTPOS scanner spawns the 'EFTPOS access code' parcel ***inside*** the chef.

The code would spawn in their backpack, or maybe on the floor.

**What happened instead**:
The code parcel gets lost inside the chef's `contents` forever.

Access code vore is weird, and it makes the scanner practically unusable (more than it is already) since you can't reset a purchase.

Spawn as chef, check your `contents` variable for a `obj/item/smallDelivery`.

Probably forever.

**Extra information**:
This is the issue:
https://github.com/ParadiseSS13/Paradise/blob/e4a1ab04c7a93533f07758a7b5ae0ab851bd70f8/code/modules/economy/EFTPOS.dm#L26-L44
The `loc` in this case is apparently the chef rather than the backpack, so everything just gets spawned inside them.
I looked into spawning it on the turf instead, but that ended up just seeming a bit odd.

<hr>

**Bonus Issue:**
The scanner makes a really loud noise whenever it prints the code (which it does on `Initialize()`), so you get jumpscared every time you spawn as a chef.

Dragonkiller93:
## What Does This PR Do
The chef's EFTPOS  code parcel now spawns in the chef's backpack, instead of inside the chef. Fixes #16031

## Why It's Good For The Game
So the chef gets their EFTPOS code parcel.

## Changelog
:cl: Iwanabeu
fix: Chef's EFTPOS code parcel now spawns in their backpack instead of inside them.
/:cl:

### DIFF
--- a/code/modules/economy/EFTPOS.dm
+++ b/code/modules/economy/EFTPOS.dm
@@ -38,7 +38,7 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(R.loc)
+	var/obj/item/smallDelivery/D = new(get_turf(R))
 	R.forceMove(D)
 	D.wrapped = R
 	D.name = "small parcel - 'EFTPOS access code'"
--- a/code/modules/economy/EFTPOS.dm
+++ b/code/modules/economy/EFTPOS.dm
@@ -41,7 +41,8 @@
 	var/obj/item/smallDelivery/D = new(get_turf(loc))
 	if(istype(loc, /mob/living/carbon/human))
 		var/mob/living/carbon/human/H = loc
-		if(H.back != null) D.forceMove(H.back)
+		if(H.back)
+			D.forceMove(H.back)
 	R.forceMove(D)
 	D.wrapped = R
 	D.name = "small parcel - 'EFTPOS access code'"
--- a/code/modules/economy/EFTPOS.dm
+++ b/code/modules/economy/EFTPOS.dm
@@ -38,7 +38,10 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(get_turf(R))
+	var/obj/item/smallDelivery/D = new(get_turf(src.loc))
+	if(istype(src.loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = src.loc
+		D.forceMove(H.back)
 	R.forceMove(D)
 	D.wrapped = R
 	D.name = "small parcel - 'EFTPOS access code'"
--- a/code/modules/economy/EFTPOS.dm
+++ b/code/modules/economy/EFTPOS.dm
@@ -38,9 +38,9 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(get_turf(src.loc))
-	if(istype(src.loc, /mob/living/carbon/human))
-		var/mob/living/carbon/human/H = src.loc
+	var/obj/item/smallDelivery/D = new(get_turf(loc))
+	if(istype(loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = loc
 		D.forceMove(H.back)
 	R.forceMove(D)
 	D.wrapped = R
--- a/code/modules/economy/EFTPOS.dm
+++ b/code/modules/economy/EFTPOS.dm
@@ -41,7 +41,7 @@
 	var/obj/item/smallDelivery/D = new(get_turf(loc))
 	if(istype(loc, /mob/living/carbon/human))
 		var/mob/living/carbon/human/H = loc
-		D.forceMove(H.back)
+		if(H.back != null) D.forceMove(H.back)
 	R.forceMove(D)
 	D.wrapped = R
 	D.name = "small parcel - 'EFTPOS access code'"


code/modules/economy/EFTPOS.dm
```diff
@@ -38,7 +38,7 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(R.loc)
+	var/obj/item/smallDelivery/D = new(get_turf(R))
```
AffectedArc07, line 41: I would rather this be put into the backpack, since that makes more sense.

S34NW:
> instead of inside the chef.

What

Dragonkiller93:
Yeah it was spawning inside them, so you could see it in the mob's contents.

Dragonkiller93:
Okay I'm pretty confused at this point, the src variable is the EFTPOS, which makes sense. Putting the package into src.loc results in the package inside the person, but I can't use src.loc.search_contents_for function to find the chef's backpack. Not sure what I'm not understanding, any help would be greatly appreciated.

code/modules/economy/EFTPOS.dm
```diff
@@ -38,7 +38,10 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(R.loc)
+	var/obj/item/smallDelivery/D = new(get_turf(src.loc))
+	if(istype(src.loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = src.loc
+		D.forceMove(H.back)
```
AffectedArc07, line 44:
```suggestion
	var/obj/item/smallDelivery/D = new(get_turf(loc))
	if(istype(loc, /mob/living/carbon/human))
		var/mob/living/carbon/human/H = loc
		D.forceMove(H.back)
```

AffectedArc07 (changes requested):

code/modules/economy/EFTPOS.dm
```diff
@@ -38,7 +38,10 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(R.loc)
+	var/obj/item/smallDelivery/D = new(get_turf(loc))
+	if(istype(loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = loc
+		D.forceMove(H.back)
```
farie82, line 44: Will have to check if `H.back` is null here as well. Edge case most likely but it's best to be safe. Otherwise, you can also try put it in his hands

farie82 (changes requested):

code/modules/economy/EFTPOS.dm
Dragonkiller93: The code that creates an eftpos, which calls this, is only found in the backpack_contents list for chef outfit, so it might not be necessary to check.

Carthusia:
Huh, so this is why they never sp awned with the code?

Id get the scanner, but never a code box

Darkmight9:
Well this is an issue I never realized was going on O_o

Dragonkiller93:
So looking at the changes, half of them are coming from being unable to access the repository, what is causing this?

S34NW:
CI is currently broken, @AffectedArc07 will trigger it again once it is no longer broken

code/modules/economy/EFTPOS.dm
AffectedArc07: Put a check in. Admins may spawn an EFTPOS.

Dragonkiller93: Okay, if there is no backpack it'll just spawn on the floor, putting in hands would have the same issue if admin tries to spawn it on the ground, floor is just easier.

farie82: it'll try to put it in the hands of the `loc`/`human`. If that fails then it'll drop it to the floor. That's how the `put_in_hands` proc works

Dragonkiller93: Oh I see, I thought you meant hands in the same way as human/back. Either way should have the intended effect.

Dragonkiller93:
This should be good, the code spawns in the backpack, unless backpack is null in which case it just spawns on the floor below. This works both with standard chefs at roundstart/late joins, and with admin spawned EFTPOS, which will create a parcel on the floor at the spawnpoint.

code/modules/economy/EFTPOS.dm
```diff
@@ -38,7 +38,10 @@
 	R.stamped += /obj/item/stamp
 	R.overlays += stampoverlay
 	R.stamps += "<HR><i>This paper has been stamped by the EFTPOS device.</i>"
-	var/obj/item/smallDelivery/D = new(R.loc)
+	var/obj/item/smallDelivery/D = new(get_turf(loc))
+	if(istype(loc, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = loc
+		if(H.back != null) D.forceMove(H.back)
```
farie82, line 44:
```suggestion
		if(H.back)
			D.forceMove(H.back)
```
DM can do falsy checks. Meaning it can directly if something is null or not (null being seen as FALSE)
We also put a new line after every if. Even though it only has one instruction in it's scope

farie82 (changes requested):

Kyep (commented):
Make it spawn in their backpack, or locker.
Having it spawn on the floor underneath them seems broken.

Dragonkiller93:
It does spawn in their backpack. It will spawn on the floor if an admin spawns a new eftpos.

Dragonkiller93:
I updated the pr text to reflect the changes, sorry about that.

dearmochi (approved):

farie82 (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #16035, opened 2021-05-15
Fixes genetics gene changes not working

farie82:
## What Does This PR Do
Fixes my fuckup from #15996.
Somehow I managed to undo one for loop and I forgot to test the monkey gene (which is the most important of them all)
It works now.

## Why It's Good For The Game
Fixes genetics

## Changelog
:cl:
fix: Fixes genetics not properly activating genes when it should
/:cl:

### DIFF
--- a/code/game/dna/dna2_domutcheck.dm
+++ b/code/game/dna/dna2_domutcheck.dm
@@ -3,7 +3,8 @@
 // M: Mob to mess with
 // flags: See below, bitfield.
 /proc/domutcheck(mob/living/M, flags = 0)
-	for(var/datum/mutation/mutation in GLOB.dna_mutations)
+	for(var/mutation_type in GLOB.dna_mutations)
+		var/datum/mutation/mutation = GLOB.dna_mutations[mutation_type]
 		if(!M || !M.dna)
 			return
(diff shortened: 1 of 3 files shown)


code/game/dna/mutations/monkey.dm
```diff
@@ -9,6 +9,7 @@
 	return ishuman(M)
 
 /datum/mutation/monkey/activate(mob/living/carbon/human/H)
+	..()
```
farie82, line 12: This actually adds the gene to the active gene pool. Same for the deactivate

code/game/dna/dna2_domutcheck.dm
```diff
@@ -3,7 +3,8 @@
 // M: Mob to mess with
 // flags: See below, bitfield.
 /proc/domutcheck(mob/living/M, flags = 0)
-	for(var/datum/mutation/mutation in GLOB.dna_mutations)
+	for(var/mutation_type in GLOB.dna_mutations)
```
farie82, line 6: No clue how I undid this change since I went through all (which weren't that many) instances of `dna_mutations`

AffectedArc07 (approved):

PR #16046, opened 2021-05-17
Tajaran Sleek Veil Flashproofing

Onyxarias:
## What Does This PR Do
Adds flash immunity to the Sleek Veil, bringing it up to spec with the standard Security HUD sunglasses. ~~Also adds species restriction to all of the veils, so that only Tajaran can wear them~~

## Why It's Good For The Game
Tajaran sec officers shouldn't have to nerf themselves in order to flavour their character with lore-specific gear. Choosing the sleek veil over a standard set of HUD sunglasses was always a disadvantage, and shouldn't be when every officer has access to them from roundstart. ~~There is no reason why another race should be able to spawn with veils, given the stated lore on them. Gives Taj a reason to follow their lore regarding veils.~~

## Changelog
:cl:
add: Adds the flash_protect ability to the Sleek Veil
/:cl:

### DIFF
--- a/code/modules/clothing/glasses/hud.dm
+++ b/code/modules/clothing/glasses/hud.dm
@@ -171,7 +171,7 @@
 	desc = "An Ahdominian made veil that allows the user to see while obscuring their eyes. This one has an in-built security HUD."
 	icon_state = "tajblind_sec"
 	item_state = "tajblind_sec"
-	flash_protect = 1
+	flash_protect = FLASH_PROTECTION_FLASH
 	flags_cover = GLASSESCOVERSEYES
 	actions_types = list(/datum/action/item_action/toggle)
 	up = 0
--- a/code/modules/clothing/glasses/hud.dm
+++ b/code/modules/clothing/glasses/hud.dm
@@ -172,7 +172,6 @@
 	icon_state = "tajblind_sec"
 	item_state = "tajblind_sec"
 	flash_protect = 1
-	species_restricted = list("Tajaran")
 	flags_cover = GLASSESCOVERSEYES
 	actions_types = list(/datum/action/item_action/toggle)
 	up = 0
--- a/code/modules/clothing/glasses/hud.dm
+++ b/code/modules/clothing/glasses/hud.dm
@@ -171,6 +171,8 @@
 	desc = "An Ahdominian made veil that allows the user to see while obscuring their eyes. This one has an in-built security HUD."
 	icon_state = "tajblind_sec"
 	item_state = "tajblind_sec"
+	flash_protect = 1
+	species_restricted = list("Tajaran")
 	flags_cover = GLASSESCOVERSEYES
 	actions_types = list(/datum/action/item_action/toggle)
 	up = 0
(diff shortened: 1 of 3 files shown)


kenken244:
There shouldn't be a species restriction. Both because I want to wear it to offend cats and also because of balance. If an anyway takes out a tator, they need to be able to use the sec hud

Aurorablade:
Species restriction is a no go i did attempt that when i ported them orignally and no. Its not something to be done, its fine without the restricts on it really.

Onyxarias:
I'll wait to see what a maintainer has to say about it. If it's an issue, I can change it.

Aurorablade:
Maintainer was the one who told me to change it when i orginally ported it. Trust me stance has not changed.

Djvoid225:
lorewise flashproof veils are more than supported, and species restriction isn't really that big of a difference when it comes down to gameplay (though I can for sure see it messing with some newer antag players, so MAYBE cut that.) otherwise I see this is an absolute win. 

More lore and rp individuality enforced by gameplay. Thumbs up from me. (and the... two people that play Tajaran.)

Wojz:
SecHUD that only tajs can use... that's not gonna fly. But I don't think the flashproof is a big deal tbh. Not like sunglasses are hard to comeby for anyone who selects those.

ZanderDoggo:
I like the idea of it, and I think it would probably be better as "loadout" locked for other species than Taj. If anyone can still wear it but just can't roundstart with it, then I think it would be fine. It allows for antags and other players to still be able to steal a fully functioning sec hud without other species spawning with veils which doesn't really make much sense overall.

S34NW:
I think adding the flash protection makes sense, but species-locking it? No thank you. Not to mention that non-tajarans can still have it as part of their loadout, but then be unable to wear it once in-game.

Onyxarias:
Removed the species restriction. Until I can figure out how to make them only SPAWN on Taj, instead of being only wearable by them, they will remain available to all.;

SabreML (changes requested):
image
Not *entirely* sure how you managed that, but you should be able to solve it by just deleting the `Paradise` folder in the PR branch and committing again.

(This one:)
image

code/modules/clothing/glasses/hud.dm
```diff
@@ -171,6 +171,7 @@
 	desc = "An Ahdominian made veil that allows the user to see while obscuring their eyes. This one has an in-built security HUD."
 	icon_state = "tajblind_sec"
 	item_state = "tajblind_sec"
+	flash_protect = 1
```
Fox-McCloud, line 174: Use the define please.

Fox-McCloud:
Please fix the unnecessary file change so this can be merged.

Onyxarias:
Oh stars, uh. I'm going to be honest, totally forgot about this.
I'm sorry for the messup, bear with me while I relearn things again and get it fixed.
Sorry ;^^

dearmochi (changes requested):
The branch seems to include the repo as a sub module, which is probably not what you want to do

variableundefined (changes requested):
This PR is approved and ready to go, except for the part where the repo is a submodule.

SabreML (approved):

code/modules/clothing/glasses/hud.dm
Onyxarias: Done and done, thank you. I used the standard sechud as a reference, which also doesn't use the define

PR #8499, opened 2017-12-31
A menu of emagged drones nerfs and buffs

Tayyyyyyy:
:cl: Tayyyyyyy
tweak: Due to increasing amounts of spaghetti code in the cryptographic sequencer, emagging maintenance drones cripples their table pathfinding and removes their ability to pathfind under people's legs. Also, Nanotrasen has patched its drones to remove the hidden diamond drill module and rewritten its drone control code so that drones that don't call home and verify their programming after a certain amount of time are destroyed automatically.
/:cl:

Wow, okay, I guess people really like emagged drives, which I still think are fundamentally wrong, but just in case I didn't make it clear enough multiple times, *these nerfs are not all intended to be merged together.* I just stuck a bunch of fixes of common emagged drone conplaints into one PR. It's my opinion that at least the first three would be great, but my opinion isn't what determines what will ultimately be nerfed.

Assuming an emag has already been purchased, the drone is something that can be gotten for 0 additional TC, the maintenance drone should NOT be far more powerful than a holoparasite, which costs 12 TC.

The idea is that the emagged drone should be an all access, mobile carded AI that can bolt/shock doors for you, as well as do construction and fight for you, but not be too powerful since there is no limit on how many emagged drones you have, and they don't have to stick close to you.

Also a minor fix to the zeroth law that's given to drones to avoid pronoun laws lawyering.

### ~~Can be blown up remotely~~
~~This is a counterplay to an emagged borg, makes sense for the emagged drone too. But this isn't enough because a powergaming drone will just deconstruct this first thing, and the engi outpost isn't a highly visible area like the borg console.~~

### ~~No more ventcrawling~~
~~The emagged drone already has all access and has AI level control over doors/station systems, no need to make it impossible to hit too. Oh, and they can unweld vents, unlike literally everything else that can ventcrawl.~~

### No more diamond drill
It already has a crowbar, fire extinguisher, and a welding tool, does it really need this?

### Dense
Cannot go onto the same tile as a person anymore, which makes it really hard to hit without hitting yourself

### No more hiding in tables
Another common complaint with emagged drones.

### Automatic self destruct
Self destruct 5 minutes after being emagged. Drone players are presumably playing for a sort of "zen mode". This will let drone players get back to their non-intrusive role after a short period of time.

### ~~Grants galactic common~~
[...]

PR #16047, opened 2021-05-17
Allows Maintenance Drones to see wire names

SabreML:
## What Does This PR Do

Changes the Abductor multitool wire interface (see image) into an item trait, and adds it to Abductor wirecutters and Maintenance drone multitools and wirecutters.
It being a trait also allows admins to add the wire vision to more or less anything. (As an example I used in testing, a dead fish.)

Also fixes maintenance drones getting a toy sword named 'Placeholder emag item' when they get emagged, which has been a bug for over three years now. #8499

## Why It's Good For The Game

Maintenance Drones are intrinsically linked to the station, so it really doesn't make much sense (to me anyway) that they need to use trial and error to deconstruct an airlock or otherwise hack a machine. Since there's a whole fancy wire name system already for abductors, that seemed like a pretty good fit for Drones too.

(Also adding it to the wirecutters is just so that you don't have to keep switching back to a multitool if you forget.)

## Images of changes

image

## Changelog
:cl:
add: Changed the Abductor multitool wire interface (Visible wire names) into an item trait, and added it to Abductor wirecutters and Maintenance Drone multitools and wirecutters.
fix: Finally removed the 'Placeholder emag item' from emagged Maintenance Drones.
/:cl:


hal9000PR:
On TG, holding the station blueprints in your hand let's you see door wires. It would make sense and give that theft objective a neat buff so that it is actually useful like other theft objectives.

code/game/objects/items.dm
```diff
@@ -100,6 +100,8 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 	var/tool_enabled = TRUE //If we can turn on or off, are we currently active? Mostly for welders and this will normally be TRUE
 	var/tool_volume = 50 //How loud are we when we use our tool?
 	var/toolspeed = 1 // If this item is a tool, the speed multiplier
+	/// Show what a machine/door wire does when held.
+	var/show_wire_information = FALSE
```
AffectedArc07, line 104: Not sure this should be put on base level `/obj/item`

SabreML: Yeah I wasn't sure about that bit either. I can't see another way of doing it though, since there's not an `/obj/item/tool` parent that it could be put on. (There probably should be though)
_Maybe_ a component or something would work better, but I've never actually looked into how those work so I don't know.

SteelSlayer: Suppose you could make it proc-based instead. For example:
```dm
/obj/item/proc/shows_wire_info()
    return FALSE

/obj/item/multitool/cyborg/drone/shows_wire_info()
    return TRUE

/obj/item/wirecutters/cyborg/drone/shows_wire_info()
    return TRUE

// etc.
```
If that's not wanted, then yeah could go component (or element) or maybe even pure signal based.

SabreML: Not sure which of these would be preferred, but I'll look into them. All of the actual wire code is there already so it should be pretty easy.

SabreML:
> On TG, holding the station blueprints in your hand let's you see door wires. It would make sense and give that theft objective a neat buff so that it is actually useful like other theft objectives.

I kinda locked myself in here with the PR and branch names so this might be out of scope, but it'd be pretty easy to do in another PR.

code/game/objects/items.dm
SabreML: Might have forgotten about this until just now, whoops.
So what *would* be the best way to do this? Changing it to use procs instead might be better, but it also has the same issue of being something on `/obj/item`, and only used for a couple of subtypes.

SteelSlayer: Besides the things I listed before, you could make a trait `SHOWS_WIRE_INFO` or such, and attach it to the items you want. Then check for the trait in the wires datum. This is probably a better way than the others.

code/_globalvars/traits.dm
```diff
@@ -57,7 +57,12 @@ GLOBAL_LIST_INIT(traits_by_type, list(
 		"TRAIT_NOGERMS" = TRAIT_NOGERMS,
 		"TRAIT_NODECAY" = TRAIT_NODECAY,
 		"TRAIT_NOEXAMINE" = TRAIT_NOEXAMINE,
-		"TRAIT_NOPAIN" = TRAIT_NOPAIN)))
+		"TRAIT_NOPAIN" = TRAIT_NOPAIN
+		),
+		/obj/item = list(
+			"TRAIT_SHOW_WIRE_INFO" = TRAIT_SHOW_WIRE_INFO
+		)
```
SteelSlayer, line 64: Nitpick: to keep it on the same indention as the `/mob = list(` above
```suggestion
	),
	/obj/item = list(
		"TRAIT_SHOW_WIRE_INFO" = TRAIT_SHOW_WIRE_INFO
	)
```

SabreML: Oh I thought I'd done that. Must've got confused by the long list.

The-Driker:
Can we add this to engineering cyborgs as well? They are also machines linked directly to the station electronics.

SabreML:
> Can we add this to engineering cyborgs as well? They are also machines linked directly to the station electronics.

I could add that pretty easily, but I was worried about possible balance issues that might cause.

Engiborgs could just go up to a door at roundstart and list off what every wire does, which somewhat removes any challenge from hacking things.
Of course, borgs can do that already since they can't be shocked by an airlock, but it still involves some trial and error. (And even despite that, I've never seen a borg give the wires over common.)

Maintenance Drones handily avoid this issue entirely by not actually being able to speak, but if there's enough demand for borgs to get it too then I wouldn't mind adding it.

SteelSlayer (approved):

Kyep (approved):

AffectedArc07 (approved):

KozackWolf:
> 
> 
> Can we add this to engineering cyborgs as well? They are also machines linked directly to the station electronics.

Abductor levels to 4 I think gives an upgrade that can give you wires for engiborgs. or alien tools for medborgs.

PR #16049, opened 2021-05-18
Removes the early classified announcement from the blob gamemode

Qwertytoforty:
##This does _not_ remove the level 5 bio hazard announcement.

## What Does This PR Do

Removes the classified announcement sent before the level 5 biohazard warning on blob rounds, and the paper that goes with it.
Also removes the declared variable, which does nothing.

## Why It's Good For The Game

Roundstart blob often has it hard, but often it is spotted quite fast. When it is not spotted fast, it has a larger issue.

image

This message will play. Without fail, even before anyone reads the paper, some crew will scream out blob, even if they do not see the paper and what it says. or the blob. This is silly, that crew instantly presume this message is a blob announcement. 
And of course, barring meta, this announcement can come really fast on a blob round.
Thus, this PR removes this announcement.

## Changelog
:cl:
tweak: Removes the early biohazard warning on blob rounds, but not the level 5 biohazard warning.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blob.dm
+++ b/code/game/gamemodes/blob/blob.dm
@@ -15,7 +15,6 @@ GLOBAL_LIST_EMPTY(blob_nodes)
 	recommended_enemies = 1
 	restricted_jobs = list("Cyborg", "AI")
 
-	var/declared = 0
 	var/burst = 0
 
 	var/cores_to_spawn = 1
@@ -177,12 +176,8 @@ GLOBAL_LIST_EMPTY(blob_nodes)
 
 		burst_blobs()
 
-		// Stage 0
-		sleep(wait_time)
-		stage(0)
-
 		// Stage 1
-		sleep(wait_time)
+		sleep((wait_time * 2))
 		stage(1)
 
 		// Stage 2
@@ -193,13 +188,10 @@ GLOBAL_LIST_EMPTY(blob_nodes)
 
 /datum/game_mode/blob/proc/stage(stage)
 	switch(stage)
-		if(0)
-			send_intercept(1)
-			declared = 1
 		if(1)
 			GLOB.event_announcement.Announce("Confirmed outbreak of level 5 biohazard aboard [station_name()]. All personnel must contain the outbreak.", "Biohazard Alert", 'sound/AI/outbreak5.ogg')
 		if(2)
-			send_intercept(2)
+			send_intercept(1)
 
 /datum/game_mode/proc/update_blob_icons_added(datum/mind/mob_mind)
 	var/datum/atom_hud/antag/antaghud = GLOB.huds[ANTAG_HUD_BLOB]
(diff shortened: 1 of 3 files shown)


AffectedArc07 (changes requested):
image
Remove this too while you're at it

AffectedArc07 (approved):

DogsParadise:
I was trying to counter act this problem as well, by creating a similar biohazard message with my mosquito pr

code/game/gamemodes/blob/blob.dm
```diff
@@ -177,12 +176,8 @@ GLOBAL_LIST_EMPTY(blob_nodes)
...
 		burst_blobs()
 
-		// Stage 0
-		sleep(wait_time)
-		stage(0)
-
 		// Stage 1
-		sleep(wait_time)
+		sleep((wait_time * 2))
 		stage(1)
```
S34NW, line 181:
```suggestion
		addtimer(CALLBACK(src, .proc/stage, 1), (wait_time * 2))
```

S34NW (commented):
Could you also move the other `sleep` calls in this proc to `addtimer` if possible?

code/game/gamemodes/blob/blob.dm
Qwertytoforty: sure

code/game/gamemodes/blob/blob.dm
```diff
@@ -75,8 +74,7 @@ GLOBAL_LIST_EMPTY(blob_nodes)
 	log_game("[key_name(blob)] has been selected as a Blob")
 	greet_blob(blobmind)
 	to_chat(blob, "<span class='userdanger'>You feel very tired and bloated!  You don't have long before you burst!</span>")
-	spawn(600)
-		burst_blob(blobmind)
+	addtimer(CALLBACK(src, .proc/burst_blob, blobmind), 600)
```
S34NW, line 77:
```suggestion
	addtimer(CALLBACK(src, .proc/burst_blob, blobmind), 60 SECONDS)
```

code/game/gamemodes/blob/blob.dm
```diff
@@ -173,33 +170,22 @@ GLOBAL_LIST_EMPTY(blob_nodes)
...
 
 		// Stage 1
-		sleep(wait_time)
-		stage(1)
+		addtimer(CALLBACK(src, .proc/stage, 1), (wait_time * 2))
 
 		// Stage 2
-		sleep(30000)
-		stage(2)
+		addtimer(CALLBACK(src, .proc/stage, 2), 30000)
```
S34NW, line 179:
```suggestion
		addtimer(CALLBACK(src, .proc/stage, 2), 50 MINUTES)
```

code/game/gamemodes/blob/blob.dm
```diff
@@ -173,33 +170,22 @@ GLOBAL_LIST_EMPTY(blob_nodes)
...
 		show_message("<span class='userdanger'>You feel like you are about to burst.</span>")
 
-		sleep(wait_time / 2)
-
-		burst_blobs()
-
-		// Stage 0
-		sleep(wait_time)
-		stage(0)
+		addtimer(CALLBACK(src, .proc/burst_blobs), (wait_time * 2))
```
dearmochi, line 173: Shouldn't it be `wait_time / 2`?

Qwertytoforty: It should be, yeah.

Kyep (approved):

AffectedArc07 (approved):

PR #16061, opened 2021-05-22
Adds rebreathers

michaelyesok:
My first PR, Made with help E-MonaRhg and SabreML. Object spirit from bay, Other spirits made by me.
## What Does This PR Do

This PR adds rebreathers, A mask created by using wirecutters on a gas mask offering less protection then average gas mask but still protection against chemical mixing while not hiding your face and/or flavor text.

## Why It's Good For The Game

A fashion item for those who still would like protection against chemicals without hiding their flavor text or face.
## Images of changes

image
image
image
image

## Changelog
:cl:
add: Added rebreathers to the game. Can be constructed by using wirecutters on a gas mask
/:cl:


code/modules/clothing/masks/gasmask.dm
```diff
@@ -20,6 +20,24 @@
...
+		return
+
+	. = TRUE
+	to_chat(user, "<span class='notice'>You start cutting [src].</span>")
+	if(!I.use_tool(user, user, 5 SECONDS, volume = I.tool_volume))
+		to_chat(user, "<span class='warning'>You decide not to cut [src].</span>")
+		return
+
+	to_chat(user, "<span class='notice'>You cut the top half off of [src], making a rebreather!</span>")
+	name = "rebreather"  // Change the name
```
S34NW, line 36:
```suggestion
	name = "rebreather"
```

code/modules/clothing/masks/gasmask.dm
```diff
@@ -20,6 +20,24 @@
...
+	desc = "Placeholder" // Change the description
```
S34NW, line 37:
```suggestion
	desc = "Placeholder"
```

Also please fill the desc out :)

code/modules/clothing/masks/gasmask.dm
```diff
@@ -20,6 +20,24 @@
...
+	icon_state = "rebreather" // Change the sprite
```
S34NW, line 38:
```suggestion
	icon_state = "rebreather"
```

code/modules/clothing/masks/gasmask.dm
```diff
@@ -20,6 +20,24 @@
 		"Plasmaman" = 'icons/mob/species/plasmaman/mask.dmi'
 		)
 
+/obj/item/clothing/mask/gas/wirecutter_act(mob/living/user, obj/item/I)
+	if(type != /obj/item/clothing/mask/gas)
+		return
```
AffectedArc07, line 25:
```suggestion
	if(!istype(src, /obj/item/clothing/mask/gas))
		return
```

code/modules/clothing/masks/gasmask.dm
```diff
@@ -20,6 +20,24 @@
...
+	if(icon_state == "rebreather")
+		return
```
AffectedArc07, line 27: Feel like you should track whether its a rebreather or not using something other than icon state.

code/modules/clothing/masks/gasmask.dm
AffectedArc07, line 37: You forgot this

AffectedArc07 (changes requested):

S34NW (commented):
I'd suggest you leave out comments for common things such as name, desc, etc. as these are fairly redundant comments.
Also, there is a missing description.

AffectedArc07:
Ok I'll bite. Whats wrong with a normal mask.
image
image

code/modules/clothing/masks/gasmask.dm
SabreML: `istype()` would also allow subtypes through as far as I know, which includes stuff like welding gas masks, syndicate masks, clown masks, etc.

AffectedArc07: Valid, ignore this suggestion

code/modules/clothing/masks/gasmask.dm
```diff
@@ -33,11 +33,11 @@
 		return
 
 	to_chat(user, "<span class='notice'>You cut the top half off of [src], making a rebreather!</span>")
-	name = "rebreather"  // Change the name
-	desc = "Placeholder" // Change the description
-	icon_state = "rebreather" // Change the sprite
-	flags_inv = null // Override the `HIDEEARS|HIDEEYES|HIDEFACE` from the standard gas mask
-	flags_cover = MASKCOVERSMOUTH // Override the `MASKCOVERSEYES` from the standard gas mask
```
AffectedArc07: These comments probably could have stayed in

DogsParadise:
Sprite doesnt fit imo.

A rebreather looks real different.
And just cutting a gasmask in half.. isn't great.
You can do lots more, creating an actual cool sprite

FlattestGuitar:
First of all, that's not what a rebreather is. That word has a very specific meaning.

Second of all, the object sprite is a bit ugly.

Third of all, this seems just silly and pointless. Don't add bloat for the sake of it, smh.

code/modules/clothing/masks/gasmask.dm
```diff
@@ -19,7 +19,27 @@
 		"Grey" = 'icons/mob/species/grey/mask.dmi',
 		"Plasmaman" = 'icons/mob/species/plasmaman/mask.dmi'
 		)
+var/gasmask_cut = FALSE
```
SabreML, line 22: Needs to be indented by one tab so that it's part of `/obj/item/clothing/mask/gas`.
```suggestion
	var/gasmask_cut = FALSE
```

LordShekelstein:
Aren't rebreathers for like... diving? Also this sprite seems incredibly lazy, especially if the intended purpose is fashion.

LiberatedWaveMan:
Seems like an interesting concept, but I feel like it could use some work. The sprite seems a bit clunky to me as a "fashion" item, and the use of the term "rebreather" is a bit misleading. Other than that it's good, but some changes to that would be nice

michaelyesok:
So yes, I will be closing this for now unil i can get better spirits for this.

PR #16062, opened 2021-05-22
Stethoscope now checks whether the target is dead.

### LINKED ISSUE #16057: Stethoscopes say that the dead breath and have a pulse
**Issue Description**:
Clicking on a dead body with a stethoscope reports that they have a pulse and breathing when they really shouldn't.

I expected the stethoscope to say that there's no breathing or pulse.

**What happened instead**:
The stethoscope said that the dead body was breathing.

Misinformation is bad.

1. Acquire a dead body.
2. Take your stethoscope and use it on the body.
3. Get confused.

I'm not sure.

**Extra information**:
The stethoscope could use some general tweaks to be more useful, like maybe detecting cardiac arrest/failure and ruptured lungs instead of only heart and lung damage.

Dragonkiller93:
## What Does This PR Do
The stethoscope will no longer read a healthy heart and lungs if the target is dead. It will now tell you that you "don't hear anything." This also works with capulettium (plus), you will not hear anything. Fixes #16057. 

## Why It's Good For The Game
The dead body is not in fact, breathing, nor is its heart beating.

## Changelog
:cl: Iwanabeu
fix: Stethoscope no longer reads dead bodies as having beating hearts and breathing.
/:cl:


Dragonkiller93:
The fix is a little weird, I originally was trying to use the is_dead proc, but that actually works through capulettium, and I figured that probably isn't a good thing.

code/modules/clothing/under/accessories/accessory.dm
```diff
@@ -141,7 +141,7 @@
 			user.visible_message("[user] places \the [src] against [M]'s chest and listens attentively.", "You place \the [src] against [M]'s chest...")
 		var/obj/item/organ/internal/H = M.get_int_organ(/obj/item/organ/internal/heart)
 		var/obj/item/organ/internal/L = M.get_int_organ(/obj/item/organ/internal/lungs)
-		if((H && M.pulse) || (L && !HAS_TRAIT(M, TRAIT_NOBREATH)))
+		if((M.pulse) && ((H || (L && !HAS_TRAIT(M, TRAIT_NOBREATH)))))
```
SabreML, line 144: Don't need that many brackets for this:
```suggestion
		if(M.pulse && (H || (L && !HAS_TRAIT(M, TRAIT_NOBREATH))))
```

dearmochi (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #16068, opened 2021-05-24
Poly can no longer grab inside disposals.

### LINKED ISSUE #16053: Poly grabs null item while in disposals
**Issue Description**:
Poly can use the "Steal from ground" button while in disposal bins or pipes and if there is item inside the bin or the pipes a "unusual" message appears

Nothing should happen since poly doesn't actually pick up the item.

**What happened instead**:
Message about poly grabbing nothing appears.

It's nothing horrible but surely not intended.

1. Put any item and poly into the disposal bin
2. Use "steal from ground" button as poly when inside bin or moving in a pipe

First spotted 20/05/2021

**Extra information**:
image

Dragonkiller93:
## What Does This PR Do
Poly will no longer grab null items while inside disposals or being flushed. Instead a message appears telling you that you are inside a disposal chute. Fixes #16053.

## Why It's Good For The Game
So players don't have to see bugged messages, and will instead receive an accurate description of why they can't perform an action.

## Changelog
:cl: Iwanabeu
fix: Poly can't grab null items in a disposal chute.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/parrot.dm
+++ b/code/modules/mob/living/simple_animal/parrot.dm
@@ -562,8 +562,8 @@
 	if(held_item)
 		to_chat(src, "<span class='warning'>You are already holding [held_item]</span>")
 		return 1
-	if(istype(loc,/obj/machinery/disposal) || istype(loc,/obj/structure/disposalholder))
-		to_chat(src,"<span class='warning'>You are inside a disposal chute!</span>")
+	if(istype(loc, /obj/machinery/disposal) || istype(loc, /obj/structure/disposalholder))
+		to_chat(src, "<span class='warning'>You are inside a disposal chute!</span>")
 		return 1
 	for(var/obj/item/I in view(1, src))
 		//Make sure we're not already holding it and it's small enough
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/parrot.dm
```diff
@@ -562,7 +562,9 @@
 	if(held_item)
 		to_chat(src, "<span class='warning'>You are already holding [held_item]</span>")
 		return 1
-
+	if(istype(loc,/obj/machinery/disposal) || istype(loc,/obj/structure/disposalholder))
+		to_chat(src,"<span class='warning'>You are inside a disposal chute!</span>")
```
SabreML, line 566: Spacing after the commas
```suggestion
	if(istype(loc, /obj/machinery/disposal) || istype(loc, /obj/structure/disposalholder))
		to_chat(src, "<span class='warning'>You are inside a disposal chute!</span>")
```

meow20:
Yey! Now where is my github contributor role hmmm

AffectedArc07 (approved):

PR #16070, opened 2021-05-24
Cleaner no longer resets Light Tiles.

### LINKED ISSUE #16063: Space cleaner resets light tiles
**Issue Description**:
Space cleaner resets light tiles

Nothing

**What happened instead**:
The light tile went white due to having its `color` var reset.

It doesnt make sense.

Colour a tile then spray it with cleaner

Since I made the tiles RGB

**Extra information**:
This also applies to cables, and we need a system that ignores these atoms when you hit them with cleaner.

Dragonkiller93:
## What Does This PR Do
Space cleaner no longer resets light tiles. This is a bit of a janky fix, because things like colorful reagent use the same method of changing color, and space cleaner does and still will reset their color. Fixes #16063 

## Why It's Good For The Game
Light tiles are changed with a multitool, spraying them with space cleaner shouldn't reset them.

## Changelog
:cl: Iwanabeu
tweak: Space cleaner no longer resets light tiles.
/:cl:


code/modules/reagents/chemistry/reagents/water.dm
```diff
@@ -75,7 +75,8 @@
 				floor_only = FALSE
 			else
 				qdel(C)
-		T.color = initial(T.color)
+		if(!istype(T,/turf/simulated/floor/light))
```
AffectedArc07, line 78:
```suggestion
		if(!istype(T, /turf/simulated/floor/light))
```
This also feels kinda snowflakey but dont know another way around that

Dragonkiller93: The only way around it would be to change the way light tiles work completely, so that they don't use the T.color variable. Or change colorful reagent and any other color changers so that they don't use T.color.

code/game/turfs/simulated/floor/light_floor.dm
```diff
@@ -80,3 +80,16 @@
...
+
+/turf/simulated/floor/light/clean(floor_only)
+	for(var/obj/effect/decal/cleanable/C in src)
+		var/obj/effect/decal/cleanable/blood/B = C
+		if(istype(B) && B.off_floor)
+			floor_only = FALSE
+		else
+			qdel(C)
+	if(floor_only)
+		clean_blood()
```
AffectedArc07, line 92: I would honestly rather a 
```dm
var/colour_cache = color
..()
color = colour_cache
```
here, as that would cut down on the duped code, and make maintaining this easier.

Dragonkiller93: Got it, also the fact that you used both spellings of color at the same time is quite something.

AffectedArc07: Mix of BYOND builtins and my native spelling

code/modules/reagents/chemistry/reagents/water.dm
```diff
@@ -67,20 +67,7 @@
...
-				floor_only = FALSE
-			else
-				qdel(C)
-		T.color = initial(T.color)
-		if(floor_only)
-			T.clean_blood()
-
-		for(var/mob/living/simple_animal/slime/M in T)
-			M.adjustToxLoss(rand(5, 10))
+	T.clean(volume >=1)
```
AffectedArc07, line 70:
```suggestion
	T.clean(volume >= 1)
```
Last nitpick

AffectedArc07 (approved):

PR #16073, opened 2021-05-24
Wires no longer gets destroyed by tesla

ppi13:
## What Does This PR Do

Stops tesla zaps from destroying wires

## Why It's Good For The Game

Wires won't be broken, and the tesla won't get loose as often as it has been.

## Changelog
:cl: ppi
tweak: The tesla will no longer destroy wires
/:cl:

### DIFF
--- a/code/modules/power/tesla/energy_ball.dm
+++ b/code/modules/power/tesla/energy_ball.dm
@@ -216,6 +216,7 @@
 										/obj/machinery/gateway,
 										/obj/structure/lattice,
 										/obj/structure/grille,
+										/obj/structure/cable,
 										/obj/machinery/the_singularitygen/tesla,
 										/obj/machinery/constructable_frame/machine_frame))
 


ManiacalSymphony:
But but, if the tesla doesn't get loose, what else am I gonna complain about

ZanderDoggo:
I would like this, I personally enjoy making 1x1 teslas and the update done to it has made trying to do so extremely boring and a pain to actually make teslas anymore. I think that doing this will make teslas a viable option for engines again. P.S. Teslas are fun and removing the functionality to make complicated setups is boring

AffectedArc07 (approved):

meow20:
I thought yesterday the tesla destroyed wires because the grounding rods were too far, is that not the case?

ppi13:
kinda, but wires are not supposed to be destroyed by tesla, it's what caused nearly every single tesla release in the past couple of months.

PR #16075, opened 2021-05-24
Mechs no longer spit out invisible air tanks.

### LINKED ISSUE #15999: Invisible air tanks that are indestructible from crowbaring broken down mechs
**Issue Description**:

Using a crowbar on a broken mech ejects an air canister that you can push, but not see via right click, and is nearly indestructible, handling dozen of pulse shots.

For a visible air tank i could easily break

**What happened instead**:

I got an invisible and invincible air tank that blocks of a tile

Do that around the captains safe is a 3x3 area, can anything in the safe can't be accessed, cause you got indestructible dense objects in the way.

1. Get a mech.
2. Break mech.
3. Crowbar mech untill you get air tank
4. Right click the tile so see the tank is not there
5. move away from the tile and move back to check it's density still checks
6. blast it with any gun, and it survives.

**Extra information**:

Dragonkiller93:
## What Does This PR Do
Fixes mech wreckage so it properly outputs its internal air tank, instead of an invisible one. Fixes #15999 
Also changes mechs to properly null their internal air tank instead of nulling their power cell twice.

## Why It's Good For The Game
So players don't have to deal with invisible air tanks.

## Changelog
:cl: Iwanabeu
fix: Tank obtained by crowbarring mech wreckage is now visible.
/:cl:

### DIFF
--- a/code/game/mecha/mecha_wreckage.dm
+++ b/code/game/mecha/mecha_wreckage.dm
@@ -56,7 +56,6 @@
 	if(crowbar_salvage.len)
 		var/obj/S = pick(crowbar_salvage)
 		S.forceMove(user.drop_location())
-		S.invisibility = 0
 		user.visible_message("<span class='notice'>[user] pries [S] from [src].</span>", "<span class='notice'>You pry [S] from [src].</span>")
 		crowbar_salvage -= S
 		return
--- a/code/game/mecha/mecha_wreckage.dm
+++ b/code/game/mecha/mecha_wreckage.dm
@@ -60,7 +60,7 @@
 		user.visible_message("<span class='notice'>[user] pries [S] from [src].</span>", "<span class='notice'>You pry [S] from [src].</span>")
 		crowbar_salvage -= S
 		return
-	to_chat(user, "<span class='notice'>You don't see anything that can be cut with [I]!</span>")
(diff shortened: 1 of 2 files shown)


code/game/mecha/mecha_wreckage.dm
```diff
@@ -56,6 +56,7 @@
 	if(crowbar_salvage.len)
 		var/obj/S = pick(crowbar_salvage)
 		S.forceMove(user.drop_location())
+		S.invisibility=0
```
AffectedArc07, line 59:
```suggestion
		S.invisibility = 0
```

S34NW (changes requested):
I'd also tidy up line 63 on `code/game/mecha/mecha_wreckage.dm` that denotes there is nothing to crowbar out. Apart from that and the AA comment, this looks fine.
```
to_chat(user, "<span class='notice'>You don't see anything that can be cut with [I]!</span>")
```
to be
```
to_chat(user, "<span class='notice'>You don't see anything that can be pried with [I]!</span>")
```

dearmochi:
I'm curious as to why the canister's `invisibility` is being set to `100` in the first place. In fact, after testing on a local server, it seems to delete itself from the salvage after a while.

image

farie82 (approved):

dearmochi (approved):

AffectedArc07 (approved):

PR #16082, opened 2021-05-25
GC ref finding improvements

AffectedArc07:
## What Does This PR Do
Partial port of:
- https://github.com/tgstation/tgstation/pull/59118
- https://github.com/tgstation/tgstation/pull/58972
- https://github.com/tgstation/tgstation/pull/57728

This PR does the following:
- Improves reference finding
- Adds a GC debug log so stuff can easily be extracted after a round
- Makes GC ref tracking able to be enabled without the TESTING define (This thing lets you edit your own permissions. No.)

Example of GC debug log:
image

## Why It's Good For The Game
The GC is *very* important to the server not being laggy, as hard deletes will add up over time and make the server unbearably slow with tons of things in the QC queue, with the biggest offender being xenobio slimes at the moment
image
This is terrifying.

## Changelog
N/A

### DIFF
--- a/code/__HELPERS/_logging.dm
+++ b/code/__HELPERS/_logging.dm
@@ -143,6 +143,14 @@ GLOBAL_PROTECT(log_end)
 /proc/log_tgui(text)
 	rustg_log_write(GLOB.tgui_log, "[text][GLOB.log_end]")
 
+#ifdef REFERENCE_TRACKING
+/proc/log_gc(text)
+	rustg_log_write(GLOB.gc_log, "[text][GLOB.log_end]")
+	for(var/client/C in GLOB.admins)
+		if(check_rights(R_DEBUG, 0, C.mob) && (C.prefs.toggles & PREFTOGGLE_CHAT_DEBUGLOGS))
+			to_chat(C, "GC DEBUG: [text]")
+#endif
+
 /proc/log_sql(text)
 	rustg_log_write(GLOB.sql_log, "[text][GLOB.log_end]")
 	SEND_TEXT(world.log, text) // Redirect it to DD too
--- a/code/__HELPERS/_logging.dm
+++ b/code/__HELPERS/_logging.dm
@@ -147,7 +147,7 @@ GLOBAL_PROTECT(log_end)
 /proc/log_gc(text)
 	rustg_log_write(GLOB.gc_log, "[text][GLOB.log_end]")
 	for(var/client/C in GLOB.admins)
-		if(check_rights(R_DEBUG, 0, C.mob) && (C.prefs.toggles & PREFTOGGLE_CHAT_DEBUGLOGS))
+		if(check_rights(R_DEBUG, FALSE, C.mob) && (C.prefs.toggles & PREFTOGGLE_CHAT_DEBUGLOGS))
 			to_chat(C, "GC DEBUG: [text]")
 #endif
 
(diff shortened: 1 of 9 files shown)


code/modules/mob/new_player/new_player.dm
```diff
@@ -19,6 +19,11 @@
 	GLOB.mob_list += src
 	return INITIALIZE_HINT_NORMAL
 
+/mob/new_player/Destroy()
+	GLOB.player_list -= src
```
farie82, line 23: Shouldn't this happen in `mob/Logout()`?

AffectedArc07: Im not sure that gets called on mob transform, but will check

farie82: it *should* happen each time a client leaves the mob and goes to another mob. This is why it's only ever removed in `mob/Logout()`. It would however be... something if this doesn't always happen

AffectedArc07: Despite what the ref says, `/mob/Logout()` is never called unless a client is DCing. Swapping from mob-to-mob doesnt call any of the things in here, including but not limited to:
- TGUI window closing
- Generating logout logs
- Admin logout notifications
- Machine unsetting

code/__HELPERS/_logging.dm
```diff
@@ -143,6 +143,14 @@ GLOBAL_PROTECT(log_end)
 /proc/log_tgui(text)
 	rustg_log_write(GLOB.tgui_log, "[text][GLOB.log_end]")
 
+#ifdef REFERENCE_TRACKING
+/proc/log_gc(text)
+	rustg_log_write(GLOB.gc_log, "[text][GLOB.log_end]")
+	for(var/client/C in GLOB.admins)
+		if(check_rights(R_DEBUG, 0, C.mob) && (C.prefs.toggles & PREFTOGGLE_CHAT_DEBUGLOGS))
```
SabreML, line 150:
```suggestion
		if(check_rights(R_DEBUG, FALSE, C.mob) && (C.prefs.toggles & PREFTOGGLE_CHAT_DEBUGLOGS))
```

code/controllers/subsystem/garbage.dm
```diff
@@ -376,58 +401,86 @@ SUBSYSTEM_DEF(garbage)
...
+		datum_container.last_find_references = search_time
+		var/list/vars_list = datum_container.vars
 
-		for(var/varname in L)
-			if(varname == "vars")
+		for(var/varname in vars_list)
+			#ifndef FIND_REF_NO_CHECK_TICK
+			CHECK_TICK
+			#endif
+			if (varname == "vars" || varname == "vis_locs") //Fun fact, vis_locs don't count for references
```
SabreML, line 447:
```suggestion
			if(varname == "vars" || varname == "vis_locs") //Fun fact, vis_locs don't count for references
```

code/controllers/subsystem/garbage.dm
```diff
@@ -376,58 +401,86 @@ SUBSYSTEM_DEF(garbage)
...
 
-	if(istype(X, /datum))
-		var/datum/D = X
-		if(D.last_find_references == last_find_references)
+	//Check each time you go down a layer. This makes it a bit slow, but it won't effect the rest of the game at all
+	#ifndef FIND_REF_NO_CHECK_TICK
+	CHECK_TICK
+	#endif
+
+	if(istype(potential_container, /datum))
```
SabreML, line 435: Surprised we don't have an `isdatum()` define yet.

dearmochi (approved):

Fox-McCloud (approved):

farie82 (approved):

PR #16084, opened 2021-05-25
Small Species Organ Refactor

SabreML:
## What Does This PR Do

Refactors the `has_organ` and `has_limbs` variables on `/datum/species` to use consistent defines, rather than just strings.

## Why It's Good For The Game

Easier to understand species code is always better, and the defines will definitely be useful for consistency. (Doing damage to `ORGAN_LIVER` rather than having to hope the species uses `"liver"`.)

## Changelog
N/A


SteelSlayer (approved):

Fox-McCloud:
Have some concerns about overlapping defines and how this could potentially impact things, in the future, given our awful handling of HUDs and such.

Putting an objection onto this to prevent accidental merging.

code/__DEFINES/mobs.dm
```diff
@@ -196,6 +196,29 @@
...
+#define ORGAN_GROIN  "groin"
+#define ORGAN_HEAD	 "head"
+#define ORGAN_ARM_L  "l_arm"
+#define ORGAN_ARM_R  "r_arm"
+#define ORGAN_LEG_L  "l_leg"
+#define ORGAN_LEG_R  "r_leg"
+#define ORGAN_HAND_L "l_hand"
+#define ORGAN_HAND_R "r_hand"
+#define ORGAN_FOOT_L "l_foot"
+#define ORGAN_FOOT_R "r_foot"
```
Fox-McCloud, line 220: Potentially a future problem creator.

Fox-McCloud (changes requested):

code/__DEFINES/mobs.dm
SabreML: Sorry, could you clarify what the problem is here? I'm not sure I understand.

SabreML: I'll put this here too so it doesn't get lost to time:
image

code/modules/mob/living/carbon/human/species/vox.dm
```diff
@@ -160,13 +160,13 @@
...
-		"kidneys" =  /obj/item/organ/internal/kidneys,
-		"cortical stack" =    /obj/item/organ/internal/brain/vox,
-		"eyes" =     /obj/item/organ/internal/eyes, //Default darksight of 2.
-		)												//for determining the success of the heist game-mode's 'leave nobody behind' objective, while this is just an organ.
+		ORGAN_HEART	   = /obj/item/organ/internal/heart/vox,
+		ORGAN_LUNGS	   = /obj/item/organ/internal/lungs/vox,
+		ORGAN_LIVER	   = /obj/item/organ/internal/liver/vox,
+		ORGAN_KIDNEYS  = /obj/item/organ/internal/kidneys/vox,
+		ORGAN_BRAIN	   = /obj/item/organ/internal/brain/vox,
+		ORGAN_EYES	   = /obj/item/organ/internal/eyes/vox, //Default darksight of 2.
```
SabreML, line 168: No mechanical changes here other than a 1.6x alcohol intensity from the liver, and all organs being sterile. This is just so that the sprites and descriptions are consistent between species.

SabreML (commented):
No possible mechanical changes other than the review below, since it's just replacing strings with defines. Unsetting from DNM.

dearmochi:
There are now two maintainer objections on this PR and thus protocol dictates that it be closed. 

The idea of using defines is good but it has been deemed inappropriate in this particular case due to future-proofing concerns (the merging of arms/hands and legs/feet, for example) as well as the potential for code inconsistency (e.g. golem organs which remain hard coded).

I apologize that it took some time to process this PR only for this outcome, your effort is still very much appreciated.

PR #16085, opened 2021-05-25
Rad Collector Scaling and Define Clarity

Arvenius169:
## What Does This PR Do
Reintroduces the feature of plasma tank contents scaling the power generated by rad collectors.
Scaling is now easily tunable to balance/preference through defines.

Defines are now more clearly named and commented.

## Why It's Good For The Game
Makes sense, engages atmos, does not break power generation balance.

## Changelog
:cl:
tweak: Rad collector output now scales on plasma tank contents.
tweak: Renamed/added defines in collector.dm for clarity.
/:cl:

### DIFF
--- a/code/modules/power/singularity/collector.dm
+++ b/code/modules/power/singularity/collector.dm
@@ -1,6 +1,7 @@
-// stored_energy += (pulse_strength - RAD_COLLECTOR_EFFICIENCY) * RAD_COLLECTOR_COEFFICIENT
-#define RAD_COLLECTOR_EFFICIENCY 80 	// radiation needs to be over this amount to get power
-#define RAD_COLLECTOR_COEFFICIENT 100
+// stored_energy += (pulse_strength - RAD_COLLECTOR_THRESHOLD) * RAD_COLLECTOR_POWER_FACTOR
+#define RAD_COLLECTOR_THRESHOLD 80 	// radiation needs to be over this amount to get power
+#define RAD_COLLECTOR_POWER_FACTOR 300	//this scales the maximum theoretical power of the collector
+#define RAD_COLLECTOR_EFFICIENCY 50	//how many moles of plasma will get you halfway closer to maximum power
 #define RAD_COLLECTOR_STORED_OUT 0.04	// (this * 100)% of stored power outputted per tick. Doesn't actualy change output total, lower numbers just means collectors output for longer in absence of a source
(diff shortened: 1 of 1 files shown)


Arvenius169:
Values are currently tuned such that from-the-rack plasma tanks will produce ~34% of old power value, full tanks of room-temperature plasma will produce ~100% of old power value, and the maximum limit that you approach as you cram more cold is 3x the old power value.

SteelSlayer:
You say in the description "does not break power generation balance", but your comment mentions you can attain "3x the old power value". I'm not sure you can increase maximum power generation by 300%, and also say it doesn't mess with power balance. 

And yes I know power is essentially meaningless at the moment, with the current SM being able to supply the station easily, but the concern still stands.

Arvenius169:
> 
> 
> You say in the description "does not break power generation balance", but your comment mentions you can attain "3x the old power value". I'm not sure you can increase maximum power generation by 300%, and also say it doesn't mess with power balance.
> 
> And yes I know power is essentially meaningless at the moment, with the current SM being able to supply the station easily, but the concern still stands.

As I said, the math can be really easily tuned to preference.  Power is essentially meaningless, but there's a plasma canister in storage specifically for the purpose of filling tanks. Using atmos to squeeze more power out of the engine is fun, and is still a feature most people remember. I feel this is a much better implementation because it doesn't scale infinitely.

dearmochi (approved):
Approved for code

SteelSlayer (approved):

Fox-McCloud:
The current power output of default operation is more than enough to power the station.

I'd much rather tinkering with the gas composition and how to manage that be the angle of "more power" rather than "I put more plasma in, so it produces more power"

👎

Arvenius169:
> 
> 
> The current power output of default operation is more than enough to power the station.
> 
> I'd much rather tinkering with the gas composition and how to manage that be the angle of "more power" rather than "I put more plasma in, so it produces more power"
> 
> 👎

It's not my intention for this to be the sole method of controlling the power of the supermatter.

It certainly makes sense to me that increasing the contents of the radiation collectors from room temperature gas to supercooled gas would improve the performance of the engine, and even more so, it makes sense to me to require filling up the plasma tanks all the way before putting them in, else you suffer a sizeable penalty to power production.

Provided you aren't disgusted with that idea, I would appreciate some input for how you personally would tune the power output in 3 cases: lazy unfilled tanks, filled tanks, and filled cold tanks. I can get back to you with an implementation of that intent.

variableundefined:

PR #16095, opened 2021-05-26
Allows (D)eswords to cut through walls like a welder

Qwertytoforty:
## What Does This PR Do

Allows energy and dual swords to cut through walls like a welder. Does not work on reinforced walls.
Changes the message slightly for plasmacutters and esword wall welding to mention what was used to cut the wall down.
Adds a toolspeed of 1 to dual eswords
Adds a can_cut_walls variable to energy melee weapons,  mainly used on the cleaving saw to be false.

## Why It's Good For The Game

Mainly because it is cool, and everyone wants to summon their inner Darth Vader to slice down a wall with their esword.

This doesn't really have any major balence issues either, welding tools already were small, and could be carried just as easy as an esword, not to mention you can find them everywhere. The esword takes just as long as a welder, it is no more effective than a welder, and can not be used to unweld or weld doors, or repair stuff like a welder. Plus, if you are welding down a wall with an esword, 50/50 someone is going to come up to you and disarm you anyway.

## Changelog
:cl:
tweak: You can now weld down walls with an energy sword or dual energy sword.
tweak: Plasmacutters no longer take extra time to deconstruct diamond walls.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -280,7 +280,6 @@
 	max_integrity = 200
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 100, "acid" = 70)
 	resistance_flags = FIRE_PROOF
-	toolspeed = 1
 	light_power = 2
 	needs_permit = TRUE
 	var/brightness_on = 2
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -280,6 +280,7 @@
 	max_integrity = 200
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 100, "acid" = 70)
 	resistance_flags = FIRE_PROOF
+	toolspeed = 1
 	light_power = 2
 	needs_permit = TRUE
 	var/brightness_on = 2
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -284,6 +284,7 @@
 	needs_permit = TRUE
 	var/brightness_on = 2
 	var/colormap = list(red=LIGHT_COLOR_RED, blue=LIGHT_COLOR_LIGHTBLUE, green=LIGHT_COLOR_GREEN, purple=LIGHT_COLOR_PURPLE, rainbow=LIGHT_COLOR_WHITE)
+	var/toy = FALSE //Is this a toy?
 
 /obj/item/twohanded/dualsaber/New()
 	..()
--- a/code/game/turfs/simulated/walls.dm
+++ b/code/game/turfs/simulated/walls.dm
@@ -371,28 +371,25 @@
 	return FALSE
 
 /turf/simulated/wall/proc/try_decon(obj/item/I, mob/user, params)
+	var/cutting = FALSE
+	var/speed = 10 //Time to cut down the wall, plasmacutter is faster
 	if(istype(I, /obj/item/gun/energy/plasmacutter))
-		to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
-		playsound(src, I.usesound, 100, 1)
-		if(do_after(user, I.toolspeed * 6 SECONDS, target = src))
-			cut_down(I, user)
-			return TRUE
-	if(istype(I, /obj/item/melee/energy))
+		cutting = TRUE
+		speed = 6
+	else if(istype(I, /obj/item/melee/energy))
 		var/obj/item/melee/energy/O = I
-		if(O.can_cut_walls && O.active)
-			to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
-			playsound(src, 'sound/items/welder.ogg', 100, TRUE)
-			if(do_after(user, O.toolspeed * 10 SECONDS, target = src))
-				cut_down(I, user)
-				return TRUE
-	if(istype(I, /obj/item/twohanded/dualsaber))
+		cutting = O.can_cut_walls && O.active
+	else if(istype(I, /obj/item/twohanded/dualsaber))
 		var/obj/item/twohanded/dualsaber/O = I
-		if(O.wielded && !O.toy)
-			to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
-			playsound(src, 'sound/items/welder.ogg', 100, 1)
-			if(do_after(user, O.toolspeed * 10 SECONDS, target = src))
-				cut_down(I, user)
-				return TRUE
+		cutting = O.wielded && !O.toy
+	if(!cutting)
+		return FALSE
+
+	to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
+	playsound(src, 'sound/items/welder.ogg', 100, TRUE)
+	if(do_after(user, I.toolspeed * speed SECONDS, target = src))
+		cut_down(I, user)
+		return TRUE
 	return FALSE
 
 /turf/simulated/wall/proc/cut_down(obj/item/I, mob/living/user, params)
--- a/code/game/turfs/simulated/walls.dm
+++ b/code/game/turfs/simulated/walls.dm
@@ -370,7 +370,7 @@
 		return TRUE
 	return FALSE
 
-/turf/simulated/wall/proc/try_decon(obj/item/I, mob/living/user, params)
+/turf/simulated/wall/proc/try_decon(obj/item/I, mob/user, params)
 	if(istype(I, /obj/item/gun/energy/plasmacutter))
 		to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
 		playsound(src, I.usesound, 100, 1)
--- a/code/game/turfs/simulated/walls.dm
+++ b/code/game/turfs/simulated/walls.dm
@@ -370,15 +370,36 @@
 		return TRUE
 	return FALSE
 
-/turf/simulated/wall/proc/try_decon(obj/item/I, mob/user, params)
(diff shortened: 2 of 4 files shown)


Cocacolagua:
I can feel miss-clicking the wall will end up with people getting blinded, so yeah.

matttheficus:
What @Cocacolagua said. Jedi didnt need eyewear, I dont think Crew needing goggles is needed. I do like the addition a lot.

Rurik123:
I second Cocacolagua's concerns there. I already blind myself over and over in fights trying to welder someone in maints cause I accidently clicked the pipe under them.

S34NW:
He's looking right at it!

image

Jokes aside, while we don't do stuff just because star wars does it, I think I agree that it should only require welding protection of the same level as the adv. welder.

Wojz:
I don't think adding the mediocre benefit of no flash from welding down a wall to an 8 TC S-class melee weapon that doesn't stun would be that big of an issue. It's just rule of cool, doesn't really change anything.

Qwertytoforty:
Eye flashing from the energy sword on walls has been removed

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -280,6 +280,7 @@
 	max_integrity = 200
 	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 100, "acid" = 70)
 	resistance_flags = FIRE_PROOF
+	toolspeed = 1
 	light_power = 2
```
SabreML, line 284: The `toolspeed` here is the same as the parent variable:
```dm
var/toolspeed = 1 // If this item is a tool, the speed multiplier
```
```suggestion
	resistance_flags = FIRE_PROOF
	light_power = 2
```

code/game/turfs/simulated/walls.dm
```diff
@@ -370,15 +370,34 @@
 		return TRUE
 	return FALSE
 
-/turf/simulated/wall/proc/try_decon(obj/item/I, mob/user, params)
+/turf/simulated/wall/proc/try_decon(obj/item/I, mob/living/user, params)
```
SabreML, line 373: Why specify `/living` here?

code/game/turfs/simulated/walls.dm
```diff
@@ -370,15 +370,34 @@
...
+			return TRUE
+	if(istype(I, /obj/item/melee/energy))
+		var/obj/item/melee/energy/O = I
+		if(O.can_cut_walls && O.active)
+			to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
+			playsound(src, 'sound/items/welder.ogg', 100, 1)
+			if(do_after(user, O.toolspeed * 10 SECONDS, target = src))
+				to_chat(user, "<span class='notice'>You remove the outer plating.</span>")
+				dismantle_wall()
+				visible_message("<span class='warning'>[user] slices apart [src] with [O]!</span>", "<span class='warning'>You hear metal being sliced apart.</span>")
```
SabreML, line 390: All of these should really be made into their own proc to avoid having the same thing copy+pasted three times.
e.g:
```dm
/turf/simulated/wall/proc/decon_wall(obj/item/I, mob/user, speed, tool_sound)
	to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
	playsound(src, tool_sound, 100, TRUE)
	if(do_after(user, speed, target = src))
		dismantle_wall()
		user.visible_message("<span class='warning'>[user] slices apart [src] with [I]!</span>", "<span class='notice'>You remove the outer plating.</span>", "<span class='warning'>You hear metal being sliced apart.</span>")
		return TRUE
```

SabreML, line 390: This `to_chat()` and `visible_message()` next to eachother means that whoever cut the wall down gets two chat messages.
It was already this way before of course, but you may as well move it to the `visible_message()` while you're here.

code/game/turfs/simulated/walls.dm
```diff
@@ -370,15 +370,34 @@
...
+				return TRUE
+	if(istype(I, /obj/item/twohanded/dualsaber))
+		var/obj/item/twohanded/dualsaber/O = I
+		if(O.wielded && !(istype(O, /obj/item/twohanded/dualsaber/toy)))
```
SabreML, line 394: Since the toy version is a subtype you could probably replace this `istype()` with an `is_toy` variable or something on the dsword, and check for that instead.

code/game/turfs/simulated/walls.dm
```diff
@@ -370,15 +370,34 @@
...
+			to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
+			playsound(src, 'sound/items/welder.ogg', 100, 1)
+			if(do_after(user, O.toolspeed * 10 SECONDS, target = src))
+				to_chat(user, "<span class='notice'>You remove the outer plating.</span>")
+				dismantle_wall()
+				visible_message("<span class='warning'>[user] slices apart [src] with [O]!</span>", "<span class='warning'>You hear metal being sliced apart.</span>")
 			return TRUE
```
SabreML, line 401: The `return TRUE` here should probably be indented under the `do_after()`, otherwise it'll still return TRUE even if it fails.

code/game/turfs/simulated/walls.dm
```diff
@@ -370,15 +370,34 @@
...
 			to_chat(user, "<span class='notice'>You remove the outer plating.</span>")
 			dismantle_wall()
-			visible_message("<span class='warning'>[user] slices apart [src]!</span>", "<span class='warning'>You hear metal being sliced apart.</span>")
+			visible_message("<span class='warning'>[user] slices apart [src] with [I]!</span>", "<span class='warning'>You hear metal being sliced apart.</span>")
+			return TRUE
+	if(istype(I, /obj/item/melee/energy))
+		var/obj/item/melee/energy/O = I
+		if(O.can_cut_walls && O.active)
+			to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
+			playsound(src, 'sound/items/welder.ogg', 100, 1)
```
SabreML, line 386: Small thing, fourth argument is a boolean.
```suggestion
			playsound(src, 'sound/items/welder.ogg', 100, TRUE)
```

code/game/turfs/simulated/walls.dm
Qwertytoforty: Fixed

code/game/turfs/simulated/walls.dm
Qwertytoforty: done

code/game/turfs/simulated/walls.dm
Qwertytoforty: Proc made

Kyep (approved):

code/game/turfs/simulated/walls.dm
AffectedArc07: Echoing this. What is the point of `living` here.

*Do not take this out of context*

code/game/turfs/simulated/walls.dm
```diff
@@ -370,19 +370,36 @@
...
+				cut_down(I, user)
+				return TRUE
+	if(istype(I, /obj/item/twohanded/dualsaber))
+		var/obj/item/twohanded/dualsaber/O = I
+		if(O.wielded && !O.toy)
+			to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
+			playsound(src, 'sound/items/welder.ogg', 100, 1)
+			if(do_after(user, O.toolspeed * 10 SECONDS, target = src))
+				cut_down(I, user)
+				return TRUE
```
AffectedArc07, line 395: This entire check is super snowflakey, but I cant think of anything better on the top of my head.

code/game/turfs/simulated/walls.dm
Qwertytoforty: Removed

code/game/turfs/simulated/walls.dm
dearmochi: +1, it might be worthwhile and not too out of scope to refactor the proc in a similar manner as:

```dm
var/slice = FALSE
if(istype(I, /obj/item/gun/energy/plasmacutter))
	slice = TRUE // Kinda interesting this doesn't check for power, but oh well
else if(istype(I, /obj/item/melee/energy))
	var/obj/item/melee/energy/O = I
	slice = O.can_cut_walls && O.active
else if(istype(I, /obj/item/twohanded/dualsaber))
	var/obj/item/twohanded/dualsaber/O = I
	slice = O.wielded && !O.toy
if(!slice)
	return FALSE

to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
playsound(src, 'sound/items/welder.ogg', 100, 1)
if(do_after(user, O.toolspeed * 10 SECONDS, target = src))
	cut_down(I, user)
	return TRUE
return FALSE
```

code/game/turfs/simulated/walls.dm
```diff
@@ -374,15 +374,32 @@
 	if(istype(I, /obj/item/gun/energy/plasmacutter))
 		to_chat(user, "<span class='notice'>You begin slicing through the outer plating.</span>")
 		playsound(src, I.usesound, 100, 1)
-
-		if(do_after(user, istype(sheet_type, /obj/item/stack/sheet/mineral/diamond) ? 120 * I.toolspeed : 60 * I.toolspeed, target = src))
```
dearmochi, line 378: Unintended change about diamond walls no longer taking more time to deconstruct?

Qwertytoforty: I removed it due to being incredibly hyper specific, with diamond walls almost never made, as well as welding tools and sonic jackhammer not having this slowdown. I can add it back in, if it is prefered

code/game/turfs/simulated/walls.dm
Qwertytoforty: Done!

Fox-McCloud:
I'm generally not inclined to buff one of the strongest melee weapons in the game.

variableundefined:

PR #16096, opened 2021-05-26
Makes Display Cases Emaggable

Leanfrog:
## What Does This PR Do
This PR adds the ability to emag display cases bypassing their access requirement such as the case containing the Captain's antique laser. Note, this does still set off an alarm

## Why It's Good For The Game
The inability to emag a display case is arguably confusing, unlike every other "locked" object or structure on the primary station display cases can only be opened with a corresponding ID, or by breaking them open which sets off an alarm if it has one built in, it's rather inconsistent. Adding the option to emag the case brings it in line with every other access restricted structure on station and adds an option for a slightly more stealthy avenue if one so wishes to take it,

## Changelog
:cl:
add: Display cases can now be emagged to remove the access requirement for opening
/:cl:

### DIFF
--- a/code/game/objects/structures/displaycase.dm
+++ b/code/game/objects/structures/displaycase.dm
@@ -35,11 +35,12 @@
 	QDEL_NULL(showpiece)
 	return ..()
 
-/obj/structure/displaycase/emag_act(user as mob)
+/obj/structure/displaycase/emag_act(mob/user)
 	if(!emagged)
 		to_chat(user, "<span class ='warning'>You override the ID lock on the [src].</span>")
+		trigger_alarm()
 
-		emagged = 1
+		emagged = TRUE
 
 /obj/structure/displaycase/examine(mob/user)
 	. = ..()
--- a/code/game/objects/structures/displaycase.dm
+++ b/code/game/objects/structures/displaycase.dm
@@ -35,10 +35,18 @@
 	QDEL_NULL(showpiece)
 	return ..()
 
+/obj/structure/displaycase/emag_act(user as mob)
+	if(!emagged)
+		to_chat(user, "<span class ='warning'>You override the ID lock on the [src].</span>")
+
+		emagged = 1
+
 /obj/structure/displaycase/examine(mob/user)
 	. = ..()
 	if(alert)
 		. += "<span class='notice'>Hooked up with an anti-theft system.</span>"
+	if(emagged)
+		. += "<span class='warning'>The ID lock has been shorted out.</span>"
 	if(showpiece)
 		. += "<span class='notice'>There's [showpiece] inside.</span>"
 	if(trophy_message)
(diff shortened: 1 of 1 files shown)


S34NW:
Balance wise, I personally think the alarm should still go off when emagging it.
This is one of the few items you can almost guarantee will be in one place (short of during a major crisis).

The ability to stealth in all the way with an emag AND stealthily steal it seems a bit much.

datlo:
I second triggering the alarm when emagging it. There's already an option for stealing it stealthily : Emag/unlock the APC and turn off the power.

code/game/objects/structures/displaycase.dm
```diff
@@ -35,10 +35,18 @@
 	QDEL_NULL(showpiece)
 	return ..()
 
+/obj/structure/displaycase/emag_act(user as mob)
+	if(!emagged)
+		to_chat(user, "<span class ='warning'>You override the ID lock on the [src].</span>")
+
+		emagged = 1
```
AffectedArc07, line 42:
```suggestion
		emagged = TRUE
```

code/game/objects/structures/displaycase.dm
```diff
@@ -35,10 +35,18 @@
 	QDEL_NULL(showpiece)
 	return ..()
 
+/obj/structure/displaycase/emag_act(user as mob)
```
AffectedArc07, line 38:
```suggestion
/obj/structure/displaycase/emag_act(mob/user)
```

AffectedArc07 (changes requested):
Please see requested changes, include S34's

Leanfrog:
Emagging the case will now set off the display cases alarm

code/game/objects/structures/displaycase.dm
```diff
@@ -35,10 +35,19 @@
 	QDEL_NULL(showpiece)
 	return ..()
 
+/obj/structure/displaycase/emag_act(mob/user)
+	if(!emagged)
+		to_chat(user, "<span class ='warning'>You override the ID lock on the [src].</span>")
```
SabreML, line 40: Remove the space here so that it's consistent with the rest of them. LGTM otherwise.
```suggestion
		to_chat(user, "<span class='warning'>You override the ID lock on the [src].</span>")
```

farie82: Can remove `the` there. Byond should add that itself if needed

Leanfrog: That's not on the emag, that's on the display case

code/game/objects/structures/displaycase.dm
```diff
@@ -35,10 +35,19 @@
 	QDEL_NULL(showpiece)
 	return ..()
 
+/obj/structure/displaycase/emag_act(mob/user)
+	if(!emagged)
+		to_chat(user, "<span class='warning'>You override the ID lock on the [src].</span>")
```
dearmochi, line 40:
```suggestion
		to_chat(user, "<span class='warning'>You override the ID lock on [src].</span>")
```

dearmochi (approved):
Fine otherwise

Kyep (approved):

AffectedArc07 (approved):

PR #16108, opened 2021-05-28
Optimizes customizable foods slightly + makes stacking beyond the limit of 41 impossible to avoid the server from dying

farie82:
## What Does This PR Do

- Makes adding ingredients to customizable snacks less performance heavy.
- Makes it impossible to add customizable snacks together if that would make it go over the ingredient limit. Sorry not sorry. This limit can be increased later on if needed. But I decided this is out of the scope of this PR for now.
- Gives the burger a bottom bun as it's supposed to. It looks ass though so if a spriter is willing to make a better one then please do.
- Fixes putting burgers on a dish from exploding into the ingredients on the dish.
- Also refactors the file a bit to make it slightly saner.
- Creating custom food now auto puts it in your active hand instead of on the floor

prevents this from happening:
```
Proc Name                                                                Self CPU    Total CPU    Real Time     Overtime        Calls
/obj/item/reagent_containers/food/snacks/customizable/proc/updateicon      33.912       33.912       33.916       33.437          489
```

Tested by making 3 full-stack burgers (41 is the limit due to the > check)
```
Proc Name                                                                Self CPU    Total CPU    Real Time     Overtime        Calls
Old:
/obj/item/reagent_containers/food/snacks/customizable/proc/updateicon      0.032        0.032        0.032        0.000          123
New:
/obj/item/reagent_containers/food/snacks/customizable/proc/add_ingredients 0.006        0.014        0.013        0.000          123
```
Small improvements here. But the actual server halting performance issue can't be reproduced again so I didn't compare those

## Why It's Good For The Game
Stops the server from hanging for a few seconds when the chef wants to make the worlds biggest ~~lag~~ burger.
Others are bug fixes.

## Images of changes
image
Known issue, the stacking on the plate does look wack. But it at least works

## Changelog
:cl:
tweak: Creating custom food now auto puts it in your active hand instead of on the floor
fix: Fixes custom burgers from lagging out the server. This does mean that they're limited in size (which was meant to be the case)
fix: Gives custom burgers their bottom sprite back
fix: Makes you able to put custom burgers on custom dishes without them exploding into their ingredients on the dish
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/food/customizables.dm
+++ b/code/modules/food_and_drinks/food/customizables.dm
@@ -58,7 +58,6 @@
 	icon_state = "breadslice"
 	baseicon = "sandwichcustom"
 	basename = "sandwich"
-	toptype = new /obj/item/reagent_containers/food/snacks/breadslice()
 
 
 
@@ -70,15 +69,23 @@
 	var/baseicon = "sandwichcustom"
 	var/basename = "sandwichcustom"
 	bitesize = 4
-	var/top = 1	//Do we have a top?
-	var/obj/item/toptype
+	/// Do we have a top?
+	var/top = TRUE
+	/// The image of the top
+	var/image/top_image
 	var/snack_overlays = 1	//Do we stack?
 //	var/offsetstuff = 1 //Do we offset the overlays?
-	var/sandwich_limit = 40
+	var/ingredient_limit = 40
 	var/fullycustom = 0
 	trash = /obj/item/trash/plate
 	var/list/ingredients = list()
 	list_reagents = list("nutriment" = 8)
+
+/obj/item/reagent_containers/food/snacks/customizable/Initialize(mapload)
+	. = ..()
+	if(top)
+		top_image = new(icon, "[baseicon]_top")
+		add_overlay(top_image)
 
 /obj/item/reagent_containers/food/snacks/customizable/pizza
 	name = "personal pizza"
@@ -303,7 +310,7 @@
 	basename = "on a plate"
 	snack_overlays = 0
 	top = 0
-	sandwich_limit = 20
+	ingredient_limit = 20
 	fullycustom = 1
 
 /obj/item/reagent_containers/food/snacks/customizable/soup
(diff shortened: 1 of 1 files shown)


code/modules/food_and_drinks/food/customizables.dm
```diff
@@ -321,71 +328,98 @@
...
-	if(contents.len > sandwich_limit)
+	if(!istype(I, /obj/item/reagent_containers/food/snacks))
+		to_chat(user, "<span class='warning'>[I] isn't exactly something that you would want to eat.</span>")
+		return
+
+	add_ingredient(I)
+
+/**
+ * Tries to add one ingredient and it's ingredients, if any and applicable, to this snack
+ * @param snack The ingredient that will be added
```
AffectedArc07, line 346: `@param` isnt autodoc syntax
---
```
/**
  * Short description of the proc
  *
  * Longer detailed paragraph about the proc
  * including any relevant detail
  * Arguments:
  * * arg1 - Relevance of this argument
  * * arg2 - Relevance of this argument
  */

```

code/modules/food_and_drinks/food/customizables.dm
```diff
@@ -321,71 +328,98 @@
...
+	name = newname()
 
-/obj/item/reagent_containers/food/snacks/customizable/proc/updateicon()
-	overlays = 0
-	var/i=0
-	for(var/obj/item/O in ingredients)
-		i++
+/**
+ * Adds a list of ingredients to the existing snack. Updates the overlays as well
+ * @param new_ingredients The new ingredients to be added
```
AffectedArc07, line 384: Ditto

code/modules/food_and_drinks/food/customizables.dm
farie82: Ah, I forgot which one it was again and found this one when searching for it. Will change

Carthusia:
slight QOL thing that maybe would fit into this PR:

Anychance you could make it so that when combining stuff into custom foods, it stays in your hand, or on the table it was on when you began the combination instead of always falling onto the floor at your feet.

Currently when you begin the combo process, say adding a tomato to a bun, or bowl or sandwich, it always appears at your feet, further additions remain where it was, but the first one always falls to your feet, which is fairly annoying when trying to make bulk custom foods.

farie82:
> slight QOL thing that maybe would fit into this PR:
> 
> Anychance you could make it so that when combining stuff into custom foods, it stays in your hand, or on the table it was on when you began the combination instead of always falling onto the floor at your feet.
> 
> Currently when you begin the combo process, say adding a tomato to a bun, or bowl or sandwich, it always appears at your feet, further additions remain where it was, but the first one always falls to your feet, which is fairly annoying when trying to make bulk custom foods.

I've noticed that as well when coding it. It should be a small change so why not!

Carthusia:
Huzzah! Death to floor foods!

dearmochi (approved):

code/modules/food_and_drinks/food/customizables.dm
```diff
@@ -1,43 +1,43 @@
+#define MAKE_CUSTOM_FOOD(snack_to_add, user, type) \
```
SteelSlayer, line 1: Interesting choice to make this a define. I'm curious, did it noticeably help performance?

farie82: Mostly made it since I didn't want to make a proc on item level to make a snack. Seeing that you can combine things on a plate (which ain't a snack). I made it mostly to reduce having to rewrite the same code multiple times :D

SteelSlayer: Ah yeah, didn't notice one of the items was a `/obj/item/trash/plate`, while others are snacks, so that makes sense.

AffectedArc07 (approved):

PR #16110, opened 2021-05-29
Adds pref to remove internals pressure from status tab, adds it to inspect text

### LINKED ISSUE #15684: Having internals on in the 'Status' tab causes a framerate drop
**Issue Description**:
Having the 'Status' tab open with internals turned on reduces the game's frame rate by a considerable amount.
Apparently GPS coordinates showing can also cause this, but I haven't tested that myself.

The game performance would stay stable, maybe with a *slight* decrease due to all of the values on the Status tab.

**What happened instead**:
Noticable framerate drop from 60 to maybe around 30. I haven't tried getting exact numbers yet.

*Video Demonstration with Internals, GPS, and Spacepod stats on a test server: (Not sure if the GPS and Spacepod actually make a difference here)*

https://user-images.githubusercontent.com/57483089/110226669-2f3def80-7ee9-11eb-9530-f5b6a277be1c.mp4

*Far more obvious example recorded on the live server:*

https://user-images.githubusercontent.com/57483089/120053296-4f310b00-c021-11eb-8560-ff87ed2b3136.mp4

It forces me to have a different tab open if I actually want to have any chance in combat, or even just walking around if it gets bad enough, which means I can't see things like the escape shuttle timer or internals pressure.

Activate internals and switch between 'Status' and any other tab. There should be an obvious difference.
The longer the round goes on for, the worse it gets.

I only noticed this problem when I started playing Vox since they have constant internals, so anywhere from about two weeks ago.

**Extra information**:
It *looks* like it's just a hardware/CPU problem on my end, but the majority of people I've spoken to about this have also experienced it.

S34NW:
## What Does This PR Do
This PR adds a preference to remove the lag-inducing internals readout from the status tab, and adds a quick way to check internals pressure by inspecting the tank.
Fixes #15684

## Why It's Good For The Game
You should not be forced to lag because you wear a tank, this will still keep it easy to read pressure but remove a source of frustration for many.
This now gives you the option of disabling the largest cause of lag on the status panel.

## Images of changes
Pref:
image

Result:
image
image

Inspecting tank:
image

Status tab with toggle on:
image

## Changelog
:cl:
add: Added a pressure reading to the inspect text of internals.
add: Added a preference to remove the detailed internals info on the status panel.
/:cl:


AffectedArc07:
This causes lag for 1 person (as far as I know), and having the status of your tank is very important to be in a stat pane.

Framerate with no tank on
image

Framerate with a tank
image

As you can see, there is no change, and this is with me playing on my laptop on battery. 

This isn't a game problem, nor is tweaking this in the game the proper solution


TLDR: Not in favour

SabreML:
This does make it slightly less convenient to see how much is left in the tank, but I at least consider that more than worth the payoff.
This *is* a game problem, since it's an unfixable issue for around half the people I've asked about it. Even if not everybody gets it, the people who *do* get it have their FPS crippled.

Having the most important stat tab pretty much unusable by a good chunk of the playerbase is more than enough reason to make a small change ingame, in my opinion.

meow20:
If im not a vox, i don't know my tank is out of pressure until im choking and getting no O2 message, and i suffer from serious lag from having status page up.

Mitchs98:
I have the same problem myself, though I hadn't actually noticed it until Sabre mentioned it. But I've tested it several times and do indeed get frame drops, quite badly, when I have internals on and the stat page open. It likely doesn't effect everyone but I have seen far more than one person mention it as an issue.

Rurik123:
Same exact problem, as I commented a while ago on the bug report.

> This isn't a game problem, nor is tweaking this in the game the proper solution

Any idea where the problem comes from, then? I'd glady fiddle with some settings to try and fix it but I honestly have no idea what would even be effecting FPS for the chunk of us, especially for something as niche as setting internals on.

Varghh:
> This causes lag for 1 person (as far as I know), and having the status of your tank is very important to be in a stat pane.

Add me and my wife too to that list too (with 3 different computers), the lag is massive when you turn on the tank with the status tab on.

shmabuu:
To add onto the count of people that get lag from this, I get really terrible frame dips when my status tab is open and I have internals enabled. 

When it's said that having the pressure of an internals tank in the status tab is important I have to disagree entirely. Very rarely do I find myself actually looking at the status tab for pressure of my internals tank, I always use the internals tank in hand to see how much is left. The pressure of internals is important but I do enjoy my framerate.

It is incredibly inconvenient how when using alt click to look at items on the floor if moved away from the game lags if internals are on.

AffectedArc07:
Dropping my objection as a lot of people seem to be impacted by this, but I still want a preference that lets you enable/disable reduced statpanels. It shouldnt be too hard, just make a new toggle, doesnt even need DB edits.

S34NW:
> 
> 
> Dropping my objection as a lot of people seem to be impacted by this, but I still want a preference that lets you enable/disable reduced statpanels. It shouldnt be too hard, just make a new toggle, doesnt even need DB edits.

I think a preference is fair enough, I'll get that updated soon

code/__DEFINES/preferences.dm
```diff
@@ -55,6 +55,7 @@
 // Yes I know this being an "enable to disable" is misleading, but it avoids having to tweak all existing pref entries
 #define PREFTOGGLE_2_REVERB_DISABLE	512
 #define PREFTOGGLE_2_FORCE_WHITE_RUNECHAT	1024
+#define PREFTOGGLE_2_SIMPLE_STAT_PANEL	2048
 
 #define TOGGLES_2_TOTAL 			2047 // If you add or remove a preference toggle above, make sure you update this define with the total value of the toggles combined.
```
AffectedArc07, line 60:
```suggestion
#define TOGGLES_2_TOTAL 			4095 // If you add or remove a preference toggle above, make sure you update this define with the total value of the toggles combined.
```

This needs updating. Also test your code with a local DB please.

code/modules/client/preference/preferences_toggles.dm
```diff
@@ -348,3 +348,11 @@
 	prefs.toggles2 ^= PREFTOGGLE_2_FORCE_WHITE_RUNECHAT
 	prefs.save_preferences(src)
 	to_chat(src, "Your runechats will [(prefs.toggles2 & PREFTOGGLE_2_FORCE_WHITE_RUNECHAT) ? "no longer" : "now"] be forced to be white.")
+
+/client/verb/toggle_simple_stat_pa()
```
AffectedArc07, line 352:
```suggestion
/client/verb/toggle_simple_stat_pane()
```

AffectedArc07 (changes requested):

code/__DEFINES/preferences.dm
S34NW: Done, works as intended

AffectedArc07 (approved):

PR #16112, opened 2021-05-29
Fixes air alarms not showing on the central atmospherics computer's map view

SteelSlayer:
## What Does This PR Do
Map standardization changed the station z-level to 2, but the atmospherics computer was still searching on z-1. This updates to the code to search on z-2.

## Why It's Good For The Game
Bug fix good.

## Images of changes
image

## Changelog
:cl:
fix: Fixes air alarms not showing on the central atmospherics computer's map view
/:cl:


tgui/packages/tgui/interfaces/AtmosControl.js
```diff
@@ -125,7 +125,7 @@ const AtmosControlMapView = (_properties, context) => {
   return (
     <Box height="526px" mb="0.5rem" overflow="hidden">
       <NanoMap onZoom={v => setZoom(v)}>
-        {alarms.filter(a => a.z === 1).map(aa => (
+        {alarms.filter(a => a.z === 2).map(aa => (
```
TDSSS, line 128: Do we have to hardcode the z level here? It might be better to just get the station z level directly.

AffectedArc07: This is TGUI so you either pass forward redundant data or hardcode it.

Normal DM procs dont work in TGUI

SteelSlayer: It will be a little annoying, but yeah I could pass in `level_name_to_num(MAIN_STATION)` from DM if its wanted.

AffectedArc07: Given how infrequently these are changed, I am deeming this not worth doing

AffectedArc07 (approved):

PR #16116, opened 2021-05-30
Adds a Cube themed gateway map

Nazsgull:
## What Does This PR Do
A matrix of rooms where each one could randomly contain anything, from hostile mobs to items to nothing at all.

Items and hostile mobs spawns:
-x6 Random tool
-x1 Toolbox
-x1 Random bookcase
-Around 50 random item spawners
-Multiple Syndicate turrets
-Multiple random slimes
-Multiple hostile mob spawners (alien sentinels, goliaths, red terror spiders)
-A opened toxins canister
-A opened N2O canister
-A small piece of lore
-x1 Energy gun

## Why It's Good For The Game
Adds variety to the pool of gateway maps, which is good because the other maps are not that many.

## Images of changes
cube

## Changelog
:cl:
add: Added the gateway map.
/:cl:


FlattestGuitar:
Revolutionary map design.

SabreML:
Personally I like the idea of a randomly generated gateway/space ruin in theory, but I'm not sure this is the best way to go about it.
Maybe having a few large rooms rather than lots of small ones would work, since then you could theme each section around something different. (One with carp, one with lavaland stuff, one with hivebots maybe, etc.)

Still might work better as a ruin than as a full gateway, but it'd be interesting at least.

Cocacolagua:
- If you are going to do this concept of multiple cubes where you "don't know what you will get!", you should randomize if rooms are going to be positive/neutral/bad rather than keeping it static and instead randomizing the exact loot.
- Giving no weapons to fight off the creatures (like the wireless laser guns the terror gateway uses) will also encourage people to bumrush the loot while ignoring everything else.
- The room design is very uninteresting, even if it is the intention it doesn't make it any better.
- One of the corners has a syndicate level4 loot table which is what the syndicate depot uses for the armory, please do not give nuclear operative guns / offensive traitor items to gateway explorers.

Overall, this concept /can/ be done well, but this is not the way to go about it.

Nazsgull:
> * If you are going to do this concept of multiple cubes where you "don't know what you will get!", you should randomize if rooms are going to be positive/neutral/bad rather than keeping it static and instead randomizing the exact loot.
> * Giving no weapons to fight off the creatures (like the wireless laser guns the terror gateway uses) will also encourage people to bumrush the loot while ignoring everything else.
> * The room design is very uninteresting, even if it is the intention it doesn't make it any better.
> * One of the corners has a syndicate level4 loot table which is what the syndicate depot uses for the armory, please do not give nuclear operative guns / offensive traitor items to gateway explorers.
> 
> Overall, this concept /can/ be done well, but this is not the way to go about it.

I'll fix these issues and try again with another PR, I wasn't very convinced either, should have given it more thought.

PR #16119, opened 2021-05-31
Rebalences teslium blob slightly

Qwertytoforty:
## What Does This PR Do

Adds a new teslium subtype, blob teslium, which has a much more stable reaction in someones body, shocking them much less.

## Why It's Good For The Game

Currently, teslium blob gets people wired, generating a shocking ammount of salt (Sorry, had to make the joke.)

In all honesty, it did need a re-balance, so I have made a tweak to the likelihood of the teslium from the blob shocking the person while it is in their user. 

Before, on average, you would be stunned every _18 seconds, or 9 cycles_, from teslium, with a 80% chance to be stunned after 12 cycles. This was too strong, and thus the subtype teslium the blob gives is heavily nerfed. Now, its a 50/50 chance you will be stunned every 20 cycles, or 40 seconds. It can be higher or lower, but on average that is how long it will be.

## Images of changes

For those curious as well, here are the current probabilities for teslium blob and spore stunning you when attacking you. Note that the first attack can not stun currently. (sorry for not rounding)

image

By the time it is 50/50 blob attacks stun you, you will have already have taken 90 damage anyway, where other blob types would have put you in crit. That is also the blob spending 25-30 chemicals on one person. Just in case anyone was curious about those chances.

## Changelog
:cl:
tweak: Teslium blob now uses a subtype of the teslium reagent, which stuns much less frequently.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/toxins.dm
+++ b/code/modules/reagents/chemistry/reagents/toxins.dm
@@ -1228,17 +1228,19 @@
 	color = "#20324D" //RGB: 32, 50, 77
 	metabolization_rate = 0.2
 	var/shock_timer = 0
-	var/normal_teslium = TRUE //Is this the base teslium, or the blob teslium. Used so it can be a subtype of teslium without inheriting the random shock from normal teslium as well as it's own
+
+	//The random values assigned to the high and low number for the random time for shocks
+	var/shock_high = 30
+	var/shock_low = 5
+	var/use_chaotic_random = TRUE //Do you change the random number every cycle, or after shock is triggered
+	var/chosen_timer = 15 //The timer chosen last cycle, or every time they are shocked
+
 	process_flags = ORGANIC | SYNTHETIC
 	taste_description = "electricity"
 
 /datum/reagent/teslium/on_mob_life(mob/living/M)
-	if(normal_teslium)
-		shock_timer++
-		if(shock_timer >= rand(5,30)) //Random shocks are wildly unpredictable
-			shock_timer = 0
-			M.electrocute_act(rand(5, 20), "Teslium in their body", 1, SHOCK_NOGLOVES) //Override because it's caused from INSIDE of you
-			playsound(M, "sparks", 50, TRUE, SHORT_RANGE_SOUND_EXTRARANGE)
+	shock_timer++
+	shock_proc(M)
 	return ..()
 
 /datum/reagent/teslium/on_mob_add(mob/living/M)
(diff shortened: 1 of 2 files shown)


code/modules/reagents/chemistry/reagents/toxins.dm
```diff
@@ -1228,16 +1228,18 @@
...
 	taste_description = "electricity"
 
 /datum/reagent/teslium/on_mob_life(mob/living/M)
-	shock_timer++
-	if(shock_timer >= rand(5,30)) //Random shocks are wildly unpredictable
-		shock_timer = 0
-		M.electrocute_act(rand(5, 20), "Teslium in their body", 1, SHOCK_NOGLOVES) //Override because it's caused from INSIDE of you
-		playsound(M, "sparks", 50, TRUE, SHORT_RANGE_SOUND_EXTRARANGE)
-	return ..()
+	if(normal_teslium)
```
SteelSlayer, line 1236: Since blob teslium will not be able to pass this check, the `return ..()` is never reached and 
`/datum/reagent/proc/on_mob_life` will not be called for it. From what I can tell, this means the reagent never decays inside the mob, which is bad obviously. You'll probably want to move the `return ..()` left a tab, so it's always called.

Qwertytoforty: Yeah, that might be a good idea

code/modules/reagents/chemistry/reagents/toxins.dm
```diff
@@ -1228,15 +1228,17 @@
...
 	process_flags = ORGANIC | SYNTHETIC
 	taste_description = "electricity"
 
 /datum/reagent/teslium/on_mob_life(mob/living/M)
-	shock_timer++
-	if(shock_timer >= rand(5,30)) //Random shocks are wildly unpredictable
-		shock_timer = 0
-		M.electrocute_act(rand(5, 20), "Teslium in their body", 1, SHOCK_NOGLOVES) //Override because it's caused from INSIDE of you
-		playsound(M, "sparks", 50, TRUE, SHORT_RANGE_SOUND_EXTRARANGE)
+	if(normal_teslium)
```
farie82, line 1236: I feel like it would be better to have a var for minimum shock timer or something. Or a proc which handles the shocking.
Both implementations are the same except for the `chosen_timer` part

farie82 (changes requested):

code/modules/reagents/chemistry/reagents/toxins.dm
AffectedArc07: I agree here. Please make it without an entire override with an if that cuts out half the proc, otherwise you may as well make it a new type instead of a subtype.

AffectedArc07 (changes requested):
See my above comment

code/modules/reagents/chemistry/reagents/toxins.dm
Qwertytoforty: Made it a proc

AffectedArc07 (approved):

PR #16123, opened 2021-06-02
Replaces between() With Native clamp()

Arvenius169:
## What Does This PR Do
Replaces a global proc with a native proc that does the same thing.
Removes the global proc.

## Why It's Good For The Game
There's absolutely no reason for between() to exist.

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -538,10 +538,6 @@ Returns 1 if the chain up to the area contains the given typepath
 	var/y = min(world.maxy, max(1, A.y + dy))
 	return locate(x,y,A.z)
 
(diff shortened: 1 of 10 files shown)


hal9000PR:
CL isn't needed here as it isn't a player focused change.

Arvenius169:
> CL isn't needed here as it isn't a player focused change.

Noted. If I should remove the now empty changelog alongside the contents, let me know.

S34NW (approved):
why someone made their own proc for this is beyond me, implementation of clamp() looks good.

AffectedArc07:
> why someone made their own proc for this is beyond me, implementation of clamp() looks good.

Clamp only became native in 513 so it's likely someone didn't notice the helper.

farie82 (approved):
Ensure to fully remove the CL tags if you have no CL entries

SabreML (approved):

dearmochi (approved):

PR #16124, opened 2021-06-02
Fixes bad inits on space tiles

### LINKED ISSUE #16080: Invisible Disposal Pipe in Botany Maintenance
**Problem Description**:
In maintenance just north of Botany and the Library, there's a small 2-square section of space between windows. Various pipes run through both these sections of space.
Lately, I've noticed these disposal pipes spawn invisible, and not functional. Trash collects in space there.
What is the problem?

Trash should continue through these disposal pipes to arrive at disposals.

**What happened instead**:
Trash ejects from these invisible chutes into the small area of space.

It means anything mailed from Cargo won't reach it's destination past Botany, and any trash from before Library won't get to disposals.

So far, this problem seems to happen upon world generation. I would attempt to restart the world, and then perhaps mail something to Engineering, then watch the pipes at Botany maintenance.

I only recently noticed the problem, it could have been happening for weeks.

I don't know if there were any special gamemodes any of the times I notice this happening.

I wish I had thought to screenshot the invisible pipes BEFORE I fixed them as a maintenance drone, but you can see the collected trash. If you alt+click the section of space there, you can see the Disposal pipe is in place, but it doesn't seem to work. I merely unweld and reweld the pipes to fix them in-game.
Botany Disposals

AffectedArc07:
## What Does This PR Do

Continuation of #15243 

I hate these things. Why is their logic based on icon state. Why is this proc even an override. This is pain.

Fixes #16080 

## Why It's Good For The Game
Fixes bugs

## Changelog
:cl: AffectedArc07
fix: Fixed a case where disposal pipes dont load properly on space tiles 
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -179,7 +179,7 @@
 // override for space turfs, since they should never hide anything
 /turf/space/levelupdate()
 	for(var/obj/O in src)
-		if(O.level == 1)
+		if(O.level == 1 && O.initialized) // Only do this if the object has initialized
 			O.hide(FALSE)
 
 // Removes all signs of lattice on the pos of the turf -Donkieyo


Dragonkiller93:
Maybe put in the changelog that this is the pipe above botany, since that's what players will recognize.

PR #16132, opened 2021-06-06
Standardizes Sec Bomber Jacket Armor And Adds To Loadout

Leanfrog:
## What Does This PR Do
This PR standardizes the security bomber jacket's armor to be in line with the standard security jacket, also adds the bomber jacket to the loadout menu

## Why It's Good For The Game
First off, why does the sec bomber jacket have more armor? It's "exclusive" to pod pilot? I disagree with this as the pod pilot already has a number of unique things to their job and the bomber jacket itself has no indication it would be stronger than the security jacket, bringing it in line with the security jacket creates a slightly more consistent choice in regards to armor. For reference the current security jacket has an armor of 
"(melee = 15, bullet = 10, laser = 15, energy = 5, bomb = 15, bio = 0, rad = 0, fire = 30, acid = 30)" 
while the bomber jacket has 
"(melee = 25, bullet = 15, laser = 25, energy = 10, bomb = 25, bio = 0, rad = 0, fire = 50, acid = 50)".
The addition of the bomber jacket to the loadout is simple, it's a rather stylish piece of equipment allowing for officers to maintain an amount of style alongside their standard protection.

## Changelog
:cl:
add: The security bomber jacket is now available in the loadout menu for security staff
tweak: Reduced the armor values of the bomber jacket to bring it in line with the standard security jacket
/:cl:

### DIFF
--- a/code/modules/client/preference/loadout/loadout_suit.dm
+++ b/code/modules/client/preference/loadout/loadout_suit.dm
@@ -108,6 +108,11 @@
 	path = /obj/item/clothing/suit/armor/secjacket
 	allowed_roles = list("Head of Security", "Warden", "Detective", "Security Officer", "Security Pod Pilot")
 
+/datum/gear/suit/secbomberjacket
+	display_name = "security bomber jacket"
+	path = /obj/item/clothing/suit/jacket/pilot
+	allowed_roles = list("Head of Security", "Warden", "Detective", "Security Officer", "Security Pod Pilot")
+
 /datum/gear/suit/ianshirt
 	display_name = "Ian Shirt"
 	path = /obj/item/clothing/suit/ianshirt
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -803,7 +803,7 @@
 	max_heat_protection_temperature = ARMOR_MAX_TEMP_PROTECT
 	strip_delay = 60
 	put_on_delay = 40
-	armor = list(melee = 25, bullet = 15, laser = 25, energy = 10, bomb = 25, bio = 0, rad = 0, fire = 50, acid = 50)
+	armor = list(melee = 15, bullet = 5, laser = 15, energy = 5, bomb = 15, bio = 0, rad = 0, fire = 30, acid = 30)
 	//End of inheritance from Security armour.
 
 /obj/item/clothing/suit/jacket/leather


S34NW (approved):

matttheficus:
As a note, the Bomber Jacket does not have a pocket like the Standard Security Jacket does. IMO its a fair trade to lose 2 slots of storage (that fit 80% of all Sec Items anyway) in exchange for some extra armor.

Leanfrog:
> As a note, the Bomber Jacket does not have a pocket like the Standard Security Jacket does. IMO its a fair trade to lose 2 slots of storage (that fit 80% of all Sec Items anyway) in exchange for some extra armor.

Are you certain? Looking at the code security jackets do not have pockets unless I missed something, and I do not happen to recall them having pockets either.

Wojz:
> As a note, the Bomber Jacket does not have a pocket like the Standard Security Jacket does. IMO its a fair trade to lose 2 slots of storage (that fit 80% of all Sec Items anyway) in exchange for some extra armor.

they don't have pockets

hal9000PR:
Why not make them all the same as sec armour so they are just cosmetic choices?

Leanfrog:
> Why not make them all the same as sec armour so they are just cosmetic choices?

Sec jackets cover the arms as well as the upper body, that's the trade off, do you want more coverage or higher armor. Also really wouldn't make sense to have a jacket that doesn't protect the arms

Stokes52:
I don't like the idea of making the jacket available to everyone. The bomber jacket is their signature uniform item to show their unique status. A bomber jacket is an aviator style jacket and the pod pilot is the "aviator" role on the station.

Besides, if you really want the jacket, you can just ask the pod pilot for their spare jacket. No opinion on the balance change, but it should not be a loadout item imo.

SabreML:
I've seen multiple people take the pod pilot role for the sole reason of getting the fancy jacket, and then play as a regular officer for the whole round, not even touching their spacepod once.
Letting these people have their fashion *without* them needing to waste a 30 karma point job slot can only be a good thing, in my opinion.

hal9000PR:
> I've seen multiple people take the pod pilot role for the sole reason of getting the fancy jacket, and then play as a regular officer for the whole round, not even touching their spacepod once.
> Letting these people have their fashion _without_ them needing to take a 30KP job slot away from people who actually want to play the role can only be a good thing, in my opinion.

The reason they never touch the pod is it is worthless. A hardsuit is much better and any officer can get that. The pod is only good during blob.

So the jacket *is* the only unique part about the job

Leanfrog:
> The reason they never touch the pod is it is worthless. A hardsuit is much better and any officer can get that. The pod is only good during blob.

I have to disagree, I've seen the pod put to good use a number of times, the weapons on it are actually quite potent against crew and it's essentially extra life that is easily replenished with a welder, not to mention that it's incredibly easy to gain a very high battery capacity on it allowing for some shenanigans with constant weapon spam, side note the pod mining laser does a very large amount of damage. It's not a matter of the pod being bad, more so people just not using it for some reason.

Varghh:
Less snowflake, more agency.
Good change

variableundefined (approved):

PR #16134, opened 2021-06-07
Fixes some turf related exploits

Qwertytoforty:
## What Does This PR Do
 
Fixes some turf related exploits in RCDS and the Theme_Warp Crystal

## Why It's Good For The Game

No. Just no. You should not be able to do this.

## Changelog
:cl:
fix: Fixes some turf related exploits.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/RCD.dm
+++ b/code/game/objects/items/weapons/RCD.dm
@@ -428,6 +428,8 @@
 	if(iswallturf(A))
 		if(istype(A, /turf/simulated/wall/r_wall) && !canRwall)
 			return FALSE
+		if(istype(A, /turf/simulated/wall/indestructible))
+			return FALSE
 		if(checkResource(5, user))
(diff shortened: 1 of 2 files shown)


ppi13:
Thats weird, you can't even rcd indestructible walls to begin with.
Also, wow, breaking into centcom with that crystal, lel.

code/game/objects/items/weapons/RCD.dm
```diff
@@ -428,6 +428,8 @@
 	if(iswallturf(A))
 		if(istype(A, /turf/simulated/wall/r_wall) && !canRwall)
 			return FALSE
+		if(istype(A, /turf/simulated/wall/indestructible))
+			return FALSE
```
ppi13, line 432: this should probably be defines under walls_indestructible.dm, if it is not already.

AffectedArc07 (approved):

PR #15447, opened 2021-02-05
Buffs Potassium Iodine and Pentetic Acid's radiation healing.

hal9000PR:
## What Does This PR Do
Increases how much radiation potassium iodine and pentetic acid purges by a factor of 10

## Why It's Good For The Game
Ever since the radiation refractor, it has been exceptionally easy to accrue a massive amount of rads during SM delams because you aren't the CE. The chamber of the SM during a delamination is, extremely hot, radioactive, and pressurised. It is best to be immune to the pressure and heat effects with an atmos hardsuit (outside of the CE's hardsuit) because this is the most direct source of damage, however this means you can easily stack up 8k+ rads. Currently that would take 1000 cycles to remove with both potassium iodine and pentetic in you at the same time. This isn't feasible. Hopefully this buff will prevent people being ancored to medbay for 40+ minutes constantly having pentetic and iodine in you and chems to counteract the negatives of pentetic.  Sitting in medbay for 40 minutes, with both chems in you constantly, to 44 HOURS with just potassium iodine in you isn't exactly fun or feasible for a poor slime person that they cannot clone or the medical doctors and chemists constantly making and supplying chems.  

## Changelog
:cl:
tweak: increased potassium iodine and pentetic adic's radiation healing by a factor of 10
/:cl:

PR #15331, opened 2021-01-22
Radiation Refactor

Fox-McCloud:
Ports and implement's TG's radiation refactor. Credit to TG for most of this

TG's radiation refactor is a far-reaching and highly impactful refactor of how rads works; instead of just being something that is snowflaked everywhere and only really impacts mobs---it now impacts *everything*.

Generally this consolidates rad handling for radiation emitting objects with `radiation_pulse`, whereby it'll bathe an area in radiation.

Radiation can now accumulate and build up on objects, making them contaminated---a contaminated object will emit radiation on its own, which will, in turn, irradiate other objects---the end result of this is that if something is highly contaminated, it'll contaminate other objects, with will, in turn, contaminate other objects, with the nonsense spreading further and further---typically, though, most radiation sources won't be enough to contaminate things---unless you're dealing with the Supermatter. To get rid of contamination, you'll have to wash things in a shower.

Note to admins: there' an investigate log to see just how crazy this gets, in case someone gets the bright idea to irradiate the heck out of a closet full of paper then drag it around to contaminate everything.

Things that are contaminated glow an eery green color and have a custom examine text:

img

Other changes in this PR:

- Implements Geiger counters, which detect radiation, or, if emagged, can release radiation
  - Hardsuit helmets have built in Geiger counters of their own. You can acquire regular geiger counters from an autolathe
- Implement's Engine Goggles
  - These are special goggles that can act as mesons, act as a t-ray scanner, or see the radioactive contamination amount of objects. These can be printed at R&D--some engineers start with some. There's also a subtype that doesn't have mesons, but has expanded t-ray scan range that atmos techs have.
- T-ray scanners have been refactored and no longer display things to the world around you; it's only you that can see "through" the floor now

Code stuffs:
- Added in a filter based  system that more easily allows adding visual filters for how things render (the radiation contamination uses this for now)
- Moved `emagged` var to be object level
- Made `water_act` an atom level proc (why wasn't it before?)
- cleaned up t-ray scanner code a bit
- Reduced memory usage by making atmos pipes created on demand as opposed to stored in pipes, themselves

This is still technically WIP--there's a couple of things I need to address and tweak:

- radioactive reagents are pretty pathetic and don't really do much
- Need to balance out the injectors and DNA modifier rad pulses too

:cl: Fox McCloud
add: Radiation has been dramatically refactored; it can now contaminate other things and spread--if you should get contaminated, it's highly recommended you wash the contamination off with a shower
/:cl:

PR #16147, opened 2021-06-10
Increases most radiation dose amounts

### LINKED ISSUE #15819: Radiation is extremely slow.
**Issue Description**:
Do note. I'm not exactly sure if this is a bug or not, it _seems_ to me that it is, but it might not be. Bare in mind this is NOT taking into mind contaminated objects, I did not test this, just straight radiation.

Essentially, damage from radiation is basically nonexistent until you reach 500 rads. Even at that point it's so extremely minor that it's a mere inconvenience, by this I mean like. 1-2 toxins here and there and nothing more. It took me roughly 18 minutes for hair to fall out, 816 rads, and toxins to reach 56 damage, bare in mind at this point it was more the toxins from the liver damage that was damaging me more than the radiation. It took 23 minutes, 1.1k toxins, and around 30 liver damage to finally enter crit. 26 minutes to kill, and that death was mostly because RNG deemed it so.

I expected radiation to be a tad more lethal, rather than something you can not care about. This test was done standing next to uranium golem, naturally radiation will scale higher with more radiation producers. The golem also did not irradiate any objects at all, other than the test mob.

**What happened instead**:
It took roughly 1/4th of the round to actually die from radiation exposure, and most of the death wasn't radiation related.

If this is a bug, it makes essentially any form of radiation barring massive doses I.E the SM trivial.

1. Get radiated.
2. Wait

Presumably since the radiation refactor.

**Extra information**:
If this ISN'T a bug, then that's my bad. Do note this isn't intended, in any capacity, to be a feature request. From my point of view, this seems like a bug.

S34NW:
## What Does This PR Do
This PR makes people take 10x as many rads from all sources bar the SM crystal. This is done by multiplying radiation by RAD_DOSAGE_MULTIPLIER (currently 10), this number is easy to tweak in future for balancing and affects all rad sources to keep them consistent.
The SM delamination rad dosage is now also 4-40x as powerful depending on the engine power (EER) at detonation.
This somewhat fixes #15819

## Why It's Good For The Game
Since the rad refactor in #15331 most radiation sources give such trivial amounts of rads, you won't feel a thing. You could make a full uranium room and be fine.
This brings radiation values in line with the SM dosages, as that was the only source that was properly implemented into this new system with the rad refactor.
This does not mean that radiation leads to instant death, and it does not tweak the SM radiation doses as they are fine, but it does mean that radiation is no longer just a joke.

The 10x value comes from looking at how many rads you get from the SM, plus how many rads you currently get. In #15447 anti-rad chems were made 10 times more effective to deal with the SM rads.
Both these chems and old rad sources didn't get updated with the refactor, so the SM rad doses were nearly impossible to fix. The change to the chems however further invalidated other rad sources. This PR balances this all out.

## Changelog
:cl:
tweak: Non-SM radiation doses are now stronger
tweak: SM delamination now more strongly irradiates the station.
/:cl:

### DIFF
--- a/code/__DEFINES/radiation.dm
+++ b/code/__DEFINES/radiation.dm
@@ -54,4 +54,4 @@ Ask ninjanomnom if they're around
 #define RAD_GEIGER_MEASURE_SMOOTHING 5
 #define RAD_GEIGER_GRACE_PERIOD 2
 
-#define RAD_DAMAGE_MULTIPLIER 10					// This brings pre-refactor rad values in line with the current danger values
+#define RAD_DOSAGE_MULTIPLIER 10					// This is used to increase radiation dosage taken during rad_act() calls
--- a/code/__DEFINES/radiation.dm
+++ b/code/__DEFINES/radiation.dm
@@ -53,3 +53,5 @@ Ask ninjanomnom if they're around
 
 #define RAD_GEIGER_MEASURE_SMOOTHING 5
 #define RAD_GEIGER_GRACE_PERIOD 2
+
+#define RAD_DAMAGE_MULTIPLIER 10					// This brings pre-refactor rad values in line with the current danger values
--- a/code/modules/mob/living/damage_procs.dm
+++ b/code/modules/mob/living/damage_procs.dm
@@ -87,7 +87,7 @@
 			Paralyse(effect * blocked)
 		if(IRRADIATE)
 			if(!HAS_TRAIT(src, TRAIT_RADIMMUNE))
-				radiation += max(effect * blocked, 0)
+				radiation += max((effect * RAD_DOSAGE_MULTIPLIER) * blocked, 0)
 		if(SLUR)
 			Slur(effect * blocked)
 		if(STUTTER)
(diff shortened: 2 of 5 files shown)


Wojz:
Once more xray laser spam, when reflected at crew, becomes an actual danger, instead of a minor annoyance. Blobs rejoice. Oh and this also fixes radblob and a bunch of chems, cool.

AsrielFBI:
:thumbsdown: just use TEGs

Wojz:
> 👎 just use TEGs

but what does this have to do with the SM, that's like the one thing that is staying the same in this PR

AsrielFBI:
> 
> 
> > 👎 just use TEGs
> 
> but what does this have to do with the SM, that's like the one thing that is staying the same in this PR

instead of using the rad of the sm for energy production, use the heat of the sm for energy production. And thats accomplished by using the chad TEG instead of the virgin rad collector

S34NW:
> 
> 
> > > 👎 just use TEGs
> > 
> > 
> > but what does this have to do with the SM, that's like the one thing that is staying the same in this PR
> 
> instead of using the rad of the sm for energy production, use the heat of the sm for energy production. And thats accomplished by using the chad TEG instead of the virgin rad collector

The actual hell are you on about

variableundefined (approved):

PR #16148, opened 2021-06-10
Adds warning to gluttony morph about not being an antag

S34NW:
## What Does This PR Do
Adds a large chat message to morphs that transform from gluttony's blessing, saying: 
`You are NOT an antagonist, as such you should never eat steal objectives or the contents of the armory.`

## Why It's Good For The Game
It's not really all that clear to non-midround morphs that they can't do these things. This clears it up with a large chat message.

## Images of changes
image

## Changelog
:cl:
add: Crew who become a morph via Gluttony's Blessing are now told they are not an antagonist.
/:cl:

### DIFF
--- a/code/datums/diseases/transformation.dm
+++ b/code/datums/diseases/transformation.dm
@@ -246,5 +246,5 @@
 	stage3	= list("<span class='danger'>Your appendages are melting away.</span>", "<span class='danger'>Your limbs begin to lose their shape.</span>")
 	stage4	= list("<span class='danger'>You're ravenous.</span>")
 	stage5	= list("<span class='danger'>You have become a morph.</span>")
-	transformation_text = "<span class='userdanger'>You are NOT an antagonist, as such you should never eat steal objectives or the contents of the armory.</span>"
+	transformation_text = "<span class='userdanger'>This transformation does NOT make you an antagonist if you were not one already. If you were not an antagonist, you should not eat any steal objectives or the contents of the armory.</span>"
 	new_form = /mob/living/simple_animal/hostile/morph
--- a/code/datums/diseases/transformation.dm
+++ b/code/datums/diseases/transformation.dm
@@ -40,13 +40,13 @@
 	if(istype(affected_mob, /mob/living/carbon) && affected_mob.stat != DEAD)
 		if(stage5)
 			to_chat(affected_mob, pick(stage5))
-		if(transformation_text)
-			to_chat(affected_mob, transformation_text)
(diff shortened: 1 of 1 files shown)


code/datums/diseases/transformation.dm
```diff
@@ -39,6 +39,8 @@
 	if(istype(affected_mob, /mob/living/carbon) && affected_mob.stat != DEAD)
 		if(stage5)
 			to_chat(affected_mob, pick(stage5))
+		if(name == "Gluttony's Blessing")
```
farie82, line 42: This is quite snowflaky.
It would be better to add a `transformation_text` variable or such. Which if not null will be displayed to the user.
This is also displayed before the `jobban_isbanned` is done. Which could lead to the message being shown while the user not becoming the mob

farie82 (changes requested):

Varghh:
I have a better idea: Just remove them.

farie82 (approved):

Cocacolagua:
should probably word it differently if the morph miner /was/ an antag before-hand

S34NW:
> 
> 
> should probably word it differently if the morph miner /was/ an antag before-hand

I had not considered this, thank you. If an antag becomes a morph, are they still expected or permitted to carry out their objectives?

dearmochi (approved):

hal9000PR:
> > should probably word it differently if the morph miner /was/ an antag before-hand
> 
> I had not considered this, thank you. If an antag becomes a morph, are they still expected or permitted to carry out their objectives?

yes

code/datums/diseases/transformation.dm
```diff
@@ -243,4 +246,5 @@
 	stage3	= list("<span class='danger'>Your appendages are melting away.</span>", "<span class='danger'>Your limbs begin to lose their shape.</span>")
 	stage4	= list("<span class='danger'>You're ravenous.</span>")
 	stage5	= list("<span class='danger'>You have become a morph.</span>")
+	transformation_text = "<span class='userdanger'>You are NOT an antagonist, as such you should never eat steal objectives or the contents of the armory.</span>"
```
S34NW, line 249:
```suggestion
	transformation_text = "<span class='userdanger'>This transformation does NOT make you an antagonist if you were not one already. If you were not an antagonist, you should not eat any steal objectives or the contents of the armory.</span>"
```

Kyep (approved):

AffectedArc07 (approved):

PR #16154, opened 2021-06-11
Fixes broken shuttle and escape pod rotation

S34NW:
## What Does This PR Do
This PR fixes a bug that was introduced in #15454 that caused incorrect shuttle wall and engine dirs when the shuttle was rotated.
I also removed the names "burst", "left", and "right" from shuttle engines, because that got confusing with rotation.

## Why It's Good For The Game
Shuttles and escape pods should never look like a garbled mess.

## Images of changes
2021-06-11 15_30_02-Window
2021-06-11 15_32_03-Window
2021-06-11 15_32_22-Window
2021-06-11 15_32_29-Window

## Changelog
:cl:
fix: Fixed a bug causing shuttle rotation to be broken
/:cl:

### DIFF
--- a/code/game/shuttle_engines.dm
+++ b/code/game/shuttle_engines.dm
@@ -6,6 +6,7 @@
 	armor = list(melee = 100, bullet = 10, laser = 10, energy = 0, bomb = 0, bio = 0, rad = 0, fire = 50, acid = 70) //default + ignores melee
 
 /obj/structure/shuttle/shuttleRotate(rotation)
+	return //This call is needed to properly rotate the object when on a shuttle that is rotated.
 
 /obj/structure/shuttle/engine
 	name = "engine"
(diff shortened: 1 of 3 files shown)


code/game/shuttle_engines.dm
```diff
@@ -6,10 +6,6 @@
 	armor = list(melee = 100, bullet = 10, laser = 10, energy = 0, bomb = 0, bio = 0, rad = 0, fire = 50, acid = 70) //default + ignores melee
 
 /obj/structure/shuttle/shuttleRotate(rotation)
```
AffectedArc07, line 8: Put a return inside this empty proc instead of just a blank override, maybe even with a comment

code/game/turfs/simulated/walls_mineral.dm
```diff
@@ -7,6 +7,8 @@
 	canSmoothWith = null
 	smooth = SMOOTH_TRUE
 
+/turf/simulated/wall/mineral/shuttleRotate(rotation)
```
AffectedArc07, line 10: Ditto

AffectedArc07 (changes requested):

dearmochi (approved):

AffectedArc07 (changes requested):

AffectedArc07 (approved):

PR #16158, opened 2021-06-11
Fixes #16157 by removing the item from the loot list

### LINKED ISSUE #16157: Dark blessing nullrod from lavaland can not be picked up
**Issue Description**:

Dark blessing nullrod from tendril chests can not be picked up

For people to pick it up, get it stuck to their arms, and be able to stab things with it

**What happened instead**:

Players report being unable to pick it up.

Makes the loot useless, and leaves an armblade on the ground no one can pick up

Open chest with dark blessing.
Attempt to pick it up.

ppi13:
## What Does This PR Do

#16021 made it so that you could not pick up or change the hand in which you held and item with nodrop on it, which lead to
 #16157 and that item is more bald bait then anything else.
 Removing dark blessing from lavaland loot is the simpler option as this is not an item that is really wanted, especially with the fact that If you pick it up, you can't drop it

## Why It's Good For The Game

Having an item you can't pick up as loot from the hard to find tendrils, does not for a fun experience make.

## Changelog
:cl: ppi
del: Removes dark blessing from lavaland loot
/:cl:


code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -48,7 +48,7 @@
 		if(14)
 			new /obj/item/nullrod/scythe/talking(src)
 		if(15)
-			new /obj/item/nullrod/armblade(src)
+			new /obj/item/changestone(src)
```
AffectedArc07, line 51: This isnt a fix. This is an alternative.

I also slated for this thing to be removed. Its not accessible and was added before para even existed.
image


If you're going to make a "fix", **actually fix this fucking problem**

AffectedArc07 (changes requested):

S34NW:
I'm not overly convinced that this is the right fix, considering the problem was introduced in a previous PR by yourself... Perhaps dark blessing can be tweaked to function properly without being abstract?

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -48,34 +48,32 @@
 		if(14)
 			new /obj/item/nullrod/scythe/talking(src)
 		if(15)
-			new /obj/item/nullrod/armblade(src)
```
AffectedArc07, line 51: Again. Please actually fix this issue instead of removing it, especially since *you* are the one who caused it.

AffectedArc07 (changes requested):

code/modules/mining/lavaland/necropolis_chests.dm
hal9000PR: All you would have to do is remove the abstract flag from the item. It isn't abstract, it is very real.

PR #16160, opened 2021-06-11
Prevents TK moving objects with people in or buckled to it

Qwertytoforty:
## What Does This PR Do

This PR adds a check to TK, that checks if the object being thrown has a mob in it, or buckled to it. If either of these cases are true, the object will not be able to be thrown.

## Why It's Good For The Game

This PR is good for the game, as it stops people throwing themself around on chairs at mach 10 with TK, going faster than most other objects, being insanely fast at navigating space, and great at escaping things like terrors after smacking them with a crowbar 50 times in a row rapidly. 

Now, while you can still move around 99% of objects still, you can not abuse it as a speed / mobility method.

## Changelog
:cl:
tweak: Telekinesis can no longer throw objects with people in them, or buckled to them.
/:cl:

### DIFF
--- a/code/_onclick/telekinesis.dm
+++ b/code/_onclick/telekinesis.dm
@@ -142,7 +142,7 @@
 		if(focus.buckled_mobs)
 			to_chat(user, "<span class='notice'>This object is too heavy to move with something buckled to it!</span>")
 			return
-		for(var/mob/M in focus.contents)
+		if(length(focus.client_mobs_in_contents))
 			to_chat(user, "<span class='notice'>This object is too heavy to move with something inside of it!</span>")
 			return
 		apply_focus_overlay()
--- a/code/_onclick/telekinesis.dm
+++ b/code/_onclick/telekinesis.dm
@@ -139,6 +139,12 @@
 
 
 	else
+		if(focus.buckled_mobs)
+			to_chat(user, "<span class='notice'>This object is too heavy to move with something buckled to it!</span>")
+			return
+		for(var/mob/M in focus.contents)
+			to_chat(user, "<span class='notice'>This object is too heavy to move with something inside of it!</span>")
+			return
 		apply_focus_overlay()
 		focus.throw_at(target, 10, 1, user)
 		last_throw = world.time


code/_onclick/telekinesis.dm
```diff
@@ -139,6 +139,12 @@
 
 
 	else
+		if(focus.buckled_mobs)
+			to_chat(user, "<span class='notice'>This object is too heavy to move with something buckled to it!</span>")
+			return
+		for(var/mob/M in focus.contents)
+			to_chat(user, "<span class='notice'>This object is too heavy to move with something inside of it!</span>")
+			return
```
AffectedArc07, line 147: I feel like there should be a better way to do this, but I also aint sure off the top of my head.

Qwertytoforty: Yeah, this is the best I could figure out

code/_onclick/telekinesis.dm
```diff
@@ -139,6 +139,12 @@
 
 
 	else
+		if(focus.buckled_mobs)
+			to_chat(user, "<span class='notice'>This object is too heavy to move with something buckled to it!</span>")
+			return
+		for(var/mob/M in focus.contents)
```
AffectedArc07, line 145:
```suggestion
		if(length(focus.client_mobs_incontents))
```
Try this. It seems a lot less hacky.

code/_onclick/telekinesis.dm
```diff
@@ -139,6 +139,12 @@
 
 
 	else
+		if(focus.buckled_mobs)
+			to_chat(user, "<span class='notice'>This object is too heavy to move with something buckled to it!</span>")
+			return
+		if(length(focus.client_mobs_in_contents))
```
AffectedArc07, line 145: Did you make sure this worked locally?

Qwertytoforty: This prevent TK if a player is in it yes, however it will not prevent if say a ckeyless mob is in it, like a monkey or xenobio slime. Unsure if you want that or not.

AffectedArc07: This is fine.

AffectedArc07 (approved):

PR #5328, opened 2016-08-04
Makes the swedish gene more swedish

pinatacolada:
957288bc2b

Also because I'm not sure if git will fuck up the encoding or not:
d4e5bd814e

:cl: Pinatacolada
add: Makes the swedish gene spit out weird swedish chars and sounds
/:cl:

PR #16168, opened 2021-06-12
Makes E-Bolas as they should be. Noncyborg variants.

RadiantFlash7:
## What Does This PR Do

This changes the type of snare given by the security bola to just be a normal energy snare.
Currently it's set to cyborg(?) variant. For some reason. 
((The coding mistake was noted by HighwayMan in discord))
## Why It's Good For The Game

This makes the limited supply Security bolas actually be better then ones you can make with stuff on station. Furthermore, this issue is well over 5 years old, and was likely made in a coding error and never realized.

## Changelog
:cl:
tweak: Makes Energy bolas work like they should per code.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/legcuffs.dm
+++ b/code/game/objects/items/weapons/legcuffs.dm
@@ -129,7 +129,7 @@
 	icon_state = "e_snare"
 	trap_damage = 0
 	flags = DROPDEL
-	breakouttime = 60
+	breakouttime = 6 SECONDS
 
 /obj/item/restraints/legcuffs/beartrap/energy/New()
 	..()
--- a/code/game/objects/items/weapons/legcuffs.dm
+++ b/code/game/objects/items/weapons/legcuffs.dm
@@ -129,6 +129,7 @@
 	icon_state = "e_snare"
 	trap_damage = 0
 	flags = DROPDEL
+	breakouttime = 60
 
 /obj/item/restraints/legcuffs/beartrap/energy/New()
 	..()
--- a/code/game/objects/items/weapons/legcuffs.dm
+++ b/code/game/objects/items/weapons/legcuffs.dm
@@ -192,7 +192,7 @@
 
 /obj/item/restraints/legcuffs/bola/energy/throw_impact(atom/hit_atom)
 	if(iscarbon(hit_atom))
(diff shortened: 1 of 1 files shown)


S34NW:
So, I don't know HOW #5328 broke this, considering it's not showing in the diff on the PR. But it definitely first appeared in this commit on that PR... https://github.com/ParadiseSS13/Paradise/pull/5328/commits/7d7c9aa6471c528b477e8d6a52326813cb44146b

S34NW (changes requested):
Please add 
`breakouttime = 6 SECONDS` 
On a new line under line 131, so that ebolas are the proper 6-second snare, not 30!

RadiantFlash7:
> Please add
> `breakouttime = 6 SECONDS`
> On a new line under line 131, so that ebolas are the proper 6-second snare, not 30!

Breakout time added to 131.

code/game/objects/items/weapons/legcuffs.dm
```diff
@@ -129,6 +129,7 @@
 	icon_state = "e_snare"
 	trap_damage = 0
 	flags = DROPDEL
+	breakouttime = 60
```
AffectedArc07, line 132:
```suggestion
	breakouttime = 6 SECONDS
```

AffectedArc07 (changes requested):

code/game/objects/items/weapons/legcuffs.dm
RadiantFlash7: RIP deciseconds. Change Applied.

S34NW (approved):

hal9000PR:
This is a pretty hefty buff to an already powerful item. They are already more powerful than makeshift bolas in a way because they fit into pockets. Also security gets what, 20 of these? So yes they are limited supply but they arent exactly rare.

RadiantFlash7:
> This is a pretty hefty buff to an already powerful item. They are already more powerful than makeshift bolas in a way because they fit into pockets. Also security gets what, 20 of these? So yes they are limited supply but they arent exactly rare.

There is currently a total of 14 availible for security to grab. Thats about 2 per officer for a 2 hour shift. Half of these are within the wardens storage and have to be handed out by the warden, seemingly rarely done in practice.

It would be a bit of a buff, but security lacks a way of acquiring more, and several officers already just make makeshift ones instead of energy guns to use.

AffectedArc07 (approved):

PR #16177, opened 2021-06-13
Gives nuke ops an ion carbine instead of an ion rifle

Qwertytoforty:
## What Does This PR Do
 

This PR swaps out the ion rifle in the nukie locker area, with an ion carbine. The only difference between the ion carbine and the rifle is a carbine is normal size, and thus fits in your bag.

## Why It's Good For The Game

Right now, there is no real point to taking the rifle. If anything, it's a noob trap. It's bulky, meaning you have to constantly hold it, or sacrifice your backpack. Some nukies take it thinking it is a fancy weapon, and then don't have space to hold a real one. Generally you need one EMP to cause a stun on cyborgs, which a grenade can do easier, or just using a normal gun shreds already. It might be useful for mechs, but its not worth sacrificing a backpack for it. This way, a nukie will actually be able to take an ion carbine and maybe possibly use it effectively, rather then it being left behind like usual.

## Changelog
:cl:
tweak: The ion rifle in the nuke ops base was replaced with an ion carbine.
/:cl:

### DIFF
--- a/_maps/map_files/generic/centcomm.dmm
+++ b/_maps/map_files/generic/centcomm.dmm
@@ -2971,7 +2971,7 @@
 /area/syndicate_mothership)
 "kL" = (
 /obj/structure/table,
-/obj/item/gun/energy/ionrifle,
+/obj/item/gun/energy/ionrifle/carbine,
 /turf/simulated/floor/plasteel/dark,
 /area/syndicate_mothership)
 "kM" = (


S34NW (approved):

Wojz:
"Monkey's paw" Qwerty you-

dearmochi (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #16183, opened 2021-06-14
minor external organ refactor

hal9000PR:
## What Does This PR Do
makes fixing and causing internal bleeding its own proc
makes internal bleeding and disfiguration use bitflags rather than being their own vars.

Unbreakable and sturdy limbs are no longer subtypes, instead they are edited depending on a trait that the species has. they are also both bitflags now of the same var `limb_flags`

## Why It's Good For The Game
having procs for internal bleeding keeps consistency so if they are edited in the future no instances of it get missed out.

Having unbreakable bones and such be a trait instead of being subtypes of organs means that it is easier to add new permanent damage types, for example, burn wounds. This is because instead of making a new subtype of organ for `unburnable` and `unbreakable/unburnable` and `unbreakable/sturdy/unburnable` for legs, arms, etc. you can just add a single trait to a species.

## Changelog
N/A


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1261,11 +1261,13 @@
...
 		for(var/obj/item/organ/external/E in bodyparts)
-			if(E.brute_dam == 0 && E.burn_dam == 0 && E.internal_bleeding == FALSE) //If there's no damage we don't bother remembering it.
+			if(E.brute_dam == 0 && E.burn_dam == 0 && !(E.status & ORGAN_INT_BLEEDING)) //If there's no damage we don't bother remembering it.
 				continue
 			var/brute = E.brute_dam
 			var/burn = E.burn_dam
-			var/IB = E.internal_bleeding
+			var/IB
+			if(E.status & ORGAN_INT_BLEEDING)
+				IB = TRUE
```
SteelSlayer, line 1270:
```suggestion
			var/IB = (E.status & ORGAN_INT_BLEEDING)
```

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1292,10 +1294,8 @@
 				if((E.type == OE.type)) // Type has to be explicit, as right limbs are a child of left ones etc.
 					var/brute = part[2]
 					var/burn = part[3]
-					var/IB = part[4]
 					//Deal the damage to the new organ and then delete the entry to prevent duplicate checks
 					E.receive_damage(brute, burn, ignore_resists = TRUE)
-					E.internal_bleeding = IB
```
FlattestGuitar, line 1298: This sorta looks like it was removed on accident

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -32,7 +32,7 @@
 	if(stat == DEAD)
 		handle_decay()
 
-	//Update our name based on whether our face is obscured/disfigured
+	//Update our name based on whether our face is obscured/
```
FlattestGuitar, line 35: `/`

code/modules/surgery/organs/organ_external.dm
```diff
@@ -638,6 +633,23 @@ Note that amputating the affected organ does in fact remove the infection from t
...
+	if(is_robotic())
+		return
+	if(HAS_TRAIT(owner, NO_BLOOD))
+		return
+	status |= ORGAN_INT_BLEEDING
+	owner.custom_pain("You feel something rip in your [name]!")
+
+/obj/item/organ/external/proc/fix_internal_bleeding()
+	if(is_robotic())
+		return FALSE
```
FlattestGuitar, line 646: I don't care enough to check, but can't a bodypart *become* robotic? 

You wouldn't want someone to be stuck with bleeding then.

hal9000PR: uhhh- through augmentation yes. *however* you cannot even start the surgery to fix IB on a robotic limb so- that probably could do with fixing.

code/modules/mob/living/carbon/human/life.dm
hal9000PR: oops

code/modules/mob/living/carbon/human/human.dm
hal9000PR: was an accident *however* this code is bugged. no damage is carried over between species swaps.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -477,7 +477,7 @@
 //Returns "Unknown" if facially disfigured and real_name if not. Useful for setting name when polyacided or when updating a human's name variable
 /mob/living/carbon/human/proc/get_face_name()
 	var/obj/item/organ/external/head = get_organ("head")
-	if(!head || head.disfigured || cloneloss > 50 || !real_name || HAS_TRAIT(src, TRAIT_HUSK))	//disfigured. use id-name if possible
+	if(!head || head.status & ORGAN_DISFIGURED || cloneloss > 50 || !real_name || HAS_TRAIT(src, TRAIT_HUSK))	//disfigured. use id-name if possible
```
farie82, line 480:
```suggestion
	if(!head || (head.status & ORGAN_DISFIGURED) || cloneloss > 50 || !real_name || HAS_TRAIT(src, TRAIT_HUSK))	//disfigured. use id-name if possible
```
Gotta be between `()`s due to the order of execution.

code/modules/surgery/organs/organ_external.dm
```diff
@@ -738,13 +750,14 @@ Note that amputating the affected organ does in fact remove the infection from t
...
-	if(disfigured)
+	if(status & ORGAN_DISFIGURED)
 		return
 	if(owner)
 		owner.visible_message("<span class='warning'>You hear a sickening sound coming from \the [owner]'s [name] as it turns into a mangled mess!</span>",	\
 							  "<span class='danger'>Your [name] becomes a mangled mess!</span>",	\
 							  "<span class='warning'>You hear a sickening sound.</span>")
-	disfigured = TRUE
+
+	 status |= ORGAN_DISFIGURED
```
farie82, line 760:
```suggestion
	status |= ORGAN_DISFIGURED
```
This is on another level of indentation

code/modules/surgery/organs/subtypes/standard.dm
```diff
@@ -250,9 +250,10 @@
 
 /obj/item/organ/external/head/receive_damage(brute, burn, sharp, used_weapon = null, list/forbidden_limbs = list(), ignore_resists = FALSE, updating_health = TRUE)
 	..()
-	if(!disfigured)
-		if(brute_dam + burn_dam > 50)
-			disfigure()
+	if(status & ORGAN_DISFIGURED)
+		return
+	if(brute_dam + burn_dam > 50)
+		disfigure()
```
farie82, line 256:
```suggestion
	if(!(status & ORGAN_DISFIGURED) && brute_dam + burn_dam > 50)
		disfigure()
```
Allows for easier adding of new code

code/modules/surgery/robotics.dm
```diff
@@ -222,7 +222,7 @@
 
 	else if(implement_type in implements_heal_brute)
 		current_type = "brute"
-		if(!(affected.brute_dam > 0 || affected.disfigured))
+		if(!(affected.brute_dam > 0 || affected.status & ORGAN_DISFIGURED))
```
farie82, line 225:
```suggestion
		if(!(affected.brute_dam > 0 || (affected.status & ORGAN_DISFIGURED)))
```
Same here as above. Order of execution

farie82 (changes requested):
Looks good for the most part. Also ensure to fix the stuff flattest pointed out

code/modules/mob/living/carbon/human/human.dm
farie82: This would give it the value `ORGAN_INT_BLEEDING`. Which is fine if it's just used for an `if` statement later which most likely will be the solution

SteelSlayer: Huh did not know that, but yeah it should be fine in this case.

code/modules/surgery/organs/organ_external.dm
```diff
@@ -638,6 +633,23 @@ Note that amputating the affected organ does in fact remove the infection from t
...
+	if(HAS_TRAIT(owner, NO_BLOOD))
+		return
+	status |= ORGAN_INT_BLEEDING
+	owner.custom_pain("You feel something rip in your [name]!")
+
+/obj/item/organ/external/proc/fix_internal_bleeding()
+	if(is_robotic())
+		return FALSE
+	if(!(status & ORGAN_INT_BLEEDING))
+		return FALSE
```
SteelSlayer, line 648: I'd say remove this. We don't need to know if they have this flag or not beforehand. If you try to remove a flag from a bitfield which does not have that flag, nothing will happen.
```suggestion
```

code/modules/mob/living/carbon/human/human.dm
hal9000PR: It also looks like this code hasn't been updated from when broken  bones and such were added. so it only carries IB over on transforms. should i fix that in this PR or move it to a different one?

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1287,14 +1285,15 @@
...
-		for(var/B in bodyparts)
-			var/obj/item/organ/external/E = B
+		for(var/obj/item/organ/external/E in bodyparts)
 			for(var/list/part in bodypart_damages)
 				var/obj/item/organ/external/OE = part[1]
 				if((E.type == OE.type)) // Type has to be explicit, as right limbs are a child of left ones etc.
 					var/brute = part[2]
 					var/burn = part[3]
-					//Deal the damage to the new organ and then delete the entry to prevent duplicate checks
+					var/IB = part [4] //Deal the damage to the new organ and then delete the entry to prevent duplicate checks
```
farie82, line 1294:
```suggestion
					var/IB = part[4] //Deal the damage to the new organ and then delete the entry to prevent duplicate checks
```
Spacing

code/modules/surgery/organs/organ_external.dm
```diff
@@ -638,6 +633,21 @@ Note that amputating the affected organ does in fact remove the infection from t
 		owner.handle_splints()
 	return TRUE
 
+/obj/item/organ/external/proc/internal_bleed()
```
farie82, line 636: Perhaps name it something like `cause_internal_bleeding()` or something? `internal_bleed` itself sounds more like a variable

farie82 (commented):
Some minor things

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1285,17 +1285,16 @@
 		dna.species.create_organs(src)
 
 		//Apply relevant damages and variables to the new organs.
-		for(var/B in bodyparts)
-			var/obj/item/organ/external/E = B
+		for(var/obj/item/organ/external/E in bodyparts)
```
farie82, line 1288:
```suggestion
		for(var/obj/item/organ/external/E as anything in bodyparts)
```
can be made to not filter as all `bodyparts` are always of this type.
Reject typeless. `as anything` is the new black

farie82 (changes requested):
Last thing

.vscode/settings.json
```diff
@@ -2,4 +2,4 @@
 	"workbench.editorAssociations": {
 		"*.dmi": "imagePreview.previewEditor"
 	}
-}
\ No newline at end of file
+}
```
hal9000PR, line 5: what

code/modules/surgery/organs/organ_external.dm
```diff
@@ -105,12 +102,23 @@
...
-		sync_colour_to_human(H)
-	get_icon()
+	if(holder)
+		if(ishuman(holder))
+			var/mob/living/carbon/human/H = holder
+			icobase = H.dna.species.icobase
+			if(istype(H))
+				replaced(H)
+				sync_colour_to_human(H)
+			get_icon()
```
SabreML, line 112: `get_icon()` needs to always be called, otherwise the limb won't have any sprites if it's spawned in separately.

Also since this doesn't really do anything if `holder` isn't a human, you should be able to compress it a bit to this:
```dm
/obj/item/organ/external/New(mob/living/carbon/holder)
	..()
	if(ishuman(holder))
		var/mob/living/carbon/human/H = holder
		icobase = H.dna.species.icobase
		replaced(H)
		sync_colour_to_human(H)
	get_icon()
```

code/modules/surgery/generic.dm
```diff
@@ -213,7 +213,8 @@
 	var/obj/item/organ/external/affected = target.get_organ(target_zone)
 	if(affected == null)
 		return 0
-	return !affected.cannot_amputate
+	if((affected.limb_flags & CANNOT_DISMEMBER))
```
dearmochi, line 216:
```suggestion
	if(affected.limb_flags & CANNOT_DISMEMBER)
```

code/modules/surgery/organs/organ_external.dm
```diff
@@ -227,7 +235,7 @@
 	var/mob/living/carbon/owner_old = owner //Need to update health, but need a reference in case the below check cuts off a limb.
 	//If limb took enough damage, try to cut or tear it off
 	if(owner)
-		if(!cannot_amputate && (brute_dam) >= (max_damage))
+		if(!(limb_flags & CANNOT_DISMEMBER) && (brute_dam) >= (max_damage))
```
dearmochi, line 238:
```suggestion
		if(!(limb_flags & CANNOT_DISMEMBER) && brute_dam >= max_damage)
```

dearmochi (approved):
LGTM, just a few nits

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1274,16 +1274,19 @@
...
 		//Loop through all external organs and save the damage states for brute and burn
-		for(var/obj/item/organ/external/E in bodyparts)
-			if(E.brute_dam == 0 && E.burn_dam == 0 && E.internal_bleeding == FALSE) //If there's no damage we don't bother remembering it.
+		for(var/obj/item/organ/external/E as anything in bodyparts)
+			if(E.brute_dam == 0 && E.burn_dam == 0 && !(E.status & ORGAN_INT_BLEEDING)) //If there's no damage we don't bother remembering it.
 				continue
 			var/brute = E.brute_dam
 			var/burn = E.burn_dam
-			var/IB = E.internal_bleeding
+			var/IB = (E.status & ORGAN_INT_BLEEDING)
```
SabreML, line 1289: Not something you've done, but for some reason this proc only tracks broken bones on *internal* organs and not external, so transforming into another species completely heals any fractures.
You should be able to fix that by just moving the checks onto the external loops since (as far as I know) it's impossible for internal organs to get broken in the first place.

SabreML (changes requested):

code/modules/mob/living/carbon/human/human.dm
hal9000PR: ORGAN_BROKEN on internal organs is used when they are necrotic. I was gonna make a separate PR to fix that at after this was merged.

SteelSlayer (approved):

PR #16188, opened 2021-06-15
Fixes a missing icon state on a floor tile under a chair on the emergency shuttle

ppi13:
## What Does This PR Do

Adds a missing icon state to a floor in the emergency escape shuttle

## Why It's Good For The Game

It's an inconsistency

## Images of changes

Before: 
image
After: 
image

## Changelog
:cl: ppi
fix: Adds correct icon state to a floor tile on the emergency escape shuttle
/:cl:


S34NW:
Was this correctly map-merged? as the key for this turf content should already exist

PR #16192, opened 2021-06-16
Prevents built APCs from being hacked, Prevents AI from loosing points from apc destruction

Qwertytoforty:
## What Does This PR Do

This PR makes a subtype of APC, constructed apcs, which can not be hacked by malf AIs. They can still have lights overloaded, power turned off, locked, ect, just like a normal AI would do to the APC
Makes it so deconstructing a malf hacked APC does not cost the malf AI points.

## Why It's Good For The Game

After discussing with a maintainer, it was decided that malf AI should be tweaked to prevent a malf AI from gaining all its points from space built APCs. We are aware however, with the limited APCs on station and off station, so with this PR it is tweaked that if crew deconstructs APCs the malf AI has hacked, they will no longer loose points, meaning once the AI starts going loud with the APC hacking, crew will not be able to destroy APCs the malf AI has hacked to subtract points from them.

For those wondering if command will start deconstructing and repairing the APC on the bridge so the AI can't malfhack it when it turns out the AI is malf, you could already cut the AI control wire to prevent this.

## Changelog
:cl:
tweak: Malf AI can no longer hack APCs that are built by crew / borgs. This does not prevent the AI from interfacing with them like normal.
tweak: Malf AI no longer looses points when Malfhacked APCs are destroyed.
/:cl:

### DIFF
--- a/code/game/objects/items/mountable_frames/apc_frame.dm
+++ b/code/game/objects/items/mountable_frames/apc_frame.dm
@@ -28,5 +28,5 @@
 	return TRUE
 
 /obj/item/mounted/frame/apc_frame/do_build(turf/on_wall, mob/user)
-	new /obj/machinery/power/apc(get_turf(src), get_dir(user, on_wall), 1)
+	new /obj/machinery/power/apc/constructed(get_turf(src), get_dir(user, on_wall), 1)
 	qdel(src)
--- a/code/game/objects/items/mountable_frames/apc_frame.dm
+++ b/code/game/objects/items/mountable_frames/apc_frame.dm
@@ -28,5 +28,5 @@
 	return TRUE
 
 /obj/item/mounted/frame/apc_frame/do_build(turf/on_wall, mob/user)
-	new /obj/machinery/power/apc/constructed(get_turf(src), get_dir(user, on_wall), 1)
+	new /obj/machinery/power/apc(get_turf(src), get_dir(user, on_wall), TRUE)
 	qdel(src)
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -105,6 +105,7 @@
 	var/global/list/status_overlays_environ
 	var/indestructible = 0 // If set, prevents aliens from destroying it
 	var/keep_preset_name = 0
+	var/constructed = FALSE //Was this APC built instead of already existing? Used for malfhack to keep borgs from building apcs in space
 
 	var/report_power_alarm = TRUE
 
@@ -130,6 +131,9 @@
 	req_access = list(ACCESS_SYNDICATE)
 	report_power_alarm = FALSE
 
+
+/obj/machinery/power/apc/constructed // APCS the crew builds
+	constructed = TRUE
 /obj/item/apc_electronics
 	name = "power control module"
 	desc = "Heavy-duty switching circuits for power control."
@@ -177,8 +181,6 @@
 /obj/machinery/power/apc/Destroy()
 	SStgui.close_uis(wires)
 	GLOB.apcs -= src
-	if(malfai && operating)
-		malfai.malf_picker.processing_time = clamp(malfai.malf_picker.processing_time - 10,0,1000)
 	area.power_light = 0
 	area.power_equip = 0
 	area.power_environ = 0
@@ -972,6 +974,9 @@
 		return
 	if(malf.malfhacking)
 		to_chat(malf, "You are already hacking an APC.")
+		return
+	if(constructed)
+		to_chat(malf, "<span class='warning'>This APC was reciently constructed, and not fully linked to station systems, and is of no use to hack.")
 		return
 	to_chat(malf, "Beginning override of APC systems. This takes some time, and you cannot perform other actions during the process.")
 	malf.malfhack = src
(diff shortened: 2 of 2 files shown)


code/modules/power/apc.dm
```diff
@@ -973,6 +975,9 @@
 	if(malf.malfhacking)
 		to_chat(malf, "You are already hacking an APC.")
 		return
+	if(constructed)
+		to_chat(malf, "<span class='warning'>This APC was reciently constructed, and not fully linked to station systems, and is of no use to hack.")
```
AffectedArc07, line 979:
```suggestion
		to_chat(malf, "<span class='warning'>This APC was reciently constructed, and not fully linked to station systems, and is of no use to hack.</span>")
```

code/game/objects/items/mountable_frames/apc_frame.dm
```diff
@@ -28,5 +28,5 @@
 	return TRUE
 
 /obj/item/mounted/frame/apc_frame/do_build(turf/on_wall, mob/user)
-	new /obj/machinery/power/apc(get_turf(src), get_dir(user, on_wall), 1)
+	new /obj/machinery/power/apc/constructed(get_turf(src), get_dir(user, on_wall), 1)
```
AffectedArc07, line 31:
```suggestion
	new /obj/machinery/power/apc/constructed(get_turf(src), get_dir(user, on_wall), TRUE)
```

code/modules/power/apc.dm
```diff
@@ -130,6 +131,9 @@
 	req_access = list(ACCESS_SYNDICATE)
 	report_power_alarm = FALSE
 
+/obj/machinery/power/apc/constructed // APCS the crew builds
```
AffectedArc07, line 134: This subtype isnt needed. Look at line 167. Just assign the var inside that if block.

Qwertytoforty: Done, thanks for the help

code/game/objects/items/mountable_frames/apc_frame.dm
Qwertytoforty: Done

code/modules/power/apc.dm
```diff
@@ -973,6 +973,9 @@
 	if(malf.malfhacking)
 		to_chat(malf, "You are already hacking an APC.")
 		return
+	if(constructed)
+		to_chat(malf, "<span class='warning'>This APC was reciently constructed, and not fully linked to station systems, and is of no use to hack.</span>")
```
S34NW, line 977:
```suggestion
		to_chat(malf, "<span class='warning'>This APC was only recently constructed, and is not fully linked to station systems. Hacking it would be pointless.</span>")
```

code/modules/power/apc.dm
```diff
@@ -105,6 +105,7 @@
 	var/global/list/status_overlays_environ
 	var/indestructible = 0 // If set, prevents aliens from destroying it
 	var/keep_preset_name = 0
+	var/constructed = FALSE //Was this APC built instead of already existing? Used for malfhack to keep borgs from building apcs in space
```
SteelSlayer, line 108: Autodoc
```suggestion
	/// Was this APC built instead of already existing? Used for malfhack to keep borgs from building apcs in space
	var/constructed = FALSE
```

Sinestia:
What happens if the AI hacks an APC in a way that's not necessarily public, but hidden enough that a borg can simply deconstruct it? 

It may be more difficult to detect in the case of say, metacomming, whereas a big bunch of new APCs is plain to see.

Qwertytoforty:
> 
> 
> What happens if the AI hacks an APC in a way that's not necessarily public, but hidden enough that a borg can simply deconstruct it?
> 
> It may be more difficult to detect in the case of say, metacomming, whereas a big bunch of new APCs is plain to see.

Very good point, will think about that

The-Driker:
If we are honestly going to be this concerned about an antag type that loses 99% of the time having advantages, I'd say it might be about the time to remove Malf. It's not fun, It's objectively not balanced and is excessively leaned in favor of the crew as is. and no this isn't salt. The mode is just plain bad and the only way to fix it would be to entirely convert it to the system TG uses for it and that would be a nightmare for whoever choose to do it.

S34NW:
> 
> 
> If we are honestly going to be this concerned about an antag type that loses 99% of the time having advantages, I'd say it might be about the time to remove Malf. It's not fun, It's objectively not balanced and is excessively leaned in favor of the crew as is, and no this isn't salt. The mode is just plain bad and the only way to fix it would be to entirely convert it to the system TG uses for it and that would be a nightmare for whoever decided to do it.

I think it's less that they get advantages, and more that we have seen engi borgs cheese it by literally spamming APCs for the AI to hack that the crew can never find.

Kepplerx:
Could not an alternative be that the AI could still hack constructed APCs, but not gain points from them?

The-Driker:
> Could not an alternative be that the AI could still hack constructed APCs, but not gain points from them?

To what end though? Hacked apcs are only a thing for the sake of their points as shunting is relic code from when you could still accomplish everything. 

Note I'm not against removing the ability to hack built apcs, I just think we are far past the point of this game mode needing to be put in the ground like we did with clockwork cult when we start cutting out things that don't give nearly the level of advantage as some would believe. Especially in comparison to the vast number of counters available when the AI starts using any of it.

Spacemanspark:
We never had clockwork cult tho

Kyep (approved):

AffectedArc07 (approved):

PR #15313, opened 2021-01-17
Stunbaton Refactor

SabreML:
## What Does This PR Do

Refactors and adds documentation to the code for Stunbatons, and changes the message for hitting yourself to "You hit yourself".

I was originally planning on making a TG-style stun delay, but gave up halfway through. So I thought why not just PR the other changes I made anyway.

## Why It's Good For The Game

Makes the code easier to read.

## Changelog
:cl:
tweak: Hitting yourself with a stunbaton will now show "You hit yourself with the baton" rather than "Name hit Name with the baton".
/:cl:

PR #16196, opened 2021-06-17
Fixes secborg batons

### LINKED ISSUE #16179: Security Cybrog Stun Baton module randomly breaks
**Issue Description**:

The Cyborg's stun baton module will sometimes depower without any clear fix aside from complete module reset. Any attempts to power it again will result in "The stun baton has no power to draw from" message despite the contrary. 
**What did you expect to happen**:  It to power up

**What happened instead**:

The module states there's no power despite there being plenty. Examining the baton results in it saying it has 5% charge. The only means by which to fix it is to completely reset most of the time

Because having a module randomly break leaves the borg with only a disabler and handcuffs. Which can make it impossible for them to apprehend certain criminals such as Carp users.

Unknown. Cause seems to be random, though latest case seemed to happen on excessive use of stunbaton while powered down

At the very least the past two weeks

**Extra information**:

SabreML:
## What Does This PR Do

Fixes #16179.
This was caused by the secborg module's stunbaton only linking to the cyborg's cell on `Initialize()`. So if the borg had their power cell replaced, the baton would stay connected to the original cell and ignore the new one.

## Why It's Good For The Game

Secborgs should be able to use their stunbaton.
*(Also this was probably caused by me in #15313 so...)*

## Changelog
:cl:
fix: Fixed secborg batons staying linked to the old cell if the borg had a new one installed.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -757,7 +757,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 			//This will mean that removing and replacing a power cell will repair the mount, but I don't care at this point. ~Z
 			C.brute_damage = 0
 			C.electronics_damage = 0
-			module.update_cells(W)
+			module?.update_cells()
 			diag_hud_set_borgcell()
 
 	else if(istype(W, /obj/item/encryptionkey/) && opened)
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -757,6 +757,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 			//This will mean that removing and replacing a power cell will repair the mount, but I don't care at this point. ~Z
 			C.brute_damage = 0
 			C.electronics_damage = 0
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -757,6 +757,7 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 			//This will mean that removing and replacing a power cell will repair the mount, but I don't care at this point. ~Z
 			C.brute_damage = 0
 			C.electronics_damage = 0
+			module.update_cells(W)
```
AffectedArc07, line 760: While youre doing this, have it null out the baton call when its taken out of the borg itself. 

Failure to do so is going to hold a ref to the cell, which if this cell is destroyed before a new one is put in, thats going to cause excess GC load.

SabreML:
Coderbrain doesn't seem to be working at the moment and I've noticed a *lot* of problems in this PR, so I'll draft this until I can get around to fixing them properly.

SteelSlayer:
This one was really _bugging_ me, I thought I caused it, and couldn't figure out the reason. 🙏

AffectedArc07 (approved):

PR #16197, opened 2021-06-17
Add weapon permit security status

S34NW:
## What Does This PR Do
This PR allows anyone with armoury access on their ID to set a "weapon permit" via the security records computer.
This permit will show on sechuds, and will stop beepsky/turrets from attacking the holder for a weapon.
The status can not be set from a sechud, but it can be cleared from one if the need arises.

This is done in addition to the ID access weapon permit, assignable from the HoP. This is done to allow job roles with permits, such as sec, to not have the permit status permanently set. It also allows antags who get AA to carry on as before without giving away the fact they have AA.

## Why It's Good For The Game
Currently, weapon permits are a very rare thing. 
Granted, the situations where they are useful are rare. But, the numerous steps involved in getting a permit, including going to the HoP after getting permission and explaining it all again, make things overly complicated.
This PR will make it significantly easier, and also provide intuitive feedback to officers who might not know about a weapon permit that was issued. At the moment, the HoS or Captain typically set a "released" or "parole" status with a note, hoping any officers bother to read before tazing someone for having a weapon. Having it show up on the sechud is far more intuitive for everyone involved, and letting the HoS set the permit themselves also makes the most sense.

## Images of changes
2021-06-17 17_21_45-Window
2021-06-17 17_22_07-Window
2021-06-17 17_22_21-Window

## Changelog
:cl:
add: Added weapon permit security status.
/:cl:


AffectedArc07:
> It also allows antags who get AA to carry on as before without giving away the fact they have AA.

This giveaway never happens. If someone who is non-security has a gun out, they are being arrested regardless of whether they have the permit or not. 

Temporary permits can already be assigned with the guest pass.


The primary issue with this is that security statuses are mutually exclusive. You cannot set someone to have a weapon permit and have them be monitored at the same time for example, which could be a good use case for this.

S34NW:
> 
> 
> > It also allows antags who get AA to carry on as before without giving away the fact they have AA.
> 
> This giveaway never happens. If someone who is non-security has a gun out, they are being arrested regardless of whether they have the permit or not.

As in, this solution (as opposed to an ID based one) means that antags who get AA don't get flagged on sechuds as having a permit.

> 
> Temporary permits can already be assigned with the guest pass.

There is no guest pass terminal in the brig for the HoS to issue a weapon permit pass.

> 
> The primary issue with this is that security statuses are mutually exclusive. You cannot set someone to have a weapon permit and have them be monitored at the same time for example, which could be a good use case for this.

This is a fair point. However, if security are having to set someone's status to something else, it doesn't make much sense that they should keep their permit. Why would you let someone on a watchlist keep a weapon?

Regardless, if you think it would be better to just map in a guest pass terminal for security, I could do that post-freeze.

S34NW:
Someone suggested an ID upgrade instead, and/or a guest pass terminal. Quite honestly, I'd forgotten the first existed, and while it doesn't solve all of the issues I have with weapon permits, it solves the worst issues.

PR #16203, opened 2021-06-18
table climbing fix

### LINKED ISSUE #16199: Table bug
**Issue Description**:

Tables can be arranged with flipping tables, to make tables that can be instantly gotten up on from one side, instead of having to crawl up

To have to crawl onto a table to get on it

**What happened instead**:

People have found ways to flip tables so you can instantly get on them

Not how tables should work, theoreticaly could be used to make 1 way slowdowns. And you know, a bug

image
Make a structure like that, with a flipped table. Entering the table from the flipped side lets you instantly get on it.

hal9000PR:
## What Does This PR Do
you can no longer use flipped tables to climb onto tables instantly.
fixes #16199 
## Why It's Good For The Game

bugs bad. can be exploited aswell for making one way barriers

## Images of changes
**NOTE** tables still smooth with flipped tables. that is a bug for another day.

## Changelog
:cl:
fix: you can no longer use flipped tables to climb onto tables instantly.
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -108,27 +108,27 @@
 
 /obj/structure/table/CanPass(atom/movable/mover, turf/target, height=0)
 	if(height == 0)
-		return 1
+		return TRUE
 	if(istype(mover,/obj/item/projectile))
 		return (check_cover(mover,target))
 	if(ismob(mover))
 		var/mob/M = mover
 		if(M.flying)
-			return 1
+			return TRUE
 	if(istype(mover) && mover.checkpass(PASSTABLE))
-		return 1
+		return TRUE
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/tables_racks.dm
```diff
@@ -120,7 +120,9 @@
 	if(mover.throwing)
 		return 1
 	if(length(get_atoms_of_type(get_turf(mover), /obj/structure/table) - mover))
-		return 1
+		var/obj/structure/table/T = locate(/obj/structure/table) in get_turf(mover)
+		if(!T.flipped)
+			return 1
```
S34NW, line 125:
```suggestion
			return TRUE
```

AffectedArc07 (approved):

PR #16204, opened 2021-06-18
Prevents ablative armor from reflecting in hand

Qwertytoforty:
## What Does This PR Do

Adds a check to ablative armor to see if the user is actually wearing it, if they are not wearing it, it will not reflect.

## Why It's Good For The Game

Bugs are bad, ablative armor should have to be worn to work, not work as a discount eshield

## Changelog
:cl:
fix: Fixes ablative armor having a chance to reflect when in hand.
/:cl:

### DIFF
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -243,7 +243,7 @@
 	var/hit_reflect_chance = 40
 
 /obj/item/clothing/suit/armor/laserproof/IsReflect()
-	var/mob/living/carbon/human/user = src.loc
+	var/mob/living/carbon/human/user = loc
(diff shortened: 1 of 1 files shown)


code/modules/clothing/suits/armor.dm
```diff
@@ -243,7 +243,8 @@
 	var/hit_reflect_chance = 40
 
 /obj/item/clothing/suit/armor/laserproof/IsReflect()
-	if(prob(hit_reflect_chance))
+	var/mob/living/carbon/human/user = src.loc
```
AffectedArc07, line 246:
```suggestion
	var/mob/living/carbon/human/user = loc
```

AffectedArc07 (approved):

PR #16205, opened 2021-06-18
Prevents every gun under the sun from lying about alt skins

Qwertytoforty:
## What Does This PR Do

Sets the unique_skin variable to false by default, as 95% of guns don;t have a unique skin, and should not have this variable true.
Remove useless times when the variable was set to true when it was true
Converts the 0's and 1's of unique reskin / name to TRUE / FALSE

## Why It's Good For The Game

The game should not tell you a gun has a unique skin when it doesn't when you alt click on it.
Bugs are bad.

## Changelog
:cl:
fix: Fixes every gun saying it has a unique skin when it does not.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -40,7 +40,7 @@
 	var/randomspread = 1
 
 	var/unique_rename = TRUE //allows renaming with a pen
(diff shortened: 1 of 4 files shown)


AffectedArc07 (approved):
image

Why

PR #16212, opened 2021-06-19
Better atom rotation

AffectedArc07:
## What Does This PR Do
Makes solars and turrets rotate fully instead of being capped to 8 dirs. 

*Ignore the maptext on the turret, that has since been removed*
image
image

Yes I know this does blur the sprites a bit. I prefer the fluid rotation as opposed to pixel perfection, but everyone has a preference.

## Why It's Good For The Game
Smooth rotation

## Changelog
:cl: AffectedArc07
add: Turrets and solars now rotate the full 360 degrees instead of only 45-degree offsets.
/:cl:


AffectedArc07:
So currently, this happens
image
However I kinda like the look of this.

React on this comment of where it should be like this :+1: or like before :-1:

AffectedArc07:
image
Turrets now reset

code/__DEFINES/math.dm
```diff
@@ -236,3 +236,6 @@
 		return round(num--)
 	else
 		return num
+
+/// Calculate the angle between two atoms
+#define ATOM_ANGLE(ATOM_A, ATOM_B) (180 + arctan((ATOM_A.y - ATOM_B.y), (ATOM_A.x - ATOM_B.x)))
```
farie82, line 241: `Get_Angle` already exists which does roughly the same. It however also keeps track of the atoms pixel shifts.
Prob best to make use of that one to avoid dupe code

farie82 (changes requested):

code/__DEFINES/math.dm
AffectedArc07: Why the hell do we have two of these which have the same name and do the same thing, except one doesnt account for pixel shifts 
image

dearmochi (approved):

farie82 (approved):
Seems fine. Could you however change `Get_Angle` to `get_angle` to make it use snake_case instead of Snake_Case?

Kyep (approved):

PR #16223, opened 2021-06-21
Makes CC cryogenic freezers silent

S34NW:
## What Does This PR Do
This PR modifies CC cryogenic freezers to be silent. That is, they do not announce those who ghost in them.

## Why It's Good For The Game
CC goings-on should not be magically known by the station AI, and can cause confusion when admins are doing things or when ERT return and enter cryosleep.

## MDB broke because I tidied variables, here is what changed:
image

## Changelog
:cl:
tweak: CC cryogenic freezers no longer announce when they are used.
/:cl:

### DIFF
--- a/_maps/map_files/generic/centcomm.dmm
+++ b/_maps/map_files/generic/centcomm.dmm
@@ -25,8 +25,7 @@
 /area/syndicate_mothership)
 "ai" = (
 /obj/item/radio/intercom/syndicate{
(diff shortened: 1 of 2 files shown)


Kyep (approved):

AffectedArc07 (approved):
Im unhappy with how big this diff is due to variable sanitize, but this will do.

PR #16224, opened 2021-06-21
Prevents final summon rune by being cleaned by normal methods

Qwertytoforty:
## What Does This PR Do

Makes Narsie summon rune not pass the is_cleanable check, meaning soap, mops, janiborgs, spray cleaner, and cleaning foam can not remove the rune. Rune can still be removed with a null rod.

## Why It's Good For The Game

Cult summoning is supposed to be hard. However, it is a bit unfair to expect in order for a cult to summon, they must prevent _ANY_ cleaning grenade from going off in a 7 tile radius around the summon rune, or anyone else with a spray cleaner or a chem grenade. The rune is still removable, by the normal methods, using a null rod as a chaplain, or cult dagger as a cultist.

## Changelog
:cl:
tweak: Narsie summon rune can now only be removed by null rod or dagger, and not normal cleaning methods.
/:cl:

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -22,8 +22,6 @@
 
 #define iseffect(A) (istype(A, /obj/effect))
 
-#define is_cleanable(A) (istype(A, /obj/effect/decal/cleanable) || istype(A, /obj/effect/rune) && !istype(A, /obj/effect/rune/narsie)) //if something is cleanable
-
 #define isclothing(A) (istype(A, /obj/item/clothing))
 
 #define is_pen(W) (istype(W, /obj/item/pen))
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -22,7 +22,7 @@
 
 #define iseffect(A) (istype(A, /obj/effect))
 
-#define is_cleanable(A) (istype(A, /obj/effect/decal/cleanable) || istype(A, /obj/effect/rune)) //if something is cleanable
+#define is_cleanable(A) (istype(A, /obj/effect/decal/cleanable) || istype(A, /obj/effect/rune) && !istype(A, /obj/effect/rune/narsie)) //if something is cleanable
 
 #define isclothing(A) (istype(A, /obj/item/clothing))
 
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -363,6 +363,9 @@ a {
 /obj/proc/cult_reveal() //Called by cult reveal spell and chaplain's bible
 	return
 
+/obj/proc/is_cleanable() //Called when you want to clean something, and usualy delete it after
+	return FALSE
+
 /obj/proc/force_eject_occupant(mob/target)
 	// This proc handles safely removing occupant mobs from the object if they must be teleported out (due to being SSD/AFK, by admin teleport, etc) or transformed.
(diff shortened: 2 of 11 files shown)


code/__DEFINES/is_helpers.dm
```diff
@@ -22,7 +22,7 @@
 
 #define iseffect(A) (istype(A, /obj/effect))
 
-#define is_cleanable(A) (istype(A, /obj/effect/decal/cleanable) || istype(A, /obj/effect/rune)) //if something is cleanable
+#define is_cleanable(A) (istype(A, /obj/effect/decal/cleanable) || istype(A, /obj/effect/rune) && !istype(A, /obj/effect/rune/narsie)) //if something is cleanable
```
SabreML, line 25: A better way of doing this maybe would be to remake this into a proc rather than a define. That way it could just be overridden for anything you don't want cleaned, rather than having to add exceptions to the define every time. E.g:
```dm
/obj/proc/is_cleanable()
	return FALSE

...

/obj/effect/rune/is_cleanable()
	return TRUE

...

/obj/effect/rune/narsie/is_cleanable()
	return FALSE
```
Etc.

Qwertytoforty: Done, just forgot to mention earlier

code/game/objects/objs.dm
```diff
@@ -363,6 +363,9 @@ a {
 /obj/proc/cult_reveal() //Called by cult reveal spell and chaplain's bible
 	return
 
+/obj/proc/is_cleanable() //Called when you want to clean something, and usualy delete it after
```
dearmochi, line 366: Just pointing out that the cleaning methods filter by `/obj/effect` so anything not of that subtype won't even be considered for cleaning.

dearmochi (approved):

The-Driker:
I don't know how I feel about making it uncounterable by anything other than the one person the cult is going to 100% have killed every round before even making the rune.

Wojz:
> I don't know how I feel about making it uncounterable by anything other than the one person the cult is going to 100% have killed every round before even making the rune.

ERT can do it.

SabreML:
> ERT can do it.

That's what I was thinking too, but there is a point to be made that stopping the cult from summoning shouldn't be reliant on an ERT being called. (Assuming they kill the chaplain)

hal9000PR:
> I don't know how I feel about making it uncounterable by anything other than the one person the cult is going to 100% have killed every round before even making the rune.

Well if they have killed everything around the rune, they deserve to win. they have successfully defended that point for over 45 seconds. one janitor with a cleaning nade shouldn't stop a cult in their tracks after that. 

Also if they draw the rune then get stopped, it becomes a big point defence thing for the crew. However I suppose they could just summon elsewhere? perhaps we could limit cultists to only making one summon rune?

Varghh:
Holy cleaner when

KozackWolf:
Leaving it to one crew member to fix, just doesn't seem like a good idea...

PidgeyThePirate:
An alternative would be to make the rune cleanable like soap but take a full minute to scrub out, so the crew have to defend the rune for an equal-ish amount of time to the cult to remove it.

Kyep (approved):

code/game/objects/objs.dm
AffectedArc07: Scope this to the `/obj/effect` level please.

AffectedArc07 (changes requested):

Qwertytoforty:
Done

AffectedArc07 (approved):

PR #16227, opened 2021-06-21
Fixes soulstones kicking players to the lobby

SabreML:
## What Does This PR Do

Fixes `Runtime in soulstone.dm,98: Cannot read null.offstation_role`, by forcing the player's ghost back into their body if they're able to reenter it, since otherwise the player's `key` would be overridden by that of a polled ghost, potentially crashing both of them and kicking one to the lobby.

<details>
<summary><b>Steps to reproduce the bug:</b></summary>

1. Join the game with two accounts, one playing and one observing.
2. Kill the living one and ghost out of it with the first account.
3. Mark the corpse in VV, spawn a soulstone, and call `transfer_soul("FORCE", marked_datum, mob_reference)`
(Same as a rune sacrifice, just skips a step)
4. The body will be polled to ghosts, despite the original player still being able to re-enter.
5. Sign up for the poll on the second observer account.
6. Re-enter the corpse on the first account.
7. Wait a few seconds, and both clients crash.
8. Rejoin the server and see that one client is inside the corpse, and one is kicked out to the lobby.

</details>

## Why It's Good For The Game

Semi-major bugfix.

## Changelog
:cl:
fix: Fixed soulstones sending players to the lobby in some circumstances.
/:cl:

### DIFF
--- a/code/game/gamemodes/wizard/soulstone.dm
+++ b/code/game/gamemodes/wizard/soulstone.dm
@@ -262,7 +262,8 @@
 	switch(choice)
 		if("FORCE")
 			var/mob/living/T = target
-			if(T.client && T.ghost_can_reenter()) // Haven't DC'd or ahudded
+			T.grab_ghost(FALSE) // If they haven't DC'd or ahudded, put them back in their body
+			if(T.client) // If there's someone in the body
 				init_shade(T, user)
 			else // Poll ghosts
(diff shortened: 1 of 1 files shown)


SabreML:
whoops

farie82 (approved):

dearmochi (approved):

AffectedArc07 (approved):

PR #16241, opened 2021-06-23
tweaks some things with energy bolas

hal9000PR:
## What Does This PR Do
E-bolas behave more like standard bolas. They no longer magically pass through shielded hardsuits/riot shield, they can now be caught and they now actually play their hitsound.

Hitting someone with a bola who is already hit with a bola no longer creates an e-beartrap at their feet

Also removed e-beartraps because they are unused and they cannot even be used as they delete themselves on the floor.

## Why It's Good For The Game
shields should block things, intended sounds should be played.

also `new`ing a beartrap and forcing them to walk over it is also kinda gross, it means that the breakout time on the actual item didn't actually affect anything resulting in confusion. 

## Changelog
:cl:
tweak: energy bolas can now be caught and blocked by shields.
fix: energy bolas now play their sound on hit.
fix: hitting someone with an e-bola who is already snared no longer drops an e-beartrap at their feet.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -883,8 +883,6 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, list(/obj/machinery/atmospherics/unary/ven
 				legcuffed = null
 				toggle_move_intent()
 				update_inv_legcuffed()
-				if(istype(I, /obj/item/restraints/legcuffs/bola/energy))
-					qdel(I)
 				return
 			else
 				unEquip(I)
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -883,6 +883,8 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, list(/obj/machinery/atmospherics/unary/ven
 				legcuffed = null
 				toggle_move_intent()
 				update_inv_legcuffed()
+				if(istype(I, /obj/item/restraints/legcuffs/bola/energy))
+					qdel(I)
 				return
 			else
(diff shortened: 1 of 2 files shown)


Varghh:
E-bolas are one time use, I personally dislike the catch change, fine with  everything else

S34NW:
`they can now be caught`
Hmmmmmmm, isn't that one of the big points of this one, vs say a regular bola?

hal9000PR:
> `they can now be caught`
> Hmmmmmmm, isn't that one of the big points of this one, vs say a regular bola?

well- they fit in pockets and they are harder to remove over a regular bola.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -883,6 +883,8 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, list(/obj/machinery/atmospherics/unary/ven
 				legcuffed = null
 				toggle_move_intent()
 				update_inv_legcuffed()
+				if(istype(I, /obj/item/restraints/legcuffs/bola/energy))
```
dearmochi, line 886: Not a super fan of this snowflaking, you may want to consider moving the `qdel` in the energy bola's `dropped` proc (only when it has hit someone)

dearmochi (changes requested):

code/modules/mob/living/carbon/carbon.dm
hal9000PR: Will look into it

hal9000PR: hopefully this is better?

hal9000PR:
I really don't care about these changes all that much myself so I am gonna go ahead and close this.

PR #16247, opened 2021-06-24
Fixes singulo hammer sucking in ghosts

hal9000PR:
## What Does This PR Do
Ghosts are no longer sucked in by the singulo hammer

## Why It's Good For The Game
bug b gone

## Changelog
:cl:
fix: ghosts are no longer pulled in by the singulo hammer
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -670,7 +670,7 @@
 
 /obj/item/twohanded/singularityhammer/proc/vortex(turf/pull, mob/wielder)
 	for(var/atom/movable/X in orange(5, pull))
-		if(isobserver(X))
+		if(X.move_resist == INFINITY)
 			continue
 		if(X == wielder)
 			continue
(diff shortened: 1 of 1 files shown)


dearmochi (approved):

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -670,6 +670,8 @@
 
 /obj/item/twohanded/singularityhammer/proc/vortex(turf/pull, mob/wielder)
 	for(var/atom/movable/X in orange(5, pull))
+		if(isobserver(X))
+			continue
```
SteelSlayer, line 674: While this does work, we should check `move_resist` instead. So something like `if(X.move_resist == INFINITY)` would be better and can cover other potential cases too, where it's pulling stuff we don't want it to.

hal9000PR: smart, will do.

SteelSlayer (approved):

AffectedArc07 (approved):

PR #16255, opened 2021-06-25
Fixes powersink not giving you a beacon on purchase.

### LINKED ISSUE #16254: Power sink does not give you a teleporter to call it in
**Issue Description**:

The powersink description in the uplink states you get a beacon to teleport the sink to later on.
`Ordering this sends you a small beacon that will teleport the power sink to your location on activation`
Unfortunately you are just given the powersink immediately.

You think you can buy it now to use later, but get handed a huge item that cant be stored instead.

S34NW:
## What Does This PR Do
Fixes #16254
Gives you a teleporter beacon to call in the powersink later, as opposed to just giving you the item.

## Why It's Good For The Game
Things should work as described, getting a powersink when you aren't ready for it because you are expecting a beacon is bad.

## Changelog
:cl:
fix: Uplink powersink purchase now properly gives you a beacon instead of the actual item.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/beacon.dm
+++ b/code/game/objects/items/devices/radio/beacon.dm
@@ -89,6 +89,10 @@
 	desc = "A label on it reads: <i>Warning: Activating this device will send a high-ordinance EMP explosive to your location</i>."
 	bomb = /obj/machinery/syndicatebomb/emp
 
+/obj/item/radio/beacon/syndicate/bomb/power_sink //Yes it's not a bomb but this code works the way we want
+	desc = "A label on it reads: <i>Warning: Activating this device will send a power sink to your location</i>."
+	bomb = /obj/item/powersink
+
 /obj/item/radio/beacon/engine
 	desc = "A label on it reads: <i>Warning: This device is used for transportation of high-density objects used for high-yield power generation. Stay away!</i>."
 	anchored = 1		//Let's not move these around. Some folk might get the idea to use these for assassinations
--- a/code/game/objects/items/devices/radio/beacon.dm
+++ b/code/game/objects/items/devices/radio/beacon.dm
@@ -74,12 +74,11 @@
 /obj/item/radio/beacon/syndicate/power_sink
 	name = "suspicious beacon"
 	desc = "A label on it reads: <i>Warning: Activating this device will send a power sink to your location</i>."
-	var/item_to_teleport = /obj/item/powersink
 
 /obj/item/radio/beacon/syndicate/power_sink/attack_self(mob/user)
 	if(user)
 		to_chat(user, "<span class='notice'>Locked In</span>")
-		new item_to_teleport(user.loc)
+		new /obj/item/powersink(user.loc)
 		playsound(src, 'sound/effects/pop.ogg', 100, TRUE, 1)
 		user.drop_item()
 		qdel(src)
--- a/code/game/objects/items/devices/radio/beacon.dm
+++ b/code/game/objects/items/devices/radio/beacon.dm
@@ -71,6 +71,19 @@
 		user.drop_item()
 		qdel(src)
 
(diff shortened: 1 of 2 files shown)


farie82 (approved):

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -89,6 +89,10 @@
 	desc = "A label on it reads: <i>Warning: Activating this device will send a high-ordinance EMP explosive to your location</i>."
 	bomb = /obj/machinery/syndicatebomb/emp
 
+/obj/item/radio/beacon/syndicate/bomb/power_sink //Yes it's not a bomb but this code works the way we want
```
AffectedArc07, line 92: This feels so wrong. I would rather a version with `/obj/item/radio/beacon/powersink` with a separate attackhand

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -71,6 +71,19 @@
 		user.drop_item()
 		qdel(src)
 
+/obj/item/radio/beacon/syndicate/power_sink
+	name = "suspicious beacon"
+	desc = "A label on it reads: <i>Warning: Activating this device will send a power sink to your location</i>."
+	var/item_to_teleport = /obj/item/powersink
+
+/obj/item/radio/beacon/syndicate/power_sink/attack_self(mob/user)
```
farie82, line 79: Since the change is made.
I think it would be better to have a spawning beacon type which does this code. But spawns a given object. Reduces the copypasta as well since now this same proc is made 3 times. 
By `/obj/item/radio/beacon/syndicate/attack_self`, `/obj/item/radio/beacon/syndicate/power_sink/attack_self` and `/obj/item/radio/beacon/syndicate/bomb/attack_self`
It would probably be better to make `/obj/item/radio/beacon/syndicate` an abstract generic class from which the others inherit.

farie82 (changes requested):

code/game/objects/items/devices/radio/beacon.dm
AffectedArc07: That can be done in another PR

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -71,6 +71,19 @@
...
+	if(user)
+		to_chat(user, "<span class='notice'>Locked In</span>")
+		new item_to_teleport(user.loc)
+		playsound(src, 'sound/effects/pop.ogg', 100, 1, 1)
```
dearmochi, line 83:
```suggestion
		playsound(src, 'sound/effects/pop.ogg', 100, TRUE, 1)
```

dearmochi (approved):

code/game/objects/items/devices/radio/beacon.dm
```diff
@@ -71,6 +71,19 @@
...
+/obj/item/radio/beacon/syndicate/power_sink
+	name = "suspicious beacon"
+	desc = "A label on it reads: <i>Warning: Activating this device will send a power sink to your location</i>."
+	var/item_to_teleport = /obj/item/powersink
+
+/obj/item/radio/beacon/syndicate/power_sink/attack_self(mob/user)
+	if(user)
+		to_chat(user, "<span class='notice'>Locked In</span>")
+		new item_to_teleport(user.loc)
+		playsound(src, 'sound/effects/pop.ogg', 100, TRUE, 1)
```
AffectedArc07, line 83:
```suggestion
/obj/item/radio/beacon/syndicate/power_sink/attack_self(mob/user)
	if(user)
		to_chat(user, "<span class='notice'>Locked In</span>")
		new /obj/item/powersink(user.loc)
		playsound(src, 'sound/effects/pop.ogg', 100, TRUE, 1)
```

AffectedArc07 (approved):

PR #16256, opened 2021-06-25
Enable Radiation HUD for ghosts

chafla:
## What Does This PR Do

While I'm on a roll I might as well try another PR.

~~Adds a ghost verb, "Toggle Radiation View",~~ Adds a new HUD to the new Ghost Hud verb that allows ghosts to see radiation levels like radiation scanner mesons. Also ~~slightly refactors the original proc for making the overlay by moving it to the mob instead of just the scanner goggles~~ updates the radiation-checking proc by adding logic to the radiation subsystem to keep track of per-tile radiation.

## Why It's Good For The Game

It's already hard enough to figure out how much radiation something has without a geiger counter, which ghosts of course don't have access to. It's also a nice bonus for ghosts to be able to see just how messed up engineers coming out of the Supermatter are.
This also makes radiation views from Mesons significantly more performant, in a way that shouldn't scale nearly as hard based on players using it.

## Images of changes

img
img

## Changelog
:cl:
add: Radiation HUD toggle in Ghost HUDs that allows seeing radiation levels.
/:cl:


Kyep (approved):
While I am okay with this, I really feel like at some point all the ghost HUDs should just be refactored into one "HUDs Panel" with a set of toggles.... there are too many HUD toggle verbs for ghosts.

SteelSlayer:
> While I am okay with this, I really feel like at some point all the ghost HUDs should just be refactored into one "HUDs Panel" with a set of toggles.... there are too many HUD toggle verbs for ghosts.

I might have to steal this idea.

code/modules/clothing/glasses/engine_goggles.dm
```diff
@@ -66,8 +66,8 @@
 		if(MODE_RAD)
 			show_rads()
 
-/obj/item/clothing/glasses/meson/engine/proc/show_rads()
-	var/mob/living/carbon/human/user = loc
+/mob/proc/show_rads(range)
+	var/mob/user = src
```
SteelSlayer, line 70: This line should be removed and the instances of `user` should be replaced with `src`, except line 92, where `user.client` would just become `client` because implicit src.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -97,13 +100,19 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
...
 
 /mob/dead/observer/examine(mob/user)
 	. = ..()
 	if(!invisibility)
 		. += "It seems extremely obvious."
 
+/mob/dead/observer/process()
+	if(seerads)
+		show_rads(30)
+	return
```
SteelSlayer, line 114: Unnecessary return.
```suggestion
```

code/modules/clothing/glasses/engine_goggles.dm
```diff
@@ -79,7 +79,7 @@
 
 	for(var/i in rad_places)
 		var/turf/place = i
-		if(get_dist(user, place) >= range * 5)	//Rads are easier to see than wires under the floor
+		if(range != -1 && get_dist(user, place) >= range)
```
SteelSlayer, line 82: What's the reason for the `range != -1` check here?

chafla: It's a special "ignore any range checks" value, essentially. I can't really think of a good reason it'd be necessary to have it larger than the screen size I suppose, and if there's a way to pull that var from the client it'd probably be a better option.

code/modules/clothing/glasses/engine_goggles.dm
chafla: Fixed in c9e796d

code/modules/mob/dead/observer/observer.dm
chafla: Fixed in c9e796d

code/modules/clothing/glasses/engine_goggles.dm
```diff
@@ -66,8 +66,7 @@
 		if(MODE_RAD)
 			show_rads()
 
-/obj/item/clothing/glasses/meson/engine/proc/show_rads()
-	var/mob/living/carbon/human/user = loc
+/mob/proc/show_rads(range)
```
dearmochi, line 69: Not the best location for a `mob` level proc. Perhaps `mob.dm`?

code/modules/mob/dead/observer/observer.dm
```diff
@@ -87,6 +88,8 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 
 	//starts ghosts off with all HUDs.
 	toggle_medHUD()
+
+	START_PROCESSING(SSobj, src)
```
dearmochi, line 92: This starts the processing of all observers, regardless of `seerads` being true or not. I would move the toggles inside the verb.

dearmochi (changes requested):

code/modules/clothing/glasses/engine_goggles.dm
SteelSlayer: I don't understand, because you're not doing `show_rads(-1)` anywhere to warrant including this. In the event that you do need this though, it should probably be placed before the loop, that way you don't have to check if range is -1 for every entry in the list.

AffectedArc07:
I have performance concerns about this, and would like to see profiler data.

code/modules/clothing/glasses/engine_goggles.dm
chafla: Yeah, I was mostly including it in case it might be useful down the line. I don't really see a use for it (since it'd probably not be worth looping over every single item when we have a range implemented) so I'll just remove it.

code/modules/mob/dead/observer/observer.dm
chafla: Fixed in 000b89a

code/modules/clothing/glasses/engine_goggles.dm
chafla: Addressed in 000b89a

code/modules/clothing/glasses/engine_goggles.dm
chafla: Removed the check in 000b89a

chafla:
> I have performance concerns about this, and would like to see profiler data.

Regarding performance issues, the only real difference should be that ghosts can use it now. I'm pretty green, so I'm not sure how performance heavy that latter part might be, I just felt like it would make sense since it was already on that loop for mesons.

I'll try running the profiler and see if anything notable comes out of that.

chafla:
Delammed a supermatter in medbay and engineering (don't really know a better way to dump rads into things easily).


Timeframe | Self CPU | Total CPU | Real Time | Overtime | Calls
ACTIVE | 0.180 | 0.498 | 0.487 | 0.000 | 752
AVG | 0.000 | 0.001 | 0.001 | 0.000 | 752

SteelSlayer:
Given those numbers, performance *could* be quite concerning, depending on the population levels, which are usually pretty high. It doesn't help that radiation component related code is already a tad expensive.

chafla:
Right, I can see that it might not scale well given how it's currently implemented per-mob. I guess it's "fine" for now because of how unlikely it is that more than four or five people (like engineers) will have it equipped in general?

I was wondering if it might make sense to have some sort of general proc that updates all ghosts a single time, rather than scanning over each radiated object for every observer. Maybe something that sets a value on the associated turf? I didn't really want to do that since it felt snowflakey (and I'm unfamiliar enough with the codebase as it is).

Anyway, I'm going to try tweaking the range parameter and see how that changes things. I think I left it obnoxiously high (30!) and I'll see if dropping it down to something more reasonable (maybe 5-10) would improve performance.

chafla:
Setting range to 5 yielded some pretty different results. After following a level 5 singularity around for a while, the CPU seems to sit at around

Timeframe | Self CPU | Total CPU | Real Time | Overtime | Calls
 -----------|-----------|------------|----------|----|----
ACTIVE | 0.225 | 0.264 | 0.259 | 0.000 | 219
AVG | 0.002 | 0.002 | 0.002 | 0.000 | 219

After making it a level 6 singularity, that seemed to skyrocket, though most of the irradiated items should have been deleted by being eaten by the singulo. Are we sure that radioactive items are being removed from the processing queue on deletion?

SteelSlayer:
>I was wondering if it might make sense to have some sort of general proc that updates all ghosts a single time, rather than scanning over each radiated object for every observer. Maybe something that sets a value on the associated turf?

Yeah I had an idea where we could store the radiation value on the tile somehow. Either just the number or create an image and then render it to anyone in range who can see radiation. I'm not sure how hard or feasible this is, just brainstorming.

>Setting range to 5 yielded some pretty different results. After following a level 5 singularity around for a while, the CPU seems to sit at around

Given that the number of calls is lower than your previous test, It seems that the CPU numbers will be higher if you let them run up to ~752 calls. Which is a bit odd considering you said that you reduced the range.

> Are we sure that radioactive items are being removed from the processing queue on deletion?

They *should* be, given the code here:

Code_PKCXsDq4dT

chafla:
In my tests, it seems like items that get absorbed by the singularity don't necessarily lose their radiation and may stay in the processing queue even after supposedly being eaten? At least from what I could see, qdel just didn't seem to be getting called? I'm not really sure what's going on there but it's out of scope for this PR, at least.

Anyway, the way that the code currently iterates over everything in the radiation subsystem rather than every tile seems like a nice jumping-off point. If there's a good way to call the `show_rads()` proc once-per-tick globally and just update a value on every tile, I'm thinking that might work out the best? The only question in my mind beyond that is where to call it, and where to store the value itself (would it be valid to just add another var to a turf?). I'm personally feeling I'm a little too green to know what would be the best course of action to tack this functionality on somewhere, but I'm still trying to think about it.

chafla:
Alright, time for take 2

I've added a new datum, `/datum/radiation_cache` that's pretty much a wrapper around two lists; one representing the tile state during the current SS tick and another representing the previous completed SS tick. Each radioactive component adds their radioactivity to the current tick state, which during the next SS tick becomes the previous tick state, and is then queried by whoever needs it I guess. 

The new rad hud code just grabs the nearest tiles in a range and checks against this list of radioactive tiles, rather than iterating over every single radioactive object in existence. This seems to be a little better for performance.

Proc | Timeframe | Self CPU | Total CPU | Real Time | Overtime | Calls
 -----|-----------|-----------|------------|----------|----|----
/datum/controller/subsystem/processing/radiation/proc/get_turf_radiation | ACTIVE | 0.034 | 0.056 | 0.051 | 0.000 | 42471
/datum/controller/subsystem/processing/radiation/proc/update_rad_cache | ACTIVE | 0.007 | 0.021 | 0.016 | 0.000 | 10784

AffectedArc07:
Could you refactor this into the new HUD manager panel?

SteelSlayer:
@chafla I can help you implement this onto the ghost HUD manager if you want. Or If you don't feel like it, and it's acceptable to maints, I can also do this myself in a follow-up PR, I don't mind.

chafla:
Should be all set

image

code/controllers/subsystem/radiation.dm
```diff
@@ -16,3 +21,46 @@ PROCESSING_SUBSYSTEM_DEF(radiation)
...
+/datum/controller/subsystem/processing/radiation/proc/update_rad_cache(datum/component/radioactive/thing)
+	return turf_cache.update_rad_cache(thing)
+/datum/radiation_cache
+	// Cache radiation levels for each turf so it doesn't need to be done iteratively
+	// turf_rad_cache is the state in the current loop, and may not be 100% representative
+	// until processing is complete.
+	// prev_rad_cache is the fully evaluated radiation state cache from the previous tick.
+	var/has_refreshed_cache = FALSE
+	var/list/turf_rad_cache = list()
+	var/list/prev_rad_cache = list()
```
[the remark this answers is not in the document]
chafla: Yeah. `has_refreshed_cache` is vestigial and can be removed.

The reason I made a new datum was so I could have the lists be updated in the radiation processing loop, since the cur-tick cache needs to be dumped into the prev rad cache and cleared every time radioactive objects update. I was thinking of adding something to `fire()` but it was awkward to get something that'd only fire once per processing loop. I similarly figured it'd be cleaner to have the lists contained with the logic that's used to update them.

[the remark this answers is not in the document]
chafla: Removed the extra var in 14cde58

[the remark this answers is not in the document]
SteelSlayer: > I was thinking of adding something to fire() but it was awkward to get something that'd only fire once per processing loop

I'm not really sure what you mean here. You can put just do:
```dm
/datum/controller/subsystem/processing/radiation/fire(resumed)
	. = ..()
	refresh_rad_cache()
```
Which is basically identical to doing:
```dm
/datum/radiation_cache/process()
	refresh_rad_cache()
```
And the old `radiation_cache` procs can be converted (already mentioned that though). With the way it's set up right now, there's this sort of middle man that doesn't need to be there. Unless I'm horribly blind and missing something.. the `radiation_cache` datum can instead stuck onto SSradiation.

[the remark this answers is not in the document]
chafla: In my testing with the debugger, `fire()` would be called once for every item in the processing queue. I tried doing that before, but it just meant that the rad cache would be cleared *on every item,* not just every tick of the radiation queue.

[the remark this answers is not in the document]
SteelSlayer: That.. should not be the case. The `fire()` proc is called only once every `wait` period.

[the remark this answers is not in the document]
chafla: That's what I saw when testing it before, but with a second look you're totally right. I'll see about refactoring out the datum.

chafla:
Should be able to clear the tgui merge conflict if this gets the go-ahead, I'm still keeping an eye on it

variableundefined:
@chafla Deconflict and DM me on discord for merge.

variableundefined (approved):

PR #16274, opened 2021-06-28
Prevents ghosts from setting off mousetraps

Qwertytoforty:
## What Does This PR Do

Adds an isobserver check to inventory code, that prevents each item trying isfound against the ghost when a storage container is open.

## Why It's Good For The Game

Ghosts should not be able to remotely set off moustraps / facehuggers / mousetrap bombs.

## Changelog
:cl:
fix: Ghosts can no longer set off mousetrap assemblies and facehuggers in storage containers.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -164,9 +164,10 @@
 	if(!user.client)
 		return
 	if(user.s_active != src) // Switching from another container
-		for(var/obj/item/I in src)
-			if(I.on_found(user)) // For mouse traps and such
-				return // If something triggered, don't open the UI
+		if(!isobserver(user)) //Don't want ghosts setting off the fun stuff
+			for(var/obj/item/I in src)
+				if(I.on_found(user)) // For mouse traps and such
+					return // If something triggered, don't open the UI
 	orient2hud(user) // this only needs to happen to make .contents show properly as screen objects.
 	if(user.s_active)
 		user.s_active.hide_from(user) // If there's already an interface open, close it.
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -164,10 +164,9 @@
 	if(!user.client)
 		return
 	if(user.s_active != src && !isobserver(user))
-		if(!isobserver(user)) //Don't want ghosts setting off the fun stuff
-			for(var/obj/item/I in src)
-				if(I.on_found(user)) // For mouse traps and such
-					return // If something triggered, don't open the UI
+		for(var/obj/item/I in src)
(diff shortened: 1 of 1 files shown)


Kyep (approved):
Oh snap. I like it.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -164,9 +164,10 @@
 	if(!user.client)
 		return
 	if(user.s_active != src) // Switching from another container
-		for(var/obj/item/I in src)
-			if(I.on_found(user)) // For mouse traps and such
-				return // If something triggered, don't open the UI
+		if(!isobserver(user)) //Don't want ghosts setting off the fun stuff
```
SteelSlayer, line 167: This check should really be put into `/obj/item/assembly/mousetrap/on_found`. Return from the proc if they're an observer.

Qwertytoforty: I did not put it there, in the event other on_founds are added, and due to facehuggers also using the same proc.

SteelSlayer: Ugh alright, I don't like how this needs to be snowflaked, but I don't have a better idea right now.

S34NW:
I ded pls nerf?

But yeah, this was a weird issue thanks for giving it attention.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -164,9 +164,10 @@
 	if(!user.client)
 		return
 	if(user.s_active != src) // Switching from another container
```
dearmochi, line 166: Reduce nesting by doing

```suggestion
	if(user.s_active != src && !isobserver(user))
```

dearmochi (changes requested):

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -163,10 +163,11 @@
...
 		return
-	if(user.s_active != src) // Switching from another container
-		for(var/obj/item/I in src)
-			if(I.on_found(user)) // For mouse traps and such
-				return // If something triggered, don't open the UI
+	if(user.s_active != src && !isobserver(user))
+		if(!isobserver(user)) //Don't want ghosts setting off the fun stuff
+			for(var/obj/item/I in src)
+				if(I.on_found(user)) // For mouse traps and such
+					return // If something triggered, don't open the UI
```
dearmochi, line 170:
```suggestion
		for(var/obj/item/I in src)
			if(I.on_found(user)) // For mouse traps and such
				return // If something triggered, don't open the UI
```

dearmochi (changes requested):

dearmochi (approved):

AffectedArc07 (approved):

PR #16275, opened 2021-06-29
Deathsquad Update

warior4356:
Note for maintainers: While this PR does edit a map file, it does not make any map changes. It is only removing a deprecated item and adjusting a type path for another. Specifically making the existing softsuits into hardsuits and removing the deathsquad suit helmets, as the suits have built-in helmets now. 

## What Does This PR Do
It's been far too long since anyone gave deathsquad any love and it shows. At this point, gamma sec ERT has better equipment than deathsquad in many ways. This PR does the following:

- Makes Deathsquad softsuits into hardsuits with a built-in jetpack, flashlight, Geiger counter.
- Removed their flashlight.
- Replaces their single piece of X4 with a box of X4.
- Gave them implants (hardened anti stun, hardened revive, hardened anti sleep, nutriment plus, sec hud, anti-drop, and basic combat (slightly worse esword, med beam, flash, stun baton)).
- Removed their esword.
- Moved their pinpointer into their PDA slot.
- Gave standard troopers a black jumpsuit with sensors on binary. Officers have a nanotrasen uniform with binary sensors.
- Moved their sidearm into their backpack and gave them an assault belt.
- Swapped the empty box in their bag with an ERT survival kit.
- Added a single BSA targeting beacon to their pack. 10-second fuse, counts out loud, replicates a BSA strike.
- Moved their rifle to their left hand to prevent friendly fire in spawn.
- Fixed adminordrazine not properly solving a stopped heart for new crit, made it stop bleeding, added blood restoration, made it heal dead mobs and revive them when mostly healed like SR would, sans gibbing.
- Made the combat nanites injector default to 10 units instead of 15 and have 5, 10, and 15 as options.
- Refactors DS to use the outfit system instead of snowflake equip statements.
- Changes the beret from the multisword officer look from the SOO beret to the regular one.

## Why It's Good For The Game

- The fact that deathsquad, the most technologically advanced force in almost any round were still using two piece soft suits was a travesty. Giving them a proper hardsuit lets us remove their flashlight too. Also giving them a jetpack helps for when the station inevitably has holes in it.
- X4 is a useful door breaching charge, but only getting one is kinda lame.
- Similar to above, gamma sec ERT has all of the implants I listed above with the exceptions of having a telebaton instead of a basic combat implant and not having anti drop or anti sleep. Deathsqaud should be stronger than gamma sec ERT. Also, the basic combat implant was unused, which is unfortunate since it's a really neat item and this seemed like a great place to put it into the game.
- They don't need an esword if their combat implant gives them one.
- Their pinpointer is very useful and so is having it in a high visibility location.
- Similarly, the bright green jumpsuits were kinda odd. Also they had random sensors. Forcing binary sensors is the perfect amount of dread for the crew I think. They know the deathsquad is there, and alive, but nothing more, besides the fact they are going to die.
- The fact they lacked a belt was also odd, and an assault belt gives them more options for storing their stuff.
- Why did they have an empty box in the first place? An ERT survival box makes more sense.
- The BSA beacon was mostly added to give them a grenade that would be on par with their normal firepower. Regular grenades are pathetic compared to their pulse rifle, but having the ability to drop a BSA strike on whatever target they want is a powerful and thematic ability for an NT kill team. It beeps, counts down out loud, and has a ten-second delay to make it very difficult to use against people, functioning more as an anti-fortification weapon to clear out things like terror nests.
- Moving their gun out of their right hand is just common sense.
- First off, the fact that new crit could not be cured by adminordrazine was dumb. That on it's own was just a bug fix. One of the few problems it couldn't solve was blood loss, so I made it stop bleeding and slowly replace missing blood so it can be used to save people who are dying. Speaking of dying, there are few things as anti-climactic as someone suicide bombing deathsquad and killing half of them, or getting chain stunned and eaten by terrors. This way, if their body can be recovered they can be brought back with adminordrazine. The drug was changed to function on the dead and the living, so it will slowly put their broken body back together, then when they reach -100 health it brings them back to life in hard crit. For reference if a human in deathsquad armor takes a max cap to the face, it takes around 30 units of adminordrazine to bring them back to life and mostly heal them, taking nearly a minute, so this is not a fast nor cheap process. If you need further justification on why should adminordrazine be able to bring back the dead? Nanomachines son.
- With light injuries or status conditions having to use 15/100 units feels a bit wasteful, so I moved the default to 10 units and gave them some options on how much they feel needs to be injected.
- Snowflake code is bad, standardization is good.
- This was a weird niche case and it was odd that a duplicate got such a powerful hat. Also, the SOO beret only exists as part of the coat now, as it is a subtype of the deathsquad armor.

## Images of changes
dreamseeker_2021-06-29_00-21-49

https://user-images.githubusercontent.com/6993506/123754786-1b3f4480-d870-11eb-9240-15411c0ce01c.mp4

https://user-images.githubusercontent.com/6993506/123754794-1ed2cb80-d870-11eb-8fb2-704eb1a61763.mp4

https://user-images.githubusercontent.com/6993506/123891374-ade1f100-d90d-11eb-866e-812e6f95fa36.mp4

## Changelog
:cl:
add: Added BSA targeting beacon.
add: Added T4 to deathsquad.
add: Added assault belt to deathsquad
add: Added PDAs to deathsquad.
add: Added ERT survival kit to deathsquad.
add: Added implants to deathsquad (hardened anti stun, hardened reviver, nutriment plus, sec hud, anti-drop, and basic combat)
add: Added adminordrazine restoring blood, fixing the dead, reviving the dead.
del: Removed flashlight from deathsquad.
del: Removed esword from deathsquad.
del: Removed X4 from deathsquad.
tweak: Made deathsquad softsuits into hardsuits.
tweak: Gave deathsquad black jumpsuits, forced suit sensors to binary.
tweak: Moved deathsquad rifles from their right hand to their left.
fix: Fixed adminordrazine not working on new crit.
imageadd: Added new helmet sprite to deathsquad for hardsuit lights.
imageadd: Added new sprite for BSA beacon.
/:cl:


hal9000PR:
Don't give them a pda. It is bloat that they dont need, if a newer player spawns and the pinpointer is in their bag instead of in their pda slot they are less likely to use it. They don't need a pda.

The BSA beacon will deal more harm to deathsquads than good. Friendly fire at and loosing limbs. I really don't think this is a good idea.

Deathsquads really shouldn't have suit sensors at all, cos of the you know, no witnesses thing. Broadcasting that they are here on suit sensors is kinda silly.

Geiger counter is also just Action button bloat that isn't needed, same with the helmet lights cos they have full night vision and are supposed to not be seen.

Also that changelog is really long and could just be condensed to `tweak: buffed and tweaked deathsquads` and then the feature ones

Aquilar:
if you wanna had a bit more crowd control immunity consider also adding the hardened neural jumpstarter as well.

shmabuu:
> if you wanna add a bit more crowd control immunity consider also adding the hardened neural jumpstarter as well.

Neural jumpstarter and CNS aren't compatible together and CNS is significantly more useful than the neural jumpstarter

Wojz:
> Neural jumpstarter and CNS aren't compatible together and CNS is significantly more useful than the neural jumpstarter

There's an admin subtype of the jumpstarter that is compatible, and I even think it's quite literally called "compatible"

S34NW:
I agree with the comments about PDA and BSA grenade, those two are going to cause confusion. DS will NEVER need a PDA, and burying the pinpointer in the bag will just complicate things.

BSA grenade is going to cause so much friendly fire, if it weren't for the shield most DS would be friendly fired. The gun they get is more than capable on its own.

hal9000PR:
They don't need T4 when their guns delete walls instantly....

Kyep (approved):

Dumbdumn5:
I could go either way on the PDA since I do love some meme potential, but having it removed for easy pinpointer slottage is fine.

I like the new BSA beacon sprite and the X4 instead of T4 is fine.

I'm impartial to the suit sensors, don't quite need them, but it doesn't hurt to have considering DS aren't so much stealth as in 'sneak aboard and never get seen' as they are 'go in and kill anything you see or that sees you period' stealth. 

I _love_ the idea of the BSA grenade though I also understand there'll be a few issues with friendly-fire, though given we get enough collateral with other nades, I'm not sure how different this is from other high power grenades on other roles.

:+1: overall

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -711,3 +711,59 @@
...
+	jetpack = /obj/item/tank/jetpack/suit
+
+	sprite_sheets = list()
+	sprite_sheets_obj = list()
+
+/obj/item/clothing/head/helmet/space/hardsuit/deathsquad/beret
+	name = "officer beret"
+	desc = "An armored beret commonly used by special operations officers."
+	icon_state = "beret_officer"
+	armor = list("melee" = 80, "bullet" = 80, "laser" = 50, "energy" = 50, "bomb" = 100, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 100)
```
S34NW, line 754:
```suggestion
```
It inherits this from the regular DS helmet, Shouldn't need to re-state it here.

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -711,3 +711,59 @@
...
+	flags =  STOPSPRESSUREDMAGE | THICKMATERIAL
```
S34NW, line 755:
```suggestion
```
Are these flags needed? As far as the game is concerned, its a hardsuit helmet. Thus you should not need to specify these things.

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -711,3 +711,59 @@
...
+
+/obj/item/clothing/suit/space/hardsuit/deathsquad/officer
+	name = "officer jacket"
+	desc = "An armored jacket used in special operations."
+	icon_state = "detective"
+	item_state = "det_suit"
+	blood_overlay_type = "coat"
+	flags_inv = 0
+	slowdown = 0
+	armor = list("melee" = 80, "bullet" = 80, "laser" = 50, "energy" = 50, "bomb" = 100, "bio" = 100, "rad" = 100, "fire" = 100, "acid" = 100)
```
S34NW, line 766:
```suggestion
```
Redundant

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -711,3 +711,59 @@
...
+	resistance_flags = FIRE_PROOF | ACID_PROOF
```
S34NW, line 767:
```suggestion
```
Redundant

code/modules/clothing/spacesuits/hardsuit.dm
```diff
@@ -711,3 +711,59 @@
...
+	strip_delay = 130
+	dog_fashion = /datum/dog_fashion/back/deathsquad
+	helmettype = /obj/item/clothing/head/helmet/space/hardsuit/deathsquad
+	jetpack = /obj/item/tank/jetpack/suit
+
+	sprite_sheets = list()
+	sprite_sheets_obj = list()
+
+/obj/item/clothing/head/helmet/space/hardsuit/deathsquad/beret
```
S34NW, line 750: By making this a hardsuit helmet, you will run into issues with it needing an icon for toggling the light. Has it been tested?

S34NW (changes requested):
Some tweaks to the DS suits now they have been ported to hardsuits.

code/modules/clothing/spacesuits/hardsuit.dm
warior4356: I removed the light and Gieger counter actions from this hat, if you notice the actions_types = list()

code/modules/clothing/spacesuits/hardsuit.dm
warior4356: Those are needed, because the default helmet flag of BLOCKHAIR needs to be "removed"

xProlithium:
If there are two things I've thought about trying to change to the DS hardsuit, they would be 
1. Having an integrated eshield setup similar to the Cyborg.
2. I get serious CTAC from Altered Carbon vibes when I think of DS. Having red lights on the front of the helmet - and nothing else would be terrifying to see in a dark hallway.

https://i.imgur.com/FyUN4Ov.png

But I dunno where to start with spriting so just pitching my idea to an amazing DS PR.

Qwertytoforty:
I like this PR, and I am tired of DS dying like idiots, however I do feel like maybe antidrop is a bit much, it's nice to have their weapon stolen now and then when they get overconfident, but that's my opinion.

warior4356:
@Qwertyotoforty Yes but anti-drop can cause grenade hilarity, like the clip shows. Plus, how many people will actually use it? Most people smart enough to use it won't stupidly lose their weapon and it would prevent things like suicide bombing from making them drop it.

Spacemanspark:
@Qwertytoforty we typically deploy Deathsquad only when we want the round to end. This is one of the few scenarios where I'm fine with something being overwhelmingly overpowered.

PR #16125, opened 2021-06-03
Addition to contribution guidelines, mapping section.

S34NW:
## What Does This PR Do
This PR adds some general rules and principles that people should follow when mapping for paradise. Most of them are common sense things, and the list is not exhaustive for every small scenario. The guidelines cover the most commonly seen mapping mistakes, and setting a point for every mistake ever is not feasible.
[There has been a feedback thread on the forum for a week with no objections.](https://www.paradisestation.org/forum/topic/20824-addition-to-the-mapping-standards-section-of-contribution-guidelines/)

## Why It's Good For The Game
Consistency will be key for paradise moving forward, especially now map rotation is implemented.
Having guidelines helps PR reviewers express that suggested changes are not due to personal preference, but are due to the standards we follow.
Having some guidelines in place will also help newer mappers feel comfortable that they arent breaking any "hidden" standards we have for mapping.

## Preview
[You can view a preview of the changes, seen in green, by clicking this link.](https://github.com/ParadiseSS13/Paradise/pull/16125/files?short_path=98e64bc#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44c)

## Changelog
:cl:
/:cl:

PR #16277, opened 2021-06-30
[MDB IGNORE] Map variable sanitisation

S34NW:
## What Does This PR Do
This PR removes all redundant variables that are declared on the map. (This means any variable declared on the map that is the same as default.)
It is similar to #15438 however touches on more variables, and does not touch the main station levels like that PR.
This brings our various non-station map files in line with proposed changes in #16125
Since it also runs them through SDMM, it will reorder variables into the most optimal order.

## Why It's Good For The Game
Map bloat bad. There is no impact for players, changes are purely backend.

### DIFF
--- a/_maps/map_files/RandomRuins/SpaceRuins/deepstorage.dmm
+++ b/_maps/map_files/RandomRuins/SpaceRuins/deepstorage.dmm
@@ -113,7 +113,6 @@
 "aj" = (
 /obj/structure/closet/cardboard,
 /obj/machinery/light{
-	icon_state = "tube1";
 	dir = 8
 	},
 /obj/item/flashlight/flare,
@@ -173,8 +172,7 @@
 "ao" = (
 /obj/structure/closet/cardboard,
 /obj/machinery/light{
-	dir = 4;
-	icon_state = "tube1"
+	dir = 4
 	},
 /obj/item/ammo_box/c9mm,
 /obj/item/ammo_box/c9mm,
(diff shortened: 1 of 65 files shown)


_maps/map_files/RandomRuins/SpaceRuins/deepstorage.dmm
```diff
@@ -342,14 +337,14 @@
 "aH" = (
 /obj/machinery/vending/clothing,
 /turf/simulated/floor/wood{
-	tag = "icon-wood-broken";
-	icon_state = "wood-broken"
+	icon_state = "wood-broken";
+	tag = "icon-wood-broken"
```
AffectedArc07, line 341: You got any way of cleansing this `tag` nonsense as well?

S34NW: hmm, are they even used in anything useful?

S34NW: ah, they are used for landmarks to specify spawn points, so filtering them out is a bit difficult

S34NW: Having said that, all of those are `tag = "start*[taghere]"` so you could run a reasonably easy regex to keep these and delete the others

S34NW:
This is now being done in #15438

PR #16278, opened 2021-07-01
Adds Fake Blindfold

shmabuu:
## What Does This PR Do
Adds a fake blindfold with a different name so people know it's fake if they examine the person wearing it. It looks exactly the same as a blindfold otherwise. It does not blind the wearer and will not prevent a vampire from glaring.

## Why It's Good For The Game
This is purely meant for fashion and games/tricks. Currently if you'd like to wear a blindfold but still be able to see, the only option is being a cultist using the zealots blindfold. Otherwise, you'd have to be blind. 

## Images of changes
blind
blind2

## Changelog
:cl:
add: Fake blindfold
/:cl:

PS: The name and crafting methods are free to change whatever may be better than what it is right now, I'm open to suggestions.


shmabuu:
My brain is huge

code/modules/clothing/glasses/glasses.dm
```diff
@@ -395,6 +395,12 @@
 	tint = 3				//to make them blind
 	prescription_upgradable = 0
 
+/obj/item/clothing/glasses/blindfold_fake
```
hal9000PR, line 398: Probably want to have the `prescription_upgradable` var here set to false. Like in normal blindfolds.

hal9000PR (changes requested):

code/modules/clothing/glasses/glasses.dm
shmabuu: Good idea

code/modules/clothing/glasses/glasses.dm
```diff
@@ -395,6 +395,13 @@
...
+	name = "tattered blindfold"
+	desc = "A see-through blindfold perfect for cheating at games like pin the stunbaton on the clown."
+	icon_state = "blindfold"
+	item_state = "blindfold"
+	prescription_upgradable = 0
```
dearmochi, line 403:
```suggestion
	prescription_upgradable = FALSE
```

Darkmight9:
You could possibly also add it to the eyewear section of the loadout though I don't know if the heads would like the possible idea of lots of people running around with the fake blindfolds.

shmabuu:
> You could possibly also add it to the eyewear section of the loadout though I don't know if the heads would like the possible idea of lots of people running around with the fake blindfolds.

This is surely an option however, I didn't do it just because I didn't think it belonged there. Also, I made it have a different name on examine for a reason. If someone takes a moment to examine someone wearing one of these they'll be able to tell it's fake

code/modules/clothing/glasses/glasses.dm
shmabuu: Ah I made it =0 like normal blindfolds have because what was suggested before was a bit conflicting in the words. I'll do this though.

dearmochi (approved):

Carthusia:
heh, use wirecutters on it to create 2 eye holes and now you can be Zoro

SabreML (approved):

code/modules/clothing/glasses/glasses.dm
```diff
@@ -395,6 +395,13 @@
 	tint = 3				//to make them blind
 	prescription_upgradable = 0
 
+/obj/item/clothing/glasses/blindfold_fake
+	name = "tattered blindfold"
+	desc = "A see-through blindfold perfect for cheating at games like pin the stunbaton on the clown."
+	icon_state = "blindfold"
+	item_state = "blindfold"
+	prescription_upgradable = FALSE
```
Fox-McCloud, line 403: This should be a blindfold sub-type that doesn't actually blind, as opposed to something completely separate.

DeveloperLoser:
Love the idea and 100/10 examine text

shmabuu:
Fork was deleted I gotta remake this

PR #16284, opened 2021-07-02
Nerfs sorium and LDM

AffectedArc07:
## What Does This PR Do
This PR makes it so sorium and LDM can only launch/pull 1 atom per 5 units of reagent created

## Why It's Good For The Game
You shouldn't be able to launch 20 hatchets into a blob and only use 1 unit of the reagent, to create instakill cheese devices with advanced release grenades. Its awful.

## Changelog
:cl: AffectedArc07
tweak: Sorium and Liquid Dark Matter now limit the amount of atoms (objects and mobs) they can throw, requiring 5 units of each reagent per atom thrown/pulled.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes.dm
+++ b/code/modules/reagents/chemistry/recipes.dm
@@ -53,8 +53,19 @@
 				for(var/j = 1, j <= rand(1, 3), j++)
 					step(S, pick(NORTH,SOUTH,EAST,WEST))
 
-/proc/goonchem_vortex(turf/T, setting_type, volume)
-	if(setting_type)
+/**
+  * Throws or pulls objects to/from a chem reaction
+  *
+  * Scales the amount of objects thrown with the volume, unless ignore_volume is TRUE
+  *
+  * Arguments:
+  * * T - The turf to use as the throw from/to point
+  * * pull - Do we want to pull objects towards T (TRUE) or push them away from it (FALSE)
+  * * volume - The volume of reagents. Used to scale the effect is ignore_volume = FALSE
+  * * ignore_volume - Do we want to ignore the volume of reagents and just throw regardless
+  */
+/proc/goonchem_vortex(turf/T, pull, volume, ignore_volume = FALSE)
+	if(pull)
 		new /obj/effect/temp_visual/implosion(T)
 		playsound(T, 'sound/effects/whoosh.ogg', 25, 1) //credit to Robinhood76 of Freesound.org for this.
 	else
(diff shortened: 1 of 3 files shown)


LordShekelstein:
Merge it. MERGE IT NOW!!!

code/modules/reagents/chemistry/recipes.dm
```diff
@@ -60,6 +60,10 @@
 	else
 		new /obj/effect/temp_visual/shockwave(T)
 		playsound(T, 'sound/effects/bang.ogg', 25, 1)
+	// PARADISE EDIT: Allow only a certain amount of atoms to be pulled per unit
+	var/units_per_atom = 5
```
dearmochi, line 64: A `#define` would perhaps be more appropriate?

code/modules/reagents/chemistry/recipes.dm
```diff
@@ -68,3 +72,6 @@
 				X.throw_at(T, 20 + round(volume * 2), 1 + round(volume / 10))
 			else
 				X.throw_at(get_edge_target_turf(T, get_dir(T, X)), 20 + round(volume * 2), 1 + round(volume / 10))
+			moved_count++
+			if(moved_count >= atoms_to_move)
```
dearmochi, line 76:
```suggestion
			if(moved_count++ >= atoms_to_move)
```

dearmochi (changes requested):

AffectedArc07:
> Merge it. MERGE IT NOW!!!
> Edit: Actually just give it a cap of how many items it can throw, or alternatively cap the damage the thrown items can do when pulled/pushed by the chemical whilst applying this fix.

The max amount of LDM or sorium you can make here is 600 units (2 bluespace beakers in a grenade), which means 120 atoms thrown, at the cost of 2 bluespace beakers and other stuff wasted. 

Most sorium abuse comes from being able to spam it in advanced release grenades, and catapult 20 things with only 5 units used, making it spammable. 

Reducing the amount that can be thrown per reagent cuts down on their, as you can either launch enough things but use way too much reagent to make it viable to be portable, or you can spam it and only move 1 thing at a time, which means it takes too long and you perish.

code/modules/reagents/chemistry/recipes.dm
AffectedArc07: Maybe, but its only scoped to one place and probably for the best to be like this to keep it in one place

LordShekelstein:
> Most sorium abuse comes from being able to spam it in advanced release grenades, and catapult 20 things with only 5 units used, making it spammable.

That's where I disagree. Feasel, the player, has filled bluespace trashbags with hatchets, placed them on the ground, drank sorium, and sent them flying into blobs, megafauna, etc, killing them instantly. With this a player could easily swallow three 100u sorium pills and do 720 brute damage. I mean, granted, they have 300u of sorium in their bellies and that's no good, but that's still enough to kill several people.

AffectedArc07:
> That's where I disagree. Feasel, the player, has filled bluespace trashbags with hatchets, placed them on the ground, drank sorium, and sent them flying into blobs, megafauna, etc, killing them instantly. 

Thats a player problem which I will handle by normal means if I have to. This nerf cuts down on the advance release grenade spam

> With this a player could easily swallow three 100u sorium pills and do 720 brute damage. 

You cant make pure sorium pills, youd have to make the other 2, and even with 200 units, youd still only move 40 atoms, which yes 40 hatchets would do damage, but the amount of prep work to empty out 40 hatchets and then hope the other player doesnt move while you do this and down the pills


If needby I can put a maximum amount of atoms moved on sorium or LDM, but we will see how this goes for now.

LordShekelstein:
> If needby I can put a maximum amount of atoms moved on sorium or LDM, but we will see how this goes for now.

Yeah, reasonable. Thanks for your input. Makes more sense to me now.

dearmochi (approved):

Kyep (approved):

code/modules/reagents/chemistry/recipes.dm
```diff
@@ -60,6 +60,10 @@
 	else
 		new /obj/effect/temp_visual/shockwave(T)
 		playsound(T, 'sound/effects/bang.ogg', 25, 1)
+	// PARADISE EDIT: Allow only a certain amount of atoms to be pulled per unit
+	var/units_per_atom = 5
+	var/atoms_to_move = round(volume / units_per_atom)
+	var/moved_count = 0
```
Fox-McCloud, line 66: May want to adjust `/datum/reagent/consumable/drink/apple_pocalypse` and `/datum/reagent/consumable/drink/grape_granade` so they actually function, post these changes.

PR #16288, opened 2021-07-02
Remove the chance for cargo shipments to lose items.

warriorstar-orion:
## What Does This PR Do

This PR removes the chance for cargo shipments to lose items.

## Why It's Good For The Game

Losing shipment items has no benefit and is not fun. Cargo on Paradise is already a mess of logistics, people moving in and out of the office, crates being dragged around, wrapped, mailed, manifests, requests, and stamping, etc. Losing shipment items is not anything that can be prevented by the crew, and only serves to frustrate both cargo and the crew waiting for their items.

The other two random events, wrong station and wrong crate count, are both fine to keep, since they don't result in the loss of any expected items, just a handful of cargo points, which is never noticed once disks come. Wrong station crates are simply rejected and sent back, and wrong count only changes the content of the manifest, not any of the shipped crates.

## Changelog
:cl:
del: Cargo items will no longer be lost in transit.
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -143,7 +143,6 @@
 
 #define MANIFEST_ERROR_NAME		1
 #define MANIFEST_ERROR_COUNT	2
-#define MANIFEST_ERROR_ITEM		4
 
 //Turf wet states
 #define TURF_DRY		0
--- a/code/modules/shuttle/supply.dm
+++ b/code/modules/shuttle/supply.dm
@@ -78,8 +78,6 @@
 			errors |= MANIFEST_ERROR_COUNT
 		if(prob(5))
 			errors |= MANIFEST_ERROR_NAME
-		if(prob(5))
-			errors |= MANIFEST_ERROR_ITEM
 		SO.createObject(T, errors)
 
 	SSshuttle.shoppinglist.Cut()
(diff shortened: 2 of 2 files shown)


TDSSS:
I disagree on it doing nothing useful. It's a potential cover for an antag in cargo to pocket some items himself.

warriorstar-orion:
While this may be true, this tactic only would work on crates that aren't locked, and looking through supplypacks.dm I'm having a hard time finding items in unlocked crates that would be of great benefit to an antag. And the ones that skirt that line (exotic seeds? oil tank?) would be benign orders in the first place—you could just order them and most Cargonians wouldn't bat an eye.

And I would argue that the benefit this provides for an antagonist doesn't outweigh the frustration it causes cargo and the recipient, even before you take the above into account. Unless the recipient is the antag's target and keeping them from getting their supplies will slowly drive them insane, all it would otherwise do is make it either difficult for someone unrelated to the antag's objectives to do their job, or delay progress on people's personal projects, none of which sound very fun.

Aurorablade:
I think the chance to lose cargo was low enough to actually not be an issue. more funny would be the chance to be shipped something labeled for another station.

warriorstar-orion:
There's two such mechanics like that, that are slightly different from each other.

*If I'm reading the code correctly* there's a 5% chance that a *legitimate* order will have the wrong destination station on it, meaning if cargo approves it and sends it to the recipient, the only punishment is Cargo getting no points for it if they approve and send the manifest back.

(In theory, it seems, you can send the legit order with the wrong station name to the right person, then reject the manifest and get the points but I'm not an expert.)

There's also a 10% chance that any shuttle with crates being delivered receives a random crate whose destination is the Cyberiad but whose recipient is a random name that doesn't appear on the manifest.

AffectedArc07:
Highly against this.

This is one of those things where is actually adds something to the job, even if its bad. Cargo should be checking the manifests and having people report if their crates have the shit in or not. Engineering can wait 5 minutes if a TEG is missing a piece. 

Removing this chance removes the ability for that inter-department communication to happen, and just makes things less interesting.


This PR has 2 maintainer objections and likely wont go anywhere.

PR #16292, opened 2021-07-03
Refactors ban logging

AffectedArc07:
## What Does This PR Do
Removes `data/ban_unban_log.txt`

There several reasons for this:
- This file is not indexed or read by anything we use 
- Its not filtered by round and is multiple years lumped into one file
- The entries are not even timestamped
- We already log this in `game.log` and in the database
- This file uses the old logging system which is very expensive, and only gets worse as the file grows in size
- This file is 2.71MB and is **never** used, just wasting CPU cycles on useless writes

## Why It's Good For The Game
See above

## Changelog
N/A

### DIFF
--- a/code/modules/admin/NewBan.dm
+++ b/code/modules/admin/NewBan.dm
@@ -138,7 +138,6 @@ GLOBAL_PROTECT(banlist_savefile) // Obvious reasons
 		log_admin("Ban Expired: [key]")
 		message_admins("Ban Expired: [key]")
 	else
-		ban_unban_log_save("[key_name_admin(usr)] unbanned [key]")
 		log_admin("[key_name_admin(usr)] unbanned [key]")
 		message_admins("[key_name_admin(usr)] unbanned: [key]")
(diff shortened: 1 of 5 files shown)


code/modules/admin/NewBan.dm
```diff
@@ -138,7 +138,6 @@ GLOBAL_PROTECT(banlist_savefile) // Obvious reasons
 		log_admin("Ban Expired: [key]")
 		message_admins("Ban Expired: [key]")
 	else
-		ban_unban_log_save("[key_name_admin(usr)] unbanned [key]")
```
AffectedArc07, line 141: You can see how all of these already have `log_admin` right below them

code/modules/admin/db_ban/functions.dm
```diff
@@ -414,6 +414,7 @@
 	qdel(query_update)
 
 	message_admins("[key_name_admin(usr)] has lifted [pckey]'s ban.")
+	log_admin("[key_name(usr)] has lifted [pckey]'s ban.")
```
AffectedArc07, line 417: Apart from this one, which now has a line for it

farie82 (approved):

PR #16296, opened 2021-07-03
Fixes an abductor camera console exploit

SabreML:
## What Does This PR Do

Fixes the abductor 'Human Observation Console' starting on the CentComm z level rather than the station one.
Because of this, abductors are currently able to reliably teleport to any location on the z level after following a few steps.

## Why It's Good For The Game

I won't go into detail on the steps to exploit this obviously, but currently abductor teams have very easy access to:

Instagib rifles (Which admittedly they can't use),
Wizard melee weapons (Which they *can* use),
Adminordrazine bottles,
Literally every syndicate implanter,
Above all access ID cards,
Sol trader equipment,
Vox raider equipment,
Sleeping carp scrolls,
(Limited) SST equipment,
And finally the 'Toolbox of Robustness'. A syndicate toolbox with a `force` value of 100,000,000.

Realistically the only thing stopping them from abusing this is the risk that an admin will be told.

## Images of changes

Prevents this from happening:
(Tested on a live server with no adminbus involved)
image

## Changelog
:cl:
fix: Fixed abductor camera consoles starting on the CentComm z level.
/:cl:

### DIFF
--- a/code/game/machinery/computer/camera_advanced.dm
+++ b/code/game/machinery/computer/camera_advanced.dm
@@ -75,7 +75,7 @@
 		for(var/obj/machinery/camera/C in GLOB.cameranet.cameras)
 			if(!C.can_use())
 				continue
-			if(C.network&networks)
+			if(length(C.network & networks))
 				camera_location = get_turf(C)
 				break
 		if(camera_location)


code/game/machinery/computer/camera_advanced.dm
```diff
@@ -75,7 +75,7 @@
 		for(var/obj/machinery/camera/C in GLOB.cameranet.cameras)
 			if(!C.can_use())
 				continue
-			if(C.network&networks)
+			if(length(C.network & networks))
```
AffectedArc07, line 78: So what happens with this now.

Does it start on Z2 or what?

SabreML: Oh yeah, probably should've explained this, huh?

Short explanation:
This is the code which decides where the camera 'Eye' is spawned initially, with this change it only selects cameras who's `network` contains one of the same things from the console's `networks`. In this case, `list("SS13", "Abductor")`.
This bug was most likely introduced by map standardisation/rotation, since they changed around the load order.

<details>
<summary><b>Long explanation:</b></summary>

In this case the camera it was always connecting to had the networks `list("ERT", "CentComm")`. Obviously neither of those match what the camera console has.
The reason for this is that the `&` operator was being used incorrectly, specifically this part:
> [If A and B are lists, the result is a list that contains only items that were in both lists, in the order of list A.](http://www.byond.com/docs/ref/#/operator/&)

Even though nothing inside the list matched, it was still returning an empty list, and the `if()` was taking that as confirmation.
```dm
// Before
var/list/L1 = list("ERT", "CentComm")
var/list/L2 = list("SS13", "Abductor")
var/list/L3 = (L1 & L2) // << Produces an empty `list()`.
if(L3) // The list exists.
	camera_location = get_turf(C)
```
```dm
// After
var/list/L1 = list("ERT", "CentComm")
var/list/L2 = list("SS13", "Abductor")
var/list/L3 = (L1 & L2) // << Produces an empty `list().
if(length(L3)) // Length of 0.
	camera_location = get_turf(C)
```

</details>

AffectedArc07: Ah listing. I thought it was bitflags at first since `&`. Makes sense.

AffectedArc07 (approved):

Feasel:
So wait. Is this why I was seeing abductors running about with AA like day or so ago?

PR #16300, opened 2021-07-04
Fixes getting TK abilities when you shouldn't

### LINKED ISSUE #16298: Inactive Telekinesis can be used
**Issue Description**:

If your telekinesis block has a value that should activate it, but didn't roll for the activation, it can still be used without problems.

For telekinesis to work only when active.

**What happened instead**:

Telekinesis is usable without problems from the activation chance, and even without instability, or visible overlay.

Mutation is not visible, nor gives instability.

1. Find TK block.
2. Mutate it above activation threshold.
3. Don't roll the 15% chance to activate it.
4. Have telekinesis without drawbacks.

Not sure

**Extra information**:

Think it's due to the check [here](https://github.com/ParadiseSS13/Paradise/blob/ec967cbaa7b3d96ce2bd0e726e514e5e04ebb30c/code/_onclick/telekinesis.dm#L42), it does check for the SE state, and not whether the mutation is in the active mutations list. Other mutations may also be affected, as ``GetSEState`` seems to be getting used for a few other things than to (de)activate mutations.

hal9000PR:
## What Does This PR Do
fixes #16298

Makes telekinesis into a trait.

## Why It's Good For The Game
bugs bad

## Changelog
:cl:
fix: you no longer get TK when the gene doesn't activate.
/:cl:


code/_onclick/telekinesis.dm
```diff
@@ -109,7 +109,7 @@
 	if(!host || host != user)
 		qdel(src)
 		return
-	if(!host.dna?.GetSEState(GLOB.teleblock))
+	if(!host.dna && HAS_TRAIT(host, TRAIT_TELEKINESIS))
```
SteelSlayer, line 112: Small logic error here. The previous code was seeing if they *did not* have telekinesis. You are now checking if they  *do* have telekinesis. Should be:
```suggestion
	if(!host.dna && !HAS_TRAIT(host, TRAIT_TELEKINESIS))
```

code/_onclick/telekinesis.dm
```diff
@@ -39,7 +39,7 @@
 /obj/item/attack_tk(mob/user)
 	if(user.stat || !isturf(loc))
 		return
-	if(user.dna?.GetSEState(GLOB.teleblock) && !user.get_active_hand()) // both should already be true to get here
+	if(user.dna && HAS_TRAIT(user, TRAIT_TELEKINESIS) && !user.get_active_hand()) // both should already be true to get here
```
SteelSlayer, line 42: Now that I think about it, I don't think you need these `if(X.dna` checks anymore. Since before they were needed to access the `GetSEState` proc. Personally, I say remove them.

code/_onclick/telekinesis.dm
```diff
@@ -109,7 +109,7 @@
 	if(!host || host != user)
 		qdel(src)
 		return
-	if(!host.dna?.GetSEState(GLOB.teleblock))
+	if(!host.dna && !HAS_TRAIT(host, TRAIT_TELEKINESIS))
```
SteelSlayer, line 112: Even if we'll be keeping the `if(X.dna` checks that I mentioned in my other review, I just realized this is backwards as well. Now this will only pass if they suddenly have a null `dna` var, which is probably not what we want.
```suggestion
	if(host.dna && !HAS_TRAIT(host, TRAIT_TELEKINESIS))
```

code/_onclick/telekinesis.dm
hal9000PR: removed the checks

SteelSlayer (approved):

AffectedArc07 (approved):

PR #16302, opened 2021-07-04
[TGUI] Adds a HUD management UI for ghosts

SteelSlayer:
## What Does This PR Do
Adds a HUD management UI for ghosts, which allows ghosts to individually set whichever combination of HUDs they want. Inspired by a comment from Kyet on #16256. Also, if and/or when that PR is merged, I plan to incorporate the radiation toggle onto this UI as well.

## Why It's Good For The Game
Cleans up code a bit. Consolidates all HUD related options into a single verb/UI. 

## Images of changes
i0PwIxi4jm

## Changelog
:cl:
add: Adds a HUD management TGUI for ghosts. This can be found by using the "Ghost HUD Panel" verb in the ghost tab.
remove: The "Toggle All/Sec/Med/Diag HUDs" and "Toggle AntagHUD" verbs have been removed in favor of the "Ghost HUD Panel" verb.
/:cl:


S34NW:
Can we see a preview of the ahud alert?

SabreML:
Feels like the 'AntagHUD' setting should have some sort of difference visually. Like a warning box, or a different colour, something like that. Even if it still gives a popup when you click it, it's better UX wise to have a distinction.

Also maybe see if you can add little icons next to each option? Don't think it's *neccesary*, but it might look cool. (Cross for medical, shield or something for sec, etc.)

SteelSlayer:
@S34NW It's just the plain white `alert()` that we currently have. I didn't have enough space to include a modal for the UI, it's too small, otherwise I was planning on that.

@SabreML I'll see what I can do about making a distinction, probably just moving them into their own <Section>. No promises on the icons.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -310,82 +309,24 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
...
 /mob/dead/observer/proc/remove_the_hud(hud_index) //remove old huds
 	var/datum/atom_hud/H = GLOB.huds[hud_index]
 	H.remove_hud_from(src)
+	data_hud_seen -= hud_index
 
-/mob/dead/observer/verb/toggle_medHUD()
+/mob/dead/observer/verb/open_hud_panel()
 	set category = "Ghost"
-	set name = "Toggle All/Sec/Med/Diag HUDs"
-	set desc = "Toggles the HUDs."
```
AffectedArc07, line 321: Do not delete this verb.

I use it a lot on a macro.

AffectedArc07 (changes requested):

code/modules/mob/dead/observer/observer.dm
SteelSlayer: If I leave in this verb while including the new one, we'd have two verbs that *essentially* do the same thing. I'd rather not keep the old ones.

Kyep (approved):

AffectedArc07 (approved):
RIP my macros

AffectedArc07:
Will merge once conflicts are resolved

SteelSlayer:
Conflicts resolved, made some slight changes to the UI based on feedback (screenshot in description updated).

tgui/packages/tgui/interfaces/GhostHudPanel.js
```diff
@@ -12,11 +12,12 @@ export const GhostHudPanel = (props, context) => {
   } = data;
   return (
     <Window resizable theme="nologo">
```
AffectedArc07, line 14: This probably doesnt need to be `resizable`

SteelSlayer: No, it does not, will remove.

AffectedArc07 (approved):

PR #16305, opened 2021-07-04
Fixes an SQL error from maximum name length

### LINKED ISSUE #16304: You can set a name longer than the DB allows
**Issue Description**:

See title

Game to cap it 

**What happened instead**:

image

People will try to save their character 50 times and spam admin logs and server error logs. This isn't hyperbole. I am writing this issue after being woken up from concerns with an admin chat window full of DB errors.

Set a long name

The dawn of time

**Extra information**:

Why is it that whenever I want to sleep stuff always breaks.

SabreML:
## What Does This PR Do

Fixes #16304 by increasing the maximum name length in the DB from 45 to 55, as anything over 45 gives SQL errors.
`MySqlError { ERROR 1406 (22001): Data too long for column 'real_name' at row 1 }`
image

## Why It's Good For The Game

Apparently someone was *really* insistent that their character name got saved, and spammed out ~50 SQL errors.
This doesn't change the actual ingame name limit either, just the backend one.

## Images of changes

image

## Changelog
:cl:
fix: Fixed an SQL error when setting a character's name to above 45 letters.
/:cl:


LordShekelstein:
Could you not just set it to 45 characters? Asking as my Kidan has a 45 character name.

SabreML:
> Could you not just set it to 45 characters? Asking as my Kidan has a 45 character name.

Affected suggested having it be two characters below the limit set on the database, and I have to agree. As far as I can tell there's no way to get it to accept a name above `MAX_NAME_LEN` so there'd probably be no risk, but as a general rule of thumb it's a bad idea to have things like this be hitting their absolute limit.

I'll look into it anyway tomorrow, assuming nobody objects to it by then, but right now it's 1AM so I'm gonna go to sleep. :P

FlattestGuitar:
@SabreML just add a db migration that will move the limit up. No point in breaking existing names.

AffectedArc07 (changes requested):
Flattest has a point.

Tweak the DB to make the max length 55. Should be a pretty easy migration.

AffectedArc07 (approved):
Will merge once TOML config is in. Which is hopefully soon.

PR #16306, opened 2021-07-04
Fixes a couple things related to miner borg upgrades and kinetic accelerator modkits

### LINKED ISSUE #16301: KA upgrades don't apply for mining borgs
**Issue Description**:
When roboticists try to apply KA upgrades to mining borgs that just fall to the floor instead of applying.

Expected upgrades to apply as all other do.

**What happened instead**:
Upgrades did nothing other than fall to the floor.

Can no longer upgrade mining borg KAs

Try to upgrade a mining borg KA

7/4/21

**Extra information**:

SteelSlayer:
## What Does This PR Do
Fixes #16301
Fixes an unlisted bug where you couldn't crowbar out a mining cyborg's  kinetic accelerator mods.

## Changelog
:cl:
fix: Fixes a bug where you could not apply kinetic accelerator upgrades properly to mining cyborgs
fix: Fixes a bug that made it impossible to crowbar out a mining cyborg's  kinetic accelerator mods.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/kinetic_accelerator.dm
+++ b/code/modules/projectiles/guns/energy/kinetic_accelerator.dm
@@ -275,6 +275,9 @@
 		return ..()
 
 /obj/item/borg/upgrade/modkit/action(mob/living/silicon/robot/R)
+	if(!..())
+		return
+
 	for(var/obj/item/gun/energy/kinetic_accelerator/cyborg/H in R.module.modules)
 		return install(H, usr)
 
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/guns/energy/kinetic_accelerator.dm
```diff
@@ -275,9 +275,6 @@
 		return ..()
 
 /obj/item/borg/upgrade/modkit/action(mob/living/silicon/robot/R)
-	if(..())
-		return
```
AffectedArc07, line 279: Surely an `if(!..()) return` would be better here?

SteelSlayer: The reason I removed it completely is that, for the moment, modkits don't need any of the procs in `/obj/item/borg/upgrade/proc/action`, since they're not real borg upgrades. I can re-add an `if(!..()) return` if you want though.

AffectedArc07: For the sake of completeness, please do

AffectedArc07 (approved):

PR #16319, opened 2021-07-06
Jobban adjustments

### LINKED ISSUE #16318: Job banned players can become ghost roles despite being banned
**Issue Description**:

Job banned players are able to become ghost roles despite being job banned.

For job banned players to not be able to play roles they are banned in.

**What happened instead**:

For banned players to not be able to play roles they are banned in.

Pretty obvious why

1. Be jobbanned from a role.
2. Spawn in an event for that role
3. Sign up for the prompt despite being jobbanned.

Has been noticed only today (2021-07-06). Though I'd imagine it has been here since the new role prompt system has been introduced.

**Extra information**:

In this case job banned player was able to become a borer despite being banned from the borer role. They also mentioned they were able to become a terror before despite being banned. Other ghost roles haven't been tested, but I'd assume its the same for them too.

SabreML:
## What Does This PR Do

Adds a jobban category for Terror Spiders, fixed Cortical Borer jobbans being ignored, and swapped out a few `jobban_isbanned()` arguments with defines.
Tested a few of the other bannable antags and these are the only two I could find with the issue.
Fixes #16318.

## Why It's Good For The Game

If someone's been banned from a role, you'd generally expect that they won't be able to play the role anymore.
Also Terror Spiders having their own category just makes sense to me, since they're a pretty big antagonist and the only way to do it currently is a full antag ban.

## Images of changes

image
*(god that UI is ugly)*

## Changelog
:cl:
add: Added a jobban category for Terror Spiders.
fix: Fixed some jobban categories not working.
/:cl:

### DIFF
--- a/code/__DEFINES/role_preferences.dm
+++ b/code/__DEFINES/role_preferences.dm
@@ -45,6 +45,10 @@
 #define ROLE_EVENTMISC			"eventmisc"
 #define ROLE_GHOST				"ghost role"
 
+// Misc jobban categories
+#define ROLEBAN_AHUD			"AntagHUD"
+#define ROLEBAN_RECORDS			"Records"
+
 //Missing assignment means it's not a gamemode specific role, IT'S NOT A BUG OR ERROR.
 //The gamemode specific ones are just so the gamemodes can query whether a player is old enough
 //(in game days played) to play that role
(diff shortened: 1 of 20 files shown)


SteelSlayer (approved):
Please, anything that makes jobban code not a confusing mess. We need to switch over to using pure defines at some point.

Kyep (approved):
I don't think I've ever wanted to job ban someone from terror spider.
If they were that bad, I'd probably just job ban them from syndicate / all antags, because if I don't trust them to play terror (which has few limits) then there's no way I'd trust them to play traitor (an antag that has a lot more limits).
This isn't an objection. I'm fine with the option existing, I just don't think it will be used.

AffectedArc07 (approved):

PR #16322, opened 2021-07-07
Removes fungus announcement

S34NW:
## What Does This PR Do
This PR removes the fungus biohazard station announcement. (The event itself remains.)

## Why It's Good For The Game
This announcement is really over the top for the severity of the event. Calling it a "biohazard" is a bit much too considering that is strongly associated with Terrors or Blob.
All it serves to do is cause confusion in newer players or just disrupt the chat log for no reason.

## Changelog
:cl:
del: Removed large station alert regarding fungus.
/:cl:

### DIFF
--- a/code/modules/events/wallrot.dm
+++ b/code/modules/events/wallrot.dm
@@ -1,6 +1,3 @@
-/datum/event/wallrot/setup()
-	endWhen = rand(1, 300)
-
 /datum/event/wallrot/start()
 	INVOKE_ASYNC(src, .proc/spawn_wallrot)
 
--- a/code/modules/events/wallrot.dm
+++ b/code/modules/events/wallrot.dm
@@ -1,9 +1,5 @@
 /datum/event/wallrot/setup()
-	announceWhen = rand(0, 300)
-	endWhen = announceWhen + 1
-
-/datum/event/wallrot/announce()
-	GLOB.event_announcement.Announce("Harmful fungi detected on station. Station structures may be contaminated.", "Biohazard Alert")
+	endWhen = rand(1, 300)
 
 /datum/event/wallrot/start()
 	INVOKE_ASYNC(src, .proc/spawn_wallrot)


Wojz:
> Crewmate victory

Is this because there's a fungus among us.... Sean...

code/modules/events/wallrot.dm
```diff
@@ -1,9 +1,5 @@
 /datum/event/wallrot/setup()
```
farie82, line 1: Can be removed.
The event only has to start. Then it can be killed again. It would be a waste to have it tick while not doing anything.
If both `startWhen` and `endWhen` are 0, which is the default, then it'll only call `start` and then afterwards instantly call `end`

code/modules/events/wallrot.dm
```diff
@@ -1,9 +1,5 @@
...
-	announceWhen = rand(0, 300)
-	endWhen = announceWhen + 1
-
-/datum/event/wallrot/announce()
-	GLOB.event_announcement.Announce("Harmful fungi detected on station. Station structures may be contaminated.", "Biohazard Alert")
+	endWhen = rand(1, 300)
```
farie82, line 2: Or at least this one can be set to `0` instead of a random value

farie82 (changes requested):

code/modules/events/wallrot.dm
S34NW: Mm, I wasn't familiar enough with event code to know if it did things constantly through the lifetime of the event, or if it just did it at the start.

dearmochi (approved):

AffectedArc07:
image

🗞️

SteelSlayer (approved):
250px-591Amoonguss

code/modules/events/wallrot.dm
```diff
@@ -1,10 +1,3 @@
-/datum/event/wallrot/setup()
```
Aurorablade, line 1: I would not remove the annoucement i would reword it. https://github.com/ParadiseSS13/Paradise/blob/b151dfb0b9b5b53e41a5a2605d1f38471a8e87df/code/game/turfs/simulated/walls.dm#L112

Wallrot weakens the damage walls can take before you can break it. so yeah it can be an issue.
Rather then biohazard perhaps something about station structural integrity?

S34NW: Do people actively hunt for wallrot after hearing the announcement? As far as I'm aware, everyone ignores it and will only (sometimes) deal with it if they organically come across it.

Sirryan2002: > Do people actively hunt for wallrot after hearing the announcement? As far as I'm aware, everyone ignores it and will only (sometimes) deal with it if they organically come across it.

people don't for the most part

Aurorablade: if chem needs fungus for something. but REGARDLESS i think a rewording would be preferable to an outright removal

Wojz: > Do people actively hunt for wallrot after hearing the announcement? As far as I'm aware, everyone ignores it and will only (sometimes) deal with it if they organically come across it.

It gets dealt with but never on purpose from what I've seen.

Kyep (approved):

AffectedArc07 (approved):
Probably the most useless announcement in the entire game

PR #16325, opened 2021-07-08
Makes all Terror spiders immune to dragging

SabreML:
## What Does This PR Do

Prevents Terror spiders being dragged or pushed by crewmembers, expanding upon #15062.
(Talked with Kyet about this beforehand)

## Why It's Good For The Game

The current possible Terror spider spawns are:
1. Five Greens.
2. One Prince.
3. Two Whites.
4. Three Princesses.
5. One Queen.

Out of all of these, number 3 is by far the most underpowered.

The eggs they can inject into people only start spawning Greens after the fourth time they trigger, which can take upwards of 12 minutes to get to and only take a minute to remove.
There's two Whites maximum, so if one or both are new to the role they won't know what to do.
Even if they *do* know what they're doing, if they pop out of a vent to bite someone there's a pretty high chance that they won't be able to go back in. Right now there's nothing stopping people from dragging the spider away from the vents and hitting it with something until it dies.
This won't solve the other two problems obviously, but it should help to make them less of an issue.

## Changelog
:cl:
tweak: Made all Terror spiders immune to dragging.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/white.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/white.dm
@@ -20,6 +20,7 @@
 	melee_damage_lower = 5
 	melee_damage_upper = 15
 	spider_tier = TS_TIER_2
+	move_resist = MOVE_FORCE_STRONG
 	web_type = /obj/structure/spider/terrorweb/white
 
 
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/white.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/white.dm
@@ -20,7 +20,6 @@
 	melee_damage_lower = 5
 	melee_damage_upper = 15
 	spider_tier = TS_TIER_2
-	move_resist = MOVE_FORCE_STRONG
 	web_type = /obj/structure/spider/terrorweb/white
 
 
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/empress.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/empress.dm
@@ -24,7 +24,6 @@
 	projectiletype = /obj/item/projectile/terrorqueenspit/empress
 	icon = 'icons/mob/terrorspider64.dmi'
 	pixel_x = -16
-	move_resist = MOVE_FORCE_STRONG // no more pushing a several hundred if not thousand pound spider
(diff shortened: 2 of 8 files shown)


Kyep (approved):
This is overdue.
OOCly, whites were meant to be able to pop in, bite someone, and get away. They're not meant to be killable by random assistants with nothing more than a spear and pull spam. They're meant to take much more preparation/effort than that to kill.
ICly, its extremely unrealistic that crew could pull any terror, especially a tier 2 terror that has hit and run attacks as its niche.

White infections are meant to be in roughly the middle of the pack in terms of infection strength. Stronger than prince (because they can reproduce), weaker than princess (because reproduction is harder for them than it is for princesses). I did not anticipate crew dragging them away from vents like this. If I had, I'd have made them pull-immune from the start. Dragging them away to capture them is not how you're meant to deal with them.

AffectedArc07 (commented):
While I would normally disagree with a terror buff, this seems sane, and kinda an oversight given the others are immune, supposedly. 

Though now I mention that, the following spiders have no move resist:
- black
- brown
- gray
- green
- mother

Ones which do have it:
- prince
- purple
- queen
- red
- white (after this PR)

@Kyep is there reason for some of these and not others having it, because I would assume royals get the move resist but reds also have it.

Is this mass oversight or reason behind?

code/modules/mob/living/simple_animal/hostile/terror_spiders/white.dm
```diff
@@ -20,6 +20,7 @@
 	melee_damage_lower = 5
 	melee_damage_upper = 15
 	spider_tier = TS_TIER_2
+	move_resist = MOVE_FORCE_STRONG
```
AffectedArc07, line 23: Spoke to Kyet on discord. Please just scope this on the top level terror class and remove the overrides on the subtypes.

More consistency and less shitcode are good things.

AffectedArc07 (changes requested):

code/modules/mob/living/simple_animal/hostile/terror_spiders/white.dm
SabreML: Fair enough 👍

farie82: It was done on purpose in #15062 just as a note

AffectedArc07 (approved):

PR #16327, opened 2021-07-08
Trait Define Refactor

SabreML:
## What Does This PR Do

Refactors the `ADD_TRAIT()`, `REMOVE_TRAIT()`, `REMOVE_TRAITS_NOT_IN()`, and `REMOVE_TRAITS_IN()` defines in order to make them actually readable, and adds documentation for good measure.
(Definitely recommend looking at the before/after files rather than the diff when reviewing this.)

Also fixes adminordrazine removing *all* traits, including inherent species ones.

## Why It's Good For The Game

Admittedly I've been staring at this code for six hours so it may just be me, but personally I find that this:
```dm
#define ADD_TRAIT(target, trait, source) \
	do { \
		LAZYINITLIST(target.status_traits); \
\
		if(!target.status_traits[trait]) { \
			target.status_traits[trait] = list(source); \
		} else { \
			target.status_traits[trait] |= list(source); \
		} \
\
		SEND_SIGNAL(target, SIGNAL_ADDTRAIT(trait), trait); \
	} while (0)
```
is a hell of a lot easier to understand than this:
```dm
#define ADD_TRAIT(target, trait, source) \
	do { \
		var/list/_L; \
		if (!target.status_traits) { \
			target.status_traits = list(); \
			_L = target.status_traits; \
			_L[trait] = list(source); \
			SEND_SIGNAL(target, SIGNAL_ADDTRAIT(trait), trait); \
		} else { \
			_L = target.status_traits; \
			if (_L[trait]) { \
				_L[trait] |= list(source); \
			} else { \
				_L[trait] = list(source); \
				SEND_SIGNAL(target, SIGNAL_ADDTRAIT(trait), trait); \
			} \
		} \
	} while (0)
```

## Changelog
:cl:
fix: Fixed adminordrazine/nanites removing ALL traits, including inherent species ones.
/:cl:

### DIFF
--- a/code/__HELPERS/traits.dm
+++ b/code/__HELPERS/traits.dm
@@ -2,88 +2,115 @@
 #define SIGNAL_REMOVETRAIT(trait_ref) "removetrait [trait_ref]"
 
 // trait accessor defines
+
+/**
+ * Adds a status trait to the target datum.
+ *
+ * Arguments: (All Required)
+ * * target - The datum to add the trait to.
+ * * trait - The trait which is being added.
+ * * source - The source of the trait which is being added.
+ */
 #define ADD_TRAIT(target, trait, source) \
 	do { \
-		var/list/_L; \
-		if (!target.status_traits) { \
-			target.status_traits = list(); \
-			_L = target.status_traits; \
-			_L[trait] = list(source); \
-			SEND_SIGNAL(target, SIGNAL_ADDTRAIT(trait), trait); \
+		LAZYINITLIST(target.status_traits); \
+\
+		if(!target.status_traits[trait]) { \
+			target.status_traits[trait] = list(source); \
 		} else { \
-			_L = target.status_traits; \
-			if (_L[trait]) { \
-				_L[trait] |= list(source); \
-			} else { \
-				_L[trait] = list(source); \
-				SEND_SIGNAL(target, SIGNAL_ADDTRAIT(trait), trait); \
+			target.status_traits[trait] |= list(source); \
+		} \
+\
+		SEND_SIGNAL(target, SIGNAL_ADDTRAIT(trait), trait); \
+	} while (0)
+
+/**
+ * Removes a status trait from a target datum.
+ *
+ * `ROUNDSTART_TRAIT` traits can't be removed without being specified in `sources`.
+ * Arguments:
+ * * target - The datum to remove the trait from.
+ * * trait - The trait which is being removed.
+ * * sources - If specified, only remove the trait if it is from this source. (Lists Supported)
+ */
+#define REMOVE_TRAIT(target, trait, sources) \
+	do { \
+		if(target.status_traits && target.status_traits[trait]) { \
+			var/list/TARGET_TRAITS = target.status_traits; \
+			var/list/SOURCES = sources; \
+			if(sources && !islist(sources)) { \
+				SOURCES = list(sources); \
+			} \
+\
+			for(var/SOURCE in TARGET_TRAITS[trait]) { \
+				if((!SOURCES && (SOURCE != ROUNDSTART_TRAIT)) || (SOURCE in SOURCES)) { \
+					if(length(TARGET_TRAITS[trait]) == 1) { \
+						SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(trait), trait); \
+					} \
+					LAZYREMOVEASSOC(TARGET_TRAITS, trait, SOURCE); \
+				} \
 			} \
 		} \
 	} while (0)
-#define REMOVE_TRAIT(target, trait, sources) \
-	do { \
-		var/list/_L = target.status_traits; \
-		var/list/_S; \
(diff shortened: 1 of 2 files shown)


code/__HELPERS/traits.dm
```diff
@@ -2,89 +2,116 @@
...
-			_S = list(sources); \
-		} else { \
-			_S = sources\
-		}; \
-		if (_L && _L[trait]) { \
-			for (var/_T in _L[trait]) { \
-				if ((!_S && (_T != ROUNDSTART_TRAIT)) || (_T in _S)) { \
-					_L[trait] -= _T \
+		if(target.status_traits && target.status_traits[trait]) { \
+			var/list/TARGET_TRAITS = target.status_traits; \
```
SteelSlayer, line 39: It seems you can remove this variable and just stick with using `target.status_traits`. Was there any reason you decided not to? Same goes for the two other `REMOVE_` defines below.

code/__HELPERS/traits.dm
```diff
@@ -2,89 +2,116 @@
...
+			if(sources && !islist(sources)) { \
+				SOURCES = list(sources); \
+			} \
+\
+			for(var/SOURCE in TARGET_TRAITS[trait]) { \
+				if((!SOURCES && (SOURCE != ROUNDSTART_TRAIT)) || (SOURCE in SOURCES)) { \
+					if(length(TARGET_TRAITS[trait]) == 1) { \
+						SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(trait), trait); \
+					} \
+					LAZYREMOVEASSOC(TARGET_TRAITS, trait, SOURCE); \
```
SteelSlayer, line 50: If I'm reading this right, you're now checking for length == 1 instead because the `LAZYREMOVEASSOC` is about to remove the last value right after this `if`.

If that's true I think it would be better to put the removal before the `if` and keep the 0 length check like how it was before.
```dm
LAZYREMOVEASSOC(TARGET_TRAITS, trait, SOURCE); \
if(!length(TARGET_TRAITS[trait])) { \
	SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(trait), trait); \
} \
```

code/__HELPERS/traits.dm
```diff
@@ -2,89 +2,116 @@
...
+				SOURCES = list(sources); \
+			} \
+\
+			for(var/TRAIT in TARGET_TRAITS) { \
+				TARGET_TRAITS[TRAIT] &= SOURCES; \
+				if(!length(TARGET_TRAITS[TRAIT])) { \
+					TARGET_TRAITS -= TRAIT; \
+					SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(TRAIT), TRAIT); \
+					} \
+				} \
```
SteelSlayer, line 78: These were indented over one tab too far.
```suggestion
				} \
			} \
```

code/__HELPERS/traits.dm
```diff
@@ -2,89 +2,116 @@
...
+				SOURCES = list(sources); \
+			} \
+\
+			for(var/TRAIT in TARGET_TRAITS) { \
+				TARGET_TRAITS[TRAIT] -= SOURCES; \
+				if(!length(TARGET_TRAITS[TRAIT])) { \
+					TARGET_TRAITS -= TRAIT; \
+					SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(TRAIT)); \
+					} \
+				} \
```
SteelSlayer, line 107: Same issue here.
```suggestion
				} \
			} \
```

code/__HELPERS/traits.dm
SabreML: In this case I believe it was to try and help with readability.
```dm
for(var/TRAIT in TARGET_TRAITS) { \
	TARGET_TRAITS[TRAIT] &= SOURCES; \
	if(!length(TARGET_TRAITS[TRAIT])) { \
		TARGET_TRAITS -= TRAIT; \
		SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(TRAIT), TRAIT); \
```
vs
```dm
for(var/TRAIT in target.status_traits) { \
	target.status_traits[TRAIT] &= SOURCES; \
	if(!length(target.status_traits[TRAIT])) { \
		target.status_traits -= TRAIT; \
		SEND_SIGNAL(target, SIGNAL_REMOVETRAIT(TRAIT), TRAIT); \
```
Although looking at it now it might actually be better the second way...

code/__HELPERS/traits.dm
SabreML: Since `LAZYREMOVEASSOC()` also nulls the list entirely if it removes the last enty, `if(!length(null[trait]))` will give a "bad index" runtime.
This could be avoided by writing out `LAZYREMOVEASSOC()` manually like it was before, but I felt that using the existing define for readability was worth the seemingly arbitrary length check.
image

SteelSlayer (approved):

variableundefined (approved):

PR #16330, opened 2021-07-09
Refactors the 'Examine' status tab

SabreML:
## What Does This PR Do

Refactors all of the code for the 'Examine' tab to fix VSCode getting confused about sources. (The tab, *not* the shift+click text.)

## Why It's Good For The Game

This fixes the *very* annoying issue of ctrl+clicking `/obj/item/gun/projectile` as well as a few other items, and being taken to a completely unrelated file rather than the correct source.

## Changelog
:cl:
tweak: Fixed ctrl+clicking on some paths in VSCode taking you to the wrong file.
/:cl:


Spacemanspark:
I use this quite a bit, actually, especially as an admin when I need to look through a few flavortexts. It's more or less a harmless feature.

Not to mention it's a decent tab to use while the status tab slows down my FPS so much.

TORE01:
I use this quite a lot actually when the status bar causes my FPS to drop for reasons BYOND my control. It's not the intended use, but I'd like to keep it unless you want to replace it with something else.

SabreML:
> I use this quite a bit, actually, especially as an admin when I need to look through a few flavortexts.

Honestly hadn't considered the use for reading flavour texts, and that is a valid point. Although considering you can open them in a new window anyway, I can't see removing this tab causing too much extra hassle.

> It's more or less a harmless feature.

Have to disagree there. If you try ctrl+clicking on `/obj/machinery/power/supermatter_crystal` or `/obj/item/gun/projectile` in VSCode you'll be taken to their examine pages rather than the actual source object, and then have to manually search through the 404 other instances of `/obj/item/gun/projectile` before you finally find the real one.
Obviously it's not doing anything bad ingame, but from a code perspective it's *very* annoying when it pops up.

<hr>

> Not to mention it's a decent tab to use while the status tab slows down my FPS so much.

> I use this quite a lot actually when the status bar causes my FPS to drop for reasons BYOND my control. It's not the intended use, but I'd like to keep it unless you want to replace it with something else.

Personally I always used the OOC tab for that, although I've not had any issues at all since #16110 was merged.
All I can say is "Why not just use a different tab instead?".

TORE01:
> Personally I always used the OOC tab for that, although I've not had any issues at all since #16110 was merged.
> All I can say is "Why not just use a different tab instead?".

Because it's so nice and empty. Unless I do toxins or engineering I barely ever examine anything. If we had an empty bar I would probably use that instead.

SteelSlayer:
> Have to disagree there. If you try ctrl+clicking on /obj/machinery/power/supermatter_crystal or /obj/item/gun/projectile in VSCode you'll be taken to their examine pages rather than the actual source object, and then have to manually search through the 404 other instances of /obj/item/gun/projectile before you finally find the real one.
Obviously it's not doing anything bad ingame, but from a code perspective it's very annoying when it pops up.

Hello devils advocate here. This can probably be solved by relocating the `description_info` var to wherever the main object is defined. I wouldn't consider this a reason for removal.

SabreML:
> Hello devils advocate here. This can probably be solved by relocating the `description_info` var to wherever the main object is defined. I wouldn't consider this a reason for removal.

I had considered doing that, but I thought that adding a bunch of very long text strings to 84 items wouldn't be worth it. The space taken up by some of the descriptions is probably the reason they were all in their own files to begin with actually.
There's also the argument of code bloat, since nobody's updated or added to the descriptions for years now, and all of the information in them is also available on the wiki.

I could add the descriptions back in if enough people want it of course, but at some point it might end up being better to just link to the wiki pages directly.

Spacemanspark:
>Honestly hadn't considered the use for reading flavour texts, and that is a valid point. Although considering you can open them in a new window anyway, I can't see removing this tab causing too much extra hassle.

Consider that we have tons of admin logs coming in very quickly, and it's a tiny 'more' button that rapidly gets throw upwards... it's a bit of a hassle. The examine tab alleviates this issue, especially if I, an admin, am looking through a lot of tickets.

>Have to disagree there. If you try ctrl+clicking on /obj/machinery/power/supermatter_crystal or /obj/item/gun/projectile in VSCode you'll be taken to their examine pages rather than the actual source object, and then have to manually search through the 404 other instances of /obj/item/gun/projectile before you finally find the real one.
Obviously it's not doing anything bad ingame, but from a code perspective it's very annoying when it pops up.

I don't feel like this is a valid enough reason to remove a feature from the game itself, sorry. I'm really not a fan of gutting features for the sake of optimization, and it's something I feel Paradise does far too much of.

>Personally I always used the OOC tab for that, although I've not had any issues at all since #16110 was merged.
All I can say is "Why not just use a different tab instead?".

On occasion I have an issue where the game likes to lock up or force me to move in a particular direction without a button being pressed. I've found that clicking on the tabs are to be helpful for this, without risking clicking on something in game or further causing problems by clicking something in the chat. Using the OOC tab for this means that, in a critical situation where I need to fix this _fast_, I'm very liable to press something I do not want to.

It would appear that steelslayer has suggested an acceptable solution. And, yes, the point of items having more descriptions is also a valid one. I particularly enjoy the descriptions on the eguns we have now, for example.

Please do not remove this feature. It sees and has valid use.

SabreML:
I'll revert the tab removal since it looks like I was *wildly* off the mark on how many people actually use it (I had assumed 0), although I'll still keep this PR up so that I can try to fix the code issue.

code/modules/examine/examine.dm
```diff
@@ -1,62 +1,54 @@
...
-	return
-
-/mob/living/get_description_fluff()
+/**
+ * Used for showing a more detailed description in the 'Examine' tab after examining the atom.
+ *
+ * Shown as blue text in the Examine tab.
+ */
+/atom/proc/detailed_examine()
+	return ""
```
AffectedArc07, line 7: Ok I am gonna say it.

Can you make this take `mob/user` as an arg, so we can tweak this text from the user, similar to `examine(mob/user)`


Also `return null` for these

AffectedArc07 (changes requested):

code/modules/examine/examine.dm
SabreML: What would be the use of having the user as an argument? I could see in the future maybe having different text for different mob types, but currently it's all set up to only have a single description.
<hr>

> Also `return null` for these

Yep 👍
I was trying to imply that it needs to return something, but `return null` works just as well for that.

Kyep (approved):

code/modules/examine/examine.dm
```diff
@@ -1,62 +1,54 @@
...
-	description_holders["antag"] = (update_antag_info)? A.get_description_antag() : ""
 
+/client/proc/update_description_holders(atom/A, update_antag_info = FALSE)
 	description_holders["name"] = "[A.name]"
-	description_holders["icon"] = "[bicon(A)]"
 	description_holders["desc"] = A.desc
 
+	description_holders["info"] = A.detailed_examine()
+	description_holders["fluff"] = A.detailed_examine_fluff()
+	description_holders["antag"] = update_antag_info ? A.detailed_examine_antag() : null
```
AffectedArc07, line 41: If these ever become hard refs, you will be shot

AffectedArc07 (approved):

PR #16337, opened 2021-07-10
Fixes telekinesis crafting appearing at mob's feet

### LINKED ISSUE #8452: interactive objects have the chance to teleport to you when using telekinesis
**Problem Description**:
If you have telekinesis and use an item that is interactive like sheets of metal, the resulting item teleports to your hand.

Not entirely sure.

**What happened instead**:
I used metal sheets that were in another room to build floor tiles and the tiles teleported into my hand from the  other room.

I wouldn't say its important, its a bug.

Get tk from genetics, find something like sheets of metal that you can interact with, try to use them to build something like you would if it was in your hand like normal.

**Possibly related stuff (which gamemode was it? What were you doing at the time? nothing out of the ordinary to put here really,

warior4356:
## What Does This PR Do
Fixes #8452.

This PR makes it so when crafting with telekinesis items appear under the stack you crafted with, not under the mob. It also makes it so the progress bar appears where the stack is, regardless if that's on the floor or in a mob's hand.

## Why It's Good For The Game
Bugs and oversights are bad. Items teleporting to the mob with telekinesis is exploitable.

## Changelog
:cl:
fix: Fixed telekinesis crafting appearing at mob's feet
/:cl:


code/game/objects/items/stacks/stack.dm
```diff
@@ -201,31 +201,31 @@
 			to_chat(usr, "<span class='warning'>\The [R.title] won't fit here!</span>")
```
SabreML, line 201: You probably want to change the argument here for line 200 as well, since it has the same issue.

code/game/objects/items/stacks/stack.dm
```diff
@@ -201,31 +201,31 @@
...
 			to_chat(usr, "<span class='warning'>\The [R.title] must be constructed on the floor!</span>")
 			return FALSE
 
-		if(R.no_cult_structure && (locate(/obj/structure/cult) in usr.drop_location()))
+		if(R.no_cult_structure && (locate(/obj/structure/cult) in get_turf(src)))
 			to_chat(usr, "<span class='warning'>There is a structure here!</span>")
 			return FALSE
 
 		if(R.time)
 			to_chat(usr, "<span class='notice'>Building [R.title] ...</span>")
```
SabreML, line 217: While you're here could you remove the space after `[R.title]`? I never actually noticed it until just now but it'll bug me now that I have.
image

code/game/objects/items/stacks/stack.dm
```diff
@@ -201,31 +201,31 @@
...
-			if(!do_after(usr, R.time, target = usr))
+			if(!do_after(usr, R.time, target = src.loc))
```
SabreML, line 218: Normally it's preferred not to use `src.` anywhere since it's always implied, but in this case I think it makes the code more readable, so sure.

code/game/objects/items/stacks/stack.dm
AffectedArc07: Please make this change

AffectedArc07 (changes requested):
See Sabre's change

AffectedArc07 (changes requested):
You still need to make the other requested change

AffectedArc07 (approved):

PR #16351, opened 2021-07-12
Adds a discord message if the last mentor logs out

SabreML:
## What Does This PR Do

Adds the 'X has logged out, no admins online' message which admins get for mentors too, to avoid the issue of nobody realising that there's not been any mentors online for the last few hours because there haven't been any mentorhelps.
Pretty much the same premise as #15533, but as more of a informational notice than an urgent warning.

## Why It's Good For The Game

Aside from being another thing to help stop this from happening:
image
There's also a pretty high chance that people don't mhelp if they notice that there's no mentors online, so in theory this will avoid that too.

## Images of changes

unknown

## Changelog
:cl:
add: Added a discord message for mentors if the last mentor logs out.
/:cl:

### DIFF
--- a/code/modules/mob/logout.dm
+++ b/code/modules/mob/logout.dm
@@ -7,14 +7,18 @@
 	create_log(MISC_LOG, "Logged out")
 	// `holder` is nil'd out by now, so we check the `admin_datums` array directly
 	//Only report this stuff if we are currently playing.
-	if(GLOB.admin_datums[ckey] && SSticker && SSticker.current_state == GAME_STATE_PLAYING)
+	if(GLOB.admin_datums[ckey] && SSticker.current_state == GAME_STATE_PLAYING)
 		var/datum/admins/temp_admin = GLOB.admin_datums[ckey]
-		// Triggers on people with banhammer power only - no mentors tripping the alarm
-		if(temp_admin.rights & R_BAN)
+		if(temp_admin.rights & R_MENTOR)
+			var/list/mentorcounter = staff_countup(R_MENTOR)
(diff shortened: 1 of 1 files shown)


code/modules/mob/logout.dm
```diff
@@ -7,14 +7,18 @@
 	create_log(MISC_LOG, "Logged out")
 	// `holder` is nil'd out by now, so we check the `admin_datums` array directly
 	//Only report this stuff if we are currently playing.
-	if(GLOB.admin_datums[ckey] && SSticker && SSticker.current_state == GAME_STATE_PLAYING)
+	if(GLOB.admin_datums[ckey] && SSticker.current_state == GAME_STATE_PLAYING)
 		var/datum/admins/temp_admin = GLOB.admin_datums[ckey]
-		// Triggers on people with banhammer power only - no mentors tripping the alarm
-		if(temp_admin.rights & R_BAN)
+		if(temp_admin.rights & R_MENTOR)
```
farie82, line 12: Not sure how the permissions are set up. But what if an admin/host with `R_MENTOR` in his permission set logs out. It will only call the mentor has logged out part. And not the admin part.
Prob want to switch these two

farie82 (changes requested):

code/modules/mob/logout.dm
AffectedArc07: Explained in discord. This is fine and works in production.

dearmochi (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #16353, opened 2021-07-12
Removes some unnecessary global variables

SabreML:
## What Does This PR Do

Removes 40(ish?) global variables which as far as I can tell either aren't used for anything, or make a lot more sense being defined locally. I went through the whole list to get these and picked out variables which looked empty or unused. Anything which had a lot of interactions or that I didn't understand I left alone.
Also removes the 'Uristrune' code since it's not actually used anymore.

I didn't notice any changes when testing, but I'd definitely still recommend a testmerge of this anyway.

## Why It's Good For The Game

Some of these are set and then never referenced, some are far better as static vars than globals, some don't seem to make any sense at all, and some are completely unused.
In all of these cases, there's no benefit to having them all clutter up the global variables list and making it even more laggy than it already is.

## Changelog
:cl:
experiment: Removed/converted some global variables.
/:cl:


code/controllers/subsystem/acid.dm
```diff
@@ -8,9 +8,14 @@ SUBSYSTEM_DEF(acid)
...
 
+	var/acid_overlay
+
 /datum/controller/subsystem/acid/stat_entry()
-	..("P:[processing.len]")
+	..("P:[length(processing)]")
 
+/datum/controller/subsystem/acid/Initialize(start_timeofday)
+	. = ..()
+	acid_overlay = mutable_appearance('icons/effects/effects.dmi', "acid")
```
AffectedArc07, line 18:
```suggestion
/datum/controller/subsystem/acid/Initialize(start_timeofday)
	acid_overlay = mutable_appearance('icons/effects/effects.dmi', "acid")
	return ..()
```

AffectedArc07 (changes requested):

code/_globalvars/configuration.dm
```diff
@@ -8,15 +8,9 @@ GLOBAL_VAR_INIT(game_year, (text2num(time2text(world.realtime, "YYYY")) + 544))
...
 GLOBAL_VAR_INIT(traitor_scaling, 1)
-//GLOBAL_VAR_INIT(goonsay_allowed, 0)
-GLOBAL_VAR_INIT(dna_ident, 1)
 GLOBAL_VAR_INIT(abandon_allowed, 0)
 GLOBAL_VAR_INIT(enter_allowed, 1)
 GLOBAL_VAR_INIT(guests_allowed, 1)
-GLOBAL_VAR_INIT(shuttle_frozen, 0)
-GLOBAL_VAR_INIT(shuttle_left, 0)
-GLOBAL_VAR_INIT(tinted_weldhelh, 1)
-GLOBAL_VAR_INIT(mouse_respawn_time, 5) //Amount of time that must pass between a player dying as a mouse and repawning as a mouse. In minutes.
```
FlattestGuitar, line 19: should be a per-mouse class var smh

code/_globalvars/lists/names.dm
```diff
@@ -10,8 +10,6 @@ GLOBAL_LIST_INIT(last_names, file2list("config/names/last.txt"))
 GLOBAL_LIST_INIT(clown_names, file2list("config/names/clown.txt"))
 GLOBAL_LIST_INIT(mime_names, file2list("config/names/mime.txt"))
 GLOBAL_LIST_INIT(golem_names, file2list("config/names/golem.txt"))
-GLOBAL_LIST_INIT(megacarp_first_names, file2list("strings/names/megacarp1.txt"))
-GLOBAL_LIST_INIT(megacarp_last_names, file2list("strings/names/megacarp2.txt"))
```
FlattestGuitar, line 14: either all of these configs are global or none of them are. What the heck is this

SabreML: All of the rest of them are used in multiple files (Except the ninja names apparently), whereas the megacarp names are used in a single file for a single rare mob. I figured that it was worth the inconsistency in this case.

code/_globalvars/configuration.dm
Aurorablade: This is honestly better kept as a global rather then a local on the chance that something down the line needs to use it or to make it eaiser to get to later.

code/_globalvars/lists/names.dm
Aurorablade: it is better to keep consistancy then to deviate off and make it inconsistent in best practice.

code/_globalvars/lists/objects.dm
```diff
@@ -48,8 +47,6 @@ GLOBAL_LIST_EMPTY(ladders)
 GLOBAL_LIST_INIT(active_diseases, list()) 			//List of Active disease in all mobs; purely for quick referencing.
 
 GLOBAL_LIST_EMPTY(mob_spawners) 		    // All mob_spawn objects
-GLOBAL_LIST_EMPTY(alert_consoles)			// Station alert consoles, /obj/machinery/computer/station_alert
```
Aurorablade, line 51: the is one of those things that should remain a global for consistency and future proof purposes.

code/_globalvars/configuration.dm
SabreML: I really can't see any situation where `mouse_respawn_timer` would ever be needed as a global variable, since it's obviously not used by anything but mice. It could maybe be a config setting, but again I can't imagine it ever needing to be changed.

code/_globalvars/lists/names.dm
SabreML: I've thought about it a bit more, and yeah it would be better to keep it as it is.

code/_globalvars/lists/objects.dm
SabreML: I definitely agree that this could be useful for in the future for something, but currently there's no benefit at all in it being there. I can't think of any admin situations where the list would be useful either, and in the event it *was* needed, the same effect can be achieved with far less hassle using an SDQL2 query.

I could change it to just be commented out instead to avoid someone needing to remake it again in the future, but personally at least I can't see much reason for this to stay.

Aurorablade:
Let me kinda put it now that my brain has kinda thought a bit. I would personally perfer if this PR just removed global vars that are not used once anywhere in the code..cuase those are they ones you can add back easier when the things that use them are later added.

now going in and changeing globals to locals and rewriting code to accommodate, No go. it breaks consistency, it breaks best use case, and belive you me..even if you think it would make more sense as a static you may later find its kinda better to be able to grab it later instead of having to go around your thumb to reach your elbow to get to the a varible. DM isn't sane. OH god the refactor PTSD here.

SabreML:
> Let me kinda put it now that my brain has kinda thought a bit. I would personally perfer if this PR just removed global vars that are not used once anywhere in the code..cuase those are they ones you can add back easier when the things that use them are later added.
> 
> now going in and changeing globals to locals and rewriting code to accommodate, No go. it breaks consistency, it breaks best use case, and belive you me..even if you think it would make more sense as a static you may later find its kinda better to be able to grab it later instead of having to go around your thumb to reach your elbow to get to the a varible. DM isn't sane. OH god the refactor PTSD here.

I think I'm misunderstanding something here, because as far as I can tell that's exactly what I've done in the PR.
The only new static vars (after I reverted the carp thing) are `advance_cures`, `blacklisted_pylon_turfs`, and the tribble and photocopier stuff.
On the offchance that someone in the future desperately needs the tribble count for a... tribble detector or something, they can easily just make them global again.

code/_globalvars/configuration.dm
dearmochi: > 
> 
> This is honestly better kept as a global rather then a local on the chance that something down the line needs to use it or to make it eaiser to get to later.

"on the chance that something needs it" isn't a valid excuse.

dearmochi (approved):

SabreML:
Found and fixed 1 runtime which I somehow missed on this line:
`timedifference_text = time2text(MOUSE_RESPAWN_TIME - timedifference,"mm:ss" MINUTES)`
Everything else seems to work, so unsetting from DNM.

SteelSlayer (approved):

SabreML:
I've deconflicted this and the commit is on the PR's branch in c709329865a2f80e027d04eca1ccde14e5203ab3, but for some reason it isn't showing up on the PR itself.
I'll give it a day or so and then just webedit it, since it only needs two files reset to master.

PR #7379, opened 2017-05-23
RnD autolathe removal + tech level overhaul

tigercat2000:
Remake of #7101 

Removes the roundstart autolathe from Science, the research division should be able to research the board and build it themselves (or just raid tech storage) rather than having it handed to them on a plate

Ports the tg RnD tech level rebalance, with many origin_tech of items changes, req_tech of designs changed to make RnD more interdepartmental, cooperative and less something that you can do just inside of science

The Experimentor can no longer raise tech levels of items, but discovering strange objects will now give them random tech levels which may be useful

The Exosuit Fab and Protolathe get their efficiency upgrades back, with the right upgrades making the costs of items lower again

RnD tech level skipping jumps 1 level further than before

Removes a few references to reliability, which was scrapped from RnD long ago

Adds an item to Cargo for Engineering research

The Destructive Analyzer will only prompt the user if deconning the item won't raise tech levels

Max tech levels for job objectives increased to 7 for all techs

Removed the forensic scanner from RnD and put it in a cargo crate

:cl:
tweak: Upgrading the Exosuit Fabricator and Protolathe with the right parts will lower item costs
add: Added an item to Cargo that RnD can use
tweak: Tweaks the tech levels for many items and the required tech levels for their designs. This is to make RnD more of a station-wide effort.
tweak: RnD tech level skipping skips 1 level further
tweak: The Destructive Analyzer will only prompt the user if tech levels will not be increased.
del: The Experimentor can no longer raise tech levels of items. It can however discover the tech levels of strange objects.
del: Removes the Autolathe from the RnD Room.
del: Removed the Forensic Scanner from RnD. Cargo can order a Forensics crate now.
/:cl:

PR #16354, opened 2021-07-13
Add autolathe to science

theo-3:
## What Does This PR Do
Adds an autolathe to the science department.

## Why It's Good For The Game
Science needs an autolathe to make stuff like tools or prox sensors for bots and they always build one as soon as possible anyways.

## Images of changes
ezgif com-gif-maker

## Changelog
:cl:
add: Added an autolathe to science/robotics
/:cl:


Spacemanspark:
Ultimately I'm neutral towards this.

On one hand, I get very annoyed with roboticists that run off to steal the autolathe board from engineering five seconds after the round start.

On the other... this can be built fairly early on in, fairly cheap.

I'm also not sure how I feel about its placement in this PR, but perhaps I just need to adjust. I typically replace the reinforced table with an autolathe whenever I run robotics, so this looks a bit off kilter to me.

Overall, eh, I can definitely understand the logic behind this PR and appreciate it. I do, however, believe that roboticists should lose access to engineering tool storage, regardless of whether or not this change goes through. There is absolutely no need for them to have it, and engineering has far more immediate use for an autolathe than robotics or science.

theo-3:
I didn't remove the table because it's useful for giving stuff (like iron or yellow cores) to robotics without dropping things on the floor or waiting for them to come open the door

theo-3:
Also, I mostly added this for scientists to print tools. I placed it in robotics because that allows both robotics and science access to it without letting scientists into robotics

Fox-McCloud:
They had this before, and it was remove for a reason; they don't need it back.

Science is already the most overtuned department; returning it to its former glory is not in the cards.

AffectedArc07:
#7379


We removed it for a reason. 

In my honest opinion, R&D should be producing their own autolathe, not stealing the one from tech storage. 

Not in favour of this PR

FreeStylaLT:
img

PR #16359, opened 2021-07-14
Simple mob GC Optimisations

### LINKED ISSUE #16358: Gateway terror spiders never GC
**Issue Description**:
Terror spiders in the gateway never garbage collect.

Taken from multiple rounds
image
image
image
image

Them to GC properly atleast some of the time

**What happened instead**:
Every gateway terror spider has a 100% GC fail rate. This makes the GC cry bloody murder and use a lot of excess CPU (see timings)

See above.

1. Load gateway terror spider mission
2. Kill some of them
3. `Debug` > `Display del() log`

Unsure

**Extra information**:
Unsure

Edan52:
## What Does This PR Do

Fixes #16358

Makes some more simple mobs GC correctly (in most cases), namely terror spiders, minebots, swarmers, and any of those that have actions.

## Why It's Good For The Game
When there are terror spiders or the terror spider gateway, the CPU spends a significant amount of time deleting them as they do not GC correctly, in a recent round there was upwards of a full minute of CPU time spent purely on this.

## Images of changes
image
image
image

## Changelog
:cl: CornMyCob
fix: More simple mobs should now GC correctly
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -140,7 +140,6 @@
 	if (T && AIStatus == AI_Z_OFF)
 		SSidlenpcpool.idle_mobs_by_zlevel[T.z] -= src
 
-	QDEL_LIST(actions)
 	return ..()
 
 /mob/living/simple_animal/handle_atom_del(atom/A)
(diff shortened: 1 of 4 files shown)


dearmochi (approved):

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -140,6 +140,7 @@
 	if (T && AIStatus == AI_Z_OFF)
 		SSidlenpcpool.idle_mobs_by_zlevel[T.z] -= src
 
+	QDEL_LIST(actions)
```
Fox-McCloud, line 143: More than just simple mobs use the actions list. We've typically handled that on the mob itself, where unique actions are handled---or on the action itself for specific special use cases.

Kyep:
Thank you Edan52 for this fix.

PR #16362, opened 2021-07-14
Prevents wand of door creation working on indestructible walls

hal9000PR:
## What Does This PR Do
Prevents wand of door creation working on indestructible walls

## Why It's Good For The Game
You cannot use the wand in the wizards layer however you could use it with an ERT shuttle to exploit and steal admin things. exploits not good

## Changelog
:cl:
fix: door creation wands respect indestructible walls.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/magic.dm
+++ b/code/modules/projectiles/projectile/magic.dm
@@ -130,9 +130,11 @@
 	. = ..()
 	var/atom/T = target.loc
 	if(isturf(target) && target.density)
-		CreateDoor(target)
+		if(!(istype(target, /turf/simulated/wall/indestructible)))
+			CreateDoor(target)
 	else if(isturf(T) && T.density)
-		CreateDoor(T)
+		if(!(istype(T, /turf/simulated/wall/indestructible)))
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile/magic.dm
```diff
@@ -130,9 +130,11 @@
 	. = ..()
 	var/atom/T = target.loc
 	if(isturf(target) && target.density)
-		CreateDoor(target)
+		if(!(istype(target, /turf/simulated/wall/indestructible)))
+			CreateDoor(target)
 	else if(isturf(T) && T.density)
-		CreateDoor(T)
+		if(!(istype(T, /turf/simulated/wall/indestructible)))
+			CreateDoor(T)
```
AffectedArc07, line 137: I hate how this check looks redundant but is actually important.

AffectedArc07 (approved):

code/modules/projectiles/projectile/magic.dm
hal9000PR: this is exactly what I thought

PR #16370, opened 2021-07-15
Adds RCD and RPD to Cryo Storage; Fixes Cryo Storage Logic

### LINKED ISSUE #16347: Non-replaceable roundstart items are lost when going to cryo
**Issue Description**:

Certain exclusive, non-obtainable items do not go into cryo storage when the holder goes to cryo.

Items such as the .38 Mars Special (detective's revolver,) Forensic Gloves, and possibly various other items are retrievable from cryo storage after the person who had them went to cryo. 

**What happened instead**:

These items are not in the storage log and are not output when the "retrieve all items" option is chosen.

Certain items, which may be vital or highly useful for the user, are completely lost and cannot be replaced. Such cases include the detective's forensic gloves, which are necessary for their work.

1. Be roundstart detective
2. Pick up revolver, forensic gloves, etc. and keep them on your person
3. Go to cryo without putting them back
4. Late-join detective doesn't have their usual tools

Not sure; probably forever

**Extra information**:

Although my personal experience with this issue is mostly with the detective's belongings, it is likely that this is an issue with other objects as well.

LiberatedWaveMan:
## What Does This PR Do

Adds the RCD, RPD, .38 Mars Special (Detective's revolver), and Forensic Gloves to the cryo "preserve" list so that they aren't obliterated when the user forgets to drop them before cryoing. Fixes #16347 
## Why It's Good For The Game

RCDs and RPDs are incredibly useful items which are inconvenient for engineering to lose, and the detective's revolver and gloves are literally irreplaceable if the roundstart detective goes to cryo with them.

## Changelog
:cl:
add: Adds the RCD, RPD, .38 Mars Special, and Forensics Gloves to the cryo storage "preserve" list.
:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -346,7 +346,7 @@
 
 
 /obj/machinery/cryopod/proc/handle_contents(obj/item/I)
-	if(I.contents.len) //Make sure we catch anything not handled by qdel() on the items.
+	if(length(I.contents)) //Make sure we catch anything not handled by qdel() on the items.
 		if(should_preserve_item(I) != CRYO_DESTROY) // Don't remove the contents of things that need preservation
 			return
 		for(var/obj/item/O in I.contents)
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -257,7 +257,7 @@
 		/obj/item/clothing/gloves/color/black/forensics
 		/obj/item/rcd
 		/obj/item/rpd
-		/obj/item/gun/projectile/revolver/detective
+		/obj/item/gun/projectile/revolver/detective,
 	)
 	// These items will NOT be preserved
 	var/list/do_not_preserve_items = list (
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -257,7 +257,8 @@
 		/obj/item/clothing/gloves/color/black/forensics,
 		/obj/item/rcd,
 		/obj/item/rpd,
-		/obj/item/gun/projectile/revolver/detective
+		/obj/item/gun/projectile/revolver/detective,
+        /obj/item/clothing/accessory/holster/armpit
 	)
 	// These items will NOT be preserved
 	var/list/do_not_preserve_items = list (
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -257,7 +257,7 @@
 		/obj/item/clothing/gloves/color/black/forensics
 		/obj/item/rcd
 		/obj/item/rpd
-		/obj/item/gun/projectile/revolver/detective,
+		/obj/item/gun/projectile/revolver/detective
 	)
 	// These items will NOT be preserved
 	var/list/do_not_preserve_items = list (
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -253,10 +253,10 @@
 		/obj/item/key,
 		/obj/item/door_remote,
 		/obj/item/autopsy_scanner,
-		/obj/item/holosign_creator/atmos
-		/obj/item/clothing/gloves/color/black/forensics
-		/obj/item/rcd
-		/obj/item/rpd
+		/obj/item/holosign_creator/atmos,
+		/obj/item/clothing/gloves/color/black/forensics,
+		/obj/item/rcd,
+		/obj/item/rpd,
 		/obj/item/gun/projectile/revolver/detective
 	)
 	// These items will NOT be preserved
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -256,9 +256,7 @@
 		/obj/item/holosign_creator/atmos,
 		/obj/item/clothing/gloves/color/black/forensics,
 		/obj/item/rcd,
-		/obj/item/rpd,
-		/obj/item/gun/projectile/revolver/detective,
-        /obj/item/clothing/accessory/holster/armpit
+		/obj/item/rpd
 	)
 	// These items will NOT be preserved
 	var/list/do_not_preserve_items = list (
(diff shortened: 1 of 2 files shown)


Necaladun:
Poor detectives and engineers, hate it when these things disappear. I approve!

code/game/machinery/cryopod.dm
```diff
@@ -253,7 +253,12 @@
 		/obj/item/key,
 		/obj/item/door_remote,
 		/obj/item/autopsy_scanner,
-		/obj/item/holosign_creator/atmos
+		/obj/item/holosign_creator/atmos,
+		/obj/item/clothing/gloves/color/black/forensics,
+		/obj/item/rcd,
+		/obj/item/rpd,
+		/obj/item/gun/projectile/revolver/detective,
```
hal9000PR, line 260: Looking at line 240, `/obj/item/gun` is already protected. Not sure if this is necessary. However, what is might be good is putting shoulder holsters here instead. If you cryo with a gun in a shoulder holster I am *pretty* sure it gets lost.

Wojz:
Add shoulder holsters to the list. They get deleted. **HoS's gun fits in them and I think that deletes it.** (Don't quote me on this though)

dearmochi (approved):

code/game/machinery/cryopod.dm
LiberatedWaveMan: This was already noted to me by steelslayer, and was updated in the most recent commit (I think, I'm not entirely sure of what I'm doing yet)

SteelSlayer: I should mention that even with holsters in this list, I'm not sure they will get found and stored if they're attached to a jumpsuit. I can't recall the cryopod code, and I can't look into this at the moment.

code/game/machinery/cryopod.dm
```diff
@@ -253,7 +253,12 @@
...
+        /obj/item/clothing/accessory/holster/armpit
```
FlattestGuitar, line 261: Adding holsters here isn't the right fix. There will be more issues with other items if cryo isn't working properly. It works for guns inside boxes, right? (or does it?)

Fix the code that deletes the contents of holsters. 
Look at `holster.dm:52`, make sure that `holstered.loc` is set properly and that the gun ends up in the holster's `contents`. 
Make sure that the code that makes use of this big list here is properly searching the contents of all objs that get thrown into cryo.

LiberatedWaveMan: Sadly, this is my first time working with DM code, and I have nowhere near an idea of how that kind of stuff works. If anyone would like to help me out, however, I will gladly test the changes and commit them to this PR.

code/game/machinery/cryopod.dm
```diff
@@ -341,21 +344,23 @@
...
-// Also make sure there is a valid control computer
+
+/obj/machinery/cryopod/proc/handle_contents(obj/item/I)
+	if(I.contents.len) //Make sure we catch anything not handled by qdel() on the items.
+		if(should_preserve_item(I) != CRYO_DESTROY) // Don't remove the contents of things that need preservation
+			return
+		for(var/obj/item/O in I.contents)
+			if(istype(O, /obj/item/tank)) //Stop eating pockets, you fuck!
+				continue
+			handle_contents(O)
```
FlattestGuitar, line 355: This line is the only real change to the logic. We just check the entire contents tree now.

Kyep (approved):

code/game/machinery/cryopod.dm
```diff
@@ -341,21 +344,23 @@
 			return CRYO_PRESERVE
 	return CRYO_DESTROY
 
-// This function can not be undone; do not call this unless you are sure
-// Also make sure there is a valid control computer
+
+/obj/machinery/cryopod/proc/handle_contents(obj/item/I)
+	if(I.contents.len) //Make sure we catch anything not handled by qdel() on the items.
```
AffectedArc07, line 349:
```suggestion
	if(length(I.contents)) //Make sure we catch anything not handled by qdel() on the items.
```

AffectedArc07 (changes requested):

SabreML (changes requested):
Currently, if the detective's revolver (or any other gun) is inside a holster when someone cryoes, the gun just will be deleted.
Even if it's not actually in the holster anymore, the gun will still be set as the `holstered` variable on the holster.
https://github.com/ParadiseSS13/Paradise/blob/f9d6300bc9cdb034fbe1196f14e2c10eaa5bb86c/code/modules/clothing/under/accessories/holster.dm#L12-L14
This means that the cryopod will follow this chain if someone cryoes with a holstered gun:
1. Search through their items until the holster, and therefore the gun, is found.
2. Guns are a preserved item, so `forceMove()` it out of the way into the cryopod.
3. Further down in `despawn_occupant()` (L377-L379), the holster is qdeleted.
4. `qdel()` calls `Destroy()` on the holster, also deleting the revolver since it's still set as a reference.
https://github.com/ParadiseSS13/Paradise/blob/f9d6300bc9cdb034fbe1196f14e2c10eaa5bb86c/code/game/machinery/cryopod.dm#L377-L379

The best way to solve this as far as I can tell would be to change the `Destroy()` code in [`code\modules\clothing\under\accessories\holster.dm`](https://github.com/ParadiseSS13/Paradise/blob/f9d6300bc9cdb034fbe1196f14e2c10eaa5bb86c/code/modules/clothing/under/accessories/holster.dm#L12-L14) so that it just drops the holstered gun rather than deleting it.
Specifically, to this or something similar:
```dm
/obj/item/clothing/accessory/holster/Destroy()
	if(holstered?.loc == src) // Gun still in the holster
		holstered.forceMove(loc)
	holstered = null
	return ..()
```
In other words:
```
if(The `holstered` variable is still set, and the holstered item is inside the holster)
	Move the item to the holster's location
Set the holstered variable to null either way, to avoid deletion issues
return and delete as normal
```
This'll make the gun properly get stored inside the cryopod, and it'll allow it to be taken out again afterwards.
<hr>
Also, if you put `Fixes #16347` in the PR description, the issue will automatically be closed when this is merged.

SabreML (approved):
RCDs, RPDs, Forensics gloves and the Detective's revolver all seem to be getting stored properly as intended, so LGTM. 👍

LiberatedWaveMan:
Since I'm in a working mood today, does anyone have suggestions for (unreplaceable) items I should add? If not I'll be finalizing this

AffectedArc07 (approved):

Issue #16360, opened 2021-07-14
No breath trait on ashwalkers conflicting with newcrit, paralyising them forever without finishing them off


shadowcrimes:
**Problem Description**:

On one round, I was playing ashwalker alongside others, one of the ashwalkers was sent into crit from damage, I resusitated them using a sleeper and trauma kits, fixing all of their damage. however newcrit cardac arrest kicked in, eventally resulting in them reaching the final stage as I was treating their damage and pumping them full of epi and saline.

However they took no damage at all from it, and due to ashwalkers not actually needing to breath at all, were completely rendered unable to die alongside being effectively paralysed forever from the stun of cardac arrest, they didn't fall unconcious either resulting in me having to toolbox them to death so the trendil would free them of their suffering.

**What did you expect to happen**: 

I expected them to die of cardac arrest like normal races do, instead of being perma stunned and paralysed on the floor.

**What happened instead**:

Cardac prison of suffering, that made them unable to move, but left them able to talk completely fine, with zero effects of cardac complete with no damage. Nothing on hand at lavaland was able to fix it either, outside of just flat out finishing the job and killing them myself.

**Why is this bad/What are the consequences:**

Being trapped in a state of unable to die but unable to act is not fun, let alone the fact ashwalkers lack compact defibs or anything to remove the cardac arrest status, it can result in otherwise fully healed ashwalkers being stunlocked to the floor.

It also doesn't knock them unconcious so you have to outright finish them off and kill them, not sure if can be succumbed out of, but its not fun to experience. especally when someone goes through all the effort of patching you up to just be completely paralysed from something that should kill you and nothing but having someone else killing you anyway can be done about it.

Trapping someone in a stasis and taking all control away from them is not good and can ruin people's rounds quite easily, even without intention to do so it can happen with this issue.

**Steps to reproduce the problem**:

1: Bring ashwalker to crit.
2: fix all brute and burn and toxin damage so they don't die from that alone.
3: have the cardac arrest status reach its final stage while healing them so they don't die of other damage.
4: obverse functional immortality and permanent suffering combined, you monster.

PR #16372, opened 2021-07-15
Give ashwalkers a handheld defib

chafla:
## What Does This PR Do

Addresses #16360 by giving ash-walkers a handheld defib in their medical crate.

## Why It's Good For The Game

If ash-walkers can still get cardiac arrest, let's at least give them a way to treat it. Handheld defibs don't revive, so this shouldn't have any balance implications, either.

## Changelog
:cl:
add: Handheld defib to ash-walker nest.
/:cl:

### DIFF
--- a/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_ash_walker1.dmm
+++ b/_maps/map_files/RandomRuins/LavaRuins/lavaland_surface_ash_walker1.dmm
@@ -195,6 +195,7 @@
 /obj/item/stack/sheet/cloth/ten,
 /obj/item/stack/medical/splint,
 /obj/item/storage/firstaid/adv,
+/obj/item/handheld_defibrillator,
 /turf/simulated/floor/plating/asteroid/basalt/lava_land_surface,
 /area/ruin/unpowered/ash_walkers)
 "aF" = (


AzuleUtama:
Ash walkers are meant to be using primitive technology due to lavaland being their home - a handheld defib unit is far too technologically advanced for them to know how to use and shouldn't be something they start out with as it wouldn't make any sense.

The primitive restrictions do mean they'll be unable to perform certain tasks or activities - healing crit being among them, but that's meant to be part of the point of playing them.

SabreML:
Not sure about this approach personally, since Ash Walkers are meant to have the whole 'primitive tribal' thing going on (in theory) and them knowing how to use a defibrillator goes against that.
It'd probably be better to instead make cardiac arrest deal damage in some other way if oxyloss isn't an option, or something along those lines.

LiberatedWaveMan:
The main two problems I'm seeing with this are 
1. The above issue that they're probably too primitive to know how a defibrillator works
and 2. This doesn't actually fix the issue of ashwalkers being crit-locked, and also relies _entirely_ upon them having someone who can defib them, which is never a given.

chafla:
Yeah, I hear you. I mostly wanted to provide a quick fix because it was pretty easy to do so and because I don't want people to be stuck stunlocked while we wait for a fix.

Some people in the discord brought up good ideas on making a defib more lore-based by making a new lightning-rod item, or just making cardiac arrest directly inflict brain damage on ashwalkers (and kill that way), rather than rely on oxyloss.

Anyway, I'm gonna close this. Thanks for the feedback!

PR #15458, opened 2021-02-07
Re-adds engineering items to secure storage and main engineering storage

Cocacolagua:
## What Does This PR Do
Re-adds six fifty-metal stacks, four fifty-glass stacks, two fifty-wood stacks, two engineering suit storages and four floodlights to secure storage.
(Half of the metal and glass was placed in engineering main storage to not overload the crate)

## Why It's Good For The Game
All of these items were removed from secure storage with the addition of the supermatter, i'm re-adding given that these are quite important for engineering, being it constructing anything else, repairing breaches or lighting a place up without having to build lights.

## Images of changes

Before:
naNgAwRhe6

After:
OHgk3sJOzO

## Changelog
:cl:
add: Adds three fifty-metal sheets stacks and two fifty-glass sheets stacks to engineering main storage.
add: Adds three fifty-metal sheets stacks, two fifty-glass sheets stacks, two fifty-wood planks stacks, two engineering suit storages and four floodlights to engineering secure storage.
/:cl:

PR #16374, opened 2021-07-15
Fully fixes the issues introduced in #15458

S34NW:
## What Does This PR Do
Properly fixes the map issue that was reverted in #16373 `Revert "Re-adds engineering items to secure storage and main engineering storage (#15458)"` 

tl;dr: Sometimes TGM map formats decide they want to break entirely


S34NW:
Can we file this PR under _"Sean is literally a monkey with a keyboard who swears he knows how to map but it really doesn't show"_?

PR #16383, opened 2021-07-17
Allows players to name seed packs to keep track of modified plants

theo-3:
## What Does This PR Do
Adds a 'variant' var to seed packs which can be set by players with a pen or with the dna modifier. This variable is shown in the pack's name, in the seed extractor, in the dna modifier and in hydroponic trays' examine. Extracting seed packs from named seeds will keep the name.

## Why It's Good For The Game
Helps keep track of what modified plants do, for example a tomato plant labeled 'healing' and another labeled 'acid', or a gaia plant labeled 'perrenial densified chemicals'.

## Images of changes

The gif doesn't show the dna modifier UI because I messed up the window capture. Sorry.
Capture1
Capture2
Capture1

## Changelog
:cl:
add: Adds a 'variant' var to seed packets which can be set by using a pen on seeds or a tray, or from the dna modifier. It shows up in the packet's name, in plant trays examine, in the seed extractor, in the dna modifier and in harvested plants. This is intended to help keep track of what modified plants do, such as wheat labeled 'healing'.
/:cl:

### DIFF
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -230,13 +230,6 @@
 		var/can_insert = disk && disk.gene && disk.gene.can_add(seed)
 		var/can_extract = disk && !disk.read_only
 
-		dat += "<div class='line'><h3>Variant</h3></div><div class='statusDisplay'><table>"
-		dat += "<tr><td width='260px'>[(seed.variant == "") ? "None" : seed.variant]</td>"
-		dat += "<td><a href='?src=[UID()];set_v=1'>Edit</a></td>"
-		if(seed.variant != "")
-			dat += "<td><a href='?src=[UID()];del_v=1'>Remove</a></td>"
-		dat += "</tr></table></div>"
-
 		dat += "<div class='line'><h3>Core Genes</h3></div><div class='statusDisplay'><table>"
 		for(var/a in core_genes)
 			var/datum/plant_gene/G = a
@@ -284,6 +277,13 @@
 			if(can_insert && istype(disk.gene, /datum/plant_gene/trait))
 				dat += "<a href='?src=[UID()];op=insert'>Insert: [disk.gene.get_name()]</a>"
 			dat += "</div>"
+
+		dat += "<div class='line'><h3>Variant</h3></div><div class='statusDisplay'><table>"
+		dat += "<tr><td width='260px'>[(seed.variant == "") ? "None" : seed.variant]</td>"
+		dat += "<td><a href='?src=[UID()];set_v=1'>Edit</a></td>"
+		if(seed.variant != "")
+			dat += "<td><a href='?src=[UID()];del_v=1'>Remove</a></td>"
+		dat += "</tr></table></div>"
 	else
 		dat += "<br>No sample found.<br><span class='highlight'>Please, insert a plant sample to use this device.</span>"
 	popup.set_content(dat)
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -279,9 +279,9 @@
 			dat += "</div>"
 
 		dat += "<div class='line'><h3>Variant</h3></div><div class='statusDisplay'><table>"
-		dat += "<tr><td width='260px'>[(seed.variant == "") ? "None" : seed.variant]</td>"
+		dat += "<tr><td width='260px'>[seed.variant ? seed.variant : "None"]</td>"
 		dat += "<td><a href='?src=[UID()];set_v=1'>Edit</a></td>"
-		if(seed.variant != "")
+		if(seed.variant)
 			dat += "<td><a href='?src=[UID()];del_v=1'>Remove</a></td>"
 		dat += "</tr></table></div>"
 	else
@@ -405,7 +405,7 @@
 	else if(href_list["del_v"])
 		if(!seed)
 			return
-		seed.variant = ""
+		seed.variant = null
 		seed.apply_variant_name()
 
 	interact(usr)
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -401,7 +401,7 @@
 	else if(href_list["set_v"])
 		if(!seed)
 			return
-		var/V = sanitize(stripped_input(usr, "Choose variant name:", "Plant Variant Naming", seed.variant, MAX_MESSAGE_LEN))
+		var/V = sanitize(stripped_input(usr, "Choose variant name:", "Plant Variant Naming", seed.variant, 64))
 		if((!Adjacent(usr)) || (!seed))
 			return
 		seed.variant = V
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -401,11 +401,7 @@
 	else if(href_list["set_v"])
 		if(!seed)
 			return
-		var/V = input(usr, "Choose variant name:", "Plant Variant Naming", seed.variant) as text|null
-		if(!issilicon(usr) && !Adjacent(usr) || isnull(V) || !seed)
-			return
-		seed.variant = copytext(sanitize(html_encode(trim(V))), 1, 64)
-		seed.apply_variant_name()
+		seed.variant_prompt(usr, src)
 	else if(href_list["del_v"])
 		if(!seed)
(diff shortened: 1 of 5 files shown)


code/modules/hydroponics/gene_modder.dm
```diff
@@ -391,6 +398,19 @@
 	else if(href_list["abort"])
 		operation = ""
 		target = null
+	else if(href_list["set_v"])
+		if(!seed)
+			return
+		var/V = sanitize(stripped_input(usr, "Choose variant name:", "Plant Variant Naming", seed.variant, MAX_MESSAGE_LEN))
```
dearmochi, line 404: Might want to use a max length lower than 1024, or you could end up with absurdly long names that fill up the chat. Perhaps 64 characters is saner.

code/modules/hydroponics/seeds.dm
```diff
@@ -317,16 +320,32 @@
...
-	if (istype(O, /obj/item/plant_analyzer))
+	if(istype(O, /obj/item/plant_analyzer))
 		to_chat(user, "<span class='info'>*---------*\n This is \a <span class='name'>[src]</span>.</span>")
 		var/text = get_analyzer_text()
 		if(text)
 			to_chat(user, "<span class='notice'>[text]</span>")
 
 		return
+	if(istype(O, /obj/item/pen))
+		var/V = sanitize(stripped_input(user, "Choose variant name:", "Plant Variant Naming", variant, MAX_MESSAGE_LEN))
```
dearmochi, line 331: Ditto

code/modules/hydroponics/seeds.dm
```diff
@@ -317,16 +320,32 @@
...
+		if(!Adjacent(user)) //prevent remote naming of seeds
+			return
+		if(V == "" && variant != "")
+			user.show_message("<span class='info'>You remove the variant designation from the [plantname].</span>")
+		else if(V != "")
+			user.show_message("<span class='info'>You designate the [plantname] as the [V] variant.</span>")
+		variant = V
+		apply_variant_name()
```
dearmochi, line 339:
```suggestion
		apply_variant_name()
		return
```

So it doesn't proceed to the default behaviour

code/modules/hydroponics/seeds.dm
```diff
@@ -317,16 +320,32 @@
...
 	..() // Fallthrough to item/attackby() so that bags can pick seeds up
 
-
+/obj/item/seeds/proc/apply_variant_name()
+	var/P = findtext(name, "\[")
+	var/V = (variant == "") ? "" : (" \[" + variant + "]") // If we have a non-empty variant add it to the name
+	if (P != 0)
+		name = copytext(name, 1, P - 1) + V // If there was already a variant in the name remove it
```
dearmochi, line 346: Consider the `initial(varname)` proc to get the initial name of a datum, would be better than text manipulation

dearmochi (changes requested):

code/modules/hydroponics/gene_modder.dm
theo-3: Thanks, I copy pasted that from another file and didn't know what the max len did

code/modules/hydroponics/seeds.dm
theo-3: Wouldn't that delete the 'experimental' from modified seeds?

dearmochi: Good point, this is fine then

theo-3:
Thanks for the help!

dearmochi (approved):

theo-3:
I improved the apply name proc. It uses initial(name) and checks if 'experimental' was there, and also reapplies labels. This also means it won't get confused if you had a label like '([][])' present before setting a variant, which would have messed up the name before.

Wojz:
Good idea.

cluwneythescourge:
man, I've been waiting for something like this forever.  Does the modified name show on the actual plant?   Like, if you grow a "healing" tomato seed, are the plants labeled?

theo-3:
The actual tomatoes are not labeled, since I thought it'd be weird to see a label on plants like that, but I could make it show up there too

theo-3:
There, I added the label to growns too

theo-3:
They're separate in plant bags now too, just tested it

theo-3:
How can I tell whether a player pressed cancel or entered an empty string? Right now pressing cancel deletes the variant instead of doing nothing

dearmochi:
> How can I tell whether a player pressed cancel or entered an empty string? Right now pressing cancel deletes the variant instead of doing nothing

Check if the result of the input is `null` or otherwise empty.

theo-3:
Yeah, I tried using !V, but that prevents a player from deleting a variant by entering an empty string. How do I tell the difference between canceling and typing an empty string?

dearmochi:
Cancelling specifically returns `null` while an empty string is, well, an empty string.

```dm
if(isnull(V))
	// Cancelled
else if(!length(V))
	// Empty string
```

theo-3:
Huh. There's a difference between isnull(V) and V == null?

theo-3:
`var/V = stripped_input(user, "Choose variant name:", "Plant Variant Naming", variant, 64)
		if((!Adjacent(user)) || isnull(V))
			[return`](url)
This isn't returning, pressing cancel deletes the variant

theo-3:
Capture
This isn't working. Pressing cancel still deletes the variant

theo-3:
Yep, it works now. The weird thing is I look at the pandemic machine and it doesn't do this. It just checks for null text without delaying sanitize.

theo-3:
Nevermind, the pandemic machine just makes unlabeled bottles if you cancel

theo-3:
There, now cancel works right

theo-3:
I moved the variant name prompt to a proc to remove code reuse from the gene modder and added naming to trays too. I'm using an adj parameter to get the proc to call Adjacent on the correct object. Is that the right way to do that?

Kyep (approved):
While I am fine with the concept, I wonder if this requires TGUI. Including raw HTML seems... not great.

theo-3:
> While I am fine with the concept, I wonder if this requires TGUI. Including raw HTML seems... not great.

All botany machines use raw HTML, don't they?

code/modules/hydroponics/grown.dm
```diff
@@ -42,6 +42,8 @@
 		seed.prepare_result(src)
 		transform *= TRANSFORM_USING_VARIABLE(seed.potency, 100) + 0.5 //Makes the resulting produce's sprite larger or smaller based on potency!
 		add_juice()
+		if(seed.variant && seed.variant != "")
```
AffectedArc07, line 45: A string of `""` evalutes as false.
image
The second check here (and all other places) is redundant.

AffectedArc07 (changes requested):

AffectedArc07:
> While I am fine with the concept, I wonder if this requires TGUI. Including raw HTML seems... not great.

I am not going to enforce a TGUI conversion in this PR since it is outside the scope.

code/modules/hydroponics/gene_modder.dm
```diff
@@ -391,6 +398,15 @@
...
 		operation = ""
 		target = null
+	else if(href_list["set_v"])
+		if(!seed)
+			return
+		seed.variant_prompt(usr, src)
+	else if(href_list["del_v"])
+		if(!seed)
+			return
+		seed.variant = ""
```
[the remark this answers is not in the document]
theo-3: Alright, I replaced all empty strings with null. Also fixed a bunch of == "" that wouldn't find null

code/modules/hydroponics/grown.dm
theo-3: Done

theo-3:
Uh. How did this break the tests?

theo-3:
Is this because I force-changed the commit message?

theo-3:
Let's try this again

theo-3:
Alright, I tested it locally and it seems to work. @AffectedArc07 Is it good now?

theo-3:
Quick question: is the stripped_input function supposed to convert nulls to empty strings? It makes it impossible to tell the difference between canceling and empty strings. I tested this with the pandemic machine, which seems to check for cancel too (`if(name == null || wait)`), and canceling creates a bottle named 'culture bottle' instead of doing nothing.

variableundefined (approved):

PR #16385, opened 2021-07-17
[DNM] Reagent ID Code Refactor

SabreML:
## What Does This PR Do

Removes all instances of the reagent `id` variable in favour of just using the typepath instead.

## Why It's Good For The Game

I know I'm not the only one who's got annoyed at some point or another by the current ID system for reagents.
For example, if you wanted to check how much damage exactly `mob.add_reagent("plasma", 5)` would do over time, or if you wanted to make sure that `"robustharvestnutriment"` is the correct ID spelling, the only way to find out would be to search for it in vscode and scroll down until you find right file.
image
This doesn't take too long to do of course, but having to do it *every* time you want to find a reagent can get pretty irritating over time.

For comparison, this is how it's done using the proposed system:
IlcZD8gIa4

The only downside that I can see to using typepaths rather than strings is the code length, but I personally believe that the upsides more than make up for that.

## Changelog
:cl:
spellcheck: Fixed the names of some reagents like 'phlogiston' not being capitalised.
experiment: Refactored all reagent code to use typepaths instead of IDs.
/:cl:


Fox-McCloud:
You really should have contacted me ahead of time before doing this.

Aurorablade:
On another note useing a shorthand or an alias to find a ID rather then a coder making sure they have the right PATH correctly in is more preferable. I understand that looking through may be an inconvenience to some but having a shorthand pointer to a thing is alot better then having to reuse a path..every..dang..time. and paths are subject to change at any point, so if the path changes? Guess what the person changing it is going to have to do.

SabreML:
> You really should have contacted me ahead of time before doing this.

@Fox-McCloud I definitely would have asked a maintainer beforehand if this was a feature PR or something considering the size of it, but as it's a refactor I didn't see a need to get specific prior approval.

I also mentioned it a bunch on #coding_chat since I started and the response seemed fully positive. https://discord.com/channels/145533722026967040/145700319819464704/865344903809597450
If you're referring to the 'Major Chemistry Changes' part on your #changes-wanted post, then I again hadn't considered that as an issue since it doesn't have any ingame changes at all, just code related. Personally I had taken the post to mean "Porting Fermichem" more than "Refactoring the code".
Worst case scenario I've lost 3 days of work, but I've also written a lot of notes about the surrounding code so it wouldn't be too bad.

<hr>

> On another note useing a shorthand or an alias to find a ID rather then a coder making sure they have the right PATH correctly in is more preferable. I understand that looking through may be an inconvenience to some but having a shorthand pointer to a thing is alot better then having to reuse a path..every..dang..time.

@Aurorablade I disagree here, although I suppose that's a given since I made the pull request.
I genuinely can't think of a single benefit to using an arbritrary string assigned to a datum rather than the actual datum itself. 
As I mentioned in the description the only downside I can see is having to type the entire typepath when referencing the reagent, and personally I consider that far less of an inconvenience than having to guess the ID and hope you didn't make a typo.
I spotted a surprising amount of typos as I was doing this actually, so it's not an unfounded concern.

The VSCode language client also means that this also adds autocomplete and ctrl+click 'Go to definition' support as well, and any typos will be caught by the compiler.
 
> and paths are subject to change at any point, so if the path changes? Guess what the person changing it is going to have to do.

That's... exactly the same as in the current system? And with everything else in the codebase for that matter.
If someone changes a typepath or identifier then generally speaking they'll have to change all other instances of that too.

Fox-McCloud:
A refactor this large and involved, especially with something that's been marked as such should be contacted about head of time, it's not something to approach willy-nilly, especially when someone is a maintainer who specializes in a particular aspect of code.

It'd be like refactoring the garbage controller without asking me about it, directly.

Refactor or not, you still should have asked--especially when there's over 150 PR's to begin with.

SabreML:
> A refactor this large and involved, especially with something that's been marked as such should be contacted about head of time, it's not something to approach willy-nilly, especially when someone is a maintainer who specializes in a particular aspect of code.
> 
> It'd be like refactoring the garbage controller without asking me about it, directly.
> 
> Refactor or not, you still should have asked--especially when there's over 150 PR's to begin with.

Well, I apologise for not asking for permission then, I hadn't known about your association with reagent code. In hindsight it *is* listed in `CODEOWNERS` but I don't exactly check that file frequently. Nobody mentioned it either when I brought up the PR idea a few days ago, but that may have been under the assumption that I had asked beforehand.

To be honest actually, I wasn't aware you were contactable at *all*. From what I understood you're busy with things in real life at the moment and don't have the time to answer SS13 related questions, as indicated by your Discord status always being offline.

SabreML:
Having gone through and tested every file now, I'm going to mark this as ready for review.
I've inevitably missed some things, but as far as I can tell everything seems to be working properly. (I definitely recommend a testmerge anyway though, if this ever gets that far)

SabreML:
After the disaster that was #16520, every one of my open PRs was set to "Do Not Merge" until I could go through each of them individually and double check that I didn't miss anything stupid.

All of them except for this and #16533 have now been unset, and I was planning do these two at some point this week.
Unfortunately, due to college and some other personal issues, I don't currently have the time nor the motivation to test each of the 386 edited files for the fourth time, as well as every possible interaction and edge case with the new code.

I will get back to these at some point, but right now it's better to just close both in the interests of getting them out of the way.

PR #16389, opened 2021-07-18
Lets revivers potentially heal you out of shock

Qwertytoforty:
## What Does This PR Do

Updates revivers description from ""revive you if you lose consciousness" to "heal you out of critical condition"

Makes revivers have a 15% chance per every other life tick to heal shock, so long as you are out of crit.

## Why It's Good For The Game

Revivers kinda suck right now. Very slow healing, kills you if emp'd, but most importantly, does not heal shock, which means if you receive shock, and have a reviver, you still will progress to stage 2 crit, and from there to heart attack, without medical attention soon. 

This PR makes the reviver better at stabilizing you, by making it potentially able to heal shock (but not stage 2 crit) if you are not longer in critical condition. This PR does not adjust healing values, or make it heal you to 25 HP as a way to cure stage 1 crit, however those could be options in potential future PRs. This just makes reviver that little bit better for stabilizing you so long as you are not in a dangerous situation.

## Changelog
:cl:
tweak: Reviver implants now stabilize the user better, by having a chance to heal shock when the user is not in critical condition.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/augments_internal.dm
+++ b/code/modules/surgery/organs/augments_internal.dm
@@ -336,7 +336,7 @@
 
 /obj/item/organ/internal/cyberimp/chest/reviver
 	name = "Reviver implant"
-	desc = "This implant will attempt to revive you if you lose consciousness. For the faint of heart!"
+	desc = "This implant will attempt to heal you out of critical condition. For the faint of heart!"
 	icon_state = "chest_implant"
 	implant_color = "#AD0000"
 	origin_tech = "materials=5;programming=4;biotech=4"
@@ -361,6 +361,11 @@
 			addtimer(CALLBACK(src, .proc/heal), 30)
 		else
 			reviving = FALSE
+			if(owner.HasDisease(new /datum/disease/critical/shock(0)) && prob(15)) //If they are no longer in crit, but have shock, and pass a 15% chance:
+				for(var/datum/disease/critical/shock/S in owner.viruses)
+					S.cure()
+					revive_cost += 150
+					to_chat(owner, "<span class='notice'>You feel better.</span>")
 			return
 	cooldown = revive_cost + world.time
 	revive_cost = 0


Wojz:
Step in the right direction I feel. You can't **really** abuse the revivers, they make you die to something quite easy to get.

code/modules/surgery/organs/augments_internal.dm
```diff
@@ -361,6 +361,11 @@
 			addtimer(CALLBACK(src, .proc/heal), 30)
 		else
 			reviving = FALSE
+			if(owner.HasDisease(new /datum/disease/critical/shock(0)) && prob(15)) //If they are no longer in crit, but have shock, and pass a 15% chance:
```
dearmochi, line 364: I HATE DISEASE CODE

dearmochi (approved):

code/modules/surgery/organs/augments_internal.dm
SabreML, line 364: So... the only reason this spawns a new `/datum/disease` is so that it can call `IsSame()` on it.
https://github.com/ParadiseSS13/Paradise/blob/c5e593b2bfc2444b13cc877c2e47592866be5cb1/code/datums/diseases/_MobProcs.dm#L2-L7
`IsSame()` of course, being nothing but a wrapper for `istype()`.
https://github.com/ParadiseSS13/Paradise/blob/c5e593b2bfc2444b13cc877c2e47592866be5cb1/code/datums/diseases/_disease.dm#L155-L158

Realistically it's probably out of scope of this PR, but if you could refactor `HasDisease()` to just check the type normally instead of... whatever it's doing currently, it'd only add a few extra changed lines and I'd be very grateful 🙏.

All seems to work fine otherwise, so 👍.

SabreML (approved):

Kyep (approved):

variableundefined (approved):

PR #16390, opened 2021-07-18
Brig cell door timer fix

SabreML:
## What Does This PR Do

Fixes this bug by removing all of the hardcoded pixel offsets on brig cell door timers:
image

Also tweaks the positioning of the Meta and Delta cell timers for consistency, and fixes some incorrectly linked lockers and flashes.

## Why It's Good For The Game

Subtypes are better than map instances in most cases, but for pixel offsets it makes a lot more sense for them to be individually edited rather than set in the code directly. Hardcoded pixel offsets on an object is just bad practice when it comes to mapping and standardisation.

### Metastation timer location change:
**Before:**
before
**After:**
after

## Changelog
:cl:
tweak: Tweaked the location of the cell door timers on Metastation.
fix: Fixed the cell 5 door timer floating in the doorway on Boxstation.
fix: Fixed some cell lockers and flashes not being correctly linked on Metastation and Deltastation.
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -36071,7 +36071,7 @@
 	dir = 1
 	},
 /obj/structure/closet/secure_closet/brig{
-	id = "Cell 4";
+	id = "Cell 2";
 	name = "Cell 2 Locker"
 	},
 /obj/machinery/atmospherics/unary/vent_pump/on,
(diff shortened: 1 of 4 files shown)


SabreML:
Just realised that the extra fixes puts this PR in all three map queues at once, so I can move the Cyberiad timer fix to a separate PR if that's requested.

Kyep (approved):

AffectedArc07 (approved):

PR #16392, opened 2021-07-18
Makes cyborgs have a dust animation, allows admins to dust smite anything

Qwertytoforty:
## What Does This PR Do

Makes cyborgs dusting call the cyborg dust animation proc.
Moves the dust smite from purely human to any mob, as any mob can be dusted.

## Why It's Good For The Game

Borgs playing dust animation is good, even if they only ever really dust when the station is nuked.

Being able to smite anything with a dust to delete it is good, especially with borgs, as deleting a borg leaves the MMI on the ground, which can be inconvenient if you need to get rid of a borg fast, for one reason or another.

## Images of changes

Old sprite, already in code, but:

dust

## Changelog
:cl:
tweak: Any mob can be smited with dust, rather than just humans.
fix: Fixes cyborgs not playing their dust animation
/:cl:

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -2180,7 +2180,6 @@
 			ptypes += "Crew Traitor"
 			ptypes += "Floor Cluwne"
 			ptypes += "Shamebrero"
-			ptypes += "Dust"
 		var/punishment = input(owner, "How would you like to smite [M]?", "Its good to be baaaad...", "") as null|anything in ptypes
 		if(!(punishment in ptypes))
 			return
@@ -2203,6 +2202,9 @@
 			if("Gib")
 				M.gib(FALSE)
 				logmsg = "gibbed."
+			if("Dust")
+				M.dust()
+				logmsg = "dust"
 
 			// These smiting types are only valid for ishuman() mobs
 			if("Brain Damage")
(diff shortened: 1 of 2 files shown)


code/modules/admin/topic.dm
```diff
@@ -2180,7 +2180,6 @@
 			ptypes += "Crew Traitor"
 			ptypes += "Floor Cluwne"
 			ptypes += "Shamebrero"
-			ptypes += "Dust"
```
SabreML, line 2183: I think you might've forgot to add this back to line 2168, cause right now it's not able to be picked on anything at all.
https://github.com/ParadiseSS13/Paradise/blob/29acd36a95db76952fc78554403512727d7dc9ed/code/modules/admin/topic.dm#L2168

code/modules/mob/living/silicon/robot/death.dm
```diff
@@ -31,6 +31,7 @@
 	canmove = 0
 	icon = null
 	invisibility = 101
```
SabreML, line 33: Not caused by this PR, but the `icon` being set to null here makes the ghost of the borg after dusting completely invisible.
Removing this line fixes the issue, so I'm honestly not sure why it was here in the first place.
```suggestion
	canmove = 0
	invisibility = 101
```

code/modules/admin/topic.dm
Qwertytoforty: Whoops

SabreML (approved):

Kyep (approved):

AffectedArc07 (approved):

PR #16401, opened 2021-07-20
Hierophant trophy wall duration nerf 

Peakwastaken:
## What Does This PR Do

Changes the duration of the Vortex Talisman's wall on mark detonation from 7.5 seconds to 1.5 seconds. 

It's still possible to have a wall up continuously (maximum two overlapping for a very short time), but you can't spam 5+ walls to box fauna/blobbers in anymore, and not being fast enough might leave you totally exposed. 

## Why It's Good For The Game

A lot of people want Hiero back on lavaland because the fight is very fun, however, with such a long wall duration, this trophy renders most lavaland fights nigh brainless - spare Bubblegum - which will indirectly impact the station even if the staff is nerfed to not work there anymore. This aims to reign in that cheese a bunch, hopefully allowing Hiero to be re-added while being healthier for the game overall.

## Changelog
:cl:

tweak: Reduced Vortex Talisman's duration from 7.5 to 1.5s

/:cl:

### DIFF
--- a/code/modules/mining/equipment/kinetic_crusher.dm
+++ b/code/modules/mining/equipment/kinetic_crusher.dm
@@ -443,4 +443,4 @@
 		new /obj/effect/temp_visual/hierophant/wall/crusher(otherT, user)
 
 /obj/effect/temp_visual/hierophant/wall/crusher
-	duration = 15
+	duration = 1.5 SECONDS
--- a/code/modules/mining/equipment/kinetic_crusher.dm
+++ b/code/modules/mining/equipment/kinetic_crusher.dm
@@ -443,4 +443,4 @@
 		new /obj/effect/temp_visual/hierophant/wall/crusher(otherT, user)
 
 /obj/effect/temp_visual/hierophant/wall/crusher
-	duration = 75
+	duration = 15


code/modules/mining/equipment/kinetic_crusher.dm
```diff
@@ -443,4 +443,4 @@
 		new /obj/effect/temp_visual/hierophant/wall/crusher(otherT, user)
 
 /obj/effect/temp_visual/hierophant/wall/crusher
-	duration = 75
+	duration = 15
```
hal9000PR, line 446:
```suggestion
	duration = 1.5 SECONDS
```

hal9000PR (approved):

Wojz:
Good PR. Doesn't change anything for skilled KC users, makes it require skill.

SabreML (approved):
Code's all working fine, but I do worry that 1.5 seconds might be slightly too far in the other direction.

The default cooldown for the crusher *also* being 1.5 seconds (plus the 0.8 second cooldown on all actions after a melee attack) means that the barrier won't be doing much to help if you're not constantly in melee range spam clicking the enemy, somewhat defeating the point of having a barrier in the first place.

Peakwastaken:
> The default cooldown for the crusher _also_ being 1.5 seconds (plus the 0.8 second cooldown on all actions after a melee attack) means that the barrier won't be doing much to help if you're not constantly in melee range spam clicking the enemy, somewhat defeating the point of having a barrier in the first place.

Appreciate the review !

That is not taking in consideration the legion skull trophy, as well these two cooldowns being mostly stackable - shoot, detonate immediately as you move back in, back off behind the wall, shoot, rinse and repeat - You can definitely keep a constant wall while dipping behind it. 

Obviously, the most impacted - and targeted - megafauna for this nerf is the colossus, which is the one that this trophy would otherwise render the most brainless. With this change, it just means that the timing of execution is tighter, that you can't try and time to dodge fauna attacks (meaning that you might get melee clipped here and there even with good execution, but with megafauna's melees not ignoring all armor now, and the BDM trophy, it's really not that much of an issue), and that you can't just heal up between marks without backing off significantly/having to dodge some attacks.

I tested it pretty thoroughly, 2.5 seconds already felt almost as good as the original 7.5 (doesn't matter if you can keep 1 or four walls up constantly), 2 was marginally better, and, even with the 1.5 seconds I settled on, I could kill the colossus very reliably, often taking very little damage, sometimes none. 

It's definitely a tight execution, and a MAJOR nerf - but I wholeheartedly believe that it is not too heavy handed. Yes, if you miss a crusher shot/ fail to trigger a mark in your cycle you WILL immediately be exposed, and potentially in trouble, but most megafauna will mostly be immobile or going in a straight line behind you when you fight them anyways - and I don't think clicking an immobile target is too much of a skill check to ask for killing megafaunas.

variableundefined:
No head or maint approval after 30+ days. Closing.

PR #16403, opened 2021-07-20
Slight nerf and tweak to the Chiropteran Screech vampire ability

Peakwastaken:
## What Does This PR Do

- Bowman headsets now reduce the screech's effect, negating the stun, but it is now also capable of breaking light bulbs (slightly further than windows).

- fixes a small grammatical issue in "aN ear piercing shriek".

## Why It's Good For The Game

With earmuffs being fixed, the screech has become downright broken against security : Either they all deafen themselves - kind of going against any and all form of RP and teamplay - or they get an entire platoon downed at the press of a single button.

This aims to tune it down, while still being an extremely competent form of crowd control against civillian crew, and getting the extra benefit of breaking lights ! (Which kiiiinda makes sense seeing as it breaks reinforced windows like paper :eyes:)

## Changelog
:cl:
tweak: Being screeched at while wearing a bowman headset now confuses rather than stun.
add: Screech now breaks lights in a six tile radius
fix: fixed a grammatical error in "You hear aN ear piercing shriek and your senses dull!"
/:cl:

### DIFF
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -265,6 +265,8 @@
 	for(var/mob/living/carbon/C in hearers(4))
 		if(C == user)
 			continue
+		if(!affects(C))
+			continue
 		if(ishuman(C))
 			var/mob/living/carbon/human/H = C
 			if(H.check_ear_prot() >= HEARING_PROTECTION_TOTAL)
@@ -275,8 +277,6 @@
 				C.Stuttering(10)
 				C.Jitter(50)
 				continue
-		if(!affects(C))
-			continue
 		to_chat(C, "<span class='warning'><font size='3'><b>You hear an ear piercing shriek and your senses dull!</font></b></span>")
 		C.Weaken(4)
 		C.AdjustEarDamage(0, 20)
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -286,7 +286,7 @@
 	for(var/obj/structure/window/W in view(4))
 		W.deconstruct(FALSE)
 	for(var/obj/machinery/light/L in view(6))
-		L.deconstruct(FALSE)
+		L.break_light_tube()
 	playsound(user.loc, 'sound/effects/creepyshriek.ogg', 100, 1)
 
 
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -272,10 +272,10 @@
 			if(H.check_ear_prot() >= HEARING_PROTECTION_TOTAL)
 				continue
 			if(H.check_ear_prot() >= HEARING_PROTECTION_MINOR)
-				to_chat(C, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
-				C.Confused(10)
-				C.Stuttering(10)
-				C.Jitter(50)
+				to_chat(H, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
+				H.Confused(10)
+				H.Stuttering(10)
+				H.Jitter(50)
 				continue
 		to_chat(C, "<span class='warning'><font size='3'><b>You hear an ear piercing shriek and your senses dull!</font></b></span>")
 		C.Weaken(4)
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -269,9 +269,15 @@
 			var/mob/living/carbon/human/H = C
 			if(H.check_ear_prot() >= HEARING_PROTECTION_TOTAL)
 				continue
+			if(H.check_ear_prot() >= HEARING_PROTECTION_MINOR)
+				to_chat(C, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
+				C.Confused(10)
+				C.Stuttering(10)
+				C.Jitter(50)
+				continue
 		if(!affects(C))
 			continue
-		to_chat(C, "<span class='warning'><font size='3'><b>You hear a ear piercing shriek and your senses dull!</font></b></span>")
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -269,16 +269,24 @@
...
 			if(H.check_ear_prot() >= HEARING_PROTECTION_TOTAL)
 				continue
+			if(H.check_ear_prot() >= HEARING_PROTECTION_MINOR)
+				to_chat(C, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
+				C.Confused(10)
+				C.Stuttering(10)
+				C.Jitter(50)
+				continue
 		if(!affects(C))
 			continue
```
hal9000PR, line 279: you want the `if(!affects(C))` check to come before this confuse bit. if not it will confuse other vampires, which is probably not what you want.

Peakwastaken: Done, thanks !

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -265,20 +265,28 @@
...
 				continue
-		if(!affects(C))
-			continue
-		to_chat(C, "<span class='warning'><font size='3'><b>You hear a ear piercing shriek and your senses dull!</font></b></span>")
+			if(H.check_ear_prot() >= HEARING_PROTECTION_MINOR)
+				to_chat(C, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
+				C.Confused(10)
+				C.Stuttering(10)
+				C.Jitter(50)
+				continue
```
SteelSlayer, line 279: Unnecessary continue, the loop will continue regardless.
```suggestion
```

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -265,20 +265,28 @@
...
+		to_chat(C, "<span class='warning'><font size='3'><b>You hear an ear piercing shriek and your senses dull!</font></b></span>")
 		C.Weaken(4)
 		C.AdjustEarDamage(0, 20)
 		C.Stuttering(20)
 		C.Stun(4)
 		C.Jitter(150)
 	for(var/obj/structure/window/W in view(4))
 		W.deconstruct(FALSE)
+	for(var/obj/machinery/light/L in view(6))
+		L.deconstruct(FALSE)
```
SteelSlayer, line 289: Another opinion, but I think that instead of the whole light being destroyed, only the bulb should be shattered, using the `break_light_tube()` proc.

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -265,20 +265,28 @@
...
+			continue
 		if(ishuman(C))
 			var/mob/living/carbon/human/H = C
 			if(H.check_ear_prot() >= HEARING_PROTECTION_TOTAL)
 				continue
-		if(!affects(C))
-			continue
-		to_chat(C, "<span class='warning'><font size='3'><b>You hear a ear piercing shriek and your senses dull!</font></b></span>")
+			if(H.check_ear_prot() >= HEARING_PROTECTION_MINOR)
+				to_chat(C, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
```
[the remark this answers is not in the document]
Peakwastaken: This one should only show when you get screamed at with a bowman headset, the other one when you're bare-eared.

code/game/gamemodes/vampire/vampire_powers.dm
Peakwastaken, line 279: From my understanding, the continue breaks the loop and skips the rest of the effects off the screech - I tested it, and indeed removing the continue will get someone with a bowman stunned.

code/game/gamemodes/vampire/vampire_powers.dm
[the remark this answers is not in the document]
SteelSlayer: You're right I misread the structure of the loop. Dismissing my other comment about the continue as well.

Peakwastaken:
Agreed, switched to the suggest light-smashing method- Thanks Steel, much more elegant !

code/game/gamemodes/vampire/vampire_powers.dm
Peakwastaken: Agreed, switched to the suggest light-smashing method- Thanks Steel, much more elegant !

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -265,20 +265,28 @@
...
+				C.Confused(10)
+				C.Stuttering(10)
+				C.Jitter(50)
```
SteelSlayer, line 278: Nitpick, but use `H` here to be clear the mob is a human.
```suggestion
				to_chat(H, "<span class='warning'><font size='3'><b>You hear a dull shriek and instinctively cover your ears!</font></b></span>")
				H.Confused(10)
				H.Stuttering(10)
				H.Jitter(50)
```

Wojz:
I like this. The screech is blatantly overpowered in its current state, and the lights breaking is fun, and possibly actually useful.

code/game/gamemodes/vampire/vampire_powers.dm
Peakwastaken: To my understanding, C stands for carbon - and H is used for human in the ear check. 

Switching C for H meant I had to remove `var/mob/living/carbon/human/H = C`, and broke the code at line 272.

SteelSlayer: Do not remove the `var/mob/living/carbon/human/H = C` line. That is required.

SabreML: What Steel means here is that, since you've casted `C` to `H` just above, it makes more sense in terms of readability if all of these subsequent procs also use `H` where they can.
image
vs
image
You should be able to just commit the suggested changes to do it.

SteelSlayer (approved):

Varghh:
Si

variableundefined:
No headmin or maint approval after 30+ days. Closing.

PR #16404, opened 2021-07-21
Fix floor painters being able to paint non-floor

dearmochi:
## What Does This PR Do

Adds a missing early return that would prevent the painting of non-floor atoms.

## Why It's Good For The Game

Bugfix, prevents items and structures from being turned invisible.

## Changelog
:cl:
fix: Floor painters can no longer paint non-floor
/:cl:


code/game/objects/items/devices/painter/floor_painter.dm
```diff
@@ -23,6 +23,7 @@
 /datum/painter/floor/paint_atom(atom/target, mob/user)
 	if(!istype(target, /turf/simulated/floor/plasteel))
 		to_chat(user, "<span class='warning'>[holder] can only be used on station flooring.</span>")
+		return
```
AffectedArc07, line 26: The fact this slipped past me is upsetting.

AffectedArc07 (approved):

PR #16408, opened 2021-07-22
Delta Disposals - Fore Section

Vi3trice:
## What Does This PR Do
This is the start of fixing up Delta's disposal network. The various sorting pipes aren't set for Paradise's current list of tagger locations (It even had a pipe for Theater, which does not exist in the list.) This also cleans up where the pipes tend to go, as they had a bad tendency of going under walls when they don't need to, or seem to be laid out with limited types of junctions in mind.

## Why It's Good For The Game

Makes things go where they should, and if something breaks, you won't have to rebuild a wall in the process (usually). Map bloat is also reduced.

## Changelog
:cl:
del: Removed redundant or unused sorting junctions
tweak: Reworked the layout of the pipes. Moved two disposal bins around in Cargo.
/:cl:


S34NW (changes requested):
The disposal unit in the mining dock is not connected to the pipe network.
Apart from that, this looks fine.

_maps/map_files/Delta/delta.dmm
```diff
@@ -25638,6 +25432,7 @@
 "bfC" = (
 /obj/machinery/atmospherics/pipe/simple/hidden/scrubbers,
 /obj/effect/decal/warning_stripes/yellow,
+/obj/machinery/disposal,
```
S34NW, line 25435: This would fix it

S34NW (approved):
LGTM

S34NW (approved):

variableundefined (approved):

PR #16422, opened 2021-07-25
Immolator lasers now immolate all mobs

Qwertytoforty:
## What Does This PR Do

Changes the check on immolators from carbon mobs to living mobs.

## Why It's Good For The Game

Blobbernaughts and cyborgs can light on fire, why not have immolators work on them too, especially when we have the code for catching on fire on the base living mob, if we want other animals to catch on fire.

## Changelog
:cl:

tweak: Immolators can now light all flammable mobs on fire, rather than just carbon mobs
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -142,7 +142,7 @@
 
 /obj/item/projectile/beam/immolator/on_hit(atom/target, blocked = 0)
 	. = ..()
-	if(istype(target, /mob/living/))
+	if(isliving(target)
 		var/mob/living/M = target
 		M.adjust_fire_stacks(1)
 		M.IgniteMob()
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -142,7 +142,7 @@
 
 /obj/item/projectile/beam/immolator/on_hit(atom/target, blocked = 0)
 	. = ..()
-	if(isliving(target)
+	if(isliving(target))
 		var/mob/living/M = target
 		M.adjust_fire_stacks(1)
 		M.IgniteMob()
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -142,8 +142,8 @@
 
 /obj/item/projectile/beam/immolator/on_hit(atom/target, blocked = 0)
 	. = ..()
-	if(istype(target, /mob/living/carbon))
-		var/mob/living/carbon/M = target
+	if(istype(target, /mob/living/))
+		var/mob/living/M = target
 		M.adjust_fire_stacks(1)
 		M.IgniteMob()
 


code/modules/projectiles/projectile/beams.dm
```diff
@@ -142,8 +142,8 @@
 
 /obj/item/projectile/beam/immolator/on_hit(atom/target, blocked = 0)
 	. = ..()
-	if(istype(target, /mob/living/carbon))
-		var/mob/living/carbon/M = target
+	if(istype(target, /mob/living/))
```
SabreML, line 145: May as well:
```suggestion
	if(isliving(target)
```

SabreML: Crap sorry, needs to be `if(isliving(target))`.

SabreML (approved):

Fox-McCloud:
Not sure if this is really a very good idea--most interactions with fire are hotspots, so a mob being on fire from this is not a particularly big deal.

That said, being able to set a mob on fire with a gun like this, even when not in a fire, may likely introduce issues with being able to put out said fire.

Wojz:
Blobbernauts can already easily deal with fire by staying on blob tiles - hell, they can weaponize it and I already set myself on fire on purpose as a blobbernaut.

Not sure what this really means for borgs if anything, but many modules are capable of putting themselves out.

Kyep (commented):
Can you explain the implications of this?
E.g:
- can non-carbon mobs resist to extinguish themselves?
- can they tell they are on fire? is there a different sprite effect?
- can OTHER PEOPLE tell they are on fire?
- how much damage do they take?
- how much damage do carbon mobs who walk into them take?
- overall, is this more dangerous to the non-carbons set on fire, or to the carbons they walk into afterwards?

Qwertytoforty:
> 
> 
> Can you explain the implications of this?
> E.g:
> 
>     * can non-carbon mobs resist to extinguish themselves?
> 
>     * can they tell they are on fire? is there a different sprite effect?
> 
>     * can OTHER PEOPLE tell they are on fire?
> 
>     * how much damage do they take?
> 
>     * how much damage do carbon mobs who walk into them take?
> 
>     * overall, is this more dangerous to the non-carbons set on fire, or to the carbons they walk into afterwards?

I will clarify, which I did not make clear in the pull request, it does not change which mobs can catch fire. Currently, only cyborgs, carbon mobs, and simple animals with the can_be_on_fire flag can catch fire. This is blob spores and blobernaughts. As such, currently this laser will now light those mobs on fire, which can already catch fire by other methods, but not animals like spiders or cows or mice. This change purely makes the laser make more sense, by lighting anything that catches on fire, rather than just carbon mobs.

1: They can not resist to extinguish themself. Fire wears off at a rate of 0.1 fire stacks per cycle (20 seconds from 1 immolator), however blobernaughts can remove 1 stack of fire each cycle while on blob. Otherwise, being hit with water, or going into an area without oxygen puts out the fire.
2: They get the status effect of fire like a person would get, and get a visual burn effect.
3: Other people can see the fire effect on the mob.
4: Blobbernaughts / blob spores take 3 burn damage per cycle while on fire. Other simple mobs currently can not catch fire at the moment, as we have this disabled. If we enabled this for a mob, they would take 2 burn per second by default.
5: Carbon mobs that walk on them catch fire, which will translates to 5 burn per cycle.
6: On one hand, the mob takes less damage than the carbon if the carbon is not fireproof. On the other hand, simple mobs will often find the fire harder to deal with then humans, and find being on fire purposefully not a good idea. In any case, if they want to catch fire on purpose from an immolator, they need to find someone using a weapon, and purposefully get hit by lasers, which is already going to hurt them. Most blobbernaughts are not going to take 20-40 burn, plus burn damage, in order to light others on fire. A welderbomb is a much better trade for them, in that regard.

variableundefined (approved):

PR #16426, opened 2021-07-25
Makes the 'Sentience' event show the mob which has been selected

SabreML:
## What Does This PR Do

Makes the 'Sentience' event alert always show what simple animal has been selected in the alert popup image and tooltip.

## Why It's Good For The Game

I don't have any actual data to back this up of course, but personally I *very* rarely take the event in case I get stuck in a maintenance tunnel as a mouse, or something like that. It being randomised could make for some fun surprises of course, but in this case I believe that actually letting the player know what they're signing up for is far better than forcing them to hope for good RNG.

This also makes it more consistent with the other ways that one can become a simple mob, namely Slime Sentience Potions and admemery.

## Images of changes

**Before:**
before alert
before text

**After:**
after alert
after text

## Changelog
:cl:
tweak: The 'Simple mob sentience' event now shows the mob which has been selected.
/:cl:


Kyep (approved):

code/modules/events/sentience.dm
```diff
@@ -4,46 +4,46 @@
...
-	SA.can_collar = 1
+	SA.can_collar = TRUE
 	SA.maxHealth = max(SA.maxHealth, 200)
 	SA.health = SA.maxHealth
 	SA.del_on_death = FALSE
 	greet_sentient(SA)
+
+	var/sentience_report = "<font size=3><b>[command_name()] Medium-Priority Update</b></font>"
+
+	var/data = pick("scans from our long-range sensors", "our sophisticated probabilistic models", "our omnipotence", "the communications traffic on your station", "energy emissions we detected", "\[REDACTED]", "Steve")
```
SteelSlayer, line 37: Isn't the `[REDACTED]` now going to display as `[REDACTED`?

dearmochi: Tested with Gearbot and it seems fine

dearmochi (approved):

code/modules/events/sentience.dm
SabreML: Yeah, you only need to escape the opening bracket. It makes more sense maybe with both being escaped, but VSCode's syntax highlighting doesn't like it:
image

AffectedArc07 (approved):

PR #16427, opened 2021-07-26
Fixes Oversight With Loadout in #16132

Leanfrog:
## What Does This PR Do
This PR removes the ability for Pod Pilots to choose the security bomber jacket in the loadout as they already start with it, oversight on my part, whoops

## Why It's Good For The Game
Pod pilots don't need to start with an extra copy of their jacket they already start with

## Changelog
:cl:
tweak: Sec Pod Pilots can no longer receive an extra sec bomber jacket if it is selected in loadout
/:cl:

### DIFF
--- a/code/modules/client/preference/loadout/loadout_suit.dm
+++ b/code/modules/client/preference/loadout/loadout_suit.dm
@@ -111,7 +111,7 @@
 /datum/gear/suit/secbomberjacket
 	display_name = "security bomber jacket"
 	path = /obj/item/clothing/suit/jacket/pilot
-	allowed_roles = list("Head of Security", "Warden", "Detective", "Security Officer", "Security Pod Pilot")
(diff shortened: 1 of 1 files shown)


SabreML (approved):

dearmochi (approved):

AffectedArc07:
Honestly I dislike this.

There's lots of jobs where you can select something in the loadout that you already get. This is just an exclusion for the sake of it.

Also keep in mind that there's only one of these, it's on their table not their bag, so a second pod pilot spawning after the first won't have one.



TLDR: Not in favour

Spacemanspark:
I don't particularly see the harm in the pod pilot being able to have a second one tbh. Hell, one could argue that it makes sense for people to carry an extra pair of clothes around, just in case.

Wojz:
So you're telling me I can start with THREE bomber jackets?!

variableundefined:
Objection for same reason as Affected. Closing.

PR #16433, opened 2021-07-26
Removes table prefixes

AffectedArc07:
## What Does This PR Do
Removes the prefixed schema and all the ingame support for it

Why?

Because maintaining two database schemas is a pain, not to mention we never use the prefixed schema in production, nor would we have any reason to. Things live in different databases for a reason, and you shouldn't have 2 separate applications share the same DB. Thats scuffed.

This PR also makes the default schema name be `paradise_gamedb` instead of `feedback`, because calling the DB `feedback` despite the fact it stores everything, not just feedback, is absolutely stupid.

## Why It's Good For The Game
See reasoning above

## Changelog
N/A

## Note
While this changes the schema, it does not constitute an SQL change as no table modifications were made, just a rename of the DB itself, which I already did in production when I tookover hosting.


AffectedArc07:
Testmerging this on live to ensure nothing breaks

Kyep (approved):
Agreed, these changes are a good idea.

dearmochi (approved):

variableundefined (approved):
I've always wondered why the database is called feedback

PR #16436, opened 2021-07-26
adds a handshaking interaction.dm

Frex-Creator:
more actions

## What Does This PR Do
adds the action of shaking hands with other players (just add this action, pat and hug are still there)

## Why It's Good For The Game
It gives us more variety in actions, thus allowing a more fluid role, increasing the quality of the game, and character relationships.

## Images of changes

image

## Changelog
:cl:
add: Added a new help intent action to shake someone's hand when targeting their arms or hands.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -251,7 +251,7 @@
 					"<span class='notice'>You pat [src] on the head.</span>",\
 					)
 				else
-					if(M.zone_selected == ("r_arm" || "r_hand" || "l_arm" || "l_hand"))
+					if(M.zone_selected in list("r_arm", "r_hand", "l_arm", "l_hand"))
 						M.visible_message(\
 						"<span class='notice'>[M] shake [src] hand.</span>",\
 						"<span class='notice'>You shake hands with [src].</span>",\
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -253,8 +253,8 @@
 				else
 					if(M.zone_selected == ("r_arm" || "r_hand" || "l_arm" || "l_hand"))
 						M.visible_message(\
-						"<span class='notice'>[M] [src] gives you a handshake.</span>",\
-						"<span class='notice'>You shakes [src] hand.</span>",\
+						"<span class='notice'>[M] [src] shakes your hand.</span>",\
+						"<span class='notice'>You shake hands with [src].</span>",\
 						)
 					else
 						M.visible_message(\
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -253,7 +253,7 @@
 				else
 					if(M.zone_selected in list("r_arm", "r_hand", "l_arm", "l_hand"))
 						M.visible_message(\
-						"<span class='notice'>[M] shake [src] hand.</span>",\
+						"<span class='notice'>[M] shakes [src]'s hand.</span>",\
 						"<span class='notice'>You shake hands with [src].</span>",\
 						)
 					else
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -259,7 +259,7 @@
 					else
 						M.visible_message(\
 						"<span class='notice'>[M] gives [src] a [pick("hug","warm embrace")].</span>",\
-						"<span class='notice'>[M] You hug [src].</span>",\
+						"<span class='notice'>You hug [src].</span>",\
 						)
 						if(ishuman(src))
 							var/mob/living/carbon/human/H = src
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -251,17 +251,40 @@
 					"<span class='notice'>You pat [src] on the head.</span>",\
 					)
 				else
-
-					M.visible_message(\
-					"<span class='notice'>[M] gives [src] a [pick("hug","warm embrace")].</span>",\
-					"<span class='notice'>You hug [src].</span>",\
-					)
-					if(ishuman(src))
-						var/mob/living/carbon/human/H = src
-						if(H.wear_suit)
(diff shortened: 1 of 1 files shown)


variableundefined:
You need a much more descriptive description and names.

variableundefined (commented):
Objection. 

Removes the hug action in favor of a massive, horrendous else if chain of copy paste code. 

Name and description extremely vague without explaining why you're removing hug.

Cocacolagua:
> Objection.
> 
> Removes the hug action in favor of a massive, horrendous else if chain of copy paste code.
> 
> Name and description extremely vague without explaining why you're removing hug.

It doesn't remove the hug, but move it after the massive else if spam.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -251,17 +251,40 @@
...
-					"<span class='notice'>[M] gives [src] a [pick("hug","warm embrace")].</span>",\
-					"<span class='notice'>You hug [src].</span>",\
-					)
-					if(ishuman(src))
-						var/mob/living/carbon/human/H = src
-						if(H.wear_suit)
-							H.wear_suit.add_fingerprint(M)
-						else if(H.w_uniform)
-							H.w_uniform.add_fingerprint(M)
+					if(M.zone_selected == "r_arm")
```
Cocacolagua, line 254:
```suggestion
					if(M.zone_selected == ("r_arm" || "l_arm" || "r_hand" || "l_hand"))
```
you can simplify all bodypart reactions by doing this

Frex-Creator: Thanks, I hadn't thought of it that way. I had another way to place the "handshakes" but I don't know if it would work in the same way

PacifistDalek:
The grammar shown in the example is quite awkward, it should be "you shake the farwa's hand". "shakes the hand" should only be used when viewing someone else doing the emote

Frex-Creator:
Thank you and sorry for those mistakes, I don't know more than a few words of English (I don't know how to make sentences)

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -251,17 +251,22 @@
...
-					"<span class='notice'>[M] gives [src] a [pick("hug","warm embrace")].</span>",\
-					"<span class='notice'>You hug [src].</span>",\
-					)
-					if(ishuman(src))
-						var/mob/living/carbon/human/H = src
-						if(H.wear_suit)
-							H.wear_suit.add_fingerprint(M)
-						else if(H.w_uniform)
-							H.w_uniform.add_fingerprint(M)
+					if(M.zone_selected == ("r_arm" || "r_hand" || "l_arm" || "l_hand"))
```
farie82, line 254: This wont work as expected. It will be the same as `M.zone_selected == "r_arm"`. A switch can do this but here you need an is in list check or if else chain

SabreML: This'll work for Farie's suggestion:
```suggestion
					if(M.zone_selected in list("r_arm", "r_hand", "l_arm", "l_hand"))
```

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -251,17 +251,22 @@
...
+						M.visible_message(\
+						"<span class='notice'>[M] shake [src] hand.</span>",\
+						"<span class='notice'>You shake hands with [src].</span>",\
+						)
```
SabreML, line 258: Should be `shakes` rather than `shake`, and `[src]'s` is better gramatically than just `[src]` for this:
```suggestion
						M.visible_message(\
						"<span class='notice'>[M] shakes [src]'s hand.</span>",\
						"<span class='notice'>You shake hands with [src].</span>",\
						)
```

code/modules/mob/living/carbon/carbon.dm
SabreML, line 254: This whole thing can be made a little neater if you move the `if()` onto the same line as the `else` before it, and indent the rest of the code accordingly. Like this:
```dm
if(M.zone_selected == "head")
	M.visible_message(\
	"<span class='notice'>[M] pats [src] on the head.</span>",\
	"<span class='notice'>You pat [src] on the head.</span>",\
	)
else if(M.zone_selected in list("r_arm", "r_hand", "l_arm", "l_hand"))
	M.visible_message(\
	"<span class='notice'>[M] shakes [src]'s hand.</span>",\
	"<span class='notice'>You shake hands with [src].</span>",\
	)
else
	M.visible_message(\
	"<span class='notice'>[M] gives [src] a [pick("hug","warm embrace")].</span>",\
	"<span class='notice'>[M] You hug [src].</span>",\
	)
```

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -251,17 +251,22 @@
...
+					else
+						M.visible_message(\
+						"<span class='notice'>[M] gives [src] a [pick("hug","warm embrace")].</span>",\
+						"<span class='notice'>[M] You hug [src].</span>",\
+						)
```
SabreML, line 263: The formatting of the arguments is very odd here, but it's also like this in the rest of the proc too and this is your first PR here, so I think it's fine to stay this way for now.

SabreML (commented):
I like the idea of adding handshaking as an interaction like this, so good first PR! 👍
The code and PR just need a little bit of tweaking before they're properly ready.
<hr>

For the PR, you need to change the title to be something more descriptive than "Update carbon.dm". If it stays like that then anyone who's just browsing the PR list on github won't understand what you're trying to edit.
You can edit the title of a PR by clicking this button here, then changing the text, and clicking 'Save':
image
image
image
It doesn't need to be "Adds a handshaking interaction" specifically of course, but just something which better explains what you're trying to add in the PR.
<hr>

Same with the changelog as well, you've put that you're adding a new help intent action, but not *what* that action is. Since the changelog is the only thing that most people see from the PR, they'll be very confused if that shows up ingame.
I'd recommend changing it from this:
```
## Changelog
:cl:
add: Added new help intent action
/:cl:
```
To something like this instead:
```
## Changelog
:cl:
add: Added a new help intent action to shake someone's hand when targeting their arms or hands.
/:cl:
```

PR #16440, opened 2021-07-27
IAA locker, spare headsets, and general locker cleanup

SabreML:
## What Does This PR Do

1. Adds an 'Internal affairs locker' to the Law Office, containing a full IAA uniform, a spare (non-bowman) headset, and all of the non-cosmetic items from the other locker.
2. Adds a spare Magistrate headset to the Magistrate's locker.
3. Changes the SOP bookcases in the NTR and IAA offices into subtypes containing the books by default, rather than them being map edited onto a regular bookcase.
4. Moves some map edited locker items out of the map and into the subtype they were placed on instead, for future-proofing.

## Why It's Good For The Game

1. I've seen a suprisingly high number of occasions where an antagonist goes after an IAA for their gear, steals their headset for sec comms, and then said IAA is left without access to the procedure channel for the remainder of the shift.
2. See above, although it's far less common for Magistrates.
3. Makes a lot more sense as a subtype than two identical piles of books placed on a bookcase.
4. If the contents need to be edited again in the future, it won't require map edits.

## Images of changes

**New IAA Locker:**
Paradise Test NSS Cyberiad 2021-07-27 133606
**IAA Locker Contents:**
Desktop Screenshot 2021 07 27 - 13 58 58 90

## Changelog
:cl:
add: Added an 'Internal affairs locker' to the Law Office, containing a full set of IAA gear and a spare headset.
add: Added a spare headset to the Magistrate's locker.
/:cl:


WingedYordle:
I am not sure this is the right PR to comment... are you able to remove the console in law office as IAA do not have access to it what so ever?

SabreML:
> I am not sure this is the right PR to comment... are you able to remove the console in law office as IAA do not have access to it what so ever?

The console's used for printing out arrest records at the moment so it's probably still good to have something there, but replacing it with something that IAAs can actually use would be a good idea.
I'll see if I can find a sprite for a new "records printer" machine or something.

SabreML:
Update: *(May have forgotten about this for a few days...)*
I've not been able to find any existing sprites that would work for it, and my attempts at making something myself didn't really end up looking like much of anything. If anyone knows of something that would work then I can use that, but otherwise I'll probably just leave the console as it is in this PR.

SabreML:
All working as intended, except for a missing flask in the HoS locker which I've now replaced.
Unsetting from DNM.

PR #16443, opened 2021-07-27
Cleans up the TODOs I left all over the codebase

AffectedArc07:
## What Does This PR Do
- Removes `command_name()`. It always returned `NAS Trurl` and there was no point in it
- Moved teleport list locations to SSmapping instead of globals
- Moved IPIntel procs to their own subsystem
- Makes karma logging use RUSTG
- Shows a "Round Time" stat in the statpane that shows the actual round time without the lobby offset

## Why It's Good For The Game
- Removes an unneeded proc
- Less shitcode
- Less shitcode
- Less shitcode
- Its nice to see a round time without the lobby offset

## Changelog
:cl: AffectedArc07
add: Added a counter for the real round time without the lobby offset
/:cl:


dearmochi (approved):

code/controllers/subsystem/ipintel.dm
```diff
@@ -3,13 +3,374 @@ SUBSYSTEM_DEF(ipintel)
...
+	if(intel <= GLOB.configuration.ipintel.bad_rating)
+		if(world.realtime < cacherealtime + (GLOB.configuration.ipintel.hours_save_good HOURS))
+			return TRUE
+	else
+		if(world.realtime < cacherealtime + (GLOB.configuration.ipintel.hours_save_bad HOURS))
+			return TRUE
+
+
+
+/**
```
SabreML, line 50: I can't comment on the rest of this since file since I have no idea what any of it does, but are the three spaces between each of these procs intentional? The standard usually seems to be 1-2.

code/game/machinery/computer/store.dm
```diff
@@ -31,7 +31,7 @@
 	var/dat = {"
 <html>
 	<head>
-		<title>[command_name()] Merchandise</title>
+		<title>Nanotrasen Merchandise</title>
 		<style type="text/css">
```
SabreML, line 35: The two in this file were replaced with "Nanotrasen" rather than "NAS Trurl"?
I mean it probably works better like that, but these seem to be the only things changed to this.

code/modules/mob/mob.dm
```diff
@@ -1009,7 +1009,8 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 
 // this function displays the station time in the status panel
 /mob/proc/show_stat_station_time()
-	stat(null, "Round Time: [worldtime2text()]") // AA TODO: Make this do "Game Time" and "Round Time" with the ROUND_TIME macro
+	stat(null, "Server Uptime: [worldtime2text()]")
+	stat(null, "Round Time: [ROUND_TIME ? time2text(ROUND_TIME, "hh:mm:ss") : "N/A"]")
 	stat(null, "Station Time: [station_time_timestamp()]")
```
SabreML, line 1014: This would be because I'm in BST currently rather than GMT, but the 'Server Uptime' on my end is underflowing to 23 hours at roundstart:
image
Might not need to be fixed since I know the actual server is on ISO 8601/UTC, but I figured I'd point this out anyway.

AffectedArc07: Thats because of daylight savings. BYOND breaks with that enabled. It should work fine in prod.

code/game/machinery/computer/store.dm
AffectedArc07: Intentional change. It makes more sense here, however in other cases (mainly command reports), NAS Trurl makes sense

code/controllers/subsystem/ipintel.dm
AffectedArc07: I wanted more separation, especially when filling in my documentation

variableundefined (approved):

PR #16447, opened 2021-07-28
Ports item hover outline

S34NW:
## What Does This PR Do

Ports tgstation/tgstation#57574 (That itself is a BeeStation port) that adds item hover outlines to stuff in your equipment slots and in inventories!
Supports themes and there is a toggle pref if you hate these sorts of new features.

## Why It's Good For The Game
Who doesn't want clearer feedback on what item you currently have your mouse over?

## Changelog
:cl:
add: Items in your inventory now have an outline when you hover over them. This can be disabled in Game Preferences.
/:cl:


code/game/objects/items.dm
```diff
@@ -418,10 +422,12 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 
 // called when this item is removed from a storage item, which is passed on as S. The loc variable is already set to the new destination before this is called.
 /obj/item/proc/on_exit_storage(obj/item/storage/S as obj)
+	in_inventory = FALSE
```
S34NW, line 425:
```suggestion
	in_storage = FALSE
```

code/game/objects/items.dm
```diff
@@ -418,10 +422,12 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
...
 	return
 
 // called when this item is added into a storage item, which is passed on as S. The loc variable is already set to the storage item.
 /obj/item/proc/on_enter_storage(obj/item/storage/S as obj)
+	in_inventory = TRUE
```
S34NW, line 430:
```suggestion
	in_storage = TRUE
```

S34NW (commented):
totally forgot to change these after I added `in_storage`

code/game/objects/items.dm
```diff
@@ -691,14 +697,22 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
...
 		var/timedelay = 8
 		var/user = usr
 		tip_timer = addtimer(CALLBACK(src, .proc/openTip, location, control, params, user), timedelay, TIMER_STOPPABLE)
+		if(!QDELETED(src))
+			var/mob/living/L = usr
+			if(usr.client.prefs.toggles2 & PREFTOGGLE_2_SEE_ITEM_OUTLINES)
+				if(istype(L) && L.incapacitated(ignore_lying = TRUE))
+					apply_outline(COLOR_RED_GRAY) //if they're dead or handcuffed, let's show the outline as red to indicate that they can't interact with that right now
+				else
+					apply_outline() //if the player's alive and well we send the command with no color set, so it uses the theme's color
```
SteelSlayer, line 710: Please use early returns here.

code/game/objects/items.dm
```diff
@@ -707,6 +721,41 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 	if(loc && I.loc == loc && istype(loc, /obj/item/storage) && loc.Adjacent(user)) // Are we trying to swap two items in the storage?
 		var/obj/item/storage/S = loc
 		S.swap_items(src, I, user)
+	remove_outline() //get rid of the hover effect in case the mouse exit isn't called if someone drags and drops an item and somthing goes wrong
+
+/obj/item/proc/apply_outline(outline_color = null)
```
SteelSlayer, line 726: I would go ahead and pass `mob/user` into this proc, so we can avoid using `usr`.

code/game/objects/items.dm
```diff
@@ -117,6 +117,10 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 	var/in_inventory = FALSE //is this item equipped into an inventory slot or hand of a mob?
 	var/tip_timer = 0
 
+	/// item hover FX
+	var/in_storage = FALSE
+	var/outline_filter
```
SteelSlayer, line 122: Consider documenting this variable, as it's not apparent what it does at a glance.

code/game/objects/items.dm
```diff
@@ -707,6 +721,41 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
...
+				outline_color = COLOR_THEME_CLOCKWORK //if you want free gbp go fix the fact that clockwork's tooltip css is glass'
+			if("glass")
+				outline_color = COLOR_THEME_GLASS
+			else //this should never happen, hopefully
+				outline_color = COLOR_WHITE
+	if(color)
+		outline_color = COLOR_WHITE //if the item is recolored then the outline will be too, let's make the outline white so it becomes the same color instead of some ugly mix of the theme and the tint
+	if(outline_filter)
+		filters -= outline_filter
+	outline_filter = filter(type="outline", size=1, color=outline_color)
```
SteelSlayer, line 752:
```suggestion
	outline_filter = filter(type = "outline", size = 1, color = outline_color)
```

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -59,6 +59,9 @@
 
 	populate_contents()
 
+	for(var/obj/item/L in return_inv()) //marks all items in storage as being such
+		L.in_storage = TRUE
+
```
SteelSlayer, line 64: This *might* be better on item init, just so we're not performing this on empty storages. The item would check it's location and see if it's in storage or not, and set the var accordingly.

Kyep (approved):

dearmochi:
Just musing, but perhaps using [image.override](https://secure.byond.com/docs/ref/index.html#/atom/var/override) and [client.images](https://secure.byond.com/docs/ref/index.html#/client/var/images) you can make the filter show for only one client as opposed to everyone. This could help protect against it showing in unintended cases and (I have no data for this) less stuff to network to other players.

S34NW:
> 
> 
> Just musing, but perhaps using [image.override](https://secure.byond.com/docs/ref/index.html#/atom/var/override) and [client.images](https://secure.byond.com/docs/ref/index.html#/client/var/images) you can make the filter show for only one client as opposed to everyone. This could help protect against it showing in unintended cases and (I have no data for this) less stuff to network to other players.

What sort of unintended cases can you think of? I'd imagine the only time you would see someone else's hover outline is when you are both rummaging through the same bag/storage container that is on the floor

Sikroth:
For unintended cases I can imagine there could be items keeping their outline when being quickly equipped/moved into a container/dropped soon after mousing over and such. I know this problem of the outlines persisting happens in another small codebase (spanish server, downstream of para), but it might be cause their implementation of the outline was not as good. I can't say for sure.

code/game/objects/items.dm
```diff
@@ -117,6 +117,12 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 	var/in_inventory = FALSE //is this item equipped into an inventory slot or hand of a mob?
 	var/tip_timer = 0
 
+	// item hover FX
+	/// Is this item inside a storage object?
+	var/in_storage = FALSE
```
SteelSlayer, line 122: I just noticed we have a variable `in_inventory` which is *very* similar to this var. That one gets set to true on equip, pickup, and when placed in storage. Basically when the item is in a place where you'd want the outline. Might be better to use that one.

S34NW: We actually check for both here, because `in_inventory` does not cover items that are on the ground or items that are nested within storage containers.

SteelSlayer: `in_inventory` should handle *all* cases where an item is inside storage. If you look at the bottom of 
`/obj/item/storage/proc/handle_item_insertion`, it gets set to true there. I can't imagine why it wouldn't work if the storage item is on the ground, or with nesting. Unless there's something I'm not seeing.

S34NW: Hmm, there might be something funky happening somewhere then, because it wasn't working properly with just that when I was testing.

variableundefined (approved):

PR #16454, opened 2021-07-29
[WIP] Refactors plasteel floors [MDB IGNORE]

S34NW:
## What Does This PR Do
This PR totally restructures how plasteel floors are defined, instead of every different floor being a varedit, leading to certain floors not showing up in map editors unless they are already used, plasteel floors now follow logical object hierarchy.
I also tidied up turfs.dmi to be less of a nightmare to look at, removing duplicates and such along the way.

TODO:

- [ ] Look into porting TG plasteel icons, might as well while I'm doing this.
- [ ] Fix things that rely on the old method, such as floor painter.
- [ ] Cry

## Why It's Good For The Game
Easier turf implementation for mappers.


AffectedArc07:
1. I literally told you earlier in design chat today to not do this, yet you did it anyway
2. You crashed mapdiffbot which had a knockon effect of crashing other services I host

S34NW:
> 
> 
>     1. I literally told you earlier in design chat today to not do this, yet you did it anyway

Since you feel the need to publicly post this, here is what you actually said:

image

That is far too vague to be taken as "do not do this", it is implied you don't think it's a good idea but that is far from a "don't do this".

Fox-McCloud:
Yeah, no.

This is the last thing in the world we need right, now---compounding the difficulty of a larger turf refactor.

This also aims to make an instance for every turf type there is, which just isn't feasible right now (and questionable if that should ever exist with dir generation), not to mention there's the distinct chance (but by no means guarantee) that we might not use this methodology and may, instead, use decals in the future.

Plus, you were literally told not to.

👎

variableundefined (commented):
Third maint objection for the same reasons as above.

AffectedArc07:
Yeah this has 3 maint objections and isnt going to go anywhere.

PR #16457, opened 2021-07-30
Re-ups terror web deletion timer to 7 seconds

### LINKED ISSUE #16455: Terror webs deleted too soon, gray terrors broken
**What is the problem?**
Terror webs are deleted within a second of crew walking into the web, rather than it taking ~7 seconds, as is intended.

When a human walks into a terror web, it should take ~7 seconds for the web to be deleted.

**What happened instead**:
It was deleted in less than a second. Almost instantly.

This cripples gray terrors.
Per /mob/living/simple_animal/hostile/poison/terror_spider/gray/spider_specialattack grays do double damage to targets standing on a web.
Obviously, if a human being on top of a web almost instantly deletes the web, then a gray attacking a human won't count as attacking the human on a web, even if they were just dragged onto a web. 
Thus, grays will never get their bonus damage, even when they play correctly, and grays are crippled as a class - their damage is reduced by 50% even when they're doing everything correctly.

1) Walk onto a terror web.
2) Observe how the web does not last 7 seconds.

When https://github.com/ParadiseSS13/Paradise/pull/15721/files changed the timer on terror webs being destroyed from 7 seconds after touch, to 0.7 seconds after touch.
That PR was labeled a fix, but it really broke the balance for gray terrors.

dearmochi:
## What Does This PR Do

- Re-ups terror web deletion timer to 7 seconds as the change was a regression
  - Fixes #16455

## Why It's Good For The Game

- Undoes a massive grey terror nerf

## Changelog
:cl:
fix: Terror spider webs no longer delete instantly
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/terror_spiders/actions.dm
+++ b/code/modules/mob/living/simple_animal/hostile/terror_spiders/actions.dm
@@ -193,7 +193,7 @@
 		M.Weaken(4) // 8 seconds.
 		if(iscarbon(M))
 			web_special_ability(M)
-			addtimer(CALLBACK(src, .proc/after_carbon_crossed, M), 0.7 SECONDS, TIMER_UNIQUE | TIMER_OVERRIDE)
+			addtimer(CALLBACK(src, .proc/after_carbon_crossed, M), 7 SECONDS, TIMER_UNIQUE | TIMER_OVERRIDE)
 
 /**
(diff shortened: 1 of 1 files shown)


AffectedArc07 (approved):
This is deemed a fix

Kyep:
Great response time, well done mochi.

PR #16464, opened 2021-08-01
Fixes runtimes caused by null client preferences

### LINKED ISSUE #16058: Clients can end up with null preference datums, which breaks a lot of stuff
**Issue Description**:
Sometimes clients can end up with null preference datums, presumably if they DC before they fully connect

The server to work as normal

**What happened instead**:
**ANYTHING** that even remotely touches preferences breaks, including but not limited to
- PMing clients
- PDA messaging
- Ghost spawns
- Radio chatter

image

A **LOT** of stuff breaks

Null out your preference datum and watch the server fall apart

Presumably since the dawn of time

**Extra information**:
If someone could make a debug verb that force reloads someones preference datum, that would be grand

SabreML:
## What Does This PR Do

Fixes #16058 (`Runtime in preferences_mysql.dm,199: Cannot read null.ckey`) and related runtimes by adding a check in `load_character()` to ensure that the client still exists after the first DB query, and deleting their preferences datum if not, so that it can be remade later.

Tested by adding a `qdel(C)` on the line above. It does also seem to work fine if the datum isn't deleted, but I can imagine that it not being able to load fully could cause problems later on.

## Why It's Good For The Game

I can only speak for the runtime logs I have downloaded of course, but based on those this seems to be an incredibly common error.

## Changelog
:cl:
fix: Fixed multiple runtimes caused by clients disconnecting before the game has fully loaded their preferences.
/:cl:

### DIFF
--- a/code/modules/client/preference/preferences_mysql.dm
+++ b/code/modules/client/preference/preferences_mysql.dm
@@ -126,10 +126,11 @@
 	qdel(query)
 	return 1
 
-/datum/preferences/proc/load_character(client/C,slot)
+/datum/preferences/proc/load_character(client/C, slot)
 	saved = FALSE
 
-	if(!slot)	slot = default_slot
+	if(!slot)
+		slot = default_slot
 	slot = sanitize_integer(slot, 1, max_save_slots, initial(default_slot))
 	if(slot != default_slot)
 		default_slot = slot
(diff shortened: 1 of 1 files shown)


dearmochi (approved):

AffectedArc07 (approved):
This will also be fixed in #16449 but that will take a bit

PR #16469, opened 2021-08-02
Can't have infinite EMP implant charges anymore

farie82:
## What Does This PR Do
Makes the EMP implant call `empulse` async. Fixing certain cases where you could make an infinite use EMP implant.
Also adds some documentation

## Why It's Good For The Game
Fixes an exploit

## Changelog
:cl:
fix: Can't make an infinite use EMP implant anymore
/:cl:

### DIFF
--- a/code/game/objects/empulse.dm
+++ b/code/game/objects/empulse.dm
@@ -1,12 +1,12 @@
 /**
- * Will cause an EMP pulse on the given epicenter.
+ * Will cause an EMP on the given epicenter.
  * This proc can sleep depending on the affected objects. So assume it sleeps!
  *
- * epicenter - The center of the EMP pulse. Can be an atom, as long as the given atom is on a turf (in)directly
+ * epicenter - The center of the EMP. Can be an atom, as long as the given atom is on a turf (in)directly
  * heavy_range - The max distance from the epicenter where objects will be get heavy EMPed
  * light_range - The max distance from the epicenter where objects will get light EMPed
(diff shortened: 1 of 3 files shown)


code/game/objects/empulse.dm
```diff
@@ -1,3 +1,13 @@
...
+ * Will cause an EMP on the given epicenter.
+ * This proc can sleep depending on the affected objects. So assume it sleeps!
+ *
+ * epicenter - The center of the EMP. Can be an atom, as long as the given atom is on a turf (in)directly
+ * heavy_range - The max distance from the epicenter where objects will be get heavy EMPed
+ * light_range - The max distance from the epicenter where objects will get light EMPed
+ * log - Whether or not this action should be logged or not. Will use the cause if provided
+ * cause - The cause of the EMP. Used for the logging
+ */
 /proc/empulse(turf/epicenter, heavy_range, light_range, log = FALSE, cause = null)
```
SabreML, line 11: Might it be better to make the individual `emp_act()` calls in here async, rather than having to always specify it separately for `empulse()`?

farie82: Those calls expect a return value which is later used. I can remove the feature and just do that. But that would change existing functionality which I am trying to avoid with this PR

SabreML: Ah right, it doesn't return anything. All good then.

SabreML (approved):

AffectedArc07 (approved):

PR #16474, opened 2021-08-02
"Crew_transfer vote" to "Crew transfer vote"

SabreML:
## What Does This PR Do

Removed the underscore from "Crew_transfer vote" to make it "Crew transfer vote" instead, since there's no reason for it to be there.

Also fixed the ordering of the admin voting panel so that the "Allowed/Disallowed" buttons actually line up with their corresponding option, and added admin logging when they get toggled.
(The buttons are for restart and gamemode votes.)

## Why It's Good For The Game

There's no reason for that underscore to be there. I think it was only added so that it worked properly with this href link, but even that doesn't really explain it:
https://github.com/ParadiseSS13/Paradise/blob/5a4dd899a961a6d3bed0b16edba2dd5e35654732/code/controllers/subsystem/vote.dm#L322-L326
For the rest of it, I'm editing things across the whole file already so might as well make a few fixes while I'm there.

## Images of changes

**Before:**
image
**After:**
image
<hr>

**Before:**
IC4DKA5xRk
**After:**
8Lits8hH5k

## Changelog
:cl:
tweak: Changed the end of round "Crew_transfer vote" to "Crew transfer vote", removing the underscore.
tweak: Cleaned up and added admin logging to the admin voting panel.
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -396,11 +396,11 @@ SUBSYSTEM_DEF(vote)
 		if("toggle_restart")
 			if(admin)
 				GLOB.configuration.vote.allow_restart_votes = !GLOB.configuration.vote.allow_restart_votes
-				log_and_message_admins("[key_name_hidden(usr)] has [GLOB.configuration.vote.allow_restart_votes ? "enabled" : "disabled"] public restart voting.")
+				log_and_message_admins("has [GLOB.configuration.vote.allow_restart_votes ? "enabled" : "disabled"] public restart voting.")
 		if("toggle_gamemode")
 			if(admin)
 				GLOB.configuration.vote.allow_mode_votes = !GLOB.configuration.vote.allow_mode_votes
-				log_and_message_admins("[key_name_hidden(usr)] has [GLOB.configuration.vote.allow_mode_votes ? "enabled" : "disabled"] public gamemode voting.")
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/vote.dm
```diff
@@ -387,14 +391,16 @@ SUBSYSTEM_DEF(vote)
 				var/votedesc = capitalize(mode)
 				if(mode == "custom")
 					votedesc += " ([question])"
-				log_and_message_admins("cancelled the running [votedesc] vote.")
+				log_and_message_admins("cancelled the running '[votedesc]' vote.")
 				reset()
 		if("toggle_restart")
 			if(admin)
 				GLOB.configuration.vote.allow_restart_votes = !GLOB.configuration.vote.allow_restart_votes
+				log_and_message_admins("[key_name_hidden(usr)] has [GLOB.configuration.vote.allow_restart_votes ? "enabled" : "disabled"] public restart voting.")
```
AffectedArc07, line 399: Remove the double key name here
```dm
/proc/log_and_message_admins(message)
	log_admin("[key_name(usr)] " + message)
	message_admins("[key_name_admin(usr)] " + message)
```

SabreML: You know, I actually checked that proc twice to make sure it was actually doing what I wanted. Not sure how I missed that.

Kyep (approved):

AffectedArc07 (approved):

PR #16481, opened 2021-08-04
Removes atmospheric pipe redundancies

ppi13:
## What Does This PR Do

Like it says on the can, it removes redundancies, by removing bits of piping that feed atmospheric to areas from multiple directions.
It also adds some missing vents and scrubbers along east and west central hallway, as well as adding a scrubber to the south-east of the engine room, and moving one scrubber in AI sat atmospherics.
Also removes some minor bloat.

## Why It's Good For The Game

Too much redundancies are nice, but from a game perspective it's not, a few pipes exploding will not cut the area after from atmospherics making simple pipe sabotage dangerous if combined with a well placed explosive.
This will encourage atmospheric technicians to properly fix broken pipes, instead of go "eh, we got redundancies."
This will also inconvenience vent crawlers, by slowing them down.

## Images of changes

Before:
pipes
After:
pipes3

## Changelog
:cl: ppi
del: Removes redundancies in atmospheric piping
/:cl:

### DIFF
--- a/_maps/map_files/cyberiad/cyberiad.dmm
+++ b/_maps/map_files/cyberiad/cyberiad.dmm
@@ -10003,9 +10003,6 @@
 	d2 = 2;
 	icon_state = "1-2"
 	},
-/obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{
-	dir = 6
-	},
 /turf/simulated/floor/plasteel,
 /area/security/lobby)
 "awc" = (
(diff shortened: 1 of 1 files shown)


AffectedArc07 (changes requested):
```
Test run failed!
Total runtimes: 1987
```

1. Test your code. This is the 9th time I have asked you to do this before making a PR
2. This only affects Box, and as per map standardisation you should do this on all maps
3. I dont like this

Kepplerx:
As a Atmos Tech, I appreciate more work, but this wouldnt really make that much more work, and even if the pipes are cut there is often enough air to make it not matter. Only looking at it from the eyes of a Atmos Tech I am slightly for. 

On the other hand, this PR has a price. There are multiple antags very reliant on pipes as it is, such as Terrors, Swarmers but also Clings and tators to a small extent. One pipe being cut would then limit them from a half of the station, or even more. Those cut pipes dont need to be on purpose either, if there is no redundancy one pipe cut by a single explosion and forgotten about is enough. Due to this I would be opposed to this PR, though I do like the intention of making more for Atmos.

ppi13:
Okay, the runtimes were caused by some updates applied incorrectly on my end. Will remake later.

PR #16488, opened 2021-08-05
Removes the Central Hallway from the list of contract locations

SabreML:
## What Does This PR Do

Removes the 'Central Primary Hallway' from the list of possible contract target locations.

## Why It's Good For The Game

The 'Central Primary Hallway' is actually *eight* separate areas, each shown in different colours on the image below. Having to run circles around the entire thing along with your target and half of security chasing you might be enjoyable sometimes, but it's a massive jump in difficulty compared to the rest of the locations and most likely not intentional.
There's also been a few mentorhelps about this recently which couldn't be properly answered since there's no actual way of knowing which area the game picked without VV diving.
image

## Changelog
:cl:
del: Removed the Central Primary Hallway from the list of possible contract extraction locations.
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/contractor/datums/objective_contract.dm
+++ b/code/modules/antagonists/traitor/contractor/datums/objective_contract.dm
@@ -100,7 +100,6 @@
 			"AI Satellite Hallway",
 			"Bar",
 			"Cargo Office",
-			"Central Primary Hallway",
 			"Chemistry",
 			"Chief Engineer's office",
 			"Chief Medical Officer's office",


hal9000PR:
it might be a good idea to add a TGUI map with the kidnap zone highlighted

S34NW:
Dare I suggest you rename the areas to be more descriptive with - [Direction] at the end? (Central Primary Hallway - Port, etc.) As opposed to just removing them all.

SabreML:
> Dare I suggest you rename the areas to be more descriptive with - [Direction] at the end? (Central Primary Hallway - Port, etc.) As opposed to just removing them all from the pool.

That would probably a better option than just removing it since it keeps the area as an extraction option if people want to go for a difficult one. Currently though, Meta and Delta both have it as one gigantic area rather than being split into smaller ones, so there might be some standardisation issues:
<details>
<summary><b>Big images:</b></summary>

**Meta:**
Desktop Screenshot 2021 08 06 - 00 15 35 52
**Delta:**
Desktop Screenshot 2021 08 06 - 00 26 44 20
</details>
I'll see about making an alternative PR with @hal9000PR's idea for now since I'm not confident enough in mapping to edit the entire central hallway for two maps, but failing that I'll see what I can do.

Kyep (approved):

AffectedArc07 (approved):

PR #16493, opened 2021-08-06
Removes Cling/Vamp CQC

matttheficus:
## What Does This PR Do

Removes the ability to get CQC as a Changeling or Vampire.

## Why It's Good For The Game

I am a very, very, very vocal user of CQC. I thoroughly enjoy using it as a stealth weapon as an Antagonist and use it quite literally every round I roll Traitor. It is a very strong skill and can easily destroy anyone who isnt expecting it or doesnt have ranged weapons to use to defend themselves at a distance. That being said, given that CQC has been changed recently to be much like CQC by making its passive buff into an activateable ability, it seems odd that one martial art is barred from use while the other is not.

Cling and Vamps already have a myriad of stuns, stims, and combat buffs. Giving them more insta-stuns, combat buffs, and a free kidnapping if done right seems to go against the very ideal that Carp sets as a precedent by being unusable.

The only issue that I can see is a Cling/Tot or Vamp/Tot rolling the Bond Kit and being unable to use the CQC Manual inside. My argument would be they can ALSO pull Carp from a surplus/super surplus and be just as badly out of luck. If the Bundle is the only thing that causes issues I can look into a possible workaround with that specific one.

## Changelog
:cl:
add: Added check for vampire/changeling to CQC
/:cl:

### DIFF
--- a/code/modules/martial_arts/martial.dm
+++ b/code/modules/martial_arts/martial.dm
@@ -292,10 +292,9 @@
 	if(user.mind && (user.mind.changeling || user.mind.vampire)) //Prevents changelings and vampires from being able to learn it
 		if(user.mind && user.mind.changeling) //Changelings
 			to_chat(user, "<span class ='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
-			return
 		else //Vampires
 			to_chat(user, "<span class ='warning'>Your blood lust distracts you from the basics of CQC!</span>")
-			return
+		return
 
 	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
 	var/datum/martial_art/cqc/CQC = new(null)
--- a/code/modules/martial_arts/martial.dm
+++ b/code/modules/martial_arts/martial.dm
@@ -289,12 +289,13 @@
 /obj/item/CQC_manual/attack_self(mob/living/carbon/human/user)
 	if(!istype(user) || !user)
 		return
-	if(user.mind && (user.mind.changeling || user.mind.vampire)) //Prevents changelings and vampires from being able to learn it
-		if(user.mind && user.mind.changeling) //Changelings
-			to_chat(user, "<span class ='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
-		else //Vampires
-			to_chat(user, "<span class ='warning'>Your blood lust distracts you from the basics of CQC!</span>")
-		return
+	if(user.mind) //Prevents changelings and vampires from being able to learn it
+		if(user.mind.changeling) //Changelings
+			to_chat(user, "<span class='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
+			return
+		else if(user.mind.vampire) //Vampires
+			to_chat(user, "<span class='warning'>Your blood lust distracts you from the basics of CQC!</span>")
+			return
 
 	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
 	var/datum/martial_art/cqc/CQC = new(null)
--- a/code/modules/martial_arts/martial.dm
+++ b/code/modules/martial_arts/martial.dm
@@ -289,8 +289,15 @@
 /obj/item/CQC_manual/attack_self(mob/living/carbon/human/user)
 	if(!istype(user) || !user)
 		return
-	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
+	if(user.mind && (user.mind.changeling || user.mind.vampire)) //Prevents changelings and vampires from being able to learn it
+		if(user.mind && user.mind.changeling) //Changelings
+			to_chat(user, "<span class ='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
+			return
+		else //Vampires
+			to_chat(user, "<span class ='warning'>Your blood lust distracts you from the basics of CQC!</span>")
+			return
 
+	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
 	var/datum/martial_art/cqc/CQC = new(null)
 	CQC.teach(user)
 	user.drop_item()


S34NW:
Considering CQC is one of the strongest stealth tools available to an antag, piling it on top of what vamps and clings already get does lead to problematic situations. I think I like this.

hal9000PR (changes requested):
You are gonna want to make CQC refundable like other items vamps/clings cannot use.

Coolrune206:
> You are gonna want to make CQC refundable like other items vamps/clings cannot use.

Remove it from the Bond bundle and then yes. Otherwise, order null crates from cargo > refund CQC from bond crates > badda bing badda boom free TC.

matttheficus:
> > You are gonna want to make CQC refundable like other items vamps/clings cannot use.
> 
> Remove it from the Bond bundle and then yes. Otherwise, order null crates from cargo > refund CQC from bond crates > badda bing badda boom free TC.

This is the only big glaring issue I see, though I'm unsure what would replace CQC in the Bond Kit/how to get around the Null spam for free TC.

Wojz:
Changelings + CQC = near-instant sleep tentacle
Vampire + CQC = near-instant AoE sleep

Yeah no thanks. I have been a CQC changeling before and it's absolutely insane. But the issue of the Bond kit does remain. Personally I wouldn't mind it being replaced by something else, Bond is well known for using a myriad of gadgets. A sleepy pen and maybe something like chameleon thermals wouldn't be a bad replacement.

Cocacolagua:
> > You are gonna want to make CQC refundable like other items vamps/clings cannot use.
> 
> Remove it from the Bond bundle and then yes. Otherwise, order null crates from cargo > refund CQC from bond crates > badda bing badda boom free TC.

For clarification's sake, you can already do this with the implant bundle, which has an uplink implant. This does mean that vampires/clings would get two options that give TC

Coolrune206:
> > > You are gonna want to make CQC refundable like other items vamps/clings cannot use.
> > 
> > 
> > Remove it from the Bond bundle and then yes. Otherwise, order null crates from cargo > refund CQC from bond crates > badda bing badda boom free TC.
> 
> For clarification's sake, you can already do this with the implant bundle, which has an uplink implant. This does mean that vampires/clings would get two options that give TC

Normal tots would also be able to refund it I believe, giving everyone more refund potential.

Peakwastaken:
> > > You are gonna want to make CQC refundable like other items vamps/clings cannot use.
> > 
> > 
> > Remove it from the Bond bundle and then yes. Otherwise, order null crates from cargo > refund CQC from bond crates > badda bing badda boom free TC.
> 
> This is the only big glaring issue I see, though I'm unsure what would replace CQC in the Bond Kit/how to get around the Null spam for free TC.

Would it be possible to make it so the one that comes in nulls is non-refundable ? Otherwize, the opposite, make a check for whether it was bought through pda/surplus to be refundable ?

This is honestly the only reason I didn't try making this PR myself, fixing that connundrum is kind of out of my reach, but this change is really needed - even if it's to risk the odd-once-in-a-blue-moon vamp/cling getting cqc in their surplus and potentially ahelping for a refund prior to finding an adequate fix to it.

hal9000PR:
Imma be honest, I'm not 100% sold on this. No matter what tot vamp/traitor will always have op stuff. Instead they will just buy an ebow and emp flash. Perhaps the issue isn't the items they can get, but instead the fact that a player can roll both. When a player rolls traitor vamp/cling they either get massively overwhelmed by the 7 objectives they get or they become next to unkillable.

Kyep (approved):

code/modules/martial_arts/martial.dm
```diff
@@ -289,8 +289,15 @@
...
 	if(!istype(user) || !user)
 		return
-	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
+	if(user.mind && (user.mind.changeling || user.mind.vampire)) //Prevents changelings and vampires from being able to learn it
+		if(user.mind && user.mind.changeling) //Changelings
+			to_chat(user, "<span class ='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
+			return
+		else //Vampires
+			to_chat(user, "<span class ='warning'>Your blood lust distracts you from the basics of CQC!</span>")
+			return
```
AffectedArc07, line 298:
```suggestion
		if(user.mind && user.mind.changeling) //Changelings
			to_chat(user, "<span class ='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
		else //Vampires
			to_chat(user, "<span class ='warning'>Your blood lust distracts you from the basics of CQC!</span>")
		return
```

AffectedArc07 (changes requested):

code/modules/martial_arts/martial.dm
```diff
@@ -289,8 +289,14 @@
 /obj/item/CQC_manual/attack_self(mob/living/carbon/human/user)
 	if(!istype(user) || !user)
 		return
-	to_chat(user, "<span class='boldannounce'>You remember the basics of CQC.</span>")
+	if(user.mind && (user.mind.changeling || user.mind.vampire)) //Prevents changelings and vampires from being able to learn it
+		if(user.mind && user.mind.changeling) //Changelings
+			to_chat(user, "<span class ='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
+		else //Vampires
+			to_chat(user, "<span class ='warning'>Your blood lust distracts you from the basics of CQC!</span>")
+		return
```
dearmochi, line 297:
```suggestion
	if(user.mind) //Prevents changelings and vampires from being able to learn it
		if(user.mind.changeling) //Changelings
			to_chat(user, "<span class='warning'>We try multiple times, but we simply cannot grasp the basics of CQC!</span>")
			return
		else if(user.mind.vampire) //Vampires
			to_chat(user, "<span class='warning'>Your blood lust distracts you from the basics of CQC!</span>")
			return
```

Looks cleaner IMO

dearmochi (changes requested):
Nit

code/modules/martial_arts/martial.dm
matttheficus: agreed, less painful to read

dearmochi (approved):

variableundefined (approved):

PR #16484, opened 2021-08-05
Fixes contractor extraction areas never being chosen due to bad names

S34NW:
## What Does This PR Do

Fixes an issue caused by #16099 where contract extraction locations were reduced due to some areas not being found after they were renamed. This should also mean contractors work properly on all maps that use the standard area names.

## Why It's Good For The Game
Contractors had an intended extraction location variety, this restores that.

## Changelog
:cl:
fix: Fixed contractor extraction locations being more limited than intended.
/:cl:

PR #15335, opened 2021-01-23
[TGUI] GPS

dearmochi:
## What Does This PR Do

- Ports GPS from Browser UI to TGUI
  - The UI now auto-refreshes as a result
  - The UI now displays distance and direction to another signal
- Adds ability to only display signals on the same Z-level
- Cleans up, refactors and documents the GPS code

## Why It's Good For The Game

TGUI good, GPS less of a pain to use

## Images of changes

image

## Changelog
:cl:
add: GPS interface has been remade and improved in TGUI
/:cl:

PR #16497, opened 2021-08-07
Adds a pointer to the extraction location for Contractor uplinks

SabreML:
## What Does This PR Do

Adds an extraction location pinpointer to the contractor uplink, as an alternative PR to #16488.

After taking a contract a yellow arrow marker will appear above the name of the extraction location, updating when the player moves to always point in the direction of the target location. When the player reaches the correct area, the marker will change into a green bullseye symbol.
*(Icon rotation code copied from #15335 so credit to Mochi for that.)*

## Why It's Good For The Game

The primary reason is explained on the other PR, so I'll just copy it here for posterity:
> The 'Central Primary Hallway' is actually eight separate areas, each shown in different colours on the image below. Having to run circles around the entire thing along with your target and half of security chasing you might be enjoyable sometimes, but it's a massive jump in difficulty compared to the rest of the locations and most likely not intentional.
There's also been a few mentorhelps about this recently which couldn't be properly answered since there's no actual way of knowing which area the game picked without VV diving.

This will not only solve that issue (as well as keeping the Central Primary Hallway as an extraction location), it will also solve it for every other location on the map too. I know there's still some confusion about where exactly the 'Aft Port Solar Maintenance' and similar areas are, and with #16484 coming I imagine there will be even more.
"Veteran" players who have the map memorised already won't gain any advantages from this since they'll know where to go anyway, so this just serves to assist people who are newer to the game or just have a bad sense of direction.

## Images of changes

<details>
<summary><b>Big images:</b></summary>

### Before taking a contract: (Same as current)
PtkDojZkjh

### During a contract:
chrD7lIgCj

### After completing a contract: (Same as current)
1mtFaVXjwW

</details>

https://user-images.githubusercontent.com/57483089/128580728-86a4645b-3ae1-484e-b0e4-501168129993.mp4

## Changelog
:cl:
add: Added an extraction location pinpointer to the contractor uplink, which points towards the target area.
tweak: Made the contractor uplink autoupdate when inside pockets.
/:cl:


code/modules/tgui/external.dm
```diff
@@ -55,10 +55,10 @@
...
  */
 /datum/proc/update_static_data(mob/user, datum/tgui/ui, ui_key = "main")
-	ui = SStgui.try_update_ui(user, src, ui_key, ui)
-	// If there was no ui to update, there's no static data to update either.
 	if(!ui)
-		return
+		ui = SStgui.try_update_ui(user, src, ui_key, ui)
+		if(!ui)
+			return // If there was no ui to update, there's no static data to update either.
 	ui.push_data(null, ui_static_data(user), TRUE)
```
SabreML, line 62: This is unrelated and shouldn't actually be in this PR, it's a leftover of some other experimentation I was doing. It *is* a code improvement though so I might just leave it.

Cocacolagua:
even if #16488 is merged, I'd like to see this anyway, it helps a lot for AI-sat rooms and confusing names in general.

Wojz:
This is a good change. Many rooms have completely unknown names to most people.

Kyep (approved):

code/modules/tgui/external.dm
AffectedArc07, line 62: Revert this change. `SStgui.try_update_ui` is required to ensure the UI re-initializes if required.

AffectedArc07 (changes requested):

code/modules/tgui/external.dm
SabreML: Ohhh, that makes a lot of sense actually.

AffectedArc07 (approved):

PR #16498, opened 2021-08-07
Adds Fake Blindfold 

shmabuu:
## What Does This PR Do
This is purely meant for cosmetic value and games/tricks. Currently if you'd like to wear a blindfold but still be able to see, options are: being a cultist using the zealots blindfold or getting chameleon glasses. Otherwise, you'd have to be blind.

## Why It's Good For The Game
This is purely meant for cosmetic value and games/tricks. Currently if you'd like to wear a blindfold but still be able to see, the only options being a cultist using the zealots blindfold or getting chameleon glasses. Otherwise, you'd have to be blind.

## Images of changes
image
image

## Changelog
:cl:
add: Adds Fake blindfold, craftable with two cloth sheets.
/:cl:

### DIFF
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -218,11 +218,13 @@
 
 /obj/effect/proc_holder/spell/vampire/mob_aoe/glare/cast(list/targets, mob/user = usr)
 	user.visible_message("<span class='warning'>[user]'s eyes emit a blinding flash!</span>")
-	if(istype(user:glasses, /obj/item/clothing/glasses/sunglasses/blindfold))
-		var/obj/item/clothing/glasses/sunglasses/blindfold/B = user:glasses
-		if(B.tint)
-			to_chat(user, "<span class='warning'>You're blindfolded!</span>")
-			return
+	if(ishuman(user))
+		var/mob/living/carbon/human/H = user
+		if(istype(H.glasses, /obj/item/clothing/glasses/sunglasses/blindfold))
+			var/obj/item/clothing/glasses/sunglasses/blindfold/B = H.glasses
+			if(B.tint)
+				to_chat(user, "<span class='warning'>You're blindfolded!</span>")
+				return
 	for(var/mob/living/target in targets)
 		if(!affects(target))
 			continue
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -219,8 +219,10 @@
 /obj/effect/proc_holder/spell/vampire/mob_aoe/glare/cast(list/targets, mob/user = usr)
 	user.visible_message("<span class='warning'>[user]'s eyes emit a blinding flash!</span>")
 	if(istype(user:glasses, /obj/item/clothing/glasses/sunglasses/blindfold))
-		to_chat(user, "<span class='warning'>You're blindfolded!</span>")
-		return
+		var/obj/item/clothing/glasses/sunglasses/blindfold/B = user:glasses
+		if(B.tint)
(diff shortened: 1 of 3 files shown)


code/modules/clothing/glasses/glasses.dm
```diff
@@ -395,6 +395,15 @@
 	tint = 3				//to make them blind
 	prescription_upgradable = 0
 
+/obj/item/clothing/glasses/sunglasses/blindfold/fake
+    name = "tattered blindfold"
+    desc = "A see-through blindfold perfect for cheating at games like pin the stunbaton on the clown."
+    icon_state = "blindfold"
+    item_state = "blindfold"
+    prescription_upgradable = FALSE
+    flash_protect = 0
```
SabreML, line 404: Don't need these three here anymore, since they're all defined on the parent `/obj/item/clothing/glasses/sunglasses/blindfold` object. Since this is a subtype of that now it'll inherit them.
```suggestion
    desc = "A see-through blindfold perfect for cheating at games like pin the stunbaton on the clown."
    flash_protect = 0
```

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -219,8 +219,10 @@
 /obj/effect/proc_holder/spell/vampire/mob_aoe/glare/cast(list/targets, mob/user = usr)
 	user.visible_message("<span class='warning'>[user]'s eyes emit a blinding flash!</span>")
 	if(istype(user:glasses, /obj/item/clothing/glasses/sunglasses/blindfold))
-		to_chat(user, "<span class='warning'>You're blindfolded!</span>")
-		return
+		var/obj/item/clothing/glasses/sunglasses/blindfold/B = user:glasses
+		if(B.tint)
+			to_chat(user, "<span class='warning'>You're blindfolded!</span>")
+			return
 	for(var/mob/living/target in targets)
```
SabreML, line 226: Would you mind changing this bit of code to avoid the use of `:`? It was like this already so it's not an obligation, but it'd be good to update the code to be in line with the [Contribution Guidelines](https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md#No-overriding-type-safety-checks).
I can't leave a github suggestion on this bit so you'll have to type it in manually:
```dm
/obj/effect/proc_holder/spell/vampire/mob_aoe/glare/cast(list/targets, mob/user = usr)
	user.visible_message("<span class='warning'>[user]'s eyes emit a blinding flash!</span>")
	if(ishuman(user))
		var/mob/living/carbon/human/H = user
		if(istype(H.glasses, /obj/item/clothing/glasses/sunglasses/blindfold))
			var/obj/item/clothing/glasses/sunglasses/blindfold/B = H.glasses
			if(B.tint)
				to_chat(user, "<span class='warning'>You're blindfolded!</span>")
				return

	for(var/mob/living/target in targets)
```

SabreML (commented):
Linking #16278 for posterity

code/modules/clothing/glasses/glasses.dm
```diff
@@ -395,6 +395,11 @@
 	tint = 3				//to make them blind
 	prescription_upgradable = 0
 
+/obj/item/clothing/glasses/sunglasses/blindfold/fake
+    name = "tattered blindfold"
+    flash_protect = 0
+    tint = 0
```
SabreML, line 401: I think the description got caught in the crossfire of the variable removal here, you probably want to add that back in.
Also this has spaces rather than tabs for indentation, so it'd be good to change those over too:
```suggestion
/obj/item/clothing/glasses/sunglasses/blindfold/fake
	name = "tattered blindfold"
	desc = "A see-through blindfold perfect for cheating at games like pin the stunbaton on the clown."
	flash_protect = 0
	tint = 0
```

SabreML (approved):

PR #16499, opened 2021-08-07
Non Christmas snow machines

AllStickyGithub:
## What Does This PR Do
Allows you to order snow machines no matter the season. The snow machine crate no longer has the special variable and any mention of it has been eradicated from the Christmas holiday event.

## Why It's Good For The Game
I think our beautiful drask players would love a way to make cold rooms cooler than they already are.

## Notes
-If you do not want snow machines in their current state to be a permanent thing I will close this PR and make a new one with adjustment to the snow machine.
-If you do not want snow machines to be a permanent thing EVER let me know so I do not waste time on coding dumb things.

## Changelog
:cl:
tweak: Snow machines can now be ordered any day of the year
/:cl:


Wojz:
Said so on the forums already, but I'll echo it here, I think limiting the existence of snow to one Christian holiday is kinda weird. These machines can be used to make interesting-looking things, Drask rooms being one, but another being, for example, an Arctic-themed room for something like a player-made zoo, which is a really cool thing I've seen built as part of DNA vaults sometimes. I like the increased amount of options they bring.

Spacemanspark:
If we make this a player creatable thing there should probably be some limitations or tweaks. Snow being able to trivially coat the entire station doesn't particularly sound great outside of the Christmas event.

I'd make a subtype of the snow machine that players get when they build one. Perhaps have some sort of tempature cool down requirement and range decrease.

The current snow machine can still be present (pun intended) for the Christmas event.

AllStickyGithub:
> If we make this a player creatable thing there should probably be some limitations or tweaks. Snow being able to trivially coat the entire station doesn't particularly sound great outside of the Christmas event.

I feel like you are over estimating the grief potential of the snow machine. I did this too on the forum before I did some testing. Using a snow machine to cover the whole station with snow takes a very very long time with a player standing next to it feeding water to machine constantly. On top of that it is also easy to clean up. All it requires is for one to turn the thermostat up and  to turn off the machine. The playerbase already have so many ways to grief some of which are round ending without admin intervention. This in comparison seems very minor and easily fixed.

Spacemanspark:
This is less 'grief' and more 'it's obnoxious and shouldn't have this much reach'.

AllStickyGithub:
Same thing really, a full snow machine will cover a space of 10x10 tiles to increase the range you need to keep it full and keep any airlocks that block it open.

code/modules/holiday/christmas.dm
```diff
@@ -9,10 +9,6 @@
 	for(var/datum/crafting_recipe/snowman/S in GLOB.crafting_recipes)
 		S.always_availible = TRUE
 		break
-	//The following spawn is necessary as both the timer and the shuttle systems initialise after the events system does, so we can't add stuff to the shuttle system as it doesn't exist yet and we can't use a timer
-	spawn(60 SECONDS)
-		var/datum/supply_packs/misc/snow_machine/xmas = SSshuttle.supply_packs["[/datum/supply_packs/misc/snow_machine]"]
-		xmas.special = FALSE
```
dearmochi, line 15: Eww.

dearmochi (approved):

code/modules/holiday/christmas.dm
AllStickyGithub: if it works 🍶

AffectedArc07:
Not in favour. Given how covered the station gets already and how this is a pain in the ass to deal with, no.

Fox-McCloud:
Not a fan of this either, for the reason @AffectedArc07 has elaborated.

AffectedArc07:
2 maints against, no one in favour from either party. Closing.

PR #10576, opened 2019-01-02
Floor Tile Updates

AffectedArc07:
**What does this PR do:**
Ports over the floor tile sprites from OracleStation, *F*
This was done for a few reasons:
- They have a cleaner aesthetic
- They are easier on the eyes
- They fit this game more (In my opinion)
- I was porting the bluespace tiles and thought it looked better with all of them

**Images of sprite/map changes:**
Instead of pasting **EVERY** single tile diff, I made a before and after webmap of the station with the new tiles. https://affectedarc07.co.uk/IconDiff/
***NOTE: This will take a while to load as its hosted on my own connection, not GitHub***

Bluespace floors (Cant be conveyed on a webmap):
dllxalxzny

**TODO:**
- [x] Port floor sprites
- [x] Update inhand sprites
- [x] Mech charger floors (In 2 files for some reason?)
- [x] Outpost floors (Again, 2 files for some reason)

**Changelog:**
:cl: AffectedArc07
tweak: Floor tiles have been given a new coat of paint
tweak: Bluespace floor tiles no longer look like they will kill you
/:cl:

PR #16503, opened 2021-08-08
TG sprite project: Floor turfs

S34NW:
## What Does This PR Do

This PR attempts to replicate TG's floor turf sprites in para.
There is one key limiting difference that calls for our awkward implementation, and not a direct port:
TG uses decals for edge colours, we use icon states with hard-baked icons for every combination of floor.

That means I've had to go and edit all current floor tiles with colours to use the TG siding decals in an icon state.
There are more TG decal styles that I have not ported, because that is an unholy amount of work that can be tackled in chunks, this just converts what we currently use to TG style floors.

Note: This might require some mapping changes down the line to tackle awkward looking areas, since we seem to love sticking corners down randomly as highlights. The only mapping changes I have had to make is moving the "Space Station 13" logo to decals so we can tidy up floors.dmi

## Why It's Good For The Game
TG sprite project brings us out of dinosaur era sprites. this will remove the ugly "compression" look we have as an attempt at shading, something that makes our turfs look _very_ dated. This also opens up the door for more cool floor options from the TG decals I have not implemented.

## Images of changes
IDB for the full list of icons, I've also tidied floors.dm up.

Comparison SS:
imageimage

Here are some "in-engine" screenshots:
image
image
image
image

### Here is an example of the new "weirdness" caused by edge icons used commonly. An easy map-edit for later.

image

## Changelog
:cl:
tweak: Ported TG floor turfs
/:cl:


AffectedArc07:
Making this a decal is going to cause problems
image

If the origin atom of this large decal is bottom left, if you walk far enough to the right its going to just clip in and out of view.

S34NW:
> 
> 
> Making this a decal is going to cause problems
> image
> 
> If the origin atom of this large decal is bottom left, if you walk far enough to the right its going to just clip in and out of view.

its not 1 decal, luckily. Each turf has its own segment.

Spacemanspark:
These... look mega fake, to be blunt. Like, a bland, plastic sort of look I don't really dig.

SteelSlayer:
I like it, other than the white & blue strip part of arrivals.

S34NW:
Note: Sprites for the station name decal for all 3 stations are on their way 🚚🚚

SabreML (commented):
I honestly can't see the proposed changes here as an improvement to the server, even after staring at it for a few hours writing this. If anything I'd say it looks more like a straight downgrade.
I understand that this is what TG uses and therefore we should do the same and all that, but this just seems like change for the sake of change, with no real benefit behind it.
* The style of the sprites hasn't been changed at all, it's still just the same regular 2x2 grid for each floor tile.
* The perspective hasn't changed to 3/4, which is usually the main reason for taking sprites from TG. 
   (Although to be fair there's only so much you can do with perspective on a flat surface.)
* There's no variation or "texture" in the colour whatsoever, it's just four completely flat squares with lines between them.
* The shading in the middle of the four tiles, as strange as it looked, has been completely removed, making everything look less like actually finished "floor tile" sprites and more like the first colour pass of a WIP sprite before it was finished.
<hr>

Lots of other servers have different flooring styles to the one that we're currently using of course, but in every example I can find they do something unique and interesting with it.
<details>
<summary><b>Examples:</b></summary>
<hr>

### Citadel:
image
Very similar to TG's version with flat colours, but with a bright outline around the edges of each square to make them better fit in with the general server art style, and a separate floor type on the side of walls.
<hr>

### Eris:
image
Again similar to TG, even simpler looking actually, but with a fancy smoothing system. This fits in very well on the server thanks to the rest of the sprites all being very consistent and sharing the same general characteristics.
(Considering some other TG sprites, it's likely that they just copied the floor idea directly from Eris.)
<hr>

### Aurora:
image
Completely different art style to other servers I've seen, and they seem to have actually made the 'Lots of sprites from different sources' issue work for them.
<hr>

### Skyrat:
image
As far as I know these sprites were taken from another server initially, but they definitely work very well with everything else surrounding them. Individually the dotted checkerboard sort of pattern can be a bit distracting, but in context with the rest of the station I think they help to add a sort of "depth".


Comparing all of those to this, and I hope it's not just me who's a bit put off by it:
image
<hr>
</details>

There's an argument to be made that when we get all of the main sprites up to date the floor sprites will fit in better. That's probably true, but personally I'd much prefer that we actually get to a state ready to support them *before* swapping over rather than after.
<hr>

There are some things in this which I do like. I think the new reinforced floors really look great when compared to the current version, and the colour lines on tiles being smooth rather than blocky looks pretty nice.
I'm not against the idea of floor sprite overhauls in general, I'm very much in favour of them if anything, I just don't think this is the correct spriteset to use for the server.

Spacemanspark:
Tbh I'd be happy with just changing out our standard floor tiles for the dark colored ones. They look so much better.

S34NW:
> 
> 
> I honestly can't see the proposed changes here as an improvement to the server, even after staring at it for a few hours writing this. If anything I'd say it looks more like a straight downgrade.
> I understand that this is what TG uses and therefore we should do the same and all that, but this just seems like change for the sake of change, with no real benefit behind it.
> 
>     * The style of the sprites hasn't been changed at all, it's still just the same regular 2x2 grid for each floor tile.
> 
>     * The perspective hasn't changed to 3/4, which is usually the main reason for taking sprites from TG.
>       (Although to be fair there's only so much you can do with perspective on a flat surface.)
> 
>     * There's no variation or "texture" in the colour whatsoever, it's just four completely flat squares with lines between them.
> 
>     * The shading in the middle of the four tiles, as strange as it looked, has been completely removed, making everything look less like actually finished "floor tile" sprites and more like the first colour pass of a WIP sprite before it was finished.
> 
> 
> Lots of other servers have different flooring styles to the one that we're currently using of course, but in every example I can find they do something unique and interesting with it.
> Examples:
> ### Citadel:
> 
> image
> Very similar to TG's version with flat colours, but with a bright outline around the edges of each square to make them better fit in with the general server art style, and a separate floor type on the side of walls.
> ### Eris:
> 
> image
> Again similar to TG, even simpler looking actually, but with a fancy smoothing system. This fits in very well on the server thanks to the rest of the sprites all being very consistent and sharing the same general characteristics.
> (Considering some other TG sprites, it's likely that they just copied the floor idea directly from Eris.)
> ### Aurora:
> 
> image
> Completely different art style to other servers I've seen, and they seem to have actually made the 'Lots of sprites from different sources' issue work for them.
> ### Skyrat:
> 
> image
> As far as I know these sprites were taken from another server initially, but they definitely work very well with everything else surrounding them. Individually the dotted checkerboard sort of pattern can be a bit distracting, but in context with the rest of the station I think they help to add a sort of "depth".
> 
> Comparing all of those to this, and I hope it's not just me who's a bit put off by it:
> image
> 
> There's an argument to be made that when we get all of the main sprites up to date the floor sprites will fit in better. That's probably true, but personally I'd much prefer that we actually get to a state ready to support them _before_ swapping over rather than after.
> 
> There are some things in this which I do like. I think the new reinforced floors really look great when compared to the current version, and the colour lines on tiles being smooth rather than blocky looks pretty nice.
> I'm not against the idea of floor sprite overhauls in general, I'm very much in favour of them if anything, I just really don't think this is the correct spriteset to use for the server.

I do like skyrat and eris floor textures, and we in fact do have those skyrat textures now in the files (just without the coloured edging for the reasons I stated at the start of the PR) so once someone implements coloured edging for those textures there will be a lot more variety.
It really is a lot of work for one person to do, because we don't use the decal system other codebases do that make colouring floor tiles ridiculously easy. Every direction and combination of colours has to have its own hand-made iconstate on our codebase. Something very time intensive.

matttheficus:
Honestly the textures look very flat and "fake" as Spark put it. It might be due to the fact that the floors are surrounded by a bunch of non-TG sprites. Perhaps this would be best added after the rest of the server's sprites are updated/ported?

E-MonaRhg:
While I still think that these are better than what we have currently I do also get the opinions of many that these sprites look "bland", "fake", etc.

As such I want to put this out there instead of doing some offtopic comments on coding chat on discord. Our neighbours over at Russian Paradise have undertook a similar project, they did however choose the Skyrat/Oracle/TeuCeti/etc sprites. Let me bring in some examples (took with ingame xray because I dont know how to turn off pipes in SDMM):

image

image

image

image

image

tiles



I do feel that these have more... personality? compared to the TG sprites.
And yes I know that this was attempted in #10576 but we weren't in a middle of a resprite project back then (or so I assume) and well, maybe it is time to give these sprites a second chance.

Spacemanspark:
Oracle's floor sprites look like someone set windows to 16 colors. The darker variants are tolerable, but anything else makes my eyes bleed.

Sinestia:
They seem like they'll fit a lot better with some of the other walls that /tg/ sprites have teased, but on their own, they don't fit.

S34NW:
bogos binted

PR #16510, opened 2021-08-10
Make bee hitbox much smaller, except if the bee attacks someone

theo-3:
## What Does This PR Do
Removes opacity from bees.
If a bee attacks someone, it will be made opaque for one minute.
If a bee becomes sentient, it will permanently become opaque.
Syndicate bees are STILL opaque.
## Why It's Good For The Game
Trying to do botany with bees is very annoying because you have to constantly alt-click when bees are on trays.
Since it's much rarer for someone to want to click on a bee than to want to click on a plant tray, it's better to make the bee hitbox smaller.
Hostile bees are temporarily made opaque, and syndicate bees are always opaque, so this won't make bees harder to fight.
Sentient bees are also made opaque so they can't cause too much trouble.

## Changelog
:cl:
tweak: Bee hitbox no longer covers the entire tile by default. If that bee attacks someone, it covers the tile for one minute. Queen, syndie and sentient bees are not changed and still cover the tile.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/bees.dm
+++ b/code/modules/mob/living/simple_animal/hostile/bees.dm
@@ -31,7 +31,6 @@
 	move_to_delay = 0
 	obj_damage = 0
 	environment_smash = 0
-	mouse_opacity = MOUSE_OPACITY_OPAQUE
 	pass_flags = PASSTABLE | PASSGRILLE | PASSMOB
 	density = FALSE
 	mob_size = MOB_SIZE_TINY
@@ -162,6 +161,7 @@
 	else
 		. = ..()
 		if(. && isliving(target) && (!client || a_intent == INTENT_HARM))
+			make_opaque()
 			var/mob/living/L = target
 			if(L.reagents)
 				if(beegent)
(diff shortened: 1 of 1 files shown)


dearmochi:
Suggestion: rather than creating a timer, perhaps you could check in the bee's `Life` if its last attack is recent enough. This way you avoid the churn of constantly creating and overriding timers and its associated memory cost.

theo-3:
Wouldn't doing this in Life cause way more checks?

theo-3:
Right now it's one check per attack, Life would do it like 20 times per second, wouldn't it?

theo-3:
Hold on, memory cost, not CPU cost? Do overridden timers remain in memory and take up space?

dearmochi:
Not really, but under the hood creating a timer is rather complex (see `SStimer`) and is way more than just "one check" -- with the `Life` method, all you'd be doing periodically is compare two numbers, which is pretty inexpensive compared to the various checks and calls associated with creating a timer.

theo-3:
Alright, is it good like this?

code/modules/mob/living/simple_animal/hostile/bees.dm
```diff
@@ -170,6 +172,21 @@
...
+
+/mob/living/simple_animal/hostile/poison/bees/Life(seconds, times_fired)
+	. = ..()
+	var/diff = world.time - last_attack
+	if(diff >= opacity_time)
+		mouse_opacity = initial(mouse_opacity)
+
+/mob/living/simple_animal/hostile/poison/bees/sentience_act()
+	. = ..()
+	make_opaque()
```
dearmochi, line 188: This will be undone when enough time has passed, you'd need another variable or something for this case

code/modules/mob/living/simple_animal/hostile/bees.dm
```diff
@@ -170,6 +172,21 @@
...
 
+/mob/living/simple_animal/hostile/poison/bees/proc/make_opaque()
+	// If a bee attacks someone, make it very easy to hit for a while
+	last_attack = world.time
+	mouse_opacity = MOUSE_OPACITY_OPAQUE
+
+/mob/living/simple_animal/hostile/poison/bees/Life(seconds, times_fired)
+	. = ..()
+	var/diff = world.time - last_attack
+	if(diff >= opacity_time)
```
dearmochi, line 183: This will constantly set the mouse opacity back, better if it was set only once

dearmochi (changes requested):

theo-3:
Whoops. Did that fix it? Sorry

dearmochi (approved):

S34NW:
🐝

hal9000PR:
> 🐝

oh god

Kyep (approved):

variableundefined (approved):

PR #16517, opened 2021-08-11
The Great Delta Secpoint Removal

LightFire53:
## What Does This PR Do

This PR removes the security checkpoints from around the NSS Kerberos, and converts them to break rooms in most cases. In the case of medical, I moved the psychiatrists office down, and converted the unused space into a maintenance closet area. I also added a radiation shower space below the psych office to make it not seem so huge. For science, since they had an existing break room, I removed the existing one and made it "run down" to fit the general grunge feel to maintenance. Engineering had a break nook that I adjusted to just be some extra tools and a free floorbot.

This PR also (as a result of me adding some extra disposals) fixes a the CE and Engineering disposal junctions, as well as making it so that garbage doesn't end up in the law office.

I also added some extra holopads to medbay where they seemed to be lacking, as I was already adding holopads to parts of medical.

## Why It's Good For The Game

Map parity. Cyberiad doesn't have security checkpoints, Kerberos shouldn't either.

## Images of changes

**Old Research**
Research Old

Research Old 2

**New Research**
Research New

Research New 2

**Old Engineering**
Engi Old 2

Engi Old

**New Engineering**
Engi New

Engi New 2

**Old Cargo**
Cargo Old

**New Cargo**
Cargo New

**Old Psych**
Psych Old

**New Psych**
Psych New

## Changelog
:cl:LightFire53
add: NSS Kerberos departmental break rooms
del: NSS Kerberos departmental Security Checkpoints
tweak: NSS Kerberos Psych Office
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -11967,11 +11967,11 @@
 /turf/simulated/floor/plasteel,
 /area/hydroponics/abandoned_garden)
 "aGO" = (
-/obj/structure/disposalpipe/junction{
-	dir = 4;
-	icon_state = "pipe-y"
-	},
 /obj/effect/decal/cleanable/dirt,
+/obj/structure/disposalpipe/junction{
+	dir = 2;
+	icon_state = "pipe-j2"
+	},
 /turf/simulated/floor/plating,
 /area/quartermaster/sorting)
 "aGP" = (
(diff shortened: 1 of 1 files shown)


variableundefined (approved):
We tried departmental security before and removed it. 

Approved.

SteelSlayer:
> Map parity. Cyberiad doesn't have security checkpoints, Kerberos shouldn't either.

This is a bad way to go about things in my opinion. These sort of features set maps apart from cyberiad. If we use the above reason to make all maps conform to it, we’ll essentially just have cyberiad v2, cyberiad v3, etc.

Coolrune206:
I like departmental Sec checkpoints - it was pretty much the only thing about the map that made me think "this is a good thing".

Map feels large and sparse enough that I think the checkpoints work.

LightFire53:
> > Map parity. Cyberiad doesn't have security checkpoints, Kerberos shouldn't either.
> 
> This is a bad way to go about things in my opinion. These sort of features set maps apart from cyberiad. If we use the above reason to make all maps conform to it, we’ll essentially just have cyberiad v2, cyberiad v3, etc.

These are also for a role we don't have/support, therefore just kinda making them dead space. By converting them to break rooms, we make them usable space, and still maintain something Cyberiad doesn't.

S34NW:
Erm, can I request you move the psych back into maints? @LightFire53 
Psych being in a main hallway is really good, because medbay psych makes it an utter pain for people to go there voluntarily.

PR #16522, opened 2021-08-11
Remove SSshuttle var referencing undefined type.

warriorstar-orion:
## What Does This PR Do
Remove SSshuttle var referencing undefined type. `/datum/round_event/shuttle_loan` appears to be a tg event, and this type isn't referred to anywhere else in the codebase.

## Why It's Good For The Game
Keeping a reference to an undefined type in the codebase feels like a bad idea, and is confusing, and, tangentially, if you happen to be writing a parser that checks for existence of all types referenced in the AST, this breaks it.

### DIFF
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -37,7 +37,6 @@ SUBSYSTEM_DEF(shuttle)
 	var/list/shoppinglist = list()
 	var/list/requestlist = list()
 	var/list/supply_packs = list()
-	var/datum/round_event/shuttle_loan/shuttle_loan
 	var/sold_atoms = ""
 	var/list/hidden_shuttle_turfs = list() //all turfs hidden from navigation computers associated with a list containing the image hiding them and the type of the turf they are pretending to be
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/shuttles.dm
```diff
@@ -37,7 +37,6 @@ SUBSYSTEM_DEF(shuttle)
 	var/list/shoppinglist = list()
 	var/list/requestlist = list()
 	var/list/supply_packs = list()
-	var/datum/round_event/shuttle_loan/shuttle_loan
```
AffectedArc07, line 40: How the hell is this passing CI when we dont have this type?


I am mystified.

warriorstar-orion: That's sort of what I would have expected, too. Perhaps it's not a case that dreamchecker checks/cares about?

SabreML: It does look like byond just ignores this sort of thing entirely for whatever reason, no errors or anything. Definitely should be removed from here though either way. 👍
image

AffectedArc07 (approved):

PR #16529, opened 2021-08-12
Ensure only pistol-type weapons can be holstered

### LINKED ISSUE #16518: Foam Enforcer does not go in the shoulder holster
**Issue Description**:

The foam enforcer, despite being a pistol, is not holsterable, unlike other pistols.

To holster my handgun.

**What happened instead**:

_This Foam Enforcer won't fit in the shoulder holster!_

Inconsistent with other pistols and doesn't make a lot of sense.

1. Get holster
2. Get foam enforcer
3. Try to holster
4. Cry

Probably always did

**Extra information**:

I'm sure this comes up a lot

### LINKED ISSUE #16542: Hybrid Turret fits in Shoulder Holster
**Issue Description**:

Hybrid Turret Guns are big, bulky, and require two hands to fire, yet you can put them in a shoulder holster.

The super big gun should not fit in a shoulder holster,

**What happened instead**:

They go into the shoulder holster.

It feels wrong that a silenced stechkin or security autorifle do not fit but the super big, two-handed gun can.

Get a hybrid turret gun, get a shoulder holster from botany, holster the gun.

I found out about a month ago but it's probably been this way for awhile.

chafla:
## What Does This PR Do

First inception of this PR fixes #16518, where foam force pistols (including the foam enforcer) could not be holstered, despite the description specifically stating that it's `A small, easily concealable toy handgun.` Turns out the rabbit hole went a bit deeper, and now we're reworking holstering in general to be disabled by default.
The base gun datum had `can_holster = TRUE`, which trickled down to all energy guns, spells, and everything else that didn't disable it by default. This ensures that items being holstered is a manual process.

(Fixes #16542 as well)

Also fixes a typo in the foam enforcer description while I'm looking there.

## Why It's Good For The Game

The purpose of a sholder holster seems to be to store small, single-hand weapons, and the notion of what can and can't be holstered seems to have been left wide open by default. This hopes to fix that (and hopefully let foam guns be holstered).

## Images of changes

img

## Changelog
:cl:
fix: Foam force pistols (including enforcer), as well as revolvers, can now be holstered.
fix: Only pistol-like energy guns can be holstered, rather than every single one.
fix: Wands and wizard items can no longer be holstered.
spellcheck: Fix grammar in foam enforcer description.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/toy.dm
+++ b/code/modules/projectiles/guns/projectile/toy.dm
@@ -26,6 +26,7 @@
 	can_suppress = 0
 	burst_size = 1
 	fire_delay = 0
+	can_holster = TRUE
 	actions_types = list()
 
 /obj/item/gun/projectile/automatic/toy/pistol/update_icon()
--- a/code/modules/projectiles/guns/projectile/toy.dm
+++ b/code/modules/projectiles/guns/projectile/toy.dm
@@ -43,7 +43,7 @@
 	..()
 
 /obj/item/gun/projectile/automatic/toy/pistol/enforcer
-	name = "Foam Enforcer"
+	name = "foam Enforcer"
 	desc = "A foam-shooting version of the Enforcer meant to be used for training new cadets who can't be trusted with rubber bullets."
 	icon_state = "enforcer"
 	mag_type = /obj/item/ammo_box/magazine/toy/enforcer
--- a/code/modules/projectiles/guns/projectile/toy.dm
+++ b/code/modules/projectiles/guns/projectile/toy.dm
@@ -44,7 +44,7 @@
 
 /obj/item/gun/projectile/automatic/toy/pistol/enforcer
 	name = "Foam Enforcer"
-	desc = "A foam shooting version of the Enforcer meant to be used for training new caddets who can't be trusted with rubber bullets."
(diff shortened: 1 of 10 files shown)


SabreML (approved):

matttheficus:
It might be a bit out of scope unless you swap the title a bit, but could you add the various versions of the revolver type to the holster-able list as well (Mateba, Golden, Nagant, Russian, Shock [this might be in already])?

chafla:
Yeah, that seems reasonable. I'll update and include that.

chafla:
So it turns out that energy weapons are holsterable by default, even chunky ones like xray lasers or turret lasers, since they never override the can_holster = TRUE set in the base datum. I'm hesitant to address that in this PR, since it seems a bit out of scope and might even be a balance issue, but I'd be happy to address that while I'm poking through here.

code/modules/projectiles/guns/projectile/toy.dm
```diff
@@ -43,7 +44,7 @@
 
 /obj/item/gun/projectile/automatic/toy/pistol/enforcer
 	name = "Foam Enforcer"
```
dearmochi, line 46:
```suggestion
	name = "foam Enforcer"
```

dearmochi (approved):
while you're at it, LGTM otherwise

Kyep (approved):

variableundefined (approved):

PR #16530, opened 2021-08-12
Fixes grammer / bugs with supermater sliver and nuke core objectives

Qwertytoforty:
## What Does This PR Do

Removes extra period on supermater sliver / nuke core objectives.

Adds a variable to steal objectives. If this is false, the protected jobs of an objective will not be listed as having the steal item in their protection. This is used for the supermatter sliver objective, as atmos techs / engineers / the CE do not carry around supermater slivers.

## Why It's Good For The Game

Bugs / extra periods bad, lying about the objective is bad.

## Changelog
:cl:
fix: Fixed extra periods in the supermater sliver / nuke core objective
fix: Fixed the supermater sliver objective saying the CE/ engineers/ atmos techs might carry the sliver around on them
/:cl:

### DIFF
--- a/code/game/gamemodes/steal_items.dm
+++ b/code/game/gamemodes/steal_items.dm
@@ -14,6 +14,8 @@
 	var/location_override
 	/// Do we have a special item we give to somewhen when they get this objective?
 	var/special_equipment = null
+	///If a steal objective has forbiden jobs, and the forbiden jobs would not be in the possession of this item, set this to false
+	var/job_possession = TRUE
 
 /datum/theft_objective/proc/check_completion(datum/mind/owner)
 	if(!owner.current)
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/steal_items.dm
```diff
@@ -14,6 +14,8 @@
 	var/location_override
 	/// Do we have a special item we give to somewhen when they get this objective?
 	var/special_equipment = null
+	///If a steal objective has forbiden jobs, and the forbiden jobs would not be in the possession of this item, set this to false
```
AffectedArc07, line 17:
```suggestion
	/// If a steal objective has forbiden jobs, and the forbiden jobs would not be in the possession of this item, set this to false
```

AffectedArc07 (approved):

code/game/gamemodes/steal_items.dm
AffectedArc07: Misclick. Pretend this was a requested change.

Qwertytoforty: already done

code/game/gamemodes/steal_items.dm
```diff
@@ -14,6 +14,8 @@
 	var/location_override
 	/// Do we have a special item we give to somewhen when they get this objective?
 	var/special_equipment = null
+	/// If a steal objective has forbiden jobs, and the forbiden jobs would not be in the possession of this item, set this to false
```
SteelSlayer, line 17: Spellcheck review :dab:
```suggestion
	/// If a steal objective has forbidden jobs, and the forbidden jobs would not be in the possession of this item, set this to false
```

PR #16532, opened 2021-08-12
Caps runechat emotes to 100 chars

dearmochi:
## What Does This PR Do

- Caps runechat emotes to 100 chars

## Why It's Good For The Game

- We don't need huge runechats above peoples head

## Images of changes

dreamseeker_VUcRBJ9nQX

## Changelog
:cl:
tweak: Runechat emotes are now capped to 100 characters
/:cl:

### DIFF
--- a/code/modules/mob/emote.dm
+++ b/code/modules/mob/emote.dm
@@ -77,7 +77,9 @@
 
 		// Type 1 (Visual) emotes are sent to anyone in view of the item
 		if(m_type & EMOTE_VISUAL)
-			var/runechat_text = copytext(input, 1, 101)
+			var/runechat_text = input
+			if(length(input) > 100)
+				runechat_text = "[copytext(input, 1, 101)]..."
 			var/list/can_see = get_mobs_in_view(1,src)  //Allows silicon & mmi mobs carried around to see the emotes of the person carrying them around.
 			can_see |= viewers(src,null)
 			for(var/mob/O in can_see)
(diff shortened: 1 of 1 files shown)


code/modules/mob/emote.dm
```diff
@@ -77,6 +77,7 @@
 
 		// Type 1 (Visual) emotes are sent to anyone in view of the item
 		if(m_type & EMOTE_VISUAL)
+			var/runechat_text = copytext(input, 1, 101)
```
AffectedArc07, line 80: May be worth appending an ellipsis if it did get cutoff.

code/modules/mob/emote.dm
```diff
@@ -77,6 +77,9 @@
 
 		// Type 1 (Visual) emotes are sent to anyone in view of the item
 		if(m_type & EMOTE_VISUAL)
+			var/runechat_text = input
+			if(length(input) > 100)
+				runechat_text = "[copytext(input, 1, 101)]..."
```
SabreML, line 82: There seems to be a system for this already in `/datum/chatmessage/proc/generate_image()` which shortens regular say messages. The only reason it's not doing the same for emotes is because they don't have a span class assigned to them, so `split_html()` doesn't return an `inner_text` value.
https://github.com/ParadiseSS13/Paradise/blob/2f5951bbaa2ddc94696c0f4a04b7bda70f232fc8/code/datums/chatmessage.dm#L100-L110
It might be better to use this existing system rather than making a new one for emotes, assuming that wouldn't end up being too convoluted.

PR #16520, opened 2021-08-11
Fix bundle #4

SabreML:
## What Does This PR Do

Please see the changelog for a list, because there's no need to type all of that again.
Each fix is in a separate commit for ease of review.

Fixes #16508.
Fixes #16471.
Fixes #16466.
Fixes #16438.
Fixes #16412.
Fixes #16279.
Fixes #16041.
Fixes #14312.
Closes #15218. (Headcrabs were removed.)
Closes #14830. (Fixed in #15535.)
Closes #14753. (Seems to be a user error.)
Closes #14506. (Fixed in #16338.)
Closes #14313. (Fixed in #16411.)
Closes #14276. (Seems to have been fixed.)
Closes #14270. (IPCs can drink reagents.)
Closes #7103. (Fixed in #14316.)
(I *think* that's all of them.)

## Changelog
:cl:
add: Added Blob-infected players and mice to the 'Antagonist' orbit menu if AntagHUD is enabled.
add: Added Revenants to the 'Antagonist' orbit menu even if AntagHUD is disabled.
tweak: Made progress bars show above regular HUD elements.
tweak: Added ant-repellant to the wrappers of vending machine food.
tweak: Made Vulpkanin mimes growl silently.
tweak: Made the Possessed Blade ghost poll actually say 'Possessed Blade', rather than 'pAI'.
fix: Fixed 'APC Access ID cards' not having a sprite set.
fix: Fixed being able to pat out a fire on yourself.
fix: Fixed gibbers vanishing after being deconstructed.
fix: Fixed using 'Epinephrine Overdose' as a changeling to wake up causing permanent blindness in some circumstances.
fix: Fixed package wrappers being able to wrap the same package multiple times.
fix: Fixed being able to teleport items around the station using the above package wrappers.
fix: Fixed Drask being unable to wear pink Hawiian shirts.
fix: Fixed being able to teleport station signs by using a screwdriver on them.
fix: Fixed various vending machines causing runtimes when reconstructed.
fix: Fixed hybrid turret inhands being flipped.
fix: Fixed some screen colour flash effects during megafauna fights acting as a cure for colourblindness.
fix: Fixed disabling night shift mode in the config not working.
spellcheck: Replaced "Targetted" with "Targeted" in the rogue drone event announcement.
/:cl:

PR #16533, opened 2021-08-12
[DNM] Cooldown timer system refactor

SabreML:
## What Does This PR Do

Ports the 'Timer cooldown' system and `world.time` timer macros from TG, and refactors most existing cooldowns to use it. (https://github.com/tgstation/tgstation/pull/51388)
*(Also adds some documentation, some stolen from TG)*

## Why It's Good For The Game

This adds two new primary systems, `addtimer()` cooldowns and `world.time` cooldowns.
Overall these make the current cooldown/timer code used across the codebase far easier to understand both in reading and writing due to using an easier standardised `START`, `END/RESET`, and `FINISHED/TIMELEFT` system for everything.

### Timer cooldowns:
**Pros:**
- Sends a signal to the parent datum at the end of the cooldown.
- Doesn't require any variables to be defined, so they're far better at tracking individual objects or rare edge-cases.
- The 'Stoppable' variant is, as the name implies, stoppable. So it can be used for more complex interactions such as interruptable channels.

**Cons:**
- Slightly less performant than the `world.time` system, since they all use `addtimer()` to track the duration.
- Unless the stoppable system is used, there's no way to check how long is left on the cooldown.
- Requires arbritrary strings as list indicies, which can get confusing if defines aren't used.

**Example:**
```dm
#define COOLDOWN_INTERACT "interact"

/mob/Initialize(mapload)
	. = ..()
	RegisterSignal(src, COMSIG_CD_STOP(COOLDOWN_INTERACT), .proc/timer_end)

/mob/proc/timer_end()
	SIGNAL_HANDLER
	do_other_thing()

/mob/proc/interact()
	if(TIMER_COOLDOWN_CHECK(src, COOLDOWN_INTERACT))
		return
	TIMER_COOLDOWN_START(src, COOLDOWN_INTERACT, 30 SECONDS)
	do_thing()

#undef COOLDOWN_INTERACT
```
<hr>

### world.time cooldowns:
**Pros:**
- Very quick and easy to use with only a couple of defines.
- Able to have their remaining time checked and modified after being started.
- Small performance impact, since it's all just variable comparisons.

**Cons:**
- No fancy interactions are possible unless called by a separate proc in `process()` or similar, these just store a number and nothing else.

**Example:**
```dm
/mob
	COOLDOWN_DECLARE(cooldown_timer)

/mob/proc/interact()
	if(!COOLDOWN_FINISHED(src, cooldown_timer))
		return
	COOLDOWN_START(src, cooldown_timer, 30 SECONDS)
	do_thing()
```

## Changelog
:cl:
experiment: Refactored the cooldown system to be more standardised and user-friendly for coders.
/:cl:


code/game/gamemodes/miniantags/guardian/types/assassin.dm
```diff
@@ -21,8 +21,9 @@
 /mob/living/simple_animal/hostile/guardian/assassin/Stat()
 	..()
 	if(statpanel("Status"))
-		if(stealthcooldown >= world.time)
-			stat(null, "Stealth Cooldown Remaining: [max(round((stealthcooldown - world.time)*0.1, 0.1), 0)] seconds")
+		var/time_left = round(COOLDOWN_TIMELEFT(src, stealth_cooldown) / 10)
```
dearmochi, line 24: Can this go below 0? If so, `if(time_left)` can be truthy even if you don't intend it to and thus show up on the stat panel

code/game/gamemodes/miniantags/guardian/types/bomb.dm
```diff
@@ -7,14 +7,15 @@
...
-	var/bomb_cooldown = 0
+	COOLDOWN_DECLARE(bomb_cooldown)
 	var/default_bomb_cooldown = 20 SECONDS
 
 /mob/living/simple_animal/hostile/guardian/bomb/Stat()
 	..()
 	if(statpanel("Status"))
-		if(bomb_cooldown >= world.time)
-			stat(null, "Bomb Cooldown Remaining: [max(round((bomb_cooldown - world.time)*0.1, 0.1), 0)] seconds")
+		var/time_left = round(COOLDOWN_TIMELEFT(src, bomb_cooldown) / 10)
```
dearmochi, line 16: Ditto

code/game/gamemodes/miniantags/guardian/types/healer.dm
```diff
@@ -40,8 +40,9 @@
 /mob/living/simple_animal/hostile/guardian/healer/Stat()
 	..()
 	if(statpanel("Status"))
-		if(beacon_cooldown >= world.time)
-			stat(null, "Bluespace Beacon Cooldown Remaining: [max(round((beacon_cooldown - world.time)*0.1, 0.1), 0)] seconds")
+		var/time_left = round(COOLDOWN_TIMELEFT(src, beacon_cooldown) / 10)
```
dearmochi, line 43: Ditto

AffectedArc07:
Leaving this here
```
It is expected that contributors discuss larger design changes on the forums prior to coding a Pull Request. The amount of time spent on any given Pull Request is not relevant. Maintainers are not responsible for contributors wasting their time creating features nobody asked for.
```

The above text applies to *any* PR, not just features


This PR is going to take a long time to review, when review capacity is already at a minimum. We have asked you before (see: reagents rework), so we will ask yet again. **Ask the maintainers before making massive refactor PRs**

code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -120,16 +120,16 @@
...
 				M.death()
 				M.splat()
 				movement_target = null
 				stop_automated_movement = 0
 				break
 		for(var/obj/item/toy/cattoy/T in view(1, src))
-			if(T.cooldown < (world.time - 400))
-				custom_emote(1, "bats \the [T] around with its paw!")
-				T.cooldown = world.time
+			if(!TIMER_COOLDOWN_CHECK(src, "cat_toy"))
```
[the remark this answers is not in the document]
SabreML: The intention here was to avoid making a whole new variable on `/mob/living/simple_animal/pet/cat` just for cat toys since that's an option now. Although that said there's already an `eats_mice` variable there, so the `world.time` ones would actually work better for this.

SabreML:
> Leaving this here
> 
> ```
> It is expected that contributors discuss larger design changes on the forums prior to coding a Pull Request. The amount of time spent on any given Pull Request is not relevant. Maintainers are not responsible for contributors wasting their time creating features nobody asked for.
> ```
> 
> The above text applies to _any_ PR, not just features
> 
> This PR is going to take a long time to review, when review capacity is already at a minimum. We have asked you before (see: reagents rework), so we will ask yet again. **Ask the maintainers before making massive refactor PRs**

I *did* ask, both before I started on it and before I posted the PR. The first time both you and @variableundefined were actively talking in the same channel so I assumed that if there *were* any issues with it then they would've been brought up.
https://discord.com/channels/145533722026967040/145700319819464704/875063389959888916
<details>
<summary><b>Message 1:</b></summary>

image
</details>

The second time, just to make absolutely sure, I asked again and then waited a full hour before submitting the PR to avoid a repeat of the reagent refactor, again with no response. I didn't think it would be neccesary to bug you all with a `@Maintainers` ping considering the lack of response to the first message.
https://discord.com/channels/145533722026967040/145700319819464704/875477411389665280
<details>
<summary><b>Message 2:</b></summary>

image
</details>

I asked twice in #coding_chat to avoid a repeat of #16385, and the messages combined were seen by multiple maintainers and PR reviewers with no negative comments. None of the modified files have code owners either, so I took that all as permission to post the PR.

There's no in-game changes which would come from this and it's not a feature so I hadn't considered it to be a "Design change" as specified in the code of conduct. If it was something like porting over span class defines from TG then obviously I'd make a forum post about it beforehand, but in this case it falls far more under the scope of just increasing the code's readability than anything else.

If this had just been a PR changing 5-6 files instead of 104 then this wouldn't be a problem. Unless there's a time limit I'm unaware of for PRs then I'm not sure what the issue is in that regard. Having reviewed the whole thing myself a couple of days ago I know that it's not exactly an easy one, but it makes no difference to me personally if it's reviewed in 2 days or 2 months so there's definitely no pressure from my side at least.

code/game/gamemodes/miniantags/guardian/types/assassin.dm
SabreML: The `COOLDOWN_TIMELEFT()` define returns a minimum of 0, so it'll either be running or finished for all of these.
```dm
#define COOLDOWN_TIMELEFT(cd_source, cd_index) max(0, cd_source.cd_index - world.time)
```

AffectedArc07:
> > Leaving this here
> > ```
> > It is expected that contributors discuss larger design changes on the forums prior to coding a Pull Request. The amount of time spent on any given Pull Request is not relevant. Maintainers are not responsible for contributors wasting their time creating features nobody asked for.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The above text applies to _any_ PR, not just features
> > This PR is going to take a long time to review, when review capacity is already at a minimum. We have asked you before (see: reagents rework), so we will ask yet again. **Ask the maintainers before making massive refactor PRs**
> 
> I _did_ ask, both before I started on it and before I posted the PR. The first time both you and @variableundefined were actively talking in the same channel so I assumed that if there _were_ any issues with it then they would've been brought up.
> https://discord.com/channels/145533722026967040/145700319819464704/875063389959888916
> 
> Message 1:
> The second time, just to make absolutely sure, I asked again and then waited a full hour before submitting the PR to avoid a repeat of the reagent refactor, again with no response. I didn't think it would be neccesary to bug you all with a `@Maintainers` ping considering the lack of response to the first message.
> https://discord.com/channels/145533722026967040/145700319819464704/875477411389665280
> 
> Message 2:
> I asked twice in #coding_chat to avoid a repeat of #16385, and the messages combined were seen by multiple maintainers and PR reviewers with no negative comments. None of the modified files have code owners either, so I took that all as permission to post the PR.
> 
> There's no in-game changes which would come from this and it's not a feature so I hadn't considered it to be a "Design change" as specified in the code of conduct. If it was something like porting over span class defines from TG then obviously I'd make a forum post about it beforehand, but in this case it falls far more under the scope of just increasing the code's readability than anything else.
> 
> If this had just been a PR changing 5-6 files instead of 104 then this wouldn't be a problem. Unless there's a time limit I'm unaware of for PRs then I'm not sure what the issue is in that regard. Having reviewed the whole thing myself a couple of days ago I know that it's not exactly an easy one, but it makes no difference to me personally if it's reviewed in 2 days or 2 months so there's definitely no pressure from my side at least.



A lack of replies means that messages likely got missed. As a rule of thumb, never treat "no response" as "no issue". You also didn't make any direct pings towards maintainers, meaning it was even easier for this to slip under.

code/game/gamemodes/blob/theblob.dm
```diff
@@ -1,19 +1,23 @@
 //I will need to recode parts of this but I am way too tired atm
 /obj/structure/blob
 	name = "blob"
+	desc = "Some blob creature thingy"
 	icon = 'icons/mob/blob.dmi'
+	anchored = TRUE
```
Aurorablade, line 6: is it needed to remove the Density and opacity values for the blob and not simply just set the current value to a true false rather then a numeric? And how is it related to the timer PR?

code/game/gamemodes/miniantags/guardian/guardian.dm
```diff
@@ -165,23 +165,21 @@
...
 		new /obj/effect/temp_visual/guardian/phase(loc)
 		reset_perspective()
-		cooldown = world.time + 30
+		COOLDOWN_START(src, manifest_cooldown, 3 SECONDS)
 
 /mob/living/simple_animal/hostile/guardian/proc/Recall(forced = FALSE)
-	if(!summoner || loc == summoner || (cooldown > world.time && !forced))
+	if(!summoner || loc == summoner || (!COOLDOWN_FINISHED(src, manifest_cooldown) && !forced))
 		return
-	if(!summoner) return
```
Aurorablade, line 180: this for me seems to be a needed last ditch fallback i could be wrong.

code/game/gamemodes/blob/theblob.dm
SabreML: In this case, both the [`density`](http://www.byond.com/docs/ref/#/atom/var/density) and [`opacity`](http://www.byond.com/docs/ref/#/atom/var/opacity) variables are already set to `0`/`FALSE` by default. Since I was editing the `/obj/structure/blob`'s variables already, I didn't see any harm in removing those two in order to clean up a bit since they're just redundant.
Code_n7DdXRYEJX Code_m6bQB09vFA

code/game/gamemodes/miniantags/guardian/guardian.dm
SabreML: It may have originally been intended that way, but as there's another `summoner` check already two lines above, this one will never actually fail.
https://github.com/ParadiseSS13/Paradise/blob/0ce6bba543b0dc8c776f8a39913424ace071a644/code/game/gamemodes/miniantags/guardian/guardian.dm#L177-L184
I've not been playing for long enough to know for sure, but if there was an "Are you sure you want to recall?" input between these two lines in the past, then that would explain the extra check. Since the `summoner` could have been deleted in the wait time.
Either way, since there's currently no sleeps or anything between L178 and L181, the check won't ever be nessecary.

SabreML:
After the disaster that was #16520, every one of my open PRs was set to "Do Not Merge" until I could go through each of them individually and double check that I didn't miss anything stupid.

All of them except for this and #16385 have now been unset, and I was planning do these two at some point this week.
Unfortunately, due to college and some other personal issues, I don't currently have the time nor the motivation to test each of the 386 edited files for the fourth time, as well as every possible interaction and edge case with the new code.

I will get back to these at some point, but right now it's better to just close both in the interests of getting them out of the way.

PR #16501, opened 2021-08-08
'Unrestricted access from direction' for windoors

SabreML:
## What Does This PR Do

Adds the 'Unrestricted access from direction' functionality to windoors.

## Why It's Good For The Game

It seems like an oversight for them not to have this feature when regular airlocks do, and it'll allow people to use windoors in builds where they wouldn't have been able to otherwise.

## Changelog
:cl:
add: Added the 'Unrestricted access from direction' functionality to windoors.
/:cl:

PR #16538, opened 2021-08-13
Fixes directional access on the Medbay and Security doors

### LINKED ISSUE #16537: All one-way access airlocks now start without the one-way access
**Issue Description**:

All the one-way access airlocks (medbay entrances, cloning, brig cells entrance) aren't properly set at shift start, and can't be opened from the inside as they should.

Bump into door, door open

**What happened instead**:

wub wub wub *blinks red

Seems to be a bug brought by a recent PR, an unintentional change.

Bump into door without medbay/sec access from inside, be sad.

First noticed on the 12/08/2021

**Extra information**:
 the airlock boards simply weren't set-up right, they were otherwise fully functional.

SabreML:
## What Does This PR Do

Fixes #16537, which was caused in #16501 due to me not noticing that firelocks and lockdown poddoors are also door subtypes.

For the record I *did* test that PR a lot, but only with new mapedited airlocks and windoors, so they never had any firelocks under them.

## Why It's Good For The Game

Mistake on my part, sorry.

## Changelog
:cl:
fix: Fixed the Medbay and Security front doors not having directional access set.
/:cl:

### DIFF
--- a/code/game/objects/effects/mapping_helpers.dm
+++ b/code/game/objects/effects/mapping_helpers.dm
@@ -83,7 +83,9 @@
 	if(!mapload)
 		log_world("### MAP WARNING, [src] spawned outside of mapload!")
 		return
-	var/obj/machinery/door/door = locate(/obj/machinery/door) in loc
+	var/obj/machinery/door/door = locate(/obj/machinery/door/airlock) in loc
+	if(!door)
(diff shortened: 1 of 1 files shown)


farie82 (approved):
Should do the job fine enough

dearmochi (approved):

AffectedArc07 (approved):

PR #16539, opened 2021-08-13
Replaces panic bunker with queue server

AffectedArc07:
You thought I wasn't being serious?

## What Does This PR Do
Replaces panic bunker with a queue server. The actual queueing implementation will be done on the bouncer, but this is the general gist of things:
- **EXISTING PLAYERS WILL NOT BE AFFECTED BY THE QUEUE**
- **I REPEAT. EXISTING PLAYERS WILL NOT BE AFFECTED BY THE QUEUE**
- If you have never been seen before, the server is polled to see if you are to be queued or not
- If the server player count is above the queue limit (dynamically settable), you are placed into the queue
- From here, you then wait in the queue until the server has room for you
- Alternatively, an admin can force-whitelist your ckey for you to queue skip
- The queue will only let in 1 player every 3 seconds to avoid a mass letin from reconnect delays
- **AGAIN. EXISTING PLAYERS WILL NOT BE AFFECTED BY THE QUEUE**

All of this is configurable via a verb requiring `R_SERVER` rights. Settings can also apply to only one round, or persist through others.

image

## Why It's Good For The Game
Panic bunker, while useful, has its meh points for me. For example, if you get denied straight away, you might not have incentive to come back. You also have to keep manually reconnecting. This system allows players to just connect to a subserver and wait for them to be let in, greatly reducing frustration. You also get to know how far along you are.

## Changelog
:cl: AffectedArc07
add: Added a server queue system
del: Removed panic bunker
/:cl:


AffectedArc07:
Drafted until I have the bouncer part done. ETA 1 day.

S34NW:
The mad lad did it

NullFag:
The oldest queue server in history of SS13.

AffectedArc07:
> The oldest queue server in history of SS13.

Thats it the queue text is orange now.

Kyep (approved):

Kyep (changes requested):
I like the idea of there being a queue.
I like the idea of people not having to manually reconnect.

But I think I still have to object to this on the grounds that it it allows the server player cap to be lowered below 150 without head approval.

Spacemanspark:
Why not, however?

The current cap of 150 is... laughable, at best. Hell, we shouldn't even be subjected to more than 100 at a time. More than that is absolutely ludicrous to deal with, and considering this won't impact new players, I fail to see the issue with GAs setting a limit they're capable of managing. Sometimes other admins just aren't able to come online, and, let me tell you, solominning 150 players as one admin is _**not fun.** _

KasparoVy:
> 
> 
> Why not, however?
> 
> The current cap of 150 is... laughable, at best. Hell, we shouldn't even be subjected to more than 100 at a time. More than that is absolutely ludicrous to deal with, and considering this won't impact new players, I fail to see the issue with GAs setting a limit they're capable of managing. Sometimes other admins just aren't able to come online, and, let me tell you, solominning 150 players as one admin is _**not fun.**_

Not fun? I don't even know how it's POSSIBLE without devolving into utter chaos D:
I'm all for GAs being able to make that call for the sake of admin sanity lol

AffectedArc07:
> considering this won't impact new players

@Spacemanspark *existing players.

AffectedArc07:
> Objection.
> 
> 
> 
> I like the idea of there being a queue.
> 
> I also like the idea of people not having to manually reconnect.
> 
> 
> 
> But I'm utterly opposed to allowing the cap to be changed from 150.
> 
> Especially persistently, by non-heads. GAs have no business changing the player cap of the server.
> 
> If the player cap of the server is to be changed, that's something that should require at the absolute minimum 2/3 of the heads to sign off on it.
> 
> It isn't something a GA should be able to do, especially not persistently.


@Kyep in regards to the changing verb. Sometimes there's situations where GAs would need to be able to adjust this for the sake of not only their sanity, but for the playerbase as well.

Think all the way back to the ssethtide, adminning 150 is painful enough (I solo adminned **180** people before even being considered an admin, it was awful). If a major tide happens again from someone who has an absolutely unbearable fanbase (Take pyrocynical for example, who by the way, is aware of this game), admins should be able to drop the cap to a lower value at once to avoid the server standards absolutely plummeting, just to quench the initial wave.

Most of the shitters would get instant RDM bans, we've seen this happen before. This just allows it to be spread over a longer time period as opposed to having to suffer through 20 people RDMing at the same time.

I personally have no issue with the option being there, hence why I coded it. There's also nothing stopping you from saying in staff announcements "Do not change this threshold unless you have a very good reason to", and given that this logs to disk, it's very easy to find if it was changed, and who did it, especially with Kibana.

AffectedArc07:
image

Bouncer component done. Not deployed yet in production.

Necaladun:
Seems a much better option that straight denial, quite in support in this. 
Being able to change it on the fly is a good, if the server is getting overwhelmed without enough GAs to handle it, it needs to be changed then - not whenever we get around to changing the config.

Necaladun:
The problem with changing it on the fly is if someone sets it low then logs out - can we get an alert to what it's set at when admins log in?
[ADMIN LOG] Queue set to (number). Use (verb) to change this!
or something

AffectedArc07:
> The problem with changing it on the fly is if someone sets it low then logs out - can we get an alert to what it's set at when admins log in?
> [ADMIN LOG] Queue set to (number). Use (verb) to change this!
> or something

Can easily display a login notice to the admin when they login:

`Server que is currently [enabled/disabled] with a threshold of [x]. This [will/will not] persist through rounds.`

AffectedArc07:
@Necaladun admins will now see this on login, below all other warnings
image

code/modules/world_topic/queue_status.dm
```diff
@@ -0,0 +1,36 @@
...
+		else // Otherwise
+			// If we have more than the threshold, queue
+			if(length(GLOB.clients) > SSqueue.queue_threshold)
+				output_data["allow_player"] = FALSE
+			else // We have less than the threshold, allow if were in the timeframe
+				if(world.time > (SSqueue.last_letin_time + 3 SECONDS))
+					output_data["allow_player"] = TRUE
+					SSqueue.last_letin_time = world.time
+				else
+					output_data["allow_player"] = FALSE
```
AffectedArc07, line 30: Going to point out the one quirk with this logic.

If 2 new players want to connect, it only allows 1 new person in every 3 seconds, regardless of client count, due to how the logic works.

- If people > threshold, disallow
- If people < threshold, allow once every 3 seconds. This is to account for connection latency. If we drop below the threshold by 1, I dont want to instantly funnel 30 people in.

This prevents a mass wave from being let in at once and does spread the load out a bit.

Denthamos:
I like the que better than the panic bunker.  A soft line versus a hard door is better, however if an admin does decide to set it/change the threshold we should be alerted in the discord as to that happening.

AffectedArc07:
> I like the que better than the panic bunker. A soft line versus a hard door is better, however if an admin does decide to set it/change the threshold we should be alerted in the staff discord as to that happening.

Also easily doable. Do you want that implemented?

Denthamos:
> > I like the que better than the panic bunker. A soft line versus a hard door is better, however if an admin does decide to set it/change the threshold we should be alerted in the staff discord as to that happening.
> 
> Also easily doable. Do you want that implemented?

Yes please, the staff should be alerted on if an admin decides to change the threshold or if the que enables.  Visibility is good.  had a talk about it with Neca on it and we are in agreement there. @Necaladun

code/modules/admin/IsBanned.dm
```diff
@@ -97,10 +97,9 @@
 			qdel(exist_query)
 		else
 			if(!exist_query.NextRow()) // If there isnt a row, they aint been seen before
-				if(GLOB.panic_bunker_enabled)
+				if(SSqueue && SSqueue.queue_enabled && (length(GLOB.clients) > SSqueue.queue_threshold) && !(ckey in SSqueue.queue_bypass_list)) // To the person who tells me in code review "Oh you should use ?. here". No. That logic isnt appropriate here. It needs a dedicated null check due to the length comparison.
```
dearmochi, line 100: `?.` would work because if it failed, the rest would simply not be evaluated.

code/modules/admin/IsBanned.dm
```diff
@@ -97,10 +97,9 @@
...
 					qdel(exist_query)
-					var/threshold = GLOB.configuration.general.panic_bunker_threshold
-					return list("reason" = "panic bunker", "desc" = "Server is not accepting connections from never-before-seen players until player count is less than [threshold]. Please try again later.")
+					return list("reason" = "server queue", "desc" = "You seem to have managed to skip the server queue, possibly due to connecting during a restart. Please reconnect in 10 minutes. If you still cant connect, please inform the server host.")
```
dearmochi, line 102:
```suggestion
					return list("reason" = "server queue", "desc" = "You seem to have managed to skip the server queue, possibly due to connecting during a restart. Please reconnect in 10 minutes. If you still can't connect, please inform the server host.")
```

code/modules/world_topic/queue_status.dm
FlattestGuitar: Have you checked how long it takes from the time this response is sent to the client actually being added to `GLOB.clients`? If it's potentially way more than those 3 seconds (I don't know how this whole flow works) you could still be overflowing by a bunch of players. 

This is a simple solution that's probably good enough, but a safer way would be to just make a list of clients that have been allowed in but haven't connected yet and add its size to the check on line 23. Then on /mob/new (or something like that) you can just remove the client from the new list. It's tiny overhead and it lets you skip the 3 second wait. The list should probably be filtered for clients that have been connecting for a sensible amount of time, just in case someone is allowed entry but doesn't end up connecting for some reason.

code/modules/admin/IsBanned.dm
FlattestGuitar: I'm curious what affected meant by this

code/modules/world_topic/queue_status.dm
AffectedArc07: > Have you checked how long it takes from the time this response is sent to the client actually being added to GLOB.clients? If it's potentially way more than those 3 seconds (I don't know how this whole flow works) you could still be overflowing by a bunch of players.

Roughly three seconds for a BYOND client to reconnect in my own testing

> This is a simple solution that's probably good enough, but a safer way would be to just make a list of clients that have been allowed in but haven't connected yet and add its size to the check on line 23. Then on /client/new (or something like that) you can just remove the client from the new list. It's tiny overhead and it lets you skip the 3 second wait. The list should probably be filtered for clients that have been connecting for a sensible amount of time, just in case someone is allowed entry but doesn't end up connecting for some reason.

This is already handled by the bouncer. If you pass the queue but fail to connect, you dont have to re-queue, you just get put in, unless you miss your 10 **minute** window, in which case, thats your own fault.

> This solution would also make the queue feasible to run 24/7, because the 3 second wait will be really bad for higher player counts connecting at the same time.

The queue only applies to never seen before players, and bouncing out new joins is something I am ok with.

dearmochi: How does that work with map restarts and auto-reconnecting with a lot of players? Will there be enough time for everyone to join?

dearmochi (approved):
Besides my singular nit, LGTM

code/modules/world_topic/queue_status.dm
AffectedArc07: > How does that work with map restarts and auto-reconnecting with a lot of players? Will there be enough time for everyone to join?

- Existing players are not put into the queue
- Map restarts do not have a larger/smaller downtime penality, its the same

AffectedArc07:
image
image

@Denthamos

Denthamos:
Giving my approval with the changes provided by AA so staff has visibility of when changes are made and who made the changes.

variableundefined (approved):

Spacemanspark:
Thank god

PR #16546, opened 2021-08-14
Remaps DeltaStation Departures Security

LightFire53:
## What Does This PR Do

Moves the security holding are of departures of the NSS Kerberos to line up with the escape shuttle instead of being on the opposite end. This came at the cost of removing customs from departures to make a hallway/public area for people to get into departures still, as well as the removal of the security checkpoint way down at the bottom. ~~Because people are going to complain about the loss of security gear to steal due to that, there is a security locker in the security holding area.~~ As a result, I was able to add some additional seating where the previous security section had been. 

I also added a dedication plaque for extra seasoning.

## Why It's Good For The Game

Map parity, but also imagine being the security officers that have to move all the prisoners to the opposite end of the shuttle through the horde of greytide.

## Images of changes

**Old departures**
Old Departures

**New departures**
image

## Changelog
:cl:LightFire53
del: Removes DeltaStations's customs from departures
tweak: Re-maps DeltaStations's departures to line security up with the shuttle.

/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -38538,6 +38538,10 @@
 	icon_state = "dark"
 	},
 /area/turret_protected/ai)
+"bGr" = (
+/obj/effect/decal/warning_stripes/yellow,
+/turf/simulated/floor/plasteel,
+/area/security/checkpoint)
 "bGs" = (
 /obj/effect/spawner/window/reinforced,
 /obj/structure/sign/electricshock{
@@ -74608,7 +74612,10 @@
 /area/medical/medbay)
 "dcj" = (
 /obj/machinery/atmospherics/pipe/simple/hidden/scrubbers,
-/obj/machinery/atmospherics/pipe/simple/hidden/supply,
+/obj/machinery/atmospherics/pipe/manifold/hidden/supply{
+	dir = 4;
+	initialize_directions = 11
+	},
 /turf/simulated/floor/plasteel{
 	dir = 8;
 	icon_state = "neutralfull"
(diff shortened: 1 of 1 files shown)


AffectedArc07:
For sake of clarity in future PRs, **please** use technical names (Box, Meta, Delta) in PR titles to save headaches.

variableundefined (approved):

SteelSlayer:
There's only one thing I don't like right now, and that's the 1 tile wide room to hold the air canister for the escape airlock. I would suggest something more like this:

image

LightFire53:
> There's only one thing I don't like right now, and that's the 1 tile wide room to hold the air canister for the escape airlock. I would suggest something more like this:
> 
> image

I wasn't sure if that I put it like that, it wouldn't suck the cannister around. Although, a glass pane would fix that as well.

SteelSlayer:
> I wasn't sure if that I put it like that, it wouldn't suck the cannister around. Although, a glass pane would fix that as well.

It *shouldn't* blow around because wrenching those onto connectors anchors them. Though it you do decide to add glass, a windoor would be nice so people can still access it if they need to.

LightFire53:
Ignore green floor tile in the middle, that was just me executing poor screenshotting etiquette

S34NW (changes requested):
This has some good bits and bad bits.

The good:
A nice little chill area north of departures, sec getting more seats for crims awaiting transfer. Sec actually having access to the right shuttle door.

The bad:
Weird airlock configuration in the bottom right, the loss of a recharger and sec records console from the checkpoint, checkpoint moved to a more prominent and visible position.

I think of the "bad" issues i listed, we really do need to add back the recharger. Everything else I can live with, but this would be a stealth nerf to antags on delta.

Also, please add a disposals somewhere near the south of departures if you can.

LightFire53:
Escape flowers now look like this:
image

I also relabeled 2 security cameras in the departures hallway (one was brig equipment storage, the other mining external access) while I was adding a camera to the departure lounge area.
I have also added a wall recharger, however since it's only a holding area and not a full on checkpoint, adding a security records console feels out of place.

PR #16548, opened 2021-08-15
Fixes 'Build Mode' layering

### LINKED ISSUE #16544: Boo button/arrow button overlap build mode
**Issue Description**:
The boo button and arrow to hide buttons in the top left of the visual screen overlap on top of admin build mode preventing you from pressing them.

Build mode should overlap on top of boo and not the other way around.

**What happened instead**:
Cannot change build mode functions due to boo.

Harder to use admin build mode.

Have admin rank and be a ghost, enable build mode.  Cannot press the build mode function and ? button due to boo buttons.

When the boo and arrow button were added to ghosts.

**Extra information**:

SabreML:
## What Does This PR Do

Fixes the admin 'Build Mode' buttons showing below regular HUD buttons like spells and actions under some circumstances.
This was caused by the cooldown overlay on action buttons using the `plane` system, which takes rendering priority over the `layer`.
(Fixes #16544)

http://www.byond.com/docs/ref/#/atom/var/plane
https://www.byond.com/docs/ref/#/{notes}/renderer
https://github.com/ParadiseSS13/Paradise/blob/d03fc7ea1ceeb67ebdde2d3d885a0e091a267fcc/code/datums/action.dm#L542-L552

## Why It's Good For The Game

Admin tools shouldn't ever be blocked by regular ingame stuff.

## Images of changes

**Before:**
Paradise Test NSS Cyberiad 2021-08-15 015619

**After:**
Paradise Test NSS Cyberiad 2021-08-15 014211

## Changelog
:cl:
fix: Fixed the admin 'Build Mode' buttons sometimes rendering below normal HUD icons.
/:cl:

### DIFF
--- a/code/modules/buildmode/buttons.dm
+++ b/code/modules/buildmode/buttons.dm
@@ -1,7 +1,7 @@
 /obj/screen/buildmode
 	icon = 'icons/misc/buildmode.dmi'
 	var/datum/click_intercept/buildmode/bd
-	layer = HUD_LAYER_BUILDMODE
+	plane = HUD_PLANE_BUILDMODE
 
 /obj/screen/buildmode/New(bld)
 	bd = bld
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -118,5 +118,8 @@
 #define SPLASHSCREEN_LAYER 23
 #define SPLASHSCREEN_PLANE 23
 
+// This should always be on top.
+#define HUD_PLANE_BUILDMODE 30
+
 ///Plane master controller keys
 #define PLANE_MASTERS_GAME "plane_masters_game"
--- a/code/_onclick/hud/_defines.dm
+++ b/code/_onclick/hud/_defines.dm
@@ -180,4 +180,5 @@
 
 #define HUD_LAYER_SCREEN 20
 
-#define HUD_LAYER_BUILDMODE 30
+// Planes are above layers, and this should always be on top.
+#define HUD_PLANE_BUILDMODE 30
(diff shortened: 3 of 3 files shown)


dearmochi (approved):

code/modules/buildmode/buttons.dm
```diff
@@ -1,7 +1,7 @@
 /obj/screen/buildmode
 	icon = 'icons/misc/buildmode.dmi'
 	var/datum/click_intercept/buildmode/bd
-	layer = HUD_LAYER_BUILDMODE
+	plane = HUD_PLANE_BUILDMODE
```
AffectedArc07, line 4: Have you tested with just `plane = ABOVE_HUD_PLANE`

I would rather use that than yet another plane.

AffectedArc07 (changes requested):

code/modules/buildmode/buttons.dm
SabreML: `ABOVE_HUD_PLANE` works, but not flawlessly:
Paradise Test NSS Cyberiad 2021-08-15 142355
It's rendering above the darkening overlay since that's set to `FLOAT_PLANE + 1` (22), but the cooldown text is `FLOAT_PLANE +1.1` (22.1) so it's still taking priority. (`#define ABOVE_HUD_PLANE 22`)
This still fixes the main issue, but build mode being on its own plane fixes everything else too and is probably better for future-proofing.

AffectedArc07: Please put the buildmode plane in the file with the rest of the planes then

AffectedArc07 (approved):

PR #16560, opened 2021-08-17
[WIP] Multi instance support

AffectedArc07:
## Important

Please keep the feedback on this PR towards the implementation of this itself, not towards the context of a server split, that is a discussion for the forums. Anyone straying discussion from this implementation to discussion on whether we should split or not will be laughed at, have their comments removed, and told to use the forums.

## What Does This PR Do

This PR implements the framework for multiple instances of Paracode sharing the same database and configuration folder. Servers will be able to use the same configuration but also allow support for overrides between each server based on port, since thats unique per server. For example, both servers will load `config.toml`, but only the server listening on port 6666 will load `overrides_6666.toml`.

Todo list:
- [x] Config overrides
- [ ] Server ID tag in config
- [ ] Preventing players from logging into both servers at the same time to game unlocks or break data
- [ ] Saving said server ID tag in the database for things such as rounds (We can lookup the server ID for notes and bans based on the round ID, treating it as a foreign key)
- [ ] A peering subsystem so players can be presented with a server hop verb as well as the servers having a means to topic with eachother
- [ ] Cross-server asay (maybe even bridge it to discord)
- [ ] Cross-server notifications (Round has ended on server X, Server Y is now starting up. Map is Z Station)

## Why It's Good For The Game
Not having multiple instance support has been a bit of a drag on us for a while, and has prevented us from experimenting with split instances and other stuff of that nature, since the 

## Changelog
No changelog, purely backend stuff.

### DIFF
--- a/code/controllers/configuration/configuration_core.dm
+++ b/code/controllers/configuration/configuration_core.dm
@@ -44,12 +44,24 @@ GLOBAL_DATUM_INIT(configuration, /datum/server_configuration, new())
 	var/datum/configuration_section/url_configuration/url
 	/// Holder for the voting configuration datum
 	var/datum/configuration_section/vote_configuration/vote
+	/// Raw data. Stored here to avoid passing data between procs constantly
+	var/list/raw_data = list()
 
 
 /datum/server_configuration/Destroy(force)
 	SHOULD_CALL_PARENT(FALSE)
 	// This is going to stay existing. I dont care.
 	return QDEL_HINT_LETMELIVE
+
+/datum/server_configuration/vv_get_var(var_name)
+	if(var_name == "raw_data") // NO!
+		return FALSE
+	. = ..()
+
+/datum/server_configuration/vv_edit_var(var_name, var_value)
+	if(var_name == "raw_data") // NO!
+		return FALSE
+	. = ..()
 
 /datum/server_configuration/CanProcCall(procname)
 	return FALSE // No thanks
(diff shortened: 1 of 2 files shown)


Spacemanspark:
Yes plz

AffectedArc07:
image
Autodiscovery works.

AffectedArc07:
image

Not 100% sold on this message format, open to suggestions, but we now have notifications for when other instances start.

Spacemanspark:
Maybe enlarge the font by a tiny bit and make a sound effect play?

AffectedArc07:
image
Tweaked the message. It also plays the announcement ding now.

lectronyx:
So what does this mean for the average player? This looks really interesting, but I'm not really sure what exactly any of it really even means... eli5 please?

AffectedArc07:
> So what does this mean for the average player? This looks really interesting, but I'm not really sure what exactly any of it really even means... eli5 please?

For average player, fuck all

For server ops, ability to run 2 instances of paracode on the same DB at the same time

AffectedArc07:
image

This entire PR now uses an in-memory MariaDB table as a means of cross-server communications instead of BYOND topics. This table is fast to query and never touches the disk, so it doesn't burn up write cycles on the host NVMe drive.

This was easier than making a BYOND redis library, and we can query this all we want for on demand data and use it as a cache. 

Peers no longer need to be manually configured in the config, simply plug this into a database and it will autodetect peers based on heartbeat info.

More to come, but we can make progress on this again.

AffectedArc07:
image

You can no longer join 2 servers at once

AffectedArc07:
DB schema changes have been made, need to actually integrate the data saving and loading stuffs. 


My motivation sorta tanked recently (hence my total lack of activity elsewhere), but I am still gonna try get this out the door.

AffectedArc07:
Cross server asay is in
image

This PR is now ready for review

PR #16561, opened 2021-08-17
Compression only CPR

hal9000PR:
## What Does This PR Do
based on https://www.paradisestation.org/forum/topic/21293-allow-compression-only-cpr/

CPR can now be done with a mask on and such at 75% effectiveness.

## Why It's Good For The Game
more realism, chest compressions make up the majority of CPR. Makes the CPR feature less clunky to use and allows races like IPC/vox/plasmamen to actually do it.

## Changelog
:cl:
tweak: CPR can be done without a mouth or while wearing a mask at 75% effectiveness
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1593,6 +1593,9 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 		return
 	..()
 
+#define CPR_CHEST_COMPRESSION_ONLY	0.75
+#define CPR_RESCUE_BREATHS			1
+
 /mob/living/carbon/human/proc/do_cpr(mob/living/carbon/human/H)
 	if(H == src)
 		to_chat(src, "<span class='warning'>You cannot perform CPR on yourself!</span>")
@@ -1600,26 +1603,19 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 	if(H.stat == DEAD || HAS_TRAIT(H, TRAIT_FAKEDEATH))
 		to_chat(src, "<span class='warning'>[H.name] is dead!</span>")
 		return
-	if(!check_has_mouth())
-		to_chat(src, "<span class='danger'>You don't have a mouth, you cannot perform CPR!</span>")
-		return
-	if(!H.check_has_mouth())
-		to_chat(src, "<span class='danger'>They don't have a mouth, you cannot perform CPR!</span>")
-		return
-	if((head && (head.flags_cover & HEADCOVERSMOUTH)) || (wear_mask && (wear_mask.flags_cover & MASKCOVERSMOUTH) && !wear_mask.mask_adjusted))
-		to_chat(src, "<span class='warning'>Remove your mask first!</span>")
-		return
-	if((H.head && (H.head.flags_cover & HEADCOVERSMOUTH)) || (H.wear_mask && (H.wear_mask.flags_cover & MASKCOVERSMOUTH) && !H.wear_mask.mask_adjusted))
-		to_chat(src, "<span class='warning'>Remove [H.p_their()] mask first!</span>")
-		return
 	if(H.receiving_cpr) // To prevent spam stacking
 		to_chat(src, "<span class='warning'>They are already receiving CPR!</span>")
 		return
-	visible_message("<span class='danger'>[src] is trying to perform CPR on [H.name]!</span>", "<span class='danger'>You try to perform CPR on [H.name]!</span>")
 	H.receiving_cpr = TRUE
+	var/cpr_modifier = get_cpr_mod(H)
+	if(cpr_modifier == CPR_RESCUE_BREATHS)
+		visible_message("<span class='danger'>[src] is trying to perform chest compressions and rescue breaths on [H.name]!</span>", "<span class='danger'>You try to perform chest compressions and rescue breaths on [H.name]!</span>")
+	else
+		visible_message("<span class='danger'>[src] is trying to perform chest compressions on [H.name]!</span>", "<span class='danger'>You try to perform chest compressions on [H.name]!</span>")
+
 	if(do_mob(src, H, 40))
 		if(H.health <= HEALTH_THRESHOLD_CRIT)
-			H.adjustOxyLoss(-15)
+			H.adjustOxyLoss(-15 * cpr_modifier)
 			H.SetLoseBreath(0)
 			H.AdjustParalysis(-1)
 			H.updatehealth("cpr")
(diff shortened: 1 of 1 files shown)


Spacemanspark:
Make it go up to 100% if the person being CPR'd has broken ribs :^)

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1593,33 +1593,29 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
...
-		return
 	if(H.receiving_cpr) // To prevent spam stacking
 		to_chat(src, "<span class='warning'>They are already receiving CPR!</span>")
 		return
-	visible_message("<span class='danger'>[src] is trying to perform CPR on [H.name]!</span>", "<span class='danger'>You try to perform CPR on [H.name]!</span>")
 	H.receiving_cpr = TRUE
-	if(do_mob(src, H, 40))
+	var/cpr_modifier = get_cpr_mod(H)
+	if(cpr_modifier == CPR_RESCUE_BREATHS)
+		visible_message("<span class='danger'>[src] is trying to perform chest compressions and rescue breaths on [H.name]!</span>", "<span class='danger'>You try to perform chest compressions and rescue breaths on [H.name]!</span>")
```
SabreML, line 1612: `"John Smith is trying to perform chest compressions and rescue breaths on Injured Crewmember!"` seems like it might be a bit too long-winded for something that'll come up once every four seconds.
Would `"John Smith is trying to perform CPR on Injured Crewmember"` work instead? Personally I would take "CPR" to cover both of those, and "Chest Compressions" to just not include the breaths.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1593,33 +1593,29 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
...
+	else
+		visible_message("<span class='danger'>[src] is trying to perform chest compressions on [H.name]!</span>", "<span class='danger'>You try to perform chest compressions on [H.name]!</span>")
+
+	if(do_mob(src, H, 4 SECONDS))
 		if(H.health <= HEALTH_THRESHOLD_CRIT)
-			H.adjustOxyLoss(-15)
+			H.adjustOxyLoss(-15 * cpr_modifier)
 			H.SetLoseBreath(0)
 			H.AdjustParalysis(-1)
 			H.updatehealth("cpr")
```
SabreML, line 1621: Github won't let me comment on the line here, but the target still gets the `"You feel a breath of fresh air enter your lungs."` message even if it's just chest compressions.
https://github.com/ParadiseSS13/Paradise/blob/ea5f25f3f9757a138b311c45bb5d3867fd4542b8/code/modules/mob/living/carbon/human/human.dm#L1617-L1627

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1633,6 +1629,18 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 		H.receiving_cpr = FALSE
 		to_chat(src, "<span class='danger'>You need to stay still while performing CPR!</span>")
 
+/mob/living/carbon/human/proc/get_cpr_mod(mob/living/carbon/human/H)
+	if((head && (head.flags_cover & HEADCOVERSMOUTH)) || (wear_mask && (wear_mask.flags_cover & MASKCOVERSMOUTH) && !wear_mask.mask_adjusted))
+		return CPR_CHEST_COMPRESSION_ONLY
+	if((H.head && (H.head.flags_cover & HEADCOVERSMOUTH)) || (H.wear_mask && (H.wear_mask.flags_cover & MASKCOVERSMOUTH) && !H.wear_mask.mask_adjusted))
+		return CPR_CHEST_COMPRESSION_ONLY
+	if(!H.check_has_mouth() || !check_has_mouth())
+		return CPR_CHEST_COMPRESSION_ONLY
```
SabreML, line 1638: (Probably) unintended side effect of this is that IPCs can be given chest compressions now when they're in crit. It doesn't actually *do* anything since they don't take oxyloss damage, but it maybe shouldn't be possible to do in the first place.

SabreML (commented):
Just nitpicks really, LGTM otherwise.

code/modules/mob/living/carbon/human/human.dm
dearmochi, line 1638: This is a bit unpleasant to read, I would shuffle things around and use `?.` to make things a bit cleaner:
```suggestion
/mob/living/carbon/human/proc/get_cpr_mod(mob/living/carbon/human/H)
	if(!H.check_has_mouth() || !check_has_mouth())
		return CPR_CHEST_COMPRESSION_ONLY
	if(head?.flags_cover & HEADCOVERSMOUTH || (wear_mask && (wear_mask.flags_cover & MASKCOVERSMOUTH) && !wear_mask.mask_adjusted))
		return CPR_CHEST_COMPRESSION_ONLY
	if(H.head?.flags_cover & HEADCOVERSMOUTH || (H.wear_mask && (H.wear_mask.flags_cover & MASKCOVERSMOUTH) && !H.wear_mask.mask_adjusted))
		return CPR_CHEST_COMPRESSION_ONLY
```

One thing you could also do is add an argument to `check_has_mouth()` (or make a new proc) that returns whether the mob's mouth not only exists, but is also non-covered by something.

Denthamos:
Hmmn.  I have no issues with this PR but I was considering on if CPR for mask wearing would only work IF internals are enabled with a connected tank that has pressure.  That way you do not get CPR being performed on someone with an empty tank connected to their mask - meaning they are getting oxygen/nitrogen/plasma from nothing.

Wojz:
> Hmmn. I have no issues with this PR but I was considering on if CPR for mask wearing would only work IF internals are enabled with a connected tank that has pressure. That way you do not get CPR being performed on someone with an empty tank connected to their mask - meaning they are getting oxygen/nitrogen/plasma from nothing.

Fairly certain you can do CPR on someone without internals in space so checking for that with internals would be inconsistent.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1633,6 +1629,18 @@ Eyes need to have significantly high darksight to shine unless the mob has the X
 		H.receiving_cpr = FALSE
 		to_chat(src, "<span class='danger'>You need to stay still while performing CPR!</span>")
 
+/mob/living/carbon/human/proc/get_cpr_mod(mob/living/carbon/human/H)
+	if((head && (head.flags_cover & HEADCOVERSMOUTH)) || (wear_mask && (wear_mask.flags_cover & MASKCOVERSMOUTH) && !wear_mask.mask_adjusted))
+		return CPR_CHEST_COMPRESSION_ONLY
+	if((H.head && (H.head.flags_cover & HEADCOVERSMOUTH)) || (H.wear_mask && (H.wear_mask.flags_cover & MASKCOVERSMOUTH) && !H.wear_mask.mask_adjusted))
```
variableundefined, line 1635: At this point, I'd spin this into a helper function called "get_chest_compression" or something similar since it is used twice and contain multiple conditional

variableundefined (changes requested):
Design looks good to me. Code styling need some cleanup. 

See what Sabre said - I think a more compact message of "CPR" is better. 

Also, can you clean up the logic as Mochi requested (and double check if it still works)

code/modules/mob/living/carbon/human/human.dm
hal9000PR: fixed in the latest commit.

variableundefined (approved):

PR #15594, opened 2021-02-23
Port TG's vis_overlays subsystem

PidgeyThePirate:
## What Does This PR Do
Ports TG's vis_overlays subsystem. Also alters monitors to use the new system and adjusts their lighting slightly to suit.

**Port Links:**
Comfy computer lighting and a vis_contents overlay system foundation 
https://github.com/tgstation/tgstation/pull/39973

Better Glowing Lights
https://github.com/tgstation/tgstation/pull/49454

Profiling the relevant procs shows that the vis_overlay fire() proc appears to average 0.003 seconds per proc over 8 or so procs for 10 minutes solo, the rest stay 0.000. How it scales remains to be seen, but I don't think it'll be too bad.

## Why It's Good For The Game
It's pretty, and opens the door for some other cool visual effects, such as partially transparent items or reagents that glow in a beaker. I've just done the monitors for now though to minimize PR conflicts.

## Images of changes
image
emissiveblock

## Changelog
:cl:
add: Monitors now glow brightly in the dark. The light they emit has been dimmed slightly as a result.
/:cl:

PR #16333, opened 2021-07-09
Ports TG emergency lights, adds special code Epsilon and Delta effects.

S34NW:
## What Does This PR Do
This PR ports TG's emergency lighting system, ~~where each light has a small battery that can be printed from an autolathe,~~ in the event of power failure to a light, a red emergency light will activate for ~~the duration of the power cell's charge~~ 10 minutes.
~~Cells recharge automatically when a light is powered, or can be manually swapped out.~~ Emergency lights can be manually disabled via an unlocked APC. They cannot be manually activated outside of power loss.

This PR makes delta alert trigger all emergency fire lighting on-station for the duration of the alert.
This PR makes all lights turn off at the start of Epsilon, playing a power failure sound. They then turn back on in fire lighting mode and the usual siren and alert are displayed.

## Why It's Good For The Game
Delta and Epsilon alerts are supposed to grab your attention, this PR adds to that idea and enhances the atmosphere of the station in these events. You know when everything is red there is something MASSIVELY wrong.

In an ideal world, they would strobe or something, but that is not performant considering how many lights are on the cyberiad.

Also, having backup power makes sense for a scientific station, it also adds a degree of creepiness to the station when power is cut or disabled to lights in an area.

### Epsilon

https://user-images.githubusercontent.com/12197162/125302320-bd0c6a00-e323-11eb-8e2e-48bfdf068016.mp4

### Delta

https://user-images.githubusercontent.com/12197162/125302339-c09ff100-e323-11eb-8a4d-8e5cb623c042.mp4

## Changelog
:cl:
add: Adds emergency lighting to all lights that activate in event of power-loss.
tweak: Fire lights now use a modified version of emergency lighting.
tweak: Lights turned off via sling/vampire spells turn back on in emergency mode after 5 minutes.
tweak: Delta and epsilon alerts now activate emergency lighting station-wide.
/:cl:

PR #16567, opened 2021-08-19
Adds a short warm-up time when turning on a light

SabreML:
## What Does This PR Do

Adds a short semi-randomised delay after turning on a lightswitch before the actual lights themselves turn on.
Also adds sounds for light switches and lightbulbs flickering on.

This only happens when they're enabled manually from a light switch or APC, so it should act the same as it does now when using other methods.

*(The idea itself was taken from Skyrat, but the sounds and most of the code is original.)*

## Why It's Good For The Game

I don't think I've ever seen a light tube just turn on instantly in real life. More often than not they'll all do [this](https://youtu.be/73HkaOoBSjc) instead.
The changes in this PR aren't nearly as prominent as that example of course, but I still find that the effect it adds to the gameplay "immersion" is surprisingly substantial.
Primarily this just adds a nice little detail into the game. The sort of thing I can imagine someone being really impressed by when they first start playing.

When #15594 is merged this'll also be able to be made even fancier, with the light tube sprite flickering slightly before turning on fully.

## Images of changes

https://user-images.githubusercontent.com/57483089/129988373-355269a7-ea67-48c0-9885-dbbddf9d3121.mp4

## Changelog
:cl:
add: Added a slight delay after clicking a light switch for the room's lights to warm up before they turn on.
soundadd: Added a new sound for light switches.
soundadd: Added a new sound for light bulbs/tubes flickering on.
/:cl:

<hr>

- Light tube sound by mmaruska on [freesound.org](https://freesound.org/people/mmaruska/sounds/232447/), heavily modified.
- Light switch sound by AffectedArc07, also modified.

Both sounds can easily be tweaked further if necessary.


code/modules/power/lighting.dm
```diff
@@ -290,62 +295,86 @@
...
-			static_power_used = brightness_range * 20 //20W per unit luminosity
+			static_power_used = brightness_range * 20 //20W per unit of luminosity
 			addStaticPower(static_power_used, STATIC_LIGHT)
 		else
 			removeStaticPower(static_power_used, STATIC_LIGHT)
 
+/**
+  * The actual proc to turn on the lightbulb.
+  *
+  * Do not call directly, use [/obj/machinery/light/proc/update] instead.
```
AffectedArc07, line 332: You may be able to use one of the spacemandmm helpers here to ensure you cant invoke this like a public method
```dm
	SHOULD_BE_PURE(TRUE)
	PRIVATE_PROC(TRUE)
	PROTECTED_PROC(TRUE)
```

One of these will work, see which one trips the vscode DM builder up
image

You can click this to test without doing the entire DM rebuild

AffectedArc07 (changes requested):
Requesting you also put the metadata back into the ogg I send

AffectedArc07:
The ogg has the metadata, windows is just being an ass

code/modules/power/lighting.dm
SabreML: Oh nice, I was wondering if it had something like that. Couldn't find any documentation for the spacemandmm stuff though so I just left it.

dearmochi (approved):

TheYeetster1:
looks great I wonder how it will look with #16333

Sinestia:
That's very pretty, nice work

variableundefined (approved):

Kyep (approved):

PR #13115, opened 2020-03-13
Adds a new logging system and a logging view

farie82:
## What Does This PR Do
Adds a new system for logging. This one currently coexists next to the old ones.
A new logging view is added that allows multiple logs to be searched from multiple mobs (with ckeys) using several settings.

Also added better logging for the fueltanks

## Why It's Good For The Game
Gives the admins an easier to use tool to quickly see what happened in the logs between several players in a given timespan.

## Images of changes
image
Log Viewer

image
Scrolling works on the table. CTRL + F is possible

image
Colors used

FIXES:
```
[2020-02-01T01:49:18] Runtime in mobs.dm,299: Cannot read null.ckey
[2020-02-01T01:49:18]   proc name: add attack logs (/proc/add_attack_logs)
[2020-02-01T01:49:18]   usr: Vin'skerasq (CKEY) (/mob/living/carbon/human)
[2020-02-01T01:49:18]   usr.loc: The plating (164,153,1) (/turf/simulated/floor/plating)
[2020-02-01T01:49:18]   src: null
[2020-02-01T01:49:18]   call stack:
[2020-02-01T01:49:18]   add attack logs(null, Cordell Walker (/mob/living/carbon/human), "Flashed with the flash", null)
[2020-02-01T01:49:18]   the flash (/obj/item/flash/cyborg): flash carbon(Cordell Walker (/mob/living/carbon/human), null, 10, 0)
[2020-02-01T01:49:18]   the flash (/obj/item/flash/cyborg): emp act(1)
[2020-02-01T01:49:18]   the engineering robot module (/obj/item/robot_module/engineering): emp act(1)
[2020-02-01T01:49:18]   MIRA (/mob/living/silicon/robot): emp act(1)
[2020-02-01T01:49:18]   MIRA (/mob/living/silicon/robot): emp act(1)
[2020-02-01T01:49:18]   MIRA (/mob/living/silicon/robot): emp act(1)
[2020-02-01T01:49:18]   the flashlight (/obj/item/flashlight/emp): afterattack(MIRA (/mob/living/silicon/robot), Vin\'skerasq (/mob/living/carbon/human), 1, "icon-x=17;icon-y=19;left=1;scr...")
[2020-02-01T01:49:18]   the flashlight (/obj/item/flashlight/emp): melee attack chain(Vin\'skerasq (/mob/living/carbon/human), MIRA (/mob/living/silicon/robot), "icon-x=17;icon-y=19;left=1;scr...")
[...]

PR #16576, opened 2021-08-20
Deletes attack_log_old and all related vars/procs

farie82:
## What Does This PR Do
Deletes the old attack log code which got replaced by the new system from #13115.
I've double-checked to see if all old logs were replaced by the new system calls and they were.

## Why It's Good For The Game
It seems no admins use it anymore, nor should they. Removing the code for this makes attack logging a bit more logical (no dupe log calls which looked weird) and will free up some memory per mob.

## Changelog
:cl:
del: Removes the old attack log code
/:cl:

### DIFF
--- a/code/modules/mob/mob_defines.dm
+++ b/code/modules/mob/mob_defines.dm
@@ -33,12 +33,10 @@
 	var/computer_id = null
 	var/lastattacker = null // real name of the person  doing the attacking
 	var/lastattackerckey = null // their ckey
-	var/list/attack_log_old = list( )
 	var/list/debug_log = null
 
 	var/last_known_ckey = null	// Used in logging
 
-	var/last_log = 0
 	var/obj/machinery/machine = null
 	var/other_mobs = null
 	var/memory = ""
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -302,10 +302,8 @@ This is always put in the attack log.
 	var/mob/MT = target
 	if(istype(MU))
 		MU.create_log(ATTACK_LOG, what_done, target, get_turf(user))
-		MU.create_attack_log("<font color='red'>Attacked [target_str]: [what_done]</font>")
 	if(istype(MT))
 		MT.create_log(DEFENSE_LOG, what_done, user, get_turf(MT))
-		MT.create_attack_log("<font color='orange'>Attacked by [user_str]: [what_done]</font>")
 	log_attack(user_str, target_str, what_done)
 
 	var/loglevel = ATKLOG_MOST
--- a/code/datums/datumvars.dm
+++ b/code/datums/datumvars.dm
@@ -40,7 +40,7 @@
 
 /datum/proc/vv_get_var(var_name)
 	switch(var_name)
-		if("attack_log_old", "debug_log")
+		if("debug_log")
 			return debug_variable(var_name, vars[var_name], 0, src, sanitize = FALSE)
 		if("vars")
 			return debug_variable(var_name, list(), 0, src)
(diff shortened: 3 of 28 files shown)


code/modules/mob/mob_defines.dm
```diff
@@ -33,12 +33,10 @@
 	var/computer_id = null
 	var/lastattacker = null // real name of the person  doing the attacking
 	var/lastattackerckey = null // their ckey
-	var/list/attack_log_old = list( )
 	var/list/debug_log = null
 
 	var/last_known_ckey = null	// Used in logging
 
-	var/last_log = 0
```
farie82, line 41: Was only used by the old attack logs weirdly enough

AffectedArc07 (changes requested):
Dropped a ping for this in adminchat and 3 people instantly responded saying they use this. 

We're not deleting it.

Spacemanspark:
>It seems no admins use it anymore, nor should they.

I use this. It's actually faster than opening up the log panel in game and having to add a bunch of filters--especially while we have upwards of 150+ players at peak hours.
Please do not delete it.

AffectedArc07:
Also as a general rule of thumb, don't go deleting admin features when you aren't aware of what we do or don't even use.

Qwertytoforty:
To be fair on henk, I was not aware anyone used this still

farie82:
> Also as a general rule of thumb, don't go deleting admin features when you aren't aware of what we do or don't even use.

I did ask a few admins and never heard any other admin using it thus far. A ping seemed a bit overkill knowing this all.

farie82:
> I use this. It's actually faster than opening up the log panel in game and having to add a bunch of filters--especially while we have upwards of 150+ players at peak hours.
> Please do not delete it. As well, if you haven't adminned like many of us, please don't try and say what we should and shouldn't do.

How exactly do you use this? Do you use the proc? Do you use the VV panel? If it is the proc, I can change it to load up the viewer with just the selected mob and the attack logs or another set of settings.

Having two logging systems in place doing the same thing is a bit stupid, so I'd rather get rid of the old one and improve the new one where needed. I'm open for improvements on the logging view as I've heard no issues with it so far other than the ones you've listed

MarsM0nd:
> How exactly do you use this? Do you use the proc? Do you use the VV panel? If it is the proc, I can change it to load up the viewer with just the selected mob and the attack logs or another set of settings.
I think it's using VV, especially while being a ghost, ctrl+shift+click opens it up, and can be quickly checked with ctrl+f, rather than opening up a persons player panel with ctrl+click, clicking the logs option in that window, and then removing the chat stuff before hitting search in another new window. VV also only takes up a part of your screen, while the log viewer does take most of it, so it's a bit harder to multi-task (watching chat/the person that's getting checked) with the log viewer in use.
Can't say for certain if it'll cover all usages, but a quick to open slim log viewer could be useful. Give one for chat, one for attacks, to cover the VV attack_log and say_log uses, and it only contains what happened. No link to anything, just the log lines who they attacked and who attacked them.

farie82:
> I think it's using VV, especially while being a ghost, ctrl+shift+click opens it up, and can be quickly checked with ctrl+f, rather than opening up a persons player panel with ctrl+click, clicking the logs option in that window, and then removing the chat stuff before hitting search in another new window. VV also only takes up a part of your screen, while the log viewer does take most of it, so it's a bit harder to multi-task (watching chat/the person that's getting checked) with the log viewer in use.
> Can't say for certain if it'll cover all usages, but a quick to open slim log viewer could be useful. Give one for chat, one for attacks, to cover the VV attack_log and say_log uses, and it only contains what happened. No link to anything, just the log lines who they attacked and who attacked them.

Thanks for the response. The use cases given do make sense indeed. I'll think about ways of making the new way work better in these cases. For now, I'll close the PR and make a new one when I finalised those cases.

PR #16577, opened 2021-08-20
AI door bolt logging

SabreML:
## What Does This PR Do

Adds a 'Misc' log and hidden fingerprints to an airlock when it's bolted remotely by a cyborg or AI, since for some reason there wasn't any before.
(Requested by Qwerty)

## Why It's Good For The Game

Logs for the log god. 🌲

## Changelog
:cl:
tweak: Added better admin logging to remotely bolting a door.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -851,10 +851,15 @@ About the new airlock wires panel:
 /obj/machinery/door/airlock/proc/toggle_bolt(mob/user)
 	if(wires.is_cut(WIRE_DOOR_BOLTS))
 		to_chat(user, "<span class='warning'>The door bolt control wire has been cut - Door bolts permanently dropped.</span>")
-	else if(lock())
-		to_chat(user, "<span class='notice'>The door bolts have been dropped.</span>")
-	else if(unlock())
+		return
+
+	if(unlock()) // Trying to unbolt
 		to_chat(user, "<span class='notice'>The door bolts have been raised.</span>")
+
+	else if(lock()) // Trying to bolt
+		to_chat(user, "<span class='notice'>The door bolts have been dropped.</span>")
+		user.create_log(MISC_LOG, "Bolted", src)
+		add_hiddenprint(user)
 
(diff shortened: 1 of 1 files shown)


dearmochi (approved):

code/game/machinery/doors/airlock.dm
```diff
@@ -851,11 +851,16 @@ About the new airlock wires panel:
 /obj/machinery/door/airlock/proc/toggle_bolt(mob/user)
 	if(wires.is_cut(WIRE_DOOR_BOLTS))
 		to_chat(user, "<span class='warning'>The door bolt control wire has been cut - Door bolts permanently dropped.</span>")
-	else if(lock())
-		to_chat(user, "<span class='notice'>The door bolts have been dropped.</span>")
-	else if(unlock())
+		return
```
AffectedArc07, line 854: If youre turning one into an early return, either do all of them or none of them

SabreML: Left it that way since `if(unlock) else if(lock)` made sense to me, but may as well change it for consistency, sure.

code/game/machinery/doors/airlock.dm
```diff
@@ -851,10 +851,16 @@ About the new airlock wires panel:
...
+
+	if(unlock()) // Trying to unbolt
 		to_chat(user, "<span class='notice'>The door bolts have been raised.</span>")
+		return
+
+	if(lock()) // Trying to bolt
+		to_chat(user, "<span class='notice'>The door bolts have been dropped.</span>")
+		user.create_log(MISC_LOG, "Bolted", src)
+		add_hiddenprint(user)
```
AffectedArc07, line 863:
```suggestion
		add_hiddenprint(user)
		return
		
	to_chat(user, "<span class='warning'>No action was performed. This may be a bug.</span>")
```

SabreML: It seems like trying to unbolt the door while the power is disabled will also trigger this, so it might be best to just leave it out.

PR #16581, opened 2021-08-21
Fix for Security camera console when reassembled(second try)

### LINKED ISSUE #14625: Not roundstart Security camera console don't list any cameras
**Issue Description**:

Newly built Security camera console don't see any cameras.

See the cameras.

**What happened instead**:

Didn't see any camera listed

Renders the console useless.

Build a security camera console OR just screwdrive one off and on.

Can't tell, I would say it come with the new GUI.

**Extra information**:

Keep this information away from antags.  <.<

### LINKED ISSUE #16568: The Security camera consoles network issue
**Issue Description**:
The security camera console, when rebuild, wont show any cameras. This is probably due to that every single security camera console on the map have had their network list value edited to show these networks.
Skärmbild 2021-08-19 092143

 And so when rebuilding a console it use the default dm that have the network list value empty.
Skärmbild 2021-08-19 092757

Adding "SS13","Research Outpost","Mining Outpost" to the list the console have connection to the networks when they're rebuild.

To be able to move the consoles.

**What happened instead**:
The console will be useless after reconstruction.

**Extra information**:

KrossarnTom:
## What Does This PR Do

When reassembling a Security cameraa console it will lose the connection to all networks, thus showing no cameras.
Each and every security camera console are edited on the map to show these networks while the default DM is blank.

By adding the networks directly to the dm we wont have this issue.

## Why It's Good For The Game

Fixes #14625
Fixes #16568

Being able to reassemble the consoles, to renovate or transport, without them losing their usefulness.

## Changelog
:cl:

fix: Fixed the security camera console losing connection to the camera networks
/:cl:

### DIFF
--- a/code/game/machinery/computer/camera.dm
+++ b/code/game/machinery/computer/camera.dm
@@ -9,7 +9,7 @@
 
 	var/mapping = 0 // For the overview file (overview.dm), not used on this page
 
-	var/list/network = list()
+	var/list/network = list("SS13","Mining Outpost")
 	var/obj/machinery/camera/active_camera
 	var/list/watchers = list()
 


code/game/machinery/computer/camera.dm
```diff
@@ -9,7 +9,7 @@
 
 	var/mapping = 0 // For the overview file (overview.dm), not used on this page
 
-	var/list/network = list()
+	var/list/network = list("SS13","Mining Outpost")
```
dearmochi, line 12: Unsure if this is the best way to go about it, I would say to keep the network list in the circuit board itself, but there's no easy way to do it...

AffectedArc07 (commented):
Apparently a while ago, when someone assembled a console, it would add networks based on the persons access.

I would check in the console `Initialize()` and see if

A) If `usr` is present
B) If `usr` is a human
C) What access `usr` has on their ID

code/game/machinery/computer/camera.dm
KrossarnTom: Although that sounds like it would be something. What i understand from the different circuit boards is that they act like a "Build path" for the Computer frame. 
Skärmbild 2021-08-21 220457

Add a security camera console(camera monitor) board and the computer frame is directed to that dm. Add a supermatter monitoring console board and the computer frame is directed to that dm. 

So the only way to change anything is the dm that is the final product. After looking through the code this is the most logical solution i could come up with.

But i wont be suprised if im dead wrong.

KrossarnTom:
> Apparently a while ago, when someone assembled a console, it would add networks based on the persons access.
> 
> I would check in the console `Initialize()` and see if
> 
> A) If `usr` is present
> B) `If `usr`is a human C) What access`usr` has on their ID

If we look at the same dm and scroll further down to the other types of camera consoles we can see that they are connected to the network by default. Like the detectives wooden tv. Even if you reassemble the tv, it will be connected to the "SS13" network by default in the dm. Engineering camera console finds areas with air alarms, fire alarms and power alarms. They are always connected to those networks. even if you reassemble them.

KrossarnTom:
Seems the access requirement for the camera consoles had been removed when TGUI were implemented, here 
#13750.

While scouting the other codebases i can see that their dm is almost identical to this one. With the only difference being that they have "SS13" added to the network list.

S34NW:
Please include
Fixes https://github.com/ParadiseSS13/Paradise/issues/14625
and
Fixes https://github.com/ParadiseSS13/Paradise/issues/16568

somewhere in your PR to autoclose these issues.

variableundefined (approved):
Make sense to me

PR #16583, opened 2021-08-21
Fixes flying/jaunting/floating mobs breaking glass tables.

S34NW:
## What Does This PR Do
This PR stops mobs with the `flying`, `floating`, or `incorporeal_move` vars from breaking glass tables when they cross them.
This includes things like scout holoparasites, jaunting slings, and anything else that doesn't actually sit on the floor.

## Why It's Good For The Game
Something that can phase through walls still breaks a glass table, this not only can unintentionally hurt the user, but it can also give them away to an unknowing crew.

## Changelog
:cl:
fix: Fixed glass tables being broken by flying, floating or jaunting mobs.
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -406,11 +406,12 @@
 	. = ..()
 	if(flags & NODECONSTRUCT)
 		return
-	if(isliving(AM))
-		var/mob/living/M = AM
-		if(M.incorporeal_move || M.flying || M.floating)
-			return
-	else
+	if(!isliving(AM))
+		return
+	var/mob/living/L = AM
+	if(L.incorporeal_move || L.flying || L.floating)
+		return
+
 		return
 	// Don't break if they're just flying past
 	if(AM.throwing)
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -406,7 +406,11 @@
 	. = ..()
 	if(flags & NODECONSTRUCT)
 		return
-	if(!isliving(AM))
+	if(isliving(AM))
+		var/mob/living/M = AM
+		if(M.incorporeal_move || M.flying || M.floating)
+			return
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/tables_racks.dm
```diff
@@ -406,7 +406,11 @@
 	. = ..()
 	if(flags & NODECONSTRUCT)
 		return
-	if(!isliving(AM))
+	if(isliving(AM))
+		var/mob/living/M = AM
+		if(M.incorporeal_move || M.flying || M.floating)
+			return
+	else
```
SabreML, line 413: You can keep the early return here since it only accepts `/mob/living` anyway.
```suggestion
	if(!isliving(AM))
		return
	var/mob/living/L = AM
	if(L.incorporeal_move || L.flying || L.floating)
		return

```

SabreML (commented):
Honestly amazed nobody's got around to fixing this before considering how annoying it can be. Thank you!

code/game/objects/structures/tables_racks.dm
```diff
@@ -407,6 +407,11 @@
 	if(flags & NODECONSTRUCT)
 		return
 	if(!isliving(AM))
+		return
+	var/mob/living/L = AM
+	if(L.incorporeal_move || L.flying || L.floating)
+		return
+
 		return
```
S34NW, line 415:
```suggestion
```

S34NW: Don't blindly accept review changes, folks!

SabreML (approved):

ppi13:
Honestly, this is a much loved feature, that should stay

AffectedArc07:
> Honestly, this is a much loved feature, that should stay

Ah yes a feature which has not only been a bug for so long, but allows blatant metagames for what way people are warping around when its meant to be a stealth move.

AffectedArc07 (approved):

PR #16586, opened 2021-08-22
[TM NEEDED] Possibly fix being unable to call contractor extraction

dearmochi:
## What Does This PR Do

- Possibly fixes being unable to call contractor extraction in some cases, extended TM required.

## Why It's Good For The Game

Bugfix, but not yet nailed down

## Changelog
:cl:
fix: Possibly fix contractor extraction not being callable
/:cl:


SabreML:
Would it be possible to also add some stack traces around there in order to narrow down what the problem is?

AffectedArc07 (approved):
Seemingly fixed

PR #16589, opened 2021-08-23
New Candle Sprites

PacifistDalek:
## What Does This PR Do
New candle sprites from [Sojourn Station](https://github.com/sojourn-13/sojourn-station/pull/1667). They're animated with dripping wax, and look gorgeous imho.

## Images of changes
candle

image

## Changelog
:cl:
imageadd: New candle sprites
/:cl:


matttheficus:
I think theyre a tad wide. They'd be nice if they were just a BIT thinner.

Fox-McCloud:
Sojourn has a completely different style and aesthetic than the one we're attempting to standardize on. This is a no go.

PacifistDalek:
> I think theyre a tad wide. They'd be nice if they were just a BIT thinner.

Fair enough! Personally i've always found ours weirdly thin, like birthday cake candles or candles from 50 years ago. These match the width of the candles I have and the church I attend uses, hence why I like them, but it is a pretty dramatic change, so preferring thinner ones makes perfect sense! 

> Sojourn has a completely different style and aesthetic than the one we're attempting to standardize on. This is a no go.

In general, yes, but these match the perspective and style that was called for in the respriting design doc. (heck Sojourn is also porting some of TG's sprites right now, so there is some overlap in style. )

S34NW:
While I think we can improve on our current candles, these look a little fat IMO

PacifistDalek:
General consensus seems to be that these are a no go! Might take a crack at a different version at some point down the line, but for now i'll just close it to keep from cluttering up the PR list.

PR #16591, opened 2021-08-23
Removes wall jaunting as we know it

S34NW:
## What Does This PR Do
This PR expands the range of valid turfs scanned when jaunting into a wall from 1 to 7.

## Why It's Good For The Game
Currently, when jaunting into a wall and/or dense turf, the game "scans" for a valid turf 1 tile away, if it cannot find this it just manifests you in the wall anyway.
This leads to AI kill cheesing by vampires, wizards and wraiths. As they are in a wall completely safe from retaliation.
With this PR, the range of "valid" turfs has been expanded to 7 from center, so the user will get popped out somewhere within that range if there is a valid turf.
If a user somehow manages to land in a space where there are no valid turfs in 7 tiles (see: deep in lavaland rock) then just manifest within the wall as usual.

## Changelog
:cl:
tweak: Tweaked jaunt to not put you inside a wall if there is a valid spot to manifest within 7 tiles.
/:cl:

### DIFF
--- a/code/game/gamemodes/vampire/vampire_powers.dm
+++ b/code/game/gamemodes/vampire/vampire_powers.dm
@@ -457,8 +457,7 @@
 		flick("mist_reappear", animation)
 		sleep(5)
 		if(!user.Move(mobloc))
-			for(var/direction in list(1,2,4,8,5,6,9,10))
-				var/turf/T = get_step(mobloc, direction)
+			for(var/turf/T in orange(7, mobloc))
 				if(T)
 					if(user.Move(T))
 						break
--- a/code/datums/spells/ethereal_jaunt.dm
+++ b/code/datums/spells/ethereal_jaunt.dm
@@ -56,8 +56,7 @@
 	qdel(holder)
 	if(!QDELETED(target))
 		if(mobloc.density)
-			for(var/direction in GLOB.alldirs)
-				var/turf/T = get_step(mobloc, direction)
+			for(var/turf/T in orange(7))
 				if(T)
 					if(target.Move(T))
 						break


Wojz:
To any wraith gamers who will surely come in here to say that this removes their only purpose: You can go anywhere in the station and break infrastructure. Go do that.

I like this change. It definitely removes a lot of hard to justify IC cheese. And no more funny jaunt behind AI and card vampires, either. Maybe the AI sat will be at least slightly defensible now.

Spacemanspark:
Christ, about time.

code/game/gamemodes/vampire/vampire_powers.dm
```diff
@@ -457,8 +457,7 @@
 		flick("mist_reappear", animation)
 		sleep(5)
 		if(!user.Move(mobloc))
-			for(var/direction in list(1,2,4,8,5,6,9,10))
```
AffectedArc07, line 460: What the fuck

hal9000PR: spellcode

Kyep: Truly magical.

Dafrek:
> To any wraith gamers who will surely come in here to say that this removes their only purpose: You can go anywhere in the station and break infrastructure. Go do that.
> 
> I like this change. It definitely removes a lot of hard to justify IC cheese. And no more funny jaunt behind AI and card vampires, either. Maybe the AI sat will be at least slightly defensible now.

Is the tile the AI on considered a full tile that wraiths cannot end their jaunt on? If its not then we still have the same problem I believe.

hal9000PR:
> > To any wraith gamers who will surely come in here to say that this removes their only purpose: You can go anywhere in the station and break infrastructure. Go do that.
> > I like this change. It definitely removes a lot of hard to justify IC cheese. And no more funny jaunt behind AI and card vampires, either. Maybe the AI sat will be at least slightly defensible now.
> 
> Is the tile the AI on considered a full tile that wraiths cannot end their jaunt on? If its not then we still have the same problem I believe.

If you can jaunt onto a tile it is considered valid for jaunt. You cannot walk into an AI core so it will push you off, likely to the top left corner, where you will get shot by turrets.

S34NW:
> 
> 
> > To any wraith gamers who will surely come in here to say that this removes their only purpose: You can go anywhere in the station and break infrastructure. Go do that.
> > I like this change. It definitely removes a lot of hard to justify IC cheese. And no more funny jaunt behind AI and card vampires, either. Maybe the AI sat will be at least slightly defensible now.
> 
> Is the tile the AI on considered a full tile that wraiths cannot end their jaunt on? If its not then we still have the same problem I believe.

Any tile with a dense object should cause the same result, so yes the AI will block jaunting onto itself.

Butlerblock:
As wraith you can break the APC before the turrets even start firing. As anything else you can get insuls and quickly screw driver and cut all the wires on the APC. Pretty much anyone can do it fast enough.

S34NW:
> 
> 
> As wraith you can break the APC before the turrets even start firing. As anything else you can get insuls and quickly screw driver and cut all the wires on the APC. Pretty much anyone can do it fast enough.

Well, that might call for another set of changes, but this PR is here to specifically tackle the fact that players can sit in a wall and be totally untouchable at all while having free shots on the AI and SMES.

PR #16592, opened 2021-08-23
Fixes runtimes in spell can_cast() checks.

S34NW:
## What Does This PR Do
This PR changes how centcom_cancast checks are made for spells, to cut back on the number of runtimes caused by people in disposals having spells that check for turf z level.

## Why It's Good For The Game
Fewer runtimes to filter through is good. There is no CL due to this being a backend change.


farie82 (approved):
Minor styling thing. Looks good otherwise

AffectedArc07 (approved):

PR #16596, opened 2021-08-24
Makes admin verbs easier to see on the context menu

S34NW:
## What Does This PR Do
This PR makes it easier to see at a glance what are admin verbs vs what are not in the right-click context menu.

## Why It's Good For The Game
You would not believe the amount of times admins misclick and heal/freeze/almost delete players. It's been one of my biggest gripes since becoming a TA. You can't even properly learn it in muscle memory because the window position shifts depending on what you are clicking.
This makes it significantly easier to see you are pushing a big uh-oh admin verb, thus making it harder to accidentally click on one.

The one downside to this is that a few verbs are also present in the event verb tab, meaning these now get shunted to the bottom of that list. Note that nothing has been deleted here.

In a ping to all online admins, 13 supported this change and 0 opposed it. While not _every_ active admin, this is a pretty strong indicator that this change is reasonably supported among those using these tools.

### Current verb list: (As a TA, I don't have _all_ the verbs a GA has, hence the discrepancy.)

2021-08-24 22_25_59-Greenshot
2021-08-24 22_26_26-Greenshot

### Proposed change:

2021-08-24 22_28_29-Greenshot
2021-08-24 22_28_36-Greenshot

## Changelog
:cl:
tweak: Tweaked how admin verbs are displayed in the right-click context menu
/:cl:

### DIFF
--- a/code/datums/datumvars.dm
+++ b/code/datums/datumvars.dm
@@ -78,7 +78,7 @@
 
 /client/proc/debug_variables(datum/D in world)
 	set category = "Debug"
-	set name = "View Variables"
+	set name = "\[Admin\] View Variables"
 
 	var/static/cookieoffset = rand(1, 9999) //to force cookies to reset after the round.
 
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -72,7 +72,7 @@ GLOBAL_VAR_INIT(nologevent, 0)
 
 /datum/admins/proc/show_player_panel(mob/M in GLOB.mob_list)
 	set category = null
-	set name = "Show Player Panel"
+	set name = "\[Admin\] Show Player Panel"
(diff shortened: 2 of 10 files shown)


code/datums/datumvars.dm
```diff
@@ -78,7 +78,7 @@
 
 /client/proc/debug_variables(datum/D in world)
 	set category = "Debug"
-	set name = "View Variables"
+	set name = "\[Admin\] View Variables"
```
AffectedArc07, line 81: This offsetting things in the verbs panel is kinda ehh

S34NW: Yeah, I can't really think of a way around this because of how these bloody verbs work. I didn't want to remove them from the verb tab since some admins might use them there for whatever reason.

variableundefined (approved):

variableundefined:
how many muscle memory complaints we'll get

PR #16608, opened 2021-08-28
Tweak to revolution conversions

S34NW:
## What Does This PR Do
This is a small PR that adds a 20-second cooldown to flash conversions by head revs.

## Why It's Good For The Game
While it's rarely forced and is not in rotation, when rev _is_ played the lack of any cooldown on conversions just leads to flash spam and an overwhelming rev force within minutes. This PR means it takes a little while to build up rev forces, and admins can tweak the cooldown var per head rev on demand.

## Changelog
:cl:
tweak: Tweaked head revs so they can only convert every 20 seconds.
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -45,7 +45,7 @@
 	var/list/targets = list()
 
 	///Tracks if this mind has been a rev or not
-	var/has_been_rev = 0
+	var/has_been_rev = FALSE
 	///Tracks if headrevs are on a conversion cooldown or not
 	var/conversion_on_cooldown = FALSE
 	///How long the cooldown timer is for conversion
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -44,12 +44,7 @@
 	var/list/datum/objective/special_verbs = list()
 	var/list/targets = list()
 
-	///Tracks if this mind has been a rev or not
-	var/has_been_rev = 0
-	///Tracks if headrevs are on a conversion cooldown or not
-	var/conversion_on_cooldown = FALSE
-	///How long the cooldown timer is for conversion
-	var/conversion_cooldown_time = 20 SECONDS
+	var/has_been_rev = 0//Tracks if this mind has been a rev or not
 
 	var/miming = 0 // Mime's vow of silence
 	var/list/antag_datums
@@ -1509,13 +1504,6 @@
 	var/obj/item/uplink/hidden/H = find_syndicate_uplink()
 	if(H)
 		qdel(H)
-
-/datum/mind/proc/begin_conversion_cooldown()
-	conversion_on_cooldown = TRUE
-	addtimer(CALLBACK(src, .proc/end_conversion_cooldown), conversion_cooldown_time, TIMER_OVERRIDE)
-
-/datum/mind/proc/end_conversion_cooldown()
-	conversion_on_cooldown = FALSE
 
 /datum/mind/proc/make_Traitor()
 	if(!has_antag_datum(/datum/antagonist/traitor))
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -45,7 +45,7 @@
 	var/list/targets = list()
 
 	///Tracks if this mind has been a rev or not
-	var/has_been_rev = FALSE
+	var/has_been_rev = 0
 	///Tracks if headrevs are on a conversion cooldown or not
 	var/conversion_on_cooldown = FALSE
 	///How long the cooldown timer is for conversion
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -44,7 +44,12 @@
 	var/list/datum/objective/special_verbs = list()
 	var/list/targets = list()
 
-	var/has_been_rev = 0//Tracks if this mind has been a rev or not
+	///Tracks if this mind has been a rev or not
+	var/has_been_rev = 0
+	///Tracks if headrevs are on a conversion cooldown or not
+	var/conversion_on_cooldown = FALSE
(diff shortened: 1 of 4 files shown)


code/datums/mind.dm
```diff
@@ -44,7 +44,12 @@
 	var/list/datum/objective/special_verbs = list()
 	var/list/targets = list()
 
-	var/has_been_rev = 0//Tracks if this mind has been a rev or not
+	///Tracks if this mind has been a rev or not
+	var/has_been_rev = 0
```
S34NW, line 48:
```suggestion
	var/has_been_rev = FALSE
```
I suppose this can be tidied

Spacemanspark:
The problem here is that it doesn't address the core issue of revs, or the security side of the problem.

S34NW:
> 
> 
> The problem here is that it doesn't address the core issue of revs, or the security side of the problem.

Indeed, that will require extensive work. Little steps.

Spacemanspark:
Yes and no.

The only thing this PR does at present is make more issues in game and alter what's already a very poor game mode into an even worse state. I'd also question if we should even be modifying the cooldown of the conversion when the conversion itself is a poor mechanic.

At the very least, I think this should be packaged in with a few other changes. I suppose it's not the worst thing in the world, seeing as the mode is out of rotation, but when it _is_ forced, this may just make things worse.

code/datums/mind.dm
```diff
@@ -44,7 +44,12 @@
 	var/list/datum/objective/special_verbs = list()
 	var/list/targets = list()
 
-	var/has_been_rev = 0//Tracks if this mind has been a rev or not
+	///Tracks if this mind has been a rev or not
+	var/has_been_rev = FALSE
+	///Tracks if headrevs are on a conversion cooldown or not
+	var/conversion_on_cooldown = FALSE
+	///How long the cooldown timer is for conversion
+	var/conversion_cooldown_time = 20 SECONDS
```
SteelSlayer, line 52: I really do not like how these are tacked onto the mind.

hal9000PR:
> when it is forced, this may just make things worse.

Would you please explain this bit a little more? I'm not quite sure I understand how/what you mean by this.

code/datums/mind.dm
S34NW: Indeed, I'm not sure where else to put it, unfortunately. It can't go on the flash since people will just carry a bag of them, and there isn't a rev datum that is assigned to the mind.

hal9000PR: You could make it a status effect. That way people know when they are on cooldown aswell.

farie82: You can make it a static thing on the flash item. Have an assoc list with the mind (UID or so) or so as key and the value being the cooldown. That way it stays on the flash item and not on the mind and it is global for all flashes

Fox-McCloud:
Trying to balance a game mode that isn't even in rotation, let alone that something like 95+% of people on paradise have no idea how to adapt to/play is a bit...silly.

When rev was in rotation *years* ago, it was largely balanced; if anything, there was a problem with stalemates as opposed to one side winning too often.

code/datums/mind.dm
dearmochi: Seconding @hal9000PR 's status effect suggestion. Let's not introduce more variables in the mind that are used in only one game mode that's not even on rotation.

dearmochi (changes requested):

S34NW:
> 
> 
> Trying to balance a game mode that isn't even in rotation, let alone that something like 95+% of people on paradise have no idea how to adapt to/play is a bit...silly.
> 
> When rev was in rotation _years_ ago, it was largely balanced; if anything, there was a problem with stalemates as opposed to one side winning too often.

When it is played, and it happens every now and then, one headrev grabs a load of flashes and just goes mental. We see half the station converted before sec even properly gears up. I'm not overly sure how that can be considered balanced.
I admit this doesn't fix many of the issues with revs, but it's one small step for something meaningful. If we aren't going to maintain it, why even keep the code around?

dearmochi (approved):

code/datums/status_effects/neutral.dm
```diff
@@ -45,3 +45,11 @@
 
 /datum/status_effect/high_five/on_remove()
 	owner.visible_message("[owner] was left hanging....")
+
+/datum/status_effect/conversion_cooldown
+	id = "conversion_cooldown"
+	duration = 20 SECONDS
+	alert_type = null
```
hal9000PR, line 52: this should have a define in `status_effects.dm` for consistency with all other status effects.

Spacemanspark:
>Would you please explain this bit a little more? I'm not quite sure I understand how/what you mean by this.

Currently there are two sides to a Rev round. Yes, revs spamming a shitty instant conversion flash mechanic in the hallways, which then gives near free reign to go murder hobo command/ security is one such issue.

The other, however, is on security. Namely, rushing down to cargo, securing it, mass ordering loyalty implants and mass implanting crew. If security can't get loyalty implants, beatings begin.

This, in effect, means that revolution rounds will be much, much shorter and more one sided to security; on a round type that's far from encouraging any semblance of RP to begin with.

I think revolution can be made into an actually interesting game mode, but it needs a massive overhaul in purpose, goals and methods of conversion. A simple change such as this doesn't do any of those things, and you can't really baby step with this--not without a concrete idea of where to take the gamemode in the first place.

SteelSlayer (approved):

variableundefined (commented):
I am pretty iffy on this. Objection.

variableundefined:
Closing due to two maint objections. See reasonings above from Fox, Spark and I

PR #16612, opened 2021-08-29
Cults now require a sacrifice before they can make mirror shields

AffectedArc07:
## What Does This PR Do
Cultists must complete their first sacrifice objective before mirror shields can be constructed inside the daemon forge.

## Why It's Good For The Game
This item is fucking garbage and needs some sort of gatekeeping, and people agree this is fine.

## Changelog
:cl: AffectedArc07
tweak: Cultists must complete their first sacrifice objective before they can produce mirror shields
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/cult_datums.dm
+++ b/code/game/gamemodes/cult/cult_datums.dm
@@ -73,6 +73,9 @@
 	var/airlock_unruned_icon_file = 'icons/obj/doors/airlocks/cult/unruned/cult.dmi'
 	var/airlock_unruned_overlays_file = 'icons/obj/doors/airlocks/cult/unruned/cult-overlays.dmi'
 
+	/// Are cultist mirror shields active yet?
+	var/mirror_shields_active = FALSE
+
 
 /datum/cult_info/fire
 	name = "Cult of Kha'Rin"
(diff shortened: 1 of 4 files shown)


code/game/gamemodes/cult/cult_datums.dm
```diff
@@ -73,6 +73,9 @@
 	var/airlock_unruned_icon_file = 'icons/obj/doors/airlocks/cult/unruned/cult.dmi'
 	var/airlock_unruned_overlays_file = 'icons/obj/doors/airlocks/cult/unruned/cult-overlays.dmi'
 
+	/// Are cultist mirror shields active yet?
+	var/mirror_shields_active = FALSE
```
SabreML, line 77: 'Available' might work better than 'Active' for this maybe? More explanatory.
```suggestion
	/// Are cultist mirror shields available yet?
	var/mirror_shields_available = FALSE
```

SabreML (approved):
I've had almost this exact thing on a todo list for nearly a year now (although triggered by the 'rise' state instead), so for what it's worth definitely 👍 from me.

Kyep (approved):

matttheficus:
Without shields it makes attacking the Captain/HoS/Warden targets rather difficult if theyre the first sacrifice goal.

Perhaps make it so they cant be the first target if they cant use shields?

dearmochi (approved):

hal9000PR:
> Without shields it makes attacking the Captain/HoS/Warden targets rather difficult if theyre the first sacrifice goal.
> 
> Perhaps make it so they cant be the first target if they cant use shields?

I think cult is just going to have to adapt to complete these objectives. Maybe after this PR we can see how it plays out and then maybe disable captain from being the first sacrifice objective. But as it stands I think cult will be fine.

Wojz:
> Without shields it makes attacking the Captain/HoS/Warden targets rather difficult if theyre the first sacrifice goal.

Eh. Not really. Three nerds with earmuffs and a multitool do the trick.

variableundefined (approved):

PR #16613, opened 2021-08-29
Removes transformation sting

Qwertytoforty:
## What Does This PR Do

Removes transformation sting as a cling ability.

## Why It's Good For The Game

While a nice ability in theory, this ability is often never used in great ways. Instead of being used as a distraction to throw sec off themself, it is often used in the following ways.

Turning everyone into a fat vulp/tarajan/wingdings grey / whatever with disabilities. (breaking advanced rules)

Turning everyone into you / a known character. (breaking advanced rules)

Using transformation sting to make officers useless / stupidly easy to kill, with no way to rectify it.

When an ability has to have species blacklisted from it, you know it is an issue. Earlier, it was brought to my attention that golems in their entirety might need blacklisting, as such, I considered removing transformation sting as an option.

## Changelog
:cl:
del: Removes transformation sting.
/:cl:

### DIFF
--- a/code/__DEFINES/genetics.dm
+++ b/code/__DEFINES/genetics.dm
@@ -74,7 +74,6 @@
 
 //Species traits.
 #define NO_BLOOD		"no_blood"
-#define NOTRANSSTING	"no_trans_sting"
 #define IS_WHITELISTED 	"whitelisted"
 #define LIPS			"lips"
 #define EXOTIC_COLOR	"exotic_blood_color"
--- a/code/game/gamemodes/changeling/powers/tiny_prick.dm
+++ b/code/game/gamemodes/changeling/powers/tiny_prick.dm
@@ -66,60 +66,6 @@
 		add_attack_logs(user, target, "Unsuccessful sting (changeling)")
 	return 1
 
-/datum/action/changeling/sting/transformation
-	name = "Transformation Sting"
-	desc = "We silently sting a human, injecting a retrovirus that forces them to transform. Costs 50 chemicals."
-	helptext = "The victim will transform much like a changeling would. The effects will be obvious to the victim, and the process will damage our genomes."
-	button_icon_state = "sting_transform"
-	sting_icon = "sting_transform"
-	chemical_cost = 50
-	dna_cost = 3
-	genetic_damage = 100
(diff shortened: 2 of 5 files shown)


variableundefined (approved):

Wojz:
Thank you. I have never seen this ability used in a way that was fun for anyone except the user.

SabreML:
I feel it's important to mention #15869 here, since that was closed a few months ago with two maintainer objections.
Obviously opinions have changed since then since one of said maintainers has approved this version, but I'm unsure what the procedure is for this sort of thing.

ppi13:
As has been said earlier, transformation sting came about as a way to trick sec into believing someone else is a cling.
However, it's far to often used in an abusive manner, and more often then not, used in a way that breaks the advanced rules.

Fox-McCloud:
Still opposed to this. If people are abusing it, then take administrative action against them---using this to trick security is more than a valid option.

I'll be honest, and **cynical** on this one. I really doubt how many people have an issue with the balance implications of this and how much opposition to this is "wahhh, I can't play the round exactly as I want because I look different".

Making it not transfer mutations would be an excellent way to nerf the combat advantages of this without ruining the ability to cause confusion among the crew.

Kyep (approved):
I approve this.
I've never once seen transformation sting used in a way I would consider a net positive for the game.
Its only ever used for grief.
That's not good enough, even for an antag ability.

ppi13:
Another way of nerfing this, instead of preventing mutation transferring, is species blacklists, or limited use per dna. so after 4-5 transform stings, ya lose the dna in the same way as you can only eat x amount of people with 1 dna.

dearmochi (approved):

lectronyx:
I have to agree on transformation sting just being there to grief. If stable mutagen is bad enough to be removed, then the only difference I see with this is that it's antag-only and only really good for griefing. It doesn't make things more interesting. It just upsets everyone except for the user.

It won't be missed.

SabreML (commented):
The `NOTRANSSTING` trait is only used for a transformation sting check, so since the sting is being removed the trait can be as well.
image

Qwertytoforty:
Done!

SteelSlayer (approved):

PR #16621, opened 2021-08-31
Makes anomalies able to spawn in more areas

Qwertytoforty:
## What Does This PR Do

Adds a check in anomalies that let them know what areas exist on a station, to be possible areas to select, rather than only selecting from a smaller list of 17 areas.

Adds more engineering subtypes that would be fine with blowing up.

## Why It's Good For The Game

Anomalies spawning in more areas then currently is good. If you have been playing the game for a while, you will notice that anomalies only ever seem to spawn in a few areas, over and over again. Be it clown / mime office, hydroponics, eva equipment, escape, janitors office, or the atmos air tanks, the areas seem to repeat a bunch.

This PR makes anomalies able to spawn anywhere that is not a restricted area on station. Security? Medical? Vacant office? The HOP office? All now possible locations. Anomalies will still not spawn near the supermatter, in maints, in solars, or in turret protected areas, like the AI sat.

## Changelog
:cl:
tweak: Anomalies can now spawn in a greater amount of areas than before.
/:cl:

### DIFF
--- a/code/modules/events/anomaly.dm
+++ b/code/modules/events/anomaly.dm
@@ -5,11 +5,13 @@
 
 /datum/event/anomaly/proc/findEventArea()
 	var/static/list/allowed_areas
-	var/static/list/existing_areas = list()
-	for(var/area/AR in world)
-		var/turf/picked = safepick(get_area_turfs(AR.type))
-		if(picked && is_station_level(picked.z))
-			existing_areas += AR
+	var/static/list/existing_areas
+	if(!existing_areas)
+		existing_areas = list()
+		for(var/area/AR in world)
+			var/turf/picked = safepick(get_area_turfs(AR.type))
+			if(picked && is_station_level(picked.z))
+				existing_areas += AR
 	if(!allowed_areas)
 		//Places that shouldn't explode
 		var/list/safe_area_types = typecacheof(list(
--- a/code/modules/events/anomaly.dm
+++ b/code/modules/events/anomaly.dm
@@ -5,6 +5,11 @@
 
 /datum/event/anomaly/proc/findEventArea()
 	var/static/list/allowed_areas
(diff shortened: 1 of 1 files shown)


variableundefined (approved):

PacifistDalek:
Sheerly out of curiosity, will this allow anomalies to spawn in new areas built and blueprinted on the edge of the existing station? It really doesn't need to, but I was curious when the check happens, and if it'd allow such shenanigans. :)

SteelSlayer (approved):

code/modules/events/anomaly.dm
```diff
@@ -5,6 +5,11 @@
 
 /datum/event/anomaly/proc/findEventArea()
 	var/static/list/allowed_areas
+	var/static/list/existing_areas = list()
+	for(var/area/AR in world)
+		var/turf/picked = safepick(get_area_turfs(AR.type))
+		if(picked && is_station_level(picked.z))
+			existing_areas += AR
```
SabreML, line 12: Since it's a static list, you can do this so that it only generates once. That way it'll use that pre-made list for any future anomalies rather than having to do the whole loop again:
```suggestion
	var/static/list/existing_areas
	if(!existing_areas)
		existing_areas = list()
		for(var/area/AR in world)
			var/turf/picked = safepick(get_area_turfs(AR.type))
			if(picked && is_station_level(picked.z))
				existing_areas += AR
```
Or alternatively, if you want it to update the list of available areas between anomaly events you could just remove the `/static` part in the variable.

Qwertytoforty: Good idea, no need to generate the list multiple times.

SabreML (approved):

Kyep (approved):

PR #16625, opened 2021-09-02
Spin Emote Changes

Fox-McCloud:
Just some tweaks to the spin emote.

This emote largely replaces the functionality lost of the ssinput system, which prevents you from rapidly spinning with the control key depressed in combination with the direction keys.

As such, I don't think it's appropriate for it to display a message, especially when it's pretty clear what you're doing (flipping is a bit more up for debate without the explicit text). This also makes it a bit less annoying if a bunch of people all start spinning at once while standing in a common location while people are trying to talk.

Also restores old functionality of the `spin` proc. The refactor wasn't worth breaking animations of existing things that used `spin`, not to mention the old way allowed for a lot more fine tuned precision for the total number of spins.

This fixes you spinning and winding up in a direction other than the one you started in.

This also makes it so you can spin while lying down---but not if stunned or weakened.

:cl: Fox McCloud
fix: Fixes spinning not resulting in you being in the same direction you started in
fix: Fixes dog tail chasing and other spin animations.
add: can spin while lying down
del: removes the spin message displayed to those around you
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -449,7 +449,7 @@
 			if(!incapacitated(ignore_lying = TRUE))
 				if(prob(5))
 					spin(32, 1)
-					to_chat(usr, "You spin too much!")
+					to_chat(src, "<span class='warning'>You spin too much!</span>")
 					Dizzy(12)
 					Confused(12)
 				else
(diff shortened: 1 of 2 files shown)


S34NW:
Does spinning while lying down animate similar to the resist when on fire animation?

Fox-McCloud:
Yes.

variableundefined (approved):

PR #16627, opened 2021-09-03
Tweaks carp martial arts projectile reflection message

Deelite34:
## What Does This PR Do
In the past, carp fu user when shot, would delete projectiles from existence, with only hint what happened appearing in chat.
Currently, what happens is hinted both visually and trough sound effect.

With both of those hints present, second part of reflection message "; [p_they()] can't be hit with ranged weapons!" feels like game is holding you by the hand, which is not needed since carp fu rework.
Also, that message suggests, that carp fu user cannot be hit with ranged weapons at all, which is not true. They are not blocked when: 
- player is not using throwing intent
- player is on the ground

## Why It's Good For The Game

Less hand holding, more place for player to interpret what's going on by himself, removal of information which is not mostly true anymore.

## Images of changes

image

## Changelog
:cl:
Tweak: tweaked the carp fu deflection message
/:cl:


SabreML (approved):
Makes sense to me. It's very obvious now what's happening when a carp user deflects a projectile, and the current text could cause confusion with players who aren't aware that there's a way around it.

hal9000PR:
> tweak: Removes redundant second part of projectile reflection by carp-fu user

This sounds unweildy and suggests it changes something to do with the actual mechanics of the deflection. Might i suggest something like `Tweak: tweaked the carp fu deflection message`

hal9000PR (approved):

variableundefined (approved):

PR #16632, opened 2021-09-04
Minor cyborg Move refactor

SteelSlayer:
## What Does This PR Do
`/mob/living/silicon/robot/Move()` is pretty bad currently. It checks if the borg has a module, then checks if that module is the janitor module. All so the borg can clean things on the tile. This PR simply extracts the cleaning code onto the janitor module and makes use of the `COMSIG_MOVABLE_MOVED` signal, so only janitor borgs trigger the cleaning code.

## Changelog
N/A

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -1147,55 +1147,18 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
 		cell = null
 	qdel(src)
 
-#define BORG_CAMERA_BUFFER 30
-/mob/living/silicon/robot/Move(a, b, flag)
-	var/oldLoc = src.loc
+#define BORG_CAMERA_BUFFER 3 SECONDS
+
+/mob/living/silicon/robot/Move(atom/newloc, direct, movetime)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1147,55 +1147,18 @@ GLOBAL_LIST_INIT(robot_verbs_default, list(
...
-								cleaned_human.shoes.clean_blood()
-								cleaned_human.update_inv_shoes()
-							cleaned_human.clean_blood()
-							to_chat(cleaned_human, "<span class='danger'>[src] cleans your face!</span>")
-				if(floor_only)
-					tile.clean_blood()
-		return
+	if(. && !updating && camera)
+		updating = TRUE
+		spawn(BORG_CAMERA_BUFFER)
```
SteelSlayer, line 1157: Yes, there is a spawn here that I left alone. I wasn't sure if it was a good idea to make a timer every time a cyborg moved (technically every 3 seconds, but still). But I can convert this if requested.

SabreML (approved):
Assuming that the spawn is fine, LGTM.

variableundefined (approved):

PR #16637, opened 2021-09-05
Medium chat message refactor

SabreML:
## What Does This PR Do

https://www.paradisestation.org/forum/topic/21473-message-code-refactor-request
- Adds documentation to the `show_message()`, `visible_message()`, and `audible_message()` procs.
- Adds defines for `MSG_VISIBLE` and `MSG_AUDIBLE`, and applies them around the codebase.
- Converts some [legacy](https://github.com/ParadiseSS13/Paradise/blob/master/.github/CONTRIBUTING.md#Legacy-Code) `show_message()` procs into `visible_message()`, `audible_message()`, and `to_chat()` where applicable.
- Adds a `vision_distance()` argument to `visible_message()`, to make the message range customisable.
- Adds an `ignored_mobs` argument to `visible_message()` and `audible_message()`, to exclude mobs from being able to see the message.
```dm
to_chat(target, "<span class='userdanger'>You are hit by [user]!</span>")
user.visible_message("<span class='danger'>[user] hits [target]!</span>",
	"<span class='danger'>You hit [target]!</span>", ignored_mobs = list(target))
```
- Fixes some incorrectly called `visible_message()` procs, where the intended user wouldn't recieve their unique message.

## Why It's Good For The Game

There's a lot of improvements that could be made to the visible and audible message systems, and this only covers a few of them. But I think it's a good start in getting them both up to a good standard of customisation and modularity.
Also replaces some magic numbers with defines and fixes some bugs/oversights, which is always a good thing.

## Changelog
:cl:
tweak: Tweaked some visible and audible chat messages to be more consistent.
/:cl:


SabreML:
I've gone through all the code again and fixed a few bugs that I had missed, with the main one being the `self_message` not getting passed to the parent proc in `/mob/visible_message()`.
Unsetting from DNM, but a review would be appreciated.

variableundefined:
Tried test merging but it conflicts with current test merges so this will have to wait

code/modules/mob/mob.dm
```diff
@@ -70,98 +70,110 @@
 	t+= "<span class='notice'>N2O: [environment.sleeping_agent] \n</span>"
 	t+= "<span class='notice'>Agent B: [environment.agent_b] \n</span>"
 
-	usr.show_message(t, 1)
+	to_chat(usr, t)
 
-/mob/proc/show_message(msg, type, alt, alt_type)//Message, type of message (1 or 2), alternative message, alt message type (1 or 2)
 
-	if(!client)	return
+/// Show a message to this mob (MSG_VISIBLE or MSG_AUDIBLE)
```
farie82, line 76: Can you make this AUTODOC approved?
```dm
/**
 * Desc here
 * Arguments:
 * * msg - desc
 */
```

code/modules/mob/mob.dm
```diff
@@ -70,98 +70,110 @@
...
+/mob/proc/show_message(msg, type, alt_msg, alt_type)//Message, type of message, alternative message, alt message type
+	if(!client)
+		return
 
 	if(type)
-		if(type & 1 && !has_vision(information_only=TRUE))//Vision related
-			if(!( alt ))
+		if((type & MSG_VISIBLE) && !has_vision(TRUE))//Vision related
```
farie82, line 82:
```suggestion
		if(type == MSG_VISIBLE && !has_vision(TRUE))//Vision related
```
Else you suggest that `type` can be both `MSG_VISIBLE` and `MSG_AUDIBLE`. While the code does not properly support that (nor should tbh)

code/modules/mob/mob.dm
```diff
@@ -70,98 +70,110 @@
...
 				return
-			else
-				msg = alt
-				type = alt_type
-		if(type & 2 && !can_hear())//Hearing related
-			if(!( alt ))
+			msg = alt_msg
+			type = alt_type
+
+		if((type & MSG_AUDIBLE) && !can_hear())//Hearing related
```
farie82, line 88:
```suggestion
		if(type == MSG_AUDIBLE && !can_hear())//Hearing related
```
Same here

code/modules/mob/mob.dm
```diff
@@ -70,98 +70,110 @@
...
+		if((type & MSG_VISIBLE) && !has_vision(TRUE))//Vision related
+			if(!alt_msg)
 				return
-			else
-				msg = alt
-				type = alt_type
-		if(type & 2 && !can_hear())//Hearing related
-			if(!( alt ))
+			msg = alt_msg
+			type = alt_type
```
farie82, line 86:
```suggestion
			type = alt_type
			alt_msg = null // alternative is now the main solution
			alt_type = null
```
Doing it like this will tell later code that you don't have an alternative. `msg` is the final stand.
(the comment can be changed ofc)

code/modules/mob/mob.dm
```diff
@@ -70,98 +70,110 @@
...
 				return
-			else
-				msg = alt
-				type = alt_type
-				if(type & 1 && !has_vision(information_only=TRUE))
-					return
+			msg = alt_msg
+			type = alt_type
+			if((type & MSG_VISIBLE) && !has_vision(TRUE))
+				return
```
farie82, line 94:
```suggestion
```
Then this code can be removed as well. Since you will trigger line 89 (`if(!alt_msg)`)

code/modules/mob/mob.dm
```diff
@@ -70,98 +70,110 @@
...
+
 	// Added voice muffling for Issue 41.
 	if(stat == UNCONSCIOUS || (sleeping > 0 && stat != DEAD))
-		to_chat(src, "<I>... You can almost hear someone talking ...</I>")
-	else
-		to_chat(src, msg)
-	return
+		if(type & MSG_VISIBLE) // Visible
```
farie82, line 98:
```suggestion
		if(type == MSG_VISIBLE) // Visible
```

farie82 (changes requested):
Lot of possible refactors in the code you touched. But don't bother. The PR is already big enough.
Good stuff btw

SabreML:
Just for an explanation on the delay:
I'm not able to test the PR ingame at the moment, and while I'm certain the suggested review code would work fine, I wouldn't feel comfortable saying that this is ready to be merged without having double checked everything myself.

SabreML:
I'm really sorry that I haven't addressed the review, but I honestly can't find the motivation to continue coding anymore. Due to that, as well as some recent depression-related issues, I'll be closing the PR.

If anyone else wants to pick this up then please feel free to do so, especially since it's been pre-approved on the forums already.

PR #16642, opened 2021-09-06
Vomiting causes nutrients to go down

montyunk:
## What Does This PR Do
Each time you vomit it will cause your nutrients to go down randomly between the values of -30 and -15

## Why It's Good For The Game
this will hopefully prevent my puketopia and make toilet drinkers suffer a little more 🤯 

## Images of changes
This will hopefully not happen:
image

## Changelog
:cl:
add: Made vomiting remove nutrients.
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1182,6 +1182,8 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 			if(!no_text)
 				visible_message("<span class='warning'>[src] pukes all over [p_them()]self!</span>","<span class='warning'>You puke all over yourself!</span>")
 			location.add_vomit_floor(TRUE)
+	if (nutrition > 0)
+		adjust_nutrition(rand(-30, -15)) //❤ Monty unk
 
 /mob/proc/AddSpell(obj/effect/proc_holder/spell/S)
 	mob_spell_list += S
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1172,8 +1172,8 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 /mob/proc/fakevomit(green = 0, no_text = 0) //for aesthetic vomits that need to be instant and do not stun. -Fox
 	if(stat==DEAD)
 		return
-	var/turf/location = loc
 	if (nutrition > 0)
+		var/turf/location = loc
 		if(istype(location, /turf/simulated))
 			if(green)
 				if(!no_text)
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1173,17 +1173,20 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 	if(stat==DEAD)
 		return
 	var/turf/location = loc
-	if(istype(location, /turf/simulated))
-		if(green)
-			if(!no_text)
-				visible_message("<span class='warning'>[src] vomits up some green goo!</span>","<span class='warning'>You vomit up some green goo!</span>")
-			location.add_vomit_floor(FALSE, TRUE)
-		else
-			if(!no_text)
-				visible_message("<span class='warning'>[src] pukes all over [p_them()]self!</span>","<span class='warning'>You puke all over yourself!</span>")
-			location.add_vomit_floor(TRUE)
 	if (nutrition > 0)
-		adjust_nutrition(rand(-30, -15))
+		if(istype(location, /turf/simulated))
+			if(green)
+				if(!no_text)
+					visible_message("<span class='warning'>[src] vomits up some green goo!</span>","<span class='warning'>You vomit up some green goo!</span>")
+				location.add_vomit_floor(FALSE, TRUE)
+			else
+				if(!no_text)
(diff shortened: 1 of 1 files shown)


meow20:
How much is 15 or 30 in grand scheme of things? Will it make drunk people die of starvation or make it a serious issue only if someone is intentionally drinking toilet water in copious amounts like on the image? Also lmao the comment.

code/modules/mob/mob.dm
```diff
@@ -1182,6 +1182,8 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 			if(!no_text)
 				visible_message("<span class='warning'>[src] pukes all over [p_them()]self!</span>","<span class='warning'>You puke all over yourself!</span>")
 			location.add_vomit_floor(TRUE)
+	if (nutrition > 0)
+		adjust_nutrition(rand(-30, -15)) //❤ Monty unk
```
hal9000PR, line 1186: don't add code comments like this.

montyunk: @hal9000PR please rereview removed the comment

code/modules/mob/mob.dm
```diff
@@ -1182,6 +1182,8 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 			if(!no_text)
 				visible_message("<span class='warning'>[src] pukes all over [p_them()]self!</span>","<span class='warning'>You puke all over yourself!</span>")
 			location.add_vomit_floor(TRUE)
+	if (nutrition > 0)
```
Sirryan2002, line 1185: Logically, I think this check could do a lot more for this proc (and add some more functionality to it).

I believe your intent of this PR is to prevent people from vomit spamming, yes this will give people hypoglycemia but it won't actually stop them from vomiting all the way until they die.

I think you could instead do this.

```
if (nutrition > 0)
    if(istype(location, /turf/simulated))
        VOMIT CODE
    adjust_nutrition(rand(-30, -15))
else
    visible_message("<span class='warning'>[src] dry heaves!</span>","<span class='warning'>You puke but there's nothing left in your stomach!</span>")
    return```

This would allow us to escape this proc quicker and remove the need for calling some procs more than we actually should.

Comment (coloredwax) [reply code/modules/mob/mob.dm:1185]: oh yeah that would also work preventing puking on empty stomach I'll update this in a bit

Comment (montyunk) [reply code/modules/mob/mob.dm:1176]: Changed it to what you suggest also just moved the **adjust_nutrition()**'s scope so you wont lose nutrition when their are no acceptable tiles/when you're missing the message telling you

Comment (montyunk) [reply code/modules/mob/mob.dm:1176]: Would be appreciated if reviewed again @Sirryan2002

Comment (Sirryan2002) [reply code/modules/mob/mob.dm:1176]: Will do, I'm omw back to uni rn so it might be s few hours

File: code/modules/mob/mob.dm [hunk 3]
```diff
@@ -1173,15 +1173,20 @@ GLOBAL_LIST_INIT(slot_equipment_priority, list( \
 	if(stat==DEAD)
 		return
 	var/turf/location = loc
```
Sirryan2002, line 1175: this should probably be moved inside the if statement as well. Other than that it looks good.

montyunk: moved it

hal9000PR (changes requested):
why not make `/datum/reagent/fishwater/on_mob_life(mob/living/M)` call `vomit` instead of editing how `fakevomit` works? the `vomit` proc has a parameter for `lost_nutrition`.

Fox-McCloud:
this was called `fakevomit` for a reason. Adding nutrition drain to it was exactly what this proc was never supposed to be.

montyunk:
I see but why even make fakevomit if you could just call vomit without a stun? @Fox-McCloud @hal9000PR only reason i changed this proc is because i havent look around codebase that much

variableundefined (commented):
What hal9000 said. 

If it is a fake vomit with real effect, why use fake vomit?

montyunk:
Hmmm I'm not against changing it to use the normal vomit proc I'm just questioning why a fake vomit proc even exists when you could just call the normal vomit proc without stuns/nutrient loss seems stupid anyone know the actual reason behind it?

Fox-McCloud:
>I see but why even make fakevomit if you could just call vomit without a stun?

Old proc, also some aspects of the `vomit` proc don't coincide well with this proc.

Also, this isn't going to do a ton to stop people who enjoy doing this kinda thing---they'll just carry food with them or deal with the consequences of being slow and/or dying from starvation in the process of things, IMO.

variableundefined (changes requested):
Please change the fish tank water to use the real vomit instead of modifying the fake vomit to become real.

Sikroth:
While I understand the thought of "maybe this'll stop annoying clowns from puking the entire station", as Fox mentioned I don't think this'll severly impede anyone dedicated enough to do it, just make them carry a bunch of soups with them.

Additionally this'll potentially turn the age old Clown forcefeeding people with toilet water from "harmless, gross prank" to potentially life threatening by inducing hypoglycemia.

variableundefined:
There has been no update for 10 days. 

I am changing my request to an objection on that ground.

variableundefined:
There has been no update in a reasonable amount of time. 

Closing this PR at my own discretion for inactivity. Please make a new one when ready.

PR #16648, opened 2021-09-07
Moves mutes to a GLOB

AffectedArc07:
## What Does This PR Do
Puts mutes on a `GLOB` assoc list instead of on the prefs datum.

## Why It's Good For The Game
This is the sole reason (that I am aware of) that prefs datums have to be cached instead of remaking them on each login, which is a necessity for when I finish the multi instance patch to avoid data inconsistencies.

## Changelog
N/A

### DIFF
--- a/code/game/objects/items/devices/megaphone.dm
+++ b/code/game/objects/items/devices/megaphone.dm
@@ -13,7 +13,7 @@
 	var/list/insultmsg = list("FUCK EVERYONE!", "I'M A TATER!", "ALL SECURITY TO SHOOT ME ON SIGHT!", "I HAVE A BOMB!", "CAPTAIN IS A COMDOM!", "FOR THE SYNDICATE!")
 
 /obj/item/megaphone/attack_self(mob/living/user as mob)
-	if(user.client && (user.client.prefs.muted & MUTE_IC))
+	if(check_mute(user.ckey, MUTE_IC))
 		to_chat(src, "<span class='warning'>You cannot speak in IC (muted).</span>")
 		return
 	if(!ishuman(user))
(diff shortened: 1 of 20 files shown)


code/game/objects/items/devices/megaphone.dm
```diff
@@ -13,7 +13,7 @@
 	var/list/insultmsg = list("FUCK EVERYONE!", "I'M A TATER!", "ALL SECURITY TO SHOOT ME ON SIGHT!", "I HAVE A BOMB!", "CAPTAIN IS A COMDOM!", "FOR THE SYNDICATE!")
 
 /obj/item/megaphone/attack_self(mob/living/user as mob)
-	if(user.client && (user.client.prefs.muted & MUTE_IC))
+	if(check_mute(user.ckey, MUTE_IC))
```
AffectedArc07, line 16: As far as I know you cant even run an attack_self without being connected so /shrug

code/modules/mob/emote.dm
```diff
@@ -112,7 +112,7 @@
 				O.show_message(message, m_type)
 
 /mob/proc/emote_dead(message)
-	if(client.prefs.muted & MUTE_DEADCHAT)
+	if(check_mute(client.ckey, MUTE_DEADCHAT))
```
AffectedArc07, line 115: Oh and for whoever asks "Why check the client ckey instead of the mob ckey"

1. Client ckey make more sense to check since its the client controlling said mob
2. Mob ckeys can sometimes be manually edited in and have scuffed casing
3. This works

variableundefined (approved):

PR #16664, opened 2021-09-09
Adds Additional Doors to DeltaStation Kitchen/Bar

LightFire53:
## What Does This PR Do

Adds an airlock to the kitchen and a windoor to the barkeep so they can access the main dining area without climbing over a table. This comes at the cost of moving an intercom by 1 tile, moving a kitchen sink to the location of that intercom, and the removal of a single bar stool at the barkeep.

Also recolours the flooring in the dining area so that it's a touch less harsh to the eyes.

## Why It's Good For The Game

Denthamos was mentioning that it's kind of a map oversight, plus it allows another route for antagonists to get into the respective areas without having to jump a table.

## Images of changes

**Old Bar**

image

**New Bar**

image

## Changelog
:cl:
tweak: Adds additional access points from the kitchen and bar to the main dining area on DeltaStation
tweak: Recolours the bar floor on DeltaStation
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -9268,7 +9268,14 @@
 /area/maintenance/fore)
 "aBE" = (
 /obj/structure/disposalpipe/segment,
-/obj/structure/table/reinforced,
+/obj/machinery/door/window{
+	dir = 2;
+	name = "Bar Door";
+	req_access_txt = "25"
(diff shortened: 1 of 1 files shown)


variableundefined (approved):

S34NW (changes requested):
While you're making changes here, it would be great if you could fix the directions for the stools

PR #16700, opened 2021-09-12
Adds an item that lets cultists turn teleport runes into portals

Qwertytoforty:
## What Does This PR Do

Adds the reality sunderer, an amulet that cultists can make from the Archives, that can be used on a teleport rune, to make a portal to any other teleport rune. When the amulet is used on a teleport rune, the teleport rune and amulet is consumed to make a portal to the chosen destination rune. The portal last 4 minutes, and makes a visible effect on the other teleport runes location, that a portal is leading there. This means hidden teleport runes are visible when a portal is connected to them.

The portal can be closed by using a cult dagger on it as a cultist, or null rod as a holy person.

The portal can only be closed from it's entrance, and not it's exit.

## Why It's Good For The Game

This item will be quite useful for cult, for a few reasons. Cult at cult base, and you need to summon. Telling everyone to teleport is great, however people sometimes get confused. Making a portal that will take you directly to the summon location can be quite useful. Raiding security? Why not make a portal in, so people don't have to line up 1 at a time, or a portal out of security, for a quick getaway, without having to stand still over a rune when you are trying to escape.

However, the item is not without downside. Crew can tell a portal is leading to a rune, and can be prepared for it, as well as if you make a portal to your cult base to escape somewhere, nothing is stopping crew from throwing a GPS in it, or sending security through it, or closing it to prevent escape. Remember, the portal can only be closed from it's opening, meaning that cult needs to close it when leaving, or they may run into issues.

A potential change that can be made to this pr, is allowing the chaplain to also close portals at the exit point, but not cultists. Leaving as is for now.

Sprite is in icon diff bot, however it may be subject to change, as a different sprite was being worked on for it.

## Changelog
:cl:
add: Adds the reality sunderer, an item cultists can make from the archives, to turn teleport runes into temporary portals, for easier travel.
/:cl:


code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,105 @@
...
+	if(!iscultist(user) && ishuman(user))
+		var/mob/living/carbon/human/H = user
+		to_chat(H, "<span class='cultlarge'>\"So, you want to explore space?\"</span>")
+		to_chat(H, "<span class='warning'>Space flashes around you as you are moved somewhere else!</span>")
+		H.Confused(10)
+		H.flash_eyes(override_blindness_check = TRUE)
+		H.EyeBlind(10)
+		do_teleport(H, get_turf(H), 5, asoundin = 'sound/magic/cult_spell.ogg')
+		qdel(src)
+
```
farie82, line 704:
```suggestion
		return

```
probably want a return here

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,105 @@
...
+/obj/effect/portal/cult
+	name = "eldritch portal"
+	desc = "An evil portal made by dark magics. Suprisingly stable."
+	icon_state = "portal1"
+	failchance = 0
+	precision = FALSE
+	var/obj/effect/temp_visual/cult_portal_exit/exit = null
+
+/obj/effect/portal/cult/New()
```
farie82, line 764: Best to use Initialize here

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,105 @@
...
+	..()
+	exit = new /obj/effect/temp_visual/cult_portal_exit
+	exit.forceMove(target)
+
+/obj/effect/portal/cult/attackby(obj/I, mob/user, params)
+	if(istype(I, /obj/item/melee/cultblade/dagger) && iscultist(user) || istype(I, /obj/item/nullrod) && user.mind.isholy)
+		to_chat(user, "<span class='notice'>You close the portal with your [I].</span>")
+		playsound(src, 'sound/magic/magic_missile.ogg', 100, TRUE)
+		qdel(src)
```
farie82, line 773:
```suggestion
		qdel(src)
		return
```
probably want a return here

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,105 @@
...
+	return ..()
+
+/obj/effect/portal/cult/Destroy()
+	qdel(exit)
```
farie82, line 777: Could it be possible that the temp effect gets deleted first?

farie82 (changes requested):

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,105 @@
...
+	return ..()
+
+/obj/effect/temp_visual/cult_portal_exit
+	name = "eldritch rift"
+	desc = "An exit point for some cult portal. Be on guard, more things may come out of it"
+	icon = 'icons/obj/biomass.dmi'
+	icon_state = "rift"
+	duration = 4 MINUTES
```
farie82, line 785: Minor thing but this is hard tied to the `attempt_portal` at the moment. If the `/obj/effect/portal/cult` ever gets made by something else for like 5 minutes then the exit will be deleted before the entrance gets deleted

code/game/gamemodes/cult/cult_items.dm
Qwertytoforty: Yup, added a check

code/game/gamemodes/cult/cult_items.dm
Qwertytoforty: Changed it from a temp visual

code/game/gamemodes/cult/cult_items.dm
Qwertytoforty: Now initializes

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,107 @@
...
+	name = "reality sunderer"
+	icon = 'icons/obj/cult.dmi'
+	icon_state = "amulet"
+	desc = "Some amulet made out of metal, bluespace crystals, and blood. Allows cultists to open portals over teleport runes, destroying the rune in the process."
+	w_class = WEIGHT_CLASS_SMALL
+
+
+/obj/item/portal_amulet/afterattack(atom/O, mob/user, proximity)
+	. = ..()
+	if(!iscultist(user) && ishuman(user))
```
farie82, line 695: Is there a case where a non-human uses this item? If so then they will be able to use the item without issues. Dunno if that is currently possible though

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,107 @@
...
+/obj/item/portal_amulet/proc/attempt_portal(obj/effect/rune/teleport/R, mob/user)
+	var/list/potential_runes = list()
+	var/list/teleportnames = list()
+	var/list/duplicaterunecount = list()
+	var/turf/T = get_turf(src) //used to tell the other rune where we came from
+
+	for(var/I in GLOB.teleport_runes)
+		var/obj/effect/rune/teleport/target = I
+		var/resultkey = target.listkey
+		if(resultkey in teleportnames)
```
farie82, line 725:
```suggestion
		if(duplicaterunecount[resultkey])
```
Could just use the assoc list here. Can remove the `teleportnames` variable if you do this.
Also can you make the names use snake_case?

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,107 @@
...
+			duplicaterunecount[resultkey]++
+			resultkey = "[resultkey] ([duplicaterunecount[resultkey]])"
+		else
+			teleportnames += resultkey
+			duplicaterunecount[resultkey] = 1
+		if(target != R && is_level_reachable(target.z))
```
farie82, line 731: Might want to do this check before the rest. Since if this isn't valid then the rune won't be shown in the list anyway

code/game/gamemodes/cult/cult_items.dm
Qwertytoforty: I believe the ishuman was needed for the eyeflash to work properly, but I will double check

code/game/gamemodes/cult/cult_items.dm
Qwertytoforty: Can do!

code/game/gamemodes/cult/cult_items.dm
Qwertytoforty: done

PacifistDalek:
I really like like this

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,107 @@
...
+/obj/effect/portal/cult
+	name = "eldritch portal"
+	desc = "An evil portal made by dark magics. Suprisingly stable."
+	icon_state = "portal1"
+	failchance = 0
+	precision = FALSE
+	var/obj/effect/cult_portal_exit/exit = null
+
+/obj/effect/portal/cult/Initialize(mapload, target, creator, lifespan)
+	..()
```
farie82, line 766:
```suggestion
	. = ..()
```

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,107 @@
...
+	if(istype(I, /obj/item/melee/cultblade/dagger) && iscultist(user) || istype(I, /obj/item/nullrod) && user.mind.isholy)
+		to_chat(user, "<span class='notice'>You close the portal with your [I].</span>")
+		playsound(src, 'sound/magic/magic_missile.ogg', 100, TRUE)
+		qdel(src)
+		return
+	return ..()
+
+/obj/effect/portal/cult/Destroy()
+	if(exit)
+		QDEL_NULL(exit)
```
farie82, line 780:
```suggestion
	QDEL_NULL(exit)
```
`QDEL_NULL` does the null check for you as well

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -682,3 +682,107 @@
...
+	name = "reality sunderer"
+	icon = 'icons/obj/cult.dmi'
+	icon_state = "amulet"
+	desc = "Some amulet made out of metal, bluespace crystals, and blood. Allows cultists to open portals over teleport runes, destroying the rune in the process."
+	w_class = WEIGHT_CLASS_SMALL
+
+
+/obj/item/portal_amulet/afterattack(atom/O, mob/user, proximity)
+	. = ..()
+	if(!iscultist(user) && iscarbon(user))
```
farie82, line 695:
```suggestion
	if(!iscultist(user))
		if(!carbon(user))
			return
```
Otherwise, the rare case still stays open where a non-carbon somehow manages to use this amulet and it succeeds since they ain't carbon nor a cultist.
Could also make it so that they don't get blinded or something but still have the teleport happen

variableundefined (approved):

PR #16703, opened 2021-09-12
Increases armour piercing for the WT550 AP magazines.

roarbark:
## What Does This PR Do
Raises the armor_penetration value for the wt550 AP projectiles from 10 to ~~30~~ 40, also lowers its damage from 17 to 15

## Why It's Good For The Game
Loosely said, armour is not very important in most situations on paradise, crew armours can be very easily circumvented by aiming for limbs and otherwise stun combat is the ruler of this land.

What follows is Nerd Shit and my full thought process behind the numbers of this change, since this is a balance change afterall.

From what I can see, concerning the autorifle, two situations are the exception, where bullet armour matters:

Nukies and to some extent Cult, since both get full body armour and can be very hard to stun.

Nukie hardsuits have 50 bullet armour, with elite getting 60 - this means an autorifle shot will deal 10 and 8 damage respectively, with AP dealing 10,2 and 8,5 damage, respectively, the benefit is literally not worth the minute or two you spend walking from cargo to science and waiting to be able to print it.

With the change from this PR the bullets will deal ~~13,6~~ 13,5  and ~~11,9~~ 12 damage. 
The AP number was chosen specifically with the amount shots until a nukie would start rolling for internal damage in mind, since nukies have very strong regenerative and direct healing options for cheap.

Breakpoints on limbs until they start rolling are:
35 head, thorax, groin
30 arms, legs
15 hands, feet

Summing three shots of the new AP means blood red takes ~~13,6/27,2/40,8~~ 13,5/27/40,5 damage and elite takes ~~11,9/23,8/35,7~~ 12/24/36, which means, taking a cycle of chems between the hits in consideration, that blood red will most likely start rolling on the third and elite on the fourth hit. Also to note, hands and feet cannot break on the first shot.

Cult gets empowered robes, which get 40 bullet armour for the body and 30 for the hood. Old AP did less (!) damage against both than current standard WT ammo (11,9 vs 12 and 13,6 vs 14, respectively). New AP should deal ~~17 against the head and 15,3 against the rest of the body~~ 15 damage against head and body. Since this doesn't affect mirror shield break chances, it's more of an afterthought, like mining armour, since that has bullet resists in line with the nukie hardsuits (lmao).

Looking forward, shotgun lethals are likely to go, which means autorifles will become the new gold standard for war rounds. Regardless if they do or not, I wanted there to be a reason to use AP and maybe get the autorifle a bit more closer to the shotgun in terms of power.

## Changelog
:cl:
tweak: increased the autorifle AP bullets armour penetration
add: there is now a new subtype for the wt550 armourpiercing bullet
/:cl:


Wojz:
This is needed and I appreciate the math proving it. As it stands, none of the alternative ammo types for the WT-550s are worth using, and with the removal of combat shotguns, there might be a need for this to be actually useful, as nukies and especially miners can get a bit silly without slugs.

Fox-McCloud:
This isn't the way to go about doing this, as it adjusts every single bullet in the game that utilizes this particular projectile. 

I'd recommend a damage value of 15 and an armor penetration value of 40, personally.

Wojz:
> I'd recommend a damage value of 15 and an armor penetration value of 40, personally.

It's probably better for math reasons alone, yeah.

variableundefined (changes requested):
@roarbark Can you change the number to 40 Armour Penetration and damage of 15. 

I'll need to check with others on whether to create another type

variableundefined:
Please change the typepath of the bullet to something like bullet/armourpiercing/wt550 and apply it only to the wt550

roarbark:
GITHUB GOT HANDS IM SO SORRY

variableundefined (approved):

PR #16710, opened 2021-09-13
Makes us use release dmdoc instead of our own snowflake version

AffectedArc07:
## What Does This PR Do
Uses release dmdoc now because the font isn't times new roman, so its actually bearable.

## Why It's Good For The Repo
Dmdoc will now be:
- Up to date
- Able to parse more
- Darkmode compatible
- Less shit
- Not be a snowflake binary we have in the repo

## Images of changes
image
Imagine this but paracode

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -8,3 +8,4 @@
 	view = "15x15"
 	cache_lifespan = 0	//stops player uploaded stuff from being kept in the rsc past the current session
 	fps = 20 // If this isnt hard-defined, anything relying on this variable before world load will cry a lot
+	name = "Paradise Station 13"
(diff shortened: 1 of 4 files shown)


code/world.dm
```diff
@@ -8,3 +8,4 @@
 	view = "15x15"
 	cache_lifespan = 0	//stops player uploaded stuff from being kept in the rsc past the current session
 	fps = 20 // If this isnt hard-defined, anything relying on this variable before world load will cry a lot
+	name = "Paradise Station 13"
```
AffectedArc07, line 11: Required so dmdoc reads the correct project name instead of just `Space Station 13`.

variableundefined (approved):

PR #16711, opened 2021-09-14
Fix - Removes twohanded wielding when equipping the item to a different slot

### LINKED ISSUE #15482: Hand being occupied when it shouldn't
**Issue Description**:
 When you dual wield a fire axe (maybe something else too) and put it on your back slot, second hand that was not holding axe item itself, keeps the occupied off-hand status until you "drop" it.

 for the "off-hand" occupied status of second hand to be removed when you put away fire axe

**What happened instead**:
 Hand keeps the status of "off-hand" so you cant pick up anything with it.

 Its obvious

 get fire axe, dual wield it, put it on the back slot.

 unsure

**Extra information**:

image

Vi3trice:
## What Does This PR Do
Adds a check for wielded items that can be also equipped onto the back or stuffed in a pocket, and frees ups the hands if that passes.

## Why It's Good For The Game
Fixes #15482

## Images of changes
https://user-images.githubusercontent.com/80771500/133180177-ee9b5284-c5ce-469c-8f7f-5b2950edb022.mp4

## Changelog
:cl:
fix: Putting a twohanded weapon on your back or pockets while holding with two hands will free your hands
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -93,13 +93,11 @@
 	return TRUE
 
 /obj/item/twohanded/mob_can_equip(mob/M, slot) //Unwields twohanded items if they're equipped on the back
-	if(!slot_flags)
-		return FALSE
 	if(wielded && slot_flags)
 		var/obj/item/twohanded/O = M.get_inactive_hand()
 		if(istype(O))
 			O.unwield(M)
-	return TRUE
+	return slot_flags
 
 /obj/item/twohanded/dropped(mob/user)
 	..()
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -92,6 +92,15 @@
 	user.put_in_inactive_hand(O)
 	return TRUE
 
+/obj/item/twohanded/mob_can_equip(mob/M, slot) //Unwields twohanded items if they're equipped on the back
+	if(!slot_flags)
+		return FALSE
+	if(wielded && slot_flags)
+		var/obj/item/twohanded/O = M.get_inactive_hand()
+		if(istype(O))
+			O.unwield(M)
+	return TRUE
+
 /obj/item/twohanded/dropped(mob/user)
 	..()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/twohanded.dm
```diff
@@ -92,6 +92,13 @@
 	user.put_in_inactive_hand(O)
 	return TRUE
 
+/obj/item/twohanded/mob_can_equip(mob/M, slot) //Unwields twohanded items if they're equipped on the back
+	if(wielded && slot_flags)
+		var/obj/item/twohanded/O = M.get_inactive_hand()
+		if(istype(O))
+			O.unwield(M)
+	return slot_flags
+
```
SteelSlayer, line 101: Returning `slot_flags` here is problematic. For example, this makes it possible to equip a fire axe into any open inventory slot (hand slot, pocket slots, etc.). Doing a `return ..()` is probably best here.

Also is there a significance of checking `slot_flags` in the `if`?

SteelSlayer (changes requested):

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -92,6 +92,11 @@
 	user.put_in_inactive_hand(O)
 	return TRUE
 
+/obj/item/twohanded/mob_can_equip(mob/M, slot) //Unwields twohanded items when they're attempted to be equipped to another slot
+	if(wielded)
```
SteelSlayer, line 96: Minor thing, but if you attempt to put a wielded two-handed item into a slot it doesn't belong, it will un-wield it regardless. It might actually be good to add an `&& slot == slot_bitfield_to_slot(slot_flags)` check here to prevent this.

But I'd like to see what other reviewers/maintainers think about this first.

Vi3trice: Yeah, that pretty much was why I commented it as such on that last revision. The `slot_flags` check works fine for the ones that do go on the back, but not for those that handle it through weight class, like smaller weapons and require manually unwielding them.

variableundefined (approved):

PR #16714, opened 2021-09-15
 Fixes the Claw Game arcade machine being impossible to deconstruct.

SteelSlayer:
## What Does This PR Do
As per title. Issue is that the `if(!freeplay)` check is true for the on-station claw games, so the code for crowbar deconstruction was unreachable.

Also allows the Claw Game to be upgraded by the standard Rapid Parts Exchange Device. It is currently only possible to use the bluespace RPED to upgrade the claw game.

## Changelog
:cl:
tweak: Allows the Claw Game to be upgrade by the standard Rapid Parts Exchange Device
fix: Fixes the Claw Game arcade machine being impossible to deconstruct.
/:cl:

### DIFF
--- a/code/modules/arcade/arcade_base.dm
+++ b/code/modules/arcade/arcade_base.dm
@@ -83,7 +83,7 @@
 	return TRUE
 
 /obj/machinery/arcade/crowbar_act(mob/living/user, obj/item/I)
-	if(!component_parts)
+	if(!component_parts || !panel_open)
 		return FALSE
 	default_deconstruction_crowbar(user, I)
 	return TRUE
--- a/code/modules/arcade/arcade_base.dm
+++ b/code/modules/arcade/arcade_base.dm
@@ -59,12 +59,8 @@
 			to_chat(user, "Someone else is already playing this machine, please wait your turn!")
 		return
 
-/obj/machinery/arcade/attackby(obj/item/O as obj, mob/user as mob, params)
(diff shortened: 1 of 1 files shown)


code/modules/arcade/arcade_base.dm
```diff
@@ -76,12 +72,22 @@
...
+/obj/machinery/arcade/screwdriver_act(mob/living/user, obj/item/I)
+	if(!anchored)
+		return FALSE
+	default_deconstruction_screwdriver(user, icon_state, icon_state, I)
+	update_icon()
+	return TRUE
+
+/obj/machinery/arcade/crowbar_act(mob/living/user, obj/item/I)
+	if(!component_parts)
+		return FALSE
```
SabreML, line 87: Just as an optional suggestion thing, since it's so inconsistent in other machines that I don't know what the standard is, and it might end up "feeling" better without it:
```suggestion
	if(!component_parts || !panel_open)
		return FALSE
```
That'll allow people to still whack the claw machine with a crowbar when the panel is closed, rather than there just being no interaction.

SabreML (approved):

code/modules/arcade/arcade_base.dm
SteelSlayer: Yeah, will that check in.

variableundefined (approved):

PR #16718, opened 2021-09-15
Orbit Refactor, Add Orbiter Count

lewcc:
## What Does This PR Do
On the frontend, this PR adds a visual counter of the number of ghosts orbiting an object of interest, inspired by /tg/.

In order to accomplish this on the backend, the orbit system was overhauled to use a component-based system (partly) ported from /tg/station. This system uses signals instead of a while loop to detect when the orbiter or orbitee have moved. The use of a new component still allows other things to be orbited (the tesla energy balls play nicely with this), and ensures everything cleans up and GCs properly as well.

Of note on the backend, this refactor allows you to set the layer that orbiters will be on while they orbit, with their original layer being restored when they stop orbiting.

Also of note is that this PR refactors tesla energy balls to use orbit signals.

## Why It's Good For The Game
Being able to see the number of ghosts around something helps to identify where the points/people of interest might be without having to jump around or follow chat context (if it exists).
On a technical side, having orbiting be component- and signal-based should hopefully cut down on some technical debt, and perhaps even be more performant than the old system.

## Images of changes

https://user-images.githubusercontent.com/89928798/140241614-2d7b3202-d2f9-4991-8aaf-4372350ec361.mp4

## Changelog
:cl:
add: Orbit Menu shows count of ghosts orbiting something.
/:cl:


Fox-McCloud:
Not too keen on this--anything a ghost orbits will now have a ref to the ghost on it, which is...well, a mess and ripe for performance issues.

TG's orbit system works a bit differently than ours does, so it cleans up a bit more gracefully than ours---putting this in is likely to result in uh...issues.

code/__HELPERS/unsorted.dm
```diff
@@ -1538,6 +1539,16 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 	SpinAnimation(rotation_speed, -1, clockwise, rotation_segments)
 
+	// Only create the list if we're adding a new item, so we don't have a bunch of empty lists on /every/ atom
+	if (!A.orbiters)
+		A.orbiters = list()
+
+	A.orbiters += (src)
```
AffectedArc07, line 1546:
```suggestion
	A.orbiters += src
```

code/__HELPERS/unsorted.dm
```diff
@@ -1550,13 +1561,18 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
...
+
 		SpinAnimation(0, 0)
-		transform = cached_transform
+		stop_orbit()
 
 
 
 /atom/movable/proc/stop_orbit()
+	if (orbiting.orbiters)
+		orbiting.orbiters -= (src)
```
AffectedArc07, line 1572:
```suggestion
		orbiting.orbiters -= src
```

code/__HELPERS/unsorted.dm
```diff
@@ -1550,13 +1561,18 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
...
+		if (orbiting.orbiters.len == 0)
```
AffectedArc07, line 1573:
```suggestion
		if (!length(orbiting.orbiters))
```

code/__HELPERS/unsorted.dm
```diff
@@ -1550,13 +1561,18 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
...
+			qdel(orbiting.orbiters)
+			orbiting.orbiters = null
```
AffectedArc07, line 1575:
```suggestion
			orbiting.orbiters = null
```

Just null it out. Calling `qdel()` on a list automatically pipes it to a hard delete.

code/modules/mob/dead/observer/orbit.dm
```diff
@@ -62,6 +62,9 @@
 			continue
 
 		serialized["ref"] = "\ref[M]"
+		var/num_orbiters = M.orbiters?.len
```
AffectedArc07, line 65:
```suggestion
		var/num_orbiters = length(orbiters)
```

AffectedArc07 (changes requested):

code/__HELPERS/unsorted.dm
```diff
@@ -1550,13 +1561,18 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
...
-		transform = cached_transform
+		stop_orbit()
 
 
 
 /atom/movable/proc/stop_orbit()
+	if (orbiting.orbiters)
+		orbiting.orbiters -= (src)
+		if (orbiting.orbiters.len == 0)
+			qdel(orbiting.orbiters)
```
Fox-McCloud, line 1574: This isn't how you null out a list--also lists can't be qdel'd.

code/__HELPERS/unsorted.dm
```diff
@@ -1538,6 +1539,16 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 	SpinAnimation(rotation_speed, -1, clockwise, rotation_segments)
 
+	// Only create the list if we're adding a new item, so we don't have a bunch of empty lists on /every/ atom
+	if (!A.orbiters)
```
Fox-McCloud, line 1543: we don't put spaces after `if` `while`, etc.

code/__HELPERS/unsorted.dm
```diff
@@ -1538,6 +1539,16 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
...
+		A.orbiters = list()
+
+	A.orbiters += (src)
+
+	// Since we're adding a ghost onto a list on any arbitrary atom, make sure we don't have any garbage collection
+	// issues if the ghost gets deleted
+	RegisterSignal(src, COMSIG_PARENT_QDELETING, .proc/stop_orbit)
```
Fox-McCloud, line 1550: we don't care what happens to `src`. That's the atom doing the orbiting. We care about what it's orbiting, which is `A`.

code/__HELPERS/unsorted.dm
lewcc: Wasn't aware of the lazylist defines; those certainly clean up what I was trying to do by a lot.

code/__HELPERS/unsorted.dm
lewcc: Fixed with lazy list define

code/__HELPERS/unsorted.dm
lewcc: Fixed in b43e1b1

lewcc:
Realizing this currently doesn't work right (orbiting someone who's already been orbited causes a stacktrace due to overridden signals), DNM in the meantime

lewcc:
yup, this is turning into a full /tg/ port. here's hoping it cleans up nicely.

lewcc:
Should be ready for review again.

lewcc:
Haven't forgotten about this. I decided to go all the way now that I've had a bit more dm experience and just rework the code to use signals instead of the old while loop construction. 

It now attaches signals to both the orbited atom and orbiter. If the orbited atom moves, all of its orbiters will be moved to that new location as well. If an orbiter moves on its own (by hitting wasd or something), the resulting signal will stop an active orbit.

This should be a good bit more performant and better architected than it previously was, which just dropped observers into their own while loop that slept every 0.6 seconds while checking for new movement.

...oh, and right! it still shows the number of people orbiting an atom.

lewcc:
For the sake of evaluating performance, I ran some (relatively unscientific) comparisons between the old code and new code by orbiting beepsky with one person for about 12 minutes.

For the old loop-based code, `/atom/movable/proc/orbit` used 0.203 self, 0.204 total, and 0 real/over time on the CPU. This was in one call, mind you, as it just sits there.

In comparison, the new event-based code's stats in `/datum/component/orbiter/proc/parent_move_react`, where the new movement update handling takes place, used 0.006 self, 0.006 total, 0.006 real, and no over time on 983 calls.

code/datums/components/orbiter.dm
```diff
@@ -0,0 +1,237 @@
...
+
+/////////////////////////////////////////
+// Atom procs/vars
+
+/// Who the current atom is orbiting
+/atom/movable/var/datum/component/orbiter/orbiting = null
+/atom/movable/var/orbit_params = 0
+
+/// who's orbiting the current atom
+/atom/var/datum/component/orbiter/orbiters = null
```
SteelSlayer, line 203: I really don't like having 3 vars on atoms/movables for this component. Atoms and components should *ideally* have no references to each other, and only interact via signals. I would love to see it refactored that way if possible. 

At the very least, the `orbit_params` var can probably move off of `atom/movable` and onto the component itself because it's only used in this file.

lewcc: That's fair! I think the old-style used to drop a few variables onto the atoms (and the reference implementation does this as well) but I agree it'd probably be nicer to have it clean up like that.

lewcc: I've been giving it a go, and while it seems to mostly work out nicely, I can't seem to completely get rid of references back to the atom currently being orbited, since that provides necessary context when an orbit is ended (how does the user know which atom to stop orbiting without scanning all other atoms?). I'm going to probably have to leave one variable attached for this case, but should be able to clean the others up.

lewcc: So I've removed most direct component references, with the only new variable being stored on an atom is a UID of the atom being followed. It should only be locate()'d when we need to stop an orbit, and its presence alone should help us know if we're orbiting something.

lewcc:
One issue noted in the TM was that orbiting the NAD (in particular, while it was being held) wasn't possible, clicking the button just teleported you to the turf it was on.

dearmochi (approved):

PR #16720, opened 2021-09-16
Makes revive, biodegrade, and freedom break grabs

Qwertytoforty:
## What Does This PR Do

This adds a check to revive, biodegrade, and freedom, that checks if the user is grabbed. If the user is grabbed, it will stun the grabber, do a bit of damage with a sound and message, allowing the user to get away

However, in the case of freedom and biodegrade, due to the stun time of the grab, under normal circumstances, the user would not be able to get away. Thus, if and only if the user is grabbed, these abilities will unstun the user, but not wake them up from sleep, or inject any chemicals. Otherwise, they do not work as antistun.

## Why It's Good For The Game

As of late, many abilities have been getting negated by constant agressive grabs. Freedoms don't work if someone has a neck grab on you, you are still stunned and stuck in the grab, which would be useless. Same with changelings, of late if they have a chance to revive in combat, security will have neck grabbed them, preventing them from doing anything. As such, this pr will break grabs when the 3 abilities are used, allowing the user to get away.

**"Why not lower stun time of grabs?**

I could, but that would still not help these users, as they would still be stunned during these effects, and not have time to get away, even if the grab is broken. (baring revive.)

**"Why is this a problem? Can't security chain stun someone with a baton?**

Indeed, security can. However, batons have limited charge. Eventually, you run out of uses, it's not free. And baton chain stuning wont work on a changeling that revives currently, or anyone that has adrenals. In comparison, neck grabs, anyone can do, it lasts forever, and can not be broken out of, unless the user grabbing you stops. It is significantly better.

## Changelog
:cl:
tweak: Changeling revive, biodegrade, and freedom implants, can now break the user out of grabs when used.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/biodegrade.dm
+++ b/code/game/gamemodes/changeling/powers/biodegrade.dm
@@ -9,7 +9,7 @@
 
 /datum/action/changeling/biodegrade/sting_action(mob/living/carbon/human/user)
 	var/used = FALSE // only one form of shackles removed per use
-	if(!user.restrained() && !user.legcuffed && !istype(user.loc, /obj/structure/closet) && !istype(user.loc, /obj/structure/spider/cocoon))
+	if(!user.restrained() && !user.legcuffed && !istype(user.loc, /obj/structure/closet) && !istype(user.loc, /obj/structure/spider/cocoon) && !user.grabbed_by)
 		to_chat(user, "<span class='warning'>We are already free!</span>")
 		return FALSE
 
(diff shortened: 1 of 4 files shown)


matttheficus:
Drastically needed for clings. Harm Grab + Saw/Knife/etc etc is a gross thing to see with zero counters.

Wojz:
Who'd win, shapeshifting ultimate lifeform, or a regular dude with wirecutters holding it in a chokehold?

Yeah this is needed.

code/game/gamemodes/changeling/powers/biodegrade.dm
```diff
@@ -57,7 +57,17 @@
...
-
+	for(var/obj/item/grab/G in user.grabbed_by)
+		var/mob/living/carbon/M = G.assailant
+		user.visible_message("<span class='warning'>[user] spits acid at [M]'s face and slips out of their grab!</span>")
+		M.Stun(1) //Drops the grab
+		M.apply_damage(5, BURN, "head", M.run_armor_check("head", "melee"))
+		user.SetStunned(0) //This only triggers if they are grabbed, to have them break out of the grab, without the large stun time. If you use biodegrade as an antistun without being grabbed, it will not work
+		user.SetWeakened(0)
+		user.lying = FALSE
+		user.update_canmove()
```
hal9000PR, line 68: `update_canmove()` is already called by `SetWeakened`

Qwertytoforty: Fixed

variableundefined (approved):

PR #16725, opened 2021-09-17
Fixes package wrapper teleportation

SabreML:
## What Does This PR Do

Fixes an exploit allowing items to be instantly teleported using a package wrapper.
This could be achieved by spam clicking the item in the alt-click examine tab. Since that has a delay between updates, the item could still be clicked and wrapped even if it was already inside a package.

(Originally fixed in #16520, but reverted due to various other issues caused by that PR.)

## Why It's Good For The Game

While I don't think it's been done yet, this could be pretty bad if used on an objective item, or on a nuke ops round. For example:

https://user-images.githubusercontent.com/57483089/133825037-df41024b-8c7d-48f3-ab47-8e2fd17d2df7.mp4

## Changelog
:cl:
fix: Fixed package wrappers being able to wrap the same package multiple times.
fix: Fixed being able to teleport package wrapped items.
/:cl:

### DIFF
--- a/code/modules/recycling/sortingmachinery.dm
+++ b/code/modules/recycling/sortingmachinery.dm
@@ -163,6 +163,8 @@
 
 	if(is_type_in_list(target, no_wrap))
 		return
+	if(is_type_in_list(A.loc, list(/obj/item/smallDelivery, /obj/structure/bigDelivery)))
+		return
 	if(target.anchored)
 		return
 	if(target in user)
--- a/code/modules/recycling/sortingmachinery.dm
+++ b/code/modules/recycling/sortingmachinery.dm
@@ -96,14 +96,12 @@
 		var/atom/A = i
 		A.emp_act(severity)
 
-/obj/item/smallDelivery/attack_self(mob/user as mob)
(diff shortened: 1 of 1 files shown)


code/modules/recycling/sortingmachinery.dm
```diff
@@ -96,14 +96,12 @@
 		var/atom/A = i
 		A.emp_act(severity)
 
-/obj/item/smallDelivery/attack_self(mob/user as mob)
-	if(wrapped && wrapped.loc) //sometimes items can disappear. For example, bombs. --rastaf0
-		wrapped.loc = user.loc
+/obj/item/smallDelivery/attack_self(mob/user)
+	if(wrapped?.loc == src) //sometimes items can disappear. For example, bombs. --rastaf0
```
SabreML, line 100: It shouldn't be possible to wrap an item more than once with the added `is_type_in_list()` check below, but just in case another item or an admin teleports it out, this `loc == src` check will stop it being moved back if the package is unwrapped.

variableundefined (approved):

PR #16726, opened 2021-09-18
Removes Honk subtype of the anomalous crystal

Sirryan2002:
## What Does This PR Do
Completely removes the Honk subtype of the anomalous crystal from the code.

## Why It's Good For The Game
If you are not aware of what the honk anomalous crystal does:
- Upon activation (which has infinite uses and a low cooldown)
- Finds every crew member within its view and unequips everything they're wearing
- Forces each of those crew members to get the roundstart clown loadout

This is just an awful item. TG has removed it for a very very good reason. It nukes mindshield and storage implants and force drops everything you have as well as giving you the Comic and Clusmy genes. This includes ERT, Command Members, Security, and literally anyone else in the vicinity of the crystal upon activation. Which means it completely neuters their ability to use weapons as well as removes their only protections against conversion without a counter. 

There is absolutely nothing fun about this and it usually just results in players being upset because of the herculean amount of time it takes to return their player to a normal state. I've only ever seen it used to grief

This has been tested on my local server with no run times or compile errors, interacting with all the crystal types did not appear to cause any runtime either. 

## Changelog
:cl:
del: Removes the Honk subtype of the Anomalous Crystal
/:cl:


Wojz:
It also makes it so you literally need a clean SE to fix it. Mutadone will return you to the clown zone. It's fucking dreadful.

Hope genetics exists and knows what a clean SE is. And that they have enough patience to print them infinitely for like 50 people.

variableundefined (approved):

Kyep (approved):

PR #16732, opened 2021-09-18
Changes the description of a drop of blood

Qwertytoforty:
## What Does This PR Do

Changes the description of a drop of blood from:

desc = "It's red."

To:

desc = "A drop of blood in an ocean of mess."

## Why It's Good For The Game

Blood drips are not always red because of different species, so the old description could be hilariously wrong, with green blood being described as red. Thus, changing the description of blood drips to be something more generic is good.

## Changelog
:cl:
fix: Blood drips no longer say "It's red.", Even when the blood is green, or white, or blue.
/:cl:


SabreML (approved):

Spacemanspark:
How poetic

variableundefined (approved):

PR #16735, opened 2021-09-19
Display the type(s) of selected antags an antag might be in player panel

variableundefined:
## What Does This PR Do

Instead of displaying "Antagonist" if someone is an antagonist, it display the type of antags they are, though it only display the most common types there are, and other antags is lumped under the "Other Antag(s)" category, because of how special_role work under the hood. 

If they are an antag, "(May be other antag) is appended to the display **no matter what**, to prevent an admin from being confused if they turn out to be more than that antag. This preserve the correctness of "Antagonist" but add preciseness to the display. 

## Why It's Good For The Game

Make it much easier for admins to see what kind of relevant antag a person is in the player panel. 

## Images of changes

Sane Case:
dreamseeker_Y0ogwhEgtN

Event Antag:
dreamseeker_lYjPPhSY51

Insane Case just for Lulz:
dreamseeker_gZ0U8zKodQ

## Changelog
:cl:
add: Player Panel now display what type of antagonist an antag is, but limited to the most common antag types due to code limitation.
/:cl:

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -662,6 +662,74 @@ GLOBAL_VAR_INIT(nologevent, 0)
 
 	return 0
 
+/**
+  * A proc that return an array of capitalized strings containing name of the antag types they are
+  *
+  * you'll get something like
+  *
+  * Arguments:
+  * * M - the mob you're checking
+  * *
+  */
+/proc/get_antag_type_strings_list(mob/M as mob) // return an array of all the antag types they are with name
+	var/list/antag_list = new/list()
+
+	if(!SSticker || !SSticker.mode)
+		return 0
+	if(!istype(M))
+		return 0
+	if(M.mind in SSticker.mode.head_revolutionaries)
+		antag_list += "Head Rev"
+	if(M.mind in SSticker.mode.revolutionaries)
+		antag_list += "Revolutionary"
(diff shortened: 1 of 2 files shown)


variableundefined:
image
Decided to refactor the is_special_character proc while at it. 

NONE of the 16 use case of is_special_character check for whether it is ==2 (aka hero fo the gamemode), so I removed those redundant code. Also added DM Doc

code/modules/admin/admin.dm
```diff
@@ -624,43 +624,95 @@ GLOBAL_VAR_INIT(nologevent, 0)
...
+  * A proc that return whether the mob is a "Special Character" aka Antagonist
+  *
+  *
+  *
+  * Arguments:
+  * * M - the mob you're checking
+  * *
+  */
+/proc/is_special_character(mob/M as mob)
+	if(!SSticker || !SSticker.mode)
```
AffectedArc07, line 637:
```suggestion
/**
  * A proc that return whether the mob is a "Special Character" aka Antagonist
  *
  * Arguments:
  * * M - the mob you're checking
  */
/proc/is_special_character(mob/M as mob)
	if(!SSticker.mode)
```

Cleans up doc and we dont need any `!SSticker` checks. `SSticker` will always exist.

variableundefined:
Updated

PR #16738, opened 2021-09-19
Adds a limit of 5 crates per round to null crates

Qwertytoforty:
## What Does This PR Do

Adds an times_ordered variable and order_limit variable to supply crates. Times ordered is increased each time a specific crate is ordered. order_limit is the number of times a crate can be ordered (and approved) before the supply of the crate runs out. All crates but one have a limit of -1, meaning they can be ordered endlessly. These variables can be changed by admins through var edit, allowing crates to be restricted or limited for an event.

Adds a order limit of 5 to null crates, allowing at most 5 to be ordered during a shift.

## Why It's Good For The Game

Null crates are strong. With purely just points from rnd, you can order 10-12 crates. Add in say, 200 plasma, and that is another 7 crates. It is quite easy to get an insane amount of items, and even TC, from these crates. All you have to do is order them, and hope no one notices for 2 minutes, and you are home free. As such, this PR limits them to 5 at most.

**"Why not just increase the price?"**

I could increase the base price, however eventually we are moving to tech webs, which will reduce the points cargo gets. While this could lower the amount of crates order, plasma still exists. I also don't want to prevent crates from being ordered at all, if mining is poor. As such, as there is no clear good number to increase it to, I thought a limit is better.

**"Why not make the price scale as more crates are ordered?"**

I considered this as well. Another potential alternative, make the crate scale up in price by a number each time one is ordered.

However, not sure how well that would work with multiple crates being ordered at once, and might confuse people when they order 3 crates for 100, but when they go to approve, they pay 450 total, or something.  As such, a flat limit seemed better.

**"Why not nerf what is in the crates?"**

Theoretically, crates could be changed to use 10 TC worth of random items. However, not sure if that is better or not, to be honest. It feels like you would get the same things over and over again, or get something like a trick revolver, 2 flags, and a cham toolbelt, or something. As such, I chose to keep the crates functioning as they currently do, while adding the limit.

## Changelog
:cl:
add: Supply packs can now have a limit of how many times it can be ordered in a round.
add: Null crates can now only be ordered 5 times in a round.
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -55,6 +55,10 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	var/list/announce_beacons = list() // Particular beacons that we'll notify the relevant department when we reach
 	var/special = FALSE //Event/Station Goals/Admin enabled packs
 	var/special_enabled = FALSE
+	/// The number of times one can order a cargo crate, before it becomes restricted. -1 for infinite
+	var/order_limit = -1
+	/// Number of times a crate has been ordered in a shift
+	var/times_ordered = 0
 	/// List of names for being done in TGUI
 	var/list/ui_manifest = list()
 
@@ -200,6 +204,7 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	containertype = /obj/structure/closet/crate
 	containername = "crate"
 	hidden = 1
+	order_limit = 5
 
 //////////////////////////////////////////////////////////////////////////////
 //////////////////////////// Security ////////////////////////////////////////
--- a/code/modules/shuttle/supply.dm
+++ b/code/modules/shuttle/supply.dm
@@ -521,17 +521,20 @@
 				visible_message("<b>[src]</b>'s monitor flashes, \"[world.time - reqtime] seconds remaining until another requisition form may be printed.\"")
 				return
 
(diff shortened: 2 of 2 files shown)


variableundefined (approved):

code/datums/supplypacks.dm
```diff
@@ -55,6 +55,10 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 	var/list/announce_beacons = list() // Particular beacons that we'll notify the relevant department when we reach
 	var/special = FALSE //Event/Station Goals/Admin enabled packs
 	var/special_enabled = FALSE
+	/// The number of times one can order a cargo crate, before it becomes restricted. -1 for infinite
+	var/order_limit = -1
```
AffectedArc07, line 59: Broader idea here. If the initial value of this is -1 on supplypack/New(), set it to a random value between 800-900, (quite literally, `rand(800, 900)`), an show the stock level on the supply console.

This also opens up opportunities for other crates to be restricted in a nicer fashion.

Spacemanspark:
Clearly every five crates ordered should ship an extra with a bomb attached inside, which goes off automatically when opened.
:^)

code/datums/supplypacks.dm
Qwertytoforty: Working on that, currently got it working, however going to try to get the number displayed to update each time an order is approved, rather than having to open and close the menu.

PidgeyThePirate:
Another possibility is limiting null crates to one or two per shipment, thus requiring that the user continually risks discovery in order to get their epic loots.

PR #15675, opened 2021-03-05
Implements Better His Grace

Fox-McCloud:
Implements TG's version of His Grace.

This version is just a lot better coded than ours---ours is heavily based on Goon, and, as such, isn't the best coded.

This version of His Grace is just easier to utilize and is far more intuitive on how to wield it.

It does function a bit differently, but the overall theme is more or less the same.

Some highlights:

- No longer do you have to fidget with grabbing someone and shoving them into the toolbox; just hitting someone who's unconscious will cause His Grace to eat them
  - This makes the merging grab+pull refactor a bit easier, yay. 
- His Grace no longer gives you 100% status effect immunity, but it still makes you immune to stuns and weakens
- The heal amount scales with His Grace's bloodlust; the more you kill, the more it heals
- Not having His Grace with you incurs an immediate punishment, as opposed to one over time
- If His Grace turns on his owner, he'll become rabid for a time, attacking anyone nearby and consuming them if they become unconscious
- His Grace can ascend; if you manage to kill 25 victims that posses a mind, he becomes....something more. His blood lust is permanently satiated, his color will take on a gold hue, and he'll deal extra damage

This also adds the wind up toolbox; grab it at the arcade for 100 tickets; it vibrates and rumbles while making an ominous noise when you wind it up.

:cl: Fox McCloud
tweak: His Grace completely reworked; it's now a higher risk, higher reward item and is a lot easier to use; just attack unconscious victims to consume them
/:cl:

PR #16741, opened 2021-09-19
Fixes unstoppable spinning/orbiting

### LINKED ISSUE #16767: Fliponium never stops flipping
**Issue Description**:
Fliponium no longer wears off, you spin until you slip or lie down.

For the spin to run out eventually when the chem runs out.

**What happened instead**:
Infinite spinny boys.

It is _extremely_ LRP to see everyone running around spinning

Probably with 514 and the other animation bugs we have.

SabreML:
## What Does This PR Do

Fixes a bug from #15675, which caused the animation from `SpinAnimate()` to be unstoppable (in some cases) if set to an infinite duration.
(Ghost orbits, fliptonium, etc.)

This was caused by the `SpinAnimate()` proc giving everything the [`ANIMATION_PARALLEL`](http://www.byond.com/docs/ref/#/proc/animate) flag by default, making the next `animate()` call stack on top rather than overriding the first.

https://user-images.githubusercontent.com/57483089/133940134-b89d2092-eb88-42f3-ac0a-bdb6b1ba03ba.mp4

https://user-images.githubusercontent.com/57483089/133940113-c4fe1838-9cb8-4fd1-b6b3-5cb39b1c93f8.mp4

## Why It's Good For The Game

^^This probably shouldn't be happening^^
(Fixes #16767)

## Changelog
:cl:
fix: Fixed a bug causing unstoppable orbiting/spinning in some circumstances.
/:cl:


Fox-McCloud:
What's the bug exactly, minus orbit?

In general, I'd rather address the spinning that's *incorrect* than flipping everything on its head for args.

SabreML:
> What's the bug exactly, minus orbit?
> 
> In general, I'd rather address the spinning that's _incorrect_ than flipping everything on its head for args.

As far as I've been able to tell, the `ANIMATION_PARALLEL` flag in `SpinAnimate()` is causing things that call it more than once to stack the animations on top of each other, rather than overriding the previous one.
Both `orbit()` and the `on_mob_life()` for fliptonium are built around that behaviour:
https://github.com/ParadiseSS13/Paradise/blob/8bcd5b8230d57521c5512c61d87d747889b47fac/code/__HELPERS/unsorted.dm#L1542-L1558
https://github.com/ParadiseSS13/Paradise/blob/8bcd5b8230d57521c5512c61d87d747889b47fac/code/modules/reagents/chemistry/reagents/drugs.dm#L568-L585

This happens for everything that calls the proc, but it's only really noticable if they have `loops` set to `-1`, since then the animation will keep going infinitely and won't expire before the next one starts.

Adding `parallel = FALSE` onto those as well *would* work as an alternative, but currently there's nothing in the code that actually uses the argument, so I thought it would be better to have it off by default to avoid someone getting confused by this again in the future.

PR #16747, opened 2021-09-20
Added safety turf proc, support guardians and swarmers can warp their targets to locations with safe levels of toxins or sleeping agent

### LINKED ISSUE #16376: Support guardian beacon too sensitive
**Issue Description**:
So this isn't strictly a bug, but functionally it appears as one to the end user.
Support guardian code checks for safe air when teleporting to a beacon:
`if(Z.oxygen >= 16 && !Z.toxins && Z.carbon_dioxide < 10 && !Z.sleeping_agent)`

`code\game\gamemodes\miniantags\guardian\types\healer.dm` line 128

Unfortunately, this also picks up trace amounts of toxins or sleeping_agent that are undetectable to the user in any way.

Guardian teleports its user to the very safe and habitable medbay.

**What happened instead**:
toxins = 0.00575326 blocked the teleport, a value that is undetectable to the user in any way. (Only looking at the air in VV panel revealed the cause.)

Since the beacon was added, AFAIK.

**Extra information**:

Vi3trice:
## What Does This PR Do
This splits the safety checks from `find_safe_turf` into its own `turf` proc, `is_safe`, and call that instead.

This allows a turf check without having to do the whole finding a random turf loop that occurs for swarmers.

This also adds the safety thresholds for toxins and sleeping agents from the lungs organ to the safety check. The threshold for sleeping agent is below 1, as is technically "safe", but induces temporary paralysis instead of full on sleep.

## Why It's Good For The Game
Fixes #16376

## Images of changes
ToxinSleep

## Changelog
:cl:
add: Safety turf check, with toxin and sleeping agent thresholds instead of detecting trace amounts
tweak: Guardian beacon warp, swarmer warp use the new safety turf check
/:cl:


farie82 (approved):
LGTM.
One request though if you're up for a challenge.
`find_safe_turf` has similar logic for finding a random safe turf. This one has the same issue. It might be a good idea to centralise the safe turf checking and make `find_safe_turf` and the PRs teleport use the same code.
There might be other instances where a turf is checked for the gasses, pressure, etc inside of it.

code/datums/helper_datums/teleport.dm
```diff
@@ -198,36 +198,27 @@
...
-					continue
-
+		if(safety_turf(random_location))
 		// DING! You have passed the gauntlet, and are "probably" safe.
-		return F
+			var/turf/simulated/floor/F = random_location
+			return F
+
+// Checks if the turf is safe to be on
+/proc/safety_turf(location)
```
farie82, line 207: Could you define this proc on `/turf` level instead? And have the turf types override the implementation?
Saves quite a lot of checks and makes it more OO.
The implementation on `/turf` can just return `FALSE` and really only floor turfs and lava turfs need to override it.
Also makes the next question easier, can you also refactor it to use boolean values `TRUE/FALSE` and use early returns etc?

Danke for taking the extra step by the way

code/game/turfs/simulated/floor.dm
```diff
@@ -82,6 +82,22 @@ GLOBAL_LIST_INIT(icons_to_ignore_at_floor_init, list("damaged1","damaged2","dama
 		if(A.level == 3)
 			return 1
 
+// Checks if the turf is safe to be on
+/turf/simulated/floor/safety_turf()
+	if(islava(src))
```
farie82, line 87: Instead of using a typecheck here you can better make an override for this proc on `/turf/simulated/floor/plating/lava`.
Be sure to call the parent as well to check the air content

code/game/turfs/simulated/floor.dm
```diff
@@ -82,6 +82,22 @@ GLOBAL_LIST_INIT(icons_to_ignore_at_floor_init, list("damaged1","damaged2","dama
...
+		var/turf/simulated/floor/plating/lava/L = src
+		if(!L.is_safe())
+			return FALSE
+
+	if(!src.air)
+		return FALSE
+	var/datum/gas_mixture/Z = src.air
```
farie82, line 94:
```suggestion
	if(!air)
		return FALSE
	var/datum/gas_mixture/Z = air
```
`src` is implied

code/datums/helper_datums/teleport.dm
```diff
@@ -198,27 +198,8 @@
 		var/z = pick(zlevels)
 		var/random_location = locate(x,y,z)
 
-		if(safety_turf(random_location))
-		// DING! You have passed the gauntlet, and are "probably" safe.
+		if(isfloorturf(random_location))
 			var/turf/simulated/floor/F = random_location
```
farie82: You can most likely just cast it to `/turf` and then call `safety_turf`

code/game/turfs/turf.dm
```diff
@@ -249,6 +249,9 @@
 /turf/proc/BeforeChange()
 	return
 
+/turf/proc/safety_turf()
```
farie82, line 252:
```suggestion
/turf/proc/is_safe()
```
Something the like is a bit more descriptive of what the proc does

code/game/gamemodes/miniantags/guardian/types/healer.dm
```diff
@@ -121,15 +121,17 @@
...
-					do_teleport(A, beacon, 0)
-					new /obj/effect/temp_visual/guardian/phase(get_turf(A))
-				else
-					to_chat(src, "<span class='danger'>The beacon isn't in a safe location!</span>")
-			else
-				to_chat(src, "<span class='danger'>You need a beacon to warp things!</span>")
-	else
-		to_chat(src, "<span class='danger'>You need to hold still!</span>")
+		if(A.anchored)
+			to_chat(src, "<span class='danger'>You need to hold still!</span>")
```
farie82, line 125: This one belonged to the `else` of the `if(do_mob(src, A, 50))` check. The anchor one had no message

farie82 (changes requested):
Might want to check also add an override for `/turf/simulated/floor/chasm` to return `FALSE`. Since those are never safe

code/datums/helper_datums/teleport.dm
Vi3trice: I don't quite understand that part

farie82: Your new proc is defined on `/turf` level. So you can call it using a `/turf` object. `random_location` should always be a `/turf` object (however it isn't casted as one currently, just `var/random_location`).
If you change it to `var/turf/random_location` then you can use `if(random_location.is_safe())` without casting it to `/turf/simulated/floor`

code/game/turfs/simulated/floor/lava.dm
```diff
@@ -50,13 +50,10 @@
...
-	var/static/list/lava_safeties_typecache = typecacheof(list(/obj/structure/lattice/catwalk, /obj/structure/stone_tile))
-	var/list/found_safeties = typecache_filter_list(contents, lava_safeties_typecache)
-	for(var/obj/structure/stone_tile/S in found_safeties)
-		if(S.fallen)
-			LAZYREMOVE(found_safeties, S)
-	return LAZYLEN(found_safeties)
+/turf/simulated/floor/plating/lava/is_safe()
+	if(find_safeties() && ..())
+		return TRUE
+	return
```
farie82, line 56:
```suggestion
	return FALSE
```

code/game/turfs/simulated/floor/chasm.dm
```diff
@@ -67,14 +67,10 @@
...
-	var/static/list/chasm_safeties_typecache = typecacheof(list(/obj/structure/lattice/catwalk, /obj/structure/stone_tile))
-	var/list/found_safeties = typecache_filter_list(contents, chasm_safeties_typecache)
-	for(var/obj/structure/stone_tile/S in found_safeties)
-		if(S.fallen)
-			LAZYREMOVE(found_safeties, S)
-	return LAZYLEN(found_safeties)
+/turf/simulated/floor/chasm/is_safe()
+	if(find_safeties() && ..())
+		return TRUE
+	return
```
farie82, line 73:
```suggestion
	return FALSE
```

code/game/turfs/simulated/floor.dm
```diff
@@ -82,6 +82,26 @@ GLOBAL_LIST_INIT(icons_to_ignore_at_floor_init, list("damaged1","damaged2","dama
...
+		return FALSE
+	return TRUE
+
+// Checks if there is foothold over the turf
+/turf/simulated/floor/proc/find_safeties()
+	var/static/list/safeties_typecache = typecacheof(list(/obj/structure/lattice/catwalk, /obj/structure/stone_tile))
+	var/list/found_safeties = typecache_filter_list(contents, safeties_typecache)
+	for(var/obj/structure/stone_tile/S in found_safeties)
+		if(S.fallen)
+			LAZYREMOVE(found_safeties, S)
```
farie82, line 102: not sure why this is made lazy but....
Just wondering. Don't have to do anything with this since it was already here.

code/datums/helper_datums/teleport.dm
```diff
@@ -196,38 +196,7 @@
...
-
-		if(extended_safety_checks)
-			if(islava(F)) //chasms aren't /floor, and so are pre-filtered
-				var/turf/simulated/floor/plating/lava/L = F
-				if(!L.is_safe())
-					continue
-
-		// DING! You have passed the gauntlet, and are "probably" safe.
-		return F
+		var/turf/random_location = locate(x,y,z)
```
farie82, line 199:
```suggestion
		var/turf/random_location = locate(x, y, z)
```

farie82 (commented):
Last minor things. Looks good to me otherwise!

code/game/turfs/simulated/floor.dm
```diff
@@ -82,6 +82,26 @@ GLOBAL_LIST_INIT(icons_to_ignore_at_floor_init, list("damaged1","damaged2","dama
...
 			return 1
 
+// Checks if the turf is safe to be on
+/turf/simulated/floor/is_safe()
+	if(!air)
+		return FALSE
+	var/datum/gas_mixture/Z = air
+	var/pressure = Z.return_pressure()
+	// Can most things breathe and tolerate the temperature and pressure?
+	if(Z.oxygen < 16 || Z.toxins >= 0.05 || Z.carbon_dioxide >= 10 || Z.sleeping_agent >= 1 || (Z.temperature <= 270) || (Z.temperature >= 360) || (pressure <= 20) || (pressure >= 550))
```
variableundefined, line 92: Check like this should be refactored out later on

variableundefined (approved):

Denthamos:
So, this would prevent a guardian holoparasite from teleporting its user to a turf if that turf is filled with plasma gas if I understand that correctly?  What if they want to teleport to such on purpose being an atmos tech traitor?

Vi3trice:
> So, this would prevent a guardian holoparasite from teleporting its user to a turf if that turf is filled with plasma gas if I understand that correctly? What if they want to teleport to such on purpose being an atmos tech traitor?

They already can't warp to a turf with trace amounts of toxins or sleeping agent, even if it is within safe levels.

PR #16748, opened 2021-09-20
Spamming scream gives you a dry throat

S34NW:
## What Does This PR Do
This PR adds a new status effect, "Dry throat".
Screaming x times in x seconds gives you this effect. (Currently 3 in 30, but that can be changed easily)
While it is active, you cannot scream any further, your character instead gasps.
Dry throat goes away after 1 minute, or when you consume a drink.

It appears as a status effect in the top right of the main game window, as with hunger and other effects.

image

IPCs have a different alert, overheating voicebox, but it is functionally identical.

## Why It's Good For The Game
Scream spamming is a pain on the ears, it's really lame, and people really overdo it whenever kidnapped. It's useful to do once or twice, but if they didn't hear you after 3 screams they aren't coming after 10.

## Changelog
:cl:
add: Added "Dry throat" status effect when screaming too much, preventing you from screaming more until you drink or wait a minute.
/:cl:


Spacemanspark:
How does it impact synths

Darkmight9:
> How does it impact synths

Voice box overload/error

S34NW:
> 
> 
> How does it impact synths

Functionally, or "lore" wise

datlo:
Add an alternative text/icon for synthetics that their voice box is currently recharging or whatever?

Spacemanspark:
>Functionally, or "lore" wise?

Both, really.

I'd like them to be included with this because fucking IPC screams are far worse than the regular human screams when spammed, but in a natural way ICly.

Carthusia:
What about vox? Will this give them dry throat as they talk? 

They scream uncontrollably potentialy 5 times within 15 seconds due to the way their speech is handled via a random % chance of a scream being made whenever they say somthing.

Not sure WHY it forces vox to scream as they talk, but it does.

matttheficus:
Thank god, i wouldnt need tape or to kill every nerd who yells their lungs out in Maints.

Cocacolagua:
I would like if attempting to scream while having a dry throat would cause you to whimper rather than nothing happening, however I cannot confirm if whimpering can be heard through walls / is too loud.

S34NW:
> 
> 
> I would like if attempting to scream while having a dry throat would cause you to whimper rather than nothing happening, however I cannot confirm if whimpering can be heard through walls / is too loud.

It currently causes gasping

S34NW:
> 
> 
> What about vox? Will this give them dry throat as they talk?
> 
> They scream uncontrollably potentialy 5 times within 15 seconds due to the way their speech is handled via a random % chance of a scream being made whenever they say somthing.
> 
> Not sure WHY it forces vox to scream as they talk, but it does.

It should not, as this handles the emote. AFAIK vox scream talking just plays the sound, as opposed to doing a full emote.

Rurik123:
> Scream spamming is a pain on the ears, it's really lame, 

If it hurts your ears, then quite frankly your volume is set too loud. Exception being on the same tile as a vox shrieking of course. 

> and people really overdo it whenever kidnapped.

And how do people "really overdo it" when kidnapping? It makes both IC and OOC sense you scream for help as loud as possible when a vampire is sucking your blood in a false wall, or changeling about to strangle you. Isn't cheesy either, since there are two very easy ways for an antag to shut you up. Duct tape or Strangulation. Ways that, unlike a flash to stun borgs, are universally available to every antag.

And even if those two methods didn't exist, if sec hears the screams and responds to false wall they wouldn't have found otherwise, then that is a **good thing!** Now you have an engagement for sec, an opportunity for the victim to escape, and action for you. That has far more value based on interaction with all 3 parties than if sec simply walked by the false wall.

> It's useful to do once or twice, but if they didn't hear you after 3 screams they aren't coming after 10.

This isn't true, at least not on box. I could maybe see it on delta or meta, but on box a minute (the cooldown after first 3 screams if I read correctly) makes the literal difference between life or death when it comes to sec locating and rescuing you.

I feel like I'm missing something crucial here because I can't quite comprehend the reasoning, nor this coming out of nowhere to pad lazy kidnappers.

Spacemanspark:
It's being made primarily because some people love sitting in public areas and spamming emotes.

Exavere:
Given that someone can just drink to clear the status effect of the Dry Throat this PR won't really change that Spark - they can still spam all they want they just need to grab one of 10,000 cans of soda available- this is more geared towards Kidnappings. 
 On that topic Rurik does have a good point that you can things to easily prevent it, hell one of the ways is literally built right into your character via Strangulation.  

I can go 50/50, but if this PR is to just make it easier on Antags they already have many easy ways to go about preventing screaming and if it's to reduce spam screaming in general than people soothing their throats makes it kind of irrelevant.

Spacemanspark:
That's a good point.

Honestly I'd think the PR would be better if the water drinking part was removed.

By the time someone manages to scream three times or so, an antag will typically have done what they needed to do to keep you quiet anyways.

variableundefined (approved):

Fox-McCloud:
Adding a human level var for *one* thing like this is a bit much.

The sound system dramatically nerfed the effective range of screaming, so I'm really not sure this is much of a problem anymore.

The CD on emotes is already what it is to add a bit of a respite from this, but if you put a bunch of people together all utilzing the same emote---well, yeah, it's going to grind some people's gears once in a while.

Fundamentally, this isn't going to change people's behavior--just shift it. Instead of screaming, it'll be spamming sneeze, cough, deathgasp, growl, howl, rumble, ping, or whatever else makes noise.

It's also pretty easy to counter as an antag---just tape someone's mouth shut---it literally takes 5 seconds to do with an item that is insanely common.

code/datums/status_effects/debuffs.dm
```diff
@@ -136,3 +136,28 @@
...
+
+/datum/status_effect/dry_throat
+	id = "dry_throat"
+	duration = 1 MINUTES
+	status_type = STATUS_EFFECT_UNIQUE
+	alert_type = /obj/screen/alert/status_effect/dry_throat
+
+/datum/status_effect/dry_throat/on_creation(mob/living/new_owner, ...)
+	if(new_owner && ismachineperson(new_owner))
+		alert_type = /obj/screen/alert/status_effect/overheating_voicebox
```
Fox-McCloud, line 148: ❄️

S34NW:
>
> Fundamentally, this isn't going to change people's behavior--just shift it. Instead of screaming, it'll be spamming sneeze, cough, deathgasp, growl, howl, rumble, ping, or whatever else makes noise.

You know, the more I think about it the more I agree with this statement. I'm not sure how to address these sorts of issues, but I suppose this way is not the right way.

PR #16751, opened 2021-09-20
Allow the game to call votes whenever. Fixes the mapping vote not happening after a shuttle vote

farie82:
## What Does This PR Do
Allows the game to call votes whenever. AKA when `usr` is null.

## Why It's Good For The Game
Allows the crew to vote if they want to leave and to vote what map they want to play next.

## Changelog
:cl:
fix: The game can now make votes whenever it wants. Fixes the mapping or any other vote not happening when another vote happened beforehand. Be it done triggered by admins or by the game itself
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -209,7 +209,7 @@ SUBSYSTEM_DEF(vote)
 
 /datum/controller/subsystem/vote/proc/initiate_vote(vote_type, initiator_key, code_invoked = FALSE)
 	if(!mode)
-		if(started_time != null && !check_rights(R_ADMIN))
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/vote.dm
```diff
@@ -209,7 +209,7 @@ SUBSYSTEM_DEF(vote)
 
 /datum/controller/subsystem/vote/proc/initiate_vote(vote_type, initiator_key, code_invoked = FALSE)
```
farie82, line 210: I ain't using the `initiator_key` since then somebody called `the server` can force a vote whenever if they manage to call this

AffectedArc07 (approved):
Please also put in the CL that this fixes end round votes not calling if admins vote earlier in the round

AffectedArc07 (approved):

PR #16758, opened 2021-09-21
Fixed SM shutter switch, now accessible to athmos

Yoekkul:
## What Does This PR Do
On Delta station: Updated access level to SM shutters switch so athmos techs can also use it (since they already have access to the rest of the room).
(This is my first PR and thus very small. Any feedback highly welcome)

## Why It's Good For The Game
Makes access levels consistent within SM room.

## Changelog
:cl:
fix: Fixed SM shutter access level
/:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -12966,7 +12966,7 @@
 	id = "engsm";
 	name = "Radiation Shutters Control";
 	pixel_y = 24;
-	req_access_txt = "24; 10"
+	req_one_access_txt = "24; 10"
 	},
 /obj/structure/cable/yellow{
 	d1 = 4;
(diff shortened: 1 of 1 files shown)


S34NW:
I think you need to make a slight tweak
`req_access_txt` calls for ALL listed accesses
`req_one_access_txt` calls for ONE of any listed accesses

you want to use the second one

SabreML:
Makes sense to me, since they've got access to everything else in the area. Atmos techs should get engine access on *every* map really, rather than just Delta, but that's something for the future.

The one change I would suggest is to make the changelog in the PR a bit more specific, since it's the only thing that actually shows up in the game.
Something like this maybe, or similar:
>:cl:
fix: Fixed atmos techs not having SM shutter access on Delta
/:cl:

PR #16759, opened 2021-09-21
Adds a webmap button

AffectedArc07:
## What Does This PR Do
Adds a webmap button
image

## Why It's Good For The Game
Fed up of "whats this map" mhelps

## Changelog
:cl: AffectedArc07
add: Added a button for the current map's webmap
/:cl:

### DIFF
--- a/code/game/verbs/webmap.dm
+++ b/code/game/verbs/webmap.dm
@@ -1,0 +1,12 @@
+/client/verb/webmap()
+	set name = "webmap"
+	set hidden = TRUE
+
+	if(!SSmapping.map_datum.webmap_url)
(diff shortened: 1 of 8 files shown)


code/game/verbs/webmap.dm
```diff
@@ -0,0 +1,12 @@
+/client/verb/webmap()
```
SabreML, line 1: Should this go into `interface\interface.dm` instead of its own file? The verbs for the other buttons all seem to be in there.

AffectedArc07: Probably

variableundefined (approved):

PR #16766, opened 2021-09-22
Allows rotating booze and soda dispensers with alt click

S34NW:
## What Does This PR Do
This PR lets you rotate unanchored drink dispenser machinery with alt-click.

## Why It's Good For The Game
This was not an issue before the new sprites as there was only one dir.
You currently have to do a funny dance to spin it that requires space you might not have, this is more intuitive and simple.

## Changelog
:cl:
tweak: Alt-clicking unanchored drinks dispensers now rotates them.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
@@ -319,6 +319,21 @@
 		return
 	ui_interact(user)
 
+/obj/machinery/chem_dispenser/AltClick(mob/user)
+	if(!is_drink)
+		return
+	if(user.incapacitated())
+		to_chat(user, "<span class='warning'>You can't do that right now!</span>")
+		return
+	if(!Adjacent(user))
+		return
+	if(anchored)
+		to_chat(user, "<span class='warning'>[src] is anchored to the floor!</span>")
+		return
+	pixel_x = 0
+	pixel_y = 0
+	setDir(turn(dir, 90))
+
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chemistry/machinery/chem_dispenser.dm
```diff
@@ -319,6 +319,21 @@
...
+	if(!is_drink)
+		return
+	if(user.incapacitated())
+		to_chat(user, "<span class='warning'>You can't do that right now!</span>")
+		return
+	if(!Adjacent(user))
+		return
+	if(anchored)
+		to_chat(user, "<span class='warning'>[src] is anchored to the floor!</span>")
+		return
```
SabreML, line 332: The `is_drink` and `Adjacent()` checks can be merged together to save a couple of lines if you want, since they're both just returning:
```suggestion
/obj/machinery/chem_dispenser/AltClick(mob/user)
	if(!is_drink || !Adjacent(user))
		return
	if(user.incapacitated())
		to_chat(user, "<span class='warning'>You can't do that right now!</span>")
		return
	if(anchored)
		to_chat(user, "<span class='warning'>[src] is anchored to the floor!</span>")
		return
```
It also stops the "You can't do that right now!" message showing even if you're not adjacent to it, since that check will come first.

S34NW: While I see your points, I don't agree since that means regular chem dispensers will ALWAYS give this message.

variableundefined (approved):

PR #16771, opened 2021-09-22
Adds Epinephrine to the Gutlunch Mob

### LINKED ISSUE #16360: No breath trait on ashwalkers conflicting with newcrit, paralyising them forever without finishing them off

Exavere:
## What Does This PR Do

Attempts to Fix #16360.   Adds the reagent Epinephrine to the Gutlunch Mob's udder, so that Ash Walkers are able to avoid getting screwed over permanently from just being near Critical for a short time.  

## Why It's Good For The Game

Currently, when an Ash Walker gets near Critical they begin to go through Cardiac Failure into Cardiac Arrest with no way of breaking out of Failure, even if they manage to heal back up to full from Critical - once it starts it doesn't stop.  Essentially, Cardiac Failure for Ash Walkers is a game ender and ruins much of their ability to play, the Trait_NOBREATH prevents oxyloss so they never accumulate the necessary oxy damage to die while in Cardiac Arrest.  

Once they enter Cardiac Arrest they just remain stunlocked on the ground for eternity auto screaming until they're finished off by external means, forcing them to eat the 5 minute respawn time and wasting resources which they don't have very much of, or suiciding.   Letting them milk Epinephrine from Gutlunch Mobs will allow them to counter this impending Cardiac Arrest Purgatory, if they're able to get it them quick enough.  That said, if they do enter full Cardiac Arrest then they're still screwed without a Defib--this at least gives them the ability to try to prevent themselves from entering Cardiac Arrest Purgatory.  

This PR also gives the Gutlunch Mob itself a much better use, it becomes more important to Ash Walkers - love your Gutlunch Ash Walkers.

## Changelog
:cl: Exavere
add: Adds Epinephrine to Gutlunches.
fix: Part of a Fix related to #16360.  
tweak: Tweaks the reagents of the Gutlunch Mob.  
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining/gutlunch.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining/gutlunch.dm
@@ -1,7 +1,7 @@
 //Gutlunches, passive mods that devour blood and gibs
 /mob/living/simple_animal/hostile/asteroid/gutlunch
 	name = "gutlunch"
-	desc = "A scavenger that eats raw meat, often found alongside ash walkers. Produces a thick, nutritious milk."
+	desc = "A scavenger that eats raw meat, often found alongside ash walkers. Produces a thick, medicinally nutritious milk."
 	icon = 'icons/mob/lavaland/lavaland_monsters.dmi'
 	icon_state = "gutlunch"
 	icon_living = "gutlunch"
@@ -104,7 +104,9 @@
 	if(prob(60))
 		reagents.add_reagent("cream", rand(2, 5))
 	if(prob(45))
-		reagents.add_reagent("salglu_solution", rand(2,5))
+		reagents.add_reagent("salglu_solution", rand(2, 5))
+	if(prob(30))
+		reagents.add_reagent("epinephrine", rand(2, 5))
 
 
 //Male gutlunch. They're smaller and more colorful!


Exavere:
Thank you @matttheficus for this idea, heckin' nerd.

variableundefined (approved):

PR #16775, opened 2021-09-22
Nerfs (S)laughter demons

Qwertytoforty:
## What Does This PR Do

Adds a new subtype of slaughter demons used for midrounds, with a max HP of 130, compared to the usual 200.
## Why It's Good For The Game

Demons are strong. With a competent demon, it is extremely hard to kill without extreme preventative cleaning to prevent it from attacking, voice cleaner grenades to instantly clean all blood so it can't escape, or shotguns with slugs to melt it before it can react to leave. Otherwise, due it's large amount of HP, full heal on kill, and high mobility with blood crawl, it almost never gets killed. 

A nerf is needed, to keep them from steamrolling the station each time one shows up.

Thus, this PR lowers the HP, to allow crew to have an easier time killing it, without having to worry anymore about demons getting 2-3 shot by a shotgun. Now, it takes 7 lasers / wt550 shots, instead of 10, making it easier to kill a demon if it has not ate in a while. I did not however want to lower laughter demons health too much below what slaughter demons have, thus I lowered laughter demon damage from 25 to 20, so they now deal 2/3rds the amount of damage slaughters do.

For those wondering why I did not lower the damage to 20 like TG did and not leave health, it would be because demons on TG have wound modifiers after multiple hits, which is not applicable here, and because a slaughter demon taking 10 hits to get someone to -100 with no armor didn't quite feel the best, especially when dealing with things like janiborgs or crew with welders.

## Changelog
:cl:
tweak: Midround slaughter demon health lowered to 130 from 200
/:cl:

### DIFF
--- a/code/modules/events/slaughterevent.dm
+++ b/code/modules/events/slaughterevent.dm
@@ -36,7 +36,7 @@
 		kill()
 		return
 	var/obj/effect/dummy/slaughter/holder = new /obj/effect/dummy/slaughter(pick(spawn_locs))
-	var/mob/living/simple_animal/slaughter/S = new /mob/living/simple_animal/slaughter/(holder)
+	var/mob/living/simple_animal/slaughter/lesser/S = new /mob/living/simple_animal/slaughter/lesser/(holder)
 	S.holder = holder
 	player_mind.transfer_to(S)
 	player_mind.assigned_role = "Slaughter Demon"
--- a/code/modules/events/slaughterevent.dm
+++ b/code/modules/events/slaughterevent.dm
@@ -1,5 +1,6 @@
 /datum/event/spawn_slaughter
 	var/key_of_slaughter
+	var/demon = /mob/living/simple_animal/slaughter/lesser
 
 /datum/event/spawn_slaughter/proc/get_slaughter()
 	var/list/candidates = SSghost_spawns.poll_candidates("Do you want to play as a slaughter demon?", ROLE_DEMON, TRUE, source = /mob/living/simple_animal/slaughter)
@@ -36,7 +37,7 @@
 		kill()
 		return
 	var/obj/effect/dummy/slaughter/holder = new /obj/effect/dummy/slaughter(pick(spawn_locs))
-	var/mob/living/simple_animal/slaughter/lesser/S = new /mob/living/simple_animal/slaughter/lesser/(holder)
+	var/mob/living/simple_animal/slaughter/S = new demon(holder)
 	S.holder = holder
 	player_mind.transfer_to(S)
 	player_mind.assigned_role = "Slaughter Demon"
@@ -46,3 +47,6 @@
 
 /datum/event/spawn_slaughter/start()
 	INVOKE_ASYNC(src, .proc/get_slaughter)
+
+/datum/event/spawn_slaughter/greater
+	demon = /mob/living/simple_animal/slaughter
--- a/code/game/gamemodes/miniantags/slaughter/slaughter.dm
+++ b/code/game/gamemodes/miniantags/slaughter/slaughter.dm
@@ -26,8 +26,8 @@
 	maxbodytemp = INFINITY
 	faction = list("slaughter")
 	attacktext = "wildly tears into"
-	maxHealth = 200
-	health = 200
+	maxHealth = 130
+	health = 130
 	environment_smash = 1
 	//universal_understand = 1
 	obj_damage = 50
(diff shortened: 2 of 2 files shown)


hal9000PR:
ezgif com-gif-maker

Fox-McCloud:
Not a fan of this at all. The crew still wins quite handedly, even during Ragin' mages.

Even so, with regular wizard, it's 1 vs the entire crew---the win-rate of wizards, even with multiple demon purchases is *abysmal*.

Game modes like this are _**NOT**_ balanced around 1 vs 1 interaction--it's the larger conflict that matters.

Aquilar:
Because the crew doesn't need to kill the demons, they just need to kill the wizard.

Not really an issue of balance any more it's more of an issue some players are too good for their own good and it drags everyone else down the nerf hole with them.

Rurik123:
Id prefer a rework entirely so a "slaughter" demon isn't entirely dependent on kiting and avoiding fights against anyone with a gun, but this is preferable to outright removal that I was expecting.

Laughter demon dying in 3 stabs to a miner katana will be funny though.

Spacemanspark:
>The crew still wins quite handedly, even during Ragin' mages.

Don't demons summoned by wizards get objectives to kill the wizard who summoned them? If so, then this is a fairly weak argument. While some demons may opt not to follow this objective (in which case, why have it?) I know I've definitely seen demons obliterate wizards and end rounds exceedingly quickly.

ppi13:
This is to harsh of a nerf, and even without shotgun, you still got miners who can basically be immune, and as long as you avoid the first hit, you can easily get away.  
Maybe wait and see how s/laughter demons handle after the shotgun nerf first?

Denthamos:
So when testing with slaughter and laughter demons during the shotgun removal TM, I ran both wizard and other game modes.  Wizard is meant to be chaos rains from above.  

The issue I noticed with the slaughter demon comes in as a mid round antagonist late into the round when the rounds antagonists are going full in on their roles.  This creates a large window of confusion and chaos for the mid round demon with which to abuse due to security being highly busy at that point and no source of burst damage with which to kill a demon before they regenerate their health after consuming someone in a blood pool.  Combined with the demon not having the objective to kill their summoner like wizard game mode, is where this becomes an issue.

So I am against a flat out nerf to demons like this, however I do believe the MAJOR EVENT demon needs a rebalance due to the shotgun removal.  I propose that a new subtype of the slaughter demon gets created that the event spawns and _that one_ get the nerf mentioned above.

Samman166:
Pulse slugs also exist still and have become very popular, while not as good as old slugs still gives good burst damage. Also I've got to agree with Rurik that a rework would be nice but that seems unreasonable for such a rare antag anyways

SweaterKittens:
> So when testing with slaughter and laughter demons during the shotgun removal TM, I ran both wizard and other game modes. Wizard is meant to be chaos rains from above.
> 
> The issue I noticed with the slaughter demon comes in as a mid round antagonist late into the round when the rounds antagonists are going full in on their roles. This creates a large window of confusion and chaos for the mid round demon with which to abuse due to security being highly busy at that point and no source of burst damage with which to kill a demon before they regenerate their health after consuming someone in a blood pool. Combined with the demon not having the objective to kill their summoner like wizard game mode, is where this becomes an issue.
> 
> So I am somewhat against a flat out nerf to demons like this, however I do believe the MAJOR EVENT demon needs a rebalance due to the shotgun removal. I propose that a new subtype of the slaughter demon gets created that the event spawns and _that one_ get the nerf mentioned above.
> 
> If wizard demon is that big of an issue then it can get said nerf as well - and during ragin mages well. Its raging mages, literally the adminbus of adminbus rounds where you are expecting the station to end up a bleeding burning mess.

I just want to second what Denthamos is saying here. I think discussing demon balance in terms of wizard rounds, where balance is already basically out the window, is not as important as how absolutely _awful_ these things are as a midround. With someone even mildly robust at the helm, they will effectively permakill people in the double digits and never die. Having both types of demon receive different balance might be worthwhile.

code/modules/events/slaughterevent.dm
```diff
@@ -36,7 +36,7 @@
 		kill()
 		return
 	var/obj/effect/dummy/slaughter/holder = new /obj/effect/dummy/slaughter(pick(spawn_locs))
-	var/mob/living/simple_animal/slaughter/S = new /mob/living/simple_animal/slaughter/(holder)
+	var/mob/living/simple_animal/slaughter/lesser/S = new /mob/living/simple_animal/slaughter/lesser/(holder)
```
Fox-McCloud, line 39: Please make two of this event that's slaughter lesser and slaughter greater.

It'd be nice to have a generic event that just pulls a random person in to be slaughter demon--sometimes its useful for events and ragin' mages shenanigans.

Qwertytoforty: Done!

PR #16784, opened 2021-09-23
[WIP] CQC Rework

matttheficus:
## What Does This PR Do
Replaces the current CQC with a reworked and remastered variant more focused on Traitor-use than Nukie-use.

### **Changes Some Base CQC Concepts**
- [x] Removes the rather unused Pressure Combo
- [x] Removes the Kick Combo, removing the insta-sleep from kicking someone on the floor
- [x] Changes Legsweep to a trip, adds stamina damage. This makes it a wonderful tool to run away rather than as an opener into a perma-stun `Weaken: 3 > 1`
- [x] Reduces the bonus damage of CQC slightly to foster a more non-lethal and stealthy playstyle `Consecutive CQC: Brute:25 > 20, Stamina: 50 > 60`, `CQC Bonus Melee: 13 > 10, Bonus Weakened: 5 > 3`
- [x] Makes neck-grab stuns less than the infinite amount of time they basically are now (`weaken(5) > weaken(3)`)
### **Introduces Restraint Combos** 
- [ ] Apply Disarm: stutters, mutes, then sleeps your victim after a variable delay
- [ ] Apply Grab: Allows you to restrain someone and move at full speed, allowing you to reposition your victim without the immediate need for restraints
- [ ] Apply Harm: Struggle with your victim for a longer period but choke them out in a more lethal manner than a normal “kill” grab. Unsure if I would make this a faster loss-breath death, or make it more like a neck snap, with a cling-style progress bar then a “crunch, death, loud noise” idea.
Another idea, though one that I’m unsure of how to implement, would be making it so you could “kill” using some sort of weapon on your person. Have a combat knife? Glass shard? Scalpel? Pencil? Each one could do a different thing, but I would need to figure out how to make it possible to use an open hand or use something in one’s pockets/etc.
- [ ] Apply Help: Mutes the victim, but keeps them conscious for interrogation, RP, etc etc.

### Ideas Currently Being Thought Of
These were mostly thought up in #codingchat while discussing possible ideas to do
Thanks to: @SteelSlayer @dearmochi @farie82 and Gummy (dunno your git whoops!)

- Make grabs not a MASSIVE stun.

- Make Sleep-Restrain tier-based: first application mute, second sleep. Maybe garble (think radio jammer), mute, sleep.

- Make the kill-grab scale timewise or damage wise with the weapon you use to apply. Still needs to be loud and let the victim scream/yell on comms/etc. Look at safety scissors code.

## Why It's Good For The Game
As I see it now, the current Martial Arts (at least from a basic Traitor's perspective) fall sort of close to each other in terms of offensive capabilities while their only big difference is their respective utility. That being said, CarpFu has a very distinct gimmick and power: you lose the ability to fight with ranged weapons, but can now completely deflect them.

CQC on the other hand might SOUND like the Metal Gear Solid sneaky-tactics powerhouse, but in reality it is more akin to a 90's Action Kung-Fu movie where you can run in, slap a bunch of people to the ground, and then knock them all out with a swift punch to the head. While this has its uses in gamemodes such as NukeOps, its rather strong from the average Traitor if used well. It also has a myriad of insta-stuns that aren't fun for the victim in any way shape or form.

This PR looks to shift CQC away from a NukeOps focus and more to being sneaky and evasive. It also gives CQC its own gimmick: restraints. That way you power is more grounded on 1v1 capabilities augmented by evasive and stamina-draining attacks to escape from multiple attackers. The idea allows you to follow your victim, attack and disable them quickly, but leaves you vulnerable if a passerby or Security shows up.

I go a bit more in-depth below if you wish for a proper design document with some notes about each idea that is in the current iteration of the combos I'm laying out.

**[CQC Thoughts.docx](https://github.com/ParadiseSS13/Paradise/files/7221233/CQC.Thoughts.docx)**

## Changelog
:cl:

add: Restraint Combos.
del: removed Pressure and CQC Kick.
tweak: made grabs weaken for 3 cycles rather than 5.
tweak: reduces CQC's bonus damage and replace it with stamina damage where applicable.
tweak: reworks legsweep by massively reducing its stun time, converting it to a trip.

/:cl:


code/modules/martial_arts/cqc.dm
```diff
@@ -51,18 +66,31 @@
...
+					D.Stuttering(4)
+				if(2)
+					to_chat(A, "<span class='notice'You notice [D]'s struggles becoming weaker!</span>")
+					D.visible_message("<span class='warning'>[A] drags [D] down to the ground!</span>")
+					D.AdjustSilence(5)
+				if(3)
+					to_chat(A, "<span class='notice'>You drag [D]'s limp body to the ground!</span>")
+					D.visible_message("<span class='danger'>[A] knocks out [D]!</span>")
+					D.AdjustSleeping(15)
+			if(!do_mob(A, D, 30))
```
hal9000PR, line 91: Use the seconds define

code/modules/mob/mob_grab.dm
```diff
@@ -178,7 +178,7 @@
 	if(state >= GRAB_KILL)
 		//affecting.apply_effect(STUTTER, 5) //would do this, but affecting isn't declared as mob/living for some stupid reason.
 		affecting.Stuttering(5) //It will hamper your voice, being choked and all.
-		affecting.Weaken(5)	//Should keep you down unless you get help.
+		affecting.Weaken(3)	//Should keep you down unless you get help.
```
hal9000PR, line 181: You also need to reduce the stun on red grabs else this is pretty pointless.

Wojz:
As far as harm is concerned I think it should be more akin to a progress bar limb snap. Like how they love to do in Metal Gear Solid. If we want throat slitting etc. I see that perhaps one day being added just as a regular thing.

This is really cool.

Denthamos:
CQC was never meant to be a regular traitor martial art in the first place.  It was built around being for nuclear operatives but for some reason it got merged into the normal traitors arsenal?  Of course its strong.  Honestly it should not be there for traitors in the first place and not liking the fact that this also nerfs the nuclear operatives version as well.

PR #16790, opened 2021-09-24
Makes uncuttable chainlink fences deconstructable

S34NW:
## What Does This PR Do
This PR makes chainlink fences that don't support hole cutting (posts, ends, doors, corners) dismantleable.
This process takes the same time as cutting three holes in a regular fence, just in one step that deletes the fence.

## Why It's Good For The Game
Currently, if you place a chainlink fence post, end, door, or corner you are screwed and have to bash it with an object until it breaks.
This is not ideal, as it makes fences stronger than walls in some ways.

With this change, even though you can't cut holes in them, you can spend the extra time to take them apart, as opposed to breaking it with a toolbox or whatever you have on hand.

## Changelog
:cl:
tweak: Non-cuttable chainlink fences can now be taken apart with wire cutters after a delay.
/:cl:

### DIFF
--- a/code/game/objects/structures/fence.dm
+++ b/code/game/objects/structures/fence.dm
@@ -95,7 +95,7 @@
 	if(!cuttable)
 		user.visible_message("<span class='warning'>[user] starts dismantling [src] with [W].</span>",\
 		"<span class='warning'>You start dismantling [src] with [W].</span>")
-		if(W.use_tool(src, user, FULL_CUT_TIME * W.toolspeed, volume = W.tool_volume))
+		if(W.use_tool(src, user, FULL_CUT_TIME, volume = W.tool_volume))
 			user.visible_message("<span class='notice'>[user] completely dismantles [src].</span>",\
 			"<span class='info'>You completely dismantle [src].</span>")
 			qdel(src)
--- a/code/game/objects/structures/fence.dm
+++ b/code/game/objects/structures/fence.dm
@@ -94,7 +94,7 @@
 			visible_message("<span class='notice'>\The [user] completely dismantles \the [src].</span>")
 			to_chat(user, "<span class='info'>You completely take apart \the [src].</span>")
 			qdel(src)
-			return
+		return
 	if(invulnerable)
 		to_chat(user, "<span class='warning'>This fence is too strong to cut through!</span>")
 		return
--- a/code/game/objects/structures/fence.dm
+++ b/code/game/objects/structures/fence.dm
@@ -93,30 +93,30 @@
 		to_chat(user, "<span class='warning'>This fence is too strong to cut through!</span>")
 		return
 	if(!cuttable)
-		visible_message("<span class='danger'>[user] starts dismantling [src] with [W].</span>")
-		to_chat(user, "<span class='danger'>You start dismantling [src] with [W].</span>")
+		user.visible_message("<span class='warning'>[user] starts dismantling [src] with [W].</span>",\
+		"<span class='warning'>You start dismantling [src] with [W].</span>")
 		if(W.use_tool(src, user, FULL_CUT_TIME * W.toolspeed, volume = W.tool_volume))
-			visible_message("<span class='notice'>[user] completely dismantles [src].</span>")
-			to_chat(user, "<span class='info'>You completely dismantle [src].</span>")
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/fence.dm
```diff
@@ -89,7 +90,10 @@
 	if(shock(user, 100))
 		return
 	if(!cuttable)
```
farie82, line 92: I'd move this check below the `invulnerable` check. Else you can remove the corners of an `invulnerable` fence

code/game/objects/structures/fence.dm
```diff
@@ -89,7 +90,10 @@
...
-		to_chat(user, "<span class='warning'>This section of the fence can't be cut!</span>")
+		if(W.use_tool(src, user, FULL_CUT_TIME * W.toolspeed, volume = W.tool_volume))
+			visible_message("<span class='notice'>\The [user] completely dismantles \the [src].</span>")
```
farie82, line 94:
```suggestion
			visible_message("<span class='notice'>[user] completely dismantles \the [src].</span>")
```

code/game/objects/structures/fence.dm
```diff
@@ -89,7 +90,10 @@
 	if(shock(user, 100))
 		return
 	if(!cuttable)
-		to_chat(user, "<span class='warning'>This section of the fence can't be cut!</span>")
+		if(W.use_tool(src, user, FULL_CUT_TIME * W.toolspeed, volume = W.tool_volume))
```
farie82, line 93: Might want to give feedback when starting in the form of text. Probably a `visible_message` for others as well

farie82 (changes requested):

code/game/objects/structures/fence.dm
```diff
@@ -89,15 +89,17 @@
 	. = TRUE
 	if(shock(user, 100))
 		return
+	if(invulnerable)
+		to_chat(user, "<span class='warning'>This fence is too strong to cut through!</span>")
+		return
 	if(!cuttable)
+		visible_message("<span class='danger'>[user] starts dismantling [src] with [W].</span>")
+		to_chat(user, "<span class='danger'>You start dismantling [src] with [W].</span>")
```
farie82, line 97: Could combine this into one `visible_message` send from the user with a self message.
I feel like `danger` is a bit overkill for what you're doing. `warning` is more suited I'd say.
I personally see `danger` as something that might or has hurt somebody other than the user directly. `warning` is something that might potentially be bad but also has a chance to be okay and just needs a bit more attention.

code/game/objects/structures/fence.dm
```diff
@@ -89,15 +89,17 @@
...
 		if(W.use_tool(src, user, FULL_CUT_TIME * W.toolspeed, volume = W.tool_volume))
-			visible_message("<span class='notice'>\The [user] completely dismantles \the [src].</span>")
-			to_chat(user, "<span class='info'>You completely take apart \the [src].</span>")
+			visible_message("<span class='notice'>[user] completely dismantles [src].</span>")
+			to_chat(user, "<span class='info'>You completely dismantle [src].</span>")
```
farie82, line 100: Same here with combining them into one message

hal9000PR:
Does it drop materials now when you dismantle them?

code/game/objects/structures/fence.dm
```diff
@@ -88,29 +89,34 @@
...
-	if(!cuttable)
-		to_chat(user, "<span class='warning'>This section of the fence can't be cut!</span>")
-		return
 	if(invulnerable)
 		to_chat(user, "<span class='warning'>This fence is too strong to cut through!</span>")
 		return
+	if(!cuttable)
+		user.visible_message("<span class='warning'>[user] starts dismantling [src] with [W].</span>",\
+		"<span class='warning'>You start dismantling [src] with [W].</span>")
+		if(W.use_tool(src, user, FULL_CUT_TIME * W.toolspeed, volume = W.tool_volume))
```
farie82, line 98:
```suggestion
		if(W.use_tool(src, user, FULL_CUT_TIME, volume = W.tool_volume))
```
The toolspeed is already processed in the proc itself

farie82 (approved):

variableundefined (approved):

PR #16793, opened 2021-09-24
Remove ways to dupe materials. Also Alien Alloy fix

botka4aet:
## What Does This PR Do
- Abductor closet cost only 1 Alien Alloy but if you disassemble it - you'll get 2. Fixed.
- 50 Plasteel from cargo cost 50. If you melt them and pick raw - you will get 50 plasma. If you send it back - you'll get 250 cargo points. Infinity dupe? Fixed.
- /obj/item/stack already have origin_tech = "materials=1" so children don't need it ?
- You can create alien alloy for 4000 metal and plasma but can't smelting back? Now it's 2000/2000

## Why It's Good For The Game
Remove ways to dupe something

## Images of changes
image

## Changelog
:cl:
add: New plasteel with low plasma
tweak: Plasteel from cargo with low plasma
fix: Abductor closet drop only 1 Alien Alloy
fix: Now alien alloy cost something
/:cl:

PS. You can build same objects from lowplasma plasteel. And if you disassemble - you will get raw plasteel. Can solve it even with new obj var.

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -1352,14 +1352,14 @@ GLOBAL_LIST_INIT(all_supply_groups, list(SUPPLY_EMERGENCY,SUPPLY_SECURITY,SUPPLY
 
 /datum/supply_packs/materials/plasteel20
 	name = "20 Plasteel Sheets Crate"
-	contains = list(/obj/item/stack/sheet/plasteel)
+	contains = list(/obj/item/stack/sheet/plasteel/lowplasma)
 	amount = 20
 	cost = 30
 	containername = "plasteel sheets crate"
 
 /datum/supply_packs/materials/plasteel50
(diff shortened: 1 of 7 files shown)


TDSSS:
It might be better to just increase plasteel costs rather than making some snowflake lowplasma plasteel

botka4aet:
> It might be better to just increase plasteel costs rather than making some snowflake lowplasma plasteel

Well, i don't think about it.
But will someone buy 50 plasteel for 260 points?

Denthamos:
Hrm, while snowflaking is bad as a QM main I gotta agree, I would _never_ buy a stack of plasteel for that amount of points unless for some reason I have 10K plus from a magical round where botany and R&G both give stuff for points and miners bring up all the minerals.  Even then I would never order more than 1-3?

PR #16796, opened 2021-09-24
Update delta.dmm

PseudOtis:
## What Does This PR Do
Revamps Medical Lobby (Deltastation)

## Why It's Good For The Game
People have complained that delta is too big and empty. I think this is an easy way to fix some of that without a complete map redesign. Plus it looks nice.

## Images of changes
medbay change

## Changelog
:cl:
tweak: Added more furniture to medbay lobby.
:cl:

### DIFF
--- a/_maps/map_files/Delta/delta.dmm
+++ b/_maps/map_files/Delta/delta.dmm
@@ -46951,6 +46951,7 @@
 /obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{
 	dir = 4
 	},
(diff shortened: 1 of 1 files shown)


S34NW:
I like this a lot, one thing i'd suggest is trimming your changelog to be just something like:
`tweak: Added more furniture to delta medbay lobby`

LightFire53:
Looks nice! Would also maybe change PR title to something closer to what's actually going on. IE: Revamps Medical Lobby (DeltaStation)

PR #16798, opened 2021-09-24
Blood overlay fix for telescopic weapons

### LINKED ISSUE #16676: Bloody Telescopic Batons are Unwashable
**Issue Description**:
When you try to clean a bloody telescopic baton (with sink, shower, space cleaner, and soap), the blood stays on.

I expected the blood to come off after using any of the cleaning methods, similar to other items.

**What happened instead**:
The blood stayed on.

I like to clean my items whenever I get a chance and this just ruins that. Also inconsistent behavior is bad.

1. Hit something with blood (I tested it on humans, monkeys, skrell, and kidan) until baton is covered in blood
2. Try to wash it, open or closed, it won't clean.

I think I used to wash telescopic batons a few months ago, but really not sure.

**Extra information**:
One time the blood stains would be in the extended position, even if the baton is closed, so you get floating blood in the closed baton sprite.

SabreML:
## What Does This PR Do

Fixes #16676 by converting the blood overlay system to use `SSoverlays` rather than directly adding and removing.
The issue seems to have been caused by the blood overlay being added by `add_overlay()`, but removed with `overlays -= blood_overlay` rather than the corresponding `cut_overlay()` proc.

Also fixes the blood overlay on telescopic scythes disappearing when the scythe gets extended or retracted.
I'm not very happy with the solution of copying the code from telescopic batons, but I couldn't find a better way to do it without a full refactor.

Also also removes the `blood_overlay_color` variable, because a mostly redundant variable only used in a single file really shouldn't be defined on `/obj/item`.

## Why It's Good For The Game

Bugfix! (Consistent overlays are good.)

## Changelog
:cl:
fix: Fixed the blood overlay on telescopic batons being impossible to remove.
fix: Fixed the blood overlay on telescopic scythes disappearing when its mode gets switched.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -789,15 +789,15 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 		var/index = blood_splatter_index()
 		var/icon/blood_splatter_icon = GLOB.blood_splatter_icons[index]
 		if(!blood_splatter_icon)
-			blood_splatter_icon = icon(initial(icon), initial(icon_state), , 1)		//we only want to apply blood-splatters to the initial icon_state for each object
-			blood_splatter_icon.Blend("#fff", ICON_ADD) 			//fills the icon_state with white (except where it's transparent)
+			blood_splatter_icon = icon(initial(icon), initial(icon_state), frame = 1)		//we only want to apply blood-splatters to the initial icon_state for each object
+			blood_splatter_icon.Blend("#ffffff", ICON_ADD) 			//fills the icon_state with white (except where it's transparent)
 			blood_splatter_icon.Blend(icon('icons/effects/blood.dmi', "itemblood"), ICON_MULTIPLY) //adds blood and the remaining white areas become transparant
 			blood_splatter_icon = fcopy_rsc(blood_splatter_icon)
 			GLOB.blood_splatter_icons[index] = blood_splatter_icon
 
 		blood_overlay = image(blood_splatter_icon)
 		blood_overlay.color = color
-		overlays += blood_overlay
+		add_overlay(blood_overlay)
 
 /atom/proc/clean_blood(radiation_clean = FALSE)
 	germ_level = 0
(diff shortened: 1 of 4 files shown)


code/game/atoms.dm
```diff
@@ -789,15 +789,15 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
 		var/index = blood_splatter_index()
 		var/icon/blood_splatter_icon = GLOB.blood_splatter_icons[index]
 		if(!blood_splatter_icon)
-			blood_splatter_icon = icon(initial(icon), initial(icon_state), , 1)		//we only want to apply blood-splatters to the initial icon_state for each object
-			blood_splatter_icon.Blend("#fff", ICON_ADD) 			//fills the icon_state with white (except where it's transparent)
+			blood_splatter_icon = icon(initial(icon), initial(icon_state), frame = 1)		//we only want to apply blood-splatters to the initial icon_state for each object
```
SteelSlayer, line 792: I was experimenting to see if there was a better way to handle the dupe code in `add_blood_overlay`. I found that if the `initial`s are removed here, then the baton and scythe `add_blood_overlay` procs are removed so they instead use this one, blood overlays still work fine for objects. I'm not really sure why the `initial`s are here in the first place, maybe I'm missing something.

It might break for items which can transform their icon state, like the baton, so code would need to be added for those. (I can't actually think of any other item besides the baton and scythe right now). But anyway, just an idea to play around with, code works fine as is.

SabreML: I had the exact same thoughts actually. I tried removing the `initial()`s entirely from the parent proc and just using that, but like you said it caused problems with other 'transforming' items like energy swords and riot shields:
image

There's more items with just a single sprite than multiple of course, and those could be tweaked individually, but I figured it would be better to fix this bug in its own PR rather than bundling it into a small refactor which would probably need a testmerge, considering the whole fix bundle thing.

PR #16799, opened 2021-09-24
Ports Cloth Golems

Fox-McCloud:
Ports Cloth Golems, from TG. Re-upload of: https://github.com/ParadiseSS13/Paradise/pull/13639 something went horrifyingly wrong with that branch and I don't want to risk it.

Cloth golems are a new type of ...well, golem, that can be made by adding cloth or gauze to a construct shell.

They have significantly less armor than regular golems (They only take 15% less damage) and are incredibly weak to burn damage---they also have no resistance to heat, either, like regular golems---additionally, they're always flammable.

In exchange they're a bit faster than regular golems....andddd---they revive on non-gibbing death.

If you die, you collapse into a pile of cloth---after 90 seconds you'll fully regenerate. If you're on fire when you die, however, you'll just turn into dust. The pile of cloth that is left behind can be hit with something that's hot and it will set it on fire--once it burns, it'll prevent the golem from reviving.

img

Also increases the max amount of gauze you can stack together to 12 (starting value is still 6)

:cl: Fox McCloud
add: Adds cloth golems
tweak: increases the max amount of gauze you can stack together from 6 to 12
/:cl:


Samman166:
How are terrors spiders and non burning oil blobs going to deal with them? They don't have any burn attacks that I can think of to destroy them with

variableundefined (approved):
Previous PR already approved.

Fox-McCloud:
>How are terrors spiders and non burning oil blobs going to deal with them? They don't have any burn attacks that I can think of to destroy them with

Blobs can easily delete teh cloth pile by expanding over top of it. 

Other mobs can still melee the cloth pile till it crumbles.

Samman166:
> 
> Blobs can easily delete teh cloth pile by expanding over top of it.
> 
> Other mobs can still melee the cloth pile till it crumbles.

Then it sounds like a cool addition, thanks!

PR #16811, opened 2021-09-26
Remaps MetaStation Escape

LightFire53:
## What Does This PR Do
 This remaps MetaStation's Escape/departures with the goal of having the shuttle be at normal docking orientation compared to being sideways. As a result, it extends departures by quite a bit, so I've had to shimmy a solar array by 90 degrees, move xenobiology so that there is space between it and the shuttle dock, added a podbay to escape for mechanic/sec pod, and added a bar that is "under construction" to fill some dead space.

## Why It's Good For The Game
 Having the shuttle dock in the same way as Delta and Box is nice not only for parity, but makes it much less jarring when the shuttle departs/docks at CC

## Images of changes

Wayyyyy too much going on here, look at Map Diff Bot for changes.

## Changelog
:cl:
tweak: Remaps MetaStation Escape
/:cl:

### DIFF
--- a/_maps/map_files/MetaStation/MetaStation.dmm
+++ b/_maps/map_files/MetaStation/MetaStation.dmm
@@ -26238,6 +26238,29 @@
 	icon_state = "neutralcorner"
 	},
 /area/hallway/primary/central)
+"bfL" = (
+/obj/effect/spawner/window/reinforced,
+/obj/structure/cable/yellow,
+/obj/structure/cable/yellow{
+	d2 = 8;
+	icon_state = "0-8"
+	},
+/obj/structure/cable/yellow{
+	d2 = 4;
+	icon_state = "0-4"
+	},
+/obj/machinery/atmospherics/pipe/simple/visible,
+/obj/machinery/door/poddoor{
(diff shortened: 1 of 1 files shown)


S34NW (commented):
I'm a bit on the fence about this, the mapping is issue-free and looks pleasing aesthetically, but the whole purpose of the remap seems a bit flimsy. The shuttle rotating happens once per shift, and is already jarring with the screen shake and such. I personally feel that adding a big open room like this is just gonna contribute to active turfs and atmos lag.

Fox-McCloud:
Don't really care for this---it has a pretty glued on look.

Meta escape is already pretty large.

I also don't see the point of it having to be vertically docked---having increased variety in maps around elements that are largely aesthetic is a good thing.

LightFire53:
It's beginning to look like I'm the only person who thinks shuttle flipping is weird/jarring, which is completely fine. More than happy to leave meta escape alone. May still try to throw a podbay into the mix in the future.

AffectedArc07:
Gonna close this. 2 maints against, and pods are now gone.

PR #16812, opened 2021-09-26
refactors butchering of humans with chainsaw/meatcleaver

hal9000PR:
## What Does This PR Do
makes butchering humans with a syndie chainsaw/meatcleaver use traits instead of a direct istype check

examining these items will also show what they do.

## Why It's Good For The Game
Traits are good. istype checks are bad, makes it easier to tell what is happening with the code.

also, bonus adminbus

## Images of changes
image

(humans was changed to people)
## Changelog
🆑 
add: admins can now adminbus any item to butcher dead corpses.
/:cl:


hal9000PR:
I am incapable of writing.

SabreML (approved):

code/game/objects/items.dm
```diff
@@ -230,6 +230,9 @@ GLOBAL_DATUM_INIT(fire_overlay, /image, image("icon" = 'icons/goonstation/effect
 		msg += "*--------*"
 		. += msg
 
+	if(HAS_TRAIT(src, TRAIT_BUTCHERS_HUMANS))
+		. += "<span class='warning'> Can be used to butcher dead humans into meat while on harm intent. </span>"
```
SteelSlayer, line 234:
```suggestion
		. += "<span class='warning'>Can be used to butcher dead humans into meat while on harm intent.</span>"
```

SteelSlayer (approved):
LGTM other than the nitpick.

Wojz:
Would be probably better to say "people" instead of "humans" since most of these are only "humans" in code.

dearmochi (approved):

PR #16824, opened 2021-09-28
TG Sprite Porting: BURGERS Edition

PacifistDalek:
## What Does This PR Do
Ports TG burger sprites
(also removes one unused variant roburger that was last referenced in a PR from 2011 :P)

## Images of changes
image

## Changelog
:cl:
imageadd: TG Burger Sprites
del: Removed unused Big Roburger
/:cl:


botka4aet:
> one unused variant roburger
You can get it from silver slime, so it's not like 'unused'

PacifistDalek:
> 
> 
> > one unused variant roburger
> 
> You can get it from silver slime, so it's not like 'unused'

Ah, ok! I couldn't find a place it was used but that makes sense. That said, i'm not sure it's worth still keeping around, as it basically just has more nanites, and it's not that hard to make nanites. I'd rather add something more unique and weird to the silver slime spawn list. Tho if people generally prefer me keep it in, i'll keep the big roburger in :)

SteelSlayer (approved):

PacifistDalek:
As a follow up, I still think the big roburger should be removed, as it functionally is just the same as multiple roburgers (which aren't that hard to make). There are other weird/interesting burgers with special effects on other codebases that we could port if we wanted more unique silver slime foods.

variableundefined (approved):

PR #16826, opened 2021-09-28
Refactors Item Blood Overlays

Fox-McCloud:
Refactors how item blood overlays are handled.

Previously, this was kinda awful. We had an icon cache for every blood overlay for every item, but we also had another image on every single item that needed a blood overlay stored, locally on it, as an image---why? Just for color.

This isn't really necessary and can be handled better---and we can do so for less CPU.

What this does:
- Makes blood overlays be based on filters, rather than rendered for each item, cached, and re-used and also held as an image on the item itself.
- Renames instances of blooed-related arg names of `color` to `b_color`.
  - Having `color` as an arg isn't great because it's a pre-existing var on all atoms.
- Cleans up some improper usage of `blood_color` and refactors how it's used--it's now incorporated into the `/obj/item` level of `add_blood` 
- Greatly simplifies blood overlay adding/removing code.

The fact this is filter based now means that it's a lot easier to add/remove things and it pushes the processing (hypothetically) to be client side as opposed to server side. This does save CPU compared to the original method, but not much. The only downside is that due to how filters are utilized (priority lists), this uses up a little less than twice the memory of the original method and about twice the iterim method I originally went with in my original commit here. That said, you're talking 5 MB total to add a blood overlay to every single item in the world vs 2.4 MB (iterim method in first commit) or 2.8 MB (the version prior to this PR).

System works fine:

img

Not CL as this doesn't really impact the player experience.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -658,7 +658,7 @@
 GLOBAL_LIST_EMPTY(blood_splatter_icons)
 
 /atom/proc/blood_splatter_index()
-	return "\ref[initial(icon)]-[initial(icon_state)]"
+	return "\ref[initial(icon)]-[initial(icon_state)]-[blood_color]"
 
 //returns the mob's dna info as a list, to be inserted in an object's blood_DNA list
 /mob/living/proc/get_blood_dna_list()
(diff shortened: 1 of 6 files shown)


dearmochi:
I'm still not a fan of this method. It might be good to, instead of generating blood overlays on the fly with icons, just adding an overlay to the item with an alpha mask filter and letting the client do the filtering out of transparent parts.

Fox-McCloud:
Used @dearmochi's idea, along with some of TG's idea combined with my own to rewrite this.

code/game/atoms.dm
```diff
@@ -726,78 +723,65 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
-			blood_splatter_icon.Blend(icon('icons/effects/blood.dmi', "itemblood"), ICON_MULTIPLY) //adds blood and the remaining white areas become transparant
-			blood_splatter_icon = fcopy_rsc(blood_splatter_icon)
-			GLOB.blood_splatter_icons[index] = blood_splatter_icon
-
-		blood_overlay = image(blood_splatter_icon)
-		blood_overlay.color = color
-		add_overlay(blood_overlay)
+		var/list/params = GLOB.blood_splatter_icons["[blood_color]"]
+		if(!params)
+			params = layering_filter(icon = icon('icons/effects/blood.dmi', "itemblood"), color = blood_color, blend_mode = BLEND_INSET_OVERLAY)
```
Fox-McCloud, line 782: This actually saves a bit of memory and CPU over just doing `add_filter("blood_splatter", 1,layering_filter(icon = icon('icons/effects/blood.dmi', "itemblood"), color = blood_color, blend_mode = BLEND_INSET_OVERLAY))`, but not much--still, this is kinda a micro-optimization to begin with....

code/game/atoms.dm
```diff
@@ -726,78 +723,65 @@ GLOBAL_LIST_EMPTY(blood_splatter_icons)
...
+			GLOB.blood_splatter_icons["[blood_color]"] = params
+		add_filter("blood_splatter", 1, params)
```
Fox-McCloud, line 784: If we were to directly just manipulate `filters`, then this would save even more CPU (0.07 vs 0.23) and the memory impact would be something like 0.1 MB.

That said, the filter system we have necessitates keeping a list of filters and it involves copying lists---so the end result is that the vast increase in CPU and memory usage is due to record keeping. Very sad.

Fox-McCloud:
went ahead and removed the blood changing colors to cut down on list operations and ensure it's as consistent with old behavior as possible.

code/game/atoms.dm
```diff
@@ -657,9 +657,6 @@
 
 GLOBAL_LIST_EMPTY(blood_splatter_icons)
 
-/atom/proc/blood_splatter_index()
```
AffectedArc07, line 660: What the fuck

AffectedArc07 (approved):

PR #16831, opened 2021-09-29
Factory protected disk, block some traits for strange seeds

botka4aet:
## What Does This PR Do
- Strange seeds can't get some traits - new var and check
- Disks with recorded genes. Factory protected(you can't overwrite it)

## Why It's Good For The Game
- Factory disk - full rewrite protect. But it can still be stolen(traitor target if disk exists?)
- Block some traits for strange seeds so you can get it only from proper plant(mutate or find)

## Images of changes
image

## Changelog
:cl:
add: Factory plant disk - full rewrite protect
tweak: Strange plants can`t get prohibited traits
/:cl:

### DIFF
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -504,9 +504,6 @@
 	..()
 	update_name() //Update names only for filled disks
 
-/obj/item/disk/plantgene/full/noreact
-	gene = new /datum/plant_gene/trait/noreact
-
 /*
  *  Plant DNA Disks Box
  */
--- a/code/modules/hydroponics/gene_modder.dm
+++ b/code/modules/hydroponics/gene_modder.dm
@@ -484,6 +484,9 @@
 		name = "plant data disk"
 
 /obj/item/disk/plantgene/attack_self(mob/user)
+	if(read_only == 2)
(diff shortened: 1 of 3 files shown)


AffectedArc07:
So you remade a PR on the same branch with the same screenshot as last time and not even editing it
image


As for the disk.... what is the point?

botka4aet:
> So you remade a PR on the same branch with the same screenshot as last time and not even editing it image
> 
> As for the disk.... what is the point?

Screenshot for example. It's shown that code works:
- disk name generates after creation
- notice about protection
Point - disks with protection, that can't be overwrited accidentally or specially
Same as golem design disk at lavaland.

AffectedArc07:
> Point - disks with protection, that can't be overwrited accidentally or specially
> Same as golem design disk at lavaland.


Ok but this disk isnt used for anything? Its just an addition for the sake of it, its dead code. 

The `strange_seed` var also makes no sense as its never overridden anywhere else.


This PR makes zero change to the game and just adds dead vars that are never used anywhere else.

Fox-McCloud:
We don't need this---and this comes across as an incremental thing of "do this first so other things can get re-added later".

👎

AffectedArc07:
We’re closing this. Further attempts to remake this PR (or ones with the same intention) will result in a repoban.

FreeStylaLT:
image

Issue #16468, opened 2021-08-01
North star gloves do not work with Chef's CQC - regardless of whether you're outside the kitchen/bar


AzuleUtama:
**Issue Description**:

If you buy north star gloves as a chef, you can't actually use them at all since the limited CQC they get overrides it regardless of whether they are actually in the kitchen or bar.

**What did you expect to happen**: 

It to work normally unless you were in either of those locations.

**What happened instead**:

Punching was at normal speed while inside arrival maintenance.

**Why is this bad/What are the consequences**:

Means chefs are unable to use this item at all if they buy it.

**Steps to reproduce the problem**:

1. Start the round or join it as a chef.
2. Spawn in / Buy north star gloves.
3. Try to punch.
4. Look in shame at your pitiful punching speed.

**When did the problem start happening**:

Noticed it last round after buying it for a gimmick and finding out it didn't work. Admins online found that the chef CQC was the cause and had it refunded.

**Extra information**:

PR #16838, opened 2021-09-30
Makes disabled martial arts not block north-star. Such as chef CQC outside of the kitchen

### LINKED ISSUE #16816: Chefs cannot use gloves of the north star
**Issue Description**:
Chefs do not benefit from the gloves of the north star attack speed bonus, regardless of where they are.

When outside the areas in which a chef has CQC, therefor using normal punches it should be boosted by the gloves.

Gloves of the north star were made to not work with martial arts, however since the chef only has martial arts in a limited area, he should be able to benefit from the bonus while he is within the remaining 95% of the station where he does not have martial arts.

**What happened instead**:
The gloves simply do not work regardless of where you are.

It makes an already underused and niche item unusable for chefs by default

Simply enter the game as a chef and put the gloves on, they will have no effect whether you are in the kitchen or not

noticed today

**Extra information**:

farie82:
## What Does This PR Do
Fixes #16816
Fixes: #16468
Makes martial arts not block the north star gloves when the user can't use the martial arts.
Tested and the damage boost is not applied when the user can't use the martial arts. So it is just like a normal punch

## Why It's Good For The Game
Bug/weirdness fix. Not sure if it is considered a bug but it is weird at least.

## Changelog
:cl:
tweak: Chefs can use the north star gloves when outside of the kitchen as any normal crew can. Disabled martial arts don't affect the north star gloves anymore
/:cl:

### DIFF
--- a/code/modules/clothing/gloves/miscellaneous.dm
+++ b/code/modules/clothing/gloves/miscellaneous.dm
@@ -165,12 +165,10 @@
 /obj/item/clothing/gloves/fingerless/rapid/Touch(mob/living/target, proximity = TRUE)
 	var/mob/living/M = loc
 
-	if(M.a_intent in accepted_intents)
-		if(M.mind.martial_art || HAS_TRAIT(M, TRAIT_HULK))
-			M.changeNext_move(CLICK_CD_MELEE)//normal attack speed for hulk, CQC and Carp.
-		else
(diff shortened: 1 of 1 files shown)


AffectedArc07 (approved):
Fuck it its a fix

PR #16839, opened 2021-09-30
Embedded objects stay embedded on dismembered limbs. Can be removed by sharp objects

### LINKED ISSUE #16817: Embed'd objects are deleted if the limb they are inside is removed
**Issue Description**:
objects that can embed when thrown will be deleted if the limb they are in is removed before it falls out or is removed through surgery. Even reattached the severed limb afterwards will not recover the embed'd item.

the object should either remain in the limb, or fall to the ground upon limb removal.

**What happened instead**:
The embed'd object is simply deleted.

This mostly affects antags that use throwing stars alongside powerful sharp weapons such as an esword, or explosions.
Its a highly specific event but I feel that throwing stars are already hard enough to recover as it is, having the risk of them vanishing isn't fun and ICly makes no sense for them to be deleted from the limb.

Embed any object into a person's leg/hand/foot/arm/head and then sever the limb. In the unlikely event that the object that embeds also severs the limb from the throw, it will still be deleted.

Noticed it a couple days ago

**Extra information**:

farie82:
## What Does This PR Do
Makes embedded objects stay embedded in dismembered limbs. Instead of them seemingly disappearing. (in practice they were just made part of the limb forever without a method of getting it out).
An examine text is added to dismembered limbs that have embedded objects in them.
You can remove them using a sharp object. Like how you remove organs from a head.

I've decided to make them stay embedded to avoid surgery shortcuts where you just amputate the limb and reattach it to remove embedded objects. It could be changed to just drop on the floor if that is desired

Fixes #16817
Also fixes a very rare edge case where embedding an organ into a limb would make it insert itself into said limb when it got reattached. Doubt this ever occurred but it was funny to find.

Also improves the safety of the embedded list.

## Why It's Good For The Game
Having items be removed from the game is a bad thing in general.

## Changelog
:cl:
tweak: Embedded items in limbs now will show this on the organ when examined
tweak: Embedded items in limbs are now removable from dismembered limbs using a sharp object
fix: Embedded items won't disappear when the limb they are in gets dismembered
/:cl:

### DIFF
--- a/code/modules/surgery/organs/organ_external.dm
+++ b/code/modules/surgery/organs/organ_external.dm
@@ -97,6 +97,14 @@
 
 	return ..()
 
+/obj/item/organ/external/examine(mob/user)
+	. = ..()
+
+	var/list/embedded_names = list()
+	for(var/obj/item/I in embedded_objects)
+		embedded_names += "[I]"
+	if(length(embedded_names))
+		. += "<span class='warning'>You can see [english_list(embedded_names)] embedded.</span>"
 
 /obj/item/organ/external/update_health()
 	damage = min(max_damage, (brute_dam + burn_dam))
@@ -571,6 +579,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 			"<span class='notice'>You begin to cut open [src]...</span>")
 		if(do_after(user, 54, target = src))
 			drop_organs(user)
+			drop_embedded_objects()
 	else
 		return ..()
 
@@ -702,8 +711,10 @@ Note that amputating the affected organ does in fact remove the infection from t
 		victim.splinted_limbs -= src
 
 	for(var/obj/item/I in embedded_objects)
-		embedded_objects -= I
+		UnregisterSignal(I, COMSIG_MOVABLE_MOVED) // Else it gets removed from the embedded list
 		I.forceMove(src)
+		RegisterSignal(I, COMSIG_MOVABLE_MOVED, .proc/remove_embedded_object)
+
 	if(!owner.has_embedded_objects())
 		owner.clear_alert("embeddedobject")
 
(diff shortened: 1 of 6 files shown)


code/modules/surgery/organs/organ_external.dm
```diff
@@ -781,14 +792,29 @@ Note that amputating the affected organ does in fact remove the infection from t
...
+
+/obj/item/organ/external/proc/drop_embedded_objects()
+	var/turf/T = get_turf(src)
+	for(var/obj/item/I in embedded_objects)
+		remove_embedded_object(I)
+		forceMove(T)
+
+/obj/item/organ/external/proc/add_embedded_object(obj/item/I)
+	embedded_objects += I
+	I.forceMove(owner)
```
SteelSlayer, line 807: Do you think we could just forceMove this into the organ itself? Code that deals with embedded objects seems to just check the `embedded_objects` list on the organ, and never looks inside the human itself. 

If this were to be changed, we can also remove the code from line 713 to 716 of the `/obj/item/organ/external/remove` proc.

farie82: I don't see an issue with it myself. However, I believe that organs and such got removed from external organs their content. Though I do not remember the full reason for that.
@Fox-McCloud could steels suggestion be implemented without issues or?

marlyn-x86: If I remember correctly [organs exist in nullspace while inside of a humanoid](https://github.com/ParadiseSS13/Paradise/blob/master/code/modules/surgery/organs/organ_external.dm#L117). So, you wouldn't want to move the item inside the organ since that would mess up anything that relied on location or coordinates.

SteelSlayer (approved):

dearmochi (approved):

PR #16844, opened 2021-10-01
Polishes up and fixes bugs with the syndicate teleporter

Qwertytoforty:
## What Does This PR Do

Removes the capitalization of the syndicate teleporter, did not look good, most items are not capitalized like that.

Fixes syndicate teleporters from vanishing from existence when empd in someones bag, instead of teleporting the user. Now, if the teleporter is not in a person directly (hands, pockets) it will teleport all living mobs on the turf, following the empd teleporter rules. If no mobs are found, as before the teleporter will teleport out of existence.

Makes it so teleporters can be forced teleported even if it has no charge via EMP, making ion rifles even more dangerous to teleporter users, as they can always be forced to move.

As part of the above teleport all living mobs, the teleporter is no longer restricted to carbon mobs, and could now be used by cyborgs, if it was put in their module list.

## Why It's Good For The Game

Strange looking item names are bad.
Teleporters vanishing from bags when empd instead of teleporting the user is bad, as such now if the teleporter is in someones bag when empd, it will teleport the user (and anyone else on the turf) rather than vanishing. EMP always working on the teleporter is good, it reinforces the downside, rather than it only working if the teleporter has charge.

## Changelog
:cl:
tweak: Tweaked the name from "Syndicate teleporter" to "syndicate teleporter"
fix: Fixed the syndicate teleported teleporting out of existence when in someones bag and empd
tweak: Syndicate teleporters will teleport everyone on the turf when empd and not in someones hands / pockets, rather then teleporting out of existence due to a bug. If no one is on the turf, then it teleports out of existence.
tweak: Syndicate teleporters will now always dangerously teleport the user when empd, even if the teleporter has no charge left.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/traitordevices.dm
+++ b/code/game/objects/items/devices/traitordevices.dm
@@ -170,7 +170,7 @@ effective or pretty fucking useless.
 		GLOB.active_jammers -= src
 
 /obj/item/teleporter
-	name = "\improper Syndicate teleporter"
+	name = "syndicate teleporter"
 	desc = "A strange syndicate version of a cult veil shifter. Warrenty voided if exposed to EMP."
 	icon = 'icons/obj/device.dmi'
 	icon_state = "syndi-tele"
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/traitordevices.dm
```diff
@@ -208,32 +208,40 @@ effective or pretty fucking useless.
...
 			attempt_teleport(user, TRUE)
-		else
-			visible_message("<span class='danger'>[src] activates and blinks out of existence!</span>")
-			do_sparks(2, 1, src)
-			qdel(src)
+		else //Well, it either is on a floor / locker, and won't teleport someone, OR it's in someones bag. As such, we need to check the turf to see if people are there.
+			var/turf/teleportturf = get_turf(src)
+			for(var/mob/living/user in teleportturf)
+				if(!teleported_something)
+					teleportturf.visible_message("<span class='danger'>[src] activates spoderaticaly, teleporting everyone around it!</span>")
```
SteelSlayer, line 221:
```suggestion
			var/turf/teleport_turf = get_turf(src)
			for(var/mob/living/user in teleport_turf)
				if(!teleported_something)
					teleport_turf.visible_message("<span class='danger'>[src] activates sporadically, teleporting everyone around it!</span>")
```

SteelSlayer (approved):

PR #16847, opened 2021-10-02
Updates Goonchat Loading Error Message

Sirryan2002:
## What Does This PR Do
Tells people not to make an ahelp when GoonChat doesn't load and instead directs them to use the goonchat troubleshooting guide on the wiki and ask questions on the discord #helpchat

## Why It's Good For The Game
This will help automate the troubleshooting process

## Changelog
tweak: Tweaks the Goonchat Loading Error Message
/:cl:

### DIFF
--- a/goon/browserassets/html/browserOutput.html
+++ b/goon/browserassets/html/browserOutput.html
@@ -31,7 +31,9 @@
 		<div>
 			Loading...<br><br>
 			If this takes longer than 10 seconds, it will automatically reload, multiple times if necessary.<br>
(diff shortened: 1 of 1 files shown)


Sirryan2002:
One part of this PR that I would like some commentary on. Is there a way for me just to make the wiki guide link changeable in the TOML config so that we don't have to PR if we ever need to change it?

AffectedArc07:
> One part of this PR that I would like some commentary on. Is there a way for me just to make the wiki guide link changeable in the TOML config so that we don't have to PR if we ever need to change it?

In this instance, not really. This is a static page that's displayed before the chat even pulls data from the server.

PR #16861, opened 2021-10-03
Fixes being able to install 0 cost upgrades on an energy crossbow

Qwertytoforty:
## What Does This PR Do

Fixes being able to install 0 cost kinetic accelerator upgrades on energy crossbows, by checking for max_mod_capacity. If max_mod_capacity is false, like on crossbows, modkits can not be installed or removed.

## Why It's Good For The Game

This is bad.
image
You should not be able to do that. Yes, that is an energy crossbow.

Also lets people potentialy make kinetic accelerators in the future with upgrades pre-installed that can not be removed.

## Changelog
:cl:
fix: Fixes being able to install zero cost KA modkits to Energy Crossbows.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/kinetic_accelerator.dm
+++ b/code/modules/projectiles/guns/energy/kinetic_accelerator.dm
@@ -21,8 +21,6 @@
 
 	var/max_mod_capacity = 100
 	var/list/modkits = list()
-	/// Can upgrades be installed on the gun?
-	var/modable = TRUE
 
 	var/recharge_timerid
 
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/guns/energy/kinetic_accelerator.dm
```diff
@@ -29,21 +31,23 @@
 /obj/item/gun/energy/kinetic_accelerator/examine(mob/user)
 	. = ..()
 	if(in_range(user, src))
-		if(max_mod_capacity)
+		if(modable)
```
FlattestGuitar, line 34: This variable is completely unnecessary. Just use `if(max_mod_capacity)` like it was used here.

FlattestGuitar (changes requested):

code/modules/projectiles/guns/energy/kinetic_accelerator.dm
Qwertytoforty: Fair point. I figured making a modable variable, rather than max_mod_capacity being zero would work better, but in the end it is an unnecessary variable when max being zero does the same. One moment.

code/modules/projectiles/guns/energy/kinetic_accelerator.dm
```diff
@@ -36,14 +36,16 @@
...
+	if(istype(I, /obj/item/borg/upgrade/modkit) && max_mod_capacity)
 		var/obj/item/borg/upgrade/modkit/MK = I
 		MK.install(src, user)
 	else
 		return ..()
 
 /obj/item/gun/energy/kinetic_accelerator/crowbar_act(mob/user, obj/item/I)
 	. = TRUE
+	if(max_mod_capacity == 0)
+		return
```
AffectedArc07, line 48:
```suggestion
	if(!max_mod_capacity)
		return
```

AffectedArc07 (approved):

PR #16864, opened 2021-10-03
Airtights turf changes

### LINKED ISSUE #2913: Directional Windows + Atmos

FlattestGuitar:
## What Does This PR Do
Fixes #2913.

A turf change slurped air from all adjacent turfs even if they were not atmos-adjacent.

## Changelog
:cl:
fix: changing a turf (for example by adding/removing plating) will no longer create atmos leaks
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -283,10 +283,10 @@
 		var/asleep = 0
 		var/ab = 0
 		var/atemp = 0
+
 		var/turf_count = 0
 
-		for(var/direction in GLOB.cardinal)//Only use cardinals to cut down on lag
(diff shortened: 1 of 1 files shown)


Fox-McCloud:
probably fine, though I spy some other issues with this proc---but uh, atmos in general needs an overhaul....as to turfs 😔

PR #16870, opened 2021-10-04
Unpowered Suit Storage Units can now be attacked

### LINKED ISSUE #13340: SSU cannot be destroyed when unpowered
**Issue Description**:
Potential oversight: A Suit Storage Unit cannot be attacked whilst the room it is in is depowered. Harm intent is ignored, leaving the user with the message 'The unit is not operational.' This makes SSU impossible to be forcefully deconstructed during a power outage or other events which cause power loss in a room with a SSU.

The player expected to be able to break the SSU and get the contents inside.

**What happened instead**:
The player, despite using an energy sword on harm intent, repeatedly received the message 'The unit is not operational.'

An antagonist, or a player in urgent need of a SSU's contents, will not be able to retrieve the SSU's contents when the room is not powered.

1. Have a SSU in an unpowered room.
2. Have a weapon, and be on any intent.
3. Attempt to attack the SSU.
4. (optional) For comparison, attack the SSU after powering the room, and note that you can successfully destroy the SSU.

Only noticed today. Unsure if/when anything has caused it.

lewcc:
## What Does This PR Do
Fixes #13340 where suit storage units were unable to be attacked if unpowered, instead yielding a message about the unit not being operational. Now, if you aren't in help intent, it'll be hit with whatever you're holding.

## Why It's Good For The Game

Especially now that SSUs drop their contents on breaking, this means that suits won't be denied just because an area is unpowered.

## Changelog
:cl:
fix: Unpowered Suit Storage Units can now be attacked
/:cl:


code/game/machinery/suit_storage_unit.dm
```diff
@@ -312,6 +312,8 @@
 	if(shocked)
 		if(shock(user, 100))
 			return
+	if (user.a_intent != INTENT_HELP)
+		return ..()
 	if(!is_operational())
 		if(panel_open)
 			to_chat(usr, "<span class='warning'>Close the maintenance panel first.</span>")
```
SabreML, line 319: This might work better under the `if(!is_operational())` check instead, that way it works the same as normal when it's powered, and checks the attack intent when it's unpowered.
Currently it only allows you to interact with it while you're in help intent, which *does* work, but it could cause some confusion:

https://user-images.githubusercontent.com/57483089/135884335-c86c8b15-125c-498a-89a2-69d0140cabcd.mp4

```suggestion
	if(!is_operational())
		if(user.a_intent != INTENT_HELP)
			return ..()
		if(panel_open)
			to_chat(usr, "<span class='warning'>Close the maintenance panel first.</span>")
```

lewcc: Gotcha. I was trying to figure out if there was a typical "standard" of sorts for how interactions between intents should work, and seeing its current behavior, that makes a lot of sense

PR #16873, opened 2021-10-04
Ghosts can no longer shuffle cards

### LINKED ISSUE #13407: ghosts can shuffle cards
**Issue Description**:

Ghosts can shuffle cards

shouldn't be able to

**What happened instead**:

nothing to happen

ghosts can get crew's attention by spamming shuffling, also show's ghosts name

ghost over deck, shuffle

forever
**Extra information**:

PacifistDalek:
## What Does This PR Do
Fixes #13407

## Changelog
:cl:
fix: Ghosts can no longer shuffle cards
/:cl:

### DIFF
--- a/code/modules/games/cards.dm
+++ b/code/modules/games/cards.dm
@@ -208,11 +208,12 @@
 
 /obj/item/deck/proc/deckshuffle()
 	var/mob/living/user = usr
-	if(cooldown < world.time - 5 SECONDS)
-		cards = shuffle(cards)
-		user.visible_message("<span class='notice'>[user] shuffles [src].</span>")
-		playsound(user, 'sound/items/cardshuffle.ogg', 50, 1)
-		cooldown = world.time
+	if(!isobserver(user))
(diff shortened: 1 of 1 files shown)


code/modules/games/cards.dm
```diff
@@ -208,11 +208,12 @@
 
 /obj/item/deck/proc/deckshuffle()
```
AffectedArc07, line 209: Change this on the `verb` level that invokes it.

AffectedArc07 (changes requested):

Darkmight9:
Going off of what was said in Coding Chat I think this could make for a fun feature! Maybe it could be part of the ghost's spook ability 🤔 ?

PacifistDalek:
Simple mobs can still shuffle cards, but I'm inclined to keep that in because the idea of Poker Dealer Ian is too appealing (plus it is the kind of trick you could teach certain animals in real life)

PR #16876, opened 2021-10-04
Moves pAI saves to the DB

AffectedArc07:
## What Does This PR Do
This PR moves pAI saves to the DB instead of on-disk savefiles. File IO is notoriously awful in BYOND, and we should be using the DB for all data storage unless absolutely necessary. 

pAI saves being savefiles is dumb and this should have been converted long ago.

This PR also includes a python script to read the existing savefiles in and dump them to the DB. I have tested this with 3000 savefiles and it seems to work fine.

## Why It's Good For The Game
See above note on why pAI saves being on disk is dumb

## Changelog
:cl:
tweak: Refactored how pAI saves are stored. Your data should still be there, though there is a chance the pAI description did not migrate properly. 
/:cl:

### DIFF
--- a/SQL/paradise_schema.sql
+++ b/SQL/paradise_schema.sql
@@ -597,5 +597,18 @@ CREATE TABLE `2fa_secrets` (
 	`date_setup` DATETIME NOT NULL DEFAULT current_timestamp(),
 	`last_time` DATETIME NULL DEFAULT NULL,
 	PRIMARY KEY (`ckey`) USING BTREE
-)
-COLLATE='utf8mb4_general_ci' ENGINE=InnoDB;
+) COLLATE='utf8mb4_general_ci' ENGINE=InnoDB;
+
+--
+-- Table structure for table `pai_saves`
+--
+CREATE TABLE `pai_saves` (
+	`id` INT(11) NOT NULL AUTO_INCREMENT,
+	`ckey` VARCHAR(50) NOT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`pai_name` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`description` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`preferred_role` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`ooc_comments` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	PRIMARY KEY (`id`) USING BTREE,
+	UNIQUE INDEX `ckey` (`ckey`) USING BTREE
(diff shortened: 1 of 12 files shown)


SQL/paradise_schema.sql
```diff
@@ -597,5 +597,18 @@ CREATE TABLE `2fa_secrets` (
...
+--
+-- Table structure for table `pai_saves`
+--
+CREATE TABLE `pai_saves` (
+	`id` INT(11) NOT NULL AUTO_INCREMENT,
+	`ckey` VARCHAR(50) NOT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`pai_name` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`description` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`preferred_role` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
+	`ooc_comments` LONGTEXT NULL DEFAULT '0' COLLATE 'utf8mb4_general_ci',
```
AffectedArc07, line 611: These should default `NULL`

code/modules/client/client_procs.dm
```diff
@@ -449,6 +451,7 @@
 	GLOB.directory -= ckey
 	GLOB.clients -= src
 	QDEL_NULL(chatOutput)
+	QDEL_NULL(pai_save)
```
AffectedArc07, line 454: GC for the GC god

code/modules/client/client_defines.dm
```diff
@@ -114,6 +114,9 @@
 	/// Is the client watchlisted
 	var/watchlisted = FALSE
 
+	/// Client's pAI save
+	var/datum/pai_save/pai_save
```
AffectedArc07, line 118: Someone (who can do this in a PR that isnt mine) can make this in a PR


Add a tab to settings to bring up the user's own pAI save and let them edit it there, since these datums now exist always.

tools/pai_savefile_converter/pai_savefile_converter.py
```diff
@@ -0,0 +1,132 @@
...
+
+for entry in all_saves:
+    if entry["name"]:
+        entry["name"] = html.unescape(entry["name"])
+    if entry["description"]:
+        entry["description"] = html.unescape(entry["description"])
+    if entry["role"]:
+        entry["role"] = html.unescape(entry["role"])
+    if entry["comments"]:
+        entry["comments"] = html.unescape(entry["comments"])
```
AffectedArc07, line 124: Could probably make this work with the above logic of `entry["name"] = html.unescape(entry["name"]) if entry["name"]`

Will try later

AffectedArc07: Update: You cant

PR #16880, opened 2021-10-05
Fix fireaxe cabinet not dropping the axe when destroyed

### LINKED ISSUE #6845: Shooting fireaxe cabinet deletes cabinet and fireaxe
**Problem Description**:
Shooting at the fire axe cabinet deletes it, and the fireaxe contained within.

Expected it to break, drop the fireaxe to the ground... or nothing at all..

**What happened instead**:
It just straight up deleted the cabinet and the fireaxe.

There are only 2 fireaxes on the map. Straight deleting one of them seems silly.

Equip gun, I chose a .357.... shot the cabinet a few times on the bridge. Despair

lewcc:
## What Does This PR Do

Fixes #6845.
The fireaxe cabinet should now properly add the fireaxe to itself on initialization, rather than in its instance definition. It seems like it was somehow getting deleted alongside the object on deletion, rather than treated as a contained object.

## Why It's Good For The Game
As noted in the initial issue, losing an item of which only two exist on the station (for box, at least) is not fun. Especially if you left your multitool in your other pants.

## Images of changes
uh

## Changelog
:cl:
fix: Fireaxe cabinets should now drop a held fireaxe
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/fireaxe.dm
+++ b/code/game/objects/structures/crates_lockers/closets/fireaxe.dm
@@ -2,7 +2,7 @@
 /obj/structure/closet/fireaxecabinet
 	name = "fire axe cabinet"
 	desc = "There is small label that reads \"For Emergency use only\" along with details for safe use of the axe. As if."
-	var/obj/item/twohanded/fireaxe/fireaxe = new/obj/item/twohanded/fireaxe
+	var/obj/item/twohanded/fireaxe/fireaxe
 	icon_state = "fireaxe1000"
 	icon_closed = "fireaxe1000"
 	icon_opened = "fireaxe1100"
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/closets/fireaxe.dm
```diff
@@ -15,6 +15,11 @@
 	locked = TRUE
 	var/smashed = FALSE
 
+/obj/structure/closet/fireaxecabinet/populate_contents()
+	fireaxe = new/obj/item/twohanded/fireaxe(src)
+	update_icon()	// So its initial icon doesn't show it without the fireaxe
+	. = ..()
```
farie82, line 21:
```suggestion
```
No need to call the parent here

AffectedArc07: Remove this parent call

AffectedArc07 (changes requested):

farie82 (approved):

PR #16881, opened 2021-10-05
Transformation diseases do the jobban check properly

### LINKED ISSUE #12700: Roboburgers ignore silicon bans
**Issue Description**:
Eating a roboburger lets you become a borg while borg banned

for the borg to get offered to ghosts or to be gibbed or something?

**What happened instead**:
player was borged normally

lets players bypass borg bans

be silicon banned
munch roboburger

No idea

**Extra information**:

Vi3trice:
## What Does This PR Do
Transformation diseases had a check for jobbans that would never be true due to passing a path to `jobban_isbanned` which expects a string or defined role. Roles are now defined for diseases if need be.

The check will kill if the player is jobbanned, as it was originally coded that way. Trying to pass it onto a ghost instead would have been a more significant rework due to the mix of special roles and silicon.

## Why It's Good For The Game
Fix #12700, and possibly other issues. Ban dodging bad.

## Images of changes
TransformDeath

## Changelog
:cl:
fix: Transformation diseases no longer bypass jobbans
/:cl:

### DIFF
--- a/code/datums/diseases/transformation.dm
+++ b/code/datums/diseases/transformation.dm
@@ -17,6 +17,7 @@
 	var/list/stage5 = list("Oh the humanity!")
 	var/transformation_text = null
 	var/new_form = /mob/living/carbon/human
+	var/job_role = null
 
 /datum/disease/transformation/stage_act()
 	..()
@@ -40,7 +41,7 @@
 	if(istype(affected_mob, /mob/living/carbon) && affected_mob.stat != DEAD)
 		if(stage5)
 			to_chat(affected_mob, pick(stage5))
-		if(jobban_isbanned(affected_mob, new_form))
+		if(jobban_isbanned(affected_mob, job_role))
 			affected_mob.death(1)
 			return
 		if(affected_mob.notransform)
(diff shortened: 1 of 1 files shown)


code/datums/diseases/transformation.dm
```diff
@@ -40,7 +41,7 @@
 	if(istype(affected_mob, /mob/living/carbon) && affected_mob.stat != DEAD)
 		if(stage5)
 			to_chat(affected_mob, pick(stage5))
-		if(jobban_isbanned(affected_mob, new_form))
+		if(jobban_isbanned(affected_mob, job_role))
 			affected_mob.death(1)
```
SabreML, line 45: Could you also change this to not have the `gibbed` argument? Currently it misses a lot of the standard 'death' things because of it: (Mob doesn't fall over, no deathgasp, etc.)
gibbed

Without the `gibbed` argument:
normal

This was in here already of course, but since you're making rest of the system actually work correctly it'd be good to do this part as well.
```suggestion
			affected_mob.death()
```

SabreML (approved):

PR #16885, opened 2021-10-05
'Invisimin' logging

SabreML:
## What Does This PR Do

- Adds database and file logging for the 'Invisimin' proc.
- Prevents admins observing and in the lobby from using the proc, since it only makes a difference on `/living` subtypes.
- Slightly refactors the proc so that the code looks nicer.

## Why It's Good For The Game

More usage data and logging for admin tools is always a good thing, and considering that this proc directly has `"Don't abuse this"` in it's description, it seems like an oversight that this is one of the only procs in which it's usage can't be tracked at all.
Even 'Big Brother Mode' is logged in multiple ways, and that one's specifically designed to be hidden from admins.

## Changelog
:cl:
tweak: Adds logging for the admin 'Invisimin' verb.
/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -362,17 +362,20 @@ GLOBAL_LIST_INIT(admin_verbs_ticket, list(
 
 	if(!check_rights(R_ADMIN))
 		return
+	if(!mob || !isliving(mob))
+		return
 
-	if(mob)
-		if(mob.invisibility == INVISIBILITY_OBSERVER)
-			mob.invisibility = initial(mob.invisibility)
-			to_chat(mob, "<span class='danger'>Invisimin off. Invisibility reset.</span>")
-			mob.add_to_all_human_data_huds()
-			//TODO: Make some kind of indication for the badmin that they are currently invisible
-		else
-			mob.invisibility = INVISIBILITY_OBSERVER
-			to_chat(mob, "<span class='notice'>Invisimin on. You are now as invisible as a ghost.</span>")
-			mob.remove_from_all_data_huds()
+	if(mob.invisibility == INVISIBILITY_OBSERVER)
+		mob.invisibility = initial(mob.invisibility)
+		mob.add_to_all_human_data_huds()
+		to_chat(mob, "<span class='danger'>Invisimin off. Invisibility reset.</span>")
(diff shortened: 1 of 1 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -362,17 +362,20 @@ GLOBAL_LIST_INIT(admin_verbs_ticket, list(
 
 	if(!check_rights(R_ADMIN))
 		return
+	if(!mob || !isliving(mob))
```
farie82, line 365:
```suggestion
	if(!isliving(mob))
```

farie82 (changes requested):

farie82 (approved):

code/modules/admin/admin_verbs.dm
```diff
@@ -362,17 +362,20 @@ GLOBAL_LIST_INIT(admin_verbs_ticket, list(
...
+		mob.invisibility = initial(mob.invisibility)
+		mob.add_to_all_human_data_huds()
+		to_chat(mob, "<span class='danger'>Invisimin off. Invisibility reset.</span>")
+		log_admin("[key_name(mob)] has turned Invisimin OFF")
+	else
+		mob.invisibility = INVISIBILITY_OBSERVER
+		mob.remove_from_all_data_huds()
+		to_chat(mob, "<span class='notice'>Invisimin on. You are now as invisible as a ghost.</span>")
+		log_admin("[key_name(mob)] has turned Invisimin ON")
+	SSblackbox.record_feedback("tally", "admin_verb", 1, "Invisimin")
```
AffectedArc07, line 378: I should point out. This isn't really logging, as its anonymised. 

Statistical logging is useful, but this isn't direct.


Not a complaint though, the verb should have this.

SabreML: The `log_admin()`s in there should do the logging part, unless I misunderstood it?

AffectedArc07: I highlighted the blackbox line, since you mentioned
`Adds database and file logging for the 'Invisimin' proc`

I was unsure if you were aware that this is anon logging in the feedback table, not an actual DB action logging system

SabreML: Ah okay, I just phrased that part of the description badly. I'll make it clearer.

PR #16895, opened 2021-10-08
Nutriment Pump Implants make you projectile vomit when EMPed

lewcc:
## What Does This PR Do
Idea emerged from a discussion in dchat; adjusts the emp behavior for nutriment plus to make you projectile vomit. This results in a minor stun from the vomiting, a significant loss in nutriment, and a temporary disabling of the implant itself so it doesn't automatically refill your stomach. 

~~IC, the pump malfunctioning is pumping your stomach with ??? anyway, and if we have a distance parameter on vomiting anyway, it seems too good not to use~~

## Why It's Good For The Game
From what I've seen, there really isn't a reason to consider not using a nutriment implant if you can get someone to put it into you, as the current only downside is that you take a little bit of toxin damage if you get EMPed.
If the stun seems too long or unnecessary with the slowdown from nutriment loss, I can easily tweak/remove it, as there is a temp slowdown from the low nutriment afterwards anyway. The nutriment also regens back to base levels pretty quickly once the EMP effect wears off, so it won't ruin you.

## Images of changes
fdgsfg

## Changelog
:cl:
tweak: Nutriment Pump/Plus implants now make you projectile vomit when EMPed.
/:cl:

### DIFF
--- a/code/modules/response_team/ert_outfits.dm
+++ b/code/modules/response_team/ert_outfits.dm
@@ -75,7 +75,7 @@
 	mask = /obj/item/clothing/mask/gas/sechailer/swat
 	cybernetic_implants = list(
 		/obj/item/organ/internal/cyberimp/eyes/hud/security,
-		/obj/item/organ/internal/cyberimp/chest/nutriment
+		/obj/item/organ/internal/cyberimp/chest/nutriment/hardened
 	)
 	belt = /obj/item/gun/energy/gun/blueshield/pdw9
 
@@ -108,7 +108,7 @@
 		)
 
 	cybernetic_implants = list(
-		/obj/item/organ/internal/cyberimp/chest/nutriment/plus,
+		/obj/item/organ/internal/cyberimp/chest/nutriment/plus/hardened,
 		/obj/item/organ/internal/cyberimp/eyes/hud/security,
 		/obj/item/organ/internal/cyberimp/brain/anti_stun/hardened,
 		/obj/item/organ/internal/cyberimp/arm/flash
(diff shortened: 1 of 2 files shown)


Denthamos:
Putting in a denial on this as there is no hardened variant for ERT/DS/SSO/Nuclear operative/SST to match the other hardened implants they get or can get.  A new version must be added for such and the roles which do have it by default adjusted to have that version.

(note currently only the Red/Gamma ERT has the nutriment implants in them by default which seems like a bit of an oversight that DS/SST/SSO does not)

Wojz:
> (note currently only the Red/Gamma ERT has the nutriment implants in them by default which seems like a bit of an oversight that DS/SST/SSO does not)

Honestly they're not gonna be on station long enough for that to matter. But, I came here to inquire about the very thing that caused Denth to object, this makes ERTs EMP vulnerable which... yeah no, so they probably would need a hardened one.

lewcc:
> Putting in a denial on this as there is no hardened variant for ERT/DS/SSO/Nuclear operative/SST to match the other hardened implants they get or can get. A new version must be added for such and the roles which do have it by default adjusted to have that version.
> 
> (note currently only the Red/Gamma ERT has the nutriment implants in them by default which seems like a bit of an oversight that DS/SST/SSO does not)

I'll add these, it isn't too much extra work. I wouldn't manually want to update DS/SST/SSO here, though, that sounds like it might be suited for another PR.

Denthamos:
Objection lifted with the existing off station roles that have the nutriment implant getting the hardened one.

variableundefined (approved):

PR #16915, opened 2021-10-11
Reworks changeling strained muscles

Qwertytoforty:
## What Does This PR Do

Moves almost all of the changeling speed leg code to a status effect so it doesn't use sleeps.

Changes the cost of speedlegs from slow stamina buildup to chemicals (by product fixes the bug where the stamina doesn't build up properly)

Makes it so you can not rapidly turn speedlegs on and off to negate the cost, by preventing it being turned off in under 3 cycles, and prevent it from being turned on when the user has genetic damage.

Lowers the dangerous stack number from > 10 to  > 6.

## Why It's Good For The Game

Right now, the changeling speed legs ability is buggy, and downsides are often ignored. In the past, people would rapidly turn the ability on and off, to avoid the buildup of stacks in the ability, allowing you to keep the rapid speed, without running into the stamina cooldown cost. Not to mention currently, the stamina buildup function of the ability is bugged, and stamina is reset every cycle.

Currently, it takes **37** cycles of the ability (which sleeps for 4 seconds, and is inconstant due to the sleeps), before the user takes enough stamina damage each "cycle" to be slowed. Even after stamina passes the normal stamina stun limit (100), the user is not stunned, and can continue moving, although changing between fast and slow due to stamina. 
Combined with a chemical to negate pain slowdown, and this ability can be always on for free.

This ability needs a nerf to make it not an always on free speed ability with no catch, so this PR changes it.

Now, once your speed legs are activated, you loose chemicals equal to (number of cycles the ability has been active * 3) every 4 seconds.

After 3 cycles, you would loose 3 + 6 + 9 chemicals, or 18. As such, while you can get a solid speed boost, it now has a cost.

You can no longer rapidly turn the ability on and off to bypass the stack buildup, you can not turn off the ability if less than 4 stacks have passed, and the ability can not be turned on when you have genetic damage.

This means that you have to pay a reasonable cost for speed, and can no longer cheese it with chemicals or turning it on and off.

Numbers may have to be tweaked to less than 3 if it is draining chemicals too fast, however a value of 2 seems too low at the moment, however live testing may change this, and I would rather not have a decimal chemical cost if possible. Fixing it to a fixed cost per cycle may also be possible as well, if we want fixed chemical cost rather than increasing as the ability is used.

## Changelog
:cl:
tweak: Changeling strained muscles ability no longer does stamina damage to the user, instead drains chemicals in an increasing amount as it is used.
tweak: Strained muscles ability can no longer be rapidly turned on and off.
/:cl:


S34NW:
While fixing bugs and code improvement is something everyone likes to see, the issue with how this works is that it's one of the few decent abilities clings have and is pretty much a default pick for a lot of cling builds. Nerfing it without reworking more about clings seems like it might just make them even weaker

datlo:
I would honestly suggest nerfing the ability as you did and making it *default* for all changelings so that we can get more points to play with other fun tools that we have to ignore to get must-haves.

variableundefined (approved):

variableundefined (approved):

